PR #143, opened 2018-06-09
Remove the * kind syntax

int-index:
The [proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0143-remove-star-kind.rst) has been accepted; the following discussion is mostly of historic interest.

---

The Haskell Report uses `*` to denote the kind of inhabited types. As we move towards DependentHaskell, it is increasingly painful to support this historical choice of name. We propose to slowly and carefully remove this syntactic oddity from the language.

[Rendered](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0143-remove-star-kind.rst)

PR #146, opened 2018-06-18
Remove TypeOperators => NoStarIsType "migration"

phadej:
This step was intended to help migration, but emprical evidence
shows it breaks more than it allows to migrate.

As StarIsType change is now merged in master, many packages broke,
which you wouldn't expect: aeson, cassava, cereal to mention some.
They are broken due use of TypeOperators for GHC.Generics code
and authors writing down the kind signatures. There * is no longer
type, which causes compilation failures.

Other packages, which I suspect would break badly are e.g.
servant, which uses TypeOperators and also a lot of kind signatures.

If the extension implication is removed, then I'm able to compile
servant-client & servant-server with only three additions
of {-# LANGUAGE NoStarIsType #-}, two in `basement` and one in `memory`
packages (they genuinely use GHC.TypeLits.*).

Also compilation of `lens` needs only few fixes: lens have one module
using TypeInType, `constraints` uses GHC.TypeLits.*.

One can argue, that the packages where the extension implication
would help, will be broken by other TypeInType related changes anyway.

TL;DR it's less work to go without TypeOperator => NoStarIsType.

### DIFF
--- a/proposals/0020-no-type-in-type.rst
+++ b/proposals/0020-no-type-in-type.rst
@@ -54,7 +54,7 @@ Motivation
 
 * If we plan to remove ``*`` from the language at some point, we should start updating
   error messages sooner than later.
-  
+
 * In truth, GHC always has ``Type :: Type``, whether you say ``-XTypeInType``
   or no. Thus, the real extension name should be ``-XPolyKinds``, because it's
   kind polymorphism that the user wants, not the always-true ``Type :: Type``.
@@ -71,7 +71,7 @@ Motivation
   The one difference between ``-XPolyKinds`` and ``-XTypeInType`` that's worth preserving
   is that the former allows easy access to the kind ``*``. The ``-XStarIsType`` extension
   is meant to preserve this difference.
-  
+
 Proposed Change Specification
 -----------------------------
 
@@ -105,9 +105,9 @@ Proposed Change Specification
 
 2. ``-XDataKinds`` would now promote GADTs and GADT constructors. This change is fully
    backward compatible; no migration would be necessary.
-      
+
 3. Two releases after this proposal is implemented, deprecate ``-XTypeInType``.
-      
+
 4. Introduce a new language extension ``-XStarIsType``, with the following behavior:
 
    a. ``-XStarIsType`` is on by default.
@@ -141,13 +141,13 @@ Effect and Interactions
   context means:
 
   1. If ``-XTypeInType`` is in effect:
-     
+
      a. If the use of ``*`` refers to ``Data.Kind.*``, then parse it as an
 	alphanumeric identifier; it means ``Type``.
      b. If ``*`` refers to some other type, it is a binary operator.
 
   2. If ``-XTypeInType`` is not in effect:
-     
+
      a. If the use of ``*`` is in a context that is syntactically understood
 	to be a kind, ``*`` is parsed as an alphanumeric identifier and means
 	``Type``.
@@ -207,7 +207,7 @@ Costs and Drawbacks
   values will have to be updated to use ``Type`` instead, as imported from ``Data.Kind``.
   This change is backward compatible to GHC 8.0. (Alternatively, they could
   use ``-XStarIsType`` and fully-qualify their uses of the binary operator ``*``.)
-  
+
 Alternatives
 ------------
 
@@ -243,10 +243,10 @@ Alternatives
      of ``*``, we'll be further from the behavior that the Report authors intended at the
      time. However, as the Reports do not specify error message text, this change does
      not bring us further from formal compliance to the letter of the Report. It would bring
-     us further from the spirit of the Report.   
-   
+     us further from the spirit of the Report.
+
 .. |star| unicode:: U+2605 .. unicode star
-   
+
 6. Currently, and in this proposal, both ``*`` and its unicode variant |star| are
    treated identically. One way to have our cake and eat it too is to follow the plan
    above for ``*`` but force |star| to always lex as an alphanumeric identifier
@@ -279,7 +279,90 @@ Alternatives
    I find both arguments compelling independently, and so I withdraw support for this
    alternative. Nevertheless, I'm keeping it in the proposal in case someone wants to
    argue in support of it.
-   
+
+Implication TypeOperators to NoStarIsType
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+One alternative is to have additional *4.f* step in introduction of ``-XStarIsType``
+
+   f. For two releases, ``-XTypeOperators`` will imply ``-XNoStarIsType``, to
+      provide a migration path for code that uses the binary operator ``*``. (After
+      two releases, this code can include ``-XNoStarIsType`` explicitly without
+      going against the three-release policy.) Users can re-enable ``-XStarIsType``
+      after ``-XTypeOperators`` is enabled if they wish.
+
+This alternative is however problematic. It's intended to help migration,
+but implementation evidence shows it causes more trouble.
+
+For example code using generics (and therefore ``-XTypeOperators``) often
+enough define helper newtypes where ``ConstraintKinds`` is needed,
+with the clause it will be broken:
+
+::
+
+  Data/Hashable/Generic.hs:116:22: error:
+      Operator applied to too few arguments: *
+      With NoStarIsType (implied by TypeOperators), ‘*’ is treated as a regular type operator.
+      Did you mean to use ‘Type’ from Data.Kind instead?
+      |
+  116 | newtype Tagged (s :: * -> *) = Tagged {unTagged :: Int}
+      |
+
+
+Another example is simple declaration (from servant):
+
+::
+
+  {-# LANGUAGE PolyKinds, TypeOperators #-}
+
+  data (:>) (a :: k) (b :: *)
+
+There is an argument that the code should be rewritten using ``Type``,
+but in that case it would be better if ``-XTypeOperators`` implied
+``-XNoStarIsType`` indefinitely:
+
+::
+
+  {-# LANGUAGE CPP #-}
+  #if __GLASGOW_HASKELL__ >= 800
+  import Data.Kind (Type)
+  #else
+  #define Type *
+  #endif
+
+  newtype Tagged (s :: Type -> Type) = Tagged {unTagged :: Int}
+  data (:>) (a :: k) (b :: Type)
+
+Without the 4.f clause, some code using ``GHC.TypeLits.*`` will need to enable
+``-XNoStarIsType`` explicitly, like
+
+::
+
+  {-# LANGUAGE CPP, KindSignatures, DataKinds, TypeFamilies #-}
+  {-# LANGUAGE TypeOperators #-}
+  #if __GLASGOW_HASKELL__ >= 805
+  {-# LANGUAGE NoStarIsType #-}
+  #endif
+
+  -- Support GHC-7.8 and GHC-7.10
+  #if __GLASGOW_HASKELL__ >= 800
+  import Data.Kind (Type)
+  #else
+  #define Type *
+  #endif
+
+  import GHC.TypeLits
+  import Data.ByteString (ByteString)
+  import Data.Coerce (coerce)
+
+  newtype Block (n :: Nat) a = Block ByteString
+
+  -- Note: in GHC-7.8 - GHC-8.4 this works even with (a :: *)
+  type family ElemSize (a :: Type) :: Nat
+
+  cast :: ((n *  ElemSize a) ~ (m * ElemSize b)) => Block n a -> Block m b
+  cast = coerce
+
 Unresolved questions
 --------------------
 
--- a/proposals/0020-no-type-in-type.rst
+++ b/proposals/0020-no-type-in-type.rst
@@ -291,12 +291,11 @@ One alternative is to have additional *4.f* step in introduction of ``-XStarIsTy
       going against the three-release policy.) Users can re-enable ``-XStarIsType``
       after ``-XTypeOperators`` is enabled if they wish.
 
-This alternative is however problematic. It's intended to help migration,
-but implementation evidence shows it causes more trouble.
-
-For example code using generics (and therefore ``-XTypeOperators``) often
-enough define helper newtypes where ``ConstraintKinds`` is needed,
-with the clause it will be broken:
+This alternative is problematic. It's intended to help migration,
+but implementation evidence shows it causes more trouble. If we consider
+``-XTypeOperators`` as the only enabled extension, then 4.f will
+indeed help migration, but there are a lot of code in the wild
+also enabling ``-XKindSignatures`` where the clause changes semantics.
 
 ::
 
@@ -334,7 +333,8 @@ but in that case it would be better if ``-XTypeOperators`` implied
   data (:>) (a :: k) (b :: Type)
 
 Without the 4.f clause, some code using ``GHC.TypeLits.*`` will need to enable
-``-XNoStarIsType`` explicitly, like
+``-XNoStarIsType`` explicitly. ``-XNoStarIsType`` is required to make
+``* :: Nat -> Nat -> Nat`` usable in the definition of ``cast``.
 
 ::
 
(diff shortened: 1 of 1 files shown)


int-index:
The difference in the amount of breakage is startling (as discussed on IRC). I will put together a Diff that reverts `TypeOperators => NoStarIsType`.

cc @RyanGlScott

goldfirere:
I'm in support of what leads to the least amount of breakage. `Type` has been available for 3 releases now, so any change we make here conforms to the goal of allowing 3 releases all to work without CPP.

int-index:
> any change we make here conforms to the goal of allowing 3 releases all to work without CPP

I don't think that's true. If we undo `TypeOperators => NoStarIsType`, then there will be code that can no longer compile without CPP: it's code that defines its own type-level binary `*`. This code will have to enable `NoStarIsType` manually, but this extension is new and must be hidden behind CPP.

So, formally speaking, we violate the three-release policy if we undo `TypeOperators => NoStarIsType`. But this is a trade-off to minimize the amount of breakage. Trying to build Hackage revealed that we'd need lots of CPP for packages that need to support GHC < 8.0

int-index:
In reality, I don't know why packages choose to support GHC < 8.0 to this day. Debian Stable has GHC 8.0.1 after all.

gbaz:
I don't think we need to hide the new extension behind CPP? Instead, we could use cabal file conditionals, no? And we could even use revisions to add such changes to existing packages to fix up prior versions already on hackage.

int-index:
I have come up with a workaround for this issue. With my workaround in place, this code will compile on all GHC versions, with `-XStarIsType` or without it:

```
{-# LANGUAGE TypeOperators, TypeFamilies #-}

module X (type (X.*)) where

type family (*) a b where { (*) a b = Either b a }
```

On GHC 8.6 with `-XStarIsType` it will produce the following warnings:

```
[1 of 1] Compiling X                ( X.hs, interpreted )

X.hs:5:14: warning:
    Desugaring ‘*’ in a binder position as a type operator despite StarIsType.
    NB. To use (or export) this operator in modules with StarIsType,
        including the definition module, you must qualify it.
  |
5 | type family (*) a b where { (*) a b = Either b a }
  |              ^

X.hs:5:30: warning:
    Desugaring ‘*’ in a binder position as a type operator despite StarIsType.
    NB. To use (or export) this operator in modules with StarIsType,
        including the definition module, you must qualify it.
  |
5 | type family (*) a b where { (*) a b = Either b a }
  |                              ^
Ok, one module loaded.
```

This means we will formally in compliance with the three-release policy. The workaround costs us 3 lines of code for its logic and 20 more for the warning. In `checkTyClHdr` I added:

```
+    go _ (HsParTy _ (L l (HsStarTy _ isUni))) acc ann fix
+      = do { warnStarBndr l
+           ; let name = mkOccName tcClsName (if isUni then "★" else "*")
+           ; return (L l (Unqual name), acc, fix, ann) }
```

int-index:
https://phabricator.haskell.org/D4865

phadej:
My rationale is, that changing `TypeOperators` to imply `NoStarIsType`, but then revert that **in two releases** doesn't make sense to me in the light how much it breaks in between.

If https://prime.haskell.org/wiki/Libraries/3-Release-Policy is the policy we are using, IMO it shouldn't be considered in this case (too strongly):

> The primary scope of this policy is the Prelude, as it affects the largest number of library authors. For other modules of base and other basic libraries, it is desirable to follow the guideline, if possible.

phadej:
And given we also have #143 , then turning `NoStarIsType` again in N releases (7 years) feels silly.

An alternative is to keep the extension implication `TypeOperators => NoStarIsType`, and *never* remove it.

hvr:
There's an additional aspect to consider here: Since we don't have versioned language pragmas, it's very nasty to change the semantics of a "released" language pragma.

We have language pragmas to setup the syntactical rules for a haskell source, i.e. if we declare

```hs
{-# LANGUAGE Haskell2010, TypeOperators #-}

-- ...
```

then this program is expected to have the non-libraries-specific language rules be stable across multiple GHC versions (unless `TypeOperators` was declared an experimental / subject-to-change extensions; in which case we shouldn't allow it on Hackage; see comment below).

Moreover, we use language pragma tokens as a dependency specification in cabal files to describe which features we require from a haskell compiler; so when you write

    other-extensions: TypeOperators

or

    default-extensions: TypeOperators

then this is considered a feature-based constraint on the compiler implementation by the cabal solver which allows to "test for features, not versions" since we ideally want to avoid having to enumerate each compatible compiler implementation and having to know the exact release version which provides a given language extensions the package needs.

In the past, we were rather cautious to only register language extensions in `lib:Cabal` which were considered stable enough and reasonably well-defined specification to be specified as depedency in a .cabal file uploaded to Hackage.

RyanGlScott:
I'm in support of this. It drastically reduces the amount of code breakage in the wild, and still supports the three-release policy. Two suggestions for the change to the proposal itself:

1. Currently, this simply deletes (4)(f) from the Proposed Change Specification. I'd recommend migrating this to the Alternatives section, and state why we ended up deciding against this migratory scheme.
2. Somewhere else in the proposal, it would be worth characterizing what sort of code breaks with this (new-and-improved) proposal. For example, with point (4)(f) in place, any code that simultaneously enabled `TypeOperators` and used `*` as a `Type` synonym would break. What breaks without (4)(f)? @phadej alluded to `servant`, `lens`, `memory`, etc. breaking in https://github.com/ghc-proposals/ghc-proposals/pull/146#issue-195406145, but it's unclear to me why exactly that is.

simonpj:
Question: if someone is using both `TypeOperators` and `*` to mean `Type`, will the author get any warning/deprecation messages to encourage them to replace those `*`'s with `Type`?

int-index:
The warning exists (`-fwarn-star-is-type`) but is not currently implied by `TypeOperators`. I think we need the approval of #143 to start moving in that direction...

goldfirere:
I'm afraid I don't understand the workaround in https://github.com/ghc-proposals/ghc-proposals/pull/146#issuecomment-397949257 . That workaround addresses only binding sites for `*`. I suppose, then, you're about to say that usage sites already have a workaround because all uses of `*` can be qualified. I guess then we're in the letter of the law, but I think requiring every use of `*` to be qualified is against the spirit of it. That doesn't mean I'm against this change -- I just want to be honest about the pain of supporting this change without CPP.

@hvr: Sadly, I believe that ship has sailed. Many extensions available on Hackage change their semantics between versions. (Indeed, this proposal is a downstream effect of a major change to `-XPolyKinds`.) I agree that each such instance is regrettable, but the alternatives seem worse. I'd be in strong support of versioned extensions if someone wants to design that feature. But until then, the community seems to support changing semantics of extensions if the benefits outweigh the costs.

phadej:
@RyanGlScott added few examples. I'm somewhat tired, so there might be something wrong.

@goldfirere changing semantics back and forth is silly. I see two options:
- removing implication from the proposal: *doesn't change `TypeOperators`*  (this PR)
- or have implication indefinitely: *change `TypeOperators`* once

Second alternative is related to the point @simonpj mentioned. We can eat the bullet *now*
and fix everything wanting to support pre-GHC-8.0 compilers with

```haskell
#if __GLASGOW_HASKELL__ >= 800
import Data.Kind (Type)
#else
#define Type *
#endif
```

for the code using `TypeOperators` today. 

The third, current  evolution of:

- TypeOperators  => NoStarIsType (GHC-8.6)
- TypeOperators => StarIsType "default" (GHC-8.6+2)
- default => NoStarIsType (GHC-8.6+N) #143 

is silly.

int-index:
@goldfirere  There's a reason why I called it a "workaround" and not a "solution". Yes, it only gives us formal compliance with the three-release policy, but I think this is good enough. It provides a way to write code for people who (a) need to support older releases, and (b) cannot use `-XCPP` for one reason or another. The rest of us can either drop older releases and turn on `-XNoTypeInType` or use `-XCPP`.

goldfirere:
We have a different understanding of "change semantics".

Here's my take:

* Current status: `TypeOperators` has no change in semantics. All programs with `TypeOperators` and no extensions that have changed semantics will continue to compile. `TypeOperators` will change in semantics when the `TypeOperators => NoStarIsType` implication is disabled.
* This proposal: `TypeOperators` changes semantics in 8.6. A program that uses `TypeOperators` to access a binary infix `*` will no longer compile.
* Have `TypeOperators` imply `NoStarIsType` indefinitely: No change in semantics.

What *has* changed is `KindSignatures`.

My "ship has sailed" comment wasn't meant to defend my proposal, which you interpret as changing semantics twice. (I see it as only changing semantics once, but I understand your point of view.) It was to defend your suggestion of changing the semantics in 8.6.

And, I fully agree that, with #143 in the mix, the proposed waffling is silly.

phadej:
@goldfirere thanks for your comment. Indeed the interplay with `ConstraintKinds` causes breakages, and the original proposal didn't mention it at all. I'll sleep over and amend the section in *Alternatives*.

To be sure, @int-index, @RyanGlScott and @goldfirere are in favour of this, or what else I have to address?

ekmett:
> So, formally speaking, we violate the three-release policy if we undo `TypeOperators => NoStarIsType`

Do we have a released GHC version with this implication in place?

Also the three release policy only states that there should generally be some way that you can write code that works across any 3 version window without CPP. It doesn't say that you can have every identifier you'd want in that code. Even if we have a released GHC with `TypeOperators => NoStarIsType`, reverting that doesn't preclude someone from simply using another symbol to get a piece of code working for that size window.

I'm not saying that rolling this back is or isn't a good idea, I'm just saying that the three-release policy doesn't enter the debate.

int-index:
> Do we have a released GHC version with this implication in place?

Not yet. This thread is to decide whether to include this implication in GHC 8.6

> It doesn't say that you can have every identifier you'd want in that code. 

Why? Exported identifiers are part of the public interface, so making some identifier unavailable  (by introducing new syntax or keywords) means that the code to support this interface can no longer be written

ekmett:
>> It doesn't say that you can have every identifier you'd want in that code.

> Why? 

Because by that reasoning we could simply never take any new identifiers, ever, without forcing everyone into qualified imports. e.g. Moving `Semigroup` into `Prelude` takes a couple of previously free identifiers from you.

int-index:
Alright, no one expressed opposition to removing the implication, at least so far. @nomeata what would be the procedure to get this adjustment to the proposal merged?

nomeata:
I think this can and should be treated just as any other proposal: Once it has been adequately discussed, @phadej (or anyone really) should put it before the committee. When it has been accepted, I simply merge the pull-request as is, changing the text of the previous proposal.

int-index:
I think there's a bit of an urgency aspect to it because the decision has to be made before the release of GHC 8.6 and my understanding is that the first release candidate must be made in June. There are also packages on head.hackage to update. Is it possible to use a different, accelerated procedure?

int-index:
My fear is that if we don't make the decision in time, the entire feature (implementation of the "Embrace TypeInType" proposal) will have to be removed from GHC 8.6, and this means that implementation of another proposal ("Treat type and kind variables identically in `forall`") will be delayed by one more release.

nomeata:
The process can be very quick if there is consensus. I guess I’d make @bgamari the Shepherd, then he has a natural interest in getting this decided in time for the 8.6 release :-)

phadej:
I think the proposal is discussed enough, but I'd like @RyanGlScott to comment on the alternative section. If it's ok, then proposal can be submitted.

proposals/0020-no-type-in-type.rst
```diff
@@ -285,7 +277,58 @@ Alternatives
    I find both arguments compelling independently, and so I withdraw support for this
    alternative. Nevertheless, I'm keeping it in the proposal in case someone wants to
    argue in support of it.
-   
+
+Implication TypeOperators to NoStarIsType
```
RyanGlScott: This section title reads strangely. I would suggest "Have `TypeOperators` imply `NoStarIsType`" as an alternative title. (Make sure to fix the forward reference to it in the Effects section as well.)

proposals/0020-no-type-in-type.rst
```diff
@@ -178,23 +172,21 @@ Effect and Interactions
...
-     use ``*`` as a binary operator, those uses would have to be qualified
-     with a module prefix.
+    mult :: forall (m :: Nat) (n :: Nat). Proxy m -> Proxy n -> Proxy (m * n)
+    mult _ _ = Proxy
 
-  2. Import ``Type`` from ``Data.Kind`` and change uses of ``*`` to ``Type``.
-     If they already have a ``Type`` in scope, they may have to use qualified
-     imports, etc.
+  will need to explicitly enable ``-XNoStarIsType``.
+  See `Implication TypeOperators to NoStarIsType`_ for the alternative.
```
RyanGlScott: "for an alternative."

proposals/0020-no-type-in-type.rst
```diff
@@ -285,7 +277,58 @@ Alternatives
...
+      provide a migration path for code that uses the binary operator ``*``. (After
+      two releases, this code can include ``-XNoStarIsType`` explicitly without
+      going against the three-release policy.) Users can re-enable ``-XStarIsType``
+      after ``-XTypeOperators`` is enabled if they wish.
+
+This alternative is problematic. It's intended to help migration, but
+implementation evidence shows it causes more trouble.  If the code being
+migrated only enables ``-XTypeOperators``, then 4.f will indeed be helpful, but
+there is a lot of code which also enables ``-XKindSignatures`` in which
+``-XNoStarInType`` changes the semantics. For example code from ``hashable`` library
```
RyanGlScott: "For example, this code from the `hashable` library:"

proposals/0020-no-type-in-type.rst
```diff
@@ -285,7 +277,58 @@ Alternatives
...
+
+::
+
+    newtype  Tagged  (s :: * -> *) = Tagged {unTagged :: Int}
+
+fails to compiler with the following error:
```
RyanGlScott: "would fail to compile with the following error:"

proposals/0020-no-type-in-type.rst
```diff
@@ -285,7 +277,58 @@ Alternatives
...
+
+Many other libraries in the wild simultaneously enable ``-XTypeOperators`` and use
+``*`` as a kind, including servant, aeson, cereal, cassava, and others, so
+including 4.f would break all of these libraries. While not having 4.f also
+results in some breakage, far fewer libraries in the wild break without 4.f
+than with 4.f, so this was ultimately decided against.
+
+These modules will suddenly have ``-XNoStarIsType`` in effect, meaning that
+their use of ``*`` will refer to a binary operator. These modules have a
+choice of how to proceed. They can either:
```
RyanGlScott: I would replace this whole paragraph with:

"If `TypeOperators` were to imply `NoStarIsType`, then any code which uses `*` as kind instead of a binary operator would have to migrate somehow. Two migration options are:"

phadej:
@nomeata Now this proposal is now ready.

hvr:
@int-cast 

> release of GHC 8.6 and my understanding is that the first release candidate must be made in June

I don't think that's realistic at all. Don't you rather mean an Alpha release of GHC 8.6 which would come without any guarantees re API stability (see below)?

See https://ghc.haskell.org/trac/ghc/wiki/WorkingConventions/Releases/NewSchedule which defines the stability guarantees defined as part of the new twice a year major-release-dance schedule:

> ### Pre-release stability
>
> * **Alpha** release: This is a newly cut release branch that is guaranteed to build on all Tier 1 platforms and passes fast validation. Beyond that, no guarantees: some features slated for this release may not have fully landed.
> * **Beta** release: The release has seen some testing beyond `./validate`. Only previously agreed on, stable and tested new functionality is allowed in. The focus is on bug fixing.
> * **Release candidate (RC)** release: Strictly no new functionality. Only bug fixes are allowed.

int-index:
Yes, I was probably thinking of an alpha.

mniip:
Just wanted to mention a quick workaround that works with the current (HEAD) ghc: if `StarIsType` is explicitly enabled *after* `TypeOperators` then GHC recognizes both type operators and `*` as kind.

int-index:
This is by design. However, older compilers do not know about this extension, so `{-# LANGUAGE TypeOperators, StarIsType #-}` would require CPP (or Cabal conditionals)

bgamari:
@int-index, do you intend on bringing this before the committee for consideration? If so it should be done sooner rather than later.

phadej:
@bgamari I already pinged @nomeata that this proposal is ready to be considered by committee.

nomeata:
Ah, thanks for the second ping :-)

PR #154, opened 2018-07-14
Unify type-level literals with their equivalent types

typedrat:
This proposal would bring the types/kinds and variety of type-level literals more in line with their term-level counterparts.

[Rendered](https://github.com/typedrat/ghc-proposals/blob/master/proposals/0000-unify-type-level-literals.rst)

### DIFF
--- a/proposals/0000-unify-type-level-literals.rst
+++ b/proposals/0000-unify-type-level-literals.rst
@@ -104,7 +104,7 @@ For the specific example I used to motivate this change, the implementation of t
 
 ::
 
-    type ValidBS str = (KnownSymbol str, AllValidChars str) 
+    type ValidBS str = (KnownSymbol str, AllValidChars (SymbolToString str))
 
     type family AllValidChars (xs :: String) :: Constraint where
         AllValidChars (x:xs) = If (IsValidChar x) (AllValidChars xs) (TypeError (InvalidCharError x))
--- a/proposals/0000-unify-type-level-literals.rst
+++ b/proposals/0000-unify-type-level-literals.rst
@@ -37,8 +37,6 @@ This is largely motivated by the discussion on ghc-proposals/ghc-proposals#124 i
 
 Proposed Change Specification
 -----------------------------
-The following is relative to the current behavior of ``DataKinds``.
-
 Grammar changes
 ^^^^^^^^^^^^^^^
 The lexical syntax would gain the following new case:
--- a/proposals/0000-unify-type-level-literals.rst
+++ b/proposals/0000-unify-type-level-literals.rst
@@ -91,7 +91,7 @@ A new type family will be added, ``Data.Type.Equality.Cmp``, to provide a unifor
 
 ::
 
-    type family Cmp (m :: a) (n :: a) :: Ordering
+    type family Compare (m :: a) (n :: a) :: Ordering
 
 Here, ``a`` is at least the five literal types, but there are many more implementations that could be added.
 
--- a/proposals/0000-unify-type-level-literals.rst
+++ b/proposals/0000-unify-type-level-literals.rst
@@ -54,10 +54,10 @@ The kinds of type-level literals will be as follows:
 
 ::
 
-    "foo" :: String
+    "foo" :: k1 -- Where 'k1' can be Symbol or String, if OverloadedStrings is enabled. Defaults to Symbol otherwise.
     'f' :: Char
     -123 :: Integer
-    123 :: k -- Where 'k' can be either Integer or Rational
+    123 :: k2 -- Where 'k2' can be Natural, Integer or Rational
     3.14 :: Rational
 
 While the ``Rational`` literals are the most dubious, they require very minimal changes (since ``Ratio Integer`` will now work correctly "for free") and are needed to satisfy the original impetus for this change.
@@ -68,6 +68,7 @@ The obvious alternatives are:
 
 * Use a syntactic tweak to show when a positive ``Integer`` rather than a ``Natural`` is desired. This is a poor option because it will mean that the term-level and type-level syntax will become *less* similar, not more. This goes against a key aim of the proposal.
 * Don't have pure syntax for integer literals, and instead have magical type families ``Positive (n :: Natural) :: Integer`` and ``Negative (n :: Natural) :: Integer`` to create them.
+* E*xtend things out to match how literals work at the term level, with ``FromInteger :: Natural -> a``, ``FromRational :: Rational -> a`` and ``IsString :: Symbol -> a`` type families allowing extension of these mechanisms. The latter, as implied above, would only be enabled if OverloadedStrings is used.
 
 More suggestions are welcome as this is one of the biggest discomfort points for the proposal as it stands.
 
--- a/proposals/0000-unify-type-level-literals.rst
+++ b/proposals/0000-unify-type-level-literals.rst
@@ -83,31 +83,21 @@ All of the mathematical type families in ``GHC.TypeNats`` will be generalized to
 
 ::
 
-    type family Add a (m :: a) (n :: a) :: a
-    type family Mul a (m :: a) (n :: a) :: a
-    type family Exp a b (m :: a) (n :: b) :: a
-    type family Sub a (m :: a) (n :: a) :: a
-    type family DivI a (m :: a) (n :: a) :: a
-    type family ModI a (m :: a) (n :: a) :: a
-    type family Log2I b (m :: b) :: b
+    type family (m :: a) + (n :: a) :: a
+    type family (m :: a) * (n :: a) :: a
+    type family (m :: a) ^ (n :: b) :: a
+    type family (m :: a) - (n :: a) :: a
+    type family (m :: a) / (n :: a) :: a
+    type family (m :: a) % (n :: a) :: a
+    type family Div (m :: b) (n :: b) :: b
+    type family Mod (m :: b) (n :: b) :: b
+    type family Log2 (m :: b) :: b
 
-    type (m :: a) + (n :: a) = Add a m n
-    type (m :: a) * (n :: a) = Mul a m n
-    type (m :: a) ^ (n :: b) = Exp a b m n
-    type (m :: a) - (n :: a) = Sub a m n
-    type (m :: a) / (n :: a) = DivI a m n
-    type (m :: a) % (n :: a) = ModI a m n
-    type Div (m :: b) (n :: b) = DivI b m n
-    type Mod (m :: b) (n :: b) = ModI b m n
-    type Log2 (m :: b) = Log2I b m
-
-A new type family will be added, ``Data.Type.Equality.Cmp``, using a similar alias-wrapper trick to provide a uniform interface for comparisons.
+A new type family will be added, ``Data.Type.Equality.Cmp``, to provide a uniform interface for comparisons.
 
 ::
 
-    type family CmpI a (m :: a) (n :: a) :: Ordering
-
-    type Cmp (m :: a) (n :: a) = CmpI a m n
+    type family Cmp (m :: a) (n :: a) :: Ordering
 
 Here, ``a`` is at least the five literal types, but there are many more implementations that could be added.
 
@@ -121,7 +111,7 @@ For the specific example I used to motivate this change, the implementation of t
 
     type family AllValidChars (xs :: String) :: Constraint where
         AllValidChars (x:xs) = If (IsValidChar x) (AllValidChars xs) (TypeError (InvalidCharError x))
-        AllValidChars '[]    = True ~ True
+        AllValidChars '[]    = ()
 
     type IsValidChar c = CmpChar c '\256' == LT
     type InvalidCharError c = ShowType c :<>: Text " is not a single-byte character."
--- a/proposals/0000-unify-type-level-literals.rst
+++ b/proposals/0000-unify-type-level-literals.rst
@@ -72,7 +72,7 @@ The obvious alternatives are:
 
 More suggestions are welcome as this is one of the biggest discomfort points for the proposal as it stands.
 
-To handle the legacy case, ``Symbol`` and ``Nat`` will become aliases for ``String`` and ``Natural``, respectively.
+To handle the legacy case, ``Nat`` will become an alias for ``Natural``.
 
 All of the mathematical type families in ``GHC.TypeNats`` will be generalized to be poly-kinded so that they work for these new numeric kinds, with the exception of a ``Rational`` implementation for  ``Div``, ``Mod``, and ``Log2``. ``/`` and ``%`` will be added for ``Rational``. Put more explicitly, their interfaces will be as listed in the block below, where ``a`` means any of ``Integer``, ``Natural``, or ``Rational`` and ``b`` means either ``Integer`` or ``Rational``.
 
@@ -119,8 +119,6 @@ In general, this makes this already well-loved feature of GHC even better, allow
 Costs and Drawbacks
 -------------------
 The development time will be fairly minimal, because the "new" functionality represents no novel codepaths or design challenges, merely adding additional parallel constructors and cases to code that already handles the existing type-level literals.
-
-One potential drawback is that type-level strings are currently efficiently represented as ``FastString``\s, but these changes require a change to keeping them always as large, troublesome type-level lists of type-level ``Char``\s. The alternatives are either to (as I have implemented currently) keep ``Symbol``\s as they are but add a type family to convert them into promoted ``String``\s or to hack type-level pattern matching so that they extensively seem to be ``[Char]`` but are actually still efficient.
 
 Alternatives
 ------------
--- a/proposals/0000-unify-type-level-literals.rst
+++ b/proposals/0000-unify-type-level-literals.rst
@@ -62,7 +62,7 @@ The kinds of type-level literals will be as follows:
 
 While the ``Rational`` literals are the most dubious, they require very minimal changes (since ``Ratio Integer`` will now work correctly "for free") and are needed to satisfy the original impetus for this change.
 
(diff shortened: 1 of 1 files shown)


proposals/0000-unify-type-level-literals.rst
```diff
@@ -0,0 +1,156 @@
...
+-----------------------
+For the specific example I used to motivate this change, the implementation of the ``ValidBS`` constraint is as follows:
+
+::
+
+    type ValidBS str = (KnownSymbol str, AllValidChars str) 
+
+    type family AllValidChars (xs :: String) :: Constraint where
+        AllValidChars (x:xs) = If (IsValidChar x) (AllValidChars xs) (TypeError (InvalidCharError x))
+        AllValidChars '[]    = True ~ True
```
int-index: You can use the unit constraint `()` here.

proposals/0000-unify-type-level-literals.rst
```diff
@@ -0,0 +1,156 @@
...
+    type (m :: a) % (n :: a) = ModI a m n
+    type Div (m :: b) (n :: b) = DivI b m n
+    type Mod (m :: b) (n :: b) = ModI b m n
+    type Log2 (m :: b) = Log2I b m
+
+A new type family will be added, ``Data.Type.Equality.Cmp``, using a similar alias-wrapper trick to provide a uniform interface for comparisons.
+
+::
+
+    type family CmpI a (m :: a) (n :: a) :: Ordering
```
int-index: Why do you need this type family? Making Cmp polymorphic is sufficient. Then the instances could look like:

```
type instance Cmp a b = CmpNat a b
type instance Cmp a b = CmpChar a b
type instance Cmp a b = CmpList a b
```

Type inference will do the rest.

proposals/0000-unify-type-level-literals.rst
```diff
@@ -0,0 +1,156 @@
...
+
+The case of positive ``Integer`` literals is challenging. The syntax above is not great, but the obvious alternatives are a subkinding relation or an explicit conversion and I'd consider both of those to be worse.
+
+To handle the legacy case, ``Symbol`` and ``Nat`` will become aliases for ``String`` and ``Natural``, respectively.
+
+All of the mathematical type families in ``GHC.TypeNats`` will be generalized to be poly-kinded so that they work for these new numeric kinds, with the exception of a ``Rational`` implementation for  ``Div``, ``Mod``, and ``Log2``. ``/`` and ``%`` will be added for ``Rational``. Put more explicitly, their interfaces will be as listed in the block below, where ``a`` means any of ``Integer``, ``Natural``, or ``Rational`` and ``b`` means either ``Integer`` or ``Rational``.
+
+::
+
+    type family Add a (m :: a) (n :: a) :: a
```
int-index: You can just remove this `a` parameter and everything will still work (it will become an invisible parameter). Then there will be no need for all the wrappers. See also my comment on `Cmp`, it's the same thing.

int-index:
I'm in support of this proposal. There's another motivation: we make types that contain `Char`, `String`, `Natural`, `Integer`, `Rational`, etc - promotable. The more types promote seamlessly, the easier type-level programming becomes. This will leave us with only two special cases of `Demote` in `singletons`: functions and types.

typedrat:
I don't know where I picked those habits up from, and I'm glad to know that neither of them are required.

int-index:
I would prefer a solution that does not involve TemplateHaskell. My view of TH is that it's a crutch for generating boilerplate due to other language deficiencies. Type families already provide a way to do type-level computation, and they will be (hopefully) enriched to become proper type-level functions in the future.

typedrat:
As would I, I said it probably at least 90% in jest. 

Still, the fact that that's something that could be seen as a genuine suggestion is a lot of why I love this language.

goldfirere:
Perhaps surprisingly, I am not in support of the proposal as it now stands (but I very much agree with the spirit of the proposal): it doesn't align with the term-level as closely as it could. Here are a few differences (under this proposal):

* Type-level numbers are not overloaded. But they can be! Define, e.g., `3 :: k` and choose the representation based on `k`. It's just like writing a type family with a polymorphic return kind; that kind is actually a parameter to the type family and can control its behavior.

* Why name it `Cmp` instead of `Compare`? More like term-level that way.

* Haskell does not parse `+3` in terms. With the overloaded numbers, it wouldn't need to in types, either.

Not a compatibility problem, but another worry:

* I'm worried about the performance implications of using `String`. Instead, we could use type-level `String` as in this proposal, but also allow `-XOverloadedString`s to affect type-level strings. Then, someone could choose to have `"blah" :: Symbol` (where `Symbol` is not a synonym for `String`), which will be more performant. There could be conversions between `String` and `Symbol`. (We could even ponder renaming `Symbol` to `Text`, as I think `Text` is the closest term-level equivalent to `Symbol`.)

int-index:
@goldfirere I considered overloaded numeric literals. My worry is that it will break some code out there that relies on monomorphism for inference.

If we seek symmetry with terms, I also think it should be not just `5 :: k` but `FromInteger (5 :: Integer) :: k`, in the same way as `Num` works in terms. This way users will be able to define their own `FromInteger` type family instances

I wouldn't consider this route without evaluating potential breakage.

And what about Symbol/String? Should `OverloadedStrings` make them desugar into `FromString (s :: String) :: k`? 

On one hand, it would be a nice symmetry with terms. On the other hand, we'd have to start considering support of `OverloadedLists` too.

Let's have a breakdown of potential design choices added to the proposal

int-index:
Ragardless of overloading choices, I agree that adding `+` and `-` to the syntax of type-level numeric literals is uncalled for as it will cause trouble for term/type parser unification. If we don't do overloading, let's add `Positive :: Natural -> Integer` and `Negative :: Natural -> Integer` type families instead of `+` and `-`.

typedrat:
I had the exact same concern with the explicit signs, I just somehow didn't think that it was even possible to do it the way that @goldfirere suggested and I considered the solution @int-index is suggesting "too ugly" but I do agree in retrospect that the current syntax proposed is absolutely unacceptable rather than just merely ugly.

typedrat:
I think that’s a good idea. I’m not sold on the implementation working well in practice but I have no conceptual concerns.

typedrat:
I fixed both of those suggestions.

typedrat:
I'm changing the "primary" design for this proposal to the type family approach, as based on my pilot work it is both much easier to implement and provides a better experience.

int-index:
I cannot speak of intent, but in its current form the proposal does not unify `Symbol` and `String`, although `String` becomes promotable (because `Char` and `[]` are promotable).

In order for `"abc" :: String` to typecheck, the user must enable `-XOverloadedStrings`, then it would desugar into `FromSymbol ("abc" :: Symbol) :: String`.

Without the extension, `"abc"` is still interpreted as `"abc" :: Symbol`, which is distinct from `String`.

Unification of `String`/`Symbol` can be done separately, closely monitoring its effects on compiler performance.

aspiwack:
I see. I had missed the relevant paragraph. Though §3 appears to unify symbol and string, which should probably be fixed.

typedrat:
That was the original intent, but making type checking programs that likely already have pretty dreadful TC performance gain a step that's O(mn) in the number and length of strings with a pretty damn big constant factor to boot... I don't think that's going to end up being a good idea.

simonpj:
I'm not sure I'm following all the details here.  The proposal would be easier going if you felt able to add an example or two to illustrate almost every point.  As it is, there is really only one example in the entire proposal, which doesn't give me enough motivation to think "oh yes, it'd be great to have this".

I'm concerned about this polykinded arithmetic. Can I write this?
```
f :: Proxy a -> Proxy b -> Proxy (a+b)
```
What kind does 'a'` have?   In the world of terms I'd have a `Num a` constraint but somehow not so at the type level? Doesn't that mean that type-level functions become too polymorphic?

If we have polykinded arithmetic, then surely we can have polykinded constants?

We have `symbolVal` and `natVal`; won't we need similar classes and functions for these new type-level strings, rationals etc?

typedrat:
>I'm not sure I'm following all the details here. The proposal would be easier going if you felt able to add an example or two to illustrate almost every point. As it is, there is really only one example in the entire proposal, which doesn't give me enough motivation to think "oh yes, it'd be great to have this".

That's a good point, it's kinda currently operating from the premise that this is "obviously" something that you want to have, here's how it should be done. More motivating examples would also likely help clarify the way that these things should work, as well.

>I'm concerned about this polykinded arithmetic. Can I write this?
>```
>f :: Proxy a -> Proxy b -> Proxy (a+b)
>```
>What kind does 'a'have? In the world of terms I'd have aNum a` constraint but somehow not so at the type level? Doesn't that mean that type-level functions become too polymorphic?

As I have it implemented right now, that would be entirely kind-polymorphic. 

It would work for any `a, b :: k` where `k` has an instance of the type family `(+) :: k -> (a :: k) -> (b :: k) -> (c :: k)` (making implicit kind variables visible), which is `Integer`, `Natural`, and `Rational` by default, but it is intentionally written to be expandable by the user. For example, `Rational` (or `Ratio Natural`) operations aren't wired in at all, they're a normal `type instance` in `GHC.TypeNats`.

The downside to this is, of course, that you can dutifully create the type `Int + Bool` and because there's no way to provide a default case for a type family there's no way to signal that this is a problem "naturally". It shouldn't be hard to special case things so GHC will complain if it sees a monomorphic but unresolvable usage of `(+)` or friends, but that's obviously not the prettiest of all solutions.

The alternative (and something that I'd support in concept, but am aware this is opening a whole new extra-extra-large kettle of fish) would be bringing ad-hoc polymorphism to the type level by making it so that it's an error to use an associated type family when it can't be reduced (because there's no associated instance), perhaps even with a constrained kind.

The dream would be something like this:
```haskell
λ> :{
λ| class TNum k where
λ|     type (a :: k) + (b :: k) :: k
λ|     type (a :: k) - (b :: k) :: k
λ|     type (a :: k) * (b :: k) :: k
λ|     type FromInteger (a :: Integer) :: k
λ| :}
λ> :kind (+)
(+) :: (TNum k) => k -> k -> k
```

>We have symbolVal and natVal; won't we need similar classes and functions for these new type-level strings, rationals etc?

This is entirely an oversight on my part.

goldfirere:
For my part, I don't need extra motivation -- I've wanted this for years. For example, type-level rationals would be very useful for checking units-of-measure. And it feels limiting not to be able to decompose type-level strings (aka symbols).

Simon is right, of course, that the design here is too polymorphic. Right now, we don't have a better way to do it. The right answer is to have [Constrained Type Families](https://repository.brynmawr.edu/cgi/viewcontent.cgi?article=1075&context=compsci_pubs), very much along the lines that @typedrat was working. I'm not convinced that constrained type families will be implementable until we have dependent types, sadly. (How do we promote dictionaries?) Perhaps once we have unsaturated type families -- and, hence, type-level lambdas -- we'll be able to make progress even without full dependent types (which @kcsongor is working on, to the best of my knowledge). But, regardless, I think constrained type families are a ways off.

The lack of constrained type families doesn't diminish my enthusiasm for this idea, though. Yes, it does mean that definitions will be too polymorphic, but no one will truly get hurt -- the problem will just be some suboptimal error messages.

typedrat:
It'll probably be op

typedrat:
I don't see why constrained type families *need* dictionaries. There already is another mechanism for resolving them into definitions in the open world case, it just becomes an error to supply them with anything other than a bare type variable unless that type is an instance.

goldfirere:
But how do we represent this restriction in Core? The natural way is to use a dictionary. I suppose we don't have to, though. Note that we can't just look for some top-level instance: we would want support for GADT-constructor-bound dictionaries and quantified constraints, I wager. I guess my comment about waiting for dependent types is that it will be straightforward to implement constrained type families then, while it would take some extra work now (and work that will be discarded later, when we have dependent types).

typedrat:
That's true, it might be possible to come up with a clever hack that has the right properties but it's a waste in the long run since it'll be obsolete when Dependent Haskell is ready.

int-index:
What if for the time being the evidence for promoted dictionaries would be simply a unit type?

typedrat:
Ooh! That's a good idea. I think that would work nicely.

goldfirere:
I'm not sure what you mean @int-index. How could a unit be evidence?

int-index:
@goldfirere Well, as long as the only thing that type-level type classes can constrain are type families, we only need to know whether a type family is allowed to reduce or not.

Let's say we have

```
class TNum k where
    type (a :: k) + (b :: k) :: k
    type (a :: k) - (b :: k) :: k
    type (a :: k) * (b :: k) :: k
    type FromInteger (a :: Integer) :: k
```

then the evidence will be

```
data TNum k = TNumEv
```

If there's a `TNumEv` in scope, allow the type family to reduce. Otherwise, report an error.

goldfirere:
Do I understand you to mean that every instance for a class also gives rise to a constant of the relevant type? It would be helpful to spell these details out...

int-index:
> Do I understand you to mean that every instance for a class also gives rise to a constant of the relevant type? It would be helpful to spell these details out...

@goldfirere Indeed. My understanding is that it's how type classes are desugared currently. If I define `TNum` and then `instance TNum Int`, here's the `-ddump-simpl` output:

```
-- RHS size: {terms: 1, types: 1, coercions: 0, joins: 0/0}
interactive:Ghci2.$fTNumInt [InlPrag=CONLIKE] :: TNum Int
[GblId[DFunId], Caf=NoCafRefs]
interactive:Ghci2.$fTNumInt = interactive:Ghci1.C:TNum @ Int
```

If `TNum` contained any term-level methods, then `$fTNumInt` would be a record that contains them. However, when we have zero methods, it degrades to a unit type.

The constructor is `C:TNum`, and we want to promote it. The trouble with promoting class dictionaries is that we may not know how to promote the methods, but in case of a type-family-only class there are no methods.

For this special case, I could state the proposal as follows:

* Allow promotion of type classes with no methods.
* Make associated type families constrained.

So that would mean that `type FromInteger (a :: Integer) :: k` would have the following kind:

```
FromInteger :: forall k. TNum k => Integer -> k
```

and reduction of `FromInteger @Int` would require an instance of `TNum Int`, a promoted version of its dictionary would be passed as an implicit argument to it.

What about classes with non-zero methods? Consider

```
class IsList l where
  type family Item l :: *
  fromList :: [Item l] -> l
```

what do we do with `Item`? If we make it constrained but do not promote the dictionary for `IsList`, then this type family won't be usable. Here I see two options:

1. Don't make associated type families constrained if the class dictionary is not promotable.
2. Promote every class as if it had zero methods.

My suggestion is to pick option (2). We are already in a world where the promoted type contains different elements from the original type. Consider `Symbol`: it is currently uninhabited in terms, but contains `""`, `"a"`, `"b"`, ... in types. For classes, that would mean that in terms we leave everything as is, and in types we pretend there are no methods to promote.

When it comes to superclasses, they are currently desugared to additional fields in the dictionary. I propose we promote them too.

All of this seems forward-compatible with the proper constrained type families described in the paper.

goldfirere:
@int-index's https://github.com/ghc-proposals/ghc-proposals/pull/154#issuecomment-409466044 is quite compelling. My big question is: if we're going to have constrained type families, shouldn't we just go all the way to the system proposed in the paper? That is, @int-index's proposal (which has really become quite separable from this ticket), will be a breaking change for anyone who uses associated type families. (Question: do data families behave similarly? I assume so.) If we're going to break code, perhaps we just want to go all the way. What would you think of this? And do you want to write a separate proposal?

Note: the paper lumps together obviously-partial type families (that is, type families whose patterns are not covering) with non-terminating ones. My thinking has evolved on this somewhat, and I want to treat (potentially) non-terminating type families as if they are total. Doing so weakens some of the results of the paper (closed type family reduction would still have to consider the possibility of infinite types), but I think that termination checking will box us into a tight corner.

typedrat:
@goldfirere, *will* it break code? At least, will it break any meaningful or correct code?

Associated type families being used with types without instances for them can't be common, and I can't think of a single circumstance where constraining them like this wouldn't allow code that would otherwise work and do something useful or even sensical.

Either way, this is a much smaller change and much more feasible I'm the short term, I believe.

int-index:
> Doing so weakens some of the results of the paper (closed type family reduction would still have to consider the possibility of infinite types)

This is very sad to hear. It was the main motivation for constrained type families, at least for me.

> which has really become quite separable from this ticket

Agreed. Let's discuss it elsewhere. I think this proposal stands on its own, and the type families need not be constrained for now.

goldfirere:
I have various thoughts on @typedrat and @int-index 's posts about constrained type families, but I'll wait for a more appropriate venue.

I agree that this ticket, with its proposed overly-polymorphic kinds, is sensible even without constrained type families.

knupfer:
Case in point of motivation:
I could improve my library `type-of-html` considerably by beeing able to compile time html escape type level String/Symbol. It would reduce runtime cost by beeing able to concat all static parts of a page at compiletime and spit it out as one blob of ByteString with `pack . symbolVal`.

typedrat:
I'm slightly less busy than I was for a couple weeks there now that I've moved, finished GSoC and gotten the initial start up at my new school taken care of, so I'm going to put some more time into improving this soon.

typedrat:
In addition, I also want to take some time to write up a proposal for constrained type families, just because that's a feature that I'm now more than before really seeing the value of.

nomeata:

nomeata:

int-index:
Oh no. @typedrat let me know if you no longer have time to pursue this, I will happily move this proposal forward.

goldfirere:
It sounds like, no matter how @typedrat responds to the above question, this PR will be pursued. Reopening so that it isn't lost.

typedrat:
This has stalled somewhat because I feel that having satisfying answers to some questions relies on tackling the bigger proposal in #177. Either way, I've learned a lot from writing that proposal so I'll want to do some amount of rewriting of this one before I'd submit it.

typedrat:
Even if #177 is, in the end, rejected, I do plan to finish this one, I just want to know the status of it either way before I fine tune the design down fully.

phadej:
There's also a [Character kind issue](https://gitlab.haskell.org/ghc/ghc/issues/11342) with then-almost ready patch.

`UnconsSymbol` and `ConsSymbol` for converting between `Symbol` and `String` (list of `Char`s) on top level is a very simple alternative.

I think that `Symbol` should stay opaque, for type-equality/TC to be efficient as possible. Comparing strings char by char for `OverloadedLabels` or #282 feels like bad idea (I might be wrong).

typedrat:
Agree that Symbol should stay opaque, ideally also gaining an efficient way to demote them directly into something ByteString adjacent, although I would prefer a primitive to go between `Symbol` and `[Char]` in one fell swoop.

phadej:
>...`Symbol` and `[Char]` in one fell swoop.

Yes, I think it would be nice, on the other hand magic functions are *magic*: trickier to let GHC itself prove things about them.

PR #128, opened 2018-04-25
Allow ScopedTypeVariables to refer to types

nomeata:

PR #126, opened 2018-04-19
Type applications in patterns

nomeata:
[Rendered](https://github.com/ghc-proposals/ghc-proposals/blob/type-applications-in-patterns/proposals/type-applications-in-patterns.rst)

This proposal gives type applications (`@…`) meaning in patterns, by making sure they behave like type signatures.

This happens to also allow binding existential type variables, and hence subsumes #96. But that is a side-product: I believe that the proposal has merits on its own and should “obviously be the case” anyways.

PR #155, opened 2018-07-16
Binding type variables in lambda-expressions

goldfirere:
The [proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0050-type-lambda.rst) has been accepted; the following discussion is mostly of historic interest.

---

[Rendered](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0050-type-lambda.rst)

### DIFF
--- a/proposals/0000-type-lambda.rst
+++ b/proposals/0000-type-lambda.rst
@@ -9,9 +9,7 @@ Binding type variables in lambda-expressions
 .. implemented:: Leave blank. This will be filled in with the first GHC version which
                  implements the described feature.
 .. highlight:: haskell
-.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
-            **After creating the pull request, edit this file again, update the
-            number in the link, and delete this bold sentence.**
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/155>`_.
 .. sectnum::
 .. contents::
 
--- a/proposals/0000-type-lambda.rst
+++ b/proposals/0000-type-lambda.rst
@@ -81,6 +81,8 @@ There are several motivating factors for this addition:
    from intuitive. Using this new syntax, all the action happens in the function
    definition.
 
+4. See crowd-sourced example `here <https://github.com/ghc-proposals/ghc-proposals/pull/155#issuecomment-459430140>`_.
+
 Proposed Change Specification
 -----------------------------
 GHC's type system is *bidirectional*, meaning that it sometimes is *inferring* a type
@@ -101,7 +103,7 @@ twice in a single lambda; this is true for type variables, too.
 
 Thus, the proposal boils down to one rule:
 
-* ``\ @a -> body``, being checked against the type ``forall a. ty``, binds the type
+* ``\ @a -> body``, being checked against the type ``forall a. ty`` (where the ``a`` is *specified*), binds the type
   variable ``a`` and then checks ``body`` against the type ``ty``. Checking an
   expression ``\ @a -> body`` against a type that does not begin with a ``forall``
   is an error. The token after the ``@`` must be a type variable name or ``_``.
@@ -215,6 +217,9 @@ Effect and Interactions
   ``\@`` is a valid name for a user-defined operator. This proposal does not change that.
   If you want to bind a type variable in a lambda, you must separate the ``\`` from the
   ``@``.
+
+* This proposal makes abstracting over type variables the dual of applying types with
+  visible type application.
 
 * This proposal is meant to dovetail nicely with other recent proposals in this space
   (`#126`_, `#128`_), but all the proposals are orthogonal. Any can usefully be accepted
--- a/proposals/0000-type-lambda.rst
+++ b/proposals/0000-type-lambda.rst
@@ -147,6 +147,9 @@ Effect and Interactions
   a function ``f :: forall {k} (a :: k). ...``, you will have to rely on the old behavior
   of ``-XScopedTypeVariables`` to bring ``k`` into scope in ``f``\'s definition. This is
   regrettable but seems an inevitable consequence of the ``{k}`` notation.
+
+.. _`26`: https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0026-explicit-specificity.rst
+.. _`#99`: https://github.com/ghc-proposals/ghc-proposals/pull/99
   
 * (technical) The `Visible Type Applications`_ (VTA) paper defines the behavior about what to
   do when checking against a polytype: it says to deeply skolemize. However, eager deep
--- a/proposals/0000-type-lambda.rst
+++ b/proposals/0000-type-lambda.rst
@@ -121,6 +121,42 @@ This new behavior will be available whenever both ``-XTypeApplications`` and
 
 This change is specified in the appendix to the `Type variables in patterns <https://cs.brynmawr.edu/~rae/papers/2018/pat-tyvars/pat-tyvars.pdf>`_ paper.
 
+Examples
+--------
+
+Here are two real-world examples of how this will help, courtesy of @int-index:
+
+1. It would be useful to eliminate ``Proxy`` in this style of proof::
+
+     class WithSpine xs where
+       onSpine ::
+         forall r.
+         Proxy xs ->
+         ((xs ~ '[]) => r) ->
+         (forall y ys.
+           (xs ~ (y : ys)) =>
+           WithSpine ys =>
+           Proxy y ->
+           Proxy ys ->
+           r) ->
+         r
+
+   Code taken `from here <https://github.com/int-index/caps/blob/2f46fc6d5480bdef0a17f64359ad6eb29510dba4/src/Monad/Capabilities.hs#L273>`_.
+
+   Compare:
+
+   a. ``@``\-style: ``withSpine @xs (onNil ...) (\ @y @ys -> onCons ...)``
+   b. ``Proxy``\-style: ``withSpine (Proxy :: Proxy xs) (onNil ...) (\(Proxy :: Proxy y) (Proxy :: Proxy ys) -> onCons ...)``
+
+2. From ```reflection`` <https://hackage.haskell.org/package/reflection-2.1.4/docs/Data-Reflection.html#v:reify>`_::
+
+     reify :: forall a r. a -> (forall (s :: *). Reifies s a => Proxy s -> r) -> r
+
+   Compare:
+
+   a. ``@``\-style: ``reify (\ @s -> ...)``
+   b. ``Proxy``\-style: ``reify (\(Proxy :: Proxy s) -> ...)``
+
 Effect and Interactions
 -----------------------
 
--- a/proposals/0000-type-lambda.rst
+++ b/proposals/0000-type-lambda.rst
@@ -40,7 +40,7 @@ extends this idea to lambda-expressions, allowing ``\ @a x -> ...``. Here are so
   higherRank :: (forall a. a -> a -> a) -> ...
   higherRank = ...
 
-  ex = higherRank (\@a x _y -> (x :: a))  -- it works in a higher-rank situation, too
+  ex = higherRank (\ @a x _y -> (x :: a)) -- it works in a higher-rank situation, too
 
   wrong @a x = x                          -- we can't do this without a type signature
 
--- a/proposals/0000-type-lambda.rst
+++ b/proposals/0000-type-lambda.rst
@@ -139,6 +139,14 @@ Effect and Interactions
 * This proposal is meant to dovetail nicely with other recent proposals in this space
   (`#126`_, `#128`_), but all the proposals are orthogonal. Any can usefully be accepted
   without the others.
+
+* Accepted proposal `26`_ (debated as `#99`_) introduces the possibility of user-written
+  specificity annotations (``forall {k} ...``). An *inferred* variable, including one
+  written by the programmer using this new notation, is not available for use with
+  any form of visible type application, including the one proposed here. If you have
+  a function ``f :: forall {k} (a :: k). ...``, you will have to rely on the old behavior
+  of ``-XScopedTypeVariables`` to bring ``k`` into scope in ``f``\'s definition. This is
+  regrettable but seems an inevitable consequence of the ``{k}`` notation.
   
 * (technical) The `Visible Type Applications`_ (VTA) paper defines the behavior about what to
   do when checking against a polytype: it says to deeply skolemize. However, eager deep
@@ -163,6 +171,11 @@ Alternatives
 If we want to bind type variables in lambda-expressions, I think this is the only way to do it.
 We don't have to, of course, but then there will still be one area in GHC/Haskell that requires
 ``Proxy``, and that's unfortunate.
+
+One alternative design would be to rearrange the extensions so that users could enable
+parts of today's ``ScopedTypeVariables`` without enabling the strange binding behavior of
+``forall``. I don't feel the need for this, myself, so I do not plan on working out this
+design, but I'm happy to accept contributions toward this end from the community.
 
 Unresolved questions
 --------------------
--- a/proposals/0000-type-lambda.rst
+++ b/proposals/0000-type-lambda.rst
@@ -148,7 +148,7 @@ Here are two real-world examples of how this will help, courtesy of @int-index:
    a. ``@``\-style: ``withSpine @xs (onNil ...) (\ @y @ys -> onCons ...)``
    b. ``Proxy``\-style: ``withSpine (Proxy :: Proxy xs) (onNil ...) (\(Proxy :: Proxy y) (Proxy :: Proxy ys) -> onCons ...)``
 
-2. From ```reflection`` <https://hackage.haskell.org/package/reflection-2.1.4/docs/Data-Reflection.html#v:reify>`_::
+2. From `\ ``reflection`` <https://hackage.haskell.org/package/reflection-2.1.4/docs/Data-Reflection.html#v:reify>`_::
 
      reify :: forall a r. a -> (forall (s :: *). Reifies s a => Proxy s -> r) -> r
 
--- a/proposals/0000-type-lambda.rst
+++ b/proposals/0000-type-lambda.rst
@@ -1,0 +1,167 @@
+Binding type variables in lambda-expressions
+============================================
+
+.. proposal-number:: Leave blank. This will be filled in when the proposal is
+                     accepted.
+.. trac-ticket:: Leave blank. This will eventually be filled with the Trac
+                 ticket number which will track the progress of the
+                 implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+.. _`#126`: https://github.com/ghc-proposals/ghc-proposals/pull/126
+   
+Proposal `#126`_ allows us to bind scoped type variables in patterns using an ``@a`` syntax.
+However, the new syntax is allowed only in *constructor* patterns ``K @a @b x y``. This proposal
+extends this idea to lambda-expressions, allowing ``\ @a x -> ...``. Here are some examples::
+
+  id :: a -> a
+  id @a x = (x :: a)                      -- no forall needed
+
+  prefix :: a -> [[a]] -> [[a]]
+  prefix @a x yss = map xcons yss
+    where xcons :: [a] -> [a]             -- this signature is impossible to write without ScopedTypeVariables
+          xcons ys = x : ys
+
+  const :: a -> b -> a
+  const @c x _ = (x :: c)                 -- names can change; you do not have to bind every type variable
+
+  fconst :: a -> b -> b
+  fconst @_ @d _ x = (x :: d)             -- order matters
+
+  pair :: forall a. a -> (a, a)           -- brings a into scope (just like today)
+  pair @b x = (x :: a, x :: b)            -- brings b into scope with the same meaning as a
+  
+  higherRank :: (forall a. a -> a -> a) -> ...
+  higherRank = ...
+
+  ex = higherRank (\@a x _y -> (x :: a))  -- it works in a higher-rank situation, too
+
+  wrong @a x = x                          -- we can't do this without a type signature
+
+  stillWrong @a (x :: a) = x              -- even here we can't
+  
+Motivation
+----------
+
+There are several motivating factors for this addition:
+
+1. There are cases where a ``Proxy`` is necessary in order for a higher-rank function argument
+   to access a type variable, such as::
+
+     type family F a
+
+     higherRankF :: (forall a. F a -> F a) -> ...
+
+     usage = higherRankF (\ (x :: F a) -> ...)
+
+   The ``(x :: F a)`` pattern signature does not work, because ``F`` is not injective. There
+   is no way to be sure that the ``a`` in ``usage`` is meant to match the ``a`` in
+   ``higherRankF``. Currently, there is simply no way for ``usage`` to get access to the
+   type variable written in the signature for ``higherRankF``. This code would have to
+   be rewritten to use ``Proxy``. Under this proposal, however, ``usage`` could be simply ::
+
+     usage = higherRankF (\ @a x -> ...)
+
+   Ah. That's better.
+
+2. With `#126`_, we can bind type variables in constructor patterns, allowing us to easily
+   capture existentials. The only other place a type variable can enter scope is in a
+   function definition, and so it's only logical to extend `#126`_ to do so.
+
+3. ``ScopedTypeVariables``\'s mechanism for binding type variables using a ``forall`` in
+   a signature has never sat well with some. (I'm in the some, but I'm not the only one.)
+   A type signature can appear arbitrarily far away from a function definition, and
+   (to me) the use of ``forall`` to induce scoping over the function definition is far
+   from intuitive. Using this new syntax, all the action happens in the function
+   definition.
+
+Proposed Change Specification
+-----------------------------
+GHC's type system is *bidirectional*, meaning that it sometimes is *inferring* a type
+and sometimes is *checking* a type. `Practical Type Inference for Arbitrary-Rank Types <http://repository.upenn.edu/cis_papers/315/>`_ is a careful introduction of the ideas, though
+GHC's algorithm is currently based on the more recent `Visible Type Applications`_. Essentially,
+bidirectionality means that the type system can distinguish (and make decisions based on)
+the difference between knowing what type to expect and not.
+
+.. _`Visible Type Applications`: https://repository.brynmawr.edu/cgi/viewcontent.cgi?article=1001&context=compsci_pubs
+
+Under this proposal, the new feature is allowed only in *checking* mode. That is, we
+always know exactly what type is expected for a function definition or lambda expression.
+
+As always, we can consider a nested lambda ``\ x y z -> ...`` to be an abbreviation for
+``\ x -> \ y -> \ z ->``. This does not change if one of the bound variables is a type
+variable (preceded by ``@``).
+
+Thus, the proposal boils down to one rule:
+
+* ``\ @a -> body``, being checked against the type ``forall a. ty``, binds the type
+  variable ``a`` and then checks ``body`` against the type ``ty``. Checking an
+  expression ``\ @a -> body`` against a type that does not begin with a ``forall``
+  is an error. The token after the ``@`` must be a type variable name or ``_``.
+
+That's it! Note that this specification assumes that the variable name in the lambda
+equals the variable name in the ``forall``. If the type begins with a ``forall``, this
+correspondence can always be made to happen because we can freely rename the bound
+type variable in a ``forall``.
+
+As usual, we can interpret a function defintion ``f <args> = body`` as
+``f = \ <args> = body``, and thus the function-definition case reduces to the lambda-expression
+case above.
+
+This new behavior will be available whenever both ``-XTypeApplications`` and
+``-XScopedTypeVariables`` are enabled.
+
+Effect and Interactions
+-----------------------
+
+* One might worry about parsing. After all, ``@`` already has a meaning in patterns. However,
+  this is all OK: whenever ``-XTypeApplications`` is enabled, ``@`` with a preceding
+  whitespace character (or comment) is parsed differently from ``@`` without a preceding
+  whitespace character (or comment). So ``f x @a`` is a good left hand side for a function
+  with type ``Int -> forall a. ...`` and ``f x@a`` simply binds both ``x`` and ``a`` to the
+  first argument to ``f``.
+
+* An astute reader will note that I put spaces after all my lambdas. That is because
+  ``\@`` is a valid name for a user-defined operator. This proposal does not change that.
+  If you want to bind a type variable in a lambda, you must separate the ``\`` from the
+  ``@``.
+
+* (technical) The `Visible Type Applications`_ (VTA) paper defines the behavior about what to
+  do when checking against a polytype: it says to deeply skolemize. However, eager deep
+  skolemization will spell trouble for this extension, as we need the lambdas to see
+  the ``forall``\s. The end of the Section 6.1 in the `extended VTA <https://cs.brynmawr.edu/~rae/papers/2016/type-app/visible-type-app-extended.pdf>`_ paper discusses
+  why we do eager deep skolemization: essentially, the alternative would be to do
+  type generalization at inflection points between checking and inference mode,
+  right before doing the subsumption check. Type generalization is hard in GHC, though,
+  and so the paper avoided it. In order to implement this proposal, we'll have to work
+  out how to do this.
+
+Costs and Drawbacks
+-------------------
+This is another feature to specify and maintain, and that's always a burden. It will take
+some creative thought about how to do generalization properly (last point in previous section),
+but I don't actually think the code will be all that challenging there.
+
+There is a potential confusion with as-patterns.
+
+Alternatives
+------------
+If we want to bind type variables in lambda-expressions, I think this is the only way to do it.
+We don't have to, of course, but then there will still be one area in GHC/Haskell that requires
+``Proxy``, and that's unfortunate.
+
+Unresolved questions
+--------------------
+None at this time.
+
+Implementation Plan
+-------------------
+I'm happy to advise and support a volunteer who wishes to implement. I might do it myself
+or work with a student on this someday, as well.
--- a/proposals/0000-type-lambda.rst
+++ b/proposals/0000-type-lambda.rst
@@ -107,10 +107,11 @@ Thus, the proposal boils down to one rule:
 That's it! Note that this specification assumes that the variable name in the lambda
 equals the variable name in the ``forall``. If the type begins with a ``forall``, this
 correspondence can always be made to happen because we can freely rename the bound
-type variable in a ``forall``.
+type variable in a ``forall``. (This "free renaming" is entirely internal; a user
+can write a different name in the type than in the pattern, always.)
 
 As usual, we can interpret a function defintion ``f <args> = body`` as
-``f = \ <args> = body``, and thus the function-definition case reduces to the lambda-expression
+``f = \ <args> -> body``, and thus the function-definition case reduces to the lambda-expression
 case above.
 
 This new behavior will be available whenever both ``-XTypeApplications`` and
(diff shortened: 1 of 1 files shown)


nomeata:
I expect the reddits and twitters to go crazy when they discover this :-)

One question that you should field: It seems “obvious” what
```
id @a (x :: a) = x -- no type signature
```
should mean. Yet the proposal rejects it, right?

Can you put in some motivation why this should be rejected? Surely the actually tricky cases are going to be more complicated that that one … so the next question is: Can the line be drawn somewhere else? Is there a easy to specify class of “easy cases” where we could allow a lambda-bound type variable like above?

goldfirere:
Yes, the proposal rejects that. I simply don't know an algorithm that accepts it. We can perhaps expand to that case later, but I don't think it will be easy. Can you think of a way to do it?

nomeata:
I am ignorant of the complications of type inference, but please don't tell anyone :-)

proposals/0000-type-lambda.rst
```diff
@@ -0,0 +1,167 @@
...
+  expression ``\ @a -> body`` against a type that does not begin with a ``forall``
+  is an error. The token after the ``@`` must be a type variable name or ``_``.
+
+That's it! Note that this specification assumes that the variable name in the lambda
+equals the variable name in the ``forall``. If the type begins with a ``forall``, this
+correspondence can always be made to happen because we can freely rename the bound
+type variable in a ``forall``.
+
+As usual, we can interpret a function defintion ``f <args> = body`` as
+``f = \ <args> = body``, and thus the function-definition case reduces to the lambda-expression
```
barrucadu: I think you meant `f = \ <args> -> body` here

barrucadu:
This is a neat idea.  I'm also in the made-uneasy-by-`ScopedTypeVariables` camp, having things happen at the function definition site is a nice change.

nomeata:
Those of you in that camp, would you want this proposal to introduce a new language extension (e.g. `TypeApplicationPatterns`) that enables this feature and (presumably) #126 and (maybe, but not necessarily) type signatures in patterns binding variables, but does *not* make `forall.` in type signatures scope over the body? This might also be helpful if the Haskell Prime committee wants to get some form of scoped type variables into the language, but without the remote-effect `forall` wart.

barrucadu:
I'd be in favour of that, but don't think it necessarily needs to be in this proposal.

chrisdone:
> Note that this specification assumes that the variable name in the lambda equals the variable name in the forall. If the type begins with a forall, this correspondence can always be made to happen because we can freely rename the bound type variable in a forall.

This seems to imply that each `@x` must have a corresponding `x` in the type. But this example from the doc indicates that you can use whatever name you want:

```haskell
fconst :: a -> b -> b
fconst @_ @d _ x = (x :: d)             -- order matters
```

So even this would be fine:

```haskell
fconst :: a -> b -> b
fconst @b @a _ x = (x :: a)             -- order matters
```

Can you clarify the relationship between the `@t` parameters and naming?

goldfirere:
I've updated the proposal, but I'll repeat the new bits here:

- No, I don't think we should have a separate extension, as I elaborate (slightly) in the proposal.

- The name bound in the type and the name bound in the definition never need to be the same, though I see why the text was confusing on that point. I've tried to clarify.

goldfirere:
Of the examples in https://github.com/ghc-proposals/ghc-proposals/pull/128#issuecomment-405152882, the first is no good (no type signature), but the last one is OK: there's no problem aliasing `b` and `c` in that manner.

AshleyYakeley:
I think there may be some benefit to a coding discipline that makes use of this but doesn't use `ScopedTypeVariables`, so I'd be in favour of a new `TypeApplicationPatterns` extension.

fsestini:
What's the reason why `a` is brought into scope here, as well as `b`?

```
pair :: forall a. a -> (a, a)
pair @b x = (x :: a, x :: b)
```

I'd expect `b` to alpha-rename `a` out of existence, after the binding.

AshleyYakeley:
@fsestini `a` is not alpha-renamed, it's merely bound to the name `b`.

goldfirere:
@AshleyYakeley has it right. The renaming in the proposal is essentially just a conceit to avoid describing that one variable is bound to another.

AntC2:
I can't see examples of repeated tyvars in binding sites. Is that allowed/what does it mean?

    -- no signature given for f, so we need to infer it
    f @a @a x y = ...
    f @a (x :: a) (y :: a) = ...               -- seems to be forbidden

Is any part of this proposal different in effect to putting `PatternSignatures`?

    f (x :: a) (y :: a) = ...

(Must admit my head's spinning from all the different proposals around `ScopedTypeVariables` and patterns: does this superseded some of Joachim's PRs?)

goldfirere:
Just like in a normal lambda, repeated variables are not allowed.

This proposal is meant to obviate the need for pattern signatures by providing a more ergonomic mechanism. (Though, to be clear, I have no intention nor desire of suggesting the removal of `PatternSignatures`, which some users might reasonably prefer.) The first example in the "Motivation" part of the proposal includes an example that cannot be expressed via a pattern signature.

I've added text to the proposal clarifying that it's orthogonal from the other proposals (#126, #128) in this space. Any of the three can usefully be accepted and implemented independently of the others, but I think they go nicely together.

Ericson2314:
I'm also for a separate extension, or even just making part of TypeApplication. This doesn't seem like the worst sort of churn anyways to me, even if I didn't really like the outcome: I'd imagine both extensions can be turned on giving us what is described here. I also like the explicit-qualification-only last bit of https://github.com/ghc-proposals/ghc-proposals/pull/128#issuecomment-405152882, which only seems natural to me if signatures and terms are completely different scopes.

jberryman:
(Mostly writing this up because I haven't seen this opinion expressed (which might mean it should be disregarded: most folks on reddit etc. seem to find `TypeApplications` perfectly natural. Also I didn't follow the discussions around that proposal).)

I suspect this is not a constructive comment, but the `TypeApplications` syntax that this and the other proposal puns/extends to other contexts has never made much sense to me. There doesn't seem to be a way to reasonably explain it without talking about _core_ or _System F_ or dictionaries or other stuff that's not haskell (but that might be totally familiar and intuitive to someone working on GHC).

On the other hand the syntax in the `PatternSignatures` bit of `ScopedTypeVariables` (and, insofar as I understand them, the changes proposed in #128) is totally intuitive to me.

Sorry if this is a dead horse that's already sailed, and as you take pains to point out both features can live alongside each other and people are free to choose whichever they prefer.

So I guess I have a question about specifically the cases in motivation pt. 1, where we can't use `ScopedTypeVariables` to express a particular type; is this just sort of incidental, i.e. it's just not the way `ScopedTypeVariables` works _now_ but it might just as easily be made more powerful in the way proposed here? I admit I don't really understand the example you give (maybe including the RHS in the example would help me?)

> ...(to me) the use of `forall` to induce scoping over the function definition is far from intuitive

I think it's worse than unintuitive, since it reuses a keyword that means "run away this type is too hard for you!" to implement behavior that [some people already naturally assumed](https://stackoverflow.com/q/5476378). Why shouldn't `ScopedTypeVariables` just bring _all_ variables into scope in the body? This would break some existing code, just as the way it works now with `forall` can break some existing code (that used `RankNTypes`). I'm sure that idea was discussed and rejected, and sorry this is straying OT..

...so I think for me, this proposal doesn't satisfy me wrt some of the criticisms I share with you of `ScopedTypeVariables` that you describe in motivation pt. 3, since I find the syntax even stranger.

proposals/0000-type-lambda.rst
```diff
@@ -0,0 +1,179 @@
...
+  fconst :: a -> b -> b
+  fconst @_ @d _ x = (x :: d)             -- order matters
+
+  pair :: forall a. a -> (a, a)           -- brings a into scope (just like today)
+  pair @b x = (x :: a, x :: b)            -- brings b into scope with the same meaning as a
+  
+  higherRank :: (forall a. a -> a -> a) -> ...
+  higherRank = ...
+
+  ex = higherRank (\@a x _y -> (x :: a))  -- it works in a higher-rank situation, too
```
int-index: I think you meant `\ @a` (with the whitespace) here.

int-index:
@goldfirere How does this proposal interact with specificity? Let's say we have

```
f :: forall {k} (a :: k). Proxy a -> ...
f = ...
```

I can call it as

```
f @a
```

hence, for symmetry, binding the variables at definition site should be

```
f :: forall {k} (a :: k). Proxy a -> ...
f @a = ...
```

rather than

```
f :: forall {k} (a :: k). Proxy a -> ...
f @k @a = ...
```

So how do we bind `k`? We need separate syntax for this. Perhaps one of:

1. ```
    f @@k @a = ...
    ```
2. ```
    f @{k} @a = ...
    ```
but then, why not allow the same syntax at call site?

Awkward!

goldfirere:
Drat. https://github.com/ghc-proposals/ghc-proposals/pull/155#issuecomment-405458733 is a good point. Right now, the only way of binding `k` is the old `-XScopedTypeVariables` mechanism, and this proposal doesn't change that. The whole point of *inferred* variables is that you can't use them with type application, and so I'm loathe to change that.

goldfirere:
@jberryman I think yours is a great comment to post. We pointy-headed types folk need to be reminded that we're not the only ones in the room sometimes.

I'm sorry that these features have been explained in terms of Core or System F or dictionaries. They shouldn't be. I hope I haven't done that.

I'll try to explain with a simple example. Simple examples are contrived, but I hope that it's not too contrived: `id`. We have `id`'s type as `a -> a`. In Haskell, putting a type signature like that implicitly means `id :: forall a. a -> a`. All the `forall` does is make the binding of `a` (where the name is introduced) more explicit. Every polymorphic type signature you've ever written starts with a hidden `forall`.

With the original `-XTypeApplications` extension, we can *instantiate* the `forall` with a type application. For example, `id @Int` will have type `Int -> Int`. Instantiation here really means substitution: since `id :: forall a. a -> a`, when I say `id @Int`, I mean to replace `a` with `Int`. So, `id @Int :: Int -> Int`. And that's it! No System F in sight.

More generally, if an expression `<expr>`'s type is `forall a. <blahblah>`, then `<expr> @<type>` has type `<blahblah>`, but with `<type>` substituted in for `a`. If you leave out the type application, GHC guesses it for you.

With the extension discussed here, roughly the opposite applies. Once again, we have `id :: forall a. a -> a`. This means that we can bind the type variable `a` in `id @a x = x`. Because `id`'s type begins with a `forall`, we're allowed to write `id @a ...` in `id`'s definition. Once again, GHC lets you leave out the type variable binding. For `id`, binding the type variable is useless, but with the more intricate examples in the proposal, there is more merit.

I hope this has been helpful!

As to the direct question above, the example in my motivation is not just a quirk of `ScopedTypeVariables`, but a fundamental limitation of pattern signatures. If you're not experienced with type families and higher-rank types, this may be hard to explain, but I'll state that no natural extension of `ScopedTypeVariables` will allow that example to be accepted.

int-index:
> Right now, the only way of binding `k` is the old `-XScopedTypeVariables` mechanism, and this proposal doesn't change that. The whole point of inferred variables is that you can't use them with type application, and so I'm loathe to change that.

Right. But this leaves us in a situation where some type variables still can be bound only with `-XScopedTypeVariables`, which goes against point (3) of the Motivation section for this proposal. And when we reapply the same reasoning to #126, it means there will be GADTs that cannot bind some of their existentials.

int-index:
For what it's worth, I do not think this is a problem with #155 or #126 — rather, it is yet more evidence that specificity is a misfeature. (I really don't think that avoiding a couple of `@_` justifies the complexity it brings). Let's just clarify in the proposal that inferred variables cannot be bound.

proposals/0000-type-lambda.rst
```diff
@@ -159,7 +160,12 @@ We don't have to, of course, but then there will still be one area in GHC/Haskel
 
 Unresolved questions
 --------------------
-None at this time.
+Q: As brought up in the GitHub trail: should we consider changes to the extension structure?
+Specifically, do we want a way to enable this feature without also enabling the fact that
+a ``forall`` in a type signature binds a type variable in a definition.
+
+A: I say "no". I would prefer that world to the one we're currently in, but I simply don't
```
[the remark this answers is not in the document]
goldfirere: I was thinking that these suggestions imply we should change the behavior of `ScopedTypeVariables`. Of course, we could redefine `ScopedTypeVariables` to actually mean a union of new extensions, such as `ExplicitForAll`, `InstanceTypeVariables` (new), `TypeVariablesInPatterns` (new), and `TypeVariablesInForAll` (new). It's the last one in this list that some dislike, and so then they could turn it off.

goldfirere:
I've incorporated the latest round of suggestions into the proposal.

Ericson2314:
Quick proposal for separate extensions, since @goldfirere wanted to defer:

- *anonymous* "Minimal Scoped Type Variables": This extension allows type variables to be captured, without saying what the binding construct is. It also prevents implicit free type variable binding where the locally-bound type variable context is non-empty. It again also allows everything `PatternSignatures` used to allow. It has no name because there's no reason to enable it on its own.
- `ScopedTypedVariables`: "Minimal Scoped Type Variables" + `forall` in signature binds over definition.
- `TermLevelTypeLambdas`: "Minimal Scoped Type Variables" + `\ @a` binds over lambda body.
- `ScopedTypedVariables` and `TermLevelTypeLambdas`: `\ @a` removes any signature-`forall`-bound variables from scope in its body, to compel the future to hurry up :). that means with both these are OK:
  ```haskell
  f :: a -> (forall b. b -> b) -> a
  f a _ = (a :: a)  -- only 'a' in local type variable context
  ```
  ```haskell
  f :: forall a c. a -> c
  f (a :: a) = f a -- more 'a' and 'c' here ok
    where f :: forall b. b -> c
          f = \ @b b -> _ -- only 'b' in local type variable context
  ```

Ericson2314:
Will there by a `\ @` analog to  `\case {}`?

goldfirere:
Thanks for these concrete suggestions!

I think your first example has a typo -- I don't see a way for the type variable `a` to be in scope.

For your "minimal" extension: would this include scoped type variables in class instances? What about in class method defaults? (In both cases, just for class variables.) These two features are enabled by `ScopedTypeVariables` but aren't covered elsewhere in your mini-proposal.

According to this proposal `TermLevelTypeLambdas` (can we have this and #126 covered by `PatternTypeApplications`?) would enable pattern signatures, which is strange.

What enables pattern signatures to bind variables?

I personally disagree with the `\ @a` drops all `forall`-bound variables from scope.

By your `\ @`-analog-of-`\case {}`, do you mean to eliminate a type variable of an empty kind? No, I don't think we can ever have this, because all kinds are trivially inhabited. In terms, `\case{}` is strict (all empty `case`s are). There's no way of making `\ @` strict in a language with type erasure, though.

int-index:
I see the proposal was updated by the typo I pointed out in https://github.com/ghc-proposals/ghc-proposals/pull/155#discussion_r202892982 is still there.

Is `\@a` (no whitespace) intentional there? In this case I don't understand the proposed syntax.

goldfirere:
The `ExplicitForAll` + `PatternSignatures` line confuses me. Why would `ExplicitForAll` induce this new implicit `forall` behavior? It seems that plain old `PatternSignatures` should have this behavior, as it means all pattern signatures are closed.

And I still must say I don't like that `ScopedTypeVariables` stops working if there is a `@a` pattern anywhere. That behavior is surprisingly non-local if a function has many equations.

int-index:
I vote for the least possible amount of interactions between extensions. A `@a` pattern (or the possibility of its presence) should not change the meaning of scoped type variables.

Besides, we cannot elaborate ScopedTypeVariables into `@a` patterns because this does not cover inferred type variables.

simonpj:
Although I do not lie awake at night worrying about the absence of explicit type abstraction, it would be a nice way to be able to say explicitly things that are currently only inferred.  As well as allowing the programmer to say explicitly what they mean, it give language in which to explain things that happen.  (The point about being able to print kind applications in types with an @ sign is an example.)

The higher-rank "you can't bring 'a' into scope here" example is not compelling to me.  We have had zero user complaints on this score.  But still, it's a point I suppose.

I'm concerned about this

> Under this proposal, the new feature is allowed only in checking mode. That is, we always know exactly what type is expected for a function definition or lambda expression.

I really don't know it means to say "the feature is only available in checking mode".  What, precisely, is "checking mode?", asks a user.   It is very far from a specification!   And It'd be very odd if
```
f @a (x::a) = x
```
didn't succeed in inference (no checking here!) and produce `f :: forall a. a -> a`.

The final bullet in Section 3, about deep skoelmisation, is interesting. It implies that there is some real technical work, not yet done, to ensure that we can implement this properly.  But I agree it would be most peculiar if this point turned out to be a show-stopper.

On the type inference front I'm also worried about
```
f @a (x::a) @b (y::b) = ...
```
Now if the RHS gives rise to constraints involving `b` we must put those inside the `@b`, to infer a type for `f` like
```
f :: forall a. Ord a => a -> forall b. Ix b => b -> blah
```
Worth pointing this out, with a concrete example (not blah).  Again, some impact on the type inference engine.   This one is a bit more of a worry; currently inferring the constraints for a let-bound function is quite tricky (`simplifyInfer`) and we do it once for a mutually recursive group of functions.  But now I'm not so sure.   Urk.  Maybe that's why you spoke about "checking mode only"?   (Well, if so, it deserves a far more thorough treatment, even a whole subsection to itself, with examples.)

One other minor point

* Please give an example showing that the type variable can have different names in different branches.  Thus
```
f @a Nothing = ...
f @b (Just (x::b)) = ...
```
  is absolutely fine.

int-index:
> The higher-rank "you can't bring 'a' into scope here" example is not compelling to me. We have had zero user complaints on this score. 

@simonpj At least one user who has wished for this in multiple occasions is me. It would be useful to eliminate `Proxy` in this style of proofs:

https://github.com/int-index/caps/blob/2f46fc6d5480bdef0a17f64359ad6eb29510dba4/src/Monad/Capabilities.hs#L273

```
class WithSpine xs where
  onSpine ::
    forall r.
    Proxy xs ->
    ((xs ~ '[]) => r) ->
    (forall y ys.
      (xs ~ (y : ys)) =>
      WithSpine ys =>
      Proxy y ->
      Proxy ys ->
      r) ->
    r
```

and in `reflection`:

https://hackage.haskell.org/package/reflection-2.1.4/docs/Data-Reflection.html#v:reify
```
reify :: forall a r. a -> (forall (s :: *). Reifies s a => Proxy s -> r) -> r
```

Compare:

1. `@`-style: `withSpine @xs (onNil ...) (\ @y @ys -> onCons ...)`
2. `Proxy`-style: `withSpine (Proxy :: Proxy xs) (onNil ...) (\(Proxy :: Proxy y) (Proxy :: Proxy ys) -> onCons ...)`.

And:

1. `@`-style: `reify (\ @s -> ...)`
2. `Proxy`-style: `reify (\(Proxy :: Proxy s) -> ...)`

int-index:
In fact, the higher-order motivation is the most compelling for me. `-XScopedTypeVariables` are not that bad in other circumstances.

Ericson2314:
@goldfirere Edited to fix both those things `ExplicitForAll` is no longer mentioned, and the explicit `@_` clobbering implicit `@_` is per equation.

goldfirere:
I've also run into this restriction wearing my "user" hat. I can't find where, sadly, but the problem hit me when I wasn't looking for it.

It's perhaps inscrutable to a user, but "available only in checking mode" is a very tight specification. See Appendix B of our [recent Haskell Symposium paper](https://cs.brynmawr.edu/~rae/papers/2018/pat-tyvars/pat-tyvars.pdf) and the [Visible Type Application paper](https://repository.brynmawr.edu/cgi/viewcontent.cgi?article=1001&context=compsci_pubs) for more discussion. The VTA paper includes what syntactic constructs and under what scenarios "checking" mode happens.

Of course, this isn't all that accessible to users. But I think we can state that the feature is only available when GHC already knows what type the lambda-expression should have -- much like what's currently done for `-XRankNTypes`, [here](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#type-inference). So I'm not too troubled on this point.

To be clear, I have no qualms rejecting `f @a (x :: a) = x` without a type signature. Yes, this case is really obvious, but I just don't know how to do this predictably and in generality.

simonpj:
**Higher rank motivation**: great to have two concrete use-cases from int-index -- let's put them into the proposal!

**Checking mode**: I'm OK with what you propose provided the proposal explains it much more carefully, not just with a one-line paragraph.  I suggest a subsection.  Needs 
* reasoning (why make this restriction), 
* a rough user-level spec for "checking mode" 
* examples of what does and does not work
* a precise pointer to the particular place in the paper where it is formalised

yav:
I can see some difficulties if trying to infer types for a mutually recursive group of definitions that feature `@a@` parameters and no type signatures, as those are usually checked using mono types, and generalized after the fact.

However, for the non-recursive case it seems that ordinary type inference might more or less "just" work", by doing something like generalization whenever we encounter `@a` parameters: it is just that we don't need to compute what to generalize, instead we simply use the provided parameters, and then do all the usual stuff to figure out what constraints end up in the schema.

Here are some more examples, that illustrate some potential corner cases with inference:
```
f @a = id   -- perhaps of type `forall b. forall a. b -> b` ?

-- Is this OK?
g @a True x = x :: a
g False @a x = x :: a 
```

I think it might be OK if we required that the type parameters all come before the other parameters.

Ericson2314:
I mentioned absurd type lamdas earlier. I was wrong, what I actually think I need is absurd constraint lambdas. I opened https://ghc.haskell.org/trac/ghc/ticket/15558 for the exact issue, in case it is just a bug to fix, but am mentioning it here in case it is a true expressiveness gap.

Ericson2314:
OK in https://ghc.haskell.org/trac/ghc/ticket/15558 no one had any ideas for properly eliminating an uninhabited constraint on the term level, so I think that indeed needs some new syntax.

```haskell
foo :: Int ~ Bool -> a
foo = \ @emptyConstraint => case @emptyConstraint of {}

foo :: Int ~ Int -> ...
foo = \ @nonEmptyConstrant => case @nonEmptyConstrant of @canOnlyReBind -> ...
```
Hopefully we can do better?

lexi-lambda:
The limitation this proposal is designed to solve came up in [a Stack Overflow question](https://stackoverflow.com/q/53769147/465378) today. Specifically, it presents the following function:

```haskell
class C a

toto :: (forall s. C s => Int) -> Int
toto k = toto k
```

This function does not typecheck, and it cannot typecheck without the introduction of a `Proxy s` argument to the first argument of `toto`. With this proposal, it could be implemented cleanly as

```haskell
toto k = toto (\ @s -> k @s)
```

(Of course, this function is clearly pretty pointless. I imagine it was reduced from a larger, more useful program for the purposes of asking the Stack Overflow question.)

treeowl:
@lexi-lambda, believe it or not, I was trying to work out whether it was possible to do something similar just today. Sounds like I have my answer. FYI, you can use `Tagged` instead of a proxy, which avoids any extra term-level arguments, but it's pretty ugly.

AshleyYakeley:
@lexi-lambda I'm not sure `toto` can be written even with this proposal. It needs to provide an instance of `C` to its argument, and there isn't one.

If you have an instance of `C`, though, you can provide it (though there is a kind wrinkle that needs to be fixed):
```
class C a

instance C Bool

toto :: (forall (s :: Type). C s => Int) -> Int
toto k = toto $ k @Bool
```

proposals/0000-type-lambda.rst
```diff
@@ -0,0 +1,287 @@
...
+While the specification above is (in my opinion) a complete specification of the proposed behavior with
+respect to the linked papers,
+I include here an expansion of the idea behind bidirectional type checking to aid understanding.
+
+**Motivation**: We need to restrict this feature to the *checking* mode of bidirectional type checking because
+it is unclear (to me) how to do better. Clearly, ``id @a x = x`` is problematic, because we don't know how to
+associate ``a`` with ``x``. But what about ``f @a (x :: a) @b (y :: b) = x == y``? That could indeed be well-typed
+at ``f :: forall a. a -> forall b. b -> (a ~ b, Eq a) => Bool``, but I don't wish to ask GHC to infer that. (Even
+without the wonky equality constraint would be hard.) Perhaps someone can sort this out and expand this feature,
+but there seems to be no need to handle the *checking* case now.
```
phadej: do you mean *inference*?

proposals/0000-type-lambda.rst
```diff
@@ -0,0 +1,287 @@
...
+
+One drawback of this proposal is that it rejects ``id @a (x :: a) = x`` if there is no
+type signature on ``id``. We could imagine extending this feature to pretend that such
+a definition comes with an implicit ``id :: forall a. a -> _`` partial type signature
+and proceeding accordingly. (The partial type signature is created from a quick syntactic
+analysis of the definition.) In this case, the definition of ``id`` would be accepted.
+However, I worry that this would be fragile as the partial-type-signature extraction would
+have to be purely syntactic. For example, would ``null @a ((_ :: a) : _) = False`` be treated
+identically to ``null @a ((_:_) :: [a]) = False`` and ``null @a (_:(_ :: [a]))``? It seems
+hard to ensure. Perhaps I'm just being pessimistic, thnough.
```
phadej: though

proposals/0000-type-lambda.rst
```diff
@@ -0,0 +1,287 @@
...
+worked out design is in `this comment <https://github.com/ghc-proposals/ghc-proposals/pull/155#issuecomment-406024481>`_.
+I'm still not convinced the complication is worth it.
+
+One drawback of this proposal is that it rejects ``id @a (x :: a) = x`` if there is no
+type signature on ``id``. We could imagine extending this feature to pretend that such
+a definition comes with an implicit ``id :: forall a. a -> _`` partial type signature
+and proceeding accordingly. (The partial type signature is created from a quick syntactic
+analysis of the definition.) In this case, the definition of ``id`` would be accepted.
+However, I worry that this would be fragile as the partial-type-signature extraction would
+have to be purely syntactic. For example, would ``null @a ((_ :: a) : _) = False`` be treated
```
phadej: How about: all type-variables have to bound if any?  I think even bad looking cases like `const @a @b x = x` aren't bad. I imagine that type-checker would generate a new type-variable for `x`, but it won't be solved, so it can error with ambiguous type error.

LeanderK:
Is this proposal abandoned? Because I think I have a use-case. I want to do essentially (from @treeowl ):

> you can't even write
> 
> useToto :: (forall s. C s => Int) -> Int
> useToto f = toto f
> which is rather disappointing. Instantiating s to a specific type as @AshleyYakeley suggests is utterly wrong for things like reflection.

the use-case is also reflection. But my formulation is a bit different, it would be:

```haskell
class Get (a::k) where
    get :: k

doReflect :: forall k r. (forall (a::k). Get a => r) -> k - > r
doReflect f k = undefined
```

which I don't think I can write today without proxy.

goldfirere:
No, not abandoned. Neglected, certainly, but I have every intention of returning to it in due course. Unfortunately, that due course may be hard to come by. We'll get there in the end, though.

RyanGlScott:
I have a use case that others have said I should leave here as a motivating example. The [`kind-generics`](http://hackage.haskell.org/package/kind-generics) library (described in the paper _Generics of All Kinds_) has experimental support for datatype-generic programming over data types with rank-_n_ fields. As a consequence of the way `kind-generics` accomplishes this, there are often situations where one needs to write functions like this:

```haskell
f :: T1 -> T2
```

Where `T1` and `T2` are as follows:

```haskell
newtype T1 = MkT1 (forall a. F a)
newtype T2 = MkT2 (forall a. F a)
type family F a
```

It turns out that `f` is surprisingly tricky to implement! One seemingly obvious solution is to try the following:

```haskell
f :: T1 -> T2
f (MkT1 x) = MkT2 x
```

But alas, GHC will flatly reject this with an error about ambiguity:

```
    • Couldn't match expected type ‘F a’ with actual type ‘F a0’
      NB: ‘F’ is a non-injective type family
      The type variable ‘a0’ is ambiguous
    • In the first argument of ‘MkT2’, namely ‘x’
      In the expression: MkT2 x
      In an equation for ‘f’: f (MkT1 x) = MkT2 x
```

Bummer. Pretty much every attempt I've made at implementing `f` ends with a similar demise. The workaround I'm currently using is to change the definition of `T1` and `T2`:

```haskell
newtype WrappedF a = WrapF (F a)
newtype T1 = MkT1 (forall a. WrappedF a)
newtype T2 = MkT2 (forall a. WrappedF a)
```

Having done this, `f (MkT1 x) = MkT2 x` will typecheck. But it's awfully indirect to throw an additional newtype into the mix just for the purpose of resolving ambiguity, and moreover, this workaround requires users of the library to be aware of `WrappedF`, which feels like a failure to separate concerns.

I believe this proposal would offer a better story here. Instead of this `WrappedF` business, we could define `T1` and `T2` as we did before and implement `f` like so:

```haskell
newtype T1 = MkT1 (forall a. F a)
newtype T2 = MkT2 (forall a. F a)

f :: T1 -> T2
f (MkT1 x) = MkT2 (\@a -> x @a)
```

-----

Alternatively, I can summarize this with a somewhat sensationalist description: this proposal would make it possible to derive `Generic` instances for data types with rank-_n_ fields and do useful things with them :)

goldfirere:
I have updated the proposal to fix some typos.

@phadej: I'm not sure what you mean by "How about: all type-variables have to bound if any". My interpretation of that statement leads me right back to my proposal (vis-à-vis bidirectional type checking). Perhaps you mean that the bound type variables are unification variables that the type checker must unify before it's done? I'm not sure.

In any case, conversation here has more than died down, and I think this is ready for submission. The proposal itself has incorporated much of the discussion above, which mainly asked for clarification. The one other thread above is about a network of extensions enabling various related mechanisms. It never all really gelled for me, and in the end, this proposal stands without that framework.

I wish to submit to the committee, @nomeata.

int-index:
One more data point. I've just written this code while working on GHC's parser:

```
            i_ecFrom $ \(Proxy :: Proxy ec) ->
```

This proposal would allow me to write

```
            i_ecFrom $ \ @ec ->
```

simonpj:
I think that this feature should only be available for Specfified type variables, not for Inferred ones.  This is the *same rule* as used for visible type application, and the duality with visible type application is super-important. This proposal is really Visible Type Abstraction.

Example
```
f @a x = x

g :: forall b. b -> b
g @d x = x
```
f's inferred type is `f :: forall {a}. a -> a`, and you cannot use visible type application.  But ty's type is `g :: forall b. b -> b` and you can use visible type application.

This looks as if it's saying the same thing as the whole infer-mode/check-mode thing, but suppose we adopt the [Explicit specificity proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0026-explicit-specificity.rst).  Then what about
```
h :: forall {a}. a -> a
h @b x = x
```
We can't use visible type application for `h` so perhaps we should not be able to use visible type abstraction.

simonpj:
PS.  For 
```
k :: a -> a
k @b x = x
```
Here `a` is a Specified type variable.  It may not scope (because of the lack of a forall) but the order of type variables is [specified in the user manual](http://downloads.haskell.org/~ghc/master/users-guide/glasgow_exts.html#visible-type-application).  So we can use VTA for `k`, and visible type abstraction should be fine too.

It would help if the proposal made this linkage clearer

int-index:
Hooray! I'm happy to see this accepted.

aspiwack:
@simonpj mentionmed, in the committee discussion, that he'd be, personally,  more comfortable if there was more vocal support for this feature. Let me.

I didn't opine, previously, because I didn't think I had much to add to the discussion as I found it. But I must say that, despite it being less often than type-variable binders in patterns, I've needed this quite a few times (though admittedly, not recently).

It also seems, to me, to improve the consistency of the language. So I'm supportive of this feature.

AntC2:
> improve the consistency of the language

There's several proposals kicking about aiming to improve consistency of signatures and bindings. I hope they are consistent with each other. (I can't keep up with them all.)

If "consistency" means removing arbitrary restrictions and special cases, that should lead to less burden of maintenance in GHC eventually(?) Will that persuade SPJ?

I agree with the motivation here of avoiding the explicit `forall` needed for `ScopedTypeVariables`. That strikes me as inconsistent with the usual rule that `forall` is optional-extra and doesn't change the semantics. (Except in `RankNTypes` where I have to remember to pronounce it `exists`.)

After the discussion @nomeata started a year or so ago, I've come round to agreeing that `PatternSignatures` is a sensible, consistent extension of H2010's rules, whereas the `forall` part of `ScopedTypeVariables` is a wart that should be quietly retired.

glaebhoerl:
> @simonpj mentionmed, in the committee discussion, that he'd be, personally, more comfortable if there was more vocal support for this feature. Let me.

And me.

Some time ago I was experimenting with whether it was possible to implement singletons in a CPS/visitor style instead of using GADTs. This was both a success (it *is* possible) and a failure (the ergonomics are, shockingly enough I know, worse). But the relevant point here was that it involved a whole lot of `RankNTypes` code which was *impossible* to write without using `Proxy`s -- a fact which I discovered only after much trial, error, and gnashing of teeth at cryptic error messages. `TypeApplications` was supposed to bring about a world where we don't need to use `Proxy`s any more. I think the fact that, even then, some higher-rank code is *impossible* to write without them is non-obvious and surprising, and we ought to finish the job.

(I guess this is just restating the motivation from the proposal, but, well, it does motivate me.)

LeanderK:
I mentioned my use-case here: https://github.com/ghc-proposals/ghc-proposals/pull/155#issuecomment-459430140

lexi-lambda:
I already commented in this thread a while back, but seeing as I didn’t really express whether or not I was in favor, let me chime in as well and explicitly say I am, too! In addition to the Stack Overflow question I mentioned above, I’ve also needed this feature in my own code at least once.

I think this is one of those things that doesn’t come up often, but when you need it, it’s very frustrating to not have it. In a Haskell with `TypeApplications`, this feature is absolutely natural and, in my mind, clearly belongs.

Icelandjack:
@RyanGlScott Without adding indirection we can `coerce @T1 @T2`

RyanGlScott:
> Without adding indirection we can `coerce @T1 @T2`

In that very simplified example, that trick happens to work out. In the original example I have, it's not quite so simple, as the newtypes look closer to [this](https://gitlab.com/trupill/kind-generics/blob/2779934ba57def7d6cc047173ef41ab04419c627/kind-apply/src/Data/PolyKinded/Atom.hs#L61-65):

```hs
newtype ForAllI (f :: Atom (d1 -> d) (*)) (tys :: LoT d) where
  ForAllI :: (forall t. Interpret f (t ':&&: tys)) -> ForAllI f tys

newtype WrappedI (f :: Atom d (*)) (tys :: LoT d) =
  WrapI { unwrapI :: Interpret f tys }
```

(There are a lot of types involved in this definitions, but don't worry, you don't have to understand what they do in order to connect these definitions back to this GHC proposal.)

`ForAllI` and `WrappedI` are analogous to `T1` and `T2` in https://github.com/ghc-proposals/ghc-proposals/pull/155#issuecomment-460333687. To complete the analogy, we need to define functions that convert from one to the other. Going in one direction is easy:

```hs
toWrappedI :: forall f tys t. ForAllI f tys -> WrappedI f (t ':&&: tys)
toWrappedI (ForAllI x) = WrapI (x @t)
```

The other direction, on the other hand, proves to be much more challenging:

```hs
fromWrappedI :: forall f tys. (forall t. WrappedI f (t ':&&: tys)) -> ForAllI f tys
fromWrappedI = ???
```

For the reasons I mention in https://github.com/ghc-proposals/ghc-proposals/pull/155#issuecomment-460333687, the obvious approach of `fromWrappedI (WrapI x) = ForAllI x` won't work. What's more, even `fromWrappedI = coerce` won't work! If you try implementing it that way, GHC will _still_ be unhappy:

```
    • Couldn't match representation of type ‘Interpret
                                               f (t0 ':&&: tys)’
                               with that of ‘forall (t :: d2). Interpret f (t ':&&: tys)’
        arising from a use of ‘coerce’
      The type variable ‘t0’ is ambiguous
    • In the expression: coerce
      In an equation for ‘fromWrappedI’: fromWrappedI = coerce
   |
   | fromWrappedI = coerce
   |                ^^^^^^
```

The only way to appease the ambiguity checker is to use an explicit type application:

```hs
fromWrappedI = coerce @(forall t. WrappedI f (t ':&&: tys))
```

This technically works, but it comes at a steep price: you must enable `ImpredicativeTypes` in order for this to work. Blegh.

Things would be much, much nicer if we could avoid faffing about with impredicative uses of `coerce` and just define `fromWrappedI` this way:

```hs
fromWrappedI x = ForAllI (\ @t -> unwrapI (x @t))
```

hseg:
Unsure of the correct procedure for this, but the proposal is missing the trac ticket number.
For future reference, this proposal is tracked at https://gitlab.haskell.org/ghc/ghc/issues/17594

nomeata:
Thanks, added to the proposal text.

barci2:
Since this has been moved to #448, shouldn't it receive an "Implemented" label just like #126?

PR #156, opened 2018-07-20
Improved Type Error Messages

NadineTheBean:
[Rendered](https://github.com/NadineTheBean/ghc-proposals/blob/master/proposals/0000-typeerrormessages.rst)

### DIFF
--- a/proposals/0000-typeerrormessages.rst
+++ b/proposals/0000-typeerrormessages.rst
@@ -19,12 +19,10 @@ For the most part, GHC’s current type error messages are visually ineffective.
 
 **Example #1**
 
-Input code
-::
+Input code::
  case1 :: a a
  case1 = undefined
-Original error message:
-::
+Original error message:::
      * Occurs check: cannot construct the infinite kind: k0 ~ k0 -> *
      * In the first argument of `a', namely `a'
        In the type signature: case1 :: a a
--- a/proposals/0000-typeerrormessages.rst
+++ b/proposals/0000-typeerrormessages.rst
@@ -130,7 +130,7 @@ New error message:
  54 | example3 (x,y) = (x, y,"")
     |                        ^^
  
-**NOTE: Several additional examples can be found `here <#additional-examples>`_.**
+NOTE: Several additional examples can be found `here <#additional-examples>`_.
 
 Proposed Change Description
 ---------------------------
--- a/proposals/0000-typeerrormessages.rst
+++ b/proposals/0000-typeerrormessages.rst
@@ -71,8 +71,8 @@ Proposed Change Specification
 The implemented change would involve the following:
 
 **1.) Removal of context phrases beginning with “In the…”**
-      *Ex. “In the expression…”
-      *Ex. “In the equation…”
+      1.) Ex. “In the expression…”
+      2.) Ex. “In the equation…”
 **2.) Reformatting and rewording of expected vs. actual phrases to use tags.** 
 The general format would look something like this for each error message:
 ::
--- a/proposals/0000-typeerrormessages.rst
+++ b/proposals/0000-typeerrormessages.rst
@@ -238,9 +238,10 @@ The implemented change would involve the following:
 
 The general format would look something like this for each error message:
 ::
- Expected something of type [E] but the expression below has type [A].
- [E] (expected type goes here, for example: [Char])
- [A] (actual type goes here, for example: Char)
+ Expected type [E] but the underlined code below has type [A].
+ [E] (expected type goes here)
+ [A] (actual type goes here)
+ (print the line in question and its number here as usual)
 
 **2.) Removal of occurs check statements** 
 
--- a/proposals/0000-typeerrormessages.rst
+++ b/proposals/0000-typeerrormessages.rst
@@ -75,9 +75,10 @@ The implemented change would involve the following:
            Ex. “In the equation…”
 **2.) Reformatting and rewording of expected vs. actual phrases to use tags.** 
 The general format would look something like this for each error message:
-Expected something of type [E] but the expression below has type [A].
-[E] (insert expected type here, for example: [Char])
-[A] (insert actual type here, for example: Char)
+::
+ Expected something of type [E] but the expression below has type [A].
+ [E] (insert expected type here, for example: [Char])
+ [A] (insert actual type here, for example: Char)
 
 **Notes:**
 •   Relevant bindings will be printed as usual.
--- a/proposals/0000-typeerrormessages.rst
+++ b/proposals/0000-typeerrormessages.rst
@@ -118,6 +118,34 @@ New error message:
     |
  32 | case4 :: Maybe
     |          ^^^^^
+     
+     * Expecting one more argument to 'Maybe'
+     
+**Example #5**
+
+Input code:
+::
+ data HighKind :: (* -> *) -> *
+ case5 :: HighKind Either
+ case5 = undefined
+     
+Original error message:
+::
+     * Expecting one more argument to `Either'
+       Expected kind `* -> *', but `Either' has kind `* -> * -> *'
+     * In the first argument of `HighKind', namely `Either'
+       In the type signature: case5 :: HighKind Either
+    |
+ 36 | case5 :: HighKind Either
+    |                   ^^^^^^
+New error message:
+::
+     * Expected a kind [E] but the expression below has kind [A]
+       [E] *
+       [A] * -> *
+    |
+ 36 | case5 :: HighKind Either
+    |                   ^^^^^^
 
 Proposed Change Description
 -----------------------------
--- a/proposals/0000-typeerrormessages.rst
+++ b/proposals/0000-typeerrormessages.rst
@@ -17,6 +17,32 @@ Motivation
 ------------
 For the most part, GHC’s current type error messages are visually ineffective. The messages are often too long and can be difficult for users to process, even when the problem at hand is actually a simple one. This is largely due to their inclusion of redundant information, and their overall structure. With some rewording, reformatting, and the removal of a few phrases, GHC’s type error messages will better facilitate the troubleshooting and overall development processes for users of all levels. This change also has the potential to improve tool integration.
 
+**Example #9**
+
+Input code:
+::
+ case9 :: String -> String
+ case9 xs = True
+     
+Original error message:
+::
+     * Couldn't match type `Bool' with `[Char]'
+       Expected type: String
+         Actual type: Bool
+     * In the expression: True
+       In an equation for `case9': case9 xs = True
+    |
+ 57 | case9 xs = True
+    |            ^^^^
+New error message:
+::
+     * Expected type [E] but the expression below has type [A]
+       [E] String
+       [A] Bool
+    |
+ 57 | case9 xs = True
+    |            ^^^^
+
 **Example #1**
 
 Input code:
@@ -243,32 +269,6 @@ New error message:
     |
  54 | case8 (x,y) = (x, y,"")
     |                     ^^
-
-**Example #9**
-
-Input code:
-::
- case9 :: String -> String
- case9 xs = True
-     
-Original error message:
-::
-     * Couldn't match type `Bool' with `[Char]'
-       Expected type: String
-         Actual type: Bool
-     * In the expression: True
-       In an equation for `case9': case9 xs = True
-    |
- 57 | case9 xs = True
-    |            ^^^^
-New error message:
-::
-     * Expected type [E] but the expression below has type [A]
-       [E] String
-       [A] Bool
-    |
- 57 | case9 xs = True
-    |            ^^^^
 
 Proposed Change Description
 -----------------------------
--- a/proposals/0000-typeerrormessages.rst
+++ b/proposals/0000-typeerrormessages.rst
@@ -128,9 +128,7 @@ The general format would look something like this for each error message:
 
 Effect and Interactions
 -----------------------
-The removal of the extra context phrases will significantly shorten the type error messages, making for more digestible feedback for GHC users. The new tagging format of the “expected vs. actual” phrases serves to establish a clearer distinction between the two mismatched types in question, allowing for easier debugging and tool integration.
-
---removing the context functions will reduce the size of the code!!
+The removal of the extra context phrases will significantly shorten the type error messages, making for more digestible feedback for GHC users. The new tagging format of the “expected vs. actual” phrases serves to establish a clearer distinction between the two mismatched types in question, allowing for easier debugging and tool integration. Additionally, this change will shorten the compiler code by removing several functions currently involved in the printing of the "In the..." context expressions.
 
 
 Costs and Drawbacks
--- a/proposals/0000-typeerrormessages.rst
+++ b/proposals/0000-typeerrormessages.rst
@@ -130,7 +130,7 @@ New error message:
  54 | example3 (x,y) = (x, y,"")
     |                        ^^
  
-NOTE: Several additional examples can be found `here <#additional-examples>`_.
+**NOTE: Several additional examples can be found** `**here** <#additional-examples>`_
 
 Proposed Change Description
 ---------------------------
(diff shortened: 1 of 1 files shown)


AntC2:
> this can go the path of a regular GHC code improvement.

Thanks @nomeata, where in that path would we get feedback from people who read the messages (and are confused by them) as opposed to those who write the messages (who are presumably less confused)?

I agree (some) messages need overhaul -- although in general GHC's messages these days do a pretty good job. I strongly dislike several of the proposals here:

* Putting the tags in `[ ]` is terrible: they look like list constructors.
* `in the underlined code below` doesn't seem less wordy than `In the ...` phrases.
* I'm not particularly in love with those `In the ...` phrases, but it is essential for GHC to narrowly point out which sub-term it's having trouble with. The proposal includes only toy examples, where the location of the problem is obvious. It needs some examples where the equation or declaration is spread over (say) a dozen lines.

Remember that any compiler messages are really a 'best guess'. All the compiler can see is an inconsistency, it doesn't know whether some earlier declaration is right, so this code is wrong vs. this code is right, the declaration was wrong.

This proposal doesn't address what is for me (as a casual Haskeller) the most exasperating about GHC's messages: I made a simple typo; GHC thinks I was trying to write some pointy-headed type-in-type higher-kinded doo-hickey. Try:

    x = length (3, 2, 1)

I get three messages of which the first complains about something called `Foldable`. The other two messages are much longer, something about `Ambiguous type variable`. The only variable I can see in the code is `x`.

If you want examples of clear error messages, try that in Hugs.

nomeata:
> Thanks @nomeata, where in that path would we get feedback from people who read the messages (and are confused by them) as opposed to those who write the messages (who are presumably less confused)?

Beat the drum on the various mailing lists, twitter, reddit, and ask people to weigh into the discussion, I guess.

goldfirere:
I encouraged my student to post here because I've found the quality of feedback and engagement to be very high. In the past, when I've sought feedback from mailing lists, I've either gotten nothing/very little or started threads that quickly went off subject.

You're right, @nomeata, that this is out of scope -- the proposal even says so. Nevertheless, I'm hoping that we can get active feedback here before this is fully implemented and merged and everyone complains.

@AntC2 is also right, in my opinion, that it would be more beneficial to fix the more complex error messages. But we have to start somewhere.

This proposal asks, essentially, for feedback about 4 design choices:

1. Should we drop the "In the ..." lines? To me, these are subsumed by the caret diagnostics at the bottom of the messages these days.

2. Should we start using tags (like `[E]` and `[A]`) in messages? I like these because they nicely separate the boring boilerplate from the part that varies in each error. They also will work well with tool integration. (Of course, there are even better ways to do tool integration, but we're not there yet.)

3. Should we aim to remove technical terms from errors? I'm thinking of "occurs check", "untouchable variable", and "skolem" here, perhaps among others. This one is a harder call (technical terms have precise meanings and may be easier to search for), but I'm generally in favor of this change.

4. Should GHC ever refer to itself as "I"? Now that GHC's [brain](https://stackoverflow.com/questions/23531566/odd-ghc-error-message-my-brain-just-exploded) explode any more, this would be new territory. I think it makes Haskell seem friendlier. You might find it creepy.

So, what do we think?

AntC2:
I agree with @goldfirere this is probably the best venue to get feedback from those who more read than write compiler messages. (I have similar experience from mailing lists, which is sad commentary, but for another venue ...)

> 2. Should we start using tags like (`[E]` and `[A]`) in messages?

I'm neutral on the idea of using tags; strongly against putting them in square brackets, because they look like list types -- especially with the word `type` right beside them, and because currently you might well see a list type. Perhaps:

* Angle brackets, as in BNF or in HTML tags `<E>`, `<A>`.
* Words, then we can accomodate @gbaz's point re _actual_

        * Couldn't match expected type with actual type
          Expected type: String
            Actual type: Bool

Should GHC express a personality in its messages? The `Couldn't match` omits who/what it is doing the matching. `GHC panic` sounds acceptably animate. But yes I do find "I" over the limit: sounds like some marketing/promotion trying to make 'friends' with me.

Avoiding technical terms is certainly valuable. (But that's down to individual messages rather than a general overhaul?) As @gbaz points out, "untouchable" and "rigid" are particularly ornery. I just know them as that error I don't understand but do know how to fix. (There's heaps of questions on StackOverflow about "untouchable"/"rigid"; perhaps we can steal some words from answers there?)

int-index:
I dislike the `[A]` and `[E]` tags.

I like proper English words as @AntC2 suggests:

```
 * Couldn't match expected type with actual type
    Expected type: String
      Actual type: Bool
```

I even think the "Couldn't match expected type with actual type" part of the message is redundant. Words "expected" and "actual" are clear enough. So we can simply have

```
 * Expected type: String
      Actual type: Bool
```

I like shorter error messages: sometimes GHC reports too much and I have to scroll.

int-index:
As to the tooling motivation, I've had enough with tools trying to parse messages intended for humans.

Parsing is the worst, and error messages don't have a formal grammar (and shouldn't have!)

Whoever wants good tooling should stop messing around and start adding JSON error message output format.

nomeata:
I agree that caret positions subsume the In... stuff, and welcome shorter, more concise error messages when it does not lose important information

mitchellwrosen:
FYI, Elm has a similar syntax and has put a lot of time into making nice error messages. We could get inspiration from there. Here's what it outputs for `example1`:

```
-- TYPE MISMATCH ------------------------------------------------------ Main.elm

The definition of `example1` does not match its type annotation.

1| example1 : String -> String
2|>example1 xs = True

The type annotation for `example1` says it always returns:

    String

But the returned value (shown above) is a:

    Bool
```

mitchellwrosen:
What I like about this particular error message: It makes no mention of _expected_ / _actual_, which requires me to think from the perspective of the compiler. I still have some trouble with this, and I much prefer the above, which just tells me what the type annotation is (expected) and what the value's type is (actual). It's wordy, but meaningful to my human brain

mitchellwrosen:
Here's an example of type inference kicking in, and again no mention of _expected_ / _actual_:

```
The right side of (++) is causing a type mismatch.

1|            "Hello" ++ True
                         ^^^^
(++) is expecting the right side to be a:

    String

But the right side is:

    Bool

Hint: With operators like (++) I always check the left side first. If it seems
fine, I assume it is correct and check the right side. So the problem may be in
how the left and right arguments interact.
```

mitchellwrosen:
@NadineTheBean, does your work cover type holes as well? I'd love to see some improvements here :)

```
hylo.hs:93:22: error:
    • Found hole: _ :: r
      Where: ‘r’ is a rigid type variable bound by
               the inferred type of f :: [a1] -> ListF a1 r
               at hylo.hs:(92,3)-(93,22)
    • In the second argument of ‘ConsF’, namely ‘_’
      In the expression: ConsF x _
      In an equation for ‘f’: f (x : xs) = ConsF x _
    • Relevant bindings include
        xs :: [a1] (bound at hylo.hs:93:8)
        x :: a1 (bound at hylo.hs:93:6)
        f :: [a1] -> ListF a1 r (bound at hylo.hs:92:3)
        g :: forall t. t (bound at hylo.hs:95:3)
        rev' :: [a] -> b (bound at hylo.hs:90:1)
      Valid substitutions include
        undefined :: forall (a :: TYPE r).
                     GHC.Stack.Types.HasCallStack =>
                     a
          (imported from ‘Prelude’ at hylo.hs:1:1
           (and originally defined in ‘GHC.Err’))
   |
93 |   f (x:xs) = ConsF x _
   |                      ^
```
I feel this can be simplified to something like:
```
hylo.hs:93:22: error:
    • Found hole

        _ :: r

    • Relevant bindings include

        xs :: [a1]
        x :: a1
        f :: [a1] -> ListF a1 r
        g :: forall t. t
        rev' :: [a] -> b

    • Valid substitutions include
        
        undefined :: forall (a :: TYPE r).
                     GHC.Stack.Types.HasCallStack =>
                     a
   |
93 |   f (x:xs) = ConsF x _
   |                      ^
```

tejon:
Count me among those who have been confused by "expected" and "actual." For example, when I'm stuck writing code that _needs_ a type signature for disambiguation *and I get that signature wrong*, the fact that my wrong signature is the "expected" type throws me off every time.

Elm's errors are perhaps more verbose than I'd prefer, but they're very clear.

int-index:
>  and I get that signature wrong, the fact that my wrong signature is the "expected" type throws me off every time.

@tejon Interesting. Why? For me it's clear: when I write a type signarute, I tell the compiler to _expect_ a term to have that type. Regardless of expectations, there's the _actual_ type which it infers. Both of these types could be correct or incorrect. I find it all very intuitive and never had a problem even as a beginner.

Could you please share your thinking process? Why are these words confusing to you? Can you think of better ones?

Here's one idea: perhaps it would be more clear with words like "specified" and "inferred"? They sound a bit more technical, though.

nomeata:
I too ignore what GHC is labeling as expected and as actual, and try to make sense of it otherwise... so I would probably benefit from a different wording.

linuxgirlie:
> I can't give you example unfortunately.

I believe that this is one of the issues with reviewing type errors. I am currently working on type errors for my PhD and trying to gather "real-life" examples is difficult as most do not keep to share or even push to github when a type error happens, when you do get examples you do not get the "feel" behind why it was difficult to solve, just that it was not understandable. I feel that having such a collection that contained these insights would be beneficial in seeing which errors are most problematic, however, understand that they would be hard to collect!

NadineTheBean:
Thank you for the feedback so far everyone, I really appreciate it! :)

- @mitchellwrosen Yes, this change would affect type holes, but it would only shorten the message by removing the "In the..." statements. I don't plan on touching the rigid type variable/relevant binding statements for now, although I believe that these could be improved in the future as well. As for the Elm syntax, I think that it's wonderfully clear, but I agree with @tejon that the messages are a bit too verbose.

- @tejon @int-index Although they make sense to me, I did consider that the "expected" and "actual" terms cause additional confusion for many users, and have been studying messages from other languages in search of replacements. I think "specified" and "inferred" are too technical and would cause a lot more confusion than "expected" and "actual" (at least for beginners.) Another idea would be to borrow Java's terms, "required" and "found". I like this, but I think pairing "expected" with "found" could be even better because they sort of imply that you need to think about it from the compiler's perspective. Any thoughts on this?

> The proposal includes only toy examples, where the location of the problem is obvious. It needs some examples where the equation or declaration is spread over (say) a dozen lines.
> 
- @AntC2 If it wouldn't be too much trouble, would you be able to provide me with a more complex example? I am still a beginner to Haskell/GHC and usually only encounter simple errors. Additionally, I can see why the [  ] may not be the best idea, and am in favor of changing it to < >.

alanz:
I think the terms `expected` and `actual` are too loaded, many times the mismatch is caused by a bug in the thing that the compiler regards as setting the expected value.

So perhaps some sort of neutral terminology would be better, using `A` and `B` instead of `E` and `A`.

int-index:
What about

```
Could not unify types
    (Bool, Char, Bool, Char, Bool, Char)
                             ^^^^  ^^^^
and
    (Bool, Char, Bool, Bool, Char, Char)
                             ^^^^  ^^^^
```

tejon:
I like `found`. It's unambiguously a verb, which if paired with `expected` implies that's a verb too, whereas alongside `actual` it infers as an adjective. This shifts the tone from denotational to imperative, which feels like it could be enough to shake me out of pure-code reverie. :)

Bonus option: pair `found` with `wanted`?

mgsloan:
How about `outer` and `inner` instead?  Type mismatch is specific to a particular sub-expression, and that sub-expression has an inferred type.  The outer expression, around the sub-expression, has a type that it expects for the hole.

`expected` / `actual` have confused me in the past as well.  I think it's good to have a discussion about improving the clarity of type errors, even if it's not strictly necessary to do it in the ghc-proposals process.  I find github issues to be far more pleasant for discussion than trac, but who knows how much of that is familiarity bias.

mgsloan:
I think I like `expected` better than `wanted`.  But `found` seems much better than `actual`.  How about:

```
     * Expected type ‘String’ but found type ‘Bool’
    |
 57 | example1 xs = True
    |               ^^^^
```

If the types are long, use `[E]` / `[F]`, I like that:

```
     * Expected type [E] but found type [F]
       [E] Proxy "This is a rather large expected type, don't you think?"
       [F] Proxy "One can be small"  
    |
 57 | example1 xs = True
    |               ^^^^
```
Not sure if the prettyprinter supports alternate forms when there's a horizontal overflow, but this is what you'd get in that case.

tejon:
I can definitely get on board with that.

ChShersh:
@mgsloan Even if `found` might be a better word than `actual` I don't like having first letters `F` and `E`. These two letters look very similar. And you might have hard times in attempts to recognise where is `F` and where is `A` on different terminals (with different colors or you might be sleepy or tired). Eventually you will just remember that `found` is on second line. And in that case having `[E]` and `[F]` letters won't really help you, so they're useless. Maybe we can use another words? Like:

* `Expected/Actual` (original)
* `Specified/Inferred` (proposed by @int-index)
* `Wanted/Found` (proposed by @tejon)
* `Goal/Found`
* `Need/Have` (bonus points for same number of letters)
* Any combination of the above or something else.

ChShersh:
My 2 cents into this proposal. If we have code like this:

```haskell
example2 :: (Bool,Char,Bool,Char,Bool,Char) -> (Bool,Char,Bool,Bool,Char,Char)
example2 (a,b,c,d,e,f) = (a,b,c,d,e,f)
```

then it would be cool to have error message like this (inspired by multiple sources):

```haskell
  • Could not unify expected type [E]
      [E] (Bool, Char, Bool, Char, Bool, Char)
                             ^^^^  ^^^^
    and actual type [A]
      [A] (Bool, Char, Bool, Bool, Char, Char)
                             ^^^^  ^^^^
    in the code below:
      C:\Users\Example\Documents\Examples.hs:44:10-15
   ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
43 ┃ example2 :: (Bool,Char,Bool,Char,Bool,Char) -> (Bool,Char,Bool,Bool,Char,Char)
   ┃                                                                ^^^^ ^^^^
44 ┃ example2 (a,b,c,d,e,f) = (a,b,c,d,e,f)
   ┃                                 ┃ ┗━ e :: [E] Char
   ┃                                 ┃         [A] Bool
   ┃                                 ┃
   ┃                                 ┗━ d :: [E] Bool
   ┃                                         [A] Char
```

NadineTheBean:
> My 2 cents into this proposal. If we have code like this:
> 
> ```haskell
> example2 :: (Bool,Char,Bool,Char,Bool,Char) -> (Bool,Char,Bool,Bool,Char,Char)
> example2 (a,b,c,d,e,f) = (a,b,c,d,e,f)
> ```
> 
> then it would be cool to have error message like this (inspired by multiple sources):
> 
> ```haskell
>   • Could not unify expected type [E]
>       [E] (Bool, Char, Bool, Char, Bool, Char)
>                              ^^^^  ^^^^
>     and actual type [A]
>       [A] (Bool, Char, Bool, Bool, Char, Char)
>                              ^^^^  ^^^^
>     in the code below:
>       C:\Users\Example\Documents\Examples.hs:44:10-15
>    ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
> 43 ┃ example2 :: (Bool,Char,Bool,Char,Bool,Char) -> (Bool,Char,Bool,Bool,Char,Char)
>    ┃                                                                ^^^^ ^^^^
> 44 ┃ example2 (a,b,c,d,e,f) = (a,b,c,d,e,f)
>    ┃                                 ┃ ┗━ e :: [E] Char
>    ┃                                 ┃         [A] Bool
>    ┃                                 ┃
>    ┃                                 ┗━ d :: [E] Bool
>    ┃                                         [A] Char
> ```

I like this idea as it provides

PR #157, opened 2018-07-24
Deprecate and remove -XTransformListComp

goldfirere:
[Rendered](https://github.com/goldfirere/ghc-proposals/blob/no-transform/proposals/0000-no-transform.rst)

### DIFF
--- a/proposals/0000-no-transform.rst
+++ b/proposals/0000-no-transform.rst
@@ -1,0 +1,91 @@
+Deprecate and remove ``-XTransformListComp``
+==========================================
+
+.. proposal-number:: Leave blank. This will be filled in when the proposal is
+                     accepted.
+.. trac-ticket:: Leave blank. This will eventually be filled with the Trac
+                 ticket number which will track the progress of the
+                 implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+Almost no one uses ``-XTransformListComp``, yet it takes quite a lot of support to keep it
+running. This proposal suggests to rip it out.
+
+Motivation
+------------
+The extension ``-XTransformListComp`` (described `here <https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#extension-TransformListComp>`_) is barely used on
+Hackage. Specifically, it is used in only 7 packages. Five of these (``llvm-pretty``, ``marxup``, ``manatee-core``, ``nettle-frp``, ``monadiccp``) do not compile on GHC 8.4. The other two
+are ``haddock`` and ``haddock-api``. Each use the extension for precisely one construct.
+
+Yet, there is quite a bit of infrastructure around this extension. Here is its 20-line declaration
+within GHC::
+
+    | TransStmt {
+      trS_ext   :: XTransStmt idL idR body, -- Post typecheck,
+                                            -- R in (>>=) :: Q -> (R -> S) -> T
+      trS_form  :: TransForm,
+      trS_stmts :: [ExprLStmt idL],   -- Stmts to the *left* of the 'group'
+                                      -- which generates the tuples to be grouped
+
+      trS_bndrs :: [(IdP idR, IdP idR)], -- See Note [TransStmt binder map]
+
+      trS_using :: LHsExpr idR,
+      trS_by :: Maybe (LHsExpr idR),  -- "by e" (optional)
+        -- Invariant: if trS_form = GroupBy, then grp_by = Just e
+
+      trS_ret :: SyntaxExpr idR,      -- The monomorphic 'return' function for
+                                      -- the inner monad comprehensions
+      trS_bind :: SyntaxExpr idR,     -- The '(>>=)' operator
+      trS_fmap :: HsExpr idR          -- The polymorphic 'fmap' function for desugaring
+                                      -- Only for 'group' forms
+                                      -- Just a simple HsExpr, because it's
+                                      -- too polymorphic for tcSyntaxOp
+    }                                 -- See Note [Monad Comprehensions]
+
+As you might imagine, parsing, renaming, and type checking this beast is quite non-trivial.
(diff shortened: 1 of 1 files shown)


nomeata:
What are the odds that this PR will get the extension on the radar of more people, gets them excited, and suddently we see an increase of uses of it :-)

treeowl:
I generally agree, but with one caveat. The least-based semantics of this extension strike me as unfortunate: lists don't support efficient SQL queries! But what if we yank that out and keep the syntax? Could this be good syntax for constructing actual database queries?

int-index:
Given the severe feature bloat of GHC, I think this proposal is well motivated. I especially welcome simplifications to the lexer and parser.

mgsloan:
I definitely think it makes sense to reduce complexity when there is evidence that a feature isn't worth its weight.  Recently when working on a change in GHC related to do notation I did a double take for `TransStmt`.  Once I dug into what it was, it was clear and I recalled the feature, but it's easy to forget that it's a thing.

When it is removed, ideally when code / cli tries to use the extension, GHC should explain that it has been removed.  Perhaps even with a mini migration guide!

goldfirere:
@nomeata: I wondered the same thing. I'd actually be fine with that result. I have no animus against this feature -- it's the wild disproportionality between the effort to maintain the feature and its uptake in the (visible/open source) community that's irksome. If either we lower that effort (e.g., remove the feature) or have people adopt it, then the problem goes away.

simonpj:
Boo hoo.  The paper turned out much more lovely than I expected, so I'd be sad to see it go.

I accept that it's not used much, more's the pity.  But I mildly question how much pain there is in keeping it.   I cant' remember any time in the last few years when I've experienced the maintenance burden.  It seems to "just work".

AntC2:
> yet it takes quite a lot of support to keep it running.

Does it? The proposal mentions volume of code, but that doesn't need 'support' in any active sense. Has this code ever interfered with developing other features, or broken upgrades? Whereas it _will_ take effort to document/warn of deprecatiing it and root out the code.  Simon's

> It seems to "just work".

Would suggest we can let sleeping dogs lie.

Perhaps one day GHC will have a decent records system such that people would consider using GHC for database-like work. We'd probably want set comprehensions rather than list comprehensions, but the Transforms idea would still apply.

gbaz:
In general I feel like there's not sufficient power-to-syntactic-weight to make the feature worth using as a consumer. The fundamental limitation is that plain old grouping, etc. on lists isn't very efficient and _also_ when it is used is easy to write and read in the standard functional style, so the sugar tends to obscure rather than clarify.

Arguably, the power of linq is A) from IDE support and B) from is more purely syntactic nature that has allowed all manner of strange overloading. And for the latter, part of the power has been that it takes place in a language that _doesn't_ have such nice and pervasive support for DSLs in a purely functional style.

In the presence of monad comprehensions, things become a bit more interesting, since you can provide arbitrary functions `m a -> n a` or `m a -> n n1 a` (for any choice of monad n, n1). But again, you don't get much more than function composition, in reverse order, with some rather unusual choices of names and syntax.

One could imagine, perhaps, using this to construct a sql-like AST that then performed execution efficiently via some other backend. But the use of this syntax directly sort of gets in the way of that, leading, as far as I know, Database Supported Haskell (https://github.com/ulricha/dsh) to need rebindable syntax anyway.

So from my standpoint it introduces a lot of hoops to jump through to get code that it harder to reason about (because more unfamiliar) than more standard dsl-like techniques.

Perhaps @jstolarek or @giorgidze would like to weigh in regarding DSH? If there's hope that DSH-style-work could find a way to put this syntax to good use, that would be a compelling reason to keep it, despite its current low usage.

(my notes above indicate by the way that the reason this didn't get pickup, I think, is that Haskell is _too nice_ a language already, but also that the utility of this work as research in clarifying how to do linq-ish things in less-nice languages is nonetheless worthwhile :-) )

gbaz:
Here is some other recent work (icfp 2018) making use of `TransformListComp` in conjunction with generalized comprehensions: https://dl.acm.org/citation.cfm?id=3236781 )

I think it is premature to decide this line of development is foreclosed just because of low current adoption.

istandleet:
Please no! TransformListComp is very nice with data analysis for concise repl based data manipulation. I use it just about every day at my job. There are the trivial:

    count xs = [(the x, length x)| x <- xs, then group by x using groupWith]

To more complex:

    findGaps :: (Eq a, Ord b, Enum b) => [(a,b)] -> [a]
    findGaps xys = 
        [ the x
        | (x,y) <- xys
        , then group by x using groupWith
        , let ys = sort y
        , and $ zipWith (==) (map inc ys) (tail ys)]

I could go on with examples, and will if desired, but I'm sure the paper has plenty. One of the ways I got my fellow data analysts to look at using Haskell for small tasks was showing the power of TransformListComp over the base python list comprehensions. 

I'm a bit surprised how underadopted TransformListComp is, but I would chalk it up to the focus on compiled Haskell (IHaskell doesn't support windows!). TransformListComp is inherently declarative, and I'm sure for production processes something like

     count = Map.unionsWith (+) . map (`Map.singleton` 1) 

    findGaps = 
        Map.filter noGap .
        fmap sort .
        Map.unionsWith mconcat .
        fmap (:[]) .
        map (uncurry Map.singleton)
      where noGap ys = and $ zipWith (==) (map inc ys) (tail ys)

Are better. But TransformListComp can be more legible (to me) for many problems, and for "execute once" repl code it's really nice to write.

goldfirere:
The two previous posts are very convincing to me. I withdraw the proposal.

The number of people who upvoted is greater than the number who downvoted. If someone else would like to adopt and re-open this proposal, please do so. I explicitly grant permission to use my proposal text.

simonpj:
I'm very happy to hear about these use-cases for `TransformListComp`!

I agree that it an under-adopted future.  Would any of you who are using `TransformListComp` like to make a blog post, or add a page to the GHC collaborative documentation describing how and why you use it?  https://wiki.haskell.org/GHC

proposals/0000-no-transform.rst
```diff
@@ -0,0 +1,89 @@
...
+Effect and Interactions
+-----------------------
+Many lines of code deleted.
+
+Costs and Drawbacks
+-------------------
+The `original paper <https://www.microsoft.com/en-us/research/wp-content/uploads/2007/09/list-comp.pdf>`_ was written by two prominent members of our community, and it would be sad if they felt
+discouraged from contributing in the future.
+
+Alternatives
```
ocharles: I *think* this could actually be implemented as a source plugin. I can't confirm that, but maybe it should be mentioned here.

PR #160, opened 2018-08-15
NoFieldSelectors

reactormonk:
This is the proposal to remove toplevel field selector generation - e.g. the toplevel functions `bar` and `baz` that are being generator from the declaration `data Foo = Foo { bar :: String, baz :: Int}`.

[Rendered](https://github.com/reactormonk/ghc-proposals/blob/master/proposals/0000-no-toplevel-field-selectors.rst)

### DIFF
--- a/proposals/0000-no-toplevel-field-selectors.rst
+++ b/proposals/0000-no-toplevel-field-selectors.rst
@@ -36,6 +36,7 @@ Possible Alternative Designs
 - namespaced lenses
 - overloaded-labels based accessors / lenses
 - row types
+- https://github.com/ghc-proposals/ghc-proposals/pull/158
 - ...
 
 Proposed Change Specification
@@ -123,9 +124,6 @@ None.
 
 Unresolved questions
 --------------------
-
-Should exporting ``Foo(..)`` also export functions based on the name of the
-field accessors?
 
 
 Implementation Plan
--- a/proposals/0000-no-toplevel-field-selectors.rst
+++ b/proposals/0000-no-toplevel-field-selectors.rst
@@ -50,9 +50,8 @@ by default to match the current behavior, but may be disabled per-module with
 the ``NoFieldSelectors`` language pragma.
 
 When ``NoFieldSelectors`` is enabled, Record construction/update syntax and
-pattern matching will work as before, and the behavior of
-``DuplicateRecordFields`` will be enabled by implication. The
-``DuplicateRecordFields`` will be deprecated in favor of ``NoFieldSelectors``.
+pattern matching will work as before, as will the disambiguation handled by
+``DuplicateRecordFields``.
 
 Record fields will still be part of a ``Record(..)`` export, or can also be
 named individually. They always have to be associated with the data type though,
@@ -192,11 +191,6 @@ Effect and Interactions
 
 `HasField` will work as before, if the corresponding field has been exported. It
 doesn't need to be exported as function.
-
-Interaction with DuplicateRecordFields
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-Setting ``NoFieldSelectors`` implies ``DuplicateRecordFields``.
 
 Breakage estimation
 ^^^^^^^^^^^^^^^^^^^
--- a/proposals/0000-no-toplevel-field-selectors.rst
+++ b/proposals/0000-no-toplevel-field-selectors.rst
@@ -55,17 +55,17 @@ pattern matching will work as before, as will the disambiguation handled by
 export, or can also be named individually. They always have to be associated
 with the data type though, because there is no more toplevel selector (see `Example`_).
 
-A Template Haskell function ``mkFieldSelector`` to go from a record field name to a selector for
-that field will be provided, because it's not possible anymore to go from field
-name directly to selector.  It takes a ``Name`` (for the constructor) and
-a ``String`` (for the field) and returns a ``Q Exp``.
+A new function will be added to Template Haskell, where the ``Name`` is a
+reference to a constructor. This function should be used in new TH even if this
+extension isn't enabled. It will fail if it's not possible to create a valid
+selector / selector doesn't exist.
 
 .. code-block:: haskell
 
     mkFieldSelector :: Name -> String -> Q Exp
 
-Example
-^^^^^^^
+Examples
+--------
 
 Given a data structure
 
@@ -76,7 +76,7 @@ The following will be available:
 1. the type constructor ``Foo``
 2. the data constructor ``Foo``
 3. the fields ``bar`` and ``baz`` for record construction, update, and patterns
-4. the two functions ``bar`` and ``baz``
+4. the two functions ``bar`` and ``baz``, which are ``Foo -> Int`` and ``Foo -> String``
 
 If the language extension ``NoFieldSelectors`` is enabled, items (1), (2), and (3)
 will still be generated, but (4) will not.
@@ -98,9 +98,30 @@ record. Without ambiguitiy, previously this was equivalent
 
 Under ``NoFieldSelectors``, these two export statements are now different. The
 first one will export the field ``baz``, but not the function ``baz``, while the
-second one will export the function ``baz`` (assuming one is defined), but not the field ``baz``. Because
-of this change, writing out all selector functions by hand is still different,
-because they all have to be exported separately.
+second one will export the function ``baz`` (assuming one is defined), but not
+the field ``baz``. Because of this change, writing out all selector functions by
+hand is still different, because they all have to be exported separately.
+
+.. code-block:: haskell
+
+    {-# LANGUAGE NoFieldSelectors #-}
+    module Exports where (Foo(Foo, bar, baz))
+    data Foo = Foo { bar :: Int, baz :: Int }
+
+    bar (Foo x _) = x
+    baz (Foo _ x) = x
+
+is different from
+
+.. code-block:: haskell
+
+    module Exports where (Foo(Foo, bar, baz))
+    data Foo = Foo { bar :: Int, baz :: Int }
+
+Because the functions in the furst example don't get exported.
+
+Let's take a module ``A`` with a function with the same name as a field, with
+the extension enabled:
 
 .. code-block:: haskell
 
@@ -119,7 +140,8 @@ Which would be equivalent to:
     baz = 42
 
 A second module, ``B``, which does not export the selector ``baz`` of
-constructor ``Foo``, but instead exports the toplevel binder ``baz``.
+constructor ``Foo``, but instead exports the toplevel binder ``baz``. The fields
+can still be used when exported (as in module ``A``).
 
 .. code-block:: haskell
 
@@ -128,16 +150,8 @@ constructor ``Foo``, but instead exports the toplevel binder ``baz``.
     data Foo = Foo { bar :: Int, baz :: Int }
     baz = 42
 
-The fields can still be used when exported (as in module ``A``).
-
-.. code-block:: haskell
-
-    import A
-    foo = Foo 23 42
-    foo { baz = 1 }
-
-This will now fail, because the field ``baz`` is not in scope anymore,
-because it is not exported by ``B``.
+Using ``baz`` when importing ``B`` will fail, because the field ``baz`` is not
+in scope anymore, because it is not exported by ``B``.
 
 .. code-block:: haskell
 
@@ -145,11 +159,11 @@ because it is not exported by ``B``.
     foo = Foo 23 42
     foo { baz = 1 }
 
-The value ``baz`` is only exported from module ``B``, not ``A``. This would fail:
+However, it is possible to use the imported variable ``baz``, because ``B`` exports it.
 
 .. code-block:: haskell
 
-    import A
+    import B
     main = print baz
 
 If you wanted to use both, you'd have to export both explicitly:
@@ -169,18 +183,6 @@ Now ``baz`` here assigns the value ``42`` to the field ``baz``.
     foo = Foo 23 1
     foo { baz = baz }
 
-
-Template Haskell
-^^^^^^^^^^^^^^^^
-
-A new function will be added to Template Haskell, where the ``Name`` is a
-reference to a constructor. This function should be used in new TH even if this
-extension isn't enabled. It will fail if it's not possible to create a valid
-selector / selector doesn't exist.
-
-.. code-block:: haskell
-
-    mkFieldSelector :: Name -> String -> Q Exp
 
 Effect and Interactions
 -----------------------
@@ -234,7 +236,10 @@ None.
 Unresolved questions
 --------------------
 
-- Which order of arguments in the new TH function?
+- What would TH code look like in the future? The disambiguating field name is
+  always the first constructor, maybe introducing a new `Newtype` might make it
+  easier to use the function? Or should all constructors work, by having TH
+  normalize to the first one?
 - Should this extension imply DuplicateRecordFields?
 
 
--- a/proposals/0000-no-toplevel-field-selectors.rst
+++ b/proposals/0000-no-toplevel-field-selectors.rst
@@ -15,8 +15,7 @@ NoToplevelFieldSelectors
 
 Enabling this Language Extension removes the default of Haskell data
 declarations to generate toplevel field accessor functions for records, such
-that the user can define their own, either via OverloadedRecordFields, or via
-Generic.
+that the user can define their own via Generic or positional extraction.
 
 Motivation
 ------------
@@ -117,6 +116,9 @@ because they all have to be exported manually.
 Effect and Interactions
 -----------------------
 
+`HasField` will work as before, if the corresponding field has been exported. It
+doesn't need to be exported as function.
+
 Breakage estimation
 ^^^^^^^^^^^^^^^^^^^
 
@@ -149,6 +151,10 @@ None.
 
 Unresolved questions
 --------------------
+
+- How exactly should the implementation look like?
+- How should the differentiation on TH work? There now needs to be a different
+  function to look up a selector function and a regular function via `Name`.
 
 
 Implementation Plan
--- a/proposals/0000-no-toplevel-field-selectors.rst
+++ b/proposals/0000-no-toplevel-field-selectors.rst
@@ -132,7 +132,7 @@ constructor ``Foo``, but instead exports the toplevel bind ``baz``.
     data Foo = Foo { bar :: Int, baz :: Int }
     baz = 42
 
-The updaters can still be used when exported (as in module ``A``).
+The fields can still be used when exported (as in module ``A``).
 
 .. code-block:: haskell
 
@@ -155,6 +155,24 @@ The value ``baz`` is only exported from module ``B``, not ``A``. This would fail
 
     import A
     main = print baz
+
+If you wanted to use both, you'd have to export both explicitly:
+
+.. code-block:: haskell
+
+    {-# LANGUAGE NoFieldSelectors #-}
+    module C where (Foo(Foo, bar, baz), baz)
+    data Foo = Foo { bar :: Int, baz :: Int }
+    baz = 42
+
+Now ``baz`` here assigns the value ``42`` to the field ``baz``.
+
+.. code-block:: haskell
+
+   import C
+    foo = Foo 23 1
+    foo { baz = baz }
+
 
 Template Haskell
 ^^^^^^^^^^^^^^^^
--- a/proposals/0000-no-toplevel-field-selectors.rst
+++ b/proposals/0000-no-toplevel-field-selectors.rst
@@ -1,0 +1,128 @@
+NoToplevelFieldSelectors
+==============
+
+.. proposal-number:: Leave blank. This will be filled in when the proposal is
+                     accepted.
+.. trac-ticket:: Leave blank. This will eventually be filled with the Trac
+                 ticket number which will track the progress of the
+                 implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+Enabling this Language Extension removes the default of Haskell data
+declarations to generate toplevel field accessor functions for records, such
+that the user can define their own, either via OverloadedRecordFields, or via
+Generic.
+
+Motivation
+------------
+
+There have been proposals to extend the usage of records in Haskell
+(DuplicateRecordFields, OverloadedRecordFields, DeriveGeneric to name a few),
+but none of them address the original issue where record fields steal the
+toplevel function name for each field for selector purposes. This proposal
+intends to open the design space to this issue by removing the generation of
+these toplevel selector functions.
+
+Proposed Change Specification
+-----------------------------
+
+Record definition no longer defines toplevel selector functions when the
+``NoToplevelFieldSelectors`` extension is enabled. A
+``NoToplevelFieldSelectors`` pragma for records will also be available, such
+that this extension can be enabled on a per-record basis.
+
+Record construction/update syntax and pattern matching will work as before, the
+disambiguation handled by ``DuplicateRecordFields``. The necessary selectors
+will be part of a ``Record(..)`` export, or can also be named one by one. They
+always have to be associated with the data type though, because there is no more
+toplevel selector.
+
+A function for Template Haskell to go from a record field name to a selector for
+that field will be provided, because it's not possible anymore to go from field
+name directly to selector.
+
+Example
+^^^^^^^
+
+Given a data structure
+
+    data Foo = Foo { bar :: Int, baz :: String }
+
+The following will be available:
+
+- the type ``Foo``
+- the constructor ``Foo``
+- the two functions ``bar`` and ``baz``
+- the names ``bar`` and ``baz`` for record construction (``Foo { bar = 3, baz = "foo" }``)
+- the names ``bar`` and ``baz`` for ``RecordWildCards``
+
+If the language extension ``NoToplevelFieldSelectors`` is enabled for the module
+or ``Foo`` specifically, all of the above will be generated, except for the two
+functions ``bar`` and ``baz``.
+
+Wildcard exports will work as before, except for the two functions. Even if
+these functions are otherwise defined, the wildcard will not export them.
+Exporting the names for record construction now has to be specific to the
+record. Without ambiguitiy, previously this was equivalent
+
+    module A where (Foo(Foo, bar, baz))
+
+    module A where (Foo(Foo, bar), baz)
+
+Because of the new semantics, these two export statements are now different.
+Maybe a wildcard export will export functions of the same name as well, this
+question isn't resolved yet.
+
+Effect and Interactions
+-----------------------
+
+Breakage estimation
+^^^^^^^^^^^^^^^^^^^
+
+Enabling this extension will break a lot of Template Haskell. Going from record
+field name to selector won't work anymore. A new way to go from record field
+name to selector has to be found.
+
+Anything that generates code with the help of Generic should be fine. The same
+functionality that generates the anonymous functions for Generic could be used
+to provide TH functionality to replace the existing toplevel functions.
+
+The record extensions NamedFieldPuns, RecordWildCards, DisambiguateRecordFields,
+and DuplicateRecordFields are unaffected by this change.
+
+
+Costs and Drawbacks
+-------------------
+
+This might cause some confusion that record fields can't be accessed by toplevel
+selectors anymore - however, that shouldn't be too big of an issue, because some
+library authors already stopped exporting these selectors so they don't have to
+break downstream software on record changes.
+
+
+Alternatives
+------------
+
+None.
+
+
+Unresolved questions
+--------------------
+
+Should exporting ``Foo(..)`` also export functions based on the name of the
+field accessors?
+
+
+Implementation Plan
+-------------------
+
+For deriving the Generic instances, either ``Record.fieldName``,
+``OverloadedRecordFields`` or an anonymous function which generates the
+corresponding selector in core will be used. (to be clarified)
--- a/proposals/0000-no-toplevel-field-selectors.rst
+++ b/proposals/0000-no-toplevel-field-selectors.rst
@@ -45,9 +45,7 @@ Proposed Change Specification
 -----------------------------
 
 Record definition no longer defines toplevel selector functions when the
-``NoToplevelFieldSelectors`` extension is enabled. A
-``NoToplevelFieldSelectors`` pragma for records will also be available, such
-that this extension can be enabled on a per-record basis.
+``NoToplevelFieldSelectors`` extension is enabled.
 
 Record construction/update syntax and pattern matching will work as before, the
 disambiguation handled by ``DuplicateRecordFields``. The necessary selectors
@@ -81,26 +79,6 @@ The following will be available:
 If the language extension ``NoToplevelFieldSelectors`` is enabled for the module
 or ``Foo`` specifically, all of the above will be generated, except for the two
 functions ``bar`` and ``baz``.
-
-The following declarations would be equivalent:
-
-::
-
-    {-# LANGUAGE NoToplevelFieldSelectors #-}
-
-    data Foo = Foo { bar :: Int, baz :: String }
-
-::
-
-    {-# NoToplevelFieldSelectors #-}
-    data Foo = Foo { bar :: Int, baz :: String }
-
-::
-
-    data Foo = Foo { bar :: Int, baz :: String }
-    {-# NoToplevelFieldSelectors Foo #-}
-
-
 
 Wildcard exports will work as before, except for the two functions. Even if
 these functions are otherwise defined, the wildcard will not export them.
--- a/proposals/0000-no-toplevel-field-selectors.rst
+++ b/proposals/0000-no-toplevel-field-selectors.rst
@@ -59,6 +59,10 @@ A function for Template Haskell to go from a record field name to a selector for
 that field will be provided, because it's not possible anymore to go from field
 name directly to selector.
 
+A new `TH` function is added which takes two `Name` and returns an `Exp` which
+is equivalent to `a -> b`, where `a` is the record type specified by the second
+argument, and `b` the contents of the field specified by the first argument.
+
 Example
 ^^^^^^^
 
@@ -153,8 +157,7 @@ Unresolved questions
 --------------------
 
 - How exactly should the implementation look like?
-- How should the differentiation on TH work? There now needs to be a different
-  function to look up a selector function and a regular function via `Name`.
+- Which order of arguments in the new TH function?
 
 
 Implementation Plan
--- a/proposals/0000-no-toplevel-field-selectors.rst
+++ b/proposals/0000-no-toplevel-field-selectors.rst
@@ -45,24 +45,27 @@ Proposed Change Specification
 -----------------------------
 
 This proposal introduces a new extension ``FieldSelectors`` that controls the
-generation of record field selector functions. The extension is enabled by
-default to match the current behavior, but may be disabled per-module with the
-``NoFieldSelectors`` language pragma.
+generation of toplevel record field selector functions. The extension is enabled
+by default to match the current behavior, but may be disabled per-module with
+the ``NoFieldSelectors`` language pragma.
 
 When ``NoFieldSelectors`` is enabled, Record construction/update syntax and
-pattern matching will work as before, as will the disambiguation handled by
-``DuplicateRecordFields``. Record fields will still be part of a ``Record(..)``
-export, or can also be named individually. They always have to be associated
-with the data type though, because there is no more toplevel selector (see `Example`_).
+pattern matching will work as before, and the behavior of
+``DuplicateRecordFields`` will be enabled by implication. The
+``DuplicateRecordFields`` will be deprecated in favor of ``NoFieldSelectors``.
 
-A new function will be added to Template Haskell, where the ``Name`` is a
-reference to a constructor. This function should be used in new TH even if this
-extension isn't enabled. It will fail if it's not possible to create a valid
-selector / selector doesn't exist.
+Record fields will still be part of a ``Record(..)`` export, or can also be
+named individually. They always have to be associated with the data type though,
+because there is no more toplevel selector (see `Example`_).
 
-.. code-block:: haskell
+Template Haskell should not rely on the selectors being present, and should use
+named pattern matching instead.
 
-    mkFieldSelector :: Name -> String -> Q Exp
+Because field labels and toplevel selectors are now different entities,
+import/export lists now behave differently. Names listed with the constructor
+(e.g. ``Record(field)``) refer to the field, whereas direct mentions ``field``
+refer to a function named ``field``. Without setting ``NoFieldSelectors``, these
+two would be equivalent.
 
 Examples
 --------
@@ -193,22 +196,13 @@ doesn't need to be exported as function.
 Interaction with DuplicateRecordFields
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Because of Record updates still being valid, this code will still fail to
-compile without DuplicateRecordFields:
-
-.. code-block:: haskell
-
-    {-# LANGUAGE NoFieldSelectors #-}
-
-    data Foo = Foo { foo :: Int }
-    data Bar = Bar { foo :: Int }
+Setting ``NoFieldSelectors`` implies ``DuplicateRecordFields``.
 
 Breakage estimation
 ^^^^^^^^^^^^^^^^^^^
 
-Enabling this extension will break a lot of Template Haskell. Going from record
-field name to selector won't work anymore. A new way to go from record field
-name to selector has to be found.
+Enabling this extension will beak Template Haskell which assumes the presence of
+ a field selector. Use named pattern matching instead.
 
 Anything that generates code with the help of Generic should be fine. The same
 functionality that generates the anonymous functions for Generic could be used
@@ -231,16 +225,6 @@ Alternatives
 ------------
 
 None.
-
-
-Unresolved questions
---------------------
-
-- What would TH code look like in the future? The disambiguating record
-  constructor is always the first constructor, maybe introducing a new `Newtype`
-  might make it easier to use the function? Or should all constructors work, by
-  having TH normalize to the first one?
-- Should this extension imply DuplicateRecordFields?
 
 
 Implementation Plan
--- a/proposals/0000-no-toplevel-field-selectors.rst
+++ b/proposals/0000-no-toplevel-field-selectors.rst
@@ -28,8 +28,11 @@ toplevel function name for each field for selector purposes. This proposal
 intends to open the design space to this issue by removing the generation of
 these toplevel selector functions.
 
-Possible Alternative Designs
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Possible Alternative Use Cases for Record Names
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+After removing the toplevel selector names, the field names could be used as
+bindings for other values.
 
 - A `generic-lens` equivalent for toplevel lenses
 - namespaced accessors
--- a/proposals/0000-no-toplevel-field-selectors.rst
+++ b/proposals/0000-no-toplevel-field-selectors.rst
@@ -109,18 +109,39 @@ because they all have to be exported manually.
 .. code-block:: haskell
 
     {-# LANGUAGE NoFieldSelectors #-}
-    module A where (Foo(Foo, bar), baz)
+    module A where (Foo(Foo, bar, baz))
     data Foo = Foo { bar :: Int, baz :: Int }
     baz = 42
+
+Which would be equivalent to:
 
 .. code-block:: haskell
 
     {-# LANGUAGE NoFieldSelectors #-}
-    module B where (Foo(Foo, bar, baz))
+    module A where (Foo(..))
     data Foo = Foo { bar :: Int, baz :: Int }
     baz = 42
 
-This will now fail, because the record updater ``baz`` is not in scope anymore:
+A second module, ``B``, which does not export the selector ``baz`` of
+constructor ``Foo``, but instead exports the toplevel bind ``baz``.
+
+.. code-block:: haskell
+
+    {-# LANGUAGE NoFieldSelectors #-}
+    module B where (Foo(Foo, bar), baz)
+    data Foo = Foo { bar :: Int, baz :: Int }
+    baz = 42
+
+The updaters can still be used when exported (as in module ``A``).
+
+.. code-block:: haskell
+
+    import A
+    foo = Foo 23 42
+    foo { baz = 1 }
+
+This will now fail, because the record updater ``baz`` is not in scope anymore,
+because the selector is not exported by ``B``.
 
 .. code-block:: haskell
 
@@ -128,16 +149,24 @@ This will now fail, because the record updater ``baz`` is not in scope anymore:
     foo = Foo 23 42
     foo { baz = 1 }
 
+The value ``baz`` is only exported from module ``B``, not ``A``. This would fail:
+
+.. code-block:: haskell
+
+    import A
+    main = print baz
+
 Template Haskell
 ^^^^^^^^^^^^^^^^
 
 A new function will be added to Template Haskell, where the ``Name`` is a
 reference to a constructor. This function should be used in new TH even if this
-extension isn't enabled.
+extension isn't enabled. It will fail if it's not possible to create a valid
+selector / selector doesn't exist.
 
 .. code-block:: haskell
 
-    lookupFieldName :: Name -> String -> Q (Maybe Name)
+    mkFieldSelector :: Name -> String -> Q Exp
 
 Additionally, ``NameSpace`` will be extended with a new constructor ``FieldName``.
 
--- a/proposals/0000-no-toplevel-field-selectors.rst
+++ b/proposals/0000-no-toplevel-field-selectors.rst
@@ -134,13 +134,21 @@ None.
 Unresolved questions
 --------------------
 
-- How exactly should the implementation look like?
 - Which order of arguments in the new TH function?
 
 
 Implementation Plan
 -------------------
 
-For deriving the Generic instances, either ``Record.fieldName``,
-``OverloadedRecordFields`` or an anonymous function which generates the
-corresponding selector in core will be used. (to be clarified)
+I'm currently on the way of implementing this extension. It's roughly as
+follows:
+
+- Add new `NameSpace` to `OccName`: `RecordSelector String`
+- Remove `flSelector` from `FieldLabel`, add an flag which denotes if it should
+  be found as `VarName`
+- Remove `FlParent`
+- Change any field lookup code to look for new `OccName`
+- Implement `ToplevelFieldSelector` flag to look for selectors if you're looking
+  for `VarName`
+- Adjust `Generic` instances
+- Add new `TH` function to access record selectors
--- a/proposals/0000-no-toplevel-field-selectors.rst
+++ b/proposals/0000-no-toplevel-field-selectors.rst
@@ -1,4 +1,4 @@
-NoToplevelFieldSelectors
+NoFieldSelectors
 ==============
 
 .. proposal-number:: Leave blank. This will be filled in when the proposal is
@@ -38,28 +38,29 @@ bindings for other values.
 - namespaced lenses
 - overloaded-labels based accessors / lenses
 - row types
-- https://github.com/ghc-proposals/ghc-proposals/pull/158
+- `<https://github.com/ghc-proposals/ghc-proposals/pull/158>`_
 - ...
 
 Proposed Change Specification
 -----------------------------
 
 Record definition no longer defines toplevel selector functions when the
-``NoToplevelFieldSelectors`` extension is enabled.
+``NoFieldSelectors`` extension is enabled. The default is ``FieldSelectors``,
+which implies the current behavior, that selector functions are being generated.
 
 Record construction/update syntax and pattern matching will work as before, the
 disambiguation handled by ``DuplicateRecordFields``. The necessary selectors
-will be part of a ``Record(..)`` export, or can also be named one by one. They
+will be part of a ``Record(..)`` export, or can also be named individually. They
 always have to be associated with the data type though, because there is no more
-toplevel selector.
+toplevel selector (see `Example`_).
 
 A function for Template Haskell to go from a record field name to a selector for
 that field will be provided, because it's not possible anymore to go from field
 name directly to selector.
 
-A new `TH` function is added which takes two `Name` and returns an `Exp` which
-is equivalent to `a -> b`, where `a` is the record type specified by the second
-argument, and `b` the contents of the field specified by the first argument.
+A new ``TH`` function is added which takes a ``Name`` (for the constructor) and
+a ``String`` (for the field) and returns a ``Maybe Name``, just like
+``lookupValueName``.
 
 Example
 ^^^^^^^
@@ -76,7 +77,7 @@ The following will be available:
 - the names ``bar`` and ``baz`` for record construction (``Foo { bar = 3, baz = "foo" }``)
 - the names ``bar`` and ``baz`` for ``RecordWildCards``
 
-If the language extension ``NoToplevelFieldSelectors`` is enabled for the module
+If the language extension ``NoFieldSelectors`` is enabled for the module
 or ``Foo`` specifically, all of the above will be generated, except for the two
 functions ``bar`` and ``baz``.
 
@@ -85,9 +86,13 @@ these functions are otherwise defined, the wildcard will not export them.
 Exporting the names for record construction now has to be specific to the
 record. Without ambiguitiy, previously this was equivalent
 
+.. code-block:: haskell
     module A where (Foo(Foo, bar, baz))
+    data Foo = Foo { bar :: Int, baz :: Int }
 
-    module A where (Foo(Foo, bar), baz)
+.. code-block:: haskell
+    module B where (Foo(Foo, bar), baz)
+    data Foo = Foo { bar :: Int, baz :: Int }
 
 Because of the new semantics, these two export statements are now different. The
 first one will export the field ``baz``, but not the function ``baz``, while the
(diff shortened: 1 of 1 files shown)


Ericson2314:
@gridaphobe On the countrary I think at least some open endedness is good. Points in the design space include:
 - top level accessesors (today)
 - top level lenses as you say
 - namespaced accessors (`MyRecord.foo`, with some rules on clashing with modules with the same name)
 - namespaced lenses
 - overloaded-label-based accessors 
 - overloaded-label-based lenses

There's a whole lot of bikeshedding for pick that appart. But the first step for all but the status quo is getting these top level accessors out of here!

gridaphobe:
Elaborating the motivation wouldn't pigeonhole us into one particular alternative, it would just demonstrate that there's at least one concrete alternative one might want to the standard record selector functions (and apparently many more in practice).

danidiaz:
How would this proposal interact with the "magic" `HasField` typeclass from [`GHC.Records`](http://hackage.haskell.org/package/base-4.11.1.0/docs/GHC-Records.html)? As the GHC User Guide [states](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html?highlight=hasfield#solving-hasfield-constraints), `HasField` constraints are only solved when the accessor is in scope:

> A field must be in scope for the corresponding HasField constraint to be solved. This retains the existing representation hiding mechanism, whereby a module may choose not to export a field, preventing client modules from accessing or updating it directly.

elaforge:
Another motivation is to be able to write `data A = A { name :: X } | B` and not have to create a partial function.  The name is useful for documentation, and still safe if you use it for construction or pattern matching.  The usual workaround is to create another type like `A !NameX | B` but sometimes that can lead to a lot of clutter.

ChShersh:
> Another motivation is to be able to write data A = A { name :: X } | B and not have to create a partial function.

This won't save you completely. Because you still can use this name in record-update syntax.

```haskell
λ: data A = A { name :: Int } | B deriving Show
λ: foo = B
λ: foo { name = 3 }
*** Exception: <interactive>:4:1-16: Non-exhaustive patterns in record update
```

But what can save you, is adding warnings like `-Wpartial-fields` or `-Wincomplete-record-updates`.

simonpj:
I like this proposal.   I think of a field name as local to the data type or even the data constructor, and it's annoying to have a short, useful name removed from my top-level namespace.

Moreover I can testify that a quite ridiculous amount of implementation complexity arises from having top-level selectors available simultaneously with `DuplicateRecordFields`.   I grant that this is an implementation consideration, but implementation considerations are significant in a compiler so overloaded (by choice) with features.

So I suggest that

* `NoRecordSelectors` is implied by `DuplicateRecordFields`; and 
* the combination of `DulicateRecorFields` with `NoNoRecordSelectors` is illegal.  I really don't want to have to support `NoRecordSelectors` as *well* as the current mess!

PS: because we use the `No` prefix to switch things off, perhaps we'd better specify that the negation of `NoRecordSelectors` is `RecordSelectors`!

adamgundry:
I'm sympathetic to this proposal as well, though I think there will be some questions of detail to work out regarding precisely how import/export works. The fundamental issue is that we need to distinguish between a field label being in scope, and the corresponding selector function being in scope, but traditional Haskell conflates the two. `DuplicateRecordFields` makes this distinction to some extent already, and perhaps we can come up with a more principled story.

@simonpj I don't immediately see how having `DuplicateRecordFields` imply `NoToplevelFieldSelectors` would make the implementation significantly simpler. Perhaps you see something I don't, though?

We would still need to do most of the name resolution/import/export stuff just for field labels, and even if we don't make selectors available in the top-level namespace, we'll still need to do comparable scope-checking of field names for record construction, pattern matching and update (albeit with a data constructor name available in the former two cases). I suppose the problem of type-checking ambiguous selectors goes away, but that's actually relatively straightforward (especially compared to ambiguous updates). 

There's also the issue that enabling `NoToplevelFieldSelectors` will potentially break code already using `DuplicateRecordFields`, and make it more difficult to enable `DuplicateRecordFields` for existing codebases. We tried fairly hard to make sure that enabling `DuplicateRecordFields` was backwards-compatible, and it would be a shame to lose that property unless we really have to.

simonpj:
> I don't immediately see how having DuplicateRecordFields

I was thinking of all the stuff around having a different selector name to the actual selector function.  Maybe I'm mistaken.

reactormonk:
@danidiaz so you would like some way how not to export record field symbols, but `HasField` should still work?

reactormonk:
So there's a few different things an export of a record field is exposing currently:

- the selector function (which this proposal supports to remove)
- the `HasField` field name
- the record update syntax field name
- record creation syntax field name (same as above?)

Did I forgot any? Then I'll add this to the list of `Unresolved Questions` if there shouldn't be a more specific control on which parts to actually export and which ones not.

adamgundry:
@reactormonk you're missing pattern matching (though perhaps that is the same as update/construction). We could certainly conceive a way to export fields for some purposes but not others, and used carefully that might preserve more abstraction, as @danidiaz commented. I worry that it is a complexity rabbit-hole, though.

ocharles:
I wonder if such an extension would help me create "read only" data types. For example,

```haskell
module Example ( T(len), mkT ) where

data T = T { str :: String, len :: Int }

mkT :: String -> T
mkT s = T s ( length s )
```

Gives my users both the ability to view `len` on `T` but also to update `len` on `T`, perhaps breaking invariants. With this extension, I can at least do:

```haskell
module Example ( T(len), mkT ) where

data T = T { str :: String, len :: Int }

mkT :: String -> T
mkT s = T s ( length s )

len :: T -> Int
len T {len} = len
```

and avoid having to invent new names, or give up record syntax for `T` (which as mentioned is usually to clarify what the slots in a record mean).

reactormonk:
@adamgundry So you would suggest this proposal only targets the selector function and ignores the rest for now?

Ericson2314:
I think for privacy, we need to go full `Type (Variant (fields...), ...)`. I'd also like to see something like Rust's record update syntax where the constructor name is there to disambiguate the field names prior to type checking. Interleaving name resolution and type checking is a nasty thing I rather avoid.

reactormonk:
@Ericson2314 I'd prefer to keep the scope of this proposal as small as possible. If you would like to add additional control over what's getting exported, feel free to write an additional proposal.

AntC2:
More motivations [from 7 years ago](https://ghc.haskell.org/trac/ghc/wiki/Records/DeclaredOverloadedRecordFields/NoMonoRecordFields)

Ericson2314:
@reactormonk That's fair. Looking at the docs for `DuplicateRecordFields` I see the existing syntax's semantics is established for just controlling record names. which is something I didn't realize.

I do think we should definitely resolve the unresolved question in saying that `Foo (..)` and friends do *not* cause any top-level definitions to be exported. I think that is forwards-compatible with more possibilities.

reactormonk:
@AntC2 I see that the names still have to be disambiguated if you use them twice - so update / constructor syntax is still bound to the namespacing. I would propose that this proposal does pretty much what it say - it just removes the toplevel function, but still requires manual addition of `DuplicateRecordFields` if you want to have the same field in different records.

AntC2:
@reactormonk, 7 years ago there was no `DuplicateRecordFields`. So my proposal then didn't need to mention anything about disambiguating.

adamgundry:
It would be nice to have per-datatype (or even per-field?) control in the case where one generally wants to use traditional record selectors, provided they are total, but wants to suppress generation of partial selectors.

I think upper-case field names would be unnecessarily confusing (see also [Trac 11671](https://ghc.haskell.org/trac/ghc/ticket/11671)).

reactormonk:
I don't see any unresolved questions left, am I missing any? Except on how I should start implementing this.

proposals/0000-no-toplevel-field-selectors.rst
```diff
@@ -0,0 +1,134 @@
...
+
+Record construction/update syntax and pattern matching will work as before, the
+disambiguation handled by ``DuplicateRecordFields``. The necessary selectors
+will be part of a ``Record(..)`` export, or can also be named one by one. They
+always have to be associated with the data type though, because there is no more
+toplevel selector.
+
+A function for Template Haskell to go from a record field name to a selector for
+that field will be provided, because it's not possible anymore to go from field
+name directly to selector.
```
adamgundry: Can you expand on the interaction with Template Haskell? What, concretely, is the new function you propose?

proposals/0000-no-toplevel-field-selectors.rst
```diff
@@ -0,0 +1,134 @@
...
+Unresolved questions
+--------------------
+
+
+Implementation Plan
+-------------------
+
+For deriving the Generic instances, either ``Record.fieldName``,
+``OverloadedRecordFields`` or an anonymous function which generates the
+corresponding selector in core will be used. (to be clarified)
```
adamgundry: I don't understand this sentence. Despite many related happenings, `OverloadedRecordFields` itself is not an extension in GHC HEAD, and I have no idea what `Record.fieldName` refers to.

More generally, I'm happy to advise on implementing this, assuming it gets accepted. My suggestion would be to make this work along the same lines as `DuplicateRecordFields`. This still generates selector functions behind the scenes, but gives them mangled names (to be unique to the module); it then converts between user-visible and mangled names as needed. You'd need to add a flag to fields indicating that they had been defined without the top-level field selectors, then the name resolution code would look at this flag and ignore such fields when resolving names of variables in expressions (but not in other contexts, e.g. record construction). Since the selector function still exists, it can be used by the existing `HasField` stuff, and its name can be used to uniquely identify the field where necessary.

proposals/0000-no-toplevel-field-selectors.rst
```diff
@@ -0,0 +1,134 @@
...
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/160>`_.
+.. sectnum::
+.. contents::
+
+Enabling this Language Extension removes the default of Haskell data
+declarations to generate toplevel field accessor functions for records, such
+that the user can define their own, either via OverloadedRecordFields, or via
+Generic.
```
adamgundry: What do you mean by "via OverloadedRecordFields"?

proposals/0000-no-toplevel-field-selectors.rst
```diff
@@ -0,0 +1,134 @@
...
+------------
+
+There have been proposals to extend the usage of records in Haskell
+(DuplicateRecordFields, OverloadedRecordFields, DeriveGeneric to name a few),
+but none of them address the original issue where record fields steal the
+toplevel function name for each field for selector purposes. This proposal
+intends to open the design space to this issue by removing the generation of
+these toplevel selector functions.
+
+Possible Alternative Designs
```
adamgundry: I think this means "possible alternative uses for field names, as motivation for removing selectors from the namespace", but it isn't completely obvious. The proposal could do with a few words of explanation here.

proposals/0000-no-toplevel-field-selectors.rst
```diff
@@ -0,0 +1,134 @@
...
+Proposed Change Specification
+-----------------------------
+
+Record definition no longer defines toplevel selector functions when the
+``NoToplevelFieldSelectors`` extension is enabled. A
+``NoToplevelFieldSelectors`` pragma for records will also be available, such
+that this extension can be enabled on a per-record basis.
+
+Record construction/update syntax and pattern matching will work as before, the
+disambiguation handled by ``DuplicateRecordFields``. The necessary selectors
```
adamgundry: s/`DuplicateRecordFields`/`DisambiguateRecordFields`?

proposals/0000-no-toplevel-field-selectors.rst
```diff
@@ -0,0 +1,134 @@
...
+
+- the type ``Foo``
+- the constructor ``Foo``
+- the two functions ``bar`` and ``baz``
+- the names ``bar`` and ``baz`` for record construction (``Foo { bar = 3, baz = "foo" }``)
+- the names ``bar`` and ``baz`` for ``RecordWildCards``
+
+If the language extension ``NoToplevelFieldSelectors`` is enabled for the module
+or ``Foo`` specifically, all of the above will be generated, except for the two
+functions ``bar`` and ``baz``.
```
adamgundry: What exactly is the per-declaration pragma syntax? Perhaps an example would help.

Do we want a per-field pragma? (That would be nice to suppress partial selectors but keep the rest. But then you might have the same name appearing in multiple places...)

proposals/0000-no-toplevel-field-selectors.rst
```diff
@@ -0,0 +1,134 @@
...
+Exporting the names for record construction now has to be specific to the
+record. Without ambiguitiy, previously this was equivalent
+
+    module A where (Foo(Foo, bar, baz))
+
+    module A where (Foo(Foo, bar), baz)
+
+Because of the new semantics, these two export statements are now different.
+Maybe a wildcard export will export functions of the same name as well, this
+question isn't resolved yet.
```
adamgundry: Looks like the proposal text needs to be updated here, as I think this question was resolved in the negative?

A consequence of this is that enabling the extension and writing selector functions out by hand is not equivalent to having the extension disabled. Perhaps that's worth observing?

proposals/0000-no-toplevel-field-selectors.rst
```diff
@@ -0,0 +1,134 @@
...
+
+Effect and Interactions
+-----------------------
+
+Breakage estimation
+^^^^^^^^^^^^^^^^^^^
+
+Enabling this extension will break a lot of Template Haskell. Going from record
+field name to selector won't work anymore. A new way to go from record field
+name to selector has to be found.
```
adamgundry: Can you give some examples of what you expect to break?

proposals/0000-no-toplevel-field-selectors.rst
```diff
@@ -0,0 +1,134 @@
...
+This might cause some confusion that record fields can't be accessed by toplevel
+selectors anymore - however, that shouldn't be too big of an issue, because some
+library authors already stopped exporting these selectors so they don't have to
+break downstream software on record changes.
+
+
+Alternatives
+------------
+
+None.
```
adamgundry: As specified, this extension applies at the definition sites of fields. One could imagine making it apply at use sites instead, so that a module could take selector names out of scope (as variables in expressions), even if they come from imported modules that don't use the extension. Is that worth considering?

proposals/0000-no-toplevel-field-selectors.rst
reactormonk: I added some words, hopefully it's enough.

proposals/0000-no-toplevel-field-selectors.rst
reactormonk: Added some examples. I don't think per-field pragmas are a good idea, a bit too finicky. If you want only to expose some, write the other functions by hand or with some TH.

proposals/0000-no-toplevel-field-selectors.rst
reactormonk: Noted and updated.

proposals/0000-no-toplevel-field-selectors.rst
reactormonk: Anything relying on using `Name` to resolve a datatype accessor. https://github.com/ghc-proposals/ghc-proposals/pull/160#discussion_r215764916 - e.g. I'd guess the codecs Aeson generates. I I'm trying to understand on how it works, but currently failing. Not 100% sure if it actually poses a problem.

proposals/0000-no-toplevel-field-selectors.rst
reactormonk: Then you'd need an orphan `Generic` instance, or to enable one of the language extensions to use them. Might be an interesting addition to use old datatypes with this approach. When creating this library, I was mostly thinking of where you have control over the declaration sites, not where you don't. I'd still go with this proposal, and consider this to be an additional feature.

proposals/0000-no-toplevel-field-selectors.rst
reactormonk: I had the idea that `OverloadedRecordFields` still generates the function names. Probably a bad idea. Lemme remove that.

proposals/0000-no-toplevel-field-selectors.rst
reactormonk: So you're suggesting to implement it by generating the functions inside `Generic` by pattern matching? (What's the difference to positional-based access btw?)

proposals/0000-no-toplevel-field-selectors.rst
adamgundry: `DuplicateRecordFields` already hits this problem, and the current solution is to build TH `Name`s using the `OccName` of the field label (which might not be module-unique) as a `NameQ`. This means that subsequently looking up the name might fail due to ambiguity, but doesn't seem to cause many problems in practice. Can't do you do the same thing?

Indeed, fixing this "properly" would presumably involve changing the TH AST to use a special type for field labels rather than reusing `Name`. That would require careful thought and cause backwards incompatibility. An alternative might be to add a new `NameFlavour`. But perhaps TH changes are better as a separate proposal?

proposals/0000-no-toplevel-field-selectors.rst
adamgundry: Right, `DisambiguateRecordFields` is the extension that permits construction and pattern matching to determine the type in use from the constructor name, even if the field names are ambiguous (or in this case, not in scope as selectors).

proposals/0000-no-toplevel-field-selectors.rst
```diff
@@ -0,0 +1,168 @@
...
+always have to be associated with the data type though, because there is no more
+toplevel selector.
+
+A function for Template Haskell to go from a record field name to a selector for
+that field will be provided, because it's not possible anymore to go from field
+name directly to selector.
+
+A new `TH` function is added which takes two `Name` and returns an `Exp` which
+is equivalent to `a -> b`, where `a` is the record type specified by the second
+argument, and `b` the contents of the field specified by the first argument.
```
adamgundry: Can't one simply use `getField` for this, without needing TH? Or it should be possible to write this function using the existing TH API, I think.

proposals/0000-no-toplevel-field-selectors.rst
reactormonk: Ah, so defer the resolution to the typechecking phase. That makes the implementation of this proposal a lot easier.

proposals/0000-no-toplevel-field-selectors.rst
reactormonk: So mandate the use of `getField` to defer the resolution to the typechecking phase, as mentioned in the other comment? Works for me.

proposals/0000-no-toplevel-field-selectors.rst
```diff
@@ -0,0 +1,168 @@
...
+::
+
+    {-# LANGUAGE NoToplevelFieldSelectors #-}
+
+    data Foo = Foo { bar :: Int, baz :: String }
+
+::
+
+    {-# NoToplevelFieldSelectors #-}
+    data Foo = Foo { bar :: Int, baz :: String }
```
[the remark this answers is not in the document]
reactormonk: You suggest `{-# NoToplevelFieldSelectors Foo #-}` instead?

proposals/0000-no-toplevel-field-selectors.rst
```diff
@@ -0,0 +1,168 @@
...
+
+::
+
+    {-# LANGUAGE NoToplevelFieldSelectors #-}
+
+    data Foo = Foo { bar :: Int, baz :: String }
+
+::
+
+    {-# NoToplevelFieldSelectors #-}
```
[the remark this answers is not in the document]
reactormonk: Good point. Would probably also simplify the implementation.

proposals/0000-no-toplevel-field-selectors.rst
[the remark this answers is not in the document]
ndmitchell: That would be consistent with INLINE. Otherwise to be consistent with UNPACK You'd put it after the `data` (or after the `=` or somewhere generally around there), and not have a named variant. It's really the existence of the named and unnamed variant that's a bit weird. (That said, as below, I'd ditch both.)

reactormonk:
Regarding imports, if a record is defined with `ToplevelFieldSelectors` and being imported, should adding the language extension remove these toplevel functions from the namespace? I'd say no.

int-index:
>  I'd say no.

I agree. I would expect `-XTopLevelFieldSelectors` to determine whether the selectors are generated for definitions in the current module, not whether they are imported from other modules.

reactormonk:
Funkier version:

```
{-# LANGUAGE DuplicateRecordFields -#}
module A where

data Foo = Foo { foo :: String }
```

```
{-# LANGUAGE DuplicateRecordFields -#}
module B where
data Foo = Foo { foo :: String }
```

```
{-# LANGUAGE DuplicateRecordFields -#}
{-# LANGUAGE NoToplevelFieldSelectors -#}
module C where
import qualified A
import qualified B

data Foo = Foo { foo :: Char }

bar = foo $ Foo 's' -- should this work? Probably not.
```

nomeata:
@reactormonk in light of the comments here, do you want to further pursue this proposal?

reactormonk:
Yes, I'm currently implementing it.

ndmitchell:
@reactormonk note that the proposal can be approved separately to the implementation, and is often done so. If you think the proposal is in a final state, or that the tweaks will be minor and implementation driven, it can be submitted to the committee.

reactormonk:
Yes, the proposal is stabilized and is ready to submitted.

proposals/0000-no-toplevel-field-selectors.rst
```diff
@@ -0,0 +1,154 @@
...
+
+Breakage estimation
+^^^^^^^^^^^^^^^^^^^
+
+Enabling this extension will break a lot of Template Haskell. Going from record
+field name to selector won't work anymore. A new way to go from record field
+name to selector has to be found.
+
+Anything that generates code with the help of Generic should be fine. The same
+functionality that generates the anonymous functions for Generic could be used
```
gridaphobe: I think you mean "the same functionality that generates the anonymous functions for **TH** could be used to provide **Generic** functionality". This is the first mention of Generic in the proposal :)

gridaphobe:

**Summary**
This proposal adds a new extension `NoToplevelFieldSelectors`, which stops GHC from generating record selector functions in the current module. Why would one want this? The selector functions GHC generates are simple and special-cased to the given record type. This can quickly lead to name clashes between records that share fields. So instead one might want to ignore the standard record selector and use a lens or `OverloadedLabels`-based accessor instead. Doing this currently requires renaming the field itself in such a way that the generated does not clash with the user-defined accessor (e.g. prefixing with a `_`), but that then creates an ugly discrepancy between the user-defined accessor and the field, which may still be used in record construction or pattern-matching. Solution: don't generate the standard accessor in the first place.

**Recommendation**
I recommend accepting this proposal. It's simple and addresses a clear wart in the experience of modern Haskell.

@ghc-proposals/ghc-steering-committee please discuss. As usual, silence will be taken as tacit agreement.

proposals/0000-no-toplevel-field-selectors.rst
reactormonk: Isn't it the same in the end? There's a function somewhere which allows you to create a record selector, and it'll be used by Generic and TH. Maybe rewrite the paragraph altogether?

gridaphobe: Sure, it's a very minor comment.

nomeata:
I concur!

goldfirere:
I worry about our habit of using language extensions to make module-wide decisions that could more sensibly be made locally. We have fixed this problem before (`-XOverlappingInstances` becoming `{-# OVERLAPPING #-}`) and still suffer from other places where we haven't (`-XRebindableSyntax`, I'm looking at you). Instead of contributing to this habit, maybe we can do this per-declaration? Or even per-field (that might be noisy):

```
data {-# NO_FIELD_SELECTORS #-} Foo = MkFoo { ... }
```

It's a bit ugly, I grant you, but it allows finer control. We also could easily use a new pseudo-keyword after `data`:

```
data noselectors Foo = MkFoo { ... }
```

Another minus with making this a language extension is that the place where the decision is made to have selectors or not is very far from the declaration -- perhaps not even in the file at all.

gridaphobe:
It appears this proposal originally included a per-declaration pragma, which was removed after [discussion](https://github.com/ghc-proposals/ghc-proposals/pull/160#discussion_r238104902).

I'd be happy to have a per-declaration pragma as well, though I don't think we should make per-declaration the only option. The comparison to `OverlappingInstances` and `RebindableSyntax` doesn't quite hit the mark IMO. As I understand it, `OVERLAPS` is preferred because `OverlappingInstances` could result in some instances being accepted that you really didn't want. In contrast, the likely outcome of a module-wide `NoFieldSelectors` is that some field selectors you tried to use do not exist, and your module fails to compile. That's much easier to diagnose than unexpected runtime behavior. And `RebindableSyntax` is about controlling use sites, while `NoFieldSelectors` is about controlling declaration sites.

Though I don't really use lenses or overloaded-labels myself, I can easily see users wanting to use them pervasively. In that case, requiring a `NO_FIELD_SELECTORS` pragma on each declaration just creates a bunch of noise and tedium.

goldfirere:
Fair point about the noise, etc., of a local switch. I still don't love the idea of putting this in a language extension though.

Is it worth thinking about having the language extension control *usage* sites? So `-XNoToplevelFieldSelectors` would just remove all selectors from scope in the module, but it doesn't affect any declarations. That localizes the effect of the extension, and it somehow feels better to me. I guess the problem there is that (presumably) the declaring module wants to export symbols that would clash with the selectors, and so all downstream modules would have to disable field selectors to make any sense at all. Urgh.

I think I'll have a hard time getting any more positive than "I won't vote against" on this one, in the end.

As I'm continue to find a way to like this more, I've stumbled upon this: What if (in some hypothetical other language) record selectors were not available by default, and instead you needed, e.g.,

```
data Foo = MkFoo { ... }
  deriving selectors
```

That seems to be much more in the spirit of Haskell (giving you *less* rope by default) than what we have currently. Of course, we can't just *do* that. But maybe thinking about that as an aspirational goal for the language will help me to support this proposal...

gridaphobe:
```
data Foo = MkFoo { ... }
  deriving selectors
```
I quite like this.

We can't just flip a switch and make all Haskell code like this, but perhaps a future language standard could default to something like this. And in that case, this proposal could be a stepping stone. Suppose that, rather than disabling selector-generation throughout the module:

1. `NoToplevelFieldSelectors` changes the **default** behavior to not generate selectors, but
2. standard field accessors can be generated per-declaration with a `deriving selectors` clause.

If this extension proves popular, a future standard could specify that selectors are no longer generated, and GHC could just enable the extension by default.

AntC2:
data Foo = MkFoo { ... }
      deriving lenses

Ericson2314:
Rust has `#[...]` pragmas on bits of syntax, and `#![...]` module-wide pragmas. Frequently the latter means "mapping" the pragma on every definition in the module. I think something like that would resolve @goldfirere's and @ndmitchell's concerns.

nomeata:
I like @gridaphobe’s idea with `deriving selectors`. And note that it is a conservative extension of `NoToplevelFieldSelectors`, so we could accept the proposal as is (which is  already useful to many), knowing that we can draft and refine a separate one for `deriving selectors`.

simonpj:
* I support the overall idea of being able to switch off the auto-generation of top level record selectors

* I'm agnostic about the stuff about having per-declaration control or `deriving selectors`.  Seems reasonable to me, but I don't have a strong opinion.

  I note that if we had `deriving selectors` then we would not need a per-declaration `norecordselectors` keyword -- a global flag would suffice.

* Before approving the proposal, though, I'd like it to be more precise.  Specifically

  * "A function for Template Haskell to go from a record field name to a selector for that field will be provided".  What is the name of the new function and what its its type?   (What was the old way to "go from a field name to a selector"?)   Does this function somehow disappear without `-XNoTopLevelFieldSelectors`.

  * "A new TH function is added which takes two Name and returns an Exp which is equivalent to a -> b, where a is the record type specified by the second argument, and b the contents of the field specified by the first argument."  What is the name of this function and what is its type? Why is it needed?   What happends if this type is ill-defined?e.g.
  ```
  data T = forall a. MkD { fld :: a, ...more stuff...}
  ```
  
  * "The necessary selectors will be part of a Record(..) export, or can also be named one by one".  I don't understand.  Give an example?

  * "The first one will export the field baz, but not the function baz, while the second one will export the function baz, but not the field baz."  Is this statement supposed to apply with or without `-XNoTopLevelFieldSelectors`?   I assume with.  Assuming so, I don't undersatnd "the second one.." because the function `baz` doesn't exist since there is no selector.  Perhaps a more complete example would help.

  * What is the negation of `-XNoTopLevelFieldSelectors`? Presumably `-XTopLevelFieldSelectors`. Please say.

nomeata:
> What is the negation of -XNoTopLevelFieldSelectors`? Presumably -XTopLevelFieldSelectors. Please say.

I would actually assume that this introduces a language extension, `TopLevelFieldSelectors`, that is on by default (i.e. implied by `Haskell98`), but can now be turned off with `NoTopLevelFieldSelectors`. Like `TraditionalRecordSyntax` and others. But yes, this should be written as such in the proposal.

BTW, why `TopLevelFieldSelectors` and not just `FieldSelectors`? Do we have non-top-level field selectors?

gridaphobe:
> why `TopLevelFieldSelectors` and not just `FieldSelectors`? Do we have non-top-level field selectors?

Agreed. `[No]FieldSelectors` seems equally precise and easier to type.

gridaphobe:
@reactormonk, @simonpj (https://github.com/ghc-proposals/ghc-proposals/pull/160#issuecomment-498671126) and @nomeata (https://github.com/ghc-proposals/ghc-proposals/pull/160#issuecomment-498682554) have made requests/suggestions for clarifications to the proposal. I would add that the proposal should mention the `deriving selectors` idea, if not as part of the actual proposal, then at least as a natural future extension. I think that would go a long way towards resolving the concern about module-wide vs per-declaration pragmas.

reactormonk:
Will do, currently a bit occupied with LambdaConf.

yav:
Seems reasonable to me

proposals/0000-no-toplevel-field-selectors.rst
```diff
@@ -38,28 +38,29 @@ bindings for other values.
...
 - overloaded-labels based accessors / lenses
 - row types
-- https://github.com/ghc-proposals/ghc-proposals/pull/158
+- `<https://github.com/ghc-proposals/ghc-proposals/pull/158>`_
 - ...
 
 Proposed Change Specification
 -----------------------------
 
 Record definition no longer defines toplevel selector functions when the
```
gridaphobe: I'd rephrase this as something like

> This proposal introduces a new extension `FieldSelectors` that controls the generation of record field selector functions. The extension is enabled by default to match the current behavior, but may be disabled per-module with the `NoFieldSelectors` language pragma.

proposals/0000-no-toplevel-field-selectors.rst
```diff
@@ -38,28 +38,29 @@ bindings for other values.
...
 A function for Template Haskell to go from a record field name to a selector for
 that field will be provided, because it's not possible anymore to go from field
 name directly to selector.
 
-A new `TH` function is added which takes two `Name` and returns an `Exp` which
-is equivalent to `a -> b`, where `a` is the record type specified by the second
-argument, and `b` the contents of the field specified by the first argument.
+A new ``TH`` function is added which takes a ``Name`` (for the constructor) and
+a ``String`` (for the field) and returns a ``Maybe Name``, just like
+``lookupValueName``.
```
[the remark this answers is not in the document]
gridaphobe: Oh, I see this is already in S2.2. Then I think a forward reference at the end of the last paragraph could replace this one entirely.

proposals/0000-no-toplevel-field-selectors.rst
```diff
@@ -85,22 +86,69 @@ these functions are otherwise defined, the wildcard will not export them.
...
+    module B where (Foo(Foo, bar), baz)
+    data Foo = Foo { bar :: Int, baz :: Int }
 
 Because of the new semantics, these two export statements are now different. The
 first one will export the field ``baz``, but not the function ``baz``, while the
 second one will export the function ``baz``, but not the field ``baz``. Because
 of this change, writing out all selector functions by hand is still different,
 because they all have to be exported manually.
 
+.. code-block:: haskell
```
gridaphobe: These code samples don't appear to be rendering at all in the rich-text view.

proposals/0000-no-toplevel-field-selectors.rst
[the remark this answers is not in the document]
gridaphobe: Taking another look at this, it's not clear to me why you changed the new TH function to return a `Name` instead of an `Exp`. How would this work? The field selector does not necessarily exist anymore so you cannot refer to it by name. You could say the semantics are that it generates a local or top-level definition and returns that `Name`, but that seems a lot more cumbersome than just returning a selector as an `Exp` directly.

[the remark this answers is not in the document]
reactormonk: Yeah, `Exp` makes more sense here.

gridaphobe:
Ping @reactormonk, it would be great to address the remaining items so we can merge this proposal.

1. `lookupFieldName` should return a selector lambda directly as an `Exp`. It should also be renamed to match the new type, perhaps `mkFieldSelector :: Name -> String -> Q Exp`. I think the semantics should be that it `fail`s in the `Q` monad if the selector cannot be created for any reason (e.g. the field does not exist, the selector would be ill-typed due to an existentially-quantified tyvar, etc.).

2. We still need to clarify the export list example for @simonpj (https://github.com/ghc-proposals/ghc-proposals/pull/160#issuecomment-498671126). I think if you just reproduce the two modules `A` and `B` with `NoFieldSelectors` enabled and the field selectors defined manually, and explain how the export lists would function in that case, that would address Simon's request.

3. Your last update seems to have broken the rendering of most of the code snippets.

reactormonk:
I forgot to push some changes from earlier :-( I hope the example explained enough.

gridaphobe:
Thanks, the extended examples do help clarify the behavior. In particular, in

```haskell
{-# LANGUAGE NoFieldSelectors #-}
module B (Foo(Foo, bar)) where
data Foo = Foo { bar :: Int }
bar = 42
```

if you want to export both the field `bar` and the top-level definition `bar`, you would need to write

```haskell
module B (Foo(Foo, bar), bar) where
...
```

right? We should add that as a final example, and also specify that such an export list will not trigger `-Wduplicate-exports` as the two `bar`s refer to different things.

proposals/0000-no-toplevel-field-selectors.rst
```diff
@@ -0,0 +1,250 @@
...
+constructor ``Foo``, but instead exports the toplevel bind ``baz``.
+
+.. code-block:: haskell
+
+    {-# LANGUAGE NoFieldSelectors #-}
+    module B where (Foo(Foo, bar), baz)
+    data Foo = Foo { bar :: Int, baz :: Int }
+    baz = 42
+
+The updaters can still be used when exported (as in module ``A``).
```
gridaphobe: I think you can just say the "field" can still be used. I've always understood "field" to refer to the occurrence of a field in record construction/update/pattern syntax, and "selector" to refer to the top-level projection function.

reactormonk: I call them `RecordUpdater` in my GHC branch, hence the name.

gridaphobe:
@reactormonk I just sent you a PR with a few more tweaks to the language. Please make sure I didn't misunderstand anything.

Once the PR is merged, I'll ping the rest of the committee again for another quick look, in particular to make sure their questions were addressed.

gridaphobe:
@ghc-proposals/ghc-steering-committee this proposal has undergone a couple changes since I first recommended acceptance:

1. The extension is now called `[No]FieldSelectors` rather than `[No]TopLevelFieldSelectors`.
2. The proposal specifies the name and type of the proposed TH function to produce an ad-hoc field selector (`mkFieldSelector :: Name -> String -> Q Exp`).
3. The examples describing the behavior of import/export lists under `NoFieldSelectors` have been elaborated.

Please take another look at the proposal (in particular @simonpj and @nomeata, since both of you made suggestions in the first round). As usual, silence will be interpreted as tacit agreement with my recommendation to accept the proposal.

goldfirere:
After "Because of this change, writing out all selector functions by hand is still different, because they all have to be exported separately." there is an example which doesn't seem to make sense there. A relic of edits, perhaps? Actually, I find it quite hard to understand the series of examples in this stretch. Maybe name each module independently, and then refer to module names in the text? Also, it would be great to move these examples out of the "Specification" section and into a new "Examples" section. The specification should be an explanation of the change, stated as precisely as possible -- but without other stuff in between. As it stands, by putting all the examples in the "Specification" section, I worry that the examples contain a new nugget that the explanation above does not.

I don't find the TH change well motivated by this proposal. It states that "a lot" of breakage will occur. Like what? Could you give an example? Also, I assume your new TH function will just produce a little anonymous function with a `case`, right? If so, should it really be in `template-haskell`? I don't think there are too many other utilities of this sort packaged with `template-haskell`.

I apologize for asking for yet more revisions, but these proposals will live a long time, and investment in making them as clear as possible will pay itself back. Thanks!

gridaphobe:
> I don't find the TH change well motivated by this proposal. It states that "a lot" of breakage will occur. Like what? Could you give an example?

Currently, if someone wants to generate code in TH to access a field, they can use the field selector as a function. Any TH code that does this would suddenly break under `NoFieldSelectors`. So TH authors should start generating little lambda functions to extract record fields so that their generated code will compile under `NoFieldSelectors` as well. I'm not a heavy user of TH so I can't say how common this scenario is, but it sure sounds like a common enough thing to do that it could break a lot of TH code.

> Also, I assume your new TH function will just produce a little anonymous function with a case, right? If so, should it really be in template-haskell? I don't think there are too many other utilities of this sort packaged with template-haskell.

I think TH should provide the function, to make it easy for authors to do the right thing. If there are other such utility functions that would similarly make it easier for authors to generate code that will work under various sets of extensions, perhaps we should add them as well!

reactormonk:
I'm not sure how writing TH would look like with this extension in the field - TH functions should work whether this extension is enabled or not. Maybe a TH author could pitch in?

proposals/0000-no-toplevel-field-selectors.rst
```diff
@@ -98,9 +98,30 @@ record. Without ambiguitiy, previously this was equivalent
...
+    baz (Foo _ x) = x
+
+is different from
+
+.. code-block:: haskell
+
+    module Exports where (Foo(Foo, bar, baz))
+    data Foo = Foo { bar :: Int, baz :: Int }
+
+Because the functions in the furst example don't get exported.
```
gridaphobe: small typo: `s/furst/first/`

proposals/0000-no-toplevel-field-selectors.rst
```diff
@@ -128,28 +150,20 @@ constructor ``Foo``, but instead exports the toplevel binder ``baz``.
...
-
-.. code-block:: haskell
-
-    import A
-    foo = Foo 23 42
-    foo { baz = 1 }
-
-This will now fail, because the field ``baz`` is not in scope anymore,
-because it is not exported by ``B``.
+Using ``baz`` when importing ``B`` will fail, because the field ``baz`` is not
```
gridaphobe: "Using `baz` *as a field*" will fail. Using `baz` as a plain old `Int` would succeed.

proposals/0000-no-toplevel-field-selectors.rst
```diff
@@ -234,7 +236,10 @@ None.
 Unresolved questions
 --------------------
 
-- Which order of arguments in the new TH function?
+- What would TH code look like in the future? The disambiguating field name is
+  always the first constructor, maybe introducing a new `Newtype` might make it
```
gridaphobe: I don't understand the second sentence here. What do you mean by "the disambiguating field name"?

goldfirere:
Thanks for the updates.

@RyanGlScott, do you have an opinion about the TH aspects of this proposal?

gridaphobe:
@goldfirere does the latest change address your concern about the clarity of the specification and examples?

Regarding the TH function, it would be nice to hear the rest of the committee's thoughts on providing a function to generate an ad-hoc selector function. To recap, the proposal provides a function

```haskell
mkFieldSelector :: Name -> String -> Q Exp
```

the semantics of which are: given

```haskell
data Foo = Foo { foo :: Int }
```

```haskell
mkFieldSelector 'Foo "foo" == [| \Foo{foo=foo} -> foo |]
```

If a valid selector cannot be generated for any reason (e.g. no such field `foo`, `foo`s type references an existentially quantified tyvar, etc.), `mkFieldSelector` will fail. The recommendation is that TH authors use `mkFieldSelector` anytime they want to generate code that would previously have used `foo` as a function.

@goldfirere has pointed out that it is out of character for TH to provide such a simple utility function. I think the likely consequence of not providing such a function is that TH authors will assume that `foo` can be used as a selector function, and their code will promptly break in modules that have enabled `NoFieldSelectors`. More generally, regardless of the history of TH's API, I think good API design entails making it easy to do the Right Thing, which is the purpose of the `mkFieldSelector` function.

Any thoughts from @ghc-proposals/ghc-steering-committee? I know in particular that @nomeata and @isovector have popular libraries that make use of TH.

goldfirere:
> @goldfirere does the latest change address your concern about the clarity of the specification and examples?

Yes. That's what I meant by "Thanks for the updates". But I should have said I'm quite happy with them, instead of just being generally grateful. Thanks for checking.

And many thanks for seeking followup about the TH question.

gridaphobe:
Sorry, that was actually a race condition on my part :) I didn't see your comment before I hit "submit".

RyanGlScott:
I'm arriving to this discussion extremely late, so I apologize if my thoughts below miss some details that have already been discussed.

> @RyanGlScott, do you have an opinion about the TH aspects of this proposal?

My initial reaction is: `mkFieldSelector` would be certainly convenient, although I'm not sure if we really need to put this in the `template-haskell` library itself. It seems relatively straightforward to implement `mkFieldSelector`—in fact, you could do so in today's `template-haskell`, unless I'm missing something obvious. Moreover, this might not even be the only utility you'd want. Perhaps you'd also want a variant that declares a top-level selector function?

```hs
mkFieldSelectorDec :: Name -> String -> Q [Dec]
```

It seems like there is potential for many different variations on this theme, which leads me to believe that this would be best served by an external library, not `template-haskell` itself.

Aside from that, I'm not sure exactly how this proposal would break existing uses of TH. The proposal states:

> Enabling this extension will break a lot of Template Haskell.

...why exactly? It's not clear to me what particular applications of Template Haskell would be affected by this.

proposals/0000-no-toplevel-field-selectors.rst
reactormonk: Replaced it with "disambiguating record constructor" - clearer?

reactormonk:
> > Enabling this extension will break a lot of Template Haskell.
> 
> ...why exactly? It's not clear to me what particular applications of Template Haskell would be affected by this.

If your TH uses the selector functions to interact with the record, that won't work anymore.

RyanGlScott:
> If your TH uses the selector functions to interact with the record, that won't work anymore.

While this is a true statement, I don't see what is special about TH here. If you have _any_ code that uses the selector functions, then it will break with `NoFieldSelectors`. TH isn't special in this regard, so why bother singling it out in the proposal?

reactormonk:
Because existing TH code cannot be used in modules which enable `NoFieldSelectors` - any other code which you reference will not have these issues. Non-TH code which you import won't expect the presence of the selector functions - e.g. `Generic`s will continue to work just fine.

RyanGlScott:
> Because existing TH code cannot be used in modules which enable `NoFieldSelectors`

I'm still not understanding what you're getting at here. Can you give a concrete example of code that uses Template Haskell which which compiles normally but won't compile with `NoFieldSelectors`? I think you had a particular pattern in mind when you wrote that comment, but it's not clear to me what that is at present.

nomeata:
> I know in particular that @nomeata and @isovector have popular libraries that make use of TH.

Do I? My library at most ab-use TH to create fake syntax, but not to do any kind of serious code generation. I consider myself definitely a non-expert on TH (and hence restrain myself from discussion TH-related issue in detail).

reactormonk:
E.g. generating a field lens as

```
foo = lens _foo (\x f -> x { _foo = x })
```

RyanGlScott:
> E.g. generating a field lens as
> 
> ```
> foo = lens _foo (\x f -> x { _foo = x })
> ```

Ah, OK. I think I'm just splitting hairs over how this proposal is worded, then. There isn't anything inherently special about Template Haskell that lets you write this code—you could just as well write `foo = lens _foo (\x f -> x { _foo = x })` by hand, and that too would break in the presence of `NoFieldSelectors`. I think it would be more direct to say something like "Enabling this extension in a random piece of Haskell code is likely to break it due to how widely used field selectors are. (For example, a common idiom is to define lenses using field selectors.) Such code would need to migrate away from using field selectors directly in order to work with `NoFieldSelectors`."

In any case, I think I should hand the conversation back over to the GHC Steering Committee, since I've already said my piece (and then some).

goldfirere:
Thanks, Ryan, for your contributions here. In light of that conversation, I still say that the TH piece should be removed. TH isn't special here, and I think keeping `template-haskell` free of utilities is good. (Why? Because `template-haskell` cannot be updated independently from GHC. Putting things in libraries buys much more flexibility.)

gridaphobe:
The difference between TH and random other code is that TH generates code *in the context of a module which it does not control*. A TH author does not control which extensions are enabled in the module in which their code will execute. I think this makes TH substantially different, because a TH must (or at least should) make their code work under a variety of environments. An author of regular Haskell code can choose themselves whether or not to enable an extension and code accordingly.

So I think TH is special, and the function should exist. But Richard's most recent comment

> Why? Because template-haskell cannot be updated independently from GHC. Putting things in libraries buys much more flexibility.

seems plausible for keeping the function out of the `template-haskell` package itself. Since it's so tightly coupled to GHC, it would be good to keep it as small as possible and let utilities built on top of `template-haskell` evolve separately. I don't know if such a library exists at the moment, but it would be nice to have. And it also raises the question of whether any current pieces of `template-haskell` should also be extracted, though that's out of scope for the current discussion.

So that all said, I think I would be ok with removing the TH function itself, and replacing it with a note in the "Effects and Interactions" section that TH authors should avoid using fields as selector functions going forward.

reactormonk:
> So that all said, I think I would be ok with removing the TH function itself, and replacing it with a note in the "Effects and Interactions" section that TH authors should avoid using fields as selector functions going forward.

I'll add that idea as soon as I figure out the snippet to select a field given a constructor and a field name.

simonpj:
> Please take another look at the proposal (in particular @simonpj and @nomeata,

I still like the proposal, but here are some suggestions for what would make it stronger and clearer

* "Proposed change specification" should, well, specify the proposed change.
  But currently it does not. It says various things that are _not_ changed,
  but does not say explicitly what _is_ changed.  It'd be good to say
  "This proposals makes three changes:

  1. The extension NoFieldSelectors suppresses the creation of a top-level record selector function.  (Of course that is obvious, but really you should say it!)
  2. The extension NoFieldselectors affects the behaviour of import and export lists in particular ways, that should be specified.  (I can _infer_
     the changes from Section 3, but that doesn't absolve the proposal from specifying the change.)
  3. A new function is added to TH. (Is that function somehow not available if the flag is off?  Or is it available unconditionally? The latter seems more plausible.)

  "No other aspect of records are affected by the flag: in particular,
   record construction, update, pattern matching are unchanged."

* In an earlier comment I suggested that we make `DuplicateRecordFields` mutually exclusive with `FieldSelectors`.   That is, if you want `DuplicateRecordFields`, you also get `NoFieldSelectors`. That would mean that when field selectors are auto-generated (a la Haskell 98) their name can be the field name. If you also have `DuplicateRecordFields` the auto-generated selector needs a funny name, which is painful in the implementation and requires [special explanation and rules](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/glasgow_exts.html#selector-functions) for the user.

* I agree with "The difference between TH and random other code is that TH generates code in the context of a module which it does not control." That is a far clearer statement than what is in the proposal "Enabling this extension will break a lot of Template Haskell".  I'm sure this bit can be improved in the light of the Discussion. The point seems to be: in TH quotations, be careful when using any construct whose semantics might change depending on the flags in scope at the splice site.  And record selectors are one of those things.

* The proposal says "Going from record field name to selector won't work anymore".  What does that mean?  Better to give the code you would have written before, and what you have to write instead now.  I _think_ (but am far from certain), that if there is no top-level selector to call, you can always generate the TH quote `case x of K { f = f } -> f` instead.  Is that it?  So then it's purely a matter of convenience whether the TH library provides such a function.  If so, spell all this out.

gridaphobe:
> In an earlier comment I suggested that we make DuplicateRecordFields mutually exclusive with FieldSelectors. That is, if you want DuplicateRecordFields, you also get NoFieldSelectors. That would mean that when field selectors are auto-generated (a la Haskell 98) their name can be the field name. If you also have DuplicateRecordFields the auto-generated selector needs a funny name, which is painful in the implementation and requires special explanation and rules for the user.

This sounds like a fine idea to me. Not only does it complicate the implementation for GHC, but the generated field selectors (for duplicated fields) are mostly inaccessible to the user in the defining module (and quite likely in downstream modules as well) due to the name collisions.

reactormonk:
> In an earlier comment I suggested that we make DuplicateRecordFields mutually exclusive with FieldSelectors. That is, if you want DuplicateRecordFields, you also get NoFieldSelectors. That would mean that when field selectors are auto-generated (a la Haskell 98) their name can be the field name. If you also have DuplicateRecordFields the auto-generated selector needs a funny name, which is painful in the implementation and requires special explanation and rules for the user

My current implementation (if it works out) doesn't need the mangling anymore (except at the code generation step). I'd rather have `NoFieldSelectors` imply `DuplicateRecordFields` semantics regarding field names and simply phase out `DuplicateRecordFields`. The funky between renamer-and-typechecker step probably won't go away because of the presence of record update syntax which is still supported.

gridaphobe:
> I'd rather have `NoFieldSelectors` imply `DuplicateRecordFields` semantics regarding field names and simply phase out `DuplicateRecordFields`.

I don't think having `NoFieldSelectors` imply `DuplicateRecordFields` is a good idea.

`DuplicateRecordFields` is more complicated than `NoFieldSelectors` in terms of how much the user has to understand about the minutiae of GHC. `NoFieldSelectors` is pretty simple, if you enable it field selectors will no longer be generated. There are some details around export lists that you many have to understand, and TH code must take care to not assume the existence of field selectors, but that's it. `DuplicateRecordFields`, on the other hand, introduces ambiguity into the record syntax and relies on GHC's bidirectional type checker to resolve it when possible. And in the case of record updates, it's often not possible to resolve the ambiguity. This makes me quite wary of recommending `DuplicateRecordFields` to newcomers to the language, which is unfortunate because it tackles a serious wart in Haskell's record system. I don't think I would have similar concerns about `NoFieldSelectors`.

`NoFieldSelectors` is also useful in the absence of `DuplicateRecordFields`, as argued by your proposal! So I would keep the implication in the direction that @simonpj suggested, primarily as a way to simplify the implementation of GHC with minimal if any user-visible effects.

reactormonk:
Yeah, record updates might be a bit wonky with `DuplicateRecordFields`. Let's leave it explicit for now then.

gridaphobe:
Thanks for removing the `NoRecordFields` => `DuplicateRecordFields` implication, but what do you think about the opposite direction that @simonpj suggested? Suppressing field selector generation under `DuplicateRecordFields` would simplify GHC's implementation, and those field selectors are quite unlikely to be useful in the first place.

I think that's the only remaining technical question to resolve. So we ought to be able to wrap this up quickly.

reactormonk:
That sounds like a breaking change to me. I'm not sure we should be doing that. The disambiguation logic will still be required for record updates, so it won't help that much.

gridaphobe:
It is a breaking change, and on second thought it's much more likely to break existing code than I thought at first, because the change would disable generation of **all** record selectors in the module (not just the ambiguous selectors). So let's not do that right now, perhaps later with a proper deprecation cycle if there's desire for it.

I'll give the rest of the committee (in particular @simonpj, who suggested the `DuplicateRecordFields` => `NoFieldSelectors` implication) a few more days to respond to this last recommendation, and then I'll mark this accepted.

/remind me on Friday to accept this proposal

isovector:
I'm in favor of solving the problem motivated by this proposal, though must admit I'm stymied by the examples and interactions between different styles of export lists.

gridaphobe:
@reactormonk thanks for the contribution!

reactormonk:
Implementation branch: https://gitlab.haskell.org/reactormonk/ghc/commits/fieldselectors

Ericson2314:
@reactormonk https://gitlab.haskell.org/ghc/ghc/merge_requests/1651#note_220137 reminds me that there should be something similar for enums. Would you like to coauthor another proposal?

TheMatten:
What's the status of implementation? Can we expect it to land in upcoming versions of GHC?

reactormonk:
I'm currently on my third try to implement it. If you're interested, I could pass on my notes.

TheMatten:
@reactormonk That would be cool - not sure if I would be able to help at the moment, but I would still find them interesting :slightly_smiling_face:

arybczak:
@reactormonk could you put these notes / your current progress (as the branch you linked to above is stale I believe) somewhere public if appropriate (or share them with me)? Perhaps I could help with the implementation.

fumieval:
@reactormonk This is a very cool idea which would benefit a large portion of the ecosystem. What are the remaining tasks in your implementation? If you don't have time, I'm willing to help

fumieval:
I talked with reactormonk and made some progress in https://gitlab.haskell.org/fumieval/ghc/-/commits/fieldselectors-fumieval. 

Record construction (`Foo { a = x, b = y }`) works for now. I'm scratching my head with updating; `disambiguateRecordBinds` needs to find the candidates from the field name, but there is no clue available in `GlobalRdrEnv` when selector functions are not created.

adamgundry:
@fumieval Great to hear you're making progress!

I haven't looked at the details of the branch yet, but I suggest you should continue to generate selector functions (even when `NoFieldSelectors` is on), giving them mangled names (just as `DuplicateRecordFields` does). In any case you will need to continue to generate the functions so that `HasField` constraint solving can work. You can then use the selector names to uniquely identify the field-within-the-type when necessary, and disambiguating record updates should work just as it does with `DuplicateRecordFields`. That is, fields under `NoFieldSelectors` would still appear in the `GlobalRdrEnv`, you would just modify the renamer's name lookup code to ignore them when looking up an occurrence of a field in an expression. See https://github.com/ghc-proposals/ghc-proposals/pull/160#discussion_r215765079

You may also be interested in https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3257 where I'm working on generating updater functions alongside selectors, to support changes to `HasField`. That requires some tweaks to `FieldLabel` as well.

I'd be happy to chat further with you and @reactormonk about the implementation of these various pieces if that would be helpful.

fumieval:
Ah, what I've said in https://github.com/ghc-proposals/ghc-proposals/pull/160#issuecomment-673435757 was inaccurate; selectors are still created with mangled names but `lookupParents` can't find them. I need to investigate how it works in DuplicateRecordFields...

I created a discord channel: https://discord.gg/EZ96GN I'm still a GHC newbie so any kind of suggestions/feedbacks would be quite helpful.

fumieval:
Tests are all green and the merge request is ready for review: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4017

georgefst:
> Tests are all green and the merge request is ready for review: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4017

Great stuff - very excited about this (especially in combination with `RecordDotSyntax`.

As someone who doesn't follow GHC development particularly closely, is this going to be too late for 9.0?

Ericson2314:
Copying a comment i left on the PR that is really about the design rather than the implementation:

> Per https://gitlab.haskell.org/ghc/ghc/-/issues/18650 and https://gitlab.haskell.org/ghc/ghc/-/issues/18650#note_298922 in particular, I think `-XNoFieldSelectors` perhaps should squeltch `-Wpartial-fields` warnings because all partiality can be caught per-use. What do you all think?

adamgundry:
@Ericson2314 To quote [myself](https://gitlab.haskell.org/ghc/ghc/-/issues/18650#note_299208):
> Even with `-XNoFieldSelectors`, if there are partial fields then one can still write a partial record update. So I think `-Wpartial-fields` should remain as-is.

Ericson2314:
@adamgundry oh sorry I would have linked that too had I not missed it / forgottne about it.

Doesn't `-Wincomplete-record-updates` serve that role, though?

adamgundry:
Approximately, though it warns at use sites rather than definition sites, and there is also possible partiality due to `setField`. Basically I don't see any reason for `-Wpartial-fields` and `-XNoFieldSelectors` to interact. It would seem very surprising to me if changing a language extension disable a somewhat-but-not-entirely-related warning.

Ericson2314:
@andrewthad Yeah I'm comfortable with `-Wpartial-fields` existing as is and taking on a lint role, I just thing it would be better if the actual sources of error throwing are picked up a more granularly, as you issue proposes. `-Wpartial-fields` to me is too far removed from the source of actual problems (some beginner says, "I only used this field in one variant, so what!?") to serve as the partial code catcher, but fine as a lint.

goldfirere:
> As someone who doesn't follow GHC development particularly closely, is this going to be too late for 9.0?

Yes. GHC 9.0's feature set has settled, as we're trying to get out release candidates by the end of the month, I think.

Akhristenko:
Does this extension allow you to write code `data MyType = MyInt { field :: Int } | MyBool { field :: Bool }`?

fumieval:
@Akhristenko Good question, currently it rejects homonymous selectors with different types.

Should this restriction be lifted in NoFieldSelectors?

cblp:
By the way, homonymous fields with different types are allowed by DuplicateRecordFields.

adamgundry:
> By the way, homonymous fields with different types are allowed by DuplicateRecordFields.

Not within a single type, they aren't. 

I don't think you can lift this restriction without disallowing record updates, which is not currently part of `NoFieldSelectors`. It also isn't clear how `HasField` constraint solving would work, and hence what `RecordDotSyntax` would mean.

Akhristenko:
> It also isn't clear how HasField constraint solving would work, and hence what RecordDotSyntax would mean.

IMHO: In case of sum types you don't need to generate getField at all. Instead, just use pattern matching. As it work in rust, for example.

getField and RecordDotSyntax still will work in case with one type constructor.

adamgundry:
That's true, I think it would be reasonable to permit such datatype declarations if you dropped record selection/update/`HasField` and supported construction and pattern-matching only. It needs a new ghc-proposal though, and I don't currently see a strong motivation for it when you could just use distinct names.

I wonder how much TH/generics code out there assumes all fields have the same type...

Ericson2314:
@adamgundry I believe everything that doesn't work when the fields don't have types is already warned via your new warnings in https://gitlab.haskell.org/ghc/ghc/-/issues/18650 ? I would support a ghc-proposal for this field name reuse, if for no other reason than that feature would keep GHC honest under the hood.

adamgundry:
> @adamgundry I believe everything that doesn't work when the fields don't have types is already warned via your new warnings in https://gitlab.haskell.org/ghc/ghc/-/issues/18650 ? I would support a ghc-proposal for this field name reuse, if for no other reason than that feature would keep GHC honest under the hood.

Well, I think the idea here is that selectors/updates/`HasField` wouldn't be available at all if you had the same field name with different types in a single datatype, whereas https://gitlab.haskell.org/ghc/ghc/-/issues/18650 proposes warnings at use sites of fields that occur in some constructors but not others.

My proposed workaround would be to encode
```hs
data MyType = MyInt { field :: Int } | MyBool { field :: Bool }
```
as
```hs
data MyType = MyInt { fieldInt :: Int } | MyBool { fieldBool :: Bool }
```
which is somewhat dangerous now, but rather safer if you have the warnings from https://gitlab.haskell.org/ghc/ghc/-/issues/18650 and use `-Werror`.

Ericson2314:
@adamgundry Yeah I mean, your stuff shows we can at least detect the things
2. `Werror=` shows that we can prevent them.
3. Codifying as a language extension makes it fair game to make those `HasField` impossible to emit
4. Now there is room to allow mixed typed fields.

Then I gain, I suppose it's perfectly possible to emit `HasField` for partial fields, but not "heterogeneous fields".

I would not like to rename the fields underneath the hood, as I think it's best of GHC thoroughly understands fields as being per-variant, and recovers the traditional per-data-type "view" of those fields only when it needs to.

Akhristenko:
@Ericson2314, @adamgundry 
Not quite so. Idea here is that selectors/uptaders/HasField don't be generated for sum types at all. And all work with this types doing via pattern matching/constructors(with RecordWildcards it quite convenient). Like it work in Rust/Swift.
This automatically solve https://github.com/ghc-proposals/ghc-proposals/pull/351, as we just can't write code which give us run-time errors(thanks to exhaustive check).
And if we need complex structure in sum type we always can do
```
data ComplexType = ComplexType { ... }
data SumType = ConstructorOne | ConstructorTwo ComplexType
```
and use record dot syntax for complex type after pattern matching.

Ericson2314:
@Akhristenko So this is getting off topic, but I've decided that "named fields for algebraic data types" and "named fields for extensible record types" are just two completely different language features that we've made the mistake of conflating in the decades-old never-quite-ending debate over records. With that in mind, special casing single-variant algebraic data types is just and awkward middle ground.

In a way, Rust (and Agda!) agree in that a single-variant `enum` and `struct` are not syntactically the same. Rust is just are skipping the extensible part as there isn't an obvious single way to manage the memory.

adamgundry:
I've been looking at the implementation of this proposal, and have come across a gap in the specification (as I understand it). Input would be appreciated! See https://gitlab.haskell.org/ghc/ghc/-/issues/18999

PR #102, opened 2018-01-06
Add Dependent Haskell quantifiers (including the new "foreach")

goldfirere:
This proposal reserves syntax for all the quantifiers of Dependent Haskell, along with introducing the `-XDependentTypes` extension and allowing the new quantifiers to be used in very limited contexts. This proposal does *not* actually add dependent types.

[Rendered](https://github.com/goldfirere/ghc-proposals/blob/pi/proposals/0000-pi.rst)

PR #161, opened 2018-08-18
Remove DatatypeContexts

monoidal:
[Rendered](https://github.com/monoidal/ghc-proposals/blob/no-datatype-contexts/proposals/0000-remove-datatypecontexts.rst)

### DIFF
--- a/proposals/0000-remove-datatypecontexts.rst
+++ b/proposals/0000-remove-datatypecontexts.rst
@@ -57,7 +57,9 @@ This will be another deviation from the Haskell 2010 standard. Note that GHC alr
 
 Packages using DatatypeContexts will break. I downloaded the entire Hackage repository, used different regexes to search for usage and manually filtered out false positives. Out of 12907 hackage packages, I counted 181 in total using DatatypeContexts; out of those, 14 are deprecated, 120 didn't have an upload within the last 2 years, there are 47 remaining. `See details <https://gist.github.com/monoidal/77c6dd7490f7c9398db35b35273cc030>`_.
 
-In general, programs with `DatatypeContexts` can be updated just by removing the context. In rare cases, programs depending on type defaulting can `change semantics <https://prime.haskell.org/wiki/NoDatatypeContexts>`_. I don't expect this to occur in practice; worst-case, type defaulting can be detected with ``ghc -Wall``.
+In general, programs with `DatatypeContexts` can be updated just by removing the context, with two exceptions:
+* In rare cases, programs depending on type defaulting can `change semantics <https://prime.haskell.org/wiki/NoDatatypeContexts>`_. I don't expect this to occur in practice; worst-case, type defaulting can be detected with ``ghc -Wall``.
+* Datatype contexts can be used as a workaround to lack of kind signatures in Haskell 2010: ``data Monad m => Foo m a = Blah``. In modern GHC, this code can be written using ``KindSignatures``.
 
 
 Alternatives
--- a/proposals/0000-remove-datatypecontexts.rst
+++ b/proposals/0000-remove-datatypecontexts.rst
@@ -28,7 +28,7 @@ As a result, ``MkT`` has type ``Num a => a -> T a``. Pattern matching on ``MkT``
 
 has type ``Num a => T a -> a``. This dictionary is discarded and not used during construction nor during pattern matching.
 
-(This can be contrasted with GADT pattern matching: in ``data T2 a = Num a => MkT2 a``, pattern matching on ``MkT2`` provides the ``Num a`` constraint, so ``f2 (MkT2 x) = x + x`` has type ``MkT2 a -> a`` even though uses addition in the function body.)
+(This can be contrasted with GADT pattern matching: in ``data T2 a = Num a => MkT2 a``, pattern matching on ``MkT2`` provides the ``Num a`` constraint, so ``f2 (MkT2 x) = x + x`` has type ``T2 a -> a`` even though uses addition in the function body.)
 
 This extension is commonly described as a misfeature - it forces type signatures to have constraints without giving anything in return.
 
--- a/proposals/0000-remove-datatypecontexts.rst
+++ b/proposals/0000-remove-datatypecontexts.rst
@@ -1,0 +1,74 @@
+Remove DatatypeContexts
+=======================
+
+.. proposal-number::
+.. trac-ticket::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/161>`_.
+.. sectnum::
+.. contents::
+
+I propose to remove datatype contexts from GHC.
+
+
+Motivation
+------------
+DatatypeContexts is a Haskell feature that allows to define a constrained datatype:
+
+::
+
+ data Num a => T a = MkT a
+
+As a result, ``MkT`` has type ``Num a => a -> T a``. Pattern matching on ``MkT`` requires the presence of the ``Num a`` dictionary: the function
+
+::
+
+ f (MkT x) = x
+
+has type ``Num a => T a -> a``. This dictionary is discarded and not used during construction nor during pattern matching.
+
+(This can be contrasted with GADT pattern matching: in ``data T2 a = Num a => MkT2 a``, pattern matching on ``MkT2`` provides the ``Num a`` constraint, so ``f2 (MkT2 x) = x + x`` has type ``MkT2 a -> a`` even though uses addition in the function body.)
+
+This extension is commonly described as a misfeature - it forces type signatures to have constraints without giving anything in return.
+
+DatatypeContexts is a part of Haskell 98 and 2010 standards. In 2011, the Haskell language committee `accepted <https://mail.haskell.org/pipermail/haskell-prime/2011-January/003335.html>`_ its removal as a new baseline for the next language standard. Starting from GHC 7.2 (released in 2011), DatatypeContexts is disabled by default and gives a deprecation warning when enabled.
+
+DatatypeContexts costs to maintain. It can be combined with newtypes, GADTs and data families. It interacts with roles, `record polymorphism <https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html?highlight=DatatypeContexts#solving-hasfield-constraints>`_, `deriving Functor <https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html?highlight=DatatypeContexts#deriving-functor-instances>`_, `partial type signatures <https://github.com/ghc/ghc/commit/d831b6f41b3b89dc4a643069d5668c05a20f3c37#diff-301d72a6bf8b97cff7a0d225b9876cea>`_ and others. It's known as "stupid theta" in GHC source: `git grep stupid` shows 80+ places where it occurs.
+
+Proposed Change Specification
+-----------------------------
+I propose that in GHC 8.8:
+
+- The flag ``-XDatatypeContexts`` continues to exist, but it has no effect other than printing a deprecation message. This behavior is the same as the current flag ``-XGenerics``.
+- Attempting to write a datatype context gives an error message stating that this feature was removed.
+- Currently, the TH constructors ``DataD``, ``NewtypeD``, ``DataInstD``, ``NewtypeInstD`` allow to declare a context. For backwards compatibility, this is not changed. Instead, attempting to splice with a nonempty context gives an error.
+- The `known bugs and infelicities <http://downloads.haskell.org/~ghc/latest/docs/html/users_guide/bugs.html>`_ list states that DatatypeContexts is not supported.
+
+
+Effect and Interactions
+-----------------------
+The proposal simplifies the language and the compiler.
+
+
+Costs and Drawbacks
+-------------------
+This will be another deviation from the Haskell 2010 standard. Note that GHC already does not enable DatatypeContexts by default; this proposal is about removing the possibility to go back.
+
+Packages using DatatypeContexts will break. I downloaded the entire Hackage repository, used different regexes to search for usage and manually filtered out false positives. Out of 12907 hackage packages, I counted 181 in total using DatatypeContexts; out of those, 14 are deprecated, 120 didn't have an upload within the last 2 years, there are 47 remaining. `See details <https://gist.github.com/monoidal/77c6dd7490f7c9398db35b35273cc030>`_.
+
+In general, programs with `DatatypeContexts` can be updated just by removing the context. In rare cases, programs depending on type defaulting can `change semantics <https://prime.haskell.org/wiki/NoDatatypeContexts>`_. I don't expect this to occur in practice; worst-case, type defaulting can be detected with ``ghc -Wall``.
+
+
+Alternatives
+------------
+A more gradual plan is to change the tone of the deprecation warning in GHC 8.6 or 8.8, and only later remove the extension. I opted for GHC 8.8 in the proposal, since we are paying interest to maintain, and the feature has been deprecated 7 years ago.
+
+
+Unresolved Questions
+--------------------
+
+
+Implementation Plan
+-------------------
+I volunteer to implement.


treeowl:
Type defaulting isn't the only reason. Datatype contexts also let you write Haskell 98 code that would otherwise require `KindSignatures`. For example,

```haskell
data Monad m => Foo m a = Blah
```

will fix `m` to have kind `Type -> Type`.

These days, that seems like a terrible way to do it, so I support this proposal, but I think that effect is worth mentioning.

monoidal:
Added, thanks.

AntC2:
> * Attempting to write a datatype context statng that this feature was removed.

I hope this means the proposal will preserve GHC's ability to parse datatype contexts. Then good. I agree with tidying up the clutter of implementing the 'stupid theta' semantics (which seems to infect a surprising amount of code).

I can't help feel there is something useful that syntax could be doing. (And several people made that point back in 2010 when it was discussed on Haskell Prime.)

Perhaps historically there was some different intent for the feature? And its implementation/spec in the H98 standard somehow missed the message? Can any 'old hands' remember/guess?

    data Num a => T a = MkT a

with the constraint appearing to left of the type name kinda suggests `Num a` should scope over _both_ the type and the data constructor. Such that `T a` appearing in a signature should expect/enforce `Num a`.

amigalemming:
@treeowl  I had a case similar to `Foo m a` and solved it like so:
~~~~
newtype Foo m a = Foo (Const (m a) ())
~~~~

amigalemming:
It is long ago that I understood that datatype contexts do not do what I need. Since then I did not use them anymore. I don't mind if they disappear.

treeowl:
What's the plan for eventually removing the flag altogether.

monoidal:
> I hope this means the proposal will preserve GHC's ability to parse datatype contexts. 

In the proposal, the parser part remains only to give a helpful error message, rather than a generic syntax error. Internally, currently `data Num a => X a` and `class Num a => X a` are parsed using the same rule `tycl_hdr`, and I think we can just keep this.

> Perhaps historically there was some different intent for the feature? And its implementation/spec in the H98 standard somehow missed the message?

DatatypeContexts went through some refinements historically if you look at previous standards, but I don't think any of those are satisfactory.

In Haskell 1.0, if `data c => T a` is present, than any typeclass instance of T must include the context: `instance c => C (T a)`. In Haskell 1.2, constraints in types of constructors, but not during pattern matching.

The main problem - requiring constraints that are nowhere used and silently discarded - is not addressed.

> What's the plan for eventually removing the flag altogether.

This is not in scope of this proposal. I think eventually we could remove deprecated flags like RelaxedPolyRec, MonoPatBinds, Generics, and DatatypeContexts (if this proposal is accepted); but I'd like this to be a separate discussion.

Ericson2314:
I'm extremely enamored with https://arxiv.org/abs/1706.09715, which makes constrained datatypss useful and necessary.

treeowl:
@Ericson2314, that paper doesn't mention constrained datatypes at all. Could you please explain how it necessitates them?

treeowl:
@Ericson2314, those are constraints on data constructors, which no one wants to get rid of. Datatype contexts are something else altogether.

amigalemming:
On Sun, 19 Aug 2018, John Ericson wrote:

> @treeowl see "7.3 Runtime Efficiency" for an example. The idea is both associated types and methods alike will
> have (promoted) constraints in their (kinds /) types. That means constructors involving an associated type need
> the constraint.

Datatype contexts are contexts of the whole datatype, whereas you mean 
contexts of single constructors.

Ericson2314:
That's true and my mistake. But there are other connections (my own reasoning, be warned):

1. Consider what that section says about no dictionary being needed. What would properly go there would be a promoted constraint / irellevant constraint quantifier. The `=>` in data contexts is effectively just that; one could imagine it being sufficient for associated types in Haskell 98-style data declarations, (while imposing the requirement of an irrelevant constraint on each constructor whether or not it is needed for an associated type with GADT syntax.)

2. The kinds of the associated type/data type constructors themselves resemble the kinds of type constructors with data contexts. See https://github.com/ghc-proposals/ghc-proposals/pull/102#issuecomment-410466409

monoidal:
@Ericson2314 Consider `data Num a => T a = MkT1 a | MkT2 a`. Even though `Num a` constraint is a property of `T`, and it syntactically looks to be putting a constraint on `T`, it is actually affects `MkT1` and `MkT2` only. It is not putting any constraints on `T`, even though it looks it's doing something. It is not changing the kind of `T`. I really think of this as `data T a = Num a ~> MkT1 a | Num a ~> MkT2 a`, where I deliberately did not use the => arrow (because it has a different meaning). The ~> arrow, unlike =>, has confusing semantics: it requires the `Num a` constraint both during construction and during pattern matching. This is why it's generally discouraged.

The paper you linked is about actually putting constraints on types like `T`, and not just on all constructors like DatatypeContexts does.

Ericson2314:
OK. To be clear I don't believe `DatatypeContexts` is useful in its current state. Rather I think with research on the horizon it would generalize into something meaningful. I'll do my best to sketch what this looks like

First, quantifiers. Some of the missing combinations of @goldfirere's quantifiers are dependent and irrelevant constraints. This would enable of the use of type classes at compile time, while erasing all dictionary passing (without specialization) at run-time. I include the table in full, with my editions at the bottom

| Quantifier  | Dependence | Relevance | Visibility    | Matchability |
| --------------- | -------------- | ------------- | ----------------- | ---------------- |
| `forall a.`     | dependent      | irrelevant    | invisible (unif)  | unmatchable      |
| `forall a '.`   | dependent      | irrelevant    | invisible (unif)  | matchable        |
| `forall a ->`   | dependent      | irrelevant    | visible           | unmatchable      |
| `forall a '->`  | dependent      | irrelevant    | visible           | matchable        |
| `foreach a .`   | dependent      | relevant      | invisible (unif)  | unmatchable      |
| `foreach a '.`  | dependent      | relevant      | invisible (unif)  | matchable        |
| `foreach a ->`  | dependent      | relevant      | visible           | unmatchable      |
| `foreach a '->` | dependent      | relevant      | visible           | matchable        |
| `ty =>`         | non-dependent  | relevant      | invisible (solve) | unmatchable      |
| `ty '=>`        | non-dependent  | relevant      | invisible (solve) | matchable        |
| `ty ->`         | non-dependent  | relevant      | visible           | unmatchable      |
| `ty '->`        | non-dependent  | relevant      | visible           | matchable        |
| --------------- | -------------- | ------------- | ----------------- | ---------------- |
| `forall (_ :: ty) =>`   | dependent      | irrelevant    | invisible (solve)           | unmatchable      |
| `forall (_ :: ty) '=>`  | dependent      | irrelevant    | invisible (solve)           | matchable        |
| `foreach (_ :: ty) =>`   | dependent      | relevant    | invisible (solve)           | unmatchable      |
| `foreach (_ :: ty) '=>`  | dependent      | relevant    | invisible (solve)           | matchable        |

The dependency without a variable (invisibility) is odd, but this is because solving takes care of that for us. If you were to look at the hypothetical Code, you'd see the dictionary being bound and used within the type. The `forall =>` variant is what I've been describing, the `foreach =>` variant is for both compile-time and run-time use, for completeness.

Next, I'd argue that what we have today is morally not `ty =>`, but `foreach (_ :: ty) =>`. So:
```haskell
data Num a => T a
  = MkT1 a
  | MkT2 a
```
to take your example, should rightfully be:
```haskell
data T :: foreach a -> Num a => Type where
  MkT1 :: forall a -> forall (_ :: Num a) => a -> T a
  MkT2 :: forall a -> forall (_ :: Num a) => a -> T a
```
[N.B. Plain `=>` is still used in the type constructor for the same reason `foreach` is. This fact is stated in https://github.com/ghc-proposals/ghc-proposals/pull/102 without much explanation. I gather it is because "relevant" doesn't absolutely mean run-time, but is rather a phase-relative statement. Type constructors applied to irrelevant arguments must necessarily evaluate to an irrelevant value.)

More interesting is the `forall (_ :: Num a)` in each constructor. Crucially, this basically matches the behavior of today! Pattern matching rightfully would provide the *irrelevant* constraint-proof, but as those don't exist today (nothing can provide or use them), we get nothing. Likewise, constructing merely requires the irrelevant constraint proof, but as that doesn't exist, the full relevant one must be provided instead. Finally, by the definition of irrelevant, no dictionary is included in the constructor as per today.

The paper comes in because the one meaningful thing one could actually do with irrelevant constraints is apply associated type/data constructors. Combining @goldfirere's thesis and paper, associated type families are functions like any other method, taking a usual constraint. When used as they are today, they are used in an irrelevant position an thus can be applied with irrelevant arguments. That in turn places a `forall (_ :: Num a) =>` obligation on the call site.

So to use the examples from https://github.com/ghc-proposals/ghc-proposals/pull/102#issuecomment-410466409, we'd have class methods:
```haskell
Foo :: foreach k . foreach (x :: k) -> A x => SomeK
Bar :: foreach k '. foreach (x :: k) '-> A x '=> SomeK
```
which then used in irrelevant position create obligations:
```haskell
forall k . forall (x :: k) -> forall (_ :: A x) => ...
forall k '. forall (x :: k) '-> forall (_ :: A x) '=> ...
```

AntC2:
@treeowl , @amigalemming 

> Datatype contexts let you also write Haskell 98 code that would otherwise need `KindSignatures`.

I believe the classic work-round for your example (within H98 but not needing Datatype contexts) is a dummy type constructor:

    data Foo m a = Blah
                 | Dummy (m a)    -- don't export this constructor!

AntC2:
@Ericson2314 I don't think this is the place to discuss what seems a totally different proposal -- perhaps kick off a github issue?

I wanted to be sure the parser continues to 'reserve' the syntax for future use. So thanks @monoidal for confirming that. Also thanks for the historical research.

monoidal:
@Ericson2314 Consider `data (Show a, Eq b) => T a b = T1 a | T2 b`. With DatatypeContexts, the `T1` constructor has only `Show a` constraint. In your proposal, it would have both `Show a` and `Eq b`. This means your idea is not backwards compatible; it's not explaining away DatatypeContexts by something else. Therefore, we cannot just change DatatypeContexts to your implementation, if you'd like to go forward with this, it's a different proposal that should have its own feature flag, say NewDatatypeContexts.

It is true that, since DatatypeContexts dictionaries are discarded, if we implemented irrelevant arguments then the dictionaries could be made irrelevant. However, this is yet another interaction of a GHC feature with DatatypeContexts that will need to be built and maintained. I don't see this as a good thing.

Ericson2314:
@monoidal OK it will need a new proposal, and maybe new research too. [I'd argue if we can remove deprecated extensions we can break them too, so maybe not a new extension name, but this is besides the point.]

My "tl;dr" invalidated, I suppose my remaining point is less that we should continue this as a user-visible feature, but that the code shouldn't be deleted until either something like what I'm saying is pursued or ruled out entirely. So it would be better to me to justify this proposal on keeping on options open with the surface language than lessening GHC's maintenance burden per-se, as I hope the code implementing this can continue to limp along for ~20 more months.

goldfirere:
While I appreciate the sentiment in https://github.com/ghc-proposals/ghc-proposals/pull/161#issuecomment-414195421, I don't fully agree with what @Ericson2314 is saying.

To wit: I think `forall (_ :: ty) =>` is hogwash. After all, we're talking about the syntax for the surface, user-written language, not Core. If we don't use a variable later in a type, it's not dependent. What happens when we translate to Core is a separate matter. I agree with @Ericson2314 that we will want dependent dictionaries in Core, but this shouldn't affect Haskell.

As someone who generally believes in removing old cruft (cf. #157), I would normally find myself in support of this proposal. However, I had a hunch that this had some interaction with Constrained Type Families, and I'm grateful to @Ericson2314 for exploring this. But I don't think the interaction is as he describes. Instead, what about this: if we have `data Ord a => Set a`, then every occurrence of `Set` induces a constraint on the argument. That means that a declaration like `f :: Set a -> Set a` would become invalid -- it would instead have to be `f :: Ord a => Set a -> Set a`, even if the function were the identity. This relates to Constrained Type Families because this constraint requirement is exactly like how that paper requires `C a` to be in scope just to mention `F a` (for `class C a where type F a`).

Perhaps this should go in a separate proposal, but would `-XDatatypeContexts` cease to be a misfeature if we had this rule (in addition to existing behavior):

* Suppose we have declared `data theta => T a`. Then, whenever `T ty` appears in a type, GHC must satisfy the constraint `theta[a |-> ty]`. (Generalize straightforwardly to multiple-argument datatypes.)

Now, a datatype context really does narrow the domain of a type constructor, always. Does a datatype context ever *increase* expressiveness? I think not. But if the argument to a `Set` should always be in the `Ord` class, it's surely nice to be able to declare this.

Ericson2314:
(We can take this elsewhere, but while I'm here...)

> To wit: I think `forall (_ :: ty) =>` is hogwash.

:D

```haskell
class Foo a where type Bar a

instance Foo Bool where Bar Bool = Int
instance Foo Word where Bar Word = Bool

f :: forall a -> forall (_ :: Foo a) => a -> (a, Bar a)
f Bool b = (b, 7)
f Word n = (n, False)
```

Sure there's no variables but *semantically* the output type is dependent on the constraint. There's no closed term inhabited by `(b, 7)` and `(n, False)`. Maybe this boils down to vocabulary quibbles, but to me an both irrelevant and non-dependent quantifier must be dead code.

treeowl:
> ... if the argument to a `Set` should always be in the `Ord` class, it's surely nice to be able to declare this.

I thoroughly disagree, because such a restriction actually *impedes* communication through parametricity. The type signature

```haskell
fromDistinctAscList :: [a] -> Set a
```

loudly proclaims that `fromDistinctAscList` completely ignores the ordering of `a` elements and works only structurally&mdash;there's no need to consider how `Ord` performance might affect that operation.

The *only* times I've found such a constraint useful have involved type families. For example, given

```haskell
type family Foo (r :: Type -> Type) :: Type
f :: Foo (Rep a) ~ ... => a -> ...
```

applying `f` to a type that has no `Generic` instance will lead to an awful error. Adding a `Generic` constraint still gives an awful error, but also gives a very useful one.

AntC2:
> If we have `data Ord a => Set a`, then every occurence of `Set` induces ...

Is not really how Datatype contexts work, as @monoidal [points out](https://github.com/ghc-proposals/ghc-proposals/pull/161?_pjax=%23js-repo-pjax-container#issuecomment-414342691), and see the example in the Language Report: the constraint attaches (or not) to the data constructors, only indirectly (or not) to the type constructor. I've never understood why Datatype contexts behaves like that. It seems eminently stupid. And the opposite of intuitive when syntactically the context appears to the left of the type constructor.

I think we should all stop trying to hijack this proposal in an attempt to make the syntax do something more useful. The time to do that was the Haskell Prime discussion 2010. What we should do now is:

* debate the proposal's proposed timescale;
* make sure the syntax will be 'reserved' for future use/a well-designed behaviour -- which @monoidal has confirmed;
* move the debate about any behaviour elsewhere.

monoidal:
I think we can discuss the alternatives here; it's helpful to explore the design space. I've learned something by checking old reports and by seeing the connection to constrained type families.

I'm fine with reusing the DatatypeContexts syntax for the proposal in https://github.com/ghc-proposals/ghc-proposals/pull/161#issuecomment-414540445. I have questions though:

* What happens when `T` occurs without `ty`? For example, given `data Apply f x = Apply (f x); data Ord a => Set a` would the signature `identity :: Apply Set a -> Apply Set a` be legal without `Ord a =>`?
* Is there motivation? Supposing we could get this version of DatatypeContexts for free as a byproduct of implementing "constrained type families", where would it be useful?
* Should this be a change in the behavior of DatatypeContexts, or a new flag, say NewDatatypeContexts? I'd prefer a new flag: (a) I think of extensions as choosing the slice of the typesystem I'm using, and this slice seems to be different for DatatypeContexts and NewDatatypeContexts; (b) This could cause silent change of behavior due to defaulting (as `data Floating a => Foo a = X` does not constrain `a` in the old version). This is unlikely to be an issue in practice though.

Ericson2314:
@monoidal 
1. Pretty sure it would require it. `Apply Set` is morally and "corely" something like `\ @_ => Apply (Set @_)`.
2. It would be required to use type families in fields, per what @treeowl said. When no type families are actually used, I suppose it allows cautious library authors to reserve the right to add them later.

AntC2:
Sorry, @treeowl, I'm not following what you think you're proclaiming

> ... such a restriction actually _impedes_ communication through parametricity. The type signature
>
>        fromDistinctAscList :: [a] -> Set a
>
> loudly ...

I see a long name composed of English words/abbreviations. I see no constraints. So whatever it's doing loudly is sound and fury, signifying nothing.

> ... proclaims that `fromDistinctAscList` completely ignores the ordering of a elements

Does it? If the list is ascending, its elements must have an ordering. But I see no `Ord a`, so the proclamation is bogus. (Let alone anything guaranteeing the List is in fact in ascending order.)

> and works only structurally—there's no need to consider how `Ord` performance might affect that operation.

I don't take @goldfirere's comment to be talking about performance. It's talking about documenting properties (and getting the compiler to check the code complies -- complies with what and how, is what's under debate).

AntC2:
> I think we can discuss the alternatives here;

Then with your permission ...

> I've learned something ...

There's quite  few questions on StackOverflow around what the datatype context means. The answers have in general not addressed the question, but have expounded that GADT's do 'it' better. Given that we have GADTs with perfectly sensible syntax, it would be a waste to treat a datatype context as merely some shorthand for a GADT.

OTOH it would be good to give the to-be behaviour in terms of already existing features.

> Should this be a change in behaviour of `DatatypeContexts`, or a new flag, say `NewDataTypeContexts`?

Nobody's arguing for continuing the current behaviour AFAICT. Yes there had better be a different flag. `New...` anything would only prompt confusion: what was the old behaviour? do I need to understand that first? So I'd prefer a specific meaningful name for the flag. What that's to be will drop out of deciding the to-be behaviour.

AntC2:
> What happens when `T` occurs without `ty`?

I say we should insert the context from `T`'s declaration. How? Let's explain it in terms of the proposed top-level sgnatures; and pattern synonyms' Required and Provided contexts. (The current behaviour doesn't do any providing of the context.) Given

    data Ord a => Set a = Set a

Treat that as:

    data Set a = Set_ a                                              -- Set_ keep abstract
    type Set :: Ord a => a -> TYPE
    pattern Set :: Ord a => Ord a => a -> Set a
    pattern Set x = Set_ x

(I might be getting my syntax wrong for top-level signatures.)

Furthermore the context is to apply whether or not the constrainted tyvar appears as a parameter to some constructor. Your earlier example `data (Show a, Eq b) => T a b = T1 a | T2 b` treat as:

    data T a b = T1_ a | T2_ b
    type T :: (Show a, Eq b) => a -> b -> TYPE
    pattern T1 :: (Show a, Eq b) => (Show a, Eq b) => a -> T a b
    pattern T1 x = T1_ x
    pattern T2 :: (Show a, Eq b) => (Show a, Eq b) => b -> T a b
    pattern T2 x = T2_ x

nomeata:

nomeata:

goldfirere:
People that care about this proposal may be interested in https://richarde.dev/papers/2019/partialdata/partialdata.pdf, which envisions a glorious new future for datatype contexts. But there are some downsides to that proposal, and I'm not fully convinced that paper should be implemented (though it's surely worth pondering, in my view). In any case, it's worth posting here.

PR #163, opened 2018-08-22
Remove an undocumented StableName guarantee

treeowl:
I believe we can make the `StableName` mechanism considerably
simpler and somewhat more efficient by removing a de facto
(undocumented) conditional guarantee about `hashStableName`
injectivity. I propose to do so.

[Rendered](https://github.com/treeowl/ghc-proposals/blob/stable-names/proposals/0000-stable-names.rst)

### DIFF
--- a/proposals/0000-stable-names.rst
+++ b/proposals/0000-stable-names.rst
@@ -31,9 +31,9 @@ a stable name table at all. We can, I believe, get away with just one hash
 table per GC generation and a radically simpler garbage collection strategy.
 
 What the stable name table really lets us do is make an *undocumented*
-guarantee: if ``sn1`` and ``sn2`` are stable names, that are currently
-*live*, and ``hashStableName sn1 = hashStableName sn2``, then
-``sn1 = sn2``.
+guarantee: if ``sn1`` and ``sn2`` are stable names, ``sn1`` was alive
+when ``sn2`` was created by ``makeStableName``, and
+``hashStableName sn1 = hashStableName sn2``, then ``sn1 = sn2``.
 
 Proposed Change Specification
 -----------------------------
@@ -84,12 +84,11 @@ For example, we could implement a map from stable names to values like so: ::
      Nothing -> pure Nothing
      Just (sn, v) -> do
        touch sn
-       touch snk
        pure (Just v)
 
 We don't need to worry about hash collisions on lookup because
-we ensure that the stable names are both alive at the end of
-the operation, and therefore the equality of their hashes implies
+we ensure (using ``touch``) that ``sn`` is alive when ``snk``
+is created, and therefore the equality of their hashes implies
 their equality.
 
 There is a clear trade-off here between complexity of code using
--- a/proposals/0000-stable-names.rst
+++ b/proposals/0000-stable-names.rst
@@ -67,6 +67,10 @@ The obvious alternative is to *document* the de facto guarantee. This would
 allow some (very carefully written) code to be simpler and/or more efficient.
 For example, we could implement a map from stable names to values like so: ::
 
+ touchStableName :: StableName a -> IO ()
+ touchStableName (StableName sn) =
+   IO $ \s -> (# touch# sn s, () #)
+
  newtype SNMap k v = SNMap (IntMap (StableName k, v))
 
  empty :: SNMap k v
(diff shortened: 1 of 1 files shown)


MarisaKirisame:
what is the "GC masters"?

treeowl:
@MarisaKirisame, I meant the people who know more about garbage collection than the other I know.

simonmar:
`hashStableName` doesn't have the guarantee mentioned in the proposal.  For example:

```
  do 
    sn1 <- makeStableName x
    sn2 <- makeStableName y
    return (hashStableName sn1 == hashStableName sn2)
```

(where `x` and `y` are distinct). This could return `True` as follows. GHC transforms it to this:

```
  do 
    sn1 <- makeStableName x
    let h1 = hashStableName sn1 
    sn2 <- makeStableName y
    let h2 = hashStableName sn2
    return (h1 == h2)
```

and then GC strikes right after `let h1 = ...`, which recovers the StableName object for `sn1`, and reuses the same entry for `sn2`.

So I think the propsal is unnecessary - the invariant already cannot be relied on, so there's nothing to change. It's entirely possible that the implementation can be made simpler, but that's a separate issue.

treeowl:
@simonmar, if you carefully keep the stable names alive as I described, you can get the guarantee.

simonmar:
Ok, I didn't notice the mention of "live". But still, I'm not sure what the purpose of this proposal is - any code relying on an undocumented invariant is broken.  Are you asking whether we should document the invariant? I don't think so, but it's nice if `hashStableName` produces as few collisions as possible, because that makes hash tables built with it more efficient. If a different implementation of stable names produced more collisions, that would be unfortunate (but not necessarily a deal breaker).

treeowl:
The implementation I'm thinking it would produce very, very few collisions (only once the word-sized counter overflows).

treeowl:
> [...] any code relying on an undocumented invariant is broken. Are you asking whether we should document the invariant? I don't think so [....]

Right. There's a trade-off here. One option is to remove the invariant, which makes the `StableName` implementation simpler and more efficient. The other option is to document the invariant, which would make some programs simpler and/or more efficient. I figured I should check in with the community to make that decision. I don't want to go replace a big, complicated piece of the run-time system with something much simpler only to discover that people would rather document the "accidental" invariant that makes it so complex!

As a worst-case example, consider user code that wants to simulate the current `StableName` operations under the new regime. It would have to use a weak hash table from stable names to integers (simulating stable name table indices). It would also need some mechanism to allocate (on `myMakeStableName`) and free (by `StableName` key finalizer) those integers. Of course, if no one (or almost no one) really wants/needs such a thing, we don't need to worry about it.

Note also that the current arrangement can also (I believe) provide an extra (not-currently-implemented) feature:

```haskell
getStableNameUnderlying :: StableName a -> IO (Maybe a)
```

taking advantage of the fact that the stable name table entry holds a weak reference to the underlying object.

This is the sort of thing where "It's free because it's already paid for", so someone other than me has to decide whether we want to stop paying for it, or whether we want to keep paying the price but get what w pay for.

treeowl:
Actually, that extra `getStableNameUnderlying` feature could also be added to the cheaper implementation, at the cost of one pointer per stable name object. So I guess that's orthogonal.

simonmar:
The need to keep the `StableName` objects live for the property to hold makes it tricky to exploit. I can't imagine documenting it in a sane way - what does "live" mean? No, let's not go that route.

treeowl:
@nomeata, please submit this to the committee.

treeowl:
I just made a last-minute edit, because I realized I could state the de facto guarantee more strongly and a bit more precisely. I still don't think it's worth the implementation complexity cost, but I want to be as fair as possible to the counterargument.

treeowl:
@simonmar, I agree with your claim that liveness is usually subtle, but thanks to the extremely limited `StableName#` API, I think we can be pretty precise about what it means here. In particular, the only sensible and reliable way to keep a stable name alive is with `touch#`. All other potential ways are fragile or involve horrors like `lazy`. And as my recent edit shows, we only need to worry about whether a given stable name is alive when a certain other one is *created*. All that is in `IO`, so we don't have to reason about graph reduction and such.

PR #165, opened 2018-08-31
Use nontotal precedence for operators.

MarisaKirisame:
[Rendered](https://github.com/MarisaKirisame/ghc-proposals/blob/patch-1/proposals/0000-nontotal-precedence.rst)

### DIFF
--- a/proposals/0000-nontotal-precedence.rst
+++ b/proposals/0000-nontotal-precedence.rst
@@ -1,13 +1,11 @@
-Proposal title
-==============
+Nontotal operator precedence
+============================
 
 .. proposal-number::
 .. trac-ticket::
 .. implemented::
 .. highlight:: haskell
-.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
-            **After creating the pull request, edit this file again, update the
-            number in the link, and delete this bold sentence.**
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/165>`_.
 .. sectnum::
 .. contents::
 
@@ -19,11 +17,11 @@ and library writer need not worry about selecting the best fixity, as it could b
 
 Motivation
 ------------
-This proposal is motivated by 'https://www.reddit.com/r/haskell/comments/27ifst/how_do_you_choose_the_fixity_of_operators_when/'.
+This proposal is motivated by https://www.reddit.com/r/haskell/comments/27ifst/how_do_you_choose_the_fixity_of_operators_when/.
 
 In the reddit post, the top comment is a redditor arguing for a partial order.
 
-The second top comment is another redditor suggesting to use :info on operators that is related, and choose the precedence accordingly.
+The second top comment is another redditor suggesting to use ``:info`` on operators that is related, and choose the precedence accordingly.
 
 In effect, what the second top comment is suggesting, is that we declare operator precedence by solving a partial ordering constraint.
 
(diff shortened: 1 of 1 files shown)


nomeata:
Having had similar thoughts before, I am looking forward to this discussion. I hope we can get input from people who know the design space a bit better.

Concretely, I am not sure if not having transitivity is the right choice. Could you add more motivation to this design choice?

How does this compare to simply allowing rational numbers as precedence levels (which allows you to always refine the total order?)

MarisaKirisame:
@nomeata thx for helping to clean up! This is my first time submitting to ghc-proposal, and I am not a good user of rst. Sorry for the inconvenience.

The two question you raise has a similar motivation. 
I think, we should be aware of 'accidental binding'. That is, when two operator has little dependency, they should simply not bind together.
It will make code easier to read, and there will be less 'I thought it is supposed to bind this way' kind of error. It is the main benefit of this approach over rational number.
Another benefit is that, when we use constraint, it is always possible to refine our operators by adding more constraint. It might break bw compat when there is loop formed, but that's it. 
I think we can prove that adding more constraint will only make some concrete syntax go from error to parsed, or parsed to error, but not parsed to a different parsed ast.
However, if we use rational, it is fixed: if something isnt bind the way you like, you cannot change it without breaking backward compatibility (it might be parsed differently now).

The argument about transitivity is the same, as the parsing mixifix operator paper comment, + and && mixing will only make sense, when + is instantiated to Bool.
It is also easier (in term of) to move from non transitive to transitive, so if transitive turn out to be what we want, we can migrate relatively more easily by adding a lot of edge automatically in the operator precedence graph.

MarisaKirisame:
@nomeata the discussions seems to ebbed down (or it was never any discussions to begin with). Is there anything I should do (or is it good in your opinion)?

treeowl:
I think non-transitive precedence seems very strange. Could you expand on why you think that makes sense? Among other things, it seems to require *lots* of declarations.

phadej:
Small comment:
- There isn't `Num Bool`, so `+` / `&&` argument doesn't work.
- @MarisaKirisame recall that committee won't read comments when considering the proposal, so all unclear things have to be clarified in the text of the proposal.

simonmar:
Speaking as someone who has implemented precedence parsing several times and worked on the specification of precedence in the Haskell report, I don't think I'm very keen on this proposal. It seems to be very verbose, for one thing.

The idea of using fractional or rational precedences has been proposed a number of times in the past, and would be an easy addition.

Can I suggest that proposals for changes to operator precedence be accompanied by an equivalent to the sample implementation in the Haskell report: https://www.haskell.org/onlinereport/haskell2010/haskellch10.html#x17-18100010.6, or at least a description of how this implementation would change?

Ericson2314:
@simonmar 

> The idea of using fractional or rational precedences has been proposed a number of times in the past, and would be an easy addition.

To me, the big disadvantage of any total preorder (being pendantic since operators can share precedence) is simply that there are too many precedence rules. Unrelated things end up being related because the total order is kluge: it's just a topological sort of the partial order in our heads. For example, who here off the top of their head knows whether `+` or `||` binds tighter?

For base, the ship has sailed (for the time being). But I can imagine someone experimenting with a beginner language where `*` < `+` and `&&` < `||`, but the arithmetic and boolean operators are unrelated.

> Can I suggest that proposals for changes to operator precedence be accompanied by an equivalent to the sample implementation in the Haskell report: https://www.haskell.org/onlinereport/haskell2010/haskellch10.html#x17-18100010.6, or at least a description of how this implementation would change?

Is OK to change the spec so instead of operator tokens carrying their own precedence and fixity, there are oracles:
```haskell
getPrec :: Op -> Prec
compareFixity :: Op -> Op -> Ordering
```
Then I believe the algorithm works at least for any preorder. So the difference between the status quo and the proposal is entirely in the definition of those two functions.

Ericson2314:
> I think non-transitive precedence seems very strange.

```
1 `rock` 2 `paper` 3 `scissors` 4
```
Can be parsed as
```
((1 `rock` 2) `paper` 3) `scissors` 4
(1 `rock` (2 `paper` 3)) `scissors` 4
1 `rock` ((2 `paper` 3) `scissors` 4)
1 `rock` (2 `paper` (3 `scissors` 4))
```
But assuming the opposite of the roshambo directed graph (winner, binds looser)
```
((1 `rock` 2) `paper` 3) `scissors` 4
1 `rock` ((2 `paper` 3) `scissors` 4)
```
are both valid parses.

Conclusion, we'll be in some sort of arrow theorem madness unless we require partial orders / take the reflexive-transitive closure.

Ericson2314:
My guess would be when the differences between `infixl` `infixr` and `infix` are taken into account, we already have a partial preorder over `(Op, Bool)`, which is the fundamental structure of the current approach. So arguably this just allows more partial orders, and doesn't even introduce the generalization to partial orders.

sgraf812:
I can see a potential drawback here, compared to just sticking to a total relation.

Consider I write two related libraries, like `twitter-conduit` and `cereal-conduit`, that don't (need to) depend on each other, but are meant to be used together. Let's assume for the sake of the example that both define operators with their precedence declared wrt. some operator in `conduit`. Now, *as a library author* of both libraries, I cannot specify the precedence relation between my two operators even if I wanted to.

This is kind of like the dreaded diamond situation we have with type classes.

sgraf812:
Also I think giving up transitivity would be a mistake, for exactly the reasons @Ericson2314 gave.

simonmar:
> who here off the top of their head knows whether + or || binds tighter? I certainly don't!

The fact that `+` and `||` are unnecessarily ordered doesn't keep me up at night.

> getPrec :: Op -> Prec
> compareFixity :: Op -> Op -> Maybe Ordering

Yes exactly, I'm just suggesting that this would be a good way to specify the proposal precisely.

Ericson2314:
@sgraf812 so first of all in this proposal 0-9 still exist as "virtual operators" in the partial order. So you can still use that to do the exact same thing.

I rather just have some way to opt into allowing orphan precedence declarations, however. Or, they can be locally scoped which doesn't pollute things they way agda-style implicits do.

Ericson2314:
@simonmar Done! (Do note I wrote my functions backwards, it's `getFixity` and `comparePrec`, in actuality).

```haskell

newtype Op = Op String
  deriving (Eq,Show)

getFixity :: Op -> Fixity
getFixity (Op "") = Nonfix
getFixity (Op "-") = Leftfix
getFixity _ = error "based on declarations"

comparePrec :: Op -> Op -> Maybe Ordering
comparePrec (Op "") _ = Just LT
comparePrec _ _ = error "based on declarations"

data Fixity = Leftfix | Rightfix | Nonfix
  deriving (Eq,Show)

data Exp = Var Var | OpApp Exp Op Exp | Neg Exp
  deriving (Eq,Show)

data Tok = TExp Exp | TOp Op | TNeg
  deriving (Eq,Show)

resolve :: [Tok] -> Maybe Exp
resolve tokens = fmap fst (parseNeg (Op "") tokens)
  where
    parseNeg :: Op -> [Tok] -> Maybe (Exp,[Tok])
    parseNeg op1 (TExp e1 : rest)
       = parse op1 e1 rest
    parseNeg op1 (TNeg : rest)
       = do guard (comparePrec op1 (Op "-") == Just LT)
            (r, rest') <- parseNeg (Op "-") rest
            parse op1 (Neg r) rest'

    parse :: Op -> Exp -> [Tok] -> Maybe (Exp, [Tok])
    parse _   e1 [] = Just (e1, [])
    parse op1 e1 (TOp op2 : rest)
       -- case (1): check for illegal expressions
       | (comparePrec op1 op2 == Just EQ) && ((fix1 /= fix2) || (fix1 == Nonfix))
       = Nothing

       -- case (2): op1 and op2 should associate to the left
       | (comparePrec op1 op2 == Just GT) || (op1 == op2 && fix1 == Leftfix)
       = Just (e1, TOp op2 : rest)

       -- case (3): op1 and op2 should associate to the right
       | otherwise
       = do (r,rest') <- parseNeg op2 rest
            parse op1 (OpApp e1 op2 r) rest'
       where
         fix1 = getFixity op1
         fix2 = getFixity op2
```

Ericson2314:
(I also added more parens so fixity resolution doesn't depend on fixity resolution :joy:)

Ericson2314:
I realized I forgot a case. Edited to fix that and usage patterns instead of guards. I prefer the resulting code because:
1. Left and right associativity is handled symmetrically.
2. Patterns make forgetting cases as I did harder

nomeata:

nomeata:

PR #166, opened 2018-09-07
List instances for a type in GHCi

xldenis:
This is a proposal to add a GHCi command that lists all valid instances for a given complete or partial type signature. 

[Rendered](https://github.com/xldenis/ghc-proposals/blob/master/proposals/0000-ghci-instances.rst)

### DIFF
--- a/proposals/0000-ghci-instances.rst
+++ b/proposals/0000-ghci-instances.rst
@@ -74,7 +74,8 @@ The primary costs for this feature are implementation time. Because it is a GHCi
 
 Alternatives
 ------------
-We don't implement this.
+
+Currently GHCi has an ``:info`` command which offers some information on instances. However, ``:info`` only works on single words, which means that ``:info Either String Int`` will return results for ``Either``, ``String``, and ``Int`` separately. This makes it a partial replacement at best for the ``:instances`` command.
 
 
 Unresolved Questions
--- a/proposals/0000-ghci-instances.rst
+++ b/proposals/0000-ghci-instances.rst
@@ -49,12 +49,12 @@ Proposed Change Specification
 This proposal adds a new command to GHCi called ``:instances`` which provides a listing of all valid instances for a given type.
 
 Syntax
-======
+~~~~~~
 
 Valid queries to ``instances`` are types, if anything other than a type is provided as an argument, the command will return an error explaining the intended usage. The command interprets everything after ``instances`` as a single type, this means there is no way to specify multiple types to lookup instances of MPTCs. Holes can be used to represent free variables in the argument type, for example: ``Either _ _``. To express relations between multiple holes, named holes can be used, for example:: ``Either _a _a``.
 
 Execution
-=========
+~~~~~~~~~
 
 Provided with a valid type, ``instances`` will attempt to match it against the heads of all visible class instances and satisfy all the implied constraints. The output will consist of a formatted listing of all matching and satisfiable instances. Each instance should be simplified as much as possible, meaning that if an instance: ``(c ~ Bool) => C c`` were found it would be presented as ``C Bool``.
 
@@ -108,6 +108,7 @@ Often when dealing with monad stacks, it can be useful to figure out exactly whi
   MonadState (StateT _s (ReaderT _r IO))
   MonadReader (StateT _s (ReaderT _r IO))
   ...
+
 
 Unresolved Questions
 --------------------
--- a/proposals/0000-ghci-instances.rst
+++ b/proposals/0000-ghci-instances.rst
@@ -56,7 +56,7 @@ Valid queries to ``instances`` are types, if anything other than a type is provi
 Execution
 ~~~~~~~~~
 
-Provided with a valid type, ``instances`` will attempt to match it against the heads of all visible class instances and satisfy all the implied constraints. The output will consist of a formatted listing of all matching and satisfiable instances. Each instance should be simplified as much as possible, meaning that if an instance: ``(c ~ Bool) => C c`` were found it would be presented as ``C Bool``.
+Provided with a valid type, ``instances`` will attempt to match it against the heads of all visible class instances and reduce constraints until only those mentioning holes in the type remain. The output will consist of a formatted listing of all matching instances that satisfy the stated condition on constraints. Each instance should be simplified as much as possible, meaning that if an instance: ``(c ~ Bool) => C c`` were found it would be presented as ``C Bool``.
 
 
 Effect and Interactions
--- a/proposals/0000-ghci-instances.rst
+++ b/proposals/0000-ghci-instances.rst
@@ -1,0 +1,87 @@
+.. proposal-number:: Leave blank. This will be filled in when the proposal is
+                     accepted.
+.. trac-ticket:: Leave blank. This will eventually be filled with the Trac
+                 ticket number which will track the progress of the
+                 implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/166>`_.
+.. sectnum::
+.. contents::
+
+List instances for a type in GHCi
+==============
+
+There is currently no mechanism to get the instances for a specific type in GHCi, `:info` command allow specifying more than a top-level constructor. This proposal suggests a new GHCi command that would allow programmers to query GHC for valid instances matching a type.
+
+Motivation
+------------
+
+Typeclass resolution and deriviation are integral parts of modern Haskell. Unfortunately, the compiler doesn't volunteer a lot of information on them during development. Figuring out whether a type has a specific instance usually requires you to work backwards: starting from the typeclass and going through every instance in Haddock, recursively, to figure out if your type satisfies it. This experience isn't optimal and can get annoying for large or nested types and classes.
+
+This is exactly the kind of task that the compiler could help us with though. Since it already has all the visible instances for a given piece of code loaded, we could just ask it to list all instances that a type matches! This proposal suggests adding a new command and related machinery to GHCi to acomplish this. An example of this command's possible execution is shown below:
+
+.. code-block:: none
+
+  >> import Data.Functor.Sum (Sum)
+  >>
+  >> :instances Sum [] []
+  Eq   a => Eq   (Sum [] [] a)
+  Ord  a => Ord  (Sum [] [] a)
+  Show a => Show (Sum [] [] a)
+  Read a => Read (Sum [] [] a)
+  Functor  (Sum [] [])
+  Foldable (Sum [] [])
+  Eq1      (Sum [] [])
+  Ord1     (Sum [] [])
+  Show1    (Sum [] [])
+  Read1    (Sum [] [])
+  FunctorWithIndex     (Either Int Int) (Sum [] [])
+  FoldableWithIndex    (Either Int Int) (Sum [] [])
+  TraversableWithIndex (Either Int Int) (Sum [] [])
+
+It would provide a new tool for Haskell developers, both new and experienced that helps during during development and debugging. It allows a new way of talking with our beloved compiler and to better understand how it thinks.
+
+Proposed Change Specification
+-----------------------------
+
+This proposal adds a new command to GHCi called `:instances` which provides a listing of all valid instances for a given type.
+
+Additionally, besides taking a fully specified type, the command will accept partial type signatures, interpreting holes (``_``) as free variables. This allows ``:instances Either _ Int`` to return ``Eq a => Eq (Either a Int)`` as part of it's listing.
+
+The original `Trac ticket <https://ghc.haskell.org/trac/ghc/ticket/15610>`_ also suggests some further work. Notably:
+
+- **Negative Results**: a second command ``:noinstance`` would output a list of the unsatisfied constraints that prevent an instance from existing for a specific type.
+- **Multi-Parameter Type Classes**: I'm not sure yet what complications supporting multi-parameter typeclasses would entail but if it turns out to significantly complicate things then they could be pushed off to a further proposal.
+- **Trace Info**: Provide the source for every instance a type has.
+
+Effect and Interactions
+-----------------------
+
+This new command gives programmers the ability to see every possible instance for a type that the compiler sees. This is useful both for beggining and  `advanced <https://github.com/Iceland_jack>`_ haskellers who are exploring types and what they can do with them.
+
+When combined with recent extensions like ``GeneralizedNewtypeDeriving`` and ``DerivingVia``, a type can 'steal' many instances from other structurally identical ones but finding those instances is currently a tedious manual process.
+
+This also provides a way for beginners to see what's possible with the types given to them. Instead of having to browse Haddocks and stumble on instances for their type, GHC can simply provide all that information on the spot!
+
+Since this is simply a GHCi command, there will be no interaction with the language itself.
+
+Costs and Drawbacks
+-------------------
+
+The primary costs for this feature are implementation time. Because it is a GHCi command, it won't have any backwards compatibility or performance issues. It should actually help make Haskell a little more accessible to beginners as well, giving a better view on what GHC sees as possible instances for their types.
+
+Alternatives
+------------
+We don't implement this.
+
+
+Unresolved Questions
+--------------------
+
+
+Implementation Plan
+-------------------
+
+If accepted, I, `@xldenis <https://github.com/xldenis>`_ intend to implement this feature.
--- a/proposals/0000-ghci-instances.rst
+++ b/proposals/0000-ghci-instances.rst
@@ -80,6 +80,34 @@ Alternatives
 
 Currently GHCi has an ``:info`` command which offers some information on instances. However, ``:info`` only works on single words, which means that ``:info Either String Int`` will return results for ``Either``, ``String``, and ``Int`` separately. This makes it a partial replacement at best for the ``:instances`` command.
 
+Examples
+--------
+
+
+.. code-block: none
+
+  >> :instances Sum f f
+  Functor f => Functor (Sum f f)
+  Foldable f => Foldable (Sum f f) ..
+
+Basic instance lookup:
+
+.. code-block: none
+  >> :instances [_]
+
+  Show _ => Show [_]
+  Eq _ => Eq [_]
+  ....
+
+Often when dealing with monad stacks, it can be useful to figure out exactly which instances are available for a given type:
+
+.. code-block: none
+  >> :instances StateT _s (ReaderT _r IO)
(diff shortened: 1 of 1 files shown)


ElvishJerricco:
This is awesome. It'd also be great if you could list the instances of a given class.

Icelandjack:
What does that mean for `Eq` @ElvishJerricco, will it list

```haskell
Eq ()
Eq Int
Eq ((), ())
Eq ((), Int)
Eq (Int, ())
Eq (Int, Int)
Eq ((), ((), ()))
Eq ((), ((), Int)) ..
Eq ((), ((), ((), ..
```

ElvishJerricco:
@Icelandjack Would that just show this?

```
Eq ()
Eq Int
(Eq a, Eq b) => Eq (a, b)
(Eq a, Eq b, Eq c) => Eq (a, b, c)
(Eq a, Eq b, Eq c, Eq d) => Eq (a, b, c, d)
...
```

Anyway, I just realized you can already get both of these behaviors with `:i DataType` and `:i ClassName`.

xldenis:
Currently the `:info` command has some heuristics to decide which instances to show. I think in general this is a problem because a class is likely to have many more instances than a type does.

RyanGlScott:
Thank you for opening a GHC proposal for this. After reading the proposal, I'm left with a number of questions:

1. What happens if you give a type class an an argument to `:instances`? (e.g., `:instances Eq`.) After reading this proposal, I get the general impression that `:instances ty` only enumerates instances where `ty` appears as an argument to some other type class, and that `:instances C` isn't intended to list all instances of a class `C`. (So if you typed `:instances Eq`, it would most likely list nothing.) Is this true? If so, it would be helpful to say so explicitly.
2. Am I correct in saying that all arguments given after `:instances` are parsed as a single type? That is to say, `:instances Sum [] []` will never be interpreted to mean "find an instance of a multi-parameter type class with `Sum`, `[]`, and `[]` as arguments"? It would be good to make note of this, especially since this would be a challenge one would have to overcome if one were to extend `:instances` in the future to support this kind of behavior.
3. How are type variables treated? For instance, what happens if you type `:instances a` or `:instances Either a b`? Would these error out due to the type variables not being bound (similarly to how `:kind a` errors out, but `:kind (forall a. a)` does not), and if so, would you require the user to instead type in `:instances (forall a. a)` or `:instances (forall a b. Either a b)`? If not, how would they behave? What is the relationship between type variables and type wildcards? (e.g., `:instances Maybe a` vs. `:instances Maybe _`.)
4. After reading the change specification, I'm unclear on what the exact scope of this proposal is. You list three intriguing ideas, but it's preceded with the phrase "some further work", which makes me unsure if these ideas are really part of the proposal at all, or if these are half-completed thoughts that are intended to be fleshed out at some separate point in the future.

   I'll go ahead and assume that these _are_ part of the proposal for the rest of this comment.
5. In your description of `:noinstance`, you say "`:noinstance` would output a list of the unsatisfied constraints that prevent an instance from existing for a specific type". I'm very unclear on what this means. Can you describe what kind of output, say, `:noinstance Eq (a -> b)`, would produce?
6. The discussion of multi-parameter type classes in the Proposed Change Specification section sounds very unsure of itself, especially since this is supposed to be a specification. Perhaps it would be better to move this to the Costs and Drawbacks section.
7. "Trace Info: Provide the source for every instance a type has." Provide an example of this, please!
8. I didn't really gain much from reading the Effects and Interactions section. The first three paragraphs of that section read like they belong in the Motivation section instead, since they only describe why someone might want this feature. As for the last paragraph, well, I'm unsure if it really has no interaction whatsoever with the language itself, which leads into my next question...
9. ...How does this feature interact with overlapping instances? For example, suppose you had the following instances:

   ```haskell
   instance (Applicative f, Monoid a) => Monoid (f a)
   instance Alternative f => Monoid (f a)
   ```

   Assuming that you can query for something like `:instances Monoid (f a)` (see question 3), will it return both of these? Only one? Will the presence of incoherent instances affect what answer is returned?

xldenis:
Hi @RyanGlScott, thanks for commenting and raising some great points. I've made some changes to the proposal which should clarify things already but I'll respond here as well. 

1. This command isn't intended to list all instances for a class if it's provided as an argument. I don't believe you can actually write instances for classes themselves even with `-XConstraintKinds` (ie `class MyClass (a :: * -> Constraint)`), but I haven't tried either. 
2. Yes, you are correct in thinking that all the arguments following `:instances` are intended to be parsed as a single type, I've clarified that in the proposal. 
3. Currently, the idea was to use wildcards `_` to demarcate those, I don't feel very strongly about this either way. I think in either case we are choosing an 'unfamiliar' syntax when talking about typeclasses. 
    * Forcing `forall` can feel unnatural since we never write them for actual instances, it may also be a foreign concept for beginners using this command (speculation, though). However, it also lets you ask `:instances forall a. Either a a` which you can't do using wildcards.
    * Using wildcards `_` is also unnatural since we don't use them in actual instance heads. It's also slightly less powerful than `forall` since you can't specify that two wildcards should be equal. 
    * There is also a third option, using an implicit `forall`. This would be most similar to the standard instance syntax. 
    * I'm going to add this to the unresolved question section

For 4-9, I've moved it to a future work section as well as expanded the sections and added an example for the traces.

Icelandjack:
Thanks for the feedback @RyanGlScott, very valuable.

What happens if you give a type class an an argument to :instances? (e.g., :instances Eq.) After reading this proposal, I get the general impression that :instances ty only enumerates instances where ty appears as an argument to some other type class, and that :instances C isn't intended to list all instances of a class C. (So if you typed :instances Eq, it would most likely list nothing.) Is this true? If so, it would be helpful to say so explicitly.

1. To make it tractable: My guiding intuition is, listed with `:instances` if listed by `:info` (as argument to type class). It is possible to write instances of constraints (they are rare but exist)¹ and one such exists in the [*constraints* library (`(:=>)`)](https://hackage.haskell.org/package/constraints-0.10.1/docs/Data-Constraint.html#t::-61--62-) and `:info []` lists the following instance (can be assumed to be a 

    ```haskell
    instance Eq a :=> Eq [a] -- Defined in ‘Data.Constraint’
    ```

   so I expect 
    ```haskell
    a
    ```

¹
* from *constraints*
  * [`Class :: Constraint -> Constraint -> Constraint`](https://hackage.haskell.org/package/constraints-0.10.1/docs/Data-Constraint.html#t:Class)
  * [`(:=>) :: Constraint -> Constraint -> Constraint`](https://hackage.haskell.org/package/constraints-0.10.1/docs/Data-Constraint.html#t::-61--62-)
  * [`Deferrable :: Constraint -> Constraint`](https://hackage.haskell.org/package/constraints-0.10.1/docs/Data-Constraint-Deferrable.html)
* [`Static :: Constraint -> Constraint`](https://www.reddit.com/r/haskell/comments/78047z/scrap_all_your_typeclasses_but_one/dorfmry/)
* [`Universally :: (k -> Type) -> (Type -> Constraint) -> Constraint`](https://github.com/andrewthad/primitive-containers/blob/3716da66541625d5a564c9a38b0414fb54568e70/src/Data/Dependent/Map/Class.hs)
* `Cartesian :: (Type -> Constraint) -> Constraint` 
    ```haskell
    class Cartesian cls where
      cartesian :: (cls a, cls b) :- cls (a, b)
    ```

Icelandjack:
Thanks for the feedback @RyanGlScott.

1. To make it tractable. My guiding intuition is, it (top-level or any subtype(?)) is listed by `:instances` if it's listed by `:info` (as an argument to a type class). It is possible to write instances of constraints (they are rare but exist).¹ In this example I expect `:instances` of both `Eq` and `Foo` to list `instance Foo (Eq Int)` because it is given by `:info`:

    ```haskell
    class    Cls (cls :: Constraint)
    instance Cls (Foo Int)

    -- >> :info Int
    -- instance Foo (Eq Int) -- ..
    -- >> :info Eq
    -- instance Foo (Eq Int) -- ..
    ```
    On the other hand, I don't expect any input to `:instances` to give `(&)` as an output.
 
    ```haskell
    class    (cls a, cls' a) => (cls & cls') a
    instance (cls a, cls' a) => (cls & cls') a
    ```
2. Yes as @xldenis says, I don't see value in having them 'float'
3. Ideally I want implicit quantification, where a type variable implies an equality constraint (this query being equal to `forall f f'. f ~ f' => Sum f f'`)

    ```
    >> :instances Sum f f
    Functor f => Functor (Sum f f)
    Foldable f => Foldable (Sum f f) ..
    ```
    and if we query `:instances Monad f => Sum f f` then that rules out `Functor`.

4. I don't care about this feature but I see at least one interesting path:
    * Following the intuition: If it appears in the original `:info` but fails for some reason, we list it:
        ```
        >> :info (,)
        instance (Eq a, Eq b) => Eq (a, b) 
        >> :instances (Int -> Int, Bool -> Bool)
        <no Eq instance>
        >> :noinstances (Int -> Int, Bool -> Bool)
        NO instance for Eq (Int -> Int, Bool -> Bool), because
          NO instance for Eq (Int -> Int)
          NO instance for Eq (Bool -> Bool)
        ```
    * (uninteresting) the user explicitly lists the expected type class `:noinstance Eq (Int -> Int, Bool -> Bool)` giving the same result as the beginning of `:noinstance (Int -> Int, Bool -> Bool)` above

¹ Examples of type classes over `Constraint`
* from *constraints*
  * [`Class :: Constraint -> Constraint -> Constraint`](https://hackage.haskell.org/package/constraints-0.10.1/docs/Data-Constraint.html#t:Class)
  * [`(:=>) :: Constraint -> Constraint -> Constraint`](https://hackage.haskell.org/package/constraints-0.10.1/docs/Data-Constraint.html#t::-61--62-)
  * [`Deferrable :: Constraint -> Constraint`](https://hackage.haskell.org/package/constraints-0.10.1/docs/Data-Constraint-Deferrable.html)
* [`Static :: Constraint -> Constraint`](https://www.reddit.com/r/haskell/comments/78047z/scrap_all_your_typeclasses_but_one/dorfmry/)
* [`Universally :: (k -> Type) -> (Type -> Constraint) -> Constraint`](https://github.com/andrewthad/primitive-containers/blob/3716da66541625d5a564c9a38b0414fb54568e70/src/Data/Dependent/Map/Class.hs)
* `Cartesian :: (Type -> Constraint) -> Constraint` 
    ```haskell
    class Cartesian cls where
      cartesian :: (cls a, cls b) :- cls (a, b)
    ```

xldenis:
I've added a section for further examples as well as clarified what happens when a typeclass is provided as an argument to `:instances`. Are there any further questions or should we proceed to notify the committee?

Ericson2314:
I might prefer writing constraints with wildcards, for sake of working with multi-parameter type classes or aggregate constraints. Basically, something like a prolog REPL

xldenis:
@RyanGlScott I know you were busy with ICFP (congrats on your talk!), do you feel like your concerns have been addressed? 

@Ericson2314 I think we'll just use the same syntax as the `:kind` command which allows writing types like `_ -> Int`, if you want to name your variables then you can bind them explicitly with `forall`.

RyanGlScott:
I haven't reviewed the changes closely, but things look reasonable from a glance. Then again, I'm not part of the committee, so you don't need my permission to continue :)

xldenis:
Well then I think I'll ping @nomeata to bring this before the comittee! 

As a summary of the changes that occurred throughout the discussion:

* Properly split out future work that could be done `:noinstance`, etc..
* Provided more examples
* Clarified behaviour when a class is provided as an argument

I don't believe there are any open objections to this proposal as it stands so it should be ready for committee review?

xldenis:
Bump @nomeata :)

nomeata:
Thanks for the bump, I guess it got lost during ICFP.

yav:
I think I get the basic idea of what we'd like to do here, and it seems useful enough.  However, I think the proposal could use a little more details on how things are supposed to work.  In particular:

1. The proposal suggests using `forall` to name variables in queries, but the meaning is not clear.  For example, would `:instances forall a. Maybe a`  match `instance C (Maybe Int)`?  
   1. If **yes**, then I don't think that `forall` is a very good choice as clearly we don't mean that the instance should work for all types.   It would seem more natural to simply allow variable without a quantifier, e.g., say `:instances Maybe a` instead.
   2. If **no**, then it seems the proposal is missing something potentially useful:  it mentions using wild cards, but those can't express that two types should be the same (yet arbitrary).

2. How exactly does matching a query work?  For the sake of concreteness, consider a query `:instances forall a. (a,b,[Int])`, where I am assuming that `a` is universally quantified (i.e., we don't want to instantiate it), and `b` is existentially quantified (i.e., it is like a named wild-card).   There are at least two dimensions in which we can make choices, when deciding if an instance matches this query:

   1. Where in the instance head can the query appear?  Here are some options, illustrate with examples:

           instance C1 (a,b,[Int])     -- A: only in single parameter classes
           instance C2 x (a,b,[Int])   -- B: as any parameter of a mult-parameter class 
           instance C3 (Maybe (a,b,[Int])) -- C: arbitrarily nested in a parameter

    2. Is it enough to match the instance head, or do we want use the instance context as well?  Here are some examples:

           instance C4 (Int,b,c)              -- Does not match: `a` universal
           instance C5 (a,Int,c)              -- Matches `b`: is existential
           instance Num c  => C6 (a,Int,c)    -- Matches head, but no instance for `Num [Int]`
           instance Num b => C7 (a,[b],c)     -- Matches head, can't simplify context (existential)
           instance Num a => C8 (a,Int,c)     -- Matches head, cant't simplify context (universal)
           instance (b ~ Char) => C9 (a,b,c)  -- context improves instance
           instance (c ~ Char)  => C10 (a,b,c) -- head matches, context unsatisfiable

Sorry for the long post, I hope this will help things more precise.

simonpj:
I rather agree with Iavor.  The "proposed change specification" is not really a specification, in the sense of precisely specifying what should and should not happen.

I think that the intent is something like this.  

* You enter `:instances <type>` to GHCi,
* Here, `<type>` can contain wildcards (underscores), but not variables (like `a`, `b`)
* GHCi should report all the instances that unify with `<type>`, treating each wildcard as a variable that can be unified.

For example
* `:instances Eq (Maybe _)` would report the instance for `Eq (Maybe a)`
* `:instances Eq (Maybe Int)` would report exactly the same
* `:instances Eq _` would report all instances for `Eq`

I don't think there is any need for the `forall` stuff, and I don't understand what is intended by it.

Optional extra: allow named wildcards.  So then `:instances C (_a, Int, _a)` would unify with `instance C (Bool, Int, Bool)` but not with `instance C (Bool, Int, Int)`.

yav:
@xldenis do you think you could make some changes to the proposal to clarify how things should work?

xldenis:
Hi @yav, yes I was travelling over the past week so I didn't have time to address your comments but I just got back today so I'll push a revision of the proposal tomorrow.

proposals/0000-ghci-instances.rst
```diff
@@ -0,0 +1,146 @@
...
+
+Syntax
+~~~~~~
+
+Valid queries to ``instances`` are types, if anything other than a type is provided as an argument, the command will return an error explaining the intended usage. The command interprets everything after ``instances`` as a single type, this means there is no way to specify multiple types to lookup instances of MPTCs. Holes can be used to represent free variables in the argument type, for example: ``Either _ _``. To express relations between multiple holes, named holes can be used, for example:: ``Either _a _a``.
+
+Execution
+~~~~~~~~~
+
+Provided with a valid type, ``instances`` will attempt to match it against the heads of all visible class instances and satisfy all the implied constraints. The output will consist of a formatted listing of all matching and satisfiable instances. Each instance should be simplified as much as possible, meaning that if an instance: ``(c ~ Bool) => C c`` were found it would be presented as ``C Bool``.
```
xldenis: I don't know if I worded this properly, but the intent is to say that given:
```
:instances T
```
We look for all unifiable instances `C T`,  for all visible classes `C`.

yav: Thanks for making the changes.  Just one more clarification:  the part about having to satisfy all implied constraints seems a little fishy to me.   If I write:

    :instances (Maybe _)

I would expect to see the `Eq` instance, correct?   The instance looks like this:

     instance Eq a => Eq (Maybe a)

When deciding if it should display it, GHC will unify the head against my query, which will succeed.  However, we can't simplify the context any further, as we are left with `Eq a` and we don't know how to simplify this.

OTOH, if I'd written `:instances (Maybe (_ -> _))`, then I probably wouldn't want to see that instance as there aren't any instances that match `Eq (_ -> _)`.

So we need some sort of a rule that will tell us which left-over constraints are OK, as in the first example, vs. not OK as in the second.

My suggestion would be that an instance should be displayed if we can unify with the head, and then simplify the instance context to constraints that only involve variables.  I haven't thought very deeply about that, so please take it with a grain of salt.

xldenis: > OTOH, if I'd written :instances (Maybe (_ -> _)), then I probably wouldn't want to see that instance as there aren't any instances that match Eq (_ -> _).

Wouldn't this just be a matter of going through the normal instance resolution process? When we attempt to reduce the context GHC should tell us that there is no instance for `Eq (_ -> _)`?

xldenis: Hmm, yea I see the problem now, I'll add that specific example and reword that sentence. What we'd like is effectively that only constraints on the type-holes remain after simplifying.

Icelandjack:
Wish list item: To generate `DerivingVia` deriving clauses, as a flag to `:instances` or `:via`

```
>> :via Cont (Result i o)
deriving
  (Functor, Applicative, Monad, MonadCont)
via
  Cont (Result i o)
```

deepfire:
A suggestion for the MPTC syntax: use `TypeApplication`'s  `@` to delineate single typeclass parameters.

Notes:

1. The syntax for the currently accepted proposal would then get an extra form, form:
    - current `:instances Type (Foo b) c`, will mean
    - will mean in the extended version: `:instances @(Type (Foo b) c)`
2. The newly-admitted MPTC queries will look like:
    `:instances @(Type1 a (f b)) @[] @(->)`
3. We're riding on/extending the concept of applying types to something that consumes types.

deepfire:
A suggestion for the MPTC syntax: use `TypeApplication`'s  `@` to delineate single typeclass parameters -- while allowing the currently accepted proposal to stand as is.

Notes:

1. The syntax for the currently accepted proposal would then get an extra form:
    - current `:instances Type (Foo b) c`
    - will mean in the extended version: `:instances @(Type (Foo b) c)`
2. The newly-admitted MPTC queries will look like:
    `:instances @(Type1 a (f b)) @[] @(->)`
3. We're riding on/extending the concept of applying types to something that consumes types.

cc @xldenis

xldenis:
I need to rebase my branch on the latest master but I expect that your example will work. 

@deepfire that seems like a pretty reasonable idea but I think we should revisit it once I've finished implementing the current proposal :)

Icelandjack:
```haskell
>> :instances Kleisli (f `Compose` Maybe)
```

PR #167, opened 2018-09-11
Control what sort of entity a deprecated pragma applies to

nineonine:
This proposal would allow to specify which entity (type or data constructor) you want to deprecate.

[rendered](https://github.com/nineonine/ghc-proposals/blob/depr-entities/proposals/0000-deprecated-entities.rst)
[gitlab issue](https://gitlab.haskell.org/ghc/ghc/issues/3427)

### DIFF
--- a/proposals/0000-deprecated-entities.rst
+++ b/proposals/0000-deprecated-entities.rst
@@ -86,8 +86,8 @@ There are currently no known drawbacks to this feature.
 Alternatives
 ------------
 * The usual workaround would be to have a module that imports one but not the other.
-Unfortunately this workaround is limited as it would only work for types, but not for data constructors.
-Another option would be to refactor data constructor names, which is not backward compatible and inefficient.
+  Unfortunately this workaround is limited as it would only work for types, but not for data constructors.
+  Another option would be to refactor data constructor names, which is not backward compatible and inefficient.
 
 * Another alternative would be to try to utilize Haddock annotations. Example: ::
 
--- a/proposals/0000-deprecated-entities.rst
+++ b/proposals/0000-deprecated-entities.rst
@@ -98,8 +98,8 @@ Unresolved Questions
 1) What specifier should be used for data constructors?
 `Initial feature request <https://ghc.haskell.org/trac/ghc/ticket/3427>`_ suggested to use `constructor` but
 using `specifiers from disambiguation in export list proposal <https://ghc.haskell.org/trac/ghc/wiki/Design/TypeNaming>`_
-seems better since it does not require new keywords to be introduced. Another disadvantage of using `constructor` is
-is that it is quite a widely used identifier which is bad for bakward compatibility
+seems better since it does not require new keywords to be introduced. Another disadvantage of using `constructor`
+is that it is quite a widely used identifier which is bad for backward compatibility
 (for example, `hsc2hs uses it <https://github.com/haskell/hsc2hs/blob/master/CrossCodegen.hs#L470>`_ )
 
 Implementation Plan
--- a/proposals/0000-deprecated-entities.rst
+++ b/proposals/0000-deprecated-entities.rst
@@ -11,7 +11,7 @@ Deprecated Entities
 .. sectnum::
 .. contents::
 
-This feature would allow to control what sort of entity a DEPRECATED pragma applies to.
+This feature would allow to control what sort of entity a ``DEPRECATED`` pragma applies to.
 
 Motivation
 ------------
@@ -31,7 +31,7 @@ Proposed Change Specification
 
 Effect and Interactions
 -----------------------
-An example of a use case for this is the following. Given the following module:
+An example of a use case for this is the following. Given the following module: ::
 
     module A where
 
@@ -44,7 +44,7 @@ An example of a use case for this is the following. Given the following module:
     data Baz = Baz
     {-# DEPRECATED data Baz "Don't use data constructor Baz" #-}
 
-When compiling the code which happens to use data constructor or type ``Foo``, we will see the following warnings:
+When compiling the code which happens to use data constructor or type ``Foo``, we will see the following warnings: ::
 
     Main.hs:5:8: warning: [-Wdeprecations (in -Wdefault)]
         In the use of type constructor or class ‘Foo’ (imported from A):
@@ -56,13 +56,13 @@ When compiling the code which happens to use data constructor or type ``Foo``, w
 
 In the same vein, if we use data constructor or type ``Bar``,
 we will be warned only when we use the type, but not constructor,
-since we specified which entity we want to deprecate in above mentioned module ``A``:
+since we specified which entity we want to deprecate in above mentioned module ``A``: ::
 
     Main.hs:8:8: warning: [-Wdeprecations (in -Wdefault)]
         In the use of type constructor or class ‘Bar’ (imported from A):
         Deprecated: "Don't use type Bar"
 
-Same logic applies to ``Baz``, we will be warned only when its data constructor is used:
+Same logic applies to ``Baz``, we will be warned only when its data constructor is used: ::
 
     Main.hs:12:7: warning: [-Wdeprecations (in -Wdefault)]
         In the use of data constructor ‘Baz’ (imported from A):
@@ -71,11 +71,11 @@ Same logic applies to ``Baz``, we will be warned only when its data constructor 
 
 When specifying several entities in one pragma,
 the sort of deprecated entity we've specified will apply to all listed entities.
-This will work - warn when these types are used(but not their constructors):
+This will work - warn when these types are used(but not their constructors): ::
 
     {-# DEPRECATED type Qux, Quux "Don't use this" #-}
 
-This will not work (parse error):
+This will not work (parse error): ::
 
     {-# DEPRECATED type Qux, constructor Quux "Don't use this" #-}
 
--- a/proposals/0000-deprecated-entities.rst
+++ b/proposals/0000-deprecated-entities.rst
@@ -113,6 +113,23 @@ such that the following example is accepted? ::
 
     {-# DEPRECATED type Qux, data Quux "Don't use this" #-}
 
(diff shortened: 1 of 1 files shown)


proposals/0000-deprecated-entities.rst
```diff
@@ -0,0 +1,112 @@
...
+
+When specifying several entities in one pragma,
+the sort of deprecated entity we've specified will apply to all listed entities.
+This will work - warn when these types are used(but not their constructors): ::
+
+    {-# DEPRECATED type Qux, Quux "Don't use this" #-}
+
+This will not work (parse error): ::
+
+    {-# DEPRECATED type Qux, constructor Quux "Don't use this" #-}
```
gridaphobe: Should the `constructor` here be `data` instead?

gridaphobe: I think it would be nicer to have the `type`/`data` qualifier apply only to the immediately-following name, such that the second example would be accepted. But I also haven't had occasion to put more than one name in a `DEPRECATE` pragma yet, so I might be wrong about the common case.

nineonine: That is correct. Thank you! Fixed.

simonpj:
I note that a module deprecation is *positional*.  One says (I think)
```
module M {-# DEPRECATED "blah" #-} where ...
```
One could do the same for data constructors, thus
```
data Baz = Baz {-# DEPRECATED "blah" #-}
               | Boo
               | Bim {-# DEPRECATED "blah" #-}
                        Int
```
or in GADT syntax
```
data Baz where
   Baz :: Int -> Baz
   Boo :: Baz
   Bim :: Baz
   {-# DEPRECATED Bim, Baz "blah" #-}
```
This positional story works well when the deprecation is attached to the *definition* of the thing.  If you want to import something, deprecate it, and re-export it, it would not work so well.  But (I think) we can't do that anyway today.

RyanGlScott:
After reading this proposal, it seems like this is limited in scope to disambiguating prefix type constructors from prefix data constructors. But what about the following scenario?

```haskell
module A where

type a <+> b = ...

(<+>) :: Int -> Int -> Int
a <+> b = ...

{-# DEPRECATED (<+>) "Don't use (<+>)" #-}
```

It's unclear which `(<+>)` I'm referring to, so I want to pick the value-level `(<+>)` in particular. But how do I say so with this proposal? Certainly not with the `type` specifier. That would suggest that `data` is the specifier to use, but that also feels a little awkward, since `(<+>)` isn't a data constructor.

The reason I'm picking this nit is because there is a separate, previously accepted proposal ([Require namespacing fixity declarations for type names and `WARNING`/`DEPRECATED` pragmas](https://github.com/ghc-proposals/ghc-proposals/blob/30eed9f8ea8d20f8401c6e80158402fcaae585d1/proposals/0008-type-infix.rst)) that also adds specifiers in a similar fashion to this proposal. However, it proposes adding specifiers named `type` and _`value`_ (instead of `type` and `data`), since `value` is a more inclusive description that can apply to more entities than `data`.

Sorry for continuing to bikeshed about this name even more, but at the very least I'd like this proposal to be consistent with the other proposal (even if we end up changing the other proposal to use `data` over `value`).

nineonine:
Thank you for all your comments!
@RyanGlScott this is a great example! `value` keyword specifier definitely sounds more inclusive than `data`. I tried to think about an edge case where we would need to distinguish between data constructor and a function (with something like `PatternSynonyms` perhaps) but wasn't able to - it looks like it is restricted on syntax level so it is probably safe to have 1 disambiguating specifier for both these namespaces. Unless someone has something else in mind, I am going to update the proposal accordingly.

nineonine:
pinging @nomeata - I think this is ready to be reviewed.

nineonine:
@nomeata  thanks for the comments.

I do agree that `pattern` looks somewhat strange in that case. But as long as everyone agrees on it, I will gladly change it.

> When specifying several entities in one pragma, the sort of deprecated entity we've specified will apply to all listed entities. This will work - warn when these types are used(but not their constructors):
 
This is coming from the implementation that I already have in mind. I just added that bit to the proposal trying to cover all the possible corner cases. Frankly speaking, I am not sure whether this feature is even widely used (specifying several entities in one pragma). So this is just an implementation side-effect, so to speak, which I included in the proposal for the sake of completeness.

nomeata:
>  So this is just an implementation side-effect…

Even if that was the easiest way to implement it, and may few people use this feature, we should still strive for consistency with the rest of the language. And if in export lists (where we already have such qualifiers) they bind more tightly than commas, then I think they should do that here as well.

nineonine:
> we should still strive for consistency with the rest of the language

@nomeata for sure, this is a good point and I wholeheartedly agree! 
Updating the proposal accordingly.

bgamari:
What is the status of this?

bravit:
@nineonine, could you please mention the idea of positional DEPRECATED pragmas as described [here](https://github.com/ghc-proposals/ghc-proposals/pull/167#issuecomment-420177163) by Simon in the list of alternatives?

bravit:
Before I am ready to make a recommendation, are there any thoughts (pro or contra) about moving to positional DEPRECATED pragmas instead? In fact, I think that this approach is better as it enables more fine-grained control over deprecated entities and we don't have to use `pattern` and other words. By the way, what about deprecating instances (see also this [feature request](https://ghc.haskell.org/trac/ghc/ticket/12014) and this [note on deprecation mechanisms](https://ghc.haskell.org/trac/ghc/wiki/Design/DeprecationMechanisms))?

nineonine:
Added positional deprecation idea.
I think this idea would work equally well and I cannot think of any bad consequences of going that way right now.
Deprecating class instances sounds like a great idea too, I added this to "unresolved questions". I wonder what others think about that..

bravit:
@nineonine will you be able to rewrite/reimplement the proposal if the Committee decides in favor of positional pragmas?

nineonine:
@bravit sounds good.

simonpj:
Hang on.  No one has addressed [Ryan's point](https://github.com/ghc-proposals/ghc-proposals/pull/167#issuecomment-420988929), where he points to the [already accepted proposal](https://github.com/ghc-proposals/ghc-proposals/blob/30eed9f8ea8d20f8401c6e80158402fcaae585d1/proposals/0008-type-infix.rst) about namespacing for `infix` and `WARNING` pragmas.

Isn't namespacing for `DEPRECATED` pragmas _precisely_ the same problem?  Can't we simply adopt precisely the same solution?

We don't want to add inconsistent solutions!

Arguably, if we like `value` in the infix/WARNING proposal, we might want to shift from `pattern` to `value` in export lists for pattern synonyms.

bravit:
@simonpj, in fact, it was addressed. We have three options here: to choose either "data", "value", or "pattern". Richard Eisenberg wrote on the [committee's mailing list](https://mail.haskell.org/pipermail/ghc-steering-committee/2019-February/000905.html): 

> Though I understand the reasons against it, I'm an unabashed supporter of using the word "data" to supplant "pattern". My principal argument is that `data` can be used freely in the syntax, given that it's a keyword that has current meaning only as the first lexeme in a top-level declaration. (Specifically, I pine for a future where types and terms mix. We can then use `type` and `data` in the middle of expressions/types to denote namespaces.) It also works nicely to mean "data constructor". I agree that it doesn't work as well with functions or the potential confusion around "data Bool = True | False" (though, in that last example, we can pretend hard that the `data` applies only to the bits after the `=`). The problem with `value` is that we would have `{-# DEPRECATED value, foo #-}` mean something very different than `{-# DEPRECATED value foo #-}`. This isn't a deal-breaker (punctuation normally is important in programming), but it's unfortunate.

The same argument applies to infix/WARNING proposal, so I think that we could change it either (anyway, it is not implemented yet afaik). I personally prefer "data" over "value", and "value" over "pattern", but I'm not ready to die for that.

simonpj:
OK -- so at let's agree, at least, that whatever is decided here applies equally to the infix/WARNING proposal.  And I would argue that it ought to apply equally to export and import lists, though that would need a careful migration path.

OK, so do we really want to write
```
infix 3 pattern `foo`

{-# DEPRECATED pattern foo "Do not use foo" #-}
foo :: Int -> Int -> Int
foo  x y = y - x
```
Obviously `foo` isn't a pattern! 

And `data` would be pretty strange  there too.  

Personally I'm fine with `value`.  It's true that
```
infixl 3 value value
```
would be pretty strange. But I don't expect it to happen in 99.999% of programs, and I care about the 99% case a lot.

nineonine:
Updated the proposal to use `value` for value level things.

nineonine:
I was thinking today about the implementation and it led me to the following question - what do we want to happen when wrong specifier was used?
```haskell
{-# DEPRECATED type ack "blah" #-}
ack = ()
```
Currently I see 3 options:
- warn user about misused disambiguator and ignore malformed deprecation
- ignore disambiguator and treat it as unqualified case
- completely ignore deprecation and do nothing

nomeata:
@nineonine did you see https://github.com/ghc-proposals/ghc-proposals/pull/214? That is a more ambitious attempt to fix all that, which likely subsumes this proposal. Please comment there as well, if you find it lacking (or to voice your support)!

nineonine:
@nomeata no I didn't! thanks

bravit:
@nineonine it turned out that the proposal on namespacing in the DEPRECATED pragma (among other things) [was already accepted](https://github.com/ghc-proposals/ghc-proposals/blob/30eed9f8ea8d20f8401c6e80158402fcaae585d1/proposals/0008-type-infix.rst) more than a year ago. That's why there is no sense in accepting this proposal now. That proposal and your proposal are now subsumed by #214. Hopefully, it will be accepted and implemented once, that will close the ticket you mentioned.

nomeata:
Right, as @bravit said: Your proposal pointed out the need for a larger reorganization around namespaces, and as such it was very helpful, thanks for that! I hope that once #214 or something similar is accepted, your itch will be sufficiently scratched as well.

PR #168, opened 2018-09-17
make rebindable fail work with overloaded strings

shayne-fletcher-da:
We propose that string literals generated by GHC and passed to user defined functions should have `fromString` applied. The only instance of such GHC generated strings we are aware of is via `fail`, so we propose to fix that case.

[Rendered](https://github.com/shayne-fletcher-da/ghc-proposals/blob/fail-rebindable-with-overloaded-strings/proposals/0000-fail-rebindable-with-overloaded-strings.rst)


nomeata:
I am sympathetic. I would be least surprised if the implicit string in the generated code would behave like a literal.

ndmitchell:
To state my opinion, I reviewed the proposal with @shayne-fletcher-da before submission, it feels like a bug fix to me, and I'm in favour :)

Ping @vincenthz, who writes the foundation library. I imagine this might be useful for him, or he might have opinions on it.

vincenthz:
I looked at this a bit diagonally, but that seems just a bug related to RebindableSyntax (similar to some of the recent one linked to the far reaching consequences). On the face value, that should keep RedindableSyntax working as expected, so :+1: from me.

ndmitchell:
It seems that there are no objections, despite posting to Twitter no real comments, so can we get a steering committee vote? In the meantime there is now a patch, which will go through the standard Phab workflow.

ndmitchell:
Is anything more required to trigger the committee review?

nomeata:
The review has been triggered by my mail to the list, and is now waiting for people to speak up. If nobody speaks up I'll accept this. I can probably do that already, but breakfast first.

nomeata:
Done, and thanks for the ping.

ndmitchell:
Thanks! Phab submission to follow soon

PR #123, opened 2018-04-18
Proposal to change the load order of `.ghci` config files

halhenke:
This is a proposal regarding the load order of ghci files & an option to change the way they evaluated such that local changes override global ones.

[Rendered](https://github.com/halhenke/ghc-proposals/blob/ghci-load-order/proposals/0000-ghci-configuration-load-order.rst)

PR #42, opened 2017-01-27
Allow signatures on pattern synonym constructors

treeowl:
* Allow type signatures on the construction functions of
bidirectional pattern synonyms.

* Change the way construction functions are typechecked when
they lack signatures.

[Rendered view](https://github.com/treeowl/ghc-proposals/blob/559df2865db8e4427b1f24d3100e59d61cb60e54/proposals/0000-bidir-constr-sigs.rst)

PR #170, opened 2018-09-20
Unrestricted OverloadedLabels

howtonotwin:
[Rendered proposal](https://github.com/howtonotwin/ghc-proposals/blob/uppercase-overloadedlabels/proposals/0000-unrestricted-overloadedlabels.rst)

[Original feature request](https://ghc.haskell.org/trac/ghc/ticket/11671)


treeowl:
This looks like an obviously good idea to me.

cartazio:
agreed, especially given the current desugaring

adamgundry:
The lowercase-only restriction originated because of the historical relationship between overloaded labels and the work on overloaded record fields, but now they have evolved somewhat independently, I think it makes sense to permit uppercase overloaded labels.

I have wondered about whether we should have a different class for interpreting uppercase labels, which might be useful in some cases for making the field/constructor distinction. But it seems that would get in the way of some use cases that have nothing to do with fields/constructors, and in any case we could recover the distinction if we had a way to deconstruct and analyse type-level expressions of kind `Symbol`. Thus I support the simplicity of the current proposal.

howtonotwin:
If there are no changes required, I guess this can be submitted to the committee. Paging @nomeata.

simonpj:
Looks sensible to me.

gridaphobe:
You could go even further and accept arbitrary strings like `#"hello there"`, analogous to lisp's treatment of symbols. I don't have a strong desire to write a label like this, but it is another type of label that the `#` syntax disallows.

ndmitchell:
When I wanted a more powerful label syntax I made it so #_Foo had a fromLabel instance that did some symbol manipulation to strip at most 1 leading underscore. I was only using the symbol at the value level though.

gridaphobe:
If we were to allow arbitrary strings in labels, we'd need some sort of delimiter. Lisp uses `|` for arbitrary symbols, e.g. `|2 + 2 = 4|` (whitespace is allowed). My example used double quotes as the delimiter, since OverloadedLabels is supposed to produce type-level Symbols. Once you have the delimiter though, it doesn't really make sense to restrict the Symbols you can produce.

I don't want to derail this proposal though. The original idea of allowing upper-case labels is very sensible.

goldfirere:
I think I would allow arbitrary alphanumeric strings. I could also see extending to `#" ... "` syntax, allowing anything that lexes as a string immediately following the `#` (without intervening whitespace). I agree that we shouldn't do arbitrary symbolic non-delimited strings.

yav:
If we were to allow arbitrary strings, would `#3` be mapped the the `Symbol` `"3"`, or to the `Nat` `3`?

howtonotwin:
@yav: `#<label>` always means `fromLabel @"<label>"`. If `#3` were to mean anything (see above), it would mean `fromLabel @"3"` with a `Symbol`, because that's the only thing `fromLabel` can handle. Someone could argue that it should desugar to something else with a `Nat`, but that's getting out there.

@simonmar + @goldfirere: Allowing alphanumeric strings (including keywords and excluding whitespace) without delimiters and anything within `#".."` has my support. If there are no further comments on that, I'll edit it in.

goldfirere:
A few clarifying questions:

1. The specification contains an example `#199.4`, but `199.4` is not an alphanumeric string. Is this a typo, or are you proposing to accept dots, too?

2. Does `#"Foo"` mean the same thing as `#Foo`? If so, this should be clarified in the proposal.

3. What about underscores? I assume you mean to include these, but the specification, as written, doesn't.

howtonotwin:
I guess when I saw "alphanumeric string" I misinterpreted that as "a nonempty string of the chars that make up variables and numbers." Going by the [report](https://www.haskell.org/onlinereport/haskell2010/haskellch10.html), alphanumeric labels would be

```none
unquotedLabel := '#' {ascSmall | uniSmall | large | digit}
```

While I was thinking

```none
unquotedLabel := '#' {small | large | digit | '\'' | '.'}
```

The latter lets some *interesting* things through, like `#'.'`, `#'abc'`. This could potentially be confusing, though we already allow `#f'a'`. Incidentally, "alphanumeric" is too restrictive anyway, not allowing some existing labels.

We can do what I was thinking; as above
```none
unquotedLabel := '#' {small | large | digit | '\'' | '.'}
```
Or we may modify it to avoid the potentially confusing `#'<str>'` labels
```none
unquotedLabel := '#' (small | large | digit | '.') {small | large | digit | '\'' | '.'}
```

howtonotwin:
Since bare labels can't be strictly alphanumeric, I've gone ahead and used my original idea from above.

goldfirere:
Thanks for drawing that out.

One last question, then: does `#Foo` mean the same as `#"Foo"`? There's nothing in the proposal to suggest that they do equal, but someone might expect them too, so I'm just checking.

howtonotwin:
`#Foo` being the same as `#"Foo"` is in table of examples. I'll also put in an explicit affirmation that, if a bare label `#<str>` exists, then `#"<str>"` is the same thing.

goldfirere:
Ah, so it is in the table. Sorry for the noise!

adamgundry:
This proposal has been implemented as https://gitlab.haskell.org/ghc/ghc/-/merge_requests/8966 and so should be in GHC 9.6. :tada:

adamgundry:
This proposal as written stops `#.` being accepted as an operator, which is a significant breaking change (see https://gitlab.haskell.org/ghc/ghc/-/issues/22821). I think we should not permit `.` as a character in labels (except in a string, of course). Any opinions? This is moderately urgent as it needs to be resolved for GHC 9.6.

nomeata:
The original issue only asked for initial upper case letters. Maybe we generalized it too much by allowing  number literals? I wouldn't necessarily expect `#1.2e3` to work. And wonder how much confusion `#01.23` vs `#1.230` vs `#123e-3` vs `#1.2_3`, i.e. the dependency on the actual format of the same number, causes.

Maybe we should cut this back to safer ground (alphanumeric labels) before the release?

int-index:
We should _at least_ revert the parts of the proposal that cannot be addressed by whitespace. It's fine to steal `f #42` because the user can write `f # 42` instead. But stealing the `#.` operator is off limits.

simonpj:
> Maybe we should cut this back to safer ground (alphanumeric labels) before the release?

Let's do that.  Numbers and letters only.  No operator symbols.  That still includes `#1e2` but maybe that's ok.

We just need someone to make a PR to modify the propposal, and we can accept it quickly

phadej:
> Let's do that. Numbers and letters only. No operator symbols. That still includes #1e2 but maybe that's ok.

The identifiers alone would be enough too.

Writing `#"1e2"` (which is also part of this proposal) would be still allowed.

treeowl:
Identifiers and explicitly quoted things seem safer.

adamgundry:
If we rule out a leading digit, then `f #123` is still accepted (as the operator application `(#) f 123`) while `f #Foo` will be the function application `f (#Foo)`. So I think it might actually be clearer to steal `#123` as a label as well, and force the use of whitespace after `#` if it is intended to be an operator? Or is `-Woperator-whitespace` enough?

nomeata:
Maybe `-Woperator-whitespace-ext-conflict` should be extended to `#`, too?

sheaf:
> > Maybe we should cut this back to safer ground (alphanumeric labels) before the release?
> 
> Let's do that. Numbers and letters only. No operator symbols. 

What about underscores?

adamgundry:
Underscores are already valid (both in the status quo and this proposal), so they need to remain so.

Apostrophes were previously allowed but not as the first character, e.g. `#a'` was a label but `#'` was not. This proposal as currently written changes things so that `#'` will be allowed. I don't have a strong opinion on whether that's useful.

adamgundry:
I've opened #573 with a proposed fix.

phadej:
FWIW, `foo #bar` ought to be reported by `-Woperator-whitespace-ext-conflict`. I'd consider it a bug for *not being reported today*.

Consider:

```haskell
Prelude> :set -Woperator-whitespace
Prelude> foo #bar

<interactive>:9:5: warning: [-Woperator-whitespace]
    The prefix use of a ‘#’ might be repurposed as special syntax
      by a future language extension.
    Suggested fix: Add whitespace around ‘#’.
```

`-Woperator-whitespace` error is incorrect, *there is existing language extension*.

PR #173, opened 2018-09-26
The dot type operator

int-index:
The `TypeOperators` language extension allows using symbols in type names, similarly to operators in terms. However, some operator names that are valid in terms are not valid in types, including the dot operator `(.)` commonly used to denote function composition. We propose to lift this restriction.

[Rendered](https://github.com/int-index/ghc-proposals/blob/dot-type-operator/proposals/0000-dot-type-operator.rst)

### DIFF
--- a/proposals/0000-dot-type-operator.rst
+++ b/proposals/0000-dot-type-operator.rst
@@ -1,0 +1,92 @@
+The dot type operator
+=====================
+
+.. proposal-number::
+.. trac-ticket::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+.. sectnum::
+.. contents::
+
+The ``TypeOperators`` language extension allows using symbols in type names,
+similarly to operators in terms. However, some operator names that are valid in
+terms are not valid in types, including the dot operator ``(.)`` commonly used
+to denote function composition. We propose to lift this restriction.
+
+Motivation
+------------
+
+At the moment, we cannot define function composition at type level::
+
+  ghci> type (f . g) x = f (g x)
+
+  <interactive>:1:9: error:
+      Illegal symbol '.' in type
+      Perhaps you intended to use RankNTypes or a similar language
+      extension to enable explicit-forall syntax: forall <tvs>. <type>
+
+There are two issues with this definition:
+
+1. The dot is not a valid type operator.
+2. The type synonyms and type families cannot be partially applied, severely
+   limiting the usability of type-level function composition.
+
+In this proposal, we seek to address the first issue. The second issue is out
+of scope, but can be addressed with defunctionalization today or by allowing
+unsaturated use type functions in the future.
+
+With a small modification to the Haskell grammar, the following code is
+accepted::
+
+  ghci> :set -XTypeOperators
+  ghci> type (f . g) x = f (g x)
+  ghci> Nothing :: (Maybe . Either Int) Bool
+  Nothing
+  ghci> Just (Left 5) :: (Maybe . Either Int) Bool
+  Just (Left 5)
+
+We can achieve this by changing a single production in ``Parser.y``, namely
+``tyapp``::
+
+  -   | tyvarop          { sL1 $1 $ TyElOpr (unLoc $1) }
+  +   | '`' tyvarid '`'  { sLL $1 $> $ TyElOpr (unLoc $2) }
+  +   | '.'              { sL1 $1 $ TyElOpr (mkUnqual tcClsName (fsLit ".")) }
+
+Proposed Change Specification
+-----------------------------
+
+Allow the use of the ``(.)`` type operator in all places where other type
+operators are legal.
+
+Effect and Interactions
+-----------------------
+
+Clasess, data types, data families, type synonyms, and type families can now be
+given the name ``(.)``.
+
+There are no known conflicts with other language features despite the use of
+the dot in type-level constructs like ``forall <tvs>. <type>``. The parser has
+enough information to disambiguate in all cases.
+
+Costs and Drawbacks
+-------------------
+
+None.
+
+Alternatives
+------------
+
+* Continue to disallow the use of the dot operator in types.
+* Find a different use for the dot syntax inconsistent with terms.
+
+Unresolved Questions
+--------------------
+
+None.
+
+Implementation Plan
+-------------------
+
+I (Vladislav Zavialov) will implement this change.
+
--- a/proposals/0000-dot-type-operator.rst
+++ b/proposals/0000-dot-type-operator.rst
@@ -53,6 +53,27 @@ We can achieve this by changing a single production in ``Parser.y``, namely
   +   | '`' tyvarid '`'  { sLL $1 $> $ TyElOpr (unLoc $2) }
   +   | '.'              { sL1 $1 $ TyElOpr (mkUnqual tcClsName (fsLit ".")) }
 
+Note that the actual implementation will be more complex to provide decent
+error messages, keep `annotations
+<https://ghc.haskell.org/trac/ghc/wiki/ApiAnnotations>`_ working, etc.
+
+We can use whitespace to disambiguate between uses of the dot as a type
+operator and as a part of a qualified name. The situation is the same as in
+terms, demonstrated by this table (courtesy of `@mstksg
+<https://github.com/mstksg>`_)::
+
+  foo . bar -- term level: (.) foo bar
+            -- type level: (.) foo bar (currently disallowed)
+
+  foo.bar   -- term level: (.) foo bar
+            -- type level: (.) foo bar (currently disallowed)
+
+  Foo . Bar -- term level: (.) Foo Bar
+            -- type level: (.) Foo Bar (currently disallowed)
+
+  Foo.Bar   -- term level: Bar imported from module Foo
+            -- type level: Bar imported from module Foo
+
 Proposed Change Specification
 -----------------------------
 
--- a/proposals/0000-dot-type-operator.rst
+++ b/proposals/0000-dot-type-operator.rst
@@ -5,7 +5,7 @@ The dot type operator
 .. trac-ticket::
 .. implemented::
 .. highlight:: haskell
-.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/173>`_.
 .. sectnum::
 .. contents::
 
--- a/proposals/0000-dot-type-operator.rst
+++ b/proposals/0000-dot-type-operator.rst
@@ -74,6 +74,21 @@ terms, demonstrated by this table (courtesy of `@mstksg
   Foo.Bar   -- term level: Bar imported from module Foo
             -- type level: Bar imported from module Foo
 
+In the ``singletons`` library, there is a `special case
+<https://github.com/goldfirere/singletons/blob/a9db6ff634d00a11a74595187e4ed935715f6626/src/Data/Singletons/Names.hs#L355-L361>`_
+for promotion of ``(.)`` and ``(!)``::
+
+    Note [Special cases for (.) and (!)]
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    Almost every infix value name can be promoted trivially. For example, (+)
+    works both at the value- and type-level. The two exceptions to this rule
+    are (.) and (!), which we promote to the special type names (:.) and (:!),
+    respectively.  This is necessary since one cannot define or apply (.) or
+    (!) at the type level -- they simply won't parse. Bummer.
+
+The issue with ``(!)`` has been fixed in the upcoming GHC 8.8, but the problem
+of ``(.)`` remains.
+
 Proposed Change Specification
 -----------------------------
 


phadej:
```
Perhaps you intended to use RankNTypes or a similar language
extension to enable explicit-forall syntax: forall <tvs>. <type>
```

could be changed (independently of whether this proposal is accepter or not) to:

```
Perhaps you intended to use ScopedTypeVariables, RankNTypes
or a similar language extension to enable explicit-forall syntax:
forall <tvs>. <type>
```

phadej:
In general I'm concerned about error reporting. In proposals example:

```haskell
type (f . g) x = f (g x)
```

GHC should  suggest `TypeOperators`. I guess it will, if parser is changed?

int-index:
> GHC should suggest `TypeOperators`. I guess it will, if parser is changed?

@phadej It suggests `TypeOperators` in my prototype:

```
Prelude> type (f . g) x = f (g x)

<interactive>:1:9: error:
    Illegal declaration of a type or class operator ‘.’
      Use TypeOperators to declare operators in type and declarations
```

It's the same error reporting codepath as for any other type operator:

```
Prelude> type (f + g) x = f (g x)

<interactive>:2:9: error:
    Illegal declaration of a type or class operator ‘+’
      Use TypeOperators to declare operators in type and declarations
```

simonpj:
I'm OK with the motivation, but unclear about what the difference
in behaviour is from today.

With your change, is there any conflict with the `forall a b. blah`
notation.  (I think not, but I'd like to be sure.)  And if not,
I wonder why `.` is not already allowed as a type-level operator;
does anyone know?


Currently we have
```
tyapp :: { Located TyEl }
        : atype                         { sL1 $1 $ TyElOpd (unLoc $1) }
        | qtyconop                      { sL1 $1 $ TyElOpr (unLoc $1) }
        | tyvarop                       { sL1 $1 $ TyElOpr (unLoc $1) }
        ...

tyvarop :: { Located RdrName }
tyvarop : '`' tyvarid '`'       {% ams (sLL $1 $> (unLoc $2))
                                       [mj AnnBackquote $1,mj AnnVal $2
                                       ,mj AnnBackquote $3] }
        | '.'                   {% hintExplicitForall' (getLoc $1) }
```
You propose to change the former to
```
tyapp :: { Located TyEl }
        : atype            { sL1 $1 $ TyElOpd (unLoc $1) }
        | qtyconop         { sL1 $1 $ TyElOpr (unLoc $1) }
        | '`' tyvarid '`'  { sLL $1 $> $ TyElOpr (unLoc $2) }
        | '.'              { sL1 $1 $ TyElOpr (mkUnqual tcClsName (fsLit ".")) }
        ...
```
Grammatically that makes no difference, does it? You've just inlined `tyvarop`.
And `tyvarop` is used nowhere else!  So what is gained by inlining it.

You are also changing something about annotations on the `tyvarid` production;
is that deliberate?

You are removing `hintExplicitForall'`; I guess that is the main change?
But surely you want to continue with that hint if `TypeOperators` is off?

Finally, you are using `mkUnqual` stuff instead of `getLoc $`.  Is that
essential to the change and what effect does it have?

int-index:
@simonpj Regarding inlining `tyvarop`, a change to annotations, and removing `hintExplicitForall'` — this was done to achieve a shorter proposal text while still conveying the general idea.

In reality I think the patch will be more complex than that, especially to achieve good error messages. No changes to annotations are planned.

As to using `mkUnqual`, isn't it the right way to construct a `RdrName`? Anyway, that's more of an implementation detail, too. 

I've shown the change to the grammar just to give a rough idea how this change could be achieved, and to prove that there is no conflict with the `forall <tvs>. <type>` syntax.

The actual change specification is this:

> Allow the use of the `(.)` type operator in all places where other type operators are legal.

goldfirere:
I agree that making `.` a type operator is unambiguous to the parser. However, it may be hard to read for us poor humans. To wit: `forall f g . f . g -> ...` It's jarring for the two dots to have wildly different interpretations.

That said, I support the proposal.

int-index:
> it may be hard to read for us poor humans. To wit: `forall f g . f . g -> ...`

Humans can take care of other humans by adding parentheses and using different whitespace conventions:

```
forall f g. (f . g) -> ...
```

There are plenty of opportunities for syntactic obfuscation and this one isn't really special.

AntC2:
-1

Allowing bare `.` as an ordinary operator at term level is a mistake we all have adapted to live with so we hardly notice. But `.` is already allowed in too many non-operator contexts: module names, `forall`.

I still hanker for the `.` suffix syntax in Simon's very original TDNR proposal "The power of the dot". (I mean the syntax, we've moved on from that specific semantics.) Allowing `.` as a type operator will put that even further out of reach.

int-index:
> Allowing bare `.` as an ordinary operator at term level is a mistake we all have adapted to live with so we hardly notice.

@AntC2 One of my goals is having a consistent language for terms and types. So, I would be happy with one of two options:

1. allowing `(.)` as a type operator
2. disallowing `(.)` in terms

I figured that option 2 is not going to happen because it would break too much code, so I proposed option 1.

> Functions at type level are spelled upper-case -- i.e. type families: that was possibly another mistake. 

Unfortunately, it's another source of inconsistency between terms and types. At the same time, it's consistent with type synonyms being upper-case. I think standard Haskell puts us into an awkward position here.

> the syntax should at least insist that `.` as operator appears with whitespace or non-alphas both sides.

This diagnostic does not sound specific to terms or types. I guess it could be added as an opt-in warning, and adding a warning doesn't even require a proposal. 

> `y :: (A.B.C.F.D.E.G) Int` That's an ambiguous parse(?)

That's an unambiguous parse, we have `G Int` where `G` comes from the module `A.B.C.F.D.E`.

mstksg:
re: the last two posts, It should be worth noting that potential confusion over things like

    (A.B.C.F.D.E.G) "hello"

is something we already have to deal with at the term level as well, and so does not actually introduce any new gotcha's or sources of confusion that was not already there in the language.

I don't mean to dismiss the potential confusion to newcomers, but rather note that this isn't some new novel confusion source brought up by this specific proposal.

AntC2:
> One of my goals is having a consistent language for terms and types. 
> 

I support the ideal, but that horse has already bolted ;-). Specifically for `.` it is _not_ like other symbols that can appear in operators: in types it appears following `forall`; anywhere it can appear in module names.

> 
> > Functions at type level are spelled upper-case -- i.e. type families: that was possibly another mistake.
> 
> Unfortunately, it's another source of inconsistency between terms and types. At the same time, it's consistent with type synonyms being upper-case.

Kinda. Now that we have type families, we like to think of type synonyms as being type families with just a single instance. But that's a bogus analogy: type synonyms are always injective/there's a structural correspondence between the two sides of the `=`. That's never the case for type families, because there's the potential for multiple instances, even if the type family is alleged to be injective. And I won't even get into overlapping/closed type families.

So now that we have pattern synonyms, which are upper-case and for stronger reasons, I think we can say type synonyms are like patterns at the type level.

> > `y :: (A.B.C.F.D.E.G) Int` That's an ambiguous parse(?)
> 
> That's an unambiguous parse, we have `G Int` where `G` comes from the module `A.B.C.F.D.E`.

So you're following a 'maximum munch' rule(?) I think the proposal should say that.

It's what I expected you'd say. Now justify it as "consistent language for terms and types".

To answer @mstksg: no we (or at least GHC) _don't_ have to deal with potential confusion at term level. The term I gave is unambiguous as far as GHC is concerned, because of the lower-case functions; and the all-upper-case must be a whole module name. Then inserting whitespace around the `.` doesn't change its meaning -- I mean: around the `.` that's the operator. It's common to compose Lenses like that with no spaces around the `.`. So I suppose Lensaholics get used to it.

Whereas for this proposal at the type level, putting whitespace around `.` does change its meaning.

mstksg:
@AntC2 the analogous situation appears in the term level and type level.  Adding a space around `.` changes the meaning at the term level as well:

    Foo . Bar

at the term level is the operator `(.)` applied to the data constructors `Foo` and `Bar`, and

    Foo.Bar

at the term level is the data constructor `Bar` imported qualified from module `Foo`.  Putting whitespace around the `.` changes the meaning at the term level.

mstksg:
Just as a reference, so we can have it all to look at while we discuss this:

    foo . bar
    -- term level: (.) foo bar
    -- type level: (.) foo bar (currently disallowed)

    foo.bar
    -- term level: (.) foo bar
    -- type level: (.) foo bar (currently disallowed)

    Foo . Bar
    -- term level: (.) Foo Bar
    -- type level: (.) Foo Bar (currently disallowed)

    Foo.Bar
    -- term level: Bar imported from module Foo
    -- type level: Bar imported from module Foo

Note that the whitespace matters for capital identifiers at both the term and type levels.

int-index:
> So now that we have pattern synonyms, which are upper-case and for stronger reasons, I think we can say type synonyms are like patterns at the type level.

@AntC2 That's an interesting point of view. I think I can agree with this line of reasoning, `type` could indeed be a type-level version of bidirectional pattern synonyms, especially given the fact that we have `{-# LANGUAGE TypeSynonymInstances #-}`. Insightful comment, thanks!

> So you're following a 'maximum munch' rule(?) I think the proposal should say that. It's what I expected you'd say. Now justify it as "consistent language for terms and types".

I added the table by @mstksg that clarifies the situation to the proposal. The parsing rules are consistent because they are based on the distinction between uppercase/lowercase names, not between functions/non-functions.

We happen to use type families to model type-level functions today, but I'm hoping this will change in the future, with regular term-level functions becoming usable on type-level. This further motivates this proposal: when we are capable of promoting functions to the type-level, not being able to use the `(.)` operator just because of its name would be an unnecessary restriction.

Ericson2314:
@AntC2 I would support an extension to disallow `.` as an operator everywhere. If this one makes the languages consistent, and that one controls dot, then it arises naturally that this has no effect when that one is disabled.

Even if it's too late to incorporate that change into regular Haskell, I think it's perfectly well situated for a "beginner haskell" which should strive to make error messages better and remove odd ambiguities at the cost of breaking with history. If the one criteria for an extension is it could conceivably end up in some language standard, that satisfies it for me .

Please write it!

AntC2:
>the analogous situation appears in the term level and type level. Adding a space around `.` changes the meaning at the term level as well:
>  ... ;-)
> 
> FWIW a similar situation surrounds `..` as well, where `[Foo..]` is interpreted as `[.]`, which is a syntax error, and `[Foo ..]` is interpreted as `enumFrom Foo`.

OK. I think there's already plenty of special-casing around `.` such that a claim that `.` is 'just an operator' is specious. So there's nothing very consistent at term level to be consistent with at type level. Note `.` at term level also appears as decimal separator. (I think `.`'s special-casing is similar to prefix `-` -- which is just hard-coded in the syntax to be `Prelude.negate` and you can't re-define prefix `-` to be your own operator.)

What _is_ reliably consistent at term level is that `.` surrounded by whitespace `(`or by parens`)` means the operator; and you can define that operator for your own purposes to override `Prelude..`

If I remember the discussion before re tight-binding postfix `.`: it was no lesser a personage than SPJ suggesting it; and there were howls of disgreement.

* People claimed that they thought of composed functions `f.g.h`as just a single function spelled with dots inside.
* In particular, Lensists claimed they thought of `point.x` as a single lens spelled with dots inside.
* But note that very often you have to surround the dotty bits with `( ... )`, otherwise you get parse errors.
* (I was perplexed: `(f . g . h)` I know as apply `h`; apply `g` to the result; apply `f` to the resut from that; which is consistent with using functions prefix throughout Haskell. Whereas `(point . x)` has to mean apply `point` to the data structure; apply `x` to the result from that; which _prima facie_ is postfix -- certainly it looks like OOP postfix dot-apply, deliberately so. I note SPJ has described lens-compose being able to use just ordinary `.`-in-reverse as "cute"; signifying to me: something of a surprise.)

Examples/question for those using names-with-dots-inside:

    map f.g.h xs
    over point.x atom

are syntax errors -- which'll probably be reported as type mis-matches. Should be

    map (f.g.h) xs
    over (point.x) atom

So the names-with-dots-inside I find a somewhat bogus claim -- it's not analagous to dotty module prefixes. Then wouldn't it be nice to be able to use the first form? So thanks @Ericson2314 for the invitation, but I don't want to disallow `.` as an operator; I want to say:

* `.` surrounded by whitespace is a regular operator.
* tightly-bound `.` is a distinct operator/special syntax, and binds tighter than function apply. 
* (My proposal will have to figure out how to support those wanting it to mean tight-binding compose vs those wanting tight-binding postfix function apply vs ....)

Then coming to this proposal at the type level:

* Probably people will want to use `.` as type-function composition.
* You can't define any type-operator to do that today, because you can't partially apply a type function.
* Assuming if/when that arrives, type functions are type families are still spelled upper-case, `A.B.C.F.D.E.G` will not behave analagously to how `A.B.C.f.D.E.g` behaves at term level, nor to the proposed behaviour at type level if `f, g` are type vars. (Although if they're type vars, they couldn't appear with module prefix??)

Then this proposal, rather than saying "we can use whitespace to disambiguate", should say "we _must_ use whitespace". IOW if you don't use whitespace (or non-alphas), that's a lexical error.

goldfirere:
To me, this proposal suggests exactly the same treatment of `.` in types as it has in terms. As has been true for a long time, a capitalized identifier, followed by a `.` (without a space), followed by another identifier (without a space) is a qualified identifier. Nothing about that behavior is suggested to change here.

Ericson2314:
At risk of detailing this further, I think what @AntC2 wants is something like Agda's lexing rules for the `.` operator. I would like that for *all* infix operators, but that is also definitely a separate proposal.

AntC2:
To be clear: I'm not trying to "derail" this proposal by adding extra syntax around `.`. The sketch for a tight-binding dot was just to show how GHC could meet a lot of diverse requirements in future. I want to keep the options open.

I'm happy to allow `.` as a type operator. The only rule I want to adjust here is to insist on whitespace around it. I'm not seeing that suggestion is out of scope for this proposal. Without it, I'd rather not have `.` as a type operator. A `.` embedded in a long string of alphas with no spaces around is already hard to spot in code you haven't written yourself. What do others think?

I see no rush to implement this proposal: partially-applied type functions and being able to compose them is probably years away.

int-index:
> The only rule I want to adjust here is to insist on whitespace around it.

The issue is that this proposal is only concerned with type-level `(.)`. So you're asking to add a new rule which would we specific to the type-level dot, which is contrary to the spirit of this proposal of having the same rules in terms in types.

If your suggestion is to require whitespace in both terms and types, then it is a breaking change and therefore out of scope of this proposal.

int-index:
@nomeata A week has passed since the last comment, so I would like to submit this proposal to the committee. Here's a summary of the discussion:

* @phadej was concerned about error message quality. I plan to address this issue in the implementation.
* @simonpj asked whether there is conflict with the `forall <tvs>. <type>` syntax. The proposal states that there is no conflict.
* @simonpj asked about the grammar change in the proposal and its specifics (a change to annotations, compiler output, inlining productions, etc). I clarified that the grammar change in the proposal is for illustration purposes and does not dictate the actual implementation.
* @goldfirere notes that this proposal creates opportunity for writing obfuscated code:
  ```
  forall f g . f . g -> ...
  ```
  I respond that there's already plenty of such opportunity and it is up to the programmer to write readable code:
  ```
  forall f g. (f . g) -> ...
  ```
* @AntC2 objects that having `(.)` as an operator even in terms was a mistake, and lists its downsides (a conflict with the syntax of qualified names). While I sympathise with the argument, the goal of this proposal is to have the same rules in terms and types. @AntC2 may want to write a proposal that disallows the `(.)` in both contexts.

* @AntC2 proposes stricter disambiguation rules for the uses of `(.)` as a type operator versus its uses in qualified names. I opt to reject this suggestion because the goal is to have the exact same disambiguation rules in types and in terms, and changing the rules for terms is out of scope of this proposal.

I hope I didn't miss anything and addressed all comments. Thanks!

PR #174, opened 2018-10-03
Lower precedence for {-# UNPACK #-}

int-index:
The `{-# UNPACK #-}` pragma is not valid inside a type, but binds more tightly than function application. This is suboptimal because it contradicts user expectations ([Trac #14761]) and forces unecessary parenthesization of types. We propose to lower the precedence of `{-# UNPACK
#-}`.

[Trac #14761]: https://ghc.haskell.org/trac/ghc/ticket/14761

[Rendered](https://github.com/int-index/ghc-proposals/blob/unpack-pragma-precedence/proposals/0000-unpack-pragma-precedence.rst)

### DIFF
--- a/proposals/0000-unpack-pragma-precedence.rst
+++ b/proposals/0000-unpack-pragma-precedence.rst
@@ -1,0 +1,87 @@
+Lower precedence for {-# UNPACK #-}
+===================================
+
+.. proposal-number::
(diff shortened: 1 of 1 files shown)


simonpj:
OK with me.

int-index:

PR #30, opened 2016-12-10
Levity-Polymorphic Type Classes

Icelandjack:
Original Trac ticket: [#12708](https://ghc.haskell.org/trac/ghc/ticket/12708).
[Reddit submission](https://www.reddit.com/r/haskell/comments/5hmk4e/ghc_proposal_levity_polymorphic_type_classes/).

 Lets us define instances like `Num Int#`

```haskell
{-# Language InstanceSigs #-}

class Num (a :: TYPE rep) where
  (+) :: a -> a -> a
  (*) :: a -> a -> a
  fromInteger :: Integer -> a

instance Prelude.Num n => Num (n :: Type) where
  ...

instance Num Int# where
  (+) :: Int# -> Int# -> Int#
  (+) = (+#)

  (*) :: Int# -> Int# -> Int#
  (*) = (*#)

  fromInteger :: Integer -> Int#
  fromInteger (fromInteger -> I# int#) = int#

instance Num Double# where
  (+) :: Double# -> Double# -> Double#
  (+) = (+##)

  (*) :: Double# -> Double# -> Double#
  (*) = (*##)

  fromInteger :: Integer -> Double#
  fromInteger (fromInteger -> D# dbl#) = dbl#
````

and

```haskell
class Show (a :: TYPE rep) where
  show :: a -> String

instance Prelude.Show a => Show (a :: Type) where
  show :: a -> String
  show = Prelude.show

instance Show Int# where
  show :: Int# -> String
  show int# = show (I# int#) ++ "#"

instance Show Double# where
  show :: Double# -> String
  show dbl# = show (D# dbl#) ++ "#"
```

[Rendered](https://github.com/Icelandjack/ghc-proposals/blob/patch-1/proposals/0000-type-class-levity.rst).

PR #175, opened 2018-10-05
Add `liftTyped` to the `Lift` typeclass

harpocrates:
[Rendered](https://github.com/harpocrates/ghc-proposals/blob/lift-typed/proposals/0000-lift-typed.rst).

### DIFF
--- a/proposals/0000-lift-typed.rst
+++ b/proposals/0000-lift-typed.rst
@@ -1,0 +1,139 @@
+Add ``liftTyped`` to the ``Lift`` typeclass
+===========================================
+
+.. proposal-number:: 
+.. trac-ticket:: 
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/175>`_.
+.. sectnum::
+.. contents::
+
+The ``Lift`` typeclass provides a way of lifting values into untyped Template Haskell expressions. ::
+
+    class Lift t where
+      lift :: t -> Q Exp
+      default lift :: Data t => t -> Q Exp
+
+However, there is no easy way to lift values into *typed* Template Haskell expressions (`Trac #14671 <https://ghc.haskell.org/trac/ghc/ticket/14671>`_). This is suboptimal since the type information required to produce such expressions is readily available. We propose to add a new method called ``liftTyped`` to ``Lift`` to take advantage of this. ::
+
+    class Lift t where
+      lift :: t -> Q Exp
+      liftTyped :: t -> Q (TExp t)
+
+See also `this thread on the libraries list <https://mail.haskell.org/pipermail/libraries/2018-January/028409.html>`_.
+
+
+Motivation
+------------
+Typed Template Haskell expressions (see ``TExp`` in `template-haskell <http://hackage.haskell.org/package/template-haskell>`_) are expression ASTs annotated with the type of the expression the AST represents. Using GHC's *typed quotes* and *typed splices* it is possible to combine and manipulate typed expressions while maintaining confidence that the generated code will typecheck properly and have a certain known type. ::
+
+    ghci> :set -XTemplateHaskell +t
+    ghci> x = [|| True && False ||]   -- this is a typed quote
+    x :: Q (TExp Bool)
+    ghci> y = [|| True && "foo" ||]
+
+    <interactive>: error:
+        • Couldn't match expected type ‘Bool’ with actual type ‘[Char]’
+        • In the second argument of ‘(&&)’, namely ‘"foo"’
+          In the Template Haskell quotation [|| True && "foo" ||]
+          In the expression: [|| True && "foo" ||]
+
+These are guarantees that we do not get from using untyped Template Haskell expressions (see ``Exp`` in `template-haskell <http://hackage.haskell.org/package/template-haskell>`_), quotes, and splices. ::
+
+    ghci> x' = [| True && False |]   -- this is an untyped quote
+    x' :: Q Exp
+    ghci> y' = [|| True && "foo" ||]
+    y' :: Q Exp
+
+Template Haskell provides an ``unsafeTExpCoerce :: Q Exp -> Q (TExp a)`` function for when a user is only able to produce an untyped ``Exp`` but knows that this expression has type ``a``. Since the ``lift`` function returns an untyped expression, it can't be spliced into a typed quote without first calling ``unsafeTExpCoerce``. This is unfortunate and completely avoidable since ``lift`` actually does know the type of its argument. ::
+
+    ghci> import Language.Haskell.TH.Syntax
+    ghci> f = False
+    ghci> x'' = [|| True && $$(unsafeTExpCoerce (lift f)) ||]
+    x'' :: Q (TExp Bool)
+
+Proposed Change Specification
+-----------------------------
+Add a ``typedLift`` method to the ``Lift`` typeclass and define default implementations of ``lift`` and ``typedLift`` in terms of each other. ::
+
+    class Lift t where
+      lift :: t -> Q Exp
+      lift = unTypeQ . liftTyped
+
+      liftTyped :: t -> Q (TExp t)
+      liftTyped = unsafeTExpCoerce . lift
+
+      {-# MINIMAL lift | liftTyped #-}
+
+
+For completeness, here is what ``Lift`` looks like today: ::
+
+    class Lift t where
+      lift :: t -> Q Exp
+      default lift :: Data t => t -> Q Exp
+      lift = liftData
+
+
+Effect and Interactions
+-----------------------
+Following the last example from the motivation section, it would now be possible to write: ::
+
+    ghci> f = False
+    ghci> x'' = [|| True && $$(liftTyped f) ||]
+    x'' :: Q (TExp Bool)
+
+Costs and Drawbacks
+-------------------
+The main drawback to this proposal is that it can break existing instances of ``Lift`` which rely on the current default implementation of the `lift` method. These cases are:
+
+* empty instances of ``Lift`` ::
+  
+       {-# LANGUAGE DeriveDataTypeable #-}
+       
+       import Data.Data (Data, Typeable)
+  
+       data Foo = MkFoo Int deriving (Typeable, Data)
+       instance Lift Foo
+  
+* instances of ``Lift`` that are derived via ``-XDeriveAnyClass`` ::
+
+       {-# LANGUAGE DeriveDataTypeable, DeriveAnyClass, DerivingStrategies #-}
+       
+       import Data.Data (Data, Typeable)
+  
+       data Foo = MkFoo Int deriving (Typeable, Data) deriving anyclass (Lift)
+
+On the other hand, ``-XDeriveLift`` has been the recommended way to derive ``Lift`` instances for a while now. Furthermore, code that would be broken could easily be unbroken in a fully backwards-compatible fashion with ``instance Lift Foo where { lift = liftData }``.
+
+Alternatives
+------------
+There are a handful of other ways to introduce ``liftTyped``.
+
+* As a standalone top-level funtion ::
+
+      liftTyped :: Lift t => t -> Q (TExp t)
+      liftTyped = unsafeTExpCoerce . lift
+
+  This means instances of ``Lift`` can't be directly defined in terms of typed expression quotations.
+
+* As the sole method of the ``Lift`` typeclass ::
+
+      class Lift t where
+        liftTyped :: t -> Q (TExp t)
+        default liftTyped :: Data t => t -> Q (TExp t)
+        liftTyped = unsafeTExpCoerce . liftData
+
+      lift :: Lift t => t -> Q Exp
+      lift = unTypeQ . liftTyped
+      
+  This breaks any existing manually defined ``Lift`` instances in a way that can only be worked around via with CPP.
+
+Unresolved Questions
+--------------------
+None that I am aware of.
+
+Implementation Plan
+-------------------
+I (Alec Theriault) will implement this change. The work is already
+done in `Phab:D5198 <https://phabricator.haskell.org/D5198>`_.


yallop:
This looks useful.  We use [a similar definition of `typedLift` in `frex`](https://github.com/yallop/haskell-frex/blob/19e5802caf5bc05a0ad1ce33f2fe5fe1258226d4/src/Data/PartiallyStatic.hs#L30-L31); it'd be better to have it as a method.

A second useful step: change the translation of splices to use `typedLift` instead.  At present, code like this

```haskell
f x = [|| x ||]
```

is (roughly speaking) translated into


```haskell
f x = unsafeTExpCoerce (lift x)
```

but it could instead be translated into

```haskell
f x = typedLift x
```

gridaphobe:
The proposal doesn’t specify any changes to `DeriveLift`, so I assume the intention is to continue deriving the untyped `lift` and use the default implementation of `liftTyped`. Wouldn’t it be better to derive `liftTyped` instead?

RyanGlScott:
What does "derive `liftTyped`" mean exactly? I ask since, AFAICT, there's no way to construct values of type `TExp a` besides (1) using typed TH quotes, or (2) using the `unsafeTExpCoerce :: Q Exp -> Q (TExp a)` function. As one example, if you write this:

```haskell
{-# LANGUAGE DeriveLift #-}
import Language.Haskell.TH.Syntax

data Pair a = MkPair a a deriving Lift
```

Then this is the `Lift` instance that would be generated under the hood (slightly cleaned up):

```haskell
instance Lift a => Lift (Pair a) where
  lift (MkPair a1 a2) = conE 'MkPair `appE` lift a1 `appE` lift a2
```

What would a derived implementation of `liftTyped` look like for `Pair`? I'm having trouble envisioning what it would be except for `unsafeTExpCoerce (conE 'MkPair `appE` lift a1 `appE` lift a2)`, but at that point, you might as well just use the proposed default implementation for `liftTyped`.

gridaphobe:
I assumed it would use the typed quotes, e.g. right now I can write

```haskell
data Pair a = MkPair a a

liftTyped :: Lift a => Pair a -> TExpQ (Pair a)
liftTyped (MkPair x y) = [||MkPair x y||]
```

which is exactly what I would want GHC to derive.

RyanGlScott:
Ah, I see what you mean. I think I ruled out that idea when originally implementing `DeriveLift` because at the time, using the TH quote syntax required enabling the `TemplateHaskell` language extension, which would have prevented the use of `DeriveLift` on compilers that didn't support `TemplateHaskell` (such as stage-1 or cross-compiled GHCs). But thankfully, there now exists a `TemplateHaskellQuotes` extension that only enables the use of TH quoting syntax without enabling the other parts of the `TemplateHaskell` extension (such as splicing) that certain compilers have trouble with.

If we simply have `DeriveLift` enable `TemplateHaskellQuotes` when typechecking the code that it generates, then the approach in https://github.com/ghc-proposals/ghc-proposals/pull/175#issuecomment-427480743 would be quite feasible. In fact, it might even _simplify_ the current implementation of `DeriveLift`, since we wouldn't need to refer to any combinators from `Language.Haskell.TH.Lib` in generated `Lift` instances. Among other things, this would make it easier to implement [this proposal](https://ghc.haskell.org/trac/ghc/ticket/14030), which I postponed implementing since I ran into annoying import cycle issues due to trying to import things from `Language.Haskell.TH.Lib` from `Language.Haskell.TH.Syntax` (since the former imports the latter as well).

RyanGlScott:
I was curious to see if the code-generation strategy in https://github.com/ghc-proposals/ghc-proposals/pull/175#issuecomment-427480743 was feasible, so as an experiment, I modified `DeriveLift` to generate code for `lift` using TH quotes. Happily, I found that it worked almost flawlessly.

...almost. There's one gotcha: currently, `DeriveLift` has some special smarts to allow deriving `Lift` instances for constructors that have fields with certain unlifted types. For example, if you write this:

```haskell
{-# LANGUAGE DeriveLift #-}
{-# LANGUAGE MagicHash #-}
module Foo where

import GHC.Exts (Int#)
import Language.Haskell.TH.Syntax

data Foo = MkFoo Int# deriving Lift
```

Then this instance will be generated:

```haskell
instance Lift Foo where
  lift (MkFoo i) = conE 'MkFoo `appE` litE (intPrimL (toInteger (I# i)))
```

Notice that this doesn't `lift` the field of `MkFoo`, but rather uses some special smarts involving `litE` and `intPrimL`. If we were to attempt to generate the `Foo` instance like so:

```haskell
instance Lift Foo where
  lift (MkFoo i) = [| MkFoo i |]
  -- lift (MkFoo i) = conE 'MkFoo `appE` lift i
```

Then this won't typecheck:

```
Foo.hs:12:20: error:
    • Couldn't match a lifted type with an unlifted type
      When matching types
        t0 :: *
        Int# :: TYPE 'GHC.Types.IntRep
    • In the first argument of ‘lift’, namely ‘i’
      In the expression: lift i
      In the expression:
        [| MkFoo i |]
        pending(rn) [<i, lift i>]
   |
12 |   lift (MkFoo i) = [| MkFoo i |]
   |                    ^^^^^^^^^^^^^
```

This is because `lift` expects its argument to have a lifted type (i.e., a type of kind `Type`), whereas `i` has type `Int#`, which has kind `TYPE IntRep`, not `Type`. This problem also applies for the proposed `liftTyped` method–in fact, the problem is even worse for `liftTyped`, since the _only_ mechanism we have for directly constructing `TExp` expressions is through TH quotes.

I can see two potential ways to resolve this issue:

1. Stop supporting the special cases for unlifted types in `DeriveLift` entirely. This would be a breaking change, but it's unclear to me how many people would be affected by this in practice.
2. Tweak `Lift` and `TExp` slightly to support unlifted types. We could leverage levity polymorphism and redefine them like so:

   ```haskell
   class Lift (a :: TYPE r) where ...
   newtype TExp (a :: TYPE r) = ...
   ```

   If we did this, then instead of hardcoding support for `Int#` _et al._ directly into `DeriveLift`, we could simply give `Lift` instances for the types that we support directly:

   ```haskell
   instance Lift Int# where
     lift i = litE (intPrimL (toInteger (I# i)))
   ...
   ```

   Once that's done we could simply derive a `Lift` instance for `Foo` above like so:

   ```haskell
   instance Lift Foo where
     lift (MkFoo i) = [| MkFoo i |]
     -- liftTyped (MkFoo i) = [|| MkFoo i ||]
   ```

   And it would Just Work™ without any further changes. It would require tweaking the definitions of `Lift` and `TExp` more, however.

What do people think of this?

gridaphobe:
Are there any downsides to option 2, aside from making the types a bit more involved? It seems like a very sensible solution to me.

RyanGlScott:
> Are there any downsides to option 2, aside from making the types a bit more involved?

Based on the feedback in #30 (a similar proposal about levity polymorphic type classes), the main concerns were:

1. Current uses of `lift` in combination with visible type application will be broken due to the introduction the `rep` type variable.
2. It's not clear what effect this would have on error messages involving `lift`.

That being said, the situation here is a bit different in that (a) this isn't a type class in `base` we're changing, so we don't have to butt heads against the Haskell Report, and (b) there's something actually depending on `Lift` becoming levity polymorphic in order to work here (as opposed to #30, which felt like somewhat of an unforced change).

That being said, I think I've wildly sidetracked from @harpocrates' original goal, which was to just introduce a new method to `Lift`. I'm not sure if it's worth updating the proposal to include all of this extra hullabaloo about changing `DeriveLift`—it could just as well be its own proposal.

harpocrates:
It seems to me that typed quotes are already sneakily leaking some levity polymorphism into `TExp` (so perhaps we could consider making `Lift` levity polymorphic as just a bug fix? seriously though... 😃).

    ghci-8.6.1> import Language.Haskell.TH.Syntax
    ghci-8.6.1> import GHC.Exts
    ghci-8.6.1> :set -XMagicHash -XTemplateHaskell +t
    ghci-8.6.1> e = [|| 1# ||]
    e :: Q (TExp Int#)
    ghci-8.6.1> e = [|| 1# ||] :: Q (TExp Int#)   -- GHC suddenly catches on... :)

    <interactive>:5:27: error:
        • Expecting a lifted type, but ‘Int#’ is unlifted
        • In the first argument of ‘TExp’, namely ‘Int#’
          In the first argument of ‘Q’, namely ‘(TExp Int#)’
          In an expression type signature: Q (TExp Int#)

In fact, I may even have personally (accidentally) taken advantage of this before, because you can still use `e :: Q (TExp Int#)`:

    ghci-8.6.1> e' = [|| I# $$(e) ||]
    e' :: Q (TExp Int)

NB: this is not just a GHCi thing.

```haskell
{-# LANGUAGE MagicHash, TemplateHaskell #-}

import GHC.Exts
import Language.Haskell.TH.Syntax

main = print =<< runQ (unTypeQ [|| I# $$e ||])
  where e = [|| 1# ||] -- :: Q (TExp Int#)
```

goldfirere:
If we had https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0026-explicit-specificity.rst, then we could do option (2) without the drawbacks. I actually advocate for doing option (2) with the drawbacks. TH users get a certain amount of breakage. This will be a smaller amount than usual, and the breakage will be in the form of a fairly-easy-to-understand kind error (`RuntimeRep /= Type`). (That is, this is easy to understand by the kinds of people who do visible type application in `Lift` instances.)

simonpj:
> we could do option (2) without the drawbacks. I actually advocate for doing option (2) with the drawbacks.

How so, precisely?  I think you mean that the runtime-rep argument can be specified as `Inferred`, so that it never appears in a visible type application?  Or did you mean something else?

Regardless, this looks like a (good) separate proposal: use levity polymorphism to make `Lift`, and perhaps TH quotes themselves, work at unlifted types.

goldfirere:
> How so, precisely? I think you mean that the runtime-rep argument can be specified as Inferred, so that it never appears in a visible type application? Or did you mean something else?

Yes, exactly.

harpocrates:
I went ahead and tried to implement option (2), following what @RyanGlScott outlines in https://github.com/ghc-proposals/ghc-proposals/pull/175#issuecomment-427611846. I have a working prototype which works fine for unlifted types, but ran into some other non-trivial issues:

  1. default implementations of methods don't work (eg. `lift x = unTypeQ (liftTyped x)`), presumably since GHC would actually have to make a different `lift` method for each runtime rep
  2. writing `class Lift (t :: TYPE r) where ...`, `instance Lift Int# where ...`, etc. requires having `ghc-prim` be a dependency of `template-haskell`

Point 1 might be relevant to the original proposal here; unless there is some sort of path forward towards making 1 possible, perhaps it is better to move `lift` out of the `Lift` typeclass... Thoughts?

RyanGlScott:
Point (2) doesn't sound too bad. `base` depends on `ghc-prim`, so `template-haskell` already depends on `ghc-prim` transitively.

I don't understand point (1). Can you provide more details on what isn't working? I would have expected something like `lift x = unTypeQ (liftTyped @r @a x)` to work (assuming that this is a default method of `class Lift (a :: TYPE r)`).

harpocrates:
> I don't understand point (1). Can you provide more details on what isn't working? I would have expected something like `lift x = unTypeQ (liftTyped @r @t x)` to work (assuming that this is a default method of `class Lift (t :: TYPE r)`).

Here is a reduced example of the problem. Also, is there somewhere else we should move this discussion?

```haskell
{-# LANGUAGE KindSignatures, PolyKinds, RankNTypes #-}
module Foo where

import GHC.Types

class Foo (t :: TYPE r) where
  foo :: t -> t
  foo x = x

-- Foo.hs:8:7: error
--﻿      A levity-polymorphic type is not allowed here:
--       Type: t
--       Kind: TYPE r
--     In the type of binder ‘x’
--   |
-- 8 |   foo x = x
--   |       ^
```

harpocrates:
Maybe that wasn't such a problem after all. There is somewhat satisfactory workaround using `Defaultsignatures`.

```haskell
class Lift (t :: TYPE r) where
  lift :: t -> Q Exp
  default lift :: (r ~ 'LiftedRep) => t -> Q Exp
  lift = unTypeQ . liftTyped

  liftTyped :: t -> Q (TExp t)
  default liftTyped :: (r ~ 'LiftedRep) => t -> Q (TExp t)
  liftTyped = unsafeTExpCoerce . lift
```

That means that unlifted instances of `Lift` will have more boilerplate, but that gets limited to `template-haskell`. The main unpleasantness is going to be the more complex types in the generated Haddocks.

How would folks feel about this? Unless someone voices an objection (or I find some other issue), I'm going to modify the initial proposal, since the definition of `Lift` listed above is quite different from what was initially proposed.

goldfirere:
That definition is awfully advanced. I understand why we want it, but we'll need to document very carefully. It makes me wonder if that should be `RRLift` and keep `Lift` to be simple.

harpocrates:
I've implemented a levity-polymorphic `TExp`, quotes, and `Lift` (along with simplifications in `DeriveLift`) in https://phabricator.haskell.org/D5220. Shall I wait until (if?) this proposal is accepted to open a new one about the levity polymorphism changes?

harpocrates:
Could you submit this to the committee for review @nomeata? It seems everyone who has replied is in support and there are no unresolved questions.

Most of the discussion has revolved on additional enhancements we could make to `Lift`, and that discussion belongs in a separate proposal.

harpocrates:
I was mostly concerned with backwards compatibility, but if the committee has any opinion on the matter I’d be happy to amend the proposal!

bgamari:
I am happy to report that this proposal has been accepted by the steering committee! Thanks, @harpocrates!

RyanGlScott:
Does this mean that this can be merged now?

nomeata:
There was some question about which version on the mailing list:
https://mail.haskell.org/pipermail/ghc-steering-committee/2019-February/000917.html
So maybe @bgamari wanted to groom the proposal before merging?

nomeata:
I have learned that this proposal was subsumed by #209, thus closing. Thanks @harpocrates for kicking off this discussion!

PR #179, opened 2018-11-05
Tweak the printing of foralls

goldfirere:
This proposal a small change to the way GHC prints `forall`-types. See the Motivation section.

[Rendered](https://github.com/goldfirere/ghc-proposals/blob/printing-foralls/proposals/0000-printing-foralls.rst)

### DIFF
--- a/proposals/0000-printing-foralls.rst
+++ b/proposals/0000-printing-foralls.rst
@@ -1,0 +1,125 @@
+Clean up printing of foralls
+============================
+
+.. proposal-number:: Leave blank. This will be filled in when the proposal is
+                     accepted.
+.. trac-ticket:: Leave blank. This will eventually be filled with the Trac
+                 ticket number which will track the progress of the
+                 implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+GHC has two infelicities around the way it prints ``forall`` types around visible type application.
+This proposes a fix for both. The two
+fixes are entirely separable, but it seems sensible to debate them together.
+
+Motivation
+----------
+With visible type application, we have a distinction between *specified* variables and *inferred* ones.
+In brief, a specified variable is one that has been written by the user, whereas an inferred variable
+is invented by GHC. Here is an example::
+
+  id :: a -> a
+  id x = x
+
+  id2 x = x
+
+The type of ``id`` contains a specified variable, ``a``, written by the user. Though ``id2`` will be
+inferred to be polymorphic, its type variable is inferred, not specified.
+
+Visible type application works only with specified variables. This is because we can rely on the ordering
+and existence of specified variables only; inferred variables might be reordered between different minor
+versions of GHC. Thus, ``id @Int 5`` is accepted, while ``id2 @Int 5`` is rejected. None of this is new.
+
+There are two problems with the status quo:
+
+1. GHC prints braces around inferred variables, but only with ``-fprint-explicit-foralls`` on. This is
+   sometimes confusing. For example::
+
+     GHCi, version 8.7.20181028: http://www.haskell.org/ghc/  :? for help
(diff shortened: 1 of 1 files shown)


simonpj:
I like this proposal.

> The drawback to change (1) is that it means GHC is printing more fancy widgets in types. 

But the truth is that GHC supports Inferred, Specified, and Required type variables; and I think it's more confusing to suppress the distinctions than to display it.   Suppressing the forall altogether is fine; but displaying the forall in two different ways is plain confusing.

Note the for Required we already show `forall k -> k -> Type` (say), with a funny widget.  Because suppressing it would be positively misleading to suppress either the forall or the "->".

int-index:
I don't like that we default to `:t +v` only when the expression is a single identifier.

Here's what we have currently. With `+v`:
```
ghci> :t +v realToFrac
realToFrac :: forall a b. (Real a, Fractional b) => a -> b
ghci> :t +v realToFrac @Double
realToFrac @Double
  :: forall b. (Real Double, Fractional b) => Double -> b
```

Without `+v`:

```
ghci> :t realToFrac
realToFrac :: forall {a} {b}. (Real a, Fractional b) => a -> b
ghci> :t realToFrac @Double
realToFrac @Double :: forall {b}. Fractional b => Double -> b
```


It's not very ergonomic, but at least it's consistent. After this proposal, we'll have a confusing situation like this:

```
ghci> :t realToFrac
realToFrac :: forall a b. (Real a, Fractional b) => a -> b
ghci> :t realToFrac @Double
realToFrac @Double :: forall {b}. Fractional b => Double -> b
```

We got what we expected in the first case, but a regeneralized type in the second case.

The issue with making `+v` the default seems to be unsolved constraints:

```
ghci> :t +v realToFrac @Double
realToFrac @Double
  :: forall b. (Real Double, Fractional b) => Double -> b
```

That is, we don't want `Real Double` to be there. Is there truly no way to solve the constraints without shuffling around the `forall`s?

goldfirere:
> Is there truly no way to solve the constraints without shuffling around the `forall`s?

Maybe, but it would make me nervous. For example, we could instantiate and solve, and then use the solved constraints to figure out which original constraints to suppress from printing. But I'm not confident we'd always get it right.

goldfirere:
For what it's worth, I'm quite unsold on the treat-single-identifiers specially. This proposal is effectively a collaboration between @simonpj and me, and I was only weakly convinced of that part of it.

monoidal:
I support idea 1. I got confused by this behavior in the linear types implementation. Given a datatype

`data T a = MkT a`

With linear types, the constructor `MkT` becomes linearity-polymorphic:

`MkT :: forall {m :: Multiplicity} a. a -->.(m) T a`

In plenty of error messages, GHC prints `forall` even if `-fprint-explicit-foralls` is not enabled. This means that we show `MkT :: forall (m :: Multiplicity)...` which is confusing. (Like with levity polymorphism, multiplicity is not printed by default.)

Additionally, this change makes `-fprint-explicit-foralls` do one thing instead of two.

I believe the implementation is:

```
--- a/compiler/iface/IfaceType.hs
+++ b/compiler/iface/IfaceType.hs
@@ -1054,10 +1054,7 @@ pprIfaceForAllCoBndrs bndrs = hsep $ map pprIfaceForAllCoBndr bndrs
 
 pprIfaceForAllBndr :: IfaceForAllBndr -> SDoc
 pprIfaceForAllBndr (Bndr (IfaceTvBndr tv) Inferred)
-  = sdocWithDynFlags $ \dflags ->
-                          if gopt Opt_PrintExplicitForalls dflags
-                          then braces $ pprIfaceTvBndr False tv
-                          else pprIfaceTvBndr True tv
+  = braces $ pprIfaceTvBndr False tv
 pprIfaceForAllBndr (Bndr (IfaceTvBndr tv) _)  = pprIfaceTvBndr True tv
 pprIfaceForAllBndr (Bndr (IfaceIdBndr idv) _) = pprIfaceIdBndr idv
```

Around 70 tests need to be adjusted: https://circleci.com/gh/tweag/ghc/2353

I have no strong opinion on idea 2 - it's a tradeoff between making the system less uniform but more user-friendly.

goldfirere:
I've revised the specification in this proposal to go in a slightly different direction. Feedback is welcome. Thanks!

simonpj:
I like the revised version. Let's do it.

goldfirere:
Sounds quiet here. Let's submit, @nomeata. Thanks.

PR #182, opened 2018-11-17
Add control flow hint pragmas.

AndreasPK:
This proposal suggests a new pragma to allow control flow hints.

In particular we can give the likelyhood that a value is constructed by a specific Constructor.
And that likelyhood that a specific case alternative is taken.

[Rendered](https://github.com/AndreasPK/ghc-proposals/blob/master/proposals/0000-likelihood-annotations.rst)

### DIFF
--- a/proposals/0569-multiline-strings.rst
+++ b/proposals/0569-multiline-strings.rst
@@ -93,11 +93,6 @@ A working prototype is available at `brandonchinn178/string-syntax <https://gith
 
 #. Post-process the string in the following steps:
 
-   #. Collapse string gaps
-
-      * See `Section 2.6 <https://www.haskell.org/onlinereport/haskell2010/haskellch2.html#x7-200002.6>`_ of the Haskell 2010 Report
-      * See the example in *Section 3.3 String gaps*
-
    #. Split the string by lexical ``newline`` characters (as defined in the `Haskell report <https://www.haskell.org/onlinereport/haskell2010/haskellch2.html#x7-160002.2>`_)
 
    #. Convert leading tabs into spaces
@@ -124,6 +119,8 @@ A working prototype is available at `brandonchinn178/string-syntax <https://gith
    #. If the last character of the string is a newline, remove it
 
 #. After parsing, it becomes indistinguishable to the equivalent single-quoted string (modulo annotations for exact-printing)
+
+In normal strings, string gaps are effectively equivalent to replacing with ``\&``. They're not simply removed, e.g. ``"\65\   \0"`` results in ``"A0"``. This behavior is preserved in multiline strings; see the examples in *Section 3.3 String gaps* for more details.
 
 Common whitespace prefix calculation
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -263,7 +260,7 @@ Because characters immediately after the ``"""`` delimiter should be included ve
 String gaps
 ~~~~~~~~~~~
 
-String gaps are collapsed first and not included in the whitespace calculation
+String gaps are collapsed before the whitespace calculation
 
 ::
 
@@ -276,6 +273,21 @@ String gaps are collapsed first and not included in the whitespace calculation
 
   -- equivalent to
   s' = "a b c d e\nf g"
+
+But a string gap starting at the beginning of a line counts as non-whitespace in the whitespace calculation.
+
+::
+
+  -- Imagine \& is substituted for the string gap
+  s =
+      """
+        a b
+      \   \  c d e
+        f g
+      """
+
+  -- equivalent to
+  s' = "  a b\n  c d e\n  f g"
 
 Mixing tabs and spaces
 ~~~~~~~~~~~~~~~~~~~~~~


gridaphobe:
Can you give a sense of how much performance is being left on the table when GHC's heuristics make the wrong decision? For example, you say

> Currently high performance code tends to vary things like constructor order manually for maximal performance.

It would be nice to include a benchmark that demonstrates the improvement from reordering constructors.

AndreasPK:
> Can you give a sense of how much performance is being left on the table when GHC's heuristics make the wrong decision?

Details are hard to quantify. But for containers the source contains a comment that changing the order made a difference of up to 10%. I can't verify that but it seems possible although likely at the extreme end.

Given that the way GHC generates code changes over time adjustments like that one can backfire if the code generation logic is changed. Something that has happened in the past already and will likely keep happening.

Actual performance gained will completely depend on what the final implementation looks like. I think gains of more than 2% are possible, and below that it's likely not worth implementing (or even merging). But I don't want to speculate on the real world impact as it's hard to estimate.

AntC2:
IIUC for data constructors, the low-order few bits of a pointer value/thunk are used to indicate which constructor -- that is for the first few constructors declared, in textual order of appearance. Then typically case despatch doesn't need to look at the value at all.

Does that mean that instead of using textual order, data constructor coding should be allocated in likelihood order? That would be a nice separation of concerns: there's a great deal of library code reorganised to put constructors textually in illogical-but-likely sequence. Pinging @treeowl .

(How many bits are available/how many constructors can be shortcut depends on machine wordsize.)

AndreasPK:
> IIUC for data constructors, the low-order few bits of a pointer value/thunk are used to indicate which constructor -- that is for the first few constructors declared, in textual order of appearance. Then typically case despatch doesn't need to look at the value at all.

This is mostly correct, however if a type has many constructors we always look at the actual value.

> Does that mean that instead of using textual order, data constructor coding should be allocated in likelihood order? That would be a nice separation of concerns: there's a great deal of library code reorganised to put constructors textually in illogical-but-likely sequence. Pinging @treeowl .
> 
> (How many bits are available/how many constructors can be shortcut depends on machine wordsize.)

Essentially yes. These are all implementation details, but currently for types with less than 8 constructors we generate code similar to:

```
if con == Con1 then e1
else if con == Con2 then e2
else if con == Con3 then e3
...
```

Depending on the actual case expression it can also end up doing binary search.

So the the first thing we can do is change the order of checks depending on the given likelihoods. Another thing you can do is turn code like `if con == Con1 then e1 else e2` into `if con /= Con1 then e2 else e1`. Which has impact on codelayout which also affects performance.

That being said, these are implementation details a user shouldn't have to care about.

cartazio:
Yeah.. as Andreask says, currently there’s some tricks that are possible on
the data type defining side. But there’s currently nothing possible on the
case expression author side

So there’s , very approximately the following scenarios underlying why this
sort of pragma is important

1) default branch path the first time the cpu sea that branch.

Modern cpus are very pipelined, and whichever ever branch is the default or
“likely” branch will start running while resolving the actual value of the
branch. After that first time or so the cpu maintains this info in a tiny
cache


How bad can this be?

I’ve written  Haskell programs where In the inner loop the branch take Is
always different from the previous step.  So pipelining  from the initial
default  OR was always incorrect.  The performance difference between that
code and one which was equvalent but didn't have the false branching was i
think about 10x.. This is an extreme case. But for large programs where
theres lots of branches (more than the predictor cache), good hinting can
make a HUGE difference, (for tight inner loops it will help make sure that
the first loop  iteration is wayy faster/ cheaper than it would other wise
be).


so the point is: I"VE WANTED THIS FOR YEARS, as have many others

one immediate gap i see in the current proposal is it doesn't talk about
ties in weights, or if those are even allowed. I'd think that these sorts
of annotations would need to be totally ordered. And also that internally
we might want these to be either nonzero positive rationals, or log domain
floating point.

probably we'd want rationals, and and either way  we could easily compose
nested case expressions by just multiplying through.

I do think that it maybe should be considered an error to have a "simple
case" expression where two branches have the same "weight"

One question i have is do these numbers act as a rank/ordering  or are
meant to model relative weight?

either way, this is a sorely lacking thing for the performance conscious
haskell and while some details may need fine tuning, this could be a very
high impact feature

cartazio:
a very good suggestion Davean raised in conversation is that "I tend to prefer profiling driven optimization for branch prediction, why can't we have a branch probability file?"

aka: this same work would immediately enable a really nice form of machine/execution driven optimization, so some way to feed statistics about branches taken into ghc to drive this information being populated from program execution data would be pretty great

goldfirere:
I would like a likelihood of 0 to mean that a branch is *never* expected to be called. (This differs from the proposal, which states that 0 would indicate rarity, not impossibility.) Perhaps (maybe with `-O2` only?) GHC needn't promise to be semantics-preserving in likelihood-0 branches. In any case, even if we're too skittish to implement that today, by reserving 0 to mean "never", we can do so in the future.

This might make a real difference! See [here](https://github.com/ghc/ghc/blob/7d9036448a394d7f2eeb158bb71d0fa694f88f56/compiler/typecheck/TcFlatten.hs#L1354) for an example of where code that should never be run affected overall performance in a very measurable where. With `{-# LIKELY 0 #-}`, GHC might be clever enough to zap the variables out of existence.

----------------

https://github.com/ghc-proposals/ghc-proposals/pull/182#issuecomment-439763816 suggests Next-Time (NT) optimization, as opposed to Just-In-Time optimization. This is a vast improvement over the current situation, which is Never optimization (that takes into account runtime metrics). Very compelling idea. Note that this proposal needn't be expanded or changed to be forward compatible with NT optimization -- all the back-end work from this proposal would be reused to implement NT. So this might fall under motivation for this proposal more than anything else.

------------------

This proposal doesn't specify (I think) what happens when both a datatype and a case statement over that datatype have `LIKELY`s. Are they multiplied? Or does one overrule the other?

cartazio:
Lest I forget : there seem to be three different types of data here

1) what encoding we choose for constructors — currently this has to do with
the order and number of constructors.

2) what default distribution / ranking we should assume on the constructors
in a case expression.  I would kinda assume we want to default to something
compatible with currently implicit adhoc choice unless we can cook up a
better approach ?

3) what ranking / distribution for a particular case expression


First : I think it’s pretty important to pin down what meaning this ranking
/ probability should have.  And what may or may not be ambiguous for a
given representation. I also feel like there’s got to be some relationship
to prefix free binary coding/compression going on here.  To explicate more :

Suppose we want to optimize a program P which has a sum type D and a set of
case expressions C over D. Suppose we have the branching statistics S for a
representative  set of executions of this program. Aka S is sort of a
probability distribution of branching weights.

 Our simplified optimization problem might be that we want to desugar our
program to maximize the fraction of  default branches that are correct. And
or minimize the branching depth over all executions.

In that case, a whole program variant of this optimization might choose an
encoding for D that allows for a useful interval of constructors branching,
so that binary search style desugaring is very effective for this program
?  Granted this becomes a sort of whole program optimization. Which I guess
might be np hard or complete to do optimally if my recollection of
complexity results related for compression and or pattern compilation are
correct

I may be skipping some steps. Not sure.

AndreasPK:
> Modern cpus are very pipelined, and whichever ever branch is the default or
“likely” branch will start running while resolving the actual value of the
branch.

What the "default" branch is depends on the CPU, and might change randomly for some of them depending on system state.

As far as I know explicit branch hinting in code is dead. There are branch hint prefixes, but these are ignored on newer models and should not be used as the increase in code size is generally not worth it.

There are heuristics but "perfectly predicting the hot path on the first iteration" is a goal that is likely not achieveable unless 5 other people suddenly show up working on this. Even then maybe not when targeting different CPUs.

However there is still a lot of "low hanging fruit" we can pick by optimizing code layout and the like. Which in general should line up with branch prediction heuristics performing better.

> One question i have is do these numbers act as a rank/ordering  or are
meant to model relative weight?

It's relative weight:

``GHC will optimize code under the assumption that the chance for the i-th alternative to be taken is Li / sum [L1 .. Ln].`` 

> I'd think that these sorts of annotations would need to be totally ordered.

Why? I definitely see a use case for allowing things like:

```
case x of
  Error ->   {-# LIKELY 0 #-} e1
  Result1 -> {-# LIKELY 1 #-} e2
  Result2 -> {-# LIKELY 1 #-} e3

```
Where users either want to rule out exceptional cases or know the other cases are equally as likely.
Now ALL branches having the same weight is pointless, but treating that in a special way doesn't seem worth it.

> And also that internally we might want these to be either nonzero positive rationals or log domain floating point.

While an implementation detail I think we should get away with Integers there. But you can convince me otherwise once I actually have patches!

> "I tend to prefer profiling driven optimization for branch prediction, why can't we have a branch probability file?"
> aka: this same work would immediately enable a really nice form of machine/execution driven optimization, so some way to feed statistics about branches taken into ghc to drive this information being populated from program execution data would be pretty great

It's certainly a long term goal of mine. Profile guided optimization will require most of the work this proposal entails. But also a lot more on top so it's just a matter of limiting scope for now.

Given that I'm a student doing this in my spare time I need to limit scope to something thats acheivable in less than a year with the time I can put in.

But I should put this into the motivation.

> I would like a likelihood of 0 to mean that a branch is never expected to be called. (This differs from the proposal, which states that 0 would indicate rarity, not impossibility.) Perhaps (maybe with -O2 only?) GHC needn't promise to be semantics-preserving in likelihood-0 branches. In any case, even if we're too skittish to implement that today, by reserving 0 to mean "never", we can do so in the future.

If we allow people to shoot themselves people will shoot themselves. There is also the risk of naive users not reading the documentation and just playing around with it suddenly getting segfaults.

A better solution would be an {-# IMPOSSIBLE #-} pragma which defaults to `error "something"` unless explicitly unlocked. This could easily imply a likelyhood of 0 internally as well.
Then we could have a flag/extension like `-fyes-i-dont-need-my-foot` allowing the compiler to optimize away code paths marked as impossible.

> This proposal doesn't specify (I think) what happens when both a datatype and a case statement over that datatype have LIKELYs. Are they multiplied? Or does one overrule the other?

That's right I forgot to put that in. I had overrule in mind. It just makes sense to me and also follows the principle of least surprise.

> I think it’s pretty important to pin down what meaning this ranking/probability should have.

I'm hesitant to give precise guarantees/semantices. Given the nature of the idea 90% of the benefit can be had without them and giving one could restrict future implementation changes in ways I didn't think of yet.

> Granted this becomes a sort of whole program optimization.

Implementation detail, but for now I want to keep to tag assignment for simplicity.
Maybe it would be easy to decouple tag and constructor order. Maybe not. 
But that's not something I plan to look into for an initial implementation.

Even so we can do a lot just by fine tuning inlining and branching/layout.

> Granted this becomes a sort of whole program optimization. Which I guess
might be np hard or complete to do optimally if my recollection of
complexity results related for compression and or pattern compilation are
correct

Building a optimal decision tree is in NP at least, so that sounds right.

chessai:
I am strongly +1 on this. GCC also has likely/unlikely pragmas, though i believe they are less flexible than what's proposed here (that's good IMO).

cartazio:
@chessai  ... by definition there will always be a normalization.

I think for now rational numbers suffice, even though there is a possible story for log domain floats having a nice compositional semantics for nested case expressions (rescalling just being addition).

chessai:
Eh, i guess. I had a brief thought that likelihoods could be 0 (for
unlikely) and 1 (for likely). But this didn't make sense given the proposal

AndreasPK:
If there is a strong preference for allowing floating point weights in the Surface Syntax then I'm happy to change that detail. 

I don't think using a logarithmic scale in the surface syntax is worth it:
* It will no doubt confuse some people no matter how well we document it.
* I don't see a scenario where if would make a codegen difference if one branch is 100x as likely versus 1000x as likely.  But if someone feels strongly about this convince me otherwise.

> Unrestricted likelihood annotations could have serious negative impacts if a user makes a typo.

Can you expand on that?

> I second this. To me it seems unintuitive to NOT equate 0 with 'never'.

I don't disagree with that. But if we treat it as a guarantee then an optimisation that should be on by default will easily cause segfaults, even worse they might only occur at certain optimisation levels.

That would be unacceptable to me.

A solution similar like the IMPOSSIBLE pragma mentioned above seems like a far better solution for people wanting this behaviour.

jmct:
Some quick thoughts:

I feel strongly that we should avoid floating point values, unless you're enforcing that all the values must sum to 1.0 you'd have to normalize the values, which you can do with integer values as well (and if you did try to enforce that they sum to 1.0 I feel like that would be a usability nightmare, imagine removing a branch!)

> A better solution would be an {-# IMPOSSIBLE #-} pragma which defaults to error "something" unless explicitly unlocked. This could easily imply a likelyhood of 0 internally as well.

This is the (IMO) cleaner solution than making 0 mean 'never'. Just have all the likelihood values go from 1 on (like the Naturals for Mathematicians), and make never very explicit.

chessai:
> 
> > Unrestricted likelihood annotations could have serious negative impacts if a user makes a typo.
> 
> Can you expand on that?

Well, if we're normalising: in a computation with three branches with annotations {-# LIKELYHOOD 1 #-} {-# LIKELYHOOD 7 #-} {-# LIKELIHOOD 9 #-}, and a user meant to write '0' for the last likelihood, but (on a QWERTY keyboard) they hit '9', then their program's performance could potentially degrade.

Then again, these benefit of a likelihood annotation already rests so much on a programmer's judgment that maybe this is nitpicky. Also there's no way at all to tell if a user made a typo. I guess my thought was that if we just let users pick any natural they want then it's more likely they'll make a typo.

goldfirere:
I'm fine with `{-# IMPOSSIBLE #-}`. I just don't want `{-# LIKELY 0 #-}` to mean "rare". Maybe use `{-# UNLIKELY #-}` or `{-# LIKELY epsilon #-}` (where I'm sure folks will rightly criticize that last one for being too jargony).

cartazio:
@chessai : typoes will happen, and as has been discussed elsewhere,
ultimately this sort of framework should ultimately be machine instantiated
with profiling data, and otherwise it should really be treated as the
Performance analogue of unsafeAccursedINlineIO, which INLINE is too
honestly :)

cartazio:
@ Richard, at least as AndreasK plans to use this information / instrument
it, its intended to be used for smarter ordering of basic blocks and branch
targets, so while it clearly *can* be treated as 0 == zero probability, it
seems like its more likely going to be used as "I dont care, this has
lowest priority, do whatevers best for everyone else"

that is: its not likelihood so much as "unnormalized priority weight",

adding a special epsilon symbol that not zero but is smaller than any other
tends to mess with things like having a unique total order! (eg cf the
complex numbers  or  the
flavor used for deriving calculus where eps^2 == 0 )

one answer you might find to your taste, AND plays well with what i believe
is  the intent of Andreas' proposal:  allow any Integer,
and define the "normalized" weights by normW_i = Exp(w_i ) / SUM_k Exp(W_k)
, where W_k ranges over the integer weights given to the case branches

this would
  1) make it impossible under the "likelihood"/"probability" semantics to
write a case as having "probability zero", because we dont allow negative
infinity as a value in the Integers :)
  2) theres no longer a floor on how low relative priority / weight can go,
so i can always make sure a new case has even lower relative weight without
changing other cases. "AKA i can always choose a more negative integer", vs
bottoming out at zero
  3) if we have a nested case expression, that we want to transform into a
single unnested case expression, the new relative weights become the sum of
the weight on the outer case branch, plus the weight  of the corresponding
inner pattern .   aka we go from
```
case  x of -> p1  {w1} -> ... ;
  ...  ;
{ pk  {wk} ->
      case x@pk of pq1 {wq1} -> ..;  ...;  pq {wqh} -> ... }
```

becomes
```
case x of  pat1 {w1 } ->  ... ;
   ... ;
  patk.patq1 {wk + wq1} -> ...
 ..
  patk.patqh {wk + wqh} -> ..

```

or maybe we do want an expressible "least" element. Does the same apply
conversely? do we want an expresibl "top" element?

AndreasPK:
> I guess my thought was that if we just let users pick any natural they want then it's more likely they'll make a typo.

If we require numbers to sum up to a specific value then adjusting weights as well as removing/adding branches becomes a major annoyance though. So to me the cure for this is worse than the disease.

> ultimately this sort of framework should ultimately be machine instantiated
with profiling data,

I agree completely.

> I'm fine with `{-# IMPOSSIBLE #-}`. I just don't want `{-# LIKELY 0 #-}` to mean "rare".

Would you be fine with values <= 0 having the defined meaning of "Assume they are never taken but preserve semantics if the branch is taken."?

Otherwise we end up with programs segfaulting/throwing exceptions if a library author sets a weight wrong somewhere. Something I want to **really** avoid.

cartazio:
Andreas : ... I think the exp transform meaning for relative weight might
be better than all negatives being some “assume never”

If we have integers we can still have an ordering with special limits for
“this is the least likely branch ever” or “this is the most likely” and
still have a total order.  Aka the integers with  symbols limits at the
positive and negative ends

I’m just worried that the wrong weight model will bite us later.

cartazio:
Whatever format is chosen should support the actual internal representation, because it'll be darn hard to otherwise to eg "ship default branch weights " from some fancy throughtly PGO work to users otherwise, AND/OR easily have single file/small number of file reproducers for optimization bugs/regressions/testing.

Ignoring the concern of how to encode "this branch will have code generated, but its like the least important branch ever" encoding/syntax concerns, 

Theres two different surface syntaxes I think are worth considering/ implicit in this discussion

```
data RelativeRawWeights = SomethingShouldProveThisUnreachable | ReachableButBasicallyNeverHappens | RelativeWeight PositiveRational
```

```
data LogScaledRelativeWeights  = ShouldProveUnreachable | ReachableButNeverHappens | LogWeight Rational
```

either way, Rational weights, written in decimal/scientific style notation, seems like a sane choice to me.

gridaphobe:
from https://github.com/ghc-proposals/ghc-proposals/pull/182#issuecomment-441788303
> Would you be fine with values <= 0 having the defined meaning of "Assume they are never taken but preserve semantics if the branch is taken."?

I think this would be the best interpretation. The current formula for computing the likelihood of a branch clearly says that `{-# LIKELY 0 #-}` is expected **never** to be taken. But, like others, I would be more comfortable with a separate pragma that allows GHC to actually elide the impossible branch.

If we all agree on that, we might as well include the `{-# IMPOSSIBLE #-}` pragma in this proposal. It's a pretty small addition to the existing proposal.

AndreasPK:
> Can you give a sense of how much performance is being left on the table when GHC's heuristics make the wrong decision?

To come back to the question of performance. I now have a prototype implementation which showed a -3.2% runtime reduction on nofib.

This is mostly from setting some default weights for constructors on List, Maybe, Either and Ord. 
The patch also does some other things like trying to detect bottoming case alternatives and setting their likelihood to zero.

There are still things to implement and decision to be made (mostly about compile/runtime tradeoffs). But I  don't expect large movement in performance numbers from where I am now.

howtonotwin:
What’s the interaction with `RebindableSyntax` for `if _ then _ else _`? It seems to me the only sane thing to do is to simply error out when there’s a liklihood pragma present and the relevant syntax is rebound:

```
{-# LANGUAGE RebinableSyntax #-}
ifThenElse True = const
ifThenElse False = const id
-- should error:
test = if cond then {-# LIKELY 10 #-} 5 else 2
```

but it would be nice to have that explicit in the proposal.

goldfirere:
In answer to questions above: yes I'm fine with the current text about `{-# LIKELY 0 #-}`: it means a branch which is never taken but preserves semantics. I still want an `{-# IMPOSSIBLE #-}`, but that can wait for another proposal. 

Great question about `if` and rebindable syntax. I think usage of this feature in such scenarios should be banned, though it is tantalizing to think of labeling arbitrary expressions with likelihoods, where the number would be a percentage chance that the expression is evaluated. (Sadly, integers don't work there.) But that's maybe a challenge for another day.

What about pattern synonyms? I assume a pattern synonym would make a `case` not simple, but the proposal should say so.

AndreasPK:
> What about pattern synonyms? I assume a pattern synonym would make a case not simple, but the proposal should say so.

Indeed I've added a blurp about unsupported constructs/extensions listing them, rebindable syntax and a few more.

> it is tantalizing to think of labeling arbitrary expressions with likelihoods, where the number would be a percentage chance that the expression is evaluated.

I think that kind of information is better gathered via a profiling and then fed back to the compiler in an automated fashion.

nomeata:

AndreasPK:
I'm fine with marking this as dormant for the moment but for the record it's still on my agenda, but I'm up to other things at the moment so it will take a while to get a proper prototype.

erydo:
The lack of guided optimization is definitely a problem, but I feel this approach would be a misstep. Essentially what's proposed here is a limited kind of PGO (great!), but where the profile information data format is "handwritten annotations embedded into the source code" (less great).

I'm surprised that there are no alternatives listed. The discussion here, and the "Unresolved Questions" in the proposal, seems mostly centered around the syntax and implementation details rather than the the merit of the approach.

### Alternative

Even if the initial support for PGO in GHC were exactly the capabilities proposed here—weighted branch hints—a separate file with `branch-id, weight` would be significantly better in several ways. _To be clear, that's a straw-man data format, not a design._

<details>
<summary><i>Alternative²</i></summary>
  * Branch IDs could identified like `module, function, [branch-indexes…]` or some other kind of indexing.
  * Or they could be annotated in source and referenced. That would the make profiling data more resilient to refactoring changes. (Is that really desirable, though? ¯\\_(ツ)_/¯). But as noted by @howtonotwin, there isn't an easy correspondence between source syntax and actual branches/case splits.
</details>

Those weights could be produced/maintained trivially automatically, published in library packages, introduced to existing packages _without_ source modifications, and (in the future) developed into a more complete format, including things like: polymorphic branch weights, stack fingerprints, etc.

## Other reasons these should not be handwritten

The inline annotation approach can't be extended in those ways. They look like a one-off dead end. Furthermore, without profiling, most developers (especially those working on libraries) have no real idea where their hot paths are. This would be a temptingly easy knob to turn as a developer, and temptingly simple advice to give to a developer, but might cause pessimal optimizations for library consumers. And if you do have profiling results available, why not provide those directly to the compiler? Doing so would be far more likely to be accurate than encouraging developers to manually encode them in the source.

Providing the wrong tool is likely to be more harmful than no tool here, I think, for a few reasons:
  1. Opportunity cost. I think the alternative would require similar levels of effort, would provide the same capabilities, but has significant roadmap + pedagogy + ecosystem advantages.
  2. "Half-measure" cost. By providing and promoting this method of optimizing Haskell programs, people are less likely to pursue better solutions. Implementing this could actually set Haskell optimization tooling back.

AndreasPK:
> Even if the initial support for PGO in GHC were exactly the capabilities proposed here—weighted branch hints—a separate file with branch-id, weight would be significantly better in several ways. To be clear, that's a straw-man data format, not a design.

The reason why I do not favour an out of line definition of weights for the proposal is complexity.

The short version is that I think having a seperate file makes only sense if you auto generate it as they will be either closely coupled or far too complex to reasonably write by hand. In order to support hand written files out of line:  
* A format has to be specified.
* The format needs a parser, and all that comes with that.
* Haskell `source` for one module is now two files instead of one, which I imagine breaks all kinds of tooling.
* Weight definitions either don't deal with refactoring of source code making any code change two changes, or you have the HARD problem of defining a mapping of weights to branches, the later likely requireing source code changes anyway.

And that's just what came to mind immediately.

> including things like: polymorphic branch weights, stack fingerprints, etc.

I'm not familiar with what "polymorphic branch weights, stack fingerprints" are and couldn't find anything in a quick search. Could you expand on that? As for the other points:

> Those weights could be produced/maintained trivially automatically,

As long as GHC does not support generating the weights itself based on profiling data generating this data is anything but trivial. 
Adding support for such a feature would be great, but is simply beyond the scope of this proposal.

>  published in library packages, introduced to existing packages without source modifications,

I don't see how publishing becomes easier if we add more files. Given that weight files would essentially become source files given they have significant impact on compilation results avoiding source modifications is to me neither a big benefit nor likely.

> Those weights could be ... (in the future) developed into a more complete format,  
> The inline annotation approach can't be extended in those ways. They look like a one-off dead end. 

This proposal doesn't stop such development in the future. There is absolutely nothing about it which prevents us from later adding a way to specify a weight file which when given supersedes the weight information contained in the .hs file. In fact it's how I expect an eventual PGO implementation to work.

> Furthermore, without profiling, most developers (especially those working on libraries) have no real idea where their hot paths are. This would be a temptingly easy knob to turn as a developer, and temptingly simple advice to give to a developer, but might cause pessimal optimizations for library consumers. And if you do have profiling results available, why not provide those directly to the compiler? Doing so would be far more likely to be accurate than encouraging developers to manually encode them in the source.

While I understand your reasoning your assumptions are wrong.  

There are important cases where you can tell the likely branch even without profiling:  
* If I have a inner loop traversing over a data structure the likely case is that the current element is not the last one.
* If I check for an error the likely case is that there was no error.

I've seen high performance code optimize for this a few times in the past. Be it by [manually reordering Constructors](https://github.com/haskell/containers/blob/533c3c8d7d6df060d04468b8847de65ce04368c4/containers/src/Data/Set/Internal.hs#L116), using <= instead of > or other silliness. And every time I came across such things it was specific to certain GHC versions and a pessimization on others.

People already tune this knob. This proposal just adds labels to the knob and stops on becoming off randomly between GHC releases.

> Opportunity cost. I think the alternative would require similar levels of effort, would provide the same capabilities, but has significant roadmap + pedagogy + ecosystem advantages.

As the person planning on implementing this, and having done first steps in that direction I wonder how you arrive at this conclusion.  
As far as I can tell the alternative you refer to is as far as tooling/frontend is concerned essentially full PGO. Which means additional requirements to:  
* Expanding profiling capabilities to gather this data properly.
* Updating tooling to properly work with the additional files.
* Updating GHC to cross reference the additional data files with the code being compiled.  
Which is far from trivial and significantly more work.

> "Half-measure" cost. By providing and promoting this method of optimizing Haskell programs, people are less likely to pursue better solutions. Implementing this could actually set Haskell optimization tooling back.

I think this risk is very low. While this proposal does reduce the benefits of full blown PGO somewhat it also reduces the cost of implementation for it. Since most of the work done for this could be reused by a PGO implementation.

erydo:
Thank you for the very thorough and thoughtful response! (Sorry for replying out of order here).

> While I understand your reasoning your assumptions are wrong.
> There are important cases where you can tell the likely branch even without profiling

I think providing more information to the compiler is useful. And I understand (and agree) that a developer _can_ do those things; I didn't claim they can't. But, by far, most developers aren't able to do it well. Furthermore, even in the cases were developers do it well, they're not typically able to maintain those well. I would be very surprised if I were wrong about that, in aggregate.

> including things like: polymorphic branch weights, stack fingerprints, etc.

Sorry, I couldn't figure out a better terminology and didn't want to be too verbose. (To be clear, this isn't something I'd suggest initially. Just highlighting future considerations).
By "polymorphic branch weights", I meant: functions in Haskell are frequently polymorphic. Their branching behaviour can depend a lot on the particular instantiation of those types, not even just the values. But you cannot express that at a per-line level (except perhaps exclusively on concrete case splits), because each line can be interpreted in multiple independent ways in the same program. By "stack fingerprints" I meant something similar: "when this function is called from X it's likely to execute this way; when it's called from Y it's likely to execute this other way.". That's trivial to pull out of a profiling dump, but hard to represent in source code.

> The short version is that I think having a seperate file makes only sense if you auto generate it, as they will be either closely coupled or far too complex to reasonably write by hand.

That's the conclusion I came to also. In other words: I acknowledge that those files could become a pain to hand maintain, but I don't think it's worthwhile to consider handwritten inline hints as the primary use case. Code and performance characteristics change way faster than comments get updated. (Even as you illustrated in your example of pessimal optimizations). They should be generated. Though generating them wouldn't be a _necessary_ first-pass requirement, either, if the format is sufficiently simple.

> I don't see how publishing becomes easier if we add more files. Given that weight files would essentially become source files, given they have significant impact on compilation results. Avoiding source modifications is to me neither a big benefit nor likely.

One benefit is that as a consumer, you could choose to compile with hinted weights specific for your application. But a library could still provide default hints for its modules, helping most people in most cases. Unlike source annotations, where there would be (AFAICT) no customizing them without forking the source.

> As far as I can tell the alternative you refer to is as far as tooling/frontend is concerned essentially full PGO. Which means additional requirements to:

The suggestion was to implement the functionality as you proposed, except sourced from somewhere that isn't just parsing the AST. (Perhaps this is just harder that I think?). Regarding tooling updates, the minimum implementation would be at most one flag to GHC: "Read this file for branch-weight hints". There's no reason Cabal/Stack implementations etc would even care, I think?

One additional argument _against_ the separate file, though: GHC currently implements `INLINE` pragma this way, so there's some consistency to that approach. (Though, I'd have argued the same position if those were newly proposed, too. Library authors shouldn't unilaterally define how what optimizations are correct for their consumers).

nomeata:
What is the status of this PR?

AndreasPK:
> I'm fine with marking this as dormant for the moment but it's still on my agenda. I'm up to other things at the moment so it will take a while to get a proper prototype.

Sadly still true!

proposals/0000-likelihood-annotations.rst
```diff
@@ -0,0 +1,264 @@
...
+communicate to the compiler. This leads to missed optimization opportunities.
+
+Consider types like ``data Either a b = Left a | Right b``
+
+Left is often used to represent an exceptional case. Making Right the common case.
+
+Based on this information and given code like ``either f g x`` we might want
+to inline g but not f for example.
+
+The same logic applies to many other cases. For example Lists are rarely empty.
```
konsumlamm, line 32: I'm doubtful about the usefulness of annotating data types, especially if they're not for a very specific purpose. `Either` is not always used for exceptions, sometimes it's just used as a generic sum type. Likewise, there are many situations where lists can likely be empty, e.g. when passing a list of extra configuration options.

proposals/0000-likelihood-annotations.rst
```diff
@@ -0,0 +1,264 @@
...
+ - Data type declarations for Sum Types.
+
+Uses where the requirements are not satisfied will result in warnings similar to
+the ones from bad UNPACK pragmas. Likelihood values must be >= 0.
+
+Weight semantics
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+Given a simple branching expression with n branches ``[A1 .. An]``,
+with likelihoods ``[L1 .. Ln]`` GHC will optimize code under the assumption that
+the chance for the i-th branch to be taken is ``Li / sum [L1 .. Ln].``
```
konsumlamm, line 63: How would you decide on the likelihoods when one branch is more likely and the other is "normally" likely? 2 and 1? 10 and 1? 100 and 1? Perhaps it would be good to allow some branches to not be annotated and choose some default likelihood for them?

proposals/0000-likelihood-annotations.rst
AndreasPK: It's not useful for every type. But for some types it's really useful.

For streams defined via Done/Yield/Step `Done` will always be very unlikely. Just is generally more likely than Nothing. `:` is more likely than `Nil` and so on.

The alternative to annotating types like these would be us expecting the user to to annotate every use of the type.

Either was probably not the best example. But I do think it would be good to attach a weak preference to `Right` in the either case nevertheless.

proposals/0000-likelihood-annotations.rst
AndreasPK: I chose the meaning of weights to be relative execution frequencies. The downside to this is that the only proper way to know what weights to use is by knowing what "normally" likely and "more" likely is. Either through profiling or domain knowledge.

> Perhaps it would be good to allow some branches to not be annotated and choose some default likelihood for them?

I don't think so. If the code is performance sensitive enough for one to care about the weights it seems justified to expect weights for all branches. If the code isn't performance sensitive enough then it's likely better not to use weights at all.

For what it's worth the proposal talks about this further down. In particular not annotating a branch:

* Will trigger a warning if any of the other branches is annotated.
* The compiler is then free to choose any weight for this branch.

The reason why I believe this should be a warning is that I don't think the compiler can properly infer a "default" weight. If the other branches have weights [1,1,0,x] there is no easy way for the compiler to know what unknown branches should default to. The highest weight? The average? The median? The lowest weight? Lowest non-zero? All of these seem reasonable in some situations. I would probably have the compiler chose the median. But what if the compiler can tell the median is likely a bad choice by some further analysis?

Overall I think it's just simpler *and* better to require full weights and leave unannotated branches unspecified/up to the implementation at this stage.

PR #183, opened 2018-11-18
integer-openssl proposal

ch1bo:
Proposal to add a fast, BSD-licensed alternative for GHC's Integer type, next to currently available `integer-gmp` (default, fast and LPGL3-licensed) and `integer-simple` (slow, but pure and BSD3-licensed) implementations. Concretely, this would use OpenSSLs "BIGNUM" arbitrary size integer library contained in `libcrypto`.

[Rendered](https://github.com/ch1bo/ghc-proposals/blob/patch-1/proposals/0000-integer-openssl.rst)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -30,7 +30,7 @@ What is the timeline of a proposal?
    `How to bring a proposal before the committee <#how-to-bring-a-proposal-before-the-committee>`_ •
    `Who is the committee? <#who-is-the-committee>`_
 
-5. One committee member steps up as a shephard, and generates consensus within the commitee within four or five weeks.
+5. One committee member steps up as a shepherd, and generates consensus within the committee within four or five weeks.
 
    `Committee process <#committee-process>`_ •
    `Review criteria <#review-criteria>`_ •
@@ -208,7 +208,7 @@ What is a dormant proposal
 
 In order to keep better track of actively discussed proposals, proposals that
 see no activity for an extended period of time (a month or two) might be marked
-as “``dormant``”. At any time the proposor, or someone else can revive the
+as “``dormant``”. At any time the proposer, or someone else can revive the
 proposal by picking up the discussion (and possibly asking `the secretary
 <#who-is-the-committee>`_ to remove the ``dormant`` tag).
 
@@ -217,7 +217,7 @@ You can see the `list of dormant proposals <https://github.com/ghc-proposals/ghc
 Who is the committee
 --------------------
 
-The current members of the GHC steering committe, who you can reach
+The current members of the GHC steering committee, who you can reach
 by email at ghc-steering-committee@haskell.org, are:
 
 ======================  ====================================================  =========
@@ -247,7 +247,7 @@ Committee process
 -----------------
 
 The committee process starts once the committee has been notified that a
-proposal is ready for decistion, and takes place on the
+proposal is ready for decision, and takes place on the
 `ghc-steering-committee <https://mail.haskell.org/cgi-bin/mailman/listinfo/>`_
 mailing list. All interested parties are invited to follow the discussion.
 
@@ -258,8 +258,8 @@ mailing list. All interested parties are invited to follow the discussion.
    makes a recommendation as to whether the proposal ought to be accepted,
    rejected or returned for revision.
 
--  Discussion among the committee ensues on the mailiing list.
-   Silence is undestood as agreement with the shepherds recommendation.
+-  Discussion among the committee ensues on the mailing list.
+   Silence is understood as agreement with the shepherds recommendation.
 
 -  Ideally, the committee reaches consensus, as determined by the secretary or
    the shepherd.  If consensus is elusive, then we vote, with the Simons
@@ -288,7 +288,7 @@ mailing list. All interested parties are invited to follow the discussion.
 
    *  **If we say yes:**
       The shepherd or the secretary announces this on the pull request
-      and lables it as
+      and labels it as
       `Accepted <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3AAccepted>`_.
       The secretary merges the pull request and assigns the final proposal
       number.
@@ -298,7 +298,7 @@ mailing list. All interested parties are invited to follow the discussion.
       oversee implementation, attract implementors, etc.
 
       The proposal authors or other implementors are encouraged to update the
-      propsoal with the implementation status (i.e. trac ticket number and the
+      proposal with the implementation status (i.e. trac ticket number and the
       first version of GHC implementing it.)
 
 
@@ -317,7 +317,7 @@ and any other relevant considerations appropriately.
    ideally offer evidence of some form.
 
 -  *Elegant and principled*. Haskell is a beautiful and principled
-   langauge. It is tempting to pile feature upon feature (and GHC
+   language. It is tempting to pile feature upon feature (and GHC
    Haskell has quite a bit of that), but we should constantly and
    consciously strive for simplicity and elegance.
 
--- a/proposals/0000-integer-openssl.rst
+++ b/proposals/0000-integer-openssl.rst
@@ -39,16 +39,16 @@ implementation of the ``GHC.Integer`` interface and is `BSD3
 slow and infeasible for contexts where big integer multiplication / division are
 in fact required, e.g. public key cryptography.
 
-So, the goal is to have a fast (enough), BSD3-licensed alternative for the
+So, the **goal** is to have a fast (enough), BSD3-licensed alternative for the
 ``Integer`` data type available in GHC.
 
 After looking for a BSD-licensed library, ``openssl`` came to mind, which
 implements arbitrary size integer arithmetic in it's `BIGNUM
-<https://github.com/openssl/openssl/tree/master/crypto/bn`_ library incorporated
+<https://github.com/openssl/openssl/tree/master/crypto/bn>`_ library incorporated
 into ``libcrypto``. It seems to ticks all boxes:
 
 * BSD-licensed: The OpenSSL license is a `BSD-style license
-  <https://tldrlegal.com/license/openssl-license-(openssl)`_.
+  <https://tldrlegal.com/license/openssl-license-(openssl)>`_.
 
 * Complete / correct: Documentation and source seems to cover all major
   operations on big integers - so only a thin layer / wrapper is suitable.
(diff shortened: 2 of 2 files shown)


simonpj:
Sounds like a good work to me.

But to an external observer it's very mysterious that you propose to call it `integer-openssl`.  That makes it sounds as if this library is something to do with crypto etc.  But I assume it's not.   I'm guessing it's because the impl uses code from openssl?

Anyway, some less confusing name might be worth considering.  Or not.  Just raising it because it took me some time to realise that this proposal was not related to crypto

cartazio:
It’s also worth noting that at least at icfp etc, there was some discussion
with Ben in investing effort

As for performance : this comparison Information and analysis needs to be
part of this proposal.  Because that’s it’s only value prop vs integer
simple.  If you need help doing that , great. But the open ssl big number c
code hasn’t had tuning in 20 years.  So you really need to do a systematic
perf comparison.

cartazio:
* at icfp Ben and I spoke about actually doing a systematic investigation
of current bsd friendly big number libs to see if any could replace gmp for
ghc as the default and keep performance parity or better on small and
asymptotic perf both.

piyush-kurur:
One of the suggested reasons for inclusion is to "allow public key algorithms". Of course you can
use that in the class room but not for serious crypto.

1. Haskell requires a integer library because it provides large arbitrary integer multiplication. But
it would be dangerous to implement pki (I am assuming RSA) based on that. A better motivation
could be representing integers in compilers.

2. For crypto purposes it would be sufficient (and better) to implement say 4026 or some fixed size integers carefully done so that there are no side channels etc.

3. More importantly if ghc has to depend on openssl that would be a pity. I do not think it is a good code base to depend on plus it bring out a lot of other useless (for ghc) stuff.

cartazio:
I agree with the spirit of what Piyush says :

Additionally dragging a big project like OpenSSL into ghc build deps might
have a huge cost.  And for big number math per se we already have great
code.

Additionally ,
https://ghc.haskell.org/trac/ghc/wiki/Design/IntegerGmp2 paves a path
towards allowing the big num lib to be to be selected in userland , though
I think Herbert might appreciate some help pushing that along. (The gmp2
design appeal is that it moves Any new big num stuff into being feasislvd
in user space )

I guess

ch1bo:
@piyush-kurur The cryptography example was given as the haskell package `cryptonite` implements public key cryptography primitives in haskell. And IIRC it is using `Integer` and operations on it. These were particularly slow in a real world scenario using `integer-simple`.

@cartazio Comparing the performance is definitely important and I tried to get early estimates how much faster than integer-simple and also how much slower than integer-gmp the prototypical implementation of multiplication and division is, I'll add these preliminary results to the PoC and link them in the proposal.
The improvements outlined in https://ghc.haskell.org/trac/ghc/wiki/Design/IntegerGmp2 are already incorporated into integer-gmp and were also considered when creating the current proof of concept of integer-openssl.

In general I'd like to say that I took the BIGNUM implementation of openssl as it was a) BSD licensed, b) complete, c) correct and d) likely available on most systems. If we find better alternatives, I am happy to update the proposal, name and implementation to use something else :)

bgamari:
My apologies if this was already clear: I think it would make the most sense to consider this as a more efficient alternative to `integer-simple`, not a replacement of `integer-gmp`. GMP seems to be the gold-standard of bignum libraries and I see no reason not to use it unless licensing concerns preclude its use.

I believe this is what the proposal as-written is proposing. Consequently, I don't think the concern that GHC will gain a dependency on OpenSSL is well-founded.

23Skidoo:
There's also https://github.com/libtom/libtommath, I wonder how it compares to OpenSSL's bignums.

cartazio:
I started a look at libtommath. But I need to fork it to make the oss
license less asnine.  I tried to reach out to the author to get it fixed
upstream and that didn’t go anywhere.  Happily it’s a license that at least
in the US is easy to fork into a sane one.

23Skidoo:
It seems to be dually licensed now, the first one is the asinine license you mention, and the other one is public domain (like SQLite).

cartazio:
Confusingling enough: public domain as a legal construct doesn’t exist is
every country. So you still want a more explicit license grant.  If you
look at the homepage for SQLite, you’ll see they sell licenses for the
occasional paranoid corporate user of SQLite that basically promises they
will vigorously defend their right to give the customer the code as public
domain.

It’s a confusing footnote. But an important one none the less.  Maybe I’ll
have a read of the lib Tom stuff this holiday.

ch1bo:
libtommath looks promising as well. After a quick glance over the header, all required ops should be there.. all but one I could not find in openssl/bn so far either: integer <-> double conversion. I am not too good in maths, but in a quick chat with @hvr on monday, he mentioned that one probably does not want to do that by hand.

Other than that, also got updated to what the original intention by @hvr with integer-gmp and link time selection of an integer library was. I'll try to incorporate that into the proposal 

Also, the https://github.com/libtom/libtommath/blob/develop/LICENSE is kinda hilarious!

minad:
@ch1bo I also suggest libtommath, in particular since it can be vendored easily into the package or GHC by copying a single c file. Concerning integer<->double conversion, I made https://github.com/libtom/libtommath/pull/123 a while ago, which is not yet part of tommath unfortunately.

piyush-kurur:
@ch1bo thanks for the response. Also @bgamari thanks for clarifying that it is not
to replace gmp. Rereading the proposal the misunderstanding was entirely my fault.
However I am uncomfortable with having an indirect suggestion of using the multi-precision
Integer at the haskell level for cryptography. Even when the underlying library is designed for crypto
I think it is tricky as I am not sure how ghc moves around large integers in its memory during GC.

cartazio:
@minad if you could help shepard that license clarification into being in master that'd be lovely :)

cartazio:
(i'm a little bit too spread thin to do more than be a resource on the sidelines atm :) )

minad:
@cartazio I will see what I can do regarding libtommath

Ericson2314:
We really ought to us backpack for `integer-*`. I'd hope with 3 implementations it would finally be worth someones time to do so.

cartazio:
Oh: for a common signature they can all instantiate?

ch1bo:
@Ericson2314 yes, this should be a perfect fit for backpack. It was also brought up by @hvr as a suggestion.

I am pretty new to backpack, but I should incorporate it into the proposal..

minad:
Tommath changed the license to Unlicense, which is essentially public domain and similar to MIT in jurisdictions which so not recognize the concept of public domain.

Furthermore additional functions were added (float conversion,...), which are also useful for a hypothetical integer-tommath.

https://www.libtom.net/news/LTM_1.1.0RC1/

piyush-kurur:
On Mon, Jan 07, 2019 at 09:28:03AM -0800, Alec Theriault wrote:
> > @Ericson2314 yes, this should be a perfect fit for backpack. It was also brought up by @hvr as a suggestion.
> > 
> > I am pretty new to backpack, but I should incorporate it into the proposal..
> 
> Since I tried this recently, here's some problems that will have to be solved:

> [snip]
>

>   * `integer-simple` and `integer-gmp` don't expose quite the same
>      functions ([I'm working on
>      that](https://gitlab.haskell.org/harpocrates/ghc/commit/d0d1fd61018296c104811e06005db251e6075a8f))

I will be happy to contribute whatever little I can if there is some
one else leading the backpaciksation.

minad:
Would this project make sense in the context of the GSOC? See https://www.reddit.com/r/haskell/comments/ae1i6z/reminder_please_submit_gsoc_ideas/edmpgit

jaspervdj:
Happy to have the implementation part of this as a GSoC project.  If anyone here is interested in mentoring, please let me know.

piyush-kurur:
On Tue, Jan 15, 2019 at 03:38:16PM +0000, Jasper Van der Jeugt wrote:
> Happy to have the implementation part of this as a GSoC project.  If
> anyone here is interested in mentoring, please let me know.

I can contribute towards mentoring but might not be able to put enough
time in the process. In any case if there is a successful GSoC project
please keep me updated.

nomeata:

ch1bo:
> Since I tried this recently, here's some problems that will have to be solved:
> 
> * there's a bunch of CPP'ed optimizations in `base` (as well as a couple in `text` and `bytestring` I think?) which require intimate knowledge of `Integer`'s particular constructors
> * there's a some GHC code that peers into the internals of `Integer`'s particular constructors
>   
>   * there's a handy core-prep optimization for turning small integers directly into calls to `S#` instead of going via `mkInteger`
>   * there's some GHCi code for cleverly printing integers with `:sp`
>   * ????
> * `integer-simple` and `integer-gmp` don't expose quite the same functions ([I'm working on that](https://gitlab.haskell.org/harpocrates/ghc/commit/d0d1fd61018296c104811e06005db251e6075a8f))

@harpocrates Thanks for that - this is already merged right?

> Ideally, we'd have `GHC/Integer.hsig` as a signature in base (and maybe even `GHC/Natural.hsig`?) and downstream consumers could instantiate `base` with `integer-*`, but it feels like there will be some work to get there.

Although I am a backpack noob, I gave this a try (also a while back) in here https://github.com/ch1bo/backpack-integer

However with the current state of backpack adoption a first iteration without backpack sounds more realistic?

nomeata:

ch1bo:
Finally created a MR for GHC here: https://gitlab.haskell.org/ghc/ghc/merge_requests/1677. This adds `integer-openssl` in the same way as `integer-gmp` - some cleanup and possibly DRYing with `integer-gmp` could/should be done - feedback welcome.

hsyl20:
I have been working on a new `ghc-bignum` package replacing both `integer-gmp` and `integer-simple` and hopefully available in GHC 8.12. Among other things, it makes the implementation of newer Integer/Natural backends way easier as each backend only has to provide a few functions: https://gitlab.haskell.org/ghc/ghc/merge_requests/2231

@ch1bo if you could try to add `libcrypto` as a backend for `ghc-bignum`, that would be great. You don't have to change anything in `bytestring`, `text` and so on anymore. I'd be glad to help as needed.

ch1bo:
@hsyl20 Your work on `ghc-bignum` looks very promising, nice! It should be quite trivial to adapt https://github.com/ch1bo/integer-openssl to a backend of your `ghc-bignum` package. Is the package also available "outside" of GHC? i.e. with a test-suite or benchmarks

hsyl20:
> Is the package also available "outside" of GHC?

Sadly not anymore. It was initially in https://github.com/haskus/packages/tree/master/haskus-integer with some tests but a lot of things have changed since then (bug fixes, etc.).

>  with a test-suite or benchmarks

I have [some benchmarks](https://github.com/hsyl20/integer-bench) which use `criterion-measurements` from [here](https://github.com/hsyl20/criterion/tree/hsyl20-no-deps) (to avoid any dependency on non boot packages).

I still have to integrate the QuickCheck tests into GHC's testsuite. Last time I tried, installing QuickCheck for the tests seemed to be a pain.

PR #185, opened 2018-12-03
Function return type signatures

timw-da:
This is a proposal to augment the existing `ScopedTypeVariables` extension to permit in-line function return type signatures, which are already recognised by the Happy grammar, but currently rejected. Such signatures can easily be given meaning and do permit very concise in-line type signature definitions.

[Rendered](https://github.com/timw-da/ghc-proposals/blob/patch-1/proposals/0000-function-return-type-signatures.rst)

### DIFF
--- a/proposals/0000-function-return-type-signatures.rst
+++ b/proposals/0000-function-return-type-signatures.rst
@@ -1,0 +1,82 @@
+Function Return Type Signatures
+===============================
+
+.. proposal-number:: Leave blank. This will be filled in when the proposal is
+                     accepted.
+.. trac-ticket:: Leave blank. This will eventually be filled with the Trac
+                 ticket number which will track the progress of the
+                 implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+
+Motivation
+------------
+
+With the ScopedTypeVariables extension enabled, GHC already supports in-line type signatures during function definition, for example:
+
+::
+
+  f (x :: Int) = g x :: String
+
+Many users coming from Scala, F#, OCaml, etc, prefer such inline signatures and would expect to be able to provide a return-type signature on the left-hand-side. For example:
+
+::
+
+  f (x :: Int) :: String = g x
+
+Such a signature would also mean only needing to specify the type annotation once when there are multiple right-hand-side equations. For example:
+
+::
+
+  f (x :: Int) :: String
+    | p x = g x
+    | otherwise = msg
+
+We propose an extension to allow the above syntactic form. It is a fairly small low-risk change, requiring no changes to the Happy grammar and parser. It provides an arguably more familiar and more succinct syntax for specifying type signatures.
+
+
+Proposed Change Specification
+-----------------------------
+
+The Happy grammar for a top-level pattern or function bind, already specifies an optional signature:
+
+::
+
+  decl_no_th
+    ...
+    | infixexp_top opt_sig rhs
+
+Currently a validation step fails if this optional signature is specified in addition to one or more pattern bindings. The proposal here is for us to not reject such a parse, but instead give it meaning.
+
+The semantics are such that any return-type signature provided on the left-hand-side is semantically equivalent to a user provided expression signature applied to the right-hand-side body.
+
+We propose that this change is enabled whenever ScopedTypeVariables is enabled; and is just a syntactic change.
+
+Effect and Interactions
+-----------------------
+
+Because the extension is a simple syntactic addition which can be easily expressed using existing features, I do not expect any unintended effects or interactions.
+
+Costs and Drawbacks
+-------------------
+
+The development/maintenance cost is small, but not insignificant as we will have to add an extra field into one of the front-end AST types.
+
+Pattern signatures currently can only introduce rigid, fully-known type variables; and we are somewhat further encouraging their use with this extension (see https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#pattern-type-signatures). However, OCaml behaves in a similar fashion for such inline signatures.
+
+Alternatives
+------------
+
+The alternative is to add an explicit expression signature on the right-hand-side of the function body, but this is counterintuitive to many beginners and new users from other languages. It also results in more boilerplate when using multiple right-hand-side expressions.
+
+Implementation Plan
+-------------------
+
+The proposal authors will implement this change.


AntC2:
Ooh yes!

Presumably this allows binding tyvars; where do constraints go?:

    f (x :: a) :: Num a => Maybe a = Just x

Hmm. That's not pretty. But then neither is:

    f (x :: a) = Just x :: Num a => Maybe a 

The main motivation I see (I'm not coming from Scala/etc) is to avoid using `ScopedTypeVariables` with that daft rule about needing explicit `forall`. That applies particularly inside `where` decls attached to a function/method.

timw-da:
Yes this would allow tyvars bound via pattern signatures as in your example, the only issue is that these are bound rigid (as discussed in the proposal). Thus we may want to permit an inline explicit forall somewhere as part of this proposal, or perhaps change the default binding behaviour for such pattern signatures under certain circumstances. Interestingly OCaml also requires explicit forall bindings for such inline signatures, if universal quantification is required.

Ericson2314:
I believe there is already a proposal to relax the rigidity requirement.

AntC2:
> @goldfirere  I propose: nowhere. [for constraints]

?So this proposal is not as conceptually simple as shunting the signature from the end of the rhs to the end of the lhs?

[Some personage is already complaining](https://ghc.haskell.org/trac/ghc/ticket/15810#comment:9) about the "veritable zoo of special cases". Is this going to be a different species of duck-billed platypus?

howtonotwin:
No, it *is* as simple as shunting the signature from the RHS to the LHS. You cannot add constraints to the type annotation of a RHS today, and you will not be able to add those same constraints on the LHS tomorrow. Using the above example,

```haskell
f (x :: a) = x + 1 :: a
```

is not a complete type signature today, and GHC will infer the `Num a` constraint into `f`'s type. More than that, there is no place to actually write the `Num a` in the first place. It is also illegal to try polymorphic recursion like:

```haskell
f (x :: a) = (if x == 0 then x else fromInteger (f (5 :: Integer)) + x) :: a
```

because that is only possible if there is a complete type signature. Instead, you must write

```haskell
f :: (Num a, Eq a) => a -> a
f (x :: a) = (if x == 0 then x else fromInteger (f (5 :: Integer)) + x) :: a
```

Similarly,

```haskell
f (x :: a) :: a = x + 1
```

has no complete type signature, and the constraint must be inferred, as there is no place to write it, and, again

```haskell
f (x :: a) :: a = if x == 0 then x else fromInteger (f (5 :: Integer)) + x
```

is rejected, as polymorphic recursion is a no-no.

simonpj:
I can see that this proposal would sometimes be convenient; but I'm not sure that it fully pays its way.

If adopted, I strongly urge that

* **This result-signature stuff should NOT have any influence on the notion of a definition having a "complete type signature"**.   Today f has a complete type signature if it has a separate type signature, and that type signature is no partial (i.e. no underscores).  Let's stick with that.  We are about to add type-level kind signatures precisely because separate type signatures have proved so valuable for terms.

* **Treat the result signature EXACTLY like a pattern signature**.  That is, it is not implicitly quantified, and may bring new type variables into scope.

* **Write out the typing rules formally**, using the framework of our Haskell Symposium paper: "Type variables in patterns", Haskell Symposium 2018

Ericson2314:
@AntC2 @goldfirere wrote the top-level signatures proposal, so I think that trac comment is just agreeing with @simonpj's point that we should have no more CUSKs.

I'd like to emphasize that doing this proposal while not adding CUSKs not only avoids problems, but also makes the language more consistent in accordance with the top-level sigs proposal. That would create near-complete "inline signatures" for all declaration types which I believe is a good thing to be consistent about: the user should need true top-level sigs for precisely signature completeness and not anything more (not being able to specify the return type otherwise) or less (there is some CUSK thing that also suffices).

AntC2:
> @howtonotwin You cannot add constraints to the type annotation of a RHS today, ... Using the above example,
>
>        f (x :: a) = x + 1 :: a

I'm confused. Yes for that signature, GHC infers the `Num a`. So I should be able to explicitly give a constraint that GHC infers anyway(?) Today GHC 8.6.1 with `ScopedTypeVariables` just accepted

    f (x :: a) = x + 1 :: Num a => a

(I have no other signature for `f`; so `a` is indeed a new type variable coming into scope.)

>  there is no place to actually write the `Num a` in the first place. 

Em, I seem to have put `Num a` into a place. Indeed I can do this 

    g (x :: a) = show $ x + 1 :: Num a => String

(That annotation looks weird. GHC infers `g :: (Num a, Show a) => a -> String`.)

I defer to the experts on whether such sigs should be considered complete/banning polymorphic recursion/CUSKs/etc.

AntC2:
>  so it's unsurprising that GHC can infer a `Show a` constraint even when you didn't write one.

Good. Yes. Quite. Thank you for your "yes". I was getting too many "no"es -- I suspect in answer to questions I did not ask and did not intend and would not understand anyway.

So please could the proposal include the following examples and say they're to be valid, so people can assess whether they're too weird/whether the extra syntax "pays its way":

    f (x :: a) :: Num a => Maybe a = Just x

    g (x :: a) :: Num a => String = show $ x + 1

And point out that GHC is to infer `g :: (Num a, Show a) => a -> String`, as currently for the annotation appearing on rhs.

With @nomeata's recent proposal to revive `PatternSignatures` as a separate extension, I was at first sceptical. But I was won round: the full-bore `ScopedTypeVariables` is just too hard. That Users Guide section 10.17.4 I understood at the time on the 4th or 5th reading; but looking at it again just now, I had to start over. It just seems easier if the pattern signature "brings the type variables into scope", and the return type signature then uses those variables, rather than all that dicking about with explicit `forall`.

Then I suggest this proposal is enabled by `PatternSignatures` alone. (And the full `ScopedTypeVariables` implies `PatternSignatures`, so also enables this proposal.)

AntC2:
>  there is a very big difference between
>
>        f1 x :: a -> a = blah
>        f2 x = blah :: a -> a

The former is not currently valid, so how can you claim what it means or what it's different from? What rule are you using? The proposal says that a return type sig on lhs is to mean the same as if it appeared on rhs. Treat that as purely syntactic sugar (to avoid the programmer repeating the sig on multiple rhs equations, and to show it more 'in your face').

> But I propose that ...

Stop proposing. If a lhs sig is going to mean something different to a rhs, I see no merit here.

> Putting any ... constraints on these pattern-bound types gets you into higher-rank types. 

Does it? @goldfirere just said it doesn't. Or by "pattern-bound types" do you mean something different to this (with a constraint, but not taken as a "complete type signature"):

    g (x :: a) :: Num a => String = show $ x + 1

You perhaps mean where a var is bound purely by a return type sig, not also bound to an argument? Then does it need a rule: return type signatures must not introduce variables? Please be clear.

The proposal section 4 says

> Pattern signatures currently can only introduce rigid, fully-known type variables; and we are somewhat further encouraging their use with this extension ...

It's not clear from the text if this is only a description, or is intended as a prescription/restriction.

AntC2:
> Why should the meaning be the same?

Because this proposal says that's what it wants, following the lead from other languages (section 2 to 3):

>> We propose that this change ... is just a syntactic change.

>> Because the extension is a simple syntactic addition which can be easily expressed using existing features, I do not expect any unintended effects or interactions.

> @int-index doing anything else would be neither consistent nor terribly useful.

(You seem to be using a different sense of "consistent". I guess: what is consistent with which? I.e. not consistent in this sense:)

> @howtonotwin  it _is_ as simple as shunting the signature from the RHS to the LHS.

Then section 3's (and @howtonotwin's ) claims are false. Then this proposal is terrible and non-useful: kill it.

int-index:
> You seem to be using a different sense of "consistent". I guess: what is consistent with which?

It is consistent with the way `ScopedTypeVariables` work. Type annotations in patterns and in expressions behave differently when it comes to type variables, and the question is:

* Do we have pattern-like behavior in function return type signatures?
* Do we have expression-like behavior in function return type signatures?

Since it's on the LHS of `=`, we probably want pattern-like behavior.

int-index:
> Then this proposal is terrible and non-useful: kill it.

I don't get where the "terrible" claim comes from, but the "non-useful" is easy to counter: it becomes useful for everything that the proposal currently claims *and* for binding parts of the return type to type variables.

shayan-najd:
cc: @goldfirere @simonpj 

This is just a side note related to this proposal to point out some related decisions in the design space. In dependently-typed programming, where GHC Haskell is moving towards to, it is often desired (although languages like Agda lack this) to let named arguments in types (to indicated term dependencies of types) to also be accessible in terms to avoid duplication. For example, in a dependent language, say Agda, we may have
~~~
f : (size : Nat) (vec : Vec size Char) ->  Vec size Char
f size vec = M 
~~~
Notice the duplication in naming the arguments (the first naming is mandatory for stating the term dependency of types).
Instead, one may want to write in the following "inductive definition" style:
~~~
f  (size : Nat) (vec : Vec size Char) 
   : Vec size Char
   = M
~~~
How the above style interacts with induction / pattern matching is an important point to consider.

Icelandjack:
So when do we allow 

```haskell
data Ty :: Type
  = TFree Name
  | Ty :-> Ty
```

nomeata:

int-index:
@timw-da Do you plan to submit this to committee?

timw-da:
Given the discussion so far, Digital Asset will reluctantly withdraw the proposal in its current form but would welcome any other contributors that might wish to take it forward.

nomeata:
Ok, will close this, but encourage anyone who wants to pick it up again to reopen.

int-index:
Thanks for the information. I will write and submit my version of the proposal.

int-index:
The new proposal is at https://github.com/ghc-proposals/ghc-proposals/pull/228

PR #186, opened 2018-12-06
GRC: Improve language for Stack code of conduct

spl:
1. Change “similar motivations ours” to “a motivation similar to ours”
2. Use language that does not require knowing the time of publication

PR #187, opened 2018-12-06
similar motivations ours → similar motivations to ours

miangraham:
@nomeata Quick fix!


nomeata:
Thanks! Already part of #186, though :-)

PR #188, opened 2018-12-07
Replace "explicit code of conduct" with "set of guidelines for respec…

Cipherwraith:
As per Simon Peyton Jones' email sent out to the "GHC-Devs" mailing list on December 6, he says that he wants to avoid calling this a "Code of Conduct" and prefers to refer to it as "guidelines for respectful communication" because "[he wants] to encourage good communication, rather than focus on bad behaviour".
Therefore, I propose updating this file to replace "explicit code of conduct" with "set of guidelines for respectful communication".

### DIFF
--- a/GRC.rst
+++ b/GRC.rst
@@ -15,7 +15,8 @@ standard, in the hope that others may voluntarily follow suit.
 Motivation
 ----------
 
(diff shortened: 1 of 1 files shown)


simonpj:
Thanks!  Done.

PR #192, opened 2019-01-08
Allow underscores in DerivingVia

Icelandjack:
Current work in progress, I invite others to propose a good way to get from `A` to `B`.

I want to allow the user to write

```haskell
 newtype F a = MkF (StateT Int IO (Maybe a))
   deriving
    (Functor, Applicative)
  via
    (_ `Compose` _)
```

and somehow infer `deriving .. via (Compose (StateT Int IO) Maybe)`.

* [Rendered](https://github.com/Icelandjack/ghc-proposals/blob/patch-4/proposals/0000-DerivingVia_underscore.rst)

### DIFF
--- a/proposals/0000-DerivingVia_underscore.rst
+++ b/proposals/0000-DerivingVia_underscore.rst
@@ -19,7 +19,7 @@ Is there a way to write
 
 ::
 
- newtype XX a = X a
+ newtype X a = X a
 
  instance Show (X a) where
   show :: X a -> String
--- a/proposals/0000-DerivingVia_underscore.rst
+++ b/proposals/0000-DerivingVia_underscore.rst
@@ -5,9 +5,7 @@
 .. trac-ticket:: 
 .. implemented:: 
 .. highlight:: haskell
-.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
-            **After creating the pull request, edit this file again, update the
-            number in the link, and delete this bold sentence.**
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/192>`_.
 .. sectnum::
 .. contents::
 
--- a/proposals/0000-DerivingVia_underscore.rst
+++ b/proposals/0000-DerivingVia_underscore.rst
@@ -1,0 +1,86 @@
+`DerivingVia` underscore
+==============
+
+.. proposal-number:: 
+.. trac-ticket:: 
+.. implemented:: 
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+This is an idea I haven't fleshed out, but I have a feeling there is a sensible way to use underscores `_` to avoid repeating yourself in `via` types.
+
+
+Motivation
+------------
+
+Is there a way to write
+
+::
+
+ newtype X a = X a
+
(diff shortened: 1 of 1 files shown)


aspiwack:
Something along these line would indeed be desirable.

When there is a single type constructor, like in your example, and this type constructor is a newtype. It is tempting to unify the right-hand side of the newtype's definition with the type being defined. (well, the right-hand side of the type being defined if it's a newtype…).

So a first attempt is to unfold all the newtypes we see once (including the type being defined), then unify, and require further annotations if it doesn't work.

It's pretty sketchy, though. And I feel there must be a better way.

Note:

- “Similarly (that's where the idea started)”: the link points to the root of haskell.org (which is probably not the intended target)

goldfirere:
Either this will have a bit of a wonky specification, or it won't work at all.

Your example is

```
newtype X a = X a
instance Show (X a) where ...
data Boolean = T | F
  deriving Show via (X _)
```

and you would like GHC to infer that `_` stands for `Boolean`. But it needn't stand for `Boolean`. It should stand for `X Boolean` (giving you `deriving Show via (X (X Boolean))`), which is just as valid. Or perhaps there are other newtypes in scope that could be arbitrarily inserted. I think you want GHC to infer the "minimal" choice for `_`... but what does "minimal" mean? (This is the wonky specification route.)

nomeata:
For the `Boolean` case it might be sufficient to introduce a syntax that stands for “the type being defined”, say `…`, to make the idiom

    data Foo = Bar deriving Class via (Newtype Foo)

have less repetition (essentially of the type has a long name, many parameters, or many such classes):

     data Foo = Bar deriving Class via (Newtype Foo)
  
. But it does not help with the former case.

goldfirere:
@nomeata's idea is indeed practicable (though I would advocate for `..` instead of `...` to avoid a new lexeme), but I personally don't think it's worth adding.

int-index:
For the `Boolean` case:

```
data Boolean = F | T
 deriving Show via wrapper X
```

where `via wrapper` is a deriving strategy that desugars into

```
data Boolean = F | T
 deriving Show via X Boolean
```

For the `F` case, 

```
 newtype F a = MkF (StateT Int IO (Maybe a))
   deriving (Functor, Applicative) via (Compose _f _g a)
```

We unwrap the newtypes in the `via` strategy as much as we need in order to unify, so we get

```
(_f (_g a)) ~ StateT Int IO (Maybe a)
--------
_f ~ StateT Int IO
_g ~ Maybe
```

and therefore derive via `Compose (StateT Int IO) Maybe`.

goldfirere:
@int-index suggests a concrete strategy... but it includes "We unwrap the newtypes". What if there is more than one newtype at some point during the unwrapping? Which one do we unwrap first? And what if this choice matters?

I'm sure we could come up with a set of rules to answer these questions unambiguously, but then those rules would very much be user-facing. This is the "wonky specification" route, to me.

mikeplus64:
Tangentially, has any thought been given to actually being able to specify the function to use for wrapping/unwrapping? e.g.

```
data Foo a = Foo | Bar | Baz a
  deriving (Monad, Functor, Applicative) via iso fromFoo toFoo

fromFoo :: Foo a -> Either Bool a
fromFoo Foo = Left False
fromFoo Bar = Left True
fromFoo (Baz a) = Right a

toFoo :: Either Bool a -> Foo a
toFoo (Left False) = Foo
toFoo (Left True) = Bar
toFoo (Right a) = Baz a
```

mikeplus64:
I guess that, to work, would require some spooky looking magic, like generating one-off `Coercible` instances from the from/to functions, think `fromFoo :: Dict (Coercible (Foo a) (Either Bool a))` `toFoo :: Dict (Coercible (Either Bool a) (Foo a))`...

nomeata:

nomeata:

PR #193, opened 2019-01-14
Make `forall` a keyword in types

goldfirere:
[Rendered](https://github.com/goldfirere/ghc-proposals/blob/forall-keyword/proposals/0000-forall-keyword.rst)

### DIFF
--- a/proposals/0000-forall-keyword.rst
+++ b/proposals/0000-forall-keyword.rst
@@ -1,0 +1,113 @@
+Make ``forall`` a keyword
+=========================
+
+.. proposal-number:: Leave blank. This will be filled in when the proposal is
+                     accepted.
+.. trac-ticket:: Leave blank. This will eventually be filled with the Trac
+                 ticket number which will track the progress of the
+                 implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+Having accepted proposals for `dependent visible quantification`_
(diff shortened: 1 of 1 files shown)


Ericson2314:
Yes, and this seems like a good candidate for the next Haskell standard too.

harpocrates:
I'm strongly in support of this.

Perhaps somewhat tangential: would it be useful to have some warning flag for all suspiciously named identifiers (everywhere, not just in types), maybe even turned on by default for `-Wcompat`? That might provide a good path towards eventually (after a bunch of release cycles) turning pseudo-keywords into full-fledged ones: `forall`, `family`, `role`, `via`, `export`, `static`, `pattern` etc.

The lexer's output should ideally not depend on which language extensions are enabled. Even better: turning these "contextual" keywords into real keywords would enable much better parse errors and language extension suggestions.

RyanGlScott:
What happens to the `ExplicitForAll` extension under this proposal? For instance, will `NoExplicitForAll` mean something?

goldfirere:
`ExplicitForAll` is unaffected, actually. This proposal merely makes `forall` a keyword in types always; it does not assign a semantics under `-XNoExplicitForAll`. That is, it's always an error without `-XExplicitForAll`.

RyanGlScott:
> `ExplicitForAll` is unaffected, actually. This proposal merely makes `forall` a keyword in types always; it does not assign a semantics under `-XNoExplicitForAll`. That is, it's always an error without `-XExplicitForAll`.

Ah, OK! It might be helpful to explicitly point this out in the proposal, since the current phrasing might mislead one into thinking that this proposal would always enable `-XExplicitForAll` (since that's currently the language extension that turns `forall` into a keyword).

goldfirere:
Discussion has wound down, with no noted dissent. I submit to the committee, @nomeata.

int-index:
This is a great change, and it also makes [the dot type operator](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0039-dot-type-operator.rst) easier to implement, see [!363](https://gitlab.haskell.org/ghc/ghc/merge_requests/363).

PR #194, opened 2019-01-14
Updated partial type signatures

goldfirere:
[Rendered](https://github.com/goldfirere/ghc-proposals/blob/updated-partial-type-sigs/proposals/0000-updated-partial-type-sigs.rst)

### DIFF
--- a/proposals/0000-updated-partial-type-sigs.rst
+++ b/proposals/0000-updated-partial-type-sigs.rst
@@ -1,0 +1,324 @@
+Update design of partial type signatures
+========================================
+
+.. proposal-number:: Leave blank. This will be filled in when the proposal is
+                     accepted.
+.. trac-ticket:: Leave blank. This will eventually be filled with the Trac
+                 ticket number which will track the progress of the
+                 implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+There are a number of points of confusion and lack of user control around the partial type signatures feature.
+These are explained in more detail in the Motivation_ section. This proposal describes an update to this feature.
+It is *not* backward-compatible, but the expectation is that partial type signatures are a tool to use during
+development, not something that should remain in a codebase for releases (though this proposal makes it easier to
+do so, if you wish).
+
+Motivation
+------------
+There are a number of infelicities around partial type signatures, described here in no particular order (but
+numbered for back-reference).
+
+1. Underscores may mean many different things.
+
+   a. In patterns, a ``_`` means a part of a pattern that matches anything, but is otherwise immaterial. Examples::
+
+        const x _ = x
+        type instance IsJust (Just _) = True   -- happens in types, too!
+
+      I will call these underscores "blanks".
+
+   b. In expressions, a ``_`` means a part of an expression for which you want GHC to print its type. Example::
+
+        tuple :: (Int, Double, Bool, Char, Float, ShowS)
+        tuple = (5, 3.14, False, _, _, _)   -- oops: I lost count, and then I forgot what ShowS meant
+
+      Note that this can be useful in types, too::
+
+        foo :: SomeMajorTypeFamily Nat Bool (_ 5) -> Bool  -- I forget: should I use 'Just or 'Left there??
+
+      There is no way to get this behavior in types today.
+
+      I will call these underscores "holes".
+
+   c. In type signature declarations, a ``_`` means a part of a type for which you want GHC to infer its value using
+      the definition of the variable being declared. Example::
+
+        addOne :: _ -> Int
+        addOne x = x + 1
+
+      Here, we want GHC to use the definition of ``addOne`` to infer that the ``_`` stands for ``Int``. This works
+      also for contraints::
+
+        addOnePoly :: _ => a -> a
+        addOnePoly x = x + 1
+
+      GHC will infer that the ``_`` stands for ``Num a``.
+
+      In all cases, GHC reports to the user what it figured out. GHC normally rejects such programs, but it will
+      accept them with ``-XPartialTypeSignatures`` enabled. GHC still reports its deductions for the wildcards
+      (these underscores are called "wildcards"), unless ``-Wno-partial-type-signatures`` is specified.
+
+      Wildcards naturally are disallowed in places where there is no expression nearby, such as in class method types
+      or instance heads.
+
+      The existence of a wildcard in a type signature drastically changes the algorithm GHC
+      uses during type-checking. As one example of this, if a declaration has a wildcard in its type, it may
+      not use polymorphic recursion, which requires a complete type signature and an utter absence of type inference.
+      Another example is that GHC defers kind-generalization until *after* checking the expression; this can
+      affect whether or not a program is accepted.
+
+   d. In visible type applications, a ``_`` means a part of a type for which you want GHC to infer its value using
+      context. Example::
+
+        x = id @_ True
+
+      GHC will infer that the ``_`` stands for ``Bool``. This form is most useful if you have multiple type
+      parameters and want to skip some. For example, we have ``const :: forall a b. a -> b -> a`` and we
+      might say ``const @_ @Bool 'x'`` to get a constant function of type ``Bool -> Char``. Note that the
+      choice for ``a`` is easily inferrable to be ``Char`` here.
+
+      I will call these underscores "elisions". These occur in visible type applications. They never
+      cause GHC to reject a program or print out further information.
+
+.. _`visible kind application`: https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0015-type-level-type-applications.rst
+      
+2. Visible kind applications don't fit well with partial type signatures. As recently merged into HEAD,
+   the `visible kind application`_ implementation treats underscores as a combination of wildcard and elision:
+   the existence of an underscore in a visible kind application has GHC treat a type signature as partial,
+   even though GHC does not stop compilation or emit warnings for such underscores. Furthermore, the value
+   of an underscore in a visible kind application (e.g., in ``Proxy @_ True``) can often be inferred from
+   context, not from an expression. This design was chosen because it's close to the treatment for visible
+   type applications, but it's an unhappy compromise.
+
+   Visible kind application (even an innocuous usage like the example in the previous paragraph) with underscores is not allowed
+   where wildcards are not allowed, such as in data constructor declarations and in instance heads.
+
+3. Named wildcards act like wildcards but are named. This allows two niceties: the user can specify that
+   the same wildcard is used twice, and output is clarified by giving a name to the wildcard (instead of
+   just ``_``). Here is an example::
+
+     foo :: _w -> _w -> _w
+     foo x 'z' = x
+
+   GHC will infer that ``foo :: Char -> Char -> Char`` (and that ``_w`` stands
+   for ``Char``), knowing that the second argument must be a ``Char`` and that
+   both arguments and the return type must be the same. This feature is
+   enabled with ``-XNamedWildCards``.
+
+   The feature is undiscoverable. If I write the code above in a module without ``-XNamedWildCards``, I get a type
+   error. This is because ``_w`` is a legal type variable name in standard Haskell. No error message in GHC suggests enabling this
+   extension.
+
+4. The current design of partial type signatures treats type generalization and constraint generalization differently.
+   Consider these examples::
+
+     ex1 :: _ -> _
+     ex1 x = x
+
+     ex2 :: _ -> _
+     ex2 x = x + 1
+
+     ex3 :: _ -> _
+     ex3 x = not x
+
+   With ``-XPartialTypeSignatures`` enabled, ``ex1`` and ``ex3`` are accepted, but ``ex2`` is rejected. (Actually, ``ex2``
+   is accepted because of type defaulting. Say ``default ()`` to disable type defaulting, and you will observe that it
+   is rejected. Avoiding this twist in the narrative would complicate the example unnecessarily.) In ``ex1``, GHC discovers
+   that ``_`` stands for an unconstrained type variable ``t``, generalizes, and gets ``ex1 :: t -> t``. In ``ex3``, GHC
+   discovers that ``_`` standards for ``Bool``. In ``ex2``, GHC discovers that ``_`` stands for ``t`` where ``Num t`` must
+   hold; however, it rejects the declaration because no context was specified. In order to accept ``ex2``, we need to
+   write ::
+
+     ex2 :: _ => _ -> _
+
+   With the possibility of a constraint, then ``ex2`` is accepted, with type ``Num t => t -> t``.
+
+   What's awkward here is that the examples are given in order of increasing specificity; each example's type is more
+   specific than the previous. Yet GHC's behavior wibbles and wobbles between them.
+
+Proposed Change Specification
+-----------------------------
+
+1. Outside of patterns, treat ``_`` as an elision everywhere. This means that ``_`` means "I don't care".
+
+   - In types, a ``_`` is treated as a fresh unification variable. This means that ``foo :: _ -> _`` is the same as
+     ``foo :: a -> b``, while ``Proxy @_ True`` is the same as ``Proxy @Bool True``. You're instructing GHC that
+     you want it to fill in the ``_`` with what is necessary for the type to kind-check. Once GHC is finished processing
+     the type, however, any opportunity to solve for ``_`` has been taken; at that point, if it is still unconstrained,
+     GHC generalizes over it, like it would a fresh normal type variable.
+
+     Elisions can appear anywhere a type can be written. They do not cause diagnostics to be printed.
+
+   - In expressions, a ``_`` is a part of the expression the author did not care to write. Currently, this means
+     that ``_`` will be replaced with ``error "elided expression"``. In this case, an error will be printed,
+     stating the inferred type of the ``_`` and suggesting to enable ``-XElidedExpressions`` if the user
+     wants to keep the ``error``\ing behavior. With ``-XElidedExpressions``, GHC will still warn; this can
+     be suppressed with ``-Wno-elided-expressions``.
+
+     In the future, GHC may support the possibility of inferring expressions. An elided expression may then
+     be filled in, not with a call to ``error``, but a correct expression. For example, we might imagine
+     that ::
+
+       id :: a -> a
+       id x = _
+
+     has its elision filled in with the only possible non-bottom value, ``x``. Any such behavior will
+     have to be specified separately and would likely be guarded by an extension flag and possible diagnostic
+     output.
+
+2. Outside of patterns, treat an unbound identifier beginning with an underscore as a named wildcard. A named wildcard
+   induces GHC to print an error with the wildcard's type and a suggested value.
+
+   - In a type signature, a named wildcard behaves much as one does today, though its kind will be printed
+     in the diagnostic along with the other information. Just like today, a suggestion will be included
+     to enable ``-XPartialTypeSignatures``.
+
+   - In a type outside a type signature (e.g., in an instance declaration or a data constructor type),
+     a named wildcard will induce a diagnostic including the wildcard's kind and any information GHC
+     can figure out about its value.
+
+   - In an expression, a named wildcard will induce a diagnostic including the
+     wildcard's type and any information GHC can figure out about its value,
+     including suggested replacements. In this way, an expression named
+     wildcard will behave like holes have.
+
+   Named wildcards will behave as such by default. This means that the ``-XNamedWildCards`` extension
+   will be on by default. Specifying ``-XNoNamedWildCards`` means that identifiers that begin with
+   underscores are treated the same as other identifiers.
+
+3. Enabling ``-XPartialTypeSignatures`` is necessary in order for GHC to accept a program with
+   named wildcards in type signatures. These signatures must also be written using the new
+   separator ``::?`` instead of the typical ``::``. That is, we would now write ::
+
+     quux ::? _w -> Bool
+     quux x = not x
+
+   The new separator would be a loud indication that the signature is *partial*. It induces GHC
+   to use its partial-type-signature algorithm instead of its typical type-checking algorithm.
+
+   Partial type signatures would work with elisions, too, allowing ::
+
+     wurble ::? _ -> _
+     wurble x = not x
+
+   Named wildcards would induce diagnostics; elisions would not.
+
+   Partial type signatures would be generalized *after* checking the function body. This would
+   allow something like the following to be accepted::
+
+     silly ::? Proxy a -> ()
+     silly (_ :: Proxy @Bool _) = ()
+
+   Note that the expression would be more specific than its type signature, if we generalized
+   the signature *before* processing the expression.
+
+   Partial type signatures forbid polymorphic recursion, as they do today.
+
+4. Partial type signatures would generalize fresh variables only when an *extra-variables* wildcard
+   is in the type. That is, the last item in the list of variables after the word ``forall`` can
+   now be an elision ``_`` or a named wildcard ``_w`` (but only in a partial type signature).
+   In either case, this means that GHC can
+   generalize over more variables than have been written in the type signature. As usual, an elision
+   produces no diagnostic, while a named wildcard does. Here are two examples::
+
+     ex4 ::? _ -> _
+     ex4 x = x
+
+     ex5 ::? forall _. _ -> _
+     ex5 x = x
+
+   Here, ``ex4`` is rejected, because we do not know what type ``x`` should have and we cannot
+   generalize. On the other hand ``ex5`` is accepted. The extension ``-XPartialTypeSignatures``
+   would have to be enabled, but there would otherwise be no diagnostic.
+
+   The use of an extra-variables wildcard anywhere other than a top-level ``forall`` in a
+   partial type signature is disallowed, must like the extra-constraints wildcard previously.
+     
+Effect and Interactions
+-----------------------
+* The new design combines the roles of holes and wildcards in the Motivation_. This means that
+  we have only 3 uses of underscores to consider.
+
+* The new design allows the user to control whether they want an elision or a wildcard, using
+  a convenient naming convention.
+
+* The new design gives users fine control over generalization, through the use of ``::?`` to
+  suppress kind generalization and the use of ``forall a b c _.`` to explicitly enable type
+  generalization.
+
+* Visible kind application now fits in nicely. Users can control whether they want elisions
+  or wildcards.
+
+* Partial type signatures have become louder, through the addition of ``::?``. This makes it
+  more sensible to keep partial type signatures in released code.
+  
+Costs and Drawbacks
+-------------------
+* Partial type signatures have proved hard to implement and with many corner cases. The new
+  design seems no simpler nor more complicated than the current, but it will take a fair amount
+  of work to re-engineer.
+
+* The new design does not adequately treat patterns. It is conceivable that a user would want
+  a wildcard (with diagnostic information) in a pattern, and this is no more achievable with this
+  proposal than it was previously.
+
+* This proposal is not backward compatible. However, migration would be straightforward, and I
+  do not expect much released code to be using partial type signatures.
+
+* This proposal rejects the Haskell98 program ::
+
+    id :: _w -> _w
+    id x = x
+
+  as accepting that would require ``-XPartialTypeSignatures``, a change from ``::`` to ``::?``,
+  and the introduction of ``forall _.``. Or, the user could just drop the underscore. Note that
+  this rejection happens even with no extensions enabled, meaning this proposal moves GHC away
+  from the standard (but only for type variables that begin with an underscore).
+
+* This proposal introduces new, wild syntax ``::?``. With two far-flung exception, this new
+  syntax does not replace any existing syntax, as ``::?`` cannot be the name of a function: it
+  starts with a ``:`` and is thus data-constructor-like. Thus, a line like ``x ::? ty`` cannot
+  be mistaken for a top-level Template Haskell declaration splice, as it would have the wrong
+  type.
+
+  Exception 1: It is conceivable to define a pattern synonym named ``::?`` that would have the
+  right type to be a top-level Template Haskell declaration splice.
+
+  Exception 2: It is conceivable to have ``::?`` as a data constructor pattern-matched against
+  as the left-hand argument to another infix operator::
+
+    data PleaseDon't a b = a ::? b
+    a ::? b /\ _ = (a, b)
+
+  In theory, this is disambiguated by the ``=`` (or guard, I suppose), but it would be hard
+  to parse.
+
+Alternatives
+------------
+* Instead of having ``::?``, we could have ``:: {-# PARTIAL #-}`` or similar. A quick grep
+  of all of Hackage (as it was last summer) finds no usage, at all, of the lexeme ``::?``.
+
+* I would welcome new syntax dealing with patterns in this framework.
+
+* Though specification parts (1) and (2) are tightly linked, the others are not, and could be
+  usefully removed from this proposal while not losing other parts.
+
+* Though there is no burning fire here (and thus "do nothing" isn't
+  unreasonable), the design of visible kind application is really quite
+  awkward. If we choose to walk away from this more comprehensive proposal, it
+  would be great to have a concrete design for underscores in visible kind
+  application, at least.
+
+Unresolved Questions
+--------------------
+* Is this really the best syntax? I am uncomfortable at stealing both underscored-idenfitiers and ``::?``.
+  How painful is it to do so?
+


Ericson2314:
Whew! At work every day I'm bitten by typed holes lagging *horribly*, so I do `(undefined :: _)` instead. Now we have a proper language support that.

I sometimes want to do `@_` and get an inferred argument out, so I would like something `@? _` for this purpose matching `::?` Does that make sense?

More broadly still, is the "partial-type-signature algorithm" something that can be switched to in type subexpressions so `:: _ -> (? _w)` would make sense, so `::? ...foo...` could be parsed as `:: (? ...foo...)` and likewise `@? ...foo...` be parsed as `@(? ...foo...)`?

xplat:
Maybe wildcards should look like `_w_` instead of `_w` so it's harder to write them by accident when just trying to use a prism.  Or maybe that's bikeshedding and should go in a separate proposal.

3noch:
This would make partial type signatures far more *useful* to general programming. Right now you can either use them for wildcards *or* diagnostics, but not both in the same file :/

Ericson2314:
Yes, combined with the lagginess we've seen with typed holes, the optic issue that @xplat brings up has meant that renaming constructors is *excruciating*.

nomeata:
On purely procedural points, I am worried about changing the default behavior, and thus deviating from our rule that ghc, by default, implements the report. We have shot down proposals before that were only (maybe) useful if enabled by default, and I think it served us well.

simonpj:
> combined with the lagginess we've seen with typed holes, the optic issue that @xplat brings up has meant that renaming constructors is excruciating.

Could you elaborate on what you mean by

* The "lagginess of typed holes"
* "renaming constructors is excruciating"

and how the new proposal will help?

adamgundry:
Thanks for writing this up, @goldfirere! I found this example a bit confusing:
```hs
silly ::? Proxy a -> ()
silly (_ :: Proxy @Bool _) = ()
```
I guess we figure out that `a` has kind `Bool` from checking the function body, and hence end up with the type `forall (a :: Bool) . Proxy a -> ()`, whereas absent the `::?` we would try to check the body against `forall {k} (a :: k) . Proxy a -> ()` and then fail. I think it would help to spell this out in the proposal, as it shows that marking partial type signatures with `::?` is useful independently of the presence of wildcards in the type.

I haven't fully grokked the proposed new defaults, but I generally agree with @nomeata that we should guard non-standard behaviour with language extensions, especially since there is an awkward interaction between the typed hole/named wildcard syntax and the convention of using underscore-prefixed names for record field selectors or prisms.

proposals/0000-updated-partial-type-sigs.rst
```diff
@@ -0,0 +1,322 @@
...
+
+     wurble ::? _ -> _
+     wurble x = not x
+
+   Named wildcards would induce diagnostics; elisions would not.
+
+   Partial type signatures would be generalized *after* checking the function body. This would
+   allow something like the following to be accepted::
+
+     silly ::? Proxy a -> ()
```
simonpj: I think you mean `silly ::? Proxy _a -> ()` or maybe `silly ::? Proxy _ -> ()`?

simonpj:
Some comments

**Extra-varaibles wildcard**
> Partial type signatures would generalize fresh variables only when an extra-variables wildcard is in the type

I'm content with this change, but you should point out that it's an orthogonal one.  GHC only allows (extra) constraints if there's an extra-constraints wildcard; you are suggesting that it should only allow extra generalisation if there's an extra-variables wildcard. 

But please specify the feature more precisely.

  * Suppose I write `f ::? forall . _ -> _`.  Since there is no extra-varaibles wildcard, presumably that prevents any generalisation taking place.  Please point this out.

  * Suppose I write `f ::? _ -> a`.  Does `a` have to be in scope already, or does this mean `f :: forall a. _ -> a`, i.e. with no extra-variables wildcard?  I assume the latter.  Please say.

**Pattterns**.  The change specification says "outside of patterns...".   What happen in patterns?  Please say.  And include what happens in type signatures in patterns.

**Vocabulary**. You introduce the vocabulary of "holes", "elisions", and "blanks".  That's good, but I suggest making it more prominent and using it consistently.  Is 1(iii) a hole, an elision, or a blank?

**Partial signatures**.   You suggest introducing `::?`, but you don't say why.  Currently the presence of a `_` is enough to cause a partial signature, but you don't want that.  I believe the principal reason is to allow elisions, correct?  So this is an ordinary, complete type signature.
```
f :: Proxy @_ True -> Int
```
Please justify the new syntax.

Another way to articulate it is this: with `::` the signature is complete resolved, including any wildcards, by looking at the signature alone.  With `::?`, we look at the signature and the function definition it describes before resolving what the wildcards mean.

**Complete signatures**.  I'm pretty certain you mean to allow elisions in complete signatures.  But please be precise.  E.g.

* `f :: _ -> _`.  Does this mean the same as `forall a b. a -> b`?
* `f :: forall a. a -> _`.   Presumably this would mean `f :: forall a a -> Any`, since we can't generalise?
* `f :: forall a _. a -> _`.  Now we are back to `f ::forall a b. a -> b`?
* Is an extra_constraints elision allowed in a complete type signature `f :: _ => a -> a`?  I assume not.

xplat:
> I haven't fully grokked the proposed new defaults, but I generally agree with @nomeata that we should guard non-standard behaviour with language extensions, especially since there is an awkward interaction between the typed hole/named wildcard syntax and the convention of using underscore-prefixed names for record field selectors or prisms.

This is a good point!  However, keep in mind that currently typed holes are not only on by default, but cannot be turned off at all.  I'm not sure it's as big a deviation from the report in theory, but the practical consequences of the proposal as written might be less.

Ericson2314:
@simonpj 
>   Could you elaborate on what you mean by
>
>   - The "lagginess of typed holes"
>   -  "renaming constructors is excruciating"
>
>   and how the new proposal will help?

For some reason, even typed wholes in programs at work slows down GHCi. Naming conventions around optics meant that there are lots of bindings of identifiers starting with `_`. When a Constructor (or field) is renamed, all those `_ident`s are no longer bound to auto-generated optics, but now are unbound variables and get treated as typed holes. That large number of typed holes ends up slowing GHCi down horrendously.

simonpj:
> For some reason, even typed wholes in programs at work slows down GHCi.

Is there a ticket?  Can you make one, with a repro case?

> Naming conventions around optics meant that there are lots of bindings of identifiers starting with _. When a Constructor (or field) is renamed, all those _idents are no longer bound to auto-generated optics, but now are unbound variables and get treated as typed holes

I'm afraid I don't understand this. Can you be more explicit, giving a concrete example or two?

treeowl:
You write "It is not backward-compatible, but the expectation is that partial type signatures are a tool to use during development, not something that should remain in a codebase for releases (though this proposal makes it easier to do so, if you wish)." That was not my previous impression. I thought one of the original motivations for partial type signatures was the fact that certain forms of heavily typed programming tend to lead to extremely large type signatures that are likely to change frequently (but in boring and limited ways) when programs are refactored.

3noch:
@treeowl I can confirm that I've both seen and used Partial type signatures in production code. In my experience you can usually find a comment nearby to the effect of `Upstream needs to export more types.` or `TODO: Figure out how to make the type of this not insane so it can be written down`. However, the fact I pointed out earlier (that you can can't practically use them for both wildcards or diagnostics in the same file) cripples their general usefulness where you might actually like to use them as a sort of hybrid inference (which seems compelling since we have inference after all).

Ericson2314:
In short, the one situation that witnesses the breaking change has so much to benefit from this, so the breakage should be worth it.

howtonotwin:
> Partial type signatures would be generalized after checking the function body. This would allow something like the following to be accepted:
> ```haskell
> silly ::? Proxy a -> ()
> silly (_ :: Proxy @Bool _) = ()
> ```
> Note that the expression would be more specific than its type signature, if we generalized the signature before processing the expression.

What is the reasoning for this? Is it even necessary? Can it not always be written in this form?

```haskell
silly ::? Proxy _a -> ()
silly (_ :: Proxy @Bool _) = ()
```

I don't really expect `::?` to disable the normal "generalize lowercase type variables" behavior.

adamgundry:
@howtonotwin the `silly` example is to do with instantiating the implicit unification variable for `a`'s kind, not `a` itself, which remains universally quantified (see https://github.com/ghc-proposals/ghc-proposals/pull/194#issuecomment-454314033).

goldfirere:
Lots to respond to!

* The slowness of typed holes is orthogonal to this proposal. It has to do with the search for valid hole fits. This feature can be affected by flags, as documented in [the manual](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/flags.html); search that page for "hole".

* > I sometimes want to do `@_` and get an inferred argument out, so I would like something `@? _` for this purpose matching `::?` Does that make sense?

    I'm afraid to say: no, it doesn't. Could you elaborate?

* The partial-type-signature algorithm either happens or it doesn't; it affects the way GHC processes a whole type signature, and does not apply only to a part of a type.

* The excruciating nature of renaming a definition that begins with an underscore is just an issue of performance, right? If the errors were produced snappily, would that problem abate?

* It turns out to be easy to rejigger things not to move away from the standard. I have updated the proposal. See the new proposed change specification, with a new numbered point explaining this detail.

* Yes, my `silly` example is quite subtle. I've elaborated at the end of the proposal.

* I have elaborated on the extra-variables wildcard. I point out in the Alternatives section that the last two points in the specification are orthogonal to the others.

* There is no change to behavior in patterns. Type signatures in patterns are not patterns. I've clarified this in the "Effect and Interactions" section.

* Yes, the chief reason for the introduction of `::?` is to allow elisions in types without forcing a type signature to become partial. I've added this to the "Effect and Interactions" section.

* See further examples of elisions at the end of the proposal.

* You may not use an elision as a constraint in a complete type signature.

I'm hoping the other questions raised above are answered in the conversation.

I *have* grown worried that this proposal makes the whole scheme over-elaborate. Is there a simpler design that's perhaps a touch less expressive but is easier to explain?

Thanks for the comments!

Ericson2314:
> The slowness of typed holes is orthogonal to this proposal. It has to do with the search for valid hole fits. This feature can be affected by flags, as documented in the manual; search that page for "hole".

Oh good! I didn't even realize it was trying to come up with things to fit, haha. Hopefully making those 0 will make the problem go away.

> I'm afraid to say: no, it doesn't. Could you elaborate?

I think i understand better now. This proposal is mainly about about better controlling which type inference algorithm is used, which is a whole-signature affair. For `@_` there may be only one algorithm, but there is still a question of whether to print the missing piece. `@?` would pun `::?` for its diagnostic purposes, not algorithm selection purposes; everything happens the same except the inferred type is printed.

simonpj:
> •	The slowness of typed holes is orthogonal to this proposal. It has to do with the search for valid hole fits.

Ericson2314: would you like to open a ticket, with a clear repro case, that illustrates the perf problem you are having?

proposals/0000-updated-partial-type-sigs.rst
```diff
@@ -0,0 +1,418 @@
...
+   - In a type outside a type signature (e.g., in an instance declaration or a data constructor type),
+     a named wildcard will induce a diagnostic including the wildcard's kind and any information GHC
+     can figure out about its value.
+
+   - In an expression, a named wildcard will induce a diagnostic including the
+     wildcard's type and any information GHC can figure out about its value,
+     including suggested replacements. In this way, an expression named
+     wildcard will behave like holes have.
+
+3. The use of a type variable that begins with an underscore will induce a warning
```
simonpj: I think you mean "In the absence of `-XNamedWildCards`, the use of..."

proposals/0000-updated-partial-type-sigs.rst
```diff
@@ -0,0 +1,418 @@
...
+do so, if you wish).
+
+Motivation
+------------
+There are a number of infelicities around partial type signatures, described here in no particular order (but
+numbered for back-reference).
+
+1. Underscores may mean many different things.
+
+   a. In patterns, a ``_`` means a part of a pattern that matches anything, but is otherwise immaterial. Examples::
```
simonpj: Head (a) with "**Blanks**" in bold font.

proposals/0000-updated-partial-type-sigs.rst
```diff
@@ -0,0 +1,418 @@
...
+
+        const x _ = x
+        type instance IsJust (Just _) = True   -- happens in types, too!
+
+      I will call these underscores "blanks".
+
+   b. In expressions, a ``_`` means a part of an expression for which you want GHC to print its type. Example::
```
simonpj: Head (b) with "*Holes*" in bold font

proposals/0000-updated-partial-type-sigs.rst
```diff
@@ -0,0 +1,418 @@
...
+      Wildcards naturally are disallowed in places where there is no expression nearby, such as in class method types
+      or instance heads.
+
+      The existence of a wildcard in a type signature drastically changes the algorithm GHC
+      uses during type-checking. As one example of this, if a declaration has a wildcard in its type, it may
+      not use polymorphic recursion, which requires a complete type signature and an utter absence of type inference.
+      Another example is that GHC defers kind-generalization until *after* checking the expression; this can
+      affect whether or not a program is accepted.
+
+   d. In visible type applications, a ``_`` means a part of a type for which you want GHC to infer its value using
```
simonpj: Head this with "**Elisions**" in bold font

proposals/0000-updated-partial-type-sigs.rst
```diff
@@ -0,0 +1,418 @@
...
+
+      Note that this can be useful in types, too::
+
+        foo :: SomeMajorTypeFamily Nat Bool (_ 5) -> Bool  -- I forget: should I use 'Just or 'Left there??
+
+      There is no way to get this behavior in types today.
+
+      I will call these underscores "holes".
+
+   c. In type signature declarations, a ``_`` means a part of a type for which you want GHC to infer its value using
```
simonpj: I'm not sure how to head this. Is it a blank, an elision, a hole, or what?

simonpj:
I think this is all helpful, but there is still a lot of vocabulary confusion.

In the motivation section you introduce "elision", "blank", "hole" and a fourth anonymous item (c). 

But then you don't use this vocabulary later, except "elilsion".  Instead you use "wildcard" or "named wildcard".  This is jolly confusing. Could you establish and use consistent vocabulary?

A table might help:
* Rows: syntax, diagnostic reports (may differ for terms and types)
* Columns: blank, hole, elision

So in the "elision" column, the syntax is "_"; the diagnostic in types is nothing; in terms it is "print type".

Maybe there is different behaviour in partial signatures?

aspiwack:
I've got a few comments about form. And then some thoughts. But let's start with the superficial comments:

- In a sense, the "blank" pattern is an elision: I'm saying to GHC that I want it to come up with a fresh name for a variable I don't care about. (this is not how GHC implements it, but most presentations of pattern matching are actually like that: variable patterns are primitive, and the underscore is an elision).
- Even though it's not hard to deduce from the current behaviour, it would be helpful if the proposal described what precisely `-XNamedWildCards` does after the proposed changes.
- In the motivating section, you distinguish between wildcard and holes. However, holes do not, I believe, make an appearance in the proposal itself. Or rather, it seems like wildcards and holes become the same thing in the proposal, don't they? In fact, I'm not sure what is the distinction you are making between the two in the motivations. My instinct would be to name them both “holes”, and reserve the name “wildcard” to the general notion of an underscore-thing (which includes elisions and holes, and possibly blank if we want to make that distinction).
- “This example is further explained below.”: the link appears to be broken.

Now for something of more substance.

It took me a bit to cut down to the essence of the `::` vs `::?` which is at the heart of this proposal. But at the end of the day, it's easy to sum up: with `::` elisions are universal variables (`foo :: _ -> _` = `foo :: a -> b`), whereas with `::?` they are existential variables (`wurble ::? _ -> _; wurble x = not x` typechecks).

Now, what makes the proposal somewhat hard to swallow is that elisions-as-universal-variables is, basically, never useful (it can always be replaced by a meaningless one-letter variable). So having a special syntax for the more standard case is pretty disconcerting.

Your (@goldfirere) real motivation for this distinction appears to be in the `silly` example. Where you reveal that, in fact, there is another kind of elision we haven't spoken about! Namely the kinds in a type signature. Most of them are invisible, but they are here, _elided_. They will usually be resolved by kind checking, but if variables remain, they will be generalised (as in `sillier`). And you want to account for the fact that you may actually not want to generalise these variables, and unify them with the term below.

Now, here is an idea I had literally while typing down the previous sentence (so not necessarily super fleshed out). Let's imagine we drop this distinction between `::?` and `::` (it's `::` when the signature is complete, and `::?` when the signature is partial). Invisible kinds are a special case of elided things which are implicitly generalised, and which we don't have (or want) syntax for. In this variant `_` is always an existentially quantified something. And we have visible kind applications, since it's already merged. Then, I think, the `silly` example would be:

```haskell
silly :: Proxy @_ a -> ()
silly (_ :: Proxy @Bool _) = ()
```

(note that there is an assumption, which I think is reasonable, that elided type expressions are resolved both when they are kind checked, and when they are involved in type checking)

RyanGlScott:
One potential issue with the `::?` syntax is that it would only work for constructs that have explicit type or kind signatures. We could envision a future in which type class instances could have named wildcards, such as this example:

```haskell
{-# LANGUAGE NamedWildCards #-}

newtype Blah a = MkBlah a
instance _w => Show (Blah a) where ...
```

As far as I can tell, this proposal would effectively prevent this program from ever seeing the light of day, since named wildcards require `::?` in order to be accepted by GHC. (Without `::?`, all you could ever do is get type diagnostics in error messages when compiling it.)

Note that standalone `deriving` declarations actually do let you write wildcard constraints in today's GHC:

```haskell

newtype Blah a = MkBlah a
deriving instance _ => Show (Blah a)
```

However, due to a lucky (or unlucky?) fluke, you currently can't use a named wildcard in place of the `_`, even with `NamedWildCards` enabled. So while this proposal wouldn't prevent you from writing any programs that today's GHC would allow, it would prevent certain avenues of future design. I don't know if I have a strong opinion on whether that is a bad thing or not, but it might be worth mentioning in the proposal.

goldfirere:
Some responses to recent comments:

* Underscores-as-universals is precisely as useful as underscores-as-patterns. Instead of writing `const x _ = x`, of course I could write `const x y = x`. But I prefer not to. It can be the same in types.

* @aspiwack is right that holes and wildcards are very similar. They are treated differently today, which is why I describe them differently.

* I disagree with @aspiwack's characterization that `::` is to universal as `::?` is to existential. See the new "Resolved Questions" section of the proposal.

* @aspiwack suggests that I can write silly as

    ```
    silly :: Proxy @_ a -> ()
    silly (_ :: Proxy @Bool _) = ()
    ```

    But that's subtly different: it requires that `_` be solved before looking at the term. This is important, because we must know, before looking at the term, whether or not to kind-generalize the type.

* I agree with @RyanGlScott that this new syntax is hard to generalize to new places. We might need new syntax in these places (`instance? _w => Show ...` anyone? or `instance _w =>? Show ...`? The first doesn't steal syntax). But I'm not very bothered by this; we can always invent new syntax.

aspiwack:
> Underscores-as-universals is precisely as useful as underscores-as-patterns. Instead of writing const x _ = x, of course I could write const x y = x. But I prefer not to. It can be the same in types.

Well, the main reason to use `_` in patterns is to signal that the name isn't used. Which is an important distinction, otherwise there wouldn't be a warning for unused bindings. There is no such warnings for used-only-in-one-place type variables. And it's hard to see how it would be valuable. I don't think anyone would be quite sad to have to write `_y` for unused pattern either. So I don't really find this point too convincing.

> holes and wildcards are very similar. They are treated differently today, which is why I describe them differently.

Is it an implementation detail, or an essential distinction? Well… actually, the real question is: does establishing this distinction yield a more useful classification/terminology than merging these two usages together?

My intuition is that they'd benefit from being merge. But I very well be missing something.

> But that's subtly different: it requires that _ be solved before looking at the term. This is important, because we must know, before looking at the term, whether or not to kind-generalize the type.

What I'm suggesting is for it not to be the case: any remaining existential variables after resolving the `_` would remain existential variables (hence no generalisation for them!) and resolved with the partial signature algorithm. I may (I probably am) missing something, but it sounds like a plausible alternative to your proposed design.

3noch:
As far as syntax, it seems a pragma might be best to handle all cases but keep shorthand around for the most common. Perhaps the pragma could also be enabled regardless of which extensions are enabled and the shorthand would be guarded by some extension.

goldfirere:
@aspiwack : Good point about underscores in patterns. I still say that `_` in a type meaning "I don't care about this" is nice. Java programmers have been able to do this for a decade, with `ArrayList<?>` meaning that we really don't care about the element type of that list, while `ArrayList<A>` implies that the type variable `A` is used elsewhere. Wildcards in Java do more than this, of course, but it's a nice bit of information I can get from looking at a type. 

Holes vs wildcards: I don't think the distinction is worthy of retaining, which is why I merged them. Right now, holes are more occupied by reporting their type, while wildcards report what they unify with. I propose that we just report both the type (kind) and value (if known).

As for `silly`: The question really is whether or not to kind-generalize. I think you might be proposing: If there are any unsolved unification variables, don't generalize; instead, wait to learn more by examining the definition. The problem with this is that it essentially means never to kind-generalize, as generalization only affects un-unified unification variables.

@3noch: Can I understand your suggestion to mean: `foo :: {-# PARTIAL #-} ...` means a partial signature, where `foo ::? ...` is a shorthand for that. But by defining the pragma, it also allows `instance {-# PARTIAL #-} _w => ...`. Seems sensible. I'll include that as an alternative.

3noch:
@goldfirere Precisely.

goldfirere:
@aspiwack's suggestion seems to be a return to the status quo. Your `?` is a proper unification variable, while `_` is like a wildcard. The difference is that my proposal allows only *named* wildcards `_w`, allowing `_` to be the user-facing syntax for your `?`. This works, of course (it's almost what we have now), but it leaves the user unable to choose between a partial type signature and a complete one. It also means that, e.g., `Proxy @_ Int` would make a signature partial, which is probably not what the user wants.

nomeata:

simonpj:
I basically like this proposal; but I can see that no one really has time to work on it, alas.  So, fair enough.

goldfirere:
I believe this proposal has incorporated the discussion here and is ready for submission to the committee. However, it's hard to see who would implement it. (I don't plan to, at least until my other upcoming tasks -- like implementing dependent types -- are complete.)

I know that proposals are meant to be independent of implementation plans, but does it serve a purpose to adjudicate proposals that have no foreseeable hope of being implemented in the next, e.g. year or two? Doing so would serve only to widen the gap between the implementation and (idealistic) "specification", meaning that downstream proposals, etc., have to take this one into account, even though it's unimplemented. I'd love feedback on this procedural point: do we want to accept proposals even when an implementor is not expected to appear?

nomeata:
> do we want to accept proposals even when an implementor is not expected to appear?

In general: yes. Accepted but unimplemented proposals are also useful food for “where can I contribute” people.

Whether we should pursue this one is up to you. I also don't think its a bad idea to leave it here until someone comes along who wants to implement it, and let them then push it, maybe even after revising the proposal to really suit their vision.

simonpj:
I agree that accepted-but-unimplemented is a fine status, and gives an opportunity for contributor who we might not otherwise know about to step forward to implement a well-specfiied thing.

If a proposal has been in that state for (say) over a year, it might be good practice just to review once more, to make sure it still fits with the language as it now is.  (Had it been implemented, the other features would have had to fit in with it.)

goldfirere:
Then I submit to the committee, @nomeata.

bravit:
In order to check that I understand this correctly, I'd like to summarize Richard's proposal once again. So, it is proposed:

* the new treatment of underscores in types (with `-XElidedTypes` and without it) — treat as elision ("I don't care")
* the new treatment of underscores in expressions (with `-XElidedExpressions` and without it) — treat as elision ("I don't care")
* the new treatment of names beginning with an underscore in types (in the presence of `-XNamedWildCards` and without it)
* the new design for partial type signatures (via fully separating them from complete ones via `::?` as opposed to `::` and rethinking an approach for generalization)

@goldfirere, am I losing something here?

It seems to me that this proposal calls for a better name, although I'm not able to suggest one incorporating all this stuff.

goldfirere:
Yes, except for the "with `-Xxyz` and without it" notes. An underscore outside of a pattern is an error without an extension. An underscored identifier outside of a pattern and without an extension is treated like any other identifier -- except for `-Wpossible-named-wildcards`, which is added to increase discoverability.

I agree this is a bit of a sprawling proposal, but (to me, at least) all the pieces fit together and I think should be considered as a clump.

Thanks.

goldfirere:
Gentle ping, @bravit, as I'm sweeping through proposals work.

nomeata:
@bravit, ping again?

int-index:
@goldfirere There's an issue I see with this proposal: it gives different treatment to `_` and `_a`. In patterns, if I write `f (MkX _ _ _)`, it is equivalent to `f (MkX _a _b _c)`, where `a b c` are fresh names.

I would expect the same to hold in all other contexts: expressions, complete type signatures, partial type signatures, and visible type applications.

Could we use this intuition as a guiding principle for the design? That is, in *any* context `_` is treated as a named wildcard `_xxx` for some fresh name "xxx".

int-index:
I'm also unsure of the merits of partial type signatures which do not have any wildcards in them.

Consider:

```
f ::? a -> a
f x = not x
```

We accept this, unifying `a ~ Bool`. However, this use case is now covered by [Function Result Type Signatures](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0228-function-result-sigs.rst):

```
f (x :: a) :: a = not x
```

This makes me wonder if the `::?` is necessary at this point.

goldfirere:
I respectfully disagree with both of @int-index's recent posts.

> [We should always treat `_` as `_x` where `_x` is fresh]

But they're different! We have a need for several different ways of leaving bits of code out, as this proposal points out. Whether or not the user offers a name is a valuable clue in what to do. I don't see a reason to ignore this information.

> With this proposal, we'd accept this, unifying `a ~ Bool`.

I don't believe the proposal suggests this. What it means is that any unconstrained kind variables in the type signature will not be generalized, but instead available for unification while type-checking the term, as exemplified in the first "[further example](https://github.com/goldfirere/ghc-proposals/blob/updated-partial-type-sigs/proposals/0000-updated-partial-type-sigs.rst#id8)".

int-index:
> But they're different!

There are things that look different but have the same meaning. E.g. `liftA2 f a b` and `f <$> a <*> b` and `pure f <*> a <*> b`. Using one or the other is a stylistic choice.

Before this proposal, `_` and `_x` were among the things that I could treat the same. Giving something a name was a stylistic choice. The proposal creates a distinction, robbing me of this choice: now it matters how I write a wildcard! Maybe I'm not adding a name because I'm too lazy to come up with a name, not because I want the compiler to read anything into it.

> We have a need for several different ways of leaving bits of code out, as this proposal points out. 

The proposal doesn't demonstrate that we have a _need_ for several ways of leaving bits of code out, only that we have the _ability_ to incorporate all these subtly different ways of leaving bits of code out into the language.

The way I see it, there are just two use cases for holes:

* when I want to leave out a piece of the program because it is redundant and the compiler can fill it in for me (you call them "elisions")
* during development, when I'm not sure what I need to write and I add a hole so that the compiler would help me (you call them "holes" and "wildcards")

The only difference is whether the compiler prints a diagnostic! That is, in both cases the compiler infers what to insert in place of the hole. What's different is whether I'm interested in what it inferred or not: I may have left it out for the sake of brevity (then don't print it out), or I may have left it out because I don't know what it should be (then please print it out).

Do we actually need to make more distinctions? I don't understand why I would want all of these hole variations as a user, and what hard-and-fast rule I could use to pick the variation I want.

With the proposed design, there are two axes of choice:

* `::` vs `:?`
* `_` vs `_x`

And this means 2*2=4 sorts of holes. I still can't quite grasp what exactly each variation means.

Instead, we could have just one sort of holes: the ones that the compiler tries to fill in, and fails at compile time if it couldn't. That's it.

> Whether or not the user offers a name is a valuable clue in what to do. I don't see a reason to ignore this information.

I don't believe it's a clue about anything. As a user, I may want to give a name to something, or I may leave it anonymous, and I don't want this choice to have any extra consequences.

> I don't believe the proposal suggests this.

I was under impression that it was exactly what that example demonstrated. The proposals says we ought to accept

```
silly ::? Proxy a -> ()
silly (_ :: Proxy @Bool _) = ()
```

I would assume this means we would also accept this:

```
silly ::? Proxy (a :: k) -> ()
silly (_ :: Proxy @Bool _) = ()
```

(If not, then I'm confused again, what's the difference?)

So we have a kind variable `k` which will get unified with `Bool`. Ok, but we make no distinction between kind variables and type variables, so I would also expect the following to get accepted:

```
silly ::? Proxy (a :: k) -> ()
silly (_ :: Proxy @Bool @True) = ()
```

And now we have `k ~ Bool` and `a ~ True`. Not that different from this example:

```
f ::? a -> a   -- a ~ Bool
f x = not x
```

------


So what do you think of this alternative design, which I believe is much simpler:

* `_` always means "fill it in for me if you know how, or fail if you don't".
* Printing out what was filled in is a tooling issue. At the moment, GHC doesn't print the types it infers, and it's up to tooling to display the type of a thing under cursor. How are holes/wildcards/etc different?

3noch:
@int-index The problem I see with your alternate proposal is the same problem we have now with PartialTypeSignatures. If I want to use them for "fill in for me" then I get drowned in warnings. If I disable the warnings then I can't use them for diagnostics. I can imagine a setting like `-fwarn-named-wildcards` that might be an interesting middle ground between your proposal and this one.

int-index:
@3noch With my alternate proposal, GHC wouldn't print anything at all if it successfully filled in a wildcard. It's up to your tooling to display what GHC inferred. You place the cursor in your text editor over the hole, and `haskell-ide-engine` tells you what GHC thinks this hole stands for.

Perhaps this doesn't sound good if you don't want to set up tooling. I, for one, don't use it, and this argument is based not on practical considerations but on ideological divide between the language and its tooling. We can provide a pragma as a workaround:

```
f (x :: _) = not x   -- fill in and don't print out

f (x :: {-# PRINT #-} _) = not x  -- fill in and print out
```

3noch:
I think "Just use an IDE" is a bit too optimistic currently. But certainly a pragma or a special option would do the trick IMO.

Ericson2314:
Somewhat related, I added something to https://gitlab.haskell.org/ghc/ghc/wikis/typechecking in hopes that the partial type signature algorithm could be closer to regular checking (vs inferring / synthing).

goldfirere:
I think @int-index may have misunderstood the core of `::?`. That separator merely tells GHC to go into "partial type signature" mode instead of its normal interpretation of types. So `foo ::? a -> a` still checks the body of `foo` against `forall a. a -> a`. The only difference is that we wait until *after* checking the expression before fixing `a`'s kind. With `::`, `a`'s kind gets fixed before checking the expression. The key point here is that type variables in partial type signatures still are proper (skolem) type variables.

Yes, a pragma would work to distinguish cases between whether to print or not. But that's very very loud. I'm looking for a syntactically cheap way of the user indicating their desires: diagnostic or no.

The idea of doing it all in tooling is quite attractive, too. We can imagine that lots of information gets heaped on any underscore in a program in, say, an HIE file. Then some tool can explore the file and get the user the information they want. If we could assume sufficient tooling, I think this proposal would look quite different, and that might suggest doing things differently, in anticipation of tooling. I'd be curious for community feedback on this point, though: how close are we to this future? Perhaps a midway point would be to add commands to GHCi to ask it about underscored things in the source. That is: no diagnostics by default, but straightforward GHCi commands produce the info we have. That design might be somewhat simpler than what I have here.

However: the tooling question does not interact, at all, with the `::` vs `::?` question. The separator question is all about instructing GHC as to what inference algorithm to use.

Ericson2314:
Ah so what i mentioned is related. With https://gitlab.haskell.org/ghc/ghc/wikis/typechecking we could perhaps make the choice per hole and not per signature after all. so `::?` wouldn't be needed but something per hole in its stead.

goldfirere:
But `::` vs `::?` has nothing to do with the difference between `_` and `_x`, so I don't understand @Ericson2314's most recent comment. Regardless, I don't see how open recursion addresses this area -- maybe I've crossed signals, but I'm finding this murky.

bravit:
As a shepherd to this proposal, I recommend **accepting** it. The rationale is as follows.

1. The proposal unifies the treatment of underscores as elisions (meaning "I don't care") everywhere outside of patterns: 

   - under `-XElidedTypes` in both complete and partial type signatures, other types and visible type applications; 
   - under `-XElidedExpressions` in expressions.

2. The proposal unifies the treatment of identifiers beginning with an underscore as named wildcards under `-XNamedWildCards` and increases the discoverability of this extension by suggesting enabling it.

3. The proposal changes the behavior of `-XPartialTypeSignatures` by 

   - requiring partial type signatures to use the `::?` separator instead of `::`;
   - providing finer control over type/kind generalization semantics in partial type signatures.

Despite breaking backward compatibility (e.g., the proposal suggests warning on Haskell98 programs which use identifiers beginning with an underscore in types) and stealing both `::?` and `_`-identifiers, I think that all this stuff is a good thing to have and is much cleaner than what we currently have.

int-index:
@goldfirere Forgive me for being obtuse, but I still don't see why `::` vs `::?` is needed to address the motivation of the proposal. What about this, much simpler, alternative?

* `_x` always means "fill it in for me if you know how, or fail if you don't". If I use `_x` several times, e.g. `id :: _x -> _x`, then I want the compiler to fill in the same thing in these positions.
* `_` always means `_xxx` for some fresh name "xxx".
* When an expression is annotated with `{-# PRINT #-}`, GHC will output the inferred contents of all holes within this expression.

If you'd like to introduce `::` vs `::?`, I would appreciate a motivation other than cleaning up the treatment of holes.

As a user, I definitely do *not* want to add a question mark to make my signature partial. It's very convenient that a signature is partial whenever there are holes in it.

proposals/0000-updated-partial-type-sigs.rst
```diff
@@ -0,0 +1,497 @@
...
+
+     This would mean the same as ``f :: forall a. forall b. a -> b`` (the second ``forall`` prevents
+     ``b`` from being in scope in the defined term). I'm sure you now ask: "Why the unrequested
+     generalization?" Because this is a complete type signature, not a partial one. The new
+     generalization behavior affects only *partial* type signatures. So, ::
+
+       f ::? forall a. a -> _
+
+     would use the definition for ``f`` to determine the result type, but that type could not
+     mention a type variable other than ``a`` (assuming we're at top-level).
```
gridaphobe: Why do partial signatures affect the generalization strategy? That doesn't seem to be explained in the proposal (nor does the need for a new `x ::? t` form for that matter), but it seems quite counterintuitive. Wouldn't it make more sense to say that the generalization strategy depends on the thing being generalized? Then we have three cases:

1. plain type variables: implicitly generalized unless there's an explicit forall.
2. type wildcards: generalized if there's a forall elision (though see my comment above).
3. type elision: implicitly generalized always.

proposals/0000-updated-partial-type-sigs.rst
```diff
@@ -0,0 +1,497 @@
...
+       f :: _ -> _
+
+     This would mean the same as ``f :: a -> b``.
+
+  2. ::
+
+       f :: forall a. a -> _
+
+     This would mean the same as ``f :: forall a. forall b. a -> b`` (the second ``forall`` prevents
+     ``b`` from being in scope in the defined term). I'm sure you now ask: "Why the unrequested
```
gridaphobe: Can a type elision introduce constraints as well? e.g. how would we treat

```
f :: forall a. a -> _ -> _
f _ x = x + 1
```

proposals/0000-updated-partial-type-sigs.rst
```diff
@@ -0,0 +1,497 @@
...
+     Elisions can appear anywhere a type can be written, but they cannot
+     appear as a constraint in a complete type signature. They do not cause
+     diagnostics to be printed. The existence of an elision does *not* cause a
+     signature to be treated as partial.
+
+     Without ``-XElidedTypes``, an underscore in a type is an error; the error message would suggest
+     either using a named wildcard to get a diagnostic or enabling ``-XElidedTypes`` to accept the
+     elision.
+
+   - In expressions, a ``_`` is a part of the expression the author did not care to write. Currently, this means
```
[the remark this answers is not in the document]
bravit: I think that "currently" in section 2 means "under this proposal", the context is specified by the first sentence of the passage.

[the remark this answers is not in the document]
goldfirere: @bravit has it right.

proposals/0000-updated-partial-type-sigs.rst
```diff
@@ -0,0 +1,497 @@
...
+   generalize over more variables than have been written in the type signature. As usual, an elision
+   produces no diagnostic, while a named wildcard does. Here are two examples::
+
+     ex4 ::? _w -> _w
+     ex4 x = x
+
+     ex5 ::? forall _. _w -> _w
+     ex5 x = x
+
+   Here, ``ex4`` is rejected, because we do not know what type ``x`` should have and we cannot
```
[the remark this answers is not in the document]
goldfirere: Because wildcards aren't something one can generalize over. Essentially, I'm saying here that you get to generalize over any out-of-scope variables in your signature -- and that's it. Instead, GHC currently generalizes over those variables and any variables it happens to need in order to fill in wildcards. This behavior is useful, but I don't think it should be the default any more than generalizing over extra constraints should be the default.

proposals/0000-updated-partial-type-sigs.rst
goldfirere: I'm OK with this change. It does seem simpler.

proposals/0000-updated-partial-type-sigs.rst
goldfirere: In this example, we have a complete type signature. That means the signature is fully understood without looking at the term. It thus becomes equivalent to `forall a. forall b c. a -> b -> c`. The definition is then rejected.

If we had a partial type signature (with `::?`), then the definition is still rejected, because there is no extra-constraints wildcard.

goldfirere:
I agree that the `::` vs `::?` distinction is not well motivated. I will fix. It's about the fact that GHC takes wildly different approaches to complete vs partial type signatures. The former are treated as authoritative: they are fully processed before looking at a definition. However, complete signatures might usefully have elisions and wildcards -- for example, in kind arguments. (Thus, we don't want their presence to change the signature to be partial.) Partial type signatures, on the other hand, are processed in concert with the definition. They do not allow polymorphic recursion, and we can use information from the term to inform how to fill in the missing bits.

Note that the use of a wildcard in a non-partial signature has GHC pretend the signature is partial in order to get a good diagnostic, but the compilation is aborted. So common workflows should still work.

I find `{-# PRINT #-}` way too noisy. And perhaps we want printing for some but not for others. The presence of a name is a very nice clue that we should take advantage of -- much like how we are now using whitespace to widen the bandwidth between users and GHC.

I will revise the proposal to better motivate `::` vs `::?`. It may take a day or two to get to, though.

int-index:
> The presence of a name is a very nice clue that we should take advantage of

But giving a name is the only way to ask GHC to unify several wildcards. How would I write something like `f ::? _a -> _a` and leave it in the program, without GHC shouting diagnostics at me every time I compile that module?

> I find `{-# PRINT #-}` way too noisy.

That's not a problem, after all, this pragma isn't going to stay in the code for long – it's only to trigger a diagnostic error message (which is, ideally, triggered by tooling like https://github.com/haskell/haskell-ide-engine). So I'm not worried at all about it being noisy. 

> However, complete signatures might usefully have elisions and wildcards -- for example, in kind arguments. 

If a kind argument can be inferred, the API must be designed to make it inferred (by using `forall k.` rather than `forall k ->`). If the author of the API decided to use `forall k ->` to make the kind argument explicitly written by the user, there's probably a reason behind this, and I don't think the user needs a tool to skip to make these parameters inferred again.

Filling a hole in a complete signature, without looking at a definition, is simply not a realistic use case (IMO) and it adds complexity to the proposal (`::` vs `::?`).

bravit:
As @goldfirere [decided to revise](https://github.com/ghc-proposals/ghc-proposals/pull/194#issuecomment-555253525) the proposal, I've updated label accordingly. Please, feel free to ping me when you are done.

aspiwack:
I'm very sympathetic to the issues at the core of this proposal (see _e.g._ [issue #16220](https://gitlab.haskell.org/ghc/ghc/issues/16220), where I have my own set of concerns).

But I can't get myself to be convinced by the proposal. I think it boils down to the fact that the solution is only useful to a GHC developer. When you know about the details of what total signatures and partial signatures mean in the GHC implementation, then it makes sense: `::` vs `::?` is about triggering one of two algorithm which is classifed by a constructor somewhere (there is a third case for absent type signature altogether). But when you are just a user of the GHC, then this is just an obscure distinction.

My current understanding is that there are two underlying issues in this proposal:
1. Deciding when to trigger the total vs the partial type signature behaviour
2. Deciding when to report the type of a hole or when it's ok to ignore it

For issue 1, the following should at least be considered (if only to be discarded, but it needs to be discussed): do kind-checking, if any free variable is left, then classify the type signature as partial, otherwise consider it total.

For issue 2, I don't have anything to propose. But I do find the `_` vs `_x` distinction to be rather obscure too.

PS: I still think that the classification of holes with 4 categories in the motivation section is more confusing than illuminating. In my views, there are only two categories (elisions and wildcard in the rest of the proposal).

int-index:
> `::` vs `::?` is about triggering one of two algorithm which is classifed by a constructor somewhere (there is a third case for absent type signature altogether). But when you are just a user of the GHC, then this is just an obscure distinction.

Even as a GHC contributor, I still find this distinction quite obscure.

> For issue 1, the following should at least be considered (if only to be discarded, but it needs to be discussed): do kind-checking, if any free variable is left, then classify the type signature as partial, otherwise consider it total.

I'm not sure I understand. Consider this signature, which is supposedly complete:

```
f :: Proxy a -> ()
```

First, we implicitly quantify over `a`. But after kind-checking, we also get a kind variable `k`, so we generalize over it:

```
f :: forall {k} (a :: k). Proxy a -> ()
```

The `::` vs `::?` distinction would control whether we generalize by considering the signature in isolation (the complete signature case) or after considering the definition (the partial signature case). In the latter case, we would accept something like this:

```
f ::? Proxy a -> ()
f = g

g :: Proxy (a :: Bool) -> ()
g = const ()
```

We'd first type-check `f = g`, unifying `k ~ Bool`, and only then do generalization of `f ::? Proxy a -> ()`, which would result in assigning `f` the type `forall (a :: Bool). Proxy a -> ()`.

On the other hand, with `::` instead of `::?`, the code above would be rejected, because we would first generalize, assigning `f` the type `forall {k} (a :: k). Proxy a -> ()`, and then reject `f = g` on the grounds of inability to unify `k ~ Bool`.

Personally, I think having a syntactic trigger for this is important. When I write a signature, I expect it to be complete by default. But then, `::` vs `::?` is an unnecessary syntactic trigger: there's hardly a real-world use case that it would serve. A much better trigger is what we have now: the signature is partial whenever there are holes in it. So the user would write one of these:

```
f :: Proxy a -> ()           -- complete
f :: Proxy (a :: k) -> ()    -- complete
f :: Proxy (a :: _k) -> ()   -- partial
```

isovector:
I aggressively _dislike_ the new behavior of `_` for expressions for several reasons:

1) I can't think of any circumstances under which this is actually desirable behavior. The combination of no diagnostics and tiny syntax and is just *asking* for runtime crashes.
2) We can already accomplish the same thing without explicit compiler support via a hypothetical `unimplemented :: Typeable a => a` that calls `show . typeRep @a`.
3) The explicit lack of diagnostics will break most people's workflows for tight REPL loops dealing with typed holes.

While the rest of the proposal looks good to me, I can't support it so long as this (mis)feature is included.

proposals/0000-updated-partial-type-sigs.rst
gridaphobe: It seems a bit odd that type elision can introduce new quantifiers but not new constraints. What's the underlying principle? It can't be avoiding spooky action at a distance because we already have that with quantifiers. Perhaps we want to avoid introducing new constraints because they have a runtime cost. I don't find this argument all that compelling either since the user opted in by writing the elision.

But regardless, the proposal should specify and motivate this behavior.

simonpj:
> the signature is partial whenever there are holes in it.

Yes, that's the trigger today.  Richard proposes instead having a louder trigger `::?`.   I can see the argument that it's better to be very explicit.   But to me that feels a bit too loud.   Today's behaviour seems quite good to me.

proposals/0000-updated-partial-type-sigs.rst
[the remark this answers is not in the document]
gridaphobe: But why can't one generalize over wildcards? I guess I'm not convinced by the distinction between an elision and a named wildcard in a type signature. Like @int-index, I feel like an elision should just be a named wildcard for some fresh name, and the main reason to use a named wildcard directly is to force a unification before handing things off to ghc.

goldfirere:
@int-index 

> But giving a name is the only way to ask GHC to unify several wildcards

Yes, I agree this is a drawback of my current proposal. I'd be happy to see a way to improve.

I've added `{-# PRINT #-}` as an alternative. But I don't like it, because I think it will discourage people from using these features for a quick query.

> A much better trigger is what we have now: the signature is partial whenever there are holes in it.

I've added a new subsection to the Motivation to clarify why I think today's scheme is inadequate.

@aspiwack 

I think the distinction between `::` and `::?` is meaningful for everyone.

* `::` means that a type signature stands alone. Just by reading the type signature for a variable, you know everything about that variable's type.

* `::?` means that a type signature must be read with a definition. Reading the type signature alone gives you only a skeleton, and you must fill in missing pieces by doing inference on a definition.

It's not just about the algorithm!

> My current understanding is that there are two underlying issues in this proposal:

This was a helpful breakdown. But when I thought about it further, I identified 6 things I was trying to do in this proposal, not just 2. :) I've documented this in the proposal; 5 are different ways that underscores are handled, and the 6th is the complete-vs-partial distinction.

> For issue 1, the following should at least be considered

But that means that, suddenly, `foo :: Proxy a -> ()` is now a *partial* signature. If we go this route, we'll eventually need to introduce syntactic clues for when a signature should really be complete. In other words, we'll need CUSTs (complete user-specified types). CUSTs are like CUSKs, and they lead to madness.

> I still think that the classification of holes with 4 categories

Yes, I had intended to use that vocabulary throughout, but then I didn't. I kept the examples, but lowered the prominence of the terms. Instead, I know have my **Summary** section, which really gets at the nub of my problem.

@isovector 

The `_` behavior in terms is guarded behind `-XElidedExpressions`, and I don't expect many to enable that extension. Furthermore, the proposal describes that warnings are printed for these by default; these warnings can be suppressed. In the summary, there was outdated text which erroneously said no diagnostic would be printed, but I've fixed this.

int-index:
> I've added a new subsection to the Motivation to clarify why I think today's scheme is inadequate.

Thanks, so there are two added examples that demonstrate the inadequacy of today's scheme.

1. ```
   blurb :: Proxy (Const @_ @Bool 5) -> ()
   ```

Here, we just want to do a visibility override for the second parameter, but we have no way to do so other than skipping the first parameter with `@_`. If anything, this simply highlights the issues with positional arguments, and would be resolved with #196 

2. ```
    hiccup :: Proxy a -> ()
    hiccup = f
    ```

Here, we want GHC to treat this signature as partial, but it looks complete. But the motivation section includes a solution: insert an underscore to trigger inference.

```
hiccup :: Proxy (a :: _) -> ()
``` 

-----

> The `_` behavior in terms is guarded behind `-XElidedExpressions`, and I don't expect many to enable that extension.

Why must it be a runtime error, though? The proposal says:

> In the future, GHC may support the possibility of inferring expressions. An elided expression may then be filled in, not with a call to error, but a correct expression. For example, we might imagine that
> ```
> id :: a -> a
> id x = _
> ```
> has its elision filled in with the only possible non-bottom value, `x`.

That's a very attractive idea, but I'd rather get a compile-time failure when GHC couldn't infer the expression, rather than a call to `error`. I could turn it into a call to `error` by using `-fdefer-type-errors`.

goldfirere:
>  insert an underscore to trigger inference.

But that would then need `-XPartialTypeSignatures` and `-Wno-partial-type-signatures`. I don't want any diagnostics. I just want to turn off kind-generalization!

> Why must it be a runtime error, though?

My thought is that `-XElidedExpressions` enables the behavior at all. But then expressions that become calls to `error` would trigger the `elided-expressions` warning. Though not written in the proposal, my hope is that when we get good expression inference, properly inferred expressions would not induce a warning. So you can get what you want with `-XElidedExpressions -Werror=elided-expressions`. Does that give you what you want?

int-index:
> I don't want any diagnostics. I just want to turn off kind-generalization!

Then don't add `{-# PRINT #-}` :-) I don't get the criticism of this pragma being too much to type – people can optimize their workflows in various ways. For example, I can bind inserting `{-# PRINT #-}` to a single key in `vim`. Ideally, this is something that `haskell-ide-engine` would provide. Simply moving the cursor to the hole in the text editor would display the diagnostics to the user. As a workaround for the lack of tooling, a pragma is the most suitable syntactic construct I can think of. Sure, it may require more keystrokes than one might desire, but so does typing something like `{-# SCC ... #-}` during profiling.

> Does that give you what you want?

Yes, but having safe behavior (compile error) by default and a flag to make it unsafe (`-fdefer-type-errors`) seems preferable to unsafe behavior by default (runtime error) and a flag to make it safe (`-Werror=elided-expressions`).

isovector:
I'm strongly in agreement with @int-index here. Hiding a magical `_ -> error` transformation behind an extension isn't safe enough IMO --- especially because it can be enabled globally in a cabal file. I'd prefer we just remove this entirely from the proposal and reconsider it if someone has a concrete usecase for it.

int-index:
I thought a bit more about `::` and `::?` and came up with a somewhat compelling argument against using the presence of holes as a syntactic trigger. Consider this program:

```
{-# LANGUAGE PartialTypeSignatures, DataKinds, PolyKinds, ExplicitForAll #-}

import Data.Proxy

f :: forall (a :: Bool). Proxy a -> ()
f = f

fancy :: Proxy a -> _
fancy = f
```

Here's what GHC has to say about it:

```
A.hs:8:21: warning: [-Wpartial-type-signatures]
    • Found type wildcard ‘_’ standing for ‘()’
    • In the type ‘Proxy a -> _’
      In the type signature: fancy :: Proxy a -> _
  |
8 | fancy :: Proxy a -> _
  |                     ^
Ok, one module loaded.
```

It compiles (loads in GHCi) with a warning. Now, one would expect that if I took GHC's inferred type for the wildcard and placed it there, the code would continue to compile. Alas, it does not:

```
fancy :: Proxy a -> ()
fancy = f
```

```
A.hs:9:9: error:
    • Couldn't match type ‘k’ with ‘Bool’
      ‘k’ is a rigid type variable bound by
        the type signature for:
          fancy :: forall k (a :: k). Proxy a -> ()
        at A.hs:8:1-22
      Expected type: Proxy a -> ()
        Actual type: Proxy a0 -> ()
    • In the expression: f
      In an equation for ‘fancy’: fancy = f
    • Relevant bindings include
        fancy :: Proxy a -> () (bound at A.hs:9:1)
  |
9 | fancy = f
  |         ^
Failed, no modules loaded.
```

That's because now the signature for `fancy` became complete and we generalized too early. Now that makes GHC a bit schizophrenic, because we simply followed its advice and now it rejects our program.

From this, I'd conclude that using the presence of holes in a signature is an inadequate tactic to determine generalization strategy.

That said, do we actually need `::?`? When #228 gets implemented, we will have two ways to annotate a function with its type:

```
-- standalone type signature
fn :: A -> B -> C
fn = ...

-- inline type signature
fn (a :: A) (b :: B) :: C = ...
```

How's that for a syntactic trigger? It is already the case that we take the definition into account to check the inline type signature.

int-index:
Another idea to trigger printing the diagnostics: two underscores instead of one.

```
fn :: Int -> _        -- no diagnostics
fn :: Int -> _result  -- no diagnostics
fn :: _a -> _a        -- no diagnostics

fn :: Int -> __        -- yes diagnostics
fn :: Int -> __result  -- yes diagnostics
fn :: __a -> __a       -- yes diagnostics
```

Adding an extra underscore is faster to type than adding `{-# PRINT #-}`.

akhra:
> Another idea to trigger printing the diagnostics: two underscores instead of one.
I don't like the fact that I like this. :laughing: It feels like an arbitrary hack. But _such_ a convenient one, and I can't see an actual downside...

3noch:
Regardless, I like the idea of a pragma, *even if* some more convenient way exists. A pragma lets tooling produce and consume these diagnostics without having to worry about name mangling, clashing, scoping, etc.

goldfirere:
I have included the idea of using `__` for a diagnostic instead of a name. It's a bit dirty, but feels very convenient. Thumbs-up this comment if you agree, and I'll update the proposal to make the `__` idea the primary proposal for this aspect of it all.

akhra:
Thumb up'd, with acknowledgement of @3noch 's point that we should also have a pragma for various reasons. (I also see an argument that the `__` shortcut should be its own extension, though I'm not personally going to campaign for that.)

akhra:
Oh... actually, one point of contention on `__`. The proposed behavior is for this to _enable_ warnings. That's a change from current behavior (always warn on `_`). I think it should be `__` to _disable_, so existing modules compiled with this extension enabled retain their current warning behavior.

goldfirere:
> The proposed behavior is for this to enable warnings. 

Yes, it's true that the proposal is not backward-compatible in this way. But backward compatibility is a non-goal here (feel free to argue that I'm wrong in this regard), as I always think of `-XPartialTypeSignatures` as a development tool and not something to keep in released code.

akhra:
Actual production code:

```

**Comment by akhra** [comment]
Actual production code:
```
# Bare-bones static type-indexed pseudo-map
data Having env where
  Having :: Having '[]
  With :: e -> Having env -> Having (e ': env)
  deriving Typeable

# Typeclass for accessing values within a Having
class EnvHas e env where
  envGet :: Having env -> e
  envSet :: Having env -> e -> Having env

# CoreEnv is an alias for a few other `EnvHas` constraints
getAppEnv :: (CoreEnv boot, EnvHas LoggingEnv boot)
  => RIO (Having boot) (Having _appEnv)
```
I wrote this as a simple type-level solution for building a runtime environment incrementally, without needing discrete structures or parameters for each stage. `_appEnv` explicitly elides a type-level list of about a dozen capabilities. It's only ever _accessed_ with the `EnvHas` interface; this is the only place that list is covariant and has to be concrete.

This module has `{-# OPTIONS_GHC -fno-warn-partial-type-signatures #-}` because indeed, I don't want to see that warning. But it's a clear example of `-XPartialTypeSignatures` being enabled in production; I'm sure it's not the only one!

Furthermore, I have it enabled globally (also likely common). If this lands, I enable it globally too (I will), and somehow have an elision in place which I _intend_ to be temporary but works as-is and due to fast-and-dirty startup development pressures is currently live, the compiler no longer reminds me with a warning.

This isn't a dealbreaker to me by any stretch, it just feels like it has the potential to be annoying in situations adjacent to mine.

**Comment by goldfirere** [comment]
There is a certain nicety to having `_` warn while `__` doesn't. It's like `foo x = 5` warns while `foo _x = 5` doesn't. Common theme: add an underscore to suppress a warning.

**Comment by goldfirere** [comment]
And thanks for illustrating your point, @akhra. Very helpful, and it changed my view of this.

**Comment by sheaf** [comment]
> goldfirere: I always think of -XPartialTypeSignatures as a development tool and not something to keep in released code.

I think there are other good uses for partial type signatures. I have one use case where users write programs in an indexed monad, say `Program i j a`, starting in state i, ending in state j, and returning a result of type a. The user can specify the start state, and the end state is inferred:

```haskell
program :: Program start _end a
program = do -- do notation for indexed monads
  ...
```

Having the end state available in the type (as opposed to existentially quantifying it away) is then useful as this information can be used at the type-level for validation. It would be extremely cumbersome to specify the end state, as it would involve spelling out all the state that has been introduced in the course of the program.    
In my use case, these programs compile to GPU shader modules, which can be combined into a graphics pipeline for rendering. Having the states available at the type-level allows type-level validation of programs, ensuring compatibility of shaders in a graphics pipeline.

int-index:
My argument for `_a` not causing a warning and `__a` causing a warning is that `_a` is less noisy and would look fine in the end program, whereas `__a` doesn't look as nice but works fine as a development tool.

akhra:
I could make the opposite argument: `__a` sticks out more visually and represents intentional emphasis, making it clear at a glance that this is a deliberate elision and not just a leftover development shortcut.

goldfirere:
OK. After a very long wait, I think this is refreshed.

/remind me to resubmit in two weeks.

goldfirere:
@bravit I think this ready for committee consideration. Thanks!

goldfirere:
@nomeata It looks like this fell through the cracks (both mine and others'). Can you reassign this?

nomeata:
I guess it kept falling through cracks. @goldfirere, is this still ready for submission, or have things changed since last year?

goldfirere:
I continue to think that there is room for improvement around partial type signatures, along the lines of this proposal. I would be happy for the committee to consider this, @nomeata.

proposals/0000-updated-partial-type-sigs.rst
```diff
@@ -0,0 +1,726 @@
...
+
+   With ``-XPartialTypeSignatures`` enabled, ``ex1`` and ``ex3`` are accepted, but ``ex2`` is rejected. (Actually, ``ex2``
+   is accepted because of type defaulting. Say ``default ()`` to disable type defaulting, and you will observe that it
+   is rejected. Avoiding this twist in the narrative would complicate the example unnecessarily.) In ``ex1``, GHC discovers
+   that ``_`` stands for an unconstrained type variable ``t``, generalizes, and gets ``ex1 :: t -> t``. In ``ex3``, GHC
+   discovers that ``_`` standards for ``Bool``. In ``ex2``, GHC discovers that ``_`` stands for ``t`` where ``Num t`` must
+   hold; however, it rejects the declaration because no context was specified. In order to accept ``ex2``, we need to
+   write ::
+
+     ex2 :: _ => _ -> _
```
gridaphobe, line 148: This seems like a plain bug to me. A type wildcard should imply a constraint wildcard for any type variables in the inferred type (but perhaps not if those tyvars appear in specified parts of the type).

proposals/0000-updated-partial-type-sigs.rst
```diff
@@ -0,0 +1,726 @@
...
+   its type signature to be complete. The ``singletons`` package runs up against this: in the code it
+   generates, it needs the expression in order to infer the kind of a type variable. Yet, short of inserting
+   an underscore (and then asking our dear users to enable ``-XPartialTypeSignatures -Wno-partial-type-signatures``),
+   there is no way of getting the behavior we want.
+
+6. Control over partial type signatures vs complete type signatures is based on the presence or absence of
+   a wildcard. This means that GHC sometimes makes the wrong decision: it is conceivable to want a partial
+   type signature without a wildcard, and to write a complete signature with one.
+
+   Here is a desired-complete signature that has a missing piece::
```
gridaphobe, line 202: What does "desired-complete" mean? The user has specified an incomplete type signature here.

proposals/0000-updated-partial-type-sigs.rst
```diff
@@ -0,0 +1,726 @@
...
+  proposal.
+  
+* 1A/2A/3C/4A/5C: **Not allowed**: ``foo :: _ -> Int; foo x = 5`` (we could just use a type variable, but the same
+  argument could apply to ``_`` patterns, and yet we really like those); allowed with this proposal.
+
+* 1A/2A/3C/4A/5D: **Not allowed**: ``data Existential where MkEx :: _ -> Existential`` (ditto comment); allowed
+  with this proposal.
+
+* 1A/2A/3C/4B/5A: **Allowed**: ``type family Equals a b where Equals _x _x = True`` (but
+  removing the underscores emits no warning; perhaps that's just a bug)
```
gridaphobe, line 272: What is the bug?

proposals/0000-updated-partial-type-sigs.rst
```diff
@@ -0,0 +1,726 @@
...
+
+a. We might usefully want to combine several possibilities in close proximity. Changing between modes with module-wide flags is unpleasant.
+
+b. As discussed in motivation point (5), we won't want the choice between complete and partial type signatures to be controlled by
+   the presence of underscores.
+  
+Proposed Change Specification
+-----------------------------
+
+1. Outside of patterns, treat ``_`` and ``__`` as elisions everywhere. This means that ``_`` and ``__`` mean "I don't care".
```
gridaphobe, line 379: What is the double-underscore `__`? This is the first I'm hearing of it.

proposals/0000-updated-partial-type-sigs.rst
```diff
@@ -0,0 +1,726 @@
...
+
+   - In types (with ``-XElidedTypes``), a ``_`` is treated as a fresh unification variable. This means that ``foo :: _ -> _`` is the same as
```
gridaphobe, line 381: Is `-XElidedTypes` a new extension? Please specify

proposals/0000-updated-partial-type-sigs.rst
```diff
@@ -0,0 +1,726 @@
...
+     diagnostics to be printed. The existence of an elision does *not* cause a
+     signature to be treated as partial.
+
+     Without ``-XElidedTypes``, an underscore in a type is an error; the error message would suggest
+     either using a named wildcard to get a diagnostic or enabling ``-XElidedTypes`` to accept the
+     elision.
+
+     Each elision spelled ``__`` will cause GHC to print out a diagnostic warning explaining how
+     the elision was filled. Writing an elision as ``_`` suppresses the diagnostic, as does
+     ``-Wno-elided-types``.
```
gridaphobe, line 398: This feels overly complex. I suggest instead:

1. Type elisions are added to the language.
2. A new warning `-Welided-types` is added that emits a warning with the inferred type. Disabled by default, but part of `-Wall`.
3. The warning can be disabled with `-Wno-elided-types`.

proposals/0000-updated-partial-type-sigs.rst
```diff
@@ -0,0 +1,726 @@
...
+
+   - In an expression, a named wildcard will induce a diagnostic including the
+     wildcard's type and any information GHC can figure out about its value,
+     including suggested replacements. In this way, an expression named
+     wildcard will behave like holes have.
+
+   In all cases, the diagnostic is suppressed by prefixing the identifier with only one underscore;
+   two or more underscores will induce the diagnostic.
+
+3. In the absence of ``-XNamedWildCards``, the use of a type variable that begins with an underscore will induce a warning
```
gridaphobe, line 439: Can we just deprecate the old behavior? I realize that takes time, but I think it would be worthwhile.

proposals/0000-updated-partial-type-sigs.rst
```diff
@@ -0,0 +1,726 @@
...
+    id x = x
+
+  Thus, this standards-conforming program would now cause GHC to bleat (but still accept,
+  with its original meaning).
+  
+* Elisions in types (``const :: a -> _ -> a``) seem less useful than other aspects of this
+  proposal, and yet they occupy prime syntactic real estate. Is there a better design around
+  this issue?
+  
+Alternatives
```
gridaphobe, line 620: After reading through this proposal again (after four years!!), I think there may be a simpler approach that would capture most of the value.

I think there are two UX modalities to consider here:

1. _"I don't **care** what goes here."_ This covers existing pattern wildcards and type expression elisions, both of which use `_` and seem to work well. There should be no warning for this, the programmer is explicitly saying they do not care. I think it also covers some use cases for partial type signatures, though sometimes the programmer does care, they just need GHC's help figuring it out.
2. _"I don't **know** what goes here."_ This covers expression holes and the other cases of partial type signatures. These both also use `_` which is confusing! The programmer is asking GHC a question here, not expressing ambivalence. We should have different syntax for this modality; I propose `?` (or `?x` for named queries) as a strawman. This is a developer tool, so the semantics should probably be that GHC prints out the inferred answer as a warning, but proceeds with compilation (injecting bottom values as necessary).

@goldfirere what do you think of this split?

gridaphobe (changes requested):
There are real issues to solve here, but I'm afraid this approach adds too much complexity. See my comment on the Alternatives section for another possible direction that I think would result in a more intuitive system.

gridaphobe:
@goldfirere just want to make sure you saw my comments.

proposals/0000-updated-partial-type-sigs.rst
goldfirere: Very happy with that. Other than your proposed concrete syntax, I don't really see the difference between what you've written and parts 1 and 2 of my proposed change. Your suggestion does not cover when we should generalize type signatures, which is a subtle but important point in practice.

gridaphobe: @goldfirere I think there are some differences, let me try to reproduce your summary table with my thinking.

  | elision (spelled `_`) | hole (spelled `?`)
-- | -- | --
in a type signature | GHC uses unification to fill in elision. If unification does not find a value for the elision, generalize. | GHC treats the signature as if it were partial and uses the definition of the identifier to solve for the hole. Issues an error with both the value GHC has discovered for the wildcard and its kind.
in a visible type application | same as above, but no generalization. If GHC is unable to figure out what an elision should be, error. | same behavior as an elision, but issues an error with the inferred type
in another type | same as in a type signature | issue an error with the kind of the hole and any information GHC can figure out about the content of the hole.
in an expression | GHC replaces the underscore with `error "elision at..."`. | Same behavior as today's holes: issue an error with the hole's type and suggestions for what it might be filled in with.

Note that elisions do not abort compilation unless GHC truly can't figure out what to do, whereas holes always abort compilation with suggestions for filling the hole. Also note that we don't need a `%Partial` modifier any more, a partial type signature is one that contains a hole `?`.

Finally, I think both elisions and holes could be extended with named variants `_x` and `?x` to allow sharing.

PR #195, opened 2019-01-15
Make Q (TExp a) into a newtype

mpickering:
[View rendered](https://github.com/mpickering/ghc-proposals/blob/code-texp/proposals/0000-code-texp.rst)

### DIFF
--- a/proposals/0000-code-texp.rst
+++ b/proposals/0000-code-texp.rst
@@ -13,12 +13,15 @@ Make Q (TExp a) into a newtype
 .. sectnum::
 .. contents::
 
-I propose to modify the typed Template Haskell API to make the ``Code`` type
+I propose to modify the Typed Template Haskell API to make the ``Code`` type
 more abstract. In particular we introduce a new data type called ``Code`` which
 is abstract such that ``Code m a`` represents an expression of type ``a`` produced
 in a monadic context ``m``. The reader should note that this proposal builds on
 top of the `overloaded quotations proposal <https://github.com/ghc-proposals/ghc-proposals/pull/246>`_ which was
 accepted before this proposal.
+
+The Untyped Template Haskell API is unmodified. This proposal is only about
+Typed Template Haskell.
 
 
 Motivation
@@ -69,7 +72,8 @@ There are three problems with the current API:
    it creates even more syntactic noise than normal metaprogramming.
 
 
-3. ``Quote m => m (TExp a)`` is ugly to read and write. This is a weak reason but one everyone
+3. ``m (TExp a)`` is ugly to read and write, understanding ``Code m a`` is
+   easier. This is a weak reason but one everyone
    can surely agree with.
 
 
@@ -93,13 +97,14 @@ Top-level splicing requires an expression of type ``Code Q T`` and produces a va
   bar :: Int
   bar = $$foo
 
-Nested splicing requires an expression of type ``(Quote m, C m) => Code m T`` and the overall
+Nested splicing requires an expression of type ``Code m T`` and the overall
 type of the quotation is a union of the constraints on all the nested splices::
 
   baz :: Quote m => Code m Int
   baz = [|| 1 + $$(foo) ||]
 
-The return type of lifting a value is changed from ``m (TExp a)`` to ``Code m a``.::
+The return type of the ``liftTyped`` method of the class ``Lift``
+is changed from ``m (TExp a)`` to ``Code m a``.::
 
   class Lift a where
     lift :: Quote m => a -> m Exp
@@ -109,9 +114,9 @@ The functions ``unsafeTExpCoerce`` and ``unTypeQ`` are modified to work directly
 with ``Code``::
 
   unsafeTExpCoerce :: m Exp -> Code m a
-  unTypeQ :: Code m a -> m Exp
+  unType :: Code m a -> m Exp
 
-A new method is added in order to perform monadic actions inside of ``Code``.::
+A new function is added to ``Language.Haskell.TH.Syntax`` in order to perform monadic actions inside of ``Code``.::
 
   liftCode :: m (TExp a) -> Code m a
   liftCode = Code
@@ -141,7 +146,7 @@ type constructors.
 Costs and Drawbacks
 -------------------
 
-The main drawback is that this will break all users of typed Template Haskell who
+The main drawback is that this will break all users of Typed Template Haskell who
 write type signatures.
 However, I feel like I am the only user so the impact will be minimal.
 
--- a/proposals/0000-code-texp.rst
+++ b/proposals/0000-code-texp.rst
@@ -80,6 +80,9 @@ There are three problems with the current API:
 Proposed Change Specification
 -----------------------------
 
+In this section, the proposed complete interface for ``Code`` and ``TExp`` is
+given.
+
 A newtype is defined called ``Code``::
 
   newtype Code m a = Code (m (TExp a))
@@ -119,8 +122,10 @@ with ``Code``::
 There are still the normal functions for interacting with ``TExp a``::
 
   unsafeTExpCoerce :: Quote m => m Exp -> m (TExp a)
+  unsafeTExpCoerce = fmap unsafeExpToTExp
   TExp :: Exp -> TExp a
   unType :: TExp a -> Exp
+  unType (TExp a) = a
 
 A new function is added to ``Language.Haskell.TH.Syntax`` in order to perform monadic actions inside of ``Code``::
 
--- a/proposals/0000-code-texp.rst
+++ b/proposals/0000-code-texp.rst
@@ -116,10 +116,15 @@ with ``Code``::
   unsafeTExpCoerce :: m Exp -> Code m a
   unType :: Code m a -> m Exp
 
-A new function is added to ``Language.Haskell.TH.Syntax`` in order to perform monadic actions inside of ``Code``.::
+A new function is added to ``Language.Haskell.TH.Syntax`` in order to perform monadic actions inside of ``Code``::
 
   liftCode :: m (TExp a) -> Code m a
   liftCode = Code
+
+And also a function which allows access to the wrapped ``TExp`` value::
+
+  examineCode :: Code m a -> m (TExp a)
+  examineCode (Code m) = m
 
 It is also useful to implement a method to modifying the underlying monadic
 representation. For example, in order to handle additional effects before running
@@ -131,6 +136,15 @@ a top-level splice::
   -- Can be used to handle a state effect
   handleState :: Code (StateT Int Q) a -> Code Q a
   handleState = hoistCode (flip runState 0)
+
+Two more useful combinators are ``bindCode`` and ``bindCode_`` which
+are versions of ``>>=`` and ``>>`` and interact nicely with QualifiedDo::
+
+  bindCode :: m a -> (a -> Code m b) -> Code m b
+  bindCode q k = liftCode (q >>= examineCode . k)
+
+  bindCode_ :: m a -> Code m b -> Code m b
+  bindCode_ q c = liftCode (q >> examineCode c)
 
 The ``Code`` data constructor is also exposed to users in case they want to
 explicitly interact with the underlying monadic computation in another manner.
@@ -153,21 +167,6 @@ However, I feel like I am the only user so the impact will be minimal.
 
 Alternatives
 ------------
-
-Instead of ``liftCode`` it might have been more intuitive to add convenient
-functions such as ``addM`` and ``addMThen`` to the API::
-
-  addM :: m () -> Code m a -> Code m a
-  addMThen :: m a -> (a -> Code m b) -> Code m b
-
-These functions can already be implemented with ``liftCode``::
-
-  addM q c = liftCode (q >> fromCode c)
-  addMThen q k = liftCode (q >>= fromCode . k)
-
-
-So can be added as library functions.
-
 
 Unresolved Questions
 --------------------
--- a/proposals/0000-code-texp.rst
+++ b/proposals/0000-code-texp.rst
@@ -110,11 +110,17 @@ is changed from ``m (TExp a)`` to ``Code m a``.::
     lift :: Quote m => a -> m Exp
     liftTyped :: Quote m => a -> Code m a
 
-The functions ``unsafeTExpCoerce`` and ``unTypeQ`` are modified to work directly
+The functions ``unsafeCodeCoerce`` and ``unTypeCode`` are introduced to work directly
 with ``Code``::
 
-  unsafeTExpCoerce :: m Exp -> Code m a
-  unType :: Code m a -> m Exp
+  unsafeCodeCoerce :: m Exp -> Code m a
+  unTypeCode :: Code m a -> m Exp
+
+There are still the normal functions for interacting with ``TExp a``::
+
+  unsafeTExpCoerce :: Quote m => m Exp -> m (TExp a)
+  TExp :: Exp -> TExp a
+  unType :: TExp a -> Exp
 
 A new function is added to ``Language.Haskell.TH.Syntax`` in order to perform monadic actions inside of ``Code``::
 
@@ -126,6 +132,9 @@ And also a function which allows access to the wrapped ``TExp`` value::
   examineCode :: Code m a -> m (TExp a)
   examineCode (Code m) = m
 
+``Code`` is still exported though so users can pattern match on it themselves
+rather than using these convenience functions.
+
 It is also useful to implement a method to modifying the underlying monadic
 representation. For example, in order to handle additional effects before running
 a top-level splice::
@@ -133,7 +142,7 @@ a top-level splice::
   hoistCode :: (forall a . m a -> n a) -> Code m a -> Code n a
   hoistCode f (Code a) = Code (f a)
 
-  -- Can be used to handle a state effect
+  -- As an example, hoistCode can be used to handle a state effect
   handleState :: Code (StateT Int Q) a -> Code Q a
   handleState = hoistCode (flip runState 0)
 
--- a/proposals/0000-code-texp.rst
+++ b/proposals/0000-code-texp.rst
@@ -1,0 +1,167 @@
+Make Q (TExp a) into a newtype
+==============
+
+.. proposal-number:: Leave blank. This will be filled in when the proposal is
+                     accepted.
+.. trac-ticket:: #16177
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+I propose to modify the typed template Haskell API to make the ``Code`` type
+more abstract. In particular we introduce a new data type called ``Code`` which
+is abstract such that ``Code m a`` represents an expression of type ``a`` produced
+in a monadic context ``m``. The reader should note that this proposal builds on
+top of the `overloaded quotations proposal<https://github.com/ghc-proposals/ghc-proposals/pull/246>`_ which was
+accepted before this proposal.
+
+
+Motivation
+------------
+
+There are three problems with the current API:
+
+1. It is hard to properly write instances for ``Quote m => m (TExp a)`` as the type is the composition
+   of two type constructors. Doing so in your program involves making your own newtype and
+   doing a lot of wrapping/unwrapping.
+
+   For example, if I want to create a language which I can either run immediately or
+   generate code from I could write the following with the new API. ::
+
+      class Lang r where
+        _int :: Int -> r Int
+        _if  :: r Bool -> r a -> r a -> r a
+
+      instance Lang Identity where
+        _int = Identity
+        _if (Identity b) (Identity t) (Identity f) = Identity (if b then t else f)
+
+      instance Lang (Code Q) where
+        _int = liftTyped
+        _if cb ct cf = [|| if $$cb then $$ct else $$cf ||]
+
+2. When doing code generation it is common to want to store code fragments in
+   a map. When doing typed code generation, these code fragments contain a
+   type index so it is desirable to store them in one of the parameterised
+   map data types such as ``DMap`` from ``dependent-map`` or ``MapF`` from
+   ``parameterized-utils``.
+
+   ::
+
+      compiler :: Env -> AST a -> Code Q a
+
+      data AST a where ...
+      data Ident a = ...
+
+      type Env = MapF Ident (Code Q)
+
+      newtype Code m a = Code (m (TExp a))
+
+
+   In this example, the ``MapF`` maps an ``Ident String`` directly to a ``Code Q String``.
+   Using one of these map types currently requires creating your own newtype and constantly
+   wrapping every quotation and unwrapping it when using a splice. Achievable, but
+   it creates even more syntactic noise than normal metaprogramming.
+
+
+3. ``Quote m => m (TExp a)`` is ugly to read and write. This is a weak reason but one everyone
+   can surely agree with.
+
+
+Proposed Change Specification
+-----------------------------
+
+A newtype is defined called ``Code``::
+
+  newtype Code m a = Code (m (TExp a))
+
+There are three constructs that the proposal affects.
+
+Quoting an expression ``e :: T`` now produces an expression of typed ``Quote m => Code m T``.::
+
+  -- foo :: Quote m => m (TExp Int)
+  foo :: Quote m => Code m Int
+  foo = [|| 5 ||]
+
+Top-level splicing requires an expression of type ``Code Q T`` and produces a value of type ``T``.::
+
+  bar :: Int
+  bar = $$foo
+
+The return type of lifting a value is changed from ``m (TExp a)`` to ``Code m a``.::
+
+  class Lift a where
+    lift :: Quote m => a -> m Exp
+    liftTyped :: Quote m => a -> Code m a
+
+The functions ``unsafeTExpCoerce`` and ``unTypeQ`` are modified to work directly
+with ``Code``::
+
+  unsafeTExpCoerce :: m Exp -> Code m a
+  unTypeQ :: Code m a -> m Exp
+
+A new method is added in order to perform monadic actions inside of ``Code``.::
+
+  liftCode :: m (TExp a) -> Code m a
+  liftCode = Code
+
+It is also useful to implement a method to modifying the underlying monadic
+representation. For example, in order to handle additional effects before running
+a top-level splice::
+
+  handleM :: (forall a . m a -> n a) -> Code m a -> Code n a
+  handleM f (Code a) = Code (f a)
+
+  -- Can be used to handle a state effect
+  handleState :: Code (StateT Int Q) a -> Code Q a
+  handleState = handleM (flip runState 0)
+
+The ``Code`` data constructor is also exposed to users in case they want to
+explicitly interact with the underlying monadic computation in another manner.
+
+
+Effect and Interactions
+-----------------------
+The proposal solves the main problem because now it is easily possible to write
+instances for the ``Code`` type because it is no longer a composition of two
+type constructors.
+
+
+Costs and Drawbacks
+-------------------
+
+The main drawback is that this will break all users of typed Template Haskell who
+write type signatures.
+However, I feel like I am the only user so the impact will be minimal.
+
+
+Alternatives
+------------
+
+Instead of ``liftCode`` it might have been more intuitive to add convenient
+functions such as ``addM`` and ``addMThen`` to the API::
+
+  addM :: m () -> Code m a -> Code m a
+  addMThen :: m a -> (a -> Code m b) -> Code m b
+
+These functions can already be implemented with ``liftCode``::
+
+  addM q c = liftCode (q >> fromCode c)
+  addMThen q k = liftCode (q >>= fromCode . k)
+
+
+So can be added as library functions.
+
+
+Unresolved Questions
+--------------------
+
+
+Implementation Plan
+-------------------
+Implementation is straightforward.
--- a/proposals/0000-code-texp.rst
+++ b/proposals/0000-code-texp.rst
@@ -13,11 +13,11 @@ Make Q (TExp a) into a newtype
 .. sectnum::
 .. contents::
 
-I propose to modify the typed template Haskell API to make the ``Code`` type
+I propose to modify the typed Template Haskell API to make the ``Code`` type
 more abstract. In particular we introduce a new data type called ``Code`` which
 is abstract such that ``Code m a`` represents an expression of type ``a`` produced
 in a monadic context ``m``. The reader should note that this proposal builds on
-top of the `overloaded quotations proposal<https://github.com/ghc-proposals/ghc-proposals/pull/246>`_ which was
+top of the `overloaded quotations proposal <https://github.com/ghc-proposals/ghc-proposals/pull/246>`_ which was
 accepted before this proposal.
 
 
@@ -41,7 +41,7 @@ There are three problems with the current API:
         _int = Identity
         _if (Identity b) (Identity t) (Identity f) = Identity (if b then t else f)
 
-      instance Lang (Code Q) where
+      instance Quote m => Lang (Code m) where
         _int = liftTyped
         _if cb ct cf = [|| if $$cb then $$ct else $$cf ||]
 
@@ -80,18 +80,24 @@ A newtype is defined called ``Code``::
 
   newtype Code m a = Code (m (TExp a))
 
-There are three constructs that the proposal affects.
+There are three main constructs that the proposal affects.
 
-Quoting an expression ``e :: T`` now produces an expression of typed ``Quote m => Code m T``.::
+Quoting an expression ``e :: T`` now produces an expression of typed ``Quote m => Code m T``::
 
   -- foo :: Quote m => m (TExp Int)
   foo :: Quote m => Code m Int
   foo = [|| 5 ||]
 
-Top-level splicing requires an expression of type ``Code Q T`` and produces a value of type ``T``.::
+Top-level splicing requires an expression of type ``Code Q T`` and produces a value of type ``T``::
 
   bar :: Int
   bar = $$foo
+
+Nested splicing requires an expression of type ``(Quote m, C m) => Code m T`` and the overall
+type of the quotation is a union of the constraints on all the nested splices::
+
+  baz :: Quote m => Code m Int
+  baz = [|| 1 + $$(foo) ||]
 
 The return type of lifting a value is changed from ``m (TExp a)`` to ``Code m a``.::
 
@@ -114,12 +120,12 @@ It is also useful to implement a method to modifying the underlying monadic
 representation. For example, in order to handle additional effects before running
 a top-level splice::
 
-  handleM :: (forall a . m a -> n a) -> Code m a -> Code n a
-  handleM f (Code a) = Code (f a)
+  hoistCode :: (forall a . m a -> n a) -> Code m a -> Code n a
+  hoistCode f (Code a) = Code (f a)
 
   -- Can be used to handle a state effect
   handleState :: Code (StateT Int Q) a -> Code Q a
-  handleState = handleM (flip runState 0)
+  handleState = hoistCode (flip runState 0)
 
 The ``Code`` data constructor is also exposed to users in case they want to
 explicitly interact with the underlying monadic computation in another manner.


simonpj:
Formatting is funny... fix?

Why just `Q (TExp a)`?  Would it not make sense to do the same for `Q Exp`?

>splices can fail to run which goes against the principle of typed staged programming. 

Can you be more explicit? What is "the principle of typed staged programming"?  What splices can fail to run?  Examples would help a lot.

> It is hard to properly write instances for Q (TExp a) 

Can you give examples of what instances you might want to write, and what evolutions you have to go through?

mpickering:
> Why just `Q (TExp a)`? Would it not make sense to do the same for `Q Exp`?

I don't have a real opinion about untyped template haskell but it seems to me that it's primarily about constructing code fragments explicitly and unsafely using operations from the `Q` monad so making it abstract wouldn't be appropriate. 

> > splices can fail to run which goes against the principle of typed staged programming.
> 
> Can you be more explicit? What is "the principle of typed staged programming"? What splices can fail to run? Examples would help a lot.

The principle is that we should never fail when running a splice. Errors should occur whilst constructing the value that we want to splice in so that we ensure that we can only construct well-scoped and well-typed terms.

For a particular situation, imagine if you call `qAddTopDecls` with a declaration which is invalid in some manner then it will cause a failure when running the splice. 

> > It is hard to properly write instances for Q (TExp a)
> 
> Can you give examples of what instances you might want to write, and what evolutions you have to go through?

I added this example of a simple DSL where it is useful to write an instance for `Code`.

```
class Lang r where                                                        
  _int :: Int -> r Int                                                    
   _if  :: r Bool -> r a -> r a -> r a                                     
                                                                            
instance Lang Identity where                                              
   _int = Identity                                                         
   _if (Identity b) (Identity t) (Identity f) = Identity (if b then t else f)
                                                                     
instance Lang Code where                                                  
   _int = liftTyped                                                        
   _if cb ct cf = [|| if $$cb then $$ct else $$cf ||] 
```

nomeata:
I guess a week is reasonable. Has this proposal been advertised anywhere (mailing lists, reddit), i.e. is there a good chance that people who care about TH are aware of it? But if it is small, and obviously desirable, then a week of silence is reasonable.

harpocrates:
Isn't this going to break pretty much every non-trivial piece of TH in existence? More importantly, it looks like there isn't going to be any easy way to fix that code in a backwards compatible fashion. It would be great if the proposal could talk a bit more about what this will break and what the recommendation will be to unbreak it...

In short: the breakage to benefit ratio seems a bit high.

mpickering:
@harpocrates Firstly the potential breakage is only in programs which use typed template haskell of which there are not that many judging by the very few bug reports and number of quite serious bugs. However, users who use the API safely by quoting and splicing would just have to change their type signatures with this modification. Users who use the operations from the `Q` monad might have to make some more serious modifications. 

Reflecting on this compared to other changes like, making `Semigroup` a superclass of `Monoid` which broke every single program containing a monoid definition, the breakage here is minimal.

mpickering:
@Ericson2314 what is your objection to this proposal? Will it break the way you use typed template haskell?

harpocrates:
> @harpocrates Firstly the potential breakage is only in programs which use typed template haskell of which there are not that many judging by the very few bug reports and number of quite serious bugs. However, users who use the API safely by quoting and splicing would just have to change their type signatures with this modification. Users who use the operations from the `Q` monad might have to make some more serious modifications.
>
>
> Reflecting on this compared to other changes like, making `Semigroup` a superclass of `Monoid` which broke every single program containing a monoid definition, the breakage here is minimal.

Yes, `Semigroup` was a painful change, but there was a straightforward way of adapting code in a backwards compatible way (add a conditional package dependency on `semigroups` or guard your `Semigroup` instance with CPP).

Also, you propose changing `Lift` to have a `liftTyped :: a -> Code a` method, but today it only has a `lift :: a -> Q Exp` method. This means that all `Lift` instances will also be broken (fixable in a backwards compatible manner only with CPP). I'm guessing you also intend to define a `lift :: Lift a => a -> Q Exp` in terms of `liftTyped`, right?

mpickering:
What happens with `Lift` depends on your proposal #175. This can be easily adapted. 

This proposal fixes a mistake in the original design, there is no very straightforward backwards compatibility story precisely because `Q (TExp a)` is not abstract currently. 

One option for a backwards compatible change (that I am strongly against) is introducing a new module which exposes the new interface and to teach GHC to be able to quote either `Code` or `Q (TExp a)`. This would introduce some kind of limited overloading in the manner I suggested in https://ghc.haskell.org/trac/ghc/ticket/16178.

gbaz:
> The principle is that we should never fail when running a splice. Errors should occur whilst constructing the value that we want to splice in so that we ensure that we can only construct well-scoped and well-typed terms.

I don't think this is a guiding principle that makes sense to me. The promise is not that we never fail -- it is that we produce well typed code, if we happen to succeed. It is always possible to fail -- e.g., exceptions, nontermination, etc.

Ericson2314:
Yes mainly I was confused on what failures there are, but secondarily what @gershom said. There are issues with `Q`'s expansiveness (e.g. IO), but I don't think they are particular to typed TH.

mpickering:
> > The principle is that we should never fail when running a splice. Errors should occur whilst constructing the value that we want to splice in so that we ensure that we can only construct well-scoped and well-typed terms.
> 
> I don't think this is a guiding principle that makes sense to me. The promise is not that we never fail -- it is that we produce well typed code, if we happen to succeed. It is always possible to fail -- e.g., exceptions, nontermination, etc.

Gershom, you are right to clarify this and I agree that I assume people write programs that terminate and don't throw exceptions but I don't see how it is related to the content of the proposal? Which is 

1. allow people to write instances for `Q (TExp` and 
2. prevent people doing unsafe things like `qAddTopDecls` which can definitely fail the guarantee regardless of how you phrase the guarantee. 

If you are referring to (2), the point is that if a user throws an exception in their program then they should expect it to fail. However, if they use the currently advertised API by calling `qAddTopDecls` then the program can fail at splice time. So it is the distinction that the advertised API shouldn't provide unsafe functions rather than preventing people shooting themselves in their own foot.

> Yes mainly I was confused on what failures there are, but secondarily exactly what @gbaz said. There are issues with `Q`'s expansiveness (e.g. IO), but I don't think they are particular to typed TH.

@Ericson2314 What do you mean by "what failures there are"?  The issue you mention about `Q` is irrelevant here. In fact, if it is harder to use operations from `Q` that will help cross-compilation cases.

Ericson2314:
> @Ericson2314 What do you mean by "what failures there are"? The issue you mention about Q is irrelevant here. In fact, if it is harder to use operations from Q that will help cross-compilation cases.

Yes, the things I brought up indeed are irrelevant here, that was my point! When you write "The principle is that we should never fail when running a splice." I couldn't think of any failures that relate to *typed* template Haskell in particular.

So if like @gbaz I think splicing failure in general has nothing to do with typed TH (the point is to prevent errors *after* splicing (type errors) not *during*), and there are no specific splicing *failures* which are relevant to typed TH, then splicing failure disciplines and the typedness of template Haskell variants are truly orthogonal.

mpickering:
So to summarise the discussion here. There is one objection of substance which is that of backwards compatibility. However, I would consider it an absolute travesty if the proposal was rejected on these grounds due to the far more invasive breakages which happen every release. Given that no users have come forward to express this concern, I believe that is should also be disregarded. 

I also grepped my old copy of Hackage (from 2017) for occurences of `[||` and found the following usages:

* http://hackage.haskell.org/package/llvm-general-quote (Author Geoffrey Mainland)
* http://hackage.haskell.org/package/pinchot
* http://hackage.haskell.org/package/network-uri-static
* http://hackage.haskell.org/package/canteven-http
* http://hackage.haskell.org/package/clash-prelude
* http://hackage.haskell.org/package/th-inline-io-action
* http://hackage.haskell.org/package/validated-literals
* http://hackage.haskell.org/package/hedgehog

int-index:
@mpickering I wonder why it must be a newtype over `Q (TExp a)` rather than `WQ (TExp a)` where `WQ` is some sort of "weakened Q" that does not offer a possibility of failure or reification.

Indeed, the name `Code` suggests that this is just an AST for some expression, and I wouldn't expect splicing it to either fail _or_ inspect its environment (by reification). It appears to me that the only thing that `WQ` needs to support is generating fresh variables, so it could be defined along the lines of

```
newtype WQ a = WQ { runWQ :: forall m. (String -> m Name) -> m a }
``` 

It is obvious how to interpret it in `Q`:

```
qRunWQ :: WQ a -> Q a
qRunWQ = runWQ mkName
```

And this opens up a possibility of writing other interpreters for `Code`.

mpickering:
@int-index My primary problem is that you can't write instances for `Q (TExp a)` easily. 

Generating fresh variables is also one of the most dangerous operations! It's very easy to end up with an AST which contains unbound variables. 

(FWIW, I have also implemented an alternative serialisation mechanism for quotes which serialises core expressions rather than to the template haskell AST, this is sufficient if you don't need to inspect it. This isn't relevant to this proposal though.)

int-index:
> Generating fresh variables is also one of the most dangerous operations! It's very easy to end up with an AST which contains unbound variables.

At least that's one dangerous operation instead of the ability to embed arbitrary `IO` via `Q`!

Generating fresh variables can be easily done in pure code with the `State Int`, and there's something satisfying about being able to splice `Code` in a pure context.

It would help me greatly to reason about what `Code` can do and what it cannot do.

In case the only motivation is merely to write instances, I suggest the name `TExpQ` which is explicit about having the full power of `Q`.

yallop:
On the original proposal: it would help in the cases where you need need an instance for `Code`, but not the cases where you need an instance for `Code T` (for some `T`).  In my (limited, possibly atypical) experience the latter are more common.

On the "weakened Q" proposal:
> It appears to me that the only thing that WQ needs to support is generating fresh variables

It's often useful to mix code generation with other effects --- in particular, effects that can make non-local transformations to code.  For example, this is from the abstract of Kameyama, Kiselyov and Shan's (2014) *Combinators for impure yet hygienic code generation*:

> Code generation is the leading approach to making high-performance software reusable. Effects are indispensable in code generators, whether to report failures or to insert let-statements and if- guards.

So I'm concerned that the weakened WQ is likely to be a bit too weak in practice.

mpickering:
You're right that it wouldn't "help" instances for `Code T` but it wouldn't hinder them either. Given the popularity of classes with higher-kinded parameters like `Functor`, `Applicative`, `Monad` and so on, the usefulness of an instance for `Code` can be easily imagined.

I think the best path to "weakened Q" is to disentangle `Q` from the quoting mechanism by making brackets and splices overloaded. This way you can then easily embed them into your own library which provides effectful combinators for manipulating the code fragments. https://ghc.haskell.org/trac/ghc/ticket/16178

int-index:
@mpickering If we make quoting and splicing overloaded, as you propose in Trac #16178, would it not subsume this proposal?

```
class IsBracket p where
  fromBracket :: Q (TExp t) -> p t

class IsSplice p where
  toBracket :: p t -> Q (TExp t)
```

```
newtype Code a = Code { unCode :: Q (TExp a) }

instance IsBracket Code where
  fromBracket = Code

instance IsSplice Code where
  toBracket = unCode
```

int-index:
I'm not sure how it would help the `WQ` case, there is seemingly no way to define `fromBracket :: Q (TExp t) -> WQ (TExp t)`.

mpickering:
@int-index I think the point would be to make `fromBracket :: WQTExp t -> p t` and then embed `WQ` actions inside `Q`. The quote generation does use some operations from `Q` but it might just be ones which make names. I would need to look at the code.

There would still be the same problems with backwards compatibility with the overloaded proposal as you can't easily write an instance for `IsBracket` for `Q . TExp`. So making `Q (TExp` a newtype and then later adding the overloading would be a possible implementation path.

int-index:
OK. I still have a few gripes about the name `Code`. It looks nice, but for me it has connotations that do not hold true for `Q (TExp t)`.

1. Types/declarations are code, but they cannot be represented by `Q (TExp t)`.
2. Doing any sort of effects (even name generation, come to think of it) is not part of the code. `TExp` is code, `Q (TExp t)` is code generation. And if `WQ` becomes a reality, then `Q (TExp t)` is just one possible way of code generation.

gbaz:
Improving support for instances is a good motivation for _something_. But this "semi-purity" stuff feels weird to me. Another reason is that I think it makes sense to have IO effects inside `Q TExp`. For example, you might want to do IO and based on that generate different code, but all the code being of the same type.

So having a weakened Q seems entirely orthogonal to the question of having a newtype for the purpose of writing instances, and I think the `fromBracket` approach is probably a much better way to tackle the latter.

mpickering:
Gershom,I think the intention is that "weakened Q"  supports only the operations that are necessary to make the representations of code. This doesn't include doing IO. Then `WQ` embeds into `Q` which supports things like `liftIO`. GHC still interprets `Q` but other consumers can more easily use quotation for runtime code generation by providing their own interpretation of `WQ` which doesn't include difficult to implement operations like reification. 

I agree you're right that this is orthogonal to the proposal.

mpickering:
Is there anything else blocking this proposal? All objections have been dealt with.

simonpj:
Earlier I wrote

> Would it not make sense to do the same for Q Exp?

To answer my own question: we also have `Q Type`, `Q Decl`, `Q Pattern`.  But we do not have `Q (TType a)` (whatever that might mean), etc.

That is, in the untyped world we construct fragments of many different syntactic categories.  But in the typed TH world we can only construct expressions.  To me that helps explain why we don't want lots of newtypes for `Q Exp` but we do for `Q (TExp a)`.

Ericson2314:
@mpickering While you pointed out correctly that `IsBracket` should use `WQ`, and we embed `WQ` into `Q` with an `IsBracket` instance., hat doesn't address @gbaz's other point that  `IsBracket`/`fromBracket` solves the same problem as `Code`. I also agree that `fromBracket` + `WQ` is the right approach. Seeing that you yourself proposed `IsBracket`, would you still want `Code` after?

mpickering:
That is a good point @Ericson2314 and it also affects how #204 is resolved. 

@bravit Do you think I should revise this proposal and incorporate it into a new proposal for #204 which solves the problems that @yav is worried about?

bravit:
> @bravit Do you think I should revise this proposal and incorporate it into a new proposal for #204 which solves the problems that @yav is worried about?

Yes, that would be the right thing to do.

nomeata:
Is this proposal going to be pursued any further?

mpickering:
Yes.

JacquesCarette:
I want to strongly re-emphasize @gbaz point that one of the first motivations for this proposal, namely "The principle is that we should never fail when running a splice" is not a sound principle. As @yallop rightly pointed out, effects during code generation are extremely useful - and I do use them in my metaocaml code. If this were implemented in Haskell, my partial port of my work to Haskell would become impossible.  To me, for this proposal to go forward, it ought to be possible to reframe it without this particular motivation in mind.

[Note that I am not against making `Code` abstract. This would, in fact, bring it closer to metaocaml!]

Motivation 2 is great; motivation 3 is weak (I agree with it, but would not put so much weight on it).

This proposal would break some of my code, but it would be easy to fix, so I don't object to that part.

Note that, if changes are going to be done to `Lift`, you may as well roll in the bigger change: you can only `lift` first-order values. You might want to add a second method to the class, say `persist` which is for ensuring cross-stage persistence of functions referable to by **name**.

yav:
I am in support of this proposal.   Here is a concrete example of why this change would be useful:

I wrote a compiler, which compiles a typed AST  (i.e. a GADT annotated with types as usual) to Haskell, using typed TH. The type of the compiler is something like this:

```
compiler :: Env -> AST a -> Code a

data AST a where ...
data Ident a = ...

type Env = MapF Ident Code

newtype Code a = Code (Q (TExp a))
```
Here `MapF` (from package `parameterized-utils`) is like `Map` but the keys and the values are parameterized on a common index. This allows the same map to contain entries mapping `Ident Int -> Code Int` and also `Ident (Int -> Int)` to `Code (Int -> Int)`

Note that the `newtype Code` is needed so that I can use `MapF`, which is needed so that the compiler can generate type correct code.

This all works fine, but since typed TH does not know about `Code` all my TH brackets look
likes this:   ``Code [|| .... ||]``   and every time I splice something it looks something like this:
```$$(unCode (...))```.     This is not the end of the world, but it is certainly clunky, and adds quite a bit of noise that serves no useful purpose.

**Comment by mpickering** [comment]
@yav has completely nailed the motivation for this proposal. I tried proposing to overload typed quotations in #246 but failed in that regard so I think we should just make it into a newtype rather than the overloaded version. 

So the plan is to update the proposal to add this additional motivation and then resubmit to the committee.

**Comment by mpickering** [comment]
I updated the motivation section now and think that the committee should consider the proposal.

**Comment by nomeata** [comment]
@bravit Can you shepherd the next round too (if not I’ll ask Iavor).

/remind me to assign a shepherd in three days

**Comment by JacquesCarette** [comment]
Nice rework - I like it.

**Comment by Ericson2314** [comment]
As @int-index points out in the other thread, https://github.com/ghc-proposals/ghc-proposals/pull/246 clashes with this, we would need to add a monad parameter to `Code`, or not modify typed template Haskell in #246.

**Comment by mpickering** [comment]
> I don't have a strong opinion on the proposal. So here is a typo.
> 
> Also, if you want to remove the `.:` at the end of some sentences, you can move the `::` alone on its own line (you also need to indent code more than the `::`).

Thanks @aspiwack

**Comment by simonpj** [comment]
Some questions, which should ultimately be answered by the proposal

* Is the data constructor `Code` visible to clients?  Can they look inside the abstraction?

* `unsafeQ` lets you turn a `Q (TExp a)` into a `Code a`. But how can you get a `Q (TExp a)` in the first place?

* The text says `unsafeQ` is added in order to perform unsafe Q actions inside of `Code`.  So I was expecting something like
  ```
  unsafeAddQ :: Q () -> Code a -> Code a
  unsafeAddQThen :: Q a -> (a -> Code b) -> Code b
  ```
  They look much more plausible to me than the current `unsafeQ`.

**Comment by yav** [comment]
@simonpj I was thinking that there would be a function `fromCode :: Code a -> Q (TExp a)`, which would be useful anyway, if you wanted to splice some typed code in an untyped setting. 

I don't have a strong preference on the API though.  @mpickering what do you think?

**Comment by isovector** [comment]
@mpickering is `Code` a monad like `Q` is?

**Comment by yav** [comment]
@isovector `Code` is not a monad because you cannot define `return` (it only works for types in the `Lift` class).

**Comment by mpickering** [comment]
* The data constructor `Code` is visible to clients but not advised to be used.
* You can get a `Q (TExp a)` just like now, by using `unsafeTExpCoerce` or by using the `TExp` constructor. 
* You can pattern match on `Code` to get the `Q (TExp a)` value out or provide an accessor function as Iavor suggests. The choice isn't very important and will be discovered during implementation. 
* As Iavor says on the mailing list, you can define the proposed API using `unsafeQ`.


> These can be defined using `fromCode` and `unsafeQ`.  If you think
> they might be useful, we could probably add them to the TH modules.
> ```
> unsafeAddQ q c = unsafeQ (q >> fromCode c)
> unsafeAddQThen q k = unsafeQ (q >>= fromCode . k)
> ```

**Comment by simonpj** [comment]
Matthew, could you take a few minutes revise the proposal so that it reflects the discussion, including your recent responses and maybe more below?   We are going to accept it, but when we do I'd love it to stand by itself -- indeed that's what our process says.  Thanks!

----------------

Some responses

> The data constructor Code is visible to clients but not advised to be used.

That's a strange position.  If its use is inadvisable, why make it available?  To what purpose might you put it?   It looks as if `unsafeQ` is essentially `Code` as a constructor; so perhaps you just want an (entirely safe) function `codeTExp :: Code a -> Q (TExp a)` as well?  Ah -- I see that in your later remarks you mention `fromCode` which isn't in the proposal.  Perhaps that's the function you want, and then `Code` can be abstract?

> You can get a Q (TExp a) just like now, by using unsafeTExpCoerce or by using the TExp constructor.

But those functions both start from `Exp` or `Q Exp`.  So maybe you want
```
unsafeEq :: Q Exp -> Code a
```
Now perhaps our TH users never need to know about `TExp` at all?   Could we simplify the API by sticking with Exp and Code?

> As Iavor says on the mailing list, you can define the proposed API using unsafeQ.

But there's a big difference!  There are two reasons that `unsafeQ` might be unsafe:

1.  It takes an `Exp`, constructed who-knows-how, and asserts that it is well typed, with a particular type.
2.  It allows effects like printing or reifying.

Now (1) is a Big Deal; very easy to screw up.  But (2) is comparatively minor.   (I'm not clear about quite how bad this is -- maybe you can remind me?)

The two combinators I suggested allow (2) without allowing (1), and that is (to me) a very useful distinction.  Asking clients to define them by using `unsafeQ` (which allows both (1) and (2)) is ... at least a lost opportunity.

**Comment by simonpj** [comment]
@mpickering I wonder if you plan to revise the proposal in the light of my comments above?  I think that would allow the committee to accept.

**Comment by yav** [comment]
@mpickering on request from the committee I've relabeled this as "needs revision" until you have a chance to modify the proposal on the points raised by @simonpj.

Otherwise the committee is positive on the change, and would likely accept the proposal.

**Comment by mpickering** [comment]
I now realise that I was very wrong about the "unsafe" nature of `Q` and `m`. So I have renamed the function which lifts a monadic value from `unsafeQ` to `liftCode`. I also deleted the section in the proposal about unsafety as it isn't right. Using effects during code generation is perfectly fine and the overloaded quotation patch makes it natural to do so. The unsafe part, as Simon points out, is coercing an `Exp` into a `TExp a`, which is performed either by using `unsafeTExpCoerce` or `TExp` directly. This also resolves the strangeness of exposing `Code` but considering it unsafe to use. 

> But there's a big difference! There are two reasons that `unsafeQ` might be unsafe:
> 
>     1. It takes an `Exp`, constructed who-knows-how, and asserts that it is well typed, with a particular type.
> 
>     2. It allows effects like printing or reifying.
> 
> 
> Now (1) is a Big Deal; very easy to screw up. But (2) is comparatively minor. (I'm not clear about quite how bad this is -- maybe you can remind me?)

`liftCode` only implements (2) as you must already have a `TExp a` in order to use it. The `TExp` constructor is still exposed to users so users can still use it to unsafely construct expressions directly. The existence of `TExp` is also more important in the world containing overloaded quotations as it is possible to implement the monad purely (without using IO) so you can extract a pure `TExp a` value from a quotation. 

-----------------

I have now updated the proposal to incorporate everyone's comments. Some of the points about unsafety are no longer as relevant as I remove the section which described using monadic computation as "unsafe".

File: proposals/0000-code-texp.rst [hunk 1]
```diff
@@ -0,0 +1,167 @@
...
+
+      class Lang r where
+        _int :: Int -> r Int
+        _if  :: r Bool -> r a -> r a -> r a
+
+      instance Lang Identity where
+        _int = Identity
+        _if (Identity b) (Identity t) (Identity f) = Identity (if b then t else f)
+
+      instance Lang (Code Q) where
```
int-index: `instance Quote m => Lang (Code m)`, perhaps? `Q` is too restrictive.

proposals/0000-code-texp.rst
```diff
@@ -0,0 +1,167 @@
...
+      newtype Code m a = Code (m (TExp a))
+
+
+   In this example, the ``MapF`` maps an ``Ident String`` directly to a ``Code Q String``.
+   Using one of these map types currently requires creating your own newtype and constantly
+   wrapping every quotation and unwrapping it when using a splice. Achievable, but
+   it creates even more syntactic noise than normal metaprogramming.
+
+
+3. ``Quote m => m (TExp a)`` is ugly to read and write. This is a weak reason but one everyone
```
int-index: Does `Quote m => m (TExp a)` have worse readability than `Quote m => Code m a`? Maybe a little bit, but not as much as to call it "ugly".

proposals/0000-code-texp.rst
```diff
@@ -0,0 +1,167 @@
...
+A new method is added in order to perform monadic actions inside of ``Code``.::
+
+  liftCode :: m (TExp a) -> Code m a
+  liftCode = Code
+
+It is also useful to implement a method to modifying the underlying monadic
+representation. For example, in order to handle additional effects before running
+a top-level splice::
+
+  handleM :: (forall a . m a -> n a) -> Code m a -> Code n a
```
int-index: This sort of operation is typically called "hoist", not "handle", so what about `hoistCode`?

int-index: https://hackage.haskell.org/package/mmorph-1.1.3/docs/Control-Monad-Morph.html#v:hoist

proposals/0000-code-texp.rst
```diff
@@ -0,0 +1,167 @@
...
+    lift :: Quote m => a -> m Exp
+    liftTyped :: Quote m => a -> Code m a
+
+The functions ``unsafeTExpCoerce`` and ``unTypeQ`` are modified to work directly
+with ``Code``::
+
+  unsafeTExpCoerce :: m Exp -> Code m a
+  unTypeQ :: Code m a -> m Exp
+
+A new method is added in order to perform monadic actions inside of ``Code``.::
```
int-index: Does it serve any purpose when the `Code` constructor is exported?

mpickering: I suppose it is not necessary but there is no harm.

proposals/0000-code-texp.rst
mpickering: Thanks.

proposals/0000-code-texp.rst
mpickering: Thanks.

Ericson2314:
I'm happy that we've resolved that `Code` is safe, I'm happy #246 is taken into account, and I hate being a broken record nag, but that this point `Code` solely exists as a means to type less. Nothing tricky is being encapsulated, and `Compose` would also work just fine for instances.

Is that justification enough? Could `Code` just be a type alias? I rather not paper over the minor difficulties that arrise in general when one has types in the form `m (t a)` in just the specific case of typed Template Haskell.

cgibbard:
Personally, I'd like to see TH move in the direction of having fewer operations that spuriously take Q-actions when all they're going to do is immediately and unconditionally execute them and then use the result. It doesn't quite seem like the right direction to me to bless that existing awkwardness into a new datatype.

mpickering:
@Ericson2314 Using `Compose` is a non-solution as you have to wrap and unwrap it everytime you use a quote or splice because the language bakes in the requirement for what type the argument to a splice has to be.

@cgibbard I'm not sure which part of the proposal you are referring to as there is nearly no mention of `Q` at all in the proposal. The only place `Q` is necessary is at the top-level splice. This proposal greatly improves the ergonomics for anyone actually trying to use typed template haskell of which there are few.

cgibbard:
I was trying to refer to the thing that the name of the proposal refers to -- the newtype of Q (TExp a).

It's possible that I just don't understand the context of typed template haskell well enough, but for instance, Language.Haskell.TH.Lib right now is filled with a whole bunch of trivial liftings of every data constructor in the entire syntax tree, with many functions taking arguments of type Q t, even though there's no real reason why they should need a Q-action and not just its result. Even though I pretty much get what the aim is there, I still think it's kind of a crazy thing to do vs. just using Applicative or do-syntax. Thankfully, those functions are mostly avoidable (except when I end up editing someone else's TH code).

Maybe I don't fully understand what this proposal is getting at yet... I apologise if this is totally out of left field.

Ericson2314:
I do think @cgibbard bring up a good point. `Language.Haskell.TH.Lib` seems designed as if `<$>` and `<*>` and `do` notation are embarrassments to hide from the user, and not that way Haskell is written everywhere else. No joke, I'd guess somebody long ago was worried about cluttering up a conference talk to a non-Haskeller audience or something similar, and wrote these "helpers" for better slides. I don't think any of us would accept them today if a PR was submitted to `template-haskell`, and as such they should be deprecated and eventually removed.

Back to my point, I see wraping and unwraping as, sure, a bit annoying, but I don't think worth a one off solution, just as `<$>` `<*>` as slightly annoying, but acceptable. I think Template Haskell would scare fewer people if it looked more like all the other Haskell, even if that means a bit noisy with infix operators and newtype constructors. If we really want to solve these problems, I would rather create some sort of auto-coerce tactic for this, and idiom brackets for applicative, to tackle these problems in their full generality, and not just in the case of Template Haskell.

Ericson2314:
Heh that got me curious. I could trace what @cgibbard is talking about as far back as the initial TH library set commit https://github.com/haskell-pkg-janitors/haskell-src/commit/733840fc, but then the trail goes cold. I have no idea what was idiomatic then, but this phrase from the commit message

> ...but the emphasis is more on simplicity here.

makes me think perhaps it wasn't then either.

mpickering:
@cgibbard If you look at proposal #246 then you will see that all the combinators no longer mention Q at all so I still don't understand which part of this proposal you want changing. 

@Ericson2314 All the combinators defined in `Language.Haskell.TH.Lib` are very useful when desugaring brackets as you have to manually construct the core terms yourself. Using typed template haskell is already very noisy and without this proposal you have to wrap and unwrap at every splice and quote. When you design a normal library you can write all your combinators in terms of `Compose` so all your library functions work smoothly together but for template haskell you have no choice as the compiler enforces the type of `quote/splice`. 

At the end of the day, this proposal is about making life easier for people who use typed template haskell in the immediate future. Anyone is welcome to propose more general mechanisms which solve the same issue but obstructing this proposal for a hypothetical future ideal is not helpful. I have already spent a disproportionate amount of time trying to argue for a single line change...

int-index:
Adding the `Code` newtype is very reasonable from the API design standpoint, it allows us to define instances for `Code m`, which is also the type that is used to type-check quotations. Without baked-in support for `Code`, one would need unnecessary wrapping/unwrapping.

However, it does seem asymmetric with untyped Template Haskell, which worries me. It seems that for symmetry, we should have:

```
newtype Code m a = Code (m (TExp a))
newtype UCode m = UCode (m Exp) 
```

mpickering:
@int-index I think the way you are supposed to construct typed template haskell programs is quite different to untyped so there is already some asymmetric.. and there could be more if you used a different representation type for typed and untyped quotations as well.

There is also the benefits vs potential breakage to consider. Modifying untyped template haskell in the same way would have none of the benefits but break a lot of programs. The typed variant is quite well motivated.

mpickering:
Anyone know the latest with this?

yav:
@mpickering we are waiting for the proposal update, to address Simon's questions.  I see you made some updates, so let me know if you think it is ready, and I'll run it by the committee again.

nomeata:
@mpickering @yav this may have fallen through someone’s cracks?

mpickering:
This has been one almighty journey and I claimed on Dec 6th to have amended the proposal for the comments, was that not sufficient?

nomeata:
@yav?

yav:
Yikes, in a couple of years we might get this done :-)   Sorry, though, I didn't realize you were done with the changes, I'll send to the committee again, hopefully this time around it will be a quick "accept".

simonpj:
Thanks for revising Matthew.  I've read it through again, and have some thoughts
* Please say super-explicitly that untyped TH quotes and splices are unaffected.  Eg `[| e |] :: Q Exp` (is that right?)

* In the motivation you say `Quote m => m (TExp a)` is ugly to read and write.  But with your change you’ll still need to write `Quote m => Code m a`.   So (truth in advertising) it would be more kosher to say "it's easier to read and wrote `Code m a` than `m (TExp a)`.   But I totally get the point about instances of Code.

* It does make me wonder about
  ```
  newtype Code m a = Code (Quote m => m (TExp a))
  ```
  or even
  ```
  newtype Code a = Code (forall m. Quote m => m (TExp a))
  ```
  c.f. the definition of Q:
  ```
  newtype Q a = Q { unQ :: forall m. Quasi m => m a }
  ```
  I'm not close enough to the details to be sure of the consequences here, but it's an obvious question to ask, so obvious that the proposal might want to explain why it's a bad idea.  

  I think you specifically want to parameterise over m, because you anticipate actively using a variety of different monads, but it's not clear to me why this is valuable.  Eg for untyped TH we just use Q. 

* "Nested splicing requires an expression of type `(Quote m, C m) => Code m T`".  What is C?  Quote is a library type class; is C?  No, I think you mean "for any C". And indeed any T.   And is it *required* that the nested splice has a Quote m constraint?  I think not.  And, besides Quote, must it have precisely one other constraint?   I think you mean that the nested splices (here, foo) can have any old constraints, but must return `Code m t`; then the constraints get unioned into the constraint for the enclosing quote.

  All of this is expressed much more naturally and precisely with typing rules; and I know you have them because I've seen them. Can you just make them part of the proposal?

* "The return type of lifting a value".  I think you mean "The return type of the `liftTyped` method of class `Lift`..."

* "A new method is added". Do you mean "method" in the sense of class method?  I think not.  I think you intend to add a new, top-level function to some library module.  But which one?

* `unTypeQ` and `liftCode` simply go back and forth across the newtype, yet they have surprisingly different names.  Maybe `toCode` and `fromCode`?   The "Q" seems particularly surprising.

* Is the Code data contructor exported?  No reason not to, since `toCode` and `fromCode` do the same thing.  But worth specifying.

* Do we need to expose `TExp` at all?   Isn't `Code` enough, leaving `TExp` as an internal implementation detail?  It would simplify the API to *replace* `TExp` with `Code` rather than *adding* `Code`.

int-index:
> I think you specifically want to parameterise over m, because you anticipate actively using a variety of different monads, but it's not clear to me why this is valuable. Eg for untyped TH we just use Q.

#246 has been accepted and implemented, so we don't "just use Q", untyped TH is monad-polymorphic as well. As to why it's valuable, the motivation in #246 applies here as well.

simonpj:
Ah indeed -- I'm so behind the times.  But don't just tell me!  Say so in the proposal!

Then we have
```
[| e |] :: Quote m => m Exp
[|| e ||] :: Quote m => Code m t
```
which is nice.  Also all the stuff about constraints in splices unioning etc, appears (I think) in that earlier proposal too -- so maybe that's not even a change?  It's even implemented, you say.   That is, right now, today,  if `foo :: blah => m (TExp Bool)`, and `bar :: blah2 => m (TExp Bool)` then
```
[|| $$foo && $$bar ||] :: (Quote m, blah, blah2) => m (TExp Bool)
```
is that right?

mpickering:
> is that right?

Yes that is right, it is already how it works and it not modified for the newtype in any special way.


> Thanks for revising Matthew. I've read it through again, and have some thoughts
> 
>     * Please say super-explicitly that untyped TH quotes and splices are unaffected.  Eg `[| e |] :: Q Exp` (is that right?)

Done.

> 
>     * In the motivation you say `Quote m => m (TExp a)` is ugly to read and write.  But with your change you’ll still need to write `Quote m => Code m a`.   So (truth in advertising) it would be more kosher to say "it's easier to read and wrote `Code m a` than `m (TExp a)`.   But I totally get the point about instances of Code.

Done

> 
>     * It does make me wonder about
>       ```
>       newtype Code m a = Code (Quote m => m (TExp a))
>       ```
>       
>       
>       or even
>       ```
>       newtype Code a = Code (forall m. Quote m => m (TExp a))
>       ```
>       
>       
>       c.f. the definition of Q:
>       ```
>       newtype Q a = Q { unQ :: forall m. Quasi m => m a }
>       ```
>       

Using effects whilst doing code generation is crucial and so having access to a monadic context makes writing generators much nicer. See the overloaded quotation proposal about why having `m` is good. 


>     * "Nested splicing requires an expression of type `(Quote m, C m) => Code m T`".  What is C?  Quote is a library type class; is C?  No, I think you mean "for any C". And indeed any T.   And is it _required_ that the nested splice has a Quote m constraint?  I think not.  And, besides Quote, must it have precisely one other constraint?   I think you mean that the nested splices (here, foo) can have any old constraints, but must return `Code m t`; then the constraints get unioned into the constraint for the enclosing quote.

Right I changed this part to hopefully be less confusing but also it is not new behavior. 


>       All of this is expressed much more naturally and precisely with typing rules; and I know you have them because I've seen them. Can you just make them part of the proposal?

I actually don't have typing rules for the overloaded quotations part but I will be producing them shortly in order to finish my thesis as they are only a small extension to what we do have. I don't think typing rules are necessary for a proposal which is about turning a type synonym into a newtype. 

> 
>     * "The return type of lifting a value".  I think you mean "The return type of the `liftTyped` method of class `Lift`..."

Done.
> 
>     * "A new method is added". Do you mean "method" in the sense of class method?  I think not.  I think you intend to add a new, top-level function to some library module.  But which one?

Done.

> 
>     * `unTypeQ` and `liftCode` simply go back and forth across the newtype, yet they have surprisingly different names.  Maybe `toCode` and `fromCode`?   The "Q" seems particularly surprising.

`unTypeQ` is the historical name, I suppose it should be changed to `unType` now. `liftCode` and
`unType` are not inverses though as `unType` produces an `Exp` but `liftCode` lifts a `TExp`. 

> 
>     * Is the Code data contructor exported?  No reason not to, since `toCode` and `fromCode` do the same thing.  But worth specifying.

Yes it is, that was already specified. 

> 
>     * Do we need to expose `TExp` at all?   Isn't `Code` enough, leaving `TExp` as an internal implementation detail?  It would simplify the API to _replace_ `TExp` with `Code` rather than _adding_ `Code`.   And it leaves more wiggle room for the implementation.

Yes we should still expose it as you can still unsafely build `TExp` using `Exp`. At least for now. 

------

I have addressed all of Simon's comments now. If everyone is happy can we please merge this, it has been quite the saga.

simonpj:
> Yes we should still expose it as you can still unsafely build TExp using Exp. At least for now.

But why?   Why not kill TExp altogether in favour of Code?   

You may say "because of back-compat".  But if it's desirable to do this we should describe the destination and then the path to get there.  (Which is standard: deprecate it for a couple of releases.)  But maybe there is an actual reason for keeping it?  If so, let's say why; if not, it's historical cruft that we should assiduously chip away.

mpickering:
@simonpj I do not want to get into this discussion 17 months after I submitted the proposal. But you still need `TExp` because otherwise, how would you write a monadic action which generated some code? The answer now is you write `m (TExp a)` and then use `liftCode`.

simonpj:
You have reason to be impatient because we dropped the ball between Jan and now, which is a bad thing.  But when the proposal was submitted to the committee by the shepherd, I responded within hours, and now I'm responding to you within minutes.  I don't want to accept a proposal I don't understand simply because it's been too long.

> The answer now is you write m (TExp a) and then use liftCode.

But how can I build those `m (TExp a)` things.  I thought the *only* way I could build code in TTH is to use quotes, which (now) makes Code things not TExp thing.  As a user how do I build a `m (TExp a)`?  (Apart from `fromCode` which would of course vanish if TExp did.)  Thanks

yav:
@simonpj  Have a look at https://github.com/ghc-proposals/ghc-proposals/pull/195#issuecomment-551517490

Note in particular the 3rd bullet, it allows you to do things like this:
```
chooseCode :: Code a -> Code a -> Code a
chooseCode (Code c1) (Code c2) =
  liftCode $ do x <- someSideEffect
                if x then c1 else c2
```

The `do` in this example is exactly of the type in question, but it is not built using quotes.

simonpj:
> Note in particular the 3rd bullet, it allows you to do things like this:

Aha!  So the goal is not to expose TExp!  Rather, the goal is to allow you to wrap monadic actions around code.  In that case, perhaps we can provide a function to do precisely that, while keep Code abstract, hiding TExp. Perhaps
```
bindCode :: m a -> (a -> Code m b) -> Code m b
```
Then we can write
```
chooseCode :: Code M a -> Code M a -> Code M a
chooseCode c1 c2 = bindCode someEffect (\x -> if x then c1 else c2)
```
Another way to say it: `Code m a` is a valuable abstraction, and I'm seeking the best API for that abstraction.

yav:
@simonpj the function `bindCode` is mentioned in the "Alternatives" section as `addM`.

Code transformation requires some sort of representation for the code.  Here is an example, using the functions from the proposal:
```
-- | Replace `f 0 x` with `x`
customCodeTransform :: Name -> Code Q Int -> Code Q Int
customCodeTransform targetF m =
  do exp <- unType
     unsafeTExpCoerce $
       case exp of
         VarE f `AppE` LitE (IntegerL 0) `AppE` x
           | f == targetF -> x
         _ -> exp
```

I am not too sure about the proposed changes to `unType` and `unsafeTExpCoerece` though.  My preferred core API would be:
```
liftCode    :: m (TExp a) -> Code m a
examineCode :: Code m a -> m (TExp a)


-- Ways to manipulate `TExp`, for now
unType     :: TExp a -> Exp
unsafeTExp :: Exp    -> TExp a
```
which is equivalent to just exposing the `Code` constructor.   I think keeping the `TExp` abstraction is useful, as it captures in what ways we can examine generated code.  This would allow us to experiment with safer ways of manipulating `TExp` in the future, or even different representations (e.g., using `core` as @mpickering talked about at the last ICFP)

Thoughts?

int-index:
I would much prefer to keep `TExp` as is. It's nice to think of `Code m` as `Compose m TExp`, not as an entirely separate abstraction. And I can also imagine a further generalization in the future such that instead of `Compose m TExp` we could have `Compose m MyExp`, to use typed quotations for DSLs richer than Haskell.

int-index:
To be a bit more precise, let's redefine `Code` as follows:

```
newtype RComp f g a = RComp (g (f a))

type Code = RComp TExp
```

Then, instead of `[|| ... ||] :: Code t`, we could talk about `[|| ... ||] :: RComp lang t` for many languages `lang`, with `TExp` as a special case.

I'm not asking to add this to the proposal. I'm just observing that the proposal as written admits such a generalization, whereas Simon's tweak does not. That's why I'd prefer to keep the proposal as is, with `TExp`.

simonpj:
> Code transformation requires some sort of representation for the code. 

But code transformation -- that is, pattern matching on the code tree -- is precisely what TExp does not allow!  You can build it, but not take it apart.   It's also why TExp is restricted to expressions, not patterns, types, declarations, as for *untyped* TH.

If you want to be unsafe, then don't use TExp: just go to Exp and unsafely back to Code:
```
codeExp :: Code m a -> m Exp
unsafeExpCode :: Exp -> Code m a   -- Unsafe because the Exp might not have type a
```

To put it another way, if we were designing Typed Template Haskell from scratch, would we have TExp as well as Code?  I think not.  There would be no compelling purpose.  (Perhaps @int-index is right that there may be a yet more glorious future, but we aren't there yet.)

Thus far, then, I have seen no compelling case for TExp.  Perhaps it boils down to this: if were were starting from scratch we would have just Code.  But for historical reasons we are going to have TExp too; it clutters up our API with another type, but it makes a smaller change from the status quo.   If that's the case, and the committee is content, I won't stand in the way.  I just want to be clear about the reasoning.

Ericson2314:
I think I am with @int-index. The lack of an exposed constructor for `TExp` is very important for correctness, while the hiding of the monad in `Code` is just for convenience, right? I am not even sure I want this proposal as I don't find `m (TExp a)` weird, but I certainly I don't want to muddle the waters with an abstraction that is trying to solve separate correctness and convenience problems at once, as
```haskell
newtype Code m a = Code (m TExp)
type role representational nominal
```
would be.

mpickering:
Another downside of the `bindCode` approach is that you can't use do notation anymore to write effectful definitions. 

I have not heard any users asking about removing `TExp` and I personally do not see the need to do so.

phadej:
A small experience report:

 I refactored a medium-sized TTH codebase to use `newtype Code m` instead of `TExpQ` . As the codebase had own `newtype` already (to have something which one can partially apply) the change greatly simplified the internal implementation of the library in question.

The downside is that as typed-quotes in released GHCs produce and require `TExpQ`, I had to add `unType` and `liftCode` around these. Otherwise I don't think I needed them anywhere. I did need `Code m a -> ExpQ` and `ExpQ -> Code m a` functions however, as some code generation I do isn't expressible with splice (recursive bindings of static-yet-variable size).

mpickering:
Would be good to accept this proposal soon so that I can implement the patch for 8.12.

simonpj:
> Would be good to accept this proposal soon so that I can implement the patch for 8.12.

Fine with me.

* My personal take is that it's unsatisfying to have both `Code` and `TExp`.  
* I would at least like to have an API that allows me to use `Code` exclusively, and never  need `TExp`.  To do that I need `bindCode` (I don't mind what it's called).   I also need a function to get from `Code` to `Exp`.  I think you have these in your proposed design, but I'd like to double-check.
* I'd love to deprecate `TExp`, explaining why.  Or somehow to encourage everyone to use `Code`
* But I don't want to hold up the proposal on these API-design grounds, especially since others are a lot closer to the TH metal than I am. 

TL;DR: could the proposal summarise, in one place, the entire API of Code, please?

mpickering:
The API for code is already in the proposal (in one place) in the "Proposed Change Specification".

yav:
@mpickering @simonpj let's add `bindCode` and `bindCode_` to the block of "convenience" functions (where `hoistCode` lives).    I think if we add them to a separate module and called them `>>=` and `>>`, then they'd work pretty nicely with the newly accepted "Qualified Do" proposal [1], as you'd be able to write code like this:
```
import qualified Language.Haskell.TH.CodeDo as Code

myExample :: Code M a -> Code M a -> Code M a
myExample opt1 opt2 =
  Code.do
    x <- someSideEffect               -- This one is of type `M Bool`
    if x then opt1 else opt2
```
[1] https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0216-qualified-do.rst

mpickering:
Thanks @yav, I didn't realise that implication of QualifiedDo. I added them into the proposal.

Ericson2314:
I'm sorry, but now I'm even more lost on why we are doing this. We are hiding the monad, and then overloading do notation to use it anyways? (I am more comfortable with, say, `Linear.do` because that seems to be actually a monad in some category, I don't know enough category theory to say `Code.do` isn't some other nice mathematical structure, but it's at least non-obvious.)

I gather the idea is that the monad for renaming is just an implementation detail, and the user should just think about composing syntax? At least this is the impression I've gotten from hearing @mpickering talk about his TTH work. Also, we are just doing this for TTH and TH because there is less worry about breakage, right? There isn't some deeper reason why `m (TExp a)` is bad and `m Exp` isn't, correct?

Yet, nothing I am surmising from above is in the motivation section. The closest item is:

> `m (TExp a)` is ugly to read and write, understanding `Code m a` is easier. This is a weak reason but one everyone can surely agree with.
 
I don't agree with this, but that's because I am skeptical of the premise that the monad is a leaky detail. Far more importantly than what I think is that there *is* an unstated premise: I suspect those that think "`m (TExp a)` is ugly to read" are near-precisely the same people who think the monad is a leaky implementation detail.

It's fine if I have the minority viewpoint---maybe I'm just overly accustomed to the monad and missing the forest for the trees. But, let's at least be explicit in the proposal about where the motivation comes from!

phadej:
@Ericson2314 `Foo a -> (a -> Bar b) -> Foo b` is a module over monad operation. https://ncatlab.org/nlab/show/module+over+a+monad#modules, there are left and right ones (depending from which side you "join").

Ericson2314:
I am just going to retract my parenthetical is anything about `Code.do` is far less important to me than the missing pieces of the motivation.

phadej:
If `TExp` is still exported, shouldn't

```haskell
unsafeTExpCoerce :: m Exp -> m (TExp a) 
```

be kept at current type, and rather new

```haskell
unsafeCodeCoerce :: m Exp -> Code m a
```

to be introduced. It's weird to have function with `TExp` in name to work on `Code`.

Also note, that we have `unType :: TExp a -> Exp` and different `unTypeQ :: TExpQ a -> ExpQ` at the moment. Proposal talks about the first one with the type of second one. Here as well, I'd try to come up with new name, rather than changing the existing signature.

phadej:
A something to consider. I just run (into my previously mentioned experiment), that

```haskell
instance (Lift a, Num a, Quote m) => Num (Quote m) where
    fromInteger x = liftTyped x
    ...
``` 

Would be nice to have. And same for `IsString`.

simonpj:
> The API for code is already in the proposal (in one place) in the "Proposed Change Specification".

My apologies -- it's looking much clearer now, thank you.
Remaining questions:

* What other operations, if any, are available for `TExp`? That is, what
  functions (other than those listed here) have `TExp` in their type
  signatures.  I thikn the answer may be "none", but I'm not certain.
  It would be great to say so explicitly in this proposal; and it
  would be good to include the API for `TExp`, if it has anything
  further.

* Do you really want to specify `handleState` in this API, since it is just a specialisation of `hoistCode`?  And if you do, do you want to specialise for `Q` or could you have
  ```
  handleState :: Code (StateT Int m) a -> Code m a
  ```

* Since we have `liftCode` and `examineCode`, can we make `Code` opaque?
  That gives a little more wiggle room if you ever want to change its implementation later.

mpickering:
@Ericson2314 Using effects if a fundamental part of code generation, for example, if you construct a code generator which constructs division, then you probably want to fail at compile time if you would construct a division by zero. Therefore having the monadic context is a big advantage that TTH has over metaocaml and dotty. Another example, I have implemented let insertion using the monadic context ( https://github.com/mpickering/let-insert/blob/master/LetInsert.hs ).

@phadej 

Thanks, I was going to let all this stuff shake out in the implementation where it becomes clearer.
So I will update the proposal to 

* Introduce `unsafeCodeCoerce`
* `unTypeCode` as the variant of `unTypeQ` and keep `unType`.

@simonpj 

* You can use `unsafeTExpCoerce`, `TExp` or `unType` to interact with `TExp`. This hasn't changed because of this proposal.
* `handleState` is an example of using `hoistCode`. I modified the comment to reflect this. 
* Making `Code` opaque is just annoying rather than serving any measurable benefit. 

I have also started the implementation.

simonpj:
Thanks Matthew.  Can you just add at the beginning of Change Specification that the section gives the complete API for both `Code` and `TExp`?

Also, I was a bit boggled by
```
unsafeTExpCoerce :: Quote m => m Exp -> m (TExp a)
TExp :: Exp -> TExp a
unType :: TExp a -> Exp
```
We really really shouldn't be exposing `TExp` as a constructor.  We should have
```
unsafeExpToTExp :: Exp -> TExp a
```
(It's a pity that `unsafeTExpCoerce` is already taken.)  I know these this is a pre-existing hole in our type system.  But this would be a good moment to fix it.

And there is no reason for `unsafeTExpCoerce` to need `Quote m`; it only needs `Functor m`.   Perhaps not a big deal.

Moreover, like `bindCode`, it's a pure abbreviation, with a two-token definition, so maybe we can give it a definition, as you do for `bindCode`:
```
unsafeTExpCoerce = fmap unsafeExpToTExp
```

mpickering:
Updated the proposal again with your comments. Patch is also available to review:

https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3358

It makes no difference other than potentially annoying users to hide `TExp`. 

The reason `unsafeTExpCoerce` takes a `Quote m` constraint is that it made the implementation a little bit easier. I think I could probably change that quite easily though but not in scope for this proposal.

simonpj:
> It makes no difference other than potentially annoying users to hide TExp.

Bu tit makes a big difference?   If I see "unsafe" something I know something is unsafe. If I see `TExp x` I have not such clue.  Yet every such occurrence is a potential seg-fault.  Asking clients to use `unsafeExpToTExp` is a positive feature, I think.

I accept that this bug predates this proposal, but the proposal refines the client API of Code and TExp.  Let's get it right.

> The reason unsafeTExpCoerce takes a Quote m constraint is that it made the implementation a little bit easier.

Interesting.  The implementation is:
```unsafeTExpCoerce :: forall (r :: RuntimeRep) (a :: TYPE r) m .
                      Quote m => m Exp -> m (TExp a)
unsafeTExpCoerce m = do { e <- m
                        ; return (TExp e) }
```
That uses `Monad`.  Does `Quote` make it easier.  And as I say `fmap unsafeExpToTExp` is shorter still.  Maybe I'm missing something.

Ericson2314:
> Using effects if a fundamental part of code generation, for example, if you construct a code generator which constructs division, then you probably want to fail at compile time if you would construct a division by zero. Therefore having the monadic context is a big advantage that TTH has over metaocaml and dotty. Another example, I have implemented let insertion using the monadic context ( https://github.com/mpickering/let-insert/blob/master/LetInsert.hs ).

Yes, I agree the monad is fundamental! That's why I think it's weird to put i in a newtype. I've never seem anyone newtype `M (G a)` where `G` isn't even a functor (i.e. has no relevant semantics WRT the Monad, it's just somewhat we happen to be producing), and yet that we are doing this here.

The motivation of "just avoiding some wrapping/unwrapping" still seems incredibly weak to me, put forth by those that *write* the code at the expensive of those that need to *read* the code later. Imagine if you need to read your coworker's or classmate's TTH code, and are new to this stuff yourself. Would you rather read something with plain old wrapping/unwrapping, or a  `Code.do`?

Ericson2314:
On a meta note, I feel bad being the chief nay-sayer here. I have been and am working with @mpickering on other things, for example. And I completely understand it's been frustrating for him to see this to drag on and on, especially when it seemed on the verge of acceptance. I take no joy in dragging it on further. But just because a proposal is overdue getting resolved doesn't mean it deserves to be accepted.

I am really surprised there really hasn't been much discussion of the motivation, one way or the other, from others. Can somebody who is *not* a current TTH user say why this is compelling to them?

mpickering:
@simonpj 

You can't get into the situation of a segfault by using `unsafeTExpCoerce` (or `TExp`) so it is less unsafe than say, `unsafeCoerce`.  Modifying `TExp` is out of scope for this proposal in my opinion. If it had been suggested at the start then perhaps I would be more willing to modify the scope but it's far too late for that. I do not wish to modify how `TExp` behaves.

> Interesting. The implementation is:

I am talking about when I implemented Overloaded Quotations internally to the compiler, not the Haskell implementation. In that case you have convenient access to a wrapper which applies the `Quote` dictionary but getting access to the `Functor ` constraint was a bit more work. If you open a GHC ticket then I could probably modify this. It is also not in scope for this proposal though, which does not modify this function. 

In this thread, a large proportion of known users indicate that this change would be beneficial and yet I have spent an inordinate amount of emotional energy attempting to make it happen. Please can we move on from this proposal!

phadej:
> Putting Code in heterogeneous collections. 

```diff
-Code [|| 1 ||] :* Code [|| True ||] :* Code [|| "foo" ||] :* Nil
+[|| 1 ||] :* [|| True ||] :* [|| "foo" |]] :* Nil
```

having written (yet unpublished) library which has code like that, the upper variant is simply not user friendly for public API (It would fine internally, and that's what it looks like now there).

Similarly, the using of EDSL would be not nice with wrapping and unwrapping in the *user code*. Instances can be ugly, but writing

`if_ (Code [|| foo $$x ||]) ... ...` has that unnecessary `Code` there. It's just noise around already quite noisy `[|| ... ||]`.

int-index:
@phadej The `:*` example is, once again, a testament to the issues of heterogeneous collections, which is not exclusive to TTH. If you have `HList Identity`, you also must write `Identity 1 :* Identity "hello" :* Nil` instead of the much nicer `1 :* "hello" :* Nil`.

It's a reason to define different APIs for heterogeneous collections, not for TTH.

As with `lookupTExp` and `insertTExp`, my suggestion would be to define another operation:

```
consTExp e ts = Code e :* ts 
```

and then write it as:

```
consTExp [|| 1 ||] $
consTExp [|| True ||] $
consTExp [|| "foo" |]] $
Nil
```

----

The `if_ (Code [|| foo $$x ||]) ... ...` indeed demonstrates the extra noise, but, once again, the same would be the case for `Identity`:

```
if_ (Identity x) ... ...
```

----

> and don't make anything worse

That's a bold claim. It does make the use of monadic effects more complicated, hence the talk about `bindCode`, `Code.do`, etc.

----

Overall, I see how `Code` makes things nicer, but baking it into the compiler still seems like a half-measure to me. A proper solution would address the extra noise both in the `Identity` case and the `Code` case.

Ericson2314:
OK thank you so much @int-index --- I no longer need to feel like the sole crazy person out here.

In fact agree with your -0.5 technical +1 social---this thing is not a big deal when way or the other, and I don't want to just waste @mpickering's time with tons of bureaucracy and TTH is basically being completely revamped.

But, if we're going to merge this for the social reasons and not the technical ones, we should at least be explicit about that. Just taking the motivation on blind faith seems no good.

Ericson2314:
> having written (yet unpublished) library which has code like that, the upper variant is simply not user friendly for public API (It would fine internally, and that's what it looks like now there).

I do think there is a desire to make the common cases for TTH as convenient as possible. But as @int-index says.

> Overall, I see how Code makes things nicer, but baking it into the compiler still seems like a half-measure to me. A proper solution would address the extra noise both in the Identity case and the Code case.

I agree completely. And let's keep in mind the risk here. If one feature takes some measures to be more usable, it's mostly fine. But If every feature (e.g. TTH) tries to hack around deficiencies in other features (e.g. heterogeneous collection ergonomics), we end up with a more complex language where disparate features view each other with suspicious.

For the sake of a whole, I am general pro *not*-hacking around things but waiting for fix the root causes of problems. We have to balance the completing interests of more beginner adoption of TTH now, vs a sensible API in conformance with the general idioms for writing Haskell later. Or, as we usually say, I am for avoiding success-at-all-costs.

yav:
Thank you everyone for you input---I think we've clarified the proposal, and maybe even come up with some new ideas for future improvements/interesting research directions.

Since this is a simple change, which names a common abstraction, and is clearly making current typed TH users happy, I think it is reasonable that we should accept it, and as the shepherd for this proposal I'll recommend to the committee that we do so.

PR #99, opened 2017-12-20
Explicit specificity

goldfirere:
This proposal introduces new syntax `typeRep :: forall {k} (a :: k). ...` (the braces are new) to allow a user to quantify a variable without affecting downstream users who might use visible type application.

[Rendered](https://github.com/goldfirere/ghc-proposals/blob/explicit-specificity/proposals/0000-explicit-specificity.rst)

PR #81, opened 2017-10-14
Syntax for visible dependent quantification

goldfirere:
This proposes a concrete syntax for a kind that has existed since GHC 8.0, but cannot currently be parsed.

[Rendered](https://github.com/goldfirere/ghc-proposals/blob/forall-arrow/proposals/0000-forall-arrow.rst)

PR #80, opened 2017-10-11
Type-level type applications

treeowl:
Allow `TypeApplications` syntax to be used in types as well as
terms.

[Rendered](https://github.com/treeowl/ghc-proposals/blob/type-level-type-app/proposals/0000-type-level-type-applications.rst)

PR #196, opened 2019-01-18
Named type arguments

bakhtiyarneyman:
This is a proposal extending type application syntax with the syntax of record field updates.

[Rendered](https://github.com/niobium0/ghc-proposals/blob/patch-1/proposals/0042-named-type-args.rst)

### DIFF
--- a/proposals/0042-named-type-args.rst
+++ b/proposals/0042-named-type-args.rst
@@ -9,7 +9,7 @@ Proposal title
 .. implemented:: Leave blank. This will be filled in with the first GHC version which
                  implements the described feature.
 .. highlight:: haskell
-.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/196>`_.
             **After creating the pull request, edit this file again, update the
             number in the link, and delete this bold sentence.**
 .. sectnum::


int-index:
I am very much in favour of this, and I had similar thoughts before. However, I would prefer round parentheses there, i.e. `f @{a = Int}`, as this syntax can be later generalized to term-level named parameters, `f (a = Int)`, without any conflict with record syntax.

int-index:
@gridaphobe That wouldn't work, as this denotes the equality constraint. We can already have

```
f :: forall (a :: Constraint). blah
```

and invoke it as

```
f @(Eq Int)
f @(String ~ [Char])
f @(a ~ Int)
```

And what's wrong with `=` anyway?

gridaphobe:
Ugh, good point.. I don't have a problem with `=`, I just thought that `~` looked more natural given its existing usage.

int-index:
> The above looks minimal and consistent to me.

That won't fly, as `f { arg = 5 }` is already taken by record update syntax.

> it is my opinion that it will be difficult (if possible) to ever have a simple, unsurprising solution for term-level named arguments.

I believe the `foreach` quantifier (proposed in #102) is a good way forward. It allows giving names to term-level parameters:

```
if' :: foreach (cond :: Bool) -> a -> a -> a
if' True t _ = t
if' False _ f = f

max a b = if' (cond = a > b) a b
```

Besides, we'll have to deal with this case even before `foreach`, when visible dependent quantification #81 comes to terms, or even sooner, due to its interaction #80

Ericson2314:
I wonder if @int-index's term-level variation can be used to define GADT records better.

aspiwack:
I'd like to point out a direct cost of this proposal: as it stands, and if I understand it properly, it implies that changing the names of type variables in a type declaration (α-renaming as it's fondly known) is _always_ compatibility-breaking.

Which, by the way, means that the mere existence of this extension (as described), changes the semantics of Haskell as a whole (in current Haskell, α-renaming is semantic-preserving).

I'd suggest, therefore, for this proposal to propose an explicit way for types to declare arguments which can be referred to by name (and hence cannot be renamed without breaking compatibility).

There is related work by the way: Coq has a syntax which works more or less as the proposal currently stands (though there is some funny business due to name shadowing). Ocaml has (for term-level things) a notion of labelled argument which can be referred to by name.

int-index:
@aspiwack There was a similar concern in #99, so I will copy my counter-argument.

----

I see the value of alpha-equivalence, but I don't think we are violating it here, just making it non-local.

We have that

```
f :: forall a. a -> a
g = f @(a = Int)
```

is equivalent to

```
f :: forall b. b -> b
g = f @(b = Int)
```

This name is not a part of program's semantics — it does not affect runtime behavior in any way. I would be more worried about something like this:

```
data D = D deriving Show
main = print D
```

which changes its behavior when I perform a renaming:

```
data T = T deriving Show
main = print T
```

and yet we seem not to worry about this.

At the same time, with named `forall` I can always change an API to use different names:

```
f :: forall a. a -> a

f' = forall b. b -> b
f' = f @(a = b)
```

so it's very well-contained.

int-index:
> I'd suggest, therefore, for this proposal to propose an explicit way for types to declare arguments which can be referred to by name (and hence cannot be renamed without breaking compatibility).

@aspiwack We are concerned with users of `TypeApplications`, who are already subject to another similar issue, namely the _ordering_ of type variables being observable.

When we have `map :: forall a b. (a -> b) -> [a] -> [b]`, we have arbitrarily chosen this variable order over `forall b a`, effectively assigning indices `0` and `1` to the types of the input and the output. 

Changing these indices into `1` for the input and `0` for the output is a breaking change in the same way:

```
map @Int @Bool   -- will break if we change `forall a b.` to `forall b a`
```

Because effectively what we have is

```
map @(0 = Int) @(1 = Bool)   -- will break if we change `forall a b.` to `forall b a`
```

It does not seem to be a huge leap to replace these numeric indices with strings.

Bottom line is, since we don't have a mechanism to hide the ordering of type variables, we should not complicate the language by adding a mechanism to hide the naming of type variables.

goldfirere:
A long time ago, we (@sweirich, Hamidhasan Ahmed, and I) debated this design when originally coming up with type applications. And the reason we avoided it was precisely @aspiwack's https://github.com/ghc-proposals/ghc-proposals/pull/196#issuecomment-456814971.

That said, I'm now ready to change my stance, in the light of experience. This does mean that library-writers now leak their variable names: changing the name of a type variable would now induce a major version bump in a library. This is regrettable, but I think the benefits outweigh the costs.

I imagine the practical cost is minimal: are library-writers frequently changing type variable names? I doubt it. Unfortunate here is that library-writers who don't know about this hypothetical addition to the language are affected, too. (Though, as pointed out above, this is true today about variable ordering.)

So I'm broadly in favor. But now, on to syntax: I think the braces notation is better, as it's more consistent with records and looks more familiar. Note that there is actually no term-level conflict (with a hypothetical future extension of this proposal to include term-level named parameters). Under that hypothetical, we would have three uses of braces in terms:

1. Constructor applications: `let record = K { field1 = 1, field2 = False }`
2. Record update: `record { field2 = True }`
3. Function application with named arguments `f { arg2 = 'z', arg1 = 5 }`

We can disambiguate these. (1) is easy: the thing to the left of the braces starts with a capital letter. But telling (2) and (3) apart is also easy: in (2), the thing to the left of the braces has a record type; in (3), that thing has an arrow type.

Furthermore, I wonder if we can use a better syntax than `foreach` for named arguments. How about

```
f :: { arg1 :: Int, arg2 :: Char } -> Double
```

Now, GADT record data constructors fit naturally. Note that the notation here wouldn't really be a type, but something else I'll call a *signature type*, allowable only at top-level in signatures. The argument names would be attached to the identifier `f`, separate from `f`'s type (which would be `Int -> Char -> Double`. Of course, dependency would be allowed, too: `g :: { k :: Bool, p :: Proxy k } -> Int`. Why not just use `foreach`? Named argument application is most useful when the arguments can appear out of order, and out-of-order application just doesn't work well with curried, one-at-a-time argument passing. So I think it's better to have a new construct. Left undesigned is how to mark the difference between relevant and irrelevant parameters in this new syntax.

Returning to the present proposal: I think sticking with braces has a nice future ahead of it (which is why I bothered with all the above), and doesn't conflict with the present.

Ericson2314:
Yes the question is not of alpha equivalence, but of scope. This is related to defining macro hygiene as *just* respecting alpha equivalence. There is the name-resolution scope of an identifier, and the binding scope of a variable. The former is prior just as name resolution is prior to evaluation. For most things, the two scope coincide, but for this and record puns, the two do not.

This gives me an idea, we scrap the explicit specificity proposal and instead use `{a :: k}` in a quantifier to indicate the binding is *exposed*. Only exposed bindings can be used to assign arguments to parameters by name. This has the very useful value of preventing people unaware of this proposal from introducing breakage. Old style:
```haskell
f :: forall a b. ...
f :: forall b a. ...
```
Only new-style:
```haskell
f :: forall {a} {b}. ...
f :: forall {b} {a}. ...
```
are not.

This also answers my earlier question of GADT records. With the term-level generalization of this feature:
```haskell
data Foo = Foo { a :: Int, b :: String }
```
is just sugar for
```haskell
data Foo where
  Foo :: forall { a :: Int} -> { b :: String } -> Foo
```
Even this could make sense if we don't regeneralize the type!
```haskell
let f = Foo
in case f { a = 1, b = "asdf" } of
  Foo { a = a, b = b } -> ...
```

Ericson2314:
Also I've had sitting almost finished a WIP proposal for a new update syntax https://github.com/Ericson2314/ghc-proposals/blob/constructor-update/proposals/0000-constructor-update.rst . That syntax would allow us to @goldfirere's deprecate (2) and avoid any type-aware disambiguation of any of the 3 use-cases those 3 syntaxes currently represent.

Now I have motivation to finish it...

int-index:
> f :: { arg1 :: Int, arg2 :: Char } -> Double

@goldfirere This syntax I would rather reserve for anonymous records.

> I think the braces notation is better, as it's more consistent with records and looks more familiar.

And how is this better, given that it has no relation to records? Not all named things have to be records.

> Note that there is actually no term-level conflict

Interesting, so you propose type-based disambiguation here. I see two drawbacks here.

First is that we now don't have the ability to talk about constructors with named parameters, `K { a = ... }` always means record construction and we have no way of saying `K (a = ...)`.

Second is that we get worse type inference, since we can't from looking at `f { a = ... }` tell if `f` is supposed to be a function or a record type.

This code will break:

```
class F t where
  f :: t

data Rec = MkRec { a :: Int } deriving Show

instance F Rec where
  f = MkRec 0

instance (arg ~ Int, res ~ Rec) => F (arg -> res) where
  f = MkRec

r :: Rec
r = f { a = 1 } -- which instance do we pick?
```

If we use the `f (a = 1)` for named arguments, then `f { a = 1 }` keeps working. Today it works because it's unambiguously a record update:

```
ghci> f {a = 1}
MkRec {a = 1}
```

> Named argument application is most useful when the arguments can appear out of order, and out-of-order application just doesn't work well with curried, one-at-a-time argument passing.

It works alright, I have https://hackage.haskell.org/package/named for this exact purpose and I'd be happy if I could replace this library with something that works out of the box.

There's value to out-of-order one-at-a-time argument passing: partial function application. For example, if I have

```
fn ::
  foreach (foo :: Int) ->
  foreach (bar :: Int) ->
  foreach (quux :: String) ->
  Int
```

I might want to

```
foldr (fn (quux = "str")) [1,2,3]
```

which wouldn't work with your all-at-once argument passing.

Ericson2314:
@int-index 

> This syntax I would rather reserve for anonymous records.

This is most convincing to me.

> And how is this better, given that it has no relation to records? Not all named things have to be records.

Under my variation records are just extra functionality for constructors + named quanitfier vars.

> Interesting, so you propose type-based disambiguation here.

See my proposal referenced above for record update with fresh syntax. We can turn off the current record update syntax in any module with the conflict.

> First is that we now don't have the ability to talk about constructors with named parameters, `K { a = ... }` always means record construction and we have no way of saying `K (a = ...)`.

You mean type parameters? What about
```haskell
K { @a = ... }
```

> It works alright, I have https://hackage.haskell.org/package/named for this exact purpose and I'd be happy if I could replace this library with something that works out of the box.

Yes. I do want that too.

int-index:
> It may be worth noting that we would not be able to support out-of-order, one-at-a-time application in types, due to the lack of type-level lambda.

We can support out-of-order, first-n-at-once in types, right? This wouldn't require type-level lambda.

```
data T a b c = ...

type T' = T (b = Int) (a = Char)
```

Not all-at-once (`c` is omitted), but still out-of-order (`b` before `a`).

int-index:
> You mean type parameters? 

@Ericson2314 No. I'd say under your variation, that would correspond to partial record construction.

Your example is this:

```
data Foo where
  Foo :: foreach { a :: Int} -> foreach { b :: String } -> Foo
```

I'd like to be able to

* `Foo (a = 10)`, inferred type `foreach { b :: String } -> Foo`
* `Foo (b = "hi")`, inferred type `foreach { a :: Int } -> Foo`

whereas `Foo {a = 10}` actually corresponds to `Foo {a = 10, b = undefined}`, inferred type `Foo`.

int-index:
> Fair, but `Foo {a = 10, b = undefined}` is like the worst sugar ever. 

@Ericson2314 This is definitely a part of the record system that I wouldn't miss, yes. It's just that we have it right now and I wouldn't want named parameters get blocked on its removal.

> With the latter, I'm tempted to try to understand what `T (b = Int)` is, all by itself, and that isn't a thing.

I'm approaching this with a "we don't have type-level lambdas *yet*" mindset, so I'd expect `T (b = Int)` become a thing in the future.

As to the `(b = Int, a = Char)`, I'm a bit concerned. With `T (a = Int) (b = Int)` if I remove the `=`, `T Int Int` is still application, but with the comma syntax if we remove the names, it's suddenly a tuple? I don't have a technical objection, though, except for the fact that it's even more syntax to support/learn.

simonpj:
Interesting discussion.

Let me make just one point: it would be sad (and odd) to provide a mechanism for named type arguments without also supporting named value arguments. This is discussed above -- but I hope the proposal itself will grow to accommodate it.

niobium0:
I find the argument about type inference to be a strong one. But I'm also unhappy about ``(a = 1, b = 2)``. It looks too much like a tuple, so to me it gives an impression of an entity that
1. Respects the order.
2. Has fixed in number of elements.

But both of these are false for the problem that we are trying to solve! This means that we are eroding the cohesiveness of syntax as it stands today, making it counterintuitive and hard to learn. 

``f (b = 2) (a = 1)`` goes even further in making the syntax confusing because I'm so used to perceive things that are next to an arrow and in parentheses to be interpreted positionally.

I see an option of using ``<-`` instead of ``=``, i.e. ``f { a <- 1, b <- 2 }`` for named arguments. There would be a small bonus in that the following would not have implicit undefineds.
````
data T = T { a :: Int, b :: Int}
f = T { b <- 1}  
````

> f :: { arg1 :: Int, arg2 :: Char } -> Double

I really like this! 
 
> @goldfirere This syntax I would rather reserve for anonymous records.

@int-index I'd like to understand what exactly are the features that you have in mind when you say "anonymous records". If it's literally just "records as we have today but with no names" maybe 
````
f :: (arg1 :: Int, arg2 :: Char) -> Double
```` 

could be a viable alternative (if we ever go in that direction). 

If you are thinking about something along the lines of row polymorphism, perhaps we could reserve 
````
f :: {arg1 :: Int, arg2 :: Char |} -> Double
g :: {arg1 :: Int, arg2 :: Char | r } -> Double
````

**Comment by int-index** [comment]
`f { a <- ... }` suggests some monadic/applicative nature of this binding, which it does not have. `=` is much better.

>  If it's literally just "records as we have today but with no names"

Yes. And I think using braces for records makes more sense than anything else we could come up with, so I suggest we don't repurpose this syntax for anything else.

**Comment by aspiwack** [comment]
A few remarks:

> Yes the question is not of alpha equivalence, but of scope

I'm afraid this is a distinction without a difference. This proposal does break the semantic property that I can rename the name of a variables in a type in isolation. I'm not saying that this should block the proposal, I'm saying that it should be documented in the _Costs and Drawbacks_ section.

> That said, I'm now ready to change my stance, in the light of experience. This does mean that library-writers now leak their variable names: changing the name of a type variable would now induce a major version bump in a library. This is regrettable, but I think the benefits outweigh the costs.
> 
> I imagine the practical cost is minimal: are library-writers frequently changing type variable names? I doubt it. 

Let me share a few thoughts on that. I've used much this feature in Coq for a long time (though Coq's feature has idiosyncrasies of its own, due to shadowing, which shouldn't affect Haskell). And I've never missed it one bit in Haskell.

I don't know exactly why. The renaming thing is, in fact, rarely a problem (though not never! even if, in Coq, it's been here since the beginning and everyone is aware of it). I think it boils down to the fact that people just don't choose their type variable for addressing (`Left (b=Int) True`…).

In Ocaml, by contrast, there is a syntax `f ~x y = …` meaning that `x` is a label which can be addressed by name (but not `y`), the name appears in the type signature and everything (which, in our particular use-case, is a non-issue). It's not use everywhere (far from it), but when you do use it, you choose the name more carefully, develop conventions and so on.

I'd like to know why such a syntax to mean that an argument is addressable by name (rather than only by position) would be a bad idea in the views of the various commenters here. The discussion should also be summarised in the _Alternatives_ section.

**Comment by int-index** [comment]
> I'd like to know why such a syntax to mean that an argument is addressable by name (rather than only by position) would be a bad idea in the views of the various commenters here.

1. Quite simply because Haskell is oversaturated with syntax and every addition needs a really compelling reason.

2. My vision is that the user must explicitly use `foreach` or `forall` in a type signature for the argument to be addressable by name, which is already enough of syntactic cues, and I don't want to write OCaml's `~` in addition to that (which is taken by irrefutable pattern matching in Haskell, by the way, see point 1).

**Comment by aspiwack** [comment]
@int-index Do remember, though, that you don't need to convince me. My opinion is immaterial, and I don't have much of one anyway. What needs to happen is a proposal document which weighs the pros and cons and present them to the committee (and, hopefully for you, convince them that you are taking the right path).

Two remarks:
- The use of `forall` seems rather to reinforce the feeling that the variable must scope over the definition and nothing else, than signify that the name is exported. It also specifies an ordering for the variables. So, in that case, is it convincing to add new syntax (as per your point 1) just to spare a few characters?
- The use of a special syntax to signify that a particular argument is addressable by name nicely extends to regular function arguments. As in https://github.com/ghc-proposals/ghc-proposals/pull/196#issuecomment-457626343 .

**Comment by jvanbruegge** [comment]
Just to leave my two cents:
- On type level I am quite indifferent. I never had the need for this, but I can imagine others do
- I absolutely dislike named value level arguments. They feel like a special form of row polymorphism, in fact you can write a type class with an associated type family to convert a function taking a record to a curried function and another one to convert back. So introducing a special syntax and semantics just for named function arguments is unneccesary

**Comment by jberryman** [comment]
So as I read it this just seems to replace one flaw of `TypeApplications` (ambiguity when constraints are reordered) with another (ambiguity when constraints are renamed) while adding more syntax. In fact it adds flaws because presumably the old ambiguous index-based syntax is not set to be deprecated...

@goldfirere writes

> This does mean that library-writers now leak their variable names: changing the name of a type variable would now induce a major version bump in a library. This is regrettable, but I think the benefits outweigh the costs.

Personally I disagree since I don't have much interest in using type applications myself at this point (don't like the syntax and they don't seem to solve a significant problem for me, though I understand there's justification for them in dependent haskell plans), and I'd rather not accept this burden as a library writer. (But I would certainly do my best if that's what the community decides.)

This is a significant departure from the way that adding extensions has gone in the past I think.

> Furthermore, I wonder if we can use a better syntax than foreach for named arguments. How about
> 
>      f :: { arg1 :: Int, arg2 :: Char } -> Double

I think this syntax is friendly and intuitive; many people get by using haskell without ever needing to really learn what a "quantifier" is , which is not a bad situation. 

By the way is there a proposal for named arguments or anonymous records that I can look at? I'm a bit confused how that relates to this discussion, but I read @int-index 's "_This syntax I would rather reserve for anonymous records_" and my first thought is these ought to be the same proposal (i.e. "named arguments" and "anonymous records" seem like they ought to be the same thing).

... then again what really is the value here? A library writer can already define a new record type paired with their function and export it and this seems nearly as lightweight for the user...

### Punning off record syntax, generally

I think using `{a = Int, b = Char }` or whatever (in whatever proposal we're talking about) is fine, but I'd like to point out that IMO the use of `=` in haskell's syntax everywhere where it doesn't mean "_= like in math_" is a wart. 

So `=` is great in functions, obviously. It's also great in `type` declarations. It's not good in record syntax (where it represents a kind of assignment, like in C), and it's terrible in `data` declarations (where it means AFAICT "here comes some different stuff").

It might be worthwhile to think carefully about all the uses of `=`, `->`, `<-`, in the language if there is a possibility to come up with something more consistent for this family of proposals.

### Alternate proposal sketch 

I think I would be in favor of doing

- deprecate `TypeApplications`
- introduce a new syntax which doesn't work off the notion of "application" (as I've written before, and which most people seem not to agree with, I simply don't think "applying" a function to types makes much sense unless you know something about GHC internals, or perhaps about dependently typed languages); this could be simply removing the space, like `fmap@Maybe`, or something... no strong opinion about alternatives
- functions with a single constraint (no ambiguity) can be used without names (like `fmap@Maybe`)
- where a function has multiple constraints the user _must_ use the named syntax (where names are taken from normal constraint signature; this proposal and `TypeApplications` in general become sort of useless if you insist that now all libraries must be modified to use some named type parameter syntax), so e.g. `mapM@(m ~ IO, t ~ [])` , or something

I think on the last point I'd agree that it's unlikely that library writers would _swap_ two constraint variables (resulting in a confusing type error for users), rather than _change_ one or both (resulting in a more informative error at the callsite, specifically related to this new extension).  I think such (hopefully rare) errors should be considered the user's problem since they are using the extension but maybe that's unreasonable :P

**Comment by int-index** [comment]
> So, in that case, is it convincing to add new syntax (as per your point 1) just to spare a few characters?

@aspiwack I don't think that I argued in favor of adding new syntax to spare a few characters. What do you mean by that?

> The use of a special syntax to signify that a particular argument is addressable by name nicely extends to regular function arguments.

For dependent functions we need a way to give names to function inputs that could be mentioned both at the term level and the type level, why don't we use the same syntactic mechanism for named parameters? And seeing parameter names with `:t` is a nice consequence of this design.

Your argument is that we need to give authors a way to hide parameter names from client code, but we already have a mechanism for hiding things: module export lists.

```
module M (f, g(arg1, arg2)) where

f :: forall arg1 arg2. blah -- parameter names hidden
g :: forall arg1 arg2. blah -- parameter names exported
```

> What needs to happen is a proposal document which weighs the pros and cons and present them to the committee

Yes, I think @niobium0 is working on that, being the author of the proposal? I could write my own proposal if that's what you are suggesting, but I would rather reach consensus here.

----

> I absolutely dislike named value level arguments. They feel like a special form of row polymorphism, in fact you can write a type class with an associated type family to convert a function taking a record to a curried function and another one to convert back. So introducing a special syntax and semantics just for named function arguments is unneccesary

@jvanbruegge I do not fundamentally oppose the idea of packaging up named arguments into anonymous records, but there are problems with this approach that I do not know how to solve yet:

1. RankN fields in an anonymous record would require `-XImpredicativeTypes`, which are notorious for not working properly. I don't know how to make them work in GHC.
2. You don't have any backwards compatibility story for code that uses positional calls. A library that decides to provide named arguments would have to drop support for positional arguments, breaking all client code.
3. Function application with records would require new syntax anyway, even if in the form of a user-defined operator. Imagine you have `f :: {a :: Int, b :: Bool} -> r`, then `f { a = 1, b = True }` is actually syntax for record updates, so perhaps you could disambiguate it with parentheses, writing `f ({a = 1, b = True})`, or with an operator, `f $ {a = 1, b = True}`. Still, this does not give you partial application, `f $ {a = 1}` is a type error, so we need some new operator, perhaps `f $$ {a = 1}`. This might suit your tastes but I would prefer the less cluttered `f (a = 1)`.
4. Since you are not opposed to named type-level arguments, it now means that our mechanisms for type-level and term-level arguments are quite different. However, the distinction between type-level and term-level is quite blurry, and with `foreach` we would have arguments available on both levels. So any vision of named arguments based on anonymous records would have to incorporate type-level arguments and dependencies between arguments (e.g. `foreach k -> foreach (a :: k) -> r), and this is beyond the capabilities of any anonymous record system that I know of.

In other words, modelling named arguments with anonymous records robs us of higher rank inputs, dependencies between arguments, convenient syntax for partial application, and backwards compatibility. In addition to that, uncurrying a function (by packaging up all its inputs in a single data structure) might change its performance characteristics.

Contrast that with natively supported named arguments (for both types and values), this is a relatively non-invasive extension to the current systems that we have.

----

> So as I read it this just seems to replace one flaw of `TypeApplications` (ambiguity when constraints are reordered) with another (ambiguity when constraints are renamed) while adding more syntax. In fact it adds flaws because presumably the old ambiguous index-based syntax is not set to be deprecated...

I think you meant "breaking change" instead of "ambiguity", there's no ambiguity in either case. Anyway, this isn't quite a flaw, how would you like in a world where we didn't use names at all and referred to library functions by their number, e.g. `f0`, `f1`, `f2`, `f3`, `f4`, etc? Having to remember the right function numbers would be pretty miserable in my opinion, and I don't see why we don't hold functions parameters to the same standard.

Names are easier than positions for humans.

> ... then again what really is the value here? A library writer can already define a new record type paired with their function and export it and this seems nearly as lightweight for the user...

@jberryman There are two issues with using non-anonymous records for function inputs:

* Namespace pollution. You end up with a type `FooArgs` and a constructor `FooArgs` for each function `foo`. And the calls become unnecessarily verbose, looking like `foo def{...}` at best and `foo FooArgs{...}` at worst.

* No support for partial application whatsoever. For `n` fields I would need `n!` records for all variants of partial calls. (We have `n` options what field to supply first, then `n-1` options what field to supply second, etc). I'm not looking forward to defining 120 records for a 5-parameter function and class instances to connect them.

These issues could be addressed by anonymous records, at the expense of introducing other issues (see above).

* It's not good in record syntax (where it represents a kind of assignment, like in C)

You mean records updates? Then I agree. But for record construction it's nothing like mutable C-style assignment, and same goes for named parameters where it's pretty close to math usage. `f (a = 1) (b = 2)` is, in fact, semantically equivalent to `let a = 1; b = 2 in f a b`, modulo ordering concerns.

**Comment by jberryman** [comment]
> You mean records updates?

Yeah, and that makes sense, thanks.

**Comment by nomeata** [comment]

**Comment by chshersh** [comment]
> it would be sad (and odd) to provide a mechanism for named type arguments without also supporting named value arguments.

I was thinking about named value arguments for Haskell for quite a long time, and I'm convinced that having them in Haskell would be a great thing and a significant improvement. But it looks like the current proposal is only about named type parameters only. It's an interesting question, whether there should be one single proposal for both values and types or two separate proposals. However, how about the following idea: in the spirit of the proposed syntax here one can use `(@a = Int)` for specifying named type arguments and `(x = 3)` for specifying named value arguments? This looks consistent with the rest of the Haskell at least.

Let me give you an example of a function that can use both type and value named arguments in some imaginary Haskell-like syntax:

```haskell
splitOn :: forall f a . (Splittable f, Eq a) => (delim :: a) -> (sequence :: f a) -> [f a]

splitPath :: FilePath -> [FilePath]
splitPath = splitOn (@f = [], delim = '/')
```

This is just an idea of how we can combine type and value named arguments. It looks like syntax is the most challenging part of such proposals.

int-index:
>  It's an interesting question, whether there should be one single proposal for both values and types or two separate proposals.

There's no need for different syntax for types and terms, as `@` is a visibility override and has only a coincidental relation to the type/term distinction. That is, I see no problem writing this, assuming VDQ in terms:

```
id :: forall a -> a -> a
id a x = x

t = id (a = Int) (x = 5) 
```

Besides, `@` clearly overrides the entire parameter and not just the parameter name, so it should be `@(a = ...)` rather than `(@a = ...)`:

```
id :: forall a. a -> a
id x = x

t = id @(a = Int) (x = 5)
```

> It looks like syntax is the most challenging part of such proposals.

The most challenging part of this proposal is whether the parameter name should be associated with an identifier or an expression. Can an anonymous lambda have named parameters? Where would it be reflected, in its type? I'm somewhat suspicious of higher-rank named parameters.

LeanderK:
This proposal did not receive much attention, but I want to say that I would very much like named type-arguments.

Type applications are different from normal function applications because, while I could use them all the time, type inference just solves the problem. I often only need to pin one type-parameter down and if that happens to be the third I need to do `f @_ @_ @a`, which I think is pretty unreadable, surprisingly hard to comprehend and easy to mess up (count the "holes" and then mentally skip number of parameters in the called function to understand what you just did). I think named type-parameters are easier even if the names end up being just one character, like `c` in `f @(c = a)`. Then I can just mentally substitute c with my parameter a in the function signature.

goldfirere:
Without re-reading this proposal: I have grown more supportive of this approach over time. If I were to go back in time to when we were designing visible type application, I probably would have used a named-argument approach instead, in light of experience.

phadej:
Agda have both (named and positional), and it's convenient. For example you might want to skip over kind variable, but then the positional application is fine:

```haskell
f :: forall k (a :: k) (b :: k) . ...
g = f @_ @'True @'False
```

could be written as 

```haskell
g = f @{a = 'True} @'False
```

So I'd be against the **instead** approach, I think both are valuable. Especially for something like `generic-lens`:

```haskell
field @"foo" -- would be almost unusable if you'd have to write it as
field @{n = "foo"}
```

Ericson2314:
> both (named and positional), and it's convenient.

But now both position *and* name cannot be changed without breakage. It is a hard-requirement for me that the definition controls the scope of the name, so I can know when they are part of my interface.

I would rather punt and just do record syntax for `DataKinds`. At least here we meet that requirement.

PR #197, opened 2019-01-19
infixr :: (treat :: right-associative-ly)

Icelandjack:
A proposal to treat :: right-associative (even if not an operator)

```haskell
'a' :: Char :: Type :: Type

Fix ::(T -> T) -> T::T
```

instead of

```haskell
'a' :: (Char :: (Type :: Type))

Fix ::((T -> T) -> T::T)
```

### DIFF
--- a/proposals/infixr_colons.rst
+++ b/proposals/infixr_colons.rst
@@ -1,0 +1,107 @@
+Notes on reStructuredText - delete this section before submitting
+==================================================================
+
+The proposals are submitted in reStructuredText format.  To get inline code, enclose text in double backticks, ``like this``.  To get block code, use a double colon and indent by at least one space
+
+::
+
+ like this
+ and
+
+ this too
+
+To get hyperlinks, use backticks, angle brackets, and an underscore `like this <http://www.haskell.org/>`_.
+
+
+``infixr ::`` 
+==============
+
+.. proposal-number:: 
+.. trac-ticket:: 
+.. implemented:: 
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+I apologize for the proposal name ``infixr`` colons, but I don't know what to call this. Code like
+
+::
+
+ 'a' :: Char :: Type
+
+fails with a parse error on ``::``. I suppose I am proposing making `::` act right-associative, despite not being an operator.
+
+Why? Well, I would like to show that ``Type -> Type`` and ``Int`` are both ``Type``, and I think it would look snappier if you could write
+
+::
+
+  Maybe :: Type -> Type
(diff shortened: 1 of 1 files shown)


Icelandjack:
Only `Type`, I could make a case of utility if we had a `Type`-hierarchy like Agda but I like to be able to write stuff like this 

```haskell
data           TyRep (a      ::    k) :: Type where
 TyRepInt   :: TyRep (Int    :: Type) :: Type
 TyRepChar  :: TyRep (Char   :: Type) :: Type
 TyRepFalse :: TyRep ('False :: Bool) :: Type
 TyRepTrue  :: TyRep ('True  :: Bool) :: Type
```

to highlight where `TyRep` fits into the picture

nomeata:
@Icelandjack do you want to pursue this further?

nomeata:
Marking this as dormant for now.

Icelandjack:
> @Icelandjack do you want to pursue this further?

@nomeata: I might actually, I was thinking the following: It's unfortunate that GHCi doesn't parenthesize the output of `:type`

```
>> :t \a -> a
\a -> a :: p -> p
```

Because `\a -> a :: p -> p` is a different expression than the user entered. For something like

```
>> :t not
not :: Bool -> Bool
```

the output is at least a valid expression. But hang on, what happens if I copy-paste that the output isn't valid because `::` is not treated as infixr:

```
>> :t not :: Bool -> Bool
not :: Bool -> Bool :: Bool -> Bool
```

`:t` prints invalid expressions for any expression headed by a type annotation!

This cause me to respond again, I feel like `:t` shouldn't print bogus expressions and I don't think it should print `(not :: Bool -> Bool) :: Bool -> Bool` either

nomeata:
The current behaviour is clearly a bug in GHC, I'd say. But

> I don't think it should print `(not :: Bool -> Bool) :: Bool -> Bool` either

is optinon ;-)

(But certainly worth discussing).

But wait, 
```
not :: Bool -> Bool :: Bool -> Bool
```
only makes sense when `::` is left-associative. So isn’t this example contradicting your rational? You can’t have both
```
not :: Bool -> Bool :: Bool -> Bool
```
and
```
not :: Bool -> Bool :: Type
```
– one of these two will need parentheses!

Icelandjack:
Good catch! Ok the motivation doesn't feel worth it now, I now change my mind and think ghci should add parentheses to `(\a -> a) :: a -> a` and `('a' :: Char) :: Char`. I would be really curious if a good motivation one way or the other exists, `::` usually isn't discussed much so it was good to think about

nomeata:
I agree it should add parentheses, and that sounds like a simple GHC patch to me, not a proposal.

aspiwack:
> I now change my mind and think ghci should add parentheses to `(\a -> a) :: a -> a` and `('a' :: Char) :: Char`.

Interestingly, if I read the Haskell 2010 report, while parentheses are needed for `('a' :: Char) :: Char`, `\a -> a :: a -> a`, ought to parse as `(\a -> a) :: a -> a`:

```
exp → infixexp::[context=>]type | …
infixexp → lexp | …
lexp → \apat1...apatn->exp | …
```

PR #203, opened 2019-01-30
Pointer Rep

andrewthad:
[Rendered link](https://github.com/andrewthad/ghc-proposals/blob/pointer_rep/proposals/0000-pointer-rep.rst)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -228,7 +228,6 @@ Manuel M T Chakravarty  `@mchakravarty <https://github.com/mchakravarty>`_
 Iavor Diatchki          `@yav <https://github.com/yav>`_
 Richard Eisenberg       `@goldfirere <https://github.com/goldfirere>`_
 Ben Gamari              `@bgamari <https://github.com/bgamari>`_
-Roman Leshchinskiy      `@rleshchinskiy <https://github.com/rleshchinskiy>`_
 Simon Marlow            `@simonmar <https://github.com/simonmar>`_            co-chair
 Simon Peyton-Jones      `@simonpj <https://github.com/simonpj>`_              co-chair
 Eric Seidel             `@gridaphobe <https://github.com/gridaphobe>`_
@@ -238,9 +237,10 @@ Members have terms of 3, 4, and 5 years.
 
 We would also like to thank our former members
 
-======================  ====================================================  =========
+======================  ==================================================== 
 Ryan Newton             `@rrnewton <https://github.com/rrnewton>`_
-======================  ====================================================  =========
+Roman Leshchinskiy      `@rleshchinskiy <https://github.com/rleshchinskiy>`_
+======================  ==================================================== 
 
 
 Committee process
--- a/proposals/0000-pointer-rep.rst
+++ b/proposals/0000-pointer-rep.rst
@@ -1,0 +1,168 @@
+Pointer Rep
+==============
+
+.. proposal-number:: Leave blank. This will be filled in when the proposal is
+                     accepted.
+.. trac-ticket:: Leave blank. This will eventually be filled with the Trac
+                 ticket number which will track the progress of the
+                 implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+GHC's ``RuntimeRep`` allows programmers to deal polymorphically with types of a
+common runtime representation. It distinguishes between lifted pointer
+representations, unlifted pointer representations, and various unlifted
+non-pointer reprenestations. There is a shortcoming of the way this
+distinction is currently drawn. There are several primitives types
+(``MutVar#``, ``MVar#``, ``Array#``, ``TVar#``, ``SmallArray#``, etc.)
+that are polymorphic in such a way that they only accept lifted types.
+The runtime actually allows them to accept unlifted pointer types as
+well, but the data constructors of ``RuntimeRep`` are not expressive
+enough to capture this possibility.
+
+Here it is proposed that the definition of ``RuntimeRep`` be changed
+from::
+
+    data RuntimeRep
+      = LiftedRep
+      | UnliftedRep
+      | IntRep
+      | ...
+
+to::
+
+    data Levity = Lifted | Unlifted
+    data RuntimeRep
+      = PtrRep Levity
+      | IntRep
+      | ...
+
+The aforementioned primitives, along with their relevant corresponding
+functions, would accept levity-polymorphic (but not
+representation-polymorphic) types. This proposal obsoletes the
+`Unlifted Array proposal`_, which was accepted but has not yet
+been implemented, since it accomplishes more with even fewer
+primitives.
+
+.. _Unlifted Array proposal: https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0021-unlifted-array.rst
+
+Motivation
+------------
+All of the reasons for which the `Unlifted Array proposal`_ proposal were
+accepted are applicable here as well. Those reasons are not reiterated
+here, so please please read the motivation section of that proposal for
+a discussion of the deficiencies of ``ArrayArray#``.
+
+.. _Unlifted Array proposal: https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0021-unlifted-array.rst
+
+What else do we get? In the discussion of the Unlifted Array proposal,
+David Feuer and Michal Terepeta `expressed interest`_ in a variant of
+``SmalArray#`` that could hold unlifted values. That proposal was
+accepted without the addition, but this proposal gives us the ability 
+to express this. Furthermore, it does this without adding any new
+primitive types. Elsewhere, David has `suggested an UnliftedRef type`_.
+This proposal gives us a way to express that without ``unsafeCoerce#``.
+
+.. _expressed interest: https://github.com/ghc-proposals/ghc-proposals/pull/112#issuecomment-385773161
+.. _suggested an UnliftedRef type: https://github.com/haskell/primitive/issues/198
+
+Where does the need for this arise? We can reduce some needless boxing
+in the event manager if we are able to store unlifted types in ``MutVar#``.
+Consider functions like::
+
+    eventManager :: IORef (IOArray Int (Maybe (ThreadId, EventManager)))
+
+If we could store the unlifted ``Array#`` type directly in a ``MutVar#``,
+we could remove the wrapper that ``IOArray`` imposes. (Actually, this
+example is more complicated. It is also necessary to replace the use of
+``IOArray`` with an array type that doesn't store the lower and upper
+bounds.)
+
+Proposed Change Specification
+-----------------------------
+The ``RuntimeRep`` data type is redefined as::
+
+    data Levity = Lifted | Unlifted
+    data RuntimeRep
+      = PtrRep Levity
+      | IntRep
+      | ...
+
+The following primitive types are given new kinds::
+
+    data Array# :: forall (v :: Levity). TYPE ('PtrRep v) -> Type
+    data MutableArray# :: forall (v :: Levity). Type -> TYPE ('PtrRep v) -> Type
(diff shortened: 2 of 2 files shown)


rrnewton:
This seems like the rare proposal that doesn't add complexity and yet cleans up existing problems and results in stronger types for primops.  If there's a downside, it has eluded me.  Looks great!

(Personally, I'm also excited about the ability to put unlifted pointer types in IORefs and MVars -- slowly, slowly, we are getting that efficient imperative/strict language hiding inside Haskell 😉.)

andrewthad:
Thanks! Also, if anyone's got any other good examples of where it's useful to put something unlifted inside an `IORef` or a `MVar`, I'd be happy to include them in the proposal. I want this from time-to-time in my own code, but I had a hard time finding any good "in the wild" examples of where this would be benefits.

andrewthad:
cc @treeowl I suspect you and interested in this, and I would appreciate any feedback you have.

chessai:
Strong +1 from me

simonpj:
> This proposal does not loosen any of the restrictions around where representation polymorphism can be used. 

I'd like this to be more explicit.  Specifically, the current restriction on any lambda-bound variable is that it may not have a type whose kind is `TYPE r` where `r` is a variable; it must have `TYPE IntRep` or `TYPE LiftedRep`, with no variables.

Here one might wonder: can you have a term `/\(l :: Levity) (a::TYPE (PtrRep l)). \(x :: a). blah`?  You might think "yes" because the representation of `x` is a pointer.  But we can't tell whether or not it's a _lifted_ pointer, and that's vital for the correct strictness semantics.

So the rule for levity polymorphism becomes that a lambda-bound variable must have a type whose kind is `TYPE rtr`, where `rtr` is a runtime-rep that has no variables (`RuntimeRep` variables or `Levity` variables).

Could make that 100% explicit.  Feel free to steal language from the levity-polymorphism paper.

andrewthad:
@simonpj I've modified that section to be more explicit. Let me know if it now has everything you are looking for.

goldfirere:
When we designed levity polymorphism, we wondered if the proposed design here was the right way to go. I think we hadn't identified any use cases for it. Now that you have, I think this is a good direction.

cartazio:
I like the flavor of this proposal ..  but why can’t unlifted and lifted reps be type or pattern synonyms under the proposal ? 

Also ... what type do I give instead of LiftedRep and unliftedRep

Ptr Lifted and Ptr unlifted?  

On the face of it this looks to be a good design though.  

It would be good to access what / if any knock on breakage it might trigger though.  Even though most code shouldn’t be impacted by this, it’s a very core / fundamental cleanup and it’s worth understanding impact.

andrewthad:
I’ve  updated the proposal with an explanation of why using type synonyms for LiftedRep and UnliftedRep doesn’t work out totally well.

treeowl:
I would hope you'd *generalize* weak references along the other axis, allowing keys to be associated with *values* of different levities.

```haskell
mkWeak# :: forall {vu :: Levity}  {vb :: Levity} (u :: TYPE ('PtrRep vu)) (b :: TYPE ('PtrRep vb)) (c :: Type).
  u -> b -> (State# RealWorld -> (#State# RealWorld, c#)) -> State# RealWorld -> (#State# RealWorld, Weak# b#)
mkWeakNoFinalizer# :: forall {vu :: Levity} {vb :: Levity} (u :: TYPE ('PtrRep vu)) (b :: TYPE ('PtrRep vb)).
  u -> b -> State# RealWorld -> (#State# RealWorld, Weak# b#)
```

Side note: `with#` can *already* be generalized along another axis, so you might as well make it

```haskell
with# :: forall {vu :: Levity} {vr :: Levity} (u :: TYPE ('PtrRep vu)) (s :: Type) (r :: TYPE ('PtrRep vr)).
  u -> (State# s -> (# State s, r #)) -> State# s -> (# State# s, r #)
```

treeowl:
Note to email followers: my last comment has been edited.

treeowl:
Edited twice, now.

andrewthad:
I was hoping it was possible to generalize `Weak` further as you've suggested. That's good to know. And I agree that `with#` should be fully representation-polymorphic in the return value. Does anyone know if `StableName#` and `StablePtr#` are able to support unlifted types? I suspect that they are, but I'd like something more familiar with their implementations to confirm this.

treeowl:
@andrewthad, yes, they certainly can! The fussy bits of stable names and pointers have to do with storage management; they really don't care about lifted vs. unlifted. That reminds me that I should turn back to my interrupted project to replace both of those systems, but that's entirely orthogonal.

andrewthad:
For sure, the  mutable constructor fields  proposal blows up MutVar.   Most high-performance code using MutVar  would be better suited by mutable constructor fields. But all of the other types generalized by this proposal  are still just as useful  in a post mutable constructor fields world.

It’s good to know that about stable names and stable pointers.  I will update the proposal to include their generalization as well.

andrewthad:
I've incorporated @treeowl's changes. My only remaining question is about painting the bikeshed. What's a better name for this: `PtrRep`, `HeapRep`, `ManagedRep`? The weird thing about `PtrRep` is that we have something else called `AddrRep`. The latter is for memory that may not be managed by GHC's runtime, and the former is for things that are definitely managed by GHC's runtime. I don't feel strongly on this, and I'll probably just punt this decision to the committee as part of the proposal, but if anyone has any thoughts, they can certainly be taken into consideration.

simonpj:
I like `PtrRep`.

iamrecursion:
I’ve nothing against `PtrRep`, but the potential for confusion with `AddrRep` is a good point. If I had to pick another name, I’d probably go with `ManagedRep`. It’s more explicit than `HeapRep` to me.

cartazio:
No matter the name, it’s gonna be something that you need to follow the
docs on

iamrecursion:
Very true, which is why I don’t feel like the exact choice of name is a big problem.

andrewthad:
We've now had a chance to have the bikeshedding discussion, and it appears that there are no more substantive issues (everyone seems to be on board with the question of "is this actually sound?"). I'd like to submit this proposal to the committee, requesting as well that thte committee decide the name of the new data constructor.

treeowl:
@nomeata, that means you.

nomeata:
Technically, I usually prefer to wait for at least about a week of silence before going to the next state. But given that this is positively received, and mostly a low-level change, I’m find with finding a shepherd directly.

cartazio:
I do think it would be good to experimentally identify if there’s
unexpected breakage from this change , though it eeems fine

andrewthad:
I'm documenting this here so that I don't forget it, but oddly `readTVarIO` doesn't generalize like all of the other `TVar` operations do. The problem is that it is implemented in cmm as:

```
stg_readTVarIOzh ( P_ tvar /* :: TVar a */ )
{
    W_ result;

again:
    result = StgTVar_current_value(tvar);
    if (%INFO_PTR(result) == stg_TREC_HEADER_info) {
        goto again;
    }
    return (result);
}
```

Since it expects an info table pointer, this only works with lifted types.

andrewthad:
Actually, there's some conditional code for `readTVar` as well that does some `UNTAG_CLOSURE` and `GET_INFO`. This may be the real problem. I think that supporting unlifted pointers in `TVar`s is beyond my ability, so I may need to retract that part of the proposal.

nomeata:
Accepted as Proposal 45.

simonpj:
> oddly readTVarIO doesn't generalize like all of the other TVar operations do. 

I'm no expert here, but this smells odd to me.  I'd like us to understand why it doesn't generalise, rather than just accept that (on the surface) it doens't.   Maybe there are some GHC-STM experts out there.  See who committed the code you point to.  Etc.

simonmar:
I don't see any problem here, @andrewthad what is it you're concerned about with `readTVar` and `readTVarIO`?

andrewthad:
Sorry, I'd been playing around with `TVar`s with boxed unlifted payloads in [primitive-unlifted](https://github.com/andrewthad/primitive-unlifted), and I accidentally used `readTVar#` where I should have used `readTVarIO#`. I've fixed this, and as everyone expected, there is not actually a problem with generalizing this primitive.

sgraf812:
I'm a little late to the game, but here are my comments:

> What's a better name for this: PtrRep, HeapRep, ManagedRep? The weird thing about PtrRep is that we have something else called AddrRep.

Yes, *very* much! In fact, I got hung up about this in the past, too: See the 3 comments starting at https://gitlab.haskell.org/ghc/ghc/issues/9279#note_159956.

@goldfirere already made precedent in his levity polymorphism work and wrote a function called [`isGcPtrRep`](https://gitlab.haskell.org/ghc/ghc/blob/b1613985252990600a2c7d789054c74470260959/compiler/types/TyCon.hs#L1419). So why not call it `GcPtrRep`? It's shorter than `ManagedRep`, which I'd prefer otherwise. Managed/Unmanaged are terms I'm very familiar with, coming from a C# background ([example](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.getmanagedthunkforunmanagedmethodptr?view=netframework-4.8#System_Runtime_InteropServices_Marshal_GetManagedThunkForUnmanagedMethodPtr_System_IntPtr_System_IntPtr_System_Int32_)). Anything, but please don't just call it `PtrRep`.

Nit: In the proposal you write

> It is possible to loosen the binder restriction, but this proposal does include such a change since it would make implementation more difficult.

But I think you specifically mean that this proposal does *not* include such a change.

Is there an MR somewhere?

andrewthad:
Unfortunately, I don't have an MR for this. A few months ago, I spent an hour or two working through all the weird `RuntimeRep`-like types that give it its wired-in magic, and I've not picked it back up since then. I don't think implementing this should be hard. It just requires having a block of maybe ten or fifteen hours to sit down and just grind through it. I've not been able to find that time. I hope to find the time at some point over the holidays, but I may not, so if anyone else wants to give this a try, go for it.

Concerning the naming, I don't know what the procedure is. I defer entirely to the committee on this issue. The committee approved the name `PtrRep`, and I don't what the procedure is for changing a proposal that has already been approved. (Perhaps you could open a PR that modifies the existing proposal?) For what it's worth, I don't particularly like `GcPtrRep`. I find it strange to mention garbage collection (an implementation detail) in the name of a type or kind. Come to think of it, I cannot believe I didn't consider the name `BoxedRep` earlier. This name actually fits in rather nicely with the existing literature around GHC. For example, this name scheme can be seen in the [Levity Polymorphism](https://www.microsoft.com/en-us/research/publication/levity-polymorphism/) paper, where Figure 1 is a four-quadrant table that illustrates the boxed vs unboxed and lifted vs unlifted distinctions.

andrewthad:
The committee has included a procedure for amending a proposal in https://github.com/ghc-proposals/ghc-proposals/pull/299. @sgraf812 Let me know what you think of `BoxedRep`. I'm inclined to amend this.

sgraf812:
I think `BoxedRep` is better, but I tend to associate "pointer" with "box". Which is probably wrong when you don't know what you point to. What's wrong about `ManagedPtr`? If that's a no go, I'd be fine with `BoxedRep`.

andrewthad:
`ManagedPtr` (which would actually need to be `ManagedRep` or `ManagedPtrRep`) is alright, but I like `BoxedRep` more because it's (1) shorter and (2) much more consistent with the language that GHC uses. The word "boxed" is already used on the GHC wiki, in published papers, and all throughout the GHC source code. The word "managed" is entirely absent from the Levity Polymorphism paper, doesn't show up in the wiki's [discussion of unboxed types](https://wiki.haskell.org/Unboxed_type), and shows up less frequently in the GHC source.

sgraf812:
Alright, `BoxedRep` it is!

sgraf812:
While implementing the unlifted datatypes proposal, I realised that this proposal leaves open the question of whether to default inferred `Levity` type variables to `Lifted`.

Should we or shouldn't we default `Levity`? Or is that a question that is out of scope for this proposal, because there is no language construct (to my knowledge) that allows us to observe defaulting?

simonpj:
Good point.   I think we should default, else we risk getting unexpectedly complicated types.  (Same with linearity I think.)

Examples would help.

sgraf812:
In https://gitlab.haskell.org/ghc/ghc/-/merge_requests/2218#note_336453, I argue that we should not default `Levity`.

nomeata:
According to https://well-typed.com/blog/2022/02/ghc-2021-12-2022-01/ this is completely implemented now. Can someone confirm that it is complete, and which GHC will contain these changes?

PR #204, opened 2019-02-03
Remove nested bracket restriction in Template Haskell

mpickering:


goldfirere:
This seems like it holds water. Why do we want the extra complexity? What (practical thing) does it allow you to do? (Preferably, answer in the proposal itself.)

Ericson2314:
I have an old WIP proposal for proper phase separation for TH (which would support cross compilation soundly). I better finish that up, because I think we should fix phase separation before we freely generate phases!

mpickering:
Here's an example of a program which I had to rewrite slightly because of the nested bracket restriction. See `codePlus`.

http://mpickering.github.io/posts/2019-02-14-stage-3.html

mpickering:
Is there anything else blocking this proposal?

nomeata:
Probably not. (Or the proposal hasn’t seen enough critical eyes, but not your fault then, I guess). Do you want to submit?

Ericson2314:
I take back my concerns. The only issue is cross stage persistence, which involves `lift` and thus is pretty well cordoned off from everything else. With cross-compilation-friendly TH, the restrictions on lift should imply the rest.

+1 on this. Nice work.

mpickering:
@nomeata Please submit this and my two other proposals.

yav:
@mpickering  I am a bit confused about the section with the multiple lifts:

```
foo2 x = [| [| $($(lift (lift x))) |] |]
```
Doesn't the outer call to `lift` require a `Lift` instance for `Q a` (where `a` is the type of `x`)?   And I don't think we have such an instance.

mpickering:
@yav Yes you need that instance. I [implemented it here](https://gitlab.haskell.org/ghc/ghc/merge_requests/259/diffs#ed3c6537b674490396f9cab5cbb014ee54e71168_709_710)

yav:
Hm, is that instance correct though?   It looks like it is doing the effects in the wrong stage, and as a result they will happen only once.   This can be bad, for example, if the effect is to generate new names.

mpickering:
What example do you have in mind?

yav:
Here's an example (omitting imports,language extensions, etc.)
```
module A where
  hello :: Q ()
  hello = liftIO (putStrLn "Hello")

  ex1 :: Q Exp {- Exp is of type Q ()-}
  ex1 = lift hello

module B where
  import A

  ex2 :: Q ()
  ex2 = $ex1

  ex3 :: Q Exp
  ex3 = ex2 >> ex2 >> [| () |]
```

I would expect that `lift m` will make a representation of `m` and splicing it back would get
the original monadic computation (but not execute it!).  Only once you splice *that* should you see
the effects.  In contrast, with this implementation, the side effects happen *once* when you load `B` (i.e., you see `"hello"`), but if you splice `ex3` you see no side effects.

This would be quite wrong if the side effects were, for example, about generating fresh names,
because instead of `ex2` generating a new name every time it is run, we would just generate one
name, and keep returning it.

If `Q` was some sort of free monad, perhaps we could write a proper instance for `Lift`, but as it is I don't know how to do it.

mpickering:
Ok @yav these are good points. I didn't consider the effects in `Q` as I never use them. It would be very unfortunate if the unprincipled metaprogramming of untyped template haskell got in the way of the typed variant so we'll have to think of a way around this. 

I will need to think more about how this instance interacts with the `newName` effect as it isn't obvious to me. 

It seems the best way to me would be to shelve this proposal and work on a revised form of #195 and #211 first so that the bracket form is more flexible and no tied to `Q`. This would have some quite far reaching consequences though as it would affect the type of `lift` (to not tie it to Q) but it seems better in the long run. `lift` and brackets are closely related as `Lift` instances are implemented using quotation.

yav:
Ok, let's wait---I don't think there is any hurry to do this, and we should figure out how to get it right.

yav:
@mpickering One idea to explore would be to add a version of `lift` that lifts across multiple stages.  This should work just fine as you only ever lift values, and there shouldn't be any need to lift actual `Q` computations.

Ericson2314:
When I saw that line I was surprised: I sort of intuitively thought it would be something more like `$(lift $(lift $(lift ...)))`, but then shrugged and assumed splicing and lifting commuted. But doesn't doing it in this way solve the problem by not needing those `Lift (Q a)`  instances?

yav:
I imaging we'd also need to change something in the TH representation of brackets too:  in the current version of the proposal splices are implemented by just pure `Exp`s in the environment of a bracket.  However, since in general we can splice arbitrary `Q` computations maybe we'd need to store `Q` things there?

Also, TH supports splicing different things, not just expressions, so I imagine more changes are needed to make this work (i.e., bracket representation in other types, and a more complex environment).

mpickering:
> When I saw that line I was surprised: I sort of intuitively thought it would be something more like `[| [| [| ... $(lift $(lift $(lift ...))) ... |] |] |]`, but then shrugged and assumed splicing and lifting commuted. But doesn't doing it in this way solve the problem by not needing those `Lift (Q a)` instances?

This is a good point @Ericson2314 and does remove the need for lifting `Q` actions.

Ericson2314:
BTW I was looking at http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.80.9887&rep=rep1&type=pdf on @xplat's recommendation, and it happened to discuss just this issue of the non-communivity of quoting and lifting.

nomeata:
This is marked as needing revision. Is someone working on this?

mpickering:
Yes.

mpickering:
A realistic example from the wild:

https://chrisdone.com/posts/static-smart-constructors/

PR #205, opened 2019-02-03
Structured module exports/imports

deepfire:
[Rendered](https://github.com/deepfire/ghc-proposals/blob/master/proposals/0000-structured-imports.rst)

### DIFF
--- a/proposals/0000-structured-imports.rst
+++ b/proposals/0000-structured-imports.rst
@@ -82,6 +82,16 @@ This constitutes a re-interpretation of module re-export statements, that
 essentially allows qualified names to flow between modules in a transparent manner,
 the same way regular names are allowed to.
 
+Module self re-export: unchanged
+''''''''''''''''''''''''''''''''
+Considering module self-re-exports::
+   module M (module M) where ...
+
+Such self re-export does not add any qualified names to the export list.
+
+The rationale is that the local qualified namespace is often quite populous,
+and so reinterpretation to include it would hinder the established usage.
+
 Export entry with a 'qualified' keyword: qualified names only
 '''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
 In section 5.2, "Export lists", *Haskell2010*, with regards to the the list of cases


deepfire:
The basic idea is to take the existing semantics for the _level-0_ names (modulo its relationship with the current aliases), and to re-use it, largely intact for the contents of _level-1_ name sets.

I agree that the presentation is fairly obscure at this point, it has suffered from too many incremental edits.. Let me take it further.

And thank you for the references!

deepfire:
Thank you for the references!

And yes, I agree, the presentation needs to be unpacked and generally made more clear.

The basic idea _should_ be simple, I think, which is to take existing semantics for the flow of _level-0_ names, and re-use it mostly unchanged at _level-1_, and indeed all that needs to be made clear.

deepfire:
@simonmar, @int-index, reworked the proposal to add missing precision and detail -- but also, sadly, significant length.

deepfire:
I see that I have utterly misread the meaning of `ModIface`'s `mi_exports` field, so adjustments to follow.

deepfire:
After an overview of things involved, I think I have a reasonable chance of being able to implement this.

simonmar:
I'd like to give more useful feedback on this proposal, but I'm afraid the main feedback I have is this: it's really hard to follow the proposal! The proposal is very verbose, mixes implementation details with motivation and general commentary, and uses unconventional terminology ("level-1 names"? what's wrong with "qualified names"?). 

The proposal seems to have a complete re-specification of the Haskell 2010 module system, but in an imprecise way and with different terminology. That makes it really hard for someone to assess the proposal. Could it instead be expressed as a delta to the existing specification? After all, the existing spec has already gone through multiple rounds of review and all the dark corners (of which there were many) have been worked out.

I'd like to point out one really crucial detail that the proposal seems to miss: module names in Haskell do not have any internal structure. `A.B.C` is just a module name - the dot is simply a character that is allowed inside a module name. Any proposal for a change to the module system should not treat module names with dots any differently from module names without dots - unless we want to deliberately break this simplifying assumption, in which case we would need a very good reason for doing so.

deepfire:
First, Simon, thank you for the feedback, it is very useful!  Indeed, in an attempt to arrive at a simplified mental picture of the problem I fell into the trap of reinventing too much of the terminology.  I'll work on reducing the unnecessary delta.

Second, I've just completed a proof-of-concept implementation, and I think I've learned enough in the process to improve the proposal.

I'll be working on improving the implementation to the point where I'm comfortable that I don't violate any invariants -- and also thinking about new tests to define boundaries of the new grounds.

After that I'll switch to reworking the proposal.

deepfire:
> module names in Haskell do not have any internal structure. A.B.C is just a module name
While I have had this in mind -- the relevant passage of the spec is cited in https://github.com/deepfire/ghc-proposals/blob/8696df477a598452786d4fe8cc05c126d2877d4a/proposals/0000-structured-imports.rst#terminology:
> This is purely a convention, however, and not part of the language definition; in this report a modid is treated as a single identifier occupying a flat namespace.

..I did not realise that this rule was without exceptions, and that `import ... as X.Y.Z` is allowed.

This, indeed, allows a dramatic simplification of the proposal.  Great news -- and thank you a lot!

deepfire:
Updated the proposal.

deepfire:
Proposal changes:

1. No new terminology.
2. Drastically reduced size.
3. Module-name-dot-agnosticism.

int-index:
How does this proposal compare to qualified re-exports? Today we can say this:

```
module M (module A, module B) where

import Stuff as A
import OtherStuff as B
```

A client of `M` also gets all of definitions in `A` and `B`:

We could extend this to

```
module M (module qualified A, module qualified B) where

import Stuff as A
import OtherStuff as B
```

Now the client of `M` also gets all of definitions in `A` and `B` qualified.

This strikes me as a much simpler alternative.

deepfire:
@int-index, the differences I see between the proposal and what you suggest are:

1. Proposal deals with sets of aliases, you suggest dealing point-wise.
2. Minor syntax details, that mostly follow from 1.

So I wouldn't say there's a drastic difference in complexity.

int-index:
> Proposal deals with sets of aliases, you suggest dealing point-wise.

Yes, this batch handling of aliases is what strikes me as unnecessary complexity. I don't see the motivation behind these constructs:

```
module C
  ( aliases                       -- Complete export of all aliases available in the current module
  , aliases (ALIAS..)             -- Export of the available alias set intersected with the specified list
  , aliases hiding (ALIAS..)      -- Export of the available alias set, modulo the specified list
  )
```

and

```
module B where
import C aliases                  -- Complete import of all aliases available in the imported module
import C aliases (ALIAS..)        -- Import of a subset of aliases available in the imported module
import C aliases hiding (ALIAS..) -- ...
```

Why can't we re-export qualified modules like any other exported entities?

deepfire:
> Why can't we re-export qualified modules like any other exported entities?

It's curious, actually. The thing is that the case we have:

> module M (module M) where

allows for transparent re-export of M's name content, without mentioning its internal structure.

The `aliases` export option that I suggest, allows the same, but for aliases -- i.e. it's actually "like any other exported entities".

Generally, the idea is to provide an option to forward names, without explicitly listing them -- thereby reducing the namespace maintenance costs.

deepfire:
> > Proposal deals with sets of aliases, you suggest dealing point-wise.
> 
> Yes, this batch handling of aliases is what strikes me as unnecessary complexity. I don't see the motivation behind these constructs:
> 
> ```
> module C
>   ( aliases                       -- Complete export of all aliases available in the current module
>   , aliases (ALIAS..)             -- Export of the available alias set intersected with the specified list
>   , aliases hiding (ALIAS..)      -- Export of the available alias set, modulo the specified list
>   )
> ```
> and
> 
> ```
> module B where
> import C aliases                  -- Complete import of all aliases available in the imported module
> import C aliases (ALIAS..)        -- Import of a subset of aliases available in the imported module
> import C aliases hiding (ALIAS..) -- ...
> ```
> Why can't we re-export qualified modules like any other exported entities?

I've deleted my previous answer, and will try to answer your question more directly.
A major theme of the proposal is reduction of module abstraction costs.

For unqualified names, we allow both set-operation-based and pointwise imports and exports -- and so in this context, not prohibiting set-wise import/export-ability for aliases seems to me exactly "like any other exported entity".

Therefore, since there is a direct contradiction between different interpretations of what you suggest, may I ask you to clarify what you intend to suggest in this?

> Why can't we re-export qualified modules like any other exported entities?

deepfire:
> > Proposal deals with sets of aliases, you suggest dealing point-wise.
> 
> Yes, this batch handling of aliases is what strikes me as unnecessary complexity. I don't see the motivation behind these constructs:
> 
> ```
> module C
>   ( aliases                       -- Complete export of all aliases available in the current module
>   , aliases (ALIAS..)             -- Export of the available alias set intersected with the specified list
>   , aliases hiding (ALIAS..)      -- Export of the available alias set, modulo the specified list
>   )
> ```
> and
> 
> ```
> module B where
> import C aliases                  -- Complete import of all aliases available in the imported module
> import C aliases (ALIAS..)        -- Import of a subset of aliases available in the imported module
> import C aliases hiding (ALIAS..) -- ...
> ```
> Why can't we re-export qualified modules like any other exported entities?

I've deleted my previous answer, and will try to answer your question more directly.
A major theme of the proposal is reduction of module abstraction costs.

For unqualified names, we allow both set-operation-based and pointwise imports and exports -- and so in this context, not prohibiting set-wise import/export-ability for aliases seems to me exactly "like any other exported entity".

Therefore, since there is a direct contradiction between different interpretations of what you suggest, may I ask you to clarify what you intend to suggest in this, @int-index?

> Why can't we re-export qualified modules like any other exported entities?

int-index:
We already can re-export all entities of some namespace like this:

```
module Foo (module MonadStuff) where

import Control.Monad.Reader as MonadStuff
import Control.Monad.State as MonadStuff
import Control.Monad.Except as MonadStuff
```

This means that anyone who writes `import Foo`, gets `MonadReader`, `MonadState`, and `MonadExcept` in scope. Anyone who writes `import qualified Foo` gets `Foo.MonadReader`, `Foo.MonadState`, and `Foo.MonadExcept` in scope.

I suggest the following design. Allow qualified re-exports:

```
module Foo (module MonadStuff) where

import Control.Monad.Reader as MonadStuff
import Control.Monad.State as MonadStuff
import Control.Monad.Except as MonadStuff
```

Now anyone who writes `import Foo`, gets `MonadStuff.MonadReader`, `MonadStuff.MonadState`, and `MonadStuff.MonadExcept` in scope. Anyone who writes `import qualified Foo` gets `Foo.MonadStuff.MonadReader`, `Foo.MonadStuff.MonadState`, and `Foo.MonadStuff.MonadExcept` in scope.

This makes the `MonadStuff` alias a first-class entity that can be exported like any other. This design doesn't require new keywords, and the import syntax is not changed at all. I fail to see what use case it wouldn't address.

nomeata:
@deepfire  in light of the comments here, do you want to further pursue this proposal?

deepfire:
@nomeata, absolutely.  More energy required, as always..

deepfire-pusher:
> We already can re-export all entities of some namespace like this:
> ```
> module Foo (module MonadStuff) where
> import Control.Monad.Reader as MonadStuff
> import Control.Monad.State as MonadStuff
> import Control.Monad.Except as MonadStuff
>```
>
> This means that anyone who writes `import Foo`, gets `MonadReader`, `MonadState`, and `MonadExcept` in scope.

@int-index, my first reaction was that of disbelief.  And indeed, it doesn't work:
```
[nix-shell:~/experiment]$ cat Foo.hs
module Foo (module MonadStuff) where

import Control.Monad.Reader as MonadStuff
import Control.Monad.State as MonadStuff
import Control.Monad.Except as MonadStuff

[nix-shell:~/experiment]$ cat Bar.hs
module Bar where

import Foo

x = MonadStuff.ask

[nix-shell:~/experiment]$ ghc Bar.hs
[2 of 2] Compiling Bar              ( Bar.hs, Bar.o )

Bar.hs:5:5: error:
    Not in scope: ‘MonadStuff.ask’
    No module named ‘MonadStuff’ is imported.
  |
5 | x = MonadStuff.ask
  |     ^^^^^^^^^^^^^^
```

deepfire-pusher:
> We already can re-export all entities of some namespace like this:
> ```
> module Foo (module MonadStuff) where
> import Control.Monad.Reader as MonadStuff
> import Control.Monad.State as MonadStuff
> import Control.Monad.Except as MonadStuff
>```
>
> This means that anyone who writes `import Foo`, gets `MonadReader`, `MonadState`, and `MonadExcept` in scope.

@int-index, my first reaction was that of disbelief.  And indeed, it doesn't work:
```
[nix-shell:~/experiment]$ cat Foo.hs
module Foo (module MonadStuff) where

import Control.Monad.Reader as MonadStuff
import Control.Monad.State as MonadStuff
import Control.Monad.Except as MonadStuff

[nix-shell:~/experiment]$ cat Bar.hs
module Bar where

import Foo

x = MonadStuff.ask

[nix-shell:~/experiment]$ ghc Bar.hs
[2 of 2] Compiling Bar              ( Bar.hs, Bar.o )

Bar.hs:5:5: error:
    Not in scope: ‘MonadStuff.ask’
    No module named ‘MonadStuff’ is imported.
  |
5 | x = MonadStuff.ask
  |     ^^^^^^^^^^^^^^
```

How could it work, if the structure used to carry exported names in module interface files is a flat list?

int-index:
> How could it work, if the structure used to carry exported names in module interface files is a flat list?

Because currently, you don't write `MonadStuff.ask`, you just write `ask`. This is my proposal exactly: export `MonadStuff` as a standalone entity by prefixing it with the `qualified` keyword. _Then_ your example should work.

int-index:
In other words, this is accepted today:

```
Reexport$ cat Foo.hs 
module Foo (module MonadStuff) where

import Control.Monad.Reader as MonadStuff
import Control.Monad.State as MonadStuff
import Control.Monad.Except as MonadStuff
Reexport$ cat Bar.hs 
module Bar where

import Foo

x1 :: MonadReader r m => m r
x1 = ask

x2 :: Foo.MonadReader r m => m r
x2 = Foo.ask
Reexport$ ghci Foo.hs Bar.hs 
GHCi, version 8.4.4: http://www.haskell.org/ghc/  :? for help
[1 of 2] Compiling Foo              ( Foo.hs, interpreted )
[2 of 2] Compiling Bar              ( Bar.hs, interpreted )
Ok, two modules loaded.
```

I propose we accept this:

```
Reexport$ cat Foo.hs 
module Foo (module MonadStuff qualified) where

import Control.Monad.Reader as MonadStuff
import Control.Monad.State as MonadStuff
import Control.Monad.Except as MonadStuff
Reexport$ cat Bar.hs 
module Bar where

import Foo

x1 :: MonadStuff.MonadReader r m => m r
x1 = MonadStuff.ask

x2 :: Foo.MonadStuff.MonadReader r m => m r
x2 = Foo.MonadStuff.ask
```

Does this make sense? @deepfire

deepfire:
@int-index, so you are proposing we add an entirely new meaning to the `qualified` keyword.  Don't we have a potential for further confusion of language users?

To cite the initial version of the proposal:

>    It's worth underscoring the effect of the ``qualified`` keyword in the non-extended language -- it is strictly negative, as it suppresses population of the top level of the local namespace. Quoting *Section 5.3.2 of Haskell2010*:
>
>```
>      If the import declaration used the *qualified* keyword, only the qualified name of the entity is brought into scope. If the *qualified* keyword is omitted, then both the qualified and unqualified name of the entity is brought into scope.
>```
>
>    With this understanding, one could (hypothetically) say that:
>
>    1. it's a misnomer, and should be called ``qualified-only`` instead, and
>    2. a more consistent name for the ``as`` keyword could be ``qualified-as``.

If I read you correctly, you're suggesting we use `qualified` in a positive meaning (include the aliases) -- whereas the analysis above makes it clear the non-extended language uses it in negative sense (exclude unqualified names).

int-index:
`qualified` has the meaning "qualify the names at use sites" in both cases. Therefore:

1. `module Foo (module MonadStuff) where` means "re-export everything from `MonadStuff` without additional qualification"
2. `module Foo (module MonadStuff qualified) where` means "re-export everything from `MonadStuff` with the `MonadStuff.` qualifier"

(1) is already the case today, I propose we add (2).

deepfire:
> `qualified` has the meaning "qualify the names at use sites" in both cases.

I respectfully disagree. Again, citing quotes Section 5.3.2 of Haskell2010 -- compare:



> If the import declaration used the *qualified* keyword, only the qualified name of the entity is brought into scope.

vs.

> If the *qualified* keyword is omitted, then both the qualified and unqualified name of the entity is brought into scope.

Clearly, `qualified` means "qualified only".  I'm not sure how it can be read otherwise.

int-index:
> So, all in all:
> Pro-aliases:
> Less user confusion.

I continue to be confused by the current proposal that talks about "aliases". You cannot list this as a "pro".

> Clearly, qualified means "qualified only". I'm not sure how it can be read otherwise.

This is exactly the meaning I am proposing. I'm not sure how I can make it more clear.

deepfire:
@int-index, I see now -- I somehow misread your proposal twice.  Thank you for your patience.

deepfire:
@int-index, I see now -- I somehow misread your proposal twice.  Thank you for your patience.

So we agree on the interpretation of `qualified`.  That's great!

deepfire:
So, I have one last problem with your suggestion, @int-index -- loss of means of control over re-exported aliases.

This should be viewed in light of the original motivation of the proposal -- reduction in the costs of module abstraction.

You seem to propose that every module that intends to export (or, potentially, re-export) a set of aliases, has to mention every single alias in the export list -- `module qualified Ali.As`.

What is your estimate of this cost on re-exporting modules?

How do you estimate this compares to re-exporting of plain names?

How do you thing that impacts the idea of  having a single prelude-like module defining a structured namespace, and having that namespace re-exported through other modules?

My personal assessment is that the effect would be highly disruptive, because of all of that pointwise mentioning of every alias at every re-export point -- but I'd like to hear your take.

--
As an aside, you said:

> I continue to be confused by the current proposal that talks about "aliases". You cannot list this as a "pro".

What, in particular, do you find confusing in the proposal, as currently stated?

int-index:
> You seem to propose that every module that intends to export (or, potentially, re-export) a set of aliases, has to mention every single alias in the export list

Not quite. This is true for an export, but not true for a re-export. For example, here's a module `Containers` that exports 4 type names (Map, Set, HashMap, and HashSet) and 4 aliases (Map, Set, HashMap, HashSet).

```
module Containers
  ( Map,
    Set,
    HashMap,
    HashSet,
    module Map qualified,
    module Set qualified,
    module HashMap qualified,
    module HashSet qualified,
  ) where

import Data.Map as Map
import Data.Set as Set
import Data.HashMap.Strict as HashMap
import Data.HashSet as HashSet
```

Indeed, this is more wordy than your version:

```
module Containers
  ( Map,
    Set,
    HashMap,
    HashSet,
    aliases
  ) where

import Data.Map as Map
import Data.Set as Set
import Data.HashMap.Strict as HashMap
import Data.HashSet as HashSet
```

However, it is more consistent with what we have today. After all, we don't write this:

```
module Containers
  ( typenames,
    aliases
  ) where

import Data.Map as Map
import Data.Set as Set
import Data.HashMap.Strict as HashMap
import Data.HashSet as HashSet
```

And I don't see why you give aliases and type names different treatment.

Now, in re-exports, I don't expect different treatment either. For example,

```
module CustomPrelude
  ( module Containers,
    module Control.Monad,
  ) where

import Containers
import Control.Monad
```

Here, `CustomPrelude` re-exports _everything_ from `Containers`, its 4 type names and its 4 aliases. Just like it re-exports everything from `Control.Monad` – its types, functions, data constructors, etc.

----

> What is your estimate of this cost on modules that do re-exports?

No cost.

> How do you estimate this compares to re-exporting of plain names?

We give aliases equal treatment (as described above).

> How do you think that impacts the idea of having a single prelude-like module defining a structured namespace, and having that namespace re-exported through other modules?

I believe this use case is addressed adequately.

int-index:
> What, in particular, do you find confusing in the proposal, as currently stated?

It proposes a completely new set of rules for handling aliases, and new syntax for these rules, instead of leveraging existing intuitions. My mental model of a module is that it's a set of entities (type constructors, data constructors, functions, pattern synonyms, etc). You can import this set of names unqualified, qualified, unqualified with an alias, and qualified with an alias:

1. `import Mod`
2. `import qualified Mod`
3. `import Mod as Alias`
4. `import qualified Mod as Alias`

My proposal is to add a new type of entity that can be exported and re-exported: a module alias. Then, at import site _all the same rules apply_. It's straightforward.

Your proposal contains new syntax and new rules to remember at import site. It is difficult and unintuitive for users of modules that contain aliases.

For example, you seem to propose that users have to write this:

```
import CustomPrelude
import CustomPrelude aliases
```

Why is it that I need to import aliases separately? With my proposal, the users simply write

```
import CustomPrelude
```

and get all the aliases they need.

Ericson2314:
@int-index what happens if you export qualified a module that has its own qualified exports?

-----------

I ask this question because right now my intuition for modules is modules can contain other modules (those qualified names), but those modules can't further contain qualified stuff. And confusingly, hierarchical module names been that both of the two layers can contain many `.`-separated components.

I'm pretty sure @int-index's proposal, and maybe also this proposal as written, both lift that restriction on 2 layers and allow free generating arbitrarily deep layers. I think this *semantic* change is the interesting place to begin, and we work through all those semantic interactions before debating syntax.

-----------

Also, I recommend looking at Rust's module system, which changed in the "2018 edition" from the 2015 original. Not because there are hugely innovative ideas there, but because following someone else's bikeshed can massively speed things up and bring more diverse ideology to the discussion.

 - Rust also has nested modules. But they are freely allowed to cyclically reference each other and so aren't semantically stand-alone at all.

 - Rust has a notion of "crates" which form a DAG and are compilation units.

 - So if you squint, Rust's notion of crates is more like our modules, and Rust's notion of modules is like our qualified imports. The latter are "pure namespacing".

 - OCaml has mli files like backpack signatures and hs-boot files. Rust follows Java where the "publicity" is part of on the definition. Haskell has a weird middle ground where the export is list is separate from the definitions, but utterly impossible to interpret without them (`module Foo` is weirdly *un*ambiguous in this respect). Haskell is maybe the best, or maybe the worst, of both worlds depending on your preferences.

- Rust has a uniform syntax for renaming things: `use deeply::nested::function as other_function;`. This works for *all* namespaces, despite it also having type aliases. In Haskell you can only bind modules under new names (qualified modules). You cannot export individual items under new names. This seems weird and arbitrary.  `import qualified ... as X` export `module X` feels like bad action-at-a-distance robotic surgery.

- The new 2018 syntax basically adds relative paths to the current create, so that a) the current create doesn't need to know its own name (as before) and b) it is always possible to get around stuff in the current module and a dependency seemingly shadowing each other. Haskell is like old Rust and does *not* have a solution. It's up to you to do an `import as ...` to avoid conflicts. I think if we get freely generated module embedding this would be a bigger problem.

-------

Everyone doesn't like the Haskell module system for at least one reason, and most discussion usually ends up with an intractable bikeshed. So I don't actually want this proposal to fix all the efficiencies I allude to above. Starting with semantics (such arbitrarily nested qualifications, allowing more free generations of things in general) is a good place to begin as I said but still may not be enough. What I really *really* rather see is somebody pull a micro-@goldfirere and sketch out something really really ambitious unifying disparate features.

For example take:
 - monomorphizing quantifier (https://gitlab.haskell.org/ghc/ghc/issues/14917)
 - first class modules
 - type classes
 - backpack (and it's embryo hs-boot)

And imagine what it means to unify all of these :astonished:. There, that's a 2030 goal after dependent Haskell. Now, work backwards and find the "VDQ" the small seemingly-superficial set of changes that clears the way for the ambitious stuff. Maybe that's this proposal! But let's get some sketches along those lines.

In short: fever dream then semantics then syntax.

int-index:
> what happens if you export qualified a module that has its own qualified exports?

My intention was that a qualified re-export appends a qualifier to a name. For example,

```
module StateMonad (module Strict qualified, module Lazy qualified) where
import Control.Monad.State.Lazy as Lazy
import Control.Monad.State.Strict as Strict
```

```
module Platform (module StateMonad qualified) where
import StateMonad
```

```
module CustomPrelude (module Platform qualified) where
import Platform
```

```
module Main where
import CustomPrelude
f = Platform.StateMonad.Strict.get 
```

* `Strict.` qualifier comes from the `module Strict qualified` export
* `StateMonad.` qualifier comes from the `module StateMonad qualified` export
* `Platform.` qualifier comes from the `module Platform qualified` export

Together, they are concatenated into `Platform.StateMonad.Strict.`.

However, note that Haskell's modules are flat, and this continues to be the case. It's just that with each re-export, the qualifiers are concatenated as strings. That is, `Platform.StateMonad.Strict` is a flat qualifier with dots in its name.

I'm usually not a fan of string-oriented solutions, but this design, with the dot being a part of the name, is already the case with the Haskell's module system. And I can't think of a way it could backfire.

----

@Ericson2314 Very nice list (#14917, first class modules, type classes, backpack, hs-boot). I, too, wonder what their unification would look like.

Ericson2314:
> However, note that Haskell's modules are flat, and this continues to be the case. It's just that with each re-export, the qualifiers are concatenated as strings. That is, `Platform.StateMonad.Strict` is a flat qualifier with dots in its name.

But does `import Platform.StateMonad.Strict` work? I wouldn't think so, and on those grounds I'd say the namespace is no longer flat. My mental model is sort of an "overlay filesystem" where modules can have child modules, or modules can be mounted at a name "stealing" it. So there is no `module Data`, but we "steal" `Data.Functor` from any would-be `Data` when we "mount" the functor module with that name. It's not a nice mental model, but I think it accounts for the past and the future.

> @Ericson2314 Very nice list (#14917, first class modules, type classes, backpack, hs-boot). I, too, wonder what their unification would look like.

I was hoping it would be the type of thing you like :).

int-index:
> But does import `Platform.StateMonad.Strict` work?

No, but this does (even today):

```
import Control.Monad.State.Strict as Platform.StateMonad.Strict
```

The alias doesn't need to map to a specific file, it is just a string, and the presence of dots in this string has no effect either.

Ericson2314:
@int-index you're very close to convincing me to stop worrying and learn to ~~love~~ accept the flat path. If/when we treat `.` in names as sugar for nested modules, we'll allow strictly more programs.

Would you allow either of these?
```haskell
import Platform (StateMonad.Strict) -- fairly conservative, fit's two level approach
```
```haskell
import Platform (StateMonad) -- more radical, but unlike my previous on the `Platform` is still a "real module"
```

int-index:
> Would you allow either of these?

1. ```
    import Platform (StateMonad.Strict)
    ```

    I'm not sure about this one. I'd say if we treat `StateMonad` as an atomic export item, it must be imported wholesale with everything inside it, so I'm inclined to reject this one (but could be convinced otherwise with a compelling use case).

2. ```
    import Platform (StateMonad)
    ```

    Not quite like that, there may be a _type_ called `StateMonad`. But we have a solution for that, `-XExplicitNamespaces`, so I would allow

    ```
    import Platform (module StateMonad)
    ```

    Just like we allow

    ```
    import Prelude (type Bool)
    ```

    Both `module` and `type` are namespace specifiers in the examples above.

----

In concrete terms, I imagine the following modification to GHC. Here's what an export item looks today:

```
-- | The original names declared of a certain module that are exported
type IfaceExport = AvailInfo

-- | Records what things are \"available\", i.e. in scope
data AvailInfo

  -- | An ordinary identifier in scope
  = Avail Name

  -- | A type or class in scope
  --
  -- The __AvailTC Invariant__: If the type or class is itself to be in scope,
  -- it must be /first/ in this list.  Thus, typically:
  --
  -- > AvailTC Eq [Eq, ==, \/=] []
  | AvailTC
       Name         -- ^ The name of the type or class
       [Name]       -- ^ The available pieces of type or class,
                    -- excluding field selectors.
       [FieldLabel] -- ^ The record fields of the type
                    -- (see Note [Representing fields in AvailInfo]).
```

I propose we define

```
data AvailInfoQ
  | AvailDef AvailInfo        -- good old AvailInfo for definitions
  | AvailMod                  -- qualified module
      ModuleName                  -- the name of the exported qualified module
      [(ModuleName, AvailInfo)]   -- qualified items

type IfaceExport = AvailInfoQ
```

For example, the `StateMonad` module would export two `AvailInfoQ`:

```
AvailMod
    "Lazy"
    [("Lazy", AvailTC "MonadState" ...), ("Lazy", Avail "runState"), ...]
       -- everything from Control.Monad.State.Lazy

AvailMod
    "Strict"
    [("Strict", AvailTC "MonadState" ...), ("Strict", Avail "runState"), ...]
       -- everything from Control.Monad.State.Strict
```

In `Platform`, these are merged into a single item:

```
AvailMod
    "StateMonad"
    [("StateMonad.Lazy", AvailTC "MonadState" ...),
     ("StateMonad.Lazy", Avail "runState"),
     ...
     ("StateMonad.Strict", AvailTC "MonadState" ...),
     ("StateMonad.Strict", Avail "runState"),
     ...
    ]
       -- everything from Control.Monad.State.Lazy
       --        and from Control.Monad.State.Strict
```

And then in `CustomPrelude`, we export

```
AvailMod
    "Platform"
    [("Platform.StateMonad.Lazy", AvailTC "MonadState" ...),
     ("Platform.StateMonad.Lazy", Avail "runState"),
     ...
     ("Platform.StateMonad.Strict", AvailTC "MonadState" ...),
     ("Platform.StateMonad.Strict", Avail "runState"),
     ...
    ]
       -- everything from Control.Monad.State.Lazy
       --        and from Control.Monad.State.Strict
```

nomeata:
What is the status of this PR?

deepfire:
@nomeata, WIP -- I will post a reply soon.

deepfire-pusher:
> > You seem to propose that every module that intends to export (or, potentially, re-export) a set of aliases, has to mention every single alias in the export list
> 
> Not quite. This is true for an export, but not true for a re-export. For example, here's a module `Containers` that exports 4 type names (Map, Set, HashMap, and HashSet) and 4 aliases (Map, Set, HashMap, HashSet).
> 
> ```
> module Containers
>   ( Map,
>     Set,
>     HashMap,
>     HashSet,
>     module Map qualified,
>     module Set qualified,
>     module HashMap qualified,
>     module HashSet qualified,
>   ) where
> 
> import Data.Map as Map
> import Data.Set as Set
> import Data.HashMap.Strict as HashMap
> import Data.HashSet as HashSet
> ```
> 
> Indeed, this is more wordy than your version:
> 
> ```
> module Containers
>   ( Map,
>     Set,
>     HashMap,
>     HashSet,
>     aliases
>   ) where
> 
> import Data.Map as Map
> import Data.Set as Set
> import Data.HashMap.Strict as HashMap
> import Data.HashSet as HashSet
> ```

This is not an apples-to-apples comparison -- you are comparing an abbreviated to a non-abbreviated version.  Under my proposal, the closest equivalent for your example would be:

```
module Containers
  ( Map,
    Set,
    HashMap,
    HashSet,
    aliases (Map, Set, HashMap, HashSet)
  ) where

import Data.Map as Map
import Data.Set as Set
import Data.HashMap.Strict as HashMap
import Data.HashSet as HashSet
```

> However, it is more consistent with what we have today. After all, we don't write this:
> 
> ```
> module Containers
>   ( typenames,
>     aliases
>   ) where
> 
> import Data.Map as Map
> import Data.Set as Set
> import Data.HashMap.Strict as HashMap
> import Data.HashSet as HashSet
> ```

The abbreviated, whole-sale `aliases` export (which means "export the entire aliased namespace") corresponds to current Haskell's self-re-export option -- which means "export the entire unaliased namespace":

```
module Containers
  ( module Containers
  ) where

import Data.Map
import Data.Set
import Data.HashMap.Strict
import Data.HashSet
```

> And I don't see why you give aliases and type names different treatment.

Now, that's the root of our disconnect -- I propose that we endow the export language for the aliases the same expressive power as that of regular names -- so we can choose to export selectively, via subsetting.

You, instead, are proposing a restricted expressivity option, only allowing either full reexport, or nothing.
 
> Now, in re-exports, I don't expect different treatment either. For example,
> 
> ```
> module CustomPrelude
>   ( module Containers,
>     module Control.Monad,
>   ) where
> 
> import Containers
> import Control.Monad
> ```
>
> Here, `CustomPrelude` re-exports _everything_ from `Containers`, its 4 type names and its 4 aliases. Just like it re-exports everything from `Control.Monad` – its types, functions, data constructors, etc.

You underscore my above point.

> > What is your estimate of this cost on modules that do re-exports?
> 
> No cost.

You reduce expressivity, so there is a cost.  Consider banning selective reexport mechanisms from the regular name export language -- as a thought exercise.

> > How do you estimate this compares to re-exporting of plain names?
> 
> We give aliases equal treatment (as described above).

It's clear to me the treatment is unequal -- but maybe you have further clarifications to your idea.

> > How do you think that impacts the idea of having a single prelude-like module defining a structured namespace, and having that namespace re-exported through other modules?
> 
> I believe this use case is addressed adequately.

Again, reduced expressivity, so I don't think that's the case.

int-index:
> Now, that's the root of our disconnect -- I propose that we endow the export language for the aliases the same expressive power as that of regular names -- so we can choose to export selectively, via subsetting.
>
> You, instead, are proposing a restricted expressivity option, only allowing either full reexport, or nothing.

This doesn't sound right. In fact, the way I see it, I'm proposing to have the same expressive power for aliases and for regular names, and you are proposing an *extended* expressivity option, which we do not have for regular names.

We have various entities Haskell programs can export: functions, type constructors, data constructors.

And we can list them one by one in an explicit export list:

```
module M
  ( isNothing,     -- function
    isJust,        -- function
    Maybe          -- type constructor
      ( Just,      -- data constructor
        Nothing )  -- data constructor
  ) where
```

Likewise, your proposal makes it possible to list aliases one by one, as you note yourself:

```
module Containers
  ( Map,         -- type constructor
    Set,         -- type constructor
    HashMap,     -- type constructor
    HashSet,     -- type constructor
    aliases
    ( Map,       -- module alias
      Set,       -- module alias
      HashMap,   -- module alias
      HashSet    -- module alias
    )) where

import Data.Map as Map
import Data.Set as Set
import Data.HashMap.Strict as HashMap
import Data.HashSet as HashSet
```

However, what your proposal does *additionally* is special syntax to re-export all aliases:

```
module Containers
  ( Map,         -- type constructor
    Set,         -- type constructor
    HashMap,     -- type constructor
    HashSet,     -- type constructor
    aliases      -- all module aliases!
  ) where

import Data.Map as Map
import Data.Set as Set
import Data.HashMap.Strict as HashMap
import Data.HashSet as HashSet
```

If we had anything of the sort for non-aliases, it would look like this:

```
module Containers
  ( types        -- all type constructors!
    aliases      -- all module aliases!
  ) where

import Data.Map as Map
import Data.Set as Set
import Data.HashMap.Strict as HashMap
import Data.HashSet as HashSet
```

----

> The abbreviated, whole-sale aliases export (which means "export the entire aliased namespace") corresponds to current Haskell's self-re-export option -- which means "export the entire unaliased namespace"

Ah, but shouldn't the self re-export also export aliases? Consider this:

```
module Containers (module Containers) where

import Data.Map as Map
import Data.Set as Set
import Data.HashMap.Strict as HashMap
import Data.HashSet as HashSet
```

It should be equivalent to

```
module Containers
  ( Map,         -- type constructor
    Set,         -- type constructor
    HashMap,     -- type constructor
    HashSet,     -- type constructor
    aliases
    ( Map,       -- module alias
      Set,       -- module alias
      HashMap,   -- module alias
      HashSet    -- module alias
    )) where

import Data.Map as Map
import Data.Set as Set
import Data.HashMap.Strict as HashMap
import Data.HashSet as HashSet
```

After all, `Containers` *does* define `Map`, `Set`, `HashMap`, and `HashSet` aliases. Self re-export should re-export these alias definitions.

deepfire-pusher:
> This doesn't sound right.

Hah, you replied to my initial message, which I then corrected. Indeed the initial message had some mistakes in it.

int-index:
> As I have just outlined, your examples, when fixed, do not show a difference in treatment.

I think we want the same feature but disagree on the minutiae. You *do* give different treatment to aliases and other entities:

> The abbreviated, whole-sale `aliases` export (which means "export the entire aliased namespace") corresponds to current Haskell's self-re-export option -- which means "export the entire unaliased namespace"

This is exactly the different treatment that I'm talking about. You say:

> current Haskell's self-re-export option -- which means "export the entire unaliased namespace"

I say that with this extension enabled, self re-export should also re-export the aliases, because there's no reason to give them a treatment. Same goes for imports. The proposal (as written) says:

> `import *modid* aliases` marks the entire set of `*modid*`'s names exported under aliases to be made available qualified by those corresponding aliases.

Why is this needed? A normal `import *modid*` could easily cover both aliased and non-aliased names.

int-index:
Let's say someone defines a custom prelude that exports aliases. Why make the user write this:

```
import CustomPrelude
import CustomPrelude aliases
```

?

Moreover, I now realize that the proposal as phrased does not allow qualified imports of aliases. As it stands, it does not seem to allow this:

```
import qualified Containers as C aliases

f = C.Set.union
g = C.Map.union
```

Under my variant, that would be a trivial special case of a normal import:

```
import qualified Containers as C

f = C.Set.union
g = C.Map.union
```

deepfire-pusher:
> > As I have just outlined, your examples, when fixed, do not show a difference in treatment.
> 
> I think we want the same feature but disagree on the minutiae. You _do_ give different treatment to aliases and other entities:
> 
> > The abbreviated, whole-sale `aliases` export (which means "export the entire aliased namespace") corresponds to current Haskell's self-re-export option -- which means "export the entire unaliased namespace"
> 
> This is exactly the different treatment that I'm talking about. You say:
>
> > current Haskell's self-re-export option -- which means "export the entire unaliased namespace"
> 
> I say that with this extension enabled, self re-export should also re-export the aliases, because there's no reason to give them a treatment.

Well, there is -- this would change interpretation of modules, when recompiled with the new language extension.

I think that while it's generally safe (the user still has to add the extension somehow), it's still unwise.

int-index:
> Well, there is -- this would change interpretation of modules, when recompiled with the new language extension.

This is true for many extensions. Take `PolyKinds` for example. If you define `data Proxy a = P` in a module without `PolyKinds`, you get `Proxy :: Type -> Type`; with `PolyKinds` you get `Proxy :: k -> Type`.

> I think that while it's generally safe (the user still has to add the extension somehow), it's still unwise.

Perhaps. But I'm not convinced (yet). All I see is better ergonomics.

deepfire-pusher:
What if the user wants to retain the old behavior for their module, and simply doesn't want to export aliases?

int-index:
> What if the user wants to retain the old behavior for their module, and simply doesn't want to export aliases?

Easy: they don't enable the extension!

You might object: but what if they want to use someone else's aliases without exporting their own?

Easy: if you don't introduce the `import aliases` stuff, using someone else's aliases doesn't need the extension.

deepfire-pusher:
> > What if the user wants to retain the old behavior for their module, and simply doesn't want to export aliases?
> 
> Easy: they don't enable the extension!
> 
> You might object: but what if they want to use someone else's aliases without exporting their own?
> 
> Easy: if you don't introduce the `import aliases` stuff, using someone else's aliases doesn't need the extension.

If we don't introduce `import aliases`, then we lose fine-grained control over our non-alias namespace.

deepfire-pusher:
> > `import *modid* aliases` marks the entire set of `*modid*`'s names exported under aliases to be made available qualified by those corresponding aliases.
> 
> Why is this needed? A normal `import *modid*` could easily cover both aliased and non-aliased names.

Because we want to give the aliased names the same treatment as the unaliased names -- we want to use the subsetting machinery for fine-grain selection of what comes into our namespace.

This is how you shape the namespace you'll be exporting later.

int-index:
> If we don't introduce import aliases, then we lose fine-grained control over our non-alias namespace.

What sort of control is lost? An alias could be an export entity just like a function or a type constructor. It can be be imported selectively or hidden.

```
import Containers hiding (module Map)
```

This would only import the `Set`, `HashMap`, and `HashSet` aliases, hiding the `Map` alias. Here, `module` is a namespace specifier (just like someone can write `import M hiding (type T)` today).

> Because we want to give the aliased names the same treatment as the unaliased names -- we want to use the subsetting machinery for fine-grain selection of what comes into our namespace.

You are talking about _aliased names_ and _unaliased names_. This is a different mental model. We can make an alias an exported entity on its own. Then everything fits together nicely and there's no need for a special treatment of aliased names.

deepfire-pusher:
> > If we don't introduce import aliases, then we lose fine-grained control over our non-alias namespace.
> 
> What sort of control is lost? An alias could be an export entity just like a function or a type constructor. It can be be imported selectively or hidden.
> 
> ```
> import Containers hiding (module Map)
> ```
> This would only import the `Set`, `HashMap`, and `HashSet` aliases, hiding the `Map` alias. Here, `module` is a namespace specifier (just like someone can write `import M hiding (type T)` today).

So you are indeed proposing different syntax for the same thing, then, I see.

> > Because we want to give the aliased names the same treatment as the unaliased names -- we want to use the subsetting machinery for fine-grain selection of what comes into our namespace.
> 
> You are talking about _aliased names_ and _unaliased names_. This is a different mental model. We can make an alias an exported entity on its own. Then everything fits together nicely and there's no need for a special treatment of aliased names.

The current proposal models (and actually implements) aliases as sets of names, tagged with module names.  But from what you say it seems as if you are proposing something else, when you say "We can make an alias an exported entity on its own".

Could you clarify that point?

int-index:
> Because this enables `import *modid* aliases (X, Y, Z)` -- we want to give the aliased names the same treatment as the unaliased names -- we want to use the subsetting machinery for fine-grain selection of what comes into our namespace.

Why introduce new syntax for this? What about:

```
import *modid* (module X, module Y, module Z)
```

This is exactly the same thing that we have today for type constructors:

```
import *modid* (type X, type Y, type Z)
```

----

> The current proposal models (and actually implements) aliases as sets of names, tagged with module names. But from what you say it seems as if you are proposing something else, when you say "We can make an alias an exported entity on its own".

Hm, sounds right to me. You are more familiar with this part of GHC than I, but I outlined my idea in https://github.com/ghc-proposals/ghc-proposals/pull/205#issuecomment-499354703

I think the main difference in expressivity between our approaches is support for qualified alias exports.

Consider (under my proposed syntax)

```
module Containers
  ( type Map, module qualified Map,
    type Set, module qualified Set
  ) where

import Data.Map as Map
import Data.Set as Set
```

Now `Containers` has 4 exported entities:

* the `Map` type constructor
* the `Map` alias
* the `Set` type constructor
* the `Set` alias

At use sites, these entities can be imported qualified:

```
module Main where

import qualified Containers as C

f = C.Set.union
g = C.Map.union
```

There seems to be no way to express this with `alias` syntax.

deepfire-pusher:
> > Because this enables `import *modid* aliases (X, Y, Z)` -- we want to give the aliased names the same treatment as the unaliased names -- we want to use the subsetting machinery for fine-grain selection of what comes into our namespace.
> 
> Why introduce new syntax for this? What about:
> 
> ```
> import *modid* (module X, module Y, module Z)
> ```
> 
> This is exactly the same thing that we have today for type constructors:
> 
> ```
> import *modid* (type X, type Y, type Z)
> ```
> 
> > The current proposal models (and actually implements) aliases as sets of names, tagged with module names. But from what you say it seems as if you are proposing something else, when you say "We can make an alias an exported entity on its own".
> 
> Hm, sounds right to me. You are more familiar with this part of GHC than I, but I outlined my idea in [#205 (comment)](https://github.com/ghc-proposals/ghc-proposals/pull/205#issuecomment-499354703)
> 
> I think the main difference in expressivity between our approaches is support for qualified alias exports.
> 
> Consider (under my proposed syntax)
> 
> ```
> module Containers
>   ( type Map, module qualified Map,
>     type Set, module qualified Set
>   ) where
> 
> import Data.Map as Map
> import Data.Set as Set
> ```
> 
> Now `Containers` has 4 exported entities:
> 
>     * the `Map` type constructor
> 
>     * the `Map` alias
> 
>     * the `Set` type constructor
> 
>     * the `Set` alias
> 
> 
> At use sites, these entities can be imported qualified:
> 
> ```
> module Main where
> 
> import qualified Containers as C
> 
> f = C.Set.union
> g = C.Map.union
> ```
> 
> There seems to be no way to express this with `alias` syntax.

So you are proposing to extend the normal name import duality (aliased, unaliased) to aliases themselves.

Using normal unrestricted (not `qualified`) imports effectively doubles the number of aliased names that will be created in the importing modules.

I think this is a non-trivial cost, but I think it's up to the committee to decide if it's an acceptable one.

int-index:
> So you are proposing to extend the normal name import duality (aliased, unaliased) to aliases themselves.

Yes, exactly. That's what I mean by giving aliases the same treatment as other export entities.

> Using normal unrestricted (not `qualified`) imports effectively doubles the number of aliased names that will be created in the importing modules.

I don't quite follow this one. Could you give an example?

deepfire-pusher:
> > Using normal unrestricted (not `qualified`) imports effectively doubles the number of aliased names that will be created in the importing modules.
> 
> I don't quite follow this one. Could you give an example?

I mean, relative to the initial proposal, where only one set of aliased names is added to the module namespace, your option will add two sets:

```
import Containers
```

will add both `Containers.Map.Map` and `Map.Map` (well and `Map` too, but that's orthogonal to the point).

Consider how this affects compilation of a regular program -- this is doubling the name-attributed memory cost of using the extension in the default case.

int-index:
> Consider how this affects compilation of a regular program -- this is doubling the name-attributed memory cost of using the extension in the default case.

Right, but the exact same thing happens for type constructors and functions. If you do `import Data.Map as Map` today you have not two but *three* ways to reference the same entity:

* `union`
* `Map.union`
* `Data.Map.union`

We're on familiar ground here.

deepfire-pusher:
That's not quite true:
```
$ cat foo.hs
import Data.Map as Map

x = Data.Map.union

$ ghci foo.hs
GHCi, version 8.6.4: http://www.haskell.org/ghc/  :? for help
[1 of 1] Compiling Main             ( foo.hs, interpreted )

foo.hs:3:5: error:
    Not in scope: ‘Data.Map.union’
    No module named ‘Data.Map’ is imported.
```

deepfire-pusher:
> > Consider how this affects compilation of a regular program -- this is doubling the name-attributed memory cost of using the extension in the default case.
> 
> Right, but the exact same thing happens for type constructors and functions. If you do `import Data.Map as Map` today you have not two but _three_ ways to reference the same entity:
> 
>     * `union`
> 
>     * `Map.union`
> 
>     * `Data.Map.union`
> 
> 
> We're on familiar ground here.

That's not quite true:

```
$ cat foo.hs
import Data.Map as Map

x = Data.Map.union

$ ghci foo.hs
GHCi, version 8.6.4: http://www.haskell.org/ghc/  :? for help
[1 of 1] Compiling Main             ( foo.hs, interpreted )

foo.hs:3:5: error:
    Not in scope: ‘Data.Map.union’
    No module named ‘Data.Map’ is imported.
```

int-index:
> That's not quite true:

Thanks, I didn't realize that adding an alias hides the original module name.

But we still agree that we're on familiar ground because we have two ways to reference `union`:

* `union`
* `Map.union`

Right? And same goes for everything else that `Map` exports. Why should exported aliases be an exception?

deepfire-pusher:
> Why should exported aliases be an exception?

Well, it does increase the cost of using the extension.

And also -- consider how often people do use the default alias (i.e. the non-`as`, non-`qualified` `Data.Map.map`).

I'd say we don't see it very often and so this cost is overwhelmingly a waste.

int-index:
> I'd say we don't see it very often and so this cost is overwhelmingly a waste.

Yes, I see what you mean. But this would happen only in case of a self re-export or implicit export list, right?

```
module M (module M) where

import Data.Map as Map
import Data.List
```

And if I understand you correctly, it's undesired to re-export the `Data.List` alias while it *is* desired to re-export the `Map` alias.

What about this: only aliases introduced with `as` are subject to this extension. Would this fix the issue?

deepfire-pusher:
> > I'd say we don't see it very often and so this cost is overwhelmingly a waste.
> 
> Yes, I see what you mean. But this would happen only in case of a self re-export or implicit export list, right?

The cost is borne out of mere existence of a name, and the names are created during import.

What you mention is a further extra cost, indeed.
 
> ```
> module M (module M) where
> 
> import Data.Map as Map
> import Data.List
> ```
> 
> And if I understand you correctly, it's undesired to re-export the `Data.List` alias while it _is_ desired to re-export the `Map` alias.
> 
> What about this: only aliases introduced with `as` are subject to this extension. Would this fix the issue?

I was about to suggest the same -- yes, that would fix the root of the problem indeed -- you'd only pay for what you asked for.

int-index:
> I was about to suggest the same -- yes, that would fix the root of the problem indeed -- you'd only pay for what you asked for.

Good. So, are we in agreement that there's no need for the `aliases` syntax? I'm particularly worried about having to write

```
import CustomPrelude
import CustomPrelude aliases
```

It's annoying enough to write `import CustomPrelude` in every module, and the `aliases` syntax doubles the cost.

deepfire-pusher:
So, I think the major choice before us is between:

1. Strictly tying new behavior to new syntax.
2. Allowing ourselves to reinterpret old syntax. With the escape hatch that "we always can disable the extension", in case we want to retain the "please don't reexport aliases too" option.

Option 2 gives us this feature of qualified aliases.

The option 2 can appear simpler in the sense of making the syntax change minimal. The reinterpretation makes that apparent simplicity questionable, though.

Option 2 slightly adds verbosity at the export site, by requiring the per-export `module` word.

Option 2 removes verbosity, by allowing us to reuse the same import statement.

The escape hatch of option 2 is slightly unflexible.

Did I miss other differences?

int-index:
I think that's an excellent summary. Thanks.

deepfire-pusher:
Ah, indeed, I've forgot the import alias subsetting.

@int-index, how do you propose we subset the set of imported aliases?

int-index:
The same way we subset imported types and functions.

```
import M (f, g, h, type F, type G, type H, module M, module N)
```

deepfire-pusher:
Oh, one import thing is that option 2 makes modules without explicit export lists _very_ expensive

int-index:
> Oh, one import thing is that option 2 makes modules without explicit export lists very expensive

They are as expensive as two imports with the `aliases` syntax

```
import CustomPrelude
import CustomPrelude aliases   -- the possibly expensive part
```

Option 2 simply bundles these two together, expensive or not. I don't see much of a problem here.

int-index:
> Oh, one import thing is that option 2 makes modules without explicit export lists *very* expensive

Wait, I see what you mean now. You're talking about export lists, not import lists. I should read more carefully.

Yes. But this means the extension must be enabled *and* there should be aliased imports in that module.

```
{-# LANGUAGE FirstClassModuleAliases #-}

module M where

import Data.Map as Map
import Data.Set as Set
```

Well, in this case it's just as expensive as the bundled alias export with the `aliases` syntax:

```
module M (aliases) where

import Data.Map as Map
import Data.Set as Set
```

I don't see much of a problem here either.

deepfire-pusher:
The difference is in reinterpretation -- something that wasn't expensive suddenly becomes so (well, if f.e. one decides they like the extension and add `StructuredImports` to the default extension list in the cabal file).

But in case of whole-sale extension use, I guess it is on the user to control those costs.

deepfire-pusher:
Heh, I guess the next thing we might want is importing extensions.

int-index:
One other idea (and I don't know if it has merit) is that if aliases are to be treated in the same way as other export entities, maybe they should be defined with the same syntax as other export entities: an `=` sign.

So, instead of new import/export syntax, add new alias definition syntax:

```
module CustomPrelude where

import Data.List as List   -- not exported

-- exported
module Containers =
  import Data.Map
  import Data.Set

-- exported
module UnorderedContainers =
  import Data.HashMap
  import Data.HashSet
```

This would correspond to the following `aliases` syntax:

```
module CustomPrelude (aliases (Containers, UnorderedContainers)) where

import Data.List as List   -- not exported

import Data.Map as Containers -- exported
import Data.Set as Containers -- exported

import Data.HashMap as UnorderedContainers -- exported
import Data.HashSet as UnorderedContainers -- exported
```

deepfire-pusher:
@int-index, after the exploration we had, I turned to like your proposed change of syntax.

Another feature it has is making the extension easier to use, by not requiring extra words/lines for the base case.

deepfire-pusher:
> One other idea (and I don't know if it has merit) is that if aliases are to be treated in the same way as other export entities, maybe they should be defined with the same syntax as other export entities: an `=` sign.

Oh, I like this idea as well, except, perhaps, for the `module` keyword, which I'm not sure if it's a good choice, since it further conflation between the notion of modules and qualified name sets.

I think a module is a module (a set of imports and exports), and name qualification is a notion that is separate, even if derived from modules.

int-index:
The `module` keyword there is to specify the namespace. We have a similar problem with [top-level kind signatures](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0054-kind-signatures.rst).

We want to write this:

```
Functor :: (Type -> Type) -> Constraint
class Functor f where ...
```

but we have to write this:

```
type Functor :: (Type -> Type) -> Constraint
class Functor f where ...
```

simply because there might be

```
data T = Functor
```

and we have to disambiguate between the data constructor namespace and type constructor namespace.

Aliases belong to the module namespace, so the disambiguation keyword is `module`.

In a world with no name conflicts I'd suggest this syntax instead:

```
Containers =
  import Data.Map
  import Data.Set
```

deepfire-pusher:
> Aliases belong to the module namespace, so the disambiguation keyword is module.

Ah, but that was actually the question I was raising -- whether they do indeed truly belong.

int-index:
> Ah, but that was actually the question I was raising -- whether they do indeed truly belong.

If they were to be put in different namespaces, what would be the meaning of `f = M.x`? Do we look up `M` in the alias namespace or the module namespace? If they are distinct, then both can be in scope.

deepfire-pusher:
> If they were to be put in different namespaces, what would be the meaning of f = M.x? Do we look up M in the alias namespace or the module namespace? If they are distinct, then both can be in scope.

I'd say aliases belong to a namespace that includes the module namespace.

But that namespace also include names.

This is complicated.

int-index:
Well, there's no need to make it complicated. Various entities can belong to the same namespace. For example:

```
class C where ...
data D = ...
newtype N = ...
type family F where ...
```

Here, `C`, `D`, `N`, and `F`, all belong to the _type_ namespace, even though one is a class, one is a data type, one is a newtype, and one is a type family.

Likewise, modules and aliases can both belong to the _module_ namespace, even though they differ in their nature.

deepfire-pusher:
> Likewise, modules and aliases can both belong to the module namespace, even though they differ in their nature.

Fair enough, I guess.

deepfire-pusher:
> Likewise, modules and aliases can both belong to the module namespace, even though they differ in their nature.

Fair enough, I guess.

However, as you exemplified, different kinds of entities are introduced into the same namespace using different keywords.

This is why I think reusing the `module` keyword is not such a good idea -- it pushes the conflation  beyond "different things in same namespace" further to "same things".

int-index:
> This is why I think reusing the `module` keyword is not such a good idea -- it pushes the conflation beyond "different things in same namespace" further to "same things".

That's a good point. Well, I don't have a strong opinion here. There are three options I see:

1. Reuse the `import ... as` syntax:

   ```
   {-# LANGUAGE FirstClassModuleAliases #-}
   
   module M where
   
   import Data.Map as Map
   import Data.Set as Set
   ```

2. Use the `module` qualifier:

   ```
   {-# LANGUAGE FirstClassModuleAliases #-}
   
   module M where
   
   module Map = import Data.Map
   module Set = import Data.Set
   ```

3. Come up with a new qualifier, e.g. `alias`:

   ```
   {-# LANGUAGE FirstClassModuleAliases #-}
   
   module M where
   
   alias Map = import Data.Map
   alias Set = import Data.Set
   ```

And there are slight downsides to each. (1) is a reinterpretation of existing syntax, (2) may cause a confusion between modules and aliases, and (3) steals syntax by introducing a new keyword.

Which one is the most benign is hard to assess. Not introducing new syntax sounds good to me, so I'd probably go with (1). What do you think?

Ericson2314:
I many times meant to get back to @int-index's proposed types, but struggled to clarify my thoughts each time. I think

1. Getting rid of whatever notion of aliases vs normal bindings we have is good, especially long term when the size of the change proposed recedes into distant memory.

2. If anything @int-index's data type isn't flat enough. I like all-flat and all-nested as two easy to understand normal forms, but @int-index's did have 1 layer. Maybe there is a reason for that that I am missing.

3. On this latest talk of `=`. Remember more than modules is needed. Two imports of identical type alises vs the same type alias imported twice are different, in that the former leads to ambiguous names but the latter won't. Like Rust, I think there should be a way of binding any sort of thing under a new name.

Ericson2314:
As to the type, what about
```haskell
data IfaceExport = IfaceExport (Maybe ModuleName) AvailInfo
```
Perhaps we should debate the types as a way to really drill in on the semantics and then return to the syntax.

Ericson2314:
For context, the maximally nested one is:
```haskell
data IfaceExport 
  = IFaceExport_Item AvailInfo
  | IFaceExport_Qualified ModuleName IfaceExport
```
Where `ModuleName` perhaps cannot contain `'.'`.

deepfire-pusher:
@Ericson2314, if you are curious, the original proposal (without the syntax change we agreed with Vladislav) has been implemented in https://github.com/deepfire/ghc/tree/structured-imports

deepfire-pusher:
> And there are slight downsides to each. (1) is a reinterpretation of existing syntax, (2) may cause a confusion between modules and aliases, and (3) steals syntax by introducing a new keyword.

Point (1) also has a downside, in that it doesn't allow the grouping you mentioned previously:
```
aliases Containers =
  import Data.Map
  import Data.Set
```

I like the option (3) the most, but it might be a harder one to agree on.

deepfire-pusher:
> And there are slight downsides to each. (1) is a reinterpretation of existing syntax, (2) may cause a confusion between modules and aliases, and (3) steals syntax by introducing a new keyword.

@int-index, point (1) also has a downside, in that it doesn't allow the grouping you mentioned previously:
```
aliases Containers =
  import Data.Map
  import Data.Set
```

I like the option (3) the most, but it might be a harder one to agree on.

deepfire-pusher:
> And there are slight downsides to each. (1) is a reinterpretation of existing syntax, (2) may cause a confusion between modules and aliases, and (3) steals syntax by introducing a new keyword.

@int-index, point (1) also has a downside, in that it doesn't allow the grouping you mentioned previously:
```
aliases Containers =
  import Data.Map
  import Data.Set
```

There is also a question of how to restrict/interpret the `import` statements subordinate to the alias declaration. I'm not sure if we should:

1. Allow:

```
aliases Containers =
  import Data.Map as Foo
  import Data.Set as Bar
```

2. Have the normal import duality in those imports.

Both of those don't make a lot of sense, I think.  What do you think?

deepfire-pusher:
@int-index, also, I think the `alias` vs. `module` choice can be deferred to the committee.

Ericson2314:
@deepfire is the only type change https://github.com/deepfire/ghc/compare/895394c26156993a9405039fe53fde3a6b1414b7...structured-imports#diff-656fbe9ca6c0892764674f039c2ab867R70 this field?

deepfire-pusher:
@Ericson2314, the link doesn't seem to point to a change..

Ericson2314:
@deepfire-pusher There was a bug in github I think, link should work now.

Ericson2314:
@deepfire on a different note, I might merge/rebase onto the latest version of the template, so the examples are in their own sectiton rather than the proposed change section.

deepfire-pusher:
@Ericson2314, there's also https://github.com/deepfire/ghc/compare/895394c26156993a9405039fe53fde3a6b1414b7...structured-imports#diff-bc8c9b8edad6309c0e6cbf2ad7e60ebeR886

int-index:
> 1. Allow:
> ```
> alias Containers as
>  import Data.Map as Foo
>  import Data.Set as Bar
> ```
> 2. Have the normal import duality in those subordinate imports (but I think we already rejected it previously).
> 
> Both of those don't make a lot of sense, I think. What do you think?

I agree that these wouldn't make much sense. It should be just `import *modname*` without the extra fluff.

deepfire-pusher:
@int-index, I just realised the idea of declaring aliases separately has a fatal flaw.

The separate `aliases` clause cannot be the only way to introduce aliases, since we cannot possibly be deprecating the normal `as`-based syntax.

So they are an addition, at best.  And if so, their power/weight ratio is probably damaged significantly.

int-index:
> The separate `aliases` clause cannot be the only way to introduce aliases, since we cannot possibly be deprecating the normal `as`-based syntax.

Wasn't this the entire point? The new-style aliases can be exported, while old-style aliases are for internal use only.

Admittedly, having two styles of aliases is questionable, which is why I said that I'm not sure if this idea has merit when I suggested it.

deepfire-pusher:
I think we still can offer standalone alias declarations as an option in the proposal.

It does solve most of the problem of re-interpretation, at least.

deepfire-pusher:
Alright, there's another problem in export list semantics, in the reinterpretation case:

Consider any traditional re-export scenario:
```haskell
{-# LANGUAGE StructuredImports #-}
module M (module X) where

import qualified M.N as X
import qualified M.O as X
import qualified M.P as X
```

..under new semantics this now introduces `X` as an alias in the export list of `M`.

deepfire-pusher:
@nomeata, the specification was updated with the new syntax proposed by @int-index.

The implementation of the changes is underway.

deepfire-pusher:
@Ericson2314, the committee has to deliberate on the proposal before we get serious with the implementation.

The current code is just a PoC quality thing, it's not serious -- but it is a useful vehicle for learning, nevertheless -- ok, at least it was for me.

chshersh:
@deepfire it is an excellent proposal and a nice feature! I think that having such a feature can increase the UX of libraries and especially alternative preludes. But I have some questions regarding the proposal because something wasn't clear for me when I read it. I hope you don't mind clarifying it a bit, and I wonder how it works :slightly_smiling_face: 

1. I see that the example usage of structured imports has the following code (compressed to only relevant parts):

```haskell
-- module C
module C
  ( module Containers qualified
  )
import Data.Map as Containers
import Data.Set as Containers

-- module B
module B where

import C (module Containers) 

foo :: Containers.Map Int String
foo = Containers.empty
```

Both `Data.Set` and `Data.Map` modules have an exported identifier called `empty`. Does the proposal somehow resolves overloaded names by type as well or the example just is written in a way to show multiple syntactic features at once?

2. Is it possible to write something like `import C (module T as Text)`? For example, if the library author provided the name for a reexported module, I don't like, and I want to use it under a different name. It's okay if the proposal is deliberately made simpler and doesn't cover such use case (or maybe there is a possible workaround for the problem using what the `StructuredImports` feature provide). I wonder whether this is supported or no :slightly_smiling_face:

deepfire-pusher:
> @deepfire it is an excellent proposal and a nice feature! I think that having such a feature can increase the UX of libraries and especially alternative preludes.

Thank you Dmitrii! : -)
 
> Both `Data.Set` and `Data.Map` modules have an exported identifier called `empty`. Does the proposal somehow resolves overloaded names by type as well or the example just is written in a way to show multiple syntactic features at once?

Assuming that `Data.Set` and `Data.Map` has a traditional structure (not qualified names exported),
the following passage from proposal applies:

> The same section of Haskell2010 describes a restriction:
> 
> > The unqualified names of the entities exported by a module must all be distinct (within their respective namespace).
>
> With regards to the qualified name exports, this restriction only applies to the individual sets of exports with individual qualifiers -- it is naturally a name clash to export different entities with the same qualified name.

So, using this wording, we have the following interpretation of your example:

1. "set of exports with individual qualifiers" -- there is just one such exported set, qualified with `Containers`.
2. since both `Data.Map` and `Data.Set` have clashing exports, they cannot be unified under the same alias.

deepfire-pusher:
> 1. Is it possible to write something like `import C (module T as Text)`?
>
>  For example, if the library author provided the name for a reexported module, I don't like, and I want to use it under a different name.
> It's okay if the proposal is deliberately made simpler and doesn't cover such use case (or maybe there is a possible workaround for the problem using what the `StructuredImports` feature provide). I wonder whether this is supported or no slightly_smiling_face

This is a very valid feature request, Dmitrii!  Indeed, the current proposal lacks options for renaming sets of qualified names.

I like your proposed syntax -- it nicely blends in with the proposal, IMO.

Will address this soon-ish, @chshersh.  Thank you!

deepfire-pusher:
@chshersh, fixed/restructured the examples.

proposals/0000-structured-imports.rst
```diff
@@ -233,35 +205,39 @@ The new semantics are to be guarded by a language pragma, such as:
...
 
-Then this table shows what names are brought into scope by the specified import statement:
+    {-# LANGUAGE StructuredImports #-}
 
-.. list-table:: Effects of import statements
-   :header-rows: 1
+    module Containers
+      ( module Map qualified          -- Export the set of names qualified with 'Map' and 'Set', qualified.
+      , module Set qualified          -- ..and the same for 'Set'.
+      , module Containers             -- And the 'Map' and 'Set' types, unqualified.
```
int-index: This has to be `Map`, `Set`, listed explicitly, as they are not defined in `Containers`.

deepfire-pusher: Done, thanks!

deepfire-pusher:
Extended the `import` non-terminal interpretation with:

1. qualifier renaming (thanks @chshersh)
2. subsetting

Added overview of changes to the import side.

deepfire-pusher:
Another pass over the semantics -- added significant precision and eliminated several ambiguities and inconsistencies.

deepfire-pusher:
@int-index, so I finally pinned down what gave me this sense of unease in our discussion from the very beginning.

And I fully acknowledge that I horribly failed to articulate it, veering down tangentials.
My apologies.

You write:

> I suggest the following design. Allow qualified re-exports:
>
> ```
> module Foo (module qualified MonadStuff) where
> 
> import Control.Monad.Reader as MonadStuff
> import Control.Monad.State as MonadStuff
> import Control.Monad.Except as MonadStuff
> ```
>
> Now anyone who writes import Foo, gets MonadStuff.MonadReader, MonadStuff.MonadState, and MonadStuff.MonadExcept in scope.

But this is not a re-export of the qualified names, this is a plain export, since the qualifier was only introduced locally,
even if its _definition_ came from elsewhere.

Haskell98's whole-sale re-export of a module doesn't need to mention names it applies to,
so if applied to qualified names it would have looked as follows:

> ```
> module Foo (module Bar) where
> 
> import Bar (module MonadStuff) -- ..or even, just:
> import Bar                     -- ..so MonadStuff is not even mentioned.
> ```

So, to facilitate those whole-sale re-exports, we need to add tracking of the
relationship between imported modules and sets of qualified names they introduce,
as the un-extended Haskell98 doesn't account for this -- it merely relies on the
local namespace for selection of re-exports.

This is a crucial point.

I'm thinking of introducing a clear separation between _exporting locally-established qualified names_,
and re-export of import-introduced sets of qualified names.

The whole-sale re-export that includes qualified names would look like as per above example,
as it is consistent with the Haskell98 re-export semantic.

Introduction of qualified names (the example you cited), on the other hand, is entirely new business,
so I think it naturally deserves a separate syntax:

```
module Foo (qualified MonadStuff) where

import Control.Monad.Reader as MonadStuff
import Control.Monad.State as MonadStuff
import Control.Monad.Except as MonadStuff
```

I.e. to maximise keyword reuse, we'd use just `qualified` in a leading position, without the `module` keyword.

As an aside, the original aliases proposal dealt with this, albeit admittedly in a less graceful manner,
by employing a wildcard `aliases` export entry.  But I don't miss old times, it wasn't pretty,
so thanks for taking effort to change my mind!

What I do regret though, is that in process of discussing I've lost sight of the bigger picture..

Another proposal rewrite pending..

deepfire-pusher:
Updated the proposal to address https://github.com/ghc-proposals/ghc-proposals/pull/205#issuecomment-531566963

Ericson2314:
@deepfire I'm a bit confused on your point. I think if anything the qualifed export is simplier than unqualified export. AFIAK haskell 98 allows all sorts of things that require lots of tracking information, like:
```haskell
module Foo (module MonadStuff) where

import Control.Monad.Reader as MonadStuff hiding (foo, bar, baz)
import Control.Monad.State as MonadStuff (foo, bar)
````

> we need to add tracking of the relationship between imported modules and sets of qualified names they introduce

But per the above we already need to track the import provenance of unqualified names. VS with just qualified exports we don't need to track anything; just grab that upper case name no worries where it came from or what it contains.

deepfire-pusher:
@Ericson2314, I believe there is a confusion of:

1. introduction of a locally-established qualified name and,
2. re-export of a set of names exported by a module.

In the above example `MonadStuff` is locally-established qualified name.
The re-export that goes on is with regards to the _bindings_ of names qualified with `MonadStuff`.
The names themselves are not re-exported, obviously -- as you cannot "re-export" something that was made anew in the local module.

In the general sense, re-export allows not mentioning the names (or, qualifiers) being re-exported.

That's not the case in the example.

Ericson2314:
Eh....I just don't think the `re-` part is that big a deal. Let's say qualifying was the default. You import a name, you export a name. Then "unqualifying" is a flattening operation that explodes and aggregate into its contents. `import Foo` "opens" `Foo` in the local module scope, `module Foo` (in an export list) "opens" Foo in export scope.

If there is an inconsistency, it's that `import Foo` still gives you a `Foo` in addition to its "contents" while `module Foo` only gives you the contents and not also `Foo` itself.

deepfire-pusher:
> Eh....I just don't think the re- part is that big a deal.

@Ericson2314, I'm not sure I understand your point.

For qualified names to have first-class support, we need to support them with operations available to the normal names:

1. introduction into the export list -- like `module Foo (a) where` for normal names and `module Foo (qualified MonadStuff) where` in the proposed extension.
2. re-export -- `module Foo (module Bar) where` -- for both cases, assuming `Foo` exports `a` and `MonadStuff.*`.

deepfire-pusher:
As another angle of exploration/scrutiny of the idea behind this proposal, I have implemented its semantics on top of @yav 's https://github.com/yav/module-98 -- in https://github.com/deepfire/module-98

deepfire-pusher:
As another angle of exploration/scrutiny of the idea behind this proposal, I have implemented (modulo tests) of its latest semantics on top of @yav 's https://github.com/yav/module-98 -- in https://github.com/deepfire/module-98

deepfire-pusher:
As another angle of exploration/scrutiny of the idea behind this proposal, I have implemented (modulo checks) its latest semantics on top of @yav 's https://github.com/yav/module-98 -- in 
https://github.com/deepfire/module-98/tree/structured-imports

deepfire:
Unless unforeseen circumstances of overwhelming force appear, I'm considering this proposal obsoleted by https://github.com/ghc-proposals/ghc-proposals/pull/283 -- except, perhaps, for the details of module namespace merging & reexports, which could potentially be migrated over.

deepfire:
There is a new, promising option in the space: https://github.com/ghc-proposals/ghc-proposals/pull/295, which makes a thorough use of the model established by https://github.com/yav/module-98.

PR #206, opened 2019-02-06
Add support for rendering proposals using sphinx

mpickering:
Documentation can be rendered using `nix-shell shell.nix --run "make html"` or just `make html` if you already have `sphinx-build` installed. 

I also added a .travis.yml file which runs this command.

This yak was shaved because I wanted to render my proposal locally before committing it. With some more care, this style could be refined so that a static page containing all the accepted proposals could be generated and hosted somewhere.


mpickering:
There are still quite a lot of warnings in the build but they don't affect the output that much. If someone wanted to clean those up as well then CI could reject proposals which introduce new warnings.

nomeata:
Neat. Could imagine using Travis and Github Pages (or a similar setup) to auto-generate that.

The theme could be better chosen. The navigation bar on the left doesn’t work well for our long titles. And “Indices and Tables” could go.

mpickering:
I selected a more suitable theme now Joachim. 

Now if someone wanted to set up github pages then they are welcome to but I don't intend to. Anyone who does so should probably spend some more time tweaking the styling/branding and making sure the markup in the proposals is reasonable. It would also be necessary to implement the custom directives which are present at the top of each file.

nomeata:
Nice theme, thanks!

PR #207, opened 2019-02-06
Fix the treatment of type variables in quotations

mpickering:
[View Rendered](https://github.com/mpickering/ghc-proposals/blob/csp-types/proposals/0000-csp-types.rst)

### DIFF
--- a/proposals/0000-csp-types.rst
+++ b/proposals/0000-csp-types.rst
@@ -1,0 +1,153 @@
+Fix the treatment of type variables in quotations
+=================================================
+
+.. proposal-number:: Leave blank. This will be filled in when the proposal is
+                     accepted.
+.. trac-ticket:: Leave blank. This will eventually be filled with the Trac
+                 ticket number which will track the progress of the
+                 implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+There is a `current bug <https://ghc.haskell.org/trac/ghc/ticket/15437>`_ in
+Template Haskell which leads to out of scope type variables in generated
+programs. We solve this problem by applying the
+same principles of cross stage persistence to types as well as values.
+
+
+Motivation
+----------
+
+Consider this simple Template Haskell program.::
+
+   import Language.Haskell.TH.Syntax (Q, TExp)
+
+   get :: forall a. Int
+   get = 1
+
+   foo :: forall a. Q (TExp Int)
+   foo = [|| get @a ||]
+
+In the definition of ``foo`` we quote the function ``get`` and apply it to
+a single type argument ``a``. However, when we splice in ``foo``, we find
+that GHC issues an error as ``a`` has escaped its scope.
+
+.. code::
+
+   Test.hs:6:8: error:
+    • The exact Name ‘a’ is not in scope
+        Probable cause: you used a unique Template Haskell name (NameU),
+        perhaps via newName, but did not bind it
+        If that's it, then -ddump-splices might be useful
+    • In the result of the splice:
+        $foo
+      To see what the splice expanded to, use -ddump-splices
+      In the Template Haskell splice $$(foo)
+      In the expression: $$(foo)
+     |
+   6 | f = $$(foo)
+     |        ^^^
+
+Now analysing ``foo``, we see that ``a`` is bound at stage 0 but used in stage
+1. The problem is that we don't check for stage violations in types like we do
+for variables. In the value case, a cross stage reference to ``x`` is interpreted
+as ``$(lift x)``. The idea is to do precisely the same for types as well.
+
+Proposed Change Specification
+-----------------------------
+
+The proposal is that we apply the principle of cross-stage to types. In order
+to do this we implement a new class called ``LiftT`` which is the type
+level analogue of ``Lift``. It's definition is as follows::
+
+   class LiftT (t :: k) where
+      liftTyCl :: Q Type
+
+What this means is that given a type, we can return the Template Haskell
+representation of that type.
+
+Then, when we see a variable which violates the stage principle we insert an
+implicit splice just as with terms. So now in pseudo-haskell, the original
+program would be modified as follows after renaming::
+
+   foo :: forall a. LiftT a => Q (TExp Int)
+   foo = [|| get @$$(liftTyCl @a) ||]
+
+There are two details to notice.
+
+1. The addition of the ``LiftT`` constraint.
+2. The stage correction by replacing ``a`` with ``$(liftTYCl @a)``.
+
+Unlike ``Lift`` we magically solve all instances of ``LiftT``. It is forbidden
+to write user defined instances as the compiler can generate representations
+for all the types it can represent. The solving works recursively, emitting
+constraints for all free type variables in the type.
+
+For example there are instances for the following type::
+
+   LiftT ()
+   LiftT Int
+   LiftT "abc"
+   LiftT 5
+   LiftT a => LiftT (Eq a)
+   LiftT (forall (a :: Type) . a -> a)
+   (LiftT a, LiftT b) => LiftT (a, b)
+
+.. note::
+   This idea is very similar to how type variables are already handled by
+   the ``StaticPointers`` extension. In that case, free variables give rise to
+   ``Typeable`` constraints rather than ``LiftT`` constraints but the principle
+   is the same.
+
+
+Effect and Interactions
+-----------------------
+
+The primary question of the proposal is why to use ``LiftT`` rather than ``Typeable``.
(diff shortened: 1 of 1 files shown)


goldfirere:
> There are likely quite a lot of user programs will break

Really? Like what?

> the monotype restriction

Does this proposal really solve that? For example, `LiftT (forall a. a -> a)` is invalid Haskell today. Furthermore, `LiftT Proxy` would also go awry, because `Proxy`'s kind is polymorphic.

Does your proposal accept the program in the motivation? Note that this doesn't have a `LiftT` constraint written.

mpickering:
@RyanGlScott showed me a program with untyped template haskell which had some strange dynamic binding property which I think would be rejected by this proposal. I can't recall the exact definition though. Perhaps a lot is quite an exaggeration.

Constraints containing polytypes are solved fine by my tests once you turn on the impredicativity extension. You can see all the examples that work in the [test](https://gitlab.haskell.org/ghc/ghc/merge_requests/166/diffs#8efee896b1723b6c95f28628e2427be3a7c2c755) I added with the patch.

If `Proxy :: k -> *` and there is a `LiftT k` constraint then it will be accepted but otherwise it has to be rejected as the kind variable will escape its scope. This is part of the design of the proposal. 

No the proposal rejects the program because it doesn't have a `LiftT` constraint. It accepts the program if you add the `LiftT` constraint.

RyanGlScott:
> @RyanGlScott showed me a program with untyped template haskell which had some strange dynamic binding property which I think would be rejected by this proposal.

Indeed, I think the only programs that would be rejected under this proposal are the ones that are doing illicit things to begin with. Here is one example that comes to mind:

```haskell
wat :: forall a. Proxy a -> Q Type
wat _ = [t| a |]
```

Currently, this quotation doesn't do what you think it does—it results in `VarT a`! This will cause disaster any time you try to splice it in:

```
λ> :k $(wat (Proxy @Int))

<interactive>:1:3: error:
    • The exact Name ‘a’ is not in scope
        Probable cause: you used a unique Template Haskell name (NameU), 
        perhaps via newName, but did not bind it
        If that's it, then -ddump-splices might be useful
    • In the untyped splice: $(wat (Proxy @Int))
```

This program will be rejected under this proposal, since there's no `LiftT a` constraint in scope. And I'd argue that that's a good thing, since whatever it was doing before was just plain unhygienic.

simonpj:
Broadly I think this makes sense, but there are lots of things I don't understand.

1. There are two entirely separate things
  (a) the existence of `liftTyCl`
  (b) a mechanism for automatically inserting `liftTyCl` splices.

Here (b) is purely a matter of convenience.  Once (a) is sorted, (b) is pretty simple.  So let's focus on (a).

2. You use `$$` so perhaps you are talking about *typed* TH?  But then you should have a guaranteed the spliced type will fit in the hole, and you give no such guarantee.  I think perhaps you are doing *untyped* TH splicing.

3. The type of `liftTyCl` looks odd, because it doesn't mention t or k.  You could stress that its type is
```
liftTyCl :: forall {k} (t :: k). LiftT t => Q Type
```
Also stress that the `Type` here is *not* `Data.Kind.Type` but rather the TH syntax-tree `Language.Haskell.TH.Syntax.Type`.  I was terribly confused.

4. What instances for `LiftT` do you envisage?  Where do they come from? 

5.  You say that polymorphic types give no problem. Can you give a concrete example?  Currently you can't say `f @(forall a. blah)`, and I think you'll want to do just that?

6.Do you have any customers?  Anyone who needs this feature?

mpickering:
Thanks Simon, responding to your points below.

2. As all the splices in this proposal are implicit the distinction between untyped and typed is not very important. It only matters if you want to support typed type splices in user programs. The one example of `$$` is meant to be indicative of some kind of splice as it's not a valid Haskell program.

3. It would be possible to define a kind-indexed type representation which would be analogous to a type indexed expression representation but I didn't want to pollute the proposal with these details as they are not important unless we also add support for splicing types in typed template haskell. 

4. It is explained in the proposal that all types get instance for `LiftT` and are solved automatically. 

5. You can say `f @(forall a . blah)`, I added a test which tries this example and produces the right result. 

https://gitlab.haskell.org/ghc/ghc/blob/0fa5339ed8c896dfe6c7263d5e8d68f7af727a4b/testsuite/tests/th/T15437.hs

6. This fixes a soundness bug, I'm not sure there needs to be any more motivation? Anyone who uses type variables in splices needs this or they will produce bad code. The only other option is to remove the ability to quote types completely by removing untyped type quotations, quoting type applications and quoting type signatures.

goldfirere:
Another solution to the soundness issue is simply not to allow the use of scoped type variables inside of TH quotes.

Also, I'd like to point out that use of polytypes requires `-XImpredicativeTypes`, which is essentially a non-feature.

mgsloan:
This is a cool idea!  It's certainly consistent with the decision to lift values in quotes.  In the past I've used a [very simple `TypeRep -> Q Type` function](https://github.com/fpco/th-utilities/blob/9fe984baf6a567e7cf28e79c05a4bad0d1b367fc/src/TH/Utilities.hs#L129) for this purpose.  The implementation there is missing most of the cases - no handling of tuples, datakinds, etc.

An implementation like this makes sense to me:

* A complete implementation of `liftTypeRep :: TypeRep -> Q Type` added to TH API.
* Reference to some type variable, `a`, from outer scope, would desugar to `$(liftTypeRep typeRep @a)`.

mpickering:
Is there anything else blocking this proposal? It is needed to fix a bad bug in Template Haskell.

nomeata:
Thanks for the proposal, but unfortunately I have to tell you that the committee has rejected it, for the following reasons:
 - While the proposal addresses a specific, real bug in TH, this is far from the simplest solution. Instead, we could just state that no variables brought into scope outside a TH quote remain in scope within the TH quote. The body of `foo` above would be rejected for this reason.
 - There is no motivation for the new feature in the proposal, other than fixing the bug.
 - The example in the proposal is one that, even with this proposal implemented, would not work: it uses *typed* TH, which does not allow type-level splices. So, even if Matt has a genuine need for this feature leading to the example in the proposal, implementing the proposal doesn't actually solve his problem.
 - The proposal mentions the possibility of using `Typeable` instead of `LiftT`, but discards this idea because we cannot solve, e.g., `Typeable (forall a. a -> a)`; in contrast, we can solve `LiftT` instances of that form. However, generating a need for a `LiftT` instance of that form requires `-XImpredicativeTypes`, which is a bug, not a feature.
 - Although Matt has kindly already implemented this feature, we all know that the initial implementation of a feature is only the tip of the iceberg in its cost. Adding this feature, as proposed, introduces yet another class that we have to have custom treatment for, along with various other points of complexity.

Perhaps with proper motivation -- and with an implementation based on Typeable -- we would be more in support. We do think this may be useful. But given the fact that the TH bug has existed since the dawn of time and no one has complained until now -- and even now, there is no practical, on the ground motivation -- we are uninclined to move forward here.

Also, there’s a simple bug-fix for now: just reject the program.  If we get lots of people yelling that they really want such programs, we can tackle it then.

mpickering:
Obviously this is an unsatisfactory conclusion for the user's of typed template haskell.

goldfirere:
> Obviously this is an unsatisfactory conclusion for the user's of typed template haskell.

Why? I don't mean to be dense, but I'm genuinely not sure how you want this feature to work in typed TH. I *do* see how it could work in untyped TH, but I don't see how the proposal motivates adding the type-variable feature instead of just rejecting the original program.

Full disclosure: I believe I am the original author of the text Joachim cribbed in his rejection. (Apologies if I've misremembered.) So feel free to respond to those points.

I do think TH (and typed TH) should grow toward being more expressive, but the motivation behind this proposal is not compelling to me given that there is a much easier way to solve the problem. When/if folks come forward with more concrete motivation, then I (for one) would be happy to reconsider this.

mpickering:
One big problem with typed template haskell currently is that you can end up splicing in an expression which fails to type check because of an ambiguous type variable. If you want to resolve the ambiguity by instantiating that variable using type applications then you have to choose a concrete type rather than a polymorphic type which can be annoying. This proposal means that you can resolve the ambiguity but still remain polymorphic which is valuable for code reuse.  It's not an ideal solution but a decent enough workaround. 

There are more invasive and correct ways to fix this issue but I didn't want to derail the proposal with those. I have a paper in preparation which discusses more complicated problems to do with implicit arguments and a more invasive way to fix it. 

I think you don't think this will work because there is no source syntax for splicing types into typed quotations? but you don't need a source syntax. The "splices" are inserted implicitly just like how cross stage persistence works for values.

goldfirere:
I'm not worried about the source syntax (though it would be a bit sad not to be able to desugar implicit splices into other source syntax). I'm worried about the fact that TH types are not indexed by their kinds, which would seem necessary to support this in a typed way. But maybe it's OK because we allow only implicit splices, where GHC does know the kind of the type variable in question? It would be helpful to spell all this out. Note that this interpretation is in conflict with Simon's https://github.com/ghc-proposals/ghc-proposals/pull/207#issuecomment-461410184 which observed that implicit splicing was merely a matter of convenience. Again, I suppose that could all work, but I'd want it spelled out to be sure (and not just in a comment: in the proposal).

It would also be very helpful (for me, at least) to have a concrete, practical use-case in the motivation. It's clear that you badly want this: why?

PR #54, opened 2017-06-06
Top-level kind signatures (instead of CUSKs)

goldfirere:
[Rendered](https://github.com/goldfirere/ghc-proposals/blob/kind-sigs/proposals/0000-kind-signatures.rst)

PR #209, opened 2019-02-11
Levity polymorphic lift

harpocrates:
[Rendered.](https://github.com/harpocrates/ghc-proposals/blob/levity-polymorphic-lift/proposals/0000-levity-polymorphic-lift.rst)

Credit goes to @RyanGlScott and @goldfirere for fleshing out most of these ideas in the discussion of https://github.com/ghc-proposals/ghc-proposals/pull/175.

### DIFF
--- a/proposals/0000-levity-polymorphic-lift.rst
+++ b/proposals/0000-levity-polymorphic-lift.rst
@@ -5,7 +5,7 @@ Levity-polymorphic `Lift`
 .. trac-ticket::
 .. implemented::
 .. highlight:: haskell
-.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/208>`_.
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/209>`_.
 .. sectnum::
 .. contents::
 
--- a/proposals/0000-levity-polymorphic-lift.rst
+++ b/proposals/0000-levity-polymorphic-lift.rst
@@ -83,14 +83,18 @@ Change ``TExp``, ``Lift``, ``unTypeQ``, and ``unsafeTExpCoerce`` to take advanta
       lift = unTypeQ . liftTyped
       
       liftTyped :: t -> Q (TExp t)
-      default liftTyped :: (r ~ 'LiftedRep) => t -> Q (TExp t)
-      liftTyped = unsafeTExpCoerce . lift
 
-      {-# MINIMAL lift | liftTyped #-}
+Two importannt observations:
 
-NB: The ``r ~ 'LiftedRep`` is needed because GHC doesn't know how to handle levity polymorphic binders.
-It does mean that instances of ``Lift`` over unlifted types will need to manually implement both ``lift``
-and ``liftTyped``.
+  1. The ``r ~ 'LiftedRep`` is needed because GHC doesn't know how to handle levity polymorphic
+     binders. It does mean that instances of ``Lift`` over unlifted types will need to manually
+     implement both ``lift`` and ``liftTyped``.
+
+  2. By not providing a default implementation of ``liftTyped``, any existing empty instances of
+     ``Lift`` will crash as opposed to loop. This does have the unpleasant side effect that there
+     will be no way to write a CPP-free and backwards-compatible manual instance of ``Lift``.
+     However, since the vast majority of instances can (and should!) be derived with
+     ``-XDeriveLift``, which *is* backwards compatible, this is not a significant problem.
 
 Add to ``Language.Haskell.TH.Syntax`` a handful of new instances of ``Lift`` which are now valid:
 
@@ -140,8 +144,8 @@ In terms of development and maintainability, the cost is small: the prototype pa
 this functionality ends up removing more lines from the compiler than it adds.
 
 The only other drawback is that beginners browsing the documentation for Template Haskell are more
-likely to be confused by the complex signature for the default implementations of the ``lift`` and
-``liftTyped`` methods.
+likely to be confused by the complex signature for the default implementations of the ``lift``
+method.
 
 Alternatives
 ------------
--- a/proposals/0000-levity-polymorphic-lift.rst
+++ b/proposals/0000-levity-polymorphic-lift.rst
@@ -108,7 +108,7 @@ Modify ``-XDeriveLift`` to generate typed and untyped splices (for ``liftTyped``
 Effect and Interactions
 -----------------------
 The proposed change would make it possible to lift values of unlifted kinds and would make it
-(officially) possibly for `TExp` to represent expressions of unlifted kinds.
+(officially) possibly for ``TExp`` to represent expressions of unlifted kinds.
 
 Costs and Drawbacks
 -------------------
--- a/proposals/0000-levity-polymorphic-lift.rst
+++ b/proposals/0000-levity-polymorphic-lift.rst
@@ -1,0 +1,148 @@
+Levity-polymorphic `Lift`
+=========================
+
+.. proposal-number::
+.. trac-ticket::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/208>`_.
+.. sectnum::
+.. contents::
+
+The `Lift` typeclass lets us lift values into typed or untyped Template Haskell expressions.
+
+::
+    class Lift t where
+      lift :: t -> Q Exp
+      liftTyped :: t -> Q (TExp t)
+
+    newtype TExp a = TExp { unType :: Exp }
+
+One shortcoming of this class is that it does not let us lift values whose kind is not `Type`.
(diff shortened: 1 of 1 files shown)


goldfirere:
It's disappointing that this proposal forces `lift` to remain in `Lift`. If we could somehow declare `lift = unTypeQ . liftTyped` and force GHC to inline it everywhere, we'd be fine. (Well, we wouldn't be fine on higher-order uses of `lift`. Blast those functional languages.)

Otherwise, I'm strongly in support.

simonpj:
> The main drawback is that uses of lift, TExp, unType, unTypeQ, and unsafeTExpCoerce which have visible type applications will be broken (since the type argument would now refer to the runtime rep).

Could the runtime-rep arg be Inferred, leaving the other as Specified?  That would resolve this niggle

simonpj:
Generally, I'm supportive of this proposal.  But I'm not sure why it's focused on `Lift`. One could make a very similar proposal for `Num`, and other classes, as the original paper about levity polymorphism says.

harpocrates:
> Could the runtime-rep arg be Inferred, leaving the other as Specified? That would resolve this niggle

Doesn't this require the [Explicit specificity in type variable binders](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0026-explicit-specificity.rst) proposal to be implemented though? If not, could you perhaps include an example of how this could be done?

> Generally, I'm supportive of this proposal. But I'm not sure why it's focused on `Lift`. One could make a very similar proposal for `Num`, and other classes, as the original paper about levity polymorphism says.

I was hoping `Lift` could serve as a first real-world guinea pig. It is nicely suited to that role because `template-haskell` users will be more tolerant of breakage than users of `base` (so we can even continue to tweak `Lift` in coming GHC releases). Besides, I think there is already [another proposal taking aim at the more ambitious goal of making `Eq`, `Num`, etc. levity polymorphic](https://github.com/ghc-proposals/ghc-proposals/pull/30).

simonpj:
> Doesn't this require the [Explicit specificity in type variable binders](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0026-explicit-specificity.rst) proposal to be implemented though?

Hmm, yes perhaps it does. But then, please point this out in the proposal; i.e. that breakage can be avoided if we do the other proposal.

> I was hoping `Lift` could serve as a first real-world guinea pig. I

Again, call that out explicitly and point to the other proposal

harpocrates:
> > Doesn't this require the [Explicit specificity in type variable binders](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0026-explicit-specificity.rst) proposal to be implemented though?
>
> Hmm, yes perhaps it does. But then, please point this out in the proposal; i.e. that breakage can be avoided if we do the other proposal.

I do call this out in the unresolved questions section.

> > I was hoping `Lift` could serve as a first real-world guinea pig. I
>
> Again, call that out explicitly and point to the other proposal

It seems a bit excessive to justify paths _not_ taken by this proposal. This is a bit of a special case though, so I'll amend the proposal.

Ericson2314:
@goldfirere tangential, but I wish us *not* to desugar `lift` now, because typed template Haskell is somewhat hostile to cross compilation, where we wish (and do today via hacks) run TH on a different platform so the types are the wrong ones.

mpickering:
I think `Lift` is a bit different to `Eq` and `Num` because the compiler could implement logic to solve instances automatically for the majority of data types. This makes it even more annoying when it's impossible to write an instance for a unlifted type.

harpocrates:
@nomeata Could we submit this to the committee for review? Discussion seems to have died down and I don't think there are any outstanding concerns.

gridaphobe:
This is somewhat tangential to the actual proposal, but the example in the motivation

```
main = print =<< runQ (unTypeQ [|| I# $$e ||])
  where e = [|| 1# ||]
```

seems like it must be a bug in GHC. 

If I query GHC for the type of the splice, I get

```
> :t [|| 1# ||]
[|| 1# ||] :: Q (TExp Int#)
```

But this doesn't make any sense, because, as the proposal points out, `TExp Int#` is ill-kinded.

simonpj:
> seems like it must be a bug in GHC.

Could you, or someone, file a bug report?

simonpj:
Generally I'm happy with this proposal.

But, before implementation, I do think that we should first decide about the [explicit specificity proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0026-explicit-specificity.rst).  If that proposal was in, we'd probably choose to make `lift` etc have an inferred runtime-rep parameter.

We don't want to get people change their libraries to add extra `@_` parameters, and then change them again to take it out!

gridaphobe:
> Could you, or someone, file a bug report? 

https://ghc.haskell.org/trac/ghc/ticket/16384

gridaphobe:
@simonpj the Explicit Specificity Proposal appears to be accepted already (though perhaps not yet implemented?).

So perhaps @harpocrates can change the proposal such that `lift` takes an inferred RuntimeRep, and the GHC patch can be merged once Explicit Specificity is implemented.

harpocrates:
> So perhaps @harpocrates can change the proposal such that lift takes an inferred RuntimeRep, and the GHC patch can be merged once Explicit Specificity is implemented.

Sure, I'll update the proposal.

@goldfirere is there anything blocking the implementation of the explicit specificity proposal? I see it was accepted 8 months ago, yet it doesn't have an associated Trac issue.

goldfirere:
No -- nothing blocking that proposal, other than the lack of time to implement it. If you want to make a Trac ticket (and update the accepted proposal text to point to it), I'd be grateful. I'd be even more grateful if you wanted to take a stab at implementing it! :) Shouldn't actually be all that hard.

harpocrates:
> No -- nothing blocking that proposal, other than the lack of time to implement it. If you want to make a Trac ticket (and update the accepted proposal text to point to it), I'd be grateful. I'd be even more grateful if you wanted to take a stab at implementing it! :) Shouldn't actually be all that hard.

Sounds good. I'd be open to implementing it, provided you can provide some pointers along the way. I'll make a Trac issue, CC you, and let's discuss there. 😄

RyanGlScott:
I must be missing something. How will explicit specificity help here? Reading [the explicit specificity proposal](https://github.com/ghc-proposals/ghc-proposals/blob/e310040d496a827aa78b5de532dc4a999081f4c2/proposals/0026-explicit-specificity.rst), it doesn't seem to suggest a syntax for controlling whether type variables in class declarations are inferred or not. The closest thing I could find is one mention in the [Unresolved questions section](https://github.com/ghc-proposals/ghc-proposals/blob/e310040d496a827aa78b5de532dc4a999081f4c2/proposals/0026-explicit-specificity.rst#unresolved-questions):

> How will this interact when we have visible type application in types (proposal #80)? For example, consider
> 
> ```haskell
> class C (a :: Proxy k) where ...
> ```
> 
> I want `C` to have only one required argument, `a`. But I also want an explicit binding site for `k`, so I can choose `k`'s kind. A nice new piece of syntax would be
> 
> ```haskell
> class C @(k :: Maybe Bool) (a :: Proxy k) where ...
> ```
> 
> This was suggested by @Saagar-A in the commentary. What if the author wanted `k` to be _inferred_? Then they would have to use a top-level kind signature, as proposed in #54. This last case should be rare enough that making it inconvenient should be OK.

Doesn't that suggest that in order to make the `r` in `lift :: forall r (a :: TYPE r). a -> Exp` inferred, we'd actually need to implement top-level kind signatures first?

harpocrates:
Good point. Really, we'd need both top-level kind signatures and explicit specificity. Assuming I'm reading both proposals right, that's

```haskell
type Lift :: forall {r :: RuntimeRep} (t :: TYPE r). t -> Constraint
class Lift (t :: TYPE r) where { ... }
```

What a mouthful.

gridaphobe:
@harpocrates sorry for the delay!

The committee is concerned about the backwards compatibility story with giving `lift` and `liftTyped` mutually recursive default definitions. This change would break existing empty `Lift` instances, as the new defaults would be non-terminating, and unfortunately the `MINIMAL` pragma issues a warning rather than an error. We had the same concern during the discussion of #175 but had decided in favor of making `liftTyped` the sole method, which is no longer a possibility as of this proposal.

So, what we propose instead is to simply not provide a default implementation of `liftTyped`. This does mean that existing empty instances would have to be modified to implement `liftTyped`, but they could also just switch to `DeriveLift`.

How does that sound to you?

harpocrates:
> So, what we propose instead is to simply not provide a default implementation of `liftTyped`. This does mean that existing empty instances would have to be modified to implement `liftTyped`, but they could also just switch to `DeriveLift.`

Sounds reasonable to me, I'll update the proposal (that might take a couple days - I'm on a trip).

nomeata:
> Sounds reasonable to me, I'll update the proposal (that might take a couple days - I'm on a trip).

Ping! :-)

nomeata:
@gridaphobe now a ping to you :-)

gridaphobe:
Thanks for updating the proposal, this looks good to me. Let's merge @nomeata.

PR #210, opened 2019-02-18
`MINIMAL` should warn about extra definitions

LeventErkok:
The [proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0044-minimal-should-warn-extras.rst) has been accepted; the following discussion is mostly of historic interest.

---

[Rendered.](https://github.com/LeventErkok/ghc-proposals/blob/master/proposals/0000-minimal-should-warn-extras.rst)


simonpj:
Looks good to me.

Can you add to "proposed spec" something along the lines of

"The MINIMAL definition requires that every method m in R is defined in the instance declaration.  Giving a default method for m, in the class declaration, is therefore redundant.  The warning encourages the programmer either to remove the default definition or to adjust the MINIMAL pragma."

I suppose that someone could argue that the default method is a disaster-recovery method -- she wants warnings at the use-site but wants a default method if those warnings fail.  But I think I agree that that scenario is less likely than the one you describe.

LeventErkok:
@simonpj Thanks! Just added your comment, and also suggested a flag that should control this behavior. `-Wminimal-redundant-methods`.

andrewthad:
This looks like a good warning to provide, and I like that you specified that it should be enabled by default. I'm not sure about the interaction with default signatures. From the proposal:

> D = set of all methods with default definitions

> If a method as a definition via the default signatures extension, then that definition should not be added to the set D as defined above. While adding it would be a strict check, I think it is likely to increase the false-positives. Though feedback is welcome on the impact of this.

Let's consider how this would work in practice:

```
class Foo a where
  bar :: a -> String
  default bar :: Show a => a -> String
  bar = show
  {-# MINIMAL bar #-}
```

According to the proposal, no warning would be generated. But, the author of the typeclass has done something utterly absurd. Authors of instances cannot actually use the default implementation. It's the same problem you have without using `DefaultSignatures`. I don't see a reason why `DefaultSignatures` should change anything. By treating such typeclass methods the same as those without a default signature, the proposal becomes more simple, and it warns typeclass authors about more mistakes.

zenhack:
I think this is generally a good idea. The [capnp][1] package actually makes use of the current behavior, so being able to disable it with a flag is important, but I agree on-by-default is the right thing.

---

The use case: the capnpc-haskell code generator generates type classes for interfaces in schema definitions, where each of the methods has a default definition that just reports "unimplemented" to the caller. The vast majority of the time you want to implement all of the methods, so we add a minimal pragma so the user still gets a warning if they miss one, but having the default methods there means that if a module providing the generated code updates to a newer version of the schema which adds methods, the user's code will continue to compile and work as it did before.

[1]: http://hackage.haskell.org/package/capnp

LeventErkok:
Thanks @zenhack. You say:

> The use case: the capnpc-haskell code generator generates type classes for interfaces in schema definitions, where each of the methods has a default definition that just reports "unimplemented" to the caller. 

I thought GHC already did that for you; that you get a default method which essentially said unimplemented. This is of course orthogonal to the `MINIMAL` issue, but I'm curious why you don't just use the auto-generated "unimplemented" function and supply your own?

LeventErkok:
Thanks @andrewthad 

You said:

> ```
> class Foo a where
>   bar :: a -> String
>   default bar :: Show a => a -> String
>   bar = show
>   {-# MINIMAL bar #-}
> ```
> According to the proposal, no warning would be generated. But, the author of the typeclass has done something utterly absurd. Authors of instances cannot actually use the default implementation. It's the same problem you have without using `DefaultSignatures`. I don't see a reason why `DefaultSignatures` should change anything. By treating such typeclass methods the same as those without a default signature, the proposal becomes more simple, and it warns typeclass authors about more mistakes.

I don't understand why you say "_Authors of instances cannot actually use the default implementation."_ If they have a type that's already an instance of `Show`, they get `bar` for free without doing anything, no? I must be missing something.

zenhack:
Quoting Levent Erkök (2019-02-24 13:30:55)

>    Thanks @zenhack. You say:
>
>      The use case: the capnpc-haskell code generator generates type
>      classes for interfaces in schema definitions, where each of the
>      methods has a default definition that just reports "unimplemented"
>      to the caller.
>
>    I thought GHC already did that for you; that you get a default method
>    which essentially said unimplemented. This is of course orthogonal to
>    the MINIMAL issue, but I'm curious why you don't just use the
>    auto-generated "unimplemented" function and supply your own?


To clarify: capnproto is an RPC protocol (in addition to a serialization
format); in this case "report as unimplemented" means send a reply to
the remote network peer indicating that the method is unimplemented, as
opposed to just throwing an exception (locally). In this case, by
"caller" I mean the remote network peer.

LeventErkok:
> To clarify: capnproto is an RPC protocol (in addition to a serialization format); in this case "report as unimplemented" means send a reply to the remote network peer indicating that the method is unimplemented, as opposed to just throwing an exception (locally). In this case, by "caller" I mean the remote network peer.

Ah, I see. Thanks for the clarification!

zenhack:
Quoting Levent Erkök (2019-02-24 16:28:10)

> Thanks for the clarification!

You're welcome!

LeventErkok:
@nomeata 

Hi Joachim.. I think this proposal is now ready for committee review, per the proposal process. Thank you!

nomeata:
Sure thing!

nomeata:
Accepted, if the flag is named `-Wredundant-minimal-methods`. I will make that change as I merge the proposal.

andrewthad:
I just realized I missed the response to my comment. Here was the code in question:

```
class Foo a where
  bar :: a -> String
  default bar :: Show a => a -> String
  bar = show
  {-# MINIMAL bar #-}
```

Responding to:

> I don't understand why you say "Authors of instances cannot actually use the default implementation." If they have a type that's already an instance of Show, they get bar for free without doing anything, no?

You cannot actually use the default implementation of bar (without getting a warning) because the `MINIMAL` pragma is there. Consequently, GHC should warn the author of this typeclass, since anyone who omits `bar` to use the `Show`-based default would get a warning about not implementing enough of the typeclass methods.

isovector:
Just ran into an edge case here and wanted to pile on, since I know this is underway.

I have a typeclass:

```haskell
class Effect e where
  {-# MINIMAL weave #-}

  weave
      :: (∀ x. s (m x) -> n (s x))
      -> e m a
      -> e n (s a)

  default weave
      :: Coercible m VoidF
      => (∀ x. s (m x) -> n (s x))
      -> e m a
      -> e n (s a)
  weave = valid only when m is unused in e

  hoist :: (m ~> n) -> e m a -> e n a
  hoist f = implemented in terms of weave
```

Here I want the `MINIMAL` pragma to make the claim "if you are writing your own instance of `Effect`, you only need to write `weave`". 

In this case, neither GHC's behavior today, nor the proposal address the problem.

GHC's behavior today will complain about the using the default instance:

```
        ‘weave’
    • In the instance declaration for ‘Effect (State s)’
    |
165 |   deriving (Functor, Effect)
    |         
```

but after the proposal, I'll just get a complaint at the definition site instead.

PR #148, opened 2018-06-20
Top-level signatures

goldfirere:
[Rendered](https://github.com/goldfirere/ghc-proposals/blob/top-level-sigs/proposals/0000-top-level-signatures.rst)

PR #214, opened 2019-03-15
Namespace specifiers

goldfirere:
This proposes using `type` and `data` liberally in programs to specify namespaces where desired. It also hypothesizes a way forward toward dealing with the Dreaded Namespace Problem when imagining dependent types in Haskell.

[Rendered](https://github.com/goldfirere/ghc-proposals/blob/namespaces/proposals/0000-namespaces.rst)

### DIFF
--- a/proposals/0000-namespaces.rst
+++ b/proposals/0000-namespaces.rst
@@ -281,13 +281,15 @@ Proposed Change Specification
 4. Namespace specifiers affect names in Template Haskell quotes.
    Note that the ``module`` namespace specifier makes sense here, too.
 
-5. ``-Wcompat`` warns on uses of ``pattern`` and ``'`` as namespace specifiers.
-   ``-Wcompat`` also warns on uses of ``''`` to denote a type-level Template Haskell
-   name quote.
+5. ``-Wcompat`` warns when:
+
+   * ``pattern`` is used as a namespace specifier.
+   * ``''`` is used to denote a type-level Template Haskell quote.
+   * ``type`` is used without a ``.`` as a namespace specifier (in an import/export list).
+   * ``'`` is used as a namespace specifier.
   
-6. Two releases after this proposal is implemented, it becomes an error to use
-   ``pattern`` as a namespace specifier in import/export lists. It similarly
-   becomes an error to use ``''`` to quote a type-level name.
+6. Two releases after this proposal is implemented, the first three bullets above
+   (``pattern``, ``''`` for TH, and ``type`` without a ``.``) become errors.
    (In contrast,
    the ``'`` syntax will not have a planned phase-out.)
 
@@ -318,6 +320,9 @@ Here are some examples with the proposed syntax::
 
   import My.Library ( type.C, type.T, data.MkT )   -- the two "type"s are redundant but harmless
 
+  module Foo ( type.C, data.T, module.X, module Y ) where ...
+    -- NB: module.X is a synonym for module X.
+
   {-# ANN module.This.Module "Something #-}        -- The "This.Module" is new; no other module may be specified
 
   false = data.not (data.True)    -- "data"s redundant but harmless
@@ -331,6 +336,9 @@ Effect and Interactions
   the oft-referred `proposal`_) are replaced by ``data``.
 
 * Disambiguation is now uniform: use ``type`` or ``data`` anywhere to disambiguate.
+
+* In export lists, the use of ``module.X`` will now work, just like ``module X``. The ``module X``
+  syntax will be retained in order to keep to the standard.
 
 * Note that the ``'`` in ``'[True, False]`` is not exactly the same ``'`` as the
   one in ``'Succ``. The latter modifies a *name*; the former doesn't have a name
--- a/proposals/0000-namespaces.rst
+++ b/proposals/0000-namespaces.rst
@@ -17,8 +17,8 @@ GHC currently maintains several different namespaces. This proposal describes an
 mechanism for giving the user control over which namespace an identifier is meant to
 come from.
 
-In brief: use ``type`` to refer to the type-level namespace, ``data`` to the term-level
-namespace, and ``module`` to refer to the module namespace. (Spoiler: this last idea
+In brief: use ``type.`` to refer to the type-level namespace, ``data.`` to the term-level
+namespace, and ``module.`` to refer to the module namespace. (Spoiler: this last idea
 doesn't yet have a practical application, but perhaps it will someday.)
 
 In the near future, ideas here can be used to clean up the current state
@@ -41,7 +41,7 @@ As the introduction states, we are currently in a sorry state of affairs.
   synonyms (its original application) and for data constructors (but not for term-level functions).
 
 * Furthermore, the new pseudo-keyword ``value`` is used for the same purpose but in different
-  contexts in a recent accepted `proposal`_.
+  contexts in a recently accepted `proposal`_.
 
 .. _`visible dependent quantification`: https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0035-forall-arrow.rst
   
@@ -247,54 +247,179 @@ prefixes can occur in many contexts and will not be listed as exceptions below.
      
 This is a mess!
 
-Proposed Change Specification (1)
----------------------------------
+Proposed Change Specification
+-----------------------------
 
 1. All features are controlled by the ``-XExplicitNamespaces`` extension.
 
 2. Let keywords ``type``, ``data``, and ``module`` be *namespace specifiers*.
 
-3. A namespace specifier may prefix a name any place a name can occur. It states
-   what namespace the name belongs to. (Namespace specifiers may *not* be used
-   where a name is bound. Occurrences only.)
+3. A namespace specifier may be used as the prefix of a ``modid`` from the `Haskell Report`_.
+   Currently, we have these productions in the Report::
 
-4. In terms and types, namespace specifiers bind as tightly as function application.
-   A namespace specifier describes the namespace for all names used in its scope.
-   (This scope may contain multiple names if a ``(`` follows the namespace specifier.)
-   That is, a namespace specifier is parsed just like any other name in an expression;
-   the namespace selection is in effect in its "argument" when considering the namespace
-   specifier as a function. It is a parse error to use a namespace specifier as an argument
-   to another function, as in the type ``Wrong data Zero``; use ``Right (data Zero)`` instead.
+     qvarid -> [modid .] varid
+     qconid -> [modid .] conid
+     ...
 
-5. Namespace specifiers affect names in Template Haskell quotes.
+   We see that ``modid`` is a module identifier. It is defined by ::
+
+     modid -> {conid .} conid
+
+   This proposal modifies this to be ::
+
+     modid -> [namespace_specifier .] {conid .} conid
+           |  namespace_specifier
+
+   In addition, a ``namespace_specifier .`` prefix is allowed wherever an
+   unqualified name occurrence can appear. (This effectively changes productions like
+   that for ``tyvar`` to ``[namespace_specifier .] varid``.) Binding sites
+   may not have a namespace specifier.
+
+   Because qualified names are defined in Haskell's *lexical* syntax, there can
+   be no spaces between the namespace specifier and the ``.``.
+
+4. Namespace specifiers affect names in Template Haskell quotes.
    Note that the ``module`` namespace specifier makes sense here, too.
 
-6. ``-Wcompat`` warns on uses of ``pattern`` and ``'`` as namespace specifiers.
+5. ``-Wcompat`` warns on uses of ``pattern`` and ``'`` as namespace specifiers.
    ``-Wcompat`` also warns on uses of ``''`` to denote a type-level Template Haskell
    name quote.
   
-7. Two releases after this proposal is implemented, it becomes an error to use
+6. Two releases after this proposal is implemented, it becomes an error to use
    ``pattern`` as a namespace specifier in import/export lists. It similarly
    becomes an error to use ``''`` to quote a type-level name.
    (In contrast,
    the ``'`` syntax will not have a planned phase-out.)
 
-8. If this is accepted before the fixities `proposal`_ is implemented, then that
+7. If this is accepted before the fixities `proposal`_ is implemented, then that
    proposal is to be amended to use ``data`` instead of ``value``.
 
-9. It is an error to use a name in a context that does not expect that kind of name.
+8. It is an error to use a name in a context that does not expect that kind of name.
    For example, the use of a type name in a term will be an error, and the use
    of a module name anywhere they cannot already be used is an error.
 
-10. Namespace specifiers are not allowed as the first lexeme at top-level.
+9. ``ANN`` pragmas for modules may now mention the module name. Omitting the
+   module name will become an error in two releases.
 
-11. ``ANN`` pragmas for modules may now mention the module name. Omitting the
-    module name will become an error in two releases.
+Examples
+--------
 
-Proposed Change Specification (2)
----------------------------------
+Here are some examples with the proposed syntax::
 
-This second proposed change specification is, essentially, an annex to the
+  oneElement :: Vec Char (data.Succ data.Zero)) -- if Succ and Zero are unambiguous, the "data" is redundant
+
+  someNames :: [Language.Haskell.TH.Name]
+  someNames = [ 'True, 'data.False, 'type.Int, 'type.Bool, 'type.Either ]
+    -- The "data" there is redundant
+
+  num :: type.Int                 -- the "type" is redundant but harmless
+
+  empty :: type.Vec type.Bool type.Zero   -- error: Zero is not in the type namespace
+
+  import My.Library ( type.C, type.T, data.MkT )   -- the two "type"s are redundant but harmless
+
+  {-# ANN module.This.Module "Something #-}        -- The "This.Module" is new; no other module may be specified
+
+  false = data.not (data.True)    -- "data"s redundant but harmless
+
+  true = data.True :: data.Bool   -- error: no Bool in data namespace
+
+Effect and Interactions
+-----------------------
+
+* Pseudo-keywords ``pattern`` (as used in import/export lists) and ``value`` (as written in
+  the oft-referred `proposal`_) are replaced by ``data``.
+
+* Disambiguation is now uniform: use ``type`` or ``data`` anywhere to disambiguate.
+
+* Note that the ``'`` in ``'[True, False]`` is not exactly the same ``'`` as the
+  one in ``'Succ``. The latter modifies a *name*; the former doesn't have a name
+  to modify. Accordingly, the syntax ``'[True, False]`` remains. (But see an Alternative
+  below for more discussion.)
+
+* The varying defaults of different contexts are not changed, as doing so would be
+  disastrous for backward compatibility. No current programs are newly rejected
+  except those that use ``pattern`` or ``''`` (the TH quote) for namespace selection
+  after two releases.
+
+* The inclusion of ``module`` in this framework is because it fits so nicely. It is not
+  yet useful, but it is my hope that this idea may spur on a proposal for first-class
+  modules.
+
+* Use of `visible dependent quantification`_ in types of terms may still require adding
+  the type level as a secondary namespace in terms. Otherwise, every type mentioned in
+  a term will have to have ``type`` nearby. This detail is left to the proposal for
+  visible dependent quantification in types of terms, which is not covered directly by
+  this proposal.
+
+Costs and Drawbacks
+-------------------
+* There is a backward compatibility annoyance around the removal of ``pattern`` as a
+  namespace specifier, but I do not think anyone will be too put out.
+
+* ``type`` and ``data`` are certainly noisy, especially if we consider ``data`` as
+  a replacement for ``'``.
+
+* Calling a function, such as ``(+)`` a ``data`` is awkward. Yet it is simply too
+  tempting to use ``data`` here, due to its status as a keyword.
+
+* I do not expect this to be all that difficult to implement.
+
+Alternatives
+------------
+* We could just drop the bit about ``module``.
+
+* We could use ``value`` as the namespace specifier for data-level variables. However,
+  we could not then use it in contexts like terms and types; it could never replace
+  ``'``, for instance.
+
+* A previous version of this proposal did not consider a namespace specifier like a module
+  identifier, but more like a function applied to its argument. This old version is `available <https://github.com/goldfirere/ghc-proposals/blob/97b625aa85f7b77b282546003522ca71b8bb0d7b/proposals/0000-namespaces.rst>`_ in the history. I like this new version more. The old version allowed
+  syntax like ``Vec Char (data (Succ Zero))`` which is quieter than the current proposal's
+  equivalent, but I think the `future work`_ section below has an idea that's even better.
+
+* We could also deprecate the current ``'`` syntax in types? There are two significant
+  stumbling blocks here: promoted lists and tuples. If I say ``[Int]`` is that ``[] Int``
+  or ``Int : <<nil>>`` (where ``<<nil>>`` unambiguously means the empty list)? Currently,
+  we write ``'[Int]`` for the latter. If we drop ``'``, then it would have to be
+  ``data [type Int]``, which is gross. If we don't allow namespace specifiers to work
+  over more than one name at a time, then it would be ``data [Int]``, which isn't terrible.
+  Similarly, do we have ``(Int, Bool) :: Type`` or ``(Int, Bool) :: (Type, Type)``. To get
+  the latter meaning (the data constructor for tuples), we currently use ``'(Int, Bool)``.
+
+  I think the use of lists and tuples is common enough that we can have special syntax
+  for these cases. The quote-mark is well-established enough. However, it does conflict
+  with ``'[]`` and ``'()`` as Template Haskell name quotes. There's no conflict when
+  there are elements in the list/tuple, so we only have to worry about four names:
+  the list type constructor, the nil data constructor, the unit type constructor, and
+  the unit data constructor. These names could just be exported by, say,
+  ``Language.Haskell.TH.Syntax``; we would advertise that these names cannot be quoted.
+
+Unresolved Questions
+--------------------
+
+None at this time.
+
+Future Work
+-----------
+
+It would be nice to be able to import names from one namespace to another wholesale.
+For example, ::
+
+  import Prelude as type as data
+
+might be new syntax to import the Prelude, but making all names available in all
+namespaces. This could cause clashes. Should these be reported? Should clashes silently
+be resolved? We leave these details for another proposal. This one stands without it,
+but allowing these kinds of cross-namespace imports would help with the noisiness
+of using ``data.`` lots in types in code where many data constructors are used in types.
+The new syntax would also be useful to import names from one namespace into the other
+within one module.
+
+Annex
+=====
+
+This is an annex to the
 proposal. It is a vision for a more expressive future. Accepting this proposal
 *does not* accept this annex. Accepting this proposal *does not* commit us to
 accepting this annex in the future. Instead, it is included here so that we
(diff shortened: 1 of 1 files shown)


int-index:
On the main proposal: I fully support it, and I think it doesn't go far enough. Let's also deprecate `'` as a namespace specifier because we need it for TemplateHaskell name quoting, which is currently impossible on the type level. By the way, we use `'` for quoting names in the data namespace and `''` for quoting names in the type namespace. What do you imagine we should use for the `module` namespace? What about `data '`, `type '`, and `module '` quoters? For example, `makeLenses (type 'Foo)` instead of `makeLenses ''Foo`.

On the annex: I don't understand the motivation behind a new `default` namespace. Just pick `type` or `data` as the default namespace and don't introduce a new entity. If anything, we need less namespaces, not more.

int-index:
Here's my take at the issue from https://github.com/ghc-proposals/ghc-proposals/pull/106#issuecomment-400936000:

* Add a new extension,` -XConsistentNamespaces`, that enables the new behavior.
* Prohibit ticks as namespace specifiers, so given `data Nat = Zero | Succ Nat` and `data P a = MkP`, writing `P 'Zero` is illegal, `P Zero` is legal.
* In modules where `-XConsistentNamespaces` is enabled, warn when declaring data constructors and type constructors with the same name. `data Proxy a = Proxy` will continue to work, but cause a warning.
* Make the data constructor namespace the default in both terms and types. Today the data constructor namespace is the default in terms, but in types it is not and we need ticks to override. With `-XConsistentNamespaces` we just don't need ticks.
* In case there's a namespace conflict, use the `type` qualifier to resolve it. So given `data Proxy a = Proxy`, writing `Proxy` always means the data constructor (regardless of context), and writing `type Proxy` always means the type constructor. Under `-XConsistentNamespaces`, `fn :: Proxy a -> ..` becomes `fn :: (type Proxy) a -> ..`, possibly without the parentheses. 
* A few more examples. `[Int]` becomes `(type [] Int)`, and `'[Int]` becomes just `[Int]`. `()` becomes `type ()` and `'()` becomes just `()`. We also define type `List = type []` and `type Unit = type ()` in `base` to avoid verbose qualifiers.

I really like that with the `type` namespace qualifier, a type synonym

```
type A = ...
```

looks just like a term level definition!

```
a = ...
```

as if it was merely a matter of the namespace (because with DependentHaskell it will actually become just a matter of namespace).

Other advantages of this approach:

* Haskell98 code continues to work, and take advantage of two namespaces
* code that needs `-XDataKinds`, `-XDependentHaskell`, and the like, can enable `-XConsistentNamespaces` and reap the benefits of having no ticks and having very simple namespacing rules: it is always the data namespace except when the `type` qualifier is used.

What other syntactic harmonies does it create? Consider associated type families:

```
class Container c where
  type Elem c :: Type

```

This does look like simply adding another method to `Container`, modulo namespaces!

```
class Container c where
  elem :: Type
```

(There is a bit of inconsistency as to whether the `c` type parameter goes to the left of `::`, but we'll have to deal with this too, I think #148 provides a way forward).

----

>  If we always prefer the data namespace, then the type of `[True, False]` becomes `type [Bool]`. That's terrible.

No, that's perfect! One can define `type List = type []` and write the type of `[True, False]` as `List Bool`. I've argued for this arrangement in https://int-index.com/posts/haskell-punning (see also the [Reddit discussion](https://www.reddit.com/r/haskell/comments/aliohu/termtype_punning_in_haskell_the_hidden_cost_of/)).

nomeata:
Nice! Especially the summary of the state of affairs. Say, who designed this language???

> Calling a function, such as (+) a data is awkward. Yet it is simply too tempting to use data here, due to its status as a keyword.

Not worried about this. Functions are first class values in a functional language. We can stress this!


What still bugs me is that after
```
data Foo = Foo
```
I can write `data Foo` and it refers to the thing on the right of the `=`!

But I see a nice way out: If you drop the rule that namespace quantifiers may not be used in binding positions, I can write
```
data type Foo = Foo
```
which is, like, perfect, because we call `Foo` a data type!

(Maybe forbid `type type Age = type Int`, and just re-interpret `type Age = type Int` as what it is)

simonpj:
Could you extend the proposal with a raft of examples, including  a representative sprinkling of all the current contexts?

Suppose
```
f :: forall a -> a -> Int
```
then it seems noisy to write
```
f (type TYPE (type LiftedRep)) True
```
I'd prefer to write
```
f (type (TYPE LiftedRep)) True
```
Does a namespace qualifier followed by an open-paren cover the entire parenthesized thing?  I hope so.

nomeata:
> Does a namespace qualifier followed by an open-paren cover the entire parenthesized thing? I hope so.
Yes, it does!

simonpj:
Currently with
```
data T = MkT
```
I can write
```
f :: S MkT -> blah
```
and since there is no `MkT` in the type namespace, GHC tries the term namespace.   You propose to stop that, correct?  So we'd ave to write
```
f :: S (data MkT) -> blah
```

simonpj:
Bikeshed: choices for the value-space keyword
* `data`
* `value`
* `term`
Each has some merit.  I'm actually not stressed about stealing the use of "value" as an identifier.  I have never used it, and it's easy to avoid.  (In contrast I sometimes want `in`, as in `i1`, `i2`... `in`.)

simonpj:
> Does a namespace qualifier followed by an open-paren cover the entire parenthesized thing? I hope so.  Yes, it does!

It's hard to tell for sure without seeing some BNF syntax.  Could that be added?  Currently there is just a highly cryptic note:  "A namespace specifier describes the namespace for all names used in its scope. (This scope may contain multiple names if a ( follows the namespace specifier.)"

int-index:
> then it seems noisy to write `f (type TYPE (type LiftedRep)) True`

I think it should be written `f (TYPE LiftedRep) True`, the `type` / `data` namespace qualifiers are only needed when there's ambiguity.

int-index:
>  I'm actually not stressed about stealing the use of "value" as an identifier

Not a good option, it's used widely: https://hoogle.haskell.org/?hoogle=value

int-index:
> Does a namespace qualifier followed by an open-paren cover the entire parenthesized thing? I hope so.

It doesn't sound like a good idea to me, as it creates more context-sensitivity when reading code. Imagine reading something that alternates between `data` and `type`!

```
type F = type [Proxy (data [type Proxy, type Identity])]
```

Just writing it out gave me a headache.

goldfirere:
I have updated the proposal:

* Add treatement of TH quotes in new mechanism.

* Raise question about deprecation of `'` in types.

* Examples added.

* New part of proposal about module-level `ANN` pragmas.

int-index:
The proposed change specification says only this:

> A namespace specifier may prefix a name any place a name can occur. It states what namespace the name belongs to. 

But then in "Effect and Interactions" there's a stronger claim:

> Namespace specifiers can now scope over a region of code, for convenience.

I don't see how it follows, and it's also not clear if this is a good idea because of the additional context-sensitivity that it creates, as I mention in https://github.com/ghc-proposals/ghc-proposals/pull/214#issuecomment-473211051. Inclusion of this "convenience" feature also encourages the style of code where names require disambiguation, while making namespace specification cumbersome encourages using names that don't require disambiguation, so I feel that it creates undesirable incentives. Can we make that a separate proposal?

int-index:
What's the treatment of variable names under this proposal? The motivation section lists visible dependent quantification in terms, so can I write this?

```
id :: forall a. a -> a
id a = id_vdq (type a) (data a)

id_vdq :: forall a -> a -> a
id_vdq _ x = x
```

I would much prefer if we unified the term/type variable namespaces instead, so let's mention it explicitly in the proposal that namespace specifiers cannot be applied to variables.

simonpj:
Thanks for the changes.   Can we please see the change to Haskell's BNF syntax?

simonmar:
>  Each has some merit. I'm actually not stressed about stealing the use of "value" as an identifier...

We do use `value` as an identifier in quite a few places in GHC, e.g. https://gitlab.haskell.org/ghc/ghc/blob/master/compiler/simplCore/CoreMonad.hs#L643 

Sure it's easy to fix in simple local variables like this, but some of the uses in the wild are in APIs that have many users, such as https://www.stackage.org/haddock/lts-12.19/aeson-1.3.1.1/Data-Aeson-Parser.html#v:value

Ericson2314:
This also opens the door to `(type MySum).constructor` and `(type MyConstructor).field`, which is a side effect I'm quite fond of.

AntC2:
If we're thinking about naming conventions, and anticipating dependent Haskell ...

I've always found it weird that if functions at the term level start lower-case, why do functions at the type level (i.e. Type Families) start upper-case? Presumably Type Family names are included in the **Background/Namespaces** point 4. 'Types and type constructors'. But Type Families are not Types, neither are they in general 'constructors', because they're not injective -- that always throws me when reading signatures.

Is there some reason Type Families couldn't start lower-case? If we could begin afresh, I'd want injective TFs to start upper-case (like pattern synonyms), non-injective to start lower.

goldfirere:
I have updated the proposal to add some BNF syntax and to recast namespace specifiers to be like module prefixes, requiring a `.`. Feel free to suggest that the old version of the proposal was better, but this new version seems simpler.

One thing lost in the new version is the ability to scope a namespace specification over several names. In its place, I have added a "Future Work" section that suggests that we find some syntax for importing a whole bunch of type names into terms and vice versa.

In response to specific points above:
- Variable (local) names are treated like any other names. I do not wish to merge the type and data namespaces here, as I want this proposal to be only about specifiers, not changing any other behavior. (Note that merging the namespaces for local variables would be a breaking change.)

- The challenge with making type families lower-case is that implicit quantification of type variables would then have to be changed to make it work only over type variables that are not spelled like a type family. This actually isn't so bad, but there is a reason type families are upper-case. I hope that, in the future, type families are gone entirely, replaced by functions that can be used in types, making this problem moot.

int-index:
The new version is great. `type.` and `data.` clearly apply to a single name and reuse the familiar dot-syntax of qualified names. I feel this syntax reuse is justified because module names in `ModName.f` and namespace specifiers `data.f` solve the same problem, they disambiguate where the name comes from.

> Variable (local) names are treated like any other names.

Does this mean that I can write this?

```
id a = data.a
```

goldfirere:
> Does this mean that I can write this?
>
> ```
> id a = data.a
> ```

Yes.

int-index:
How do export lists look under the new version?

```
module (type T, module Mod) where  -- Option (A)
module (type.T, module.Mod) where  -- Option (B)
```

goldfirere:
I think it should be (B). I will clarify in the proposal. But we will continue to be backwards compatible with the standard.

int-index:
Under this proposal, do we still need top-level kind signatures as a separate syntactic construct? We could say that this is merely a matter of the namespace and say things like this

```
type.Id :: forall k. k -> k
type family Id x where
  Id x = x
```

Note the dot.

goldfirere:
> Under this proposal, do we still need top-level kind signatures as a separate syntactic construct?

That's a very interesting suggestion. I suppose we could. But what do we think about this?

```
foo :: forall k. k -> Type
data foo a = MkFoo
```

Here, `foo` would be in the data namespace, and it has a lowercase identifier. Including this possibility isn't directly implied by your idea, but it seems a natural consequence.

Indeed, why do we distinguish between upper-case identifiers and lower-case ones at all? I can think of only one reason: they allow us to easily spot what's a binding site and what's an occurrence site. For example, if I write

```
bar x = ...
```

that could mean several things (allowing for no distinction between upper-case and lower-case identifiers): It could be a function `bar` pattern-matching on a constructor `x`, it could be a pattern-binding for `x` after matching on a constructor `bar`, it could be a function `bar` binding the variable `x`, and it could be a (useless) pattern-match against `bar x` (where both are constructors). Right now, we know which of these it is because both identifiers are lower-case. But if we allow mixing of cases, we'd have to look to see what's in scope already to interpret. Worse, we'd have to determine whether, e.g., `x` is bound as a constructor or a variable: if it's a variable, then this new `x` shadows the old one.

While nothing in the paragraph above is at all difficult for GHC to sort out, it's very annoying for us mere humans. So, as we imagine blurring the lines between namespaces, it seems we want to keep the lines between cases. This could be enforced by GHC (as it is now) or by convention. For now, at least, I favor having GHC give us a hand here by enforcing it.

So, these musings all lead to this proposed design rule:
 * Upper-case identifiers (including symbols that begin with a `:`) are *matchable* (in the sense of my [thesis](https://repository.brynmawr.edu/cgi/viewcontent.cgi?article=1074&context=compsci_pubs), section 4.2.4), while lower-case identifiers are not necessarily.

Note that we already violate this rule by allowing, e.g., `+` to be a type-level matchable symbol. But we rarely bind symbolic names locally, so perhaps that's why this doesn't bite all that badly. (Still, we may wonder about repairing it.)

None of these musings directly affect the proposal, nor do they really respond to @int-index's comment.... but once we start blurring lines, we need to know which ones we want to keep.

Thus, I support @int-index's thought here and would be happy allowing such declarations even if the name is at the data level (albeit capitalized).

int-index:
>  they allow us to easily spot what's a binding site and what's an occurrence site

They are not sufficient for this. Imagine we promote functions to the type level, then what is the meaning of this signature?

```
lookup :: foreach (xs :: [a]) -> Fin (length xs) -> a
```

Do we treat `length` the same way we treat `a` and implicitly quantify it?

```
lookup :: forall a length. foreach (xs :: [a]) -> Fin (length xs) -> a
```

That's not what the user probably meant! Instead, we want this:

```
lookup :: forall a. foreach (xs :: [a]) -> Fin (length xs) -> a
```

How do we know to do this? That's because `length` is already in scope, I say. So I'm not convinced that using capitalization to determine binding sites works in general.

iamrecursion:
> So I'm not convinced that using capitalization to determine binding sites works in general.

In designing Luna's new syntax we ran into this problem and you're right that it doesn't work in general, unless we missed something. We ended up using the following scoping rule: names get looked up in the context (module scope, in most cases), and if introduced in the scope of the signature we then quantify them. 

Using that, like you mentioned, would yield the correct result, where `length` is a function. 

```hs
lookup :: forall a. foreach (xs :: [a]) -> Fin (length xs) -> a
```

goldfirere:
Thanks for pointing out the problem with type signatures. My understanding is that Idris also follows the rule @iamrecursion describes in Luna. If we did that, too, that would defeat some of my argument about capitalization. Somehow, I'm not as bothered in a type signature as I am by having this problem in a pattern.... though I can't quite articulate why it seems better in a signature.

It might be worth having a new extension `-XRequiredForAll` that never does implicit lexical generalization in types, just to avoid this problem. Then, people could easily opt out of this behavior. (Such an extension would also be useful today, for that matter, as type signatures nested within functions can both implicitly quantify type variables and also use type variables brought into scope somewhere else. There is no local way to tell the difference.)

iamrecursion:
> It might be worth having a new extension `-XRequiredForAll` that never does implicit lexical generalization in types, just to avoid this problem. Then, people could easily opt out of this behavior. 

It really _would_ have to be an opt-out, though. Nevertheless, I can see why this extension would be useful. Worth a separate proposal?

goldfirere:
I've addressed the recent round of comments in an update, including the change from `type T ::` to `type.T ::`.

int-index:
TLKSs apparently have a syntactic conflict with associated types. Consider `C`:

```
class C a where
  type F :: Type -> Type -- TLKS
  type F a :: Type       -- type family header
```

We can tell the TLKS apart from the type family header by the presence of binders on the LHS. We can't do this for nullary type classes:

```
class N where
  type F :: Type  -- TLKS?
  type F :: Type  -- type family header?
```

Fortunately, this proposal resolves this issue:

```
class N where
  type.F :: Type  -- TLKS.
  type F :: Type  -- type family header.
```

I have a prototype implementation of TLKSs, so I urge we submit this proposal to the committee to resolve this issue.

goldfirere:
This is ready for submission to the committee, @nomeata. Thanks!

simonpj:
Top level kind signatures means just that!  Shouldn't we have
```
type C :: Type -> Constraint    -- TLKS
type F :: Type -> Type          -- TLKS

class C a where
  type F a :: Type       -- type family header
```
And now this is no ambiguity

int-index:
@simonpj Yes, however this is not what the accepted proposal says:

> Associated types may be given kind signatures within their classes.

It is the "within their classes" bit that creates trouble. Relevant discussion here: https://mail.haskell.org/pipermail/ghc-steering-committee/2018-August/000750.html

simonpj:
Ok, well, I'm questioning that aspect of the TLKS proposal.   The simple, clear thing is for top level kind signatures to be, well, top level.

@goldfirere ?

goldfirere:
Personally, I think the kind signature for an associated type should be within the class. This is simple in a different way: a kind signature for a datatype is in the same enclosing construct as the definition. Perhaps the name "top-level" kind signature is the problem. If we just say "kind signature", the problem melts away. Note that if this proposal is accepted, then the clash @int-index describes goes away. If this proposal is rejected, perhaps we'll have to re-think.

simonpj:
For some definition of "melt away" :-).   I am not looking forward to explaining this to newcomers:
```
class N where
  type.F :: Type  -- TLKS.
  type F :: Type  -- type family header.
```
It's very strange.  After all, the name-space specifier is just that, about name spaces. So presumably I couls also write
```
class N where
  type.F :: Type  -- TLKS.
  type type.F :: Type  -- type family header.
```
Moreover, in the presence of scoped type variables, the very meaning of a TLKS in a nested scope is strange.
```
class N a b where
  type F :: a -> b
  type.F :: b -> a
```
where the latter perhaps means `type.F :: forall b a. b -> a`, which is indeed a good signature.  Do we have a "hole in the scope" of the tyvars from the class header.  

It seems so much simpler just to have a separate, top level signature.  Less elegant, I grant you, but simple, unambiguous.

nomeata:
Richard, did you ponder the possibility of having
```
modid ->{conid .} conid
      | {conid .} conid . [namespace_specifier .
      |  namespace_specifier
```
rather than
```
modid -> [namespace_specifier .] {conid .} conid
      |  namespace_specifier
```
i.e. put the namespace specifier last?

The small reason why this might be preferrable? If you see code like
```
… type.Prelude.Maybe … data.Prelude.Just …  data.Prelude.id …
```
then it looks like there are two different name spaces, and in each there is a module `Prelude`. In other words, it looks like there are two prelude modules that may or may not have something in common! But with 
```
… Prelude.type.Maybe … Prelude.data.Just … Prelude.data.id …
```
it is clear that there is _one_ module that contains things in different namespaces.

Oh, and it also makes it easier to write 
```
… module.Prelude.type.Maybe … module.Prelude.data.Just … module.Prelude.data.id …
```
should one feel inclined to namespace-qualify all the things, including module names.

int-index:
@nomeata This thought has crossed my mind, I agree with this:

> then it looks like there are two different name spaces, and in each there is a module `Prelude`.

Indeed, `Prelude.type.Maybe` looks better to me than `type.Prelude.Maybe` for this exact reason. Thanks for articulating it!

I am less convinced of the value and validity of `module.Prelude.type.Maybe`. This double qualification looks suspicious (is starts with the `module` namespace specifiers but in the end it refers to a `type`), and I see no use for it.

But changing `type.Prelude.Maybe` to `Prelude.type.Maybe` would be good.

int-index:
@simonpj I agree that it is simpler to put the TLKS on the top level. It also need not be limited to associated types, as there's already an extended proposal in this spirit https://github.com/ghc-proposals/ghc-proposals/pull/148 (which is, ironically, by @goldfirere, who argues against this design :)

simonpj:
Here's a possible principle:

* At _binding sites_ the syntactic context should determine the name space
* At _occurrence sites_ we may use a namespace qualifier to distinguish

So currently in Haskell we have
```
gendecl  ::= vars :: [context =>] type
```
Note `vars` not `qvars`.  You can say `f :: Int -> Int` but not `M.f : Int -> Int`.

So I say that the decl
```
f :: Int -> Int
```
means a term-level signature, and you should write
```
type T :: Type -> Type
```
if you want a kind signature.

_Occurrences_, in terms, types, or export lists, are a different matter.

Does that principle make sense to others?

simonpj:
The proposal says

> However, we could not then use it in contexts like terms and types; it could never replace ', for instance.

I don't understand that.  It seems to me that this namespace qualifier notation can (and arguably should) replace the quote notation in TH.

int-index:
> So I say that the decl `f :: Int -> Int` means a term-level signature, and you should write `type T :: Type -> Type` if you want a kind signature.

@simonpj It runs contrary to one of my goals, which is to collapse these two constructs into one. In particular, I'd like to write things like:

```
not, type.Not :: Bool -> Bool

not True = False
not False = True

type family Not b where
  Not True = False
  Not False = True
```

yav:
I am in agreement that it would be nice to unify the terminology for specifying namespaces in ambiguous contexts, but I am unconvinced that this proposal has the right approach at present.  Here are some concrete thoughts:

  * I fully agree with @simonpj that we shouldn't need this at binding sites.

  * I would go further, and say that we only need special notation for use sites
    in ambiguous contexts (e.g., import/export lists, pragmas, etc.).  However,
    in terms, type signatures, data/class declarations (i.e., what we write
    most of the time!) we shouldn't need this notation.

  * I think restricting the contexts where we need this would allow us to have
    a more readable notation.  Does anyone really plan on writing things like
    `a data.: b data.: data.[]`?  The notation used by `ExplictNamespaces`
    (which I think this proposal started with?) seems quieter and more readable.

  * To me, the ambiguity due to datatype promotion
    (I mean that need to write `'`) suggests that we may want to revisit some
    of the design choices related to that extension, rather than adding more
    syntax to work around the resulting issues.

  * I much prefer the original notation for top-level kind signatures.

int-index:
> At binding sites the syntactic context should determine the name space
> At occurrence sites we may use a namespace qualifier to distinguish

Note that this principle means that we still get to write `type.T :: ...` for a TLKS, because a signature is an occurrence site, not a binding site! Why is it not a binding site? Consider:

```
x = 1

y :: Int
y = 1
```

`x` has no type signature, but it is bound. Thus a definition is a binding occurrence. But then do we say that `y` is bound in the signature or the definition? I cannot justify giving a different answer here, thus I would say that both `x` and `y` are bound by their definition, and the signature for `y` simply mentions this definition.

On the other hand, can we bind names by signatures without definitions? Yes, in `.hs-boot` and `.sig` files. So this binding/occurrence separation becomes suspiciously inconsistent.

What about a different principle: namespace specifiers are allow anywhere. That's simple and gives good control. We have already seen the desire for this control in https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0032-type-data.rst and if we allow namespace specifiers anywhere, they would subsume that proposal. And I find it elegant that a single design addresses many issues at once.

goldfirere:
See #227 about the change to top-level kind signatures in-class vs out-of-class.

goldfirere:
Actually, the proposal *does* address `'[]` and tuples in the fourth "effect and interaction" and in the final "alternative". I have nothing more to add there.

goldfirere:
NB: Comment https://github.com/ghc-proposals/ghc-proposals/pull/214#issuecomment-490502363 was actually made at around 10:30am EDT (1430 GMT) on May 8. GitHub gave it the wrong timestamp, confusingly misordering the comments. I've reported a bug to GH.

yav:
I still remain quite unconvinced that this is the right design:

  1.  syntactically---the notation is quite noisy, and it introduces a 3rd interpretation for `.` (in addition to being an operator and a module-qualified name).  On the above point:

> Does anyone really plan on writing things like a data.: b data.: data.[]? 
> > I hope no one writes that! But I also see no reason to forbid it. Note that, because it is not ambiguous, we can write type Cons a b = a : b today

Well, perhaps this indicates that we should use a different approach more along those lines---when would you recommend one should use these explicit qualifiers in a code, say, to improve readability?

   2. I think treating data constructors as a separate 3rd name space is a mistake, and we don't need to dig ourselves deeper in that direction.   I find it quite unsatisfactory that if I write `data.X` that might be a type or it might be a value, and I can't specify which one I mean.

goldfirere:
Thanks for shepherding, @yav.

1. Yes, the notation is noisy. Each namespace qualification takes an extra 5 characters, and I agree that this is a lot. But I worry that anything shorter will steal too much syntax and/or be context-dependent. There are so many places we need namespace disambiguation, and the goal is unification.

2. I'm not sure what you mean by a 3rd namespace for data constructors. Could you elaborate?

int-index:
For the record, I voiced my support for this proposal before (by saying "I fully support it") but my thinking on the matter has evolved. I, too, find these namespace specifiers unpleasant, and I have an alternative design in mind.

yav:
@goldfirere I got confused:  in the beginning of the proposal data constructors are listed as a separate namespace, and so when I was looking at `data.T` in the examples I thought we were refereeing to *that* namespace.

Now I see that there are only two specifiers: `type` and `data` (not counting the `module` one, which seems quite orthogonal to the whole discussion) and `data` is used to refer to all value level things.  But also promoted data constructors, which are actually types. I find this quite confusing.

I think that, really, my main objection to the proposal (apart from the syntax) is the treatment of promoted constructors.  I think that the notation should emphasize *what* a thing is, and not how the thing happened to be defined.  To me, using the notation `data.C` for a type, is similar to saying that we should use `primitive.Int` to refer to built-in types.

int-index:
> and I have an alternative design in mind

To follow up on that, I briefly described this alternative idea, namespace-qualified imports, in #236:

```
import Data.Proxy type qualified as T
import Data.Proxy data qualified as D

a = f T.Proxy   -- refers to the type constructor
b = f D.Proxy   -- refers to the data constructor
```

I will see how the discussion evolves there and write a full proposal if the overall plan is met with agreement.

simonpj:
I am very sympathetic to the goal of solving the infelicities described so well in the proposal.

But I can't say I find the propsed solution satisfying.  It's noisy, and (although I understand the reasoning) I alwsy find that `data` makes me think of data type declarations (especially when juxtaposed with `type`) rather than of the term level namespace.

I wonder if a simpler (albeit still not great) alternative might be to use single-quote to mean "the other namespace".  So in types, the quote means the term namespace (as now), but the change would be to make ti work the other way around.   So `f 'Int` woudl apply `f` to a Required `Int` argument.

I'm inclined to say "wait" on this proposal.  As in, let's wait for a better idea to come along.

simonmar:
I dislike the single quote:
* it looks ugly
* syntax highlighters almost always get it wrong
* the bad interaction with lexing of characters that @goldfirere points out in the proposal

So I'm in favour of deprecating the single quote in favour of a complete and regular way to specify namespaces. 

I actually don't find this current proposal too noisy, but it seems sensible to hold fire while @int-index 's alternative poposal is being discussed.

LeanderK:
To give the perspective from a "normal" user: I really don't like the single quote because it makes haskell very hard to understand if you don't know whats going on. Especially if dealing with the namespaces will become more common. To comment on @int-index: I think namespace-qualified imports are usable because it gives you a hint what's happening and somewhere to start if you want to ask what the heck this code is doing ("why are there type and data imports in this code?").

I don't want to comment on the technical point of view for @goldfirere proposal, only for the usability from the point of view as an interested observer. I think what the proposal gets right is that the resulting code can be dealt with as a ordinary user. I think if you glance over the problem, the code is somewhat intuitive and can be understood without deep knowledge of the namespace problem (because it's very explicit). Also, if you don't understand what the code is doing, there's a clear, unambiguous question you can ask ("what's the data in false = data.not (data.True)?"). If the syntax can appear in another context it's way hard to ask a question about it. Although I agree that it seems very noisy.

I think beside elegance, there should still be possibility for someone not familiar with the extension to understand it or at least ask a question about it. Otherwise you create a language that only those already familiar with it can use.

goldfirere:
@LeanderK's observations made me realize that I would use this new syntax when teaching Haskell, emphasizing the difference in the `T`s in `data T = T`. (No, I don't use types like that in my own code when teaching, but `aeson` is very very useful, and it has lots of punning.)

I agree that this proposal may be suboptimal and am happy to let it sit dormant while other proposals in this space mature. I would hope, though, that other proposals seek to unify the mess that's elaborated in this proposal.

monoidal:
> Introduce a new proposal for -XRequiredForAll that disables GHC's habit of implicit lexical generalization in types. 

This could be called `-XNoImplicitForAll`; it's feels consistent with `-XNoImplicitPrelude` and `-XExplicitForAll`. Furthermore, "required" forall might cause ambiguity with required variables (as opposed to inferred/specified). One downside is that this extension would probably imply ExplicitForAll, and we don't have a negative flag implying a positive one yet.

(I know, this is only a possible future direction, not strictly a part of the proposal.)

yav:
I think this proposal is ready to go to the "Committee discussion" phase and I will label it as such.  I would kindly ask that further comments are limited to members of the committee.

**Summary**  Names in Haskell are organized in two namespaces---values and types---and some language constructs may contain members of either namespace.   This presents a problem when a name which inhabits both namespaces is used in such a context.   At present, different language constructs have different ways of addressing this ambiguity, many of which look similar but differ in what are probably unimportant ways.   This proposal aims to address this problem by allowing names to be qualified with an explicit namespace qualifier: for example `data.T` would refer to the name in the value namespace, while `type.T` would refer the type namespace.

**Recommendation** I think that the problem which this proposal aims to address is important and worth addressing.  However, I do not like the proposed solution, for the following reasons:
   * I find the notation to be quite noisy, and can't honestly think of situations where I'd use it
   * I find the use of `data` for the value namespace to be quite confusing, as in Haskell it is already strongly associated with declaring datatypes.   I could probably be retrained on this, but I don't think it is great design to use the same notation with rather different meaning---the notation does not seem intuitive and is likely to be confusing to users not familiar with it.
   * I find it unsatisfying that `data.T` may be either a type or a value. I understand that we are just saying we mean `T` in the value namespace, which due to design choices in `DataKinds` can actually be a type, but it is still very confusing.

Based on this I'd say that we should table this for now, but continue thinking of ways to solve the underlying issue.

simonpj:
Thanks Iavor.

My current view ([elaborated above](https://github.com/ghc-proposals/ghc-proposals/pull/214#issuecomment-498045198)): I'm inclined to say "wait" on this proposal. As in, let's wait for a better idea to come along.

goldfirere:
I'm happy to wait, as well, especially in light of #236.

yav:
Ok.  We don't have a "wait" status, but I'll re-label it as "needs revision" for now (as opposed to "reject") indicating that we'd still like to solve the underlying problem.    Depending on what comes out of the other proposals we can re-activate this or just close it later.

goldfirere:
My enthusiasm for this has waned, as #236 suggests that better designs exist. I retreat to rejoin the fight another day.

PR #215, opened 2019-03-22
DerivingAlias

chshersh:
[Rendered](https://github.com/ChShersh/ghc-proposals/blob/chshersh/deriving-alias/proposals/0000-deriving-alias.rst)

A proposal to introduce a way to define and use aliases for deriving statements.

**Example:**

```haskell
deriving alias CoreClasses =
    (Show, Generic)
    newtype  (Eq, Ord, Hashable, FromField, ToField, ToMustache)
    anyclass (FromJSON, ToJSON)

newtype Email = Email { unEmail :: Text }
    deriving alias CoreClasses

newtype Subject = Subject { unSubject :: Text }
    deriving alias CoreClasses
```

### DIFF
--- a/proposals/0000-deriving-alias.rst
+++ b/proposals/0000-deriving-alias.rst
@@ -1,0 +1,246 @@
+Deriving aliases
+================
+
+.. proposal-number::
+.. trac-ticket::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+
+The Deriving mechanism has become increasingly powerful in Haskell. The initial
+goal may have been to ease the burden of developers from writing boilerplate
+code, but so much code is being derived that this method of removing boilerplate
+became boilerplate itself (though in a much more comprehensible way). This
+document is a proposal to improve the deriving mechanism by allowing users to
+define deriving aliases.
+
+
+Motivation
+------------
+
+Haskell applications usually involve multiple layers of communication between
+the components (of the app): database serialization, JSON encoding and decoding,
+Mustache templates, CSV serialization and very often much more. It is very
+common to use newtypes to squeeze more guarantees from the compiler. The
+nice thing about ``newtype`` is that it is possible to derive a lot of
+typeclasses almost for free. Here are the examples of some data types from a
+real world application: ::
+
+  -- | Wrapper over the text of an email.
+  newtype Email = Email { unEmail :: Text }
+      deriving (Show, Generic)
+      deriving newtype (Eq, Ord, Hashable, FromField, ToField, ToMustache)
+      deriving anyclass (FromJSON, ToJSON)
+
+  -- | Subject of the email.
+  newtype Subject = Subject { unSubject :: Text }
+      deriving (Show, Generic)
+      deriving newtype (Eq, Ord, Hashable, FromField, ToField, ToMustache)
+      deriving anyclass (ToJSON, FromJSON)
+
+  -- | Wrapper for primary keys in a database table.
+  newtype Id a = Id { unId :: Int }
+      deriving (Show, Generic)
+      deriving newtype ( Eq, Ord, Hashable, FromField, ToField, ToMustache
+                       , FromHttpApiData, ToHttpApiData)
+      deriving anyclass (FromJSON, ToJSON)
+
+The more data types you have, the more repetitive deriving statements become.
+Each time you introduce a new component in the application, you need to manually
+change the definition of each data type and add more derivings. This makes
+declaring and maintaining new data types a remarkably cumbersome process.
+
+This proposal suggests reducing the problems highlighted above by allowing
+aliases for multiple deriving clauses and a new deriving strategy that can make
+use of this construct. Using these new features the above code can be simplified
+to: ::
+
+  deriving alias CoreClasses =
+      (Show, Generic)
+      newtype  (Eq, Ord, Hashable, FromField, ToField, ToMustache)
+      anyclass (FromJSON, ToJSON)
+
+  newtype Email = Email { unEmail :: Text }
+      deriving alias CoreClasses
+
+  newtype Subject = Subject { unSubject :: Text }
+      deriving alias CoreClasses
+
+  newtype Id a = Id { unId :: Int }
+      deriving newtype (FromHttpApiData, ToHttpApiData)
+      deriving alias CoreClasses
+
+Proposed Change Specification
+-----------------------------
+
+The name of the language extension: ``DerivingAlias``. This extension, when
+enabled, allows us to:
+
+1. Declare new synonyms for reusable deriving definitions.
+2. Use the ``alias`` deriving strategy.
+3. Export and import deriving aliases.
+
+The syntax of declaring a new ``deriving alias`` is the following:
+
+1. ``deriving`` keyword.
+2. Followed by the ``alias`` keyword.
+3. Followed by the name of the alias.
+4. Followed by zero or more type variables.
+5. Followed by the ``=`` operator.
+6. Followed by the list of deriving definitions similar to what is used with
+   data types but without the ``deriving`` prefix.
+
+Several examples of deriving aliases: ::
+
+  deriving alias EnumBundle = (Show, Read, Eq, Ord, Enum, Bounded, Ix)
+
+  deriving alias FAM = newtype (Functor, Applicative, Monad)
+
+  deriving alias MReader env =
+      alias FAM
+      newtype (MonadIO, MonadReader env)
+
+  deriving alias ToElm t =
+      stock (Generic)
+      (Elm, ToJSON, FromJSON) via ElmStreet t
+
+A new strategy named ``alias`` is introduced to allow using these definitions in
+both scenarios — inside other deriving aliases (as shown above) and for data
+type deriving. ::
+
+  data Status = Approved | Rejected | Pending
+      deriving alias EnumBundle
+
+  newtype App a = App { unApp :: ReaderT Env IO a }
+      deriving alias MReader Env
+
+Since ``deriving alias`` would be a new entity in the Haskell language, it
+should be possible to export and import it. The ``deriving`` keyword would need
+to be added as a prefix for exported and imported ``deriving alias`` symbols
+(similar to ``pattern`` from `PatternSynonyms <https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#pattern-synonyms>`_
+and ``type`` from
+`ExplicitNamespaces <https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#explicit-namespaces-in-import-export>`_): ::
+
+    module Foo (deriving MReader) where
+
+    import Bar (deriving FAM)
+
+
+Effect and Interactions
+-----------------------
+
+The proposal aims to remove even more boilerplate from Haskell projects by
+providing the ability to capture common deriving patterns into their own
+entities that can be reused later. Additionally, it becomes easier to introduce
+changes in large applications: once you realize that you need to derive another
+typeclass for all your core data types, you can simply patch relevant deriving
+synonym. Moreover, since Haskell already has a huge number of features in the
+area of type-level computation, ``deriving`` clauses can get quite complex, like
+the one below
+(`original example and discussion <https://www.reddit.com/r/haskell/comments/8y98yt/typedriven_safe_derivation_of_tojson_and_fromjson/>`_): ::
+
+  data OtherConfig = OtherConfig { otrNameOfProcess :: Maybe String
+                                 , otrArgsToProcess :: [String]
+                                 }
+    deriving (Read, Show, Eq, Ord, Generic)
+    deriving (ToJSON, FromJSON)
+         via WithOptions '[ FieldLabelModifier     '[CamelTo2 "-", Drop 3]
+                          , ConstructorTagModifier '[CamelTo2 "-", UserDefined Init]
+                          , SumEnc                  TwoElemArr
+                          , TagSingleConstructors  'True
+                          , OmitNothingFields      'True
+                          ]
+                          OtherConfig
+
+After this proposal is implemented, it will become possible to define common
+deriving definitions, put them into packages and export them to increase the
+code reusability and make the overall user experience more pleasant.
+
+Costs and Drawbacks
+-------------------
+
+The proposed feature increases the power of the
+`DerivingStrategies <https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#deriving-strategies>`_
+extension. Like any other feature, it makes the entrance threshold higher for
+the language and requires Haskell developers to learn more in order to
+understand code that uses this feature. However, since it is very close to mere
+syntax-sugar, it shouldn't be too difficult to learn.
+
+
+Alternatives
+------------
+
+CPP macros
+^^^^^^^^^^
+
+It's already possible to achieve an almost similar level of the reusability
+using the
+`CPP macros <https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/phases.html#standard-cpp-macros>`_.
+You can consolidate common parts of ``deriving`` under a new custom CPP macro
+and use it wherever you need. However, abusing CPP macro-system for this feature
+would likely be a poor solution. Furthermore, it is a well-known fact that
+tooling usually handles CPP macros in a bad way. Since GHC is fed with the code
+after preprocessing, it's not possible to perform a huge bunch of analysis forms
+for macros and their usage.
+
+Template Haskell
+^^^^^^^^^^^^^^^^
+
+It is possible to define a
+`TH macro <https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#template-haskell>`_
+that uses the
+`StandaloneDeriving <https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#stand-alone-deriving-declarations>`_
+extension to derive a lot of typeclasses for you. Like this: ::
+
+    boilerplateAnnigilator :: Name -> Q [Dec]
+    boilerplateAnnigilator typeName = do
+        let typeCon = conT typeName
+        [d|deriving instance Show $(typeCon)
+           deriving instance Eq   $(typeCon)
+           deriving instance Ord  $(typeCon)
+          |]
+
+This approach `was proposed here <https://stackoverflow.com/questions/45113205/is-there-a-way-to-shorten-this-deriving-clause>`_.
+
+It works okay for basic and simple cases. However, it's extremely difficult to
+implement generic logic that will work for every typeclass since standalone
+deriving requires you to specify the context for type variables. With this
+approach, one needs to reimplement internal GHC logic but in Template Haskell.
+
+Reuse type aliases
+^^^^^^^^^^^^^^^^^^
+
+`Quite recently <https://www.reddit.com/r/haskell/comments/9dx6s9/proposal_data_deriving_synonyms/>`_
+it was discussed to reuse existing ``type`` synonyms of kind ``Constraint``,
+like in the example below: ::
+
+  type RS a = (Read a, Show a)
+
+  data Foo = Bar | Baz
+      deriving RS
+
+However, this approach is quite limited:
+
+1. It doesn't allow one to specify a deriving strategy.
+2. It is not possible to integrate
+   `DerivingVia <https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#deriving-via>`_
+   with this approach.
+3. It becomes cumbersome to use when multiple type variables are involved.
+
+Unresolved Questions
+--------------------
+
+1. Should it be a simple syntax-sugar extension where the definition of
+   ``deriving alias`` is checked only syntactically and a more thorough analysis
+   is performed only when the alias is actually used?
+
+
+Implementation Plan
+-------------------
+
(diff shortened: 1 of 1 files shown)


mpickering:
It feels to me that a proposal which makes deriving via work with type families better (similar to option 3) would be preferable and more powerful. Adding specific new syntax for this seems to not have a very good power to weight ratio.

int-index:
Here's something else we could do – a first-class encoding for deriving strategies.

```
data Deriving =
  forall k. Newtype k   |
  forall k. Stock k     |
  forall k. AnyClass k  |
  forall k. k `Via` Type
```

Then we accept `[Deriving]` in data type declarations:

```
newtype CompM a = CompM (Either Err (a, [String]))
  deriving '[ Stock Functor,
              Stock Foldable,
              Stock Traversable,
              Newtype Show,
              Newtype Eq,
              Newtype Ord ]
```

The last step is to allow using type families and type synonyms to build this deriving list:

```
type CoreClasses =
    '[ Stock Generic,
       Newtype Eq,
       Newtype Ord,
       Newtype Show,
       Newtype Hashable,
       Newtype ToField,
       Newtype FromField,
       AnyClass FromJSON,
       AnyClass ToJSON ]

newtype Email = Email { unEmail :: Text }
  deriving CoreClasses

newtype Subject = Subject { unSubject :: Text }
  deriving CoreClasses

newtype Id a = Id { unId :: Int }
  deriving newtype (FromHttpApiData, ToHttpApiData)
  deriving CoreClasses
```

Fresheyeball:
How does this interact with warn-redundant-constaints ?

ChShersh:
@int-index That's an interesting idea. I like turning `deriving` arguments into first-class values. However, it does seem like a much bigger change.

ChShersh:
@Fresheyeball Maybe I misunderstood you, but it looks like this has nothing to do with `-Wredundant-constraints`. Redundant constraints warning is about constraints in the type signatures for functions, not deriving. 

* https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/using-warnings.html#ghc-flag--Wredundant-constraints

However, I think that I can see your concerns. Consider the following code:

```haskell
{-# LANGUAGE DerivingStrategies #-}

data Foo = Foo
    deriving (Show)
    deriving (Show)
```

Currently it doesn't compile due to duplicate instance error. So it might be not clear what to do if you derive two `alias`es that contain the same typeclass (so it will appear twice in the `deriving`). I see two possible solutions for this problem:

1. Do nothing, so users will see the same compiler error as now.
2. When `alias` deriving strategy is used, remove all duplicates during substitution of alias.

int-index:
> However, it does seem like a much bigger change.

In what sense? I see it as a smaller change, as it doesn't introduce new top-level syntactic constructs or keywords.

ChShersh:
@int-index Sorry, it wasn't completely clear from your example. If you are proposing to change behavior of the existing `DerivingStrategies` extension then it does look like a bigger change to me. However, if you're talking about completely new proposal like `FirstClassDeriving` then I can't comment much on the difference in the implementation since I don't have that much experience working with GHC. In my opinion, both language extensions (`DerivingAlias` and `FirstClassDeriving`) are good. And, probably, it doesn't make much sense to have both of them implemented. I think that both proposals have their own benefits and drawbacks.

RyanGlScott:
This is an idea that I've pondered before, but I'm still not convinced that the payoff here is worth the additional complexity.

First, I've typed up some notes as I've read the proposal, which should hopefully paint a picture for why I'm somewhat skeptical of the overall idea.

> 2. Use the `alias` deriving strategy.

I'm very hesitant to call `alias` a "deriving strategy" in the same vein as existing deriving strategies. `alias` is rather peculiar in that it is only well defined when used in the context of `deriving` clauses. What about `StandaloneDeriving`, however? The meaning of something like this:

```haskell
deriving alias FAM = newtype (Functor, Applicative, Monad)

newtype T f a = MkT (f a)
deriving alias instance FAM (T f)
```

Is rather unclear. Does this mean that you should derive a `Functor`, `Applicative`, _and_ a `Monad` instance? That would be breaking new ground since currently, you can only derive one instance per standalone `deriving` declaration. Moreover, what would the instance context for these instances be? Presumably, you'd want a different context for each of the `Functor`, `Applicative`, and `Monad` instances, but it's far from obvious to me how you would specify that syntactically.

> The syntax of declaring a new deriving alias is the following:
>
> 1. `deriving` keyword.
> 2. Followed by the `alias` keyword.
> 3. Followed by the name of the alias.
> 4. Followed by zero or more type variables.
> 5. Followed by the `=` operator.
> 6. Followed by the list of deriving definitions similar to what is used with data types but without the `deriving` prefix.

I'm left with several questions after being presented this syntax.

1. The following is a perfectly valid use of `deriving`:

   ```haskell
   data T f a = MkT (f a)
     deriving Functor
     deriving Show
   ```

   Suppose I wanted to group those two `deriving` clauses into a single `alias`. How would I do so? Judging from (6), it would be something like this:

   ```haskell
   deriving alias FunctorShow = Functor Show
   ```

   But this is rather confusing: `Functor Show` looks like a single, ill kinded type! We surely don't want to permit this sort of thing in the syntax for `alias`.

   My suggestion would be to break up each cluster of derived classes with some sort of delimiter. One possibility is to use the `deriving` keyword for this purpose:

   ```haskell
   deriving alias FunctorShow = deriving Functor deriving Show
   ```

   I'm sure there are other delimiters one could choose as well.

   Another possibility would be to simply require prefixing each cluster of derived classes with an explicit deriving strategy so as to always ensure the existence of a syntactic delimeter that separates each cluster. (In other words, require `deriving alias FunctorShow = stock Functor stock Show`.) This would mean that translating uses of `deriving` clauses to `alias`es might require some thought, but perhaps that is a good thing.
2. Since `alias`es can be parameterized by type variables, does this mean that `alias`es have types? For instance, in the following example:

   ```haskell
   deriving alias MReader env =
     (MonadReader env)
   ```

   Is it the case that `MReader :: Type -> Alias`? (Where `Alias` is some abstract data type?) If so, should users be able to write this type signature themselves at the top level?

   ```haskell
   deriving alias MReader :: Type -> Alias
   deriving alias MReader env =
     (MonadReader env)
   ```
3. Can there be higher-order `alias`es? For instance, is the following legal?

   ```haskell
   deriving alias Product alias1 alias2 =
     alias alias1
     alias alias2
   ```

   Where `Product :: Alias -> Alias -> Alias`?
4. How far can you go with this? Should the following be legal?

   ```haskell
   deriving alias Bot = alias Bot
   ```

   Any use of `Bot` in a `deriving` clause would surely fail to terminate. (At compile-time!)
5. Since the following is legal:

   ```haskell
   newtype T f a = MkT (f a)
     deriving newtype (Functor, Applicative, Monad, MonadReader env)
   ```

   Presumably, the following is also legal?

   ```haskell
   deriving alias FAMReader =
     newtype (Functor, Applicative, Monad, MonadReader env)
   ```

   Notice that `env` is not bound by `FAMReader`, but is instead implicitly quantified in the `deriving` clause. That is to say, just as `deriving` clauses have multiple binding sites for type variables in today's GHC, so too can `alias`es have multiple binding sites. It would be good to clarify this wrinkle in the proposal.

> It is possible to define a [TH macro](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#template-haskell) that uses the [StandaloneDeriving](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#stand-alone-deriving-declarations) extension to derive a lot of typeclasses for you. Like this:
>
> ```haskell
> boilerplateAnnigilator :: Name -> Q [Dec]
> boilerplateAnnigilator typeName = do
>     let typeCon = conT typeName
>     [d|deriving instance Show $(typeCon)
>        deriving instance Eq   $(typeCon)
>        deriving instance Ord  $(typeCon)
>       |]
>
> This approach [was proposed here](https://stackoverflow.com/questions/45113205/is-there-a-way-to-shorten-this-deriving-clause).
>
> It works okay for basic and simple cases. However, it's extremely difficult to implement generic logic that will work for every typeclass since standalone deriving requires you to specify the context for type variables. With this approach, one needs to reimplement internal GHC logic but in Template Haskell.

This section of the proposal mischaracterizes the power of a TH-based solution. In fact, you _aren't_ required to specify any context at all! You can take the approach that the [`derive-topdown`](http://hackage.haskell.org/package/derive-topdown) library takes and simply leave the contexts unspecified:

```haskell

boilerplateAnnihilator :: Name -> Q [Dec]
boilerplateAnnihilator typeName = do
    let typeCon = conT typeName
    [d|deriving _ => instance Show $(typeCon)
       deriving _ => instance Eq   $(typeCon)
       deriving _ => instance Ord  $(typeCon)
      |]
```

That's it! Now GHC will figure out the contexts for you when you splice in a use of `boilerplateAnnihilator`. Of course, this requires you to enable the `PartialTypeSignatures` extension, which is a bit funny, but there is a [separate GHC proposal](https://github.com/ghc-proposals/ghc-proposals/pull/194) underway to clean that up.

(When I first opened this proposal, my first thought was to wonder why you simply didn't use this solution.)

> [Quite recently](https://www.reddit.com/r/haskell/comments/9dx6s9/proposal_data_deriving_synonyms/) it was discussed to reuse existing type synonyms of kind Constraint, like in the example below:
>
> ```haskell
> type RS a = (Read a, Show a)
> ```
>
> ```haskell
> data Foo = Bar | Baz
>     deriving RS
> ```
>
> However, this approach is quite limited:
>
> 1. It doesn't allow one to specify a deriving strategy.
> 2. It is not possible to integrate [DerivingVia](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#deriving-via) with this approach.
> 3. It becomes cumbersome to use when multiple type variables are involved.

I'm glad you mentioned this here for the sake of completeness, since it's an approach that I think is fundamentally flawed. I would go even further and point out an even more glaring weakness of this approach:

4. It's extremely unclear on what the typing rules are. For instance, you can write this today:

   ```haskell
   data T f a = MkT (f a)
     deriving Functor
     deriving Show
   ```

   But can we actually abstract that out into a type synonym? A first attempt would be something like `type FunctorShow = (Functor, Show)`, but surely that can't be well kinded. Perhaps then it should be `type FunctorShow f a = (Functor f a, Show a)`, but then it becomes unclear on how to instantiate it:

   ```haskell
   data T f a = MkT (f a)
     -- deriving FunctorShow ??
     -- deriving (FunctorShow f) ??
     -- deriving (FunctorShow f a) ??
   ```

-----

Addressing comments that others have made on this proposal:

1. I concur with @Fresheyeball that some thought needs to be put into specifying what happens when you derive with multiple `alias`es that contain overlapping classes. I don't have a strong opinion on what the desired behavior should be, but regardless of what is decided, it should be mentioned in the proposal.
2. @int-index's proposal is intriguing, although I agree that it ought to be considered separately from this proposal, since his proposal has its own set of design questions (which I'll omit here, since this reply has already become far too long-winded).

theindigamer:
For the CPP section:

> However, abusing CPP macro-system for this feature would likely be a poor solution. Furthermore, it is a well-known fact that tooling usually handles CPP macros in a bad way. Since GHC is fed with the code after preprocessing, it's not possible to perform a huge bunch of analysis forms for macros and their usage.

1. Could you explain why CPP is a poor solution for this? Are tooling and analysis the two key problems with CPP, or is there something else (because you've written 'furthermore')?
2. W.r.t. tooling, I know brittany barfs in the presence of CPP. What other tools do you have in mind here?
3. What does "huge bunch of analysis forms for macros and their usage" mean here? What kind of analysis do you have in mind here?

I'm not a fan of CPP either, but IMHO, it does seem like the right tool for the job here. The fact that it doesn't work with tools like brittany is a bit unfortunate though :frowning_face:.

goldfirere:
Thanks, @RyanGlScott, for such an in-depth analysis!

I wonder if I might propose another alternative that I think would be considerably simpler: allow Template Haskell splices in `deriving` clauses.

Currently, we allow TH splices in precisely 4 contexts: expressions (`let foo = $(splice) in foo x y`), types (`foo :: Int -> Bool -> $(resultType 3)`), declarations (`$(genDecls 42)` at top-level), and patterns (`foo $(mkPat "a" "b") = a + b`). It would be in keeping with the spirit and implementation of TH to allow another context: the `deriving` clause of a datatype declaration: `data X = Y | Z deriving (Show, Read) deriving $usefulAlias`. The type of a deriving-splice would be `Q [DerivClause]` (allowing a list so that multiple strategies could be used); a deriving-splice could be used only after the word `deriving` (i.e. not in a larger list) and could be used only without an explicit strategy. We could also allow deriving-quotes, though that is merely a convenience, not an increase in expressiveness.

Would this alternative achieve your goals? I admit that TH is never as clean as a feature that doesn't use TH, but the design issues above seem to be sidestepped somewhat by the TH approach. (We might want to stipulate that redundant `deriving`s are ignored.)

Looking at this problem in an even larger way, I've always been a bit uncomfortable with the fact that `deriving` and TH are both mechanisms for automated code generation, yet they are entirely separate... I sometimes wonder if `deriving` could somehow usefully replaced wholesale with TH, but that's really a conversation for another day.

mpickering:
A big problem with the TH approach suggested in https://github.com/ghc-proposals/ghc-proposals/pull/215#issuecomment-476004380 is that the quotes will be untyped. 

There are many different forms of compile time evaluation possible with GHC but one involving the constraint solver as @int-index suggests in https://github.com/ghc-proposals/ghc-proposals/pull/215#issuecomment-475631466 is the most appropriate for this domain I think.

goldfirere:
@int-index's approach (https://github.com/ghc-proposals/ghc-proposals/pull/215#issuecomment-475631466) seems to suffer from the same untypedness as my TH approach (https://github.com/ghc-proposals/ghc-proposals/pull/215#issuecomment-476004380). Am I missing something here?

Of course, either approach could, in theory, be made more typed, if someone wants to describe the type system.

aspiwack:
A bigger problem with Template Haskell seems, to me, to be the ordering limitation that Template Haskell imposes on declaration. Could we even do mutually recursive types with derived definition this way?

RyanGlScott:
> Could we even do mutually recursive types with derived definition this way?

It could, at the expense of having to order things more carefully:

```haskell
boilerplateAnnihilator :: [Name] -> Q [Dec]
boilerplateAnnihilator typeNames = concat <$> mapM decs typeNames
  where
    decs typeName =
      let typeCon = conT typeName in
      [d|deriving instance _ => Show $(typeCon)
         deriving instance _ => Eq   $(typeCon)
         deriving instance _ => Ord  $(typeCon)
        |]

-- ...

data A = A1 | A2 B
data B = B1 | B2 A
$(boilerplateAnnihilator [''A, ''B])
```

ChShersh:
@RyanGlScott Thank you very much for such a detailed analysis of the proposal! I will try to answer some questions here.

Regarding the following use case:

> ```haskell
> deriving alias FAM = newtype (Functor, Applicative, Monad)
> 
> newtype T f a = MkT (f a)
> deriving alias instance FAM (T f)
> ```

I think in that case `alias` deriving strategy could be just a syntax sugar for the following code which exploits `PartialTypeSingatures` in the deriving context:

```haskell
newtype T f a = MkT (f a)

deriving newtype instance _ => Functor (T f)
deriving newtype instance _ => Applicative (T f)
deriving newtype instance _ => Monad (T f)
```

> Another possibility would be to simply require prefixing each cluster of derived classes with an explicit deriving strategy

I guess my description of the syntax wasn't precise enough. In case of something like this:

> ```haskell
> data T f a = MkT (f a)
>   deriving Functor
>   deriving Show
> ```

You need to define alias as follows:

```haskell
deriving alias FunctorShow = (Functor, Show)
```

Probably I can try to define precise grammar using `happy`-like syntax. But the idea was to keep close to the existing `deriving` syntax. Currently you are not required to specify deriving strategy. However, I can see reasons why some people always prefer to specify deriving strategy explicitly. I was thinking about enforcing use of the strategies in `deriving alias` but that looked too restrictive to me for a simple syntax-sugar extension.

> Since aliases can be parameterized by type variables, does this mean that aliases have types? 

In my proposal – no. `deriving alias` is just a syntax sugar construction. Type variables are treated merely as substitution arguments.

> Can there be higher-order aliases? For instance, is the following legal?

I haven't thought about that. And it doesn't seem easy patch my proposal to support higher-order aliases. I can see how inability to define higher-order deriving aliases can become annoying pretty-quickly. My proposal claims to solve boilerplate-related issues of the `deriving` mechanism, but it seems like this issue can't be completely solved without adding support for higher-order aliases. I think here it makes more sense to start thinking about some way to support first-class encoding for deriving.

> You can take the approach that the derive-topdown library takes and simply leave the contexts unspecified:

Very nice library! That's a cool approach indeed. Ability to write such code via TH increases threshold for language extensions like `DerivingAlias`.

<hr>

@theindigamer 

> 1. Could you explain why CPP is a poor solution for this? Are tooling and analysis the two key problems with CPP, or is there something else (because you've written 'furthermore')?

GHC compiler takes Haskell source code after preprocessing stage. GHC knows almost nothing about CPP macros except the presence of `-XCPP` extension. This means that tools that use GHC to analyse Haskell source code (like `brittany` or `ghc-exactprint` or `smuggler`) are doomed to never be able to look through CPP unless they implement their own version of the Haskell compiler. This can be beneficial for authors of IDE because you might want to have context-sensitive autocompletion or error-reporting. Reusing GHC for such situations makes more sense to me at this stage.

> I'm not a fan of CPP either, but IMHO, it does seem like the right tool for the job here.

You can't export CPP macros from the Haskell module where it is defined. If you want to have reusable deriving clause, you need to put it into separate C header file and include this file into your project. This is simply just not convenient.

<hr>

At the end I want to clarify that my goal is to not push `DerivingAlias` to GHC no matter what but instead have a discussion about possible restrictions that stop GHC from having this feature. The question about having reusable deriving clauses pops up from time to time. But before this ghc-proposal there was no a single place where you see detailed reasons why such feature doesn't exist yet.

RyanGlScott:
> Regarding the following use case:
> 
> > ```haskell
> > deriving alias FAM = newtype (Functor, Applicative, Monad)
> > 
> > newtype T f a = MkT (f a)
> > deriving alias instance FAM (T f)
> > ```
> 
> I think in that case `alias` deriving strategy could be just a syntax sugar for the following code which exploits `PartialTypeSingatures` in the deriving context:
> 
> ```haskell
> newtype T f a = MkT (f a)
> 
> deriving newtype instance _ => Functor (T f)
> deriving newtype instance _ => Applicative (T f)
> deriving newtype instance _ => Monad (T f)
> ```

This is certainly a valid interpretation, but it's worth noting that this interpretation vastly restricts the power of `StandaloneDeriving`, which gives you full power to specify instance contexts. This can be useful when:

1. One wishes to use an instance context that is stronger than what a `deriving` clause would infer.
2. One wishes to manually specify a context that GHC would otherwise refuse to infer, such as in the following example:

   ```haskell
   type family F a
   data T a = MkT (F a)
   deriving instance Show (F a) => Show (T a)
   ```

Neither of these scenarios are possible with standalone-derived instances that use `alias`, as far as I can tell.

> I guess my description of the syntax wasn't precise enough. In case of something like this:
> 
> > ```haskell
> > data T f a = MkT (f a)
> >   deriving Functor
> >   deriving Show
> > ```
> 
> You need to define alias as follows:
> 
> ```haskell
> deriving alias FunctorShow = (Functor, Show)
> ```

I ask this question because `deriving alias FunctorShow = Functor Show` is a natural consequence of interpreting the sixth item in your description of the syntax:

> 6. Followed by the list of deriving definitions similar to what is used with data types but without the `deriving` prefix.

If you take `deriving Functor deriving Show` and remove the `deriving` keywords, then `Functor Show` is what you're left with. If that wasn't your intention, then it would be good to change the description of the syntax to clarify this point.

> In my proposal – no. `deriving alias` is just a syntax sugar construction. Type variables are treated merely as substitution arguments.

If `deriving alias` declarations don't have types (and, as a consequence, aren't typechecked), then it seems like it would be quite easy to construct nonsensical `alias`es like the following:

```haskell
deriving alias E env = (MonadReader env, Show env)
```

Every attempt to use the `E` alias would result in a type error, since there is no possible way for `deriving (Show env)` to be well typed. That might be a tradeoff that you're willing to make, but if so, it would be good to make a note of this drawback in the proposal. 

> I haven't thought about that. And it doesn't seem easy patch my proposal to support higher-order aliases. I can see how inability to define higher-order deriving aliases can become annoying pretty-quickly.

To be fair, I think it's quite a reasonable position to say "we simply don't support higher-order aliases". This is a position that GHC has adopted for other language features, such as type synonyms and type families (at least, at present). Again, it would be good to make a note of this in the proposal, since I'm sure there are others are wondering the same thing.

nomeata:
@chshersh in light of the comments here, do you want to further pursue this proposal?

chshersh:
@nomeata Thanks for asking! I don't have solutions to all problems raised here, and at this moment I can't continue my work on the proposal. However, I don't mind if somebody else will take it from here :slightly_smiling_face:

nomeata:
Thanks for the update. Allow me to close this, just for tidyness; anyone is welcome to reopen and restart this.

chshersh:
Maybe it's too late, but I just thought of an alternative solution that requires to introduce less syntactic constructs. Instead of writing `alias` as a separate entity, why not just copy all derived instances for some existing `newtype`? So proposed here syntax:

```haskell
deriving alias CoreClasses =
    (Show, Generic)
    newtype  (Eq, Ord, Hashable, FromField, ToField, ToMustache)
    anyclass (FromJSON, ToJSON)

newtype Email = Email { unEmail :: Text }
    deriving alias CoreClasses
```

will become:

```haskell
newtype CoreClasses a = CoreClasses a
    deriving (Show, Generic)
    deriving newtype  (Eq, Ord, Hashable, FromField, ToField, ToMustache)
    deriving anyclass (FromJSON, ToJSON)

newtype Email = Email { unEmail :: Text }
    deriving as CoreClasses Email
```

In that case `deriving as Foo Bar` becomes syntactic sugar for `deriving (... every instance for Foo ...) via Foo Bar`. Not sure if this is 100% correct, just an idea of simplification.

PR #217, opened 2019-04-01
Dynamic typing

goldfirere:
This proposal describes a way we can opt out of Haskell's pesky type system, in a bid to increase both our expressiveness and overall popularity.

[Rendered](https://github.com/goldfirere/ghc-proposals/blob/dynamic/proposals/0000-dynamic-types.rst)

### DIFF
--- a/proposals/0000-dynamic-types.rst
+++ b/proposals/0000-dynamic-types.rst
@@ -132,3 +132,8 @@ Unresolved Questions
   perhaps we can get by simply by ignoring the input string once we have such an error up
   until the next line that begins at column 0. With this approach, if an infrequently-used
   function has a parsing error, the application will still build and be useful.
+
+Unacknowledgments
+-----------------
+
+This proposal bears striking similarity to content available `here <https://www.reddit.com/r/haskell/comments/b7p4xg/finally_dynamically_typed_programming_in_haskell/>`_, but was developed independently.
--- a/proposals/0000-dynamic-types.rst
+++ b/proposals/0000-dynamic-types.rst
@@ -1,0 +1,136 @@
+Dynamic Typing
+==============
+
+.. proposal-number:: Leave blank. This will be filled in when the proposal is
+                     accepted.
+.. trac-ticket:: Leave blank. This will eventually be filled with the Trac
+                 ticket number which will track the progress of the
+                 implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+Haskell is widely regarded as a leading strongly typed, lazy, functional programming language.
+Of course, we all know that it is also an excellent *imperative* programming language.
+With ``-XStrict``, we can consider it to be an excellent *strict* programming language.
+This proposal completes the cube by describing how Haskell can be an excellent *dynamic* programming
+language by allowing users to opt out of our pesky type system.
+
+Motivation
+------------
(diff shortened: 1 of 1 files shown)


nomeata:
Can we finally have a decent JavaScript FFI story in GHCJS now?

gbaz:
> Perhaps we want -XDynamicTypes to also defer scoping errors.

That sounds like a terrible idea! Scope errors are very important, and it goes against the spirit of this proposal to defer them instead of handling them properly. We should have a complimentary extension, `-XDynamicScope`.

Ericson2314:
Can `TYPE any` choose a register at run time? I think we could avoid GC then.

gary12377:
lgtm 👍

goldfirere:
Dynamic kinding works today, though it's inconvenient:

```
{-# LANGUAGE ExplicitForAll, PolyKinds, DataKinds #-}

module DynamicKind where

import Prelude hiding ( ($), (.) )
import GHC.Exts

class Rep r where
  ($) :: forall (r2 :: RuntimeRep) (a :: TYPE r) (b :: TYPE r2). (a -> b) -> a -> b
  (.) :: forall (r3 :: RuntimeRep) (a :: TYPE r) (b :: TYPE r) (c :: TYPE r3).
         (b -> c) -> (a -> b) -> a -> c

instance Rep LiftedRep where
  f $ a = f a
  (f . g) x = f (g x)

instance Rep UnliftedRep where
  f $ a = f a
  (f . g) x = f (g x)

instance Rep IntRep where
  f $ a = f a
  (f . g) x = f (g x)

instance Rep DoubleRep where
  f $ a = f a
  (f . g) x = f (g x)

twice :: forall (a :: TYPE r). Rep r => (a -> a) -> a -> a
twice f = (f $) . (f $)
```

It would take an extension of this proposal to handle such things more idiomatically. The (almost) above is because both kinds in the type of `.` are the same, but this could probably be worked out with more sweat.

treeowl:
I think correctable scoping errors should produce runtime warnings. I also suspect there are a lot of other potential applications for runtime instance lookup; that idea might be worth fleshing out in a paper.

nomeata:
> I think correctable scoping errors should produce runtime warnings

The program should just pause and ask the user for what to do. TeX has been doing that for years!

mstksg:
This proposal is very well thought-out from a design and engineering standpoint, and I'd be (truly) happy to see this in GHC.  It's a good demonstration of how sound everything is even from a dynamically typed standpoint.

gbaz:
Actually, on consideration I tried to derive a segfault and couldn't. So I think that this doesn't really give the expressive power I'm looking for in a dynamic typing extension.

int-index:
What about dynamic pattern matching? The proposal must be extended to handle this code:

```
f 5 = True
f "hello" = True
f _ = False
```

goldfirere:
> It's a good demonstration of how sound everything is even from a dynamically typed standpoint.

It indeed scared me to think how easy it would be to implement this. And I actually think it would work swimmingly, and would probably beat the performance pants off other dynamically typed languages, especially if we added a few `RULES` that eliminated redundant `wrapDyn`/`fromDynamic` pairs. This would mean GHC's type checker would figure out which dynamic checks could be performed statically.

> Actually, on consideration I tried to derive a segfault and couldn't.

Good point -- but we already have this feature: just type `unsafeCoerce True False`.

> What about dynamic pattern matching? 

This is not spelled out in the proposal, but this case is implicitly covered. The function `f` would take a `Dynamic` and then do runtime checks in its patterns, induced by the type mismatch that would ensue.

-----------------------------

It's sad to say, but the sunlight of another day has diminished my enthusiasm for this proposal, despite the great response it received. I withdraw it. But do reopen if you feel otherwise. :)

gbaz:
> would probably beat the performance pants off other dynamically typed languages

I think the more serious lisp compilers do this sort of thing whenever possible already actually :-)

iamrecursion:
It's a touch terrifying just how feasible this proposal is.

PR #218, opened 2019-04-03
"ViaApplications" -- fold @_ @(Int via Product a)

Icelandjack:
**Note**: The idea is in its infancy, help me find problems with it.

`{-# Langage ViaApplications #-}` allows writing

```haskell
>> sort @(Int via Down Int) [1..5]
[5,4,3,2,1]

>> fold @(Int -> Int via Dual (Endo Int)) [(* 10), (+ 5)] 10
105

>> fold @(Int -> Int via Dual (Dual (Endo Int))) [(* 10), (+ 5)] 10
150
```

and defining 

```haskell
sum, product :: forall a. Num a => [a] -> a
sum     = fold @_ @(a via Sum     a)
product = fold @_ @(a via Product a)

any, all :: (a -> Bool) -> ([a] -> Bool)
any = foldMap @_ @(Bool via Any)
all = foldMap @_ @(Bool via All)

foldl f z t = foldMap @​_ @(a -> a via Dual (Endo a)) (flip f) t z
bifoldMapDefault = bitraverse @t @(m via Const m)
```
instead of

```haskell
foldl f z t = appEndo (getDual (foldMap (Dual . Endo . flip f) t)) z

bifoldMapDefault :: forall t m a b. (Bitraversable t, Monoid m) => (a -> m) -> (b -> m) -> t a b -> m
bifoldMapDefault = coerce (bitraverse :: (a -> Const m ()) -> (b -> Const m ()) -> t a b -> Const m (t () ()))
```

The last definition also won't compile without a type signature, on the other hand the correct type of `bifoldMapDefault = bitraverse @t @(m via Const m)` is inferred

### DIFF
--- a/proposals/ViaApplications.rst
+++ b/proposals/ViaApplications.rst
@@ -1,0 +1,175 @@
+ViaApplications
+==============
+
+.. proposal-number:: 
+.. trac-ticket:: 
+.. implemented:: 
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+Allow ``@(<ty> via <viaty>)`` which instantiates ``@<viaty>`` (using ``<viaty>`` instances) and then coerces as if we had written ``@<ty>``
+
+
+Motivation
+------------
+
+Allow writing 
+
+::
+
+ foldMap @f @(Bool via Any) @a
+
+(handwave) where we instantiate its second type parameter to ``Any`` (``foldMap @f @Any @a``) but coerced to the type of ``foldMap @f @Bool @a``, somehow ignoring that ``Monoid Bool`` does not hold
+
+::
+
+ coerce $ foldMap @f @Any @a :: forall f a. Foldable f => (a -> Bool) -> (f a -> Bool)
+
+This would give us the following 
+
+::
+
+ sum :: forall a. Num a => [a] -> a
+ sum = fold @[] @(a via Sum a)
+
+ product :: forall a. Num a => [a] -> a
+ product = fold @[] @(a via Product a)
+
+ minimum :: foldl
+
+ any :: (a -> Bool) -> ([a] -> Bool)
+ any = foldMap @[] @(Bool via Any)
+
+ all :: (a -> Bool) -> ([a] -> Bool)
+ all = foldMap @[] @(Bool via All)
+
+And allows an alternative default ``foldl`` definition which makes use of composing types and their (``Monoid``) functionality
+
+
+like `Data.Foldable.foldl <https://hackage.haskell.org/package/base-4.12.0.0/docs/src/Data.Foldable.html#foldl/>`_:
+
+::
+
+ foldl :: (b -> a -> b) -> (b -> t a -> b)
+ foldl f z t = appEndo (getDual (foldMap (Dual . Endo . flip f) t)) z
+
+ -- vs
+
+ foldl :: (b -> a -> b) -> (b -> t a -> b)
+ foldl f z as = foldMap @(b->b via Dual (Endo b))
+   (flip f)
+   as
+   z
+
+::
+
+ >> mappend @(Int via Ap [] (Product Int)) [1,2,3] [1,11,111]
+ [1,11,111,2,22,222,3,33,333]
+
+`Data.Semigroup.Option <https://hackage.haskell.org/package/base-4.12.0.0/docs/src/Data.Semigroup.html#line-521/>`_:
+
+::
+
+ instance Semigroup a => Semigroup (Option a) where
+  (<>) = coerce ((<>) :: Maybe a -> Maybe a -> Maybe a)
+
+ -- vs
+
+ instance Semigroup a => Semigroup (Option a) where
+  (<>) = (<>) @(Option a via Maybe a)
+
+`Numeric.Natural.Natural <https://hackage.haskell.org/package/base-4.12.0.0/docs/src/GHC.Enum.html#line-968/>`_:
+
+::
+
+ instance Enum Natural where
+  enumFromThen x y
+    | x <= y    = coerce (enumFromThen :: Integer -> Integer -> [Integer]) x y
+    | otherwise = enumFromThenTo x y (wordToNaturalBase 0##)
+
+  enumFrom       = coerce (enumFrom     :: Integer -> [Integer])
+  enumFromTo     = coerce (enumFromTo   :: Integer -> Integer -> [Integer])
+  enumFromThenTo = coerce (enumFromThenTo :: Integer -> Integer -> Integer -> [Integer])
+
+ -- vs
+
+ instance Enum Natural where
+  enumFromThen x y
+    | x <= y    = enumFromThen @(Integer via Natural) x y
+    | otherwise = enumFromThenTo x y (wordToNaturalBase 0##)
+
+  enumFrom       = enumFrom @(Integer via Natural)
+  enumFromTo     = enumFromTo     @(Integer via Natural)
+  enumFromThenTo = enumFromThenTo @(Integer via Natural)
+
+`Data.Bitraversable.bifoldMapDefault <https://hackage.haskell.org/package/base-4.12.0.0/docs/src/Data.Bitraversable.html#bimapDefault/>`_:
+
+::
+
+ bifoldMapDefault :: forall t m a b. (Bitraversable t, Monoid m) => (a -> m) -> (b -> m) -> t a b -> m
+ bifoldMapDefault = coerce
+   (bitraverse :: (a -> Const m ())
+               -> (b -> Const m ()) -> t a b -> Const m (t () ()))
+
+ -- vs
+
+ bifoldMapDefault :: forall t m a b . (Bitraversable t, Monoid m) => (a -> m) -> (b -> m) -> t a b -> m
+ bifoldMapDefault = bitraverse @(m via Const m)
+
+
+Proposed Change Specification
+-----------------------------
+Specify the change in precise, comprehensive yet concise language. Avoid words like should or could. Strive for a complete definition. Your specification may include,
+
+* allows ``@(<ty> via <viaTy>)`` which means coercing the term ``@<viaTy>`` to the type if it had been ``@<ty>``
+
+
+Effect and Interactions
+-----------------------
+Detail how the proposed change addresses the original problem raised in the motivation.
+
+Discuss possibly contentious interactions with existing language or compiler features. 
+
+
+Costs and Drawbacks
+-------------------
+Give an estimate on development and maintenance costs. List how this effects learnability of the language for novice users. Define and list any remaining drawbacks that cannot be resolved.
+
+
+Alternatives
+------------
+The alternative is writing a cumbersome ``coerce`` with explicit annotations (which is what libraries do)
+
+::
+
+ bifoldMapDefault :: forall t m a b. (Bitraversable t, Monoid m) => (a -> m) -> (b -> m) -> t a b -> m
+ bifoldMapDefault = coerce
+   (bitraverse :: (a -> Const m ())
+               -> (b -> Const m ()) -> t a b -> Const m (t () ()))
+
+
+Unresolved questions
+--------------------
+
+Unfortunately this doesn't work, 
+
+::
+
+ bimapDefault :: forall t a b c d. Bitraversable t => (a -> b) -> (c -> d) -> t a c -> t b d
+ bimapDefault = coerce
+   (bitraverse :: (a -> Identity b)
+               -> (c -> Identity d) -> t a c -> Identity (t b d))
+
+ -- vs
+
+ bimapDefault :: forall t a b c d. Bitraversable t => (a -> b) -> (c -> d) -> t a c -> t b d
+ bimapDefault = bitraverse @(? via Identity)
+
+
+Implementation Plan
+-------------------
+(Optional) If accepted who will implement the change? Which other ressources and prerequisites are required for implementation?


Icelandjack:
This replaces some use cases for [`ala`](http://hackage.haskell.org/package/lens-4.17/docs/Control-Lens-Wrapped.html#v:ala)

luqui:
Cool, I like this!

Icelandjack:
Thanks man, I would love to find more examples compositions; here is one I found on IRC

```haskell
>> (($ []) . appEndo . getConst . traverse (\x -> Const (Endo (x:)))) "hello"
"hello"
>> traverse @_ @(String -> String via Const (Endo String)) "hello" []
"hello"
```

gridaphobe:
Would this allow us to break the invariants guaranteed by global instance coherence? For example, would I be allowed to write the following?

```haskell
s :: Set Int
insert @(Int via Down Int) s 5
```

It seems that your suggestion is to desugar `ViaApplications` to applications of `coerce`, so presumably this would desugar to something like:

```haskell
coerce (insert @(Down Int)) s 5
```

I would hope that the work on roles and safe coercions would save us here, and indeed it does.

```haskell
ghci> coerce @(Down Int -> Set (Down Int) -> Set (Down Int)) @(Int -> Set Int -> Set Int) insert s 5
<interactive>:1:1: error:
    • Couldn't match type ‘Down Int’ with ‘Int’
        arising from a use of ‘coerce’
    • In the expression:
        coerce
          @(Down Int -> Set (Down Int) -> Set (Down Int))
          @(Int -> Set Int -> Set Int)
          insert
          s
          5
```

The proposal should address this potential issue, alongside other fleshing out (like specifying what the desugaring plan is :)

nomeata:
Interesting idea, worthwhile exploring!

(shameless plug) reminds me of https://www.reddit.com/r/haskell/comments/765ogm/multiple_type_class_instances_for_the_same_type/doc8enp/

Icelandjack:
Vote for names: [`-XApplyingVia`](https://twitter.com/fried_brice/status/1113593543275175936)

Icelandjack:
Vote for names: `-XViaApplications`

Icelandjack:
Given [`Fold`](https://gist.github.com/sjoerdvisscher/6155858) this allows us to specify the existential `Monoid` independently (apart from representation) from the type we want to provide

```haskell
data Fold :: Type -> Type -> Type where
 Fold :: Monoid mm
      => { tally     ::  a -> mm
         , summarize :: mm ->  b
         }
      -> Fold a b
```

```
>> :t Fold 
.. :: Monoid mm => (a -> mm) -> (mm -> b) -> Fold a b
```

We might want our `Monoid` to be `Endo _` (as in `fromRightFold`) or `Dual (Endo _)` as in `fromLeftFold`

```
>> :t Fold @(Endo _)
.. :: (a -> Endo _) -> (Endo _ -> b) -> Fold a b
>> :t Fold @(Dual (Endo _))
.. :: (a -> Dual (Endo _)) -> (Dual (Endo _) -> b) -> Fold a b
```

```
>> :t \@x -> Fold @(x->x via Endo x)
.. :: (a -> (x -> x)) -> ((x -> x) -> b) -> Fold a b
>> :t \@x -> Fold @(x->x via Dual (Endo x))
.. :: (a -> (x -> x)) -> ((x -> x) -> b) -> Fold a b
```

iamrecursion:
I like this idea, and think it's definitely worth exploring further here! The fact that `via` is starting to become a way of expressing relationships that can't otherwise be inferred is quite cool.

Icelandjack:
@iamrecursion Thank you! It is exciting

@gridaphobe This is based on `coerce` like `-XDerivingVia` and inherits safety guarantees from the role system so it shouldn't introduce any issues (as demonstrated, `Set`'s nominal argument cannot be `coerce`d)

It works with `Data.List.insert` as `[a]` is representational in `a`

```haskell
>> insert @(Int via Down Int) 0 [1..4]
[1,2,3,4,0]
>>
>> (coerce $ insert @(Down Int) :: Int -> [Int] -> [Int]) 0 [1..4]
[1,2,3,4,0]
```
funky :zipper_mouth_face: 

@nomeata Thanks, it's all built on your work on `Coercible`! Biggest part of `via` is informing `Coercible` what parameters should be left alone. That's exactly what your "shameless plug" `*With` functions do manually, they match up the type variables but it requires a lot of annotations on "both ends" of `coerce @((c -> a) -> ([c] -> a)) @((c -> b) -> [c] -> b)` when the only change is `a` to `b`. This aims to capture that with `@(b via a)`

Jashweii:
So the idea is that
```Haskell
fun :: forall a. F(a)
fun @(X via Y)
```
is instantiated to F(Y) and then coerced to F(X), correct?
What does this do if some constraints depend on and aren't satisfied by Y? Would they remain as constraints on Y?
```Haskell
example :: forall s m . (MonadState s m, IsList s, Ord s) => ...
example' = example @(Int via Down Int)
```
should example' be 
```Haskell
forall m . (MonadState (Down Int) m, IsList (Down Int), Ord (Down Int)) => ...
```
(Ord (Down Int) is solved, of course)

DanBurton:
I'm a little uneasy with it.

Deriving via introduced codegen in a context where we were already doing codegen (the deriving mechanism). Applying via introduces codegen in a place where it was not found previously (type application).

I think instead of type coercions, what I actually want to see is something like explicit typeclass dictionary applications. So if the function calls for (Ord a) I want to be able to say "sort @(Ord a via Down a)". (Insert much handwaving here about how to tell which "a" is intended if there are multiple Ord constraints). I am more able to wrap my head around what this does, because it's just choosing a different typeclass dictionary to pass to the function.

Of course this is a Bad Idea to do with Data.Set functions. Perhaps that detail should be addressed in this proposal as well?

Icelandjack:
Why do you call this codegen @DanBurton?

goldfirere:
Despite its being extraordinarily useful, I've never loved deriving-via, essentially for the reasons @DanBurton describes. The key trick in deriving-via is to use newtypes to stand in for named instances. Clever use of `coerce` makes this work. And work it does. However, I find it indirect.

A case in point is a recent usage of deriving-via: A class I wanted instances of didn't have the right generic default for my purposes. So I wrote a newtype, an instance for that newtype, and then used a lot of deriving-via. But I'd much rather have done this without the newtype (which wasn't otherwise useful): if I could name the instance directly, then this would be simpler. One could argue that the newtype acts as the name of an instance, and that one would be right -- but again, let's not have play-acting: let's just name the instance.

This proposal (which I think holds water and may prove to be just as useful in practice as deriving-via) expands our use of newtypes as proxies for named instances. As an alternative, I'd love to consider [Coherent Explicit Dictionary Application for Haskell](https://dl.acm.org/citation.cfm?id=3242752) by Thomas Winant and Dominique Devriese. I liked the presentation at last year's Haskell Symposium, but I haven't studied this in detail. Would it lead to a cleaner way to do all this?

DanBurton:
I call it "codegen" because it is basically a macro to insert "coerce" at certain points in the code. Coerce might have 0 runtime cost, but it is still "code" with a semantic meaning.

Reflecting on this, I should state one further cause for unease, which is that type applications have historically served only to "narrow" the type of a function call. Erasing the type application will either a) be inferred to the same type, or b) be too ambiguous to infer. ApplyingVia allows type applications to _change_ the type of a function call, so that erasing the type application c) _changes_ the behavior of the code.

It may be worthwhile to do, but it is something that type applications have not been able to do before. If turning on this extension means that type applications cannot be "safely erased", then this is worth noting. (Define "safely erased" to mean "either a) the code runs the same, or b) it fails to compile"; the latter being a more tedious case to deal with, but still "safe" in the sense that you don't end up with different runtime code.)

DanBurton:
> I do think this (type-based approach) is the right point in the design space since instance resolution is type-directed but I need a more precise description of what you have in mind before I say more

I don't have a precise definition in mind, I just have a few reference points.

* Something comparable to Scala's explicit typeclass applications. (Not saying this is an ideal point in the design space, but it's worth discussing.)
* In GHC core, instance dicts are just named values and can be referred to by name
* As @goldfirere said, "let's just name the instance". I'm imagining things like
  * `deriving instance named SumInt as Monoid Int via Sum Int`
  * `instance named Blah as Monoid Int where ...`
  * `mconcat @{SumInt}` or some other unique new syntax for instance applications. This is probably the point you want to debate but I don't yet have a clear idea of what this should be. I apologize for not giving you something concrete to debate against.

So what I'm saying is, if you're going to make a case for ApplyingVia, then there should be some attempt at comparing/contrasting the other points in the design space for similar features. In this case, that would be Coherent Explicit Dictionary Application in Haskell, as mentioned by @goldfirere (which I have not yet reviewed but it sounds like exactly what I'm picturing).

I'd also love to see a more explicit comparison/contrast to `ala` & similar term-level tools, which are already mentioned briefly in the proposal. ApplyingVia is certainly more flexible, in theory.

AntC2:
foldMap @f @(Bool via Any) @a
    ...
    fold @[] @(a via Sum a)

This is presumably part of the continuing campaign to make Haskell indistinguishable from line noise. Please stop with the hier@glyphics already.

What's become of the succinct yet readable, elegant language by which I came to love functional programming? 

Is this intended merely as a shorthand for something I could write today? From the proposal's "(handwave)" I'm not seeing what; so I can't see what all those `@`s are intended to represent, nor if there's much saving of code.

Is there any benefit for users of packages as opposed to writers of packages? Will users get faced with these hier@glyphics (perhaps in error messages)? I'm experiencing that GHC today with all its extensions is just too big of a language to be usable. The language is in danger of disappearing up its categorial fundament.

Icelandjack:
I'll make time to address comments about *names* and what that entails later.

@AntC2: Thanks for sharing your opinions. The notation fits my view of Haskell, `@` existed long before and the often referenced "Coherent Explicit Dictionary Application" proposes a similar notation `@{downOrd}`.

> Is this intended merely as a shorthand for something I could write today? 
Yes a shorthand for `coerce` like `-XDerivingVia`

Not limited to library authors. Doesn't appear in error messages.

klapaucius:
>  succinct ..., elegant language

```
(($ []) . appEndo . getConst . traverse (\x -> Const (Endo (x:)))) "hello"
```

Icelandjack:
I will describe Coherent Explicit Dictionary Application (CEDA from here on) and its approach. It's a good paper and these are my preliminary observations.

There is an obvious appeal of exposing dictionaries to users as CEDA does, a case-insensitive `nub` can be written with `nub @{ Eq.Dict { (==) = \str str' -> map toLower str == map toLower str' }`. To be clear `via` is more verbose and can't straight-forwardly match this example until we get dependent types.

What does CEDA entail?

We generate a `.. -> Type` per `.. -> Constraint`, the paper introduces single-method `Eq` and `Ord` classes which generate dictionary records suffixed with `.Dict`:

```haskell
newtype Eq.Dict a = Eq.Dict { (==) :: a -> a -> Bool }

data Ord.Dict a = Ord.Dict { parent1 :: Eq.Dict a, (<=) :: a -> a -> Bool }
```

fields named `parent{1,2,..}` store superclasses in the order they appear in the type declaration. Now order matters, so changing
```haskell
class (Num a, Ord a) => Real a ..
-- to
class (Ord a, Num a) => Real a ..
```
seems like a breaking change. The design of CEDA does not support associated types and `MINIMAL` pragmas and `default` methods are ignored, so the dictionary records of `Eq` and `Ord` from base require more than the minimal fields needed in the paper

```haskell
data Eq.Dict a = Eq.Dict { (==) :: a -> a -> Bool, (/=) :: a -> a -> Bool }

data Ord.Dict a = Ord.Dict { parent1 :: Eq.Dict a, compare :: a -> a -> Ordering, (<) :: a -> a -> Bool, (<=) :: a -> a -> Bool, (>) :: a -> a -> Bool, (>=) :: a -> a -> Bool, max :: a -> a -> a, min :: a -> a -> a }
```

CEDA recommends writing smart constructors, and case-insensitive `nub` becomes

```haskell
smartEq :: (a -> a -> Bool) -> Eq.Dict a
smartEq (===) = Eq.Dict { (==) = (===), (/=) = \a b -> not (a === b) }

nub @{ smartEq (\str str' -> map toLower str == map toLower str') }
``` 

Because sometimes we want to specify the default dictionary record they optionally propose a special constraint, whose instances are autogenerated by the compiler 

```haskell
class HasDict (c :: Constraint) where
  type DictOf c :: Type
  getDict :: c => DictOf c
```

so you don't have to write by hand

```haskell
Ord.Dict 
  { parent1 = Eq.Dict
    { (==) = (==)
    , (/=) = (/=)
    }
  , compare = compare
  , (<)     = (<)
  , (<=)    = (<=)
  , (>)     = (>)
  , (>=)    = (>=)
  , max     = max
  , min     = min
  }
```

If we have a function with *two* constraints of the same type class

```haskell
foo :: (Eq a, Eq b) => (a, b) -> (a, b) -> ..
```

how do we override only `Eq a` or only `Eq b`? Now the variable names in the function's type signature start to matter as CEDA introduces the notation `foo @{eq1 as Eq a} @{eq2 as Eq b}` to refer explicitly to these unordered constraints. Alpha-renaming the type of `foo` to `(Eq b, Eq a) => (b, a) -> (b, a) -> ..` is now a breaking change. 

That is a big proposal with many choices, it requires compiler generated dictionaries and optional compiler generated constraint where superclass order and alpha-equivalent types start to matter.

----

`via` isn't as exciting and takes from its surroundings (instances, `newtype`s and `coerce`). As a result `MINIMAL` pragmas, default methods and associated type classes all work as it does in normal Haskell (as it is) so we only need to specify `(==)` or `(/=)`.

Because it is *type*-directed rather than *dictionary*-directed we never need to disambiguate between constraints (we do need a type signature for `foo` just as CEDA which is interesting to allow `-XTypeApplications`)

```haskell
foo :: forall a b. (Eq a, Eq b) => (a, b) -> (a, b) -> ..
```

since type variables are already ordered: `foo @(x via Eq1 x) @(y via Eq2 y)` (unlike CEDA the order is not dependent on the name of the type variables `a`, `b` as they appear in the type signature of `foo`).

There are some other questions but I'm tired so I won't answer them right now! Idk not really a conclusion

goldfirere:
As I'm thinking about this more, I'm not sure what the exact specification of this would be. I thought we would treat `f @(T1 via T2)` as `coerce @(type of f @T2) @(type of f @T1) f`, but that's not quite right. For example, if we have `sort :: forall a. Ord a => [a] -> [a]`, then `sort @(Int via Down Int)` is like `coerce @(Ord [Down Int] => [Down Int] -> [Down Int]) @(Ord Int => [Int] -> [Int]) sort`, but that's ill-typed, because `Ord`'s role is nominal (like just about all classes).

So then we can say that `f @(T1 via T2)` is really `coerce @(maximally instantiated type of f @T2) @(maximally instantiated type of f @T1) f`, where maximal instantiation instantiates all Inferred arguments (like class constraints). This means that `sort @(Int via Down Int)` becomes `coerce @([Down Int] -> [Down Int]) @([Int -> Int]) sort`, which is probably what you want.

But what if more specified variables intervene? Then the maximal instantiation trick doesn't work. And what if there are class constraints that mention multiple variables? It's not quite clear to me. So I think we might need a full specification before going much further...

Icelandjack:
similarly we could write

```haskell
newtype Const a b = Const a

instance Functor (Const a) where
  fmap :: (b -> b') -> (Const a b -> Const a b')
  fmap _ (Const a) = Const a

instance Monoid a => Applicative (Const a) where
  pure :: b -> Const a b
  pure _ = Const mempty

  (<*>) :: Const a (b -> b') -> Const a b -> Const a b'
  Const a <*> Const a' = Const (a <> a') 
```

as 

```haskell
instance Functor (Const a via a) where
  fmap :: (b -> b') -> (a -> a)
  fmap _ = id

instance Monoid a => Applicative (Const a via a) where
  pure :: b -> a
  pure _ = mempty

  (<*>) :: a -> a -> a
  (<*>) = (<>) 
```

Icelandjack:
An observation

```haskell
nub     :: forall a.               Eq  a => [a] -> [a]
nubOrd  :: forall a.               Ord a => [a] -> [a] 
nubHash :: forall a. Hashable a => Eq  a => [a] -> [a]
```

CEDA is dictionary-based, we cannot change the underlying function to one with different constraints without updating the dictionary to agree. Now we are hoist by the necessity to construct and name dictionaries (maybe you want to vary them separately?)

```haskell
nub     @{eqCaseInsensitive}
nubOrd  @{ordCaseInsensitive}
nubHash @{hashCaseInsensitive} @{eqCaseInsensitive}
```

The type-focused `-XApplyingVia`:

```haskell
nub     @(String via CaseInsensitive)
nubOrd  @(String via CaseInsensitive)
nubHash @(String via CaseInsensitive)
```

Same thing with `liftM`, `liftA` and `fmap`

```haskell
liftM @{myMonad}
liftA @{parent1 myMonad}
fmap  @{parent1 (parent1 myMonad)}
```

compared to

```haskell
liftM @(m via My m)
liftA @(m via My m)
fmap  @(m via My m)
```

So when Richard says

> One could argue that the newtype acts as the name of an instance, and that one would be right -- but again, let's not have play-acting: let's just name the instance.

As a library author CEDA means I export one dictionary record for each instance?

```haskell
module CEDA (eqCaseInsensitive, ordCaseInsensitive, hashCaseInsensitive, ..)
```

While `-XApplyingVia` only exports `CaseInsensitive(..)`; so it doesn't seem the case that `CaseInsensitive` is the name of a *single* instance but rather a single tag that indexes into *many* instance names.

goldfirere:
I'm not so worried about the covariance between function name and dictionary name in CEDA. We could always have a way of abstracting over that, by (say) having a class that classifies dictionary types, so the correct dictionary could be selected in a type-directed fashion.

Icelandjack:
The show stopper for `-XApplyingVia`; while I can apply `traverse` at some monoid `m` via `Const m` to obtain `foldMap`

```haskell
   (\@m -> traverse @_ @(m via Const m))
:: Traversable t => Monoid m => (a -> m) -> (t a -> m)
```

What about via `Identity` to obtain `fmap`?

```haskell
   traverse @_ @(?? via Identity)
:: Traversable t => (a -> b) -> (t a -> t b)
```

If all uses of the Applicative functor `f` are fully saturated, we could maybe write

```haskell
traverse @_ @(forall x. x via Identity x)

type Id a = a
traverse @_ @(Id via Identity)

type family Id a where Id a = a
traverse @_ @(Id via Identity)

traverse @_ @((\x -> x) via Identity)
```

but I'm not happy with those designs :) just putting it out there

Icelandjack:
This cool implementation of a SAT solver https://julesh.com/2016/09/22/abusing-the-continuation-monad/, the idea to replace `\k -> k (k True)` with [`\k -> k True || k False`](https://twitter.com/sjoerd_visscher/status/1040895496255156224) came from Sjoerd 

```haskell
sat :: Int -> ([Bool] -> Bool) -> Bool
sat n = runCont $ replicateM n $ cont $ \k -> k True || k False
```

Here it is in longer form with no `runCont` and `cont`

```haskell
sat n = replicateM @(forall x. (x -> Bool) -> Bool via Cont Bool x) n $ \k -> k True || k False

type Cont_ res a = ((a -> res) -> res)
sat n = replicateM @(Cont_ Bool via Cont Bool) n $ \k -> k True || k False
```

Icelandjack:
Keeping track of examples

```haskell
>> sequenceA @[] @[] ["ABC", "123", "!"]
["A1!","A2!","A3!","B1!","B2!","B3!","C1!","C2!","C3!"]

>> sequence @([] via ZipList) @[] ["ABC", "123", "!"]
["A1!","A2!","A3!","B1!","B2!","B3!","C1!","C2!","C3!"]

>> sequence @[] @([] via ZipList) ["ABC", "123", "!"]
["A1!"]

>> sequence @([] via ZipList) @([] via ZipList) ["ABC", "123", "!"]
["A1!"]
```

DanBurton:
sat n = replicateM @(forall x. (x -> Bool) -> Bool via Cont Bool x) n $ \k -> k True || k False

Is there a way to leverage inference so that this doesn't require such verbosity (or a type alias)? So that you can write something like:

    @(_ via Cont Bool)

What if the type application and the via application could be written side-by-side instead?

    @_@(via Cont Bool)

And then the type could be omitted

    @(via Cont Bool)

DanBurton:
One reason I'm inclined to nudge the proposal in this direction is that it may help distinguish Via Applications as a separate sort of thing, rather than being a new special case in the "type" space.

    instance Monoid a => Applicative (Const a)@(via a) where

Icelandjack:
@DanBurton 

> Is there a way to leverage inference so that this doesn't require such verbosity (or a type alias)?

The information we want to convey is `forall a r.` we are applying to `(a -> r) -> r` via `Cont r a`. Ignoring that `(a -> r) -> r` is not matchable in `a`, there is a place that gives exactly that information: The function `cont`

```haskell
cont :: forall a r. ((a -> r) -> r) -> Cont r a
```

The way [`ala` works is](https://hackage.haskell.org/package/newtype-0.2/docs/src/Control-Newtype.html)

> [..] the first parameter is _completely ignored_ on the value level, meaning the only reason you pass in the constructor is to provide type information.

We can do the same where `@(via f)` means applying to the domain of `f` _via_ the codomain of `f`, but discarding the term `f`

```haskell
replicateM @(Cont Bool)         :: Int -> Cont Bool a           -> Cont Bool [a]
replicateM @(via cont @_ @Bool) :: Int -> ((a -> Bool) -> Bool) -> (([a] -> Bool) -> Bool)
```

The `sat` example becomes (`@Bool` is inferred), I think this works for most of the examples and tends to be as short as I can reasonably expect

```haskell
sat :: Int -> ([Bool] -> Bool) -> Bool
sat n = replicateM @(via cont) n (\k -> k True || k False)
```


Similarly it could work for any "proxy" `f :: proxy ty viaTy`

Icelandjack:
Using the same value-level witness for instance declarations, it would probably be the other way round

```haskell
runContT :: ContT r m a -> ((a -> m r) -> m r)
```

note: the body of `fmap` doesn't mention `ContT` or `runContT`

```haskell
instance Functor (via runContT) where
 fmap :: (a -> b) -> (((a -> m r) -> m r) -> ((a' -> m r) -> m r))
 fmap = flip (.) . flip (.)
```

Icelandjack:
(proxy means [`@(via Coercion @ty @viaTy)`](http://hackage.haskell.org/package/base/docs/Data-Type-Coercion.html) works and [`@(via id)`](https://www.reddit.com/r/haskell/comments/4rxxpv/interesting_useful_neat_applications_of_id/d559j7n/) is a no-op)

Icelandjack:
```haskell
import Data.Semigroup renaming
 ( (+)    from (<>) @(via Sum)
 , (*)    from (<>) @(via Product)
 , (&&)   from (<>) @(via All)
 , (||)   from (<>) @(via Any)
 , union  from (<>) @(Set _)
 , union' from (<>) @(Map _ _)
 , const  from (<>) @(via First)
 , tsnoc  from (<>) @(via Last)
 , liftS  from (<>) @(Ap _ _)
 )
```

Icelandjack:
- [ ] Richard in [ghc-devs thread](https://ghc-devs.haskell.narkive.com/NXBBDXg1/suppressing-false-incomplete-pattern-matching-warnings-for-polymorphic-pattern-synonyms) discusses
    > So you're viewing complete sets as a type-directed property, where we can take a type and look up what complete sets of patterns of that type might be.
    > Then, when checking a pattern-match for completeness, we use the inferred type of the pattern, access its complete sets, and if these match up.

    I want to explore `-XMatchingVia`, maybe a type can be matched by any constructor of the (representationally) same type. How it relates to completeness and GADTs.   

I'm not very excited about `-XImportingVia` or what ever.

TheMatten:
May I ask, what's the current status of this proposal? Here and there, cases pop up in my code where `ApplyingVia` would make `newtype` transformations much more concise and readable.

tysonzero:
Personally this feels like an abuse of typeclasses.

Typeclasses are specifically designed around coherence and having a canonical instance, whereas regular value-level Haskell is designed around composing and building pieces in incredibly flexible ways.

This seems like trying to force typeclasses to do things they are not designed for, and I don't find it very readable or beginner friendly.

[Here](http://h2.jaguarpaw.co.uk/posts/simplifying-typeclasses/) is a blog post I wrote relating to this that I think is quite relevant to this proposal.

Icelandjack:
Can you turn your blog post into a proposal tysonzero

tysonzero:
@Icelandjack 

To be honest part of my proposal is just to ask the community to not overuse typeclasses, and try and build types first, and then add typeclasses when canonical/globally-coherent values of these types are needed. For example I am sad that `Monoid`/`Group`/`Ring` are much more available as classes than as types.

With that said, actually doing the above is more awkward then it needs to be, so I can definitely put together a proposal to go after the current ergonomic issues:

E.g.

```
class FooBarBaz a where
    foo :: Foo a
    bar :: Bar a
    baz :: Baz a
```

Is ergonomically nicer but less useful than:

```
data FooBarBaz_ a = FooBarBaz
    { foo_ :: Foo a
    , bar_ :: Bar a
    , baz_ :: Baz a
    }

class FooBarBaz a where
    fooBarBaz :: FooBarBaz a

foo :: FooBarBaz a => Foo a
foo = foo_ fooBarBaz

bar :: FooBarBaz a => Bar a
bar = bar_ fooBarBaz

baz :: FooBarBaz a => Baz a
baz = baz_ fooBarBaz
```

I would like the former to enable me to do everything I can do with the latter, just like the underlying core, although of course without any of the `_` namespacing hacks. I think `RecordDotSyntax` is somewhat of a prereq for getting particularly far with this.

I also have some more things to consider such as how to deal with superclasses, as they make it so the structure you can get out of an instance is larger than the structure you have to put in to make the instance in the first place, assuming the superclasses have already been defined.

Icelandjack:
Apart from the community changing did you look at Coherent Explicit Dictionary Application for Haskell? You both propose a dictionary solution, I think it's the wrong way but it sounds like it solves what you want

tysonzero:
I haven't read through it particularly thoroughly but it doesn't seem overly relevant. I am not looking for coherence or global uniqueness, I want different behavior based on what parameter you pass in, which is pretty antithetical to the former (and to typeclasses in general).

I just don't see the appeal of using something that is essentially a much less flexible and powerful version of passing in a parameter. Particularly when the syntax and desugaring is quite beginner unfriendly and magical.

There are a variety of useful things that you can do with a parameter that will be quite problematic when trying to use `ApplyingVia` instead. Some examples off the top of my head:

```
-- parameterized structures
foo :: (Foldable f, Integral a) => f a -> a
foo = fold_ (additiveModulo 7)
```

You could technically do the above using type level naturals, but this will not generalize to more complex parameters.

```
-- deferring the choice of structure
foo :: (Foldable f, Integral a) => a -> f a -> a
foo n = fold_ (additiveModulo n)
```

You could do the above using a parameter at the type level, but then you are in for some serious pain if you want to get that value based on information present at runtime.

```
-- use two different structures in a single function
foo :: Num a => Monoid_ a -> Monoid_ a -> a
foo x y = fold_ x + fold_ y
```

Not sure how you would go about using ApplyingVia in the above, as there is only one `a` to play with.

```
-- Make a structure out of two structures
mkSemiring :: Monoid_ a -> Monoid_ a -> Semiring_ a
mkSemiring = ...
```

This seems like it would be pretty painful, as `Sum` and `Product` and so on operate on just one underlying type and its instances. You'd need a newtype that takes in two type variables for each of the monoid instances, plus probably a type variable for the underlying type that is actually stored, it must also then require that all 3 of these are coercible to each other.

```
-- First class usage of these structures
myIntMonoids :: [Monoid_ Int]
myIntMonoids = [additive, multiplicative]
```

This is just a small example, but the general idea is that these instances are very much second class citizens. You can only do the hard coded things with them that various extensions say you can do. They aren't real values.

I enjoy higher order functions and algebraic structures and am very happy that they are a big part of Haskell, and all of this feels like a step in the opposite direction. I want to be able to talk about things like monoids and rings and serializers and monads as first class values. I then also want to seamlessly transition from these first class objects to and from instances, when the convenience and power of global coherence and canonicity are beneficial to me. I don't want to have to only talk about one single blessed instance per type, and then try to carefully manage a variety of extensions to coerce them from one to another.

Icelandjack:
> I just don't see the appeal of using something that is essentially a much less flexible and powerful version of passing in a parameter. 

You can't beat values in terms of easy of use, people like [Gabriel](http://www.haskellforall.com/2012/05/scrap-your-type-classes.html) have written about this. Why *not* use first-class values, I will try to answer these questions in a blog post but I think the choice is not obvious

Icelandjack:
> I just don't see the appeal of using something that is essentially a much less flexible and powerful version of passing in a parameter. 

You can't beat values in terms of easy of use, people like [Gabriel](http://www.haskellforall.com/2012/05/scrap-your-type-classes.html) have written about this. Why *not* use first-class values, I will try to answer these questions in a blog post but I think the choice is not obvious

Icelandjack:
Your examples are possible except the last one which doesn't have to do with the extension, Dependent Haskell will add a lot of power

tysonzero:
> Your examples are possible except the last one which doesn't have to do with the extension,

Can you show me what they would look like?

> Dependent Haskell will add a lot of power

I like dependent types, but this is absolutely not where I want to use them. `\n -> fold_ (additiveModulo n)` is simple stuff that shouldn't require dependent types.

LeanderK:
Just to give another perspective. I am really excited for the proposal and never ever wanted more value-level haskell via dictionary passing like @tysonzero is proposing, I just feel like it doesn't address the problems i have. I see applying-via as a great tool when a single blessed instance is not possible and we have the situation that once we have chosen the correct instance, we want to use it coherently in the function we're calling. Passing dictionaries is more flexible, but I (almost?) never want the full flexibility. Once i've committed, i only want to use the single instance.

Working with type-classes makes writing abstract code so much more enjoyable. I can say: I have an abstract thing with property 1,2,3 and here's what I want to do with it. It's really a joy to write. But type-classes show their pain when you have things that have multiple, equally important instances. All those new types makes code really hard to read and understand. For me, applying via solves a real problem.

Since I am doing some computational math with Haskell, the perfect example is the ring with two possible monoid-instances: One for addition, one for multiplication. I explicitly don't want dictionary passing, any additional flexibility can actively harm me, I still want to rely on coherency when writing abstract code. But I have two equally important instances when the thing i have is a ring.

For me, a blessed canonical instances works 95% of the time, but it sometimes happens that I have to juggle a lot of newtypes. This proposal would really reduce the pain and i really hope it moves forward!

tysonzero:
@LeanderK 

I'm not proposing getting rid of typeclasses or typeclass-based functions. I'm just saying that when you want that flexibility you should just accept a parameter, and whenever you want coherence instead then use typeclasses.

`ApplyingVia` is a weird hybrid where you are using a bunch of code generation to get something that feels like an incoherent instance to get custom local behavior, but that technically doesn't directly involve incoherence due to all the implicit coerces.

LeanderK:
@tysonzero while i may be not as clear as i wanted to be, i read your blogpost.

_ApplyingVia is a weird hybrid where you are using a bunch of code generation to get something that feels like an incoherent instance to get custom local behavior, but that technically doesn't directly involve incoherence due to all the implicit coerces._

i was trying to argue that this is in fact sometimes very useful with the, because it complements the usually typeclass-mechanism (with the ring-example).

masaeedu:
@Icelandjack

> Can you turn your blog post into a proposal tysonzero

I can't speak for @tysonzero but perhaps https://github.com/ghc-proposals/ghc-proposals/pull/324 is something like what you're looking for?

For whatever little it's worth I am strongly in the camp of avoiding further incremental patching of the deriving mechanism. There's an incredible wealth of use cases for "deriving" instances that cannot be served by the baked in notion of representational type equality available in the compiler. Sometimes you want to map an instance along an isomorphism, sometimes merely along a homomorphism, and very often these morphisms are not statically known (or at least it is inconvenient to shunt enough of the logic into the type system to reify them statically).

IMO the best and most beautiful features of any system are the ones that emerge spontaneously from the interaction/composition of simpler features. I think that `DerivingVia` and this proposal are good workarounds given the state of affairs of the typeclass feature today. But if we need to keep adding `DerivingThis`, `DerivingThat`, and `DerivingTheOther`, and each such feature shunts a little more of the user's logic into the type level, perhaps it is a good time to pause and reflect on whether a radically different approach is justified, at least before we proceed too far along this road.

LeanderK:
@Icelandjack I always come back thinking about this proposal. I personally really like it. 

I wonder....can this be implemented as an GHC-plugin? I am not exactly sure as it's a bit over my head...but I that it's possible.

PR #221, opened 2019-04-18
Process tweak: Involve authors in a possible rejection

nomeata:
Based on recent bad experiences, this is a proposal to tweak our process. The main idea is that if the shepherd intends to reject the proposal, they have to first discuss this with the authors, to give the authors a chance to respond, tweak, or point out misunderstandings in the shepherd’s perception. 

So the process of accepting proposals is unchanged, and still on the list, rejecting a proposal is now a two step process: the shepherd discusses it with the authors on GitHub (and anyone else interested, of course), and then there is a shorter discussion on the mailinglist, with hopefully all facts clarified.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -269,11 +269,11 @@ mailing list. All interested parties are invited to follow the discussion.
 -  Based on the proposal text (but not the GitHub commentary), the shepherd
    decides whether the proposal ought to be accepted or rejected or returned for
    revision.
-   
+
 -  If the shephard thinks the proposal ought to be rejected, they post their
    justifications on the GitHub thread, and invite the authors to respond with
    a rebuttal and/or refine the proposal. This continues until either
-    
+
    * the shepherd changes their mind and supports the proposal now,
    * the authors withdraw their proposal,
    * the authors indicate that they will revise the proposal to address the shepherds
@@ -283,19 +283,19 @@ mailing list. All interested parties are invited to follow the discussion.
      positions, even if they disagree on the conclusion.
 
 -  Now the shepherd proposes acceptance or rejectance, by email to the mailing
-   list. Discussion among the committee ensues on the mailing list, and 
+   list. Discussion among the committee ensues on the mailing list, and
    silence is understood as agreement with the shepherd's recommendation.
 
 -  Ideally, the committee reaches consensus, as determined by the secretary or
    the shepherd. If consensus is elusive, then we vote, with the Simons
    retaining veto power.
 
--  The decision is announced, by the shepherd to the mailing.
+-  The decision is announced, by the shepherd to the mailing list.
 
 -  The secretary tags the pull request accordingly, and either merges or closes it.
    In particular
-  
-   *  **If we say no:** 
+
+   *  **If we say no:**
       The pull request will be closed and labeled
       `Rejected <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3Arejected>`_.
 
@@ -306,9 +306,9 @@ mailing list. All interested parties are invited to follow the discussion.
       GHC, it will be reverted.
 
    *  **If we say yes:**
-       The pull request will be merged, numbered and labeled
+      The pull request will be merged, numbered and labeled
       `Accepted <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3AAccepted>`_.
-      
+
       At this point, the proposal process is technically
       complete. It is outside the purview of the committee to implement,
       oversee implementation, attract implementors, etc.
--- a/README.rst
+++ b/README.rst
@@ -271,8 +271,8 @@ mailing list. All interested parties are invited to follow the discussion.
    revision.
    
 -  If the shephard thinks the proposal ought to be rejected, they post their
-   justifications on the GitHub thread, and invites the authors to respond with
-   a rebuttal and/or refine the proposal. This continues until the either
+   justifications on the GitHub thread, and invite the authors to respond with
+   a rebuttal and/or refine the proposal. This continues until either
     
    * the shepherd changes their mind and supports the proposal now,
    * the authors withdraw their proposal,
@@ -284,7 +284,7 @@ mailing list. All interested parties are invited to follow the discussion.
 
 -  Now the shepherd proposes acceptance or rejectance, by email to the mailing
    list. Discussion among the committee ensues on the mailing list, and 
-   silence is understood as agreement with the shepherds recommendation.
+   silence is understood as agreement with the shepherd's recommendation.
 
 -  Ideally, the committee reaches consensus, as determined by the secretary or
    the shepherd. If consensus is elusive, then we vote, with the Simons
--- a/README.rst
+++ b/README.rst
@@ -267,23 +267,36 @@ mailing list. All interested parties are invited to follow the discussion.
    the discussion.
 
 -  Based on the proposal text (but not the GitHub commentary), the shepherd
-   makes a recommendation as to whether the proposal ought to be accepted,
-   rejected or returned for revision.
+   decides whether the proposal ought to be accepted or rejected or returned for
+   revision.
+   
+-  If the shephard thinks the proposal ought to be rejected, they post their
+   justifications on the GitHub thread, and invites the authors to respond with
+   a rebuttal and/or refine the proposal. This continues until the either
+    
+   * the shepherd changes their mind and supports the proposal now,
+   * the authors withdraw their proposal,
+   * the authors indicate that they will revise the proposal to address the shepherds
+     point. The shepherd will label the pull request as
+     `Needs Revision <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3A"Needs+revision">`_.
+   * the authors and the shepherd fully understand each other’s differing
+     positions, even if they disagree on the conclusion.
 
--  Discussion among the committee ensues on the mailing list.
-   Silence is understood as agreement with the shepherds recommendation.
+-  Now the shepherd proposes acceptance or rejectance, by email to the mailing
+   list. Discussion among the committee ensues on the mailing list, and 
+   silence is understood as agreement with the shepherds recommendation.
 
 -  Ideally, the committee reaches consensus, as determined by the secretary or
-   the shepherd.  If consensus is elusive, then we vote, with the Simons
+   the shepherd. If consensus is elusive, then we vote, with the Simons
    retaining veto power.
 
--  The decision is announced, by the shepherd or the secretary, to the mailing
-   list and the pull request commentary. In particular:
+-  The decision is announced, by the shepherd to the mailing.
 
-   *  **If we say no:** The shepherd comments on the GitHub pull request with the
-      reasons for rejection.
-      The pull request will be closed, by the shepherd or the secretary, with
-      label
+-  The secretary tags the pull request accordingly, and either merges or closes it.
+   In particular
+  
+   *  **If we say no:** 
+      The pull request will be closed and labeled
       `Rejected <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3Arejected>`_.
 
       If the proposer wants to revise and try again, the new proposal should
@@ -292,19 +305,10 @@ mailing list. All interested parties are invited to follow the discussion.
       In the case that the proposed change has already been implemented in
       GHC, it will be reverted.
 
-   *  **If we return for revision:** If during the debate, the need for
-      substantial changes does arise, then the shepherd summarizes the
-      issues on the GitHub pull request and labels it
-      `Needs Revision <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3A"Needs+revision">`_.
-      The author is encouraged to address the issues and re-submit.
-
    *  **If we say yes:**
-      The shepherd or the secretary announces this on the pull request
-      and labels it as
+       The pull request will be merged, numbered and labeled
       `Accepted <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3AAccepted>`_.
-      The secretary merges the pull request and assigns the final proposal
-      number.
-
+      
       At this point, the proposal process is technically
       complete. It is outside the purview of the committee to implement,
       oversee implementation, attract implementors, etc.


ndmitchell:
Might I suggest a Skype or similar? I find that much faster for resolving or clarifying differences opinion.

iamrecursion:
I think a 'face-to-face' discussion could really help, though as somebody who watches the proposals process I would like to see any salient points from said discussion clarified on the Pull Request in question.

gbaz:
I have no particular skin in the game here, but in my experience, scheduling a call at the same time between people on multiple continents, etc. seems rather difficult. It could be nice, if possible, but baking it into the process rather than something async seems rather onerous to all involved?

ndmitchell:
Certainly not a requirement, just a "offer it up if it works" kind of thing :)

int-index:
I can see why a Skype call might seem like a good idea, but there are also significant drawbacks:

1. A text-based discussion can be re-read in the future, a voice call leaves no trace unless recorded.
2. Spoken English is not the same skill as written English, non-native speakers might manage to write a proposal but not defend it during a realtime conversation.

nomeata:
Obviously, everybody is welcome to chat and talk to whoever they want (and ideally leave a note of the result in the github commentory). But I don’t think we need to enshrine that in the official process text.

README.rst
```diff
@@ -267,23 +267,36 @@ mailing list. All interested parties are invited to follow the discussion.
...
+   silence is understood as agreement with the shepherds recommendation.
 
 -  Ideally, the committee reaches consensus, as determined by the secretary or
-   the shepherd.  If consensus is elusive, then we vote, with the Simons
+   the shepherd. If consensus is elusive, then we vote, with the Simons
    retaining veto power.
 
--  The decision is announced, by the shepherd or the secretary, to the mailing
-   list and the pull request commentary. In particular:
+-  The decision is announced, by the shepherd to the mailing.
```
gridaphobe: mailing list.

README.rst
```diff
@@ -267,23 +267,36 @@ mailing list. All interested parties are invited to follow the discussion.
...
+    
+   * the shepherd changes their mind and supports the proposal now,
+   * the authors withdraw their proposal,
+   * the authors indicate that they will revise the proposal to address the shepherds
+     point. The shepherd will label the pull request as
+     `Needs Revision <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3A"Needs+revision">`_.
+   * the authors and the shepherd fully understand each other’s differing
+     positions, even if they disagree on the conclusion.
+
+-  Now the shepherd proposes acceptance or rejectance, by email to the mailing
```
gridaphobe: The committee discussion does sometimes uncover concerns the shepherd did not consider :) And in such a case, it would be unfortunate if the process turned into 

1. the shepherd raises concerns with the author,
2. the author addresses the concerns,
3. the committee raises other concerns,
4. the author addresses the new concerns,
5. the committee makes a decision.

I think we can collapse 1/2 and 3/4 into two steps: (1) the committee discusses and provides an initial response, (2) the author responds to concerns. In particular, we don't have to reach consensus to provide feedback, it seems perfectly reasonable for the initial response to be "the committee is divided on this proposal, specifically on the following concerns...".

AntC2:
To respond to a q on the Steering Committee discussion

> I wonder how many non-member are actively following this list.

I actively (if reading them is "active") follow those discussions. In general I think it's appropriate the great unwashed don't take part in them. I do find it frustrating occasionally when criticisms/suggestions for the proposal content get discussed there rather than on github where the authors/contributors can respond/correct misunderstandings.

I'd expect Committee discussion to focus not on elaborating the content of the proposal; but on its power-to-weight ratio (esp the insights about long-term costs of maintenance); and its 'fit' to other proposals and the future direction for GHC. (OK, how to keep that focus without discussing content? Do I contradict myself? Then I contradict myself.)

On that future direction: I think it would save proposers a great deal of wasted effort/annoyance if they could get a feel for where GHC is going; and therefore what proposals might align with that. You're always asking for people to contribute to GHC. Is there some kindly but firm way to tell people before they put effort into formally writing up a proposal, that it's a waste? The Committee process is not democracy, so getting upvotes or positive feedback on Github is no indicator of likely success. I dimly perceive that I am going in a different direction to GHC, but I don't understand why. (Perhaps those who understand the terrain better than me could suggest some other language I should concentrate on. I'm looking for an anonymous/extensible records system with well-principled typing that's not an embarrassment.)

There's still proposals open on the list dating back to the start of using Github. Are they effectively dead because the Committee will never accept them? And there's plenty more that got withdrawn without being submitted to the Committee, even after substantial discussion. Could we have avoided that waste?

GHC Haskell these days feels just too big and complex, full of extensions that are not-quite orthogonal. Is the current sprawl of type/kind features (with so many different ways to achieve the same thing) some sort of interim/stepping stone to the bright future? When will the cruft fall away and we get a compact, elegant language again?

If GHC is heading to a language that is essentially syntactic sugar over System FC, might it be easier to start a new language with that very narrow brief?

tejon:
> I actively (if reading them is "active") follow those discussions.

Likewise, for the data point.

README.rst
goldfirere: I don't quite see how @gridaphobe's tweak to the tweak leaves us in a different place than we are now -- except that we don't blast out the "reject" label so quickly and perhaps post back on GitHub sooner. Perhaps one way to avoid the unfortunate 5-step process is to have the shepherd post their concerns on both GitHub and the list in tandem, inviting other committee members to join in. I'm not sure we'll always get this right, though.

goldfirere:
> Is there some kindly but firm way to tell people before they put effort into formally writing up a proposal, that it's a waste?

This makes sense, in order to avoid a good deal of wasted effort. But I'm not sure it would work in practice, for several reasons:
 - The committee does not think as a unit. So to do this right, we would have to work to achieve a kind of early consensus on a proposal. But such a consensus is harder to achieve until some discussion has played out.
  - Proposals often go through a good deal of revision on their way to the committee. It would be poor form if we gave early discouragement to a proposal that could be revised into an acceptable state.
  - While I agree that the process isn't properly democratic, user opinion counts greatly. There have been several proposals I have been skeptical of, but my opinion has been swayed by a groundswell of support. Early discouragement would short-circuit this.

So perhaps the best way to get this feedback is to use the feedback gained from frequent posters as a quick barometer. It's imperfect, but I'm not sure another process would be better.

As to proposals that have languished, my belief is that these have been abandoned by their authors. They can be revived and submitted at any point.

iamrecursion:
To add another datapoint, I actively follow the process both here and on GitHub. 

_ara

README.rst
gridaphobe: The main goal of my tweak is to produce a faster feedback cycle. A few different people have raised the issue of delays on the committee's end, and I believe a decent chunk of that is due to stagnating debates within the committee. So my hope is that by collecting feedback from the committee as a whole and then allowing the author to respond, we can keep the discussions alive.

nomeata: I agree that my proposed tweak works well if the shepherd can foresee all complains, but less so when the committee discussion reveals flaws.

Having all discussions on Github would solve that.

Another take on it could be that the shephard acts like an advocate in both direction, representing the author towards the commttee, presenting and defending the proposal, but also representing the commitee towards the authors, relaying and discussing new concerns raised the by the committee as they come up, not just at the end of the discussion.

goldfirere:
Let's pretend we can snap our fingers and create reasonable technology. It sounds like the ideal scenario would be that each proposal would be discussed on a freshly-created mailing list (with public archives), and to which the proposal author was subscribed and could contribute. This scenario would seem to be a nice middle ground: participants in the discussion are limited, but include the proposal author. (I suppose the official author should be allowed to invite co-authors, too.)

Two questions:
1. Do others agree that this is the best of both worlds?
2. If so, how can we reasonably approximate this behavior?

gbaz:
Why not just encourage discussion to be raised on the ticket? Does limiting the discussion to a narrower group really help/make a difference for most proposals? Or is it more a matter of people just distinguishing/signaling when they raise concerns _as committee members_ vs otherwise?

goldfirere:
> Does limiting the discussion to a narrower group really help/make a difference for most proposals?

I'm assuming that the answer to this is "yes", but I don't know. We could always try doing it as an experiment and then we'd be able to move forward from a position of experience.

One real problem about having the discussion on GitHub: while I follow this repo, proposals that are of less interest to me don't grab my attention. On the other hand, I make it a priority to look at mail on the mailing list. (This aspect of the mailing list has also been raised by others.) So we'd need to do something to grab (and keep) the attention of committee members.

gbaz:
Members could simply subscribe to the tickets they are interested in to get notifications. And if they already subscribe to all but only pay attention to some, they could "star" the threads they are interested in, using their mail client of choice?

goldfirere:
I think the problem is that committee members should be actively involved in every proposal under committee consideration -- not just those proposals that naturally interest them. (This is an ideal that is not quite achieved today, to be fair.) Today, a committee member can easily distinguish these cases by any number of means, given that committee deliberations happen on a different platform than general discussion. But how can we do this if the conversation is on GitHub? I agree that it's sensible to ask committee members to add some mail-reader configuration, but ideally this would be something done once, not per proposal-under-consideration.

nomeata:
> We could always try doing it as an experiment and then we'd be able to move forward from a position of experience.

I’d be happy to conduct such an experiment.

> So we'd need to do something to grab (and keep) the attention of committee members.

I sometimes unsubscribe from noisy proposal discussions where I can’t contribute much, knowing that GitHub will notify and re-subscribe me once someone mentions me (usually to bring the proposal forward to the committee).

A similar scheme could work here as well:
* Committee members may ignore and/org unsubscribe proposals under discussion (or configure GitHub to not even subscribe them to new PRs).
* When the author wants to bring it forward, they mention and thus notify me.
* When I have found a shepherd, I mention and thus notify them (and they can start deliberating with the authors – I am still keen on the “shepherd must raise concerns with the author first before recommending to reject it” aspect of the present proposal). 
* When they have a recommendation, they mention _all_ committee members (we’d have the list somewhere for easy copy’n’paste), and now all committee members are expected to follow the discussion, until a decision has been made.

It’s a Github-heavy process that might not work in a less techy environment, but most (all?) of the committee members have an higher-than-average familiarity with GitHub, so it might actually work.

goldfirere:
Aha! The key detail I was unaware of in Joachim's suggestion is that I can unsubscribe from a thread even if I watch a repo. (Yes, I know that every email from GitHub invites me to mute a thread. But I never used the button and so its existence faded from my consciousness.) Indeed, I was wondering how Joachim executed his secretary duties -- I assumed some mail-reader option that looked for the string "@nomeata" in emails from this repo.

I do think the process he proposes would work. Instead of just deleting emails from uninteresting threads (as I do now), I can officially unsubscribe. Then, when I am @-mentioned, I will rejoin the conversation.

simonpj:
> On the other hand, I make it a priority to look at mail on the mailing list. 

For me this is the key point.  I just don't feel I have the bandwidth to contribute actively to every "live" proposal.  Indeed a major reason for the entire GHC proposal process is to widen the undesirable bottleneck caused by old(er) people like me, and get more people involved in the process of developing and refining proposals.

But we do still need a gatekeeper and, doubtless imperfectly, the committee serves that role.  So, like Richard, I treat the committee mailing list as a much higher priority feed.

But I still feel uncomfortable raising technical points on that list, because they feel as if they "belong" on the original Github discussion.  So I sometimes post to the Github discussion and follow up with an email to the committee list pointing to my Github post.

It doesn't feel entirely satisfactory.  So, addressing the priority feed point, *could the Github email come with a subject line (or early line in the message) saying "Under committee review"*? Then I could filter for that, but the discussion would still all be in one place.

The role of the shepherd in summarising periodically -- for everyone's benefit -- where consensus is slow to evolved, remain critical.

nomeata:
> It doesn't feel entirely satisfactory. So, addressing the priority feed point, could the Github email come with a subject line (or early line in the message) saying "Under committee review"? Then I could filter for that, but the discussion would still all be in one place.

The subject line of the email notifications is the title of the PR; we could manually edit it and prepend “Under Review:”.

gridaphobe:
Editing the title to make it filterable sounds like a pretty good compromise to me.

I think it's worth giving the GitHub workflow a try. We can always return to this discussion if it doesn't work out.

aspiwack:
_I may be repeating what is being said elsewhere, apologies if so, this thread is suprisingly hard for me to follow._

The bottom-line, I think, is that whatever conversation happens on the committee mailing list should be transcribed in the github thread. Presumably, the shepherd would have the task of summarising the committee's opinion.

If something doesn't sit right with the committee, the concern would be thus raised to the proposal author. Which could choose to address them, or deem them insurmountable and abandon the proposal (which would then be marked as rejected).

It doesn't mean there can't be ground for immediate rejection as well, if the core principle of the proposal is deemed so unsuitable that any fix would better be a new proposal.

nomeata:
If I may paraphrase @aspiwack, then he is offering another approach: The committee (usually) doesn’t reject proposals right away, but comes back to the author with “here is why we plan to reject it, please comment/fix”. (In contrast, my idea was that this is done at shepherd recommendation phase).

Pro: The author get to rebut the verdict of the full committee.
Con: The discussion on the committee is still mostly opaque to the author.

goldfirere:
I'm in favor of a practice of prepending "Under review:" to titles under review and trying the GitHub-based discussion.

bravit:
I like this prepending-"Under review:"-workflow idea. It seems more convenient to keep the technical discussion on GitHub. And I like that it keeps authors and the wider community involved.

nomeata:
I think we have enough momentum in the committee in favor of a fully Github-centric approach. I will create a proposal to that end, and put it up for discussion.

BTW, TIL I learned that we can notify the whole committee with @ghc-proposals/ghc-steering-committee, which makes a good workflow for when the shepherd wants to notify and subscribe the committee.

aspiwack:
> In contrast, my idea was that this is done at shepherd recommendation phase

I must confess that I didn't realise that I was proposing something different. I'm not getting this point for reading comprehension…

At any rate, the core of my point was that whatever motivated a committee decision should be mirrored on the Github thread. Mainly because the Github thread is accessible from the proposal itself (or the closed issues list when it's a reject), hence can be perused by the curious onlooker. It is not the case of the mailing list. Since the goal of a proposal, in my eyes, is to document the design process and decisions, it seems natural to me that the committee's reasoning, wherever it happens in practice, be reported in the Github thread. As documentation.

It doesn't have to happen all at once, or the committee may not choose to have their internal discussion on the Github thread. But I thought it may be a good opportunity to ask the proposal authors for comments.

nomeata:
I started a vote on the mailing list.

nomeata:
Superceded by #225

PR #222, opened 2019-04-19
Proposal 47 has been implemented

RyanGlScott:
In https://gitlab.haskell.org/ghc/ghc/merge_requests/789.

(Side note: the label "trac-ticket" is somewhat misleading now that GHC tracks issues on GitLab instead of Trac. Perhaps this should be renamed to something else, like "gitlab-issue"? Although in this case, we're actually linking to a merge request, not an issue...)


nomeata:
Maybe just “ticket”? Seems generic enough to me.

PR #223, opened 2019-04-21
Migrate Trac links to GitLab

RyanGlScott:
Now that GHC's issue tracker is hosted on GitLab instead of Trac, there are a number of things in the accepted GHC proposals that deserve to be updated. In particular:

* Explicit references to Trac have been updated, as Trac is no longer the primary source for GHC issues.
* Links to Trac issues and wiki pages have been converted to their GitLab counterparts.
* The `trac-ticket` label on each proposal is now an anachronism, so I've changed this to `ticket-url`.


nomeata:
Thanks a lot, very helpful!

I wonder if plain full urls in the header field for tickets would be simpler, and easier to follow for uninitiated visitors.

RyanGlScott:
That would work for me. Granted, the name `ticket` doesn't really give a clear indication that it's expecting a URL. Perhaps the label should be named `ticket-url`?

nomeata:
Yeah, that might avoid confusion.

RyanGlScott:
OK, we now use `ticket-url` everywhere.

nomeata:
Thanks a lot!

PR #225, opened 2019-04-26
Process tweak: GitHub discussion

nomeata:
This is a variant of #221: It retains the extra step that the shepherd must confer with the author before proposing to reject a proposal, but it additionally moves GitHub discussion onto the mailing list.

A nice side-effect of this is that the committee mailing list will be mostly only status updates mails (new proposal, new recommendation, decision), and my status summary emails will be much less tedious to assemble :-)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -269,11 +269,11 @@ mailing list. All interested parties are invited to follow the discussion.
 -  Based on the proposal text (but not the GitHub commentary), the shepherd
    decides whether the proposal ought to be accepted or rejected or returned for
    revision.
-   
+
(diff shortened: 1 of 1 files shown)


isovector:
This sounds like a big improvement on the process; I felt like a few of my proposals were rejected without a clear understanding by the committee of the pain points they solved --- but while I could watch the discussion go off the rails, I had absolutely no venue to try and explain myself. Thanks for putting this together @nomeata (and @mpickering for getting the conversation started!)

nomeata:
I started a vote on the mailing list.

nomeata:
Ok, time to make it official!

PR #226, opened 2019-05-03
Typo in Levity Polymorphic Lift Proposal

mpickering:
`DeriveLift` generates quotes, not splices.


nomeata:
Ok, I will bring this proposal before the committee shortly.

nomeata:
(just kidding of course)

PR #177, opened 2018-10-15
Simple constrained type familes

typedrat:
This proposal offers a simpler path to constrained type families, as inspired by discussion in #154.

PR #227, opened 2019-05-08
Put associated-type kind sigs outside classes

goldfirere:
This is a small tweak to the [accepted proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0036-kind-signatures.rst) on top-level kind signatures. The only change is to move associated-type kind signatures outside their class declarations. This is for two reasons:

1. If the signature is within the class, we have some awkwardness around scoping. The kind signature is purportedly "top-level", but if the class variables are in scope there, what does that really mean? If the class variables are not in scope, then the scope of the class variables has a "hole", which is awkward.

2. I would like to see associated type families become more like methods. These would be [constrained](https://github.com/typedrat/ghc-proposals/blob/constrained-type-families/proposals/0000-simple-constrained-type-families.rst) and more tightly associated with their class. By putting signatures *outside* the class, we know that the signature is really the kind of the type family.

This tweak was, in part, fueled by challenges in implementation.

### DIFF
--- a/proposals/0036-kind-signatures.rst
+++ b/proposals/0036-kind-signatures.rst
@@ -75,7 +75,7 @@ Proposed Change Specification
    is done before ever examining the full declaration, just like how GHC treats type
    signatures.
 
-   Associated types may be given kind signatures within their classes.
+   Associated types may be given kind signatures outside their classes.
 
    Unlike type signatures, the type variables brought into scope in a type-level kind
    signature do *not* scope over the type definition.
@@ -164,6 +164,9 @@ Alternatives
   made on the original), but it still means that types have a different treatment from
   terms, which is aesthetically displeasing to me.
 
+* Put associated-type signatures inside their classes. Doing so has unfortunate consequences
+  with respect to the scoping of class variables (do they scope over the kind signature or
+  not?).
 
 Unresolved questions
 --------------------


int-index:
The decision here should be tightly coupled with the design of #148: if we put associated type kind signatures on the top level, we should allow proper top-level signatures for other class methods

Some minutes ago I proposed `-XClassTopScope` there https://github.com/ghc-proposals/ghc-proposals/pull/148#issuecomment-491214983

Now I realize that it would also solve the scoping issues mentioned here. Under this extension, the class variables simply shouldn't scope over the class definition: neither methods nor associated types. And it's how GADTs work, so we are in a familiar territory.

I do not have a strong opinion on what design we use in the end, but I feel strongly about using the *same* design for associated types and methods.

int-index:
In light of #148 getting closed, I think we should close this as well. We seek to unify type families with functions, and, by extension, associated types with class methods. #177 is a step in this direction, let's not make a step in the opposite one! As long as class methods have their signatures inside the class, associated types should too.

What about the syntactic conflict?

```
class C k where
  type T :: k   -- TLKS?
  type T :: k   -- declaration header?
```

The stopgap solution is quite simple, even if not satisfactory: we are under no obligation to support TLKSs for associated types. I have implemented it in my prototype and it was straightforward.

The longterm solution requires more thought, but I think one promising direction is to add syntax for matchable quantifiers and injective functions, then we won't need declaration headers for type families.

simonpj:
>  We can simply consider an associated type to have a CUSK as long as its parent has a TLKS:

That's OK provided we document exactly what the implied TLKS for the associated type is.  What is the rule?  Eg
```
type C :: Type -> Constraint
class C a where
  type T a b
```
There is no explicit kind attribution for `b`.   I'm sure we can invent more complicated scenarios like
```
class C a where
  type T a k (c :: k a)
```
or something wild like that.  

(Maybe this example is bogus, but I still want that specification!)

int-index:
@simonpj The specification is a small modification to the rule we have today. To cite the GHC User's Guide,

> An associated type or data family declaration has a CUSK precisely if its enclosing class has a CUSK.

I propose we append "or a TLKS" to this sentence:

> An associated type or data family declaration has a CUSK precisely if its enclosing class has a CUSK or a TLKS.

That's it. Then we continue to use the existing rules:

* class variables mentioned in the parent class have their kind from the class CUSK (TLKS)
* annotated type variables either have a kind annotation or are defaulted to `Type`
* undetermined kind variables are quantified over

So, let's look at what happens to this example _today_:

```
class C (a :: j) where
  type T a b (c :: Proxy p)
```

GHCi 8.6.5 reports:

```
T :: forall k j (r :: k). j -> Type -> Proxy p -> Type
```

This is according to the rules outlined above:

* `a` is annotated in the parent class, so it has pre-determined kind `j`
* `b` is unannotated, so its kind defaults to `Type`
* `c` is annotated and we use the user-specified kind
* `p` has an undetermined kind `k` which we quantify over

With `-XTopLevelKindSignatures`, the exact same thing will happen if the user writes

```
type C :: j -> Constraint
class C a where
  type T a b (c :: Proxy p)
```

Here, the parent class `C` has a TLKS instead of a CUSK, but we use the old CUSK rules for `T`.

simonpj:
> I propose we append "or a TLKS" to this sentence:

But that sentence talks about CUSKs, which we are busily deprecating in favour of TLKSs!  In a non-CUSK world, what will we say?

Are we _allowing_ a top-level signature for an associated type?  Or disallowing?

I think maybe you are saying: a TLKS for a class C implicitly provides a TLKS for each of C's associated types.  But if so we must specify precisely what the implied TLKS is.

int-index:
> But that sentence talks about CUSKs, which we are busily deprecating in favour of TLKSs!

Right, but let's be careful here. At the moment we have:

1. a set of heuristics to determine whether a type constructor has a CUSK
2. two kind-checking algorithms, `kcLHsQTyVars_NonCusk` and `kcLHsQTyVars_Cusk`.

I'd say all the trouble comes from heuristics (1), but the kind-checking algorithms in (2) are both well-behaved. My proposal is to use `kcLHsQTyVars_Cusk` for associated types as long as its parent class has a TLKS. No heuristics!

And of course, we won't need the word "CUSK" anymore. It all can be worded in terms of "TLKS" as you ask.

> Are we allowing a top-level signature for an associated type? Or disallowing?

We disallow it, and derive the kind of the associated type from kind annotations and the parent class TLKS.

> But if so we must specify precisely what the implied TLKS is.

OK, I can just remove the mention of a "CUSK" from these rules. Here we go:

* class variables have their kind determined by the parent TLKS
* other type variables either have a kind annotation or are defaulted to Type
* undetermined kind variables are quantified over

Therefore, in the following example

```
type C :: j -> Constraint
class C a where
  type T a b (c :: Proxy p)
```

The derived TLKS for `T` is

```
T :: forall k j (r :: k). j -> Type -> Proxy p -> Type
```

Reasoning:

* `a` has its kind determined by the parent TLKS
* `b` is unannotated, so its kind defaults to `Type`
* `c` is annotated and we use the user-specified kind
* `p` has an undetermined kind k which we quantify over

simonpj:
But what is an "undetermined kind variable"?

In your example, you don't quantify over p, which is odd, prorbably a mistake.

p has no explicit binder so what it its kind?

All this seems quite unfortunate.  The glorious thing about TLKSs is that you can look at it and say "ah, that's the kind of C".  But now, in the slightly corner case of associated types, we have a somewhat elaborate mechanism (sufficiently elaborate that I do not yet understand it) for deducing the full kind of T.  I'm quite uncomfortable about this.

An alternative would be: T has a TLKS iff it has a top level kind signature.   That's nice and easy to explain!

int-index:
> In your example, you don't quantify over p, which is odd, prorbably a mistake.

Ah yes, `r :: k` should be `p :: k` – edited.

>  `T` has a TLKS iff it has a top level kind signature. That's nice and easy to explain!

But then we end up with a different special case for associated types: their signatures are top-level, but their declaration headers and default equations are not! That's not nice, at least as long as class methods don't have type signatures outside classes.

Let's think one step ahead, with dependent types, associated types are the same thing as class methods! (Just like closed type families are the same thing as functions). I believe that the right thing is to work towards their convergence, not divergence.

> The glorious thing about TLKSs is that you can look at it and say "ah, that's the kind of C". 

Yes, I know, but another thing about TLKSs is that they make the declaration header for type families mostly redundant. For example,

```
type Id :: k -> k
type family Id a where
  Id x = x
```

Here, the `type family Id a where` doesn't serve much of a purpose, and `a` doesn't scope over the body. And how come we don't need it in terms?

```
id :: k -> k
id x = x
```

The reasons for having a type family header are listed in the proposal, we need it to determine:

1. whether the type is generative and/or injective
2. whether the type is open or closed
3. whether the type must be applied to a certain prefix of arguments (arity)

If we put this information into the kind signature, then we're fine without a type family header. And I'm quite certain that this is the direction in which we want to go:

0. having this information in the TLKS makes it a sufficient description of a type family for `.hs-boot`
1. matchability syntax is needed for [Higher-order Type-level Programming in Haskell](https://www.microsoft.com/en-us/research/uploads/prod/2019/03/ho-haskell-5c8bb4918a4de.pdf)
2. arity is already a very pressing issue in the current TLKS implementation, as it stands TLKSs cannot properly replace CUSKs without syntax for arity (see ["Matchability and arity inference"](https://gitlab.haskell.org/ghc/ghc/wikis/ghc-kinds/kind-inference/tlks#matchability-and-arity-inference))
3. associated types are open, so the open/closed distinction isn't an issue here (even longer-term, I'd say we should deprecate non-associated type families for reasons described in ["Constrained Type Families"](https://cs.brynmawr.edu/~rae/papers/2017/partiality/partiality-extended.pdf), and then every non-associated type family can be considered closed)

OK, once this is done, we don't need type family headers. This is the point at which we can safely allow kind signatures for associated types:

```
type C :: j -> Constraint
class C a where
  T :: j -> Type -> Proxy p -> Type
```

Then the derived top-level kind of `T` is

```
T :: forall k j (p :: k). forall (a :: j) -> C a => Type -> Proxy p -> Type
```

Which is essentially the same thing as we would infer for a term-level class method `t` if we used visible dependent quantification for otherwise ambiguous type variables:

```
class C (a :: j) where
  t :: j -> Type -> Proxy p -> Type

t :: forall j.
     forall (a :: j) ->
     C a =>
     forall k (p :: k). j -> Type -> Proxy p -> Type
```

That's it, equivalent type/kind inference rules for class methods and associated types.

But this is all very far fetched and will require numerous proposals. For now, I only say that it's a reasonable solution to simply keep the system that works well enough for associated types today, until we get enough bells and whistles to give associated types their proper kind signatures.

int-index:
I guess that one way to put it is that TLKSs outside classes for associated types are indeed a design that has its merits, but I'm looking for a design that would work for *both* associated types *and* class methods. And after the discussion in #148 I am not sure at all that this design is to put all signatures outside classes, I hesitate to move in this direction.

goldfirere:
I support @int-index's instincts here: that what's good for class methods must be good for associated types. (I, too, see them as two different manifestations of the same underlying idea.) So, either both are allowed to have signatures outside of classes, or neither is. Right now, we're trending toward "neither is".

So then the challenge is to figure out what to do instead. That is, when we see an associated type declaration, what is its kind? (And can it participate in polymorphic recursion?)

**Definition:** A type/kind is *stable* when it is completely known. A stable type has no metavariables, and all quantification has been sorted out.

**Corollary:** A top-level stable type is closed.

**Observation:** Only entities with stable types may participate in polymorphic recursion.

The whole reason CUSKs came about and that we want TLKSs is so that we can bestow stable kinds on datatypes. Today, iff a type has a CUSK, it has a stable kind. Tomorrow, if a type has a TLKS, it also has a stable kind.

**Question:** If an associated type cannot be given a TLKS (and we have no CUSKs anymore), when can it have a stable kind?

**Proposed answer:** An associated type has a stable kind iff its enclosing class has a stable kind.

Note that top-level open type families *always* have a stable kind. The specification is this:
* All un-annotated Required variables are assigned kind `Type`
* All Specified variables (e.g., the `k` in `type family F (a :: k Bool)`) have their kind inferred and generalized.

For associated types, we just add one more bullet:
* In-scope type variables (from the enclosing class) retain their kind from the class declaration.

One important detail to remember here is that a primary source of difficulty in inferring kinds comes when we infer a type's kind by looking at its definition. (For example, we look at a class's definition to determine the kinds of its type variables.) But for open type families (associated or not), there *is* no definition. All of the kind information must be in the header.

simonpj:
You make the point that TLKS => stable kind.

But my point is different: I want to be able to write down the TLKS of a type constructor as machine-checked documentation, so that I, the programmer, know what it is.   CUSKs are horrible in part because you have a deal of type inference to do before you know the stabld kind of the type constructor.  TLKSs mean that you simply write down that stable kind.

 And I think I can write down its TLKS.    TL;DR: when you say "the specification of the stable kind of a top-level family is..." I think ok ok, but please please don't stop me writing its kind signature directly."   I don't want to have to execute your specification (which involves kind inference) in my head.  I just want to write the TLKS. And I can!

What about associated types?   Look at class methods, you say -- you can't write a top-level signature for them!   No, but for class methods there is no inference to do.  Given 
```
class blah => C a b  where
   op :: ty
```
the toplevel type of `op` is `op :: C a b => ty`.  If you like, if we had a top-level signature for `op` that is exactly what it'd be.

But, sadly, there is no such simple rule for associated types.   That's why I'm so uncomfortable.  I want to write its stable kind. 

I supopse we could allow both a signature and a header in a class decl:
```
class C a b where
   type F :: forall k. k -> Type -> Type   -- The top-level kind
  type F x b   -- Defined arity, and establishes connection with the class
```

int-index:
> But, sadly, there is no such simple rule for associated types. That's why I'm so uncomfortable.

And I'm uncomfortable because you are proposing a _different_ simple rule for them. Presumably, the rule for the class is to quantify over class variables and add the class constraint. That's what your `C a b => ty` is about. But if we try the same with associated types, it's not as straightforward.

Consider this example:

```
class C (a :: k) where
  type F a :: r
```

Here we have `F :: k -> r`. Or, using VDQ, `F :: forall (a :: k) -> r`. So, presumably, with TLKS you are proposing that we write

```
type C :: k -> Constraint
class C a where
  type F :: k -> r
  type F a
```

except if we apply the rules we use for classes, i.e. adding a constraint and implicitly quantifying class variables, we get `F :: forall a. C a => k -> r`, which is clearly wrong! OK, you might say that associated types differ in that they don't introduce a class constraint, but it's the "implicitly quantify" part that bothers me. If we remove the constraint, we get `F :: forall a. k -> r`, which is _still_ wrong. And if we remove quantification as well, then we are removed even farther from the rules of class methods.

Instead, let's add the following rule:

* class variables ambiguous in the method signature are quantified visibly

For terms, this would mean that

```
class C a where
  m :: ty
```

would translate into

```
m :: forall a. C a => ty    -- if  ty  determines  a  (ambiguity check)
m :: forall a -> C a => ty  -- if  ty  doesn't determine  a
```

This will be possible when we get VDQ in terms. Then, applying the exact same rule to associated types, we should write

```
type C :: k -> Constraint
class C a where
  type F :: ki
```

Then we get

```
m :: forall a. C a => ki    -- if  ty  determines  a  (ambiguity check)
m :: forall a -> C a => ki  -- if  ty  doesn't determine  a
```

Except we remove the `C a =>` constraint part because we don't have constraints in types yet. When we get constraints in types, we do add the constraint (#177).

-----

So, in your example

```
class C a b where
  type F :: forall k. k -> Type -> Type   -- The top-level kind
  type F x b   -- Defined arity, and establishes connection with the class
```

this isn't right, we're not using the same rules. It should rather be

```
type C :: k -> Type -> Type
class C a b where
  type F :: Type
```

and then we get

```
F :: forall (a :: k) -> forall (b :: Type) ->          Type
F :: forall (a :: k) -> forall (b :: Type) -> C a b => Type  -- with #177 
```

Except we can't do it just yet, because for everything to fit together we need syntax for specifying arity, and we need constraints in types, and VDQ in terms, etc. So, for now I propose we keep the machinery that already works, and then direct our effort towards a design which is consistent across types and terms (associated types and class methods).

simonpj:
The trouble with "class variables ambiguous in the method signature are quantified visibly"  is that it is non-syntactic.  My goal is to be able to make the program source give the stable kind of a type constructor, as an ordinary signature.   (And ditto for terms.)  So that the programmer can just read it from the program source text.

That rule holds for data constructors, if you write them in GADT style.  (I don't say that the type/kind can _always_ be read off; just that it should be possible to write the program so that it can be read off.  So "if you wrte in GADT style" is fine.)

It holds for every type constructor where you can write a top-level kind signature.

Sadly it does not hold for class methods, but the rule  is purely syntactic: just add "forall a. C a =>" to the front of the written type.

I'm looking for something similarly simple for associated types.  And indeed my straw man treats associated types more like GADT-style data constructors than like class methods.   (That is, my straw man says that the kind signature, if you give it, is a stand-alone top-level signature, with no scoping from the header, just like GADTs.)  I can see it's un-satisfying to treat associated types and methods differently though.

I sometimes wonder if associated types are worth the bother!

int-index:
> Sadly it does not hold for class methods, but the rule is purely syntactic: just add "forall a. C a =>" to the front of the written type.

Except this rule isn't good enough, we get ambiguous types this way. For example:

```
class Sized a where
  sizeOf :: Int
```

It isn't very useful that we get `sizeOf :: forall a. Sized a => Int` out of this declaration because any occurrence of `sizeOf` not followed by a type application results in cryptic errors.

I would rather get `sizeOf :: forall a -> Sized a => Int` and use it as `sizeOf Bool` and `sizeOf String` (with VDQ in terms). `sizeOf :: Int` would tell me that I haven't supplied enough arguments instead of an error about ambiguous type variables. The rule is not syntactic, but it's no good to have a purely syntactic rule that produces ambiguous types.

goldfirere:
I'm personally not a fan of the "quantify visibly" rule -- I would find it quite unexpected as a programmer. Following Simon, I tend to prefer "predictable but stupid" (i.e. giving the ambiguous type to the class method) over "smart but inscrutable" (i.e. figuring out what the ambiguous variables are and then quantifying them visibly).

However, I wouldn't call the current rule for class methods purely syntactic. Examples:

```
type C :: forall k. k -> Constraint
class C a where
  m :: Proxy a -> ()

type C2 :: forall k1 k2. k2 -> k1 -> Constraint  -- NB: order backwards
class C2 a b where
  m2 :: Proxy a -> Proxy b -> ()   -- what are the Specified parameters to m2?

type C3 :: forall k -> k -> Constraint   -- NB: VDQ
class C3 k a where
  m3 :: Proxy a -> ()  -- ???
```

Sitting here, I'm actually not even sure what the full, top-level types of `m2` and `m3` are expected to be. I have guesses, but I'm not sure.

Separately, as #148 points out, field selectors also cannot be given their full types anywhere; instead, we have to do some syntactic rearrangement to get its type.

simonpj:
> Sitting here, I'm actually not even sure what the full, top-level types of m2 and m3 are expected to be

The syntantic transformation would give
```
m2::  C a b => Proxy a -> Proxy b -> ()
m3 :: C3 k a => Porxy a -> ()
```
Are those wrong?

goldfirere:
What about the kind variables? They seem Specified to me, from the TLKS.

simonpj:
They are Specified in the kind of C, but (according to my simple-minded syntactic story) not Specified in the type of `m2`.  If you wanted to make them Specified, you'd hyave to say somethign like
```
class C2 a b where
   m2 :: Proxy (a :: k1) -> Proxy (b::k2) -> ()
```
so that the syntactic transformation gave you
```
m2::  C a b => Proxy (a ::k1) -> Proxy (b::k2)-> ()
```
Or you could specify them in the `class` header.  Either way, it's clear what `m2`'s type is.

goldfirere:
I've lost enthusiasm for this. Why? Because it occurs to me that it's not really so important for associated types to behave like class methods. The motivation for doing so is that, in the limit, associated types will merge with class methods. This, I almost agree with. My "almost" qualification is that, in the limit (i.e. when we have dependent types), we'll have no need of associated types, and thus we can just have class methods. So: the treatment of associated types won't matter so much.

Thus: I support whatever is easiest here. I *think* that's putting the TLKS outside the class, but I'm happy to be swayed.

int-index:
> we'll have no need of associated types, and thus we can just have class methods

That's one way to put it, but another way to put it is that with dependent types, these two notions should coincide. You seem to propose to support both of:

* dependent class methods
* associated types

And you say it is not important to have consistent treatment for them: it's OK to parse, rename, and typecheck them differently. But this means we condemn ourselves to support two different ways to parse, rename, and typecheck two language constructs which in principle should be the same thing.

Sure, we may deprecate associated types eventually, but if that's the plan then it is wasted effort to support other features based on associated types. Do we add TLKSs for associated types outside classes only to deprecate them later in favor of dependent class methods? It's a stillborn feature then.

Instead, I see this as an opportunity to come up with a design that works both for associated types and dependent class methods. I made an attempt with the "quantify visibly if ambiguous" rule, I still stand by it (despite the expressed disapproval). Coupled with VDQ in terms, It rids us of the need for `-XAllowAmbiguousTypes`, and provides a way for convergence of associated types and class methods, rather than their divergence and then deprecation.

simonpj:
I'm content with any solution that allows me, as a programmer, to read off the kind of the associated type from the text of the program, with at most a syntactic transformation (as for methods).

I think that means that
* The user should be able to specifiy the fulll, top-level (i.e. closed) kind of the assocaited type
* I don't mind too much whether that full TLKS is actually at top level, or syntactically nested in the class decl.  (I mildly prefer the latter)

int-index:
> I'm content with any solution that allows me, as a programmer

OK @simonpj, I wrote down a design that fits this requirement: #236 

Short term, it implies having signatures for associated types on the top-level (outside of the class).

int-index:
> I'm content with any solution that allows me, as a programmer, to read off the kind of the associated type from the text of the program, with at most a syntactic transformation (as for methods).

OK @simonpj, I wrote down a design that fits this requirement: #236 

Short term, it implies having signatures for associated types on the top-level (outside of the class). I'm content with this because I give class methods equal treatment.

Ericson2314:
I prefer convergence over deprecation because I want to someday use today's old code not caring about whether an associated type or function was used.

goldfirere:
We now have an implementation of standalone kind signatures (full credit to @int-index) that simply omits kind signatures for associated types. I'm OK with this state of affairs, but will start a new proposal that officially removes support for signatures on associated types.

goldfirere:
New proposal to remove associated type signatures at #259.

PR #119, opened 2018-04-06
Resurrect PatternSignatures

nomeata:
[Rendered](https://github.com/ghc-proposals/ghc-proposals/blob/patternsignatures/proposals/new-proposal.rst)

PR #228, opened 2019-05-09
Function result type signatures

int-index:
We propose to allow function result type signatures, shown as ``:: r`` in the following example:

```
f (x :: a) (y :: b) :: r = rhs
```

Note that this is an annotation on the LHS, not on the RHS:

```
f (x :: a) (y :: b) = rhs :: r   -- not the same!
```

It is a spiritual successor of #185, incorporating feedback from @simonpj

[Rendered](https://github.com/int-index/ghc-proposals/blob/fn-res-sig/proposals/0000-function-result-sigs.rst)

### DIFF
--- a/proposals/0000-function-result-sigs.rst
+++ b/proposals/0000-function-result-sigs.rst
@@ -172,9 +172,21 @@ Proposed Change Specification
 
 Allow function result type signatures on the left-hand side.
 
-**Syntax.** These signatures are already a part of the grammar, and a
-validation step rejects them. This check will be removed, and ``FunBind``
-extended with a result type.
+**Syntax.** Take the Haskell 2010 function left-hand side grammar as the
+starting point::
+
+  funlhs -> var apat {apat}
+          | pat varop pat
+          | ( funlhs ) apat {apat}
+
+The change is to remove the mandatory ``apat`` from the first rule and add to
+an optional type annotation::
+
+  funlhs -> var {apat}
+          | pat varop pat
+          | ( funlhs ) apat {apat}
+
+  funlhs' -> funlhs :: [context =>] type
 
 **Semantics.** The result type signature is unified with the inferred type of
 the function body. It does not enable polymorphic recursion.
@@ -207,6 +219,10 @@ allow scoped type variables::
 
   x :: [a] -> [a] = reverse    -- accepted as FunBind
 
+This is the result of this grammar change::
+
+  - funlhs -> var apat {apat}
+  + funlhs -> var {apat}
 
 Costs and Drawbacks
 -------------------
@@ -246,7 +262,7 @@ In general, the situation with parenthesization in bindings is quite elaborate::
   (f a)   = ...  -- parse error
   (f a) b = ...  -- FunBind
 
-A possible solution is change this parsing rule in the Haskell Report::
+A possible solution is change this parsing rule::
 
   funlhs ->
     ...
@@ -262,3 +278,7 @@ Implementation Plan
 -------------------
 
 I (Vladislav Zavialov) will (attempt to) implement.
+
+The function result signatures are already a part of the ``Parser.y`` grammar,
+and a validation step rejects them. This check will be removed, and ``FunBind``
+extended with a result type.
--- a/proposals/0000-function-result-sigs.rst
+++ b/proposals/0000-function-result-sigs.rst
@@ -179,7 +179,7 @@ starting point::
           | pat varop pat
           | ( funlhs ) apat {apat}
 
-The change is to remove the mandatory ``apat`` from the first rule and add to
+The change is to remove the mandatory ``apat`` from the first rule and to add
 an optional type annotation::
 
   funlhs -> var {apat}
@@ -243,9 +243,8 @@ Alternatives
 Unresolved Questions
 --------------------
 
-What treatment do we give to the following example?
+What treatment do we give to the following example? ::
 
-::
   (x :: [a] -> [a]) = reverse
 
 At the moment, it is parsed as a ``PatBind`` and rejected because scoped type
--- a/proposals/0000-function-result-sigs.rst
+++ b/proposals/0000-function-result-sigs.rst
@@ -1,0 +1,208 @@
+Function Result Type Signatures
+===============================
+
+.. proposal-number::
+.. ticket-url::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/228>`_.
+.. sectnum::
+.. contents::
+
+We propose to allow function result type signatures, shown as ``:: r`` in the
+following example::
+
+  f (x :: a) (y :: b) :: r = rhs
+
+Note that this is an annotation on the LHS, not on the RHS::
+
+  f (x :: a) (y :: b) = rhs :: r   -- not the same!
+
+Motivation
+----------
+
+There are several reasons to add function result type signatures:
+
+1. Code style
+2. Fine-grained control over type checking
+3. Consistency with types
+
+Let us take a closer look at each of these points.
+
+1.  **Code style.** There are two ways to think of a type signature: it is a
+    standalone summary of a definition, or it is an integral part of it.
+
+    The first way of thinking leads to idioms such as top-level type signatures,
+    comma-separated names in signatures, and interface/implementation sections::
+
+      -- Interface
+      even, odd :: Integer -> Bool
+      concat :: [[a]] -> [a]
+      reverse :: [a] -> [a]
+
+      -- Implementation
+      even = ...
+      odd = ...
+      concat = ...
+      reverse = ...
+
+    The second way of thinking leads to inline type annotations on patterns and
+    function bodies::
+
+      even (a :: Integer) = ... :: Bool
+      odd (a :: Integer) = ... :: Bool
+      concat (xs :: [[a]]) = ... :: [a]
+      reverse (xs :: [a]) = ... :: [a]
+
+    Both styles have their merit. Haskell has good support for the former,
+    while the latter would benefit from moving the annotation to the left-hand
+    side. For one thing, it is what Scala, F#, OCaml, and some other languages
+    do, so we would better accomodate users coming from these languages by
+    allowing this style::
+
+      def fib(n: Int): Int = ...        -- Scala
+      let rec fib (n : int) : int = ... -- F# and OCaml
+      fib (n :: Int) :: Int = ...       -- Haskell (proposed)
+
+    For another thing, in the presence of guards a type annotation on the
+    right-hand side must be duplicated, or else it would only annotate only one
+    of the equations::
+
+      wet (x :: a)
+        | p x = g x      :: [a]   -- duplicated [a]
+        | otherwise = [] :: [a]
+
+      dry (x :: a) :: [a]         -- no duplication
+        | p x = g x
+        | otherwise = []
+
+2.  **Fine-grained control over type checking.** Type annotations in
+    patterns and in expressions behave differently when it comes to type
+    variables.
+
+    In patterns, the type variables are bound to the inferred types::
+
+      f :: (Bool -> Bool) -> Bool
+      f (g :: a -> a) = g True    -- OK, a ~ Bool
+
+    In expressions, the type variables are implicitly quantified::
+
+      f :: (Bool -> Bool) -> Bool
+      f g = (g :: a -> a) True    -- Couldn't match type ‘a’ with ‘Bool’
+
+    We do not unify ``a ~ Bool``, quantifying over ``a`` instead. Thus, there
+    is an important design decision: should type variables in function result
+    type signatures on the left-hand side exhibit pattern-like (unification) or
+    expression-like (quantification) behavior?
+
+    We propose that type signatures on the left-hand side are subject to
+    unification, and type signatures on the right-hand side are subject to
+    quantification. This is a simple, consistent principle, which gives more
+    control to the user: now it is possible to control whether unification or
+    quantification is used by putting the type annotation in the appropriate
+    position.
+
+    In particular, consider the following example::
+
+      qf = <rhs> :: a -> a
+      uf :: a -> a = <rhs>
+
+    ``qf`` and ``uf`` are vastly different. ``qf`` requires ``rhs`` to be fully
+    polymorphic, so there is only possible implementation::
+
+      qf = id :: a -> a   -- OK
+      qf = not :: a -> a  -- Couldn't match type ‘a’ with ‘Bool’
+
+    ``uf`` may unify ``a`` with another type, so ``uf`` may be defined as any function where
+    the monaid is equal to the codomain::
+
+      uf :: a -> a = id       -- OK
+      uf :: a -> a = not      -- OK, a ~ Bool
+      uf :: a -> a = ('x':)   -- OK, a ~ String
+
+    We expect programmers to make use of this power.
+
+3.  **Consistency with types**. In data declarations, we can add kind
+    annotations to the bound type variables::
+
+      data Vec (a :: Type) (b :: Nat) where <defn>
+
+    We can also add a result kind annotation::
+
+      data Vec (a :: Type) (b :: Nat) :: Type where <defn>
+
+    Note that all of these annotations are in the declaration header, not in
+    the definition.
+
+    With the recently accepted proposal about top-level kind signatures, one
+    may write a top-level kind annotation::
+
+      data Vec :: Type -> Nat -> Type
+      data Vec a b where <defn>
+
+    Therefore, in types we have:
+
+    * Kind annotations on binders: Yes.
+    * Top-level kind signatures: Yes. (Implementation in progress)
+    * Result kind annotations: Yes.
+
+    In terms, the situation differs. We have top-level type signatures, and
+    with ``-XScopedTypeVariables``, we may write type annotations for the bound
+    variables::
+
+      f :: Bool -> Integer
+      f (a :: Bool) = <defn>
+
+    However, the result type annotations are not allowed::
+
+      f (a :: Bool) :: Integer = <defn>
+                    ^^^^^^^^^^
+
+    Therefore, in terms we have:
+
+    * Type annotations on binders: Yes.
+    * Top-level type signatures: Yes.
+    * Result type annotations: No. (Proposed)
+
+    This discrepancy is slightly annoying from the language design standpoint.
+
+
+Proposed Change Specification
+-----------------------------
+
+Allow function result type signatures on the left-hand side.
+
+**Syntax.** These signatures are already a part of the grammar, and a
+validation step rejects them. This check will be removed, and ``FunBind``
+extended with a result type.
+
+**Semantics.** The result type signature is unified with the inferred type of
+the function body. It does not enable polymorphic recursion.
+
+
+Costs and Drawbacks
+-------------------
+
+This is one more feature to implement and maintain.
+
+
+Alternatives
+------------
+
+* We could treat ``f :: t = <rhs>`` equivalently to ``f = <rhs> :: t``, but
+  this is neither consistent nor terribly useful.
+
+* We could detect CUSKs as we do in types to enable polymorphic recursion, but
+  this makes little sense as we are in the proccess of their deprecation.
+
+
+Unresolved Questions
+--------------------
+
+None.
+
+
+Implementation Plan
+-------------------
+
+I (Vladislav Zavialov) will (attempt to) implement.
--- a/proposals/0000-function-result-sigs.rst
+++ b/proposals/0000-function-result-sigs.rst
@@ -114,7 +114,7 @@ Let us take a closer look at each of these points.
       qf = not :: a -> a  -- Couldn't match type ‘a’ with ‘Bool’
 
     ``uf`` may unify ``a`` with another type, so ``uf`` may be defined as any function where
-    the monaid is equal to the codomain::
+    the domain is equal to the codomain::
 
       uf :: a -> a = id       -- OK
       uf :: a -> a = not      -- OK, a ~ Bool
--- a/proposals/0000-function-result-sigs.rst
+++ b/proposals/0000-function-result-sigs.rst
@@ -182,11 +182,14 @@ starting point::
 The change is to remove the mandatory ``apat`` from the first rule and to add
 an optional type annotation::
 
-  funlhs -> var {apat}
-          | pat varop pat
-          | ( funlhs ) apat {apat}
+  funlhs' -> var {apat}
+           | pat varop pat
+           | ( funlhs' ) apat {apat}
 
-  funlhs' -> funlhs :: [context =>] type
+  funlhs -> funlhs' :: [context =>] type
+
+This results in an ambiguity with pattern bindings, which is resolved in favor
+of function bindings.
 
 **Semantics.** The result type signature is unified with the inferred type of
 the function body. It does not enable polymorphic recursion.
@@ -221,8 +224,8 @@ allow scoped type variables::
 
 This is the result of this grammar change::
 
-  - funlhs -> var apat {apat}
-  + funlhs -> var {apat}
+  - funlhs  -> var apat {apat}
+  + funlhs' -> var {apat}
 
 Costs and Drawbacks
 -------------------
@@ -263,15 +266,15 @@ In general, the situation with parenthesization in bindings is quite elaborate::
 
 A possible solution is change this parsing rule::
 
-  funlhs ->
+  funlhs' ->
     ...
-    | ( funlhs ) apat {apat}
+    | ( funlhs' ) apat {apat}
 
 We can treat all of the examples above as ``FunBind`` by removing the mandatory ``apat``::
 
-  funlhs ->
+  funlhs' ->
     ...
-    | ( funlhs ) {apat}
+    | ( funlhs' ) {apat}
 
 Implementation Plan
 -------------------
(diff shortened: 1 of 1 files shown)


phadej:
How this interacts with warnings, such top level definition without type signature. Do we need *CUST* concept, such when all argumentaatioon and return type are annotated, those act as top-level signature?

int-index:
I mention this in the proposal:

> It does not enable polymorphic recursion.

And elaborate further in the "Alternatives" section:

> We could detect CUSKs as we do in types to enable polymorphic recursion, but this makes little sense as we are in the proccess of their deprecation.

In other words: no, inline signatures are *never* considered equivalent to top-level signatures, they still warn under `-Wmissing-signatures`, and do not enable polymorphic recursion.

Ericson2314:
This is really nicely written (love the ratio of motivation to specification). Even in a pessimistic zero-sum-game light, I think consistency with the rest of the language is more important than having one way to do things. And I can imagine a positive sum game where this someday clarifies highly dependent code where different equations unify with the signature in interesting ways.

simonpj:
I like this now.  Thank you.

I urge adding language that says explicitly that result type signatures behave just like pattern signatures, as in `\ (x :: a->a) -> blah`.    That is: there is no implicit quantification; its' a binding site for `a` (in this example); and the `a` might be bound to any type e.g `Int`.   The details of pattern signatures are worked out in the paper [Type variables in patterns](https://www.microsoft.com/en-us/research/publication/type-variables-patterns/), which is worth citing.

Indeed best of all would be to augment the typing rules of that paper to cover result signatures; not hard I think.

int-index:
> Indeed best of all would be to augment the typing rules of that paper to cover result signatures; not hard I think.

That would be something I'd be happy to do if I understood these rules in the first place, and at the moment I do not. I definitely agree that it's a good idea to formalize this proposal, although I also hope that lack of formalization is not a showstopper.

AntC2:
To check my understanding of quantification vs unification ... With these examples being different

    qf = <rhs> :: a -> a
    uf :: a -> a = <rhs>

How does it go with

    rf :: a -> a = <rhs> :: a -> a

Presumably the rhs signature is referencing the tyvar `a` bound on the lhs(?) So `rf`'s rhs as `not` is acceptable; and uses unification; and the signature has different semantics to `qf`, even though the rhs looks the same. Writing it bare like that is daft repetition, but it might turn up in a form like

    rf2 :: a -> a = case blah of
                         blahA :: b -> case blah2 of
                                            blahB :: c -> <rhsB> :: a -> a
                                            ...
                         ...
                    where foo :: a = <rhsfoo> :: a

The `a` bound to `foo` is fresh. Or is it(?) So for a reader to make sure what that deep-nested `<rhs> :: a` refers to might need a bit of sleuthing.

int-index:
That's a very interesting example, @AntC2 

```
rf :: a -> a = <rhs> :: a -> a
```

To find out its semantics, let us consider a similar example where `a` is pattern-bound:

```
pf (g :: a -> a) = g . not :: a -> a
```

This compiles and GHC infers:

```
pf :: (Bool -> Bool) -> Bool -> Bool
```

At least, this is what GHC 8.8 and GHC HEAD do with `-XScopedTypeVariables` on. You may see a different result with older GHCs.

The principle is that function result type signatures behave like pattern signatures. Therefore, this analysis of yours is correct:

> Presumably the rhs signature is referencing the tyvar `a` bound on the lhs(?) So `rf`'s rhs as `not` is acceptable; and uses unification; and the signature has different semantics to `qf`, even though the rhs looks the same. 

As to sleuthing: to make sure a variable is fresh, just use an explicit `forall`. Consider this:

```
pf' (g :: a -> a) = g . not :: forall a. a -> a
```

Now GHC barfs that it `Couldn't match type ‘a1’ with ‘Bool’`. Good. And no need to inspect the LHS to find out if `a` is quantified.

shayne-fletcher-da:
Brilliant! Way to go @int-index !!

AntC2:
There's an argument in favour of this proposal that doesn't appear in the write-up. Is that because I'm expecting too much?

I'm expecting that with this proposal, code can avoid the artificial `forall` quantification in stand-alone signatures of `ScopedTypeVariables`, and will just need the `PatternSignatures` part. Because you can use  a combo of pattern signatures binding to the arguments, function result type signatures (this proposal) and `@` bindings in patterns (some other proposal) to get access to all the types mentioned in the function's signature.

I'm scanning the 'Type Variables in Patterns' paper for an example where binding tyvars in function result signatures would help, but can't find anything so far.

int-index:
> You can use a combo of pattern signatures binding to the arguments, function result type signatures (this proposal) and `@` bindings in patterns (some other proposal) to get access to all the types mentioned in the function's signature.

This thought has not occurred to me, that's a good observation! Indeed, in case this proposal is accepted, this observation could be used as motivation to resubmit #119

That said, I'd like to avoid feature creep, and this idea depends on #119, so I'd rather pursue it there, not here. (But I do like the thought! I've grown to dislike the `forall` part of `-XScopedTypeVariables` lately after I had to deal with it during implementation of #54, on the other hand `the -XPatternSignatures` is great).

goldfirere:
I agree that binding type variables with `@` ([accepted proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0050-type-lambda.rst)) gets rid of the need for the `forall` behavior of `ScopedTypeVariables`, but I don't see how that interacts with this proposal, really. Note that, for the `@`-bindings to work, we need already to know the type of the function. Maybe that restriction can be relaxed later, but it's not yet obvious how.

int-index:
@nomeata I suppose enough time has passed to submit this, what do you think?

simonpj:
This proposal has been submitted to the committee. I am the shepherd.  We welcome all authors to join the discussion, but kindly ask others to refrain from posting, except to correct egregious errors or offer important clarification.

I recommend acceptance.  It's a fairly simple extension, and very much in the spirit of what we already have.

I really, really would like to see the proposal formalised in the style of [Type variables in patterns](https://www.microsoft.com/en-us/research/publication/type-variables-patterns/).  I don't think it would be hard, given access to the source code of the paper.

Committee members: please comment.

simonpj:
One matter of detail. Consider this
```
(x :: [Int]->[Int], y :: Bool) = e
```
This is legal today.    But this is not:
```
(x :: [a]->[a], y :: Bool) = e
```
we get:
```
    • You cannot bind scoped type variable ‘a’
        in a pattern binding signature
    • In the pattern: x :: [a] -> [a]
```
This restriction (no unbound variables in a pattern binding) is not properly documented in [the relevant user manual section](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/glasgow_exts.html#pattern-type-signatures).

But what about
```
x :: [a] ->[a] = e
```
Is that
(a) a 0-ary function binding, falling under this proposal, with 'a' scoping over the RHS
(b) a degnerate pattern binding, rejected by the above restriction.

I think (a).  But the proposal should point out the ambiguity and say how it is resolved.

Presumably if you wite
```
(x :: [a] -> [a]) = e
```
now it is definitely (b)

int-index:
@simonpj Indeed, and there's a note about this – `Note [FunBind vs PatBind]`. However, today's arrangement is quite strange:

```
f       = ...  -- FunBind
f a     = ...  -- FunBind
(f a)   = ...  -- PatBind
(f a) b = ...  -- FunBind
```

I'd rather have `(f a)` be a `FunBind` instead of a parse error! Thus I'm not inclined to cement the current arrangement in this proposal by saying that `(x :: [a] -> [a]) = e` is a `PatBind` – I don't think it should be.

But I will clarify that `x :: [a] ->[a] = e` is indeed a `FunBind`.

simonpj:
I agree that it's odd that `(f a) = ...` is a pat-bind.

I think it would be helpful to add a BNF to the proposal, so we know what is proposed.  (Saying "it's implemented" doesn't help a programmer!)   Can yo do that?   You can start from the BNF in the report.

int-index:
Done.

* Specified that `x :: [a] ->[a] = e` is to be treated as a `FunBind` in "Effect and Interactions"
* Described the problem of `(x :: [a] ->[a]) = e` and `(f a) = e` in "Unresolved Questions".

If the committee decides that we should change the treatment of `(f a) = e`, I shall move it from "Unresolved Questions" to "Proposed Change Specification".

simonpj:
> Specified that x :: [a] ->[a] = e is to be treated as a FunBind in "Effect and Interactions"

Perhaps it belongs in Specification.  Also BNF?  THanks

int-index:
Do you ask for a Haskell Report based BNF or the `Parser.y` based BNF? The former does not cover enough, as it does not appear to cover pattern signatures at all. The latter already contains the required productions (should I copy/paste?)

simonpj:
As mentioned [above](https://github.com/ghc-proposals/ghc-proposals/pull/228#issuecomment-497270432) I was mainly thinking of starting from the (human-readable) Report BNF.  It's just the changes to function bindings we need to cover, not the full syntax of patterns.   

Stuff about `Parser.y` could still be interesting though; you could summarise under Implementation.

int-index:
>  It's just the changes to function bindings we need to cover, not the full syntax of patterns.

Great. I added the grammar change to the proposed change specification. Haskell 2010 Report grammar:

```
  funlhs -> var apat {apat}
          | pat varop pat
          | ( funlhs ) apat {apat}
```

Proposed grammar:

```
  funlhs -> var {apat}
          | pat varop pat
          | ( funlhs ) apat {apat}
 
  funlhs' -> funlhs :: [context =>] type
```

> Perhaps it belongs in Specification.

I added the grammar change to the specification but kept the explanation of its consequences in "Effect and Interactions" to keep the specification short.

> Stuff about `Parser.y` could still be interesting though; you could summarise under Implementation.

I moved what was my prior specification of syntax to the "Implementation Plan" section.

simonpj:
Thanks.  But
* Where is `funlhs'` used?   I think you should probably swap the names of `funlhs` and `funlhs'`, so that the existing uses of `funlhs` now get a result signature.

* Removing that mandatory `apat` now makes the syntax of `funlhs` overlap with that of `pat`, which makes the grammar ambiguous.   Indeed that's probably exactly why that mandatory `apat` was there in the first place!
```
decl  -> gendecl 
        |  (funlhs | pat) rhs
```  
Either you'll need to add a verbal side condition "choosing funlhs over pat in the var case") or you could have a `non_var_pat` non-terminal.

int-index:
> Where is `funlhs'` used? 

It was implied that `funlhs'` is to be used in place of `funlhs`, but I swapped as you ask to make it more clear.

> which makes the grammar ambiguous.

I know, but none of this is new, we already have this ambiguity in the `Parser.y` grammar and it is already resolved in favor of `FunBind`. There's `Note [FunBind vs PatBind]` about it. Writing this in "Proposed Change Specification" feels post hoc. I added it nevertheless.

simonpj:
> Writing this in "Proposed Change Specification" feels post hoc. Nevertheless, I added a verbal side condition:

Fine thanks,  The Specification for every proposal should be read relative to the language report, and the GHC user manual.  It should not assume familiarity with the implementation.  (We should say this in proposal guidance, Joachim!)

nomeata:
> Fine thanks, The Specification for every proposal should be read relative to the language report, and the GHC user manual. It should not assume familiarity with the implementation. (We should say this in proposal guidance, Joachim!)

Just done in f3f6c7e.

nomeata:
JFTR: I concur with Simon’s proposed acceptance.
(I think I will occasionally stumble over `foo :: a -> a = not` not being polymorphic, but it is consistent with what happens in patterns.)

shayne-fletcher-da:
@int-index This [DAML PR](https://github.com/digital-asset/ghc/pull/24) may be of some use to you. Although it targets 8.8.1-alpha1, it might serve some use as starting point for your implementation.  The `RdrHsSyn.hs` code is somewhat different when targeting `HEAD`. You can get sense of what that looks like by reviewing https://github.com/digital-asset/ghc/blob/da-master/compiler/parser/RdrHsSyn.hs. Hope this helps!

yav:
Overall, I like the idea.

It would be nice to add some examples of how this interacts with declaring infix functions.  For example, I believe the two examples below should be equivalent, correct?
```
f ~> x :: (Int,Int) = (f,x)

(f ~> x) :: (Int,Int) = (f,x)
```

int-index:
> I believe the two examples below should be equivalent, correct?

They are not equivalent today if we omit the type annotation:

```
f ~> x = (f,x)      -- OK
(f ~> x) = (f,x)    -- Parse error in pattern: f ~> x
```

The proposal does not change this.

> It would be nice to add some examples of how this interacts with declaring infix functions.

This problem affects both infix and prefix functions, and the prefix case is currently covered in the "Unresolved Questions" section. 

If the committee decides that we should change the treatment of `(f a) = e` and `(f ~> x) = e`, I shall move it from "Unresolved Questions" to "Proposed Change Specification".

nomeata:
@simonpj, nobody objected to your recommendation to accept. Are there open questions or tweaks to be done?

simonpj:
Thanks for the ping Joachim.  Yes, let's accept this.

Thanks to Vlad for pushing it through.   It fills out a corner case nicely.

I think we'd be open to another proposal addressing the annoyances in "Effects and interactions" but I don't want to tangle that up with accepting this proposal.

Joachim pls merge.

Ericson2314:
@int-index It just occurred to me, what about lambdas?
```
f = \ (x :: a) (y :: a) :: a -> x && y
```

int-index:
@Ericson2314 This does not work syntactically. My intuition for function LHS is based on pattern-matching, so if we have

```
-- definition site
f (x :: a) :: b -> b = ...

-- call site
g = f (True :: Bool) :: Int -> Int
```

then we can find a substitution such that the LHS at the definition site is equal to the RHS. In this case, the substitution is:

* `x :-> True`
* `a :-> Bool`
* `b :-> Int`

This is a bit fast-and-loose because the call site does not necessarily have the explicit type annotations, but hopefully the idea is clear. I don't think the same principle can be applied to a lambda.

Ericson2314:
@int-index Sure, but is it super bad in this regard? The `\` maps to the application as the `f` to the `f``.
```haskell
-- call site
g =  (\(x    :: a)    (y    :: [a])  :: (b   -> b)   -> id)
       (True :: Bool) (True :: Bool) :: (Int -> Int)
```

My motivation a long term goal to "close the gap" between pattern and function bindings; I would like desugaring from the latter to the former to be as lightweight as possible. Then I would propose a `-XNoFunctionBinds` to enshrine the now-usable sub language.

Ericson2314:
I'm also changed my mind on the part of this proposal that changes function binds. I like
```haskell
x (ss :: [a]) :: [a] = reverse @Int ss
```
but not
```haskell
x :: [a] -> [a] = reverse @Int
```
because my intuition is only bindings in *negative* position capture stuff:
```haskell
x (ss :: [a]) :: [a] = reverse @Int ss
f = x [1,2,3,4]
-- =>
f = (\(ss :: [a]) :: [a] -> reverse @Int ss) [1,2,3,4]
-- =>
f = let
  a = Int
  ss :: [a] =  reverse @Int ss
  in  reverse @Int ss :: [a]
```
If there was no `a` bound with the parameter, the second `a` would be uncaptured and over-generalized. I recalls @goldfirere's original reasoning for why it's OK to bind type parameters without screwing up interface had to do with such a `let` intuition.

Ericson2314:
Lastly I am also suspicious of the optional context in 
```haskell
funlhs -> funlhs' :: [context =>] type
```
```haskell
f (a :: at) (b :: bt) :: X bt => (at, bt) = ...
```
look to me like it has a type like
```haskell
f :: concreteAt -> concreteBt -> X concreteBt => (concreteAt, concreteBt)
```
But do we want the higher rank? I rather have some sort of faux-dictionary-patterns so the `=>` can go wherever I want.

Ericson2314:
On a process note, I am sorry to come up with all these qualms *after* the proposal was accepted. I guess my initial reaction was so euphoric (https://github.com/ghc-proposals/ghc-proposals/pull/228#issuecomment-490990818) I didn't think hard enough :). I would be happy to make a PR against the existing proposal instead if it's better to let this thread rest.

goldfirere:
The last few comments look like either a diff against this proposal text or a new proposal entirely. I favor letting this thread rest.

fosskers:
I had voiced opposition to this on Twitter after-the-fact, not having realized that `ScopedTypeVariables` already allows syntax like:
```haskell
foo (n :: Int) = n + 1
```
Having written a lot of Scala in the past, I don't see beauty in this pattern. Having variables marked with their types in the same line becomes harder to visually parse (and enjoy) in production code bases with sufficient complexity. However, since it's already possible, and this Proposal here just extends that paradigm, there isn't much more to say about the matter. It will have to be sufficient that I myself won't choose to write Haskell this way (which in itself is a bit of a C++ problem - multiple syntactic "dialects" within the same language).

Ericson2314:
Wait, @int-index is there no named language extension for this?? Also you have `:: [context =>] type` but I don't think you wish for the result type to be *mandatory*.

Ericson2314:
I would fix both of those in my PR against this, but you rightfully have the honor of picking the name :).

int-index:
> is there no named language extension for this? ... you rightfully have the honor of picking the name

Good point. Does `FunctionResultSignatures` sound good? 

> Also you have :: [context =>] type but I don't think you wish for the result type to be mandatory. :)

Ah yes, I meant to write two productions there:

```
funlhs -> funlhs' :: [context =>] type
        | funlhs'
```

goldfirere:
How about `PatternSignatures`? :)

simonpj:
Vlad, this proposal has two components

1.   Allow a result signature on the LHS of a function declaration
2.  Change the syntax of `funlhs` from `var apat {apat}` to `var {apat}`.

I realise that I do not understand the conequences of (2).  Can you update the proposal to explain, explicitly, the reason for (2)?   

I think that the sentence "This results in an ambiguity with pattern bindings, which is resolved in favor of function bindings." is important here, though the proposal does not explain the ambiguity.   I htink you want
```
  f :: Int -> Int = \x -> x
```
to be a function binding?   Without your change, based on the Haskell Report (which has no signatures in patterns), the above binding would be syntactically illegal.  Is that your point?  Could you make it explicit please?

Moreover, changing from a "pattern binding" to a "function binding" in the Report has some impact: it affects the monomorphism restriction.   Consider
```
   f1 = \x -> x
   f2 :: a -> a = \x -> x
```
`f1` in Haskell today is a pattern binding, and subject to the MR.

`f2` under your proposal would be a function binding, and hence _not_ subject to the MR.

That's a rather unexpected consequence of adding a rsult signature.   At very least the proposal should draw attention to it.   But another alternative would be to redefine a "pattern binding" (for the purposes of the MR) to include a nullary function binding.

int-index:
@simonpj Due to the way the proposal evolved, I started confusing the things as they are defined in the Haskell Report, and as they are implemented in GHC `Parser.y` (which I am more familiar with).

In particular, the change of `funlhs` from `var apat {apat}` to `var {apat}` was motivated by the desire to allow `x :: [a] -> [a] = reverse`, as described in "Effect and Interactions". In #254, @Ericson2314 has convinced me that we should drop this part. But now that you mention that Haskell Report does not define pattern signatures I have realized that this change is also needed to allow `f :: Int -> Int`!

This is hard. The report does not have pattern signatures. But the proposal is meant to be an _extension_ of pattern signatures to function signatures! And for this reason, it is written as an extension to `Haskell2010 + PatternSignatures`.

And I have not thought about the monomorphism restriction at all. I am now even more inclined to simply drop the second component of the proposal, as #254 does.

Let's not support `f :: Int -> Int` as a _function result signature_, let's support it as a _pattern signature_. And allowing `f :: a -> a = reverse` should be a matter of extending `ScopedTypeVariables` support to pattern signatures, rather than reclassifying it as a function binding.

int-index:
There's another misguided piece of prose in the proposal at the moment. I give the following table in "Unresolved Questions":

```
f       = ...  -- FunBind
f a     = ...  -- FunBind
(f a)   = ...  -- parse error
(f a) b = ...  -- FunBind
```

However, it is only true of the GHC implementation. According to the report, `f = ...` is a PatBind, not a FunBind.

I think #254 also resolves this question.

simonpj:
> Let's not support f :: Int -> Int as a function result signature, let's support it as a pattern signature.

Fine.   I think you are proposing to remove (2) in my list above.   Will you update the proposal? 

Ditto your "unresolved questions" point.  This proposal is only about the language, not about GHC.

AntC2:
Um, I'm a lot confused by what's going on here. This proposal is accepted but discussion is still (it seems) making substantive changes. Does that mean the Committee purdah is lifted and the great unwashed can comment further?

Permission or not, I wanted to comment on my recent experience writing a proposal (which applies to 'signatures in patterns' aot), wrt

> @simonpj ..., based on the Haskell Report (which has no signatures in patterns), the above binding would be syntactically illegal. ...
>
> Moreover, changing from a "pattern binding" to a "function binding" in the Report has some impact: it affects the monomorphism restriction. Consider ...
> ...
> This proposal is only about the language, not about GHC.

It is really, really hard to express proposals as a diff at a detailed level. Diff to the Report? Diff to GHC with common/stable extensions? Diff to GHC with some peculiar combination of extensions switched on and off? All of the above?

Simon has taken recently to asking proposers include BNF diffs. But BNF diffs to the Report are almost useless. And the BNF for what GHC currently accepts needs proposers to chase down all sorts of rabbit-holes of what turn out to be extensions not relevant to their proposal. In the end, I included what BNF diffs I was sure of, and appended a bullet list of 'other stuff' for which I couldn't figure out GHC's current syntax.

I'd rate this much more of a barrier to entry than ReST formatting. I might say my proposal stood for a couple of months with some supportive comments but no detailed critique. Only now that it's gone to the Committee have the detailed requests turned up. This is probably not good use of the Committee's time. OTOH there's probably only a handful of people can apply the needed level of quality oversight -- if you're going to require that level of compliance before accepting a proposal.

Then how did this #228 proposal get accepted without reaching the level of compliance?

goldfirere:
What type is inferred here:

```hs
foo ([] :: [a]) :: Num a => String = "hi"
foo (_ :: [a]) :: Show a => String = "bye"
```

Proposed rules around this at https://gitlab.haskell.org/rae/haskell suggest that this be rejected. What do you think should happen?

int-index:
Consider this program:

```
{-# LANGUAGE RankNTypes, ScopedTypeVariables #-}

f ([] :: Show a => [a]) = "hi"
f (_ :: Num a => [a]) = "bye"
```

Here's what GHC 8.8.1 has to say about it:

```
GHCi, version 8.8.1: https://www.haskell.org/ghc/  :? for help
[1 of 1] Compiling Main             ( A.hs, interpreted )
Ok, one module loaded.
*Main> :t +v f
f :: Show a => [a] -> [Char]
```

Weird. Where did `Num` go? Is this a bug? In any case, whatever we do in this example, the same strategy should apply to the result type signature.

goldfirere:
Except that the polarity is different. In your case (which is strange -- see https://gitlab.haskell.org/ghc/ghc/issues/17514), we need to make sure that the inferred argument type is more general than the specific type you're giving to the argument in each equation. In a result signature, we need to have the inferred type more *specific* than the types written in a result signature. I believe this means that, in practice, we'll have to forbid constraints in result signatures when there is more than one equation.

Ericson2314:
@int-index I'd think that would be different? Try pattern matching on the bottom and see if. That makes a difference. (On phone, will explain myself fully later.)

int-index:
That makes sense. I don't see a problem with rejecting constraints in result signatures.

goldfirere:
Do you mean in all result signatures? Or just those with multiple equations? Either way, would you mind updating the proposal? I don't think this requires a full review.

int-index:
> Do you mean in all result signatures?

I do.

> would you mind updating the proposal?

I would prefer if someone with a better understanding of the issue would update the proposal.

simonpj:
I think it's a good idea to forbid quantification and constraints in result type signatures. It's simpler, obviously.   We can always make it more expressive later, but the reverse is harder.

But do we allow
```
f x :: Int -> Ord a => a -> a = blah
```
or 
```
g x :: Int -> forall a. a -> b = blah
```
where the quantification or contstraints are to the right of an arrow?  (For now let's assume that we adopt the proposal about making arrows invariant.)

It would be helpful to have typing rules.  And we have some -- right @rae?

goldfirere:
We have typing rules: https://gitlab.haskell.org/rae/haskell (see link at the bottom). Those rules use the subsumption relation. If that relation changes, then it's conceivable to allow constraints to the right of arrows.

But I would rather not. It's very strange to allow them to the right of arrows but not at the top of a result sig.

simonpj:
> But I would rather not. It's very strange to allow them to the right of arrows but not at the top of a result sig.

I agree with that.

So, what do the rules say?   Do they allow quantification and constraints in result signatures, or not?

goldfirere:
The rules currently allow this quantification, but I will amend them not to. I have amended the proposal.

AntC2:
heh heh just adding a historical note. This 'extension' seems to be already available in Hugs 2006 (in HugsMode)

    Hugs> let {fib (x :: Int) :: Int | x == 0 = 1 | x == 1 = 1  | otherwise = fib (x - 1) + fib (x - 2) } in fib 6
    ===> 13

This doesn't allow constraints in result signatures (nor in pattern signatures).

The [Hugs documentation](https://www.haskell.org/hugs/pages/users_guide/type-annotations.html) (last bullet) has me baffled (beware the warning that the doco is "with some gaps")

* GHC also allows _result type signatures_, where a type signature is attached to the left side of a function definition, but Hugs does not.

As at 2006 it was the other way round(?) Or did this extension used to be in GHC and the Hugs team so liked it they emulated it but failed to document; whereas the GHC team took it out(??)

Ericson2314:
@AntC2 Thanks for bringing that up! Also, until you did, I had forgotten about guards, and didn't realize the result type sig doesn't always come before an `=`.

Ericson2314:
I didn't see a tracking issue, so I made one in https://gitlab.haskell.org/ghc/ghc/issues/18203

PR #229, opened 2019-05-13
Simplify parsing of (~) and (!)

int-index:
We propose to simplify GHC internals and the lexical syntax of Haskell by replacing ad-hoc parsing rules with whitespace-based disambiguation:

```
a ! b = <rhs>  -- Infix (!)
f !a = <rhs>   -- Bang pattern
```

[Rendered](https://github.com/int-index/ghc-proposals/blob/whitespace-bang-patterns/proposals/0000-whitespace-bang-patterns.rst)

### DIFF
--- a/proposals/0000-whitespace-bang-patterns.rst
+++ b/proposals/0000-whitespace-bang-patterns.rst
@@ -142,7 +142,7 @@ Proposed Change Specification
 
 * As a consequence of these rules, ``(- x)`` is now an operator section,
   ``(-x)`` is prefix negation. This change is to be guarded behind a new
-  language extension ``-XPrefixNegation``.
+  language extension ``-XLexicalNegation``.
 
 * Add a new warning, ``-Woperator-whitespace``, disabled by default, that warns
   on prefix, suffix, and tight infix uses of operators that do not have a
@@ -196,10 +196,13 @@ implementation.
 Alternatives
 ------------
 
-If this proposal is rejected, the implementation will need another hand-written
-state machine, which is hard to extend and maintain. This state machine will
-not be able to handle some corner cases which whitespace-based disambiguation
-handles easily.
+* If this proposal is rejected, the implementation will need another hand-written
+  state machine, which is hard to extend and maintain. This state machine will
+  not be able to handle some corner cases which whitespace-based disambiguation
+  handles easily.
+
+* We could classify ``if``, ``do``, ``case``, ``let``, and other keywords as
+  opening.
 
 Unresolved Questions
 --------------------
--- a/proposals/0000-whitespace-bang-patterns.rst
+++ b/proposals/0000-whitespace-bang-patterns.rst
@@ -78,10 +78,10 @@ Proposed Change Specification
 
 * When ``!`` or ``~`` is preceded by whitespace and not followed by
   whitespace, consider it a prefix occurrence, otherwise an infix occurrence.
-* A prefix occurence is treated as bang/lazy pattern in term-level patterns, or
-  as a strictness annotation in types.
-* An infix occurence is treated as an infix operator in terms, or an infix type
-  operator in types.
+* A prefix occurrence is treated as bang/lazy pattern in term-level patterns,
+  or as a strictness annotation in types.
+* An infix occurrence is treated as an infix operator in terms, or an infix
+  type operator in types.
 * For the purposes of these rules, comments are considered whitespace.
 
 Effect and Interactions
--- a/proposals/0000-whitespace-bang-patterns.rst
+++ b/proposals/0000-whitespace-bang-patterns.rst
@@ -195,7 +195,7 @@ proposal does not entail accepting this appendix.
   on prefix, suffix, and tight infix uses of operators that do not have a
   meaning override at the moment. Users who desire forward compatibility may
   enable this warning in case we create new operator meaning overrides in the
-  future.
+  future. Enabled by ``-Weverything`` but not ``-Wall`` or ``-Wcompat``.
 
 * The operator meaning override system has lower precedence than other lexical
   rules that steal operator syntax:
--- a/proposals/0000-whitespace-bang-patterns.rst
+++ b/proposals/0000-whitespace-bang-patterns.rst
@@ -76,12 +76,15 @@ as-patterns, and it works remarkably well in practice.
 Proposed Change Specification
 -----------------------------
 
-* A starting character is an ``$idchar`` (a digit, a letter, or an underscore
+* An opening character is an ``$idchar`` (a digit, a letter, or an underscore
   ``_``), an opening bracket ``(``, ``[``, ``{``, or a quotation mark ``"``,
-  ``'``.
-* In the lexer, when ``!`` or ``~`` is preceded by whitespace or a comment, and
-  followed by a starting character but not a comment, consider it a prefix
-  occurrence, otherwise an infix occurrence.
+  ``'``, excluding the occurrences of ``{`` that start a comment or a pragma.
+* An closing character is an ``$idchar`` (a digit, a letter, or an underscore
+  ``_``), a closing bracket ``)``, ``]``, ``}``, or a quotation mark ``"``,
+  ``'``, excluding the occurrences of ``}`` that end a comment or a pragma.
+* In the lexer, when ``!`` or ``~`` is not preceded by a closing character, and
+  is followed by an opening character, consider it a prefix occurrence,
+  otherwise an infix occurrence.
 * A prefix occurrence is treated as bang/lazy pattern in term-level patterns,
   or as a strictness annotation in types.
 * An infix occurrence is treated as an infix operator in terms, or an infix
@@ -141,3 +144,70 @@ Implementation Plan
 I (Vladislav Zavialov) will implement this change. The idea is to add tokens
 ``BANG`` and ``TILDE`` in addition to ``'!'``, ``'~'``, akin to ``TYPEAPP`` vs
 ``'@'``.
+
+Appendix A: Extended Proposed Change Specification
+--------------------------------------------------
+
+As a basis for a future proposal, here's an outline of a more general
+whitespace-based parsing framework for other operators. Accepting the main
+proposal does not entail accepting this appendix.
+
+* Operator occurrences are classified into four groups, based on preceding and
+  following characters::
+
+    a . b   -- a loose infix occurrence
+    a.b     -- a tight infix occurrence
+    a .b    -- a prefix occurrence
+    a. b    -- a suffix occurrence
+
+  ``a`` and ``b`` stand for closing and opening characters respectively,
+  whitespace stands for all other characters.
+
+* A loose infix occurrence should always be considered an operator. Other types
+  of occurrences may be assigned a special per-operator *meaning override*:
+
+  +-------------------+---------------------+--------------------------------------------+
+  | Operator          | Occurrence          | Meaning override                           |
+  +===================+=====================+============================================+
+  | ``!``, ``~``      | prefix              | strictness annotation in types,            |
+  |                   |                     | bang/lazy pattern in term-level patterns   |
+  +-------------------+---------------------+--------------------------------------------+
+  | ``$``, ``$$``     | prefix              | untyped/typed Template Haskell splice      |
+  +-------------------+---------------------+--------------------------------------------+
+  | ``@``             | prefix              | type application                           |
+  +-------------------+---------------------+--------------------------------------------+
+  | ``@``             | tight infix, suffix | as-pattern                                 |
+  +-------------------+---------------------+--------------------------------------------+
+  | ``-``             | prefix              | negation                                   |
+  +-------------------+---------------------+--------------------------------------------+
+
+  This wouldn't be a backward compatible change in every corner case, but the
+  migration path does not require ``-XCPP``.
+
+* As a consequence of these rules, ``@`` (loose infix) and ``~`` (suffix, loose
+  infix, tight infix) are now proper infix operators.
+
+* As a consequence of these rules, ``(- x)`` is now an operator section,
+  ``(-x)`` is infix negation. This change is to be guarded behind a new
+  language extension ``-XPrefixNegation``.
+
+* Add a new warning, ``-Woperator-whitespace``, disabled by default, that warns
+  on prefix, suffix, and tight infix uses of operators that do not have a
+  meaning override at the moment. Users who desire forward compatibility may
+  enable this warning in case we create new operator meaning overrides in the
+  future.
+
+* The operator meaning override system has lower precedence than other lexical
+  rules that steal operator syntax:
+
+  * ``#`` under ``-XMagicHash`` or ``-XOverloadedLabels``
+  * ``?`` under ``-XImplicitParams``
+  * ``.`` as module qualification
+
+  We choose not to handle these cases under the new framework because their
+  rules do not apply to arbitrary subexpressions:
+
+  * ``(f x)#`` is not a proper use of magic hash
+  * ``(f x).id`` is not proper module qualification
+  * ``?(f x)`` is not an implicit parameter
+  * ``#(f x)`` is not an overloaded label
--- a/proposals/0000-whitespace-bang-patterns.rst
+++ b/proposals/0000-whitespace-bang-patterns.rst
@@ -188,7 +188,7 @@ proposal does not entail accepting this appendix.
   infix, tight infix) are now proper infix operators.
 
 * As a consequence of these rules, ``(- x)`` is now an operator section,
-  ``(-x)`` is infix negation. This change is to be guarded behind a new
+  ``(-x)`` is prefix negation. This change is to be guarded behind a new
   language extension ``-XPrefixNegation``.
 
 * Add a new warning, ``-Woperator-whitespace``, disabled by default, that warns
--- a/proposals/0000-whitespace-bang-patterns.rst
+++ b/proposals/0000-whitespace-bang-patterns.rst
@@ -103,7 +103,8 @@ accepted::
   x !y = x == y         -- standard interpretation: infix operator (!)
   data T = MkT ! Int    -- standard interpretation: strict field !Int
 
-The migration strategy is to adjust whitespace::
+This may break existing programs. The migration strategy is to adjust
+whitespace::
 
   f ~a ~b = <rhs>
   x ! y = x == y
--- a/proposals/0000-whitespace-bang-patterns.rst
+++ b/proposals/0000-whitespace-bang-patterns.rst
@@ -144,6 +144,10 @@ Proposed Change Specification
   ``(-x)`` is prefix negation. This change is to be guarded behind a new
   language extension ``-XLexicalNegation``.
 
+* Under ``-XLexicalNegation``, prefix ``-`` binds tighter than any infix
+  operator, so that ``-a % b`` is parsed as ``(-a) % b`` regardless of the
+  fixity of ``b``.
+
 * Add a new warning, ``-Woperator-whitespace``, disabled by default, that warns
   on prefix, suffix, and tight infix uses of operators that do not have a
   meaning override at the moment. Users who desire forward compatibility may
--- a/proposals/0000-whitespace-bang-patterns.rst
+++ b/proposals/0000-whitespace-bang-patterns.rst
@@ -87,17 +87,7 @@ Proposed Change Specification
 Effect and Interactions
 -----------------------
 
-It is a slight deviation from the standard which dictates the following to be
-interpreted as irrefutable patterns::
-
-  f ~ a ~ b = <rhs>
-
-The migration strategy is to remove unnecessary whitespace::
-
-  f ~a ~b = <rhs>
-
-
-The users also regain the ability to define infix ``(!)`` even when
+The users regain the ability to define infix ``(!)`` even when
 ``-XBangPatterns`` are enabled::
 
   {-# LANGUAGE BangPatterns #-}
@@ -105,6 +95,19 @@ The users also regain the ability to define infix ``(!)`` even when
 
 Costs and Drawbacks
 -------------------
+
+It is a slight deviation from the standard which dictates the following to be
+accepted::
+
+  f ~ a ~ b = <rhs>     -- standard interpretation: lazy (irrefutable) patterns
+  x !y = x == y         -- standard interpretation: infix operator (!)
+  data T = MkT ! Int    -- standard interpretation: strict field !Int
+
+The migration strategy is to adjust whitespace::
+
+  f ~a ~b = <rhs>
+  x ! y = x == y
+  data T = MkT !Int
 
 This already matches the style of most Haskell users and will simplify the
 implementation.
--- a/proposals/0000-whitespace-bang-patterns.rst
+++ b/proposals/0000-whitespace-bang-patterns.rst
@@ -1,0 +1,119 @@
+Simplify parsing of (~) and (!)
+===============================
+
+.. proposal-number::
+.. ticket-url::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/229>`_.
+.. sectnum::
+.. contents::
+
+We propose to simplify GHC internals and the lexical syntax of Haskell by
+replacing ad-hoc parsing rules with whitespace-based disambiguation::
+
+  a ! b = <rhs>  -- Infix (!)
+  f !a = <rhs>   -- Bang pattern
+
+
+Motivation
+------------
+
+At the moment, Haskell has inconsistent and complicated rules to decide whether
+a ``~`` or a ``!`` is used as an infix operator or as a prefix annotation. It
+poses a significant burden on the parser implementation without any benefits
+to the users.
+
+We have six cases to consider:
+
+* ``!`` as a bang pattern
+* ``~`` as a lazy (irrefutable) pattern
+* ``!`` as a strictness annotation in type declarations
+* ``~`` as a strictness annotation in type declarations
+* ``!`` as an infix operator
+* ``~`` as an infix operator
+
+GHC puts a great deal of effort into distinguishing these cases, and still does
+a poor job at it (for example, see `#1087
+<https://gitlab.haskell.org/ghc/ghc/issues/1087>`_). Here are the rules we have as of GHC 8.8:
+
+* In term-level patterns, ``!`` is considered either an infix operator or a
+  bang pattern, depending on the module-wide ``-XBangPatterns`` flag.
+* In term-level patterns, ``~`` is always considered a lazy (irrefutable) pattern.
+* In term-level expressions, ``!`` is always considered an infix operator.
+* In term-level expressions, ``~`` is never valid.
+* In types, ``!`` and ``~`` are considered type operators or if they have no LHS, or strictness
+  annotations otherwise.
+
+Here are the consequences of these rules:
+
+* We cannot write both of these definitions in the same module::
+
+    a ! b = <rhs>  -- Infix (!)
+    f !a = <rhs>   -- Bang pattern
+
+* One of these definitions is valid, the other is not::
+
+    ~a + ~b = <rhs>   -- Valid
+    !a + !b = <rhs>   -- Invalid (#1087)
+
+* We cannot use ``~`` as a term-level operator.
+
+The implementation of the current rules is not pretty. In terms, we always
+parse ``!`` as an infix operator and rejig later, and it doesn't cover all
+possible cases. In types, we have a hand-written state machine to detect the
+"no LHS" -- it is more robust but difficult to maintain.
+
+In order to unify term-level and type-level parsers, a milestone towards
+Dependent Haskell, we will need to parse the ``(~)`` operator in terms, but it
+will further complicate parser implementation under the current arrangement.
+
+Users typically write bang patterns and lazy (irrefutable) patterns without
+whitespace after it, and we can make use of this information during parsing. A
+similar rule is used to distinguish between ``@`` in type applications and
+as-patterns, and it works remarkably well in practice.
+
+Proposed Change Specification
+-----------------------------
+
+* When ``!`` or ``~`` is preceded by whitespace and not followed by
+  whitespace, consider it a prefix occurrence, otherwise an infix occurrence.
+* A prefix occurence is treated as bang/lazy pattern in term-level patterns, or
+  as a strictness annotation in types.
+* An infix occurence is treated as an infix operator in terms, or an infix type
+  operator in types.
+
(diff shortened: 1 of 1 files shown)


AntC2:
It's always fascinating to read these proposals, to discover there are extensions I've never heard of.

* `~` as a strictness annotation in type declarations

Is it? What extension is that? Oh, it's a laziness annotation when you've switched on [Strict-by-default Datatypes](https://downloads.haskell.org/~ghc/8.6.5/docs/html/users_guide/glasgow_exts.html#strict-by-default-data-types).

I note that syntactically it behaves as a unary prefix, as it is for irrefutable patterns.

I don't really think of `~` in constraints as being an operator: it's special syntax; just as `->` is special syntax. Does it make any sense to treat it as an operator: can it be partially-applied? can it be passed as an argument to a higher-order type function?

`~` in the Report is a `<reservedop>`. Do we really want to make it available as an infix operator in terms? The only sensible use would be for something parallel to its role in Equality constraints. What would that be?

If `~` is to appear in terms, we might consider making it available as a unary prefix -- same syntax rules as prefix `-`/`negate`.

int-index:
> Does it make any sense to treat it as an operator: can it be partially-applied? can it be passed as an argument to a higher-order type function?

Yes, both of this is possible for `(~)`. There's no reason to have it as special syntax, for example its cousin `~~` is a regular class: https://hackage.haskell.org/package/base-4.12.0.0/docs/Data-Type-Equality.html#t:-126--126-

> The only sensible use would be for something parallel to its role in Equality constraints. What would that be?

With Dependent Haskell, or even VDQ alone, any type could appear in terms, including equality constraints. Imagine I defined

```
data Dict c where
  MkDict :: forall c -> c => Dict c
```

Note the `forall c ->` syntax, this is [VDQ] which is only available in types at the moment but will come to terms.

[VDQ]: https://ryanglscott.github.io/2019/03/15/visible-dependent-quantification-in-haskell/

Then one could write something like this:

```
string_is_list_of_char_proof = MkDict (String ~ [Char])
```

Now, `(String ~ [Char])` is still a type, but we can't tell before typechecking, so the parser and the renamer will have to assume it's a term. That's why we will need to bring all type-level syntax to terms, eventually.

> same syntax rules as prefix `-/negate`.

Frankly, I'm not a fan of these rules. I would even consider making them whitespace-sensitive in the same way this proposal treats `!` and `~`. Many a beginner have been bitten by this:

```
sqrt -x    -- oops, it's parsed as (sqrt - x)
```

goldfirere:
GHC currently supports at least two whitespace-sensitive operators: `-XTypeApplications`' `@` and `-XTemplateHaskell`'s `$`. But, perversely, the rules for each are different.

`$` decides what it means based on whether or not it is followed by either an identifier or an open-paren, with no whitespace. Examples:

```
x$y   -- $y is a splice
x$ y  -- $ is an infix operator
x $ y -- $ is an infix operator
x $y  -- $y is a splice
x $(y)  -- $(y) is a splice
x $[y]  -- $ is an infix operator
x ${- hi -}y  -- $ is an infix operator
```

`@` decides what it means based on whether or not the preceding character can plausibly end an identifier. Examples:

```
x@y    -- as-pattern
(x)@y   -- type application
x@ y   -- as-pattern
x @y   -- type application
x{- hi -}@y  -- type application
x@{- hi -}y  -- as-pattern
```

This proposal suggests to add `!` and `~` into the mix. This kind of disambiguation is really useful, and I generally support it. But let's bring more consistency to the whole affair. Can we unify the treatment of all of these operators?

Also, the proposal discusses whitespace. Should a comment be understood to be whitespace? Or is it ignored? (I prefer the former, but the proposal should specify.)

simonpj:
> This kind of disambiguation is really useful, and I generally support it. 

I agree.  White space is very significant to humans, and it wastes information bandwidth not to take advantage of it.

Richard has good points about `$` and `@`.

AntC2:
> White space is very significant to humans,

Heh, heh. So do we grasp the livewire of white space around `.`? I remember the uproar around Simon's 'The Power of the dot'. And Haskell's inconsistencies have now gotten more ossified with all the code using tight-binding `.` for nested lenses.

    Prelude.length         -- module qualifier
    double.length          -- function composition
    Just .length           -- function composition
    Just. length           -- function composition
    Just . length          -- function composition
    Just.length            -- module qualifier, but no such module `Just'

    shape.centre.xcoord    -- function composition (a nested lens)
    screenobj.shape.centre.xcoord   -- function composition, but probably ill-typed because screenobj is not a lens
    screenobj.^shape.centre.xcoord  -- probably what was meant,
                                    --  all the dots are function composition except .^ 

I always put space both sides of `.` as function composition, and I hate seeing it inline when it's not module qual. I plain don't get the Lensists claim that they're trying to look like OOP: the precise place where you'd go `object.method` or `struct.component` is what Lenses don't do.

goldfirere:
Ah yes, I had forgotten poor old `.`. It, too, is whitespace-sensitive, and with yet different rules.

int-index:
>  Should a comment be understood to be whitespace?

Yes, I think so, – added it to the proposal.

> Richard has good points about `$` and `@`.

Indeed. But there's a lot of existing whitespace-sensitive syntax to consider:

* `$` and `$$`: TH splices
* `@`: as-patterns or type applications
* `.`: module qualification
* `#`: overloaded labels
* `?`: implicit parameters

Unified rules are possible, but they might cause too much breakage. I don't want to get too ambitious with the current proposal. Let's do one thing at a time.

int-index:
For the sake of a future proposal, I think we should define the unified rules in the following terms:

```
a . b -- a loose infix occurrence
a.b   -- a tight infix occurrence
a .b  -- a prefix occurrence
a. b  -- a suffix occurrence
```

A loose infix occurrence should always be considered an operator. (This frees up `@` as an operator). Other types of occurrences may have special per-operator meaning, or produce a warning when there isn't any special meaning assigned at the moment. This gives us a wealth of new syntax for future language extensions.

AntC2:
Thanks that's useful terminology. It doesn't fully explain what happens with `.`:

    A.b      -- module qualifier, because upper-case A
    a.b      -- function composition
    A.B.c.d  -- both together
    A.B.c.D  -- also both together

That's why I always write composition as "loose infix".

Bj0rnen:
@AntC2 So I think the idea is that the last three would raise a warning. I could see two approaches for GHC to reach that conclusion:
1. `.` only has a special meaning in a tight infix occurrence when the preceding lexical term is a capitalized identifier. So the special "rule" doesn't trigger and there is a warning because it wasn't a loose infix occurrence.
2. The special `rule` does trigger in any case of a tight infix occurrence of `.`. But the rule itself says to do different things depending on the capitalization of the preceding identifier. So if that isn't capitalized, it's treated as an infix term-level operator, plus a warning is emitted.

It depends on the responsibility/power of these "rules" and what can be used as a criteria for whether a "rule" triggers. I don't know if this is a rabbit hole worth pursuing in this ticket. But I think that I generally like the idea of warning on/eventually freeing up those ambiguous-looking infix operator placements.

Although, it would mean that expressions like this would be discouraged/eventually even stop compiling:
```haskell
i :: Integer
i = 1 + 2^3
```
So, you can't use whitespace to indicate tighter binding for readability. But you can still use parentheses for that.

goldfirere:
Though I see its appeal, I'm against warnings for omitting whitespace for operators, in general. This will break gobs and gobs of code (knowing that many Haskellers consider a warning to be breakage), and it doesn't seem quite necessary.

One new realization here: this proposal means that GHC would reject (more) Haskell98 programs, given that `x !y = x + y` is valid Haskell98, but would be rejected under this proposal (presumably with advice to enable `-XBangPatterns`). Similarly, `data T = MkT ! Int` would be rejected, even though that, too, is Haskell98. This should definitely be listed as a drawback, along with the fact that existing programs may break.

int-index:
> This should definitely be listed as a drawback, along with the fact that existing programs may break.

Done, @goldfirere. In theory, we could implement `-Wcompat` warnings for this and wait for a couple of releases, but this sounds rather pointless given that the migration strategy is to so simple and requires no conditional compilation.

AntC2:
BTW, operator `(!)` for factorial is the standard example with [`-XPostfixOperators`](https://downloads.haskell.org/~ghc/8.6.5/docs/html/users_guide/glasgow_exts.html#postfix-operators).

For any syntax change you make, be careful it doesn't invalidate the example in the Users Guide. That shows `(e !)`, which is a "loose infix" usage, I guess(?). Currently "tight infix/suffix" also works. The section parens are required, so I guess it's moot.

    (!) :: (Ord a, Num a) => a -> a
    (!) 0 = 1
    (!) n | n > 0 = n * ((n - 1)!)

int-index:
Note that in the current proposal, we treat `!` as a bang pattern only if it's both *preceded by whitespace* and *not followed by whitespace*, in other words, a prefix occurrence. A suffix or a tight/loose infix occurrence is considered an operator.

`-XPostfixOperators` continue to work.

int-index:
Ah, I see now, `(e !)` is indeed a case when this rule fires, and we would have to consider it an invalid bang pattern. I will try to think of a better rule.

int-index:
I glanced over `Lexer.x` and `Parser.y` and formulated more precise rules. Now `(e !)` is an infix `!` as intended.

jvanbruegge:
I support this
> Many a beginner have been bitten by this `sqrt -x    -- oops, it's parsed as (sqrt - x)`

I wouldn't describe myself as beginner, but I've been bitten by minus parsing too
> For the sake of a future proposal, I think we should define the unified rules in the following terms: loose/tight infix, prefix/suffix occurence

THIS. So much THIS. I always put spaces before and after infix operators and rules/warnings would surely improve the situation here. It would also make a future record access syntax `foo.label` possible.

> Although, it would mean that expressions like this would be discouraged/eventually even stop compiling: `i = 1 + 2^3`

One possibility would be to add a pragma, that basically hints that this operator is intended to also be used as a tight infix

goldfirere:
> One possibility would be to add a pragma, that basically hints that this operator is intended to also be used as a tight infix

But what operators should get that pragma? Just about every operator binds tighter than *something*.

Separately, I personally would have a hard time supporting this proposal unless it addresses at least `$` and `@` as well, with unified rules for all of them. (Or reasons that unified rules cannot work.) I would be thrilled if we could include `.` in the mix, but that may be harder.

Ericson2314:
@goldfirere I feel like this proposal at least moves things in the direction of unified rules for all of them—we can just exclude programs like `p@(..)`  until the rules are all the same. I don't mind taking baby steps as long as the trajectory is good. Are you worried about the good sapping the momentum of the better?

goldfirere:
> I feel like this proposal at least moves things in the direction of unified rules for all of them

How so? There is a framework, but I see no suggestions for how this framework should apply to `$` and `@`. Maybe it's obvious -- to be honest, I haven't given the details much thought. But even if this proposal does not actually implement the changes, seeing what they would be would be helpful.

Ericson2314:
That's fair. It's not obvious, and perhaps the proposal should spell it out the a chain of restrictions even if only one step is being taken to make it clear.

To quickly through together a straw-man convergence point:

1. When an token is both an infix operator and something else, the operator form must be loose infix. So that means only `foo $ bar`, `foo . bar`, `foo ~ bar`, and `foo - bar`. `foo @ bar` too if `@` gets allowed as an infix operator.

2. When a token acts as a prefix operator, or prefix "meta operator" (special syntactic form not user code) it must be in prefix form (preceded by a space, and followed by the character class from this proposal (`$idchar` (a digit, a letter, or an underscore `_`), an opening bracket `(`, `[`, `{`, or a quotation mark `"`, `'`). That would mean `foo !bar`, `foo ~bar`, `foo $templateBar`, `foo $(templateBar templateBaz)`, `@typeBar` or `-numberBar`.

3. When a token acts as a meta infix operator, it must not be white-space on either side, so only `moduleFoo.bar` or `patternFoo@bar`.

AntC2:
> @Ericson2314 the operator form must be loose infix. So that means only ... `foo . bar`, ...

This I love. This I have always wanted for operator ` . `. This will cause a riot just as soon as the Lensists find out about it.

A couple of your other examples aren't quite right:

> `moduleFoo.bar` or `patternFoo@bar`

Module names and pattern names must begin upper case. Specifically for modules, that's how to tell a module prefix vs "tight infix" function composition.

Ericson2314:
@AntC2

> `moduleFoo.bar` or `patternFoo@bar`

Fixed those. `patternBar` (what the `@` RHS now says) is supposed to be any pattern, not an identifier for a pattern synonym.

AntC2:
For the record with ` . `, previous discussions beware

* `Prelude..`  -- module qualification of operator `.`, so this is legal

        Just Prelude..foo $ bar

   (I find that unreadable.)

* `[5..7]` ==> `[5,6,7]` -- that `..` can be "tight" or "loose" infix or "prefix" or "suffix", using @int-index's terms.

One of the earlier suggestions was

* Loose infix `foo . bar` is always function composition (as currently) -- or rather always whatever's in scope for operator `.`.
* Tight infix with upper-case to left `Foo.bar` is always module qual (as currently)
* Tight infix with lower-case to left `foo.bar` is a distinct operator spelled `.` but somehow distinguished, again whatever's in scope (which might be the same as function composition, to keep the Lensists happy).
* Tight-`.` could optionally have an infix binding tighter than function application and optionally a different association (however we declare that).

        infixl 11 ?.?                -- assuming function applic is precedence 10

        obj.subobj.meth baz          -- parses as

        ((obj ?.? subobj) ?.? meth) baz

   Now Lensists could go

        struct.lens baz              -- to mean struct.^lens applied to baz

   Or traditional recordists could go (this is the Type-Directed Name Resolution proposal, so should fit with `DuplicateRecordFields` resolution)

        datatype.field baz           -- to mean datatype's field applied to baz
                                     -- i.e. x.f to mean (f x)
                                     -- then we get virtual fields: person.fullname
                                     --                           = person.first ++ " " ++ person.last

Ericson2314:
`[Prelude..foo]` parses as `[(Prelude.) .foo]`, but that's another rule: I'll amend my post above to require` .. ` too.

int-index:
> I personally would have a hard time supporting this proposal unless it addresses at least `$` and `@` as well, with unified rules for all of them.

@goldfirere I added an appendix that handles `$`, `$$`, `@`, and `-`. I want to keep the limited scope of the current proposal, but the committee may decide to accept the appendix as well.

Does this look reasonable? If so, I'd like to submit to the committee now.

goldfirere:
Yes, I'm happy now. Thanks!

int-index:
Great! I submit this to the committee, then, @nomeata

goldfirere:

**Summary**

This proposal describes uniform lexical rules for the understanding of operators. It makes whitespace significant in lexing operators, treating, e.g., `x !y` to have a bang pattern but `x!y` and `x ! y` to be uses of an infix `!`. The proposal covers `~` similarly, while opening it up as a possible infix operator in terms. The proposed change is *not* guarded by an extension flag, and it will reject some Haskell98 programs, such as `data T = MkT ! Int`.

The proposal includes an appendix, which extends the uniform treatment to Template Haskell's `$` and `$$`, type application's `@`, and numeric negation's `-`. (The new treatment of `-` is guarded behind a new `-XPrefixNegation`.)

This appendix also includes a new warning, `operator-whitespace`, which warns whenever an operator is used without whitespace surrounding it. The rationale of this warning is that it allows Haskellers to avoid future syntax stealing.

The committee is encouraged to read the specification in the proposal and appendix. These specifications are nicely terse.

**Recommendation**

As the shepherd of this proposal, I recommend acceptance, but only if we also accept the appendix.

Why accept?

* There is a good deal of delicate code in the parser today to deal with the strange behavior of `!` and `~`.

* This code is buggy and requires yet more intricacy to fix.

* The description in this proposal is clean, and it should echo common usage patterns.

Why require the appendix?

* GHC already has too many rules to remember. Adopting the appendix is a *simplification* of the specification of existing features.

* Both the main proposal and the appendix are breaking changes. (These changes are easy to accommodate in a backward-compatible way; just add or remove some spaces.) It's better to lump similar breaking changes together than do them piecemeal.

* This will set a standard for future attempts to steal operators.

I am lukewarm on the addition of a warning for use of operators without spaces. I see the appeal, but I worry that Haskellers will take the warning too seriously. I would be against putting it in even `-Wall`, but having it in `-Weverything` may not be so bad.

As usual, silence is agreement. Ping @ghc-proposals/ghc-steering-committee

nomeata:
I am fine with the clean-up about `~` and `!`, which doesn’t affect code that is not using these symbols.

I am on the fence about the merits of `-Woperator-whitespace`, as that would step on the toes of every programmer out there who likes to use their operators in a tight infix form. Maybe `-Weverthing` is fine for it, though.

simonpj:
I support the proposal.  White space is extremely significant for humans, and ignoring it forces us into clumsy circumlocution.  Let's celebrate it!

The proposal explcitly does not cover "." becuse the taxonomy of "tight infix" or "prefix" etc occurrences appllies to arbitrary expressions.   But distinguishing record field selection , "e.f" (where f is a field name (not arbitrary expression)  from infix function composition  "e . f" is another use of the same "white space is important for humans" principle.  This proposal leads us in that direction, whcih I'm happy about.

Gaining "~" and "@" as ordinary operators is a nice bonus.

simonmar:
Is it possible to define this extension as a modification to the lexical syntax of Haskell, without introducing ad-hoc side conditions (things like "is not preceded by....")?

int-index:
> Is it possible to define this extension as a modification to the lexical syntax of Haskell

Not in a way that I would know of. The specification language of the report is not context-sensitive. Consider `(!a)` vs `(!)` and `x!y`: the first one is a bang pattern, the other two are infix operators. The sequences of non-terminals we want to generate are:

1. for `(!a)`: `(`  `bang` `varid`  `)`
2. for `(!)`: `(`  `varsym`  `)`
3. for `x!y`: `varid`, `bang`, `varid`

How do we choose between `bang` and `varsym`? The most straightforward way is to look at the preceding and following tokens.

That said, now I realize that changing the language of the proposal to talk about preceding and following _tokens_ instead of _characters_ would be an improvement.

int-index:
>  I recommend acceptance, but only if we also accept the appendix.

I have merged the appendix into the main "Proposal Change Specification".

> Is it possible to define this extension as a modification to the lexical syntax of Haskell

I have also rewritten the specification in terms of lexical non-terminals defined in the Haskell 2010 report instead of individual characters. Unless I made a mistake, the proposed change is the same.

However, it is still a side condition for reinterpretation of unqualified `varsym`.

Ericson2314:
Small nit. `-XPrefixNegation` strikes me as a slightly confusing name since `-expr` (no space) is negation either way. Also, all things equal rather have the `-XNo*` case being that one that makes the language more ad-hoc or forgiving. Something like `-XNoLooseNegationTightSubtraction` or `-XNoContextualNegation` (with the default being `-XLooseNegationNoLooseNegation` / `-XContextualNegation`) avoids both of those a feels a bit better.

int-index:
> since `-expr` (no space) is negation either way

I don't think that's the case? At the moment, `f -x` is subtraction, same as `f - x`. With `-XPrefixNegation` it would be negation, same as `f (-x)`.

Ericson2314:
Err i mean `-x` not `f -x`. We always had some prefix negation, but with this extension no we loose prefix subtraction and loose infix negation. Per your terms, "prefix" is about the lex context rather than parse context, whereas the current rules today disambiguate in the parser. `-XPrefixNegation` sounds like it is changing the lex rules or being more permissive, rather than switching from parse to lex side-conditions wholesale and mainly being more restrictive (many previously accepted programs and now rejected, but some are instead accepted differently).

Ericson2314:
I guess I am also confused by the extent of the changes about `-`. Do we have two separate `-` tokens in the lexer? Or perhaps we are still disambiguating in the parser? If the latter. we also have cases where (in the current parse) there is no previous token---the "true" prefix/suffix cases. How are things handled in that case?
```haskell
foo :: Int -> Int
foo = - 1
fooAlt :: Int
fooAlt = - 1
bar :: Int
bar = (-) -1
baz :: Int
baz = 1- -1
```

int-index:
> Err i mean `-x` not `f -x`

`-x` is a substring of `f -x`, but I get what you're saying. `-XLexicalNegation`? Note that there's already an extension with a similar effect, `-XNegativeLiterals`, which makes `f -1` interpreted as `f (-1)`, but has no effect on `f -x`. What about `-XNegativeAtoms`, where "atoms" refers to atomic expressions? (`-XAnions`? Jk.)

> Do we have two separate `-` tokens in the lexer? 

Yes.

> we also have cases where (in the current parse) there is no previous token---the "true" prefix/suffix cases.

That's only at the start of the file and at the end of the file, and I wouldn't be worried about it. the spec defines three token groups: opening, closing, and other; just toss the lack of a preceding/following token into the "other" group.

Now, to elaborate on your examples.

1. `foo = - 1` and `fooAlt = - 1`: that's a parse error just like `foo = + 1` is a parse error today:
    ```
    ghci> foo = + 1
    <interactive>:1:7: error: parse error on input ‘+’
    ```
    The type signature does not have any effect.
2. `bar = (-) -1`: this is parsed as `bar = (-) (-1)`:
    ```
    ghci> bar = (-) (-1)
    ghci> :t bar
    bar :: Num a => a -> a
    ```
3. `baz = 1- -1`: since we have no meaning override for a suffix occurrence of `-`, this is parsed as `baz = 1 - (-1)`:
    ```
    ghci> baz = 1 - (-1)
    ghci> baz
    2
    ```

yav:
@int-index shouldn't some keywords (e..g, `if`, `do`, `case`, `let`) also be considered "opening"?

int-index:
@yav Hm, what do what do we think of these?

1.  `10 -let a = 5 in a+a`
2. `even -let a = 5 in a+a`

Today (1) computes `0`, (2) is a parse error. If we classify `let` as opening, then (1) is a parse error, (2) computes `True`. Neither looks terribly compelling.

It's a judgement call how to classify these keywords. I lean towards not classifying them as opening, but I'm happy to leave the final decision to the committee.

int-index:
By the way, it seems that GHC diverges from the report at the moment. This is the grammar defined in the report:

```
infixexp -> lexp qop infixexp      (infix operator application)
          | - infixexp             (prefix negation)
          | lexp
```

If I'm reading this correctly, it means `- 1 + 2` should be parsed as `- (1 + 2)` and evaluate to `-3`, but GHC parses it as `(-1) + 2` and computes `1`.

That's because the GHC grammar defines `'-' fexp`, not `'-' infixexp`. Changing it to `'-' infixexp` actually makes the amount of shift/reduce conflicts drop from 236 to 179!

Is this an oversight? Should I fix?

int-index:
> Is this an oversight? Should I fix?

Apparently, the report also gives an example that is consistent with GHC behavior:

```
- f x + y     parses as    (- (f x)) + y
```

There's more to this than the grammar, it appears, infix operators also have precedence in the report: "Negation is the only prefix operator in Haskell; it has the same precedence as the infix - operator defined in the Prelude".

And of course, GHC gets it right when I use an operator that binds more tightly:

```
ghci> infixr 9 .+; (.+) = (+)
ghci> - 1 .+ 2
-3
```

Therefore, I withdraw my last comment. GHC is way smarter than me, it appears 🙂

Ericson2314:
@int-index I like `LexicalNegation`. That gets at the heart of the matter. If people think the alternative is some god awful "dynamic negation" so be it!

yav:
@int-index That's a good point, I certainly like the current behavior better.    I was thinking of expressions like `1+if x then 2 else 3`, though. If `if` is not an opener, would `+` be treated as suffix?

I guess this would be the same as `1+ if x then 2 else 3`, but I am not very clear if we are proposing to actually add suffix operators, or the classification is only for the meaning override, and if there is no override, then the operator is treated as infix.

int-index:
> I am not very clear if we are proposing to actually add suffix operators, or the classification is only for the meaning override, and if there is no override, then the operator is treated as infix.

It's the latter. We may introduce a meaning override for an operator, but short of that, it is treated as infix. There's no meaning override for a suffix occurrence of `+`, so both of `1+if x then 2 else 3` and `1+ if x then 2 else 3` continue to work as today (with a warning if the user has enabled `-Woperator-whitespace`, disabled by default).

simonmar:
I have mixed feelings about this proposal. Yes there are warts in the current design, but there are things about the proposed design that bother me:

1. It's not clear to me that it's a simplification. Having to classify every token into 3 buckets, and classifying occurrences of certain operators based on whitespace-sensitive rules feels more complex than the current situation.

2. Personally I don't like to use whitespace sensitivity in this way. It worries me that `a@b` and `a @ b` would be different, but `a+b` and `a + b` would not be, for example.

3. If implementing this change in the Haskell report would not be possible without ad-hoc side conditions, that should give us pause, no?

int-index:
> It's not clear to me that it's a simplification. Having to classify every token into 3 buckets, and classifying occurrences of certain operators based on whitespace-sensitive rules feels more complex than the current situation.

From the implementation perspective, it should be a simplification. The lexer will take a complexity hit, but simplifications in the parser will make up for it.

From the user's perspective, it is less about complexity and more about intuition. And as a user, I definitely consider the current system unintuitive:

```
ghci> -1 + 2
1
ghci> 2 + -1

<interactive>:3:1: error:
    Precedence parsing error
        cannot mix ‘+’ [infixl 6] and prefix `-' [infixl 6] in the same infix expression
```

For the reasons of parsing, `+` does not appear commutative. That's a shame.

> Personally I don't like to use whitespace sensitivity in this way. It worries me that `a@b` and `a @ b` would be different, but `a+b` and `a + b` would not be, for example.

That's already the case with `-XTypeApplications`. Consider these patterns:

1. `f id@x = ...`
2. `f id @x = ...`
3. `f id @ x = ...`

(1) is an as-pattern, (2) and (3) bind a type variable (assuming [GHC Proposal 50](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0050-type-lambda.rst) is implemented).

Admittedly, this proposal takes it a bit further, so that (1) is an as-pattern, (2) binds a type variable, and (3) defines the `@` operator.

However, I think this context-sensitivity is a fair price to pay. After all, I definitely expect this to be an operator definition, not a bang pattern:

```
a ! b = ...
```

Unfortunately, at the moment enabling `-XBangPatterns` turns it into `a (!b) = ...`. Whitespace is important to humans, I just want to find a good way to make use of it in the parser.

> If implementing this change in the Haskell report would not be possible without ad-hoc side conditions, that should give us pause, no?

That is indeed unfortunate. But the specification language of the current Haskell report is not equipped to deal with this: the reason Haskell 2010 supports `~` in patterns is that `~` is not a valid operator in expressions! Imagine we wanted to implement `-XBangPatterns` in the report, would that steal the `!` operator from the users? I believe that whitespace is how a person can easily tell a bang pattern from an infix operator:

```
f !a = ...
a ! b = ...
```

That's why I believe the parser should use the same principle.

----

@simonmar You make good points, especially about complexity. I've got carried away with the specification, not realizing that classifying tokens into three buckets is a major source of complexity.

I'd be happy to hear ideas how to simplify the specification if it continues to work in the same spirit (whitespace-sensitivity) and addresses all the listed use cases.

nomeata:
Richard, now that POPL papers have been submitted: Can you move this forward?

goldfirere:
Yes, let's get this settled, one way or another.

@simonmar, you've been the loudest detractor, and you've made some good points. My enthusiasm for this proposal has diminished a bit because of your points, but I still favor it. What about others?

An important point here for me is that we've dug ourselves into this hole. That is, the current treatment of several operators (`!`, `~`, and `@`) is painful without this proposal. (In particular, the first two.) It's unclear whether less than a Herculean effort will fix all the problems with the first two. (It's essentially writing a parsing state machine by hand, and then powering the machine with coal-fired steam.) This proposal cleans all that up quite nicely, and makes a convenient avenue for future extensions. If we could rewind time before those troublesome operators, I might not be in favor, as whitespace-sensitivity can be surprising. But I think that, given where we are, we should follow it all to its natural conclusion: this proposal.

yav:
I don't have a strong opinion one way or another.   While the current system is a little clunky I don't think it has really been a big source of difficulty, so I am not sure we need to fix it.  OTOH, the approach in the proposal could open the door to future extensions (e.g., postfix poerators?), although I am not really sure that we'd want these in the language...

int-index:
> While the current system is a little clunky I don't think it has really been a big source of difficulty

Depends on what counts as big, it has definitely cost me weeks of my time spent on parser engineering:

* https://github.com/ghc/ghc/commit/bd7898537768f936d05c0c83eef1cd9b00933347
* https://github.com/ghc/ghc/commit/a78e23b8bb614ded2ff842e3a5c2dc51db1fa790
* https://github.com/ghc/ghc/commit/2f4af71e73ac3b59f4faba5bf1b25774b1008898
* https://github.com/ghc/ghc/commit/2f4af71e73ac3b59f4faba5bf1b25774b1008898

And there's seemingly no end to this, there are numerous issues that remain... The current system is a continuous source of difficulty.

Ericson2314:
FWIW the approach of "hacking up the lexer" rather than "hacking up the parser" is done but Rust too. I think it's good to get the messiness "out of the way": the the lexer is simpler and whitepsace rules easier to understand than parser tricks, so I'd hope the end user experience is better, not just developer experience.

simonpj:
We should decide about this.  I'm in favour.  @simonmar do you object?

nomeata:
@simonmar, ping? We don't have consensus yet, it seems.

isovector:
I'm in favor of this!

simonmar:
In general I'm not in favour of whitespace-sensitivity. Yes I know that there is already whitespace-sensitivity in the Haskell grammar (e.g. `1e9`), and that various GHC extensions have added more whitespace-sensitivity (`TypeApplications` and more). And yes we've already made a mess with `!` and `~`.  If we were starting from scratch I would probably argue for making `!` a reserved symbol.

So given where we currently are perhaps this is an improvement. I still worry that having to classify all tokens is a burden (both on the implementor and the user), and that we can't concisely specify, and therefore explain, the parsing rules. Those are my objections, if people still think this is worthwhile then OK.

goldfirere:
Yes. Accept.

goldfirere:
Final home: https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0229-whitespace-bang-patterns.rst

nomeata:
I thought we’d give the other committee members some time to react to our lunch decisions before acting? Also, we should have done that together with putting in the renaming/redirects 

Ah, I see you are just following what @gridaphobe did in #160.

nomeata:
Also, you forgot to update the top of this PR, like the other accepted PRs are (yes, there are many many manual steps involved here.)

tseenshe:
Sorry for the noise, but I just wanted to say that as an editing tooling contributor (see https://gitlab.com/tseenshe/haskell-tng.el) I very much appreciate this change.

PR #230, opened 2019-05-17
Propose atomicModifyIORef_, atomicModifyIORef'_

bollu:
I'm sorry for this possibly superfluous GHC proposal. I'm not entirely sure if this is the right venue for this change. If it is not, please point me to the correct place and I'll gladly move this. 

We propose the addition of `atomicModifyIORef_` and `atomicModifyIORef'_`, with the standard semantics those names imply. 

[Rendered link](https://github.com/bollu/ghc-proposals/blob/patch-1/proposals/0000-introduce-atomic-modify-IO-ref-underscore.rst)

### DIFF
--- a/proposals/0000-introduce-atomic-modify-IO-ref-underscore.rst
+++ b/proposals/0000-introduce-atomic-modify-IO-ref-underscore.rst
@@ -10,7 +10,7 @@
 .. sectnum::
 .. contents::
 
-We introduce two new functions in `Data.IORef`, `atomicModifyIORef_, atomicModifyIORef'_ :: IORef a -> (a -> a) -> IO`.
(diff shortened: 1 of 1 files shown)


cartazio:
@bollu the libraries list and a small patch to base on gitlab suffice for userland core library changes. that said, this is wayyyy better detail than many proposals for the libraries list :)

RyanGlScott:
I agree with @cartazio. The scope of this idea is narrow enough that I think it falls under the bar for a full GHC proposal and could instead suffice as a discussion on the [Haskell libraries mailing list](https://mail.haskell.org/mailman/listinfo/libraries), especially since this doesn't require any changes to the compiler itself.

nomeata:
We agreed that this does not need a full proposal, closing. Thanks nevertheless!

PR #231, opened 2019-05-17
Record `with` syntax

rohanjr:
We propose a layout syntax for records, triggered by a new keyword `with`. This means that record fields in type declarations, expressions and patterns can be formatted using indentation instead of punctuation. The feature removes syntactic noise and makes large records easier to read and modify. It is consistent with the use of layout for other constructs that span several lines (e.g. `do` blocks, branches of `case` expressions).

[Rendered](https://github.com/rohanjr/ghc-proposals/blob/record-with-syntax/proposals/0000-record-with-syntax.rst)

### DIFF
--- a/proposals/0000-record-with-syntax.rst
+++ b/proposals/0000-record-with-syntax.rst
@@ -1,0 +1,327 @@
+Record ``with`` syntax
+=======================
+
+.. proposal-number:: Leave blank. This will be filled in when the proposal is
+                     accepted.
+.. ticket-url:: Leave blank. This will eventually be filled with the
+                ticket URL which will track the progress of the
+                implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+We propose a layout syntax for records, triggered by a new keyword ``with``.
+This means that record fields in type declarations, expressions and patterns can be formatted using indentation instead of punctuation.
+The feature removes syntactic noise and makes large records easier to read and modify.
+It is consistent with the use of layout for other constructs that span several lines (e.g. ``do`` blocks, branches of ``case`` expressions).
+
+
+Motivation
+----------
+Record syntax in Haskell uses curly braces and commas. For example, you can declare a record type as follows:
+::
+
+ data Rec = Rec { field1 :: Int, field2 :: Bool }
+
+You can construct and pattern match on expressions of this type using the syntax
+::
+
+ Rec { field1 = f1, field2 = f2 }
+
+
+When records have a large number of fields, programmers typically spread declarations and expressions over several lines, e.g.
+::
+
+ data BigRec = BigRec
+   { field1 :: Int
+   , field2 :: Bool
+   , field3 :: String
+   }
+
+Similarly when constructing a big record:
+::
+
+ BigRec
+   { field1 = 1
+   , field2 = True
+   , field3 = "hello"
+   }
+
+Indentation of the fields inside curly braces is not required, but is usually maintained by the programmer (with varying styles of formatting).
+
+Let's compare this situation to ``do`` blocks in monadic code.
+You could write
+::
+
+  echo = do {
+    line <- getLine;
+    putStrLn line;
+  }
+
+But the much more standard style is to use indentation and write
+::
+
+  echo = do
+    line <- getLine
+    putStrLn line
+
+The latter version avoids meaningless characters, and alignment of statements in the ``do`` block is required by the parser.
+
+The goal of this proposal is to use the same technique and introduce indentation-based record syntax to Haskell.
+
+We propose a keyword ``with`` that induces layout syntax for record fields.
+Using ``with``, our big record declaration from earlier can be written as
+::
+
+  data BigRec = BigRec with
+    field1 :: Int
+    field2 :: Bool
+    field3 :: String
+
+We can use indentation for record construction
+::
+
+  Rec with
+    field1 = 1
+    field2 = True
+
+as well as record field updates
+::
+
+  r with
+    field1 = field1 r + 1
+    field2 = True
+
+Note that when using indentation, the lines corresponding to fields are formatted identically.
+This means that reordering or deleting fields does not require adjusting braces or commas.
+Not only is this less hassle to write, but it also leads to simpler patches and code review.
+
+On the whole, record ``with`` syntax removes syntactic noise and makes large records easier to read and modify.
+The feature is well precedented by layout syntax in other Haskell constructs.
+
+
+Proposed Change Specification
+-----------------------------
+We propose a language extension called ``RecordWith`` offering layout syntax for records.
+With the extension enabled, there is an additional keyword ``with`` which induces layout syntax for record fields.
+
+The structure and formatting options are very similar to monadic ``do`` blocks (and other layout-inducing constructs).
+Specifically, ``with`` can be followed either by indentation and newline delimiters or by curly braces ``{ }`` and semicolon ``;`` delimiters.
+For example, the following 3 declarations are all valid, equivalent ways to define a record type ``Rec``:
+::
+
+  data Rec = Rec with
+    field1 :: T1
+    field2 :: T2
+    field3 :: T3
+
+  data Rec = Rec with
+    { field1 :: T1
+    ; field2 :: T2
+    ; field3 :: T3 }
+  
+  data Rec = Rec with field1 :: T1; field2 :: T2; field3 :: T3
+
+These are all equivalent to the traditional syntax
+::
+
+  data Rec = Rec { field1 :: T1, field2 :: T2, field3 :: T3 }
+
+(which can also be spread across multiple lines).
+
+The ``with`` syntax can occur in any language construct where traditional record syntax usually appears.
+Namely, we need to handle the following features:
+
+1. Data type declarations
+2. Record construction
+3. Record field updates
+4. Record patterns
+
+We showed an example of the first case above.
+Note that ``with`` should work in data types with several constructors, as well as constructors with zero arguments.
+So the following declaration is valid:
+::
+
+  data ComplexRec = Con1 with f1 :: Int; f2 :: Bool | Con2 with | Con3 String
+
+(``Con2 with`` is permitted for consistency with ``Con2 {}``.)
+
+We can also use ``with`` to construct record expressions.
+We can use newlines and indentation
+::
+
+  Rec with
+    field1 = expr1
+    field2 = expr2
+    field3 = expr3
+
+or we can use delimiting punctuation
+::
+
+  Rec with
+    { field1 = expr1
+    ; field2 = expr2
+    ; field3 = expr3 }
+
+Record update expressions look similar:
+::
+
+  r with
+    field2 = newExpr2
+    field3 = newExpr3
+
+(where ``r`` is a record of type ``Rec``.)
+
+The last construct in which ``with`` can appear is a record pattern.
+For example, we can pattern match on a record ``r`` of type ``Rec`` as follows:
+::
+
+  case r of
+    Rec with
+      field1 = f1
+      field2 = f2
+      field3 = f3
+        -> someExpression
+
+Or more concisely using the ``NamedFieldPuns`` extension:
+::
+
+  case r of
+    Rec with field1; field2; field3 -> someExpression
+
+As this last example suggests, ``with`` should work seamlessly with syntax introduced by other language extensions.
+The major ones to consider are:
+
+1. ``NamedFieldPuns``
+2. ``RecordWildCards``
+3. ``GADTs``
+
+Working with ``RecordWildCards`` means we can construct and pattern match on records using the syntax ``Rec with ..``.
+We can also write variations on this such as constructing a record with some fields explicit and some taken from scope:
+::
+
+  let field2 = expr2
+      field3 = expr3
+  in
+  Rec with
+    field1 = expr1
+    ..
+
+Record ``with`` syntax can also work with GADTs in a fairly predictable way.
+The following example shows a GADT declaration using ``with`` syntax.
+::
+
+  data G a where
+    Con1 :: with
+        i :: Int
+        j :: Int
+      -> G Int
+    Con2 :: with { b :: Bool } -> G Bool
+
+Note that to write ``Con2`` on a single line, the curly braces are required to separate the field type from the result type.
+
+
+Effect and Interactions
+-----------------------
+The proposed change directly addresses a lack of language support for formatting records over several lines.
+
+The change has very little interaction with other language features.
+It introduces a new keyword but otherwise reuses established layout syntax.
+Moreover, it does not clash with but rather complements existing extensions like ``NamedFieldPuns`` and ``RecordWildCards``.
+
+Note that we have implemented record ``with`` syntax in a fork of GHC used for the `DAML <https://github.com/digital-asset/daml>`_ language.
+We have found the user experience to be intuitive and did not encounter unexpected interactions with other features.
+
+
+Costs and Drawbacks
+-------------------
+The main maintenance cost I can forsee is due to additional productions in the parser and extra information (a Boolean flag) in the AST nodes for records.
+For example, future changes to record parsing or pretty printing would bear a slightly higher complexity.
+However we do not consider this a significant cost.
+
+On the user side, the proposed change could in fact be *more* intuitive for novice users.
+The ``with`` syntax is consistent with other occurrences of layout syntax and promotes the use of indentation across the board.
+It avoids questions around the placement of braces and commas as well as the need for alignment conventions.
+
+The possible drawback is that there would be two different syntax styles to be aware of.
+The mixture of semicolons and commas may be confusing, as
+``Rec with { x :: Int; y :: Bool }`` looks quite similar to ``Rec { x :: Int, y :: Bool }``,
+but the field delimiters are different.
+This probably won't be much of an issue as ``with`` syntax will mostly be used with indentation instead of punctuation.
+
+
+Alternatives
+------------
+There is no alternative syntax for records that I am aware of.
+However, one could avoid large record expressions and patterns by using existing extensions such as ``NamedFieldPuns`` or ``RecordWildCards``.
+For example, you can bind field names using a ``let`` or ``where`` block outside of a record expression:
+::
+
+  let field1 = expr1
+      field2 = expr2
+      field3 = expr3
+  in
+  Rec {..}
+
+Another example is using field punning for an update expression (assuming a variable ``r`` of type ``Rec``):
+::
+
+  let field1 = newExpr1
+      field2 = newExpr2
+  in
+  r { field1, field2 }
+
+This is consistent with the Haskell idiom of using identation for blocks of name bindings.
+
+These techniques are fairly limited, however.
+For one, they do not apply to record type declarations.
+Second, they rely on bound variables being exactly the same as the field names of the record in question.
+Therefore the syntactic convenience cannot be used for multiple record expressions of the same type in the same scope.
+(The same is true if you need to pattern match on multiple records of the same type in a single definition, as the field names would clash.)
+Moreover, using ``RecordWildCards`` to construct records can be error prone as it is easy to use a variable from scope when you should have set the value of the field explicitly.
+
+
+Unresolved Questions
+--------------------
+
+With the ``RecordWith`` extension enabled, how flexible should we be in mixing traditional and layout syntax?
+For example, should we allow commas when using ``with`` (as in the following)?
+::
+
+  data Rec = Rec with field1 :: T1, field2 :: T2, field3 :: T3
+
+If so, should we also allow the reverse scenario?
+::
+
+  data Rec = Rec { field1 :: T1; field2 :: T2; field3 :: T3 }
+
+Allowing these cases may be more forgiving for users switching between the two syntaxes on different projects.
+
+We do propose that declaring multiple fields of the same type should only use commas regardless of the syntax choice, e.g.
+::
+
+  data Rec = Rec with field1, field2 :: T12; field3 :: T3
+
+Should we allow mixing commas and semicolons more generally (as in the following)?
+::
+
+  data Rec = Rec with field1 :: T1, field2 :: T2; field3 :: T3
+
+
+Implementation Plan
+-------------------
+I volunteer to implement this feature with my collaborators.
+As mentioned, we have a prototype implementation for the `DAML <https://github.com/digital-asset/daml>`_ language that we can use as reference.
+We do not require any changes to GHC in advance of the feature.
+
+The implementation entails the following changes to the compiler frontend:
+
+1. Extra ``with`` token in the lexer (conditional on the ``RecordWith`` extension being set), enabling layout syntax
+2. Extra productions in the parser for ``with`` followed by record fields within virtual braces, storing source locations as usual
+3. Boolean flags in AST nodes involving records, indicating if ``with`` syntax was used
+4. Embellishment of pretty printers to respect layout syntax (e.g. when reporting error messages)


AntC2:
Thank you @rohanjr . That's a clearly written and precisely scoped proposal.

TBH, I find the offside rule more ornament than use; and you can find Landin admitting as much in his remarks on ISWIM. Of the many things other languages have copied from FP, the layout rule has never found much appeal.

I do know there's a great deal of complexity in the lexer and parser to cope with layout, I'd be reluctant to add to that.

Layout is great if you're quickly knocking together a few lines of code. If I'm writing the meatier stuff you're aiming at, I always put explicit bracketing and separators.

>  reordering or deleting fields does not require adjusting braces or commas. Not only is this less hassle to write, but it also leads to simpler patches and code review.

I'm not seeing hassle. To take your example

>        data BigRec = BigRec
>          { field1 :: Int
>          , field2 :: Bool
>          , field3 :: String
>          }

You can freely shuffle the lines beginning `,`. It's sensible, as you've shown, to put the closing `}` on a separate line, so you can freely insert extra stuff. Yes the first line starts with `{` so that might be awkward to move around, but I usually make it that the first field in a record is the most important; that's not likely to change. I must admit that starting a line with a `,` or `;` looked weird to me when I first came across Haskell. I've long since gotten used to it.

The big advantage of explicit bracketing in refactoring is that you can cut'n'paste the whole code block and it's still syntactically valid, even if its level of indentation is 'wrong' for wherever it's landed.

Specifically in terms, for record construction or update or pattern matching, it's likely that code is nested (inside a `do` block and/or `case` expression). Then you must make its indentation level deeper. It really is hassle to rejig the indentation if I want to insert an extra layer of nesting, or copy code from somewhere else. At least with explicit bracketing, the code is still syntactically valid, even if its indentation is ragged. I can pretty it up later when I've got it working.

Ericson2314:
@AntC2 given existing diff tools' line-orientation (vs proper AST diffing), layout syntax indeed helps cope with merge conflicts. As you concede, the fact that that the first line is different, be it a `[` or `{`, means changing it is quite a pain.

> but I usually make it that the first field in a record is the most important; that's not likely to change.

I don't think that's true. I find in quickly iterating "business code"---relatively arbitrary and mutable requirements and goals---the first line is as likely to change as any other.

----

On the other hand, the other proposal for leading and trailing comments also helps with this.

vrom911:
The proposal doesn't seem to cover how it would interact with the functions named `with` (as the new suggested keyword). Could you elaborate on that?

phadej:
How this interacts with sums of records

```haskell
data SumOfR
  = A { foo :: Int }
  | B { foo :: Int, foo' :: Char }
```

or GADTs + Records:

```haskell
data Ex where
  MkEx :: { bar :: a, baz :: a -> Int } -> Ex
```

---

As far as I can see, the proposal makes only small (though common) usage case (subjectively) nicer, yet the old syntax would have to be there still.

Thus :-1:

shayne-fletcher-da:
> The proposal doesn't seem to cover how it would interact with the functions named `with` (as the new suggested keyword). Could you elaborate on that?

I expect it is implemented as a "pseudo keyword" -- that is, IIUC, only lexed as a keyword only when the `RecordWith` extension is in effect.

goldfirere:
I think this proposal makes sense (as in, I believe the resulting language would be unambiguous, etc.) but I, personally, don't think it's worth it. There are some noteworthy drawbacks of this proposal (numbered only for easy backreference):

- The `;`/`,` confusion described in the proposal.
- Another extension GHC aficionados need to know.
- Potential confusion if a programmer in one style sees the other.
- Necessary rules that will arise around functions named `with` (piggybacking on @vrom911's https://github.com/ghc-proposals/ghc-proposals/pull/231#issuecomment-493645267)
- More implementation complexity.

None of these drawbacks is so terrible on its own, but I don't see that the gain is worth the pain here.

shayne-fletcher-da:
> I do know there's a great deal of complexity in the lexer and parser to cope with layout, I'd be reluctant to add to that.

It really doesn't add complexity there. The layout stuff in the lexer is designed to be trivially extensible in this way. The parser does get more productions no way around that - they follow well established patterns though and are just variants of the existing `field-decl` productions - fairly minimal. As the proposal points out, there will also have be a field(s) in the AST to keep track of usage and updated pretty printing to take into account whether the syntax is in effect or not. That small work shouldn't prove deep either and has plenty of precedent.

shayne-fletcher-da:
> None of these drawbacks is so terrible on its own, but I don't see that the gain is worth the pain here.

We do. Small improvements add up!

shayne-fletcher-da:
> Another extension GHC aficionados need to know.

This is unavoidable. It is a rule that any accepted syntax changes (past the Haskell2010 specification) be under a language extension.

AntC2:
>> None of these drawbacks is so terrible on its own, but I don't see that the gain is worth the pain here.

> @shayne-fletcher-da Small improvements add up!

Yes, add up to confusion. Many different ways to write the same code (improvements or not, small or not) add up to an increasing mental load on reading programs. (Most code gets read and modified far more than it's written.) It's not that any one proposal is much burden considered alone; it's that all of them together (which any one module might or might not be using) gives a combinatorial explosion of possible idioms the reader has to bear in mind.

I just hate having to figure out code that uses deep nesting layout. If all the indenting doesn't fit on a single screen, it's just impossible to figure out what brackets with which. Gone are the days when you would print the code on lineflow and draw vertical margins and indents. So I won't use this extension in my own code, but still I would need to know about it when reading others' code.

shayne-fletcher-da:
> * The `;`/`,` confusion described in the proposal.

I personally advocate that the two syntaxes be allowed to freely interoperate. I respect that there will likely be differences of opinion on that.

shayne-fletcher-da:
> So I won't use this extension in my own code, but still I would need to know about it when reading others' code.

Perfectly fair! I do think though that the syntax is natural enough and consistent with the other multi-line constructs to the extent that it justifies this apparent burden and shouldn't in practice evoke surprise.

shayne-fletcher-da:
> I just hate having to figure out code that uses deep nesting layout. If all the indenting doesn't fit on a single screen, it's just impossible to figure out what brackets with which.

Yep. Couldn't agree more -- in all things moderation of course.

gbaz:
A key consideration why I think this is not worth it is that the gain over `RecordWildCards` seems very small to me. One way to recover most of the gain without a _language_ extension would be to add an optional warning for when binding wildcards captures things that come from somewhere outside of the immediate scope. Combined with warning on incomplete records this gives most of the benefit of this extension without altering the core language at all.

ndmitchell:
As someone who has used this language feature in anger (via experience in DAML), I find it's ergonomic and has a nice "feel" to it. I think it makes Haskell records just a bit more elegant, which has a big pay off for record-heavy code.

> One way to recover most of the gain without a language extension would be to add an optional warning for when binding wildcards captures things that come from somewhere outside of the immediate scope

I am unconvinced. To recover all the gain you'd need to define `let x = ...; y = ... in R{..}` to be "local scope", and effectively use a very constrained syntax and slightly odd warning to invent syntax. You also would still struggle to define say a record containing nested records of the same type with different field values.

cartazio:
I kinda think a compound word like “record where” as a sort of flavored
where clause, would be more palatable than eating with for use in an
extwnsion that afaik doesn’t serious impact expressivity.

Maybe I need to re read some expositions. But I get the sense this is
mostly a bunch of sugar that consumes an otherwise nice lexeme

nomeata:
I am generally fond of layout over punctuation, so I am sympathetic.

mrBliss:
Will type signatures for fields be supported when constructing/modifying a record? For example:

```haskell
data Rec = Rec with
    field1 :: Int
    field2 :: Int -> Int

Rec with
    field1 :: Int
    field1 = 1

    field2 :: Int -> Int  -- †
    field2 x = x
```
† Would you also accept a signature like `a -> a`, which is more general than the expected field type `Int -> Int` (similar to `-XInstanceSigs`)?

rohanjr:
> Will type signatures for fields be supported when constructing/modifying a record?

@mrBliss I wasn't anticipating usage like that. Is it possible to do that using traditional `{ }` and `,` syntax? I was proposing feature parity with traditional record syntax.

rohanjr:
> How this interacts with sums of records
> 
> ```haskell
> data SumOfR
>   = A { foo :: Int }
>   | B { foo :: Int, foo' :: Char }
> ```

@phadej I proposed `with` to work with sums of records too. Your example would look like
```haskell
data SumOfR
  = A with foo :: Int
  | B with
      foo :: Int
      foo' :: Char
```

> or GADTs + Records:
> 
> ```haskell
> data Ex where
>   MkEx :: { bar :: a, baz :: a -> Int } -> Ex
> ```

As mentioned at the end of the Proposed Change section, I envision `with` working with GADTs as well. You could write your example as
```haskell
data Ex where
  MkEx :: with
      bar :: a
      baz :: a -> Int
    -> Ex
```
Note that you need to write `-> Ex` on a new line to separate it from the type of `baz` (otherwise you still need the `{ }`).

> As far as I can see, the proposal makes only small (though common) usage case (subjectively) nicer, yet the old syntax would have to be there still.

I agree that the old syntax will have to remain, but hope I've addressed your points about the proposal only improving one small use case. :)

AntC2:
> @mrBliss Will type signatures for fields be supported when constructing/modifying a record?

As I understand it, this proposal is only introducing syntactic sugar for current functionality, not adding new functionality. Are you asking for something that can't be done now? Do you have a use case?

Or are you saying the proposed syntax (without the `{  ,  }` decoration) could be confused with indented code in a `do` block or `case` body?

> For example:

>        ...
>        Rec with
>            field1 :: Int
>            field1 = 1
>
>            field2 :: Int -> Int  -- †
>            field2 x = x

Today you can go

    myrec = Rec 
              { field1 = 1 :: Int
              , field2 = (\x -> x) :: Int -> Int
              }

> † Would you also accept a signature like a -> a

You can put `field2 = (\x -> x) :: a -> a`. But that won't change the actual signature of `field2` returned in `myrec`. You'd need existential types or GADTs for that.

The signatures appearing in record field declarations are somewhat bogus.

    data Rec = Rec { field1 :: Int, ... }

Doesn't mean you get anything named `field1` with type `Int`. You get a field selector function `field1 :: Rec -> Int`. And you get validation that in record construction/update `Rec { field1 = e, ...}`, `e`'s type unifies with `Int`.

Construction/update syntax wants bare field name followed by `=`. So this is not syntactically valid

    ... Rec { field2 x = x }

because there's nothing named `field2` of type `Int -> Int`. There is a field selector `field2 :: Rec -> (Int -> Int)`.

rohanjr-da:
> I think this proposal makes sense (as in, I believe the resulting language would be unambiguous, etc.) but I, personally, don't think it's worth it. There are some noteworthy drawbacks of this proposal (numbered only for easy backreference):

@goldfirere Thanks for the nice summary response. I'll respond to your points individually.

> * The `;`/`,` confusion described in the proposal.

We can get around the `;`/`,` confusion by allowing field lists to use either delimiter, i.e. allow both `Rec with f1 = 1, f2 = 2` and `Rec { f1 = 1; f2 = 2 }` (allowing mixed `;` and `,` in the same field list is debatable - my personal preference is not to)

> * Another extension GHC aficionados need to know.

I argue that the barrier to learning this extension is very low. It is quite likely one could guess the meaning of `with` constructs without reading any documentation, and if not a quick skim of the docs should give an understanding in a few minutes. I think this a cost we can accept for users to adapt to evolution in the language.

> * Potential confusion if a programmer in one style sees the other.

I think the potential confusion is addressed in the last two points. Learning the new style takes minutes, and we can be somewhat flexible in mixing the two syntaxes. This should mean that the switching cost is very low.

> * Necessary rules that will arise around functions named `with` (piggybacking on @vrom911's [#231 (comment)](https://github.com/ghc-proposals/ghc-proposals/pull/231#issuecomment-493645267))

It is a valid point that using the extension consumes a potentially useful lexeme. I can’t immediately think of when I would use `with` as an identifier, but we shouldn’t consume it without a good reason. I think this feature is worth it. We could instead attempt to reuse an existing layout keyword such as `where` or `of`. However I believe this would lead to either ambiguities during parsing or more confusing code.

> * More implementation complexity.

As stated in the Costs and Drawbacks section, I don’t expect the implementation to have undue complexity. I think some added complexity is inevitable in evolving the language.

> None of these drawbacks is so terrible on its own, but I don't see that the gain is worth the pain here.

Your assessment is fair, but certainly subjective. As we've seen in the comments so far, there are differing opinions on the utility of layout syntax in general. So I suspect this proposal will divide those who like layout (and want to extend it in a logical way) and those who question it in the first place.

goldfirere:
@rohanjr-da:

> Your assessment is fair, but certainly subjective.

Indeed! Thanks for responding to my points. Your arguments are all roughly of the form "Yes, that's a drawback, but this feature is worth that cost." My argument is essentially "Yes, that would indeed be nice -- and perhaps would have been nice from the beginning -- but the cost is too high." These are subjective arguments, both.

My subjective assessment has not yet been swayed by your arguments. But I'm only one person! So don't be discouraged just because of my one opinion -- you don't really need to convince me. However, I am somewhat strengthened in my position by the reactions to my post. The way to win battles of taste is to have a big army. If there are lots of people clamoring for this new syntax, that would help to convince me to stand down.

rohanjr-da:
@goldfirere 
> These are subjective arguments, both.

Yes, I agree! I wasn't trying to imply that my arguments are not also subjective :)

> So don't be discouraged just because of my one opinion -- you don't really need to convince me. However, I am somewhat strengthened in my position by the reactions to my post.

Yes, in responding to your points I was also trying to sway those who agreed with you. I'm working on the army :P

shayne-fletcher-da:
> sugar that consumes an otherwise nice lexeme

OCaml uses `with` for record (update) expressions. Prior art  - Haskell record syntax, while different, avoids unnecessary impedance mismatch by using it here.

fommil:
Haskel2010 has whitespace provisions for `where`, `let`, `do`, `of`... or WLDO. It seems like a missed opportunity not to use the keyword `admit` so we can finally spell WALDO correctly in our acronym! :laughing: 

More seriously though, I voted :-1: because this pushes a burden onto lexing/parsing tools (including all text editors, and especially indentation rules) and I don't think the benefit of "slightly cleaner definitions of a subset of data types" (as @phadej points out) are worth it for that level of breakage.

gridaphobe:
Given that one of the main concerns with this proposal (which I share) is stealing `with`, I think it would be helpful for the proposal to spell out why existing keywords cannot be reused for this extension. As an example, `where` seems like a very natural replacement for `with`, but it would make (at the very least) record updates ambiguous, e.g.

```haskell
r where
  field2 = newExpr2
  field3 = newExpr3
```

As to where one might use `with` as an identifier, I have recently found it very nice to export resource-allocation functions called `with` that are meant to be imported qualified. Then you end up writing code like

```haskell
Consumer.with $ \consumer ->
Producer.with $ \producer -> do
  ...
```

AndreasPK:
When I think of Haskell2030 I would rather have fewer than more record syntax variants.

I think the big sell here is not the ability to do define records using layout but this:

> Note that when using indentation, the lines corresponding to fields are formatted identically. This means that reordering or deleting fields does not require adjusting braces or commas. Not only is this less hassle to write, but it also leads to simpler patches and code review.

Something which ideally would be possible with the old form by allowing trailing commas. Something which has been discussed [in the past](https://github.com/ghc-proposals/ghc-proposals/pull/87) and which I think is reasonably to expect to be accepted in some for eventually.

So we could eventually use syntax like below and get the same effect.
```
data Foo = {
  f1 :: T1,
  f2 :: T2,
  }
```

Once this happens the remaining value of this proposal is in the ability to define records just via layout. Which to me is not worth the cost of having two different kinds of syntax. But that is subjective.

Even so I think this has a few weak points.

* `where` is a existing keyword which seems just as natural to me.
* The sum types example actually seems like it would be harder to parse using this approach.

```
data SumOfR
  = A with foo :: Int
  | B with
      foo :: Int
      foo' :: Char
```

* The interaction in GADT Syntax where we either need curly braces or move the result type to a new line will likely also lead to a lot of head scratches over time.

rohanjr-da:
@gridaphobe @AndreasPK Thanks for the helpful feedback. I agree it is worth exploring whether we could use existing layout keywords instead of `with`. The only ones I can see making sense are `where` and `of`.

`where` reads quite naturally for record constructs. For example, you could declare a record like
```
data R = R where f1 :: Int; f2 :: Int
```
However, there are serious ambiguities when it comes to record construction and update. Using the previous declaration of `R`, the following term declaration has two type-correct meanings:
```
e = R where f1 = 2; f2 = 2
```
(one being a record of type `R` and the other being the constructor of type `Int -> Int -> R` ignoring the local `f1` and `f2` bindings).

Similarly, we get the ambiguity mentioned by @gridaphobe for record updates:
```
r = R 1 2
e = r where f1 = 11; f2 = 22
```
As far as I understand, these issues are dealbreakers for reusing `where` for record fields.

`of` may be more plausible. A declaration looks like
```
data R = R of f1 :: Int; f2 :: Int
```
However there may be confusion (both for users and for the GHC parser) when trying to parse a record expression within a `case` expression.
```
c = case R of f1 = 1; f2 = 2 of R of f1; f2 -> f1
```
There is only one successful parse of this example, but I'm not sure if it will cause an error in parser generation or parsing itself.

You can see further issues for a record update within a `case` expression.
```
r = R of f1 = 1; f2 = 2
c = let f1 = 11 in case r of f1 of f1 -> f1
```
The first `of f1` in `c` is a field update using `NamedFieldPuns` whereas the second `of f1` starts a pattern binding a variable `f1`. I think there is only one successful parse of this example too, but again I expect this to cause trouble for the parser.

Can anyone confirm whether these examples show a technical issue with parsing `of`? If there is no technical issue, then there is only the readability concern, but I would say these examples are very much contrived, i.e. you usually wouldn't construct or update a record directly in a case scrutinee.

Sorry for the long comment! I can add this to the proposal once I have answers to the technical questions.

cartazio:
so heres my thoughts atm 👍 

defining a record type

```
data record Foo ... where 
   f1 :: ...
...
```

defining creating a record foo values
```
let record myVal where 
   f1  = ...
...
```

on the other hand, this whole proposal seems pretty hefty for whats (essentially) the trailing/leading commas for record syntax by another name.

cartazio:
so my current stance is "nope, this doesn't really make anything genuinely simpler or easier"

in fact, it seems to be kinda subtly wrong:

 1) when we move into the dependently typed world (as ghc is moving) with general recursion/coinduction, you find yourself with TWO different but both valuable  notions of records
 2) https://agda.readthedocs.io/en/v2.6.0.1/language/record-types.html -- usual dependent record
and 
 3) https://agda.readthedocs.io/en/v2.6.0.1/language/coinduction.html -- COINDUCTIVE records

in this enriched setting, the only records which have "permutation" invariance in data type definition field order would be non-dependent/simply typed records AND coinductive records.

Its worth pointing out, in DAML (where the `with` proposal originates), HAS coinductive records with effects, they just use a "template" keyword. and call them objects ;) 

for records, its useful to talk about `construction` (defining a record value) and `destruction` (projection of a field!)

for constructing records and composing building the, COPatterns are a fantastic tool! 
https://agda.readthedocs.io/en/v2.6.0.1/language/copatterns.html -- has a LOT of nice examples. and that syntax works VERY nicely.  

4) how to desugar copatterns where you have ordered clause semantics and fancy types is documented  in https://jesper.sikanda.be/files/elaborating-dependent-copattern-matching.pdf and the agda codebase

5) i've found it SUPER helpful for writing genuinely complicated statemachines to even simulate/fake copattern matching https://www.reddit.com/r/haskell/comments/4aju8f/simple_example_of_emulating_copattern_matching_in/ is an exposition from a few years ago i shared that has some toy examples. 


# summary
 1) this proposal doesn't result in improved expressivity now or when ghc gets dependent types, 
 2) borrowing agda syntax for fancy record definitions seems more palatable to me because we can probably safely avoid eating up names/keywords, or just have `data record Sigma ... where` and perhaps `data coinductive record MyStateMachine where .. `
 3) scope order of fields does matter for Dependent records, which also let you model modules/various other modularity notions quite nice
 4) copattern matching is fantastic and its been on my "would be nice to add to ghc" list for a long time

Syntax that doesn't come with added expressivity is a heavy weight to carry forever, and thats what i worry may happen here.

cartazio:
nb: i'm genuinely wanting what i sketch out above, and i'd be super happy to help @rohanjr and @ndmitchell understand and execute such an alternative if theres the appetite for that

cartazio:
heres some examples
# example  1

 a concrete example: consider infinite streams! Lets mark them coinductive so they have CBN / non memoizing and  nonstrict semantics!
```
data coinductive record Stream (a :: Type) where 
    head :: a 
    tail :: Stream a 

-- lets define zipWith using copattern syntax!
zipWith :: (a -> b -> c) -> Stream a -> Stream b -> Stream c
.head (zipWith f stA stB) = f (head stA) (head stB)
.tail (zipWith f stA stB) = zipWith f (tail stA) (tail stB)
```
I think thats pretty nice!  and doesn't really require any type system changes to do today (a simulation of that in current haskell is the reddit link from my previous comment https://www.reddit.com/r/haskell/comments/4aju8f/simple_example_of_emulating_copattern_matching_in/)


## dependent records
this is really about supporting *dependent* existentials

### in agda 
```agda
record Σ {a b} (A : Set a) (B : A → Set b) : Set (a ⊔ b) where
  constructor _,_
  field
    fst : A
    snd : B fst
```
notice how the second field *uses* the value of the *first* field in its *type*!
also notice how theres machinery to provide syntax for the record former! Here its `v1 , v2` aka the comma `_,_` is an infix operator. (its a nice factoring of type name, value former, and field projection names!)

one could imagine writing stuff like 

```
data record SizedList (a :: Type) (s :: Nat) where
    list : [a]
    lengthPf :  (s :~: lengthNat list)
```
where the second field will just (ultimately) be a `Refl` proof that the value in the field `list` has the finite size `s`

cartazio:
amusingly: github's haskell syntax highlighting works pretty well on my two haskell examples in the preceding comment

cartazio:
nonfancy version of streams types

```haskell
data coinductive record Stream a where 
    head :: a 
    tail :: Stream a 

-- lets define zipWith using copattern syntax!
zipWith :: (a -> b -> c) -> Stream a -> Stream b -> Stream c
.head (zipWith f stA stB) = f (head stA) (head stB)
.tail (zipWith f stA stB) = zipWith f (tail stA) (tail stB)
```

rohanjr-da:
@cartazio This looks interesting, but well beyond the scope of this proposal. Looks like it could be a compelling proposal of its own though!

rohanjr:
I've included the concern about consuming the `with` keyword to the Drawbacks section, and added a subsection in Alternatives with my analysis of reusing an existing layout keyword.

Note that I've also removed the unresolved questions around mixing semicolons and commas between syntaxes, e.g. allowing `Rec with f1 = 1, f2 = 2` (with a comma). This is due to ambiguity with tuple syntax, e.g. `(rec with f1, f2)` could be either a pair of type `(Rec, Int)` or a single record update of type `Rec`.

My perception is that the main concerns with the proposal are the consumption of a keyword, as well as the cognitive overhead of having multiple record syntaxes (when learning the language, reading code or switching between codebases). However it seems that many people do see the benefits in terms of readability and ease of editing.

@nomeata, I think this proposal is ready for committee review!

simonpj:
I would love

* A clear statement (in the proposal) that `with` is a new "layout herald".  (I'm not sure where this term is defined, but it does a specific meaning; e.g. `let` and `where` are both layout heralds, as is `\case`.)

* A clear statement that it behaves exactly like other layout heralds wrt indentation and semicolons.  So, for example
  ```
  Rec with
    field1 = expr1
    ..
  ```
  means precisely `Rec with { field1 = expr1 ; .. }`.   (Note the semicolon before the `..`.

* A BNF for the language after the layout mechanism has inserted the curly braces and semicolons.

I wonder whether we should allow semicolons in the non-with situation, thus `Rec { x = 3; y = 4 }`, as well as the current commas.  I don't see any reason not to and it would make it all more uniform.

I'm not a big supporter, but nor am I a detractor if a signficant constituency finds it useful.

shayne-fletcher-da:
> "layout herald"

Very nice. Definitely replacing the awkward "layout inducing keyword" phrase I'd invented for this notion in my vocabulary 😃

cartazio:
Simon raises an interesting point :

; as record field separator doesn’t have the ambiguity of meaning , has!
So we could just do lead or trailing ; for the same easy editing that this
proposal makes as it’s goal

rohanjr:
Thank you @simonpj for the useful feedback. I have pushed a commit to address your requests, and in particular have taken your suggestion to allow semicolon field delimiters in the non-`with` case. I agree it makes the syntax more uniform overall.

@simonmar @nomeata Please consider the updated proposal in the committee process.

nomeata:
We are still waiting for Simon Marlow for a recommendation anyways, so I guess he can now just look at the updated proposal.

simonmar:
@ghc-proposals/ghc-steering-committee

I propose that we accept this proposal. Rationale:
* it fits well with the existing syntax and doesn't have any unpleasant interactions, as far as I can see.
* it's an addition of syntactic sugar only and therefore low risk

On a personal level, having written a lot of code that uses large records I imagine I would use this if it were available.

nomeata:
Part of me wants to be conservative about adding purely syntactic features. Other parts really prefer layout over separators (and the discussion at https://github.com/ghc-proposals/ghc-proposals/pull/87 is convincing me that separators are a bad idea in general). So I am slightly on the fence, but not firmly, so Simon pushed me over, so I concur with him.

goldfirere:
I remain against, for the reasons I articulated [above](https://github.com/ghc-proposals/ghc-proposals/pull/231#issuecomment-493715771). (To be fair, the author responded eloquently in several following comments.) My bottom line is that I just don't think this syntactic nicety is worth it.

I sometimes wonder how to change the equation around syntax. Over the years, we've identified a number of small concrete-syntax changes that have fizzled as "not worth it". The one rare one that made it over the line was `-XBlockArguments` (a.k.a. `-XArgumentDo`, which I can't seem to overwrite in my memory). Even bits of syntax that we think are unambiguously better than the status quo meet resistance. What can be done about this? My best guess is to perhaps have a meta-ticket where small syntax changes can be accumulated and then accepted en masse. Yet I don't know whether that would make them more palatable or not. Maybe syntax is just too hard to change?  It's just that rejecting a proposal such as this one on subjective grounds, with no way forward, seems limiting.

nomeata:
> My best guess is to perhaps have a meta-ticket where small syntax changes can be accumulated and then accepted en masse.

goldfirere:
> Or just that accepting them feels more like a “significant step”?

It's more this... but I'm dubious anyway. Your questions are why I waffled after the suggestion. I want there to be ways to make incremental syntactic improvements, but I just can't see a good way to do so.

proposals/0000-record-with-syntax.rst
```diff
@@ -105,49 +105,72 @@ The feature is well precedented by layout syntax in other Haskell constructs.
...
   data Rec = Rec with field1 :: T1; field2 :: T2; field3 :: T3
 
-These are all equivalent to the traditional syntax
+Here the layout mechanism inserts the curly braces for us.
+
+We can also use semicolon delimiters without the ``with`` layout herald:
 ::
 
-  data Rec = Rec { field1 :: T1, field2 :: T2, field3 :: T3 }
+  data Rec = Rec { field1 :: T1; field2 :: T2; field3 :: T3 }
```
gridaphobe: This option, using semicolon delimiters without the `with` herald seems inconsistent with the other layout-aware keywords in the language. Perhaps we should limit the use of semicolon delimiters to the new `with` syntax?

nomeata: Could go either way, that does sound reasonable to me. I assume that most people don’t think about `;` when they see layout code, so emphasizing the `;` here too much does not seem to be useful.

gridaphobe:
My main concern with this proposal has been that it steals a nice identifier for resource-allocation functions, `with`. I've just assumed that this syntax theft would even prevent me from using a qualified import of a function named `with` (e.g. https://github.com/ghc-proposals/ghc-proposals/pull/231#issuecomment-497324583). The insidious thing here is that this would mean that this proposal would burden even libraries that do not use the extension, as enabling it in a client module could make a library that exports a `with` function unusable.

But I believe I was mistaken! It turns out that GHC will happily compile the following:

```haskell
module Foo where

rec :: Int -> Int
rec = undefined
```

```haskell
{-# LANGUAGE RecursiveDo #-}
module Bar where
import qualified Foo

bar :: Int -> Int
bar = Foo.rec
```

This works for other pseudo-keywords, and even regular keywords like `if` or `let` that you could never use as an identifier.

So, as long as the same will hold for the new `with` keyword, I withdraw my objection and support this proposal.

yav:
As someone who uses Haskell's records a lot I would certainly use this if it were available.

simonpj:
> So, as long as the same will hold for the new with keyword, I withdraw my objection and support this proposal.

I dont' think it will. Today you can say
```
with = True
foro = Just with
```
but I think that under this new proposal the `with` would trigger layout etc.  (The proposal is silent on whether the binding `with = True` would be illegal.  I assume it would -- but the propsal should say.

So your question is whether a library exporting `with` would become unusable.  The propsoal doesn't say, but perhaps you could use the imported `with` in a _qualified_ way, thus:
```
import M( with )          -- Not a use of the 'with' keyord
foo = Just M.with         -- Not a use of the 'with' keyord
bar = R with { x = 4 }    -- A use of the 'with' keyord
```
I'd love all this to be explained in the proposal.

gridaphobe:
@simonpj right, my expectation based on the way other keywords currently work in GHC is that an imported `with` would be usable **qualified**, which is good enough for me.

I agree that it would be good to discuss this in the proposal itself.

simonpj:
Here's another question.  Today I can write
```
MkT { f = \x -> x+1, p = fst t }
```
and with this proposal I'll be able to write
```
  MkT with
      f = \x -> x
      p = fst t
```
But as soon as it's in that form people are going to ask for
```
  MkT with
      f x = x+1
     (p,_) = t
```
That it, the `with` clause looks like a bunch of bindings, and I anticipate pressure to make them behave that way.  Don't you?

And indeed
```
   let f x = x+1
       (p,_) = t
   in MkT {..}
```
does exactly that.  

Maybe `with` should be just a postfix form of `let`. Thus you can write `e with b` instead of `let b in e`.   Then your examples would look like `MkT {..} with <the bindings>`.

Just brainstorming.  I'm sympathetic to non-invastive syntactic extenstions because they are superficial and easy to explain. But I'm unfomfortable about this one, and I'm trying to articulate why.

gridaphobe:
> Here's another question. Today I can write
> 
> MkT { f = \x -> x+1, p = fst t }
> and with this proposal I'll be able to write
> 
>   MkT with
>       f = \x -> x
>       p = fst t
> But as soon as it's in that form people are going to ask for
> 
>   MkT with
>       f x = x+1
>      (p,_) = t
> That it, the with clause looks like a bunch of bindings, and I anticipate pressure to make them behave that way. Don't you?
> 
Yes, probably. But there have already been requests for the same pattern bindings in regular records, so I don’t really think this proposal would change the situation that much. 
> Maybe with should be just a postfix form of let. Thus you can write e with b instead of let b in e. Then your examples would look like MkT {..} with <the bindings>.
> 
This is an interesting alternative, though it seems like this version of `with` is just another name for `where`. Still, I think many people who like the current proposal would not like your alternative because it relies on RecordWildCards.

yav:
@gridaphobe the `with` that @simonpj is suggesting would be different from `where` in that `e with ds` is an expression, whereas `where` can be attached only to declarations.

simonmar:
To clarify: with the extension enabled, `with` would become a keyword. As per the lexical syntax of Haskell 2010, a qualified keyword such as `M.where` is not a valid qualified identifier (`qvarid`), therefore it would lex as the three tokens `M`, `.`, and `where`.

`with` is currently used as an identifier in [a few places](https://www.stackage.org/lts-12.19/hoogle?q=with), most notably [Foreign.Marshal.Utils.with](https://www.stackage.org/haddock/lts-12.19/base-4.11.1.0/Foreign-Marshal-Utils.html#v:with) which is part of the Haskell 2010 standard libraries.

I think @simonpj's [point](https://github.com/ghc-proposals/ghc-proposals/pull/231#issuecomment-502760687) about the potential confusion between sets of bindings and `with` syntax is a good one and I'm persuaded that this is not as straightforward a win as I thought.

Imagine allowing `<expr> where <binds>` as an expression form, and then we would have `Rec{..} where <binds>` instead of `Rec with <binds>`, but in a more general and uniform way. I'm sure there must be unforseen complications arising from allowing `where` in expressions, though.

simonpj:
>  I'm sure there must be unforseen complications arising from allowing where in expressions, though.

At the moment `where` scopes over multiple guarded RHSs
```
f x | x > foo   = e1
    | g x       = e2
    | otherwise = e3
  where
     foo = h x
     g y = y>7
```
Here the bindings for `foo` and `g` in the `where` clause scope over all the guards.  We don't want to mess with this.

However I think there would be no technical difficulty with using a new keyword, such as `with`.  The  new expresion form could be `<expr> with <bindings>`, which behaves just like `let` but with the bindings written postfix. 

Like this proposal, it woud be tehnically trivial, and arguably sometimes stylistically attractive.  Personally I'd prefer it to the current proposal, because it's useful in many more situations (beyond records), and because it allows arbitrary bindings (includign for records).   But it's still one more thing.

gridaphobe:
> To clarify: with the extension enabled, with would become a keyword. As per the lexical syntax of Haskell 2010, a qualified keyword such as M.where is not a valid qualified identifier (qvarid), therefore it would lex as the three tokens M, ., and where.

This is how I expected GHC to lex it, but GHC currently treats qualified keywords as identifiers. For example, https://repl.it/repls/BrownOrneryFolders contains the following program to experiment with:

```haskell
-- Foo.hs
module Foo where

proc :: Int -> Int
proc = (+1)

-- Main.hs
{-# LANGUAGE Arrows #-}
import qualified Foo

main :: IO ()
main = print (Foo.proc 1)
```

GHC accepts this program with no complaints, even though `Arrows` introduces a new `proc` keyword. You can also replace `Foo.proc` with e.g. `Foo.where`; it will not compile of course since `Foo` does not export `where`, but it **does** parse.

Perhaps this is a bug in GHC's lexer and parser, but it's a rather handy bug when you have to deal with extensions that introduce new keywords. It also occurs to me that this is not the first extension to introduce a new keyword that conflicts with standard library functions. `Arrows` introduces the `proc` keyword, but [`proc`](https://hackage.haskell.org/package/process-1.6.5.1/docs/System-Process.html#v:proc) is also a function in the `process` package, which is distributed with GHC.

nomeata:
(unhelpful tangent:)
Ha, if we had #214 and would allow them also in binding sites, we could even say that
```
data where :: String
data where = "Look, ma, I have something that has a keyword as a name"
```

simonmar:
> > To clarify: with the extension enabled, with would become a keyword. As per the lexical syntax of Haskell 2010, a qualified keyword such as M.where is not a valid qualified identifier (qvarid), therefore it would lex as the three tokens M, ., and where.
> 
> This is how I expected GHC to lex it, but GHC currently treats qualified keywords as identifiers. 

My memory of this is a bit hazy, but I think I did this because it was just too annoying to implement the strict Haskell 2010 lexical syntax. It's even sort-of documented: https://downloads.haskell.org/ghc/latest/docs/html/users_guide/bugs.html#lexical-syntax

and we have a test, which is marked as "should compile" even though strictly speaking it should fail:

https://gitlab.haskell.org/ghc/ghc/blob/master/testsuite/tests/parser/should_compile/read023.hs#L4

I fear we have stumbled into a dark corner here :) Really this should be an extension, with a flag...

gridaphobe:
> I fear we have stumbled into a dark corner here :) Really this should be an extension, with a flag...

If we create a new extension to cover this difference from the Haskell Report, should `RecordWith` (and `Arrows`, `RecursiveDo`, etc.) imply it? I'm inclined to say yes, to give users a way to use imported functions that conflict with the extra keywords. Or perhaps this is a case where a friendly suggestion "Did you mean to enable `QualifiedKeywords`?" that editors can recognize would suffice?

yav:
Heh, I think that according to the report `M.where` should be lexed as `M.wher e` due to the maximum munch rule.  I wouldn't say this is something we particularly need to prioritize fixing :)

I can confirm that there are no problems with expressions like `e with ds` as we have them in Cryptol, which has a fairly similar expression language.

simonmar:
> Heh, I think that according to the report `M.where` should be lexed as `M.wher e` due to the maximum munch rule. I wouldn't say this is something we particularly need to prioritize fixing :)

Ah yes, I'd forgotten that! I suspect the reason we didn't implement it in GHC is because this aspect of the lexical syntax is arguably wrong. I wish I'd written that down for my future self...

gbaz:
> Like this proposal, it woud be tehnically trivial, and arguably sometimes stylistically attractive. Personally I'd prefer it to the current proposal, because it's useful in many more situations (beyond records), and because it allows arbitrary bindings (includign for records). But it's still one more thing.

I like this. I've often wanted a `where` that works in expressions in general.

rohanjr:
Thanks everyone for the good discussion so far! I thought I'd chime in on two points that have come up:
1. How `with` can be used as an identifier
2. @simonpj's proposed [alternative](https://github.com/ghc-proposals/ghc-proposals/pull/231#issuecomment-502760687) of having `with` be a postfix form of `let`, for use with record wildcards.

On 1, it's true the proposal was light on details about if `with` can ever be used as an identifier. The assumption I made is that, in a file using the `RecordWith` extension, `with` cannot be defined as an identifier anywhere in the file. The question is what happens when a module that does not use `RecordWith` exports an identifier called `with`. Like @gridaphobe, I'd love to be able to import that identifier qualified and use it in a file with `RecordWith` turned on. From what @simonmar is saying, I understand that `M.with` should parse with the current handling of keywords, but that it violates the Haskell 2010 spec. I'm all for amending the spec to account for this use case, if there are no other implications, as it would increase usability of the extension (and other extensions that introduce keywords). However that is probably an issue to be discussed separately.

On 2, I don't agree that a combination of postfix `let` and record wildcards would solve the original use cases or is a more general solution. I see how it can help with record constructions, but how does it help with declarations, patterns or updates? For example, my proposal allows me to write the pattern match
```haskell
case r of
  Rec with
    f = [Just x]
    g = (n, m) -> ...
```
and the record update
```haskell
r with
  f = newF
  g = newG
```
Using the postfix `let` version, I don't know how I'd write the pattern example, and for the update example I would have to write
```haskell
let Rec{..} = r
in  Rec{..} with
      f = newF
      g = newG
```
which is less effective.

Further, when using record wildcards to construct records, it is very easy to accidentally use a variable that is in scope instead of specifying a field value. For example,
```haskell
let Rec{..} = r -- use wildcard pattern match
    ...
 in Rec{..} with -- construct new Rec
      f = exprF
      -- forget to set g as it's already in scope
```

I agree that allowing general binding forms for record fields is a possible approach, but as they aren't supported in the current record syntax I see it as a separate proposal.

cartazio:
I def have wanted a better where as post fix let many many times and I’ve
also seen where not being that bite / surprise folks many many times.

Record wildcard stuff could be addressed with a copatterns plus this binder
stuff of a post fix let

simonmar:
Discussion appears to have stalled. To summarise the current state:

* There seemed to be general agreement that the feature would be useful,
* @simonpj raised the point that, in expressions, the `with` form looks like a list of bindings, but in fact it would be much more restrictive. Furthermore, having a way to introduce a list of bindings with a prefix identifier within expressions would be useful, and in conjunction with `RecordWildCards` it addresses part of the motivation for this proposal, but with a construct that's more generally useful.
* @rohanjr points out that this only addresses record construction, and not pattern matching or record update.
* Stealing `with` as a keyword (albeit only when the extension is turned on and only when used unqualified) would have negative consequences and break some existing code.

At this point I don't have a clear opinion one way or the other, I'd like to hear what others think.

simonpj:
As usual when we are on the fence, I lean towards no-op.   The option of a post-fix let doesn't do everything that @rohanjr wants, but it does quite a lot _else_ that would be attractive.  Yet we can't sensibly do both.

To me, the gain seems small.  Not zero, I grant, and clearly important to some.  But as it stands I don't feel convinced.  Let's see what others think.

goldfirere:
I still lean against. It's not that I don't like the idea, but I don't think it's worth the "one more thing" burden.

nomeata:
Yeah, after the experience with `ExtraCommas` I am also leaning towards taking uncertainty as a sign that a proposal should probably be rejected swiftly.

tejon:
I do like the postfix `let` concept, but losing `with` is too steep a price. I would suggest `letting` but also, that's a different proposal.

simonmar:
The discussion appears to have died down, with the consensus tending towards rejecting the proposal.

I think despite the result, the discussion has been useful and has certainly pushed forwards our understanding of the design space. Hopefully this will result in an improved proposal in the future.

Any further comments before we close as rejected?

ndmitchell:
Unfortunately I think the discussion somewhat drifted off into postfix let, when I think that record wildcards is not a workable alternative to record updates. I don't think that a postfix let would answer the issues introduced in motivation.

I suggest any future proposal also incorporate the insights from @int-index in https://github.com/ghc-proposals/ghc-proposals/pull/87#issuecomment-511915852 about providing block syntax for more things than just record creations/updates.

nomeata:
> providing block syntax for more things than just record creations/updates

isovector:
@nomeata I suggest we not get too wrapped up in the herald syntax; regardless of what it looks like, it is going to be a great improvement to all codebases. Let's spend a few days, collect some ideas, run a quick poll on what the user base likes best (but *not* have an "other" option), and then just make it happen :)

sofusmortensen:
(I am very biased, to say the least, since I brainstormed this syntax for DAML together with ekmett back around Christmas 2017)

With DAML the motivation was to create a language easy to learn (for both non-developers and curly braces developers), yet powerful when mastered - we called that "low bar, high ceiling". Haskell definitely has "high ceiling" - but "low bar" it does not. 

We based DAML off Haskell with a few initiatives to simplify the learning experience: easy-to-use IDE, a simpler Prelude and a handful of syntactical desugaring with records being the most significant. 

I have plenty of experience with the result myself, but more importantly experience with dozens of people learning DAML (and unknowingly Haskell) for the first time. It is my experience that equivalent of this proposal had a significant impact making DAML both easier to learn and more pleasing to work with. 

Having left Digital Asset, but still being an eager (but novice) Haskell hobbyist, I am hoping for Haskell too to get a lower entry bar. Haskell too should be for everybody. I believe this proposal was a move in that direction.

goldfirere:
> I believe this proposal was a move in that direction.

Why? (I'm asking honestly -- not to goad.)

I would imagine that the braces and commas would be the easiest part of Haskell's record system. This proposal replaces easily-visible braces and commas with invisible whitespace, making indentation (and spaces vs tabs) suddenly relevant. Having taught Haskell several times now, indentation-awareness was always a major stumbling block. If the new syntax were somehow newline-aware but indentation-agnostic, I would believe a claim that it makes things simpler. But that ship has sailed for Haskell. (Full disclosure: I've never really liked layout -- it's just too easy for the existence or absence of one space to throw things off.)

sofusmortensen:
@goldfirere 

I found the `with` keyword was easy to explain as it had a "literate programming" feeling to it.  Objective was for a non-developer (someone smart with relevant domain expertise) to be able to guess what what code did (at least if written with that purpose in mind).  I found that newcomers found both record construction and copy/update to be particularly easy-to-read (ie. `foo with x = 1; z = 3`). 

Layout was already something our users needed to learn anyway (`where`, `do`, `let`, `of`) so there wasn't really an extra toll to pay here in connection with records - and the result was more (subjectively) pleasing to the eye. I am a fan of layout, and did push for it.  Sure there's been a few times faced with an obscure parser error where I have been in doubts whether that was the really right choice, but then I have reminded myself that it is optional and that it is (usually) making the result easy to read for the non-developer. 

I must admit that a significant part of what made DAML easier in connection with Records is captured in a previous proposal (https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0042-record-set-field.rst).

BTW I enjoyed your keynote at ZuriHac. Thanks for that.

goldfirere:
Very interesting. My extraction from @sofusmortensen's https://github.com/ghc-proposals/ghc-proposals/pull/231#issuecomment-512351274 is that the advantage to newcomers is the `with` keyword: this convenient English word makes the relationship between the constructor/record-to-be-updated and the fields clearer to non-developers. I buy that, and I see how a non-developer wouldn't get the same semantic clues from braces. But actually, I see this observation as layout-agnostic. That is, I think the same argument would apply to `MkRec with field1 = val1, field2 = val2 end` (where `end` is a keyword, too). I'm *not* proposing that, just thinking aloud.

Separately, you observe that layout hasn't been a problem for you and your students/trainees/domain experts. I wonder whether my struggle with layout is that my student body is different than yours: mine generally have not been domain experts: they have been college students, mostly without a great deal of technical experience. A professional engineer non-dev, on the other hand, will more easily be able to apply and understand the rules around layout.

Thanks for your very helpful explanation, @sofusmortensen!

rohanjr:
The summary from me is that I still believe ``with`` syntax is a simple improvement to usability for many Haskell users, both novice and experienced. It seems that the main barrier to acceptance is losing ``with`` as an identifier. As discussed (e.g. [here](https://github.com/ghc-proposals/ghc-proposals/pull/231#issuecomment-503533198)), there is an issue when defining a ``with`` identifier or using it unqualified (qualified works) with the extension enabled. I don’t think that’s reason enough to stop an optional and potentially widely appreciated extension. That is my opinion, but of course I’ll respect the decision of the committee.

isovector:
Losing `with` as an identifier is apparently not very painful: https://codesearch.aelve.com/haskell/search?query=%5Cbwith%5Cb%3D&filePath=.hs&insensitive=off&space=off&precise=off&sources=on

gridaphobe:
Hmm, your query is missing quite a few results.

https://codesearch.aelve.com/haskell/search?query=%5Ewith%20%3A%3A&filePath=.hs&insensitive=off&space=off&precise=off&sources=on

Still, I'm much less concerned about the loss of `with` as long as we preserve GHC's current behavior of lexing qualified keywords as identifiers (or at least make it an option).

simonmar:
> The summary from me is that I still believe `with` syntax is a simple improvement to usability for many Haskell users, both novice and experienced. It seems that the main barrier to acceptance is losing `with` as an identifier. As discussed (e.g. [here](https://github.com/ghc-proposals/ghc-proposals/pull/231#issuecomment-503533198)), this is an issue when defining `with` as an identifier or using it unqualified (qualified works) with the extension enabled. I don’t think that’s reason enough to stop an optional and potentially widely appreciated extension. That is my opinion, but of course I understand the opposing view that's been stated here.

I agree on both counts - the `with` syntax is a simple improvement that would probably be widely used, and that stealing `with` as a keyword is not necessarily a blocker.  However, I don't think these are the main concerns here - apologies if I'm misrepresenting those committee members who spoke up (please contradict me if so) but my impression is that we're not sure that this is the right extension. There appears to be a design space here with the potential to find an extension that would be more general and serve more uses than the current proposal, so it seems prudent to explore the design space before settling on something.

nomeata:
> There appears to be a design space here with the potential to find an extension that would be more general and serve more uses than the current proposal, so it seems prudent to explore the design space before settling on something.

Are your referring to

1.  a more general “layout for everything” extension, as brainstormed on https://gitlab.haskell.org/ghc/ghc/wikis/All-things-layout (which wasn’t a good choice, because there are no easy commenting and page watching features there), or
2. a more powerful “record with” syntax, which would for example allow type signatures and function definitions, e.g. moving it closer to `Foo {...} = where …`, albeit still with the checking that all defined names are indeed fields of the record? (As [observed by SPJ](https://github.com/ghc-proposals/ghc-proposals/pull/231#issuecomment-502760687).

So there are two ways into which we can make this one more general, and they go into different directions. 

(I am inclined to agree with Simon that, since this looks very much like bindings, _and_ using pattern matching in these bindings would genuinely add expressiveness and be very useful,while still being quite natural, that a `RecordWith` extension probably ought to do something like that.)

ndmitchell:
Please note that of the two possible future avenues number 1. would answer the motivation and desires expressed in the original proposal. In contrast, 2. is of no applicability to the original motivation at all. Telling people to give up on record syntax for wildcards is a dangerous thing to do, and I would consider it a very bad idea to have the "nice but dangerous" and "ugly but safe" syntaxes.

nomeata:
I am not sure I parse what you say about 2? I do _not_ propose using WildCards here. I mean that  a natural extension of 
```
some code that takes a record Foo where
  field1 = bar
  field2 = \x y -> x +y
```
is
```
some code that takes a record Foo where
  field1 :: String
  field1 = bar
  field2 x y = x + y
```
and my sense is that the latter, while making the extensions no longer a purely syntactic one, provides “more bang”, while satisfying the original motivation (because it is an extension over the proposed extensions).

ndmitchell:
Ah, my bad @nomeata - I missed that you are coming up with a new extension, distinct from the one SPJ proposed. Adding type signatures would still match the "feel" of the original proposal. However, adding additional functions would be dangerous once more (assuming that the additional functions had the same syntax as fields, and were thus thrown away if not needed). You would then have to introduce a way to "declaring" the functions, e.g. with let.

That said, I've never wanted the type signatures, and the idea of mixing record updates and defining helper functions seems unexpected to me, so I'm not sure 2. as described by @nomeata actually adds anything that people want. It's possible with more advanced forms of record overloading the types on fields might be handy, but I'd rather write them `field1 :: String = bar`, and I guess with other extensions that's now quite reasonable.

gridaphobe:
I interpreted @nomeata's extension as a rather conservative thing that would allow you to write 

```
some code that takes a record Foo where
  field1 :: String
  field1 = bar
  field2 x y = x + y
```

provided that `Foo` was defined as

```
data Foo = Foo
  { field1 :: String
  , field2 :: Int -> Int -> Int  -- or similar
  }
```

ie. it doesn't let you define arbitrary helper functions, but just provides a more convenient syntax for functional fields. I've seen people ask for this before, and agree that it would be nice to have.

ndmitchell:
Thanks @gridaphobe - I hadn't made that connection. That does start to look a worthwhile extension which gives more bang for buck and is very much inline with the original proposal.

nomeata:
As @gridaphobe says :-)

simonmar:
Kicking proposal into needs-revision given the preceding discussion.

PR #232, opened 2019-05-22
Ambiguous Type per-signature pragma

AntC2:
Per-signature pragma ``{-# AMBIGUOUS #-}`` to allow that signature's type to be ambiguous; instead of the module-wide ``-XAllowAmbiguousTypes``, which dangerously lifts ambiguity checking on all signatures.

Also tweak the error reporting to avoid recklessly suggesting users turn on ``-XAllowAmbiguousTypes``; and provide a flag ``-Wallowed-ambiguous-types`` that shows the ambiguity message as a warning.

[Rendered](https://github.com/AntC2/ghc-proposals/blob/patch-2/proposals/AmbiguousType-pragma.rst)

### DIFF
--- a/proposals/AmbiguousType-pragma.rst
+++ b/proposals/AmbiguousType-pragma.rst
@@ -91,7 +91,7 @@ Proposed Change Specification
 
 4. The error reporting from the ambiguity check that currently suggests ``To defer the ambiguity check to use sites, enable AllowAmbiguousTypes`` must make clear this is likely to entail using ``TypeApplications`` at usage sites, and that there are several possible approaches to avoid ambiguous type variables.
 
- Precise wording to be arrived at in discussion of this PR. (Prefer not mentioning ``AllowAmbiguousTypes`` at all.) Starting bikeshed::
+   Precise wording to be arrived at in discussion of this PR. (Prefer not mentioning ``AllowAmbiguousTypes`` at all.) Starting bikeshed::
  
      The ambiguity might be resolvable through TypeApplications at use sites. Then mark this signature as AMBIGUOUS
  
--- a/proposals/AmbiguousType-pragma.rst
+++ b/proposals/AmbiguousType-pragma.rst
@@ -77,15 +77,22 @@ Then this proposal firstly aims to avoid users blundering blindly into ambiguous
 Proposed Change Specification
 -----------------------------
 
-1. There is to be a pragma ``{-# AMBIGUOUS #-}``, to appear immediately after the ``::`` of a function or method definition's signature (so before the type). Not applicable for term type annotations beginning ``::``, nor for pattern signatures. Examples
+1. There is to be a pragma ``{-# AMBIGUOUS #-}``, to appear immediately after the ``::`` (so before the type), in principle wherever ``::`` can appear. That is a function or method definition's signature; term type annotations beginning ``::``; pattern signatures (and pattern synonyms). The "in principle" would also include ``::`` introducing kind signatures (type of types); and possibly places ready for not-yet-developed usages of type ``@`` applications. Examples
 ::
 
     f :: {-# AMBIGUOUS #-} C a => Int
 
     class Sized a  where
       sizeOf :: {-# AMBIGUOUS #-} Integer
+      
+    apply (x :: {-# AMBIGUOUS #-} forall a. (Read a, Show a) => String -> String) = x @Int "01"
+    
+    norm :: {-# AMBIGUOUS #-} forall a. (Read a, Show a) => String -> String
+    norm = show @a . read                                     -- needs type-lambda #155
+    
+    data T :: {-# AMBIGUOUS #-} F a -> Type                   -- ambiguous kind signature
 
-2. Signatures marked ``AMBIGUOUS`` are to be validated as if ``-XAllowAmbiguousTypes`` is set, for that signature only. (If that is already set module-wide, the pragma has no further effect.)
+2. Signatures marked ``AMBIGUOUS`` are to be validated as if ``-XAllowAmbiguousTypes`` is set, for that signature only. (If that is already set module-wide, the pragma has the effect of suppressing the ``-Wambiguous-type`` warning, see 5.)
 
 3. This does not change the validation for ambiguous types/type variables at usage sites.
 
@@ -95,24 +102,36 @@ Proposed Change Specification
  
      The ambiguity might be resolvable through TypeApplications at use sites. Then mark this signature as AMBIGUOUS
  
-5. There is to be a flag ``-Wallowed-ambiguous-types`` controlling whether a warning is raised for ambiguous types -- allowed either from the ``AMBIGUOUS`` pragma or ``-XAllowAmbiguousTypes``.
+5. There is to be a flag ``-Wambiguous-types`` controlling whether a warning is raised for ambiguous types -- as allowed by ``-XAllowAmbiguousTypes``. If the signature is also marked ``{-# AMBIGUOUS #-}``, then don't issue the warning (see "migration path" below). If a signature marked ``{-# AMBIGUOUS #-}`` is not in fact ambiguous, ignore.
 
+6. The pragma can only appear with an explicit ``::`` signature; not for terms where the inferred signature is ambiguous such as toplevel functions or instances::
+
+    data Option a
+    class C a
+    instance C (F b) => C (Option a)
+    
+   For those cases, the user must contrive an explicit signature (with ``-XInstanceSigs`` if necessary).
 
 
 
 Effect and Interactions
 -----------------------
+
 By lifting the ambiguity check only for signatures deliberately flagged, this ensures ambiguity checking does apply for the bulk of the signatures in the program *at the definition site*. Then ambiguity is less likely to manifest at *usage* sites, where it is more difficult to diagnose -- particularly if that is in a separate module.
 
 The proposed behaviour affects only validation and error/warning messages, not type checking rules or type inference.
 
 Existing code using ``AllowAmbiguousTypes`` is not affected. That is, ambiguities are not checked. The migration path is:
 
-* Switch on ``-Wallowed-ambiguous-types``; compile the module to examine signatures that are currently ambiguous.
+* Switch on ``-Wambiguous-types``; compile the module to examine signatures that are currently ambiguous.
 
-* If their ambiguity is expected and understood; mark as ``{-# AMBIGUOUS #-}``. Otherwise diagnose and correct.
+* If their ambiguity is expected and understood; mark as ``{-# AMBIGUOUS #-}`` (and that will suppress the warning). Otherwise diagnose and correct.
 
 * Remove the ``LANGUAGE AllowAmbiguousTypes`` setting and recompile.
+
+* In due course (not within scope of this proposal), deprecate the ``AllowAmbiguousTypes`` extension. (Same idea as introducing the ``OVERLAPPABLE`` and friends pragmas; then deprecating ``OverlappingInstances``/``IncoherentInstances``.)
+
+* Discussion on this proposal is going on in parallel with #234 'Local Warning Pragmas'. Because ``{-# AMBIGUOUS #-}`` is a language extension (an alternative to ``-XAllowAmbiguousTypes``), not merely controlling warnings, I see this as outside the scope that #234 has evolved to. (The ``-Wambiguous-types`` point 5. warning might fall within the 'Local Warning' in the sense of #234, but note that ``{-# AMBIGUOUS #-}`` in effect is a local suppression of that warning.)
 
 
 Costs and Drawbacks
@@ -140,7 +159,9 @@ Unresolved questions
 
 * Re pragmas that change semantics (such as the ``{-# OVERLAPPABLE #-}`` series), there has been comment they're difficult for source tooling utilities to observe. As well as the ``AMBIGUOUS`` pragma per signature, should there be a module-wide ``LANGUAGE`` setting? ``-XAllowAmbiguousTypesPragma``.
 
-* For modules containing more ambiguous types than not, so with ``AllowAmbiguousTypes`` switched on, should there be a per-signature pragma ``{-# NOAMBIGUOUS #-}`` that *does* apply the ambiguity check?
+* From discussion, decide against this: For modules containing more ambiguous types than not, so with ``AllowAmbiguousTypes`` switched on, should there be a per-signature pragma ``{-# NOAMBIGUOUS #-}`` that *does* apply the ambiguity check? That would prevent in future deprecating ``AllowAmbiguousTypes``.
+
+* If a signature marked ``{-# AMBIGUOUS #-}`` is not in fact ambiguous ...? A comment suggested warning of the non-ambiguity. Leave that to the implementer's judgment.
 
 
 Implementation Plan
--- a/proposals/AmbiguousType-pragma.rst
+++ b/proposals/AmbiguousType-pragma.rst
@@ -101,7 +101,13 @@ Proposed Change Specification
  
      The ambiguity might be resolvable through TypeApplications at use sites. Then mark this signature as AMBIGUOUS
  
-5. There is to be a flag ``-Wambiguous-types`` controlling whether a warning is raised for ambiguous types -- as allowed by ``-XAllowAmbiguousTypes``. If the signature is also marked ``{-# AMBIGUOUS #-}``, then don't issue the warning (see "migration path" below). If a signature marked ``{-# AMBIGUOUS #-}`` is not in fact ambiguous, ignore.
+5. There is to be a flag ``-Wambiguous-types`` controlling whether a warning is raised for ambiguous types -- as allowed by ``-XAllowAmbiguousTypes``. That is:
+
+   - If ``-XAllowAmbiguousTypes`` is not set, reject ambiguous signatures/don't also warn.
+   - If ``-XAllowAmbiguousTypes`` is set and the signature is also marked ``{-# AMBIGUOUS #-}``, then don't issue the warning (see "migration path" below).
+   - If a signature marked ``{-# AMBIGUOUS #-}`` is not in fact ambiguous, ignore.
+
+   ``-Wambiguous-types`` is to be in bin-of-warnings ``-W`` "normal warnings", on grounds an ambiguous signature is outside Haskell 2010. (If some time in future, ``-XAllowAmbiguousTypes`` is to be deprecated in favour of per-signature pragmas, move ``-Wambiguous-types`` into the ``-Wcompat`` bin.)
 
 6. The pragma can only appear with an explicit ``::`` signature; not for terms where the inferred signature is ambiguous such as toplevel functions or instances::
 
@@ -120,7 +126,7 @@ By lifting the ambiguity check only for signatures deliberately flagged, this en
 
 The proposed behaviour affects only validation and error/warning messages, not type checking rules or type inference.
 
-Existing code using ``AllowAmbiguousTypes`` is not affected. That is, ambiguities are not checked. The migration path is:
+Existing code using ``AllowAmbiguousTypes`` is not affected. That is, ambiguities are not checked. The migration path away from the module-wide setting for modules with ambiguous signatures is:
 
 * Switch on ``-Wambiguous-types``; compile the module to examine signatures that are currently ambiguous.
 
@@ -128,7 +134,11 @@ Existing code using ``AllowAmbiguousTypes`` is not affected. That is, ambiguitie
 
 * Remove the ``LANGUAGE AllowAmbiguousTypes`` setting and recompile.
 
-* In due course (not within scope of this proposal), deprecate the ``AllowAmbiguousTypes`` extension. (Same idea as introducing the ``OVERLAPPABLE`` and friends pragmas; then deprecating ``OverlappingInstances``/``IncoherentInstances``.)
+After this proposal is in place, with experience of how onerous or intrusive is the per-signature ``{-# AMBIGUOUS #-}``, a possible future migration path away from module-wide ``AllowAmbiguousTypes`` for GHC is
+
+* In due course (not within scope of this proposal), deprecate the ``AllowAmbiguousTypes`` extension in GHC, and eventually remove it. (Same idea as introducing the ``OVERLAPPABLE`` and friends pragmas; then deprecating ``OverlappingInstances``/``IncoherentInstances``.)
+
+Also note
 
 * Discussion on this proposal is going on in parallel with #234 'Local Warning Pragmas'. Because ``{-# AMBIGUOUS #-}`` is a language extension (an alternative to ``-XAllowAmbiguousTypes``), not merely controlling warnings, I see this as outside the scope that #234 has evolved to. (The ``-Wambiguous-types`` point 5. warning might fall within the 'Local Warning' in the sense of #234, but note that ``{-# AMBIGUOUS #-}`` in effect is a local suppression of that warning.)
 
@@ -160,7 +170,7 @@ No contrary feedback received for these questions, so left here as visible for C
 
 * [Implementor's judgment:] Re pragmas that change semantics (such as the ``{-# OVERLAPPABLE #-}`` series), there has been comment they're difficult for source tooling utilities to observe. As well as the ``AMBIGUOUS`` pragma per signature, should there be a module-wide ``LANGUAGE`` setting? ``-XAmbiguousTypesPragma``.
 
-* [From discussion, decide against this:] For modules containing more ambiguous types than not, so with ``AllowAmbiguousTypes`` switched on, should there be a per-signature pragma ``{-# NOAMBIGUOUS #-}`` that *does* apply the ambiguity check? That would prevent in future deprecating ``AllowAmbiguousTypes``.
+* [From discussion, decide against this:] For modules containing more ambiguous types than not, so with ``AllowAmbiguousTypes`` switched on, should there be a per-signature pragma ``{-# NO[T_]AMBIGUOUS #-}`` that *does* apply the ambiguity check? That would prevent in future deprecating ``AllowAmbiguousTypes``.
 
 * [Implementor's judgment:] If a signature marked ``{-# AMBIGUOUS #-}`` is not in fact ambiguous ...? A comment suggested warning of the non-ambiguity. 
 
--- a/proposals/AmbiguousType-pragma.rst
+++ b/proposals/AmbiguousType-pragma.rst
@@ -208,7 +208,7 @@ Appendix: BNF diff for appearences of ``{-# AMBIGUOUS #-}``
     exp     → infixexp :: [ {-# AMBIGUOUS #-} ] [context =>] type      (expression type signature)
             | infixexp
             
-    fexp    → [fexp] [@([ {-# AMBIGUOUS #-} ] type)] aexp	       (function application with type applicn)
+    fexp    → [fexp [@([ {-# AMBIGUOUS #-} ] type)] ] aexp	       (function application with type applicn)
 
     gendecl → vars :: [ {-# AMBIGUOUS #-} ] [context =>] type          (type signature)
             | fixity [integer] ops                                     (fixity declaration)
@@ -223,11 +223,11 @@ Appendix: BNF diff for appearences of ``{-# AMBIGUOUS #-}``
             
     topdecl → ...
             | data simpletype where [gconstrs] [deriving]              (GADT style data or newtype)
-            | newtype [context =>] simpletype where [newgconstrs] [deriving]
+            | newtype simpletype where [newgconstrs] [deriving]
             | ...
             
     gconstrs → gconstr1 | … | gconstrn	                               (n ≥ 1)
-    gconstr → con [{ fielddecl1 , … , fielddecln }] :: [ {-# AMBIGUOUS #-} ] [context =>] type  
+    gconstr → con :: [ {-# AMBIGUOUS #-} ] [{ fielddecl1 , … , fielddecln }] [context =>] type  
                                                    	               (n ≥ 0, GADT style constructor)
                                                                        (GADT style newtype constr likewise)
 
--- a/proposals/AmbiguousType-pragma.rst
+++ b/proposals/AmbiguousType-pragma.rst
@@ -91,20 +91,20 @@ Proposed Change Specification
     
         data T :: {-# AMBIGUOUS #-} F a -> Type                   -- ambiguous kind signature
         
-   Also ``{-# AMBIGUOUS #-}`` can appear after ``::`` in a ``data`` or ``newtype`` declaration -- either for fields or GADT-style constructors; or in H2010-style immediately after the constructor::
+   Also ``{-# AMBIGUOUS #-}`` can appear after ``::`` in a ``data`` or ``newtype`` declaration -- either for fields or GADT-style constructors; or in H2010-style immediately ~after~ before the constructor (that is, after ``=`` or ``|``, because no ``::``)::
    
         data D a where                                            -- GADT style
           MkD :: {-# AMBIGUOUS #-} (forall b. C a b => a)
 
-        data D2 a = MkD2  {-# AMBIGUOUS #-} (forall b. C a b => a) (Show a => a) 
+        data D2 a = {-# AMBIGUOUS #-} MkD2  (forall b. C a b => a) (Show a => a) 
                                                -- H2010 style; note no ::; note the second arg is not ambiguous
                                                
-        newtype N a = MkN {-# AMBIGUOUS #-} { foo :: {-# AMBIGUOUS #-} (forall b. C a b => a) } 
+        newtype N a = {-# AMBIGUOUS #-} MkN { foo :: {-# AMBIGUOUS #-} (forall b. C a b => a) } 
                                                -- AMBIGUOUS not ambiguous, but one is redundant
 
-   ``{-# AMBIGUOUS #-}`` is also to be allowed in the body of a type application::
+   For type applications, see 7. <strike>``{-# AMBIGUOUS #-}`` is also to be allowed in the body of a type application::
 
-        x = foo @({-# AMBIGUOUS #-} forall b. C a b => a) bar
+        x = foo @({-# AMBIGUOUS #-} forall b. C a b => a) bar</strike>
    
    See Appendix giving a diff to the BNF for the language (wrt the Report, and affected extensions).
 
@@ -133,6 +133,8 @@ Proposed Change Specification
     instance C (F b) => C (Option a)
     
    For those cases, the user must contrive an explicit signature (with ``-XInstanceSigs`` if necessary).
+
+7. Type applications are to be allowed ambiguous signatures, no warning or error, without needing ``{-# AMBIGUOUS #-}`` (nor needing ``LANGUAGE AllowAmbiguousTypes``).
 
 
 
--- a/proposals/AmbiguousType-pragma.rst
+++ b/proposals/AmbiguousType-pragma.rst
@@ -131,7 +131,7 @@ Proposed Change Specification
     
    For those cases, the user must contrive an explicit signature (with ``-XInstanceSigs`` if necessary).
 
-7. ``-XTypeApplications`` is to be changed, so that when enabled, ``f @t`` is valid regardless of whether ``t`` is ambiguous. That is, Type applications are to be allowed ambiguous signatures, no warning or error, without needing ``{-# AMBIGUOUS #-}`` (nor needing ``LANGUAGE AllowAmbiguousTypes``). For example, this is currently rejected without ``AllowAmbiguousTypes``:
+7. ``-XTypeApplications`` is to be changed, so that when enabled, ``f @t`` is valid regardless of whether ``t`` is ambiguous. That is, Type applications are to be allowed ambiguous signatures, no warning or error, without needing ``{-# AMBIGUOUS #-}`` (nor needing ``LANGUAGE AllowAmbiguousTypes``). For example, this is currently rejected without ``AllowAmbiguousTypes``::
 
         x = foo @(forall b. C a b => a) bar
 
@@ -178,7 +178,7 @@ Do nothing. That is, continue with the module-wide ``AllowAmbiguousTypes`` setti
     
 I would disagree with that "useless". I see the confusion they cause as harmful. Especially because that follows from the error message's misleading ``enable AllowAmbiguousTypes``.
 
-Re Type Applications [Section 2 Point 7.] possibly changing to silently accept ambiguous signatures might be considered too radical of a change. (Although this is the opposite of a breaking change: it will accept more programs, without needing ``LANGUAGE AllowAmbiguousTypes``.) Then Point 7. could require an ``{-# AMBIGUOUS #-}`` pragma after the ``@``, and syntactically would need the signature and pragma enclosed in parens, for example:
+Re Type Applications [Section 2 Point 7.] possibly changing to silently accept ambiguous signatures might be considered too radical of a change. (Although this is the opposite of a breaking change: it will accept more programs, without needing ``LANGUAGE AllowAmbiguousTypes``.) Then Point 7. could require an ``{-# AMBIGUOUS #-}`` pragma after the ``@``, and syntactically would need the signature and pragma enclosed in parens, for example::
 
     x = foo @({-# AMBIGUOUS #-} forall b. C a b => a) bar
 
--- a/proposals/AmbiguousType-pragma.rst
+++ b/proposals/AmbiguousType-pragma.rst
@@ -30,29 +30,82 @@ Proposal title
 .. sectnum::
 .. contents::
 
-Text for pull request [move there when PR created].
+.. Text for pull request [move there when PR created]:
 
-Per-signature pragma ``{-# AMBIGUOUS #-}`` to allow that signature's type to be ambiguous; replacing the module-wide ``-XAllowAmbiguousTypes``, which dangerously lifts ambiguity checking on all signatures.
+Per-signature pragma ``{-# AMBIGUOUS #-}`` to allow that signature's type to be ambiguous; instead of the module-wide ``-XAllowAmbiguousTypes``, which dangerously lifts ambiguity checking on all signatures.
 
 Also tweak the error reporting to avoid recklessly suggesting users turn on ``-XAllowAmbiguousTypes``; and provide a flag ``-Wallowed-ambiguous-types`` that shows the ambiguity message as a warning.
 
-To do: Here you should write a short abstract motivating and briefly summarizing the proposed change.
+.. Here you should write a short abstract motivating and briefly summarizing the proposed change.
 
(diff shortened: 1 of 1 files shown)


goldfirere:
I'm in strong support of this well-written proposal, but please permit me to offer some suggestions:

- I think `{-# AMBIGUOUS #-}` should be allowed wherever type annotations are allowed. Beyond just variable and method signatures, this includes in expressions, patterns, and pattern synonym signatures. For an example of when you need this in a pattern:

```
apply (x :: forall a. (Read a, Show a) => String -> String) = x @Int "01"
```

Though we can't use this today, with [my "type-lambda" proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0050-type-lambda.rst), we could write

```
norm :: forall a. (Read a, Show a) => String -> String
norm = show @a . read
```

and then call `apply (\ @a -> norm @a)`.

- With `-Wallowed-ambiguous-types`, do you mean that a warning would be raised even if the user includes `{-# AMBIGUOUS #-}`? That seems off to me. One guideline I have for a warning is that there should be some action the user can take to make the warning go away (other than just saying `-Wno-xyz`). I *could* see warning on ambiguous types with `-XAllowAmbiguousTypes` though.

- I would support a more ambitious proposal that sets a deprecation strategy for `-XAllowAmbiguousTypes`. Though maybe some programs have so many ambiguous types that this would be too annoying. In that case, I like the `{-# NOT_AMBIGUOUS #-}` idea (spelling slightly changed).

- Unlike `{-# OVERLAPPABLE #-}` and friends, which do change semantics and probably shouldn't be pragmas (but, instead, something else, like a keyword), this one does not change semantics. It just skips a check. Yes, yes, that could be called a change in semantics, but given a program accepted with `{-# AMBIGUOUS #-}` either (1) the same program would have been accepted with the same meaning but without the pragma, or (2) the program would be rejected without the pragma. In no case does the pragma change the name of an accepted program, and so (to me) a pragma is appropriate here.

Thanks for writing this up @AntC2.

int-index:
> either (1) the same program would have been accepted with the same meaning but without the pragma, or (2) the program would be rejected without the pragma.

Same goes for `{-# OVERLAPPABLE #-}`, no?

adamgundry:
I echo @goldfirere's support, thanks @AntC2!
* Perhaps `-Wambiguous-types` is a shorter and equally clear alternative to `-Wallowed-ambiguous-types`? (The extension could really have been called `AmbiguousTypes` in the beginning -- we don't say `AllowRankNTypes` after all!)
* I agree that a warning should be issued when `AllowAmbiguousTypes` is used to permit an ambiguity, but not when the `{-# AMBIGUOUS #-}` pragma is used. If I'm converting a codebase to this wonderful new world, it would be nice to set `-Werror=ambiguous-types`, fix the errors and then stop. One can always `grep` for `AMBIGUOUS`, after all.
 * Presumably `-Wall` will not include the new warning, at least initially?
 * Do I understand correctly that a type marked `{-# AMBIGUOUS #-}` is permitted to not, in fact, be ambiguous? Or should it be rejected with an error telling the user to remove the redundant pragma?
 * I'm not convinced `{-# NO[T_]AMBIGUOUS #-}` is worth the candle.

AntC2:
> @goldfirere Unlike `{-# OVERLAPPABLE #-}` and friends, which do change semantics and probably shouldn't be pragmas (but, instead, something else, like a keyword), this one does not change semantics. ... either (1) the same program would have been accepted with the same meaning but without the pragma, or (2) the program would be rejected without the pragma. ... and so (to me) a pragma is appropriate here.

Yes. My chief concern is with sourcecode tooling and people importing this program as a package: they'll need to know they're in for `TypeApplications`. So do we provide some indication at the top of the program/in the `LANGUAGE` settings? I guess the tooling could go looking for pragmas, but they're a little tricky to detect mechanically because they're treated as comments and they might have arbitrary whitespace inside.

> @int-index Same ["does not change semantics"] goes for `{-# OVERLAPPABLE #-}`, no?

Actually no: if an instance is marked `OVERLAPPABLE`, GHC somewhat avoids usage sites committing to it. So you can [somewhat avoid orphan instance problems](https://gitlab.haskell.org/ghc/ghc/issues/16241). As opposed to marking the `OVERLAPPING` in the importing instance, which merely accepts the instance and imports the orphan. (Not a lot of people know that; and it's highly unreliable/dependent on optimisation settings, for example.)

AntC2:
> @goldfirere I think `{-# AMBIGUOUS #-}` should be allowed wherever type annotations are allowed. 

Ah, OK. I was thinking `AMBIGUOUS` wouldn't make sense in a pattern signature/synonym. I don't have your ingenuity to think of weird places to put @pplications. ;-)

> warning would be raised even if the user includes `{-# AMBIGUOUS #-}`? That seems off to me.
> [and @adamgundry agrees]  it would be nice to set `-Werror=ambiguous-types`, fix the errors and then stop

OK fair point re "and then stop". I was trying to keep it simple for the implementer.

> @goldfirere  would support a more ambitious proposal that sets a deprecation strategy for `-XAllowAmbiguousTypes`.

Yes that was my thinking in the Interactions section "migration path". Without saying it out loud for fear of upsetting @pplication-aholics, who I expected to say this:

>  I like the `{-# NOT_AMBIGUOUS #-}` idea

That only makes sense if we _don't_ deprecate `-XAllowAmbiguousTypes`. Or am I not understanding?

> @adamgundry Do I understand correctly that a type marked {-# AMBIGUOUS #-} is permitted to not, in fact, be ambiguous?

;-) That's the current behaviour with `-XAllowAmbiguousTypes`, because the alternative would be hugely annoying (and probably impossible). Then I was trying to keep proposed behaviour as close as possible to current.

> Or should it be rejected with an error telling the user to remove the redundant pragma?

I'm happy to take a recommendation. Error or only warning? How easy to implement?

RyanGlScott:
I'm definitely in support of controlling the scope of `AllowAmbiguousTypes` to be on a finer granularity than what it currently is. I'm not entirely sure if a pragma is the best way to accomplish this.

In the past, there have been many proposals to allow controlling warnings on a per-identifier basis, such as [this GHC issue](https://gitlab.haskell.org/ghc/ghc/issues/13841) suggesting an `{-# ADOPT #-}` pragma for suppressing orphan instance–related warnings. By response to this is usually of the form: instead of concocting new pragmas for each warning we wish to suppress, why not have a general mechanism that could work for _any_ warning by delimiting regions of code that should have the warning disabled? This is the subject of GHC issue [#602](https://gitlab.haskell.org/ghc/ghc/issues/602), and while no one has yet proposed a concrete design for how this would look, I think there is a lot of merit in considering this.

Granted, `AllowAmbiguousTypes` isn't really about warnings _per se_, so it's in a slightly different territory. But at the same time, I do think it would be worthwhile to consider it as a possible way to achieve the end goal of enabling ambiguity in controlled bursts. For instance, this proposal states that:

> There is to be a pragma `{-# AMBIGUOUS #-}`, to appear immediately after the `::` of a function or method definition's signature (so before the type). Not applicable for term type annotations beginning `::`, nor for pattern signatures.

It is worth noting that this specification prevents the use of `{-# AMBIGUOUS #-}` in certain places where ambiguity can bite, such as in the following code:

```
type family F a

g (x :: F a) = x
```

Compiling this without `AllowAmbiguousTypes` will yield:

```
    • Couldn't match type ‘F a0’ with ‘F a’
      Expected type: F a -> F a
        Actual type: F a0 -> F a0
      NB: ‘F’ is a non-injective type family
      The type variable ‘a0’ is ambiguous
    • In the ambiguity check for the inferred type for ‘g’
      To defer the ambiguity check to use sites, enable AllowAmbiguousTypes
      When checking the inferred type
        g :: forall a. F a -> F a
  |
7 | g (x :: F a) = x
  | ^^^^^^^^^^^^^^^^
```

However, since `g` does not have a type signature, one has no way of marking it as ambiguous with an `{-# AMBIGUOUS #-}` pragma. Of course, one might argue that having to write an explicit type signature is the price one must pay for using `{-# AMBIGUOUS #-}`. Fair enough, although there are other places where ambiguity can creep in that are less obvious how to fix:

1. Instance declarations, as in the following example:

   ```hs
   data Option a
   class C a
   instance C (F b) => C (Option a)
   ```

   There is no `::` in an instance declaration, so (as far I can tell) it cannot be combined with an `{-# AMBIGUOUS #-}` pragma.
2. Data declarations, such as `data T :: F a -> Type` (example taken from [this issue](https://gitlab.haskell.org/ghc/ghc/issues/14419)) don't seem to be compatible with `{-# AMBIGUOUS #-}` pragmas. This would be possible if we had [top-level kind signatures](https://github.com/ghc-proposals/ghc-proposals/pull/54), I suppose, but until they are implemented, this poses a challenge.

-----

Having written all of this up, I realize that it's somewhat unfair to oppose this proposal (which, by all accounts, is very well motivated and clear in its scope) on the grounds that it should be superseded by a more ambitious proposal without a clear design. In that case, I suppose I could live with my idea simply being mentioned as an alternative. If that is what is decided, however, I would like to see the proposal amended to mention the following:

1. A mention of places where `{-# AMBIGUOUS #-}` pragmas can _not_ be used (e.g., functions without top-level signatures, instance declarations, etc.), and
2. For each such place, a justification for why it is acceptable not to allow `{-# AMBIGUOUS #-}` pragmas there.

AntC2:
Thanks @RyanGlScott, re your specific point

> It is worth noting that this specification prevents the use of `{-# AMBIGUOUS #-}` in certain places ... [pattern signatures]

Yes, @goldfirere said the same. That's exposing my paucity of imagining how devious `@`pplicationists can be. I wanted the proposal to be narrowly scoped to start with. So I've left the proposal as written for a while to see other feedback. But I'm happy to change it to: `{-# AMBIGUOUS #-}` can appear wherever `::` can appear.

I wanted to narrow the scope because there's many causes/places GHC says `The type variable ‘a0’ is ambiguous`. By no means all of them can be suppressed with `AllowAmbiguousTypes`.

As audience for this change I have in mind not GHC HQ/the people who typically respond to these proposals, but the people who ask questions on StackOverflow/are confused and hoodwinked by GHC saying `To defer the ambiguity check to use sites, enable AllowAmbiguousTypes`/can't discriminate when that's sensible vs when it'll just make for more trouble.

The sort of people who say ["I'm far from an expert on type inference, so I can't really offer a
better inference algorithm that would avoid this problem. The best you can do, as far as I can tell, is to enable `AllowAmbiguousTypes` and hope for the best."](https://mail.haskell.org/pipermail/ghc-devs/2019-May/017664.html). A compiler really shouldn't leave people in the lurch to 'hope for the best'. Or ponder "If we `AllowAmbiguousTypes`, then the definition of `Q3` is accepted, but
`q3` fails with a similar error, complaining that GHC cannot match `a` with `Bool` because it is "untouchable". This seems confusing--" [same thread at GHC HQ]

As to your more general (and frankly rather ineffable point) about "delimiting regions of code" ... Thanks for drawing attention to #602. I've put a response on there, in general (but not ineffably general) terms.

> Granted, `AllowAmbiguousTypes` isn't really about warnings _per se_, ...

That's right. It's suppressing what would otherwise be a rejection/error, not a warning. Furthermore, it exhibits what I might call 'action at a distance': we are in effect not _suppressing_ an error check at the definition site so much as _mandating_ `@`pplication at the usage site -- which might be in another module, certainly not in the "regions of code" where the ambiguous definition appears.

The more I thought about #602, the more it seemed to me there were plenty of examples of this 'action at a distance' such that "regions of code" is not a helpful delimitation. (Try arriving at a sensible delimitation of overlapping instances in terms of "regions of code".)

Whether or not something more effable can be rescued from #602, currently it seems to me poorly argued, unsupported by use cases, and just plain wrong as stated. I see no merit. Also `AllowAmbiguousTypes` -- and particularly GHC's current recklessly suggesting it -- is doing clear harm and confusion currently. (See my recent comments on proposal #148 for concrete examples.) Waiting for #602 (a decade so far, and still counting) would be a clear case of making the (arguably) perfect the enemy of saving continued harm.

AntC2:
Ah, to sweep up a couple of @RyanGlScott's other examples (seeking feedback whether this makes sense)

> Instance declarations, as in the following example:
>
>        data Option a
>        class C a
>        instance C (F b) => C (Option a)
> There is no `::` in an instance declaration,

With `-XInstanceSigs` there is, so assuming `class C` has a method

    class C a  where methC :: blah
    instance C (F b) => C (Option a)  where
      methC :: {-# AMBIGUOUS #-} C (F b) => ... (Option a) ...

? If `C` doesn't have a method, it can't appear in `@`pplications, only constraints/signatures. What would be a use case? Should that instance be accepted anyway?


Following the "wherever `::` can appear" new rule:

    data T :: {-# AMBIGUOUS #-} F a -> Type

(Whatever it is you're trying to do with `T`.)

> functions without top-level signatures,

Sorry, don't follow: if there's no signature, there can't be a tyvar to be ambiguous. (If the function's type is given via pattern signatures/term annotations, the "wherever `::` can appear" new rule would cover?)

AntC2:
Thanks @goldfirere , @RyanGlScott. I've incorp'd your feedback as follows:

* `{-# AMBIGUOUS #-}` to be valid following any `::` -- even where such an ambiguous signature only makes sense with future features (such as type-lambda) or with kind `@`pplications. (The argument being such sites are today allowed to be ambiguous with `AllowAmbiguousTypes`. See examples at point 1.)
* But not for other inferred types. That is, you need to put an explicit sig for toplevel functions or `-XInstanceSigs`.

* Rejected the possibility of a `{-# NO[T_]AMBIGUOUS #-}` pragma: that would interfere with the long-term goal of deprecating `AllowAmbiguousTypes`.

* `{-# AMBIGUOUS #-}` to suppress warning messages for specific sigs that would otherwise appear with `-XAllowAmbiguousTypes` and `-Wambiguous-types`.

I'm surprised feedback on this has dried up so quickly. I'll be ready to pass for committee consideration within a couple of days, unless there's violent disagreement. (At that point, I'll make more executive decisions re the 'Unresolved questions')

goldfirere:
Thanks for this updates. I'm in support.

AntC2:
Hi @nomeata, please submit this for Committee discussion.

No further feedback received on the 'Unresolved questions', so I've frozen in a resolution for each, but left them visible for discussion. No strong objections from anybody, but OTOH very few explicitly supporting.

There was [another example on StackOverflow this month](https://stackoverflow.com/questions/56487391/multiparamtypeclasses-why-is-this-type-variable-ambiguous) of somebody writing a class with an ambiguous method signature; getting the rejection message `blah blah, enable AllowAmbiguousTypes`; and consequently getting a landslide of "confusing to me" ambiguity rejections at usage sites; then only _after_ that coming to ask what was ambiguous.

goldfirere:
I will be shepherding this proposal. Before making my recommendation (which is expected to be "accept"), I would like @AntC2 to clarify a few issues:

* The migration path. That section of the proposal appears to mix actions the user is expected to take and actions that GHC developers are supposed to take. Could you clarify this?

* What bin of warnings with the new warning go in? It could be any of on-by-default, `-W`, `-Wall`, `-Wcompat`, I think.

Thanks!

AntC2:
Thanks @goldfirere, before I clarify the proposal could I clarify ...

* This proposal is not aiming to start a path towards deprecating module-wide `AllowAmbiguousTypes`. (I say that's out of scope/a possible future.) That's because the feedback (especially from you) was that some people expect all/most signatures to be ambiguous, then festooning everything with `{-# AMBIGUOUS #-}` would be onerous and distracting. I think we need some road miles with the per-sig pragma to find a consensus. Then most of the 'migration path' is how to migrate a module with a few ambiguous signatures towards using per-sig pragmas. I'll split that 'migration path' section to reflect more strongly what's future scope.

* In that case re bin of warnings, I think `-Wcompat` would not be appropriate until/unless we're on a deprecation path. (I presume it's relatively easy to switch a warning into a different bin later?) `-Wambiguous-types` only makes sense if `LANGUAGE AllowAmbiguousTypes` is set. If not allowed, ambiguous types get rejected. It would be daft to get both a rejection and a warning for a signature. Then on-by-default would be too annoying. I'm thinking the appropriate bin is `-W` "normal warnings", on grounds an ambiguous signature is outside Haskell 2010.

I'll clarify the wording to say that if `Wambiguous-types` is on but not `AllowAmbiguousTypes`, don't issue warnings/do issue rejections.

goldfirere:
It sounds like you've clarified on your own. :) You're zeroing in on exactly the changes I wanted. Thanks!

AntC2:
OK done. Over to @goldfirere ...

goldfirere:

**Summary**

1. Proposed is to add a new `{-# AMBIGUOUS #-}` pragma, usable in prefix wherever types appear. When this pragma is in place, the ambiguity check is skipped.

2. Additionally, proposed is to add a warning `-Wambiguous-types`, part of `-W`, that triggers whenever an ambiguous type is accepted without the pragma (this can happen only with `-XAllowAmbiguousTypes`).

3. The proposal also suggests rewording of the current suggestion to add `-XAllowAmbiguousTypes`, informing the user that they probably want to fix their type, not enable an extension.

**Recommendation**

I recommend acceptance. I am wholly in support of points (1) and (3) above, and weakly in support of (2). Module-wide language settings should either be impossible (or very difficult) to specify locally, or should be agreed to be benign. `AllowAmbiguousTypes` is neither: a user should know precisely where they're writing ambiguous types (users will have to use type application, after all), and it is an advanced feature with the capability of causing obscure type errors in clients.

My lukewarm support of (2) is that it's a bit annoying for users to have to specify both `-XAllowAmbiguousTypes` *and* `{-# AMBIGUOUS #-}`. I do see the motivation to do so, however. And users who don't wish to do this can just `-XAllowAmbiguousTypes -Wno-ambiguous-types`, so it's not so terrible. In the end, I do think obviating `-XAllowAmbiguousTypes` would be a good thing, which is why I have any support at all for (2).

I look forward to your discussion, committee. As usual, silence will be taken as agreement.

goldfirere:
@ghc-proposals/ghc-steering-committee see immediately above for my recommendation.

simonpj:
I like this proposal, for the reasons set out in the proposal and Richard's note above.

My one ask is that the proposal gives a diff to the BNF for the language (wrt the Report).  Please!

Richard says

> My lukewarm support of (2) is that it's a bit annoying for users to have to specify both -XAllowAmbiguousTypes and {-# AMBIGUOUS #-}

This doesn't seem right.  Users don't have to specify both.   As the proposal says, given a module that _already_ has `-XAllowAmbiguousTypes` a user can 
* Switch on the warning
* Suppress the warnings on a case by case basis with `{-# AMBIGUOUS #-}`
* Finally remove the module wide `-XAllowAmbiguousTypes`.

No need to "specify both".

AntC2:
Thanks @simonpj 

> My one ask is that the proposal gives a diff to the BNF for the language (wrt the Report).

Um? The only change to syntax is to introduce `{-# AMBIGUOUS #-}`. Syntactically that's a comment, so already allowed by the BNF. (Then strictly speaking it could occur anywhere a comment could occur. Only if it immediately follows `::` does it trigger this proposal.) 

Perhaps somebody could point me to how the BNF goes with (say) the `{-# OVERLAPPABLE #-}` pragma? I guess ... (the optional `[ {-# AMBIGUOUS #-} ]` is the change):

    exp     -> infixexp :: [ {-# AMBIGUOUS #-} ] [context =>] type      (expression type signature)
             | infixexp

    gendecl -> vars :: [ {-# AMBIGUOUS #-} ] [context =>] type          (type signature)
             | fixity [integer] ops                                     (fixity declaration)
             |                                                          (empty declaration)


Also following `::` in these, none of which are Report syntax:

* a GADT style constructor decl (does that follow `gendecl vars`?);
* a datatype field name signature with `forall` and/or context -- i.e. existentials;
* a pattern signature;
* a pattern synonym signature;
* a kind signature (in class/instance heads).

Post-thought: the rhs of a `type` synonym decl with a context (also not Report syntax).

goldfirere:
It should also be allowed in the body of a type application (also no BNF to make a delta from).

simonmar:
Is `-Wambiguous-types` included in `-Wall` or not?

AntC2:
> Is `-Wambiguous-types` included in `-Wall` or not?

`-Wambiguous-types` is to be included in `-W`, says Section 2 point 5.

I'm going by [the User Guide](https://downloads.haskell.org/~ghc/8.6.5/docs/html/users_guide/using-warnings.html), which I presume means that everything in `-W` is _ipso facto_ in `-Wall`.

Is that not so? The User Guide isn't very explicit in the way a literal-minded implementer would like.

treeowl:
I really don't like the idea of including this in -Wall.

AntC2:
> I really don't like the idea of including this in `-Wall`.

Does that mean you're happy it's in `-W`? Are there examples of other warnings that are in `-W` but not in `-Wall`? That seems strange.

Note if what you mean is you don't like the idea of getting warnings for ambiguous types, then flag those signatures as `{-# AMBIGUOUS #-}` to suppress the warning.

treeowl:
I don't know anything about -W. I would prefer to have to turn the warning
on manually.

AntC2:
> @treeowl I really don't like the idea ...  I would prefer ...

Could you explain the use case(s) you have in mind; you're being enigmatic. Re `-W`, please read the User Guide at the link I gave. (I'm trying to follow policy.)

Similarly, is @simonmar's q motivated by a use case?

What's motivating me is the amount of confusion and harm GHC's error message is doing currently for learners and moderately experienced users. Even for people deliberately using type `@`pplication extensively, I suspect they're giving plenty of signatures they don't intend to be ambiguous. Wouldn't it be better to warn of unintended ambiguity at the declaration site rather than delayed to the usage site?

simonpj:
>  I guess ... (the optional [ {-# AMBIGUOUS #-} ] is the change):

That is helpful thank you.  In the [Hakell 2010 report syntax](https://www.haskell.org/onlinereport/haskell2010/haskellch10.html#x17-18000010.5) there are four other occurrences of "::", in
* foreign import
* foreign export
* newconstr
* fielddecl
I wonder if the pragma has any effect there?  I think the answer is "yes, for all", but it would be great to say so explicitly.

I think perhaps you mean that {-# AMBIGUOUS #-} is recognised precisely and only after any occurrence of "::" in any language construct. 

For type synonyms see `Note [When we don't check for ambiguity]` in `TcValidity` which says that we never check for ambiguity there.

Making this precise in the proposal will guide both implementors and users.  Thanks!

AntC2:
Thank you again @simonpj I might need a little more guidance.

As I understand it, you can only get an ambiguous signature if it has a constraint (or an existential tyvar). The constraint might be implied rather than explicit, as for class method sigs or instance sigs.

Then the constructs in your 4 bullets don't allow constraints in H2010. (And for fielddecls, I'd already said that -- I intended it to include fields in newconstrs.) I'm happy to add those 4, but is the syntactically-based rule to be

a) anywhere following any appearance of `::`; or
b) only those appearances of `::` which could be followed by an ambiguous signature?
c) Does GHC make any such syntactic distinction; or
        does GHC potentially parse a context on any signature then later reject it if disallowed semantically?

I mean 'rule to be' so that the spec can " guide both implementors and users".

* foreign import } can a foreigner carry a context? (not according to the syntax)
* foreign export }    how would the FFI handle it?
* newconstr  
* fielddecl

Those last two raise another difficulty darn it! consider

    newtype N a = MkN { foo :: (forall b. C a b => a) }

    data D a = MkD (forall b. C a b => a)

For `newtype` or `data` + existential , whether or not there's a fielddecl with an explicit `::`, you might get an ambiguous sig, and GHC currently volunteers `enable AllowAmbiguousTypes`. But we want to avoid ambiguous syntax for the pragma, consider

    data D a = MkD {-# AMBIGUOUS #-} (forall b. C a b => a)            -- OK, but no ::

    newtype N a = MkN {-# AMBIGUOUS #-} { foo :: {-# AMBIGUOUS #-} (forall b. C a b => a) }  -- yeuch

    data D2 a = MkD2  {-# AMBIGUOUS #-} (forall b. C a b => a) (Show a => a)    -- hmm?

`MkD2`'s second arg is not ambiguous.

> For type synonyms ... we never check for ambiguity there.

Ok that's great thank you.

AntC2:
Expanded Section 2 point 1 to identify more places ``AMBIGUOUS`` might occur (with examples), per recent discussions.

Added Appendix with a stab at the BNF diff. (I had to guess quite a bit of the BNF for post-H2010.)

For an ambiguous `conop` declared infix, the syntax is going to look weird. Any suggestions for improvement? (My head is hurting.)

    data D2 a = MkD2  {-# AMBIGUOUS #-} (forall b. C a b => a) (Show a => a) 
              | {-# AMBIGUOUS #-} (Show a => Int) :&& Bool

goldfirere:
Let's take a small step back here. Instead of looking for every place we can shoehorn the pragma, let's consider where it's useful.

1. Identifiers with ambiguous types are useful: they can be specialized by type application. Thus: everywhere a type describes the type of an identifier is a good place for the pragma.

2. An expression `(expr :: ty)` behaves like `let fresh :: ty; fresh = expr in fresh`. Thus an expression signature is like an identifier and reduces to case (1).

3. Ambiguity in type applications is questionable: it can happen only in the presence of the `-XImpredicativeTypes` bug. I would actually argue that, given that this type is just being used to substitute somewhere else (as opposed to being assigned to an identifier), we should skip the ambiguity check in type applications always.

4. Non-record constructors fields don't need an independent ambiguity check, but it is useful to have per-field ambiguity checks on record fields (see (1)).

I'd like to propose that, for H98-syntax constructors, we put the pragma after the `=` or `|`. Seems more obvious that it applies to the whole constructor; I had thought from the example that it applied only to the first field (like an `UNPACK` pragma would). This also makes infix constructor syntax less strange. For records: an ambiguous field implies that the whole constructor is ambiguous; do we want to require pragmas on both the field and the constructor? Or do we want any pragma'd field to imply a pragma'd constructor?

AntC2:
Thank you @goldfirere for some sensible thinking. The `AMBIGUOUS` applies to the signature, the signature applies to the identifier (explicit or created fresh for an expression).

In the clear-cut cases the syntax goes `<identifier-or-expr> :: <type>`. It's not until we see the `::` that we know a type is coming. So I thought to infix the pragma between `::` and type.

More generally, there's other clues that a type is coming; there may be no `::`; the `::` might be in the middle of a type; there might be other `::` within the type. With data constrs my thought was they reduce to GADT style, and in clear-cut cases you get `<constr name> :: type`. But more generally you might get field decls before the `::`. Then 

> I'd like to propose that, for H98-syntax constructors, we put the pragma after the `=` or `|`. 

Yes. And why not also for GADT-style constructors, after the `where` or `|`. Thus:

    data D a where                         
        {-# AMBIGUOUS #-} MkD ::  (forall b. C a b => a)
      | {-# AMBIGUOUS #-} MkDb {bar :: Show a => Int} :: D a

    data D3 a  where                                                       -- not like this
      MkD3 {baz :: forall b. C a b => a} :: {-# AMBIGUOUS #-} D3 a         -- looks daft

> For records: an ambiguous field implies that the whole constructor is ambiguous; do we want to require pragmas on both the field and the constructor? Or do we want any pragma'd field to imply a pragma'd constructor?

Now that I'm less fixated on the `::`, I'm (mildly) inclined to say put the pragma on/before the constructor, rather than within each field. (The `::` in field decls has always seemed a bit bogus to me, because it's not giving the type of the identifier -- i.e. the field selector function.)

> 3. Ambiguity in type applications is questionable: ...

Does that mean you're reversing [your earlier comment](https://github.com/ghc-proposals/ghc-proposals/pull/232#issuecomment-508893381)? I'm happy to include removing the ambiguity check; or should that be a separate proposal? (Potentially it has more impact on existing code?)

goldfirere:
> And why not also for GADT-style constructors, after the `where` or `|`.

(Quibble: There is no `|` for GADT-syntax.) I still prefer these after the `::`. We have the `::`, so let's use it.

> ```
> data D3 a  where                                                       -- not like this
>   MkD3 {baz :: forall b. C a b => a} :: {-# AMBIGUOUS #-} D3 a         -- looks daft
> ```

Forgetting about the pragma, this is not the syntax for GADT record declarations. Try out a few examples, and then see where you think the pragma fits best.

And, yes, I'm reversing (mostly) my earlier comment about type applications. Today, we always do an ambiguity check on type applications; I'm proposing now never to do it. I honestly don't think that it deserves a separate proposal.

AntC2:
Thanks @goldfirere, I've fixed the dumbness wrt GADT syntax (in Appendix), and included an example ambiguous GADT decl.

Re potentially `AMBIGUOUS` type `@`pplications, I've added Section 2 point 7 for the Committee to assess; but also left in (for now) the syntax.

goldfirere:
Thanks, @AntC2 for the updates.

I'm happy with this now. Let's wait a few more days to see if anyone else in the committee has a comment. If not, I'll mark this as accepted.

aspiwack:
I don't understand point 7 in the specification:

> Type applications are to be allowed ambiguous signatures, no warning or error, without needing {-# AMBIGUOUS #-} (nor needing LANGUAGE AllowAmbiguousTypes).

Maybe this would benefit from a rewording.

Other than that, this all sounds reasonable to me. All things being equal, I prefer local switches over global switches.

AntC2:
Hi @aspiwack, there's an example `@`pplication left over at the end of point 1. That signature is ambiguous, and currently requires `LANGUAGE AllowAmbiguousTypes` to compile. (I would `<strike>` that example if only ReST allowed me.)

Point 7 is saying: with this proposal you won't need any switches set (not even `{ AMBIGUOUS }` pragma) for that code to compile.

aspiwack:
> I would `<strike>` that example if only ReST allowed me.

Why would you  `<strike>` that example?

Anyway, even in the light of the example, and your explanation, I don't understand point 7, I'm quite sorry about that. Is it about allowing `{-# AMBIGUOUS #-}` inside `@`-applications?

AntC2:
For any signature following type application `@`, if that signature is ambiguous (such as the example at the end of point 1):

* Current GHC behaviour is to reject the code; so
* to get the code to compile needs `LANGUAGE AllowAmbiguousTypes`.
* Proposed behaviour: accept the code without needing `AllowAmbiguousTypes`, nor `{-# AMBIGUOUS #-}`.

> Why would you `<strike>` that example?

Because a) it exhibits a now-superseeded version of the proposal; but b) a valid possible design to put `{-# AMBIGUOUS #-}`, if the Committee's view is that silently accepting ambiguous sigs after `@` is too dramatic of a change.

aspiwack:
@AntC2 If it's been superseded, then it ought to go in the _Alternatives_ section. Explain there what it is. Explain how it's superseded by the rest of the proposal. Explain why it could still be consider but you think it is unnecessary.

You can pretty much use the wording of your last comment.

I finally figured out what 7 meant, by the way. But it also needs to be explained clearly in the proposal. Something like

> `-XTypeApplications` is changed, so that when `-XTypeApplications` is enabled, `f @t` is valid regardless of whether `t` is ambiguous. <plus an example which is rejected now, but would be accepted after this change>

AntC2:
Done. See Changelog.

Another month, [another confused learner bamboozled by GHC's error message into unnecessary `AllowAmbiguousTypes`](https://mail.haskell.org/pipermail/haskell-cafe/2019-July/131330.html). With added non-injective Type Families red herring.

gridaphobe:
Sorry to be so late to the discussion. I also like the direction of this proposal, but have a couple questions about the specifics.

1. The proposal introduces `{-# AMBIGUOUS -#}` as a positional pragma, but in most cases it's applied to types that are being ascribed to some identifier. This makes it seem like a good candidate for a named pragma. The exceptions are expressions and type applications. Richard is suggesting we remove the ambiguity check from type applications entirely, and I haven't seen an argument for allowing ambiguous types in ascriptions to arbitrary expressions. Does anyone have an example where this is necessary?

2. The `-Wambiguous-types` warning seems unnecessary to me. It's main purpose appears to be to provide a migration path for code currently using `-XAllowAmbiguousTypes`, where the warning tells you which signatures must be annotated before you remove the global extension. But it would be easier to just remove the extension first, and then add the pragma wherever a type error appears!

AntC2:
> @gridaphobe 1. ... I haven't seen an argument for allowing ambiguous types in ascriptions to arbitrary expressions. 

The proposal is to touch everywhere that's currently touched by `AllowAmbiguousTypes`. Currently this is rejected unless you set that extension

    x = undefined :: (Num a, Ord b) => a

(To be precise, with `AllowAmbiguousTypes` set, that's rejected for other reasons ;-). And introducing `TypeApplications` to fix them seems to kick out some weird kind errors GHC 8.4.4.)

You might want to argue ambiguous expression ascriptions shouldn't be allowed. Take that up with the original author of the extension. That'd be a different proposal.

> 2. The -Wambiguous-types warning seems unnecessary to me. ...

[@simonpj gets it.](https://github.com/ghc-proposals/ghc-proposals/pull/232#issuecomment-508865791) You seem to be of the brutalist school of software evolution.

gridaphobe:
AllowAmbiguousTypes allows any user-written type in the module to be ambiguous. The whole point of this proposal is that that’s painting with far too large a brush. So I think this is a natural time for us to ask where desired ambiguity arises. 

The main instance I’m aware of is in class methods that are meant to be @pplied to the type rather than a value. It’s hard to imagine a case where I want an ambiguous type ascribed to an arbitrary expression. We’re already at the use-site, which means the ambiguity must be resolved immediately, so why not just resolve it in the ascribed type?

If we disallow ambiguity in ascriptions to expressions (and type applications) we could use named pragmas instead of positional pragmas, which would be a win for readability IMO. This is not a major issue, but it does seem like an opportunity to clean up a bit more than the current proposal does. 

Regarding the warning, I’m one of those people who likes to enable -Werror in CI, so Simon and my proposed migrations would be functionally equivalent for me, it would all have to happen in a single PR. Does that make me a brutalist? :)

gridaphobe:
But your point regarding the warning is well-taken, we should accommodate people who want a more gradual migration.

AntC2:
> @gridaphobe `AllowAmbiguousTypes` allows any user-written type in the module to be ambiguous. The whole point of this proposal is that that’s painting with far too large a brush. 

I think you've misunderstood the motivation. It'd be informative to read through the thread I link to in July's cafe. The major point of this proposal is that GHC suggests switching on `AllowAmbiguousTypes` when it's not needed (as in that case). With that switched on, that it exacerbates the situation by indeed painting with far too large a brush is a consequence, and not "the whole point".

> So I think this is a natural time for us to ask where desired ambiguity arises.

I've answered that: the ambiguity is _not_ desired. It's bogus, as a consequence of GHC's mis-leading message.

> The main instance I’m aware of is in class methods that are meant to be @pplied to the type rather than a value. ...

The cause I'm addressing is novice users who neither need nor understand ambiguous types. Their use case is not "meant to" anything. Their use case is because they're too trusting of GHC suggestions.

> If we disallow ambiguity in ascriptions to expressions ...

That would be a different proposal, that potentially breaks currently accepted code. By all means write it yourself. If the Committee judges it would be preferable to this proposal, I'll abandon this.

The people putting `{-# AMBIGUOUS #-}` pragmas are non-novice, who presumably know what an ambiguous signature is about, and know they're going to need type applications, and presumably also know that for expressions they might as well "resolve it in the ascribed type".

> (and type applications)

No this proposal (Point 7) is not disallowing ambiguity in type applications. To the contrary, it's _allowing_ ambiguity without any rejection/extension set/nor pragma. So it's the opposite way round to your counter-proposal for expression signatures.

gridaphobe:
> > @gridaphobe `AllowAmbiguousTypes` allows any user-written type in the module to be ambiguous. The whole point of this proposal is that that’s painting with far too large a brush.
> 
> I think you've misunderstood the motivation. It'd be informative to read through the thread I link to in July's cafe. The major point of this proposal is that GHC suggests switching on `AllowAmbiguousTypes` when it's not needed (as in that case). With that switched on, that it exacerbates the situation by indeed painting with far too large a brush is a consequence, and not "the whole point".

It seems to me that the motivation is to aid both novice and expert users. If the only issue were that GHC makes a bad suggestion, the proposal could be as simple as "stop suggesting that users enable `AllowAmbiguousTypes`". But this proposal takes the time (rightly!) to shrink the scope of the allowed ambiguity, which is presumably there to aid expert users who know that they want the ambiguous signature, but don't want other ambiguous signatures sneaking in accidentally. So I think my question about where we might reasonably want an ambiguous signature stands. 

> > (and type applications)
> 
> No this proposal (Point 7) is not disallowing ambiguity in type applications. To the contrary, it's _allowing_ ambiguity without any rejection/extension set/nor pragma. So it's the opposite way round to your counter-proposal for expression signatures.

You're quite right on that point, I mispoke.

AntC2:
> @gridaphobe  I think my question about where we might reasonably want an ambiguous signature stands.

I was re-stating my motivation as wanting to avoid/minimise the scope of the ambiguity. I'm not claiming to understand why someone wants ambiguity. I manage to write all my code without it, and without type applications. So in the absence of someone chipping in here on use cases, I see three scenarios:

1. Ascribing ambiguous types at an expression is dumb; and nobody's doing it. So this proposal has no effect.

2. Ascribing ambiguous types at an expression is dumb and a mis-feature; but somebody's doing it because you can currently. (Or they've switched on module wide `AllowAmbiguousTypes` and given an ambiguous expression signature unintentionally/because GHC won't tell them.) Then this proposal helps by identifying the location, and gets them to think whether it's needed. They can change the signature; then no need to be `{-# AMBIGUOUS #-}`

3. Ascribing ambiguous types at an expression has sensible use cases. Then this proposal allows marking as `{-# AMIBGUOUS #-}`. But removing the ability breaks the code.

What worries me about 'named pragms' -- where do I go to find out what that means exactly? Is that an existing feature or a pipe-dream? Is:

* Amongst the names with ambiguous signatures are datatype field names.
* Those are not uniquely identified now that we have `DuplicateRecordFields`.
* There's also the longer-standing `DisambiguateRecordFields` and `NamedFieldPuns`.

Those combined means it not necessarily easy to get from a field name to its definition to some pragma about the name.

Then how do we persuade users to move from the module-wide `AllowAmbiguousTypes` -- which, BTW, some people seem to advocate switching on globally -- to a complicated name-specific pragma? Putting a pragma slap against the signature that GHC is highlighting seems a modest burden.

Could I suggest the applicability of ambiguous expression signatures gets revisited as and when `AllowAmbiguousTypes` gets considered for deprecating (outside scope of this proposal)?

gridaphobe:
> **Addit:** if by 'named pragmas' you mean something along the lines of [`DEPRECATED` or `WARNING`](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#warning-and-deprecated-pragmas)

Yes, that's precisely what I meant, sorry for the confusion. Concretely, I would like to be able to write

```haskell
class Foo a where
  foo :: a -> a
  bar :: Int
  {-# AMBIGUOUS bar #-}
```

> note this from the Users Guide
> 
> > You can only attach to entities declared at top level in the module being compiled, and you can only use unqualified names in the list of entities. A capitalised name, such as `T` refers to _either_ the type constructor `T` or the data constructor `T`, or both if both are in scope.
> 
> Then consider you might have `AMBIGUOUS` signatures for non-top-level names; for a data constructor but not same-named type constructor; for one field name but not another same-named.

These are all places where `AllowAmbiguousTypes` would currently accept ambiguity, but I also find them unconvincing, or we have ways of dealing with them. It's not clear why you would have a local binder with an ambiguous signature, for the same reason as for arbitrary expressions. The use-site is local, so why not just make the signature unambiguous? There are also other named pragmas like `INLINE` that can apply to local binders, so there would be precedent for doing the same with `AMBIGUOUS` if necessary. For constructors there's an accepted [proposal] that enables disambiguating which constructor a pragma should apply to. For duplicate fields I don't think we have an existing way to disambiguate them, though I'm also perplexed as to why you would want ambiguity there.

[proposal]: https://github.com/ghc-proposals/ghc-proposals/blob/30eed9f8ea8d20f8401c6e80158402fcaae585d1/proposals/0008-type-infix.rst

I guess my takeaway from this is that it would be nice to have a better understanding of where Haskellers desire ambiguous types. The only case I'm aware of is typeclass methods like `bar` above (which feels like the Haskell analogue of a static method from object-oriented languages). What are other scenarios where ambiguity is useful? An answer would help us better target the extension/pragma to the desired usecases.

That being said, I don't mean to stand in the way of this proposal. I think it would be a marked improvement over the status quo. I just saw a potential opportunity to improve it further.

AntC2:
> @gridaphobe it would be nice to have a better understanding of where Haskellers desire ambiguous types.

I'm happy to say I have no understanding of where Haskellers desire ambiguous types. I have no such desire. And I observe (most) novices don't either. So I'm trying to avoid the harm GHC is currently doing. (Most of your first para has gone over my head. I'm not trying to redesign the ambiguity in signatures. Wouldn't those considerations have been apparent to the originator?)

>  I don't mean to stand in the way of this proposal.

Good, because so far it's seemed you want to make the perfect the enemy of the vastly better. I'm not going to hang around for other proposals that claim to disambiguate pragmas; but seem to require code beakages; and seem to have gone nowhere for nearly 2 years; and anyway haven't anticipated a generalised way to attach pragmas to names.

goldfirere:
The underlying question @gridaphobe is asking is (I believe -- sorry if I'm putting the wrong words into your mouth): Is a type ambiguous, or is an identifier ambiguous? I think this is a good question. If a type is ambiguous, then the syntax proposed here is the right one. If an identifier is ambiguous, then a named pragma might be better.

I'm in favor of calling a type ambiguous. The identifier can sometimes be a bit hard to pin down (examples: what do we do in an `InstanceSig`? do we have to label all fields of an ambiguous record constructor as ambiguous if the constructor type is?) and I think a correct program can include an ambiguous expression signature.

Given the long length of time this has been under consideration and the considerable silence-is-agreement among the committee, I'm going to make a deadline of Friday, at which point I will accept (and acknowledging @gridaphobe's statement that he wishes not to block acceptance).

/remind me Friday to accept this

proposals/AmbiguousType-pragma.rst
```diff
@@ -0,0 +1,242 @@
...
+   
+   See Appendix giving a diff to the BNF for the language (wrt the Report, and affected extensions).
+
+2. Signatures marked ``AMBIGUOUS`` are to be validated as if ``-XAllowAmbiguousTypes`` is set, for that signature only. (If that is already set module-wide, the pragma has the effect of suppressing the ``-Wambiguous-type`` warning, see 5.)
+
+3. This does not change the validation for ambiguous types/type variables at usage sites.
+
+4. The error reporting from the ambiguity check that currently suggests ``To defer the ambiguity check to use sites, enable AllowAmbiguousTypes`` must make clear this is likely to entail using ``TypeApplications`` at usage sites, and that there are several possible approaches to avoid ambiguous type variables.
+
+   Precise wording to be arrived at in discussion of this PR. (Prefer not mentioning ``AllowAmbiguousTypes`` at all.) Starting bikeshed::
```
aspiwack: This sentence should be removed before merge.

My attempt:

> If you intend to resolve the ambiguity at each use site (typically via TypeApplications), prefix this signature with the AMBIGUOUS pragma.

AntC2: Done. Thank you.

goldfirere:
I hereby accept. @nomeata, I believe you do the next steps?

nomeata:
Done, although you should/could notify the mailing list about the status change.

PR #233, opened 2019-05-23
Higher-order roles

goldfirere:
This proposal describes a way we can use existing features to describe the roles on the arguments to datatype parameters, thus permitting, e.g., `data T f a = MkT (f a)` to assign a representational role to both its parameters. It requires only a small tweak to role inference and is inspired by a problem "in the wild" from production code.

[Rendered](https://github.com/goldfirere/ghc-proposals/blob/higher-roles/proposals/0000-higher-roles.rst)

### DIFF
--- a/proposals/0000-higher-roles.rst
+++ b/proposals/0000-higher-roles.rst
@@ -48,9 +48,8 @@ Int``. This is sad, though, because we don't want to export the constructor of
 ``GList``, which is necessary in order to make those representational
 equalities. And it's also sad that we need a newtype here.
 
-This issue was directly inspired by a need in real software. (I'm not sure if
-the original reporter to me [in person] wishes to be identified. But that person
-is real and writes Haskell for a company you have heard of.)
+This issue was directly inspired by a need in real software, as reported by
+Lennart Augustsson of Google Research.
 
 Proposed Change Specification
 -----------------------------
--- a/proposals/0000-higher-roles.rst
+++ b/proposals/0000-higher-roles.rst
@@ -120,6 +120,12 @@ Alternatives
   Now, users can make their own decisions. This seems strictly better than either previous
   proposal.
 
+* An `alternative proposal <https://github.com/ghc-proposals/ghc-proposals/pull/248>`_
+  introduces an entirely new mechanism for ``Coercible``. At this point, that proposal
+  has yet to become convincingly sound, and it is *much* more invasive than this one.
+  It is worth reading as a counterpoint, but I don't think it's yet ready to be counted
+  as a viable.
+  
 Unresolved Questions
 --------------------
 None at this time.
--- a/proposals/0000-higher-roles.rst
+++ b/proposals/0000-higher-roles.rst
@@ -53,8 +53,8 @@ Lennart Augustsson of Google Research.
 
 Proposed Change Specification
 -----------------------------
-During role inference of a type with a datatype context, bring the datatype context
-into scope as a given.
+During role inference of a type with a datatype context, assume the constraints
+in the datatype context as givens.
 
 Then, when inferring the role of the argument of a type ``ty`` (which may be a type
 variable or some other type):
--- a/proposals/0000-higher-roles.rst
+++ b/proposals/0000-higher-roles.rst
@@ -17,6 +17,9 @@ If we write ``data T f a = MkT (f a)``, GHC assigns a nominal role to ``a``, bec
 as the argument of a type variable, and we cannot know how this argument will be used. This proposal
 describes a way to assign a representational role to ``a`` via the use of a quantified constraint.
 
+There are two proposed ways to fix this problem, each described in its own "Proposed Change
+Specification". Only **one** of these changes should be implemented.
+
 Motivation
 ----------
 Consider::
@@ -51,8 +54,8 @@ equalities. And it's also sad that we need a newtype here.
 This issue was directly inspired by a need in real software, as reported by
 Lennart Augustsson of Google Research.
 
-Proposed Change Specification
------------------------------
+Proposed Change Specification (1)
+---------------------------------
 During role inference of a type with a datatype context, assume the constraints
 in the datatype context as givens.
 
@@ -69,22 +72,49 @@ variable or some other type):
 3. Otherwise, mark the argument of ``ty`` as having a nominal role.
 
 Lastly, GHC will not warn when ``-XDatatypeContexts`` is specified.
-   
+
+Proposed Change Specification (2)
+---------------------------------
+When role inference inspects the type of a data constructor, it will assume the
+data constructor context as a given.
+
+Then, when inferring the role of the argument of a type ``ty`` (which may be a type
+variable or some other type):
+
+1. Try to solve ``forall a b. Coercible (ty a) (ty b)``. If solving succeeds, then
+   add no constraints on the role of the argument of ``ty``.
+
+2. Try to solve ``forall a b. Coercible a b => Coercible (ty a) (ty b)``. If solving
+   succeeds, then mark the argument of ``ty`` as having at least a representational
+   role.
+
+3. Otherwise, mark the argument of ``ty`` as having a nominal role.
+
 Effect and Interactions
 -----------------------
-1. We can now write ::
+1. We can now write (with change (1)) ::
 
      data (forall c d. Coercible c d => Coercible (v c) (v d)) => GList v a = Mk Metadata (v a)
      type role GList representational representational
 
-2. You might be worried (I was) that the fact that datatype contexts are broken would
+   or (with change (2)) ::
+
+     data GList v a where
+       Mk :: (forall c d. Coercible c d => Coercible (v c) (v d)) => Metadata -> v a -> GList v a
+     type role GList representational representational
+
+2. (1) You might be worried (I was) that the fact that datatype contexts are broken would
    make this unsound. That is, even if we have ``G :: Type -> Type`` with ``type role G nominal``, one
    can still talk about the type ``GList G Int``. However, we would never have a value
    of that type, and so no threat to soundness can come of the type's existence.
 
+3. (2) It would appear that the ``Mk`` constructor in this example would now have to hold
+   evidence for the implication constraint. However, it seems quite possible to detect this
+   shape of dictionary and eliminate it.
+   
 Costs and Drawbacks
 -------------------
-1. This uses ``-XDatatypeContexts``, which many people don't like. But it's exactly what
+1. (1) This uses ``-XDatatypeContexts``, which many people don't like. But it's exactly what
    we need here.
 
 2. This is potentially non-performant, requiring running the solver a lot during role inference.
@@ -99,6 +129,9 @@ Costs and Drawbacks
    abstraction barrier might be compromised. There is also a remote chance that I will quantum-tunnel
    through the floor and break my leg on the floor below. Neither of these is likely.
 
+5. (2) Newtypes will not be able to take advantage of this new facility, as newtype
+   constructors cannot have contexts.
+
 Alternatives
 ------------
 * Instead of using a quantified constraint and invoking the solver, we could come up with a more
@@ -106,27 +139,18 @@ Alternatives
   This would mean we wouldn't run the solver during role inference. However, such a feature seems
   sadly non-orthogonal when we have the features to express this idea already.
 
-* We could configure the warning about datatype contexts in some other way.
+* (1) We could configure the warning about datatype contexts in some other way.
 
-* Instead of using datatype contexts at all, we could use data constructor contexts,
-  which avoid the controversy of datatype contexts. Role inference necessarily looks at
-  the usages of type variables in each constructor. When doing so, we could assume the
-  constraints on the data constructor as givens, and then do the checks as described
-  in the Specification, above. This means that a datatype with many constructors would
-  have to repeat the context many times. It also means that, by default, an unnecessary
-  dictionary would be stored by the constructor; however, I think it would not be hard
-  to optimize that dictionary away completely.
+* I have orchestrated an unofficial vote on the two alternatives `on the GitHub trail <https://github.com/ghc-proposals/ghc-proposals/pull/233#issuecomment-498672441>`_.
 
-  One problem with this alternative is that it means that newtypes cannot take advantage
-  of these higher-order roles. Given that the constructor of ``Linked`` in the Motivation
-  would be hidden in clients, it means that this alternative would not actually permit
-  the motivating program to be accepted.
-
-  I have orchestrated an unofficial vote on this alternative `on the GitHub trail <https://github.com/ghc-proposals/ghc-proposals/pull/233#issuecomment-498672441>`_.
+* Considering (1) vs (2), both have the potential for repetition: In (1), the datatype
+  context may have to be written in many type signatures and instance heads, etc. In (2),
+  the context may have to be written in many constructors. In the end, (1)'s repetition
+  is unbounded, while (2) is more manageable.
 
 Unresolved Questions
 --------------------
-None at this time.
+\(1) or (2)?
 
 Implementation Plan
 -------------------
--- a/proposals/0000-higher-roles.rst
+++ b/proposals/0000-higher-roles.rst
@@ -1,0 +1,120 @@
+Simulate higher-order roles via cleverer role inference
+=======================================================
+
+.. proposal-number:: Leave blank. This will be filled in when the proposal is
+                     accepted.
+.. ticket-url:: Leave blank. This will eventually be filled with the
+                ticket URL which will track the progress of the
+                implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+If we write ``data T f a = MkT (f a)``, GHC assigns a nominal role to ``a``, because it appears
+as the argument of a type variable, and we cannot know how this argument will be used. This proposal
+describes a way to assign a representational role to ``a`` via the use of a quantified constraint.
+
+Motivation
+----------
+Consider::
+
+  -- in a library
+  data GList :: (Type -> Type)   -- concrete representation
+             -> Type             -- element type
+             -> Type where
+    Mk :: Metadata      -- some metadata; details irrelevant
+       -> v a           -- contents
+       -> GList v a
+
+  newtype Linked a = MkLinked (GList [] a)
+
+  -- in a client
+  newtype Age = MkAge Int
+  change :: Linked Age -> Linked Int
+  change = coerce
+
+It would be nice if this could work. The problem is that we have ``type role
+GList representational nominal``, with the ``a :: Type`` parameter getting a
+nominal role. This happens because we can't know the role of ``v``\'s argument,
+and so GHC (correctly) conservatively chooses nominal.
+
+The only solution I could offer is to make ``GList`` a newtype (holding a tuple)
+so that GHC could look through the newtype to its representation and then
+discover that ``Linked Age ~R (Metadata, [Age]) ~R (Metadata, [Int]) ~R Linked
+Int``. This is sad, though, because we don't want to export the constructor of
+``GList``, which is necessary in order to make those representational
+equalities. And it's also sad that we need a newtype here.
+
+This issue was directly inspired by a need in real software. (I'm not sure if
+the original reporter to me [in person] wishes to be identified. But that person
+is real and writes Haskell for a company you have heard of.)
+
+Proposed Change Specification
+-----------------------------
+During role inference of a type with a datatype context, bring the datatype context
+into scope as a given.
+
+Then, when inferring the role of the argument of a type ``ty`` (which may be a type
+variable or some other type):
+
+1. Try to solve ``forall a b. Coercible (ty a) (ty b)``. If solving succeeds, then
+   add no constraints on the role of the argument of ``ty``.
+
+2. Try to solve ``forall a b. Coercible a b => Coercible (ty a) (ty b)``. If solving
+   succeeds, then mark the argument of ``ty`` as having at least a representational
+   role.
+
+3. Otherwise, mark the argument of ``ty`` as having a nominal role.
+
+Lastly, GHC will not warn when ``-XDatatypeContexts`` is specified.
+   
+Effect and Interactions
+-----------------------
+1. We can now write ::
+
+     data (forall c d. Coercible c d => Coercible (v c) (v d)) => GList v a = Mk Metadata (v a)
+     type role GList representational representational
+
+2. You might be worried (I was) that the fact that datatype contexts are broken would
+   make this unsound. That is, even if we have ``G :: Type -> Type`` with ``type role G nominal``, one
+   can still talk about the type ``GList G Int``. However, we would never have a value
+   of that type, and so no threat to soundness can come of the type's existence.
+
+Costs and Drawbacks
+-------------------
+1. This uses ``-XDatatypeContexts``, which many people don't like. But it's exactly what
+   we need here.
+
+2. This is potentially non-performant, requiring running the solver a lot during role inference.
+   However, it will happen only on types that have a datatype context. If we're worried about
+   performance, we could require users to write a role annotation to confirm that they want GHC
+   to do the extra work.
+
+3. It has not been proved sound. Even so, I would be comfortable adding this to GHC without a proof.
+   There are already 4 publications about roles!
+
+4. There is a remote chance that this would relax some roles in existing code, meaning that a library-writer's
+   abstraction barrier might be compromised. There is also a remote chance that I will quantum-tunnel
+   through the floor and break my leg on the floor below. Neither of these is likely.
+
+Alternatives
+------------
+* Instead of using a quantified constraint and invoking the solver, we could come up with a more
+  restrictive syntax for this idea, which expands out to a quantified constraint on every constructor.
+  This would mean we wouldn't run the solver during role inference. However, such a feature seems
+  sadly non-orthogonal when we have the features to express this idea already.
+
+* We could configure the warning about datatype contexts in some other way.
+
+Unresolved Questions
+--------------------
+None at this time.
+
+Implementation Plan
+-------------------
+I suppose I could implement.
--- a/proposals/0000-higher-roles.rst
+++ b/proposals/0000-higher-roles.rst
@@ -108,6 +108,22 @@ Alternatives
 
 * We could configure the warning about datatype contexts in some other way.
 
+* Instead of using datatype contexts at all, we could use data constructor contexts,
+  which avoid the controversy of datatype contexts. Role inference necessarily looks at
+  the usages of type variables in each constructor. When doing so, we could assume the
+  constraints on the data constructor as givens, and then do the checks as described
+  in the Specification, above. This means that a datatype with many constructors would
+  have to repeat the context many times. It also means that, by default, an unnecessary
+  dictionary would be stored by the constructor; however, I think it would not be hard
+  to optimize that dictionary away completely.
+
+  One problem with this alternative is that it means that newtypes cannot take advantage
+  of these higher-order roles. Given that the constructor of ``Linked`` in the Motivation
+  would be hidden in clients, it means that this alternative would not actually permit
+  the motivating program to be accepted.
+
+  I have orchestrated an unofficial vote on this alternative `on the GitHub trail <https://github.com/ghc-proposals/ghc-proposals/pull/233#issuecomment-498672441>`_.
+
 Unresolved Questions
 --------------------
 None at this time.
--- a/proposals/0000-higher-roles.rst
+++ b/proposals/0000-higher-roles.rst
@@ -9,9 +9,7 @@ Simulate higher-order roles via cleverer role inference
 .. implemented:: Leave blank. This will be filled in with the first GHC version which
                  implements the described feature.
 .. highlight:: haskell
-.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
-            **After creating the pull request, edit this file again, update the
-            number in the link, and delete this bold sentence.**
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/233>`_.
 .. sectnum::
 .. contents::
 
--- a/proposals/0000-higher-roles.rst
+++ b/proposals/0000-higher-roles.rst
@@ -17,9 +17,6 @@ If we write ``data T f a = MkT (f a)``, GHC assigns a nominal role to ``a``, bec
 as the argument of a type variable, and we cannot know how this argument will be used. This proposal
 describes a way to assign a representational role to ``a`` via the use of a quantified constraint.
 
-There are two proposed ways to fix this problem, each described in its own "Proposed Change
-Specification". Only **one** of these changes should be implemented.
-
 Motivation
 ----------
 Consider::
@@ -115,9 +112,9 @@ Alternatives
   This would mean we wouldn't run the solver during role inference. However, such a feature seems
   sadly non-orthogonal when we have the features to express this idea already.
 
-* Previous versions of this proposal had the constraints either in the datatype context or
-  in the constructor contexts, but that decision was made at the level of the proposal.
-  Now, users can make their own decisions. This seems strictly better than either previous
+* Previous versions of this proposal had the constraints either in the datatype context (only) or
+  in the constructor contexts (only).
+  In this version of the proposal, allowing constraints in both places, users can make their own decisions. This seems strictly better than either previous
   proposal.
 
 * An `alternative proposal <https://github.com/ghc-proposals/ghc-proposals/pull/248>`_
--- a/proposals/0000-higher-roles.rst
+++ b/proposals/0000-higher-roles.rst
@@ -54,29 +54,10 @@ equalities. And it's also sad that we need a newtype here.
 This issue was directly inspired by a need in real software, as reported by
 Lennart Augustsson of Google Research.
 
-Proposed Change Specification (1)
----------------------------------
-During role inference of a type with a datatype context, assume the constraints
-in the datatype context as givens.
-
-Then, when inferring the role of the argument of a type ``ty`` (which may be a type
-variable or some other type):
-
-1. Try to solve ``forall a b. Coercible (ty a) (ty b)``. If solving succeeds, then
-   add no constraints on the role of the argument of ``ty``.
-
-2. Try to solve ``forall a b. Coercible a b => Coercible (ty a) (ty b)``. If solving
-   succeeds, then mark the argument of ``ty`` as having at least a representational
-   role.
-
-3. Otherwise, mark the argument of ``ty`` as having a nominal role.
-
-Lastly, GHC will not warn when ``-XDatatypeContexts`` is specified.
-
-Proposed Change Specification (2)
----------------------------------
-When role inference inspects the type of a data constructor, it will assume the
-data constructor context as a given.
+Proposed Change Specification
+-----------------------------
+When role inference inspects the type of a data constructor, it will assume both the
+data constructor and datatype context as a givens.
 
 Then, when inferring the role of the argument of a type ``ty`` (which may be a type
 variable or some other type):
@@ -92,44 +73,39 @@ variable or some other type):
 
 Effect and Interactions
 -----------------------
-1. We can now write (with change (1)) ::
-
-     data (forall c d. Coercible c d => Coercible (v c) (v d)) => GList v a = Mk Metadata (v a)
-     type role GList representational representational
-
-   or (with change (2)) ::
+1. We can now write::
 
      data GList v a where
        Mk :: (forall c d. Coercible c d => Coercible (v c) (v d)) => Metadata -> v a -> GList v a
      type role GList representational representational
 
-2. (1) You might be worried (I was) that the fact that datatype contexts are broken would
+2. You might be worried (I was) that the fact that datatype contexts are broken would
    make this unsound. That is, even if we have ``G :: Type -> Type`` with ``type role G nominal``, one
    can still talk about the type ``GList G Int``. However, we would never have a value
    of that type, and so no threat to soundness can come of the type's existence.
 
-3. (2) It would appear that the ``Mk`` constructor in this example would now have to hold
+3. It would appear that the ``Mk`` constructor in this example would now have to hold
    evidence for the implication constraint. However, it seems quite possible to detect this
    shape of dictionary and eliminate it.
+
+4. Unlike previous versions of this proposal, this version does *not* change the status
+   of ``-XDatatypeContexts``.
    
 Costs and Drawbacks
 -------------------
-1. (1) This uses ``-XDatatypeContexts``, which many people don't like. But it's exactly what
-   we need here.
-
-2. This is potentially non-performant, requiring running the solver a lot during role inference.
+1. This is potentially non-performant, requiring running the solver a lot during role inference.
    However, it will happen only on types that have a datatype context. If we're worried about
    performance, we could require users to write a role annotation to confirm that they want GHC
    to do the extra work.
 
-3. It has not been proved sound. Even so, I would be comfortable adding this to GHC without a proof.
+2. It has not been proved sound. Even so, I would be comfortable adding this to GHC without a proof.
    There are already 4 publications about roles!
 
-4. There is a remote chance that this would relax some roles in existing code, meaning that a library-writer's
+3. There is a remote chance that this would relax some roles in existing code, meaning that a library-writer's
    abstraction barrier might be compromised. There is also a remote chance that I will quantum-tunnel
    through the floor and break my leg on the floor below. Neither of these is likely.
 
-5. (2) Newtypes will not be able to take advantage of this new facility, as newtype
+4. Newtypes will have to use datatype contexts, as newtype
    constructors cannot have contexts.
 
 Alternatives
@@ -139,18 +115,14 @@ Alternatives
   This would mean we wouldn't run the solver during role inference. However, such a feature seems
   sadly non-orthogonal when we have the features to express this idea already.
 
-* (1) We could configure the warning about datatype contexts in some other way.
-
-* I have orchestrated an unofficial vote on the two alternatives `on the GitHub trail <https://github.com/ghc-proposals/ghc-proposals/pull/233#issuecomment-498672441>`_.
-
-* Considering (1) vs (2), both have the potential for repetition: In (1), the datatype
-  context may have to be written in many type signatures and instance heads, etc. In (2),
-  the context may have to be written in many constructors. In the end, (1)'s repetition
-  is unbounded, while (2) is more manageable.
+* Previous versions of this proposal had the constraints either in the datatype context or
+  in the constructor contexts, but that decision was made at the level of the proposal.
+  Now, users can make their own decisions. This seems strictly better than either previous
+  proposal.
 
 Unresolved Questions
 --------------------
-\(1) or (2)?
+None at this time.
 
 Implementation Plan
 -------------------


AntC2:
> This uses `-XDatatypeContexts`, which many people don't like

Thanks @goldfirere . It's not so much "don't like" as 'semantics aren't useful'/stupid theta. So it's in process of being deprecated #161 . Then we can bring it back with sensible semantics. (Supposing we can specify what that looks like -- opinions vary. Mostly I've seen it suggested as some sort of shorthand for a GADT with constraint, and/or the fancy dual constraints for pattern synonyms.)

So are you saying `DatatypeContexts` is not stupid? What semantics do you expect for this higher-order roles business? Is that consistent with other expectations/can we express it in System FC?

AntC2:
> some sort of shorthand for a GADT with constraint

Ah, I discover you can't mix `DatatypeContexts` with GADT-style syntax, even if you're declaring a constructor that's regular H98. So you want that context on the type `GList` rather than on the constructor `Mk`.

AFAICT, the most-favoured non-stupid semantics for the rise-from-the-ashes `DatatypeContexts` is as a GADT-style constraint prefixed to every data constructor. Then how does your proposed `GList`/`Mk` differ from (this accepted today)

    data GListG v a where
         MkG :: (forall c d. Coercible c d => Coercible (v c) (v d))
                => Metadata -> (v a) -> GListG v a

    type role GListG representational nominal          -- we want both representational here

What if this proposal ran: to accept the wanted `type role GList ...` every data constructor within `GList` must have that Quantified/`Coercible` constraint? (Presumably we need GHC to recognise `Coercible` as special/magic.)

I have a lot of use cases where I'd like to put `~` in an Implication constraint. To be precise, GHC accepts that but it doesn't achieve anything. How would this be different to `Coercible`?:

    data (forall c d. c ~ d => (v c) ~ (v d))
     => GListEq v a = MkEq Metadata (v a)

RyanGlScott:
Would these quantified `Coercible` constraints only have the semantics you described if used in the same position that is currently occupied by `DatatypeContexts`? I'm imagining someone writing the following when [top-level kind signatures](https://github.com/ghc-proposals/ghc-proposals/pull/54) become a thing:

```hs
data GList :: forall (v :: Type -> Type)
           -> (forall c d. Coercible c d => Coercible (v c) (v d))
           => Type -> Type
data GList v a = Mk Metadata (v a)
type role GList representational representational
```

Would this mean the same thing as this version of `GList`?

```hs
data (forall c d. Coercible c d => Coercible (v c) (v d)) => GList v a = Mk Metadata (v a)
```

Or would that mean something different?

RyanGlScott:
Moreover, I'm not sure that using the word "`representational`" is really the best way to describe the role of `GList`'s first argument. If I understand the proposal correctly, it's more like "`representational`, but only if certain conditions are met". Perhaps we should consider printing `representational*` or somesuch to make it more obvious that it is in fact a higher-order role, not an ordinary `representational` role.

simonpj:
Alas, I have no idea what is going on in this proposal.

I can't build a connection in my head (of any kind) between section 1 (motivation) and section 2 (specification).  Nor do I understand what role the data type context plays.

We really need more.  And examples of the proposal in use.  Im really lost!

(Plus I'm deeply suspicous of anything involving datatype contexts, a mis-feature IMHO.)

Ericson2314:
I wish we could just get rid of role annotations and some sort of `where` clause for all the `Coercible` axioms the user wants.

Icelandjack:
Does this tie in with [Partial Type Constructors](https://cs.brynmawr.edu/~rae/papers/2019/partialdata/partialdata.pdf)?

simonpj:
> I can't build a connection in my head (of any kind) between section 1 (motivation) and section 2 (specification). Nor do I understand what role the data type context plays.

Richard asked me to clarify, but I'm so lost it's hard.  Let's start with

> During role inference of a type with a datatype context, bring the datatype context into scope as a given.

I have no idea what it means to "bring the datatype context into scope during role inference"..

Probably an example would help.   Why not use the example from section 1.  GList doesn't _have_ a data type context.  Perhaps you intend that it should get one? But what?  `Eq a`?  Unlikely.  Something else perhaps?

> Then, when inferring the role of the argument of a type ty (which may be a type variable or some other type):

I think we'll need a reminder of how we "infer the role of an argument of a type".  Do we do it one argument at a time?   And in `data T a b = ...` aren't the arguments `a` and `b` always type variables?

goldfirere:
I've tweaked the wording at the beginning of the specification to be clearer: "scope" was the wrong word to use there. I want to assume the datatype context as a given.

The connection between the motivation and the specification is the "Effect and Interactions", which gives the example you seek.

Role inference is described in Section 4.6 of the [Coercible paper](https://repository.brynmawr.edu/cgi/viewcontent.cgi?referer=&httpsredir=1&article=1010&context=compsci_pubs). Briefly, it assumes all the variables are phantoms, looking for information to contradict that assumption (like a usage of the variable), working toward a fixpoint. This proposal suggests that, instead of just assuming that arguments to type variables are nominal, to use the constraint solver instead.

Does this help? My guess is that the bit in "Effect and Interactions" will make it all much clearer.

simonpj:
I'm getting a faint and distant glimmer of what you are doing now, thanks.

But in the proposal you give a proposed change, but you give no reason to suppose it is sound.  I'm not asking for a _proof_ (though I assume you'd do one before we adopted this) but I would like an _argument_!

Guessing wildly, perhaps your argument goes like this:

* The type of the data constructor `Mk` is
```
  Mk :: (forall c d. Coercible c d => Coercible (v c) (v d)) => v a -> GList v a
```
* So any value built with `Mk` must have used a type for `v` that satisfied that quantified constraint..
* In this case that limits `v` in just the way we want.

What if we had written
```
data GList v a where
  Mk :: (forall c d. Coercible c d => Coercible (v c) (v d)) => v a -> GList v a
```
Should that not have the same property?   Should you not take into account the context of the data constructor?

And if so, you don't really need that datatype context after all except possibly as an abbreviation.  And we can abbreviate that big context with
```
type Representational v = forall c d. Coercible c d => Coercible (v c) (v d)
```
I'd like this proposal much more if it didn't mess around with datatype contexts.

Incidentally H2010 specifies that datatype contexts become Wanted constraints during pattern matching, as well as a use of the contructor.  I'm sure you don't particularly want that!

goldfirere:
We could do this via GADT-style constructor contexts. But to do it soundly (and, yes, your argument for soundness is exactly what I was thinking), you would have to make sure that the constraint is on every constructor. So role inference would have to somehow take the intersection of all the contexts. This could be done by checking the implication for each context, in turn. Yet putting the constraint in the datatype context more directly expresses what we mean here: it's not a constraint on the data constructor but a constraint on the type itself.

As for generating Wanteds on pattern-matches: I suppose I don't need that behavior, but it certainly doesn't hurt anything. The check helps us to avoid unreachable code (which could otherwise come about by writing a function that pattern-matches on an impossible type -- without the Wanted, GHC would never check).

You'll find that I don't share your dislike of datatype contexts: see this [ICFP reject / expected POPL submission](https://cs.brynmawr.edu/~rae/papers/2019/partialdata/partialdata.pdf). Don't take that paper too seriously -- I really don't know if it's such a hot idea in reality -- but it is an interesting (and sound) thought experiment.

Datatype contexts are also something that we'll naturally want if we adopt #177. #177 can work without them, but it's much better with them.

simonpj:
> somehow take the intersection of all the contexts

I don't think so. Each constructor already independently places constraints on the role of the arguments of the type constructor; we already ensure that all those constraints are met -- that's the intersection.

> As for generating Wanteds on pattern-matches: I suppose I don't need that behavior, but it certainly doesn't hurt anything

It does though, doesn't it?  It places extra constraints -- and one that we never infer -- in the context of every pattern match on any data constructor of the type.  And that dictionary is never used, which feels all wrong.

Now I think about it, those data consturctors will also give rise to constraints we can't infer. So
```
f x = (GList x, Glist x)
```
will be rejected.

I'm not fundamentally against data type contexts.  But I am troubled by the idea of computing dictionaries that we never use.   In System FC we very carefully pass evidence around, even when it is evenutally erased just before runtime.  That stops us making unsound program transformations, as we point out in the original paper, and at the time I thought that was totally brilliant. A quick example
```
data T a where { T1 :: T Bool; T2 :: T a }
```
f :: T a -> a -> Int
f x y = case x of
             T1 -> let v = not y in blah1
            T2 -> blah2
```
It looks as if we could float that let:
```
f :: T a -> a -> Int
f x y =  let v = not y in 
           case x of
             T1 ->blah1
            T2 -> blah2
```
but that's ill-typed.  When w realise that T1 binds a coercion that we use to cast y, all is well.

I can't give an example, but I worry that anything that relies on a proof that we immediately discard will be vulneralbe to unsound program transformations.

**Comment by goldfirere** [comment]
Your point about the fact that we check data constructors separately *already* is valid. I had been thinking of roles as a property of a type, but it's really some kind of least upper bound on a property of constructors. Moving the constraint to the constructors would mean repeating it a lot -- and storing a dictionary that will never be used. (I don't think it would be hard to convince the optimizer to blot it out, though. The dictionaries will always be vacuous, as all `Coercible` dictionaries are.)

What do others think? I'll make a separate comment to enable voting.

**Comment by goldfirere** [comment]
Give a "thumbs up" to this comment if you prefer the coercible **constraints on constructors**.

Give a "thumbs down" to this comment if you prefer the coercible **constraints in datatype contexts**.

And, if you like, give a "confused" to this comment if you're intensely relaxed about this choice.

**Comment by RyanGlScott** [comment]
I think I prefer constraints in datatype contexts, but I'm hesitant to formally cast my vote because I still don't quite understand how it would work. Having an answer to my question in https://github.com/ghc-proposals/ghc-proposals/pull/233#issuecomment-495730654 would be helpful.

**Comment by goldfirere** [comment]
Ryan, thanks for reminding me about your https://github.com/ghc-proposals/ghc-proposals/pull/233#issuecomment-495730654. A constraint in a datatype kind is very much like a datatype context. Interestingly, in order to have the regularity property, we'll need the wanted constraints on pattern-matches just like we do for datatype contexts.

**Property (Regularity).** If an expression `e` is well-typed at `t`, then `t` is well-kinded at `TYPE r` for some `r`.

**Comment by Ericson2314** [comment]
@goldfirere indeed that correspondence is the one thing that got me about https://cs.brynmawr.edu/~rae/papers/2019/partialdata/partialdata.pdf, as that makes type constructors with data contexts different than "regular" generative functions returning types.

**Comment by AntC2** [comment]
As @RyanGlScott says, we don't currently have a semantics for datatype contexts. They were removed from the H2010 Language Report -- but after it was published. strangely. They're on a schedule for deprecation in GHC. So the official position AFAICT is that they're not a thing.

So I put a thumbs-up to @goldfirere's request -- which was really a vote _against_ a non-thing rather than a vote _for_ constraints on constructors.

* Constraints on constructors (i.e. GADTs) we know will work; although it'll be tedious repetition.

* I think it's putting the cart before the horse to advocate for datatype contexts: I'm all for bringing back that syntax and giving it sensible semantics. (Different to the pre-H2010 semantics.) Perhaps the favoured semantics will amount to 'treat as if a constraint repeated on each constructor'. Perhaps it'll be more like the dual constraints on pattern synonyms.

But we need to see a proposal/argument for a sensible-theta rather than getting the semantics painted into a corner by this Higher-order roles proposal.

**Comment by AntC2** [comment]
I'm sold by the 'Partial Data Constructors' paper. I'm particularly struck by the Historical note in Section 7 _Datatype contexts in Haskell_: original Haskell Report 1.0 tried to define (informally) a semantics for datatype contexts (and for contexts on type synonyms). That answers a question I raised a long time ago.

It does make sense that (for example) on a BST implementing a set, or a monad transformer, we want a constraint on the underlying datatype to make it eligible for the class; rather than repeating constraints in the class and all derived classes/methods. Then a type function yielding such a constraint is also unnecessarily indirect. 

That semantics got taken out in report v1.1, and replaced by 'stupid theta'. I guess in 1990, without System FC and a mature/formal semantics of evidence-passing in qualified types, nobody knew how to specify the behaviour.

I'm intrigued by the paper's hints that a datatype context might intersect with the desired semantics for _Constrained type families_.

**Comment by goldfirere** [comment]
I've updated the proposal with the "in data constructors" alternative.

**Comment by AntC2** [comment]
As a counterpoint to the "in data constructors" alternative, consider a variation on [@simonpj's example](https://github.com/ghc-proposals/ghc-proposals/pull/233#issuecomment-498561878)

    data T3 a where
      T3 :: Integral a => a -> T3 a            -- Integral implies Eq
      T4 :: Eq a => a -> T3 a 

    f3 :: T3 a -> a -> Int
    f3 x y = case x of
      (T3 x') -> let v3 = x' == y in blah3
      (T4 x') -> let v4 = x' == y in blah4

    g3 :: Eq a => T3 a -> a -> Int
    g3 = f3

With GADTs only, does/could GHC see it is well-typed to float out `v3` and `v4`? Does it see it can then eliminate the common subexpression?

With `g3` does/could it recognise `T3 a` already carries an `Eq` dictionary, so that could be shared, even though there's no explicit `Eq` in `f3`'s signature or in `T3`'s decl?

IOW _if_ the risen-from-the-ashes `DatatypeContexts` is no more than shorthand for repeating the constraint on every data constructor (and thereby forcing them to be GADTs), does that throw away opportunities for type refinement and optimisations?

**Comment by goldfirere** [comment]
Good point, @AntC2. `f3` couldn't allow CSE, exactly as you surmise. There's no way that would be (provably) sound.... and there's no dictionary to use without pattern-matching (though that's an implementation concern).

`g3` could conceivably use CSE. I don't know if GHC is clever enough to do so, as it will have the choice of two `Eq a` dictionaries at the occurrence of `==`. However, I'm confident we could arrange to fix this if provoked.

So: I agree that datatype contexts are superior to repeated GADT constraints in this way.

**Comment by AntC2** [comment]
> @goldfirere datatype contexts are superior to repeated GADT constraints

Datatype contexts might give `representational` roles. Does that give datatypes that are usable? Consider

    data (forall c d. Coercible c d => Coercible (v c) (v d)) => GList v a = Mk Metadata (v a)
    type role GList representational representational

    instance (forall c d. Coercible c d => Coercible (v c) (v d)) => Functor (GList v)  where
      fmap f ( Mk m xs ) = ...

    newtype LinkedS a = MkLinkedS (GList Set a)        -- operations on Set usually need Ord a

    instance ( ?? ) => Functor LinkedS  where
      fmap f (MkLinkedS (Mk m xs)) = ...               -- Ord a in here somehow?

Usually with datatype contexts, the context must be repeated everywhere the type constructor appears -- so in the `instance ... => Functor`. Or are you saying we can assume them away?

Ideally we'd want `Set` to have a datatype context `Ord a` (or in general, the container type imposes constraints on the contents), but that makes it impossible to declare instances for many common classes. Then getting the contents `representational` might be the least of the worries.

**Comment by Ericson2314** [comment]
@AntC2 In the paper, the constraints automatically propagate from the data context to constraint the type arguments (and thus eventually any variables used in the arguments).

**Comment by AntC2** [comment]
> @Ericson2314 In the paper,  ...

Yes, it was reading the paper prompted my musings. @goldfirere is not proposing the whole of the paper's mechanisms here (or am I very confused?). Then the programmer would have to apply the propogating by hand, as currently with datatype contexts(?) And you can't always do that -- for example not constraining the element type for a constructor class(?) is what I'm asking.

**Comment by goldfirere** [comment]
I am *not* proposing the entirety of that paper here.

And you're right that the datatype-context approach means a lot of tiresome repetition of constraints. Datatype contexts might, in this way, induce *more* repetition than constructor constraints. At least the number of repetitions with constructor constraints is bounded.

For the concrete example in https://github.com/ghc-proposals/ghc-proposals/pull/233#issuecomment-501688208, you wouldn't be able to write that final `Functor` instance, because `Set`'s parameter is not representational. Indeed, any type that depends on a type class for operations the way `Set` does will have (if written correctly) a nominal parameter. So this problem goes away.

However, in thinking about the repetition engendered by doing this with datatype contexts, I wonder if maybe it would be silly to use datatype contexts without simultaneously (or previously) proposing that [paper](https://cs.brynmawr.edu/~rae/papers/2019/partialdata/partialdata.pdf).

If and when datatype contexts ever become not-stupid, it will be sensible to assume the datatype context as a given when examining constructor types. Thus, we can set up this proposal to be forward-compatible with such a change. It would require constructor contexts today, but work with non-stupid datatype contexts should they ever arise.

**Comment by chshersh** [comment]
I'm very interested in seeing this proposal implemented though it looks too technical for me so I can't comment on the best way how to implement it. But I can provide a real-life use case for this feature need.

I'm working on a [bidirectional TOML serialisation](https://kowainik.github.io/posts/2019-01-14-tomland) and the core of the [`tomland`](https://github.com/kowainik/tomland) library is the following  data type:

```haskell
data Codec r w c a = Codec
    { codecRead  :: r a
    , codecWrite :: c -> w a
    }
```

I would like to improve the library by allowing users to have `Generic` codecs for their data types. Moreover, I want to provide nice UX with `DerivingVia` so developers can have generic TOML codecs with as less boilerplate as possible. However, due to the current limitation of the role system I can't use `DerivingVia` to derive my typeclasses that return `Codec` as an argument because `DerivingVia` uses `coerce` underhood and GHC tells me that it cannot `coerce` values of type parameter `a` for `Codec`.

> I've [asked on Reddit](https://www.reddit.com/r/haskellquestions/comments/c602b2/cannot_coerce_data_type/) about this problem before commenting here.

goldfirere:
In an attempt to refresh this, I have integrated the data-constructor possibility into the proposal itself. I will submit to the committee soon.

AntC2:
Thanks @goldfirere ; something not very clear from the Proposed Change Spec (2)/GADT data constructor option (or am I being dim).

Roles are in the types namespace; data constructors are in the terms namespace. (Leaving aside for now promoted GADTs.) Furthermore a datatype might have many data constructors, of which some are GADTs, some not. So when the spec says

> ... inspects the type of a data constructor, it will assume the data constructor context as a given ...

* rather than _the_ data constructor, shouldn't it say _all/every_? What if different constructors have different contexts? Is it possible the rules can infer role `representational` for the type params from one constructor, but only `nominal` from other constructors? Should the rules say to default to `nominal` unless `representational` can be inferred for all type params for all constructors' contexts?

* what if the datatype is declared in another module, and the type is imported but not its constructors (or not all of them)? Can role inference still see the contexts of all constructors, even though they're not in scope in the importing module?

goldfirere:
Roles are indeed a property of the type, not the constructors. However, role inference works on a per-constructor basis. If one constructor forces an argument to be nominal (and all others allow it to be representational), then the argument is nominal. So the proposal means what it says here: each constructor is processed individually, with its own context assumed as a given. These individual analyses are combined to give the roles for the type. Role inference is done once for a type, at its declaration. Importing/exporting does not change the roles, nor does it trigger a re-do of role inference, so module boundaries aren't really an issue here.

I hope this helps!

cartazio:
Would this allow having join resume being in the monad type class ?

goldfirere:
> Would this allow having join resume being in the monad type class ?

This is already possible today. See https://gist.github.com/goldfirere/947a5d674c76b22fe508ece901d538b6, and you can uncomment the bottom section to see that it works without manual use of `coerce`.

Ericson2314:
```haskell
data (forall c d. Coercible c d => Coercible (v c) (v d)) => GList v a = Mk Metadata (v a)
type role GList representational representational
```
Doesn't just limit what things can be provided for `v`? Don't we want instead to condition the role for `d` on the choice of `v`? 

What I meant in https://github.com/ghc-proposals/ghc-proposals/pull/233#issuecomment-495818972 is something like:
```haskell
data GList :: (Type -> Type)   -- concrete representation
           -> Type             -- element type
           -> Type where
  Mk :: Metadata      -- some metadata; details irrelevant
     -> v a           -- contents
     -> GList v a
  instance Coercible (v c) (w c) => Coercible (GList v c) (GList w c)
  instance Coercible (v c) (v d) => Coercible (GList v c) (GList v d)
```
Now I am clearly not limiting the domain of `GList` or it's constructors, but instead declaring the `Coercible` instances I expect.
```haskell
data Some :: (Type -> Type)   -- given existential arg
           -> Type where
  Some ::  f a           -- contents
     -> Some f
  instance (forall a. Coercible (f a) (g a) => Coercible (Some f) (Some g)
```
Again here, I don't limit what `Some` can be applied to, but just what coercing I can do based on it's application.

Don't get me wrong, it is useful to restrict the domain of types, and that should be done in kind signatures / data contexts, but I don't feel like that's the heart of higher-order roles.

goldfirere:
The problem is that we can't have that fine level of control over `Coercible` instances: the relationship between coercible components and coercible wholes is baked into GHC's language of coercions. Maybe it can be done, but it would require a whole new theory and implementation. The best we can do now is to influence the role annotations on types, which has to be done by restricting the type.

Ericson2314:
Huh, I'm sort of surprised at that. It was my understanding that:
1. *Evalated* coercisions have no information at run time. Types are erased and coercing is `O(1)` so what could it have?
2. Coercsions are thus like any other computationally irrelevant proof.
3. The soundness of coercersions is just a special case of evaluating computationally irrelevant proofs to stay sound in a non-total language.
4. Everything "interesting" about coersions in particular must be in their static semantics.
5. The surface syntax `Coercible` is rather simpler than the special treatment of coercisons in core, and yet sound.
6. If only the static semantics is intresting, then we should drop all the magic from core keeping `Coercible` and remembering to evaluate our proofs.
7. We can change the implementation without changing the theory (`Coercible` all the way done no core magic). And the change the theory without changing the implementation (Can we admit new axioms like I propose without breaking the static semantics?)

I don't mean to dismiss all the good hard work that went into coercions, and don't doubt that my reasoning or assumptions is flawed, but just want to throw all that out there so I learn my exact mistake.

goldfirere:
Coercions are indeed challenging to understand from a Haskell user point of view. I'll try to respond to your points.

> 1. *Evalated* coercisions have no information at run time. Types are erased and coercing is `O(1)` so what could it have?

I'm not sure what an evaluated coercion is. Coercions are inert; they do not evaluate. We do have a coercion optimizer that looks for coercions that can be expressed more simply and updates them, but this optimizer is a purely internal device that should have no effect at all on understanding a Haskell program.

Types are erased. Coercions are erased. Coercing is `O(0)`, if I'm allowed to say that: the act of a coercion is simply missing from the compiled executable.

> 2. Coercsions are thus like any other computationally irrelevant proof.

Sure.

> 3. The soundness of coercersions is just a special case of evaluating computationally irrelevant proofs to stay sound in a non-total language.

I'm not exactly sure what you mean here, but yes: We need to have *some* proofs that we know are logically consistent. Normal Haskell terms aren't, but we make sure coercions are.

> 4. Everything "interesting" about coersions in particular must be in their static semantics.

Yes.

> 5. The surface syntax `Coercible` is rather simpler than the special treatment of coercisons in core, and yet sound.

The syntax is simpler, but the semantics isn't: I believe all the rules baked into Core coercions can be witnessed in surface Haskell.

> 6. If only the static semantics is intresting, then we should drop all the magic from core keeping `Coercible` and remembering to evaluate our proofs.

I'm afraid I can't make much meaning of this. Perhaps you mean: don't have coercions in Core. Instead, use the theory behind coercions to guide type inference, making sure that the produced Core is sound -- but drop the coercions themselves.

The problem with dropping coercions is that it effectively makes Core untyped. We have types in Core for a reason: they catch gobs and gobs of bugs in GHC. With `-dcore-lint`, GHC re-checks the output Core after every optimization pass. Without types, many of these passes would be riddled with (more) bugs. If we don't have coercions, then we have no way of converting from one type to another even when we "know" they're equal. So make any progress at all, we'd have to drop types, too. So we absolutely need coercions in Core.

> 7. We can change the implementation without changing the theory (`Coercible` all the way done no core magic).

I'm again not sure what you mean here. Changing the implementation without changing the theory is very dangerous: how would we know that the new implementation is sound?

(I admit: This proposal does not describe writing a new proof with the changes proposed. This is perhaps wrong.)

> And the change the theory without changing the implementation (Can we admit new axioms like I propose without breaking the static semantics?)

I don't know. Maybe. Someone would have to work out the details and write up a proof. It would be a different (and more complicated) approach to roles than we've taken previously. I don't see anything obviously wrong, but it's at least a full-length paper away from a notion I would trust.

---------------

Reading back, I think the fundamental difference in our understanding is around the importance of coercions in Core. We absolutely cannot live without them and have `-dcore-lint` work. And we absolutely cannot live without `-dcore-lint` and have GHC work.

Ericson2314:
> I'm not sure what an evaluated coercion is. 

Sorry yes coercions are indeed inert. I was thinking `Coercible`. Here's some polymorphic recursion that indeed builds up an arbitrarily big `Coercible` dictionary.
```haskell
{-# LANGUAGE FlexibleContexts #-}
import Data.Coerce
import Data.Functor.Identity
import Data.Word

polymorphicRecurCoercsion :: Word -> Int -> Int
polymorphicRecurCoercsion count n = go count n
  where
    go :: Coercible a Int => Word -> a -> Int
    go 0 n = coerce n
    go count n = go (count - 1) (Identity n)
```

But yes we want to remove all those thunks and still have the evidence in  the form of erased `~R` stuff. And I agree we definitely don't want to drop `~R`.

-----

All that said, I think I have some decent proof that we don't need to change  the implementation for my thing:

```haskell
{-# LANGUAGE GADTs #-}
{-# LANGUAGE FlexibleContexts #-}
{-# LANGUAGE ConstraintKinds #-}
{-# LANGUAGE KindSignatures #-} 
{-# LANGUAGE QuantifiedConstraints #-} 

import Data.Coerce
import Data.Functor.Identity
import Control.Monad.Trans.Identity
import Data.Kind
import Unsafe.Coerce

data Metadata = Metadata
  deriving Show

data Dict c = c => Dict

data Dict' v w c d = (Coercible (v c) (w d) => Coercible (GList v c) (GList w d)) => Dict'

data GList :: (Type -> Type)   -- concrete representation
           -> Type             -- element type
           -> Type where
  Mk :: Metadata      -- some metadata; details irrelevant
     -> v a           -- contents
     -> GList v a

deriving instance Show (v a) => Show (GList v a)

coerce' :: forall v w c d. Coercible (v c) (w d) => GList v c -> GList w d
coerce' = case asdf @v @w @c @d of Dict' -> coerce

asdf :: Dict' v w c d
asdf = unsafeCoerce (Dict @(Coercible Int Int))

x :: GList (IdentityT []) (Identity Int)
x = coerce' (Mk Metadata [1] :: GList [] Int)

main = print x
```

Ericson2314:
Also, I forget about `deriving instance` when I made through out my syntax above. I'm changing it to use that.

goldfirere:
`-dcore-lint` is purely a check on GHC: it makes sure that GHC doesn't do (certain) silly things. It will *not* check that you do not do silly things. And, I'm afraid to say that you have. Append the following to your example:

```
data Phant a where
  MkPhant :: Int -> Phant a

y :: GList Phant Bool
y = coerce' (Mk Metadata (MkPhant 5) :: GList Phant Int)

oops :: Bool -> Int
oops n = let coPhantBoolPhantInt :: Coercion (GList Phant Bool) (GList Phant Int)
             coPhantBoolPhantInt = case asdf @Phant @Phant @Bool @Int of Dict' -> Coercion

             extract :: Coercion (GList a b) (GList c d) -> Coercion b d
             extract Coercion = Coercion

             coBoolInt :: Coercion Bool Int
             coBoolInt = extract coPhantBoolPhantInt
         in
         coerceWith coBoolInt n

main = print (oops True)
```

and you will see the integer representation of `True`. Oops!

Ericson2314:
@goldfirere you can only write extract because you are using GHC's inferred role-based axioms in addition to mine. But mine is supposed to replace it. Declare all roles nominal to emulate that, and then soundness should be restored. Will edit to elaborate once I'm off my phone.

goldfirere:
Ah -- yes, it's possible that there's a different way to skin this cat. That is, one might come up with a different set of rules for coercions that will allow what you want (but, likely, will allow less of what we have today). Designing it is a non-trivial amount of work.

Ericson2314:
Unless I'm missing something I think it's the optimal way to skin the cat. It works just like DeriveFunctor for but N arguments. You look at every field, substituting `paramN` for `paramN'` to derive the per-field constraints, and accumulate. This seems sound and complete to me.

Ericson2314:
Ah! OK. So I can also write:
```haskell
extract :: Coercion (GList a b) (GList c d) -> b :~: d
extract Coercion = Refl
```
Normally in Haskell you can only derive superclass constraints, but here we are deriving what's analogous to *instance* constraint stemming from the nominal role. That throws up `Coercision` becoming an actual class (though only with derived instances) but some things still would work with a scheme like mine. I think this is OK:
```haskell
data Nomin a where
  MkNomin :: Int -> Nomin a
type role Nomin phantom

extract :: Coercion (GList Nomin b) (GList Nomin d) -> Coercion b d
extract Coercion = Coercion
```
In particular, we would have axioms like:
```haskell
deriving instance Coercible (v c) (w d) => Coercible (GList v c) (GList w d)
deriving instance a ~ b => Coercible (Nomin a) (Nomin b)
```
And yes, there is a backwards compatability problem, namely that while my way might be complete with respect to coersions themselves, that does mean less deriving of equalities from lack of coerciability. More `Coercisible` less `~` is just a fundamental tradeoff. I think the way to get around it is to say by default only the first-order axioms like today are derived, and the higher order ones must be opted-in with standalone deriving. Users then get to opt-in to the breaking change themselves.

goldfirere:
There are a lot of fun ideas that you're proposing, but there is a *ton* of work to see these to completion -- and we might discover that the ideas do not actually work. The key to `Coercible` is that coercing is zero-cost: there is no runtime behavior. That, in turn, relies on having evidence in Core for coercibility. This evidence must be absolutely reliable, meaning it needs to have a syntax outside that of normal Haskell. (Normal Haskell risks non-termination, for example.) These proposals have not described that syntax or how any of this would work, in detail. I'm not inclined personally to revisit what we have, as this is really quite substantial. But if someone else were to do the work of formalizing all of this (somewhat along the lines of [this paper](http://repository.brynmawr.edu/cgi/viewcontent.cgi?article=1010&context=compsci_pubs)), I'd be very interested to look through and see how we can update GHC.

I'm not trying to just shut off conversation on alternative designs -- just pointing out that a suggestion made in a thread like this doesn't quite have enough weight to make GHC change direction here.

However (and separately), I fear this conversation has wandered off course. Other than new role systems, are there other concerns/suggestions about this proposal?

Ericson2314:
@goldfirere 

> But if someone else were to do the work of formalizing all of this (somewhat along the lines of this paper), I'd be very interested to look through and see how we can update GHC.

I won't say anything further in this thread, and make my own proposal.

> But if someone else were to do the work of formalizing all of this (somewhat along the lines of [this paper](http://repository.brynmawr.edu/cgi/viewcontent.cgi?article=1010&context=compsci_pubs)),

I've started reading it. By any chance are the appendices' proofs made with the help of a proof assistant, and is the code open? I might have some fun trying to modify it.

> ...That, in turn, relies on having evidence in Core for coercibility...

That was why I mentioned the `-dcore-lint` lint before: not because it would sign off on my `unsafeCoerce` (it certainly won't!) but to show that there is indeed core evidence which is good enough to pass the lint for the use of `asdf`. I've included part of the simplifier dump below. Conveniently, everything up to `asdf` was inlined into `main` so it shows all the safe bits and none of the unsafe bits.

<details><summary> `-ddump-simpl` output</summary>

```haskell
main :: IO ()
[GblId]
main
  = print
      @ (GList (IdentityT []) (Identity Int))
      (Main.$fShowGList
         @ (IdentityT [])
         @ (Identity Int)
         (Control.Monad.Trans.Identity.$fShowIdentityT
            @ [] @ (Identity Int) Data.Functor.Classes.$fShow1[] $dShow_r39o))
      (case asdf_rJJ @ [] @ (IdentityT []) @ Int @ (Identity Int) of
       { Dict' df_a2O6 ->
       case df_a2O6
              (GHC.Types.MkCoercible
                 @ *
                 @ [Int]
                 @ (IdentityT [] (Identity Int))
                 @~ (([Sym (Data.Functor.Identity.N:Identity[0]
                                <Int>_R)])_R ; Sym (Control.Monad.Trans.Identity.N:IdentityT[0]
                                                        <*>_N <[]>_R) <Identity Int>_N
                     :: [Int] ~R# IdentityT [] (Identity Int)))
       of
       { GHC.Types.MkCoercible v_B3 ->
       (Main.Mk
          @ []
          @ Int
          Main.Metadata
          (GHC.Types.: @ Int (GHC.Types.I# 1#) (GHC.Types.[] @ Int)))
       `cast` (v_B3
               :: GList [] Int ~R# GList (IdentityT []) (Identity Int))
       }
       })
```
</details>

goldfirere:
> I've started reading it. By any chance are the appendices' proofs made with the help of a proof assistant, and is the code open? I might have some fun trying to modify it.

No proof assistant; just me thinking hard. That said, I was doing a bunch of Coq proofs at the time, and used the same process in writing these paper proofs. If there were support for names (Isabelle?), it might be relatively easy to mechanize this.

I've sent you a buildable tarball by email.

Ericson2314:
@goldfirere thanks!!

Ericson2314:
I threw up a very WIP https://github.com/ghc-proposals/ghc-proposals/pull/248. Will refine over time.

treeowl:
I support the per-constructor version. It's purely an improvement, and I think we'll want something like that whatever else we do.

I oppose the `DatatypeContexts` version. I think it's wrong for the same reason `DatatypeContexts` has *always* been wrong. We've always said "put the constraints on the function, not the type," and I think that's exactly what we want to do here too. That program, I think, starts with something like #248. What I'd like, ultimately, is to be able to *name* coercion axioms and either "open" them in a module or use them in more local contexts. But if people start using `DatatypeContexts` to enhance coercibility, we'll be stuck with a bunch of types we can't use how we want. Suppose someone put such a constraint on `ReaderT`. Then `ReaderT` will *stop working* with `Monad` instances that have the wrong type roles. That's a disaster, so it won't happen, and this whole thing will accomplish nothing.

treeowl:
As for unpacking coercions, you already know I want that. My proposal to find a way got stuck in the mud, but I don't think that needs to be the end of it!

Ericson2314:
@treeowl FWIW I think both the data contexts and the constructor version would not work for `ReaderT` for those reasons.

But I think that raises a separate good point, there's no reason we couldn't have *both* datatype contexts and constructor contexts have an effect, and that strikes me as the most correct.

treeowl:
> @treeowl FWIW I think both the data contexts and the constructor version would not work for `ReaderT` for those reasons.

Exactly! That's because the solution to `ReaderT` is something like your own WIP proposal. But Richard's idea makes good sense in some other contexts.

treeowl:
And I don't see any expressivity benefit to the `DatatypeContexts` approach; in combination with your proposal it becomes a mere convenience in some situations but an impediment in others. Better, I think, to make sure that it's convenient enough to express what we need in constraints at use sites. Those use sites may, of course, be in GADT constructor types, which is where this proposal can step in and help.

goldfirere:
I have updated the proposal to allow these new constraints either in the datatype context or in the constructor context. Seems strictly better than either previous alternative, and I'm a little embarrassed at not proposing it this way from the start.

Anyway, conversation that would affect the proposal has ebbed. I plan on submitting soon.

AntC2:
> @goldfirere I'm a little embarrassed at not proposing it this way from the start.

There were reasons why not; and I'm still worried by them; and I don't think you've answered [@simonpj's I am troubled by the idea of computing dictionaries that we never use.](https://github.com/ghc-proposals/ghc-proposals/pull/233#issuecomment-498561878)/general distaste for datatype contexts.

The datatype context syntax is very suggestive (of what, seems to be debatable); we can at least agree H98 semantics isn't delivering the suggestion. So the syntax is on track for deprecating. We've kept the syntax reserved so it can come back with improved semantics.

If this proposal blesses or half-blesses the syntax; what semantics is it blessing? Is that semantics consistent with whatever will be the future semantics? Will the future design get hobbled by the need for backwards compatibility with this proposal?

goldfirere:
You're right that this design still forces computation of unused dictionaries. I believe these can be optimized away, as the proposal itself argues. And it's appropriate to be skeptical of such a design, but I don't have a better one right now.

As for datatype contexts, I have a [concrete proposal](https://richarde.dev/papers/2019/partialdata/partialdata.pdf). I don't know if that's really the right answer, but it's completely compatible with this proposal.

AntC2:
> @goldfirere As for datatype contexts, I have a concrete proposal. ...

Deliberately I haven't looked at that proposal, because:

Then are you turning this proposal into a design for datatype contexts? It's neither the time-pressure "I plan on submitting soon"; nor the place (how many interested in datatype contexts would know the design is going on under 'Higher-order roles'); nor the timescale in the deprecation process; nor the open/transparent approach intended for Github proposals.

I hate finding myself in the role of Mordac the Refuser here. I think this proposal (and documentation for this feature) should at least be explicit that you're making assumptions about the future of datatype contexts; and that anybody exploiting datatype contexts to specify Higher-order roles (as opposed to exploiting per-constructor GADT contexts) should beware some other design for datatype contexts might eventuate whereupon they might need to refactor their code.

goldfirere:
> Then are you turning this proposal into a design for datatype contexts? 

No. I am not proposing the content of that paper, and I have no concrete plans to do so. You expressed a concern about the eventual semantics of datatype contexts. This makes sense: the feature is clearly "in transition" (to what, we're not quite sure) and so we want to know if this new idea is future-compatible. I link to that paper on datatype contexts solely to make concrete one possible future. Of course, we cannot plan for *all* possible futures, so we just do our best for future-compatibility.

The current proposal is future-compatible with three known possible futures of datatype contexts: the future where we don't have them any more at all, the future that looks just like the present (i.e., keep the feature deprecated in perpetuity), and the future that looks like that paper. To me, that's a convincing argument that this proposal is future-compatible with datatype contexts. Of course, you're free to imagine another future for datatype contexts that is not compatible with this proposal.

goldfirere:
Conversation has died down here, and I think the proposal has landed in a good place. It solves the motivating problem and allows flexibility in whether users employ datatype contexts or data constructor contexts.

I'd like to submit to the committee, @nomeata.

proposals/0000-higher-roles.rst
```diff
@@ -0,0 +1,135 @@
...
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/233>`_.
+.. sectnum::
+.. contents::
+
+If we write ``data T f a = MkT (f a)``, GHC assigns a nominal role to ``a``, because it appears
+as the argument of a type variable, and we cannot know how this argument will be used. This proposal
+describes a way to assign a representational role to ``a`` via the use of a quantified constraint.
+
+There are two proposed ways to fix this problem, each described in its own "Proposed Change
+Specification". Only **one** of these changes should be implemented.
```
aspiwack: I believe this paragraph doesn't hold in the current version of this document.

proposals/0000-higher-roles.rst
```diff
@@ -0,0 +1,135 @@
...
+------------
+* Instead of using a quantified constraint and invoking the solver, we could come up with a more
+  restrictive syntax for this idea, which expands out to a quantified constraint on every constructor.
+  This would mean we wouldn't run the solver during role inference. However, such a feature seems
+  sadly non-orthogonal when we have the features to express this idea already.
+
+* Previous versions of this proposal had the constraints either in the datatype context or
+  in the constructor contexts, but that decision was made at the level of the proposal.
+  Now, users can make their own decisions. This seems strictly better than either previous
+  proposal.
```
aspiwack: This is not super clear. After a handful of reading I now assume that you mean: in previous versions of this proposal, the context used by role inference was fixed either to be from the datatype context or the constructor context; the proposal is more flexible than either, as you can use both.

aspiwack (commented):
Some comments on form.

isovector:
This addresses a real-world problem I've also run into (unfortunately can't remember the details), so I'm genuinely interested in a solution. 

That being said, this whole business of packing a datatype context of a quantified coercible constraint seems like a pretty big hack. The underlying issue appears to be a byproduct of the fact that we don't have role arrows/poly roles; in the example here it seems like we'd like to have the `a` in `f a` to be at role `representational`.

Just spitballing here, but what if I could instead write a role signature:

```haskell
type role GList ({- 1 -} a -> {- 2 -} representational)  {- 3 -} a
```

in which `2` describes the fact (like today) that `f` itself is representational, but `1` is a variable corresponding to the role of the eventual argument of (concrete) `f`, and `3` states that `a` will have same role as that.

I'd understand if such a change feels like more work than it's worth for the problem, or if it comes with soundness issues or what-have-you. But I think it's worth keeping in mind that this proposal treats more of a symptom than a cause IMO.

Ericson2314:
@isovector I think you are basically moving towards asking for my https://github.com/ghc-proposals/ghc-proposals/pull/248. I do intend to return to that, but I need to spend time hunting down more research (about what the patternson conditions mean for the termination of dictionary evaluation, as opposed to the termination of the solver).

I still think this proposal is a good step, as @treeowl says above. In all formulations, the correctness of some coercion rule (be it the roles or a more free-form axiom) needs to be decidable. Currently, it just uses global information, which is weird and unlike the other class/constraint parts of the language. This makes it use local information too, which is more consistent. I think the consistency is good even if users rarely/never need to resort to the constructor or data kind contexts in the end.

goldfirere:
@isovector :

> this proposal treats more of a symptom than a cause 

I agree. But really doing roles right is very involved -- we'd have to account for arbitrary nested of functions, for example. Working that out would take at least another peer-reviewed publication, if not more. (The current design took two!) This proposal takes the current, somewhat weak design for roles and tweaks it just a bit.

isovector:
@goldfirere understood, thanks. No more objection on my end.

simonpj:
I'm the shepherd for this proposal.  I discussed it with RIchard (author) today, and we agreed to put it back into "Discussion" state so that
* He can add a (much) more precise specification of the change.
* He can think about what is definitely sound. (We discussed how we can be confident about soundness in System FC if we end up discarding evidence that is crucial for soundness.)

Icelandjack:
Someone mentioned specifying axioms for coercible directly, which made me think of it as a new deriving strategy `role`

```haskell
data GList f a where
  Mk :: .. -> GList f a
 deriving
 role forall a b. Coercible => Coercible (f a) (f b)
```

Ericson2314:
@Icelandjack that seems closer to what I wrote in https://github.com/ghc-proposals/ghc-proposals/pull/248, unless you wish to constrain the derivied "instances" to be of the "shapes" we get today?

cartazio:
Question : how would this proposal impact or help eg mtl or transformers or vector , from either a library author or library user perspective ?

AntC2:
Re the [vote back here](https://github.com/ghc-proposals/ghc-proposals/pull/233#issuecomment-498672441):

> Give a "thumbs up" to this comment if you prefer the coercible **constraints on constructors**.
> 
> Give a "thumbs down" to this comment if you prefer the coercible **constraints in datatype contexts**.

I'm not sure it's either-or: perhaps we need both. Because _pace_ SPJ's comment:

>  I'm deeply suspicous of anything involving datatype contexts, a mis-feature IMHO.

you can get what seems [equally mis-feature-ish behaviour out of GADTs](https://stackoverflow.com/questions/63187598/confused-about-gadts-and-propagating-constraints): two function definitions that are morally equivalent, but have different inferred type depending where the data constructor's constraint gets swallowed.

cartazio:
whats going on with this or other discussions? it'd re really nice to be able to do the following sans unsafe coerce, even though we "know" it should be safe


```
{-#INLINE coerceThroughFunctor #-}
coerceThroughFunctor :: forall a b f.  (Coercible a b, Functor f) => (Coercion (f a) (f b))
coerceThroughFunctor = (unsafeCoerce (Coercion :: Coercion a b  )) :: (Coercion (f a) (f b))

```

goldfirere:
This, sadly, is in need of a proof of type safety. I would like to have such a proof, but I don't as of now, and I don't wish to promise a timeframe for when I might develop one. If someone else would like to prove the safety of this system, that would be fantastic.

Ericson2314:
I don't think this will work as well as we hope because the quantified constraints are boxed (and furthermore a function between boxed things). A coercion cannot force a box because that would constitute doing work. And creating unboxed *true* coercion implications is a rather taller order theory wise, though perhaps an incremental step we can do after.

But by getting rid of `Co_Nth` so the push rules instead work field wise, it then becomes actually quite easy to prove that one can take advantage of the coercion fields that have already been coerced. It's only when our only hammers are `Co_Nth` and `[[_]]` that we don't have a chance to access to the intermediate coercion fields because the proof structure is one-size-fits-all and tackles each field in parallel.

(Relatedly, https://repository.brynmawr.edu/cgi/viewcontent.cgi?article=1010&context=compsci_pubs coerces coercion like this: `syn [[c1]]; c0; [[c1]]`, which is to say it always tries to substitute the variables by coercion. But just as we have a `Tm_Contra` we could have a `Co_Contra` that allows coercion an absurd coersion into any coercion. This wouldn't help with Functors but would help with GADTs, where we want to be able show that certain constructor are impossible to skirt the obligation of coercing them.)

All this is why in https://mail.haskell.org/pipermail/libraries/2021-January/031096.html I say the thing to start with is formalizing https://github.com/ghc-proposals/ghc-proposals/pull/276. It's perhaps "not enough" payout it's own, but I'm pretty sure it gets us closer even though I am not quite sure what to do for the "last mile" of axioms / pseudo-instances / solving.

noughtmare:
How about just letting the user specify roles on the function arrow type just like we have linearity annotations? That should be relatively simple now that the modifiers proposal is implemented.

I'm imagining writing this:

```haskell
data GList :: forall (r :: Role). 
    (Type %r -> Type)   -- concrete representation
  -> Type %r            -- element type
  -> Type
  where
    Mk :: Metadata      -- some metadata; details irrelevant
       -> v a           -- contents
       -> GList v a
```

I also think this could be combined with matchability annotations from #242 (i.e., add an `unmatchable` role which is like `nominal` but without the implication `f a ~ g b => (f ~ g, a ~ b)`).

Edit: or we don't even need polymorphism for the main example:

```haskell
type R :: Role
type R = Representational
data GList :: (Type -> %R Type) %R
           -> Type %R
           -> Type where
  Mk :: Metadata
     -> v a
     -> GList v a
```

AntC2:
> roles on the function arrow

A couple of (related) questions. 1) If the annotation applies to the arrow (rather than the bits between the arrows [**]), shouldn't we show `GList` as

```
data GList :: forall (r :: Role). 
       (Type %r -> Type)   -- concrete representation
     -> Type               -- element type
  %r -> Type               -- result
  where ...
```

2) Could the role apply for the resulting `Type`? (That is, to the right of all arrows.) What role applies for the result `Proxy a`, whose type param is Phantom/how to encode that? (In particular so that a function `unProxy` won't be in danger of returning anything Coercible.)

```
data Proxy a  where
     Proxy :: Proxy a

unProxy :: Proxy a -> a
unProxy = const (error "no usable value possible here")
```

Or ... The answer to 1) is that the `%r` applies not to the arrow but to the `Type` preceding it (in which case the encoding you show is more felicitous); but then the answer to 2) might allow:

```
data Proxy :: forall (r :: Role). 
    Type %r -> Type %r
 where
     Proxy :: Proxy a
```

[**] [`LinearTypes` doco](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/linear_types.html) seems to intend the annotation is part of the arrow:

> If `UnicodeSyntax` is enabled, the `%1 ->` arrow can be written as `⊸`.

noughtmare:
Yes, I intend the annotation to apply to the arrow, just like linearity annotations (and the proposed matchability annotations which this would subsume). I prefer the (I believe common) style of writing them on the line of the argument, though.

I don't understand the motivation for question 2. Why would the role annotations affect what you could return from `unProxy`? Currently role annotations only affect when you can use `coerce` (and at the type level, how you can use certain type variables).

I also don't think this is very relevant to my proposal. Even with the current role annotations you could imagine making it possible to write `type role Proxy phantom -> phantom` or something like that to signify a "return type role annotation" if you can find a good use case for it.

PR #234, opened 2019-05-25
Local control of warning flags

Pluralia:
We propose to add functionality for control warnings locally, in particular, add pragmas `WARN`, `IGNORE` and `ERROR`. Consider this code:
```
{-# OPTIONS_GHC -Wname-shadowing #-}

main :: IO ()
main = {-# IGNORE name-shadowing #-} do
  let x = "some data"
  let x = "updated version of that data"
  let x = "further changed version of data"
  print x
```

Here in each case, you are effectively throwing away the previous version of the data without any warning but if you remove IGNORE you get `-Wname-shadowing`. You can rewrite some lines to rid of it:
```
let x1 = "some data"
let x2 = "updated version of that data"
let x3 = "further changed version of data"
```
then you have all of `x1`, `x2` and `x3` in scope and you may make an error by picking an older numbered x, when conceptually you really only want one x in scope.

[Rendered](https://github.com/pluralia/ghc-proposals/blob/opt-loc-prag/proposals/0000-local-warn-pragmas.rst)


isovector:
Love it. I'm not sure if your proposal is to also add local extensions? (if so, *fuck yeah*!)

What is the behavior of setting local options for things that probably can't be local? I'm thinking particularly about the optimization flags and `-fplugin`? Silently ignore them, or a warning or something?

kirelagin:
@isovector I think, section “1.2 Using language extensions locally” explicitly gives an example of a local use of an extension. And the proposal, in general, is talking about allowing _any_ compiler flags (that is, including the ones that enable/disable extensions).

To which you rightly point out that there are also things like optimisation flags and plugins. Which means that the proposal is actually to allow overriding all flags _for which it makes sense_... and I am wondering how to explain this in a proposal, other than by enumerating all “kinds” of flags 🤔.

harpocrates:
Language pragmas get a pretty special treatment from GHC, so it might be a bit tougher to have those be turned on only locally. Before GHC even starts parsing a file, it'll try to figure out what language extensions are enabled. Part of the reason for this is that extensions often impact parsing...

int-index:
@harpocrates Indeed, but this is purely an engineering problem, not a design issue. We'd have to change the `P` monad to use `DynFlags` instead of `ExtsBitmap` and then set and unset the flags appropriately, which is possible because we know when we reach the end of an expression, a type, or a declaration.

harpocrates:
For some extensions this will work fine. For others it will be impossible (ex: `-XCPP`). I also fear that some of the layout extensions will be particularly difficult to get right...

Pluralia:
@isovector I updated proposal for cases of misuse. I think some warning can be appropriated. White list of use can be created in the process of implementation.

kirelagin:
@harpocrates Again, for the case of `CPP`, I can’t see anything that prevents switching it locally, _in theory_. It is just the current implementation with a separate phase makes things difficult.

Initially, I was going to add that it is fine, because switching `CPP` locally makes no sense... but then I realised that someone might have `OverloadedLabels` and then want to use a label called `#ifdef`. So, again, in principle, there is nothing wrong with proposing to do this – it is just that the implementation will be challenging.

kirelagin:
In general, there is a whole lot of questions, and the most appropriate strategy seems to be to solve them on a case-by-case basis. Example:

I think it is a great idea to enable `RebindableSyntax` locally. However, currently it implies `NoImplicitPrelude`, for which it makes absolutely no sense to switch it locally, right? I see two possible solutions:

1. If `RebindableSyntax` is enabled even locally, enable `NoImplicitPrelude` globally. I find this clumsy and counter-intuitive.
2. Remove the implication, and assume that if someone uses `RebindableSyntax` and did not enable `NoImplicitPrelude`, then it is their problem. I think it is the right thing to do, but it breaks compilation, so it will definitely require a proposal of its own.

AntC2:
Now there's a weird coincidence. I've just written a [heavy criticism of this whole idea](https://gitlab.haskell.org/ghc/ghc/issues/602#note_200935). In particular, the example for flagging orphan instances is unnecessary, use `{-# OVERLAPPABLE #-}`, and get GHC to treat it better. I won't bother to repeat that explanation.

nomeata:
I agree with the motivation: Disabling certain warnings very locally is certainly useful, and I can see how some language extensions might work locally as well.

But I think “*command line* flags” are the wrong level of abstraction here. Note that we already have `{-# LANGUAGE FooBar #-}` so that people don’t have to use `{-# OPTIONS_GHC -XFooBar #-}`, for good reasons. So I would advocate for coming up with declarative pragmas for the various use-cases, instead of just saying “anything goes, let the compiler complain if it doesn’t make sense”.

For warnings, GHC, already has a first-class notion of the name of a warning (`orphans`, `unused-do-bind`) together with a little DSL to turn them on and off or into warnings (`-W`, `-Wno-`, `-Werror-`). So I might be able to support adding declarative, local pragmas that target warnings (`{-# WARN orphans #-}`, `{-# IGNORE unused-to-bind #-}`, … feel free to come up with separate names).

Can such warning pragmas scope over individual import statements? (I guess they are just declarations for this purpose?)

int-index:
@nomeata When working on GHC, I imagine adding `{-# OPTIONS_LOCAL -ddump-tc-trace #-}`, what would be the alternative with your syntax?

nomeata:
First, I find it odd to have temporary options that affect dumping inside the source file (I specify these things on the command line).

Then I am not sure if it even makes sense to do these options locally: isn’t type inference a module-wide thing? Maybe something can be done locally … but I am not sure if the engineering for that is worth it. And even if it is: yes, a dedicate pragma might be better.

bgamari:
I strongly agree with @nomeata here. Command-line flags are a great interface for configuring a program on the command-line. They are not the right concept for source code. Not only does specifying this pragma in terms of command-line flags require us to handle (e.g. by failing) inherently global flags (e.g. `-optc` or `-fspecialise`) but it also makes it hard for us to document (and for the user to discover) which flags are valid.

AntC2:
From recent comments, is there some support for a limited local pragma applying for warnings only? (That is, not for `LANGUAGE` extensions or object code generation/run-time behaviour.) If GHC issues a warning, it's going to accept the program anyway, and use the same type inference and code generation.

I know some development shops have a rule that production code must compile without warnings. Some warnings it must be really hard to avoid. Would they regard suppressing warnings locally as 'cheating'? or useful?

I am still not seeing compelling use-cases for a _general_ facility, as opposed to ad-hoc pragmas for specific features. The proposal demos `-Wname-shadowing`. Yes that's a useful warning. I can't imagine why you'd want to switch it off locally: just renaming your local vars would be no more trouble (? isn't that what IDEs are good at).

For a general facility, there should be lots of use cases; and there should be a generic specification with easy-to-understand rules about the scope of a `{-# OPTIONS_LOCAL ... #-}`. For the `-Wno-unused-do-bind` example, my intuition expects the pragma would appear immediately after the `do`. (That follows the idea that pragmas appear immediately after a keyword and before the 'meat': `instance {-# OVERLAPPABLE #-} ...` ) Appearing after the `=` of a function definition doesn't seem right; what about a series of equations with pattern matches for a function definition? Where do I switch off a warning for the whole function?

nomeata:
> From recent comments, is there some support for a limited local pragma applying for warnings only? 

Yes, absolutely. I have recently used this in Ocaml (disabling a “module unused” warning for an single local module), and I could see this being useful in GHC.

> I can't imagine why you'd want to switch it off locally: just renaming your local vars would be no more trouble (? isn't that what IDEs are good at).

Maybe for a subset of your module, you want to use the `do x <- …; x <- return $ …` idiom? I am sure there are valid reasons to disable this warning, otherwise it wouldn't be a warning in the first place, but an error.

> I am still not seeing compelling use-cases for a general facility, as opposed to ad-hoc pragmas for specific features.

When you say *general facility*, you mean mean general as in “all flags” or “all warnings”?

AntC2:
> @nomeata When you say _general facility_, you mean mean general as in “all flags” or “all warnings”?

No, I mean where you could put an arbitrary list of warnings inside each `{-# OPTIONS_LOCAL ... #-}`.

>  I am sure there are valid reasons to disable this warning [`-Wname-shadowing`], otherwise it wouldn't be a warning in the first place, but an error.

That would go rejecting valid H2010 programs, which GHC doesn't do. But yes your example of pseudo-destructive assignment to a var in imperative style within a `do` block makes sense.

goldfirere:
I tend to fall on the side of preferring specific solutions for specific problems in this area. (This is counter to my usual preference for very general solutions!)

1. Doing this right requires enumerating every warning (resp. flag) and describing what its semantics should be. (For example, can we give a reasonable local semantics to the `type-defaults` warning? There may be other challenging ones, too.)

2. A developer (resp. shop) may have reasons to enable certain warnings. Allowing local defeat of these warnings seems counterintuitive. I worry that soon we'll need to enable the existence of local warnings via a language pragma (that could be forbidden).

3. An important design principle behind most warnings is that they are easily suppressed. Examples: names never *need* to be shadowed, `_ <- blah` works nicely to suppress unused results, binding `_x` suppresses the warning about unused variables, `import Foo ()` suppresses the warning about an unused import, etc. If a warning is not easily suppressed (e.g., orphans), then I'd support some fresh pragma to deal specifically with that warning.

scott-fleischman:
Here are some motivations for this proposal that I haven't seen expressed yet.

1. Dealing with **other people's code**. With a large codebase that uses lots of libraries and limited developer resources we need to respond to changes in libraries as we update to more recent versions. We may not agree with decisions of various libraries, but we do have to respond to them, and we may not be able to make the fully correct response immediately.
2. Allowing **local exceptions** to warnings allow us to turn on warnings globally but allow local exceptions that we can document where they came from and why we are not able or willing to change them in the short term. This could be because it's not technically possible or because we are not willing to invest the time and effort to make the changes now. (We can file a ticket to improve it later.)
3. We can **quarantine deprecations** more easily. We turn on the warning for use of deprecated code, but often libraries make choices that make it hard to immediately remove the deprecated code. One case I ran into recently was a library deprecated a record field that is still even used internally by the library. The library disabled the deprecation warning in the entire module in their own code, and we are forced to also disable deprecations in our modules that use the field, or to quarantine our use of that field to a separate smaller module that only has code using the deprecated field. It would have been nicer to indicate which deprecated field that we are intentionally using to avoid allowing any deprecated code to be used in the module.
4. A local declaration also provides **documentation** about which warnings we are disabling and why. In particular if the syntax for local pragmas is unique enough, it makes common search/replace an easy way to gauge how large a task it would be to update all of it at a future time.
5. With **variable shadowing**, generally it is useful to turn on that warning globally because it can help avoid a certain class of bugs. But there are a few cases where shadowing can help avoid another class of bugs. Turning off the shadowing warning locally, could help in cases for code that looks more like
```
  let x = some data
  let x = updated version of that data
  let x = further changed version version of data
```
where in each case, you are effectively throwing away the previous version of the data. If you write that code as:
```
  let x1 = some data
  let x2 = updated version of that data
  let x3 = further changed version version of data
```
then you have all of `x1`, `x2`, and `x3` in scope and you may make an error by picking an older numbered x, when conceptually you really only want one `x` in scope.
6. **Orphans**. The case I have run into with more down-to-earth code is that for whatever reason we need to define an orphan instance for some type in an external library. I think it serves a nice **documentation**-like purpose to keep those instances local to avoid allowing any orphan in an entire module. Later we can search for the local instance declarations and revisit the decision to use them.

tejon:
Orphans by far the most common warning I suppress on a per-module basis, and I would love to do it per-instance instead.

Another example from my actual production code is unused top-level bindings warnings from records which are only declared to enable autogenerated Aeson instances.

goldfirere:
I see all the motivations in @scott-fleischman's https://github.com/ghc-proposals/ghc-proposals/pull/234#issuecomment-496661390 -- thanks for writing that up. But I wonder if some small subset of warnings are the ones in contention here. For example, I'm in full support of a way of writing an orphan without a warning, and I think there should be a way of suppressing a deprecation. However, I think these should be specific cases that we address.

For example, in the case of deprecations, I worry that switching off deprecation warnings in a region of code is too big a hammer: perhaps you're aware of one deprecated entity (which you wish to use regardless) but another entity in that region of code also gets deprecated (and you don't wish to use). By switching off the deprecation warning, you can't see that this second deprecation happened. Instead, I think (in the case of deprecations) that it would be much better to have `{-# LOCALLY_UNDEPRECATE foo #-}` (that's a terrible name) than `{-# NO_WARNING_LOCAL deprecations #-}` (or similar). By treating warnings case-by-case, I think we can build the best design for each.

In some cases, perhaps the warnings themselves are the wrong level of granularity. For example, maybe one warning shouldn't encompass both `f x = let x = blah in ...` and `do x <- blah; x <- blah2`.

Concretely, I'm proposing that the proposal list (and specify the behavior of) a few specific such pragmas. The commentary can guide the expansion of such a list. Of course, each pragma could be a separate proposal, but that seems overwrought.

scott-fleischman:
I am ok with a narrow scope for which local pragmas are allowed.

As far as a concrete list goes, personally I have been motivated to want a local override for the following warnings. These are more-or-less in order of priority (lower number = more important). There might be a few others in the list, but they aren't things I have thought about recently.

1. deprecations
2. orphans
3. name-shadowing
4. unused-top-binds
5. redundant-constraints
6. incomplete-patterns

I mentioned the motivations for most in https://github.com/ghc-proposals/ghc-proposals/pull/234#issuecomment-496661390.

`unused-top-binds` as motivated by @tejon in https://github.com/ghc-proposals/ghc-proposals/pull/234#issuecomment-496673537 (fields that are unused in one respect, but used by template haskell or generics).

A redundant constraint can be useful to express the intention of the code for purposes of **clarity** even when not strictly necessary.

`incomplete-patterns`: we might choose one place where we reluctantly decide that we don't want to match all patterns (due to other people's code, or maybe your own poor choice in the past but not able/willing to fix right now), but we certainly want to check for complete patterns everywhere else in the module.

It seems a common pattern among these is that it:
1. can be attached to a specific name (binding, typeclass name, record field)
2. the warning occurs at the definition site

One exception to 2) might be the shadowing case, which is more about where the next identical binding would override it--I wouldn't want to disable shadowing everytime I make a new shadowed version of the same binding.

Pluralia:
The idea to add local language extensions was rejected - please, see updated proposal.
Summary:
- local warnings remain and now we use [an idea](https://github.com/ghc-proposals/ghc-proposals/pull/234#issuecomment-495977461) to split one pragma to subpragmas by @nomeata.
- motivation was updated pursuant to [the remarks](https://github.com/ghc-proposals/ghc-proposals/pull/234#issuecomment-496661390) by @scott-fleischmanI
- as alternatives - pragma `OPTIONS_LOCAL`
- add black- and whitelist of warnings
- there are local language extensions in uresolved questions

kirelagin:
I like the idea of a specialised pragma focused on warnings much more.

While the choice of the name `WARN` is clearly perfect, my concern is that `IGNORE` and `ERROR` sound too generic, non-descriptive, and their names do not really relay the purpose. Personally I would go with

* `WARN`
* `NOWARN`
* `WERROR`

Or something something along these lines. For example, it can be `WARN_NO` to, sort of, mimic the cli syntax, but `NOWARN` sounds more consistent with `NOINLINE` and `NOUNPACK`.

Pluralia:
@AntC2 I apologize. This proposal had been written, but hadn't been committed yet by the time you wrote the critics - your arguments reminded me of the inconsistency of the local extensions idea and became one of the reasons to reject it for now. I think they will be useful when this question is raised again. As well as remarks by @harpocrates @kirelagin @int-index @nomeata

Pluralia:
@nomeata Thanks for an idea to separate pragmas name!

> Can such warning pragmas scope over individual import statements? (I guess they are just declarations for this purpose?)
Yes, you're right.

Pluralia:
> what about a series of equations with pattern matches for a function definition? Where do I switch off a warning for the whole function?

@AntC2 If I’m correct in understanding of your question, pragma placed before declaration can solve this problem.

Pluralia:
@goldfirere 

> Doing this right requires enumerating every warning (resp. flag) and describing what its semantics should be.

You're right. But warnings aren't language extensions and are easier to make, I suppose. In the updated version of the proposal you can find particular warnings that can be worked with locally.

> An important design principle behind most warnings is that they are easily suppressed.

This remark made me add a list of pragmas for which there is a local suppression syntaxin in addition to the black and whitelist. Wheather it's worth it to make pragma for them is an open question. I think it is, at least because such pragma demonstrates suppression more clearly than the underscore.

Pluralia:
@scott-fleischman Thanks for nice examples!

Pluralia:
> Another example from my actual production code is unused top-level bindings warnings from records which are only declared to enable autogenerated Aeson instances.

@tejon Can I ask you to give a code example?

Pluralia:
> While the choice of the name WARN is clearly perfect, my concern is that IGNORE and ERROR sound too generic, non-descriptive, and their names do not really relay the purpose.

@kirelagin Maybe you're right. `WARN`, `IGNORE` were chosen from [the remark](https://github.com/ghc-proposals/ghc-proposals/pull/234#issuecomment-495977461) by @nomeata. `ERROR` is continuation such naming. I like these names because they seem to be easier to read. They have the warning names as arguments and it turns out to be the following: “warn about name-shadowing”, “ignore name-shadowing”, “name-shadowing error”.

goldfirere:
I find the blacklist/whitelist very helpful. (Though it would be great to define these terms -- as in you wish to allow this new pragma to apply to the whitelist but not the blacklist.)

However, as I go through, I find myself with many questions:

* How do we make the pragma scope over a top-level declaration? For example:

```
{-# IGNORE missing-signatures #-}
f True = False
f False = True
```

How does that parse? Using the normal layout rule, the `IGNORE` is a declaration on its own -- nothing connects it to what follows. This is clearly not what is meant. If we perhaps tweak the layout rule to not insert a `;` after an `IGNORE` (or similar), then the `IGNORE` applies only to the first equation, begging the question of what this means:

```
{-# IGNORE missing-signatures #-}
f True = False
{-# WARN missing-signatures #-}
f False = True
```

With custom pragmas for each suppressable warning, we get a more sensible

```
{-# MISSING_SIGNATURE f #-}
f False = True
f True = False
```

* Why blacklist `dodgy` things? When an upstream package changes an export list, you can get a dodgy import -- and suppressing a warning about this seems to fit one of the motivations.

* Why is `unrecognized-pragmas` blacklisted? That seems useful if you want code to work over multiple GHC versions.

* As I've argued elsewhere, I think disabling `warnings-deprecations` is often not what you want. Instead, you want to undeprecate/unwarn a particular entity, not a particular region of code.

* The pattern-match warnings would often want to apply to a top-level pattern-match in a function declaration. How will this parse correctly?

* How will the `inaccessible-code` one work? Presumably, you would want to put it in the inaccessible code. But the warning is actually generated on the pattern-match. Yet if you put it on the pattern-match, then you might be suppressing a warning for inaccessible code in an unexpected branch of the pattern-match -- it's another case of using too big a hammer.

* For `missing-fields`, you often have a particular record construction that you know is missing fields. Yet if you suppress the warning over an expression, then you might miss an incomplete nested record construction. Much better would be to have `{-# MISSING_FIELDS #-}` right in the `{ ... }` of the constructor you know is missing some fields.

I could continue along these lines. Looking at these individual warnings makes me feel even stronger that we're better off with a case-by-case treatment than a general one. (Perhaps best is something in the middle: group warnings into logical clumps and then treat each clump uniformly.)

AntC2:
> @goldfirere * How do we make the pragma scope over a top-level declaration? For example: ...

That was answered [here](https://github.com/ghc-proposals/ghc-proposals/pull/234#issuecomment-497166013) in response to same q from me. But I'm not convinced by the answer. Suppose

    {-# IGNORE missing-signatures #-}
    f True = False
    f False = True
    g (Just x) = ...
    g Nothing = ...

Does that `IGNORE` cover every decl following? Until what terminator? As Richard asks, then guesses

>        {-# MISSING_SIGNATURE f #-}

A toplevel name isn't applicable for every pragma, so we'd quickly get to custom pragmas with custom syntax. Then (as I said before) I don't think we have a generic approach here. Then there's going to be lots of ad-hoc syntax/code inside the compiler, adding to the cost of maintenance. And each pragma or clump of pragmas would need justifying case-by-case.

Those who are comparing to Java or Rust: is there a lot of ad-hockery inside those implementations? Or is there a generic approach we could learn from?

nomeata:
> Personally I would go with `WARN`, `NOWARN`, `WERROR`

yes please. I wrote`IGNORE` without spending time to think of a good name, but it felt bad already then. These three are much nicer.

scott-fleischman:
Most of the warnings I am interested in suppressing have a name attached to them that I could refer to when suppressing them. I am less interested in suppressing a warning for a region of code.

* If I am importing a deprecated entity or an orphan, I could name the entities I want to disable the deprecation warning and/or orphan. Or possibly even in the import statement.
* If I declare an orphan, I can name what instance I want to ignore.
* If a top binding is unused or is lacking a type signature, I can name which one
* If there is redundant constraint, I could either name the entity that it is attached to and/or the entity and the constraint itself, possibly even marking the constraint itself as redundant.

These warnings could be one of the "logical clumps" of warnings that could be dealt with similarly. They are the more common ones I have seen suppressed in our codebase.

Pluralia:
@goldfirere @AntC2 

> How do we make the pragma scope over a top-level declaration?

I propose to solve this problem like

```
{-# IGNORE "f" missing-signatures #-}
f True = False
f False = True
```

This way can help to avoid increasing of "custom pragmas with custom syntax" and adds anything to the cost of maintenance :)

Pluralia:
@goldfirere @AntC2 

> How do we make the pragma scope over a top-level declaration?

I propose to solve this problem like

```
{-# IGNORE "f" missing-signatures #-}
f True = False
f False = True
```

This way can help to avoid increasing of "custom pragmas with custom syntax" and adds anything to the cost of maintenance :)

scott-fleischman:
I'm curious what the state of this PR is. I would like to see more control over warnings so they can be applied to more specific locations rather than across an entire module, similar in spirit to the recently accepted proposal for [ambiguous types](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0232-AmbiguousType-pragma.rst).

I wonder if this PR as-is is to broad in scope? 

I think name shadowing by itself could be its own proposal, since it is a pretty delicate subject. There are reasons why one might want to enable shadowing warnings for larger blocks of code, and there are reasons why one might want to disable it for specific bindings or for specific sections. Having a separate proposal that outlines motivations for both cases would probably help create a proposal that is more universally acceptable.

On the other hand something like more control over deprecations doesn't seem as complicated. If it were it's own proposal it potentially could be accepted more easily since there aren't as many situations to consider. Likewise there may be a few more warnings that are less involved, and perhaps having a separate proposal for the simpler ones would help move the process along.

goldfirere:
There has been much conversation here, but I think this has been a healthy debate of the design space. What are the next steps here? @Pluralia ? Is it ready for committee consideration?

Pluralia:
> I wonder if this PR as-is is too broad in scope?
@scott-fleischman Maybe you're right and I like your argumentation. The thing which is stopped me to split this proposal to a few proposals is splitting implementation logic. In any cases I think about splitting one more time.

Pluralia:
@goldfirere I agree on the need for а more detailed design space. The following steps will be taken:
1) Some notes for every warning
2) Grouping warnings by <something>
3) Open a new proposal / proposals

Unfortunately, now I haven't enough time due to studying for finishing such big job. So I hope nobody will take this topic and I will be able to finish it later. As for this PR - I close it

goldfirere:
>  I hope nobody will take this topic

With all respect, I don't think this is very helpful for us. It's perfectly understandable that you don't have the time to continue this now. But as it stands, it seems like you're asking for an unbounded wait until you have time to resume, thus preventing forward progress in the meantime. While I do understand your desire to see this through to the end, I think that desire needs to be balanced with promoting the possibility of having others build on your work.

Thanks for your welcome efforts thus far on this!

Pluralia:
Sorry, I wasn't going to block the work until I could continue myself. I meant, I'd like to resume work before someone else takes up it for <natural> reasons.

int-index:
Use case: https://twitter.com/ndm_haskell/status/1234555185420263426

ndmitchell:
@int-index: Note that my complaint is not really asking for this feature, although it would possible address my concerns. What I'd really likeis for a warning that can't be determined whether it's applicable or not to default to not showing anything, rather than warning me. For example, I don't expect a warning about "unused import" to read "I ran Z3 and after 17 iterations I had no idea if your import was used or not, so instead I'm going to give a warning anyway you can't remove". Instead it says (quite reasonably) "I'm not sure about orphan instances, shove a () to hide it". In all other cases, if the warning isn't sure, it says nothing.

PR #235, opened 2019-05-26
-XTopLevelKindSignatures

int-index:
We already have `-XDefaultSignatures`, `-XKindSignatures`, and `-XPartialTypeSignatures`. Let's be consistent and have a proper name for top-level kind signatures.


nomeata:
This is a minor tweak to an accepted proposal, and I don’t think it needs the full process. @goldfirere, as the author of that, I am happy to merge that with your approval (or you can just press the button yourself).

goldfirere:
I have pushed the button.

PR #236, opened 2019-06-01
Grand Class Unification

int-index:
This is a "big picture" proposal, prompted by several other threads on the GHC Proposals platform. It is not intended to be submitted to the committee in its current form; rather, it outlines a general direction of travel for other, less ambitious proposals, to take into account.

The proposed features are:

* top-level signatures for class methods
* remove the `type` prefix of top-level kind signatures
* visible dependent quantification in terms
* visible quantification of ambiguous class variables
* constrained type families #177

### DIFF
--- a/proposals/0000-grand-class-unification.rst
+++ b/proposals/0000-grand-class-unification.rst
@@ -1,0 +1,160 @@
+The Grand Plan for Unification of Class Methods and Associated Types
+====================================================================
+
+.. proposal-number::
+.. ticket-url::
+.. implemented::
+.. highlight::
+.. header::
+.. sectnum::
+.. contents::
+
+NB. This is a "big picture" proposal, prompted by several other threads on the
+GHC Proposals platform. It is not intended to be submitted to the committee in
+its current form; rather, it outlines a general direction of travel for other,
+less ambitious proposals, to take into account.
+
+The proposed features are:
+
+* top-level signatures for class methods
+* remove the ``type`` prefix of top-level kind signatures
+* visible dependent quantification in terms
+* visible quantification of ambiguous class variables
+* constrained type families
+
+In more detail:
+
+* Add top-level signatures for class methods that serve as machine-checked
+  documentation::
+
+    class Eq a where
+      eq :: a -> a -> a
+
+    eq :: forall t. Eq t => t -> t -> t
+
+  The rule for checking the top-level signature is as follows. Let us refer to
+  the in-class method as ``$eq``, then we assume the following definition::
+
+    eq = $eq
+
+  This allows us to reorder and rename invisible type variables, and to
+  use a subclass instead of the actual parent class.
+
+* Remove the ``type`` prefix of top-level kind signatures::
+
+    Functor :: (Type -> Type) -> Constraint
+    class Functor f where
+      fmap :: (a -> b) -> f a -> f b
+
+  This unifies TLKS and regular, term-level signatures. In case of namespace
+  ambiguity, reject::
+
+    data T = T
+    T :: ...   -- not allowed.
+
+* Allow visible dependent quantification in terms::
+
+    idv :: forall a -> a -> a
+    idv t x = x :: t
+
+    x = idv Double  42     -- x :: Double
+    n = idv Integer 42     -- n :: Integer
+
+  The difference with ``-XTypeApplications`` is in the presence of a visibility
+  override ``@`` at use sites.  This implies a unification of the type and
+  expression parsers, and a change to name lookup rules.
+
+* Quantify class variables visibly when ambiguous::
+
+    class C a where
+      m :: ty
+
+    m :: forall a. C a => ty    -- if  ty  determines  a  (ambiguity check)
+    m :: forall a -> C a => ty  -- if  ty  doesn't determine  a
+
+  This removes the need for ``-XAllowAmbiguousTypes``, which are to be deprecated.
+  Note that the programmer may give a top-level signature to the methods, so the
+  readers of the code need not figure out which class variables are ambiguous.
+
+* Constrain the domain of associated types by the class::
+
+    class Container c where
+      type Item c
+      toList :: c -> [Item c]
+
+    Item :: forall c -> Container c => Type
+
+  Note that we use the same "quantify visibly if ambiguous" rule.
+
+Examples
+--------
+
+*Example 1*, ``Storable``. Definition site::
+
+  class Storable a where
+    sizeOf :: Int
+    alignmentOf :: Int
+
+  sizeOf, aligmentOf :: forall a -> Storable a => Int
+
+Use site::
+
+  ghci> sizeOf Int
+  8
+  ghci> sizeOf Bool
+  4
+
+*Example 2*, tagged accessor class ``HasLens``. Definition site::
+
+  class HasLens tag s a | tag s -> a where
+    lensOf :: Lens' s a
+
+  lensOf :: forall tag -> forall s a. HasLens tag s a => Lens' s a
+
+Use site::
+
+  ghci> struct ^. lensOf UserInfo . lensOf UserName
+  "Jack Sparrow"
+
+*Example 3*, ``MonadReader`` in the style of ``monads-tf``. Definition site::
+
+  class MonadReader m where
+    type Env :: Type
+    ask :: m (Env m)
+
+  type Env :: forall m -> MonadReader m => Type
+  ask :: forall m. MonadReader m => m (Env m)
+
+Use site::
+
+  doStuff :: Env AppM -> AppM r -> IO r
+
+
+Motivation
+----------
+
+TBD.
+
+
+Proposed Change Specification
+-----------------------------
+
+TBD.
+
+
+Effect and Interactions
+-----------------------
+
+TBD.
+
+
+Costs and Drawbacks
+-------------------
+
+TBD.
+
+
+Unresolved Questions
+--------------------
+
+TBD.


int-index:
The reason I put together this plan is that there's delicate interplay between all of the proposed features. We may make suboptimal decisions if we consider them in isolation.

proposals/0000-grand-class-unification.rst
```diff
@@ -0,0 +1,167 @@
...
+
+* top-level signatures for class methods
+* remove the ``type`` prefix of top-level kind signatures
+* visible dependent quantification in terms
+* visible quantification of ambiguous class variables
+* constrained type families
+
+In more detail:
+
+* Add top-level signatures for class methods that serve as machine-checked
```
RyanGlScott: There has been at least one proposal about this idea previously, correct? If so, mention it, and say if your vision corresponds exactly to what is put forth in the proposal, or if your vision differs on some points.

proposals/0000-grand-class-unification.rst
```diff
@@ -0,0 +1,167 @@
...
+
+  The rule for checking the top-level signature is as follows. Let us refer to
+  the in-class method as ``$eq``, then we assume the following definition::
+
+    eq = $eq
+
+  This allows us to reorder and rename invisible type variables, and to
+  use a subclass instead of the actual parent class.
+
+* Remove the ``type`` prefix of top-level kind signatures::
```
RyanGlScott: Why is it important to remove the `type` prefix? There must be some reason why you felt this warranted inclusion here.

I think your motivation here might be in reference to another, previous proposal, but others might not realize that. Better to be explicit about it.

proposals/0000-grand-class-unification.rst
```diff
@@ -0,0 +1,167 @@
...
+    class Functor f where
+      fmap :: (a -> b) -> f a -> f b
+
+  This unifies TLKS and regular, term-level signatures. In case of namespace
+  ambiguity, reject::
+
+    data T = T
+    T :: ...   -- not allowed.
+
+* Allow visible dependent quantification in terms::
```
RyanGlScott: Again, I think there's at least one previous GHC proposal that has discussed allowing visible dependent quantification syntax in terms, so mention that.

proposals/0000-grand-class-unification.rst
```diff
@@ -0,0 +1,167 @@
...
+      m :: ty
+
+    m :: forall a. C a => ty    -- if  ty  determines  a  (ambiguity check)
+    m :: forall a -> C a => ty  -- if  ty  doesn't determine  a
+
+  This removes the need for ``-XAllowAmbiguousTypes``, which are to be deprecated.
+  Note that the programmer may give a top-level signature to the methods, so the
+  readers of the code need not figure out which class variables are ambiguous.
+
+* Constrain the domain of associated types by the class::
```
RyanGlScott: Constrained type families... hasn't there been another proposal about this? Mention it.

proposals/0000-grand-class-unification.rst
```diff
@@ -0,0 +1,167 @@
...
+    idv t x = x :: t
+
+    x = idv Double  42     -- x :: Double
+    n = idv Integer 42     -- n :: Integer
+
+  The difference with ``-XTypeApplications`` is in the presence of a visibility
+  override ``@`` at use sites.  This implies a unification of the type and
+  expression parsers, and a change to name lookup rules.
+
+* Quantify class variables visibly when ambiguous::
```
RyanGlScott: As far as I can tell, this is a completely new idea. As a result, I think this deserves more justification for why it is the right thing to do. Personally, I'm somewhat suspicious of this idea, since it's very easy to write Haskell98 class declarations that are ambiguous, and having those declarations imply visible dependent quantification feels like a step too far (at least, without some sort of language extension to guard it). But perhaps I'm simply not seeing all of the details of the design you imagine for this feature.

proposals/0000-grand-class-unification.rst
```diff
@@ -0,0 +1,167 @@
...
+    ghci> struct ^. lensOf UserInfo . lensOf UserName
+    "Jack Sparrow"
+
+* **Example 3**, ``MonadReader`` in the style of ``monads-tf``. Definition site::
+
+    class MonadReader m where
+      type Env :: Type
+      ask :: m (Env m)
+
+    type Env :: forall m -> MonadReader m => Type
```
RyanGlScott: Wait, `type Env`? Didn't you say earlier that we should drop the `type` prefix in top-level kind signatures?

proposals/0000-grand-class-unification.rst
int-index: Is there? I couldn't find it.

proposals/0000-grand-class-unification.rst
int-index: > it's very easy to write Haskell98 class declarations that are ambiguous, and having those declarations imply visible dependent quantification feels like a step too far 

Haskell 98 and Haskell 2010 reject these ambiguous declarations. We would fail with an error about ambiguous type variables unless either `-XVDQ` is enabled.

proposals/0000-grand-class-unification.rst
RyanGlScott: There's #102, and possibly others.

proposals/0000-grand-class-unification.rst
RyanGlScott: > unless either `-XVDQ` is enabled.

Ah, I had no idea you intended to have a dedicated language extension for this! That's a very important detail to mention.

int-index:
> it would be helpful to state explicitly what the common thread of logic behind each of these ideas is

Indeed. I added have extended the text with the reasoning:

> The main observation, and premise of this plan, is that with dependent types (presumably, with `-XDependentHaskell`), the notion of associated types should coincide with class methods, just like closed type families coincide with functions. The stated goal is to design a path toward their convergence.
>
> We propose to add features that allow us to give equal treatment to associated types and class methods. The reasoning goes like this:
>
> ...

AntC2:
The silent syntax for VDQ in terms really worries me (as a reader). If I (or indeed the compiler) see

    x = f Foo bar

I don't know if `Foo` is a data constructor or a type name. There might be both in scope. Suppose I (or the compiler) haven't yet come across a declaration for `f` (maybe it's imported, maybe it's a parameter from an enclosing function). At least with explicit `@` I know 'here comes a type' within the term.

I then come across

    y = f baz bar

Could `baz` be a type variable? (If not, why not? How does referential transparency go for tyvars?) Could there be both a term-level and type-level var `baz` in scope? Does it need some Kind-directed name resolution?

Thank you for attempting to present a big picture. I've been feeling for some time with the drip, drip of fancy type/kind proposals that whatever language it is that GHC is moving to accept, it has no more right to call itself Haskell than many other Haskell-inspired languages. Oh, I suppose it's still Haskell because its record system continues to be an embarrassment.

int-index:
@AntC2 I completely share your concern for the readers of VDQ-code. Hence the `-XDefaultNamespace` extension. To answer your questions:

> Could `baz` be a type variable?

Yes.

> Could there be both a term-level and type-level var `baz` in scope?

* Under `-XVDQ` alone: yes. We pick the one appropriate for the current context (using today's rules to decide whether we are in a type-level or a term-level context). For example:

  ```
  data T = Single | Double
  f :: forall (a :: Type) -> ...

  ghci> f Double
  ```
  Here, we use the term-level `Double` and the user faces an error that `Double :: T`, but `Double :: Type` was expected.

* Under `-XVDQ` and `-XDefaultNamespace`: no. If there is `baz` in both namespaces, its occurrence is an error. The user must use namespace-qualified imports to resolve this.

> Does it need some Kind-directed name resolution?

This is a possibility, but I don't think it's a good idea. Postponing namespace resolution to the typechecking phase would be a complex, fragile design. Therefore, this plan offers an alternative solution (`-XDefaultNamespace`).

> whatever language it is that GHC is moving to accept, it has no more right to call itself Haskell than many other Haskell-inspired languages

Neither `-XTypeFamilies` nor `-XAllowAmbiguousTypes` are part of the Haskell 2010 report, therefore this plan does not affect the "Haskell" part of language GHC accepts. Every feature listed here is guarded behind a language extension (relative to `-XHaskell2010`).

In any case, I think this change is a move in the right direction. My goal here is to consolidate the features that GHC has, to provide a single design for things that are currently distinct (associated types and class methods). We should end up with a simpler language in the end, than the plethora of ad hoc extensions we have right now.

int-index:
> Edit: Ah, you've added this after I originally wrote: ...
> So I can't pun the names of the type constructor/data constructor?
> ```
> data Quux a = Quux a
> ```
> That's a very common idiom for types with a single constructor. Hmm, that's a -1 from me.

Indeed, it is. And there's plenty of code using it. That's why the change is guarded behind a flag, `-XDefaultNamespace`.

You can continue using this idiom, currently I often do it in my code as well. I just don't think it's possible to reconcile it with full-blown dependent types, where terms and types may be mixed arbitrarily.

I think users who need a lot of dependently typed features will end up using `-XDefaultNamespace` and just live without `data Quux a = Quux a`. It's ok to write `data Quux a = MkQuux a`.

Users who prefer classic Haskell2010 will continue their code with punning.

AntC2:
> @int-index We should end up with a simpler language in the end, than the plethora of ad hoc extensions we have right now.

Consolidating the drip, drip/plethora into one big extension to rule them all doesn't simplify anything. We'll only get a simpler language (that looks quite a bit unlike Haskell) if we can take out a bunch of syntax (and associated extensions). So what will `DependentHaskell` make redundant, that can eventually get deprecated? That's the big picture I can't see.

int-index:
> So what will `DependentHaskell` make redundant, that can eventually get deprecated?

Dependent Haskell will make `-XTypeFamilies` redundant. For example, these definitions from `base`:

```
type family Not a = res | res -> a where
  Not 'False = 'True
  Not 'True  = 'False

class IsList l where
  type Item l
  fromList  :: [Item l] -> l
  toList :: l -> [Item l]
```

will become

```
Not :: foreach (a :: Bool) -> Bool  |  not a -> a
Not False = True
Not True = False

class IsList l where
  Item :: Type
  fromList  :: [Item l] -> l
  toList :: l -> [Item l]
```

`Not` and `Item` are just functions at this point (with capitalized names, but this is inconsequential). That's because closed type families correspond to functions, associated type families correspond to class methods.

We're back in the land of functions and classes.

The idea behind this plan is to prepare ground for this transition.

cartazio:
one funny knock on consequence is certain functions would have "checked" invertibility, but that will only be visisble when used at type checking right? I couldn't reflect that into some magic "nnot = invert not" could i?

int-index:
@cartazio Even at the type level, there's no `Invert` magic. That is, given

```
data Color = R | G | B
type family F (c :: Color) = r | r -> c where
  F R = G
  F G = B
  F B = R
```

there isn't a mechanism to automatically produce an inverse of `F`:

```
type family G (c :: Color) = r | r -> c where
  G = INVERSE(F)         -- not possible
```

Likewise, I expect no such feature on the term level.

Functional dependencies and type family dependencies are only used to guide type inference. There is no term inference at the moment, so this annotation would have no effect at the term level.

cartazio:
are you sure there isn't some way to flip an injective type family using Functional dependencies and a type class?
something like (and i'm sure this will subtly not work)

```haskell
-- have MPTC and FunDeps enabled for this
class (output ~ f input ) => Related f input out |  f input -> out, f out -> input where
   forward :: Proxy f -> Proxy input -> Proxy output
   forward _ _ = Proxy 

  backward :: Proxy f -> Proxy output -> Proxy input 
  backward _ _ = Proxy 

--- lets assume Not as an injective type family
instance Related Not a b 


```

int-index:
> `instance Related Not a b`

Instances for type families are not possible.

cartazio:
good point! but could the be for unary injective type families? :)

int-index:
> good point! but could they be for unary injective type families?

No, you'd have to check functions for equality.

```
type family Id a = r | r -> a where
  Id a = a

type family Id' a = r | r -> a where
  Id' a = a

instance C Id
instance C Id'  -- is this a duplicate instance?
```

cartazio:
we could just do "are the names of the type familes the same or not" as a conservative equality relation

int-index:
> we could just do "are the names of the type familes the same or not" as a conservative equality relation

Does this mean I could observe the difference between `Id` and `Id'` at runtime?

```
class S a where
  s :: Proxy a -> String

instance S Id where
  s _ = "Id"

instance S Id' where
  s _ = "Id'"

main = print (s (Proxy @Id) == s (Proxy @Id'))
```

Does this print `False`? That's definitely not something I could support. And to be honest, I don't see what's the point of this conversation – if you want type family dependencies to produce inverses, that's a feature we can discuss. There's no need to get into the murky waters of function equality for this.

For example, we could give names to functional dependencies:

```
type family F a = r | G r = a where
  F True = 1
  F False = 0
```

This would generate

```
type family G r = a where
  G 1 = True
  G 0 = False
```

However, this is completely orthogonal to the topic of this proposal, which is to unify associated types and class methods. I created #237 to give this spin-off conversation a venue.

AntC2:
>  It is an error to use the same name for a type-level entity and a term-level entity.
>> ... pun the names of the type constructor/data constructor

We're still kicking ideas around at this stage, right? Then is there a strong reason we can't use prefix `'` to distinguish namespace?

    x = f Foo bar             -- look in data namespace for Foo, then in the type namespace
    y = f 'Foo bar            -- look in the type namespace, then error

    z = f 'baz bar            -- must be type variable baz
    u = f ''Type bar          -- double-' looks in the kind namespace

I appreciate those could be confused lexically with ill-formed `Char` literals, so it would have to be guarded behind an extension. (Or with an ill-formed `'` embedded or suffixed in a name -- but that already applies with `Char` literals. Then there must be a preceding whitespace/closing `$idchar`, ref #229 .)

Ericson2314:
@AntC2 I sort of think that type of discussion should go in the specific proposal on that issue.

@int-index 
Perhaps to denormalize the proposals to guide the discussion, this proposal could be abstract to the exact features proposed? I'm thinking e.g. 
 - "the syntax for VDQ function application will be such that type and term arguments can be parsed, name resolved, and disambiguated before the type checking of the applied function.
 - "The type langauge should be a syntact language of the term language so only namespace resolution is different."
The specific example could be kept for clarity, made non-normative just to illustrate one way to achieve those abstract goals.

Maybe my idea is bad because it will make a confusing document that's hardly easier to follow than the ideas split among individual proposals, but just I'd through it out there.

AntC2:
> @Ericson2314 I sort of think that type of discussion should go in the specific proposal on that issue.

Have just done, but

a) there is no specific proposal for usage sites of VDQ. Which raises the question where did @int-index get that usage syntax? And has anybody agreed to it?
b) the PR #102 linked from here is re definition/quantifying sites only AFAICT.
c) PR #102 was another outline/big picture bunch of stuff, now closed. "The committee has decided to table this proposal until there is a more crying need. "

[That's the U.S. sense of "table", as in take off the table. As opposed to the British sense: put on the table.]

int-index:
> We're still kicking ideas around at this stage, right?

Absolutely.

> Then is there a strong reason we can't use prefix `'` to distinguish namespace?

@AntC2 I see two problems with this.

First, `'` and `''` is the syntax TemplateHaskell uses for name quotations. You've probably seen code that looks like this, `makeLenses ''Foo`, here `''Foo :: Name`, a name looked up in the type namespace.

Second, it may lead to noisy code with a lot of namespace specifiers like this, e.g. `f (''Foo ''Bar ''Baz)` instead of `f (Foo Bar Baz)`, I'd rather motivate people to avoid punning to achieve cleaner code in the end.

These aren't "strong" reasons as you ask, but they are annoying enough that I prefer `-XDefaultNamespace`.

> there is no specific proposal for usage sites of VDQ. Which raises the question where did @int-index get that usage syntax? And has anybody agreed to it?

Well, there's basically _no_ syntax proposed for VDQ here, it's just mixing terms and types without explicit syntax. There's no special use site syntax for VDQ in types. I believe @goldfirere has the same plan for VDQ syntax in terms, for example he mentioned it in the motivation section of #214:

> The recently-implemented visible dependent quantification proposal allows types to take required, dependent variables:
>
> ```
> data ProxyVis :: forall k -> k -> Type where ...
> ```
>
> To use `ProxyVis`, we must apply `ProxyVis` to a kind and then a type, thus: `ProxyVis Bool True`.
>
> Today's theory of Haskell (that is, without dependent types) is strong enough to allow this syntax in terms, except that the namespaces are all wrong. For example, we might imagine
>
> ```
> data Proxy :: forall k. k -> Type where
>   MkProxy :: forall a -> Proxy a
> ```
>
> Note the type of `MkProxy` takes a visible, dependent argument. With this, we could write `MkProxy True` or `MkProxy Int` instead of today's `MkProxy :: Proxy True` and `MkProxy :: Proxy Int` or `MkProxy @True` and `MkProxy @Int`. After all, what's the point of a proxy if we don't specify the choice of type?

I think we should tackle this design challenge now because the exact same problem would arise with `foreach`. And I don't want to end up with noisy syntax for `foreach`, it's the quantifier that Dependent Haskell is about.

I could be convinced otherwise, but for now I think `-XDefaultNamespace` is the optimal compromise. Sacrifice some punning for readable use sites.

----

> Perhaps to denormalize the proposals to guide the discussion, this proposal could be abstract to the exact features proposed? I'm thinking e.g.
> * "the syntax for VDQ function application will be such that type and term arguments can be parsed, name resolved, and disambiguated before the type checking of the applied function.
> * "The type langauge should be a syntact language of the term language so only namespace resolution is different."

@Ericson2314 I agree with both bullets. In fact, the reason I created this plan is to make sure everyone is on the same page, so I shall put these clarifications into the document.

potato4444:
The `-XVDQ` seems to be doing two things here:
 - Allowing visible dependent quantification.
 - Using visible dependent for ambiguous type variables in classes.

I support the first of these features, but am less sure of the second. At least I think they are different enough that they should be enabled separately. The behaviour for classes should be it's own separate pragma or extension.

int-index:
> At least I think they are different enough that they should be enabled separately.

@potato4444 You're right. I split it into two extensions, `-XVDQ` and `-XVisibleClassVariables`.

AntC2:
Re punning of names in terms and types, and the reason GHC is already using `'` prefix ... I couldn't see how `DataKinds` fit with `-XVDQ` and `-XDefaultNameSpace` (apologies if I'm being dumb)

    {-# LANGUAGE  DataKinds, VDQ  #-}

    data Bool = False | True

    x = f False                           -- term or type?
    y = f 'False                          -- definitely type
    z = f (False :: Bool)                 -- ? type/kind annotation no help

int-index:
@AntC2 The treatment of `'` (and a few other syntactic constructs) is a delicate matter. I planned to tackle this question in the dedicated proposals for `-XVDQ` and `-XDefaultNamespace`.

For now, I imagine the following arrangement:

* Under `-XVDQ` alone, both `'` and `''` continue to have their current meaning. That is, `'` in types is a namespace specifiers, `'` and `''` in terms are TemplateHaskell name quotation syntax.

* Under `-XDefaultNamespace`, `'` is TemplateHaskell name quotation syntax both in types and in terms, `''` is deprecated.


So, with your example:

```
{-# LANGUAGE  DataKinds, VDQ  #-}

class False   -- a type-level False to spice things up.

data Bool =
    False    -- term-level False
  | True     -- term-level True

x = f False                           -- term-level False
y = f 'False                          -- TemplateHaskell name
z = f (False :: Bool)                 -- term-level False
```

There is no way to refer to the type-level `False` in this arrangement. This is the reason to introduce `-XDefaultNamespace`.

glaebhoerl:
How would `-XDefaultNamespace` deal with type/data namespace punning in the builtin syntax? (`(a,b)`, `[a]`, I'm not sure if there are others.)

int-index:
> How would `-XDefaultNamespace` deal with type/data namespace punning in the builtin syntax? 

This is to be specified in the proposal for `-XDefaultNamespace`. The design I am leaning towards at the moment is that built-in syntax would be defaulted to the data namespace, for example `[a]` is a singleton list, not the type of lists. An additional module, `BuiltinTypes`, would be explicitly imported alongside `Prelude`, with `Unit :: Type`, `List :: Type -> Type`, and `Tuple :: List Type -> Type`, such that:

```
() :: Unit
[1,2] :: List Integer
(1,"hello") :: Tuple [Integer, String]
```

goldfirere:
Unsurprisingly, I'm broadly in support, but with some specific thoughts:

* Despite retracting my proposal (#148) on the issue, I'm in support of top-level signatures. But my sense was that others wanted a different syntax, and then it all got mired in debate, so I gave up. I don't feel strongly.

* I can't see `-XVDQ` without thinking that the next proposal will suggest `-XKCD`. `-XTypeArguments` maybe? That's a bit similar to `-XTypeApplications`, I suppose.

* I like `-XDefaultNamespace`. It's like my #214 but giving the user the choice (and obligation) of specifying how to disambiguate. Without some more structure in the module system, though, it may prove painful to have lots of qualified imports everywhere. I may have other thoughts once we have the details written down.

* I still dislike `-XVisibleClassVariables`, and I don't see the need for it.

* I don't think we need `-XMatchableQuantifiers`. Instead, `:->` could just be imported.

I worry about blowing our complexity budget with extensions here. It might somehow be simpler just to have an extension `-XDependentTypes` (NB: not `-XDependentHaskell`, which is redundant) that evolves over several releases.

I don't think it's unfair to call it a new language, given the amount of churn here. The new language might well still be called Haskell, but someone who stepped away for a few years will have a hard time recognizing it, and we're at the risk of outdating a *lot* of written material. I'm certainly not saying these are blockers, but that I'm more sympathetic to the viewpoint that this is a new language than many folks would probably think.

Thanks, @int-index for writing this up -- it's very helpful to have this out there.

int-index:
> my sense was that others wanted a different syntax, and then it all got mired in debate, so I gave up

The version I am proposing works in addition to the normal signatures, so the top-level signature is always redundant. That's a bit different from the current iteration of #148, in a way other than syntax.

> without thinking that the next proposal will suggest `-XKCD`

That's reserved for "Kind-Checking Determinism", which must be disabled to promote `IO`.

>  it may prove painful to have lots of qualified imports everywhere

Perhaps. Yes, I do expect it to result in quite a bit of qualified imports, but that does not sound catastrophic to me. Qualified imports are nice.

> I still dislike `-XVisibleClassVariables`, and I don't see the need for it.

There's a reason I like it. Here's a question:

* How does the user choose between visible and invisible quantification?

My answer:

* Visible quantification is needed when the variable would be otherwise ambiguous.

And that's precisely the rule we're implementing here. That's GHC producing a good API for class methods out of the box. And no ambiguous types as a bonus.

You say it's not necessary, but then what would be a way to quantify class variables visibly (in class methods)? Ah, perhaps you could bring up the current iteration of #148 to answer this, but then we're dealing with those spicy scoping rules:

```
class C a b where
  toplevel m :: forall a b -> C a b => Int
```

Here, `forall a b` binds the variables, but `C a b` in the header mentions them. Outside the `forall`! Ouch.

I understand that it'd work, but `-XVisibleClassVariables` seems like a much more ergonomic choice to me.

> I don't think we need `-XMatchableQuantifiers`. Instead, `:->` could just be imported.

I'm hazy on the details here, but I thought `-XMatchableQuantifiers` is needed to change the meaning of `:->` rather than to bring `:->` into scope. That is:

* With `-XMatchableQuantifiers`: `->` is unmatchable, `:->` is matchable
* Without `-XMatchableQuantifiers`: matchability of `->` is inferred by GHC like today

And we'd also need some syntax for `forall a :.` and `forall a :->`, so I imagined it would be guarded behind `-XMatchableQuantifiers` as well.

>  It might somehow be simpler just to have an extension `-XDependentTypes` <...> that evolves over several releases.

Having less knobs to tweak would be surely easier to implement, but I'm aware that not everyone is on board with evolving extensions: https://github.com/ghc-proposals/ghc-proposals/pull/146#issuecomment-398052024

I share the concerns expressed there. Unless we call this extension something like `-XExperimentalDependentTypesDontEnable`, it doesn't feel right to make it evolve in drastic ways.

> Thanks, @int-index for writing this up -- it's very helpful to have this out there.

Thanks for taking the time to review!

It seems that the main point of tension is `-XVisibleClassVariables`, but I continue to believe it is the sweet spot in the design space.

goldfirere:
Good point about `-XMatchableQuantifiers`.

Other points you've made are essentially about taste -- and I want to watch others debate to see if that evolves my thinking.

RyanGlScott:
Some scattered thoughts:

> Associated types may quantify visibly over class variables, while class methods cannot. To bridge the gap, we introduce a new "quantify visibly if ambiguous" rule, which deprecates `-XAllowAmbiguousTypes` along the way.

Surely this doesn't completely deprecate `AllowAmbiguousTypes`? There are places besides class methods where ambiguity can arise (e.g., `f :: forall a. F a -> ()`, where `F` is a non-injective type family), and I don't foresee how `VisibleClassMethods` would help there.

More broadly, while I wouldn't go as far to say that I flat-out dislike `VisibleClassMethods`, I can't say I'm excited about it either. I like the fact that `TopLevelSignatures` lets you verify that the type variables that you believe will be ambiguous will in fact be quantified visibly, but by the same token, I'm somewhat disappointed that ambiguous type variables are the _only_ ones that can be quantified visibly. I feel like there ought to be a more general mechanism that lets you make any of the class variables be quantified visibly (although I don't have any concrete suggestions to make in pursuit of achieving this goal).

> Guarded by `-XVisibleForall`

If we're going to introduce a language extension guarding the use of visible dependent quantification in terms, should we require the same language extension for its use in kinds? Note that we can already use visible dependent quantification in kinds in today's GHC (guarded only behind the preexisting extensions `RankNTypes` and `PolyKinds`).

AntC2:
> @RyanGlScott  Note that we can already use visible dependent quantification in kinds in today's GHC

Hmm? Ever since `TypeinType` kinds just are types: they occupy the same namespace. And any kind signature will appear in a place we know we're looking in the types namespace.

Whereas up 'til now Haskell has kept strong syntactic demarcation between which namespace applies for terms vs types/kinds. Doesn't this go all the way back through the Barendregt cube and the various typed lambda calculi? Then punning data constructor names with their datatype, or punning in `DataKinds` are just symptoms of something more pervasive. Something I find really valuable. Something I can see no justification for breaking.

> @goldfirere I don't think it's unfair to call it a new language, given the amount of churn here. 

Indeed. But a 'Tale of two languages' would be a terrible outcome. There's not enough hands/brains to support one Haskell. And if there were two Haskells, I know where all the brains would go to. (And I'd ask them to rip out a load of typeness-in-progress from Haskell Haskell before they left.)

Explain it again: if we want to put types in terms, what's wrong with prefixing `@`? That seems a very small lexical price to pay. And retains the two namespaces cleanly. Then we don't need the abomination `forall a -> ...`; the usual `forall a. ...` works to specify the order of tyvars for `@`pplication.

If (for example) a term has `sizeOf` not `@`pplied, that'll be rejected at the usage site as ambiguous.

Ericson2314:
@AntC2 that gets confusing with `foreach a ->` and using terms in types. It's better if we design for all combinations, and then maybe make "enforc style guides" in there is demand for that once the features have landed.

AntC2:
> @Ericson2314  that gets confusing with ... using terms in types.

Isn't that going to run into the same ambiguity? Does `Bool` appearing in a type mean the type or the promoted `DataKind`? We resolve that by prefixing `'` to mean the kind. (Then I could imagine prefixed `'True` appearing in a type to mean the promoted type.) So I'm being consistent: if you want the type in a context where it's unexpected or could be ambiguous, prefix it.

Let's consider `foreach a ->` (also looks like an abomination) after we've decided on `forall a ->`.

>  It's better if we design for all combinations, ...

That's what I'm trying to do. Whereas

a) you haven't answered my "what's wrong with prefixing `@`"

b) you're justifying one abomination to be consistent with another abomination that also hasn't been justified.

I'd be persuaded by almost any type-astronaut's abomination if it got us to Haskell having a decent records system. All the Type Families work a decade ago was promised to do that, but hasn't. We now have a Row Polymorphism proposal using Type Families, but that's not leading to a Records system either.

int-index:
> Surely this doesn't completely deprecate `AllowAmbiguousTypes`? There are places besides class methods where ambiguity can arise (e.g., `f :: forall a. F a -> ()`, where `F` is a non-injective type family), and I don't foresee how `VisibleClassMethods` would help there.

Without `-XAllowAmbiguousTypes`, that's just an error due to an ambiguous variable. The user has the option to write `f :: forall a -> F a -> ()` instead. The problem with class methods is that the choice is not up to the user.

Therefore, I do think `-XVisibleClassMethods` allows us to deprecate `-XAllowAmbiguousTypes`.

> If we're going to introduce a language extension guarding the use of visible dependent quantification in terms, should we require the same language extension for its use in kinds?

Not necessarily. The situation in types is a mess, we should be actually using `foreach` there because we don't have parametricity. Cleaning this up is a story for another day.

> I'm somewhat disappointed that ambiguous type variables are the only ones that can be quantified visibly. I feel like there ought to be a more general mechanism that lets you make any of the class variables be quantified visibly

I understand the desire for control, and at first I was unsatisfied with this idea as well. Then I thought: why make an argument visible if it can be inferred from the rest of the type? Sure, it might be nice to specify it at call site, but that's what `-XTypeApplications` are for – just defer this choice to the call site. With ambiguous type variables, there's no choice to defer – not specifying it is an error. The inherent connection between ambiguity and visibility is why I like this design.


----

@AntC2 I have to agree with @Ericson2314 here. The thing we want in the end is `foreach a ->`, it is the goal of Dependent Haskell. We'll have to solve the exact same host of challenges for it. `forall a ->` is just a smaller feature that allows us to tackle only some of the challenges, so that we have less to worry about when implementing `foreach`: the parsing and name resolution issues will be settled by then.

int-index:
@simonpj @goldfirere Thank you for the feedback about the "quantify visibly if ambiguous rule". I have thought about it for a long time now, and I agree we can do better than this. In particular, I remember Richard mentioning that we can support class method headers which would mirror the headers of associated types... and now I realize that this is exactly the missing piece of the puzzle.

I've updated the proposal a bit, although it may have become a bit more cryptic, as I haven't elaborated the idea of class method headers in sufficient detail. However, I plan to write a full proposal quite soon (as it has immediate pay off even without considering other proposed features: we get the ability to reorder class variables).

This will give us a much better backwards compatibility story, more control, and (probably) a syntactic rule to determine the full type of a class method.

int-index:
Class method headers are now proposed in #267

Ericson2314:
BTW you can cross out #54 from the OP, as it's now merged.

PR #238, opened 2019-06-06
Introduce -XTypeAbstractions, limiting -XScopedTypeVariables

goldfirere:
Recent discussion (on only [tangentially related features](https://gitlab.haskell.org/ghc/ghc/issues/16734#note_203412)) has brought to light that the accepted proposal for binding type variables does not fly. I have amended the proposal to fix the problem. Basically, we cannot have both `\ @a -> ...` and the scoping behavior of `forall`. This new version introduces a new extension `-XTypeAbstractions` that will control the new syntax, as well as disable the feature of `-XScopedTypeVariables` that brings `forall`d variables into scope.

[Original, accepted proposal](https://github.com/ghc-proposals/ghc-proposals/blob/79f48248ae31b1a490deb1b019c206efa0be89da/proposals/0155-type-lambda.rst)
[Discussion on original proposal](https://github.com/ghc-proposals/ghc-proposals/pull/155)
[Rendered new proposal](https://github.com/goldfirere/ghc-proposals/blob/type-abstractions/proposals/0050-type-lambda.rst)

### DIFF
--- a/proposals/0050-type-lambda.rst
+++ b/proposals/0050-type-lambda.rst
@@ -424,12 +424,33 @@ Effect and Interactions
   
 Costs and Drawbacks
 -------------------
-* The new small zoo of flags is yet more flags to be aware of. It is possible to do away
-  with this part of the proposal of the cost appears to outweigh the benefit. (Earlier
-  versions of this proposal had ``-XTypeAbstractions`` imply ``-XNoScopedForAlls``, but
-  that does not work with *inferred* variables. In any case, this historical fact is why
-  these two ideas are bundled in one proposal. They could be separated, but I want both
-  parts, so I've kept them both here.)
+* The new small zoo of flags is yet more flags to be aware of. It is possible
+  to do away with this part of the proposal if the cost appears to outweigh
+  the benefit. (Earlier versions of this proposal had ``-XTypeAbstractions``
+  imply ``-XNoScopedForAlls``. In any case, this historical fact is why these
+  two ideas are bundled in one proposal. They could be separated, but I want
+  both parts, so I've kept them both here.)
+
+* Note that the second part of this proposal (introducing ``-XTypeAbstractions``)
+  is *not* backward-compatible, because it rejects expressions like ::
+
+    ((\x -> (x :: a)) :: forall a. a -> a)
+
+  which are accepted today. No migration period is proposed, because it is
+  very hard to imagine how ``-XTypeAbstractions`` and ``-XScopedForAlls`` should
+  co-exist peacefully here. Instead, we can issue a specific error message telling
+  users how to migrate their code in this case.
+
+  My hope is that constructs such as this one are rare and would not impact many
+  users.
+
+  If necessary, we could imagine taking the expression ``expr :: forall ... . ty``
+  and looking proactively to see whether ``expr`` ever uses a type variable
+  pattern from this proposal. If not, ``-XScopedForAlls`` could trigger (and we
+  issue a warning with ``-Wcompat``). But, if a type argument appears anywhere
+  in ``expr``, then ``-XScopedForAlls`` is disabled. This would be backward-compatible,
+  but unfortunately non-local and annoying. I prefer just to skip this
+  migration step.
 
 * ``-XTypeAbstractions`` is another feature to specify and maintain, and
   that's always a burden. It will take some creative thought about how to do
--- a/proposals/0050-type-lambda.rst
+++ b/proposals/0050-type-lambda.rst
@@ -14,6 +14,7 @@ Binding type variables in lambda-expressions
 .. _`#128`: https://github.com/ghc-proposals/ghc-proposals/pull/128
 .. _`#119`: https://github.com/ghc-proposals/ghc-proposals/pull/119
 .. _`Haskell 2010 Report`: https://www.haskell.org/onlinereport/haskell2010/haskellch10.html#x17-18000010.5
+.. _`#285`: https://github.com/ghc-proposals/ghc-proposals/pull/285
 
 Proposal `#126`_ allows us to bind scoped type variables in patterns using an ``@a`` syntax.
 However, the new syntax is allowed only in *constructor* patterns ``K @a @b x y``. This proposal
@@ -95,13 +96,16 @@ Proposed Change Specification
 -----------------------------
 
 1. Introduce ``-XPatternSignatures``. With ``-XPatternSignatures``, we would
-   allow type signatures in patterns. These signatures could mention type
-   variables, but their scope would be limited to only that very signature.
+   allow type signatures in patterns. These signatures could mention in-scope
+   type variables as variable occurrences, but could not bind type variables.
 
-2. Introduce ``-XPatternTypeVariables``. With ``-XPatternTypeVariables``, any
-   type variables bound in a pattern signature would remain in scope over the
+2. Introduce ``-XPatternSignatureBinds``. With ``-XPatternSignatureBinds``, any
+   out-of-scope type variables written in a pattern signature would be bound there
+   and would remain in scope over the
    same region of code that term-level variables introduced in a pattern scope
-   over. This implies ``-XPatternSignatures``.
+   over. This extension is on by default (but is useless without ``-XPatternSignatures``).
+   (This point is actually not a change, but is a part of accepted proposal
+   `#285`_.)
 
 3. Introduce ``-XMethodTypeVariables``. With ``-XMethodTypeVariables``, type
    variables introduced in an instance head would scope over the bodies of
--- a/proposals/0050-type-lambda.rst
+++ b/proposals/0050-type-lambda.rst
@@ -492,8 +492,7 @@ Effect and Interactions
   concerned).
 
   Accepting the ``@(..)`` syntax does *not* entail accepting this new, separate
-  ``(..)`` syntax, though it is good to know that the idea is forward compatible
-  with.
+  ``(..)`` syntax, though it is good to know that the idea is forward compatible.
 
   A ``@(..)`` argument counts as a type argument when asking whether ``-XScopedForAlls``
   affects a function equation.
--- a/proposals/0050-type-lambda.rst
+++ b/proposals/0050-type-lambda.rst
@@ -501,6 +501,12 @@ Effect and Interactions
   lambda-expressions, or anywhere other than a function binding with a type
   signature. This is because doing so would require propagating type
   information into scoping, which is problematic.
+
+  Some have argued on GitHub that it may be best to hold off the ``@(..)`` until
+  we gain more experience here: adding new features is easier than removing them.
+  While I agree that this could be done, the ``@(..)`` construct makes for a very
+  easy migration from today's ``-XScopedTypeVariables`` and is thus tempting to
+  be around from the start. I don't feel strongly.
   
 Costs and Drawbacks
 -------------------
--- a/proposals/0050-type-lambda.rst
+++ b/proposals/0050-type-lambda.rst
@@ -394,7 +394,8 @@ Effect and Interactions
 
 * The optional extra ``@(..)`` notation seems like a convenient middle ground,
   allowing for an easy transition from the old-style ``-XScopedTypeVariables``
-  to the newer ``-XTypeAbstractions``. It brings inferred variables into
+  to the newer ``-XTypeAbstractions``. It brings the *inferred* variables (from `#99`_)
+  into
   scope, quite conveniently. This new notation also allows type variables to
   be brought into scope without the ``forall`` keyword in the type, in case
   the user does not want to trigger ``forall``\ -or-nothing behavior.
--- a/proposals/0050-type-lambda.rst
+++ b/proposals/0050-type-lambda.rst
@@ -238,7 +238,83 @@ C. Typing rules for the new construct are as in a `recent paper
        Multiple equations can bind type variables in different places,
        as we have a type signature to guide us.
 
-D. ``-XTypeAbstractions`` and ``-XScopedForAlls`` have a fraught relationship,
+#. Typing rules for pattern synonym bindings are complicated, as usual.
+
+   i. A visible type abstraction in a pattern synonym binding that lacks
+      a type signature is rejected. (While we could, at some cost, work
+      out what should happen here, please just use a type signature.)
+
+   ii. (Background information; no new specification here.)
+       Pattern synonym type signatures have a restricted form that looks
+       like this::
+
+         pattern P :: forall universal_tvs.   required_context =>
+                      forall existential_tvs. provided_context =>
+                      arg1 -> arg2 -> ... ->
+                      result
+
+       `The GHC manual <https://downloads.haskell.org/ghc/latest/docs/html/users_guide/exts/pattern_synonyms.html#typing-of-pattern-synonyms>`_ has the details for how parts
+       of this signature can be left out; I will not repeat these rules here.
+       The key observation is that all quantified type variables occur
+       *before* any required term-level arguments.
+
+       Furthermore, pattern synonym bindings may be specified in two parts,
+       for explicit bidirectional pattern synonyms::
+
+         pattern P <- pat
+           where P = expr
+
+        Call the top line the *pattern synonym pattern binding*, while
+        the second line is the *pattern synonym expression binding*.
+
+        In an implicitly bidirection pattern synonym binding, the
+        pattern synonym pattern binding and pattern synonym expression
+        binding are written with one bit of syntax. For the purposes
+        of this proposal, though, we consider type-checking this
+        bit of syntax *twice*, once as a pattern synonym pattern binding,
+        and once as a pattern synonym expression binding.
+       
+   iii. With ``-XTypeAbstractions``, a pattern synonym pattern binding may
+        include any number of type abstractions (such as ``@a`` or ``@_``)
+        directly after the pattern synonym name. (Such a binding must be written
+        in prefix notation, not infix.)
+        These bindings correspond to a prefix of the *specified* *universal* type variables
+        in the pattern synonym's type. It is an error to write more type
+        abstractions than there are specified universal variables.
+
+        Each type abstraction binds a local name to the corresponding
+        universal type variable. These names are available in the right-hand
+        side (after the ``<-`` or ``=``).
+
+        (Existentials are excluded here because an existential type variable
+        is bound by the pattern in the right-hand side. There appears to be
+        no motivation for being able to name these on the left.)
+
+        The rules for the usage of such variables on the right-hand side are
+        unchanged from the way scoped type variables work in pattern synonyms
+        today.
+
+   iv. With ``-XTypeAbstractions``, a pattern synonym expression binding
+       may include any number of type abstractions (such as ``@a`` or ``@_``)
+       directly after the pattern synonym name. (Such a binding must be written
+       in prefix notation, not infix.) These correspond to a prefix of
+       the concatentation of the specified universal and specified existential type variables
+       written in the pattern synonym type signature. It is an error
+       to write more type abstractions than there are specified universal
+       and specified existential type variables.
+
+       Each type abstraction binds a local name to the corresponding
+       universal or existential type variable. These names are available in the
+       right-hand side (after the ``=``).
+
+       (Existentials are included here because a pattern synonym used as an
+       expression takes existentials as arguments from call sites, and it is
+       sensible to bind these on the left.)
+
+       The rules for the usage of such variables on the right-hand side are
+       just as they exist for ordinary function bindings.
+       
+#. ``-XTypeAbstractions`` and ``-XScopedForAlls`` have a fraught relationship,
    as both are trying to accomplish the same goal via different means. Here are
    the rules keeping this sibling rivalry at bay:
 
@@ -249,7 +325,7 @@ D. ``-XTypeAbstractions`` and ``-XScopedForAlls`` have a fraught relationship,
 
    ii. If ``-XScopedForAlls`` is enabled,
        in an equation for a function definition for a function ``f`` (and similar
-       for pattern synonym bindings):
+       for pattern synonym pattern bindings and pattern synonym expression bindings):
 
        * If ``f`` is written with no arguments or its first argument is not
          a type argument (that is, the next token after ``f``
@@ -280,7 +356,11 @@ E. (Optional extra) If ``-XTypeAbstractions`` is in effect, then a function
 
    would have ``a`` and ``b`` in scope in the ``RHS``.
 
-   The ``@(..)`` construct works for both *specified* and *inferred* variables.
+   The ``@(..)`` construct works for both *specified* and *inferred* variables,
+   and is additionally avaialable in pattern synonym pattern bindings (where it
+   brings into scope only universals) and pattern synonym expression bindings
+   (where is brings into scope both universals and existentials). (In an implicitly
+   bidirectional pattern synonym, the ``@(..)`` brings into scope only universals.)
 
 Examples of new behavior of scoped type variables
 -------------------------------------------------
--- a/proposals/0050-type-lambda.rst
+++ b/proposals/0050-type-lambda.rst
@@ -31,8 +31,8 @@ extends this idea to lambda-expressions, allowing ``\ @a x -> ...``. Here are so
   fconst :: a -> b -> b
   fconst @_ @d _ x = (x :: d)             -- order matters
 
-  pair :: forall a. a -> (a, a)           -- brings a into scope (just like today)
-  pair @b x = (x :: a, x :: b)            -- brings b into scope with the same meaning as a
+  pair :: forall a. a -> (a, a)           -- no longer brings a into scope
+  pair @b x = (x :: b, x :: b)            -- brings b into scope
   
   higherRank :: (forall a. a -> a -> a) -> ...
   higherRank = ...
@@ -98,7 +98,7 @@ As always, we can consider a nested lambda ``\ x y z -> ...`` to be an abbreviat
 variable (preceded by ``@``). We do require, as usual, that we do not bind the same variable
 twice in a single lambda; this is true for type variables, too.
 
-Thus, the proposal boils down to one rule:
+Thus, the core of the proposal boils down to one rule:
 
 * ``\ @a -> body``, being checked against the type ``forall a. ty`` (where the ``a`` is *specified*), binds the type
   variable ``a`` and then checks ``body`` against the type ``ty``. Checking an
@@ -115,12 +115,16 @@ As usual, we can interpret a function defintion ``f <args> = body`` as
 ``f = \ <args> -> body``, and thus the function-definition case reduces to the lambda-expression
 case above.
 
-This new behavior will be with ``-XTypeApplications``. Naturally, scoped type variables
-work only with 
-``-XScopedTypeVariables`` enabled, so using this feature without ``-XScopedTypeVariables``
-would enable only ``@_`` abstractions.
+One wrinkle is that, according to the rules in the `Visible Type Application`_ paper,
+type variables quantified by a ``forall`` in a type signature get brought into scope
+*before* we check the term. We thus add the following rule:
 
-This change is specified in the appendix to the `Type variables in patterns <https://cs.brynmawr.edu/~rae/papers/2018/pat-tyvars/pat-tyvars.pdf>`_ paper.
+* The new extension ``-XTypeAbstractions`` enables binding type variables as described
+  in this proposal. Additionally, it disables the behavior of ``-XScopedTypeVariables``
+  that brings ``forall``\-quantified variables into scope in the body of the ``forall``.
+  Examples are below.
+
+This change is specified in the appendix to the `Type variables in patterns <https://cs.brynmawr.edu/~rae/papers/2018/pat-tyvars/pat-tyvars-extended.pdf>`_ paper.
 
 Bidirectional type checking
 ---------------------------
@@ -162,10 +166,41 @@ In the context of the GHC implementation, we have these definitions::
 *Checking* mode is precisely when the ``ExpType`` passed to ``tcExpr`` is a ``Check``.
 *Inference* mode is precisely when the ``ExpType`` passed to ``tcExpr`` is an ``Infer``.
   
-  
+Examples of new behavior of ``-XScopedTypeVariables``
+-----------------------------------------------------
 
-Examples
---------
+::
+
+   f :: forall a. a -> a
+   f x = (x :: a)      -- rejected with -XTypeAbstractions
+
+   g :: forall a. a -> a
+   g @a x = (x :: a)   -- accepted
+
+   h = ((\x -> (x :: a)) :: forall a. a -> a)
+     -- accepted with previous -XScopedTypeVariables, but rejected
+     -- if -XTypeAbstractions is in effect
+
+   i = ((\ @a x -> (x :: a)) :: forall a. a -> a)
+     -- accepted with -XTypeAbstractions
+
+Note that limiting the behavior of ``-XScopedTypeVariables`` is necessary if we
+think about where type variables are brought into scope. Are they brought into
+scope by the ``forall``? Or by the ``@a``? It can't be both, as there is no
+sensible desugaring into System F. Specifically, if we have ``expr :: forall a. ty``,
+that gets desugared into ``/\ a -> expr``. If we have ``(\ @a -> expr) :: forall b. ty``,
+what does it get desugared into? It would have to be ``/\ b -> /\ a -> expr``, but then
+``b`` and ``a`` are different.
+
+Here might be another way of thinking about it. Suppose we're checking ``expr`` against
+the pushed-down (known) type ``forall a. ty``. If we bring ``a`` into scope, what type
+do we check ``expr`` against? Is it ``forall a. ty`` again? That's very awkward if ``a``
+is *already* in scope. If we check ``expr`` against ``ty`` and ``expr`` looks like
+``\ @b -> expr'``, then we check ``\ @b -> expr'`` against ``ty`` -- not against
+``forall a. ty``.
+     
+Motivating Examples
+-------------------
 
 Here are two real-world examples of how this will help, courtesy of @int-index:
 
@@ -242,7 +277,7 @@ Effect and Interactions
   right before doing the subsumption check. Type generalization is hard in GHC, though,
   and so the paper avoided it. In order to implement this proposal, we'll have to work
   out how to do this.
-
+  
 Costs and Drawbacks
 -------------------
 This is another feature to specify and maintain, and that's always a burden. It will take
--- a/proposals/0050-type-lambda.rst
+++ b/proposals/0050-type-lambda.rst
@@ -163,9 +163,6 @@ Type abstractions
 Introduce ``-XTypeAbstractions``. With ``-XTypeAbstractions``, users
 could write a pattern like ``@a`` to the left of the ``=`` in a function
 binding or between the ``\`` and ``->`` in a lambda-expression. 
-In addition, the ``-XScopedForAlls`` extension now affects only *inferred*
-variables, never *specified* variables. Note that *inferred* variables
-can be mentioned in source Haskell with accepted proposal `#99`_.
 
 A. Here is the BNF of the new form (cf. The `Haskell 2010 Report`_)::
 
@@ -241,20 +238,64 @@ C. Typing rules for the new construct are as in a `recent paper
        Multiple equations can bind type variables in different places,
        as we have a type signature to guide us.
 
+D. ``-XTypeAbstractions`` and ``-XScopedForAlls`` have a fraught relationship,
+   as both are trying to accomplish the same goal via different means. Here are
+   the rules keeping this sibling rivalry at bay:
+
+   i. ``-XScopedForAlls`` does not apply in expression type signatures. Instead,
+      if users want a type variable brought into scope, they are encouraged to
+      use ``-XTypeAbstractions``. (It would not be hard to introduce a helpful
+      error message instructing users to do this.)
+
+   ii. If ``-XScopedForAlls`` is enabled,
+       in an equation for a function definition for a function ``f`` (and similar
+       for pattern synonym bindings):
+
+       * If ``f`` is written with no arguments or its first argument is not
+         a type argument (that is, the next token after ``f``
+         is not a prefix ``@``), then ``-XScopedForAlls`` is in effect and
+         brings type variables into scope.
+
+       * Otherwise, if ``f``\'s first argument is a type argument, then
+         ``-XScopedForAlls`` has no effect. No additional type variables
+         are brought into scope.
+
+E. (Optional extra) If ``-XTypeAbstractions`` is in effect, then a function
+   binding may use ``@(..)`` on its left-hand side. Here is the BNF (cf. the
+   `Haskell 2010 Report <https://www.haskell.org/onlinereport/haskell2010/haskellch4.html#x10-800004.4>`_, Section 4.4.3), recalling that braces mean "0 or more"::
+
+     funlhs  →  var apat { apat }
+             |  pat varop pat
+             |  '(' funlhs ')' apat { apat }
+             |  funlhs '@' '(' '..' ')'
+
+   The last line is new, and we assume the ``@`` is in prefix form. This construct
+   is available only when the function being defined has a type signature.
+   The new construct brings into scope all type variables brought into scope
+   at that point in the signature. Note that implicitly quantified type variables
+   are brought into scope at the top of a signature, and so ::
+
+     f :: a -> b -> a
+     f @(..) = -- RHS
+
+   would have ``a`` and ``b`` in scope in the ``RHS``.
+
+   The ``@(..)`` construct works for both *specified* and *inferred* variables.
+
 Examples of new behavior of scoped type variables
 -------------------------------------------------
 
 ::
 
    f :: forall a. a -> a
-   f x = (x :: a)      -- rejected with -XTypeAbstractions
+   f @b x = (x :: a)   -- rejected, because -XScopedForAlls is disabled here
 
    g :: forall a. a -> a
    g @a x = (x :: a)   -- accepted with -XTypeAbstractions
 
    h = ((\x -> (x :: a)) :: forall a. a -> a)
      -- accepted with previous -XScopedTypeVariables, but rejected
-     -- if -XTypeAbstractions is in effect
+     -- now
 
    i = ((\ @a x -> (x :: a)) :: forall a. a -> a)
      -- accepted with -XTypeAbstractions
@@ -330,7 +371,8 @@ Effect and Interactions
   written by the programmer using this new notation, is not available for use with
   any form of visible type application, including the one proposed here. If you have
   a function ``f :: forall {k} (a :: k). ...``, you will have to rely on the old behavior
-  of ``-XScopedTypeVariables`` to bring ``k`` into scope in ``f``\'s definition. This is
+  of ``-XScopedTypeVariables`` to bring ``k`` into scope in ``f``\'s definition, or
+  you will have to use a pattern signature. This is
   regrettable but seems an inevitable consequence of the ``{k}`` notation.
 
 .. _`#99`: https://github.com/ghc-proposals/ghc-proposals/pull/99
@@ -349,6 +391,36 @@ Effect and Interactions
   right before doing the subsumption check. Type generalization is hard in GHC, though,
   and so the paper avoided it. In order to implement this proposal, we'll have to work
   out how to do this.
+
+* The optional extra ``@(..)`` notation seems like a convenient middle ground,
+  allowing for an easy transition from the old-style ``-XScopedTypeVariables``
+  to the newer ``-XTypeAbstractions``. It brings inferred variables into
+  scope, quite conveniently. This new notation also allows type variables to
+  be brought into scope without the ``forall`` keyword in the type, in case
+  the user does not want to trigger ``forall``\ -or-nothing behavior.
+
+  Note that this notation is forward compatible with visible dependent quantification
+  in terms::
+
+    f :: foreach (count :: Int) (label :: String) (is_paid_for :: Bool) -> Invoice
+    f (..) = -- here, count, label, and is_pair_for are all in scope
+
+  This style allows for more perspicuous types while avoiding redundancy. The particular
+  example here uses ``foreach`` to denote arguments that are available at runtime, but
+  nothing about ``foreach`` is required to make this all work (as far as scoping is
+  concerned).
+
+  Accepting the ``@(..)`` syntax does *not* entail accepting this new, separate
+  ``(..)`` syntax, though it is good to know that the idea is forward compatible
+  with.
+
+  A ``@(..)`` argument counts as a type argument when asking whether ``-XScopedForAlls``
+  affects a function equation.
+
+  The new ``@(..)`` notation does *not* work with expression type signatures,
+  lambda-expressions, or anywhere other than a function binding with a type
+  signature. This is because doing so would require propagating type
+  information into scoping, which is problematic.
   
 Costs and Drawbacks
 -------------------
--- a/proposals/0050-type-lambda.rst
+++ b/proposals/0050-type-lambda.rst
@@ -1,13 +1,17 @@
 Binding type variables in lambda-expressions
 ============================================
 
-.. proposal-number:: 50
-.. trac-ticket::
-.. implemented::
-.. highlight:: haskell
-.. header:: This proposal was `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/155>`_.
 .. sectnum::
-   :start: 50
+.. author:: Richard Eisenberg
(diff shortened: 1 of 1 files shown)


int-index:
I agree with what's written here, except for one bit.

> This new version introduces a new extension `-XTypeAbstractions` that will control the new syntax, as well as disable the feature of `-XScopedTypeVariables` that brings `forall`d variables into scope.

These emergent interactions between extensions aren't good. And what is left of `-XScopedTypeVariables` at this point? Pattern signatures!

Let's just accept #119 for `-XPatternSignatures`, and mark `-XScopedTypeVariables` incompatible with `-XTypeAbstractions`.

int-index:
> And what is left of `-XScopedTypeVariables` at this point? Pattern signatures!

Actually, this is not correct. Pattern signatures alone would not enable the expected scoping behavior:

```
f (x :: a) =  ...   -- either rejected or 'a' not in scope
```

Therefore, disregard my previous comment. I'm OK with the proposed amendment as is.

Ericson2314:
Oh this is great! I preferred this design for aesthetic reasons anyways. 

> And what is left of `-XScopedTypeVariables` at this point? Pattern signatures!

I do still like this idea.

Is there precedent for one extension disabling part of another? I do generally like it when the set of extensions mapped to the set of accepted programs is monotonic, but, perhaps this isn't worth the additional extensions needed. 

Perhaps a better argument is it would be nice to deprecate `forall` binding in term, but that's awkward to do if we don't want to change the meaning of `-XScopedTypeVariables` and yet require both `-XScopedTypeVariables` and `-XTypeApplications` to get the desired behavior.
 
> Actually, this is not correct. Pattern signatures alone would not enable the expected scoping behavior:
> 
> ```
> f (x :: a) =  ...   -- either rejected or 'a' not in scope
> ```

That makes sense to me. I'd think one would have to write something along the lines of:

> ```
> f @(... a ...) (x :: a) =  ...   -- accepted, 'a' in scope
> ```

Likewise, I'd expect `id :: a -> a` would be an error to me if we had an extension which required explicit binding in types. (And I wish to write that extension if for no other reason than teaching.)

int-index:
> That makes sense to me. I'd think one would have to write something along the lines of:
> ```
> f @(... a ...) (x :: a) =  ...   -- accepted, 'a' in scope
```

That's not equivalent, though. Compare:

```
f :: Bool -> Bool
f (x :: a) = True    -- ok, a ~ Bool

g :: Bool -> Bool
f @a (x :: a) = True     -- rejected, no 'forall' to go with the @a binder
```

**Comment by goldfirere** [comment]
> I do generally like it when the set of extensions mapped to the set of accepted programs is monotonic.

That's already not true for a number of reasons. `-XMonoLocalBinds`, `-XRebindableSyntax`, `-XTemplateHaskell` (which steals syntax), etc.

What *is* odd about this new proposal is that `-XTypeAbstractions` *always* will override the `-XScopedTypeVariables` behavior, regardless of the ordering of extensions. In contrast, `-XTypeFamilies` implies `-XMonoLocalBinds`, but a user can say `-XTypeFamilies -XNoMonoLocalBinds` and get the behavior they want. (I did this once and it made my life better.)

Of course, we could allow `-XScopedTypeVariables` to dislodge (part of) the `-XTypeAbstractions` behavior. The only part that's problematic is when we use a type abstraction near a `forall`:

```
f :: forall a. a -> a
f @a x = ...   -- problem

g = ((\ @a x -> ...) :: forall a. a -> a)  -- problem

hr :: (forall a. a -> a) -> ...
hr = ...

j = hr (\ @a x -> x)   -- hunky dory. The `forall` does not scope here!
```

So maybe that suggests this breakdown:

* `-XPatternSignatures`: allow `::` in a pattern
* `-XPatternTypeVariables`: implies `-XPatternSignatures` and allows the binding of a type variable in a pattern. These type variables will be bound to arbitrary types (not only other variables).
* `-XMethodTypeVariables`: allows type variables in instance heads to scope over methods and type variables in a class head to scope over method defaults
* `-XScopedForAlls`: allows `forall` to bring tyvars into scope in function bodies and expressions with signatures
* `-XScopedTypeVariables`: Catch-all extension that implies all of the above
* `-XTypeAbstractions`: Enables binding type variables with `@a` syntax. Implies `-XNoScopedForAlls`.

With this approach, a user could specify `XTypeAbstractions -XScopedForAlls` and then get type abstractions in the higher-rank scenario, but not in any scenario with an explicit `forall`. This would work just fine.

I would also favor coming up with a deprecation plan for `-XScopedForAlls`. Or maybe not, since it's used *so* widely. At least new code could be written to avoid it, and individual style guides could forbid it, if they so choose.

**Comment by int-index** [comment]
I'm skeptical of supporting a configuration where `-XScopedForAlls` and `-XTypeAbstractions` are enabled simultaneously, but the latter is limited to the higher rank scenario. We're back to the non-local interactions between extensions (in one way or the other).

Let's just roll with the current proposed amendment.

**Comment by goldfirere** [comment]
> We're back to the non-local interactions between extensions (in one way or the other).

Not really. What `-XScopedForAlls` means is that `xyz :: forall a b c. blah` really means `\ @a @b @c -> xyz`. Indeed, this is essentially the specification of `-XScopedForAlls`.

The code implementing `-XTypeAbstractions` won't even have to check whether `-XScopedForAlls` is enabled. Suppose we've written `(\ @a x -> (x :: a)) :: forall b. b -> b` with `-XTypeAbstractions -XScopedForAlls`. GHC will bring `b` into scope in the body of the annotation, and then check `\ @a x -> (x :: a)` against the type `b -> b`, where `b` is a skolem. This will fail, because we allow `\ @a` only when checking against a `forall` type. With `-XScopedForAlls` off, we would check against `forall b. b -> b`, and then all would work.

**Comment by int-index** [comment]
What's the behavior of `-XScopedForAlls` with regards to TLKSs? Does it take arity into account or does it instantiate blindly?

That it, does this code

```
type T :: forall b. b -> Maybe b
type T = Just :: forall a. a -> Maybe a
```

get desugared into

```
type T :: forall b. b -> Maybe b
type T @b = Just :: forall a. a -> Maybe a
```

and fail?

It would work with the `AScopingErr` trick we discussed.

I need to know the full story, both for terms and types, to make progress on the implementation of TLKSs.

I don't have too strong an opinion on the matter, but I'd like something simple rather than six extensions with intricate interplay between them (`-XPatternSignatures`, `-XPatternTypeVariables`, `-XMethodTypeVariables`, `-XScopedForAlls`, `-XScopedTypeVariables`, `-XTypeAbstractions`).

Perhaps:

* The `-XScopedTypeVariables` flag continues to have its current meaning: scoped `forall` in terms, no effect whatsoever in types.
* Type abstractions do not need a new extension. There's a straightforward desugaring for scoped `forall`s, as you have shown:
  ```
  xyz :: forall a b c. blah  ==>  \ @a @b @c -> (xyz :: blah)
  ```
  Just keep at it in spite of user-written type abstractions. For example,
  ```
  f :: forall a. a -> a
  f @b x = ...   -- problem
  ```
  is mapped to
  ```
  f @a @b x = (... :: a -> a) -- no problem, just a type error
  ```
  It is a type error because the `@b` binder does not have a matching `forall` in the user written signature.
  Given the fact that `-XScopedTypeVariables` only looks at the first `forall`, we can disable it with an empty `forall.`, therefore the following is valid:
  ```
  f :: forall. forall a. a -> a
  f @b x = ...   -- no problem
  ```
  This time `@b` is a binder for `forall a.`.
* To get rid of the `forall.` workaround, introduce a new extension `-XScopedForAll` (not `-XScopedForAlls`, it's singular for consistency with `-XExplicitForAll`) tightly coupled with `-XScopedTypeVariables`, which controls the `forall` scoping part of it. That is, `-XNoScopedForAll` disables `forall` scoping in terms (there's nothing to disable in types) and `forall.` is no longer needed.
* In a future release make `-XNoScopedForAll` the default.

**Comment by simonpj** [comment]
Gah. I'm lost again

> Recent discussion (on only tangentially related features) has brought to light that the accepted proposal for binding type variables does not fly

OK so where is "the accepted proposal"?

*  I start on [the GHC proposals home page](https://github.com/ghc-proposals/ghc-proposals).   
* From there I can see a link to the [list of accepted proposals](https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3AAccepted).  
* Scrolling down that list I see ["Binding type variables in lambda-expressions "](https://github.com/ghc-proposals/ghc-proposals/pull/155).  Maybe that's the one.
* Bother, no, that's the _discussion_ about the proposal.  Click on the [Rendered](https://github.com/goldfirere/ghc-proposals/blob/type-lambda/proposals/0000-type-lambda.rst) link.
* And I get a 404 page not found.

*There has to be an easier way than this to find the text of an accepted proposal!*

Plus, maybe this is a snapshot of the exact accepted propsoal, and Richard has a modified version.

**Comment by goldfirere** [comment]
The [accepted proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0050-type-lambda.rst) and a [rendered version of the update](https://github.com/goldfirere/ghc-proposals/blob/type-abstractions/proposals/0050-type-lambda.rst). 

Apologies that these were not more readily available. And I've updated the link in #155.

**Comment by goldfirere** [comment]
I say `-XScopedForAlls` (yes, I know that `-XExplicitForAll` is singular, and that's always bothered me. I didn't see a need to repeat that mistake, but others may feel differently) should affect types just like terms. So the desugaring is exactly as you suggest in https://github.com/ghc-proposals/ghc-proposals/pull/238#issuecomment-499660404.

But that's just my current opinion at the moment, and this all may yet evolve.

**Comment by simonpj** [comment]
> There has to be an easier way than this to find the text of an accepted proposal!

What is the canonical way to do this?  (Admittedly, this is a tangent to the main purpose of this ticket.)

**Comment by simonpj** [comment]
The more I think about the big-lambda proposal, the less I like it.   This unexpected interaction with forall-scoping has taken me aback.

----------------------
Leave aside scoped type variables for the moment.   Here are some questions.
```
f :: forall a b. a -> b -> b
f = let { stuff } in \@a. 
      case thing of
         True -> \x y. y
         False -> \@b. \ p q. q
```
I assume this is supposed to typecheck (but only if there is a signature!).   But note that
* Currently `TcExpr.tcPolyCheck` with a `CompleteSig` (which is the relevant code path) always skoelmises the type immediately.  Now it can't do that.  
* Well, it can (and must) if there are any argument on the LHS.  So we we need an extra case when there are no arguments, then we must pass in a non-skolemised type polytype, bypassing `tcMatchesFun` (which also skolemises).
* Then we'l get to `tcPolyExpr`.  Oh!  It skolemises too, and we mustn't do that either.
* But then when we get to the lambda, or anything else, we must skolemise.

Probably this is all do-able.  But it's a change that potentially affects a lot of code, much more than I realised.
--------------------------
What about pattern bindings?  Am I allowed
```
f :: forall a. a ->a
(f,g) = (\@a \(x:a). x,  blah)
```
I'm guessing not, but the proposal is silent.  (Forall-scoping doesn't work for pattern bindings, incidentally.)
-----------------
I was on the fence about this big-lambda thing; now I'm mildly against.   Esp as it only works in "checking mode", so it doesn't satisfy the "you can do System F" goal.   We have so many other fish to fry, I rather think we should park this one.  (To mix metaphors.)

**Comment by goldfirere** [comment]
I *do* expect there to be a non-trivial implementation burden, but I don't think the resulting code will be any worse (i.e. more complex) than what we have now. To implement type applications, we needed lazy instantiation. To implement type abstractions, we naturally need lazy skolemization.

If you're bothered by the "checking" mode requirement, we could think harder about @gridaphobe's suggestion in the [committee discussion](https://mail.haskell.org/pipermail/ghc-steering-committee/2019-March/000935.html):

>  Why not say that `\ @a -> body` infers the type `forall a. ty` where `body :: ty`, and let unification handle the rest?

I think that might work, but I'd want to draw some typing rules to be sure. Would that sway you?

**Comment by simonpj** [comment]
I think some typing rules would be a very good thing!

I'm not dead set against this.  I just feel uneasy.   And that makes me wonder if we might be better spending our finite budget of intellectual cycles elsewhere, since this has turned out not to be the "quick win" that I initially thought it might be.

**Comment by gridaphobe** [comment]
I've updated the description of this proposal with a link to the [rendered rich-text diff](https://github.com/ghc-proposals/ghc-proposals/pull/238/files?short_path=09b9e6a#diff-09b9e6a0c9d0d707a9e97cea3f11f49a), which is pretty handy.

> Apologies that these were not more readily available. And I've updated the link in #155.

I think this suggests a useful addition to the protocol for accepting (or rejecting!) a proposal. The link at the time of discussion will point to the author's fork of this repo, which they might well update or delete in the future. We should update the "Rendered" link in the PR with a link to the permanent home of the proposal. This suggests that we ought to also keep a store of rejected proposals too for easy future reference. (ping @nomeata)

---

I'm still pretty confused as well though. The proposal doesn't seem to explain what the problem is with having both `forall` binders and `@` binders, unless it's in this single sentence:

> One wrinkle is that, according to the rules in the `Visible Type Application`_ paper, type variables quantified by a forall in a type signature get brought into scope before we check the term.

But this doesn't sound like a problem to me. It sounds like the `forall` binder would be brought into scope first, and then the `@` binder would come into scope, possibly shadowing the `forall` binder.

Oh, there's also this sentence that I missed at first.

> Note that limiting the behavior of -XScopedTypeVariables is necessary if we think about where type variables are brought into scope. Are they brought into scope by the forall? Or by the `@a`? It can't be both!

Why not? We allow shadowing at the term level, and also at the type level. This feels very much like shadowing to me.

**Comment by nomeata** [comment]
> > There has to be an easier way than this to find the text of an accepted proposal!
> 
> What is the canonical way to do this?

Thanks for pointing this out, I change the [List of accepted proposal](https://github.com/ghc-proposals/ghc-proposals/tree/master/proposals) so something accurate (albeit less pretty).

**Comment by int-index** [comment]
> I say -XScopedForAlls should affect types just like terms.

@goldfirere That's impossible, we don't have Λ in types. You seem to suggest that

```
Just :: forall a. a -> Maybe a
```

maps to

```
\ @a -> (Just :: a -> Maybe a)
```

and gets rejected for lack of Λ?

Or do we map to `AScopingErr`? Then we should map to `AScopingErr` in TLKSs, too. Therefore, `-XScopedForAlls` would not accept any new type-level programs, it would simply change the error message from "not in scope" to "scoping error".

**Comment by goldfirere** [comment]
About TLKSs: I say that having `forall` in a TLKS works just like it does in terms. For our purposes, we do have a lambda in types here, because we can bind type variables in the LHS of data type declarations. `forall` outside of a TLKS should do the `AScopingErr` thing.

I have added some text to the proposal about why we can't have both. I was secretly hoping that vigorous waving of hands (and with an exclamation point!) would allay concerns, but I was rightly mistaken.

**Comment by int-index** [comment]
> About TLKSs: I say that having `forall` in a TLKS works just like it does in terms.

In terms, we don't have to worry about the arity. I say we want the following code to be accepted:

```
type T :: forall a. a -> Maybe a
type T = Just :: forall a. a -> Maybe a
```

And it's easy to make it accepted: map `a` to `AScopingErr` in the TLKS.

I think the rule in types should this:

* Scoped `forall` works like in terms for variables in TLKSs that fall within the arity of the tycon. Other variables are mapped to `AScopingErr`.

"Other variables" are variables in TLKSs that cannot be instantiated due to the inferred arity, and variables outside TLKSs.

This accepts more programs than the naive "just like it does in terms" rule.

**Comment by goldfirere** [comment]
I think the TLKS issue was resolved elsewhere, and I'd like to return this conversation to the proposal at hand, which is concerned with the term-level, not the type level. Any further commentary before we go to committee? Thanks!

**Comment by Ericson2314** [comment]
@goldfirere what about your alternative with many more extensions? (Meta: how do alternatives go to the committee when an existing proposal is being amended?)

**Comment by simonpj** [comment]
In [my comment above](https://github.com/ghc-proposals/ghc-proposals/pull/238#issuecomment-499681597) I asked about pattern bindings. The proposal remains silent.   Could you add language to say what is supposed to happen?

**Comment by goldfirere** [comment]
This needs attention from me, but I don't have attention to give this week. It is not abandoned, though. I shall return! (Next week, likely.)

**Comment by sgraf812** [comment]
I'm not sure if this is the right place to put this, but after talking to @goldfirere I'll post my own motivation for this: https://stackoverflow.com/questions/50159349/type-abstraction-in-ghc-haskell

```haskell
{-# LANGUAGE AllowAmbiguousTypes    #-}
{-# LANGUAGE RankNTypes             #-}
{-# LANGUAGE TypeApplications       #-}
{-# LANGUAGE TypeFamilies           #-}

module Foo where

f :: Int -> (forall f. Functor f => Secret f) -> Int
f x _ = x

g :: (forall f. Functor f => Secret f) -> Int
g = f 4

type family Secret (f :: * -> *) :: * where
  Secret f = Int
```

This (the binding for `g`, in particular) currently doesn't type-check. We'd need a binding construct and 'eta-expand' the existential type parameter `@f`. Thinking I hit a bug (I had expected that the eta-reduced form should work anyway), I opened https://gitlab.haskell.org/ghc/ghc/issues/15119, but this was closed because we didn't have the expressivity at the time required to accept the program.

I'd really like this to become a thing, if only for reasons of symmetry with `-XTypeApplications`.

simonpj:
> I'd really like this to become a thing, if only for reasons of symmetry with -XTypeApplications.

To be clear, you are advocating [this already-accepted proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0050-type-lambda.rst).   This discussion thread is, I think, a diff to that proposal.

Moreover I think that with the type-lambda proposal you'd have to write
```
g :: (forall f. Functor f => Secret f) -> Int
g x = f 4 (\@f. x @f)
```
That would specify that you want to instantiate `x` at the same type over which `f 4`'s argumetn is abstracted.  Is that what you had in mind?

ulysses4ever:
This is not the first time a proposal interacts closely with some earlier proposal (return type signatures has this story too, I think). Maybe there should be a formal process to amend existing proposals?

sgraf812:
>  `f 4 (\@f. x @f)` Is that what you had in mind?

Yes, exactly.

cartazio:
while I'm not 100% following everything in this proposal, when will we have lexical scoping for type variables?!

the most common question people would ask me about Haskell202X has been "will scoped type variables become the default", and I definitely remember Richard saying this would be a bad idea owing to how that extension is designed due to historical way it worked.

does this proposal pave the way for lexical scoping being the universal default for variables in ghc programs?

Ericson2314:
@cartazio I would think so!

goldfirere:
I have refreshed this proposal, as there is now interest in implementing #155, which this proposal amends. There is also now a paper that gives typing rules (though this feature is not a focus of that paper). See the new Proposed Change Specification.

Note to careful readers: one typing rule (Pat-CheckType) in the [paper](https://richarde.dev/papers/2021/stability/stability.pdf) appears as if it may be slightly wrong, a problem I noticed while writing this. I am checking with my co-author @Gertjan423 to see if he agrees and hope to update the posted version of the paper soon.

/remind me in two weeks to submit if conversation has died down.

Ericson2314:
In the paper on line 382 you wrote `swizzle = undefined` but I think you both meant `swizzle = undef`?

Ericson2314:
Also, I think `PatternTypeVariables` is the same as `PatternSignatureBinds` from my (accepted) #285?

goldfirere:
> Also, I think `PatternTypeVariables` is the same as `PatternSignatureBinds` from my (accepted) #285?

Close, but not quite. But many thanks for connecting these two -- I had forgotten about `-XPatternSignatureBinds`.

My `-XPatternTypeVariables` is about the scope of the variables, while your `-XPatternSignatureBinds` is about the presence of the variables at all.

```hs
f (Left x :: Either a a) = ... 
  -- accepted with my -XPatternSignatures without any further extensions,
  -- but would be invalidated by your -XNoPatternSignatureBinds
```

My `-XPatternTypeVariables` means that the `a` above would scope over the RHS.

I don't feel strongly about this particular corner of the design. In light of the fact that #285 is already accepted, I'll revise this proposal accordingly.

Ericson2314:
> My `-XPatternTypeVariables` is about the scope of the variables, while your `-XPatternSignatureBinds` is about the presence of the variables at all.

Ah, I see.

> I don't feel strongly about this particular corner of the design. In light of the fact that #285 is already accepted, I'll revise this proposal accordingly.

Thanks! 

From the proposal:

> This extension is on by default (but is useless without `-XPatternSignatures`).

That matches what I proposed in #285, but I actually like your original proposal of having the extension merely be implied by `-XScopedTypeVariables` rather than on by default.

 - Without `-XPatternSignatures` existing there is no difference, but with it existing there is, and since i am skeptical of binding in pattern signatures binding variables, I them being opt-in not opt-out until `-XScopedTypeVariables` is turned on.
 - The new `-XMethodTypeVariables` and `-XScopedForAlls` are both opt-in but implied by `-XScopedTypeVariables`, so with the original defaulting we keeps the symmetry between all 3.

simonpj:
Section numbering is funny (51, 52 etc).

Formatting wrong in bullet 4 of Proposed change spec

It would help to elucidate the connection with #285 in the proposal itself.

Under proposed change spec there are two big things:
1.  `-XTypeAbstractions`, the main payload
2. Making `-XScopedTypeVariables` more fine-grain. 

I'm not against (2) but 
* I'd love to see it laid out like this two-pronged way.  At the moment there are 6 proposed changes, of which 5 concern (1) and only one concerns (2).  Could you group them more clearly.
* What's the motivation for (2)?  I see that you want `-XTypeAbstractions` to change the meaning of `-XScopedForAlls` (and hence `-XScopedTypeVariables`), but I don't see that forces this finer-grain flaggery.  Why not just do (1)?

Ericson2314:
To me the motivation is that, with `-XTypeAbstractions`, `-XScopedForAlls` and `-XPatternSignatureBinds` become arguable misfeatures with replacements, but `-XMethodTypeVariables` is (perhaps unfortunately) still needed. It would be possible to skip `-XScopedForAlls` and have that just be the remaining core feature of `-XScopedTypeVariables`, but I do sort of like `-XScopedTypeVariables` *just* enabling other features, so we have more decomposition into composable parts. I think that might help with teaching.

`-XPatternSignatures` is also nice in order to separate out the least controversial part of all of this. It's a knob we don't need only because I very much hope it can uncontroversial become part of the next version of Haskell (likewise `-XExplicitForAll`) while the rest of this stuff would probably would get more haggling.

Ericson2314:
I made https://github.com/ghc-proposals/ghc-proposals/pull/420 to make the changes to `-XPatternSignatureBinds` inspired by the `-XPatternTypeVariables` previously in this proposal that I described above in https://github.com/ghc-proposals/ghc-proposals/pull/238#issuecomment-820079174

That small would free this proposal to drop 2) entirely --- we would just be continuing a process of moving functionality from `-XScopedTypeVariables` into other extensions implied by it --- which I think would be nice to narrow the discussion here.

goldfirere:
I've updated this in light of recent comments, splitting up the specification section and changing the default setting of `-XPatternSignatureBinds` (in sympathy with #420).

simonpj:
> splitting up the specification section 

Much better split up like this!

Please give at least one example for each point in Proposed Change Spec. Such examples might be unnecessary if the spec were complete and formal, but as it is I desperately need examples.

I did not understand
```
f :: forall a. a -> a
f x = (x :: a)      -- rejected with -XTypeAbstractions
```
Why rejected?   Where in the Change Spec does it say that.  Do you intend that -XTypeAbstractions *requires* explicit type abstractions everywhere in the module. Could they not be optional?

cdsmith:
The "no forks" philosophy seems to suggest that since `TypeAbstractions` is incompatible with most of `ScopedTypeVariables`, it should only be added if one expects a migration _away_ from scoped type variables for existing code.  The only alternative offered here requires visible type application, with its unfortunate dependence on positional type parameters, for that use case instead.  Positional type parameters are, of course, quite problematic in conjunction with type inference, to the point that I think it's a mistake to use them in any place where there's more than one type variable in question.

`ScopedTypeVariables` is quite widely used!  Do we really expect Haskell to migrate so completely to positional type parameters that `ScopedTypeVariables` could be abandoned?

Ericson2314:
Type *parameters* have been positional since we had explicit `forall`. It's only type applications that exposes the positionality to the caller, via the caller's positional *arguments*. This proposal, being about parameters and not arguments, doesn't leak any new information.

cdsmith:
> Type _parameters_ have been positional since we had explicit `forall`. It's only type applications that exposes the positionality to the caller, via the caller's positional _arguments_. This proposal, being about parameters and not arguments, doesn't leak any new information.

I'm not sure what you mean, but I've thought about this, and it's less problematic than I originally thought.  There's still a type signature on the previous line with the `forall` staring at you, so in this particular instance you do know the order of type variables.  It's just that I'm accustomed to thinking of type variables as not having an order (even though they obviously do have an order in core - but that's just implementation detail - and in Haskell with `TypeApplications`, which I avoid), and this shatters that illusion.  This particular instance doesn't really cause any concrete problems, but just pushes the programmer into a more complex mental model.

goldfirere:
We could imagine something like

```hs
f :: forall a b c. ...
f @q @(..) ... = ...
```

This uses new syntax `@(..)` (noting that `@..` is available as a binary operator and so would steal syntax) to say "bring all the rest of the type variables into scope, using the names as written in the type signature". In the concrete example, what is called `a` in the type signature would be called `q` in the body, and what is called `b` and `c` would be called `b` and `c` (respectively) in the body.

This `@(..)` syntax would *also* be available if the user does not write `forall` in the type signature. Viz:

```hs
id :: a -> a
id @(..) x = (x :: a)
```

I quite like this.

Open question:

```hs
hr :: (forall a. a -> a) -> ...
hr = ...

x = hr (\ @(..) -> {- here -})
```

Is `a` in scope in `here`? Note that the type signature for `hr` can be arbitrarily far away, including imported. The consistent thing is "yes".

What do others think of `@(..)`?

Ericson2314:
As long as `(..)` could also automatically bind all the visibly dependent quantified type variables, for symmetry, I'm for it.

cdsmith:
> What do others think of `@(..)`?

While I haven't had much time to think about it, my initial reaction is that I like it.

What if there is no type annotation?  Is `@(..)` an error in that case?  (I'd think so.  It would be terrible to bind the arbitrary type variable name that GHC makes up or grabs from some unclear source elsewhere via inference.)

Ericson2314:
Yes, it should be an error.

goldfirere:
> As long as `..` (without the `@`) could also automatically bind all the _visibly_ dependently quantified type variables, for symmetry, I'm for it.

Hm. Yes. I think that will do nicely:

```hs
frob :: foreach (multiplier :: Double) (num_rows :: Int) (num_cols :: Int) (desc :: String) -> Result
frob (..) = -- now, multiplier, num_rows, num_cols, and desc are all in scope
```

I like. Semantics: `(..)` brings into scope all variables up to the next non-`forall ->` non-`foreach ->` argument.

cdsmith:
Assuming this is going to be accepted, and therefore the decision is made to eviscerate `ScopedTypeVariables` in its current form... is it too late to remove it from GHC2021 before the 9.2 release?  It would be unfortunate to carve it in stone and create an even greater migration burden.

Ericson2314:
@cdsmith Hehe, I didn't bring that very point up in https://github.com/ghc-proposals/ghc-proposals/pull/380#issuecomment-731606787. The response was basically "no one said 2021 -> 2021 wouldn't be a breaking change".

cdsmith:
I'm nitpicking a bit here, but I strongly believe that the UI matters, so I'll raise these concerns, at least.  I've just noticed the weird interaction with inferred versus explicit type variables.  I definitely got on-board with the idea that `forall` now binds variables only within the type, and not the function body, and that we now have an explicit `@t` parameter (including the `@(..)` form) to do the binding in the function body.  I think that's better.  But having the old way still work for *inferred* variables honestly seems like the worst of both worlds!  The meaning of `forall` and the means of binding variables in the function body is now wildly different for the two, and you have to think about *both* syntaxes, and the means of knowing when each means what.

In this case, I'd really like to see `ScopedForAlls` (that is, the main meaning of the old `ScopedTypeVariables`) be an error when used together with `TypeAbstractions`.  This is a far better failure mode than changing the meaning of the extension and thus possibly **silently changing the meaning** of existing code!  Yuck.

- Do we need a way to bring inferred variables into scope at all?  Probably so, since one of the stated goals of #99 was to decouple the choice of which types are explicit vs inferred, and which can be named in type signatures in the implementation.  It would be walking back that progress if adding the right type signature to a `where` clause required making a type variable explicit so that it could be referred to by name. :(
- Could `@(..)` also bind inferred variables?  Certainly, positional `@a` cannot; this would be hideously at odds with the meaning of visible type application.  But I don't see the same issue with `@(..)`.  I don't know that I'd *expect* it to do so, exactly, since it looks like a visible type application.  Nevertheless, at least it seems much, much less bad than half-retaining the `forall` behavior from `ScopedTypeVariables` but only for some names.
- Or, is there another option that's less bad than either?

cdsmith:
> This is a far better failure mode than changing the meaning of the extension and thus possibly **silently changing the meaning** of existing code!

I guess I owe a demonstration that silently disabling the binding behavior of `ScopedTypeVariables` doesn't just cause errors, but possibly change the meaning of code.

https://code.world/haskell#PPqiKatOMwtyCXq_e_ZBO_w

Ericson2314:
So if we could do `@{..}` to I think that would suffice for the explicit-inferred variable case --- I gather the modified `ScopedTypeVariables` was mainly because those otherwise could be unbindable. If so, then I agree we would be better off just making `ScopedForAlls` and `TypeAbstractions` mutually exclusive.

simonpj:
I'm not keen on this `@(..)` business.  Anything that brings a bunch of unspecified variables into scope is tricky to reason about.  (Record puns are a case in point.)

What's the motivation?  Can't we wait and see how pressing it is?  We don't have to do everything at once.

Ericson2314:
I also want to add that not only do a only do I support `@(..)`, but it greatly relieves me. I was worried I was going to sacrifice much convenience on the alter to system F, making people bind their variables twice because "type signatures and their terms are different scopes, dammit!", and this boilerplate was going to usher in a massive thermidorian reaction. But `@(..)` keeps the principle mostly intact, while dispensing with the repetition. Whew!

cdsmith:
> I also want to add that not only do a only do I support `@(..)`, but it greatly relieves me.

I feel very much this way, too.  After a few very tedious experiences fighting `MonoLocalBinds`, and starting to see similar reactions to simplified subsumption in GHC 9.2 alphas (which I've yet to use myself), I'm growing especially tired of "oh, it's fine, and easy to work around" justifications attached to making Haskell less convenient to use.

simonpj:
> Since this proposal is going to make things less convenient

Ah, that's interesting.  Could you just articulate explicitly the loss of convenience?  Perhaps that is what I am missing. After all, you could continue to use Haskell precisely as now.

One proposal that we discussed in the early days of Haskell's design was the repetition of parameter names.  We write
```
f pingle pongle [] = blah1
f pingle pongle (a:as) = blah2
```
We discussed abbreviating the names, something like this
```
f pingle pongle [] = blah1
f "          "          (a:as) = blah2
```
But we decided to make it less convenient and require repetition, because it make it harder to determine exactly where the variables mentioned in `blah2` are bound.   Perhaps this is the same kind of choice?

Of course no two situations are quite alike.

cdsmith:
Well, previously one could write:

```

f :: forall m t a. ...
f = ... {- code mentioning m, a, and t in types #-}
```

But according to this proposal, `TypeAbstractions` would prevent that from working so that one must write:

```
f :: forall m t a. ...
f @m @t @a = ... {- code mentioning a, b, and c in types #-}
```

This is not only more typing; it also forces programmers to view order of type variables as significant, even if they are not using `TypeApplications`.  (If one accidentally writes `f @t @m @a` then... disaster.  If you're lucky, a type error.)

Certainly, it's true that one can just not use `TypeAbstractions`, and this isn't an issue.  But from the review criteria:

> - Does not create a language fork. By a "fork" we mean
>    - It fails the test "Is this extension something that most people would be happy to enable, even if they don't want to use it?";
>    - And it also fails the test "Do we think there's a reasonable chance this extension will make it into a future language standard?"; that is, the proposal reflects the stylistic preferences of a subset of the Haskell community, rather than a consensus about the direction that (in the committee's judgement) we want to push the whole language.
>
> The idea is that unless we can see a path to a point where everyone has the extension turned on, we're left with different groups of people using incompatible dialects of the language. A similar problem arises with extensions that are mutually incompatible.

This proposal explicitly creates two such mutually incompatible extensions: `TypeAbstractions` disables most of `ScopedTypeVariables`.  (I argue above that since this can silently change the meaning of existing programs, it should really be an error to enable both, instead.)  One of the stated goals of the proposal is also to deprecate `Proxy`, and doing so would essentially force the use of `TypeAbstractions` on some portion of Haskell code.  So with this in mind, it makes sense to view this proposal as effectively deprecating `ScopedTypeVariables` with `TypeAbstractions` as its replacement, and that's where the loss of convenience comes from.

Of course, one could accept the proposal and implement it _just to see_ whether it's adopted or not, if that's what you meant.  This would be an exception to the rule about forks.  I'd be a bit concerned that the adopters would not be the same kinds of people who would be affected by the lost convenience in much more basic Haskell programming. `ScopedTypeVariables` is used by a much larger part of the Haskell community than those who would typically need `TypeAbstractions`.

> But we decided to make it less convenient and require repetition, because it make it harder to determine exactly where the variables mentioned in blah2 are bound. Perhaps this is the same kind of choice?

This is interesting!  I do agree it's the same kind of choice, but the details matter a bit.  As I mentioned earlier, needing to name the type variables also means needing to think of them as having an order, which is not necessary in Haskell today unless one is using `TypeApplications`.  So in this case, instead of just being more explicit about a concept that's already there, it's asking for the programmer to change how they understand the meaning of their type.  Instead of a logical statement "for all types `m`, `t`, and `a`, my function can have the following type" into a statement like "my function takes hidden type arguments `m`, then `t` , then `a`".  The latter is closer to System F, but I don't want to program in System F.

simonpj:
Aha, thanks for explaining.

How about this: any use of an explicit type abstraction on the LHS switches off the forall-scoping.  Thus
```
f :: forall a b c. Colour -> blah
f                    Red    = ...a,b,c in scope...
f @x @y @z Green = ...x,y,z in scope...
f @x              Blue   = ...only x in scope....
```

cdsmith:
@simonpj Does that solve the motivating problem?  To be clear, I don't actually understand that motivating problem, but to quote Richard:

> The behavior of `ScopedTypeVariables` appears incompatible with the ability to bind type variables using a `@a` pattern. (What would `(\ @a -> expr) :: forall b. ty` mean? What would its desugaring be? What's the relationship between `a` and `b`?)

Here, the use of `ScopedTypeVariables` is not on the same binding as the use of `TypeAbstraction`.

Actually, wait... is that a use of ScopedTypeVariables at all?  Okay, I'm confused... @goldfirere or someone else may be able to clarify.

Ericson2314:
@simonpj I would still prefer `@(...)`.

1. It's more local. With multiple equations, as you point out, the type signature can be quite far away. Having the per-equation `@(..)` says "look up!".
2. It allows getting rid of the forall-or-nothing rule. With the new quantifiers we would get for dependent haskell, the `forall-or-nothing` rule is still well defined, but less obvious (see https://gitlab.haskell.org/ghc/ghc/-/issues/18660 and related issues). `@(..)` works perfectly well with free-var-based implicit binders.
3. `@(..)` + free-var-based implicit binders is the best combination for traditionalists that don't want to think of type vars having an order. This is because both the signature and term bindings are implicit about the order. I think it's pretty neat that this proposal can make things better for the "System F" and "Hindly Milner" camps.

Ericson2314:
> The @(..) construct works for both specified and inferred variables.

I think you might want to add back the reference to the other proposal / extension to indicate you mean explicit inferred variables? Because "actually" inferred ones (whether from no sig checking mod or poly kinds) won't have a name to bind.

goldfirere:
After a conversation with @simonpj, I've sketched out a middle ground that might make everyone happy.

One unfortunate consequence is that we unconditionally lose `((\x -> (x :: a)) :: forall a. a -> a)` as an expression. Hopefully, using an expression signature to bring a type variable into scope in this way is rare. The proposal sketches out a plausible backward-compatible alternative (by heuristically looking for the presence of a type argument to control the behavior), but I dislike it.

goldfirere:
@cdsmith Hopefully this new version -- which does not gut `-XScopedTypeVariables` nearly as much, is more acceptable.

Separately, I'm not exactly sure what you're asking about, above. Perhaps this will help: in the expression `(\ @a -> expr) :: forall b. ty`, suppose we wish to bring `b` into scope. Then, what type do we check `\ @a -> expr` against? Is it just `ty`? Then the `@a` part is surely not going to work out. On the other hand, if we check `\ @a -> expr` against `forall b. ty`, what does it mean to have `b` in scope when we have a type `forall b. ty` around, where `b` is locally bound? I really don't know. I thus propose simply not to bring `b` into scope (and to check `\ @a -> expr ` against `forall b. ty`). And, in this new version, that's the *only* change to `-XScopedTypeVariables`.

RyanGlScott:
(I'm reading this proposal for the first time a bit late, so I apologize if I ask questions that have already been discussed.)

I want to make sure that this proposal is able to encompass all of the use cases that currently require `ScopedTypeVariables`. Based on my reading of [this GHC User's Guide section](https://downloads.haskell.org/ghc/9.0.1/docs/html/users_guide/exts/scoped_type_variables.html#extension-ScopedTypeVariables), that consists of:

1. Declaration type signatures
2. Expression type signatures
3. Pattern type signatures
4. Class and instance declarations

As I understand it, `XMethodTypeVariables` completely covers use case (4), and `-XPatternSignatureBinds` completely covers use case (3). Use case (2) is discussed in the proposal in the sense that there is a way to migrate expression type signatures using `ScopedTypeVariables` over to using `TypeAbstractions`.

I'm least sure about use case (1). The particular case of _function_ declaration type signatures is covered at length in this proposal, but there's surprisingly little in the way of _pattern synonym_ declaration type signatures. For example, suppose I define a pattern synonym that uses `ScopedTypeVariables`/`ScopedForAlls`:

```hs
pattern J :: forall a. a -> Maybe a
pattern J x <- Just (x :: a)
```

Is it possible to convert this code over to use `TypeAbstractions` instead? My hunch is that you would be able to write something like this:

```hs
pattern J @a x <- Just (x :: a)
```

However, the proposal only discusses the changes needed for the `funlhs` BNF production, which covers function LHSes but not pattern synonym LHSes. Is this omission intentional? I would suspect that you would need to change the [`pattern_synonym_lhs` Happy production](https://gitlab.haskell.org/ghc/ghc/-/blob/24a9b1708cee95670e7ec2a6ceb68e29fc376cf7/compiler/GHC/Parser.y#L1628-1631) in GHC's parser as well, but this assumes that the proposal is meant to cover this use case.

JakobBruenker:
From the proposal:
>  If you have a function f :: forall {k} (a :: k). ..., you will have to rely on the old behavior of -XScopedTypeVariables to bring k into scope in f's definition, or you will have to use a pattern signature. This is regrettable but seems an inevitable consequence of the {k} notation.

Is there a reason not to allow `@{k}` in in the lhs of an equation? It admittedly is not reflected by how inferred variables work with visible type application, but then, neither is `@(..)`.

cdsmith:
> @cdsmith Hopefully this new version -- which does not gut `-XScopedTypeVariables` nearly as much, is more acceptable.

@goldfirere For my part, I'm happy with either of the two accommodations: adding `@(..)` or making sure that `forall` binding still works when `TypeAbstractions` is enabled (but not used).  Doing both is even better, since it make migration easier while also leaving a path to a more consistent world, since it's definitely a wart that `forall` will now bind type variables in some cases but not others.  So this definitely addresses my concerns, twice over.

Me earlier confusion was because I didn't realize that expression signatures bound variables with `ScopedTypeVariables` at all, and it seems very odd that they do so.  I certainly don't object to its no longer doing so in this proposal.

maralorn:
I am a bit surprised by the `@(..)` solution. Aren‘t record puns with `{..}` considered harmful?

Is that a niche opinion or how might problems with that be different on the type level?

ramirez7:
> I am a bit surprised by the `@(..)` solution. Aren‘t record puns with `{..}` considered harmful?
> 
> Is that a niche opinion or how might problems with that be different on the type level?

`RecordWildCards` is a pretty commonly-used extension (for good reason - it's nice!)

There are some corners of Haskell-dom that "consider it harmful," but they are mostly driven by corporate-style programming where you must accommodate an ever-changing large-ish team. Such a situation incentivizes working with a niche but standardized subset of the language. I wouldn't say `RecordWildCards` is harmful as a rule, but I can see how in such a situation you would want to avoid it. The same goes for implicit imports too, but they too are still plenty useful in Haskell programming.

That said, the main reason for the "considered harmful" afaiu is that wildcards potentially pollute scope in hard-to-understand ways (especially when used with `let` and `where`), which doesn't seem like as much of an issue here with type variables.

davidsd:
Forgive me if this is tangential to this proposal, but something that has always confused me is: why isn't the scoping behavior of `ScopedTypeVariables` on by default, even without the `forall`? That is, why can't we have

    id :: a -> a
    id x = (x :: a)

To me, this is the most intuitive meaning of the type signature on the first line, and it makes that type signature the "single source of truth" for the names of the types below.

Ericson2314:
@maralorn the `@(...)` is strictly less harmful than the scoped type variables we have today: unlike with record wildcards the explicit field names are somewhat local (next to the definitions in the same module), and the `@(..)` is at least a reminder saying "look up!" where one might miss whether there was a `forall` above otheerwise.

Ericson2314:
@davidsd Originally the choice was I believe due to backwards comparability. But while I'm happy to discard that, i am still not terribly found of automatic free variable binding like that. It is quite fragile, and all the more so when the variables are bound with a larger scope.

Note with this proposal
```haskell
id :: a -> a
id @(..) x = (x :: a)
```
will work, so you can skip the `forall` if you want.

davidsd:
@Ericson2314 Could you give an example of what you mean by fragile?

Often when editing a function, I realize I need access to one of the types of the arguments. I then have to scroll up to the type signature and add `forall m a b p q r . ...`, making the type of that one function a bit uglier than the others in the module. I also often forget to list a type variable that's in the signature somewhere, and have to contend with a few rounds of `ghc` errors.

Under this proposal, `@(..)` at the definition sites would replace `forall m a b p q r . ...` in the type signature. This makes me happy because it avoids the problem of scanning the type signature for free variables and listing them out in a `forall`. It would also make the type signatures of my program look more uniform, and this aids readability. On the other hand `@(..)` is a somewhat noisy string of symbols, and I am not very excited to add extra symbolic noise to the definitions.

What I really want almost 100% of the time is the ability to access type variables from the type signature in the body of my functions. `forall m a b p q r.` is an annoying incantation to achieve this. The proposed replacement `@(..)` is perhaps less annoying, but I still don't understand the need for an incantation at all (unless one wants to modify the order of the type arguments).

(By the way, I am not trying to argue against the main motivating examples of this proposal, such as `usage = higherRankF (\ @a x -> ...)`, which are very compelling.)

cdsmith:
> Often when editing a function, I realize I need access to one of the type variables in the arguments. I then have to scroll up to the type signature and add `forall m a b p q r . ...`, making the type of that one function a bit uglier than the others in the module.

This is, indeed, a bit annoying.  It's the combination of the forall-or-nothing rule (which makes perfect sense without scoped type variables) with the unfortunate second use of `forall` to bind type variables.  The proposal's suggested `@(..)` to bring variables into scope is slightly nicer because you don't have to repeat the entire variable list and it's specific to just one binding where you might need the variables.  That said, it still brings everything into scope when you only really wanted one.  The proposal also (as its core suggestion) offers positional type variables as an alternative to bring just one into scope.  Especially if it's the first type argument, but if not then you can write `f @_ @_ @_ @m`.  It's rather ugly, but oh well.

At the risk of too much piling on random ideas, I think it would be awesome to have something like `@{m}` that means "bind the variable called `m` in the type signature, regardless of its position".  This is consistent with the use of braces to mean a non-positional type variable.  I think I might even like this better than `@(..)`

(There are some subtleties: This should work for _all_ type variables, regardless of their explicitness.  It probably should not consume the argument, so that `f @{a} @b` would mean that `a` refers to the variable called `a` in the type signature, and `b` refers to the first positional type variable, whatever its name in the signature.  If `a` is itself the first positional type variable, then `a ~ b`.  IMO, syntactic consistency would then suggest using `@{..}` instead of parentheses for @goldfirere's wildcard operator, if it's still implemented.)

goldfirere:
@RyanGlScott 

Pattern synonyms. Yes. Fixed now.

@JakobBruenker 

> Is there a reason not to allow `@{k}` in in the lhs of an equation?

The whole reason *inferred* variables exist is because we cannot pin down their ordering. Allowing `@{k}` in the LHS of an equation means there would be an ordering. Here is an example:

```hs
data Proxy a = MkProxy

f :: Proxy a -> Proxy b
f @{k} = ...
```

What does the `k` bind to? `f`'s type has two inferred variables: the kind of `a` and the kind of `b`. (It also has two specified variables: `a` and `b`.) But now we have no idea which one should be called `k` in the body of `f`. You might say that you have to write the quantification over the inferred variable explicitly in the type signature in order to bind it in the definition, but now you've just introduced a middle ground between *specified* and *inferred*: *explicitly inferred*, where a definition can bind explicitly inferred variables with `@{k}`, but not inferred inferred variables. While possible, I do not want this.

@maralorn 

I personally find record puns questionable (though I still use them sometimes) for two reasons:

1. A record pun introduces shadowing whenever there are field selectors (which is most of the time). This is confusing, because the punned name even has a different type than the unpunned name. I have been tripped up by this in practice.
2. A variable brought into scope by a record pun has no easily found binding site. When I've been reading code I'm less familiar with, I see a variable and want to know where it came from. If it came from a record pun, I have to search for a long time. I have been tripped up by this in practice.

Neither problem occurs with this new syntax:
1. There is no similar shadowing.
2. While the location a variable is brought into scope is a bit murky here, a textual search will find the variable mentioned in a binding site -- it's just that the binding site is in the type instead of in the definition. This is mildly confusing, but not nearly as bad as record puns. And, as @Ericson2314 points out, this is still an improvement from the status quo, where at least a `@(..)` tells you that something special is happening.

@davidsd 

Yes, if we were happy to toss out backward compatibility, perhaps we would just always bring those type variables into scope. When I've programmed in dependently typed languages (that have a feature like the one I'm proposing here), I still sometimes wish the type variables just popped into scope without any effort. But `@(..)` is an improvement on those other languages, which lack such a construct. I'm not totally sold on automatic scoping always, but I can duck the question here because doing so would be backward incompatible anyway.

@cdsmith 

Yes, I see the virtue in your @{m} construct. I'm not excited to have both mechanisms, though. Consider

```hs
f :: forall a b. Bool -> a -> b -> ...
f @b True (x :: b) _ = ...
f @{b} True _ (y :: b) = ...
```

In the two equations, the `b` means something different! The first equation binds `b` to the first type parameter of `f`, while the second equation binds `b` to its second. That's too subtle for my taste. I could imagine something more like `@{m1 = m2}`, where `m1` is the name used in the type signature, and `m2` is the local name bound here. (They, of course, could be the same.) Idris and Agda allow this. But such notation is just as useful at call sites as it is at binding sites, and I'd rather thinking about a holistic approach to named parameters first. Furthermore, I've been thinking for some time now that it's good to reserve `{ blah = wurble }` syntax for anonymous records -- which might exist in types and thus want a `@` prefix.

My bottom line here: yes, I agree that would be nice, but 1) I don't see a good way forward here and 2) the feature you want would be generally useful elsewhere, too, so it probably belongs in a separate proposal.

cdsmith:
> My bottom line here: yes, I agree that would be nice, but 1) I don't see a good way forward here and 2) the feature you want would be generally useful elsewhere, too, so it probably belongs in a separate proposal.

Sure.  I might argue that the similarity between `@b` and `@{b}` isn't really so bad... but that conversation is definitely a but far afield of this proposal's intent.  I wonder if the right answer here is to go ahead and just accept that in the short term, this is going to be a poor fit with ScopedTypeVariables, but that we have some reasonable ideas about what to do to fix that, and then defer that entire discussion (including `@(..)`, which Simon didn't seem too keen on, either) to a later proposal.  In order to accept this proposal, IMO, all that's needed is to conclude that we're unlikely to end up with an indefinite fork.

goldfirere:
I don't see how the proposal, as currently written, introduces even a short-term fork. This version is, I believe, compatible with today's `-XScopedTypeVariables`, with the exception of their use in expression signatures (which should be rare). I like `@(..)` and would want to keep it in, but I'm also OK with losing it, if we agree that such an idea is better kept until later.

Is there a forky aspect of what's here now that I'm missing?

Ericson2314:
BTW, there is a conflict, I think because of the way the proposals were renumbered long ago to patch the PR numbers.

goldfirere:
I think this is ready for review, @nomeata .

nomeata:
I'm currently away from a laptop for 3 weeks, and while I could do the steps by email, it's a bit tedious. Maybe @serras can be going through the motions (label, assign, send mail to list), and either shepherd it yourself, or pick one of your liking.

nomeata:

nomeata:

nomeata:
Obliviated by https://github.com/ghc-proposals/ghc-proposals/pull/448

PR #240, opened 2019-06-10
Compile with threaded RTS by default

ulysses4ever:
Compile with `-threaded` by default. For those in need of the non-threaded RTS, provide the new `-single-threaded` flag.

[Rendered](https://github.com/ulysses4ever/ghc-proposals/blob/master/proposals/0000-threaded-by-default.rst).

[Reddit thread](https://www.reddit.com/r/haskell/comments/byvq5w/ghc_proposal_compile_with_threaded_rts_by_default/).

### DIFF
--- a/proposals/0000-threaded-by-default.rst
+++ b/proposals/0000-threaded-by-default.rst
@@ -19,18 +19,24 @@ Parallel hardware is here. The overwhelming majority of GHC users probably has a
 
 * Blocking FFI calls do not yield to other threads. Common use cases:
 
-  * ``System.Process.waitForProcess`` blocks other threads.
+  * ``System.Process.waitForProcess`` blocks other threads. 
+  
+  This becomes especially hard to debug when used transitevly trough some library (e.g. for `conduit-extra <https://github.com/nh2/sourceProcessWithStreams-nonthreaded-problem>`_).
 
 * Foreign export and foreign import wrapper cannot be called by other threads.
 
 * The IO Manager only exists with ``-threaded``, which means that:
 
-  * Control.Concurrent.STM.registerDelay aborts compilation without ``-threaded``.
+  * ``Control.Concurrent.STM.registerDelay`` aborts compilation without ``-threaded``.
   * Other uses of the event manager, such as ``getSystemEventManager`` fail.
   * ``System.Timeout`` falls back to an inefficient implementation.
   * Parallelism (``+RTS -N`) doesn't work.
-    
-* Not a problem per se, but a matter of consistency: GHCi is built with ``-threaded`` by default, so using the same for binaries by default seems more consistent.
+
+Here is a couple of more points, which cannot be classified as problems per se but still suggest going for the switch:
+
+* GHCi is built with ``-threaded`` by default, so using the same for binaries by default seems more consistent.
+
+* The new I/O manager for Windows (`!1224 <https://gitlab.haskell.org/ghc/ghc/merge_requests/1224>`_) explicitly states that it is meant to be used with the threaded RTS.
 
 The proposal suggests avoiding these subtle problems by defaulting on the threaded RTS, and provide a way to fallback for users in need of single-threaded RTS mode for certain reasons (determinism, debugging, limited hardware, etc.). The latter can be done by the means of the new ``-single-threaded`` flag.
 
@@ -44,9 +50,9 @@ The GHC compiler should start in ``-threaded`` mode by default. To fallback to t
 Effect and Interactions
 -----------------------
 
-The users of GHC will get their programs compiled with the threaded RTS by default. This follows principle of the least surprise: many programs implicitly depending on tools such as `forkIO` can deadlock or crush with no good explanation with the single-threaded RTS. As surprising as it may sound, the threaded RTS comes closer to the principle of the least surprise than the single-threaded RTS.
+The users of GHC will get their programs compiled with the threaded RTS by default. As unexpected as it may sound, the threaded RTS comes closer to the principle of the least surprise than the single-threaded RTS: many programs implicitly depending on tools such as ``forkIO`` can deadlock or crush with no good explanation with the single-threaded RTS. 
 
-One of the main concerns with regards to the threaded RTS is parallel GC, which, anecdotally, almost certainly brings performance losses when not tuned carefully (cf., e.g. `#9221 <https://gitlab.haskell.org/ghc/ghc/issues/9221>`_). It is important to understand that parallel GC **is not implied** by ``-threaded``. The threaded mode still starts the program with one capability by default (unless ``-N`` is explicitly provided by the user), and GC runs sequential.
+One of the main concerns with regards to the threaded RTS is parallel GC, which, anecdotally, almost certainly brings performance losses when not tuned carefully (cf., e.g. `#9221 <https://gitlab.haskell.org/ghc/ghc/issues/9221>`_). It is important to understand that parallel GC **is not implied** by ``-threaded``. The threaded mode still starts the program with one capability by default (unless ``-N`` is *explicitly* provided by the user), and GC runs sequential.
 
 Those who passed ``-threaded`` before will get a new warning, which might be painful (e.g. in the ``-Werror`` setting).
 
@@ -56,7 +62,7 @@ Costs and Drawbacks
 
 The main concern of this change is sudden changes to performance of compiled programs. In some cases it will improve, but in certain cases it may degrade, i.e. programs with no concurrent IO or programs running on single-threaded architectures might observe performance degradation upon updating to a newer GHC, and we should make sure to advertise the option to opt-out (via ``-single-threaded``) in the release notes. Changes in performance shouldn't be significant on a typical architecture given that parallel GC is not implied by ``-threaded``.
 
-Implementation has a very low cost and mostly concern with figuring out necessary adaptations in the GHC test suite.
+Implementation has a very low cost and mostly concerns with figuring out necessary adaptations in the GHC test suite.
 
 
 Alternatives
@@ -64,11 +70,11 @@ Alternatives
 
 One alternative is status quo: default to the non-threaded RTS. This is a plausible option but feels outdated as of now.
 
-Another alternative suggested by Cris Done:
+Another alternative suggested by Chris Done:
 
     GHC could determine at the Core/STG phase whether in the call graph from main, directly or transitively, there was a reference to ``fork#`` and other primops related to threading, and if neither ``-threaded`` nor ``-single-threaded`` was specified, it could warn "Your program may use multi-threaded code, please specify a mode by either: ``-threaded`` or ``-single-threaded``".
 
-This, in fact, does not exclude the option to switch the deafult.
+This, in fact, is orthogonal to switching the deafult, as well as, requiring sugnificant implementation effort, allegedly.
 
 There is also a minor concern about the fallback flag name. Possible options that have been suggested so far are ``-single-threaded`` and ``-non-threaded``.
 
--- a/proposals/0000-threaded-by-default.rst
+++ b/proposals/0000-threaded-by-default.rst
@@ -1,0 +1,78 @@
+Notes on reStructuredText - delete this section before submitting
+==================================================================
+
+The proposals are submitted in reStructuredText format.  To get inline code, enclose text in double backticks, ``like this``.  To get block code, use a double colon and indent by at least one space
+
+::
+
+ like this
+ and
+
+ this too
+
+To get hyperlinks, use backticks, angle brackets, and an underscore `like this <http://www.haskell.org/>`_.
+
+
+Make RTS threaded by default
+==============
+
+.. proposal-number:: <blank>
+.. trac-ticket:: 16126
+.. implemented:: <blank>
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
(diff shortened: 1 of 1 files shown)


guibou:
I'm mitigated about this idea. I'm afraid the surprise performance regression may be important.

Most programs out there does not have any explicit source of parallelism. The only implicit source of parallelism which is available everywhere is the threaded garbage collector. That one is known (at least from my experience) to introduce performance and memory regression when not correctly tuned. (e.g. number of threads, usually less than the number of core. Size of the nursery, ...). So I'm afraid that turning on threaded-rts by default will result in a lot of performance regressions.

This can be mitigated by turning on the threaded-rts, but disabling the parallel GC. This way, no performance change will be observed on non parallel programs.

For now the only benefit I'm seeing of turning `-threaded` on by default is performance improvement / less surprise for programs which exploit explicit parallelism and for which the `-threaded` flags was not turned on. I'll be surprised if there is a lot of them outside.

One positive note however: turning `-threaded` by default may exhibits more the threaded-rts to the real world, resulting in more bug report / use cases, which may or may not results in improvement.

merijn:
I strongly oppose this proposal unless significant benchmarking is done. The proposal mentions:

> In the lack of objective data at hand, we can only speculate that: 1) typical architecture of a GHC user has multiple cores; 2) switch to -threaded will most likely improve performance in such a setting.

Point 2 is wrong. I use -threaded on manycore (24 cores, 48 hyperthreads) machines and some casual benchmarks and day to day experience show that enabling -threaded leads to *crippling* performance regressions, unless you disable parallel GC and tweak the GC a bit. On IRC I've talked with several people who have had the same results.

Not only that, many operations are *substantially* slower in the threaded RTS. We're talking one or more orders of magnitude slowdown of MVar operations, for example. I've got the benchmarking numbers from a library I was working on comparing the threaded and single threaded runtimes here:
    http://files.inconsistent.nl/sync.txt
    http://files.inconsistent.nl/sync-threaded.txt

AntC2:
I think there's a general principle here: default/out-of-the-box GHC (esp GHCi) should be aimed at newbies, to be easy to run without a load of obscure configuration.

It doesn't sound hard to switch on `-threaded` if you know what that involves. Whereas asking a newbie to switch to `-single-threaded` or to understand a load of stuff about threading is just another burden that will turn people off Haskell. (There's already too many gotchas for newbies.)

chrisdone:
I think everyone's at least once accidentally forgot the `-threaded` flag in a project and been confused that a program just hangs because it's running in the unthreaded runtime. That can confuse intermediates and experts, and certainly confuses beginners. Multithreaded code is very common in Haskell projects. 

However, as others have said, there's no evidence that the multithreaded runtime is faster, and no reason to assert that the single threaded runtime should be phased out. The single threaded runtime is useful for two things: (1) speed, (2) determinism. It's valuable. It's not going anywhere.

The annoying problem is that GHC doesn't give you a heads up about it. The setting is dynamic and it should be static. This leads us to having to pick a default.

I've thought about this problem a lot. I think I'd prefer it if GHC could determine at the Core/STG phase whether in the call graph from `main`, directly or transitively, there was a reference to `fork#` and other primops related to threading, and if neither `-threaded` nor `-single-threaded` was specified, it could warn "Your program may use multi-threaded code, please specify a mode by either: -threaded or -single-threaded" (or whatever). I'd prefer that. But it takes more work than just flipping a flag.

nh2:
As @chrisdone says, any program that uses `forkIO` and calls blocking FFI code essentially needs `-threaded` to work correctly as expected (with the two IO actions running concurrently).

It is equally important to allow interrupting IO with functions like `timeout`. The fact that `timeout` works in some cases in the non-threaded RTS is a [lucky accident](https://phabricator.haskell.org/D42#118444).

In general, many programs work on non-threaded on Linux only because of "lucky accidents". On other OSs those break, and once we improve how e.g. the signal works on non-threaded, they will likely also break on Linux.

You can use non-threaded when you have full control over any FFI calls that the program makes. If that is not the case (e.g. you use some library that does IO, and it is reasonable to assume that most programs do that), then `-threaded` is likely a more correct default.

I did observe in multiple projects that non-threaded can give significantly better performance in some cases (up to 2x in my cases). It is also easier to debug most Haskell programs on non-threaded (single-threaded cooperative multitasking programs are always easier to debug than multi-threaded preemptive ones), and it makes many programs more deterministic (`forkIO` based programs that don't depend on time become deterministic under the nonthreaded RTS).

For these reasons, I think that non-threaded should continue to be supported in GHC, and all libraries that can should strive to work very well on non-threaded.

* I am not opposed to `-threaded` being default.
* I would certainly welcome a `-single-threaded` flag to be able to be explicit (independent of which RTS is the default).
* I like @chrisdone's idea with the warning when you weren't explicit.

Finally, note that `-single-threaded` does not mean that the program is really single-threaded!

For example, on OSX, not passing `-threaded` [results in pthreads being used](https://phabricator.haskell.org/D42#128355):

> On OSX, even when using the non-threaded RTS, GHC uses two threads: One for the timer signal, and one for everything else. (I found this a bit surprising given that it's called "non-threaded RTS", but the approach is still sensible and I guess it should just be clearly documented.)

And I would like it to work on Linux this way as well, to remove some of the "lucky accidents" and make non-threaded behave similarly on all the pthreads-supporting platforms, and remove syscall spam.

3noch:
Having been bitten by *both* sides of this (forgetting -threaded in some cases and wishing I hadn't added it in others), I heartily agree that it should be explicit.

andrewthad:
In this context, I think of performance as a secondary concern and behavior as the primary concern. The problem I see with the nonthreaded runtime is that there are several situations where its behavior is inferior to that of the threaded runtime

- If you call a foreign function with the safe FFI and it blocks, the nonthreaded runtime blocks every thread. The threaded runtime doesn't do this. This is documented, but it's a huge behavioral difference that makes certain programs only work correctly with the threaded runtime.
- As a consequent of the first problem, `registerDelay` doesn't work with the nonthreaded runtime (it just crashes with an error telling the user to use the threaded runtime).
- The nonthreaded runtime imposes a file descriptor limit when dealing with event registration. The nonthreaded runtime doesn't do this.

My preference for the threaded runtime aside, I agree with [Chris Done's comment](https://github.com/ghc-proposals/ghc-proposals/pull/240#issuecomment-500412517) that it would be neat to have a way to take advantage of this knowledge at compile time. I would go a step further though. I wish that your hand was just forced at compile time, disallowing switching out the runtimes dynamically. This would probably mean you'd need a different install of GHC for the thread and nonthreaded runtimes, which seems unfortunate. But this way, with the nonthreaded runtime, you'd be compiling against a `base` which simply didn't have functions like `registerDelay`, so you would never get those pesky runtime exceptions.

simonmar:
I'm strongly in favour of this. At work we have `-threaded` baked into the build system and I probably haven't used the single-threaded RTS for years.  There are just too many things that don't work unless you use `-threaded`, which is particularly important for new users: the failure modes are obscure and hard to debug (e.g. deadlocks). We should make things work first, and fast second.

I expect the main source of slowdown that people experience with `-threaded` is doing a lot of context switching between the main thread, which is bound by default, and `forkIO` threads, which are unbound and therefore run in a separate OS thread. In this case, adding `-threaded` causes local context switching on a single OS thread to become context-switching between OS threads, which is orders of magnitude slower.

We might investigate whether changing the main thread to be unbound would be possible.

* You can always get back the current behaviour with `runInBoundThread`
* One fewer performance gotcha
* But, it's a behaviour change that might have weird consequences for code that assumes main is bound.

simonmar:
> Point 2 is wrong. I use -threaded on manycore (24 cores, 48 hyperthreads) machines and some casual benchmarks and day to day experience show that enabling -threaded leads to _crippling_ performance regressions, unless you disable parallel GC and tweak the GC a bit. On IRC I've talked with several people who have had the same results.

So you must have also been using `+RTS -N`, right? The proposal here does not make that the default, only `-threaded`. The cases in which `-threaded` by itself can make things slower are far fewer than the cases in which `+RTS -N` can make things slower.

> Not only that, many operations are _substantially_ slower in the threaded RTS. We're talking one or more orders of magnitude slowdown of MVar operations, for example. 

See my other [comment above](https://github.com/ghc-proposals/ghc-proposals/pull/240#issuecomment-502028699) for the most likely reason for this. If you're seeing orders of magnitude slowdown with `-threaded` and it's *not* due to bound/unbound thread communication, then that's definitely a bug and we should investigate.

cartazio:
Simon: what’s the default selection of N and gc flags for threaded these
days? I know you did some changes in the past year to have N set to the
number of physical cores rather than hypertheads , which corresponds to
half of the historical default (logical core count. ) aka on the 48 core ,
pick 24.

On the other hand, my experience historically has been that there’s usually
diminished/ negative improvement to application if N is set above say
=min(12,number of physical cores ).

Mind you, this was also a parallel data aggregation style computation,
where most memory lives longer so more stuff gets promoted out of nursery.
But it’s definitely true that gc pause frequency is proportional to global
allocation rate, and higher N / number of rts caps can back fire hard

Mind you it’s been at least 3 years since I had to deal with this myself so
my experiences may be out of date.

I believe the objection some folks have isn’t to threaded as such, but
rather that the default setting for N is too aggressive and easily results
in massive perf regressions when an application is moved to a putatively
more performant computer


https://gitlab.haskell.org/ghc/ghc/issues/9221 is a very very relevant
ticket I created 4.5 years ago that’s still relevant and open. And has lots
of interest but is still unaddressed.

If threaded is to be the default, it might be good to clip the default
number of caps when not user toggles to something like the min of (12,
number of physical cpus )

ulysses4ever:
@simonmar @nh2 what would be a good example of a program stuck on `-single-threaded` and fine on `-threaded`? I want to add one in the proposal. I tried something like
```
import Control.Concurrent
import Control.Monad

main = do
    _ <- forkIO $ loop 1
    forever $ 
        print "********** PARENT THREAD *************" >>
        threadDelay 1000000
  where
    loop i = print i >> threadDelay 1000000 >> loop (i+1)
```
I also tried to add `getchar >>` after either `print` (main or child thread): none got stuck in the default RTS, and I was able to put in chars while the other thread was happily printing (GHC 8.6.4, 2 physical core + 2 HT machine).

More generally, would it be fair to say that the new default would likely make for more stable builds **only** for programs that use `Control.Concurrent` or its reverse dependencies? If not, what would be a better way to describe the class of programs potentially benefiting from it? I think something like this should be added to the proposal.

gridaphobe:
Almost all of my code at work would be similarly broken in the non-threaded RTS, specifically due to the lack of `registerDelay`.

simonmar:
Yes, it would be useful to have a list of things that are broken without `-threaded` in the proposal.  Here's a start off the top of my head:

* Blocking FFI calls do not yield to other threads. Common use cases:
  * `System.Process.waitForProcess` blocks other threads
* `foreign export` and `foreign import "wrapper"` cannot be called by other threads
* The IO Manager only exists with `-threaded`, which means that:
  * `Control.Concurrent.STM.registerDelay` fails without `-threaded`
  * Other uses of the event manager, such as `getSystemEventManager` fail
  * `System.Timeout` falls back to an inefficient implementation

Also worth making the point that GHCi is built with `-threaded` by default, so using `-threaded` by default for binaries would be consistent. 
* Parallelism (`+RTS -N`) doesn't work

3noch:
As @chrisdone and @andrewthad have said, a separate and possibly far more useful proposal would be a way to encode, at compile time, the limitations of the runtimes. A CPP flag could come along with `-threaded`. Conditional compilation is a pain, but perhaps a typeclass with selective instances could be layered on top for better ergonomics. If we had that, this proposal would become significantly less important and/or appealing.

cartazio:
One gotcha with the cpp approach is the long term plan to allow later /
launch time  rts flavor / integer impl selection style concerns. So in some
respects it might be closer to how do you juggle runtime per
microarchitecture implementation concerns.

ulysses4ever:
@simonmar I merged your suggestions. If nothing more comes up we should probably submit soon, as the discussion seems to fade.

nh2:
> @nh2 what would be a good example of a program stuck on `-single-threaded` and fine on `-threaded`?

@ulysses4ever https://github.com/nh2/sourceProcessWithStreams-nonthreaded-problem

ulysses4ever:
@nomeata it seems that the discussion has faded. I incorporated some feedback I had here and would like to submit the proposal to the Committee.

simonmar:
Committee shepherd here.

I propose we accept the proposal. Basically the argument comes down to correctness vs. performance: there are a number of correctness gotchas that are fixed by using `-threaded`, while there are some performance gotchas that are avoided (or worked around) by using `-single-threaded`. Changing the default to favor correctness over performance is the right thing.  

Moreover, the performance gotchas with `-threaded` can be avoided while keeping `-threaded` turned on, while the correctness issues with `-single-threaded` don't have workarounds - the only way to use `waitProcess` without blocking all other threads is to add `-threaded`, for example.

simonmar:
Committee discussion has started, I propose acceptance:

https://github.com/ghc-proposals/ghc-proposals/pull/240#issuecomment-509109635

nomeata:
I concur.

bravit:
I agree with the recommendation to accept this proposal.

yav:
Seems reasonable to me.

nomeata:
This seems to be steering towards consensus.

ulysses4ever:
@nomeata is there some kind of timeline here? It's been several weeks since the last bit of Committee discussion. Generally, I think, a clear timeline for the proposal lifecycle would be a good thing to have.

isovector:
I'm pro. Let's merge this.

nomeata:
I think it’s up to the shepherd to maintain the timeline (and maybe I should advertise that we have a reminder bot installed for this repository that people can use). But in general everyone is welcome to nudge and thus share the load.

But in this case we have clear consensus, so I will mark this accepted!

ulysses4ever:
Thanks a lot for everyone contributed!

PR #243, opened 2019-06-20
Stage Hygiene for Template Haskell

Ericson2314:
Template Haskell currently doesn't work well with cross compilation. This is a big nuisance for anyone making phone or browser software. To fix this, enforce stronger separation between the stages.

There are some TODOs, but I am purposefully leaving them there now to be resolved during discussion.

[Rendered](https://github.com/Ericson2314/ghc-proposals/blob/th-stage-hygiene/proposals/0000-th-stage-hygiene.rst)

### DIFF
--- a/proposals/0000-th-stage-hygiene.rst
+++ b/proposals/0000-th-stage-hygiene.rst
@@ -23,7 +23,7 @@ Motivation
 
 Template Haskell currently doesn't work well with cross compilation.
 
-The external interpreter made it at least possible, relatively automatically, but doesn't work if you explicitly want side affects to be run on the compiler's platform.
+The external interpreter made it at least possible, relatively automatically, but doesn't work if you explicitly want side effects to be run on the compiler's platform.
 Also, while same-OS cross can sometimes be fairly lightweight
 —e.g. by having QEMU translate syscalls so the native kernel can be used—
 differnet-OS requires harder to provision virtual machines or real devices.
--- a/proposals/0000-th-stage-hygiene.rst
+++ b/proposals/0000-th-stage-hygiene.rst
@@ -362,14 +362,15 @@ Bindings interleave stages
 Forward references across splices
    The intra-module staging restriction is gone, but that's separate from the prohibition on referencing bindings.
    It just avoids the need to topologically sort splices based on references from the quotations inside them, or break cycles à la ``*.hs-boot``.
-   Nevertheless, implementing that is not trivial so it is good to decouple relaxing the restriction from this already-large proposal.
+   Nevertheless, allowing circular intra-module dependencies is not trivial so it is good to decouple relaxing that restriction from this already-large proposal.
    Hopefully a future proposal will tackle this.
 
 Speeding up builds
-  Modules and libraries are relative in that their exported stage 0 may not be imported at stage 0.
-  There is no notion of a global "true" stage 0.
-  This is good in that we can share build artifacts without breaking abstractions.
-  For example, in the common native case, a library that needs another library in stage 0 and stage -1 can load the *same* build of the library in both of those stages.
+  Modules' and libraries' number of stages is relative.
+  Specifically, their exported stage 0 may not necessarily be imported at stage 0.
+  There is no notion of a global "true" stage 0, which would have to be something the entire dependency graph agrees on.
+  This is good in that we can share build artifacts more widely without breaking abstractions.
+  For example, in the mostly-common native case (build == host), a library that needs another library in stage 0 and stage -1 can load the *same* build of the library in both of those stages.
   By virtue of the explicit stage attached to the import, the definitions do not unify even though the underlying build is the same.
   This can be compared to repeated abstract interfaces in backpack being instantiated with the same concrete module.
 
@@ -391,7 +392,7 @@ Speeding up builds
   To satisfy that, we just `hi` files with `-fexpose-all-unfoldings` file, along with a `"naive" Core interpreter`_ which can evaluate those unfoldings.
   Splices are typically small and numerous, so it seems likely that the lower latency of starting the interpreter is worth the cost of slower evaluation once it is started.
   https://gitlab.haskell.org/ghc/ghc/issues/10871, originally made for Backpack, enshrines `hi` files with `-fexpose-all-unfoldings` as a separate "fat" interface file format.
-  This ideal duel to the "naive" core interpreter to ensure errors are caught as soon as possible.
+  This is an ideal complement to the "naive" core interpreter to ensure we do no more work than necessary.
 
 Template Haskell in GHC
   The motivation evokes the specter of ecosystem splits.
@@ -441,7 +442,7 @@ Costs and Drawbacks
 
 - I don't know of precedent for extensions that prevent modules from being linked together.
 
-- Most existing libraries with commonly used TH helpers (`lens`, `aeson`) have the TH in the same call component but in a different module.
+- Most existing libraries with commonly used TH helpers (`lens`, `aeson`) have the TH in the same Cabal component but in a different module.
   To leverage this proposal, we would have to refactor them to put those modules in a separate library component.
   It would take decent amount of conditional code to still support old GHCs, and even more to not be a breaking change on those old libraries.
 
@@ -451,6 +452,7 @@ Alternatives
 There is no fundamental reason modules couldn't export non-stage-0 items, and libraries expose non-stage-0 modules.
 At the cost of more complexity, there could be a `.lib` or `.so` for each exposed stage, and imports would be offset to match the ``#import <offset>`` literal.
 But in fairness, this might allow a smoother transition form how libraries are structured today.
+Not only would GHC need to learn more tricks, but also Cabal and other tools.
 For example, one could do ``#import 1 Control.Lens.Lens`` in ``Control.Lens.TH`` while exposing ``Control.Lens.TH`` from the same library just like today.
 I decided against this as a matter of taste.
 I think it good to enforce the normal form that the "main" stage is stage 0.
--- a/proposals/0000-th-stage-hygiene.rst
+++ b/proposals/0000-th-stage-hygiene.rst
@@ -623,7 +623,7 @@ Rigid vars
 ~~~~~~~~~~
 
 Why bother with these abstract platforms?
-Why not have something like separate Linux vs Windows ``import``s and ``build-depends``, or separate native vs cross ones?
+Why not have something like separate Linux vs Windows ``import`` and ``build-depends``, or separate native vs cross ones?
 Many systems in fact work this way; I changed Nixpkgs and `Meson`_ to *not* work this way as best as I could.
 The simple answer is there's a combinatorial explosion.
 There's many arches * many OSes * many libcs * many linking strategies that may sadly be observable, etc.
--- a/proposals/0000-th-stage-hygiene.rst
+++ b/proposals/0000-th-stage-hygiene.rst
@@ -98,67 +98,14 @@ It has even been informally proposed for Haskell by @ezyang in `<http://blog.ezy
 Enforcing that separation means restricting programs we currently allow,
 by assigning bindings to stages and restricting what kinds of references between stages are allowed.
 Least surprisingly, normal bindings, and normal imports in the module cannot be used in splices or quotes.
-But there are other constructs that more surprisingly tangle stages too.
-Typed Templated Haskell is one.
-First of all, there is name leakage.
-::
-  [|| ... :: IosOnlyType ||] :: Q (TExp IosOnlyType)
-This can't work unless we are building *on* and *for* iOS.
-Otherwise the ``IosOnlyType`` will be out of scope in one of its two usage sites.
-If we aren't compiling for iOS (iOS is not host OS), then ``IosOnlyType`` is not in scope in the quote.
-If we aren't compiling on iOS (iOS is not the build OS), the ``IosOnlyType`` is not in scope as the argument for ``TExp``.
-The latter one is the show-stopper, presumably we are compiling for ``iOS`` if we want to do this.
-Typed TH in affect assumes that any host type can be mapped back to a build type for sake of the phantom param.
-As shown, this is not always the the case.
-
-But even if we work around that, there's also will be semantic leakage.
-In the near future there would be
-::
-  AppE <$> [|| ... :: foreach (x :: Int) -> F x ||] <*> [|| 2 ^ 36 :: Int ||] :: Q (TExp (F ???))
-How do we type the whole expression, or ``AppE`` in particular?
-``F (2 ^ 36)``?
-But say the platform the compiler runs on (build platform) has a 32-bit ``Int``, while the platform the spliced code runs on (host platform) has a 64-bit ``Int``?
-The code when eventually spliced will have a type of ``F (2 ^ 36)``, but the quote has a type of ``TExp (F 0)``.
-This ruins the guarantees of typed Template Haskell.
-Even today with CPP'd type families:
-::
-  #if mingw_HOST_OS
-  type instance F Bool = []
-  #else
-  type instance F Bool = Tree
-  #endif
-Say we are compiling the following from linux to mingw.
-::
-   AppE <$> [|| ... :: forall a. a -> F a ||] <*> [|| True ||] :: Q (TExp (F Bool))
-We'll have ``F Bool = []`` when the code is eventually spliced, but ``TExp (F Bool) = TExp Tree`` for the quote itself.
-
-Finally, ``Lift`` is problematic for similar reasons.
-Consider
-::
-  lift (linuxOnlyValue)
-This will evaluate through something like
-::
-  lift (LinuxOnlyConstructor arg0 ...argn)
-All good so far.
-But that in turn evaluates as
-::
-  [| LinuxOnlyConstructor $(lift arg0) ...$(lift argn) |]
-If we aren't compiling to ``Linux``, ``LinuxOnlyConstructor`` will be out of scope.
-The overflowing issue doesn't break type safety, but is still nastily non-confluent.
-::
-  lift (2 ^ 25 >= 0) /= [| $(lift $ 2 ^ 25) >= 0 |]
-  ==>
-  [| 0 >= 0 {- on 32-bit build platform -} |] /= [| 1 >= 0 {- on 64-bit host platform -} |]
-There's no non-determinism since ``lift`` doesn't automatically commute like that,
-but the lack of bijectivity is still a foot-gun.
-
-I would love to, instead of outright banning Typed Template Haskell and ``Lift``, come up with a flexible way to associate types and terms between stages.
-For ``Lift``, at a minimum, we just need to map *values* preserving type, though bijectivity is still nice, at least as something one can but manually opts out of.
-For typed Template Haskell, I think we additionally need to map type *expressions* such that evaluation commutes with the mapping.
-Adding language support for such mappings is lots of extra work—borderline research—for a proposal which already is no small task.
+But Typed Templated Haskell and ``Lift`` also entangle stages too, and I propose banning them with this feature for now
+I would love to, instead of outright banning Typed Template Haskell and ``Lift``, come up with more flexible ways to restrict / opt into them,
+But doing that is lots of extra work—borderline research—for a proposal which already is no small task.
 I therefore think banning these constructs for now to start solving the problems people have with cross compilation as soon as possible is prudent.
 Because this proposal is breaking change vs Template Haskell today, a variant extension is used anyways, so no program breaks.
 Instead, users are just temporarily presented with a choice to either support cross compilation or have ``Lift`` and typed TH.
+Remember, this is still strictly better than today when the choice is cross compilation vs all of TH.
+Untyped TH is liberated from the fragmentation, and hopefully the others follow.
 
 As a final side benefit, now that Template Haskell will be defined and implemented in terms of stages, we can relax ``-XTemplateHaskellQuotes``.
 For example, the following current prohibited:
@@ -322,6 +269,66 @@ end user code:
   $(unneededBinding)
 
 A few misc implementation notes:
+
+Banning ``Lift`` and typed TH
+  let's start with the ways typed Templated Haskell entangles the stages.
+  First of all, there is name leakage.
+  ::
+    [|| ... :: IosOnlyType ||] :: Q (TExp IosOnlyType)
+  This can't work unless we are building *on* and *for* iOS.
+  Otherwise the ``IosOnlyType`` will be out of scope in one of its two usage sites.
+  If we aren't compiling for iOS (iOS is not host OS), then ``IosOnlyType`` is not in scope in the quote.
+  If we aren't compiling on iOS (iOS is not the build OS), the ``IosOnlyType`` is not in scope as the argument for ``TExp``.
+  The latter one is the show-stopper, presumably we are compiling for ``iOS`` if we want to do this.
+  Typed TH in affect assumes that any host type can be mapped back to a build type for sake of the phantom param.
+  As shown, this is not always the the case.
+
+  But even if we work around that, there's also will be semantic leakage.
+  In the near future there would be
+  ::
+    AppE <$> [|| ... :: foreach (x :: Int) -> F x ||] <*> [|| 2 ^ 36 :: Int ||] :: Q (TExp (F ???))
+  How do we type the whole expression, or ``AppE`` in particular?
+  ``F (2 ^ 36)``?
+  But say the platform the compiler runs on (build platform) has a 32-bit ``Int``, while the platform the spliced code runs on (host platform) has a 64-bit ``Int``?
+  The code when eventually spliced will have a type of ``F (2 ^ 36)``, but the quote has a type of ``TExp (F 0)``.
+  This ruins the guarantees of typed Template Haskell.
+  Even today with CPP'd type families:
+  ::
+    #if mingw_HOST_OS
+    type instance F Bool = []
+    #else
+    type instance F Bool = Tree
+    #endif
+  Say we are compiling the following from linux to mingw.
+  ::
+     AppE <$> [|| ... :: forall a. a -> F a ||] <*> [|| True ||] :: Q (TExp (F Bool))
+  We'll have ``F Bool = []`` when the code is eventually spliced, but ``TExp (F Bool) = TExp Tree`` for the quote itself.
+
+  Finally, ``Lift`` is problematic for similar reasons.
+  Consider
+  ::
+    lift (linuxOnlyValue)
+  This will evaluate through something like
+  ::
+    lift (LinuxOnlyConstructor arg0 ...argn)
+  All good so far.
+  But that in turn evaluates as
+  ::
+    [| LinuxOnlyConstructor $(lift arg0) ...$(lift argn) |]
+  If we aren't compiling to ``Linux``, ``LinuxOnlyConstructor`` will be out of scope.
+  The overflowing issue doesn't break type safety, but is still nastily non-confluent.
+  ::
+    lift (2 ^ 25 >= 0) /= [| $(lift $ 2 ^ 25) >= 0 |]
+    ==>
+    [| 0 >= 0 {- on 32-bit build platform -} |] /= [| 1 >= 0 {- on 64-bit host platform -} |]
+  There's no non-determinism since ``lift`` doesn't automatically commute like that,
+  but the lack of bijectivity is still a foot-gun.
+
+  The alternative to outright banning these is some sort of flexible way to associate types and terms between stages.
+  For ``Lift``, at a minimum, we just need to map *values* preserving type, though bijectivity is still nice.
+  Perhaps unbijective mappings would take an extra opt-in.
+  For typed Template Haskell, I think we additionally need to map type *expressions* such that evaluation commutes with the mapping.
+  The type for ``AppE`` uses stage n rather than stage n + 1 (type) application, so we can't just concern ourselves with the mapping of type values.
 
 Relaxing the stage restriction is hygiene at work
    We can fearlessly interpret all n - 1 code to fill in splices in stage n without the risk of encountering splices that depend on themselves.
--- a/proposals/0000-th-stage-hygiene.rst
+++ b/proposals/0000-th-stage-hygiene.rst
@@ -1,4 +1,4 @@
-Stage Hygiene in Template Haskell
+Stage Hygiene for Template Haskell
 ==============
 
 .. proposal-number:: Leave blank. This will be filled in when the proposal is
--- a/proposals/0000-th-stage-hygiene.rst
+++ b/proposals/0000-th-stage-hygiene.rst
@@ -260,7 +260,9 @@ Cabal
    N.B ``build-tool-depends`` can be thought of as a stage -1 executable dependencies list.
    "Those executables are executed at build time, like top-evel splices, and so need to be built for the build platform."
    `<https://github.com/haskell/cabal/issues/5411>`_ asks for a ``run-tool-depends`` which would be nothing but a stage 0 executable depends.
-   ``setup-depends`` can also be thought of as a stage -1 executable dependencies list.
+   ``setup-depends`` can also be thought of as a stage -1 dependencies list.
+   In both cases though, the metaphor is not perfect because separate executable are only loosely coupled.
+   E.g libraries for Template-Haskell must be built with the same compiler / same ABI, but executable deps (including `./Setup` itself), need not be.
 
 #. Connect today's "qualified goals" to stages.
    [TODO exact formalism, is it in scope?]
--- a/proposals/0000-th-stage-hygiene.rst
+++ b/proposals/0000-th-stage-hygiene.rst
@@ -191,15 +191,16 @@ GHC
 #. Let there be a notion of stages assigned to the integers.
    All existing rules outside of TH on binding/name resolution are retaken to act independently per stage.
    (i.e. identifiers in stage *n* resolve to bindings in stage *n*, all syntax in the rule is parameterized with the stage.)
-   The top level is always stage 0.
-   A consequence of the above is all non-TH syntax in is also stage 0.
+   bindings (with existing, regular syntax) on the top level are always in stage 0.
+   As a consequence, all non-TH syntax in is also stage 0.
 
-#. Redefine quoting and splicing as acting on adjacent stages. Specifically, quoting quotes code from the next stage:
+#. Redefine quoting and splicing as acting on adjacent stages.
+   Specifically, quoted code from the next stage:
    ::
      G ⊢(n + 1) syntax
      -----------------------
      G ⊢(n) [| syntax |]
-   and splicing splices code from the previous stage:
+   and spliced code from the previous stage:
    ::
      G ⊢(n - 1) syntax
      -----------------------
@@ -213,24 +214,27 @@ GHC
    This means import a module in stage *n* instead of stage 0 as per normal.
    ::
      <decl> ::= $let <integer-literal> <<existing syntax>> = <<existing syntax>>
+   This means bind identifiers in stage *n* instead of stage 0 as per normal.
    In both case the ``$`` must not be followed by whitespace, both to avoid conflicts with other syntax and to be consistent with splices.
-   The means bind identifiers in stage *n* instead of stage 0 as per normal.
-   Module exports however are restricted to stage 0.
+
+#. Module exports, however, are restricted to stage 0.
+   There is no syntax analogous to that of definitions and imports to overcome what is for them merely a default of stage 0.
 
 #. The current "stage restriction" on splices using bindings from the current module is abolished.
-   Any stage n + 1 binding in a stage n splice is fair game.
+   Any stage n - 1 binding in a stage n splice is fair game.
 
-#. Relax ``-XTemplateHaskellQuotes`` to instead allow Template Haskell constructs, but restrict their usage so all syntax is in stages >= 0.
+#. Relax ``-XTemplateHaskellQuotes`` to instead allow splices, but restrict their usage so all syntax is in stages >= 0.
 
 #. Introduce ``-XTemplateStagePersistence``.
    Which is implied by ``-XTemplateHaskellQuotes`` (and thus plain ``-XTemplateHaskell``) for backwards compatibility.
    It allows the current behavior where we blur the distinction between stages.
    In particular, with `TemplateStagePersistence` enabled:
 
-   - Stage 0 identifiers bound in another module can be used in stage 1 (quotes) and stage -1 (splices).
-   - Stage 0 identifiers bound anywhere can be used in stage 1, and are automatically.
+   - Stage 0 identifiers bound in another module can be used in stage -1 (splices).
+   - Stage 0 identifiers bound at the top level can be used "by reference" in stage 1.
    - Typed template haskell is allowed.
    - The ``Lift`` type class and all its associated definitions are made available.
+   - Stage 0 identifiers bound anywhere can be used "by value" in stage 1, via an implicit ``lift``.
 
    These are always permitted today.
    But with ``-XNoTemplateStagePersistence``, overriding the default, all of those are *disabled*.
@@ -318,6 +322,17 @@ end user code:
   $(unneededBinding)
 
 A few misc implementation notes:
+
+Relaxing the stage restriction is hygiene at work
+   We can fearlessly interpret all n - 1 code to fill in splices in stage n without the risk of encountering splices that depend on themselves.
+   The stages enforce a guardedness condition.
+   Inter-module infinite stages are still possible via e.g. a library that depends on itself in stage -1, but Cabal catches that rather than GHC.
+
+"True" splices vs splices within quotes
+  The new rules for ``-XTemplateHaskellQuotes`` instead require that "all syntax is in stages >= 0".
+  This means every splice is within a quote.
+  Those nested splices effectively cancel out with their parent quote.
+  Splices from stages <= 0 (i.e. caused by syntax in stages < 0) are the "true" splices which actually force evaluation.
 
 Bindings interleave stages
   Note that ``$let`` can appear outside the top-level, including in contexts where a variable of later stage is bound.
--- a/proposals/0000-th-stage-hygiene.rst
+++ b/proposals/0000-th-stage-hygiene.rst
@@ -9,9 +9,7 @@ Stage Hygiene for Template Haskell
 .. implemented:: Leave blank. This will be filled in with the first GHC version which
                  implements the described feature.
 .. highlight:: haskell
-.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
-            **After creating the pull request, edit this file again, update the
-            number in the link, and delete this bold sentence.**
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/243`_.
 .. sectnum::
 .. contents::
 
--- a/proposals/0000-th-stage-hygiene.rst
+++ b/proposals/0000-th-stage-hygiene.rst
@@ -543,6 +543,16 @@ I think it good to enforce the normal form that the "main" stage is stage 0.
 As to the specific example, I would rather packages leverage public Cabal sub-libraries for Template Haskell anyways;
 I think that's a cleaner way to package code.
 
+It would be nice to go straight implement typed TH by quoting types.
+One would do something like::
+  [|| e... :: t... ||] :: Q (TExp [t| t... |])
+This correctly gets ``t...`` from the right for the argument to ``TExp``.
+The problem though is the parameter is type syntax rather than a type; we've thrown away all the semantics.
+We can explicitly normalize, however::
+  [|| e... :: t... ||] :: Q (TExp (Normalize [t| t... |])
+``Normalize`` could be some sort of magic type family that just tells GHC to do its thing, rather than a implementation of Haskell.
+It's good to know there are options going forward, but I think there is just too much uncertainty here to commit to anything at this time.
+
 Unresolved Questions
 --------------------
 
--- a/proposals/0000-th-stage-hygiene.rst
+++ b/proposals/0000-th-stage-hygiene.rst
@@ -26,7 +26,7 @@ Haskell has an excellent culture of code reuse, but that means that arbitrary so
 That lack of being able to write idiomatic Haskell code with popular libraries for all platform is as big of a loss as not being able to Template Haskell itself.
 The risk of library fragmentation between between those writing code for weird platforms and those not is also bad for everyone.
 
-The external interpreter made it at least possible, relatively automatically, but doesn't work if you explicitly want side effects to be run on the compiler's platform.
+The external interpreter made it at least possible, relatively automatically, but doesn't work if you explicitly want side effects to be run on the platform doing the compiling (build platform).
 Also, while same-OS cross can sometimes be fairly lightweight
 — e.g. by having QEMU translate syscalls so the native kernel can be used —
 different-OS requires harder to provision virtual machines or real devices.
@@ -75,7 +75,7 @@ In the near future there would be
 ::
   AppE <$> [|| ... :: foreach (x :: Int) -> F x ||] <*> [|| 2^36 :: Int ||] :: Q (TExp (F ???))
 How do we type the whole expression, or ``AppE`` in particular?
-And say the compiling platform has 32-bit `Int`s?
+And say the platform the compiler runs on (build platform) has 32-bit `Int`s?
 The dependent function will have different result types due to overflow, which ruins the guarantees of typed Template Haskell.
 Even today we have similar problems with CPP'd type families:
 ::
@@ -285,7 +285,7 @@ Speeding up builds
   By virtue of the explicit stage attached to the import, the definitions do not unify even though the underlying build is the same.
   This can be compared to repeated abstract interfaces in backpack being instantiated with the same concrete module.
 
-  In the cross case, there is no getting around needing separate native and foreign builds for different stages, but there are still performance improvements.
+  In the cross case, there is no getting around needing separate builds for the different platform used in each stage, but there are still performance improvements.
   As said in the motivation, we only need what is needed when it is needed, versus everything twice with splice dumping and loading.
   This reduces the size and improves the parallelism of the build plan.
   More subtly, and perhaps more importantly, are benefits with rebuilds during development.
--- a/proposals/0000-th-stage-hygiene.rst
+++ b/proposals/0000-th-stage-hygiene.rst
@@ -20,6 +20,9 @@ To fix this, enforce stronger separation between the stages.
 Motivation
 ------------
 
+Hacking TH today
+~~~~~~~~~~~~~~~~
+
 In the olden days, one couldn't use Template Haskell with cross compilation at all.
 This is bad in itself, but also bad from an ecosystem perspective.
 Haskell has an excellent culture of code reuse, but that means that arbitrary software is very likely to depend on Template Haskell.
@@ -33,7 +36,7 @@ different-OS requires harder to provision virtual machines or real devices.
 
 Another alternative is dumping and loading splices, where one builds natively, dumping splices, and the builds cross with those dumped splices.
 This became easier with `this patch <https://github.com/reflex-frp/reflex-platform/blob/master/splices-load-save.patch>`_.
-Still, this requires building every package twice, and worse doesn't work if the macro is target specific.
+Still, this requires building every package twice, since we redo the entire compilation on both platforms, and worse doesn't work if a top-level splice is target specific.
 For example, imagine some code like
 ::
   #ifdef ios_HOST_OS
@@ -41,20 +44,48 @@ For example, imagine some code like
   $(iosBoilerplateHelper ''SomeIosFfiType)
   #endif
 If the splice is within the ``ifdef``, it won't be dumped.
-But if it is outside the ``ifdef``, the native one won't build!
+When we compile to dump splices, compilation occurs on the native platform, and so the splice will be removed at preprocessing time before dumping.
+And deleting the CPP is no quick fix — if it is outside the ``ifdef``, the native one won't build!
+If it is outside the ``ifdef``, then when we compile to dump splices we will actually fail, since we're now trying to build ios-specific code on the wrong platform.
 
-What we need instead is a way to say is different platforms:
+What we need instead is a way to say different code runs on different platforms.
 
-- Splices alone run on the build platform (of the module being built, not GHC).
+Referencing platforms
+~~~~~~~~~~~~~~~~~~~~~
+
+First, an aside about naming platforms.
+Long ago, the autoconf invented the terms "build" platform, "host" platform, and "target" platform:
+
+- the build platform of some code is where it is built.
+- the host platform of some code is where it runs.
+- the target platform of some code is the host platform of code produced by this code.
+
+Yes, the names are vague, and yes, the naming of one inductive step (target) is awkward, but the beauty of these names is they work for native and cross compilation alike.
+Rather than thinking about concrete platforms and where they are used, they turn the problem around by thinking about the holes where concrete platforms go in.
+These "abstract" platforms are thus parameters.
+And whereas most designs are native only, and then hurridly retrofitted for cross, these names naturally lead to not assuming any of the 3 platforms are the same.
+In other words, native compilation is the special case, not cross.
+
+Explicit Staging
+~~~~~~~~~~~~~~~~
+
+With that in mind, what we are looking for is:
+
+- Splices alone run on the build platform.
 - Normal code, as usual, runs on the host platform.
 - Quoted code runs on the target platform.
 
-This solves all the above problems:
+This solves all the problems of the first section:
 
-- No need to emulate any other platforms, since evaluation only happens within splices.
+- No need to emulate any other platforms.
+  Recall TH-induced evaluation only happens within top-level splices; splices within brackets just build bigger expressions.
+  That means only TH only induces build platform splicing, which is native by defintion!
+
 - No need to build everything twice.
-  Just what is needed in each phase, and just when it's needed.
-- No all-or-nothing CPP problem.
+  Just what is needed in each phase is built, and just when it's needed.
+
+- No risk of CPPing away the splice, because we aren't faking it with build == host native builds.
+  ``ios_HOST_OS`` is independent of any ``*_BUILD_OS`` macro.
 
 To do this, we need to cleanly separate the stages induced by quoting and splicing.
 This is not a new idea for programming languages in general.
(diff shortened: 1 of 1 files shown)


proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,369 @@
...
+
+   These are all already always permitted today.
+   But with ``-XNoTemplateStagePersistence``, overriding the default, all of those are *disabled*.
+
+#. Extend the command line [TODO bikeshed!!] with a way to specify per-stage package dependencies and the like.
+   If/when GHC becomes multi-target, by default stages >= 0 take GHC's target platform / the packages host platform (where compiled code runs), while stages < 0 take GHC's host platform / the packages build platform (where GHC runs).
+   But, the emitted platform can still be specified per-stage like the other flags.
+   This is needed when building TH functions to be used from cross compiled code.
+
+#. Add a Core "way" to GHC, which basically amounts to `-fexpose-all-unfoldings` and doing no work no past core [TODO bikeshed/clarify].
```
michaelpj: This sounds like another use for https://gitlab.haskell.org/ghc/ghc/issues/10871.

Ericson2314: Awesome!! Commented on that issue, and will incorporate it into this.

Ericson2314: Updated the text. Hope it looks good now re this, and is more honest about what needs to be built as we discussed on IRC!

mpickering:
There are some good ideas in this proposal and I would like to work with you on getting this proposal into shape. At the moment the key ideas are a bit lost in the weeds so it would be good to distinguish the fundamental from the accidental.

goldfirere:
I have not thought deeply about the cross-compilation problem for TH, but this seems plausible. I'm broadly in support of something like this, as long as it doesn't make TH harder to use for those not worried about cross-compilation (unless there's a very compelling reason to burden us). The motivation section of the proposal is compelling to me.

michaelpj:
Do we need to expose the full complexity of stage offsets in user-visible syntax? This comes up in a couple of places:
- In source Haskell you propose stage-offset imports and bindings.
- In Cabal you propose stage-offset dependencies.

However the other existing systems seem like they can get away with only exposing *two* phases to the user:
- In Racket, you have `define-for-syntax` and `import-for-syntax`, but nothing more.
- In other systems that handle cross compilation you have "native build dependencies" and nothing more.

I think both of these correspond to exposing the -1 and 0 stages and no others, and they have the advantage of being relatively easy to describe, whereas I am worried that the full stage offsetting machinery will be hard for people to understand (I certainly found it hard to understand!).

While it might make sense for the underlying system to know about phases in full generality, I'd be interested in how much expressiveness we would lose if we were to restrict the surface syntax (at least initially - we could add a fuller version later if we needed it).

For example, I think it would make it impossible to do anything too fancy *within* a single package (component?), but this might be an acceptable price to pay.

Ericson2314:
@mpickering has talked to me quite a lot already, and I would like to thank him for putting up with me stumbling through a rather large backlog of hitherto unexpressed ideas.

@goldfirere 
> as long as it doesn't make TH harder to use for those not worried about cross-compilation (unless there's a very compelling reason to burden us)

You are on the money. I am definitely imposing a burden, but I hope to justify it. 

@michaelpj

> In Racket, you have `define-for-syntax` and `import-for-syntax`, but nothing more.

Ah but see https://docs.racket-lang.org/guide/phases.html. There is a `for-meta` with an explicit phase offset just like here! Heh, I haven't looked at the Racket docs in a while....turns out they count the phases in the other direction. oops!

> While it might make sense for the underlying system to know about phases in full generality, I'd be interested in how much expressiveness we would lose if we were to restrict the surface syntax (at least initially - we could add a fuller version later if we needed it).

So do note that I restrict the user to stages -1, 0, and 1 in a single module. This correspond's to Racket's `for-syntax`, `for-meta`, and `for-template`. As you say, it is important that the overall system understand things more broadly since a composition of modules via offset imports can generate more stages than those 3. I even agree 2 at a time is most likely...except sometimes it's -1 and 0, and sometimes it's 0 and 1.

Ericson2314:
@angerman I think this is our best bet for a relocatable GHC long term. What's cool is if this is accepted and implemented (along with the naive core interpreter) we can use it immediately within GHC: stage 1 can always use TH, and stage 2+ is always built with a new GHC.

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+-------------------
+
+- This is a huge amount of work.
+  But I am fine chipping away it over a long period of time.
+
+- Even a temporary conflict between typed TH and this could slow typed TH's adoption.
+
+- I don't know of precedent for extensions that prevent modules from being linked together.
+
+- Most existing libraries with commonly used TH helpers (`lens`, `aeson`) have the TH in the same call component but in a different module.
```
michaelpj: What is a "call component"?

michaelpj: Cabal component, maybe?

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+  It would take decent amount of conditional code to still support old GHCs, and even more to not be a breaking change on those old libraries.
+
+Alternatives
+------------
+
+There is no fundamental reason modules couldn't export non-stage-0 items, and libraries expose non-stage-0 modules.
+At the cost of more complexity, there could be a `.lib` or `.so` for each exposed stage, and imports would be offset to match the ``#import <offset>`` literal.
+But in fairness, this might allow a smoother transition form how libraries are structured today.
+For example, one could do ``#import 1 Control.Lens.Lens`` in ``Control.Lens.TH`` while exposing ``Control.Lens.TH`` from the same library just like today.
+I decided against this as a matter of taste.
```
michaelpj: I think this also makes cabal's job easier.

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+
+- Even a temporary conflict between typed TH and this could slow typed TH's adoption.
+
+- I don't know of precedent for extensions that prevent modules from being linked together.
+
+- Most existing libraries with commonly used TH helpers (`lens`, `aeson`) have the TH in the same call component but in a different module.
+  To leverage this proposal, we would have to refactor them to put those modules in a separate library component.
+  It would take decent amount of conditional code to still support old GHCs, and even more to not be a breaking change on those old libraries.
+
+Alternatives
```
michaelpj: Are there any alternatives that give us the benefit without introducing this infinite set of stages? Or without exposing them as much to the user? I can see stages have significant precedent, but I'd be interested to hear if there are any other approaches.

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+   Nevertheless, implementing that is not trivial so it is good to decouple relaxing the restriction from this already-large proposal.
+   Hopefully a future proposal will tackle this.
+
+Speeding up builds
+  Modules and libraries are relative in that their exported stage 0 may not be imported at stage 0.
+  There is no notion of a global "true" stage 0.
+  This is good in that we can share build artifacts without breaking abstractions.
+  For example, in the common native case, a library that needs another library in stage 0 and stage -1 can load the *same* build of the library in both of those stages.
+  By virtue of explicit stage of the import, the definitions do not unify even though the underlying build is the same.
+  This can be compared to repeated abstract interfaces in backpack being instantiated with the same concrete module.
```
michaelpj: Possibly silly question: could we implement it in exactly that way? i.e. use the machinery we have for Backpack to handle the multiple "stage instantiations" of a module?

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+Motivation
+------------
+
+In the olden days, one couldn't use Template Haskell with cross compilation at all.
+This is bad in itself, but also bad from an ecosystem perspective.
+Haskell has an excellent culture of code reuse, but that means that arbitrary software is very likely to depend on Template Haskell.
+That lack of being able to write idiomatic Haskell code with popular libraries for all platform is as big of a loss as not being able to Template Haskell itself.
+The risk of library fragmentation between between those writing code for weird platforms and those not is also bad for everyone.
+
+The external interpreter made it at least possible, relatively automatically, but doesn't work if you explicitly want side effects to be run on the compiler's platform.
```
michaelpj: You only mention TH's side-effects here, afaict. That can matter a lot. I assume the idea is that side-effects will run where they are compiled, so that code which is compiled for the host will run on the host etc?

An extension which I have considered for a while (which would be possible following https://github.com/ghc-proposals/ghc-proposals/pull/246) would be to have something like `PureTemplateHaskell`, which would restrict the type of splices to be `NameSupply m => ...` rather than `Quasi m => ...` i.e. ban IO from quotes. This would play quite nicely with this proposal, I think.

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+For example, imagine some code like
+::
+  #ifdef ios_HOST_OS
+  data SomeIosFfiType
+  $(iosBoilerplateHelper ''SomeIosFfiType)
+  #endif
+If the splice is within the ``ifdef``, it won't be dumped.
+But if it is outside the ``ifdef``, the native one won't build!
+
+What we need instead is a way to say is different platforms:
```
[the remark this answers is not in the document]
Ericson2314: Agreed

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+To do this, we need to cleanly separate the stages induced by quoting and splicing.
+This is not a new idea for programming languages in general.
+Racket (and probably some schemes) do this.
+The work-in-progress (?) `OCaml macro system <https://github.com/ocamllabs/ocaml-macros>`_ does this.
+It has even been informally proposed for Haskell by @ezyang in `<http://blog.ezyang.com/2016/07/what-template-haskell-gets-wrong-and-racket-gets-right/>`_.
+
+Enforcing that separation means restricting programs we currently allow.
+Least surprisingly, normal bindings, and normal imports, in the module cannot be used in splices or quotes.
+But there are other constructs that more surprisingly tangle stages too.
+Typed Templated Haskell is one.
```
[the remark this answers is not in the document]
Ericson2314: Good idea.

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+Haskell has an excellent culture of code reuse, but that means that arbitrary software is very likely to depend on Template Haskell.
+That lack of being able to write idiomatic Haskell code with popular libraries for all platform is as big of a loss as not being able to Template Haskell itself.
+The risk of library fragmentation between between those writing code for weird platforms and those not is also bad for everyone.
+
+The external interpreter made it at least possible, relatively automatically, but doesn't work if you explicitly want side effects to be run on the compiler's platform.
+Also, while same-OS cross can sometimes be fairly lightweight
+— e.g. by having QEMU translate syscalls so the native kernel can be used —
+differnet-OS requires harder to provision virtual machines or real devices.
+
+Another alternative is dumping and loading splices, where one builds natively, dumping splices, and the builds cross with those dumped splices.
```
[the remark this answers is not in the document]
Ericson2314: I think "native" is useful for saying "I can run the thing I built", but I added a build = host paranthetical.

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+- Splices alone run on the build platform (of the module being built, not GHC).
+- Normal code, as usual, runs on the host platform.
+- quoted code runs on the target platform.
+
+This solves all the above problems:
+
+- No need to emulate any other platforms, since evaluation only happens within splices.
+- No need to build everything twice.
+  Just what is needed in each phase, and just when it's needed.
+- No all-or-nothing CPP problem.
```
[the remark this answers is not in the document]
Ericson2314: I switched this to prose, but let me know if we still want back references, say to the dumping and loading paragraph.

proposals/0000-th-stage-hygiene.rst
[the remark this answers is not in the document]
Ericson2314: Well actually, this "things we need to disable" is the last bit of the motivation before the concluding thought. Maybe that is OK?

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+``Lift`` is similarly problematic.
+::
+  lift (.. :: LinuxOnlyType)
+This relies on native compilation to Linux or a scoping violation must also be induced.
+::
+  lift (2^^25)
+This though is fine as regardless of overflowing on either side an ``Int`` can be kept an ``Int``, and overflowing is already defined behavior.
+
+I would love to, instead of outright banning Typed Template Haskell and ``Lift``, come up with a flexible way to associate types and terms between stages.
+To be "complete" in that module scoping everything is complete is still possible, we would need slightly different requirements for each.
```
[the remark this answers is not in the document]
Ericson2314: Sheesh it doesn't make sense to me either now!

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+For ``Lift`` we just need to map *values* preserving type, while typed Template Haskell we need to map type *expressions* such that evaluation commutes with the mapping.
```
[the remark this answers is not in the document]
Ericson2314: I elaborated the previous examples, which hopefully adds some context. This is just a conjuncture---if I knew exactly what to do I might propose doing it! Couched my language to admit as much.

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+Adding language support for such a mapping is lots of extra work—borderline research—for a proposal which already is no small task.
+I therefore think banning for now to start solving the problems people have with cross compilation as soon as possible is prudent.
+Because this is breaking change no matter what, a variant extension is used anyways, so no program breaks.
+Instead, users a just temporarily presented with a choice to either support cross compilation or have ``Lift`` and typed TH.
```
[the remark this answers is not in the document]
Ericson2314: I hope this is only temporary, but I think it's still an improvement in the short term. There's already fragmentation, and this makes untyped TH fine, leaving a smaller fragmentation of just lift/typed TH vs cross, not all of TH vs cross.

Remember that functions dealing with quotes defined this style can still be imported from regular TH.

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+------------
+
+GHC
+~~~~~~~~~~~~
+
+#. Let there be a notion of stages assigned to the integers.
+   All existing rules outside of TH on binding/name resolution are retaken to act independently per stage.
+   (i.e. identifiers in stage *n* resolve to bindings in stage *n*, all syntax in the rule is parameterized with the stage.)
+   The top level is always stage 0.
+   A consequence of the above is all non-TH syntax in is also stage 0.
```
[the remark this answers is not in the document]
Ericson2314: The "All existing rules outside of TH" means that from the top-level binds, everything else is constrained to also be stage 0.

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+Proposed Change Specification
+------------
+
+GHC
+~~~~~~~~~~~~
+
+#. Let there be a notion of stages assigned to the integers.
+   All existing rules outside of TH on binding/name resolution are retaken to act independently per stage.
+   (i.e. identifiers in stage *n* resolve to bindings in stage *n*, all syntax in the rule is parameterized with the stage.)
+   The top level is always stage 0.
```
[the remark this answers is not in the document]
Ericson2314: You can think of this as sort of a "in the beginning before TH, there were stages but everything was stage 0".

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+   A consequence of the above is all non-TH syntax in is also stage 0.
+
+#. Redefine quoting and splicing as acting on adjacent stages. Specifically, quoting quotes code from the next stage:
```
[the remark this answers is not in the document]
Ericson2314: oops that's me waffling over tense, screw those two I'll actually go with past tense!

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+
+#. Relax ``-XTemplateHaskellQuotes`` to instead allow Template Haskell constructs, but restrict their usage so all syntax is in stages >= 0.
+
+#. Introduce ``-XTemplateStagePersistence``.
+   Which is implied by ``-XTemplateHaskellQuotes`` (and thus plain ``-XTemplateHaskell``) for backwards compatibility.
+   It allows the current behavior where we blur the distinction between stages.
+   In particular, with this enabled:
+
+   - Stage 0 identifiers bound in another module can be used in stage 1 (quotes) and stage -1 (splices).
+   - Stage 0 identifiers bound anywhere can be used in stage 1, and are automatically.
```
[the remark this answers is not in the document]
Ericson2314: I meant to call out implicit lifts, but yeah I didn't do it right and there is overlap.

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+     <impdecl> ::= $import <integer-literal> <<existing syntax>>
+   This means import a module in stage *n* instead of stage 0 as per normal.
+   ::
+     <decl> ::= $let <integer-literal> <<existing syntax>> = <<existing syntax>>
+   In both case the ``$`` must not be followed by whitespace, both to avoid conflicts with other syntax and to be consistent with splices.
+   The means bind identifiers in stage *n* instead of stage 0 as per normal.
+   Module exports however are restricted to stage 0.
+
+#. The current "stage restriction" on splices using items from module is abolished.
+   Any stage n + 1 binding in a stage n splice is fair game.
```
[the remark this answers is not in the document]
Ericson2314: I will put in effects and interactions. This is huge proposal so I want to keep the normative part very terse.

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+
+   - Stage 0 identifiers bound in another module can be used in stage 1 (quotes) and stage -1 (splices).
+   - Stage 0 identifiers bound anywhere can be used in stage 1, and are automatically.
+   - Typed template haskell is allowed.
+   - The ``Lift`` type class and all its associated definitions are made available.
+
+   These are all already always permitted today.
+   But with ``-XNoTemplateStagePersistence``, overriding the default, all of those are *disabled*.
+
+#. Extend the command line [TODO bikeshed!!] with a way to specify per-stage package dependencies and the like.
```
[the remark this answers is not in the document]
Ericson2314: Different splices have different dependencies. If this weren't the case, there would be no finite build plans!

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+   If/when GHC becomes multi-target, by default stages >= 0 take GHC's target platform / the packages host platform (where compiled code runs), while stages < 0 take GHC's host platform / the packages build platform (where GHC runs).
+   But, the emitted platform can still be specified per-stage like the other flags.
+   This is needed when building TH functions to be used from cross compiled code.
+
+#. When importing modules/packages, after applying the import offset ensure that the platforms match.
+   Note that while each module only has bindings in its own stage 0, those bindings can contain quotes from stages greater than 0.
+   All such quoted platforms need to match.
```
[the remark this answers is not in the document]
Ericson2314: I made the language not suck, but it's still terse, let me know if you want more in effects/interactions.

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+
+#. Just as GHC defines ``*_HOST_OS`` and similar CPP identifiers today, it would define ``*_BUILD_*`` ones if you have any stage -1 package imports, and ``*_BUILD_*`` if you have any stage 1 package imports.
```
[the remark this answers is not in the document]
Ericson2314: So as not to needlessly invalidate builds where this stuff shouldn't be observed; I'm thinking about real or maybe hypothetical incremental build systems. Added a little not here within my terse budget :).

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+
+Cabal
+~~~~~~~~~~~~
+
+#. Extend the ``build-depends`` syntax with a stage integer offset parameter.
+   N.B ``build-tool-depends`` can be thought of as a stage -1 executable dependencies list.
+   `<https://github.com/haskell/cabal/issues/5411>`_ asks for a ``run-tool-depends`` which would be nothing but a stage 0 executable depends.
+   ``setup-depends`` can also be thought of as a stage -1 executable dependencies list.
+
+#. Likewise extend ``other-modules`` with a stage integer offset parameter, to support intra-package ``$import``.
```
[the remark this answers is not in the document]
Ericson2314: You can have a -1 `other-module` which is just used for a splice, rather than exported. But yeah I am not wild about this feature and wouldn't mind dropping it.

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+   Leave ``exposed-modules`` as is, however. Libraries should only expose stage 0 modules, just as modules only expose stage 0 definitions.
+   Restrict the ``other-modules`` offset to be <= 0, as positive stage code is either pointless or would escape via references from quotes causing build system havoc.
```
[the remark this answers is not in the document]
Ericson2314: I realized one would could use stage 1+ files for quotes, and those would need to be installed. I'm just going to nix this feature. Just use another component.

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+   Unexposed negative stage modules need not be installed at all, as there is no way for stage 0 to reference them (splices eliminate references).
```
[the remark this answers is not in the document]
Ericson2314: Just skipping this feature.

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+
+A few misc implementation notes:
+
+Bindings interleave stages
+  Note that ``$let`` can appear outside the top-level, including in contexts where a variable of later stage is bound.
+  At first glance, binding a compile-time variable within a run-time variable might seem like a staging violation:
+  ::
+    f x = $huh
+       where foo = ...
+             $let -1 huh ... = ... [| x |] ... [| foo |] ...
```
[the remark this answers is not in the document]
Ericson2314: See "Bindings interleave stages" in effects and interactions.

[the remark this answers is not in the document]
Ericson2314: Yeah it is part of the `where`.

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+  But remember that later stage syntax can just be used in quotes; it is inert and cannot be evaluated.
+  ``huh`` is trivially lifted outside of ``f`` since captures the syntactic ``x`` which is static at compile-time.
```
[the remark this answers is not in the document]
Ericson2314: Exactly.

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,397 @@
...
+  This is comparable to today's trick of renaming `containers` to `kontainers` so we can tune it and re-benchmark without rebuilding criterion and friends.
+
+  There are *still* more tricks we can do for overall build size and parallelism.
+  Stage 1 code doesn't need to be evaluated, just composed correctly.
+  As such, we just need the interface of imports, and don't care about the definitions behind those declarations.
+  That means we just need to build as far as today's `hi` files to resolve those imports.
+  Stage -1 code does need to be run, but still not compiled in the final binary since it cannot be exported.
+  To satisfy that, we just `hi` files with `-fexpose-all-unfoldings` file, along with a `"naive" Core interpreter`_ which can evaluate those unfoldings.
+  Splices are typically small and numerous, so it seems likely that the lower latency of starting the interpreter is worth the cost of slower evaluation once it is started.
+  https://gitlab.haskell.org/ghc/ghc/issues/10871, originally made for Backpack, enshrines `hi` files with `-fexpose-all-unfoldings` as a separate "fat" interface file format.
```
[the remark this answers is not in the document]
Ericson2314: Yeah we're all good here.

proposals/0000-th-stage-hygiene.rst
Ericson2314: Good point.

proposals/0000-th-stage-hygiene.rst
Ericson2314: I in fact restrict to stage -1, 0, 1 in the proposed change specification. But (perhaps confusingly) all the design components assumes the full integers to hold them to a higher standard. I don't want anything to be allowed to be sloppy or ad-hoc by doing a 3 stage case analysis, rather than having to deal uniformly with arbitrarily many stages.

proposals/0000-th-stage-hygiene.rst
Ericson2314: Yes I hope! In general I want to leverage backpack in behind the scenes in a variety of ways. I think it's just better to type check against an interface and only later plug in the machine code.

proposals/0000-th-stage-hygiene.rst
Ericson2314: Yes that dovetails. I think it's good to separate the stages in any event, it's also nice to turn off effects. With either enabled, the other is just a "human convenience", but a good one.

proposals/0000-th-stage-hygiene.rst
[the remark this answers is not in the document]
michaelpj: Can't we just take the union of all the dependencies needed across all splices?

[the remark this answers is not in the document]
Ericson2314: Sor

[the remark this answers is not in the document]
Ericson2314: Sorry I meant to write different *stages*.

proposals/0000-th-stage-hygiene.rst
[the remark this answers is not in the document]
Ericson2314: Ah I can move to effects and interactions. duh.

proposals/0000-th-stage-hygiene.rst
michaelpj: Sure, but I do think you should explicitly say what happens with TH side effects in this proposal.

angerman:
I’m clearly in favor of this! I am however not exceptionally happy with the Syntax; I’m specifically unhappy about the integer stages. I’ll try to get some more experience with racket as it has been mentioned a few times as doing this better.

JacquesCarette:
I am very strongly against getting rid of typed template haskell. To me, this would be a very clear step backwards. The problem is not really TTH (it seems) but rather `Lift`, which allows cross-stage persistence. But of course cross-stage persistence is incompatible with some parts of cross-compilation.

I think the solution is straightforward: when doing meta-programming, one needs to know whether it is homogeneous (and then CSPs are fine) or heterogeneous (where the situation is more complex).

Ericson2314:
1. I'm just proposing a temporary conflict. I definitely want to make everything work together longer term.

2. Perhaps I need to clarify my examples, but typed TH really is problematic too. The crux of it is the type parameter to TExpr needs to be evaluated on acording to the host platform while the type as a whole evaluates on the build platform.

Ericson2314:
Hmm so I don't think this is ready for commit up/down per se--they're are still some TODOs revolving around the Cabal stuff. But, I am curious how the committee feels on the rough thrust of the change here. It's fairly large change to GHC and Cabal, and a big shift to make everyone work together on bootstrapping (of which i consider cross compilation a special case) rather than everyone but the GHC devs blissfully ignorant. That isn't to say I think I will inconvenience most users with more work---I don't think is a harder system to use, but there will be a decent amount of work to GHC and Cabal, and then a lot of submitting PRs to packages to get them to separate their TH and main code into separate components. It's a one-time culture shift and ecosystem overhaul for sure.

I'll summarize the thread so far. Metaprogramming and cross are not really mainstream topics within are community, and so far the commenters have been those that have encountered these issues and agree with the motation, not those that haven't and therefore might be more skeptical of goals. Mainly people are skeptical of the syntax (the "raw" integers), and needing more than 2 or 3 stages. I referenced the Racket precedent for n stages. The vast majority of the comments have been @michaelpj's detailed reviews (thanks again!), but this was tightening up various small issues rather than overall criticism. I also communicated off-thread with @luite briefly who was wondering whether something simpler might suffice but didn't yet have a concrete counter-proposal, and @mpickering who is keen on cross stage persistence and the other features I disable. In the proposal body I say I hope the feature "schism" is temporary, and there is a way to opt in to those features in the end.

CC @nomeata

nomeata:
I guess you are asking for a prelimary assessment from the committee, to see if it is worth pursuing? Guess I’ll have to make up some process for that :-)

simonpj:
I regret to say that I am complete baffled.   I do not understand the problem we are trying to solve.   I do not understand the poposed solution (at all).  This may primarily reflect my own shortcoming, but it's an obstacle to my making any meaningful contribution to this conversation,still less a preliminary assessment.  And perhaps it is a shortcoming shared by others.

To get started in a concrete way, I wonder if you could consider illlustraing the motivation section of the people iwith a range of examples that demonstrate how inadequate the current state of afffairs is?  Perhpas starting with a tiny examples, and then a more realistic ones.  Then you could use those same examples to exemplify your proposed solution.   Thanks!

tydeu:
I like the idea and I believe I understand the utility of stage separation. I am, however, a bit confused with some of the examples.

In your rational for banning Lift and Typed Template Haskell, you state this is because some types may be out of scope. How common is this? Wouldn't you generally want all of the types for all of the platforms to be available, but only *use* the type for the current stage's platform?

Ericson2314:
@simonpj thanks I will rewrite the motivation and effects sections. The examples I do have tend to come after what they describe; I'll make sure they come first making the problem clear.

Ericson2314:
@tydeu well yes, it is in scope, but in the wrong stage. This stuff really matters, because the seemingly same thing may evaluate differently in different stages.

tydeu:
@tydeu I understand that it is possible. I was more curious about how common it was in practice, 

It seems like it would be better to to have types (like your example of `IosOnlyType`) available across all platforms (just like it is useful to be able to read and write little and big endian datatypes even though a platform is generally just one or the other). Instead, one would simply not use a given variation in code that isn't hosted on said platform.

I would think the build stage should know about every piece of functionality (including types and functions) supported by any platform (at some level of abstraction), but the host stage would only use the types relevant to it (as those are the only ones supported on the platform it is running on).

More concretely, based on the examples in your proposal, why can't I do something like the following instead:

```haskell
type LinuxType = ...
type MingwType = ...
type IosType = ...

#if linux_HOST_OS
type HostType = IosType
#endif

#if mingw_HOST_OS
type HostType = MingwType
#endif

#if ios_HOST_OS
type HostType = IosType
#endif 
```

This would allow a every type to always be avaliable regardless of the actual host platform.
However, I imagine there might be some cases were this approach doesn't work. If so, I would like to see (either in this discussion or in the proposal) some more precise examples of when that is. I think it would help improve the Motivation.

On a somewhat unrelated note: why use CPP macros instead of types to store platform information?

goldfirere:
I'm afraid I'm squarely in @simonpj's camp here: I have very little idea of the problem this is trying to solve. And so I didn't even look at the solution, because it wouldn't have made much sense. I understand, in vague terms, that cross-compilation is a challenge with TH (and why) and ought to be fixed, but that's about as far as I get.

I think this proposal (indeed, just about all proposals) are made stronger with an example, early on, of something that is impossible today. Ideally, the solution is nearby, but sometimes that has to be delayed until after the specification part. I also find that this proposal tries to use abbreviations that readers won't necessarily understand. For example, what is QEMU? What does `host := build` mean? (I think it means "where the host platform is the build platform" but I'm not sure. Spelling it out would help!)

Sorry to be dense here, but I'm honestly not sure how to proceed.

michaelpj:
@Ericson2314 I'd be interested in collaborating on a revised version. I think I could help with the exposition.

Ericson2314:
@michaelpj That would be great!

yav:
@goldfirere QEMU is a hardware emulator, which is one way to execute something built for one platform on another.

While I have some experience with autoconf and TH, and I could follow the terminology, I also had quite a bit of trouble understanding the problem and the proposed solution.   For me, the proposal would be improved by having some very concrete examples illustrating 1) what we'd like to do, and 2) what part of TH makes this difficult/impossible to do (I understand this is supposed to be `lift` but I don't see why)

For context, my understanding of the cross-compilation problem in the context of TH is as follows:

Suppose that I'd like to compile Haskell to ARM, but I want to do it on an X86 machine.  To do so, I'd need a binary `ghc_for_x86_producing_arm`

If, however, the Haskell I am compiling also contains TH, then I'd also need to *execute* some Haskell during compilation.  One way to do so would be to have `ghc_for_x86_producing_x86`.  I could use it to compile TH splices to `x86` code, dynamically load it  into `ghc_for_x86_producing_arm`'s process and execute the result to get some AST, then proceed as usual.

Obviously this is non-trivial and probably requires non-trivial amount of engineering work, but I don't (yet) see why it requires changes to TH.

Ericson2314:
OK the first section describing the status quo (problem and current hacks), is now full of examples. I still need to flesh out the other sections, but I figured this is worth mentioning because if the first section is still confusing, the others still won't make sense no matter what I do!

yav:
Thanks for the updates!   I am afraid I still don't quite follow though, probably due to lack of understanding on the details of how things work.   In particular, here is where I am getting confused:

It seems to me that if we are going to be executing some Haskell code during compilation (on the build platform), then we need to either a) compile the code, or b) interpret it.

As I read it, the introduction seems to rule out a) because it requires two versions of the libraries (one for the build platform and one for the target platform), but it also rules out b) because "side effects" are not supported.

This is confusing to me:
   * if we want compilation (option a), I don't see how we could possibly avoid having two builds of some libraries, no matter what changes we make to template Haskell
  * if we are going for interpretation (option b), why can't we perform side effects on the build platform (e.g., loading files, etc.)?

Ericson2314:
@yav Ah! OK I will clear up the wording but let me answer directly: we absolutely do need different versions of the library for build and host, and I wholly embrace that. What I don't want is building two versions of *every* library. That is wasteful. [It also rules out bootstrapping with a different version of a library (say base wanted to to use TH?).

I want both compilation and interpreting to work. I think they can both do side effects on the build platform just fine. I think we still in practice need to two versions of the library for interpreting too, because of CPP, conditional cabal flags, etc are not in scope yet for GHC proper, and so still would pre-eliminated. (Yes, @mpickering and I did talk about making at least CPP more part of GHC, but that's independent, and furthermore it probably makes sense to interpret optimized platform-specific core for dependencies.)

goldfirere:
I still have a bit of a hard time getting through the motivation, but I managed to. The specification section reads fairly nicely. It's all a bit much for me to have an informed opinion on, but here are a few thoughts:

* The `AppE` examples w.r.t. typed TH don't type-check. That is, `AppE :: Exp -> Exp -> Exp`, and you seem to be using it at `AppE :: TExp (a -> b) -> TExp a -> TExp b`. So I skipped that bit.

* I wonder if we can recover `Lift` and typed TH with a notion of a staged type. Something like `[|| xyz ||] :: Q (TExp (Staged 1 "HostSpecificType"))`. Note that `"HostSpecificType"` is a type-level string (`Symbol`), meant to be interpreted at stage-1. Perhaps even `Staged` is a type family such that `Staged 0 "Ty"` is just `Ty` (by some magic). `Lift` could work similarly, by interpreting a value from a stage-0 type into a stage-1 type. We'd need negative integer literals in types to do this properly, and I'm sure other hiccups would ensue. But it strikes me as a possibility. Actually, even better: `[|| xyz ||] :: Q (TExp [|| HostSpecificType ||])`. I've now used a typed quote in a type. Perhaps this just desugars to `Staged`, but I like it.

As you admit, this is *big*. I personally would want a sizeable chorus of voices in strong support before moving forward. This will be a very costly feature to specify and maintain.

mboes:
I have only skimmed through the details of this proposal. I'm not endorsing it per say (yet). But I do want to say that as a build system author, I support all efforts to achieve proper stage hygiene, for the reasons given by @ezyang in http://blog.ezyang.com/2016/07/what-template-haskell-gets-wrong-and-racket-gets-right/, cited in this proposal. Cross compilation is an important use case, but it is for the most part relevant to only some users. I think the benefits of this proposal go beyond just cross-compilation.

When GCC compiles a `.c` file, it only needs the transitively included header files as inputs, along with the `.c` file itself. Multiple `.c` files can thus be build in parallel. If one `.c` file changes, GCC does not need to rebuild the object code for all other `.c` files. In Haskell, things aren't so simple, but things are also worse than they really should be.

When GHC compiles a `.hs` file, it needs the `.hi` files of all imported modules as inputs (whether from the current package or from dependencies), along with the `.hs` file itself. We cannot in general build multiple `.hs` files in parallel. If one `.hs` file changes, GHC might need to rebuild the object code for other `.hs` files. But because of Template Haskell, in reality things are actually worse.

The `.hi` files are not the only inputs required when calling GHC. Because of Template Haskell, you also need the object code for all libraries that are dependencies. Example: if module `Data.Foo` in package `foo` imports `Data.Bar` from package `bar`, then `libHSbar.so` or `libHSbar.a`, respectively the shared and static libraries for package `bar`, are required inputs. This is because `Data.Foo` *could* have splice code generated by a function defined in `bar`.

That's a lot more inputs required for GHC relative to GCC, and a commensurate pessimization for build times. Creating a shared library takes time (sometimes several seconds). No build system can start building package `foo` until the shared library for `bar` has been created, even though only the `.hi` files (which become available shortly after type checking) are required. This means that we are adding several seconds to the critical path, per levels of depth in the dependency tree. Worse, for caching systems like Nix, Bazel and Buck, this means that any change whatsoever to the content of `libHSbar.so`, even in unexported symbols, will invalidate the build cache for `foo`, since `libHSbar.so` is an input to the build action.

So again, if it were not for Template Haskell, we could:

1. have fewer and less bulkier inputs to each build action,
1. interleave building `foo` with linking dependencies as shared libraries (i.e. more parallelizable builds).

TH splices are not the common case. The common case is that a module does not use any Template Haskell. It's fine that building a module that requires Template Haskell (i.e. the slow case) requires more inputs. But we should make the common case as fast as possible. I believe @Ericson2314's proposal, or anything in the spirit of what @ezyang proposed in his blog post, would give the build system enough information that the common case can be made faster. And make the build system cache less brittle and more useful.

There is another point here, which is that it's currently very hard to avoid having to build *both* static libraries and shared libraries for every Cabal package. This is because the default on Linux is for GHCi to use the dynamic linker. This means that Template Haskell needs the shared libraries of all dependencies when evaluating a splice. Even when the user ultimately wants the static libraries to statically link them in their final executable. Here again, I believe that with this proposal it becomes feasible to build the shared libraries on-demand, only when strictly required, because the user can now explain to the build system that when `foo` depends on `bar` and `baz`, it depends on `bar` also for TH splices but on `baz` only at runtime.

Ericson2314:
@mboes Absolutely. Thanks for raising this issue in greater detail than I've talked about in the proposal. I've commented elsewhere that I've recently come to the conclusion that `ghc --make` dependency graph out to be finer-grained and per stage, which makes these types of things much more straight-forward to implement + once GHC cleans up its act external build systems will have an easier time imposing more discipline on it.

aspiwack:
@Ericson2314 don't hesitate — Indeed feel encouraged — to reproduce @mboes's arguments in the proposal. At the risk of belabouring a point: a proposal should ideally be appreciated on its own without reference to the discussion (I do, personally, love the historical insight provided by proposal discussions, but that's besides the point :wink: ). So the proposal should reflect all the arguments which were discussed.

Ericson2314:
@mboes does it wait for the `ho` today, or just the machine code of other packages' modules?

mboes:
@Ericson2314 Did you mean `.hi`, or perhaps `.so`? I don't know what a `ho` is. The graph looks something like this:

```
.hs needed to produce .o's and .hi's.
.o's needed to produce .so
.so + .hi's needed by dependent packages
```

Ideally it would be:

```
.hs needed to produce .o's and .hi's
.o's needed to produce .so
.hi's needed by dependent packages (+ .so in the case of TH)
```

Ericson2314:
@mboes err sorry not `.ho` but just `.o` then. I mean *within* a package if one module provides TH for another. Does GHC always wait for the `.o` of all the local package module dependencies for TH's sake?

Ericson2314:
@mboes ping :) I think within a package we do needlessly depend on the `.o` and not just `.hi` too, but I'd love to confirm that with someone else.

Ericson2314:
CC @nh2 I wrote up @mboes's point about easier static linking, which should interest you.

-------

OK, so with that I think I've included all the new ideas, and hopefully clarified the background with a appendix on cross compilation in general.

nh2:
@Ericson2314 Since I was just pinged here I haven't read everything yet, but here is another idea in case that isn't included yet:

I believe most TH is some very simple "pure" syntax generation, like generating lenses.

"Pure" means it doesn't do any `runIO` and it is architecture independent.

Thus it can always be run on the host in the interpreter; no need for external interpreters or any object code generation.

Thus I propose: Let's make a "pure" variant of TH that isn't allowed to do any `runIO`. Then we can always run that on the host

Ericson2314:
@nh2 Yes I am sympathetic to that, but to be clear I don't think it is a substitute for what is being proposed here. For example in the static linking use-case, that wouldn't help only build SOs when needed (for TH), and ensure they weren't part of the run-time closure (to use the Nix term).

michaelpj:
@nbp I [agree](https://github.com/ghc-proposals/ghc-proposals/pull/243#discussion_r305961940). But we'd need to be careful in specifying what we mean by "architecture-independent" (getting rid of `IO` seems comparatively easy).

JacquesCarette:
I think that @nh2 's idea of having a "pure" part of TH is an excellent, and relatively small, first step in the direction of this proposal.  It could be done independently, and might gather a lot more support. And is in line with other current proposals about TH.  This current proposal is so huge, breaking it down into small steps is likely the best way of actually making forward progress.

Ericson2314:
I'm worried about the opposite. I think most people won't know what they are missing and declare that good enough, sapping support for this proposal.

Besides the other benefits like build incrementality and static linking, saying pure TH is good enough is saying there is some subset of Haskell where the build and host coincide. But I'm interested long term in cross compiling to really exotic targets: circuits, embedded systems with no dynamic memory allocation, etc etc.

I *really* rather rip off the bandaid and start with no assumption the build and host have anything in common. It's the correct programming model.

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,680 @@
...
+Proposed Change Specification
+------------
+
+GHC
+~~~~~~~~~~~~
+
+#. Let there be a notion of stages assigned to the integers.
+   All existing rules outside of TH on binding/name resolution are retaken to act independently per stage.
+   (i.e. identifiers in stage *n* resolve to bindings in stage *n*, all syntax in the rule is parameterized with the stage.)
+   bindings (with existing, regular syntax) on the top level are always in stage 0.
```
facundominguez, line 194:
```suggestion
   Bindings (with existing, regular syntax) on the top level are always in stage 0.
```

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,680 @@
...
+   As a consequence, all non-TH syntax in is also stage 0.
```
facundominguez, line 195:
```suggestion
   As a consequence, all non-TH syntax is also in stage 0.
```

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,680 @@
...
+  data MyType = ...
+  makeLenses ''Foo -- dump this
+
+::
+
+  -- build for host platform (local host platform := overall host platform, cross)
+  data MyType = ...
+  $(...) -- file in with dumped splice, no eval needed
+
+This became easier with `this patch <https://github.com/reflex-frp/reflex-platform/blob/master/splices-load-save.patch>`_.
```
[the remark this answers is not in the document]
Ericson2314:
```suggestion
This became easier with `this patch <https://github.com/reflex-frp/reflex-platform/blob/master/haskell-overlays/splices-load-save/load-splices.nix>`_.
```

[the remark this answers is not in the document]
Ericson2314: Oh wait that is *not* it. Indeed it moved a lot.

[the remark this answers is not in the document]
Ericson2314:
```suggestion
This became easier with the changes on `this branch of GHC <https://gitlab.haskell.org/obsidiansystems/ghc/-/tree/wip/abrar/splices-8.6.5>`_.
```

lexi-lambda:
This proposal seems somewhat dormant. But since it was linked in a recent Twitter discussion, I figured I’d still chime in.

I mostly want to say this: *please* consult the Racket prior art for this. There are a handful of fairly tricky details here, but Racket gets all of them right and has successfully used a phased compilation model for almost two decades now. The canonical publication about this is [Composable and compilable macros: you want it *when?*](https://www.cs.utah.edu/plt/publications/macromod.pdf), which explains the core issues concisely and outlines a clear solution.

I can say from personal experience that Racket’s model gets all of this overwhelmingly right. It handles everything I’ve ever thrown at it (which is a lot), it’s easy to use, and it is conceptually simple. What’s more, it has since been successfully extended with additional features that one could, in theory, imagine Haskell programmers finding useful as well, such as nested modules, discussed in [Submodules in Racket: You Want it *When*, Again?](https://www.cs.utah.edu/plt/publications/gpce13-f-color.pdf).

Racket’s phasing system is sophisticated enough to support local macro definitions (i.e. elimination of the GHC stage restriction), predictable compilation with arbitrary, user-defined, compile-time state, and full macro hygiene. Fully adopting all the features of Racket’s module system is out of scope for TH in the short term, but there is no reason to not adopt a common subset, as many of those features would be useful in Haskell as well. In particular, I would recommend making positive numbers “more compile-time” phases and negative ones “closer to runtime” phases, simply because the choice is wholly arbitrary and opposite conventions would be needlessly confusing. (Also, in practice, references to positive phases seem to occur much more often in Racket than negative ones.)

Ericson2314:
Being like Racket is absolutely the goal here. The order of the integers was an unconscious holdover from Nixpkgs where I last did this sort of thing; I'm happy with flipping the order. 

This is mainly been dormant as we [mainly @hsyl20!] chip away at https://github.com/hsyl20/ghc-cross-compilation. The prereqs for this are less controversial, and after those are done I think this proposal will be easier to discuss because the marginal cost will be substantially less.

wz1000:
My suggestion would be to start small, with the simplest, most conservative change that yields tangible benefits.

For me, this would be something like `-XStagedImports`, which would require "compile time" imports to be explicitly marked.

For example

```haskell
{-# LANGUAGE StagedImports #-} 
module MyModule where

import {-# META #-} Data.Aeson.TH
import {-# UNRESTRICTED #-} Data.Singletons.TH
import Control.Lens.TH

data Foo = ...
deriveJSON defaultOptions ''Foo -- This is legal
genSingletons [''Foo] -- This is too

foo = genSingletons -- This also

makeLenses ''Foo -- This is not

someOptions = defaultOptions -- Neither is this
```

Packages making use of `StagedImports` can be backwards compatible if they switch the extension on conditionally in the cabal file, and use `-Wno-unrecognised-pragmas` for older GHCs.

This would be a major improvement for the purposes of ghcide. We usually don't need to compile code and can just get away with typechecking, but in the presence of any module using `-XTemplateHaskell` we have to conservatively go and compile everything below it in the module graph, even if nothing downstream from the package is actually being used in a splice, and TH is only used for things like `deriveJSON` etc.

With `-XStagedImports`, the impact of this can largely be limited to what is strictly necessary.

Ericson2314:
@wz1000 I'm sympathetic to the use-case and urgency, but I don't think that will be substantially easier than the whole thing, as I think most of the work GHC will be wrangling the internally separate modules for each stage.

wz1000:
> as I think most of the work GHC will be wrangling the internally separate modules for each stage.

Don't we just need to split the `GlobalRdrEnv` into two for my proposed change?

Ericson2314:
My plan was just to split the whole module in two, so mainly changes in driver but dealing with each module stays the same.

mpickering:
I think I agree with @wz1000 here that starting like he proposes is best and then we can adapt a more complicated schema later. The majority of the benefit for normal users is gained from his proposal.

Ericson2314:
The spit modules are needed for cross, but not just to solve the incrementally issues. For this one I would keep that approach, but for @mpickering's I also agree just do something simpler like @wz1000 says. I think that's a good division of the work required.

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,680 @@
...
+
+Cabal
+~~~~~~~~~~~~
+
+#. Extend the ``build-depends`` syntax with an optional stage integer offset parameter.
+   The default is stage 0.
+   N.B ``build-tool-depends`` can be thought of as a stage -1 executable dependencies list.
+   "Those executables are executed at build time, like top-evel splices, and so need to be built for the build platform."
+   `<https://github.com/haskell/cabal/issues/5411>`_ asks for a ``run-tool-depends`` which would be nothing but a stage 0 executable depends.
+   ``setup-depends`` can also be thought of as a stage -1 executable dependencies list.
```
shlevy, line 263: `setup-depends` is libraries needed to compile `Setup.hs`, not executables.

Also, this is perhaps overly pedantic, but IMO calling it `stage -1` is a bit misleading, as in principle the execution context of `Setup` can be arbitrarily decoupled from the execution context of `GHC` (e.g. one could imagine someone implementing the `Setup` interface backed by a heterogeneous-architecture build farm)

Ericson2314:
```suggestion
   ``setup-depends`` can also be thought of as a stage -1 dependencies list.
   In both cases though, the metaphor is not perfect because separate executable are only loosely coupled.
   E.g libraries for Template-Haskell must be built with the same compiler / same ABI, but executable deps (including `./Setup` itself), need not be.
```

Ericson2314: Hows that?

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,682 @@
...
+-------------------
+
+- This is a huge amount of work.
+  But I am fine chipping away it over a long period of time.
+
+- Even a temporary conflict between typed TH and this could slow typed TH's adoption.
+
+- I don't know of precedent for extensions that prevent modules from being linked together.
+
+- Most existing libraries with commonly used TH helpers (`lens`, `aeson`) have the TH in the same Cabal component but in a different module.
```
konsumlamm, line 538:
```suggestion
- This is a huge amount of work.
  But I am fine chipping it away over a long period of time.

- Even a temporary conflict between typed TH and this could slow down typed TH's adoption.

- I don't know of a precedent for extensions that prevent modules from being linked together.

- Most existing libraries with commonly used TH helpers (``lens``, ``aeson``) have the TH in the same Cabal component but in a different module.
```

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,682 @@
...
+  To leverage this proposal, we would have to refactor them to put those modules in a separate library component.
+  It would take decent amount of conditional code to still support old GHCs, and even more to not be a breaking change on those old libraries.
+
+Alternatives
+------------
+
+There is no fundamental reason modules couldn't export non-stage-0 items, and libraries expose non-stage-0 modules.
+At the cost of more complexity, there could be a `.lib` or `.so` for each exposed stage, and imports would be offset to match the ``#import <offset>`` literal.
```
konsumlamm, line 546:
```suggestion
At the cost of more complexity, there could be a ``.lib`` or ``.so`` for each exposed stage, and imports would be offset to match the ``#import <offset>`` literal.
```

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,682 @@
...
+But in fairness, this might allow a smoother transition form how libraries are structured today.
```
konsumlamm, line 547:
```suggestion
But in all fairness, this might allow a smoother transition from how libraries are structured today.
```

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,682 @@
...
+Not only would GHC need to learn more tricks, but also Cabal and other tools.
+For example, one could do ``#import 1 Control.Lens.Lens`` in ``Control.Lens.TH`` while exposing ``Control.Lens.TH`` from the same library just like today.
+I decided against this as a matter of taste.
+I think it good to enforce the normal form that the "main" stage is stage 0.
```
konsumlamm, line 551:
```suggestion
I think it's good to enforce the normal form that the "main" stage is stage 0.
```

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,682 @@
...
+I think that's a cleaner way to package code.
+
+It would be nice to go straight implement typed TH by quoting types.
+One would do something like::
+  [|| e... :: t... ||] :: Q (TExp [t| t... |])
+This correctly gets ``t...`` from the right for the argument to ``TExp``.
+The problem though is the parameter is type syntax rather than a type; we've thrown away all the semantics.
+We can explicitly normalize, however::
+  [|| e... :: t... ||] :: Q (TExp (Normalize [t| t... |])
+``Normalize`` could be some sort of magic type family that just tells GHC to do its thing, rather than a implementation of Haskell.
```
konsumlamm, line 562:
```suggestion
``Normalize`` could be some sort of magic type family that just tells GHC to do its thing, rather than an implementation of Haskell.
```

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,682 @@
...
+
+Quotes in ``-XTemplateStagePersistence`` modules cannot reliably be used from ``-XNoTemplateStagePersistence`` modules without introducing scoping errors.
+Need some way to prevent that outright, or catch those errors early, perhaps by tainting any quote with cross-stage persisted syntax.
+[Thankfully the other direction is fine.
+Libraries can experiment with this extension without forcing an ecosystem split.]
+
+Implementation Plan
+-------------------
+
+I volunteer to chip away at this, thought it will take quite a while for one person to do it all.
```
konsumlamm, line 576:
```suggestion
I volunteer to chip away at this, though it will take quite a while for one person to do it all.
```

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,682 @@
...
+
+#. Parameterize dependency data types (for module and package dependencies) to track dependencies per stage.
+
+#. Refactor the implementation of Template Haskell to use the per-stage data-types.
+
+Appendix: "build", "host", and "target" terminology, and some opinions
+------------------------------------------------------------------------
+
+These terms come from GNU Autoconf `<https://gcc.gnu.org/onlinedocs/gccint/Configure-Terms.html>`_.
+You may also want to compare Nixpkgs's documentation, `<https://nixos.org/nixpkgs/manual/#ssec-cross-platform-parameters>`_, as Nixpkgs's use of these terms very much influenced this proposal.
```
konsumlamm, line 592:
```suggestion
You may also want to compare Nixpkgs' documentation, `<https://nixos.org/nixpkgs/manual/#ssec-cross-platform-parameters>`_, as Nixpkgs' use of these terms very much influenced this proposal.
```

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,682 @@
...
+  If one built the cross compiler themselves, we have ``build = host != target``.
+  Confusion between cross compilers and cross-compiled compilers, has muddled this.
+
+- As hinted in the notes on the target platform we have a law where if ``a `builds` b``, then ``host b`` = ``build a``.
+
+This is a rather colorful appendix, but I really hope to convey a shift in perspective that the dry definitions alone may fail to do.
+Despite Make's early popularizing of call-by-need semantics (if you squint) at build system authors, they reverted to conventional imperative thinking for cross compilation and bootstrapping.
+If you focus on the building, the work, and who does it, then the build platform is given undo importance:
+Native is normal, cross is weird; ``if cross then ... else ...`` code abounds.
+But try focusing on the needing, the "why", so the subject and object are switched if you imagine being a anthropomorphized dependency node and looking at your now-flipped edges and new adjacent nodes like I do!
```
konsumlamm, line 630:
```suggestion
But try focusing on the needing, the "why", so the subject and object are switched if you imagine being an anthropomorphized dependency node and looking at your now-flipped edges and new adjacent nodes like I do!
```

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,682 @@
...
+
+- "Cross" I use to mean ``build != host``, but a cross compiler is ``host != target``.
+  If one built the cross compiler themselves, we have ``build = host != target``.
+  Confusion between cross compilers and cross-compiled compilers, has muddled this.
+
+- As hinted in the notes on the target platform we have a law where if ``a `builds` b``, then ``host b`` = ``build a``.
+
+This is a rather colorful appendix, but I really hope to convey a shift in perspective that the dry definitions alone may fail to do.
+Despite Make's early popularizing of call-by-need semantics (if you squint) at build system authors, they reverted to conventional imperative thinking for cross compilation and bootstrapping.
+If you focus on the building, the work, and who does it, then the build platform is given undo importance:
```
konsumlamm, line 628:
```suggestion
If you focus on the building, the work, and who does it, then the build platform is given undue importance:
```

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,682 @@
...
+Native is normal, cross is weird; ``if cross then ... else ...`` code abounds.
+But try focusing on the needing, the "why", so the subject and object are switched if you imagine being a anthropomorphized dependency node and looking at your now-flipped edges and new adjacent nodes like I do!
+The host platform regains its rightful primacy:
+We "need" something to be built that "run" on our platform, we don't care where it is built.
```
konsumlamm, line 632:
```suggestion
We "need" something to be built that "runs" on our platform, we don't care where it is built.
```

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,682 @@
...
+If the needer doesn't care where the dependency is built, it should care even less where the compiler that built the dependency is built.
+If the compiler is multi-target, and the runtime and standard libraries can be built separately, then "target" almost isn't needed at all:
+every node in the dependency graph is uniquely determined (say for caching purposes) by its build and host platforms (and dependency closure).
+
+Platforms and stages
+~~~~~~~~~~~~~~~~~~~~
+
+I said target "'target' almost isn't needed" because ironically, given my general disdain for abstract platforms other than build and host, this proposal makes them relevant again.
+While stages less than 0, corresponding to "build", "pre-build", "pre-pre-build", etc, should never leak in the library's interface,
+stages greater than 0---quotes and nested quotes---corresponding to "target, "post-target", "post-post-target", do and must influence the interface.
```
konsumlamm, line 643:
```suggestion
stages greater than 0 --- quotes and nested quotes --- corresponding to "target, "post-target", "post-post-target", do and must influence the interface.
```

proposals/0000-th-stage-hygiene.rst
```diff
@@ -0,0 +1,682 @@
...
+From the building perspective this is simple: quotes are not eliminated by compilation, they remain in the interface, the package is thinking further and further ahead on how it is used.
+From the needing perspective, things are more subtle.
+If I have a ``-1`` import on ``Foo``, and splice some code from ``Foo``, I could end up with a quote form ``Foo`` in exposed in stage 0 of myself.
```
konsumlamm, line 646:
```suggestion
If I have a ``-1`` import on ``Foo``, and splice some code from ``Foo``, I could end up with a quote from ``Foo`` exposed in stage 0 of myself.
```

shlevy:
This is strictly speaking out of scope for TH, but dependent types does break this model a bit. On the one hand, values in type semantically need to match the target, on the other hand the typechecker needs to evaluate them on the host (or, more precisely, in the internal interpreter, if we have some kind of bytecode for the typechecker rather than using the results of native codegen). I raise it here only because a proper solution might involve changes relevant to just the TH case.

Thanks to @goldfirere and @lexi-lambda for helping clarify the constraints here.

shlevy:
> This is strictly speaking out of scope for TH, but dependent types does break this model a bit.

Actually I haven't looked in detail but perhaps this is basically the same problem with typed template haskell?

Ericson2314:
@shlevy 

> if we have some kind of bytecode for the typechecker rather than using the results of native codegen

We do. Anything that is "transparent" to DH must have an unfolding. and then the type checker can munch on that. If you use data kinds and export a type synnonym the moral equivalent of that is happening.

As long as there no effects at the type level, we should be fine.

> Actually I haven't looked in detail but perhaps this is basically the same problem with typed template haskell?

Yeah I don't expect typed Template Haskell to work too well. It doesn't have an IMO good enough phase separation. I think https://davidchristiansen.dk/pubs/tyde2020-predictable-macros-abstract.pdf is better, giving regular template Haskell enough power that we can not miss typed template haskell.

Ericson2314:
Excitedly closing in favor of #682! The cabal parts would become an https://github.com/haskell/ecosystem-proposals

PR #244, opened 2019-06-29
Add a bullet about the voting process

gridaphobe:
Importantly, note that the shepherd should lock the thread for the duration of the vote.

TODO:
- [ ] we should commit to a method for counting the votes as well. @nomeata has suggested the [Schulze method](https://en.wikipedia.org/wiki/Schulze_method) elsewhere, which seems fine to me. Any objections?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -314,6 +314,13 @@ proposal is ready for decision.
    Ideally, the committee reaches consensus, as determined by the secretary or
    the shepherd. If consensus is elusive, then we vote, with the Simons
    retaining veto power.
+   
+-  If consensus eludes us, the shepherd may call for a vote. The shepherd may 
+   propose any number of alternative resolutions, which the committee members
+   will rank in order of preference. For the duration of the vote, the shepherd
+   will lock the thread, to ensure that only committee members participate.
+   Once voting is complete, the shepherd will tally the votes and announce the
+   result.
 
 -  The decision is announced, by the shepherd or the secretary, on the Github
    thread and the mailing list.


goldfirere:
I think it's good to codify the voting procedure. I have never studied different voting mechanisms, so I have no opinion on what method -- but I do think that selecting one particular method and sticking to it is good form.

As for community feedback: there is simply no good way to do this. Like any healthy open-source community, ours is fragmented across many communication media, lag time in response, and interests. Any method we choose to collect feedback will be suboptimal. Reaching out to various venues will get more response, but there is no way of qualifying that response as representative. Reaching out to various venues is also expensive, both in the time it takes to do the reaching out, and the inevitable survey fatigue. To me, we have a choice between collecting feedback poorly or not at all. I think that collecting it poorly is better than not at all.

One particular problem with the recent vote is that it was unclear to a casual participant that the vote was intended to be committee-only. This caused some people not to participate. We can fix this problem. But, I think collecting feedback on these tickets is still valuable, should we wish to do it. I do *not* think that this procedure needs to be codified.

One substantive change to this proposal, though: In the recent vote, there was a 4th option (https://github.com/ghc-proposals/ghc-proposals/pull/87#issuecomment-503266938) that got missed, both by the initial call to vote and by most voters. (Those that saw it seemed to like it.) Perhaps before a vote is called, there can be a brief (< 1 week) time for the committee to settle on the list of candidates. Perhaps not, as it creates more bureaucracy.

README.rst
```diff
@@ -314,6 +314,13 @@ proposal is ready for decision.
    Ideally, the committee reaches consensus, as determined by the secretary or
    the shepherd. If consensus is elusive, then we vote, with the Simons
    retaining veto power.
+   
+-  If consensus eludes us, the shepherd may call for a vote. The shepherd may 
+   propose any number of alternative resolutions, which the committee members
```
nomeata: Maybe remove the last sentence of the previous paragraph then?

nomeata:
This was the first vote on a proposal we had in the committee, as far as I recall. I’m inclined to learn from our mistakes, and refine it further when it comes up again (e.g. allow for some time to collect more options, and/or allow members to update their votes as new options emerge (easily possible with GitHub); locking the issue or using a voting service like https://www.condorcet.vote/ which can be restricted to certain people), without enshrining a particular process yet.

gridaphobe:
> But, I think collecting feedback on these tickets is still valuable, should we wish to do it. I do _not_ think that this procedure needs to be codified.

Good point, we probably don't need to codify that procedure. But I still think it would be useful to have a list of active communities that we might want to collect feedback from.

> One substantive change to this proposal, though: In the recent vote, there was a 4th option ([#87 (comment)](https://github.com/ghc-proposals/ghc-proposals/pull/87#issuecomment-503266938)) that got missed, both by the initial call to vote and by most voters. (Those that saw it seemed to like it.) Perhaps before a vote is called, there can be a brief (< 1 week) time for the committee to settle on the list of candidates. Perhaps not, as it creates more bureaucracy.

Like @nomeata, I'd rather avoid overly codifying the process. I think the crucial steps based on #87 are:

1. make it clear that the vote is only for committee members,
2. if the votes are to be collected on GitHub, lock the thread.

I think as long as we're collecting the committee discussion on GitHub, collecting the votes on GitHub makes sense too. That way everything is in one place, the votes of individual members are public, and anyone can verify the count if they desire.

To your point about choosing the list of candidates, I don't think we need to codify this, at least not yet. With only the committee participating, it ought to be easier to address additional candidates as they arrive. Perhaps we call off the vote and start another, perhaps we just update our existing votes in place.

nomeata:
> perhaps we just update our existing votes in place.

My thinking precisely.

gbaz:
Let me also address the community feedback point. I think the time for that is _well before_ voting begins. The feedback should not be in the form of up/down votes, but substantive discussion, and typically this already occurs. I also think that it is largely the responsibility of proposers to seek feedback and attention for their proposals. The committee should of course seek to direct the attention of interested parties or subject-matter experts to the proposals as well.

The guidelines do make the point that "Ideally, the committee reaches consensus, as determined by the secretary or the shepherd. If consensus is elusive, then we vote, with the Simons retaining veto power."

Working on getting voting "just right" only goes so far. What is harder and requires more work is working on trying to manage discussion and balance concerns so that the committee can more often come to a consensus. When votes do occur, that means that the committee, despite having tried hard to see all sides and reach a common understanding, is really balancing some irresolvable tradeoffs, and further, that both options are "probably ok" else there would be clear weight one way or the other. As such, at this point opening the way for more discussion, broader if unbalanced input, etc. seems to be counterproductive. In my experience, once you're in such a tangle, no amount of broader input will untangle you unless someone arrives at some fantastic left-field technical solution that makes the apparent source of conflict vanish. What matters more, in my opinion, is to get things resolved cleanly, decisively, and erring in the absence of other indicators towards conservatism, so that the bigger ticket and less contentious things (which are often far more meaty) can get the attention they deserve.

I think it is also worth noting that, only surprising by its predictability, the proposal  that eventually did lead the committee to call a vote was, naturally, under suitable interpretation, regarding lexical syntax of whitespace. [1]

[1] https://wiki.haskell.org/Wadler's_Law

gridaphobe:
Thanks @gbaz for your feedback!

Like you said, this is not about the normal case of discussion and consensus building, it's about what to do when the happy path fails. 

What I had in mind re: community feedback is that (1) a round of community feedback on the specific options the committee is considering might sway the committee (@simonpj indicated that it might sway him at least), and (2) soliciting feedback on a restricted set of options might get responses from a broader swath of the community than the usual discussion and deliberation process does (which seems mostly dominated by a small number of highly dedicated participants).

We don't need to codify any of this, but if we end up in a similar situation in the future and we want get feedback, it would be nice to have a list of active communities to solicit feedback from. Incidentally, this list would also be useful for proposal authors for the main discussion.

goldfirere:
@gridaphobe Any further action to take here?

PR #246, opened 2019-07-03
Overloaded Quotation Brackets

mpickering:
[Rendered](https://github.com/mpickering/ghc-proposals/blob/overloaded-proposal/proposals/0000-overloaded-bracket.rst)

### DIFF
--- a/proposals/0000-overloaded-bracket.rst
+++ b/proposals/0000-overloaded-bracket.rst
@@ -15,7 +15,7 @@ Overloaded Quotations
 .. sectnum::
 .. contents::
 
-This proposal is about making quotation brackets polymorphic. The motivation
+This proposal is about making quotation brackets (typed and untyped) polymorphic. The motivation
 is so that quotes can be used without being tied to the ``Q`` monad and to
 enable effects during code generation.
 
@@ -263,7 +263,7 @@ make this change possible.
     i = [|| \x -> x + 1 ||]
 
 
-   If at a later point ``Q (TExp a)`` is turned into a newtype then an extra
+   If at a later point `(Proposal 195) <https://github.com/ghc-proposals/ghc-proposals/pull/195>`_ ``Q (TExp a)`` is turned into a newtype then an extra
    parameter to indicate the monad used will be added to the wrapper::
 
     i :: Quote m => Code m (Int -> Int)
@@ -272,6 +272,13 @@ make this change possible.
    The monad will be exposed in the newtype to support user-defined effects
    during code generation but retaining the newtype so that the typed representation
    can still be placed into maps and instances defined easily for it.
+
+7. The types of ``untypeQ`` and ``unsafeTExpCoerce`` are generalised in the natural
+   manner::
+
+    untypeQ :: Quote m => m (TExp a) -> m Exp
+    unsafeTExpCoerce :: Quote m => m Exp -> m (TExp a)
+
 
 
 Effect and Interactions
@@ -284,6 +291,17 @@ ambiguity in the composition.
 
 It doesn't seem to me that there will be any problems with ambiguity here as
 the types of splices is not overloaded in the same manner.
+
+Due to the monomorphism restriction, unannotated top-level bindings will no
+longer typecheck by default::
+
+  module A where
+
+  -- Fails to typecheck due to unsolved constraint m
+  foo = [| 5 |]
+
+It is easy to workaround this in a backwards compatible way by either adding a
+type signature or turning on ``NoMonomorphismRestriction``.
 
 
 Interaction with Lift
--- a/proposals/0000-overloaded-bracket.rst
+++ b/proposals/0000-overloaded-bracket.rst
@@ -16,7 +16,8 @@ Overloaded Quotations
 .. contents::
 
 This proposal is about making quotation brackets polymorphic. The motivation
-is so that quotes can be used without being tied to the ``Q`` monad.
+is so that quotes can be used without being tied to the ``Q`` monad and to
+enable effects during code generation.
 
 
 Motivation
@@ -44,7 +45,12 @@ implementation is a name generation effect.
 Detaching quotations from ``Q`` makes way for a form of "pure" template haskell
 so there is no need to invoke ``Q`` in order to create the representation of an
 expression. The most immediate application is the ability to purely
-manipulate ``Exp`` values in user libraries.
+manipulate ``Exp`` values in user libraries::
+
+  lamPlus1 :: Exp
+  lamPlus1 = (runParse us [| \x -> x + 1 |])
+
+
 
 Another benefit is that in a cross compilation setting a "pure" quote can be
 fully evaluated on the host and then the generated code compiled for the target.
@@ -52,9 +58,53 @@ Certain effects in the ``Q`` monad mean that currently all splices have to be
 evaluated on the target which leads to significant complication when
 cross-compiling.
 
-The proposal doesn't deal with anything to do with Typed Template Haskell at
-all. This generalisation is expected to follow in a future proposal.
+In a similar fashion, we can overload the type of a typed quotation::
 
+  lamPlus :: TExp (Int -> Int)
+  lamPlus = (runParse us [|| \x -> x + 1 ||])
+
+Due to the implementation of a typed quotation being already in terms of
+untyped syntax, the implementation of this is natural.
+
+Motivation 2: Effectful Code Generation
+---------------------------------------
+
+Jamie Willis provides additional motivation for the generalisation of the
+quotation bracket. Whilst writing multi-stage programs it is almost inevitable
+that you will need to perform effects whilst doing the code generation.
+For example, this is from the abstract of Kameyama, Kiselyov and Shan's (2014) Combinators
+for impure yet hygienic code generation:
+
+    Code generation is the leading approach to making high-performance software reusable. Effects are indispensable in code generators, whether to report failures or to insert let-statements and if- guards.
+
+To be precise, in his parser combinator library which is implemented using
+typed template haskell the following effects are necessary.
+
+1. Use the ``Reader`` monad in order to carry around an environment.
+2. Use a let-insertion effect to automatically insert lets to avoid code duplication.
+3. Use an exception monad to automatically insert missing dependencies for mutually recursive code generation.
+
+Even using the simple ``Reader`` monad is awkward at the moment::
+
+   generateLoop :: String -> Reader CodeMap (Q Exp)
+   generateLoop name = ask (\codeMap -> [|
+     let loopyCode x =
+       $(runReader loopBody (Map.insert name [|loopyCode|] codeMap))
+     in loopyCode ...
+     |])
+
+The effect must be explicitly run in each splice. For state or other more complicated
+effects this approach doesn't work. With the proposal we would hope to write something like::
+
+   generateLoop :: (MonadReader CodeMap m, Quote m) => String -> m Exp
+   generateLoop name = [|
+     let loopyCode x =
+       $(local (Map.insert name [|loopyCode|]) loopBody)
+     in loopyCode ...
+     |]
+
+and directly use the ``local`` function inside the nested splice just like normal
+monadic programming.
 
 
 Proposed Change Specification
@@ -111,6 +161,11 @@ make this change possible.
 5. The types of type, pattern and declaration quotes will also
    be generalised in the same manner.
 
+6. Type quotations are similarly generalised::
+
+   i :: Quote m => m (TExp (Int -> Int))
+   i = [|| \x -> x + 1 ||]
+
 
 Effect and Interactions
 -----------------------
@@ -147,7 +202,8 @@ Definition of Quote
 Richard observes that ``Language.Haskell.TH.Lib.Internal.numTyLit`` calls
 ``fail`` from the ``Q`` monad. This call to ``fail`` can be replaced with
 a call to ``error``. It will still be executed at compile-time but with a
-potentially slightly worse error message.
+potentially slightly worse error message. The alternative is to
+also add this effect to the ``Quote`` type class.
 
 Other Effects
 .............
@@ -175,7 +231,11 @@ Costs and Drawbacks
 Alternatives
 ------------
 
-* Just keep things the way they are.
+* The main alternative to the design would be to only require a ``Quote``
+  constraint when the quotation requires the ``newName`` effect. For example,
+  ``[| 5 |] :: Applicative m => m Exp``. I am opposed to this direction as it
+  breaks abstraction. The implementation detail of how ``[| 5 |]`` is desugared
+  leaks to the user.
 
 Unresolved Questions
 --------------------
--- a/proposals/0000-overloaded-bracket.rst
+++ b/proposals/0000-overloaded-bracket.rst
@@ -22,10 +22,12 @@ is so that quotes can be used without being tied to the ``Q`` monad.
 Motivation
 ------------
 
-Quoting an expression ``[| e |]`` yields its representation. In the current implementation
-the type of representations is ``Q Exp``. This proposal is changing the type of
-the representation to the polymorphic ``Quote m => m Exp``. The ``Quote`` interface
-defines the operations which are necessary to construct the code representation::
+Quoting an expression ``[| e |]`` yields its representation. In the current
+implementation the type of representations is ``Q Exp``. This proposal is
+changing the type of the representation to the polymorphic
+``Quote m => m Exp``.
+The ``Quote`` interface defines the operations which are necessary to construct
+the code representation::
 
    class Applicative m => Quote m where
       newName :: String -> m Name
@@ -40,14 +42,27 @@ construct a representation of expressions, the only effect which is used in the
 implementation is a name generation effect.
 
 Detaching quotations from ``Q`` makes way for a form of "pure" template haskell
-which can be fully evaluated on the host without having to run effects on the target.
+so there is no need to invoke ``Q`` in order to create the representation of an
+expression. The most immediate application is the ability to purely
+manipulate ``Exp`` values in user libraries.
+
+Another benefit is that in a cross compilation setting a "pure" quote can be
+fully evaluated on the host and then the generated code compiled for the target.
+Certain effects in the ``Q`` monad mean that currently all splices have to be
+evaluated on the target which leads to significant complication when
+cross-compiling.
+
+The proposal doesn't deal with anything to do with Typed Template Haskell at
+all. This generalisation is expected to follow in a future proposal.
+
 
 
 Proposed Change Specification
 -----------------------------
 
-The goal of the changes is for an expression ``e : T`` to give the representation
-``[| e |] : Quote m => m Exp``. Several steps are necessary to make this change possible.
+The goal of the changes is for an expression ``e : T`` to give the
+representation ``[| e |] : Quote m => m Exp``. Several steps are necessary to
+make this change possible.
 
 1. Define the interface for ``Quote``::
 
@@ -57,36 +72,56 @@ The goal of the changes is for an expression ``e : T`` to give the representatio
    These are all the operations which are necessary to build the representation
    of expressions.
 
-2. Generalise all the combinators to build syntax in ``Language.Haskell.TH.Lib``.
-   Due to an `audit <https://github.com/ghc-proposals/ghc-proposals/issues/211#issuecomment-472092412>`_ conducted by Richard, it was found that the only effect from
-   ``Q`` which was used is the ``newName`` function which generates a fresh name.
-   All the other combinators can be defined using the ``Applicative`` operations.
+2. Generalise all the combinators which build syntax in ``Language.Haskell.TH.Lib``.
+   Due to an `audit <https://github.com/ghc-proposals/ghc-proposals/issues/211#issuecomment-472092412>`_
+   conducted by Richard, it was found that the only effect from
+   ``Q`` which was used is the ``newName`` function which generates a fresh
+   name. All the other combinators can be defined using the ``Applicative``
+   operations.
 
 
 3. Generalise the ``Lift`` type class::
 
       class Lift a where
          lift :: Quote m => a -> m Exp
-         liftT :: Quote m => a -> m (TExp a)
 
    This is necessary so that implicit lifting can continue to work without
    enforcing strong constraints on the type of the bracket.
 
-4. Refine the rules to do with splicing. A top-level splice instantiates ``m ~ Q``
-   and operates as before. The use of nested brackets doesn't enforce any
-   specific constraints on ``m``. A quotation which contains splices inherits
-   the constraints on the representation.
+4. Refine the rules to do with splicing.  The type of
+   a quotation depends on the types of the nested splices inside it::
+
+      -- Add a redundant constraint to demonstrate that constraints on the
+      -- monad used to build the representation are propagated when using nested
+      -- splices.
+      f :: (Quote m, C m) => m Exp
+      f = [| 5 | ]
+
+      -- f is used in a nested splice so the constraint on f, namely C, is propagated
+      -- to a constraint on the whole representation.
+      g :: (Quote m, C m) => m Exp
+      g = [| $f + $f |]
+
+   A top-level splice still requires its argument to be of type ``Q Exp``.
+   So then splicing in ``g`` will cause ``m`` to be instantiated to ``Q``::
+
+    h :: Int
+    h = $(g) -- m ~ Q
+
+5. The types of type, pattern and declaration quotes will also
+   be generalised in the same manner.
+
 
 Effect and Interactions
 -----------------------
 
-When making an interface more general it is important to think about whether
-it will affect type inference. If there are functions where we have to generalise
-the argument type but not the result then generalisation can result in ambiguity
-in the composition.
+When making an interface more general it is important to think about whether it
+will affect type inference. If there are functions where we have to generalise
+the argument type but not the result then generalisation can result in
+ambiguity in the composition.
 
-It doesn't seem to me that there will be any problems with ambiguity here as the
-types of splices is not overloaded in the same manner.
+It doesn't seem to me that there will be any problems with ambiguity here as
+the types of splices is not overloaded in the same manner.
 
 
 Interaction with Lift
@@ -98,13 +133,13 @@ The main breakage from this patch comes from modifying the type signature for
 Instances defined using ``DeriveLift`` will continue to work because they are
 defined in terms of quotation brackets.
 
-Instances written in terms of the combinators from ``Language.Haskell.TH.Lib`` will
-continue to work because these combinators will be generalised.
+Instances written in terms of the combinators from ``Language.Haskell.TH.Lib``
+will continue to work because these combinators will be generalised.
 
 Instances written in terms of ``Q`` will no longer work. For users to migrate
-an additional class ``LiftQ`` could be defined which has the old interface. This
-would mean users need to explicitly lift but there are likely only a few instances
-which fall into this category if any at all.
+an additional class ``LiftQ`` could be defined which has the old interface.
+This would mean users need to explicitly lift but there are likely only a few
+instances which fall into this category if any at all.
 
 Definition of Quote
 ...................
@@ -134,8 +169,8 @@ Costs and Drawbacks
 * The generalisation of untyped brackets does not seem like it will cause
   any significant breakage but it's hard to predict.
 * The modification to the ``Lift`` interface could cause user-written instances
-  to break but users should not define their own instances anyway. ``DeriveLift``
-  is the blessed manner in which to define a ``Lift`` instance.
+  to break but users should not define their own instances anyway.
+  ``DeriveLift`` is the blessed manner in which to define a ``Lift`` instance.
 
 Alternatives
 ------------
@@ -147,8 +182,8 @@ Unresolved Questions
 
 * Carter points out that if you want to achieve "pure" template haskell then
   you still need to deal with the fact that different platforms have different
-  representations of primitive data types. This is out of scope of this proposal
-  though.
+  representations of primitive data types. This is out of scope of this
+  proposal.
 
 * It would also be possible to make ``Quote`` a superclass of ``Q`` but
   this hierarchy refactoring seems unecessary.
--- a/proposals/0000-overloaded-bracket.rst
+++ b/proposals/0000-overloaded-bracket.rst
@@ -29,7 +29,6 @@ defines the operations which are necessary to construct the code representation:
 
    class Applicative m => Quote m where
       newName :: String -> m Name
-      fail :: String -> m Name
 
 An instance for ``Quote`` can be implemented
 for ``Q``, retaining backwards compatibility, but also ``State NameSupply`` to
--- a/proposals/0000-overloaded-bracket.rst
+++ b/proposals/0000-overloaded-bracket.rst
@@ -105,14 +105,15 @@ Interaction with Lift
 The main breakage from this patch comes from modifying the type signature for
 ``lift``.
 
-Instances defined using ``DeriveLift`` will continue to work.
+Instances defined using ``DeriveLift`` will continue to work because they are
+defined in terms of quotation brackets.
 
 Instances written in terms of the combinators from ``Language.Haskell.TH.Lib`` will
-continue to work.
+continue to work because these combinators will be generalised.
 
-Instances written in terms of ``Q`` will no longer work. For user's to migrate
+Instances written in terms of ``Q`` will no longer work. For users to migrate
 an additional class ``LiftQ`` could be defined which has the old interface. This
-would mean user's need to explicitly lift but there are likely only a few instances
+would mean users need to explicitly lift but there are likely only a few instances
 which fall into this category if any at all.
 
 Interaction with Typed Template Haskell
--- a/proposals/0000-overloaded-bracket.rst
+++ b/proposals/0000-overloaded-bracket.rst
@@ -1,0 +1,205 @@
+Overloaded Quotations
+=====================
+
+.. proposal-number:: Leave blank. This will be filled in when the proposal is
+                     accepted.
+.. ticket-url:: Leave blank. This will eventually be filled with the
+                ticket URL which will track the progress of the
+                implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+This proposal is about making quotation brackets polymorphic. The motivation
+is so that quotes can be used without being tied to the ``Q`` monad.
+
+
+Motivation
+------------
+
+Quoting an expression ``[| e |]`` yields its representation. In the current implementation
+the type of representations is ``Q Exp``. This proposal is changing the type of
+the representation to the polymorphic ``Quote m => m Exp``. The ``Quote`` interface
+defines the operations which are necessary to construct the code representation::
+
+   class Applicative m => Quote m where
+      newName :: String -> m Name
+
+An instance for ``Quote`` can be implemented
+for ``Q``, retaining backwards compatibility, but also ``State NameSupply`` to
+enable a pure way to extract expressions.
+
+``Q`` is not necessary for the implementation of quotations.
+It provides much more power than is necessary. Quotations are used in order to
+construct a representation of expressions, the only effect which is used in the
+implementation is a name generation effect.
+
+Detaching quotations from ``Q`` makes way for a form of "pure" template haskell
+which can be fully evaluated on the host without having to run effects on the target.
+
+
+Proposed Change Specification
+-----------------------------
+
+The goal of the changes is for an expression ``e : T`` to give the representation
+``[| e |] : Quote m => m Exp``. Several steps are necessary to make this change possible.
+
+1. Define the interface for ``Quote``::
+
+      class Applicative m => Quote m where
+         newName :: String -> m Name
+
+   These are all the operations which are necessary to build the representation
+   of expressions.
+
+2. Generalise all the combinators to build syntax in ``Language.Haskell.TH.Lib``.
+   Due to an `audit <https://github.com/ghc-proposals/ghc-proposals/issues/211#issuecomment-472092412>`_ conducted by Richard, it was found that the only effect from
+   ``Q`` which was used is the ``newName`` function which generates a fresh name.
+   All the other combinators can be defined using the ``Applicative`` operations.
+
+3. Overload the type of typed quotations as well. ``[|| e ||] :: QuoteT p => p T``::
+
+      class QuoteT p where
+         fromBracket :: (forall m . Quote m => m Exp) -> p a
+
+   ``fromBracket`` is a generalisation of ``unsafeTExpCoerce`` which allows the
+   type of resulting representation to be modified. This is not a backwards
+   compatible change because the current type of quotatations is ``Q (TExp a)`` for which
+   you can't write an instance of ``QuoteT``.
+
+4. Generalise the ``Lift`` type class::
+
+      class Lift a where
+         lift :: Quote m => a -> m Exp
+         liftT :: QuoteT m => a -> m a
+
+   This is necessary so that implicit lifting can continue to work without
+   enforcing strong constraints on the type of the bracket.
+
+5. Refine the rules to do with splicing. A top-level splice instantiates ``m ~ Q``
+   and operates as before. The use of nested brackets doesn't enforce any
+   specific constraints on ``m``. A quotation which contains splices inherits
+   the constraints on the representation.
+
+
+Effect and Interactions
+-----------------------
+
+When making an interface more general it is important to think about whether
+it will affect type inference. If there are functions where we have to generalise
+the argument type but not the result then generalisation can result in ambiguity
+in the composition.
+
+It doesn't seem to me that there will be any problems with ambiguity here as the
+types of splices is not overloaded in the same manner.
+
+
+Interaction with Lift
+.....................
+
+The main breakage from this patch comes from modifying the type signature for
+``lift``.
+
+Instances defined using ``DeriveLift`` will continue to work.
+
+Instances written in terms of the combinators from ``Language.Haskell.TH.Lib`` will
+continue to work.
+
+Instances written in terms of ``Q`` will no longer work. For user's to migrate
+an additional class ``LiftQ`` could be defined which has the old interface. This
+would mean user's need to explicitly lift but there are likely only a few instances
+which fall into this category if any at all.
+
+Interaction with Typed Template Haskell
+.......................................
+
+The bracket form for Typed Template Haskell should also be generalised but should
+use a different interface to normal quotation brackets to allow for the type
+index. Given an expression ``e : T``,
+``[|| e ||] : QuoteT m => m T``.
+
+This is not backwards compatible with the existing typed template haskell
+implementation because you are forbidden from writing an instance for `Q (TExp a)`
+without using ``Compose`` or a specific newtype. There is more discussion about
+why making this change is a good idea anyway in `proposal 195 <https://github.com/ghc-proposals/ghc-proposals/pull/195>`_.
+
+Connection with StaticPtrs
+..........................
+
+This form of overloading is already implemented for static pointers, a similar
+metaprogramming facility to template haskell.
+
+Definition of Quote
+...................
+
+Richard observes that ``Language.Haskell.TH.Lib.Internal.numTyLit`` calls
+``fail`` from the ``Q`` monad.
+
+It needs to be decided whether ``fail`` should also be added as a member of
+``Quote``. Desugaring to ``error`` results in a run-time error rather than a
+compile-time error as in the correct implementation.
+
+Other Effects
+.............
+
+Vlad points out that you don't need to very strict about the types of
+expressions in splices. Each nested splice could have different constraints::
+
+      f :: Quasi m => m Exp
+      g :: MonadIO m => m Exp
+      [| putStrLn $(f) >> putStrLn $(g) |] :: (Applicative m, Quasi m, MonadIO m) => m Exp
+
+If one of the nested splices has a specific type, for instance ``Q Exp``, then
+the type of the whole expression is fixed to be ``Q Exp``.
+
+
+Costs and Drawbacks
+-------------------
+
+* The generalisation of untyped brackets does not seem like it will cause
+  any significant breakage but it's hard to predict.
+* The modification to the ``Lift`` interface could cause user-written instances
+  to break but users should not define their own instances anyway. ``DeriveLift``
+  is the blessed manner in which to define a ``Lift`` instance.
+* The changes to the typed template haskell quotation is in line with the untyped
+  variant but will be a breaking change to user programs. The number of users of
+  typed template haskell is however very small so this shouldn't stop the adoption of
+  the proposal.
+
+Alternatives
+------------
+
+
+
+Unresolved Questions
+--------------------
+
+* Carter points out that if you want to achieve "pure" template haskell then
+  you still need to deal with the fact that different platforms have different
+  representations of primitive data types. This is out of scope of this proposal
+  though.
+
+* The definition of `QuoteT` does not allow constraints to propagate in the same
+  manner as untype quotations. Perhaps the definition should instead be::
+
+      class QuoteT p where
+         fromBracket :: (forall m . c m => m Exp) -> p c a
+
+      data Code c a where
+         Code :: (forall m . c m => m Exp) -> Code c a
+
+      runCode :: c m => Code c a -> m Exp
+      runCode (Code e) = e
+
+* It would also be possible to make ``Quote`` a superclass of ``Q `` but
+  this hierarchy refactoring seems unecessary.
+
+Implementation Plan
+-------------------
+
+* I will implement this.
--- a/proposals/0000-overloaded-bracket.rst
+++ b/proposals/0000-overloaded-bracket.rst
@@ -93,7 +93,7 @@ This proposal has three parts to it:
 * Define a dedicated class for fresh name generation, with an operation much
   like the existing ``qNewName :: Quasi m => String -> m Name``::
 
-   class Applicative m => Quote m where
+   class Monad m => Quote m where
       newName :: String -> m Name
 
   The notable difference is that this is not bundled with other operations of
@@ -208,7 +208,7 @@ make this change possible.
 
 1. Define the interface for ``Quote``::
 
-      class Applicative m => Quote m where
+      class Monad m => Quote m where
          newName :: String -> m Name
 
    These are all the operations which are necessary to build the representation
@@ -218,12 +218,12 @@ make this change possible.
    Due to an `audit <https://github.com/ghc-proposals/ghc-proposals/issues/211#issuecomment-472092412>`_
    conducted by Richard, it was found that the only effect from
    ``Q`` which was used is the ``newName`` function which generates a fresh
-   name. All the other combinators can be defined using the ``Applicative``
+   name. All the other combinators can be defined using the ``Monad``
    operations.
 
    For example, the ``appE`` combinator which constructs an application is
    generalised to ``Quote m => m Exp -> m Exp -> m Exp``, the ``varE`` function
-   to ``Name -> m Exp`` and the ``lamE`` function to ``Quote m => [m Pat] -> m Exp -> m Exp``.
+   to ``Quote m => Name -> m Exp`` and the ``lamE`` function to ``Quote m => [m Pat] -> m Exp -> m Exp``.
    In general any ``ExpQ`` type is replaced with ``m Exp``, ``PatQ`` with ``m Pat`` and so on.
 
 3. Generalise the ``Lift`` type class::
@@ -340,7 +340,7 @@ Alternatives
 
 * The main alternative to the design would be to only require a ``Quote``
   constraint when the quotation requires the ``newName`` effect. For example,
-  ``[| 5 |] :: Applicative m => m Exp``. I am opposed to this direction as it
+  ``[| 5 |] :: Monad m => m Exp``. I am opposed to this direction as it
   breaks abstraction. The implementation detail of how ``[| 5 |]`` is desugared
   leaks to the user.
 
--- a/proposals/0000-overloaded-bracket.rst
+++ b/proposals/0000-overloaded-bracket.rst
@@ -29,6 +29,7 @@ defines the operations which are necessary to construct the code representation:
 
    class Applicative m => Quote m where
       newName :: String -> m Name
+      fail :: String -> m Name
 
 An instance for ``Quote`` can be implemented
 for ``Q``, retaining backwards compatibility, but also ``State NameSupply`` to
@@ -53,6 +54,7 @@ The goal of the changes is for an expression ``e : T`` to give the representatio
 
       class Applicative m => Quote m where
          newName :: String -> m Name
+         fail :: String -> m Name
 
    These are all the operations which are necessary to build the representation
    of expressions.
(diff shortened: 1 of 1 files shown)


int-index:
Thanks for writing this up! Since we have discussed it extensively already, needless to say I support the proposed changes.

proposals/0000-overloaded-bracket.rst
```diff
@@ -0,0 +1,239 @@
...
+      instance TypedQuote Expr where
+         fromBracket = Leaf . Code
+
+      expr = [|| 5 ||] `Add` [|| 10 ||]
+
+Example 2: A stage polymorphic interpreter::
+
+      data Code m a = Code (m (TExp a))
+
+      instnace TypedQuote Code where
```
int-index: s/instnace/instance/

Ericson2314:
I still feel the Typed Template Haskell changes should be a separate proposal depending on the original. 

I agree that `Code` should be possible to write without involving `Q`, but I still don't understand that motivations that we must overload it breaking unwrapped `m (TExp a)`. This is unlike the first part which should really be uncontroversial to everyone.

proposals/0000-overloaded-bracket.rst
```diff
@@ -0,0 +1,239 @@
...
+
+      weaken :: Code WQ a -> WQ (TExp a)
```
Ericson2314: `WQ` is no longer in the proposal, want to replace with `State NameSupply` or bind that as `WQ`?

proposals/0000-overloaded-bracket.rst
```diff
@@ -0,0 +1,239 @@
...
+
+There is another useful axis of polymorphism for typed quotations.
+It is useful to be polymorphic in the type constructor
+``p``. This can be used to embed quotations directly into an EDSL and write
+instances for typed quotations which isn't convenient without the polymorpism.
+
+Example 1: Creating a partially static data type::
+
+      data Expr m a where
+         Leaf :: Code m a -> Expr a
```
int-index: Shouldn't it be `Leaf :: Code m a -> Expr m a`? Same goes for the other constructors, `Int` and `Add`.

proposals/0000-overloaded-bracket.rst
```diff
@@ -0,0 +1,239 @@
...
+         Int :: Int -> Expr Int
+         Add :: Expr Int -> Expr Int -> Expr Int
+
+      data Code m a = Code (m (TExp a))
+
+      instance TypedQuote Expr where
+         fromBracket = Leaf . Code
+
+      expr = [|| 5 ||] `Add` [|| 10 ||]
```
int-index: Admittedly, I don't fully understand this use case. OK, now we have `expr :: Quote m => Expr m Int`, how would one instantiate `m` and make use of this `expr`?

And why would one prefer it over this variant, where `Expr` does not mention `m`:

```
data Expr a where
  Leaf :: TExp a -> Expr a
  Int :: Int -> Expr Int
  Add :: Expr Int -> Expr Int -> Expr Int

newtype ExprGen m a =
  ExprGen (m (Expr a))  -- like Code for TExp

instance TypedQuote ExprGen where
  fromBracket = ExprGen . fmap Leaf

add :: Applicative m => ExprGen m Int -> ExprGen m Int -> ExprGen m Int
add (ExprGen x) (ExprGen y) = ExprGen (liftA2 Add x y)

expr = [|| 5 ||] `add` [|| 10 ||]
```

Ericson2314:
Static pointers are mentioned for comparision, but the crucial difference is https://hackage.haskell.org/package/base/docs/GHC-StaticPtr.html#t:IsStatic supports the identity instance. If that is to be the inspiration, I rather make it more directly so:
```haskell
class IsExp p where
   fromExp :: Exp -> p
instance IsExp Exp where
  fromExp = id

class IsTExp p where
   fromTExp :: TExp a -> p a
instance IsTExp TExp where
  fromTExp = id
```

@int-index's alternative code example (https://github.com/ghc-proposals/ghc-proposals/pull/246#discussion_r300340013) is changed to:

```haskell
data Expr a where
  Leaf :: TExp a -> Expr a
  Int :: Int -> Expr Int
  Add :: Expr Int -> Expr Int -> Expr Int

instance IsTExp Expr where
  fromTExp = Leaf

add :: Applicative m => m (Expr Int) -> ExprGen m (Expr Int) -> m (Expr Int)
add = liftA2 Add

expr = [|| 5 ||] `add` [|| 10 ||]
```

This doesn't look too verbose to me. If anything now `add` is hardly needed!

goldfirere:
I'm broadly in support, at least of the untyped TH changes.

1. I think we can remove `fail` from `Quote`. You say that having `numTyLit` call `error` would be a run-time error, but I don't see how that's the case: the `error` would run at the same stage as the `fail` does now. The `error` would just produce (perhaps) a worse error message. But this is such a corner case that I don't think it should drive the overall design.

2. I don't understand specification point (4), which mentions top-level splices. This proposal is all about quotes, so I see no change at all to top-level splices. A *nested* splice might indeed change the type of the surrounding quote. And I can't extract any meaning from "A quotation which contains splices inherits the constraints on the representation.", I'm afraid.

3. Could I ask you to refactor the typed TH section? The specification should be exactly that; you have some motivation mixed in there as well.

4. The typed TH proposal generalizes along a different axis than the untyped proposal. I don't see why untyped quotes wouldn't similarly benefit from this form of generalization. The class name `TypedQuote` threw me off, because it is almost unrelated to `Quote`. It's almost `TypedLang`, but I don't really like that name either. I almost want the typed TH proposal to be completely separate from the rest of this proposal, as it seems orthogonal. Yet I'm sympathetic to the overhead that proposal-writing causes.

5. What's `QuoteT`?

6. The example under "Other Effects" suggests that nested splices induce a `Quasi` constraint instead of setting `m ~ Q`. Which is it? (I think the `Quasi` constraint is easier, but maybe I'm missing something.)

7. Should we consider banning hand-written `Lift`? Are there non-trivial hand-written `Lift` instances out there? Or, at least, we can explicitly say that `Lift` will not aim to preserve backward-compatibility: use `deriving Lift` if you want compatibility.

8. I agree that it's OK to break typed TH.

9. Why do we need the typed TH part of the proposal? Can't you do what you want by just adding a function call on your quote? I don't see why this generalization needs to happen inside TH: it looks like everything would just be done in a post-processor, anyway.

mpickering:
@Ericson2314 I'm also fine with that but I was under the impression that making `Q (TExp` a newtype was contentious. 

@goldfirere 

1. That seems correct. I will remove `fail`.

2. This is the point made by @int-index. If you have `c :: C1 m => m Exp` and `c' :: C2 m => m Exp` then the idea is that the quotation has the union of the constraints of the constituent
parts `[| $c $c' |] : (C1 m, C2 m) => m Exp`.  

3. I will split the section about typed template haskell into another proposal.

4. The reason not to generalise untyped quotations is to retain backwards compatibility. 

5. `QuoteT` is supposed to be `TypedQuote`. Typo.

6. What the constraint is depends on the type of the nested splice. The point is supposed to be that if
one of the nested splices is specifically of type `Q Exp` then the whole quote must also be of that type.

7. There are some hand-written lift instances so I would prefer to not ban it. 

8. Thanks :)

9. The reason for the existence of the typed template haskell part of the proposal is that in #195 the proposal was shelved as everyone seemed to agree that this axis of generalisation was better than just making `Q (TExp` a newtype. I then got confused about these two different axis and so started writing this proposal in order to solve #195 but then realised that the `WQ` stuff was completely separate. At this point I was getting sick of it so I just ploughed on and hoped for clemency. 

It is getting a bit frustrating. I have been trying to make `Q (TExp` into a new type for many months!
I would be happy at this point if I could just make it into a new type rather than this other generalisation but given that (1) the other commenters believed this generalisation was better and (2) it is already how static pointers work, I thought it was better to try to propose to make it polymorphic. 

So I will remove the parts about typed template haskell as I need to write a proposal anyway which completely revamps it. Hopefully by the end of the summer I will have done that.

proposals/0000-overloaded-bracket.rst
mpickering: That code also seems fine. I think you could choose to design it either way?

proposals/0000-overloaded-bracket.rst
mpickering: Yes. Thanks.

proposals/0000-overloaded-bracket.rst
mpickering: Thanks.

proposals/0000-overloaded-bracket.rst
mpickering: thanks.

proposals/0000-overloaded-bracket.rst
int-index: Ah, but then there's `fmap`, so this class

```
class TypedQuote p where
   fromBracket :: m (TExp a) -> p m a
```

must become

```
class TypedQuote p where
   fromBracket :: Functor m => m (TExp a) -> p m a
```

mpickering:
@goldfirere Also the point about top-level splices is that you still need to use a concrete type at the top-level splice and the maximal type of a quote is `Q Exp`, therefore a top-level splice expects an argument of type `Q Exp`.

int-index:
The "Rendered" link is outdated now, perhaps use https://github.com/mpickering/ghc-proposals/blob/overloaded-proposal/proposals/0000-overloaded-bracket.rst instead of a link to a specific commit?

int-index:
In the current proposal, I see these definitions:

```
class Lift a where
   lift :: Quote m => a -> m Exp
   liftT :: Quote m => a -> m (TExp a)
```

Question: what instances of `lift` would make use of `Quote m` methods? Why can't we have a pure `Lift`?

```
class Lift a where
   lift :: a -> Exp
   liftT :: a -> TExp a
```

mpickering:
`DeriveLift` is implemented in terms of quotation brackets, so if we want to do that it's more natural to generalise the type of `Lift`.

int-index:
Okay, then why the `Quote` constraint instead of `Applicative` or `Monad`? I'd favor some constraint that could be satisfied by `Identity`.

mpickering:
Because I think it's more consistent to always always give quotations a `Quote m` constraint rather than relying on the internal implementation detail of how the quotation is desugared. 

The `Quote` constraint can be implemented purely so I don't think it's any significant difference from `Applicative` or `Monad`.

Ericson2314:
@mpickering wait are you OK with https://github.com/ghc-proposals/ghc-proposals/pull/246#issuecomment-508536543 ? It *precludes* a newtype combining the syntax and the monad from being an instance of `IsTExp`. But if so, I'm happy to factor out second half of this proposal according to that design on your behalf.

Ericson2314:
> Because I think it's more consistent to always always give quotations a `Quote m` constraint rather than relying on the internal implementation detail of how the quotation is desugared.

I think the quote-everywhere is more forwards-compatible too.

Off topic. FWIW In some more dependent future I want to make a Reader + Writer thing based off https://cs.brown.edu/~sk/Publications/Papers/Published/pkw-inf-scope-syn-sugar/paper.pdf where you pass around references to the binding sites in a statically keyed map (as in row-types, or some equivalent). The macro evaluation does it's own name resolution, alpha-equivalent by construction, and custom binding syntax is expressible and enforceable.

mpickering:
@Ericson2314 I am not happy until it's possible to write instances for `Code`, that's the whole point of the `newtype`. And looking at your suggestion again, that doesn't seem to be possible.  

So no, I'm not actually in favour of it it seems.

Ericson2314:
@mpickering OK drat, but yeah that makes more sense too :).

goldfirere:
@mpickering 

2. Yes, that makes sense. But the proposal still talks about "top-level splices" and so it's not clear to me how the proposal matches your intent.

4. I think the back-compat breakage from generalizing untyped quotes would be minimal. There might be a few ambiguous types, I suppose. In any case, it sounds like this other idea is becoming a separate proposal, so we can debate it there.

6. OK. I think I understand this better. But it still seems that the proposal text is contradictory on this point, between specification point (4) and "Other effects".

7. Could you give a real-world example of a hand-written `Lift`? Why is it hand-written? And what do you feel about abandoning back-compat for such instances (while still retaining support, perhaps via `CPP`)?

> Also the point about top-level splices is that you still need to use a concrete type at the top-level splice and the maximal type of a quote is `Q Exp`, therefore a top-level splice expects an argument of type `Q Exp`.

But this proposal looks, on the surface, that it doesn't touch top-level splices at all. So I'm unsure what to infer from this description about them.

> The "Rendered" link is outdated now

I took the liberty to fix this myself.

Thanks for your love and care to TH (and its typed variant). It is much appreciated!

mpickering:
@goldfirere 

2. A top-level splice still needs to expect something of type `Q Exp`. Is that what you are asking? 
Later on in your comment you say that it doesn't touch top-level splices but that is explained in bullet point 4.

> A top-level splice instantiates `m ~ Q` and operates as before.

I also see that I explained this in a previous comment so perhaps you could explain with an example what you think the confusion is? Top-level splices have to instantiate `m` to `Q` so that there is a concrete type at the top level. Nested splices don't need to do this.

5. There are some (potentially dubious) hand-written instances in `th-lift-instances` (http://hackage.haskell.org/package/th-lift-instances-0.1.13/docs/src/Instances.TH.Lift.html). I just don't see a particular reason to make it impossible to define for users, it doesn't present a safety problem like `Typeable`.

simonpj:
* I'm generally sympathetic to making things more general.  But generalisation has a price, and without compelling use-cases I'm wondering if the cost is justified.  You say "Detaching quotations from Q makes way for a form of "pure" template haskell which can be fully evaluated on the host without having to run effects on the target." But I don't really understand that; it's far from a concrete use-case.   Can you perhaps give a few concrete examples of how useful this feature could be?  For example, it gives you access to a pure Haskell parser, doesn't it?

* I think the proposal would be stronger if it explicitly said it was about _untyped_ TH, and explicitly makes no change to _typed_ TH.   I spent some time hunting for the  answer to "what happens about typed TH?".

* Expression quotes are mentioned but nothing about type, pattern, or declaration quotes.  Are they covered?  Perhaps the proposal could be explicit, and give examples.

* If it doesn't cover typed TH, why is the type of `liftT` changed?  (Maybe `lift` and `liftT` should be in different classes?)

* I think the discussion about typing could readily be resolved simply by giving the typing rules.  Can you do that, in the proposal?  For example:
| A top-level splice instantiates m ~ Q and operates as before.
  All you need say is that the typing rule for a top-level splice is 100.00% unchanged:
  ```
     G |- e :: Q Exp, C
     runQ e --> e'
     G |- e' :: ty, D
     ------------
     G |- $e :: ty, C u D
  ```
  That is, in a top-level splice `$e` we expect that `e :: Q Exp`.  There is no `m` to instantiate.  (There might be if it so happened that `e :: forall m, Quote m => m Exp`, but there no requirement that it has such a type.

  The rule for nesed splices does change though.  Looking at the rules in the [original paper](https://www.microsoft.com/en-us/research/publication/template-meta-programming-for-haskell/) I think we need some adjustment; I think we need to carry along which 'm' is beind used for this bracket.   Perhaps the "State" in Fig 2 needs to be (B m) rather than just B.

In general, we lack a single place where the typing rules for untyped and typed TH are laid out.  (Not your fault, of course.) As we debate this proposal, another about nested quotatations, and perhaps another about overloading typed TH quotes, I acutely feel the lack of such a specification.  Matthew: you are absolutely on top of all this -- could you help us along by writing the typing rules?  Thanks!  (Probably Latex is best for that, rather than markdown. The original paper predates the typed/untyped TH split, and so may need adjustment.)

mpickering:
> In general, we lack a single place where the typing rules for untyped and typed TH are laid out. 

Yes I can write down a type system which contains both untyped and typed template haskell but probably not until after ICFP. There needs to be some additional care taken to deal with top-level splices but I have thought about how to account for this.

goldfirere:
Simon says:

> All you need say is that the typing rule for a top-level splice is 100.00% unchanged:

Is that true? Then that answers my question about "instantiating `m ~ Q`". My problem there is that, if I write `$( return (LitE (IntegerL 3)) )`, there is no `m` that I can see. So I don't know what you mean by instantiating.

In any case, I agree with Simon that your specification point (4) is in sore need of a typing rule. I can't really understand it as is. At the risk of seeming rude, I will explain how I can't understand it, trusting that you would want me to: "Refine the rules to do with splicing." This looks simply introductory. The second sentence, if I understand it now, is a no-op, saying that top-level splices don't change. "The use of nested brackets..." Now I'm lost, because the introductory sentence is about splices, but here you're talking about brackets. Maybe this sentence is to avoid readers thinking about quotes here? "A quotation that contains splices..." Now we're getting somewhere, but the sentence refers to the "constraints on the representation" and I don't know what these are. (The representation of what?) It would all be cleared up with a typing rule.

int-index:
> My problem there is that, if I write `$( return (LitE (IntegerL 3)) )`, there is no m that I can see. So I don't know what you mean by instantiating.

```
f :: Monad m => m Exp
f = return (LitE (IntegerL 3))
```

When one writes `$f` inside a bracket, e.g. `[| $f + $f |]`, it remains polymorphic, so we have `[| $f + $f |] :: Monad m => m Exp`.

On the other hand, if one writes a top-level splice, `m` gets instantiated to `Q`:

```
x :: Int
x = $f   -- m ~ Q
```

mpickering:
Thanks @goldfirere, I can now see that I made a mistake in this section. "nested brackets" should be "nested splices". Nested brackets are not relevant to this proposal as they are currently forbidden. I'll try to clarify this section tomorrow.

goldfirere:
@int-index yes, that was what I had understood from Simon's post. But I have to say that this is far from clear from the proposal. Thanks for writing that up.

mpickering:
@simonpj I think the argument about being able to execute splices on the host is good enough on its own. It's quite a serious problem with the current design which uses the `Q` monad. 

@goldfirere I updated (4) now, can you see whether you think it's clearer?

simonpj:
> I think the argument about being able to execute splices on the host is good enough on its own. It's quite a serious problem with the current design which uses the Q monad.

It satisfies you, but I don't understand it.  Perhaps I (and likely other readers) need a short tutorial.  "Here is a TH program that I cannot write today, but with this proposal it will work fine.  And here is why these kind of programs are important in practice."

Also, while that may be persuasive, adding further motivation may help to motivate further groups of users, who don't care about your main resaon, but may be excited by other things.  For example, with this proposal I think I could write
```
(runParse us [| \x -> x + 1 |]) :: Exp
```
where `runParse :: UniqueSupply -> P Exp -> Exp`, and `P` is a custom monad supporting `Quote`, to get an `Exp` in my hand that is parsed by the full Haskell parser.  Right?   That's nice.  Why not say so?

goldfirere:
Yes, (4) is much clearer now. It's still not quite a specification -- it just presents a characteristic example -- but that's fine for now. I still say the `m ~ Q` part is confusing, but I can safely skip over it given the clear statement about top-level splices above.

Ericson2314:
Wait why did we remove the analogus changes to typed Template Haskell? I'm all for moving the `p` stuff for a separate proposal but what about `liftT` and splicing `TExp`?

Ericson2314:
I guess it was @simonpj's comment

> I think the proposal would be stronger if it explicitly said it was about untyped TH, and explicitly makes no change to typed TH. I spent some time hunting for the answer to "what happens about typed TH?".

I thought that had to do with the removing of the `p` stuff for later, and there maybe being some prose left behind. I can't think why we would want to require `Q` for typed TH but not for untyped TH, when neither needs it more than the other.

J-mie6:
>  "Here is a TH program that I cannot write today, but with this proposal it will work fine. And here is why these kind of programs are important in practice."

I have a concrete use case that I can't do in Haskell right now, but causes me no end of pain and this proposal would fix. 

TLDR; Sometimes you really want to be able to push effects through code generation!

Suppose for instance, that you wanted to generate a bunch of code for a DSL that might be recursive. The way I would be doing this is to carry around a map of names to code (QExp) with me (in a `Reader` monad), and when I find a recursion point referencing a specific name, I pull the code out of my hat and splice it in. When I need to tie the knot, I can write something like this (for a loop say):

```
generateLoop :: String -> Reader CodeMap (Q Exp)
generateLoop name = ask (\codeMap -> [|
  let loopyCode x = 
    $(runReader loopBody (Map.insert name [|loopyCode|] codeMap))
  in loopyCode ...
  |])
```

That's all well and good, but hopefully you're already thinking that it would be better to be able to write something like:

```
generateLoop :: String -> Reader CodeMap (Q Exp)
generateLoop name = [|
  let loopyCode x = 
    $(local (Map.insert name [|loopyCode|]) loopBody)
  in loopyCode ...
  |]
```

Unfortunately, you can't, because the local splice _doesn't_ have a type `(MonadReader CodeMap m, Quote m) => m Exp`, but perhaps we can make our peace with that. With the proposal, however, this would be written as:

```
generateLoop :: (MonadReader CodeMap m, Quote m) => String -> m Exp
generateLoop name = [|
  let loopyCode x = 
    $(local (Map.insert name [|loopyCode|]) loopBody)
  in loopyCode ...
  |]
```

But this gets worse! Now suppose you might have mutually recursive things to generate code for: one of them is going to not exist in the map when it's asked for! So perhaps you decide that you make the stack a bit bigger: `ReaderT CodeMap (Except MissingDependency) a`. The general idea is that you can handle the "exception" generated and compile a new binding on the fly, nice. However, the `generateLoop` example from before doesn't suffer from this dependency problem, but because the Except monad isn't part of type of the code we want to splice in, we have to always handle our exceptions inside the splice and return around the outside again, duplicating the error handling code to places it doesn't need to be.

Obviously, my use case is actually in Typed-Template Haskell (slightly more complex code required), but the same principles apply there! ;)

goldfirere:
> I have a concrete use case that I can't do in Haskell right now, but causes me no end of pain and this proposal would fix.

Great -- exactly what we're looking for.

> Obviously, my use case is actually in Typed-Template Haskell (slightly more complex code required), but the same principles apply there! ;)

Oh. :(  So this proposal, silent on typed TH, wouldn't fix?

While I see how the same principles apply, there are *lots* of things that could be improved, in principle. Concrete use cases are very helpful in prioritizing effort, and it would be a shame to put in effort in designing, implementing, and maintaining this feature without people who are actually using it. In addition, perhaps porting these principles to the typed setting will reveal new complication (or simplification) that changes the design of the feature so that it's actually useful for your use case.

J-mie6:
> Oh. :( So this proposal, silent on typed TH, wouldn't fix?

Well... Like I said, same principles apply. I could just as easily be doing what I'm doing in TTH in UTH, all I'd do would be to remove the types! The real meat is that I would love to be able to thread effects through as I described in my comment, and it's really a problem to the point that I'd even consider giving up the types if this was implemented. The first step towards making this a reality is implementing for ntyped Template Haskell, evidently, but that will serve as a stepping stone towards supporting it in TTH as well. Correct me if I'm wrong @mpickering.

goldfirere:
>  I could just as easily be doing what I'm doing in TTH in UTH, all I'd do would be to remove the types!

Indeed! But this still leaves the possibility open that we'll go through hard work to build a feature that, as implemented, no one would use. Perhaps you would switch to use it, and I wouldn't ask you to commit to that, but perhaps not.

> The first step towards making this a reality is implementing for untyped Template Haskell, evidently, but that will serve as a stepping stone towards supporting it in TTH as well.

I don't see the untyped implementation necessarily as a stepping stone for a typed implementation. They could be independent. If anything, you're arguing *against* this proposal, as it doesn't address typed TH. Can this approach work in typed TH? Quite possibly, but we'd need to see the design laid out to be sure. Then, if we observe that the two designs are very similar, and that implementing the untyped proposal would be easier, we can view this current untyped proposal as a stepping stone. It's just hard to consider this a step on a path when we haven't yet defined the destination we're going to.

Just to be clear: I'm not at all arguing against the merits of this proposal. I'm just observing that we haven't yet satisfied Simon's "Here is a TH program that I cannot write today, but with this proposal it will work fine. And here is why these kind of programs are important in practice." I think having a concrete, boots-on-the-ground reason to add a feature to GHC is important.

J-mie6:
> I'm just observing that we haven't yet satisfied Simon's "Here is a TH program that I cannot write today, but with this proposal it will work fine.

Well, really I have, it's just there are two parallel versions of every concrete example you can come up with, one that exists in UTH and the other in TTH. Just because my concrete use case would ideally be TTH doesn't make this any less of an issue that would affect the same stuff being done by people working with UTH.

Ericson2314:
@goldfirere Per my https://github.com/ghc-proposals/ghc-proposals/pull/246#issuecomment-512334476 we just need to go back to a slightly earlier version of the proposal that did typed TH too, and then we have our example.

simonpj:
To be clear, I'm fine with a proposal that covers (a) UTH, or (b) TTH, or (c) both.  I would just like
*  it to be clear what is covered and what is not. 
* to have concrete, compelling use-cases that demonstrate how useful the feature would be
* to have as precise a description as possible -- thank you so much, Matthew, for ofrering to write typing rules

I think UTH and TTH have turned out to be quite substantially different.   UTH can handle type, pattern, and declaration splices; it runs in the renamer so you can see much less through reification; and you can decompose expressions as well as compose them.   TTH gives much much stronger typing guarantees, but only works on expressions and you can only compose terms (built by quotes) never decompose them.   I feel guilty that we have no articulate explanation of these two approaches; I would LOVE someone to write one.

But the differences make me think that it is not a small step to go from (a) to (c).

Interesting @j-mie6's example needs a monad that is _richer_ than Q; while as I understand it Matthew has in mind using a monad that is _less rich_ than Q.   Both are interesting!

Ericson2314:
@simonpj I agree they are different, and have no problem with them becoming more different (as say my https://github.com/ghc-proposals/ghc-proposals/pull/243 would allow). But I think this change cleanly generalizes to both of them, and benefits both of them alike. I wrote https://github.com/mpickering/ghc-proposals/pull/1, restoring the removed `quoteT` and adding more prose to explain its usage, to this end.

JacquesCarette:
I am amused that no one seems to have noted that the current proposal moves in the direction of 'tagless final' style encoding of meta-programming. Being polymorphic on the effects that can be encountered while doing code manipulation is quite useful indeed.  [Even more so, in my case, in TTH, but it's useful in UTH too].

I would like a slightly different route forward than @simonpj asks for regarding UTH vs TTH: I would love a type system that internalizes what it means to have typed types, typed patterns and typed declarations; that way TTH could 'catch up' to UTH.  [I have some ideas on how to do all three, but I'd need a partner-in-crime to write it all up, as I'm not sure it would rise high enough on my 'ToDo' priority list to get done in less than a decade...]

Ericson2314:
@JacquesCarette I wouldn't quite call it tagless final, as one still can eliminate syntax the usual way.

That said, I do agree with you that UTH and TTH are not so fundamentally different. If we had GADTs for each of `HsExpr Rn` and `HsExpr Tc` that enforced all the scoping guarantees, I think TTH would be exactly the same except for the (reflected) IR being outputted---we wouldn't need to be at all careful about the typing rules as the GADT indices would keep us safe.

Also, I would perhaps be interested in exploring these things with you. I have been working with @mpickering already, and I think the wider gamut that is explored the better the eventual design will be :).

JacquesCarette:
Sorry for the super slow response time - got overloaded.  It's not quite tagless final (I think I ought to know that...), but if you look at various later papers on the topic, you'll see that this is one short step away. The point is to be polymorphic over representation, which allows for multiple interpretations.

I'm still quite interested in all of that. I just can't promise fast turn-around. But if you're willing to live with that, then I'm happy to help explore the design space.

tydeu:
If we are making quotations polymorphic, why force them to be a monad? It would be very convenient to use the quotation syntax for functions that take pure TH. 

For example, in your proposal you have the line:

> ```haskell
> f :: (Quote m, C m) => m Exp
> f = [| 5 | ]
> ```

Why can't we do something like:

```haskell
f :: Exp
f = [| 5 |]
```

mpickering:
@tydeu This is explained in the proposal. You need a name generation effect in order to create a value of type `Exp`. You can run it purely if you wish using `State` to implement the effect.

tydeu:
@mpickering 

Why is name generation needed to implement quotations? In practice, you don't always need access to `newName`  to make an expression (for instance, in my previously given example). In fact, I would argue that you don't need it most of the time. When I am coding TH, I generally use non-monadic functions and use `mkName` instead.

Is the argument that that is what GHC currently does? From my understanding of the proposal it seems to be something of a revamp of the quotation syntax altogether, so I don't see why it couldn't be changed to not require that. Or am I misunderstanding?

int-index:
> In fact, I would argue that you don't need it most of the time.

You need it to bring fresh names into scope, so `[| \x -> ... |]` or `let x = ... in ...` will use `newName` to generate a fresh `x`.

> you don't always need access to `newName` to make an expression (for instance, in my previously given example)

Yes, `[| 5 |]` should actually have the type `forall m. Applicative m => m Exp`, without the `Quasi` constraint, because it only uses `pure`. Then you can run it with `runIdentity` (instantiating `m ~ Identity`) without any name generation.

tydeu:
> > In fact, I would argue that you don't need it most of the time.
> 
> You need it to bring fresh names into scope, so `[| \x -> ... |]` or `let x = ... in ...` will use `newName` to generate a fresh `x`.

Couldn't `[| \x -> ... |]` or `let x = ... in ...` just as easily use `mkName "x"` instead of generating a fresh name? Sure that could cause name shadowing and/or name conflicts, but for some use cases that is perfectly fine (or even desirable). In fact, it is what I generally do when generating TH. It also makes more sense (visually at least) for `[| \x -> ... |]` to actually produce a lambda with variable named `x` rather than `x_<some-unique-id>`. It could even clean up a lot of the identifiers in the documentation of TH-heavy libraries (like `singletons`).

Personally, I would just like the option of producing pure expressions rather than monadic ones from quotations (it would make me use them a lot more often instead of rolling my own with custom build functions).

Potentially, we could get the best of both worlds by desugaring quotations in a manner akin to many other literals -- with a `fromQuote` method defined similar to this:

```haskell
class Quote m => IsQuote m x a where
  fromQuote :: m x -> a
```

With instances for `x = Exp`, `x = Type`, etc. However, I imagine there may be some complications with this approach (I haven't really thought about it that deeply yet).

mpickering:
I have come back to this proposal with the intent to finish it. To do so I consulted @simonpj's last comment about what the proposal needs to be in order to be accepted.

> * it to be clear what is covered and what is not.
> 
> * to have concrete, compelling use-cases that demonstrate how useful the feature would be
> 
> * to have as precise a description as possible -- thank you so much, Matthew, for ofrering to write typing rules

I can update the proposal to cover the first two points but producing typing rules will not be completed before christmas. It is a significant amount of work that I will complete but I don't think the proposal should be blocked on this. Other proposals are not required to write typing rules for more complicated extensions. 

So therefore, I am going to update the proposal in the following ways.

1. Add back in overloading of typed quotations as well because that's what we really care about. 
2. Add Jamie's motivation about doing effects during complicated staged programming. 

Performing effects whilst doing code generation is widely accepted as necessary so it's a shame that this proposal has languished so badly for months.

mpickering:
I have updated the proposal now and think the committee should review it.

michaelpj:
Rereading this proposal, I realise that I also suffer from the problem described in Motivation 2 in a similar setting: generating typeclass instances with TH, but with some monadic effects required. I have some quite hairy code that would be significantly simpler with this.

JacquesCarette:
Generally, I'm quite supportive of this.

Would you however consider a 2 class alternative?  I don't want to suggest names now, so let's go with the awful `QuoteA` and `QuoteB` for the sake of this comment.  `QuoteA` can be used for all the parts of `TExp` that do not actually need a name supply at all (so really are pure) while `QuoteB` would be for the rest (lambda, etc).

I guess this is closer to something comonadic: you're asking for 'environment effects' to be available for the computation to proceed.  It's not really effectful, in the sense that you are not asking for effects, it's the operational semantics that insist on it.  All the pure stuff can be immediately lifted.

simonpj:
We had a useful conversation about this proposal on Monday.  Vlad is going to write up some notes and/or offer a PR.

int-index:
> We had a useful conversation about this proposal on Monday. Vlad is going to write up some notes and/or offer a PR.

Right. Notes first.

-----

A Template Haskell quotation `[| ... |]` is just syntactic sugar for combinators from `Language.Haskell.TH.Lib`.

For example, the following expressions are equivalent:

* `[| f 5 $(spl) |]`
* `appE (appE (varE 'f) (litE (integerL 5))) spl`

A quotation `[| ... |]` has type `Q Exp`, and it is also the type on which the combinators operate:

```
appE :: Q Exp -> Q Exp -> Q Exp
varE :: Name -> Q Exp
litE :: Lit -> Q Exp
integerL :: Integer -> Lit
``` 

The `Q` monad is defined as follows:

```
newtype Q a = Q { unQ :: forall m. Quasi m => m a }
```

And `Quasi` is a class that supports various Template Haskell operations:

* Name generation: `qNewName :: Quasi m => String -> m Name`
* Error reporting: ` qReport  :: Quasi m => Bool -> String -> m ()`
* Reification: `qReify :: Quasi m => Name -> m Info`
* Input/output: `qRunIO :: Quasi m => IO a -> m a`

In order to get the `Exp` out of `[| ... |] :: Q Exp`, the user needs to run `Q`, and that requires an instance that would implement all of the operations above:

```
main = do
  exp <- runQ [| 1 + 2 |]
  print exp

-- output:
--   InfixE (Just (LitE (IntegerL 1))) (VarE GHC.Num.+) (Just (LitE (IntegerL 2)))
```

The available instances are `Quasi Q` and `Quasi IO` (the latter panics on reification). 

However, one might want to extract `Exp` from a quotation without `IO`. And this is easy to do with a few fake instances:

```
instance MonadFail Identity   -- no method implementations!
instance MonadIO Identity     -- no method implementations!
instance Quasi Identity       -- no method implementations!

exp :: Exp
exp = runIdentity (runQ [| 1 + 2 |])
```

This works because the operations of `Quasi` are *not used* in quotations, except for `qNewName` which is used to bind new names (e.g. to generate a fresh `x` in `[| let x = ... in ... |]`).

**This proposal is about giving `[| ... |]` a more truthful type, so that these unsafe instances for `Identity` are not needed.**

So, how can we achieve this? In a few steps

1. Unwrap the `Q` newtype in the types of quotations, so that `[| ... |] :: forall m. Quasi m => m Exp` rather than `[| ... |] :: Q Exp`. This could be easily achieved by changing the types of Template Haskell combinators slightly:

   ```
   appE :: Quasi m => m Exp -> m Exp -> m Exp
   varE :: Quasi m => Name -> m Exp
   litE :: Quasi m => Lit -> m Exp
   ```

   This is practically the same as what we have today, except we have inlined the definition of `Q` into the signature.

2. Relax the typing rules for quotations to allow constraints other than `Quasi`. That is, instead of `[| ... |] :: forall m. Quasi m => m Exp` we say that `[| ... |] :: forall m. C m => m Exp` where the constraint `C` is the conjunction of all constraints required by the splices within the quotation and the syntactic constructs used within the quotation.

   Let's say we have `spl1 :: MonadState s m => m Exp` and `spl2 :: MonadReader r m => m Exp`, then

   ```
   [| $(spl1) $(spl2) |] :: (Quasi m, MonadState s m, MonadReader r m) => m Exp
   ```

    Why this type? Easy: consider the desugared version

    ```
    appE spl1 spl2
    ```

    Here, GHC would emit `Quasi m` from the use of `appE`, `MonadState s m` from the use of `spl1`, and `MonadReader r m` from the use of `spl2`, resulting in:

    ```
    appE spl1 spl2 :: (Quasi m, MonadState s m, MonadReader r m) => m Exp
    ```

    The same process happens with the `[| ... |]` syntactic sugar.

3. Relax the constraints required by Template Haskell combinators. Instead of `appE :: Quasi m => m Exp -> m Exp -> m Exp`, we can just say `appE :: Applicative m => m Exp -> m Exp -> m Exp`, and similarly for other combinators. Then we would have:

    ```
    [| f 5 |] :: Applicative m => m Exp
    ```

    Why this type? Easy: consider the desugared version

    ```
    appE (varE 'f) (litE (integerL 5))
    ```

    The combinators `appE`, `varE`, and `litE` would impose the `Applicative` constraint. `varE` and `litE` would make use of `pure`, and `appE` would make use of `<*>`, but not much else is needed here.

    At this point, the `Identity` example would work without the shady instances:

    ```
    exp :: Exp
    exp = runIdentity [| 1 + 2 |]
    ```

    Sure enough, `Identity` has an `Applicative` instances, and `Applicative` is sufficient for this quotation.

4. The final step is to improve support for quotations that require name generation, but not other features of `Q`. Consider:

    ```
    exp :: Exp
    exp = _runSomething [| let x = 2 in x + x |]
    ```

    What could `_runSomething` be? Not `runIdentity` because `let x = ...` uses `qNewName` to generate a fresh name, and that would impose the `Quasi m` constraint. And yet, `Quasi` is an overkill here: we don't need reification, we don't need input/output, etc. So we could move `qNewName` into a superclass of `Quasi`, which the proposal calls `Quote`:

    ```
    class Applicative m => Quote m where
      qNewName :: String -> m Name

    class (Quote m, MonadIO m, MonadFail m) => Quasi m where
      ...
    ```

    `Quote m` could be satisfied by something stronger than `Identity` but weaker than `Q` or `IO`. For example, `State NameSupply` is a perfectly fine monad that could provide name generation (where `type NameSupply = Int` or similar).

int-index:
@simonpj Does the summary above cover all important points? @mpickering Would you prefer to incorporate this text into the proposal yourself, or would you prefer me to make a PR?

mpickering:
I agree with @int-index's summary but I'm also a bit confused as it contains all the details which are already in the proposal. So if you want to make a PR to clarify some of the points then please feel welcome.

The only part I disagree with still is that the type of the combinators should reflect only on the parts of the interface they use. For example the type of `[| 5 |]` should still be `Quote m => m Exp` rather than `Applicative m => m Exp`. Any other choice exposes the implementation details of a particular combinator to the user. 

No one has provided any substantial argument about why breaking abstraction is a good thing here. The fact that the desugaring of an application doesn't require name generation is an implementation detail which shouldn't be exposed to the user. Other metaprogramming frameworks (LMS) desugar all expressions into ANF which requires fresh names for application.

simonpj:
I found @int-index's writeup extremely helpful.   I that in your mind it's all there, Matthew, but I totally didn't get it.

To take one example, the second sentence of the proposal says

> This proposal is changing the type of the representation to the polymorphic Quote m => m Exp

But that's not true, is it?  You plan to add all the constraints from the splices.  Moreover, it's not a fixed `Quote m` but rather the conjunction of constraints.  Or is that an extension that Vlad is proposing?

Anyway, Vlad, do offer a PR!

mpickering:
@int-index Are you going to offer a PR to clarify some of your points? I am not going to modify the proposal.

int-index:
> Are you going to offer a PR to clarify some of your points? 

I was going to make the PR but then I realized that I didn't know how to incorporate this point:

> The only part I disagree with still is that the type of the combinators should reflect only on the parts of the interface they use. For example the type of `[| 5 |]` should still be `Quote m => m Exp` rather than `Applicative m => m Exp`. Any other choice exposes the implementation details of a particular combinator to the user.

I'm not strongly opposed to that, but it's not the story that I'm telling in https://github.com/ghc-proposals/ghc-proposals/pull/246#issuecomment-548031138 (where I explicitly mention `Applicative` and `Identity`), and I haven't had the time to rewrite the thing.

If you're asking because you want to submit to the committee, do not let me hold you up.

int-index:
Judging by https://github.com/ghc-proposals/ghc-proposals/pull/246#issuecomment-543112610, I would guess @mpickering intended to submit this to the committee. Pinging @nomeata

nomeata:
There was a significnat amount of discussion afterwards, so I’d rather have an explicit “latest discussion is reflected in proposal, if applicable, please submit” from @mpickering

mpickering:
Yes the latest discussion should be reflected in the proposal. Please consider it!

proposals/0000-overloaded-bracket.rst
```diff
@@ -0,0 +1,394 @@
...
+   conducted by Richard, it was found that the only effect from
+   ``Q`` which was used is the ``newName`` function which generates a fresh
+   name. All the other combinators can be defined using the ``Applicative``
+   operations.
+
+
+3. Generalise the ``Lift`` type class::
+
+      class Lift a where
+         lift :: Quote m => a -> m Exp
```
aspiwack: I'm wondering if there should be a comparison with

```haskell
class Quote m => Lift m a where
  list :: Quote m => a -> m Exp
```

This would allow for more flexible `Lift` instances, which would work in some quotations but not other. On the other hand, we may actively not care about such instances, and maybe type inference wouldn't be as good?

Still, it probably makes sense to mention it in the alternatives.

proposals/0000-overloaded-bracket.rst
```diff
@@ -0,0 +1,394 @@
...
+Unresolved Questions
+--------------------
+
+* Carter points out that if you want to achieve "pure" template haskell then
+  you still need to deal with the fact that different platforms have different
+  representations of primitive data types. This is out of scope of this
+  proposal.
+
+* It would also be possible to make ``Quote`` a superclass of ``Quasi`` but
+  this hierarchy refactoring seems unecessary.
```
aspiwack: On the other hand, is there any situation where we may want ``Quasi`` without ``Quote``?

simonpj:
In the [typeset proposal](https://github.com/mpickering/ghc-proposals/blob/overloaded-proposal/proposals/0000-overloaded-bracket.rst) the "discussed at this pull request" link is broken.   Could this be fixed please?

simonpj:
> Extracting the Exp from Q Exp requires unsafe partial functions (see Appendix A: PureQ).

I had trouble making sense of this.   What I think you mean is this.  Given a quote `[| blah |]` with no splices, it should be possible to extract an `Exp` with nothing more than a name supply.  But actually one has to fake up a Quasi monad with many error stubs, which is unsafe (see Appendix PureQ).  This is undesirable.

simonpj:
Some other questions/thoughts

* What about Typed TH quotes.  Are they affected?  Should htye be?  Please say.

* What about quotes that do not need a name supply, like `[| map inc [1,2,3] |]`?   The combinators for that don't even need a name supply, so do we get away with Applicative for them? 

* The idea of gathering up constraints from splices is a good one.  But it's a significant new thing.  In parsentational terms think that "Consequently, generalize the type of quotation brackets ..." is far to quiet a way to introduce it.   Better to say that this proposal has three parts not two (Motivation section)

* I would love to have a list of all the API changes.  Yes, I know there are a lot.  It's be fine, say, to do `appE` and then list all the others that are generalised in exactly the same way, rather than give the new signature for each individually.    But there may be more API changes, eg Lift-related?

proposals/0000-overloaded-bracket.rst
```diff
@@ -0,0 +1,394 @@
...
+  ``DeriveLift`` is the blessed manner in which to define a ``Lift`` instance.
+
+Alternatives
+------------
+
+* The main alternative to the design would be to only require a ``Quote``
+  constraint when the quotation requires the ``newName`` effect. For example,
+  ``[| 5 |] :: Applicative m => m Exp``. I am opposed to this direction as it
+  breaks abstraction. The implementation detail of how ``[| 5 |]`` is desugared
+  leaks to the user.
```
nomeata: I don’t quite understand why this is worse: Why not have a less restrictive type if possible?

Can we draw an analogy, where `do e` does _not_ have a `Monad` constraint; the constrait only gets added if we add more lines. And `MonadFail` is only added if there is an incomplete pattern match? Seems analogous to this case.

nomeata: I guess this ties to Simon’s question

> What about quotes that do not need a name supply, like [| map inc [1,2,3] |]? The combinators for that don't even need a name supply, so do we get away with Applicative for them?

which is partly answers by this section, but maybe not sufficiently motivated?

gridaphobe: There's value in presenting a consistent API. Tagging all TH syntax combinators with the `Quote` constraint indicates that they are meant to be used together, much like the current definition of everything in terms of `Q` does.

There's also value in requesting the minimal required capabilities as it makes the API more flexible, but this usually becomes important when the capabilities are heavyweight somehow. A database connection might be heavyweight, but a name supply seems pretty light. So I'd probably lean towards using `Quote` consistently.

tydeu: @gridaphobe 
TH already has a consistent API, namely `Q`. The whole point of this proposal (as I see it) is to reduce the heavyweight `Q` to the lighter weight `Quote`. This will make the API inconstent, as reification and other such things will still need `Q`. Since this proposal is requesting that we reduce `Q` to its minimal required capabilities (at least for splices), why not go all the way and reduce it to `Applicative` in the cases where that is possible?

gridaphobe:
There are two parts to the motivation. The more compelling piece to me is making the API polymorphic in the monad, to support more complex use cases. That could be done without introducing a new class, just define everything directly in terms of the existing Quasi class.

But the proposal also points out that constructing syntax trees doesn’t require anything beyond a name supply, and you might want to do this in pure contexts. If you use Quasi here, you’d have to ‘error’ out a large portion of the interface, which is ugly. So we introduce a new class Quote to better support this use case. 

We could go further and use the minimal constraints for each individual function, but that makes the API even less consistent and doesn’t provide an obvious benefit to me. Is there a concrete use case that would be better supported by doing this?

nomeata:
>  Is there a concrete use case that would be better supported by doing this?

I certainly have none. I am only raising it because the motivation for not going there wasn’t very clear  to me. If it is to you (as it seems to be), then that’s good enough for me!

Ericson2314:
TBH, I feel like the "how fined-grained are the constraints" issue is easier to tweak over time, and I personally really have little-to-no opinion. On the other hand, the issue of what to do with typed-TH seems more pressing, especially since #195 and this seem to conflict more severely than the conversion for either suggests (though I'd love to be wrong!), making me nervous. I'd hope we can sort the big picture regarding both proposals, and only then drill down into the details.

int-index:
This proposal and #195 can be combined by changing

```
newtype Code a = Code (Q (TExp a))
```

to

```
newtype Code m a = Code (m (TExp a))
```

And the constraints such as `Quote`, `Quasi`, `MonadReader r`, etc, are to be placed on `m`.

Ericson2314:
#195 certainly can be so changed, but I worry that the change is against it's spirit if not it's letter. I see #195 as an attempt to hide "implementation details" that also mildly interfere with invariants. But if we expose a monad parameter like that, we're saying the monad is a rightful part of the interface rather than implementation detail, in which case the `Code` datatype (IMO) appears like an odd attempt to ram together two independent axes of variation into one type, very much unlike the way we usually deal with monads.

simonpj:
Dear Matthew,

There is general support from the Steering Committee for this proposal, so it's set to be accepted.

But before doing so, could you please update the proposal to respond to the disucssion above, starting around [this comment](https://github.com/ghc-proposals/ghc-proposals/pull/246#issuecomment-554449591)?

While you do that I'll change it back to "needs revision", but with the understanding that it'll be accepted once you've had time to make the proposal reflect the points above, and re-submit it.

Thanks!

Simon

proposals/0000-overloaded-bracket.rst
mpickering: > #195 certainly can be so changed, but I worry that the change is against it's spirit if not it's letter. I see #195 as an attempt to hide "implementation details" that also mildly interfere with invariants. But if we expose a monad parameter like that, we're saying the monad is a rightful part of the interface rather than implementation detail, in which case the `Code` datatype (IMO) appears like an odd attempt to ram together two independent axes of variation into one type, very much unlike the way we usually deal with monads.

@Ericson2314 I think the primary concern was with people using effects from ``Q``, using effects during code generation is a completely natural thing to do in both typed and untyped metaprogramming.

proposals/0000-overloaded-bracket.rst
mpickering: I'm not aware of anyone definition a `Quasi` instance for themselves so it seems difficult to answer.

proposals/0000-overloaded-bracket.rst
mpickering: This is quite a big breaking change which the proposal has made quite an effort to avoid. It would be better to do this later in an wholesale change to `Lift` rather than incrementally.

mpickering:
I updated the proposal I think to cover the points from the discussion.

* A little discussion on why we use `Quote` all the time rather than sometimes just `Applicative`. With a small comparison to `MonadFail`. 
* Reword section about extracting `Exp` based on Simon's suggestion.
* Move the section about collecting constraints into its own section in the proposed changes.
* Provide some more examples of how the combinators generalise. I didn't list the whole API as it is a mechanical translation from `ExpQ` to `m Exp`. 
* Expand section about typed quotations to take into account what would happen if #195 is accepted.

proposals/0000-overloaded-bracket.rst
goldfirere: I've hand-written `Quasi` instances profitably. See https://github.com/goldfirere/singletons/blob/master/src/Data/Singletons/Single/Monad.hs#L55. And I've used GeneralizedNewtypeDeriving twice: https://github.com/goldfirere/singletons/blob/master/src/Data/Singletons/Promote/Monad.hs#L49 and https://github.com/goldfirere/th-desugar/blob/master/Language/Haskell/TH/Desugar/Reify.hs#L206. So it's possible.

simonpj:
Thanks.  That's much improved.

> the varE function to `Name -> m Exp`

Shouldn't that have a `Quote` in it?

> 5.1  Instances written in terms of Q will no longer work

Could you add an example?   

> 5.3 Other effects

Isn't this covered under "Proposed changes (4)"?    I couldn't figure out what new wrinkle was being discussed here.

In Richard's audit, which you cite, he says

> unTypeQ and unsafeTExpCoerce: These are only about typed TH. Let's tackle untyped TH first.

But in fact this proposal does embrace typed TH.   So what about these functions?

More generally, could you make it clear, from the start, that the proposal covers Typed TH quotes as well as untyped ones?   (Otherwise we only find out in Proposed changes (5).)

> If at a later point Q (TExp a) is turned into a newtype 

Could you cite (and link to) the proposal on this?  It's in "recommend accept" state.

Once you are content, do re-submit to the committee.

mpickering:
@simonpj Do you have any opinion about whether we should add some special defaulting rules to stop errors about annotated top-level quotes. 

For example

```
foo = [| 1 |]
```

Does not typecheck without a type signature or `NoMonomorphismRestriction`. 

```
[nix-shell:~/ghc/overloaded-th]$ _build/stage1/bin/ghc A.hs

[1 of 1] Compiling A                ( A.hs, A.o )

A.hs:4:7: error:

    • Ambiguous type variable 'm0' arising from a static form

      prevents the constraint '(Language.Haskell.TH.Syntax.Quote

                                  m0)' from being solved.

      Relevant bindings include

        foo :: m0 Language.Haskell.TH.Syntax.Exp (bound at A.hs:4:1)

      Probable fix: use a type annotation to specify what 'm0' should be.

      These potential instance exist:

        one instance involving out-of-scope types

        (use -fprint-potential-instances to see them all)

    • In the expression: [| 1 |]

      In an equation for 'foo': foo = [| 1 |]

  |

4 | foo = [| 1 |]

  |       ^^^^^^^
```

simonpj:
This is just the monomorphism restriction, right?   I'd be inclined to leave it for now and see if we get user pressure to change.  

Perhaps your concern is back-compat?   But this is the least serious back-compat problem: it can be fixed by doing something *that would compile with the old compiler too*.   In this case, adding a type signature to `foo` woud render it compilable with both old and new compiler

mpickering:
When I was implementing it it was much easier to use `Monad` rather than `Applicative` for the superclass. This decision could be revisited later.

simonpj:
> When I was implementing the proposal it was much easier to use Monad rather than Applicative for the superclass

Can you elaborate on why?  Do you see it as a stop-gap measure, or are you inclined to change the proposal?

mpickering:
> > When I was implementing the proposal it was much easier to use Monad rather than Applicative for the superclass
> 
> Can you elaborate on why? Do you see it as a stop-gap measure, or are you inclined to change the proposal?

If you look at `DsMeta.wrapGenSyms` then it's not clear to me how to refactor that to use applicative or whether it's possible. The desugared code makes heavy use of `bindQ`.

mpickering:
> > 5.1  Instances written in terms of Q will no longer work
> 
> Could you add an example?

I don't know any examples and also Ryan didn't know any either so they must be exceedingly rare if they exist as well.

simonpj:
> I don't know any examples and also Ryan didn't know any either so they must be exceedingly rare if they exist as well.

Worth saying that!   Also an artificial example would be fine to demonstrate the potential problem you are drawing attention to.  It's doesn't have to be from the wild

simonpj:
Just to be clear, I'm waiting for you to say "I've completed my revisions of the propsal and want to resubmit it to the committee".   This way we are sure that the final proposal is in good shape, and accept it.  Thanks!

mpickering:
I have completed the revisions and wish to resubmit the proposal to the committee. The implementation is also finished an ready for review.

simonpj:
I have recommended acceptance to the committee.  Will commit on Friday.

//remind Friday

yav:
@mpickering it looks like this conflicts with #195 for the typed case as `Code` is not an `Applicative`, unless I am missing something?

I was wondering though, do we need the `Quasi` class to have `Applicative` as a super class?  Perhaps just a single method like this might be sufficient:
```
withNewName :: String -> (Name -> m a) -> m a
```
Thoughts?

mpickering:
I don't see how this conflicts with #195 at all. The fact that `Code` is neither an `Applicative` or `Monad` doesn't matter. You define `newtype Code m a = Code (m (TExp a)` like before and use the polymorphic `unsafeTExpCoerce` and `unTypeQ` functions to get at the actually monadic part to compose the fragments together, the desugaring will not need to change much.

yav:
Oh, I see, that makes sense---I hadn't realized you were thinking of changing `Code` also.   Please add that to #195 when you get a chance to update it.

Just to be perfectly clear,  typed expressions would look something like this, right? 
```
example :: Quasi m => Code m Bool
example = [|| True ||]
```

mpickering:
Yes that's exactly right.

nomeata:
/remind @simonpj about accepting this on friday.

simonpj:
Thank you Matthew for revising the proposal.    I declare it accepted.

Let's try to get #195 revised and accepted to; it would make sense to implement both at once.

Simon

nomeata:
But you still want me to merge this as it is now, right?

simonpj:
> But you still want me to merge this as it is now, right?

I believe so.  Matthew has already updated it.

Matthew: to double check: do you have any pending changes to this proposal?

mpickering:
Yes no blockers from me.

simonpj:
Tom Sydney Kerckhove (NorfairKing) asks: would the authors of this PR please weigh in on the discussion at [commercialhaskell/path#170](https://github.com/commercialhaskell/path/pull/170#issuecomment-775066947) 

    In our testing so far we've not found any Lift instances relying on any special methods of Q.

The path library seems to be an example of this, and I would like to ask for some advice on how to proceed over there.

int-index:
Possibly related: https://gitlab.haskell.org/ghc/ghc/-/issues/17565

yav:
I posted a possible solution on that thread.

By the way, as I mentioned there, it would be convenient to add a function that reinterprets a `Typeable` type as a Template Haskell type, as I couldn't find one, but maybe it already exists?

josephcsible:
This caused some breakage in `recursion-schemes`: recursion-schemes/recursion-schemes#132

Short explanation: that package has a typeclass like this:
```haskell
class MakeBaseFunctor a where makeBaseFunctor :: a -> DecsQ
instance MakeBaseFunctor a => MakeBaseFunctor [a] where -- ...
instance MakeBaseFunctor a => MakeBaseFunctor (Q a) where -- ...
instance MakeBaseFunctor Name where -- ...
instance MakeBaseFunctor Dec where -- ...
```

(Note that the implementations all eventually use `reify` internally, so they all really do need to run in `Q`.)

Anyway, all of the following used to work:
```haskell
makeBaseFunctor ''Identity
makeBaseFunctor [''Identity, ''Maybe]
makeBaseFunctor [d| instance Recursive (Identity a) |]
makeBaseFunctor [[d| instance Recursive (Identity a) |], [d| instance Recursive (Maybe a) |]]
```

But now the latter two fail with an ambiguous type error, since the types of the quotations are inferred to be `Quote m => m Decs`, which isn't enough information to pick the `MakeBaseFunctor (Q a)` instance. I'm not sure if there's a way that this can be fixed entirely within `recursion-schemes`, or if users of it will necessarily have to change their code that calls `makeBaseFunctor`.

int-index:
Consider changing

```
instance MakeBaseFunctor a => MakeBaseFunctor (Q a) where -- ...
```

to

```
instance {-# OVERLAPPABLE #-} (m ~ Q, MakeBaseFunctor a) => MakeBaseFunctor (m a) where -- ...
```

josephcsible:
Unfortunately, that doesn't help:
```
    • Overlapping instances for MakeBaseFunctor
                                  (m0 Language.Haskell.TH.Lib.Internal.Decs)
        arising from a use of ‘makeBaseFunctor’
      Matching instances:
        instance [overlappable] [safe] (q ~ Q, MakeBaseFunctor a) =>
                                       MakeBaseFunctor (q a)
          -- Defined in ‘Data.Functor.Foldable.TH’
        instance [safe] MakeBaseFunctor a => MakeBaseFunctor [a]
          -- Defined in ‘Data.Functor.Foldable.TH’
      (The choice depends on the instantiation of ‘m0’
       To pick the first instance above, use IncoherentInstances
       when compiling the other instance declarations)
```
Bizarrely, that error persists even if I do turn on `IncoherentInstances`.

int-index:
That’s too bad. What we want here is some form of defaulting, so that the ambiguous type variable would be instantiated to `Q`. But I don’t know if there’s a way to get it. Even a recent accepted GHC proposal #409 seems insufficient to handle this use case.

@blamario What would it take to extend #409 in such a way that it could handle this?

blamario:
I don't see why #409 would not cover this as-is. Admittedly the motivation and the examples included in the proposal are mostly about solving the `IsString` ambiguity, but there are also examples like

    default Foldable ([])

As [section 2.1](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0409-exportable-named-default.rst#21naming-the-class) states

> The types may belong to any kind, but the class must have a single parameter.

there's absolutely nothing preventing a declaration like

    default MakeBaseFunctor (Q)

In fact it might be a good idea to recommend

    default Quote (Q)

in general.

int-index:
> I don't see why #409 would not cover this as-is. Admittedly the motivation and the examples included in the proposal are mostly about solving the `IsString` ambiguity, but there are also examples like
> 
> ```
> default Foldable ([])
> ```
> 
> As [section 2.1](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0409-exportable-named-default.rst#21naming-the-class) states
> 
> > The types may belong to any kind, but the class must have a single parameter.
> 
> there's absolutely nothing preventing a declaration like
> 
> ```
> default MakeBaseFunctor (Q)
> ```
> 
> In fact it might be a good idea to recommend
> 
> ```
> default Quote (Q)
> ```
> 
> in general.

But wouldn’t this be ill-kinded? `Q` isn’t an instance of `MakeBaseFunctor`, but `Q a` is.

josephcsible:
I retract my previous comment. I'm not sure what I did wrong, but I just tried again and it worked this time.

int-index:
> my previous comment. I'm not sure what I did wrong, but I just tried again and it worked this time.

Now that’s surprising, because the error message you posted makes sense! I think if you got it working, then only thanks to `IncoherentInstances` (overlappable alone isn’t sufficient here), and that’s a fragile arrangement.

blamario:
> But wouldn’t this be ill-kinded? `Q` isn’t an instance of `MakeBaseFunctor`, but `Q a` is.

You're right, sorry I replied too quickly. I was lead from the class name to think it had the same kind as `Functor`. Still the other declaration I suggested should work to resolve the ambiguity:

    default Quote (Q)

int-index:
Yes, `default Quote (Q)` sounds like a great idea.

PR #248, opened 2019-07-09
WIP: Coercions without roles

Ericson2314:


treeowl:
A small point: when producing the Core coercion, we need to apply the transitivity enhancement from the paper:

```haskell
Coercible a b => Coercible (F a) (F b)
====>
(Coercible a b, Coercible (F a) q) => Coercible q (F b)
```

or something like that.

A somewhat separate element is that I really want to be able to offer these coercions in a local context. With `QuantifiedConstraints`, that tends to produce overlap problems. So I'd really like to be able to refuse to "export" certain of these coercion axioms, while still being able (in that module) to write functions exposing them. I can explain that a bit more later.

treeowl:
Oh, and I definitely agree with you; this is much more like what I'm after personally than #233, though the ideas are actually complementary; #233 would enhance this proposed system the same way it enhances the role system.

Ericson2314:
@treeowl Yes I agree. I actually coincidentally edited the top part! https://gitlab.haskell.org/ghc/ghc/issues/8177#note_212318 is in to me the killer app for #233.

Oh, and I should use your reflex PRs as part of the motivation. Nobody should have to work around GHC that much!

Ericson2314:
> transitivity enhancement 

What is this? I couldn't immediate find it in https://repository.brynmawr.edu/cgi/viewcontent.cgi?article=1010&context=compsci_pubs. I would have thought no special support is  needed:
```
c0 :: Coercion a b => Coercible (F a) (F b)
|- 
/\ (c1 :: Coercion a b) (c2 :: Coercible (F a) q) =>
  (syn c2 ; (c0 c1) :: Coercible q (F b))
```

treeowl:
So the local idea is pretty orthogonal. What I want to do is let people write things like

```haskell
deriving coercion foo :: ... => Coercible (T f x) (T g y)
deriving coercion bar :: .... => Coercible (T f x) (T g y)
```

naming alternative `Coercible` axioms. These names can then be exported/imported, either "open" or not. And the names can easily be treated as (CPS-style) functions as well, to apply that axiom locally. By not opening an axiom in a module, we should (I hope) be able to avoid some overlapping problems. I should probably write my own proposal for that, but I won't have time soon, I don't think.

treeowl:
@Ericson2314 I don't quite remember *why*, but that does turn out to be important. And it is in the paper.

Ericson2314:
@treeowl Yeah from your reflex PRs I started wanting better control over axiom export too. The constructor in scope rule is very all-or-nothing

treeowl:
I wonder if that transitivity thing was from an earlier version of the paper.... Could be. Yeah, I don't like that all-or-nothing, and not being able to say "No, you stupid compiler, don't pick the global `a ~ b => Coercible (T a) (T b)`! I have `Coercible a b => Coercible (T a) (T b)` in scope!"

treeowl:
Actually, I think this proposal probably *does* supplant the `DatatypeContexts` version of Richard's proposal, but not the one covering constructor constraints.

goldfirere:
I'm having a hard time understanding the details of this proposal in its current form. I'm looking, in particular, at the suggested changes to Core (and, therefore, Haskell's type safety). Should I ask detailed questions now or wait until the WIP is removed?

Fair warning: I will have lots of questions! They are made out of love, and with deep appreciation to @Ericson2314 for tackling this. I actually quite dislike roles in their current form, but yet much has been invested in them (in both design and proof). It will thus take quite a bit of technical detail (likely including a type safety proof) to convince me to move away from them.

Ericson2314:
@goldfirere sure, lay into it! What's I'm most aware is missing is an actual proof, more surface language change detail, and motivation, so if you have questions on the core that's great!

Ericson2314:
@goldfirere sure, lay into it! What's I'm most aware is missing is an actual proof, more surface language change detail, and motivation, so if you have questions on the core that's great!

mgsloan:
I haven't dug into the details here, but from a skim I like the idea of having an explicit list of constraints on Coercible instances instead of roles.  A misc bit of feedback think it would be good to add more examples to this doc, particularly the declarations that would correspond to the current roles system, and a motivating example for the new functionality.

Specifically, this might help resolve a pet problem of mine discussed in [a reddit post](https://www.reddit.com/r/haskell/comments/8h2lon/could_coercible_be_extended_to_work_on/) , which is to be able to coerce `Map Int v` to `Map WrappedInt v` if GND was used to derive the `Ord` instance for `WrappedInt`.  So, it might look like:

```haskell
deriving instance (Coercible k k', Coercible v v', SameInstances (Eq k, Ord k) (Eq k', Ord k')) => Coercible (Map k v) (Map k' v')
```

Currently map keys have nominal roles to prevent `coerce` from breaking Map data invariants.  This would allow specifying a more precise constraint which permits coercion of the key type when it uses identical instances (probably only GND instances).

I realize that this is well out of scope of this proposal, I don't want to distract too much from the main topic.  However, it may well be worth considering what other sorts of constraints might be useful here.

Ericson2314:
@mgsloan sure that is interesting and good to keep in mind. I think what you might really want for `SameInstances` is
```haskell
deriving instance
  ( Coercible k k'
  , Coercible v v'
  , Coercible (Eq k) (Eq k')
  , Coercible (Ord v) (Ord v')
  ) => Coercible (Map k v) (Map k' v')
```
This is good because it keeps class constraints out of the instance head, which is good keeping coercion cost-free since we don't want to have to evaluate anything.

In fact, this turns this into a purely orthogonal proposal. `Coercion :: Coercion (Eq a) (Eq a)` already is inhabited, to my pleasant surprise. `Eq` has a nominal role. So the tasks for that proposal are roughly
 1. Block `coNth` / `coPremise` from yielding `forall a b. Coercion (Eq a) (Eq b) => a ~ b`
 2. Add more axioms based on GND.

mgsloan:
@Ericson2314 Yes in my original proposal in the reddit post I reused `Coercible` on constraints.  @goldfirere [commented](https://www.reddit.com/r/haskell/comments/8h2lon/could_coercible_be_extended_to_work_on/dygxvj2/) on the post pointing out that using `Coercible` in that way might be imprecise.  Quoting his comment:

> I don't think `Coercible (Hashable Int) (Hashable N)` is exactly the right constraint: that constraint says that the runtime representation of `Hashable Int` is the same as `Hashable N`. However, this is always true, regardless of the instance for `Hashable N`.

Ericson2314:
@mgsloan Right that is true, but a) we already have overly conservative coercible instances today for a variety of reasons b) `forall (c :: k -> Constraint) (a :: k) (b :: k). Coercible a b => Coercible (c a) (c b)` "surely" can ruin cannonicity somehow so we should never admit it anyway.

treeowl:
I agree with @goldfirere that `Coercible (Eq a) (Eq b)` doesn't get the right point across. `SameInstance` makes more sense. In fact, I believe it makes sense to allow arbitrary constraints as "instance constraints" on the `Coercible` instance (not letting you get `Num a` out, for example). One significant challenge, I think, is working out how reasoning backwards needs to work with different `Coercible` instances in scope, especially if we're trying to do some things locally.

treeowl:
I think the solution to the reverse reasoning problem is probably just not to do it, or to do it in a very restricted way. Currently, given

```Haskell
data T a
type role T nominal
```

we can write

```haskell
f :: Coercible (T a) (T b) => a :~: b
f = Refl
```

But if we want to be able to import coercion axioms selectively, and write them flexibly, I think we need to give it up. It's a loss, but I think a fairly minor one.

Ericson2314:
@treeowl in the response to @mgsloan I was referring just to just backwards reasoning with classes, but yes I am tempted to give it up for all `Coercible`, as that is does seem to be the chief difference between it and a normal class, and the reason for much caution. I do fear it is too big a breaking change though. I should put it in the alternatives.

treeowl:
I don't actually think much real code is likely to break if the reverse reasoning is removed. But there may be a way around it: allow the user to specify the reverse axioms, with some syntactic help for fully bidirectional ones. But require that each reverse axiom be proven (by the compiler) compatible with *all* forward axioms.

treeowl:
Aside from the serious problems it causes for local use, reverse reasoning only ever works for `data`, not `newtype`. This is because there's no way to disable the newtype wrapping/unwrapping axiom in the defining module. That certainly suggests a compatibility check for injectivity axioms and the ability to disable wrapping/unwrapping even in the defining module.

goldfirere:
OK -- now that it's after the POPL deadline, I can, as you say, "lay into this". :)

Numbered only for backreference.

1. Motivation and/or examples will be needed at some point.

2. Your point (1): is this a change from your proposal? You say it's "not an implementation change"... is it a specification change? I just don't know how to interpret the text.

3. By `phi`, I suppose you mean the grammar for constraints. (It would help to say this explicitly.)

4. Presumably, you're keeping "old" with "new" in the description for `phi`. Thanks to quantified constraints, this all is true already. So I think point (1) doesn't say much.

5. I understand from (2) that every datatype declaration comes along with axioms now. OK.

6. Newtypes and type families create axioms today. But they are *not* available for `nthCo`. Is this a typo? I'm not sure what you mean here.

7. "The axiom is admissible under these conditions." I'm not sure what this sentence means. What axiom? What conditions?

8. Under **Shape**: What is "It"? Presumably, the axiom that arises when declaring `H`. Better to say so.

9. This axiom still uses representational equality. But if there are no roles (as the proposal title suggests), what does representational equality mean?

10. Under **Saturation**, I presume you're not talking about the existing axiom that arises from newtypes -- just the new ones. OK.

11. "If one looks at the type of each field under the two applications, they will accumulate per-field `f_n ~_R f'_n` constraints." The "two application" are presumably the `H a*` and `H b*` from a couple of lines ago. But what the "each field"s? Do you mean the definition of `H`, suitably substituted by the `a*` and the `b*`? This should be spelled out. Again, I'm very curious what `~R` means without roles.

12. **Not-too-incompleteness**: What is `C` there?

13. "roughly": We'll need to be quite exacting here, in the end.

14. `coTyConApp`: Presumably, `C` is the name of one of your axioms.

15. The rule needs a bunch more detail (knowing where the kinds come from, etc.). But I get the idea.

16. It's starting to make a bit more sense now. You're not proposing the abolishment of roles. That is, you still have two equality relations (nominal and representational) in your system. But some of the rules are different, and type constructors no longer are associated with a list of roles.

17. `premise` basically says that, at least for datatypes, the axioms are the *only* way to produce representational equality. This may be troublesome, in that I don't see a restriction that only one axiom is associated with a type declaration. Was that meant to be in the mix here?

18. I'm not sure how you would want to implement this. In its current form, it doesn't seem implementable as zero-cost. Specifically: constraints `phi` are used to imply representational equality. But constraints are *lifted*: they might be bottom. So we have to evaluate them in order to get the representational equality. Of course, evaluating anything takes time, and so we've lost our zero-cost guarantee. You might then argue that constraints shouldn't be lifted... and you might even have a good point there. But now we must certainly throw out quantified constraints, which really are functions. And these quantified constraints appear to play a pivotal role in your design.

19. I'm not going to worry so much about surface syntax until we know the idea is implementable.

20. "This is OK because we can rewrite any closed term with the correct type into one where the Axiom is at it's head!" I'm not sure at all what that sentence means. It sounds like you want to rewrite a term into something with an axiom... but that's a coercion. Ah -- maybe you're rewriting coercions to coercions (implied by your next point).

21. You describe that you can pull axioms to the front, but what if you're using one axiom as evidence for use in another? They can't both be at the front. And it's not clear that this trick will work with transitivity and symmetry mixed in. This proof is hard.

22. "I should rigorously show that this is true" You're already planning on it, but *should* must become *must*. This is the cornerstone of Haskell's type safety.

23. "If we required the axioms be complete, we could get rid of the no newtype restriction for Co_Premise." What does *complete* mean here? I actually can't picture what you mean in that alternative.

24. I don't think this will be so easy to implement. Roles are pretty well baked in at this point.

---------------------------------

Some of the quibbles above are about the writeup. My questions are honest (that is, I've been confused for real, not just pointing to nice clarifications for the sake of clarification) as I've read through, but by the end I do think I know what's going on. So, to me, this really boils down to (18) and (22): I'm very far from convinced that this plan holds water. Note that today's language of coercions does not contain functions (or anything like functions), and this is an essential property of coercions: they do not evaluate. If they did, they might not terminate, and then we'd be in deep trouble. (The syntax for coercions have things that look like functions. They are not functions if you look at how the typing rules work out.)

Again, I don't want to be overly discouraging. And I'll admit I have bias against this because I'm deeply invested (3 published papers) in the current solution. But it would be no good for any of us to move forward with a proposal such as this without a rock-solid proof of correctness. So there's a bit of a hill to climb here.

treeowl:
@goldfirere functions can be okay if they're guaranteed to terminate. These are functions written by the compiler in a very restricted language; we can make very sure they terminate!

goldfirere:
> we can make very sure they terminate

Famous last words -- at least in my world. :)

One challenge is that we can't simply "be careful" when constructing the functions. For Core Lint to really deliver on its promise of ensuring type safety, we need to be able to check for termination (and quickly). So we'd probably need some kind of restricted syntax that uses only primitive recursion, not general recursion.

treeowl:
@goldfirere Remember: we only need "more powerful than the current type roles" to start winning expressiveness. There's a lot of literature out there on termination checking and terminating languages. And since these terms are always erased, there's no need to worry about how it's safe to transform them.

Ericson2314:
> Your point (1): is this a change from your proposal? You say it's "not an implementation change"... is it a specification change? I just don't know how to interpret the text.

> Presumably, you're keeping "old" with "new" in the description for phi. Thanks to quantified constraints, this all is true already. So I think point (1) doesn't say much.

Right it's not an implementation change because quantified constrains already gives us this stuff. I like "specification change": just from deeming a larger portion of `Type` as the constraint proposition language than before, we have to recheck all theory.

Ericson2314:
> Newtypes and type families create axioms today. But they are not available for nthCo. Is this a typo? I'm not sure what you mean here.

I am just trying to describe a restriction we already have today, per https://repository.brynmawr.edu/cgi/viewcontent.cgi?article=1010&context=compsci_pubs . One can only use `nthCo` 

> This axiom still uses representational equality. But if there are no roles (as the proposal title suggests), what does representational equality mean?

> It's starting to make a bit more sense now. You're not proposing the abolishment of roles. That is, you still have two equality relations (nominal and representational) in your system. But some of the rules are different, and type constructors no longer are associated with a list of roles.

Ah right, roles have a broader meaning in the theory since they index the equality relations. In GHC I got used to thinking of roles as "just" the things associated with the type parameters, since there is no built-in surface syntax to range just over `~` and `Coercible`.

Ericson2314:
> Under Saturation, I presume you're not talking about the existing axiom that arises from newtypes -- just the new ones. OK.

I realized also that I implied that only data types get the new ones, when newtypes do too. Fixed it so both do.

> Under Saturation, I presume you're not talking about the existing axiom that arises from newtypes -- just the new ones. OK.

Yeah. Hopefully will be clearer when I push.

> "roughly": We'll need to be quite exacting here, in the end.

I just removed the "roughly". The new replacement rules naturally allow unsaturated `H`, and the rest stay exactly the same.

Ericson2314:
> `premise` basically says that, at least for datatypes, the axioms are the only way to produce representational equality. This may be troublesome, in that I don't see a restriction that only one axiom is associated with a type declaration. Was that meant to be in the mix here?

I added a uniqueness requirement. It's not hard to do @treeowl's thing whether there are multiple axioms with different scopes, but a most general one which is exclusively used by `coPremise`, but I'll not worry about that just yet.

-------

Overall thanks so much @goldfirere! Agreed that 18 and 22 are the big sticking points. For 22,  I just have to do the proof. For 18 I'll need to think more, but don't the existing decidable instance rules help, since there is no way to undecidable instances to affect anything? (note to self: call that out when I flesh out the surface syntax.) Do I basically just need to mash up the `~> *Q*` rules from your paper and the quantified constraints paper, and prove that everything still terminates?

Ericson2314:
@goldfirere Oh for 22 I think the best thing I can do prior to the proof is explain 23 better (see my current attempt, but let's discuss if it still doesn't make sense). This is a different intuition than your paper used to explain that `nthCo` restriction, so there might be some interesting to communicate.

goldfirere:
Thanks for the updates!

> GHC already lumps coercions under `Type`

How so? In the internal language, coercions are entirely separate from types. This may be a central source of confusion between us.

> can already be tricked

How so? Yes, you can use `unsafeCoerce` to build a coercion from any type to any type, but I can't get a function to be a coercion using `unsafeCoerce`.

> We are just deeming ...

Aha. Now I'm beginning to understand better. Quantified constraints, etc., exist in *Haskell* but do not exist at all in *Core*. So this would really be a big change. I had not understood this previously. (I think that's my oversight, given that this is in the Core section. Sorry.)

> admissible

You and I may be using different definitions for this word. I use "admissible" (when talking about type theory) to mean a property that is implied by other properties of the system. I think you mean to say that axioms that obey certain requirements would be acceptable.

> `CP :=` ...

I think you just want `:` there. You've given the type of the axiom, not its definition. (As an axiom, it doesn't really have a definition.)

> The following must be derivable

What is `t` in this rule? What are the free variables of `ft`, anyway? This doesn't quite seem right.

> lossless

I understand this much better now. Can users restrict the form of the axioms, though? If they can, then we're back in trouble.

> Do I basically just need to mash up the `~> *Q*` rules from your paper and the quantified constraints paper, and prove that everything still terminates?

Those rules are all about the constraint solver. Having a constraint solver that terminates sure is nice, but it's not the issue I'm worried about. I'm worried about type safety. If coercions aren't evaluated at runtime (necessary to keep them zero-cost), then we need to make sure anything that looks like evaluation in a coercion terminates. This is all about Section 4 of the paper, not about Section 5.

Ericson2314:
> How so? In the internal language, coercions are entirely separate from types. This may be a central source of confusion between us.

Oh, well I'm glad to be wrong. I was worried it was like `Constraint` and `Type` all over again! I got that idea from https://gitlab.haskell.org/ghc/ghc/wikis/commentary/compiler/fc#coercions-and-coercion-kinds, but that is out of date?

> You and I may be using different definitions for this word.

Right mine is wrong I'll go get another word.

> I think you just want `:` there.

I had it `C : ...` before, but i was using `C` as the type not the term, so I switched to `CP` for "constraint proposition" to be the type so `CP => (forall a* b*. ...)` makes sense.

I don't like this more logic than type-theory style, or extraneous higher ranked business (`CP => (forall a* b*. ...)`), though! There's probably a simpler and better way of doing it I'm missing.

> I understand this much better now.

Oh good!

> Can users restrict the form of the axioms, though? If they can, then we're back in trouble.

Right they wouldn't be able to for newtypes.

> Those rules are all about the constraint solver. 

Right, but don't have have the loose discipline today that if solver terminates then the solution terminates? I was hoping to steal some of that that. I'll look more again at section 4 though.

------

To settle all my core confusion, I think I need to really understand what is going on with the 
```haskell
 -- `-ddump-simpl` output

main :: IO ()
[GblId]
main
  = print
      @ (GList (IdentityT []) (Identity Int))
      (Main.$fShowGList
         @ (IdentityT [])
         @ (Identity Int)
         (Control.Monad.Trans.Identity.$fShowIdentityT
            @ [] @ (Identity Int) Data.Functor.Classes.$fShow1[] $dShow_r39o))
      (case asdf_rJJ @ [] @ (IdentityT []) @ Int @ (Identity Int) of
       { Dict' df_a2O6 ->
       case df_a2O6
              (GHC.Types.MkCoercible
                 @ *
                 @ [Int]
                 @ (IdentityT [] (Identity Int))
                 @~ (([Sym (Data.Functor.Identity.N:Identity[0]
                                <Int>_R)])_R ; Sym (Control.Monad.Trans.Identity.N:IdentityT[0]
                                                        <*>_N <[]>_R) <Identity Int>_N
                     :: [Int] ~R# IdentityT [] (Identity Int)))
       of
       { GHC.Types.MkCoercible v_B3 ->
       (Main.Mk
          @ []
          @ Int
          Main.Metadata
          (GHC.Types.: @ Int (GHC.Types.I# 1#) (GHC.Types.[] @ Int)))
       `cast` (v_B3
               :: GList [] Int ~R# GList (IdentityT []) (Identity Int))
       }
       })
```
from my example program https://github.com/ghc-proposals/ghc-proposals/pull/233#issuecomment-508841271.

I made it inline `asdf`
```haskell
main :: IO ()
[GblId]
main
  = print
      @ (GList (IdentityT []) (Identity Int))
      (Main.$fShowGList
         @ (IdentityT [])
         @ (Identity Int)
         (Control.Monad.Trans.Identity.$fShowIdentityT
            @ [] @ (Identity Int) Data.Functor.Classes.$fShow1[] $dShow_r2YQ))
      (case unsafeCoerce
              @ (Dict (Coercible Int Int))
              @ (Dict' [] (IdentityT []) Int (Identity Int))
              (Main.Dict
                 @ (Coercible Int Int)
                 (GHC.Types.MkCoercible
                    @ * @ Int @ Int @~ (<Int>_R :: Int ~R# Int)))
       of
       { Dict' df_a2El ->
       case df_a2El
              (GHC.Types.MkCoercible
                 @ *
                 @ [Int]
                 @ (IdentityT [] (Identity Int))
                 @~ (([Sym (Data.Functor.Identity.N:Identity[0]
                                <Int>_R)])_R ; Sym (Control.Monad.Trans.Identity.N:IdentityT[0]
                                                        <*>_N <[]>_R) <Identity Int>_N
                     :: [Int] ~R# IdentityT [] (Identity Int)))
       of
       { GHC.Types.MkCoercible v_B3 ->
       (Main.Mk
          @ []
          @ Int
          Main.Metadata
          (GHC.Types.: @ Int (GHC.Types.I# 1#) (GHC.Types.[] @ Int)))
       `cast` (v_B3
               :: GList [] Int ~R# GList (IdentityT []) (Identity Int))
       }
       })
```
But It's not going to inline the `unsafeCoerce`.  That mean's It cannot eta-reduce the `MkCoercible` either, so I can't trick GHC in making a true unboxed higher order coercion, just higher order wrapper. Drat!

goldfirere:
> but that is out of date?

Oof, yes. Very. I've added a disclaimer to the page.

> if solver terminates then the solution terminates

Not necessarily. Regardless, a major goal of having a typed core language is so that we can independently verify the constraint solver. If the soundness of core depends on a correct constraint solver, we'll have lost quite a lot.

Ericson2314:
> Not necessarily.

But almost?

>  Regardless, a major goal of having a typed core language  is so that we can independently verify the constraint solver. If the  soundness of core depends on a correct constraint solver, we'll have  lost quite a lot.

Oh I absolutely agree. I was thinking of merely taking inspiration from it. If the solver which terminates produces terms whose evaluations terminate, that great, but I must characterize the range of the sovler independent of the solver to use it in this proposal.

goldfirere:
> > Not necessarily.

> But almost?

The termination of the solver and the termination of the produced language aren't related, I think. Of course, we want both to terminate, and it's possible some misthinking in the solver leads both not to. But I can imagine a terminating solver producing a non-terminating core expression and a non-terminating solver that, when it terminates, is guaranteed to produce terminating core coercions.

Regardless, type safety must depend only on the termination of the produced core. For the sake of this issue, we can pretend that the solver is completely opaque. We just need to examine what it produces.

Ericson2314:
@goldfirere So giving it some more thought I have these two conjectures

1.  If all axioms follow the Paterson conditions, generalized for quantified constraints, then all inferred terms all follow the Patterson conditions. (i.e. all possible derivations are composed of paterson-condition-abiding subterms, and are themselves paternson-condition-abiding terms.)

2. The evaluation trace on such terms is redexes whose types only *grow*. This means that a well typed term must evaluate in a finite number of steps, because if it kept on evaluating, the additionally redexes and any final value have be too big a type.

Perhaps I'm just bad at looking, but I'm surprised I couldn't find anything like this in the literature I saw already. I would think that we do care that no Haskell program diverges evaluating dictionaries.

goldfirere:
I hate to be closed-mindedly clinging to one issue, but https://github.com/ghc-proposals/ghc-proposals/pull/248#issuecomment-512943645 still seems to be talking about the solver. Core has no quantified constraints, core coercions do not evaluate (I think these are the "terms" you discuss in your point (2)), and core does not have superclass axioms.

The question is: in your proposed new core language, given a well-typed term (not a coercion, but a normal evaluating term), is it guaranteed not to get stuck while evaluating? That is, can you prove progress and preservation? (Progress is the hard one for the changes you're considering.) Then, and only then, can we start to discuss the solver, which is a component of the compiler that produces core.

Ericson2314:
BTW I have realized via coincidental discussion on https://gitlab.haskell.org/ghc/ghc/issues/17295 that (local) quantified constraints involving equalities would need to become bidirectional for this to work.

Ericson2314:
I think I finally got two feet on the ground with this, and have talked a bit with @cgibbard until we reached an understanding. @treeowl we might want to bring in your nice examples from #276 too.

PR #249, opened 2019-07-10
Allow proposals to be written in markdown

isovector:
[Rendered](https://github.com/ghc-proposals/ghc-proposals/blob/87a2339d191027135537f90923f6d903c494c244/proposals/0000-allow-markdown.md)

### DIFF
--- a/proposals/0000-allow-markdown.md
+++ b/proposals/0000-allow-markdown.md
@@ -189,9 +189,3 @@ number of `=`s as characters in the line above it.
 Note that markdown doesn't require indentation of the code block, nor does it
 require blank lines before and after it.
 
-```diff
-@@ inline code
-- ``foo :: Int``
-+ `foo :: Int`
-```
-
--- a/proposals/0000-allow-markdown.md
+++ b/proposals/0000-allow-markdown.md
@@ -1,0 +1,195 @@
+---
+proposal-number: ""
+ticket-url: ""
+implemented: ""
+---
+
+# Allow Proposals to be Written in Markdown
+
+The current proposal process requires proposals to be written in ReST. This is a
+frustrating barrier to entry, and doesn't seem to pay its rent in real-life
+proposals.
+
+This will not deprecate writing ReST, should a proposal author wish to do so.
+
+
+## Motivation
+
+Markdown is a widely adopted document format that we can expect users to know;
+it's supported on Github, reddit, Slack, and to some degree, Facebook Messenger.
+Furthermore, it's is the default format for most non-WYSIWYG blogging software.
+Markdown makes the common cases easy.
+
+ReStructuredText, on the other hand, doesn't have nearly the market-share. The
+proposal template implicitly acknowledges this, in that it starts with a section
+users are expected to delete, explaining how to do basic things in ReST.
+
+A quick Github code search agrees with this. There are [180M Markdown
+documents](https://github.com/search?l=&q=filename%3A%2A.md+language%3AMarkdown&type=Code)
+on Github, compared with only [14M for
+ReST](https://github.com/search?l=&q=filename%3A%2A.rst+language%3AreStructuredText&type=Code).
+
+Furthermore, ReST is significantly more verbose than Markdown.  For a comparison
+of everyday constructs in both languages, see the appendix at the end of this
+document.
+
+The ghc-proposals README offers this in the way of explanation:
+
+> Note that proposals are written in ReStructuredText rather than Markdown for
+> its expressiveness and ease of integration into other GHC infrastructure.  See
+> the GHC Users Guide for a brief introduction to ReStructuredText.
+
+
+### Nobody actually uses ReSt's expressiveness
+
+To see if anyone actually uses this added expressiveness, I forked the
+`ghc-proposals` repository and grepped around in it for the constructs mentioned
+in the GHC Users Guide. The results are not favorable for ReST.
+
+We're told we can reference other documents in the Users Guide (a feature which
+obviously can't work when the proposal is *designed to be viewed on Github*.)
+This comes in the form `:ref:`, as well as package haddocks in the form
+`:<package>-ref:`.
+
+```shell
+$ grep -r ':\([A-Za-z-]*\)ref:' *
+index.rst:* :ref:`genindex`
+index.rst:* :ref:`modindex`
+index.rst:* :ref:`search`
+```
+
+We're told we can do math via `:math:`:
+
+```shell
+$ grep -r ':math:' * | wc -l
+0
+```
+
+Even if math were used, it isn't rendered in either the [GHC Users
+Guide](http://downloads.haskell.org/~ghc/latest/docs/html/users_guide/editing-guide.html#math),
+nor on Github.
+
+The other features described in the Users Guide have a similarly absent usage:
+
+```shell
+# index entries
+$ grep -r '^\.\. \+index::' * | wc -l
+0
+
+# citations
+$ grep -r ']_' * | wc -l
+0
+
+# admonitions
+$ grep -r '^\.\. \+\(attention\|error\|note\|caution\|hint\|tip\|danger\|important\|warning\)::' * | wc -l
+0
+
+# command-line options
+$ grep -r '^\.\. \+\(rts-flag\|ghc-flag\)::' * | wc -l
+0
+
+# ghci options
+$ grep -r '^\.\. \+\(rts-flag\|ghc-flag\)::' * | wc -l
+0
+```
+
+*Literally nobody* is using the extra expressiveness of ReST when writing
+proposals.
+
+
+### Tying into other GHC workflows
+
+As best I can tell, there is no integration of the proposal process with any
+existing GHC infrastructure. Even if there were, I doubt that the utility gained
+from it is more than the very real frustration in writing ReST.
+
+One potential argument to this proposal is that "since we're writing GHC's
+documentation in ReST, we should use it for the proposal process too." I don't
+think this holds water. If a proposal is passed, the proposal text is never
+going to be sufficient for direct inclusion into the Users Manual.
+
+Furthermore, there's a lazy-evaluation solution here. Rather than force everyone
+to write ReST *just in case*, let's only require someone to write ReST as part
+of a documentation MR on GHC itself. There's no reason to leak GHC
+implementation details into the proposal process.
+
+
+
+## Proposed Change Specification
+
+We will copy `0000-template.rst` to `0000-template.md`, and replace the ReST
+formatting with Markdown formatting.
+
+We will then remove the
+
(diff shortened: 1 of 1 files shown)


mpickering:
I consider it important that there is an official website which hosts accepted proposals so that there is somewhere canonical to cite accepted proposals. The advantage of the current process is that it is easy to use sphinx to create a reasonable looking site automatically.

https://ghc-proposals.readthedocs.io/en/latest/index.html

It would be possible to create a custom solution using hakyll which rendered the markdown documents so it's not an argument for or against the proposal but please carefully consider this when implementing the outcome.

proposals/0000-allow-markdown.md
```diff
@@ -0,0 +1,197 @@
...
++ foo = 5
++ ``
+```
+
+Note that markdown doesn't require indentation of the code block, nor does it
+require blank lines before and after it.
+
+```diff
+@@ inline code
+- ``foo :: Int``
```
neongreen: `@@ inline code` is duplicated.

parsonsmatt:
> It would be possible to create a custom solution using hakyll which rendered the markdown documents so it's not an argument for or against the proposal but please carefully consider this when implementing the outcome.

Given that GitHub renders Markdown quite well, we could just link to the documents on GitHub. I don't see this as a compelling point in favor of a format that is way less common than Markdown.

Ericson2314:
:-1: I found it easier to learn *docbook* to write Nix documentation, and RST for GHC proposals, than keep on indenting things carefully to get nested syntax in markdown to work. The latter is just black magic. https://spec.commonmark.org/0.29/ is highly suspicious because there is no formal grammar. I conclude markdown makes no sense.

Ericson2314:
I also savor the eventually glory of dropping an accepted, implemented proposal into a GHC documentation section. The GHC manual is written in...RST. ~~Translating would take all the fun away /s!~~ It is nice to have all GHC documentation written in the same way.

mpickering:
> > It would be possible to create a custom solution using hakyll which rendered the markdown documents so it's not an argument for or against the proposal but please carefully consider this when implementing the outcome.
> 
> Given that GitHub renders Markdown quite well, we could just link to the documents on GitHub. I don't see this as a compelling point in favor of a format that is way less common than Markdown.

GitHub also renders RST but I don't feel comfortable linking to the rendered document on github in a citation. 

Let me please reiterate that this isn't an argument about whether to use RST or markdown but if this proposal is accepted then please don't make it impossible to create a nicely rendered document/website from the proposals which people can feel comfortable citing.

parsonsmatt:
That's fair - fortunately [`pandoc`](https://avilpage.com/2014/11/pandoc-best-way-to-convert-markdown-to.html) handles Markdown to RST :smile:

Ericson2314:
@parsonsmatt Our ideomatic RST uses constructs with no markdown equivalent (e.g. definition lists).

nomeata:
> I also savor the eventually glory of dropping an accepted, implemented proposal into a GHC documentation section. The GHC manual is written in...RST. ~Translating would take all the fun away!~ It is nice to have all GHC documentation written in the same way.

This was the original motivation for using RST rather than Markdown, and I think it was and is the only motivation (readthedocs actually supports `.md`, IIRC from another project).

It turns out that proposal texts rarely enter the manual directly. Or do they?

AntC2:
AFAICT, the barrier to entry is deliberate. If you really want some change to GHC, learning ReST is the least of the effort. Toughen up!

What's more annoying is that some formatting you can do in markdown isn't available in ReST.

> It turns out that proposal texts rarely enter the manual directly.

I can think of one extension for which the proposal text went straight into the User Guide. The proposal text had not been updated to reflect actual development, so the User Guide was quite a bit misleading.

parsonsmatt:
> AFAICT, the barrier to entry is deliberate. If you really want some change to GHC, learning ReST is the least of the effort. Toughen up!

Yeah this attitude is why I'm not going to be contributing to GHC. I doubt I'm alone.

simonpj:
> AFAICT, the barrier to entry is deliberate

GHC has no deliberate barriers to entry, at least none that I know about., nor should it.

It is hard for a large community fo volunteers, all passionate, all trying to make Haskell better, but with different backgrounds and priorities, to work together seamlessly.   GHC's ecosystem is far from seamless: there are plenty of barriers.  But we shouls eeek to remove them, not erect them.   Of course, matters are seldom stratightforward -- in many cases there are costs and benefits to balance.  But we should start from a presumption of doing all we can to make it easy for people to contribute.

goldfirere:
I just want to second Simon's comment above. We don't try to make barriers. But we do sometimes choose to optimize for regular contributors over passers-by. For example, Phabricator had excellent code-review facilities, which worked great for us regulars, but I know posed a problem for new contributors. One of the benefits we sought in switching to GitLab was to lower this barrier.

In any case, https://github.com/ghc-proposals/ghc-proposals/pull/249#issuecomment-510254985 saddens me. I wish I knew what I/we could do to change that sentiment.

simonmar:
> GHC has no deliberate barriers to entry, at least none that I know about., nor should it.

I just wanted to second that - whatever barriers exist are generally the result of people trying to do the right thing, or solve a different problem, sometimes with unintended consequences. Nothing is set in stone and we should absolutely fix problems in the processes and tooling if we can understand what the problems are and how to fix them.

Let's not forget that we've already made deliberate (and costly, in terms of effort) changes specifically to lower the barrier to entry: the move to GitLab, which happened despite misgivings on the part of multiple regular GHC contributors, and this proposal process itself which was instituted in response to concerns from contributors. We're definitely open to making more changes if necessary.

mpickering:
We could also probably allow people to submit proposals in either markdown or RST and then convert the markdown into RST at the end. (RST being the richer format) Supporting both formats would probably end up being necessary anyway as all the proposals so far are written in RST.

nomeata:
As all discussions about syntax, this thread has the potential to be toxic and waste too many resources. Clearly, @isovector cares about it and put in thought to write down such an elaborate proposal. And we really do not want to deter people from contributing. 

So with my secretary powers, I declare that this change does not need a full proposal decision process, and that yes, we should just do that.

@isovector can you create a PR adding the other template (also update the template embedded in the link in README), and maybe point out next to the links in the README that RST has the advantage that it can be copied to the users guide, but otherwise either is fine.

takenobu-hs:
Everyone, please remember.

The GHC proposal process was started with the aim of lowering the barriers:

https://mail.haskell.org/pipermail/ghc-devs/2016-July/012468.html

gridaphobe:
I think @mpickering's idea to accept proposals in Markdown but convert them to RST upon merging is a very nice middle ground. Contributors get to write in a more familiar markup language, and we retain a single markup language that can easily be copied into the User's Guide on the backend.

isovector:
@nomeata Excellent, thanks! I'll spin up the PR.

@gridaphobe @mpickering I'm more than happy to write tooling to perform the markdown -> ReST transformation, but will hold off on doing so until we have a concrete plan for needing it. Give me a two weeks' lead time before such a website will go live, and I'll get you that software!

nomeata:
I don't think we need software. I (or whoever) can just run pandoc over the file before merging.

AntC2:
> @parsonsmatt Yeah this attitude is why I'm not going to be contributing to GHC.

This being the internet, I'm not sure what irony setting I should be reading that with. In case you didn't know, I am not one of the 'regular contributors' @goldfirere mentions, more of a regular observer/commenter.

My remark was not trying to turn anything toxic. And I'm genuinely surprised it's sparked such controversy. I thought I was merely stating the facts on the ground.

GHC is a large, hairy beast with 40 years of high intellectual input. Suggesting any change to it (that hasn't already been long churned over) is bound to be hard. Specifying the change in sufficient detail for an implementor, and considering all the interactions with the huge GHC infrastructure is an enormous challenge. Then a lesser effort is complying with the template -- which is a good discipline. And learning ReST is the least of the effort.

I presume the reason for the Github proposals process is that mixing up proposals with Trac tickets lead to too many half-baked 'bright ideas' and too many tickets that started off as (alleged) bugs then morphed to 'please GHC turn this into not-a-bug'. So Github is deliberately to raise the standard of proposals. I.e. a barrier to entry of half-baked ideas.

> @simonmar Let's not forget that we've already made deliberate (and costly, in terms of effort) changes specifically to lower the barrier to entry: the move to GitLab, which happened despite misgivings on the part of multiple regular GHC contributors, and this proposal process itself which was instituted in response to concerns from contributors. 

Sorry, GHC HQ, but that sounds like a dumb move. The GHC proposal/change/bugfix infrastructure should be optimised to minimise awkwardness for our most precious resource: the time/effort of people who are competent to change GHC. Certainly not for me. The move to Gitlab has significantly raised the barrier to entry for me, to the extent I've more or less given up on it -- but that's OK I didn't expect it to benefit me. If it hasn't helped 'regular GHC contributors', why do it?

I plain don't know what 'concerns from contributors' anybody was trying to address. I don't remember seeing any discussion around it. I rather took it that GHC HQ had its reasons, and that non-regular contributors were expected to sink or swim with it (mostly I've been sinking).

Perhaps I was overstating it to say "deliberate". More a case of: there were deliberate reasons for the change, and the concerns of non-regular contributors were rated of less importance. And that would be the right grounds.

@parsonsmatt, If you want a friendly/fun Haskell compiler to hack inside, have you considered Hugs? I'm enjoying working on it, and I've solved a problem that blocked the HList guys ~15 years ago. If you're going to say Hugs is hopelessly out of date, then consider why that is and what it means for the complexity that's gone into GHC in that 15 years.

AntC2:
> @AntC2 I plain don't know what 'concerns from contributors' anybody was trying to address. I don't remember seeing any discussion around it.

Thank you @takenobu-hs for posting that link to ghc-devs. I'm not a ghc developer, so I don't feel eligible to post there. And I see a fair amount of discussion before that went on on reddit. I find reddit incomprehensible and a gross waste of time for arriving at any sense. Time I don't have. My choice/preference; I'm explaining, not complaining.

simonmar:
@AntC2 thanks a lot for the honest feedback. We genuinely wanted to make things more transparent and have a level playing field for those who wanted to propose changes to GHC; to hear that this had the negative consequence of raising the barrier to entry for some is disappointing.

AntC2:
@simonmar, (let's not flog this to death but I can't let your comment be the last word here) I see no reason to be disappointed.

The Github proposal process _is_ more of a level playing field and more transparent. You've levelled the field by filling in the pot-holes, spreading crushed stone on the low-lying parts and churned-up mud. You've levelled it _up_.

Then naturally we (nearly) all must step up to the new level. ReST format might represent a bit of 'unnecessary complexity' compared to other markup/down formats; but it's only the height of verdant turf compared to the groundworks.

PR #250, opened 2019-07-11
Add a Markdown template

isovector:
Here are my suggested changes for enabling Markdown proposals. Any feedback is welcome!


README.rst
```diff
@@ -125,17 +125,12 @@ g. *Implementation Plan* (Optional): If accepted who will implement the change?
   This is also a good section to outline the changes to the implementation, or otherwise include insights that assume familiarity with the implementation.
 
 
-Note that proposals are written in `ReStructuredText
-<http://www.sphinx-doc.org/en/stable/rest.html>`_ rather than Markdown for its
-expressiveness and ease of integration into other GHC infrastructure. See the
-`GHC Users Guide
-<http://downloads.haskell.org/~ghc/latest/docs/html/users_guide/editing-guide.html>`_
-for a brief introduction to ReStructuredText.
+Proposals may be written in either `ReStructuredText <http://www.sphinx-doc.org/en/stable/rest.html>`_ or `Markdown <https://github.github.com/gfm/>`_. While the proposal process itself has no preference, keep in mind that the `GHC Users Guide <http://downloads.haskell.org/~ghc/latest/docs/html/users_guide/editing-guide.html>`_ uses ReStructuredText exclusively. Accepted proposals written in ReStructuredText thus have the slight benefit that they can be more easily included in the official GHC documentation.
```
aspiwack:
```suggestion
Proposals are written in either `ReStructuredText <http://www.sphinx-doc.org/en/stable/rest.html>`_ or `Markdown <https://github.github.com/gfm/>`_. While the proposal process itself has no preference, keep in mind that the `GHC Users Guide <http://downloads.haskell.org/~ghc/latest/docs/html/users_guide/editing-guide.html>`_ uses ReStructuredText exclusively. Accepted proposals written in ReStructuredText thus have the slight benefit that they can be more easily included in the official GHC documentation.
```

It sounds better doesn't it?

isovector:
> How about providing a one-liner to convert a markdown file into a RST file in the proposal format? 

I'd prefer we keep the author-facing bureaucracy to a minimum. Asking authors to do this means they no longer have a good source of truth for their proposal, and means they're going to need to spend an hour compiling pandoc. I think we should just deal with this ourselves when we _actually need a unified file format._

aspiwack:
> I'd prefer we keep the author-facing bureaucracy to a minimum.

I meant it for the benefit of @nomeata . Who will presumably have to do this translation.

nomeata:
@isovector note that the `README.rst` has an link that loads the template into the Github add new file UI, and there is a python one-liner in a comment in “How to start a new proposal” on how to generate that link. We now need two such links, and maybe an updated comment.

mpickering:
The metainfo header is missing fields for author and date.

isovector:
@mpickering fixed, thanks

@nomeata I tried changing the links in the README, but couldn't for the life of me figure out how to do it. I even broke out the official ReST reference and still couldn't work out why my link anchors weren't loading properly. The reversion of the change is https://github.com/ghc-proposals/ghc-proposals/pull/250/commits/bfeaaf552405885d7cb89afbce6ae746722bdd21 if you understand this language better than I do. I'm happy to merge this as-is, however; I suspect anyone who isn't familiar with  git probably doesn't have a strong preference between ReST and Markdown.

nomeata:
That should work now. Merge at will.

isovector:
Excellent, thanks so much!

PR #251, opened 2019-07-12
Add "Examples" section to template

goldfirere:
Many proposals are best illustrated with examples. With our experience reviewing proposals, I have observed two common suboptimal (to me) ways in how these examples are presented:

1. No (or too few) examples are provided. This almost inevitably leads to someone asking for examples to be provided.

2. The examples are mixed in with other text, often the specification. This often leads to someone asking for a more tightly-stated specification.

This PR introduces a new *Examples* section of the template, encouraging proposers to include examples of their new language construct. This is explicitly *not* part of the specification, which has been amended to specifically exclude examples. I believe introducing this new section will improve the readability of proposals and cause fewer (sometimes tiresome) requests for revision of proposals.

### DIFF
--- a/proposals/0000-template.md
+++ b/proposals/0000-template.md
@@ -12,50 +12,77 @@ the link, and delete this bold sentence.**
 
 # Proposal title
 
-Here you should write a short abstract motivating and briefly summarizing the proposed change.
+Here you should write a short abstract motivating and briefly summarizing the
+proposed change.
 
 
 ## Motivation
 
-Give a strong reason for why the community needs this change. Describe the use case as clearly as possible and give an example. Explain how the status quo is insufficient or not ideal.
+Give a strong reason for why the community needs this change. Describe the use
+case as clearly as possible and give an example. Explain how the status quo is
+insufficient or not ideal.
+
+A good Motivation section is often driven by examples and real-world scenarios.
 
 
 ## Proposed Change Specification
 
-Specify the change in precise, comprehensive yet concise language. Avoid words like should or could. Strive for a complete definition. Your specification may include,
+Specify the change in precise, comprehensive yet concise language. Avoid words
+like "should" or "could". Strive for a complete definition. Your specification
+may include,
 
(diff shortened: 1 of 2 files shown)


int-index:
Having an example-driven exposition definitely helps, but isn't it what the Motivation section is for? That is, perhaps it's better to make the Motivation section example-driven rather than have two separate sections?

After all, there must be some motivation behind each of the examples, no?

goldfirere:
Often the motivation section has more "real world" examples, and in many proposals, they are written *without* the new syntax. In some sense, my new Examples section is meant to complement the Motivation by writing examples in the new syntax -- perhaps contrived ones. If we settle on this, I can update the PR to make this explicit.

isovector:
I'm very pro on this!

int-index:
> my new Examples section is meant to complement the Motivation by writing examples in the new syntax -- perhaps contrived ones

So, if I understand correctly

* "Motivation" should contain "real-world" examples which serve to motivate the feature
* "Examples" should contain examples crafted to clarify the specification

Yes, then it's a great idea. Initially I (mis)interpreted the proposal such that all the examples would have to go to the "Examples" section, but I'd rather interleave examples with motivation.

goldfirere:
I have tweaked some of the commentary. Let me know what you think.

int-index:
> I have tweaked some of the commentary. Let me know what you think.

I should've read it more carefully in the first place, there was no ambiguity to start with. Indeed, it states that "This section illustrates the _specification_". But having a section called "Examples" made me immediately think "that's where all the examples go". The new wording makes it more clear that this is not necessarily the case, thanks.

nomeata:
Obviously a great idea, merge when you are happy with it, @goldfirere.

nomeata:
This only updates the RestructuredText template, not the markdown template, (added after my approval, but before the merge.)

I also, I guess I need to updates the links in the README, which embed the templates)

Guess I’ll quickly fix that.

nomeata:
Nevermind, both files are affected, I misread the diff.

PR #252, opened 2019-07-12
Add author and date metadata to all proposals

### LINKED ISSUE #247: Proposals should contain author information and date
The proposals metadata should contain information about the author and also the date the proposal was accepted. 

This information seems hard to work out from the rendered proposal page:

https://ghc-proposals.readthedocs.io/en/latest/proposals/0045-pointer-rep.html

and people often refer to I, which is not timeless when decoupled from author information.

isovector:
This was automatically generated by using `git blame` to determine who wrote the last line in the file (the author), and then finding the date the file was created (presumably the accepted date.)
It was generated via

```bash
#!/bin/bash

for FILE in proposals/*; do

AUTHOR=$(git blame $FILE | tail -n1 | cut -f2 -d'(' | sed 's/  \+/#/g' | cut -f1 -d'#' | cut -f1,2 -d' ' | cat <( echo -n ".. author:: ") -)
DATE=$(git log --format="format:%ai" $FILE | tail -n1 | cut -f'1' -d' ' | cat <( echo -n ".. date-accepted:: ") -)

sed -i "5i$AUTHOR" $FILE
sed -i "6i$DATE" $FILE

done
```

Closes #247

### DIFF
--- a/proposals/0032-type-data.rst
+++ b/proposals/0032-type-data.rst
@@ -2,7 +2,7 @@ Define Kinds Without Promotion
 ==============================
 
 .. proposal-number:: 32
-.. author:: Gabor Greif
+.. author:: Iavor Diatchki
 .. date-accepted:: 2018-09-12
 .. ticket-url::
 .. implemented::
--- a/proposals/0032-type-data.rst
+++ b/proposals/0032-type-data.rst
@@ -2,6 +2,8 @@ Define Kinds Without Promotion
 ==============================
 
 .. proposal-number:: 32
+.. author:: Gabor Greif
+.. date-accepted:: 2018-09-12
 .. ticket-url::
 .. implemented::
 .. header:: This proposal was `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/106>`_.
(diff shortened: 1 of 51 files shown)


int-index:
It seems that the `author` field contains the full name if available, and GitHub username otherwise. If this is policy, perhaps we should document it somewhere? And, perhaps, include both the full name and the GitHub username if available?

This additional meta-info is already useful as is, thanks for gathering it. I'm just wondering if it can be made more complete.

isovector:
@int-index, probably this shouldn't be policy. I'll just hand-edit the few github names that are present, if I can do it without doxxing anybody. Thanks for the catch!

nomeata:
Nice approach!

Yeah, let's not make policy, and let's not dig up names that people did not already volunteer on their Github profile.

proposals/0032-type-data.rst
```diff
@@ -2,6 +2,8 @@ Define Kinds Without Promotion
 ==============================
 
 .. proposal-number:: 32
+.. author:: Gabor Greif
```
nomeata: Shouldn’t this be Iavor?

isovector: Looks like yes. Blame is really messy on that file! I'll fix it. Nice catch!

PR #253, opened 2019-07-19
Clean treatment of contexts in "Function Result Type Signatures"

Ericson2314:
The proposed change had a context in the grammar, but this was never
commented on in the design description, and has funny interactions.

This removes it from the base extension (which is lacking a name! but
hopefully that is fixed), but allows it (via inaction) with Rank2Types.

CC @int-index. I hope this is the most acceptable to you of the 3 quibbles I had :).

[Rich diff](https://github.com/ghc-proposals/ghc-proposals/pull/253/files#diff-2ac9f0c3a207e16c928ba1a54a2ab2f7) warning this link might go stale, but I'm not sure how to do better.

### DIFF
--- a/proposals/0051-function-result-sigs.rst
+++ b/proposals/0051-function-result-sigs.rst
@@ -188,7 +188,7 @@ an optional type annotation::
            | pat varop pat
            | ( funlhs' ) apat {apat}
 
-  funlhs -> funlhs' :: [context =>] type
+  funlhs -> funlhs' [:: type]
 
 This results in an ambiguity with pattern bindings, which is resolved in favor
 of function bindings.
@@ -242,8 +242,36 @@ Alternatives
   this is neither consistent nor terribly useful.
 
 * We could detect CUSKs as we do in types to enable polymorphic recursion, but
-  this makes little sense as we are in the proccess of their deprecation.
+  this makes little sense as we are in the process of their deprecation.
 
+* We could allow contexts in the return type.
+  An earlier version of this proposal had::
+
+    funlhs -> funlhs' [:: [context =>] type]
+
+  But the usage of this can be confusing.
+  For example::
+
+    f (x :: a) :: Eq a => a
+
+  looks like it it might have type::
+
+    f :: a -> Eq a => a
+
+  Which is gratuitously higher-ranked. Base Haskell 2010 never reorders
+  arguments currently, so just moving the ``Eq a`` to the front seems overly
+  magical.
+
+  However, with ``-XRank2Types``, contexts are allowed in types and so this
+  naturally becomes legal again. In GHCi today we have::
+
+    Prelude> :set -XRank2Types
+    Prelude> f :: Int -> Eq a => a; f = undefined
+    Prelude> :t f
+    f :: Eq a => Int -> a
+
+  i.e. the "out of order" ``Eq a`` is allowed and moved to the front.
+  By analogy, the return signature ``Eq a`` would also be so moved.
 
 Unresolved Questions
 --------------------


int-index:
This looks reasonable to me.

proposals/0051-function-result-sigs.rst
```diff
@@ -242,8 +242,36 @@ Alternatives
...
+  But the usage of this can be confusing.
+  For example::
+
+    f (x :: a) :: Eq a => a
+
+  looks like it it might have type::
+
+    f :: a -> Eq a => a
+
+  Which is gratuitously higher-ranked. Base Haskell 2010 never reorders
```
goldfirere: Nitpick: This is not a higher-rank type. It is non-prenex, but there is no polytype to the left of an arrow here.

proposals/0051-function-result-sigs.rst
```diff
@@ -242,8 +242,36 @@ Alternatives
...
+  arguments currently, so just moving the ``Eq a`` to the front seems overly
```
goldfirere: Haskell 2010 is happy to reorder arguments like this: `f _ = abs` gets the type `f :: Num a => p -> a -> a`. Note the `Num a` moved.

proposals/0051-function-result-sigs.rst
```diff
@@ -242,8 +242,36 @@ Alternatives
...
+  magical.
+
+  However, with ``-XRank2Types``, contexts are allowed in types and so this
+  naturally becomes legal again. In GHCi today we have::
+
+    Prelude> :set -XRank2Types
+    Prelude> f :: Int -> Eq a => a; f = undefined
+    Prelude> :t f
```
goldfirere: This is a quirk of `:t`, not of inference. Use `:t +v` to get the real type, which will have `Eq a` in its declared place.

proposals/0051-function-result-sigs.rst
```diff
@@ -242,8 +242,36 @@ Alternatives
...
+
+  looks like it it might have type::
+
+    f :: a -> Eq a => a
+
+  Which is gratuitously higher-ranked. Base Haskell 2010 never reorders
+  arguments currently, so just moving the ``Eq a`` to the front seems overly
+  magical.
+
+  However, with ``-XRank2Types``, contexts are allowed in types and so this
```
goldfirere: I don't think we need the `RankNTypes` extension to enable contexts. It should just work.

Ericson2314: Without the `-XRank2Types`
```
Prelude> f :: Int -> Eq a => a; f = undefined

<interactive>:1:6: error:
    • Illegal qualified type: Eq a => a
      Perhaps you intended to use RankNTypes or Rank2Types
    • In the type signature: f :: Int -> Eq a => a
```

goldfirere: That's true. The extension is slightly misnamed. Perhaps `FlexibleQuantification` would have been better. But I still think this new behavior can be enabled without the extension: the type you write at the end of the LHS will be prenex, and the inferred type of the function will be prenex (because of reshuffling the order), and so no one gets hurt.

Ericson2314: I...wish to be conservative here. It just occurred to me, say in some dependent future we have 
```haskell
f (a :: k) :: Constraint k a => f a = ...
```
This should infer to
```haskell
f :: forall k. foreach (a :: k) -> Constraint k a => f a
```
We can always be more flexible about rearranging things later, but once the cat is out of the bag we cannot put it back.

goldfirere: The rearranging is all up to type inference -- it's not something that even needs to be specified here. Instead, we just say that the result has a certain type, and then type inference will either check a type signature against that or infers a type; either way, it all just works.

Ericson2314: @goldfirere I suppose with the normal "it just needs to unify" LHS treatment that is true.

So it sounds like you don't think anything should be changed in the proposal then? Just the wording cleaned up? We always allow contexts, but they and the return type signatures are optional?

phadej:
Thinking about this proposal as a whole,

```haskell
double (x :: a) :: Num b => b = x + x
```

would that be accepted or not?

Is it like writing

```haskell
double :: (Num b, _) => a -> b
double x = x + x
```

which GHC currently rejects with 

```
    Couldn't match ‘b’ with ‘a’
      both bound by the partial type signature:
        double :: (Num b, _) => a -> b
  |
2 | double x = x + x
  | ^
```

but that's arguable.

On the other hand, that translation IMHO makes sense to describe what should or shouldn't pass the type-checker.

Ericson2314:
Right for all these sorts of things, I am just more comfortable binding new non-rigid variables only in negative position, on analogy with other pattern matching.

int-index:
@phadej It would definitely be accepted, unifying `a` and `b`. Your translation to a top-level function signature is not valid, just like it would not be valid for pattern signatures:

```
f (x :: a) (y :: b) = x==y

g :: _ => a -> b -> Bool
g x y = x==y
-- Rejected:
--    Couldn't match ‘a’ with ‘b’
--      both bound by the partial type signature: g :: _ => a -> b -> Bool
```

Ericson2314:
https://github.com/ghc-proposals/ghc-proposals/pull/254/files has the more radical changes.

goldfirere:
This amendment draws attention to function-result contexts. Simon and I weren't sure exactly how these would work, so I drew up some typing rules. See https://gitlab.haskell.org/rae/haskell for the source, and a link from the README there to a PDF containing rules (bottom of page 2) for result signatures. (I know that most people here don't read these rules -- they are not required reading, but the rules are a wonderfully precise way to state what is going on.) The PDF has a lack of commentary, as it's mostly for my benefit, but I can build it out if there is demand.

Ericson2314:
Oh that's good to have! So it's looking like nothing in this PR is really needed.

goldfirere:
> So it's looking like nothing in this PR is really needed. Is that your conclusion too?

Yes, I think so. I will close. But it helped spark good conversation. Thanks!

PR #254, opened 2019-07-23
Scale back "Function Result Type Signatures" to keep pattern sigs as-is

Ericson2314:
Keep pattern vs function binds matching the spec, rather than changing the division as the original proposal proposed to expose a hitherto hidden implementation detail in GHC.

I don't know why pattern variables couldn't just bind variables on the right side of the `::` just as they do on the left. But, I grant this is a big change, because the scope of the binding would be large. The conservative choice to keep all options  open is just to keep today's restriction of no binding variables in pattern binding signaturess.

[Rendered new version](https://github.com/Ericson2314/ghc-proposals/blob/fn-res-sig-bind/proposals/0228-function-result-sigs.rst)


int-index:
Today, GHC treats a variable binding as a nullary function binding:

```
x = 5   -- accepted as FunBind
```

Why is that? Here are some comments from GHC sources:

```
Reason 1: Special case for type inference: see 'TcBinds.tcMonoBinds'.
Reason 2: Instance decls can only have FunBinds, which is convenient.
          If you change this, you'll need to change e.g. rnMethodBinds
```

When a signature is added to a variable binding, is interpreted as a pattern binding, as function bindings cannot have signatures:

```
x :: Int = 5  -- accepted as PatBind, cannot be a FunBind due to the signature!
```

However, with #228 accepted, we do have signatures in function bindings, and it is a hard sell that `x = 5` is a FunBind while `x :: Int = 5` is a PatBind.

That was my motivation behind reclassifying `x :: [a] -> [a] = reverse` as FunBind.

If you are proposing to drop this part of the proposal, we should also reclassify `x = 5` as a PatBind. And I think it would be the right thing to do. Then we could drop the strictness flag from FunBind, which frankly does not belong there.

However, we need to come up with solutions to the `Reason 1` and `Reason 2` from that GHC comment. Both of them sound more like engineering issues than something fundamental, so I think we can pull this off?

Ericson2314:
Glad you like this change! I started plugging away at making GHC parse them as pattern binds in https://gitlab.haskell.org/ghc/ghc/merge_requests/1474 .

goldfirere:
Ah -- so MR 1474 is an implementation of this proposal. In the future, if you have a proposal related to an MR, please make this prominently clear in the MR. It's fantastic to offer an implementation, but reviewing that MR absent this discussion was a bit confusing. Thanks!

We indeed need to handle the issues Vlad cites.

Ericson2314:
Well I want #1474 to stand on it's own; it just cuts the baby on the other axis. I think this proposal could be implemented with or with it it too, though it definitely makes more sense with it.

proposals/0051-function-result-sigs.rst
```diff
@@ -244,6 +225,68 @@ Alternatives
 * We could detect CUSKs as we do in types to enable polymorphic recursion, but
   this makes little sense as we are in the proccess of their deprecation.
 
+* An earlier version of the propsal changed the boundary between function binds
+  and pattern binds via not requring an arugment::
```
int-index: s/arugment/argument/

goldfirere:
Now I'm very confused. Are you advocating `x = 5` to be a `FunBind` or a `PatBind`? It would be very helpful to lay out what you are proposing, why, what goes right, and what goes wrong, all in one place. It doesn't have to be a diff against a proposal or some such, but I'm honestly quite lost here. Thanks!

int-index:
The standard says:

* `x = 5` is a `PatBind`
* `x :: Int = 5` is a `PatBind`

The GHC implementation does not match this specification at the moment:

* `x = 5` is a `FunBind`
* `x :: Int = 5` is a `PatBind`

Under the "function result type signatures", the specification is amended such that:

* `x = 5` is a `FunBind`
* `x :: Int = 5` is a `FunBind`

The argument that @Ericson2314 makes is that

* We should fix GHC to respect the specification regardless of any proposals, so `x = 5` must be a `PatBind` according to the standard

* We should revert the part of the "function result type signatures" that reclassifies these constructs as `FunBind`

These points are related but could be evaluated independently.

Although I believe considering them independently only adds to the confusion. My requirements are such that

* The implementation should match the specification.
* `x = 5` and `x :: Int = 5` should be of the same variety, either both `PatBind` or both `FunBind`.

goldfirere:
Very helpful, Vlad -- thanks!

Why does this matter? I think it matters because of the following argument:

Consider

```
let x :: a = 5 in ...
```

What is the scope of `a`? If we have a `FunBind`, the scope of `a` would be `5`. If we have a `PatBind` (and `PatBind`s supported scoped type variables), then the scope of `a` would be the `...` (the body of the `let`).

Odd here is that certain constructs permit *only* pattern bindings (such as `do` notation); a bare variable in those constructs is always a pattern binding.

Also odd is that, with result signatures, `x _ :: a = 5` is perfectly sensible. But if `x :: a = 5` is a pattern binding, then that would be rejected. So this is a user-facing distinction. I believe that the implementation in GHC today (where `x = 5` is a `FunBind`), is done for internal expediency and the belief that no one would ever know. Suddenly, that's no longer the case.

It's probably easiest just to stick with the standard, and have `x = 5` be a `PatBind`. This reduces the set of accepted programs (`x :: a = 5` is rejected).

We then have to address points (1) and (2) of https://github.com/ghc-proposals/ghc-proposals/pull/254#issuecomment-514415455. As Vlad suggests, these are small engineering-y issues and can probably be resolved. That said, if it takes contortions to resolve them, I might change my mind here.

Does that get us going again? Or maybe we were going this whole time, but at least now I understand where we're going and why. :)

simonpj:
> It's probably easiest just to stick with the standard, and have x = 5 be a PatBind. 

Yes, I agree with this.

> As Vlad suggests, these are small engineering-y issues and can probably be resolved

I'm pretty sure they can.   Lets's call a pattern binding that has a simple variable (perhpas with a bang) on the LHS a "simple pattern binding".   The places we'd need to adjust are

* `rnMethodBinds`, since an instance method can now have a (simple) pattern binding.
* `tcMonoBinds`: see `Note [Single function non-recursive binding special-case]`.  The speciao case shoiuld apply to simple pattern bindings too.
* `decideGeneralisationPlan`, esp the defn of `restricted`.

> this reduces the set of accepted programs (x :: a = 5 is rejected).

Well `x :: Int = 4` would be accepted as a `PatBind` with pattern `(x :: Int)`.  The resaon it's rejected is becaues we don't allow pattern bindings to bring type variables into scope.

Everyone happy with this?

Ericson2314:
Really glad we all agree! Thanks @int-index especially for wording it much better than I did!

Hope this isn't off topic but in the interest of end-to-end designing, along with clarifying what is in my PR:

With the agreed-upon route taken, I would then propose that pattern bindings be allowed to bind variables. The scope as @goldfirere points out is the crucial destinguishing feature, and I think a useful one. Basically I hope we end up in a state where "all metavariables can be written down".

goldfirere:
I think that the last time someone tried to allow type-variable bindings in `let`, Simon's [brain exploded](https://stackoverflow.com/questions/23531566/odd-ghc-error-message-my-brain-just-exploded). Perhaps these would be sensible in local, non-recursive `let`s. But it's not obvious to me that this works with recursion. And I also wonder what happens at the top-level: do we want to export such bound type variables? Maybe we do, but that goes far, far beyond what we're looking at here. So my thought here is to sort out the local troubles first, and then perhaps return, with a brand new proposal, about type-variable bindings in `let`.

simonpj:
Yes currently GHC simply **rejects any pattern binding that introduces fresh type variables**.  E.g.
```
f x = let (y::a, z::b) = (x,x) in (z,y)
```
If `a` and `b` were brought into scope here, they would _have_ to have the same scope as `y` and `z`.  There are many complications there, and GHC now rejects such programs utterly.    You may want otherewise, but no one has reported it as a problem for them, and we should not hold up the train for this.

TL;DR: this a GHC proposal but has ended up as a proposal ofr user-invisible refactoring.  Is that right?  There'll be no user-visible change?

Ericson2314:
Yeah I meant propose as a separate proposal, not as part of this one.

> TL;DR: this a GHC proposal but has ended up as a proposal of a user-invisible refactoring. Is that right? There'll be no user-visible change?

That's only because the accepted proposal is not yet implemented. This proposal *does* scale back that proposal in a user-visible way, because
```haskell
(f :: a) = ...
```
*is* currently accepted with `f` and `a` having differing scopes in the pre-this-PR accepted version of the proposal.

The PR doesn't mention the non-user-visible refactor to GHC we've discussed changing `x :: a =` to a pattern bind in accordance with the spec, but I think that is the correct choice. Future readers shouldn't be confused reading about what will become historical obsolete implementation details that were at no point user visible, just to learn about a language extension.

int-index:
@Ericson2314 Could you please extend this patch to drop the "Unresolved Questions" section? Its premise is stated as follows:

```
x :: [a] -> [a] = reverse    -- accepted as FunBind
(x :: [a] -> [a]) = reverse  -- rejected as PatBind
```

However, with your adjustment, both of these are `PatBind`, so there are no unresolved questions.

Ericson2314:
@int-index Good eye. I still fill like it's a bit silly that `(f a) = ...` doesn't parse, but that's entirely separate from this proposal with this PR.

simonpj:
I'm confused about this thread.  It's in the GHC proposals repo, but it doesn't appear to be a proposal -- at least I can't see one.   If there is a propsoal here, would be possible to write it up in the usual format?

Ericson2314:
Did you see the diff https://github.com/ghc-proposals/ghc-proposals/pull/254/files ? I could have made a new proposal but seeing that this one is just newly accepted that seemed like overkill.

simonpj:
A proposal itself is a diff on the Haskell language.  This is a diff on a diff.  My head has exploded.

Let's see: could you possibly work with Vlad on producing a single revised proposal, superseding the previous one, that embodies all the change you want to see?  That way I think it'll be easier to engage a broad constituency of people in reviewing.

int-index:
> A proposal itself is a diff on the Haskell language. This is a diff on a diff. My head has exploded.

Here's a page that shows the original proposal with @Ericson2314 changes applied: https://github.com/Ericson2314/ghc-proposals/blob/fn-res-sig-bind/proposals/0051-function-result-sigs.rst

> Let's see: could you possibly work with Vlad on producing a single revised proposal, superseding the previous one, that embodies all the change you want to see?

@Ericson2314 Are there other changes you'd like to see or does #254 cover it?

Ericson2314:
@int-index yes since we all agreed to close the other one it's just this.

proposals/0051-function-result-sigs.rst
```diff
@@ -244,39 +225,75 @@ Alternatives
 * We could detect CUSKs as we do in types to enable polymorphic recursion, but
   this makes little sense as we are in the proccess of their deprecation.
 
+* An earlier version of the propsal changed the boundary between function binds
```
int-index: s/propsal/proposal/

goldfirere:
What's the status here? Should this be sent to the committee?

Ericson2314:
@goldfirere Well, I think the original implicit plan was I was going to finish https://gitlab.haskell.org/ghc/ghc/merge_requests/1474, to make sure GHC's implementation details can match the spec. But, I got sucked back to the cross/build system stuff that is my main task with the 8.10 branch looming. So I suppose we can send to committee now, especially if we assume !1474 or something like it can be debugged with out hiccups and just a bit more time and effort.

goldfirere:
> I suppose we can send to committee now,

Is this a request to send to the committee? If so, please notify Joachim (username: nomeata). Thanks!

proposals/0051-function-result-sigs.rst
Ericson2314:
```suggestion
* An earlier version of the proposal changed the boundary between function binds
```

proposals/0051-function-result-sigs.rst
Ericson2314:
```suggestion
  and pattern binds via not requring an argument::
```

Ericson2314:
OK I cleaned up some sloppiness on my end, and I believe per the above @int-index is OK with these changes (correct me if I'm wrong!), so yes let's send it to the committee! @nomeata

int-index:
Yes, I'm fine with these changes, but I'd like @Ericson2314 to be added as a proposal co-author, as at this point not all of the wording is mine.

nomeata:
> I'd like @Ericson2314 to be added as a proposal co-author,

Quite appropriate. Just add him to the metadata on top :-)

simonpj:
I'm the shepherd for this proposal, which is a modification
to an earlier, accepted about result signatures for function
definition.  I recommend  acceptance. Some reminders:

* [The existing (accepted) proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0228-function-result-sigs.rst), rendered.
* [The new verison](https://github.com/Ericson2314/ghc-proposals/blob/fn-res-sig-bind/proposals/0228-function-result-sigs.rst), after this modification is applied.
* [The diff](https://github.com/ghc-proposals/ghc-proposals/pull/254/files)

The original proposal added result signatures, so you can write
```
f x :: (a->a) = \y -> (y :: a)
```
which says that the call `f x` returns a result of type `a -> a`; and moreover brings `a` into scope in just the same way as in a pattern signature.

The question is: what happens if the number of arguments is not two, or one, but zero:
```
f :: a -> a = blah
```
Is that a (nullary) function binding (with a return signature), or a pattern binding (with the pattern `(f :: a -> a)`?  It matters; for example, in a pattern binding a pattern is not allowed to bind new type variables.

The original proposal treated the above definition as a nullary function binding with a return signature. But that leads to problems as [the (revised) proposal](https://github.com/Ericson2314/ghc-proposals/blob/fn-res-sig-bind/proposals/0228-function-result-sigs.rst) discusses in the (bizarrely numbered) Section 232 Alternatives.

Moreover, minor variants like
```
(g :: a -> a)  = blah
!(h :: a -> a) = blah
```
look a lot more like pattern bindings, but it'd be odd for the rules to change suddenly.

So the current proposal draws back a bit, and regards
```
f = e
f :: type = e
```
as pattern bindings (as in the report) and therefore entirely out of scope for
the "function return signature" proposal.  In short, the new function return signature
propoals now applies only to function definitions with at least one argument.

To me this seems like an easy accept.

simonpj:
I will note that, in the original proposal, I'm still anxious about what we should make of
```
f x :: Num a => a -> a = blah
```
or
```
g y :: forall b. C a b => a ->b -> b = blah
```
That may come out of [Richard's evolving static semantics work](https://gitlab.haskell.org/rae/haskell).  But it doesn't affect accepting this proposal.

Ericson2314:
I didn't see a tracking issue, so I made one in https://gitlab.haskell.org/ghc/ghc/issues/18203

PR #257, opened 2019-08-05
Unlifted Data through `Strict`

sgraf812:
[Rendered](https://github.com/sgraf812/ghc-proposals/blob/unlifted-strict/proposals/0000-unlifted-strict.rst)

### DIFF
--- a/proposals/0000-unlifted-strict.rst
+++ b/proposals/0000-unlifted-strict.rst
@@ -66,8 +66,8 @@ the caller:
 ::
 
  tsum :: Strict (Tree Int) -> Int
- tsum Force Leaf           = 0
- tsum Force (Branch l x r) = tsum l + x + tsum r
+ tsum (Force Leaf)           = 0
+ tsum (Force (Branch l x r)) = tsum l + x + tsum r
 
 Note that this particular example would be less of an issue if we had
 strictness analysis and worker/wrapper transformation work for sum types: The
--- a/proposals/0000-unlifted-strict.rst
+++ b/proposals/0000-unlifted-strict.rst
@@ -46,7 +46,7 @@ properly tagged.
 For ``Nat`` above, having ``x :: Strict Nat`` is enough to know that the
 inhabitant really only can be a natural number.
 
-For another example, consider strict lists of ``Int``s:
+For another example, consider strict lists of ``Int``:
 
 ::
 
@@ -57,10 +57,11 @@ For another example, consider strict lists of ``Int``s:
  isum INil         = 0
  isum (ICons x xs) = x + isum xs
 
-Since ``isum undefined`` is a possible call site of ``isum``, codegen can't omit
-a tag check on ``isum``s parameter, although in all recursive calls the tail
-should be properly tagged. Giving ``isum`` the following type, reflecting its
-strictness, gets rid of any tag checks, offloading the burden to the caller:
+Since ``isum undefined`` is a possible call site of ``isum``, codegen can't
+omit a tag check on the parameter of ``isum``, although in all recursive calls
+the tail should be properly tagged. Giving ``isum`` the following type,
+reflecting its strictness, gets rid of any tag checks, offloading the burden to
+the caller:
 
 ::
 
--- a/proposals/0000-unlifted-strict.rst
+++ b/proposals/0000-unlifted-strict.rst
@@ -1,0 +1,244 @@
+Unlifted Data through ``Strict``
+================================
+
+.. author:: Sebastian Graf
+.. date-accepted:: Leave blank. This will be filled in when the proposal is accepted.
+.. proposal-number:: Leave blank. This will be filled in when the proposal is
+                     accepted.
+.. ticket-url:: Leave blank. This will eventually be filled with the
+                ticket URL which will track the progress of the
+                implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+This proposal is about introducing to `ghc-prim` the unlifted data type
+
+::
+
+ data Strict :: Type -> TYPE 'UnliftedRep where
+   Force :: !a -> Strict a
+
+that deprives itself and its argument of ⊥.
+
+Motivation
+----------
+To quote the `unlifted data types wiki page <https://gitlab.haskell.org/ghc/ghc/wikis/unlifted-data-types#proposal-b4-levity-polymorphic-functions>`_:
+
+Bob Harper `has written <https://existentialtype.wordpress.com/2011/04/24/the-real-point-of-laziness/>`_:
+
+    Haskell suffers from a paucity of types.  It is not possible in Haskell to
+    define the type of natural numbers, nor the type of lists of natural numbers
+    (or lists of anything else), nor any other inductive type!
+
+The reason, of course, is that whenever you write ``data Nat = Z | S !Nat``, you
+define a type of strict natural numbers, AS WELL AS bottom. Ensuring that an
+``x :: Nat`` is never bottom requires all use-sites of this type to do strict
+pattern matching / force the value appropriately. It would be nice if there was
+some type-directed mechanism which specified that a value ``x`` was always
+evaluated. This would give benefits, e.g. for code generation, where we can
+assume that a pointer never points to a thunk or indirection and is thus
+properly tagged.
+
+For ``Nat`` above, having ``x :: Strict Nat`` is enough to know that the
+inhabitant really only can be a natural number.
+
+For another example, consider strict lists of ``Int``s:
+
+::
+
+ data IntList = ICons !Int !IntList
+              | INil
+
+ isum :: IntList -> Int
+ isum INil         = 0
+ isum (ICons x xs) = x + isum xs
+
+Since ``isum undefined`` is a possible call site of ``isum``, codegen can't omit
+a tag check on ``isum``s parameter, although in all recursive calls the tail
+should be properly tagged. Giving ``isum`` the following type, reflecting its
+strictness, gets rid of any tag checks, offloading the burden to the caller:
+
+::
+
+ isum :: Strict IntList -> Int
+ isum (Force INil)         = 0
+ isum (Force (ICons x xs)) = x + isum xs
+
+Note that this particular example would be a non-issue if we had strictness
+analysis and worker/wrapper transformation work for sum types: The argument
+would turn into an unboxed sum with arguably even better performance
+characteristics.
+The point is that we can do this for *any* strictly used argument of lifted
+kind! There's an opportunity for worker/wrapper here.
+
+Proposed Change Specification
+-----------------------------
+Add a new data type ``Strict :: Type -> TYPE 'UnliftedRep`` to ``GHC.Exts``
+with a single constructor ``Force :: !a -> Strict a``.
+
+As ``Strict`` is just one more unlifted data type, its semantics follow from 
+the semantics of unlifted data types. In particular:
+
+* When wrapped in a field (e.g. ``data T = MkT (Strict ())``), the semantics are
+  identical to a field with a bang pattern (``data T = MkT !()``), modulo
+  packing and unpacking of the ``Force`` constructor.
+
+* In argument position (``f (Force a)``) and in the right-hand side of let
+  bindings (``let x = Force e1 in e2``), the argument ``Force a`` is evaluated
+  before beta reduction and the right-hand side ``Force e1`` is evaluated
+  before the body. Since ``Force`` is strict in its field, this forces
+  evaluation of the wrapped lifted expressions ``a`` and ``e1``. 
+
+Examples
+--------
+Every unlifted data type will become syntactic sugar over a combination
+of unlifted newtypes and ``Strict``, like
+
+::
+
+ data Ptr a = Ptr Addr#
+ newtype Ptr# :: Type -> TYPE 'UnliftedRep where
+   Ptr# :: Strict (Ptr a) -> Ptr# a
+
+ newtype UPair :: Type -> Type -> TYPE 'UnliftedRep where
+   UPair :: Strict (a, b) -> UPair a b
+
+We can even recover ad-hoc forms of `unboxed strict tuples <https://gitlab.haskell.org/ghc/ghc/issues/17001>`_:
+
+::
+
+ (#! Either Int Bool, Char#, ByteArray# !#)
+ ==>
+ (# Strict (Either Int Bool), Char#, ByteArray# #)
+
+In fact, ``Strict`` is somewhat similar to the unit unboxed strict tuple. It
+crucially is a specialisation to lifted types, though, meaning it still has a
+boxed representation. This is important for later endeavours into levity
+polymorphism (rather than the current boxity polymorphism) over lifted and
+unlifted types.
+
+Another example again concerns the worker/wrapper transformation. Consider
+
+::
+
+ data SPair a b = SPair !a !b
+ 
+ foo :: Int -> SPair Int Int
+ foo x
+   | even x
+   = SPair (x+1)  x
+   | otherwise
+   = case foo (x-1) of
+       SPair a b -> SPair (a+1) (b+1)
+  data SPair a b = SPair !a !b
+
+CPR analysis will discover that ``foo`` has the constructed product result
+property. Hence WW will turn this function into (ignoring strictness and
+inlining for the sake of simplicity)
+
+::
+
+ foo :: Int -> SPair Int Int
+ foo x = case $wfoo x of (# a, b #) -> SPair a b
+
+ $wfoo :: Int -> (# Int, Int #)
+ $wfoo x
+   | even x
+   = (# (x + 1), x #)
+   | otherwise
+   = case $wfoo (x-1) of
+       (# a, b #) -> (# a+1, b+1 #)
+
+Compared to the original definition of ``foo``, ``$wfoo`` lost knowledge of the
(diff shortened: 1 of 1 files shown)


goldfirere:
Thanks for writing this up! I'm looking forward to the design debate around the different alternatives. I was tempted to say "Let's just have unlifted data types", but you observe that your approach here separates the surface-level language concerns from code generation. I agree, and this will simplify matters. Perhaps, in the end, your `Strict` becomes user-definable, but that's backward-compatible with this proposal. I also wonder about strict unboxed tuples, but I don't have a strong enough sense of what would be best here -- hence my anticipation of the design debate.

As a smaller note: I don't understand your `tsum` example. Syntactically, I think you're missing some parentheses in pattern-matching. But also I don't see how strictness avoids the tag check, given that there are two tags. I *do* see how strictness can eliminate a tag check, but only when there is one tag.

sgraf812:
Thanks Richard, great to have you on board :)

> Perhaps, in the end, your `Strict` becomes user-definable

Yes, I hope so. But on the other hand, I'm not sure if we can eliminate the indirection in the general (that is, almost-newtype) case. Also inference of `Force`, which could be the goal of a future proposal, would be hard to achieve/justify for all almost-newtypes. Anyway, removing the indirection is unimportant from a semantic perspective, so actually pretending that `Strict` was user-defined as in the abstract would be a reasonable model to work with.

> I don't understand your tsum example

Yes, syntax was off, indeed I missed parentheses.

Semantically, my whole proposal hinges on  [#16970](https://gitlab.haskell.org/ghc/ghc/issues/16970), specifically the fact that all strict constructor fields (`Force`s in particular) are tagged*. Hence both the `Force` thing and its wrapped value should be properly tagged. I cleared up that dependence by mentioning it in the motivation.

* Note that we also silently assume so far that `Force` things (the dreadful box) are always tagged, which might not be true for the similar reasons that strict constructor fields aren't always tagged, I guess. Although there's no `STATIC_IND` for unlifted things, since they can't be defined at the top-level. Whatever.

goldfirere:
I'm still not following the example, I'm afraid. I think part of my confusion is that the strict example looks like it needs to check two levels of tag: both `Force` and e.g. `Leaf`. What tag check are you presuming to eliminate?

int-index:
I don't follow the `tsum` example either. It says:

```
data Tree a = Branch !(Tree a) !a !(Tree a)
            | Leaf

tsum :: Strict (Tree Int) -> Int
tsum (Force Leaf)           = 0
tsum (Force (Branch l x r)) = tsum l + x + tsum r
```

but this does not typecheck. `l, r :: Tree Int` while `tsum` takes `Strict (Tree Int)` as input, so you need to wrap them:

```
tsum :: Strict (Tree Int) -> Int
tsum (Force Leaf)           = 0
tsum (Force (Branch l x r)) = tsum (Force l) + x + tsum (Force r)
```

Unless there's an implicit wrapping mechanism implied here?

AndreasPK:
> I also wonder about strict unboxed tuples, but I don't have a strong enough sense of what would be best here -- hence my anticipation of the design debate.

Having suggested strict unboxed tuples I think this is the next best, if not an equal approach.
This proposal could be considered slightly less powerful as Force is restricted to lifted types. But in practice it's a power difference that doesn't matter much.

sgraf812:
> What tag check are you presuming to eliminate?

Your confusion probably stems from the fact that by "tag check" I don't mean the actual `switch`ing involved, but rather the [zero check for the pointer tag](https://gitlab.haskell.org/ghc/ghc/issues/16820), where in the taken branch we would first have to enter the object. This check is unnecessary when we can statically prove that the tag could never be zero. Of course we still have to dispatch on the particular tag, we just don't need to enter the thing anymore to know it. Did my changes clarify this?

> so you need to wrap them

Of course, makes sense, fixed. Thanks!

goldfirere:
We're getting closer! The key question: what is the "zero check"? The proposal links to a ticket, but the first page of the ticket does not define a "zero check", and I'm not at all familiar with this term. Thanks!

sgraf812:
Added an explanation with a small C-- example to the motivation.

isovector:
At a high level, this looks good to me! But I am painfully aware about how little I know in the space of unlifted reps --- offhand can anyone point me in the direction of good reading for understanding the nitty gritty in this proposal?

goldfirere:
Thanks very much @sgraf812! I now fully understand that example. I will try to paraphrase, both to check my understanding and lightly translate to folks unfamiliar with code generation (like me):

The blob of memory that represents a lifted datatype might be in head normal form (HNF), meaning that (at least) the outermost constructor is fully evaluated. In this case, the blob contains a *tag* that corresponds to the choice of constructor. If the datatype is not in HNF, it is a thunk. This thunk will have 0s in its lower 3 bits, as it is a pointer to the code that represents evaluating the thunk. (We thus have an invariant that no tag can be divisible by 8, as 0 in the lowest three bits means "thunk".) When case-matching on a datatype, even before we look at which constructor we have, we must check whether we have any constructor at all -- in other words, we must evaluate before scrutinizing. The check for whether an object is evaluated or not is called the "zero tag check". It can be skipped if we know, statically, that an object is evaluated in HNF.

Is this summary correct? The rest of my post assumes it is.

A few questions:
* Is it possible to do the zero-tag check and scrutinize at the same time? In C, this would look something like

```c
void match(byte* obj)
{
  switch(*obj)
  {
    case 0: (*obj)();   // evaluate thunk
       match(obj); // try again
       break;
    case 1: // first constructor
    case 2: // second constructor
    ...
  }
}
```

The zero tag check is on even footing as all the other ones. Hopefully a backend can produce code that implements the check in terms of, say, local jumps and needn't do a linear search to find the matching number.

If this is done or can be done, then eliminating the zero tag check for a datatype with multiple constructor buys very little.

* Even if we can't do the zero tag check and scrutinize in the same step, it seems possible to take something like `tsum` and make a separate function `tsum_strict` that is used for the recursive occurrences. The `tsum_strict` code could certainly skip the zero tag check. Do we do this now? Would it be hard to do? If we already do this (or could easily), then the change to `tsum` saves only the initial check -- still perhaps worthwhile, but less convincingly so. (Maybe this is the worker/wrapper transformation you mention, but it doesn't require inlining unboxed sums, so I think it's different.)

-------------

I should say that I like the idea of `Strict` even if `tsum` isn't a great motivating example (it might be -- I'm still learning about this and really don't know what I'm talking about). But I do want to make sure we're clear on what this will and won't do for us.

goldfirere:
> offhand can anyone point me in the direction of good reading for understanding the nitty gritty in this proposal?

The [Levity Polymorphism](http://repository.brynmawr.edu/cgi/viewcontent.cgi?article=1065&context=compsci_pubs) paper offers Section 2, which should be an accessible introduction. There may be other, better resources, but this one first comes to mind.

sgraf812:
Thanks, Richard!

> Is this summary correct?

Yes, although I should mention that the blob could already have been evaluated, but the pointer not tagged yet (i.e. lower bits 0). That's what happens when a shared thunk is evaluated in one code path and then scrutinised again in another: Then we have to enter the memoised value once more to find the constructor tag.

> do the zero-tag check and scrutinize at the same time

I suppose you could. I gave this a try to see what code GHC generates for an enum of size 7. This is the result:

```
           if (R1 & 7 != 0) goto c1dK; else goto c1dL;
       c1dL: // global
           call (I64[R1])(R1) returns to c1dK, args: 8, res: 8, upd: 8;
       c1dK: // global
           switch [1 .. 7] (R1 & 7) {
               case 1 : goto c1dO;
               case 2 : goto c1dP;
               case 3 : goto c1dQ;
               case 4 : goto c1dR;
               case 5 : goto c1dS;
               case 6 : goto c1dT;
               case 7 : goto c1dU;
           }
```

So this does one extra comparison and conditional jump for the zero case, and in that case dispatches over the same jump table. We could probably get by without those additional two instructions and a backwards jump instead (punishing the presumably cold path)! I'm not sure why we aren't, but I'm not terribly invested in the back end. I'd expect LLVM to optimise this redundancy away.

> eliminating the zero tag check for a datatype with multiple constructor buys very little

Well, there's a lot of dead code and unnecessary stack movement (see the "low-level" example) which would go away completely. This probably matters for hot loops. @AndreasPK did some experiments on `containers` and found improvements of 10% on `Map.member`ship tests.

> make a separate function tsum_strict that is used for the recursive occurrences

Yes, I've thought about that a little. Let's pick up that thought once more.

It sounds a lot like constructor specialisation, only that we specialise for strictness/taggedness now, the `Force` constructor, so to speak. Or even better: We could do this as part of worker/wrapper, i.e. when we found out that a function is strict in a parameter that can't be unboxed, emit a `seq` on that parameter in the wrapper. That wrapper is then inlined at call sites and hopefully cancels away with existing statically visible evaluation, like in the recursive case of `tsum`. It's a little hard to do this without types guiding the transformation (after this proposal we could give the worker `Strict` parameters), though: How do we communicate to the worker that this particular argument is properly tagged? That's what we need `Strict` for.

> Maybe this is the worker/wrapper transformation you mention

Yes, that's exactly the kind of transformation I had in mind! But we need some means of communicating the analysis info. Worse: If we had the information available and did the specialisation, and  through some program transformation called `tsum_strict` with unevaluated arguments (think: regression in the simplifier), this wouldn't break in CoreLint at compile time but result in a segfault instead.

AndreasPK:
> do the zero-tag check and scrutinize at the same time:

Jump tables are generally worse if we only have few possible targets, so your code snippet is slightly misleading in terms of performance.
The current code we generate is in fact a compiled representation of such a switch, but optimized for the unevaluated case. There are reasons for doing this but they are beyond this proposal. See [here](https://gitlab.haskell.org/ghc/ghc/issues/8905) for some discussion.

But either way the issue of zero-tag checks exists even for types with a single Constructor.

We might, for example, want to take out the second field of a Constructor like a tuple. If we know we are given a pointer to the constructor itself the Cmm code will be:

```C
           R1 = P64[R1 + 7]; //Offset of field - tag
           call (I64[R1])(R1) //"return R1"
```

However currently we have to check if the argument is evaluated first so we get instead:

```C
//Set up the stack, check the tag, evaluate argument if needed.
       c1dr: // global
           if ((Sp + -8) < SpLim) (likely: False) goto c1dv; else goto c1dw;
       c1dv: // global
           R2 = R2;
           R1 = Foo.foo_closure;
           call (stg_gc_fun)(R2, R1) args: 8, res: 0, upd: 8;
       c1dw: // global
           I64[Sp - 8] = c1do;
           R1 = R2;
           Sp = Sp - 8;
           if (R1 & 7 != 0) goto c1do; else goto c1dp;
       c1dp: // global
           call (I64[R1])(R1) returns to c1do, args: 8, res: 8, upd: 8;
       c1do: // global
           I64[Sp] = c1du;
//Finally, get the field and return it.
           R1 = P64[R1 + 7];
           call (I64[R1])(R1) //"return R1"
```

The cost here is not simply the branch, but we also spill to the stack, adjust the stack pointer, potentially spill live variables. If other code does not require spills to the stack it will also incur another branch for a stack check.

There is no point in pretending that this will be a huge improvement for a typical haskell program. But there is a small number of code patterns which do suffer from this overhead and they will benefit greatly.

> Even if we can't do the zero tag check and scrutinize in the same step, it seems possible to take something like tsum and make a separate function tsum_strict that is used for the recursive occurrences. The tsum_strict code could certainly skip the zero tag check. Do we do this now?

In short no. 

@sgraf812 answered most of this already but to be explicit there is currently no way to define tsum_strict this way. Users can't define it this way.
The compiler won't try to transform it into something like this either, primarily because GHC does internally have no notion of "boxed but guaranteed unlifted" as Force would provide/require.

This proposal would allow users to write such functions by hand, and make transformations towards this goal in the compiler easier.

goldfirere:
Thanks @sgraf812 and @AndreasPK! This all makes much more sense to me now. The key bit for me was about the fact that there would be no way to communicate to `tsum_strict` not to evaluate its argument. Somehow I thought there was a mechanism for that, but clearly there isn't. And indeed `Strict` provides precisely that mechanism.

My support for this proposal has grown stronger. Many thanks for taking the time to educate me!

AndreasPK:
In 3.3 some trailing code seems to have snuck into the code block.

simonpj:
I'm interested in discussing the tradeoff between

1. A single unlifted `Force` type as you propose
2. Allowing users to declare new unlifted types, as [proposed here](https://gitlab.haskell.org/ghc/ghc/wikis/unlifted-data-types#proposal-1-allow-data-types-to-be-declared-as-unlifted).

Advantages of (1)

* One new built-in primitive type
* You can make an existing type strict by wrapping it in `Force`

Disadvantages of (1)

* All this pattern-matching on `Force` and use of `Force` as a constructor (which you can see in `tsum`) adds a lot of clutter to the source code.  Indeed, in the defn of `Ptr` in 3.1, the `Addr#` is wrapped in no fewer than _three_ data constructors: `Ptr`, `Force` and `Ptr#`!

* All this pattern-matching on `Force` and use of `Force` as a constructor may add runtime cost.  In particular, a data type like `Strict` alwasya allocates a box, and that is expensive -- which is particularly unacceptable in the performance critical code for which this feature is intended.

  Alternatively, perhaps we can arrange that `Force` is implemented by nothing at all, a bit like `coerce`??   But that really would be entirely new.   And this "coercion" would be between types of kind `TYPE LiftedRep` and `TYPE UnliftedRep` respectively, and generally we don't let you coerce between types with different runtime-representations.

Advantages of (2)

* All the additional pattern matching and construction goes away.  For example, we can just say
  ```
  data unlifted Ptr a = Ptr Addr#
  ```
  and we are done.

* `Strict` is definiable as a special case.  (But it will be boxed unless we do more work.)

Disadvantages of (2) (maybe)

* The proposal says "This would have to anticipate all possible interactions with existing ways to introduce (generalised) algebraic data types, from surface language issues regarding syntax to code generation issues".   I don't see the problem here.  We'd just need the new user-defined unlifted data type to have kind `TYPE UnliftedRep`.   _We already have type constructors of this kind_ so nothing new is needed to deal with them.  I think it might "just work".

To me (2) seems rather attractive.

sgraf812:
I started implementing a `Strict#` primtype with `mkForce#`/`getForce#` pseudoops and noticed that it's quite hard to come up with all the builtin rewrite rules, when all we are basically saying is that `mkForce#` and `getForce#` are `DataCon`s. Implementing `Strict` as a data constructor is probably the proper solution here, at which point we could just give the user the additional bit of syntax (well, omission of semantic check, rather) to declare them themselves. I'm a little wary about unforeseen interactions, but you seem pretty convinced that the hard bits should already be hammered out. So why not? I'll prepare a counter proposal for that.

I actually consider removal of the box essential to making `Force` a useful thing, by the way. We could apply this optimisation to any 'newtype' that turns lifted into unlifted pointers.

> Alternatively, perhaps we can arrange that Force is implemented by nothing at all, a bit like coerce?? But that really would be entirely new. And this "coercion" would be between types of kind TYPE LiftedRep and TYPE UnliftedRep respectively, and generally we don't let you coerce between types with different runtime-representations.

Yes, exactly. I'm not sure this would be covered by the current `Coercible` mechanism, because they are supposed to be zero-cost (which a `coerce` from lifted to unlifted rep isn't, it's a `seq`). But maybe we could introduce a super-class of `Coercible` for that (`CostlyCoercible` 🎉), or introduce an entirely new mechanism.

sgraf812:
`Force` is still useful as a library solution, by the way: Strict fields with pointer representation should desugar to a `DataCon` wrapper that forces strict field arguments and a worker that actually has `Force`-wrapped arguments. Example

```
data SPair a b = T !a !b
-- introduces T :: Strict a -> Strict b -> SPair a b
-- with the following wrapper
$WPair :: a -> b -> TPair a b
$WPair x y = T (Force a) (Force b)
```

goldfirere:
I think we can implement `Strict` without a box. We'll just need primitives

```
forceCoerce :: forall (a :: TYPE LiftedRep). a -> Strict a
   -- just forces. No box.

unpackStrict :: forall (a :: TYPE LiftedRep). Strict a -> a
  -- does nothing at all at runtime
```

Here, we take advantage of the fact that (as I understand it) the set of possible representations for a value of type of kind `TYPE UnliftedRep` is a subset of the equivalent for `TYPE LiftedRep`: that is, the former just doesn't have thunks. So as long as we know something isn't a thunk, we can just change its kind. Unpacking the lifted value out of a `Strict` is then a no-op.

sgraf812:
> We'll just need primitives

This is basically the approach I tried implementing in https://github.com/ghc-proposals/ghc-proposals/pull/257#issuecomment-523073017. I found it unsatisfying, because I have the fear that the primitives will trip up the simplifier and other optimisations. For example, I *think* we'd need the rewrite rules `forall a. forceCoerce (unpackStrict a) = a` and ``forall v. let a = forceCoerce v in f (unpackStrict a) = v `seq` f v``. The reasoning would all come for free if `Strict` was just a `DataCon`.

I think eliminating the box is purely a CodeGen optimisation, so we probably shouldn't introduce new pseudoops that basically are just data constructors.

treeowl:
Since I made a [similar suggestion](https://mail.haskell.org/pipermail/ghc-devs/2018-February/015356.html), I'm in favor. I do think it would be nice to guarantee that there's no box. That seems even more important than guaranteeing no zero-tag check.

treeowl:
One gripe: every other primitive unlifted thing I know of has a hash at the end of its name. Shouldn't this too?

sgraf812:
Here is the 'Unlifted Data' counter proposal: https://github.com/ghc-proposals/ghc-proposals/pull/265

code5hot:
I have a need to be able to trivially define an unlifted type for any datatype with exactly one constructor, we do this for datatypes with one constructor *and* one field using `newtype` but the constraint should just be "exactly one constructor" and otherwise be semantically similar to `newtype`. I started a proposal at #268 but it's not fully refined yet.

treeowl:
Speaking of one-way coercions: if we ever get those, we'll also want

```haskell
oneWayCoerce :: a -> (# a #)
```

which I've mentioned before. One-way coercions need to be careful about covariance and contravariance.

sgraf812:
Closing this as superseded by #265 which already has a propotype implementation

PR #258, opened 2019-08-06
TopLevelKindSignatures -> StandaloneKindSignatures

monoidal:
This is an addendum to the accepted proposal 34, top-level kind signatures.

I propose to rename the extension to `StandaloneKindSignatures`. The reasons are:

(1) Some day we might introduce local types (types in `where` clauses). In this case, it would still make sense to add a kind signature, but the name "top-level" would no longer be suitable.
(2) This name is consistent with `StandaloneDeriving`. A standalone deriving declaration allows users to separate a datatype declaration from its derived classes; a standalone kind signature allows users to separate a datatype declaration from a kind signature.

The idea comes from discussion at gitlab: https://gitlab.haskell.org/ghc/ghc/merge_requests/1438#note_215740

Since this change concerns only renaming, I am not creating a new document.

The extension has not landed in GHC yet, so there's no worry about backwards compatibility.


goldfirere:
I'm in strong support.

isovector:
I like this.

RyanGlScott:
I'll admit that I'll be somewhat sad to part with the acronym TLKSs, since saying it out loud ("tolks") gave it a very middle-earth flair. I'm not quite sure how to pronounce SKSs yet. ("sicks"? "skizz"?)

In all seriousness though, I'm in support.

Ericson2314:
"suckses"!

goldfirere:
I believe someone (@monoidal?) suggested somewhere SAKS. I propose, though, that we should just use SAK, to avoid plurality confusion.

monoidal:
@nomeata The proposal is ready for committee review.

nomeata:
Thanks, assigning this to @bitemyapp

int-index:
It's been a week since this was assigned to a shepherd.

I have a merge request that uses "top-level kind signatures" all throughout: in the User's Guide, in the Notes and other comments, in function names, in data declaration names, in error messages, language flags, etc.

Do I rename it or not? I'd like to strike it off my todo list.

int-index:
In fact,

* it's been a month since this renaming was suggested in the [merge request discussion](https://gitlab.haskell.org/ghc/ghc/merge_requests/1438#note_215750)
* three weeks since this proposal was written on the `ghc-proposals` platform
* a week since it was assigned to a shepherd

and the current status is not even "under committee consideration".

I'm a bit annoyed at this pace of reaching consensus for a simple name change. There isn't even much to debate here: let's just have a vote by the committee and move on.

int-index:
This was first suggested a month ago in the [discussion of the implementation](https://gitlab.haskell.org/ghc/ghc/merge_requests/1438#note_215750). In addition to changing the extension name, I will need to update the User's Guide, numerous code comments, names of functions and data types, error messages, and so on. If we get a decision on this sooner, it will help me move the implementation forward. Thanks!

simonpj:
Chris, you are the shepherd.  Do you have a recommendation?

nomeata:
After some internal discussions, we decided that this doesn't need the full process, it is more or less a simple uncontroversial editorial fix. Weilding my process administrative super powers, I’ll merge it as such. Sorry for not paying better attention initially.

PR #259, opened 2019-08-06
No kind signatures on associated types

goldfirere:
This amendment to the accepted proposal on standalone kind signatures removes the ability to put a kind signature on associated types, as we could not choose a good design for this feature and the feature does not seem all that pressing.

* [Original, unchanged proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0036-kind-signatures.rst)
* Discussion of original proposal at #54.
* [Rendered version of updated proposal](https://github.com/goldfirere/ghc-proposals/blob/no-assoc-sak/proposals/0054-kind-signatures.rst)
* [Rendered version of diff](https://github.com/goldfirere/ghc-proposals/compare/8a2f26408decd4be7799179213b3d3416509eb18...no-assoc-sak?short_path=bf40b0d#diff-bf40b0d802d1ac546bd04c73dfd23dd6)
* Previous amendment to move associated type signatures to top-level at #227.

### DIFF
--- a/proposals/0054-kind-signatures.rst
+++ b/proposals/0054-kind-signatures.rst
@@ -70,14 +70,14 @@ Proposed Change Specification
      type <name_1> , ... , <name_n> :: <kind>
 
    (where ``n >= 1``) It is distinguishable from a type synonym by the lack of an ``=`` on the line. A
-   type-level declaration with a top-level kind signature may use polymorphic recursion;
+   type-level declaration with a standalone kind signature may use polymorphic recursion;
    one without is considered to have an inferred kind and may not use polymorphic recursion. One signature can apply to many names, just like a type signature.
 
    The kind given is checked against the declaration of the type. All kind generalization
    is done before ever examining the full declaration, just like how GHC treats type
    signatures.
 
-   Associated types may not be given top-level kind signatures. (See "Costs and Drawbacks" for discussion.)
+   Associated types may not be given standalone kind signatures. (See "Costs and Drawbacks" for discussion.)
 
    Unlike type signatures, the type variables brought into scope in a type-level kind
    signature do *not* scope over the type definition.
@@ -118,6 +118,10 @@ to give, e.g., ``data ProxyVis k (a :: k)`` a kind signature.
 
 Costs and Drawbacks
 -------------------
+
+Implementation Cost
+~~~~~~~~~~~~~~~~~~~
+
 Implementation should be rather straightforward, as this is a new syntactic construct.
 
 Parsing may be slightly complicated by the similarity to a type synonym, but I doubt this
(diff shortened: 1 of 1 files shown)


goldfirere:
This is a small tweak, debated at some length during the implementation.

/remind me in two weeks to submit to the committee.

isovector:
To clarify, is this removal only in the context of `-XTopLevelKindSignatures`? Or is the intent to break existing code that puts kind signatures on assoc tyfams?

proposals/0036-kind-signatures.rst
```diff
@@ -77,7 +77,7 @@ Proposed Change Specification
    is done before ever examining the full declaration, just like how GHC treats type
    signatures.
 
-   Associated types may be given kind signatures within their classes.
+   Associated types may not be given kind signatures.
```
RyanGlScott: It would be helpful to give a forward reference to the Costs and Drawbacks section here, since someone not in the know will likely find this sentence quite surprising without any context to explain why this choice was made.

goldfirere:
I've clarified the proposal to answer the above (good) questions.

RyanGlScott:
Thanks! We should make sure to replace the use of "top-level" with "standalone" in the prose if #258 is accepted.

simonpj:
I support this proposal.

goldfirere:
Time to submit, @nomeata. Thanks!

yav:
This change seems reasonable, given the lack of preferred design choice.

Unrelated:   shouldn't we allow a single top-level signature to specify the kinds for multiple type things (to match how normal signature works)?  I mean, make the rule like this:

```Haskell
type <name> (, <name>)* :: <kind>
```
This would allow us to write things like:
```haskell
type T, S :: Type -> Type
```

goldfirere:
Good point about the multiple names, @yav. I've just updated the proposal directly. (I thought of making a PR and such, but that just seemed like a waste of everyone's time.)

Ericson2314:
> ...waste of everyone's time

Except for yours with git :D. N.B. You have a merge conflict now.

goldfirere:
What's the status here? @simonpj ?

simonpj:
> What's the status here? @simonpj ?

On 14 Sept [I said to the committee](https://mail.haskell.org/pipermail/ghc-steering-committee/2019-September/001213.html) "I would love to hear from more of you.  But this particular proposal is a small delta, and if I don't hear any objections I'll declare it accepted on Weds 18 Sept."

But then I dropped the ball.  No one responded.  So I declare this proposal accepted.

@nomeata, can you execute?  Thanks

TheMatten:
One question - would signatures in `instance`s in style of `InstanceSigs` introduce any problems?

goldfirere:
@TheMatten I suppose those would be ok. If we saw `type F :: xyz`, we'd know it's a signature.

TheMatten:
Is support for signatures in instances worth considering then?

goldfirere:
If you're so inclined, you could write it up. I don't think it's quite worth it, for several reasons:

* `-XInstanceSigs` allows (at least) two nice things: (1) scoped type variables via a top-level `forall` and (2) the ability to make the method definition *more* polymorphic than necessary. Neither of these nice things carry over to type families. (1), it has been decided, does not apply to type-level declarations, and (2) the type-level subtyping relation (that is, the "more-polymorphic-than" relation) is weaker than the term-level one. (Though that may be changing: #287) Thus (2) is less applicable.
* Without constraints, etc., it seems easier just to put kind signatures inline than to do the same for type signatures.
* My hope is that support for proper dependent types will eclipse type families in the not-too-distant future. At that point, features such as the one you're supposing will become redundant.

PR #261, opened 2019-08-09
Classify instances by deriving strategy / declaration (:classify)

Icelandjack:


Icelandjack:
IDEs could allow collapsing but there should be a more principled way of getting an instance's metadata to allow more sophisticated filtering

```haskell
[-] newtype: ..
[x] via:
  [-] Co Point: ..
  [-] Compose Identity (Co Point): ..
[-] stock: ..
[x] standalone:
  instance Representable f => Representable (Point f)
  type instance Rep (Point f) = Rep f
```

monoidal:
I would not use "standalone" to avoid confusion with standalone deriving. Maybe "manual"?

If we extended `:info`, we could just use existing syntax:

```
Prelude> :i Int
data T
deriving newtype instance Eq T -- Defined in Foo:4:22
instance Ord T -- Defined in Foo:6:10
```

Issue #162, opened 2018-08-20
Add naive core interpreter so TH and GHCi work even with ABI changes


Ericson2314:
external interpreters and `--target` help with the cross case, but that still leaves the ABI compatibility problem. The end result for GHC development itself is the same: TH and GHCi are not availible in certain bootstrapping scenarios, with restricts the way GHC and core libs must be developed in myriad obscure and highly new-contributor-hostile ways.

Also, it seems like a dedicated core interpeter would be a good thing to have in general as a "living quasi-specification", including giving something authoritative for people to read, oracle for tests, etc.

goldfirere:
Apologies if I'm being dense, but I really do not understand what this is proposing. What is `--target`? What is the "ABI compatibility problem"? How do `--target` and external intepreters help? What don't they do? How would a Core interpreter help? What's "highly hostile" about today's setup?

angerman:
@goldfirere, you need to look at this through the cross compilation lenses.

- `-target` refers to ghc being eventually able to produce code for multiple targets from the same compiler by providing the relevant `-target` flag to ghc. Similar to what `clang` allows.

- `-fexternal-interpreter` runs a secondary process (called `iserv`) as the interpreter. This allows a stage1 GHC to evaluate TH splices via the external interpreter process.

TH in its current (unrestricted) form is a major hurdle to get cross compilation working with GHC.

Ericson2314:
@angerman tells me about the source plugins new to GHC 8.6. These are very cool, and I think better embody the current implementation strategy. A plugin by nature accesses far too vast an API to be worth the manual virtualization and mashalling the interpreter method uses. Using the old GHC (worst case) to make code ABI-compatible with the new GHC is the necessary price to access all the bells and whistles of the compiler library.

simonmar:
Fun fact: GHCi was originally implemented as a Core interpreter. But that implementation never made it as far as a release, because we couldn't get it to fully work. The sticking point was unboxed types (and would probably be unboxed tuples these days, too). It would be possible if the interpreter didn't have to interact directly with compiled code, but for GHCi it does.  Without that constraint the interpreter could choose its own representation for unboxed values (i.e. it could use boxed ones). 

But I presume the idea here is to interpret everything, including the base package. For that you would need some way to persist the Core when compiling modules, and load it into the interpreter.

Ericson2314:
[...]

PR #262, opened 2019-08-14
White-box interface files

michaelpj:
[Rendered](https://github.com/michaelpj/ghc-proposals/blob/white-box-interface-files/proposals/0000-white-box-interface-files.rst)

### DIFF
--- a/proposals/0000-white-box-interface-files.rst
+++ b/proposals/0000-white-box-interface-files.rst
@@ -1,0 +1,454 @@
+White-box interface files
+=========================
+
+.. author:: Michael Peyton Jones
+.. date-accepted:: Leave blank. This will be filled in when the proposal is accepted.
+.. proposal-number:: Leave blank. This will be filled in when the proposal is
+                     accepted.
+.. ticket-url:: Leave blank. This will eventually be filled with the
+                ticket URL which will track the progress of the
+                implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/262>`_.
+.. sectnum::
+.. contents::
+
+GHC is a tremendous resource for language designers to work with. GHC is very
+friendly, and allows users to reuse many parts of it, especially with the advent
+of compiler plugins. But support is not so good for users who want *white-box*
+access to compiled modules.
+
+This proposal suggests a new kind of interface file to remedy that issue. It is
+a resurrection and evolution of Edward Yang’s `"fat interface files" proposal
+<https://gitlab.haskell.org/ghc/ghc/issues/10871>`_.
+
+Motivation
+----------
+
+GHC Haskell programs are typically put together by compiling together some
+number of modules. These can be local source modules, or they can be separately
+compiled.
+
+Separately compiled modules look rather different to source modules. We have two
+files:
+
+1. An object file, which provides compiled code for the target platform that can
+   be linked against.
+2. An interface file, which provides the information that the typechecker needs
+   to typecheck against the compiled module.
+
+   - Interface files can contain other things, in particular we will have more
+     to say about unfoldings later.
+
+This is all well and good so long as those are the only two things that you want
+to do with a separately compiled module: link against it and typecheck against
+it.
+
+However, sometimes we want to do more with our dependencies than this. A
+particularly common desire is to see the *Core* for a compiled module. This
+allows us to do *white-box compilation*.
+
+White-box compilation
+^^^^^^^^^^^^^^^^^^^^^
+
+A white-box compiler relies on seeing more information about compiled modules
+than is typically provided. [1]_
+
+.. [1] Some people talk about “whole program” compilation or optimization. I
+   haven’t used this term because I find it somewhat ambiguous: if you do
+   link-time optimization when
+   linking your object files, are you now using a whole-program compiler? What
+   about if you do static linking? "White-box" indicates the key thing, which is
+   that we want to see into parts of the compilation process that are normally
+   opaque. This is still a matter of degree: after all, even "black-box" normal
+   compilation has interface files.
+
+Here are some examples which I claim do some kind of white-box compilation. I
+have attempted to give a brief summary of some very complicated projects based
+on conversations I have had: any errors are my own.
+
+CLaSH
+"""""
+
+CLaSH compiles Haskell code into a low-level hardware description language. It
+works from Core.
+
+CLaSH currently gets this information from unfoldings, compiling dependencies
+``-fexpose-all-unfoldings`` to ensure they are present.
+
+GRIN
+""""
+
+GRIN is an experimental backend for GHC, which converts STG into its own IR. It
+works from STG.
+
+GRIN currently adds a GHC compiler plugin, which generates its own, new,
+interface files during compilation of dependencies. They patch Cabal to ensure
+that their interface files are installed. Then at the end they can load the GRIN
+interface files for dependencies and assemble the final result.
+
+Plutus
+""""""
+
+The Plutus compiler compiles Haskell into Plutus Core, a small System F-derived
+target language. This is done by a compiler plugin, which compiles targeted
+subsets of the Haskell program and embeds the compiled programs inside the main
+Haskell program as data, where they can be used at runtime. It works from Core.
+
+The Plutus compiler currently gets this information from unfoldings.
+Dependencies must explicitly mark functions which are to be used as ``INLINABLE``
+to ensure that unfoldings are present.
+
+Backpack-that-was
+"""""""""""""""""
+
+In the original issue for fat interface files, Edward Yang describes a system
+whereby indefinite units could be separately compiled and installed. But to
+actually instantiate the indefinite unit we need to do some work, and this
+requires more information than is in interface files, indeed it requires
+recompiling the indefinite unit.
+
+Backpack avoided this problem by becoming more interdependent with Cabal, so
+that particular instantiations of units could be requested and Cabal would
+handle rebuilding the unit and instantiating it.
+
+Other use cases
+^^^^^^^^^^^^^^^
+
+Dumping splices
+"""""""""""""""
+
+Obsidian Systems have `patched
+<https://gitlab.haskell.org/obsidiansystems/ghc/commits/wip/abrar/splices-8.6.5>`_
+GHC so that dumped splices (as from ``-ddump-splices``) can be loaded in when
+compiling to a non-native platform.
+
+This technique could also be used to "snapshot" compilation after TH has been
+eliminated. This would allow build systems to be more incremental.
+
+If there were support for compiling modules to Core and loading them, then we
+would avoid having to write additional code for this use case.
+
+
+Stage -1 Template Haskell
+"""""""""""""""""""""""""
+
+The `Staged TH proposal
+<https://github.com/ghc-proposals/ghc-proposals/pull/243>`_ would benefit from
+the ability to compile dependencies to Core and interpret them directly.
+Interpreting core is necessary for TH to be usable in GHC, due to ABI changes
+between stage 0 and stage 1.
+
+White-box interface files
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+I think Edward and GRIN have the right idea here: if we want to preserve
+information about separately compiled dependencies then we should add a new kind
+of artifact that contains it. This is an approach we have taken before:
+interface files and profiling outputs are prominent examples.
+
+We call this new kind of interface file a "white-box interface file", and it
+will have extension ``.hi-wb``. (Name and extension completely up for grabs!)
+
+What should go in a white-box interface file? I propose the following slogan:
+    
+    If a module has been loaded from a white-box interface file it should be as
+    if that module had been compiled from source.
+
+This avoids us trying to pick some set of information that will suffice in all
+cases. Instead, we preserve *everything*.
+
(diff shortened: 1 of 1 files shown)


simonmar:
This sounds like a good idea. 

I'm still not sure about using fat interface files to implement Backpack ("Backpack-that-was"). The rationale is all in the other thread that you linked to: https://gitlab.haskell.org/ghc/ghc/issues/10871

But I don't think this is your main motivation, right? The idea is to support full introspection of the compiler's internal representation for a previously compiled module. In that respect fat interface files are fine.

I prefer to put extra information into the `.hi` file itself, to avoid needing changes in external tooling.

One thing worth remembering is that we can't just defer all compilation until link time, due to `TemplateHaskell`. And we might have foreign code that depends on Haskell code, that is called by more Haskell code, that is called at compile time by `TemplateHaskell`.... essentially it's very difficult to disentangle compilation from Cabal at this point. That doesn't affect your goal of better introspection though.

michaelpj:
> I'm still not sure about using fat interface files to implement Backpack ("Backpack-that-was"). The rationale is all in the other thread that you linked to: https://gitlab.haskell.org/ghc/ghc/issues/10871
>
> But I don't think this is your main motivation, right?

That's right. I'm not at all qualified to judge whether Backpack-that-was would have been a good idea, and anyway that ship has definitely sailed. I mainly wanted to note that this is a feature of sufficient power that it would have let us implement it.

> One thing worth remembering is that we can't just defer all compilation until link time, due to TemplateHaskell.

Thanks, that's a good point. So any system which did want to do this thoroughly would have to have either a single-module mode or a core interpreter, as suggested [elsewhere](https://github.com/ghc-proposals/ghc-proposals/issues/162).

All the existing approaches get around this by working alongside normal codegen so we can use that for TH. This is not a bad approach (and for e.g. Plutus it is actively desirable), but something like GRIN might well eventually want to be the sole codegen output.

gridaphobe:
I like this proposal, but I think that (1) we should extend the `.hi` file rather than introduce another file, and (2) what we should really do is implement the alternate proposal for extensible (AKA fat) interface files.

The proposal already motivates the extensible interface files with GRIN; the main proposal would not address GRIN's issue as GRIN produces its own metadata that needs to be provided to downstream modules. Extensible interface files would also solve a long-standing pain point for using LiquidHaskell across packages, which faces the same problem as GRIN.

simonpj:
I'm sympathetic to the goal of having GHC spit out more info, useful to downstream clients of various kinds.  Some thoughts:

* I'm very unconvinced about making a distinction btween HIE files, HI files, and "whitebox" files.  They all do essentially the same thing -- albet with different goals leading to different kinds of information being stored.  Let's go for "fat interfacee files". Much easier conceptually, adn much easier engineering-wise -- e.g. no need to fiddle with Cabal.   (Conceptually, all this should be in the .o file, and technically that is possible, but I'm not arguing we shoulud go that far.)

* You state the goal of "serialise ModGuts"; but that's exactly what HI files are: the signature of `mkIface` is
  ```
  mkIface :: ... -> ModGuts -> IO (ModIface, Bool)
  ```
  So `mkIface` does precisely serialise most of `ModGuts` already!  Perhaps we should serialise more of it. That's a fine goal: let's just specify what "more" rather than invent a new mechanism.

* The `ModGuts` slogan is ambiguous, because `ModGuts` changes as you go down the compilation pipeline.  Do you want the program just after desugaring, or after optimisation, or somewhere in the middle?  I believe this matters in the Plutus case.

* The more you serialise, the more often the fat interface file will change, so the more recompliation will be triggered.  Mind you, GHC is carefult to track which _parts_ of the interface file it uses, and aborts compilation if the relevant bits haven't changed.  (You could argue that multiple interface files might reduced this problem, but I think the cure is worse than the disease.)

angerman:
@simonpj 
> (Conceptually, all this should be in the .o file, and technically that is possible, but I'm not arguing we shoulud go that far.)

@simonmar mentioned that as well, the other day. I am not absolutely sure this is true. For traditional object files yes. But what if our object files are not traditional object files, but say javascript files? IR files? ...

I'm all for making `.hi` files extendable by turning them into a bunch of sections. (E.g. serialise something of type `String -> ByteString` or similar and have every tool/consumer be responsible for coding/decoding their own section).  I am not convinced that we should force this capability into object files as we have way less control over the format/constraints object files for new architectures may lay upon us. For `.hi` files we have all the control we want.

simonpj:
>  I am not convinced that we should force this capability into object files

Yes I wasn't seriously suggesting doing so.   But I am suggesting that we should trreat that as the "right" plan, and keep as close to it as possible by putting all our own stuff into one fat .hi file, not several.

Ericson2314:
I generally like the "one extensible format" idea. I'd want to leverage trees that grow perhaps for this, e.g. for the different varieties of mod guts. My only concern is we don't end up in a postel's law mire of "the section you want might be missing, and you should be robust to that." Let's have everything just fail fast; we will need Cabal to be aware of the flags that change the sections as cache-invalidating, for example.

For using external build systems that work on the file system level, having separate files is preferable, but perhaps we can easily write something to split up the sections into additional files and recombine them again. Then we don't need to force multiple files or big files on any tool.

simonmar:
One reason we didn't put the contents of `.hi` files into a section in the `.o` file is that the timestamps on these two files can diverge - if we compile a module and its `.hi` file doesn't change, then we don't want to touch it, so that the build system won't recompile modules that depend on it, but we do want to touch the `.o` file to indicate that we ran the compiler. Of course this only applies to build systems that use GHC in one-shot mode, and anyway timestamps are really the wrong way to do this and we should be using content hashes or something else, but nevertheless this is one reason why we can't just combine the two files today.

michaelpj:
Okay, it sounds like there's a consensus that "fat" `.hi` files are the way to go. Great - I thought that might be unpopular, but I agree that it's a nicer solution.

Should I make a separate proposal for that, or should I incorporate it into this one?

I do think there is content to this proposal beyond just having fat interface files, because I still want to teach GHC to emit this particular kind of interface section. (As an alternative, one could imagine allowing plugins somehow to add things to interface files, but then you have to arrange for all your dependencies to depend on and activate the plugin.)

Regarding recompilation, at least for the use-cases I've listed, I think recompiling if any part of the interface has changed is the right behaviour. If you depend on the "white-box" interface content, then you definitely want to be re-run if that changes. Now, this may mean that if you ask for white-box content you're setting yourself up for more frequent recompilation, but that seems okay to me.

Perhaps a sufficiently clever compiler would be able to only update the white-box content and skip later stages, but I don't think we should rely on that.

> You state the goal of "serialise ModGuts"; but that's exactly what HI files are

I meant *faithfully* serializing `ModGuts`, i.e. serializing it in such a way that we can read back the original value. Current `.hi` files  aren't a faithful serialization, since you only get back a `ModIface`.

> The ModGuts slogan is ambiguous, because ModGuts changes as you go down the compilation pipeline. Do you want the program just after desugaring, or after optimisation, or somewhere in the middle?

Ugh yes, this is just an omission. I meant just after desugaring.

The reason for picking this point is that tools which want access to later stages can recover that by running the compilation pipeline again after loading the file. So the heuristic is "as soon as possible after the frontend is done".

I know John is interested in emitting code at other points too, but I think being able to get the code at the start of the pipeline and run forwards from there is a good starting point.

> My only concern is we don't end up in a postel's law mire of "the section you want might be missing, and you should be robust to that." Let's have everything just fail fast; we will need Cabal to be aware of the flags that change the sections as cache-invalidating, for example.

I certainly agree that changing sections should be cache-invalidating.

I think failing if the section you require is not present is reasonable. That does put it on the user to ensure that all their dependencies have been compiled with the right flags, and maybe eventually we'd want to teach Cabal about that.

gridaphobe:
I think it makes sense to extend the current proposal to include the fat interface file change.

nomeata:
> Regarding recompilation, at least for the use-cases I've listed, I think recompiling if any part of the interface has changed is the right behaviour. If you depend on the "white-box" interface content, then you definitely want to be re-run if that changes. Now, this may mean that if you ask for white-box content you're setting yourself up for more frequent recompilation, but that seems okay to me.
>
> Perhaps a sufficiently clever compiler would be able to only update the white-box content and skip later stages, but I don't think we should rely on that.

Currently, GHC has the very nice property that even if a dependency changes, it will not recompile a downstream module that only uses unchanged definitions from the changed module. Will this now be gone a soon as there is any additional data in the interface file?

michaelpj:
> Currently, GHC has the very nice property that even if a dependency changes, it will not recompile a downstream module that only uses unchanged definitions from the changed module. Will this now be gone a soon as there is any additional data in the interface file?

In the particular case of white-box content, you should be able to abort compilation in the same circumstances, since if the definitions it uses are unchanged then the white-box content it produces should be the same.

But I'm not sure this is safe for *consumers* of white-box content: a plugin that used white-box content could look at definitions that *weren't* directly used, and so might require recompilation even if apparently unused definitions are used. 

In fact, this seems like it might already be true today: what happens if a plugin looks at e.g. the unfoldings of a binding that is otherwise unused in the module being compiled? Will we get incorrect recompilation avoidance? Perhaps the problem is plugins!

simonpj:
> Current .hi files aren't a faithful serialization, since you only get back a ModIface.

...and you can typecheck that ModIface to get a ModDetails.  If you look
at ModDetails vs ModGuts you'll see that the latter is pretty much a superset of
the latter.   So to achieve what you want, it might be just a question of
beefing up what we have already, to serialse/deserialse a bit more.

> Ugh yes, this is just an omission. I meant just after desugaring.

Ah, now this is a *much* bigger difference from current .hi files.  Current .hi files have stuff
from the *end* of compilation, and must do.  You want stuff from somwhere
near the beginning.  Why?

And "near the beginning" is vague.   The desugarer generates lots of
junk, which is removed by a run of `CoreOpt.simpleOptPgm`.  You probably
don't mean "right after the desugarer"; you probably mean "after the simple
optimiser".  But why is that step magic?   Why not after further optimisation?

And other clients might want the bindings after some (but perhaps not
all) optimisation passes.

Somehow it seems a bit ad hoc to store three copies of the program in the .hi file:
* the original source code (I think, for IDE tooling)
* just after desugaring (for Plutus)
* after optimisation (for GHC separate compilation)

ndmitchell:
FWIW, the DAML Compiler (which isn't a million miles away from Plutus) definitely does want the output just after desugaring, before running `simpleOptPgm`. DAML uses the GHC API to get that, and does its own compilation driver, but a fat `.hi` could enable lots of things like TemplateHaskell that are otherwise tricky.

Ericson2314:
I am having second thoughts about the "one extensible file" idea. If the interesting bits are caching and cache invalidation and other compilation manager concerns. Putting everything in one file doesn't so much abstract over unimportant details as shove the all the important details under the rug! Interfaces vs definitions (including today's unfoldings) are invalidated and accessed differently. More files simplifies this by eliminating the dynamism in the dependency graph before parsing.

Remember, the whole raison d'etre of the proposal was that loading core from whitebox compilation files to finish compilation should be simpler. This does including GHC itself loading core, not just Plutus and other tools. We don't want to load a bunch of hi files, realize they all contain complete core for all deps, and then wildly change the compilation plan.

Thinking back to #245 we already have serious issues that the behavior (let alone implementation) of `ghc --make` is very complex and hard to specify. I want to leverage more files *statically* to break things up into smaller, more numerous, but simpler steps. If we use big extensible files, it's just too tempting to not statically decide which sections those files should contain, but see what they actually do contain make numerous of dynamic planning decisions.

Ericson2314:
Take https://gitlab.haskell.org/ghc/ghc/merge_requests/1633#note_219575 (the PR and @mpickering's) comment in particular. I too an confused by the `-fno-code` + `-fwrite-interface` interaction; I think modifying generated interface files would wreck havoc on `ghc --make` alternatives, but putting this post-code-gen info in a separate file is the best of both worlds:

 - Get information as soon as possible (only wait for type checking when relevant)

 - No mutation of files (good for file-based incremental computation)

 - In conjunction with this proposal supports highly incremental file-based builds:

   1. GHC invocation 1: parse, type check, desugar, serialize

   2. GHC invocation 2: load core, stg, cmm, code gen, write caf info / lf info / etc.

angerman:
@Ericson2314 note that fat/white-box interface files and extendable interface files are not necessarily the same. While you can model some form of fat interface files with extendable interface files, extendable interface files let you just attach additional data in `.hi` files. What you do with that is completely up to you. If it's a good fit for your specific use case or not, is up to you in the end. Are extendable interface files the ultimate solution? I have my doubts, but I think it's a minimal invasive change to ghc, that provides additional avenues to explore and experiment with.

Ericson2314:
I'd agree with that, but with the caveat that "minimally invasive" isn't necessary good.

Based on this discussion, #245, and my refactoring for multi-package and fixing some backpack bugs, I'm starting to think that the underlying tech debt that's entangled with all of this is that ModuleGraph is too course grained. Instead of leveraging the dependency graph as much as possible for caching, we have lots of ad-hoc incrementality cache invalidation code. More dependency nodes  naturally works well woth more files, with all the benefits I mentioned above. I might write a proposal for this.

So back to the extensible interface files, it is a small change, but one that further combinatorially grows the possible work denoted by the one dependency node. I worry it would be harder to remove once it's introduced too, and what logic might spring up downstream in support of it. Conversely if we can add new functionality *while* clarifying the monster that is `ghc --make` so other build systems can better immitate it, I'm willing to take a risk making a bigger up front change.

michaelpj:
After some discussion, we've decided that this isn't a good fit for the proposal process. I've started a [GHC wiki page](https://gitlab.haskell.org/ghc/ghc/wikis/Core-interface-section) which has an updated version based on the idea of having a new section of interface files with the Core in it.

PR #263, opened 2019-08-14
WIP: Multi package

Ericson2314:
[Rendered](https://github.com/Ericson2314/ghc-proposals/blob/multi-package/proposals/00-multi-package.rst)

This is only WIP because the exact command line needs to be bikeshed.


Ericson2314:
CC @DanielG

phadej:
This is complete opposite development than in https://github.com/ghc-proposals/ghc-proposals/pull/245 which would make it possible to use GHC as single shot compiler.

Note, we don't ever, ever want to **compile** multiple packages at once, as we don't really want to compile even single one "at once".

Here's a a very difficult impedance mismatch what's good for compilation, and for interactive use. Yet, I think that with *extended dependency generation*, tools could operate `ghci` like process.

---

I.e. rather than adding all possible "build tool" functionality into GHC, I'd rather move into direction into moving everything not-strictly necessary out of it.

---

As a concrete challenge: If I work on multi-component Backpack project. How this would work?

Ericson2314:
@phadej the short answer is, yes, I'm afraid I think it's best to refer to peruse these contradictory dependencies in the long term. As a big nix user (where all caching is at least process granularity) who is constantly trying to make packages leverage the build system rather than reinvent the wheel trust me I highly value that goal. But in the short term I think multi-repl is just too important to put off. There's a lot engineering needed to make build systems low latency enough for GHCi.

Longer term I think the contradiction can go away. @ulysses4ever was talking about a persistent daemon for build systems that would do Haskell builds *and* leverage #245 to tell the build system about dependencies. Even longer term I hope we will have libraries that do build system / incremental computation with extremely fine grained atoms and yet ironclad hashing (think serializing a tiny process in memory, hashing it, and then forking it).

@ezyang pointed out the contradictions long ago http://blog.ezyang.com/2015/12/the-convergence-of-compilers-build-systems-and-package-managers/ so it's certainly an issue that's been building for a while.

gbaz:
"What #10827 glosses over however is if you do the data structure change (part 1) then there is no need for part 2 because part 1 subsumes it. We therefore go straight for the data structure change."

Is this true? I don't understand the data structures and their usage enough to be confident in this...

Ericson2314:
@gbaz OK yes I should clarify
```
(HomePackageTable, DynFlags, UnitID)
```
Part 1 complains we have just one `HomePackageTable`, and part 2 complains we have 1 `DynFlags`, so that's orthogonal. It proposes the stop gap for part 2

> since we already track DynFlags on a per-module basis, we just have to arrange for a "package" level DynFlag to be applied to the correct ModSummarys.

The PR we have switches to
```
Map UnitId (HomePackageTable, DynFlags)
```
which solves both problems 1 and 2 in one fell swoop, and obviates the need for that per-module hack.

mgsloan:
Great stuff @Ericson2314 , I'm glad you are digging into this!

One way to approach designing the CLI for this might be to design GHCi commands to support this.  A minimum viable implementation of this wouldn't need to make the feature available from commandline flags - instead `stack repl` and other implementations can generate a `.ghci` script to load.

One way to slice this might be to have the following GHCi commands:

* `:configure PATH_0 ... PATH_N` which takes the current set of dyn flags and perhaps things like current directory and associates them with the modules specified by path.  The reason to specify these by path rather than module name is that it would be very nice for this to support multiple modules with the same name (package qualified imports).

  - It might also be useful to have these paths be source directories.  This would be quite useful for manual usage.

* `:unset-all` which resets the current set of flags.  There's likely some trickiness here - some flags are currently not unsettable - likely flags which also have some trickiness for being per-module.  The `.ghci` files generated by tools would use this to reset the flags before setting the flats used for each component.

Then, subsequent uses of `:module` or `:load` will use the flags configured for the modules.  I just made up the names `:configure` and `:unset-all`, certainly feel free to bikeshed.

I think this should cover most of it.  The only tricky case I can think of right now is where multiple packages use the same source files.  I believe this can be resolved via an extension to the idea above, which is to allow multiple `DynFlag` configurations for a particular module, disambiguated by `PackageName`.

> Note, we don't ever, ever want to compile multiple packages at once, as we don't really want to compile even single one "at once".

Why not?  It seems reasonable to me to expect that `ghc --make` can be much more efficient than invoking ghc multiple times.  As @Ericson2314 describes, such a feature would be quite handy for implementing a persistent ghc build daemon which keeps info in memory.

Ericson2314:
@mgsloan 

While I agree most build system would benefit from not using `ghc --make`, I see no reason to prevent `ghc --make` from working with this. OTOH that means we're probably best off going strait to a CLI as an ghci-only interface won't allow testing that.

----

@ezyang Could the `.bkp` file processing used for tests perhaps be normal with this, too?

ezyang:
@Ericson2314 Not sure what you mean by "be normal" with this :)

Module stability computation in `compiler/main/GhcMake.hs` will need to be adjusted in a multi-HPT world, as stability computation must extend across home packages.

Ericson2314:
@ezyang that all sounds correct. Basically I was wondering about the `unitIdsToCheck` to function used in `GhcMake`, and likewise the multiple units in a `.bkp` file. Should each of those units be a home package with this?

ezyang:
Yeah, I think that would make sense.

mpickering:
This change would benefit projects such as `h-i-e` purely from the fact that there would be a common language to specify multiple dependent components. Therefore I am strongly against any interface such as that suggested by @mgsloan which ties any of these flags into a GHCi script or the like. 

At the moment supporting multi components projects in tooling is close to impossible as extracting the interdependencies between the different components and what should update when requires a lot of help from the build tool. Tools which just operate on the GHC level tend to be much simpler and robust as you don't have to support N different build tools to work out the same information. 

@phadej is right that there is another proposal pulling in the other direction but whilst GHCi is still a thing built into GHC, then you still need both. I don't think #245 proclaims anything about GHCi?

mgsloan:
@mpickering How else are build tools supposed to express how to load multiple packages into GHCi?  I don't believe that HIE does its own build plans, and instead defers to other tools like cabal and stack to prepare the environment and specify the GHCi flags.  I don't follow your reasoning about why such an interface would be undesirable or what such a "common language" would constitute.

Also, I think it is is highly desirable to be able to support this for the combination of vanilla ghci with an external tool.  Adoption of tools like ghcid, dante, stack repl, cabal repl, etc is good evidence that folks still use vanilla ghci for more complicated development scenarios.

Ericson2314:
@mgsloan I think @mpickering is agreeing on the meat of the change. I think is criticism of the ghci-script thing is merely then that only ghci (and not regular `ghc --make` or Haskell DIE Engine) can use it-.

Ericson2314:
>  I don't believe that HIE does its own build plans, and instead defers to other tools like cabal and stack to prepare the environment and specify the GHCi flags

Right, but if those tools can produce the multi component ghci flags or whatever, then hie can just use that. The point is that stack and cabal don't need to think ghci vs hie vs other tool. There just needs to be one way to "declare the `HscEnv` setup" for all ghc-involved things. 

I would actually say that flags and ghic scripts are both poorly structured, and it would be better to have haskell data + json schema for that stuff.

Ericson2314:
Random aside, I stumbled on https://gitlab.haskell.org/ghc/ghc/commit/61d2625ae2e6a4cdae2ffc92df828905e81c24cc. I feel like this proposal is about finishing what that started.

Ericson2314:
I updated the description of https://gitlab.haskell.org/ghc/ghc/merge_requests/935 a bit; made more progress getting backpack things to work.

Do take note of the bit about wanting to do https://github.com/ghc-proposals/ghc-proposals/pull/266 for `HscEnv`. The data structures are rather monolithic, and with this change even more poorly factored. Tracking the flow of information more finely with `Has*` constraints or similar would make it easier to see at a glance what parts of the code need what, and greatly inform moving fields around.

One example of the problem is it's weird to store some unit id information in the dynflags, and some in the key to the map of `InternalUnitEnv`s. `HscEnv.hsc_currentPackage` and `DynFlags.thisPackage` do the same thing.

mgsloan:
Oh yeah, my suggestion was not to only expose this through GHCi, but rather that might be the first place to expose it.  This is why I said "minimum viable implementation".  Certainly it would also be desirable to expose it in the CLI.  My reasoning for this is that mutation of some context is more natural as GHCi commands.  That said, there is precedent for GHC's commandline to be order dependent, so perhaps the focus should instead be on CLI flags for this.

Ericson2314:
I cannot decide whether the keys of my map should be `UnitId`s or `InstatlledUnitIds`. Anyone know?

andreabedini:
I believe this has been implemented?

https://gitlab.haskell.org/ghc/ghc/-/merge_requests/6805

Ericson2314:
Oh good catch I forgot about this. And the recent Cabal work from @mpickering is making it more usable too.

Issue #198, opened 2019-01-25
Unidirectional Coercible


int-index:
For newtypes that guarantee some invariant, we might want to coerce in one direction (unwrap) but not the other.

```
newtype NonNegative a = NonNegative a

coerce :: NonNegative a -> a -- safe
coerce :: a -> NonNegative a -- unsafe
```

Right now the only option is to not export the constructor, which makes both directions of `coerce` unavailable.

nomeata:
This has been in the back of our heads for a while. But there are a few can of worms, such as now type constructors have not only roles, but also variances....

goldfirere:
I agree that this would be nice, but doing it right would take at least one peer-reviewed publication, perhaps more. Indeed we would need variances, but we also need a new theory of FC/Core that can deal with non-symmetrical coercions. This would be interesting for me to pursue (and it's related to ideas I've wanted to write about), but no one should hold their breath.

monoidal:
Linear types could use unidirectional `coerce :: (a ->. b) ->. (a -> b)`.

goldfirere:
Is this ticket achieving any goal right now? I move to close it. This won't be solved anytime soon, I don't think.

int-index:
> Is this ticket achieving any goal right now?

It's a feature request.

> This won't be solved anytime soon, I don't think.

If this is ground for closing, so be it. I think it's good to have it open as a reminder that this is a desired feature, but I don't feel strongly.

goldfirere:
I actually agree with

Ericson2314:
The axioms proposed to be removed in #274 could be added back as one-way coercions. This is better than the situations both with and without #274 because it avoids the cost and semantic change of eta-expansion.

Icelandjack:
With dependent types is there a way to somehow add a proof obligation? Taking the `NonNegative example as an example, pretending `Coercible` has user instances:

```haskell
instance Coercible (NonNegative a) a
```

but in the other direction.. I assume this won't make sense but it doesn't hurt to ask

```haskell
instance pi (x :: a). (Num a, Ord a, x >= 0) => Coercible a (NonNegative a)
```

Ericson2314:
I like this.

One small tweak:
```haskell
instance pi (x :: a). (Num a, Ord a, (x >= 0) ~ True) => Coercible a (NonNegative a)
```

ju1m:
For the record, I tried the following hack:
```haskell
{-# Language RankNTypes #-}
[...]

PR #265, opened 2019-08-25
Unlifted Datatypes (under review)

sgraf812:
[Rendered](https://github.com/sgraf812/ghc-proposals/blob/unlifted-data/proposals/0000-unlifted-datatypes.rst)


nomeata:
Thanks!

> Implement the Strict data type only. Doing so provides the same semantics at the cost of more syntactic overhead.

More syntactic overhead, but also more composable! I don’t need `Bool` and `StrictBool` and `Maybe` and `StrictMaybe`, I can use `Strict Bool` and `Strict Maybe`.

At first glance, this makes me lean much more towards the `Strict` proposal.

nomeata:
Ah, but this allows you to implement `Strict`, so we can have our cake and eat it too. Compelling!

sgraf812:
Yes, the only real drawback compared to implementing `Strict` only would have been that this is a superset of the functionality #257 provides, so ideally should be implemented in a separate proposal. But having done a little prototyping, I think that we would have to make `Force` an actual `DataCon` anyway, so that it is properly handled by the simplifier and other passes. At this point adding the tiny bit of surface syntax is just a stone throw away, hence @simonpj argued in favor of this proposal instead.

sgraf812:
> cross-kind UNPACK is slightly novel and should be explicitly spelled out in the proposal

What do you mean by cross-kind? It's just unpacking the regular constructor fields like before, that's entirely orthogonal to the semantic concern of there being no bottom as an inhabitant of that constructor's type.

Unless you mean the implicit unpacking of constructors like `Force`, that's an entirely different matter but with no consequences for perceived semantics, it's just an optimisation affecting operational semantics. Note that this doesn't UNPACK `Force`s argument, rather we want to coerce `Strict a` to `a` in Core like we do for newtypes, except that one way (`a` to `Strict a`) isn't actually zero-cost, but erase to a `seq`. See [Simon's earlier comment](https://github.com/ghc-proposals/ghc-proposals/pull/257#issuecomment-523063835) and my remarks in https://gitlab.haskell.org/ghc/ghc/wikis/unlifted-data-types#proposal-b4-levity-polymorphic-functions.

> I hope someday every data type can have an associate unboxed version connected by a type family (let's call it Unbox)

There are plans for having levity polymorphic (unlike the current meaning of levity polymorphism, which is more like representational/shape polymorphism) data types. See https://gitlab.haskell.org/ghc/ghc/wikis/unlifted-data-types#proposal-b3-levity-polymorphic-data-types for a design I would aim at.

simonpj:
Generally I like the idea here, and I don't think it'd be all that hard to implement.  We already have unlifted data types like `Array#`; this just makes them user-definable.

On syntax, a postfix kind signature is rather a quiet signal for such an important change.  I'd suggest using a prefix keyword:
```
data unlifted SPair = SP Int INt
```
(This "keyword" would have significance only after `data` so it wouldn't prevent you using "unlifted" as, say, a variable name.

sgraf812:
@Ericson2314 If I understand correctly, then I don't think what you want to have is possible. You can't unpack polymorphic fields, at least not without specialisation of the data type, at that's certainly out of scope for this proposal. To exemplify, what is the static offset of the `b` field in `UPairStrict`? If we'd unpack the `a` field, then it depends on the size of the representation of `a`, which means you can't use `UPairStrict` in code polymorphic over `a`.

`UNPACK` concerns the fields of a constructor (`Force`, in particular). After unpacking, the heap object still has its own header. I was rather aiming at making the constructor `Force` have no representation at runtime at all, because its only field is conveniently of the exact same representation as the constructor itself (i.e. a pointer to the heap).

This is probably most similar in spirit to turning a lifted data type with a single strict boxed field (So as `Strict` above, but with regular `'LiftedRep`) into a `newtype`, except that that would be broken semantics because a pattern match on the constructor actually evaluates the field (as opposed to newtypes).

simonmar:
The proposal mentions that we could have an unlifted `Ptr` type. But that wouldn't be a drop-in replacement for the current `Ptr` type, because it has an unlifted kind and therefore can't be passed to polymorphic functions or stored in polymorphic data structures, right? e.g. you can't have a `[Ptr a]` when `Ptr` is unlifted.  That should probably be mentioned as a drawback. 

(aside from this I'm strongly in favour of the proposal)

sgraf812:
@simonmar Yes, currently we have no means of encoding the necessary levity polymorphism. See https://gitlab.haskell.org/ghc/ghc/wikis/unlifted-data-types#proposal-b3-levity-polymorphic-data-types in that regard. I think levity polymorphism (i.e. lifted vs. unlifted, but still boxed pointer to GC'd heap, as opposed to the current interpretation, which is more like representation/shape polymorphism) should be discussed in a subsequent proposal.

Or maybe it should be part of this proposal, I've no idea what the guidelines are wrt. incrementality of GHC proposals. But I'd rather implement it in separate steps.

Ericson2314:
@sgraf812 Right, I completely forgot about in normal cases need a concrete constructor to unpack, so no polymorphism there.

But perhaps we do anything special with `data AlmostNewtype a = AlmostNewtype !a` today? By code gen, whether matching on `AlmostNewtype` forces anything should be irrelevant, and the same unboxing opportunities should be available. Perhaps we should, so the `Strict` unpacking is less of a one-off optimization.

Back to my original, question I assume plain:
```haskell
data Strict :: TYPE 'UnliftedRep where
  Strict :: {-# UNPACK #-} !Concrete -> Strict a
```
```haskell
data UPairStrict :: TYPE 'UnliftedRep where
  UPairStrict :: {-# UNPACK #-} !Concrete0 -> {-# UNPACK #-} !Concrete1 -> SPair a b
```
works fine?

-----

I do really really want "template" / "monomorphized" polymorphism, but I think should be separate proposal. There is much to decide about that alone.

code5hot:
Not knowing about this proposal I just added #268. I noticed that we already have support for unlifted types with just one constructor via newtype and that it can be extended to more types with possible compiler simplification by implementing case analysis in two ways, one for `data` and one for `newtype` instead of mere coercing (I think I read that's what you do).

sgraf812:
Coercing between unlifted and lifted rep are not actually essential to the proposal. Levity polymorphic coercions might be interesting to have in Core to support a zero-cost `Strict` data type, but that's irrelevant to Haskell as a language. It *might* become relevant later on when we want to have levity polymorphism (in the sense of lifted vs. unlifted, not boxed vs. unboxed) in the surface language.

I don't see how implementing case analysis in two ways leads to compiler simplification. Note that `case` already works on terms of unlifted rep, so my proposal wouldn't actually imply a change to semantics at all. Everything is already in place! We just offer the needed vocabulary to the programmer to define her own unlifted data types.

code5hot:
What are the semantics of Strict?

sgraf812:
> By code gen, whether matching on `AlmostNewtype` forces anything should be irrelevant, and the same unboxing opportunities should be available. Perhaps we should, so the `Strict` unpacking is less of a one-off optimization.

Hmm. I felt a strong urge to disagree but couldn't come up with an example. So perhaps eliminating indirections for types like `AlmostNewtype` (and `Strict` of course) is indeed a sound transformation, if a little surprising for debugging purposes.

> Back to my original, question I assume plain:
> 
> ```haskell
> data Strict :: TYPE 'UnliftedRep where
>   Strict :: {-# UNPACK #-} !Concrete -> Strict a
> ```
> 
> ```haskell
> data UPairStrict :: TYPE 'UnliftedRep where
>   UPairStrict :: {-# UNPACK #-} !Concrete0 -> {-# UNPACK #-} !Concrete1 -> SPair a b
> ```
> 
> works fine?

Yes, it should just unbox the fields, as today.

> I do really really want "template" / "monomorphized" polymorphism, but I think should be separate proposal. There is much to decide about that alone.

The best bet is probably a `backpack` inspired solution. See [`unpacked-containers`](https://hackage.haskell.org/package/unpacked-containers) for example.

sgraf812:
> What are the semantics of Strict?

[This is the definition](https://github.com/sgraf812/ghc-proposals/blob/unlifted-data/proposals/0000-unlifted-data.rst#31the-strict-data-type). Its semantics follow straight-forward from the semantics of unlifted expressions (i.e. everything unlifted is evaluated by default, because that's the only way to make sure there is no bottom anywhere).

See also #257.

code5hot:
can I pattern match on `x : Strict (GHC.Tuple.Unit a)`?

    case x of {Force (Unit y) = y}

what is `undefined :: Strict (GHC.Tuple.Unit a)` ? is it still `_|_`?

code5hot:
Oh I see you can pattern match as above, the tree example is missing parentheses.

sgraf812:
I updated the Change Specification to be more formal. There's an open TODO discussion concerning syntax, so I expect this thread to be on fire very soon 🔥

niobium0:
Great proposal! I wish it at least sketched the path to the zero-cost `Strict` data type. 

The worker-wrapper example is what motivates me write this. Here's the status quo example from the proposal.
```
foo :: Int -> SPair Int Int
foo x = case $wfoo x of (# a, b #) -> SPair a b

$wfoo :: Int -> (# Int, Int #)
$wfoo x
  | even x
  = (# (x + 1), x #)
  | otherwise
  = case $wfoo (x-1) of
      (# a, b #) -> (# a+1, b+1 #)
```
To me it's a bit surprising that the type of `$wfoo` is not `Int# -> (# Int#, Int# #)` (maybe I'm overlooking something trivial). I'd like to see worker-wrapper transform that produces the second type signature, and I'm assuming it will happen at some point. And it would be a shame if `Strict` would get in the way of it because it cannot accept unboxed arguments.

sgraf812:
> Great proposal! I wish it at least sketched the path to the zero-cost Strict data type.

Thanks! I think zero-cost `Strict` can be an implementation detail (without which w/w to it makes no sense, though).

> To me it's a bit surprising that the type of $wfoo is not Int# -> (# Int#, Int# #)

What you want is [Nested CPR](https://gitlab.haskell.org/ghc/ghc/issues/1600) ([WIP MR](https://gitlab.haskell.org/ghc/ghc/merge_requests/1866)). It's not entirely trivial to unbox the inner `Int`s.

andrewthad:
I really like the meat of the proposal, but I'm not convinced that all the applications are accurate. I do think that we should be able to write `data Foo :: TYPE 'UnliftedRep` (or something like that). It's a clear win. I need this all the time.

However, here are two of the applications that I have qualms with. One is the example with `Ptr`. `UnliftedNewtypes` will be in GHC 8.10. There's a section [in the UnliftedNewtypes proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0098-unlifted-newtypes.rst) that talks about an unlifted `Ptr` type as an application. Specifically, you will be able to write:

    newtype Ptr# :: Type -> TYPE 'AddrRep where
      Ptr# :: Addr# -> Ptr# a

The allocation-related guarantees provided by this type are strictly better than the guarantees provided by the type suggested in this proposal:

    data Ptr a :: TYPE 'UnliftedRep where
      Ptr :: !Addr# -> Ptr a -- also there shouldn't be a need for the bang here

In particular, `Ptr#` leads to data constructor allocations while `Ptr` does in situations where worker-wrapper doesn't work (like when you pass a function as an argument to a function that doesn't inline). So I don't find that particular use case motivating.

The other problem I have is with the `Strict` data type. I'm troubled by the idea of turning a value strict by wrapping it in another data constructor. I know that the bang inside it means that the argument gets forced everywhere. It's just strange to me that it accomplishes strictness by using a mechanism that lives outside of the type system (bangs on data constructor fields), which inserts `seq` everywhere the constructor is applied. Similar to the `Ptr` example, this also doesn't work correctly with higher-order functions that don't inline. In fact, it would leads to *more* allocations than you would have seen with the lazy type.

I have a [half-baked proposal](https://gist.github.com/andrewthad/90a4b23051870d14afb4ba064699bc5b) that puts strictness in kinds instead and doesn't suffer from this problem, but it's more complicated, changes the type of `seq`, and would introduce coercions all over the place in GHC Core, since data types would all be desugared to newtypes over the real under-the-hood strictness-polymorphic data type.

Ultimately, I don't think that dual-purpose types that communicate both strictness and laziness is really that big of a deal though. I just don't really need thunks very often. With most types I deal with, I want values to be strict all of the time, and as it stands, this proposal gives me a way to get that guarantee. So, +1.

Ericson2314:
The optimization that works to unbox
```haskell 
data AlmostNewtype a = MkAlmostNewtype !a
data AlmostNewtype' a :: Type 'UnliftedRep = MkAlmostNewtype' !a
```
I think is a good one, even if `Strict` is replaced with something a bit more straightforward like your `forget` and `seqTyped` (which, BTW, I don't see why we couldn't use `LiftedRep` and `UnliftedRep` if we someday retrofit existing data definitions to be more polymorphic). As such, I don't really mind `Strict` in the meantime as it's not actually a hack for a single nominal type. Ultimately we'll want https://github.com/ghc-proposals/ghc-proposals/issues/198 to clean all this stuff up.

sgraf812:
> However, here are two of the applications that I have qualms with. One is the example with `Ptr`.
>
> In particular, `Ptr#` leads to data constructor allocations while `Ptr` does in situations where worker-wrapper doesn't work (like when you pass a function as an argument to a function that doesn't inline). So I don't find that particular use case motivating.

Yes, your observation is spot on. There's no compelling reason to choose `Ptr` over `Ptr#`, other than that `Ptr` is of a different kind. But that also means we can't get rid of the indirection in higher-order scenarios. I think I'll just delete that example.

> The other problem I have is with the `Strict` data type. I'm troubled by the idea of turning a value strict by wrapping it in another data constructor. I know that the bang inside it means that the argument gets forced everywhere. It's just strange to me that it accomplishes strictness by using a mechanism that lives outside of the type system (bangs on data constructor fields), which inserts `seq` everywhere the constructor is applied. 

Right. One could argue that we should rather formulate bang patterns in terms of `Strict`, but that horse is out the barn. Semantically, `Strict` is exactly what we want, but operationally we really don't want it to be a data constructor. Good news is that this part could entirely be a follow-up proposal.

> Similar to the `Ptr` example, this also doesn't work correctly with higher-order functions that don't inline. In fact, it would leads to _more_ allocations than you would have seen with the lazy type.

Yes. In a levity polymorphic world, we might want to call `dataToTag#` on a `Strict` thing in a polymorphic scenario, in which case we can't elide the data con:

```haskell
f :: (forall a :: TYPE 'UnliftedRep). a -> Int#
f x = dataToTag# x
```

If we call this on `Force True`, I'd expect to get back `0#`, but with the optimisation that tries to remove the indirection we would get `1#`. Bugger. So we can't do the optimisation if the `Strict` escapes into a polymorphic context.

It really seems like we need a coercion-inspired approach (like in your proposal or in https://gitlab.haskell.org/ghc/ghc/wikis/unlifted-data-types#proposal-4-allow-unlifting-existing-data-types-with-no-overhead) to solve this.

> I have a [half-baked proposal](https://gist.github.com/andrewthad/90a4b23051870d14afb4ba064699bc5b) that puts strictness in kinds instead and doesn't suffer from this problem, but it's more complicated, changes the type of `seq`, and would introduce coercions all over the place in GHC Core, since data types would all be desugared to newtypes over the real under-the-hood strictness-polymorphic data type.

I left some comments there.

> Ultimately, I don't think that dual-purpose types that communicate both strictness and laziness is really that big of a deal though. I just don't really need thunks very often. With most types I deal with, I want values to be strict all of the time, and as it stands, this proposal gives me a way to get that guarantee. So, +1.

Yes, I think this proposal can be largely orthogonal to what you sketch out over in yours. This proposal doesn't try to tackle levity polymorphism just yet, it's just some mostly uncontradictory ground-work that needs to be done first, together with some ideas of how to use it.

andrewthad:
It sounds like we're more-or-less in agreement about the important things. Thanks for writing this up and trying to tackle this. One other minor concern is that the proposal suggests both a GADT-syntax and a H98 syntax for unlifted data types:

    data unlifted Foo = ... --- H98
    data Foo :: TYPE 'UnliftedRep where -- GADT

My concern is that the `unlifted` keyword is optional when using GADT syntax. Because of the kind signature, the meaning is still clear, but this would establish a precedent that a kind signature can change the runtime behavior of a program. If that user had instead written:

    data Foo where

The runtime behavior of their code would be different. CUSKs on data types can currently influence where or not the compiler accepts them, but the never change the operational semantics of a program. (I could be wrong about this though.) I think it would be more consistent to require:

    data unlifted Foo :: TYPE 'UnliftedRep where -- GADT

I don't think this is big deal. It's my aesthetic preference, but I thought I'd mention it in case it resonated with anyone else.

sgraf812:
I didn't have a strong opinion on this before and wasn't sure what would be the best solution. But I agree now that we should probably require the `unlifted` keyword either way and that the KindSignature (if any) should match that keyword.

simonpj:
Under "motivation" the proposal says
```
data unlifted Tree a
  = Branch !(Tree a) a !(Tree a)
  | Leaf
```
But in this defintion the ! are redundant.  Just omit them!

sgraf812:
Ha, good point!

simonpj:
> Now for the rest of this proposal, we assume that the plan outlined in #16970 has landed, in particular for its operational consequence that strict constructor fields always contain tagged pointers (meaning their tag is never zero, like for thunks).

I think that patch is 100% irrelevant to this proposal.

In this proposal, if T is delcared with `data unlifted T =...`, then a value of type T is always represented by a properly-tagged pointer to the value, period.    Just like `Array# t` today.   The MR #16970 concerns *strict fields* of a data constructor, where the strict field is a *lifted* data type, say `Bool`.  Thus:
```
data Ordinary = MkOrdinary !Bool Int
```
Here the patch adds a new invariant that although some Bool-values can be bottom, and represented by an untagged pointer, the Bool in a MkOrdinary is always always a properly-tagged pointer to a boolean value.

Nothing to do with this proposal!

simonpj:
> topdecl -> 'data' 'instance' [ 'unlifted' ] [ context => ] ...

No, I don't think you can do this for data instances -- unless you make the data *family* (and hence all its instances) unlifted.  OK you can do that, but please spell it out.

sgraf812:
> I think that patch is 100% irrelevant to this proposal.

Correct. I deleted that paragraph.

> No, I don't think you can do this for data instances -- unless you make the data family (and hence all its instances) unlifted. OK you can do that, but please spell it out.

We don't have to do anything specific for data families (well, other than making sure result kinds match up), the generalisation already landed with the UnliftedNewtypes proposal. I added a paragraph to the static semantics section:

> Data family instances may be declared unlifted, in which case the data family application's result kind must reduce to TYPE 'UnliftedRep. See [the section on data families in the UnliftedNewtypes proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0098-unlifted-newtypes.rst) and Note [Implementation of UnliftedNewtypes] for details involving type-checking the parent data family.

sgraf812:
Here's a very rough [prototype](https://gitlab.haskell.org/ghc/ghc/tree/wip/unlifted-data) that miraculously just works on the following file:

```haskell
{-# LANGUAGE UnliftedDatatypes #-}
{-# LANGUAGE MagicHash #-}

module Lib where

import GHC.Exts

pack :: Bool -> Bool -> Int#
pack False False = 0#
pack False True  = 1#
pack True  False = 2#
pack True  True  = 3#

data unlifted SBool = STrue | SFalse

spack :: SBool -> SBool -> Int#
spack SFalse SFalse = 0#
spack SFalse STrue  = 1#
spack STrue  SFalse = 2#
spack STrue  STrue  = 3#

data Tree a
  = Branch !(Tree a) a !(Tree a)
  | Leaf

tsum :: Tree Int -> Int
tsum Leaf           = 0
tsum (Branch l x r) = tsum l + x + tsum r

data unlifted STree a
  = SBranch (STree a) a (STree a)
  | SLeaf

stsum :: STree Int -> Int
stsum SLeaf           = 0
stsum (SBranch l x r) = stsum l + x + stsum r
```

And it generates *exactly* the Cmm we expect it to generate (so no zero tag checks for unlifted variants).

AndreasPK:
It occurred to me that while we want these types to be unlifted, I think we don't want them to be UnliftedRep as it's more restrictive than desirable.

In particular it's not uncommon to have top level static data. Either explicitly user written, or floated to the top during code transformations.  
Something not possible for things of UnliftedRep currently.

sgraf812:
I think the solution to the top-level problem is not declaring them lifted, but rather making sure that we only allow unlifted normal forms (NB: not only WHNF) at the top-level. Problem solved!

sgraf812:
@nomeata This is now ready for review by the committee. Is there anything more I have to do?

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,507 @@
...
+
+Costs and Drawbacks
+-------------------
+Thanks to previous work on unlifted types and ``-XUnliftedNewtypes``, this
+proposal seems rather easy to implement, with the majority of changes happening
+in the parser and type-checker. Notably the backend is not only affected **at
+all**. Very good cost to benefit ratio.
+
+As for the risk of making the language harder to learn: Beginners won't come in
+touch with unlifted newtypes at all. Unless they crave for better performance
```
simonmar: s/unlifted newtypes/unlifted datatypes

simonmar:
I'm having trouble with the worker-wrapper example. The proposal suggests that we could end up with 

```
foo :: Int -> SPair Int Int
foo x = case $wfoo x of (# Force a, Force b #) -> SPair a b

$wfoo :: Int -> (# Strict Int, Strict Int #)
$wfoo x
  | even x
  = (# Force (x + 1), Force x #)
  | otherwise
  = case $wfoo (x-1) of
      (# Force a, Force b #) -> (# Force (a+1), Force (b+1) #)
```

But every `Force` here is a box, so this doesn't actually help unless there's also some mechanism to elide these boxes.  Indeed the proposal calls out this lack in the section on Effect and Interactions:

> This proposal consciously left out [...] details of whether we should eliminate the indirection in constructors like Force (we certainly should!) and to what degree we could infer and let the user omit Force constructors.

so given that this isn't part of the current proposal, is the work-wrapper example relevant here?

Moreover, the `pack` example suggests

> we can wrap all Bool s in Strict (see above) or define a new unlifted SBool 

but this doesn't work with `Strict`, due to the box.

simonmar:
> It occurred to me that while we want these types to be unlifted, I think we don't want them to be UnliftedRep as it's more restrictive than desirable.

It's not clear to me how you could have unlifted types without them having kind `UnliftedRep`. It's precisely the restrictions on types of kind `TYPE 'UnliftedRep` that give us the guarantee that a value of that type is always a tagged pointer to the object, never a thunk or indirection.

sgraf812:
Re: WW: Yes, I wrote the WW stuff when I was writing https://github.com/ghc-proposals/ghc-proposals/pull/257, which is basically succeeded by this proposal. At the time I was pretty convinced that we could always get rid of the extra indirection, which turns out not to be true (see the code block in https://github.com/ghc-proposals/ghc-proposals/pull/265#issuecomment-552865750). WW'ing for `Strict` doesn't make sense if we would have to introduce a box. Same for the `pack` example. I'll update accordingy.

Re: UnliftedRep, yes, I'm not sure what Andreas was alluding to. But this was before the discussion in [#17521](https://gitlab.haskell.org/ghc/ghc/issues/17521), so I think that comment is no longer relevant.

goldfirere:
* Given that we cannot have top-level unlifted bindings, are we allowed to have nullary constructors of unlifted datatypes?

* The proposal mentions kind `#`, which was gone as of GHC 8.0. I think you mean `TYPE UnliftedRep`. (Actually, all `RuntimeRep`s other than `LiftedRep` are call-by-value, but we have no common way to refer to the lot of them.)

* This potentially steals syntax, in case someone has a datatype context that begins with a type variable named `unlifted`. No one has ever done this, I'm sure. But there may be a small parsing hiccup.

* I'm a bit bothered by the fact that `-XUnliftedNewtypes` uses a kind annotation (optionally) to control the kind of the type, while this proposal uses a new keyword. I would prefer that this proposal uses a kind annotation `data UPair a b :: TYPE UnliftedRep where ...`. (Yes, I know this means that H98 syntax is not allowed. Although there's no reason we cannot have kind annotations on H98 syntax. I think.) We could provide a type synonym to avoid the scary `TYPE`: `data UPair a b :: UnliftedData where ...`.

* If we are going to have `data unlifted`, I'd prefer to also have `data lifted`. I'm frequently bothered by GHC's asymmetry between laziness and strictness and would sometimes prefer to have GHC require me to label everything as either lazy or strict (and thus force me to think about what's better). This isn't possible today. (For example, we can write `data MixedPair a b = MP ~a !b` only with `-XStrictData`, which sadly also changes the default.) More symmetry is better! But I'd still prefer using kind annotations.

andrewthad:
I'd like to push back a little bit on the fourth point Richard makes (the one about using a kind annotation instead of a keyword). Earlier in this thread, referring to using just a kind signature to change the levity of a data type, I [wrote](https://github.com/ghc-proposals/ghc-proposals/pull/265#issuecomment-552879495):

> ... this would establish a precedent that a kind signature can change the runtime behavior of a program. If that user had instead written ... the runtime behavior of their code would be different. CUSKs on data types can currently influence where or not the compiler accepts them, but they never change the operational semantics of a program. (I could be wrong about this though.)

Although UnliftedNewtypes uses optional kind annotations, the big difference is that, in that setting, there is only one possible kind for a newtype declaration, so adding the kind signature does not influence the behavior of a well-typed program. With UnliftedDatatypes, this isn't the case. If the levity of a data type can be specified by solely the kind signature, then that kind signature could change the behavior of a well-typed program. And the fact that that could happen bothers me (only a little bit).

sgraf812:
> - Given that we cannot have top-level unlifted bindings, are we allowed to have nullary constructors of unlifted datatypes?

Bummer! Indeed that is a problem for data con workers. Currently, if we define

```haskell
{-# LANGUAGE UnliftedDatatypes #-}
{-# LANGUAGE GADTs #-}

module Lib where

data unlifted T a where
  TInt :: T Int
```

We get the following data con wrapper (with the prototype implementation):

```
Lib.$WTInt [InlPrag=INLINE[0]] :: T Int
[GblId[DataConWrapper],
 Caf=NoCafRefs,
 Str=m,
 Unf=Unf{Src=InlineStable, TopLvl=True, Value=True, ConLike=True,
         Guidance=ALWAYS_IF(arity=0,unsat_ok=True,boring_ok=True)
         Tmpl= Lib.TInt @ Int @~ (<Int>_N :: Int GHC.Prim.~# Int)}]
Lib.$WTInt = Lib.TInt @ Int @~ (<Int>_N :: Int GHC.Prim.~# Int)
```

which is illegal with the current unlifted top-level binding policy. In the implementation it just seems to work, but it's hard to test because the wrapper will be inlined immediately. It could *potentially* break because the binding might not be properly tagged. My gut says this is highly unlikely because we can always see the constructor it evaluates to, but I'm not an authority on how pointer tagging for top-level data is implemented.

Also note that this is purely an implementation detail: Haskell-the-language doesn't have the concept of data con wrappers and thus doesn't need the very limited form of unlifted top-level bindings we have here.

> kind `#`

Oh, right. Will get rid of it.

> This potentially steals syntax, in case someone has a datatype context that begins with a type variable named unlifted. No one has ever done this, I'm sure. But there may be a small parsing hiccup.

I thought I already had written something in that regard, but apparently I only did so [in the implementation](https://gitlab.haskell.org/ghc/ghc/blob/e49102709ae17439e20578696fe8f8111f40b817/compiler/parser/Parser.y#L219-243). You are right, we're turning a well-formedness error into a parsing error. It's the same situation as with `instance`.

> I'm a bit bothered by the fact that -XUnliftedNewtypes uses a kind annotation (optionally) to control the kind of the type, while this proposal uses a new keyword.

That's a discussion I'm willing to have, but I'm leaning towards the keyword a little bit. It's also a lot less verbose.

> If we are going to have data unlifted, I'd prefer to also have data lifted.

Me too. I'll amend the proposal and wait for someone to contest.

goldfirere:
Annotations affecting runtime behavior has been with us from the beginning. If I write

```
module Main where
x = 9223372036854775807
y = x + 1
main = print y
```

and run, I see `9223372036854775808`.

But if I add

```hs
x :: Int
```

and run, I get `-9223372036854775808`. I do believe that (short of trickery involving type families) this may be the first time a *kind* annotation affects runtime behavior, but that doesn't seem like much of a stretch.

sgraf812:
One thing that's really weird about having the `unlifted` keyword is described in the proposal as an interaction with `-XUnliftedNewtypes`:

> -XUnliftedNewtypes introduces unlifted newtypes, but does so simply by inferring the kind of its single constructor's field type, no unlifted needed. Now with the new unlifted keyword, we could potentially allow syntax like newtype unlifted Foo (a :: TYPE r) = Foo a. What are its semantics? Can we still have Foo Int#? That wouldn't exactly be UnliftedRep (which this proposal is all about), but the unlifted, unboxed runtime-rep IntRep. Similarly, do we allow Foo Int? That would be boxed and lifted, seemingly contradicting the declaration.
>
> So I suggest that we (somewhat ironically) disallow unlifted syntax for Newtype delcarations and instead suggest to activate -XUnliftedNewtypes, which will automatically infer the generalised kind.

If we don't add the contextual keyword, we can just say that the return kind must evaluate to `TYPE 'UnliftedRep`. That may exclude H98 decls, but with `-XStandaloneKindSignatures` we can have the shorter syntax for H98 decls, too, right? That way we wouldn't even have to introduce new syntax.

Otherwise, I like the keyword for its terseness.

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,457 @@
...
+``TYPE 'UnliftedRep`` when there was a leading ``unlifted`` keyword.
+
+Data family instances may be declared unlifted, in which case the data family
+application's result kind must reduce to ``TYPE 'UnliftedRep``. See
+`the section on data families in the UnliftedNewtypes proposal <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0098-unlifted-newtypes.rst>`_
+and ``Note [Implementation of UnliftedNewtypes]`` for details involving
+type-checking the parent data family.
+
+The static semantics of other types of unlifted kind, such as the inability to
+delare them at the top-level, apply. The top-level restriction is not
```
aspiwack:
```suggestion
declare them at the top-level, apply. The top-level restriction is not
```

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,457 @@
...
+proposal seems rather easy to implement, with the majority of changes happening
+in the parser and type-checker. Notably the backend is not only affected **at
+all**. Very good cost to benefit ratio.
+
+As for the risk of making the language harder to learn: Beginners won't come in
+touch with unlifted datatypes at all. Unless they crave for better performance
+in a custom data structure, at which point I wouldn't consider them beginners
+anymore. There's precedent in going from unlifted to lifted by
+`Idris <http://docs.idris-lang.org/en/latest/tutorial/typesfuns.html>`_ with its
+``Lazy`` data type.
```
aspiwack: Or Ocaml's `Lazy.t`.

simonmar:
Putting the proposal into "needs revision" state while the outstanding issues that arose during the committee review phase are addressed.

proposals/0000-unlifted-datatypes.rst
sgraf812: Fixed in https://github.com/ghc-proposals/ghc-proposals/pull/265/commits/356566d78fb08e6703c1c4acb808ddfdf1119efa.

proposals/0000-unlifted-datatypes.rst
sgraf812: Fixed in https://github.com/ghc-proposals/ghc-proposals/pull/265/commits/356566d78fb08e6703c1c4acb808ddfdf1119efa.

sgraf812:
I have updated the proposal to look at the kind signature instead to determine levity.

The WIP implementation is out of sync now, but I think that can be fixed at a later point, I think.

simonmar:
Any other outstanding issues, or are we ready to accept this?

aspiwack:
@sgraf812 I'll be a bit of a bore, but as proposal need to be self-contained, may I ask you to add the `unlifted` keyword story in the _Alternatives_ section.

I like to phrase this sort of subsection as “a previous version of this proposal used […] but it was decided against for [such and such]”.

This way you can keep your possible grammar change for future documentation, but it is also documents the reason for the particular design choices without having to trawl through the Github conversation. Finally, as much as possible, the committee is supposed to take a decision based solely on the proposal document, it must therefore figure all discussed alternatives as accurately as possible.

simonpj:
I would love the proposal to talk about levity polymorphism.  That is

* Does it allow `data T l :: TYPE (BoxedRep l)  where ...`?

I suspect that the answer is "not in this proposal", but since the design has changed specifically to alow the possiblity of this future development, it would be good to discuss this future path explicitly.

Also, wny *not* support it now?

goldfirere:
A few more details from me:

* The proposal should be (slightly) rephrased to be compatible with #301/#203, where we now say `BoxedRep Lifted` instead of `LiftedRep`.

* The proposal should clarify that, even with `-XUnliftedDatatypes`, datatypes will be defaulted to have kind `TYPE (BoxedRep Lifted)`. Indeed, the *only* way to get an unlifted datatype is via a kind signature (standalone or otherwise).

* The sentence "The static semantics of other types of unlifted kind, such as the inability to declare them at the top-level, apply." seems to say the wrong thing. After beta-reduction, I interpret this to say "The inability to declare other types of unlifted kind at the top-level applies." I don't think this is what you mean.

* The specification does not say whether nullary constructors of an unlifted type are allowed. This isn't me nitpicking -- I don't know the answer! (At least, not from the specification section; I have not reread the entire proposal.)

* I second Simon's question about levity polymorphism. I think it should be easy to accommodate here.

*

sgraf812:
> Does it allow data T l :: TYPE (BoxedRep l) where ...?

I actually see no technical reason not to support it as part of this proposal. But it wouldn't be very useful without levity polymorphism in expressions. Well, apart from being able to re-use the data structure, which is quite a boon. Hmm... I'll include it in the proposal. I also added an example on data structure re-use. @rae (or someone else being able to make sense of it) could you check that I got the right intuition about visible kind application etc.? Do we "default" `Levity` to `Lifted`? :thinking: 

> nullary constructors

The proposal doesn't say anything about nullary constructors because they are allowed, just like regular constructors. I don't consider nullary constructors top-level bindings (should I?), hence there is no interaction with the top-level restriction. Under "Effects and Interactions" I recognise the interaction with nullary data-con wrappers, though, but I think that's just an implementation detail. We could instead choose to inline all data con wrappers unconditionally and wouldn't have that problem.

Anyways, I added a pointer to the static semantics section.

I also addressed the remaining points. (Maybe look at the commit diff for easier review).

klapaucius:
can unlifted data be allocated directly into compact region?

sgraf812:
> can unlifted data be allocated directly into compact region?

No. It has to be copied in the same manner as lifted data.

sgraf812:
@simonmar I think I resolved all requests for revision in my last commit, at least I hope so. Can we hand over to the committee again, please?

sgraf812:
@simonmar Could consider handing this back to the committee? I'm not currently aware of any open TODOs on my side.

simonmar:
@simonpj @goldfiere could you check that your comments are addressed now?

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,465 @@
...
+
+Henceforth, data type declaration refers to both data type and data family instance declarations.
+
+Static semantics
+~~~~~~~~~~~~~~~~
+
+Kind signatures for regular data type declarations must have a return kind of
+``TYPE (BoxedRep Lifted)``. Activating ``-XUnliftedDatatypes`` will lift this
+restriction to allow any return kind that unifies with ``TYPE (BoxedRep _)``. A
+data type constructor explitly *may* have levity polymorphic kind
```
goldfirere:
```suggestion
data type explitly *may* have levity polymorphic kind
```

I thought, for a sec, that you were talking about data constructors here.

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,465 @@
...
+``forall l. TYPE (BoxedRep l)``.
```
goldfirere:
```suggestion
``TYPE (BoxedRep l)`` for some type variable ``l``.
```

This allows for the possibility of type arguments.

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,465 @@
...
+
+Note that Haskell98-style data declarations can use top-level kind signatures
```
goldfirere:
```suggestion
Note that Haskell98-style data declarations can use standalone kind signatures
```

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,465 @@
...
+to specify the return kind. If the user provides no kind signature, the default
+remains that the data type has lifted return kind (``TYPE (BoxedRep Lifted)``).
+Specifying a kind signature is the only way to declare an unlifted data type.
```
goldfirere:
```suggestion
Specifying a kind signature is the only way to declare an unlifted or levity-polymorphic data type.
```

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,465 @@
...
+
+The same applies to data family instances, in which case the data family
+application's result kind must reduce to ``TYPE (BoxedRep _)``. See
+`the section on data families in the UnliftedNewtypes proposal <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0098-unlifted-newtypes.rst>`_
+and ``Note [Implementation of UnliftedNewtypes]`` for details involving
+type-checking the parent data family.
+
+As usual for types of unlifted kind, they may not be declared at the top-level.
```
goldfirere:
```suggestion
As usual for types of unlifted kind, values of unlifted data types may not be declared at the top-level.
```

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,465 @@
...
+The top-level restriction is not fundamental (see
+`#17521 <https://gitlab.haskell.org/ghc/ghc/issues/17521>`_), but best
+discussed in a separate proposal. There is an interesting interaction of this
+restriction with nullary data-con wrappers (cf. "Effects and Interactions"),
+which we consider to be an implementation detail.
```
goldfirere: I'm not convinced here. Whether or not nullary constructors are allowed is part of this specification.

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,465 @@
...
+Example:
+
+::
+
+ data UPair a b :: TYPE (BoxedRep Unlifted) where
+   UPair :: a -> b -> UPair a b
+
+* When occuring in a constructor field (e.g.
+  ``data T = MkT (UPair Int Bool)``), the semantics are identical to a field
+  with a bang pattern (``data T = MkT !(Int, Bool)``).
```
goldfirere: This isn't really a bang pattern -- it's not a pattern (and is unaffected by `-XBangPatterns`).

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,465 @@
...
+
+* In an application ``f (UPair a b)``, the argument ``UPair a b`` is evaluated
```
goldfirere: I think this is a poor example, because evaluating `UPair a b` does nothing. (Note that it evaluates neither `a` nor `b`.) Instead, maybe you want a function that returns something of type `UPair a b`.

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,465 @@
...
+  before the application is beta reduced. So call-by-value instead of
+  call-by-need.
+
+* In a let binding ``let x = UPair a b in e``, the right-hand side ``UPair a b``
```
goldfirere: Again, evaluating `UPair a b` does nothing here.

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,465 @@
...
+  is evaluated before the body.
+
+The Strict data type
+~~~~~~~~~~~~~~~~~~~~
+
+We get to define ``Strict``
```
goldfirere: Why does this have to be a `data` type, allocating an extra box? Why can't it be a newtype? I know that newtypes typically have the same kind as their argument, but it actually seems just fine to have a newtype of kind `BoxedRep Unlifted` wrap a representation of kind `BoxedRep Lifted`. The type system would make sure that the value of type `StrictNewtype` would be evaluated eagerly. Perhaps that's a separate proposal, but one worth considering next, I think.

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,465 @@
...
+-----------------------
+
+Introduction of user-defined unlifted data types means we can finally write
+code processing data types that can be compiled as if we were in a strict
+language.
+
+**Strict constructor fields** share considerable overlap with ``Strict``, yet they
+proved insufficient for encoding invariants for efficient code generation.
+
+This proposal consciously left out further work like a new specification for
```
goldfirere: Not true any more.

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,482 @@
...
+
+Henceforth, data type declaration refers to both data type and data family instance declarations.
+
+Static semantics
+~~~~~~~~~~~~~~~~
+
+Kind signatures for regular data type declarations must have a return kind of
+``TYPE (BoxedRep Lifted)``. Activating ``-XUnliftedDatatypes`` will lift this
+restriction to allow any return kind that unifies with ``TYPE (BoxedRep _)``. A
+data type explitly *may* have levity polymorphic kind
```
simonpj: explicitly

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,482 @@
...
+``TYPE (BoxedRep l)`` for some type variable ``l``.
```
simonpj: This desperately needs an example:
```
data PList (l :: Levity) a :: TYPE (BoxedRep l) where
   PNil :: PList l a
   PCons :: a -> PList l a -> PList l a
```

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,482 @@
...
+
+Note that Haskell98-style data declarations can use standalone kind signatures
+to specify the return kind. If the user provides no kind signature, the default
```
simonpj: For example
```
type SList :: Type -> TYPE (BoxedRep Unlifted)
data SList a = SNil | SCons a (SList a)
```

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,482 @@
...
+remains that the data type has lifted return kind (``TYPE (BoxedRep Lifted)``).
+Specifying a kind signature is the only way to declare an unlifted or levity-polymorphic data type.
+
+The same applies to data family instances, in which case the data family
+application's result kind must reduce to ``TYPE (BoxedRep _)``. See
```
simonpj: Give an example.

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,482 @@
...
+`the section on data families in the UnliftedNewtypes proposal <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0098-unlifted-newtypes.rst>`_
+and ``Note [Implementation of UnliftedNewtypes]`` for details involving
+type-checking the parent data family.
+
+As usual for types of unlifted kind, values of unlifted data types may not be
+declared at the top-level. The top-level restriction is not fundamental (see
+`#17521 <https://gitlab.haskell.org/ghc/ghc/issues/17521>`_), but best
+discussed in a separate proposal. Note that we don't consider nullary
+constructors to be top-level bindings in that regard, so nullary unlifted
+constructors are allowed. There *is* an interesting interaction of this
```
simonpj: Example
```
zs :: SList a
zs = SNil
```
I have no idea why you want to allow this.  It seems... odd.

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,482 @@
...
+restriction with nullary data constructor *wrappers* (cf. "Effects and
+Interactions"), though, which we consider to be an implementation detail.
```
simonpj: Deeply mysterious.   Omit?

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,482 @@
...
+~~~~~~~~~~~~~~~~~
+
+Unliftedness (i.e., the absence of divergence) in general implies the need for
+an eager evaluation semantics, which GHC implements in expressions of kind
+``TYPE r``, where ``r`` is not ``BoxedRep Lifted`` (so
+``TYPE (BoxedRep Unlifted)`` in particular).
+
+Thus, call-by-value semantics are already well established within GHC. The
+novelty is pattern matching on and construction of unlifted data types, but
+that's exactly the same as it is for lifted data types.
```
simonpj: I don't know what this paragraph means.  Give an example or two.

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,482 @@
...
+restriction with nullary data constructor *wrappers* (cf. "Effects and
+Interactions"), though, which we consider to be an implementation detail.
+
+Dynamic Semantics
+~~~~~~~~~~~~~~~~~
+
+Unliftedness (i.e., the absence of divergence) in general implies the need for
+an eager evaluation semantics, which GHC implements in expressions of kind
+``TYPE r``, where ``r`` is not ``BoxedRep Lifted`` (so
+``TYPE (BoxedRep Unlifted)`` in particular).
```
simonpj: Example
```
f x = let { y :: SList Int; y = g x
        in ...y...
```
Since the binding for `y` is unlifted, the let is evaluated eagerly, without building a thunk.,

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,482 @@
...
+As usual for types of unlifted kind, values of unlifted data types may not be
+declared at the top-level. The top-level restriction is not fundamental (see
+`#17521 <https://gitlab.haskell.org/ghc/ghc/issues/17521>`_), but best
+discussed in a separate proposal. Note that we don't consider nullary
+constructors to be top-level bindings in that regard, so nullary unlifted
+constructors are allowed. There *is* an interesting interaction of this
+restriction with nullary data constructor *wrappers* (cf. "Effects and
+Interactions"), though, which we consider to be an implementation detail.
+
+Dynamic Semantics
```
simonpj: I the TL;DR here is that the proposal makes no change to GHC's existing dynamic semantics.  Namely:
* Valuse of unlifted type are always computed eagerly.
* No let-binding or non-HNF argument can have a levity otherthan `Lifted` or `Unlifted`.

 Say so!

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,482 @@
...
+does the Right Thing, namely disallowing unevaluated boxes to be passed around
+by using call-by-value. This way we can have the guarantee that all unlifted
+boxes are always properly tagged.
+
+Preliminary `benchmarks <https://gitlab.haskell.org/ghc/ghc/merge_requests/2218#note_239994>`_
+suggest that the unlifted variant is 5-10% faster, just due to omission of the
+zero tag check.
+
+Proposed Change Specification
+-----------------------------
```
simonpj: Please say that this entire proposal is expressed under the assumption that [the pointer rep proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0203-pointer-rep.rst) and [its ammendment](https://github.com/ghc-proposals/ghc-proposals/pull/301) are both implemented.  Otherwise it's hard to grok.

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,482 @@
...
+
+Henceforth, data type declaration refers to both data type and data family instance declarations.
+
+Static semantics
+~~~~~~~~~~~~~~~~
+
+Kind signatures for regular data type declarations must have a return kind of
+``TYPE (BoxedRep Lifted)``. Activating ``-XUnliftedDatatypes`` will lift this
+restriction to allow any return kind that unifies with ``TYPE (BoxedRep _)``. A
```
simonpj: For example:
```
data List a :: TYPE (BoxedRep Lifted) where
   Nil :: List a
   Cons :: a -> List a -> List a

data SList a :: TYPE (BoxedRep Unlifted) where
   SNil :: SList a
   SCons :: a -> SList a -> SList a

File: proposals/0000-unlifted-datatypes.rst [hunk 26]
```diff
@@ -0,0 +1,482 @@
...
+
+``Strict`` is the very essence of this proposal: Every unlifted data type can
+be defined in terms of lifted data types and ``Strict``, at the cost of an
+additional indirection.
+
+Whether we could define ``Strict`` as a ``newtype`` -- which would then mean we
+coerce between kinds, meaning coercions aren't zero-cost -- is left for a
+separate proposal.
+
+Low-level code
```
simonpj: YOu give Haskell code, Core code, and C-- code for the bad version.  But, oddly, you do not give the Haskell code or Core code for the good version -- just the C-- code!   That makes it really hard to grok.

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,482 @@
...
+* For data family instances, this is only possible if the parent data family
+  was kind polymorphic. Plus it's a strange thing to do change kinds of a
+  declaration just by switching on a language extension.
+
+So rather dreadfully, we probably shouldn't "augment" ``-XStrict``.
+
+**Lazy pattern matches** ``let ~t = ... in ...`` should not be allowed if ``t``
+is unlifted. That is exactly the behavior that is currently implemented in GHC.
+
+**Data con wrappers** of nullary constructors like
```
simonpj: This stuff about data con wrappers is *only* about GHC's implementation, nothing to do with the specification.  Please say so.   Examples woudl be useful.

simonpj:
> @simonpj @goldfiere could you check that your comments are addressed now?

I have submitted a raft of drafting suggestions, all designed to make the proposal easier to understand.

I'm content with the technical content.  So yes, once these edits are made to Simon M's satisfaction, I'm happy to approve.

proposals/0000-unlifted-datatypes.rst
sgraf812: I thought I'd leave examples to the Examples section, where I give

```
  data List a :: TYPE (BoxedRep l) where
    Nil :: List a
    Cons :: a -> List a -> List a

  mapLifted :: (a -> b) -> List a -> List b
  mapLifted f Nil         = Nil
  mapLifted f (Cons x xs) = Cons (f x) (mapLifted f xs)
  
  mapUnlifted :: (a -> b) -> List @Unlifted a -> List @Unlifted b
  mapUnlifted f Nil         = Nil
  mapUnlifted f (Cons x xs) = Cons (f x) (mapLifted f xs)
 ```

I can give that example earlier, but my goal was to keep the Change Specification as concise as possible.

File: proposals/0000-unlifted-datatypes.rst (hunk 17 above)
Comment (sgraf812) [reply proposals/0000-unlifted-datatypes.rst]: I'll add an example to the "data structure re-use" example.

File: proposals/0000-unlifted-datatypes.rst (hunk 16 above)
Comment (simonpj) [reply proposals/0000-unlifted-datatypes.rst]: I understand that goal, but feedback is that the Change Specification section is nigh incomprehensible without the examples.     The examples can be carefully chosen to illustrate the particular point being made by each sentence.   More distant examples just don't do that.

(In another world, you could imagine a change spec that included formal BNF and typing rules.  But even then examples are helpful.)

File: proposals/0000-unlifted-datatypes.rst (hunk 18 above)
Comment (sgraf812) [reply proposals/0000-unlifted-datatypes.rst]: Added the following example:

```hs
data family DF a :: TYPE (BoxedRep l)
data instance DF Int :: TYPE (BoxedRep UnliftedRep) where
  TInt :: Int -> DF Int -- unlifted!
data instance DF Char :: TYPE (BoxedRep LiftedRep) where
  TChar :: Char -> DF Char -- lifted!
```

proposals/0000-unlifted-datatypes.rst
sgraf812: That's exactly what is not allowed. I wrote

>  we don't consider nullary constructors to be top-level bindings in that regard, so nullary unlifted constructors are allowed

to mean that nullary constructors aren't affected by the top-level restriction, so defining a nullary unlifted datacon is just fine. I'll change that to

> we don't consider nullary constructors to be top-level bindings in that regard, so *defining* nullary unlifted constructors *is* allowed

Happy to use different words here if that's still confusing, but I don't know which.

Or maybe just get rid of the section beginning at "Note that we don't"?

proposals/0000-unlifted-datatypes.rst
sgraf812: > No let-binding or non-HNF argument can have a levity otherthan Lifted or Unlifted.

Are you referring to the let/app invariant? If so, doesn't it also apply to Unlifted things? I'm confused.

proposals/0000-unlifted-datatypes.rst
sgraf812: Yes, that seems like the simplest way forward.

proposals/0000-unlifted-datatypes.rst
simonpj: > we don't consider nullary constructors to be top-level bindings in that regard, so defining nullary unlifted constructors is allowed

Oh I totally misunderstood.   You mean that
```
data T :: TYPE (BoxedRep Unlifted) where
  T1 :: T
```
is allowed.  Yes, indeed.  It never occurred to me that you might consider that that to be a "top level binding".   I thought you meant that `t1 = T1` is allowed!  But you don't.   

I think you can just say "As usual for types of unlifted kind, values of unlifted data types may not be declared at the top-level."; then give an example of what is and is not allowed (I offered a suggestion); then your following sentence "The top-level restriction is not fundamental (see #17521), but best discussed in a separate proposal."  Then stop

proposals/0000-unlifted-datatypes.rst
sgraf812: I don't know what I meant here, either. I mean, it's exactly the same, but then this paragraph is just unnecessary. I'll just delete it.

proposals/0000-unlifted-datatypes.rst
simonpj: I just mean that
```
let (x :: ty :: TYPE r) = e in ...
```
is illegal unless `r` = AddrRep or IntRep or (BoxedRep Lifted) or (BoxedRep Unlifted).  It can't have variables or type functions.

proposals/0000-unlifted-datatypes.rst
sgraf812: Well, the diff in Core isn't particularly interesting (just substituting `Bool` for a new `SBool`). But I'll add it.

proposals/0000-unlifted-datatypes.rst
sgraf812: Added an example and clarified that this is about the implementation.

proposals/0000-unlifted-datatypes.rst
sgraf812: Hmm... What about `DoubleRep`, etc.? I think you could say that for every concrete rep. So it's rather that they must have kind `TYPE r` for a known/concrete runtime-rep `r`. Also I feel like we are repeating the runtime rep polymorphism restrictions here, which feels a little out of place.

In fact we don't touch dynamic semantics at all, we just give data types a different kind and exploit that GHC already knows what to do.

proposals/0000-unlifted-datatypes.rst
sgraf812: Done.

proposals/0000-unlifted-datatypes.rst
sgraf812: I rewrote the section. How about

```
There is no change in GHC's existing dynamic semantics, namely

- Values of unlifted type are always computed eagerly
- Only values of types with a concrete ``RuntimeRep`` can be let-bound

Example:

::

  f x = let y = if odd 42 then SNil else SCons 42 SNil
        in ... y ...

Since the binding for ``y`` is unlifted, the ``let`` binding (is legal and) is
evaluated eagerly, without building a thunk.

This proposal simply allows data declarations to have kinds other than
``TYPE LiftedRep`` and the existing dynamic semantics of GHC takes care of the
rest.

```

proposals/0000-unlifted-datatypes.rst
sgraf812: Done

proposals/0000-unlifted-datatypes.rst
simonpj: So much better!

proposals/0000-unlifted-datatypes.rst
```diff
@@ -0,0 +1,590 @@
...
+  -- alternative using a SAKS:
+  -- type List :: Type -> TYPE (BoxedRep l)
+
+  mapLifted :: (a -> b) -> List a -> List b
+  mapLifted f Nil         = Nil
+  mapLifted f (Cons x xs) = Cons (f x) (mapLifted f xs)
+  
+  mapUnlifted :: (a -> b) -> List @Unlifted a -> List @Unlifted b
+  mapUnlifted f Nil         = Nil
+  mapUnlifted f (Cons x xs) = Cons (f x) (mapLifted f xs)
```
hsyl20: Typo: should be `mapUnlifted`

simonmar:
Great, I think we can now accept this.

noughtmare:
I hope this is the right place to ask this question. 

On the [GHC wiki page about unlifted data types](https://gitlab.haskell.org/ghc/ghc/-/wikis/unlifted-data-types) the motivation mentions that unlifted data types allows us to program with types that don't have the pesky bottom inhabitant. That should make it easier to reason about unlifted data types and write API's that cannot be abused. 

Personally, I have been looking at [the source code of the `stream-fusion` package](http://hackage.haskell.org/package/stream-fusion-0.1.2.5/docs/src/Data-Stream.html). I am fascinated by the use of a custom [`Unlifted` type class](http://hackage.haskell.org/package/stream-fusion-0.1.2.5/docs/src/Data-Stream.html#Unlifted) to make sure that no unwanted memory leaks are introduced in the streams. I am hoping that this new unlifted data types proposal would make it possible to use these unlifted representations instead of that `Unlifted` class to avoid memory leaks. And I think that this will also improve the performance because you won't need to use `expose` anymore.

In this proposal the motivation section only mentions a slight performance improvement because we can skip checks for unevaluated thunks. Which makes me wonder: does this proposal still give us the reasoning benefits and the ability to avoid unintended memory leaks? Would we be able to write polymorphic functions over unlifted types like in the second part of the proposal on [the GHC wiki](https://gitlab.haskell.org/ghc/ghc/-/wikis/unlifted-data-types#proposal-2-polymorphism-over-a-new-unlifted-kind), e.g.:

```Haskell
data UList (a :: UnliftedType)
  = UNil | UCons a UList

umap :: forall (a :: UnliftedType) (b :: UnliftedType).
        UList a -> (a -> b) -> UList b
umap UNil f = UNil
umap (UCons x xs) = UCons (f x) (umap xs f)
```

(I guess you can already write something like that in GHC today, but there are not many interesting unlifted types)

andrewthad:
> Does this proposal still give us the reasoning benefits and the ability to avoid unintended memory leaks?

Absolutely. Any time you see a type of kind `UnliftedType`, it's guaranteed to be in WHNF. Whether or not this is also NF (normal form) depends on whether or not the data has any lazy fields in it. To go back to your example, there are actually two ways to write `UList`. In the one you wrote, the list is strict in the elements but still potentially lazy in the spine. You may instead want:

    data UList :: UnliftedType -> UnliftedType where
      UCons :: a -> UList a -> UList a
      UNil :: UList a

Which is strict in both the elements and the spine.

Also, Ben Gamari and Sebastian Graf landed the (related) BoxedRep MR in GHC HEAD recently. Sebastian has been doing a bit of work on the Unlifted Datatypes proposal recently.

sgraf812:
> Does this proposal still give us the reasoning benefits and the ability to avoid unintended memory leaks?

As Andrew said, yes. The example in the motivation section describes a scenario in which it is *impossible* to achieve maximum performance today, even with strict fields present and very careful performance optimisation. `UnliftedType` achieves a speedup there that wouldn't be possible without ([well... so I thought](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/2218#note_337457)). If anything, `UnliftedType` means you don't have to worry about evaluatedness anymore, as evaluatedness is an invariant encoded in the kind and maintained by the compiler.

Ultimately, if every data type was levity-polymorphic, you could probably define `expose` (or `seq`) to go from the `Lifted` form of a data type to its `Unlifted` form. That will not be possible for the foreseeable future, though.

aspiwack:
I second the choice of (1)+(2).

simonpj:
I have just realised that in this accepted, and implemented proposal we have done something entirely new.  Consider this (from https://gitlab.haskell.org/ghc/ghc/-/issues/20204)
```
type IntU ::  Bool -> Wombat

...LOTS OF CODE...

data IntU a b = IntU Int

...MORE CODE...

type Wombat =Type -> TYPE UnliftedRep
```
The kind signature for `IntU` *completely changes the semantics of the `data IntU` declaration*, and yet can be separate from it.  Even if it is adjacent, the fact that it's unlifted is quite subtle... you have to look to the right of the arrows, and then through the distant (and perhaps imported) type synonym `Wombat`.

I'm not very happy with a distant kind signature having such a profound effect on the semantics of the data type.  Indeed in [my comment above](https://github.com/ghc-proposals/ghc-proposals/pull/265#issuecomment-525705557) I suggested a keyword
```
data unlifted IntU a b = IntU Int
```
to signal that it's an *unlifted* data type.   But then I went AWOL and didn't pursue the matter.  I don't know why I was so negligent.

So this post is to ask: does anyone else think this is bizarre?   I'm inclined to make a proposal to add the keyword, but I thought I'd test the waters first.

sgraf812:
See the 5ish comments staying at https://github.com/ghc-proposals/ghc-proposals/pull/265#issuecomment-560375954 for why we decided to omit the keyword.

JakobBruenker:
Instead of adding a keyword, this could be another case where modifiers could come in handy? As in,

```haskell
-- Existing type already exposed via GHC.Exts:
-- | Whether a boxed type is lifted or unlifted.
data Levity = Lifted | Unlifted
 ```
 
 ```haskell
%Unlifted
data SMaybe a = SJust !a | SNothing
```

noughtmare:
> The kind signature for `IntU` completely changes the semantics of the `data IntU` declaration, and yet can be separate from it.

Don't we already have similar behavior with normal functions and type classes:

```haskell
x :: Double
-- x :: Int

-- lots of code

x = 1

main = print x
```

Will print `1` or `1.0` depending on the distant type signature. In this case the semantic change is very small, but more complicated functions and classes can change the semantics entirely.

I don't think this is very strange.

goldfirere:
I'm sympathetic to Simon's concern here, but I'm not as bothered: I don't feel the need to include a new keyword.

On the other hand, I'd be happy with a modifier (thanks for the idea @JakobBruenker), along with a `-Wall` warning that bleated whenever the modifier is absent for a non-lifted type. (Non-lifted = not necessarily lifted; unlifted = unlifted. That is, non-lifted includes levity-polymorphic.)

sgraf812:
I might also add that as tooling gets better, the *textual* data type definition becomes less and less relevant while understanding code. Heck, even if I hover over a type like `CoreExpr` in my editor *today*, I get the tooltip

```
> CoreExpr :: * 

 *Defined in ‘GHC.Core’*
```

Note the `*`. For `IntU`, I'd get `Bool -> Wombat`, as expected. It's no longer the job of the programmer to find and put signatures and definitions together, you can just let the tool do it. 

I'm not against a keyword or modifier if people find the current behavior confusing, but I still want to add that as soon as you want to define a levity-polymorphic data type, you are stuck with a separate signature again.

goldfirere:
I agree with @sgraf812's observation. More generally: as our language gets more intricate, our tools need to keep apace. This is why e.g. Agda and Idris are out ahead of Haskell with tooling. Those languages are more intricate, and so the tooling must be up to the task. For a long time, Haskell has limped along with sub-par tooling, because, well, I don't know why. As the tooling catches up, though, we can new flexibility in design -- for example, by making things less obvious in the code text if they can be clarified by tools.

@simonpj: you prodded this conversation to start. How do you feel about this all now?

simonpj:
Ugh.   When I started this conversation had forgotten about
```
type T :: forall (l :: Levity) -> TYPE (BoxedRep l)
data T l = MkT Int
```
Now `(T Lifted)` is lifted, while `(T Unlifted)` is unlifted; so clearly we can't put a fixed keyword on the data type declaration.

That does leave me a bit stuck.  A keyword isn't going to work.

The only viable alterantive is to insist that if the return kind isn't `Type` then you must declare it *in the declaration*, thus
```
data T l :: TYPE (BoxedRep l) where
  MkS :: Int -> T l
```
I forget -- was that possibility discussed?   (I don't think we have a way to declare a return kind in H98 syntax, do we?)

I'm still clinging to the idea that we should be able to infer a kind from a declaration; a signature should make it less polymorphic, but not more. (And yes I know that polymorphic recursion breaks this, but not badly.)

Any other views?

sgraf812:
> I don't think we have a way to declare a return kind in H98 syntax, do we?

That is exactly the problem. I find H98 quite nice for declaring simple data types such as `SMaybe`, but YMMV. @RyanGlScott and I discussed it very briefly in https://gitlab.haskell.org/ghc/ghc/-/merge_requests/2218#note_335773, but the proposal already included SAKS at the time, probably from a decision I made [here](https://github.com/ghc-proposals/ghc-proposals/pull/265#issuecomment-562042372).

Another gripe I'd have with it is that `-XUnliftedNewtypes` works with SAKS, so why wouldn't `-XUnliftedDatatypes`?

Again, I'm not strongly against disallowing SAKS, but it seems awfully surprising not to allow them from a design perspective. *Especially* given that they work for all lifted data types. Also to me separating the SAKS from the data type declaration is a code smell to begin with (wasn't syntactic distance the reason you gave in https://github.com/ghc-proposals/ghc-proposals/pull/265#issuecomment-896626642?). Arguably, the mix up in https://gitlab.haskell.org/ghc/ghc/-/issues/20204 could have happened just as well for a CUSK. I think our eyes merely aren't trained to give SAKS the attention they deserve yet.

simonpj:
I am absolutely not suggesting that we disallow a standalone kind signature!  I'm only suggesting that you can infer the kind from the declaration.  It may be narrowed by the SAKS.

With unlifted newtypes you *can* infer the signature, right?
```
newtype N = MkN Int#
```
I can see that is unlifted without looking at its signature.

noughtmare:
> With unlifted newtypes you can infer the signature, right?

You can write this:

```haskell
type T :: TYPE (BoxedRep r) -> TYPE (BoxedRep r)
newtype T a = T a
```

simonpj:
Ah yes.  So in my straw-man idea, that would be illegal because looking at the newtype decl alone we'd default to lifted.  If you want something more general you'd have to decorate the newtype decl itself.

But maybe I am a lone voice.

goldfirere:
> But maybe I am a lone voice.

Perhaps. But you've been at this for some time and have had a modicum of success, so it's a voice we should listen to.

What about my modifier story?

```hs
type N :: TYPE (BoxedRep l) -> TYPE (BoxedRep l)
%HasUnusualKind
newtype N a = MkN a
```

The `%HasUnusualKind` modifier is optional, but we could create a warning that would bleat if you leave it out. It's a marker to the reader to look for the SAKS. (Ditto it all for datatypes.) `%HasUnusualKind` requires the presence of a SAKS.

My own preference is not to have `%HasUnusualKind` and just to say that the type signature is part of the declaration, but I think the modifier approach could be a way forward toward compromise.

simonpj:
But why would you invent new modifiers rather than use what we already have: kind signatures?

JakobBruenker:
Feels weird to introduce something that goes against the general trend of moving away from CUSKs and towards SAKs, no?

aspiwack:
> I'm not arguing for a _complete_ user supplied kind (CUSK). The CUSK story was principally to suport polymorphic recursion ([user manual entry](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/poly_kinds.html?highlight=cusk#extension-CUSKs)). That's not at all waht I am talking about here.
> 
> I'm only talking about defaulting of the return kind of a data type: to be other than `Type`, you need an explicit kind signature.

Do we agree, though, that it's already what the proposal says?

AntC2:
> @tomjaguarpaw FWIW I'm very surprised to learn that type signatures can be separated from definitions like this

> @gridaphobe [on the Steering Committee discussion last month] I think the distance between the `type IntU` and `data IntU` is the crux of the issue. If we were forced to keep the signature and declaration together, ...
> I've never been particularly fond of Haskell's tolerance for separating signatures and definitions. 

I agree, and it afflicts `-XScopedTypeVariables` too: the significance of the explicit `forall` is not something you'd pay attention to if the decl is a long way from the binding.

We need a syntactic way to group [**] a type(-of-type) decl with its binding (and other stuff). Some ideas that don't clash with current syntax:

    let type IntU ::  Bool -> Wombat                          -- top-level `let`
        type Wombat =Type -> TYPE UnliftedRep
     in data IntU a b = IntU Int

       type IntU ::  Bool -> Wombat                           -- these must be indented deeper than the `in`
       type Wombat =Type -> TYPE UnliftedRep
    in data IntU a b = IntU Int

    { type IntU ::  Bool -> Wombat
      type Wombat =Type -> TYPE UnliftedRep
      data IntU a b = IntU Int
    }

    type IntU ::  Bool -> Wombat
    ;; type Wombat =Type -> TYPE UnliftedRep
    ;; data IntU a b = IntU Int

(The last is stolen from BCPL's 'very binding semicolon'. It actually seems to be currently valid Haskell syntax, with apparently the same meaning as a single semicolon in the same position/allowing the following decl to be indented as far as you like; or same as no semicolon at all without indenting.)

The semantics is to be as currently/proposed here if these were each stand-alone decls. (So `Wombat` gets top-level scope.) Except that if we supported something like this for `-XScopedTypeVariables`, the explicit `forall` would be optional.

For this proposal, if the three decls were given separately stand-alone: infer the usual type (usual liftedness) for `IntU` from the `data` decl alone (possibly with an embedded/inline signature); later discover the `type IntU ...; type Wombat ...`; _check_ `IntU`'s inferred type against the decl; reject if not consistent.

[**] another place we could use grouping is for `pattern` sigs with their binding -- that would avoid the need to repeat the `pattern` keyword, and help signal that pattern sigs are bizarre, as compared with function sigs.

mrgutkun:
I believe this should have an Implemented label.

(Should there be an explicit policy that random passers-by may ping someone to add missing labels?

nomeata:
Thanks! You guessed the policy correctly; where should it be made explicit so that you would have seen it? Feel free to propose a PR

noughtmare:
@sgraf812 I just found out that the motivation in this proposal is not that strong, because you can get the same speedup by simply making the recursive fields of `IntSet` strict too:

```haskell
data IntSet
  = Branch !IntSet !Int !IntSet
  | Leaf
```

Perhaps the motivation should focus more on avoiding the need to put bang patterns everywhere and that we now have bona fide algebraic data types.

AndreasPK:
> @sgraf812 I just found out that the motivation in this proposal is not that strong, because you can get the same speedup by simply making the recursive fields of `IntSet` strict too:
> 
> ```haskell
> data IntSet
>   = Branch !IntSet !Int !IntSet
>   | Leaf
> ```
> 
> Perhaps the motivation should focus more on avoiding the need to put bang patterns everywhere and that we now have bona fide algebraic data types.

I believe this is mostly a result of tag inference, something that hadn't been merged yet when this proposal was written.

There are also some edge cases where tag inference doesn't help (things involving boot files, functions not being W/Wed and others) so I think the spirit of the motivation still holds. Even if the exact example isn't as striking anymore.

PR #266, opened 2019-08-28
WIP: Don't speak of `DynFlags`

Ericson2314:
``DynFlags`` hampers the modularity of GHC.
Instead of referring to it directly, refer to a type parameter constrained to have just the information necessary.
The benefits of this are somewhat uncertain, so merge the refactor on an experimental basis, and revert it if they aren't realized.

[Rendered](https://github.com/Ericson2314/ghc-proposals/blob/dynflags-bad/proposals/0000-dynflags-bad.rst)
WIP MR: https://gitlab.haskell.org/ghc/ghc/merge_requests/1033

CC @wz1000 @bgamari

### DIFF
--- a/proposals/0000-dynflags-bad.rst
+++ b/proposals/0000-dynflags-bad.rst
@@ -1,0 +1,106 @@
+Notes on reStructuredText - delete this section before submitting
+==================================================================
+
+The proposals are submitted in reStructuredText format.  To get inline code, enclose text in double backticks, ``like this``.  To get block code, use a double colon and indent by at least one space
+
+::
+
+ like this
+ and
+
+ this too
+
+To get hyperlinks, use backticks, angle brackets, and an underscore `like this <http://www.haskell.org/>`_.
+
+
+Proposal title
+==============
+
+.. author:: Your name
+.. date-accepted:: Leave blank. This will be filled in when the proposal is accepted.
+.. proposal-number:: Leave blank. This will be filled in when the proposal is
+                     accepted.
+.. ticket-url:: Leave blank. This will eventually be filled with the
+                ticket URL which will track the progress of the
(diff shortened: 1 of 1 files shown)


proposals/0000-dynflags-bad.rst
```diff
@@ -0,0 +1,129 @@
...
+
+Motivation
+----------
+
+``DynFlags`` has no meaning.
+It is just a roll up of every configuration option anything needs ever.
+This is bad for modularity: it leaks information to components of the compiler that they shouldn't care about, and prevents adding additional components with more configuration options.
+
+The easiest way to replace it today is to use type variables and ``Has*`` classes to project smaller records.
+[Perhaps in ea future with row types we would do something different, but we can refactor to do that then.]
```
Anrock: > ea future

`near`?

Ericson2314:
@bgamari Ah a bunch of concrete types. Sorry if I misunderstood, and that is what you meant all along when we talked in person.

A few thoughts:

1. That does deliver all the bulk benefits, unlike one big record. 

1. It definitely makes the non-inlining case faster. But it could be slower in the fully inlined case is one projects from `DynFlags` a bunch intermediate records, vs inlining the instance for DynFlags so all that's left is loading the wanted field from DynFlags itself with no allocations.

>  I think we should start by enumerating the bits of DynFlags that the pretty-printer actually uses.

Hopefully I'll get at that by finishing the refactor, even if we don't go with that exact design.

nomeata:
> I suspect this is out of scope of the proposals process since it is a change to GHC's internals. That being said, I'm happy to discuss it here.

I agree, but be our guest using this space.

Ericson2314:
Thanks to @hsyl20 a lot of this has already happened, and continues to happen.

PR #129, opened 2018-04-30
TypeApplications for Overloaded Literals 

isovector:
[Rendered](https://github.com/isovector/ghc-proposals/blob/type-apps/proposals/0000-type-apply-literals.rst)

[Reddit discussion](https://www.reddit.com/r/haskell/comments/8g083q/ghc_proposal_typeapplications_for_overloaded/)

Currently, types are unable to be applied to overloaded literals via `TypeApplications`. This proposal slightly changes the behavior of applied types in the presence of overloaded literals allowing us to do so.

PR #267, opened 2019-08-31
Class method headers (reordering class variables)

int-index:
We propose to allow class variable binders on the left of `::` in class method signatures:

```
class C a b where
   f @b @a :: (a, b) -> Bool

-- f :: forall b a. C a b => (a, b) -> Bool
```

The ordering of these variables may differ from their ordering in the class declaration header.

The syntax mirrors that of associated type headers and [Type variable binders](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0155-type-lambda.rst).


RyanGlScott:
I generally like this proposal, but there's one stipulation in the proposal that I find puzzling:

> * No other variables can be bound in the class method header.

The reason this puzzles me is that it limits the amount of type variable reordering you can do. I think this is better explained with an example, so here's one:

```hs
class C a b where
  m :: c -> a -> b
```

Under this proposal, I believe there are two possible type variable orderings:

1. `m :: forall a b. C a b => forall c. c -> a -> b`. This is the default ordering, as well as the ordering that you would obtain if you typed out `m @a @b :: a -> b -> c` using `ClassMethodHeaders`.
2. `m :: forall b a. C a b => forall c. c -> a -> b`. This is the ordering you would obtain if you typed out `m @b @a :: c -> a -> b` using `ClassMethodHeaders`.

However, there are other orderings that may be desirable, such as this one:

3. `m :: forall c a b. C a b => c -> a -> b`.¹

As far as I can tell, it is impossible to obtain this ordering using `ClassMethodHeaders`. One might envision being able to type `m @c @a @b` to achieve this, but the "No other variables can be bound in the class method header" expressly forbids this.

Another reason I find this restriction odd is because it is not a restriction that applies to associated type families. That is to say, you can certainly write this:

```hs
class C a b where
  type M (x :: c) (y :: a) :: b
```

And here the kind of `M` will be `forall c a b. c -> a -> b`. (Note that `c` is not bound in the class declaration header.) If the goal is to eventually unify the treatment of class methods and associated types, wouldn't it make sense to permit this sort of thing in class methods as well?

-----

¹ There are three other permutations of `a`, `b`, and `c` where `c` does not appear last in the `forall` telescope, but I've omitted them for the sake of brevity.

int-index:
@RyanGlScott The proposal is deliberately conservative. There are a few things that headers of associated types can do, but the proposed class method headers cannot:

1. Binding non-class variables (as you point out):

    ```
    class C a where
      type F a b  :: blah -- ok
      f a b :: blah  -- error: 'b' is not a class variable
    ```

2. Not binding all of class variables:

    ```
    class C a b where
      type F a :: blah -- ok
      f a :: blah  -- error: 'a' is not bound
    ```

3. Adding inline kind annotations on binders:

    ```
    class C a where
      type F (a :: k) :: blah  -- ok
      f (a :: k) :: blah  -- syntax error
    ```

The proposal as written does not exclude the possibility of giving reasonable semantics to these forms in the future. However, I'm not sure if we should follow the footsteps and allow all of these in terms, or if we should deprecate them in types.

Lifting these rules would have consequences that I have not fully evaluated:

1. Binding non-class variables would violates the principle that the part of the signature to the right of `::` defines the field type in the dictionary. Consider:
   
   ```
   class C a b where
     f :: fsig
     g :: gsig
   ```

   We can rewrite this as a record type:

   ```
   data CDict a b = MkCDict
     { f :: fsig,
       g :: gsig }
   ```

    Note that the field types `fsig` and `gsig` correspond exactly to the class method types. Reordering or changing the visibility of class variables maintains this property, while adding new variables does not.

2. Not binding all of class variables would be akin to a violation of the "forall-or-nothing" rule. We can say that unmentioned class variables are bound implicitly, but then we have to say where exactly: before or after the explicitly bound variables?

    ```
    class C a b where
      f @b :: blah

    -- is it   f :: forall a b. blah  ?
    --    or   f :: forall b a. blah  ?
    ```

    Or do we, perhaps, interleave these variables in a way that would make the signature well-scoped?

    ```
    class C (a :: k) (b :: j) where
      f @a @j :: blah

    -- f :: forall k (a :: k) j (b :: j) :: blah
    ```

    This is an intricate design question which we do not have to answer now (or, perhaps, at all).

3. Adding inline kind annotations on binders is plausible, but it would complicate the specification (and the implementation) with details about pattern-like meta-variables in these kind annotations. Basically, this is left as future work.

I understand that these rules outlaw some variable orderings that may be desirable, but I'd rather be cautious lest we discover that these orderings aren't such a good idea (as elaborated above).

RyanGlScott:
That explanation makes sense to me, thanks. Do you mind adding this discussion to the Alternatives/Unresolved Questions section? I'm not sure which section would be most appropriate, but it definitely deserves to live _somewhere_ in the proposal.

int-index:
@RyanGlScott Good idea. I added the answer to a new section, "Limitations".

Ericson2314:
I sort want something like NoMethodSelectors and a way to get the dictionary of a class (but not vice versa!). Then one can write their own wrapper. Of course this is a big feature which would take a lot of planning, but it's a single big feature for maximal expressive power.

isovector:
At risk of derailing the discussion, for which I apologize greatly in advance --- while we're fiddling with these levers, can we somehow get a solution for type-applying overloaded literals out of this?

int-index:
I don't believe the proposal in its current form would allow type-applying overloaded literals. This is because it does not change the fact that _class variables come first_:

```
fromInteger :: forall a. Num a => Integer -> a
fromString :: forall a. IsString a => String -> a
```

Here, `forall a.` is at the front of the type. Would it be a good idea to relax this restriction? I'm far from sure. Currently, the part of the signature to the right of `::` defines the type of the method as stored in the dictionary. Consider:

```
class Num a where
  fromInteger :: Integer -> a

class IsString a where
  fromString :: String -> a
```

We can desugar to

```
data NumDict a = MkNumDict
  { fromInteger :: Integer -> a }

data IsStringDict a = MkIsStringDict
  { fromString :: String -> a }
```

However, for type-applying literals we would want these types:

```
fromInteger :: Integer -> forall a. Num a => a
fromString :: String -> forall a. IsString a => a
```

But now, can you tell from these signatures what the method dictionaries store? It's no longer a straightforward syntactic transformation! I'm very uncomfortable with this.

In fact, #129 looks like the right way to do achieve what you're asking for (and now I've noticed that it was you who proposed it). I'm not happy that the committee rejected it.

JakobBruenker:
So, to compare it to https://github.com/ghc-proposals/ghc-proposals/pull/148, the main advantage seems to be that it mirrors the syntax for parameters of associated types, whereas the main disadvantage seems to be that it only allows reordering within the class variables. (And thus for example not allowing type-applying overloaded literals.) Is that correct?

As we move towards unifying associated types and methods, beyond retaining similarity to the current type parameter syntax, is there a reason to prefer one of these syntaxes (toplevel or headers) over the other?

int-index:
> the main advantage seems to be that it mirrors the syntax for parameters of associated types

Yes.

> the main disadvantage seems to be that it only allows reordering within the class variables. (And thus for example not allowing type-applying overloaded literals.)

Yes, but I don't consider it a disadvantage, it's a conscious design decision. The reasons are elaborated in the "Limitations" section and https://github.com/ghc-proposals/ghc-proposals/pull/267#issuecomment-526949512. We could extend the proposed syntax further and allow something like

```
class Num a where
  fromInteger (x :: Integer) @a :: a
```

and that would result in

```
fromInteger :: Integer -> forall a. Num a => a
```

The proposal is forward-compatible with such an extension. In fact, this would bring us even closer to what associated types allow today. I'm just not sure if it's a good idea to allow non-class variables in the header. The thing that bothers me:

* How do we figure out what to store in the method dictionary from such a declaration?

Is there an algorithm, or at least a declarative specification of thereof? This problem is not specific to the choice of syntax, it affects both #148 and this proposal.

----

> As we move towards unifying associated types and methods, beyond retaining similarity to the current type parameter syntax, is there a reason to prefer one of these syntaxes (toplevel or headers) over the other?

Hm, good question. Beyond matters of taste, #148 is also more complex, introducing a type equivalence check based on mutual subtyping (see its "Proposed Change Specification"). Here, the semantics are defined by simple reordering of parameters.

JakobBruenker:
One thing that bothers me a little bit is that it seems to mirror the syntax of type equations rather than of kind signatures, but since type family headers also seem to have inherited their syntax from type equations, it might be too late to get upset about that. Though presumably (and if I'm not misunderstanding anything), once unsaturated type families are allowed, `F a :: b` will be equivalent to something like `F :: a -> b`, taking away a bit from the justification of mirroring type family header syntax.

Aesthetically it seems a bit strange to me to write type variables on the left of the `::` when there's already a perfectly usable type signature on the right of the `::` that could be extended to include this, but I'm also not *really* happy by having to use use a keyword like `toplevel` to enable this extension, so I'm not really sure what my ideal solution would look like here.

simonpj:
* I like the conservative-ness of this proposal: it adds a bit of complexity to get a bit of expressiveness.

* I like that the bit after the `::` continues to express the type of the field (as the proposal says).

> Class method definitions in instance declarations may explicitly write out class variable instantiations

* Presumably those "class method header" arguments must appear first (unlike associated types), and identically to the instance header (like associated types)?   You need to say this; and give the new syntax for instance decls.

So it looks feasible to me, but I'm a bit unconvinced about the cost/benefit ratio.  Is the lack of this causing Real Pain in Real-world Applications?   Who actually needs this?

I note that one could simplify a bit by omitting the ability to put the type args in instance decls -- that aspect adds no new expressiveness, but it does add implementation and specification complexity.

int-index:
> Presumably those "class method header" arguments must appear first (unlike associated types), and identically to the instance header (like associated types)? You need to say this; and give the new syntax for instance decls.

Right. I will extend the specification.

> I note that one could simplify a bit by omitting the ability to put the type args in instance decls -- that aspect adds no new expressiveness, but it does add implementation and specification complexity.

It's important to have this bit because we are very unlikely to change this for associated types. Assuming visible `forall` in terms, consider:

```
class C a where
  type F a :: a
  f a :: a

instance C Bool where
  type F Bool = True
  f Bool = True
```

We have to write `f Bool` to mirror the `type F Bool` syntax. And, in fact, this is the only reason I included it, my initial version left instances untouched.

goldfirere:
This is a very nicely written proposal. Thanks for taking such care.

However, I'm lukewarm... perhaps leaning against.

1. The wrapper workaround fully nabs this. And the workaround itself isn't too expensive. In your motivation, you say "This means that users have to define instances using original method names, but call the methods using the wrappers, which leads to more complicated APIs." This is true, but (to me) that one sentence isn't enough to drive this whole proposal.

2. The proposal doesn't solve a problem we actually have: the type of `fromInteger`. While this solves problems we don't have (and I don't mean that to be quite as snarky as it sounds), it's a shame it doesn't solve the problem we actually have.

I'm willing to be convinced otherwise, but I'm dubious of yet another subtle feature like this.

int-index:
> "This means that users have to define instances using original method names, but call the methods using the wrappers, which leads to more complicated APIs." This is true, but (to me) that one sentence isn't enough to drive this whole proposal.

The wrapper workaround would need `-XAllowAmbiguousTypes` for the VDQ use case:

```
class Storable a where
  _sizeOf :: Int
  _alignmentOf :: Int

sizeOf :: forall a -> Storable a => Int
sizeOf a = _sizeOf @a

alignmentOf :: forall a -> Storable a => Int
alignmentOf a = _alignmentOf @a
```

I agree this isn't the worst thing in the world, but it's still quite a bit of boilerplate compared to

```
class Storable a where
  sizeOf a :: Int
  alignmentOf a :: Int
```

goldfirere:
> The wrapper workaround would need `-XAllowAmbiguousTypes` for the VDQ use case:

I thought you might say that. Yes, it would. And I agree that's bad. I just don't quite think it's worth fixing. It's a close judgment call.

gridaphobe:
I'm also leaning against this proposal at the moment. It seems to me that a general solution to the `fromInteger` issue -- i.e. a method to control where class variables are quantified -- would subsume this proposal, but I don't see how this proposal could be extended to solve the general problem. So I'm reluctant to introduce new syntax that would ideally be made quickly obsolete.

On a meta-note: proposals that are accepted are enshrined in the main repo and eventually (ideally) in the official GHC User's Guide. But sometimes we reject a proposal because there's a good enough workaround (e.g. the wrappers @goldfirere suggested). It would be nice to document these alternatives in a more official (and discoverable!) manner.

int-index:
> I don't see how this proposal could be extended to solve the general problem

As I mentioned in https://github.com/ghc-proposals/ghc-proposals/pull/267#issuecomment-527012953, that wouldn't be hard:

```
class Num a where
  fromInteger (x :: Integer) @a :: a
```

In fact, the proposal contains an explicit clause to disallow this:

> No other variables can be bound in the class method header.

Why is this disallowed? See the "Binding non-class variables" part of "Limitations" section. 

The proposal is forward-compatible with such an extension! The goal here is to lay a stable foundation.

gridaphobe:
Ah, sorry I missed that.

I suppose you could keep moving parameters to the LHS of the `::`, but it feels pretty unnatural given the conventional way of writing signatures in Haskell. I know we've recently accepted [Function Result Signatures](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0228-function-result-sigs.rst), which extends support for this style of signatures, but we've also accepted [Standalone Kind Signatures](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0054-kind-signatures.rst), which extends support for the more traditional style.

So I guess the consistent thing to do then would be to expect support for both standalone and inline signatures in future extensions like this one. But it seems then we run right back into the issues of #148.

int-index:
>  I know we've recently accepted Function Result Signatures, which extends support for this style of signatures, but we've also accepted Standalone Kind Signatures, which extends support for the more traditional style.

Function result signatures are about inline signatures to the left of `=`, and here I'm proposing inline signatures to the right of `::`. So it's a different feature. In fact, standalone kind signatures could be extended to support this as well:

```
type D (a :: Type) :: F a -> Type
data D a b = ...
```

This would the equivalent to

```
type D :: forall (a :: Type) -> F a -> Type
data D a b = ...
```

I'm not proposing this because there's no need for this. Moving binders to the left of `::` is not that useful outside of classes. But it *is* useful inside classes, and in fact, this proposal is a manifestation of the following thought process:

* What if the declaration header for an associated type *is* its signature?

This would nicely resolve #259

goldfirere:
> This would nicely resolve #259

You're getting dangerously close to proposing that we have CUSKs! In some ways, a CUSK is just a signature with some variables to the left of the `::` (for a datatype, at least).

int-index:
> You're getting dangerously close to proposing that we have CUSKs! In some ways, a CUSK is just a signature with some variables to the left of the `::` (for a datatype, at least).

I'm close, but I'm not crossing the line. What's the issue with CUSKs? It's that we have heuristics to determine when we have one:

```
-- has a CUSK
data T1 (k :: Type) (a :: k) :: Type where ...

-- doesn't have a CUSK
data T2 k (a :: k) :: Type where ...
```

These rules are delicate, as if we were laying a trap for the programmer.

However, if we simply push some of the binders to the left of `::`, we get something that *looks* like a CUSK but shares none of its issues:

```
-- has a "CUSK"-looking standalone type signature
type T1 (a :: Type) (a :: k) :: Type
data T1 a k where ...

-- doesn't have it.
data T2 a k where ...
```

goldfirere:
I've been prodded to take another look at this proposal. While I think it has some merit, I don't see that the specification/documentation/implementation costs are recouped by the gains. Maybe if we had dependent types and could use this proposal to get rid of associated type families (or some similar win), I would be more in favor. As of now, my previous position (https://github.com/ghc-proposals/ghc-proposals/pull/267#issuecomment-528445375) remains current.

int-index:
>  Maybe if we had dependent types and could use this proposal to get rid of associated type families (or some similar win), I would be more in favor.

Indeed, that's the end-goal. Associated type families allow reordering class variables and changing their visibility, so if we are to replace them with class methods, we'd need a similar feature for class methods.

In particular, it answers the question of SAKS for associated types: the header is the signature.

Alternatively, we could go in the opposite direction and limit what associated types can do. Use the same rules as we have for class methods today: class variables are added in front with an implicit `forall`:

```
class C a b where
  type F :: Type

-- top-level kind: F :: forall a b. Type
```

Unfortunately, this seems suboptimal to me. Consider this use case:

```
class IsList l where
  type Item l
  fromList  :: [Item l] -> l
  toList :: l -> [Item l]
```

We'd have to change it to:

```
class IsList l where
  type Item :: Type
  fromList  :: [Item @l] -> l
  toList :: l -> [Item @l]
```

Note the added `@`. That's because we no longer control the visibility.

int-index:
>  I don't see that the specification/documentation/implementation costs are recouped by the gains

The informal specification is easy:

* Class method signatures can mention class variables on the left-hand side. These occurrences define the ordering and visibility of such variables, and have no other effect.

And I'd like to stress that those are occurrences, not binding sites. The variable is bound by the class header:

```
class C a b where  -- binding sites for a, b
  type F b @a :: Type -- occurrences of a, b
```

Therefore, I don't believe the implementation would be very complicated. It would just set the specificity and the order of variables that are quantified. Easy!

int-index:
Immediately after release of GHC 9.10 we see programmers use intricate workarounds to get visible `forall` in class methods:

```haskell
class NewStorable2 a where
  sizeOf :: forall a' -> a ~ a' => Int
```

Source: https://zenn.dev/mod_poppo/articles/whats-new-in-ghc-9-10

In light if this, I decided to refurbish this proposal. @simonpj, @goldfirere, you're welcome to take another look before I submit to the committee.

AntC2:
> use intricate workarounds

Is that out of a 'business need', or merely because they're kicking the tyres on a new release? IoW does @goldfirere's earlier comment still apply?

> [I don't see that ... costs are recouped by the gains](https://github.com/ghc-proposals/ghc-proposals/pull/267#issuecomment-578860258)

The workaround shows there's a workaround. It doesn't look any more intricate than what comes with the territory for Haskell these days.

int-index:
> Is that out of a 'business need', or merely because they're kicking the tyres on a new release? 

It seems like they're just playing around with the new features, but it's an independent discovery of the problem, which I find somewhat significant (you may disagree). In any case, this is a gaping hole in the design of `RequiredTypeArguments`, and I'm eager to find a proper solution.

> The workaround shows there's a workaround.

Fair enough, but it makes type signatures more complex (e.g. if you check with `:t` in GHCi) and the equality constraint has a performance overhead because it's not erased.

> IoW does @goldfirere's earlier comment still apply?

I suppose it does, but since it hinges on a cost/benefit analysis, context matters. Now that `RequiredTypeArguments` is released, the proposal has more weight (or so it seems to me).

AntC2:
> the problem, which I find somewhat significant (you may disagree). In any case, this is a gaping hole in the design of `RequiredTypeArguments`,

Yes I do agree this is a gaping hole, and I said so at the time of the discussion round `RequiredTypeArguments`:

* The `forall a -> ...` is mixing up different concerns:
1. **Ordering** of the type params to the method or class. (As you say in the o.p.)
2. **Usage** whether the type param appears as the type of a term or is required as a type expression.
3. **Ordering** of the appearance of that required type expression amongst the arguments to the method.

* And the `forall a -> ...` syntax I continue to find perverse. (Just as I find the arrow in `ViewPatterns` perverse.)

Re point 3. above, here's an independent motivation: suppose I have a function/method that takes a list (or listlike) as arg. Usually the type of the elements can be inferred from the list content, and it would be annoying to specify that type separately. But if the list is empty, you do need to specify the content type. I'd like to express that as a type expression _after_ the list -- because that's where `::` signatures appear:

```
x = foo [] Int
y = foo [True] _           -- infer the type for me
z = foo [1,2] Float        -- 1, 2 ambiguous as to which Num type

-- foo :: forall a. Foo a => [a] -> @a -> ...
-- foo :: forall a. Foo a => [a] -> type a -> ...     -- alternative syntax
```

9.10 has only just been released/non-one's yet using it in anger. I suggest you revisit the design of `RequiredTypeArguments` before we get a design baked in with a "gaping hole". If you want to use `@` for this purpose [**] (although I'm flipping the usage of `@` in terms, so that's not great), I suggest

```
class C a b c where
   type F b c a :: @c -> (a, b, c) -> Bool
   f b c a :: @c -> (a, b, c) -> Bool

-- f :: forall b. forall c. forall a. @c -> C a b c => (a, b, c) -> Bool
-- F :: forall b. forall c. forall a. @c ->            (a, b, c) -> Bool

-- or alternative syntax, for the same typings
class C a b c where
   type F b c a :: type c -> (a, b, c) -> Bool
   f b c a :: type c -> (a, b, c) -> Bool

```

[**] I find the proposed syntax here perverse again: I'm required to mark with `@` the type arguments that _aren't_ required -- that is, are behaving as per usual Haskell, not partaking in this new-fangled TypeArguments gig.

simonpj:
Five years ago (how time passes) I wrote

> it looks feasible to me, but I'm a bit unconvinced about the cost/benefit ratio

RequiredTypeArguments strengthens the case.  

I'm still unconvinced about changing instance declarations.

int-index:
> I'm still unconvinced about changing instance declarations.

It's about making definition sites and use sites match up. For example, if we have

```
class C a where
  f a :: a -> Bool
```

the calls to `f` will look like:
* `f Int 10`
* `f Bool True`

so I'd expect the LHS of `f's` equations to have those `Int` and `Bool` arguments too, thus

```
instance C Int where
  f Int x = ...

instance C Bool where
  f Bool b = ...
```

And this is how it's done with associated types, so there's an established precedent.

vagarenko:
In motivation you write:
> At the moment, the signature of a class method is prefixed with class variables and a class constraint:

```
class C a b c where
  f :: (a, b, c) -> Bool

-- f :: forall a b c. C a b c => (a, b, c) -> Bool
--      ^^^^^^^^^^^^^^^^^^^^^^^^
--      compiler-generated prefix
```

Why not just force the user to write that prefix manually?
```
class C a b c where
  f :: forall a b c. C a b c => (a, b, c) -> Bool
```

int-index:
> Why not just force the user to write that prefix manually?

That'd be a move in the direction of #148, which is a plausible alternative.

VitWW:
Splitting `forall` into 4 different categories (`forall.`, `forall->`, `foreach.`, `forreach->`) it is a bad design.
It is better to have alternative syntax - just one `forall` and special marks for different properties.
Like next:
```haskell
foo :: forall {k} (a :: k) (type b) c (dynamic d). some_signature
--             ^    ^        ^      ^    ^
--             |    |        |      |    unerased
--             |    |        |        ordinary
--             |    |          visible
--             |    ordinary, but kinded
--              inferred
```

int-index:
@adamgundry I'd like to submit this proposal to the committee.

proposals/0000-class-method-headers.rst
```diff
@@ -0,0 +1,437 @@
...
+
+    fWrapper :: forall b. forall c -> forall a. C a b c => (a, b, c) -> Bool
+    fWrapper @b c @a = f @a @b @c
+
+Our proposal is to take the LHS of the wrapper, ``fWrapper @b c @a``, and
+integrate it into the original ``f``'s declaration.
+
+This approach avoids the main problem of the wrapper workaround: the
+introduction of unnecessary named entities to the APIs of libraries (it is
+generally desirable to keep APIs simple and easy to browse).
```
JakobBruenker, line 111: In the comments you explained that the wrapper would also require `-XAllowAmbiguousTypes`. Might be worth mentioning, if you think it adds to the motivation.

proposals/0000-class-method-headers.rst
```diff
@@ -0,0 +1,437 @@
...
+      Expected: f @Int
+        Actual: f @Bool
+    • In the method definition for ‘f’
+      In the instance declaration for ‘C Int’
+
+  This is the same message as one would get for associated types.
+  The implementation may opt to provide a different error message in the same spirit.
+
+Effect and Interactions
+-----------------------
```
JakobBruenker, line 308: This section should mention whether there's any interaction with `-XInstanceSigs`, which one might reasonably expect to be affected

JakobBruenker (commented):
Thanks for the proposal, it's clearly written and most of the questions I had about were already answered by the proposal.

I was initially somewhat apprehensive about introducing a bit of syntax that's specific to methods and does not apply to other functions, but given the discussion it seems like this could eventually become part of a broader push to have a mostly unified class of syntax that's available to things inside classes - that being methods and associated types - and one class of syntax (standalone type/kind signatures) that is available to functions and type families outside of classes.

Given the difficulty we've had in adapting standalone kind signature syntax to associated types, that may not be the worst thing.

Before I decide on my recommendation to the committee, I left a few clarifying questions in the review comments, and have one more question:

The proposal doesn't really need to have an answer to this given it's not part of the language at this point, but I'm curious if you have any thoughts about whether this syntax would also work somehow with erasure quantification (keeping in mind that the syntax for that has not actually been decided on yet).
I suppose it may just be the case that, even ignoring syntax, it won't make sense to allow users to choose the erasure of class variables on methods? I'm not really familiar enough with how class dictionaries work to have an answer for this.

sgraf812:
I find the proposal well-written and decently motivated, so I'm generally in favor.

However, a few items left me puzzled. This might be because I need to catch up on the implied Dependent Haskell vision.

1. Although the proposal mentions associated types in the motivation, the rest of the proposal never mentions them again. That left me thinking whether `class C a b c where type F a @b c :: Type` is currently an allowed declaration. Is it? (I think it is.) That would strengthen the case of the proposal, because then it is just a matter of parity between associated type families and methods.
2. You argue that "changing the ordering of class variables in the class declaration header has the downside that now the class constructor has a different ordering of variables, not just its method". But similarly, your proposal has the downside of changing the ordering of *all* uses of the function! What if I'd prefer a different order? I would still need to write a wrapper, so your argument is rather weak to me. However, I think this is very similar to the situation with term-level binders, where it is the job of the library author to recommend one argument order that can be overriden with combinators such as `flip` (which I don't think I've ever used). And consistency with the term-level is a relatively stronger argument to me. Thus, I think it would help to bring this argument in the proposal instead!
3. That brings me to more of a meta-critique of Haskell: The lack of named arguments. I think that named arguments are far superior to min/maxing ordered arguments. See also [Lean 4](https://github.com/leanprover/lean4/blob/master/doc/lean3changes.md#function-applications) which adopted named arguments when transitioning from Lean 3 to Lean 4.

Ideally, I would rather see general focus shift towads named arguments (3), but I can see that there is value in improving and making what we have more consistent, independent of introducing a language feature with which we would be better served. That is, IMO consistency (with term-level and associated types) is the big argument in favor of this proposal, not so much the use cases it enables (but that might be just me not having had those use cases so far).

sgraf812 (changes requested):

JakobBruenker:
> I just figured that it's not really my job do a review yet -- sorry, @JakobBruenker!

@sgraf812 I think it's perfectly fine for anyone to review before, during, or after a proposal has the "Pending shepherd recommendation" label - doing it sooner might mean more work for you since you might have to take another look later if any other changes are made before the shepherd recommendation, but I think for everyone else it's probably better to be made aware of any concerns sooner rather than later.

proposals/0000-class-method-headers.rst
```diff
@@ -0,0 +1,437 @@
...
+
+  This means that users have to define instances using original method names,
+  but call the methods using the wrappers, which leads to more complicated
+  APIs.
+
+The problem is further exacerbated by the introduction of ``RequiredTypeArguments``
+(visible forall in types of terms). If we want to visibly quantify over some of
+class variables, there are also two workarounds:
+
+* Dummy type variable with an equality constraint:
```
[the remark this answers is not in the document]
JakobBruenker: One interesting note about this workaround, which doesn't apply to the others, is that since arguments will still need to be in dependency order, this workaround will still have a place in cases where that's not desirable, as in e.g.
```haskell
ghci> (.:) :: forall a. a -> forall b -> a ~ b => a; x .: _ = x
ghci> 4 .: Float
4.0
```
(Not a class method, but illustrates the point I think.)

proposals/0000-class-method-headers.rst
```diff
@@ -0,0 +1,437 @@
...
+       f @a :: a
+
+     instance C Int where
+       f @Int = 42
+
+   The type in the method equation must be identical to the one in the instance
+   head.
+
+#. In instances, method definitions *must* explicitly write out
+   instantiations of visible class variables::
```
aspiwack, line 177: Fair enough. But, unless I'm mistaken, it doesn't look like requiring visible arguments to be specified is strictly necessary. Would you discuss the reason for this choice in the _Alternatives_ section? It would be helpful

proposals/0000-class-method-headers.rst
```diff
@@ -0,0 +1,437 @@
...
+
+     class C a where
+       f :: a         -- note: no @a
+
+     -- f :: forall a. C a => a
+
+     instance C Int where
+       f @Int = 42    -- still allowed
+
+#. Class variables bound in methods always precede other arguments.
```
aspiwack, line 208: This seems to answer my question above. I guess that turning this extension on can break programs which are implemented with `-XTypeAbstractions`, correct? If so, maybe a little word in the _Effect and interaction_ section would be warranted.

int-index:
Thank you for the reviews! It will take me some time to properly address all comments, maybe I'll get to it by the end of this or next week.

goldfirere:
I took a quick read-through. I did not examine others' detailed feedback first, so apologies if there are duplicates here.

1. I agree that context matters in my previous cost-greater-than-benefit opinion. And the benefit has increased. The cost has not. I'm more positive now in my opinion here.... but still not totally blown over. (For those not keeping score at home: I am no longer on the committee, and so my opinion matters less than it once did.)
2. I would be more excited about this proposal if it solved all problems in the class of reordering/relabeling class method arguments. But it doesn't. I can't, for example, get `fromInteger :: Integer -> forall a. Num a => a`, which is what I'd dearly want there. The Drawbacks section of the proposal highlights the fact that you can't quantify a variable local to one method's type in front of the class variables, but I did not immediately see a mention of the fact that you can't move a term argument in front either.
3. This proposal does not allow you to mark a variable as *inferred*, though this would be very easy to add, with some braces.
4. Kind annotations on LHS binders could serve to bring into scope kind variables: `f @a @(b :: k -> k) c :: a -> Proxy (b c) -> F k`. In this example, `k` would not be explicitly bound in the class header, and thus could not be just brought into scope on its own to the left of the `::`.
5. It looks like the syntax allows LHS binders on ordinary variable signatures (not just in classes). Presumably these would be rejected later in the pipeline, but it would be good to say so.

Summing it together from me: I would be in support of this proposal if it were a prefix of some larger proposal that did allow arbitrary swizzling of clsas method types. We don't need to hammer out the details all at once, and we don't need to implement it all at once. But casting this as the first part of a grander plan helps give us assurance that we're not painting ourselves into a corner.

goldfirere:
@sgraf812 One day, maybe about 10 years ago, I stood in @sweirich's office as we were designing what would become `-XTypeApplications`. We wondered about named arguments vs. positional ones. We decided that named arguments would leak details of the type variables used in a type signature, which previously were purely local. So we decided to go with positional arguments. It was sensible enough at the time, but in the light of experience, I think we got that one wrong.

More broadly, I would support a general way to add labeled arguments to Haskell. OCaml has them, and they're pretty fantastic.

JakobBruenker:
I also had some comments on named arguments, and so thought it might be good to have a dedicated place for that discussion, in lieu of derailing this proposal: https://github.com/ghc-proposals/ghc-proposals/discussions/654

simonpj:
Personally I'm not yet convinced that the pain (language complexity, implementation complexity) is worth the gain.

JakobBruenker:
If this proposal is accepted and implemented, will there still be any legitimate use-cases for `-XAllowAmbiguousTypes` that can't easily be replaced by `-XRequiredTypeArguments`?

If so, that seems like a point in favor to me, considering that `-XAllowAmbiguousTypes` is still de facto module-wide.

AntC2:
> get rid of _all_ remaining ... uses

> [from a now-gone version of @JakobBruenker's comment] considering that ... is still de facto module-wide.

`ScopedTypeVariables` is the module-wide extension I'd like to get rid of all its remaining uses. Does this tweak help with that? (Please say so/give an example.)

PR #271, opened 2019-09-06
Set clearer expectations about the proposal process

simonpj:
This PR is my attempt to write down clearer expectations about the GHC proposals process.

Particularly

1. The need for precision
2. The idea that a proposal should not languish under committee review for long; but may well be returned for revision without any negative implication.  (As I did yesterday for Richard's role-inference proposal, which I basically like.)
3. Clearly stating that a proposal should be accepted/rejected as-is.  If it needs modification (a typical case) it should be returned fro revision.
4.  Adding explicit timescale expectations so authors know wnen to time out.

All of this is, of course, for the committee to debate.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -75,7 +75,7 @@ This section outlines what stages a proposal may go through. The stage is identi
    `≡ List of proposals under review <https://github.com/ghc-proposals/ghc-proposals/pulls?q=is%3Aopen+is%3Apr+label%3A%22Pending+committee+review%22>`_
 
 6. Eventually, the committee rejects a proposal (label: Rejected), or passes it back to the
-   author for review (label: `Needs revision <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3A%22Needs+revision%22>`_), or accepts it (label: `Needs revision <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3A%22Accepted%22>`_).
+   author for review (label: `Needs revision <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3A%22Needs+revision%22>`_), or accepts it (label: `Accepted <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3A%22Accepted%22>`_).
 
    Acceptance of the proposal implies that the implementation will be accepted
    into GHC provided it is well-engineered, well-documented, and does not
--- a/README.rst
+++ b/README.rst
@@ -108,6 +108,8 @@ Proposals are written in either `ReStructuredText <http://www.sphinx-doc.org/en/
 
 Poposals should follow the structure given in the `ReStructuredText template <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0000-template.rst>`_, or the `Markdown template <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0000-template.md>`_.  (The two are identical except for format.)
 
+See the section `Review criteria <#review-criteria>`_ below for more information about what makes a strong proposal, and how it will be reviewed.
+
 To start a proposal, create a pull request that adds your proposal as ``proposals/0000-proposal-name.rst`` or ``proposals/0000-proposal-name.md``. Use the corresponding ``proposals/0000-template`` file as a template.
 
 If you are unfamiliar with git and github, you can use the GitHub web interface to perform these steps:
--- a/README.rst
+++ b/README.rst
@@ -148,12 +148,12 @@ Here are some characteristics that a good proposal should have
 
 * It should offer evidence of utility.  Even the strongest proposals carry costs:
 
-  * For programmers: they usually make the language just a bit more complicated;
-  * For GHC maintainers: they make the implementation a bit more complicated;
-  * For those wanting to proposals in the future: they usually add new back-compat burdens, and consume syntactic design space.
-  * Moreover, all these costs contitute a permanent tax on every future programmer, langauge designer, and GHC maintainer.
+  * For programmers: most proposals make the language just a bit more complicated;
+  * For GHC maintainers:  most proposals make the implementation a bit more complicated;
+  * For future proposers:  most proposals consume syntactic design space add/or add new back-compat burdens, both of which make new proposals harder to fit in.
+  * It is much, much harder subsequently to remove an extensiuon than it is to add it.
 
-  The tax may well be worth it (a language without polymorphism
+  All these costs contitute a permanent tax on every future programmer, langauge designer, and GHC maintainer.  The tax may well be worth it (a language without polymorphism
   would be simpler but we don't want it), but the case should be made.
 
   The case is stronger if lots of people jump in saying "yes, this would be so
--- a/README.rst
+++ b/README.rst
@@ -106,7 +106,7 @@ How to start a new proposal
 
 Proposals are written in either `ReStructuredText <http://www.sphinx-doc.org/en/stable/rest.html>`_ or `Markdown <https://github.github.com/gfm/>`_. While the proposal process itself has no preference, keep in mind that the `GHC Users Guide <http://downloads.haskell.org/~ghc/latest/docs/html/users_guide/editing-guide.html>`_ uses ReStructuredText exclusively. Accepted proposals written in ReStructuredText thus have the slight benefit that they can be more easily included in the official GHC documentation.
 
-Poposals should follow the structure given in the `ReStructuredText template <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0000-template.rst>`_, or the `Markdown template <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0000-template.md>`_.  (The two are identical except for format.)
+Proposals should follow the structure given in the `ReStructuredText template <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0000-template.rst>`_, or the `Markdown template <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0000-template.md>`_.  (The two are identical except for format.)
 
 See the section `Review criteria <#review-criteria>`_ below for more information about what makes a strong proposal, and how it will be reviewed.
 
--- a/README.rst
+++ b/README.rst
@@ -15,53 +15,57 @@ GHC.
 * `≡ List of implemented proposals <https://github.com/ghc-proposals/ghc-proposals/pulls?q=is%3Apr+label%3A%22Implemented%22>`_
 * `≡ List of all proposals <https://github.com/ghc-proposals/ghc-proposals/pulls?q=>`_
 
-What is the timeline of a proposal?
+The life cycle of a proposal
 -----------------------------------
 
-1. The author drafts a proposal.
+This section outlines what stages a proposal may go through. The stage is identified by a GitHub label, which is identified in the following list.
+
+1. (No label.) The author drafts a proposal.
 
    `What is a proposal? <#what-is-a-proposal>`_ • `What should a proposal look like? <#what-should-a-proposal-look-like>`_
 
-2. The author submits the proposal to the wider Haskell community for discussion, as a pull request against this repository.
+2. (No label.) The author submits the proposal to the wider Haskell community for discussion, as a pull request against this repository.
 
    `How to submit a proposal <#how-to-start-a-new-proposal>`_
 
-3. The wider community discusses the proposal in the commit section of the pull
+3. (No label.)  The wider community discusses the proposal in the commit section of the pull
    request, while the author refines the proposal. This phase lasts as long as necessary.
 
    `Discussion goals <#discussion-goals>`_ •
    `How to comment on a proposal <#how-to-comment-on-a-proposal>`_ •
    `≡ List of proposals under discussion <https://github.com/ghc-proposals/ghc-proposals/pulls?q=is%3Aopen+is%3Apr+no%3Alabel>`_
 
-4. Eventually *the proposal author* brings the proposal before the committee for review.
+4. Label: `Pending shepherd recommendation <https://github.com/ghc-proposals/ghc-proposals/pulls?q=is%3Aopen+is%3Apr+label%3A%22Pending+shepherd+recommendation%>`_.  Eventually *the proposal author* brings the proposal before the committee for review.
 
    `How to bring a proposal before the committee <#how-to-bring-a-proposal-before-the-committee>`_ •
-   `Who is the committee? <#who-is-the-committee>`_
+   `Who is the committee? <#who-is-the-committee>`_   •
+   `≡ List of proposals waiting for shepherd <https://github.com/ghc-proposals/ghc-proposals/pulls?q=is%3Aopen+is%3Apr+label%3A%22Pending+shepherd+recommendation%22>`_ •
 
-5. One committee member steps up as a shepherd, and generates consensus within the committee within four or five weeks.
+5. Label: `Pending committee review <https://github.com/ghc-proposals/ghc-proposals/pulls?q=is%3Aopen+is%3Apr+label%3A%22Pending+committee+review>`_.  One committee member steps up as a shepherd, and generates consensus within the committee within four or five weeks.
 
    `Committee process <#committee-process>`_ •
    `Review criteria <#review-criteria>`_ •
-   `≡ List of proposals waiting for shepherd <https://github.com/ghc-proposals/ghc-proposals/pulls?q=is%3Aopen+is%3Apr+label%3A%22Pending+shepherd+recommendation%22>`_ •
    `≡ List of proposals under review <https://github.com/ghc-proposals/ghc-proposals/pulls?q=is%3Aopen+is%3Apr+label%3A%22Pending+committee+review%22>`_
 
-6. Eventually, the committee rejects a proposal, or passes it back to the
-   author for review, or accepts it.
+6. Eventually, the committee rejects a proposal (label: Rejected), or passes it back to the
+   author for review (label: `Needs revision <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3A%22Needs+revision%22>`_), or accepts it (label: Accepted).
 
    Acceptance of the proposal implies that the implementation will be accepted
    into GHC provided it is well-engineered, well-documented, and does not
    complicate the code-base too much.
 
-   `≡ List of accepted proposals <https://github.com/ghc-proposals/ghc-proposals/tree/master/proposals>`_
+   `≡ List of accepted proposals <https://github.com/ghc-proposals/ghc-proposals/tree/master/proposals>`_ •
+   `≡ List of proposals being revised <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3A%22Needs+revision%22>`_ •
+   `≡ List of rejected proposals <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3A%Rejected%22>`_
 
-7. If a proposal sees no activity for along time, they are marked as “dormant”,
+7. Label: `Dormant <https://github.com/ghc-proposals/ghc-proposals/pulls?q=is%3Aopen+is%3Apr+label%3A%22Dormant>`_.  If a proposal sees no activity for along time, it is marked as “dormant”,
    and eventually closed.
 
    `What is a dormant proposal <#what-is-a-dormant-proposal>`_ •
    `≡ List of dormant proposals <https://github.com/ghc-proposals/ghc-proposals/pulls?q=is%3Apr+label%3A%22dormant%22>`_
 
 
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -124,6 +124,45 @@ g. *Implementation Plan* (Optional): If accepted who will implement the change?
...
+  The tax may well be worth it (a language without polymorphism
+  would be simpler but we don't want it), but the case should be made.
+
+  The case is stronger if lots of people jump in saying "yes, this would be so
+  useful to me".  The committee is often faced with proposals that are reasonable,
+  but where there is a suspicion that no one other than the author cares.
+  Defusing this suspicion, by describing use-cases and inviting support from others,
+  is helpful.
+
+* It should be compiously iillustrated with examples, to aid understanding.
```
nomeata:
```suggestion
* It should be compiously illustrated with examples, to aid understanding.
```

README.rst
```diff
@@ -309,17 +356,58 @@ proposal is ready for decision.
...
      discussion has started.
 
--  Discussion among the committee ensues on the discussion thread.
-   Silence is understood as agreement with the shepherd's recommendation.
+-  Discussion among the committee ensues, in two places
+
+   * *Technical discussion* takes place on the discussion thread, where others may
+     continue contribute.
+
+   * *Evaluative discussion", about whether to accept, reject, or return the
```
nomeata:
```suggestion
   * *Evaluative discussion*, about whether to accept, reject, or return the
```

nomeata:
I think you are omitting the maybe biggest change:

> For acceptance, a proposal must have at least *some* enthusiastic support from member(s) of the committee

This is a deviation from our current handling, which is more like “nobody takes offense with a proposal and we can imagine it being useful to some”.

simonpj:
> This is a deviation from our current handling, which is more like “nobody takes offense with a proposal and we can imagine it being useful to some”.

I would not say it's a devitation -- it's just something we have never explicitly discussed, which is why I'm raising it here. 

The difficulty I see is that if we adopt every proposal for which no one has a strong technical objection,  but no one is actually enthusiastic, we will gradually end up wiith a dogs breakfast, full of features that no one actively hates, but which few actually like.

For peer review, Oscar Nierstrasz's paper [Identify the champion](http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.28.1243) has been quite influential.

But in practice we have found it hard to reject a proposal unless someone actively argues against it.   In the long run I don't think that criterion will serve us well.   So yes, I am proposing to articualte something new, namely that we need a champion on the committee.

Someone might be a champion because, while they personally don't feel strongly, they have become persuaded that there is an important constituency that does.

simonmar:
> The difficulty I see is that if we adopt every proposal for which no one has a strong technical objection, but no one is actually enthusiastic, we will gradually end up wiith a dogs breakfast, full of features that no one actively hates, but which few actually like.

Yes! This is exactly what I've been trying to articulate at various times. It's closely related to the point that I've mentioned before about retaining a "coherent language".

I think the committee should have a much higher bar for proposals that fork the language rather than simply extend it. What I mean by that is, we apply a simple test: is this an extension that most people would not mind having enabled by default, even if they don't actively use it? If the answer is yes, then this is a language extension, whereas if the answer is no then this is a language fork: an extension that only a subset of users would enable, and would likely never be a part of the language standard unless a clear majority of users supported it, and even then we have a migration problem.

README.rst
```diff
@@ -124,6 +124,45 @@ g. *Implementation Plan* (Optional): If accepted who will implement the change?
...
+  a BNF for the changes.
+
+  Ultimately, the necessary degree of precision is a judgement that the committee
+  must make; but authors should try hard to offer precision.
+
+* It should offer evidence of utility.  Even the strongest proposals carry costs:
+
+  * For programmers: they usually make the language just a bit more complicated;
+  * For GHC maintainers: they make the implementation a bit more complicated;
+  * For those wanting to proposals in the future: they usually add new back-compat burdens, and consume syntactic design space.
```
simonmar: this sentence no verb?

README.rst
```diff
@@ -124,6 +124,45 @@ g. *Implementation Plan* (Optional): If accepted who will implement the change?
...
+  * Moreover, all these costs contitute a permanent tax on every future programmer, langauge designer, and GHC maintainer.
```
simonmar: Maybe point out that removing extensions is much, much, harder than adding them?

README.rst
```diff
@@ -124,6 +124,45 @@ g. *Implementation Plan* (Optional): If accepted who will implement the change?
...
+
+  The tax may well be worth it (a language without polymorphism
+  would be simpler but we don't want it), but the case should be made.
+
+  The case is stronger if lots of people jump in saying "yes, this would be so
+  useful to me".  The committee is often faced with proposals that are reasonable,
```
simonmar: Do we really want to invite this noise on the github threads? I liked the earlier suggestion we had about named supporters in the proposal itself.

simonpj:
> I think the committee should have a much higher bar for proposals that fork the language rather than simply extend it. What I mean by that is, we apply a simple test: is this an extension that most people would not mind having enabled by default, even if they don't actively use it?

That's an interesting idea. 

I  think you are specifically *excluding* the fact that many extensions use up a bit more syntactic space, and thereby reduce future wiggle room.   That's not something that "most people" will directly care about.

Examples:

* postfix `qualified` in import statements.  Having that on all the time does me no harm.
* extra-commas in tuples.  Having that on all the time messes up tuple sections; so it would be fork-like. 

Articulating pricniples like this is helpful to authors, because it helps them to know how to frame their proposals.

simonpj:
> I liked the earlier suggestion we had about named supporters in the proposal itself.

I'd forgotten about this.  For the benefit of others, the idea is this:

* Explicitly encourage people who like, and would actively use, the proposal to add their name and affiliation to a new "Endorsements" section.   (We should make it clear just how to do this -- a PR on the PR?)

The idea is it allows enthusiastic supporters to add their support in a clear way -- but one that they won't do on a whim because it'll durably attach their name to the  proposal.   It also gives the committee a way to gauge support.

Lack of endorsements is not a reason for rejection; but their presence is a positive sign.

What do others think?

README.rst
simonpj: I have reworded slightly

README.rst
simonpj: I have reworded slightly

README.rst
Anrock: There are message reactions in Github like +1 and -1 made specifically to allow expressing "this would be useful to me" without polluting discussion at the same time. Not sure if it's okay to depend on Github-specific features, tho.

goldfirere:
Please read my commit message -- my "tweaks" are perhaps more substantial than the name implies. You can see the message by clicking on the `...` to the right of the message title in GitHub.

README.rst
nomeata: We built the process around GitHub delibarately, so yes, we can do that.

proposals/0000-template.md
```diff
@@ -81,6 +81,24 @@ Hopefully this section will be empty by the time the proposal is brought to
...
+   external contributions are meant to clarify my intent, not change it, and I retain
+   the ability to shape the final proposal.
+   
+2. Writing precise specifications is hard. Accordingly, I seek to increase my
+   ability to perform this skill and welcome constructive, detailed criticism
+   in how to make my specification more precise. I recognize that doing so may
+   take many rounds and will interpret others' advice as a way to improve my
+   craft. If I get frustrated with the process, I will speak up early.
+
+
```
nomeata: Something about the way this is set up feels odd; it seems we are putting rather personal, almost emotional wording into other people’s mouths.

And do we really want such boiler plate code in all proposals, even after acceptance?

Maybe the shepherd should check if the proposal even needs help in getting more precise, and if so, use the GitHub thread or private communication to see what kind of help the author needs and wants?

goldfirere: Perhaps the wording can be improved. I agree that it's odd to ascribe emotions to others. And I agree that keeping the boilerplate in after acceptance is poor. Perhaps we remove it then, or perhaps this communication is made on another frequency.

But I do want the communication. This is an attempt to resolve a fundamental tension in our proposals process: we want to be inclusive of, say, people without PhDs in PL theory; and we want very precise specifications that can last in perpetuity. And individuals differ in how they want to bridge this tension: some would prefer a proposal to be their own work, no matter how many times it has to be revised; and some would prefer a person more experienced in this kind of thing to just fix it up. If we don't know where an individual sits on this spectrum, we'll offend some people (and discourage future proposals from them). And it can be awkward to ask this question in the middle of technical debate. So I just wanted it up front.

I'm very happy for other mechanisms to achieve this goal -- or of course I could be making much out of little here.

README.rst
```diff
@@ -302,24 +328,67 @@ proposal is ready for decision.
...
+   it might connote "we absolutely love this proposal but we want it
+   to be clear on these points".
+
+   In fact, this should happen if *any* substantive technical debate
+   takes place.  The goal of the commitee review is to say yes/no to a
+   proposal *as it stands*.  If new issues come up, they should be
+   resolved, incorporated in the proposal, and the revised proposal
+   should then be re-submitted for timely yes/no decision.  In this way,
+   *no propoosal should languish in the committee review stage for long*,
+   and every proposal can be accepted as-is, rather than suject to a raft
```
gridaphobe: s/suject/subject/

README.rst
```diff
@@ -302,24 +328,67 @@ proposal is ready for decision.
...
+   proposal for revision is not a negative judgement; on the contrary
+   it might connote "we absolutely love this proposal but we want it
+   to be clear on these points".
+
+   In fact, this should happen if *any* substantive technical debate
+   takes place.  The goal of the commitee review is to say yes/no to a
+   proposal *as it stands*.  If new issues come up, they should be
+   resolved, incorporated in the proposal, and the revised proposal
+   should then be re-submitted for timely yes/no decision.  In this way,
+   *no propoosal should languish in the committee review stage for long*,
```
gridaphobe: s/propoosal/proposal/

proposals/0000-template.md
gridaphobe: Perhaps instead of putting this boilerplate language in the template, we explain in the README that drafting the proposal should be a collaborative process, that the committee understands that not everyone has the background to specify a change with the precision the committee desires, and that we're happy to help with the actual drafting in addition to providing feedback.

I think the point of putting this language in the README instead of the template is that it sets a general expectation, rather than putting words in the author's mouth.

simonmar:
> Examples:
> 
>   * postfix `qualified` in import statements.  Having that on all the time does me no harm.
> 
>   * extra-commas in tuples.  Having that on all the time messes up tuple sections; so it would be fork-like.

Yes. One might argue that postfix qualified is somewhat fork-like in that the intention is to eventually migrate to the new syntax, however the current proposal isn't technically a fork because it allows both forms of syntax.

On the other hand, the default-qualified-import proposal most definitely *is* a fork, because it can't be enabled without authors making many changes to their code.

> Articulating pricniples like this is helpful to authors, because it helps them to know how to frame their proposals.

Yes.

simonpj:
I've added some stuff about "language forks" in response to Simon's suggestions on another thread.

Is everyone happy with this patch?

README.rst
```diff
@@ -212,11 +230,12 @@ comment or proposal, feel free to use GitHub's "Reactions"
...
 
-When the discussion has ebbed down and the author thinks the proposal is ready, he
+When the discussion has ebbed down and the author thinks the proposal is ready, he or she
 
-1. reviews the discussion thread and ensure that the proposal text accounts for
-   all salient points.
-2. adds a comment to the a pull request, briefly summarizing the major points raised
+1. Reviews the discussion thread and ensure that the proposal text accounts for
+   all salient points. *Remember, the proposal must stand by itself, and be understandable
+   without reading the disucssion thread.*
```
hsyl20: s/disucssion/discussion

README.rst
```diff
@@ -302,24 +328,67 @@ proposal is ready for decision.
...
+   there is a presumption that it should be rejected, or at least "parked".
+   (See "evidence of utility" above, under "What a proposal should look like".)
+
+-  A typical situation is that the committee, now that they have been asked
+   to review the proposal in detail, unearths some substantive technical issues.
+   This is absolutely fine -- it is what the review process is *for*!
+
+   If the technical debate is not rapidly resolved, the shepherd
+   should return the proposal for revision. Further technical
+   discussion can then take place, the author can incorporate tha
```
hsyl20: s/tha/the

README.rst
```diff
@@ -302,24 +328,67 @@ proposal is ready for decision.
...
+   to be clear on these points".
+
+   In fact, this should happen if *any* substantive technical debate
+   takes place.  The goal of the commitee review is to say yes/no to a
+   proposal *as it stands*.  If new issues come up, they should be
+   resolved, incorporated in the proposal, and the revised proposal
+   should then be re-submitted for timely yes/no decision.  In this way,
+   *no propoosal should languish in the committee review stage for long*,
+   and every proposal can be accepted as-is, rather than suject to a raft
+   of ill-specified futher modifications.
```
hsyl20: s/futher/further

README.rst
```diff
@@ -370,6 +439,13 @@ and any other relevant considerations appropriately.
    solutions, none of which have that "aha" feeling of "this is the Right
    Way to solve this"; in that case we might delay rather than forge ahead
    regardless.
+   
+-  *Does not create a language fork*.  By a "fork" we mean
+
+  * It fails the test "Is this extension something that most people would be happy to enable, even if they don't want to use it?"
+  * It fails the rest "Do we think there's a reasonable chance this extension will make it into a future language standard?"; that is, the propoosal reflects the stylistic preferences of a subset of the Haskell community, rather than a consensus about the direction that (in the committee's judgement) we want to push the whole language.
```
Topsii: s/rest/test

perhaps clarify whether it has to fail just one or both tests to be a fork. At first I thought just one.

glaebhoerl:
Just for clarity, is it the case that if they were proposed today (or after this proposal goes in), `Strict` and `StrictData` would not be accepted?

simonpj:
> Just for clarity, is it the case that if they were proposed today (or after this proposal goes in), Strict and StrictData would not be accepted?

You mean because it is fork-like?    That's a good question.  But NONE of the criteria is an absolute bar.  They are just, well, criteria; we have to balance conflicting goals.   But it helps to know what those goals *are* consciously; otherwise they may lurk in an unspoken way (perhaps even unconsciously) at the back of people's minds, which makes the discusison harder to have.

Would it help to add words to this effect?

README.rst
simonpj: Good question.  I think SImon intended that if the first test fails, then apply the second.

glaebhoerl:
Yes and yes. :)

README.rst
simonmar: One way to clarify this could be to change the second bullet to read 

* And It also fails the test "Do we think ...

Leaving open the possibility that we might accept an extension that breaks a lot of code, if we think the change will ultimately be adopted widely. (which is also what the second sentence says, I like the wording there)

README.rst
simonmar: Shall we suggest that people just thumbs-up a proposal if they like it, or comment if they have some more substantive evidence?

README.rst
```diff
@@ -212,11 +230,12 @@ comment or proposal, feel free to use GitHub's "Reactions"
 How to bring a proposal before the committee
 ---------------------------------------------
 
-When the discussion has ebbed down and the author thinks the proposal is ready, he
+When the discussion has ebbed down and the author thinks the proposal is ready, he or she
```
simonmar: s/he or she/they

README.rst
```diff
@@ -274,6 +293,13 @@ Committee process
 The committee process starts once the secretary has been notified that a
 proposal is ready for decision.
 
+The steps below have timescales attached, show that everyone shares
```
simonmar: s/show/so

README.rst
```diff
@@ -302,24 +328,67 @@ proposal is ready for decision.
 
    * post their recommendation, with a rationale, on the GitHub discussion thread,
    * label the pull request as ``Pending committee review``,
-   * notify the committee by mentioning ``@ghc-proposals/ghc-steering-committee``,
-   * include the sentence “This proposal will now be discussed by the committee.
-     We welcome all authors to join the discussion, but kindly ask others to
-     refrain from posting.” in the comment
+   * re-title the proposal, appending ``(under review)`` at the end. (This enables easy email filtering.)
```
simonmar: If the proposal ends up reverting to the discussion phase, `(under review)` should be removed again.

README.rst
```diff
@@ -302,24 +328,67 @@ proposal is ready for decision.
 
    * post their recommendation, with a rationale, on the GitHub discussion thread,
```
simonmar: Since evaluative discussion is now happening on the mailing list again, should this be on the mailing list instead?

proposals/0000-template.rst
```diff
@@ -100,6 +100,23 @@ Hopefully this section will be empty by the time the proposal is brought to
 the steering committee.
 
 
+Proposal Collaboration
```
simonmar: This is duplicated from above?

README.rst
goldfirere: How about this: we invite people to thumbs-up if they like it, but to post if they have a fresh example to share. Or, if their example is proprietary, they can just say that Company X has an example that would benefit from this. I find new examples very, very helpful.

proposals/0000-template.rst
goldfirere: Yes, because we have two different templates which should remain in sync.

goldfirere:
I've removed those sentences inviting collaboration from the templates, as I agree with the comments that they were a poor fit there. Sadly, I didn't see a good place in the README document to put the essential idea: that this proposal process is always in tension between the goals of inclusiveness and precise specification. (Why are these in conflict? Because not everyone has the training / experience to write precise specifications, which is a specific, learned skill.) I would be grateful if someone else could suggest a way forward here.

I also took the liberty to apply several of the comments suggesting small improvements, as Simon has said that he welcomes my assistance.

simonpj:
> How about this: we invite people to thumbs-up if they like it, but to post if they have a fresh example to share. Or, if their example is proprietary, they can just say that Company X has an example that would benefit from this. I find new examples very, very helpful.

We also discussed the following idea:

* Add a section "Endorsements" to the proposal that gives people to say that they support the proposal, with a sentence or two about why.

The idea here is that these endorsements would form a permanent part of the proposal.  A quick thumbs-up costs nothing; an endorsement associates your name with something.  So it's a more substantial form of support.  It also allows distinguishing "I'd like this personally" from "The lack of this feature has real costs in my workplace; if we had it, it'd change what we do".  It gives a tangible and durable way to express support.

I think I'd intended to add this but failed to do so.  Any opinions?

gridaphobe:
I like the endorsements idea. My only reluctance was to the earlier idea that proposals should need a certain number of endorsements before going before the committee. I'm concerned that a formal requirement would end up being somewhat arbitrary and would preclude obvious improvements (like #277) that are targeted at a particular niche.

isovector:
I like the idea of endorsements, though not as a mandatory part of the process. The endorsements would add strength to a proposal, but we shouldn't gatekeep on it.

simonpj:
I have updated with a section in the templates on endorsements.  I have also combined the two widely-separeated but overlapping sections "what makes a godo propsoal" and "review criteria" into one.

It woudl be good to commit all this.  Is everyone happy?

README.rst
```diff
@@ -15,53 +15,84 @@ GHC.
...
 
    `Committee process <#committee-process>`_ •
    `Review criteria <#review-criteria>`_ •
-   `≡ List of proposals waiting for shepherd <https://github.com/ghc-proposals/ghc-proposals/pulls?q=is%3Aopen+is%3Apr+label%3A%22Pending+shepherd+recommendation%22>`_ •
    `≡ List of proposals under review <https://github.com/ghc-proposals/ghc-proposals/pulls?q=is%3Aopen+is%3Apr+label%3A%22Pending+committee+review%22>`_
 
-6. Eventually, the committee rejects a proposal, or passes it back to the
-   author for review, or accepts it.
+6. Eventually, the committee rejects a proposal (label: Rejected), or passes it back to the
+   author for review (label: `Needs revision <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3A%22Needs+revision%22>`_), or accepts it (label: `Needs revision <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3A%22Accepted%22>`_).
```
gridaphobe: The second “(label: Needs revision” should be “(label: Accepted”.

nomeata:
```suggestion
   author for review (label: `Needs revision <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3A%22Needs+revision%22>`_), or accepts it (label: `Accepted <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3A%22Accepted%22>`_).
```

PR #274, opened 2019-09-09
Quick Look Impredicativity

serras:
This proposal gives `ImpredicativeTypes` a clear specification, based on [quick look impredicativity](https://www.microsoft.com/en-us/research/publication/a-quick-look-at-impredicativity/).

[Rendered](https://github.com/serras/ghc-proposals/blob/quick-look/proposals/0000-quick-look-impredicativity.md)


treeowl:
You write (modulo punctuation and alpha renaming)

> Take for example the functions
> 
> ```haskell
> choose :: forall a. (a -> Bool) -> (a -> Bool) -> Bool
> f :: (Bool -> Bool) -> Bool
> g :: (forall b. b -> b) -> Bool
```
> In theory we can accept the expressions `choose f g` and `choose g f` if we instantiate `a` with `Int -> Int`, due to contravariance.

I'm *guessing* you meant to instantiate `a` with `Bool -> Bool`, but that is surely wrong: `g` does not have type `(Bool -> Bool) -> Bool`. The only way it could sensibly work would be to instantiate `a` to `forall b. b -> b`.

**Comment by serras** [comment]
For the record, our collaborator Jurriaan Hage has made a [list of packages](https://github.com/serras/ghc-proposals/blob/quick-look/proposals/quick-look-list-pkgs.txt) from Stackage LTS 13.6 which require `ImpredicativeTypes`, `RankNTypes` or `Rank2Types`, and which may be affected by this change. The goal is to minimize the list as much as possible.

**Comment by isovector** [comment]
The specification isn't clear on whether this will alleviate the need for the impredicativity trick:

```haskell
newtype Impredicative = Impredicative ( forall x. blah )
```

Hopefully the answer is yes, and some explicit examples would be much appreciated. If it does help get rid of the need for the trick above, does it work in all cases? Can I pass around dicts in impredicative types?

goldfirere:
I'm in strong support. This "quick look" approach does a beautiful thing: it gets maybe 98% of the cases where we want impredicativity but is actually relatively simple. And for constructs where "quick look" is powerless, users can use a normal type application.

About contravariance: I'm not sure how much this will affect in practice. But it wouldn't be hard to find out. Disabling contravariance (without other changes) shouldn't be that hard -- indeed, this is already done in the branch. Then, I believe there is infrastructure for having our CI setup try to build head.hackage to a certain commit. We can see where the failures come (if at all). It wouldn't surprise me if no one uses contravariance of `(->)` in the wild.

simonpj:
> The specification isn't clear on whether this will alleviate the need for the impredicativity trick:

The answer is yes!  (Alejandro, you might want to point this out, and indeed add a bunch more examples.

simonpj:
I really like this propsosal for the reasons Richard gives.   I've been trying to do type inference for impredicativity for over a decade -- and FINALLY we have a solution that is
* Simple to describe
* Simple to implement
* Handles all the examples we've thought of

Moreover, unlike many propsals
* It is backward compatible, and involves no new syntax
* It is fully impelemented
* We have a formal specification

It simply allows more programs to typecheck than before.   

Since GHC already has a flag `-XImpredicativeTypes` you could even regard this proposal as more of a bug-fix than a fresh langauge feature.   But I think it's a Good Idea to make it a proposal because it will only be improved by discussion by members of the Haskell community.

proposals/0000-quick-look-impredicativity.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+Creating a list of such lenses, `[Lens a Int]`, requires impredicatively instantiating the constructors `(:)` and `[]`. Also if we want to replace an expression such as `view l` with `view $ l`, we need to instantiate the types of `($) :: forall a b. (a -> b) -> a -> b` impredicatively (this case is so common that GHC contains a special case for them).
+
+## Proposed Change Specification
+
+This proposal is structured in two parts. A precise specification can be found in this [paper draft](https://www.dropbox.com/s/hxjp28ym3lptmxw/quick-look-steps.pdf?dl=0), which would most surely lead to an actual paper submitted to a research conference. Below you can find a more approachable description, which could ultimately lead to a section in the GHC Users Guide.
+
+Any programmer which does not enable `ImpredicativeTypes` is unaware of the contents of this proposal (with a small caveat for users of `RankNTypes`, see below): impredicative instantiation is *not* allowed (rank-n types do not contradict that statement, since there is no instantiation going on during their type checking). In particular, type checking an application `e_0 e_1 ... e_n` in done in the following steps:
```
robrix: I’m having trouble figuring out what the caveat this paragraph refers to is. Am I missing something obvious…?

robrix:
If I’m understanding correctly, does this mean that with `-XImpredicativeTypes` enabled, GHC wouldn’t need to handle `$` specially? If so, does that imply that `-XImpredicativeTypes` should allow `.` to compose rank-_n_ functions without instantiating the rank-_n_ type variable(s)?

treeowl:
I don't see where the proposal explains what changes for `RankNTypes` when `ImpredicativeTypes` is *not* enabled.

serras:
I have made some changes to the text to clarify each of the interactions (@treeowl you may look for "RankNTypes and lack of contravariance"). I have also added some discussion about `($)`.

atennapel:
This looks very similar to the n-ary extension proposed in Daan Leijen's HMF paper. I wonder if it behaves the same. I think in your comparison table you do not consider this extension to HMF?

serras:
There are some similarities, indeed.
But HMF does another thing: it tries to *infer* types for the arguments, then builds a joint substitution, and finally does a subsumption check. It also mentions about their `subsumeN` that "the trick is now to subsume the parame-ter and argument pairs in the right order to disambiguate correctly", which reminds of the guardedness condition in quick look unification.
You can think of our proposal as trying to *infer* information from the "simple", unambiguous, parts of the program, and then push that information to the rest (lambdas, pattern matching). This is an important difference, because most difficulties in impredicative type inference come from the non-simple cases, so we don't even try to look at those in the quick look pass.

simonpj:
Perhaps, given the long history, the proposal would be made stronger by an explicit (non firewalled) link to our paper "Guarded impredicative polymorphism" (PLIDI'18).  The proposal starts from the same idea as the paper (n-ary applications, making use of invariance of some types); but the proposal is dramatically simpler.

Also pointing to the related-work section of the paper would give interested folk a good "way in" to the extensive literature.

serras:
Done. Now the introduction refers to both out paper draft and the work on guarded impredicativity.

LPTK:
> can we drop this special case from the compiler?
>
> A few versions ago, the answer would have been no. The reason was that ($) was the only function able to operate with both lifted and unlifted kinds. But nowadays, this fact is reflected in its type:
>    ```haskell
>    ($) ::
>      forall (r :: GHC.Types.RuntimeRep) a (b :: TYPE r).
>      (a -> b) -> a -> b
>    ```

Is there a reason why the type of this function is not simplified to the following?

```haskell
($) :: forall a. a -> a
```

After all, it's an identity on function types. But it does not _need_ to be restricted to function types. If we stop talking about the function types, we can stop talking about whether the argument type is lifted or unlifted:

```haskell
Prelude> :set -XRankNTypes -XImpredicativeTypes
Prelude> ($$) :: forall a. a -> a; ($$) = undefined
Prelude> :set -XMagicHash
Prelude> import GHC.Prim
Prelude GHC.Prim> :t (+#)
(+#) :: Int# -> Int# -> Int#
Prelude GHC.Prim> :t ($$) (+#)
($$) (+#) :: Int# -> Int# -> Int#
Prelude GHC.Prim> :t ($$) (+#) 1#
($$) (+#) 1# :: Int# -> Int#
Prelude GHC.Prim> :t (+# 1#) $$ 2#
(+# 1#) $$ 2# :: Int#
```

proposals/0000-quick-look-impredicativity.md
robrix: @serras: Thanks for clarifying this! 🙇

serras:
The fact that `($)` has the type `(a -> b) -> a -> b` and not other gives a lot of information for the type inference process. In particular, we know that the first argument *must* be a function, which means that any type variable `alpha` in that position can be refined to `alpha ~ beta -> gamma`. In other words, there are situations in which `f $ x` type checks but `id f x` does not.

proposals/0000-quick-look-impredicativity.md
```diff
@@ -0,0 +1,158 @@
...
+
+As most languages based on the Hindley-Damas-Milner typing discipline do, Haskell 2010 distinguishes between *monomorphic types* such as `Int -> Int`, which contain no `forall`s, and *type schemes* or *polymorphic types* like `forall a. a -> a`. Each expression in a program must, in principle, be given a *monomorphic* type -- if a variable has a polymorphic type it is *instantiated* beforehand to a *variable*, whose value is found by type inference. For example, when we write `id True`, the `a` in the type of `id` is instantiated to the type `Bool`, which means that that specific occurrence of `id` has type `Bool -> Bool`.
+
+On the other side of the spectrum we find System F, which forms the basis of GHC Core, in which instantiation is not restricted to monomorphic types. Using the syntax from `TypeApplications`, we can write `id @(forall a. a -> a)` to obtain a function of type `(forall a. a -> [a]) -> (forall a. a -> [a])`. We say that such instantiation with a polymorphic type is an *impredicative* instantiation. System F is able to instantiatiate impredicatively simply because every instantiation there is *explicit*, as one can witness when looking at GHC Core.
```
gridaphobe: this should be `id @(forall a. a -> [a])`, right? (note the list brackets)

gridaphobe:
Can we use the `head.hackage` infrastructure to get a sense for how many packages might be broken by the loss of function contravariance?

serras:
@gridaphobe I am trying to get it running. As far as I know, some important consumers of `RankNTypes` such as `syb` work perfectly fine :)

treeowl:
Could you give an example or two where contravariance will break *without* `ImpredicativeTypes`? I find that idea quite troubling.

serras:
Suppose you have the following functions:

```haskell
f :: ((forall a. a -> a) -> R) -> S
g :: (Int -> Int) -> R
```

The question is whether `f g` is well-typed. In particular this amounts to the question:

```
(Int -> Int) -> R  <=  (forall a. a -> a) -> R  ?
```

If we have full co- and contra-variance, the following question is equivalent to the following pair of questions:

```
forall a. a -> a  <=  Int -> Int  ?
               R  <=  R           ?
```

This succeeds, because we can indeed instantiate `forall a. a -> a` with `a := Int` to get `Int -> Int`.

However, if we do *not*. have contravariance, the original question is equivalent to (note the change to an equality):

```
forall a. a -> a  ==  Int -> Int  ?
               R  <=  R           ?
```

This check does not succeed, which means that `f g` is rejected. This example does not require any impredicative instantiation, only `RankNTypes`.

However, as explained in the proposal, this restriction is easy to work around by eta-expanding the offending function, `f (\x -> g x)` works perfectly fine.

Of course, one may ask "why does this trick work?". The answer is that by eta-expanding you are implicitly stating that you do *not* want quick look to consider those arguments, but instead push the information to the argument of the abstraction.

treeowl:
But why would the quick look even come into play in a module that doesn't decree `{-# language ImpredicativeTypes #-}`? Eta expansion isn't free, either notationally or (in some cases) operationally.

serras:
There is no quick look in that example. However, to make quick look *sound* we need to *drop contravariance* from function types. To keep coherence in the type system, we have to drop it also when we are not using the extension; otherwise code that compiles could cease to do so if you enable `ImpredicativeTypes`.

Note however that in the current implementation GHC is eta-expanding your code under the hood whenever real contravariance comes into play. Thus there is a notational overhead but not any new operational overhead.

Furthermore, it is arguable that eta-expanding functions without programmers' knowledge is not the best option, since it changes the laziness properties of the code. For example, `f` may throw an exception, but `\x -> f x` not.

quasicomputational:
Looking at prior art, enabling `GADTs` can also break code due to needing monomorphic local bindings.

Maybe `ImpredicativeTypes` can do something similar: introduce a new extension, say `ContravariantFunctionTypes`, which is currently enabled by default and implied by `ImpredicativeTypes`. Then you can have `NoContravariantFunctionTypes` if you want to write code that'll work with `ImpredicativeTypes`.

Or is that not an option because `ImpredicativeTypes` can have effects even across module boundaries, wheras `MonoLocalBinds` really is limited to the modules it's enabled in?

serras:
I guess this is possible, if the GHC Commitee agrees to do so. If there is agreement on this, I will update the proposal to reflect it.

However, I think that in that case the right solution would be to make `ImpredicativeTypes` imply `NoContravariantFunctionTypes`. The problem is that the combination `ImpredicativeTypes` + `ContravariantFunctionTypes` would then be unsupported, because such combination makes inference brittle, as discussed above.

mstksg:
I'd be happy to have dependable support for impredicative types in Haskell; hopefully maybe even in a way that would support CPS-encoded existential types like `[(forall a. Show a => a -> r) -> r]` being usable as a stand-in for `data Showable = forall a. Show a => Showable a` and `[Showable]`.

However, it might be worth noting that the examples in the proposals from the lens library do not match the actual design of the library as it stands today:

* `LensLike` is not a newtype wrapper (even though it could be), but instead a separate Rank-1 type that is isomorphic to the Rank-2 lens.
* `view` also does not take strictly a `Lens`, so is not RankN (so wouldn't require any special-casing on `$`).  `view` instead takes a specific monomorphisation of `Lens`, and so is also Rank-1.

Both of these cases *could* have been implemented in the way mentioned by the proposal text, nevertheless.  However, as it stands the library is built in a way to use different workaround methods than the proposal specifically states.

Not that this is of much material consequence to the proposal, just wanted to state this here in case people try to investigate these examples in real code and get disappointed :)

Ericson2314:
> There is no quick look in that example. However, to make quick look *sound* we need to *drop contravariance* from function types.

Wait, are we saying that we loose the ability to infer some things relying on contravariant arguments, or are we really left with a coarsened subsumption relation? Put differently, can I `@` my way out of actually needing eta-expand?

serras:
In the current proposal, the subsumption relation itself is restricted.

serras:
Up to now, the only problematic package has been `Cabal` (and `cabal-doctest`), which defines its own variant of `IO` with a higher-rank type:

```haskell
type CabalIO a = HasCallStack => IO a
```

This means that if this proposal goes through, Cabal would need some big amount of eta-expansion.

serras:
It might be a bit hidden in the draft paper, but there is a way to keep contravariance at the expense of (way) worse inference. One core idea of the proposal is that for almost any injective type constructors, you have that:

```
T t1 <= T t2 <==> T1 t1 ~ T t2 <==> t1 ~ t2
```

since they are invariant. However, this does not hold for function types. The result is that from knowing that `t1 -> t2 <= t3 -> t4`, you cannot infer *anything* in terms of equalities. By making the function type invariant in its first argument, we can at least infer that `t1 ~ t3`. But we could well drop it, and say "unless there is a type constructor other than `(->)` we cannot infer".

What are the consequences? Esentially, we lose the ability to infer usages of `($)`, `(.)` and those kind of function combinators. If you look at the type of the former:

```haskell
($) :: (a -> b) -> a -> b
```

none of the type variables appear in a position in which we can infer. Thus, an application such as `runST $ ...` would be rejected, since we would not be able to infer that `a` must be instantiated with `forall s. ST s v`.

serras:
All this discussion makes me very happy with @quasicomputational proposal:
- Keep contravariance under a `ContravariantFunctionType` flag, which is enabled by default;
- Make `ImpredicativeTypes` imply `NoContravariantFunctionType`.
That way we ensure that any code using `RankNTypes` keeps working in the current set-up.

simonpj:
> `type CabalIO a = HasCallStack => IO a`

Alejandro, can you explain why the lack of contravariance is a problem here?

serras:
Sorry, I gave too little information. Cabal uses a lot of functions of the form:

```haskell
f :: ... -> ((Package -> ... -> CabalIO a) -> ...) -> ... -> CabalIO a
```

where `CabalIO a` is defined as above. However, most of the times the functions given as arguments are defined in a `let` block and using `do` notation, so they are assigned a type of the form:

```haskell
((Package -> ... -> IO a) -> ...)
```

so they cannot be fed to the function.

For example, when patching `cabal-doctest`, I had to eta-expand a line to read:

```haskell
getBuildDoctests (\pkg lbi f -> withLibLBI pkg lbi (\lib clbi -> f lib clbi))
                 mbLibraryName exposedModules (const Nothing) libBuildInfo
```

Before the change the first argument would simply be `withLibLBI`.

goldfirere:
What happens when we have `-XImpredicativeTypes -XContravariantFunctions` (I've shortened the new extension name a bit)? It means that type inference becomes unpredictable, right? Is that the worst consequence? If wonky type inference is the worst result (as opposed to producing bad Core or something truly horrible) then we do have precedent: `-XGADTs -XNoMonoLocalBinds` leads to similarly wonky inference -- but it's not actually prohibited. (I used that combination once to good effect, even.)

Echoing conversations from other proposals, we do not want to have two extensions that are simply incompatible. Perhaps we don't have a tight specification for their interaction, but combining them shouldn't halt compilation or launch any missiles.

One possible idea here is that `-XContravariantFunctions` disables the ability of quick-look to look to the left of an arrow. That would mean that `runST $` wouldn't work. We could still have `-XImpredicativeTypes` imply `-XNoContravariantFunctions` (so that `runST $` continues to work), but if a user specifies `-XImpredicativeTypes -XContravariantFunctions`, type inference is still predictable. It just puts the power in `-XRankNTypes` instead of in `-XImpredicativeTypes`. Users get to choose where to sprinkle the type inference magic. Happily, supporting these different modes is generally quite easy (I think).

The big minus in these designs is that they are fork-like, as @simonmar describes in https://github.com/ghc-proposals/ghc-proposals/pull/271#issuecomment-529336548. That said, it's quite a small fork.

A related question: given that we have contravariance now and that `runST $` works today, is there some wonky type inference going on? Is there a way to get GHC deeply confused? If not, why not? Maybe the features don't have to be mutually exclusive!

serras:
@goldfirere yes, the only problem is that type inference may be a bit unpredictable, or at least would depend on the order of arguments.

So right now the idea is that:

* `ContravariantFunctions` is enabled by default (but unless you enable `RankNTypes`, there is no way to know about it),
* `ImpredicativeTypes` implies `NoContravariantFunctions`,
* `RankNTypes` + `ContravariantFunctions` works exactly as GHC currently does,
* `ImpredicativeTypes` + `NoContravariantFunctions` performs quick look at the left of an arrow,
* `ImpredicativeTypes` + `ContravariantFunctions` does not look at that part.

About `($)`, it relies on the special typing rule for that case. If you look carefully, `runST $ ...` would actually require `ImpredicativeTypes`, but GHC only asks for `RankNTypes`. Maybe it makes sense to keep that rule anyway, given that `f $ x` is a very common idiom and we really want it to work as `f x`.

goldfirere:
But with `$` and contravariance, is there a way to cause strange behavior? For example, is there a case that should morally be accepted, but GHC gets the type inference wrong somehow? If there exists no case, what is special about `($)` that makes it OK to be impredicative while retaining contravariance?

serras:
The example from the [comment above](https://github.com/ghc-proposals/ghc-proposals/pull/274#issuecomment-530117571) is also relevant here, replacing `f g` by `f $ g`. The possible outcomes are:

* If contravariance is not allowed, that function does not type check, we need to eta-expand `g`.
* If contravariance is allowed, and as a result the left-hand side of the arrows do not take part of quick look, the type cannot be inferred, but we still may add some kind of annotation to help the compiler.

It may be the case that the type of `($)`, in which each type variable appears both in a covariant and in a contravariant position, means that effectively the type is invariant on both variables.

Ericson2314:
> In the current proposal, the subsumption relation itself is restricted.

Understand

> It might be a bit hidden in the draft paper, but there is a way to keep contravariance at the expense of (way) worse inference.....

Ah I see. We want to derive that equality. This sort of thing crops up a lot; e.g. the two different notions of generativity from https://www.imperial.ac.uk/media/imperial-college/faculty-of-engineering/computing/public/1718-ug-projects/Csongor-Kiss-Higher-order-type-level-programming-in-Haskell.pdf . More worryingly, it also reminds me of the `co_nth` stuff @treeowl and I were wrestling with, where it has been annoying for abstraction when a weaker introduction is coupled with a stronger elimination in some context; which means it's unsafe to use the "true" stronger introduction anywhere else.

> To be completely clear: none of these proposals affect the soundness of the type system. It is just about the predictability (changing order of parameters does not affect acceptance, enabling ImpredicativeTypes may cause your program to be rejected) and the strength of type inference.

OK now I am confused. Are you saying there's no way observe some `a -> c <=  b -> c` in a `-XContravariantFunctions` module and eliminate it in a `-XNoContravariantFunctions`, leading to an incorrect `a ~ b` conclusion? That would be a relief!

serras:
@Ericson2314 the co- and contra-variance of functions in a module changes how all functions (both in the module and from other modules) are treated. It doesn't matter for others whether a module has been compiled with or without `-XContravariantFunctions`.

In any case, what may happen with the combination of Quick Look and `ContravariantFunctions` is that the first "quick look" phase guesses an impredicative instantiation which is then found to be wrong in the second, real type-checking pass. It is never the case that a wrong equality is accepted by the type checker.

Ericson2314:
I think I am talking about something different. Suppose we could make a `:<=:` analogous to `:~:`, that witnesses subsumption. If I construct a ` a -> b :<=: c -> d` in a `-XContravariantFunctions` module, and eliminate it in a `-XNoContravariantFunctions` module, I'll get a `a ~ c` proof in the eliminating module that could be untrue.

LukaHorvat:
Would this enable `Generic` instances for types with higher-rank fields? Perhaps I'm mistaken but I think the reason we can't derive `Generic` for a type like `data T = T { field :: forall a. a -> a }` is because the `Rep` would end up with a `K1 (forall a. a -> a)` somewhere.

serras:
@Ericson2314 if we go down that alley, we can get a nice system (we have described it in [our paper our guarded impredicativity](https://www.microsoft.com/en-us/research/publication/guarded-impredicative-polymorphism/)) but unfortunately one that involves big changes to the compiler. The main advantage of Quick Look Impredicativity is that it acts as an additional inference phase, but orthogonal to the usual type checking.

@LukaHorvat I see no problem on your usage, in fact that definition is already possible with the current GHC if you steer the compiler enough:

```haskell
> :t K1 @_ @(forall a. a -> a) (\x -> x)
K1 @_ @(forall a. a -> a) (\x -> x) :: K1 w (forall a. a -> a) p
```

Note that something that `-XImpredicativeTypes` does *not* give us is the ability to define GADTs which refine the index with a polymorphic type. That is, still we cannot do:

```haskell
> data X a where X :: Int -> X (forall a. a -> a)

<interactive>:4:16: error:
    • Illegal polymorphic type: forall a. a -> a
```

serras:
I have pushed a new version of the proposal based on the discussion we have had. It now includes two alternatives for contravariance on function types.

serras:
> You do not discuss this possiblity
> 
>     * Use no-contravariance for the quick-look; but continue to use contravariance for the main type subsumption check.

I have updated the document to discuss this possibility (which is in fact a double possibility, because we could decide to consider left-hand sides of arrows as guarded, or not).

simonpj:
Also let me note that I have never been really happy with the whole idea of supporting contravariance.  It may be convenient **but it changes the semantics of the program**.   It causes implicit eta-expansion which is not semantics-preserving in Haskell.

Let me put it this way: suppose we didn't have contravariance, and someone made a GHC proposal for it as a programming convenience.   I think it'd go down in flames in short order.  Type inference that changes semantics?  Not good!  And yet that's what we have. 

So it's arguable that removing contravariance for higher rank stuff is Fixing A Bug, and we should just, well, fix it.

PS  Yes, I know that I am the original perpertrator of this particular infelicity.

serras:
Particularly, I would strongly prefer to remove contravariance (as @simonpj proposes), or otherwise keep contravariance under a flag. Those are alternatives 1 and 3 in the proposal.

| Alternative | Pros | Cons |
| ----------- | ---- | ------|
| Remove contravariance | "Fixes a bug" (no laziness changes under the hood) | Not backwards-compatible |
| | Keeps unified language (no fork-like proposal) | Needs (mechanical) changes to `Cabal` codebase |
| Contravariance under a flag | Completely backwards-compatible `RankNTypes` | Fork-like proposal (adding a flag may make GHC reject the program) |
| | | The combination contravariance + impredicativity is not formally studied |

How should we proceed here? What is the opinion of the rest of the people interested in this proposal?

serras:
Maybe there is yet another alternative:

* Drop contravariance from function types regardless of the situation,
* Leave a `ContravariantFunctions` flag (disabled by default) to get back contravariance during subsumption. Quick Look would be completely oblivious to this flag.

This means the Bug Is Fixed, while leaving a escapeway for those libraries which would require too many invasive changes (again, I am mostly thinking about Cabal).

simonpj:
> Leave a ContravariantFunctions flag (disabled by default) to get back contravariance during subsumption. Quick Look would be completely oblivious to this flag.

Yes, that's exactly what I was [suggesting above](https://github.com/ghc-proposals/ghc-proposals/pull/274#issuecomment-532130748).   Sounds quite attractive to me.

serras:
I have updated the document with that alternative to contravariance: remove it by default, leave it under a `ContravariantFunctions` flag.

serras:
@nomeata I think that the proposal is ready for the commitee.

Ericson2314:
It just occurred to me that that we could use unidirectional coercions (#198) someday, which would neither get in the way of type inference or be pointlessly costly / observationally observable.

My main worry about loosing contravariance was I want to enforce the PvP someday, I and I wish for the PvP to have a broad definition of subsumption so as to have as few breaking changes as possible. But I don't mind coercing whole modules (!) to do this---it's the sort of thing that explicit PvP enforcement vs implicit PvP enforcement ("does it still typecheck, cross fingers") is actually in a position to do.

isovector:
I read only the proposal, and not the paper, but I'm very excited about this change. Some questions:

The proposal say that `($)` no longer needs special support in the compiler, but it's not clear to me exactly how that works. The _definition_ of `($)` isn't impredicative, only it's instantiation, right? So does that mean I'd need to enable `-XImpredicativeTypes` in any module that wants to use `($)` impredicatively?

gridaphobe:
My understanding is that GHC currently has a special rule to allow impredicative instantiations in applications of `($)`.

serras:
@isovector @gridaphobe Consider a common use of `($)`,

```haskell
runST $ do ...
```

Since `runST :: forall v. (forall s. ST s v) -> v`, using it in this fashion would require `ImpredicativeTypes` for the instantiation of `($)`. However, GHC is fine if you only enable `RankNTypes`. Essentially, the special typing rule allows using `($)` as if it was not there.

Note that before levity polymorphism (when we had `*` and `#` and a supertype of both called `OpenType`) that typing rule was even more special, because `($)` was the only function for which one could substitute a variable by an *unlifted* type.

Still, my patch keeps the rule for `($)` for backwards compatibility (so you don't need to enable `ImpredicativeTypes` if that is the only function which needs it), but in theory it could be removed.

isovector:
@serras got it, thanks. That's unfortunate; letting things like `.` and `flip` be used impredicatively is my #1 biggest excitement for this stuff.

serras:
Hopefully, you should be able to use (.) and flip impredicatively as easily
as ($)! If you have some examples you would like to test, either you or me
can test it in the GHC branch which implements this proposal. But only the
built-in ($) benefits from the special rule "allow impredicative
instantiation even with ImpredicativeTypes off".

isovector:
Dang. Is there any hope of putting an "always allow impredicative instantation" pragma on such things? The idea of putting `foralls` inside lists is exciting, but I definitely want to do it less often than call `flip @(forall x. ...)`. Anyway, I'm happy to take this discussion offline if you feel it's out of scope of scope here.

aspiwack:
Dear all,

I am the shepherd for this proposal. I will take a bit of time to understand this proposal well. And I will report here some time next week. Don't be surprised by my silence until then.

serras:
@isovector this proposal doesn't include any pragma on that direction. But if this proposal gets accepted, feel free to open a new one about that same topic! I would be happy to help and jump in the discussion.

goldfirere:
> Still, my patch keeps the rule for `($)` for backwards compatibility (so you don't need to enable `ImpredicativeTypes` if that is the only function which needs it), but in theory it could be removed.

I don't see this special treatment in the proposal. Should it be there? I vote "yes", but with an appropriate phase-out plan. `runST $ do ...` really is impredicative, and users (in my opinion) should have to enable `-XImpredicativeTypes` (eventually) to do that.

gridaphobe:
Another interesting point is that this particular impredicative instantiation in `runST $ do ...` is easier to avoid now with `-XBlockArguments` (I think that's what it ended up being called). So that gives users another alternative if we decide to remove the special case for `($)`.

adamgundry:
I'd suggest adding a small new extension (`ImpredicativeApplication`?), enabled by default and implied by `ImpredicativeTypes`, that controls whether `$` is magical. That makes it possible for users to explicitly disable the magic (perhaps if their style guide prefers `BlockArguments`). Then later we can consider disabling the magic by default but leave the option of re-enabling `ImpredicativeApplication` without the rest of `ImpredicativeTypes`.

isovector:
I feel like we've probably gotten off-topic here; this proposal is about `ImpredicativeTypes`. Apologies for the derailment, though it's pretty clear the design space is sufficiently open for later proposals around what to do about magic `$` and friends. Let's leave the discussion there until someone wants to write that proposal.

serras:
> > Still, my patch keeps the rule for `($)` for backwards compatibility (so you don't need to enable `ImpredicativeTypes` if that is the only function which needs it), but in theory it could be removed.
> 
> I don't see this special treatment in the proposal. Should it be there? I vote "yes", but with an appropriate phase-out plan. `runST $ do ...` really is impredicative, and users (in my opinion) should have to enable `-XImpredicativeTypes` (eventually) to do that.

There is some [discussion of `($)`](https://github.com/serras/ghc-proposals/blob/quick-look/proposals/0000-quick-look-impredicativity.md#the-typing-rule-for-) in the proposal. The two problems that may arise with doing so are:

* There is potentially a lot of code (about 200 packages which use `RankNTypes`) that could break. But as you say, this has been done before (like the deprecation of `fail` from `Monad`).
* I am not 100% sure about how this interacts with levity polymorphism. The current rule for `($)` contains some notes about this. Right now, in the branch of GHC with quick look impredicativity we augment the current behavior (which essentially amounts to looking at the first argument, and pushing the type to the second), but keep the old code. But it could well be deleted.

proposals/0000-quick-look-impredicativity.md
```diff
@@ -0,0 +1,181 @@
...
+
+In addition, "quick look" only unifies a type variable when it is found under a type constructor different than arrows, because invariance ensures that equality is the only way to make the expression type check. It never tries to perform any complicated analysis on other types: impredicativity must be the *only obvious* solution to make the program type check (or as we say in the paper, "impredicativity is never guessed").
+
+## Examples
+
+Several examples can be found in the [paper draft](https://www.dropbox.com/s/hxjp28ym3lptmxw/quick-look-steps.pdf?dl=0). Let us review the main example, `(\x -> x) : ids`, where `ids :: [forall a. a -> a]` from the eyes of the "approachable" description.
+
+1. We infer the type of the head of the application, `(:) :: forall a. a -> [a] -> [a]`.
+2. We expose two arguments, which leads to the type `alpha -> [alpha] -> [alpha]` where `alpha` is a fresh unification variable.
+3. Now we do *quick look* by checking the "simple" expressions against those types:
```
aspiwack: Instead of relying on quotes, let me suggest you to define a term (be it _simple expression_ or another term) in your specification section. Then you can simply refer to it in your example.

proposals/0000-quick-look-impredicativity.md
```diff
@@ -0,0 +1,181 @@
...
+
+As discussed several times throughout this proposal, its goal is to give a clear and simple specification of impredicative instantiation within GHC. Whereas "simple" is a subjective matter, the availability of a set of typing rules defines a clear specification.
+
+### `TypeApplications`
+
+We deem the interaction with `TypeApplications` as a very important one; the goal is for our specification to benefit from user-written types as much as possible. We want `(:) @(forall a. a -> a) (\x -> x) []` to work, without the need of an additional type application in `[]`. The draft paper details the interaction between those features in depth.
+
+### `RankNTypes` and lack of contravariance
+
+The main disadvantage of making "quick look impredicativity" official as described in the [paper draft](https://www.dropbox.com/s/hxjp28ym3lptmxw/quick-look-steps.pdf?dl=0) is that it collides with *contravariance of function types*. That is, it no longer holds that:
```
aspiwack: What is “described in the paper”? This sentence doesn't make it clear.

proposals/0000-quick-look-impredicativity.md
```diff
@@ -0,0 +1,181 @@
...
+
+`(Int -> Int) -> R is a subtype of (forall a. a -> a) -> R`
+
+The problem is of technical nature, but without this restriction we would not be able to infer impredicative instantiations of functions like `($)`. **This restriction affects not only users of `ImpredicativeTypes`, but also users of `RankNTypes`, effectively making this proposal not 100% backwards compatible.**
```
aspiwack: Explain why contravariance precludes inferring impredicative instanciations of `($)`.

proposals/0000-quick-look-impredicativity.md
```diff
@@ -0,0 +1,181 @@
...
+    • Couldn't match type ‘Bool -> Bool’ with ‘forall a. a -> a’
+      Expected type: (Bool -> Bool) -> Bool
+        Actual type: (forall a. a -> a) -> Bool
+    • In the second argument of ‘choose’, namely ‘g’
+      In the expression: choose f g
+Prelude> :t choose g f
+choose g f :: Bool
+```
+
+There is a simple workaround that works most of the time: eta-expansion (you might need to eta-expand more than once, though). Taking the `choose`, `f`, and `g` functions defined above,in a non-contravariant world we need to eta-expand `f` to make it work: `choose (\x -> f x) g`; now the `x` is assigned type `forall a. a -> a`, which is then instantiated to `Int -> Int` before passing it to `f`.
```
aspiwack: Is eta-expansion a workaround under the current sytem, under this proposal, or both?

proposals/0000-quick-look-impredicativity.md
```diff
@@ -0,0 +1,181 @@
...
+
+### The `ContravariantFunctions` extension
+
+After a preliminary assessment of which packages still compile under this new policy, it looks like most of them do not need it. Except one big and important one: `Cabal` (and anything that uses its API). This library often uses functions of the form:
```
aspiwack: > do not need it

What's “it”?

proposals/0000-quick-look-impredicativity.md
```diff
@@ -0,0 +1,181 @@
...
+
+```haskell
+f :: ( (... -> CabalIO a) -> ...) -> ...
+```
+
+where `CabalIO a` is defined as `HasCallStack => IO a`. That means that whereas before we could pass a function `g` as argument, now we have to do something akin to `\... t ... -> g (\... x ... -> t ... x ...)`, which is quite annoying.
+
+We thus propose to add another extension to GHC, called `ContravariantFunctions`, which should be disabled by default. When the extension is enabled, subsumption checking is done using full contravariance of function types (but during quick look contravariance does not enter the game). This means that `RankNTypes` + `ContravariantFunctions` behaves the same as the old `RankNTypes. This has the benefit of asking people to eta-expand their functions, while having a escape way for those libraries which would require too many changes.
```
aspiwack: If that's part of the proposal, `-XContravariantFunctions` should be in the _Specification_ section. If it is not, it should be in the _Alternatives_ section.

> but during quick look contravariance does not enter the game

I don't understand this sentence. Does it mean that `-XContravariantFunctions` is incompatible with `-XImpredicativeTypes`? Or something else?

proposals/0000-quick-look-impredicativity.md
```diff
@@ -0,0 +1,181 @@
...
+
+### The typing rule for `($)`
+
+Currently, GHC contains a hard-coded typing rule for `($)` (you can find it in [this file](https://gitlab.haskell.org/ghc/ghc/blob/master/compiler/typecheck/TcExpr.hs), looking for `dollarIdKey`) which ensures that `f $ e` works even when `($)` would need to be impredicatively instantiated. The million dollar question is: can we drop this special case from the compiler?
+
+A few versions ago, the answer would have been **no**. The reason was that `($)` was the only function able to operate with both lifted and unlifted kinds. But nowadays, this fact is reflected in its type:
```
aspiwack: > A few versions ago

A few version of what so ago?

proposals/0000-quick-look-impredicativity.md
```diff
@@ -0,0 +1,181 @@
...
+```
+
+where `CabalIO a` is defined as `HasCallStack => IO a`. That means that whereas before we could pass a function `g` as argument, now we have to do something akin to `\... t ... -> g (\... x ... -> t ... x ...)`, which is quite annoying.
+
+We thus propose to add another extension to GHC, called `ContravariantFunctions`, which should be disabled by default. When the extension is enabled, subsumption checking is done using full contravariance of function types (but during quick look contravariance does not enter the game). This means that `RankNTypes` + `ContravariantFunctions` behaves the same as the old `RankNTypes. This has the benefit of asking people to eta-expand their functions, while having a escape way for those libraries which would require too many changes.
+
+### The typing rule for `($)`
+
+Currently, GHC contains a hard-coded typing rule for `($)` (you can find it in [this file](https://gitlab.haskell.org/ghc/ghc/blob/master/compiler/typecheck/TcExpr.hs), looking for `dollarIdKey`) which ensures that `f $ e` works even when `($)` would need to be impredicatively instantiated. The million dollar question is: can we drop this special case from the compiler?
```
aspiwack: Here is a permalink to the typing rule. It's probably better than asking the reader to search: https://gitlab.haskell.org/ghc/ghc/blob/795986aaf33e2ffc233836b86a92a77366c91db2/compiler/typecheck/TcExpr.hs#L368-397

proposals/0000-quick-look-impredicativity.md
```diff
@@ -0,0 +1,181 @@
...
+
+```haskell
+(++) @(forall a. a -> a) ids ([] @(forall a. a -> a))
+```
+
+Finally, we could also enable `ContravariantFunctions` whenever `RankNTypes` is enabled but `ImpredicativeTypes` is not. The main disadvantage is that code that compiles under the former flag may stop compiling if you add the latter.
+
+## Unresolved Questions
+
+* What should be the error messages in those cases in which "quick look" was unsuccessful and did not infer enough information for impredicative instantiation?
```
aspiwack: I'm afraid that I don't understand this question.

proposals/quick-look-list-pkgs.txt
```diff
@@ -0,0 +1,199 @@
+For extensions, based on LTS 13.6:
+RankNTypes, Rank2Types and ImpredicativeTypes
```
aspiwack: Did you intend to commit this file?

proposals/0000-quick-look-impredicativity.md
```diff
@@ -0,0 +1,181 @@
...
+
+A few versions ago, the answer would have been **no**. The reason was that `($)` was the only function able to operate with both lifted and unlifted kinds. But nowadays, this fact is reflected in its type:
+
+```haskell
+($) ::
+  forall (r :: GHC.Types.RuntimeRep) a (b :: TYPE r).
+  (a -> b) -> a -> b
+```
+
+So the initial answer would be **yes**. This should be taken with a grain of salt, though, since the specification in the draft paper does not consider polymorphic type representations. However, for the usual case of `b` having kind `*`, the rule may be dropped. This also means that other combinators such as `(&)` or `(.)` no longer are second-class with respect to impredicativity.
```
aspiwack: I'd like to see, here, an example which works out how `runST $ do {…}` would typecheck without the special rule.

aspiwack: By the way, a discussion which is missing is that in current GHC:

```haskell
> import Control.Monad.ST
> runST $ return 0
0
```

Works without extension. If the special rule for `($)` is removed, then the above will require `-XImpredicativeTypes`.

proposals/0000-quick-look-impredicativity.md
serras: Moved to the "Alternatives" section. The sentence you quoted does not make much sense at that point, so I have removed it.

proposals/0000-quick-look-impredicativity.md
serras: No, but I thought that would be useful during the discussion.

proposals/quick-look-list-pkgs.txt
serras: No, but I thought it would be useful during the discussion.

proposals/0000-quick-look-impredicativity.md
serras: Added discussion to the proposal.

proposals/0000-quick-look-impredicativity.md
serras: Rephrased it. But the main idea is: what should we do with error messages related to impredicative instantiation? can we do better at diagnosing problems and suggesting fixes?

serras:
Thanks @aspiwack for all the comments. I have pushed a new version which should handle all of them.

serras:
The discussion has discovered another possible source of incompatibility. It turns out that GHCi today accepts this *without any extension*:

```haskell
> import Control.Monad.ST
> runST $ return 0
0
```

The [section in the GHC User's Guide](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#extension-RankNTypes) is not very clear about whether you need `RankNTypes` to merely *consume* rank-n types, but the instantiation of `($)` is definitely impredicative.

I would be happy to keep treating `($)` differently in the compiler to this respect (and the current implementation does so). However, from another perspective we are again Fixing A Bug (using @simonpj's terminology) and just asking programmers to enable the right extensions.

goldfirere:
I would favor a transition period with regard to `($)`. Let's pretend that turning on `-XImpredicativeTypes` in, say, GHC 8.8 would immediately stop compilation. (That is, pretend `-XImpredicativeTypes` is a new extension.) Then we would need some time to let everyone enable it without CPP. I think we should do that here, but require `-XImpredicativeTypes` in the future.

An alternative would be to introduce some pragma or other widget on `$` to say that it always allows impredicativity. (I don't like this idea.) Then users could use that to mark their own definitions as wanting impredicative instantiations.

simonpj:
> The section in the GHC User's Guide is not very clear about whether you need RankNTypes to merely consume rank-n types, but the instantiation of ($) is definitely impredicative.

The intention (perhaps seldom made explicit) is that you should require an extension to _define_ a function using an extension, but not to _use_ it.  Eg you can call a rank-n function without having `RankNTypes` on.  Ditto overlapping instances.

For impredicative types, the extension really is to do with _occurrences_; at a particular call site, can we instantaite the function impredicatively. That seems inherently to do with consumption.

A plausible position for `$` is

* Leave today's behaviour unchanged, regardless of flag settings.
* Point out that with `-XImpredicativeTypes` the special-case behaviour is unnecessary, which is good for other higher order functions like `(.)`.

proposals/0000-quick-look-impredicativity.md
```diff
@@ -170,11 +170,27 @@ Another possibility is to drop impredicative type inference altogether, and expe
...
+
+```haskell
+f :: ( (... -> CabalIO a) -> ...) -> ...
+```
+
+where `CabalIO a` is defined as `HasCallStack => IO a`. That means that whereas before we could pass a function `g` as argument, now we have to do something akin to `\... t ... -> g (\... x ... -> t ... x ...)`, which is quite annoying. At the moment of writing, though, only `cabal-doctest` seems to be really problematic, as many other packages in the ecosystem relies on it.
+
+We thus propose to add another extension to GHC, called `ContravariantFunctions`, which should be disabled by default. When the extension is enabled, subsumption checking is done using full contravariance of function types. This means that `RankNTypes` + `ContravariantFunctions` behaves the same as the old `RankNTypes. This has the benefit of asking people to eta-expand their functions, while having a escape way for those libraries which would require too many changes.
```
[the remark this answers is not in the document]
serras: Done. I have moved the last paragraph to the *Proposed Change Specification* section.

atennapel:
A bit off-topic maybe, but would love to hear the opinion of the people involved in this proposal on this idea for impredicative type inference: https://github.com/AndrasKovacs/elaboration-zoo/tree/master/experimental/poly-instantiation, reddit discussion here: https://www.reddit.com/r/haskell/comments/dg1vm7/an_elegant_approach_for_impredicative_polymorphism/.

aspiwack:
Just a quick word to say that I've made my recommendation to the committee. I'll report back here promptly on the outcome of this discussion.

simonpj:
Re the contravariance issue, let me note that

* Both contravariance *and covariance* threaten semantics.  For example, suppose
  ```
  f :: (forall a. Int ->a -> a) -> blah
  g :: Int-> forall a. a->a
  ```
  Is `(f g)` well typed?  Today the asnwer is yes, but only by elaborating to `f (/\a. \x:Int. g x a)`. Note the eta-expansion, which changes the semantics of the argument.

* So I think it is entirely arguable that we should abandon both covariace and contravariance, along with deep skolemisation and deep instantiation.  If we didn't have them, and a GHC proposal proposed to add them, it'd be shot down in flames because type inference changes the semantics!

* We should get data on what the impact of such a change would be.  Jur is doing that.  (At worst it'd require manual eta-expansion.)

* The whole issue is somewhat orthogonal to the core idea of Quick Look.  The more co/contra stuff we have, the less powerful Quick Look is, but regardless it is still jolly useful.

gridaphobe:
A few more comments:

- The current version of the proposal says that the new `-XContravariantFunctions` will be disabled by default. Although the goal may be to remove the extension entirely, I think we should initially enable it by default to preserve the existing behavior. Contravariance is required at the call-site rather than the definition-site, which means that, unlike `-XRankNTypes`, users will pay the cost. It also makes the scope of potential breakage much greater. I think I'm fine with the direction of removing contravariance (though the eta-expansion feels like a hack), I just want us to move a bit slower. Specifically, we should have a proper deprecation cycle with warnings emitted for modules that rely on contravariance.

- Reading the proposal in isolation, my immediate question would be "how well does the Quick Look work in practice?" The discussion suggests that it works extremely well, and that we may not be aware of a single case where Quick Look (- contravariance) is insufficient. This should really go in the proposal itself, as it's a crucial part of the argument for this particular technique. (And if there are any real-world cases where Quick Look fails, the proposal should discuss them.)

- The proposal refers a few times to a draft paper for a more formal specification. I think that's fine, but it would be nice to host the draft some place a bit more permanent than a Dropbox folder. Perhaps upload a copy to Arxiv? Or just include it in this PR?

simonpj:
Concering contravariance, I have written a [new GHC proposal](https://github.com/ghc-proposals/ghc-proposals/pull/287).

gridaphobe:
@serras the committee is considering moving all discussion of contravariance to a separate proposal (#287). If we go that route, it would mean this proposal would introduce QuickLook without `-X[No]ContravariantFunctions`, and it would be a bit less useful due to the interaction with contravariance. `($)` is a prime example of the limitations of QuickLook + contravariant functions, but it of course has a special rule in the type checker. Other function combinators like `(.)` and `flip` would be similarly affected, but I'm not sure how often they are instantiated impredicatively.

Based on your survey of the uses of `-XImpredicativeTypes`, do you think QuickLook would still be useful on its own if we split off the discussion of contravariance?

serras:
I prefer to know the outcome of this proposal + #287 at once, since they complement each other quite well. Also in terms of implementation, as it does not make much sense to have this proposal accepted and implemented, just to only remove contravariance by the other proposal.

Ericson2314:
I suspect the idea with two proposals is first we'll decide costful contravariance is bad and should be removed, and then this will be a shoe-in.

aspiwack:
@serras Here is my reasoning: separating the two issues lets this proposal be evaluated on its own merit (everybody, to be clear, is pretty keen). Without being bogged down in the conversation about `-XContravariantFunctions` and its deprecation policy. Which #287 is, probably, a better place to discuss.

A reviewed proposal would cut the discussion about `-XContravariantFunctions` and let the examples relying on `(->)` invariance (`($)`, `(.)`, …) depend on #287.
nc
A few people, on the committee, believe that this proposal makes a lot of sense on its own, even with `(->)` being invariant (after all, if `(->)` isn't invariant, then the typing rule for `($)` is not principal!).

But if you say: no, it really does need invariant functions, I'll stop pushing. Instead we will probably have to talk about the future of `-XContravariantFunctions`.

It is not to say that you can't implement this proposal and #284 in one fell swoop, were both to be accepted!

simonpj:
I agree with Arnaud. This quick-look proposal makes perfect sense even without invariant functions.  It's just a bit less powerful.  I don't think we need condition one on the other.

And, as he says, we can still implement them together if both are accepted.

serras:
@gridaphobe @aspiwack should I then push a new version of the proposal? should I account on it for the two possible outcomes of #287?

gridaphobe:
Yes, I'd suggest removing the proposed changes around function contravariance, and replacing it with a brief discussion of the limitations that contravariance imposes. A link to #287 noting how it would strengthen this proposal would also be great.

Thanks!

aspiwack:
@serras May I enquire as to your status on this one?

serras:
You can take this as dormant. While developing the formalization further, we have discovered some glitches we are working on resolving. Once the formalization is stable, I'll send a message here to continue with the process.

aspiwack:
Ok. Looking forward to hearing from you again.

serras:
@aspiwack @nomeata I would like to revive this proposal, with our [new version](https://www.microsoft.com/en-us/research/publication/a-quick-look-at-impredicativity/) of the Quick Look impredicativity formalization. Furthermore, since #287 has been accepted, there's no reason to talk about contravariance on arrows in the proposal anymore, we just take it for granted.

aspiwack:
@serras I'll be reviewing next week.

Everyone is, of course, invited to comment on this new version. If the conversation doesn't bring in new elements, I'll make a recommendation to the committee around the 20th of May.

aspiwack:
@serras No activity, so I will recommend acceptance as soon as I'm convinced that all the piece are there. I'm running a bit late on my review, though, quite sorry about that. Can you help me out a bit and summarise what didn't work in the previous iteration and how you changed it?

PS: I can only encourage you to add a link to #287 somewhere (I can't see one): it's useful documentation, you probably want to explain that (some of) your example depend on this new rule. Remember that the proposal stays around after it has been accepted. So all the design consideration should appear in the text as much as possible.

serras:
It was not that the previous one didn't work, as much as during the review+rewriting process from PLDI to ICFP we found a dramatic simplification of the type rules. The core idea is still the same: having a quick look at the arguments before doing the actual type checking, and only allowing impredicative instantiation during that phase. However, the little details -- which are relevant for the implementation -- are much simpler now.

@simonpj gave a talk explaining the general idea and the most important details -> https://www.youtube.com/watch?v=ZuNMo136QqI

serras:
@aspiwack I have added a mention to #287 as you requested. I have also updated the fact that @simonpj has an implementation of the proposal in [MR 3220](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3220).

proposals/0000-quick-look-impredicativity.md
```diff
@@ -0,0 +1,146 @@
...
+
+One important feature of Haskell's type system that "quick look" uses is the invariance of type constructors. In short, the subsumption rules ensure that if, for example, `Maybe t` is a more polymorphic than `Maybe s`, it must be the case that `t` equals `s`. This property holds for every type constructor except for function types, which require a slightly more complex handling. The proposed inference algorithm never tries to perform any complicated analysis on other types: impredicativity must be the *only obvious* solution to make the program type check (or as we say in the paper, "impredicativity is never guessed"). Note that since [proposal 287, *Simplify Subsumption*](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0287-simplify-subsumption.rst) has been accepted, function types are also considered invariant.
```
aspiwack, line 71: Is it actually the case that function types require more complex handling? Or are they handled uniformly thanks to #287 ?

proposals/0000-quick-look-impredicativity.md
```diff
@@ -0,0 +1,146 @@
...
+
+1. We infer the type of `id` by merely looking up in the environment, `forall a. a -> a`,
+2. We expose one argument, and for that we instantiate the type above with a new type variable, leading to `alpha -> alpha`,
+3. Now we check the argument `True` against `alpha`:
+   a. We infer the type of `True`, which is `Bool`,
+   b. Steps (2) and (3) are not needed here, since there are no arguments,
+   c. We perform the subtype check `alpha <= Bool`, leading to the unification `alpha := Bool`;
+4. The result of the inference is `alpha`; after zonking the result is `Bool`.
+
+When `ImpredicativeTypes` is on we introduce an additional step between (2) and (3), which we call *quick look*. Quick look traverses the arguments of the application, and tries to infer as much impredicative instantiations as possible. The results are applied before step (3), which means that arguments are type checked with the correct impredicative instantiations. The "quick look" pass only traverses *simple expressions*, which we define to be (possibly nested) applications of variables in the environment.
```
aspiwack, line 67: What's missing here is a summary of how the quick-look pass infers impredicative instantiations. It's find to punt to the paper for details. But, as a programmer, I shouldn't need to refer to the paper to build an intuitive understanding of the typing rules.

I think it's reasonable to apply the rule of thumb that the proposal should contain enough information for a programmer to use as its main documentation.

proposals/0000-quick-look-impredicativity.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Effect and Interactions
+
+As discussed several times throughout this proposal, its goal is to give a clear and simple specification of impredicative instantiation within GHC. Whereas "simple" is a subjective matter, the availability of a set of typing rules defines a clear specification.
+
+### `TypeApplications`
+
+We deem the interaction with `TypeApplications` as a very important one; the goal is for our specification to benefit from user-written types as much as possible. We want `(:) @(forall a. a -> a) (\x -> x) []` to work, without the need of an additional type application in `[]`. The draft paper details the interaction between those features in depth.
+
+### The typing rule for `($)`
```
aspiwack, line 96: Are you or are you not proposing that the `($)` rule be dropped? This section (and the corresponding alternative) don't unclear to me.

proposals/0000-quick-look-impredicativity.md
serras: I have removed the sentence about arrows. As you say, since #287 they are treated as any other type.

proposals/0000-quick-look-impredicativity.md
serras: I have reworded this a bit. I propose to remove the rule. However, I want to make clear that due to GHC's lenient behavior, things like `runST $ ...` which only required `RankNTypes` now may require `ImpredicativeTypes`.

proposals/0000-quick-look-impredicativity.md
serras: I had tried to describe the workings intuitively as "two passes": one over simple expressions where impredicativity can be inferred, and a second one where everything is checked, but impredicativity can no longer be inferred.

@aspiwack @simonpj maybe you could help me in rewording this part?

simonpj:
I have done some rewording to address Arnaud's questions.  Arnaud, ok?

phadej:
I kindly ask to add `QuickLookImpredicativity` language extension (alias) when this proposal is implemented.

GHC has `--supported-extensions` feature which is great for feature discovery. Yet if the meaning of old extensions changes, without any other hint than GHC version, it greatly reduces value of `--supported-extensions`.

`--supported-extensions` list is used by Cabal solver together with `other-extensions:` `.cabal` field declarations to find better install plans, or to give good solver errors ("your compiler doesn't support `QuickLookImpredicativity`). If there (and there probably) will be packages which rely on `QuickLookImpredicativity` power, please let the package authors specify that declaratively and precisely, and not by mere opaque "you need GHC >=9.2". It's better to be able to say "you need compiler with `QuickLookImpredicativity`".

simonpj:
> I kindly ask to add QuickLookImpredicativity language extension (alias) when this proposal is implemented.

Ok, so the proposal as written, and accepted, changes the (unspecified) meaning of the existing `-XImpredicativeTypes`. I believe that you are suggesting:

1. Add `-XQuickLookImpredicativity`
2. Make `-XImpredicativeTypes` be a synonym for `-XQuickLookImpredicativity`
3. Deprecate `-XImpredicativeTypes` and eventually remove it
4.  I'm not sure whether or not you want to remove `ImpredicativeTypes` from `--supported-extensions`, and if so when.  (It has never been "supported".)

At least I'm not sure if you are proposing (3), but it seems funny to have two names for the same thing, and would encourage any library that uses the existing `-XImpredicativeTypes` to review how it is used.

All of this is OK with me, as an amendment to the existing proposal.   (You should turn it into a PR for the existing proposal, so we end up with a final proposal that says what we mean.)

But of course it is extra work.   Perhaps people can post here if they think the work is justified -- that is, that it would make their lives better?

phadej:
> but it seems funny to have two names for the same thing

Depending on how much you want to break things. An example is `TypeInType` which is mere alias nowadays as well (and IMO somewhat obscure alias, "more powerful" `PolyKinds` + something I don't ever remember).

monoidal:
I can see the rationale for the new name, but I'd like to point at some downsides.

Until QL, `ImpredicativeTypes` was never really officially "supported". The manual was pretty clear that it can change between versions of GHC, and people who use it should expect breakage between versions.

In GHC 9.0, there will be a LinearTypes extension. However, this extension is just its initial version and I fully expect GHC 9.2 to have better support and be able to compile more programs. I would not want to be adding LinearTypesV2 extension.

We generally don't use algorithm names in extensions - I'd consider this implementation detail. For example, GADTs, not OutsideIn; DependentHaskell, not Bake.

simonpj:
@phadej say yes, @monoidal  says no.  We need to hear from more people.

phadej:
I don't feel strongly about this. I consider major changes in extensions functionality to be undesirable, but I have no better option than coming up with new names.

I can only say that `LinearTypes` is not a good example. It's a fresh extension.

If GHC-9.2 would change how `OverloadedStrings` or `OverloadedLists` (e.g. to solve `bytestring` ASCII-ness problem or to allow `Vec` creation, respectively), that would be problematic, won't it?

serras:
One problem is that once Quick Look impredicativity is merged, the code for the old `ImpredicativeTypes` will be gone. In fact, #287 (on top of which this proposal builds) is not backwards-compatible, and at that point no additional extension was added.

Otherwise, what should we go? Leave two different implementations of `ImpredicativeTypes`, one supported and formalized, and the other officially discouraged?

simonpj:
I would argue very strongly indeed against maintaining (in any form) the current un-specified, unprincipled, and generally bogus implementation of impredicativity.  We've advertised for years that it's an unsupported extension and might change without notice.

In contrast I'm totally relaxed about what name to use for the extension flag that wakes up Quick Look.

Ailrun:
What about `ImpredicativeInference`? I have thought the `ImpredicativeTypes` is bit vague, as the problem of impredicativity occurs only when a user doesn't specify type explicitly.

Ericson2314:
@Ailrun but without this extension one cannot even specify it manually with e.g. a `@(foral ....)` type application.

goldfirere:
I think we're conflating two compiler features that needn't be the same:

1. The feature allowing users to declare what language they are writing in.

2. The feature allowing tools and users to query what features a compiler supports.

Right now, both (1) and (2) are done by using extension names. This is a reasonable design, as (1) and (2) are closely related. But they need not be the same mechanism. For example, we could imagine `--supported-extensions` producing extra information for each extension. This could either be a simple version number or it could be a list of capabilities. The version number is simpler. But the list of capabilities is nice in that we could simply extend the list when an extension becomes stronger, or remove a capability when appropriate. We could even append capabilities with version numbers, allowing fine-grained control. These capabilities would be listed in cabal files -- but not in source files. (It's actually quite like how we specify packages in cabal files but not in source files. That is, the cabal file provides some more global truth than the source file does.)

As regards this particular extension name: I'd very much like to keep `-XImpredicativeTypes`. The fact that it's implemented with Quick Look is something of an implementation detail. (I say "something of" because the detail leaks into the specification.)

simonpj:
> GHC has --supported-extensions feature which is great for feature discovery. Yet if the meaning of old extensions changes, without any other hint than GHC version, it greatly reduces value of --supported-extensions.

Was `-XImpredicativeTypes` in "supported extensions"?  It should not have been.  The user manual explicitly described it as not-supported.

phadej:
> Was -XImpredicativeTypes in "supported extensions"? It should not have been. The user manual explicitly described it as not-supported.

```
% ghc-7.0.4 --supported-extensions|grep Impre
ImpredicativeTypes
NoImpredicativeTypes
```

```
% ghc-7.8.4 --supported-extensions|grep Impre
ImpredicativeTypes
NoImpredicativeTypes
```

```
% ghc-8.8.4 --supported-extensions|grep Impre
ImpredicativeTypes
NoImpredicativeTypes
```

PR #276, opened 2019-09-17
Remove the Coercible decomposition rule

treeowl:
Remove the rule that coercibility of `data` types implies
coercibility or equality of some of their parameters.

[Rendered](https://github.com/treeowl/ghc-proposals/blob/no-coercible-decomp/proposals/0000-no-coercible-decomp.md)


Ericson2314:
Nice! So here's an idea from when I was last thinking about this: `nthCo` is only used for the push rule for `case`, so what we really want is not to decompose aggregate coercions into *parameter* coercions, but rather *field* coercions. This is great as any coercion must imply the field coercions to be sound, and we only need to push through a case (and know fields) when we know the constructors being matched. So the core rules and surface rules can actually be similar and both hinge on constructor opacity.

Ericson2314:
Incidentally my proposal is basically the dual of all that: introducing coercions from the (potentially weaker) field coercions rather than parameter coercions, but I think this proposal is the better first step. If the new elimination rules are "obviously sound" and the introduction rules are the same, then we still have a sound system partially liberated from parameter roles. After, my proposal is left as a smaller second step.

treeowl:
@Ericson2314, oh, that's interesting. Do you have a suggestion for modifying my proposal? It sounds like you do! Perhaps you could give a couple more details, or open a PR against my branch?

One cool thing about the simple "just yank them out" approach is that I believe it removes one potential source of infinite loops in the type checker. Pushing them in to fields, on the other hand, has the potential to make things *more* complicated. How do you know when to stop? Roughly following what goes on in Core, we could expose the decomposition-induced coercion axioms under a pattern match, but I'm still not so sure that buys us enough comprehensible power.

goldfirere:
> Treat types as having their inferred roles wherever all their constructors are in scope.

I don't know how to do this without losing type safety. Sure, we can drop an axiom from Haskell, but we need that axiom in Core, as the proposal admits.

I'm afraid the burden is on you to prove this new version of roles type-safe.

phadej:
isn't `mapMonotonic (coerce :: Int -> Down Int) = coerce` just wrong, i.e. `coerce` isn't always monotonic.

For me this demotivates whole proposal.

Ericson2314:
@phadej `mapMonotonic (coerce :: Int -> Down Int) = coerce` is correct because the two sides do the same thing. Either they both are correct or they both are wrong.

Ericson2314:
WIP: OK page 53 we need a rule for `(->)` for `S_Push`, since it doesn't have a data constructor. It can be a restricted form of `nth`, since there really isn't anything better to do. (For dependent functions there is, but that's out of scope.) For `S_CPush` we need the same but for `=>`.

soundlogic2236:
I argued further for a limited version of this here: https://discourse.haskell.org/t/the-current-coercible-rules-are-dangerously-misleading/10563/26

PR #277, opened 2019-09-19
Winapi pseudo calling conv

Mistuke:
Re-opening of #66 , github seems to not allow me to re-open the PR.


Mistuke:
[Rendered](https://github.com/ghc-proposals/ghc-proposals/blob/0b23ba94c76d4e279aa1e872947151edbbbfed93/0000-winapi-callconv.rst)

ndmitchell:
This is an awesome idea. You aren't wrong that this code crops up all the time, e.g. https://github.com/ndmitchell/shake/blob/master/src/Development/Shake/Internal/FileInfo.hs#L126-L133. I think the majority of people doing this get it wrong, so it probably brings conciseness and correctness benefits.

Mistuke:
@nomeata Since this is a continuation from #66 I think it's ready.

isovector:
Is there a reason for exporting with winapi to not just be an error, rather than a warning? Besides that point I'm in favor.

ndmitchell:
What do CALLBACK values require? They are stdcall on 32 bit, and I assume fastcall on 64bit? If so isn't there a very strong need to have winapi on exports?

Mistuke:
callback functions are created with `import1 `wrapper` functions as they need to be pinned by the GC. so export is not needed for them.

I didn't prohibit exports as there may be legitimate uses for it, but also the Haskell report does not seem to allow for "import only" calling conventions. So having it only be `import` would then require this to be a language extension afaik.

I'm also perfectly happy not to emit a warning for export as the semantics are sound.

ndmitchell:
I think no warning makes sense. It's a consistent semantics and could be useful in some cases.

Ericson2314:
Can't a non-closure callback be created with an export?

Mistuke:
> Can't a non-closure callback be created with an export?

I don't quite follow... My understanding of a callback is a function pointer. You can export a function but that's not a function pointer and doesn't have the same semantics. They differ in e.g. what happens when an exception occurs and when GC may happen and what the GC is able to remove.

example 

> Note that a callback into the Haskell system implies that a garbage collection might be triggered after an external entity was called, but before this call returns. Consequently, objects other than stable pointers (cf. Section 36) may be moved or garbage collected by the storage manager. 

The Haskell report has strict rules for which types are acceptable in an FFI declaration https://www.haskell.org/onlinereport/haskell2010/haskellch8.html#x15-1690008.6 and importantly it doesn't allow higher-ordered functions or Functions inside data-types. 

In particular the set of basic datatypes allowed in between an arrow `->` are

```
8.4.2 Foreign Types

The following types constitute the set of basic foreign types:

    Char, Int, Double, Float, and Bool as exported by the Haskell Prelude as well as
    Int8, Int16, Int32, Int64, Word8, Word16, Word32, Word64, Ptr a, FunPtr a, and StablePtr a, for any type a, as exported by Foreign (Section 24).
```
So the only way to provide a function pointer to the foreign world, or accept a function pointer from the outside world is through a `FunPtr` and the documentation of `FunPtr` and the report specify how to create a `FunPtr` https://hackage.haskell.org/package/base-4.12.0.0/docs/Foreign-Ptr.html#t:FunPtr  i.e. only through the `import` statement's `wrapper` and `dynamic` methods.

`dynamic` for when you get a callback function from a `C` function, (i.e. one you got from an `import` to a C function returning a callback) and `wrapper` is for when you want to pass a callback to a `C` function. (i.e. what you get when you call a C function expecting a callback or providing an export to a Haskell function returning a callback).

In both cases the actual callback `must` be created though an `import`.

As far as I'm aware, there is no other way to create or use callbacks that have a well defined semantic.

Ericson2314:
https://stackoverflow.com/questions/48892763/returning-a-struct-from-a-haskell-ffi-callback the question is unrelated but look at the export + import at the bottom. I'd think this sort of callback would be faster because there's no trampoline or whatever it does to snuggle in a closure.

But none of this matters as your proposal now points out another use-case for `foreign export`.

Mistuke:
That example just proves my point, the FunPtr which is the callback is created using an import wrapper. You have to. It's the only way to get the right rts support. The the wrapper is the one that need the right calling convention not the exported function.

What you're describing is is just a trampoline. And I don't think it's actually faster at all. You just introduce an extra indirection.

So you still don't need `winapi` for that export case. And that question is factually wrong. Haskell's FFI supports returning structs and unions just fine. That's what `Storable` is for.

isovector:
Thanks for the explanation, @Mistuke. I'm happy to recommend this for acceptance.

Mistuke:
Thanks @isovector!

nomeata:
Pinging @ghc-proposals/ghc-steering-committee just for good measure

Ericson2314:
[It's no longer relevant to this proposal but I'll try to clarify the situation for sake of the GHC docs elsewhere. I think the export is doing all the work, the import should just box the function's address into a `FunPtr`. Compare if one called the higher order function in C with the address of the export. No Haskell-specific magic could happen in this case.]

Mistuke:
> [It's no longer relevant to this proposal but I'll try to clarify the situation for sake of the GHC docs elsewhere. I think the export is doing all the work, the import should just box the function's address into a `FunPtr`. Compare if one called the higher order function in C with the address of the export. No Haskell-specific magic could happen in this case.]

The implementation of `dsFExportDynamic` disagrees with you. The wrapper is what provides the needed code to get an rts capability and actually create and schedule the TSO. 

The documentation of the wrappers literally says:

> @foreign import "wrapper"@ (previously "foreign export dynamic") lets
> you dress up Haskell IO actions of some fixed type behind an
> externally callable interface (i.e., as a C function pointer). Useful
> for callbacks and stuff.

This is also my last comment on the subject as this is getting nowhere.

hsyl20:
> And that question is factually wrong. Haskell's FFI supports returning structs and unions just fine. That's what `Storable` is for.

Note that on x86-64, small structures (< 32 bytes) may be passed into SSE registers: https://uclibc.org/docs/psABI-x86_64.pdf (page 18-19)

isovector:
@nomeata I think we're good to merge this! It's been a week now, and there has been no dissent.

nomeata:
Will do. Will you send the mail to the list?

isovector:
Done. Thanks @nomeata !

PR #278, opened 2019-09-26
Better Custom Type Errors

isovector:
The `GHC.TypeLits.TypeError` machinery is much too eager to emit `TypeError`s.
This proposal loosens them up significantly so that they:

* can produce better error messages
* don't get in the way of more important type errors
* aren't emitted at definition sites

[Rendered](https://github.com/ghc-proposals/ghc-proposals/blob/5edc1ab7849dc482bc206d7b8db914a531133c7d/proposals/0000-type-errors.md)

### DIFF
--- a/proposals/0000-type-errors.md
+++ b/proposals/0000-type-errors.md
@@ -1,0 +1,343 @@
+---
+author: Sandy Maguire
+date-accepted: ""
+proposal-number: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/0>).
+**After creating the pull request, edit this file again, update the number in
+the link, and delete this bold sentence.**
+
+# Better Custom Type Errors
+
+The `GHC.TypeLits.TypeError` machinery is much too eager to emit `TypeError`s.
+This proposal loosens them up significantly so that they:
+
+* can produce better error messages
+* don't get in the way of more important type errors
+* aren't emitted at definition sites
+
+This proposal is motivated by [a lot][lots] of experience with `TypeError`s
+going wrong.
+
+[lots]: https://github.com/polysemy-research/polysemy/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3A%22custom+errors%22+
+
+
+## Motivation
+
+`TypeErrors` are littered with non-compositionality, and seem to explode at
+every chance they get. Here are a few aggressively use-cases:
+
+### TypeErrors are emitted at their definition sites, not their use sites
+
+The following module will not compile:
+
+```haskell
+module Example where
+
+type MyError = TypeError ('Text "This module doesn't compile")
+```
+
+(instead you'll see the error `This module doesn't compile`)
+
+This is not just a contrived example; even [boot libraries run into this
+problem.][containers]
+
+[containers]: https://github.com/haskell/containers/blob/403cf3cb7ef365961269cebe9d4eeb7221edd927/containers/src/Utils/Containers/Internal/TypeError.hs#L17-L52
+
+The eagerness of `TypeError` means we can't easily provide good error messages
+for default implementations of classes (reported as a [GHC bug][bug]):
+
+[bug]: https://gitlab.haskell.org/ghc/ghc/issues/16906
+
+```haskell
+class Foo a where
+    foo :: a
+    default foo :: TypeError
+        (Text "Please define foo manually. Suggested definitions:" :$$:
+         Text "  foo = something complicated" :$$:
+         Text "  foo = something else complicated")
+        => a
+    foo = error "unreachable"
+```
+
+The behavior of `TypeError` today prevents this *class* from compiling, rather
+than triggering on *instances* which forget to implement the `foo` method. Many
+libraries provide several different generic means of implementing their class
+methods --- because there is a choice, `DeriveAnyClass` can't be used, but it
+would be helpful to inform the user of their options. [Early versions of
+`polysemy`, for example, had multiple defaults.][polysemy]
+
+[polysemy]: https://github.com/polysemy-research/polysemy/blob/b23f7d2561e2be7440f124c65218247f3afe849e/src/Polysemy/Internal/Effect.hs#L85-L124
+
+There is a workaround for all of the above, but it depends on an undocumented
+implementation detail of how type families expand.
+
+
+### TypeErrors get in the way of good, useful errors
+
+When used in value-level code, `TypeError`s will often *prevent legitimate type
+errors* from being emitted. Consider the following encoding of a row type:
+
+```haskell
+type family Lookup (needle :: k) (haystack :: [k]) where
+  ...
+
+type Member effect row = Lookup effect row
+```
+
+We can write some code that uses the type level list:
+
+```haskell
+get :: Member (State s) row => Effect row s
+get = ...
+```
+
+and attempt to use it:
+
+```haskell
+wrong1 :: Effect row Int
+wrong1 = get
+```
+
+Here we'd like to get a custom error message saying `You should try adding
+a (Member (State Int) row) constraint`, but instead we'll see something about a
+stuck `Lookup effect row` constraint.  There is [a trick][detecting] for
+producing a type error iff some type is stuck, which we can do here:
+
+[detecting]: https://kcsongor.github.io/report-stuck-families/
+
+```haskell
+type MyUsefulTypeError needle haystack =
+  ... something about `'ShowType haystack`
+
+type Member needle haystack =
+  ErrorIfStuck (MyUsefulTypeError needle haystack) (Lookup needle haystack)
+```
+
+and this appears to work, in the sense that it will give the right error message
+for `wrong1`. But it has a dreadful implication --- it will completely suppress
+real type errors. Consider `wrong2`:
+
+```haskell
+wrong2 :: Member (State (Maybe Int)) r => Effect row Bool
+wrong2 = isJust $ get @(Maybe Int)  -- should be `fmap isJust get`
+```
+
+Unfortunately `wrong2` will also emit `MyUsefulTypeError` rather than producing
+the usual `Could not match 'Maybe Int' with 'Effect row (Maybe Int)'` type
+error! Not only this, but `MyUsefulTypeError` will also be emitted instead of
+typed holes, completely destroying hole-driven development workflows!
+
+**There is no workaround for this problem.** We are stuck choosing between
+terrible type errors that leak implementation details, and nice type errors that
+completely squelch GHC's usual errors.
+
+
+
+## Proposed Change Specification
+
+There are three separate pieces to the proposed change. The first two correspond
+to the issues raised above, and the third is for quality of life improvements
+since we're here anyway.
+
+### Part 1) Don't Emit User-Written Type Errors
+
+`TypeError`s will no longer be emitted when appearing in user-written types.
+Instead, they will be emitted whenever they appear in *inferred* types.
+
+#### Part 1.5) A note on backwards compatibility
+
+The original use case for `TypeError` was as a genuine error case in a `type
+family` match. For example:
+
+```haskell
+type family Pred (n :: Nat) :: Nat where
+  Pred 0 = TypeError ...
+  Pred n = n - 1
+```
+
+When a user writes `Pred 0` monomorphically directly in a type, we *probably
+still want to emit it,* despite the proposed changes above. To that end, we will
+also add two more rules for when to emit type errors:
+
+1. Whenever a type family reduces to a `TypeError`
+2. Whenever a type family scrutinizes a `TypeError`
+
+Rule 1 gives us today's behavior for eagerly reporting `TypeError`s that result
+as genuine errors in type family expansion.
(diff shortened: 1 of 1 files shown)


treeowl:
I haven't read through this whole thing yet, but I have also encountered the *opposite* problem. That is, GHC will be trying to show that `a ~ MyFamily b`, hit a `TypeError` in `MyFamily b`, and just tell me it can't show `a ~ MyFamily b`. Grrrr.... See [this comment](https://github.com/love-haskell/coercible-utils/blob/0c634f98af9e492960521e481dcb781759198f46/src/CoercibleUtils/Newtype.hs#L107) for an example.

phadej:
Any improvements with what we can do with `TypeError`s are welcome.

On the other hand, I'd like the better motivations. Changing how
type-checker works should be motivated well.

## TypeErrors are emitted at their definition sites, not their use sites

There are various ways to deal with an issue.
Simplest one is to

```haskell
type MyErrorMessage = 'Text "This module doesn't compile"
```

You should also discuss relation with other unsatistifiable constraints:
To repeat

```haskell
{-# LANGUAGE ConstraintKinds, GADTs, DataKinds #-}
import GHC.TypeLits
type Wrong = TypeError ('Text "err")
```

errors

```
GHCi, version 8.8.1: https://www.haskell.org/ghc/  :? for help
Loaded GHCi configuration from /home/ogre/.ghci
[1 of 1] Compiling Main             ( TypeAlias.hs, interpreted )

TypeAlias.hs:5:1: error:
    • err
    • In the type synonym declaration for ‘Wrong’
  |
5 | type Wrong = TypeError ('Text "err")
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

That might look inconstent, because if you do

```haskell
{-# LANGUAGE ConstraintKinds, GADTs #-}

type Wrong = Int ~ Char
```

it doesn't error.

You can also define

```haskell
foo :: Wrong => a
foo = foo
```

and it wont' error either. But for obvious reasons it's impossible to use
that function:

```
GHCi, version 8.8.1: https://www.haskell.org/ghc/  :? for help
Loaded GHCi configuration from /home/ogre/.ghci
[1 of 1] Compiling Main             ( Equality.hs, interpreted )
Ok, one module loaded.
*Main> foo

<interactive>:1:1: error:
    • Couldn't match type ‘Int’ with ‘Char’ arising from a use of ‘foo’
    • In the expression: foo
      In an equation for ‘it’: it = foo
```

Yet, there's still a way to work with `TypeError`: type families.

```haskell
{-# LANGUAGE ConstraintKinds      #-}
{-# LANGUAGE DataKinds            #-}
{-# LANGUAGE GADTs                #-}
{-# LANGUAGE TypeFamilies         #-}

import Data.Kind
import GHC.TypeLits

type family Wrong :: Constraint where
    Wrong = TypeError ('Text "err")

foo :: Wrong => a
foo = foo
```

loads into GHC nicely. And if one tries to use it, we get a `TypeError`:

```
GHCi, version 8.8.1: https://www.haskell.org/ghc/  :? for help
Loaded GHCi configuration from /home/ogre/.ghci
[1 of 1] Compiling Main             ( TypeFamily.hs, interpreted )
Ok, one module loaded.
*Main> foo

<interactive>:1:1: error:
    • err
    • In the expression: foo
      In an equation for ‘it’: it = foo
```

So if you use type family you can alias `TypeError`s.
I kind of understand the difference between `type` and `type families`
behaviour, but not deeply.

### Default implementations

I don't understand, if I have

```haskell
{-# OPTIONS_GHC -Werror=missing-methods #-}

class Foo a where
  foo :: a -> a

instance Foo Bool
```

errors with built-in error.

```
GHCi, version 8.8.1: https://www.haskell.org/ghc/  :? for help
Loaded GHCi configuration from /home/ogre/.ghci
[1 of 1] Compiling Main             ( Class.hs, interpreted )

Class.hs:6:10: error: [-Wmissing-methods, -Werror=missing-methods]
        ‘foo’
    • In the instance declaration for ‘Foo Bool’
  |
6 | instance Foo Bool
  |          ^^^^^^^^
Failed, no modules loaded.
```

I think that making

- `-Werror=missing-method` the default behaviour
- Somehow being able to amend built-in errors/warnings with extra hints (how to define the default implementation)

is the better way forward. Trying to tweak `TypeError` to do something which arguably GHC should do better itself is IMO wrong way to improve the situation.

### TypeErrors get in the way of good, useful errors

The example is too complicated, does the below one highlight the issue:

```haskell
{-# LANGUAGE ConstraintKinds #-}
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE ScopedTypeVariables  #-}

import Data.Kind
import Data.Proxy
import GHC.TypeLits

type family IfOK (b :: Bool) :: Constraint where
    IfOK 'True  = ()
    IfOK 'False = TypeError ('Text "there's an error")

type AliasOfIfOK b = IfOK b

fun :: AliasOfIfOK b => Proxy b
fun = Proxy

use :: forall (b :: Bool). Proxy b
use = fun
```

errors with

```
GHCi, version 8.8.1: https://www.haskell.org/ghc/  :? for help
Loaded GHCi configuration from /home/ogre/.ghci
[1 of 1] Compiling Main             ( Simpler.hs, interpreted )

Simpler.hs:22:7: error:
    • Could not deduce: IfOK b arising from a use of ‘fun’
    • In the expression: fun
      In an equation for ‘use’: use = fun
    • Relevant bindings include
        use :: Proxy b (bound at Simpler.hs:22:1)
   |
22 | use = fun
   |       ^^^
```

I'm not sure that using `ErrorIfStuck` **hack** is the right solution
to begin with.

I'm thinking that in this case when type-checker needs to satisfy
a constraint which isn't in a weak-head-normal-form, i.e
it's some unreduced type family -- and there's nothing else it can do,
there could be away to report an additional hint to built-in error,
like in previous case.

So in non-existing syntax:

```haskell
type family IfOK (b :: Bool) :: Constraint where
    IfOK 'True  = ()
    IfOK 'False = TypeError ('Text "there's an error")
    {-# warn-if-stuck IfOK b = 'Text "there's another error" #-}
```

treeowl:
One problem I've encountered is that in many cases a type error occurs because a type is ambiguous or too polymorphic to allow a type family to reduce or an instance to be

treeowl:
One big problem I have with the current mechanism is that it gives us *nothing* when we can't match on the types involved, either because they're ambiguous or because they're too polymorphic. (Yes, it's sometimes possible to hack around this with overlapping instances, as @kcsongor has shown, but that has some nasty knock-on effects.) Very often, I want something a bit more in the nature of a stack trace. Vaguely speaking, I may want to register a special interest in whether a particular constraint can be resolved. Rather than *just* reducing it to a set of wanteds, I want to say "If you can't satisfy this constraint, I want to know explicitly."

```haskell
class C m => D m
instance Q x => D (P x)

f :: ICareAbout (D m) => ...
-- or even
f :: D m `OnFailure` ('Text "Whoops! " :<>: 'ShowType m) => ...
```

If GHC tries to resolve `D (P x)` and can't satisfy `Q x` or `C (P x)` for whatever reason, I explicitly want a (possibly custom) error about `D (P x)`.

isovector:
Thanks for all of your feedback!

@treeowl this is definitely an idea I'm interested in. It's strongly reminiscent of @serras's [`WithMessage`](https://github.com/serras/ghc-proposals/blob/master/proposals/0000-custom-type-errors.rst#custom-type-errors). It's a cool idea, but as the proposal points out, now you want to be able to order when those errors are emitted.

@phadej thanks for all of your input! Let me try to tackle as much of it as I can page in my brain.

> There are various ways to deal with an issue.

Absolutely we can get around these things, but it's tedious and non-composable and I have yet to meet someone get any of it right on their first try. That being said, I'll make sure to point out this in the `Alternatives` section.

> I kind of understand the difference between type and type families behavior, but not deeply.

There isn't a meaningful implementation behind this. The check as it's written is performed only as part of `reportWanteds`, which doesn't expand type families. For whatever reason, `type` synonyms hit this codepath, and `type instance`s do not. Though as best I can tell, today's behavior is _completely accidental._

> Somehow being able to amend built-in errors/warnings with extra hints (how to define the default implementation)

On one hand, I agree it would be better to have more specialized machinery for this. On the other hand, `-XDefaultSignatures` and a well-behaving `TypeError` are sufficient to implement this for ourselves --- and indeed is a completely sensible implementation!

> The example is too complicated

Thanks for calling this out. In the process of trying to justify it, I came up with this much simpler example:

```haskell
foo :: Deferred ('Text "blah") => ()
foo = ()

-- Doesn't compile. Error message is `blah` rather than `Expected type String, but actual type ()`
bar :: String
bar = foo
```

> I'm not sure that using ErrorIfStuck hack is the right solution to begin with.

Strongly agreed. I like your pragma idea, especially if it can be attached to any type/type synonym --- I often want to determine whether `a ~ b` is stuck, for example. Is your suggestion that the `warn-if-stuck` message is emitted whenever the constraint should be discharged, but is stuck? I'm not exactly sure what the semantics would be here, but I do like the direction.

phadej:
> On the other hand, -XDefaultSignatures and a well-behaving TypeError are sufficient to implement this for ourselves --- and indeed is a completely sensible implementation!

That will hide the built-in GHC error. I'm personally strongly against completely replacing compilers' built-in errors. People (and tools!) learn to recognize them.

---

A comparison to `Deferred` example is

```haskell
{-# LANGUAGE GADTs #-}

foo :: Int ~ Char => ()
foo = ()

bar :: String
bar = foo
```

errors with 

```
GHCi, version 8.6.5: http://www.haskell.org/ghc/  :? for help
[1 of 1] Compiling Main             ( Foo.hs, interpreted )

Foo.hs:7:7: error:
    • Couldn't match type ‘Int’ with ‘Char’ arising from a use of ‘foo’
    • In the expression: foo
      In an equation for ‘bar’: bar = foo
  |
7 | bar = foo
  | 
```

Exactly the same as with Deferred example. We get `Int` is not `Char` -error. Not `String` is not `()`.

---

Yet

```haskell
foo :: Ord a => a -> String
foo _ = 42

bar :: ()
bar = foo (succ :: Bool -> Bool)
```

with

```
    • Couldn't match type ‘[Char]’ with ‘()’
      Expected type: ()
        Actual type: String
    • In the expression: foo (succ :: Bool -> Bool)
      In an equation for ‘bar’: bar = foo (succ :: Bool -> Bool)
  |
5 | bar = foo (succ :: Bool -> Bool)
  |
```

and not with

```
 No instance for (Ord (Bool -> Bool)) arising from a use of ‘foo’
```

I hope that there would be a good story, so I as library writer could understand which type errors will happens when. E.g. if I rewrite `Member` in original proposal to be a `class`, will things work as expected (i.e. only when GHC decides to pursue finding the `Member` instance it will run into `TypeError`?)

---

My current intuition is that `TypeError` is `error` on type-level, so when GHC while evaluating types tries to `seq` the `TypeError` it will fail. Is this correct, or not? How more dependent stuff will change that?

---

> Is your suggestion that the warn-if-stuck message is emitted whenever the constraint should be discharged, but is stuck? I'm not exactly sure what the semantics would be here, but I do like the direction.

I have very operational feeling to it only. When the type-checker is searching for an instance or type-equality, and the operand is type-family application: then in addition to reporting *Cannot satisfy `IsOK b`* or *Couldn't match `Foo a` with `'True`*, GHC could report something extra.

tydeu:
From the proposal:

> Do we really need to worry about the backwards compatibility changes in Part 1.5? Things would be drastically simpler without them.

Yes, please worry about this. I use `TypeError` often in my type family heavy code to provide helpful errors when the family isn't used properly, so I would very much like to keep that functionality. Otherwise it would be very hard to know what went wrong when a type family does not reduce as expected.

In general, though, I like this proposal. It would be very useful to be able to define types synynoms of type errors. My current approach has been to use `TemplateHaskell`, but that is kind of overkill.

isovector:
Closing, because I don't have enough energy to push this through the committee. If anyone's interested in taking up the mantle, I'd be all in favor.

adamgundry:
@isovector thanks for the work you've put into this to date!

I think it would be great to see some improvements around custom type errors, if anyone has bandwidth to make progress in this area. There are a few suggestions on the GHC tracker (e.g. https://gitlab.haskell.org/ghc/ghc/-/issues/17168, https://gitlab.haskell.org/ghc/ghc/-/issues/18978). And in particular I made a suggestion on https://gitlab.haskell.org/ghc/ghc/-/issues/18310#note_295410:
> we could add a variant of `TypeError` with kind `Constraint`, that provides a function to witness the unsatisfiability of the constraint. That is, add something like:
> ```hs
> class Unsatisfiable (e :: ErrorMessage) where
>   unsatisfiable :: a
> ```
> The behaviour of this class would be rather better defined than instantiating the polymorphic `TypeError` to kind `Constraint`, because it is clear when to report the custom error: instead of an unsolved constraint error, when the constraint solver fails to solve a wanted `Unsatisfiable e`.
> 
> In particular, there is nothing wrong with mentioning `Unsatisfiable` in, say, a type family that computes a constraint; it is only when trying to solve the constraint that the error is reported. Whereas with `TypeError` there are few guarantees that the error won't be reported too early.

PR #279, opened 2019-10-01
Create 0000-prevent-incoherent-instances.rst

tysonzero:
Currently it is possible to cause incoherent instance behavior without `-Worphans` giving any warning and without even enabling `Overlaps`/`Overlappable` let alone `Incoherent`.

This proposal introduces a flexible and general mechanism that prevents this incoherence.

[Rendered](https://github.com/tysonzero/ghc-proposals/blob/patch-2/proposals/0000-prevent-incoherent-instances.rst)

### DIFF
--- a/proposals/0000-prevent-incoherent-instances.rst
+++ b/proposals/0000-prevent-incoherent-instances.rst
@@ -1,0 +1,168 @@
+Prevent Incoherent Instances
+============================
+
+.. author:: Daniel Smith
+.. date-accepted:: Leave blank. This will be filled in when the proposal is accepted.
+.. proposal-number:: Leave blank. This will be filled in when the proposal is
+                     accepted.
+.. ticket-url:: Leave blank. This will eventually be filled with the
+                ticket URL which will track the progress of the
+                implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+Currently it is possible to cause incoherent instance behavior without `-Worphans` giving any warning and without even
+enabling `Overlaps`/`Overlappable` let alone `Incoherent`.
+
+This proposal introduces a flexible and general mechanism that prevents this incoherence.
+
+Motivation
+----------
+As seen [here](https://pastebin.com/wyVMdRkc) and [here](https://pastebin.com/MQ4wd17Y), it is currently possible to break
+coherence without the use of the `Incoherent` pragma, without `-Worphan` warnings, and even without any usage of
+`Overlaps`/`Overlappable`.
+
+As seen above this incoherence can break various invariants without warning, and cause unintuitive and antimodular behavior,
+and thus we should prevent it.
+
+Proposed Change Specification
+-----------------------------
+We define a new pragma, currently named `Refine`, that can be attached to `class` and `instance` declarations to specify
+how [overlapping-]instances should be declared.
+
+`Refine` takes in a list of zero or more type variables from the class/instance head.
+
+Instances of this class, or instances that overlap with this instance, must match on a top level concrete type constructor
+for all of the above type variables, and they must have ownership of at least one of those type constructors. Any type
+variables not mentioned in the `Refine` clause have no restrictions and can be left bare if desired.
+
+For completeness a `noinstance` declaration will be added that does not actually define an instance, but exists solely to
+attach a `Refine` pragma too.
+
+By default classes and noinstances will have a `Refine` that specifies all type variables in the class/instance head,
+instances however will default to an empty `Refine` which disallows any overlapping instances.
+
+Instances that violate the above must be closely tracked to prevent incoherence, as soon as any two instances have
+unintentional overlap not declared with `Refine`, compilation should fail. Unless the `Incoherent` pragma is explicitly used.
+`Overlaps`/`Overlappable` would no longer be needed and can be phased out.
+
+Instances that violate the above should ideally not be defined in the first place, and should warn with `-Worphans` to avoid
+having to constantly watch out for unintentional overlap. However they should not be completely prevented to allow for
+situations like `data-default` and `data-default-class` where all the rules would be followed except for the rule that one
+of the concrete type constructors must be owned.
+
+The above prevents any unintentional overlap, leaving only intentional overlap from defining `Refine` on an `instance`.
+
+Currently the way `Overlaps`/`Overlappable` is handled we can cause incoherence even with only intentional overlapping.
+This is because the type checker is willing to commit too early to an instance which is later overlapped. Since we now
+must always specify in advance that an instance can be overlapped, we know exactly when to not commit too early to an
+instance. In those cases where an instance can be overlapped, constraints on top level values must not commit to that instance
+and must instead keep a FlexibleContexts-style expanded constraint that will be able to receive any overlapped instance.
+
+Examples
+--------
+
+Refine example indicating which instances would be considered orphans:
+
+```
+module Foo where
+
+class Foo a where
+
+instance Foo [a]
+
+[no]instance Foo (a -> b)
+   {-# Refine b #-}
+
+[no]instance Foo (a, b)
+    {-# Refine a b #-}
+
+class Baz a b
+    {-# Refine b #-}
+
+class Qux a b
+    {-# Refine a b #-}
+```
+
+```
+module Bar where
+
+import Foo
+
+data Bar
+
+instance Foo Bar
+
+instance Foo Int -- orphan
+
+instance Foo a -- orphan
+
+instance Foo [Bar] -- orphan
+
+instance Foo (a -> Bar)
+
+instance Foo (Bar -> a) -- orphan
+
+instance Foo (Int, Bar)
+
+instance Foo (Bar, Int)
+
+instance Foo (a, Bar) -- orphan
+
+instance Foo (Bar, a) -- orphan
+
+instance Baz a Bar
+
+instance Baz Bar a -- orphan
+
+instance Qux Int Bar
+
+instance Qux Bar Int
+
+instance Qux a Bar -- orphan
+
+instance Qux Bar a -- orphan
+```
+
+Example of adjusted type checking:
+```
+instance Show a => Show [a]
+    {-# Refine a #-}
+
+-- Rejected to prevent incoherence
+showInList :: Show a => a -> String
+showInList x = show [x]
+
+-- Accepted and prevents incoherence
+showInList :: Show [a] => a -> String
+showInList x = show [x]
+```
+
+Effect and Interactions
+-----------------------
+This proposal would prevent incoherence whilst keeping a large amount of flexibility and expressiveness.
+
+Costs and Drawbacks
+-------------------
+Some classes would require `Refine` to be specified immediately to continue being usable. For example `IsLabel` and `HasField`
+would want to specify that the `Symbol` type variable is not in the `Refine` list to avoid requiring owning a given Symbol.
+
+It is not as flexible as the status quo of allowing incoherence.
+
+Alternatives
+------------
+Doing nothing and accepting that instances are not guaranteed to be coherent once you leave Haskell2010.
+
+Unresolved Questions
+--------------------
+Is there a better pragma name than Refine?
+
+Is the above proposal truly coherent in all cases?
+
+Is there any reasonable and desirable existing code which cannot be easily adjusted to fit this proposal?


isovector:
Looks like you wrote a markdown document but saved it as rst --- the formatting is all wonky!

tysonzero:
Should be fixed now.

isovector:
For what it's worth, the ghc-proposal process now supports markdown, so in the future you need not translate one to the other :)

phadej:
The proposal doesn't even mention that problems starts **only** when you have `FlexibleInstances` enabled.

---

And quite often you can make helper classes to avoid FlexibleInstances,
if you really need.

IMO proposal should discuss that in the alternatives section.

```haskell
module Foo were

instance Foo [a]

class Foo a where

instance Foo [a]

class FooArr b => Foo (a -> b) where
```

```haskell
module Bar where

data Bar

instance Foo Bar -- ok
instance Foo Int -- orphan

instance Foo a -- needs UndecidableInstances, overlaps instance Foo [a]

instance Foo (a -> Bar) -- flexible instance, overlaps instance Foo (a -> b)

instance FooArr Bar -- ok, induces Foo (a -> Bar) instances

instance Foo (Bar -> a) -- flexible instance, overlaps Foo (a -> b)
```

AntC2:
Thanks @tysonzero, I'm having trouble putting my finger on what you're trying to address. (The `IncoherentInstances`/`INCOHERENT` pragma covers a multitude of sins.)

Could your Motivation section actually include the critical bits of code from the Pastebin examples, together with a narrative for what goes wrong/your explanation for what you think is wrong about it?

In general, GHC is far too willing to accept instance decls that taken together would make an inconsistent program theory; but take the risk that the usage sites won't ask for an instance at the inconsistent/overlapping combo of types. Contrast that Hugs has much tighter validation of instance decls; which means you don't get inconsistent program theory; but there are programs you want to write that it won't accept.

It would be good to link this proposal to the applicable tickets raising issues to do with incoherence -- there's gazillions of them. Can you show how/why such programs are rejected; and just as important show how very similar programs are accepted OK.

@phadej makes an important point that the difficulties start with `FlexibleInstances` -- including repeated tyvars in the instance head. Furthermore for Overlapping Instances, what GHC accepts is more liberal than the original design for Overlaps (see Mark Jones' paper/the Hugs manual); and if instance decls are more restricted like that, some of the difficulties go away. (You can still write all the same programs, you just have to do it in a rather convoluted style.)

Is the `noinstance` decl any different to doing

    instance (TypeError "no Foo (a -> b)") => Foo (a -> b)

tysonzero:
@phadej 

> The proposal doesn't even mention that problems starts only when you have FlexibleInstances enabled.

You also have an identical set of problems with MultiParamTypeClasses.

I originally had a more restrictive rule that you must always own the top level type constructor which would fix the FlexibleInstances problem, but then I noticed I still needed a more powerful solution to the MultiParamTypeClasses problem, at which point I noticed that this generalized solution also solved precisely the problems introduced by FlexibleInstances, so I relaxed the rule to be the same for both.

The advantage of allowing users to not have to define a helper class is to avoid extra namespace bloat that doesn't really add any additional value, as `FooList Bar` is IMO no better than `Foo [Bar]`.

If it weren't for MultiParamTypeClasses I would probably be more in favor of requiring the helper class, but since the `Refine` pragma naturally generalizes to FlexibleInstances I don't see much benefit in being more restrictive there.

One clear indication of how tightly linked the FlexibleInstances problem and the MultiParamTypeClasses problem are, is to look at how you can always write any multi param type class as a single param type class over a tuple, and how you can always rewrite a flexible instance using helper typeclasses which may or may not need to have multiple parameters.

AntC2:
> @tysonzero You also have an identical set of problems with `MultiParamTypeClasses`.

Actually no (but this is a common misconception). MultiParams give no problems with `NoFlexibleInstances` -- in particular note that prohibits repeated tyvars in instance heads, which probably makes MultiParams pretty much useless, but anyhoo.

>  you can always write any multi param type class as a single param type class over a tuple

Yes-ish, but without repeated tyvars within the tuple, you can't write the instances you want.

tysonzero:
> Could your Motivation section actually include the critical bits of code from the Pastebin examples, together with a narrative for what goes wrong/your explanation for what you think is wrong about it?

Yeah good idea.

Basically I am only defining lawful `Ord` instances, but still managed to break `Set`. Or in other words I was able to select a different `Ord` instance for the same type in two different locations without `Incoherent` and without a `-Worphan` warning.

So basically it all comes to the idea that currently GHC allows incoherent code to compile without warnings or errors, and I want to make it so that you need to explicit tell GHC you are ok with incoherence for that to happen.

> Is the noinstance decl any different to doing

```
instance (TypeError "no Foo (a -> b)") => Foo (a -> b) 
```

Hmm I suppose it isn't!

This is probably a more reasonable approach to go than a new keyword, so I will most likely edit the original proposal, although I think it needs a bit of polish:

1. You still have to fill out a body even though the code is unreachable.

It would be nice if you could omit the definition and ghc would be fine with it knowing that it is unreachable. Similar to the following Agda:

foo : Void -> String
foo ()

2. You have to enable `UndecidableInstances`.

tysonzero:
> Actually no (but this is a common misconception). MultiParams give no problems with NoFlexibleInstances -- in particular note that prohibits repeated tyvars in instance heads, which probably makes MultiParams pretty much useless, but anyhoo.

Ah that's a good point.

Due to the restrictiveness of only using MultiParamTypeClasses without FlexibleInstances I don't think this makes me want to change the proposal much, but I will have to keep this in mind when deciding how to phrase things and talk about alternatives and such.

AntC2:
A couple more ideas to include in the Alternatives section:

* **Closed Type Classes**, section 5.1 in [Morris & Eisenberg 2017](https://repository.brynmawr.edu/cgi/viewcontent.cgi?article=1075&context=compsci_pubs).

* [Instance Apartness Guards proposal](https://github.com/AntC2/ghc-proposals/blob/instance-apartness-guards/proposals/0000-instance-apartness-guards.rst) (now withdrawn), which essentially tries to avoid ever getting overlapping instances (for an extended sense of 'instance overlap'); and therefore avoids the dangers of orphans. 

> the MultiParamTypeClasses problem ... you can always write any multi param type class as a single param type class over a tuple,

I think you should consider how that interacts with `FunctionalDependencies`, or the equivalent using `TypeFamilies` and equality constraints: using a tuple instead of MultiParams would mean losing the ability to improve types, and the ability to write accurate signatures for methods; or you `AllowAmbiguousTypes` and need `TypeApplications` everywhere.

The Instance Apartness Guards proposal has a lot of material about interaction with FunDeps. Also some of the objections to that proposal would get raised here. (For example, the helper classes @phadej suggests.)

And before you put in a heap more effort here, something I learnt from experience with that Instance Guards: who's going to implement this proposal? If you don't have the skills/knowledge, you're reliant on GHC HQ. Then my take is that nobody at GHC HQ has any appetite for further work on Overlaps or FunDeps or `UndecidableInstances` (EKmett had a comprehensive proposal for that, which was also eventually withdrawn). Clearly the drift towards dependent Haskell is that we handle all type improvement logic as type-level functions; and overlap logic with Closed Type functions/Families.

Without wanting to put words in anybody's mouth, I think the claim would be that GHC can handle your examples with helper classes and Closed Type Families. Yes that would give namespace bloat. That pain is not severe enough for the additional compiler complexity in a makeover for the overlapping/incoherent instances machinery in GHC -- which is already a dog's breakfast, by all accounts.

tysonzero:
> A couple more ideas to include in the Alternatives section:

Hmm it seems like those are specifically to deal with overlapping instances?

The overlapping instances part of this is honestly an afterthought, I never use them personally. I just want the coherence issue in GHC fixed, and to try and prevent the occasional problematic instances I have seen appear in libraries.

> I think you should consider how that interacts with FunctionalDependencies, or the equivalent using TypeFamilies and equality constraints:

Yeah perhaps I was a little bold in my initial wording of that. My point was mostly just that this solution seems to generalize to both of these cases quite well. Although if some of this extra nuance does break the proposal in some way then that would of course be very important information. I'm not trying to get rid of MultiParamTypeClasses or anything like that.

 > Then my take is that nobody at GHC HQ has any appetite for further work on Overlaps or FunDeps or UndecidableInstances

> Without wanting to put words in anybody's mouth, I think the claim would be that GHC can handle your examples with helper classes and Closed Type Families.

I mean my approach is not trying to express new things that we can't already express. So I'm not particularly considered with what new features we can add to express these instances.

I'm trying to fix what I consider to be a pretty significant bug, namely that you can cause incoherence without any warning and with extensions generally considered to be benign.

Perhaps most consider coherence to be of fairly low importance, but I had assumed otherwise when making this proposal.

AntC2:
>> A couple more ideas to include in the Alternatives section:

> Hmm it seems like those are specifically to deal with overlapping instances?
>
> The overlapping instances part of this is honestly an afterthought, I never use them personally. I just want the coherence issue in GHC fixed, ...

I'm now very confused as to your motivation. I suggest you write out the Motivation section with examples before making any more proposals. Why do you think this is to do with _anything else_ than overlapping instances? The `INCOHERENT` pragma/`IncoherentInstances` is all and only to do with overlaps. If you "never use" overlapping instances, you can't possibly be bumping into coherence issues. Orphan instances are only problematic if an instance declared in the orphan module overlaps or is overlapped by instance(s) in the module that declares the class/datatype.

I'd say that far more realistic use cases are people who use and understand overlaps, and never use `INCOHERENT` because it is just as unpredictable as it sounds. OK avoiding `INCOHERENT` does need work-rounds and can lead to code bloat, but there's usually a way.

> and to try and prevent the occasional problematic instances I have seen appear in libraries.

Then please give examples of those, and clearly describe why they're "problematic". It's true that for some libraries, because you can't intrude into the instances they declare, you do end up with potential incoherence. That is all and only because of overlap. Please explain with an example:

> you can cause incoherence without any warning 

You perhaps don't realise (though I said it already) that just declaring instances that overlap (and without `OVERLAPS` etc or `INCOHERENT` pragmas) doesn't "cause incoherence" and doesn't give warnings -- as GHC sees it, that's merely creating the _potential_ for overlap/incoherence. What triggers incoherence is at a usage site. If what you're proposing amounts to rejigging the compiler phasing to validate and reject potential overlap/incoherence at instance decls rather than at usage, that's a major re-engineering of GHC.

tysonzero:
> I'm now very confused as to your motivation. I suggest you write out the Motivation section with examples before making any more proposals.

> You perhaps don't realise (though I said it already) that just declaring instances that overlap (and without OVERLAPS etc or INCOHERENT pragmas) doesn't "cause incoherence" and doesn't give warnings -- as GHC sees it, that's merely creating the potential for overlap/incoherence. What triggers incoherence is at a usage site.

I linked [this](https://pastebin.com/wyVMdRkc) example in that section which I thought very clearly demonstrated the problem.

I never used `Overlaps`/`Overlapping`/`Incoherent` and I had `-Worphans` enabled, but I still got a completely incoherent result and absolutely did "cause incoherence".

I apologize if I sound frustrated but I don't know what I need to do to make it clearer. The `baz` set has 6 elements including multiple duplicates, completely violating the invariants of `Set` without defining a single unlawful `Ord` instance.

nil-ptr:
> I linked [this](https://pastebin.com/wyVMdRkc) example in that section which I thought very clearly demonstrated the problem.

I don't think there's any incoherent instance resolution happening in that example at all. The result is obviously bogus, but that's because `Data.Set` assumes that all values of type `Set a` must have been constructed using *the same* `Ord` instance. Not merely lawful ones. That is not the case in your example.

In `OrphansBar` `bar` uses the `Ord (Tuple a Bar)` instance because that's the only possible choice at that `Data.Set.fromList` call site. No other applicable instances are in scope. And similarly in `OrphansFoo` `foo` uses the `Ord (Tuple Foo b)` instance, because that's the only possible choice there. 

The only constraints that need to be solved for in `baz` are an `Ord Foo` constraint from the type of `bar` with `a ~ Foo` and an `Ord Bar` constraint from the type of `foo` with `b ~ Bar`. 

None of the above involves overlapping or incoherent instances.  `Ord (Tuple a Bar)` and `Ord (Tuple Foo b)` *do* overlap with each other in `OrphansBaz`, but ghc is ok with this because neither instance is ever actually used in that module.

If you want to trigger overlapping instance errors, you need to "delay" instance resolution until there really are overlapping instances in scope. In this case, "delaying" just means changing the types of `foo` and `bar`slightly:

```haskell
-- In OrphansFoo:
foo :: Ord (Tuple Foo b) => b -> b -> Set (Tuple Foo b)
-- In OrphansBar:
bar :: Ord (Tuple a Bar) => a -> a -> Set (Tuple a Bar) -> Set (Tuple a Bar)
```

That moves the choice of `Ord (Tuple a b)` instances to the `foo` and `bar` call sites in `baz`. Which should be enough to make ghc complain about overlap. Note that this requires enabling `FlexibleContexts` in `OrphansFoo` and `OrphansBar`.

tysonzero:
> I don't think there's any incoherent instance resolution happening in that example at all.

> but that's because Data.Set assumes that all values of type Set a must have been constructed using the same Ord instance.

That's the exact definition of coherence.

Coherence means that for a given typeclass and type(s), you will always resolve it to the same instance dictionary within the same program.

So the fact that two different `Set a` were constructed with different instances, by the very definition of the word, means that instances were selected in an incoherent manner.

> If you want to trigger overlapping instance errors, you need to "delay" instance resolution until there really are overlapping instances in scope. In this case, "delaying" just means changing the types of foo and bar slightly:

Right, if you look at the "adjusted type checking" part of the proposal I suggest that GHC needs to do exactly that, so that we can keep coherence and either get errors or correct and coherent overlap.

AntC2:
> I linked this example in that section which I thought very clearly demonstrated the problem.

That's 60 lines of code without a skerrick of comment or narrative. a) people might not wade through it; b) they might not see any "problem"; c) they might see a problem that is not the problem you think is "very clearly demonstrated"; d) If GHC accepted it without error or warning and it runs without throwing an exception, then there can't be a problem.

Thank you to @nil-ptr for making more effort than I did. The answer seems to be d).

> I never used Overlaps/Overlapping/Incoherent and I had -Worphans enabled, but I still got a completely incoherent result and absolutely did "cause incoherence".

Then please define what you mean by "incoherence" (in the Motivation section -- you've now explained it in the reply to @nil-ptr, why did it take this much asking?). It's clearly not the same as what GHC's `INCOHERENT` pragma detects.

BTW you _do_ have overlapping instances for `Ord` -- just not declared in the same module. It's exactly the kind of partial overlap described and warned of at length in the Users Guide at "the potential for overlap". One possible fix (this sometimes works, but is very unreliable) put `OVERLAPPABLE` pragma on _both_ those instances. Maybe that'll delay instance resolution, in the manner @nil-ptr's talking about.

> I apologize if I sound frustrated but I don't know what I need to do to make it clearer.

Exactly as if this were a bug report: describe the expected behaviour (in English); give a minimal piece of code that exhibits the "problem"; say what GHC does and why that's 'wrong' in your opinion.

> The baz set has 6 elements including multiple duplicates, completely violating the invariants of Set without defining a single unlawful Ord instance.

Does it? Oh, so I need to run this code and inspect the output to decide whether it exhibits any problem. Your 'bug report' should show that output.

nil-ptr:
Ah. Well, we're using different definitions of what it means for an instance to be coherent then. Because to me every instance selection in that code is perfectly reasonable. 

But I think I get what you're actually after now: you want *global* instance coherence, correct? Or package/project-global coherence, anyway.

tysonzero:
> If GHC accepted it without error or warning and it runs without throwing an exception, then there can't be a problem.

Uhh... what? If `1 + 1` returned `3` without error or warning or exception, that would be a problem... In fact I would say the biggest problems are the ones that involve no errors or warnings or exceptions.

> Then please define what you mean by "incoherence" (in the Motivation section -- you've now explained it in the reply to @nil-ptr, why did it take this much asking?)

I honestly assumed there was a single agreed upon definition of coherence. The same one used in [Type Classes vs. the World](https://www.youtube.com/watch?v=hIZxTQP1ifo) and the same one used in the top results when you search [typeclass coherence](https://www.google.com/search?q=typeclass+coherence). Although I see some disagree on the definition so I should have been more precise.

> It's clearly not the same as what GHC's INCOHERENT pragma detects.

I mean it basically is. The incoherent pragma allows you to violate the above definition of coherence. It just turns out that there are other ways to violate coherence, orphans (which was known and detected by `-Worphans`) and FlexibleInstances (which was relatively unknown as far as I can tell and why I'm bringing up this issue).

> BTW you do have overlapping instances for Ord -- just not declared in the same module.

Right. I should have been more clear with my wording. I don't use "OverlappingInstances" the intentional feature made up of `Overlaps`/`Overlappable`, but you are right that I am discussing "instances that overlap".

What I want is for ghc to properly enforce that there is no unintentional overlap that leads to incoherence. *Intentional* overlap I am less concerned about as I do not use it, but I wanted to make sure the proposal didn't get in the way of those who do like intentional overlap.

> Exactly as if this were a bug report: describe the expected behaviour (in English); give a minimal piece of code that exhibits the "problem";

Yeah alright that's a fair point. I came at this too much from a feature proposal point of view, when I should have treated it as a bug report first and foremost and a feature proposal as the solution.

> Does it? Oh, so I need to run this code and inspect the output to decide whether it exhibits any problem. Your 'bug report' should show that output.

There is a comment at the bottom with the contents of the set.

> But I think I get what you're actually after now: you want global instance coherence, correct? Or package/project-global coherence, anyway.

I was under the impression that is what everyone wanted/expected (except when using `Incoherent` or not being careful with orphans).

When I google "instance hiding" or "instance coherence" all the discussions care about it on a project-wide level, not on an a local level.

I mean what's the point of local-only coherence? With only local coherence you can break `Set` and `Map` and `To/FromJSON`, but with global coherence (and lawful instances) you cannot.

gbaz:
the ghc manual agrees on the notion of coherence. c.f.

https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#overlapping-instances

I don't this proposal prevents it however! It just gives certain control over more granular warnings. I don't know if I'm particularly happy with the mechanism proposed for the warnings, which feels ad-hoc, complicated, and confusing. And I think that the discussion in the proposal itself could use some more clarity in establishing  exactly what problem its trying to solve, etc. As it is it feels unformed. But as a general goal: warning about potential conflict in instances more thoroughly seems like a sound problem to attack.

@phadej can you elaborate more clearly on the problem with flexible instances? e.g. the users guide warns about overlaps, but not flexible instances in this regard. can you provide an example that shows the same behavior with just flexible?

AntC2:
> I was under the impression that is what everyone wanted/expected ...
> I mean what's the point of local-only coherence?

Oh, there are regular requests/proposals for local instances. For example a module that does pretty `Show` for complex data structures with your carefully crafted datatypes embedded, whereas elsewhere you want terse/diagnostic `Show`. Of course poly-instance `Show` doesn't do the harm that poly-instance `Ord` might, but you might want case-insensitive `Ord` for `[Char]` or for `MyPrettyStringType` as part of pretty `Show`.

You can somewhat achieve that effect by exploiting orphan instances. But everybody knows that's flakey. (I'm playing devil's advocate here: those proposals disintegrate when you give some pointed examples of local vs exported functions that pollute instance resolution globally.)

> The incoherent pragma allows you to violate the above definition of coherence.

Here's a sensible use of instances that seem incoherent (but aren't `INCOHERENT` haha)

    instance {-# OVERLAPPABLE #-} Ord b => Ord (Tuple Foo b)

    instance {-# OVERLAPPABLE #-} Ord a => Ord (Tuple a Bar)

    instance Ord (Tuple Foo Bar)               -- (TypeError "panic the impossible happened") =>

The first two `OVERLAPPABLE` instances (I've taken from your code) do overlap in an irresolvable/incoherent fashion. **But** GHC will never detect that because the third instance sits at exactly the point of intersection. So you don't need an `INCOHERENT` pragma -- indeed putting that pragma will produce a program that might be incoherent (by your measure), whereas with the `OVERLAPPABLE` pragmas you get (with luck) consistent instance selection or a rejection if your usage sites stuff up. You do need to make sure that most-specific instance is in scope everywhere either of the other two are, but you don't need the first to be in scope everywhere the second is, and v.v.

> I honestly assumed there was a single agreed upon definition of coherence.

There might be for those who know the topic closely. Never the less, your proposal should give a definition (or at least reference one). Then the examples in the proposal can specifically say: this code is incoherent by that definition; and is to be rejected, by this part of the proposed change.

AntC2:
>  ... ways to violate coherence, ... `FlexibleInstances` (which was relatively unknown ...

>  can you elaborate more clearly on the problem with flexible instances?

There's not so much a problem with `FlexibleInstances` as poor documentation. And since most people switch on `FlexibleInstances` everywhere, it really only matters when we get to the sort of nitty-gritty discussions happening here.

* Without `FlexibleInstances`, you can't write instances that overlap. You can write instances with the `OVERLAPS` and friends pragma, but those pragmas are unnecessary ... except maybe when you import instances from modules that do have `FlexibleInstances`.

  Note that with `MultiParamTypeClasses` but not `FlexibleInstances`, you still can't write instances that overlap: each type in the instance head must be of the form `T tv1 tv2 ...` in which `T` is a type constructor and the `tvn` are type variables (`n` >= 0) and there are no repeated tyvars across the whole instance head.

* Even with `FlexibleInstances`, you won't necessarily get (or want) instances that overlap, for example

        instance Ord (Tuple Foo a)
        instance Ord (Tuple Bar b)
        instance Ord (Either c c)

The 'problem' is that there's no way to tell GHC: I want `FlexibleInstances`, but reject any instance decls whose heads overlap.

The far-and-away clearest way to demonstrate this is to run up Hugs. In Hugsmode you get `FlexibleInstances`; but you need to [`:set +o`](https://www.haskell.org/hugs/pages/users_guide/class-extensions.html#OVERLAPPING-INSTANCES) to allow instance decls to overlap. Even with `+o` Hugs doesn't allow the combo of instances for `Ord (Tuple ... ...)` in @tysonzero's Pastebin example: the rule is that for any overlapping instance heads, there must be one that is strictly more specific. (And that means that my example with a third instance at the intersection is still rejected.)

viercc:
I agree with @tysonzero on the opinion it is a real problem that global coherence can be compromised silently.

I don't mind there is a way to intentionally break (even local) coherence (`Overlappable` etc.)
I don't mind it is possible to define overlapping instance. When an ambiguity happens, GHC shouts at me anyway, unless I explicitly wrote `Overlappable` etc. I think these facts are irrelevant to the motivation of this proposal.

To me, the problem is simple: if `FlexibleInstances` is used, the logic behind `-Worphans` is too weak to detect it may cause (global = inter-module) incoherence. One purpose of `-Worphans` existed was to warn orphan instances may cause (global) incoherence due to separate compilation GHC employed, wasn't it?

Anyway Haskell2010 does not allow any form of overlap. Also requires global coherence by [clearly stating](https://www.haskell.org/onlinereport/haskell2010/haskellch4.html#x10-750004.3):

> A type may not be declared as an instance of a particular class more than once in the program.

ekmett:
Would a simpler fix be to have this sort of flexible instance trigger `-Worphans`?

Here the instances for `Tuple Foo a` and `Tuple a Bar` are not being defined in the same module as `Tuple` and so if you were to build them out of smaller "helper" classes, the result would involve orphan instances.

nil-ptr:
> I honestly assumed there was a single agreed upon definition of coherence.

As did I. I took it to mean a related but different thing than what I now realise you meant. I'd type up an explanation, but some googling revealed that @ezyang wrote a [blog post](http://blog.ezyang.com/2014/07/type-classes-confluence-coherence-global-uniqueness/) about this same confusion, which explains things much more clearly than I think I would've. Admittedly, having read that, I think my own understanding of the term was a little off too.

Anyway, returning to the issue at hand: I'm inclined to agree with @ekmett 's suggestion. That would be a far less invasive change. And should probably be done regardless, in my opinion. I was rather surprised the `Set (Tuple Foo Bar)` example didn't trigger any orphan warnings.

tysonzero:
@ekmett 

In such a case how would `IsLabel` (and similar) be handled?

The following are both safe and do not risk overlap, but both of them don't own the first type parameter.

```
instance ... => IsLabel x Foo
instance ... => IsLabel "foo" Bar
```

You could use the kind as a guide, but I would say it is morally reasonable to define a custom `IsLabel`-like class that uses `Type` instead of `Symbol` to allow for an extensible set of keys.

The known ways to get incoherence with `-Worphans` and without `Incoherent` are:

1) FlexibleInstances using instances that I would consider to be orphans

2) OverlappingInstances using instances that I would consider to not be orphans

And any solution must keep in mind at least the following considerations:

1) Use cases like `Cast a b` where all type variables are going to be matched on

2) Use cases like `IsLabel x a` where at least one type variable might be left bare

3) Use cases like `Show [Char]` where coherent overlap is desired

Trying to handle all of the above is what lead to the solution I proposed:

I couldn't find a way to support both `Cast` and `IsLabel` without requiring an explicit declaration (although you don't need the declaration for `Cast` as it is the default).

I couldn't find a way to support choosing which of `Foo (Int, MyType)` or `Foo (MyType, Int)` (or both) is the orphan without the `noinstance` stuff.

I couldn't find a way to fix `OverlappingInstances` without the type-checker changes.

ekmett:
Is there still stuff that you can't handle safely once you've extended `-Worphans`? Sure. But at least now you'd be told you're doing something that could be dangerous.

You have a concrete proposal here that you feel addresses the issue. I'm not so sure. It has a lot of moving parts. But I do feel that we should figure out a way to shut the barn door at least, and then more carefully consider what things can be safely let out.

Out of your 3 examples, I'd argue that `Show` is handled "correctly" in the `String` case by defining a helper inside of the `Show` class to avoid the need for overlap. It is morally defined at the same time as the `[]` instance by the way the helper is baked in, and isn't any sort of orphan or overlap situation. 

An attempt to define your own fancy flexible `instance Foo (Bar Baz)` really would be an orphan, unless you did so in the module that defined `Foo` or `Bar`.

Going back to the start of your enumeration:

`IsLabel` is just bad. As defined there is really no "safe" way to use that class. Every instance will have to be an orphan as none are defined with it, and as you note, nobody owns symbols. 

As it stands, I see three rather disjoint sets of clients for it today. One group tries to use it to implement extensible records or named parameters or some other trick use as it offers very cheap syntax. Another wants to build instances that look like field accessors. Another wants lenses. All three of these are incompatible, and since all choices are orphaned, any of their usages of the class fragments the community into mutually incompatible fragments. 

Saying you're going require instances to refine the type of the result term for `fromLabel` eliminates the "trick" users, but doesn't do anything to discharge the disconnect between the functional and lens-style users, both of which are trying to define incompatible refinements of that one term.

The only winning move is not to play.

Without a data type you're defining in the same module on the outside of each of the arguments to an MPTC (or if you want to get fancy, at least all of the ones that determine the others via functional dependency, have fun specifying that formally), those instances are bad for the same global instance resolution reasons that we rule out "orphans".

`Cast`-like classes are also a trash fire. n^2 complexity, no way to get transitive closure, and no clear ownership lines. There is no way to define them that doesn't risk conflicting instances when spread across multiple modules unless you can introduce some kind of canonical normal form and split the instances in half allowing casts between things with the same normal form. `Coercible` relies on compiler magic for fairly good reasons, as doing this right in general, especially spread across multiple modules, is hard.

Another example of this sort of problem that comes to mind for me is

```haskell
class Monoid m => MonoidAction m s where
  // act mempty = id
  // act (m <> n) = act m . act n
  act :: m -> s -> s
```
You'd like to have things like

```haskell
instance MonoidAction m (Endo m)
```

that can work for all monoids, but for any concrete monoid you probably also want

```haskell
instance MonoidAction Foo Foo
```

but then we also have cases where you'd want to lock down the monoid and still vary the set independently. The lack of fundep/ordering/refinement clause whatever, means you suffer both when it comes to type inference as it tends to be pretty floaty to use requiring `Proxy` arguments or explicit type applications, and instances tend to clash just as in the `IsLabel` and `Cast` cases when you orphan them in separate modules. 

My (sketch of a) proposal doesn't try to go as far as yours, but I'm not off hand very sure that your proposal as written works to serve its stated end goals. It is a lot of machinery and at least to my superficial read, doesn't seem to actually help in those scenarios above. 

In the event that I'm wrong about this knee jerk reaction, at the very least improving `-Worphans` does no harm here. 

I remain hopeful that we could with a relatively straightforward refinement of the existing stated rule for `-Worphans` restore the general sense of security that you'd get when turning on that flag. 

I'd like to ensure that we don't wind up with users facing multiple instances for a given class for a given type without explicitly asking for something dangerous.

If you want to construct better safeties for the guns that users use to shoot themselves in the foot, great. That doesn't strike me as incompatible with putting up warning signs at the firing range.

viercc:
If I'm not missing something, there is a case not considered by the proposal.

``` haskell
module Foo where
  class C a b
  {-# Refine a #-}
  
  -- Is this declaration allowed?
  noinstance C a [b]
  {-# Refine b #-}

module Bar where
  import Foo
  
  data Bar
  
  instance C a [Bar]
  
module Baz where
  import Foo
  
  data Baz
  
  instance C Baz b
```

We shouldn't allow this, but the proposal is not clear about it.
Clearly it violates `Refine a` attached to the `class C a b`, but the example showed
in the proposal also violates.

``` haskell
-- Quoted from the proposal
module Foo where

class Foo a where

noinstance Foo (a -> b)
```

`noinstance Foo (a -> b)` violates `Refine a` attached to `class Foo a` (which is the default),
because `Foo` does not own `(->)`.

viercc:
[Also I'll add one more alternative on the table](https://github.com/viercc/kitchen-sink-hs/blob/master/dominance.md)

I compared @ekmett's simple fix and @tysonzero's this proposal. If there are misunderstandings, I'm happy to be corrected.

Also I forgot it on the comparison: all of them does not give an answer to "how FunDeps interact with it?"

tysonzero:
> Out of your 3 examples, I'd argue that Show is handled "correctly" in the String case by defining a helper inside of the Show class to avoid the need for overlap. It is morally defined at the same time as the [] instance by the way the helper is baked in, and isn't any sort of orphan or overlap situation.

I would argue that the `showList` approach is less general. As it means that you can have a `Show` that changes once you put that type in a `[]`/`Vector` etc. but it means that you cannot simultaneously have a `Show` that changes once you put that type in a `Maybe` or a `Show` that changes once you put that type in a `Tree`. You would need `showMaybe`, `showTree` etc. all in the `Show` class.

With something like the above you can have `Show` or the third party library define  `[no]instance Show a => Show (Foo a) where {-# Refine a #-}`. I still think `showList` is not a bad approach and i'm not rushing to have it changed, but it doesn't seem particularly general to me.

> An attempt to define your own fancy flexible instance Foo (Bar Baz) really would be an orphan, unless you did so in the module that defined Foo or Bar.

That's a reasonable approach to take, but it essentially classifies safe extensible/modular uses of overlapping instances as orphans. It also forces the use of helper classes to avoid orphan warnings, even if those helper classes solely exist as a way to hang instances off of a constructor. All of this is reasonable and it wouldn't break my personal workflow, but it does cause things that are safe and that others may use to start throwing warnings.

> IsLabel is just bad. As defined there is really no "safe" way to use that class. Every instance will have to be an orphan as none are defined with it, and as you note, nobody owns symbols.

> As it stands, I see three rather disjoint sets of clients for it today. One group tries to use it to implement extensible records or named parameters or some other trick use as it offers very cheap syntax. Another wants to build instances that look like field accessors. Another wants lenses. All three of these are incompatible, and since all choices are orphaned, any of their usages of the class fragments the community into mutually incompatible fragments.

I would argue that if the way you want to use it requires an orphan (direct lenses, field accessors) then it is pretty clearly not what you should be doing with it.

There are some legitimate and IMO quite nice uses of it that don't need orphans. For example the `optics` library, and the `Name` instance of the `named` library.

It seems as though `HasField` will end up handling the record use case, with `{}`-based and `.`-based syntax sugar potentially in the future. For lenses I think `optics` or a `lens #foo` function could build off of `HasField` to handle the lens use case.

So based on the above I would actually argue that `IsLabel` is best used for constructors/sums as well as the above non-orphan use cases. In other words `base` should define one of:

```
noinstance IsLabel x (a -> b)
    {-# Refine b #-}

-- Need to manually recurse for multiple argument constructors
-- noinstance IsLabel x (a -> b -> c)
--     {-# Refine c #-}
```

```
class IsLabelOutput x b where
    type Input x b :: Type
    funcFromLabel :: Input x b -> b

instance (IsLabelOutput x b, Input x b ~ a) => IsLabel x (a -> b) where
    fromLabel = funcFromLabel @x @b

-- Need to manually recurse for multiple argument constructors
```

```
type family Uncurry (as :: [Type]) (b :: Type) :: Type where       
    Uncurry '[] b = b                                              
    Uncurry (a ': as) b = Uncurry as (a -> b)                      
                                                                   
type family Tail (as :: [Type]) :: [Type] where
    Tail (a ': as) = as

class IsLabelOutput (x :: Symbol) a where
    type Inputs x a :: [Type] 
    funcFromLabel :: Uncurry (Inputs x a) a
    
instance (IsLabelOutput x b, Inputs x b ~ (a ': as)) => IsLabelOutput x (a -> b) where
    type Inputs x (a -> b) = Tail (Inputs x b)
    funcFromLabel = funcFromLabel @x @b

-- Complex but handles arbitrary argument constructors
```

This would also allow the remaining (useful) instances in `named` to stop being orphans, and would allow for `#just 5` and `#left True` and more importantly `#customCurrentlyPrefixedConstructor "foo"`. Essentially solving the namespace issue of constructors.

> Cast-like classes are also a trash fire. n^2 complexity, no way to get transitive closure, and no clear ownership lines. 

I appreciate this sentiment and I won't claim a manual `Cast` is particularly elegant by any means, but honestly I would say I would benefit in practice from just having things like `Cast Text String`, `Cast T.Text TL.Text`, `Cast Int Integer` and so on. I can always reach for some other more function or typeclass in the cases when an instance I would like to have is missing.

If we ignore `Cast` there are still classes like `LeftModule` and others that could benefit from that same "at least one type constructor must be owned, the rest cannot be bare" approach. Particularly since it's what you get currently with `-XNoFlexibleInstances` combined with `-Worphans`.

> but then we also have cases where you'd want to lock down the monoid and still vary the set independently.

Yes, in cases where you want `Foo Bar a` and also `Foo a Baz`, my proposal doesn't help you, because those two instances overlap (and not in a safe way).

Likewise combining `Foo Bar a` and also `Foo Int Baz` is not supported, because in principal since you can define `Foo Int Baz` aka `Foo NotOwned Owned`, you should be able to define `Foo Bar MyType` which would be overlap that `Foo Bar a` did not give permission for.

Now these cases could potentially be handled by increasing the complexity, the former by requiring another instance that handled the overlapping case explicitly, and the latter by requiring `Foo Bar a` allow overlap in `a`. But the complexity is really spiking at this point.

> In the event that I'm wrong about this knee jerk reaction, at the very least improving -Worphans does no harm here.

I mean I would like the `optics` library to continue compiling with `-Worphans -Werror` as morally it really isn't doing anything orphan-y, similarly I would like the non-orphan instances in `named` to continue compiling, as well as any `HasField`/`IsLabel` instances I end up defining.

> I'd like to ensure that we don't wind up with users facing multiple instances for a given class for a given type without explicitly asking for something dangerous.

I agree, and I do think all the proposals mentioned successfully achieve that. It's just about what functionality we want to allow to continue to compile without warnings.

> If I'm not missing something, there is a case not considered by the proposal.

```
-- Is this declaration allowed?
noinstance C a [b]
{-# Refine b #-}
```

Good catch, I should have specified! It would not be allowed! If you own `C` you automatically own every type constructor, but you don't gain ownership over bare type variables.

> Also I'll add one more alternative on the table

> If there are misunderstandings, I'm happy to be corrected.

```
  instance C (F x) [y]   -- All of (A) (B) (C) accept
  instance C [x] (F y)   -- (A) (B) only
  instance C (F x) y     -- (C) only, but (B) can be customized to allow it
  instance C x (F y)     -- Nothing accept, but (B) can be customized to allow it
  instance C [F x] [F x] -- (C) only
```

Correct except for the last one, it should be `(C) only, but (B) can be customized to allow it`:

```
class C a b
    {-# Refine a b #-}

noinstance C [a] [b] where
    {-# Refine a b #-}
```

Your alternative seems interesting, although one big thing I notice that spooks me a little is that it seems like it would be incompatible with `IsLabel` and `HasField`, that both currently exist in `base`.

> Also I forgot it on the comparison: all of them does not give an answer to "how FunDeps interact with it?"

It seems like my proposal can more or less ignore them, and typically users would just choose to not put the type variables on the "output" side of the FunDep in the `Refine` clause. Which as avoids the fundep complexity issue noted by @ekmett.

I suppose you could try and adjust the defaults when there is a FunDep involved, but personally i'm not overly motivated to do such a thing due to the complexity. ^^^Besides I prefer TypeFamilies

viercc:
> Correct except for the last one, it should be (C) only, but (B) can be customized to allow it:

Ah, yes. But ...

* Required number of `noinstance`s can be unbounded. `C (F a) [b]`, `C [F a] [b]`, `C [[F a]] [b]`, ...
* What if that instance used some type constructor `F`, unknown to the module `Foo`,
instead of `[]`? You can, theoretically, demand the owner of `F` to include that `noinstance`. But I don't think it is feasible at all.

> `IsLabel` will be incompatible

(C) does not rule out `IsLabel` as a class. I know, you meant useful instances are ruled out.

I wrote (C) can employ per-class customization on ordering of parameters of MPTCs.
Because that note wasn't meant to be a complete proposal, I didn't make up syntax for it. Let me write `{-# ParamOrder  a b c #-}` for now.

* `IsLabel` instance for raw van Laarhoven lenses is orphan to start with. The only way to prevent silent incoherence with it is including this instance in `GHC.OverloadedLabels`.
* An workaround using auxiliary type class exists.

```haskell
class IsLabel (name :: Symbol) a
  {-# ParamOrder a name #-} where

class IsLabelFn (name :: Symbol) a b
  {-# ParamOrder b a name #-} where
  fieldFn :: a -> b

instance IsLabelFn name a b => IsLabel name (a -> b) 
```

This is a hack, maybe even a bad know-how. But it buys simpler semantics of the language.

I think your proposal still works, if you removed `noinstance`, using this workaround. **Is there a benefit of having `noinstance` in your proposal, other than better-looking code?**

viercc:
Eureka! Probably you can define the semantics of `noinstance` by transpilation.

```haskell
module Foo where
  class C a b {-# Refine a b #-} where
    foo :: a -> b -> Foo
  noinstance Eq a => C a (b,c) {-# Refine a b c #-}
```
This translates into
```haskell
module Foo where
  class C a b {-# Refine a b #-} where
    foo :: a -> b -> Foo
  -- 12345 is unique id generated by the compiler
  class Eq a => C_12345 a b c {-# Refine a b c #-} where
    foo_12345 :: a -> (b,c) -> Foo
  instance C_12345 a b c => C a (b,c) where
    foo = foo_12345
```
Then, another instance which overlaps to `C a (b,c)` also gets translated
```haskell
{-
Automatically generated from
instance C T (U,V) where
  foo = <impl>
-}
instance C_12345 T U V where
  foo_12345 = <impl>
```

Though I withhold the judgement whether this machinery is good or bad. Probably I don't like.

tysonzero:
> Ah, yes. But ...

Sure, but `(C) only, but (B) can be customized to allow it` is still the most reasonable comment to put next to it.

> I wrote (C) can employ per-class customization on ordering of parameters of MPTCs.
Because that note wasn't meant to be a complete proposal, I didn't make up syntax for it. Let me write {-# ParamOrder a b c #-} for now.

I think at that point it'd be very hard to argue it's actually simpler than my proposal, I would put them at fairly comparable complexity.

In general I really don't think my proposal is particularly complex, you just have to own at least one type constructor (or the class) for the variables mentioned in `Refine`, and the rest cannot be bare.

It also seems like your proposal is incompatible with the current `-XNoFlexibleInstances -Worphans` behavior, which is a little unfortunate, as that behavior is sound/coherent and useful for a good chunk of MPTC uses.

The follow works with `-Worphan -XNoFlexibleInstances` currently, but as far as I can tell it is impossible to make it work with your proposal, even with `ParamOrder`.

```
newtype MyInt = MyInt Int

instance Cast MyInt Int
instance Cast Int MyInt
```

> I think your proposal still works, if you removed noinstance, using this workaround. Is there a benefit of having noinstance in your proposal, other than better-looking code?

It's mainly to avoid namespace bloat and verbose/ugly code. It's also to be consistent with how things work for intentional overlap, so that switching from `noinstance Show a => Show [a] {-# Refine a #-}` to `instance Show a => Show [a] {-# Refine a #-}` or vice versa is painless.

```
noinstance IsLabel x (a -> b)
    {-# Refine b #-}

noinstance IsLabel x (a -> b -> c)
    {-# Refine c #-}

noinstance IsLabel x (a -> b -> c -> d)
    {-# Refine d #-}
```

```
class IsLabelOutput1 (x :: Symbol) b where
    type Input1 x b
    funcFromLabel1 :: Input1 x b -> b

instance (IsLabelOutput1 x b, Input1 x b ~ a) => IsLabel x (a -> b) where
    fromLabel = funcFromLabel1 @x @b

class IsLabelOutput2 (x :: Symbol) c where
    type Input2a x c
    type Input2b x c
    funcFromLabel2 :: Input2a x c -> Input2b x c -> c

instance (IsLabelOutput2 x c, Input2b x c ~ b) => IsLabelOutput1 x (b -> c) where
    type Input1 x (b -> c) = Input2a x c
    funcFromLabel1 = funcFromLabel2 @x @c

class IsLabelOutput3 (x :: Symbol) d where
    type Input3a x d
    type Input3b x d
    type Input3c x d
    funcFromLabel3 :: Input3a x d -> Input3b x d -> Input3c x d -> d

instance (IsLabelOutput3 x d, Input3c x d ~ c) => IsLabelOutput2 x (c -> d) where
    type Input2a x (c -> d) = Input3a x d
    type Input2b x (c -> d) = Input3b x d
    funcFromLabel2 = funcFromLabel3 @x @d
```

viercc:
> > Ah, yes. But ...
> 
> Sure, but (C) only, but (B) can be customized to allow it is still the most reasonable comment to put next to it.

I don't deny I was wrong (Let me take a time to update that note). I meant that (B) and (C) have incomparable set of safely acceptable instances. Even with carefully wrote `noinstance`s.
That example I wrote did not to show it successfully.

> Even the addition of noinstance isn't complex,

I was seeing the complexity of your proposal mainly from `noinstance`, and being able to attach `Refine` to `[no]instance`. Not from `Refine` attached to the class only.

Your proposal can be broken into some steps.

1. We have to make `-Worphans` stricter. (@ekmett's) (A) is the simplest.
2. For MPTCs, it is good to be able to choose which parameters are required to be refined.
   (Rather than all parameters must be refined.)
   Introduce `{-# Refine #-}`.
3. It is safe to accept specific subset of instances, characterised by being specialized type of
   another instance (e.g. `C (a -> Foo)` against `C (a -> b)`). Let's mark such subset by `noinstance`,
   and make it `{-# Refine #-}`-able.
4. (I'm not sure you still want it, but) Change type checker to provide more safety for users of `Overlappable` etc.

To ensure safety, 1. is enough. 2. and 3. is to allow more without losing safety. I'd say 2. is simple enough, but 3. is too complex. Can you describe 3. formally? Can you show why that formal definition keeps safety of 1. and 2.?

Also, the instances 2. allows have no other way to declare without disabling `-Worphans`. The instances 3. allows have workaround, even if it may be ugly.

> It also seems like your proposal is incompatible with the current `-XNoFlexibleInstances -Worphans` 

Ah, I overlooked it. **You are right.** But not with that example.
(C) treats both of these instances as non-orphan, regardless of ParamOrder.

This is the disparity. MPTCs + having type variable make the distinction.

```haskell
module Foo where
  class Cast a b
module Bar where
  import Foo
  newtype F x = F [x]
  instance Cast x y => Cast (F x) [y]
  instance Cast x y => Cast [x] (F y)
```

gbaz:
What I find weird about this proposal is we have a pragma that attaches to a class that _only_ affects when warnings are fired off about instances of the class. We've never had pragmas that only affect warnings before (outside of the ones that just turn them on and off entirely).

Furthermore, as far as I can tell, this means that its possible to write warning-free instances that lack global coherence anyway, since the pragma may allow it. It's just intended that the pragma indicate how one thinks the class "should" be used. So I don't like that either, since it doesn't even prevent what it says it prevents -- just gives a complicated mechanism that pushes authors and consumers alike to understand it.

I like Edward's proposal -- just make the warning cover more cases. Just like today, if the warning covers a case, but you still want the case, you can just add a local pragma to turn off the warning. Simple!

tysonzero:
> We've never had pragmas that only affect warnings before

http://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#warning-and-deprecated-pragmas

> Furthermore, as far as I can tell, this means that its possible to write warning-free instances that lack global coherence anyway, since the pragma may allow it.

No... that is not possible. That's pretty much the entire point of the proposal, so if you think you can write warning-free code that is incoherent, then please do, so that I can find and fix the proposal appropriately.

tysonzero:
> I don't deny I was wrong (Let me take a time to update that note). I meant that (B) and (C) have incomparable set of safely acceptable instances. Even with carefully wrote noinstances.
That example I wrote did not show it successfully.

It would probably be worth trying to find such an example, ideally one that has practical value. Currently this proposal seems to cover every coherent use case I can envision (besides inherently incoherent/hard-to-keep-coherent ones mentioned by @ekmett) but your proposal would not cover use cases I would very much appreciate having available to me.

> We have to make -Worphans stricter. (@ekmett's) (A) is the simplest.

> To ensure safety, 1. is enough.

I mean yes we have to make it stricter, but I'm not a fan of the attitude that overshooting significantly is perfectly ok. I am quite excited for increased usage of `HasField` and `IsLabel` and so on, and it would feel like a bit of a slap in the face if I need to use `-fno-warn-orphans` everywhere.

The simplest way to make a dynamic language type safe is to add a fully monomorphic language to it, but even I would choose a dynamic language over a fully monomorphic language.

It's also worth noting that such an overshoot would be backwards incompatible and break perfectly safe and coherent code that has `-Worphans -Werror` enabled.

> Can you describe 3. formally? Can you show why that formal definition keeps safety of 1. and 2.?

I could probably do that at some point, don't have a whole lot of free time so can't guarantee it will be particularly timely.

> This is the disparity. MPTCs + having type variable make the distinction.

Hmm, that seems very unintuitive to me.

If I have sufficient ownership over a type constructor to put it in a constructor, e.g `Foo ... SomeType ...` and I decide to add a completely unused phantom type variable to `SomeType` and then adjust the instance to `Foo ... (SomeType a) ...` it would be very surprising if that declaration suddenly became invalid.

To me there are a few different levels of ownership that have clear reasoning for being different than one another:

1) Owned type constructor: another instance can overlap with this if they use a bare type variable or import this module, strongest form of ownership.

2) Type constructor: another instance can overlap with this if they use a bare type variable or the unowned type constructor.

3) Bare type variable: another instance can overlap with this if they use a bare type variable or an unowned type constructor or even an owned type constructor, weakest form of ownership.

Whether or not the owned or unowned type constructor contains a type variable does not matter to me, except when you explicitly define an instance with a `Refine` on that type variable to explicitly allow for going deeper.

As a side note which instance is the orphan in the following case?

```
module Bar where

data Bar

instance Foo Bar a
```

```
module Baz where

import Bar

data Baz

instance Foo Bar Baz
```

It seems like when analyzed independently by your proposal, neither are considered orphans?

So this would mean that adding a new instance could retroactively break and/or orphan-ize an existing legitimate instance?

viercc:
> It would probably be worth trying to find such an example, ideally one that has practical value.
 
https://gist.github.com/viercc/d51b0c4c7ae8333d9f82e59af2aa34c7

Yesterday I tried to search such examples. In wild. What I managed to collect are either

* I couldn't judge, since levity and kind polymorphism is used.
* All of (A), (B), (C) accepts
* Only (C) accepts
  * Theoretically (B) can accept, but the amount of required changes in dependencies (which can be across packages) makes it impractical.
  * `FromJSON` instances found in `stack` repo are notable

> It's also worth noting that such an overshoot would be backwards incompatible and break perfectly safe and coherent code that has -Worphans -Werror enabled.

You can't be perfect. I will stress again, your proposal and (C) have incomparable set of safely acceptable instances. I think there is no practical solution which does not break all safe use of FlexibleInstances. So looking at, or asking people, how much of codes in use will be affected.

> Hmm, that seems very unintuitive to me.

Yeah, I admit that as afterthought. But I'll point out (C) is not that bad.

```haskell
module Foo where
  class C a b
module Bar where
  import Data.Functor.Const
  import Foo
  
  data F
  data G x
  
  -- Day 1. These were happily accepted
  instance C Int F
  instance C F   Int
  -- Day 2. Suddenly we have to replace F with (G x). They are still accepted.
  instance C Int (G x)
  instance C (G x) Int
  -- Day 3. Suddenly we have to replace Int with Const Int x. Then it happened.
  instance C (Const Int x) (G x) -- Orphan
  instance C (G x) (Const Int x) -- Accepted

  -- Because, in unknown module Baz, there can be
  --   data Z
  --   instance C (Const Int Z) b
```

* Replacing an owning constructor with parametric one does not change orphan-ness. Always.
* Replacing an not-owning constructor with parametric one can change orphan-ness.

"It is a perfectly valid transformation in current Haskell" -- it is not. On `Day 2`, there were no possibility of incoherence. On `Day 3`, incoherence was possible. It looked like harmless, though.

Your proposal warns to module Baz regardless of whether there is a overlap. And (C) warns to the change on `Day 3`, regardless of whether there is a overlap.

> As a side note ...

Yes, (C) treats both of them as non-orphan. It can retroactively cause compile error. It is still safe (because it's compile error, not some bogus result.) I don't think this problem is in the scope of this proposal.

tysonzero:
Saw the change you made, the new set of instances is not quite right:

```
instance C (M (F x)) (M (F x)) -- (C) only
```

This should be:

```
instance C (M (F x)) (M (F x)) -- (C) only, but (B) can be customized to allow it
```

As you just need:

```
noinstance C (M a) (M b)
    {-# Refine a b #-}
```

gbaz:
We've never had pragmas that only affect warnings before **_(outside of the ones that just turn them on and off entirely)_**

>  if you think you can write warning-free code that is incoherent,  then please do so

Anything along the lines of

```
module Foo where

class Foo a 
   {-# Refine #-}
    where

module Bar where

import Foo
instance Foo Int
```

i.e. you rely on the author to give the "right" refine declaration for their intended usage. And what if... they don't?

viercc:
@tysonzero, that `noinstance` have to be in the module either `Foo` or `Middleware.Specific.Types`. Both of these modules may not be in the same package to `Bar`. If you are not the maintainer of one, you have to send patch and pray. Or if that was OSS, clone it to local repo and patch yourself. This is what I repeatedly said "theoretically possible but practically impossible".

int-index:
> We've never had pragmas that only affect warnings before (outside of the ones that just turn them on and off entirely)

[`{-# MINIMAL ... #-}`](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#minimal-pragma)?

tysonzero:
> Yesterday I tried to search such examples. In wild. What I managed to collect are either

Can you break down the examples into small representative pieces? So far the two "(C)-only" examples I have seen were very compatible with this proposal.

I have given `Convert` and `LeftModule` as examples that don't work at all with (C), and `IsLabel` and `HasField` which don't work without your ordering addendum (that I would say makes it just as complicated as this proposal).

> You can't be perfect. I will stress again, your proposal and (C) have incomparable non-comparable sets of safely acceptable instances.

I still haven't seen an instance that your proposal accepts that this proposal does not. I'm not saying I doubt there existence entirely, but I would like to see one, and I would like to assess it's practical significance.

> "It is a perfectly valid transformation in current Haskell" -- it is not. On Day 2, there were no possibility of incoherence. On Day 3, incoherence was possible. It looked like harmless, though.

I should have clarified that by current Haskell I meant without the broken FlexibleInstances. Yes a lot of transformation are not valid with FlexibleInstances. My bad.

> Your proposal warns to module Baz regardless of whether there is a overlap. And (C) warns to the change on Day 3, regardless of whether there is a overlap.

Well with this proposal you would choose what you want when you define `class Foo a b` and based on that choice only one of those instances would be valid: `{-# Refine a #-}` would support the `Bar` instance, and `{-# Refine a b #-}` or `{-# Refine b #-}` would support the `Baz` instance.

> Yes, (C) treats both of them as non-orphan. It can retroactively cause compile error as just like today. Even when -Wno-orphans is specified, GHC do and should reject it. I don't think this problem is in the scope of this proposal.

GHC/Haskell today does NOT suffer from this unless you turn on `FlexibleInstances`, defining a new non-orphan non-flexible instance will not break an existing non-orphan non-flexible instance.

This is also quite important to me and part of why I wrote this proposal. I do not want it to ever be a race between two libraries to define an instance, where the first to define it makes the other instance invalid to define.

My proposal (and existing GHC/Haskell without FlexibleInstances) achieves that goal.

tysonzero:
> We've never had pragmas that only affect warnings before (outside of the ones that just turn them on and off entirely)

Neither of those really "turn them on and off entirely", they just decide that a certain action (using the specific value) causes a warning. It's pretty much the same idea, although I will agree it's obviously more complex than just "if you use this value then emit a warning".

Also `MINIMAL` as mentioned.

> Anything along the lines of ...

That would absolutely throw a warning, unless you disabled `-Worphans`, which is known to risk coherence and not something any proposal mentioned is trying to address.

I'm not sure what exactly you think this proposal does? Because throwing a warning when given code like that is pretty much the exact point of this proposal.

tysonzero:
> at noinstance have to be in the module either Foo or Middleware.Specific.Types. Both of these modules may not be in the same package to Bar. If you are not the maintainer of one, you have to send patch and pray. Or if that was OSS, clone it to local repo and patch yourself. This is what I repeatedly said "theoretically possible but practically impossible".

This apparent pain is all for very very good reason. What if at some point in the future `M` decided it wanted to define it's own `instance C (M a) (M b)`? Adding a very sane and reasonable instance is now a breaking change, which as a side note violates the [PVP](https://pvp.haskell.org/) which declares adding a non-orphan instance to be non-breaking.

If you really want that instance with my proposal you can just add `-fno-warn-orphans`, and if/when `M` adds that instance, you will have to deal with it. This will hopefully also prompt you to tighten up the dependency version range on the library that defines `M` so that you don't violate the PVP.

gbaz:
> Refine takes in a list of zero or more type variables from the class/instance head.
> 
> Instances of this class, or instances that overlap with this instance, must match on a top level concrete type constructor for all of the above type variables, and they must have ownership of at least one of those type constructors. Any type variables not mentioned in the Refine clause have no restrictions and can be left bare if desired.

The above has an empty refine clause and so instances can be left bare, as per the proposal.

tysonzero:
> The above has an empty refine clause and so instances can be left bare, as per the proposal.

>  and they must have ownership of **at least one** of those type constructors

In such a case you clearly do not have ownership of "at least one" of those type constructors, in fact an empty `Refine` clause just means you can't define any further instances, hence why it is the default for `instance` clauses where overlap is the exception and not the rule.

gbaz:
In my reading of the proposal, instances must own a constructor for one of the type variables named in the refine clause of the class. The class has an empty refine clause.

This is why the proposal you wrote does _not_ specify the default refine clause for a class as empty, but instead as specifying all type variables: "By default classes and noinstances will have a Refine that specifies all type variables in the class/instance head"

tysonzero:
> In my reading of the proposal, instances must own a constructor for one of the type variables named in the refine clause of the class. The class has an empty refine clause.

Right... so you can't define any instances... because you can't own a constructor for **one** of the type variables named in the refine clause.

Just because a condition is impossible to meet, does not mean you can ignore it.

gbaz:
So if there are two variables in the clause, one must own for two. If there is one variable in the clause, one must own for one. If there are zero in the clause, by induction, one must own for...?

I guess you intend for it to mean something other than the proposal specifies.

Here is a further question: in the proposal there is, `instance Foo [Bar] -- orphan`. Is this an orphan because of the `Foo [a]` instance already existing, or would it be an orphan regardless?

tysonzero:
> So if there are two variables in the clause, one must own for two. If there is one variable in the clause, one must own for one. If there are zero in the clause, by induction, one must own for...?

Right, by "induction" one must also own for zero, since you must always own one, that never changes.

This is of course impossible. Which is a good thing and exactly what I want, and exactly what the proposal specifies.

> I guess you intend for it to mean something other than the proposal specifies.

No. I mean exactly what the proposal specifies. You **must** own **one** type constructor mentioned in the `Refine` clause. So if you do not own **one** type constructor then it is not a valid instance.

If I say that to call a function you must give me a number greater than 5 and less than 3, you can't just give me 7 and say "well it's impossible for it to be less than 3, so i'm ignoring that", you just can't call that function.

> Here is a further question: in the proposal there is, instance Foo [Bar] -- orphan. Is this an orphan because of the Foo [a] instance already existing, or would it be an orphan regardless?

It's an orphan because you do not own `[]`, and there is no `noinstance Foo [a]` or `instance Foo [a] {-# Refine a #-}` instance available for you to utilize.

So it would be an orphan regardless, you explicitly would need the owner of the class (or `[]`) to give you permission to define such an instance via the above mechanism.

gbaz:
Ok, this clears up the confusion. 

The argument seems to run as follows: If one owns the head of a single parameter class, all is good. In a multiparameter class, we could require that one owns the head for at least one parameter, and the others are at least narrowed down. This is the naive warning. For certain special classes, it may be that we want things to be controlled by one specific parameter or collection of them, so refine lets the compiler "pretend" that the class is only about those specific parameters for the purposes of detecting overlap.

I think that an alternate approach could be to specify the fundep interaction better, and "secretly" attach a refinement clause to mptcs with fundeps that refines only the variables on the left hand side of fundeps. That way there is no surface syntax at all, but there are perhaps a few warnings people might not like on mptcs without fundeps (but those cases should be rare, and the warnings can be turned off).

viercc:
> So far the two "(C)-only" examples I have seen were very compatible with this proposal.

Your proposal can cover any finite examples by "just" writing at most O(n) `noinstance`s. (C) is not perfect (I said it every time) but accepts *infinite* instances, which require infinite `noinstance`s to cover.

Let's stop "Who can say the largest number" game since that's not productive at all.

(Side note: discussion here went very long, and tracking common basis like alternatives and concrete examples became difficult. Could you update the proposal itself?)

> Well with this proposal you would choose...

First of all, this discussion already has conclusion: (C) does not comply with current MPTCs without `FlexibleInstances`. Your proposal does, that's an advantage.

Sorry, that's me who derailed discussion by rushed mind. You replied along with derailed discussion (which is more customizable / intuitive?)

> Adding a very sane and reasonable instance is now a breaking change, ...

Suppose there are developers of `instance C (M x) (M x)` (suppose her name is Lea, library developer) and `instance C (M (F x)) (M (F x))` (suppose her name is Alice, application developer).

* With your proposal, Alice can't write that instance without warning, until her request to Lea is done.
  * This request can be rejected if Lea didn't want that instance. Lea might not want to depend to package containing `C`, and that's reasonable. Lea might got similar request for Binary, FromJSON, ToJSON, Hashable, Storable, FromYAML, Lift, ...
  * Alice didn't request but did `-Wno-orphans`. Then what happens is same to current GHC.
* With (C), or currently, Alice's application suddenly stops to build because of *minor* version up made by Lea. Alice complained Lea and is told to put `OVERLAPPING` pragma.

Comparing both scenario, I would say (C) and the current situation is better, though that's concerning it was minor version up. Same can happen when any unqualified import is used, so you can say instances like Alice defined is as bad as unqualified imports.

tysonzero:
> I think that an alternate approach could be to specify the fundep interaction better, and "secretly" attach a refinement clause to mptcs with fundeps that refines only the variables on the left hand side of fundeps.

I don't see how that helps, none of these MPTCs (`Cast`, `IsLabel`, `HasField`) want to have a functional dependency, they would all be more or less ruined if you added one.

tysonzero:
> Your proposal can cover any finite examples by "just" writing at most O(n) noinstances. (C) is not perfect (I said it every time) but accepts infinite instances, which require infinite noinstances to cover.

> Let's stop "Who can say the largest number" game since that's not productive at all.

I'm fine with that, but I still would like you to adjust your proposal to acknowledge the fact that for any case with a depth bounded by some K, (B) can do everything (C) can do. So the example in the README (`instance C (M (F x)) (M (F x))`) should be updated.

> With your proposal, Alice can't write that instance without warning, until her request to Lea is done.

Well no... she can just use `-fno-warn-orphans` and define it immediately.

If she cares about the PVP and general correctness/compatibility, and she thinks its reasonable for Lea to depend on `C`, then she submits the request to Lea and waits.

If she sort of cares but not THAT much, she can use `-fno-warn-orphans` and submit the request to Lea, then remove the pragma once Lea accepts the proposal (or leave it if she never does).

All of the above seems pretty clearly better than (C) or the status quo. Alice doesn't accidentally violate the PVP or break her users code without warning, but she can easily opt out of this mechanism the second it's too inconvenient.

gbaz:
`HasField` tends to have a fundep: https://hoogle.haskell.org/?hoogle=HasField&scope=set%3Astackage

`Cast` would warn on both, just as in your proposal.

`IsLabel` isn't helped by this, as Edward describes above.

tysonzero:
Sorry forgot that `HasField` was (still) `FD`-based and not `TF`-based.

Regardless the FD in `HasField` doesn't help you. As you still need to use Symbols you don't own and sometimes a bare type variable for the Symbol slot, and that Symbol slot appears on the LHS of an FD.

The biggest red flag for me as to why an FD-based approach won't be satisfying, is that personally I never use FD's, but yet I still have the need for the proposal, so clearly the FD-based solution will leave me in the exact same spot that I am now.

tysonzero:
> (Side note: discussion here went very long, and tracking common basis like alternatives and concrete examples became difficult. Could you update the proposal itself?)

Yeah I can probably do so in not too long, just will require a bit of planning so I can cover the various things we discussed.

gbaz:
> Regardless the FD in HasField doesn't help you. As you still need to use Symbols you don't own and sometimes a bare type variable for the Symbol slot, and that Symbol slot appears on the LHS of an FD.

What if we solve this and the `IsLabel` thing in one swoop and just don't enforce the "specify top level constructor" on things of kind symbol, since it makes no sense there/is not necessary?

AntC2:
> (Side note: discussion here went very long, ...

It has, and sorry I've abandoned hope of keeping up. I did see:

> The biggest red flag for me as to why an FD-based approach won't be satisfying, is that personally I never use FD's, ...

`HasField` nearly got defined using Type Families rather than FD's. Really the decision turned on practicalities of generating the auto-instances, and supporting user-defined instances.

In terms of type improvement and interaction with overlapping/orphan instances, there's no difference between FD's vs Type Families (including with `~` constraints). If you "never use FD's" and never use Type Families, I think you're not in a good position to tackle the topic of incoherence amongst sets of instances.

I don't like to blow my own trumpet, but the proposal for 'Instance Apartness Guards' (see my earlier comment) needed to go into a great deal of detail. FD's/TF's were tractable only because the guard was explicit on each instance. Note that you might have multiple FD's on a class (or multiple superclass constraints using Type Families -- esp because `~` constraints point in both directions at the same time), which mean that it's difficult/impossible to ascribe an owner to a class's parameter position.

There's no help from any of the type theory here: the academic papers essentially all assume no overlapping instances; and assume global visibility of all instances.

tysonzero:
> What if we solve this and the IsLabel thing in one swoop and just don't enforce the "specify top level constructor" on things of kind symbol, since it makes no sense there/is not necessary?

How would that interact with `PolyKinds`? I would be open to the idea of having `Refine` default to all type variables with "open" kinds (like `Type`), and ignore those with "closed" kinds (like `Symbol`).

tysonzero:
> In terms of type improvement and interaction with overlapping/orphan instances, there's no difference between FD's vs Type Families (including with ~ constraints). If you "never use FD's" and never use Type Families, I think you're not in a good position to tackle the topic of incoherence amongst sets of instances.

I never use `FD`'s because I DO use `TypeFamilies` and generally prefer them. I also understand FD's and TF's well enough to know my proposal works as-is with them. I am aware that the default `Refine` might be inconvenient for `FD` users, and for those using `TF`'s + `~` for `FD`-like behavior, but I don't think it's worth the complexity to try and address that issue.

gbaz:
> How would that interact with PolyKinds? I would be open to the idea of having Refine default to all type variables with "open" kinds (like Type), and ignore those with "closed" kinds (like Symbol).

I would assume we just check at instantiation site in such a case.

Note, btw, that this proposal seems to only affect type classes, and as far as I can tell has no relationship to type families, which always have stricter rules regardless.

tysonzero:
> I would assume we just check at instantiation site in such a case.

How would that work in the following case?

```
class Foo (a :: k)
class Bar (a :: k) b
instance Foo a => Bar a Int
```

> and as far as I can tell has no relationship to type families

It's important to consider how well it plays with associated type/data families and with FD's, even if it is not directly targeted at them.

PR #280, opened 2019-10-02
Extended Typed-Holes

Tritlo:
This proposals suggests extending the current typed-holes functionality of GHC.

[Rendered](https://github.com/Tritlo/ghc-proposals/blob/master/proposals/0000-extended-typed-holes.md)


isovector:
@TOTBWF you might be interested in this.

goldfirere:
Under the restriction that GHC is used as a single-shot or batch compiler, this seems reasonable. But I wonder what riches would await if we could interact with GHC more interactively.

I suppose, given that these holes won't remain in production code, forward compatibility is less of a problem.

simonpj:
I though typed holes were OK -- but I had no idea how popular they'd be.

For this extension I'm sceptical that it's really worth all this work.  But I have been proved wrong before; and the cost is not pervasive -- it's well localised to error handling.  So I would not want to stand in the way.

But I'd love to see evidence of typed-holes users standing up to say "yes, this would really improve my life".

TOTBWF:
I've been working on a tactic system for haskell on and off for a while now, and this would be _huge_. Having access to the type at a given point is a big win already, but being able to suggest the completion would bring this from a neat proof of concept to something I would legitimately use.

robrix:
I rely on typed holes quite heavily in my day-to-day work, but I’ve yet to use any sort of plugin to interact with them. Having a richer interface to interact with plugins might encourage more of that in the short-term, but either way, I think the potential value of this feature will be hard to assess up front, because so much of it will be delivered by the plugins using these interfaces.

Given that it’s well localized (per @simonpj’s comment, above), and given that it’s enabling fairly open-ended improvements to interactions with plugins, I’d be in favour of letting the ball roll—I’d love to see what this ends up enabling plugin developers to do.

I do wonder if having it gated behind a language extension is a) necessary, given that these expressions aren’t going to end up in running programs (at least, without deferred errors?), and b) impairing the utility of the feature given that it’s already gated on having an appropriate plugin available and selected. At the very least, as a relative newcomer to `ghc`’s implementation, I’d enjoy learning more about the rationale for requiring the extension.

Very creative, very cool, and I am super curious about the potential for improving the interface between my editor and `ghc`!

mpickering:
It doesn't seem likely to me that it will be worthwhile to annotate each individual hole with all this precise information if you are using a traditional `save -> compile` loop. Each hole is ephemeral so you only temporarily type all this information and then you'll spend more time writing out the hole each time than it takes to look up what definition you meant by some external means. 

I can think of two other designs.

1. A way to configure how typed holes should behave on a module by module basis. In fact, this already exists with plugins. 
2. If you want the user to direct the resolution of the hole then the "program source" is not the correct place to do it. You really want the editor to interact with a typed hole plugin in order to invoke and search different possibilities interactively. One possible implementation is to defer resolving all typed holes until after typechecking is complete, then write a typechecker plugin which traverses the source looking for the holes and then starts a dialog with the user about what should go into the hole. 

Overall I think it would be better to concentrate on providing the right API for an interactive editor experience than adding new syntax to the language for this niche.

Tritlo:
@mpickering 

> It doesn't seem likely to me that it will be worthwhile to annotate each individual hole with all this precise information if you are using a traditional `save -> compile` loop. Each hole is ephemeral so you only temporarily type all this information and then you'll spend more time writing out the hole each time than it takes to look up what definition you meant by some external means.
> 

Part of the idea is that the syntax for extended typed-holes matches that of template haskell and paranthesized expression. By introducing an underscore before a parenthisized expression, you can send that expression along to the typed-hole plugin for analysis and to further inform the plugin what you might want. Then, most of the contents would remain in the hole, but could be swapped out. 

> I can think of two other designs.
> 
> 1. A way to configure how typed holes should behave on a module by module basis. In fact, this already exists with plugins.

This does not allow you to send more information about particular holes, at least not in any nice way that I can think of.

> 2. If you want the user to direct the resolution of the hole then the "program source" is not the correct place to do it. You really want the editor to interact with a typed hole plugin in order to invoke and search different possibilities interactively. One possible implementation is to defer resolving all typed holes until after typechecking is complete, then write a typechecker plugin which traverses the source looking for the holes and then starts a dialog with the user about what should go into the hole.
> 

This also works, but would require a lot of integration with an editor, and then we'd have to go into the whole "which editor to support" debate, which I don't think we should be having. 

> Overall I think it would be better to concentrate on providing the right API for an interactive editor experience than adding new syntax to the language for this niche.

Yes, it would be really nice to have a proper API for an interactive editor experience, but these extended typed-holes give you a lot already, and they're completely editor agnostic.

mnn:
I am confused, this PR is open, but a link in the first post returns 404. Last commit seems to be a deletion of the whole document?

Tritlo:
Ah sorry, that was my bad. A bit of a git mixup when renaming it to "Non-Empty" instead of "Extended".

dpvanbalen:
I'm confused: the example includes the following suggestions

`Valid hole fits include`
`        (\ (_, a) -> a)`
`        (\ (_, a) -> g (head (cycle (([]) ++ ([]))), a))`

Since when does GHC infer such complicated fits? Is this related to the proposal or something else that I missed?

Tritlo:
This is from the `Djinn` part of the [`NonEmptyHoles` plugin](https://github.com/Tritlo/ExampleHolePlugin/tree/master/non-empty-holes-plugin),  being invoked by:
```haskell
h = _$$( execTyped $ do
         filterBy "Prelude"
         invoke Djinn)1
```

googleson78:
Now that we have things like [hls support for agda-style refinement](https://github.com/haskell/haskell-language-server/releases/tag/1.0.0) (introduce constructor for example), this feature would be even more useful.

Some example common usages of this in Agda include:
* ! giving names of arguments on which we want to do a case split
* giving hints about which identifiers to use to the auto program search mechanism
* giving information/flags to the auto program search mechanism (e.g.  "search with a fixed depth")
* giving a name for a function which to  "refine with" - if the return type of the function would match, Agda inserts the function with holes for its arguments (this is actually way more convenient than it sounds in my experience)

PR #284, opened 2019-10-15
Bidirectional Instanes

KoenP:
An extension that lets the compiler treat selected type class instances as bi-implications, which improves the interaction between type classes and GADTs.

[Rendered](https://github.com/KoenP/ghc-proposals/blob/patch-1/proposals/0000-bidirectional-instances.md)

### DIFF
--- a/proposals/0000-bidirectional-instances.md
+++ b/proposals/0000-bidirectional-instances.md
@@ -1,0 +1,334 @@
+---
+author: Koen Pauwels
+date-accepted: ""
+proposal-number: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/0>).
+**After creating the pull request, edit this file again, update the number in
+the link, and delete this bold sentence.**
+
+# Bidirectional Type Class Instances
+
+The interaction between type classes and GADTs can lead to some unpleasant
+surprises. The source of this problem is the fact that GHC treats
+instance contexts only as sufficient conditions, while any
+(non-overlapping) instance is also automatically a necessary one.
+
+## Motivation
+
+Consider the following GADT:
+```haskell
+data Term a where
+  Con :: b -> Term b
+  Tup :: Term b -> Term c -> Term (b,c)
+```
+
+A (simplified) implementation for a `Show` class instance might look like this:
+```haskell
+instance Show d => Show (Term d) where
+  show (Con x)   = show x
+  show (Tup x y) = concat ["(", show x, ",", show y, ")"]
+```
+
+Although the implementation seems quite reasonable at first glance, it is
+rejected by the compiler. The wanted constraints from the `Tup` clause
+`(Show b, Show c)` 
+cannot be deduced from the given constraints `(Show d, d ~ (b,c))`.
+This is because GHC considers instance declarations to be unidirectional
+implications from the instance context to the instance head. So the `Show`
+instance for pairs only introduces an axiom of the form
+`∀ a b. (Show a, Show b) => Show (a,b)`.
+But since overlapping instances are
+disallowed by default, instance declarations are in fact bi-implications (the
+conditions in the instance context are not only necessary but also sufficient).
+Our proposal is to allow the compiler to treat them as such (by way of a
+per-instance declaration pragma), which would be enough to resolve the constraints
+for the given example and type check the program.
+
+## Proposed Change Specification
+
+A detailed account of the changes to type system and elaboration required to
+support this extension can be found in our paper
+[Bidirectional Type Class Instances](https://arxiv.org/abs/1906.12242).
+We summarize the key ideas here.
+
+### Type Inference
+Each instance declaration tagged as bidirectional will give rise to an additional
+"converse" instance axiom for each constraint in its instance context.
+More concretely,
+an instance declaration of the form `instance (C1, ..., Cn) => C t`
+(where `t` is a metavariable for any monotype, and
+`C`, `C1`, ..., `Cn` are metavariables that stand in for constraints)
+gives rise to the `n` converse axioms `C t => C1`, ..., `C t => Cn`.
+
+These converse instance axioms are treated in much the same way as superclass
+axioms for the purpose of type checking; that is, they are used to extend
+the set of given constraints.
+
+### Elaboration
+Each type class declaration
+```haskell
+class ... => C a where
+  f :: T
+```
+(where `T` is a metavariable)
+generates a type declaration declaration `type F_C a` in the intermediate language.
+In Core, the dictionary data type associated with each type class is extended
+with an additional "context" field, the type of which is given by the
+type family declaration.
+In pseudo-Core:
+```haskell
+data Dict_C a = Dict_C { f   :: T'    -- T' is the elaborated version of T
+                       , ctx :: F_C a
+                       }
+```
+
+Each instance declaration `instance (C1, ..., Cn) => C t` generates a type
+equality axiom asserting the equality between the types `F_C t` and a product of
+dictionary data types `(D1, ..., Dn)` corresponding to the constraints `C1, ... Cn`.
+
+The context field is used to define the dictionary transformers associated with the
+converse instance axioms. 
+To construct a dictionary transformer for, say, the axiom `C t => C1`, the context
+field is extracted from the dictionary for `C t` and then coerced to the type of the context
+dictionary tuple `(D1, ... Dn)`, from which the relevant dictionary (`D1`) is projected.
+
+### Syntax
+We propose the addition of a new pragma `{-# BIDIRECTIONAL #-}` which is used 
+in a similar way to the `{-# OVERLAPPING #-}` pragma: it tags an individual
+instance declaration as bidirectional.
+
+## Examples
+
+### Structural Induction Over Indexed Data Types
+The problem with the motivating example is solved when the `Show` instances for 
+pair types and `Term` types are tagged as bidirectional.
+(We use a simplified `Show` class for brevity, where `show :: a -> String`
+is the only method).
+```haskell
+instance {-# BIDIRECTIONAL #-} (Show a, Show b) => Show (a,b) where
+  show (a,b) = concat ["(", show a, ",", show b, ")"]
+```
+
+This gives rise the following type equality axiom in the intermediate language:
+```haskell
+axiom gShowPair a b : F_Show (a,b) ~ (Dict_Show a, Dict_Show b)
+```
+
+The `Show` instance for pairs is elaborated in the following dictionary,
+where the `gShowPair` axiom is used to store the instance context in the dictionary:
+```haskell
+dShowPair :: Dict_Show a -> Dict_Show b -> Dict_Show (a,b)
+dShowPair da db = Dict_Show
+  { show = \(a,b) -> concat ["(", show da a, ",", show db b, ")"]
+  , ctx  = (da, db) ▷ sym gShowPair
+  }
+```
+
+The `Show` instance for the `Term` GADT is then elaborated into:
+```haskell
+dShowTerm :: Dict_Show d -> Dict_Show (Term d)
+dShowTerm d = Dict_Show
+  { show = \t -> case t of 
+      Con x   -> show d x
+      Tup x y -> let (da, db) = ctx d ▷ gShowPair
+                 in concat ["(", show da x, ",", show db y, ")"]
+  , ctx = undefined -- only defined if the Show instance for Term is made bidirectional
+  }
+```
+
+### Example: Functional Dependencies and Associated Types
+Bidirectionality also raises the expressive power of functional dependencies.
+Consider this definition of length-indexed vectors:
+```haskell
+data Nat :: * where
+  Z :: Nat
+  S :: Nat -> Nat
+
+data Vec :: Nat -> * -> * where
+  VN :: Vec Z a
+  VC :: a -> Vec n a -> Vec (S n) a
+```
+
+In order to define an `append` function for this vector data type, we have to
+define addition at the type level. This can be encoded with a multi-parameter
+type class and a functional dependency (a.k.a. playing "type-level Prolog").
+```haskell
+class Add (n :: Nat) (m :: Nat) (k :: Nat) | n m -> k
+instance              Add Z     m m
+instance Add n m k => Add (S n) m (S k)
+
+append :: Add n m k => Vec n a -> Vec m a -> Vec k a
+append VN        ys = ys
+append (VC x xs) ys = VC x (append xs ys)
+```
+However, this also fails to compile without bidirectional instances, because
+the second clause of the `append` definition requires the constraint
+`Add n' m k'`, while the type signature provides `Add (S n') m (S k')`.
+
+This is easily resolved with bidirectionality; the only change needed is to
+add bidirectionality to the "successor" instance of `Add`:
+```haskell
+instance {-# BIDIRECTIONAL #-} Add n m k => Add (S n) m (S k)
+```
+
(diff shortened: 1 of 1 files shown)


phadej:
Note: there's a workaround to `append` problem:

```haskell
{-# LANGUAGE GADTs, FunctionalDependencies #-}
{-# LANGUAGE DataKinds, KindSignatures #-}
{-# LANGUAGE FlexibleInstances, UndecidableInstances #-}

data Nat :: * where
  Z :: Nat
  S :: Nat -> Nat
data Vec :: Nat -> * -> * where
  VN :: Vec Z a
  VC :: a -> Vec n a -> Vec (S n) a

class Add (n :: Nat) (m :: Nat) (k :: Nat) | n m -> k where
instance              Add Z     m m -- one could also have m ~ k => Add Z m k instance, which behaves better
instance Add n m k => Add (S n) m (S k)

class Add n m k => Append (n :: Nat) (m :: Nat) (k :: Nat) | n m -> k where
    append :: Vec n a -> Vec m a -> Vec k a

instance Append 'Z m m where
    append VN        ys = ys

instance Append n m k => Append ('S n) m ('S k) where
    append (VC x xs) ys = VC x (append xs ys)
```

int-index:
Bidirectional type classes can be encoded in today's Haskell with `UndecidableSuperClasses`. Here's an encoding of the example from the proposal:

```
{-# LANGUAGE GADTs, TypeFamilies, ConstraintKinds,
             UndecidableInstances, UndecidableSuperClasses #-}

import qualified Prelude
import Prelude hiding (Show(show))
import Data.Kind (Type, Constraint)

type family ShowCtx a :: Constraint

-- Bidirectional 'Show'
class ShowCtx a => Show a where
  show :: a -> String

type instance ShowCtx Bool = ()
instance Show Bool where
  show = Prelude.show

type instance ShowCtx (a, b) = (Show a, Show b)
instance (Show a, Show b) => Show (a, b) where
  show (x, y) = "(" ++ show x ++ "," ++ show y ++ ")"

data Term :: Type -> Type where
  Con :: a -> Term a
  Tup :: Term a -> Term b -> Term (a,b)

type instance ShowCtx (Term a) = Show a
instance Show a => Show (Term a) where
  show t = case t of
    Con x -> unwords ["Con", show x]
    Tup x y -> unwords ["Tup", show x, show y]
```

```
ghci> show (Tup (Con True) (Con False))
"Tup Con True Con False"
```

Note that the presence of the type family rules out overlapping instances, and the operational semantics are the exact match of your proposal (an additional field in the dictionary).

----

My conclusion is that the proposal can be rephrased as new syntax for an existing feature.

KoenP:
@int-index That is very interesting! I am not yet familiar with GHC's internals, so forgive the potentially ignorant question: is it possible to implement your approach without implicitly turning on `UndecidableSuperClasses` in every module that uses a bidirectional instance?

int-index:
> That is very interesting! I am not yet familiar with GHC's internals, so forgive the potentially ignorant question: is it possible to implement your approach without implicitly turning on `UndecidableSuperClasses` in every module that uses a bidirectional instance?

For this encoding, `UndecidableSuperClasses` is only needed at class definition site, for this part:

```
type family ShowCtx a :: Constraint
class ShowCtx a => Show a where
  show :: a -> String
```

Modules that define instances only require `TypeFamilies` to define `ShowCtx` equations:

```
type instance ShowCtx (a, b) = (Show a, Show b)
instance (Show a, Show b) => Show (a, b) where
  show (x, y) = "(" ++ show x ++ "," ++ show y ++ ")"
```

And modules that use these instances require no extensions.

proposals/0000-bidirectional-instances.md
```diff
@@ -0,0 +1,360 @@
...
+rejected by the compiler. The wanted constraints from the `Tup` clause
+`(Show b, Show c)` 
+cannot be deduced from the given constraints `(Show d, d ~ (b,c))`.
+This is because GHC considers instance declarations to be unidirectional
+implications from the instance context to the instance head. So the `Show`
+instance for pairs only introduces an axiom of the form
+`∀ a b. (Show a, Show b) => Show (a,b)`.
+But since overlapping instances are
+disallowed by default, instance declarations are in fact bi-implications (the
+conditions in the instance context are not only necessary but also sufficient).
```
ulysses4ever: Not that it's very important, but I think it should be the other way around: the conditions used to be just sufficient and now they are considered necessary, therefore they can be assumed given the head.

goldfirere:
I agree with @int-index in that this proposal essentially provides syntactic sugar for the approach in https://github.com/ghc-proposals/ghc-proposals/pull/284#issuecomment-542322728. That's not to malign the proposal -- after all, `do` is "just" syntactic sugar. But I believe that the two approaches are entirely equivalent.

However, this suggests a possible improvement: you could think of labeling a *class* as `{-# BIDIRECTIONAL #-}`. This would add the bidirectionality machinery to the class dictionary type. Only bidirectional classes can have bidirectional instances. If a class isn't labeled as bidirectional, then no bidirectional instances are allowed. This would also mitigate the performance worry, because it makes the whole system opt-in. Perhaps a bidirectional class implies that all instances are also bidirectional, but it also might be nice to have some non-bidirectional instances of a bidirectional class, in case a user wants some overlapping instances. (In the constraint-kinds approach, this just means a few missing associated type instances.)

One other concern: GHC currently compiles one-element classes into newtypes. (Here, "one-element" means either one method xor one superclass.) This feature is relied upon to implement e.g. the `reflection` package. Although GHC does not formally advertise this compilation strategy, it would be disruptive to change it, yet this proposal changes it (by adding another superclass to every class).

KoenP:
@int-index @goldfirere
I agree that the two approaches are essentially equivalent, but I have some lingering concerns about rephrasing the implementation part of the proposal as syntactic sugar.

Firstly, I'm not entirely sure I understand @int-index' comment about `UndecidableSuperClasses` being needed only at class definition site. While this is certainly true, I estimate it is not a small change to the proposal and de facto identical to enabling `UndecidableSuperClasses` everywhere, since
1. we enable `UndecidableSuperClasses` in every module with a class declaration, and
2. `UndecidableSuperClasses` only affects how the compiler handles class declarations.

In other words, the only modules where the extension would not be enabled are those modules that would be unaffected by it either way. @goldfirere 's suggestion of also having a class-level pragma for opting in or opting out of supporting bidirectional instances alleviates some of this, but I think disallowing bidirectional instances at the class level should be a rare occurrence.

Secondly, I'm not sure how a syntactic sugar implementation would interact with error messaging. Can an error in the class declaration be displayed to the user without showing the generated Foo_Ctx constraint? Can we still emit an error message if the user's class declaration violates one of the constraints that is disabled by `UndecidableSuperClasses`? If these issues can be addressed, will the solution be cleaner than a direct implementation?

goldfirere:
The "syntactic sugar" part is (for me, at least) merely an observation. I'm not saying that the proposal should be reworded in that way or that the implementation should be designed around a desugaring. Certainly, I don't want to enable `-XUndecidableSuperClasses` everywhere. If anything, the syntactic-sugar angle strengthens this proposal, in that we have a model for how to understand and implement it.

Separately, I see two reasons an individual class might want to avoid the new behavior:

1. The new behavior requires adding a field to every dictionary of the class. This shouldn't lead to a very high performance burden, but it is a runtime performance burden that everyone will pay for (for the benefit of only a few).

2. Without a way to opt in or out, I don't see how this proposal is compatible with e.g. the `reflection` package's treatment of class-newtypes.

KoenP:
Ah, I see, indeed I misunderstood the comment about the syntactic sugar as a suggestion for an implementation method.

The small but ubiquitous runtime overhead is probably the hardest sell in the proposal, so I very much agree with your suggestion to add an opt-in or opt-out at the class level.

I'm currently picturing it more as an opt-out. I think that one of the core features of type classes is their openness: the author of the type class does not need to know which types will be made instances, therefore in the common case, the class author will not have the information needed to make a call on whether bidirectional instances should be allowed or not for their class. There are exceptions, where the class author has a very clear idea of what their class is going to be used for, in these cases they may optimize by opting out of bidirectional instances at the class level.

I think an analogy with strictness/unpack annotations for data type fields is apt here: the default behavior (non-strict, boxed data) incurs a performance penalty to be as general as possible, but the data type author may choose to give up the freedom to have a certain field computed lazily to ensure a more efficient representation.

proposals/0000-bidirectional-instances.md
KoenP: Thanks!

goldfirere:
I looked at the updated proposal, but I still don't see this opt-out mechanism. Also, the specification does not say what will happen with non-`{-# BIDIRECTIONAL #-}` instances. Furthermore, I think it would be a bit awkward to have classes be opt-out, while instances are opt-in.

Bj0rnen:
Just a thought from skimming this. Would it be possible to have the compiler generate instances, on-demand, for something that is basically the reverse of the `constraints` library's [`:=>` operator](https://hackage.haskell.org/package/constraints-0.12/docs/Data-Constraint.html#t::-61--62-)?

Let's define it as
```haskell
class b :<= h | h -> b where
  ctx :: h :- b
```
Then one could work around the problem from the proposal just like this:
```haskell
data Term a where
  Con :: b -> Term b
  Tup :: Term b -> Term c -> Term (b,c)

instance Show d => Show (Term d) where
  show (Con x)   = show x
  show (Tup (x :: Term b) (y :: Term c)) =
    concat ["(", show x, ",", show y, ")"]
    \\ ctx @(Show b, Show c) @(Show (b, c))
```
Much like `Typeable`, GHC would just conjure the instance
```haskell
instance (Show a, Show b) :<= Show (a, b) where
  ctx = ...  -- Compiler magic.
```
as long as it can find the matching instance head. It seems, then, that overlapping instances won't cause any troubles; only one instance could have a matching head. And the cost will be local to the use-site of `ctx` + a slightly larger binary I suppose. Again, just like `Typeable`.

Now, I'm not sure that this particular way is the best one to expose this functionality (it certainly can't depend on `constraints`). But if what I'm showing here is truly possible and correct, then it at least suggests that there is a way that doesn't require the foresight of annotating classes or even instances, and which only carries a runtime cost where it's being used.

Am I missing something?

Icelandjack:
I didn't feel the need for this extension until [I needed `FunctorOf (Kleisli m) (->) f` to entail `Monad m`](https://gitlab.haskell.org/ghc/ghc/issues/17981). Knowing `Functor f` means that `Category (Kleisli m)` and `Category (->)`. `Kleisli m` is only a category when `Monad m` holds.

```haskell
class (Category (Dom f), Category (Cod f)) => Functor (f :: d -> c) where
 type Dom (f :: d -> c) :: Cat d
 type Cod (f :: d -> c) :: Cat c
 fmap :: Dom f a a -> Cod f (f a) (f a')
--       =====        =====
--         ^   arrows   ^

class    (Functor f, dom ~ Dom f, cod ~ Cod f) => FunctorOf dom cod f
instance (Functor f, dom ~ Dom f, cod ~ Cod f) => FunctorOf dom cod f
```

vdukhovni:
> Bidirectional type classes can be encoded in today's Haskell with `UndecidableSuperClasses`. Here's an encoding of the example from the proposal:
> 
> ```
> {-# LANGUAGE GADTs, TypeFamilies, ConstraintKinds,
>              UndecidableInstances, UndecidableSuperClasses #-}
> 
> [...]
> type instance ShowCtx Bool = ()
> instance Show Bool where
>   show = Prelude.show

However, it is not by accident that the [example](https://github.com/ghc-proposals/ghc-proposals/pull/284#issuecomment-542322728) defines `show` for `Bool`, and then only demonstrates `show` for a `Term` with Bool values.  The `Show` class defined there is a new class, distinct from the `Show` in Prelude.  And one would have to implement instances of the new `Show` for all the basic types one might use to construct a `Term`.  Hardly worth calling it `Show` in the first place, I think that just creates confusion:
```
$ ghci -v0 Tup.hs 
λ> :info Term
type role Term nominal
type Term :: * -> *
data Term a where
  Con :: a -> Term a
  Tup :: Term b -> Term c -> Term (b, c)
        -- Defined at Tup.hs:24:1
instance [safe] Show d => Show (Term d) -- Defined at Tup.hs:29:10
type instance ShowCtx (Term d) = Show d         -- Defined at Tup.hs:28:15
λ> Tup.show (Con True)
"True"
λ> Tup.show (Tup (Con True) (Con False))
"(True,False)"
λ> Tup.show $ Con "foo"

<interactive>:4:1: error:
    • No instance for (Show [Char]) arising from a use of ‘show’
    • In the expression: show $ Con "foo"
      In an equation for ‘it’: it = show $ Con "foo"
```
So how would one in fact define an **actual** `Show` instance for `Term`, that could leverage existing `Show` instances for the underlying types out of which Terms are constructed?

vdukhovni:
The best solution I can come up with is to bypass the whole issue, by changing the definition of `Term`:
```haskell
{-# LANGUAGE GADTs #-}
module Tup (Term(..)) where

data Term d where
  Con :: b -> Term b
  Tup :: Term p -> Term q -> Term (Term p, Term q)

instance Show d => Show (Term d) where
    show (Con b) = show b
    show (Tup x y) = show (x,y)
```
This just works, provided one is willing to accept more deeply layered types for compound terms.

ulysses4ever:
@vdukhovni the overlapping solution is [discussed in the proposal](https://github.com/KoenP/ghc-proposals/blob/patch-1/proposals/0000-bidirectional-instances.md#use-overlapping-instances).

vdukhovni:
> @vdukhovni the overlapping solution is [discussed in the proposal](https://github.com/KoenP/ghc-proposals/blob/patch-1/proposals/0000-bidirectional-instances.md#use-overlapping-instances).

Good point.  While perhaps non-obvious, it is I think reasonably clean.  But really I think that the right/natural solution is the one with `Term (Term p, Term q)`.

Lev135:
I've implemented something like the solution discussed here as a class type class with associated type family and TH function to write boilerplate code for

Lev135:
I've implemented something like the solution discussed here as a class 
with associated type family and TH function to generate boilerplate code 
(not too much, but I think it's anyway may be useful):
https://hackage.haskell.org/package/bidirectional-instances-0.1.0.0#readme

Certanly GHC extension would be better for some resons:
- no need for TH in dependencies
- more intuitive syntax
- no need for language pragmas (no these are necessary for my solution to work:
`TemplateHaskell`, `FlexibleContexts`, `UndecidableInstances`, 
`MultiParamTypeClasses` and `TypeFamilies`).
- maybe perfomane (I'm not skilled enough to reason about it)

But it has one great advantage over this proposial: it is implemented and it's
implementation is quite simple. And comparing with ones suggested here, 
it's more general and (for me) looks more natural. 

Full code for the problem with GADT:
```hs
{-# LANGUAGE GADTs #-}
{-# LANGUAGE FlexibleContexts #-}
{-# LANGUAGE MultiParamTypeClasses #-}
{-# LANGUAGE TypeFamilies #-}

module Example where

import Data.Kind (Type)
import Control.Bidirectional ( BidirectionalRec, makeBidirectionalInstances) 

data Term :: Type -> Type where
  Con :: a -> Term a
  Tup :: Term b -> Term c -> Term (b, c)

makeBidirectionalInstances [d|
    instance (Show b, Show c)  => Show (b, c)
  |]

instance BidirectionalRec Show a => Show (Term a) where
  show (Con x) = show x
  show (Tup x y) = unwords ["(", show x, ",", show y, ")"]
```
TH actually expands in
```hs
instance Constr Show (b, c) => Bidirectional Show (b, c) where
  type Constr Show (b, c) = (Bidirectional Show b, Bidirectional Show c)
```
so it can be written manually without too much work, but I prefer the first 
varriant as more intuitive

I hope this may be useful to overview this problem and also may be used as a
solution for the current state.

PR #270, opened 2019-09-05
Support pun-free code

Hithroc:
This proposal introduces `-Wpuns` and `-Wpun-bindings` and
introduces namespace-qualified imports and module aliases.

These changes should help the users to write pun-free code to take advantage of
*Syntactic Unification Principle* described in #378.

[Rendered](https://github.com/hithroc/ghc-proposals/blob/master/proposals/0000-support-pun-free-code.md)

PR #285, opened 2019-10-17
`-XNoImplicitForAll`, `-XNoPatternSignatureBinds`

Ericson2314:
Provide a way to opt out of implicit ``forall`` binding of free variables in type signatures. Compliments the unified namespace proposal (#270) by making all bindings usable in type signatures.

[Rendered](https://github.com/Ericson2314/ghc-proposals/blob/no-implicit-forall/proposals/0000-no-implicit-binds.rst)


int-index:
A few minor remarks:

* The capitalization should be `NoImplicitForAll` for consistency with `ExplicitForAll`.

* The actual extension should be `ImplicitForAll`, enabled by default, so the user gets to disable it with `NoImplicitForAll`. If the extension is `NoImplicitForAll`, then disabling it would be `NoNoImplicitForAll`, which doesn't look right to me.

* Disabling `ImpilicitForAll` should probably automatically enable `ExplicitForAll`.

proposals/0000-no-implicit-forall.rst
```diff
@@ -0,0 +1,103 @@
...
+.. sectnum::
+.. contents::
+
+Provide a way to opt out of implicit ``forall`` binding of free variables in type signatures.
+
+Motivation
+----------
+
+There are two independent motivations, education and a unified namespace.
+Compliments the unified namespace proposal (`#270`_) by making all bindings usable in type signatures.
```
JakobBruenker: s/Compliments/Complements

proposals/0000-no-implicit-forall.rst
```diff
@@ -0,0 +1,103 @@
...
+
+Education
+~~~~~~~~~
+
+Some people think that implicit binding is bad for people learning Haskell.
+All other variables are explicitly bound, and the inconsistency means more to learn.
+Also, implicit syntax in general allows the beginner to not realize what they are doing.
+What are tedious tasks for the export may be helpful learning steps to them.
```
JakobBruenker: s/export/expert

Ericson2314:
>     * The capitalization should be `NoImplicitForAll` for consistency with `ExplicitForAll`.

Done

>     * The actual extension should be `ImplicitForAll`, enabled by default, so the user gets to disable it with `NoImplicitForAll`. If the extension is `NoImplicitForAll`, then disabling it would be `NoNoImplicitForAll`, which doesn't look right to me.

That as my intent. Maybe I should re-title it?
 
>     * Disabling `ImpilicitForAll` should probably automatically enable `ExplicitForAll`.

Mmm, I prefer extensions to be monotonic where possible. Also this made me think some teachers may which to prevent polymorphism altogether with `-XNoExplicitForAll` `-XNoImplicitForAll` `-Wmissing-signatures` and `-Wmissing-local-signatures`. I'll add that to the moivation.

monoidal:
With `-XNoImplicitForAll`, will `instance C a => D a` be allowed, or do we require `instance forall a. C a => D a`? I guess the latter.

Should `-XNoImplicitForAll` imply `-fprint-explicit-foralls`?

treeowl:
I think `NoImplicitForAll` should turn on `ExplicitForAll`, but users can override it.

neongreen:
This would be the first precedent where a `{-# LANGUAGE NoX #-}` pragma enables an extension. I'd rather avoid it.

People using a Dependent Haskell kitchen sink likely already have `{-# LANGUAGE ExplicitForAll #-}` transitively enabled (via `ScopedTypeVariables`, `RankNTypes`, etc), and people using "teacher's edition" of Haskell will either have a `default-extensions` list handed to them, or won't even need `ExplicitForAll` as per the discussion in the proposal:

> Further, the most beginnning students may be taught with both `-XNoImplicictForall` and `-XNoExplicitForall`

JakobBruenker:
> This would be the first precedent where a `{-# LANGUAGE NoX #-}` pragma enables an extension. I'd rather avoid it.

Isn't `-XNoMonomorphismRestriction`the same thing?

neongreen:
I meant "enables a different extension". AFAIK `NoMonormorphismRestriction` doesn't enable anything else (judging by https://gitlab.haskell.org/ghc/ghc/blob/d584e3f08cfee6e28b70bf53c573d86e44f326f8/compiler%2Fmain%2FDynFlags.hs#L4641).

int-index:
> I meant "enables a different extension"

Ah, so it's an argument against `NoImplicitForAll` implying `ExplicitForAll`. Good point.

Ericson2314:
Thanks @neongreen, you made my "Mmm, I prefer extensions to be monotonic where possible" a lot clearer.

Ericson2314:
@monoidal great point, I also forgot data declarations.

Ericson2314:
@int-index Yeah I feel better now. `class forall b (x :: b). Foo (x :: b)` is garbage because a class has parameters (binders) while a instance has binders. I'll remove the question.

@int-index so the overly-conservative thing was that in other proposals places we say a type signature in a pattern binds a variable that is unified (it is fresh "syntactically" but not "semantically"). I don't want to fight against that completely, but I rather there at least be a way to opt-out of it. I've written this proposal such that you'd have to opt *in* if you have `NoImplicitQuantififation`.

goldfirere:
I generally prefer fewer implications among extensions -- fewer implications means that I have less to remember. And an implication from a `-XNo...` seems like a bad precedent.

Ericson2314:
OK cleaned up the language so the classes case is handled like @int-index said. Also made the prohibition on binding variables in binding signatures clear.

Ericson2314:
OK, @nomeata, I think I am ready to submit to the committee.

A summary of the discussion: First people found typos (thanks!). Then, we discussed where else implicit binding can occur besides term binding signatures (`instance` and `data` declarations, pattern signatures, but *not* `class` declarations). I was initially confused re class declarations, and no agree with everyone else. Finally, we discussed whether `-XNoImplicitParams` should imply anything else. This was the only point of disagreement, some people including myself argue for monotonicity (the turning *off* of a feature should never turn *on* anything), whereas others argued for convenience. I stayed with monotonicity, but added the convenient implications as alternatives.

simonpj:
I'm not against this proposal.   But one point troubles me.  In a single example, never discussed or repeated we see
```
f (x :: k) = ... -- error: k is not bound
```
This is a *pattern signature*, which is a special and slightly odd feature of GHC's version of Haskell.  It is NOT considered shorthand for
```
f (x :: forall k. k) = ....
```
Rather, it brings k into scope, binding it to x's type (whatever that is).   If, with the new flag, this was outlawed, there is no convenient workaround.

I would advocate narrowing the scope, to *not* include pattern signatures.  (Granted, there may be a separate conversation to be had about those.)

goldfirere:
Somehow I missed the bit about pattern signatures (which is in the specification section). I agree that this proposal shouldn't touch those -- there's no implicit `forall` near pattern signatures.

Ericson2314:
I agree it's not an implicit `forall`, but (to me) it is an "implicit binding" which I'll call any sometime-reference / sometime-binding. I think I'll add a second on-by-default extension to cover it, as my goal with this proposal really is to make a firm distinction between bindings and uses, but if still the consensus is the other extension should become another proposal, I'll move it.

proposals/0000-no-implicit-binds.rst
```diff
@@ -0,0 +1,233 @@
...
+                ticket URL which will track the progress of the
+                implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/285>`_.
+.. sectnum::
+.. contents::
+
+Provide a way to strictly separate bindings and use of varariables, so the destinction never depends on what is in scope.
```
simonmar: *distinction

proposals/0000-no-implicit-binds.rst
```diff
@@ -0,0 +1,233 @@
...
+This means opting out of implicit ``forall`` binding of free variables in type signatures, and opting out of the binding of variables in pattern signatures.
+The two can be controlled separately.
+
+Motivation
+----------
+
+There are two independent motivations, education and a unified namespace.
+Complements the unified namespace proposal (`#270`_) by making all bindings usable in type signatures.
```
simonmar: "This proposal complements the unified namespace proposal..."

proposals/0000-no-implicit-binds.rst
```diff
@@ -0,0 +1,233 @@
...
+@Monoidal asks whether `-XNoImplicitForAll` should imply `-fprint-explicit-foralls`.
+I am also sympathetic, but again worried about non-monotonicity.
+That said, warnings are more freeform than extensions so I am less worried than with the above suggestion.
+
+Unresolved Questions
+--------------------
+
+No unresolved questions.
+
+seems something that ought to be prohibited because ``b`` is unbound.
```
simonmar: what does this refer to?

aspiwack:
I found it pretty hard to understand what is being proposed here.

A few suggestion to improve clarity:

- Expand the _Specification_ section to specify the meaning/behaviour variables which are not “explicitly bound” under each of your new extension.
- Give examples, in the introduction to your section 3, with and without each extension turned on. To illuminate what everything does.

The _Alternatives_ section doesn't mention it, so: what is the motivation for having two extensions rather than a single one?

isovector:
I don't understand the difference between these two flags --- it doesn't seem to have any examples in the proposal. In addition, the messiness between the parameters of classes and arguments of instances makes me nervous --- as a user, I don't want to care about which things GHC considers a binder and which it doesn't.

Ericson2314:
Sorry, my revision into making it two extensions, based on the comments, was a bit hurried. I'll separate which extensions apply to which examples, and include the original one extension proposal in the alternatives.

Variables which are not explicitly bound are unbound, and that's an error, but make it more explicit. I mentioned the `forall.` trick to simulate `-XNoExplicitForall`, but there isn't a corresponding trick for `-XNoPatternSignatureBinds` and I should mention that too.

@isovector If you don't feel like caring about what's is and isn't a binder, then don't turn these extensions off! I don't expect their use to be widespread in most code initially (before DH), mainly educational code that never reached hackage. I think that's fine. If optional sugar allows more people to improve their understanding, even if they eventually turn it back on, that's victory to me. Compare https://terrytao.wordpress.com/career-advice/theres-more-to-mathematics-than-rigour-and-proofs/ (though I do wish professional math was more formal!).

int-index:
>  there isn't a corresponding trick for `-XNoPatternSignatureBinds` and I should mention that too.

There is, it's #155! Sure, it is not implemented yet, but that shouldn't affect the proposed design.

simonpj:
Some thoughts

* I like separating `-XExplicitForAll` from `-XPattternSignatureBinds`.  Though, as you say, the proposal does not adequately describe the latter.

* You remark that `data T (a :: k) = ...` becomes illegal.  But please add a bit to explain how you can make it legal.   We don't want to force you to write `data T k (a::k) = ...` because then we have to add an argument to every use of T.  We want something like `data T @k (a::k) = ...`.  Do we have a proposal for that?   

  And can you write `data T @{k} (a::k) = ...` to get an Inferred rather than Specified argument (see [accepted proposal on Explicit specificity](https://github.com/goldfirere/ghc-proposals/blob/explicit-specificity/proposals/0000-explicit-specificity.rst)).

* int-index points to #155 (Binding type variables in lambdas), but actually that proposal does not let you bind existential type varables.   I think that the accepted [poposal on type application in patterns](https://github.com/ghc-proposals/ghc-proposals/blob/type-applications-in-patterns/proposals/type-applications-in-patterns.rst) is the one you want.  Consider
  ```
  data T where MkT :: a -> (a->a) -> T
  f x = case x of MkT (x::b) f -> f x :: b
  ```
  Here `b` is existentially bound; bringing it into scope is the number-one reason that pattern signatures were introduced in the first place, many moons ago.  If the implicit binding becomes illegal, we need another way, and the type-application-in-patterns does the job, thus:
  ```
   f x = case x of { MkT @b x f -> f x :: b }
  ```
  This is a connection worth making.

* In general, it would be useful if the proposal could explicitly state how every program rendered illegal by the proposed change could be made legal again.   (Of course you can always switch off the extension, but the proposal is implicitly arguing for a particular style of explicit binding, so we had better make sure that we can always do that.)

Ericson2314:
The rest all sounds good and I'll do that, but let me clarify myself on the "forall `forall.` trick". T

he `.` is crucial; I mean with scoped type variables you can put such an "empty" `forall.` at the beginning of the expression to prevent any free type variables from being automatically bound. This allows one to mimic `-XNoImplicitForAll` today. 

Pattern signature bindings are *not* syntax-driven sugar for universals or existentials, or moreover any proposed `@var` binding. i.e. none of
```haskell
f (_ :: t) = ...
f @t (_ :: t) = ...
f (MkFoo (_ :: t)) = ...
f (MkFoo @t (_ :: t)) = ...
```
are necessarily equivalent. The inferred definition of `t` can always be rewritten into some combination of those things (and both types of `@` binding are necessary, in general), but the de-sugar is not syntax directed. (Interestingly, pre-#128 we might have been able to say that only a single `@` was needed, though perhaps in a larger scope; still the desugar wouldn't have been syntax directed.)

I'm thus prepared to admit while explicit `forall` cleanly replaces implicit `forall`, there is no negligible human effort replacement for pattern signature binds, and that would await e.g. a a future proposal for e.g. `type a = _` (where we bind a variable explicitly, but say it is defined by unification with its use-sites).

I'll add that admission to the drawbacks.

simonmar:
Putting the proposal into "needs revision" state after recent comments. As I understand it these are mostly requests for clarifications, but we'll take another look once they're actioned.

Ericson2314:
Thanks for the reminder. The examples are greatly expanded now, I believe covering all outstanding concerns.

nomeata:
What’s the status here, @Ericson2314 and @simonmar?

Ericson2314:
I'm hoping my last revisions succeeded in clarifying things!

simonmar:
Looks good to me. @aspiwack and @simonpj you requested changes, can you check whether the updates address your concerns?

simonpj:
I'm pretty content, thank you.  Just some nits

* Section 3.1 item 2 says that the old thing can be rewritten
  ```
  f @t (x :: t1) = ....
  ```
  I think you mean `(x :: t)`.   But also this `f @t` stuff isn't in Haksell yet.  There is a proposal; please point to it explicilty.

* In 3.2 you refer to a "hypothetical not-yet-proposed" syntax for data type `data T @k (a:k) = ....`.  Is it really not yet proposed?  Please write a proposal!  We obviously want it!

Thanks!

Ericson2314:
OK the first part is done. I'll try to write that second proposal soon too.

Ericson2314:
I realized I missed some typo comments; fixed those too.

proposals/0000-no-implicit-binds.rst
```diff
@@ -0,0 +1,446 @@
...
+``-XNoImplicitForAll`` alone would *not* cause ``t`` in the signature to be bound from in the above; it would be unbound causing and error.
+But, as a newly freed variable, it is now ready to be captured by whatever is proposed in `#270`_.
+I think this is a good separation of concerns.
+
+Proposed Change Specification
+-----------------------------
+
+Create ``-XImplicitForAll`` to allow automatically capturing free variables in an outer ``forall`` as is always done today.
+It is on by default for backwards compatibility.
+When using ``-XNoImplicitForAll``, all variables in regular signatures, instances, and data declerations must be explicitly bound.
```
aspiwack:
```suggestion
When using ``-XNoImplicitForAll``, all variables in regular signatures, instances, and data declarations must be explicitly bound.
```

proposals/0000-no-implicit-binds.rst
```diff
@@ -0,0 +1,446 @@
...
+     data Some where
+       MkSome :: forall t. t -> Some
+
+     f (MkSome (x :: t)) = ... -- error: `t` is not bound
+
+   This could be rewritten as::
+
+     -- {-# LANGUAGE NoImplicitForAll #-} -- Does not matter whether enabled or disabled
+     {-# LANGUAGE NoPatternSignatureBinds #-}
+     {-# LANGUAGE ExistentialQuantification #-}
```
aspiwack:
```suggestion
     {-# LANGUAGE GADTs #-}
```
(just because you are using that syntax, rather than the existential syntax)

proposals/0000-no-implicit-binds.rst
```diff
@@ -0,0 +1,446 @@
...
+     {-# LANGUAGE TypeApplications #-}
```
aspiwack: Don't you need the type-application-in-pattern extension from a recent proposal rather than `-XTypeApplications`?

proposals/0000-no-implicit-binds.rst
```diff
@@ -0,0 +1,446 @@
...
+
+   Note that since there is no ``class F :: ...`` syntax analogous to ``data F :: ...``,
+   so ``-XStandaloneKindSignatures`` are the only way to write explicitly kind-polymorphic classes.
+
+Note that the variables to the left of the ``::`` are are deemed explicit bindings analogous to ``f (y :: x) (z :: z) = ...`` and permitted.
+However ``x`` to the right of the ``::`` is a use, not otherwise bound, and thus implicit binding today.
+It is not permitted as-is, and must be explicitly bound or discarded as done in the working alternatives.
+
+Empty `forall` "desugar"
+~~~~~~~~~~~~~~~~~~~~~~~~
```
aspiwack: You should probably move this subsection to the `Cost` and drawback as it is a justification that this proposal oughtn't be costly to implement.

proposals/0000-no-implicit-binds.rst
Ericson2314: https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0126-type-applications-in-patterns.rst#proposed-change-specification there's isn't one, actually.

proposals/0000-no-implicit-binds.rst
Ericson2314: Moved.

proposals/0000-no-implicit-binds.rst
Ericson2314: I linked it for good measure

Ericson2314:
I think we are all good now?

nomeata:
> I think we are all good now?

I assume this question is to @aspiwack and @simonpj, pinging explicitly.

aspiwack:
On my end, this comment still stands:

> I'd still like to see, in the alternative section, a discussion of one extension vs two extensions for these effects. Other than that the proposal is solidly written.

Ericson2314:
Oh sorry I forgot about that.

Ericson2314:
OK discussed the alternative of a single extension.

aspiwack:
I'm good then.

nomeata:
@simonpj are your comments addressed as well?

simonpj:
Yes, good to go.

nomeata:
@simonmar do you want to present the updated proposal to the committee for a final review?

Ericson2314:
Any word?

Ericson2314:
From https://mail.haskell.org/pipermail/ghc-steering-committee/2020-January/001410.html

> I'm still not convinced about the fact that there are two extensions for
this. Which seems more motivated by technical details of the
implementation, rather than by semantics. But it is not a hill that I'm
willing to die on… so if I'm alone in this, I'm not going to argue further.

As I put in the proposal, I only lean slightly towards 2 extensions :). I can't speak for the other committee members though.

So currently today, when the variable is free, both pattern sigs and implicit foralls bind an identifier, and when the identifier it is not free, it is captured. So in terms of name resolution they do work the same. 

However in terms of the semantics, the type variable is a fresh quantification / rigid var in the latter, but free to unify with things / non-rigid with the former. This is a true semantic difference, not just implementation difference.

And yet... with the both `-XNo...` flags turned on, type variables in regular type sigs (that aren't explicit binds) and regular sigs are both captured (or a scoping violation) (same name resolution) and both take on the type value of the binding that captures them (same semantics).

So that means if we had `-XNo` flag, while it changes two distinct semantics, the "destination" is the same semantics. But that is the destination in the `-XNo` direction. If you ignore what's on by default and just worry about the "positive" (`-X` not `-XNo`) direction, it's one flag starting with a uniform simple semantics and enabling two distinct semantics.

All this is to say the situation is somewhat muddled and it depends whether you care more about the `-XNo` -> `-X` transition or default -> manual override transition.

Ericson2314:
It's been a while, but I recall I purposefully kept the list of items out of the proposed change and in examples because I assume I would miss one :).

The normative part assumes "pattern signatures" was well-defined enough to include it and perhaps that was a mistake.

The `(a :: k`) in `data T (a :: k)` certainly *looks* to me like other RHS / binding `(a :: k)` so I lean towards treating it the same way following the examples. It would be good to get our our terminology consistent (glossary in the users guide?) regardless, though.

Ericson2314:
in particular,
```haskell
{-# LANGUAGE WhateverIsNeededForBraveUnifiedNamespace #-}

k = Type

data T (a :: k) -- captured now
```
nicely illustrates all the issues following from the violation principle (in your proposal) are at play, so I would like to be able to fix that.

tek:
probably sensible to link the implementation MR: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/8629

Ericson2314:
It helped to see the place where this came up: I left https://gitlab.haskell.org/ghc/ghc/-/merge_requests/8629#note_444076 and so "cross posting" that here.

IMO the confusion stems from the fact that https://github.com/ghc-proposals/ghc-proposals/pull/425 is the one type variable proposal *not* to be subsumed into https://github.com/ghc-proposals/ghc-proposals/pull/448. I suggest we subsume it too; we need a unified spec that gets rid of all the silly semantic drift between things which ought to work the same way.

barci2:
Since this has been moved to https://github.com/ghc-proposals/ghc-proposals/pull/448, shouldn't it receive an "Implemented" label just like https://github.com/ghc-proposals/ghc-proposals/pull/126?

PR #287, opened 2019-10-17
Simplify subsumption

simonpj:
This proposal simplifies GHC's subsumption judgement

[Rendered](https://github.com/ghc-proposals/ghc-proposals/blob/wip/spj-deep-skol/proposals/0000-simplify-subsumption.md)

### DIFF
--- a/proposals/0000-simplify-subsumption.md
+++ b/proposals/0000-simplify-subsumption.md
@@ -6,9 +6,7 @@ ticket-url: ""
 implemented: ""
 ---
 
-This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/0>).
-**After creating the pull request, edit this file again, update the number in
-the link, and delete this bold sentence.**
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/287).
 
 # Simplify subsumption
 
--- a/proposals/0000-simplify-subsumption.md
+++ b/proposals/0000-simplify-subsumption.md
@@ -24,7 +24,7 @@ However, *all four change the semantics of Haskell*, by performing eta-expansion
 
 * **The cost is large**.  There is extra complexity in the compiler.  There is extra complexity in the formal description of the type system, and in any paper we write about it.  But, worst of all, the features change the semantics of the program.
 
-* A recent new cost is that the [Quick Look Impredicativity proposal]( https://github.com/ghc-proposals/ghc-proposals/pull/274) gains extra power by not having contravariance.
+* A recent new cost is that the [Quick Look Impredicativity proposal]( https://github.com/ghc-proposals/ghc-proposals/pull/274) gains extra power by not having contravariance.   More details in the [Quick Look paper](https://www.microsoft.com/en-us/research/publication/a-quick-look-at-impredicativity/)
 
 Here is a thought experiment.  Suppose GHC lacked all four features, and someone proposed adding them.  That proposal would never leave the launchpad. A minor change in programming convenience, in exchange for changing language semantics?  No.  If that’s true, then the only issue would be back-compat issues: how many libraries would be affected, and how painful they would be to fix.  We need to get data on that.
 
@@ -138,6 +138,7 @@ See Motivation above.
 
 ## Effect and Interactions
 
+* See Section 7 of the [Quick Look paper](https://www.microsoft.com/en-us/research/publication/a-quick-look-at-impredicativity/) for a detailed analysis of the practical impact of these changes.
 * Everything (specification, implementation) becomes a bit simpler
 * Quick Look Impredicativity gains more power
 
@@ -197,7 +198,8 @@ The solution is simple, and improves the code: just give `getBuildDoctests` a ty
 
 ## Alternatives
 
-Status quo. But the the status quo is extremely unsatisfactory.
+1.  Status quo. But the the status quo is extremely unsatisfactory.
+2.  Change GHC's internmediate language to base it on [System F-eta](https://www.sciencedirect.com/science/article/pii/0890540188900090) (Mitchell, 1988).  In this language, everything is done modulo eta expansion/contraction.  Apart from the huge engineering consequences, it's not clear that F-eta as an intermediate language is compatible with Haskell, which distinguishes bottom from (\x.bottom).
 
 
 ## Unresolved Questions
--- a/proposals/0000-simplify-subsumption.md
+++ b/proposals/0000-simplify-subsumption.md
@@ -18,7 +18,7 @@ This proposal argues that GHC’s type system is a bit too liberal; in particula
 
 All four features are implemented in GHC.
 
-However, **all four change the semantics of Haskell*, by performing eta-expansion, changing bottom into non-bottom.  That’s a pretty serious change, which needs pretty serious motivation.  But actually all we get in exchange is a minor improvement in programming convenience.  So I argue that:
+However, *all four change the semantics of Haskell*, by performing eta-expansion, changing bottom into non-bottom.  That’s a pretty serious change, which needs pretty serious motivation.  But actually all we get in exchange is a minor improvement in programming convenience.  So I argue that:
 
 * **The benefit, in terms of programming convenience, is small**.  In particular, while using these features accepts some more programs, any such program is easily accepted without these features, after a small edit.
 
--- a/proposals/0000-simplify-subsumption.md
+++ b/proposals/0000-simplify-subsumption.md
@@ -143,11 +143,15 @@ See Motivation above.
 
 ## Costs and Drawbacks
 
-The main cost is that some existing programs will require some manual eta-expansion.
-
+The main user-facing cost is that some existing programs will require some manual eta-expansion.
 Jurriaan is gathering data on how many libraries are affected.
 
 Ideally we'd like a deprecation saying "you are using this feature, and it's going to disappear".  But this would have false positives: just because deep instantiation does eta-expansion does not imply that using only shallow instantiation would make the program un-typable.
+
+There are some implementation consequences:
+
+* `TcUnify.matchExpectedFunTys` would need to be extended to deal with the possiblity of a `forall`.  Very strightforward.
+* The ambiguity check would need a bit more code than at present.  Currently, we just check whether `ty <= ty` using the existing subsumption check: if this check fails, the type is ambiguous. With but with a simpler subsumption check `Int -> forall a. String` would be a sub-type of itself, even though it's plainly ambiguous. So we'd have to write a proper ambiguity checker.  Not hard!
 
 ### Examples of back-compat issues
 
--- a/proposals/0000-simplify-subsumption.md
+++ b/proposals/0000-simplify-subsumption.md
@@ -22,9 +22,9 @@ All four features are implemented in GHC.
 
 However, **all four change the semantics of Haskell*, by performing eta-expansion, changing bottom into non-bottom.  That’s a pretty serious change, which needs pretty serious motivation.  But actually all we get in exchange is a minor improvement in programming convenience.  So I argue that:
 
-* The benefit, in terms of programming convenience, is small.  In particular, while using these features accepts some more programs, any such program is easily accepted without these features, after a small edit.
+* **The benefit, in terms of programming convenience, is small**.  In particular, while using these features accepts some more programs, any such program is easily accepted without these features, after a small edit.
 
-* The cost is large.  There is extra complexity in the compiler.  There is extra complexity in the formal description of the type system, and in any paper we write about it.  But, worst of all, the features change the semantics of the program.
+* **The cost is large**.  There is extra complexity in the compiler.  There is extra complexity in the formal description of the type system, and in any paper we write about it.  But, worst of all, the features change the semantics of the program.
 
 * A recent new cost is that the [Quick Look Impredicativity proposal]( https://github.com/ghc-proposals/ghc-proposals/pull/274) gains extra power by not having contravariance.
 
@@ -36,16 +36,18 @@ The baseline for this proposal is [Practical Type Inference for Arbitrary Rank T
 * **Covariance** of the function arrow (Fig 7, rule FUN)
 * **Contravariance** of the function arrow (Fig 7, rule FUN)
 * **Deep skolemisation** (Fig 7, rule DEEP-SKOL)
-* **Deep instantiation**.  Actually, the section (4.7.3) says that we don’t need deep instantiation, but GHC actually does deep instantiation too, for reasons sketched below.
+* **Deep instantiation**.  Section (4.7.3) says that we don’t need deep instantiation, but GHC actually does that too, for reasons sketched below.
+
+The proposal argues to remove all four.  The following subsections descuss each in turn.
 
 ### Deep skolemisation
 
-Suppose (paper, 4.6.1)?
+Consider these types (paper, 4.6.1):
 ```
 f :: ∀ab.a → b → b)
 g :: (∀p.p → (∀q.q → q)) → Int
 ```
-Is `(g f)` well typed?  Notice that `g` requires an argument with a forall to the right of an arrow. With deep skolemisation, the answer is “yes”; but without the program would be rejected.  But remember that GHC elaborates to System F.  Here is its elaboration of `(g f)`:
+Is `(g f)` well typed?  Notice that `g` requires an argument with a forall to the right of an arrow. With deep skolemisation, the answer is “yes”.  But remember that GHC elaborates to System F.  Here is its elaboration of `(g f)`:
 ```
 g (/\p. \(x:p). /\q.\(y:q). f @p @q x y)
 ```
@@ -56,7 +58,7 @@ BUT suppose `f` and `g` are defined like this:
 f = bottom
 g f = f `seq` 0
 ```
-You would expect `g f` to diverge, since it seq’s on bottom.  But it won’t  it’ll return 0.  Yikes.
+You would expect `g f` to diverge, since it seq’s on bottom.  But it won’t! it’ll return 0.  Yikes.
 
 Without deep skolemisation, `(g f)` is rejected.  But the programmer can easily repair it by manual eta-expansion, to
 ```
@@ -66,11 +68,11 @@ and now, of course, it is not surprising that the expression evaluates to 0.
 
 ### Deep instantiation
 
-Suppose `f :: Int -> forall a. a -> a`.   Notice the forall to the right of the arrow.  Now consider this definition, which lacks a type signature:
+Suppose `f :: Int -> forall a. a -> a`.   Again, notice the forall to the right of the arrow.  Now consider this definition, which lacks a type signature:
 ```
 g x = f
 ```
-What type would you expect to infer for `g`?   The obvious answer is
+What type would you expect to infer for `g`?   The obvious answer (and the one we'd get without deep instantiation) is
 ```
 g :: forall b. b -> Int -> forall a. a -> a
 ```
@@ -78,7 +80,7 @@ But GHC actually deeply instantiates `f` (for no very well-explained reason), so
 ```
 g :: forall b a. b -> Int -> a -> a
 ```
-Perhaps that type is a tiny bit more explicable to the programmer.  But again, to produce that type, GHC must elaborate `g` to
+with the buried foralls pulled to the top.  Perhaps that type is a tiny bit more explicable to the programmer.  But again, to produce that type, GHC must elaborate `g` to
 ```
 g = /\ b a. \(x:b). \(i:Int). f i @a
 ```
@@ -101,11 +103,13 @@ Well, according to rule FUN of Figure 7, using contravariance of `(->)`, that is
 ```
 forall a. a -> a    <=      Int -> Int
 ```
-and that is certainly true.  But again, GHC needs to eta-expand during elaboration.  We get this elaboration of `(g f)`:
+and that is certainly true.  But again, to witness that proof GHC needs to eta-expand during elaboration.  We get this elaboration of `(g f)`:
 ```
 g (\(h : forall a. a->a).  f (h @Int))
 ```
-Again we have changed the semantics.  Yikes.  Again, lacking covariance the program would be rejected, but is easily fixed by manual eta-expansion, thus `g (\h -> f h)`
+Again we have changed the semantics.  Yikes.
+
+Again, lacking covariance the program would be rejected, but is easily fixed by manual eta-expansion, thus `g (\h -> f h)`
 
 ### Covariance
 
--- a/proposals/0000-simplify-subsumption.md
+++ b/proposals/0000-simplify-subsumption.md
@@ -42,7 +42,7 @@ The proposal argues to remove all four.  The following subsections descuss each 
 
 Consider these types (paper, 4.6.1):
 ```
-f :: ∀ab.a → b → b)
+f :: ∀ab.a → b → b
 g :: (∀p.p → (∀q.q → q)) → Int
 ```
 Is `(g f)` well typed?  Notice that `g` requires an argument with a forall to the right of an arrow. With deep skolemisation, the answer is “yes”.  But remember that GHC elaborates to System F.  Here is its elaboration of `(g f)`:
@@ -147,15 +147,15 @@ The main cost is that some existing programs will require some manual eta-expans
 
 Jurriaan is gathering data on how many libraries are affected.
 
-Ideally we'd like a deprecation saying "you are using this feature, and it's going to disappear".  But this would have falies positives: just because deep instantiation does eta-expansion does not imply that using only shallow instantaition would make the program un-typable.
+Ideally we'd like a deprecation saying "you are using this feature, and it's going to disappear".  But this would have false positives: just because deep instantiation does eta-expansion does not imply that using only shallow instantiation would make the program un-typable.
 
 ## Alternatives
 
-Status quo. But the the stauts quo is extremely unsatisfactory.
+Status quo. But the the status quo is extremely unsatisfactory.
 
 
 ## Unresolved Questions
 
 ## Implementation Plan
 
-Implemenation is relatively easy.  I can do it, or Richard, or Alejandro.
+Implementation is relatively easy.  I can do it, or Richard, or Alejandro.
--- a/proposals/0000-simplify-subsumption.md
+++ b/proposals/0000-simplify-subsumption.md
@@ -149,6 +149,48 @@ Jurriaan is gathering data on how many libraries are affected.
 
 Ideally we'd like a deprecation saying "you are using this feature, and it's going to disappear".  But this would have false positives: just because deep instantiation does eta-expansion does not imply that using only shallow instantiation would make the program un-typable.
 
+### Examples of back-compat issues
+
+We found an example of a back-compat problem in `cabal-doctest`.  In `Cabal:Distribution/Compat/Prelude` we have:
+```
+type CabalIO a = HasCallStack => IO a
+
+```
+(actually the definition re-uses `IO` as the name, but that's just confusing, so I've renamed it `CabalIO` here.) Then in `Cabal:Ditribution.Simple.LocalBuildInfo` we have
+```
+withLibLBI :: PackageDescription -> LocalBuildInfo
+           -> (Library -> ComponentLocalBuildInfo -> CabalIO ()) -> CabalIO ()
+```
+Finally, in `cabal-doctest`, a function has a local definition, with no type signature,
+looking like
+```
+   let getBuildDoctests withCompLBI mbCompName compExposedModules
+                        compMainIs compBuildInfo = ...
+   in
+   ...(getBuildDoctests withLibLBI ...)...
+```
+Now, lacking a type signature on `getBuildDoctests`, GHC infers the type of the function
+to have plain arrows in its type, something like
+```
+   getBuildDoctests :: (PackageDescription -> LocalBuildInfo -> blah -> IO ())
+                    -> ...blah...
+```
+but in the call the actual argument `withLibLBI` has type
+```
+withLibLBI :: PackageDescription -> LocalBuildInfo -> blah -> HasCallStack => IO ()
+```
+And the function and its argumetnt do not agree about the placement of the `HasCallStack` constraint.
+With deep skolemisation, GHC would eta-expand the call to
+```
+   getBuildDoctests (\ a b c. withLibLBI a b c)  ...
+```
+but, as discussed, that is unsound in general.
+
+Moreover, there is a *nested* use of `CabalIO` in `withLibLBI`'s third argument, so GHC has to use contravariance and more eta expansion to make that line up.
+
+The solution is simple, and improves the code: just give `getBuildDoctests` a type signature!
+
+
 ## Alternatives
 
 Status quo. But the the status quo is extremely unsatisfactory.
--- a/proposals/0000-simplify-subsumption.md
+++ b/proposals/0000-simplify-subsumption.md
@@ -26,7 +26,7 @@ However, *all four change the semantics of Haskell*, by performing eta-expansion
 
 * A recent new cost is that the [Quick Look Impredicativity proposal]( https://github.com/ghc-proposals/ghc-proposals/pull/274) gains extra power by not having contravariance.   More details in the [Quick Look paper](https://www.microsoft.com/en-us/research/publication/a-quick-look-at-impredicativity/)
 
-Here is a thought experiment.  Suppose GHC lacked all four features, and someone proposed adding them.  That proposal would never leave the launchpad. A minor change in programming convenience, in exchange for changing language semantics?  No.  If that’s true, then the only issue would be back-compat issues: how many libraries would be affected, and how painful they would be to fix.  We need to get data on that.
+Here is a thought experiment.  Suppose GHC lacked all four features, and someone proposed adding them.  That proposal would never leave the launchpad. A minor change in programming convenience, in exchange for changing language semantics?  No.  If that’s true, then the only issue would be back-compat issues: how many libraries would be affected, and how painful they would be to fix.  We have collected data on this, presented in 
 
 ## Motivation
 
@@ -128,9 +128,9 @@ by removing
 * Deep skolemisation
 * Deep instantiation
 
-This could be under control of a flag, but I propose that we
-take a clear path to removing these features altogether, i.e.
-ultimately, without a flag that enables them.
+Thinking about a transition, it is very difficult to accept all current programs, while providing a warning for programs that will need to be changed when the propsal is adopted. Doing so would amount to compiling every program twice, which does not seem acceptable.
+
+It would be possible to offer a flag that restored the old behaviour, but that still means changing the .cabal file, or adding a LANGUAGE pragma. It seems more straightforward simply to change the source code to work with the new restrictions.  These changes turn out to be extremely minor, and fully backward compatible.
 
 ## Examples
 
@@ -142,12 +142,14 @@ See Motivation above.
 * Everything (specification, implementation) becomes a bit simpler
 * Quick Look Impredicativity gains more power
 
+Key conclusions of the practical impact (details in the paper) are:
+* Where programs require changes under this proposal, those changes are simple, local, and arguably desirable anyway.
+* The changes are backward-compatible: if you change a package to accommodate this proposal, it'll still compile with earlier GHC's too.
+
+
 ## Costs and Drawbacks
 
 The main user-facing cost is that some existing programs will require some manual eta-expansion.
-Jurriaan is gathering data on how many libraries are affected.
-
-Ideally we'd like a deprecation saying "you are using this feature, and it's going to disappear".  But this would have false positives: just because deep instantiation does eta-expansion does not imply that using only shallow instantiation would make the program un-typable.
 
 There are some implementation consequences:
 
--- a/proposals/0000-simplify-subsumption.md
+++ b/proposals/0000-simplify-subsumption.md
@@ -1,0 +1,159 @@
+---
+author: Simon Peyton Jones
+date-accepted: ""
+proposal-number: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/0>).
+**After creating the pull request, edit this file again, update the number in
+the link, and delete this bold sentence.**
+
+# Simplify subsumption
+
+This proposal argues that GHC’s type system is a bit too liberal; in particular, the subsumption judgement is too liberal.  It adds a little bit of convenience for the programmer, but at quite a high cost in terms of changing semantics and type-system complexity.   So I propose that we should simplify the type system, by removing
+* Covariance of function types
+* Contravariance of function types
+* Deep skolemisation
+* Deep instantiation
+
+All four features are implemented in GHC.
+
+However, **all four change the semantics of Haskell*, by performing eta-expansion, changing bottom into non-bottom.  That’s a pretty serious change, which needs pretty serious motivation.  But actually all we get in exchange is a minor improvement in programming convenience.  So I argue that:
+
+* The benefit, in terms of programming convenience, is small.  In particular, while using these features accepts some more programs, any such program is easily accepted without these features, after a small edit.
+
+* The cost is large.  There is extra complexity in the compiler.  There is extra complexity in the formal description of the type system, and in any paper we write about it.  But, worst of all, the features change the semantics of the program.
+
+* A recent new cost is that the [Quick Look Impredicativity proposal]( https://github.com/ghc-proposals/ghc-proposals/pull/274) gains extra power by not having contravariance.
+
+Here is a thought experiment.  Suppose GHC lacked all four features, and someone proposed adding them.  That proposal would never leave the launchpad. A minor change in programming convenience, in exchange for changing language semantics?  No.  If that’s true, then the only issue would be back-compat issues: how many libraries would be affected, and how painful they would be to fix.  We need to get data on that.
+
+## Motivation
+
+The baseline for this proposal is [Practical Type Inference for Arbitrary Rank Types]( https://www.microsoft.com/en-us/research/publication/practical-type-inference-for-arbitrary-rank-types/), and specifically Section 4.6 which concerns the subsumption judgement.  That section discusses
+* **Covariance** of the function arrow (Fig 7, rule FUN)
+* **Contravariance** of the function arrow (Fig 7, rule FUN)
+* **Deep skolemisation** (Fig 7, rule DEEP-SKOL)
+* **Deep instantiation**.  Actually, the section (4.7.3) says that we don’t need deep instantiation, but GHC actually does deep instantiation too, for reasons sketched below.
+
+### Deep skolemisation
+
+Suppose (paper, 4.6.1)?
+```
+f :: ∀ab.a → b → b)
+g :: (∀p.p → (∀q.q → q)) → Int
+```
+Is `(g f)` well typed?  Notice that `g` requires an argument with a forall to the right of an arrow. With deep skolemisation, the answer is “yes”; but without the program would be rejected.  But remember that GHC elaborates to System F.  Here is its elaboration of `(g f)`:
+```
+g (/\p. \(x:p). /\q.\(y:q). f @p @q x y)
+```
+The lambdas do the impedence matching to turn `f` into an argument of exactly the type that `g` expects.
+
+BUT suppose `f` and `g` are defined like this:
+```
+f = bottom
+g f = f `seq` 0
+```
+You would expect `g f` to diverge, since it seq’s on bottom.  But it won’t  it’ll return 0.  Yikes.
+
+Without deep skolemisation, `(g f)` is rejected.  But the programmer can easily repair it by manual eta-expansion, to
+```
+g (\x y. f x y)
+```
+and now, of course, it is not surprising that the expression evaluates to 0.
+
+### Deep instantiation
+
+Suppose `f :: Int -> forall a. a -> a`.   Notice the forall to the right of the arrow.  Now consider this definition, which lacks a type signature:
+```
+g x = f
+```
+What type would you expect to infer for `g`?   The obvious answer is
+```
+g :: forall b. b -> Int -> forall a. a -> a
+```
+But GHC actually deeply instantiates `f` (for no very well-explained reason), so we get
+```
+g :: forall b a. b -> Int -> a -> a
+```
+Perhaps that type is a tiny bit more explicable to the programmer.  But again, to produce that type, GHC must elaborate `g` to
+```
+g = /\ b a. \(x:b). \(i:Int). f i @a
+```
+GHC has eta-expaned `f`, which changes the semantics.  Yikes.
+
+### Contravariance
+
+Suppose you have
+```
+g :: ((forall a. a -> a) -> R) -> S
+f :: (Int -> Int) -> R
+```
+Now, is `(g f)` well typed?   That depends on whether
+```
+(Int -> Int) -> R   <=    (forall a. a -> a) -> R
+```
+where `<=` is pronounced “is more polymorphic than”  see the paper sections 4.4. and 4.6.
+
+Well, according to rule FUN of Figure 7, using contravariance of `(->)`, that is true if
+```
+forall a. a -> a    <=      Int -> Int
+```
+and that is certainly true.  But again, GHC needs to eta-expand during elaboration.  We get this elaboration of `(g f)`:
+```
+g (\(h : forall a. a->a).  f (h @Int))
+```
+Again we have changed the semantics.  Yikes.  Again, lacking covariance the program would be rejected, but is easily fixed by manual eta-expansion, thus `g (\h -> f h)`
+
+### Covariance
+
+Fig 7 in the paper also supports covariance of the function arrow, but exactly the same eta-expansion issues arise.
+
+#
+
+
+## Proposed Change Specification
+
+There are no syntactic changes.
+
+The changes to the type system is to simplify the subsumption judgement
+by removing
+
+* Covariance of function types
+* Contravariance of function types
+* Deep skolemisation
+* Deep instantiation
+
+This could be under control of a flag, but I propose that we
+take a clear path to removing these features altogether, i.e.
+ultimately, without a flag that enables them.
+
+## Examples
+
+See Motivation above.
+
+## Effect and Interactions
+
+* Everything (specification, implementation) becomes a bit simpler
+* Quick Look Impredicativity gains more power
+
+## Costs and Drawbacks
+
+The main cost is that some existing programs will require some manual eta-expansion.
+
+Jurriaan is gathering data on how many libraries are affected.
+
+Ideally we'd like a deprecation saying "you are using this feature, and it's going to disappear".  But this would have falies positives: just because deep instantiation does eta-expansion does not imply that using only shallow instantaition would make the program un-typable.
+
+## Alternatives
+
+Status quo. But the the stauts quo is extremely unsatisfactory.
+
+
+## Unresolved Questions
+
+## Implementation Plan
+
+Implemenation is relatively easy.  I can do it, or Richard, or Alejandro.


proposals/0000-simplify-subsumption.md
```diff
@@ -0,0 +1,161 @@
...
+* Everything (specification, implementation) becomes a bit simpler
+* Quick Look Impredicativity gains more power
+
+## Costs and Drawbacks
+
+The main cost is that some existing programs will require some manual eta-expansion.
+
+Jurriaan is gathering data on how many libraries are affected.
+
+Ideally we'd like a deprecation saying "you are using this feature, and it's going to disappear".  But this would have falies positives: just because deep instantiation does eta-expansion does not imply that using only shallow instantaition would make the program un-typable.
```
atennapel: typo: `falies` and `instantaition`

proposals/0000-simplify-subsumption.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Alternatives
+
+Status quo. But the the stauts quo is extremely unsatisfactory.
```
atennapel: typo: `stauts`

proposals/0000-simplify-subsumption.md
```diff
@@ -0,0 +1,161 @@
...
+
+
+## Unresolved Questions
+
+## Implementation Plan
+
+Implemenation is relatively easy.  I can do it, or Richard, or Alejandro.
```
atennapel: typo: `Implemenation`

simonpj:
Thanks @JakobBruenker.  I have pushed typo fixes.

simonpj:
Everyone: I have erred on the side of brevity.  Please do collaborate as co-authors, and offer more examples and text for me to add!

LeanderK:
I am just a user trying who noticed the Quick Look Impredicativity proposal. It mentions problems with the `ContravariantFunctions`-extension, because "`CabalIO a` is defined as `HasCallStack => IO a`". I sometimes use this pattern and define a type-aliases for a type with a constraint when I have a constraint that i use everywhere. What are the consequences for this pattern? I am still unsure where exactly it will break?

simonpj:
>  It mentions problems with the ContravariantFunctions-extension, because "CabalIO a is defined as HasCallStack => IO a". 

I added a section to the proposal discussing this example.

jvanbruegge:
I am sorry but I am not so familiar with the edge cases of higher rank polymorphism. Does this change create cases where a pointfree version is not the same than an expanded version, missing type signatures excluded?

simonpj:
> Does this change create cases where a pointfree version is not the same than an expanded version

Yes indeed.  The proposal has several examples where `(g f)` is not well typed but `(g (\x. f x))` is.

jvanbruegge:
What is the reason for this? I don't really understand what makes the two expressions different. I always assumed that due to laziness eta expansion does not matter

simonpj:
>  I always assumed that due to laziness eta expansion does not matter

No indeed.  The proposal discusses this, but obviously not clearly enough.

Consider
```
f1, f2 :: Int -> Int
f1 = error "urk"
f2 = \x. f1 x
```
Here `f2` is the result of eta-expanding `f1`.  But they don't behave the same: 
```
f1 `seq` True     calls error "urk"
f2 `seq` True     returns True
```

treeowl:
Is there a way to express these axioms in Core so we won't need to eta-expand operationally?

simonpj:
> Is there a way to express these axioms in Core so we won't need to eta-expand operationally?

Alas no.  That's the point.   You could just about imagine it for type arguments, which are erased, although you'd need to extend System F in some as-yet-unpsecified way.   But even that won't work for dictionary arguments, which exist at runtime.

In any case, it really is not worth losing sleep over.  The point about this proposal is that the cost of giving these unsound eta-expansions is very slight in practice.   As I say in the proposal, if we didn't all the implicit eta-expansion, and someone proposed to add it, at the cost of silently changing the semantics of some programs, the propsal would never leave the launchpad :-).

Ericson2314:
> Is there a way to express these axioms in Core so we won't need to eta-expand operationally?

I left a comment in the impredicativity proposal, but it got lost. If we had 1-way Coercible, we could just add these back as 1-way coercion axioms.

serras:
Just to add some information, I am currently working on a branch of GHC which implements this proposal and #274.

aspiwack:
> > Is there a way to express these axioms in Core so we won't need to eta-expand operationally?
> 
> Alas no. That's the point. You could just about imagine it for type arguments, which are erased, although you'd need to extend System F in some as-yet-unpsecified way.

As a mostly theoretical aside, in System F (with implicit type abstractions, rather than explicit as in Core), you can express a subtyping relation

σ ⩽ τ ⟺ x:σ ⊢ x:τ

Except it's not quite a true subtyping relation. So there is an extension of F, called Fη, which adds the following typing rule

Γ ⊢ \x.u x : τ
    ------------------
Γ ⊢ u : τ

This essentially buys you contravariance of the left-hand argument of the arrow. It's a fairly well behaved system (where, however, pretty much everything is undecidable).

-----

In a system with non-termination, dictionaries, and decidable type checking, though, it seems that all these complications involve true, unavoidable, η-expansions. Which are not semantically valid.

These η-expansions are not very dangerous: they only make some diverging program converge. Though, of course, it may be surprising if a twiddling in the type signature of a function causes another function diverge. This may be a hard one to tease out.

But it also seems to be the case that these deep instanciations and skolemisation are regularly brought up in conversations about new features, in which they appear to cause problems. It may just be that these are feature which don't combine well with other type features! And is would rather be a problem.

goldfirere:
I'm in support of exploration about this -- curious for data about it all before committing.

But I thought of one complication: doing this would mean that the ambiguity check would have to become more intricate. Currently, we just check whether `ty <= ty`: if it doesn't, the type is ambiguous. With this change, `Int -> forall a. String` would be a sub-type of itself, even though it's plainly ambiguous. So we'd have to write a proper ambiguity checker. Not much of a problem, but it's an unwanted effect of this proposal.

simonpj:
> But I thought of one complication: doing this would mean that the ambiguity check would have to become more intricate

Yes, I agree.  But I also agree that it's not a reason to push this proposal one way or another, just a consequence.

proposals/0000-simplify-subsumption.md
```diff
@@ -0,0 +1,207 @@
...
+
+All four features are implemented in GHC.
+
+However, *all four change the semantics of Haskell*, by performing eta-expansion, changing bottom into non-bottom.  That’s a pretty serious change, which needs pretty serious motivation.  But actually all we get in exchange is a minor improvement in programming convenience.  So I argue that:
+
+* **The benefit, in terms of programming convenience, is small**.  In particular, while using these features accepts some more programs, any such program is easily accepted without these features, after a small edit.
+
+* **The cost is large**.  There is extra complexity in the compiler.  There is extra complexity in the formal description of the type system, and in any paper we write about it.  But, worst of all, the features change the semantics of the program.
+
+* A recent new cost is that the [Quick Look Impredicativity proposal]( https://github.com/ghc-proposals/ghc-proposals/pull/274) gains extra power by not having contravariance.
```
aspiwack: Also by not having covariance, I believe. Though maybe less dramatic?

proposals/0000-simplify-subsumption.md
```diff
@@ -0,0 +1,207 @@
...
+by removing
+
+* Covariance of function types
+* Contravariance of function types
+* Deep skolemisation
+* Deep instantiation
+
+This could be under control of a flag, but I propose that we
+take a clear path to removing these features altogether, i.e.
+ultimately, without a flag that enables them.
```
aspiwack: This sentence is very unclear to me.

simonpj:
> This has lingered for a bit, is it about time to submit this proposal to the committee?

We are awating data on the impact of the proposal on Haskell libraries, which Jurriaan is gathering.  This week I expect.

simonpj:
[paper.pdf](https://github.com/ghc-proposals/ghc-proposals/files/3968695/paper.pdf)

aspiwack:
Ok, so, to summarize: all of Stackage except two packages require relatively trivial changes (and a total of ~300 of these changes).

It'd be interesting to understand what changes are needed for the two package where it is non-trivial. Just in case the effect is rare but very strong.

It is fairly clear that if we were doing a new language, these results would be ample evidence that the simplified rules are quite sufficient, and that there would likely never be strong enough reasons to add deep instantiation and deep skolemisation later.

So maybe, even if the effect is rather dramatic on this handful of remaining packages, we should consider this proposal to be worth it.

simonpj:
Thanks for the reminder Arnaud.  Which "two packages" were you referring to?  Can you point to the particular bit of the paper you have in mind?  Here's an [up to date link](https://www.microsoft.com/en-us/research/publication/a-quick-look-at-impredicativity/)

I've updated the proposal with a pointer to the paper.

Secretary Joachim: I'd like to submit the proposal to the committee.

goldfirere:
@nomeata ^^

aspiwack:
>  Which "two packages" were you referring to?

Ah, sorry! In Section 7 of the paper (both the old and new link), it is mentioned that all but two packages were easy to fix. But, because you have basically given up on fixing these remaining packages, there is no measure of how hard it is to fix them. This worries me a little.

aspiwack:
A question: currently, the proposal stipulates that we should make a plan for phasing out these features, should we turn this into a concrete plan before merging?

As the shepherd for this proposal, I will give it a good read in a few days, and make a recommendation to the committee.

simonpj:
I believe you are referring to the sentence "two had issues we could not solve, e.g., because
of Template Haskell. We leave these two to the authors of these packages" Correct?

Suppose that the fix is to eta-expand a definition.  Then for these packages we must change TH code to *generate* eta-expanded code.  There's no difficulty here; just that you have to find the right place in the *generating* code, whereas for the other packages GHC would point direct to the source code line to change.  No more than that, I think.   Jurriaan may comment

simonpj:
> A question: currently, the proposal stipulates that we should make a plan for phasing out these features, should we turn this into a concrete plan before merging?

This is actually quite difficult. Ideally we'd like to say "Warning: in the next release of GHC this program won't compile, although it does now".  But that is super-hard to implement, short compiling every program twice, which our users will not like.

We could, I guess, produce a release iwth a flag saying "-XLessSubsumption" (modulo flag renaming), and ask users to switch that flag on themselves; the incentive being that in the next release it'll be on by default.

But I bet they never use it, so it'd just be wasted delay and effort.

We could offer a way to recover the old behaviour, so that they had a "quick fix" when the change happens.  But an equally quick fix is to do eta-expansion!

  I'm inclined to implement it early in a release cycle, advertise it heavily, and explictly warn users of the packages we've tried and fixed.  Advice welcome.

aspiwack:
This is a bit of a special situation, because, while backwards incompatible, code which compile after this change will also compile before it. So this is not a change which hurts those who want to write code which spans several compiler versions.

Given that, it seems acceptable, to simply roll the change without a change.

The alternative is to create a flag `-XLessSumption` in release 1, turn it by default in release 2, and delete the code only in release 3.

I think I'm going to be siding with your inclination on this. 

The latter is more polite, leaving to everybody control over their schedule, but then again, it means unnecessary code will linger around for a year.

Probably I'd favour swift deletion too.

simonpj:
> This is a bit of a special situation, because, while backwards incompatible, code which compile after this change will also compile before it. So this is not a change which hurts those who want to write code which spans several compiler versions.

Yes that's a key point.  Making the change to your source code is fully backward compatible.

goldfirere:
I agree with doing this quickly. It's not that the house is on fire, but the delay has real engineering costs and will have to be remembered by our users for some time. With the quick fix, users get a small surprise, a quick fix, and resume their programming with nary an unhappy memory.

jurriaanhage:
Simon asked me to comment, being the one who did the experiments. About the eta-expension, my guess is it is best for the authors to just do them. I did it for all of them, and the fixing took little time, it was mostly the compiling of dependencies. In all the eta-expension cases, the type error message basically underlines the expression to eta-extend. 

The two where we needed a signature were clear from the message too, in that the error was for a locally defined identifier without a signature. 

About those two hard nuts: one of these indeed involved TH, and the main problem is then that the code that is generated is not sufficiently eta-extended. Given our "success" with the other packages, I do not expect other issues to pop up. The reason i did not attempt this, is that it needs intimate knowledge of the package that I do not have, but the author of the package should. 

The other one may have straightforward fixes, but I was not familiar enough with the type system extensions being used to make those changes. I tried a few things that worked in other situations, like eta-extending various subexpressions but got nowhere. The problems were all comparable, but I simply did not get what the code was achieving. I could have asked the author of the package to see if he can, but did not. Just in case you are interested to know what the packages are, 
these are `singletons` and `labels'.
 
If you need to know anything else, please let me know.

proposals/0000-simplify-subsumption.md
```diff
@@ -0,0 +1,209 @@
...
+
+* **The benefit, in terms of programming convenience, is small**.  In particular, while using these features accepts some more programs, any such program is easily accepted without these features, after a small edit.
+
+* **The cost is large**.  There is extra complexity in the compiler.  There is extra complexity in the formal description of the type system, and in any paper we write about it.  But, worst of all, the features change the semantics of the program.
+
+* A recent new cost is that the [Quick Look Impredicativity proposal]( https://github.com/ghc-proposals/ghc-proposals/pull/274) gains extra power by not having contravariance.   More details in the [Quick Look paper](https://www.microsoft.com/en-us/research/publication/a-quick-look-at-impredicativity/)
+
+Here is a thought experiment.  Suppose GHC lacked all four features, and someone proposed adding them.  That proposal would never leave the launchpad. A minor change in programming convenience, in exchange for changing language semantics?  No.  If that’s true, then the only issue would be back-compat issues: how many libraries would be affected, and how painful they would be to fix.  We need to get data on that.
```
aspiwack: > We need to get data on that

We now have :slightly_smiling_face:

proposals/0000-simplify-subsumption.md
```diff
@@ -0,0 +1,209 @@
...
+## Effect and Interactions
+
+* See Section 7 of the [Quick Look paper](https://www.microsoft.com/en-us/research/publication/a-quick-look-at-impredicativity/) for a detailed analysis of the practical impact of these changes.
+* Everything (specification, implementation) becomes a bit simpler
+* Quick Look Impredicativity gains more power
+
+## Costs and Drawbacks
+
+The main user-facing cost is that some existing programs will require some manual eta-expansion.
+Jurriaan is gathering data on how many libraries are affected.
```
aspiwack: This last remark is no longer accurate, and is superseded by the link to the paper above.

simonpj:
THanks.  I have addressed the nits.

gridaphobe:
> This is actually quite difficult. Ideally we'd like to say "Warning: in the next release of GHC this program won't compile, although it does now". But that is super-hard to implement, short compiling every program twice, which our users will not like.

Would it be possible instead to make the change in GHC, but detect type errors that would likely be fixed by eta-expansion and issue a fixit hint to the user? I'm not sure if that's any easier than what GHC currently does, but it would make the change in behavior a lot more palatable to users.

simonpj:
> detect type errors that would likely be fixed by eta-expansion and issue a fixit hint to the user?

I don't think it'd be easy to identify errors likely fixed by eta-expansion.  We could try.

gridaphobe:
One possibility, since we already have the deep skolemization etc code, might be

1. perform the new check
2. if (1) fails, perform the old check
3. if (2) succeeds, emit a fixit hint suggesting eta expansion

This would mean keeping the old code around for a bit, but maybe one major release cycle would be sufficient?

simonpj:
> if (1) fails, perform the old check

Alas no.  We generate constraints and solve them later.  We don't know if they are going to fail.  I don't think it is AT ALL easy to tell when the old thing would work and the new one would not. It's not just reluctance to keep the old code around.

aspiwack:
Dear all. This proposal has been accepted by the committee. Including the early deletion of the rules being deprecated.

@nomeata are you responsible for the merging thing, or should I?

Icelandjack:
The suggestion for deep skolemization is to eta expand `g (\x y -> f x y)`, for that particular example in the description does `g (\x -> f x)` suffice?

goldfirere:
In talking with @Gertjan423, I just ran across a troubling scenario. Suppose I write

```hs
f x = (+)
```

in a program. What is `f`'s type? The type of `f`'s RHS is `forall a. Num a => a -> a -> a`. I might expect then that `f`'s type is `forall {b}. b -> forall a. Num a => a -> a -> a`. Today, GHC infers `forall {a} {b}. Num a => b -> a -> a -> a`: with the `forall` brought to the top. GHC's habit of doing this is, today, inconsequential: the two possible types for `f` are equivalent with respect to today's subtyping relation. But they are *not* equivalent with respect to the subtyping relation proposed here. So our choice matters.

Which will it be?

I think it has to be `forall {b}. b -> forall a. Num a => a -> a -> a`, despite the strange shape of that type. Otherwise, we lose substitutivity: imagine replacing `f` with `\x -> (+)`, which will now have a different (and incomparable) type.

simonpj:
Richard, Gert and I discussed.  In the end we concluded that the inferred type should be
```
f :: forall {a} {b}. Num a => b -> a -> a -> a
```
Two contributing factors.   
* If it was just
  ```
  f = \x -> (+)
  ```
  then it's legal Haskell today, and is subject to the Monomorphism Restriction.  We need to instantiate that `(+)` to see that fact.

* If it was
  ```
  f True = (+)
  f False = error "Uk"
  ```
  with two branches, then today we insist that the two RHSs have monotypes (see `TcMType.tauifyExpType`).  So we will definitely get
  ```
  f :: forall {b}. Bool -> b -> b -> b
  ```
  It would be jolly odd if adding a catch-all case changed the type!

goldfirere:
It's actually not at all clear from the Report (https://www.haskell.org/onlinereport/haskell2010/) that the monomorphism restriction applies to `f = \x -> (+)`. It says that the MR stops type variables from being generalised. But I don't see a place where the Report says that a bare `(+)` is instantiated. So, while the MR certainly applies to `f = \x -> (+) x`, it's defensible to say that it doesn't apply to `f = \x -> (+)`.

I'm not changing the conclusion above, and the second point in Simon's post is quite true. But perhaps the MR is a bit of a red herring.

simonpj:
> But I don't see a place where the Report says that a bare (+) is instantiated

Well the report is not fully formalised.  But wihtout instantiating `(+)`, f woudl have a tuype like
```
f :: forall a. a -> forall b. Num b => b -> b -> b
```
and this is simply not a legal Haskell 98 type.  So I think that yes, Haskell 98 does require instantiation.

goldfirere:
I think I'm changing my mind here, on the second point in https://github.com/ghc-proposals/ghc-proposals/pull/287#issuecomment-579853918.

The argument is this. Suppose we had lazy instantiation (as we do today), meaning that `f (x :: Bool) = id` would infer type `Bool -> forall b. b -> b` for `f`. (This is not the type currently inferred because we instantiate deeply. We will only be able to instantiate shallowly with this proposal.) Simon observes that if we instead wrote

```hs
f True = id
f False = id
```

then we would infer `forall {b}. Bool -> b -> b`. This is true. What I disagree with is that this is a bad thing.

Today, if I write

```hs
higher :: (forall a. a -> a) -> ()
higher _ = ()

f1 (x :: Bool) = higher

f2 True  = higher
f2 False = higher
```

then `f1` is accepted with type `Bool -> (forall a. a -> a) -> ()` while `f2` is rejected, as we cannot have a polytype as the return type of a `case` (an implicit one here). `f1` is accepted because we treat one-equation non-recursive definitions specially. And we have for some time, without complaint. The treatment above that Simon objects to is the same: a special treatment for a single defining equation. So my claim is that if special treatment is OK for this second example, it should be OK for the first.

So I have defeated both of Simon's arguments. The MR restriction need not apply at `f = \x -> (+)` with `RankNTypes` enabled (without it, we should always brutally `tauify` as we type-check), and the "strange" behavior around `case`-expressions is deemed acceptable.

Why do I care?

Witness various strangenesses with Simon's approach (in concert with this proposal):

1.

If I write

```hs
f = ((\x -> x) :: forall a. a -> a)
```

GHC assigns type `forall {a}. a -> a` to `f`. Even though I said what type it should have!

In terms of a formal presentation, this means that an annotated let is different from a `let` of an annotated expression, which is, well, strange.

2.

Suppose I write

```hs
g1 :: Int -> forall a. a -> a
g1 = ...

h1 = g1
```

GHC assigns type `Int -> forall a. a -> a` to `h1`. After all, there is no way to assign a *different* type without eta-expansion, which this proposal forbids.

Yet now consider

```hs
g2 :: forall a. a -> a
g2 = ...

h2 = g2
```

GHC assigns type `forall {a}. a -> a` to `h2`. Note that the variable specificity has changed! And variables might be reordered, too. Why this discrepancy? It's unexpected.

3.

If I have

```hs
g3 :: Int -> forall a. Num a => a -> a -> a
g3 _ = (+)

h3 = g3
h4 = g3 42
```

GHC assigns type `Int -> forall a. Num a => a -> a` to `h3` (effectively bypassing the MR). Yet it assigns `Integer -> Integer -> Integer` to `h4` because of the MR. Deeply strange.

No. Let's stop this madness. Just instantiate lazily, and all will be well. This does mean that, sometimes, GHC will infer types like `forall a. -> forall b. b -> b` (if we typed `f _ = id`, for example), but that type accurately describes what the user wrote. It is easy to arrange for such exotic types only when the user has turned on `-XRankNTypes`: without the extension, just skip the `InferResult` mechanism GHC uses to deal with higher-rank types. If we always use unification variables, we'll get nice types, having instantiated them and regeneralized. (Note that all my "strange" examples are prohibited without `-XRankNTypes`, so they don't apply in that fragment of GHC.) In addition, lazy instantiation jibes with the most recent publication on the matter ([Visible Type Application](https://repository.brynmawr.edu/cgi/viewcontent.cgi?article=1001&context=compsci_pubs)) and is a simpler type system, without the need for a special form for checking heads of expressions. (See https://gitlab.haskell.org/ghc/ghc/issues/17173, which explores eager instantiation.)

simonpj:
About the "special case for single equations" point, currently the difference between
```
f1 True = \(y :: forall a. a->a). blah

f1 True = \(y :: forall a. a->a). blah
f2 False = error "urk"
```
is that `f1` is accepted while `f2` is rejected.   Moreover this difference shows up only with RankNTypes.  In Haskell 98 there is no difference.

With lazy instantiation, the difference between
```
g1 True = (+)

g2 True = (+)
g2 False = erorr "urk"
```
is that both are accepted, but with different types:
```
g1 :: Bool -> forall b. Num b => b -> b -> b
g2 :: forall {b}. Num b => Bool -> b -> b -> b
```
Moreover, even though *no extensions are specified at all*, `g1` has a type that is not even legal in Haskell 98.  Probably no one will observe that difference, but `:info` or `:type` will show it.

Finally, [as I say above](https://github.com/ghc-proposals/ghc-proposals/pull/287#issuecomment-580245837), `g3 = \x -> (+)` is subject to the MR in Haskell 98, and so (absent extensions) it really should be in GHC.

Also: we will soon have a special form for checking heads of expressions anyway, for impredicativity; and in fact we do already (to some extent) in an effort to improve error messages.

In short
* There are bumps in the road either way. 
* Probably no one will notice these changes in specificity.
* For example, the single-equation case  has been there for years without causing trouble, even though it means that single-equation definitions are treated subtly differently -- indeed it is positively useful.  (For interested folk, it's `Note [Single function non-recursive binding special-case]` in `TcBinds`.)
* To adopt your proposal would mean we did not adhere strictly to Haskell 98, even absent extensions.  But again maybe no one will notice.

It's not a decision with deep technical consequences; it's a decision with smaller user consequences.  I don't feel terribly strongly.  I'm less fussed than you are about your 1,2,3.  But I'd be happy to get advice from others.

goldfirere:
I disagree with some of your examples.

The first `f1`/`f2` example is possible only with `-XRankNTypes` enabled, because it mentions a `forall` outside of a top-level type signature. So the behavior of that example in Haskell98 is moot.

With, `g1` and `g2`, you claim "Moreover, even though no extensions are specified at all...." But without `-RankNTypes`, I am proposing that both are inferred to have the type you have assigned to `g2`. While instantiation is lazy in all modes, without `-XRankNTypes`, I am proposing to rigidly infer every expression to have a monotype. (In terms of the implementation, this means to invent a unification variable and check against that variable, bypassing the `InferResult` mechanism. Very easy, and local, to implement.) This will make instantiation effectively eager in Haskell98. This means that `g1` and `g2` get the same type, and that the MR is applied appropriately. But, when we enable `-XRankNTypes`, we infer `forall`s to the right of arrows. I see that as a feature, not a bug.

So:

> To adopt your proposal would mean we did not adhere strictly to Haskell 98, even absent extensions. But again maybe no one will notice.

I roundly disagree. My proposal does mean that inference changes between a blank slate and `-XRankNTypes` (in a way unlikely to be noticed). This is different to the status quo, where (I believe) `-XRankNTypes` changes nothing about any existing programs.

> It's not a decision with deep technical consequences; it's a decision with smaller user consequences. I don't feel terribly strongly. I'm less fussed than you are about your 1,2,3. But I'd be happy to get advice from others.

I partially agree here. There *are* technical consequences, in that eager instantiation seems harder to specify, meaning a whole new typing judgment, for example. I don't know if users will notice that, but @Gertjan423 and I do. I, too, am happy to get advice from others. Perhaps a good next step is to create a new ticket in this repo, laying out the problem and the possible solutions, and seeking input.

simonpj:
>  But without -RankNTypes, I am proposing that both are inferred to have the type you have assigned to g2.

Aha.  Either you didn't say that, or I missed it.   So without `RankNTypes` we get eager instantiation (in effect, albeit implemented in the way you describe), while with `RankNTypes` we get lazy instantation.   That changes things; indeed, I agree that, since with no extensions we get eager instantiation, we make not change to H98 behaviour. 

So without `RankNTypes`, the definition `g3 = \x -> (+)` is subject to the MR; but with `RankNTypes`
*  `g3 = \x -> (+)` is not subject to the MR
*  `g4 = (+)` is not subject to the MR
* `g5 = \x -> (+) x` is subject to the MR
* `g6 = if debug then (+) else (-)` is subject to the MR.

I agree that eager instantiation involves treating application heads specially.  But so does impredicative polymorphism.  It's all laid out in [the paper](https://www.microsoft.com/en-us/research/publication/a-quick-look-at-impredicativity/) (including VTA).   

Hmm.  I have not thought recently about whether quick-look impredicativity is compatible with lazy instantiation.  I suspect that, at best, it would complicate both the specification and implementation.   In particular, the INST rule in Fig 4 returns a rho-type not a sigma type. At one pont it was a sigma, but I remember it all became much simpler when we moved to rho.

NorfairKing:
Would the authors of this PR please weigh in on the discussion at https://github.com/commercialhaskell/path/pull/170#issuecomment-775066947

> In our testing so far we've not found any Lift instances relying on any special methods of Q.

The `path` library seems to be an example of this, and I would like to ask for some advice on how to proceed over there.

serras:
@NorfairKing Shouldn't this be #246 instead?

NorfairKing:
@serras WHOOPS, yes it looks like it.
Sorry, I don't know how this happened.

Bodigrim:
> See Section 7 of this draft paper
> 
> [paper.pdf](https://github.com/ghc-proposals/ghc-proposals/files/3968695/paper.pdf)
> 
> which quantifies the impact of this proposal.

Quoting from the paper: 

> To substantiate the claim, we implemented Quick Look on top of the latest incarnation of  GHC (ghc-8.9.0.20191101). ... We started from collection of packages for a recent version of GHC, obtained from Stackage (LTS 13.6), containing a total of 2,332 packages. We then selected only the 607 packages that used the extensions `RankNTypes`, `Rank2Types` and `ImpredicativeTypes`; the other 1,725 packages are certainly unaffected.

One does not need to enable `RankNTypes` to be affected. E. g., if

```haskell
{-# LANGUAGE RankNTypes #-}

f3 :: (Int -> forall a b. a -> b -> b) -> Bool
f3 x = undefined
g3b :: forall x. Int -> forall y. x -> y -> y
g3b x = undefined
```

are exported from a package, other packages do not need to enable `RankNTypes` to use `f3 g3b`.

> Of the 607 packages, 150 fail to compile for reasons unrelated to Quick Look – they depend on external libraries and tools we do not have available; or they do not compile with GHC 8.9 anyway.

This statement is surprising at the first glance: the very purpose of Stackage LTS is to be fully compatible and consistent. However, the authors took LTS 13.6 which expects GHC 8.6, but used GHC 8.9. A proper testing would require either using LTS 15 (compatible with GHC 8.8) or impementing Quick Look changes atop of GHC 8.6 instead of GHC 8.9. The approach used indeed excludes most complex and fragile packages from the examination.

Ericson2314:
Thank you @Bodigrim for that investigation!

I am excited for https://github.com/haskellfoundation/tech-proposals/pull/27 to ensure this sort of thing doesn't happen again, because the experimental method for doing such regression tests is clearly documented / automated / well-known, and not something researchers have to put put together on the fly.

noughtmare:
>  proper testing would require either using LTS 15 (compatible with GHC 8.8) or impementing Quick Look changes atop of GHC 8.6 instead of GHC 8.9.

LTS 15 didn't exist yet and I can imagine implementing Quick Look on top of an old version of GHC and later "forward-porting" it to GHC 8.9 would have been too much effort. Perhaps a nightly version could have been used, but that also doesn't seem like it is obviously a good choice.

Ericson2314:
@noughtmare I think backporting Quick Look would have been annoying, but backporting simplified assumption alone not nearly as much. Still, part of the idea of https://github.com/haskellfoundation/tech-proposals/pull/27 (as I understand it) is that `head.hackage` should *also* be in better shape (and we could imagine a draft Stackage for GHC head likewise) so back-porting ought not to be necessary.

Bodigrim:
@noughtmare there’ve been already https://www.stackage.org/nightly-2019-12-16 with GHC 8.8 available at the moment of the impact assessment. 

Essentially, the impact assessment has shown that 59 packages out of 458 analysed were broken by the simplified subsumption, which is 15%. If we agree to extrapolate this rate over all Haskell ecosystem, the ongoing public outcry is no longer surprising.

gbaz:
Characterizing this as "ongoing public outcry" is a bit much. GHC 9 has been out for a very long time and many people and libraries adapted with at most some minor grumbling. Only in the last few days have there been two very ranty threads seemingly designed to stir up outrange, and which succeeded in garnering a number of people agreeing with that outrage.

This is not an ongoing outcry. Its some ongoing grumbles, and then one person kicking up a giant fuss and in so doing getting a lot more people a lot more outraged than before, including people who were relatively unaware of and unaffected by the ongoing migrations but now who are convinced that there is some "lost principle of purity of the language" which has been taken by them.

Bodigrim:
@gbaz sorry, it might be that my English is at fault here. I used “ongoing” in the sense of [“being actually in process”](https://www.merriam-webster.com/dictionary/ongoing), “current”, which seems to me an accurate description.

chrisdone:
> Only in the last few days have there been two very ranty threads seemingly designed to stir up outrange

This seems an uncharitable interpretation.

> GHC 9 has been out for a very long time and many people and libraries adapted with at most some minor grumbling. 

For what it’s worth, my work projects and personal projects are all still on GHC 8.10.7 to this day. I think that “very long time” is relative: open source packages move at a certain pace, and private and industry work moves at another pace. I’m not taking this opportunity to comment on what that says about GHC’s release pace, but rather that your dismissal of “outrage” based on the passage of time seems to lack this nuance.

goldfirere:
While I can't say I like outrage, I'm happy to see the feedback that is being communicated. Some years ago, I remember a proposal that would cause breakage (sadly, I forget which one), and we maintainers were worried that the breakage was too big. A handful of industry folks piped up saying that the improvements were worth the breakage. This surprised me at the time. The industry folks were right: whichever proposal it was (the evidence is on GitHub somewhere if you'd like to trawl through old proposal conversations) did not cause outrage when it landed. But the experience taught me not to fear breakage so much. On the other hand, this new experience communicates to me, as a language maintainer (= member of the GHC Steering Committee), more about the cost of breakage. This is helpful feedback -- I'm inclined to listen to our users, here.

gbaz:
The most interesting thing I've learned from the lengthy threads, along with deficiencies in the testing process for ecosystem breakage, is how much absolute confusion there is over how rank-n-types work, and the limitations and complexities of algorithms pertaining to them. Now that they are in widespread use and on by default in ghc, it seems like an important case where user education is warranted. I would hope that in the long run, simplified subsumption helps in this education.

While a few things jumped out at me in terms of confusions, I think the most important is that there is a naive view that type signatures should not affect the meaning of terms -- certainly this held true in the early days of Haskell, but that time is long past. The fact that rank-n-types mean that most general unifiers do not necessarily exist is certainly not widely understood. Similarly, the fact that equivalent terms may have different cost profiles is also not so well understood -- and the fact that typing _constraints_ may affect this even less so.

In general, this was also a collision caused by an advanced feature with sharp edges getting pushed into widespread industrial/advanced-beginner usage as a way to simplify code, and to prevent users from needing to think about constraints. So perhaps there is another way to make constraints "less scary" that is also less fragile? Or a richer system than current type aliases for letting people write more expressive type signatures? Or perhaps it would be nice to have syntactic sugar for "eta expand and fully instantiate this quantified/polymorphic object".

xplat:
That syntax would certainly be nice for eta expanding things that don't have any explicit arguments to begin with...

Ericson2314:
> is how much absolute confusion there is over how rank-n-types work, and the limitations and complexities of algorithms pertaining to them.

There is confusion, but there is also disagreement among non-novices about what exactly is in play, too.

## Other ideas

### Patched-up eta expansion

@xplat should me that by simply doing eta expansion as 
```haskell
f ====> let f' = f in f' `seq` \x -> f' x
```
we fix the operational semantics. Certainly this is not ideal: the non-0 cost in extra allocations is unsavory, and can rise to practical problem in the presence of recursion in the form of repeated eta expansions wasting memory in `O(1)` ways. Still, this simple solution directly address the issues raised in the proposal --- not including it as an alternative feels like rigidly characterizing the status quo so as to reduce it to a straw men.

To be clear, I am *absolutely not* insinuating @simonpj and @serras thought of this and purposely didn't mention it. Rather, I am wondering why this counter-proposal didn't come up in the course of process evaluation. (Certainly I could have made a bigger stink raising @xplat's idea then, too.)

### Fancy coercions

Secondly there is the matter of ideal solutions. While as I said just above, I neglected to mention @xplat's simple solution. I did mention very rough idea far fancier ideas of one-way coercions that would make non-simple subsumptions

Of course all, that is super tiring to formalize / implement, and we can't just ask @serras to double the length of his PhD. However, we *should* acknowledge the fundamental correctness of regular user's intuitions:

Ignoring the details of *today's* type system what fancy subsumption is trying to do is fundamentally something that need not have any cost. This follows from the basics of type erasure, and why we chose to call these the removed features "subsumption" despite their sub-par implementation in the  first place. Thus a desire "to have our cake and eat it too" on the part of users that "don't know the details" is, in fact, fundamentally reasonable, and not an optimism stemming from ignorance to be discarded.

### Fancy type systems

> The fact that rank-n-types mean that most general unifiers do not necessarily exist is certainly not widely understood.

However, in alternative type systems like MLF, such terms *do* have principal types. The funny MLF types with their subsumption-bounded type variable quantification in fact ought to correspond to the the one-way coercions I was vaguely gesticulating at.

While MLF is an actual published idea, and thus far easier to discuss than rough undirectional coercion ideas, the practical hurdles to implementing it in GHC are no less, because it is such a departure from the current (80,000 line!!!) type checker. And MLF has a bit of a reputation of being hard to confirm to existing languages, or, worse, a research dead end. One again, of course we cannot ask @serras to double the length of his PhD, especially as Haskellers have looked at MLF with an eye for GHC before. Still, once again the "naive optomism" of beginners does has more credence than we might give it, once we broaden the survey of existing research to take into account.

## Conclusions

### It's just Muddled

Underlying @gbaz's post I think is a notion that widespread appreciation for the truth should bulldoze through the controversy. But I would argue based upon the above that the truth is more muddled. In classic post modern-fashion, there are all sorts of "Map–territory" problems in that the more one imagines we could retrofit our existing programs to fit different "alternative" type systems, the less clear what our conclusions should be becomes.

### Focus on testing

In light of that, I think it is best to mainly focus on the testing store. The theory story is extremely complicated, and hard to convey --- I imagine my attempt to write this are going to make very little sense for most readers!

Conversely, the testing/empiricism post-mortem @Bodigrim touched upon is extremely clear, and begets extremely clear action-items with https://github.com/haskellfoundation/tech-proposals/pull/27. I don't think there is any disagreement about what is needed to ensure those problems will not happen again, and indeed future research will be *easier* without researches having to come up with the regression testing protocol on the fly.

For the sake of community cohesion and tackling the low hanging fruit first, I really want to **not** discuss both the "people don't know enough" points and spooky "alt type theories raise question" points I wrote about until these consensus-building measures are underway.

Ericson2314:
Also, as I have written elsewhere, I really like how https://richarde.dev/papers/2021/stability/stability.pdf tries to formally characterize what we want out of a type system. I think this sort of "academic approach to ergonomic concerns" work is vastly underutilized, and can help us resolve some industry-vs-academy tensions we face.

Any future proposals like QuickLook and simplified subsumption I hope would leverage the "principles" document we now have in this ghc-proposals repo to incorporate a good deal of that sort of material.

serras:
To give some context to where this discussion is coming from, here's the Reddit thread https://www.reddit.com/r/haskell/comments/ujpzx3/was_simplified_subsumption_worth_it_for_industry/

ParetoOptimalDev:
> GHC 9 has been out for a very long time and many people and libraries adapted with at most some minor grumbling.

Applications with up to hundreds of dependencies are now being adapted that weren't before. Very long time is subjective enough here to not make a strong point IMO.

> Only in the last few days have there been two very ranty threads seemingly designed to stir up outrange, and which succeeded in garnering a number of people agreeing with that outrage.

@gbaz Wowwwwww... okay. Conspiracy theories.

Not designed to stir up outage, posted to give an experience report and share frustration from a perspective that didn't seem considered enough.

> which succeeded in garnering a number of people agreeing with that outrage.

Perhaps in confirming your reality everyone was drunk on outrage you missed the various people who said it bothered them, but they didn't have the energy or confidence they could make a difference?

> and then one person kicking up a giant fuss 

Should I not post how a change inconvenienced me, my teammates, and other Haskellers I know? Are real world experience reports not valuable?

If you're trying to discourage me and others like me from contributing... You are doing a great job of sowing a fear of anyone who dares to share their less than perfect experiences being interpreted so uncharitably as to sound like a petulant child.

Thanks for making my day worse @gbaz.

> people who were relatively unaware of and unaffected by the ongoing migrations but now who are convinced that there is some "lost principle of purity of the language" which has been taken by them.

You have a very low opinion of the Haskell community outside of your in-group it seems.

goldfirere:
I just want to [reiterate](https://github.com/ghc-proposals/ghc-proposals/pull/287#issuecomment-1120342880) that I see this discussion, including the Reddit thread, as valuable feedback. I also see a number of misunderstandings in the various posts; I interpret these as an indication that there is an outreach / education piece missing somewhere.

Internet communication is hard. For example, one post on Reddit claims that simplified subsumption breaks the set-interpretation of constraints, claiming that `(Eq a, Show a) =>` is now somehow different than `(Show a, Eq a) =>`. I started to rebut... but this statement is actually true, when the constraints are buried in a higher-rank type in just the right way. (The statement was also true before simplified subsumption, in even more obscure setups.) The problem is that a casual reader could see the headline claim ("constraints are broken!") and easily think that this statement applies broadly, instead of in a very narrow and unusual setup. I don't know how to fix this, but it's these false generalizations that I worry about in outrage threads -- not so much the more targeted unhappiness about a particular feature (or lack thereof).

I think this all boils down to trust: I would like to think we can trust our community to complain about issues that are affecting them -- in specific, accurate ways -- and I would hope that they can trust maintainers of core infrastructure that we would take the complaints seriously. (Evidence: https://gitlab.haskell.org/ghc/ghc/-/issues/21548.) There are always sub-optimal behaviors on both ends, but the trust part, for me, means that we see these as exceptions, rather than the rule.

simonpj:
John suggests "patching up" eta expansion thus
```
f ====> let f' = f in f' `seq` \x -> f' x
```
On the face of it, that seems like quite a sensible suggestion, and indeed I myself thought that it could be made to work.
But sadly it doesn't, at least not in general.
Consider the application `(f (g expensive))`
```
f  :: Int -> (forall b. Eq b => [b] -> [b])
g  :: forall a. Eq a => Int -> Int -> a -> a
```
where `expensive` is some complicated expression.  The application is rejected under simplified subsumption.  But with deep subsumption we can elaborate to
```
f (\(x:Int). /\b. \(d :: Eq b).
   let d2 :: Eq [b] = $dfEqList d  -- Make an Eq [b] dict from an Eq b dict
   in g @[b] d2 expensive x)
```
We can't pull out that entire argument `(g expensive)` into a let-binding, as John suggests, because it mentions `d2`.  We could, in principle pull out the subexpression `expensive`, but in general `f`'s argument may be arbitrarily complicated so it's not clear how to do this in general.

And it gets worse.  Consider the call `(f (g2 27))` where
```
g2,g3,g4 :: forall a. Eq a => Int -> Int -> a -> a
g2 n | expensive n = g3
     | otherwise   = g4
```
You might hope that `g2 27` will evaluate `(exensive 27)` just once, and choose `g3` or `g4`
depending on the result.  But the elaboration is precisely as above:
```
f (\(x:Int). /\b. \(d :: Eq b).
   let d2 :: Eq [b] = $dfEqList d  -- Make an Eq [b] dict from an Eq b dict
   in g2 @[b] d2 27 x)
```
Now we can't let-bind `(g2 @[b] d2 27)` outside the `\x` because neither `b` nor `d2` are in scope there.  Nor can we lambda-abstract over `b` and `d2` because that would defeat the sharing we are trying to recover.

**Summary**.  I know of no general way to do deep subsumption without losing sharing.

simonpj:
The original proposal said *"It would be possible to offer a flag that restored the old behaviour, but that still means changing the .cabal file, or adding a LANGUAGE pragma. It seems more straightforward simply to change the source code to work with the new restrictions."*  Maybe that was a bad decision, and we should have added an extension, say `DeepSubsumption` to restore the old behaviour.

I have done enough work on [GHC MR !8210](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/8210) to convince myself that with hundred extra lines of code, almost entirely confined to one module, we could implement `-XDeepSubsumption`.

I don't recommend using it, for the reasons discussed in the original proposal.  (As mentioned [here](https://github.com/ghc-proposals/ghc-proposals/pull/287#issuecomment-1125419649), I don't think the silent loss of sharing and change of semantics is avoidable.) But it would, I suppose, allow users to say "I'm prepared to trade the occasional silent loss of sharing for the convenience".  Rather as `NoMonoLocalBinds` allows you to say "I'm prepared to trade a less predictable type inference for best-efforts generalisation on local bindings".  There could be some dark corners, but the idea would be to get 90% of the benefit with a small, well-contained change.

Would adding `DeepSubsumption` be a step forward?  I'm pretty agnostic.
* I really think that deep subsumption was a mis-feature.  To be clear, I do not blame library authors for using it, because it can be convenient; rather, I blame myself for implementing it in the first place.
* If it wasn't in GHC and someone wrote a GHC proposal to suggest adding it, I'm sure we would never do so.  That is a useful thought experiment.
* But the fact is that we have lots of libraries that use it, so if it'll make our ecosystem happier I would not stand in the way.  On the contrary, motivated by theseexpressed concerns, I have invested some hours in the feasibility study I mention above.  (I am afraid that `DeepSubsumption` would need to be on in *clients* of these libraries too.)

How to move forward? It's open to anyone to propose a delta to the existing proposal that reverses that choice, and adds a language extension `DeepSubsumption` to restore the original behaviour.  That need not take long, and the implementation (if accepted) is mostly done.  But it would be useful to establish a consensus about whether this is the path we should follow.  E.g. maybe the ship has sailed.

There are doubtless lessons about advance warning of breakage, and increasting the set of libraries that we use for "smoke-testing" GHC before any release.  But that's another story.

treeowl:
When there are no superclass relationships involved in the subsumption, some of the problems are less inherent. This applies, for example, to instantiating parametric polymorphic functions and (as a special case) to instantiating multiplicity-polymorphic functions. In these cases, there's no *inherent* reason that operational eta expansion has to happen; are the necessary changes to Core too difficult?

Ericson2314:
@simonpj correct me if I am wrong, but the issue with your examples fundamentally stems from the fact that constraints are not erased, right? I think @xplat's plan (don't give me credit) works for erased stuff just fine, but there is no solution for unerased stuff.

This makes me think we should destinguish two sorts of `-XDeepSubsumption`, because one of them is perfectly safe, and one of them can never be.

Ericson2314:
(On a more meta note, if I didn't make a mistake above, I think this an excellent example about how the DH principle from @goldfirere of disentangling visibility vs erasure is a very important thinking aid.)

Ericson2314:
Also, I think you example is ill-typed. `f  :: (Int -> forall b. Eq b => [b] -> [b]) -> _` worked for me in GHCi; did you mean that?

Ericson2314:
@simonpj

@xplat and I were exploring and well....there's a lot going on

## Erasure

I am not sure, but I probably misspoke, and erasure is quite enough for this sort of eta expansion (though it could help with fancier things).

Still, I am wondering where do I read about Core's evaluation semantics? (That seems like something that should be easy to Google and yet I am completely failing!)

Being able to evaluate under big lambdas helps (I think I read that in the paper from Richard I was pouring over), but there are two cases based on whether the big lambda is being pulled out vs pushed in, and I keep on messing up in my head what exactly I need to prove.

## Pulling apart arguments more fancily

You write

> We could, in principle pull out the subexpression `expensive`, but in general `f`'s argument may be arbitrarily complicated so it's not clear how to do this in general.

@xplat taught me about `-ffull-laziness` which does seem relevant! It is more theory- and implementation- heavily lifting to use that, but it might be useful to keep in mind as a back up plan.

## `g2 27`

That example struct me as funny, and it took me a second to figure out why but consider this program:

```haskell
import Debug.Trace

expensive' :: Int -> Bool
expensive' t = trace "oof" $ t `mod` 23 == 0

g2 :: forall a. Eq a => Int -> Int -> a -> a
g2 n | expensive' n = \_ -> id
     | otherwise    = \_ -> id

foo :: forall a. Eq a => Int -> a -> a
foo = g2 27

test = foo 0 [22] ++ fmap (\_ -> 33) (foo 1 [(), (), ()])
```

evaluating `test` prints "oof" twice, with simplified subsumption (GHC 9.2)! 

In ways @xplat me teaches me relate to old monomorphism restriction debates (though the explicit signature side-steps that) we have an implicit eta expansion for the `Eq a` dictionary that thus makes `foo = g2 27` not a CAF. At least we have the explicit signature to indicate something is up, but it still is a bit odd from simplified subsumption's philosophical standpoint.

In any event, the fact that `g2 27` is weird regardless of context makes me a little unsure where the "goal posts" should be finding a satisfactory handling of your second example.

tomjaguarpaw:
> @xplat taught me about `-ffull-laziness` which does seem relevant!

I don't know if you're aware, but `-ffull-laziness` is catastrophic from the point of view of introducing unwanted sharing, and in consequence, space leaks.  See, for example, https://gitlab.haskell.org/ghc/ghc/-/issues/8457.

simonpj:
> Also, I think your example is ill-typed. f :: (Int -> forall b. Eq b => [b] -> [b]) -> _ worked for me in GHCi; did you mean that?

Correct, thanks.  Fixed,

LeanderK:
I was very annoyed by simplified subsumption and would very much like the option to turn it on again. Even if I loose the occasional sharing

Ericson2314:
@tomjaguarpaw this would be an extremely targeted use of the `-ffull-laziness` algorithm only to share things as much as they were prior to eta expansion. So no *net new* sharing / space leaks. However, that as a solution breaks down when the expression being eta-expanded is abstract (e.g. in another module, without an unfolding) so I rather not rely on the `-ffull-laziness` algorithm not regress in that department.

(You can think of the `-ffull-laziness` algorithm as trying to statically approximate head normal form (no "weak") evaluation. For commuting non-erased quantifiers (e.g. constraints) I think HNF eval is the only solution and a complete non-starter in any world like today. But erased ones seem easier and that is why I am focused on them.)

gbaz:
I suspect that "putting deep subsumption back" isn't really solving the problem people need solved.

The problematic cases seem to arise from two things.

Most notably: type aliases of the form `type Foo = forall a. X` (or `type Foo = forall a. (Constraint a) => X`).

What do people actually intend with such aliases? They don't intend, I suspect, from what I've seen, that these type aliases actually expand out directly into a fancy quantified thing. Really, they want to be able to write something like:

`bar :: Int -> String -> Foo`

and have that expand to

`bar :: forall a. (Constraint a) => Int -> String -> X`.

So they're not worried about higher ranked types (and often intended a rank one type all along!) but they're just trying to find some way to use type synonyms to package up quantification and constraint conveniently into a single alias.

Rather than make the type inference system more complicated, I think it would be better if there was just a convenient syntax to provided the functionality that people desired all along.

Secondly, and less pressing, there's the issue of records of the form `data Foo = Foo {getFoo :: forall a. X}` and that `getFoo` accessor having a higher ranked type rather than being instantiated immediately at each use site. I would tend to think that people always expect the latter -- but it might be too confusing and break too much just to change how such quantified accessors work at this point?

That is to say, I suspect people find the following surprising:

```
ghci> data Foo = Foo {getFoo :: forall a. Maybe a -> Int}
ghci> :t getFoo
getFoo :: Foo -> forall a. Maybe a -> Int

```

And that they would much more expect `getFoo :: forall a. Foo -> Maybe a -> Int`, so perhaps accessors should always be generated that way?

simonpj:
> I cannot forsee all the consequences, but it is at least easy to specify and easy to implement.

One consequence is that visible type application is affected.

gbaz:
I was thinking of a variant of simon's idea -- pardon the janky syntax. The idea would be to not have a flag that makes forall "more magical" but instead to have a syntactic distinction between "floating forall" and normal ones. So what users get is always normal foralls, but also some sugar for "nonlocal type alias action"

`type Foo = forall a b. type forall c d. [...]`.

The idea is that the material to the left of the `type` is "syntactically floating" and moves to the far left of any type the alias appears in. This would float through _everything_ (an improvement on the situation for many cases, since they could write `[Foo]` without getting an error about impredicativity), and be understood as just a fancy rewrite rule.

Then, the material to the right of the `type` would be substituted and handled normally.

Now, aliases of the form `type Foo = forall a b. [...]` would be treated the same, with or without the `FloatingForall` flag, but turning it on would simply enable the extended syntax to allow these fancier aliases to be used as well.

Reusing the `type` keyword is not particularly delightful, but it seems not terribly misleading, and it doesn't require any new keywords.

noughtmare:
If it is just for type synonyms then I think this syntax is more natural:

```haskell
type forall a b. Foo = ...
```
Is it is ever necessary to have floating foralls outside of type synonyms?

goldfirere:
This still doesn't quite add up for me. What if I do this:

```hs
type Endo a = a -> a
type Weird = forall b. type Int -> b -> b

f1 :: Endo Weird
f2 :: forall a. Endo a
f3 = ... f2 @Weird ...
```

I would imagine that `f1` would get the type `forall b2. (forall b1. Int -> b1 -> b1) -> Int -> b2 -> b2`. (You said that `forall`s float out of *everything*, but you didn't mean past arrows from the left, did you?) But `f2 @Weird` couldn't get that type: it would get `(forall b. Int -> b -> b) -> forall b. Int -> b -> b`, because there's no real opportunity to float in that case.

I have to say this plan reminds me of 
image: all corner cases.

serras:
It might be that I'm too attached to this idea of simplifying the subsumption, but I would like to offer two counter-points for re-introducing deep subsumption:

1. I don't this there's a clear story regarding type application (and maybe in the future, abstraction) when interacting with a subsumption which "looks through" foralls. As an extreme example, if type abstraction was available, should the following compile?

    ```haskell
    f :: forall a. a -> forall b. b -> a
    f @a @b x y = x
    ```

2. I don't think we're going to alleviate most of the pain in the community by re-introducing deep subsumption. As I see it, the problem is that *right now* code which used to compile in GHC 8.x does not compile in GHC 9.x. Fixes are already been deployed, and by the time a potential GHC 9.4 or 9.6 reaches Stackage, maybe every package had already been updated.

PR #288, opened 2019-11-03
Proposal 54(36) and 190(49) have been implemented

takenobu-hs:
See the following:

Proposal 54(36)
* https://gitlab.haskell.org/ghc/ghc/merge_requests/853

Proposal 190(49)
* https://gitlab.haskell.org/ghc/ghc/merge_requests/1438


nomeata:
Thanks! This is very helpful.

PR #289, opened 2019-11-06
The variable which isn't brought into scope by the pattern that

cgibbard:
actually matters for the discussion which follows is b.


nomeata:
Quite right, thanks!

PR #281, opened 2019-10-10
Visible `forall` in types of terms, and types in terms

int-index:
The [proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0281-visible-forall.rst) has been accepted; the following discussion is mostly of historic interest.

----------------------------------------------

We propose to allow visible irrelevant dependent quantification, written as `forall x ->`, in types of terms.

[Rendered](https://github.com/int-index/ghc-proposals/blob/visible-forall/proposals/0000-visible-forall.rst)

PR #242, opened 2019-06-16
Unsaturated type families (under review)

kcsongor:
This is a proposal to allow partial application of type families, as
described in the paper [Higher-Order Type-level Programming in Haskell](https://www.microsoft.com/en-us/research/publication/higher-order-type-level-programming-in-haskell)

[Rendered](https://github.com/kcsongor/ghc-proposals/blob/unsaturated-type-families/proposals/0000-unsaturated-type-families.rst)

PR #291, opened 2019-11-08
Simplify scoping for type applications in patterns

Ericson2314:
This simplifies the scoping of type applications in patterns so they:

 - Match regular variables in patterns

 - Keep us on the path to DH where `@<foo>` vs `<foo>` is minimally different

 - Allows a subset of Haskell where there's a strict separation between
   identifier binding and use with no loss of expressive power

It is an amendment to #126 that grew out of my comment https://github.com/ghc-proposals/ghc-proposals/pull/126#issuecomment-550515751

[Rendered](https://github.com/Ericson2314/ghc-proposals/blob/simpler-scope-for-type-application-in-patterns/proposals/0126-type-applications-in-patterns.rst)

### DIFF
--- a/proposals/0126-type-applications-in-patterns.rst
+++ b/proposals/0126-type-applications-in-patterns.rst
@@ -181,7 +181,7 @@ so it always has a variable on its left, whereas a type application is always he
 
 `Proposal #238`_ would introduce an ``-XTypeAbstractions``, which completes this in adding “top level” ``@ty`` patterns, while this proposal adds nested ``@ty`` patterns.
 It also restricts ``-XScopeTypeVariables`` so that ``forall`` no longer binds type variables over the term with the ``forall`` type.
-`Proposal 285`_ would introduce a ``-XNoImplcitForAll``, which would prohibit the implicit binding of free variables in regular and pattern signatures.
+`Proposal 285`_ would introduce a ``-XNoImplicitForAll``, which would prohibit the implicit binding of free variables in regular and pattern signatures.
 The combination of these proposals is rather nice; we can “write System F”, with no superfluous free var binding sugar or missing explicit syntactic constructs.
 This should be good for teaching, and also get us closer to a fully faithful resugaring of GHC’s core.
 
--- a/proposals/0126-type-applications-in-patterns.rst
+++ b/proposals/0126-type-applications-in-patterns.rst
@@ -18,6 +18,9 @@ We have ``TypeApplications`` in terms. This proposal brings them to patterns in 
 Motivation
 ------------
 
+Symmetry between terms and patterns
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
 ``TypeApplications`` are a convenient and natural way to specifying types of polymorphic functions. Consider::
 
  data Foo a where MkFoo :: forall a. a -> Foo a
@@ -35,14 +38,49 @@ but not::
 
     foo (MkFoo @ty x) = …
 
-This proposal fills this gap: It allows type applications in patterns, and specifies them to behave “just like type signatures”.
+This proposal fills this gap.
 
-The intention of the following specification is that the following holds: For a constructor with type like ``C :: forall a. a -> …`` the meaning of ``C @ty x`` should coincide with the existing form ``C (x :: ty)``.
+Simpler scoping rules
+~~~~~~~~~~~~~~~~~~~~~
+
+The scoping rules from ``ScopedTypeVariables`` are quite complicated. With plain::
+
+    foo :: Foo -> SomethingElse ty …
+    foo (MkFoo (x :: ty)) = …
+
+``ty`` is freshly bound. But with::
+
+    foo :: forall ty. Foo -> SomethingElse ty …
+    foo (MkFoo (x :: ty)) = …
+
+``ty`` on the second line is now captured by the implicit binding on the first line.
+I made the first line’s implicit binding explicit with a ``forall``::
+But, ``ScopedTypeVariables`` gives me nothing analogous to ``forall`` to write an explicit, uncapturable binding in patterns.
+We cannot change ``ScopedTypeVariables`` without breaking lots of code, but we can give ``@`` that role.::
+
+    foo :: forall ty. Foo -> SomethingElse ty -> …
+    foo (MkFoo @ty x) = …
+
+Here, the ``ty`` in ``@ty`` is always a fresh binding, in this case shadowing the ``ty`` from above.
+
+Future collapsing of types and terms
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+With Dependent Haskell, ``@a`` vs plain ``a`` is not supposed to distinguish between type and terms, but rather the visibility of the quantifier.
+Simply put, ``@a`` is supposed to mean an optional ``a``, no more, no less.
+Baring namespace issues that are being sorted out other proposals, this is ready to be true for terms.
+I think this is especially salient in the case of the relevent, dependent quantification (``foreach x.``).
+When bind types today (say with pattern signatures), which are erased, we cannot do much computation and mainly are "guiding" inference.
+It can feel more like annotations than the program proper, and thus separate scoping rules don't feel so off.
+When pattern matching a constructor with such a field, we really pattern matching a runtime value we can then compute with perfectly normally.
+
+With this proposal, it we are ready to say ``@a`` is just an optional ``a`` in patterns too.
+Baring the affected namespace, the scoping for ``@a`` and ``a`` is exactly the same.
 
 Proposed Change Specification
 -----------------------------
 
-When both ``TypeApplications`` and ``ScopedtypeVariables`` are enabled, then type application syntax is
+When ``TypeApplications`` is enabled, then type application syntax is
 available in constructor patterns. Concretely, the grammar for constructor pattern is extended from::
 
   pat   → gcon apat1 … apatk --  (arity gcon  =  k, k ≥ 1)
@@ -54,11 +92,18 @@ to::
         …
   tyapp → @ atype
 
-A pattern ``C @a``, where ``C`` is a data constructor and ``a`` is a type variable that is not yet in scope, matches if ``C`` matches. It brings ``a`` into scope so that ``a`` stands for the corresponding type that was passed to ``C`` upon construction.
+However, without ``ScopedTypeVariables``, no variables can be bound in the type, only closed patterns (which include underscores) are permitted.
 
-More complicated cases such as ``C @ty``, where ``ty`` is not just a plain type variable, are handled like `type signatures in patterns <https://downloads.haskell.org/~ghc/8.4.3/docs/html/users_guide/glasgow_exts.html#pattern-type-sigs>`_. A full and formal description of the scoping and typing rules for this feature can be found in `“Type variable in pattern” by Richard Eisenberg, Joachim Breitner and Simon Peyton Jones <https://arxiv.org/abs/1806.03476>`_.
+A pattern ``C @a``, where ``C`` is a data constructor and ``a`` is a type variable that is not yet in scope, matches if ``C`` matches.
+It binds a fresh ``a`` so that ``a`` stands for the corresponding type that was passed to ``C`` upon construction.
 
-An underscore in a type signature or type application in a pattern is allowed, does not bind any variables, and is not treated as a partial type signature (i.e. does not cause warnings with ``-Wpartial-type-signatures``).
+Just like with regular (term) identifiers, it is an error to bind the same type variable twice in a sequence of patterns.
+
+More complicated cases such as ``C @ty``, where ``ty`` is not just a plain type variable, are handled like `type signatures in patterns <https://downloads.haskell.org/~ghc/8.4.3/docs/html/users_guide/glasgow_exts.html#pattern-type-sigs>`_. A full and formal description of the typing rules for this feature can be found in `“Type variable in pattern” by Richard Eisenberg, Joachim Breitner and Simon Peyton Jones <https://arxiv.org/abs/1806.03476>`_.
+However, note the scoping rules are simplified from what is in the paper.
+
+An underscore in a type application is not treated as a partial type signature (i.e. does not cause warnings with ``-Wpartial-type-signatures``).
+A pattern signature however is treated as a partial type signature, as that is a signature first, which only contains bindings by analogy to the normal implicit free variable binding rule that applies to regular signatures.
 
 Examples
 --------
@@ -85,16 +130,20 @@ A more complex example is this (also inspired by `#15050 <https://gitlab.haskell
       MkT3 :: forall a b.            T a
       MkT4 :: forall a b. b ~ Int => T a
       MkT5 :: forall a b c. b ~ c => T a
+      MkT6 :: forall a b. ClassWithFunDep a b => T a
+
+    class ClassWithFunDep a b | a -> b
 
     foo :: T (Int, Int) -> ()
     foo (MkT1 @(Int,Int))  = ()
     foo (MkT2 @x)          = (() :: x ~ Int => ())
-    foo (MkT3 @_ @x)       = (() :: x ~ x => ()) -- nb: x is in scope
+    foo (MkT3 @_ @x)       = (()) -- x is unconstrained
     foo (MkT4 @_ @x)       = (() :: x ~ Int => ())
     foo (MkT4 @_ @Int)     = ()
-    foo (MkT5 @_ @x @x)    = (() :: x ~ x => ())
+    foo (MkT5 @_ @x @y)    = (() :: x ~ y => ())
+    foo (MkT6 @x @y)       = (() :: ClassWithFunDep x y => ()) -- hard to bind b otherwise!
 
-All of these equations type-check (just like they would if added value arguments of type ``a``, ``b``,... to the constructors and turned the type applications into type signatures).
+All of these equations type-check.
 
 Note that the ``@_`` are not treated like partial type signatures.
 
@@ -103,50 +152,38 @@ Note that it is usually a type error to supply a non-tyvar type, or an in-scope 
 Scoping
 ~~~~~~~
 
-The scoping works just like with ``ScopedTypeVariables``. Just for reference, here are some examples of how that feature works now::
+The scoping works just like regular variables, but in the type namespace::
 
  f1 :: forall a b. ([a], b) -> Int
- f1 (x :: [c], y) = ...
-
-brings ``c`` into scope, together with ``a`` and ``b``, which are already in scope. The type variables ``c`` and ``a`` refer to the same type.
-
-But the pattern in::
-
+ f1 ((,) @[c] x y) = ...           -- c ~ a
  f2 :: forall a b. ([a], b) -> Int
- f2 (x :: [b], y) = ...
-
-does not bring ``b`` into scope; here ``b`` refers to the ``b`` from the type signature. This leads to an type error, because in general ``a`` and ``b`` do not refer to the same types.
-
-And the pattern in::
-
+ f2 ((,) @[b] x y) = ...           -- c ~ b
  f3 :: forall a b. ([a], b) -> Int
- f3 (x :: [c], y :: c) = ...
-
-brings one ``c`` into scope; the second occurence in the pattern does not shadow the first one, but rather refers to the same type. This example, too,  leads to a type error because ``c`` needs to be equal to both ``a`` and ``b``.
-
-The same rules apply for type applications, so the above examples could be re-written as follows, with identical behaviour::
-
- f1 :: forall a b. ([a], b) -> Int
- f1 ((,) @[c] x y) = ...
- f2 :: forall a b. ([a], b) -> Int
- f2 ((,) @[b] x y) = ...
+ f3 ((,) @[c] @c x y) = ...        -- error, c bound twice
  f3 :: forall a b. ([a], b) -> Int
- f3 ((,) @[c] @c x y) = ...
+ f3 ((,) @[c] c d) = ...           -- c ~ a
 
 Effect and Interactions
 -----------------------
-We answer the question “what should ``@ty`` mean in patterns” based on an existing feature (“what should ``_ :: ty`` mean in patterns”. This fills an obvious hole in the syntax in a way that is consistent with existing features. In particular analogy between type applications and type signatures that we currently have in terms will now hold the same way in patterns.
+We answer the question “what should ``@ty`` mean in patterns” based on the analogous features (“what should ``ty`` mean in patterns”, and “what does ``@ty`` means in terms”.
+This “completes the square”, so we have visible and invisible constructor application and decomposition.
 
 Furthermore, type application arguments to ``C`` refer to the corresponding parameters in both terms and types, irregardless of whether they are considered universal or existential variables (this is not the case for alternative proposals, as explained below under “Alternatives”).
 
-This proposals allows the binding of existential type variables of constructors, and hence subsumes `Proposal #96 <https://github.com/ghc-proposals/ghc-proposals/pull/96>`_.
+This proposals allows the binding of existential type variables of constructors, and hence subsumes `Proposal #96`_.
 
 There is almost a syntactic ambiguity with as-patterns, but in fact there is not: The grammar of as-pattern is::
 
-  apat 	→ 	var [ @ apat] 	    (as pattern)
+  apat  →   var [ @ apat]       (as pattern)
         …
 
 so it always has a variable on its left, whereas a type application is always headed by a constructor.
+
+`Proposal #238`_ would introduce an ``-XTypeAbstractions``, which completes this in adding “top level” ``@ty`` patterns, while this proposal adds nested ``@ty`` patterns.
+It also restricts ``-XScopeTypeVariables`` so that ``forall`` no longer binds type variables over the term with the ``forall`` type.
+`Proposal 285`_ would introduce a ``-XNoImplcitForAll``, which would prohibit the implicit binding of free variables in regular and pattern signatures.
+The combination of these proposals is rather nice; we can “write System F”, with no superfluous free var binding sugar or missing explicit syntactic constructs.
+This should be good for teaching, and also get us closer to a fully faithful resugaring of GHC’s core.
 
 Costs and Drawbacks
 -------------------
@@ -154,9 +191,9 @@ Given that the specification is inspired by an existing feature, I expect the im
 
 For users who want this mainly to instantiate existential variables may find that they have to write ``C @_ @x`` to
 go past the universial variables, which is mildly inconvenient. It may be fixed in some cases by changing the order
-of the type variables of ``C``. This is unavoidable if we want to preserve the symmetry between terms and types, though. A mitigation for this is offerend in `proposal #99 (explicit specificity) <https://github.com/ghc-proposals/ghc-proposals/pull/99>`_.
+of the type variables of ``C``. This is unavoidable if we want to preserve the symmetry between terms and types, though. A mitigation for this is offerend in `proposal #99`_.
 
-A possible future proposal that extends as-patterns to allow patterns on both sides of the `@` would now introduce ambiguities, e.g. in `Nothing @ a`, and will require disambiguation. This disambiguation could be
+A possible future proposal that extends as-patterns to allow patterns on both sides of the ``@`` would now introduce ambiguities, e.g. in ``Nothing @ a``, and will require disambiguation. This disambiguation could be
 
 * extra parenthesis: ``(Nothing) @ a`` is an as-pattern, vs. ``Nothing @ a`` is a type application.
 * using a helper pattern synonym::
@@ -174,10 +211,28 @@ Open Questions
 
 Alternatives
 ------------
-`Proposal #96 <https://github.com/ghc-proposals/ghc-proposals/pull/96>`_ proposes a variant where ``@x`` may only mention type variables and only existential type variables may be
+`Proposal #96`_ proposes a variant where ``@x`` may only mention type variables and only existential type variables may be
 bound this way. See there for an in depth discussion; a summary of the main criticism that the proposal at hand tries
 to fixes preserving the symmetry between type applications in terms and patters, and preserving the analogy between
 type applications and type signatures, and also in Section 6.1 of `the paper <https://arxiv.org/abs/1806.03476>`_. Furthermore, it does not introduce new concepts (e.g. the distinction between
 existential and universal parameters) to the Haskell programmer.
 
-The existing restriction of ``ScopedTypeVariabes`` that type variables in pattern signatures may only be bound to type variables, and not types, carries over to type variables in type applications. One could discuss lifting this restriction, but this question is completely orthotogonal to the proposal at hand, and should be discussed elsewhere (e.g. in `Proposal #128 <https://github.com/ghc-proposals/ghc-proposals/pull/128>`_ and `ticket #15050 <https://gitlab.haskell.org/ghc/ghc/issues/15050#note_152286>`_).
+The existing restriction of ``ScopedTypeVariables`` that type variables in pattern signatures may only be bound to type variables, and not types, carries over to type variables in type applications. One could discuss lifting this restriction, but this question is completely orthotogonal to the proposal at hand, and should be discussed elsewhere (e.g. in `Proposal #128 <https://github.com/ghc-proposals/ghc-proposals/pull/128>`_ and `ticket #15050 <https://gitlab.haskell.org/ghc/ghc/issues/15050#note_152286>`_).
+
+We could allow ``-XTypeAbstractions`` to bind variables even without ``-XScopeTypeVariables``.
+Unlike turning ``-XScopeTypeVariables`` on by default, this would not break any Haskell 98 programs as there are no type application patterns in Haskell 98.
+This allows strictly more programs, so it seems fine to consider later.
+
+A previous version of this proposal had the scoping rules work just like that for pattern signatures.
+This had the benefit of allowing so-called “non-linear patterns”, where the variables must be proven equal for GHC to accept the program, and also prohibited potentially-confusing shadowing.
+The downside is the scoping rules are more complex, and widen the gulf that we claim we want to shrink between visible and invisible quantifiers.
+They also break the “write System F” result of this + ``-XTypeAbstractions`` + ``-XNoImplcitForAll``.
+
+I will concede with respect to the old version of the proposal that permitting shadowing allows strictly more programs, giving us less room in the future to maneuver.
+If we disallow non-linear patterns and shadowing, we keep the door open to both the old and current version of the proposal.
+
+As final alternative based on the old version of the proposal, in the further future when we have equality constraints, we could create a new warning (or extension) that requires all shadowing and non-linear patterns of any sort to obligate that the two variables be constrained to be equal.
+This would effectively apply the old version of the proposal to all bindings, also restoring the symmetry between visible and invisible quantifiers.
+
+.. _`Proposal #96`: https://github.com/ghc-proposals/ghc-proposals/pull/96
+.. _`Proposal #238`: https://github.com/ghc-proposals/ghc-proposals/pull/238
--- a/proposals/0126-type-applications-in-patterns.rst
+++ b/proposals/0126-type-applications-in-patterns.rst
@@ -143,7 +143,7 @@ A more complex example is this (also inspired by `#15050 <https://gitlab.haskell
     foo (MkT5 @_ @x @y)    = (() :: x ~ y => ())
     foo (MkT6 @x @y)       = (() :: ClassWithFunDep x y => ()) -- hard to bind b otherwise!
 
-All of these equations type-check.
+All of these equations type-check (just like they would if added value arguments of type ``a``, ``b``,... to the constructors and turned the type applications into type signatures).
 
 Note that the ``@_`` are not treated like partial type signatures.
 


aspiwack:
I'll admit that it is a bit hard to get a good feel about what's happening from the diff.

Could you give a summary of the changes in this thread: what is the current behaviour (or behaviours) which you are amending, and what is the new behaviour (resp. behaviours), preferably with example? I think it would be tremendously helpful.

proposals/0126-type-applications-in-patterns.rst
```diff
@@ -85,16 +130,20 @@ A more complex example is this (also inspired by `#15050 <https://gitlab.haskell
...
-    foo (MkT3 @_ @x)       = (() :: x ~ x => ()) -- nb: x is in scope
+    foo (MkT3 @_ @x)       = (()) -- x is unconstrained
     foo (MkT4 @_ @x)       = (() :: x ~ Int => ())
     foo (MkT4 @_ @Int)     = ()
-    foo (MkT5 @_ @x @x)    = (() :: x ~ x => ())
+    foo (MkT5 @_ @x @y)    = (() :: x ~ y => ())
+    foo (MkT6 @x @y)       = (() :: ClassWithFunDep x y => ()) -- hard to bind b otherwise!
 
-All of these equations type-check (just like they would if added value arguments of type ``a``, ``b``,... to the constructors and turned the type applications into type signatures).
+All of these equations type-check.
```
nomeata: Did you remove this sentence because itis no longer true?

proposals/0126-type-applications-in-patterns.rst
```diff
@@ -103,60 +152,48 @@ Note that it is usually a type error to supply a non-tyvar type, or an in-scope
...
 
-  apat 	→ 	var [ @ apat] 	    (as pattern)
+  apat  →   var [ @ apat]       (as pattern)
         …
 
 so it always has a variable on its left, whereas a type application is always headed by a constructor.
 
+`Proposal #238`_ would introduce an ``-XTypeAbstractions``, which completes this in adding “top level” ``@ty`` patterns, while this proposal adds nested ``@ty`` patterns.
+It also restricts ``-XScopeTypeVariables`` so that ``forall`` no longer binds type variables over the term with the ``forall`` type.
+`Proposal 285`_ would introduce a ``-XNoImplcitForAll``, which would prohibit the implicit binding of free variables in regular and pattern signatures.
```
nomeata:
```suggestion
`Proposal 285`_ would introduce a ``-XNoImplicitForAll``, which would prohibit the implicit binding of free variables in regular and pattern signatures.
```

simonpj:
> Could you give a summary of the changes in this thread: what is the current behaviour (or behaviours) which you are amending, and what is the new behaviour (resp. behaviours), preferably with example?

I believe I have reverse-engineered from the diff what John is proposing.

TL;DR: John wants the type variables in type application in patterns to behave like term variables:

* they shadow
* they cannot be repeated
* they can be alpha-reaned freely

and *not* like the type variable in partial type signatures as [the original proposal](https://github.com/ghc-proposals/ghc-proposals/blob/type-applications-in-patterns/proposals/type-applications-in-patterns.rst) specified.

If I'm right, it would be really helpful to make the proposal cover the points below.

------------------

Consider:
```
f :: forall a.  Maybe [a] -> a
f (Just @[b] y) = y :: b
```
Clearly the `@[b]` binds `b`.  But to what?  Presumably to `a`, since the type of the argument is `Maybe [a]`.

Now consider:
```
f :: forall a.  Maybe [a] -> a
f (Just @[a] y) = y :: a
```
What is the `'a` in the `Just @[a]` pattern?  In the proposal it's the same `a`.   In John's proposed change it's a fresh `a`, shadowing the old one.  That might make a difference:
```
g :: forall a b. Maybe a -> Maybe b -> b
g x (Just @a y) = y :: a
```
This would be rejected by the original proposal, because the `a` in the pattern is the `a` that is in scope from the signature.  John proposes instead that patterns always bind fresh variables, so this is really a fresh `a` that *shadows* the outer `a`.   Here `Just @a` binds a fresh `a` to `b`.   We can freely alpha-rename, so the definition is equivalent to
```
g :: forall a b. Maybe a -> Maybe b -> b
g x (Just @c y) = y :: c
```
where I have replaced `a` by `c`.

I assume that, as in the original proposal, this definitions would be accepted
```
f1 :: forall a. Maybe [a] -> [a]
f1 (Just @b y) = y :: b
```
Here `b` is bound by the pattern `Just @b`, and is bound to the type `[a]`.

But this would be rejected:
```
f2 :: forall a. Maybe a -> blah
f2 (Just @[b] y) = ...
```
because the type of the first argument is `Maybe a` but the pattern tries to make it have type `Maybe [something]`.

John would also reject this:
```
f3 :: Mabye (a,a) -> blah
f3 (Just @(b,b) x) = ...
```
because `b` is bound twice in the same pattern.  The original proposal would have accepted it.

simonpj:
Why did the original proposal use "pattern-type-signature scoping"?  Consider
```
k (x::a) (y::a) = x
```
Today this is accepted, and we get the inferred type `k :: forall a. a -> a -> a`.

Now consider
```
k2 (Just @[a] x) (Just @[a] y) = x
```
Under the existing proposal this too is accepted with inferred type `k2 :: Maybe [a] -> Maybe [a] -> [a]`.  But under John's proposal it is rejected becuase it binds `a` twice in the same LHS.

I think I'm ok with this, but it's worth understanding the tradeoff.

goldfirere:
John suggested this change elsewhere (https://github.com/ghc-proposals/ghc-proposals/pull/126#issuecomment-550515751) and I think I understand the main payload, so I'm responding to that (and Simon's summary at the top of https://github.com/ghc-proposals/ghc-proposals/pull/291#issuecomment-551450307, which matches my intuition), not reading the full diff.

The idea behind binding type variables in patterns was that they would behave identically to type variables in type annotations. That was appealing, because we were building a new feature in terms of an old one -- and almost all uses of the new feature could be desugared into the old one. (Not all, because of ambiguity.) Now, you're suggesting to reframe the new feature (variables bound in `@` bits) to be quite detached from annotations.

To me, the crux of the argument is this: we don't want to treat patterns with a visibility override `@` differently from patterns without `@`. (Note: no as-patterns in sight here.) I think that's quite a compelling design point, and one I hadn't considered previously.

I'm in favor of the idea. If others are too, I can take a look at the proposal text more closely.

simonpj:
> To me, the crux of the argument is this: we don't want to treat patterns with a visibility override @ differently from patterns without @

I have no idea what this means!  WHat is a "visibility override"?  Examples?  (Which again shoud make their way into the proposal.

nomeata:
Is this a adequate summary of the question:

Given
```
foo @c (a :: b) = …
```
should `c` scope/shadow/bind like `a` or like `b`.

simonpj:
> Is this a adequate summary of the question:

Alas no.  Neither the proposal nor the modificaiton allows
```
foo @c x = ...
```
The `@` form is only in constructor patterns, as it stands.

nomeata:
sorry, I simplifed too much. Edited my post.

goldfirere:
Haskell still retains the Hindley-Milner coincidence that types are invisible in terms. That is, types are inferred. We currently have the ability to override this choice in visible type application: by writing a `@`, we take an invisible (type) argument and make it visible. It so happens that all invisible arguments are types and all visible arguments are not types, but that may not always be (#281). (Actually, implicit parameters are invisible, non-type arguments, too.)

We thus have multiple (valid) interpretations of `@`:

1. `@` is a visibility override: it takes an invisible argument and makes it visible. *Inferred* arguments are not available for overriding, while *specified* ones are.

2. `@` is a marker that a type comes next in order to instantiate a specified argument.

I prefer interpretation (1), as it's more general. But interpretation (2) is valid today... though I personally hope for it not to be valid in the future.

With the un-amended proposal, a `@` in a pattern does two things: it's a visibility override (making an invisible argument visible), and it changes scoping behavior. (Incidentally, it also says "a type come next".) This amendment means that it no longer changes scoping, making it conceptually simpler.

nomeata:
I agree it is simpler. But we are losing the ability to refer to types in scope in `@` patterns.

There is a connection with term patterns and pattern synonyms: It is an obvious hole that we don’t have patterns that are parametrized by terms. We can write
```
Prelude> pattern SquareOfTwo <- ((== (2::Int)^2) -> True)
Prelude> :info SquareOfTwo
pattern SquareOfTwo :: Int 	-- Defined at <interactive>:15:1
Prelude> let foo 2 SquareOfTwo = True; foo _ _ = False
```
but not
```
Prelude> pattern SquareOf a <- ((== (a::Int)^2) -> True)
Prelude> :info SquareOf
pattern SquareOfTwo :: Int -> Int -- or something like that
Prelude> let foo a (SquareOfT a) = True; foo _ _ = False
```
and at least one reason why we failed to do so is that in the last equation it us unclearif `a` is a binderor or an occurrence.

So mabye there is a solution for that problem that would
 * work for both terms and types
 * allow us to use John’s proposal here, and still be able to use `@` to specify types in patterns that refer to type variables
 * open the path for term-abstracted pattern synonyms.

(No concrete suggestion, though).

proposals/0126-type-applications-in-patterns.rst
Ericson2314: Thanks!

proposals/0126-type-applications-in-patterns.rst
Ericson2314: Yes, but I was wrong. Those are all fine; will revert.

Ericson2314:
```suggestion
All of these equations type-check (just like they would if added value arguments of type ``a``, ``b``,... to the constructors and turned the type applications into type signatures).
```

Ericson2314:
@nomeata I did think of that a little bit. It's an extra hard problem because for relevant fields someone might expect fallthrough / decidable equality. 

With matchability polymorphism, we could in fact just quantity of things we know how to match on too. Just requiring two variables to be equal according to `Eq` is easy to write more explicitly, so I am not so sympathetic to that, but passing a data constructor which is then matched is harder to desugar.

While this may seem far-fetched / loony, consider if a student manually beta-reduces the parameterized pattern, they would get a pattern that looks fallible. I suppose we already have this problem with bottoms, but it would make it more noticeable.

Ericson2314:
Yes, a key as @goldfirere says is to think past Milner's Coincidence.

If we have:
```haskell
data Foo where
  MkFoo :: foreach (x :: Int). Foo

foo  :: Foo -> Int
foo (MkFoo @x) = x + 1
```
`x` is a real runtime field being pattern matched here. One could say the information flow is "operationally" determined, unlike irrelevant fields which we unify and do other fancy tricks with at compile time beyond good ole' close term evaluation. 

There's also the opposite case where we can do:
```haskell
data Foo t where
  MkFoo :: forall t -> t -> Foo t

foo :: Num t => Foo t -> t -> t
foo (MkFoo t x)  y = x + y + 1
```
`x` looks like a normal field today, but is a another spooky erased thing we only play with at compile time. If I like the proposal as written, might I also want the `t` in the pattern to reference the `t` in the signature rather than shadow?

goldfirere:
I think term abstractions (though very nice) are a red herring here.

Let's imagine briefly that term variables scoped like type variables do. Then we could write this:

```hs
foo :: Maybe Bool -> ()
foo x = case x of
  Just b -> case x of
    Just b -> ()

bar :: Maybe Bool -> Maybe Bool -> ()
bar x y = case (x, y) of
  (Just b1, Just b2) -> case (x, y) of
    (Just b2, Just b1) -> ()
```

If term variables worked like type variables do, then `foo` would be accepted while `bar` would be rejected (because we don't know that `b1` and `b2` are the same). (Of course, we could use `Eq` instances to allow even more, but that's still not quite what happens with types.)

The difference is that scenarios where we want this with terms are very rare, while they're not so rare with types.

If I remember correctly, Coq deals with this issue by requiring a symbol prefix (`@` maybe?) in order to bind a variable in a certain construct; otherwise, the variable name is an occurrence.

I don't have any good ideas, but I do have a bad one, included here in case it sparks someone else's good idea: In view patterns, we have `(expr -> pattern)`. That means that variables to the left of `->` are occurrences while those to the right are binders. So we could imagine

```hs
isSquare :: Int -> Int -> Bool
isSquare (-> a) (SquareOf (a ->)) = True 
isSquare _ _ = False
```

where `SquareOf` is from https://github.com/ghc-proposals/ghc-proposals/pull/291#issuecomment-551760822. The `(-> ...)` declares (redundantly) that something is a pattern. Perhaps we don't allow that. It's like unary `+` in C. But more interesting is `(SquareOf (a ->))`, where we have to denote, at usages of `SquareOf`, that its argument is an *expression*, not a pattern. Mnemonic: the expression is fed "into" the pattern. (Of course, we don't *have* to denote this. But it's a service to our users to.)

The problem with my syntax idea is that it doesn't quite work here, as some type arguments will have a mix of occurrences and binders. It really needs to be at the level of an individual variable.

Hey wait! What about labeling a binder with `\`? We're used to putting binders after `\`. It's still kinda unsatisfactory, though, because we don't want to label every (term!) binder with `\`. We really want to label the occurrences. I'm all out of ideas for now.

Ericson2314:
@goldfirere I don't think we necessarily need to come up with new syntax at this point. The alternative of banning shadowing and multiple bindings within patterns keeps both versions of proposal a non-breaking change, and two different ways to keep `@` just a visibility override with Dependent Haskell. On top of either of those two options, we can add decide to add bind/reference overrides. I'm more concerned with not backing ourselves into a corner we'll come to regret, than solving *all* the problems :).

aspiwack:
> If I remember correctly, Coq deals with this issue by requiring a symbol prefix (`@` maybe?) in order to bind a variable in a certain construct; otherwise, the variable name is an occurrence.

“bind a variable in a certain construct” is a rather vague turn of phrase, so I'm not sure what you are referring to. But I don't believe Coq has this sort of facility. I believe you may be thinking of Agda, where, if I remember correctly, in:

```agda
f x .x = …
```

The first `x` (without the `.`) is a binding occurrence, whereas the second one is just a term (I believe an arbitrary term can be used after the `.`, but I haven't Agda'd for many years, so don't take my word for it), and incurs an equality constraint.

Ericson2314:
From https://mail.haskell.org/pipermail/ghc-steering-committee/2020-January/001410.html

> I'm still not convinced about the fact that there are two extensions for
this. Which seems more motivated by technical details of the
implementation, rather than by semantics. But it is not a hill that I'm
willing to die on… so if I'm alone in this, I'm not going to argue further.

As I put in the proposal, I only lean slightly towards 2 extensions :). I can't speak for the other committee members though.

So currently today, when the variable is free, both pattern sigs and implicit foralls bind an identifier, and when the identifier it is not free, it is captured. So in terms of name resolution they do work the same. 

However in terms of the semantics, the type variable is a fresh quantification / rigid var in the latter, but free to unify with things / non-rigid with the former. This is a true semantic difference, not just implementation difference.

And yet... with the both `-XNo...` flags turned on, type variables in regular type sigs (that aren't explicit binds) and regular sigs are both captured (or a scoping violation) (same name resolution) and both take on the type value of the binding that captures them (same semantics).

So that means if we had `-XNo` flag, while it changes two distinct semantics, the "destination" is the same semantics. But that is the destination in the `-XNo` direction. If you ignore what's on by default and just worry about the "positive" (`-X` not `-XNo`) direction, it's one flag starting with a uniform simple semantics and enabling two distinct semantics.

All this is to say the situation is somewhat muddled and it depends whether you care more about the `-XNo` -> `-X` transition or default -> manual override transition.

cgibbard:
The alternative of disallowing shadowed variables for now seems like a nice way to preserve the option of taking any of the proposed behaviours later.

Personally, I do think it's nicer to be able to think of the @ as allowing access to the next otherwise-invisible parameter vs. specifically treating it as a thing about types. (A cute thing I just noticed is that it's kind of like an as-pattern where the left side is invisible! I'm pretty sure that wasn't intentional, but if it was, bravo.)

So on the face of it, I'm leaning in favour of this proposal.

But at the same time I have to admit that a lot of the usual downsides to allowing repeated variables in patterns don't quite apply when we're not going to be rejecting the patterns at runtime on the basis of whether the things they match are equal or not (and there's no bottoms to worry about).

Despite the fact that I'm the one who has lately been tasked with getting enough of a grasp on how the typechecker works to implement this stuff, I do question a little bit what its overall purpose is. We already kind of had two ways of binding those type variables, and we're adding another, which feels slightly excessive to me, particularly when we have no intention of rejecting a pattern at runtime on the basis that a type doesn't match. Making the language of patterns more closely mimic the language of terms seems like a good thing, so I suppose that's a decent enough reason. Perhaps seeing the use-cases that people have in mind for the feature would help to distinguish which option is better though.

simonpj:
Cale has been implementing it (see [ticket 11350](https://gitlab.haskell.org/ghc/ghc/-/issues/11350) and [MR 2464](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/2464)).    It’s nearly done.
But to complete it we need to resolve the issue of this proposal, which is a diff from the accepted [Type applications in patterns](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0126-type-applications-in-patterns.rst).

Let me summarise how I understand it.
Consider first the existing pattern-signature mechanism (not the new feature):
```
data T a where
MkT :: forall a b. a -> b -> T a

f1 :: forall p. T [p] -> blah
f1 (MkT (x :: a) pi) = blah

f2 :: forall p. T [p] -> blah
f2 (MkT (x :: p) pi) = blah
```
In f1, the pattern (x :: a) brings ‘a’ into scope, binding it to the type [p].  But in f2, since p is already in scope, the pattern (x :: p) does not bring anything new into scope.  Instead it requires that x have type p, but actually it has type [p], so f2 is rejected.
Notice that a pattern signature brings a new variable into scope only if it isn’t already in scope.  Notice how this differs from the treatment of term variables; the ‘pi’ in the pattern brings ‘pi’ into scope unconditionally, shadowing the existing ‘pi’ (from the Prelude).

OK, now let’s look at the new feature. Consider
```
g1 :: forall p. T [p] -> blah
g1 (MkT @a x y) = blah

g2 :: forall p. T [p] -> blah
g2 (MkT @p x pi) = blah
```
Question: should the variables free in these type arguments be treated like pattern signatures, or like term variables?
* (A) **Like pattern signatures**.   In this design, in g1, ‘a’ is not in scope, so this brings it into scope, bound to [p].   But in g2, ‘p’ is in scope, so this is rejected (because MkT is instantiated at [p] not p.
* (B)  **Like term variables**.  In this design, all the variables free in type patterns are fresh, and brought into scope. In g2,  a new ‘p’ is brought into scope, shadowing the existing ‘p’; indeed the new ‘p’ is bound to [old_p].

The original paper, and hence the accepted proposal, adopts (A).   But this proposal advocates (B).  

This is not a terribly big deal, but it would be good to settle it.

simonpj:
I'm veering towards "reject"; that is, stick to (A).

* (A) is more expressive. You can always achieve (B) manually by renaming.

* It seems funny to have two different rules for the two different places a type can occur in a pattern. Eg. `Just @[a] (x :: [b])`.  Those two types `[a]` and `[b]` would be treated uniformly by (A) but differently by (B).

* (A) allows you to say very natural things like
  ```
  f :: forall a. Maybe [a] -> Maybe [a] -> blah
  f (Just @[a] x) (Just @[a] y) = blah
  ```
  where all the `a`s mean the same thing.  Under (B) this would actually be rejected because (just like term variables) you can't bind the same variable twice in a singe LHS.  Nor can you write
  ```
  f :: Maybe (a,a) -> blah
  f (Just @(b,b) x) = blah
  ```
  because again `b` is bound twice.  You'd have to say `Just @(b,c)` even though both `b` and `c` are bound to the same type `a`.

nomeata:
I concur with Simon. I agree it’s not pretty that this choice is not in all cases intuitive, but I strongly want to keep the strict symmetry with pattern signatures, and the ability to annotate patterns (via type application or signatures) with types that contain type variables  that are in scope.

gridaphobe:
I also agree with Simon. In

```
f :: forall a. Maybe [a] -> Maybe [a] -> blah
f (Just @[a] x) (Just @[a] y) = blah
```

the `@[a]` annotations should be read as typing constraints that refer to the same (perhaps implicitly bound) type variable `a`.

The correct place to actually bind `a` in the definition would be as the first parameter, matching `f`s signature.

```
f :: forall a. Maybe [a] -> Maybe [a] -> blah
f @a (Just @[a] x) (Just @[a] y) = blah
```

It seems to me that most of the confusion here is due to the fact that we have implicit binders for type variables. If we required people to explicitly bind tyvars (and maybe had different syntax for binding vs applying) I think the behavior would be very clear.

Ericson2314:
I'm really worried that between this and the discussion in https://github.com/ghc-proposals/ghc-proposals/pull/281, where on track to get a Dependent Haskell that has all the new complexity, but none of the new simplicity. CC @int-index.

Going back to to @goldfirere's thesis, `@`, was *just* to effect visibility, and nothing else. Now we have the namespaces difference to also work out, but if we add this difference too, we'll be further away from that goal. I think all these arbitrary symmetry-breaks will really make future Haskell harder to each.

@gridaphobe 

> It seems to me that most of the confusion here is due to the fact that we have implicit binders for type variables.

Under the `@`-is just about visibility interpretation, and given the existing behavior of bindings, these *are* explicit binders. And with (not yet implemented) `-XNoPatternSignatureBinds`, pattern signatures are explicitly uses. The roles of each are nicely separated.

What was originally planning on doing before returning to this thread was updating the motivation since https://github.com/ghc-proposals/ghc-proposals/pull/285 to make it easier to follow. I'll do that now.

Ericson2314:
Also, the thing we actually implemented isn't this change, but a conservative middle ground where shadowing is outright prohibited, so we don't have to pick any semantics for it. Given the uncertainty about the overall direction of dependent Haskell the discussion of #281 has brought up, maybe it is prudent to stick with that conservative option and make both (A) and (B) forwards-compatible changes we can do in the future.

I do think we need to have a higher level discussion on exactly what we are looking for in Dependent Haskell, because I'm not sure the current divergence in philosophies is fully grasped, and I'm wary of incrementally litigating this stuff proposal-by-proposal and still combining up with a coherent plan.

gridaphobe:
> Under the `@`-is just about visibility interpretation, and given the existing behavior of bindings, these _are_ explicit binders.

I disagree, because they are in the wrong position to be explicit binders. Type variables are bound (and applied) separately from term variables, and the scoping rules dictate that type variables must come before any types that depend on them. If you wanted to write the definition of `f` in its most explicit form, it would be 

```
f = /\ @a -> \ (Just @[a] x) -> \ (Just @[a] y) -> blah
```

which makes it clear (to me at least) that the `@[a]` are constraints like `Just`, not binders like `x` and `y`.

The only place where an explicit type binder would make sense as part of a term-level pattern is when you're binding an existential, because there's nowhere else to bind it.

> I do think we need to have a higher level discussion on exactly what we are looking for in Dependent Haskell, because I'm not sure the current divergence in philosophies is fully grasped by all parties, and I'm wary of incrementally litigating this stuff proposal-by-proposal and still combining up with a coherent plan.

That sounds like a good idea. I don't think the committee has ever expressed a formal opinion on Dependent Haskell as a direction for the language.

Ericson2314:
@gridaphobe 

So I think the "naturality" hinges on the universal vs existential distinction. If we have
```haskell
data SomeNum where
  MkSomeNum :: forall a. Num a => a -> SomeNum
```
and
```haskell
double = \ (MkSomeNum @a (x :: a)) -> MkSomeNum @a (2 * x :: a)
```
it's at least as natural that `@a` work as a binder in `double`.

For your example, I admit using bound identifiers as patterns is useful, but I would rather do that with an explicit syntax.

```haskell
f = \ @a -> \ (Just @[= a] x) -> \ (Just @[a] y) -> blah
```

That means we could even use it a the term level:
```rust
FFI_TAG_XX = ...
FFI_TAG_YY = ...

f = \case
   {-# = FFI_CONST_XX, ... #-} -> ...
   {-# = FFI_CONST_YY -> ... #-} -> ...
```
N.B. my example is inspired by Rust which always allows constants in patterns at the term level https://doc.rust-lang.org/reference/patterns.html#path-patterns. This is useful but also leads to confusing errors when one shadows by mistake. The explicit syntax should help avoid the accidental uses.

> That sounds like a good idea. I don't think the committee has ever expressed a formal opinion on Dependent Haskell as a direction for the language.

Glad to be in agreement on that :)

gridaphobe:
> So I think the "naturality" hinges on the universal vs existential distinction. If we have
> 
> ```haskell
> data SomeNum where
>   MkSomeNum :: forall a. Num a => a -> SomeNum
> ```
> 
> and
> 
> ```haskell
> double = \ (MkSomeNum @a (x :: a)) -> MkSomeNum @a (2 * x :: a)
> ```
> 
> it's at least as natural that `@a` work as a binder in `double`.

Exactly! But that's because we're following the structure of the types to determine where binders can occur, which also suggests that (explicit) universal binders should be separate.

This discussion also made me think about having a separate syntax for constraining patterns as opposed to binding patterns, like you said it would be handy for values too. I worry a bit that introducing more syntax will make things too noisy, but I'm open to suggestions.

nomeata:
> This discussion also made me think about having a separate syntax for constraining patterns as opposed to binding patterns, like you said it would be handy for values too. I worry a bit that introducing more syntax will make things too noisy, but I'm open to suggestions.

I’d expect such a syntaxto apply to single type variable in a pattern, e.g. in `foo Just @(a, b) =`, the only one of the two variable might be a binding, and the other an occurrence. 

I definitely recall that we discussed that when we wrote the paper, but I don't remember which options we considered, and why we discarded them back then…

simonpj:
> Also, the thing we actually implemented isn't this change, but a conservative middle ground where shadowing is outright prohibited, so we don't have to pick any semantics for it. 

I think that is a reasonable proposal.  So now now we have
* (A) Like pattern signatures.  The accepted proposal
* (B) Like term variables.   This proposal.
* (C) No shadowing (for type *arguments* in patterns; type *signatures* in patterns are unaffected)

I don't mind much myself, but as it stands (A) is the accepted proposal. Perhaps the committee will prefer (C).  If so, fine; let's see.

NB: (C) does not allow some rather natural things like
```
f :: forall a. Maybe a -> blah
f (Just @a x) = ...
```
And
```
g1,g2 :: Maybe a -> Maybe a -> blah
g1 (Just @a x) (Just (y::a)) = ...   -- Accepted
g2 (Just (x::a)) (Just @a y) = ...  -- Rejected (shadowing)
```

yav:
I think (A) is the right choice here---it is simple and consistent with how the language already works.  It also ensures that all types in patterns handle variables in the same way.  For example, it would be quite surprising if these two worked differently for some reason:

```
h1 (Nothing :: Maybe [p]) = ...
h2 (Nothing @[p]) = ...
```

goldfirere:
For me, this sentence expresses my thoughts exactly:

@Ericson2314 

> I'm really worried that between this and the discussion in #281, we're on track to get a Dependent Haskell that has all the new complexity, but none of the new simplicity.

One of the complaints I regularly hear at conferences, etc., about Haskell is that it's too complicated -- especially once we get into type-level manipulations. My conversation partner then always goes on to say how Agda and Idris are just so much simpler. If we continue to treat types and terms differently in new constructs, we will be propagating that complexity, requiring our users to remember what syntactic bit has what property.

I agree that (A) fits better with today's Haskell. And I agree that (A) is more expressive. But I feel very strongly that we will come to regret the decision later, once we have more dependent types, and less of a difference between terms and types. 

Here's another thought experiment. It won't be long until someone proposes that we allow e.g.

```hs
data T where
  MkT :: forall a -> F a -> T
```

Note that we have here a *visible* `forall`, of #81 and #281 fame. What should be the binding semantics of a pattern-match here? If I write `f (MkT x y)`, it would be very silly of us to have different binding semantics for `x` than for `y`. You might say that `a` is an existential and must be instantiated with a type variable in a pattern. But now what about

```hs
data T2 a where
  MkT2 :: forall a -> F a -> T a
```

Now `a` is universal. Should it be treated differently? One of the main observations in the *Type Variables in Patterns* paper is that we needn't distinguish between universal and existential. (And, in the presence of equality constraints, doing so reliably is hard.) So I don't want `MkT2` to be different from `MkT` in this scenario.

Here's another example which may help clarify this point:

```hs
data List a where
  Mk :: forall (n :: Nat) -> Vec n a -> List a
```

`List` packs a linked list with its length, and the type system works to make sure that the length is accurate. Is `n` a type? Hardly -- it's a natural number. Is it dependent? Surely. Is it relevant? (That is, can we write `len (Mk n _) = n`?) Well, it can't be today, but there is a [large crowd](https://github.com/ghc-proposals/ghc-proposals/pull/378) of people who want that tomorrow. So: what should the binding semantics of the first argument of `Mk` be? Any semantics that differs from the binding semantics of the second argument seems purposefully confounding.

We don't have `forall ... ->` in data constructors yet. (Interestingly, #281 is completely silent on this point. Is that an omission? @int-index may want to think about this.) So one could argue that these examples are specious: "let's wait until we have `forall ... ->` in data constructors before worrying about this." I think that's quite short-sighted, and thus only answer (B) will work smoothly in all of these cases.

Barring agreement about (B), I advocate for (C), which can be fixed (to (B)!) later. If the committee *still* thinks that (A) is the correct answer, I would advocate waiting to implement until we have more related proposals on the table, so that we realize that (B) -- or perhaps (C) -- is the correct answer. Don't worry -- I'm not going to stonewall (i.e. become obstructionist) over this point; that's not my style. But I do feel quite strongly, as you see.

@gridaphobe, in the GHC SC thread, offers a tantalizing other solution:

> There is another option for unifying the behavior of type and value parameters in patterns, make values more like types. That is, allow repeated occurrences of term variables in patterns, with the semantics that the pattern only matches if the values are the same. This could be implemented as a trivial desugaring
>
> ```hs
> f x x = e  ~>  f x x' | x == x' = e
> ```
>
> and it would give the surface language a nice extra declarative feeling.

I'm in full support of this direction. My problem isn't the mixing of binding and occurrences, per se -- it's treating types and terms differently. If we allowed it for terms (allowing both repeated variables and already-in-scope variables, via a desugaring to an equality check) then I'm very happy to allow it for types. This might even be a good idea more generally! And if anyone thinks it might be worth exploring, then choosing (C) is very sensible, while we work out the details.

nomeata:
I was, and still am, quite attached to the idea that `Con @t x` and `Con (x :: t)` should work the same. So whenever someone bring up an issue about how that behavior of `@t` is wrong or undesirable, I hink, “well, then fix pattern signatures as well”…

So it that a way forward? Have `DHCompatiblePatternSignatures` (tmp name) that has a different rules for how type variables in pattern signatures could work, matching your eventual vision (for `@t` and `::t`)?

My other heistation here is that this this discussion seems to be fundamentally about distinguishing binders from ocurrences. And that problem seems to apply to types and terms alike! For terms, so far we only have occurrences on the LHS of a view pattern, but I really would like to see pattern synonyms with value-parameters, and Eric’s non-linear term patterns also has variables that are not all binders.

Is there a chance we can step back and consider a solution for distinguishing variable binders and variable occurrences, that works uniformly for term and type variables (very much in the spirit of #378) and will open a cleaner way to `forall ->`, to parametrized pattern synonyms, to non-linear term patterns etc…?

If there is such a chance, then maybe (A) (i.e. `@t` behaves like `:: t`) is _still_ the right choice here, with the hope for a uniform fix to the binder/occurrence confusion that applies to `@t` and `:: t` and term variables alike.

int-index:
> Is there a chance we can step back and consider a solution for distinguishing variable binders and variable occurrences

Here's a thing that I've tried myself as a beginner (when I was learning about pattern matching) and have seen other Haskell beginners attempt:

```
a = 42

f a = True
f _ = False
```

Here, the expected semantics are `f a' | a' == a = True`. And it makes a great deal of sense. But there's no way to explicitly tell the compiler whether it's a binding or an occurrence.

Personally, I would expect a variable to be a binding if it's a fresh variable name, and an occurrence if it has been bound earlier. We already have `-Wname-shadowing`, so we can easily tell which case it is. Too bad this is incompatible with H98.

simonpj:
> Personally, I would expect a variable to be a binding if it's a fresh variable name, and an occurrence if it has been bound earlier.

I don't think that would be a good idea. Just imagine if we added `x=3` to the Prelude.  Suddenly, every single function that uses x as a parameter would change its meaning --  the program might still typecheck, but it'd mean something utterly different.

I know the this is the current rule for type variables in pattern signatures, but type variables are different in two ways
* They are always local.  No top-level type variable bindings, let alone imported ones.
* If you brought 'a' into scope by adding a binding (perhaps in an outer type signature) that might cause a static type error, but it cannot silently alter runtime behaviour.

If you want to go this route, you'll need an annotation on each occurrence site in the pattern -- and then I think the cure becomes worse than the disease.

nomeata:
Right! It seems that my line of thought leads to one of these directions:

 1. We introduce explicit syntax to distinguish binders from ocurrences in patterns (`f a` vs. `f .a`) or something, the default being “binder” to not breaking a lot of code, but breaking some existing uses of pattern signatures.
 2. We introduce explicit syntax to distinguish binders from ocurrences in patterns (`f a` vs. `f .a`) or something, with complicated rules around what the default is in which contexts, to keep backward compat (in terms and `@`, default is binder, in `::` default is occurrence)
 3. We embrace the direction taken by pattern synonyms as they are now, and _always_ distinguish binders and ocurreces based on what’s in scope. With that extension on (which would be a requirement for `TypeAppInPatterns`), shadowing a term variable in a pattern simply is an error (not just a warning). We have uniformity across types, type apps and type signatures, paving the path to ergonoic DH.

3  seems to be a bit radical, so I wonder if I am missing a more clever, elegant solution to this problem.

nomeata:
> I know the this is the current rule for type variables in pattern signatures, but type variables are different in two ways

I assume that this kind of reasoning should be avoided upon in light of #378, and (if we accept that proposal’s premise) that we should design new featurs in a way that does _not_ rely on the fact that type and term variables are different _now_.

goldfirere:
I'm pleased to see a ferment of ideas of how to get past the impasse here. Thank you.

@nomeata 

> well, then fix pattern signatures as well

Yes, please!

> And you are not worried about the loss of that expressivity?

I would say disappointed, but not worried. If the need for this expressivity is great enough, someone will come up with some syntax to support it. 

> I am still also quite attached to be able to take [example]

That example was, at first, compelling. Now, not so much. For example:

```hs
f 5 = 5
f _ = 2
```

We don't expect to be able to abstract over `5` to get

```hs
f x = x
f _ = 2
```

Constants and variables really are different. :)

@simonpj

> [Type variables] are always local. No top-level type variable bindings, let alone imported ones.

I don't see this being true once we have dependent types, as we'll be able to use ordinary variables in type signatures. So (like Idris) we'll have to distinguish between not-yet-in-scope variables and ones that are in scope.

I've been wanting (1) from @nomeata's https://github.com/ghc-proposals/ghc-proposals/pull/291#issuecomment-728812082 for some time. But, sadly, I don't think we can use `.` -- it looks too much like a record selector. How about `,a` or `` `a ` or `'a` (though the last looks too much like a TH name). If we use `*a`, we can make C programmers feel nicely at home. It's too bad that "binder" really should be the default, because `?a` makes for a decent way to denote a binder. 

By the way, if we use a sigil to denote an occurrence, I don't think that we do so immediately in terms. Perhaps in the future, we would allow the new notation only in irrelevant (that is, erased) arguments, anyway.

int-index:
We're running out of ASCII, but `^x` is an unused sigil.

goldfirere:
A conversation with @sweirich clarified an important aspect of all this.

The original #126 describes a semantics for variables that differs in type from what it means in terms. This #291 observes (in the spirit of #378) that treating types and terms differently is troublesome. One real worry is that, one day, type arguments may become required arguments as in #281 (with no need for `@`), and then treating types and terms differently will be very hard, indeed.

Keeping with the spirit of #378 forbids me from suggesting that types are different from terms. And we surely don't want patterns with the invisibility override `@` (my interpretation for that sigil) to be different from patterns without. But what about the difference between *relevant* patterns and *irrelevant* ones?

A relevant pattern is one that affects the execution of the pattern-match and may bring relevant (today: term-level) variables into scope. An irrelevant pattern does not. My [thesis](https://richarde.dev/papers/2016/thesis/eisenberg-thesis.pdf#subsection.4.2.2), Section 4.2.2, contains a good introduction to relevance; you can read that section divorced from the rest of the tome. Today, terms are relevant and types are not.

Consider matching `e` against the pattern `(Just @Bool True)`. The `Bool` and `True` play very different roles in the match. The `Bool` is simply informative: it tells both readers and the type inference engine that the type of `e` must be `Maybe Bool`. The `Bool` influences compile-time behavior but does *not* have any influence on runtime behavior. The `True`, on the other hand, *does* influence run-time behavior: it says what bit to look for in memory when examining the result of evaluating `e`. These roles are really different and will remain so.

Now, let's consider for a moment the idea of using already-in-scope variables in patterns (both in irrelevant ones and relevant ones). Consider matching `e` against the pattern `(Just @a x)`, where the type variable `a` and the term variable `x` are already in scope. If we allowed matching in term-level (ahem, relevant) patterns, this would mean that the pattern is equivalent to `Just @a x' | x == x'`, which would induce an `Eq a` constraint and call the `==` method. On the other hand, the occurrence of `a` in that pattern causes none of these: it just influences type inference, and performs a check to make sure that, if `e :: ty`, then `ty ~ Maybe a`. We could call both of these behaviors "equality checks" (and they both are), but they are *really different*. Having the same syntax for them might even be considered bonkers.

So: I propose a new way out of this morass. Use a different notation for irrelevant patterns than for relevant ones. Irrelevant patterns allow the occurrence of in-scope variables (unless we specify `-XNoPatternSignatureBinds` of #285, which should probably be renamed in light of type applications in patterns); relevant ones do not. Irrelevant patterns are syntactically delineated in code as irrelevant (with any of the sigils here proposed -- I will use `^`). So we would write `Just ^(@Bool) x` or `Just ^(@a) x` or maybe `Just ^@Bool x` if we're willing to steal more syntax (`^@` is a valid operator). If we had visible `forall` in data constructors, the `^` would be required for all irrelevant arguments. In a sense, it's redundant: type inference tells us precisely where the `^`s should go. But it's still extraordinarily helpful both to the reader and to GHC to know which patterns affect execution, and which are merely informative. The rule would be this: provided inference succeeds (without any type defaulting), changing a `^` pattern to be `_` would never affect runtime behavior.

nomeata:
Thanks, that’s a helpful phrasing! I always felt that there was some confusion here that’s blocking us, and the confusion between “informs the typechecker” and “does relevant matching” might be the one.

How does this `^p` differ from `.p` in Agda?

Would this be valid? (Or maybe just “morally valid, but we woudn’t go there”)
```
data Unit = Unit
data Foo = Foo !Unit
foo (Foo ^Unit) = "hey"
```
given that the compiler can just as well infer that `Unit`, and changing `^Unit` to `^_` would never affect runtime behavior?

gridaphobe:
This sounds like a plausible resolution to me as well, I think the relevant/irrelevant distinction nicely matches our existing intuition about how type annotations should behave. I have one question though, do we really need a new sigil to denote matching on irrelevant arguments? At the moment, type arguments are both invisible **and** irrelevant, so it seems like we could just declare that `@` actually means matching on an irrelevant argument. In fact, according to the [quantifier table](https://gitlab.haskell.org/ghc/ghc/-/wikis/dependent-haskell#quantifiers) you put together, the only place we currently distinguish between visibility and relevance is with constraints, which are invisible but relevant, and notably cannot be matched on with `@`.

int-index:
> Having the same syntax for them might even be considered bonkers. Use a different notation for irrelevant patterns than for relevant ones.

I like the idea, but seeing `@` and `^` together really makes me question the syntax. Do we really need more sigils?

Ericson2314:
Overall I would say while that while it's true that irrelevant and relevant patterns are different, I'm not sure we want to double down on this with...another axis of variation?

Consider these weird things:

1. I would say an irrelevant pattern match and a irrefutable pattern match are pretty similar. When the variable is existential / unconstrained, it is a "fresh" thing to bind even if it isn't actually provided at run-time.

2. A related question is whether we will keep our lower-case vs upper-case conventions with dependent Haskell. (Note that, if one imagines "demoting" type families and type synonyms, we have already broken this somewhat.) Because if we dispense with that question, then even with relevant patterns we have a similar issue telling apart 0-arg patterns vs fresh variables.

3. If we get equality constraints between values, we can have a funny situation where we have two relevant values and a proof they are equal, we might want to "0 cost" pattern match on one of them using the equality proof to discharge having to do any work.

So I still think it's better to just...not have extra notation / shorthand for now, or have a `= a` sort of thing that gets at the heart of what the users intent is "something is equal", rather than the mechanism that makes this possible "irrelevance and equality being discharged statically".

cgibbard:
Yeah, I really don't (yet) want even more notation here -- it's already kind of hard to think of examples where I personally would use type applications in patterns, and among those examples, I can think of none where inducing type equalities between variables from outer scopes and the matched parts of the type is what would be needed (though there are a good few where I would want to discover a new equality afterward, by pattern matching on a GADT).

But maybe people have some fun ideas about how they want to apply this feature.

My own preference is to make pattern matching work as similarly as possible for types and terms here, so variables would always be fresh. I don't particularly like the thought that moving a function definition that uses type applications in patterns into a where clause for another function that perhaps uses ScopedTypeVariables would cause the patterns to mean something completely different.

Maybe someday we'll also get explicit invisible function parameters, and the @ notation can be used at the term level for mentioning invisible things then as well.

Ericson2314:
Not to open another axis of contention to worry about, but with matchable functions from #242, we might someday want
```haskell
f :: (A -> @M B -> @M C) -> C -> ...
f (g :: A -> @M B -> @M C) (g (Asdf :: A) (x :: B) :: C) = ....x....
```
Where the data constructor `g` that is being matched on a is not a constant. Now, perhaps matchable things should always be capitalized, so we have
```haskell
f :: (A -> @M B -> @M C) -> C -> ...
f (G :: A -> @M B -> @M C) (G (Asdf :: A) (x :: B) :: C) = ....x....
```
But now it looks like `G` is a constant, matched in the first pattern (perhaps refutably). No we have the same problem at the term level; the grand coincidence that capitalization, matchability, and bind-vs-use all agree is broken some more.

I am certainly *not* saying we should decide what we want not, trying to fully plan dependency and first-class matchability all at once is a recipe for madness. Rather, I'm hoping to provide another piece of evidence that we ought to tread carefully, and be wary of the continued resilience of the old correlated dichotomies.

int-index:
While discussing this proposal with @goldfirere and @simonpj, we made the following observations

1. There’s no loss of expressiveness as previously assumed in the discussion here. Define `asserteq` of the following type:
    ```haskell
    asserteq :: forall a b c. (a~b) => c -> c
    ```
    Now it can be used to create equalities using `asserteq @a @b blah`. So instead of:

    ```haskell
    f :: forall a. Maybe [a] -> Maybe [a] -> ...
    f (Just @[a] x) (Just @[a] y) = ...
    ```

    One would be more explicit and write:

    ```haskell
    f :: forall a. Maybe [a] -> Maybe [a] -> ...
    f (Just @[a] x) (Just @[b] y) = asserteq @a @b ...
    ```

    Although a bit more wordy, it matches what we have in terms (where we need to use `==` explicitly rather than repeat the same variable to check for equality).

2. With visible foralls (#281), the renamer can't tell the differnece between the two args in
    ```haskell
    f (K (a,b) (x,y)) = ...
    ```
    where

    ```haskell
    data T p where
      K :: forall p -> p -> T p
    ```

    So the scoping for `a,b`, and `x,y` have be the same. But then, what if we used an ordinary `forall`?

    ```
    data T p where
      K :: forall p. p -> T p

    f (K @(a,b) (x,y)) = ...
    ```

    Suddenly the scoping for `(a,b)` is different (unless we accept this proposal). No good.

Ericson2314:
https://gitlab.haskell.org/ghc/ghc/-/issues/19847 brings up *explicitly* passing information into pattern synnonyms, just as we (rather hackily) support today *implicitly* with the "two `=>`" rule. I think that is compelling, but, in order to uphold the principle that name resolution should not be type directed, I think should require a call site disambiguator.

So I think we should still accept this amendment for the reasons @int-index gave above, but we might later:

1. Plan out a system for visible parameter for type synonymy (rather than just "outputs" to be marched)
2. Consider reusing any "input argument" disambiguation syntax from part 1 as a shorthand for the "assert eq" case defined above, as discussed earlier in this thread.

goldfirere:
What's the status of this proposal? Is it ready to submit?

Ericson2314:
Ah right, forgot it wasn't because there was that discussion mentioned above. Let's submit!

nomeata:
It seems that the change specification list of cases does not cover `@a` where `a` is a simple type variable that's in scope. Is that simply forbidden?

@serras , can you play secretary here?

Ericson2314:
On second thought, I think I can harmonize this with #238 a bit more. Just the slight interaction with scoped type variables.

@nomeata that is not my intent, I'll double check.

nomeata:

nomeata:

nomeata:
Good point!

PR #282, opened 2019-10-11
RecordDotSyntax language extension proposal

shayne-fletcher-da:
We propose a new language extension `RecordDotSyntax` that provides syntactic sugar to make the features introduced in [the `HasField` proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0158-record-set-field.rst) more accessible, improving the user experience.

[Rendered](https://github.com/shayne-fletcher-da/ghc-proposals/blob/record-dot-syntax/proposals/0000-record-dot-syntax.md)

PR #293, opened 2019-11-14
Tweaks to #229 (under review)

int-index:
1. Include keywords in the "opening" and "closing" token categories. This makes the implementation *much* easier and more efficient. For the users, it means they will be getting different error messages if they write something like `f = $case x of ...`. There should be no other effect. (I know I've said otherwise in https://github.com/ghc-proposals/ghc-proposals/pull/229#issuecomment-498397349, but that comment is wrong because I didn't take parsing errors into account).

2. Clarify that the overrides for `@`, `$`, `$$`, are guarded behind the appropriate extensions. This was the original intent anyway, but now it's spelled out.

3. Introduce `-Woperator-whitespace-ext-conflict`, a weaker form of `-Woperator-whitespace`, which would address the use case in https://github.com/ghc-proposals/ghc-proposals/pull/282#issuecomment-553534628 (cc @cdsmith)

4. A suffix occurrence of `@` (e.g. `foo f@ (Just x) = ...`) is a parse error, as per @simonpj request.


nomeata:
Is this just an editorial change/fix? @goldfirere (as the original shepherd), can you assess that?

int-index:
No, it's a change that requires (however brief) a committee review. The changes are user-facing.

nomeata:
Thanks for the clarification! :-)

goldfirere:
Not just an editorial change. But I would hope that this could be accepted with only the briefest review. :)

Given the lack of commentary and the value of expediency: I hereby submit this proposal to the committee and assign myself as shepherd.

I recommend acceptance. The difference is to correct an oversight regarding keywords and to forbid suffix uses of `@`. In the previous version of this proposal, `foo x@ (Just y)` was an as-pattern. Now it's a parse error. I think this is a clear improvement, even if it takes us one tiny step further from Haskell98.

goldfirere:
The majority of the committee has agreed to this minor change, with no dissent, and a week has passed. I declare victory and shall merge.

PR #283, opened 2019-10-13
Local modules (under review)

goldfirere:
This is an alternative to #205 that appears to be more compositional.

[Rendered](https://github.com/goldfirere/ghc-proposals/blob/local-modules/proposals/0000-local-modules.rst)

PR #295, opened 2019-11-15
WIP: First class modules

michaelpj:
This proposal suggests a conceptual reinterpretation to make modules first-class citizens in the module system. This then unlocks a number of further improvements straightforwardly.

This is not nearly done enough for submission to the committee, but I think that most of what I want to say is here, even if it could do with more justification, specification, and examples. I hope this will start some discussion especially with @goldfirere in the context of https://github.com/ghc-proposals/ghc-proposals/pull/283. If people think this is a promising direction I will work on it more.

[Rendered](https://github.com/michaelpj/ghc-proposals/blob/imp/first-class-modules/proposals/0000-first-class-modules.rst)

### DIFF
--- a/proposals/0000-first-class-modules.rst
+++ b/proposals/0000-first-class-modules.rst
@@ -1,0 +1,612 @@
+.. author:: Michael Peyton Jones
+.. date-accepted:: Leave blank. This will be filled in when the proposal is accepted.
+.. proposal-number:: Leave blank. This will be filled in when the proposal is
+                     accepted.
+.. ticket-url:: Leave blank. This will eventually be filled with the
+                ticket URL which will track the progress of the
+                implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+First class modules
+===================
+
+The Haskell module system helps us manage the names of the *entities* of a
+Haskell program. That includes values, types, typeclasses, but notably
+*not* modules. We can justifiably say that modules are second-class
+citizens of their own system.
+
+This proposal makes modules just like everything else in a Haskell program,
+so they can be named and managed in the same way that we already do successfully
+for other entities. In doing so, we also solve some existing ergonomic
+issues with the module system.
+
+Motivation
+----------
+
+Modules are not first-class in the Haskell module system. They cannot be
+named in the way that other entities can, and this makes them difficult
+to manipulate. From the point of view of the Haskell report, modules barely
+even exist - the only real thing is qualified names! This is frustrating
+for users, since modules *seem* to have names, and they *seem* to have
+a logical role as entities which are “containers” of bindings. [#]_
+
+.. [#] This gives them a distinct similarity to objects, with Scala
+   taking this to the logical extreme of identifying modules and objects.
+
+This proposal suggests that modules should be entities in the module
+system. This requires a conceptual reimagining of the current system, but it
+allows us to make a number of further changes which unlock a lot
+of power.
+
+There are many examples of problems with the current module system in
+[ghc283]_ and [ghc205]_  which I
+will not repeat here: I believe this proposal solves them all.
+
+Proposed change specification
+-----------------------------
+
+Outline
+~~~~~~~
+
+This proposal consists of a series of changes, many of which could be
+adopted separately, but which build upon a central change which is
+described in the main section. However, the additions are important for
+the motivation of this proposal, and I would not recommend that it be
+accepted without at least some of them.
+
+This proposal is conceptually based on [hsmods]_. We assume that
+this provides an accurate representation of the current state of the
+Haskell module system and we describe our changes with respect to it. [#]_
+
+.. [#] I find the relational setting very amenable for this work, since it makes it
+       easy to describe the rules while gracefully handling un- or multiply-defined
+       bindings.
+
+The text here is mostly expository. There is a `Datalog implementation`_ of some
+of the rules.
+
+The proposed changes are:
+
+1. `Modules as entities`_
+2. `Module export specifiers`_
+3. `Local modules`_
+4. `Module aliases`_
+
+Terminology
+~~~~~~~~~~~
+
+-  A *compilation unit* is the unit of Haskell compilation, currently a
+   single ``.hs`` file.
+-  A *file module* is a module which is defined at the top level of a
+   ``.hs`` file. Currently this corresponds to a compilation unit.
+-  A *module* is a named scope, of which file modules are currently the
+   only examples.
+-  A *simple name* is a name that is given to something when it is
+   defined.
+-  A *qualified name* is a more complex form of name which can *refer*
+   to an entity. We will say more about qualified names later.
+-  A *dotted name* is a (simple) module name that contains ``.``\ s, e.g
+   ``Data.Set``. We will want to distinguish these from qualified names.
+-  An *entity* is something that can be referred to with a name.
+   Currently this means types, values, field accessors, etc.
+-  A *binding* is a mapping from a name to an entity.
+
+Modules as entities
+~~~~~~~~~~~~~~~~~~~
+
+This change makes modules into entities, and allows them to be referred
+to. However, the only place that they can be referred to is in the
+qualifiers of qualified names, so this changes little.
+
+I believe this would be entirely internal, and would have no effect on
+what names resolve to in existing Haskell programs. As such, it could be
+implemented without a language extension flag, and is completely
+compatible with existing code.
+
+Modules become entities
+^^^^^^^^^^^^^^^^^^^^^^^
+
+Modules become entities. All scopes now have a *module namespace*, which
+contains the name bindings for modules. [#]_
+
+.. [#] I will ignore namespaces for the rest of this proposal. They don't
+       interact with this proposal much, since it is easy to distinguish which
+       bindings belong to which namespaces by classifying the *entities* according
+       to which namespace they belong to. Then we can have a "unified" namespace and
+       simply filter out bindings that refer to the wrong kind of entities. 
+
+Qualified names become more structured
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Currently, qualified names consist of a simple name, possibly qualified
+by a module name. That is, they could be represented by a type like:
+
+::
+
+   data QualName = Qualified ModuleName Name | Simple Name
+
+We propose to replace this with a type like:
+
+::
+
+   data QualName = Qualified QualName Name | Simple Name
+
+That is, a qualified name is a simple name with a (possibly empty) list
+of qualifying names. These correspond to the dot-separated segments of
+the textual name.
+
+Scopes contain simple names only
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Currently, the ``inScope`` relation relates *qualified* names to
+entities, with the following signature:
+
+::
+
+   inScope :: Rel Module QualName Entity
+
+We propose that instead it relates *simple* names to entities, and we
+add a new relation ``refersTo`` which relates qualified names to
+entities in a scope. [#]_
+
+.. [#] By analogy to programming languages: the bindings in scope are like an environment,
+   simple names are like variables, qualified names are like expressions, and ``refersTo``
+   is ``evalExpr``.
+
+::
+
+   inScope :: Rel Module Name Entity
+   refersTo :: Rel Module QualName Entity
+
+``refersTo`` works as follows:
+
+- If the qualified name has no
+  qualifiers, then it refers to whatever that simple name maps to in the
+  scope.
+- If the qualified name has a qualifier, then we work out what
+  the qualifying name refers to. If it refers to a module, then we look at
+  the bindings exported by the module to find one matching the name.
+
+That is, we interpret ``A.b`` by first working out what ``A`` refers to, as a
+module, and then looking inside it to find ``b``. ``A`` may itself be
+qualified (since modules may export module bindings, just like any other
+bindings), hence the need for the recursive lookup.
+
+Exporting module names
+^^^^^^^^^^^^^^^^^^^^^^
+
+Module names *cannot* be exported, since there are no export specifiers that can
+refer to them. [#]_ The existing ``module M`` export specifier
+continues to work as it is defined in the Haskell report.
+
+.. [#] See `Module export specifiers`_ for the changes to allow this.
+
+Imports
+^^^^^^^
+
+Imports at the top level of a file module are treated specially. Their
+target is interpreted as a *dotted* name, which is resolved using the
+current rules for locating external Haskell modules. [#]_
+
+.. [#] For imports in other settings, see `Local modules`_.
+
+Importing a module brings all the exported bindings of the target module
+into scope, as described in [hsmods]_. It also brings
+the target module into scope bound to its declared simple name (which may be
+a dotted name if it is a file module).
+
+If a module is imported ``qualified`` then the target module is no
+brought into scope.
+
+If a module is imported ``as`` a simple name, then the target module is brought
+into scope with that simple name instead.
+
+If multiple modules are imported ``as`` the same name, then we retain the current
+behaviour, which is to “merge” modules. [#]_ The semantics is roughly that:
+
+.. [#] See `What should happen if multiple modules are defined
+       with the same name?`_ for discussion.
+
+::
+
+   import A as M
+   import B as M
+
+means the same as
+
+::
+
+   module M where
+       import A
+       import B
+
+This retains the current behaviour with respect to clashing names, since
+Haskell does not issue ambiguity warnings on import, only on usage. Thus
+the hypothetical definition of ``M`` is error-free even if ``A`` and
+``B`` export ``n``, but an attempt to use ``M.n`` will result in an
+ambiguity error. [#]_
+
+.. [#] This explication of the semantics uses a local module definition,
+   but we could wire the behaviour in as a special case if we decide
+   not to support local modules.
+
+Module export specifiers
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+The previous section allows us to refer to modules by name, but not to
+export those names.
+
+We propose adding a language extension ``ExportModuleNames`` which
+*changes* the behaviour of the ``module M`` export specifier. It now
+exports the name ``M`` as a module binding.
+
+This ``ExportModuleNames`` behaviour is the “right” behaviour as far as
+this proposal is concerned, but we gate it behind a language extension
+for backwards compatibility.
+
+Local modules
+~~~~~~~~~~~~~
+
+We propose adding a language extension ``LocalModules``. This allows
+module definitions within modules, with the same syntax as in [ghc283]_. Local
+modules are very like file modules, in particular:
+
+- They can have imports.
+
+  - The imports can only target modules which can be referred to
+    with a qualified name in the current scope. Their target module name is *never*
+    interpreted as a dotted name.
+
+- They can have nested module definitions.
+- They can have export specifiers.
+- They can *not* have dotted names.
+
+The rules for local modules are mostly the same as for file modules,
+but:
+
+- A local module definition adds a binding for its declared name to
+  its enclosing scope, as usual for definitions.
+- Bindings from the
+  enclosing scope are in scope in the local module, but are shadowed if
+  there is a local definition of the same name. [#]_
+
+.. [#] See the `Recursion`_ section for some subtleties.
+
+Local modules *cannot* import file modules. This ensures that we can
+always compute the dependency graph of the compilation units of a
+Haskell program by looking at the top-level imports alone. [#]_
+
+.. [#] See `Do we need different syntax for "local" imports?`_ for discussions
+   of the consequences.
+
+Module aliases
+~~~~~~~~~~~~~~
+
+We propose adding a language extension ``ModuleAliases``. This allows a
+new kind of top-level definition: ``module <name> = <qualname>``
+
+A module alias adds a binding in the enclosing scope from the new name
+to the module referred to by the right hand side.
+
+Examples
+--------
+
+Exporting small utility modules
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+::
+
+   module A (f, module Unsafe) where
+
+   f = Unsafe.g …
+
+   module Unsafe where
+       g = … 
+
+Exporting qualified names
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This proposal allows exporting modules, but doesn’t say they can be
+exported under *qualified* names, which is something that people want to
+do. Consider the following `proposed
+example <https://github.com/ghc-proposals/ghc-proposals/pull/283#issuecomment-541525245>`_:
+
+::
+
+   module MyPrelude ( qualified module BL
+                    , qualified module BS
+                    , Set
+                    , qualified module Set ) where
+
+
+   import qualified Data.ByteString.Lazy as BL
+   import qualified Data.ByteString as BS
+   import Data.Set ( Set )
+   import qualified Data.Set as Set
+
+This could be achieved as follows:
+
+::
+
+   module MyPrelude (module Data, Set) where
+
+   import qualified Data.ByteString.Lazy as BL
+   import qualified Data.ByteString as BS
+   import Data.Set ( Set )
+   import qualified Data.Set as Set
+
+
+   module Data where
+       module ByteString where
+           import BS
+           module Lazy where
+               import BSL
+       module Set where
+           import Set
+
+Because qualified names correspond to module-qualification, we can
+always create that structure ourselves. This is somewhat laborious for
+the author, but does allow the desired experience for the consumer.
+
+This example is somewhat unsatisfactory, since one would really like to
+re-export the “module” ``Data.ByteString``, but this is not a module as
+far as this proposal is concerned. See `Can we make dotted names less
+special?`_ for more discussion.
+
+Effect and interactions
+-----------------------
+
+I believe this proposal would solve all the problems described in [ghc283]_ and [ghc205]_.
+Moreover, because it makes modules just like other entities, improvements to the module
+system will equally benefit management of modules.
+
+Interactions
+~~~~~~~~~~~~
+
+`Modules as entities`_ should have no interaction with any other
+features. We do not even need to change interface files, since module
+names cannot be exported.
+
+`Module export specifiers`_ has slightly more effects:
+
+- Interface files must be able to handle the possibility that an exported name refers to a
+  module. This may have some interaction with Backpack.
+- Code which *imports* such modules can always treat it as if it
+  exports qualified names. In particular, they do not need to be able to
+  *import* exported module names to use them.
+
+`Local modules`_ and `Module aliases`_ should again have few interactions
+and has effects only in the module where it is used. Consumers again can
+again interact with nested modules only using qualified names.
+
+Costs and drawbacks
+~~~~~~~~~~~~~~~~~~~
+
+This proposal requires changing the way name resolution works in
+Haskell. This is a critical part of the compiler, and the current
+implementation is very battle-tested. Replacing it would be hard work.
+
+However, it *should* be relatively easy to test the new implementation:
+all existing modules should have no change to their interfaces.
+
+This proposal would close off some other approaches to improving the
+module system.
+
+Alternatives
+------------
+
+The primary alternative is the “Local Modules” proposal [ghc283]_. This
+proposal aims to support most of the same things, but I will point out a
+few differences:
+
+- Export specifiers
+
+  - The export specifiers in [ghc283]_ are quite complex, and there are questions about how they should
+    behave for nested modules etc. The export specifiers in this
+    proposal are *extremely* simple: they just allow exporting a
+    module name. All additional structure must be added by structuring
+    the exported module.
+
+- Module aliases
+
+  - Module aliases barely make sense in the world of [ghc283]_
+    since there are no such things as modules.
+
+Conceptually, [ghc283]_ is a bit 
+like an “unrolled” version of this proposal, where we try and always
+work with the fully explicated list of qualified names. The problems
+above arise because we sometimes *want* to talk about modules
+themselves, but there are also problems with working with the full set
+of qualified names:
+
+- Recursive modules could cause non-termination.
+
+  - If ``A`` exports itself, then arguably it should export an
+    infinite set of qualified names: ``A``, ``A.A``, ``A.A.A``\ …
+  - [ghc283]_ avoids this by simply banning recursive local modules, but we don't
+    have to do this.
+  - See `Termination`_ below for why this isn’t a problem for this
+    proposal.
+
+Ultimately, I think this proposal is just *simpler*, in that it makes
+modules more like everything else in the module system, which then
+allows us to solve our problems with the namespacing tools we already
+have for other bindings.
+
+There is also an earlier proposal “Structured module exports/imports” [ghc205]_
+I won't respond to it in detail, but it takes the same approach of continuing to
+focus on qualified names rather than giving modules an identity.
+
+Unresolved questions
+--------------------
+
+What should happen if multiple modules are defined with the same name?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+[ghc283]_ effectively proposes that multiple modules with the same name should be
+*merged*. I would like to say that multiple modules with the same name
+should be *ambiguous*, just like all other names.
+
+I think the argument for merging is:
+
+- We have some existing cases where this happens (``import as``), which we need to support.
+- It might be convenient to “mix in” additional names into an existing module.
+
+The argument for ambiguity is:
+
+- Ambiguity is consistent with the way names work for everything else, so makes the system work.
+- It is not very inconvenient in practice
+
+  - Anecdotal evidence: I wrote the Semmle QL module system this way, and I never had anybody ask
+    for module merging.
+  - It is easily to “manually” merge modules if you have local module definitions.
+
+If we thought ambiguity was more consistent with "modules as entities", we could add a language
+extension `AmbiguousAs` which makes such `import as` statements produce ambiguity errors instead.
+
+Do we need any tricky rules for typeclasses/type family instances?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Consider:
+
+::
+
+   module M where
+       instance Ord T where …
+       f :: T -> T -> Bool
+       f = compare
+
+   g = M.f
+
+That is, we can use ``f`` without importing its defining module, and
+hence the typeclass instance is not in scope. Possibly this is fine, but
+I’d like someone who knows more about how typeclasses work to think
+about this with me!
+
+An alternative would be for typeclass instances to be associated with a
+file module rather than any module. So if you import a module which
+contains an instance inside a nested module, then that instance is used
+during resolution even if it is not strictly in scope. But this would be
+somewhat unhygienic.
+
+Can we make dotted names less special?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+File modules declared with dots in them are awkward. The dots *look*
+like the qualifier we use in qualified names, but they behaves
+differently: there is a module called ``Data.Set``, but there is no
+module ``Data`` which contains a module ``Set``.
+
+We could change the system so there *was* such a ``Data`` module. [#]_
+However, there is also a scoping
+problem. If we want the dotted-name import ``import Data.Set`` to work
+as a qualified-name import, then there must be a module ``Data`` *in scope*
+at that point. But we probably don’t want ``Data`` to be unconditionally
+in scope in the whole module! (Or do we?)
+
+.. [#] This is the case in Scala, for example.
+
+One possibility would be to have a special scope in the file module
+header where the “external modules” are in scope. I don’t know whether
+this would be nicer overall.
+
+Do we need different syntax for "local" imports?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Consider:
+
+::
+
+   module A where
+       import qualified B
+       module C where
+          import B.C
+
+From a user perspective, the operations carried out in the two import
+statements look very similar. However, there are differences because of the
+way we're handling dotted names and file module imports. We *could* signal this
+by having a different syntax for the new kind of imports, e.g. ``open`` (to mimic ML).
+I don't *think* this is necessary, but I might be wrong, and it might be clearer
+anyway.
+
+Datalog implementation
+----------------------
+
+The text of this proposal is mostly expository. I have written a model
+of the rules for the old and the new system in Souffle
+[souffle]_, a Datalog variant, which is a
+nice format for writing executable logical rules. This can be found at
+[hsmods-logic]_.
+
+So far, I have only implemented the simplest versions of the systems, in
+particular ignoring qualified imports and import/export specifiers. I
+will add qualified imports given more time, but I don’t plan to add
+import/export specifiers, since they aren’t very interesting and just
+act as filters on the imported/exported bindings.
+
+Since it is a logic programming language, the Datalog rules could easily
+be turned into inference rules to give a formal specification of the
+scoping relations of Haskell.
+
+The Datalog implementation also provides *an* implementation strategy: we
+can directly copy the evaluation model of Datalog. However, this is
+unlikely to be what we would do in a real compiler, and a procedural
+version is certainly possible.
+
+Recursion
+~~~~~~~~~
+
+In both systems the defining relations are naturally recursive. This
+represents the possibility of recursive modules. Writing them in Datalog
+gives us this “for free”, since it has good support for recursion via
+fixpoint iteration. [#]_
+
+.. [#] The fact that the Haskell module system could be
+   made recursive via a fixpoint computation is observed in
+   [hsmods]_, but the Datalog version is much simpler.
+
+While supporting recursive modules across compilation units would be
+challenging, supporting recursion between *local* modules (see `Local modules`_)
+would be much less challenging and might be worth considering
+if they are implemented.
+
+In order for the recursion to be well-defined, we must not recurse
+through a negation. This actually has some design implications! *Name
+shadowing* is implemented by saying that a binding from an external
+scope is visible if it is *not* defined locally. We therefore could not
+allow *imported* bindings to shadow enclosing ones, as that would make
+the recursion non-monotonic. Fortunately, we probably don’t want to do
+that.
+
+Termination
+~~~~~~~~~~~
+
+Datalog is a terminating language (Souffle allows for some features that
+can cause non-termination, but I have not used them). So we are
+guaranteed that the systems both terminate, despite the recursion.
+
+This may seem surprising: given recursive modules, could there not be an
+infinite number of ``refersTo`` facts? E.g. if ``A`` exports itself then
+all of ``A``, ``A.A``, ``A.A.A`` etc. refer to ``A``.
+
+However, what grounds the recursion is that we must provide the
+relations that define all the qualified names we are interested in up
+front. This is something we can do in reality too: we are only
+interested in the qualified names that appear in the program. Or if we
+implement this in a procedural way, we can implement ``refersTo`` as a
+function that computes references on demand.
+
+References
+----------
+.. [hsmods] `A formal specification of the Haskell 98 module system <https://web.cecs.pdx.edu/~mpj/pubs/hsmods.pdf>`_
+.. [ghc283] `Local Modules <https://github.com/ghc-proposals/ghc-proposals/pull/283>`_
+.. [souffle] `Souffle <https://souffle-lang.github.io/>`_
+.. [hsmods-logic] `<https://github.com/michaelpj/hsmods-logic>`_
+.. [ghc205] `Structured module exports/imports <https://github.com/ghc-proposals/ghc-proposals/pull/205>`_
--- a/proposals/0000-first-class-modules.rst
+++ b/proposals/0000-first-class-modules.rst
@@ -223,7 +223,7 @@ means the same as
 
 ::
 
-   module M where
+   module M (module A(..), module B(..)) where
        import A
        import B
 
@@ -243,9 +243,14 @@ Module export specifiers
 The previous section allows us to refer to modules by name, but not to
 export those names.
 
-We propose adding a language extension ``ExportModuleNames`` which
-*changes* the behaviour of the ``module M`` export specifier. It now
-exports the name ``M`` as a module binding.
+We propose adding a language extension ``ExportModuleNames`` which adds the following
+export specifiers:
+
+- ``module M`` *changes* to exports the name ``M`` as a module binding.
+- ``module M(a, b, .. c)`` exports ``M`` and some of its names.
+- ``module M(..)`` exports ``M`` and all of its names. [#]_
+
+.. [#] See `Do we need additional syntax for re-exporting all bindings from a module?`_ for more discussion.
 
 This ``ExportModuleNames`` behaviour is the “right” behaviour as far as
 this proposal is concerned, but we gate it behind a language extension
@@ -343,12 +348,12 @@ This could be achieved as follows:
    import qualified Data.Set as Set
 
 
-   module Data where
-       module ByteString where
+   module Data (module ByteString, module Set) where
+       module ByteString (module Lazy, module BS (..)) where
            import BS
-           module Lazy where
+           module Lazy (module BSL (..)) where
                import BSL
-       module Set where
+       module Set (module Set (..)) where
            import Set
 
 Because qualified names correspond to module-qualification, we can
@@ -505,6 +510,12 @@ differently: there is a module called ``Data.Set``, but there is no
 module ``Data`` which contains a module ``Set``.
 
 We could change the system so there *was* such a ``Data`` module. [#]_
+This would require constructing such virtual modules fora all the packages in
+the package database and then merging them. We can do this (as for the "ambiguous
+as" case), but it might create ambiguous names. However, we already have a mechanism
+for resolving this: `PackageImports`. An import with a package specifier should look
+in the *non*-merged virtual module from that package.
+
 However, there is also a scoping
 problem. If we want the dotted-name import ``import Data.Set`` to work
 as a qualified-name import, then there must be a module ``Data`` *in scope*
@@ -535,6 +546,17 @@ way we're handling dotted names and file module imports. We *could* signal this
 by having a different syntax for the new kind of imports, e.g. ``open`` (to mimic ML).
 I don't *think* this is necessary, but I might be wrong, and it might be clearer
 anyway.
+
+Do we need additional syntax for re-exporting all bindings from a module?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This proposal suggests using the export specifier ``module M(..)`` to re-export all the
+names defined in ``M``. This is useful when you want to effectively just rearrange the
+module structure without changing what is in a particular module.
+
+However, by analogy with the export specifier ``C(..)``, this should also export ``M``.
+I'm not sure if this is what we want, and if we don't, then perhaps we should have a
+different syntax.
 
 Datalog implementation
 ----------------------
--- a/proposals/0000-first-class-modules.rst
+++ b/proposals/0000-first-class-modules.rst
@@ -202,8 +202,8 @@ into scope, as described in [hsmods]_. It also brings
 the target module into scope bound to its declared simple name (which may be
 a dotted name if it is a file module).
 
-If a module is imported ``qualified`` then the target module is no
-brought into scope.
+If a module is imported ``qualified`` then the exported bindings of the target
+module are not brought into scope.
 
 If a module is imported ``as`` a simple name, then the target module is brought
 into scope with that simple name instead.
@@ -296,6 +296,7 @@ Module aliases
 
 We propose adding a language extension ``ModuleAliases``. This allows a
 new kind of top-level definition: ``module <name> = <qualname>``
+(where ``<name>`` is a non-dotted module name).
 
 A module alias adds a binding in the enclosing scope from the new name
 to the module referred to by the right hand side.
@@ -336,34 +337,38 @@ example <https://github.com/ghc-proposals/ghc-proposals/pull/283#issuecomment-54
    import Data.Set ( Set )
    import qualified Data.Set as Set
 
+The goal of this is to re-export everything in `Data.ByteString.Lazy` qualified
+under `BL`, so for example `BL.ByteString` would be `Bytestring` from `Data.Bytestring.Lazy`.
+And similarly for the other modules.
+
 This could be achieved as follows:
 
 ::
 
-   module MyPrelude (module Data, Set) where
+   module MyPrelude (module BL, module BS, Set, module Set) where
 
    import qualified Data.ByteString.Lazy as BL
    import qualified Data.ByteString as BS
    import Data.Set ( Set )
    import qualified Data.Set as Set
 
+This example is simple, because we only want a single level of qualification.
+For example, suppose that we instead want to re-export `Data.Bytestring.Lazy`
+as `Data.Bytestring`. Then we need to write:
 
-   module Data (module ByteString, module Set) where
-       module ByteString (module Lazy, module BS (..)) where
-           import BS
-           module Lazy (module BSL (..)) where
-               import BSL
-       module Set (module Set (..)) where
-           import Set
+::
 
-Because qualified names correspond to module-qualification, we can
-always create that structure ourselves. This is somewhat laborious for
-the author, but does allow the desired experience for the consumer.
+   module MyPrelude (module Data) where
 
-This example is somewhat unsatisfactory, since one would really like to
-re-export the “module” ``Data.ByteString``, but this is not a module as
-far as this proposal is concerned. See `Can we make dotted names less
-special?`_ for more discussion.
+   import qualified Data.ByteString.Lazy as BL
+
+   module Data (module ByteString) where
+       module ByteString = BL
+
+Here, we have to construct a `Data` module with the appropriate inner structure.
+This is how we can export "deeply-qualified" names: instead of exporting a "qualified name"
+per se, we export a module whose structure allows us to access the names we want to access
+with the qualifiers that we want.
 
 Effect and interactions
 -----------------------
--- a/proposals/0000-first-class-modules.rst
+++ b/proposals/0000-first-class-modules.rst
@@ -1,16 +1,10 @@
 .. author:: Michael Peyton Jones
-.. date-accepted:: Leave blank. This will be filled in when the proposal is accepted.
-.. proposal-number:: Leave blank. This will be filled in when the proposal is
-                     accepted.
-.. ticket-url:: Leave blank. This will eventually be filled with the
-                ticket URL which will track the progress of the
-                implementation of the feature.
-.. implemented:: Leave blank. This will be filled in with the first GHC version which
-                 implements the described feature.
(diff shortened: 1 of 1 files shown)


proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,612 @@
...
+current rules for locating external Haskell modules. [#]_
+
+.. [#] For imports in other settings, see `Local modules`_.
+
+Importing a module brings all the exported bindings of the target module
+into scope, as described in [hsmods]_. It also brings
+the target module into scope bound to its declared simple name (which may be
+a dotted name if it is a file module).
+
+If a module is imported ``qualified`` then the target module is no
```
int-index:
```suggestion
If a module is imported ``qualified`` then the target module is not
```

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,612 @@
...
+::
+
+   import A as M
+   import B as M
+
+means the same as
+
+::
+
+   module M where
```
int-index: Is this supposed to be `module M (module A, module B) where`? It's not clear to me (at this point in the proposal, maybe it's clarified later?) why importing `A` and `B` would re-export their contents from `M`.

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,612 @@
...
+   not to support local modules.
+
+Module export specifiers
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+The previous section allows us to refer to modules by name, but not to
+export those names.
+
+We propose adding a language extension ``ExportModuleNames`` which
+*changes* the behaviour of the ``module M`` export specifier. It now
```
int-index: Under this extension, how does one get the old behavior? Would it be `module M(..)`?

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,612 @@
...
+
+Can we make dotted names less special?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+File modules declared with dots in them are awkward. The dots *look*
+like the qualifier we use in qualified names, but they behaves
+differently: there is a module called ``Data.Set``, but there is no
+module ``Data`` which contains a module ``Set``.
+
+We could change the system so there *was* such a ``Data`` module. [#]_
```
int-index: `containers` defines `Data.Set` and `Data.Map`.
`unordered-containers` defines `Data.HashSet` and `Data.HashMap.`

How would these `Data` modules get merged when I depend on both packages?

proposals/0000-first-class-modules.rst
michaelpj: Ugh, yes, I was operating on an older model of mine where there were automatic re-exports. Yes, this should have something like `module (module A, module B)`, although I think I'll adopt your suggestion of `module A(..)`.

proposals/0000-first-class-modules.rst
michaelpj: I haven't included something that provides the old behaviour, but I think `module M(..)` would be a sensible syntax for that.

proposals/0000-first-class-modules.rst
michaelpj: We could merge them just as we do for the "multiple import as" case. In particular, we don't have to have a rule that the resulting modules must have no ambiguous names. We would need something like `PackageImports` to resolve that (as we do today).

proposals/0000-first-class-modules.rst
michaelpj: I've added that provisionally along with a note that it needs more thought since I haven't thought about it enough.

proposals/0000-first-class-modules.rst
michaelpj: I've added something to that effect. I think it would actually play nicely with `PackageImports`.

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,634 @@
...
+
+.. [#] For imports in other settings, see `Local modules`_.
+
+Importing a module brings all the exported bindings of the target module
+into scope, as described in [hsmods]_. It also brings
+the target module into scope bound to its declared simple name (which may be
+a dotted name if it is a file module).
+
+If a module is imported ``qualified`` then the target module is no
+brought into scope.
```
deepfire: @michaelpj, if the intent is to replicate the status quo with the new terminology, then I think it should say:

```
If a module is imported ``qualified`` then the bindings of the target module are not brought into scope directly.
```

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,634 @@
...
+
+- ``module M`` *changes* to exports the name ``M`` as a module binding.
+- ``module M(a, b, .. c)`` exports ``M`` and some of its names.
+- ``module M(..)`` exports ``M`` and all of its names. [#]_
+
+.. [#] See `Do we need additional syntax for re-exporting all bindings from a module?`_ for more discussion.
+
+This ``ExportModuleNames`` behaviour is the “right” behaviour as far as
+this proposal is concerned, but we gate it behind a language extension
+for backwards compatibility.
```
deepfire: I think, that to properly retain backward compatibility, the import side also needs to be protected -- so that even if the upstream module is upgraded, the downstream will not need to change, unless it opts into the extended module semantics -- by the means of `ExportModuleNames` or whatever the the pragma should be.

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,634 @@
...
+
+- They can have imports.
+
+  - The imports can only target modules which can be referred to
+    with a qualified name in the current scope. Their target module name is *never*
+    interpreted as a dotted name.
+
+- They can have nested module definitions.
+- They can have export specifiers.
+- They can *not* have dotted names.
```
deepfire: Why disallow dotted names?  File modules can have them, and the un-extended language has no problem re-exporting those.

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,634 @@
...
+Haskell program by looking at the top-level imports alone. [#]_
+
+.. [#] See `Do we need different syntax for "local" imports?`_ for discussions
+   of the consequences.
+
+Module aliases
+~~~~~~~~~~~~~~
+
+We propose adding a language extension ``ModuleAliases``. This allows a
+new kind of top-level definition: ``module <name> = <qualname>``
```
deepfire: What is your proposal on the restriction on the dotted-ness of the new binding?

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,634 @@
...
+
+The previous section allows us to refer to modules by name, but not to
+export those names.
+
+We propose adding a language extension ``ExportModuleNames`` which adds the following
+export specifiers:
+
+- ``module M`` *changes* to exports the name ``M`` as a module binding.
+- ``module M(a, b, .. c)`` exports ``M`` and some of its names.
+- ``module M(..)`` exports ``M`` and all of its names. [#]_
```
deepfire: > `M` and all of its names

This doesn't appears to include all modules that `M` exports, right?

Did you consider allowing transparent re-export of modules-exported-by-modules?

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,634 @@
...
+   import qualified Data.Set as Set
+
+
+   module Data (module ByteString, module Set) where
+       module ByteString (module Lazy, module BS (..)) where
+           import BS
+           module Lazy (module BSL (..)) where
+               import BSL
+       module Set (module Set (..)) where
+           import Set
```
deepfire: I'm not sure that the examples correspond to each other:

The former example doesn't export any module prefixed with `Data`, but does export modules named `BS` and `BL`.

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,634 @@
...
+
+   module Unsafe where
+       g = … 
+
+Exporting qualified names
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This proposal allows exporting modules, but doesn’t say they can be
+exported under *qualified* names, which is something that people want to
+do. Consider the following `proposed
```
deepfire: > exported under *qualified* names

Are qualified names necessary, or will dotted name suffice?

If dotted names will suffice, I think there's a nice simplification to be made?

A further argument for dotted-instead-of-qualified, is that in the end, the compiler will have to dumb down all this nice structure to something that can be serialised.

And as the implementation experience in https://github.com/ghc-proposals/ghc-proposals/pull/205 has shown, it can be a simple step to:
```
-        tcg_exports :: [AvailInfo],     -- ^ What is exported
+        tcg_exports    :: [AvailInfo],                     -- ^ What is exported
+        tcg_exports_aliases :: [(ModuleName, AvailInfo)], -- ^ ..aliased.
```

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,634 @@
...
+~~~~~~~~~~~~
+
+`Modules as entities`_ should have no interaction with any other
+features. We do not even need to change interface files, since module
+names cannot be exported.
+
+`Module export specifiers`_ has slightly more effects:
+
+- Interface files must be able to handle the possibility that an exported name refers to a
+  module. This may have some interaction with Backpack.
```
deepfire: There's the question of the serialisation -- should it be:

1. A reference to a module (that'll have consequences that are difficult to predict, I'm afraid), or
2. A flat list of names-qualified-with-dotted-names -- which appears to be much more limited in impact -- as per https://github.com/ghc-proposals/ghc-proposals/pull/295/files#r347111957

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,634 @@
...
+
+::
+
+   data QualName = Qualified ModuleName Name | Simple Name
+
+We propose to replace this with a type like:
+
+::
+
+   data QualName = Qualified QualName Name | Simple Name
```
Ericson2314: Did you consider the "right-recursive" rather than "left-recursive" 
```haskell
data QualName = Qualified Name QualName | Simple Name
```

I think this makes `refersTo` a bit clearer, as it's now tail recursive, and frankly, all things equal, left-to-right is easier given that's the way Haskell text flows.

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,634 @@
...
+
+Qualified names become more structured
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Currently, qualified names consist of a simple name, possibly qualified
+by a module name. That is, they could be represented by a type like:
+
+::
+
+   data QualName = Qualified ModuleName Name | Simple Name
```
Ericson2314: Keep in mind that `import Prelude as A.B` works today! I was saddened to to learn this in https://github.com/ghc-proposals/ghc-proposals/pull/205. While I am not sure it is fatal, I think we should probably deprecate at as a separate step / limit the use of dotted names in a single `Qualified _ _`.

proposals/0000-first-class-modules.rst
Ericson2314: @deepfire I agree that we should constrain dotted names as much as possible. That is the defining characteristic of this proposal vs the others, and I think that the middle ground between them all isn't as coherent a design as the boundary.

proposals/0000-first-class-modules.rst
Ericson2314: I say it shouldn't be allowed. if it was, it should be sugar for nested local modules as Nix does with ` { a.b.c.d = ...; }`

proposals/0000-first-class-modules.rst
Ericson2314: I'll note that at this point I would switch from being anti-#220 (wart fix is too ad-hoc / bikesheddy) to pro-#220 (imports should match export).

I am likewise sympathetic to the `qualified module` syntax from #205 for matching the status quo in a non-obvious way, but I agree that with the first class modules / "dotted names are bad" approach, the status quo is too harmful to further in enshrine in new features.

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,634 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+The previous section allows us to refer to modules by name, but not to
+export those names.
+
+We propose adding a language extension ``ExportModuleNames`` which adds the following
+export specifiers:
+
+- ``module M`` *changes* to exports the name ``M`` as a module binding.
+- ``module M(a, b, .. c)`` exports ``M`` and some of its names.
```
Ericson2314: I'll note that at this point I would switch from being anti-#220 (wart fix is too ad-hoc / bikesheddy) to pro-#220 (imports should match export).

I am likewise sympathetic to the `qualified module` syntax from #205 for matching the status quo in a non-obvious way, but I agree that with the first class modules / "dotted names are bad" approach, the status quo is too harmful to further in enshrine in new features.

proposals/0000-first-class-modules.rst
Ericson2314: I'll note that at this point I would switch from being anti-#220 (wart fix is too ad-hoc / bikesheddy) to pro-#220 (imports should match export).

I am likewise sympathetic to the `qualified module` syntax from #205 for matching the status quo in a non-obvious way, but I agree that with the first class modules / "dotted names are bad" approach, the status quo is too harmful to further in enshrine in new features.

proposals/0000-first-class-modules.rst
Ericson2314: > A further argument for dotted-instead-of-qualified, is that in the end, the compiler will have to simplify all this nice structure to something that can be serialised efficiently.

Oh I would *really* like GHC to not flatten things with this new design. I think not doing so would basically destroy the proposal. Yes we're making language changes, but I think the associated rich/flexible/structured internal datatype are just as important on their own in the long run.

proposals/0000-first-class-modules.rst
Ericson2314: I actually think the backpack situation will be too bad. Modules can already reexport siganture's items, for example. If I had to make a guess, the export will be restricted to be a module interface with the items in the original signatures, but the types of those items will be richer due to later instantiations.

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,634 @@
...
+
+- Ambiguity is consistent with the way names work for everything else, so makes the system work.
+- It is not very inconvenient in practice
+
+  - Anecdotal evidence: I wrote the Semmle QL module system this way, and I never had anybody ask
+    for module merging.
+  - It is easily to “manually” merge modules if you have local module definitions.
+
+If we thought ambiguity was more consistent with "modules as entities", we could add a language
+extension `AmbiguousAs` which makes such `import as` statements produce ambiguity errors instead.
```
Ericson2314: I think we we all agree with this proposal as our road map, This and something like #220 would be good preparatory steps.

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,634 @@
...
+act as filters on the imported/exported bindings.
+
+Since it is a logic programming language, the Datalog rules could easily
+be turned into inference rules to give a formal specification of the
+scoping relations of Haskell.
+
+The Datalog implementation also provides *an* implementation strategy: we
+can directly copy the evaluation model of Datalog. However, this is
+unlikely to be what we would do in a real compiler, and a procedural
+version is certainly possible.
```
Ericson2314: Datalog nudges us towards principled incremental compilation. Probably not a short term benefit, but differently a long term one. CC @kmicklas @ekmett.

simonmar:
A couple of points from a quick read through the proposal, which at first glance seems like a promising direction.

* It's confusing to first say "Module names cannot be exported, since there are no export specifiers that can refer to them." and then to relax that restriction, especially when part of the motivation is to relax exactly that restriction. I'm wondering whether all of this should be a single extension.

* Perhaps the problems of "dotted modules" could be solved with the usual ambiguity rules. e.g. `A.B.c` means both the name `c` from the module `A.B` and the name `c` from the module `B` from the module `A`, and if both of these interpretations exist then the qualified name is ambiguous.  Obviously this could get confusing, but my guess is that in most cases the meaning will be fairly obvious to the reader.

* I don't understand the example in 1.3.2 at all.

proposals/0000-first-class-modules.rst
michaelpj: Whoops, wrote the opposite of what I meant.

proposals/0000-first-class-modules.rst
michaelpj: Yes, I think my philosophical position is that dotted names shouldn't exist, and dot should *always* mean accessing a module. The only reason they're in this proposal is to accommodate existing behaviour, and if we can come up with a way to get rid of them I'd be keen.

proposals/0000-first-class-modules.rst
michaelpj: Yes, no dots.

proposals/0000-first-class-modules.rst
michaelpj: If you mean:
- `M` has an old-style `module A` export, then that just turns into a bunch of normal exports, so is covered by this.
- `M` has a new-style `module A` export, then that just exports `A` under its name, so that name is covered by this.

Put it this way: I gloss a `M (..)` export as `forall n e . exports(m, n, e), exports(<current module>, n, e)`

proposals/0000-first-class-modules.rst
michaelpj: Yes, I don't think we should flatten, and I'm not keen on dotted names. (I thought I had an example of where flattening makes something go wrong, but I can't reconstruct it: I'll think about it.)

proposals/0000-first-class-modules.rst
michaelpj: I don't know that much about how interface files currently work. However, here's a sketch of a way to structure them that would work fine for exporting modules, even recursive ones.

```
type Scope = Map Name Symbol
type SymbolTable = Map Symbol Entity
data Entity = Type ... | Value ... | Module Scope
-- The `Scope` is the exported scope of the module
data Interface = Interface Scope SymbolTable
```

proposals/0000-first-class-modules.rst
michaelpj: Interesting, I hadn't thought of that. The current way makes more sense to me as an evaluator (evaluate the qualifier and then access it), but in the end it's an implementation detail.

However, there is one quirk. I realise that subconsciously I was preparing myself to add a `DottedName ....` alternative. In which case we can no longer always assume that the first segment is going to be an independent `Name`, whereas we can *always* assume that the last segment is.

proposals/0000-first-class-modules.rst
michaelpj: Gosh, that's annoying. I'd be happy to deprecate it, but I think we're going to have to live in the world where simple names include dotted names for backwards compatibility anyway.

proposals/0000-first-class-modules.rst
michaelpj: I don't quite follow your first paragraph, could you elaborate?

proposals/0000-first-class-modules.rst
michaelpj: Do we all agree that this is bad? In my experience people use this quite heavily, particularly for re-exporting, e.g.
```
module A (module Export) where
import B as Export
import C as Export
...
```

So I don't think removing it would be uncontroversial.

michaelpj:
> It's confusing to first say "Module names cannot be exported, since there are no export specifiers that can refer to them." and then to relax that restriction, especially when part of the motivation is to relax exactly that restriction. I'm wondering whether all of this should be a single extension.

Perhaps it wasn't clear enough, but my goal was to have one entirely behaviour-preserving change, which could be implemented without a language extension, and then subsequent changes which *would* require extensions to enable.

That means we could adopt the "plumbing" of this proposal unconditionally, whereas it might be annoying to be supporting two different kinds of plumbing.

In order to be behaviour-preserving `module M` exports have to work as now: hence the odd dance of keeping them as they are and requiring an extension to get the new behaviour.

> Perhaps the problems of "dotted modules" could be solved with the usual ambiguity rules

Yes, I think this is a good way to go if we allow "dotted simple names", which we may have to. I would still like to get rid of them entirely, but I think it's important to have a good story if we can't.

I need to revise the examples.

proposals/0000-first-class-modules.rst
michaelpj: It looks like I just misunderstood what the first example was trying to achieve. Let me fix that.

michaelpj: I think I've got it right now.

proposals/0000-first-class-modules.rst
michaelpj: I didn't call it out in the proposal but this is true. Although if you wanted to do incremental computation, you'd be modifying the input relations (e.g. `modDefines` & co), which isn't standard. Fortunately, I wrote [a paper](https://link.springer.com/chapter/10.1007/978-3-030-17184-1_19) on how to do that :)

proposals/0000-first-class-modules.rst
Ericson2314: Oh actually I completely take back my suggestion. `.` would be right-associative if we had other expression forms we had to disambiguate with, so your way is right.

proposals/0000-first-class-modules.rst
Ericson2314: I still think its better to (systematically) eliminate which simple names can be dotted, even if some are kept around.

proposals/0000-first-class-modules.rst
Ericson2314: I'm basically saying that saying that if we "fix" the  export syntax, then the import syntax's oddities are a bit more pronounced, and more worth fixing.

proposals/0000-first-class-modules.rst
Ericson2314: sorry "if we" was what I meant to write, not "we we". Even if `AmbiguousAs` isn't deprecated, it would make clear what is sugar and what isn't.

proposals/0000-first-class-modules.rst
Ericson2314: Oh! What a coincidence :)

goldfirere:
While I know it's obscure, I'm worried about interactions with `-XPackageImports`. If I say

```hs
import "pack1" A.B
import "pack2" A.B

foo = A.B.y   -- y is exported only by "pack1"
```

do I get ambiguity because the `A.B` is ambiguous? This would be backward-incompatible.

> That is, a qualified name is a simple name with a (possibly empty) list of qualifying names. These correspond to the dot-separated segments of the textual name.

Not quite, given that dotted names are simple. But I agree fully in spirit.

> Thus the hypothetical definition of M is error-free even if A and B export n,

We must tread carefully here. Today, if I have `module M (module A, module B)` and `A` and `B` have a clash, then `M` is rejected. You have written `M` in your new syntax (which I have not yet internalize), and so there's not necessarily trouble here, but I'm worried.

> 1.2.4   Module export specifiers

This section means that all `module A` exports today would have to be rewritten to `module A(..)` to retain their intended meaning. (The new form would also export `A`, which is new. But I'm more worried about `A`'s contents.)

> Local modules cannot import file modules.

But I think this would be OK and useful:

```hs
module Top where

import qualified A.B

module Local where
  import A.B
```

Right?

> (where <name> is a non-dotted module name).

(in the Module Aliases section) Why this restriction?

>  ```hs
>    module MyPrelude (module Data) where
>
>    import qualified Data.ByteString.Lazy as BL
>
>    module Data (module ByteString) where
>        module ByteString = BL
>    ```

I believe you could also say

```hs
module MyPrelude (module Data) where

import qualified Data.ByteString.Lazy as BL

module Data (module ByteString) where
  module ByteString ( module BL(..) ) where
    import BL
```

Acutally, my version also exports the internal `BL`. But otherwise it should be like yours. Does this reduce the motivation for module aliases? I don't see them as quite necessary.

> typeclass instances

Following the design of file modules, I think instances are always exported and always in scope. We thus don't have to associate instances with file modules: they are just exported from their local module and then transitively exported from their file module.

> But this would be somewhat unhygienic.

Typeclass instances are always somewhat unhygienic. :)

> 1.6.4   Do we need different syntax for "local" imports?

I think we do. Consider

```hs
-- A.hs:
{-# LANGUAGE ExportModuleNames #-}
module A ( module B ) where
  module B where ...

-- A/B.hs:
module A.B where
  ...

-- C.hs:
module C where
import A
import A.B   -- what does this mean?
```

On the surface, we could just say "it's ambiguous", but that's not quite good enough. We need the file-dependency analyzer to know whether `C.hs` depends on `A/B.hs`.

## Order dependence

And that example makes me worried about order dependence of local module imports. (This same problem exists in my proposal.)

If, in the middle of a module, we have

```hs
import M1
import M2
```

what does that mean? (No ambiguity around file dependencies here; we're in the middle of a module.) Perhaps `M1` exports `M2`. Or perhaps `M2` exports `M1`. Perhaps both! So I struggle what to make of this rats' nest. Maybe we declare that, within a module, any `import`s are processed top-to-bottom? That's a little out of the spirit of Haskell, but otherwise reasonable. With that extra rule, the first `import` is unambiguous. If `M2` is already in scope and `M1` exports `M2`, then the second `import` would be ambiguous. Even if `M2` exports a module `M1`, then the first `import` remains unambiguous. Further references to `M1` would be ambiguous, however.

> However, by analogy with the export specifier C(..), this should also export M. I'm not sure if this is what we want, and if we don't, then perhaps we should have a different syntax.

I've always been a little unhappy about the current status quo on this. I don't have any concrete suggestions. My proposal suffers from the same problem.

> 1.7.1   Recursion

I'm afraid I don't understand this section. Could you include some examples?

## Scope of proposal

Do your modules affect compilation processes other than naming? Specifically, can definitions one local module be seamlessly mutually recursive with definitions in another local module? Would local modules affect e.g. Template Haskell's stage restriction?

## Conclusions

One drawback of this proposal is that it is not backward compatible. That is, enabling the new extensions changes the meaning of existing constructs. This should be mentioned explicitly.

I identify a few points of difference with my proposal:

* Multiple modules with the same name are not merged in your proposal; they are ambiguous. My proposal merges them.

* My proposal goes into a few twists and turns to retain the current behavior of the `module M` export item. Yours chooses backward incompatibility (but much simpler exports) instead.

* My proposal has a few extra bells and whistles (`import`ing a local module at its definition, anonymous local modules, automatic modules in type declarations, maybe a few more). Yours is more bare-bones (except for module aliases, which strikes me as an inessential bell/whistle). Your proposal is correspondingly a good deal simpler than mine.

Let's strip all the extras away from both proposals. I think in that state, mine favors backward compatibility and ease of use (i.e. merging of modules); yours favors a more principled approach, where modules are resolved more like other entities.

I can't say I have a clear preference. In my view, they're both good. Maybe mine has a bit of bloat, but this bloat is easy to remove. I'd be happy, say, to submit both to the committee (in due course) and see what comes of it.

Thanks for writing this up!

michaelpj:
Thanks for the detailed comments, Richard. You've got a lot of good points, so I'm going to go through them thoroughly.

> While I know it's obscure, I'm worried about interactions with -XPackageImports.

Thanks for pointing this out. I think this should be handled like the "ambiguous as" case: the two imported modules should merged in the base proposal (to retain backwards compatibility), and give an ambiguity error with a hypothetical `NoModuleMerging`. Users can always work around this by using `import as` to pick unambiguous names for the two modules.

> We must tread carefully here. Today, if I have module M (module A, module B) and A and B have a clash, then M is rejected.

Ah, good point.

Module merging is not a hard problem (you just take the union of the exports of the arguments!). I should just define a module merge operator: it's well-behaved and monotone in both arguments. I was trying to avoid doing so, but I think we can just have it as a "semantic" operator that we don't necessarily expose in Haskell syntax.

Then all cases of module merging should just use that operator.

> This section means that all module A exports today would have to be rewritten to module A(..) to retain their intended meaning. (The new form would also export A, which is new. But I'm more worried about A's contents.)

Correct! That's why it's behind a language extension.

> But I think this would be OK and useful (importing a file module brought into scope qualfied)

You're quite right. I think I was wrong to key this restriction off the *module* that's being imported. What I really want to say is that local imports cannot trigger the "look up an external module from a package" behaviour, because I want to still be able to understand all the external interactions by looking at the top-level header imports.

If we adopted your proposal of having a new syntax for the "new style" of imports (`import module`), then we could say that:
- `import` *always* targets a dotted name, and *never* a qualified name, and is only allowed at the top level.
- `import module` *always* targets a qualified name, and *never* a dotted name

I actually really like this: this gives us two clearly syntactically distinct types of import that do two different kinds of thing. That corrals dotted names even further, and allows us to determine at parse time what kind of names we're dealing with. I think this would clarify the dotted name part of the proposal significantly, which I think is currently the most confusing part.

(This might entail biting the bullet that you can't do `import M as A.B; import module A.B`, but I'd accept that. Perhaps another language extension is needed to disable importing as a dotted name!)

> Why this restriction? (can't define new modules with dotted names)

The goal is that dotted names should be associated with file modules *only*. So we shouldn't allow them anywhere else.

> I believe you could also say (example without modules aliases). Does this reduce the motivation for module aliases? I don't see them as quite necessary.

You're quite right: module aliases are very inessential to this proposal, and are purely a quality-of-life feature. The point was mostly to show that this model allows us to do things like aliases easily.

> Following the design of file modules, I think instances are always exported and always in scope. We thus don't have to associate instances with file modules: they are just exported from their local module and then transitively exported from their file module.

Note that in my proposal names exported by a local module are not necessarily in scope in their enclosing module, unless that module imports them. So there is no transitive export to save us!

I think what I'm proposing is that all modules do the equivalent of `import M ()` for all of their locally defined sub-modules. This ensure that all instances get sucked to the top level, but nothing else unless it's explicit. Hideous, but I'm not sure what else to do.

> I think we do [need new syntax for new imports]. [example with ambiguity at the top level]

I agree, this combined with the earlier point I've come around to thinking that new syntax would be actively good.

> Order dependence

There is no order dependence! I think this is a very important property and we should *not* sacrifice it (but: what about TH?). My Datalog version has no ordering, if you want evidence.

The way we would resolve your example is by doing multiple passes in which we discover increasingly more information. Suppose that `M1` exports `M2`:  initially we would discover that `M1` is in scope but not `M2`. Then we would add the exports of `M1` to the scope, which would reveal that `M2` was now in scope. Then we would add the exports of `M2` to the scope.

Where does this "multi-pass" behaviour arise from? The answer is as usual: recursion. `refersTo` and `inScope` are mutually recursive, because
- What a qualified name refers to depends on what's in scope, and 
- What's in scope depends on what is imported, and
- What is imported depends on what the qualified names targeted by imports refer to!

But this dependency is all monotonic (all that can happen is that we discover *new* bindings, and we just check for ambiguity at the end), so we can just compute the fixpoint in the usual way. This happens transparently in Datalog but would need to happen explicitly in another implementation.

This is an important point, though, because it illustrates that there's an important recursion *apart* from the recursion that we get due to recursive modules.

> Recursion

Yes, this is a bit compressed. I'll expand.

> Do your modules affect compilation processes other than naming? 

No. In particular, we retain the property that you can convert the program into a module-free program.

> Specifically, can definitions one local module be seamlessly mutually recursive with definitions in another local module? 

Yes. It's purely about renaming.

> Would local modules affect e.g. Template Haskell's stage restriction?

No. I believe the stage restriction to be more about *compilation units* than *modules* per se. This proposal has nothing to say about compilation units.

> One drawback of this proposal is that it is not backward compatible. That is, enabling the new extensions changes the meaning of existing constructs. This should be mentioned explicitly.

I thought this was clear, but I'll say more. I'm certainly not proposing that all the extensions become enabled by default. The core change ("Modules as entities") *should* be backwards compatible, and I propose that it be effected unconditionally.

Crucially, everything *is* backwards-compatible for consumers of a module that uses these extensions.

> I think in that state, mine favors backward compatibility and ease of use (i.e. merging of modules); yours favors a more principled approach, where modules are resolved more like other entities.

I agree with this, although I've tried quite hard to ensure that my proposal can be enacted in a maximally backwards compatible way, with only those who want the new behaviour opting in to it.

goldfirere:
> doing multiple passes

Suppose I write

```hs
module Top where

module A where
  module A where
    x = 5

import A
```

Would that trigger an ambiguity error? I suppose I could say `import Top.A` to avoid the problem. (I had originally thought there was no way out.)

But how can you do the ambiguity check at the end? Suppose module name `M` is ambiguous and I write `import M`. Then what have I imported? We have to stop right now and report an ambiguity.

> backwards-compatible

Sorry if I didn't express myself clearly. I understood that your proposal would continue to accept all programs that are currently accepted. But enabling your extensions would cause some programs to break.

I think we're basically in agreement about where we stand.

michaelpj:
> Would that trigger an ambiguity error?

Lovely example, thanks. Yes, in my proposal this should trigger an ambiguity error.

`Top.A` actually doesn't help you here, because Top has two `A` bindings: the one from the module definition, and the one from the import. So `Top.A` is also ambiguous. Note that the user is under no obligation to `import A`, they can use `A` qualified and `A.A` is perfectly unambiguous (local definitions shadow external ones, so the outer `A` is shadowed by the inner `A`). So I'm not too bothered if this doesn't work.

However, the *right* solution in my opinion is for local definitions to *shadow* imported definitions. This was the case when I implemented a system like this in Semmle QL, and avoids this sort of thing without adding any more non-monotonicity. Unfortunately, that would be quite a big departure from how things work in Haskell today, so I didn't propose it.

> But how can you do the ambiguity check at the end? Suppose module name M is ambiguous and I write import M. Then what have I imported? We have to stop right now and report an ambiguity.

We do not have to stop! That's the beauty of the relational model (and the `hsmods` paper uses this too): ambiguity is no particular problem, it just means your set of possible answers has more than one entry. You "import both of them" i.e. add entries to the `imports` relation for both.

This may of course make more names ambiguous, but that's fine.

We probably want to "cut" this process, i.e. stop and make sure everything is unique/issue errors, at various points to improve error reporting. However, every time we do this we introduce non-monotonicity, which would have consequences for recursive modules.

One safe example would be to resolve all module names, issue ambiguity errors, and then not go any further until we have resolved those. That would prevent "cascading" ambiguity, for example (assuming `NoModuleMerging`):

```
import Data.ByteString as BS
import Data.ByteString.Lazy as BS

-- Could be `Data.ByteString` or `Data.ByteString.Lazy`, so we put the exports for 
-- both in scope.
-- Also, we'll issue an ambiguity error at the end.
import module BS 

-- Both `Data.BytesString.ByteString` and `Data.ByteString.Lazy.ByteString` are
-- in scope, so this is also ambiguous if we kept going
foo :: ByteString
```

Uncontroversially, I think we should also "cut" the process at file module boundaries. This is something we'd need to relax if we did do mutually recursive file modules.

goldfirere:
> local definitions shadow external ones

But you can't access outer definitions qualified this way, can you? For example:

```hs
module Top where

x = 5
module A where
  y = 6

z = A.x
```

Would that work? I wouldn't expect it to.

Also, what about this:

```hs
module Top where

x = 5
module A (module Top) where
```

Does this mean I can say `Top.A.Top.A.Top.A.x`? Maybe that's your "recursive modules" bit. What if we remove the export spec?

And here's another fun one:

```hs
module Top (x) where   -- NB: export spec

x = 5
module A where
  y = 6
z1 = Top.A.y
z2 = A.Top.A.y
```

What do we think here?

And about ambiguity, what about this:

```hs
module Top where

import qualified M1 as A (x)
import qualified M2 as A (y)

import A

z1 = x
z2 = y
```

Note that both `x` and `y` can refer to only one entity each. But for both to be accepted, the `import A` has to do two different things. Or is this allowed?

michaelpj:
Note: some of the things I said about dotted names in a previous comment were wrong, I've edited it to reflect that. Dotted names are still difficult, I need to think more about them and try to present some more coherent options.

----

> But you can't access outer definitions qualified this way, can you? For example:

Outer definitions are like imported definitions in that sense: they're not automatically re-exported unless you say so. I think your example *should* work if we had `module A (x, y)`, for example.

> Does this mean I can say Top.A.Top.A.Top.A.x? ... What if we remove the export spec?

Yes to the first, no to the second (same reason as the previous point).

> And here's another fun one ... What do we think here?

`Top` does not export `A`, so all the `Top.A` uses are errors. `A` also doesn't export `Top` (as above), so `A.Top` is also an error.

> And about ambiguity, what about this:

Yes, this is an interesting case. If just run the whole process, then when we do our ambiguity check, we would end up saying that:
- The use of `A` is ambiguous.
- The uses of `x` and `y` are not ambiguous.

Now, you *could* have a system that accepted this. Since the modules only exist to resolve the names of terms and types, as long as those are unambiguous it's fine, right? However, I think this is never really going to be what anyone wants, and so we should indeed give an error that `A` is ambiguous.

I also think we should do this *before* trying to say anything about `x` or `y` (this is what I was saying above about "cutting" the process). Term and type names depend on module names, but not vice versa (but: what if e.g. records had associated modules?0, so we might as well sort out the first properly before we move on to the second.

goldfirere:
> `Top` does not export `A`, so all the `Top.A` uses are errors.

But this works today:

```hs
module A () where  -- NB: empty export spec

x = 5
y = A.x
```

That is, we can use a name qualified by the module we're defining, even if that module doesn't export the name. Weird, but true. So, would the following work:

```hs
module A () where

x = 5
module B where
  y = A.x
```

I would assume it would have to, by monotonicity. Sadly, the following fails today:

```hs
module A (x) where
x = 5
```

```hs
module B (x) where   -- NB: exports x
import A
y = B.x
```

That is, we can use the defining module as a qualifier only for those entities it defines. This seems somewhat at odds with your proposal, sadly.

michaelpj:
> That is, we can use a name qualified by the module we're defining, even if that module doesn't export the name. Weird, but true.

> That is, we can use the defining module as a qualifier only for those entities it defines. This seems somewhat at odds with your proposal, sadly.

The joys of trying to reinterpret an existing system... :see_no_evil:

I had one idea, but I don't think it works: essentially, we could allow some modules to "see" all the bindings inside other modules, rather than just the exported ones. The reason I think this won't work is that I expect this works today:

```haskell
module A (B.x) where
import qualified B

x = ...
y = A.x -- refers unambiguously to the local declaration
```
However, if we were looking at *both* the local bindings and the exported ones, this would be ambiguous.

Another ugly solution: for every top-level module, we add a shadowing definition of a module which contains only the local definitions. We add a language extension `NoStrangeLocalQualification` to turn this off.

e.g. we would end up with things like this
```
module A () where
module A (x) where -- shadows the outer definition

x =5 

module B where
    y = A.x -- refers to the shadowing definition
```

proposals/0000-first-class-modules.rst
neongreen: > Yes, I think my philosophical position is that dotted names shouldn't exist, and dot should always mean accessing a module.

Do you have a sketch of how this would look?

E.g. [servant](https://hackage.haskell.org/package/servant) has `module Servant.API` and `module Servant.API.Alternative` (though not `module Servant`). How would module names, file layout, and module layout look in this situation, in an ideal world?

michaelpj: Does the section "How could we make dotted names less special?" answer your question?

In short: defining a file module with a dotted name would give rise to a "virtual" parent module containing it. In an ideal world, such modules would be created per-package and be ambiguous if they appeared in multiple packages, and we would use `PackageImports` to disambiguate them. However, this would be very annoying in practice (`Data` would be massively ambiguous), and so probably we should just merge all such "virtual" parent modules across packages. (This inconsistency is further evidence to me that this position is not fully baked.)

neongreen: Okay, what about ByteString and ByteString.Lazy, then? In this situation you can't create a virtual module because it's already taken. Do you just merge the real top-level Bytestring module and the virtual one arising from the existence of ByteString.Lazy?

goldfirere:
Let me ask a controversial question: why is making modules entities an improvement? I agree with you that making modules to be entities is theoretically satisfying and adds uniformity to our understanding of entities. But how will it improve life for Haskellers?

If we examine the core of our two proposals (and my proposal could do much better at identifying its core), I think the salient difference lies in export specifiers. My semantics are quite a bit more complicated than yours -- but my approach keeps with the precedent of not having modules as entities. It is also backward compatible, even with `-XLocalModules` enabled.

Do you agree with this being the salient difference?

Another difference is that, under your proposal, `A.x` has the same meaning inside `A` as outside `A`. That's really nice -- but different from what we have today.

How shall we proceed from here? I'd be happy if we submit both to the committee at one go, as we definitely don't want to accept both proposals. I'm also happy to wait if more ingenuity is cooking at your (for any "you") end. I'm truly in no rush here -- just don't want to wait around for the sake of waiting around.

michaelpj:
> Let me ask a controversial question: why is making modules entities an improvement? I agree with you that making modules to be entities is theoretically satisfying and adds uniformity to our understanding of entities. But how will it improve life for Haskellers?

This is an excellent question. My answer is twofold: 
- By their fruits shall ye know them; and 
- Accept no substitute for saying what is really going on

For the first:  we expend a lot of ink reinterpreting what modules are and what they do, and adding epicycles to ensure we have backwards compatibility. But then we get the payoff: the specification of e.g. local modules is is easy.

My intuition is that this will keep paying off: that because the bones of the new system are cleaner, adding more features will be similarly simpler. This is difficult to justify (although I included module aliases as an example), but I think it is a design intuition that we commonly follow in the Haskell world.

For the second: I think that our intuitions about modules (and the way we talk about them) match much more closely to modules-as-entities than nothing-but-qualified-names. But that means that we have to translate our ideas into the world of nothing-but-qualified-names to get things done.

This bites us as ordinary Haskellers. Every now and again I have to go and look at the Haskell report to work out how to translate what I want to do into terms that Haskell's current module system will understand. I do think there's a pedagogical and conceptual benefit for our users in having the system match the way they think. (Obvious caveat: this is entirely unrigorous and based on my own impressions.)

It also bites us as system designers. I think "local modules" is actually a pretty simple concept, and part of the goal of this proposal was to show that if we set up our system the right way, then it can be simple to implement as well. In contrast, your proposal has to squidge the concept into a conceptual model that doesn't quite fit, which is harder work.

> I think the salient difference lies in export specifiers. 

Hopefully the above has conveyed that I think there's more at stake. Export specifiers are a visible consequence, but I would still care if we had a language with no export specifiers.

> [My approach] is also backward compatible, even with -XLocalModules enabled.

I have the same goal: if you turn on `-XLocalModules` but also leave on all the backward compatibility widgets I keep adding, then you should get a sensible system! But I would encourage people to turn them off, of course :)

> How shall we proceed from here?

I'm not happy with this proposal. In particular, I want to have a core proposal that is *completely* backwards-compatible (and thank you again for your examples, which have been very helpful in pointing out some dark corners I need to cover). I also want to have a better story about dotted names. Unfortunately, I think there are a few directions I could go with that, and I don't currently know which one is the best.

I will try and do some more work on this over the next month or so. I'm also in no hurry to submit anything to the committee - after all, neither of us are volunteering to implement our proposals (yet). And it might even be sensible to await a prototype implementation before accepting, since that often throws up issues.

proposals/0000-first-class-modules.rst
michaelpj: Yes, in that situation we would have to merge them. Thanks, that's a pretty ugly case, and pushes me away from this idea. 

(That is: while I would *like* dots to always mean accessing a module, I think it may be impossible for Haskell and we'll have to have some kind of uneasy coexistence with dotted names regardless.)

goldfirere:
>  My answer is twofold

... and convincing. But your stance is most convincing at its most pure: no epicycles. That is, maybe it's better to embrace a coherent design and not even design backward-compatibility widgets. The key back-compat question is this: can the lost expressiveness be recovered by an easy-to-write construct that is both backward- and forward-compatible? If the workaround is easy -- and would compile in older versions of GHC, then a few programs that no longer compile is OK. I do worry that this answer won't come easily, as `module A () where loopy = A.loopy` is accepted today, struggles with your approach, and does not have an obvious workaround (if another `loopy` is in scope).

Actually, that opens up a question. Suppose I have

```hs
module A (y) where

import B (x)

x = 5
y = A.x + B.x
```

This compiles today. How would I write it under your proposal? I can't write `A.x`, because `x` is not exported from `A`. Just writing `x` is ambiguous. Unless locally defined names shadow imported ones -- at which point we're OK again technically, but perhaps confusing. We could also forbid the local `x` in this scenario; it's purely internal, so requiring a renaming is feasible. Note I'm not looking for back-compat here -- just expressiveness.

Regardless, happy to wait until you have time to polish this up.

michaelpj:
> But your stance is most convincing at its most pure: no epicycles. That is, maybe it's better to embrace a coherent design and not even design backward-compatibility widgets.

The problem with this is that modules-as-entities will probably require a fair amount of changes to GHC inside the renamer. I don't want GHC to have to support two *actually* different models, I think it's much better if we just have one and can simulate the old one with the new one.

> [more examples]

I do *not* think that we should try and make it so that people can always refer to e.g. local definitions unambiguously. People will write things where the names are ambiguous (as they already can), and there are lots of ways to get around it. I'm more concerned with having clear and consistent rules about what's ambiguous.

So my answer is: if it hurts, don't do it. I don't think it's a problem to require people to write this:
```
module A (y) where

import qualified B (x)

x = 5
y = x + B.x
```

(Of course, that example *is* a problem for my backwards compatibility shims, but I think you'd already shown me that in https://github.com/ghc-proposals/ghc-proposals/pull/295#issuecomment-558265324)

akhra:
Doesn't that solution still rule this out?
```haskell
module A (C.y) where

import qualified B (x)

x = 5

module C (y) where
    x = 6
    y = A.x + B.x + x
```

michaelpj:
> Doesn't that solution still rule this out?

Yes. Then you've made life unnecessarily difficult for yourself and I suggest Not Doing That.

As an aside, you should only be able to export things that are in scope, and `C.y` is not in scope outside `C`, but your example works perfectly well with `module A ()`.

akhra:
> I suggest Not Doing That

This is fine in a world where you get to design all your modules from the ground up, but I can imagine situations like this emerging from relocation and refactoring all the time.

An early migration step for a lot of projects will be consolidating things that are already written as separate modules; and in an industry environment, it's not always viable to do things like that in one go. The more we can enable turning a module into a submodule without renaming things, the better.

In short, sometimes objectively bad style is still a good tool. Not having this isn't a showstopper, but I'm very sure it will slow adoption.

michaelpj:
> This is fine in a world where you get to design all your modules from the ground up, but I can imagine situations like this emerging from relocation and refactoring all the time.

Sure, in much the same way that naming ambiguity from imports etc can emerge when you do refactoring. But we have lots of tools for dealing with that.

As a bazooka, if you really want to have two modules with exactly the same names in one file module, you can always do this:

```
module A where

module B where
    x = ...
    y = ...

module C where
    x = ...
    y = ...
```
and indeed, if you're refactoring by merging multiple modules into one, that's probably the best place to start, as it mirrors the namespacing of the previous situation the most closely.

michaelpj:
I've pushed a substantial revision of this proposal. Highlights:
- Separate syntax for non-external imports (need better terminology...).
- Embrace dotted-names-as-simple-names, with the corresponding increase in ambiguity. This is disappointing, but I can't make any of the other approaches work.

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,733 @@
...
+
+::
+
+   data QualName = Qualified ModuleName Name | Simple Name
+
+We propose to replace this with a type like:
+
+::
+
+   data QualName = NonEmpty Name
```
goldfirere: I think you want `type` here.

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,733 @@
...
+a dotted name if it is a file module).
+
+If a module is imported ``qualified`` then the exported bindings of the target
+module are not brought into scope.
+
+If a module is imported ``as`` a simple name, then the target module is brought
+into scope bound to that simple name instead.
+
+Module export specifiers
+~~~~~~~~~~~~~~~~~~~~~~~~
```
goldfirere: You might want `^^^` instead of `~~~` here.

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,733 @@
...
+
+The behaviour of the ``module`` export specifiers are changed as follows.
+
+- ``module M`` *changes* to exports the name ``M`` as a module binding.
+- ``module M(a, b, .. c)`` exports ``M`` and the listed in-scope bindings.
+- ``module M(..)`` exports ``M`` and all of its exported bindings. [#]_
```
goldfirere: What if `M` is the current module? Then this definition is badly recursive.

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,733 @@
...
+The extension ``FlatModuleExports`` is added and enabled by default. It has the effect
+that the ``module M`` export specifier has the behaviour given in [hs2010]_. [#]_
+
+.. [#] See `Flat module exports - reasoning`_ for reasoning.
+
+Import module merging
+^^^^^^^^^^^^^^^^^^^^^
+
+The extension ``ImportModuleMerging`` is added and enabled by default. It has the effect that
+when multiple modules are imported ``as`` the same name, we merge the modules. [#]_
```
goldfirere: It is sadly possible to have

```
import A
import B as A
```

The first `import` does not import `A` `as` anything, so this sentence might want to be generalized a bit, if I suppose your intended meaning correctly.

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,733 @@
...
+.. [#] See `Import module merging - reasoning`_ for reasoning. 
+   multiple modules are defined with the same name?`_ for discussion
+   about what the primary behaviour should be.
+
+Privileged local module access
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The extension ``PrivilegedLocalModuleAccess`` is added and enabled by default. It has the effect
+that for every file module ``M``, we add a new module declaration with the same name inside ``M``,
+which exports all and only the local declarations of ``M``. [#]_
```
goldfirere: But, presumably, not `M` itself. That is, this shim module named `M` does not export `M`. Otherwise, we might say

```
ones = 1 : M.M.M.ones
```

and I don't think we want to allow that (it's not allowed today).

goldfirere: Also, I can write this nonsense today:

```
module M where
import A as M
```

with merging semantics. So this might want to be taken into account in your backward compatibility allowances...

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,733 @@
...
+
+Module imports
+^^^^^^^^^^^^^^
+
+We add the language extension ``ImportModule``, which introduces a a new
+syntax ``import module`` [#]_ for importing modules in other contexts.
+The target of an ``import module`` statement is a *qualified* name, resolved
+as normal for qualified names, and will *not* resolve to an external module as
+is possible for normal ``import``s. [#]_ . ``import module``
+otherwise behaves as ``import``.
```
goldfirere: I don't think this is exactly what you mean. Here is an example:

```hs
module Top where

import qualified A

import module A.Local (..)
```

The key question: is `Local` in scope here? A normal `import` brings the named module into scope, but I don't think you want `import module` to do that. (That is, I *don't* think `Local` should come into scope.)

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,733 @@
...
+Additional features
+~~~~~~~~~~~~~~~~~~~
+
+The following extensions add features beyond existing Haskell, building upon the previous changes.
+
+Module imports
+^^^^^^^^^^^^^^
+
+We add the language extension ``ImportModule``, which introduces a a new
+syntax ``import module`` [#]_ for importing modules in other contexts.
```
goldfirere: Where is this syntax allowed? Presumably, not intermixed with normal `import` statements. But is it allowed in `let` and `where` clauses?

proposals/0000-first-class-modules.rst
michaelpj: Good question! I think this is in fact fine with the right kind of recursive implementation. We can get the same kind of issue with recursive modules:
```
module A
    import B
    x = 1

module B
   import A
```

(note the lack of export specifiers, you could also give them `B(..)` and `A(..)` respectively.)
My implementation will say that both `A` and `B` export `x`. In the same vein, `M(..)` when `M` is the current module will just have no effect (it simply never tells us any new `exports` facts).

proposals/0000-first-class-modules.rst
michaelpj: Sigh. Yes, it sounds like this needs to cover all cases where module names are introduced as part of import statements.

proposals/0000-first-class-modules.rst
michaelpj: `M` is not a local declaration of `M`, but rather of the "enclosing scope". So indeed it is not included.

(I realise that I have been implicitly acting as though there is a top-level file scope in which the file module declaration lives, but I should make that explicit).

michaelpj: Thanks - another one to handle.

I am gradually realising that *every* place where module names are currently introduced is a place where this proposal requires a backwards compatibility shim. I think that's quite a wart, and the more backwards compatibility shims we need, the less appealing it becomes. Perhaps "all-or-nothing" is better after all.

proposals/0000-first-class-modules.rst
michaelpj: Why not? I think it should behave like `import` as much as possible. I'd rather take a slightly odd piece of behaviour than the inconsistency.

proposals/0000-first-class-modules.rst
michaelpj: I guess the reason not to allow them in the initial block of `import` statements is to not break scanners that make assumptions about the structure of modules? I do think it would be good to allow them to be at the file-module level. But we could ban them in the initial block of `import`s.

Initially I had thought that we could have these only at the module level, but allowing them in other scopes would be very natural too. What about local module declarations in `let` and `where` clauses too?

michaelpj:
> The current rule that all instances are always exported works just fine with your proposal, I believe. Is this true? So I don't see a reason to forbid instances in local modules...

I don't think so: things which are exported by local modules are *not* necessarily exported by their *enclosing* modules. So for example:

```
module A where -- does not export the instance, as it is not in scope (B is in scope, but not its exports)
    module B where -- exports the instance
        instance Foo T where
```

So it would be possible to use parts of `A` without having imported the instance. We could have a rule whereby the nested instances get sucked out magically and exported by the enclosing module, although that's a bit ugly.

The thing that actually made me want to ban them was looking at the Local Types proposal. It seems quite possible to me that we'll come up with some sensible meaning for locally scoped types/instances. In that case we could potentially use the same strategy for those in local modules. I don't want to preclude that choice by giving them a behaviour for the sake of it.

The issue would be made even sharper if we allowed local module declarations in `let` or `where` clauses - then even the module is not in scope in the enclosing module, let alone any instances that it declares.

----

> But this proposal has now become heavier, proposing 6 new extensions -- three on by default and three off by default. While each is well motivated in the context of the overall theme here, I'm still not leaping up and down thinking this is The Right Answer.

I agree. I've gradually come to think that this actually cuts quite hard against the grain of the way that names work in Haskell today. So this is *a* way that we could go, but I had hoped that it might be a) easier and b) unifying. But alas, I think it is neither.

I still like it as a "revolutionary" proposal, but that's a much harder sell than an incremental one. And then perhaps I should ditch all the backwards compatibility wrangling and just push the full version, changed behaviour and all.

proposals/0000-first-class-modules.rst
goldfirere: Consider

```hs
module Top where

import Local     -- some top-level module perversely named "Local"
import A

import module A.Local (..)
```

Now, any `Local.` prefix is ambiguous. I could use `as` to rename the `Local` top-level module, but I don't want to, as I really don't need `A.Local` in scope at all.

More generally, I wish Haskell had better ways of bringing components of structures into scope without bringing the structures themselves.

Maybe `import module A.Local (..) hiding (Local)` would do it.

michaelpj:
> But this is not true for instances. All instances in scope are always exported, even if those in-scope instances are in-scope via importation.

But in my example, `B` is not imported into `A`. It is *defined* there, but not imported. The names inside it are not in scope, although they can be accessed qualified (which is, now that I think about it, another reason why the instances would need to be "around" in the outer scope).

So I still think the current rule for how instances works isn't sufficient. We'd need to say something like: "An instance is exported if it is defined in a module *or any module defined within it*". Which would be okay, but is another odd special case. Given the possibility that we'd be stealing something that could get more sensible semantics later, my inclination is to just not allow it.

> The problem is that revolutionary proposals are very hard to evaluate without a working implementation that can check for real-world breakage.

Realistically, that's not something I'm going to do any time soon, so I'm not really sure what to do with this proposal. Perhaps I should just tidy it into its best "revolutionary" form and leave it for some future keen people.

proposals/0000-first-class-modules.rst
michaelpj: `import module A.Local as SomethingElse`?

I don't disagree that this is annoying, but I think this is an annoyance about the way `import` works, which I think is orthogonal to this proposal.

goldfirere:
> The names inside it are not in scope, although they can be accessed qualified (which is, now that I think about it, another reason why the instances would need to be "around" in the outer scope).

And this exposes another key difference between our proposals. I would have said that local modules *do* export their contents -- but these contents are exported qualified (by default). Whereas in your terminology, "exported qualified" makes no sense. Incidentally, it would seem this notion has implications for a top-level import like

```
import qualified Furble
```

According to your new semantics (if I understand correctly), this imports `Furble` but not `Furble`'s contents. But what if `Furble` exports instances? In today's Haskell, these instances are imported. So I think you might need a special case to deal with this, even if you eschew the special case for locally defined instances by disallowing them.

> Realistically, that's not something I'm going to do any time soon

I didn't mean to imply that it was... just that it's hard to evaluate the effectiveness of a revolution until afterwards (both in programming language design and beyond).

I do support your plan of rephrasing this to be revolutionary. Then, perhaps, the committee can decide where it wants to go with this all.

michaelpj:
> And this exposes another key difference between our proposals. I would have said that local modules do export their contents -- but these contents are exported qualified (by default). Whereas in your terminology, "exported qualified" makes no sense. Incidentally, it would seem this notion has implications for a top-level import like ...

Right, this is something I hadn't thought out fully. My vague plan was to say something like:
- Instances are handled on a per-*compilation-unit* level.
- Normal-import-ing an external module also "links" in the compilation unit, which causes its instances to be available during typechecking.

Indeed, the more I think about it, the more I think that instances are going to need special rules. Even the statements we've been making about instances "being in scope" or "being exported" don't make sense in my system: only *bindings* are in scope or exported, and instances aren't visible under a name binding, they're just "around". The Haskell report does talk about instances being imported or exported, but without names - very odd.

So we perfectly well could have special rules for pulling local instances to the top level, which would be consistent with the idea that instances are per-compilation-unit. But that rather nixes any other ideas for local instances, which is why I'm hesitant.

proposals/0000-first-class-modules.rst
```diff
@@ -0,0 +1,733 @@
...
+as `Data.Bytestring`. Then we need to write:
+
+::
+
+   module MyPrelude (module Data) where
+
+   import qualified Data.ByteString.Lazy as BL
+
+   module Data (module ByteString) where
+       module ByteString = BL
```
vdukhovni, line 373: How does this differ from just importing "DBL" as "DB", already possible today?
```
$ ghci
GHCi, version 8.6.3: http://www.haskell.org/ghc/  :? for help
Prelude> import Data.ByteString.Lazy as Data.ByteString
Prelude Data.ByteString> :t Data.ByteString.toStrict 
Data.ByteString.toStrict
  :: ByteString -> Data.ByteString.Internal.ByteString
```
Editing the tree at intermediate nodes looks like a rather tricky business to me, the dots in the middle of module names don't imply any sort of parent/child or scoping relationships today, and introducing that could make things rather more fragile...

michaelpj: > the dots in the middle of module names don't imply any sort of parent/child or scoping relationships today, and introducing that could make things rather more fragile...

Yes, well, that's precisely the thing that this proposal wants to change. In what way would this make things fragile?

michaelpj:
This proposal is functionally dormant. Thanks to @goldfirere 's impressive ability to generate fiendish corner-cases, I've been convinced that this isn't viable as an incremental proposal. 

For me, that kills a lot of the motivation: I don't believe that a revolutionary version of this proposal would be (or indeed should be) accepted. The changes are too extensive, and the marginal benefit is too small. It's too late for Haskell - at best this proposal could be a set of notes on how you might do it differently in a different programming language. (Not to mention that the implementation would likely require extensive surgery that would be required on the innards of GHC.)

As such, I think that if the goal is improving Haskell today, we are better served by something like https://github.com/ghc-proposals/ghc-proposals/pull/283.

chreekat:
Although this is now dormant, I'll still make a meta-proposal regarding first class modules.

Breaking the "compilation unit = file = module" monolith is obviously really hard. Trying to break it up by changing the definition of a module will require a lot of complexity to maintain backward-compatibility.

One solution to this dilemma is to just leave modules alone, and introduce an entirely new first-class entity. Call it an `object`. Then the strategy would be to introduce objects, describe how to map between modules and objects, and slowly expand the usefulness of objects until they cover all of module's current use cases.

This might allow a staged evolution that minimizes backward compatibility problems, particularly if it's really easy to define a module in terms of an object.

Ericson2314:
If we don't have a incremental path to the revolutionary thing, I'm a bit tempted to do nothing at all, and exclusively focus on cleaning up GHC behind the scenes to return to this later.

PR #306, opened 2020-01-14
errors-as-values proposal

alpmestan:
[Rendered](https://github.com/alpmestan/ghc-proposals/blob/alp/error-values/proposals/0000-error-values-proposal.rst#errors-as-values).

I realize this is not exactly a typical GHC proposal, as it's about a change almost entirely internal to GHC. But the "almost" is very important in this case, because a major motivation for this line of work is to make the life of GHC API users a lot easier, whenever they're interested in doing anything non-trivial with the errors that GHC emits. With this in mind, `ghc-proposals` seems (to me) like a great place to discuss the proposal and get some feedback, including from GHC API users.

This is part of a larger plan, with #307, to provide GHC API users with easier solutions to problems that they are facing when dealing with the errors that GHC emits, especially in the IDE/tooling space. When "composed" together, these two proposals would give us proper ADTs to represent errors/warnings and a flexible way to consume the error documents that are derived from those ADTs, with the ability to embed structured data in the middle of documents and leave the final rendering of those bits to users.

### DIFF
--- a/proposals/0000-error-values-proposal.rst
+++ b/proposals/0000-error-values-proposal.rst
@@ -107,10 +107,10 @@ GHC subsystem. For example (simplified):::
 We could even split error types further if necessary, making it a
 slightly more elaborate/deep hierarchy. The exact shape of the said hierarchy
 has yet to be determined, as it will be best informed by staring at the
-error generation code that GHC has today. We would also provide a toplevel
-error data type, ``GHCError``, which would be a sum of the error types from all
-subsystems. This would allow us to store and more generally treat
-uniformly errors from different systems:::
+error generation code that GHC has today.
+
+We would also provide a toplevel error sum type, ``GHCError``, but only in the
+driver.::
 
     data GHCError
       = PsError ParseError
@@ -118,29 +118,41 @@ uniformly errors from different systems:::
       | TcError TypecheckerError
       | ...
 
-One can then define helper functions such as
-``notInScope :: OccName -> Name -> GHCError`` to be able to easily construct
-error values from within, say, the guts of a renamer function, without
-having to make the intermediate layers and wrapping visible there. We would
-create, manipulate and store ``GHCError`` values until the very last moment,
-when it is time to render the errors and report them. This would require
-implementing ``errorMessage :: GHCError -> ErrDoc``, and would be equivalent
-to all the ``ErrDoc`` building code that GHC has right now.
+Each subsystem would stick to its dedicated error type and the driver
+could then easily wrap those up appropriately to get a ``GHCError``. This would
+allow us to store and more generally treat uniformly errors from different
+systems at the top of the call stack (therefore avoiding potential cycles), and
+GHC API users would be free to handle the error values differently.
 
-Error consumers (the GHC program, GHC API users) would be presented with
-``GHCError`` values, and would be free to just call ``errorMessage`` on them
-to generate error message documents, or do something more interesting with
-some or all of the error values, using good old pattern matching to provide a
-specific interpretation for the errors of interest.
+We would quite want to likely define many helpful little functions to assist in
+the construction of error values, for each subsystem, e.g
+``notInScopeError :: OccName -> Name -> RenamerError`` or
+``occursCheckError :: Type -> Type -> TypecheckerError``, with some possibly
+requiring a suitable monadic context to gather extra information to stick in the
+error value. We may or may not want to move that logic to a dedicated place.
+
+We could then have an ``IsError`` class, defined as follows:::
+
+  class IsError e where
+    errorMessage :: e -> ErrDoc
+
+coming with instances for all the subsystem-specific error types, as well as
+``GHCError``, so that any value of one of those types can be used directly to
+generate an error document without having to resort to any kind of
+wrapping/unwrapping. This would all be equivalent to the entire ``ErrDoc``
+building code that GHC has right now. Most of that code would in fact simply
+move without undergoing any serious change.
+
+Error consumers (the GHC program, GHC API users) could be presented with
+``ParserError``, ``RenamerError``, ``TypecheckerError`` or ``GHCError`` values,
+depending on the parts of the API that are used. They would be free to just call
+``errorMessage`` on them to generate error message documents, or do something
+more interesting with some or all of the error values, using good old pattern
+matching to provide a specific interpretation for the errors of interest.
 
 For error producers, the main change is that the different subsystems will
-define error types and helper functions to build error values. In order for
-``GHCError`` to be able to refer to all the different error types, and for
-those types to use names from the module where ``GHCError`` is defined, we will
-quite likely have to introduce ``.hs-boot`` files to work around
-the import cycles induced by such an architecture. The exact details should be
-figured out at implementation time, but there might be a way to get away with
-just one ``.hs-boot`` file (must be confirmed).
+define error types and helper functions to build error values, along with the
+(existing) error document building code.
 
 It is important to note that ``errorMessage`` ties this proposal back with
 the existing system. Right now, GHC immediately emits error messages
@@ -155,12 +167,11 @@ as values of suitably defined algebraic data types, with all the
 expressions, types, contexts, suggestions and more stored in fields of
 those ADTs.
 
-If necessary, we could define a separate sum type for warnings and
+If necessary, we could define a separate hierarchy for warnings and
 update the definitions of ``ErrorMessages`` and ``WarningMessages``
 given earlier as follows:::
 
-    -- defined as a direct sum of the warning data for each warning, or
-    -- as a subsystem-driven hierarchy (like for errors) if required
+    -- defined as a subsystem-driven hierarchy (like for errors) if required
     data GHCWarning
       = UnnecessaryImport ModuleName
       | ...
@@ -174,15 +185,24 @@ given earlier as follows:::
 (The alternative being to just store ``GHCError`` values in both bags and
 augment ``GHCError`` with a constructor dedicated to warnings.)
 
+These types would only be used in the driver... except that the ``TcRn``
+monad and other pieces of code seem to be relying on ``Messages``. This
+suggests that we might have to either parametrize ``Messages`` on the
+concrete warning and error types, or replace those uses of ``Messages`` with
+an accumulation of values of the corresponding subsystem's error and warning
+types, essentially reimplementing the ``Messages`` machinery. The option of
+just using ``GHCError`` in all those places is dismissed because of the import
+cycles that this would come with.
+
 Finally, we would have to update some error reporting infrastructure
-to take ``GHCError`` values as arguments instead of ``ErrDoc``. That is
-the point at which the actual rendering of error messages would happen,
+to work with ``GHCError`` or ``IsError e => e`` values instead of ``ErrDoc``.
+That is the point at which the actual rendering of error messages would happen,
 under this proposal, right before calling the code that logs the said errors.
 
-A consequence of this is that the ``Messages`` type that GHC API users
-consume would now carry error and warning **values** that they can render
-but also inspect, without parsing. A lot of the work would be about
-actually moving all the error rendering code away from where we create
+A consequence of implementing this proposal would be that the ``Messages`` type
+that GHC API users consume would now carry error and warning **values** that
+they can render but also inspect, without parsing. A lot of the work would be
+about actually moving all the error rendering code away from where we create
 errors, and defining suitable types that carry the data around until
 it is time to report the errors to the user.
 
@@ -226,12 +246,7 @@ be made a lot easier if the current proposal is accepted.
 Costs and drawbacks
 -------------------
 
-The potential import cycles induced by this architecture and the ``.hs-boot``
-file(s) that we might add to work around them are going to add a little bit of
-maintenance overhead which we believe is largely compensated by having errors
-become proper values.
-
-Another drawback is that the wrapping in ``ParserError``, ``GHCError`` and
+One drawback is that the wrapping in ``ParserError``, ``GHCError`` and
 friends can be a bit verbose, becoming more verbose still as we introduce
 additional levels to the hierarchy. This can be mitigated by going for a
 rather flat hierarchy like the one presented above, with ``GHCError`` at the
@@ -245,6 +260,17 @@ The major cost of implementing this proposal is the sheer amount of
 refactoring that will be necessary to emit error values and move the
 rendering to much later, essentially delegating this work to each subystem
 and combining everything in the implementation of ``errorMessage``.
+Some complications are anticipated around the ``TcRn`` monad (and perhaps
+others), because this is a case of two subsystems using common code. A
+straightforward update to the ``Messages`` related types, used by ``TcRn``,
+would leave us with those two subsystems depending on ``GHCError``, introducing
+import cycles. A better idea would probably involve parametrizing ``TcRn`` (and
+the other types involved here) by the error/warning type(s) stored/accumulated,
+with functions to change the error type. This would surely involve quite a bit
+of refactoring but would give us precise types as far as errors are concerned,
+in a similar spirit to how ``mtl`` users write
+``MonadExcept SomeErrorType m => ...``, to say that a piece of code possibly
+throws errors of type ``SomeErrorType``.
 
 Alternatives
 ------------
@@ -264,10 +290,9 @@ This would allow us to work around the whole import cycles problem,
 at the price of being a lot more cumbersome to use: error consumers would
 have to use ``Typeable`` to implement specific behaviours for some types of
 errors. This price is likely a higher one to pay in the long run than the
-import cycles that we would work around when implementing the current proposal,
-as the cost will likely be non-trivial when/if we implement the proposal, but
-very small afterwards, especially with a flat hierarchy. GHC does not get a
-new subsystem nor an error infrastructure redesign all that often.
+refactorings necessary to implement this proposal, or than the
+import cycles that would come with using this proposal's ``GHCError`` everywhere.
+we would work around when implementing the current proposal,
 
 Unresolved Questions
 --------------------
@@ -275,7 +300,7 @@ Unresolved Questions
 We have not fully fleshed out the entire list of error types that would have
 to be defined, since we believe this is something that will be best done by
 scanning GHC's code, looking for functions that emit error messages and trying
-to adapt them to emit a suitably wrapped error **value**. This however did not
+to adapt them to emit a suitable error **value**. This however did not
 seem very relevant to describing the idea behind this proposal, as it is mostly
 about determining what constructors we should have in the "leaf error types"
 and which pieces of data have to be stored in those constructors, while the
--- a/proposals/0000-error-values-proposal.rst
+++ b/proposals/0000-error-values-proposal.rst
@@ -1,0 +1,291 @@
+Errors as values
+================
+
+.. author:: Alp Mestanogullari
+.. date-accepted::
+.. ticket-url::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. contents::
+
+Define and use a hierarchy of data types for all the errors,
+warnings and suggestions that GHC emits. Only render the error values
+in GHC-the-program, at the very last minute, to allow GHC API users to
+grab those error values instead of the current error documents. This new
+representation would be a lot easier to work with for IDE tooling developers
+and GHC API users in general: pattern matching on the new data types would
+replace parsing of the error messages.
+
+Motivation
+----------
+
+Up until this day, the errors that GHC emits are represented as mere
+textual documents. They often involve fancy layout constructs but GHC's
+code never manipulates errors, warnings or suggestions as values of some
+algebraic data types that are later rendered in the user's terminal; errors
+only ever exist as documents.
+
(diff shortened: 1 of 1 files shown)


ndmitchell:
I would find this proposal super helpful for the `ghcide` development work. To address a few things in the proposal:

* You can find many examples of this string munging in https://github.com/digital-asset/ghcide/blob/master/src/Development/IDE/LSP/CodeAction.hs. It's dependent on whether GHC has decided to use smart quotes or normal quotes. It's horrible. This proposal (or any variant on it) solves that beautifully.
* Warnings in GHC can be promoted to errors. I imagine the error type still wants a field saying whether its a warning or an error, and perhaps whether it started as an error originally, and perhaps which warning flag triggered it. Perhaps something along the lines of `data Warning a = Warning {cause :: WarningFlag, severity :: ErrorOrWarning, inner :: a}`
* I don't see how "using Typeable to implement specific behaviours" is vastly worse. It's slightly worse (you don't get type checking that you are ending up at a valid type), but `f (LexingError MismatchedBrackets{..})` vs `f (cast -> Just MismatchBrackets{..})` doesn't seem meaningfully different.
* If you have to centralise all the error message printing code, is that a good thing or a bad thing? Makes it easier to be more consistent, but less compositional?

simonpj:
I can see the attraction, but there woudl be substantial engineering costs.

* One humungous data type for all errors (transitively) which woudl have to import all data types involved in errors.  At least this means hs-boot files for all those data types, since the error type will be near the bottom of the hierarchy.

* Before long someone will want that data to be an instnace of Data which will generate monster code.

* The redering code needs lots of utility functions to pretty-print the payload, so it can't be down with the data type decl.  I don't konw what this will do to the module structure.

* Adding an error message will become painful.   Add a constructor + rendering, recompile everythign in sight.  Currently the rendering is right next to the place the error is generated, which is good.

I worry about he additional bureaucracy.   And about the cost of the switchover.

But I can see the attraction.  I'd like to hear a lot more from the consumers (people like Neil) about what the pain pionts really are.   I don't want to invest a lot of work in something that doesn't quite solve the problem.

I'd love to know what Idris does. They have "smart documents" that can unfold more info; but do they have this big enumeration as well?  The two seem to have overlapping functionality

goldfirere:
Though perhaps I'm out of date, @david-christiansen was at one point the point person for this stuff in Idris. Would you mind giving us a quick overview of Idris's approach?

goldfirere:
As to the humongous datatype approach: I've advocated for that in offering early feedback on this proposal, as it makes the different errors more discoverable. And I thought the `Typeable` approach would be horrid for consumers. But Neil points out that `Typeable` wouldn't be so bad, so maybe I was wrong there.

I had, at one point, convinced myself that the big datatype could live comfortably in the middle of the hierarchy: above all the basic types (`Type`, `CoreExpr`, `DataCon`, etc) but below the modules that actually do work (type-checker, renamer, etc.). And I'm a big proponent of the idea of error codes for easy online searching, etc., which would be easier with a big-datatype approach. I won't die on this particular sword, but I don't really feel the problems of the big-datatype approach. (Maybe the worst, for me, is textually separating the generation of the error and its text.)

int-index:
The approach as described is inherently anti-modular and will require `.hs-boot` files. To recap, the current proposal suggests this architecture:

```
    data GHCError
      = PsError ParseError
      | RnError RenamerError
      | TcError TypecheckerError
      | ...
```

Then, say, the parser, in order to produce `GHCError`, will transitively depend on `TypecheckerError`, and therefore on the typechecker-specific data types. Thus, _every component depends on every other component_.

Instead, why don't we do this instead:

1. The parser generates `ParseError`
2. The renamer generates `RenamerError`
3. The typechecker generates `TypecheckerError`
4. The driver then wraps these errors in a sum type.

In pseudo-code:

```
parse :: String -> Either ParseError ParsedSource
rename :: AST -> Either RenamerError RenamedSource
typecheck :: RenamedSource -> Either TypecheckerError TypecheckedSource
```

Then it's easy to wrap the errors in a sum type:

```
mapLeft f (Left e) = Left (f e)
mapLeft f (Right a) = Right a

mapLeft PsError . parse :: String -> Either GHCError ParsedSource
mapLeft RnError . rename :: Either GHCError RenamedSource
mapLeft TcError . typecheck :: RenamedSource -> Either GHCError TypecheckedSource
```

So the idea here is that we wrap component-specific errors into a sum type at a later point in the dependency graph, avoiding inter-component dependencies and `.hs-boot` files.

ndmitchell:
As an example of the functionality, if the user gets a warning that an import is redundant, we want to give the user a suggestion to remove that bit, and a textual edit that does the job. However, sometimes the user has written `import Foo(bar,baz)` and only `bar` is redundant, which is a similar but slightly different message. As a result, we have:

```
suggestRemoveRedundantImport :: ParsedModule -> Maybe T.Text -> Diagnostic -> [(T.Text, [TextEdit])]
suggestRemoveRedundantImport ParsedModule{pm_parsed_source = L _  HsModule{hsmodImports}} contents Diagnostic{_range=_range@Range{..},..}
--     The qualified import of ‘many’ from module ‘Control.Applicative’ is redundant
    | Just [_, bindings] <- matchRegex _message "The( qualified)? import of ‘([^’]*)’ from module [^ ]* is redundant"
   ... snip ...

-- File.hs:16:1: warning:
--     The import of `Data.List' is redundant
--       except perhaps to import instances from `Data.List'
--     To import instances alone, use: import Data.List()
    | _message =~ ("The( qualified)? import of [^ ]* is redundant" :: String)
        = [("Remove import", [TextEdit (extendToWholeLineIfPossible contents _range) ""])]
    | otherwise = []
```

We enable smart quotes everywhere with DynFlags settings, to ensure we have a somewhat stable regex. Things like the `qualified` note at the start are things we learn by experimentation. If GHC changes the error message, it breaks. Does it work for package imports? No idea, maybe GHC puts that info in the error, maybe not. If a line break happens in the pretty printing we probably miss the message.

bgamari:
@ndmitchell, regarding your point:

>     * I don't see how "using Typeable to implement specific behaviours" is vastly worse. It's slightly worse (you don't get type checking that you are ending up at a valid type), but `f (LexingError MismatchedBrackets{..})` vs `f (cast -> Just MismatchBrackets{..})` doesn't seem meaningfully different.

The problem is not that the syntax is meaningfully different. The problem is rather that it becomes very error-prone to document and evolve the error type since no-where is there a single list of the various errors that GHC can throw. This is the same problem that our extensible exceptions mechanism suffers from, but with the difference that GHC's error type will likely be more volatile than the exceptions typically thrown by most libraries.

Ericson2314:
@int-index i agree with @int-index that we do

> The parser generates ParseError
> The renamer generates RenamerError
> The typechecker generates TypecheckerError
> The driver then wraps these errors in a sum type.

Which *improves* modularity, rather than making it worse, and that's really good!

We we also return a parameter `e` which is constrained by constraints like `FromParseError e` `FromRenamerError e`, etc. This makes it less bureaucratic to combine errors if we need to do it in many places. In https://github.com/ghc-proposals/ghc-proposals/pull/266 I was attacking something related to this problem, and I found such classes (but their dual, the lensy version rather than primsy version) quite useful.

Speaking of https://github.com/ghc-proposals/ghc-proposals/pull/266 another big example of this is we can throw errors without having a `Outputable` instance. This means to break the `DynFlags` an `Outputable` cycles, in addition to my extant approach of making `Outputable` not hard-code `DynFlags`, we could also make more `.Ppr` modules with orphan `Outputable` instances, and not get messed up by error handling.

phadej:
From GHC plugin developers' POV, I'd like the "big sum" to have extension point with some `Typeable t =>` trickery, so one could teach before-mentioned `ghcide` to work also with non-GHC originated errors.

Yet otherwise, I'd like the "big sum". I don't think that loops are a real problem. In the definition of the error type one would need only to know that type exists, and that it has say `Outputable` instance. The needed `.hs-boot` files would be quite small. **There is already a `Type.hs-boot` and `Expr.hs-boot`**: https://github.com/ghc/ghc/blob/master/compiler/types/Type.hs-boot and https://github.com/ghc/ghc/blob/ad1fe274411129f82e7bbfd51bde49775288ae4b/compiler/GHC/Hs/Expr.hs-boot for example.

Ericson2314:
@phadej with the error type variable approach, we I'd hope plugin devs can instead use the ghc as a library and instantiate everything with their own error wrapper types.

In general, really think plugins are a major kludge working around GHC not being good as a library, and trying to support the plugin use-case vs just improving the library usually leads to dubious overly dynamic design decisions.

alpmestan:
Thanks everyone for the feedback

@ndmitchell Thanks for the links to the error parsing code. Regarding the dynamic errors, I have to admit I like the static hierarchy better (clearer, easier to consume). We can probably still leave a dynamic escape hatch as suggested later in the thread, for plugin authors or GHC API users in general.

@int-index Your suggestion is something that crossed my mind at some point, and could definitely work. As pointed out in the proposal, I'm not quite convinced yet that the cycles would indeed be that bad, but what you suggest is IMO a step in the right direction, to make GHC more modular.

Ericson2314:
> We can probably still leave a dynamic escape hatch as suggested later in the thread, for plugin authors or GHC API users in general.

At the very least, it would be nice to have a lint that regular GHC doesn't use it.

> but what you suggest is IMO a step in the right direction, to make GHC more modular.

Yeah I think it's less that cycles are bad, than modularity is good. The status quo of printing with `Outputable` which uses `DynFlags` is certainly bad even outside of the cycles it causes, in that it also isn't extensible.

aspiwack:
@alpmestan I'm really not in favour of inducing a dependency cycle which spans the entirety of the compiler. Smaller strongly-connected components are to be preferred. So I'd rather use the open-union. (and, as other points out, it makes it possible to define the error message close to where you use it)

With a bit of boilerplate, one can devise a way to make it more explicit and not rely on `Typeable` (or only rely on it for an escape hatch), if that's preferable.

alpmestan:
@simonpj

> I'd love to know what Idris does. They have "smart documents" that can unfold more info; but do they have this big enumeration as well? The two seem to have overlapping functionality

[I revised the initial post with a link to a related proposal (#307, to embed structured data/annotations into error documents) and a description of the overall plan/goal.]

Idris has the following error type(s) (`src/Idros/Core/TT.hs`):

``` hs
-- | Idris errors. Used as exceptions in the compiler, but reported to users
-- if they reach the top level.
data Err' t
          = Msg String
          | InternalMsg String
          | CantUnify Bool (t, Maybe Provenance) -- Expected type, provenance
                           (t, Maybe Provenance) -- Actual type, provenance
                           (Err' t) [(Name, t)] Int
               -- Int is 'score' - how much we did unify
               -- Bool indicates recoverability, True indicates more info may make
               -- unification succeed
          | InfiniteUnify Name t [(Name, t)]
          | CantConvert t t [(Name, t)]
          | CantSolveGoal t [(Name, t)]
          | UnifyScope Name Name t [(Name, t)]
          | CantInferType String
          | NonFunctionType t t
          | NotEquality t t
          | TooManyArguments Name
          | CantIntroduce t
          | NoSuchVariable Name
          | WithFnType t
          | NoTypeDecl Name
          | NotInjective t t t
          | CantResolve Bool -- True if postponed, False if fatal
                        t (Err' t) -- any further information
          | InvalidTCArg Name t
          | CantResolveAlts [Name]
          | NoValidAlts [Name]
          | IncompleteTerm t
          | UniverseError FC UExp (Int, Int) (Int, Int) [ConstraintFC]
            -- ^ Location, bad universe, old domain, new domain, suspects
          | UniqueError Universe Name
          | UniqueKindError Universe Name
          | ProgramLineComment
          | Inaccessible Name
          | UnknownImplicit Name Name
          | CantMatch t
          | NonCollapsiblePostulate Name
          | AlreadyDefined Name
          | ProofSearchFail (Err' t)
          | NoRewriting t t t
          | At FC (Err' t)
          | Elaborating String Name (Maybe t) (Err' t)
          | ElaboratingArg Name Name [(Name, Name)] (Err' t)
          | ProviderError String
          | LoadingFailed String (Err' t)
          | ReflectionError [[ErrorReportPart]] (Err' t)
          | ReflectionFailed String (Err' t)
          | ElabScriptDebug [ErrorReportPart] t [(Name, t, [(Name, Binder t)])]
            -- ^ User-specified message, proof term, goals with context (first goal is focused)
          | ElabScriptStuck t
          | RunningElabScript (Err' t) -- ^ The error occurred during a top-level elaboration script
          | ElabScriptStaging Name
          | FancyMsg [ErrorReportPart]
  deriving (Eq, Ord, Functor, Data, Generic, Typeable)

type Err = Err' Term

throwError :: Err -> Idris a
catchError :: Idris a -> (Err -> Idris a) -> Idris a
...
```

which is somewhat close to what _this_ proposals suggests, and then, for the pretty-printing side of things:

``` hs
data OutputAnnotation = AnnName Name (Maybe NameOutput) (Maybe String) (Maybe String)
                        -- ^^ The name, classification, docs overview, and pretty-printed type
                      | AnnBoundName Name Bool
                        -- ^^ The name and whether it is implicit
                      | AnnConst Const
                      | AnnData String String -- ^ type, doc overview
                      | AnnType String String -- ^ name, doc overview
                      | AnnKeyword
                      | AnnFC FC
                      | AnnTextFmt TextFormatting
                      | AnnLink String -- ^ A link to this URL
                      | AnnTerm [(Name, Bool)] (TT Name) -- ^ pprint bound vars, original term
                      | AnnSearchResult Ordering -- ^ more general, isomorphic, or more specific
                      | AnnErr Err
                      | AnnNamespace [T.Text] (Maybe FilePath)
                        -- ^ A namespace (e.g. on an import line or in
                        -- a namespace declaration). Stored starting
                        -- at the root, with the hierarchy fully
                        -- resolved. If a file path is present, then
                        -- the namespace represents a module imported
                        -- from that file.
                      | AnnQuasiquote
                      | AnnAntiquote
                      | AnnSyntax String -- ^ type of syntax element: backslash or braces etc.
  deriving (Show, Eq, Generic, Ord)

instance Show Err where ...
instance Pretty Err OutputAnnotation where ...
```

The `Pretty` class:

``` hs
class Pretty a ty where
  pretty :: a -> Doc ty
```

looks exactly like the modified `Outputable` that would come with #307. All in all with these two proposals we would get closer to the Idris' error architecture. The main difference being that we have `Doc` and `SDoc` while Idris only has `Doc`, from [a hackage library](http://hackage.haskell.org/package/annotated-wl-pprint-0.7.0/docs/Text-PrettyPrint-Annotated-Leijen.html#t:Doc). This particular variant is what #307 refers to as "scoped annotations" in the "Alternatives" section, where annotations are nodes instead of leaves, they always wrap an arbitrary sub-document. If you want to continue this discussion or have questions, we can perhaps follow up in #307.

---

Regarding the cycles, I'm hoping that we can find a sweet spot in the module hierarchy that's above the core AST types and below the typechecking/etc code. As pointed out by @int-index this becomes a lot easier to work with when varying the error type for each phase.

Also, technically, I'm not entirely sure we _do_ need to specify the interpretation of error values as documents right next to the error ADTs, in `Outputable` instances. This could all just be specified as a `GHCError -> SDoc` function (current version of the proposal) or `{PsError, RnError, TcError} -> SDoc` functions (@int-index variant), either a fixed definition or a configurable one that would live in `DynFlags` or a similar place and defaults to what would otherwise be the fixed definition. The said default definition could live somewhere under `compiler/main/` and could perhaps use pieces that are defined below it in the module structure, close to where the errors are emitted. The definitions of the error types would however indeed have to live somewhere in the middle, not right next to where we produce those errors.

Just to be clear though, the dynamic approach has at some point been the one we were focused on, and if it's the one that most error producers and consumers want to go with, I'd be happy to revise the proposal accordingly. I (perhaps wrongly) anticipated that the easy/trivial consumption (with straightforward pattern matching) would be a big selling point.

goldfirere:
In case I hadn't said it previously, I've also advocated for the static approach over the dynamic (`Typeable`) one. Though I agree that we should have an escape mechanism for e.g. plugins to embed errors we did not anticipate; `Typeable` would work well here.

A slightly separate question is: if we go with the static approach, where do we define the datatype? I do believe there would be a "sweet spot" -- or, if not, it would take only a little work to make one. But this approach *is* inherently anti-modular. The approach in https://github.com/ghc-proposals/ghc-proposals/pull/306#issuecomment-574585808 by @int-index seems to be a strict improvement in this regard. Is there any reason not to go with that design? The internets seem to like it, too.

alpmestan:
I'm personally quite enthusiastic about @int-index's suggestion, it feels right. All the error types would be defined "where they belong", including the sum type in the driver. The only drawback that I see is that this makes the proposal a little bit more invasive, as we might have to e.g add type params to some types to allow the error type to vary, etc. But I think it's worth it.

parsonsmatt:
I've thought quite a bit about errors and ergonomics relating to them. The approach that I give and document in the [`plucky`](https://hackage.haskell.org/package/plucky-0.0.0.1/docs/Data-Either-Plucky.html) library has quite good ergonomics, extensibility, and modularity while retaining static type safety and the ability to discharge all errors.

I personally don't like the 'mega error' type, as it has many usability and semantic issues [that I cover here](https://www.parsonsmatt.org/2018/11/03/trouble_with_typed_errors.html).

alpmestan:
I tried to update the proposal to integrate @int-index's suggestion. Feedback/comments welcome!

bitc:
This is going on a bit of a tangent, but I would like to offer my perspective.

@simonpj said:
> [...] I'd like to hear a lot more from the consumers (people like Neil) about what the pain pionts really are. I don't want to invest a lot of work in something that doesn't quite solve the problem.

And then @ndmitchell replied that one of the use cases for structured errors is to be able to apply an automatic "Code Action" that will fix the error (for example, a redundant import compiler error will be parsed by the IDE and then a Code Action can automatically remove the redundant import).

For this use case, there are other ways to accomplish this without having fully structured errors. One way is: for each compiler error, the compiler could also emit a list of "Code Actions" that can be used to fix the error automatically. This is how eslint works: <https://eslint.org/docs/developer-guide/working-with-rules#applying-fixes>

The advantage of this is that this "autofix" algorithm only needs to be implemented once for each error in the compiler itself, and then it is very easy for every editor plugin and IDE to hook into it. As another example, GHC sometimes detects when you made a typo and recommends other names that you might have meant instead. It it were to also emit "autofixes", then it would simply be one click in your editor to apply the correct typo fix.

Moving on. Another use case for structured error messages is to provide hyper-links for relevant parts of the error message. For example, if the error message mentions a type, it would be nice for the IDE to underline the type so you could click it and get more information or jump to its definition. This use case also doesn't require fully structured error messages. Instead, we can use a richer markup language (like XML) for the error messages, so instead of an error like:

    The type `Integer' doesn't match the type `Double'

we get an error like:

```xml
The type <Type qualified-name="Data.Integer">Integer</Type> doesn't match the type <Type qualified-name="Data.Double">Double</Type>
```

This could still be rendered as plain text for regular command-line usage of the compiler, but an IDE would be able to render it with "intelligence" and extra magic sauce applied. I believe converting all of the error messages inside the compiler to emit such a markup (instead of plain text) would be fairly straight-forward

ndmitchell:
@bitc - having Markup in the types seems quite nice.

Having GHC generate the suggestions would also be lovely - but I'm not sure it's feasible. Imagine that `Integer` is not in scope, a fairly simple error. One way to resolve that would be by importing `GHC.Integer`, another by importing `Prelude`. You could do it via `import Prelude(Integer)`, or just `import Prelude`. A better way would be to look at adjacent modules in this project, or perhaps set of projects, and see what they did to get Integer into scope - maybe most of the do `import Intero(Integer) as Q` and use `Q.Integer`, so maybe that pair of solutions is now the best. Or maybe I've got a config file that ways how to resolve Integer. Or maybe I search Hoogle for it. I don't see a clean API to bake into the compiler that lets the IDE have this flexibility, and I can't imagine GHC HQ wants to shove all that logic into their API (because it really doesn't belong)..

alpmestan:
(@bitc Note that there is a _separate_ proposal about enriching GHC's documents to (hopefully easily) enable the use case you described at #307. GHC wouldn't emit the markup but bits of ASTs, embedding `HsExpr` values and so on, that tools could then render however they want.)

goldfirere:
This conversation has essentially moved to https://gitlab.haskell.org/ghc/ghc/-/issues/18516, various associated merge requests, and a [wiki page](https://gitlab.haskell.org/ghc/ghc/-/wikis/Errors-as-(structured)-values).

I'm not sure this PR is serving a purpose at this point. Shall we close? @adinapoli may have an opinion (he is the one working on this).

adinapoli:
> This conversation has essentially moved to https://gitlab.haskell.org/ghc/ghc/-/issues/18516, various associated merge requests, and a [wiki page](https://gitlab.haskell.org/ghc/ghc/-/wikis/Errors-as-(structured)-values).
> 
> I'm not sure this PR is serving a purpose at this point. Shall we close? @adinapoli may have an opinion (he is the one working on this).

@goldfirere Indeed we seem to have _a lot_ of information when it comes to this piece of work, and I agree that this proposal was a good place to kickstart the conversation but at the moment it's not clear to me what this adds to the plate _on top_ of the resources you linked: sure, it still provides good historical context, but that's it I think. 

Part of the sentiment stems from the fact this was never merged, so technically it doesn't even show up in the [accepted GHC proposals](https://ghc-proposals.readthedocs.io/en/latest/), so it's in this weird limbo where it's neither an "authoritative" source (i.e. "look, we have an approved proposal for this") nor reflects the most updated state-of-the-art of the work being done.

So yes, probably closing it makes sense.

goldfirere:
Yes, I believe this was intended as a place to kick-start the conversation. I will close.

Passers-by: please continue the conversation at https://gitlab.haskell.org/ghc/ghc/-/issues/18516. We'd love to have your input.

PR #307, opened 2020-01-24
richer error documents proposal

alpmestan:
[Rendered](https://github.com/alpmestan/ghc-proposals/blob/annotated-documents/proposals/0000-richer-error-documents.rst)

Just like #306, this proposal is a bit unusual in that it is about a change that's almost internal to GHC. And just like #306, it is about making the life of GHC API users easier. Both proposals are about providing them with easier solutions to problems that they are facing when dealing with the errors that GHC emits, especially in the IDE/tooling space. When "composed" together, these two proposals will give us proper ADTs to represent errors/warnings and a flexible way to consume the error documents that are derived from those ADTs, with the ability to embed structured data in the middle of documents and leave the final rendering of those bits to users.

### DIFF
--- a/proposals/0000-richer-error-documents.rst
+++ b/proposals/0000-richer-error-documents.rst
@@ -1,0 +1,539 @@
+Richer (error) documents
+========================
+
+.. author:: Alp Mestanogullari
+.. date-accepted:: Leave blank. This will be filled in when the proposal is accepted.
+.. ticket-url::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. contents::
+
+Enrich GHC's document types (``Doc`` and ``SDoc``) with support for
+annotations of some arbitrary type ``a``, and use this new capability
+to emit error messages with bits of expressions, types and more
+embedded in the documents, as AST values, as opposed to their
+textual form which is not amenable to any form of inspection or
+post-processing.
+
+This proposal drew a lot of inspiration from the discussion at
+`ghc#8809 <https://gitlab.haskell.org/ghc/ghc/issues/8809>`_.
+
+Motivation
+----------
+
+GHC's whole pretty printing infrastructure is based around the following
+types:::
+
+    data Doc
+      = Empty                                            -- empty
+      | NilAbove Doc                                     -- text "" $$ x
+      | TextBeside !TextDetails {-# UNPACK #-} !Int Doc  -- text s <> x
+      | Nest {-# UNPACK #-} !Int Doc                     -- nest k x
+      | Union Doc Doc                                    -- ul `union` ur
+      | NoDoc                                            -- The empty set of documents
+      | Beside Doc Bool Doc                              -- True <=> space between
+      | Above Doc Bool Doc                               -- True <=> never overlap
+
+    data SDocContext = SDC
+      { sdocStyle      :: PprStyle
+      , sdocLastColour :: PprColour
+      , sdocDynFlags   :: DynFlags
+      }
+    newtype SDoc = SDoc { runSDoc :: SDocContext -> Doc }
+
+``SDocContext`` gives access to all sorts of pretty-printing and compiler
+configuration. These types allow us to specify how sub-documents should be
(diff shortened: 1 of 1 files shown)


Ericson2314:
I must admit I think I am more excited about #306 alone. I once tried making a "trees-that-grow-ified" doc type like this, but the only uses for it boil down to mixing content and presentation together. That's never been considered a very a good idea. #306 alone however completely decouples the content from the presentation, *forcing* one to create a variant for each sort of error, which I think is a good thing.

int-index:
@Ericson2314 has put it nicely. In my opinion, this proposal mixes content and presentation. 

With #306 alone, we have a clear dichotomy:

* structured data for tools to process
* formatted text for humans to read

This proposal seems to advocate for a mix of structured data and formatted text for tools to process. It seems to me that this simply enables tools with worse architecture.

goldfirere:
I guess the question is what do tools want? If we have #306, then we absolutely will have the ability for tools to get our datatypes directly. They can then implement all their own error messages / do custom processing / whatever. The problem is that this puts a huge, ongoing burden on tools: they need to render all error messages themselves. Every (minor!) release of GHC will require updating these error messages. This seems very flexible, but utterly unsustainable.

The goal in this proposal is precisely to allow for a mixed economy. Yes, that is often a bad idea. But I think it's the best bad idea in this space. With this proposal (or something like it), GHC can partially render errors, providing much of the text, while still embedding rich content. This rich content can then be made interactive by tools. Tools can opt to deal with whatever embedded content they like. For example, a tool might work with embedded types and terms, but not deriving strategies. Thus, deriving strategies would just be rendered by GHC, while the others would be handled by the tool. The idea is that the shape of the embedded rich content would be *much* more stable over time than the overall error messages. This seems to provide a nice middle ground.

To be clear: I agree that this mixes content and presentation, which is bad. But "all content" (that is, #306 alone) puts an impossible, ongoing burden on tool developers. And "all presentation" (what we have now) severely limits what tools can do. So we embrace the content+presentation as the best solution to a hard problem.

Ericson2314:
@goldfirere

 > The problem is that this puts a huge, ongoing burden on tools: they need to render all error messages themselves. 

There's not reason the external tool cannot use `Outputable` to render anything from GHC it doesn't understand. If that is not good enough, I would make this proposal orthogonal to #306, making it about a richer `SDoc`/`Outputable` and nothing to do with errors.

GHC suffers from a huge non-modularity problem which rarely goes discussed except as sort of pie-in the-sky wishful thinking. This really irks me, and I think #306 vs this could have a big effect:

 - If we do this, not only will we continue to mix the compiler logic with various pretty printing concerns, but that pretty printing logic will be even more complex and---I wouldn't be surprised--liable to have more `hs-boot` madness as we try to put the syntax in `Outputable` instances.

 - If we do #306, not only can we break cycles, but we can move `SDoc`/`Outputable` out of the main ghc library altogether (!), and into a separate library used by `ghc` the executable to render caught exceptions or whatever.

goldfirere:
This is meant to be orthogonal to #306. It's not either/or. It's both/and. Yes, the proposal talks about errors as a concrete example, but that carries over to the #306 world, as well.

int-index:
> we can move `SDoc`/`Outputable` out of the main ghc library altogether (!)

This indeed sounds nice. But there's at least one stumbling block to consider: debug output and compiler panics. We'd likely need some sort of formatting there. And when I `-ddump-tc-trace` I don't want to get machine-readable JSON or some such.

Ericson2314:
@goldfirere Ah whew. So there's no inherent tension with I said? That's a relief.

@int-index Isn't it enough for GHC the executable to use the pretty-printing library for that?

int-index:
> Isn't it enough for GHC the executable to use the pretty-printing library for that?

Well, if you introduce some sort of structured logging, then you could move pretty-printing to the executable. Otherwise, all the `traceTc` and `traceRn` will need a way to render the relevant parts of the AST.

Ericson2314:
Yes, using the ideas in this proposal for logging, or hell, a simplified version with `Typeable`, seems like the far less bad place to mix content and presentation.

Boarders:
What is the latest on this? There seems to have been no activity for months, are there still conversations going on about what the best move forward is?

alpmestan:
@Boarders I don't think there has been anything new, nor has this been discussed elsewhere. Progress has been made on trying out the ideas from #306, but not this proposal (yet).

PR #310, opened 2020-02-14
Disambiguate record update

Ericson2314:
A conservative plan. Can contrasted with #282 as something in the spirit #160 hewing closer to the semantics of records today.

 - If you like today's record semantics, you might simply prefer this.

 - If you wish we had structural record types, you may *still* prefer this to free up syntax for them now so a bold plan in the future doesn't need to contend with a complex transition period.

[Rendered](https://github.com/Ericson2314/ghc-proposals/blob/disambiguate-record-update/proposals/0000-disambiguate-record-update.rst)


Ericson2314:
CC @reactormonk

adamgundry:
As a first observation, note that the proposed syntax is ambiguous in certain cases involving data families. If two instances of a data family declare the same field name (potentially in separate modules) it is necessary to specify the entire type, not just the type constructor, to identify which is meant. And we have a syntax for that already! One can simply write either of
```hs
(r :: MyRecord) { a = b }
r { a = b } :: MyRecord
```
Admittedly this gets a bit painful if `MyRecord` is a type with many parameters, but `PartialTypeSignatures` helps. Thus I'm not very convinced a new syntax is justified.

Ericson2314:
> data families

Ugh. Yes those are a real pain. I wanted a purely-renaming-based implementation to keep the implementation simple, provide simple error messages, and be extremely easy to read. Normalizing data families brings in tons of complexity and undermines all that.

Ericson2314:
> One can simply write either of

```
(r :: MyRecord) { a = b }
r { a = b } :: MyRecord
```

But note that unless we *heavily* special case this compound syntax, we are stuck in the status quo of mixing renaming and type checking together for record fields, even for non-data-families.

Ericson2314:
@adamgundry Sorry, I somehow missed your thorough review of this. Thanks for that. The data family issue is irksome, as the only solution seems to be going back to data constructors with the issues of sum types with record syntax, so I am not quite sure what to do with this one.

AntC2:
Heh heh I seem to have come up with a [competing proposal](https://github.com/ghc-proposals/ghc-proposals/issues/328). At discussion stage for the time being, but I have a prototype. Furthermore that prototype co-habits with H98 (as @adamgundry says, we've gotta be backwards-compatible) and with at least one structural record type implementation (Trex). Not that I'm proposing any structural records as part of it.

Here's my proposed record update syntax -- this is as-well-as H98 syntax, not instead-of. There are a few clashes with H98 records that are a little noisy to disambiguate, see [formal syntax](https://github.com/ghc-proposals/ghc-proposals/issues/328#issuecomment-622927157) discussion.

    aPoint = {Point | x = 1.0, y = 2.5 }                 -- record build syntax

    bar r = {r | x = 3.4}                                -- record update syntax
    bar r = {r :: Point | x = 3.4}                       -- record update with signature

That prominent position for the signature is precisely to `DisambiguateRecordFields`.

I strongly dislike H98's update syntax and pattern matching, for the reasons you lay out. Your proposal here keeping the constructor outside the `{ ... }` really doesn't reduce the surprise element.

I'm proposing syntax very similar to current, just 'hopping over the fence' the constructor or value for update. For that reason the implementation cost is low: it's syntactic sugar only -- I've implemented the whole prototype in yacc. 

I'm not currently addressing one of the improvements you're trying to make: for a record update, what if the constructor is uninferrable. Hmm ripping some syntax out of nowhere:

    bar r = {Point | r | x = 3.4}                        -- record update with constructor doesn't need signature

Implementing that isn't simply sugar.

Ericson2314:
@AntC2 my initial 30 second reaction is your proposal is much better :)

Ericson2314:
@AntC2 To the last bit of your argument, record update today doesn't bias any constructor---it's total as long as all fields appear in all variants. So I don't even hold that against your proposal. I think the next step is to look at exactly how constructor syntax + data families works today, which I assume is going to be a mess, and then come up with something like your thrown-out `{Point | r | x = 3.4}`.

Ericson2314:
@AntC2 I noticed you made an issue rather than PR. Does that mean you proposal as-written is somewhat tentative? I would be happy to just turn this PR into your proposal with you if you are looking for a co-author.

AntC2:
> @Ericson2314  I noticed you made an issue rather than PR. ...

Thanks John, I'd be delighted to work with a co-author. I should warn you the reason I don't delve into the insides of GHC, and prefer developing with Hugs, is because I don't understand anything of the instructions even to install/build GHC from source, let alone change anything. It might be significant that I work on Windows (ref the recent discussion on the cafe). And I enjoy developing in Hugs because I find the code easy to read -- I say that even though it's written in C/C++, a language I've never used commercially. yacc is easy-peasy.

I started that proposal as a 'Issue', chiefly to throw down my thinking while it was still fresh in my head; and I was flushed with success of mastering yacc.

Also because my previous experiences writing a full-blown proposal were so negative: it was a lot of effort filling in material that was not essential to understanding the design IMO; that attracted a lot of nit-picking that failed to consider the main changes; consequently the guts of the proposal didn't get any close examination until it was brought before the Committee (not a good use of their time, and by nearly a year after I first wrote it up, I'd just lost the will to live).

The Committee's responses were polite; but really not interested. Clearly GHC Haskell at some point is going to abandon Functional Dependencies and Overlapping Instances (which is what I was trying to improve). From GHC's point of view, it must be annoying that so much code (especially Lenses) continues to embrace FunDeps, and looks balefully at Type Families with all their powerful but superfluous machinery. Avoid success at all costs!

Given that H98's record system is still alive despite that comment back in 2006 (which would be conventional wisdom at the time); and despite the bits and pieces of improvement in GHC that trickle along achingly slowly despite the passions/interest that record proposals fire up; I suspect I'm going to find that the Committee will again be polite but uninterested in anything for H98 records.

So a discussion around an 'Issue', focussing on core functionality may be a better way to test the water without so much effort. With my previous proposal, I couldn't offer a prototype; so I couldn't assess the cost/benefit. With this proposal, the prototype shows there's no cost in touching actual code/it's all syntactic sugar. And it makes the code I'm writing look prettier IMO; so the effort is already paying dividends for me whether or not the proposal goes anywhere.

Ericson2314:
@AntC2 Your proposal can also be be motivated as "putting H89 records out to pasture"----i.e. freeing up syntax for new fancy experiments and retiring Haskell 98 records to some other syntax that doesn't get in the way. I hope that can widen support to include both people that like and dislike the current semantics.

I forget what happened with your other proposal, but there is nothing technically challenging about this as you, so I don't think any prototype would be needed. I suspect this would go a lot faster and be worth the effort.

AntC2:
>> I'm not currently addressing one of the improvements you're trying to make: for a record update, what if the constructor is uninferrable. 

> come up with something like your thrown-out `{Point | r | x = 3.4}`.

I'm thinking we probably have enough hooks already, if we mesh together enough GHC extensions (and similarly mix our metaphors ;-)

    foo (rr@{Point | x, .. } :: r) = { Point :: r | x = ..., .. }

On the rhs of `foo` we have a as-pattern variable `rr` for the whole record; we've pattern-matched the constructor; we have a pattern-signature type variable `r` for the record type; and we've `FieldPuns/RecordWildCards`-bound all the labels, so we can again use `RecordWildCards ..` to complete the record build/update.

Admittedly that expression is not easy on the eye, but from what Adam's saying, I don't think Data Families is such a common use case that we need to work hard to improve it.

> "putting H89 records out to pasture"----i.e. freeing up syntax for new fancy experiments and retiring Haskell 98 records to some other syntax 

I would indeed like to put H98 syntax out of the way. But realistically there's so much legacy code, we'll have to live with it for a long time. So to be clear: the advantage here is that everything record-y is compatible with H98 syntax and can run alongside. You can even mix the different syntaxes within the same program -- as you can with Trex and H98. There's no need to find "some other syntax" for H98 records: it can enjoy a long and graceful retirement. At some time in the remote future we might gently draw down its eyelids.

Some of the criticisms of H98 records are coming from OOP/procedural languages with update-in-situ semantics. Then they want or think of update as poking a value into a location pointed to by a object-instance reference. Haskell (thank crikey) doesn't have reference semantics like that. We create a new value by copying (but preserving pointers to thunks/existing values). What might be inefficient (and this is what worries me about `setField` proposals) is if you're updating two fields at the same time: because `setField` works field-at-a-time, it'll create a copy for the first field (including copying the useless current content of the second field), then immediately throw that copy away to create a copy for the second field (repeating the copy of the first-updated field). If these fields are deeply-nested this gets worse and worse. Then adding syntax for deep-nesting record labels makes it look less fussy but does not add one drop of efficiency. Lenses likewise hide the fact of copying around half-built records.

That's why I'm a strong advocate for Extensible records: don't nest, do extend! Keep everything flat -- exactly as Ted Codd figured out almost exactly 50 years ago.

Ericson2314:
@AntC2 

> I would indeed like to put H98 syntax out of the way. But realistically there's so much legacy code, we'll have to live with it for a long time. So to be clear: the advantage here is that everything record-y is compatible with H98 syntax and can run alongside.

Ah but compatibility is a devilish game. The trick is usually on-by-default extensions.

Let's say we make a `-XPipeRecords`, and call the existing thing `-XH98Records`. The former is off by default, the latter is on by default, all 4 combinations (`-X...`/`-XNo...`) are valid. 

Now, let's suppose suppose we want to have `{...}` anonymous records. Those still syntax. What do we do? Make a `-XAnonymousRecords` which conflicts with `-XH98Records`---one have enable one of each or neither, but not both.

Just as you say, H98 records are going to be with us for a while. I would H98 record *semantics* will be with us far longer than the syntax, too: one could imagine bulk-converting the syntax of all of Hackage, but absolutely not bulk-converting the semantics.

So what does this buy us? Say you want to use whatever the new records are, but need to also use some older library that uses the old records, and don't want to fall back on positional code. `-XAnonymousRecords` doesn't conflict with `-XPipeRecords`, so this is in fact the only way to make a seamless transition without throwing away ergonomics.

In short, the legacy interface doesn't need to defined using the same syntax it is consumed with.

PR #311, opened 2020-02-17
`data` without parameters with -XStandaloneKindSignatures

JakobBruenker:
This proposal suggests that it should be possible to write a GADT declaration `data Foo` without additional parameters if the kind of `Foo` is given in a standalone kind signature.

[Rendered](https://github.com/JakobBruenker/ghc-proposals/blob/no-data-params/proposals/0000-no-data-params.rst)

### DIFF
--- a/proposals/0000-no-data-params.rst
+++ b/proposals/0000-no-data-params.rst
@@ -138,7 +138,7 @@ Costs and Drawbacks
 -------------------
 
 The proposed change is a fairly minor one that doesn't affect the Parser and is
-essentially identical to already existing behaviour for inline kind signatures,
+very similar to already existing behaviour for inline kind signatures,
 so the implementation and maintenance costs should be low.
 
 Alternatives
--- a/proposals/0000-no-data-params.rst
+++ b/proposals/0000-no-data-params.rst
@@ -79,11 +79,24 @@ In ``data`` or ``newtype`` declarations with GADT-syntax and a standalone kind
 signature, if no kind signature follows the parameter list following the
 ``data`` or ``newtype`` keyword, the status quo is that ``n`` parameter names
 must be given in this list, assuming the standalone kind signature specifies an
-``n``-ary kind. These parameters may have inline kind signatures matching the
+``n``-ary kind. These parameters may have individual inline kind signatures, matching the
 standalone kind signature.
 
 The change suggested in this proposal instead allows between zero and ``n``
-(inclusive) parameter names to be given.
+(inclusive) parameter names to be given, regardless of whether or not a kind signature
+follows the parameter list.
+
+Alternatively, the inline kind signature of the declaration may be
+a wildcard (``_``), which allows users to more explicitly state that the kind is determined by the
+standalone kind signature.
+
+In ``data instance`` declarations with GADT syntax, this also works: If a standalone kind signature has been given
+to the ``data family`` declaration, then a data instance declaration without a kind signature (or with a
+wildcard as kind signature) is assumed
+to have the most general kind possible, i.e. the one given by the standalone kind signature. Note that
+this will result in an error if the kind specified by the standalone kind signature has a non-* return kind.
+
+In the context of this proposal, declarations without constructors are also treated as GADT-style declarations.
 
 Examples
 --------
@@ -96,12 +109,14 @@ are all accepted:
   data Foo
   data Foo a
   data Foo a b
+  data Foo :: _
+  data Foo a :: _
+  data Foo a b :: _
   data Foo :: Type -> Bool -> Type
   data Foo a :: Bool -> Type
   data Foo a b :: Type
 
-As well as any variations that include inline kind signatures for ``a`` or
-``b``.
+As well as any variations that include inline kind signatures for ``a`` or ``b``.
 
 There are some illustrative real-world examples in the singletons library that
 already use standalone kind signatures today. Old lines are prefixed with
@@ -122,10 +137,21 @@ already use standalone kind signatures today. Old lines are prefixed with
   + newtype WrappedSing where
       WrapSing :: forall k (a :: k). { unwrapSing :: Sing a } -> WrappedSing a
 
-    -- Empty data declaration (also GADT-syntax)
+    -- Empty data declaration (also treated as GADT-syntax)
     type TyFun :: Type -> Type -> Type
   - data TyFun a b
   + data TyFun
+  
+  
+While less useful, this can also be used for data families:
+
+::
+
+  data family F1 :: Bool -> Type
+  data instance F1 -- kind: Bool -> Type
+  
+  data family F2 :: Bool -> k
+  data instance F2 -- error due to non-* return kind
 
 Effect and Interactions
 -----------------------
@@ -137,17 +163,23 @@ implicitly guarded behind the ``-XStandaloneKindSignatures`` extension.
 Costs and Drawbacks
 -------------------
 
-The proposed change is a fairly minor one that doesn't affect the Parser and is
-very similar to already existing behaviour for inline kind signatures,
-so the implementation and maintenance costs should be low.
+This is a fairly minor change, since the compiler already has to ensure that the standalone kind
+signature matches the inline kind signature. As such, the implementation and maintenance costs
+should be low.
 
 Alternatives
 ------------
-None aside from keeping the status quo.
+- Fully partial kind signatures could be allowed, rather than just wildcards. Given the marginal
+  benefit, this does not seem to be worth it.
+- Between making it possible to leave off the kind signature, and to have it be a wildcard, one of these
+  two could be dropped. The main reason to keep the first is brevity, the main reason to keep the second
+  is allowing the user to be explicit.
+- These changes could only apply to declarations that have not been given any paramaters, rather than zero to
+  ``n`` parameters.
 
 Unresolved Questions
 --------------------
-None.
+Should ``-XPartialTypeSignatures`` be required for Wildcards?
 
 
 Implementation Plan
--- a/proposals/0000-no-data-params.rst
+++ b/proposals/0000-no-data-params.rst
@@ -11,9 +11,7 @@
 .. implemented:: Leave blank. This will be filled in with the first GHC version which
                  implements the described feature.
 .. highlight:: haskell
-.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
-            **After creating the pull request, edit this file again, update the
-            number in the link, and delete this bold sentence.**
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/311>`_.
 .. contents::
 
 When a data declaration with GADT-syntax has been given a standalone kind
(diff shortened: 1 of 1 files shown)


Ericson2314:
Good idea. This was bugging me too.

int-index:
The proposed behavior is what the initial implementation used to do, but then @goldfirere and @simonpj convinced me to require `n` parameters in the header. 

There are some outdated discussions at https://gitlab.haskell.org/ghc/ghc/wikis/ghc-kinds/kind-inference/tlks. Search for "object strenuously" to see the relevant section :smile:

JakobBruenker:
@int-index Thank you, I will take a look

Ericson2314:
I wouldn't mind an extension to just bans standalone family and GADT implicit signatures.

Class heads and Haskell 98 LHS seems actually part of a term in some sense. (Furthermore if you think of making elaborating the dictionary type Haskell 98 style, the former turns into the latter.) But GADT implicit signatures and standalone family "heads" really just are weird signatures. The term is definitely just the family pattern matching / GADT constructors, and not the implicit signatures.

I think those signatures just overlap awkwardly with standalone kind signatures, and by allowing one to turn them off we avoid the inference pitfall (you just check against the standalone kind signature, or infer from the closed instance / constructor return types). I would even deprecate these implicit features in the future.

JakobBruenker:
Regarding the link @int-index shared, I agree with the arguments made in it regarding Haskell 98 syntax, which seems to be the main objection, and of course is why this proposal only affects GADT-style declarations.

I agree with @Ericson2314 that the heads of GADT-style declarations and standalone data/type families are weird (in fact, this weirdness extends to associated families, hence the need for https://github.com/ghc-proposals/ghc-proposals/pull/259).

An extension to turn off inline kind signatures for these heads seems fine to me.

Note that data families, which are pretty similar to GADTs, already allow the following:

```haskell
type Foo :: Bool -> Type
data family Foo
data instance Foo True = Bar -- i.e. Bar :: Foo True
```

Here, `Foo :: Bool -> Type`, even though it has no inline signature. Given that, it seems odd to disallow this:

```haskell
type Foo :: Bool -> Type
data Foo where
  Bar :: Foo True
```

The main point regarding GADTs from @int-index's link was this one from @simonpj:

> But even in GADT syntax, I'm not sure we want to accept a data type declaration that (on the face of it) appears to have kind `T1 :: Type`. The kind signature might be far away.  And, absent a TLKS if we say
```haskell
data T1 where
  ...
```
> then, regardless of the `...` we'll infer a kind for `T1` ending in `Type`.  Let's be consistent with that.

But all of this is *already* the case for data families.

I find it somewhat unfortunate that (especially open) type families rely on the parameter names to determine their arity, otherwise those could just be removed from everything (i.e. data/type families and GADTs) with such an extension as well, but I don't really see a way around that.

int-index:
> But all of this is *already* the case for data families.

Data families are quite a different beast. Reasoning here: https://gitlab.haskell.org/ghc/ghc/issues/16756#note_203567

In short, for data families we have the restriction as well, but it applies to data *instances*. The principle is that we require `n` parameters for things that introduce data constructors:

* `data` declarations
* `data` family instances
* `class` declarations (because they introduce a dictionary in Core)

JakobBruenker:
@int-index hm, you're right, and it looks like data families instances throw a bit of a wrench into this - I hadn't considered that data family instances can *also* use GADT-syntax. I'll have to think about this.

Ericson2314:
@JakobBruenker Hmm that is a wrench. But I have hope for the future. In particular, per original constrained type classes paper, open families will turn into closed type classes with associated types. Also, per dependent type common sense, repeating the instance head arguments on associated types is ludicrous because we are putting a single value, and not function, in the dictionary. Combine these two things, and the situations is fixed!

```haskell
data family T :: Type -> k
data instance T Int where
  Foo :: Int -> Int -> Int  :: T Int
data instance T Bool a b where
  Bar :: Bool -> a -> b  :: T Bool a b
```

becomes

```haskell
type TC :: Type -> Constraint
type TK :: forall (x :: Type) -> TC x => Type -- user chooses visibility
-- ^ oh hey, now we know a reason this can go outside
class closed TC x where
  type TK :: Type
  data T :: TK x
  instance TC Int where
    type TK = Type
    type T :: TK Int -- optional
    -- ^ note on the definition side we drop the extra args but not on the use side. 
    -- likewise we should ban `return @classVar = ...` since the binding should use the "internal" not "external" type.
    -- name resolution already distinguishes between internal and external uses, so that helps.
    data T where
      Foo :: Int -> Int -> Int :: T Int
  instance TC Bool where
    type TK = Type -> Type -> Type
    type T :: TK Bool -- optional
    data T where
      Foo :: Bool -> a -> b :: T Bool a b
```

Now, there is a clean separation between the instance-selection arguments and extra parametric arguments stemming from the kind polymorphism.

Now, doing all the above in one proposal would be *huge*, but we can still go with my "banning plan", and also just ban "family syntax" altogether with that same extension to punt. On this own, that extension/proposal looses expressiveness and is bad. But as one plank of a the whole platform that goes into making the above reality, it is a useful building-piece and nice small separated-out bit of work (the change is self-contained even if the motivation isn't). And by separating things out we make it easier to collaborate on the whole.

JakobBruenker:
@Ericson2314 Interesting. I'll have to look at this a bit longer to fully understand it, but if you want, feel free to make a PR to this proposal or to make your own proposal to reflect your banning plan, before then. (Otherwise I'll likely change this one myself once I'm at that point.)

Ericson2314:
I have recently learned that everything I thought about data families is wrong. One consequence of that is I am not longer worried about what data families do for this proposal. I say we simply allow wildcards in legacy signatures:

```haskell
data A :: _ -- error, no SAKS so valid kinds are unbounded

Type A :: Type ->Type ->Type
data A :: _ -- OK!

Type A :: Type ->Type ->Type
data A :: _ -> _ -- OK!

Type A :: Type ->Type ->Type
data A :: _ -> Type -> _ -- OK!

type F1 :: Bool -> Type
data family F

data instance F :: _ -- OK! Bool -> Type

type F2 :: k
data family F

data instance F :: _ -- error, despite family SAKS, kinds are still unbounded
```

The very last one we could allow if we made a class to enforce the "must in in Kind" restriction rather than doing it with an ad-hoc check, but that is a separate issue for a separate proposal.

JakobBruenker:
@Ericson2314 To be clear, am I understanding correctly that something like

```haskell
type Foo :: Type -> Bool -> Type
data Foo
```
should remain disallowed and wildcards are the replacement for this?

Ericson2314:
@JakobBruenker 

Yes. I thought about going further for GADT notation, but `data Foo` is valid as a both GADT and classic ADT, muddling things. I still hope the situation can be further improved as the language evolves, but I hope this can be a good and far less controversial step that gets us on our way.

JakobBruenker:
@Ericson2314 It does seem like a good step. I'll see about adjusting the proposal text tomorrow.

Ericson2314:
@int-index I should tag you as this is based on your arity idea for type families. I guess type families already don't need the `:: _`, but maybe the kind inference expectations around `type family F` were never as strong.

JakobBruenker:
@Ericson2314 I was in the process of changing the proposal but that made me realize that I'm actually not really able to properly write down what makes this less controversial than the original proposal.

If I remember my original reasoning about why GADTs in data families are a problem correctly, that is taken care of by your `F1` example, where you resolve the ambiguous kind by assuming the most general one. This approach seems like it could also be used with the original version of this proposal, I think?

Is the only thing that makes the wildcard necessary the ambiguity between GADT vs non-GADT syntax of something like `data Foo` (without any constructors)? If so, why is this ambiguity a problem? Couldn't we simply say that it works as a GADT (which would make more sense to me personally, since `where` can usually be omitted but `=` usually cannot)? Or are there downsides to this that make it the controversial part?

Ericson2314:
@JakobBruenker Sounds convincing me to me! Especially in combination with
```haskell
type F :: SomethingFancy
type family F
```
already working as I mentioned.

JakobBruenker:
@Ericson2314 I've been spending some time thinking about how far this should go.

The easiest solution would be to just say that you can have, for `Foo :: Type -> Type -> Type`, `data Foo :: _ where` and `data Foo where`, but not `data Foo t :: _ where` or `data Foo b where`.

The most complete and consistent solution would be the one where you could have both of those, as well as things like `data Foo :: Type -> _ -> Type` and `data Foo b :: _ -> _`. Especially with these genuine partial kind signatures (as opposed to just `_`) I'm not sure if there's issues in implementing I'm not thinking of, or if it's worth doing just for consistency.

I'm wondering if you have an opinion on this.

Ericson2314:
Well, partial type signatures have been a morass in the past. I never been satisfied that the implementation complexity of them is at all essential, but I haven't disproved it either.

\[My hunch in detail is just don't have separate "synthesizing" and "checking" algorithms, but instead always check against a type with holes, where the one-big-hole case amounts to synthesizing as before. Sure the whole thing might be more complicated, but the combing the two disparate code paths ought to make up for it.]

@goldfirere and @int-index have also made a good point in distinguishing `_` which can be deduced from the rest of the signature alone, vs those which need to be compared to the thing being given a type.

Make of that what you will.

JakobBruenker:
@Ericson2314 I went with not allowing fully partial kind signatures for the time being. Do you think using a wildcard should require `-XPartialTypeSignatures`? (Mostly I'm wondering if you had an example in mind where that's not necessary when you wrote "no signature can be the same as :: _, just like the rest of the language!')

Ericson2314:
I am not really sure. I've been making many claims; I hope some else can jump in here to offer their perspective.

JakobBruenker:
Rendered largely obsolete by https://github.com/ghc-proposals/ghc-proposals/pull/448, which has in variables in the header of a GADT declaration scope over all constructor declarations, thus giving them a use. Thus, closing for now.

Ericson2314:
@JakobBruenker no I don't think so! @nomeata and I both think this is a gap in @goldfirere's reasoning, and I at least see #311 as good reason to hope we are not mistaken :).

JakobBruenker:
@Ericson2314 fair enough; I suppose I'll reopen at the very least until that's resolved

JakobBruenker:
One thing that I'm not sure how to resolve here is that the variables in a GADT header can be used in deriving clauses.

Ericson2314:
Hmm? Is there an example earlier in the thread?

JakobBruenker:
@Ericson2314 No, I just happened to see it mentioned in https://github.com/ghc-proposals/ghc-proposals/pull/448#issuecomment-957591916.

Here is an example:

```haskell
data Test a where
  A :: Test b
  deriving Show via Identity (Test a)
```

The `a` in the deriving clause is bound in the header.

JakobBruenker:
Wait, disregard - this still works if I rename a

JakobBruenker:
@Ericson2314 Okay, sorry for the previous noise, here's a better example:
```haskell
newtype Pair a b
  where Pair :: (x, y) -> Pair x y
  deriving Show via (a, b)
```

works. If you change the variable names, it doesn't work, because the types don't match anymore, as in

```haskell
newtype Pair a b
  where Pair :: (x, y) -> Pair x y
  deriving Show via (c, d)
```

Ericson2314:
Ugh. I'm tempted to just be stern and say one had to use standalone deriving.

JakobBruenker:
I'm wondering if it would be possible to uniquely infer the correct type variables for the deriving clauses in all cases where you would need the variables from the header. I suspect that's not the case, but without any justification.

In any case, it wouldn't be too big a deal to not handle this case, even ignoring the possibility to use standalone deriving instead, since this proposal doesn't actually disallow adding type variables to the header, so you could always add them only where they're really needed for a deriving clause.

PR #314, opened 2020-02-27
Enable `-Wnoncanonical-monad-instances` and `-Wnoncanonical-monoid-instances` by default (under review)

fumieval:
[Rendered](https://github.com/fumieval/ghc-proposals/blob/master/proposals/0000-warn-noncanonical-monad-instance-by-default.md)

### DIFF
--- a/proposals/0000-warn-noncanonical-monad-instance-by-default.md
+++ b/proposals/0000-warn-noncanonical-monad-instance-by-default.md
@@ -62,8 +62,8 @@ With this proposal implemented, it would produce the following warning:
 
 /path/to/example.hs:12:3: warning: [-Wnoncanonical-monoid-instances]
     Noncanonical ‘mappend’ definition detected
-    ‘mappend’ will eventually be removed from the ‘Monoid’ class in favour of ‘(<>)’.
     in the instance declaration for ‘Monoid (P a)’.
+    ‘mappend’ will eventually be removed from the ‘Monoid’ class in favour of ‘(<>)’.
     Either remove definition for ‘mappend’ (recommended) or define as ‘mappend = (<>)’
     See also: https://gitlab.haskell.org/ghc/ghc/-/wikis/proposal/semigroup-monoid
    |
--- a/proposals/0000-warn-noncanonical-monad-instance-by-default.md
+++ b/proposals/0000-warn-noncanonical-monad-instance-by-default.md
@@ -5,9 +5,7 @@ ticket-url: ""
 implemented: ""
 ---
 
-This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/0>).
-**After creating the pull request, edit this file again, update the number in
-the link, and delete this bold sentence.**
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/314).
 
 # Enable `-Wnoncanonical-monad-instances` by default
 
--- a/proposals/0000-warn-noncanonical-monad-instance-by-default.md
+++ b/proposals/0000-warn-noncanonical-monad-instance-by-default.md
@@ -53,19 +53,24 @@ With this proposal implemented, it would produce the following warning:
 ```
 /path/to/example.hs:12:3: warning: [-Wnoncanonical-monad-instances]
     Noncanonical ‘return’ definition detected
-    in the instance declaration for ‘Monad P’.
-    Either remove definition for ‘return’ or define as ‘return = pure’
+    ‘return’ will eventually be removed from the ‘Monad’ class in favour of ‘pure’.
+    Either remove definition for ‘return’ (recommended) or define as ‘return = pure’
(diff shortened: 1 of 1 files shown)


RyanGlScott:
Is there a reason to enable `-Wnoncanonical-monad-instances` by default but not `-Wnoncanonical-monoid-instances`? (The latter warns about `Monoid` implementations where `mappend` is not implemented as `(Data.Semigroup.<>)`.)

fumieval:
Why not :) I updated the proposal

proposals/0000-warn-noncanonical-monad-instance-by-default.md
```diff
@@ -0,0 +1,78 @@
...
+  fmap _ _ = P
+
+instance Applicative P where
+  pure _ = P
+  _ <*> _ = P
+
+instance Monad P where
+  return _ = P
+  _ >>= _ = P
+```
```
RyanGlScott: It would be helpful to include a `Monoid` instance that triggers `-Wnoncanonical-monoid-instances` in this example.

RyanGlScott:
Don't forget to update the title of the PR itself!

BlackCapCoder:
I would not enable a warning that is trying to notify me of one specific thing that I am currently unaware of, nor enable it once I have become aware and have already adapted.

I might have the catch all -Wdeprecations on, but it does not warn about non-canonical instances.

Perhaps there should be a '-Wbase-deprecations' flag or similar- which should absolutely be on by default for consistency with deprecated language extensions.

fumieval:
How am I supposed to move this forward?

fumieval:
This proposal is ready for review @nomeata

nomeata:
Thanks, assigned to @i-am-tom

i-am-tom:
This proposal seems wholly sensible to me:
- The change is likely to cause any huge disruption to existing code.
- The warnings already exist in the compiler, so the implementation cost is very low.
- Should the proposal be accepted, users can recover previous behaviour if need be.

Ultimately, I agree wholeheartedly with the motivation you've cited (for both `Monad` and `Monoid`), so I'd like to recommend that this proposal be accepted. 👍

I'll update the issue accordingly, and open up a discussion with the rest of the committee.

simonpj:
Fine with me!

goldfirere:
I do not support this proposal in its current form: the impact on Haskell learners is too great, for me. Essentially any Haskell resource written before ~2 years ago will have `return` definitions for `Monad`. I think I would be happier with a friendlier error message, perhaps without the word "noncanonical" in it. Best would be an error message with a URL that describes the historical arc of `Monad`, putting the suggestion to strike the `return` definition in context.

With an improvement in the error message, I would support this proposal.

fumieval:
I actually started writing a patch to see what kind of change we need. Now I'm unsure which URL it should refer to; GHC wiki pages point at https://prime.haskell.org/wiki/Libraries/Proposals/SemigroupMonoid for SMP, but it seems to be dead.

fumieval:
@goldfirere I'd argue that this proposal doesn't make it much worse in that respect, since learners who use GHC >=8.4 are already affected by `Monad` and `Monoid` being subclasses of `Applicative` and `Semigroup` respectively. Would you be happier with something like this?

```
/path/to/example.hs:12:3: warning: [-Wnoncanonical-monad-instances]
    Noncanonical ‘return’ definition detected
    in the instance declaration for ‘Monad P’.
    ‘return’ will eventually be removed from the ‘Monad’ class in favour of ‘pure’.
    Either remove definition for ‘return’ (recommended) or define as ‘return = pure’
    See also: ${URL}
   |
12 |   return _ = P
   |   ^^^^^^^^^^^^
```

goldfirere:
@fumieval Yes, I know about the superclass stuff. I just figure that's easier to search for, etc. And the error message doesn't use a technical term that many learners might not yet know.

(Somewhat separately, even knowing what non-canonical means, I don't think it's correct here. For example, if I write `newtype Reader' r a = Reader' (r -> a)`, there is one canonical `Monad` instance. If I write a different one (but still avoid defining `return`), I won't get this warning. And if I write `return = coerce (const @a @r)`, that *is* canonical, but I get the warning. What we mean is "non-forward-compatible". But I suppose the warning names are already enshrined and certainly not worth changing.)

In any case, the suggested error message above is a vast improvement. With that in place, I support this proposal. Thanks.

fumieval:
Thank you for supporting. I agree that it's not worth changing the term "noncanonical" although it's a bit misleading. Anyway I updated the warning examples; meantime I scavenged the semigroup monoid proposal [back in the wiki](https://gitlab.haskell.org/ghc/ghc/-/wikis/proposal/semigroup-monoid).

i-am-tom:
Hi @fumieval,

Thanks for your patience. I've discussed the proposal with the committee, and the consensus is that the proposal in its current form should be accepted! @nomeata, I think we're all set to merge.

Thanks for your proposal!

fumieval:
For the record: this is implemented in https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3174 and will be included in GHC 8.12.1

PR #316, opened 2020-03-08
DefaultExportAll option to change default module headers

cdsmith:
[Rendered](https://github.com/cdsmith/ghc-proposals/blob/master/proposals/0000-default-export-all.md)

### DIFF
--- a/proposals/0000-default-export-all.md
+++ b/proposals/0000-default-export-all.md
@@ -5,9 +5,7 @@ ticket-url: ""
 implemented: ""
 ---
 
-This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/0>).
-**After creating the pull request, edit this file again, update the number in
-the link, and delete this bold sentence.**
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/316>).
 
 # Default Export All
 
--- a/proposals/0000-default-export-all.md
+++ b/proposals/0000-default-export-all.md
@@ -50,9 +50,12 @@ implied module header just `module Main where` instead, without the export list)
 
 ## Proposed Change Specification
 
-A new language extension will be defined called `DefaultExportAll`.  When
-enabled, the default module header will become `module Main where` instead
-of `module Main (main) where` as it is now.
+A new language extension will be defined called `DefaultExportAll`.  When this is
+enabled and GHC compiles a module that does *not* contain an explicit module header,
+the default module header will become `module Main where` rather than
+`module Main (main) where` as it is in the Haskell Report.
+
+If the module has an explicit module header, there is no change in behavior.
 
 ## Examples
 
--- a/proposals/0000-default-export-all.md
+++ b/proposals/0000-default-export-all.md
@@ -68,8 +68,15 @@ I am aware of no unwanted interactions.
 
 The implementation itself should be approximately a single-digit number of
 lines of code.  The documentation and test cases will be slightly larger.
-The greatest cost is probably just keeping track of yet another language
-extension name.
+
+A second cost is just keeping track of yet another language extension.
+
+Finally, there's a potential performance cost to enabling this.  Joachim
+Breitner has found, in a different situation, [a gain of 11% in performance]
+(http://www.joachim-breitner.de/blog/762-Faster_Winter_4__Export_lists)
+just by adding a module header with explicit exports.  This is because GHC
+can cheaply inline definitions that are only used once, but exporting a
+definition from the module breaks this.
 
 ## Alternatives
 
@@ -122,6 +129,9 @@ straight-forward, since it means a compiled module needs to remember whether it 
 an explicit module header or not.  This seems like overkill for something that I
 would guess is going to break a single-digit (where that digit might be 0) number
 of people's code.
+
+On the other hand, the inlining performance cost mentioned above would apply to a lot
+more programs, so it's likely this shouldn't be the default.
 
 ## Unresolved Questions
 
--- a/proposals/0000-default-export-all.md
+++ b/proposals/0000-default-export-all.md
@@ -89,12 +89,12 @@ The first seems unlikely, but the second is known to happen.  A typical case
 involves something like this:
 
 ``` haskell
-module Main (main) where
-
+-- Main.hs
 main = putStrLn "Doing work..."
 ```
 
 ``` haskell
+-- Test.hs
 module Test where
 
 import Main
@@ -106,11 +106,11 @@ testMain = do
   putStrLn "Check assertions"
 ```
 
-Now the tests can be run by compiling with `-main-is Test.testMain`.
+The test can be run with `-main-is Test.testMain`.
 
 This pattern is used by some people who notice when it's broken.  For instance, see https://gitlab.haskell.org/ghc/ghc/issues/15702.  Neither the examples here nor there
-would be broken by the change, but they are both vulnerable to name collisions after
-this change.
+would be broken outright by this change, but they would become vulnerable to name
+collisions after this change is made.
 
 So it's possible this might break some code, but we cannot find any examples of code
 that would actually break.  If it did break, then there are two easy fixes: either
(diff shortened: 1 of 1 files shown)


erikd:
> When enabled, the default module header will become module Main where instead of module Main (main) where as it is now.

I think it may help to explicitly state that this proposal *only* comes into play in the absence of a `module` header.

cdsmith:
@erikd Thanks for the suggestion.  I've updated the wording.

goldfirere:
In your imperfect options, you mention the possibility of students writing a module header with an export list. Why must they include the export list? If they were just instructed to write `module Main where` at the top, that would exactly mimic your desired extension. I agree that this is three more words than you might want, but they're not particularly scary words. And they give you an opportunity to describe the benefits of modularity, which is (in my opinion) a concept that can conceivably be mentioned (if not internalized) on day one. (To wit: "A 'module' is just one small part of a big piece of software. In this class, you'll be writing the main part of the program, so you'll write `module Main where` at the top of every file." Explaining `where` is harder, especially if students don't know what a dependent clause is in English, I admit.)

If we were to accept this proposal, I would advocate for just having this be the new behavior, with no extension. It might break code, but the breakage can be fixed in a backward-compatible way. And the new behavior makes more sense to me: you would expect that an omitted module header would mean the same thing as a module header with an omitted export list. It's weird that omitting the export list and omitting the whole header move the "how much is exported?" needle in opposite directions.

nomeata:
> If we were to accept this proposal, I would advocate for just having this be the new behavior, with no extension.

That would be in voliation of the Haskell report: ([last sentence of 5.](https://www.haskell.org/onlinereport/haskell2010/haskellch5.html#x11-990005.1))

goldfirere:
> That would be in voliation of the Haskell report

Yes, I know. I should have clarified that I was advocated for going against the Report. It is not the first time.

cdsmith:
Yes, you're right that students could add an explicit header with no export list.  I would still make the choice not to implement the features I want rather than as the boilerplate.  (If you're adding boilerplate anyway, the boilerplate may as well be of the form "main = realMain foo bar", rather than "module Main where".)

I agree that changing the default behavior would be better if we were starting from scratch, but this seemed like the first step to that.  Surely we can't just break people's projects suddenly; there need to be a migration period and warnings and such?  Though we're probably talking about an extremely small amount of broken code here.

goldfirere:
We can and do break people's projects suddenly. :) But we try to contain the breakage. And we try desperately to make it possible for projects to work over several versions of GHC, without the use of CPP. In this case, the fix is to add a module header that says what you want. This is fully backward compatible. And so it's acceptable to break code in this way. If it weren't for the Report, I'd personally not even consider the idea of adding an extension to control this behavior... but contravening the Report does give me pause.

cdsmith:
I've updated the alternatives to expand the argument for not having a flag at all.  Also clarified the oversight @goldfirere mentioned in motivation.

nomeata:
There is a benefit to the `module Main(main) where` default. The most common use case for omitting the header is when you are writing a program, where you really only care about `main` as the entry point.

If the header is `module Main(main)`, the compiller can freely inline all other definitions and do other optimizations that are only possible if all use-sites are visible.

If the header is `module Main`, then some optimizations will be prevented.

The effect can be significant in some cases (e.g. I observed [11% improvement in one instance](http://www.joachim-breitner.de/blog/762-Faster_Winter_4__Export_lists)).

Maybe the compiler can be taught to recognize when it is really building the main of a program, and be smart about it. But unless that is the case, it seems that changing the default to `module Main` would possibly slow down some existing programs out there, and that seems to be a bad trade off.

cdsmith:
@nomeata Very good point!  Thanks for bringing that up.  I think that settles the question of turning it on by default, then.  Clearly we shouldn't if it would lead to significant performance costs in a significant amount of existing code.

The remaining question is whether GHC is interested in having this as an option.  The other alternative is that I could patch the GHCJS build that's used for CodeWorld, and maintain the patch indefinitely on my side.  Since it's neither a lot of code nor a fast-changing part of the compiler, that's actually okay with me.

nomeata:
> performance costs in a significant amount of existing code.

I didn't say that :-). Maybe “performance costs in an insignificant amount of existing code, and inperformance costs in an significant amount of existing code”…

> I could patch the GHCJS build that's used for CodeWorld

While I am sympathetic towards CodeWorld, I would argue that without at least another independent use case, this might be the better course of action.

Ericson2314:
@nomeata I do not understand your performance argument. The compiler is always free to do as much intra-module inlining as it wants, provided it also provided outlined versions of all exported functions. Perhaps it is less likely to inline when it must also provide the outlined version, but this is a matter of changing the heuristic---not terribly difficult or fundamental.

Ericson2314:
There's a related (subsuming?) change I've wanted, but haven't bothered to write a proposal for: not having modules know their own names, so one writes `module (...) where` instead of `Module Foo (...) where`. The name would instead come from the file path, and only be relevant for downstream modules that import oneself. Knowing your own name is almost always *bad thing*, and this also means that that irregularly named files can be anonymous root modules rather than be given the arbitrary name `Main`.

phadej:
:-1: on module name coming from file path. We import module by names, so their names are very well relevant. Where they are in the file system is implementation detail (e.g. which report doesn't talk about).

Fun, and not well known fact:

```haskell
% ls                     
Foo.Bar.hs  Main.hs

% cat Foo.Bar.hs         
module Foo.Bar where

answer :: Int
answer = 42

% cat Main.hs            
import Foo.Bar

main = print answer

% ghci Foo.Bar.hs Main.hs
GHCi, version 8.6.5: http://www.haskell.org/ghc/  :? for help
[1 of 2] Compiling Foo.Bar          ( Foo.Bar.hs, interpreted )
[2 of 2] Compiling Main             ( Main.hs, interpreted )
Ok, two modules loaded.
*Foo.Bar> Main.main
42
```

;)

nomeata:
The compiler is free to have inline and keep the exported version around, but it is expensive (code size) and thus the compiler might not want to inline in these cases.

Ericson2314:
@phadej Let me clarify, I am *not* advocating that report should talk about file paths. I am instead saying that the "mysterious out of band data structure that holds all the modules" should also name them.

> Perfectly, build tool should give all the source files to compiler, and it will not do any discovery.

That actually meshes really well with my plan.

 If the compiler is told "no don't look for things", it *has* to be at the mercy of the build tool as to which module is which, whereas with names it is free to follow the letter of the all ("I don't look for files!") but not the spirit ("hmm, these modules have these names, let me `upsweep`")

Ultimately, we both (I think) don't just want the build tool to know the set of modules, but know all their dependencies and do all the `ghc --make` stuff, so GHC is the "dumn compiler" the does no build planning. Anonymous modules are perfect for this.

cdsmith:
@Ericson2314 I'm definitely on board with this in spirit.  In fact, I just moved mountains to implement an import-by-hash feature in my educational environment (e.g., https://code.world/haskell#PpTBzvD_QX9ITJZYMI_m6ag) which doesn't have file management features.  Unfortunately, naming a module after its hash in the source code is a tricky problem, so I had to parse, rewrite, and reserialize the source code with a different module name.  Ugh!  Would much rather the source didn't care which default name it's imported as, and just let me identify it in the build tool.

However, it seems certainly much too late in Haskell's lifecycle for a change like that!

Ericson2314:
@cdsmith Oh, I don't think it's too late! Maybe I'll write one then :). [Also, in the meantime you could use some of the backpack renaming features perhaps to help with that.]

cdsmith:
@Ericson2314 I'm going to go ahead and close this, since it doesn't look worth doing.  Adding special behavior for the main module when deciding on inlining goes beyond my offer to implement the change anyway, given that I don't know GHC that well.

Don't take that as a lack of support for other alternatives.  Just a recognition that this one isn't going to happen.

cdsmith:
In case anyone else ends up here and is interested in this: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/292/diffs appears to have actually made almost this same change, independent of this proposal.  One difference is that the pull request only defaults to exporting all symbols if the module does not define `main`.  This addresses Joachim's concern, and also only makes the compiler more permissive, rather than changing the meaning of existing valid code.

PR #320, opened 2020-03-26
Ignore class defaults for associated type instances in instances that occur in backpack and hs-boot files (under review)

cgibbard:
[Rendered](https://github.com/obsidiansystems/ghc-proposals/blob/signature-type-instances/proposals/0000-signature-instances.md)


goldfirere:
Very nice write-up. It's unfortunate that the handling of defaults is backwards in regular files vs boot files/signatures, but life is sometimes backwards.

I support.

Ericson2314:
I like to mumble "polarity" to myself until I'm numb to the backwards part :).

cgibbard:
Yeah, signatures/hs-boot files are essentially module-shaped patterns rather than expressions, so you expect some differences to arise.

Thus far, this seems uncontroversial and relatively tiny. Any way that we can get it moved along so that the corresponding GHC PR can get merged?

goldfirere:
Just submit this to the committee, I think.

cgibbard:
Pinging @nomeata I suppose? :)

nomeata:
/remind @goldfirere to check for progress in two weeks

goldfirere:
As shepherd, I recommend acceptance.

This proposal addresses an expressiveness gap in Haskell: a class with an associated-type default cannot suppress that default in an hs-boot or backpack signature file. The solution is to ignore associated-type defaults in hs-boot and backpack signature files.

Upsides:
 * Solves the problem.
 * Simple specification.
 * Forward-compatible with support for associated type instances in hs-boot and signature files.

Downsides:
 * Yet another small wrinkle in the different handling of hs-boot/signature files from normal Haskell.

Let's do it.

goldfirere:
/remind me in one week to accept unless debate has sprung up.

simonpj:
The change specification says:

> In backpack signatures and hs-boot files, any instance item which occurs without mention of an class item simply places no constraint on that type family, regardless of whether the class has a default for that item.

But 
* What is "an instance item"?  I'm guessing a class instance declaration is intended here?
* What is a "class item"?  What is "that type family" referring to?
* The wording implies that a class instance decl in a hs-boot or backpack sig *can* sometimes mention a class item.  I think the intent is that hs-boot class instances simply have no 'where' part.  Ever. Is that right?

-------------

Separately, the immediately following Example (just after Change Spec) is confusing. It speaks of stuff in the where clause of of a class instance, but I think the real burden of the proposal is that the class instance (with no where part) carries no type-family-instance stuff at all.

Can there be standalone type-family instance decls in a hs-boot or backpack file?

Ericson2314:
@simonpj 

> The change specification

Sorry the proposed change was all messed up. This was an emitting error on my part when we generalized from just talking about associated type families to class items in general. Should be clear now.

> Can there be standalone type-family instance decls in a hs-boot or backpack file?

Not today, but there are issues open for implementing this. The basic idea is any definition can be interpreted as an equality constraint instead, which actually makes to put in a signature. Type family definitions are already desugared into quotienting equality constraints, so this is especially natural for them, but the same reasoning applies to any other sort of definition.

Relatedly, instances are really wanted constraints, so "instances" for `(~)`, `Coercible`, and other magic classes also can make sense.

Notwithstanding everything I just wrote above, I do think it is still up in the air whether we want to allow any more definitions in hs-boot files. \[Maybe we we even allow too much already!\] The `(~)` instance syntax is indeed more flexible, and perhaps makes clearer what is really going on, so that might be preferred, for example. 

> It speaks of stuff in the where clause of of a class instance, but I think the real burden of the proposal is that the class instance (with no where part) carries no type-family-instance stuff at all. So the example isn't an example of what the proposal is about. Unless I have misunderstood.

Given that it is unclear whether we want those definitions, this proposal isn't meant to weigh in one way or the other on whether definitions ought to be allowed in hs-boot files and signatures. It's supposed to just be about what the semantics definition within an instance, regardless of whether omitting the definition is mandatory or obligatory.

simonpj:
Would it be possible to improve the proposal so that it makes all this clear?  Then I'll take another look at it.

Could you also state clearly (in the proposal) what that status quo is concerning:
* Standalone type instance decls
* Class instances with a where clause

Maybe these decisions could be changed, but to read *this* proposal I need to be clear about what the baseline is, and I'm regrettably paged on in this area.  Thanks!

Ericson2314:
@simonpj Sure. I already reworded that the proposal's one sentance to actually make sense. We'll go work the rest, I suppose in the effects and interactions section.

Ericson2314:
We'll will put them in the proposal, but in the meantime if anyone is curious here are the tickets I mentioned:

 - https://gitlab.haskell.org//ghc/ghc/issues/12680 Permit type equality instances in signatures

 - https://gitlab.haskell.org//ghc/ghc/issues/8441 Allow family instances in an hs-boot file

cgibbard:
I made some edits to hopefully clarify the proposal. This proposal really has nothing to do with allowing items in instances, though we want to recognise that there's already discussion in that direction (especially for things such as type equalities which are "proof-irrelevant").

simonpj:
Did you manage to account for my request?

> Could you also state clearly (in the proposal) what that status quo is concerning: (a) Standalone type instance decls, (b)  Class instances with a where clause

I can't see it.  Moreover, the Example after Change Specification has one of these instance decls with a where clause in a hs-boot file.  I think they are illegal, and in any case the example doesn't illustrate the effect of the proposal. Could you give an example that *does*.   

I *think* it may look like this:
```
A.hs
         module A where
         class C a where
             type F a
             type F a = a -> a    -- Default instance
             op :: a -> a

         foo :: C a => [a] -> F a
         foo = blah

B.hs-boot
        module B where
         import A
         instance C Int

C.hs
       module C where
       import {-# SOURCE #-} B

       wim :: [Int] -> F Int
       wim = foo                 -- OK

       wam :: [Int] -> Int -> Int
       wam = foo                 -- NOT OK

B.hs
        module B where
        import A
        instance C Int where
             type F Int = [Int]
             op = reverse
```
The `instance C Int` in B.hs-boot is legal, does *not* imply `instance F Int = Int -> Int`, despite the default associated type decl in the original defn of C.    Indeed, the final `instance C Int` in `B.hs` (shown above) might use an entirely different `type instance` than the default!

This is all the same as class methods.  The existence of a default method for `op` in the defn of C does not imply that the C Int instance (defined in B.hs presumably) uses that default method. 

I believe that this "legal but does not imply" is the entire burden of this proposal.

An example like this would be fantastically helpful.

Ericson2314:
@simonpj Sorry, we both missed your comment somehow. That example looks great so we'll put i in there.

cgibbard:
I added the example to the proposal, replacing the previous one.

simonpj:
The "discussed at this pull request" link in the rendered proposal is broken

simonpj:
> I added the example to the proposal, replacing the previous one.

OK that makes it much better.

I think that (1) is always what was intended.  I think it's a plain bug in GHC that there's any monkeying with family instances in hs-boot or hsigs.   So I support the proposal, but I see it mainly as clarifying what is supposed to be the case already, and bringing to light a current GHC bug.  It's not a language change.  

But still, having it written up a bit helps.  Let's just fix the bug!

FWIW, I didn't find alternatives (2) and (3) comprehensible.  As I say in the example, I can't make sense of having method or associated-type instances in hsig files, period.

Anyway, onward!

goldfirere:
This proposal is now accepted.

barci2:
Since the proposal has been implemented, shouldn't it receive an "Implemented" label?

nomeata:
Thanks!

PR #321, opened 2020-03-28
Implicit module names

Ericson2314:
Allow plain `module where`, inferring the name from context.

[Rendered](https://github.com/Ericson2314/ghc-proposals/blob/implicit-module-names/proposals/0000-implicit-module-names.rst)

### DIFF
--- a/proposals/0000-implicit-module-names.rst
+++ b/proposals/0000-implicit-module-names.rst
@@ -1,0 +1,112 @@
+Notes on reStructuredText - delete this section before submitting
+==================================================================
+
+The proposals are submitted in reStructuredText format.  To get inline code, enclose text in double backticks, ``like this``.  To get block code, use a double colon and indent by at least one space
+
+::
+
+ like this
+ and
+
+ this too
+
+To get hyperlinks, use backticks, angle brackets, and an underscore `like this <http://www.haskell.org/>`_.
+
+
+Proposal title
+==============
+
+.. author:: Your name
+.. date-accepted:: Leave blank. This will be filled in when the proposal is accepted.
+.. ticket-url:: Leave blank. This will eventually be filled with the
+                ticket URL which will track the progress of the
+                implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. contents::
+
+Here you should write a short abstract motivating and briefly summarizing the proposed change.
+
+
+Motivation
+----------
+Give a strong reason for why the community needs this change. Describe the use
+case as clearly as possible and give an example. Explain how the status quo is
+insufficient or not ideal.
+
+A good Motivation section is often driven by examples and real-world scenarios.
+
+
+Proposed Change Specification
+-----------------------------
+Specify the change in precise, comprehensive yet concise language. Avoid words
+like "should" or "could". Strive for a complete definition. Your specification
+may include,
+
+* BNF grammar and semantics of any new syntactic constructs
+* the types and semantics of any new library interfaces
+* how the proposed change interacts with existing language or compiler
+  features, in case that is otherwise ambiguous
+
+Note, however, that this section need not describe details of the
+implementation of the feature or examples. The proposal is merely supposed to
+give a conceptual specification of the new feature and its behavior.
+
+Examples
+--------
+This section illustrates the specification through the use of examples of the
+language change proposed. It is best to exemplify each point made in the
+specification, though perhaps one example can cover several points. Contrived
+examples are OK here. If the Motivation section describes something that is
+hard to do without this proposal, this is a good place to show how easy that
+thing is to do with the proposal.
+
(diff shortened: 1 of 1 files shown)


proposals/0000-implicit-module-names.rst
```diff
@@ -0,0 +1,146 @@
...
+
+Implementation Plan
+-------------------
+
+I (@Ericson2314) can do it, though I am happy to help someone else do it as it's a good easy training ticket.
+
+Endorsements
+-------------
+
+.. _`Proposal 316`: https://github.com/ghc-proposals/ghc-proposals/pull/a316
```
Anrock: Link is borked due to stray `a`.

nomeata:
There are use-cases for referring to the current module inside this module; export lists and qualified names.

So can you still do something the equivalent of
```
module A (module A) where …
```
or 
```
module A where
tail = undefined
foo = A.tail -- disambguate
```
with your extension? Do you then again need to “guess” the right module name, or do you envision some way of referring to “the present module” without knowing its name?

proposals/0000-implicit-module-names.rst
Ericson2314: Thanks1

Ericson2314:
@nomeata That is a good question. The simplest implementation would allow one to guess the module name, but I think I would rather have a dedicated way to refer the the current module. I haven't thought of a name yet, but I guess that could be a part of this proposal.

Ericson2314:
What about plain `module`? Kinda silly, but I think ti works:

```haskell
module A (module) where …
```
```haskell
module A where
tail = undefined
foo = module.tail -- disambguate
```

iamrecursion:
Just a brief thought, `here` could also work as an identifier for the current module. 

```haskell
module where
tail = undefined
foo = here.tail -- disambguate
```

goldfirere:
I don't find the motivation convincing here. I acknowledge the comment at the end of the proposal that aims to be more future-oriented than history-oriented, but still don't see enough of the eventual benefit to warrant yet another extension.

Also: the module name matters, because it is what is used in `import` statements. It's disappointing that this proposal means that the text of my program is not only a set of strings (the module contents) but now, necessarily, a mapping from strings (file paths) to strings (file contents).

Ericson2314:
@iamrecursion I was hoping to avoid introducing yet another contextual keyword. Also, if the record dot notation passes the `.` grammar will be far more complex and `here.thing` will probably be taken.

@goldfirere 

> Also: the module name matters, because it is what is used in `import` statements. It's disappointing that this proposal means that the text of my program is not only a set of strings (the module contents) but now, necessarily, a mapping from strings (file paths) to strings (file contents).

I don't like strings or the file system either? I ideally would like to edit the AST of my entire package a thoroughly de-unixed IDE, but as long as modules contain their names that AST is going to be denormalized---I'm too "constructivist" to think of a package as a set of modules whose names happen to not overlap, I really want the map as part of the AST, and ideally nested maps for hierarchical module names.

Ericson2314:
That last bit reminds me, there is also some relation to the local modules / first class modules / structured imports exports proposals. There was some interest in a uniform binding syntax like `module <name> = <sub module|import>` from @int-index starting in https://github.com/ghc-proposals/ghc-proposals/pull/205#issuecomment-524619444, to make clear that nested modules and qualified imports bind in the same namespace (and also make the combining of qualified imports explicit).

galenhuntington:
How about using `as`?

```haskell
module as A where
undefined = A.undefined
```

That could also be useful in named modules:

```haskell
module My.Long.Module.Name.That.Might.Change as A where
undefined = A.undefined
```

Ericson2314:
At first I wasn't excited about `module as`, but then it came up again completely coincidentally in https://github.com/ghc-proposals/ghc-proposals/pull/270 / https://github.com/ghc-proposals/ghc-proposals/pull/340, so now I'm on board if those go through.

soundlogic2236:
> It's disappointing that this proposal means that the text of my program is not only a set of strings (the module contents) but now, necessarily, a mapping from strings (file paths) to strings (file contents).

I would like to point out that this is already dubious in practice. While one can use ghc to directly compile individual files such that their file name is ignored, ghc itself contains the code to cause a compile failure in other cases, and while I do not have actual data, I expect the latter mode is overwhelmingly more common.

This compile failure means that one cannot in fact determine the result of compilation from just the set of module contents, merely narrow it down to two possibilities: Either everything is in its "proper" place and you compile, or something isn't and you don't. 

While a compiler that makes no claims about if the system it is running on has a file system is an interesting concept, and perhaps in the era of its creation could even have some use for interpreting things like printed out academic papers defined by nothing more than prose separating blocks of code with modules, I think the hierarchical file system's use in programming is at this point very much committed to, as demonstrated by the fact that ghc contains code to respond differently based on its status.

Alternatively, a file system agnostic ghc could also be aimed for, rejecting that trend. A system that permits multiple modules per file and module names that don't match the filename would also make sense, though it seems to me to have a lot of disadvantages. 

This middle path however seems much harder to justify than either of these. If modules unlinked from files are meant to be the primary structure, why is `EOF` treated uniquely in the grammar, when we have already gone to the work of defining a way to separate modules with the `module` syntax? Even in the mode where a name that doesn't match the filename is permitted, such a thing is currently forbidden for no clear reason.

Finally, one color of bikeshed that seems to me to have precedent but doesn't seem to have been listed here is:
```haskell
module _ (foo, bar) where
```

PR #324, opened 2020-04-09
Allow direct access to underlying concrete class dictionaries

tysonzero:
[Rendered](https://github.com/tysonzero/ghc-proposals/blob/concrete-class-dictionaries/proposals/0000-concrete-class-dictionaries.md)

### DIFF
--- a/proposals/0000-concrete-class-dictionaries.md
+++ b/proposals/0000-concrete-class-dictionaries.md
@@ -1,0 +1,266 @@
+---
+author: Daniel Smith
+date-accepted: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/0>).
+**After creating the pull request, edit this file again, update the number in
+the link, and delete this bold sentence.**
+
+# Concrete class dictionaries
+
+Currently typeclass dictionaries are second class citizens in Haskell.
+They are used internally but they are not exposed properly to the user.
+
+This has led to a proliferation of extensions to help you define instances
+and convert instances between one another. Even with these extensions the
+ergonomics are still significantly worse than regular value level Haskell.
+
+For this reason I think we should instead focus on making the underlying
+dictionaries firsts class and accessible to developers.
+
+
+## Motivation
+
+There has been a large proliferation of typeclass related extensions, and yet
+the ergonomics for interacting with typeclass dictionaries is still poor.
+
+Something fundamentally quite simple such as defining a series of newtype
+wrappers over Int for modular arithmetic, with each using a different modulo for
+Num, is very painful.
+
+The specifical example that motivated me to write this proposal is the one
+given [here](https://mail.haskell.org/pipermail/haskell-cafe/2020-April/132083.html)
+and looks as follows:
+
+```Haskell
+{-# LANGUAGE DerivingVia, MultiParamTypeClasses #-}
+{-# LANGUAGE ScopedTypeVariables, TypeOperators #-}
+
+import Data.Function (on)
+import Data.Coerce (coerce)
+
+data Foo = Foo deriving Eq via (Foo `InjectedInto` Bar)
+data Bar = Bar deriving Eq
+
+foo2bar :: Foo -> Bar
+foo2bar Foo = Bar
+
+newtype InjectedInto a b = InjectedInto a
+
+class Injective a b where
+    -- Law: to x = to y => x = y
+    to :: a -> b
+
+instance Injective Foo Bar where
+    to = foo2bar
+
+instance (Eq b, Injective a b) => Eq (a `InjectedInto` b) where
+    (==) = (==) `on` (to :: a -> b) . coerce
+```
+
+With some follow up discussion about making libraries of various type level
+utilities to help with the above.
+
+Everything in the above example is really just manipulating value level code
+and thus I think it should be done on the value level.
+
+I would expect something closer to the following:
+
+```Haskell
+{-# LANGUAGE ConcreteClassDictionaries #-}
+
+import Data.Function (on)
+import Data.Coerce (coerce)
+
+data Foo = Foo deriving (Eq = eqOn foo2Bar)
+data Bar = Bar deriving Eq
+
+foo2bar :: Foo -> Bar
+foo2bar Foo = Bar
+
+eqOn :: ???
+```
+
+
+## Proposed Change Specification
+
+The proposal introduces a new extension `ConcreteClassDictionaries` that
+enables direct manipulation of typeclass dictionaries.
+
+First we add a new syntactic construct that allows for defining a type and
+constructor for classes:
+
+```Haskell
+class Eq a where
+    data dict EqDict a = Eq
+    (==) :: a -> a -> Bool
+    (/=) :: a -> a -> Bool
+
+class Semigroup a where
+    newtype dict SemigroupDict a = Semigroup
+    (<>) :: a -> a -> a
+```
+
+The above examples would create the following types:
+
+```Haskell
+data EqDict a = Eq
+    { (==) :: a -> a -> Bool
+    , (/=) :: a -> a -> Bool
+    }
+
+newtype SemigroupDict a = Semigroup
+    { (<>) :: a -> a -> a
+    }
+```
+
+The field names would collide with the class methods if they were generated,
+so we apply `NoFieldSelectors` to these data types.
+
+In the case of a user already having a data type available and wanting to
+create a class over the top of it, we define the following alternative syntax:
+
+```Haskell
+data Fold f = Fold
+    { foldr :: forall a b. (a -> b -> b) -> b -> t a -> b 
+    , foldl :: forall a b. (b -> a -> b) -> b -> t a -> b
+    }
+
+class Foldable f = fold :: Fold f
+
+class Default a = def :: a
+```
+
+This syntax does not need to use `NoFieldSelectors` as no class methods are
+created besides the singular `fold :: Foldable a => Fold a`.
+
+We now provide syntax for defining new instances directly from an appropriate
+concrete dictionary:
+
+```Haskell
+fromEq :: (a -> a -> Bool) -> EqDict a
+fromEq eq = EqDict
+    { (==) = eq
+    , (/=) = \a b -> not (eq a b)
+    }
+
+data Foo = Foo
+
+instance Eq Foo = fromEq $ \_ _ -> True
+
+instance Default Foo = Foo
+
+data Bar = Bar
+    deriving (Eq = eqOn foo2Bar, Default = Bar)
+
+eqOn :: Eq b => (a -> b) -> EqDict a
+eqOn f = fromEq ((==) `on` f)
+
+foo2Bar :: Foo -> Bar
+foo2Bar Foo = Bar
+```
+
+
+## Examples
+
+In the proposed changes I already covered the motivating equality example.
+So I will give the `Num` example as well:
+
+```Haskell
+class Num a where
+    data dict NumDict a = Num
+    (+) :: a -> a -> a
+    (-) :: a -> a -> a
+    (*) :: a -> a -> a
+    negate :: a -> a
+    abs :: a -> a
+    signum :: a -> a
+    fromInteger :: Integer -> a
+
+modularNum :: Num a => a -> NumDict a
+modularNum m = NumDict
+    { (+) = \x y -> (x + y) `mod` m
+    , (-) = \x y -> (x - y) `mod` m
+    , (*) = \x y -> (x * y) `mod` m
+    , negate = \x -> negate x `mod` m
+    , abs = \x -> x
+    , signum = \x -> signum x
+    }
+
+newtype Int7 = Int7 Int
+
+instance Num Int7 = coerce (modularNum @Int 7)
+
+newtype Int13 = Int13 Int
+    deriving (Num = coerce (modularNum @Int 13)
+```
+
+## Effect and Interactions
+
+With these additions, we have the full power of value level Haskell available
+to us when defining various ways to create and transform typeclass
+dictionaries.
+
+There is no longer any need to build up complex and verbose type level
+utilities in order to define various desired typeclass instances.
+
+We do not explicitly remove any existing typeclass machinary in this proposal.
+With that said this should significantly slow any desire to add new machinary
+around typeclass instance manipulation. In future we can revisit existing
+extensions to see which, if any, are obviated by this extension.
+
+Associated type and data families will largely be ignored by this proposal.
+The concrete dictionary types created will not contain the type or data
+families themselves, they will simply reference them as before.
+
+
+## Costs and Drawbacks
+
+I do not have a deep enough understanding of GHC to be able to precisely
+estimate the development and maintainence costs.
+
+With that said the above changes are purely syntax level, and should not have
+any effect on core or the underlying runtime.
+
+I do not anticipate significant learnability difficulties. I actually prefer
+this approach to the status quo for learnability, as it makes it clear that
+classes and instances are basically just extensible partial functions from the
+type level to the valeu level.
+
+
+## Alternatives
+
+* I originally considered making heavier use making heavier use of data families
+for the dict types to avoid having to declare a dict type. I ultimately decided
+against it on the basis that you would still need to decide on a constructor
+name and thus may as well have full control of both the type name and the
+constructor name.
+
+* I also considered providing an `IsClass (c :: Constraint)` class with a
+`use :: (IsClass c, c) => Dict c` but decided to do away with it once I decided
+I didn't want to use a data family. Particularly since there would be no way to
+avoid exporting this conversion without specifying extra rules about how that
+would work.
+
+
+## Unresolved Questions
+
+* Is it worthwhile to adjust/extend the first syntax to also allow for defining
+a function that converts directly from an instance to the underlying dict.
+Equivalent to `fold :: Foldable f => Fold f` from the second syntax.
+
+* Is it worthwhile to add an optional flag to the second syntax that will
+automatically create top level class methods for every field of the underlying
+type. This would allow you to avoid writing out the fields/methods twice just
+like we can do with the first syntax.
+
+
+## Implementation Plan
+
+No implementation plan at this time.
+
+
+## Endorsements
--- a/proposals/0000-concrete-class-dictionaries.md
+++ b/proposals/0000-concrete-class-dictionaries.md
@@ -136,8 +136,8 @@ class Eq a where
 
 defaultEq :: Eq a => EqDict a
 defaultEq = Eq
-  { (==) = \x y -> x /= y  -- using (Eq a)
-  , (/=) = \x y -> x == y
+  { (==) = \x y -> not (x /= y)  -- using (Eq a)
+  , (/=) = \x y -> not (x == y)
   }
 
 instance Eq T = defaultEq
--- a/proposals/0000-concrete-class-dictionaries.md
+++ b/proposals/0000-concrete-class-dictionaries.md
@@ -5,9 +5,7 @@ ticket-url: ""
 implemented: ""
 ---
 
-This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/0>).
-**After creating the pull request, edit this file again, update the number in
-the link, and delete this bold sentence.**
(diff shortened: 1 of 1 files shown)


edsko:
Related: https://dl.acm.org/doi/10.1145/3299711.3242752

PhilArmstrong:
I like this, but fear the unforeseen consequences!

cartazio:
One concern I have is that this Approach as defined could dictate internal dictionary representation choice.  Additionally my understanding is that type class dictionary data types have different strictness properties than normal userland datatypes

tysonzero:
@cartazio

The internal dictionary representation can stay the way it is and be different than the concrete dictionary representation if desired.

It just means the conversion between concrete dictionary and implicit dictionary wouldn't be a no-op, but I don't foresee that being a significant performance concern.

tysonzero:
For an example from production code:

```
instance PersistStoreWrite Environment where
    insert = contramapReaderT eBackend . insert
    insert_ = contramapReaderT eBackend . insert_
    insertMany = contramapReaderT eBackend . insertMany
    insertMany_ = contramapReaderT eBackend . insertMany_
    insertEntityMany = contramapReaderT eBackend . insertEntityMany
    insertKey = (contramapReaderT eBackend .) . insertKey
    repsert = (contramapReaderT eBackend .) . repsert
    repsertMany = contramapReaderT eBackend . repsertMany
    replace = (contramapReaderT eBackend .) . replace
    delete = contramapReaderT eBackend . delete
    update = (contramapReaderT eBackend .) . update
    updateGet = (contramapReaderT eBackend .) . updateGet
```

Could become:

```
instance PersistStoreWrite Environment = contramapPersistStoreWrite eBackend
```

Given a library-provided `contramapPersistStoreWrite :: PersistStoreWrite b => (a -> b) -> PersistStoreWriteDict a` function.

mikeplus64:
This would be really really good to have.

I think it might still be better to have a type/data family that automatically exposes the data constructor for the class instead of needing any boilerplate for it. I'm not overly fond on obligatory boilerplate that will looks identical 99% of the time. That said it's also probably not a great idea for `base` to introduce a swathe of new names that probably have (value-level) use already like `Num` (so even in the current iteration of the proposal good names should probably be decided). 

Maybe dict constructors can live in a magical module that exposes the class dicts of the classes that are sufficiently in scope already (i.e. it doesn't have hidden members you can't see) or even have a special syntax?

So roughly, magical module:
```haskell
module Foo where
import GHC.Dict (Dict(..), Nice(..)) -- ???!
-- import qualified GHC.Dict as X

-- data family Dict (x :: Type)

class Nice a where
  nicely :: a -> String

fakeNice :: Dict (Nice a)
fakeNice = Nice { nicely = const "oh no" }
-- fakeNice = X.Nice { nicely = const "oh no" }
```

Special syntax:
```haskell
fakeNice :: Dict (Nice a)
fakeNice = (=>Nice) { nicely = const "oh no" }
```

noughtmare:
I think many of the problems that this proposal aims to solve can also be solved by using pseudo dependently typed Haskell, for example the Num with a modulo can be implemented as follows:

```Haskell
{-# LANGUAGE GADTs #-}
{-# LANGUAGE KindSignatures #-}
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE PolyKinds #-}
{-# LANGUAGE InstanceSigs #-}

import GHC.TypeLits

data ModInt :: Nat -> * where
  ModInt :: Int -> ModInt n

instance KnownNat n => Num (ModInt n) where
  (+) :: forall n . KnownNat n => ModInt n -> ModInt n -> ModInt n
  ModInt a + ModInt b = ModInt ((a + b) `mod` fromIntegral (natVal @n undefined))

  (-) :: forall n . KnownNat n => ModInt n -> ModInt n -> ModInt n
  ModInt a - ModInt b = ModInt ((a - b) `mod` fromIntegral (natVal @n undefined))

  (*) :: forall n . KnownNat n => ModInt n -> ModInt n -> ModInt n
  ModInt a * ModInt b = ModInt ((a * b) `mod` fromIntegral (natVal @n undefined))

  negate :: forall n . KnownNat n => ModInt n -> ModInt n
  negate (ModInt a) = ModInt (negate a `mod` fromIntegral (natVal @n undefined))

  abs :: forall n . KnownNat n => ModInt n -> ModInt n
  abs (ModInt a) = ModInt a

  signum :: forall n . KnownNat n => ModInt n -> ModInt n
  signum (ModInt a) = ModInt (signum a)

  fromInteger :: forall n . KnownNat n => Integer -> ModInt n
  fromInteger n = ModInt (fromIntegral n `mod` fromIntegral (natVal @n undefined))

type Int7 = ModInt 7

type Int13 = ModInt 13
```

xplat:
the library `reflection` already provides most of this, and some things this doesn't, do we really need it at the language level?  At the very least the proposal should explain why `reflection` doesn't cut the mustard, if it doesn't...

tysonzero:
@mikeplus64 

The reason I wanted to allow for user defined dictionary types is because sometimes the underlying dictionary will be a meaningful type of its own worthy of a proper name. For example `Traversal` for `Traversable`, `PersistStoreWriter` for `PersistStoreWrite`, `a` for `Default a`.

It's also worth noting that `Dict (Nice a)` would not have a `Contravariant` instance but `NiceDict a` would be able to support one.

With that said I'm not against looking into ways to allow people to access the dict of a fully exported class without requiring changes to the class, as long as we preserve the ability to do all of the above.

@noughtmare

A large part of my job description is onboarding people onto Haskell, so I hope you understand that I don't consider that to be an acceptable solution. This is similar to what I was referring to in the motivation section of verbose and complicated type level programming to solve fundamentally value-level problems.

The `persistent` example doesn't seem to be have much in relation to `PersistStoreWrite` so it's not clear to me what it is attempting to do. Could you give an example that includes all the original methods of `PersistStoreWrite` thats shows what the library code would look like (it's a heavily used library so backwards compatibility would be strongly preferred) and what the user code would look like.

@xplat 

I'm not a reflection expert, so would you care to convert the examples I gave in the proposal to `reflection` to show me what that would like. My assumption is that they would be significantly more verbose and less beginner friendly. I also don't see anything in `reflection` that involves using the library to define new instances, so I was under the assumption that doing so wasn't possible.

mikeplus64:
> @mikeplus64
> 
> The reason I wanted to allow for user defined dictionary types is because sometimes the underlying dictionary will be a meaningful type of its own worthy of a proper name. For example `Traversal` for `Traversable`, `PersistStoreWriter` for `PersistStoreWrite`, raw `a` for `Default a`.
> 
> It's also worth noting that `Dict (Nice a)` would not have a `Contravariant` instance but `NiceDict a` would be able to support one.
> 
> With that said I'm not against looking into ways to allow people to access the dict of a fully exported class without requiring changes to the class, as long as we preserve the ability to do all of the above.

Fair enough. Being able to write standard instances against the concrete dict is a compelling argument. I guess with a magic module you could still have

```haskell
import GHC.ConcreteDict (Nice (..))
-- data Nice a = ...
instance Contravariant Nice where {...}
```

The other way I can imagine is an even more magical type/data family like `Dict :: (args ~> Constraint) -> args ~> Type` where `type family args ~> x where { '[] ~> r = r; (a ': as) ~> r = a -> (as ~> r) }`. Maybe there is an injective way to express that so it can be useful.

Alternatively just use `Compose Dict Nice` (and require infinity `ComposeN` wrappers as well.)

noughtmare:
> The persistent example doesn't seem to be have much in relation to `PersistStoreWrite` so it's not clear to me what it is attempting to do. Could you give an example that includes all the original methods of `PersistStoreWrite` thats shows what the library code would look like (it's a heavily used library so backwards compatibility would be strongly preferred) and what the user code would look like.

I don't know what your production code looks like. I mainly need to know how your `Environment` is defined. Without knowing that I cannot really give a more specific example.

The persistent example that I gave was based on the current implementation of the SQL backend in the persistent library. I understood that the main idea was that there are two classes: `PersistentStoreRead` and `PersistentStoreWrite`. This allows you to indicate a read only database in the types if you only require the `PersistentStoreRead` class. The persistent library provides three types: `SqlWriteBackend`, `SqlReadBackend` and the general `SqlBackend`. They implement the instances once for the general `SqlBackend` and then have 'redundant' instances that simply reuse the implementation of the `SqlBackend`.

My proposed approach is to use a more detailed underlying type that allows you to define just a single instance and then the rest of the types can be defined as type synonyms (or newtypes with derived instances using `GeneralizedNewtypeDeriving`).

Type classes are fundamentally a link between types and values. You keep repeating that you think this is really "fundamentally value-level code", but I disagree. It is possible to use richer types to achieve the same effect.

Looking back at my persistent example I realize it is quite silly. The same effect can be achieved with newtypes and `GeneralizedNewtypeDeriving`:

```Haskell
{-# LANGUAGE GeneralizedNewtypeDeriving #-}

data Environment = Environment Int

class Readable a where
  read :: a -> Int

class Writeable a where
  write :: Int -> a -> a

instance Readable (Environment a) where
  read (Environment n) = n

instance Writeable (Environment a) where
  write n (Environment _) = Environment n

newtype ReadEnvironment = ReadEnvironment Environment deriving Readable
newtype WriteEnvironment = Environment deriving Writeable
newtype RWEnvironment = Environment deriving (Readable, Writeable)
```

So, I guess my main point is that I cannot give more detailed examples without knowing more about your production code. Please give a more detailed, ideally self-contained, example.

cartazio:
Xplat makes a very good point , this seems entirely expressible via the reflection library today, or at least I can’t think of a nontrivial example to the contrary. (Though I welcome being wrong )

tysonzero:
@mikeplus64 

How would the `GHC.ConcreteDict` import know which `Nice` class is being referred too?

With proper support for anonymous extensible records I could see that being useful for classes without an explicit annotation.

As it stands I still lean towards requiring an explicit annotation for the names and values.

@noughtmare 

My `Environment` type is *not* a newtype over `SqlBackend`. It contains a variety of data including an `eBackend :: SqlBackend` field.

I don't want to be rude but I have to say I find your new example still quite frustrating. It does not mention `insert`, `insert_`, `insertMany` or any of the other `PersistStoreWrite` methods. I'm not trying to fundamentally restructure `persistent` and break everyones code. I just want to allow the library to define helper functions for creating instances.

Can you please show me how to create an instance of the existing `PersistStoreWrite` class with my `Environment` type that is significantly more concise than my example, without doing anything backwards-incompatible to `persistent`.

@cartazio 

Can one of you then please demonstrate how my examples can be made concise and beginner friendly with the `reflection` library.

goldfirere:
I'm sympathetic to this idea. Not yet convinced that this new feature is necessary, given the many ways that it can be mocked up today.

I agree that this is a feasible idea -- it doesn't break soundness/coherence or anything like that.

If there is a dictionary `CDict` for `C`, is there a function generated like `getDict :: C a => CDict a`? Or do you have to write that yourself with a bunch of boilerplate?

tysonzero:
@goldfirere 

The main reason I see the need for this feature is to handle the following fairly basic and simple desire in a beginner friendly and concise way:

* I have a class with a bunch of methods

* I want to expose various helper functions to concisely and easily define instances of that class

You essentially have three options as of now:

* You can define one single unparameterized "helper function" via default signatures. (e.g. `ToJSON`)

* You can also define a verbose and beginner unfriendly network of type level programming to handle more complex and parametrized cases. (e.g. `InjectedInto` + `Injective`)

* You can define one helper function for each member of the class. (e.g. `genericToEncoding ` + `genericToJSON`)

Currently interacting with classes and instances does not feel very friendly or concise outside the one happy path where you can immediately derive the exact instance you want.

> `getDict :: C a => CDict a`

This is mentioned in the unresolved questions section. It does seem like it would be worthwhile. I just couldn't come up with a nice syntax for it. This is part of why personally I will personally prefer to use the second syntax, but I anticipate the first syntax being more practical for a lot of existing libraries.

mpickering:
I think this is a great idea. Perhaps you could instead use a generic n-ary product rather than making a new datatype for each dictionary. There may be some issues with impredicativity but those may be solved by the recent work to improve that story.

It would also work really nicely with Typed Template Haskell.

simonpj:
I'm sympathetic to the idea, but not yet convinced that the power-to-weight ratio is enough to justify it.

> This has led to a proliferation of extensions that help you define instances and convert instances between one another.

Can you enumerate this proliferation, one by one, with examples to illustrate how each would be better handled with your scheme?

------------------------

Can you make your Change Specification be closer to a specification?  At the moment it gives an example or two but that's it.  A specification at least specify the changes in syntax pretty precisely, and should cover all cases.

One case I wonder about is this. Suppose I have
```
class C a where
   data dict CDict a = C
   op1 :: ...
   op2 :: ...
```
and I want to replace
```
instance C a => C [a] where
   op1 = ...
   op2 = ...
```
with
```
instance C a => C [a] = wubble
```
Now, what is the type of wubble?  Perhaps it must be
```
wubble :: C a => CDict [a]
```
It would be good to be specific.

-----------------------

The namespace-cluttering thing is indeed annoying.   Maybe we want an injective type family to convert from the classs to the data type, and vice versa?    But then what is the data constructor called?   I'm not sure how to resolve this.

----------------------
One dark corner: see `Note [Recursive superclasses]` in `GHC.Tc.TyCl.Instance`.  This describes a very subtle way in which we could get bottom-valued dictionaries, and how we avoid doing so.  But the avoidance strategy relies heavily on the fact that every instance decl immediately returns a data constructor.  Maybe we could come up with another strategy (although it's very tricky), but as it stands when you say (in your new proposal)
```
instance C T = wubble
```
we'd have to generate
```
$dfCT = MkC (case wubble of MkC a b c -> a)
                      (case wubble of MkC a b c -> b)
                      (case wubble of MkC a b c -> c)
```
Maybe that's ok, but worth pointing out the issue.

tysonzero:
> Can you enumerate this proliferation, one by one, with examples to illustrate how each would be better handled with your scheme?

[This blog post](http://h2.jaguarpaw.co.uk/posts/simplifying-typeclasses/) I wrote a while ago should more or less cover this.

> Now, what is the type of wubble? Perhaps it must be

> `wubble :: C a => CDict [a]`

Yeah that seems correct. Given:

```
instance ctx => C T = x
```

It types checks if and only if the following typechecks:

```
x :: ctx => Dict C T
```

> The namespace-cluttering thing is indeed annoying. Maybe we want an injective type family to convert from the class to the data type, and vice versa? But then what is the data constructor called? I'm not sure how to resolve this.

I am actually becoming increasingly convinced that the separate naming is a good thing.

The type and the typeclass have different semantic meaning, and with good naming conventions will generally not collide.

Types should generally be nouns (`Fold`, `Monoid`, `List`, `Number`, `Int`, `Ring`) and classes should generally be adjectives (`Foldable`, `IsList`, `Coercible`, `Numeric`, `Monoidal`?, `Ringed`?).

Even without this feature I would argue that in an ideal world the types equivalent to various classes should exist. As it stands it's not very convenient to actually programmatically do categorical things in Haskell, as currently you can generally only talk about one algebraic structure per carrier set.

This also allows you to define two separate classes over the same underlying type, for example `Additive` and `Multiplicative` might both have `Monoid` or `Semigroup` as their underlying dict.

> One dark corner: see `Note [Recursive superclasses]` in `GHC.Tc.TyCl.Instance`.

Hmm, if you are using the second syntax to define a class directly on top of a sum type this could be problematic. What is the underlying reason why bottom valued dictionaries are more problematic than single-member dictionaries that contain only bottom (`instance Default Void`)?

Lysxia:
To address the comments about reflection: the reflection package addresses a different issue through the creation of "local instances". It does not aim to unify and generalize deriving mechanisms currently in use, for defining normal, "global" instances, as this proposal does.

tysonzero:
Thanks @Lysxia for all the improvements!

@simonpj this should hopefully address your questions!

Lysxia:
The first comment is very relevant

> Related: https://dl.acm.org/doi/10.1145/3299711.3242752

Indeed, this proposal is similar to the one in that paper, [*Coherent Dictionary Applications in Haskell*](https://dl.acm.org/doi/10.1145/3299711.3242752), by Thomas Winant and Dominique Devriese, in Haskell Symposium 2018.

Here's a comparison between "this proposal" and "the paper". **The similarity** is that this proposal centers on two features, also found in the paper:

1. to expose type class dictionaries (Section 2.1)
2. to define instances using dictionaries ("dictionary instances", Section 2.3)

However, the paper focuses mainly on a third feature, "coherent dictionary applications", which is not part of this proposal.

**The differences** in the two relevant features are:

1. this proposal adds new syntax for naming the new record constructor; the paper derives the name from the class `MyClass.Dict`;
2. this proposal ignores superclasses because they are not needed for defining instances; the paper requires the full run-time information carried by class dictionaries, including superclasses, to be able to apply functions to them;
3. this proposal desugars dictionary instances in terms of the current syntax for instances; the paper uses the dictionary directly, which allows dictionaries
   to not terminate (it is unclear whether that is an issue, but it's one difference) and if the superclass fields are allowed to be overriden, that will be **incoherent**. (If that is indeed an oversight rather than my misunderstanding, I would chalk it up to "dictionary instances" being merely a secondary point of the paper.)

Items (2) and (3) point to there being a fundamental difference in need between "dictionary instances" and "dictionary applications", which justifies our choice of ignoring superclasses. It would in fact be detrimental to the former---this proposal---to expose the complete underlying representation of dictionaries including superclasses.

Leaving out superclasses will not lock us out of any future "dictionary application" proposal: one way in is to build a tuple containing our dictionaries, plus the superclass instances.

Lysxia:
@tysonzero The `data dict` syntax is a very hard sell. It means libraries must enable the extension for anyone to use this new feature. This extension will gain adoption much more easily if it does not require any change from existing libraries. I would much prefer to infer the constructor name, even if that involves new syntax for expressions/identifiers.

Ideas mentioned in previous comments for that are worth considering. The `MyClass.Dict` syntax from the Haskell Symposium paper mentioned earlier and addressed in my previous comment won't work, since it would be ambiguous with a constructor `Dict` in another module. To throw another idea on the pile, one variant is to replace `Dict` with a keyword, for example: `MyClass.instance`.

tysonzero:
It seems like you could use `MyClass` as the constructor name, and then (similar to DataKinds) use `’MyClass` as the type name.

If the module exports a constructor with the same name as the class then there would be still be ambiguity that would need to be resolved, but otherwise you’d be ok.

You could potentially add an optional extra `’` to resolve the above conflict. Giving you `’MyClass` as the constructor and `’’MyClass` as the type.

With the above in place I wouldn’t be overly concerned if we dropped the first syntax.

I would very strongly prefer to keep the second syntax though, as I generally develop in a type-first style (e.g `XMLParser`) and only add in a class later (e.g `FromXML`) when I decide I want canonicity. Particularly given how frustrating I have found it to a work with libraries that do the opposite.

It’s also worth noting that most of the benefit from this will come from libraries defining instances and helper functions on these dict types (such as my Persistent example) so they’d need this extension in that case anyway.

noughtmare:
I feel like a very lightweight alternative would be to simply allow pattern bindings in instance declarations. Combined with `RecordWildcards` it would allow many of the things proposed here, e.g.

```haskell
module EqDict where

data EqDict a = EqDict
  { (==) :: a -> a -> Bool
  , (/=) :: a -> a -> Bool
  }

defaultEq :: Eq a => EqDict a
defaultEq = EqDict
  { (EqDict.==) = \x y -> x Prelude./= y  -- using (Eq a)
  , (EqDict./=) = \x y -> x Prelude.== y
  }

instance Eq T where
  EqDict {..} = defaultEq { (EqDict.==) = eqT }
```

This requires some repetition with the `EqDict` data type and the annoying qualifiers, but I think the implementation would be very easy.

proposals/0000-concrete-class-dictionaries.md
```diff
@@ -0,0 +1,631 @@
...
+```haskell
+class Eq a where
+  data dict EqDict a = Eq
+  (==) :: a -> a -> Bool
+  (/=) :: a -> a -> Bool
+
+defaultEq :: Eq a => EqDict a
+defaultEq = Eq
+  { (==) = \x y -> x /= y  -- using (Eq a)
+  , (/=) = \x y -> x == y
```
noughtmare, line 140: Shouldn't this be `not (x /= y)` and `not (x == y)`?

Ericson2314:
I wrote up https://github.com/ghc-proposals/ghc-proposals/discussions/467 about how to do the "combinator soup" for quantified constraints explicitly when GHC cannot figure it out. I thought it would be like this but actually it turns out it is a completely orthogonal feature.

ashokkimmel:
Yes, but I'm wondering why patterns aren't allowed in instance declarations.

PR #131, opened 2018-05-02
Allow type variables to declared out of dependency order

goldfirere:
[Rendered](https://github.com/goldfirere/ghc-proposals/blob/reordered-tyvars/proposals/0000-reordered-tyvars.rst)

PR #326, opened 2020-04-23
WIP: Invisible params for declarations

Ericson2314:
[Rendered](https://github.com/obsidiansystems/ghc-proposals/blob/decl-invsible-param/proposals/0000-decl-invisible-params.rst)

### DIFF
--- a/proposals/0000-decl-invisible-params.rst
+++ b/proposals/0000-decl-invisible-params.rst
@@ -335,7 +335,7 @@ Implementation Plan
 
 I have begun this in `GHC MR 3145`_.
 I have some bugs but it has not been hard so far.
-@int-index's syntax work as provided a very good foundation.
+@int-index's syntax work has provided a very good foundation.
 
 Endorsements
 -------------
--- a/proposals/0000-decl-invisible-params.rst
+++ b/proposals/0000-decl-invisible-params.rst
@@ -151,18 +151,23 @@ These declaration forms also now allow parameters of the form ``@{var}``, where 
 Renaming
 ~~~~~~~~
 
-Any such declaration with an invisible parameter must only bind variables via explicit parameters in its head.
-There is no implicit binding of free variables in this case.
+``@``\ -prefixed parameters bind their variables just like normal ones do, in the same namespace and with the same scopes.
+\[The same rules on shadowing, duplication, and the mixing of implicitly and explicitly bound variables apply, as all follow from the choice of scope variables are bound in.\]
 
 Type checking
 ~~~~~~~~~~~~~
 
 An invisible parameter is given a invisible forall quantifier (``forall ... .`` kind).
+Invisible parameters need not all be given, the rule will be dual to type applications, and analogous to type lambdas:
 
-An invisible parameter in braces can be used only when the type being declared
-also has a standalone kind signature (SAK). The parameter name in braces must
-exactly match the name of a parameter bound with the ``forall {var}``
-construct in the type's SAK.
+ - An invisible parameter must match a corresponding invisible quantifier in the same position before, after, or between visible quantifiers.
+
+ - Invisible parameter in those positions must match a prefix of invisible quantifiers.
+
+Explicit and implicit visible parameters are both type-checked in the same way.
+
+An invisible parameter in braces can be used only when the type being declared also has a standalone kind signature (SAK).
+The parameter name in braces must exactly match the name of a parameter bound with the ``forall {var}`` construct in the type's SAK.
 
 Examples
 --------
@@ -176,13 +181,14 @@ Data type
 
 ::
 
-  type F :: forall k. k -> Type
-  data F @k :: k1 -> Type -- k1 is not bound
+  type F :: forall k -> k -> Type
+  data F @k :: k1 -> Type -- Rejected: doesn't match kind signature
 
 ::
 
-  type F :: forall k -> k -> Type
-  data F @k :: k1 -> Type -- dosen't match kind signature
+  type F :: forall k1 k2. forall (a :: k1) -> k -> Type
+  data F @kA @(a :: kB) :: k1 -> Type
+    -- Rejected: implicit binding for kB comes after kA, so kind is not matched.
 
 ::
 
@@ -315,7 +321,9 @@ None known at this time.
 Alternatives
 ------------
 
-Allow implicit variable binding always, or some in-between (such as allowing based off whether the first parameter is visible).
+Prohibit implicit bindings if explicit visible bindings are used.
+This was deemed to draconian.
+@Ericson2314 would argue that thanks to the simplifications to the arity rules, we arguable have extra "complexity budget" to "spend" on the interaction between implicit and explicit parameters.
 
 Unresolved Questions
 --------------------
--- a/proposals/0000-decl-invisible-params.rst
+++ b/proposals/0000-decl-invisible-params.rst
@@ -23,11 +23,11 @@ Scoping and parameter order in declaration heads
 
 When someone sees declarations like::
 
-  data Foo_0 a :: forall (b :: k). Type
+  data Foo_0 a :: Type -> forall (b :: k). Type
 
-  data family Foo_1 a :: forall (b :: k). Type
+  data family Foo_1 a :: Type -> forall (b :: k). Type
 
-  type family Foo_2 a :: forall (b :: k). Type
+  type family Foo_2 a :: Type -> forall (b :: k). Type
 
 They might think the signature works just like a term signature, and they can explicitly bind ``k`` without changing the meaning like::
 
@@ -41,7 +41,7 @@ But this is in fact not the same: the ``k`` parameter comes before the ``a`` in 
 At least, we can use standalone kind signatures to illustrate the difference if not fix the problem::
 
   type Foo_0, Foo_1, Foo_2 ::
-    forall k. Type -> forall (b :: k). Type
+    forall k. Type -> Type -> forall (b :: k). Type
 
 versus::
 
@@ -100,14 +100,14 @@ There is also a notion of arity: the number of arguments a type synonym family m
 We have the rough intuition that the number of parameters with patterns is the arity.
 The first problem is that implicitly bound variables in the kind, in floating to the LHS, also count::
 
-  -- arity 1
-  type Foo_0 :: forall (b :: k). Type
-  -- arity 2
-  type family Foo_1 a :: forall (b :: k). Type
+  -- arity [invisible]
+  type Foo_0 :: Type -> forall (b :: k). Type
+  -- arity [invisible, visible]
+  type family Foo_1 a :: Type -> forall (b :: k). Type
 
-  -- arity 0
+  -- arity []
   type Foo_0 :: forall k (b :: k) -> Type
-  -- arity 1
+  -- arity [visible]
   type family Foo_1 a :: forall k (b :: k) -> Type
 
 Now, since there is no body here, unlike the previous section we don't have a scoping problem or problem defining with ambiguous kinds.
--- a/proposals/0000-decl-invisible-params.rst
+++ b/proposals/0000-decl-invisible-params.rst
@@ -9,9 +9,7 @@ Invisible parameters for declarations
 .. implemented:: Leave blank. This will be filled in with the first GHC version which
                  implements the described feature.
 .. highlight:: haskell
-.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
-            **After creating the pull request, edit this file again, update the
-            number in the link, and delete this bold sentence.**
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/326>`_.
 .. contents::
 
 There are certain implicit bindings that cannot be written explicitly.
--- a/proposals/0000-decl-invisible-params.rst
+++ b/proposals/0000-decl-invisible-params.rst
@@ -23,11 +23,11 @@ Scoping and parameter order in declaration heads
 
 When someone sees declarations like::
 
-  data Foo_0 a :: Type -> forall (b :: k). Type
+  data Foo_0 a :: forall (b :: k). Type
 
-  data family Foo_1 a :: Type -> forall (b :: k). Type
+  data family Foo_1 a :: forall (b :: k). Type
 
-  type family Foo_2 a :: Type -> forall (b :: k). Type
+  type family Foo_2 a :: forall (b :: k). Type
 
 They might think the signature works just like a term signature, and they can explicitly bind ``k`` without changing the meaning like::
 
--- a/proposals/0000-decl-invisible-params.rst
+++ b/proposals/0000-decl-invisible-params.rst
@@ -1,5 +1,5 @@
 Invisible parameters for declarations
-==============
+=====================================
 
 .. author:: John Ericson (@Ericson2314)
 .. date-accepted:: Leave blank. This will be filled in when the proposal is accepted.
@@ -18,32 +18,122 @@ This adds the syntax to fix that not already proposed.
 Motivation
 ----------
 
-Implicit binding is confusing to users.
-Even advanced users.
-It is far from obvious that:
+Scoping and parameter order in declaration heads
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-::
+When someone sees declarations like::
 
-  data F :: forall (a :: k) -> Type
+  data Foo_0 a :: forall (b :: k). Type
 
-  data family F :: forall (a :: k) -> Type
+  data family Foo_1 a :: forall (b :: k). Type
 
-do not mean the same thing (respectively) as:
+  type family Foo_2 a :: forall (b :: k). Type
 
-::
+They might think the signature works just like a term signature, and they can explicitly bind ``k`` without changing the meaning like::
 
-  data F :: forall k. forall (a :: k) -> Type
+  data Bar_0 a :: forall k (b :: k). Type
 
-  data family F :: forall k. (a :: k) -> Type
+  data family Bar_1 a :: forall k (b :: k). Type
 
-For both the ``data`` and ``data family`` declaration, the order in which the ``k`` parameter is quantified changes, but perhaps that isn't so bad on its own.
-Worse is that for the family definition, the implicitly bound ``k`` is something instances can scrutinize, while the other ``k`` is purely parametric.
+  type family Bar_2 a :: forall k (b :: k). Type
 
-Making matter worse still, the latter forms of each cannot be written with all variables explicitly bound.
-How are we suppose to teach these subtle things made even by no explicit way to highlight their differences!
+But this is in fact not the same: the ``k`` parameter comes before the ``a`` in the first group, but after the ``a`` and before the ``b`` in the second group.
+At least, we can use standalone kind signatures to illustrate the difference if not fix the problem::
 
-The solution is ``@``-prefixed "invisible parameter", as they are known, just as is already proposed for constructor patterns, in `Proposal 126`_, and lambdas, in proposal `Proposal 155`_.
-The uniformity between all of these should round out the language according to expectations.
+  type Foo_0, Foo_1, Foo_2 ::
+    forall k. Type -> forall (b :: k). Type
+
+versus::
+
+  type Bar_0, Bar_1, Bar_2 ::
+    Type -> forall k (b :: k). Type
+
+(pretending for concision's sake we had multiple name SAKS syntax).
+
+Scoping in declaration bodies
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+These all work (with SAKSs for just clarity---they can be removed)::
+
+  type Foo :: Type -> Type
+  data Foo k = forall (b :: k). MkFoo
+
+  type Foo :: forall k. k -> Type
+  data Foo (a :: k) = forall (b :: k). MkFoo
+
+  type Foo :: forall k. k -> Type
+  type Foo (a :: k) = k
+
+  type C :: Type -> Constraint
+  class C k where
+    type F :: k
+
+  type C :: forall k. k -> Constraint
+  class C (a :: k) where
+    type F :: k
+
+But these don't::
+
+  type Foo :: forall k. Type
+  data Foo = forall (b :: k). MkFoo -- k is unbound
+
+  type Foo :: forall k. Type -> Type
+  data Foo a = forall (b :: k). MkFoo -- k is unbound
+
+  type Foo :: forall k. k -> Type
+  type a = k
+
+  type C :: forall k. k -> Constraint
+  class C a where
+    type F :: k
+
+The problem is while free variables in the LHS / head are implicit bound, free variables in the RHS / body are not.
+The ``forall k`` in the SAKSs corresponds the invisible parameter we'd like to use.
+In the ambiguous cases like the first failing example, where no visible parameter references the invisible parameter, there is in fact no way to make the experiment type-check today.
+
+Arity
+~~~~~
+
+For type synonyms and type families, the kind doesn't yet tell the whole story.
+There is also a notion of arity: the number of arguments a type synonym family must be applied if the use of it is to be allowed.
+
+We have the rough intuition that the number of parameters with patterns is the arity.
+The first problem is that implicitly bound variables in the kind, in floating to the LHS, also count::
+
+  -- arity 1
+  type Foo_0 :: forall (b :: k). Type
+  -- arity 2
+  type family Foo_1 a :: forall (b :: k). Type
+
+  -- arity 0
+  type Foo_0 :: forall k (b :: k) -> Type
+  -- arity 1
+  type family Foo_1 a :: forall k (b :: k) -> Type
+
+Now, since there is no body here, unlike the previous section we don't have a scoping problem or problem defining with ambiguous kinds.
+We can always just write down the any quantifiers we want in the kind, and then control should they count towards the arity or not.
+The problem is, because there is no pattern syntax for invisible binders, GHC uses the legacy explicit return kind to determine the *non*\ -arity part of the kind signature from the legacy inline kind signature, and *just* for invisible binders!.
+This is convoluted to both learn and implement, and requires the user to write more than they should::
+
+  type F :: forall k. Maybe k
+  type family F
+            -- arity = [invisible]
+
+  type G :: Type -> forall k. Maybe k
+  type family G
+            -- arity = []
+
+  type H :: Type -> forall k. Maybe k
+  type family H a
+            -- arity = [visible, invisible]
+
+  type H :: Type -> forall k. Maybe k
+  type family H a :: forall k. Maybe k
+            -- arity = [visible]
+
+The solution for all of these problems ``@``\ -prefixed patterns for invisible parameters, just as is already proposed for constructor patterns, in `Proposal 126`_, and lambdas, in proposal `Proposal 155`_.
+We can use the same simple argument order, scoping, and arity rules for both visible and invisible parameters, and complete the obseletion of inline kind signatures and CUKSs.
+Finally, this proposal completes what the other two proposals started, bringing a much-needed uniformity to the language, and matching users expectations after they first encounter ``@``\ -bindings in one context and try to use in another.
 
 Proposed Change Specification
 -----------------------------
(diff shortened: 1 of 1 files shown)


int-index:
I have two additional bits of motivation for this change.

1. Specifying the arity of type families and type synonyms. With SAKS, this is done in a roundabout way by comparing the amount of `forall`-bound variables in the standalone kind signature and the inline kind signature.

   ```haskell
   type F :: forall k. Maybe k
   type family F
             -- arity = 1

   type G :: forall k. Maybe k
   type family G :: forall k. Maybe k
             -- arity = 0
   ```

    Firstly, the kind signature is duplicated. Secondly, it is an extremely arcane bit of knowledge that an inline kind signature can have an effect on the arity.

    With `@`-binders, we could specify the arity using the amount of binders:

   ```haskell
   type F :: forall k. Maybe k
   type family F
             -- arity = 0

   type G :: forall k. Maybe k
   type family G @k
             -- arity = 1
   ```

2. Scoping of class variables. Consider this declaration:

    ```
    class C (a :: k -> k -> k) where
      type F :: k
    ```

    One could (mistakenly) think that the equivalent declaration with a SAKS would look like this:

   ```
   type C :: (k -> k -> k) -> Constraint
   class C a where
     type F :: k
   ```

   But in fact, these declarations differ, as `k` in `type F :: k` is no longer class-bound! And this results in an error:

   ```
     • The associated type ‘F k’
       mentions none of the type or kind variables of the class ‘C k a’
     • In the class declaration for ‘C’
   ```

   Today there's one way to mitigate this, which is to duplicate the kind signature on `a`:

   ```
   type C :: (k -> k -> k) -> Constraint
   class C (a :: k -> k -> k) where
     type F :: k
   ```

    Another unfortunate case of duplicated kinds. With `@`-binders, there's a much better way to bind `k`:

   ```
   type C :: (k -> k -> k) -> Constraint
   class C @k a where
     type F :: k
   ```

   I faced this issue when working on deprecating CUSKs in favor of SAKSs. GHC's test suite revealed the need for `@`-binders.

So I'm hugely in favor of this proposal. This is the best way forward if we want to get rid of CUSKs. Solves two corner-cases which are poorly handled by SAKSs today.

goldfirere:
Thanks @Ericson2314 for writing this up! I've been meaning to do this for ages!

* I agree with the high-level points in the motivation. But I'm sorry to say that the examples you chose don't exhibit the problem. Because there are no arguments before the `::`, the kinds of the declarations both with and without the `forall` are the same. And, because data families don't really have an arity, they can scrutinize all of their arguments, whether declared before or after the `::`. It's only type families where it matters.

* In your examples, it would be helpful to be very explicit about which you mean to accept and which you mean to reject. Also, please give each a unique name for easy discussion. For example, the `k1 is not bound`: is that a good thing or a bad thing? I would think that `k1` would be brought implicitly into scope, and then the kind wouldn't match the SAK. But maybe you want to disable implicit scoping somehow? It's not clear to me.

* Your last two `class` examples use `->`. Did you mean `=>`?

* I'm uncomfortable using the presence of an implicit parameter to enable/disable automatic scoping of variables. Instead, how about a `-XNoImplicitQuantification` that subsumes `-XNoImplicitForAll` and covers this case too?

* While it's clear SPJ wants *some* proposal of this form, I personally would refrain from including a note saying he has endorsed this specific proposal.

* What about inferred variables? This proposal deals only with specified ones. Perhaps we don't want to allow this syntax with inferred variables, but the omission should be noted in the proposal.

* All your examples have SAKs. Is that a requirement? If not, we are setting up a tension with the term level. For example:

```hs
id @a (x :: a) = x    -- no type signature; rejected under #155
type Id @a (x :: a) = x    -- accepted under this proposal
```

So I think a little refining around the edges would help this proposal, but I'm very strongly in support of this general idea.

Ericson2314:
First of all, marking this WIP, which it should have been all long. I really need to bring out the arity parts, which are every bit (if not more) as important as the scoping. That the examples are all scoping focused is a symptom of arity being swept under the rug.

> While it's clear SPJ wants some proposal of this form, I personally would refrain from including a note saying he has endorsed this specific proposal.

That is how I intended to use the quote---about the vague idea being "natural" in everyone's mind and not the specific proposal. So if that was unclear I will indeed remove.

Ericson2314:
```
Prelude> data family F :: forall k. k
Prelude> data instance F :: * -> *
Prelude> data instance F :: * -> * -> *
```
OK this works with GHC 8.10, which is very surprising to me. @int-index I like your arity proposal a lot and will adopt it, but I don't think I understand the status quo very well with which I need to compare it with!

int-index:
The concept of arity is meaningful when talking about type synonyms and type families, as those cannot be partially applied.

For data types, newtypes, classes, and data families, we don't need to talk about arity at all. And as the last example shows, it's hard to even say what arity a data family has, as it varies per instance.

Ericson2314:
> And, because data families don't really have an arity, they can scrutinize all of their arguments, whether declared before or after the `::`. It's only type families where it matters.

Oh, there is the answer. How odd.

Ericson2314:
@int-index 

```haskell
type C :: (k -> k -> k) -> Constraint
class C a where
  type F :: k
```

Is this something we even want to work? It wasn't expressible before was it? I liked how in the type family case all the scrutable variables come first. But with classes there is no ordering. Even crazier:

```haskell
type C
  :: forall k
  .  (Proxy k)
  -> forall (a :: k)
  .  (Proxy k)
  -> forall (b :: k)
  .  Constraint
class C a b

instance C @Bool 'Proxy @False 'Proxy @False
instance C @Bool 'Proxy @False 'Proxy @True
```

The bindings are interspersed, and I can use them to make those instances not overlap.

I suppose semantics are "indexed class families", but is this intentional?!?

simonpj:
I am indeed in favour of a proposal along these lines!

We need to think about specificity.   Suppose we have a SAKS
```
data D :: forall {j}. forall k. j -> k -> Type
```
The braces mean that the `j` parameter is always inferred, and you can't use visible kind application.  Thus you can say  `D Int Maybe` or `D @(Type->Type) Int Maybe`.  But you can't fix j in with a visible kind application.

Now, in D's declaration what do we say.  The natural thing would be
```
data D @{j} @k a b = MkD (...j..k...a...b...)
```
Those @ parameters are optional, but useful to bring j and k into scope.   The question is: do we need to be explicit about the braces in the data declaration.  I think we could say
* Yes, be explicit. And without a SAKS you'd infer something with the desired specificity
* No, don't be explicit (no braces in the data decl).  Just get specificity from the SAKS; if there is no SAKS, make them all Specified.

I'm not sure which is better.

goldfirere:
*Inferred* bits are problematic. The key quality about inferred variables is that their order does not matter. That's the whole point of *inferred*! So, for any type, writing `forall {a} {b} . ...` should be indistinguishable from writing `forall {b} {a} . ...`. Yet that won't be the case if we can bind inferred variables with `@{a}`, because suddenly the order matters.

Here's an easier way to see this problem:

```
type T :: forall a b. Proxy a -> Proxy b -> Type
data T @{k1} @{k2} @(a :: k1) @(b :: k2) p1 p2
```

Is this correct? Or have I gotten my `k1` and `k2` confused? If this syntax were allowed, the answer would depend on the order GHC chooses for `k1` and `k2` when inferring their quantification. We could say that the `@{k1}` syntax is allowed only when the corresponding kind as a `forall {k1}` in it... but that means inventing a *third* specificity: *explicitly inferred*, where we can bind the variable with `@{...}` but never instantiate it visibly. I don't want this.

So I'm stuck on these inferred variables.

#131 is looking more attractive, as it would effectively be a replacement for *inferred* variables. Or requiring named instantiation.

@Gertjan423 and I thought about all this a bit over the past few months, and he may be interested here.

int-index:
> Is this something we even want to work? It wasn't expressible before was it?

Wasn't expressible before what? As I said, the only problem is that `k` from the SAKS does not scope over the `where` clause. We can make it scope by repeating the kind signature:

```
type C :: (k -> k -> k) -> Constraint
class C (a :: k -> k -> k) where
  type F :: k
```

This works in GHC 8.10

With `@`, we would avoid duplication:

```
type C :: (k -> k -> k) -> Constraint
class C @k a where
  type F :: k
```

As to your crazier example:

```
type C
  :: forall k
  .  (Proxy k)
  -> forall (a :: k)
  .  (Proxy k)
  -> forall (b :: k)
  .  Constraint
class C a b

instance C @Bool 'Proxy @False 'Proxy @False
instance C @Bool 'Proxy @False 'Proxy @True
```

I see no problem with this. Invisible kind parameters are on equal grounds with other parameters. Today you could write the same thing, just a bit differently:

```
instance C ('Proxy :: Proxy False) ('Proxy :: Proxy True)
instance C ('Proxy :: Proxy True) ('Proxy :: Proxy False)
```

goldfirere:
Tiny nitpick: the pull request link in the proposal should be updated to point to this discussion.

Ericson2314:
@int-index Sorry ignore all that. I thought there was another arity-like scrutinized vs not scrutinized variable situation going on, but it is just scope.

goldfirere:
@simonpj and I had a chat about this proposal, where we discussed my concerns https://github.com/ghc-proposals/ghc-proposals/pull/326#issuecomment-618909674 about inferred binders. I have formatted my thoughts as a diff against this PR and posted
https://github.com/obsidiansystems/ghc-proposals/pull/4

int-index:
As additional motivation for this proposal, there are users requesting this feature: https://gitlab.haskell.org/ghc/ghc/issues/17705

proposals/0000-decl-invisible-params.rst
```diff
@@ -18,61 +18,156 @@ This adds the syntax to fix that not already proposed.
...
+~~~~~
+
+For type synonyms and type families, the kind doesn't yet tell the whole story.
+There is also a notion of arity: the number of arguments a type synonym family must be applied if the use of it is to be allowed.
+
+We have the rough intuition that the number of parameters with patterns is the arity.
+The first problem is that implicitly bound variables in the kind, in floating to the LHS, also count::
+
+  -- arity 1
+  type Foo_0 :: forall (b :: k). Type
```
int-index, line 104: Doesn't this fail the forall-or-nothing check?

Ericson2314: Indeed!

int-index: There's a way to evade the check at the cost of making the example more contrived:

```
type Foo_0 :: forall. forall (b :: k). Type
```

Ericson2314: Actually that's still an explicit forall at the beginning, so I think it would still be disallowed. I threw in a leading `Type ->` for now.

int-index: Actually, that did not work. But here's something that does:

```
type Foo_0 :: Type -> forall (b :: k). Type
data Foo_0 a
```

RyanGlScott:
One thing that isn't mentioned in this proposal (at least, not that I can tell from a quick read) is that it would let us simplify the specification in [Proposal 103](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0103-no-kind-vars.rst) (Treat kind variables and type variables identically) further. That proposal aims to unify the treatment of type and kind variables, but there is actually an exception to this treatment. Namely, kind variables are handled differently than type variables in the context of type synonyms and type family instances:

```hs
type N1 = Nothing :: Maybe a

type family N2 :: Maybe a
type instance N2 = Nothing :: Maybe a
```

Note that the right-hand sides of `N1` and the `N2` instance mention a variable `a` that is not explicitly bound on the left-hand side. Normally, this would be an error, but Proposal 103 treats these specially since (1) `a` is a kind variable, and (2) it is mentioned in an outermost kind signature on the right-hand side. As a result, `a` is implicitly quantified over, and both `N1` and `N2` are accepted.

When Proposal 103 was being implemented, we considered what it would take to remove these special cases. Removing the special case for `N2` would actually be quite straightforward, since type family instances already allow you to bind invisible arguments:

```hs
type instance N2 @a = Nothing :: Maybe a
```

On the other hand, no such syntax existed for type synonyms, so we couldn't easily remove the special casing for them. Having a special case for type synonyms but not type family instances felt incongruous, so we decided to just keep outermost-kind-signature exception for both types of declarations.

With this proposal, however, we can revisit this design, since we can now write:

```hs
type N1 @a = Nothing :: Maybe a
```

I'm not sure if this proposal itself is the best place to mention this simplification or if it would be better to amend Proposal 103 instead, but I thought it was worth mentioning this nonetheless.

int-index:
Good point @RyanGlScott! This is also documented in `Note [Implicit quantification in type synonyms]` in the `GHC.Rename.HsType` module, which I will cite here for completeness:

```
Note [Implicit quantification in type synonyms]
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
We typically bind type/kind variables implicitly when they are in a kind
annotation on the LHS, for example:

  data Proxy (a :: k) = Proxy
  type KindOf (a :: k) = k

Here 'k' is in the kind annotation of a type variable binding, KindedTyVar, and
we want to implicitly quantify over it.  This is easy: just extract all free
variables from the kind signature. That's what we do in extract_hs_tv_bndrs_kvs

By contrast, on the RHS we can't simply collect *all* free variables. Which of
the following are allowed?

  type TySyn1 = a :: Type
  type TySyn2 = 'Nothing :: Maybe a
  type TySyn3 = 'Just ('Nothing :: Maybe a)
  type TySyn4 = 'Left a :: Either Type a

After some design deliberations (see non-taken alternatives below), the answer
is to reject TySyn1 and TySyn3, but allow TySyn2 and TySyn4, at least for now.
We implicitly quantify over free variables of the outermost kind signature, if
one exists:

  * In TySyn1, the outermost kind signature is (:: Type), and it does not have
    any free variables.
  * In TySyn2, the outermost kind signature is (:: Maybe a), it contains a
    free variable 'a', which we implicitly quantify over.
  * In TySyn3, there is no outermost kind signature. The (:: Maybe a) signature
    is hidden inside 'Just.
  * In TySyn4, the outermost kind signature is (:: Either Type a), it contains
    a free variable 'a', which we implicitly quantify over. That is why we can
    also use it to the left of the double colon: 'Left a

The logic resides in extractHsTyRdrTyVarsKindVars. We use it both for type
synonyms and type family instances.

This is something of a stopgap solution until we can explicitly bind invisible
type/kind variables:

  type TySyn3 :: forall a. Maybe a
  type TySyn3 @a = 'Just ('Nothing :: Maybe a)

Note [Implicit quantification in type synonyms: non-taken alternatives]
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Alternative I: No quantification
--------------------------------
We could offer no implicit quantification on the RHS, accepting none of the
TySyn<N> examples. The user would have to bind the variables explicitly:

  type TySyn1 a = a :: Type
  type TySyn2 a = 'Nothing :: Maybe a
  type TySyn3 a = 'Just ('Nothing :: Maybe a)
  type TySyn4 a = 'Left a :: Either Type a

However, this would mean that one would have to specify 'a' at call sites every
time, which could be undesired.

Alternative II: Indiscriminate quantification
---------------------------------------------
We could implicitly quantify over all free variables on the RHS just like we do
on the LHS. Then we would infer the following kinds:

  TySyn1 :: forall {a}. Type
  TySyn2 :: forall {a}. Maybe a
  TySyn3 :: forall {a}. Maybe (Maybe a)
  TySyn4 :: forall {a}. Either Type a

This would work fine for TySyn<2,3,4>, but TySyn1 is clearly bogus: the variable
is free-floating, not fixed by anything.

Alternative III: reportFloatingKvs
----------------------------------
We could augment Alternative II by hunting down free-floating variables during
type checking. While viable, this would mean we'd end up accepting this:

  data Prox k (a :: k)
  type T = Prox k
```

Ericson2314:
@RyanGlScott I originally made this proposal very strict in prohibiting implicitly quantified variables when `@`-binders were used, but I was convinced to let things for now. While it is no good to distinguish type for kind variables, I think it is OK-ish to distinguish variables in signatures from from variables that aren't, as we do that on the type/term level too.

```haskell
type instance N2 @a = Nothing :: Maybe a
```

I should point out that neither of those are proper binders, actually, because we have
```haskell
type instance forall a. N2 @a = Nothing :: Maybe a
```
now.

int-index:
> I should point out that neither of those are proper binders, actually, because we have
> `type instance forall a. N2 @a = Nothing :: Maybe a`

In type families, yes. In type synonyms, no.

For type synonyms, you do not have that extra `forall`, and the LHS has actual binders. So I'd love if this proposal removed the special case for handling the RHS of type synonyms.

Ericson2314:
Agreed. I meant the two `a`s in the `type instance` example, not the two examples; I should have been more clear on that..

cgibbard:
I'm confused about a few of the examples

    type F6 :: forall {k1} {k2} a b. Proxy a -> Proxy b -> Type
    data F6 @{k2} @{k1} @a @b p1 p2

is rejected because the variables are in the wrong order, but I sort of think it's weird that we're allowing explicit binding of curly-braced variables here in the first place. They're fine in the SAKS, but I think they should perhaps be disallowed altogether in the data declaration. I'm not sure about that though, and could possibly be convinced of the importance of allowing that by an example. The left hand side of a data declaration is meant to mimic a usage site in a way that makes this uncomfortable.

Another confusing thing about these examples: are k1 and k2 meant to be new unspecified type arguments, or are they meant to capture the unspecified type arguments which might already be present from the kinds of a and b?

goldfirere:
Here is an example of the `@{k}` notation:

```hs
type Proxy :: forall {k}. k -> Type
data Proxy @{k} (a :: k) = Proxy
```

This keeps the `k` as inferred while still allowing me to use it. The principle here is that I might want to bring a `k` into scope for the definition of a datatype. This desire should not require that I change the *interface* to the type. Without `@{k}` notation, there would be no way to bring `k` into scope (other than implicit binding, which @Ericson2314 is rightly fighting against).

Icelandjack:
I missed this conversation and made a ticket (https://gitlab.haskell.org/ghc/ghc/issues/18139) about it which discusses data and type constructors, classes and class methods, type families and synonyms.

Ericson2314:
So like @int-index I always felt iffy on `{..}` in general. But I'll try to not wade into those weeds for now.

-----

From the amended proposal

> The parameter name in braces must exactly match the name of a parameter bound with the forall {var} construct in the type's SAK.

I guess I am confused why the names need to match for the `{..}` bindings. Since only explicit ones in the SAKSs in play, how come it can't be positional like everything else? Just to be cautious future-proof?

In general, having names in signatures have anything to do with names in bodies feels like a legacy of `-XScopedTypeVariables` I think I want to dispense with.

------

> ```haskell
> forall {k}.
> ```

Hmm reminds me of a old idea, what if we allowed `@` `forall @a b @c d ->`? We could even deprecate the `.` quantifier syntaxes!

goldfirere:
To understand why we want the names to line up, consider

```hs
type T1 :: forall a b. Proxy a -> Proxy b -> Type
type T2 :: forall {k1} {k2} (a :: k1) (b :: k2). Proxy a -> Proxy b -> Type
```

After kind-checking and generalization, these signatures have the exact same meaning -- except that we cannot know what order the kind variables are quantified in `T1`. If we allowed binding them with `data T1 @{k1} @{k2} ...`, we wouldn't know which variable goes where. You could say: well don't! That is, disallow binding the variables in `T1` but allow it in `T2`. The problem there is that it requires essentially having a *fourth* category of quantified dependent variable: "inferred but user-written". These inferred-but-user-written variables (as in `T2`) would be allowed to be bound, but the inferred-but-really-inferred variables (as in `T1`) would not. I desperately don't wish to do this. The name-lining-up seemed an easy-enough compromise.

Ericson2314:
@goldfirere I'm a bit confused by your answer. 

> If we allowed binding them with data T1 @{k1} @{k2} ..., we wouldn't know which variable goes where.

Completely agree.

> I desperately don't wish to do this. The name-lining-up seemed an easy-enough compromise.

I don't get what the names not lining up helps us avoid?

Whats less bad about `forall {k}.` is yes `{k}` is a fourth type of binding syntax, but it's not a 4th type semantically---(non-core) quantifiers still fall into specified vs inferred vs required. It's the binding associated with `T2` just checks whether 1) there's a SAKS, 2) does the SAKS have enough `{...}` bindings. Users of `T2` still have no idea whether an inferred variable has an explicit lambda or not.

I don't get how the name alignment thing helps simplify the above. It just seems like one extra requirement on top of everything else.

goldfirere:
> 2) does the SAKS have enough {...} bindings.

How is this implemented? Maybe you're suggesting that the kind signature induces both a kind and an inferred-arity. When we check the definition, we must ensure that the inferred-arity is at least as many `@{...}` binders as appear. Because the definition is always in the same module as the signature, I guess that's possible. But it does mean that the specification of the behavior must have an extra little tidbit that says that the number of `@{...}` binders in the definition must be no more than the number of `{...}` `forall`-bound variables in the kind signature. It's not really so different than an extra side condition about name-matching. But maybe numbers are easier than names.

Ericson2314:
Yes that sounds right to me. Either way we're smuggling a "pst! the signature has hidden <blah> so you, the definition, alone can do <blah>". Just the minor details of the steganography are different.

goldfirere:
OK. I relent on my name-matching idea. We can just go with numbers, as you would doubtless prefer. :)

goldfirere:
Is this ready for the committee to review?

int-index:
One thing I'm unsure of, in this proposal, is the treatment of inferred type variables. Introducing `@x`-binders is obviously the right thing to do, because currently we have explicit type applications, but not explicit type abstraction, and we're looking to rectify this deficiency.

However, when we introduce `@{x}`-binders, we are introducing an asymmetry instead of removing it. Now inferred variables will have binders, but not applications!

The reason inferred variables are not subject to type applications is that we don't want their order to matter. That is, consider the following type signatures:

```
f :: forall {k1} {k2} (a :: k1) (b :: k2). Proxy a -> Proxy b
g :: forall {k2} {k1} (a :: k1) (b :: k2). Proxy a -> Proxy b
h :: forall a b. Proxy a -> Proxy b
```

`f`, `g`, and `h` are term-level here for simplicity, but the same reasoning applies to SAKS and type declarations, and I wager that if we introduce `@{x}`-binders in types, we'll have to amend [#155](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0155-type-lambda.rst) accordingly.

So, my understanding is that we want `f`, `g`, and `h`, to be indistinguishable. That's because in case of `h`, we can't possibly know in which order `k1` and `k2` will be implicitly quantified, and exposing it to the user would leak implementation details.

But `@{k}` stand in the way somewhat. We are now talking about different treatment for `f`, `g`, and `h`:

```
f :: forall {k1} {k2} (a :: k1) (b :: k2). Proxy a -> Proxy b
f @{k1} @{k2} Proxy = Proxy 

g :: forall {k2} {k1} (a :: k1) (b :: k2). Proxy a -> Proxy b
g @{k1} @{k2} Proxy = Proxy 

h :: forall a b. Proxy a -> Proxy b
h @{k1} @{k2} Proxy = Proxy 
```

Do we allow this?

* In case of `h`, we outright reject, because it's unpredictable what variables `k1` and `k2` are. Maybe we're getting `a :: k1, b :: k2`, or maybe we're getting `a :: k2, b :: k1`. That's a non-starter.

* In case of `f` and `g`, we have user-written binders, so the situation is more nuanced. There have been two matching strategies put forward: by name and by ordering. Notably, these strategies lead to different results.

Both proposed strategies introduce more spooky action at a distance (i.e. information that travels from the type signature to the declaration). We already have other sources of action at a distance (ScopedTypeVariables and type family arity), and they're bad enough, and I even think we're trying to get rid of them. So let's not introduce more side channels for moving information from the signature to the definition.

Can we just remove them from the proposal? I think yes. Let's consider the examples from the proposal:

```
type D4 :: forall {k1} {k2} a b. Proxy a -> Proxy b -> Type
data D4 @{k1} @{k2} @a @b p1 p2   -- OK

type D5 :: forall {k1} {k2} a b. Proxy a -> Proxy b -> Type
data D5 @{k1} @{k2} @p @q x y     -- OK: specified type variables can be renamed
```

But there's an easy way to rewrite those:

```
type D4 :: forall {k1} {k2} a b. Proxy a -> Proxy b -> Type
data D4 @(a :: k1) @(b :: k2) p1 p2

type D5 :: forall {k1} {k2} a b. Proxy a -> Proxy b -> Type
data D5 @(p :: k1) @(q :: k2) x y
```

Just bind those variables using inline kind signatures. This rewrite should be always possible for any inferred variable, because inferred variables cannot be ambiguous, and therefore always appear in types/kinds of other binders.

I can think of two objections to this: 

1. Sometimes we'll need to duplicate the type of a binder. For example:
    ```
    f :: forall {a}. VeryLargeType a -> ()
    f (x :: VeryLargeType a) = ...
    ```

    I say: yes, but I don't expect the need for binding inferred variables to crop up very often (in fact, seems like a very niche use case). That's still much better than the complexity that `@{x}`-binders add to the language.

2. We lose the guarantee that those variables will be skolems, as they may unify with other types. I say: yes, but it's not a terribly important guarantee. However, if we deem it to be important, we can introduce `@{x}`-binders without spooky action at a distance, as "skolemicity" markers, so to say. For example, we would allow something like this:

    ```
    f @{k} (x :: Proxy (a :: k)) = ...
    ```

    We don't even need the signature of `f` to check this. First, we pretend that the `@{k}`-binder isn't even there, effectively checking this declaration:

    ```
    f @{k} (x :: Proxy (a :: k)) = ...
    ```

    And then we introduce a check that `k` must have unified with a variable, not with a type. We even have a mechanism for this in GHC: use `TyVarTv` instead of `TauTv` for the unification variable.

-------

To summarize, here's what I propose:

* Drop `@{x}`-binders from this proposal, they don't buy us much.
* If we later discover that we actually want them to enforce skolemicity, introduce them non-intrusively (without action at a distance), as skolemicity markers.

Thoughts?

Ericson2314:
> Just bind those variables using inline kind signatures.

I am a bit nervous because `-XNoPatternSignatureBinds` would take that away. But overall, yes, I am happy to just punt on this inferred binder stuff, as we can always come back to it later compatably.

int-index:
> I am a bit nervous because `-XNoPatternSignatureBinds` would take that away.

Right, with `-XNoPatternSignatureBinds` we'd need `@{x}`-binders, but they still could work as I described (as skolemicity markers). Essentially the trick is to introduce it as a metavariable and then check that it unified with some other skolem.

RyanGlScott:
It might be worth mentioning [GHC#15782](https://gitlab.haskell.org/ghc/ghc/-/issues/15782) and [GHC#17705](https://gitlab.haskell.org/ghc/ghc/-/issues/17705) somewhere in this proposal, as this proposal would fix those two issues.

simonpj:
@Ericson2314, what's your plan for this proposal?  I really think we should have it, and indeed I have written #386 which builds on it.

But this one is marked WIP... do you plan to finish it and submit?  Thanks!

Ericson2314:
@simonpj Well, we recently rebased the WIP implementation, now that many things @RyanGlScott was doing in the same area came to pass and the time seems ripe.

I'm not sure whether I'll be able to work on this (proposal or MR) as part of work. I'm fine with any of 

1. waiting a bit longer to confirm
2. preemptively handing off the proposal and/or MR completely
3. preemptively setting the proposal and/or MR up so many people can contribute.

Your call.

simonpj:
Thanks. 

Does anyone feel able to take over the proposal and/or MR?  Being able to say
```
data T @k (a::k) = MkT (Proxy @k a)
```
seems so obviously desirable!  It fills out an important corner of our design space, in a nice uniform way.   We just need to agree the proposal and get it done. @RyanGlScott you are very active in this space...would you have time?  Or anyone else?

RyanGlScott:
> you are very active in this space...would you have time?

I don't. Sorry.

simonpj:
@Ericson2314 or indeed anyone else (since John says it's not really in-scope for him): would anyone feel able to take over authorship and push this proposal through to acceptance?  There is no doubt (in my mind) that we want this proposal: it fills an obvious hole in our story.

int-index:
Very well, I will see this to completion.

Ericson2314:
Thanks, @int-index! Sorry again to be abandoning this one.

simonpj:
> Very well, I will see this to completion.

@int-index are you really taking this over? That's great!

Can you incorporate #386 as part of it?   The text is all written!

If you want any help from me, let me know.

int-index:
> are you really taking this over?

Yes. I’m going to write a proposal that has the same effect as this one, but in my own words and with minor adjustments.

> Can you incorporate #386 as part of it?

Alright, it looks reasonable.

Ericson2314:
I'll just close this ahead of you opening the new one

PR #334, opened 2020-05-22
Add draft proposal for extending array primitives

vilem-ff:
We propose to expand the arrays API in [GHC.Exts](https://hackage.haskell.org/package/base-4.14.0.0/docs/GHC-Exts.html) by picking the low hanging fruit of missing primitives which would improve the performance of key libraries, such as [unordered-containers](https://github.com/tibbe/unordered-containers/tree/d0b78afc028ef307a3cdfe1dd95feaf4ec47f195).

[Current RFC rendered here.](https://github.com/buggymcbugfix/ghc-proposals/blob/master/proposals/0000-extend-array-primitives.md)

### DIFF
--- a/proposals/0000-extend-array-primitives.md
+++ b/proposals/0000-extend-array-primitives.md
@@ -1,0 +1,80 @@
+---
+author: buggymcbugfix
+date-accepted: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/335).
+
+# RFC: Expand arrays APIs with efficient primops
+
+## Motivation
+
+Operations that allocate new arrays (`Array#`, `MutableArray#`, `SmallArray#`,
+`SmallMutableArray#`) must ensure that the array never contains garbage values
+during safe points (when the GC runs); otherwise the GC would interpret these as
+pointers and try to follow them and Undefined Behaviour would ensue. Thus, for
+safety, many library-defined array operations have to conservatively initialise
+array slots at the cost of performance.
+
+We propose to expand the arrays API in
+[GHC.Exts](https://hackage.haskell.org/package/base-4.14.0.0/docs/GHC-Exts.html)
+by picking the low hanging fruit of missing primitives which would improve the
+performance of key libraries, such as
+[unordered-containers](https://github.com/tibbe/unordered-containers/tree/d0b78afc028ef307a3cdfe1dd95feaf4ec47f195).
+
+## Proposed Change Specification
+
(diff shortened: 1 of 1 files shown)


treeowl:
I don't see why we would need `singleton#`. `newArray#` and `unsafeFreezeArray#` seem sufficient. The other proposed functions would be valuable.

proposals/0000-extend-array-primitives.md
```diff
@@ -0,0 +1,80 @@
+---
+author: buggymcbugfix
+date-accepted: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/334>).
```
AndreasPK, line 8:
```suggestion
This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/334).
```

AndreasPK:
It would be good to give examples of the fully qualified names for the proposed primops for at least one array type.

AndreasPK:
> I don't see why we would need `singleton#`. `newArray#` and `unsafeFreezeArray#` seem sufficient. The other proposed functions would be valuable.

`new >>= unsafeFreeze` currently writes the info table twice. This could be avoided with a pure singleton# function. It could also be avoided by a sufficiently smart backend and carries little overhead. So maybe still not worthwhile.

proposals/0000-extend-array-primitives.md
```diff
@@ -0,0 +1,80 @@
...
+[GHC.Exts](https://hackage.haskell.org/package/base-4.14.0.0/docs/GHC-Exts.html)
+by picking the low hanging fruit of missing primitives which would improve the
+performance of key libraries, such as
+[unordered-containers](https://github.com/tibbe/unordered-containers/tree/d0b78afc028ef307a3cdfe1dd95feaf4ec47f195).
+
+## Proposed Change Specification
+
+Below is the current list of proposed new primops, in order of priority.
+
+- `insert# :: array# a -> Int# -> a -> array# a`
```
AndreasPK, line 31: Please state the the expected semantics for the primops in the proposal itself.

vilem-ff:
Sorry, I opened this with my wrong GitHub account. I've reopened it here: https://github.com/ghc-proposals/ghc-proposals/pull/335

proposals/0000-extend-array-primitives.md
buggymcbugfix: Oops, I never defined what `concat` is meant to do. But for the other ones, is there any value in copying and pasting the `unordered-containers` code?

PR #335, opened 2020-05-23
Add draft proposal for extending array primitives

buggymcbugfix:
We propose to expand the arrays API in [GHC.Exts](https://hackage.haskell.org/package/base-4.14.0.0/docs/GHC-Exts.html) by picking the low hanging fruit of missing primitives which would improve the performance of key libraries, such as [unordered-containers](https://github.com/tibbe/unordered-containers/tree/d0b78afc028ef307a3cdfe1dd95feaf4ec47f195).

[Current RFC rendered here.](https://github.com/buggymcbugfix/ghc-proposals/blob/master/proposals/0000-extend-array-primitives.md)

### DIFF
--- a/proposals/0000-extend-array-primitives.md
+++ b/proposals/0000-extend-array-primitives.md
@@ -1,0 +1,80 @@
+---
+author: buggymcbugfix
+date-accepted: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/335).
+
+# RFC: Expand arrays APIs with efficient primops
+
+## Motivation
+
+Operations that allocate new arrays (`Array#`, `MutableArray#`, `SmallArray#`,
+`SmallMutableArray#`) must ensure that the array never contains garbage values
+during safe points (when the GC runs); otherwise the GC would interpret these as
+pointers and try to follow them and Undefined Behaviour would ensue. Thus, for
+safety, many library-defined array operations have to conservatively initialise
+array slots at the cost of performance.
+
+We propose to expand the arrays API in
+[GHC.Exts](https://hackage.haskell.org/package/base-4.14.0.0/docs/GHC-Exts.html)
+by picking the low hanging fruit of missing primitives which would improve the
+performance of key libraries, such as
+[unordered-containers](https://github.com/tibbe/unordered-containers/tree/d0b78afc028ef307a3cdfe1dd95feaf4ec47f195).
+
+## Proposed Change Specification
+
+Below is the current list of proposed new primops, in order of priority.
+
+- `insert# :: array# a -> Int# -> a -> array# a`
+  - Defined [here](https://github.com/tibbe/unordered-containers/blob/d0b78afc028ef307a3cdfe1dd95feaf4ec47f195/Data/HashMap/Array.hs#L357-L374) in
+  unordered-containers.
+- `pair# :: a -> a -> array# a`
+  - Defined [here](https://github.com/tibbe/unordered-containers/blob/d0b78afc028ef307a3cdfe1dd95feaf4ec47f195/Data/HashMap/Array.hs#L266-L271) in
+  unordered-containers.
+- `singleton# :: a -> array# a`
+  - Defined [here](https://github.com/tibbe/unordered-containers/blob/d0b78afc028ef307a3cdfe1dd95feaf4ec47f195/Data/HashMap/Array.hs#L258-L264) in
+  unordered-containers.
+- `delete# :: array# a -> Int# -> array# a`
+  - Defined [here](https://github.com/tibbe/unordered-containers/blob/d0b78afc028ef307a3cdfe1dd95feaf4ec47f195/Data/HashMap/Array.hs#L442-L458) in
+  unordered-containers.
+- `concat# :: array# a -> array# a -> array# a`
+  - Is this useful? Andrew Martin mentioned that the need to concatenate two
+    arrays may indicate a flaw in how the arrays are being used, as the two
+    arrays should have been created as one in the first place. Can we find a
+    counterexample/compelling use case?
+
+## Examples
+
+Please see the links to unordered-containers in the previous section.
+
+## Effect and Interactions
+
+Purely additive and backward-compatible.
+
+## Costs and Drawbacks
+
+There will be more primops to maintain. However, there is only a very small overhead to this.
+
+## Alternatives
+
+Implementing these primops in a separate package, like [array-primops](https://hackage.haskell.org/package/array-primops-0.2.0.1).
+
+Drawbacks of this approach ([courtesy of @treeowl)](https://github.com/treeowl/ghc-proposals/blob/array-creation/proposals/0000-array-creation.rst#11motivation):
+
+1. Its cmm implementations won't be inlined, so they have a performance penalty for very short arrays (like the ones used in Data.HashMap, for example).
+2. Since it's written in cmm, it needs to track GHC internals particularly carefully, and it's not available to non-cmm-based backends.
+
+Both of these aspects make it an unattractive dependency.
+
+## Unresolved Questions
+
+Which primops are the most interesting?
+
+## Implementation Plan
+
+@buggymcbugfix will implement this as part of [his Summer of Haskell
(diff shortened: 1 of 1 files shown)


awpr:
I believe the primops `insert#`, `delete#`, and `concat#` could all be subsumed by the following (fairly ad-hoc) primop, which would additionaly support efficient rotation:

```
spliceArray#
  :: Array# a -- ^ source1
  -> Int# -- ^ source1 start offset
  -> Int# -- ^ source1 length
  -> a -- ^ filler
  -> Int# -- ^ filler length
  -> Array# a -- ^ source2
  -> Int# -- ^ source2 start offset
  -> Int# -- ^ source2 length
  -> Array# a
```

If we want to be even slightly more general, I think it's possible to make the following work:

```
type ArrOff a = (# Array# a, Int# #) -- Array and offset
type MutArrOff a = (# MutableArray# a, Int# #)
-- Spec of how to fill one contiguous chunk of an array:
-- * the destination offset
-- * the chunk length
-- * where to get the data: either a value, or a slice of an array
type ArrBlit a = (# Int#, Int#, (# ArrayOff a | MutArrOff a | a #) #)

-- Subsumes singleton# and current cloneArray#
blitArray1# :: Int# -> ArrayBlit a -> Array# a
-- Subsumes pair#, concat#, and delete#; additionally implements rotate, cons, and snoc
blitArray2# :: Int# -> ArrayBlit a -> ArrayBlit a -> Array# a
-- Subsumes insert#; additionally implements inserting an array slice and other (more questionable) 3-way concatenations.
blitArray3# :: Int# -> ArrayBlit a -> ArrayBlit a -> ArrayBlit a -> Array# a
-- I don't know of a use case for four or more.
```

A moderate downside is that it's possible to misuse these in a way that ruins the GC, but that's already true of other mutable array ops, if you do out-of-bounds reads or writes.  It would be the caller's responsibility to ensure that the set of `ArrayBlit`s cover all indices of the resulting array.

These could all share the core array-copying logic of their implementations, since it's just doing the same thing one, two, or three times.

buggymcbugfix:
@AndreasPK, @andrewthad, @chessai, I'm not managing to reconstruct how we can get the single operation `slice` which we discussed on Monday to support insertion, deletion and replacement (IIRC we used this to refer to something like a copy, followed by a write?):

    -- Disclaimer: I'm sketching--expect off-by-one errors etc. Also Omitting #s and State#.

    slice
      :: Array a -- source
      -> Int     -- slice beginning
      -> Int     -- slice length
      -> Int     -- element insertion index???
      -> a       -- element to insert
      -> Array a

This gives us

    insert
      :: Array a
      -> Int
      -> a
      -> Array a
    insert arr ix = slice 0 (sizeOf arr) ix

But then how do we define deletion or replacement in terms of this?

@awpr's suggested a `spliceArray#` operation, which I will call `slice2` for now. It is more clear to me how we could implement a bunch of operations in terms of this:

    slice2
      :: Array a -- source1
      -> Int     -- source1 start offset
      -> Int     -- source1 length
      -> a       -- filler
      -> Int     -- filler length
      -> Array a -- source2
      -> Int     -- source2 start offset
      -> Int     -- source2 length
      -> Array a

    insert arr ix el = slice2 arr 0 ix el 1 arr (ix + 1) (sizeOf arr - ix)

    delete arr ix = slice2 arr 0 (ix - 1) undefined 0 arr (ix + 1) (sizeOf arr - (ix + 1))

    replace arr ix el = slice2 arr 0 (ix - 1) el 1 arr (ix + 1) (sizeOf arr - ix) 

    append arr1 arr2 = slice2 arr1 0 (sizeOf arr1) undefined 0 arr2 0 (sizeOf arr2)

    cons arr = insert arr 0

    snoc arr = insert arr (sizeOf arr)

    rotate ...

    grow ...

buggymcbugfix:
Related to this proposal: https://gitlab.haskell.org/ghc/ghc/-/wikis/Doubleton-Arrays

buggymcbugfix:
Related MR: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3372

andrewthad:
What we discussed on Monday wasn't to back all three by a single primop. It was to have three separate primops, each of which perform slice-and-X:

* sliceAndInsert
* sliceAndReplace
* sliceAndDelete

andrewthad:
I'll copying in the contents of an email that motivates these three:

I've done some more digging into unordered-containers and thought more about other cache-coherent data structures, and here is what I now perceive to be the maximum-value set of immutable-array-to-immutable-array-single-element-modifying primops:

    insertArray :: Array a -> Int -> Int -> Int -> a -> Array a
    replaceArray :: Array a -> Int -> Int -> Int -> a -> Array a
    deleteArray :: Array a -> Int -> Int -> Int -> Array a

All three of three interpret the first three arguments as a slice into an array. The next argument, always an Int, is the index, after performing slicing, to target. In more natural language, you could say that all of these slice before performing a single-element modification. The important question to answer about this design is why? Why build slicing in when we already have `cloneArray` available to us? The reason is that we really want to be able to make the length of the targeted area explicit so that GHC can unroll insert/modify/delete. Such an optimization is only available when the length of the slice is statically known, but cache-coherent data structures tend to offer a number of situations where the length is statically known. I'll provide a concrete motivating example. In unordered-containers, the HashMap data constructors include a `Full` data constructor for 16-element arrays:

    data HashMap k v
      = Empty
      | BitmapIndexed !Bitmap !(A.Array (HashMap k v))
      | Leaf !Hash !(Leaf k v)
      | Full !(A.Array (HashMap k v))
      | Collision !Hash !(A.Array (Leaf k v))
      deriving (Typeable)

Further down in the code, there are some functions for updating these 16-element arrays. For this code, we want GHC to be able to unroll the memcpy needed to build the new array. Such an optimization is only practical if the primop includes the length of the slice to be replaced.

The utility of the slice+modify primops shows up again in the context of considering B+ trees. B tree nodes always split at the same size. If you are inserting into and splitting a node in a B tree of degree 8, you would end up doing this (assuming the element belonged in the right half of the tree).

    let lowerNode = cloneArray node 0 4
    let upperNode = insertArray node 4 4 newElementIx newElement -- creates an array of length 5

GHC is already smart enough to unroll the memcpy that backs cloneArray. If insertArray includes slicing information as it does above, GHC can be made smart enough to unroll the `memcpy` that backs `insertArray` as well.

buggymcbugfix:
Thanks @andrewthad for clearing up my misunderstanding. Would these primops have any advantages vis-à-vis @awpr's proposal of a single primop that does all of this?

awpr:
Certainly they're simpler and easier to implement.  I was trying to make sure the more-complex-but-more-flexible part of the design space gets explored, even if making targeted primops for each individual important case ends up being the right call.

For completeness I'll apply the changes described in my edit to the `blitArrayN#` set of primops. 
 To try to pare down some of the complexity and make them more directly comparable to other suggestions, I'll drop MutableArray support and curry the outermost unboxed sums, too:

```
-- Reference to suffix of an array
type ArrayOff a = (# Array# a, Int# #)

-- What to put in a particular chunk of an array
type BlitSrc a = (# ArrayOff a | a #)

-- blitArray2# x xn y yn creates an array of length xn+yn, where the first xn elements are from x, and the rest are from y
blitArray2# :: BlitSrc a -> Int# -> BlitSrc a -> Int# -> Array# a

-- etc. for blitArray1# and blitArray3#
```

In this form, they no longer have the problem of potentially creating arrays with uninitialized gaps.

There are some variations available on this theme: for one, `blitArray3#` can subsume the primops with fewer arguments: just pass `(# | undefined #)` and `0` for the unused blit chunks.  Additionally, we could unroll the unboxed sums into separate primops, if unboxed sums can't be used in primops / are too annoying in primops / interfere with being able to unroll memcpys.  For example, `spliceArray#` I sketched above is what you'd get by specializing `blitArray3#` to an array for its first and third chunks and an element for its second chunk.

buggymcbugfix:
Update: we have a working sketch for a variadic `*ArrayOf#` primop: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3571. I will update the proposal depending on feedback. CC @andrewthad @AndreasPK @chessai

treeowl:
I agree with the importance of filler. Something like `a1 ++ a2 ++ [x]` (and other permutations) is pretty likely to pop up.

PR #338, opened 2020-05-30
Replace AllowAmbiguousTypes with AMBIGUOUS pragma

jkoppel:
AllowAmbiguousTypes is almost necessary for effective use of TypeApplications, but, as it applies to an entire file, creates risk of bugs. 

The inspiration for this request comes from my collaborators at ByteAlly, which banned AllowAmbiguousTypes in their public-facing APIs after getting burned.

This proposal aims to de-risk AllowAmbiguousTypes by replacing it with a per-declaration pragma, similar to the treatment given to OverlappingInstances.

### DIFF
--- a/proposals/0000-ambiguous-pragma.rst
+++ b/proposals/0000-ambiguous-pragma.rst
@@ -1,0 +1,88 @@
+Replace AllowAmbiguousTypes with AMBIGUOUS pragma
+============================================
+
+.. author:: James Koppel
+.. date-accepted:: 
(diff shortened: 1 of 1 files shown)


int-index:
What about [#232](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0232-AmbiguousType-pragma.rst)? It's accepted but needs implementation.

jkoppel:
I swear I searched for that beforehand.

PR #339, opened 2020-05-30
Rename 229 and change handling of prefix @

int-index:
* Clarify the scope of the proposal by renaming it.
* Change the treatment of `f @t`, such that it's always treated as a type application regardless of enabled extensions. This is in violation of the Haskell Report, but the proposal as it stands already diverges from the report, whereas this amendment makes it possible to recover good error messages.

The corresponding change to the implementation is here: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3387

It was prompted by @phadej's observation that the existing implementation of the proposal produces low quality error messages.


phadej:
Thanks!

PR #341, opened 2020-06-09
Simplify subsumption has been implemented

monoidal:
Implemented in https://github.com/ghc/ghc/commit/2b792facab46f7cdd09d12e79499f4e0dcd4293f.


nomeata:
I think the implemented field is typically used to indicate GHC releases. Which release will this be in?

monoidal:
You're right, I've fixed it.

PR #92, opened 2017-11-10
Library-defined language extensions!

nomeata:
This is a rough first draft, and not near submission quality. I hope that the community will help me to flesh this out and detect and fix all the problems with it.

[Rendered](https://github.com/ghc-proposals/ghc-proposals/blob/lib-lang-exts/proposals/new-proposal.rst)

PR #343, opened 2020-06-24
Enable BangPatterns by default

AndreasPK:
[Rendered](https://github.com/AndreasPK/ghc-proposals/blob/bang_by_default/proposals/0343-bang-by-default.rst)

### DIFF
--- a/proposals/0343-bang-by-default.rst
+++ b/proposals/0343-bang-by-default.rst
@@ -1,0 +1,154 @@
+Enable BangPatterns by default
+===============================
+
+.. author:: Andreas Klebinger
+.. date-accepted:: Leave blank. This will be filled in when the proposal is accepted.
+.. proposal-number:: Leave blank. This will be filled in when the proposal is
+                     accepted.
+.. ticket-url:: Leave blank. This will eventually be filled with the
+                ticket URL which will track the progress of the
+                implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/343>`_.
+.. sectnum::
+.. contents::
+
+BangPatterns are a useful, popular and mostly harmless extensions
+that should be enabled by default.
+
+In the current eco system not doing so carries a higher cost than changing the default.
+
+Motivation
+----------
+
+The BangPatterns extension being disabled by default carries a real cost
+as it's often used by users, but it's hardly ever beneficial to disable it.
+
+BangPatterns are incredible useful and enabling them comes at a very low cost
+of giving up the ability to define the ``!`` operator in a certain way.
+
+They give a convenient syntax to specify strict arguments which
+as far as I can tell has been well received and is in wide use.
+
+It seems to be the custom that code is in general written without
+the extension enabled at first. With users only enabling as needed.
+Common tools like hlint even encourage users to delete the pragma if
+not currently used. The result is usually one of the following:
+
+1. The extension is always enabled by the build tool.
+2. The extension is added/removed in the source as needed.
+3. The pragma is used independent of it's use.
+
+All three options have drawbacks which are discussed further down.
+
+I argue that the real world cost of BangPatterns being disabled
+by default is higher than the cost of enabling the extension by default
+would be.
+
+For this reason I consider it the pragmatic thing to enable the extension by default
+and propose as much.
+
+Proposed Change Specification
+-----------------------------
+
+* When invoked GHC should behave as if -XBangPatterns has been passed,
+* unless a language standard is requested. (Haskell98 or Haskell2010)
+* GHC provides a new (unstandardized) language level -XGHC2020 which is
+  based on Haskell2010 but includes the proposed change in behaviour.
+
+Examples
+--------
+
+These invocations would enable the BangPatterns extensions:
+
+::
+
+ ghc WithBangs.hs
+ ghc WithBangs.hs -XBangPatterns
+ ghc WithBangs.hs -XGHC2020
+
+
+These would not:
+
+::
+
+ ghc NoBangs.hs -XNoBangPatterns
+ ghc NoBangs.hs -XHaskell98
+ ghc NoBangs.hs -XHaskell2010
+
+Effect and Interactions
+-----------------------
+
+This would allow use of BangPatterns without having to explicitly enable the extension.
+I'm not aware of any significant interactions with other GHC extensions.
+
+Build tools currently passing -XHaskell2010 would need to be updated to support the new
+language level.
+
+Costs and Drawbacks
+-------------------
+
+Code defining the ``!`` operator infix will break if this proposal is accepted.
+This is easy to work around for actively maintained projects, but it is a cost.
+
+Implementation and maintenance is hardly an issue as the extension is already implemented
+and maintained.
+
+This proposal would signify a change between from GHC towards the haskell standard.
+
+Alternatives
+------------
+
+Defy -XHaskell2010
+~~~~~~~~~~~~~~~~~~~~~
+
+As an alternative to respecting the Haskell2010 standard we could change the default
+behaviour even when -XHaskell2010 is requested.
+
+While easier for build tools. It does not seem worthwhile overall.
+
+Enabling of BangPatterns by users via build tools
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This can cause issues as the interpretation of source files becomes dependent
+on external configurations. While it reduces the risk of files being compiled
+without BangPatterns enabled it does not eliminate it.
+
+Enabling of BangPatterns users in the source as needed
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+I generally use this approach (and it seems common). However I regularly find myself
+hitting build errors as I forgot to enable the extension for a particular file.
+It has also happened that I pushed changes to CI only to discover far later
+that CI failed since the file in question did not have BangPatterns enabled.
+
+In a similar fashion refactoring code can sometimes remove all bang patterns in a file.
+As a consequence the now unused BangPatterns extension should be removed from the file.
+
+Until eventually it's used again. At which point the build might fail again if I forgot
+to re-add the extension.
+
+Enabling of BangPatterns by the user independent of actual use
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+While this seems like a reasonable solution it still allows users to
+forget about adding the extension pragma to new source files.
+
+Users might also still get warnings from toolings about unused pragmas
+and many consider unused pragmas bad practice.
+
+Wait for a new Haskell standard
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+I do not expect a new standard to be released in the next few years.
+
+Unresolved Questions
+--------------------
+
+None
+
+Implementation Plan
+-------------------
+I would implement this change.


DanBurton:
Regarding the Drawbacks section, I don't think it would even break very much code with the (!) operator, since it only involves ! in a pattern context. Possibly no breakage at all.

nomeata:
Hmm, most of the discussion would apply to many other extensions that are popular and would not “hurt” to be on by default. Where would you draw the line?

phadej:
There is precedence

https://github.com/ghc/ghc/blob/master/compiler/GHC/Driver/Session.hs#L1626-L1640

So if one want to draw a line, it would make sense to limit extensions to one existed for 10-15 years ;)

AndreasPK:
> Hmm, most of the discussion would apply to many other extensions that are popular and would not “hurt” to be on by default. Where would you draw the line?

I've never had code fail to compile because I enabled BangPatterns. I had plenty of code fail to compile because it wasn't enabled. All other extensions I can think of either have a higher chance of causing failure or less benefit.

Do I think it would be better to enable a few more extensions by default? Yes!  
Do I think a proposal for doing so is likely to be accepted in a reasonable timeframe? No!

For this reasons **this** proposal draws the line at BangPatterns.

harendra-kumar:
To add a data point, we tried enabling all extensions used in the [streamly repo](https://github.com/composewell/streamly/pull/566) by default for all files, using GHC 8.8.3, it includes the `BangPatterns` extension as well. We did not see any code breakage or performance degradation for any of the extensions used except for one extension, the `TypeFamilies` extension. `TypeFamilies` caused the code to not compile in one case and caused large perf regressions in many cases.

goldfirere:
I'm not sure about this one.

While I'm in support of enabling `-XBangPatterns` by default *in practice*, I don't think this proposal achieves that goal. The problem is that cabal requires you to specify a language. If you specify `Haskell2010`, then `-XHaskell2010` is passed to GHC, and we lose the effect of this proposal. Presumably to get around this complication, the proposal also includes a new language level `GHC2020`, which is simply `Haskell2010` + `BangPatterns`. But why define a whole new language for that combination? `.cabal` files can always enable `-XBangPatterns` as a package-wide setting.

So what would happen if this were to be accepted, implemented, and released? For all the cabalized code out there, nothing would change. (I don't know what happens with stack, but I believe stack still uses `.cabal` files. Perhaps I can be educated away from this assumption.) All `.cabal` files would have to be updated to use `GHC2020` instead of `Haskell2010`, assuming `cabal` is updated reasonably quickly to support `GHC2020`. But can't all those `.cabal` files just enable `-XBangPatterns` directly?

And what happens the next time someone writes a proposal to enable an extension by default? Do we get `GHC2020b`? I don't want the perfect (a new standard) to get in the way of the good (a default `-XBangPatterns`), but this question adds to my discomfort with this proposal. If this were my only objection, it wouldn't be enough, but it adds to the list of cons, for me.

AndreasPK:
> But can't all those .cabal files just enable -XBangPatterns directly?

> But why define a whole new language for that combination? .cabal files can always enable -XBangPatterns as a package-wide setting. 

Having options is no excuse for bad defaults. It also doesn't cover the case of running ghc[i] on a file directly which I do surprisingly often.

> While I'm in support of enabling -XBangPatterns by default in practice, I don't think this proposal achieves that goal. The problem is that cabal requires you to specify a language. If you specify Haskell2010, then -XHaskell2010 is passed to GHC, and we lose the effect of this proposal. Presumably to get around this complication, the proposal also includes a new language level GHC2020, which is simply Haskell2010 + BangPatterns.

This is intentional. A change to the behaviour of -XHaskell2010 devalues Haskell2010 as a standard which is why I want to avoid it. But it's definitly the easier fix for the problem at hand.

In the end I only see two ways to achieve a new default going forward:
* Add a new language level which users are encouraged to use going forward
* Change the behaviour of -XHaskell2010

Both achieve the goal of new default for *new* projects, assuming GHC2020 is being adopted by the community.

Changing the interpretation of an existing standard [has been done before](https://mail.haskell.org/pipermail/haskell-prime/2011-January/003335.html) so it's not unheard of either. If there is more support for that approach I would be happy with that too.

AndreasPK:
> And what happens the next time someone writes a proposal to enable an extension by default? Do we get GHC2020b? I don't want the perfect (a new standard) to get in the way of the good (a default -XBangPatterns), but this question adds to my discomfort with this proposal. If this were my only objection, it wouldn't be enough, but it adds to the list of cons, for me.

I forgot to address this. I personally don't see a problem here. Should GHC decide to adopt another extension to be on by default I would welcome a GHC2021 in the future.

phadej:
@goldfirere 

> The problem is that cabal requires you to specify a language. If you specify Haskell2010, then -XHaskell2010 is passed to GHC, and we lose the effect of this proposal. 

with `cabal-version: 3.4` you will be able to omit `default-language` and thus have whatever "language" the compiler has by default

Anrock:
> default-language: GHC2020

Sounds like interesting idea. Given that Haskell20xx isn't happening in the foreseable future and GHC-Haskell is almost another language already and there are a lot of widely adopted de-facto-standard language extensions it kinda makes sense to bundle some of them under some GHC20XX quasi-standard-language.

Ericson2314:
a `GHC2020` like GCC's `-std=gnu++20` could be worth it, but should not be bikeshed on an extension-by-extension basis.

I agree having good defaults default, by I think junk in base causes *far* more trouble than off-by-default language extensions. Fewer language extensions means fewer choices, and that's generally good for beginners. Whereas actively harmful choices in base (as oppose to a lack of good choices) lead people stray.

`head` and `tail` have destroyed minds but I've never heard of beginners going crazy writing `seq` all over the place for lack of knowing about `!`.

goldfirere:
> with cabal-version: 3.4 you will be able to omit default-language and thus have active whatever "language" the compiler has by default

Aha! Then this proposal becomes more appealing to me.

Anrock:
> I think junk in base causes _far_ more trouble than off-by-default language extensions

Oh yes. I wonder how bad / hard it would be implement first-class Prelude-module support in ghc/cabal. Something like optional `prelude-module` field in cabal file that by default is `prelude-module:  Base.Prelude` and tied to `default-language`. With that it would more convenient to use alternative preludes and also will allow `default-language: GHC2020` to override `prelude-module` to something like `Base.PreludeGHC` without all those `head`s and `tail`s and other bad things.

Sorry for derailing this proposal. Maybe somebody is willing to bringup that GHC2020 thing as proper proposal? I still don't feel confident enough with Haskell and GHC to propose something nor do I have an ability to implement this.

Ericson2314:
[semi off topic]

@phadej 

> with cabal-version: 3.4 you will be able to omit default-language and thus have active whatever "language" the compiler has by default

Is there an issue for this? Because I would like to comment there (if it hasn't been already been done) that it seems like a terrible idea to increase the non-determinism of builds just to save a little bit of typing.

stephen-smith:
> * Change the behaviour of -XHaskell2010

I'm strongly against this alternative.

---

I mildly pro the proposal as is, and I think GHC20xx "languages" is a fine approach.

neongreen:
I feel like a language standard is a “big enough” entity that I don’t feel creating a new one is justified for the benefit this proposal brings. I understand that you want to limit the proposal’s scope. I’m still going to feel bad about the existence of a language standard that is literally “Haskell2010 + one extension”.

On the other hand, it will make it easier for GHC2021 to appear, which is definitely a good thing. Even if it’s just one extension per year, I’m happy with that. 

I have convinced myself in favor of this proposal.

AndreasPK:
I checked a large subset of hackage by simply enabling -XBangPatterns by default.

Out of 2471 packages I had:
* 2414 packages built successful
* 17 errors related to missing C dependencies
* 1 package which will need fixing with this change
* 17 packages which will work with the parsing changes in 8.14
* 2 packages which failed because certain executables were not on path (protobuff and something else I forgot)
* 1 package failed with a ghc-internal error
* 19 packages which I didn't categorize in detail. A good deal is bang pattern related but I think most should build with 8.14

AndreasPK:
@nomeata I believe this proposal is ready for review.

While a number of points has been raised in the discussion I think they are all covered by the proposal itself.

Ericson2314:
One issue with bang patterns in particular is that there are various other efforts nudging towards using the unlifted pointer runtime rep to demonstrate strictness. *Currently* that is not ergonomic like bang patterns and so it's essentially for a completely different use-case, but I don't know what the farther future holds. This makes me wary to stabilize bang patterns---they are a great extension one should feel free to use currently, but my bar for making something part of the language is higher than that.

simonpj:
My main worry here is that I don't seen any reason to give special attention to `BangPatterns`.  There are plenty more extensions that (like this one) don't hurt you if you don't use them, but are quite widely used.   E.g. `MultiParamTypeClasses`, `ExplicitForalls`, `ScopedTypeVariables`. 

Many Haskell modules have lots of extensions specified at the top, does one more or less make a difference?

I'd feel more comfortable if there were a lot of actual Haskell users yelling in support.

yav:
I don't think we should introduce `Haskell2020` for just this change, and I also don't think that the actual change is worth on its own, as it is easy enough to just enable in the `cabal` file, if one so desires.   I don't find the (counter) argument that your "code depends on the external environment" particularly compelling, as your code always depends on the external environment---in this particular case you'd get code that would work one way with some versions of GHC and another ways wither others, which seems more trouble than it's worth.

I would support, however, a separate proposal to define some sort of "GHC2020" "language" that enables a bunch of common and non-intrusive extensions by default, and I'd be fine with including `BangPatterns` in that.  I know that there is likely not 100% on what constitutes a "common non intrusive extension", but I have faith that the community can come to a reasonable agreement.

bravit:
I agree with @yav. I don't see any need to do something special about this particular extension.

gridaphobe:
Introducing a GHC2020 standard that could form the basis for, or inform, future official Haskell standards feels consistent with our principle that Good Proposals should have a path towards inclusion in a future standard. But I agree with @yav and others that BangPatterns alone doesn't feel like enough to warrant the new standard. 

But we have a good community of practitioners and educators involved here in the GHC Proposals process, so I think this would be a great place to work out the details of a broader set of extensions to call GHC2020.

AndreasPK:
> My main worry here is that I don't seen any reason to give special attention to BangPatterns. There are plenty more extensions that (like this one) don't hurt you if you don't use them, but are quite widely used. E.g. MultiParamTypeClasses, ExplicitForalls, ScopedTypeVariables.

GHC already has special treatment for BangPatterns. `-Wspace-after-bang` is enabled by default independent of the extension (since 8.8) and warns about code incompatible with this change. Going from warning about code incompatible with an extension, to enabling it by default seems like a small step to me.

Personally I also experienced more errors because of disabled `BangPatterns` than I have from the rest of your list combined.

But most of all I think this is a good change for people writing Haskell, independent of whether or not there are other good changes that coud be made.

> Many Haskell modules have lots of extensions specified at the top, does one more or less make a difference?

I have run into this pattern a bit over 10 times last month
* use a bang pattern
* compile/reload
* `Did you forget to enable BangPatterns?`
* enable extension
* recompile/reload

People say "just enable it in the cabal file". But [in practice](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3615) when I did so for ghc it wasn't just opening an MR/PR. But people wanted to collect opinions, mutiple people reviewed the change. Changes to the MR were made. 

As long as the default is at it is a lot of future projects will go through similar steps. And adding it to the cabal file doesn't even solve *all* problems that are solved by changing the default.

goldfirere:
@AndreasPK I agree with much of your last comment: changing the default would be a fine thing. But I agree with @yav's comment https://github.com/ghc-proposals/ghc-proposals/pull/343#issuecomment-661966636 that we should take a broader look at the problem, instead of doing something just specially for `BangPatterns`.

neongreen:
> we should take a broader look at the problem, instead of doing something just specially for BangPatterns.

I feel that /starting/ by doing something specifically for BangPatterns — as a path to the broader look — has a higher chance of success than aiming at a broader look right from the start.

simonmar:
It's quite a big step for GHC's default language to deviate from the most recent standard. The starting point in my opinion should be to discuss whether we want to do *that*, and *then* if we agree that's a good idea, discuss what set of extensions should be in that language.

phadej:
GHC's default languages **deviates from a standard**, it already disables `DatatypeContext` (which is accepted by Haskell Prime committee, but never made into any standard).

treeowl:
It also enables RelaxedPolyRec.

cartazio:
bang patterns was one of the uncontroversial parts of the H2020 discussions that did happen, but as the overall effort was under managed/under resourced, that wasnt a h2020 revision

monoidal:
The whitespace syntax for bang patterns has changed recently (#229). Previously, `f ! x` and `f !x` were the same thing. Now, only the latter denotes a bang pattern, even if `-XBangPatterns` are off (and then you're prompted for the extension). This change is a part of GHC 9.0 - it has not yet been released.

I'm not convinced we should enable an extension by default if it has recently undergone nontrivial changes. I'd expect more stability.

cartazio:
ok, thats a very good point.
in a certain philosophical sense, i guess we're moving to agda's sensibilities about whitespace seperation for infix operators?

nomeata:
> I would support, however, a separate proposal to define some sort of "GHC2020" "language" that enables a bunch of common and non-intrusive extensions by default, and I'd be fine with including `BangPatterns` in that. I know that there is likely not 100% on what constitutes a "common non intrusive extension", but I have faith that the community can come to a reasonable agreement.

Let’s pick up this line of thought. What would happe if, every year (or every two years) the GHC Steering Committee creates `GHC20xx`, simply as curated set of kinda-stable and non-intrusive extensions, and make GHC support that alias? Here are some observations:

 * We would not need to create a new report, the bar for extensions would be lower (don’t need an integration in the report; a section in the user library is fine).
 * Features would still live as extensions in GHC, and you can still choose Haskell2010+Ext, or even GHC2020+NoExt. We would _not_ be able to simplify GHC by no longer having to support all combinations.
 * The implementation is cheap; it’s just another langauge extension which implies other extensions.
 * Many library authors will not be able to use GHC2020 until a few years later, when the oldest GHC they want to support knows about the alias. Program authors might be able to do that.
 * As a commitee we still sometimes allow us to tweak extensions. But if an extenion has been “blessed” into these GHC-pseudo-language-standards, would we want to be more strict?
 * Reminds me of the proposal about library-defined sets of extensions (#92) – this way `ghc2020` could be a Hackage package, and even released GHCs _would_ understand the latest set! But it would also lead to a proliferation of extensions-sets (a bit like the proliferation of `base` packages; maybe not entirely a bad thing).
 * It might actually help Haskell2030 to be a thing, if some of the bikeshedding is taken care of, and the process to Haskell2030 might be simplified by looking at a GHC202x in total.
 * The bike-shedding about which extensions go in is confined to the commitee, and hopefully more efficient this way. The process could be: Every member can nominate any number of extensions, to include, maye a small rationale and then we do one round of independent approval voting, requiring a supermajority to really only pick uncontested extensions.

adamgundry:
I like the idea. This feels like a good extension of the work of the GHC Steering Committee, and a good way to signal "generally accepted" extensions with low overhead and a consensus mechanism for resolving disputes. A few thoughts:
 * I think there should be a venue for public ~bikeshedding~ discussion of the proposed list of extensions to be included, before the committee makes a decision. Perhaps something like the proposals process, with an MR including the proposed list along with a rationale for each extension, which could be discussed before the committee vote?
 * IMHO the bar should be higher than "widely used", it should be "widely used, relatively free from awkward corners, and no reasonable alternatives are known". So `BangPatterns` would make the cut for me (the minor whitespace rule tweaks notwithstanding), but e.g. `MultiParamTypeClasses` would not (too many awkward corners) and neither would `ScopedTypeVariables` (as some kind of explicit type abstraction would be a plausible alternative).
 * GHC's existing default language (which as others noted is already non-standard, something like `Haskell2010`+`NoDatatypeContexts`+`RelaxedPolyRec`+...) should be given a name (perhaps `GHC2019`, or whatever year it was last modified?).
 * The default language wouldn't necessarily need to be the latest one understood by GHC, e.g. GHC 10.x could support `GHC2025` but default to `GHC2020` if no language is specified.

treeowl:
Note: `RelaxedPolyRec` is a bit special in that it's not actually possible to turn it off.

nomeata:
> Perhaps something like the proposals process, with an MR including the proposed list along with a rationale for each extension, which could be discussed before the committee vote?

Yes, agreeed. In a way, it _is_ an extension proposal, namely for the “extension” GHC2020. So it makes sense to follow the same process in applicable parts – main difference would be that the voting is mandatory, and is actually a vote on each nominated extension.

> widely used, relatively free from awkward corners, and no reasonable alternatives are known

Not sure about only blessing things that are free of alternatives. Maybe it is precisely the value of a standard to say “yes, there are many ways to do this, but it’s better for all of us to commit to one variant”.

I think the bar should be “a supermajority of the committee things that this extensions could be in the next language standard”, and that among the committee, all possible requirements are enforced by at least some members. Of course in each concrete discussions, arguments like “it’s unclear whether an alternative not just exists, but might be better” are valid and will be considered.

treeowl:
When it comes to `ScopedTypeVariables`, there are some pretty prominent voices who'd prefer to see it done a different way. But yes, I think you're generally right. I'm not so sure `MultiParamTypeClasses` itself is an issue, so much as exactly which of the relaxations bundled together as `FlexibleInstances` should go along with it. Yes, I am saying `FlexibleInstances` should probably be split.

nomeata:
While we will appreciate such input, this is _not_ the place to discuss anything else but `BangPatterns` and/or the general idea of naming sets of extensions in a GHCxxxx langauge.

goldfirere:
> IMHO the bar should be higher than "widely used", it should be "widely used, relatively free from awkward corners, and no reasonable alternatives are known".

This statement looks sensible on its surface, but will cause trouble. One of the reasons that Haskell2020 foundered is that this standard is remarkably hard to meet. Even `-XFlexibleContexts` has awkward corners.

adamgundry:
> > IMHO the bar should be higher than "widely used", it should be "widely used, relatively free from awkward corners, and no reasonable alternatives are known".
> 
> This statement looks sensible on its surface, but will cause trouble. One of the reasons that Haskell2020 foundered is that this standard is remarkably hard to meet. Even `-XFlexibleContexts` has awkward corners.

@goldfirere that's why I included the weasel word "relatively". :wink: But yes, I'd argue for a fairly conservative approach to `GHC20xx` extension sets, including leaving out `FlexibleContexts`/`FlexibleInstances` as too awkward, and hence probably not taking `MultiParamTypeClasses` either.

This might help to create pressure to simplify and shore up existing extensions where part but not all of the extension is suitable for `GHC20xx`, e.g. `PatternSignatures` vs `ScopedTypeVariables` (#119).

@nomeata "could be in the next language standard" is a better way of putting what I was trying to get at by "no reasonable alternatives are known". Indeed it isn't the absence of alternatives that is important, but a general feeling that the extension is likely to be the preferred approach for a standard, rather than one option amongst many.

hanshoglund:
I strongly support the notion of `GHC20xx` (making `GHC2020` the first instance).

That is quite a different idea from _this_ proposal as it stands though. Maybe @AndreasPK or @nomeata would like to make a separate proposal for `GHC2020` (as per above), including a tentative list of included extensions?

nomeata:
Sorry for the late notice, but the proposal at hand (BangPatterns on by default) was rejected by the committee – we want to keep up the priciple that deviations from the standard are guarded by a language pragma. But the idea of `GHC2020` got some positive feedback, and is worth pursueing, in a separate proposal.

PR #344, opened 2020-06-29
Improve NegativeLiterals

int-index:
At the moment, `NegativeLiterals` has the following issue, documented in the
User's Guide:

> One pitfall is that with `NegativeLiterals`, `x-1` will be parsed as `x`
> applied to the argument `-1`, which is usually not what you want. `x - 1` or
> even `x- 1` can be used instead for subtraction.

We propose to fix this.

[Rendered](https://github.com/int-index/ghc-proposals/blob/negative-literals-improved/proposals/0000-negative-literals-improved.md)

### DIFF
--- a/proposals/0000-negative-literals-improved.md
+++ b/proposals/0000-negative-literals-improved.md
@@ -51,7 +51,8 @@ Changing the behavior of an existing extension may not be the best practice.
 
 ## Alternatives
 
-Don't change anything.
+* Don't change anything.
+* Deprecate `NegativeLiterals` in favor of `LexicalNegation`.
 
 ## Unresolved Questions
 
--- a/proposals/0000-negative-literals-improved.md
+++ b/proposals/0000-negative-literals-improved.md
@@ -1,0 +1,52 @@
+---
+author: Vladislav Zavialov
+date-accepted: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/344).
+
+# Negative Literals Improved
+
+At the moment, `NegativeLiterals` has the following issue, documented in the
+User's Guide:
+
+> One pitfall is that with `NegativeLiterals`, `x-1` will be parsed as `x`
+> applied to the argument `-1`, which is usually not what you want. `x - 1` or
+> even `x- 1` can be used instead for subtraction.
+
+We propose to fix this.
+
+## Motivation
+
+1. Having `x-1` parsed as `x (-1)` may violate users' expectations.
+2. Fixing this infelicity would make `NegativeLiterals` a subset of `LexicalNegation`.
+
+## Proposed Change Specification
+
+When lexing a negative literal, require that it is *not preceded by a closing
+token*. See
+[#229](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0229-whitespace-bang-patterns.rst)
+for the definition of a closing token.
+
+## Effect and Interactions
+
+Code that relies on `x-1` being parsed as `x (-1)` will break.
+
+## Costs and Drawbacks
+
+Changing the behavior of an existing extension may not be the best practice.
+
+## Alternatives
+
+Don't change anything.
+
+## Unresolved Questions
+
+None at the moment.
+
+## Implementation Plan
+
+After the `LexicalNegation` patch is merged, the implementation will be
+trivial: modify `negLitPred` slightly.
--- a/proposals/0000-negative-literals-improved.md
+++ b/proposals/0000-negative-literals-improved.md
@@ -18,9 +18,14 @@ User's Guide:
 
 We propose to fix this.
 
+Additionally, unboxed literals (e.g. `-1#`) have the same issue even without
+the `NegativeLiterals` extension. We propose to fix this, too.
+
 ## Motivation
 
 1. Having `x-1` parsed as `x (-1)` may violate users' expectations.
+   As evidence, see [#18022](https://gitlab.haskell.org/ghc/ghc/-/issues/18022).
+
 2. Fixing this infelicity would make `NegativeLiterals` a subset of `LexicalNegation`.
 
 ## Proposed Change Specification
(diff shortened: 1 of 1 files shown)


Ericson2314:
I would like to end up with `x - whatever` (subtraction) and `x -whatever` (application to negated thing) being the only allowed options, and `x-whatever` and `x- whatever` being disallowed, whenever `NegativeLiterals` or `LexicalNegation` is enabled. But I understand that may to too much of a breaking change.

This is the next best thing, and certainly better than the status quo (of accepted but not-necessarily-implemented behavior). :+1:

int-index:
> But I understand that may to too much of a breaking change.

Well, `LexicalNegation` is to land in 8.12, and 8.12 hasn't been released yet. So if you get this amendment to #229 accepted real quick, then it won't be hard to implement. No breakage.

Ericson2314:
Oh well I mean what I proposed would be a bigger change to `NegativeLiterals` (with or without `LexicalNegation`).

int-index:
`NegativeLiterals` can't be concerned with `x-whatever` or `x- whatever` at all, as it's about literals. But `LexicalNegation` can be amended to disallow these forms.

Ericson2314:
Well `NegativeLiterals` *could* do more than what it says on the tin, but I see your point :).

galenhuntington:
Hmph, well I've been working on my own proposal in this domain over the last couple of weeks (I wrote the quoted part of the User's Guide above), which was to be much more detailed.  I suppose I won't bother continuing.  But a couple of comments on this one:

1.  The proposal says it will make `NegativeLiterals` a subset of `LexicalNegation`, but I don't see how that's true.  Even with this proposal, `-1 :: Foo` will still desugar as `negate (fromInteger 1) :: Foo`, whereas with `NegativeLiterals` it would be `fromInteger (-1) :: Foo`, which is not the same.

2.  I'm opposed to outlawing `x-whatever`.  It's inconsistent to allow `x*2` but not `x-1`.  And what would be the point of this change if it weren't allowed anyway?

int-index:
1. `LexicalNegation` has a special case for literals and desugars `-1` as `fromInteger (-1)`, just like `NegativeLiterals`.

2. Alright. I don't have much of an opinion on outlawing `x-y`, but it's not part of this proposal anyway.

int-index:
> I wrote the quoted part of the User's Guide above

Oh, somehow I assumed the disclaimer must've been there from the beginning. Thanks for piquing my curiosity: I've now discovered (through your commit message) the issue [`#18022`](https://gitlab.haskell.org/ghc/ghc/-/issues/18022). And by reading the discussion there, I've realized that my mini-proposal doesn't handle `-1#`, i.e. `x-1#` is still parsed as `x (-1#)`. This should also be fixed, I believe.

monoidal:
I support the proposal.

While we are at it, `LexicalNegation` currently has the same issue with unboxed integers (`\x -> x-1#` is not parsed as `\x -> x - 1#`).

int-index:
@monoidal This wasn't accounted for in the initial proposal, but it is reflected in the current version (and in the implementation, too).

goldfirere:
I support this proposal, and recommend that it be submitted to the committee so that the implementation can be released with 8.12.

int-index:
@nomeata I'd like to submit the proposal to the committee.

LeanderK:
I use haskell for more numerical work and would really welcome this proposal!

nomeata:
/remind me to check status in a week

simonpj:
Sounds like a good proposal to me.  But where is the original NegativeLiterals proposal?  I'd love for us to finish up with a single proposal that describes the final design, rather than a succession of deltas.

nomeata:
`NegativeLiterals` long predates the proposal process:

```
commit ef739635a8c2646112d2a1fa4c8715704aff1f1f
Author: Ian Lynagh <igloo@earth.li>
Date:   Wed Jul 31 18:43:11 2013 +0100

    Add NegativeLiterals extension
    
    I'd been meaning to do this for some time, but finally got around to it
    due to the overflowing literals warning. With that enabled, we were
    getting a warning for
        -128 :: Int8
    as that is parsed as
        negate (fromInteger 128)
    which just happens to do the right thing, as
        negate (fromInteger 128) = negate (-128) = -128
```

simonpj:
> NegativeLiterals long predates the proposal process:

Ah yes, thanks. Barely user-visible, except for `-128`.

> Fixing this infelicity would make NegativeLiterals a subset of LexicalNegation.

I'm not sure what this point (in the proposal) means.  (LexicalNegation is #229.)   It would great if the proposal said, or perhaps expressed as a diff to #229.

int-index:
> I'm not sure what this point (in the proposal) means.

It means enabling `-XLexicalNegation -XNegativeLiterals` has the same effect as just enabling `-XLexicalNegation`.

simonpj:
> It means enabling -XLexicalNegation -XNegativeLiterals has the same effect as just enabling -XLexicalNegation.

Super. Perhaps say this in the proposal, and let's deprecate NegativeLiterals.

Ericson2314:
> and let's deprecate NegativeLiterals.

Part of me wants to deprecate prefix negation (!) and *just* have `NegativeLiterals`. But this proposal is an improvement to negative literals either way. I also hope a future Haskell Report goes with `NegativeLiterals` (as in I dislike `NoNegativeLiterals` more than I dislike prefix negation).

galenhuntington:
> Part of me wants to deprecate prefix negation (!) and _just_ have the negative `NegativeLiterals` extension.

I was initially going to float this idea as well.  There are only negative literals; to negate an arbitrary expression, use `negate`.  It's simple, instead of having this one operator hardcoded into the language syntax; and consistent, with no subtle distinction between `-5` and `let x = 5 in -x` (as there is with these extensions).  But, I figured it would be a hard sell.  The current proposal at least makes negation work reasonably, with less change.

And I agree it would be good to have something like this in Haskell2030 (or whatever).  I also used to do a lot of numerical Haskell, and the negation situation was an annoyance.  And, `NegativeLiterals` as it is now was probably always a nonstarter due to the `x-1` problem.

simonpj:
We already have a MR !3624 that implements this (tiny) proposal.   If we get a decision this week, we could have this in 8.12 and be done with it.

nomeata:
The commitee has just accepted it

isovector:
Thank you so much for putting this together.

PR #346, opened 2020-07-14
Bracket-sensitive layout rule

cdsmith:
This short proposal changes the layout rule to allow a kind of formatting of close brackets that's common in most other programming languages.

[Rendered](https://github.com/cdsmith/ghc-proposals/blob/loose-layout/proposals/0000-bracket-sensitive-layout.md)

### DIFF
--- a/proposals/0000-bracket-sensitive-layout.md
+++ b/proposals/0000-bracket-sensitive-layout.md
@@ -127,7 +127,18 @@ A new language extension is added, called `BracketSensitiveLayout`.  When
 enabled, the following rule is added to the beginning of the layout algorithm in
 [section 10.3 of the Haskell 2010 Report](https://www.haskell.org/onlinereport/haskell2010/haskellch10.html#x17-17800010.3):
 
-* L (\<n\> : t : ts) (m : ms) = L (t : ts) (m : ms), **if** m = n, and t is one of: ), ], or }.
+* L (\<n\> : t : ts) (m : ms) = L (t : ts) (m : ms), **if** m = n, and t is a closing bracket.
+
+Without language extensions, the phrase "is a closing bracket" means one of `}`, `)`, and `]`.  More
+generally, using the token classifications in
+[proposal 229](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0229-whitespace-bang-patterns.rst#proposed-change-specification),
+a closing bracket is defined as a token that is "closing", but not "opening".  This extends
+closing brackets to include:
+
+* `#)` when `UnboxedTuples` are enabled.
+* `|]` and `||]` when `TemplateHaskellQuotes` is enabled.
+* `|)` when `Arrows` is enabled.
+* `⟧` and `⦈` when `UnicodeSyntax` is enabled along with any of `TemplateHaskellQuotes` or `Arrows`, respectively.
 
 ## Examples
 
--- a/proposals/0000-bracket-sensitive-layout.md
+++ b/proposals/0000-bracket-sensitive-layout.md
@@ -7,12 +7,12 @@ implemented: ""
 
 This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/346).
 
-# Looser Layout Rules
+# Bracket-Sensitive Layout Rule
 
 This proposal makes the following valid Haskell code:
 
 ```
-{-# LANGUAGE LooseLayout #-}
+{-# LANGUAGE BracketSensitiveLayout #-}
 
 foo = (
   1,
@@ -44,16 +44,16 @@ choose.
 
 ## Proposed Change Specification
 
-A new language extension is added, called `LooseLayout`.  When enabled, the
-following rule is added to the beginning of the layout algorithm in
+A new language extension is added, called `BracketSensitiveLayout`.  When
+enabled, the following rule is added to the beginning of the layout algorithm in
 [section 10.3 of the Haskell 2010 Report](https://www.haskell.org/onlinereport/haskell2010/haskellch10.html#x17-17800010.3):
 
 * L (\<n\> : t : ts) (m : ms) = L (t : ts) (m : ms), **if** m = n, and t is one of: ), ], or }.
 
 ## Examples
 
-With the `LooseLayout` extension enabled, the following are all valid, while
-they were previously syntax errors:
+With the `BracketSensitiveLayout` extension enabled, the following are all valid,
+while they were previously syntax errors:
 
 ```
 foo = (
@@ -162,8 +162,7 @@ a solution.
 
 ## Unresolved Questions
 
-The extension name is terrible and doesn't communicate what it's doing very
-well.  I cannot yet think of a better name.
+None yet.
 
 ## Implementation Plan
 
--- a/proposals/0000-bracket-sensitive-layout.md
+++ b/proposals/0000-bracket-sensitive-layout.md
@@ -81,8 +81,24 @@ mylist = [
 ```
 
 In Haskell, due to the layout rule, this is not allowed most of the time.
-Haskell has, therefore, developed some unique formatting conventions,
-such as:
+
+```
+foo = (
+  1,
+  2
+)
+
+test.hs:4:1: error:
+    parse error (possibly incorrect indentation or mismatched brackets)
+  |
+4 | )
+  | ^
+```
+
+There are two consequences of this:
+
+**Divergence in formatting:** Haskell has, therefore, developed some unique
+formatting conventions, such as:
 
 ```
 indentedCloseBracket = (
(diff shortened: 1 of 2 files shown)


tfausak:
That was quick! See this discussion on Reddit: https://np.reddit.com/r/haskell/comments/hr5c2n/haskell_style_guide/fy2k681/?context=2

erikd:
As someone who codes Haskell professionally in a large team (that is geographically dispersed, with a huge range of Haskell experience and skill) the number of syntactic bike shedding proposals like this is incredibly frustrating and scary.

Yes, I am well aware that language pragmas like this are opt in, but the ever increasing number of language pragmas will cause difficulties for teams like mine which seems to have a lot of difficult enforcing and maintaining standards across a huge code base, with a geographically dispersed team, with a wide range of skills.

To give an example, I was recently reviewing some code that looked like:
```
    liftIO do
        something
        someOtherThing
```
First thing that struct me was "how does that even compile?". Then I looked to the top of the file and found a language pragma I had not seen before `BlockArguments`, which I then looked up in the documents: https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#extension-BlockArguments . This language pragama saves two charcters, a space and a `$` which to me is a an infinitesimally small benefit. Yes, this could be addressed with policy "thou shall not use the BlockArguments pragma", but in my organisation agreeing to and enforced such policy has been incredibly difficult. If `BlockArguments` had been available from the start of Haskell and baked into the language I would have no problem with it. The fact that it is optional **IS** the problem.

I will therefore be opposing this and other similar syntactic bikeshedding proposals.

cdsmith:
I agree with @erikd as far as to say that this proposal should be rejected if the general feeling is that this should not eventually become part of a future Haskell Report.

DanBurton:
Imo just build it into ghc, let the Haskell Report catch up later.

cdsmith:
@DanBurton Yes, that would definitely be the plan.  I just mean that the GHC committee should make their decision on a GHC implemention with an eye toward whether this would likely be adopted in a future Haskell Report.  If they feel it will not, so that this would be only a GHC extension, then GHC probably shouldn't implement it.  Changes always have short-term costs, but the benefit here is not great enough to justify forking Haskell syntax in the long run.

tomjaguarpaw:
My initial reaction is a negative one.  I have a few observations and questions.

1. I am pessimistic about the value of small syntax extensions in general.  See "[Some arguments against small syntax extensions in GHC](https://osa1.net/posts/2020-01-22-no-small-syntax-extensions.html)".

2. I would go even further than @erikd.  Syntax extensions can be actively harmful even when they are not enabled!  (e.g. `BlockArguments`, see http://h2.jaguarpaw.co.uk/posts/unhelpful-ghc-error-messages/).  Therefore I could only support this proposal if it is guaranteed that it doesn't make error messages less helpful (and indeed either way presence or absence of such behaviour ought to be noted under "Effect and Interactions").

3. I don't understand the motivation.  Amongst the justification is that "Partly [b]ecause of [this syntax restriction], Haskell has developed some unique formatting conventions, such as leading commas".  I don't understand the connection between comma-prefix style and not being able to use a closing token in a column that other languages allow. @tfausak [gave this as a rationale on the Reddit discussion](https://np.reddit.com/r/haskell/comments/hr5c2n/haskell_style_guide/fy2hpc5/) but I'm afraid I don't see the connection between the two things.

4. The proposal seems to support only a very weak form of what other languages allow.  It would allow `foo1` but not `foo2`.  Python *does* allow `foo2`. Arguably this proposal does what Python *should* have done, that is, require indentation inside multi-line list/tuple/dict literals too, but I think we need to see more examples of what this syntax supports and does not support.

```haskell
foo1 = (
  1,
  2
)
```

```haskell
foo2 = (
1,
2
)
```

5.   I *think* this proposal would allow `foo3` but I am not certain.  I think it would *not* allow `foo4`. Could someone clarify?

```haskell
foo3 = ( (
  1,
  2
), (
  3,
  4
) )

```haskell
foo4 = ( (
  1,
  2
),
(
  3,
  4
) )

**Comment by AntC2** [comment]
I have some sympathy with this proposal, after all the way to avoid layout annoyances is explicit `{  ;  ;  }`. Then if I use explicit `(  ,  ,  )` or `[  ,  ,  ]`, why do I still fall over layout complaints? Shouldn't the compiler take everything from `(` to `)` as one syntactic unit, indented as per the opening `(`.

Calling @int-index : I believe there's some work in hand to recognise in the lexing what are opening brackets and their corresponding close. Note there's all sorts of combo-brackets used by Template Haskell, also by Rebindable Syntax IIRC.

I also have some sympathy that fiddling with the syntax is not high-value, but does add cognition costs for Haskellers, even if the implementation and maintenance costs inside GHC are small.

As to the "future Haskell Report": I don't see any likelihood there will be another Haskell Report. There's supposed to be a Haskell 2020. The Haskell-Prime process/team are supposed to be progressing that. Posts to that mailing list so far over halfway into the final year for delivery? None.

**Comment by AntC2** [comment]
BTW, this doesn't fall foul of layout rules, seems inconsistent with the treatment of `(  ,  ,  )`

    data Foo = MkFoo{ foo1, foo2 :: Int }
    
    foo12 = MkFoo{
      foo1 = 1,
      foo2 = 2
    }                        -- doesn't need to be indented

To explain why the leading comma style is popular (as asked in the reddit thread): it's easier for code maintenance; if you need to cut/paste lines within comma-separated sequences and the comma is at the end of the line, and you paste it just before `)`, then you need to edit out the `,` (and probably insert a comma somewhere else such as at the end of the line above). That would apply just as much for comma-sequences in any C/Java-style programming languages. Don't they use leading commas?

There's another proposal to ease this a bit -- for example in import/export lists: allow a trailing comma before `)` and ignore it.

Re @erikd 's comments, I'm surprised anyone " who codes Haskell professionally in a large team" uses layout or relies on it to read large codebases. I've always seen layout as at best a gimmick and at worst a pain in the backside. I would expect codeshops to insist nobody uses layout/always puts explicit `{  ;  ;  }`. Layout is great for hacking together a few lines of code; not for industrial strength.

**Comment by cdsmith** [comment]
@tomjaguarpaw Thanks for the suggestion about considering error messages.  That should definitely be included in the proposal.  I am tempted to cite it as a possible advantage of the proposal, though, since it gives the lexer an opportunity to detect a specific common mistake, and more clearly explain why the parse error happens (versus today's message, which just says something vague about "mismatched brackets or incorrect indentation").  Do you foresee any problems here?

If it was unclear, it isn't a goal to accept anything that specific other languages would accept.  Rather, it is a goal to accept a particular style of indent that is commonly used in practice in most other languages.

I agree with you that `foo3` would be accepted.  It is clearly not formatted well, but it remains unambiguous what was intended. 
 `foo4` would still be rejected because a semicolon is inserted before the last `(`.

**Comment by tomjaguarpaw** [comment]
Thanks for the response @cdsmith.

If the error messages improve then so much the better! I don't have any specific knowledge there, but as a user I certainly don't want error messages to get worse.

I was probably unclear with my comparison to Python.  I'm trying to say something like "users might be surprised that `foo1` is permitted by not `foo2` because they seem to have a similar structure, especially to those familiar with other languages".

```haskell
bar1 = (
  1,
  2
)
```

```haskell
bar2 = (
  1,
  2,
3 )
```

Thanks for the clarification regarding `foo3` and `foo4`.

cdsmith:
@AntC2 Thanks for your comments.

As regards the Haskell Report, I understand that Haskell 2020 has mostly stalled, but I'm optimistic enough to not assume that there will never be another Haskell Report.  In any case, it seems reasonable then to make decisions conditionally upon there being a future path for the language; if there isn't, then it doesn't matter so much in the long run, anyway.

I think this proposal is actually quite separate from the leading-commas style, and it was my mistake to tie the two together so strongly.  I have reworded the motivation section to (I hope) more clearly address that it's more about *not* being able to use the formatting style that is dominant in most other programming languages, rather than about which specific convention has replaced it.

Your last example is a weird one.  In fact, even this is valid:

```
data Foo = MkFoo{ foo1, foo2 :: Int }

foo = 42
  where
                       foo12 = MkFoo {
  foo1 = 1,
  foo2 = 2
                       }
```

This is because curly braces effectively disable most other layout rules by pushing a special layout context on the stack.  I don't know if this was intentional, but it seems like a bug in the Haskell Report, to be honest.

ndmitchell:
@erikd I've managed geographically dispersed teams with varying skill sets, and language extensions wasn't even slightly a problem. In case it's useful, I've written out how we did it: https://neilmitchell.blogspot.com/2020/07/managing-haskell-extensions.html. Making the rules was pretty pleasant (one time we did it in the pub). Enforcing them was easy with HLint (although the first time I did have to modify HLint to enable enforcement).

alanz:
As someone who works on tooling, and in particular things like ghc-exactprint, and all the various formatters and transformation tools built off it, this kind of proposal fills me with misgivings.  It potentially means we have to add another layer of special processing, which must be aware that this rule is in effect, and contributes to the combinatorial explosion of possibilities.

Ericson2314:
> bug in the Haskell Report, to be honest.

Honestly the report's layout rules seem kinda crazy for me, or at least it's one of the few things about Haskell that seems completely intuitive despite their use being intuitive.

I would like a from-scratch reimagining at some point, perhaps this feature can wait for that. It is a feature I would like.

------

> combinatorial explosion of possibilities.

We really need to get to a place where features are independent things that compose and parametricity saves us. GHC today is mephatically not that place.

------

It doesn't help in the short term, but i would encourage people to join me in writing more `-XNo*` proposals! If we can remove a bunch of stuff in future language reports, it will free up budget for other things.

tomjaguarpaw:
> i would encourage people to join me in writing more `-XNo*` proposals!

Have you already written some? If so could you provide links? I'd be interested to see them!

AntC2:
> As someone who works on tooling, ..., this kind of proposal fills me with misgivings. It potentially means we have to add another layer of special processing, which must be aware that this rule is in effect, and contributes to the combinatorial explosion of possibilities.

Or ...

> This is because curly braces effectively disable most other layout rules by pushing a special layout context on the stack. I don't know if this was intentional, but it seems like a bug in the Haskell Report, to be honest.

The current layout rule needs 'special processing'; and it gives inconsistency or special status to `{  ;  ;  }` and `{  ,  ,  }` vs other syntax that uses some form of brackets and separators.

The lexing work I was trying to remember is [proposal #229 ](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0229-whitespace-bang-patterns.rst). It categorises lexemes as Opening, Closing or Separators aot. Perhaps this proposal should be adjusted to apply to all sequences of Opening-Separators-Closing, as per that lexing.

Then that should give _simpler_ tooling. If @alanz wants simple-as-possible tooling, I suggest just getting rid of the layout rule altogether. To satisfy @Ericson2314, we can call that `-XNoLayout`. (BTW the first para of John's comment makes no sense to me; should there be a 'not' somewhere?)

cdsmith:
@AntC2 Thanks for the reference.  At least it would make sense to say that the new layout rule proposed here should apply to all tokens that are "closing but not opening" in the language of proposal 229.  That matches the list here when there are no language extensions, but it has the advantage of automatically extending the same rule to handle TH brackets, unicode syntax, and presumably other future extensions as well.

Ericson2314:
@AntC2 Sorry, I edited to fix my sentence---my browser doesn't think "unintuitive" is a word and then i hastily removed the "un-".

To elaborate on that sentence, most things in Haskell are a) intuitive after a while b) specified in a way that matches the intuition, and being able to connect the intuition and formalism is a great joy. Layout completely defies that (to me), and that makes me sad.

cdsmith:
@Ericson2314 I appreciate your ideas about making a larger change to the layout rule.  I agree that the layout rule today seems unintuitive, though I think it could be fixed without doing anything too crazy.  At first glance, I think the root of the problem is just trying to collapse too many states, because the Report tries to shoehorn everything into one integer per layout level, which is not an expressive enough type.  I will spend some time thinking about how one might rewrite the layout algorithm to do what one would expect of it, and throw something about it into the alternatives section.

Ericson2314:
@cdsmith Yes that sounds like a much more tractable way to make the rules clearer. We could, even if don't have the token tree datastructure that witnesses this, still track nesting, and ensure opening and closing braces and increases and decreases in indentation alike match. Good luck!

Also, one other place layout came up before was with `\of`—maybe we can make the one rule `\` to rule them all with new layout rules after all.

cdsmith:
@Ericson2314 @AntC2 

I threw out a first draft of rewriting the layout rules to treat all brackets consistently, as well as noting some initial advantages and disadvantages.  I'm not yet confident that it's correct or complete.  Feedback is welcome.

cdsmith:
This has fizzled out without lots of support, so I'll close it for cleanliness.

PR #216, opened 2019-03-27
Local do

aspiwack:
This proposal introduces a mechanism to override the meaning of a do-notation block (a bit like with `-XRebindableSyntax`, but on individual `do`-s).

```haskell
do @IndexedMonad.builder
x <- someIndexMonadExpression
…
```

It superseeds a previous attempt ( #78 ).

This proposal was authored by @Tritlo and @aspiwack .

[rendered](https://github.com/tweag/ghc-proposals/blob/local-do/proposals/0000-local-do.rst)

#### Todos

- [x] See https://github.com/ghc-proposals/ghc-proposals/pull/216#issuecomment-477164098 : sum up arguments for either using the default behaviour without do or using the ambiant `builder` binding.
- [x] See https://github.com/ghc-proposals/ghc-proposals/pull/216#issuecomment-477168538 : “Worth examining carefully what methods are allowed in F#'s computation expressions.”
- [x] See https://github.com/ghc-proposals/ghc-proposals/pull/216#issuecomment-477168538 : consider support for list comprehension (in which case a field for parallel list comprehension would be warranted).
- [x] See https://github.com/ghc-proposals/ghc-proposals/pull/216#issuecomment-477524138 : add alternative syntax: `do via builder`.
- [x] See https://github.com/ghc-proposals/ghc-proposals/pull/216#issuecomment-477646013 : explain the reasoning behind the `@` syntax.
- [x] See https://github.com/ghc-proposals/ghc-proposals/pull/216#issuecomment-477168538 : be more explicit that we are considering linear-IO example from the paper.
- [x] See https://github.com/ghc-proposals/ghc-proposals/pull/216#issuecomment-477168538 : be more precise on the semantics
- [x] See https://github.com/ghc-proposals/ghc-proposals/pull/216#issuecomment-477168538 : add support for recursive do expressions (by way of an `mfix` field in builders)
- [x] See https://github.com/ghc-proposals/ghc-proposals/pull/216#issuecomment-477173618 : add link to F# computational expressions doc.

PR #350, opened 2020-07-27
Negative literals and meaning-preserving SCC implemented

monoidal:
While we are at it: there's an "Implemented" label that is currently not always used. I think it's pretty convenient to see implemented proposals with a tag. Could you add it to #29, #37, #54, #71, #99, #103, #111, #176, #190, #195, #216, #246, #287, #344?

### DIFF
--- a/proposals/0176-scc-parsing.rst
+++ b/proposals/0176-scc-parsing.rst
@@ -3,8 +3,8 @@ Meaning-preserving parsing rules for SCC annotations
 
 .. author:: Vladislav Zavialov
 .. date-accepted:: 2019-04-17
-.. ticket-url::
-.. implemented::
+.. ticket-url:: https://gitlab.haskell.org/ghc/ghc/-/issues/15730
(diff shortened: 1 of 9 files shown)


nomeata:
Many thanks for the convenient list, added labels to all of them.

proposals/0176-scc-parsing.rst
```diff
@@ -3,8 +3,8 @@ Meaning-preserving parsing rules for SCC annotations
 
 .. author:: Vladislav Zavialov
 .. date-accepted:: 2019-04-17
-.. ticket-url::
-.. implemented::
+.. ticket-url:: https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0176-scc-parsing.rst
```
nomeata, line 6: The ticket url should be something in the gitlab (either an issue or a MR)

monoidal: Fixed.

PR #351, opened 2020-07-29
NoIncomplete

Ericson2314:
Provide an extension to disallow syntax that would desugar to run-time errors without a `-fdeferred-*` flag.

[Rendered](https://github.com/Ericson2314/ghc-proposals/blob/no-sugared-incompleteness/proposals/0000-no-incomplete.rst)


proposals/0000-no-sugared-incompleteness.rst
```diff
@@ -0,0 +1,144 @@
...
+Provide an extension to disallow syntax that would desugar to run-time errors without a ``-fdeferred-*`` flag.
+
+Motivation
+----------
+
+Going back to Milner's "Well-Typed expressions do not go wrong", typed functional programmers, and Haskellers in particular, usually prefer their programs to be free of run-time failures stemming from programmer error.
+For those learning Haskell, learning this mindset of being fastidious and aiming for code that is correct-by-construction about even banal errors is sometimes as important and novel as learning the language itself.
+We should help those learners by trying to guide them in that direction every step of the code they write.
+
+Unfortunately, the compiler and language do always help with this.
```
nomeata, line 26: ?
```suggestion
Unfortunately, the compiler and language do not always help with this.
```

maralorn:
Can you add a rendered link?

TheMatten:
I can imagine this name being hard to write - maybe somethihing like `[No]PartialSugar` would work better?

proposals/0000-no-sugared-incompleteness.rst
```diff
@@ -0,0 +1,144 @@
...
+Why write yet another proposal, for yet another language extension, when this is solved problem?
+
+Well, I just don't think ``-Werror=...`` is the right mechanism to rely on.
+This is, of course, a matter of opinion, but to me warnings are for "extra" analyses, not "essential" ones like type checking which enforce the language proper.
+Deciding whether e.g. pattern matching is exhaustive feels to me like one of those essential analyses.
+I am asking that we switch our vantage point and instead of considering programs with implicit partiality well-formed programs that are iffy, consider them ill-formed programs that we can permissively accept anyways.
+
+So, what's the practical upshot of this lawyering?
+And what mechanism might we use rather than ``-Werror=...``?
+There is another mechanism for controlling run-time errors which I think is a perfect model for how these programs should be dealt with: ``-fdefered-type-errors``.
```
monoidal, line 41: There's no `-fdefered-type-errors`, only `-fdefer-type-errors` and `-Wdeferred-type-errors`.

proposals/0000-no-sugared-incompleteness.rst
Ericson2314: Thanks!

Ericson2314:
> Can you add a rendered link?

Done, thanks.

> I can imagine this name being hard to write - maybe somethihing like `[No]PartialSugar` would work better?

I think I subconsciously chose the biggest mouthful to hint at the need to choose a better name :).

goldfirere:
I would just have the extension be `-XIncomplete`.

If we started from a complete language and someone suggested allowing these forms of incompleteness, I imagine we would want separate control over e.g. pattern-match incompleteness and incomplete record updates. But maybe it's not worth splitting hairs here.

Why have the new `-fdefer-*` flags? How are these different from `-XIncomplete`?

I'm broadly in support.

Ericson2314:
> Why have the new `-fdefer-*` flags? How are these different from `-XIncomplete`?

Basically, I want everyone to be able to turn off the extension (i.e. do `-XNoIncomplete`) all the time. Rather than doing `{-# LANGUAGE Incomplete #-}`, I want people to do `{-# OPTIONS_GHC -fdefer-* #-}`, just as they would do `{-# OPTIONS_GHC -fdefer-type-errors #-}`

Why? To me `LANGUAGE` pragmas indicate a consistent intent to use an extended/restricted Haskell variant, while `OPTIONS_GHC` pragmas indicate "I've been debugging stuff so long I'm going to just write down the flags I'm using". I don't really believe most people are/should be indefinitely committing to using implicit partial syntax as language choice, and the extended debugging framing covers what they are actually doing much better.

Basically, the same reason [`-XDynamicTypes`](https://github.com/ghc-proposals/ghc-proposals/pull/217) is an April fools joke and not a replacement for the `-fdefer-type-errors` flag. :)

hanshoglund:
I support this strongly.

Suggestion: assuming #319 is accepted, have this extension imply `-XNoFallibleDo`. I suspect nobody would wish for incompleteness  _in do-blocks only_.

Ericson2314:
@hanshoglund Glad you like it so much!

Regarding your suggestion, this without `-XNoFallibleDo` means continuing to call `MonadFail.fail`, which some people may not consider partiality. On the other hand, `-XNoFallibleDo` without this does mean throwing `PatternMatchFail`, which is definitely is considered partiality. Perhaps you meant the opposite of what you said, i.e. having `-XNoFallibleDo` imply this, which avoids that latter case?

hanshoglund:
I meant what I said. My reasoning:  this is about defining a stricter/more reliable subset of the language. I see no reasons to special-case do blocks in the subset.

gridaphobe:
I was also struck by the singling out of incomplete patterns in do-notation. In #319 we seemed to agree that a good motivation for the proposal was resolving the existing inconsistency between do-notation (uses `fail`) and case/lambda/etc (aborts). I think it would be unfortunate to reintroduce an inconsistency here.

But I also think there's a difference between incomplete patterns in general (which feel like an explicit assertion to me) and the other sources of partiality this proposal targets, like incomplete record construction/update and missing type-class methods (which I would happily remove).

Ericson2314:
So my notion of "implicit partiality" for this proposal is quite formal: code that may or may not desugaring into synchronous exception throwing based on definitions elsewhere.

By that definition we have:
1. ```haskell
   let Foo x = ...
   ```
   Is still deemed implicit, because it might be fine if `Foo` is the only data constructor for the type. c.f. 
   ```haskell
   let Foo { x = y, ... } = ...
   ```
   which is idiomatic and definitely not expected to be partial.
2. ```haskell
   do Foo x <- ...
      ...
    ```
   is not implicit partiality (without `-XNoFallibleDo`) because it does not itself sugar to throwing, but rather to a user-defined function call (which may or may not throw).

So in response:

@hanshoglund

>  this is about defining a stricter/more reliable subset of the language. I see no reasons to special-case do blocks in the subset.

I do like that this extension fixes one well-defined, narrow problem. I do fully intended to use both `-XNoFallibleDo` and `-XNoSugaredIncompleteness` wherever I'm allowed to, but I like that they both "do one thing". It's nice to motivate them together, but in terms of documenting their actual behavior, I think they are pretty well separate.

@gridaphobe

> I think it would be unfortunate to reintroduce an inconsistency here.

I don't think I am reintroducing anything? As I understand it, only the "Proposed Change Specification", and I don't mention do notation at all there. The "Effect and Interactions" section calls out an interaction that might be surprising, but nevertheless should completely follow from the "Proposed Change Specification".

I still think that it's an inconsistent design, I just think that's `-XNoFallibleDo`'s problem to solve. That makes the design consistent; this makes the design good; together it is consistent and good.

gridaphobe:
> As I understand it, only the "Proposed Change Specification", and I don't mention do notation at all there. 

I was referring to this bit from the Change Specification:

> Incomplete pattern matching, **except in do-notation where it is always allowed** ...

But I now see that you already addressed the interaction with #319 below in Effects and Interactions:

> The combination of -XNoFallibleDo and -XNoSugaredIncompleteness would make those incomplete patterns errors like all the others under -XNoSugaredIncompleteness.

So that all sounds good now, thanks for clarifying!

gridaphobe:
Separately, I wonder if disallowing all uni-patterns, even for single-constructor types, is going a bit too far. These are complete patterns, but they may become incomplete at a later time. In particular they can only become incomplete when the type definition changes. But if this type is defined in the same module (or package even), then it's impossible for the type to change independently of the pattern, which means we can rest assured that the author will have an opportunity to change the pattern before releasing anything. So in that case, it seems like rejecting uni-patterns only for sums (unless the pattern and type definition occur in different packages) would suffice.

hanshoglund:
> It's nice to motivate them together, but in terms of documenting their actual behavior, I think they are pretty well separate.

Fair enough. I'd hope for both to be accepted (and possibly becoming the default), but the separate proposals make sense.

Ericson2314:
@gridaphobe 

> Incomplete pattern matching, except in do-notation where it is always allowed ...

Ah! I'm embarrassed that I did just misread my own proposal and think that do notation is only covered in "effects and interactions". At least that section clarified things, whew.

> Separately, I wonder if disallowing all uni-patterns

This proposal doesn't aim to do that; I agree that is way too much. I am curious what gave you the notion that it did, so I can fix it.
```haskell
let MyProduct ... = ...
```
is perfect valid to me. I wish to allow it, and allow it no matter where the product is defined, in fact.

Only the uni-patterns which are fallible (i.e. which would be warned `-Wincomplete-uni-patterns`) are prohibited.

------

Rust has annotations on data declarations that say "I reserve the right to add more fields/variants in future versions", and then downstream libraries are compelled to act accordingly. I think that is a great idea, and is the right solution to the "sum in other module" problem you mention. In particular, I think package- and not module-boundaries are the salient ones here, since that is the granularity at which we version interfaces----put differently, a package shouldn't need to be robust against a future version of itself!

gridaphobe:
It wasn't anything in the proposal itself, but your https://github.com/ghc-proposals/ghc-proposals/pull/351#issuecomment-677872436

> So my notion of "implicit partiality" for this proposal is quite formal: **code that may or may not desugaring into synchronous exception throwing based on definitions elsewhere**.

that gave me that impression. It's not entirely unreasonable either to give people the ability to warn/reject such patterns (they're not robust to dependency upgrades), so long as we have a precise enough check to identify cases where the implicit partiality cannot arise. Swift does something similar where they require wildcard patterns when you match against external sums since there's no guarantee the match will still be complete once your dependencies move.

Ericson2314:
Ah right. So in those cases, it's not the whole construct I want to reject for being "implicitly partial". Rather, I just want to disallow just the situations where the construct would desugar synchronous exception throwing. Then just the total situations remain, and the construct is thereby no longer implicitly partial.

TheMatten:
> Rust has annotations on data declarations that say "I reserve the right to add more fields/variants in future versions", and then downstream libraries are compelled to act accordingly.

Maybe `COMPLETE` pragma could be extended with use of wildcard like

```hs
{-# complete _ :: Foo #-}
```

to say that matching on `Foo` has to always contain `_`.

goldfirere:
Is this ready to be submitted to the committee? I still would simplify the name of the extension, but that's up to the author.

Ericson2314:
@goldfirere You are happy then to my answer in https://github.com/ghc-proposals/ghc-proposals/pull/351#issuecomment-666429207 to your question?

I do want to change the name too. `-XIncomplete` I worry isn't really clear, and also flips the direction of the extension (I prefer `-XNo...` to mean we are accepting fewer programs). Most people seem to talk about "partial" rather than "incomplete", so something with that word was my original preference, but then the GHC user manual didn't use that word with any of the existing warnings spooking me that it was perhaps technically inappropriate for some reason.

goldfirere:
Yes, I'm happy about your answer about `-fdefer-` flags.

I don't think `-XIncomplete` changes the polarity of the extension. Users would still have to say `-XNoIncomplete` to disable incompleteness.

I favor talking about incompleteness instead of partiality, as the opposite of partiality is totality, and that implies a termination checker.

Ericson2314:
Ah sorry, I somehow found myself thinking you had proposed `-XComplete` not `-XIncomplete`.

Ericson2314:
Name changed.

> Is this ready to be submitted to the committee?

Ok, I submit it! CC @nomeata.

nomeata:
/remind @yav to check progress in two weeks.

simonpj:
The "discussed at this pull request" link on the [rendered proposal](https://github.com/Ericson2314/ghc-proposals/blob/no-sugared-incompleteness/proposals/0000-no-sugared-incompleteness.rst) is broken.

This proposal is entirely to do with knobs, so it's a UX issue not technical one, and I defer to users (from whom I would love to hear).

I didn't find the specification easy to read.  

* What does "With NoIncomplete those things are disallowed, regardless of warnings." mean?  What does "disallowed" mean?  Program is rejected unconditionally?  I don't think so.

* I don't understand the interaction of the language extension with the new (or indeed old) flags.   Maybe
  * with `-XNoIncomplete`, an incomplete pattern is an error; but you can turn it into a warning with `-fdefer-incompleteness-errors`; and you can suppress that warning with `-Wno-deferred-incomplete-patterns`
  * with `-XIncomplete`, an incomplete pattern is a warning, but you can turn it into an error with `-Werror=deferred-incomplete-patterns`; or you can suppress it with `-Wno-deferred-incomplete-patterns`

  Is that right. Worth laying it out like that.


* What is a "warning category".  Ah... I think it may be the xxx in `-Wxxx`

Ericson2314:
@simonpj 

> The "discussed at this pull request" link on the rendered proposal is broken.

Thanks! I always forget that backlink :/. Fixed.

> This proposal is entirely to do with knobs, so it's a UX issue not technical one, and I defer to users (from whom I would love to hear).

True, the proposal does not involve adding any interesting new functionality to GHC, But just to be clear I think it does pose some more interesting questions for Haskell itself.

This language extension, like any other, presents the question of whether it might be part of a future language report. I would indeed like the report to say "class declarations cannot omit required methods", "record construction must include all fields", etc., and so the new `-fdefered-*` flags are an implementation-specific extension. Perhaps that's a bit more interesting than knobs.

Of course, you want are still free to "sit this one out", just want anyone to think I "slipped it by the committee".

> What is a "warning category". Ah... I think it may be the xxx in `-Wxxx`

Yes. Happy to call that something else, I don't know what the conventional phrasing is.

--------

As to the wording issues, agreed it wasn't yet clear enough. Perhaps it is better now? I want to introduce the language extension prior to introducing the `-fdefer-incompleteness-errors` mechanism, since the former could one day become part of Haskell proper while the latter is an implementation extension that depends on the format. That meant a different organization than what you proposed, but I hope it's not fatally less clear because of that?

Also, inspired by your alternative phrasing, I changed my mind on one of the examples. `deferred-incomplete-patterns` only makes sense when there is something that is being deferred, so only with `-XNoIncomplete -fdefer-incompleteness-errors`. But the existing warning categories are free to either matter or not matter with `-XNoIncomplete`. I had them have an effect before, but now choose to have them not have an effect. This makes a nice symmetry:

 -  `-XIncomplete`: only old `-Wincomplete-* warnings matter
 - `-XNoIncomplete`: only new `-Wdeferred-incomplete-* warnings matter

I think that's better.

adamgundry:
Sorry for not spotting this proposal before it was submitted.

I recently opened https://gitlab.haskell.org/ghc/ghc/-/issues/18650 about warnings for incomplete field selection and update (especially in the presence of `RecordDotSyntax` or similar). Note that we do not currently have a warning at use sites for incomplete field selectors, although it would be a good idea.

> We cannot get rid of accessor functions defined in other modules, but we can at least limit record syntax in the current module

This sounds a bit like you are proposing not to warn about occurrences of imported partial accessors. Is that the right interpretation? Why shouldn't the warning apply across modules? We'll have to serialize an "is this selector incomplete" bit in interface files, but that should be easy.

Those details aside, I'm not really convinced that a new language extension pays its way here. Why not simply add a new option group (like `-Wcompat`) for enabling all the incompleteness warnings at once? That is, we could have a `-Wincompleteness` option that enables `-Wincomplete-uni-patterns`, `-Wincomplete-record-updates`, `-Wpartial-fields`, `-Wmissing-fields`, `-Wmissing-methods`... It would be easy to set `-Werror=incompleteness` and get much the same effect as `-XNoIncomplete`, but without complicated interactions between enabled language extensions and warning flags, nor a lengthy process of renaming and deprecating existing warnings.

Ericson2314:
Oh https://gitlab.haskell.org/ghc/ghc/-/issues/18650 looks really good!

> We'll have to serialize an "is this selector incomplete" bit in interface files, but that should be easy.

We could, but I sort of feel icky adding special cases the accessors which are supposed to "just" be regular functions. I kinda feel between all the other ways we have to access fields now (regular patterns, record dot, puns, overloaded labels, ...) that the community should just avoid accessors altogether via`-XNoFieldSelectors` to solve this.

> Those details aside, I'm not really convinced that a new language extension pays its way here. ....a lengthy process of renaming and deprecating existing warnings.

You're right it's more work / head-spinning, but it's worth it to me. My argument is unabashedly ideological.

As you say 

> These warnings are intended to support a style of Haskell programming where uses of any partial functions are discouraged/minimized. This is a common (if not universal) approach.

I believe in that approach and agree it's is predominant, and I think Haskell itself should help promote it.

Right now we say this incomplete syntax is part of the language, but then encorage people to use `-Werror` to prevent its use.
I think that's a confusing and muddled message.

With is proposal, we're on track to just saying this incomplete syntax is not part of the language, with  `-fdefer-*` as GHC-specific mechanism to extend the domain of the operational semantics to include it. I think that's a stronger and clearer message.

goldfirere:
I'm not sure I'm fully sold on the `-fdefer-xyz` plan herein, but I agree with the spirit of the proposal (and do not disagree with the letter). Right now, some basic mistakes are not even flagged with an on-by-default warning (like incomplete patterns). If one of Haskell's strengths is in its static analysis (via its type system), then we should capitalize on this by catching more errors. This proposal allows us to do so quite nicely.

One use case: If I were still teaching intro Haskell, I would have all my students enable `-XNoIncomplete` from day 1.

simonpj:
I'm still struggling  a bit with the Proposed Change Specification.

* "Incomplete pattern matching, excluding...".  This rather presupposes that we are sure
  what "incomplete pattern matching" is.  Is this legal, with `-XComplete`?
  ```
  f x | x > 0 = e1
      | x <= 0 = e2
  ```
  Or is it rejected with an error?  In other words, on which side do we err?  Pattern synonyms and COMPLPETE pragmas complicate the picture.

  I think it would help to list what this bullet includes.  I think it is
  * Function definitions e.g.
    ```
    f [] = e1
    f (x:xs) = e2
n    ```
  * Case-expression alternatives e.g.
    case y of
       [] -> e1
       (x:xs) -> e2
  * Pattern bindings e.g.
    ```
    let (x,y) = e in ...
    ```
  Anything else?

* "Incomplete record construction, field selection, or update.".  Let's just have three separate bullets, and be fully explicit:
  * Incomplete record construction, such as `K { x = e }` where constructor `K` has a field other than `x`.
  * Incomplete record update, such as `e { x = e2 }` where the data type being updated has at least one constructor that does not have a field `x`.
  * Incomplete record selection: any use of a record selector that can fail, such as a use of `x`
    for a data type that has at least one constructor that does not have a field `x`.

  This last bullet answers Adam's question:

* "Missing items.."  Better "Missing class methods or associated types in a type class instance declaration, when the class declaration does not specify a default method or default associated type respectively.".

  If there is a MINIMAL pragma, is it an error (or still a warning)(

* What if the *defining* module for a data type has `-XNoIncomplete` but the
  module that (say) uses the incomplete construction, update, or selection
  has `-XIncomplete`.   Then the warnings should fire, right?  Say so.

* I keep getting stuck in double-negatives. "Incomplete" is already a
  negative, while "NoIncomplete" is a double negative.  Maybe `IncompletePatterns`,
  and `CompletePatterns` as a synonym for `NoIncompletePatterns`?

* I am troubled by having both `-Wincomplete-patterns` and `-Wno-deferred-incomplete-patterns`.
  I think the former is relevant only with `Incomplete` and the latter only with `NoIncomplete`.
  The plethora of combinations, some of them vacuous, seems unsettling.

* Is the star notation e.g. `Wno-deferred-*` currently in GHC.  Or is it part of the extension you propose? Or is it meta-notation in the specification?

I'm sorry to be so picky, but precision is all.  Rather than answering me blow by blow, you might just want to update the Spec and say "look now".

Ericson2314:
@simonpj I'm happy write a more complete specification. That said, to me the true "living spec" is

> with `-XNoComplete` the compiler must never emit synchronous exception throws that weren't explicit in the surface syntax, except for `NonTermination`.

Everything else follows from that.

It absolute is bad to defining a language by 1) implementation-specific-behavior 2) in negative terms, and this short version both crimes, but it does have the advantage of being concise, and specifying the behavior of `-XNoIncomplete` going forward with interactions we cannot forsee. GHC's current exhaustiveness checking is also quite complex and constantly evolving, and I know I would fail to specify it all here.

I'm genuinely confused on how to navigate writing a spec that's robust on one hand, and direct one on the other.

simonpj:
> to me the true "living spec" is...

Great. Write that down too!  It illuminates the specifics.

> I'm genuinely confused on how to navigate writing a spec that's robust on one hand, and direct one on the other.

Language specification is not easy.  But the only thing worse than wading through the swamp of writing a spec is trying to implement something to an ill-defined spec!  So I believe that it is effort well invested.

> GHC's current exhaustiveness checking is also quite complex and constantly evolving, and I know I would fail to specify it all here.

Actually it's simple: GHC's exhaustiveness checking is, and will always be, imprecise.  The x>3, x<=3 example is a case in point.  So we know for sure there will be cases when the checker errs.  The point here is not to specify in which cases it may err (that is hard, and changing) but rather to specify what happens if it does err.  Is the program accepted, or rejected?

int-index:
> The point here is not to specify in which cases it may err (that is hard, and changing) but rather to specify what happens if it does err. Is the program accepted, or rejected?

Assuming I understand the intend behind the proposal, the program will be rejected:

```
f x | x > 0 = e1
    | x <= 0 = e2
```

The fix is to add a redundant clause to calm down the exhaustiveness checker:

```
f x | x > 0 = e1
    | x <= 0 = e2
f _ = error "f: impossible case"
```

Typically, GHC would insert a similar clause automatically, but with `-XNoIncomplete`, this sort of exception-throwing is never sneakily added behind the user's back.

But I think it's important that with `-XNoIncomplete`, GHC would suggest this workaround in its error message.

JakobBruenker:
> The fix is to add a redundant clause to calm down the exhaustiveness checker:

can't these cases usually (always?) be fixed simply by using `otherwise`? As in

```Haskell
f x | x > 0 = e1
    | otherwise = e2
```

Some might prefer the `error` fix since it has a nice symmetry in the actual cases, but for me, I'm happier using -XNoIncomplete (and more sympathetic towards this proposal) knowing that `error` doesn't (usually?) have to be thrown in, than I would be otherwise.

Ericson2314:
Yeah I would prefer @JakobBruenker's fix too. But the warning could describe the unhandled-trace ("can match first pattern but none of the guards") rather than biasing one approach over the other.

ndmitchell:
A few questions:

* Currently some form of incompleteness warnings are handled in the desugarer where the pattern match checker runs. Does this proposal aim to produce such errors before that phase? I can't think of any other language pragmas that have an impact that late in the compilation pipeline (are there any?). Does defining a language pragma at such a level have practical consequences?
* Pattern match checking is exponential. The GHC pattern match checker uses a "fuel" style limit to avoid taking an indefinite amount of time to complete. Does this proposal require an infinite fuel? What if that fuel runs out? Is the fuel part of the spec?
* The `otherwise` function is one we happen to know always returns true. Is it special? Is `Prelude.otherwise` special? Or can I define `myOtherwise` in my alternative Prelude? Such details are handled by GHC's current pattern match checker, which seems to be rewritten from scratch every so often. Does this proposal imply the current GHC pattern match checker? Some abstract pattern match checker with specific properties?
* How do things like view patterns interact? Guards? Implicitly it seems to be following the GHC pattern match checker.

Overall, I'm a little concerned that this language pragma is encoding a lot of GHC specific heuristics that change regularly. For a warning, that's absolutely fine - and I think an overhaul of warnings in this area could be valuable. You correctly observe that people want to say "no incompleteness" and get concrete guarantees on runtime behaviour, rather than wiggling a bunch of separate flags to get only guarantees on specific source code forms. But the above proposal feels more like a diff to GHC, rather than a diff to the Haskell report, and I'm not sure how easy it would be to implement in a different compiler. Perhaps try and concretely specify what "incomplete" does at the level of Haskell syntax - what does the idealised checker required to adhere to the spec look like? What properties must it satisfy?

Ericson2314:
@ndmitchell 

Check out my comment https://github.com/ghc-proposals/ghc-proposals/pull/351#issuecomment-692330133, which will be incorerpated in the next draft. I think that's basically me voicing the same concerns!

I totally agree that if we try to codify what must be allowed, we'll be stuck trying to set in stone those "GHC-specific heuristics that change regularly", but if we make this about what the implemention can *not* do --- producing synchronous exception throws that the user didn't write --- the situation becomes much a clearer: Another implementation of Haskell doesn't need to do all the same tricks GHC does, and is free to permit fewer programs, so long as those throws are not produced.

I will look at the report, and see if I can some up with a very basic exhaustiveness check (e.g. ignoring GADTs, treating all pattern synonyms as fallible, etc.) to specify as a bare minimum level of sophistication (i.e. programs passing these conservative criteria must be allowed), while allowing fancier analyses that except more programs so long as they are still sound.

ndmitchell:
> Another implementation of Haskell doesn't need to do all the same tricks GHC does

But it does need to match those tricks to permit the same programs GHC does. I think I'm making two assumptions, either or both of which might not be commonly agreed upon:

1. Language pragmas are things that could be standardised into the Haskell report one day, rather than some which might inherently be "GHC" language pragmas rather than Haskell language pragmas. 
2. The Haskell report should define exactly which programs are/aren't accepted by the standard, rather than a conservative lower bound.

Note that for warnings, we already accept they are conservative in unpredictable ways, and aren't appropriate for the language standard, so I set far less stringent conditions upon them.

Ericson2314:
@ndmitchell I agree with that, and I definitely do want `-XNoImplicit` to become part of the language in a future iteration in report---in fact it might be the extension I most want to see ratified.

I think implementations that can except more programs aren't necessarily in violation of the spec, but those implementations ought to have a "strict mode" where they except no mroe and no less than the spec, so it's good phrase things into those "invariant" terms.

Ericson2314:
So Haskell 2010 in "3.17.3 Formal Semantics of Pattern Matching" basically assumes all pattern matching is infallible---omitting the `error "No match"` is left as one of many implementation-specific patterns. This fine as the simplest way to present the operational semantics, but offers no guidance whatsover on what `-XNoIncomplete` should be standardized as, ugh.

 I guess I'll see what OCaml or SML do? I'm a bit wary of trying to codify what it is no practical concern to implementing this in GHC --- GHC already fails to offer such a strict mode where nothing not allowed in the standard is acceptd.

Ericson2314:
The revised (1997) definition of standard ML (https://smlfamily.github.io/sml97-defn.pdf), talks about "irredundant" and "exhaustive" patterns, but only very informally.

Ericson2314:
https://caml.inria.fr/pub/docs/manual-ocaml/expr.html just talks about when `Match_failure` is raised and doesn't make any mention of completeness checking.

simonpj:
> offers no guidance whatsover on what -XNoIncomplete should be standardized

We know that the compiler cannot possibly always get it right.  Consider
  ```
  tricky x | alwaysTrue x = e
  ```
  where `alwaysTrue` is a complicated function that, in fact, always returns True.  No way can any compiler guarantee to work out whether this pattern match is complete or not.

So the questions are:
  
1.  With `-XCompleteMatching` (I struggle with the double negative of `-XNoIncomplete`), should `tricky` be (a) rejected or (b) accepted?  More generally, in cases where the compiler can't decide if matching is complete or incomplete.
 
2.  In some cases the compiler can't reasonably decide; but in others, it definitely can, such as
```
length [] = 0
length (x:xs) = 1 + length xs
```
The exact boundary between "can" and "can't" is quite hard to specify, but we might want to specify a large class of cases where the compiler can guarantee to decide.

monoidal:
> The `otherwise` function is one we happen to know always returns true. Is it special? Is `Prelude.otherwise` special?

In the current GHC, `otherwise` and `True` are special - user synonyms won't work.

I share the worry that without a clear definition, the set of programs accepted with `-XIncomplete` will be in flux - especially if the plan is to make it the default later. For example, GHC 9.0 will check for long-range pattern matching: `case x of False -> ...` will not be flagged if `x` was matched against `True` in the surrounding context.

yav:
Overall, I am sympathetic with this proposal, and am likely to use such a feature.   However, it seems like the proposal is not quite in its final form, and still needs some work, especially when it comes to specifying pattern matching completeness.  It would also be nice to be more explicit about exactly which language constructs will be affected by this, and how (e.g., are pattern bindings affects?  what are the interactions with lazy pattern matching?).

When making choices, my suggestion would be to favor things that are likely to help programmers rather then aim for an abstract "right way" of doing things.   In particular, I don't think that partial pattern bindings should be covered by this, as if I write `let Just x = e1 in e2` I probably have a reason to believe that this won't crash, and I am unlikely to do this by accident.   Otoh, missing an alternative in a case is pretty easy to do, so I want an error.

For the actual algorithm, my suggestion would be to make a conservative algorithm that only looks at patterns, and never analyzes expressions with the exception of `True` and `otherwise`.  Thus, view patterns and pattern guards would fail if their pattern may fail, and an ordinary guard may always fail, unless the expression is `True` or `otherwise`.

I find the trickiest case to be "competences based on type" which arises with GADTs and I am not sure what's the right thing to do there.

adamgundry:
This discussion is making it increasingly clear to me why I would prefer a new warning option group `-Wincompleteness` over `-XNoIncomplete`: the definition of completeness is too dependent on specific analyses that change over time, which is fine for warnings whereas language extensions should be more stable. Yes, we could specify a simple analysis and make that part of the language specification. But IMHO it seems more honest to regard incompleteness checking as a (fallible) warning mechanism rather than a part of the language itself.

What if in the future someone was to implement a decent termination checker for (a subset of) Haskell programs? Then it would make perfect sense to have `-Wnontermination` with the understanding that it would sometimes reject perfectly good programs, and maybe even have `-Wincompleteness` imply `-Wnontermination` if the checker was good enough. But we probably wouldn't want to change the (extended) language itself to say that as of GHC 42.0, potentially-nonterminating programs are erroneous albeit with a deferrable error.

yav:
@adamgundry I kind of agree wrt to completeness of pattern matching.  However, the record based incompleteness and missing class methods are not something I'd every purposefully use, so I'd be quite happy if they were errors rather than just warnings.

hanshoglund:
> record based incompleteness and missing class methods are not something I'd every purposefully use, so I'd be quite happy if they were errors rather than just warnings.

Strongly agree!

> completeness of pattern matching

This is clearly trickier. One option might be to exclude pattern matching from the proposed extension. 

Another one might be to (formally) specify a very conservative subset of patterns treated as complete by this extension. This would include things like constructor completeness but not "tricky" cases (e.g. anything involving literals, guards or GADTs). To write such programs with the extension enabled the user would _always_ be forced to write out a catch-all clause. The user of course may just invoke `error` in the clause, but it is much better to make this choice explicitly!

Here are some  I'd expect to be rejected, and what one would have to write instead.

```haskell
-- No
{ True -> 1 }
-- Yes
{ True -> 1; _ -> 2 }

-- No
case 22 of { x | odd x -> 1 }



```

---

>  if I write let Just x = e1 in e2 I probably have a reason to believe that this won't crash

That may be true in the case of `Maybe`. But consider:

```haskell
-- Version 1 of A.hs
data Foo = Foo Int

-- Version 1 of B.hs
{-# LANGUAGE NoIncomplete #-}
f :: Foo -> Int
f x = let Foo a = x in a 

-- Version 2 of A.hs
data Foo = Foo Int | Bar
```

As `A` is updated from v1 to v2, I would expect `B` to no longer compile. Thus permitting the `let A x = ...` form for single-constructor types only can catch bugs, and feels like default for the stricter language we're discussing here.

hanshoglund:
Under the conservative scheme, I'd expect:

```haskell
-- Reject
{ True -> ... }
-- Accept
{ True -> ...; _ -> ... }
-- Accept
{ True -> ...; x -> ... } -- for any identifier x
-- Accept
{ True -> ...; False -> ... }

-- Reject
{ x | odd x -> ... }
-- Reject
{ x | odd x -> ... | even x -> ... }
-- Accept
{ x | odd x -> ... | even x -> ...; _ -> ... }

-- Reject
{ x | p x -> ... } -- forall p
-- Accept
{ x | p x -> ...; _ ->  } -- forall p

-- Reject
{ 2 -> ... }
-- Accept
{ n -> ... }
-- Accept
{ 1 -> ...; 2 -> ...; n ->  }

```

proposals/0000-no-sugared-incompleteness.rst
```diff
@@ -0,0 +1,211 @@
...
+   To accomplish this, we will mainly rely on existing analysis.
+   Any program that would emit warnings from the following warning categories is prohibited:
+
+   - ``incomplete-patterns``
+   - ``incomplete-uni-patterns``
+   - ``incomplete-record-updates``
+   - ``missing-fields``
+   - ``missing-methods``
+
+   Additionally, with ``-XFieldSelectors`` (also on by default), any program that would emit warnings with ``-Wpartial-fields`` is also prohibited.
```
adamgundry, line 67: Or we could implement `-Wincomplete-record-selectors` and add it to the list above, then `-Wpartial-fields` would be safe? After all, `-Wpartial-fields` doesn't itself introduce incompleteness, so it would be a shame to forbid it if we can be more precise.

proposals/0000-no-sugared-incompleteness.rst
```diff
@@ -0,0 +1,211 @@
...
+Possible Future Work
+--------------------
+
+The drawbacks are serious enough that I feel compelled to sketch some future plans that would mitigate them.
+
+The chief drawback with pattern matching today as it relates to completeness checking is that the Haskell implementation cannot tell whether a user skipped a pattern because they think it's impossible or because they forgot.
+If we had something like Agda's "absurd patterns", however, the user could use those to use those cover the impossible the cases, making clear that anything else the user really did forget.
+Firstly, this should help make code more self-documenting and error/warning messages better.
+But especially relevant to the problem at hand, this allows simpler Haskell implementations that aren't so sophisticated that they can derive proofs of pattern impossibility very well on their own, but can verify user-written arguments.
+This allows extension conjunctions like ``NoIncomplete`` and ``GADTs`` to be specified in ways that are less onerous on the Haskell implementation.
```
adamgundry, line 169: While Agda-style absurd patterns would be nice, I don't think they actually help much here, because you have to specify when they can be used. An absurd pattern on `x` is essentially equivalent to requiring that `case x of {}` covers all cases, so checking it doesn't make the implementation simpler.

In Agda, one has to use an absurd pattern when an impossible pattern is nested somewhere inside a larger pattern, but in Haskell such cases are normally trivially accessible because of laziness.

adamgundry (commented):
> @adamgundry I kind of agree wrt to completeness of pattern matching. However, the record based incompleteness and missing class methods are not something I'd every purposefully use, so I'd be quite happy if they were errors rather than just warnings.

@yav that's true, I guess it would be a reasonable design to say that those sources of incompleteness are forbidden by default (deferrable via a flag), whereas general pattern matching completeness warnings are still warnings. It's certainly a lot easier to specify. But IMHO from where we are now it is better to just default to `-Werror=missing-methods` etc. and leave language extensions out of it. :wink:

Ericson2314:
OK I've pushed some updates that I hope help address recent conversation.

First, a disclaimer that I still need to read over the rendered view, where I always catch more mistakes than reading the markdown :flushed:.

Highlights:

 - I precisely specify `NoIncomplete` *for GHC*, using existing analyses and a 1-2 new novel restrictions taken from https://gitlab.haskell.org/ghc/ghc/-/issues/18650.

 - I make not that's not specifying `NoIncomplete` as diff on the Haskell report, sketch what would be done were `NoIncomplete` in fact to be submitted for a future Haskell report, but for now explicitly punt on that.

 - I add the drawbacks people have discussed

 - I sketch some future work which I hope will stop those drawbacks from biting.

proposals/0000-no-sugared-incompleteness.rst
Ericson2314: @adamgundry I totally agree, I just felt it would be cheating specify with reference to an open issue. I would love to tighten this spec up with your new warnings!

proposals/0000-no-sugared-incompleteness.rst
Ericson2314: I guess I should rephrase this to make it clear I don't mean exactly like Agda absurd patterns, but something along those lines.

> In Agda, one has to use an absurd pattern when an impossible pattern is nested somewhere inside a larger pattern, but in Haskell such cases are normally trivially accessible because of laziness.

Yes when it's just a nested `Void`-like thing, Haskell unintentionally makes one "prove it" which is great. But in the case GADTs type constraint stuff, Haskell is magical again. We'll soon (:crossed_fingers:) have type applications in patterns, I think I would want to put something in there (think nested absurd patterns for *irrelevant* fields. which aren't lazy) to get closer to an explicit proof for deriving the type inequalities.

simonpj:
> First, a disclaimer that I still need to read over the rendered view

Perhaps just give us a ping when you are ready for a new review?  Thanks!

yav:
To keep things orderly I'll mark this with "requires reversion" for now, as it seems that the proposal is still being revised, and the discussion is in progress.

@Ericson2314 please let us know when the dust has settle again, but I'd encourage others to continue the discussion *before* submitting to the committee, so that hopefully the next round will be more stable.

Ericson2314:
OK tweaked a words and expanded upon my second reply to @adamgundry in the "future work" section. This is ready more review.

proposals/0000-no-incomplete.rst
```diff
@@ -0,0 +1,268 @@
...
+do something like:
+
+::
+
+  f :: G Bool -> ...
+  f GBool -> ...
+  f (ABSURD @ GInt)
+
+Firstly, this should help make code more self-documenting and should allow for better error/warning messages.
+But more relevant to the problem at hand, this could possibly allow simpler Haskell implementations that aren't so sophisticated that they can derive proofs of pattern impossibility very well on their own, but can verify user-written arguments.
```
adamgundry, line 221: Sorry, I still don't get this. How does `ABSURD` help simplify the coverage checking problem?

Presumably
```hs
f (ABSURD @ GInt)
```
is (imaginary syntax for) a programmer assertion that the `GInt` case is impossible. Is the idea that this assertion will be unchecked, and itself constitutes the programmer's explicitly specified reason for the program failing at runtime, should it in fact turn out to be invalid? Or if it is checked, why is that easier than treating the omission of the constructor as such an assertion?

I can imagine extending Core/System FC with a language of checkable "apartness evidence" that would turn the coverage checker into an elaborator. That would allow a simpler lint algorithm to verify the absence of incomplete matches in Core. But I highly doubt the evidence terms would be anything a programmer would want to read or write, and that doesn't seem to be what you are proposing in the short term (though it comes up in the following section)?

proposals/0000-no-incomplete.rst
```diff
@@ -0,0 +1,268 @@
...
+Deciding whether e.g. pattern matching is exhaustive feels to me like one of those essential analyses.
+I am asking that we switch our vantage point and instead of considering programs with implicit partiality well-formed programs that are iffy, consider them ill-formed programs that we can permissively accept anyways.
+
+So, what's the practical upshot of this lawyering?
+And what mechanism might we use rather than ``-Werror=...``?
+There is another mechanism for controlling run-time errors which I think is a perfect model for how these programs should be dealt with: ``-fdefer-type-errors``.
+As I mentioned above, nobody would propose type checking is some extraneous analysis a Haskell implementation need not do by default.
+The "defer" in the name makes clear we are taking what is normally a compile-time error and deferring it to run-time behind additional branching.
+Whether for quick slap-dash programming, or arguing with proponents of untyped languages, ``-fdefer-type-errors`` allows GHC to both express its opinion and yet also not straight-jacket the user into any one policy.
```
adamgundry, line 42: Without type-checking, a Haskell program doesn't really have any meaning (more formally, it doesn't have a sensible denotational or operational semantics). In the absence of types an expression like `mempty` is simply meaningless;  `-fdefer-type-errors` throws in runtime errors in an implementation-defined way to patch it up. This is very different to (gradually) dynamically typed languages, where untyped expressions have a more-or-less sensible semantics.

In contrast, even with the coverage checker disabled (or buggy), a Haskell program has perfectly well-defined semantics: it throws a runtime error if an uncovered pattern occurs during evaluation. Thus I think that for Haskell, coverage checking really is an "extra" analysis, whereas type-checking is an "essential" one.

One can certainly imagine a language for which coverage checking really is necessary as part of the language specification, but Haskell isn't such a language. Some dependently-typed languages even require totality, which allows optimizing away runtime computations because there is no requirement to check proofs at runtime, but then they have undefined semantics if a non-total program slips past the compiler. 

Sorry to harp on. :wink: I think we are in agreement in preferring a programming style where the absence of incompleteness is obvious to the compiler, with the compiler warning us if it can't understand why a program is complete. I'm just not convinced this motivates a new language extension and a lot of churn around warning names.

Ericson2314: > In contrast, even with the coverage checker disabled (or buggy), a Haskell program has perfectly well-defined semantics.

We *choose* to give it those semantics. We could also choose not too. I agree the math doesn't force us to exclude such programs, but it doesn't need us to include them either. It's a matter of opinion.

I want there to be a generation of Haskell programs who've never not had incomplete patterns be an error (by default), and who would be shocked to learn the Haskell ever worked differently. If you have an alternative way to get there, I'm all ears.

proposals/0000-no-incomplete.rst
Ericson2314: > Or if it is checked, why is that easier than treating the omission of the constructor as such an assertion?

It is meant to be checked. `(ABSURD @ GInt)` is a simple example, but imagine if this was buried between deep within a more complicated pattern. The compiler can just follow the given pattern, rather than try to come up with it.

I will admit perhaps it's more saving the compiler work doing in exhaustive search than allowing it to be less sophisticated, as this doesn't show how to combine the equalities to get the absurdity. Still perhaps it could be augmented with something along those lines (type applications in patterns helps a bit) that gets us closer.

goldfirere:
@Ericson2314 

> I want there to be a generation of Haskell programs who've never not had incomplete patterns be an error (by default), and who would be shocked to learn the Haskell ever worked differently.

I agree. :)

Ericson2314:
@simonpj 

> Perhaps just give us a ping when you are ready for a new review? Thanks!

Oh, oops. I said it was ready but forgot to ping.

simonpj:
> Oh, oops. I said it was ready but forgot to ping.

Thanks for the ping. It is clearer now.

>  More important than the specific specification is the guiding principle behind that specification. It is this:..

Good to have a guiding principle!  But then please follow up the principle with the specific specification!  We don't want to ask the reader to think about every source of synchronous exceptions. (E.g. divide-by-zero is not covered by this proposal.)  Rather you have some very specific exceptions in mind. Can you enumerate them in the spec?

Let me try to help.  I think you intend to cover precisely and only these cases:

* A value is not matched by a pattern or patterns in:
  * A function definition, e.g.
    ```
    f (Just s) = rhs
    -- But no f Nothing case
    ```
  * A case expression, e.g.
    ```
    case expr of
       Just x -> rhs
       --- But no Nothing case
  * A pattern binding e.g.
    ```
    let Just x = blah in body
    ```

* A record update, when the record value to be updated does not have all the updated fields, e.g.
  ```
  data T = T1 { x:: Int } | T2
  f r = r { x=3 }   -- Fails if T2 is passed to f
 
* Misssing class methods in an instance declaration, with no default method  e.g.
  ```
  class C a where
    op1 :: a -> a
    op2 :: a -> [a]

  instance C Int where
    op1 = rhs
    -- No defn for op2, and no default method in the class decl
  ```

* With `-XFieldSelectors` (which is on by default), the selector for any field that
  does not occur in all constructors.  Example: data type T above.

I think that enumerating the cases explicitly would make the proposal much clearer.

> To accomplish this, we will mainly rely on existing analyses.

It's confusing to *specify* a new feature by reference to existing analyses.  I think the
specification is this:
* With `-XComplete`, if the program compiles without error, then none of the above
  exceptions can possibly occur at runtime.

Is that what you intend? Or do you intend that:
* With `-XComplete`, if the program is rejected for one of the above reasons,
  then there is *definitely* a completeness bug: a value for which the match will fail.

(One has false positives, the other has false negatives.)
A consequence of the former (and we should say this in the spec) the compiler is required to err on
the conservative side. I think this is your intent, but the draft is not clear.  For example
```
f x | x > 0  = rhs1
    | x <= 0 = rhs2
```
Few compilers will be able to *guarantee* that this equation is
complete (especially since `f :: (Ord a, Num a) => a -> blah`, so the
behaviour of comparison and numeric literal depends on the type), so
most compilers will reject the program.  In general, no compiler can perfectly
separate incomplete from complete programs.

This would make this the *only* feature in GHC for which we know there is no
possible (reasonable) specification for exactly which programs are accepted
and which are rejected.  There is an issue of principle here.

-----------

The "proposed change spec" seems to include quotations from comments on earlier drafts.
That makes it hard to read as a spec. Perhaps just focus on the spec?


-------------------

As an implementation plan (not part of the spec) just making `-XComplete` equivalent to
`-Werror=incomplete-patterns` `-Werror=incomplete-uni-patterns` etc?

goldfirere:
I can think of a few more concrete cases that ought to be listed:

* Record construction that does not include a declared field.

* Patterns (first of Simon's bullets) in `-XMultiWayIf`.

I presume this proposal does *not* affect type family completeness.

cgibbard:
> ```
> f x | x > 0  = rhs1
>     | x <= 0 = rhs2
> ```

Thinking about this example from Simon a bit, I think it's actually okay if we steer in the direction of complaining about code of this form. Apart from those with a fall-through pattern match, any final guard which is something other than otherwise/True is either expressing redundant work that could be avoided, or is responsible for the pattern match being incomplete. That is, in the case of this example, there should be no need to test that x <= 0, unless the function is actually incomplete (because some Ord instance fails to be total). We might hope that <= is defined in a way that allows CSE to occur here, but I'm not sure I'd really expect or want to rely on that.

It doesn't seem too onerous to be required to write:

```
f x | x > 0     = rhs1
    | otherwise = rhs2
```

instead, or more generally, to replace whatever the last guard was with `otherwise`. If you're being extra defensive and want a runtime error to occur in the case that none of your expected-to-be-complete conditions holds (perhaps due to an invariant being broken by a change elsewhere in your program), it would still be possible to write an `otherwise` branch along with an explicit error. The redundancy checker may warn about it though.

hanshoglund:
@cgibbard I was about to make the same proposal.

More succinctly I guess one could say:

> A guard without a fall-through pattern match is considered incomplete. A fall-through pattern is either `True` or `Prelude.otherwise`.

For comparison this is [exactly what Purescript does](https://github.com/purescript/documentation/blob/master/language/Pattern-Matching.md#guards).

cgibbard:
@hanshoglund Of course, there's one other possibility we should probably allow: an incomplete set of guards is okay so long as a subsequent pattern in the same pattern match is at least as general as the pattern that the guards were applied to.

For example:

```
foo xs | someCondition xs = ... 
foo [] = ...
foo (y:ys) = ...
```

is still complete.

hanshoglund:
@cgibbard I think this is compatible with the brief definition.

Incomplete:

```haskell
foo xs | someCondition xs = ... 
```

Complete:

```haskell
foo [] = ...
foo (y:ys) = ...
```

Complete:

```haskell
foo xs | someCondition xs = ... 
foo [] = ...
foo (y:ys) = ...
```

cgibbard:
The brief definition defined a fall-through pattern to either be "True" or "Prelude.otherwise" though. But I think we basically agree on what should be accepted, anyway. :)

Ericson2314:
I agree with these last few posts, and also thanks @hanshoglund for pointing out the prior art in Purescript. It's always nice to note when others reached the same conclusion.

goldfirere:
What if we have

```hs
f x
  | some_condition = ...
  | not some_condition = ...
```

? By default, this would be flagged (rightly) as incomplete. Then the user would add a `| otherwise = ...` clause. But what if the compiler one day gets smarter and learns about the semantics of `not`? This isn't inconceivable. Then the `| otherwise = ...` clause would become redundant. Even worse, what if the new smarts weren't terribly reliable?

What I'm getting at is that there should be a way to semantically label a clause as erroring. An erroring redundant clause would perhaps trigger a different warning (off by default) than a non-erroring redundant clause.

Or do we say that it would be Against The Rules for GHC to ever learn about how `not` behaves?

hanshoglund:
My thinking is that warnings are out of scope for this proposal. The proposal is about making a clearly defined set of programs previously accepted to be rejected. Error messages and warnings can always be improved and the fact that we're "downgrading" some programs from "accepted with a warning" to "rejected" is somewhat accidental.

Concretely, under `NoIncomplete` this should always be rejected.

```haskell
f x
  | p = ...
  | not p = ...
```

This should always be accepted (with a warning or not, up to the implementation):

```haskell
f x
  | p = ...
  | not p = ...
  | otherwise = ...
```

Ericson2314:
@goldfirere

> Or do we say that it would be Against The Rules for GHC to ever learn about how not behaves?

I don't think that.

> Even worse, what if the new smarts weren't terribly reliable?

I think we should only have the new smarts the influence the "hard" exhaustiveness checker once we're convinced they are reliable. I don't think so restricting ourselves will be too hard?

> What I'm getting at is that there should be a way to semantically label a clause as erroring. An erroring redundant clause would perhaps trigger a different warning (off by default) than a non-erroring redundant clause.

Most of this time, I was thinking people might write
```haskell
f x
  | some_condition = ...
  | complementary_condition = ...
#if __GLASGOW_HASKELL__ >= 9999
  | otherwise = error "shouldn't be possible to get here"
#endif 
```
to try to leverage compiler analyses as hard as possible without sacrificing comparability.

Maybe we would have some fancier annotations and warnings, but I agree with @hanshoglund that's basically out of scope. In particular I want the "no footguns" worldview of `-XNoIncomplete` to be a simple baseline. The "user vs proof assistant totality checking games" interactions I rather only get into once everyone is on board with the "no footguns" part. :).

------

All that aside, I hope the real savior is not more compiler magic but Dependent Haskell, with things like [Agda's `Trichotomy`](https://github.com/agda/agda-stdlib/blob/d67e53e09c577695b1100d17d22a27fbe0396c2a/src/Relation/Binary/Definitions.agda#L102) data type allowing us to safely avoid the lack of clarity of `True`, `otherwise`.

cgibbard:
If a user, faced with a complaint from GHC that their guards are `b` and `not b` and that they should make use of an `otherwise` branch to avoid the potential for incompleteness, then adds a third branch to their function for the `otherwise` case, at that point, we'll have GHC send the error message not to the user, but to the user's employer.

(joking of course)

In the case where you literally have guards `b` and `not b`, it's understandably a little bit frustrating that the compiler might complain at you (apart from the fact that when `b` is not a constant expression this is code that will be unfortunate if CSE doesn't catch it for whatever reason), but if we imagine the general case where the compiler is doing a bunch of boolean algebra to let you get away with writing guards considered complete that have complicated combinations of logical expressions, I'm not sure there's really a great argument for why we should allow that rather than insist that the user avoid the cryptomorphically-complete expression. Also, the guards `b` and `otherwise` are pretty comparable in readability to `b` and `not b`.

If a future GHC discovers through boolean algebra that your `otherwise` is redundant, then presumably it was always redundant, and you can just remove that branch, and replace the second last guard with `otherwise`.

goldfirere:
> If a future GHC discovers through boolean algebra that your otherwise is redundant, then presumably it was always redundant, and you can just remove that branch, and replace the second last guard with otherwise.

For me, that's the key bit: that users should be careful not to write a redundant guard. OK. But redundant guards can be useful as documentation (especially if the redundant guard is far from other guards and/or is non-trivial)... though perhaps I should put my documentation in a comment, not in the code.

Ericson2314:
> though perhaps I should put my documentation in a comment, not in the code:

Yeah event today I might write something like:
```haskell
| otherwise {- bool expr for what remains -} -> ...
```

I wish it was code too, (and I think I do because I also wish there was a mode that required all alternatives to be disjoint, because fall-through is so imperative and ugly), but solving the first part and making a `-XNoFallThrough` I think are problems for other proposals :)

cgibbard:
I added a specification of the cases which the proposal aims to prohibit, paraphrasing Simon's comment.

cgibbard:
Should this still be marked needs-revision, or does the revision I made suffice?

yav:
@cgibbard It's up to you---if you think that the revised version has stabilized, let me know, and I'll ready trough it again, and submit it to the committee for review.

cgibbard:
I'll leave that decision up to @Ericson2314, but I think I'm reasonably happy with it. It would of course be nice to have a proper formalisation of some of this, but I'm not sure that it's absolutely necessary.

ndmitchell:
My objection still remains that this doesn't rise to the level of formalism or precision where it could ever be implemented by any other compiler, or where it can be relied upon to remain stable over multiple GHC versions. It describes a lower bound on a language feature, and then allows an implementation to permit arbitrarily more programs that satisfy certain properties - I'm unaware of any other language feature that has such a property. Thus, it feels like a souped up warning, rather than a language feature. If it were a warning, I'd have no problem with it.

Ericson2314:
@yav Yes, let's resubmit to the committee.

With this last round we just incorporated @simonpj's and @goldfirere's suggestions on how to clarify, verbatim. I'm happy to keep on incorporating concrete suggestions of that sort.

-------

@ndmitchell

 - Didn't `FlexibleInstances` or  `FlexibleConstraints` come about because GHC was accidentally more permissive than the spec?
 - `ImpredicativeTypes` was incubated knowingly-broken for many years until "quick look" got us all the way there.

I don't think it's *ideal* to land an open ended extension, but I think the scourge of implicitly partial code is a far greater evil.

I don't think anyone has the budget to audit, categorize, and rework everything our pattern match checker does today --- in fact, precisely because the pattern match checker was considered "cool shit just for lower-stakes warnings", I think we've been somewhat lax about what what's gone in it.

I am hoping this proposal and extensions would be the, beginning, not end, of a transition period in which that auditing, categorizing, and reworking *would* happen. *After* that, we could get separate extensions for the "lower bound" that is most ready for standardization and other tricks GHC can do.

Perhaps we could do some forwards-compatible tricks to keep this more by the book and assuage your concerns? For example, by simultaneously defining `NoIncomplete` and e.g. `FlexibleCompleteness` at the same time, but require them to be enabled or disabled together, so the meaning of `NoIncomplete` itself remains indeterminate.

ndmitchell:
@Ericson2314 - For instances/constraints GHC was more permissive than the spec, so exactly how it was more permissive got formalised, and turned into extensions. For impredicative types, a known broken extension limped on for a while, breaking peoples code on a semi regular basis with random hacks around `$` in the type checker. I'm not sure I'd use either of those as models of how we should do things in future.

Rather than land an open ended ambiguous extension, why not add an open ended ambiguous warning? Many warnings have those properties. Once you have the warning in a robust and specified shape, then it could become a language pragma. Why is `-Wincomplete` meaningfully worse than `NoIncomplete`? Usually for warnings it's worse because the warning doesn't have the properties of stability, definition etc - but for `NoIncomplete` you don't have those anyway.

Splitting it into two language pragmas that only exist in tandem seems worth than either alternative. And if everyone disagrees with me and wants the extension anyway, that's OK. I think it's probably a mistake, but its not a massively costly one.

Ericson2314:
> Once you have the warning in a robust and specified shape, then it could become a language pragma. Why is `-Wincomplete` meaningfully worse than `NoIncomplete`?

It all goes back to how this proposal is about norms. I think it's crazy that GHC is inserting run-time errors i didn't write into my programs, as do others. We can package up existing warnings into a new convenience `-Wincomplete`, but I don't think that's going to move the needle on the norms between GHC, its users, and its developers. As long as the pattern match checker is just for warnings, it's not going to register as a problem that it's behavior is ad-hoc.

As long as users are turning on warnings, they're going to think *they* are doing bad things writing incomplete patterns, not *GHC* is doing bad things allowing them too, just as programers of untyped languages blame them selves about run-time "type" errors, or programmers without `Maybe` blame themselves for "null pointer exception".

> I'm not sure I'd use either of those as models of how we should do things in future.

They are not ideal models, but I think they'll suffice. We have decent test coverage that the pattern match checker, however unspecified, doesn't start complaining about code it used to accept. That should avoid the practical problems `ImpredicativeTypes` had. But also, `ImpredicativeTypes` had a disclaimer that it's behavior wasn't fully formalized and was subject to change.

I guess I invite you to propose an alternative roadmap to change these norms on a comparable budget, because all I envision are plans where nothing changes until someone spends 3 months writing formalizations of a hierarchy of pattern checkers checkers, and then spends another 3 months refactoring the pattern match checker to implement them. And, as much as I wish it were otherwise, the problems here are too boring and too beginner-focused ("those in the know just use `-Werror` and get on with their lives") to attract that level of investment any time soon.

adamgundry:
> We can package up existing warnings into a new convenience `-Wincomplete`, but I don't think that's going to move the needle on the norms between GHC, its users, and its developers.

I think what matters here are defaults, rather than whether the more-restrictive behaviour is activated by `-XNoIncomplete` or by `-Wincomplete`. Personally I'd favour adding a `-Wincomplete` option, and then arguing that `-Wwarn=incomplete` or even `-Werror=incomplete` should be enabled by default.

ndmitchell:
> I guess I invite you to propose an alternative roadmap to change these norms on a comparable budget, because all I envision are plans where nothing changes until someone spends 3 months writing formalizations of a hierarchy of pattern checkers checkers, and then spends another 3 months refactoring GHC to implement them.

I spent 530 comments and 8 months on just one single prong of the RecordDotSyntax proposal :P. This process is not meant to be cheap, it's meant to be somewhat formalised.

I guess I'm still not convinced your plan gets us to the end state you want cheaply. It adds an extension. It doesn't account for evangelising and changing of social norms where I think 3 years would be to little (btw, I spent a PhD writing a pattern match checker and trying to change norms - so I dropped at least 2 years on this crusade without moving the needle). My suggestion would be add a warning (no more effort than your plan) then evangelise (no more effort than your plan). I think your plan assumes that a language extension gives you a significant amount of that evangelising for free, but I suspect not.

I don't think I've got anything more to say, so I leave it to the committee to decide. (This isn't me rage quitting, I just think we have genuinely held differing views, so leaving it to an independent arbiter makes sense.)

Ericson2314:
@adamgundry I'm willing to admit that a `-Werror=incomplete` default packs the same punch, though I don't think a `-Wwarn=incomplete` default does. On the other hand, a `-Werror=incomplete` default is a big breaking change that I assume is totally out of the question. (I as said above, I wouldn't submit `NoIncomplete` to the report until it was precisely specified and not just lower-bounded.)

Beyond the existing connotations of language extensions between petitions to change Haskell, and warnings not being that, there is also the advantage that an extensions is all or nothing (off or error), whilest the `-fdefer-*` debugging aid is separate probably discovered after. That makes the "order of intentionality" [status quo, compile-time error, run-time error], which I think is right and hard to create by other means.

-------

@ndmitchell Ah, I didn't know your thesis was on that! I'll check it out, and appreciate that you've been going after the same goal, too.

simonpj:
> With this last round we just incorporated @simonpj's and @goldfirere's suggestions on how to clarify, verbatim. I'm happy to keep on incorporating concrete suggestions of that sort.

Some more suggestions, under "Proposed change specification"

* You first list some warning categories (`incomplete-patterns` etc) and then have a list of bullets after "More specifically, the cases we intend to make illegal ...".  I believe that every bullet corresponds precisely to one of the preceding warning categories.  Is that right? If so, please add the category to each of the bullets so it's clear.

* "For a report, the guiding principle behind NoIncomplete only requires that the Haskell implementation's pattern match completeness checking by sound, not sophisticated."  I'm not sure what you are meaning here.  This is the specification, let's say what the specification specifies.  For example, I think you mean:

  *The implementation must have no false negatives.  That is, with `-XNoIncomplete` if the program compiles without error, it must not fail at runtime with any of the above errors.  However, the implementation is permitted to have false positives: it may report an error when in fact the pattern match is complete.   A good implementation will strive to reduce the number of false positives."

  Is that what you mean?  Let's say so!  (I'm not sure whether or not the present implementation makes these guarantees.  Ask Sebastian Graf.)

* If you say `-XNoIncomplete` and `-fdefer-incompleteness-errors`, all the incompleteness errors turn into warnings.  But isn't that *precisely* what you get with `-XIncomplete`?  My head spins with all these variations of errors vs warnings.  Could it be made simpler?

Ericson2314:
@simonpj The first two sound good to me, and we'll do those.

The overlapping functionality in the third one is intentional.

We could just not introduce the `-fdefer-*` mechanisms. But then the user would have to turn on and off `-XNoIncomplete` to get back the old functionally for debugging, and that seems very unbecoming for a language extension. I also really want `-XNoIncomplete` to be something that one would turn on permanently to best enshrine the new norm/"new covenant" that GHC isn't going to go be secretly inserting run-time errors. Constantly wigging it would seem to muddle establishing that? This all goes back for this proposal being a bit more in service of social than technical ends compared to what we're used to.

If it's any relief, the old warnings only affect `-XIncomplete`, and the new `-fdefer-*` flags only affect `-XNoIncomplete`, so there at least isn't a combinatorial explosion of interactions.

simonpj:
> But then the user would have to turn on and off -XNoIncomplete to get back the old functionally for debugging, and that seems very unbecoming for a language extension

This is the bit I don't get.  By deferring the errors you are saying "I want an incomplete language. Warn me at compile time, and fail at runtime".   Adding a raft of new defer flags to achieve what we already have seems unnecessary complexity. Or at least complexity for which the proposal should make a strong case.

(If nothing else, it would help to point out that `NoIncomplete` + deferring = `Incomplete`.  It took me some time to work that out.

hanshoglund:
> We could just not introduce the -fdefer-* mechanisms. But then the user would have to turn on and off -XNoIncomplete to get back the old functionally for debugging

Consider simplifying the proposal by removing them. They do not buy you much over just writing out `error "TODO"` explicitly. 

An interactive/IDE-style workflow to fill in the stubs would be nice, but that seems out of scope here :D.

Ericson2314:
> They do not buy you much over just writing out `error "TODO"` explicitly.

In my opinion they do, just the same as writing `unsafeCoerce` everywhere is no substitution for `-fdefer-type-errors`. The `-fdefer-*` options are O(1) for humans, while the manual ones are `O(n)` for humans.`

That said, we will better explain that the new `-fdefer-*` is only for `-XNoIncomplete`, while the old `-W...` are for `-XComplete`, so there isn't a "permanent" combinatorial blow-up.

hanshoglund:
@Ericson2314 do you plan to revise this further, or is it ready to be resubmitted?

(Ping @yav and @nomeata)

Ericson2314:
I cleaned up a few things, let's resubmit! I always forget that part.

nomeata:
/remind @yav to check for progress in two weeks

bravit:
Hi @Ericson2314, I'm very sorry for the delays. I'll review this proposal in a couple of days and then run our regular procedure.

bravit:
I've recommended to accept this proposal. I think it's ready for that.

goldfirere:
I'm unsurprisingly sympathetic to the design goal of this proposal -- make dynamic errors into static ones -- but I find the design here to be too ornate. (Disclaimer: I have not re-read the thread above. Feel free to refer me to previous comments if they address my points.)

I'd like to make a very simple counterproposal: Let `incomplete` be a new class of warnings, like `all` or `compat`, and let `incomplete` include `incomplete-patterns`, `incomplete-uni-patterns`, `incomplete-record-updates`, `missing-fields`, and `missing-methods`. That's it!

Now, what you want can be achieved with `-Werror=incomplete`.

Do you agree that this would have the same net effect?

I can see an objection that this new approach does not adequately signal to programmers that these incomplete features are problematic in the language -- and I agree with this objection. But first, I want to make sure that my understanding of the proposal is accurate, in that it just amounts to taking existing warnings and making them errors.

If that is the case, and if we agree that we want to somehow make this a language pragma instead of a set of warnings, we must make a case for that desire: is it simply about a communication of values? With my `incomplete` warning set, it would seem that any programmer who wants to eliminate the possibility of these exceptions has a way to do so. Turning this into a language pragma then is "just" cosmetic. My inclination is actually in favor of this language pragma approach, but I also want us to be honest in how superficial this desire is.

If we do decide to go the language-pragma route, I would want a consistent story between the features here, `-fdefer-type-errors`, and `-XPartialTypeSignatures`. Right now, GHC has (at least) three broadly similar mechanisms for mediating between errors, warnings, and silence:

* For the massive bulk of warnings, you can get silence with `-wxyz`, the warning with `-Wxyz`, and the error with `-Werror=xyz`. Each warning has a default of either silence or warning. Some warnings (the ones at issue in this proposal) appear as errors at runtime if the program compiles.
* For type errors, you can get silence with `-fdefer-type-errors -Wno-deferred-type-errors`, the warnings with `-fdefer-type-errors`, and the error by default or with `-fdefer-type-errors -Werror=deferred-type-errors` (though the precise error text depends on which approach you use to get the error). The warnings appear as errors at runtime if the program compiles.
* For partial type signatures, you can get silence with `-XPartialTypeSignatures -Wno-partial-type-signatures`, the warnings with `-XPartialTypeSignatures` and the error by default or with `-XPartialTypeSignatures -Werror=partial-type-signatures` (though the precise error text depends on which approach you use to get the error). There is no runtime effect if the program compiles.

This is not the hallmark of good design! The proposal, as written, essentially adds a fourth bullet to this list:

* For incompleteness warnings, you can get silence with `-wxyz` or with `-XNoIncomplete -fdefer-xyz -Wno-deferred-xyz` or (in some cases) by default, the warning with `-Wxyz` or with `-XNoIncomplete -fdefer-xyz` or (in other cases) by default, the error with `-Werror=xyz` or with `-XNoIncomplete` or with `-XNoIncomplete -fdefer-xyz -Werror=deferred-xyz`. In all cases, there is a runtime error if the program compiles. In all cases, the different ways of getting the error message / warning message produce slightly different text (in that the warning flag printed in the text would be different). 

This is dizzying, and I do not wish to count the number of negations in the last paragraph (building on the two already present in `-XNoIncomplete`).

So, for my part, I would want either the very simple counterproposal I made above -- just a new warning group! -- or a more thorough redesign and simplification of this small zoo, so that all these different features have parallel interfaces.

-----------------------------

Separate from my main objection above (complexity):

* I'm not as bothered as some about the tightness of the specification. In the deep end of the language, we have a number of not-quite-specified features. Adding to this list is sad, but I don't think we should let the perfect become the enemy of the good here. I like how the proposal specifies that there shall be no false negatives. In an ideal world, there would be a very local mechanism to squelch a false positive (that is, without enabling, say, `-fdefer-incompleteness` for the whole module just because GHC slips up on one function definition), but I don't require that local mechanism to feel able to proceed here.
* I don't understand the Examples section at all. It lists flags and results, but no code. And it seems to be inconsistent, claiming that `NoIncomplete` leads to errors while `-XNoIncomplete` leads to warnings, but surely there is no difference between `{-# LANGUAGE NoIncomplete #-}` and `{-# OPTIONS_GHC -XNoIncomplete #-}`. So I skipped this section.

Ericson2314:
@goldfirere Thank you for the nice review! I am actually happy doing `-Wpartial` for now, so long as

- it doesn't sap our ability to do `-XNoIncomplete` later, because no one wants to deal with the optics/values in isolation later.
- https://github.com/ghc-proposals/ghc-proposals/pull/319 still have a chance to go forward, despite being much more ergonomic with `-Wpartial` or `-Werror=partial` than without. (I was concerned hitching a language extension's best use to warnings would appear worse than hitching it to another language extension.)

simonpj:
>  I am actually happy doing -Wpartial for now

Richards points seem quite compelling to me, so this looks like a good way to go.

nomeata:
Does this mean that this proposal can be closed, and instead a new proposal with `-Wpartial` will be created (or, because warnings are not language features, “simply” be implemented without proposal process)?

Ericson2314:
Whether we need a new/modified proposal or not, I am hoping someone could response to my two points in https://github.com/ghc-proposals/ghc-proposals/pull/351#issuecomment-860817242. 

I do agree with @goldfirere there are two many knobs, but I also think `-XNoIncomplete` raises some good uncomfortable questions about "pragmatics" and appearences. I wouldn't those issues to to fester because we do `-Wpartial` and never return to this stuff again.

simonmar:
I'm very happy with `-Wpartial`, but I think there's a clear distinction between warnings and language features that is in danger of being muddied here. 

* Warnings are allowed to be unspecified, implementation-dependent, and can freely vary in behaviour from release to release of the compiler.
* For language features, while they're not always fully specified in practice, have the expectation that we *could* write down a specification if we wanted to, and stick to it. We do sometimes change the meaning of extensions and break backwards compatibility, but we take that seriously and document it in the release notes - it's like changing the language, not just the compiler.

This distinction is important in practice: that's why warnings are warnings, and `-Werror` is a code quality tool rather than something you turn on in your production build. Cabal/Hackage rejects `-Werror` in released packages, IIRC.

So `-Wpartial` can happily be a warning, but the bar for a language extension is a lot higher IMO.

Ericson2314:
@simonmar

> and `-Werro` is a code quality tool rather than something you turn on in your production build.

Ah but this is precisely what sways it for me! I absolutely *don't* want GHC inserting runtime errors in my production build to patch over missing code. `-Werror=partial` is required to ensure that all possibilities have been accounted for my the programmer, or we're not deploying.

That the absence of `-Wpartial` has operational effect in inserting extra runtime checks makes it *not* like other warnings. It does make it exactly like `-fdefer-type-errors` though.

I do think `-Wpartial` is an improvement in that it groups together these scattered warnings that have a runtime effect when disabled, so we are used to thinking about them together. But I don't think it goes far enough, I would still propose `-XNoIncomplete`, just with a simple description now that I could refer to `-Wpartial` and not it's constituents in defining this language extension.

bravit:
It seems that there is a strong opposition within the committee to solving the problem in this proposal via a GHC extension, so the proposal in its current form doesn't have chances to be accepted. 

As @nomeata points out, proposing and implementing new warnings doesn't change the language so this doesn't require going through the GHC proposal process at all. Though we can still do that after modifications.

Unfortunately, I have to send this proposal back to revision to reflect the current situation with it as we cannot go further along our process.

simonmar:
> @simonmar
> 
> > and `-Werro` is a code quality tool rather than something you turn on in your production build.
> 
> Ah but this is precisely what sways it for me! I absolutely _don't_ want GHC inserting runtime errors in my production build to patch over missing code. `-Werror=partial` is required to ensure that all possibilities have been accounted for my the programmer, or we're not deploying.

I understand why you'd want that. But in exchange for the guarantee that "if it compiles, it isn't `_|_`", you lose the guarantee that it compiles at all. It might not compile with a future GHC, because the implementation of `-Wpartial` might change such that it has a new false positive with your code. I think you're OK with that; I'm just quite dubious that we want a first-class language feature to have that property.

> That the absence of `-Wpartial` has operational effect in inserting extra runtime checks makes it _not_ like other warnings. It does make it exactly like `-fdefer-type-errors` though.

I think these are different. `-Wpartial` is unspecified at compile-time, while `-fdefer-type-errors` is unspecified at runtime. If there's a type error in your code and you use `-fdefer-type-errors` we don't specify the runtime behaviour: it may or may not be `_|_`. However we can specify precisely whether a program that has warnings with `-Wpartial` will actually be `_|_` at runtime.

> I do think `-Wpartial` is an improvement in that it groups together these scattered warnings that have a runtime effect when disabled,

Agreed!

Ericson2314:
> you lose the guarantee that it compiles at all...

Well, this does happens with the type system some some regularity. That is, of course, quite analogous to me.

> I think you're OK with that; I'm just quite dubious that we want a first-class language feature to have that property.

Fair enough. I don't love it but to me it's more about properly accounting for an existing debt than creating a new one, since I think people should feel free to do `ghc-options: -Werror=partial` in their cabal file in the absence of change.

> If there's a type error in your code and you use -fdefer-type-errors we don't specify the runtime behaviour: it may or may not be _|_. However we can specify precisely whether a program that has warnings with -Wpartial will actually be _|_ at runtime.

That is true, and a good point.

-------

For my first question about whether abandoning this will effect that outcome of https://github.com/ghc-proposals/ghc-proposals/pull/351, I resubmitted that proposal so we can just discuss that there.

Something that just re-occurred to me is that if absurd patterns are figured out per https://github.com/ghc-proposals/ghc-proposals/discussions/423 (there is new strong case to get them working so empty-`case` is not incongruously strict), `-XNoIncomplete` could requires "traditional Haskell" complete patterns + absurd patterns for any GADT funny business.

I bring it up think that the language feature can in fact by specified after all: The type checker is still doing some spooky magic on the absurd patterns, but the pattern exhaustiveness checker can just "trust" the absurd patterns that the type checker signs off on and treat them as `_ -> undefined` for pattern exhaustiveness purposes, which brings us back wihtin traditional easy cases!

simonpj:
Richard said

> I'd like to make a very simple counterproposal: Let incomplete be a new class of warnings, like all or compat, and let incomplete include incomplete-patterns, incomplete-uni-patterns, incomplete-record-updates, missing-fields, and missing-methods. That's it!  Now, what you want can be achieved with -Werror=incomplete.

I think others have then used `partial` instead of `incomplete` as the name of the warning class.

I think that there seems to be some consensus that this is a simpler path forward.

John, are you minded to take that path?

John asked about [his two questions here](https://github.com/ghc-proposals/ghc-proposals/pull/351#issuecomment-860817242).  My response (without having deeply studied the NoFaillibleDo thing) is that we simply have to take each proposal on its merits.  Adding a `incomplete`/`partial` warning class is a small change, and isn't going to affect future proposals much.

goldfirere:
Though I don't think this is the proposal to do so, I'd like to note here in public that I'd be enthusiastic about a proposal that cleaned up the deferred-type-errors/partial-type-signatures mess described in my https://github.com/ghc-proposals/ghc-proposals/pull/351#issuecomment-849654517.

Ericson2314:
>  Adding a `incomplete`/`partial` warning class is a small change, and isn't going to affect future proposals much.

Yeah I am not worried about that. Just was worried about *not* enshrining it as an extension and that proposal, since it's the desire to get to get good incomplete errors/warnings that shapes the design space e.g. ruling out qualified do as-is.

Ericson2314:
Like I said before, I'll revisit this with absurd patterns, so closing this as desired.

PR #111, opened 2018-02-13
Linear types

aspiwack:
This proposal introduces a notion of linear function to GHC. Linear functions are regular functions that guarantee that they will use their argument exactly once. Whether a function f is linear or not is called the multiplicity of f. We propose a new language extension, `-XLinearTypes`. When turned on, the user can enforce a given multiplicity for f using a type annotation.

[rendered](https://github.com/tweag/ghc-proposals/blob/linear-types2/proposals/0000-linear-types.rst)

This proposal has already been through [one round of review](https://github.com/ghc-proposals/ghc-proposals/pull/91). If you are new to this, you don't need to read the conversation there: it has been integrated into this newer version of the proposal. For a full understanding of the proposal, reading the [companion article](https://arxiv.org/abs/1710.09756) is recommended. For a bird's eye view of the proposal, you can browse the [wiki page](https://ghc.haskell.org/trac/ghc/wiki/LinearTypes).

If you already read the previous version of the proposal, here is a summary of the main changes:

- The main technical change is that we've removed the multiplicity 0 from the proposal: it adds quite a bit of complication, and was only useful in interaction with Dependent Haskell. Maybe. And it was too much complication for something which was only potentially useful. (0 is discussed in the alternatives, as help if it's ever needed)
- We added examples based on external contributions and on the [linear-base library](https://github.com/tweag/linear-base/) that we've been implementing to demonstrate the programming style that this proposal heralds.
- We discuss how (deep, multi) pattern-matching is type-checked.
- We added a discussion on linearity vs affinity.
- We expanded vastly the discussion on exceptions.
- We added a (beginning of a) discussion on the effect on Core-to-Core passes.
- We changed the policy regarding unrestricted newtype definitions and GADT syntax when the extension is turned off

By @aspiwack, @facundominguez, @mboes, @jyp and @simonpj

### Ongoing discussion with the committee 

- [x] Defining consume exactly once: interaction with evaluation and primitive unboxed types
- [x] Linear pattern synonyms
- [ ] Printing multiplicity
- [ ] Multiplicities in let bindings
- [x] Case syntax
- [ ] Of the use of linear types and resources

PR #352, opened 2020-07-29
Update a couple missed cases to the new linear arrow syntax

mb64:
I believe a couple arrows were missed when changing the linear arrow syntax. (Of course, I may be mistaken – I might have missed something myself, since there was tons of discussion and I didn't read it all.)

The syntax of linear arrows underwent a couple of revisions over the course of the proposal, and I think updating these few cases was probably overlooked by accident.

The syntax section specifies the current accepted syntax as `#->` and `# p ->`. For reference, here's [the comment changing `:p->` to `#p->`](https://github.com/ghc-proposals/ghc-proposals/pull/111#issuecomment-475681083) and [the comment changing `->.` to `#->`](https://github.com/ghc-proposals/ghc-proposals/pull/111#issuecomment-519175000).

### DIFF
--- a/proposals/0111-linear-types.rst
+++ b/proposals/0111-linear-types.rst
@@ -282,7 +282,7 @@ This proposal adds two new syntactical constructs:
     ::
 
       type family F (a :: *) :: Multiplicity
-      f ::  forall (a :: *). Int  :(F a)-> a -> a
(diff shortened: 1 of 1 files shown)


nomeata:
@aspiwack can you have a look at merge if good?

PR #353, opened 2020-08-15
RST rendering fixes

shayne-fletcher:
Saw some issues arising from the mapping from markdown to ReST.
- Since ReST doesn't support inline markup, some renderings were broke; fix that;
- Formatting of section 2.3 didn't survive the transition that well; use some ReST role stuff to recover the intended rendering.

Cc @ndmitchell


ndmitchell:
Look reasonable to me, thanks @shayne-fletcher

nomeata:
Sorry for the late response, I was traveling, and thanks for cleaning up!

PR #356, opened 2020-08-24
Change syntax for linear arrows

mboes:
This is not a full proposal. It's a small change to a recently accepted proposal, ahead of the implementation making it into a release.

[Link to original proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0111-linear-types.rst).

In the linear types proposal as accepted, we proposed to introduce a new type constructor `FUN p a b` where `p :: Multiplicity`, along with *two* new syntactic forms in types:

* `a #p -> b` stands for `FUN p a b`;
* because multiplicity `One` is so common, `a #-> b` stands for `FUN
  'One a b`.

Based on implementation experience writing `linear-base`, this change removes the second syntactic form and use `^` where `#` was used previously. The motivations are:

* A prefix `#` in identifiers is already used for overloaded labels. Since overloaded labels may one day appear in types, we deem it wise to avoid any potential for conflict in the future, by picking a different character instead. This PR proposes `^` instead of `#`, though `%` is also under consideration.
* From a didactic point of view, we prefer to convey the idea that there really is only one function space in Haskell, denoted `a -> b`. The arrow is parameterized in various ways, including by a multiplicity. The `a ^p -> b` syntax is in keeping with this idea.
* `#->` is 3 characters long where `::` and `->` are of width 2. This makes aligning a mix of all three lexemes vertically in a type signature awkward.

We propose to henceforth denote a linear function with type `a ^1 -> b`. For now, this is technically a new syntactic form that replaces `a #-> b`. In the future, this may or may not remain special. If the type-level is further unified with the term-level as some have plans for, it is possible that integer literals in types will become overloaded, like they are in terms. If that happens, we'll have `1 :: GHC.Types.Nat` or `1 :: Multiplicity` depending on context, and `a ^1 -> b` will no longer be a special syntactic form. Anything between `^` and `->` will then truly be an arbitrary type in all cases.

We have tentative plans for other multiplicities in the future, like multiplicity `0`.

Overall, we view this change to the syntax of linear arrows as a simplification: the user need only learn one new syntactic form, not two.

### DIFF
--- a/proposals/0111-linear-types.rst
+++ b/proposals/0111-linear-types.rst
@@ -105,7 +105,7 @@ Previous condition
 - We must work out an acceptable syntax for this all. In particular, ``:`` in
   types is taken by the list-cons operator, so we'll need something new.
 
-This condition has been met, by using a syntax around ``#``, as described in
+This condition has been met, by using a syntax around ``^``, as described in
 the Syntax_ section below.
   
 Motivation
@@ -219,7 +219,7 @@ as follows).
 - Consuming a function exactly once means applying it and consuming
   its result exactly once
 
-The type of linear functions from ``A`` to ``B`` is written ``A #1 ->
+The type of linear functions from ``A`` to ``B`` is written ``A ^1 ->
 B`` (see Syntax_).
 
 Linearity is a strengthening of the contract of the regular function
@@ -246,7 +246,7 @@ polymorphic functions may have variable multiplicity (see also Syntax_), *e.g.*
 
 ::
 
-  map :: (a #p -> b) -> [a] #p -> [b]
+  map :: (a ^p -> b) -> [a] ^p -> [b]
 
 Without polymorphism, we would need two implementations of ``map`` with
 the exact same code: one for ``p=1`` and one for ``p=ω``. Function
@@ -255,7 +255,7 @@ hence, would require four identical implementations:
 
 ::
 
-  (.) :: (b #p -> c) -> (a #q -> b) -> a #(p ':* q) -> c
+  (.) :: (b ^p -> c) -> (a ^q -> b) -> a ^(p ':* q) -> c
 
 Syntax
 ------
@@ -265,45 +265,46 @@ Syntax
 This proposal adds two new syntactical constructs:
 
 - The multiplicity annotated arrow, for polymorphism, is written ``a
-  #p -> b`` (where ``a`` and ``b`` are types and ``p`` is a
-  multiplicity). This steals syntax as ``(#)`` is a valid
-  type operator. That is the syntax entry for types becomes:
+  ^p -> b`` (where ``a`` and ``b`` are types and ``p`` is
+  a multiplicity). We add a new production to the grammar for
+  ``type``:
 
   ::
 
-    type -> btype [[# btype] -> type]
+    btype PREFIX_CARET btype -> ctype
 
+  The ``PREFIX_CARET`` means that the ``^`` character and the
+  multiplicity following it should never have a space in between.
 
-
-  - In ``a #p -> b``, ``p`` can be any type expression of kind
+  - In ``a ^p -> b``, ``p`` can be any type expression of kind
     ``Multiplicity`` (see below). So that the following is legal
     (though see Alternatives_):
 
     ::
 
       type family F (a :: *) :: Multiplicity
-      f ::  forall (a :: *). Int  #(F a) -> a -> a
+      f ::  forall (a :: *). Int  ^(F a) -> a -> a
 - When ``-XScopedTypeVariables`` is switched on, binders can also be annotated with a multiplicity:
 
   ::
 
-    \x :: A # 'One -> x
+    \x :: A ^'One -> x
 
   is the identity function at type ``A ⊸ A``. A binder can be
   annotated with a multiplicity without a type like this
 
   ::
 
-    \x # 'One -> x
+    \x ^'One -> x
 
   This modifies the syntax entry for pattern with signature annotation
   as follows as follows
 
   ::
 
-    pat -> pat [# btype] [:: type]
+    pat -> pat [PREFIX_CARET btype] [:: type]
 
-  where the ``btype`` after the ``#`` must be of kind ``Multiplicity``
+  where the ``btype`` after the ``^`` must be of kind ``Multiplicity``
   (see below).
 
   This form is disallowed for:
@@ -312,25 +313,25 @@ This proposal adds two new syntactical constructs:
 
     ::
 
-      forall (a # 'One). a -> Int -- rejected
+      forall (a ^'One). a -> Int -- rejected
   - Top-level signatures (though, see `Toplevel binders`_)
 
     ::
 
-      foo # 'One :: A -> B -- rejected
+      foo ^'One :: A -> B -- rejected
       foo x = …
 
   The form is however permitted in records (see `Records`_ below)
 
   ::
 
-    data R = R { unrestrictedField # 'Many :: A, linearField # 'One :: B }
+    data R = R { unrestrictedField ^'Many :: A, linearField ^'One :: B }
 
   This modifies the field declaration syntax to
 
   ::
 
-    fielddecl -> vars [# btype] :: (type | ! atype)
+    fielddecl -> vars [PREFIX_CARET btype] :: (type | ! atype)
 
 In the fashion of levity polymorphism, the proposal introduces a data
 type ``Multiplicity`` which is treated specially by the type checker,
@@ -359,22 +360,22 @@ A new type constructor is added
 
     FUN :: Multiplicity -> forall (r1 r2 :: RuntimeRep). TYPE r1 -> TYPE r2
 
-``FUN`` is such that ``FUN p a b ~ a #p -> b``.
+``FUN`` is such that ``FUN p a b ~ a ^p -> b``.
 
 The linear and unrestricted arrows are aliases:
 
 - ``a -> b`` is an alias for ``FUN 'Many a b``,
 - ``a ⊸ b`` (Unicode syntax) is an alias for ``FUN 'One a b``.
 
-The type ``a # 'One -> b``, being such a common case, can also be
-written ``a #1 -> b`` for brevity, where ``#1`` is a single token.
-Like ``#``, this steals syntax, since with the ``-XDataKinds``
+The type ``a ^'One -> b``, being such a common case, can also be
+written ``a ^1 -> b`` for brevity, where ``^1`` is a single token.
+Like ``^``, this steals syntax, since with the ``-XDataKinds``
 extension, ``1`` is a valid type literal. However, integer literals
 are already overloaded at the term level and this syntax is
 forward-compatible with any future proposal to overload literals at
 the type level as well. If and when overloaded integer literals in
-types become available, `a #1 -> b` would be parsed as 4 lexemes: the
-type `a`, the symbol `#`, the type literal `1`, the symbol `->` and
+types become available, `a ^1 -> b` would be parsed as 4 lexemes: the
+type `a`, the symbol `^`, the type literal `1`, the symbol `->` and
 the type `b`.
 
 Printing
@@ -385,18 +386,18 @@ multiplicities are printer: ``-fprint-explicit-multiplicities``. It is
 turned off by default.
 
 When ``-fprint-explicit-multiplicities`` is turned on, every arrows
-are printed in the form ``#p ->``. For instance, the type of the
+are printed in the form ``^p ->``. For instance, the type of the
 unrestricted ``fmap`` function from ``base`` will be printed as:
 
 ::
 
-    fmap :: Functor f => (a #'Many-> b) #'Many-> f a #'Many-> f b
+    fmap :: Functor f => (a ^'Many-> b) ^'Many-> f a ^'Many-> f b
 
 And a linearised ``List.map`` would be printed as:
 
 ::
 
-    lmap :: (a #'One-> b) #'Many-> [a] #'One-> [b]
+    lmap :: (a ^'One-> b) ^'Many-> [a] ^'One-> [b]
 
 When ``-fprint-explicit-multiplicities`` is turned off (as is the
 default), the shorthands are used when available. The above examples
@@ -405,7 +406,7 @@ are printed as
 ::
 
     fmap :: Functor f => (a -> b) -> f a -> f b
-    lmap :: (a #1 -> b) -> [a] #1 -> [b]
+    lmap :: (a ^1 -> b) -> [a] ^1 -> [b]
 
 Where no shorthand is available, as is the case for multiplicity
 polymorphic arrows, then the long form is used in both cases. So a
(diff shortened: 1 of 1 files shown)


anka-213:
What is the motivation for adding a space between the arrow and the number? `a #1 -> b` instead of `a #1-> b`? 

To me, this new syntax makes it seem like the `#1` is an argument to `a` instead of being connected to the arrow.

If you write a function like this
```haskell
f :: a #1
  -> b
  -> c #1
  -> d
```

would it be a correct interpretation that the multiplicity `#1` is associated with the arguments `a` and `c`?

(Sorry for the bikeshedding)

mboes:
We could allow both. Allowing a space before the arrow makes it possible to
vertically align arrows at the beginning of each line in type signatures.
It's not the style I use, but I know it's a very common style.

int-index:
In the examples, `#p` and `#1` are often written without a space, but this won't work if the user enables `-XOverloadedLabels` (in particular, when https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0170-unrestricted-overloadedlabels.rst is implemented).

So maybe the recommended style should be to write `a # 1 -> b` rather than `a #1 -> b` to avoid this conflict.

int-index:
Also, this proposal introduces `a # p -> b` as a ternary operator, while `#` is also a valid binary type operator. So there's now a difference between:

1. `a # p -> b`
2. `(a # p) -> b`

In the latter case, it's a type operator.

I'd like to offer a syntax that resolves both conflicts (`-XOverloadedLabels` and `-XTypeOperators`):

```
a #[p] -> b
```

The `#` must be _prefix_ (as defined in #229). Thus:

1. No conflict with overloaded labels because an overloaded label cannot start with `[`
2. Minor conflict with type operators, but type operators are typically with spaces on both sides

I also think this syntax will be nicer to implement, with a dedicated parser production for the `a #[p]` construct.

anka-213:
> We could allow both. Allowing a space before the arrow makes it possible to vertically align arrows at the beginning of each line in type signatures. It's not the style I use, but I know it's a very common style.

My main question is if it makes semantic sense to align it like that? Is it correct to associate the `#1` with the argument instead of associating it with the arrow?

If we write it as `#->` or as `#1->`, it is very clear that it is a part of some kind of funky arrow, instead of leading readers to believe that `#` is a binary operator or that `#1` is some kind of label. What would be the least confusing and misleading for people who see this syntax for the first time?

I also believe that `Foo # 1 -> Bar` is already valid syntax, so that also makes me a bit more skeptical about the choice to allow/encourage spaces:
```haskell
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE PolyKinds #-}

type a # b = a

foo :: a # 1 -> a
foo = id
```

----

I am more positive to `a #[p] -> b`, since it clearly shows that something strange is going on. (It is a tad verbose though.) But I still wonder if
```haskell
a #[p] -- associate the multiplicity with the argument and not the arrow
  -> b
```
is a correct reading of it? I'm not sufficiently familiar with linear types to say if the answer is yes or no.

mboes:
@anka-213 Yes, "associating the multiplicity with the argument" is a correct reading. In fact reading it that way makes it possible to continue using both styles of type signatures:

```haskell
foo ::
  a #1 ->
  b ->
  c
```

and

```haskell
foo
  :: a #1
  -> b
  -> c
```

anka-213:
Ah, I see. Then it makes a lot more sense.

int-index:
Just another syntax idea to avoid a conflict with OverloadedLabels: `a %p -> b`.

mboes:
That latest syntax suggestion makes a lot of sense to me, Vlad. No conflict
but still short. Thanks for suggesting it. If we forbid the interstitial
space, than just like with overloaded labels, % with the space can still be
used as a binary operator.

goldfirere:
A few clarifying questions:

* In the original proposal, is `#1` a new lexeme? That is, would `a # 1 -> b` work? If `#1` is not a new lexeme, then what kind of production exists between the `#` and the `->`? If it's a type, `1` would be a type-level `Nat`. If it's something completely new, then how do we write, for example, a type family application in there? Ditto `%1`.

* In the `#[p]` version, same question.

Maybe we just have a rule that `1` parses like a type-level number, but some later pass special-cases this to be a multiplicity.

I'm in support of a tweak of this form -- I just want to understand what the new BNF would be.

int-index:
Here's what I proposed with the `#[p]` syntax:

```diff
 type
   : btype               { ... }
   | btype '->' ctype    { ... }
+  | btype PREFIX_HASH '[' mult ']' '->' ctype { ... }
   | btype '#->' ctype   { ... }
```

And here's the `%` version:

```diff
 type
   : btype               { ... }
   | btype '->' ctype    { ... }
+  | btype PREFIX_PERCENT mult '->' ctype { ... }
   | btype '#->' ctype   { ... }
```

int-index:
One more option just came to mind:

* `|One| a -> b`
* `|Many| a -> b`
* `|p| a -> b`

Grammar:

```diff
 type
   : btype               { ... }
   | btype '->' ctype    { ... }
+  | PREFIX_BAR mult SUFFIX_BAR btype '->' ctype { ... }
   | btype '#->' ctype   { ... }
```

There are actually many viable options if we're willing to employ `PREFIX_` and/or `SUFFIX_` tokens.

int-index:
Also, isn't it a bit strange that **mult**iplicity uses something other than the **mult**iplication sign? Another option:

```
*One a -> b
*Many a -> b
*p a -> b
```

Grammar:

```diff
 type
   : btype               { ... }
   | btype '->' ctype    { ... }
+  | PREFIX_STAR mult btype '->' ctype { ... }
   | btype '#->' ctype   { ... }
```

buggymcbugfix:
I welcome @mboes's postfix multiplicities suggestion. This emphasizes the type first and then gives the multiplicity, which is how I write my shopping lists (milk x2, apples x3, etc.)! We also opted for postfix multiplicities in Granule for this reason.

I also fully support making things more uniform and removing the special case. `#` evokes the box modality from modal logic, which is analog to `!`, so it's _not_ linear! I'm ok with `#1` as this is like box graded at `1`; written `a [1] -> b` in Granule's syntax.

The downside here is  `#` is already used for overloaded labels _and_ magic hash; e.g. `#1` and `1#` look confusingly similar. Maybe it doesn't matter because I don't foresee `1#` ever occuring in types, yet it does seem somehow confusing.

Example:

```hs
newArray# :: Int# #1 -> a -> (State# d) #1 -> (# State# d, MutableArray# d a #) 

-- and vertically...
newArray#
  :: Int# #1
  -> a
  -> (State# d) #1
  -> (# State# d, MutableArray# d a #) 
```

I think Idris's syntax is quite nice when you use literals (e.g. `1 a -> b`) but not sure if this is great once you introduce variables: `p a -> b`(!?), plus I'm not a fan of the quantity being prefix, especially in complicated types.

@int-index you suggested using `*`, which I find rather cute, on the other hand, doesn't this get a bit confusing/cluttered in the presence of more complex multiplicities containing multiplication?

```hs
(.) :: forall p q. (b *p -> c) *1 -> (a *q -> b) *p -> a *(p*q) -> c

(.)
  :: forall p q
   . (b *p -> c) *1
  -> (a *q -> b) *p
  -> a *(p*q)
  -> c
```

Now, if we are making math analogies here, then really wouldn't `^` be the appropriate operator? After all `a ^3 -o b` converts to `(a * a * a) -o b` a.k.a. `(a, a, a) -o b`: using a value 3 times is like using a triple of that value once. (It's not an isomorphism as it only works this way round: the triple must have thrice the same value.)

So we might write
```hs
(.) :: forall p q. (b ^p -> c) ^1 -> (a ^q -> b) ^p -> a ^(p*q) -> c

-- and vertically...
(.)
  :: forall p q
  .  (b ^p -> c) ^1
  -> (a ^q -> b) ^p
  -> a ^(p*q)
  -> c

newArray# 
  :: Int# ^1
  -> a 
  -> (State# d) ^1
  -> (# State# d, MutableArray# d a #) 
```
I think this looks quite good, plus if you want pretty LaTeX, you could have that too:

Screenshot 2020-09-06 at 00 10 15

I think having a space between the type and the quantity is a bit easier to read (`a ^p -> b` vs `a^p -> b`), but perhaps we could leave that as a matter of good judgment.

int-index:
@buggymcbugfix `a ^p -> b` sounds good to me.

 type
   : btype               { ... }
   | btype '->' ctype    { ... }
+  | btype PREFIX_CARET mult '->' ctype { ... }
   | btype '#->' ctype   { ... }

dorchard:
I agree that using something other than `#`  is a good idea because of the conflation with magic hash, although I find `#` quite please visually even if it makes some examples ugly (like the one shown by @buggymcbugfix ). I also think postfix is a good approach for the reasons @buggymcbugfix explains and `^` is a nice idea re the LaTeXability.

ulysses4ever:
The `^` notation looks really cute to me for the reason described by Vilem.

buggymcbugfix:
@basile-henry pointed out to me that the `a ^p -> b` syntax proposal clashes with the [`^` type family](https://hackage.haskell.org/package/base-4.14.0.0/docs/GHC-TypeLits.html#t:-94-). I don't think this is a deal breaker, but we should think about what to do about this in case we go for this syntax.

int-index:
As I mentioned previously, we should use `PREFIX_CARET` rather than a usual caret here:

* With `LinearTypes` enabled, the users of the `^` type family will be able to write `a ^ b`, `a^b`, and `a^ b`, but not `a ^b`.
* Without `LinearTypes`, nothing changes.

mboes:
Summary of the discussion: there is consensus around the idea that `a #p -> b` won't fly, because of the conflict with overloaded labels, which isn't a problem per say yet, but will be as soon as arbitrary terms show up in types. So we need a different character than `#`, leaving the rest as-is. This syntax conflict affects current GHC `master` too, so we need to agree on a fix before the 9.0 release.

Based on an idea by @int-index, @buggymcbugfix proposes `^` in lieu of `#` above. I updated this PR to reflect that, and I used @int-index's proposed syntax specification. Namely, a multiplicity must be written `^p`, with no intervening spaces. `p` can be an arbitrary base type, with parens required around it if it contains any spaces.

Because this PR addresses a syntax issue currently present in the upcoming GHC 9.0 release, @nomeata I would like to fast track this change to committee for review now. It's not a full proposal. Just an amendment to a previously accepted proposal whose implementation has not yet made it into any release. The proposed amendment has itself changed in response to community feedback, and there appears to be good support.

The one question I would like to ask the committee simultaneously to submitting this amendment is: would you prefer `a ^p -> b` or `a %p -> b` as the syntax for the multiplicity annotated arrow? @buggymcbugfix suggested the former. @int-index initially suggested the latter. I'm happy with either, though my concern with `a ^p -> b` is that many fonts use a small, hard to see glyph for the caret character.

proposals/0111-linear-types.rst
```diff
@@ -265,45 +265,46 @@ Syntax
...
-  multiplicity). This steals syntax as ``(#)`` is a valid
-  type operator. That is the syntax entry for types becomes:
+  ^p -> b`` (where ``a`` and ``b`` are types and ``p`` is
+  a multiplicity). We add a new production to the grammar for
+  ``type``:
 
   ::
 
-    type -> btype [[# btype] -> type]
+    btype PREFIX_CARET btype -> ctype
```
int-index, line 274: That should be `atype` after the `PREFIX_CARET` (or `mult`, which, in turn, should be defined as an `atype`).

int-index:
I implemented this proposal, see the patch here: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4020

Works quite well, and there aren't any conflicts in the grammar. Although I marked it as WIP because it requires a bit more engineering effort to improve error messages when the user writes `a ^1-> b` without `LinearTypes` enabled, and also to get parser API annotations right.

However, I believe it's a good proof-of-concept, demonstrating that the approach proposed here is viable, and the implementation will be available on time before the GHC 9.0 release.

buggymcbugfix:
I made a comparison of `%` vs `^` in different common monospace fonts as well as TeX equations following @mboes suggestion that `^` may be hard to read in some fonts. `^` is actually a bit subtle in the TeX typewriter font, although I think it's still hard to miss.

Including the pdfs below in case anyone wants to have a look:

[linear-percent.pdf](https://github.com/ghc-proposals/ghc-proposals/files/5201378/linear-percent.pdf)
[linear-caret.pdf](https://github.com/ghc-proposals/ghc-proposals/files/5201379/linear-caret.pdf)
[percent-vs-caret.pdf](https://github.com/ghc-proposals/ghc-proposals/files/5201380/percent-vs-caret.pdf)

simonpj:
Personally I think ^ is a bit to quiet, FWIW

buggymcbugfix:
What is happening to unicode syntax? Keeping `a ⊸ b` and allowing something like `a ^1 → b` or `a¹ → b` kind of goes against the proposal's idea of having just one way of expressing linear arrows.

mboes:
> I implemented this proposal, see the patch here: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4020. [...] it's a good proof-of-concept, demonstrating that the approach proposed here is viable, and the implementation will be available on time before the GHC 9.0 release.

@int-index you're a star. Many thanks!

anka-213:
Would exponents on type numerals make things more confusing with this new proposed syntax?
https://hackage.haskell.org/package/base-4.14.0.0/docs/GHC-TypeLits.html#t:-94-

```
type family (m :: Nat) ^ (n :: Nat) :: Nat
```

goldfirere:
I'm happy to shepherd this proposal quickly.

The current diff in this PR uses the `^` syntax, but I understand `%` should also be under consideration by the committee. Is that accurate? Are there other alternatives that should be considered? It would be helpful to e.g. update the original post in this PR to summarize the current state of play for committee review.

mboes:
> The current diff in this PR uses the ^ syntax, but I understand % should also be under consideration by the committee. Is that accurate?

Yes.

> Are there other alternatives that should be considered?

No one has expressed strong support for any other alternative. In the interest of time, and since the stakes are pretty low on this particular question (so long as we avoid conflict with any existing prefix sigil in type- and term-level identifiers), I suggest we keep it to `^` or `%` - though the committee may feel strongly that yet another character is more appropriate.

> It would be helpful to e.g. update the original post in this PR to summarize the current state of play for committee review.

Done.

goldfirere:
I've forwarded this to the committee, but I realized it would be unfair to make a solid recommendation, as I am in conflict with the proposer. That said, I will try to shepherd this to a quick conclusion, as I'm in strong support of sorting this out soon!

nomeata:
JFTR, I briefly considered the conflict, but deemed this proposal to be likely (hopefully) non-controversial enough that “was the previous shepherd” trumps “may be in conflict”.

mboes:
Just to reassure any onlookers: "to be in conflict with" is a technical term from academia meaning that there exists a potential conflict of interest that might colour one's judgement (say when reviewing a paper). @goldfirere and I get along fine. ;)

goldfirere:
> Just to reassure any onlookers: "to be in conflict with" is a technical term from academia meaning that there exists a potential conflict of interest that might colour one's judgement (say when reviewing a paper). @goldfirere and I get along fine. ;)

Oh gosh yes. "conflict" in this way is so standard in the peer-review process that I just used it without thinking. Apologies if anyone misread it! It's all about the fact that a reviewer of a submission (be it a paper for publication or a GHC proposal) should not stand to benefit or be hurt by any particular decision -- and even that it not appear to be the case. Specifically, the peer review process forbids a reviewer to be employed at the same institution as the author, and I extended that rule to this new venue. @mboes and I indeed have no conflict, in the more general sense, with one another. :)

And @nomeata I considered declining shepherding due to the conflict, but I decided that a quick resolution was better, and that I could deliver that quick resolution even without making a recommendation. We're at the murky end of the rules here, but I think this is low-stakes enough that it's all OK.

goldfirere:
This proposal has been accepted, with the choice of `%`, not `^`. Thanks, all!

int-index:
> This proposal has been accepted, with the choice of `%`, not `^`. Thanks, all!

I've updated the implementation accordingly.

PR #357, opened 2020-08-26
Add specializable proposal

fgaz:
We propose a new `SPECIALIZABLE` pragma that has the effect of exposing the
unfolding of the marked binding for specialization purposes, and does not
conflict with inline pragmas.

Part of GSoC 2020

[Rendered](https://github.com/fgaz/ghc-proposals/blob/specializable/proposals/0000-specializable-pragma.rst)

/cc @cartazio @AndreasPK


maralorn:
The two last sections of this proposal currently contain links which are broken in the rendered version of the document.

fgaz:
Thanks, fixed

simonpj:
Under "Motivation" the link about `elem` doesn't lead to anything relevant to elem.

simonpj:
I'm quite supportive of this proposal.  From a user point of view, "SPECIALISABLE" is much more explicit than "INLINEABLE".

That said, I'd love the proposal to give more concrete examples about when the distinction is useful.  In particular,

*  In many cases, the function you want to specialise is recursive (that's why we can't inline it), so you don't need to say NOINLINE.  Examples of when we specifically *don't* want to inline but *do* want to specialise would be useful.   Perhaps a function with rewrite rules, so we want to say "don't inline until phase 0, after than feel free to" (ie NOINLINE[0])?

* If you say INLINE you definitely don't want to specialise, because inlining effectively specialised the function for that single call site.   Or so you might think.  But if you say `map f xs`, then `f` may not inline, even if it has an INLINE pragma, because it's not applied to any arguments; so you might still want to specialise it.

Enumerating all the cases when you might want both a SPECIALISE and a INLINE/NOINLINE pragma would be very helpful motivation.

fgaz:
> Under "Motivation" the link about elem doesn't lead to anything relevant to elem.

Oops, some clipboard mishap happened. fixed it (and also the double inline markup), thank you.

goldfirere:
I would prefer simply to allow `NOINLINE` alongside `INLINEABLE`. That is, we have two pragmas today that, in conjunction, give us exactly the desired behavior. The only problem is that GHC has a rule that forbids using the two together. Just drop that rule!

Sadly, it does look a bit odd to combine `INLINEABLE` and `NOINLINE`. So maybe we rename `INLINEABLE`. (`TRANSPARENT`? `EXPOSE`? `RETAIN_RHS`? I don't like any of these, really.) But I'm against adding a new pragma whose semantics is simply the combination of two existing ones.

simonpj:
> But I'm against adding a new pragma whose semantics is simply the combination of two existing ones.

That's not suggested here. Rather, SPECIALISABLE just means well, what it says. And it's orthogonal to INLINE/NOINLINE.

I don't know any *other* use for INLINEABLE.  (Though I remember @simonmar arguing for it ages ago.) So we could deprecate it, or make it a synonym for SPECIALISABLE.

goldfirere:
You're right. I guess I meant that I don't want a new pragma that means the same as an existing one, less one restriction.

I suppose I'm fine if `INLINEABLE` becomes a (non-preferred) alternate spelling of `SPECIALISABLE`. Is it worth deprecation and churn? I'm not sure it is.

fgaz:
Either way to get the `NOINLINE`+`SPECIALIZABLE` behavior, `SPECIALIZABLE`/`INLINEABLE` _has_ to be a separate pragma from the `INLINE` family, so why not give it a more appropriate name? This also makes it possible to extend it with other specialization-specific stuff later, for example transitive specialization and more importantly phase control (`INLINEABLE`'s phase control affects both inlining and specialization).

simonpj:
> for example transitive specialization and more importantly phase control (INLINE(ABLE)'s phase control affects both inlining and specialization).

Can you say more about both of these points?  You are arguing for some future wiggle room, and that might affect our choices for this proposal.

arybczak:
Will SPECIALIZABLE pragma expose unoptimized or optimized unfolding?

treeowl:
@simonpj, doesn't `INLINABLE` expose the unfolding for `GHC.Magic.inline` as well as for specialization? Sometimes you really do want to inline something very large in one or two particular places.

simonpj:
> @simonpj, doesn't INLINABLE expose the unfolding for GHC.Magic.inline as well as for specialization?

I was puzzled about what you mean, but I think you intend this
```
module M where
  f = <big rhs>
  {-# INLINABLE f #-} 

module A where
  import M( f )
  import GHC.Magic( inline )

  bar = ...(inline f x)....
```
Then yes, f will be inlined at its call site in A (provided it is non-recursive)    But if you didn't have an INLINABLE pragma, module M might not expose an unfolding for f at all, since the RHS is big.

I imagine that SPECIALISABLE would (unless we stop it) have this behaviour too.  Inlining is just call-site-specific version of specialisation, after all

cartazio:
so i feel like perhaps some of the confusion might be best addressed by having a table of how things work 

@simonpj  would something like this (filled out suitably) help with reducing the confusion / faciliate communication?

 property/pragma        |   always exposes optimized unfolding |  always unopt unfolding | may inline  | always inlines | may inline specialization 
---------------------- | ----------------------------- | --------------------------------- | -------- | -------------- |  ------- | 
Specialize                   |   
Inline                          |    ?                             | ?                                      
NoInline                    | no
No such pragma      | ?                                    |  ?                               | yes, per optimizer | depends? | depends? | 
Inlineable                 |
Specializable            | 
 Specializable Inline | 
Specialiable Noinline | ? | ? | no

cartazio:
or fliped around another way we may wanna write down a product of "ways we inline" and "ways we specialize" and talk about which combinations of those features are expressible as pragmas today 

eg as a starting point for this discussion, (which fgaz should fact chat before using / adapting to the proposal) 


**source module** where its defined
**client module** where a definition is used, via imports etc

_ | No Unfolding                   | Optimized/default unfolding may inline | unoptimized unfolding may inline | unfolding in some form but cant be used for inlinining | WILL inline always  |
--- |   ---------------------- |  ------------------ | ----------------------- | -----------------------------------  | ---------- 
can specialize in **source** module |       specialize + noinline in source module      |      specializeable        |      inlineable ?               |   specializable noinline  |   inline
can **NOT** specialize in a **client** module |  noinline |        no pragma   in source module  / INLINE (?)  | ?      |  specialize pragma in source module |    inline                  
can specialize in **client** module | impossible | specializable | inlineable ? | specializable noinline | ?

cartazio:
i guess the `will inline` and `can specialize` sortah  maps to `specialize inline`

cartazio:
NB: i think i'm slightly conflating "can i write a valid specalization pragma" and "will code specialize" in this table, but i still think its a good starting point ?

sgraf812:
Will SPECIALISABLE store optimised unfoldings like INLINABLE or the original RHS like INLINE? Why? I think that's missing in the proposal.

sgraf812:
In general, I think it may make sense to specify whether or not the original RHS should be inlined, but that's a separate proposal.

cartazio:
@sgraf812 
```hs

         -> (gbl_id `setIdUnfolding` inlinable_unf, rhs) -- TODO review it
```
at https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3630/diffs#65e2067fbb7377818b6cadff0531f44029499948_390_388 , so atm, I think it mimics the logic of INLINABLE for its RHS unfolding, if i'm reading the diff correctly. to be clear, as the comment in the current MR states, thats open to discussion / feedback

cartazio:
the full snippet is 

```haskell

  = case (inlinePragmaSpec inline_prag, specializable_prag) of
          (NoUserInline, NoUserSpecializable) -> (gbl_id, rhs)
          (NoUserInline, Specializable      )
            -> (gbl_id `setIdUnfolding` inlinable_unf, rhs) -- TODO review it
          (NoInline,     NoUserSpecializable) -> (gbl_id, rhs)
          (NoInline,     Specializable      )
            -> (gbl_id `setIdUnfolding` specializable_noinline_unf, rhs)
          (Inlinable,    _                  )
            -> (gbl_id `setIdUnfolding` inlinable_unf, rhs)
          (Inline,       _                  ) -> inline_pair
```

fgaz:
> Can you say more about both of these points? You are arguing for some future wiggle room, and that might affect our choices for this proposal.

Sorry, I've been busier than I thought. Here's some more about it:

##### Transitive specialization

https://gitlab.haskell.org/ghc/ghc/-/issues/12463#transitive-specialisation

Basically, treat polymorphic bindings where a specializable binding is used as specializable themselves

##### Phase control

Currently [phase control](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#phase-control) can be used [to specify the specialization activation for `SPECIALIZE`](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#specialize-pragma), but there's no way to directly do that for automatically specializations, neither with `SPECIALIZABLE` nor with `INLINEABLE`.

Right now (I don't think this is documented anywhere outside of the source) the specialization's activation is inherited from the `INLINE`/`NOINLINE`/`INLINEABLE`'s one, so there's no way to specify it separately, or to specify it altogether without being forced to use an inline pragma.

---

Also I guess I might as well undraft this pr

fgaz:
As @cartazio says, I conservatively used the same type of unfolding as inlineable (which does receive some light optimization iirc), but open to discussion.
So for now the two "may inline" columns should probably be coalesced

cartazio:
@fgaz  like this? (again, please view this table of mine as by no means authoritative, but as an attempt to synthesize my own understandings, hopefully a more authorizative one can be eventually in the final version of this proposal and perhaps as an addition to the ghc manual on these pragmas?)

_ | No Unfolding | may inline | unfolding in some form but cant be used for inlinining | WILL inline always
-- | --- | -- | -- | --
can specialize in source module | specialize + noinline in source module | specializeable or  inlineable ? | specializable noinline | inline
can NOT specialize in a client module | noinline | no pragma in source module / INLINE (?)  | specialize pragma in source module | inline
can specialize in client module | impossible | specializable or inlineable ? | specializable noinline | specialize inline

simonpj:
I think this proposal is best seen as two separate proposals.

1.  Allow NOININE and NOINLINE[n] to be specified at the same time as INLINABLE

2.  Introduced SPECIALISABLE as a synonym for INLINABLE. Maybe even deprecate the latter.

I agree with Richard that we don't want SPECIALISABLE to mean something subtly different from INLINABLE. I see no motivation for doing so.

Under the above, you can mark a function as NOINLINE or NOINLINE[1], say, and simultaneously SPECIALISABLE (or equivalently INLINABLE).  The SPECIALISABLE pragma means that call sites (in this or other modules) can see its original definition (via the exposed unfolding), and can specialise it.

If the function is marked INLINE or INLINE[n], adding INLINABLE/SPECIALISABLE is a no-op.  The unfolding is exposed anyway.
Whether we want to make it illegal to co-specify these is a design choice. It could just silently be a no-op.  But I think it's
an error today.

I think that's all.  It's pretty simple.  Is that indeed what you intend?  Or have I misunderstood?

phadej:
@arybczak question

> Will SPECIALIZABLE pragma expose unoptimized or optimized unfolding?

Is still not answered in the proposal nor discussion.

IMO we need a way to expose **optimized** unfolding. `INLINEABLE` expose unoptimized one, which is bad, as it the specialized code is reoptimized over and over.

`-fexpose-all-unfoldings` (which exposes optimized unfoldings) is the only  way to expose optimized unfoldings.

Proposal SHOULD be explicit and say "exact RHS" or "optimized" or "non-optimized", currently it's ambiguous. And probably will be the "exact RHS" (as that's what INLINEABLE does), which is bad, and doesn't solve the problems raised in motivation section.

---

The GHC manual says

> Like INLINE, the INLINABLE pragma retains a copy of the original RHS for inlining purposes, ...

---

IN summary, I don't agree that `INLINEABLE` is not expressive enough

PR #358, opened 2020-08-28
Clarify language in the proposal template

aspiwack:
This is an attempt at clarifying the language in the template a tiny bit. To try and pre-empt some of the advice that the steering committee find itself giving often. There may be other advice that I missed, so don't hesitate to bring points to clarify. I'll make attempts.

I've only modified the rst template for the review. I'll port to the markdown template when we agree on the language.
- [x] Port to markdown.

### DIFF
--- a/proposals/0000-template.rst
+++ b/proposals/0000-template.rst
@@ -52,12 +52,23 @@ may include,
 * how the proposed change interacts with existing language or compiler
   features, in case that is otherwise ambiguous
 
-The ideal specification is described as a modification of the
-`Haskell 2010 report <https://www.haskell.org/definition/haskell2010.pdf>`_.
+Strive for *precision*. The ideal specification is described as a
+modification of the `Haskell 2010 report
+<https://www.haskell.org/definition/haskell2010.pdf>`_. Where that is
+not possible (e.g. because the specification relates to a feature that
+is not in the Haskell 2010 report), try to adhere its style and level
+of detail. Think about corner cases. Write down general rules and
+invariants.
 
-Note, however, that this section need not describe details of the
-implementation of the feature or examples. The proposal is merely supposed to
-give a conceptual specification of the new feature and its behavior.
+Note, however, that this section should focus on a precise
+*specification*; it need not (and should not) devote space to
+*implementation* details -- there is a separate section for that.
+
+The specification can, and almost always should, be illustrated with
+*examples* that illustrate corner cases. But it is not sufficient to
+give a couple of examples and regard that as the specification! The
+examples should illustrate and elucidate a clearly-articulated
+specification that covers the general case.
 
 Examples
 --------
--- a/proposals/0000-template.rst
+++ b/proposals/0000-template.rst
@@ -92,8 +92,8 @@ the reasons for choosing the proposed change over these alternative:
 *e.g.* they can be cheaper but insufficient, or better but too
 expensive. Or something else.
 
-The PR discussion often raises other potential designs, they should be
-added to this section. And, similarly, if the proposed change
+The PR discussion often raises other potential designs, and they should be
+added to this section. Similarly, if the proposed change
 specification changes significantly, the old one should be listed in
 this section.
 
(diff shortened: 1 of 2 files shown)


nomeata (approved):
LGTM

proposals/0000-template.rst
```diff
@@ -52,6 +52,9 @@ may include,
 * how the proposed change interacts with existing language or compiler
   features, in case that is otherwise ambiguous
 
+The ideal specification is described as a modification of the
+`Haskell 2010 report <https://www.haskell.org/definition/haskell2010.pdf>`_.
```
goldfirere, line 56: Yes, but I think this is too high a bar. In particular, a good deal of proposals (maybe the majority) are amendments to existing extensions, which are usually not very finely documented.

So maybe this suggests:

```suggestion
The ideal specification is described as a modification of the
`Haskell 2010 report <https://www.haskell.org/definition/haskell2010.pdf>`_.
However, that is not always possible, if the feature you are building from is not part of Haskell 2010 or if you cannot see how to modify the report to include your feature.
```

proposals/0000-template.rst
```diff
@@ -83,9 +86,16 @@ drawbacks that cannot be resolved.
...
-List existing alternatives to your proposed change as they currently exist and
-discuss why they are insufficient.
-
+List alternative designs to your proposed change. Both existing
+workarounds, or alternative choices for the changes. Explain
+the reasons for choosing the proposed change over these alternative:
+*e.g.* they can be cheaper but insufficient, or better but too
+expensive. Or something else.
+
+The PR discussion often raises other potential designs, they should be
```
goldfirere, line 95:
```suggestion
The PR discussion often raises other potential designs, and they should be
```

proposals/0000-template.rst
```diff
@@ -83,9 +86,16 @@ drawbacks that cannot be resolved.
...
+added to this section. And, similarly, if the proposed change
```
goldfirere, line 96:
```suggestion
added to this section. Similarly, if the proposed change
```

proposals/0000-template.rst
aspiwack: That was the meaning that I intended of the word “ideal”. I'm not sure we should qualify this sentence too much, lest it be unconsciously ignored.

goldfirere (approved):

proposals/0000-template.rst
```diff
@@ -52,6 +52,9 @@ may include,
 * how the proposed change interacts with existing language or compiler
   features, in case that is otherwise ambiguous
 
+The ideal specification is described as a modification of the
+`Haskell 2010 report <https://www.haskell.org/definition/haskell2010.pdf>`_.
+
 Note, however, that this section need not describe details of the
 implementation of the feature or examples. The proposal is merely supposed to
 give a conceptual specification of the new feature and its behavior.
```
simonpj, line 60: replace "conceptual" with "precise".   Also "need not" should be "should not".  And we should distinguish examples.  How about this?

"Note, however, that this section should focus on a precise *specification*; it need not (and should not) devote space to *implementation* details -- there is a separate section for that.

"The specification can, and almost always should, be illustrated with *examples* that illustrate corner cases.  But it is not sufficient to give a couple of examples and regard that as the specification!  The examples should illustrate and elucidate a clearly-articulated specification that covers the general case."

proposals/0000-template.rst
simonpj: How about this?

"Strive for *precision*.   The ideal specification is described as a modification of the Haskell 2010 report. Where that is not possible (e.g. because the specification relates to a feature that is not in the Haskell 2010 report), try to adhere its style and level of detail.   Think about corner cases.   Write down general rules and invariants.  "

aspiwack:
Ok, I've ported to markdown. We don't have a way to generate the markdown from the rst that I could find, so I had to manually port. I triple-checked my changes, so I'm pretty confident. It did require that I find a bit of focus time on this, hence me merging only now.

Thanks everyone for your feedback.

PR #360, opened 2020-09-01
Add committee bylaws

goldfirere:
A meeting of the committee at ICFP led to an understanding that we need to codify our practices a bit. This is, thankfully, not a result of anything having gone wrong, but an agreement that our continued success requires a well-articulated foundation.

This PR is a stab at drafting bylaws. The committee did *not* discuss particulars, which are my own invention, and about which I expect suitable debate.

Note that this PR is set up to allow edits by the committee members; please feel free to edit my branch directly, especially with typos and formatting goofs.

[Rendered bylaws](https://github.com/goldfirere/ghc-proposals/blob/bylaws/committee.rst), but there are a few other changes with this PR as well.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -4,7 +4,7 @@ GHC Proposals
 This repository contains specifications for proposed changes to the
 `Glasgow Haskell Compiler <https://www.haskell.org/ghc>`_.
 The purpose of the GHC proposal process and of
-the GHC Steering Committee, is to broaden the discussion of the evolution of
+the GHC Steering Committee is to broaden the discussion of the evolution of
 GHC.
 
 * `≡ List of proposals under discussion <https://github.com/ghc-proposals/ghc-proposals/pulls?q=is%3Aopen+is%3Apr+no%3Alabel>`_
@@ -14,7 +14,7 @@ GHC.
 * `≡ List of rejected proposals <https://github.com/ghc-proposals/ghc-proposals/pulls?q=is%3Apr+label%3A%22Rejected%22>`_
 * `≡ List of implemented proposals <https://github.com/ghc-proposals/ghc-proposals/pulls?q=is%3Apr+label%3A%22Implemented%22>`_
 * `≡ List of all proposals <https://github.com/ghc-proposals/ghc-proposals/pulls?q=>`_
-
+  
 What is a proposal?
 -------------------
 
@@ -250,7 +250,8 @@ Alejandro Serrano       `@serras <https://github.com/serras/>`_               20
 Arnaud Spiwack          `@aspiwack <https://github.com/aspiwack/>`_           2019/07
 ======================  ====================================================  =======  =========
 
-The committee members have committed to adhere to the `Haskell committee guidelines for respectful communication <GRC.rst>`_.
+The committee members have committed to adhere to the `Haskell committee guidelines for respectful communication <GRC.rst>`_ and are subject to the
+`committee bylaws <https://github.com/ghc-proposals/ghc-proposals/blob/master/committee.rst>`_.
 
 We would also like to thank our former members:
 
@@ -266,6 +267,8 @@ Christopher Allen       `@bitemyapp <https://github.com/bitemyapp>`_          20
 
 Committee process
 -----------------
+
+The `committee bylaws <https://github.com/ghc-proposals/ghc-proposals/blob/master/committee.rst>`_ are posted.
 
 The committee process starts once the secretary has been notified that a
 proposal is ready for decision.
@@ -395,33 +398,9 @@ is a polite ping/enquiry.
       proposal with the implementation status (i.e. ticket URL and the
       first version of GHC implementing it.)
 
-      Concretely, a committee member must perform the following steps:
-
-      1. Add a new commit on top of the PR branch that:
-
-         a. Changes the filename of the proposal to correspond to the PR number.
-
-         b. Updates any metadata fields that may have changed in the template on ``master`` since
-            the PR branch split off.
-
-         c. Fills in these metadata fields as appropriate, including changing "is discussed"
-            to "was discussed".
-
-      2. Merge the PR branch into master, and push.
-
-      3. Update the PR description to start
-         with the text "The proposal has been accepted; the following discussion is mostly of historic interest."
-         where the word "proposal" links to the final rendered version pushed above.
-
-      4. If the PR title has "(under review)", remove it.
-         
-      5. Set the PR to have the "Accepted" label.
-
-      6. Comment on the PR that the proposal was accepted.
-
-      7. Close the PR if GitHub has not detected the merge.
-
-      8. Announce on the committee mailing list.
+      Committee members should see the `acceptance page <https://github.com/ghc-proposals/ghc-proposals/blob/master/acceptance.rst>`_ for a checklist
+      to be applied to accepted proposals and the steps necessary in
+      order to mark a proposal as accepted.
 
 Review criteria
 ---------------
--- a/README.rst
+++ b/README.rst
@@ -265,10 +265,8 @@ Christopher Allen       `@bitemyapp <https://github.com/bitemyapp>`_          20
 ======================  ====================================================  =================
 
 
-Committee process
------------------
-
-The `committee bylaws <https://github.com/ghc-proposals/ghc-proposals/blob/master/committee.rst>`_ are posted.
+Committee process for responding to a proposal
+----------------------------------------------
 
 The committee process starts once the secretary has been notified that a
 proposal is ready for decision.
--- a/acceptance.rst
+++ b/acceptance.rst
@@ -1,0 +1,63 @@
+How to Accept a Proposal
+========================
+
+This document describes the technical process for accepting a proposal,
+assuming the proposal has gone through the `proposal process <https://github.com/ghc-proposals/ghc-proposals/#committee-process>`_.
+
+Acceptance checklist
+--------------------
+
+Before accepting, the shepherd should check for the following
+(extracted from the `review criteria <https://github.com/ghc-proposals/ghc-proposals/#review-criteria>`_):
+
+1. Is the proposal self-standing? It should not refer explicitly to the GitHub
+   trail, for instance.
+
+2. Is the proposal precise? The specification section should fully describe
+   the proposal payload. In particular:
+
+   a. If there is new syntax, is it described in BNF notation?
+
+   b. If applicable, are there typing rules? If there are not, could
+      typing rules be reasonably extracted from the proposal text?
+
+   c. Are any changes to runtime behavior precisely specified?
+
+   d. Is there a reasonable description of how the new feature might
+      be documented in Haddock? Linted by HLint? Treated by other tools?
+      (These will not apply to every proposal.)
+
+3. Is the proposal sufficiently illustrated with examples? Examples help
+   us understand the details that will be needed during implementation.
+
+Acceptance steps
+----------------
+
+A committee member accepts a proposal by following this sequence of
+steps:
+
+1. Add a new commit on top of the PR branch that:
+
+   a. Changes the filename of the proposal to correspond to the PR number.
+
+   b. Updates any metadata fields that may have changed in the template on ``master`` since
+      the PR branch split off.
+
+   c. Fills in these metadata fields as appropriate, including changing "is discussed"
+      to "was discussed".
+
+2. Merge the PR branch into master, and push.
+
+3. Update the PR description to start
+   with the text "The proposal has been accepted; the following discussion is mostly of historic interest."
+   where the word "proposal" links to the final rendered version pushed above.
+
+4. If the PR title has "(under review)", remove it.
+   
+5. Set the PR to have the "Accepted" label.
+
+6. Comment on the PR that the proposal was accepted.
+
+7. Close the PR if GitHub has not detected the merge.
+
+8. Announce on the committee mailing list.
--- a/committee.rst
+++ b/committee.rst
@@ -14,7 +14,7 @@ who wish to volunteer for this service. It is our aim that this committee
 be diverse; by representing different viewpoints, we will make decisions
 that benefit larger segments of our community.
 
-Specifically, we ideally would be able to represent the following
+Specifically, we would like the committee to represent at least the following
 constituencies:
 
 * GHC developers
@@ -30,13 +30,13 @@ choosing new members, we aim to correct any imbalances.
 Committee Roles
 ---------------
 
-* The committee has two co-chairs. This role is more ceremonial
+* **Co-chairs.** The committee has two co-chairs. This role is more ceremonial
   than practical, as all members of the committee have the right
   to set the agenda and move our business forward. The co-chairs
-  are the stop of last resort for any issues of contention, but
-  are otherwise undistinguished from other members of the committee.
+  are the stop of last resort for any issues of contention, but are
+  not otherwise distinguished from other members of the committee.
 
-  Agreement among the co-chairs can initiate a process to force a member off
+  Agreement among the co-chairs can initiate a process to remove a member from
   the committee. If this happens, the co-chairs must explain their rationale
   to the entire committee, including the member that has been removed. In
   order to avoid public embarrassment, it is expected that this communication
@@ -46,10 +46,10 @@ Committee Roles
   expectation is that if the co-chairs abuse their power, the committee will
   vote them out of their chairships.
 
-* The committee has one secretary. The secretary is the first stop for keeping
+* **Secretary.** The committee has one secretary. The secretary is the first stop for keeping
   our repository up-to-date and running our selection process.
 
-* The committee has one nudger. The nudger's job is to make sure other
+* **Nudger.** The committee has one nudger. The nudger's job is to make sure other
   committee members are fulfilling their duties by nudging them (often
   in public) to do so.
 
@@ -107,6 +107,16 @@ term is ending is free to re-nominate themselves. Nominations include a brief
 bio of the nominee and reasons why they would be appropriate for the
 committee.
 
+The main criteria for becoming a member of the committee are:
+
+ * A track record of constructive contribution to public discussion of GHC proposals
+ * A track record of constructive contribution to one or more of the communities outlined above (users, tool authors etc)
+ * An expressed willingness to respond to GHC proposals in a timely manner.
+
+These criteria are not exhaustive -- nominations and self-nominations are free
+to strengthen the case in whatever way they deem appropriate --- but a record
+of thoughtful contribution in a public space carries the most weight.
+
 Continuing members of the committee vote on new membership via a ranked voting
 system; the precise details (i.e. election algorithm) are recommended by the
 secretary but can be vetoed by a majority vote. The ranking system includes a
@@ -118,7 +128,7 @@ rationale for this decision be published to the community.
 The voting process may result in a number of new members not equal to
 the number of outgoing members. This is fine; the size of the committee
 is not fixed. However, a new nomination process begins only when
-membership drops below nine members.
+membership is expected to drop below nine members.
 
 The nomination and voting process is kept private, by using direct
 email to committee members, not the mailing list. When the secretary's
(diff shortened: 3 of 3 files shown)


committee.rst
```diff
@@ -0,0 +1,152 @@
...
+secretary but can be vetoed by a majority vote. The ranking system includes a
+method for choosing multiple winners. It is expected that the committee
+favor newcomers in preference to re-nominated continuing members. If
+an expiring member continues, the secretary shall ensure that a
+rationale for this decision be published to the community.
+
+The voting process may result in a number of new members not equal to
+the number of outgoing members. This is fine; the size of the committee
+is not fixed. However, a new nomination process begins only when
+membership drops below nine members.
```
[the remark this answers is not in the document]
gridaphobe: Also, I believe our goal was to keep the committee size around 10 members, so this last sentence should read

> However, a new nomination process begins only when membership drops below **ten** members.

README.rst
```diff
@@ -395,33 +398,9 @@ is a polite ping/enquiry.
...
-         
-      5. Set the PR to have the "Accepted" label.
-
-      6. Comment on the PR that the proposal was accepted.
-
-      7. Close the PR if GitHub has not detected the merge.
-
-      8. Announce on the committee mailing list.
+      Committee members should see the `acceptance page <https://github.com/ghc-proposals/ghc-proposals/blob/master/acceptance.rst>`_ for a checklist
+      to be applied to accepted proposals and the steps necessary in
```
phadej, line 402: Is this *accepted* meant to be *submitted*?

AntC2:
>  This is, thankfully, not a result of anything having gone wrong, ...

(I appreciate this isn't really the place to discuss the below, but I feel the the Committee has gone wrong; and one of the wrong things is there's no venue for feedback.)

The Committee mailing list has just started a discussion about Haskell 2020. There was a separate Haskell Prime committee and process supposed to be doing that; but it has been moribund all this year. I rather doubt that the _GHC_ Committee are the right people/priorities to discuss the Haskell standard -- even granted that GHC represents a de facto standard.

@goldfirere has started a github page to record/discuss decisions. How do I contribute to that? I'm asking because there's two screamingly wrong things I see on it already.

gridaphobe:
> The Committee mailing list has just started a discussion about Haskell 2020.

The discussion we just started was about a **GHC** 2020 "standard", which would just be a set of existing language extensions that can be enabled as one. We have no desire to take over the Haskell standard, but there have been many requests to reduce the boilerplate of repeatedly enabling all the myriad extensions people use, and a GHC 2020 standard is one solution that was floated recently and had broad support from the committee.

> @goldfirere has started a github page to record/discuss decisions. How do I contribute to that?

I agree that a wiki is not the best way to discuss which extensions belong in the hypothetical GHC 2020 standard, this concern was also raised on the list. One idea I just had was to create a separate repo in the `ghc-proposals` org and use github issues to discuss extensions individually, with an issue per nominated extension. This would preserve all the useful features of github discussions that have made the Proposals process so successful, without requiring another mega-proposal. The alternative that was originally discussed was to just issue a request for GHC 2020 proposals and use the standard process to select one, but I wonder if that won't become unwieldy when people inevitably get into protracted discussions about particular extensions. But regardless of the specific process, we're not going to unilaterally create a GHC 2020 standard. The goal is to get the same level of community involvement as with any regular proposal.

AntC2:
Thanks @gridaphobe, and for your post to the Committee list.

> ... unwieldy when people inevitably get into protracted discussions about particular extensions.

I agree we don't want protracted discussions. So you don't want Uncle Tom Cobbley and all editting the wiki page. The decision should be only: is such-and-such an extension as currently implemented in GHC to be in or out? Thank you to Richard for the explicit criteria 1-5. I think those are about right, but possibly need tightening up because they allow too much room for interpretation/subjectivity. Specifically ...

So what I see as wrong is there's two extensions, as currently implemented, that RAE evaluates as meeting his criteria 1-5; but which I evaluate as failing on at least 3 of those criteria.

I'll mention there are other compilers for Haskell than GHC; and that at least one of them implements both those extensions; and implements them differently (in dark corner cases [**]); and indeed had already implemented them well before the H2010 standard. It was precisely because they fail on the sort of criteria RAE suggests that neither extension made it into H2010.

[**] In those corner cases, my experience is that Hugs' implementation is better-disciplined than GHC's, and rejects programs that can lead GHC into type-incoherence. If we were to get into protracted discussions, I'd say Hugs is too conservative, GHC is too laissez-faire.

goldfirere:
To be clear: I never noticed that the wiki wasn't world-editable. I thought it was. Regardless, the conversation about GHC2020 really does belong elsewhere, and it seems the committee is looking at figuring out where. May I suggest that we return this thread back to the proposed bylaws?

acceptance.rst
```diff
@@ -0,0 +1,63 @@
+How to Accept a Proposal
+========================
+
+This document describes the technical process for accepting a proposal,
+assuming the proposal has gone through the `proposal process <https://github.com/ghc-proposals/ghc-proposals/#committee-process>`_.
+
+Acceptance checklist
+--------------------
+
+Before accepting, the shepherd should check for the following
```
aspiwack, line 10: Also: before sending the proposal to the committee to review.

In my view, the first job of the shepherd, when handed a proposal, is to help make sure that the checks below pass. This means both ensuring that the conditions are met before sending the proposal over to the committee, and, when they aren't, to help guide the proposal's authors towards meeting said conditions.

goldfirere: Yes, but it's very easy at that stage to focus on the substance of the proposal, instead of these details. And I personally think that's OK. But we should make sure the details are attended to in the end.

aspiwack: I guess what I'm getting at is that we should use this opportunity to set the expectations that we have for the shepherd. I'm guessing that we don't have the same image of the role. When I'm shepherding, I try to suspend my judgement as much as possible and focus on ensuring that the proposal is the best proposal it can be, regardless of whether I'm going to recommend acceptance or rejection.

committee.rst
```diff
@@ -0,0 +1,152 @@
...
+
+Committee Composition
+---------------------
+
+The committee is formed of roughly 10 members of the Haskell community
+who wish to volunteer for this service. It is our aim that this committee
+be diverse; by representing different viewpoints, we will make decisions
+that benefit larger segments of our community.
+
+Specifically, we ideally would be able to represent the following
```
simonpj, line 17: Specifically, we would like the committee to represent at least the following constituencies:

committee.rst
```diff
@@ -0,0 +1,152 @@
...
+choosing new members, we aim to correct any imbalances.
+
+Committee Roles
+---------------
+
+* The committee has two co-chairs. This role is more ceremonial
+  than practical, as all members of the committee have the right
+  to set the agenda and move our business forward. The co-chairs
+  are the stop of last resort for any issues of contention, but
+  are otherwise undistinguished from other members of the committee.
```
simonpj, line 37: but are not otherwise distinguished from ...

committee.rst
```diff
@@ -0,0 +1,152 @@
...
+
+  Agreement among the co-chairs can initiate a process to force a member off
```
simonpj, line 39: a process to remove a member from the committee

committee.rst
```diff
@@ -0,0 +1,152 @@
...
+* Researchers
+* Tooling developers
+
+We recognize that we may not always live up to this ideal, but when
+choosing new members, we aim to correct any imbalances.
+
+Committee Roles
+---------------
+
+* The committee has two co-chairs. This role is more ceremonial
```
simonpj, line 33: **Co-chairs**.   The committee has two co-chairs. ...

committee.rst
```diff
@@ -0,0 +1,152 @@
...
+  the committee. If this happens, the co-chairs must explain their rationale
+  to the entire committee, including the member that has been removed. In
+  order to avoid public embarrassment, it is expected that this communication
+  be by private email, not via a mailing list of public record. (The chairs
+  may choose, however, to use public channels, in egregious circumstances.)
+  The committee does not need to weigh in during this process, though the
+  expectation is that if the co-chairs abuse their power, the committee will
+  vote them out of their chairships.
+
+* The committee has one secretary. The secretary is the first stop for keeping
```
simonpj, line 49: ** Secretary**.  The committee has ...

committee.rst
```diff
@@ -0,0 +1,152 @@
...
+Any member of the committee is free to step down at any time. Doing so
+causes a nomination process to begin (if the resulting size of the committee
+would be less than nine).
+
+There are two exceptions to the three-year term rule for two "key members"
+of the committee: in recognition of
+their historical and current importance to GHC, both Simon Peyton Jones
+and Simon Marlow will be expected to be retained on the committee when
+their terms end. If either wishes to continue serving on the committee
+when their terms end, and with the support of the majority of the committee,
```
simonpj, line 137: terms end, and if the majority of the rest of the committee supports this outcome, no public nomination process...

committee.rst
```diff
@@ -0,0 +1,152 @@
...
+no public nomination process needs to take place to replace them.
+
+These key members can be stripped of their status as key members by a
+majority vote of the committee, and other individuals can be made into
+key members by a unanimous decision of the committee. In both cases,
+changes to the list of key members will be accompanied by a public
+rationale.
+
+Note that there is no process for members of the public at large to
```
simonpj, line 146: Remove "note that"

committee.rst
```diff
@@ -0,0 +1,152 @@
...
+
+Any community thrives best by continuing to refresh itself with new members.
+Accordingly, membership on the committee comes with a three-year term. Terms
+begin/end in October. Toward the end of an individual's term, the secretary of
+the committee will put out a public call for nominations for people to join
+the committee. Self-nominations are welcome and expected. The member whose
+term is ending is free to re-nominate themselves. Nominations include a brief
+bio of the nominee and reasons why they would be appropriate for the
+committee.
+
```
simonpj, line 109: The main criteria for becoming a member of the committee are:
* A track record of constructive contribution to public discussion of GHC proposals
* A track record of constructive contribution to one or more of the communities outlined above (users, tool authors etc)
* An expressed willingness to respond to GHC proposals in a timely manner.

These criteria are not exhaustive -- nominations and self-nominations are free to strengthen the case in whatever way they deem appropriate --- but a record of thoughtful contribution carries most weight.
*

README.rst
```diff
@@ -267,6 +268,8 @@ Christopher Allen       `@bitemyapp <https://github.com/bitemyapp>`_          20
 Committee process
 -----------------
 
+The `committee bylaws <https://github.com/ghc-proposals/ghc-proposals/blob/master/committee.rst>`_ are posted.
```
simonpj, line 271: What does this sentence mean?  Maybe "Here are the committee bylaws"?   

But in any case that seems irrelevant to the process of dealing with a proposal

README.rst
```diff
@@ -267,6 +268,8 @@ Christopher Allen       `@bitemyapp <https://github.com/bitemyapp>`_          20
 Committee process
```
simonpj, line 268: Committee process for responding to a proposal

README.rst
goldfirere: Agreed. I've dropped this line; the bylaws are linked elsewhere.

acceptance.rst
```diff
@@ -0,0 +1,63 @@
...
+
+   a. Changes the filename of the proposal to correspond to the PR number.
+
+   b. Updates any metadata fields that may have changed in the template on ``master`` since
+      the PR branch split off.
+
+   c. Fills in these metadata fields as appropriate, including changing "is discussed"
+      to "was discussed".
+
+2. Merge the PR branch into master, and push.
```
nomeata, line 49: I usually merge master into the PR, then update the metadata, then push to master. But we probably don’t care too much about the various ways of achieving the same result, so no change needed.

acceptance.rst
```diff
@@ -0,0 +1,63 @@
...
+
+3. Update the PR description to start
+   with the text "The proposal has been accepted; the following discussion is mostly of historic interest."
+   where the word "proposal" links to the final rendered version pushed above.
```
nomeata, line 53:
```suggestion
   where the word "proposal" links to the final rendered version, as found on https://github.com/ghc-proposals/ghc-proposals/tree/master/proposals
```

committee.rst
```diff
@@ -0,0 +1,152 @@
...
+  be by private email, not via a mailing list of public record. (The chairs
+  may choose, however, to use public channels, in egregious circumstances.)
+  The committee does not need to weigh in during this process, though the
+  expectation is that if the co-chairs abuse their power, the committee will
+  vote them out of their chairships.
+
+* The committee has one secretary. The secretary is the first stop for keeping
+  our repository up-to-date and running our selection process.
+
+* The committee has one nudger. The nudger's job is to make sure other
```
nomeata, line 52: I guess this is worth an experiment

committee.rst
```diff
@@ -0,0 +1,152 @@
...
+membership drops below nine members.
+
+The nomination and voting process is kept private, by using direct
+email to committee members, not the mailing list. When the secretary's
+term is expiring (but that individual wishes to stay on the committee),
+one co-chair fills in for the secretary to run this process.
+
+Any member of the committee is free to step down at any time. Doing so
+causes a nomination process to begin (if the resulting size of the committee
+would be less than nine).
```
nomeata, line 130:
```suggestion
Any member of the committee is free to step down at any time. Doing so
causes a nomination process to begin, if the resulting size of the committee
would be less than nine.
```

committee.rst
```diff
@@ -0,0 +1,152 @@
...
+most likely pertaining to individuals. In particular, discussions of selecting
+new members and of violations of the guidelines for respectful communication
+will not appear on the mailing list.
+
+Term Limits and Committee Selection
+-----------------------------------
+
+Any community thrives best by continuing to refresh itself with new members.
+Accordingly, membership on the committee comes with a three-year term. Terms
+begin/end in October. Toward the end of an individual's term, the secretary of
```
nomeata, line 103:
```suggestion
Accordingly, membership on the committee comes with a three-year term. 
Toward the end of an individual's term, the secretary of
```

It seems that later we say we call for nominations when we drop below 9. Doesn’t that naturally lead to more organic start/end of terms?

goldfirere: That's a good question. I like the idea of spacing out when terms expire, so that we have a regular cadence of comings and goings. If someone resigns early, does their replacement get a 3-year term when they arrive? or do they fill out the remainder of the resigner's term? It's not clear to me what's best.

nomeata: Seems clear to me: a full three year term. Why should a new member be treated differently based on the event that triggered the nomination…

goldfirere: Because staggering ends of terms has a countervailing benefit to treating members equally without regard to when they step in.

I honestly don't feel strongly -- and you seem to -- but I'd like one other voice to speak up on this point before updating the text.

nomeata: > Because staggering ends of terms has a countervailing benefit to treating members equally without regard to when they step in.

Sorry, I don’t understand what you are saying

goldfirere: There is a tension between two nice attributes of a the committee:

1. Terms end in a staggered fashion, with roughly 1/3 of the committee expiring every year, on a predictable schedule.
2. Every member selected to the committee is granted a 3-year term, regardless of the conditions around their nomination.

Both of these are nice things to have. However, I think we cannot have both. In having members who replace early resigners fill out the unfinished term, we keep property (1). In having all members start with a 3-year term, we keep property (2), but we may slowly lose (1), depending on when people resign.

I originally phrased the bylaws to implicitly favor (1), whereas you are advocating for (2).

Does this clarify?

bravit: I agree with Ioachim. A full three-year term sounds reasonable. It's relatively cheap to track ends of terms.

nomeata: Yes, I don’t really care about (1). I expect that a fair fraction of terms will expire earlier than three years (people losing interest…). And because of our “accept many, re-nominate when <10” rule, even (1) might not lead to predictable schedule. So when I have to choose between a (maybe) predictable schedule, and treating all new members equally, I strongly go for the latter :)

goldfirere:
I've updated the proposal to remove the reference to October and to specify that all new members get three-year terms.

nomeata:
Did you push?

goldfirere:
> Did you push?

Yes, but to the wrong place. Should be fixed how.

goldfirere:
I'd like to submit this for a vote, @nomeata.

README.rst
nomeata: This is indeed just the steps to take to merge a PR

committee.rst
```diff
@@ -0,0 +1,164 @@
...
+new members and of violations of the guidelines for respectful communication
+will not appear on the mailing list.
+
+Term Limits and Committee Selection
+-----------------------------------
+
+Any community thrives best by continuing to refresh itself with new members.
+Accordingly, membership on the committee comes with a three-year term.
+Toward the end of an individual's term, the secretary of
+the committee will put out a public call for nominations for people to join
```
nomeata, line 104: Sorry for re-reading this carefully only now, but I am a bit confused: Do we call for nominations when

 * any time a member's term ends
 * when the number of members drop below 9 (as written further down)
 * in both events?

If only when the number drops, that does mean that re-nomination is only possible when enough members have dropped out?

goldfirere: Good point. I've amended the text to clarify that the answer is "both". It seems annoying for an existing member not to be able to stay on the committee if they and the rest of the committee wishes for them to, just because of these rules.

nomeata: If this was an IR, we’d say it has too many ad-hoc typing rules ;-)

But we can refine as we go.

goldfirere: Yes, it's a bit fiddly. But I don't see a better option here, other than perhaps fixing the size of the committee, which would make this point disappear. Yet I really like the idea of giving ourselves the freedom to accept more people than we lose, when we have a pool of good willing volunteers.

nomeata: Another option might be to extend terms until enough members have their 3 year term lapsed to put the total below 9.

Or put differently: 
 * Calls for nominations happen when there are less than 9 active members who have served less than 3 years (or maybe a bit before that happens)
 * Terms end with the first call for nominations after 3 years.
 
Also, should expiring members stop being members not at 3 years exactly, but at the end of the nomination and voting period, so that the committee isn’t suddenly empty, and that rejoining members are effectively serving without interruption?

goldfirere: I somehow have a hard time wrapping my head around the rules you propose. I'm inclined to keep the rules as I've written them in this regard. Do others have opinions here?

I'll clarify that departing members should stick around until their replacements arrive.

nomeata: Yeah, I have a hard time to write them down in a simple way, although I think they are, in fact, simple.

The current state of the PR has some odd behavor. For example, if we have a 11 member committee, and Joe’s term expires. It now depends soley on Joe’s whim whether there is a public nomination or not. Moreover, assume that Joe wants to stay on, then we have a round of nomination with a large committee, maybe taking on additional members. This can happen over and over again, and the committee grows out of bounds.

With my rules, the committee is bounded by “9 + number of people joining after a _single_ round of nominations”.

Let me try to rephrase the proposed rules, by distinguishing “active” members (<3 since joining or re-joining), “lame duck” members (≥ 3 years, but still active; don’t use this term in the bylaws) and “retired”. Then the rule are stated:

 * After 3 years, active members become lame ducks (but nothing happens pe se)
 * A public nomination happens when there are less than 9 active members.
 * Lame duck members retire (or rejoin) at the end of the next nomination process after they turned lamed ducks.

Is that clearer?

gridaphobe: Another way to think about @nomeata's rules is that we'd have fewer, larger nomination cycles as we batch up members whose terms have formally expired. I think that's probably a good thing, but haven't thought about it too carefully.

aspiwack: My inclination is to favour @nomeata 's proposal.

goldfirere: OK. I have updated the proposal accordingly. Please double-check to make sure it means what we described here. This is somehow trickier than I would have thought.

committee.rst
```diff
@@ -117,6 +109,19 @@ These criteria are not exhaustive -- nominations and self-nominations are free
...
+member who has served for over three years an "expiring" member.
+A nomination process is triggered when the number of unexpiring members is
+about to drop below 9.
+
+When a nomination process is triggered, the secretary of
+the committee will put out a public call for nominations for people to join
+the committee. Self-nominations are welcome and expected. Expiring members are
+free to re-nominate themselves. Nominations include a brief
+bio of the nominee and reasons why they would be appropriate for the
+committee.
```
nomeata, line 123: Just a reodering (else a reader not interested in self-nominations might stop reading )
```suggestion
When a nomination process is triggered, the secretary of
the committee will put out a public call for nominations for people to join
the committee. Nominations include a brief
bio of the nominee and reasons why they would be appropriate for the
committee. Self-nominations are welcome and expected. Expiring members are
free to re-nominate themselves. 
```

PR #361, opened 2020-09-04
Variadic Array Creation Primops

buggymcbugfix:
Proposal rendered [here](https://github.com/buggymcbugfix/ghc-proposals/blob/arrayOf/proposals/0000-variadic-array-creation-primop.md).

At the risk of being a bit gaudy, I allow myself to include some benchmark results here:

### Ubuntu 16.04 16-core i9
arrayOf-bench

The primop discussed here gives rise to the red and blue lines.
Coarse granularity (intervals of 50), with static allocation optimization (blue).

### MacBook Pro 16-core i9
arrayOf-benchmark

The primop discussed here gives rise to the blue line.
Fine granularity (intervals of 1).

### DIFF
--- a/proposals/0000-variadic-array-creation-primop.md
+++ b/proposals/0000-variadic-array-creation-primop.md
@@ -1,0 +1,174 @@
+---
+author: Vilem-Benjamin Liepelt (@buggymcbugfix)
+date-accepted: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/361).
+
+# Variadic Array Creation Primops
+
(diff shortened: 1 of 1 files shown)


Boarders:
I think this would be a very good change and in general I would love Haskell to move towards having more `constexpr` type behaviors (without having to jump through template Haskell hoops).

Ericson2314:
> I would love Haskell to move towards having more `constexpr` type behaviors.

Yes we absolutely do need that. I think this, as with just about anything that uses runtime reps, could be a bit better if we had a proper version of some sort of `constexpr` system (maybe http://www.jonmsterling.com/pdfs/modal-mltt.pdf would help??) but I won't fault this on that not being ready yet, and hopefully this will provide further motivation for that.

treeowl:
I believe you can improve the types *slightly*:

```haskell
arrayOf# :: forall rs (a :: TYPE ('TupleRep rs)) b. a -> Array# b
```

This is still pretty untyped. We'd really like the type checker to verify that the types all have the right `RuntimeRep`.

PR #362, opened 2020-09-08
Add incrementally cleared arrays

treeowl:
Enable data structures with worst-case time bounds to use arrays
by adding support for incremental array initialization.

[Rendered proposal](https://github.com/treeowl/ghc-proposals/blob/incrementally-cleared/proposals/0000-incrementally-cleared.md)


treeowl:
Ed Kmett has suggested a substantial improvement (at least for many/most situations); I'll edit the proposal to incorporate that idea.

PR #363, opened 2020-09-15
Start new proposal

Icelandjack:
Allow kind signatures without an associate definition

### DIFF
--- a/proposals/standalonestandalone.md
+++ b/proposals/standalonestandalone.md
@@ -1,0 +1,60 @@
+---
+author: Iceland_jack
+date-accepted: ""
+proposal-number: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/0>).
+**After creating the pull request, edit this file again, update the number in
+the link, and delete this bold sentence.**
+
+# Standalone StandaloneKindSignatures
+
+Many papers like [A Monadic Framework for Delimited Continuations(https://legacy.cs.indiana.edu/~dyb/pubs/monadicDC.pdf) introduce a library like this
+
+```haskell
+data CC a -- Abstract
+instance Monad CC
+
+data Prompt a -- Abstract
+data SubCont a b -- Abstract
+
+type Obs = ... -- Arbitrary but fixed
+```
+
+Given that we have `StandaloneKindSignatures` I think we should be able to write (standalone) standalone kind signatures. 
+Currently it gives an error "The standalone kind signature for ‘..’ lacks an accompanying binding":
+
+```haskell
+type CC      :: Type -> Type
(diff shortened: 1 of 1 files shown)


ndmitchell:
Can you rename the PR title?

Icelandjack:
Yup done

Ericson2314:
I'm a bit on the fence.

This is nice during development, but one inevitably wants it at the term level too, where we can't use `Void` tricks. Backpack signatures also overlap, and have nicer semantics since the types are actually abstract. Lastly, my https://github.com/ghc-proposals/ghc-proposals/pull/351 also would recommend making keeping illegal in "real code", but allowed with an `-fdefer-*` (e.g. `-fdefer-missing-definition-errors) mechanism.

I think I would feel better if someone could put on the dependent haskell goggles and figure out in general when we can use skolemnization vs when we need partialiity, without falling back on a types vs terms distinction. (e.g. at first I though it had to do with matchability, but now I think it has more to with relevance. I'm not sure). If we can something based on that, with the `-fdefer-*` presentation, without assuming `Void` and empty classes, I would feel a lot better about this.

Icelandjack:
> This is nice during development, but one inevitably wants it at the term level too

I would want that as well! https://gitlab.haskell.org/ghc/ghc/-/issues/393

Icelandjack:
> This is nice during development, but one inevitably wants it at the term level too

I would want that as well! https://gitlab.haskell.org/ghc/ghc/-/issues/393

Icelandjack:
But it was just an idea that came to my mind, it reminds me of the proposal to allow a trailing "bind" in do-notation

```haskell
do as
     bs
     c <- cs
```

to mean the same as `do as; bs; cs`. Just a neat trick to require fewer `undefined`s during development

int-index:
When you declare a type constructor this way, is it generative? In other words, would this be valid?

```
type F :: Type -> Type
type family Map (f :: Type -> Type) :: [Type] -> Type where ...
type X = Map F [Int, Bool, Char]
```

If you say yes, then what happens if I then define `F` like this?

```
type family F a where ...
```

And if you say no, then it seems you're about to disallow a lot of valid programs.

So, I would think, this proposal depends on #242. Then the arity of a type constructor will be reflected in its kind signature, and only then we'll be able to talk about making them truly standalone.

Icelandjack:
Good call with #242

goldfirere:
Yes, I would say this depends on #242. And I'd want to tie this proposal with the idea of doing this in terms.

And we would want to be careful. Under this proposal, the following would be different:

```
type Void1 :: Type

type Void2 :: Type
data Void2
```

The difference is that

```hs
f :: Void2 -> Unicorn
f = \case {}
```

is valid and warning-free. The same would not be true of `Void1`, which should really be an abstract data type, where we have no knowledge of its constructors (or lack thereof).

PR #364, opened 2020-09-15
Unify `Nat` and `Natural` (under review)

goldfirere:
This proposal is an attempt to unblock https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3583, which is user-facing and requires some broader consultation.

[Rendered](https://github.com/goldfirere/ghc-proposals/blob/natural/proposals/0000-unify-natural.rst)

### DIFF
--- a/proposals/0000-unify-natural.rst
+++ b/proposals/0000-unify-natural.rst
@@ -15,15 +15,22 @@ the two by making ``GHC.TypeLits.Nat`` a synonym for ``Numeric.Natural.Natural``
 
 Motivation
 ----------
+
 GHC should not have two different types to represent natural numbers. One is
-enough.
+enough. The existence of the separate kind ``Nat`` and the impossibility of
+using ``Natural`` at compile time (and data constructors with ``Natural`` fields) is
+an unnecessary complication. It forces us to create
+redundant types with ``Nat`` fields in their constructors (making these types
+uninhabited at runtime) only for use at compile time. This is more cumbersome
+than the promotion of ordinary data types.
 
 Proposed Change Specification
 -----------------------------
 * Add ``type Nat = Natural`` in ``GHC.TypeNats``.
 
-* Re-export ``Natural`` from ``GHC.TypeLits`` (along with the current re-export
-  of ``Nat``)
+* Re-export ``Natural`` from ``GHC.TypeNats`` and
+  ``GHC.TypeLits`` (along with the current re-export
+  of ``Nat``).
 
 * Assign numbers in types to have kind ``Natural`` instead of kind ``Nat``.
 
@@ -33,7 +40,15 @@ Effect and Interactions
   is implied by ``FlexibleInstances``. Or, they could be written to use ``Natural``
   instead of ``Nat``.
 
-Otherwise, this change should be backward compatible.
+* It is possible for someone to have separate instances today for ``Nat`` and
+  ``Natural``, though it is unclear how an instance on ``Nat`` would be useful.
+  Having two separate instances for these types will not be possible after this
+  proposal is implemented.
+
+Otherwise, this change should be backward compatible. In particular, this proposal
+does *not* change parsing or other aspects of numbers written in types. For example,
+even though ``(-5 :: Natural)`` parses today (and throws a runtime error), that
+expression will *not* parse in types.
 
 
 Costs and Drawbacks
--- a/proposals/0000-unify-natural.rst
+++ b/proposals/0000-unify-natural.rst
@@ -1,0 +1,61 @@
+Unify ``Nat`` and ``Natural``
+=============================
+
+.. author:: Richard Eisenberg, inspired by Rinat Striungis
+.. date-accepted:: 
+.. ticket-url:: 
+.. implemented:: 
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. contents::
+
+GHC currently uses ``GHC.TypeLits.Nat`` to describe compile-time natural numbers
+but ``Numeric.Natural.Natural`` to describe runtime ones. This proposal unifies
+the two by making ``GHC.TypeLits.Nat`` a synonym for ``Numeric.Natural.Natural``.
+
+Motivation
+----------
+GHC should not have two different types to represent natural numbers. One is
+enough.
+
+Proposed Change Specification
+-----------------------------
+* Add ``type Nat = Natural`` in ``GHC.TypeNats``.
+
+* Re-export ``Natural`` from ``GHC.TypeLits`` (along with the current re-export
+  of ``Nat``)
+
+* Assign numbers in types to have kind ``Natural`` instead of kind ``Nat``.
+
+Effect and Interactions
+-----------------------
+* Instances written about ``Nat`` may now need ``TypeSynonymInstances``, which
+  is implied by ``FlexibleInstances``. Or, they could be written to use ``Natural``
+  instead of ``Nat``.
+
+Otherwise, this change should be backward compatible.
+
+
+Costs and Drawbacks
+-------------------
+* We might be painting ourselves into a corner, having not yet worked out the
+  way we will support other types of compile-time literals. But I (Richard)
+  think this will be OK.
+
+
+Alternatives
+------------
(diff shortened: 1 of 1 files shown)


proposals/0000-unify-natural.rst
```diff
@@ -0,0 +1,59 @@
...
+Motivation
+----------
+GHC should not have two different types to represent natural numbers. One is
+enough.
+
+Proposed Change Specification
+-----------------------------
+* Add ``type Nat = Natural`` in ``GHC.TypeNats``.
+
+* Re-export ``Natural`` from ``GHC.TypeLits`` (along with the current re-export
```
RyanGlScott, line 25: Mention `GHC.TypeNats` as well, since that's where the definition is `Nat` is. (`GHC.TypeLits` just re-exports it.)

proposals/0000-unify-natural.rst
```diff
@@ -0,0 +1,59 @@
...
+  of ``Nat``)
+
+* Assign numbers in types to have kind ``Natural`` instead of kind ``Nat``.
+
+Effect and Interactions
```
RyanGlScott, line 30: One thing that may be worth emphasizing is that making `Nat = Natural` will not affect the existing behavior of type-level numeric literals. You might expect this to be the case, since `Natural`s can be negative but `Nat`s cannot, but if I understand this proposal correctly, then the operations that GHC exposes over type-level numeric literals won't permit them to be negative anyway.

RyanGlScott (commented):
I think the cut of this proposal's jib. I've left two minor suggestions inline.

proposals/0000-unify-natural.rst
phadej: Since when `Natural` can be negative?

RyanGlScott: Sorry, I should have been more precise when I wrote that. What I mean is that GHC will happily parse this:

```hs
f = (-1 :: Natural)
```

On the other hand, this will not parse:

```hs
type F = -1
```
```
error:
    Illegal literal in type (type literals must not be negative): -1
```

Similarly while running the following term-level program will throw an error:

```hs
1 - 2 :: Natural
*** Exception: arithmetic underflow
```

The corresponding type-level program will be stuck instead:

```
λ> :kind! 1 - 2
1 - 2 :: Nat
= 1 - 2
```

I don't believe this proposal should affect these differences between terms and types, so I wanted that point to be mentioned explicitly.

int-index: `Natural` can't be negative, unlike `Integer`, which can.

proposals/0000-unify-natural.rst
```diff
@@ -0,0 +1,59 @@
...
+-----------------------
+* Instances written about ``Nat`` may now need ``TypeSynonymInstances``, which
+  is implied by ``FlexibleInstances``. Or, they could be written to use ``Natural``
+  instead of ``Nat``.
+
+Otherwise, this change should be backward compatible.
```
int-index, line 36: That's not the full story, the following code will also break:

```
class C a
instance C Nat
instance C Natural
```

Now these instances conflict with one another.

proposals/0000-unify-natural.rst
Ericson2314: Off topic, but if `1 - 2` could be a type error, at least until we get constrained type familiies (in which case we can do proper partial functions!), that would be nice.

phadej: @RyanGlScott I can only comment that `Num Natural` is utterly broken, `negate` and `fromInteger` are just wrong. Also the fact we have `-` on type level, and not [`Monus`](https://en.wikipedia.org/wiki/Monus) is IMHO also a wart, as if we were writing Agda or Coq we wouldn't (be able to) define such partial function.

proposals/0000-unify-natural.rst
```diff
@@ -0,0 +1,59 @@
...
+
+GHC currently uses ``GHC.TypeLits.Nat`` to describe compile-time natural numbers
+but ``Numeric.Natural.Natural`` to describe runtime ones. This proposal unifies
+the two by making ``GHC.TypeLits.Nat`` a synonym for ``Numeric.Natural.Natural``.
+
+Motivation
+----------
+GHC should not have two different types to represent natural numbers. One is
+enough.
+
```
Haskell-mouse, line 20: I would add something like: 

Existence of separated kind Nat and the impossibility of the promotion of usual Natural data type (and data constructors with Natural fields) looks like an unnecessary and inconvenient complication. It forces us to create redundant types with Nat fields in their constructors (it makes these types uninhabited in terms)only for promotional purposes. This is more cumbersome than the promotion of ordinary data types.

int-index: Indeed, currently one must go through some hoops to get data types with numeric fields promoted. Consider this data type:

```
data T = MkT [Bool] Natural
```

The `[Bool]` field can be promoted just fine, but the `Natural` field can't be:

```
x = MkT [True, False, True] 42  -- works
type X = MkT [True, False, True] 42   -- doesn't work because of 42
```

So one must introduce extraneous type variables as a workaround:

```
data T n = MkT [Bool] n
x = MkT [True, False, True] (42 :: Natural)
type X = MkT [True, False, True] (42 :: Nat)
```

Here, `n` is instantiated to either `Natural` or `Nat`, depending on whether it's used as a type or as a kind. This issue basically affects any user-defined data type that has a numeric field, and this proposal solves it.

int-index: Of course, there's also a similar issue with `Symbol` / `Text`, but I'd claim fixing one annoyance at a time is not unreasonable.

int-index:
@ekmett appears to be asking for this exact change in https://twitter.com/kmett/status/1306872837588754433

goldfirere:
Any more commentary before submitting this to committee? Seems pretty straightforward.

int-index:
> Any more commentary before submitting this to committee?

None from me or @Haskell-mouse. Looks good, thank you!

goldfirere:
Let's submit then, @nomeata .

nomeata:
/remind @serras to ensure progress in two weeks

aspiwack:
The cost of this proposal is mightily low (there is a small, already ready MR, which has essentially 0 net lines of code).

So I'd see the bar to pass to be just as low.

However, I think that the motivation would be quite a bit more readable, and will stand the test of time better, with a tiny bit of code illustrating the difficulty.

jvanbruegge:
Are there any plans concerning Symbol/String (or maybe Text)? They have basically the same issue. Or would we want to wait for something that allows overloading type level strings? I remember that there was a proposal for this, but no idea what the status is

int-index:
> Are there any plans concerning Symbol/String (or maybe Text)?

There's ongoing work to make the promotion of `Char` possible, this way you also get `String`.

However, `Symbol` is more efficient than `String`, so we can't just unify them as we unify `Nat` and `Natural` here. This will require more thought, and a separate proposal. Personally, I like the idea of making `Symbol` inhabited at the term level by UTF-8 encoded strings.

goldfirere:
@int-index You and colleagues (I don't have their GitHub usernames) are the instigators here. Would you care to write a better motivation section? Feel free just to push to my branch -- it seems you already have access.

serras:
My recommendation is to accept this proposal.

Apart from the small comment by @goldfirere, I think it would be nice for the proposal to point our that neither `Natural` nor the old `Nat` kind implement Peano naturals, so people should not expect to be able to decompose them.

int-index:
@aspiwack I've added an example to the Motivation section. How is it looking now?

Ericson2314:
> > Are there any plans concerning Symbol/String (or maybe Text)?
> 
> There's ongoing work to make the promotion of `Char` possible, this way you also get `String`.
> 
> However, `Symbol` is more efficient than `String`, so we can't just unify them as we unify `Nat` and `Natural` here. This will require more thought, and a separate proposal. Personally, I like the idea of making `Symbol` inhabited at the term level by UTF-8 encoded strings.

Yeah to me the "right" way to do this is dramatically more work than this proposal and should be discussed separately. (I would want promoted `IsString` so we don't have to bloat base further.)

aspiwack:
> @aspiwack I've added an example to the Motivation section. How is it looking now?

It's the sort of things I had in mind. Thanks.

int-index:
@nomeata @serras What's the status here?

nomeata:
There seems to be pretty clear consensus on the commitee, I’ll mark this as accepted.

PR #84, opened 2017-10-18
DuplicateRecordFields simplification

adamgundry:
[Rendered](https://github.com/adamgundry/ghc-proposals/blob/duplicaterecordfields-simplification/proposals/0000-duplicaterecordfields-simplification.rst).

PR #366, opened 2020-09-25
DuplicateRecordFields without ambiguous field access (under review)

adamgundry:
This proposal addresses an unsatisfactory aspect of `DuplicateRecordFields`, namely the unclear rules around when a field selector or update will be accepted, by entirely removing the type-directed name resolution aspect. This proposal is more restrictive than the previous (dormant) simplification proposal #84, allowing fewer programs, but correspondingly simpler. The proposal is similar to and reuses parts of #84, but I decided to open a new proposal as the suggested simplification here is more radical.

[Rendered](https://github.com/adamgundry/ghc-proposals/blob/no-ambiguous-selectors/proposals/0000-no-ambiguous-field-access.rst).

### DIFF
--- a/proposals/0000-no-ambiguous-field-access.rst
+++ b/proposals/0000-no-ambiguous-field-access.rst
@@ -23,7 +23,7 @@ The ``DuplicateRecordFields`` extension is useful to allow multiple datatypes to
   module M where
 
   data Person = MkPerson { name :: String, pets :: [Pet] }
-  data Pet    = MkPet    { name :: Text   }
+  data Pet    = MkPet    { name :: Int }
 
 It is now entirely unproblematic to use ``name`` in a record construction or pattern match, because the constructor name disambiguates which field is meant::
 
@@ -145,6 +145,61 @@ Since this proposal will break existing code using ``DuplicateRecordFields``, we
 This transition period will give time for users of ``DuplicateRecordFields`` to adapt their code (using ``RecordDotSyntax`` or otherwise), or raise concerns about the proposed changes and request a stay of execution.  Our expectation is that step 2 will be taken in the GHC release immediately following step 1, but this can be changed if feedback from users indicates that the removal of the feature is causing substantial pain.
 
 The warning produced by ``-Wambiguous-fields`` should mention the specific selector and type that were determined by the disambiguation rules, rather than just complaining about the ambiguity.  This should make it easier for affected users to adapt their code.
+
+
+Migration strategy
+^^^^^^^^^^^^^^^^^^
+Code that is broken by this proposal because it relies on ambiguous field occurrences can be fixed in one of the following ways:
+
+1. Where the field is defined in a different module, use qualified imports, import hiding and/or aliases to remove the ambiguity.  For example, here is a `technique using import aliases <https://gist.github.com/chrisdone/d7a8f9e91e2ac111fac6ab72cc480f78>`_::
+
+    {-# LANGUAGE DuplicateRecordFields #-}
+    module M1 where
+      data Person = MkPerson { name :: String, pets :: [Pet] }
+      data Pet    = MkPet    { name :: Int }
+
+    module N where
+      import M1 as Person (Person(..))
+      import M1 as Pet (Pet(..))
+
+      getPersonName :: Person -> String
+      getPersonName = Person.name
+
+      setPersonName :: String -> Person -> Person
+      setPersonName n p = p { Person.name = n }
+
+   The new version of the code is completely backwards-compatible, and its meaning is clear. The downsides are that this approach cannot be used within a single module, barring enhancements to the module system, and it requires boilerplate imports.
+
+2. Use ``RecordDotSyntax`` when it is available::
+
+    {-# LANGUAGE DuplicateRecordFields, RecordDotSyntax #-}
+    module M2 where
+      data Person = MkPerson { name :: String, pets :: [Pet] }
+      data Pet    = MkPet    { name :: Int }
+
+      getPersonName :: Person -> String
+      getPersonName p = p.name
+
+      setPersonName :: String -> Person -> Person
+      setPersonName n p = p { name = n }
+
+   This works within a single module. However it requires a new as-yet-unreleased extension, and will not work for fields with higher-rank or unboxed types.
+
+3. Use explicit pattern-matching and record construction, possibly in combination with ``NamedFieldPuns`` or ``RecordWildCards``::
+
+    {-# LANGUAGE DuplicateRecordFields, NamedFieldPuns #-}
+    module M2 where
+      data Person = MkPerson { name :: String, pets :: [Pet] }
+      data Pet    = MkPet    { name :: Int }
+
+      getPersonName :: Person -> String
+      getPersonName MkPerson{name} = name
+
+      setPersonName :: String -> Person -> Person
+      setPersonName n MkPerson{pets} = MkPerson {name = n, pets }
+
+   This works in a single module and does not require any new extensions, but it may require additional boilerplate, especially if a type has many constructors and/or fields.
+
 
 
 Effect and Interactions
--- a/proposals/0000-no-ambiguous-field-access.rst
+++ b/proposals/0000-no-ambiguous-field-access.rst
@@ -44,6 +44,8 @@ When ``DuplicateRecordFields`` is disabled, ambiguous field selector/update occu
 
 Note that if an identifier refers to both non-fields and fields, it is always rejected by the renamer as excessively ambiguous, regardless of ``DuplicateRecordFields``.
 
+This proposal does not affect field selector/update occurrences that are not ambiguous.
+
 
 Existing rules for selectors
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


int-index:
I agree with the proposed direction of travel. The way I see it, there are two approaches to name overloading:

1. Type-driven name resolution
2. Class-based polymorphism

Haskell normally uses (2), but in rare cases ends up using (1), one such case being `DuplicateRecordFields`. We should aspire to use (2) always, and the `HasField` machinery is exactly that.

Ericson2314:
I feel a bit silly bringing this up, but might you rename the proposal something like 
> DuplicateRecordFields without ~~ambiguous selectors~~ *field access*"

Since you define "field selector occurrence" and "field update occurrence", and the proposal applies to both, I think it's confusing to title it with the word "selector".

--------

I agree the "Type-driven name resolution" that's done is bad. I'm not sure "Class-based polymorphism" is good, but I remain a little hopeful the "Name-driven" resolution I went for in #310 becomes viable after all. I like this proposal because it just pushes us away from type-driven resolution, and doesn't bias one alternative over another.

adamgundry:
Regarding type-directed name resolution, my feeling is that it is in principle a useful approach, particularly in languages that are heavily slanted towards bidirectional type-checking (e.g. Idris, Agda). That is, I think TDNR done right can work well. But the use in `DuplicateRecordFields` is clearly not "TDNR done right", and indeed the interactions with a wider context that prioritises class-based resolution mean that it is better to just drop the TDNR aspect from `DuplicateRecordFields`.

(A good bidirectional/TDNR story for record selectors really needs dot-notation, because then you can infer the type of the record being accessed and use that to disambiguate the selector. But with `RecordDotSyntax` taking the class-based approach I think that ship has sailed.)

> I feel a bit silly bringing this up, but might you rename the proposal something like
> 
> > DuplicateRecordFields without ~ambiguous selectors~ _field access_"
> 
> Since you define "field selector occurrence" and "field update occurrence", and the proposal applies to both, I think it's confusing to title it with the word "selector".

@Ericson2314 thanks, that's a good point. (I originally wrote the proposal for selectors, then decided to add updates, and of course I forgot to change the title. :facepalm:) I've updated the PR.

simonpj:
I'm strongly in favour of this proposal.   The current design is an egregious hack; I don't think it would every have got past this committee!   Moreover, it has a pretty significant implementation cost; that would be OK if it were in service of a principled goal, but it isn't.

Let's tidy this up.

goldfirere:
I'm also in strong support of this proposal. Indeed, I would advocate for a shorter, more abrupt transition period. Maybe we warn for a release, but then just rip out the existing machinery. Any change to code is fully backward compatible.

Users could easily convince me that a slower transition is better, but I don't think we designers should assume that users need the slower transition.

ekmett:
I'm not formally a part of the approval process, but I did want to chime in with a strong +1 in favor of this proposal. `DuplicateRecordFields` is a horrible wart, and this seems to shave it down to a reasonable nub.

adamgundry:
> I would advocate for a shorter, more abrupt transition period. Maybe we warn for a release, but then just rip out the existing machinery.

@goldfirere the reaction to this proposal is making me agree we should rip it out ASAP. So default to `-Wwarn=ambiguous-fields` for a release, then gone in the next release? If everything goes perfectly we might aim to get `RecordDotSyntax`, `NoFieldSelectors` and `-Wwarn=ambiguous-fields` into 9.2, then drop the hack completely in 9.4?

(Incidentally, [this gist](https://gist.github.com/chrisdone/d7a8f9e91e2ac111fac6ab72cc480f78) from @chrisdone shows how the module system can be used nicely to avoid field occurrence ambiguity in modules other than the defining module. I half wonder if we should allow modules to import themselves qualified and with an explicit import list, so this trick would work in the defining module as well...)

goldfirere:
Yes, I support the more rapid removal @adamgundry proposed.

AntC2:
>  this gist from @chrisdone shows how the module system can be used nicely to avoid field occurrence ambiguity in modules other than the defining module.

cute! Isn't there a more general feature lurking here, not specific to ambiguous record field names? (They merely bite the worst.) For example I've seen loose talk about multi-module modules: that is, one source file with multiple `module` decls, mutually importing each other.

There are for example StackOverflow questions asking why you can't use duplicate names for data constructors, especially for enumerated types: `Orange` as a colour vs as a fruit; `Dec, Oct` as a month vs as a number base. And I guess there's some generic method names (`make, build, empty, singleton, size, length`) that are likely to be ambiguous amongst classes.

> I half wonder if we should allow modules to import themselves qualified and with an explicit import list, so this trick would work in the defining module as well

mmm that seems heavy machinery. Would something like this be more succinct (but using the same name prefixing under the covers)

    data Person qualified = Person { name :: String, age :: Int } deriving Show
    data Company qualified as Co = Company { name :: String, age :: Int } deriving Show

    data Colour qualified = Red | Orange | Yellow | Green | Blue | Indigo | Violet

    class Shape a qualified  where { area :: ...; perimeter :: ...; centre :: ...    }

in which if the `as`-part is omitted `qualified as <name to my left>` is assumed.

(hmm hmm I used `qualified` because it's already a reserved word, not likely to be used in existing code; whereas bare `as` is liable to clash with tyvar names. OTOH `import qualified` means the name must always be used qualified -- which is probably onerous.)

klapaucius:
@AntC2 https://github.com/ghc-proposals/ghc-proposals/pull/283

adamgundry:
@AntC2 yes, local modules would be another way to be able to disambiguate field occurrences, and there might indeed be some merit in having datatype declarations imply local modules. FWIW, being able to refer to the enclosing module with a different prefix and an explicit import list has come up as desirable before (https://gitlab.haskell.org/ghc/ghc/-/issues/15432). But in any case this is out of scope of the current proposal.

---

So far the only dissatisfaction I've seen with this proposal is a single :-1: from @klapaucius. Would you be willing to elaborate?

I've thrown together a rough implementation of the new warning at https://gitlab.haskell.org/ghc/ghc/-/commit/abf0fb3138bd05a94fe69fb887cf308e51d3a7d4 ... anyone fancy bikeshedding the warning text?

```
overloadedrecfldsfail06.hs:15:22: error: [-Wambiguous-fields (in -Wdefault), -Werror=ambiguous-fields]
    The record update u {x = True} with type U is ambiguous.
    This will not be supported by -XDuplicateRecordFields in future releases of GHC.

overloadedrecfldsfail12.hs:13:5: error: [-Wambiguous-fields (in -Wdefault), -Werror=ambiguous-fields]
    The field ‘foo’ belonging to type T is ambiguous.
    This will not be supported by -XDuplicateRecordFields in future releases of GHC.
    You can use a qualified import or explicit case analysis to resolve the ambiguity.

overloadedrecfldsfail12.hs:16:5: error: [-Wambiguous-fields (in -Wdefault), -Werror=ambiguous-fields]
    The field ‘foo’ belonging to type S is ambiguous.
    This will not be supported by -XDuplicateRecordFields in future releases of GHC.
    You can use explicit case analysis to resolve the ambiguity.
```

klapaucius:
@adamgundry 

I prefer that existing features aren't removed unless they are really problematic. Such removals induce unnecessary anxiety about extensions and I'm not looking forward to discussions with my colleagues what we shouldn't use some extensions because "they can be removed at any moment just like TDNR in DuplicateRecordFields".
I agree that TDNR is somewhat problematic, but I believe that it shouldn't be fixed with [an even more problematic extension](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0282-record-dot-syntax.rst).
If the breakage is necessary I would prefer to disambiguate selectors with a type or constructor name.
I believe that it would be better to wait until something like [proposal 283](https://github.com/ghc-proposals/ghc-proposals/pull/283) is accepted and implemented.

adamgundry:
@nomeata I believe this proposal is ready for review. From the discussion:
 * Several people expressed support for this change. The proposal originally suggested a possibly longer transition period, but this was amended in the light of feedback.
 * @klapaucius raised concerns about the impact of breaking existing working features (see below).
 * There was some discussion of possible extensions to the module system that might help with name ambiguity problems, but I don't think these need hold up this proposal and they can be discussed elsewhere.

----

@klapaucius thanks for the feedback. I do acknowledge that breaking changes are always painful, but if we don't make them we are stuck with dubious design decisions forever, and here I think the consensus is that this feature is one we could do without. I've added a line to the "Costs and Drawbacks" section, and added a new section documenting migration strategies for fixing existing code.

adamgundry:
@nomeata ping; I would like to submit this proposal for committee review.

nomeata:
Thanks for the ping!

i-am-tom:
I'm in favour of this proposal for several reasons, all of which have already been given above. I think the current behaviour is unintuitive, and this proposal's explicit presentation of the rules (under which the current TDNR works) is the first time I've really thought about them beyond hand-waving or trial and error. I'm therefore going to recommend that we accept this, pending committee review.

i-am-tom:
Hey! We've discussed this as a committee, and we agree that it should go ahead. This proposal will now be marked as accepted, and @nomeata will merge it in for us. Thanks for submitting the proposal!

PR #367, opened 2020-09-26
Clarify primops using unboxed sums

treeowl:
* Instead of hackishly representing primop results using
  unboxed tuples, flags, and undefined values, use unboxed
  sums.

* Split `addCFinalizerToWeak#` into two primops.

[Rendered proposal](https://github.com/treeowl/ghc-proposals/blob/unboxed-sum-primops/proposals/0000-unboxed-sum-primops.md)

### DIFF
--- a/proposals/0000-unboxed-sum-primops.md
+++ b/proposals/0000-unboxed-sum-primops.md
@@ -98,21 +98,14 @@ naturally returns an unboxed sum. Unfortunately, it is not documented in
 
 ### Migration plan
 
-In version `v`, GHC will add the new primops using names suffixed with the
-letter "S" (except for the finalizer adding primops): `tryTakeMVarS#`,
-`tryReadMVarS#`, `deRefWeakS#`, `finalizeWeakS#`, and `getSparkS#`. Compatibility
-wrappers will be added to `GHC.Exts`.
+In version `v`, we will add a module `GHC.NewPrimops` with implementations
+of the future primops using the current ones. We will deprecate the current
+primops.
 
-In version `v+1`, the wrappers will be deprecated.
+In version `v+1`, we will make the new primops actual primitives, and deprecate
+the `GHC.NewPrimops` module.
 
-In version `v+2`, the wrappers will be removed.
-
-In version `v+3`, the new primops will be renamed to remove the final "S", with
-"S"-suffixed synonyms added to `GHC.Exts` for compatibility.
-
-In version `v+4` the "S"-suffixed synonyms will be deprecated.
-
-In version `v+5`, the "S"-suffixed synonyms will be removed.
+In version `v+2`, we will remove the `GHC.NewPrimops` module.
 
 ## Examples
 
--- a/proposals/0000-unboxed-sum-primops.md
+++ b/proposals/0000-unboxed-sum-primops.md
@@ -42,8 +42,7 @@ I think we should bring that clarity to the primops.
 ## Proposed Change Specification
 
 I propose to replace the primops that produce product-encoded sums
-with ones that produce proper unboxed sums, and to add compatibility
-wrappers in `GHC.Exts` with the old types.
+with ones that produce proper unboxed sums.
 
 The primop `addCFinalizerToWeak#` effectively takes an unboxed sum as
 an argument. For simplicity and clarity, I propose to split it into
--- a/proposals/0000-unboxed-sum-primops.md
+++ b/proposals/0000-unboxed-sum-primops.md
@@ -159,6 +159,8 @@ not many will be affected.
 
 ## Alternatives
 
+### Finalizer addition
+
 We could write a version of `addCFinalizerToWeak#` that takes an unboxed sum.
 This would match the current API a little more closely, but the ergonomics seem
 fairly lousy compared to the splitting option:
@@ -176,6 +178,19 @@ like the idea in
 [#361 (Variadic Array Creation Primops)](https://github.com/ghc-proposals/ghc-proposals/pull/361),
 but it's hard to justify the effort to do so for an operation that cannot be
 expected to be used very heavily.
+
+### Unit type
+
+I've used `(# #)` as the "unboxed unit type", which strikes me as the
+most logical. We *could* instead use `Void#`, which is somewhat more
+traditional for `GHC.Prim`. It has two downsides, however:
+
+1. It's impossible to match on a value of type `Void#`, so users are
+   forced to use a wildcard. `f (# _ :: Void# | #) = ...` doesn't seem
+   as nice as `f (# (# #) | #)` for making the type apparent.
+
+2. The name `Void#` seems to suggest an unboxed version of `Data.Void.Void`,
+   but it is not one.
 
 ## Unresolved Questions
 
--- a/proposals/0000-unboxed-sum-primops.md
+++ b/proposals/0000-unboxed-sum-primops.md
@@ -1,0 +1,170 @@
+---
+author: David Feuer
+date-accepted: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/367).
+
+# Use unboxed sums for primops
+
+Several primops are expressed rather awkwardly, because they naturally
+produce sum types, and historically those have not been available to
+primops. Now that we have unboxed sums, we can use those.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


simonpj:
Looks plausible to me, because it would make the type of the primop better reflect its semantics.

Some work needed on transition plan. 
* Are you suggesting a permanent change of name, from `tryTakeMVar#` to `tryTakeMVarS#`?  I can see transitional arguments for that, but it means that the final status quo will have a strangely-named primop.
* Presumably the old primop will be deprecated, and eventually removed?

treeowl:
@simonpj, I'm not fussy about it. If the old onea are eventually removed, then the new ones can eventually be renamed. That'll probably only take what, ten years? I'm happy to let others make that decision.

simonpj:
I'm not fussy either. But (in due course at least) the proposal should specify what the plan is.

treeowl:
@simonpj, I've updated the proposal with a migration plan.

treeowl:
@nomeata, while there are still unresolved questions (mostly minor), no one has commented recently to help resolve them. I therefore submit this proposal to the committee.

nomeata:
/remind @simonmar to check progress in two weeks

simonpj:
I'm generally in support of this, although the syntax for unboxed sums is terrible.  (My fault I know.)

I'm not sure why the deprecation plan has to be so long.  e.g. Why not deprecate the old things immediately (in version 'v')?

Also I'm not sure about these S-suffix things.  Why not export `takeMVar#` with the new type from `GHC.NewPrimops` (or some suitable module), but with no S suffix. So the changes required are:
* Change `import GHC.Primops`  to `import GHC.NewPrimops` and change calls to `takeMVar#` to account for its new type
* After a while... change back to `import GHC.Primops`

That seems less fiddly than adjusting every call site a second time.

treeowl:
@simonpj, I'm okay with pretty much any migration plan the committee will accept. Should I modify the proposal to match your suggestion above, or wait for the committee to complete its review?

treeowl:
@simonpj, I'm also not so clear about your suggestion. `GHC.Prim` is a fake module full of primitive stuff. I have no idea what would be involved in adding a `GHC.NewPrim` module like that.

simonpj:
Perhaps
```
module GHC.NewPrimops( module GHC.Prim, ... ) where
  import GHC.Primops
  ...extra definitions...
```

But I have not thought about it carefully. Migration planning is painstaking.  I was just asking whether it needs to take five releases of GHC.  Perhaps it does, but perhaps you can think of a more rapid path.

The principle is, I think, that wherever possible a new release should work with the old code, but emit deprecation warnings suggesting what to do.  Then the next release can implement those deprecations.

treeowl:
Oh, I see. So start by implementing the new versions as library code, then make the real switch. Yes, that should work.

treeowl:
@nomeata, @simonmar, do you think there will be a shepherd recommendation in the next week?

simonpj:
David, could you clarify the migration plan.

> In version v, we will add a module GHC.NewPrimops with implementations of the future primops using the current ones. We will deprecate the current primops.

What will be the *names* of the new primops?  Same name, different types, and different exporting module?  An example would be super helpful.

> In version v+1, we will make the new primops actual primitives, and deprecate the GHC.NewPrimops module.

What does that mean?   Do you mean that both `GHC.Primops` (is that the right module name) and `GHC.NewPrimOps` will export the new primop with the new type?

A rationale for the migration plan would help.  I think it is this:

* During version v, responding to deprecation warnings, users can 
  * Replace `import GHC.Primops` with `import GHC.NewPrimops`
  * Change uses of the affected primops to use the new primop types
  They do not need to change the *name* by which the primop is called.  Old programs continue to work, albeit with deprecation warnings.

* During version v+1, responding to deprecation warnings about `import GHC.NewPrimops`,users can change `import GHC.NewPrimops` back to `import GHC.Primops`.

Is that more or less it? Spelling it out would help.

treeowl:
@simonpj, yes, that's the idea. I'll make the changes today or tomorrow.

treeowl:
@simonpj, I'm sorry about the delay. I just updated the proposal in response to your suggestions.

simonmar:
In principle this is the right thing, I'm just not sure that it's worth it.

* Does our Cmm parser and compiler have all the bits it needs to support returning unboxed sums? I'm not all that familiar with the implementation of unboxed sums, but I wouldn't be surprised if there is real work to do here.
* It's not actually causing any problems.
* There has to be a migration, there will be legacy backwards compat stuff floating around for a while to cause confusion, and there's room for things to go wrong (performance regressions due to the backwards compat layer)
* Is someone volunteering to do all the work here?

treeowl:
> In principle this is the right thing, I'm just not sure that it's worth it.
> 
> * Does our Cmm parser and compiler have all the bits it needs to support returning unboxed sums? I'm not all that familiar with the implementation of unboxed sums, but I wouldn't be surprised if there is real work to do here.

I don't imagine we'd need to change Cmm. We represent unboxed sums in Cmm as unboxed products, a lot like the primops I'm proposing to change. There's no immediate need to change that. I expect the hardest part will be adding support for expressing unboxed sum types in `primops.txt.pp`. I don't know what might be involved there.

> * There has to be a migration, there will be legacy backwards compat stuff floating around for a while to cause confusion, and there's room for things to go wrong (performance regressions due to the backwards compat layer)

I don't believe performance regressions will be a big issue for this particular proposal. The primops involved are all out of line anyway, and mostly on the expensive side.

> * Is someone volunteering to do all the work here?

I'd be happy to give it a go, but I'd need some help understanding the `primops.txt.pp` parsing.

simonmar:
> > In principle this is the right thing, I'm just not sure that it's worth it.
> > 
> > * Does our Cmm parser and compiler have all the bits it needs to support returning unboxed sums? I'm not all that familiar with the implementation of unboxed sums, but I wouldn't be surprised if there is real work to do here.
> 
> I don't imagine we'd need to change Cmm. We represent unboxed sums in Cmm as unboxed products, a lot like the primops I'm proposing to change. There's no immediate need to change that. I expect the hardest part will be adding support for expressing unboxed sum types in `primops.txt.pp`. I don't know what might be involved there.

As I understand it, unboxed sums are translated away before code generation. So there are two alternatives for writing primops that use unboxed sums, either (a) have some syntax for returning an unboxed sum in Cmm and apply the compiler's translation to it, or (b) write the primops to return the translated types directly. The right way would be (a), but I think you're suggesting (b) - it would work, but the concern is that if we ever change the translation we'll have to also change the primop implementations, and it diffuses the knowledge about the translation into more parts of the compiler.

> > * There has to be a migration, there will be legacy backwards compat stuff floating around for a while to cause confusion, and there's room for things to go wrong (performance regressions due to the backwards compat layer)
> 
> I don't believe performance regressions will be a big issue for this particular proposal. The primops involved are all out of line anyway, and mostly on the expensive side.

`tryTakeMVar` is *very* performance-critical.

I'm not saying you can't avoid a performance penalty. But the moment you have a compatibility layer, the possibility for accidental regressions creeps in, and who is going to notice? Only users, when we ship the compiler with a regression. 

Question: would we actually need a compatibility layer? I don't think GHC's primops are covered by the 3-release policy are they? 

> > * Is someone volunteering to do all the work here?
> 
> I'd be happy to give it a go, but I'd need some help understanding the `primops.txt.pp` parsing.

Very kind of you to offer. 

At the moment I'm still not sure I'm convinced this is worth it, though.

simonmar:
Not to leave this discussion unfinished, I'll suggest that we reject this proposal, on the grounds that the gains are small and there are several potential problems: complexity, migration, and performance problems could arise as mentioned above.

nomeata:
> Not to leave this discussion unfinished, I'll suggest that we reject this proposal, on the grounds that the gains are small and there are several potential problems: complexity, migration, and performance problems could arise as mentioned above.

Is this your “official” recommendation, or just a heads-up to allow the authors to rebut? In the former case, please also send it to the mailing list.

nomeata:
The committee has followed Simon’s recommendation to reject this proposal, for the following reasons:
 * The implementations of the primops would bake in the translation of unboxed sums. Right now our unboxed sum translation is defined in one place - CorePrep - but this proposal would leak it into the implementations of a few primops.
 * Migration would entail adding a backwards-compatibility layer, with the associated complexity and risk of performance regressions.
 * While the current situation is perhaps inelegant, it's not broken and it's easy to understand.


Nevertheless, thanks for your work on this proposal, @treeowl!

PR #368, opened 2020-09-29
Warn on prefix/suffix operators (under review)

goldfirere:
This is an amendment to an existing proposal, saying that we should warn (in `-Wall` and `-Wcompat`) when an operator is used in a suffix or prefix occurrence.

Motivation: we might steal this syntax in the future, so it's good to warn against using it now.

Current, accepted proposal: https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0229-whitespace-bang-patterns.rst
Amended proposal: https://github.com/goldfirere/ghc-proposals/blob/operators/proposals/0229-whitespace-bang-patterns.rst (the only change is in the third-to-last specification bullet)

Below this line is the original text of this PR, retained only so that the comments below can be interpreted.

-----------------------------------------------
We sometimes like to steal syntax by treatment the whitespace next to operators as significant. This proposal suggests we should warn when some syntax might theoretically be stolen in the future.

Note that I, myself, am not convinced that this is a good idea, but I thought having this proposal would give the community a good opportunity to debate this general direction of travel.

[Rendered](https://github.com/goldfirere/ghc-proposals/blob/88690a542ada41dccd0f58d252ed6d392140034d/proposals/0000-prefix-suffix-operators.rst)


Ericson2314:
I'm tempted to say we also warn on "tight infix". I'm not very excited about stealing more syntax, but I am jealous of Agda's super simple lexing which felt it it was just spiting on whitespace and deliminators, so it needs a far smaller list of special tokens. This moves us in that direction.

int-index:
Note that the original proposal already adds two warning flags, but neither is implemented at the moment.

* `-Woperator-whitespace-ext-conflict` covers syntax that is stolen by extensions that already exist but are not enabled.
* `-Woperator-whitespace` covers all prefix, suffix, and tight infix usages, enforcing the best practice of using the loose infix form for operator occurrences.

These are already accepted, so I think this proposal should be a simple amendment that puts `-Woperator-whitespace` into `-Wall`.

int-index:
> Note that the only operator with a special tight infix meaning is `@`, which cannot be used as a user-definable operator.

Oh, but it can, if you write it in loose infix form. That's one of the consequences of the original proposal.

goldfirere:
> Note that the original proposal already adds two warning flags, but neither is implemented at the moment.

If neither is implemented, then we should note that in the proposal. I will do so shortly, along with recharacterizing this proposal as an amendment.

goldfirere:
I've updated this to be a simple amendment to the old proposal. Thanks, @int-index, for the pointer.

anka-213:
I think the main cause of annoyance for this proposal would be infix sections, where we can no longer write code like this:
```
fmap (+1) [1..5]
```
without getting compiler warnings. Or is it just intended to warn about binary use of the operators, not sections? (If so, the proposal should mention this)

On another note, has anyone done a search on Hackage for how much code would be affected by this warning? It's not a very big deal though, since it should be easy to automatically fix this warning.

int-index:
> fmap (+1) [1..5]

Somebody less familiar with Haskell might read this as an explicit marker that `1` is a positive number here (i.e. `-1` vs `+1`), rather than an operator section, so it would be a good change to nudge the programmer towards writing it as `fmap (+ 1)`, I believe.

goldfirere:
I'd like to submit this to the committee, @nomeata.

i-am-tom:
Having read the original proposal, I think this is a fairly uncontroversial change. It seems like a good idea; I'll put it to the committee!

simonpj:
No objections from me

i-am-tom:
As no member of the committee voiced a concern or voted against this proposal, I think we ought to mark it accepted :)

Paging @nomeata

phadej:
Why `c*(4)` is considered acceptable? We can write

```
Prelude GHC.Exts> Ptr(plusAddr#("foo"#)(3#))
0x0000000001e3d8b3
```

which has different meaning without `MagicHash`.

---

Also I think there is a bug in GHC as (with  `MagicHash`) the following is "accepted":

```
Prelude GHC.Exts> plusAddr#foo

<interactive>:15:10: error: Variable not in scope: foo :: Addr#
```

I.e. there is no whitespace in between, but it is still an application. I understand that it is clear from the token stream perspective, but then `a-b` could be either `a - b` or `a- b` if `-` would be a post-sigil.

int-index:
> Why `c*(4)` is considered acceptable?

Because of https://gitlab.haskell.org/ghc/ghc/-/issues/19372

phadej:
Let me rephrase the question

Why `c+(4)` is considered acceptable?

int-index:
Is it? That's a "tight infix" use of `+` and should trigger `-Woperator-whitespace` (I'm not at a computer right now to verify)

phadej:
Screenshot from 2021-02-15 11-39-37

I interpret the last point as not being specific to `*`, but apply to any varsym/consym.

int-index:
Ah, the `Rendered` link does not lead where it should. That's an outdated version of the proposal. We should also check if the committee members who accepted this proposal were not misled by the ourdated link.

i-am-tom:
Maybe due to some bizarre muscle memory, I get to the rendered version via "view file" on the "files changed" tab, so I seem to have accidentally managed to avoid this pitfall. I can't speak for the others, though I suspect that's a question for the initial proposal, rather than this specific amendment to the error/warning convention.

PR #369, opened 2020-10-04
Add sumToTag# primop (under review)

treeowl:
Add a `sumToTag#` primop for extracting the numerical value
of an alternative of an unboxed sum type.

[Rendered proposal](https://github.com/treeowl/ghc-proposals/blob/sum-to-tag/proposals/0000-sum-to-tag.md)

### DIFF
--- a/proposals/0000-sum-to-tag.md
+++ b/proposals/0000-sum-to-tag.md
@@ -122,6 +122,20 @@ case a of
   (# ||_ #) -> e2
 ```
 
+One long-term option this primop gives us is to stop returning
+`Int#` from primops whose values represent only two or three
+specific values. For example, we would have the option of
+writing
+
+```haskell
+(==#) :: Int# -> Int# -> (# (##) | (##) #)
+```
+
+which more faithfully represents what the return value can be,
+while still being able to work with the `Int#` representation
+directly using `sumToTag#`. I do *not* propose making such a
+change at this time.
+
 ## Costs and Drawbacks
 
 Implementing this primop should be extremely cheap. Implementing
--- a/proposals/0000-sum-to-tag.md
+++ b/proposals/0000-sum-to-tag.md
@@ -71,12 +71,21 @@ needed is a simple decrement operation.
 
 ## Proposed Change Specification
 
-Add a `sumToTag#` primop much like `dataToTag#`. Much like `dataToTag#` can be
-compiled only when the *type constructor* of its argument is known and is an
-algebraic datatype, `sumToTag#` can be compiled only when the *runtime
-representation* of its argument is known and is `'SumRep xs` for some known
-`xs`. `sumToTag#` will return the 0-based index of the sum alternative of its
+Add a `sumToTag#` inline primop, analogous to `dataToTag#`. Approximately,
+
+```haskell
+sumToTag# :: forall (xs :: RuntimeRep) (a :: TYPE ('SumRep xs)). a -> Int#
+```
+
+`sumToTag#` will return the 0-based index of the sum alternative of its
 argument.
+
+The `forall` is a bit of a lie: much like `tagToEnum#` can be compiled only
+when the *type constructor* of its argument is known and is an algebraic
+datatype, `sumToTag#` can be compiled only when the *runtime representation* of
+its argument is known and is `'SumRep xs` for some known `xs`. This restriction
+doesn't really hurt in practice, because bindings can't be
+representation-polymorphic anyway.
 
 ## Examples
 
@@ -106,9 +115,16 @@ bar = sumToTag#
 
 ## Effect and Interactions
 
-It will be possible to extract tags efficiently. It would be
-beneficial to implement a special rule for `case` of `sumToTag#`, in
-case that should appear in the course of optimization.
+It will be possible to extract tags efficiently. With the current
+implementation of unboxed sums, we would compile to code looking
+something like
+
+```haskell
+sumToTag# (# tg1, ... #) = tg1 -# 1#
+```
+
+It would be beneficial to implement a special rule for `case` of `sumToTag#`,
+in case that should appear in the course of optimization.
 
 ```haskell
 case sumToTag# a of
--- a/proposals/0000-sum-to-tag.md
+++ b/proposals/0000-sum-to-tag.md
@@ -145,6 +145,13 @@ the special `case` rule will probably not be terribly hard; its
 structure should generally follow that of the rule for `case` of
 `dataToTag#`.
 
+The main (theoretical) drawback I see is that `sumToTag#` weakens
+parametricity. Today, we can reason that a function polymorphic over its type
+can't inspect its argument. With `sumToTag#`, it can do so, to a limited
+extent. In practice, however, primops already break parametricity, and users
+who don't have access to principled ones like `tagToSum#` may well reach for
+more dangerous unsafe coercions to get the job done.
+
 ## Alternatives
 
 Joachim Breitner asks, "Could this be solved by an optimization in a later
--- a/proposals/0000-sum-to-tag.md
+++ b/proposals/0000-sum-to-tag.md
@@ -81,10 +81,12 @@ argument.
 ## Examples
 
 ```haskell
-sumToTag# ((# | 2# | #) :: (# (# Int, Char #) | Int# | Bool #)) = 1#
+sumToTag# ((# | 2# | #)
+             :: (# (# Int, Char #) | Int# | Bool #)) = 1#
 
 -- No need to pin down specific types
-sumToTag# ((# | x | #) :: (# a | x | b :: TYPE 'UnliftedRep #)) = 1#
+sumToTag# ((# | x | #)
+             :: (# a | x | b :: TYPE 'UnliftedRep #)) = 1#
 ```
 
 This will compile, but can never ever be called:
@@ -176,6 +178,35 @@ b + sumToTag# a
 This all seems much too hard.
 
 ## Unresolved Questions
+
+Should we also offer `tagToSum#`, mirroring `tagToEnum#`? Conceptually,
+
+```haskell
+tagToSum#
+  :: forall (a :: TYPE ('SumRep '[ 'TupleRep '[]
+                                 , 'TupleRep '[], ...])).
+  Int# -> a
+
+-- For example:
+
+tagToSum# :: Int# -> (# (##) | (##) | (##) #)
+tagToSum# 0# = (# (##) || #)
+tagToSum# 1# = (# | (##) | #)
+```
+
+where the type-level list has at least one element. Including `tagToSum#` seems
+nice for symmetry's sake, but I have not yet found a situation where it's truly
+necessary. Note that unlike `sumToTag#`, `tagToSum#` would be unsafe in two
+different ways:
+
+1. The user could supply an out-of-bounds `Int#` value.
+2. The user could supply an uninhabited type:
+
+    ```haskell
+    newtype Uninhab :: TYPE ('SumRep '[ 'TupleRep '[]
+                                      , 'TupleRep '[]]) where
+      Uninhab :: Uninhab -> Uninhab
+    ```
 
 ## Implementation Plan
 
--- a/proposals/0000-sum-to-tag.md
+++ b/proposals/0000-sum-to-tag.md
@@ -74,17 +74,17 @@ needed is a simple decrement operation.
 Add a `sumToTag#` inline primop, analogous to `dataToTag#`. Approximately,
 
 ```haskell
-sumToTag# :: forall (xs :: RuntimeRep) (a :: TYPE ('SumRep xs)). a -> Int#
+sumToTag# :: forall (xs :: [RuntimeRep]) (a :: TYPE ('SumRep xs)). a -> Int#
 ```
 
 `sumToTag#` will return the 0-based index of the sum alternative of its
 argument.
 
-The `forall` is a bit of a lie: much like `tagToEnum#` can be compiled only
-when the *type constructor* of its argument is known and is an algebraic
+The `forall` feels like a bit of a lie: much like `tagToEnum#` can be compiled
+only when the *type constructor* of its argument is known and is an algebraic
 datatype, `sumToTag#` can be compiled only when the *runtime representation* of
 its argument is known and is `'SumRep xs` for some known `xs`. This restriction
-doesn't really hurt in practice, because bindings can't be
+doesn't actually have any effect in practice, because bindings can't be
 representation-polymorphic anyway.
 
 ## Examples
--- a/proposals/0000-sum-to-tag.md
+++ b/proposals/0000-sum-to-tag.md
@@ -194,10 +194,8 @@ tagToSum# 0# = (# (##) || #)
 tagToSum# 1# = (# | (##) | #)
 ```
 
-where the type-level list has at least one element. Including `tagToSum#` seems
-nice for symmetry's sake, but I have not yet found a situation where it's truly
-necessary. Note that unlike `sumToTag#`, `tagToSum#` would be unsafe in two
-different ways:
+where the type-level list has at least one element. Unlike
+`sumToTag#`, `tagToSum#` would be unsafe in two different ways:
 
 1. The user could supply an out-of-bounds `Int#` value.
 2. The user could supply an uninhabited type:
@@ -207,6 +205,20 @@ different ways:
                                       , 'TupleRep '[]]) where
       Uninhab :: Uninhab -> Uninhab
     ```
+
+Including `tagToSum#` seems nice for symmetry's sake, but I have not yet found
+a situation where it's truly necessary. In the `Trit#` example above, it would
+be a natural way to deal with certain arithmetic results:
+
+```haskell
+lowTrit :: Int# -> Trit#
+lowTrit x = tagToSum# (x `remInt#` 3#)
+```
+
+But when would we really benefit from this? Most of the time we'll just
+perform a `case` match, and case-of-case will clean everything up. The
+rest of the time, we'll probably be allocating memory and the conversion
+will be the least of our worries.
 
 ## Implementation Plan
 
--- a/proposals/0000-sum-to-tag.md
+++ b/proposals/0000-sum-to-tag.md
@@ -145,7 +145,35 @@ structure should generally follow that of the rule for `case` of
 
 ## Alternatives
 
-I do not know of any alternative approach.
+Joachim Breitner asks, "Could this be solved by an optimization in a later
+phase that detects this pattern, and produces the code that you want?"
+Perhaps. Such an optimization would have to be in the C-- stage, which
+is rather late in the game. It would have to be reminiscent of
+common subexpression elimination, but I believe it would be considerably
+more complicated. Imagine what this would look like in Haskell: If we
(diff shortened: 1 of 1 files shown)


proposals/0000-sum-to-tag.md
```diff
@@ -0,0 +1,165 @@
...
+## Costs and Drawbacks
+
+Implementing this primop should be extremely cheap. Implementing
+the special `case` rule will probably not be terribly hard; its
+structure should generally follow that of the rule for `case` of
+`dataToTag#`.
+
+## Alternatives
+
+I do not know of any alternative approach.
```
nomeata, line 148: Could this be solved by an optimization in a later phase that detects this pattern, and produces the code that you want?

treeowl: @nomeata that was the first thing I wondered, but @bgamari didn't think it would likely be worth the trouble. I don't personally know what that would look like, but it would presumably require somewhat complicated analysis of almost-common subexpressions in CPP. Even if I knew how to do it, it sounds expensive at compilation time.

treeowl: @nomeata, I've addressed your question in the proposal text.

nomeata: Thanks!

treeowl:
@nomeata, as a week has gone by with no comments, I submit this proposal to the committee.

treeowl:
@nomeata, will you please assign a shepherd?

nomeata:
Ups, sorry, this must have slipped through my radar, thanks for the nudge.

aspiwack:
@treeowl in the change specification section, can you specify the precise type of `sumToTag#`?

treeowl:
> @treeowl in the change specification section, can you specify the precise type of `sumToTag#`?

Like `dataToTag#`, it's impossible to give it a precise type in Haskell. The best we can do is

```haskell
sumToTag# :: forall xs (a :: TYPE ('SumRep xs)). a -> Int#
```

But the `forall` is a lie—`xs` must be fully determined at compile time.

aspiwack:
This reminds me of [#17201 on Gitlab](https://gitlab.haskell.org/ghc/ghc/-/issues/17201). I don't think there is anything wrong with `dataToTag#`, actually, maybe you're speaking of `tagToEnum#` (whose specification I'm not sure of)?

Nevertheless, the specification section should not leave the reader guessing as to what, precisely, you are proposing.

treeowl:
> This reminds me of [#17201 on Gitlab](https://gitlab.haskell.org/ghc/ghc/-/issues/17201). I don't think there is anything wrong with `dataToTag#`, actually, maybe you're speaking of `tagToEnum#` (whose specification I'm not sure of)?
> 
> Nevertheless, the specification section should not leave the reader guessing as to what, precisely, you are proposing.



> This reminds me of [#17201 on Gitlab](https://gitlab.haskell.org/ghc/ghc/-/issues/17201). I don't think there is anything wrong with `dataToTag#`, actually, maybe you're speaking of `tagToEnum#` (whose specification I'm not sure of)?
> 
> Nevertheless, the specification section should not leave the reader guessing as to what, precisely, you are proposing.

Unfortunately, I don't understand this comment. Could you expand? How does it relate to that GitHub issue? What aspect of the proposal don't you understand? Yes, I think I did mix up an aspect of `dataToTag#` with one of `tagToEnum#`. The proposed primop would have type checking requirements analogous to `tagToEnum#`. I will edit the proposal to reflect that.

aspiwack:
> Unfortunately, I don't understand this comment. Could you expand? How does it relate to that GitHub issue?

Sure. In the Gitlab issue link, the discussion veers towards creating a new kind of constraint `FixedRuntimeRep :: RuntimRep -> Constraint`, whose role is precisely to ensure that a runtime representation argument is fully determined. It's a similar constraint to what you would need to give a precise type to `sumToTag#` if I understand correctly.

treeowl:
@aspiwack, I hope the commit I pushed addresses your concerns.

proposals/0000-sum-to-tag.md
```diff
@@ -0,0 +1,259 @@
...
+will be extra code and unnecessary conditional branches when all that's
+needed is a simple decrement operation.
+
+## Proposed Change Specification
+
+Add a `sumToTag#` inline primop, analogous to `dataToTag#`. Approximately,
+
+```haskell
+sumToTag# :: forall (xs :: RuntimeRep) (a :: TYPE ('SumRep xs)). a -> Int#
+```
```
aspiwack, line 78: Why “approximately”? I assume it will be the type of `sumToTag#` in `GHC.Prim`. Just say so.

proposals/0000-sum-to-tag.md
```diff
@@ -0,0 +1,259 @@
...
+
+`sumToTag#` will return the 0-based index of the sum alternative of its
+argument.
+
+The `forall` is a bit of a lie: much like `tagToEnum#` can be compiled only
+when the *type constructor* of its argument is known and is an algebraic
+datatype, `sumToTag#` can be compiled only when the *runtime representation* of
+its argument is known and is `'SumRep xs` for some known `xs`. This restriction
+doesn't really hurt in practice, because bindings can't be
+representation-polymorphic anyway.
```
aspiwack, line 88: Does it ever hurt, actually? The representation polymorphism restriction is precisely made to make functions which depend on a representation to be implementable.

I'm guessing that the innovation here is that `sumToTag#` is possibly the first non-data-constructor function to be representation polymorphic. But does it really have any real consequence?

proposals/0000-sum-to-tag.md
treeowl: It seems weird to me to call this `forall xs` when it feels a bit more like `foreach xs` or something. But I'm not an expert.

proposals/0000-sum-to-tag.md
treeowl: I think you're probably right.

gridaphobe: Are there applications of `sumToTag#` that would type-check but nonetheless have to be rejected? This aside about the `forall` being a lie and the talk about `tagToEnum#` make it sound like you think the answer is yes, but I'm not so sure. It is possible to have representation polymorphic expressions, e.g. `error "urk"`, but they all diverge. `sumToTag#` is strict in its argument, so it seems to me that the type you've given to `sumToTag#` is exactly right.

proposals/0000-sum-to-tag.md
gridaphobe: The type should be 

```hs
sumToTag# :: forall (xs :: [RuntimeRep]) (a :: TYPE ('SumRep xs)). a -> Int#
```

in order to be well-kinded, right?

proposals/0000-sum-to-tag.md
```diff
@@ -0,0 +1,259 @@
...
+The values look like this:
+
+```haskell
+(# (##) | #)   ~=   (# 1#, padding #)
+(# | b #)      ~=   (# 2#, b #)
+```
+
+Note: the choice to make the first alternative `1#` rather than `0#`
+is poorly motivated, and should probably be changed, but that's beyond
+the strict scope of this proposal.
```
gridaphobe, line 46: Is the one-indexed tag for unboxed sums currently exposed in any way? I agree that zero-indexing makes more sense, but it could be pretty confusing if the actual representation were already exposed.

proposals/0000-sum-to-tag.md
```diff
@@ -0,0 +1,259 @@
...
+For example, I might write
+
+```haskell
+newtype Trit# = Trit# (# (##) | (##) | (##) #)
+
+tritToInt# :: Trit# -> Int#
+tritToInt# (Trit# (# (##)|| #)) = 0#
+tritToInt# (Trit# (# |(##)| #)) = 1#
+tritToInt# (Trit# (# ||(##) #)) = 2#
+```
```
gridaphobe, line 32: Do you have an example of where you might want to call `tritToInt#` in real code? Consistency with the API around boxed sums and `dataToTag#` is good enough for me to motivate this proposal, but I'm genuinely curious about where these functions become useful.

proposals/0000-sum-to-tag.md
treeowl: Yes, I'll fix that.

proposals/0000-sum-to-tag.md
treeowl: I think you're right. It *feels* like it's not really `forall`, but I think in practice it might as well be.

proposals/0000-sum-to-tag.md
treeowl: No, it's not. I strongly suspect it should be changed to 0-indexed internally, but that's out of the scope of the proposal process.

proposals/0000-sum-to-tag.md
treeowl: No, I don't have anything immediate. But in #367 I speculate about possibly using unboxed sums to express, e.g., Boolean results from primops. If that should ever happen, then we may well want to be able to write things like

```haskell
3# +# sumToTag# (x ==# 12)
```

Even today, some people may choose to write functions producing such unboxed sums (to express the finite number of return values), and may not want them to have to inline those to get free conversion to `Int#` when they want to do arithmetic.

gridaphobe:
Thanks for the updates @treeowl.

@ghc-proposals/ghc-steering-committee I recommend accepting this proposal. It's a simple addition that fills a gap in the API for unboxed sums (we already have `dataToTag#` for lifted types). 

The Unresolved Questions section discusses a possible `tagToSum#` function analogous to `tagToEnum#`. For the same reasons mentioned in the proposal, I recommend we do not implement `tagToSum#` at this time.

simonpj:
> The forall feels like a bit of a lie: much like tagToEnum# can be compiled only when..

I think you mean `dataToTag#`.

>  the type constructor of its argument is known and is an algebraic datatype, sumToTag# can be compiled only when the runtime representation of its argument is known and is 'SumRep xs for some known xs.

What does "known" mean?   I think you mean, more precisely
* the first (type) argument to `sumToTag#` must be closed (i.e. have no free variables)

So (after type inference is complete) you can have 
* `sumToTag# @[IntRep]` or
* `sumToTag# @[IntRep, IntRep, PtrRep]`,
* `sumToTag# @[IntRep, TupleRep[PtrRep,PtrRep]]`
but NOT
* `sumToTag# @r`, or
* `sumToTag# @[r,IntRep]`

> This restriction doesn't actually have any effect in practice, because bindings can't be representation-polymorphic anyway.

I'm really not sure what this means.  You could say `sumToTag# @r (error "urk")`, where `r` is some in-scope type variable.

simonpj:
I'm ok with this proposal, modulo the clarifications above.

simonpj:
> Note: the choice to make the first alternative 1# rather than 0# is poorly motivated, and should probably be changed, but that's beyond the strict scope of this proposal.

I think it's because data constructor tags start at 1; we use 0 for "thunk".

treeowl:
> > The forall feels like a bit of a lie: much like tagToEnum# can be compiled only when..
> 
> I think you mean `dataToTag#`.

No, `dataToTag#` isn't at all fussy, is it? But we need to know the result type to compile `tagToEnum#`.

> > the type constructor of its argument is known and is an algebraic datatype, sumToTag# can be compiled only when the runtime representation of its argument is known and is 'SumRep xs for some known xs.
> 
> What does "known" mean? I think you mean, more precisely
> 
> * the first (type) argument to `sumToTag#` must be closed (i.e. have no free variables)

That could work.

> So (after type inference is complete) you can have
> 
> * `sumToTag# @[IntRep]` or
> * `sumToTag# @[IntRep, IntRep, PtrRep]`,
> * `sumToTag# @[IntRep, TupleRep[PtrRep,PtrRep]]`
>   but NOT
> * `sumToTag# @r`, or
> * `sumToTag# @[r,IntRep]`

Yes.

> > This restriction doesn't actually have any effect in practice, because bindings can't be representation-polymorphic anyway.
> 
> I'm really not sure what this means. You could say `sumToTag# @r (error "urk")`, where `r` is some in-scope type variable.

Ah, I suppose you could! And that would actually be fine, wouldn't it? So maybe my well-intentioned restriction isn't actually needed at all!

treeowl:
> > Note: the choice to make the first alternative 1# rather than 0# is poorly motivated, and should probably be changed, but that's beyond the strict scope of this proposal.
> 
> I think it's because data constructor tags start at 1; we use 0 for "thunk".

Yes, I have heard it was by analogy to the way it's done for lifted sums, but we don't *have* thunks for unboxed sums, so we don't need that. If we make them 0-indexed instead, and work from low to high instead of high to low, I believe we should sometimes be able to drop `cmp` instructions (probably in peephole optimization). Again, out of scope for the proposal.

gridaphobe:
> What does "known" mean? I think you mean, more precisely
> 
> * the first (type) argument to `sumToTag#` must be closed (i.e. have no free variables)
> 
> So (after type inference is complete) you can have
> 
> * `sumToTag# @[IntRep]` or
> * `sumToTag# @[IntRep, IntRep, PtrRep]`,
> * `sumToTag# @[IntRep, TupleRep[PtrRep,PtrRep]]`
>   but NOT
> * `sumToTag# @r`, or
> * `sumToTag# @[r,IntRep]`

Are there any non-diverging representation-polymorphic expressions? The only way I know of to produce a representation-polymorphic expression is with `error` and `undefined`. That's why I suggested above that the restriction, while well-intentioned, seems ultimately unnecessary.

gridaphobe:
> > Note: the choice to make the first alternative 1# rather than 0# is poorly motivated, and should probably be changed, but that's beyond the strict scope of this proposal.
> 
> I think it's because data constructor tags start at 1; we use 0 for "thunk".

Are the 1-indexed tags exposed via `dataToTag#`, or do we reindex from 0 since `dataToTag#` requires its argument to be evaluated? I think consistency between the two APIs would be good.

treeowl:
> Are there any non-diverging representation-polymorphic expressions? The only way I know of to produce a representation-polymorphic expression is with `error` and `undefined`. That's why I suggested above that the restriction, while well-intentioned, seems ultimately unnecessary.

No, there are not. I think it *should* be possible, in principle, to support something like

```haskell
f :: forall (xs :: [RuntimeRep]) (a :: TYPE ('SumRep xs)). (() -> a) -> Int#
f err = sumToTag# (err ())
```

But how does it compile, exactly? The first step is easy:

```haskell
-- STEP 1
f err = case err () of {
  x -> sumToTag# x }
```

Since the result of `err` is polymorphic in runtime representation, it must throw an exception, so we can rewrite this to

```haskell
f err = case err () of { }
```

This is fine, but we've already hit a problem in step 1: `x` is a representation-polymorphic binding, which is prohibited. So we have to push the bottom-detection up *earlier* in the compilation process: somewhere right around type checking or desugaring. I'm not a compiler developer, but this all smells bad to me.

treeowl:
> Are the 1-indexed tags exposed via `dataToTag#`, or do we reindex from 0 since `dataToTag#` requires its argument to be evaluated? I think consistency between the two APIs would be good.

No, the 1-indexed tags for lifted types are reindexed to 0 by `dataToTag#`.

treeowl:
Let me be totally clear:

```haskell
dataToTag# False = 0#
dataToTag# True = 1#
```

In the current proposal,

```haskell
sumToTag (# (##) | #) = 0#
sumToTag (# | (##) #) = 1#
```

They are consistent.

gridaphobe:
> I think it _should_ be possible, in principle, to support something like
> 
> ```haskell
> f :: forall (xs :: [RuntimeRep]) (a :: TYPE ('SumRep xs)). (() -> a) -> Int#
> f err = sumToTag# (err ())
> ```
> 
> But how does it compile, exactly?

We can ask GHC right now!

```hs
sumToTag :: forall (xs :: [RuntimeRep]) (a :: TYPE ('SumRep xs)). a -> Int
sumToTag = undefined

urk :: forall (xs :: [RuntimeRep]) (a :: TYPE ('SumRep xs)). (() -> a) -> Int
urk err = sumToTag (err ())
```

```
> ghc-8.10 Tags.hs
[1 of 1] Compiling Tags             ( Tags.hs, Tags.o )

Tags.hs:14:20: error:
    A levity-polymorphic type is not allowed here:
      Type: a
      Kind: TYPE ('SumRep xs)
    In the type of expression: (err ())
   |
14 | urk err = sumToTag (err ())
   |                    ^^^^^^^^
```

It looks like GHC already has a blanket restriction on representation-polymorphic arguments. We can't generate code for passing the argument without knowing its representation, so this makes sense. And it helpfully sidesteps the question of whether `sumToTag#` should accept representation-polymorphic arguments, it already can't!

With that in mind, I would revise the last paragraph in the "Proposed Change Specification" to explain that the question of how to deal with representation-polymorphic arguments is already addressed by existing restrictions. And you can also remove the "Approximately" when giving the type, the type is exactly right. I think that will address the concerns raised so far.

simonpj:
> It looks like GHC already has a blanket restriction on representation-polymorphic arguments. We can't generate code for passing the argument without knowing its representation, so this makes sense.

Yes exactly.  You could refer (in the proposal) to Section 5 of [Levity Polymorphism](https://www.microsoft.com/en-us/research/publication/levity-polymorphism/)

proposals/0000-sum-to-tag.md
```diff
@@ -0,0 +1,259 @@
...
+
+`sumToTag#` will return the 0-based index of the sum alternative of its
+argument.
+
+The `forall` feels like a bit of a lie: much like `tagToEnum#` can be compiled
+only when the *type constructor* of its argument is known and is an algebraic
+datatype, `sumToTag#` can be compiled only when the *runtime representation* of
+its argument is known and is `'SumRep xs` for some known `xs`. This restriction
+doesn't actually have any effect in practice, because bindings can't be
+representation-polymorphic anyway.
```
gridaphobe, line 88:
```suggestion
It may seem bizarre to have a function that is polymorphic in the runtime representation of its argument; we could only call such a function when we know the concrete representation of its argument, i.e. when the `xs :: [RuntimeRep]` type argument is *closed*. However, GHC already imposes a universal ban on levity-polymorphic arguments (see Section 5.1 of [Levity Polymorphism][]) so we do not have to worry about this case.

[Levity Polymorphism]: https://www.microsoft.com/en-us/research/publication/levity-polymorphism/
```

gridaphobe (commented):
@treeowl the feedback from the Committee has been positive, with a couple comments that the Specification could be clarified a bit more. I've made some suggested edits that I hope will tighten the language. If they look good to you (and @simonpj and @aspiwack), I think we can merge this proposal.

aspiwack:
As far as I'm concerned, the suggested edits are good.

gridaphobe:
@treeowl have you had a chance to look at my suggested edits? I think that's the only thing blocking us from merging this proposal.

nomeata:
@treeowl ping?

treeowl:
Oy. I'll try to revise tonight.

gridaphobe:
@treeowl I went ahead and applied the wording changes we suggested in https://github.com/ghc-proposals/ghc-proposals/pull/369/commits/d0ecec097994ce940c54378e2ee3c1cdf33ae1ba. Unless you have any objections, we'll consider this proposal accepted and merge.

treeowl:
No objection, and thanks.

PR #370, opened 2020-10-12
A syntax for Modifiers

goldfirere:
The [proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0370-modifiers.rst) has been accepted; the following discussion is mostly of historic interest.

---
This proposes a new syntax for *modifiers*, a prefix word or expression that modifies the way a bit of syntax behaves. The only modifier available right now is the multiplicity of an arrow, but #242 will want a new modifier, and the proposal outlines other ways this new feature might be used.

The primary motivation behind this is to create a palatable way to combine linear types with unsaturated type families, and is the first time I feel happy about the way these two features will interact.

[Rendered](https://github.com/goldfirere/ghc-proposals/blob/modifiers/proposals/0000-modifiers.rst)

### DIFF
--- a/proposals/0000-modifiers.rst
+++ b/proposals/0000-modifiers.rst
@@ -98,6 +98,9 @@ Proposed Change Specification
     Any other use of a modifier is an error.
 
 12. ``-XLinearTypes`` implies ``-XModifiers``.
+
+13. GHC will never infer quantifying over a ``Modifier`` constraint. No modifier
+    polymorphism!
   
 Examples
 --------
@@ -166,7 +169,7 @@ Effect and Interactions
 * Some other features that have had tortuous and torturous syntax
   discussions may have enjoyed having the modifier option. For example,
   this might have been used instead of ``type role`` for role annotations:
-  ``data Tagged (%nominal t) a = Tagged a``. Or it might have been an
+  ``data Tagged (%Nominal t) a = Tagged a``. Or it might have been an
   alternative for ``-XDerivingStrategies``.
 
 * Though not proposed here, we can imagine extensions allowing abstractions
@@ -187,7 +190,8 @@ Costs and Drawbacks
 -------------------
 * The loss of the inferred kind of ``m`` in multiplicity polymorphism is a
   drawback. However, a user seeing ``Int %m -> Bool`` is hard-pressed to
-  understand what is going on. On the other hand, ``Int %(m :: Multiplicity) -> Bool``
+  understand what is going on. On the other hand, labeling ``m :: Multiplicity``
+  explicitly (either in the binding for ``m`` or in a usage site)
   is much more perspicuous.
 
 * Any feature has a maintenance burden, but this one should be fairly small.
--- a/proposals/0000-modifiers.rst
+++ b/proposals/0000-modifiers.rst
@@ -13,8 +13,8 @@ Syntax for Modifiers: a generalization of linear-types syntax
 .. contents::
 
 This proposal introduces a new form of syntax ``%blah`` that defines a *modifier*.
-Modifiers somehow change the meaning of the next token. The ``blah`` must have
-a type which is a member of the new class ``Modifier``. For now, all modifiers
+Modifiers somehow change the meaning of the next token. The ``blah`` is parsed
+and renamed as a type. For now, all modifiers
 will be built in, but we might imagine making an extensible feature later.
 
 As of the writing of this proposal, there will be precisely one modifier: the
@@ -79,28 +79,24 @@ Proposed Change Specification
 7. Reserve the use of ``%`` in a prefix occurrence to be used only for modifiers;
    though this proposal does not do so, we can imagine extending the modifier syntax
    to apply to further syntactic situations (e.g. term-level operators, declarations,
-   import lists, etc.). The one exception is the syntax ``%1`` for a linear function,
-   which continues to be allowed.
+   import lists, etc.).
 
-8. Introduce a new type-level constant ``Modifier :: Type -> Constraint``, exported
-   from ``GHC.Exts``.
+8. The type of a modifier is determined only by synthesis, never by checking.
+   That is, in the bidirectional type-checking scheme used by GHC, we find the
+   type of the modifier by running the synthesis judgment. Effectively, this
+   means that if we consider a modifier to be some head (constructor or
+   variable) applied to a sequence of arguments (possibly none), the head must
+   have a known type: constructors always have a known type, and variables
+   have a known type if declared with a type signature. Alternatively, the
+   modifier may have a top-level type signature.
 
-9. Let the constraint ``Modifier Multiplicity`` be satisfiable; let no other
-   ``Modifier`` constraint be satisfiable.
+9. A modifier of type ``Multiplicity`` changes the multiplicity of the following arrow,
+   preceding pattern-bound variable (but only on the top level of a lambda pattern),
+   or preceding record field.
+   Multiple modifiers of type ``Multiplicity`` on the same arrow are not allowed.
+   Any other use of a modifier is an error.
 
-10. During constraint generation, let an occurrence ``%(ty)``, where ``ty :: ki``,
-    emit a constraint ``Modifier ki``.
-
-11. A modifier of type ``Multiplicity`` changes the multiplicity of the following arrow,
-    preceding pattern-bound variable (but only on the top level of a lambda pattern),
-    or preceding record field.
-    Multiple modifiers of type ``Multiplicity`` on the same arrow are not allowed.
-    Any other use of a modifier is an error.
-
-12. ``-XLinearTypes`` implies ``-XModifiers``.
-
-13. GHC will never infer quantifying over a ``Modifier`` constraint. No modifier
-    polymorphism!
+10. ``-XLinearTypes`` implies ``-XModifiers``.
   
 Examples
 --------
@@ -108,14 +104,16 @@ Here are some examples that will be accepted or rejected with this proposal::
 
   f1 :: Int %1 -> Bool    -- unaffected, actually: that "%1" is one lexeme, and
                           -- is not a modifier. See more on this below.
-  f2 :: Int %Many -> Bool -- accepted: Many :: Multiplicity, and Modifier Multiplicity holds
-  f3 :: Int %m -> Bool    -- rejected: the kind of m is ambiguous
-  f4 :: Int %(m :: Multiplicity) -> Bool   -- accepted
+  f2 :: Int %Many -> Bool -- accepted: Many :: Multiplicity
+  f3 :: Int %m -> Bool    -- rejected: the kind of m is undeclared
+  f4 :: Int %(m :: Multiplicity) -> Bool   -- accepted with a type signature
   f5 :: Int %One %Many -> Bool   -- rejected (although it will parse)
   f6 :: Int %Many %Many -> Bool  -- rejected
+  f7 :: Int %(m :: Multiplicity) -> Int %m -> Int
+    -- rejected: the second use of '%m' has an unknown type
 
   map :: forall (m :: Multiplicity). (a %m -> b) -> [a] %m -> [b]
-    -- the kind annotation is really on m, not on the modifier
+    -- accepted: m has a known type
 
 The syntax (and semantics) for modifiers on patterns and record fields is exactly
 as described in the `linear types proposal`_.
@@ -124,8 +122,8 @@ as described in the `linear types proposal`_.
   
 Effect and Interactions
 -----------------------
-* It is expected that the matchability of `#242`_ will have a kind ``Matchability``,
-  and that ``Modifier Matchability`` will be satisfiable. Then, users will be able
+* It is expected that the matchability of `#242`_ will have a kind ``Matchability``.
+  Then, users will be able
   to write ``Int %Many %Matchable -> Bool`` or ``Int %Matchable %Many -> Bool``.
   The details are left to `#242`_ (assuming this proposal is accepted first).
   The author of `#242`_, Csongor Kiss, was involved in the conceptualization of
@@ -135,15 +133,15 @@ Effect and Interactions
   is not expected to matter (though that would be up to other proposals to
   spell out).
 
-* Let's assume we have overloaded numbers at the type level, and then consider
-  ``%1``. Under this proposal, we would have ``1 :: a`` where ``Num a`` and
-  ``Modifier a`` must hold. If we have ``Multiplicity`` specified at the end of
-  the ambient ``default``\ing list, then ``Multiplicity`` will be the first
-  (and only) member of that list that satisfies both ``Num`` and ``Modifier``.
-  Accordingly, GHC will default ``a`` to be ``Multiplicity``, and all will be
-  well. (We may want this case to avoid activating ``-Wtype-defaults``, but
-  that's a conversation for later.)
+* The ``%1`` will remain a single lexeme and does not participate with this
+  proposal. We may want more exceptions to the general scheme in the future.
 
+* The key action of this proposal is to carve out a new syntax space, anchored
+  by a prefix occurrence of ``%``. Ideally, there would be few exceptions to
+  the general scheme (but ``%1`` is one such exception). It is possible that
+  future extensions to this idea will be disambiguated before the type checker
+  gets a chance to do its work.
+  
 * This proposal means that ``Int %m -> Bool``, acceptable today as a
   multiplicity-polymorphic function, would be rejected. The user would need
   to add a kind annotation to tell us that ``m`` is a multiplicity (and not,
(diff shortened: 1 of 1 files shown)


jvanbruegge:
I like this, it cleans up this interaction quite well. I also support not defaulting to Multiplicity as I for example would use Matchability more than Multiplicity (and it just feels wrong to arbitrarily default to one)

Just one question: The `m` is a normal type variable, isn't it? So instead of a kind signature directly with the modifier anywhere else should be fine too, is it? So it is just that the modifier by itself does not allow ghc to infer a kind and thus the kind needs to be user supplied.

int-index:
Yes, it would be helpful to show that one `:: Multiplicity` annotation is sufficient:

```
map :: forall (p :: Multiplicity) a b. (a %p-> b) -> [a] %p-> [b]
```

jvanbruegge:
And I  guess map would probably written more like this:
```
map :: forall {p :: Multiplicity} a b. (a %p -> b) -> [a] %p -> [b]
```
(With an inferred argument)

goldfirere:
Prodded by @monoidal, I added support for modifiers in more places, to keep up with the linear types proposal. I've also added the `map` example above.

simonpj:
In the first bullet of "Costs" you imply that you have to say `Int %(m :: Multiplicity) -> Bool`.  But you don't, as the last example in Examples shows.   Confusing!

`data Tagged (%nominal t) a = Tagged a`.  I didn't understand this one.  Where is `nominal` (presumably a type variable) bound? A and how is this decoration reflected in the type it decorates?

There may a tricky bit implementation-wise.  We don't know whether the modifier is a multplicity or a matchability thing until after kind inference has completed.  Another reason for kind-checking to return a HsType not a Type.

Presumably the kind of the modifier must be one of a fixed collection.  No modifier polymoprhism!  Say this.

Modifiers are always types, right, even in terms?  `prefix% atype`.

goldfirere:
I've added a few clarifications based on @simonpj's comment.

> Modifiers are always types, right, even in terms?

I interpret this question as: "what parser and namespace are used in modifiers?" For now, yes, they're types. But there's no reason this has to remain true, as we continue to find uses for this new syntax.

aspiwack:
I don't understand why the `f5` example would parse (even to be rejected). I don't think it follows from your specification, which I read as allowing a single `%mod` to the left of an arrow.

int-index:
@aspiwack In EBNF, `{ ... }` means repetition, i.e. "zero or more". So `{modifier}` accepts a sequence of modifiers.

aspiwack:
Ah, indeed. Thanks.

kcsongor:
I'm strongly in favour if this proposal! (somehow missed it until now, thanks for writing it up)
Concerning #242, there is also a need to annotate `forall`s, e.g.:
```haskell
type Foo :: forall k. @U k
type family Foo :: k where ...
```

could/should this be handled by rule 3 in your proposal?

```haskell
type Foo :: forall (%U k). k
```

The current version of #242 uses a single annotation for the whole telescope of forall-bound variables, but more fine-grained control might be better.

goldfirere:
This has been quiet. I'd like to submit to the committee @nomeata . I would normally say that it should be considered alongside #242, but I'm shepherding that, and so it seems improper to also shepherd this one.

serras:
I really think this is the way to go, since more and more proposals are going to add syntax. Let me ask a few questions about the proposal:
- I feel like @simonpj's question about inference have not been fully answered. The proposal says that a constraint is emitted, but having a variable lying around may negatively affect how constraint solving proceeds.
- Could we introduce a generic way to say which is the "default" of each modifier. I mean, when we parse `a -> b`, somehow we have to know that a lone arrow means matchability X, linearity Y, and so on.
- Do we need to decide on syntax for the term language, or would it be better to focus simply on the type level?

goldfirere:
> 
>     * I feel like @simonpj's question about inference have not been fully answered. The proposal says that a constraint is emitted, but having a variable lying around may negatively affect how constraint solving proceeds.

Yes. I'm unsure about this. I can't seem to quite figure out how it would all play out.

> 
>     * Could we introduce a generic way to say which is the "default" of each modifier. I mean, when we parse `a -> b`, somehow we have to know that a lone arrow means matchability X, linearity Y, and so on.

I don't see this as a property of the modifier, but rather of the `->`. I could imagine writing e.g. `default Matchability M` or `default Multiplicity One` or some such, but I think that's separate from this proposal.

> 
>     * Do we need to decide on syntax for the term language, or would it be better to focus simply on the type level?

I would like to universally reserve prefix-`%` for modifiers, meaning that no other proposal should take this space, unless that proposal is compatible with a modifier interpretation. Beyond reserving term-level syntax, I don't need this proposal to affect the term level.

serras:
I am recommending acceptance on this proposal. I'll leave up to the Committee to decide whether the prefix should also be reserved at the term-level.

proposals/0000-modifiers.rst
```diff
@@ -0,0 +1,233 @@
...
+8. Introduce a new type-level constant ``Modifier :: Type -> Constraint``, exported
+   from ``GHC.Exts``.
+
+9. Let the constraint ``Modifier Multiplicity`` be satisfiable; let no other
+   ``Modifier`` constraint be satisfiable.
+
+10. During constraint generation, let an occurrence ``%(ty)``, where ``ty :: ki``,
+    emit a constraint ``Modifier ki``.
+
+11. A modifier of type ``Multiplicity`` changes the multiplicity of the following arrow,
```
gridaphobe, line 94: The idea of constraint resolution changing the type of a neighboring syntactic element feels really strange to me. And it's very unclear how we would extend this to user-defined modifiers without a user-definable semantics for "running" a modifier, which I think is a must for a feature like this (see other existing mechanisms like `ANN` pragmas, the various plugins we support, even template haskell and quasiquoters). I think I'd be happier with this proposal if it specified a user-definable mechanism and then showed how we could use it to solve the multiplicity/matchability problem.

proposals/0000-modifiers.rst
```diff
@@ -0,0 +1,233 @@
...
+
+     fexp     ::= {modifier} aexp | fexp aexp
+
+5. With ``-XModifiers``, introduce postfix modifier syntax on patterns as follows::
+
+     lpat     ::= apat {modifier}
+
+6. With ``-XModifiers``, introduce postfix modifier syntax on record field declarations as follows::
+     
+     fielddecl ::= vars '::' (type | '!' atype) {modifier}
```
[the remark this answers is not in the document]
goldfirere: This is meant to be compatible with the current syntax for linear types. I'm open to other placements.

proposals/0000-modifiers.rst
```diff
@@ -0,0 +1,233 @@
...
+  like ``type ManyMatch = '(Many, Matchable); foo :: Int %ManyMatch -> Bool``.
+
+* Though not proposed here, we can imagine a large extension to this
+  mechanism allowing for *user-written* ``Modifier``\s. Perhaps a
+  ``Modifier`` type supports some function call to the GHC API that
+  transforms the meaning of bit of syntax. The possibilities are
+  tantalizing.
+  
+* These modifiers recall Java's `Annotations <https://en.wikipedia.org/wiki/Java_annotation>`_
+  mechanism, which were a direct inspiration.
```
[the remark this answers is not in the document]
goldfirere: I think it depends on what we want the semantics of a pragma to be. I like the idea that pragmas should be out-of-band instructions to the compiler. Inlining, rewrite rules, warning enable/disable, and deprecation (among others) fit nicely here. On the other hand, annotations that change the typing or execution (beyond performance) of the program are really not out-of-band: they are an essential part of the program. These, in my opinion, should not be pragmas -- instead, these should be modifiers. The `ANN` mechanism you mention here thus seems similar to, but not the same as, the idea of modifiers.

proposals/0000-modifiers.rst
goldfirere: Yes, I've grown uncomfortable about this, too. Modification of the proposal to appear shortly.

serras:
@goldfirere given that changes are expected here, I'll move this proposal back to the revision state. Looking forward to see those changes, I think that this proposal -- in whatever shape ends up being -- is really needed for the upcoming changes in the language.

goldfirere:
I've uploaded my changes, removing `Modifier` and simplifying the proposal somewhat. Really, this is all about reserving syntactic space, and I agree that the overloading mechanism I tried to build was a step too far.

Given the technical nature of this one and the general lack of commentary, I'd like to send this pack to the committee, @serras.

proposals/0000-modifiers.rst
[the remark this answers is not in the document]
gridaphobe: I think changing program semantics or not may be a useful distinction to make, but this proposal already blurs the lines by suggesting `inline` and `oneShot` as candidates for future modifiers. Neither changes the semantics of your program, only how GHC optimizes it. And it's also not super clear where `ANN` pragmas fall. They don't change the semantics on their own, but seem designed to aid plugins that may well use them to change your program's semantics!

Here's another example, I've long wanted to give LiquidHaskell refinements a more first-class feeling by teaching GHC about them to some extent. We considered using `ANN` pragmas at one point but it didn't quite work for some technical reason (I think it was surmountable, we just didn't follow through..). But would they fit better as modifiers? From GHC's perspective refinements don't change the semantics of your program (at least not at the moment), but they are an essential piece from the programmer's perspective.

proposals/0000-modifiers.rst
```diff
@@ -0,0 +1,214 @@
...
+
+1. Introduce a new extension ``-XModifiers``.
+
+2. With ``-XModifiers``, introduce modifier syntax on arrows as follows (cf.
+   the Haskell 2010 Report for all BNF syntax, and recall that we use braces
+   to denote "zero or more")::
+
+     type     ::= btype [ {modifier} -> type ]
+     prefix%  ::= '%'    -- only in prefix position
+     modifier ::= prefix% atype
```
gridaphobe, line 61: It still feels strange to say that a modifier is a type. It sounds more like a function `Type -> Type` (or `Expr -> Expr` or `Pat -> Pat` in the other cases), but now that sounds a lot like a Template Haskell splice. Should modifiers maybe be a nicer syntax for running TH splices?

proposals/0000-modifiers.rst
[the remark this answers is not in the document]
gridaphobe: One possibility would be to follow Haddock's precedent and have one syntax for modifying the next syntactic element, and another for modifying the previous element, but then we'd have to steal more syntax.

Also, I should note that the second sentence of the proposal says:

> Modifiers somehow change the meaning of the **next** token.

proposals/0000-modifiers.rst
```diff
@@ -0,0 +1,214 @@
...
+    and its typing rules.
+  * a mechanism for suppressing warnings over one region of a program,
+    instead of at the module level: ``%(suppress "uni-complete-patterns") (case x of ...)``.
+    This could also be done with a pragma.
+
+* Some other features that have had tortuous and torturous syntax
+  discussions may have enjoyed having the modifier option. For example,
+  this might have been used instead of ``type role`` for role annotations:
+  ``data Tagged (%Nominal t) a = Tagged a``. Or it might have been an
+  alternative for ``-XDerivingStrategies``.
```
gridaphobe, line 171: Another possibility that just came to mind, should we have levity modifiers?

proposals/0000-modifiers.rst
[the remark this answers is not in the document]
goldfirere: These are all really good points -- the line is blurrier than I realized. And you're right about `oneShot` and `inline`: these really should be pragmas, not modifiers. It's just that they are now magic functions, which seems worse than either.

For what it's worth, I would say that an LH annotation would be a modifier, because it changes the compile-time semantics of the program (taking a program that might have otherwise been accepted to be rejected).

proposals/0000-modifiers.rst
goldfirere: That's an interesting way to phrase them, but no. If we imagine that TH supports a larger syntax than Haskell, we could also imagine implementing lots of features via TH that have no syntax in "pure" Haskell. But I would rather keep TH and Haskell aligned with each other -- and to avoid needing the TH machinery to, say, specify a linear type.

And a modifier is not a type. It's parsed/renamed as one, but then interpreted separately.

proposals/0000-modifiers.rst
goldfirere: We could have levity modifiers, but as levity is nicely captured by the kind system, there seems to be only weak motivation to introduce a new mechanism for them.

proposals/0000-modifiers.rst
gridaphobe: > But I would rather keep TH and Haskell aligned with each other -- and to avoid needing the TH machinery to, say, specify a linear type.

Agreed, but I assumed there would be a modifier-free way of specifying things like linear types (possibly horribly ugly) and that modifiers would act more as a shorthand for the full incantation, which feels more like TH.

proposals/0000-modifiers.rst
gridaphobe: Another feature that in retrospect seems like a modifier: strictness annotations.

aspiwack:
There seem to be a similar flavour from this new variant of the proposal than the _fully settled_ idea previously discussed at https://github.com/ghc-proposals/ghc-proposals/pull/216#issuecomment-598744392 . Which makes sense to me.

nomeata:
> There seems to be a similar flavour from this new variant of the proposal than the _fully settled_ idea previously discussed at [#216 (comment)](https://github.com/ghc-proposals/ghc-proposals/pull/216#issuecomment-598744392) . Which makes sense to me.

So if we had had these ~annotations~ modifiers earlier, I wonder if that might have swayed us to the builder-design for local do, using the  `%builder do` synax? But probably otiose to muse about that.

simonpj:
> This proposal means that Int %m -> Bool, acceptable today as a multiplicity-polymorphic function, would be rejected.

Would it? What about
```
f :: forall (m :: Multiplicity). Int %m -> Bool
```
surely that's clear enough? Why can't we accept it?

> Future modifiers will also seamlessly work with existing ones, where order is not expected to matter (though that would be up to other proposals to spell out).

Can we specify, up front, that order doesn't matter?

**Syntax**: the form `btype [ {modifier} -> type ]` seems to allow one or zero modifiers but not two.  Or do I misunderstand the braces?  Spell it out please

simonpj:
Generally I'm in favour, but I'd mark this as experimental syntax.  We might end up wanting to change it.  But the only we can find out is to try it.

kcsongor:
I see that there's no implementation plan outlined in the proposal. May I volunteer to implement it? Given that unsaturated type families will be the primary beneficiary of this feature in the short term, I think it would make sense to bundle the implementation effort.

goldfirere:
Yes -- I think implementing this within the implementation for #242 would be ideal. Thanks!

goldfirere:
@simonpj 

> Syntax: the form `btype [ {modifier} -> type ]` seems to allow one or zero modifiers but not two. Or do I misunderstand the braces? Spell it out please

Directly above this production, the proposal says

> recall that we use braces to denote "zero or more"

Does that specify the grammar precisely?

simonpj:
>     recall that we use braces to denote "zero or more"

Well, I missed that.  Sorry.  Maybe put the parenthesis *after* the BNF block rather than before?

goldfirere:
I'm inclined to leave this part alone -- I think introducing notation before using it makes more sense than the other way around.

Is there anything else blocking a merge?

googleson78:
I'm interested in implementing this but the best I can do currently is a "a couple of hours on some weekends" sort of commitment. I'd also appreciate someone to "mentor" me, mostly on design points. 

@kcsongor are you still working/planning to work on this, or can I potentially pick it up at some point?

goldfirere:
Happy to help work out a design. And the time you're offering here is a lot more than we've got, so it's very welcome. Thank you! If you want to go ahead, make a GHC ticket, tag me, and then it's also probably best to email, just to make sure it gets my attention.

Icelandjack:
This could be used to give field or constructor specific deriving behaviour, as a continuation of a [prior idea](https://www.reddit.com/r/haskell/comments/pbq9rn/via_fields_finer_granularity_in_deriving/). Specifying `%(Via Sum Int) Int` hints that we want derive using the implementation for `Sum Int`.

```haskell
data XY = MkXY
  { x :: %(Via Sum Int) Int
  , y :: %(Via Sum Int) Int }
  deriving stock Generic
  deriving (Semigroup, Monoid) via Generically XY
```

The `!`-field modifier could then be viewed as shorthand for this sort of `%Bang` modifier.

googleson78:
I've also been constantly thinking about what @Icelandjack said - a lot of other languages have some pragma that says things like "omit this field from the json rep" or "this field is going to be the db key" and modifiers have the potential to fill in this exact role verry well if we expose modifier info to `Generic` and TH.

adamgundry:
I've marked this and its amendments as implemented, since https://gitlab.haskell.org/ghc/ghc/-/merge_requests/14781 has landed. Thanks @ChickenProp!

PR #371, opened 2020-10-16
Export ~ from Data.Type.Equality

int-index:
The [proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0371-non-magical-eq.md) has been accepted; the following discussion is mostly of historic interest.

-------------------------------

The `~` type operator is magical, in that it's always in scope and does not require `TypeOperators` to use.
But `~~`, `:~:`, and `:~~:` are not magical in the same way, they are
exported from `Data.Type.Equality` and require `TypeOperators`.

This is a historical accident which we propose to correct.

[Rendered](https://github.com/int-index/ghc-proposals/blob/non-magical-eq/proposals/0000-non-magical-eq.md)

### DIFF
--- a/proposals/0000-non-magical-eq.md
+++ b/proposals/0000-non-magical-eq.md
@@ -1,0 +1,76 @@
+---
+author: Vladislav Zavialov
+date-accepted: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/371).
+
+# Export `~` from `Data.Type.Equality`
+
+
+## Motivation
+
+The `~` type operator is magical, in that it's always in scope and does not require `TypeOperators` to use.
+But `~~`, `:~:`, and `:~~:` are not magical in the same way, they are
+exported from `Data.Type.Equality` and require `TypeOperators`.
+
+This is a historical accident which we propose to correct.
+
+## Proposed Change Specification
+
+* Do not bring `~` into scope magically, export it from `Data.Type.Equality`.
+
+* Re-export `~` from `Prelude`.
+
+* Drop the requieremnt to enable `TypeFamilies` or `GADTs`
+  to use the `~` type operator.
+
+* Require `TypeOperators` to use the `~` type operator.
+
+### Migration story
+
+Let `n` be the GHC version that implements this proposal.
+
+Starting with GHC `n`, include a compatibility fallback:
+
+* When the lookup for `~` fails because it is not in scope,
+  assume it refers to `Data.Type.Equality.~`.
+  When this happens and `-Wcompat` is in effect, emit a warning.
+
+Starting with GHC `n+2`, enable the warning by default.
+Either starting with GHC `n+8` or with the next major compiler version bump (GHC
+10), whichever comes first, remove the compatibility fallback.
+
+## Effect and Interactions
+
+* One less quirk.
+
+* The users are allowed to define their own ``~``.
+
+* The compiler internals are simplified, in particular things described in
+  ``Note [eqTyCon (~) is built-in syntax]``.
+
+* Haddock documentation for ``~`` will get fixed (it is currently missing in
+  the generated HTML).
+
+## Costs and Drawbacks
+
+This is a breaking change:
+
+* Users will need to enable `TypeOperators`
+* Authors of custom preludes will likely want to re-export `(~)`
+
+## Alternatives
+
+* Keep `~` special for historical reasons to avoid a breaking change.
+* A different migration schedule (e.g. include the warning in `-Wall` first).
+
+## Unresolved Questions
+
+None.
+
+## Implementation Plan
+
+There's a prototype implementation in [MR 4313](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4313).


treeowl:
I like *most* of this. But doesn't `TypeOperators` sometimes require other code to change? Breaking vast numbers of packages by requiring it for `~` strikes me as too high a price for very little benefit.

int-index:
> But doesn't `TypeOperators` sometimes require other code to change?

I can't think of an example. Looking at the implementation, it seems that we always parse them, but then require the user to enable `TypeOperators` in a post-hoc check:

```
  = do { ops_ok <- xoptM LangExt.TypeOperators
       ; op' <- rnTyVar env op
       ; unless (ops_ok || op' `hasKey` eqTyConKey) $
           addErr (opTyErr op overall_ty)
```

With this proposal, we remove the `op' `hasKey` eqTyConKey` special case from the check (and from a few other places).

Ericson2314:
Maybe @treeowl was thinking of `-XNoStarIsType` being implied by `-XTypeOperators`?

int-index:
Ah, yes, but that was removed very quickly when we found out about the extent of breakage: https://github.com/ghc-proposals/ghc-proposals/pull/146/commits/eab388b1b57a26ff9704e87ba3add8b561833be4

AntC2:
I'm not convinced `~` is an operator at all.

* In patterns it's not infix, it's prefix (lazy match).
* In constraints, it can only appear at top level.
* What would it mean to have `(a ~ b)` as an argument to some other type operator or class?
* In terms `~` is not allowed at all -- or did I miss that memo?

If we allow `~` as a type operator, does that mean people can define their own `~` and override it being a type equality constraint? That seems to me a good way to puzzlement for anybody else. IIUC, `~`'s behaviour as a type equality constraint is wired-in/you couldn't define some other operator or class to get the same behaviour.

The Motivation section here doesn't include anything that motivates me. It's a bureaucrat nit-picking.

phadej:
> In terms ~ is not allowed at all -- or did I miss that memo?

You did. https://github.com/ghc-proposals/ghc-proposals/pull/229 

> As a consequence of these rules, `@` (loose infix) and `~` (suffix, loose infix, tight infix) are now proper infix operators.

goldfirere:
@AntC2 
 
* In patterns it's not infix, it's prefix (lazy match).

My understanding is that this is about the operator `~`, completely different from, and related only by superficial similarity to, the `~` used to denote a lazy pattern match.

* In constraints, it can only appear at top level.

I don't think this is accurate. Or maybe I misunderstand your meaning?

* What would it mean to have `(a ~ b)` as an argument to some other type operator or class?

The same thing passing any constraint would mean. For example, we can have `Dict (a ~ Int)` if we like. (This is true today.)

Internally, `(~)` is actually just a fairly ordinary class, with a non-ordinary superclass (GHC's real, internal equality) and non-ordinary "instance". But `(~) :: forall k. k -> k -> Constraint` is true today.

@int-index 

This means a breaking change for anyone who says `import Prelude ()` or `-XNoImplicitPrelude` and uses `~` (beyond the more obvious breaking change around `-XTypeOperators`). And the workaround won't be backward compatible. I'm not convinced the motivation is worth this -- seems like intentional busywork. Yes, the language you're moving us toward is more uniform than the one we have today, but not every step in that direction is inherently worthwhile. I'm thus on the fence, but leaning against.

Ericson2314:
I hope if this is too much breakage we could still do a `-XNoImplciitEqualityProp` or something.

int-index:
> And the workaround won't be backward compatible.

@goldfirere How would you feel about the change if it had a migration story?

1. allow exporting, importing, and defining `(~)` as any other type operator
2. when `(~)` is not in scope, its occurrence is interpreted as `eqTyCon`, with a warning
3. in some amount of releases, drop (2) to simplify compiler internals

goldfirere:
Yes, I'm happy with that migration story. The warning would have to be born in `-Wcompat` (only) and then eventually migrate to on-by-default. But I like.

treeowl:
I'm happy with `-XNoImplcitEqualityProp`, but not terribly happy with *yet another* language extension to have to type in just to do basic Haskell programming. I very often *don't* need any other type operators. And there's sure to be lots of annoying breakage ... to what end, really?

phadej:
@treeowl to not get stuck in inconsistent place where `@` can be operator, but `~` is hardcoded syntax.

In retrospective using `~` for type-equality constraint was a bad choice without *first* doing what @int-index is cleaning up afterwards. I really thank @int-index for making possible for `~` to have a fixity which one would expect from equivalence relation. This is very natural follow up to me.

I'm :+1: to the goal, even there will be awkward intermediate phase.

The proposal could (and IMHO) should reflect on `-XStarIsType`. This is very similar in spirit. At some point, you cannot have `*` in kinds, you have to import and use `Type` (I don't remember details, will `Type` be brought to `Prelude`). I'd value the `~` having similar treatment.

Ericson2314:
@treeowl Well, my thought is to queue up these `-XNo` ones for the next language standard :). Also, it's not clear (to me at least) whether @int-index's migration plan in fact uses a new extension?

adamgundry:
I've always felt it odd that writing an equality constraint requires at least one of `GADTs` or `TypeFamilies` to be enabled (https://gitlab.haskell.org/ghc/ghc/-/issues/10431). Generally speaking GHC doesn't require any language extensions to be enabled at use sites that merely mention "special" imported constraints (e.g. `Typeable`, `Coercible`). If `~` is just a (type) operator, why should it be different from `Coercible`?

Could we instead say that writing `~` doesn't require any extensions, but does require an explicit `import Data.Type.Equality ((~))`, i.e. `~` would not be re-exported by the `Prelude`? (With an appropriate migration story so that users are gently prodded to add the import if it will become necessary in the future.)

int-index:
I updated the proposal to reflect the discussion:

* Do not require `GADTs` and `TypeFamilies` for `~` because `Typeable`, `Coercible`, and other ‘special’ constraints don’t require any extensions either

* Do not re-export `~` from Prelude, the users will import it from `Data.Type.Equality` explicitly

* Include a migration story (see https://github.com/ghc-proposals/ghc-proposals/pull/371#issuecomment-711127708)

Did I miss anything?

goldfirere:
I'm still not happy with the migration story -- sorry. I think the warning should move from `-Wcompat` to on-by-default for several releases before we drop the current support.

My bottom line here: there's no fire we're trying to put out. Instead, we're just trying to do a bit of housekeeping. But, since this is really for GHC's and language-lawyers' benefit, we should be as gentle to users as possible, and thus have a long, slow migration story.

I could be easily convinced not to do this at all, but with a gentle enough migration story, I suppose I'm in favor.

int-index:
> there's no fire we're trying to put out

I agree. The fallback is not a huge burden on the implementation either. We can wait as long as necessary, I just think it’d be nice to clean it up eventually.

goldfirere:
I'm happy with the new migration story. Thanks. When the time comes, we might decide to kick the can down the road further, but we can wait for the light of experience.

int-index:
@nomeata, I’d like to submit this proposal.

goldfirere:

/remind me in 2 weeks to accept unless there has been dissent.

nomeata:
The proposal originally suggested to export `(~)` from `Prelude`, which, together with the fact that GHC2021 includes `TypeOperators`, would mean that many users don’t have to do anything. What is the argument against going that route?

int-index:
> What is the argument against going that route?

It's a special, magical constraint, like `Coercible` and `Typeable`. It's good to have an explicit indication that it is used in the code.

It's a valid reason, but not a strong one. I am ambivalent about this.

nomeata:
I see. It’s a trade off between “if you happen to be using GHC2021 or TypeOperators anyways, it’ll be a non-breaking change” and “ideally people loudly opt-into using `~`”.  I am also ambivalent on this one.

phadej:
People need to opt-in for `~` at the moment too.

```
<interactive>:1:12: error:
    • Illegal equational constraint a ~ b
      (Use GADTs or TypeFamilies to permit this)
```

GHC2021 doesn't include either of extensions. So even with requiring `import Data.Type.Equality` this proposal will require "less".

I have written code where I enabled `GADTs` just to allow me to write `a ~ b`, e.g. something like:

```haskell
instance a ~ Char => IsString [a] where
    fromString = id
```

nomeata:
I was referring to the state after this proposal (needing `TypeOperators`, but not `GADTS` or `TypeFamilies`). I fully agree that I found it odd that I was told to enable `GADTs` to get `~`.

My musing about exporting `~` from the Prelude is not about whether people have to opt in less or more, but merely if we can make this less disruptive this way (even if we woudn’t export it from the prelude if this was introduced as a new feature now)

goldfirere:
I've thought about exporting `~` from the Prelude in the past, but it was an attempt to avoid any backward-incompatibility. It fails, because some users might have `-XNoImplicitPrelude` or `import Prelude ( ... )`. But it certainly lessens the backward-incompatibility problem (while not eliminating it). Perhaps that would be an improvement here: export `~` from `Data.Type.Equality` and re-export it from `Prelude`.

goldfirere:
I'm having a hard time reading the temperature of the room around exporting `(~)` from `Prelude`? I personally think that would strengthen this proposal, in that it would make back-compat less troublesome for more users and seems to have no downside (other than the fact that it doesn't completely eliminate the back-compat problem). Nothing else in the proposal would change.

What do we think? Thumbs-up this post if you think exporting `(~)` from `Prelude` would make this better; thumbs-down this post if you think it would make the proposal worse.

nomeata:
I am not hot about it (hence not temporature to read), but it seems that exporting `(~)` from the Prelude seems to be the choice that’s considerate of our users time, as it breaks less code (it seems), at a reasonably small cost

simonpj:
> I'm having a hard time reading the temperature of the room around exporting (~) from Prelude? 

This is exclusively a user-experience issue, not a technical one, and I feel ill-qualified to have an opinion. So I'm explicitly neutral on this one.

goldfirere:
Hmm. I don't know about dissent, per se, but it seems there is energy toward exporting `~` from the Prelude as part of this proposal. After that's done, I'll send this back through the committee, hopefully for a definitive up or down. Thanks!

int-index:
@goldfirere Ok, let’s resubmit.

goldfirere:
I've re-recommended acceptance.

/remind me in 1 week to accept

monoidal:
> Drop the requieremnt to enable TypeFamilies or GADTs to use the ~ type operator.

Question: There are two places where GHC requires either TypeFamilies or GADTs:

(a) direct usage `type T = (~)`

(b) pattern match on a GADT `case x of T a -> ...`

Is the proposal affecting (a) only or (a)+(b)?

int-index:
The proposal is about (a) only, as (b) is not even an occurrence of the `~` operator.

proposals/0000-non-magical-eq.md
```diff
@@ -0,0 +1,76 @@
...
+
+* Haddock documentation for ``~`` will get fixed (it is currently missing in
+  the generated HTML).
+
+## Costs and Drawbacks
+
+This is a breaking change:
+
+* Users will need to enable `TypeOperators`
+* Authors of custom preludes will likely want to re-export `(~)`
```
nh2, line 63: How does one do that re-export of `~` in custom preludes with the common `as` don't-repeat-yourself pattern?

When I use

```hs
module Myprelude (
  module AllReexports
) where

import Data.Type.Equality as AllReexports (type (~))
-- other imports `as AllReexports` here
```

in GHC 9.10 I get warning

```
The import of ‘Data.Type.Equality’ is redundant
  except perhaps to import instances from ‘Data.Type.Equality’
To import instances alone, use: import Data.Type.Equality()
```

Was this in scope of the proposal / implementation?

## Workaround

I spotted this in [`relude`](https://hackage-content.haskell.org/package/relude-1.2.2.2/docs/src/Relude.Base.html):

```hs
module Myprelude (
  module AllReexports,
  module Data.Type.Equality
) where

import Data.Type.Equality (type (~))
```

int-index: I've tried to load this module in GHCi but haven't got a warning

```haskell
{-# LANGUAGE ExplicitNamespaces, TypeOperators #-}
{-# OPTIONS -Wall #-}

module Myprelude (
  module AllReexports
) where

import Data.Type.Equality as AllReexports (type (~))
```

But in your example, there's

```
-- other imports `as AllReexports` here
```

Could be that one of those "other imports" also reexports `~`?

PR #372, opened 2020-10-22
The GHC20xx process  (under review)

nomeata:
This is based on an early draft by @serras on https://github.com/serras/ghc-proposals/blob/ghc-2020/proposals/0000-ghc-2020.md

[Rendered](https://github.com/ghc-proposals/ghc-proposals/blob/ghc20xx-process/proposals/0000-ghc-extensions.rst)

### DIFF
--- a/proposals/0000-ghc-extensions.rst
+++ b/proposals/0000-ghc-extensions.rst
@@ -68,7 +68,7 @@ Process
 * In order to gather data on the criterium “widespread usage”, the secretary creates a tally of which extensions are used how often on hackage.
 
 * In order to gather data on the criterium “community support”, the secretary runs a public poll on a suitable platform for one week where anyone
-  can vote in favor or against the inclusion of a given extension.  
+  can vote in favor or against the inclusion of a given extension, or points the committee to a suitable existing survey result.
 
 * Within two weeks of the start of the process, every committee member is expected to send an initial list of which extensions they expect to be in GHC202x to the mailing list.
    
--- a/proposals/0000-ghc-extensions.rst
+++ b/proposals/0000-ghc-extensions.rst
@@ -7,9 +7,7 @@ The GHC20XX process
 .. ticket-url::
 .. implemented::
 .. highlight:: haskell
-.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
-            **After creating the pull request, edit this file again, update the
-            number in the link, and delete this bold sentence.**
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/372>`_.
 .. sectnum::
 .. contents::
 
--- a/proposals/0000-ghc-extensions.rst
+++ b/proposals/0000-ghc-extensions.rst
@@ -24,7 +24,7 @@ It has almost become a meme that any slightly-complex Haskell module starts with
 
 For that reason, it's benefitial to pack those extensions which have "graduated" in terms of stability and usefulness into a single language version.
 
-We aim to lean on the conservative side in our selection of extensions: Not including an extension that we should have is relatively harmless (users of that extension will have to still set that manually, but they alreay have to do that, and we can add it alter), while including an extension that we should not have hurts more (any detriminal effect on developer experience now affects many users).
+We aim to lean on the conservative side in our selection of extensions: Not including an extension that we should have is relatively harmless (users of that extension will have to still set that manually, but they already have to do that, and we can add it alter). Conversely, including an extension that we should not have hurts more (any detriminal effect on developer experience now affects many users).
 
 We also aim to have a process that is efficient. For each extension, we are all very capable of having very long discussions about the merits, up-sides and downsides, and almost anybody will have an opinion on almost any extensions. But aiming to include only _uncontroversial_ extensions, we aim to avoid much of these discussion by construction. The process deliberately does not provide room for long per-extensions deliberations, neither within the committee or with the overall community, while still involving the community.
 
--- a/proposals/0000-ghc-extensions.rst
+++ b/proposals/0000-ghc-extensions.rst
@@ -47,7 +47,7 @@ affects many users).
 We also aim to have a process that is efficient. For each extension, we are
 all very capable of having very long discussions about the merits, up-sides
 and downsides, and almost anybody will have an opinion on almost any
-extensions. But aiming to include only _uncontroversial_ extensions, we aim to
+extensions. But aiming to include only *uncontroversial* extensions, we aim to
 avoid much of these discussion by construction. The process deliberately does
 not provide room for long per-extensions deliberations, neither within the
 committee or with the overall community, while still involving the community.
--- a/proposals/0000-ghc-extensions.rst
+++ b/proposals/0000-ghc-extensions.rst
@@ -11,7 +11,7 @@ The GHC20XX process
 .. sectnum::
 .. contents::
 
-This proposal sets up a process to introduce new language versions `GHC20XX`, extending `Haskell2010` as defined by the Report with several extensions that are considerd harmless, helpfun and uncontroversial.
+This proposal sets up a process to introduce new language versions `GHC20XX`, extending `Haskell2010` as defined by the Report with several extensions that are considerd harmless, helpful and uncontroversial.
 
 
 Motivation
--- a/proposals/0000-ghc-extensions.rst
+++ b/proposals/0000-ghc-extensions.rst
@@ -38,7 +38,7 @@ version.
 We aim to lean on the conservative side in our selection of extensions: Not
 including an extension that we should have is relatively harmless (users of
 that extension will have to still set that manually, but they already have to
-do that, and we can add it alter). Conversely, including an extension that we
+do that, and we can add it later). Conversely, including an extension that we
 should not have hurts more (any detriminal effect on developer experience now
 affects many users).
 
@@ -104,18 +104,18 @@ criteria, but often subjective, and not all need to be satisfied in all cases.
 Process
 ^^^^^^^
 
-* 4 months before the expected GHC spring release day of 202x, the committee
-  Secretary starts the GHC202x process.
+* 4 months before the expected GHC spring release day of 20xx, the committee
+  Secretary starts the GHC20xx process.
 
   They inform the committee, in an email to the mailing list, of all language
-  extensions supported by the latest released GHC that are not in GHC202(x-1),
-  which could be added. They also list all extensions *in* GHC202(x-1), which
-  might be omitted in GHC202(x-1) (likely a rare thing).
+  extensions supported by the latest released GHC that are not in GHC20(xx-1),
+  which could be added. They also list all extensions *in* GHC20(xx-1), which
+  might be omitted in GHC202(xx-1) (likely a rare thing).
 
   The secretary also creates a PR with a proposal saying (roughly)
    
-    GHC202x contains the following extensions in addition to those in
-    GHC202(x-1):
+    GHC20xx contains the following extensions in addition to those in
+    GHC20(xx-1):
     
     * (none yet)
     
@@ -124,7 +124,7 @@ Process
     * (none yet)
 
 * In order to gather data on the criterium “widespread usage”, the secretary
-  creates a tally of which extensions are used how often on hackage.
+  creates a tally of which extensions are used how often on Hackage.
 
 * In order to gather data on the criterium “community support”, the secretary
   runs a public poll on a suitable platform for one week where anyone can vote
@@ -133,7 +133,7 @@ Process
 
 * Within two weeks of the start of the process, every committee member is
   expected to send an initial list of which extensions they expect to be in
-  GHC202x to the mailing list.
+  GHC20xx to the mailing list.
    
   Committee members are expected to take the Hackage statistics and the
   community vote into account.
@@ -156,7 +156,7 @@ Process
   simply not be included.
 
 * After these four weeks, the proposal with the current tally gets accepted by
-  the secretary, and defines GHC202x
+  the secretary, and defines GHC20xx
 
 Cadence
 ^^^^^^^
@@ -167,7 +167,7 @@ uncontroversial extensions; so the next iteration will likely only make a
 smaller, but more controversial change.
 
 Therefore, this proposal does *not* commit to a fixed cadence. Instead, 6
-months after the first release of a version of GHC that supports a GHC20XX
+months after the first release of a version of GHC that supports a GHC20xx
 set, we evaluate the outcome, the process, and the perceived need of a next
 release. At that time we will refine the processes, if needed, and set a
 cadence.
--- a/proposals/0000-ghc-extensions.rst
+++ b/proposals/0000-ghc-extensions.rst
@@ -74,7 +74,7 @@ Process
    
   Committee members are expected to take the Hackage statistics and the community vote into account.
    
-  These mails may contain justifications for why a certain extension is or is not included, but not required (or even expected).
+  These mails may contain justifications for why a certain extension is or is not included, but this is not required (or even expected).
 
   After these two weeks, the PR is continuously updated by the secretary to reflect the *current* tally of votes: An extension is included if it is listed by at least ⅔ (rounded up) of committee members.
 
--- a/proposals/0000-ghc-extensions.rst
+++ b/proposals/0000-ghc-extensions.rst
@@ -32,7 +32,7 @@ We also aim to have a process that is efficient. For each extension, we are all 
 Proposed Change Specification
 -----------------------------
 
-We propose to introduce a process to develop new "language versions", dubbed ``GHC20XX``, which extend ``Haskell2010`` with some default extensions. This proposal does not set *which* extensions are included in the first version,, rather seeks to provide a framework to reach a conclusion in a constructive and timely manner.
+We propose to introduce a process to develop new "language versions", dubbed ``GHC20XX``, which extend ``Haskell2010`` with some default extensions. This proposal does not set *which* extensions are included in the first version, but rather seeks to provide a framework to reach a conclusion in a constructive and timely manner.
 
 Criteria
 ^^^^^^^^
--- a/proposals/0000-ghc-extensions.rst
+++ b/proposals/0000-ghc-extensions.rst
@@ -53,7 +53,7 @@ Process
 
 * 4 months before the expected GHC spring release day of 202x, the committee Secretary starts the GHC202x process.
 
-  They inform the committe, in an email to the mailing list, of all language extensions supported by the latest released GHC that are not in GHC202(x-1), which could be added. They also list all extensions *in* GHC202(x-1), which could be removed.
+  They inform the committee, in an email to the mailing list, of all language extensions supported by the latest released GHC that are not in GHC202(x-1), which could be added. They also list all extensions *in* GHC202(x-1), which might be omitted in GHC202(x-1) (likely a rare thing).
 
   The secretary also creates a PR with a proposal saying (roughly) 
    
--- a/proposals/0000-ghc-extensions.rst
+++ b/proposals/0000-ghc-extensions.rst
@@ -72,7 +72,7 @@ Process
 
 * Within two weeks of the start of the process, every committee member is expected to send an initial list of which extensions they expect to be in GHC202x to the mailing list.
    
-  Committe members are expected to take the Hackage statistics and the community vote into account.
+  Committee members are expected to take the Hackage statistics and the community vote into account.
    
   These mails may contain justifications for why a certain extension is or is not included, but not required (or even expected).
 
--- a/proposals/0000-ghc-extensions.rst
+++ b/proposals/0000-ghc-extensions.rst
@@ -100,7 +100,7 @@ The implementation cost seems small.
 
 The cost of a GHC20xx extension is that, upon reading a file with ``{-# LANGUAGE GHC20xx #-}``, the reader does not immediatelly know the set of enabled extensions; this may hamper readability of code.
 
-The costs of this process is that it binds volunteer time, and there is is risk of unpleasant, heated debates, because everybody has opinions. The process tries to err on the conversative side and rather add too few than too many extensions.
+The costs of this process is that it binds volunteer time, and there is is risk of unpleasant, heated debates, because everybody has opinions. The process tries to err on the conservative side and rather add too few than too many extensions.
 
 Alternatives
 ------------
--- a/proposals/0000-ghc-extensions.rst
+++ b/proposals/0000-ghc-extensions.rst
@@ -2,7 +2,7 @@ The GHC20XX process
 ===================
 
 .. author:: Alejandro Serrano, Joachim Breitner
-.. date-accepted:: 
+.. date-accepted::
 .. proposal-number::
 .. ticket-url::
 .. implemented::
@@ -121,17 +121,15 @@ Process
   in favor or against the inclusion of a given extension, or points the
   committee to a suitable existing survey result.
 
-**Alternative 1:**
-
 * At the start of the process, the secretary creates a PR with a proposal saying (roughly)
-   
+
     GHC20xx contains the following extensions in addition to those in
     GHC20(xx-1):
-    
+
     * (none yet)
-    
+
     and removes these extensions
-   
+
     * (none yet)
 
     This PR is a suitable place, besides the poll, for the wider
@@ -140,7 +138,7 @@ Process
     archive, and if the committee is missing an important piece of
     information (e.g. more code breaking than expected), to raise
     such a point.
- 
+
     We hope, however, that the community poll is sufficient to convey
     the level of community support and demand for specific extensions
     have, and want to discourage lengthy, opinion-based discussions of
@@ -152,10 +150,10 @@ Process
 * Within two weeks of the start of the process, every committee member is
   expected to send an initial list of which extensions they expect to be in
   GHC20xx to the mailing list.
-   
+
   Committee members are expected to take the Hackage statistics and the
   community vote into account.
-   
+
   These mails may contain justifications for why a certain extension is or is
   not included, but this is not required (or even expected).
 
@@ -168,7 +166,7 @@ Process
 
   It is absolutely ok to change one’s mind based on the explanations in the
   other members’ emails, or the general comments on the PR.
-   
+
   Long discussions of individual extensions are discouraged at this point. If
   there is controversy around an extension, it is a strong sign that it should
   simply not be included.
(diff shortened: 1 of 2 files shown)


simonpj:
> The extension is not intentional opt-in as a gate-keeper of advanced or potentially-unsafe feature. IncoherentInstances should not be enabled by default because of its potential unsafety, and the point of MagicHash is precisely that it it unlocks names not there by default.

Hard to understand.  Perhaps "The extension is one that users might plausibly want to be on all the time. This excludes experimental extensions that deliberately enable potentially-unsafe or unstable features, such as `IncoherentInstances`."

I'm not sure what the bit about MagicHash is about.

nomeata:
> Hard to understand. 

Yes, struggled with that one too, not sure if I made it any better. Will try again.

> I'm not sure what the bit about `MagicHash` is about.

Enabling `MagicHash` by default would defeat it’s purpose of being a (syntatic, artificial) barrier to low-level features, so we definitely would not want to add it to `GHC20xx`, right?

proposals/0000-ghc-extensions.rst
```diff
@@ -0,0 +1,120 @@
...
+
+1. The extension is (mostly) *conservative*: All programs that were accepted without the extension remain accepted, and with the same meaning. We prefer to take this as a strict criteria, and grant exceptions on a case-by-case basis.
+2. *New failure modes* that become possible with the extension are *rare* and/or easy to diagnose. These failure modes include new error messages, wrong inferred types, and runtime errors, for example. In other words, the *developer experience* should not get worse by adding this extension to the default mix.
+3. The extensions *complement* the design of standard Haskell. For example, ``MultiParamTypeClasses`` extends what's already there, and works well with the rest of features. If we suddenly decided you can use Lisp-like syntax, that would not complement the current design.
+4. The extension has been – and can reasonably be predicted to remain – *stable*.
+5. The extension is *not* intentional opt-in as a *gate-keeper* of advanced or potentially-unsafe feature. ``IncoherentInstances`` should not be enabled by default because of its potential unsafety, and the point of ``MagicHash`` is precisely that it it unlocks names not there by default.
```
nomeata, line 46:
```suggestion
5. The extension is one that users might plausibly want to be on all the time. This excludes experimental extensions that deliberately enable potentially-unsafe or unstable features, such as ``IncoherentInstances`` or ``MagicHash``.
```

int-index:
> Enabling `MagicHash` by default would defeat it’s purpose of being a (syntatic, artificial) barrier to low-level features, so we definitely would not want to add it to `GHC20xx`, right?

I didn't realize that gatekeeping was the reason for its existence. I though it was there simply because a postfix `#` steals syntax, e.g. `f# x` means different things with and without `-XMagicHash`.

proposals/0000-ghc-extensions.rst
```diff
@@ -0,0 +1,120 @@
...
+It has almost become a meme that any slightly-complex Haskell module starts with a dozen lines of ``{-# LANGUAGE #-}`` pragmas. This poses several problems:
+
+- beginners often see features behind an extension as "advanced" or "obscure", even though some like ``LambdaCase`` or ``NamedFieldPuns`` are quite simple;
+- some sorts of programming in Haskell – mostly type-level oriented – require enables lots of extensions (think of the usual combo ``MultiParamTypeClasses`` + ``FlexibleInstances`` + ``FlexibleContexts``). Sometimes the fact that there are different extensions, instead of a single `ExtendedTypeClasses` is mostly historical.
+
+For that reason, it's benefitial to pack those extensions which have "graduated" in terms of stability and usefulness into a single language version.
+
+We aim to lean on the conservative side in our selection of extensions: Not including an extension that we should have is relatively harmless (users of that extension will have to still set that manually, but they alreay have to do that, and we can add it alter), while including an extension that we should not have hurts more (any detriminal effect on developer experience now affects many users).
```
robrix, line 27:
```suggestion
We aim to lean on the conservative side in our selection of extensions: Not including an extension that we should have is relatively harmless (users of that extension will have to still set that manually, but they alreay have to do that, and we can add it later), while including an extension that we should not have hurts more (any detriminal effect on developer experience now affects many users).
```

robrix: This is a bit of a run-on sentence, and I found it a bit hard to parse. Maybe breaking it up after the first parenthetical:

> …we can add it later). Conversely, including an extension which we shouldn’t have hurts more…

proposals/0000-ghc-extensions.rst
```diff
@@ -0,0 +1,120 @@
...
+.. date-accepted:: 
+.. proposal-number::
+.. ticket-url::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/372>`_.
+.. sectnum::
+.. contents::
+
+This proposal sets up a process to introduce new language versions `GHC20XX`, extending `Haskell2010` as defined by the Report with several extensions that are considerd harmless, helpfun and uncontroversial.
```
robrix, line 14:
```suggestion
This proposal sets up a process to introduce new language versions `GHC20XX`, extending `Haskell2010` as defined by the Report with several extensions that are considerd harmless, helpful and uncontroversial.
```

proposals/0000-ghc-extensions.rst
```diff
@@ -0,0 +1,120 @@
...
+
+Motivation
+----------
+
+It has almost become a meme that any slightly-complex Haskell module starts with a dozen lines of ``{-# LANGUAGE #-}`` pragmas. This poses several problems:
+
+- beginners often see features behind an extension as "advanced" or "obscure", even though some like ``LambdaCase`` or ``NamedFieldPuns`` are quite simple;
+- some sorts of programming in Haskell – mostly type-level oriented – require enables lots of extensions (think of the usual combo ``MultiParamTypeClasses`` + ``FlexibleInstances`` + ``FlexibleContexts``). Sometimes the fact that there are different extensions, instead of a single `ExtendedTypeClasses` is mostly historical.
+
+For that reason, it's benefitial to pack those extensions which have "graduated" in terms of stability and usefulness into a single language version.
```
robrix, line 25:
```suggestion
For that reason, it's beneficial to pack those extensions which have "graduated" in terms of stability and usefulness into a single language version.
```

FWIW I think this idea of a feature “graduating” is an _excellent_ litmus test.

proposals/0000-ghc-extensions.rst
```diff
@@ -0,0 +1,120 @@
...
+
+We aim to lean on the conservative side in our selection of extensions: Not including an extension that we should have is relatively harmless (users of that extension will have to still set that manually, but they alreay have to do that, and we can add it alter), while including an extension that we should not have hurts more (any detriminal effect on developer experience now affects many users).
+
+We also aim to have a process that is efficient. For each extension, we are all very capable of having very long discussions about the merits, up-sides and downsides, and almost anybody will have an opinion on almost any extensions. But aiming to include only _uncontroversial_ extensions, we aim to avoid much of these discussion by construction. The process deliberately does not provide room for long per-extensions deliberations, neither within the committee or with the overall community, while still involving the community.
```
robrix, line 29: I’m curious what “uncontroversial” means, but haven’t finished reading yet. I don’t know if that suggests the doc should discuss this more up front or that it’s working well as a hook, tho 😄

proposals/0000-ghc-extensions.rst
```diff
@@ -0,0 +1,120 @@
...
+
+
+Proposed Change Specification
+-----------------------------
+
+We propose to introduce a process to develop new "language versions", dubbed ``GHC20XX``, which extend ``Haskell2010`` with some default extensions. This proposal does not set *which* extensions are included in the first version,, rather seeks to provide a framework to reach a conclusion in a constructive and timely manner.
```
robrix, line 35:
```suggestion
We propose to introduce a process to develop new "language versions", dubbed ``GHC20XX``, which extend ``Haskell2010`` with some default extensions. This proposal does not set *which* extensions are included in the first version, but rather seeks to provide a framework to reach a conclusion in a constructive and timely manner.
```

proposals/0000-ghc-extensions.rst
```diff
@@ -0,0 +1,120 @@
...
+
+Criteria
+^^^^^^^^
+
+We define a set of criteria that an extension should guide the committee members in their process of picking suitable extensions. These are not hard criteria, but often subjective, and not all need to be satisfied in all cases.
+
+1. The extension is (mostly) *conservative*: All programs that were accepted without the extension remain accepted, and with the same meaning. We prefer to take this as a strict criteria, and grant exceptions on a case-by-case basis.
+2. *New failure modes* that become possible with the extension are *rare* and/or easy to diagnose. These failure modes include new error messages, wrong inferred types, and runtime errors, for example. In other words, the *developer experience* should not get worse by adding this extension to the default mix.
```
robrix, line 43: Is it worth giving an example? e.g. something that changes the semantics of existing programs (e.g. how `-XScopedTypeVariables` can cause type variables to start unifying in e.g. a `where` clause in an `instance` declaration)

proposals/0000-ghc-extensions.rst
```diff
@@ -0,0 +1,120 @@
...
+3. The extensions *complement* the design of standard Haskell. For example, ``MultiParamTypeClasses`` extends what's already there, and works well with the rest of features. If we suddenly decided you can use Lisp-like syntax, that would not complement the current design.
+4. The extension has been – and can reasonably be predicted to remain – *stable*.
+5. The extension is one that users might plausibly want to be on all the time. This excludes experimental extensions that deliberately enable potentially-unsafe or unstable features, such as ``IncoherentInstances`` or ``MagicHash``.
+6. The extension has *widespread* usage.
```
robrix, line 47: …as measured by code on e.g. hackage, presumably?

proposals/0000-ghc-extensions.rst
```diff
@@ -0,0 +1,120 @@
...
+7. The extension is favored by the community, with many in favor, and very few opposed to its inclusion.
+
+
+Process
+^^^^^^^
+
+* 4 months before the expected GHC spring release day of 202x, the committee Secretary starts the GHC202x process.
+
+  They inform the committe, in an email to the mailing list, of all language extensions supported by the latest released GHC that are not in GHC202(x-1), which could be added. They also list all extensions *in* GHC202(x-1), which could be removed.
```
robrix, line 56:
```suggestion
  They inform the committee, in an email to the mailing list, of all language extensions supported by the latest released GHC that are not in GHC202(x-1), which could be added. They also list all extensions *in* GHC202(x-1), which could be removed.
```

Removed from 202x, right? Not retroactively dropped from 202(x-1)?

proposals/0000-ghc-extensions.rst
```diff
@@ -0,0 +1,120 @@
...
+    
+    * (none yet)
+    
+    and removes these extensions
+   
+    * (none yet)
+
+* In order to gather data on the criterium “widespread usage”, the secretary creates a tally of which extensions are used how often on hackage.
+
+* In order to gather data on the criterium “community spport”, the secretary runs a public poll on a suitable platform for one week where anyone
```
robrix, line 70:
```suggestion
* In order to gather data on the criterium “community support”, the secretary runs a public poll on a suitable platform for one week where anyone
```

proposals/0000-ghc-extensions.rst
```diff
@@ -0,0 +1,120 @@
...
+  can vote in favor or against the inclusion of a given extension.  
+
+* Within two weeks of the start of the process, every committee member is expected to send an initial list of which extensions they expect to be in GHC202x to the mailing list.
+   
+  Committe members are expected to take the Hackage statistics and the community vote into account.
```
robrix, line 75:
```suggestion
  Committee members are expected to take the Hackage statistics and the community vote into account.
```

proposals/0000-ghc-extensions.rst
```diff
@@ -0,0 +1,120 @@
...
+   
+  These mails may contain justifications for why a certain extension is or is not included, but not required (or even expected).
```
robrix, line 77:
```suggestion
  These mails may contain justifications for why a certain extension is or is not included, but this is not required (or even expected).
```

proposals/0000-ghc-extensions.rst
```diff
@@ -0,0 +1,120 @@
...
+Therefore, this proposal does *not* commit to a fixed cadence. Instead, 6 months after the first release of a version of GHC that supports a GHC20XX set, we evaluate the outcome, the process, and the perceived need of a next release. At that time we will refine the processes, if needed, and set a cadence.
+
+Costs and Drawbacks
+-------------------
+
+The implementation cost seems small.
+
+The cost of a GHC20xx extension is that, upon reading a file with ``{-# LANGUAGE GHC20xx #-}``, the reader does not immediatelly know the set of enabled extensions; this may hamper readability of code.
+
+The costs of this process is that it binds volunteer time, and there is is risk of unpleasant, heated debates, because everybody has opinions. The process tries to err on the conversative side and rather add too few than too many extensions.
```
robrix, line 103:
```suggestion
The costs of this process is that it binds volunteer time, and there is is risk of unpleasant, heated debates, because everybody has opinions. The process tries to err on the conservative side and rather add too few than too many extensions.
```

robrix:
Posted some notes, mostly typo fixes ☝🏻 I think this is an excellent proposal, although I’m ambivalent about the idea of GHC20xx-bundles themselves (TL;DR: it seems pragmatic but opaque and I tend to prefer semantic groupings). Specifically, I think this addresses the worry I had coming in that “uncontroversial” could be particularly difficult to satisfy for almost _any_ extension.

Nice work!

proposals/0000-ghc-extensions.rst
serras: Yes, and also about what people say in the polls.

serras:
@robrix "uncontroversial" is indeed a bit vague, but it's supposed to cover several scenarios:
- when the community-as-a-whole seems to agree, even though some people may not use the extension;
- when there's a consensus that a feature has the right design. For example, I think that `FlexibleContexts`, which is rooted on academic papers and tested through decades, is a good example.

On a more personal note, I used to share your feeling of "everything is going to be controversial". But there's so much low-hanging fruit, extensions which everybody enables and would help us explain GHC/Haskell as a more unified story.

proposals/0000-ghc-extensions.rst
serras: Maybe `ScopedTypeVariables` is not the best example, because I guess it might be one of the extensions for which some of the requirements are waived due to its overwhelming use.

serras:
My two cents on the unresolved questions is:
1. Yes, we should make it a Cabal language too. In fact, this would be my preferred option altogether, but we need to keep the `-XGHC2021` and `{-# language GHC2021 #-}` for those places where updating Cabal it not an option.
2. I would say yes to `ghci` too. In some sense, approving this new language is also stating that this is the preferred language for working with GHC, so it makes sense to make it the default in new versions. In fact, running `ghc MyFile.hs` should also default to the latest `GHC20XX`.

proposals/0000-ghc-extensions.rst
nomeata:
```suggestion
We aim to lean on the conservative side in our selection of extensions: Not including an extension that we should have is relatively harmless (users of that extension will have to still set that manually, but they already have to do that, and we can add it alter). Conversely, including an extension that we should not have hurts more (any detriminal effect on developer experience now affects many users).
```

proposals/0000-ghc-extensions.rst
nomeata: Ah, no fun?

proposals/0000-ghc-extensions.rst
nomeata: Hmm, is it unclear? My intention is that it means “there is no (or little) controversy about turning this on by default”. Maybe it gets clearer later… if not, please do suggest improvements to the wording!

proposals/0000-ghc-extensions.rst
nomeata:
```suggestion
  They inform the committee, in an email to the mailing list, of all language extensions supported by the latest released GHC that are not in GHC202(x-1), which could be added. They also list all extensions *in* GHC202(x-1), which might be omitted in GHC202(x-1) (likely a rare thing).
```

nomeata:
> Posted some notes, mostly typo fixes ☝🏻

Thanks a lot for that!

jvanbruegge:
ghci is probably the biggest win for me with `GHC202x`.  Often I fire up ghci outside of a project to try something out quickly and it gets quite annoying to enable 6 extensions at minimum

erikd:
> ghci is probably the biggest win for me with GHC202x.

You can have this **now** with a `~/.ghci` file.

bChiquet:
From reading the `Motivation` section, I am not sure I understood what this proposal aims to address. Here are some possible interpretations:
* newcomers may be intimidated by the "wall of pragmas" at the beginning of a file.
* Some language features emerged gradually, and there are multiple pragmas for them. Merging pragmas that are commonly used together would add clarity to this.
* having lots of pragmas at the start of a file is perceived as a problem.
* There is a lack of sharing about what constitutes "common" pragmas, and a debating space would help float a consensus about it.
Are these interpretations correct ?

In the `Alternatives` sections, the variable cycle length seems to be a variant of the proposal rather than an alternative.

AntC2:
Re Motivation, 1st bullet: the most scary-looking extension for beginners is `UndecidableInstances` And yet: nobody's been able to demonstrate any program that goes wrong with that switched on (providing it compiles without sending the compiler into a loop); it's almost always needed if you're using any of the other extensions likely to qualify for `GHC20xx`.

>  (think of the usual combo `MultiParamTypeClasses` + `FlexibleInstances` + `FlexibleContexts`).

Yes that combo's usual, but not usually sufficient on its own; and the usual complements are problematic, going by your Criteria:

* `MultiParamTypeClasses` in practice needs `TypeFamilies` and/or `FunctionalDependencies`, both of which are controversial, especially combined with ...
* `FlexibleInstances` puts you in a position to write overlapping instances. (That's to say: GHC won't stop you writing instances that overlap, and won't warn they're overlapping until something upsets its instance resolution algorithm; program incoherence can lurk undiscovered for a long time.) Those interact badly with `MultiParamTypeClasses`.
* `FlexibleContexts` almost always end up needing `UndecidableInstances`, because they evade the coverage conditions -- which we all agree are too strict, but it's difficult to see how to relax them and retain coherence.


So I disagree with your comment in Criteria 3. " `MultiParamTypeClasses` ... works well with the rest of features." Precisely because it _doesn't_ combine well is why it wasn't included in H2010, although it's been available and relatively stable in two compilers since ~2000.

I'd suspect the Criteria are going to be _too_ conservative, with the result any beginner will find they need to switch on extensions additional to `GHC20xx` (a different set of extensions depending on what their program is doing). Then they're still going to have every "Haskell module starts with a dozen lines of {-# LANGUAGE #-} pragmas. " -- maybe a half-dozen + `GHC20xx`. All this has done is added an extra layer of indirection.

I'm lukewarm as to whether the `GHC20xx` exercise is worth the effort.

A couple of data points:

* GHC vintage ~2006 had a `glasgow-exts` option. That represented a coherent and effective set of functionality. I'd want to see a very strong argument before excluding from `GHC20xx` any extension that was included in that. But I doubt they'd all meet the Criteria.
* Hugs of same vintage has 'HugsMode', fairly similar in scope to `glasgow-exts`. `OverlappingInstances` is additional; `IncoherentInstances` is further additional.

proposals/0000-ghc-extensions.rst
```diff
@@ -0,0 +1,120 @@
...
+    * (none yet)
+    
+    and removes these extensions
+   
+    * (none yet)
+
+* In order to gather data on the criterium “widespread usage”, the secretary creates a tally of which extensions are used how often on hackage.
+
+* In order to gather data on the criterium “community support”, the secretary runs a public poll on a suitable platform for one week where anyone
+  can vote in favor or against the inclusion of a given extension.
```
tfausak, line 71: It's not exactly the same, but I have asked "Which language extensions would you like to be enabled by default?" in the state of Haskell survey for the past three years:

- https://taylor.fausak.me/2019/11/16/haskell-survey-results/#s2q5
- https://taylor.fausak.me/2018/11/18/2018-state-of-haskell-survey-results/#question-038
- https://taylor.fausak.me/2017/11/15/2017-state-of-haskell-survey-results/#question-21

I plan to ask it again this year. If there's something I can do to make it more useful to this GHC20xx process, please let me know!

nomeata: @simonmar actually just mentioned that, and maybe that is simply enough to get some community data. We could make the process flexible in this point:
```suggestion
* In order to gather data on the criterium “community support”, the secretary runs a public poll on a suitable platform for one week where anyone
  can vote in favor or against the inclusion of a given extension, or points the committee to a suitable existing survey result.
```

nomeata: If I’d do a dedicated poll, I might ask for “Yes please”/“Don’t care”/“No, please don’t” to also gauge how much resistance it might have. But maybe not worth the bother.

tfausak: I like how that sounds! Other people have asked for something like this before. I've held off because it's already an intimidating question. But I think I could create a simple enough UI/UX if "Yes please" is 👍/upvote, "No, please don't" is 👎/downvote, and "Don't care" is no vote at all.

goldfirere:
I've added an alternative to this proposal, outlining a process (originally based on one @serras included in an email to the committee) that has a structured place to debate individual extensions. I worry about endless debate... but I also worry that if we don't make a home for the endless debate, then it will show up where we don't want it.

serras:
@AntC2 lets try to keep this conversation focused on the process itself. Maybe we could remove the examples to not steer the process already in one direction?

AntC2:
@serras I don't think we'll be able to clarify the process -- particularly the Criteria -- without seeing how they apply to particular examples.

So @nomeata (and some of the discussion on the Steering Committee) put some examples, to keep the discussion concrete. All I did was tease out the consequences of those examples wrt the Criteria. I wasn't advocating for any particular extension being in or out. My points were more like: if you're going to include extension X (which the Proposal considers), it doesn't make sense without extension Y, but the Criteria don't seem to allow extension Y. So the Criteria need some extra grounds like 'makes a coherent and expressive set of functionality'.

Particular example: `MultiParamTypeClasses` doesn't make sense without `FlexibleInstances`. (Both are considered in the Proposal.)  `FlexibleInstances` introduces _new failure modes_, contra Criteria 2 -- namely overlapping instances and nasties like orphan instances.

goldfirere:
It sounds like perhaps @AntC2 is suggesting a new criterion: that the set of extensions form a cohesive whole. Maybe this can be encapsulated by changing the existing "complement the design" criterion like this:

* The extensions, taken alongside standard Haskell, form a cohesive whole.

I've removed the example, because anything I think of would create more controversy, not less.

More generally, though, I'm not very fussed about the exact criteria. As the proposal says, the criteria are guidelines and may end up being contravened.

nomeata:
> I've removed the example, because anything I think of would create more controversy, not less.

Looks like you didn’t push yet.

goldfirere:
I didn't intend to update the proposal. That bullet point was posted in that comment only, to facilitate discussion. I didn't want to update the proposal without further agreement on it, first.

nomeata:
Ah, I didn't get that. All well then :-)

serras:
I agree with @goldfirere's suggestion, that's a good way to phrase the criterion.

serras:
Is the difference between alternative 1 and 2 simply whether discussion takes place in a single PR or in a different repo?

nomeata:
Superficially yes. But I think fundamentally it is the question: do we expect/encourage/faciliate/give room to a discussion of the inclusion of each individual extensions by the communit, or do avoid/discourage that.

serras:
I would prefer alternative 2, as it gives an official place to have this discussion with a broader community.

serras:
Any further opinions on the proposal? Should we send this to the Committee for review?

nomeata:
I’ll do that now – it’s one way to get opinions :-)

simonpj:
Looking at the proposal, I'd love it to be more explicit about when and how community input is possible.   Clearly in the survey yes, but they can also contribute to the debate about which extensions should be in/out can't they? Let's say that explicitly, making clear when and how they can contribute.

proposals/0000-ghc-extensions.rst
```diff
@@ -0,0 +1,249 @@
...
+   
+    GHC20xx contains the following extensions in addition to those in
+    GHC20(xx-1):
+    
+    * (none yet)
+    
+    and removes these extensions
+   
+    * (none yet)
+
```
nomeata, line 136:
```suggestion

    This PR is a suitable place, besides the poll, for the wider commity to weigh in.
```
@simonpj  like this?

nomeata:
```suggestion

    This PR is a suitable place, besides the poll, for the wider
     community to weigh in. The community is invited to follow the
     committee votes, rationales and discussion on the public email
     archive, and if the committee is missing an important piece of
     information (e.g. more code breaking than expected), to raise
     such a point.
 
     We hope, however, that the community poll is sufficient to convey
     the level of community support and demand for specific extensions
     have, and want to discourage lengthy, opinion-based discussions of
     the merits of extensions.
 
     If you miss your favorite extension in the list, please remember
     that you can still use it (by setting the flag explicitly), and
     that it can still go in next round.
```

ByteEater-pl:
A question: if this is going to be on by default in GHCi, is there going to be an off switch too, e.g. `NoGHC2021` (only necessary for the first one, since specifying an older one would AFAIU disable those added in later ones)?

There's indeed a long standing controversy around `FlexibleInstances`. And although there apparently hasn't been much discussion about them recently, I remember proposals (for new features or extensions) which would enable (or even subsume) many use cases desired by users of `FlexibleInstances`, I believe that's the best way forward. And it's also a more conservative alternative for a step forward than including this controversial extension in `GHC2021`. So I suggest mentioning this consideration somewhere (the proposal is already accepted, so I'm not sure where). And perhaps even emphasizing that discussion of less controversial alternatives (even at a stage of lesser maturity, without a proposal, or still in early research) to the extensions proposed for inclusion is actually welcome. (If those alternatives aren't yet implemented, sometimes it may be better to wait and then decide, possibly ending up with a better set of included extensions a few years later than paint ourselves into a corner by adding something like `FlexibleInstances`.) As an additional benefit, the perspective of inclusion may also encourage people to put more work into discussed alternatives.

nomeata:
> A question: if this is going to be on by default in GHCi, is there going to be an off switch too, e.g. NoGHC2021 (only necessary for the first one, since specifying an older one would AFAIU disable those added in later ones)?

I’d expect that `-XHaskell98` or `-XHaskell2010` would achieve that.

ByteEater-pl:
Indeed, I'd expect that too. But that's still different from the present default. Which it would likely be useful to have available with a single switch.

adamgundry:
I think `NoGHC2021` would be rather confusing, because it isn't clear what it *does* enable. Just as we don't currently have e.g. `NoHaskell2010`. But it would make sense to retrospectively come up with a new name for the current default behaviour (e.g. `GHCLegacy` for the sake of argument), so it can be explicitly requested in the future after the default changes.

ByteEater-pl:
Yup, the name `NoGHC2021` was somewhat tongue in the cheek and actually intended for bikeshedding. How about simply `GHC2020`?

PR #373, opened 2020-10-28
Type family role annotations

treeowl:
Add role annotations for type and data families.

[Rendered proposal](https://github.com/treeowl/ghc-proposals/blob/family-roles/proposals/0000-family-roles.md)


goldfirere:
I think this is feasible. This has actually been proposed before (though it pre-dates this ghc-proposal process): See https://gitlab.haskell.org/ghc/ghc/-/issues/8177 and https://gitlab.haskell.org/ghc/ghc/-/wikis/roles#proposal-roles-for-type-families, which include both the main proposal and roles on data family instances.

For this proposal, though, I think it would need more detail in its specification about what is and is not allowed. For example, a representational parameter must not only be unmatched in the LHS, but there are restrictions of how it is used on the RHS.

The other improvement I would want is a more compelling motivation. As the ticket above shows, we've known about this problem and how to fix it for a long time -- but we haven't fixed it, lacking a compelling motivation. The motivation currently written is contrived. Is there a real use-case? Is there more than one? These would be necessary for me to be in support.

Ericson2314:
I like this, but I must admit it will expose more of the existing orphan issues with fancy constraints in open type families? (Say one does conflicting instances of newtypes of the same underlying type in different modules.)

That said, we could also restrict to just `nominal` and `phantom`, for a first version, which avoids all that.

treeowl:
> I think this is feasible. This has actually been proposed before (though it pre-dates this ghc-proposal process): See https://gitlab.haskell.org/ghc/ghc/-/issues/8177 and https://gitlab.haskell.org/ghc/ghc/-/wikis/roles#proposal-roles-for-type-families, which include both the main proposal and roles on data family instances.
> 
> For this proposal, though, I think it would need more detail in its specification about what is and is not allowed. For example, a representational parameter must not only be unmatched in the LHS, but there are restrictions of how it is used on the RHS.
> 
> The other improvement I would want is a more compelling motivation. As the ticket above shows, we've known about this problem and how to fix it for a long time -- but we haven't fixed it, lacking a compelling motivation. The motivation currently written is contrived. Is there a real use-case? Is there more than one? These would be necessary for me to be in support.

Yes, the RHS restrictions are presumably the same as the ones under data constructors for current role annotations. My actual use case involves a tree whose nodes are unlifted arrays.... I thought that was a little too much machinery for a motivation section, but I can use it if you like!

treeowl:
@goldfirere, my actual use-case is perfectly height-balanced trees. Here's the idea.

First, type-level naturals and fake (but very efficient) singletons for them.

```haskell
-- | Type-level naturals
data Nat = Z | S Nat

-- | This type pretends to be defined
-- @
-- data SNat (n :: Nat) where
--   SZero :: SNat 'Z
--   SSucc :: !(SNat n) -> SNat ('S n)
-- @
newtype SNat (n :: Nat) = SNat Int
type role SNat nominal
pattern SZero :: () => n ~ 'Z => SNat n
pattern SSucc :: () => n ~ 'S m => SNat m -> SNat n
```

Now we could imagine defining a tree like this:

```haskell
data Tree a = forall n. Tree !(SNat n) !(Tree_ n a)
data Tree_ (n :: Nat) (a :: Type) where
  InternalNode :: !(Array (Tree_ n a)) -> Tree_ ('S n) a
  LeafNode :: !(Array n a) -> Tree_ 'Z a
```

Unfortunately, this representation is somewhat inefficient: there's a heap object representing a tree node between each array and its child. We can get around that using unlifted arrays (see the master branch of `primitive-unlifted`; hopefully these will soon be supported natively in GHC via `BoxedRep`) and a type family:

```haskell
data Tree a = forall n. Tree !(SNat n) (Tree_ n a)

type family Tree_ (n :: Nat) (a :: Type) :: TYPE 'UnliftedRep where
  Tree_ 'Z a = Array# a
  Tree_ ('S n) a = UnliftedArray# (Tree_ n a)
```

This representation roughly *halves* the number of pointers that must be followed to reach a leaf. But ... oops ... we suddenly have a `nominal` role for the element type, which sucks. A little caveat: even with type family roles, we'll still have trouble using this definition in practice because of [GHC #10016](https://gitlab.haskell.org/ghc/ghc/-/issues/10016), but let's not let one problem block up an unrelated one.

Please let me know if I should incorporate this whole thing into the proposal.

goldfirere:
> Please let me know if I should incorporate this whole thing into the proposal.

I think so, yes. Keep the existing much simpler example, but the example above concludes that this proposal would halve the number of pointers in a structure -- a very nice win.

phadej:
I don' think that example needs `MagicHash` (one can keep the example simple). Transforming from GADT to data family already produces better core (there won't be any sums), and allows *lazy pattern matching*. This is why [`vec`](https://hackage.haskell.org/package/vec) has `DataFamily` definition as well. Sadly, I cannot `coerce` `DataFamily.Vec`.

treeowl:
@phadej, I think I'll go with "both". For flow, I'll probably stick your application and my own in the "Examples" section rather than "Motivation", but I'll refer to them from there.

treeowl:
Or maybe "effects"? Whatever.

treeowl:
@phadej, unfortunately, I'm not familiar enough with `vec` to be able to use it without aid. Do you think you could write a blurb?

phadej:
@treeowl, I don't know how much

```haskell
data Vec n a where
    VNil  ::                 Vec 'Z     a
    (:::) :: a -> Vec n a -> Vec ('S n) a
```

needs an introduction. That type (sans renamings) is well-known.

The *lazy pattern matching* I was referring comes up in (surprisingly common?)
problems where one uses `Vec n` to model `n` things which definitions
are interconnected: We use `mfix` or `RecursiveDo` to wire them up.
A concrete example is `letrec` bindings in a well-scoped AST type.

```haskell
calculateSomeFixpoint :: MonadFix m => SNat n -> Arguments n -> m (Vec n a)
calculateSomeFixpoint n args = mdo
    -- we calculate some kind of fixpoint
    xs <- case snat of
        -- zero case is trivial
        SZ -> return VNil 

        -- cons case is where something interesting happens
        SS -> case xs of
             -- here we must be lazy while matching on `xs`,
             -- and we know it should always be cons, :::
             ~(x ::: xs') -> ... -- but GHC would disagree with us.
             
    return xs
```

With `data family` version of `Vec`:

```haskell
data family   Vec (n :: Nat) a
data instance Vec 'Z     a = VNil
data instance Vec ('S n) a = a ::: !(Vec n a)
```

the lazy pattern match is possible, as there is no refinement of indices going on
in `case`: you have to know the index before you can `case`.
But this `Vec` cannot be `coerce`d.
How big problems this is in practice? Slightly annoying.
The GADT and data-family `Vec` module interface is for most parts
one-to-one, roles is one of differences, and probably the subtlest one.

Similar example can be made about `HList`, height indexed perfect trees etc.

Ericson2314:
Off topic, but @phadej, since there are no tags/sums, does this work?
```haskell
data family   Vec (n :: Nat) a
data instance Vec 'Z     a = VNil
data instance Vec ('S n) a = a ::: {-# UNPACK #-} !(Vec n a)
```

PR #375, opened 2020-11-03
Update GRC.rst

blackheaven:
My goal with this pull request is the change the emotional state responsibility.

When I read this sentence:

> [..] we avoid forms of expression that might make our dialogue partner feel attacked [..]

I conclude that I have two responsibilities:

1. I should not use "aggressive" language
2. I have to worry of my partner feelings

In my opinion, while the first point is perfectly valid, I can control what I say, the second one is at least annoying.

 1. It can easily lead to the [mind reading](https://en.wikipedia.org/wiki/Jumping_to_conclusions) cognitive distortion which can increase anxiety.
 2. It does encourage emotional ownership, since it would imply outsourcing its control to someone else.

Finally, as a person, I am responsible of my emotional state, I treat other as responsible of theirs, and I hope to be recognized as such.

### DIFF
--- a/GRC.rst
+++ b/GRC.rst
@@ -48,8 +48,8 @@ Specifically, we aspire to these qualities:
 * We strive to be scrupulously polite at all times.  There should be no
   rudeness, name-calling, or harassment in our communication.
 
-* Where we disagree with someone, we avoid forms of expression that might make
-  our dialogue partner feel attacked, humiliated, demeaned, or marginalised.
+* Where we disagree with someone, we avoid forms of expression which intends to
+  attack, humiliate, demean, or marginalise our dialogue partner.
   Our critique should always be of specific statements and claims, never of
   people.
 


goldfirere:
I am sympathetic to your motivation here, but I disagree with the change you've suggested: the new text suggests that anyone is in conformance with the GRC as long as they (claim to) intend not to offend. The current text suggests that communication which "might" offend falls outside the guidelines. This perhaps is problematic, in that any text "might" offend.  But if someone were appealing to this GRC in saying that some communication falls outside the guidelines, then the e.g. chair of the relevant committee would have to assess whether they believed the communication could reasonably be understood to be offensive. It's always going to come down to judgment, but I support the current text in how it frames this judgment.

Put another way: You argue that the current text requires the writer to mind-read how the reader will feel. I argue that the new proposed text requires that the reader mind-read to determine the writer's intent. In deciding between these two, I favor putting the burden on the writer; then one person holds the burden instead of the many people who will read the text and have to discern the writer's intent individually.

simonpj:
I agree with Richard, but I wonder if it might be helpful to unpack the guidance a bit.  One guideline I use is this: never use language that attributes bad motives or attributes to the *person* you are addressing; instead focus on the concrete, technical pros and cons of what they are saying.

So, for example, it's unhelpful to say "you are an idiot" even if you don't intend to be offensive, because it suggests that the person *themself* is somehow bad or inadequate.   You don't need to read minds to figure out if they'll be offended -- rather, because it concerns the person rather than the idea, it's just best not to do it.

Of course, things are often more nuanced.   For example, it's also unhelpful to say "that is an idiotic idea", because it suggests that the other party is an idiot, even though it doesn't say so directly.  Similarly "If you cared to read the manual you would know..." suggests that they are lazy or negligent.

Better to concentrate on the facts.   For example, "I would like to explore the consequences of your approach for case X... bad thing Y would happen.  Can you see any way to mitigate that?"

Whether or how to incorporate these thoughts in the guidelines is not clear to me.

blackheaven:
@goldfirere Thank you for having spent your time to write this explanation, but the motivations are not really clear to me.

It does not really matter because you have made me realized that, since I have not ran a big community for many years, I have no idea about the situations you have to arbitrate, and how you use the GRC.

I was going to close my PR but it seems to have triggered a debate :) feel free to close it as soon as it is no longer useful.

Anrock:
It seems to me that we're reinventing [Graham's Hierarchy of Disagreement](https://upload.wikimedia.org/wikipedia/commons/7/7c/Graham%27s_Hierarchy_of_Disagreement.svg) from [How to Disagree](http://www.paulgraham.com/disagree.html) here.

PR #380, opened 2020-11-16
GHC2021

nomeata:
[Rendered](https://github.com/ghc-proposals/ghc-proposals/blob/ghc2021/proposals/0000-ghc2021.rst)

This PR will go along the process of defining GHC2021, as per the process outlined in [#372](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0372-ghc-extensions.rst).

This PR is a suitable place, besides the poll, for the wider community to weigh in. You are invited to follow the committee votes, rationales and discussion on the public email archive, and if the committee is missing an important piece of information (e.g. more code breaking than expected), to raise such a point.

We hope, however, that the community poll is sufficient to convey the level of community support and demand for specific extensions have, and want to discourage lengthy, opinion-based discussions of the merits of extensions.

If you miss your favorite extension in the list, please remember that you can still use it (by setting the flag explicitly), and that it can still go in next round.


nomeata:
I have opened this PR already to have a scratch space to let people know about the process and progress. My plan is to bring the data from hackage and the poll into a nice presentable form, and then start the actual committee deliberation.

proposals/0000-ghc2021.rst
```diff
@@ -0,0 +1,63 @@
...
+
+The ``GHC2021`` language extension set comprises of the following language extensions:
+
+* Everything in ``Haskell2010``, plus
+* TBD 
+
+Data
+----
+
+The following table contains the information that was gathered to help the committee decide, based on an analysis of hackage and a community poll, as well as the tallied votes. At a committe size of 11, **8** votes are necessary to inclue an extension in ``GHC2021``:
```
Anrock, line 25: `inclue` -> `include`

Ericson2314:
What is the story for backwards compatibility with these `GHC<year>`? In particular, I think it's quite likely that `ScopedTypeVariables` and `TypeFamilies` would end up with in a `GHC2021`, but I would *not* want them in a `GHC2023` because `ScopedTypeVariables` should be replaced by `\ @a ->` and `TypeFamilies` by promoted functions.

maralorn:
The way I read the process proposal it is definitely okay to drop extensions again, if it's decided an error was made. (I guess one intention is to make entering an extension a little bit less final -> less controversial.)

Ericson2314:
> I guess one intention is to make entering an extension a little bit less final -> less controversial.

But that intention is antithetical to the idea that extensions quite useful today can be a bad idea tomorrow.

nomeata:
> The way I read the process proposal it is definitely okay to drop extensions again, if it's decided an error was made. (I guess one intention is to make entering an extension a little bit less final -> less controversial.)

Exactly. And unlike the old `-fglasgow-exts` flag, if GHC2022 has less extensions than GHC2021, code that opts into GHC2021 still works (if the extensions themselved stay compatible, of course)

nomeata:
I have started the voting period, see this thread: https://mail.haskell.org/pipermail/ghc-steering-committee/2020-November/001902.html

I will update the table in the rendered proposal irregularly, and as votes come in.

Topsii:
[From the mailing list:](https://mail.haskell.org/pipermail/ghc-steering-committee/2020-November/001906.html)

> > InstanceSigs: no
> >   -- ^ It does feel mostly innocuous on the surface, but it has a
> >   probably surprising semantics. Considering that it's quite rarely
> >   used, probably better left out.
> 
> Interesting, I don’t see a mention of this on
> [https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#extension-InstanceSigs](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#extension-InstanceSigs)
> Can you elaborate?

According to the documentation the signature can be different from the signature in the class declaration.
> The type signature in the instance declaration must be more polymorphic than (or the same as) the one in the class declaration, instantiated with the instance type.

Essentially it is (sometimes?) possible to omit constraints in the signature. This is even possible for class constraints other than the class being instantiated. [Here's an example ](https://gitlab.informatik.uni-bremen.de/haslop/min-depth-sn/-/blob/84cee9067bb5d5d03a0f0faf9bfbcf45c7585b61/src/MinDepthSN/Vars.hs#L159-166)from my code that shows what I mean. After I wrote that code I wondered for a moment if this instance signature would trigger some kind of erasure of the constraint. I suspect this is not the case and the instance signature does never affect the compiled output. But it confused me.

nomeata:
Thanks for the example! I don’t find anything confusing here. Based on the documentation, I expect you can even do the following:
```
Prelude> :set -XInstanceSigs
Prelude> class Foo a  where foo :: a -> Int -> Int
Prelude> instance Foo Bool where foo :: a -> b -> b; foo = const id
```
and you can – leaving out constraints is just one way a function can be “more polymorphic”.

But if it is confusing to some, it _might_ be a reason to leave it out (although the fact that it's mostly educators asking for it is significant for me)

Topsii:
In the class definition the constraint of the class that is being defined (`Foo a =>`) is missing from the signature `foo :: a -> Int -> Int`.
From reading the documentation I conjecture that the rationale for allowing a more polymorphic signature in the instance definition is to support the use case of similarly omitting the constraint of the class that is being instantiated (`Foo a =>`).

If I see a type signature `foo :: a -> b -> b` in the context of an instance for `Foo`, it might be plausible to infer from the context that the signature is actually `foo :: Foo a => a -> b -> b` (though why not `Foo b => a -> b -> b`?  `class Foo a where` clearly specifies that the constraint applies to `a` and not `b` in the class definition, but nothing similarly resolves this ambiguity in the instance definition).

Is there another rationale for allowing a more polymorphic signature? It says the code for the method is typechecked against the (more polymorphic) signature. That is not what confuses me though. This typechecks in my head too. The confusion starts when I think of a situation where `foo` (the version from the `Bool` instance) is called and I (naturally or naively?) assume the signature `foo :: a -> b -> b` applies to this situation. I imagine a situation where someone reads the instance definition and profits from the (more polymorphic) instance signature in the sense that he then does not need to bother anymore to look up the (actually less polymorphic) signature in the class definition. After all 
> One stylistic reason for wanting to write a type signature is simple documentation.

```
*> :set -XInstanceSigs
*> class Foo a  where foo :: a -> Int -> Int
*> instance Foo Bool where foo :: a -> b -> b; foo = const id
*> foo True True

<interactive>:4:10: error:
    • Couldn't match expected type ‘Int’ with actual type ‘Bool’
    • In the second argument of ‘foo’, namely ‘True’
      In the expression: foo True True
      In an equation for ‘it’: it = foo True True
*> :t foo
foo :: Foo a => a -> Int -> Int
*> :t foo :: Bool -> _hole

<interactive>:1:1: error:
    • Couldn't match type ‘_hole1’ with ‘Int -> Int’
      ‘_hole1’ is a rigid type variable bound by
        an expression type signature:
          forall _hole1. Bool -> _hole1
        at <interactive>:1:8-20
      Expected type: Bool -> _hole1
        Actual type: Bool -> Int -> Int
    • In the expression: foo :: Bool -> _hole
```
I'd like to know: How is this not confusing? I agree that InstanceSigs help a lot to work out the types. And I think it is rather unusual (well, definitely not the norm) if a class method implementation admits a more polymorphic signature than the method signature in the class definition. But it worries me that one can easily give a signature unrelated to the actual signature if your implementation admits it. Actually personally I don't mind `InstanceSigs` being in GHC2021. I just think it's a dangerous detail that is easily overlooked and someone more experienced should assess the risk.

nomeata:
>  I just think it's a dangerous detail that is easily overlooked and someone more experienced should assess the risk.

Duly noted! I’m tempted to respond in detail, but this is not the right place to discuss design choices of language extensions that have been created a long time…

goldfirere:
While I laud @nomeata's restraint in not cluttering this conversation, I lean the opposite direction, in order to correct the (permanent) record being built here.

Tersely: An InstanceSig does not give a new type to the method. Its effect is restricted only to the method definition in the instance. Any variables brought into scope in the instance head are in scope in an InstanceSig, and the class variable must be instantiated according to the class variable instantiation implied in the instance head.

More drawn out: It seems that @Topsii is worried that an InstanceSig will somehow make a method easier to call (more polymorphic methods are easier to call), say by dropping constraints. This won't happen: the type of the method is the one as given in its declaration in the class (prepended by the class constraint). An InstanceSig will not change this. Let's look at an example:

```hs
class C a where
  foo :: Show a => a -> Int -> Int

instance C [b] where
  foo :: [b] -> c -> c
  foo _ x = x
```

This will be accepted. The type of `foo` is `(C a, Show a) => a -> Int -> Int`. The instantiation implied by the instance head is `a |-> [b]`. Thus, the instantiated type of `foo` is `(C [b], Show [b]) => [b] -> Int -> Int`. The given type of `foo` (`[b] -> c -> c`) is more general than this instantiated type. (That is, if we have `x :: (C [b], Show [b]) => [b] -> Int -> Int` and `y :: [b] -> c -> c`, then `x = y` type-checks.) Furthermore, the definition of `foo` in the instance type-checks against its signature. So all is well.

We might then ask: will the following work?

```hs
z :: Bool
z = foo ['x'] False
```

Given the fact that the first argument to `foo` is a list, we will use our instance -- and the type given in the instance can be specialized to `[Char] -> Bool -> Bool`, as required in the definition for `z`. Yet this will be rejected, because of a mismatch between `Int` and `Bool`. This is because the InstanceSig does not change the type of occurrences of `foo` -- the problem I believe @Topsii is worried about.

Bottom line: I don't think there is a risk here. But there may be some confusing text in the manual. @Topsii, if you believe the current explanation of InstanceSigs to be confusing, would you open a ticket at https://gitlab.haskell.org/ghc/ghc/? We can then update the manual. The most helpful way to open this ticket is to suggest new wording -- as the confused person, you're in the best seat to understand what might have avoided the confusion in the first place. Thanks!

Topsii:
I opened a [ticket](https://gitlab.haskell.org/ghc/ghc/-/issues/19017) though in my opinion it's not about the documentation. Feel free to give in to temptation there, @nomeata.

AntC2:
a plea against `ScopedTypeVariables`, but for `PatternSignatures`

I see Richard's started a [thread on the Committee mailing list](https://mail.haskell.org/pipermail/ghc-steering-committee/2020-December/001973.html). You might remember there was a proposal a long time ago from @nomeata IIRC to re-introduce the `PatternSignatures` part of `ScopedTypeVariables`. I think this is the part Richard likes (as do I); it's the delta between `ScopedTypeVariables` and as-was `PatternSignatures` that introduces the `forall` scoping of tyvars from a signature.

Note also there's soon to be `ResultSignatures`.

Currently if you ask for `PatternSignatures`, you get a deprecation warning and GHC treats it as if `ScopedTypeVariables`. Would it stretch beyond the scope of GHC2021 to ask to re-introduce the as-was `PatternSignatures`?

nomeata:
> Currently if you ask for `PatternSignatures`, you get a deprecation warning and GHC treats it as if `ScopedTypeVariables`. Would it stretch beyond the scope of GHC2021 to ask to re-introduce the as-was `PatternSignatures`?

I think so. But, as I ~threatened~ elsewhere already: If `ScopedTypeVariables` does not make it into GHC2021, I will re-propose #119 so that it can maybe become part of GHC2022 or so.

JakobBruenker:
Perhaps I missed it, but one thing that's not quite clear to me: Is this an extension that has to be explicitly enabled or does it replace the default -XHaskell2010 that GHC currently uses when launching it without any extensions?

nomeata:
I expect it’d become the new default mode. One of the main benefits we want get out of this is that using ghci is more convenient, i.e. you don't have to enable lots of extensions there (where you have to do it every time you restart `ghci` and it, in my opinion, thus hurts more than when editing a file where you add the pragma once).

JakobBruenker:
@nomeata That sounds very reasonable. I also just noticed that https://github.com/ghc-proposals/ghc-proposals/pull/372 does mention this:

> When ghc is used without an explicit language choice, the latest GHC20xx known to GHC is used. This applies in particular to uses of ghci.

phadej:
Recently Safe Haskell resurfaced in discussions (on reddit).

One argument that Safe Haskell poses the hard constraint on which exceptions could be in `GHC202x`: *GHC202x should be Safe Haskell compatible*. The alternative means phasing out Safe Haskell. [For example, `GeneralizedNewtypeDeriving` is not Safe Haskell compatible at the moment.](https://gitlab.haskell.org/ghc/ghc/-/issues/8827#note_115766)

goldfirere:
I generally agree that GHC2021 should be compatible with Safe Haskell. But what is Safe Haskell? It's an honest question: do we have an up-to-date (or reasonably so) spec?

My recollection around GND and SH is that there was discomfort around role inference. For example, if I write

```hs
module Set ( Set, unitSet, unionSet, elemSet ) where
newtype Set a = Set [a]   -- INVARIANT: no two elements are ==
unitSet :: a -> Set a
unionSet :: Eq a => Set a -> Set a -> Set a
elemSet :: Eq a => a -> Set a -> Bool
```

and do not include a role annotation, then a client could write

```hs
newtype Ignore = Mk Int
instance Eq Ignore where
  _ == _ = True
sneaky :: Set Ignore
sneaky = coerce (unitSet (1 :: Int) `unionSet` unitSet (2 :: Int))
```

to create a `Set` that violates its invariant. The solution is to include a role annotation on any abstract type that has an invariant that depends on a class-based operation:

```hs
type role Set nominal
```

With that role annotation, the `coerce` in the definition of `sneaky` would fail to type-check.

The problem vis-à-vis Safe Haskell is (if I recall correctly) that users might not know to write the role annotation. That is, before roles, an abstract type was simply one that did not export its constructors. With roles, an abstract type does not export its constructors *and* has a role annotation appropriate to any invariants. (Before roles, you could still break the invariant with `-XGeneralizedNewtypeDeriving`, which is why the extension has never been Safe.)

So, it all comes down to what we expect of users. Note that a very low expectation might say that users would not know how to make a type abstract via an export list; in this interpretation any module without an explicit export list might be considered Unsafe. A high expectation is that users know about role annotations, under which `-XGeneralizedNewtypeDeriving` should be Safe.

Maybe this suggests that we include `-XGeneralizedNewtypeDeriving` in GHC2021 only if we also include `-XRoleAnnotations` -- a decision I would be fine with.

Thanks for bringing this up, @phadej!

phadej:
> But what is Safe Haskell? It's an honest question: do we have an up-to-date (or reasonably so) spec?

No. It's indeed a problem. Is it so that Safe Haskell programs don't segfault (as they could if you use `unsafeCoerce`), if so, maybe that's what it should say.

If that is the definition, then your example is ok even without hiding the constructor. Yes, invariants will be broken and results might be complete non-sense (and one could run into non-termination), but I cannot write `unsafeCoerce` or something not referentially transparent.

https://simonmar.github.io/bib/papers/safe-haskell.pdf says 

> Though Haskell is predominantly type-safe, implementations contain a few loopholes through which code can bypass typing and
module encapsulation.

`unsafeCoerce` is bypassing typing, that is clear. Module encapsulation is defined as

> Haskell provides an effective module
system that is used to control access to functions and data
types. In the safe language these module boundaries are strictly
enforced; a user of a module that contains an abstract data type
is only able to access or create values through the functions
exported.

Here, I think, it's exactly on GHC Committees turf to decide whether GND as is Safe (Haskell), and programmers (of abstract data-types) just need to know of `coerce`, or whether changing default roles to `nominal` is the way to make GND and Safe Haskell compatible.

phadej:
Full Safe Haskell guarantees from the paper above

> Subject to this, Safe Haskell provides the following guarantees for code compiled in the safe language:
> 
> - Type safety. In Milner’s famous phrase, well-typed programs do not go wrong.
> - Referential transparency. Functions in the safe language must be deterministic. Evaluating them should not cause any side effects, but may result in non-termination or an exception.
> -  Module encapsulation. Haskell provides an effective module system that is used to control access to functions and data types. In the safe language these module boundaries are strictly enforced; a user of a module that contains an abstract data type is only able to access or create values through the functions exported.
> -  Modular reasoning. Adding a new import to a module should not change the meaning of existing code that doesn’t directly depend on the imported module.
> - Semantic consistency. Any valid Haskell expression that compiles both in the safe language and in the full language must have the same meaning.

goldfirere:
@phadej thanks for digging that out -- very helpful. So I think it becomes a judgment call: those bullet points do not define "abstract data type". We could say that a type without exported constructors is abstract, or we could also require a role annotation for a type to be considered abstract.

I'm not sure I have a good sense of what's better. I think it would be helpful to hear from users of Safe Haskell to see what they would find better.

Ericson2314:
FWIW I hope via something like https://github.com/ghc-proposals/ghc-proposals/pull/276 we can obviate this problem by not exporting any introduction or elimination rules for coercions for "abstract data types".

phadej:
I noticed that there are some extensions on the list which are enabled currently (e.g. `ForeignFunctionInterface`, `MonadFailDesugaring`, `NonIncreasingIndentation`, `CUSKs`). Is it possible that GHC2021 will disable them?

I think you cannot even disable `MonadFailDesugaring`, not sure about e.g. `CUSKs`.

Also `StarIsType` is currently enabled. I guess vote is to not disable it (i.e. `NoStarIsType`). That is inconsistent with having `NoMonomorphismRestriction` on the list.

For the references, snippet from `Session.hs`

```haskell
-- | The language extensions implied by the various language variants.
-- When updating this be sure to update the flag documentation in
-- @docs/users_guide/exts@.
languageExtensions :: Maybe Language -> [LangExt.Extension]

languageExtensions Nothing
    -- Nothing => the default case
    = LangExt.NondecreasingIndentation -- This has been on by default for some time
    : delete LangExt.DatatypeContexts  -- The Haskell' committee decided to
                                       -- remove datatype contexts from the
                                       -- language:
   -- http://www.haskell.org/pipermail/haskell-prime/2011-January/003335.html
      (languageExtensions (Just Haskell2010))

   -- NB: MonoPatBinds is no longer the default

languageExtensions (Just Haskell98)
    = [LangExt.ImplicitPrelude,
       -- See Note [When is StarIsType enabled]
       LangExt.StarIsType,
       LangExt.CUSKs,
       LangExt.MonomorphismRestriction,
       LangExt.NPlusKPatterns,
       LangExt.DatatypeContexts,
       LangExt.TraditionalRecordSyntax,
       LangExt.NondecreasingIndentation
           -- strictly speaking non-standard, but we always had this
           -- on implicitly before the option was added in 7.1, and
           -- turning it off breaks code, so we're keeping it on for
           -- backwards compatibility.  Cabal uses -XHaskell98 by
           -- default unless you specify another language.
      ]

languageExtensions (Just Haskell2010)
    = [LangExt.ImplicitPrelude,
       -- See Note [When is StarIsType enabled]
       LangExt.StarIsType,
       LangExt.CUSKs,
       LangExt.MonomorphismRestriction,
       LangExt.DatatypeContexts,
       LangExt.TraditionalRecordSyntax,
       LangExt.EmptyDataDecls,
       LangExt.ForeignFunctionInterface,
       LangExt.PatternGuards,
       LangExt.DoAndIfThenElse,
       LangExt.RelaxedPolyRec]
 ```

**Comment by phadej** [comment]
Also to corrent `Session.hs` (I'll make a MR for that)

```
           -- strictly speaking non-standard, but we always had this
           -- on implicitly before the option was added in 7.1, and
           -- turning it off breaks code, so we're keeping it on for
           -- backwards compatibility.  Cabal uses -XHaskell98 by
           -- default unless you specify another language.
```

is not be true with upcoming `Cabal-3.4`. If you don't specify `default-language`, used language will be whatever is the default for the compiler you use.

**Comment by nomeata** [comment]
Thanks for double checking!

> I noticed that there are some extensions on the list which are enabled currently (e.g. `ForeignFunctionInterface`, `MonadFailDesugaring`, `NonIncreasingIndentation`, `CUSKs`). Is it possible that GHC2021 will disable them?

Yes, this confusion was my fault: When preparing the ballot I didn’t think about what set of extensions the ballot should be based on. And that is not even a simple question, as GHC already has four different language sets (Haskell98, Haskell2010, the “default” mode, and the “default” mode with changes for GHCi). I wasn’t even aware of this complexity before 🤦

The ballot took the polaritiy of options from the GHC documentation, but that is itself inconsistent (e.g. it lists `NoImplicitPrelude` but `StarIsType`). Well, now it’s done and we’ll have to deal with it. Next round, the base should certainly be the previous GHC20xx, so then at least the question is clear.

That said, it _is_ possible that GHC2021 might disable some of them. It was argued (but not universally held) that it is beneficial that  uses of `ForeignFunctionInterface` are explicitly heralded at the top of the file. At the moment it looks like it will make it in.

`MonadFailDesugaring` currently only has 6 votes. I assume that most “no” votes were passed without full awareness that this would change the status quo (mine isn’t, I should correct that). I will bring that to the committe.

There was a brief discussion of `NonincreasingIndentation`. It seems is currently part of default and Haskell98, but not Haskell2010. Assuming it was intentionally removed in Haskell2010, it seems reasonable if it stays out in GHC2021. But it maybe worth digging up the justification why it is still on by default, despite Haskell2010 removing it. 

`CUSKs` is documented to be a legacy feature and supposed to be superceded by `StandaloneKindSignatures`; it seems that GHC2021 is headed towards including that switch. 

> is not be true with upcoming Cabal-3.4. If you don't specify default-language, used language will be whatever is the default for the compiler you use.

That’s an interesting fact. With the GHC20xx plan we are _hoping_ that we can evolved these extensions sets and defaults relatively liberally, assuming that most “properly packaged code” specifies a `default-language`. Will Cabal-3.4 at least warn or otherwise nudge people to specify one? Or maybe Hackage will enforce one?

**Comment by phadej** [comment]
> “properly packaged code”

Is a sensitive topic. I'd rather avoid it.

> Will Cabal-3.4 at least warn or otherwise nudge people to specify one? Or maybe Hackage will enforce one?

Clarification: omitting `default-language` would only be ok in `cabal-version: 3.4` cabal files.

Note for context. Cabal-3.4 was feature frozen in July 2020.
The issue about making `default-language` optional field is https://github.com/haskell/cabal/issues/6288

When that "decision" was made there weren't any talks about GHC20xx.

So to answer your question: No, Cabal-3.4.0.0's `cabal check` won't warn about `cabal-version: 3.4` packages without `default-language`.

That can be changed.

**Comment by nomeata** [comment]
Hmm, I don't think I can already say what's best here. Was there previously a perception that GHC's “default” mode was in some sense stable, and is that expectation broken with GHC20xx? 

Was the motivation for being able to use the default mode more from people who pin their GHC anyways, and wished to not fix the language set separately? Or what was the motivation?

(It seems odd to me to encourage developers to add version bounds or even pins to the libraries they want to be using, but not be explicit in which language they want to be using, but maybe I don't understand the full picture here)

**Comment by maralorn** [comment]
Just to throw in a thought here: I thought GHC20xx could be a chance for more stability. i.e. we could refrain from changing defaults (like i.e. MonadFailDesugaring, which broke quite a lot of projects) and projects would always opt-in to changing the language they use by specifying it in their cabal file. It would be easier to change defaults in a new GHC20xx, if existing code would never run into it without opting in.
That is quite exactly how rust editions work. (The biggest disadvantage I see with this is, that we could never drop deprecated features from the ghc code.)

Of course, if cabal would always default to the latest GHC20xx we loose that possibility. One slightly unsatisfying solution to get this behavior would be for ghc to default to the newest GHC20xx, when used alone or for ghci but cabal to always default to Haskell2010, if not specified. I don‘t know what the best solution is, but I would keep this in mind.

**Comment by phadej** [comment]
> Was there previously a perception that GHC's “default” mode was in some sense stable, and is that expectation broken with GHC20xx?

The default have not changed for a decade (sans how interpretation of Haskell2010 have changed) https://github.com/ghc/ghc/commit/18a0ead12b605fd20c13f5255d78cfbf699fe0c4

**Comment by phadej** [comment]
> Just to throw in a thought here: I thought GHC20xx could be a chance for more stability. i.e. we could refrain from changing defaults (like i.e. MonadFailDesugaring, which broke quite a lot of projects) 

That is not something GHC is engineered to support now. If `base` library provided `Monad` doesn't have `fail`, GHC really doesn't have much options how to desugar patterns. Similarly AMP (or SMP) proposals were not guarded by language extensions as these are library changes (though `Monad` is defined in report), though indeed broke some stuff, but OTOH would fragmented the ecosystem if there were choice to not define the new superclasses. The upcoming whitespace sensitivity change of `~`, `!` in 9.0 is another small, pragma-unguarded, yet forced change.

All of the changes I highlighted changed how GHC interprets both Haskell98 and Haskell2010. The https://downloads.haskell.org/~ghc/9.0.1-alpha1/docs/html/users_guide/bugs.html#haskell-standards-vs-glasgow-haskell-language-non-compliance section grows over time. I think it's safe to expect that GHC2021 interpertation will slightly change over time as well.

File: proposals/0000-ghc2021.rst [hunk 2]
```diff
@@ -11,13 +11,35 @@ GHC 2021
...
+* Types:  `GADTSyntax`_, `RankNTypes`_, `StarIsType`_, `TypeApplications`_, `PolyKinds`_, `StandaloneKindSignatures`_, `ExistentialQuantification`_, `TypeOperators`_, `ConstraintKinds`_
+* Type signatures: `ExplicitForAll`_, `KindSignatures`_, `NamedWildCards`_, `ScopedTypeVariables`_
+* Class and instances declarations: `FlexibleContexts`_, `FlexibleInstances`_, `MultiParamTypeClasses`_, `ConstrainedClassMethods`_, `InstanceSigs`_, `TypeSynonymInstances`_
+* Deriving mechanism: `DeriveDataTypeable`_, `DeriveFoldable`_, `DeriveFunctor`_, `DeriveTraversable`_, `StandaloneDeriving`_, `EmptyDataDeriving`_, `DeriveLift`_, `GeneralisedNewtypeDeriving`_, `DeriveGeneric`_
 
-* Everything in ``Haskell2010``, plus
-* TBD
+As a reminder, the following extensions are part of `Haskell2010`_, and continue to be enabled:
+
+* ``ImplicitPrelude``, ``StarIsType``, ``MonomorphismRestriction``, ``TraditionalRecordSyntax``, `EmptyDataDecls`_, `ForeignFunctionInterface`_, ``PatternGuards``, ``DoIfThenElse``,  ``RelaxedPolyRec``.
```
phadej, line 36: GHC's interpretation of Haskell2010 includes `MonadFailDesugaring`.

(Also `StarIsType` is in both lists: in to be included and this remainder).

nomeata: > GHC's interpretation of Haskell2010 includes `MonadFailDesugaring`.

Although not listed under https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/control.html?highlight=haskell2010#extension-Haskell2010. Even if it isn’t an technical change of status quo, I think it makes sense to treat it like a change to the specification. It’s all a bit squshiy when it comes to report vs. `base` issues…

Thanks for `StarIsType`, fixed.

phadej: It won't. It's then better to omit whole extension, as it will be removed from future GHCs. See https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4767

nomeata: If we _don’t_ list it in this section it looks as thought the committee voted against `MonadFailDesugaring`. I think this is fine, if at some point _flag_ goes away in GHC, effectively turning it on constantly, that’s in accordance with what GHC2021 says it should look like, doesn't it?

phadej: I argue that the flag is turned on for the GHC interpretation of Haskell2010, which is what I understand `Haskell2010` refers to. I think we should be very clear that GHC Haskell 2010 diverges from report *considerably*, and there is no point trying to hide that. 

https://ghc.gitlab.haskell.org/ghc/doc/users_guide/bugs.html#divergence-from-haskell-98-and-haskell-2010 is lengthy.

nomeata: I clarified in the proposal text; better now?

nomeata:
Hmm. It seems that it either means
 * we maintain the `*` for those who opt into `GHC2021` (Rather than `GHC202x` without `StarIsType`)
 * we remove support for `GHC2021` 
 * we break a bunch of code.

I think it’s the simply same options we have for how we want to treat those using `StarIsType` explicitly or via `Haskell2010+KindSignatures`.

goldfirere:
One possibility is to have `-XGHC2021` imply `-XNoStarIsType`, but to have the default be `-XGHC2021 -XStarIsType`. This way, people who explicitly opt into the new language standard are forced to write code that will remain valid after we drop support for `-XStarIsType`, but it keeps with the spirit of the status quo in keeping `-XStarIsType` on by default.

nomeata:
I think it was a mistake so far for GHC to, by default, have a selection of flags that doesn’t have a name… with that in mind, your suggestion translates to into having slightly different `GHC2021`  and `GHC2021Legacy`, the latter on by default, and with `StarIsType` (and `NondecreasingIndentation`?) on.

I am not quite sure I like this extra complexity… can’t we really have a single base line language that works both as “this is what people should use” and “this is what people get to use by default”?

goldfirere:
Some discussion this morning helped me to realize that we are not expecting rigid stability in the `-XGHC2021` language, and so the change around `*` in the future does not necessarily pose a problem. Happy to drop this point.

monoidal:
I'm fine with deciding that we might ultimately break code using * under GHC2021, though I would put this fact directly in the proposal.

Note that GHC 9.0 will already issue a warning on * with -Wall.

PR #378, opened 2020-11-14
Support the design for dependent types (under review)

goldfirere:
This proposal describes the tension between two important criteria the committee (and broader community) have implicitly been using to evaluate new features. In several proposals (#270, #281, and #291), these criteria -- essentially, cohesiveness of the language today vs. cohesiveness of a hypothetical future with dependent types -- appear to be at odds, with no mutually pleasing design. The proposal payload is to explicitly acknowledge the importance of designing with ergonomic dependent types in mind, and accepting this proposal (or rejecting it) will hopefully clarify our deliberations on other proposals.

Put more simply: Support this proposal if you think GHC should design with an eye toward adopting ergonomic dependent types; argue against this proposal if you think GHC should focus more on maintaining cohesion in the short term, possibly closing off future avenues toward ergonomic dependent types.

[Rendered](https://github.com/goldfirere/ghc-proposals/blob/dependent-types/proposals/0000-dependent-type-design.rst)

Issue #237, opened 2019-06-01
Named functional dependencies


int-index:
In https://github.com/ghc-proposals/ghc-proposals/pull/236#issuecomment-497959410, @cartazio brings up the topic of generating function inverses from functional dependencies. I think it should be possible by giving them a name.

Today we can write:

```
type family F a = r | r -> a where
  F True = 1
  F False = 0
```

I propose we write this instead:

```
type family F a = r | G r = a where
  F True = 1
  F False = 0
```

Then GHC would derive the following definition:

```
type family G r = a where
  G 1 = True
  G 0 = False
```

int-index:
Similarly, for classes, instead of:

```
class MonadReader r m | m -> r

instance MonadReader AppEnv AppM
instance MonadReader r m => MonadReader r (StateT s m)
```

One could write:

```
class MonadReader r m | Env m = r
instance MonadReader AppEnv AppM
instance MonadReader r m => MonadReader r (StateT s m)
```

And GHC would derive:

```
class (r ~ Env m) => MonadReader r m where
  type Env m

instance MonadReader AppEnv AppM where
  type Env AppM = AppEnv

instance MonadReader r m => MonadReader r (StateT s m) where
  type Env (StateT s m) = Env m
```

That's an associated type family for free!

Ericson2314:
Woah! So this was a neat trick in general, basically making sure we don't waste the opportunity to have a "propositional *x* for every judgmental *x*". But that we could re-use the same syntax to desugar functional dependencies in general is *really* cool. The idea that we'll end up with just functions (including associated familes), equality constraints, and some basic guarantees that GHC can use constraints in tactic-like ways is very cool!

cartazio:
+1, thanks for distilling my fuzzy idea a tad.  This seems like it would
give a mechanism for augmenting injective type families too, right ?
Currently we only have unary ones. Or at least mechanizing part of such a
desugaring

One thing I’m not up to date on is that for the past few years ghc has has
the constraint/coercion’s in core blowup.  Would this interact with that?
 (To be clear, that’s not an obstacle, just preexisting ghc debt,  I think
Ben merged in a don’t keep coercion terms unless core lint is enabled
change in recent history too )

Ericson2314:
I think this would also "force" us to solve the untracked fun dep orphans problem; the fun dep families would have heads that are blatantly orphans by the current approach, so we would catch them for free.

cartazio:
[...]

PR #381, opened 2020-11-27
Generalised type family injectivity

cgibbard:
About 4 years ago, there was a push to generalise type family injectivity constraints, to make them more like the functional dependencies that type classes enjoy. We're interested in reviving this effort.

There's a GHC issue about it [here](https://gitlab.haskell.org/ghc/ghc/-/issues/10832), as well as a fairly detailed [wiki page](https://gitlab.haskell.org/ghc/ghc/-/wikis/injective-type-families#user-content-type-c-injectivity-aka-generalized-injectivity).

There are some details or finer points left to be worked out, but hopefully we can get a discussion going.

### DIFF
--- a/proposals/0000-generalised-type-family-injectivity.rst
+++ b/proposals/0000-generalised-type-family-injectivity.rst
@@ -144,7 +144,7 @@ With regard to learnability, I think this removes an inconsistency given the sim
 functional dependencies might not expect the restriction on injectivity specifications for type families. That said, it enables clever
 applications of these constraints to discover the instantiations of type variables from type family instances in a way that might be
 tricky to do in one's head. Haskell users will need to exercise some care that they're not making life any harder for others who read their 
-code. (This concern could already be levelled against functional dependencies in general though.)
+code. (This concern could already be levelled against tricky uses of functional dependencies in general though.)
 
 Alternatives
 ------------
--- a/proposals/0000-generalised-type-family-injectivity.rst
+++ b/proposals/0000-generalised-type-family-injectivity.rst
@@ -18,6 +18,10 @@ Type families presently enjoy a very restricted form of functional dependency / 
 namely, one is only allowed to specify that the result of the type family uniquely determines some of its
 arguments. We propose to extend this to allow arbitrary functional dependencies.
 
+There was an attempt to implement this 4 years ago, and we're interested in reviving the effort. There's
+`a GHC wiki page <https://gitlab.haskell.org/ghc/ghc/-/wikis/injective-type-families#user-content-type-c-injectivity-aka-generalized-injectivity>`_
+as well as an existing `GHC issue (#10832)<https://gitlab.haskell.org/ghc/ghc/-/issues/10832>`_ regarding
+the aim of this proposal.
 
 Motivation
 ----------
@@ -161,6 +165,44 @@ The explanation provided is informal and might be somewhat open to interpretatio
 as possible to class functional dependencies.
 
 In addition to that, formally specifying the change to unification might be helpful.
+
+Consider the type family and instance::
+
+  type family F a b = c | a -> b
+  type instance F Int Char = Bool
+
+Given::
+
+  x :: F Int b
+
+we know that there can be no other instance than the one for F Int Char, so from this, we might deduce that b ~ Char and x :: Bool.
+However, an interesting thing about this is that a similar-looking signature::
+
+  y :: forall b. F Int b
+
+must surely be invalid, as no other b will do. There are some questions about when exactly these constraints are generated and how to apply them in all cases.
+
+Consider as well::
+
+  z :: forall b. F Bool b
+
+where there is no instance matching F Bool b, but the functional dependency seems to indicate that at most one type b would do.
+Presumably if the signature for y is invalid, then so should be the signature for z, so perhaps looking at instances is irrelevant, and only the functional
+dependency itself restricts our ability to quantify the variable.
+
+Existential quantification may be fine though::
+
+  data SomeF a = forall b. MkSomeF (F a b)
+  
+  e :: SomeF Int
+
+and we can presumably recover the quantified variable when pattern matching::
+
+  case e of
+    MkSomeF (u :: F Int b) -> {- b ~ Char here because of the functional dependency on F -}
+      Refl :: (b :~: Char)
+
+While this looks sensible, we're not entirely sure how to specify things in a way that makes it entirely clear what is valid.
 
 Implementation Plan
 -------------------
--- a/proposals/0000-generalised-type-family-injectivity.rst
+++ b/proposals/0000-generalised-type-family-injectivity.rst
@@ -1,0 +1,172 @@
+Generalised Type Family Injectivity
+===================================
+
+.. author:: Cale Gibbard and Dan Bornside on behalf of Obsidian Systems on behalf of MIRI
+.. date-accepted:: Leave blank. This will be filled in when the proposal is accepted.
+.. ticket-url:: Leave blank. This will eventually be filled with the
+                ticket URL which will track the progress of the
+                implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. contents::
+
+Type families presently enjoy a very restricted form of functional dependency / injectivity constraint,
+namely, one is only allowed to specify that the result of the type family uniquely determines some of its
+arguments. We propose to extend this to allow arbitrary functional dependencies.
+
+
+Motivation
+----------
+
+Consider the following small example with sized-vectors, often one of the first examples of dependently-typed programming::
+
+  {-# LANGUAGE TypeFamilies #-}
+  {-# LANGUAGE DataKinds #-}
+  {-# LANGUAGE GADTs #-}
+  {-# OPTIONS_GHC -Wall -Werror -Wno-unticked-promoted-constructors #-}
+
+  data Nat = Zero | Succ Nat
+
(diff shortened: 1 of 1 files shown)


AntC2:
This is a hard problem to solve in general, so good luck! Maybe the approach should be to look at examples where there's a clear need and easily-understandable rules for inference, and tackle only those.

* That ticket has languished for years, with some of the difficulties.
* GHC's FunDeps, as soon as you liberalise some of the rules from the '[FunDeps via CHRs](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/jfp06.pdf)' paper, turns into a hot mess: there's heaps of tickets and surprises and quite a bit of debate over whether certain programs should be accepted, let alone what they should do.

The difficulty in context of TFs is:

* The TF rules stipulate each LHS of the TF equations must be apart. But when you read the equation 'from right to left', the RHSs are not necessarily apart.
* Or for a Closed TF, any lack of apartness in the LHSs is resolved top-to-bottom, whereas reading 'right to left', the top-to-bottom priority might be different. Take your example

        type family Add (n :: Nat) (m :: Nat) :: Nat where
          Add Zero b = b
          Add (Succ a) b = Succ (Add a b)

Reading right to left for FunDep `r m -> n`, GHC can't tell that `r ~ b` is apart from `r ~ Succ (Add a b)`.

Even for FunDep `r n -> m`, where the `n`s are apart in the two equations, it needs to solve `r ~ Succ (Add a b)` to figure out `b`.

Ericson2314:
@AntC2 I should first make clear that this isn't a normative part of the proposal at this time, but we are very aware of the discssionn in #237.

1. The current implementation is sadly unrelated to class fun deps, which is also unrelated to type families. It would be nice to deduplicate all this so we don't get out of sync functionality in the first place!
2. If we have trouble with GHC figuring out the inverse, it would be nice to allow humans to provide it. Then GHC can just worry about proving the equalities and implications. Maybe we can even allow uses to write the proof too.

I think #237 could help with both of these, if approaching this problem as has been done in the past proves too difficult or complicated.

AntC2:
Thanks @Ericson2314, it's difficult to comment further, because the proposal currently doesn't contain a detailed change spec. It's more of a wish-list. Yes it'll need to allow the TF head to give arbitrary dependency arrows, in the style of FunDeps. But we know that's not going to be sufficient to solve type improvement without specifying something about the form of instances. Some points I notice:

> There is also probably some sensible way in which it ought to interact with associated type families.

I guess it's going to be easier to see the type improvements for a TF if it's Closed, with all the equations/instances bunched together and no opportunity to add further instances in another module. Then the TF couldn't be Associated.

> Perhaps functional dependencies on the type class should be translated to constraints on any associated type families?

I think it's easier to lean on the work already done to translate FunDeps into superclass constraints with TF calls and `~` equalities. That would be more 'function-y', and in line with #237. It'd only be ergonomic if the compiler generated the instances for those other TFs. Anyhoo, the proposal as stands is hand-wavey.

> There's some degree of potential to reduce future maintenance costs by sharing machinery between class and type family functional dependencies, 

Is a really bad idea: the code for Functional Dependencies is "bogus" [SPJ on Trac #10675], tries to support overlapping instances, by accident supports [dysfunctional dependencies](https://gitlab.haskell.org/ghc/ghc/-/issues/18759), [risks non-confluence](https://gitlab.haskell.org/ghc/ghc/-/issues/18851), undecidability, ... Your examples " the following combinations of instances are invalid" can all be accepted as class instances by GHC, given sufficient ingenuity/malice in reformulating them.

So re @Ericson2314's "The current implementation is sadly unrelated to class fun deps,": that's a Good Thing, nothing to be sad about.

A better strategy would be to approach injectivity requirements with a clean design, then see if the learnings could be applied to FunDeps.

AntC2:
One way (credit to Oleg) to get multi-way injectivity with typeclasses for your Add example

    class (Add' n m r, Add' m n r) 
          => Add (n :: Nat) (m :: Nat) (r :: Nat)  | n m -> r, n r -> m, m r -> n     -- three FunDeps
    instance (Add' n m r, Add' m n r)
          => Add n m r 
    
    class Add' (n :: Nat) (m :: Nat) (r :: Nat) | n m -> r, n r -> m                  -- two FunDeps
    
    instance (m ~ r) => Add' Zero m r
    
    instance ((Succ r') ~ r, Add' n' m r') => Add' (Succ n') m r 

Class `Add` is the wrapper, `Add'` is the worker providing injectivity only one way `n r -> m`. `Add`'s superclass constraints use two `Add'`s, with the params flipped.

Note that `Add'`'s instances recursively descend on the `n` parameter only, with bare distinct tyvars for the other parameters, to ensure instance selection is as eager as possible. All of the type improvement is in the instance constraints, even for the `Zero` case.

goldfirere:
I think having this would be great. But I sadly agree that the current proposal is really more of a wish-list than a proper proposal. It specifies the syntax nicely, but is too quiet on the semantics. The proposal knows this shortcoming, but it would be hard to accept (or really evaluate) the proposal without more info.

In particular, see [Section 4.1 of the Injective Type Families](https://richarde.dev/papers/2015/injective/injective-type-families.pdf#%5B%7B%22num%22%3A37%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22Fit%22%7D%5D) paper. The fact that these cases are awkward was a great surprise to us in writing that paper. I have to imagine similar surprises will come up with generalized injectivity, and we cannot implement a proposal such as this without figuring out what those cases are.

The implementation will not be difficult here -- but the specification is a real challenge.

cgibbard:
@goldfirere We've been thinking about some of the awkward cases in the paper this morning. For awkward case 2 specifically, it appears the two examples are fundamentally different. The type family W1:

    type family W1 a = r | r -> a
    type family W1 [a] = a

is injective on the set of types for which it is actually defined, while for W2, well, perhaps it's really injective because it is in actuality the empty function, but there is no way that it could ever be defined anywhere and remain injective.

    type family W2 a = r | r -> a
    type instance W2 [a] = W2 a

If we assume that W2 [a] is defined, then we would have W2 [[a]] = W2 [a] = W2 a which would contradict injectivity immediately. So this instance for W2 in some way causes an actual violation of the injectivity constraint.

Let's think some more about W1 though: the form of injectivity that we're deriving for the instance looks like:

    W1 [a] ~ W1 [b] => a ~ b

but really, we should only conclude this if W1 [a] and W1 [b] are both defined. As the example in the paper shows, we have:

    W1 [W1 Int] ~ W1 Int

per the beta reduction of the instance itself, but if we're allowed to apply injectivity, we'll have

    [W1 Int] ~ Int

The problem here is that we're dealing with functions whose domain hasn't been correctly modelled and trying to conclude things using injectivity for points that don't actually lie in the domain of the function. What if we had a new constraint predicate `defined` that could be applied to a type-level expression involving type families, and which could only be satisfied once no type family applications or type variables occur in it? (This latter restriction because type variables might be instantiated to type family applicatons.) Then the injectivity law we would infer for the W1 instance would be:

    (defined (W1 [a]), defined (W1 [b]), W1 [a] ~ W1 [b]) => a ~ b

We could immediately evaluate the W1 [a] and W1 [b] occurring under the defined predicates to obtain:

    (defined a, defined b, W1 [a] ~ W1 [b]) => a ~ b

So now if we apply this with a |-> W1 Int and b |-> Int, we obtain:

    (defined (W1 Int), defined Int, W1 [W1 Int] ~ W1 Int) => W1 Int ~ Int

which we can reduce to:

    (defined (W1 Int)) => W1 Int ~ Int

But W1 Int is a stuck type, so the `defined (W1 Int)` constraint will never go away. Moreover, we could potentially observe that any instance defining W1 Int would violate the functional dependency, but that's beyond what is required to prevent this source of potential unsoundness.

While there's still much more to think about, this may be at least one missing ingredient. What do you think?

goldfirere:
I think you may have reinvented [Constrained Type Families](https://richarde.dev/papers/2017/partiality/partiality.pdf). I believe this all gets much simpler with constrained type families. However, I don't think we can implement that properly without dependent types, due to some unfortunate interactions with data families (if memory serves).

cgibbard:
@goldfirere It would be good to hear more about any unfortunate interactions here, if you can remember what they were.

The `defined` predicate I'm thinking of above seems like it would be easy enough to implement, as the rule for its satisfaction is practically a syntactic check (after type family reduction and unification have done their work).

The real question is though: Is knowledge about the domains on which type families are actually defined the only obstruction to soundness here?

I think if that is so, then introduction of a new sort of constraint of some kind is the right approach to handling the fact that we will often need to defer decisions about it. (While closed type classes would *also* be great, and closed type families are as well, there are enough cases where the openness is rather essential to an application.)

It might be worth mentioning that the way I came to this idea is by trying to consider what the semantics of an open type family might look like. If we think of it as a mathematical function, the issue that seems to be at the heart of some of these examples of unsoundness is that the type family is being applied to points that are ultimately not in the domain of that set-theoretical function. Open type families (or open functions more generally) are functions A -> B for which questions of the form "Is t in A?" can remain unanswerable until some future module.

There's also the question on the other side, of finding an algorithm to check that the injectivity constraints are respected by type family instances in the first place. I'd kind of hope that the remainder of the problem was similar enough to the work that's already been done for functional dependencies, but I'm somewhat wary of this as well. I haven't thought enough about it yet.

Any further known hard examples or obstructions of the sort in the Injective Type Families paper would be appreciated from anyone.

cgibbard:
In a discussion I was having with co-workers, it was pointed out that my `defined` predicate is essentially the same thing as `Typeable`, only with no requirement for runtime evidence.

goldfirere:
>  it was pointed out that my `defined` predicate is essentially the same thing as `Typeable`, only with no requirement for runtime evidence.

I agree completely, and had not thought of that previously. Interesting.

Ericson2314:
Taking a step back a bit, and assuming #378 is accepted or we are otherwise on the road to Depedent Haskell, there is a question of whether to add further functionality to type families as they exist, or work on their replacement. I've written #383 to try to kick off sketching out how to get the relevant DH goodies sooner, in the hopes we might end up two shorter-term roadmaps to decide between, rather than the DH alternative meaning "wait 3 years".

PR #382, opened 2020-11-29
GHCi Clear Command proposal

cateroxl:
Hello! 

Adding `:clear [variable identifier]` that only applies to variable identifiers or functions defined in GHCi. Here's the link to my [proposal](https://github.com/cateroxl/ghc-proposals/blob/patch-1/proposals/ghci-clear-command.md).

### DIFF
--- a/proposals/ghci-clear-command.md
+++ b/proposals/ghci-clear-command.md
@@ -1,0 +1,88 @@
+---
+author: Cate Roxl
+date-accepted: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/0>).
+**After creating the pull request, edit this file again, update the number in
+the link, and delete this bold sentence.**
+
+# Make learning and playing with Haskell through GHCi easier through adding `:clear [variable identifier]`.  
+
+Here you should write a short abstract motivating and briefly summarizing the
+proposed change.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


proposals/ghci-clear-command.md
```diff
@@ -0,0 +1,86 @@
...
+
+```
+Prelude> rev x = reverse x
+Prelude> rev "Haskell"
+"lleksaH"
+```
+
+Now there are other ways to write such function, which we can play and study with. But because of immutability, once we declare `rev = reverse x` or any variable identifier to a variable: it stays that way. 
+
+Unless. . . We quit GHCi and start it up again.
```
nomeata, line 34: Does it really stay that way? GHCi allows you to shadow earlier bindings:
```
Prelude> foo x = x
Prelude> foo ()
()
Prelude> foo x = x + 1
Prelude> foo 1
2
```

Of course it can still be confusing that the old function is still around:
```
Prelude> foo x = x + 1
Prelude> foo 1
2
Prelude> bar x = foo x
Prelude> bar 1
2
Prelude> foo x = x + 2
Prelude> foo 1
3
Prelude> bar 1
2
```
but I that is, even maybe initially confusing, explainable.



```

cateroxl: Oh! I tried it and it works. Thank you for this. There are statements in my proposal that are wrong. It doesn't stay that way as I thought. It may just be the books and tutorials that I've come across when they discuss immutability. It's my mistake for not considering that the GHCi allows changes in definition. I've always thought of it being immutable right there in the GHCi.

cateroxl: I'm closing the pull request. Thank you once again.

PR #386, opened 2020-12-08
Simplify type synonym and type instance scoping

simonpj:
New proposal to simplify type synonym and type instance declarations

[rendered](https://github.com/ghc-proposals/ghc-proposals/blob/wip/spj-instance-scoping/proposals/0000-instance-scoping.rst)

### DIFF
--- a/proposals/0000-instance-scoping.rst
+++ b/proposals/0000-instance-scoping.rst
@@ -39,7 +39,7 @@ Proposed Change Specification
   (Implementation note: see ``Note [Implicit quantification in type synonyms]`` in ``GHC.Rename.HsType``.)
 
 * Change #2.   Introduce the following rule: in a type family instance declaration,
-  the instantiation of the family instance fully determined by the left hand side, without
+  the instantiation of the family instance is fully determined by the left hand side, without
   looking at the right hand side.
 
 Examples and motivation
--- a/proposals/0000-instance-scoping.rst
+++ b/proposals/0000-instance-scoping.rst
@@ -1,0 +1,190 @@
+Notes on reStructuredText - delete this section before submitting
+==================================================================
+
+The proposals are submitted in reStructuredText format.  To get inline code, enclose text in double backticks, ``like this``.  To get block code, use a double colon and indent by at least one space
+
+::
+
+ like this
+ and
+
+ this too
+
+To get hyperlinks, use backticks, angle brackets, and an underscore `like this <http://www.haskell.org/>`_.
+
+
+Proposal title
+==============
+
+.. author:: Simon Peyton Jones
+.. date-accepted:: Leave blank. This will be filled in when the proposal is accepted.
+.. ticket-url:: Leave blank. This will eventually be filled with the
+                ticket URL which will track the progress of the
+                implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. contents::
+
+In a couple of places GHC is being too clever when dealing with type synonym
+or type-family instance declarations.  This proposal suggests two
+related simplifications, that make the language easier to read, less ad hoc,
+and simpler to implement.
+
+It depends on proposal #326 (Inivisible parameters for declarations).
+This proposal has a draft implementation in MR !3145.
+
+Proposed Change Specification
+-----------------------------
+
+1. Introduce the following rule: in
+   * a type synonym declaration, or
+   * a type family instance declaration
+   every type variable mentioned on the RHS must be bound on the LHS.
+
+   To match this change, in the user manual, remove the text in
+   `Implicit quantification in type synonyms and type family instances
+   <https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/poly_kinds.html>`_.
+   from the beginning down to
+   "Kind variables can also be quantified in visible positions...".
+
+   (Implementation note: see ``Note [Implicit quantification in type synonyms]`` in ``GHC.Rename.HsType``.)
+
+2. Introduce the following rule: in a type family instance declaration,
+   the instantiation of the left hand side is fully determined, without
+   looking at the right hand side.
+
+Examples and motivation
+-----------------------
+
+Change #1
+~~~~~~~~~
+
+Consider::
+
+  type T1 = 'Nothing :: Maybe a
+  type T2 = 'Just ('Nothing :: Maybe a)
+  
+``T1`` is currently legal, yielding ``T :: forall a. Maybe a``. The general rule is that the
+free variables of a *top-level* kind signature on the RHS are brought into scope
+implicitly, and will be quantified in the final kind of the type constructor.
+
+So ``T2`` is currently illegal, because the kind signature is not at the top level.
+
+With this proposal, both declarations woudl be illegal.  Instead you must write:::
+
+  type T1 @a = 'Nothing :: Maybe a
+  type T2 @a = 'Just ('Nothing :: Maybe a)
+
+so that all the variables occurring on the RHS are bound on the LHS.
+
+
+Change #2
+~~~~~~~~~
+
+Consider::
+
+  type family F a :: k
+
+  type instance F Int = Char
+  type instance F Int = Maybe
+
+From the family declaration we see that ``F :: forall k. Type -> k`.
+The two ``type instance`` declarations appear to have an identical head, but by
+looking at the RHS we can infer that the invisible kind argument of ``F`` is
+``Type`` in the first instance, and ``Type -> Type`` in the second.  It would
+be much clearer to write::
+
+  type instance F @Type         Int = Char
+  type instance F @(Type->Type) Int = Maybe
+
+and indeed this is already legal.
+
+This proposal requires that the type instance be fully determined by the LHS,
+so that the programmer sees two visibly distinct instance heads.  For the purpose
+of determining the LHS, the RHS is ignored.  So under this proposal the
+instance::
+
+   type instance F Int = Char
+
+would mean::
+
+   type instance F @k Int = Char
+
+(where the LHS instantation is at an unconstrained kind ``k``).
+Now the kind of the RHS if fixed to be ``k``, and the kind of ``Char`` does
+not match that, so the declaration is rejected.
+
+The principle is that it should be possible to see what instance the
+programmer intended by looking only at the instance head (the LHS).
+This property already holds for *data* family instances.  Suppose
+``D`` is a data family of kind::
+
+   D :: forall k. (k->Type) -> k -> Type
+
+Now consider ::
+
+   data instance D p q where
+      MkD :: forall r. r Int -> T r Int
+
+So what kind do ``p`` and ``q`` have?  No clues from the header, but from
+the data constructor we can clearly see that ``r :: Type->Type``.  Does
+that mean that the the *entire data instance* is instantiated at ``Type``
+like this::
+
+   data instance D @Type (p :: Type->Type) (q :: Type) where
+      ...
+
+Or does it mean that the GADT data constructor specialises that kind argument,
+thus::
+
+   data instance D @k (p :: k->Type) (q :: k) where
+     MkD :: forall (r :: Type -> Type).
+            r Int -> T @Type r Int
+
+(It might be specialised differently in some other data constructor ``MkD2``).
+GHC avoids this question by determining the instance header solely from the
+header.  This proposal simply extends the same principle to type family instances.
+
+Effect and Interactions
+-----------------------
+
+These changes will make fewer programs compile.
+
+* For change #1 the approved new programming style requires proposal #326,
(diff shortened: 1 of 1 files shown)


int-index:
Rendered: https://github.com/ghc-proposals/ghc-proposals/blob/wip/spj-instance-scoping/proposals/0000-instance-scoping.rst

int-index:
On one hand, this change is very reasonable and I'd like to see it accepted.

On the other hand, it introduces an inconsistency with terms, where this continues to be accepted:

```
x = const :: a -> Bool -> a
```

I would, by analogy, expect this to be accepted:

```
type X = Const :: a -> Bool -> a
```

But with this proposal it will be rejected since `a` is not bound on the LHS. Well then, the term-level `x` should be rejected by the same reasoning.

proposals/0000-instance-scoping.rst
```diff
@@ -0,0 +1,190 @@
...
+  To match this change, in the user manual, remove the text in
+  `Implicit quantification in type synonyms and type family instances (6.4.11.12)
+  <https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/poly_kinds.html>`_,
+  from the beginning of that section down to
+  "Kind variables can also be quantified in visible positions...".
+
+  (Implementation note: see ``Note [Implicit quantification in type synonyms]`` in ``GHC.Rename.HsType``.)
+
+* Change #2.   Introduce the following rule: in a type family instance declaration,
+  the instantiation of the family instance fully determined by the left hand side, without
```
JakobBruenker, line 42:
```suggestion
  the instantiation of the family instance is fully determined by the left hand side, without
```

proposals/0000-instance-scoping.rst
```diff
@@ -0,0 +1,190 @@
...
+so that all the variables occurring on the RHS are bound on the LHS.
+This proposal exploits #326 to allow a nice, simple, uniform rule.
+As the manual says, "The reason for this exception [the strange, ad-hoc,
+top-level-kind-signature rule] is that there may be no other way to bind k".
+
+The same rule applies to type family instances::
+
+   type family F a :: k
+
+   type instance F Int = Any :: k -> k
```
JakobBruenker, line 76: To be clear, the `k` in `k -> k` is different from the `k` in `type family F a :: k`, right? Perhaps it'd be a good idea to rename one of these `k`s to make that more obvious

proposals/0000-instance-scoping.rst
```diff
@@ -0,0 +1,190 @@
...
+
+Consider::
+
+  type family F a :: k
+
+  type instance F Int = Char
+  type instance F Int = Maybe
+
+From the family declaration we see that ``F :: forall k. Type -> k`.
+The two ``type instance`` declarations appear to have an identical head, but by
```
JakobBruenker, line 96: The formatting here thinks that `The two` is part of a code block; I'm not quite sure how one would fix that in `.rst`

JakobBruenker:
I ran into something that seems relevant a couple weeks ago, so I'm curious how it would interact with this example:

```haskell
data Relation n m = MkR

type Trans :: forall a b c . Relation a b -> Relation b c -> Relation a c
type family Trans pa pb where
  Trans rel MkR = rel -- this type checks but is a partial type family
  -- Trans @a @b @c rel MkR = rel -- this would not type check
  -- Trans @a @b @b rel MkR = rel -- this would type check
```

Am I understanding correctly that the equation that's not commented out here would not type check under this proposal? If so, I'd very much appreciate that change, since I found it extremely confusing when a type family that looked total judging by the lhss turned out not to be.

simonpj:
> On the other hand, it introduces an inconsistency with terms, where this continues to be accepted

It doesn't *introduce* such an inconsistency, it merely continues it. 

When you write `(e :: a -> a)` in a term, if `a` isn't already in scope, the forall-or-nothing rule adds foralls, thus `e :: forall a. a->a`.

But that doesn't happen after the `::` in kind signatures.  Consider
```
f :: forall a. Proxy (a :: k->k) -> Int
```
You might think that would mean
```
forall a. Proxy (a :: forall k. k->k) -> Int
```
but it doesn't.  The outer `::` wins, and we get
```
f :: forall k. forall (a :: k->k). Proxy a -> Int
```
We might consider changing that but I didn't think this was the place to do so.

proposals/0000-instance-scoping.rst
simonpj: done

simonpj:
> Am I understanding correctly that the equation that's not commented out here would not type check under this proposal?

Yes you are!  I've taken the liberty of adding your example to the proposal.

Ericson2314:
I do like seeing these things discussed :).

I was thinking we'd get there via #326 + https://github.com/ghc-proposals/ghc-proposals/pull/285 's `NoPatternSignatureBinds`, which would restore the consistency @int-index worries about by prohibit the implicit binding in terms and types alike.

One of the interesting things to think about is that:
```haskell
t2 = Just (Nothing :: Maybe a)
```
works not by immediately turning into
```haskell
t2 @a = Just (Nothing :: Maybe a)
```
but by becoming
```haskell
t2 = Just (Nothing :: forall a. Maybe a)
```
and then type inference instantiating that.

proposals/0000-instance-scoping.rst
```diff
@@ -0,0 +1,225 @@
...
+
+Summary
+---------
+
+In a couple of places GHC is being too clever when dealing with type synonym
+or type-family instance declarations.  This proposal suggests two
+related simplifications, that make the language easier to read, less ad hoc,
+and simpler to implement.
+
+It depends on proposal #326 (Inivisible parameters for declarations).
```
RyanGlScott, line 21:
```suggestion
It depends on proposal #326 (Invisible parameters for declarations).
```

proposals/0000-instance-scoping.rst
```diff
@@ -0,0 +1,225 @@
...
+
+Here's an example from the wild (thanks Jakob Bruenker)::
+
+  data Relation n m = MkR
+
+  type Trans :: forall a b c . Relation a b -> Relation b c -> Relation a c
+  type family Trans pa pb where
+    Trans rel MkR = rel -- this type checks but is a partial type family
+
+In current GHC this typechecks, but the type family is not total?  Why?
```
RyanGlScott, line 163:
```suggestion
In current GHC this typechecks, but the type family is not total.  Why?
```

proposals/0000-instance-scoping.rst
```diff
@@ -0,0 +1,225 @@
...
+    Trans @a @b @b rel MkR = rel
+
+Notice the repeated ``b`` on the LHS. The author was entirely unaware
+that the resulting type family was partial, because the equation he wrote
+looked total.  With Change #2, the original program::
+
+    type family Trans pa pb where
+      Trans rel MkR = rel
+
+would be rejected. WHy? Because the LHS imposes no kind constraints, so
```
RyanGlScott, line 176:
```suggestion
would be rejected. Why? Because the LHS imposes no kind constraints, so
```

proposals/0000-instance-scoping.rst
```diff
@@ -0,0 +1,225 @@
...
+
+These changes will make fewer programs compile.
+
+* For change #1 the approved new programming style requires proposal #326,
+  and there is no backward compatible workaround.  So the phase-in will
+  need to be planned.
+
+* For change #2 there is a backward-compatible workaround, so we could
+  perhaps bring it in immediately.  It would be somwhat tricky to implement
+  a deprecation cycle, beucause we'd have to figure out whether the instantiaon
```
RyanGlScott, line 198:
```suggestion
  a deprecation cycle, because we'd have to figure out whether the instantiation
```

proposals/0000-instance-scoping.rst
```diff
@@ -0,0 +1,225 @@
...
+  was driven by the RHS
```
RyanGlScott, line 199:
```suggestion
  was driven by the RHS.
```

proposals/0000-instance-scoping.rst
```diff
@@ -0,0 +1,225 @@
...
+-----------------------
+
+These changes will make fewer programs compile.
+
+* For change #1 the approved new programming style requires proposal #326,
+  and there is no backward compatible workaround.  So the phase-in will
+  need to be planned.
+
+* For change #2 there is a backward-compatible workaround, so we could
+  perhaps bring it in immediately.  It would be somwhat tricky to implement
```
RyanGlScott, line 197:
```suggestion
  perhaps bring it in immediately.  It would be somewhat tricky to implement
```

RyanGlScott:
Besides some inline comments, there's a typo in the title of this PR itself (s/syononym/synonym/).

goldfirere:
I'm in support. This is a nice simplification, enabled by the introduction of new syntax.

simonpj:
No comments for a month. @nomeata I'd like to submit this to the committee.

nomeata:
Hmm, I am confused. The proposal says “It depends on proposal #326 (Invisible parameters for declarations).” but #326 is still in draft state. Seems to be too early to submit #386 then?

Also, since #386 fixes problems that can’t be fixed without #326, if I am not mistaken, would it make sense to just make them part of #326?

simonpj:
> Also, since #386 fixes problems that can’t be fixed without #326, if I am not mistaken, would it make sense to just make them part of #326?

I'd forgotten that.  And John has dropped #326 for lack of time (NOT because it is controversial or unimportant).  Bother maybe someone can pick up #326.  Maybe even me.  Anyone else?  

I'd be happy to see them combined.

RyanGlScott:
While modifying some code in GHC related to this proposal, I realized that [Change #1](https://github.com/ghc-proposals/ghc-proposals/blob/26c075a093dfaaab6c3aa706d7cf1f3163b6bd01/proposals/0000-instance-scoping.rst#id4) in this proposal doesn't really say what should happen to data family instances. For example, would this be illegal?

```hs
data family D (a :: k)
data instance D :: Maybe a -> Type
```

That is, you'd instead be expected to write this?

```hs
data instance D @(Maybe a) :: Maybe a -> Type
```

I think that's the case, based on my understanding of [Change #1](https://github.com/ghc-proposals/ghc-proposals/blob/26c075a093dfaaab6c3aa706d7cf1f3163b6bd01/proposals/0000-instance-scoping.rst#id4)—the internals of GHC treat the `:: Maybe a -> Type` bit as part of the data instance's RHS. But this example was _just_ different enough from the examples in this proposal that I'd thought I'd double-check.

simonpj:
>  I realized that Change #1 in this proposal doesn't really say what should happen to data family instances.

Ah yes, very true.  It should be absolutely specific.

I think that *data instances should be distinguishable (from each other) from the argument(s) to the left of the `::`*.  So yes, these are all OK
```
data instance forall b. D @(Maybe b) :: Maybe b -> Type where ...
data instance D @[c] (d :: [c]) where ...
data instance D (e :: (p,q)) where ...
```
(The `forall` in the first is optional, just added for variation.)

But these are not OK.
```
data instance D :: Maybe b -> Type where ...
data instance D :: [c] -> Type where ...
data instance D :: (p,q) -> Type where ...
```

JakobBruenker:
Just wanted to note that I'm still interested in this proposal; I just ran into a subtle bug again because of the status quo:

```haskell
type All :: (a -> Constraint) -> [a] -> Constraint
type family All c xs where
  All c '[] = ()
  All c (x:xs) = (x, All c xs)
```

This is accepted without warning, and at first glance might look fine - it did to me.

But in fact, the last line should have `(c x, All c xs)`, not `(x, All c xs)`; with just `x`, it gets stuck unless `a ~ Constraint`.

Not spotting this can lead to rather confusing type errors down the line.

int-index:
@JakobBruenker The changes proposed here have been incorporated into [GHC Proposal #425](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0425-decl-invis-binders.rst) and accepted. With regards to your example, see the "Proposed Change Specification: Instances" section, and more specifically this part of it:

> **Instantiation**
> In type family and data family instances, the instantiation is fully determined by the left hand side, without looking at the right hand side.

In fact, this change is tracked in [GHC ticket #23515](https://gitlab.haskell.org/ghc/ghc/-/issues/23515), where you'll also find a link to a prototype implementation in [MR !10672](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/10672). The change itself is relatively small and self-contained, so it should not be a problem to rebase it onto current master.

The actual problem is the breakage it causes. Turns out that relying on the RHS to select type instances is not as niche as we thought, even `base` does it in its definition of type-level `==`. Basically we just need someone motivated to pick this up and do the hard/boring part of the job: adapt the ecosystem (libraries shipped with GHC, `head.hackage`, etc).

JakobBruenker:
@int-index ah, thank you, that's good to know. I suppose this proposal could be closed, then.

I will spend some time on impact assessment to see if I can find the motivation to do the boring work.

simonpj:
> I will spend some time on impact assessment to see if I can find the motivation to do the boring work.

This would be a great thing to do.  The current situation is a mess, and matching only on the LHS is a real clean-up.  Thank you!

aspiwack:
@simonpj , @int-index to clarify, there's nothing left in this proposal that isn't part of #425 ? Are we good with closing this?

int-index:
That is my understanding, though it'd be great if Simon could confirm that I didn't miss anything when writing #425.

aspiwack:
Closing, then. Thanks.

PR #387, opened 2020-12-21
The Char Kind (under review)

DanielRrr:
We propose to introduce the built-in `Char` kind with such additional type families as `CmpChar`, `ConsSymbol` and `UnconsSymbol`.

[Rendered](https://github.com/serokell/ghc-proposals/blob/master/proposals/0000-char-kind.rst)

### DIFF
--- a/proposals/0000-char-kind.rst
+++ b/proposals/0000-char-kind.rst
@@ -8,74 +8,137 @@ The ``Char`` kind
 .. highlight:: haskell
 .. contents::
 
-At the moment, we have built-in type-level strings and natural numbers in GHC.TypeLits and
-GHC.TypeNats respectively. We also would like to introduce the ``Char`` kind.
-This proposal describes our updates that introduce the build-in character kind.
-This proposal also provides a solution for the issue `#11342 <https://gitlab.haskell.org/ghc/ghc/-/issues/11342>`_.
-Here we only provide changes making the character kind built-in as `Nat` and `Symbol`.
-This affects the type-checker, the parser, GHC core, and several libraries such as Template Haskell,
-Typeable, and TypeLits.
+Haskell has support for type-level data in the form of the ``DataKinds``
+extension, which promotes algebraic data types such as ``Bool``, ``Maybe``, or
+even user-defined ADTs. It also has special support for two non-algebraic
+data types: ``Symbol`` and ``Natural``, representing type-level strings and
+type-level numbers respectively.
 
+We propose to add support for one more non-algebraic data type, namely
+``Char``, thereby resolving the long-standing issue `#11342
+<https://gitlab.haskell.org/ghc/ghc/-/issues/11342>`_.
 
 Motivation
 ----------
 
-The purpose of the proposed changes is to provide a possibility of analysing type-level strings
-as we can make it with term-level ones. This feature allows users to implement such programs as
-type-level parsers. See examples above.
+The existing support for type-level strings is limited to:
 
-At the moment, we have no instruments allowing one to process type-level strings
-addressing to their characters. ``Symbol`` is a promoted ``FastString`` data type, not a usual
-list of chars for efficiency reasons. However, we still have no direct access to characters of a
-type-level string since there is no ability for pattern matching as in the term-level case. For that,
-one needs to have the full-fledged support for type-level characters similarly to strings and natural numbers.
-We are going to introduce the built-in ``Char`` kind. Secondly, we are going to provide several type families
-for the built-in ``Char`` kind. At this stage, we decided to include ``CmpChar`` for
-type-level comparison of type-level characters. We also introduce the type families called
-``ConsSymbol`` and ``UnconsSymbol`` that connect `Char` and `Symbol` with each other.
+1. String literals::
+
+     ghci> :kind "myString"
+     "myString" :: Symbol
+
+2. String concatenation::
+
+     ghci> :kind! AppendSymbol "hello" "world"
+     AppendSymbol "hello" "world" :: Symbol
+     = "helloworld"
+
+3. String comparison::
+
+     ghci> :kind! CmpSymbol "hello" "world"
+     CmpSymbol "hello" "world" :: Ordering
+     = 'LT
+
+There are no ways to decompose or analyse a type-level string in terms of its
+constituent characters. Constrast that with the API of `Text
+<https://hackage.haskell.org/package/text/docs/Data-Text.html>`_, which offers
+a multitude of functions such as ``uncons``, ``map``, ``splitAt``, and so on.
+
+We could try to extend the API of ``Symbol`` accordingly. For example, the
+type-level counterpart of ``uncons :: Text -> Maybe (Char, Text)`` could be a
+built-in type family such as ``UnconsSymbol :: Symbol -> Maybe (Char,
+Symbol)``.
+
+Notice that the return type of the proposed ``UnconsSymbol`` mentions ``Char``.
+However, there's currently no support for type-level characters::
+
+  ghci> :kind! 'x'
+  <interactive>:1:1: error: parse error on input ‘'’
+
+We propose to fix this omission. The ``Char`` kind and the accompanying
+built-in type families will make it possible to implement type-level parsers
+(see the "Examples" section below).
 
 Proposed Change Specification
 -----------------------------
 
-Proposed changes are the following:
+1. Extend the grammar of type-level literals
+   with character literals::
 
-1. That's how the grammar of type literals looks like with our changes::
+     tylit ::=
+         INTEGER
+       | STRING
+       | CHAR       (NEW)
 
-    tylit ::= nat | symbol | char
+   The lexical syntax matches that of term-level character literals: the
+   character enclosed in single quotes, e.g. ``'X'`` or ``'\n'``.
 
+2. Extend the ``GHC.TypeLits`` module
+   with the following built-in type families::
 
-2. We extend the GHC.TypeLits module with the following built-in type families::
+     type family CmpChar (a :: Char) (b :: Char) :: Ordering
+     type family ConsSymbol (a :: Char) (b :: Symbol) :: Symbol
+     type family UnconsSymbol (a :: Symbol) :: Maybe (Char, Symbol)
 
-    type family CmpChar (a :: Char) (b :: Char) :: Ordering
+   * The semantics of ``CmpChar`` match that of ``compare @Char``.
+   * The semantics of ``ConsSymbol`` and ``UnconsSymbol`` match that of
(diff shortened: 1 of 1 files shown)


proposals/0000-char-kind.rst
```diff
@@ -0,0 +1,315 @@
...
+2. Extend the ``GHC.TypeLits`` module
+   with the following built-in type families::
+
+     type family CmpChar (a :: Char) (b :: Char) :: Ordering
+     type family ConsSymbol (a :: Char) (b :: Symbol) :: Symbol
+     type family UnconsSymbol (a :: Symbol) :: Maybe (Char, Symbol)
+
+   * The semantics of ``CmpChar`` match that of ``compare @Char``.
+   * The semantics of ``ConsSymbol`` and ``UnconsSymbol`` match that of
+     ``Data.Text.cons`` and ``Data.Text.uncons`` respectively.
```
phadej, line 86: What about surrogates?

```
Prelude Data.Text> cons '\xD800' "foo"
"\65533foo"
Prelude Data.Text> 0xd800
55296
```

int-index: Very interesting. I didn't know about this, but here's a relevant excerpt from `text` sources:

```
-- | Map a 'Char' to a 'Text'-safe value.
--
-- UTF-16 surrogate code points are not included in the set of Unicode
-- scalar values, but are unfortunately admitted as valid 'Char'
-- values by Haskell.  They cannot be represented in a 'Text'.  This
-- function remaps those code points to the Unicode replacement
-- character (U+FFFD, \'&#xfffd;\'), and leaves other code points
-- unchanged.
safe :: Char -> Char
safe c
    | ord c .&. 0x1ff800 /= 0xd800 = c
    | otherwise                    = '\xfffd'
```

I think `ConsSymbol` should not reduce (i.e. it should get stuck) in this case. Either that, or a produce a proper type error.

phadej: FWIW, `Symbol` is not `Text`, so it can contain surrogate codepoints, where (safely) constructed `Text` cannot. `Symbol` is really `String` at the moment i.e a list of (possibly invalid) codepoints/`Char`s.

```haskell
Prelude GHC.TypeLits Data.Text> "\xD800" :: String
"\55296"
Prelude GHC.TypeLits Data.Text> "\xD800" :: Text
"\65533"
Prelude GHC.TypeLits Data.Text> case someSymbolVal "\xD800" of SomeSymbol p -> symbolVal p
"\55296"
```

phadej: ... so IMHO just going with ` (:)` and `uncons` semantics is fine IMHO. But if someone wants to fix `Symbol` to represent only correct Unicode text, that would be another proposal...

int-index: > `Symbol` is really `String` at the moment

Internally, `Symbol` is represented as a `FastString`, defined thus:

```
{-| A 'FastString' is a UTF-8 encoded string together with a unique ID. All
'FastString's are stored in a global hashtable to support fast O(1)
comparison.

It is also associated with a lazy reference to the Z-encoding
of this string which is used by the compiler internally.
-}
data FastString = FastString {
      uniq    :: {-# UNPACK #-} !Int, -- unique id
      n_chars :: {-# UNPACK #-} !Int, -- number of chars
      fs_sbs  :: {-# UNPACK #-} !ShortByteString,
      fs_zenc :: FastZString
      -- ^ Lazily computed Z-encoding of this string. See Note [Z-Encoding] in
      -- GHC.Utils.Encoding.
      --
      -- Since 'FastString's are globally memoized this is computed at most
      -- once for any given string.
  }
```

Note that it says "UTF-8 encoded". So either the implementation of `Symbol` has a bug (and it shouldn't use `FastString`), or its surface API has a bug (and it shouldn't represent byte sequences which are not valid UTF-8).

phadej: *UTF-8 encoded* is not correct. It's a Modified UTF-8 (= it doesn't contain zero-character). And it also can contain surrogates.

Modified-UTF8:

```haskell
-- | Gives the Modified UTF-8 encoded bytes corresponding to a 'FastString'
bytesFS, fastStringToByteString :: FastString -> ByteString
bytesFS = fastStringToByteString

{-# DEPRECATED fastStringToByteString "Use `bytesFS` instead" #-}
fastStringToByteString f = SBS.fromShort $ fs_sbs f

fastStringToShortByteString :: FastString -> ShortByteString
fastStringToShortByteString = fs_sbs
```

Surrogates: Because it has to, because non-abstract `String = [Char]` can contain them.

phadej: https://www.unicode.org/versions/Unicode13.0.0/ch03.pdf

Page 77.

> C1) A process shall not interpret a high-surrogate code point or a low-surrogate code point as an abstract character.
> 
> • The high-surrogate and low-surrogate code points are designated for surrogate code units in the UTF-16 character encoding form. They are unassigned to any abstract character.

I'm not sure if UTF-16 can even represent surrogate codepoints (which `text` uses atm).

---

This problem is very  deep in Haskell / GHC.

String literals like "\xD800" are wrong to begin with, they are not *Unicode Text*. Yet Haskell Report doesn't claim they are, so we shouldn't here either.

We can have debate elsewhere whether this trade-off  of "string literals are not unicode text because we need to represent also binary yadda-yadda" elsewhere. More productive (short-term) is just accept that strings in "core" Haskell are not "Unicode Text", but a list of codepoints (including non-abstract characters).

I speak from the experience of low-level problems these kind of hacks impose. For example, if `text` were to change to UTF-8 represention internally tomorrow, it would still need to do conversion from FastString Modified-UTF8 to recode zero-character and filter surrogates, so literals won't become any cheaper.

int-index: OK. So `Symbol` should be isomorphic to `[Char]`, and the semantics of `ConsSymbol` and `UnconsSymbol` should match that of `(:)` and [`GHC.List.uncons`](https://hackage.haskell.org/package/base-4.14.1.0/docs/GHC-List.html#v:uncons). At the same time, internally it's a `FastString` that stores Modified UTF-8, therefore its performance characteristics are different from that of `[Char]`.

Is that right?

phadej: @int-index Yes, that is right.

goldfirere:
I'm in strong support.

proposals/0000-char-kind.rst
Ericson2314: Has it been ever proposed that we fix `Char`? `ByteString` `[WordXx]` etc. are there to handle the other cases, IMO.

Ericson2314: On the other hand, https://simonsapin.github.io/wtf-8/ is how rust encodes surrogate codepoints as an extension of UTF-8. This could be used for fast string.

DanielRrr: Hmmmm
```
ghci> :set -XDataKinds
ghci> :set -XTypeFamilies
ghci> import GHC.TypeLits
ghci> type X = ConsSymbol '\xD800' "foo"
ghci> :kind! X
X :: Symbol
= "\55296foo"
```

DanielRrr: If I'm not mistaken, there exist strings with no counterparts in `FastString`. Is it crucial? If so, it seems like a problem for a new issue.

DanielRrr: > Has it been ever proposed that we fix Char?
How exactly?

DanielRrr: Now we have the following test case
```
testUncons :: ConsSymbol '\xD800' "foo" :~: "\55296foo"
testUncons = Refl
```

Ericson2314: > > Has it been ever proposed that we fix Char?
> 
> How exactly?

By only allowing valid unicode.

phadej: @Ericson2314 it's a swamp, e.g. `FilePath = String` forces us to have 'leaky" strings, which allow invalid unicode - because filepaths could be such.

https://github.com/rust-lang/rust/issues/12056 is "fun" read of some issues to consider...

Ericson2314: Well, time to advocate for `FilePath != String` then :D

kirelagin: > `FilePath = String` forces us to have 'leaky" strings

I thought filepath `ByteString -> String` decoding was using some weird rounde-trippable encoding that essentially allows to represent any sequence of bytes as a sequence of Unicode codepoints? I.e. the fact that a `FilePath` should have been a `ByteString` is an entirely different (and already solved) issue.

DanielRrr:
@nomeata

I would like to submit this proposal to the committee.

int-index:
@nomeata ping?

nomeata:
Thanks! Must have missed the notification

serras:
My recommendation is to accept the proposal: it fills a gap in our current kind-level machinery in the most minimal way.

simonpj:
I'm in support of this proposal.

But would another alternative be to redefine
````
type Symbol = [Char]
```
Then we would not need `ConsSymbol` and `UnConsSymbol`.  And it would make it like term level.

At least could this be listed under Alternatives, with reasons for rejection.

One reason for reject is that we use `Symbol` for record fields etc, and it'd impose a heavy overhead on the type checker to navigate nested cons-cells.  And we'd have to worry about type functions inside, things like
```
HasLabel T ('x' : F y : G z) ty
```
which simply don't arise today. Ugh!

phadej:
@simonpj I'd rather move `base` into other direction, where `Symbol` would be also an *opaque* (internally storage efficient) string type usable on term level.

int-index:
With this patch, `type String = [Char]` is promoted, so it can be used instead of `Symbol`. But we keep `Symbol` around for type-checking performance (and I agree with @phadej that it’d be nice to make it inhabited at the term level as well).

Ericson2314:
Yes, bringing `Symbol` into the fold should wait for more resolution of the `Text` vs `String` controversy, not anticipating it.

serras:
Is everybody fine then with accepting this proposal as is, with `Symbol` separate from `[Char]`?

In this direction, would it make sense to provide type families `ToSymbol` and `ToCharList`? They could be part of the `GHC.TypeLits` module, for example.

goldfirere:
I believe those type families could be built on top of the primitives in this proposal. Maybe we should provide them, but I think we can separate that library issue from the language issue here.

serras:
@DanielRrr everything looks good to me. Just one small thing: could you add the item in "unresolved questions" somewhere else? The idea is that once a proposal is accepted, nothing is unresolved any more.

int-index:
@serras Done, moved it to Alternatives

Icelandjack:
I see it's been accepted, I was wondering if there is a need for type families mapping between `[Char]` and `Symbol`.

PR #389, opened 2021-01-07
Better ambiguous variables

serras:
Right now the extension `AllowAmbiguousTypes` is too invasive. This document proposed a more fine-grained annotation `{-# AMBIGUOUS #-}` which can be applied to variable in a type signature.

[Rendered](https://github.com/serras/ghc-proposals/blob/ambiguous/proposals/0000-better-ambiguous-vars.md)

### DIFF
--- a/proposals/0000-better-ambiguous-vars.md
+++ b/proposals/0000-better-ambiguous-vars.md
@@ -6,13 +6,11 @@ ticket-url: ""
 implemented: ""
 ---
 
-This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/0>).
-**After creating the pull request, edit this file again, update the number in
-the link, and delete this bold sentence.**
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/389).
 
 # Better ambiguous variables
 
-Right now the extension `AllowAmbiguousTypes` is too invasive. This document proposed a more fine-grained annotation `{-# AMBIGUOUS #-}` which can be applied to variable in a type signature.
+Right now the extension `AllowAmbiguousTypes` is too invasive. This document proposes a more fine-grained annotation `{-# AMBIGUOUS #-}` which can be applied to variable in a type signature.
 
 ## Motivation
 
--- a/proposals/0000-better-ambiguous-vars.md
+++ b/proposals/0000-better-ambiguous-vars.md
@@ -1,0 +1,97 @@
+---
+author: Alejandro Serrano Mena
+date-accepted: ""
+proposal-number: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/0>).
+**After creating the pull request, edit this file again, update the number in
+the link, and delete this bold sentence.**
+
+# Better ambiguous variables
+
+Right now the extension `AllowAmbiguousTypes` is too invasive. This document proposed a more fine-grained annotation `{-# AMBIGUOUS #-}` which can be applied to variable in a type signature.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


phadej:
Shouldn't class declaration rather be

```haskell
-- Usage in a type class declaration
class forall {-# AMBIGUOUS #-} (n :: Nat). ToInt n where
```

or alternatively ordinary type-signatures allow annotation in use-site.

(Yes I know it's more verbose this way, but I think uniformity is more important).

int-index:
We have the accepted (but not implemented) [#232](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0232-AmbiguousType-pragma.rst).

I have not read either proposal closely, but they appear quite similar.

phadej:
> Should this proposal be guarded under an extension? Following the lead of `{-# OVERLAPPING #-}` and friends, the answer would be 'no'.

I think that is bad design, and better avoided. The same code fails to compile with GHC (/compiler) version which knows about `{-# OVERLAPPING #-}` vs one which doesn't. Compare with `{-# UNPACK #-}`. If compiler doesn't recognize `{-# UNPACK #-}` the code will still compile (and even run, but maybe not as well.

phadej:
@int-index I like this proposal more as it's more granular. It tells exactly which variable (of many) can be ambiguous.

serras:
@int-index I failed to find that proposal. If that's accepted, I guess my proposal does not make much sense...

goldfirere:
I agree that this might be an alternative to #232. It's different, so still worth considering, but it should be explicit about the comparison with #232.

One idea that was raised (sorry -- I forget by whom and where) was to aim for a future where there are no ambiguous variables, ever. That is, make ambiguity a hard error. Instead, use #281 to make call sites required to supply the instantiation for variables that would otherwise be ambiguous. An ambiguous variable essentially needs to be specified at call sites anyway, and so using #281 makes this fact even more obvious than other techniques, which are a bit round-about.

Ericson2314:
I agree with @phadej that annotations should be tied to language extensions. (I would support a proposal to do this retroactively, too.)

AntC2:
[meta]
> @serras  I failed to find that proposal. ...

I too had to search quite hard, and I knew it was here because I authored it.

simonpj:
Indeed #232 describes the choice made under "Alternatives" in the current proposal.

Per-type-variable ambiguity, as suggested here, is nice because it tells you how many type arguments are needed.  Eg here's a silly type:
```
f :: forall {-# AMBIGUOUS #-}  a
             {-# AMBIGUOUS #-} b
             c.
    (Num a, Num b) => c -> c
```
Every call of `f` should supply two types, but the third is optional.   So it's precise but quite heavyweight.

With visible dependent quantification, is there ever a reason to have an invisible ambiguous argument. Since you have to supply it at every call site, you may as well make it visible, no?

adamgundry:
> With visible dependent quantification, is there ever a reason to have an invisible ambiguous argument. Since you have to supply it at every call site, you may as well make it visible, no?

Even if an argument is ambiguous, there may exist call sites at which it need not be supplied (the [user's guide](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html?highlight=ambiguous#ambiguous-types-and-the-ambiguity-check) gives an example). Most of the time visible dependent quantification is probably preferable to ambiguity, but we should keep the latter around as an option.

simonpj:
> Most of the time visible dependent quantification is probably preferable to ambiguity

So in documenting `{-# AMBIGUOUS #-}` we should explicitly say that you should think about using VDQ instead.  (Once we get VDQ of course.)  In general, it's probably a better plan.

So if VDQ is best in all but really obscure cases, there's no great merit in allowing very fine-gran per-type-variable ambiguity.  Per-function is just fine.

Ericson2314:
When is VDQ not the better option? just calling pre-VDQ higher order functions?

serras:
Given that #232 is already accepted, and as @simonpj VDQ is a better tool in the middle and long-term, I'm closing this proposal. If somebody feels that per-variable ambiguity is still desirable, I'm happy to keep working on it, though.

PR #374, opened 2020-10-30
COVERED per-instance pragma

arybczak:
Provide a per-instance ``{-# COVERED #-}`` pragma that allows disabling the coverage condition for specified type variables.

 [Rendered proposal](https://github.com/arybczak/ghc-proposals/blob/covered-pragma/proposals/0000-covered-pragma.rst)

PR #390, opened 2021-01-07
Fine-grained pragmas for undecidability

serras:
This proposal subsumes the job of `UndecidableInstances` and `UndecidableSuperClasses`, making it more fine-grained.

[Rendered](https://github.com/serras/ghc-proposals/blob/instance-pragmas/proposals/0000-fine-grained-undecidable.md)

### DIFF
--- a/proposals/0000-fine-grained-undecidable.md
+++ b/proposals/0000-fine-grained-undecidable.md
@@ -1,0 +1,86 @@
+---
+author: Alejandro Serrano Mena
+date-accepted: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/0>).
+**After creating the pull request, edit this file again, update the number in
+the link, and delete this bold sentence.**
+
+# Fine-grained pragmas for undecidability
+
+Currently when one needs to "escape" the termination checker, this is only possible by enabling `UndecidableInstances` and `UndecidableSuperClasses` in a per-module basis. However, this means losing those checks for every single type class, family, or instance defined in that module. This proposal introduces new `{-# UNDECIDABLE #-}` pragmas to mark a specific definition, instead of the whole module.
+
+## Motivation
+
+Once a developer starts using type level computation, either in the form of type classes or type families, it's easy to end up in a situation in which the instances go outside of the Paterson conditions from `FlexibleInstances` or the coverage conditions from functional dependencies.
+
+```haskell
+{-# language DataKinds, KindSignatures #-}
+{-# language FunctionalDependencies, FlexibleInstances #-}
+
+data Nat = Z | S Nat
+class Sum (n :: Nat) (m :: Nat) (r :: Nat) | n m -> r
+instance Sum 'Z m m
+instance (Sum n m r) => Sum ('S n) m ('S r)
+```
+
+As GHC's error message point out, one can escape those checks by using `UndecidableInstances`.
+
+```
+<interactive>: error:
+    • Illegal instance declaration for ‘Sum ('S n) m ('S r)’
+        The coverage condition fails in class ‘Sum’
+          for functional dependency: ‘n m -> r’
+        Reason: lhs types ‘'S n’, ‘m’
+          do not jointly determine rhs type ‘'S r’
+        Un-determined variable: r
+        Using UndecidableInstances might help
+    • In the instance declaration for ‘Sum ('S n) m ('S r)’
+```
+
+However, this is quite a drastic measure. In order to accept _one_ instance, we need to disable the checks for _every_ instance defined in the module. One could imagine a situation in which that leads to an incorrect instance being accepted, which would have been caught if the check was accepted.
+
+```haskell
+instance (Sum 'Z m m) => Sum 'Z m m
+```
+
+## Proposed Change Specification
+
+Type classes, instances, or type families can be annotated with `{-# UNDECIDABLE #-}` to signal that the termination and coverage checks should be waived for that particular definition.
+
+This proposal subsumes the job of `UndecidableInstances` and `UndecidableSuperClasses`, but makes it more fine-grained. This has been done in the past with the `OverlappingInstances` extension, which was refined into a series of instance pragmas.
+
+## Examples
+
+The instance in the previous example would be written as follows:
+
+```haskell
+instance {-# UNDECIDABLE #-} (Sum n m r) => Sum ('S n) m ('S r)
+```
+
+The [example in the documentation](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#extension-UndecidableSuperClasses) for `UndecidableSuperClasses` would now be written:
+
+```haskell
+type family F a :: Constraint
+class {-# UNDECIDABLE #-} F a => C a where
+```
+
+## Effect and Interactions
+
+By using this pragmas the programmer does not have to enable the `UndecidableInstances` or `UndecidableSuperClasses` extensions, but rather mark _each_ undecidable definition explicitly. This makes it harder to accept a definition by mistake.
+
+## Costs and Drawbacks
+
+For modules using a lot of type level computation, there might be a large amount of undecidable instances or type families. In that case marking every single undecidable definition could be hard.
+
+## Unresolved Questions
+
+None as of now.
+
+## Implementation Plan
+
+I (Alejandro) volunteer to implement this proposal once / when this proposal is accepted.
+


phadej:
I repeat my comment https://github.com/ghc-proposals/ghc-proposals/pull/389#issuecomment-756375262

I think that is bad design, and better avoided. The same code fails to compile with GHC (/compiler) version which knows about `{-# UNDECIDABLE #-}` vs one which doesn't. There should *also* be a language pragma. Compare with `{-# UNPACK #-}`. If compiler doesn't recognize `{-# UNPACK #-}` the code will still compile (and even run, but maybe not as well).

phadej:
I also think `UNDECIDABLE` is too scary. `{-# TERMINATING #-}` is what we mean. (and what Agda etc use to turn-off termination checker.)

serras:
@phadej I see.. I guess any deprecation strategy would involve some pre-processor magic as follows.

```haskell
#if GHC_VERSION > ...
{-# language TerminationPragmas #-}
#else
{-# language UndecidableInstances #-}
#endif
```

I am not tied to `UNDECIDABLE`, I just took that name because it's used in GHC docs. Note however that `TERMINATING` does not talk about the failure of the coverage condition from type families.

phadej:
@serras and maybe it is good to not have "yes yes trust me in *all* aspects* pragma.

---

re GHC_VERSION, that was the case with `OverlappingInstances`. In fact GHC still has `{-# LANGUAGE OverlappingInstances #-}`. It just issues warnings when you use it. (And as you point out, sometimes it might sense to enable it module-wide, a bit like `-Wno-orphans` vs. potential per instance orphanage specifier.).

goldfirere:
I'm very much in favor of finer-grained control over these features, but I disagree with the particulars in the design here. Thanks for starting this conversation!

### Current state of affairs

By grepping in the source code, I gather the following facts:

1. `UndecidableSuperClasses` omits the call to `checkClassCycles` for all class declarations in a module. This allows for arbitrary predicates in a superclass constraint.

2. `UndecidableInstances` omits the termination check for all local quantified constraints in a module.

3. `UndecidableInstances` allows a type family in a superclass context. (I think this is a bug: https://gitlab.haskell.org/ghc/ghc/-/issues/19186)

4. `UndecidableInstances` allows non-type-variable arguments to classes in an instance context. (I think this is a different bug: https://gitlab.haskell.org/ghc/ghc/-/issues/19187)

5. `UndecidableInstances` skips the termination check for class instances.

6. `UndecidableInstances` skips the termination check for type family equations (both in open type family instances and closed type family equations).

7. `UndecidableInstances` allows a liberalization of the check for injective type family consistency. (See [the Note](https://gitlab.haskell.org/ghc/ghc/-/blob/master/compiler/GHC/Tc/Instance/Family.hs#L795).)

8. `UndecidableInstances` allows an instance of a class with fundeps to use the liberal coverage condition instead of the coverage condition.

Accordingly, a proposal to remove these extensions must account for all the actions the extensions cause today.

### Pragmas vs modifiers

I agree with @phadej above, in that we shouldn't accept/reject a program based on the presence/absence of a pragma. Pragmas are comments that happen to be interpreted by the compiler. As such, they are good for denoting when to `UNPACK`, specifying rewrite `RULES`, and enabling `OPTIONS`. But I don't think they are good for the purpose here.

Instead, I propose we consider [modifier](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0370-modifiers.rst) syntax.

### Concrete counter-proposal

1. Introduce a new modifier `%terminating`.

2. Putting `%terminating` before a class declaration skips its superclass-cycle check. Example: `%terminating class C (F a) => C a where ...`

3. Putting `%terminating` before the `forall` in a quantified constraint skips its termination check. Example: `f :: forall a. (%terminating forall b. C b a => C a a) => a -> a` or `g :: forall a. (%terminating forall . D a a => C a) => a -> a`. Note we need an empty `forall` to be the target of the `%terminating` modifier.

4. Putting `%terminating` before an instance declaration skips its termination check. Example: `%terminating instance Eq (Tree a a) => Eq (Rose a) where ...`

5. Putting `%terminating` before a `type instance` (or `type` in an associated type instance) skips the equation's termination check and allows for the more liberal injectivity consistency check. Examples: `%terminating type instance F [a] = G a a`, `instance D (Maybe a) where %terminating type F (Maybe a) = G a a`

6. Putting `%terminating` before a closed type family declaration skips the termination check for all of its equations and allows the more liberal injectivity consistency check. Example: `%terminating type family F a where F [a] = G a a; F Int = Bool`

7. Accept some form of #374 to deal with locally changing the coverage condition on a class instance.

### Alternative counter-proposal

If we really want e.g. class declarations to start in column 0, we could imagine putting the `%terminating` at the end (right before the `where`). The key thing for modifiers is that they should affect the thing written right afterwards. If we go with e.g. `class %terminating C (F a) => C a`, then I'd worry about a later clash with a modifier that would affect a context.

We could also allow a `%terminating` to be on a previous line:

```hs
%terminating
class C (F a) => C a
```

but that may be hard to parse.

Ericson2314:
@goldfirere to be fair, I don't think @phadej is against the annotation syntax, just making the annotation be always accepted. That is wouldn't be annotations mixing up obligatory and discretionary concepts, but the liberal acceptance of annotations causing users to be unaware of what their dependencies are. They are both fine issues, just distinct.

I personally don't mind the mixup that much. Annotations only concern me when they are for more "first-class" concepts or interfaces. For example if `=> {-# TERMINATING #-}` was sub-type that would be too far. But as long as they just tweak what's allowed in a "in-universe unobservable" way, they are fine with me.

------

On to more serious matters, yes `UndecidableInstances` is still *egregiously* overloaded. For many years I thought it only effected whether the solver terminated, but then I learned it also affected whether boxed evidence is guaranteed to evaluate. I think this is a very unfortunate state of affairs because IMO the latter is fair more meaningful, interesting, and consequential, and yet the former received top billing.

If we do reformulate what controls what, I would love an organization that makes the "go wrong at compiler time" vs "go wrong at runtime" breakdown clear.

serras:
Thanks @goldfirere for the extensive summary of affairs. If you don't mind, I'll take your counter-proposal as a base for refining the proposal.

serras:
Looking at #374, they also discussed there about a pragma/modifier for using the Patterson condition. Given that it seems quite plausible that #380 (GHC2021) would get `FlexibleInstances` and `FlexibleContexts` in the default mix, I think we should not go that way, and only add the "undecidable" pragma.

serras:
I've updated the proposal in several directions:
- Clarified the many roles of `Undecidable` extensions,
- Changed from pragmas to modifiers,
- Introduce both `%terminating` and `%covered` for cater for the different use cases.

serras:
> On to more serious matters, yes `UndecidableInstances` is still _egregiously_ overloaded. For many years I thought it only effected whether the solver terminated, but then I learned it also affected whether boxed evidence is guaranteed to evaluate. I think this is a very unfortunate state of affairs because IMO the latter is fair more meaningful, interesting, and consequential, and yet the former received top billing.
> 
> If we do reformulate what controls what, I would love an organization that makes the "go wrong at compiler time" vs "go wrong at runtime" breakdown clear.

Would you mind elaborating on that? I don't completely understand what you mean by "go wrong at runtime": isn't all of this just problems for the type checker, which may not terminate?

arybczak:
What is the relation of this proposal to #374 ?

I'm confused whether this one subsumes my proposal, but with a different syntax, considering that is has `%covered`. What does `%covered` do? Does it lift the coverage condition?

serras:
@arybczak my idea was this proposal to include the "safe" part of #374 (simply enabling the liberal condition), but not the "dysfunctional" part. Here `%covered` only enables the former check.

arybczak:
I see. If we want to add `%covered` as a means of selective relaxing of the coverage condition to its liberal version, then we might as well add `%dysfunctional` that disables it altogether, since at this point the implementation overhead is pretty much nil. That is, unless you don't want the scope creep.

BTW, the proposed modifier syntax looks nice, but what about backward compatibility? Pragmas have a huge advantage here, because with modifiers you need CPP at each definition site. I doubt people will be happy about that.

It also introduces inconsistency with existing OVERLAPS/INCOHERENT pragmas. If you say you want to convert them to modifiers at some point, then we'll have even more breakage.

serras:
The problem with pragmas is that they should be used only when removing the pragma would not affect the meaning of the program, they should rather be optimisations or information for the compiler. We have decided as of #370 to use a different syntax (with `%`) for new elements which modify another one.

Unfortunately, the ship has already sailed for OVERLAPS, although if this proposal gets accepted I plan to open a new one to have modifier-syntax for `%overlaps` and so forth.

arybczak:
> they should be used only when removing the pragma would not affect the meaning of the program

Can you elaborate on that? I see people say this, but I don't know why.

jvanbruegge:
If you write a class that would fail the termination check your code throw a compiler error. If one would add a pragma that disables the termination check the code would suddenly compile.
Compare that with rewrite rules, `INLINE` or `UNPACK`. Even if you would remove those pragmas, the code would still compile and run, just not as efficient.

arybczak:
Right, so some pragmas affect compilation and some not. I still don't see this as a problem.

On the other hand making maintenance of backward compatibility much more annoying is a real problem.

simonpj:
I'm generally in support of this direction of travel.   Some thoughts

* Please refer explicitly to #370.

* Please articulate the principle that we use modifiers because it affects whether the program is accepted or not.  (Pragmas should not.)

* I'd much prefer to say
  ```
  class %terminating  C (F a) => C a 
  ```
  with the modifier after the keyword. I see no reason to prohibit this.

* I wonder about allowing perhaps optional parense or brackets:
  ```
  instance( %terminating, %liberal )  C a a => C [a] [b] where ...
  ```
  to delimit the modifiers.

* Please give proposed changes to the BNF.

* I'm unsure about `%terminating` as the magic word.  You intend it to mean `%I_promise_this_is_terminating`.  But it could easily be read as `%check_this_is_terminating`.   Indeed the "I promise" stuff is like `unsafeCorece` which means `IpromiseThisIsTypeSafe`.

  Perhaps for superclasses it should be
  ```
  class(%LiberalSuperClasses) C (F a) => D a
  ```
  And for instances
  ```
  instance(%Liberal, %RiskNonTermination) ...
  ```
  `%RiskNonTermination` seems much better than `%terminating`.

goldfirere:
> * I'd much prefer to say
>   ```
>   class %terminating  C (F a) => C a 
>   ```
>   with the modifier after the keyword. I see no reason to prohibit this.
> 

The #370 modifiers proposal did not lay this out (it should have), but a key design principle for modifiers is that they modify the thing that comes next, if anything. By sticking to that organizing principle, we can be sure that modifiers won't interfere with each other. Imagine we had a different modifier `%erasable` that, for example, said that an instance constraint is compile-time only. If we had `instance %erasable (C a b) => D (Maybe a) [b]`, then the solver would simplify `[W] D (Maybe a) [b]` to `C a b`, but the `C a b` constraint would not be available in any methods defined in the instance. (Let's please not debate whether this example is worthwhile.) Critically, the modifier applies to the constraint. If we allowed your syntax, then we would have `instance %terminating %erasable (C a b) => ...`, and the modifiers get mixed up.

On the other hand, if modifiers always go before the thing they modify, then this cannot occur.

Exception: a modifier can go before a syntactic separator, like `;` or `where` or `)`, as there is no potential ambiguity there. Accordingly, I offer alternative syntax putting the `%terminating` right before the `where`. Yet a further alternative is to put the `%terminating` in the body:

```hs
instance ... where
  %terminating
  meth1 = ...
```

This won't work for open type family instances, though.

> * I'm unsure about `%terminating` as the magic word.  You intend it to mean `%I_promise_this_is_terminating`.  But it could easily be read as `%check_this_is_terminating`.   Indeed the "I promise" stuff is like `unsafeCorece` which means `IpromiseThisIsTypeSafe`.

Fair point. `%assumeTerminating`? `%noTerminationCheck`? `%uncheckedTerminating`?

Separately: should the keyword begin with a capital letter? Maybe.

goldfirere:
I've just submitted a fresh PR clarifying #370 a bit. See #392. In light of this clarification, it seems clear that the new keywords should indeed start with a capital letter -- they are in the type-level namespace.

Ericson2314:
@serras   

> @Ericson2314 would you mind elaborating on that? I don't completely understand what you mean by "go wrong at runtime": isn't all of this just problems for the type checker, which may not terminate?

See https://gitlab.haskell.org/ghc/ghc/-/issues/19045. Though actually the situation is a bit more muddled than I thought because only *some* of those example still work on 8.10? I guess Richard and I were both trying with an older compiler before.

simonpj:
> Fair point. %assumeTerminating? %noTerminationCheck? %uncheckedTerminating?

I quite like `%RiskNonTermination`.

arybczak:
> > Fair point. %assumeTerminating? %noTerminationCheck? %uncheckedTerminating?
> 
> I quite like `%RiskNonTermination`.

I'd prefer to stay with `%Undecidable` at this point. It captures the idea and is a single word.

serras:
> I'd prefer to stay with `%Undecidable` at this point. It captures the idea and is a single word.

The problem with "undecidable" is that it doesn't tell you what _is_ undecidable of all possible conditions!

serras:
I've updated the proposal with the diff in parse rules. I've gone with `%NoTerminationCheck`, `%LiberalCoverage`, and `%LiberalInjectivity`. So now it looks as follows:

```haskell
instance %NoTerminationCheck Eq (Tree a a) => Eq (Rose a) where ..
```

The only weird corner case is when you have overlapping pragmas and modifiers:

```haskell
instance %NoTerminationCheck {-# OVERLAPS #-}
```

serras:
It seems that most people in the discussion agree with the idea of splitting the extensions into fine-grained "annotations" of some sort, right? I got this feeling since we are mostly discussing the specifics of the syntax.

goldfirere:
I'm fine with the choice of keywords here, but I still say that the location of the keywords is troublesome: a modifier should always precede the thing it modifies. (Or we could use extra punctuation as @simonpj has suggested. But I prefer just putting modifiers prefix.)

simonpj:
> a modifier should always precede the thing it modifies

But why?   We can decide what we like?  I quite dislike a long string of verbiage preceding the keyword.  It's so simple to put it after the keyword.  (And the parens help to highlight this.)

> instance %NoTerminationCheck {-# OVERLAPS #-}

Perhaps we could take the opportunity to regularise the syntax thus
```
instance( %NoTerminationCheck, %Overlaps ) ....
```

arybczak:
For the record, I'd prefer the modifier before the keyword, because class/instance/type familiy declarations are verbose as is and I'd probably just put the modifiers on their own line, like this:

```haskell
%NoTerminationCheck %Overlapping
instance ...

%NoTerminationCheck
type family ...
```

this is much more readable to me that putting them somewhere in the middle.

simonpj:
> For the record, I'd prefer the modifier before the keyword, because that allows to put them on a separate line:

I can see merit in that, provided I don't need to indent the next line.  Thue
```
%NoTerminationCheck %Overlapping
instance ...
```
and not
```
%NoTerminationCheck %Overlapping
    instance ...
```
So with layout there'd be a semicolon, thus
```
%NoTerminationCheck %Overlapping ; instance ...
```
Maybe that'd fine.  After all, the successive equations of a function defn belong together, but are separated by semicolons.

------------

Reminiscent of *attributes* in OO langauges,  e.g. C#
```
[Serializable]
public class SampleClass { ... }
```

goldfirere:
> Reminiscent of attributes in OO langauges

Yes, the idea of modifiers is directly inspired by Java annotations. If we can work out the parsing for putting the modifiers (optionally) on the line before, that seems ideal.

arybczak:
This proposal has put #374 into awkward position because I considered it pretty much ready for submission to the committee (with a few minor adjustments), but now this one overlaps with wrt. the LIBERAL pragma, but with a different design, effectively blocking it.

However, since the overlap is only about the LIBERAL pragma, I think the optimal route here is to cut the scope of #374 to only the DYSFUNCTIONAL pragma. Then #390 can simply change the pragma syntax to the modifier if #374 is accepted.

Does that make sense?

serras:
@arybczak we could also split it as it was, although I think that the discussion about `DISFUNCTIONAL` is going to take longer, and should be separate from `LIBERAL` altogether.

serras:
I've updated the proposal to have the annotations before the corresponding keywords. I like the fact that you can choose to have its on its own line, but also on the same line with a `;` in between.

serras:
@simonpj has mentioned the possibility of regularizing the syntax for `OVERLAP` pragmas. I think this is a great idea, so I've included it as part of the specification.

Should we even take a more ambitious position and also regularize [`AMBIGUOUS`](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0232-AmbiguousType-pragma.rst) before it is implemented?

simonpj:
> but now this one overlaps with wrt. the LIBERAL pragma

Can you be more specific?  What is the conflict? Purely syntactic?

simonpj:
> @simonpj has mentioned the possibility of regularizing the syntax for OVERLAP pragmas. I think this is a great idea, so I've included it as part of the specification.

Good -- but I could not see any mention of OVERLAP in the proposal.

serras:
> > @simonpj has mentioned the possibility of regularizing the syntax for OVERLAP pragmas. I think this is a great idea, so I've included it as part of the specification.
> 
> Good -- but I could not see any mention of OVERLAP in the proposal.

I've mentioned it (briefly) in the first paragraph and in point 4 of _Proposed Change Specification_

simonpj:
> I've mentioned it (briefly) in the first paragraph and in point 4 of Proposed Change Specification

Ah... buried in the BNF not mentioned in the text.  This needs bringing out explicitly.  And presumably the pragmas are deprecated for a cycle or two.

Also the BNF is a bit mixed up.  `cls_mods` in some places, `class_mods` in others; `cls_mod` shouldn't have a semicolon if `cls_mods` does.  And you have yet another semicolon in `cl_decl`.

Please also refer explicitly to the modifiers proposal, which is the source of the syntax offered here.

arybczak:
> I think that the discussion about DISFUNCTIONAL is going to take longer, and should be separate from LIBERAL altogether.

@serras I think the discussion about `DYSFUNCTIONAL` in #374 is pretty much settled (well, excluding potential committee discussion), whereas this proposal introduces a new syntax and discussions about syntax have an unfortunate tendency to last forever ;) But let's hope this time it'll be better.

> Can you be more specific? What is the conflict? Purely syntactic?

@simonpj Yes, purely syntactic. A pragma `{-# LIBERAL #-}` vs a modifier `%LiberalCoverage`.

serras:
I've just pushed a new version taking care of some of @simonpj's concerns. About the semicolon situation, some of those are actually part of Happy syntax - the diff is taken from the actual GHC source code.

@arybczak I think the syntactic discussion would affect both proposals anyway, so it's better to clarify whether we want modifiers or pragmas.

serras:
@nomeata this is ready for Committee review.

simonpj:
There is still something wonky in the BNF.  Look at `cl_decl`, and consider this derivation
```
cl_decl
==>  cls_mods                            ';' 'class' tycl_hdr fds where_cls
==> cls_mods ';' cls_mod ';'       ';' 'class' tycl_hdr fds where_cls
==> {- empty -} ';' %NoTerminationCheck ';'    ';'       ';' 'class' tycl_hdr fds where_cls
```
Wow!  No fewer than three bogus semicolons.

Or maybe the semicolons in `cls_mod` and `cls_mods` are meta-syntax, in which case just delete them.  They are horribly confusing.

simonpj:
> For that reason, this proposal also regularizes the syntax, turning those pragmas into modifiers.

But that is all the proposal says about this; and it contradicts the earlier "We introduce new modifiers %NoTerminationCheck, %LiberalCoverage, and %LiberalInjectivity".  Actually we are introducing %Overlaps and %Overlapping as well.

Better to move the "For that reason" stuff into Motivation, and make overlap stuff a proper first class citizen of "Proposed Change Spec".

serras:
Updated with the proposed changes

arybczak:
What's the status?

nomeata:
The status is that I missed the message that this is ready for committe deliberation, sorry

bravit:
Hi, I've just recommended accepting this proposal. I am sure this is the right thing to do!

goldfirere:
I like where this has gone, but I still have a few more smallish suggestions:

* #370 suggests that modifiers are parsed and renamed as types. That is, instead of just a single name appearing after the `%`, we might have something more involved. I don't think this proposal needs to support having non-atomic phrases after the `%`, but I do think it's nice for users to have a uniform understanding of modifier syntax. I thus propose to update this proposal with adding the modifiers (as empty datatypes) to a new module `GHC.Modifiers`.

* There is nothing in this proposal about phasing out the current approaches. Maybe we want to keep both, in order to have backward compatibility. But maybe not. In any case, I do think the proposal should be explicit on this question. I have a hard time coming up with a solid opinion on what the choice should be.

nomeata:
>  I thus propose to update this proposal with adding the modifiers (as empty datatypes) to a new module `GHC.Modifiers`.

Would that have to be imported then?

serras:
> #370 suggests that modifiers are parsed and renamed as types. That is, instead of just a single name appearing after the `%`, we might have something more involved. I don't think this proposal needs to support having non-atomic phrases after the `%`, but I do think it's nice for users to have a uniform understanding of modifier syntax. I thus propose to update this proposal with adding the modifiers (as empty datatypes) to a new module `GHC.Modifiers`.

I don't fully grasp the consequences of this. It seems to me that modifiers for declarations are a bit different than those for types, in that the latter go into the constraint solving phase, in which we can work out the modifier stuff.

How do you envision parsing and renaming working in the case of `instance` or `type`? Is it possible import a module which states `type Weird = Undecidable`, for example?

goldfirere:
The idea behind modifiers is that, mostly, they are like keywords. But by parsing and renaming them as types, we are allowed to encode arbitrary compile-time reasoning into modifiers if and when we choose to do so. All the modifiers proposed here would just be keywords, and it's hard to see why we might want to expand this later... but maybe we will. As a strawman idea, maybe we want named instances at some point. We might write

```hs
%(Named "wildebeest") %NoTerminationCheck
instance Foo Bar where ...
```

I don't have to introduce new syntax to get `Named "wildebeest"` to be accepted.

Yes, the modifiers would have to be imported. I think this is a feature, not a bug. Whenever we have new features that don't need to be imported, it turns out to be a bad idea later. (I'm looking at you, `~` and `*`.)

Do we accept type synonyms? I don't much care whether we do or don't. Parsing and renaming as types does not imply that we have to follow type synonyms. Though if we imagine allowing `%(NoTerminationCheck, LiberalCoverage)` to mean the conjunction of the two (a reasonable imagination), then we could have a synonym for that, and this would be useful, perhaps. I don't feel strongly.

Actually, I don't feel strongly about any of it, but I do think there's something nice about having a uniform treatment of modifiers. It will ease the burden on readers, knowing how any phrase after a `%` is handled.

proposals/0000-fine-grained-undecidable.md
```diff
@@ -0,0 +1,185 @@
...
+    %NoTerminationCheck ; class C (F a) => C a where ...
+    ```
+
+    ```diff
+    +cls_mod  : '%NoTerminationCheck'
+    +cls_mods : {- empty -} | cls_mods cls_mod
+
+    cl_decl :
+    - : 'class' tycl_hdr fds where_cls
+    + : cls_mods ';' 'class' tycl_hdr fds where_cls
```
simonpj, line 77: This can't be right.  It suggests that if there are no `cls_mods` then the class decl still needs to be preceded by a semi-colon.

proposals/0000-fine-grained-undecidable.md
```diff
@@ -0,0 +1,185 @@
...
+    +forall_mod  : '%NoTerminationCheck'
+    +forall_mods : {- empty -} | forall_mods forall_mod
+
+    forall_telescope : 
+    - :             'forall' tv_bndrs '.'
+    + : forall_mods 'forall' tv_bndrs '.'
+    - |             'forall' tv_bndrs '->'
+    + | forall_mods 'forall' tv_bndrs '->'
+
+3. Putting `%NoTerminationCheck` before an instance declaration skips its termination check. Putting `%LiberalCoverage` allows for the more liberal coverage condition.
```
simonpj, line 99: "instance"  -->  "type class instance"

proposals/0000-fine-grained-undecidable.md
```diff
@@ -0,0 +1,185 @@
...
+
+    ```haskell
+    %NoTerminationCheck ; instance Eq (Tree a a) => Eq (Rose a) where ..
+    ```
```
simonpj, line 103: Put the BNF changes for cls_inst_mod here.

proposals/0000-fine-grained-undecidable.md
```diff
@@ -0,0 +1,185 @@
...
+    ```diff
+    +cls_inst_mod  : '%NoTerminationCheck' | '%LiberalCoverage'
+    +              | '%Overlapping' | '%Overlaps' | '%Overlappable'
+    +cls_inst_mods : {- empty -} | cls_inst_mods cls_inst_mod
+    +ty_fam_mod    : '%NoTerminationCheck' | '%LiberalInjectivity'
+    +ty_fam_mods   : {- empty -} | ty_fam_mods ty_fam_mod
+
+    inst_decl
+    - :               'instance' overlap_pragma inst_type where_inst
+    + : inst_mods ';' 'instance' overlap_pragma inst_type where_inst
```
simonpj, line 125: Should we not deprecate `{-# OVERLAPS #-}`  and `{-# OVERLAPPABLE #}` etc in favour of this new stuff?  The proposal should say.

proposals/0000-fine-grained-undecidable.md
```diff
@@ -0,0 +1,185 @@
...
+    - |             'forall' tv_bndrs '->'
+    + | forall_mods 'forall' tv_bndrs '->'
+
+3. Putting `%NoTerminationCheck` before an instance declaration skips its termination check. Putting `%LiberalCoverage` allows for the more liberal coverage condition.
+
+    ```haskell
+    %NoTerminationCheck ; instance Eq (Tree a a) => Eq (Rose a) where ..
+    ```
+
+5. Putting `%NoTerminationCheck` within a type instance (or type in an associated type instance) skips the equation's termination check. Putting `%LiberalInjectivity` allows for the more liberal injectivity consistency check.
```
simonpj, line 105: What does "within" mean here?

Also you are talking about open type families here; mention that.

In fact, for each of these numbered items could you start with a heading.  Like

5. **Open type families**.   Putting...

That would make the structure much clearer

proposals/0000-fine-grained-undecidable.md
```diff
@@ -0,0 +1,185 @@
...
+    inst_decl
+    - :               'instance' overlap_pragma inst_type where_inst
+    + : inst_mods ';' 'instance' overlap_pragma inst_type where_inst
+    - |                 'type' 'instance' ty_fam_inst_eqn
+    + | ty_fam_mods ';' 'type' 'instance' ty_fam_inst_eqn
+    - |                 data_or_newtype 'instance' capi_ctype datafam_inst_hdr ...
+    + | ty_fam_mods ';' data_or_newtype 'instance' capi_ctype datafam_inst_hdr ...
+    ```
+
+6. Putting `%NoTerminationCheck` within a closed type family declaration skips the termination check for all of its equations. Putting `%LiberalInjectivity` allows for the more liberal injectivity consistency check.
```
simonpj, line 132: Again "within".  I think you mean "immediately before".

proposals/0000-fine-grained-undecidable.md
```diff
@@ -0,0 +1,185 @@
...
+    ```haskell
+    %NoTerminationCheck ; type family F a where
+      F [a] = G a a
+      F Int = Bool
+    ```
+
+    ```diff
+    ty_decl
+    - |                 'type' 'family' type opt_tyfam_kind_sig ...
+    + | ty_fam_mods ';' 'type' 'family' type opt_tyfam_kind_sig ...
```
simonpj, line 143: Again compulsory semicolon is wrong if there are no mods.

simonpj:
> The idea behind modifiers is that, mostly, they are like keywords. But by parsing and renaming them as types, we are allowed to encode arbitrary compile-time reasoning into modifiers if and when we choose to do so.

I quite like this.  Yes, we'd need to import, but saying `import GHC.Modifiers( Overlapping )`  is quite illuminating.

For now, we could say that they are parsed and renamed as type *constructors* (`qtycon`).   We could later extend types if we wanted.

bravit:
Hi @serras, it looks like we are not ready to decide on the proposal right now. What do you think if we set the "needs revision" label and then you have time to address these issues?

int-index:
I agree that `%` must be followed by a type, and these types could be imported from `GHC.Modifiers`. The proposed change specification should include declarations of these types, e.g.:

```haskell
data TerminationCheck =
  Terminating | NoTerminationCheck

data OverlapCheck =
  Overlapping | Overlappable | Overlaps | NoOverlap
```

Notice also that I included constructors to explicitly specify the default behavior, e.g. `%Terminating` and `%NoOverlap` explicitly enforce the respective properties.

proposals/0000-fine-grained-undecidable.md
serras: The proposal states:

> Note that the old pragmas are _not_ deprecated by this proposal.

although this is of course open to discussion.

serras:
It's taken some time, but I've finally updated the proposal to include the suggestions.

@int-index I've decided _not_ to include the negative constructors (`NoOverlap` and friends) because it was unclear how to rule out cases like `%NoOverlap %Overlaps ; instance ...`.

@simonpj I've added headings to the sections, and moved all of the grammar rules into a central location. I think the rule in the updated proposal is now correct and does not force the programmer to always write `;` before declarations.

int-index:
> because it was unclear how to rule out cases like `%NoOverlap` `%Overlaps`

I assumed we’d only allow one modifier per kind, i.e. at most one `TerminationCheck` modifier and at most one `OverlapCheck` modifier.

The reason I assumed that is that when linear arrows are reconceptualized as modifiers (i.e. `a %m -> b`), we’d have to give a reasonable error message if the user wrote `a %Many %One -> b` or something like that. So we’ll definitely need “one per kind” logic here.

You decided to pursue a different design, perhaps allowing more than one modifier of any given kind. That, however, will not work for multiplicity annotations, and we should be clear about this inconsistency.

simonpj:
You syntax says
```
+modifiers : {- empty -} | ('%' qtycon)* ';'
```
that is, always with a trailing semicolon.

But I don't think you intend that for the `forall_telescope` do you?   At least not in the examples you give.

serras:
> But I don't think you intend that for the `forall_telescope` do you? At least not in the examples you give.

You are completely right. I've just pushed a new revision.

proposals/0000-fine-grained-undecidable.md
```diff
@@ -0,0 +1,198 @@
...
+- `UndecidableInstances` allows a liberalization of the check for injective type family consistency.
+- `UndecidableInstances` allows an instance of a class with functional dependencies to use the liberal coverage condition instead of the coverage condition.
+
+## Proposed Change Specification
+
+The modifier syntax is taken from this [approved proposal](https://github.com/ghc-proposals/ghc-proposals/pull/370), with the additional [change](https://github.com/ghc-proposals/ghc-proposals/pull/392) of modifiers appearing _before_ the modified thing. Since modifiers ought to be types, we introduce the following declarationg in the `GHC.Generics` module.
+
+```haskell
+data TerminationCheck =
+  Terminating | LiberalCoverage | LiberalInjectivity
```
audunska, line 67:
```suggestion
  NoTerminationCheck | LiberalCoverage | LiberalInjectivity
```

proposals/0000-fine-grained-undecidable.md
```diff
@@ -0,0 +1,198 @@
...
+- `UndecidableInstances` skips the termination check for type family equations (both in open type family instances and closed type family equations).
+
+There are also tweaks to the coverage conditions for coverage / injectivity in type families and functional dependencies.
+
+- `UndecidableInstances` allows a liberalization of the check for injective type family consistency.
+- `UndecidableInstances` allows an instance of a class with functional dependencies to use the liberal coverage condition instead of the coverage condition.
+
+## Proposed Change Specification
+
+The modifier syntax is taken from this [approved proposal](https://github.com/ghc-proposals/ghc-proposals/pull/370), with the additional [change](https://github.com/ghc-proposals/ghc-proposals/pull/392) of modifiers appearing _before_ the modified thing. Since modifiers ought to be types, we introduce the following declarationg in the `GHC.Generics` module.
```
audunska, line 63: Why GHC.Generics in particular?

serras: Oops, this was meant to be `GHC.Modifiers` instead.

serras:
@nomeata I think I've fixed all the problems mentioned in the thread. I would like to submit the proposal to the Committee.

arybczak:
Ping. What's the status?

nomeata:
Sorry, I must have missed @serras’s ping.

nomeata:
@bravit, you are still the shepherd, so please have a look and submit your recommendation to the committee.

serras:
Hi @bravit, any update on this?

bravit:
I've repeated my recommendation to accept this proposal.

goldfirere:
I'm quite happy with this. Thanks!

How would we feel about making the `;`s optional? That would allow, for example

```hs
class C Int where
  %LiberalInjectivity type F Int = ...
```

instead of the current

```hs
class C Int where
  %LiberalInjectivity
  type F Int = ...
```

Maybe it's not worth the complication in the parser?

int-index:
Modifiers are identified by their type, e.g. on an arrow the plan is to have a `Multiplicity` and a `Machability` modifier, like so:

```
f :: forall (p :: Multiplicity) (m :: Matchability). a %p %m -> b
```

The way I see this, it’s a bit like a record of properties:

```
{
  multiplicity = ...,
  matchability = ...
}
```

except it’s indexed by the modifier type, not by a field name.

And we have an _identity_ modifier, so to speak. `a %Many -> b` is the same as `a -> b`.

So I expect this proposal to follow this design, so that we could characterize, say, an instance declaration with a record of properties:

```
{
  terminationCheck = ...,  -- enabled/disabled
  liberalCoverage = ..., -- enabled/disabled
  liberalInjectivity = ..., -- enabled/disabled
  overlapping = ..., -- enabled/disabled
  overlappable = ..., -- enabled/disabled
}
```

The default is that the termination check is enabled, overlap is disabled. So I could write identity modifiers like so:

```
%TerminationCheckEnabled %NoLiberalCoverage %NoLiberalInjectivity %NoOverlap
instance C Int where
  ...
```

This would be the same as writing no modifiers, but I can be explicit about my choices.

To achieve this, we could define the following in `GHC.Modifiers`:

```
data OverlapConfig = Overlap { overlapping :: Bool, overlappable :: Bool }

type Overlappable = Overlap False True
type Overlapping = Overlap True False
type Overlaps = Overlap True True
type NoOverlap = Overlap False False

data LiberalCoverageConfig = NoLiberalCoverage | LiberalCoverage
data LiberalInjectivityConfig = NoLiberalInjectivity | LiberalInjectivity

data TerminationCheck = NoTerminationCheck | TerminationCheckEnabled
```

I would prefer that over the definitions currently proposed. What do others think?

proposals/0000-fine-grained-undecidable.md
```diff
@@ -0,0 +1,198 @@
...
+    %NoTerminationCheck ; class C (F a) => C a where ...
+    ```
+
+2. **Type class instance**: the `%NoTerminationCheck` modifier before an type class instance declaration skips its termination check. The `%LiberalCoverage` modifier allows for the more liberal coverage condition.
+
+    ```haskell
+    %NoTerminationCheck ; instance Eq (Tree a a) => Eq (Rose a) where ..
+    ```
+
+3. **Quantified constraint**: the `%NoTerminationCheck` modifier before the `forallp  in a quantified constraint skips its termination check.
```
simonpj, line 118: Typo here "forallp"

proposals/0000-fine-grained-undecidable.md
```diff
@@ -0,0 +1,198 @@
...
+- `UndecidableInstances` skips the termination check for type family equations (both in open type family instances and closed type family equations).
+
+There are also tweaks to the coverage conditions for coverage / injectivity in type families and functional dependencies.
+
+- `UndecidableInstances` allows a liberalization of the check for injective type family consistency.
+- `UndecidableInstances` allows an instance of a class with functional dependencies to use the liberal coverage condition instead of the coverage condition.
+
+## Proposed Change Specification
+
+The modifier syntax is taken from this [approved proposal](https://github.com/ghc-proposals/ghc-proposals/pull/370), with the additional [change](https://github.com/ghc-proposals/ghc-proposals/pull/392) of modifiers appearing _before_ the modified thing. Since modifiers ought to be types, we introduce the following declarationg in the `GHC.Modifiers` module.
```
simonpj, line 63: Please can we not talk about proposal 370 and a change to it?   We can just talk about the Final Version, with the change applied.  I'm not sure how to link to that, but let's just stick to the one, final, version.

simonpj:
I'm in support of this proposal.

But I would like us to drive libraries more actively towards using the new syntax, by deprecating the old `{-# OVERLAPS #-} ` etc.   Maybe 
* Add the new stuff in release X
* Deprecate the old stuff in release X+1
* Remove the old stuff in release X+3

or something like that. Let's actively push towards a coherent language, giving people time to adapt.

goldfirere:
I really like @int-index's refactoring. We could also then error if two modifiers of the same type are given on the same entity, as these modifiers would be contradictory.

i-am-tom:
I like this proposal, as this solves a problem I've run into before. Also want to +1 @int-index's suggestion - I think it makes the writer's intentions a lot more obvious.

serras:
I've updated the proposal with @int-index suggestions (they are great!) and the small things from @simonpj.

Unfortunately I'm not versed enough in GHC's parsing machinery to know whether @goldfirere's proposal of dropping the `;` requires lots of changes to the parser. I think that's the last missing piece to complete the proposal.

simonpj:
I think this proposal is back in author-revision state now, is that right?   I don't know how to change the status, but it currently says "committee review".

On the content
* I doubt that making the semicolons optional would be hard. @int-index is our expert on parsing.
* I think then the syntax can be
  ```
  modifiers ::= /* empty */  |   modifier modifiers
  modifier ::= atype [ ; ]
  ```
  That is, modifiers is zero or more modifiers; and each modifier is an atype, followed by an optional semicolon.

* Note 'atype' not 'qtycon'.   Is there anything wrong with `Overlap { overlapping = False, overlappable True }` as a modifier?

* The proposal should be specific about what types are legal in the modifers it describes
  * You can have the data constructors `NoLiberalCoverage`, `LiberalCoverage` ... (specify the full list)
  * You can have a type synonym that expands so those data constructors
  * But nothing else.  No type families, no newtypes...

* The proposal should say in which module these modifier types are defined; or at least the canonical module from which they should be imported.

arybczak:
What's the status of this proposal? Asking since #374 is semi-dependent on it. @serras is there a plan to get it through the finish line?

nomeata:
Bravit was the shepherd, but he has rotated out of the committee. So if @serras (as the author) picks it up again and says he has addressed all comments (e.g. Simons above), please ping me so that I can re-assign it.

serras:
Right now I don't have the energy to address the comments, so you can consider this proposal as dormant.

PR #392, opened 2021-01-08
Clarify modifiers design principle

goldfirere:
This PR is an amendment to the previously accepted Modifiers proposal #370.

Changes:

* Clarify that modifiers *precede* the thing they modify.

* Correct the syntax for linear record fields, which was different than the syntax in the accepted linear types proposal.

* Allow a limited facility for future compatibility.

Also:

* Change the syntax for annotating a lambda-expression with a multiplicity. This did not conform to the general scheme for modifiers. Given that this feature is also unimplemented, it seemed like a good time to tweak the syntax to allow for a more uniform treatment for modifiers. This patch also comes with a tweak to the linear-types proposal to mirror the update here.

Original modifiers proposal: https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0370-modifiers.rst
Updated modifiers proposal: https://github.com/goldfirere/ghc-proposals/blob/clarify-modifiers/proposals/0370-modifiers.rst
Diff: https://github.com/goldfirere/ghc-proposals/compare/7bb3ca581771fcac4753140e51e8df33bc10c0c0...clarify-modifiers  You can view a "rich diff", showing the changes in a rendered version, by clicking on the "page" icon to the top-right of the modified files on that page.

### DIFF
--- a/proposals/0111-linear-types.rst
+++ b/proposals/0111-linear-types.rst
@@ -307,8 +307,6 @@ This proposal adds two new syntactical constructs:
   where the ``btype`` after the ``%`` must be of kind ``Multiplicity``
   (see below).
 
-  This form is allowed only when the lambda binds precisely one variable.
-
 - Record fields can be labeled with a multiplicity. This modifies
   the syntax for record fields as follows::
      
--- a/proposals/0111-linear-types.rst
+++ b/proposals/0111-linear-types.rst
@@ -288,24 +288,25 @@ This proposal adds two new syntactical constructs:
 
   ::
 
-    \x :: A %'One -> x
+    \ (%'One x) :: A -> x
 
   is the identity function at type ``A ⊸ A``. A binder can be
   annotated with a multiplicity without a type like this
 
   ::
 
-    \x %'One -> x
+    \ %'One x -> x
 
   This modifies the syntax entry for lambda expressions
   as follows
 
   ::
 
-    lexp -> \ apat1 ... apatn [PREFIX_PERCENT btype] -> exp
+    lpat -> [PREFIX_PERCENT btype] lpat
 
   where the ``btype`` after the ``%`` must be of kind ``Multiplicity``
-  (see below).
+  (see below). The ``lpat`` must be a bare variable, or an error
+  occurs.
 
 - Record fields can be labeled with a multiplicity. This modifies
   the syntax for record fields as follows::
--- a/proposals/0111-linear-types.rst
+++ b/proposals/0111-linear-types.rst
@@ -284,7 +284,7 @@ This proposal adds two new syntactical constructs:
 
       type family F (a :: *) :: Multiplicity
       f ::  forall (a :: *). Int  %(F a) -> a -> a
-- When ``-XScopedTypeVariables`` is switched on, binders can also be annotated with a multiplicity:
+- Lambda-bound variables can also be annotated with a multiplicity:
 
   ::
 
@@ -297,40 +297,21 @@ This proposal adds two new syntactical constructs:
 
     \x %'One -> x
 
-  This modifies the syntax entry for pattern with signature annotation
-  as follows as follows
+  This modifies the syntax entry for lambda expressions
+  as follows
 
   ::
 
-    pat -> pat [PREFIX_PERCENT btype] [:: type]
+    lexp -> \ apat1 ... apatn [PREFIX_PERCENT btype] -> exp
 
   where the ``btype`` after the ``%`` must be of kind ``Multiplicity``
   (see below).
 
-  This form is disallowed for:
+  This form is allowed only when the lambda binds precisely one variable.
 
-  - Type variables
-
-    ::
-
-      forall (a %'One). a -> Int -- rejected
-  - Top-level signatures (though, see `Toplevel binders`_)
-
-    ::
-
-      foo %'One :: A -> B -- rejected
-      foo x = …
-
-  The form is however permitted in records (see `Records`_ below)
-
-  ::
-
-    data R = R { unrestrictedField %'Many :: A, linearField %'One :: B }
-
-  This modifies the field declaration syntax to
-
-  ::
-
+- Record fields can be labeled with a multiplicity. This modifies
+  the syntax for record fields as follows::
+     
     fielddecl -> vars [PREFIX_PERCENT btype] :: (type | ! atype)
 
 In the fashion of levity polymorphism, the proposal introduces a data
--- a/proposals/0370-modifiers.rst
+++ b/proposals/0370-modifiers.rst
@@ -66,20 +66,22 @@ Proposed Change Specification
 
      fexp     ::= {modifier} aexp | fexp aexp
 
-5. With ``-XModifiers``, introduce postfix modifier syntax on patterns as follows::
+5. With ``-XModifiers``, introduce modifier syntax in lambda expressions as follows::
 
-     lpat     ::= apat {modifier}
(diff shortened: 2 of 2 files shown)


simonpj:
The proposal would be stronger if every bullet in the "change spec" list had at least one accompanying example.  Many do not.

There no syntax to attach a modifier to a class, instance, data type, or type instance decl.  That's facility is needed by #390 so it would be good to lay it out here.

You can have un-ambiguity without following a rigid prefix form.  E.g.
```
class_decl ::= 'class'  [ '(' modifier1, ... , modifiern ')' ]   ..then as now...
```
with an optional modifier list in parens (and actually similarly after any keyword) would work just fine woudn't it?

arybczak:
Could we also restrict them to begin with a capital letter? One thing less to discuss when adding new modifiers.

goldfirere:
I think this is ready to submit to the committee @nomeata

serras:
This seems a natural extension of #370 to handle some additional usages of modifiers we found along the way. My recommendation is acceptance.

proposals/0111-linear-types.rst
```diff
@@ -297,40 +297,21 @@ This proposal adds two new syntactical constructs:
...
 
-  This modifies the syntax entry for pattern with signature annotation
-  as follows as follows
+  This modifies the syntax entry for lambda expressions
+  as follows
 
   ::
 
-    pat -> pat [PREFIX_PERCENT btype] [:: type]
+    lexp -> \ apat1 ... apatn [PREFIX_PERCENT btype] -> exp
```
aspiwack, line 305: I find this change surprising. I thought the direction we were taking was to be able to specify, for each pattern in a lambda-expression to be an arg-type as we've been calling it (that is, everything that goes to the left of an arrow).

But here we need several lambda-expressions if we have several multiplicities to bind. Maybe it's the right thing to do? I suppose linearity annotations on nested patterns don't make a ton of sense. Also it precludes giving explicit multiplicities in equations. Usually the multiplicities come from the type signature, but this seems to run contrary to the variable-in-patterns story and the idea of not relying on the `forall p.` binders in type signature to bind names. So actually, linearity annotations on nested patterns do make sense: they let me bind multiplicity variables.

At the very least I'd like some motivation. Maybe a paragraph in the _Alternatives_ section (“a previous version of this proposal had <previous syntax> but <justification>” would help?

proposals/0111-linear-types.rst
```diff
@@ -297,40 +297,21 @@ This proposal adds two new syntactical constructs:
...
-  The form is however permitted in records (see `Records`_ below)
-
-  ::
-
-    data R = R { unrestrictedField %'Many :: A, linearField %'One :: B }
-
-  This modifies the field declaration syntax to
-
-  ::
+  This form is allowed only when the lambda binds precisely one variable.
```
aspiwack, line 310: Why? This sounds arbitrary.

aspiwack (commented):
I have questions about the change on the lambda-syntax.

proposals/0111-linear-types.rst
goldfirere: My main motivation is so that modifiers always precede the thing they modify... and this new version seemed simpler, because I didn't think that a multiplicity annotation on a nested pattern made any sense. But you're right that a multiplicity annotation on nested patterns might indeed make sense.

What would you think of

```
pat -> [PREFIX_PERCENT btype] pat [:: type]
```

? This would mean we write ` \ (%1 x) -> x` or `\ (%1 x :: a) -> (x :: a)`. Not so different from Idris, actually.

proposals/0111-linear-types.rst
goldfirere: I think I didn't want the multiplicity to apply to all bound variables / patterns, but it seems, in retrospect, unnecessary.

proposals/0111-linear-types.rst
serras: I've had assumed that the modifier is on the _arrow_ (since the modifier always precedes the modified thing). This is why `x %1 -> ...` made sense to me. In fact, I find a bit weird to allow other patterns, I would expect having to write the following for a linear function that matches:

```
\x %1 -> case x of ...
```

aspiwack: After further thought there is a difficulty with using multiplicity modifier to bind variables in patterns: the current specification would not let me infer a modifier's type. If I wrote (in the notation you are proposing)

```haskell
\(%p x) -> u
```

Then GHC would go “you naughty, naughty person, I don't know what kind of modifier this is”. And it would be quite right too.

So maybe modifiers are not the right way to do multiplicity variable binding…

I'm honestly not quite sure what conclusion to draw from this.

goldfirere: Is `p` in scope? Ah, no: I see you want to *bind* `p`. No dice. I say: we just don't allow this. Use the syntax from #155/#238 to do that.

aspiwack: Ah yes. I got confused. I was thinking: I need multiplication annotation to bind existentially-bound multiplicity variables. But it's just not true: if they are existentially bound, then they are an argument of a constructor, and we can bind it at this site.

Nevermind this objection, then.

Then probably I like `\ (%1 x) -> x`. Maybe it makes more sense as part of the lambda syntax? Rather than nested in patterns. It's a bit fuzzy in my mind right now.

goldfirere: My goal here is to have a consistent way to understand `%blah` in Haskell source code, and that is for the `%blah` to affect the next thing in the source. I don't have a strong sense for the best syntax in this spot, so happy to follow any suggestion of yours as long as the `%1` precedes something reasonably understood to be part of the lambda.

aspiwack: I have difficulties paging all this back in active memory. But maybe it doesn't matter: the design principle is now laid out, and if we find the need to adjust in the future, we can, we don't have to think about everything right now. (though we have to admit that record fields are a bit inconsistent, ah well…)

My favoured syntax for this particular case is `\ (%1 x) ->`, it's probably sufficient. Maybe not, but probably. So I suggest we go for it.

goldfirere:
OK. I've updated the proposal to satisfy @aspiwack's hopes and dreams. Off to you @serras.

simonpj:
I think I'm happy with this.  I like
* The consistent prefix form
* That modifiers are types

Things I am bothered about
* It seems that you can only have *one* modifier on a type, or declaration.  Why one?  I suggest allowing zero or more modifiers, with optional semicolons separating them (to allow them to be on a new line in declaration syntax.

Yes, let's make sure this is all consistent with #390

goldfirere:
I've updated this to allow semicolons between modifiers and the declarations they modify.

@simonpj The syntax in the proposal allows for multiple modifiers, through the braces convention as used in the Haskell Report. This is also signposted in point 2 of the specification.

The update does not address the specific ideas above, but I do think the syntax encompasses these possibilities, to be pursued in other proposals.

Ericson2314:
Inspired by https://github.com/ghc-proposals/ghc-proposals/pull/512, https://mail.haskell.org/pipermail/ghc-steering-committee/2022-December/003026.html is rethinking modifiers a bit.

I would want to just throw out that https://gitlab.haskell.org/ghc/ghc/-/wikis/annotations isn't mentioned in the modifiers RFC, and by virtual of being a declaration not expression oriented syntax might be a bit easier to work with? https://doc.rust-lang.org/reference/attributes.html from Rust has also proved to be very nice in practice.

PR #393, opened 2021-01-08
No `forall ->` in data constructors

goldfirere:
This strikes a small piece of #81 that was never implemented and deserves more careful study than this proposal gives it.

### DIFF
--- a/proposals/0081-forall-arrow.rst
+++ b/proposals/0081-forall-arrow.rst
@@ -105,9 +105,10 @@ of rejected constructs):
 
     isJust (x :: forall a -> Maybe a) = ...
 
-A data constructor *can* use ``forall ... ->`` in its type (as given in
-GADT-syntax) or arguments, but any use of such a constructor in terms (as
-opposed to in a type) will be an error.
+* A GADT data constructor type::
+
+    data T where
+      MkT :: forall a -> a -> T
 
 Naturally, the new syntax is forbidden anywhere that ``forall`` is currently
 forbidden (for example, in an argument position of a type family).


monoidal:
Is this still active, considering #402?

goldfirere:
I think so, in that we're moving this idea to a different proposal.

@nomeata, does this need full committee action, or can we just pull the trigger? (I myself am split on this.) In any case, I'd like to move this forward.

simonpj:
I think we can just pull the trigger.  It's very much a dark corner.   Any objections from anyone?

int-index:
My problem with this amendment is that it explicitly disallows `forall ->` in GADT constructors, while #402 allows it. So I think they are mutually exclusive. I think we should merge this amendment if #402 is rejected, but discard the amendment if #402 is accepted.

simonpj:
This amendment is for #81, which is about kinds.  It says

> This new construct will be rejected in any context that is unambiguously a type for a term. 

It previously had an ill-thought-out bit about data constructors that are used only in types (in promoted form). This edit conservatively removes that bit.

Nothing stops us adding `forall a -> ` to the types of terms and data constructors later.  Until we do that, #402 might still be what we want but we should take the `forall a-> ` part out until we do.

TL;DR: I see no conflict here.

int-index:
But a GADT constructor signature is not "unambiguously a type for a term", as when the constructor is promoted, the signature specifies its kind, where `forall ->` is allowed, today.

> ill-thought-out bit about data constructors 

It was not ill-thought-out, it was insufficiently thought out, and #402 provides the missing bits (nested foralls in constructor signatures)

nomeata:
> @nomeata, does this need full committee action, or can we just pull the trigger? (I myself am split on this.) In any case, I'd like to move this forward.

I can’t tell. Simon PJ says go ahead, so maybe do? :-)

simonpj:
For this particular edit, I think we can just go ahead.

PR #396, opened 2021-01-23
Defaulting plugins

abarbu:
[Rendered](https://github.com/abarbu/ghc-proposals/blob/master/proposals/0000-defaulting-plugins.rst)

There's a lot of interest in better defaulting, but no agreement on what to do. Let's move forward in a lightweight way with defaulting plugins. Easy to implement and doesn't constrain what we can do in the future.

### DIFF
--- a/proposals/0000-defaulting-plugins.rst
+++ b/proposals/0000-defaulting-plugins.rst
@@ -6,9 +6,7 @@ Defaulting plugins
 .. ticket-url:: 
 .. implemented:: 
 .. highlight:: haskell
-.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
-            **After creating the pull request, edit this file again, update the
-            number in the link, and delete this bold sentence.**
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/396>`_.
 .. contents::
 
 Defaulting has many useful applications, and is particularly critical for large
--- a/proposals/0000-defaulting-plugins.rst
+++ b/proposals/0000-defaulting-plugins.rst
@@ -1,0 +1,209 @@
+Defaulting plugins
+==================
+
+.. author:: Andrei Barbu
+.. date-accepted:: 
+.. ticket-url:: 
+.. implemented:: 
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. contents::
+
+Defaulting has many useful applications, and is particularly critical for large
+APIs with many type parameters like those used for machine learning or
+games. Instead of burdening the language with syntax that might be too divisive
+or limited, we propose a defaulting plugin extension. Like type plugins,
+defaulting plugins run while type checking. They offer up new hypotheses before
+GHC declares a type to be ambiguous.
+
+Motivation
+----------
+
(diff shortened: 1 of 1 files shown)


goldfirere:
It has been a busy stretch for me, and so I apologize for not reading this before now. This is a thoughtful proposal, and it even comes with an implementation. I'm in support.

If I were you, I'd just go ahead and submit this to the committee for consideration. Silence does not mean dissent! :)

abarbu:
Thanks for giving it a read! Wasn't sure if the silence was a bad sign.

@nomeata I'd like to submit this for consideration.

int-index:
Hello. Thank you for the well-written proposal. Good news: I don’t think it needs committee’s approval. The change falls outside the GHC Proposals process. Here’s an excerpt from the README:

> Changes to the GHC API or the plugin API are not automatically within the scope of the committee, and can be contributed following the usual GHC workflow

Overall, I like the idea. However, I haven’t implemented any GHC plugins myself, so I’m not very familiar with the plugins API. It would be great if other developers more familiar with this part of GHC API could comment on low-level details. I believe the right platform for this discussion would be either the `ghc-devs` mailing list or GitLab comments (when you submit a merge request).

@abarbu How does that sound? Let me know if you want this to be handled by the committee anyway.

abarbu:
@int-index I'm happy to follow whatever the correct procedure is. I thought review was necessary since the source plugins proposal went through review.

I'm happy to just submit a merge request on GitLab if that's the better way forward. Whatever you prefer.

int-index:
> I thought review was necessary since the source plugins proposal went through review.

The “Source Plugins” proposal was written in Jan 2018, and the sentence I quoted was added to the README the next month. These were the early days of `ghc-proposals` and best practices were still in the process of getting established.

@nomeata Do you perhaps remember what compelled you to add https://github.com/ghc-proposals/ghc-proposals/commit/ea57828a1368657eb8bf8626fe80a398ac75d20c?

Also, hole fit plugins (https://github.com/ghc/ghc/commit/c311277bf640a4aeb929f3080eaaf656c0e0611c) seem to have been added without going through the GHC Proposals process, or at least I cannot locate the proposal for them.

> I can just submit a merge request on GitLab if that's the better way forward.

You’ll need to do that anyway to get your code merged upstream. And since you already have an implementation, I see no reason to hesitate. Go for it!

From reading the patch that you currently have, I understand that your goal is to add a way to run user-written code inside `applyDefaultingRules`. Seems like a very reasonable addition to me.

On the other hand, I’m less certain about the exact definitions that you have for `FillDefaulting`, `DefaultingPlugin`, and `DefaultingPluginResult`. And I surmise that you are much more likely to get concrete technical feedback of this sort on GitLab.

nomeata:
> @nomeata Do you perhaps remember what compelled you to add ea57828?

I think to me, the GHC API isn't really part of the “language” implemented by GHC, and just like base library changes are not in scope this process (which arguably have much more severe implications!), `ghc` library changes need not be.

And I still think it makes sense for changes to that API to start on GitLab, with the GHC devs of course having the authority to accept or reject – or, if they think it’s deeded, defer to the commitee.

nomeata:
I’ll remove `Pending shepherd recommendation` for now, assuming that this can indeed be discussed and merged into GHC without official committee involvement. Feel free to use this space, of course.

goldfirere:
To be clear, I think a good next step for our intrepid @abarbu is to submit their implementation on GitLab for review. We don't really have a forum for discussing these sorts of changes beyond GitLab, but I don't want our "out-of-scope" label to be too discouraging.

Put another way: I want to see this (broadly -- it's possible I will disagree in some detail later) in GHC, and so I'd like to make an avenue for that to happen.

goldfirere:
I think this ticket can now be closed.

PR #397, opened 2021-01-23
Clarify renomination process bits

goldfirere:
This is a small clarification PR. The only substantive change is that it allows any committee member to play secretary when the secretary's term expires. I expect this to be wholly uncontroversial. If it looks good (and you agree that this is uncontroversial), then I recommend any member of the committee just merge.


gridaphobe (approved):
makes sense to me!

nomeata:
As the secretary who is not a chair, I ensorse this message.

PR #399, opened 2021-02-03
Type directed COMPLETE pragmas

cgibbard:
[Rendered](https://github.com/ghc-proposals/ghc-proposals/blob/9ba3c5192e8a2ba1be633086246c6926cfb1ae47/proposals/0000-type-directed-complete-pragmas.rst)

### DIFF
--- a/proposals/0000-type-directed-complete-pragmas.rst
+++ b/proposals/0000-type-directed-complete-pragmas.rst
@@ -1,0 +1,136 @@
+Type-directed COMPLETE pragmas
+==============================
+
+.. author:: Cale Gibbard and Dan Bornside on behalf of Obsidian Systems and MIRI
+.. date-accepted:: Leave blank. This will be filled in when the proposal is accepted.
+.. ticket-url:: Leave blank. This will eventually be filled with the
+                ticket URL which will track the progress of the
+                implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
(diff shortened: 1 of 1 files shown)


sgraf812:
Just before opening #400, I had look around and found this proposal. If only I had found it *before* writing up mine!

I think your end goal is the same, our examples even have the same general theme ("lists" + subtypes).

I like my approach of specifying a Constraint a bit more, because it allows to write *one* COMPLETE set for `AsEmpty` and *one* for `AsCons` and all instances will automatically profit from the COMPLETE set. I don't see how that's easily possible with your proposal. You could allow `{-# COMPLETE Empty, Cons :: AsCons a as => as #-}`, I guess, but I find the analogy to type signatures flawed. Just so that you don't have to grep through my proposal in order to imagine what your code would look like, here it is:

```hs
{-# COMPLETE[forall as. AsEmpty as] Empty #-}
{-# COMPLETE[forall as. AsCons a as] Empty, Cons #-}
```

Urgh, I just realised that you want to single out `[a]` from the first `COMPLETE` set. Well, then better model that through a subclass relationship, so you have a constraint that `Proxy a` satisfies, but `[a]` doesn't. Alternatively, if you insist on the first signature only applying to `[a]`, then write

```hs
{-# COMPLETE[as. (as ~ Proxy a)] Empty #-}
```

But `as` is not really a type constructor now. Sigh. I'm beginning to see why we might actually want to have `cts => T a b c` syntax.

Anyway, I'm quite keen on having a good story for constraints, which allows COMPLETE sets to be bundled with type class instances.

cgibbard:
Just leaving a comment here that we had a go at implementing this: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/5095

PR #401, opened 2021-02-08
RecursiveLet

ocharles:
This is a proposal to allow Haskell developers to distinguish bindings that are self-referentially recursive from non-recursive bindings through the new `RecursiveLet`/`NoRecursiveLet` language extensions.

[Rendered](https://github.com/ocharles/ghc-proposals/blob/letrec/proposals/0000-letrec.md)

### DIFF
--- a/proposals/0000-letrec.md
+++ b/proposals/0000-letrec.md
@@ -1,0 +1,116 @@
+---
+author: Ollie Charles
+date-accepted: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/0>).
+
+# `(No)RecursiveLet`
+
+This proposal changes GHC to distinguish between (a subset of) recursive and non-recursive `let`-bindings with the new `RecursiveLet` extension. When enabled, `let` bindings can refer to themselves by name (this is the current behavior), but when this extension is turned off (with `NoRecursiveLet`), recursive let bindings are only allowed if the author uses `let rec`.
+
+## Motivation
+
+Currently, whenever we write `let` bindings in Haskell we gain the ability to write recursive definitions - a definition can refer to itself. Such definitions are obviously useful, but this power is not always needed. The cost of accidentally using this "power" is significant - a lapse of concentration while writing Haskell can cause a program to be accepted that at runtime simply spins and fails to terminate.
+
+These concerns are not theoretical. This proposal was ultimately motivated by the author making a seemingly innocent change. A small `let` binding began as:
+
+```haskell
+let panelisation = flipPanelisation unreflectedPanelisation
+```
+
+Later, requirements changed such that we should only call `flipPanelisation` if the `side` variable takes the value `Bot`. Noting this, we go ahead and add a guard and go off on our merry way:
+
+```haskell
+let panelisation
+      | side == Bot = flipPanelisation unreflectedPanelisation
+      | otherwise    = panelisation
+```
+
+Can you see the problem? The author certainly couldn't! This program compiles fine, and was happily deployed. This innocent change resulted in a partial outage later, as a batch-processing daemon had got stuck and a server was pinned at 100% CPU usage.
+
+The problem is that when `side /= Bot`, we actually have:
+
+```haskell
+let panelisation = panesilation
+```
+
+This was not intended at all! The intended code was:
+
+```haskell
+let panelisation = unreflectedPanelisation
+```
+
+But that's not what was written. The author of the change didn't notice, the compiler didn't notice, and a colleague didn't notice when they did a code review.
+
+With this proposal, if the author had disabled recursive let bindings with `-XNoRecursiveLet`, the definition would have been rejected. The options would be to change `let` to `let rec`, or to pause, reflect, and hopefully write the intended code.
+
+## Proposed Change Specification
+
+* A new language extension is added: `RecursiveLet`. This can be enabled or disabled via the normal mechanisims for extensions. By default the extension is enabled.
+
+* When `RecursiveLet` is _enabled_ (the default, or explicitly with `{-# language RecursiveLet #-}`), there are no changes.
+
+* When `RecursiveLet` is _disabled_ (`{-# language NoRecursiveLet -#}`), compilation of `let` expressions changes. With `NoRecursiveLet`, group of `let`-bound bindings cannot refer to themselves, either directly or mutually. For example, with `NoRecursiveLet`, the following will now fail to compile:
+
+  ```
+  let iterate f x = x : iterate f (f x) in ...
+  ```
+
+* If the user does want to write such a definition, they need use `let rec` when they create their bindings, rather than just `let`:
+
+  ```
+  let rec iterate f x = x : iterate f (f x) in ...
+  ```
+
+  By adding `rec` after `let`, all bindings in this `let` expression (or statement, if `let rec` is used within `do`) can refer to each other, enabling self-recursive definitions and mutually recursive definitions.
+
+## Effect and Interactions
+
+One major interaction with this proposal is that it means that `rec` cannot be used as an identifier name. Currently, the following is accepted by GHC:
+
+```
+let rec x = x in rec True
+```
+
+Here `rec` is just the identity function, albeit with a terribly misleading name. With `NoRecursiveLet`, this program will fail to compile, as the body of the `let` expression doesn't have a variable called `rec` in scope - `rec` is now a keyword, so this `let rec` expression actually binds `x = x` (a recursive, but permitted, definition).
+
+## Costs and Drawbacks
+
+This change appears to be fairly low cost (though this is unverified):
+
+* The change to GHC should be localised to the renamer
+
+* As the extension is turned on by default, all existing code continues to compile.
+
+In terms of drawbacks:
+
+* With `NoRecursiveLet`, a `let` binding can no longer bind the name `rec`, as this becomes a keyword.
+
+* This extension is intentionally simple, and there are many classes of recursion it does _not_ catch. It is intended to catch accidentally self-recursive definitions, but does not rule out non-termination by other means (e.g., calling `fix`).
+
+## Alternatives
+
+### Add a new `letrec` keyword
+
+This proposal currently tries to avoid adding new syntax, but an alternative could be to use `letrec`. This would allow `rec` to be bound as an identifier, as `rec` would no longer parse as a keyword. The downside is that the lexical structure of Haskell increases.
+
+### Use Warnings Instead of a Language Extension
+
+Another suggestion when this propsal was being drafted was to do this with warnings. A "warn about recursive let bindings" warning (name TBC) would be added, and would fire whenever a `let` bound name referred to itself. There remains an open question about what to do with legitimate uses of recursive let bindings, though. That is, if we have `let ones = 1 : ones` and we want this, how do we silence the warning?
+
+## Unresolved Questions
+
+* Should this extend to `where` clauses, too?
+
+## Implementation Plan
+
+The author of this proposal (Ollie Charles) offers to do the work to get this proposal into GHC.
+
+## Endorsements
+
+While not a particularly accurate number, a subset of the community resonated to the frustration of introducing accidentally recursive `let` bindings in this tweet:
+
+https://twitter.com/acid2/status/1346851203599249413


ocharles:
I'm kind of throwing this one over the fence as I lost steam towards the end of it... hopefully that's OK. I know people are interested in seeing *something* happen, so hopefully this is a good way to get a semi-formal conversation started.

Ericson2314:
Hmm, I think I could get behind this. If nothing else, it's good as a stepping-stone for teaching. I would also mention recursive do as precedent for the `rec` keyword.

I feel like it should also apply to `where` clauses too, as otherwise this is a rather ineffective guardrail for beginners. Maybe if we can have multiple `where`s like `let .. in let ... in let ... in` (but reversed?) it would be a less draconian restriction.

int-index:
It’s good to be explicit about recursion. I would also suggest a compiler warning that identifiers whether a particular let binding can be moved out of the recursive group:

```
let rec
  a = ... b ...
  b = ... a ...
  c = ... no mention of a or b
in
  ...
```

Then it would suggest to define `c` outside of `let rec`.

amesgen:
This would also enable to opt-in into shadowing via let bindings. I really like this in [Rust](https://doc.rust-lang.org/book/ch03-01-variables-and-mutability.html#shadowing).

Right now, one can do
```haskell
doStuff foo = do
  foo <- preprocess foo
  -- now the "old" foo can't get used accidentally
```
but not
```haskell
doStuff foo = do
  let foo = preprocess foo -- this loops!
  -- somewhat ugly workaround:
  -- foo <- pure $ preprocess foo
```

goldfirere:
I'm in strong support of this direction, but I have a few small suggestions:

* It's a bit unclear to me what the intended scoping rules for `let` is under the new syntax. Which of the following would be allowed under `-XNoRecursiveLet`?

```hs
let x = x + 1 in ....   -- presumably no
let x = y + 1; y = 2 in ....   -- maybe?
let y = 2; x = y + 1 in ....   -- maybe?
let x = 1; x = x + 1 in ....   -- gee, I hope not
let x = 1 in let x = x + 1 in ....  -- this one might be more sensible, but then enabling -XRecursiveLet would break this
```

Note that the last example seems to be what @amesgen is asking for.

* I definitely think the new treatment should apply to `where` clauses.

* The specification should make clear that the layout context of a `let rec` starts at the first non-comment, non-whitespace character after the `rec`.

* Should the use of `rec` as an identifier cause a warning with `-Wcompat`? Not sure.

Thanks for writing this up!

Ericson2314:
Talking to @danbornside, it would also be nice to be able to get evidence from GADT matches out. For example:
```haskell
let Refl = ... in {-# use equality #-}
```
(Neither of us totally understands why we don't get the evidence after the `in`, even if we do understand why out of ease and caution no other definition's RHS gets the evidence is today. So maybe we are missing some important detail.)

Basically, I think that with this change one should no longer need the current workarounds of:
 - Single alternative case
 - Do notation bind with pure / Identity

for any purpose, be that shadowing, evidence-gathering, or anything else.

simonpj:
But `let` is lazy.  Considre
```
let Refl = error "urk" :: Int ~ Bool in ....{- use equality Int = Bool -}...
```
I'm sure we don't want that.   If you want equality you need strictness.

It woudl be OK to support this if you use a bang:
```
let !Refl = ... in ...{- use equalty -} ...
```
because that's precisely the same as
```
case .... of Ref -> ...{- use equality -}...
```

Ericson2314:
Oh right...duh. Yes would need a `!`, and making the evidence scoping depend on the `!` is quite arguably bad taste.

ndmitchell:
I think that `letrec` as a keyword is preferable to `let rec`, which is really a "two word" keyword, which is a bit grim. It also makes layout complex, since what does the layout anchor to. A single keyword solves all that.

I think a strong argument in favour of this proposal is that it lets us reflect GHC Core more precisely into Haskell. I've always found that distinction a bit unpleasant.

> I definitely think the new treatment should apply to where clauses.

I disagree - personally I mean `letrec` about once every few years, but write recursive functions under a `where` pretty regularly. If this treatment applies to where, what does that mean? Where would the `rec` go? I would hope we're treating everything that comes after a `module X where` as "underneath a where" too, meaning we have to annotate all recursive functions? That seems very unlike Haskell.

phadej:
Re `let` or `letrec`

For recursive do we have both, new keyword `mdo` and `rec` blocks.

```haskell
justOnes = mdo { xs <- Just (1:xs)
               ; return (map negate xs) }
```

or

```
justOnes = do { rec { xs <- Just (1:xs) }
              ; return (map negate xs) }
```

So I think that

```haskell
let xs = True : ys; ys = False : xs in xs
[True,False,True,False,True,False,True,False,True,False]
```

```haskell
take 10 $ let { xs = True : ys; ys = False : xs } in xs
[True,False,True,False,True,False,True,False,True,False]
```

could be written in both ways

```haskell
take 10 $ letrec { xs = True : ys; ys = False : xs } in xs
```

or

```haskell
take 10 $ let { rec { xs = True : ys; ys = False : xs } } in xs
```

where `rec` allows more granularity of recusive blocks i.e.

```haskell
let rec xs = True : ys
        ys = False : xs

    zs  = map not xs
in ...
```

When this won't work:

```haskell
let zs = map not xs
    rec xs = True : ys
        ys = False : xs
in ...
```

But single keyword `letrec` would:

```
letrec zs = map not xs
       rec xs = True : ys
           ys = False : xs
in ...
```

---

so

```haskell
let x1 = ...
    x2 = ... x1 ...
    rec x3 = ... x2 ... x3 ...
    x4 = ... x3 ...
    rec x5 = ... x4 ... x5 ...
in ...
```

---

The `letrec` and `let rec` would be confusing; latter is more expressive,
i'd use that.

goldfirere:
I would be disappointed if this new idea didn't extend to `where` clauses, which I use *much* more often than `let` expressions.

----------------------------

I'm in favor of @phadej's reinterpretation. Concretely:

```
expr ::=
    'let' '{' let_decls '}' 'in' expr
  | ...
let_decls ::= {- empty -} | let_decl ';' let_decls
let_decl ::=
    type_signature
  | function_decl
  | ...
  | 'rec' '{' let_decls '}'
function_decl ::= ... function_equation ...
function_equation ::= ... where_clause
where_clause ::= {- empty -} | 'where' '{' let_decls '}'
```

It is understood that the lexer inserts braces and semicolons according to the layout rule, where `rec` is a layout herald. (Does this pose problems with its existing use in `mdo`? Not sure. I don't know `mdo` offhand.)

Open questions:
1. Do we allow later declarations in a `let` (without the `rec`) refer to earlier ones?
2. Do we allow later declarations in a `where` (without the `rec`) refer to earlier ones? Or do we perhaps reverse the direction in a `where`?
3. What to do at top-level? Maybe we want `rec` there, too! With this new design, such an idea is very reasonable.

gridaphobe:
Thank you for writing this up @ocharles! I've had it in the back of my mind to propose this extension for quite a while. I eventually settled on the same variant that @phadej and @goldfirere propose, where `rec` becomes a new keyword that introduces a recursive binder group.

One issue that I think we need to address (part of the reason I never wrote up the proposal myself) is that this extension introduces a fork in the language. We have a few such extensions already (most notably `Strict` and `StrictData`), but my impression is that they would not make it through the Committee today. What do we do about that? Perhaps the argument is that implicitly-recursive `let` was a mistake and should be removed, and this extension provides a migration path? That would be a huge change to the language though..

My answers to @goldfirere's open questions would be

> Do we allow later declarations in a let (without the rec) refer to earlier ones?

No, plain `let` should behave like Scheme's `let*`, i.e. `let { x = 1; x = x + 1 in e }` is equivalent to `let x = 1 in let x = x + 1 in e`.

> Do we allow later declarations in a where (without the rec) refer to earlier ones? Or do we perhaps reverse the direction in a where?

I'm somewhat ambivalent about `where`. We should keep in mind that `let` and `where` are technically redundant, and yet `where` is one of the best additions to Haskell's syntax! So while consistency would favor making them behave similarly with respect to `rec`, what we really need to consider is how people use `where`.

> What to do at top-level? Maybe we want rec there, too! With this new design, such an idea is very reasonable.

It would be reasonable, and simple and consistent, to have `rec` for top-level binders as well, but I think it would be a mistake. I tend to structure my top-level definitions in a somewhat topological order, so you have the main entrypoints towards the top of the module and helper/utility functions towards the bottom. If we apply `NoRecursiveLet` to the top-level, this would no longer work, and I'd have to enclose the entire module in a `rec` block, with needless indentation. Unless maybe you're suggesting we might start writing

```haskell
module Foo where rec

...
```

Ericson2314:
I am not so sure about the `let* semantics`. Implicit sequentially feels antithetical to what Haskell is about. I also agree the reverse order for where and the top level is very common.

I am not sure what is right, I was thinking multiple `where` to support the reverse order was nice above (and perhaps I wasn't clear enough about that), but not everyone does the reverse order and this would enforce it.

Anything that breaks the `let` and `where` symmetry of groups being handled the same makes me very wary. Maybe they are used differently, idiomatically, but beginners don't know that. The `rec` difference is meaningful and prevents mistakes, but a `let`-`where` difference would be unsemantic complexity that's just a burden to learn.

int-index:
> Maybe they are used differently, idiomatically, but beginners don't know that. 

I still choose which one to use at random.

proposals/0000-letrec.md
```diff
@@ -0,0 +1,116 @@
...
+      | side == Bot = flipPanelisation unreflectedPanelisation
+      | otherwise    = panelisation
+```
+
+Can you see the problem? The author certainly couldn't! This program compiles fine, and was happily deployed. This innocent change resulted in a partial outage later, as a batch-processing daemon had got stuck and a server was pinned at 100% CPU usage.
+
+The problem is that when `side /= Bot`, we actually have:
+
+```haskell
+let panelisation = panesilation
```
JakobBruenker, line 37:
```suggestion
let panelisation = panelisation
```

JakobBruenker (changes requested):

klapaucius:
alternative: let-no-rec as in Clean (sans syntax)
```
writeABC file
  # file = fwritec 'a' file
    file = fwritec 'b' file
  = fwritec 'c' file
```

JakobBruenker:
> Do we allow later declarations in a let (without the rec) refer to earlier ones?

To me it would feel a lot more Haskell-like to allow this. It would also minimize the difference between `let` with and without `rec` (to whether or not recursion is allowed), which seems like a good thing to me from an understandability and refactorability perspective, and doing the same with `where` would make the question of whether or not to handle the order in the same way obsolete.

Kleidukos:
Thanks for this proposal @ocharles !!
I see this as a beautiful opportunity to unify the FP community around something that OCaml has championed and which was quite successful, both in terms of syntax (`let rec`) and practices (explicitly enabling recursion in let bindings).

ethercrow:
> This would also enable to opt-in into shadowing via let bindings. I really like this in [Rust](https://doc.rust-lang.org/book/ch03-01-variables-and-mutability.html#shadowing).
> 
> Right now, one can do
> 
> ```haskell
> doStuff foo = do
>   foo <- preprocess foo
>   -- now the "old" foo can't get used accidentally
> ```
> 
> but not
> 
> ```haskell
> doStuff foo = do
>   let foo = preprocess foo -- this loops!
>   -- somewhat ugly workaround:
>   -- foo <- pure $ preprocess foo
> ```

In this situation I'd use ViewPatterns:

```
> ```haskell
> doStuff (preprocess -> foo) = do
>   ...
> ```
```

amesgen:
@ethercrow Yes, I also use that often, but it is not always the case that `preprocess` is actually just a short function name (or rather part of a lengthy computation involving other bindings before the "redefinition" of `foo`), or that `foo` is explicitly introduced, e.g.

```haskell
data Foo = Foo {foo :: Int, bar :: Text}
doStuff Foo{..} = do
  let foo = preprocess foo -- this loops!
```

Either way, this is only a minor point IMO.

ethercrow:
Current behavior causes me to lose on the order of 10 minutes roughly once a year when I introduce a loop by accident.

If this PR is implemented I expect to lose much more time in aggregate:

1. Hundreds of times a year I'd have to notice that a definition became recursive and go and insert `rec` nearby.
2. When copy-pasting a fragment of code with a `let` binding from some other file or project I'd have to think about which "rec-ness" it used and if I have to adjust it.
3. When opening an unfamiliar project I'd sometimes have to figure out which "rec-ness" it uses via pragmas in a file or maybe even in a cabal file.

Overall personally I don't think it's worth it.

Kleidukos:
> Hundreds of times a year I'd have to notice that a definition became recursive and go and insert rec nearby.

Only if you enable this extension, though? 

> When copy-pasting a fragment of code with a let binding from some other file or project I'd have to think about which "rec-ness" it used and if I have to adjust it.

Everyone has to submit their soul to the Gods of Copypasta, you are no exception. :)

> When opening an unfamiliar project I'd sometimes have to figure out which "rec-ness" it uses via pragmas in a file or maybe even in a cabal file.

Exploration of a new project is a big task in itself and I think it would be foolish to try think we can avoid making it tedious at all. 

@ethercrow  Exploring a new codebase is also exploring the psyche of the various people who contributed to it. I don't understand how it is different from any other extension that has been proposed here.

ethercrow:
> I don't understand how it is different from any other extension that has been proposed here.

Most other extensions introduce something new, not change how something fundamental like `let` works.

```
main = do
  let fib n = 42
  print $ fib  5
  let fib n = if n < 2 then 1 else fib (n - 1) + fib (n - 2)
  print $ fib  5
```

prints 42 and 8. But with "NoRecursiveLet" and shadowing like proposed in this thread the same piece of code is still valid, but means something completely different and prints 42 and 84 now. This scares me.

ndmitchell:
I get @phadej's arguments for a separate `rec` keyword - and they makes sense - so that seems the right way to go.

> I would be disappointed if this new idea didn't extend to where clauses, which I use much more often than let expressions.

I would be disappointed if it _did_ extend to where clauses (and probably just not use it - despite having been one of the people who persuaded Ollie to write this up). I _also_ use where clauses much more often than let expressions, and regularly write both recursive and non-recursive where clauses. And yet, based on my personal experience and Ollie's motivating example, it's usually a `let` that causes the blackhole, and I approximately never mean to write a recursive let. Making `where` clauses require explicit `rec` without doing it to top-level definitions is inconsistent. Annotating recursive functions with `rec` makes it less natural to write recursive functions and blows off the alignment between function and type.

phadej:
> I would be disappointed if it did extend to where clauses (and probably just not use it -

The proposed extension is `RecursiveLet`, there could be `RecursiveWhere` too.

I"m strongly against having unifying `RecursiveBindings`, `let` and `where` are used differently by people.

ocharles:
Just a quick update from me - I haven't managed to read a single reply here yet. I'm of course happy for the discussion to continue, though! I'll hopefully catch up next week.

goldfirere:
Aha! https://github.com/ghc-proposals/ghc-proposals/pull/401#issuecomment-778347431 has betrayed your underlying assumption: you perhaps have internalized a distinction "`let` is not recursive; `where` is". Maybe this comes from growing up writing ML? (I don't know whether that's true.) On the other hand, I do not operate with this distinction -- I see `let` and `where` as wholly synonymous, modulo where the blocks occur in the syntax... and thus would find it highly peculiar to see them treated differently.

My point here isn't that "you're wrong", or something; it's that you (for some reason) have a distinction between `let` and `where`. Others (like me) may not. And so we may have a hard time agreeing to a design here, having come to the table from different starting points.

phadej:
`let` appears in `do` statement blocks. There `let` is definitely not interchangeable with `where`. It is in fact about only place where I use `let`.

```haskell
action x0 = do
   x1 <- ... x0 ...
   let x2 = ... x1 ...
   x3 <- ... x2 ...
   ...
```

So, there is `let in`, `let` and `where`:

* `let in` is an expression
* `let` is a statement
* `where` is something I don't know name for

To me these are very different.

Ericson2314:
I agree with @phadej that it's `let in`, `do let` (I'll rename to help disambiguate) and `where`, and I also just use `let` and `where`, and never `let in`. But like @rae I see `let in` and `where` as interchangeable --- indeed that's why I always use `where` instead of `let in`! Likewise, I use `do let` because I need some `do`-bound variables, not because I am or am not writing recursive things.

I do agree I've been tripped up by accidental cycles with `do let` more than `where`, but that doesn't override by my strong desire to avoid making things that were the same different. Separate extensions does help a bit (one can choose to vary them together) but also feels like too many knobs. And I would hate a situation where `-XGHC2022` or whatever flipped one of them but not the other.

phadej:
To clarify. I'm not against having `where rec`. I'm not sure whether it will affect me. Probably not, I write things top to bottom.

---

An additional design choice is do what Agda does:

```haskell
foo :: [Bool]
bar :: [Bool]

foo = True : bar
bar = False : foo
```

Here the recursive group is clear. `foo` can use `bar` as its existence is declared by a type signature.

This would work in `let in` too

```haskell
let foo :: [Bool]
    bar :: [Bool]
    
    foo = True : bar
    bar = False : foo

in ... foo ... bar
```

maralorn:
I think I`d prefer `let ... in`, `do let` and `where` to all be affected. But I see the big counter argument of actually the whole file being in a `where`. So unless we apply this to top level bindings there will be an inconsistency.
Since `let ... in` and `where` really are (should be) the same. In that case I would argue for only `do let` to be affected. That would also agree a bit more with the more procedural and less declarative feel of this proposal.

Funny question what happens when you have `RecursiveDo` enabled and do
```
main = do rec let {ones = 1:ones}; print ones
```
is that allowed?

phadej:
```
main = do rec let {ones = 1:ones}; print ones
```

and

```
main = do let rec {ones = 1:ones}; print ones
```

are *clearly* different. So IMHO no.

tomjaguarpaw:
> I see the big counter argument of actually the whole file being in a `where`

I'm surprised to see it so definitively claimed that a whole module is implicitly in a `where` as opposed to a `let``. According the Report `let` and `where` both just contain *decls*. Is there some reason that the semantics makes it better to think of them as bound under a `where` rather than bound in a `let`?

* https://www.haskell.org/onlinereport/decls.html#sect4
* https://www.haskell.org/onlinereport/exps.html#sect3.12
* https://www.haskell.org/onlinereport/decls.html#sect4.4.3

maralorn:
> I'm surprised to see it so definitively claimed that a whole module is implicitly in a `where` as opposed to a `let`. 

I admit I hadn’t really looked into the report. I just observed that files start with `module ... where`.

But I assume in truth there is more than one difference between top-level bindings and where statements. (I can‘t list them all but I‘d be surprised if there weren’t any.) I think they are a lot less similar than `where` and `let ... in` in bindings.

tomjaguarpaw:
> I just observed that files start with `module ... where`

Hah, well that's a good reason! I guess that I'm so used to typing it that it has become subconscious and I didn't even notice it :)

In any case I don't think we have to treat decls in a `where` the same as top-level decls just because the latter occur under a syntactic `where` at the top of the module. I don't think Haskell would somehow be a different language if the syntax for introducing a module were `module MyModule = let <topdecls>`

phadej:
@tomjaguarpaw we have

```haskell
module Foo (foo1, foo2) where

foo1 = ...
foo2 = ...
```

The `let-in` variant would be

```haskell
let foo1 = ...
    foo2 = ...
in module Foo (foo1, foo2)
``` 

When we bikeshed about syntax, we should be precise.

phadej:
Expanding on my previous comment. `where` is not `let`.

Top-level, `module ... where ...` can contain other module imports
data, class, etc definitions, not only bindings.

I remember seeing discussions (not sure if they ever ended as a proposal)
to have local imports

```haskell
foo x = Bar.foo x
  where
    import qualified Bar
```

or even "local datatypes"

```haskell
foo x = ...
  where
    data P = P! !Int !Int
```

So it's possible `where` would get more features. I don't see it happening to `let..in` expression.

tomjaguarpaw:
Interesting. I never understood syntactic `where` in module and instance declarations to be the "same thing" as `where` in function bodies. The common style is to write the former `where`s on the same line as `module` or `instance` and the latter `where` *beneath* the function name. That probably contributed to my interpretation (misinterpretation?).

Perhaps I'm in the minority on that interpretation. In retrospect I do see the benefit of treating everything beneath a `where` in a regular manner. On the other hand I wouldn't want a superficial similarity to get in the way of a good proposal.

> I don't see it happening to `let..in` expression.

Interesting. I do wonder why not, though perhaps this is not the place to elaborate. I will think on it further.

tomjaguarpaw:
And to return to issues that are more concrete, I really do want to be able to make top-level decls non-recursive by default. One of my most frustrating bugs recently, as measured by the ratio of silliness of the cause to the severity of the impact came from a misapplied recursion. I had a tree transformation

```haskell
foo1 = \case
    Branch1 a b c -> <slow code> ... (foo1 a) ... (foo1 b) ... (foo1 c) ...
    ...
    BranchN x y z -> <slow code> ... (foo1 x) ... (foo1 y) ... (foo1 z) ...
```

I wanted to make a faster version so I copied it, renamed it `foo2` and changed some of the implementation

```haskell
foo2 = \case
    Branch1 a b c -> <fast code> ... (foo1 a) ... (foo1 b) ... (foo1 c) ...
    ...
    BranchN x y z -> <fast code> ... (foo1 x) ... (foo1 y) ... (foo1 z) ...
```

I was really surprised it didn't get faster. Why didn't it? Because although I changed the implementation I didn't change the recursive calls! The fact that the names `foo1` and `foo2` were so similar hindered me tracking down the cause.

Because of this experience I am leaning away from syntactic recursion and towards implicit recursion via a fix combinator when writing large functions with many recursive calls, at least until the design is settled. I would like the option of disabling recursion at the top level in order to nudge me towards this style.

```haskell
foo1 = fix $ \recurse -> \case
    Branch1 a b c -> <slow code> ... (recurse a) ... (recurse b) ... (recurse c) ...
    ...
    BranchN x y z -> <slow code> ... (recurse x) ... (recurse y) ... (recurse z) ...
```

klapaucius:
> I remember seeing discussions (not sure if they ever ended as a proposal)
> to have local imports

https://github.com/ghc-proposals/ghc-proposals/pull/283

> or even "local datatypes"

https://github.com/ghc-proposals/ghc-proposals/pull/273

phadej:
@klapaucius thanks!

phadej:
@tomjaguarpaw 

> And to return to issues that are more concrete,

That is why many people define `go` or `aux`, so the recursive call is not depending on the name of the top-level binder.

tomjaguarpaw:
> That is why many people define `go` or `aux`, so the recursive call is not depending on the name of the top-level binder.

Yes, that's a nice alternative to a call to `fix`, although it points more strongly to disabling top-level recursion than local recursion!

> Note: you need recursive bindings to call `foo2` from `foo2`

Indeed, therefore I wish I had the option to disable recursive bindings to nudge me away from syntactic recursion.

maralorn:
I think the answer is no, but If we were to disable syntactic recursion at top-level by default would that also apply to things like (mutually) recursive data types?

gbaz:
> Because of this experience I am leaning away from syntactic recursion and towards implicit recursion via a fix combinator when writing large functions with many recursive calls, at least until the design is settled. I would like the option of disabling recursion at the top level in order to nudge me towards this style.

As a datapoint, this is what we do in the EAQL language -- no bindings are recursive, but we have a builtin `rec` combinator (`fix` by another name) for explicit recursion. Our motivation was to steer end-users new to the language away from explicit recursion and towards more efficient builtin functions. As a bonus, this reduced implementation complexity in some regards as well...

Note however that in a Full Grown language, disabling top-level recursion also makes _mutual_ recursion more of a pain to write, as `fix` is much less pleasant in such circumstances (especially when passing through e.g. typeclass methods, etc).

danidiaz:
> One issue that I think we need to address (part of the reason I never wrote up the proposal myself) is that this extension introduces a fork in the language.

> Most other extensions introduce something new, not change how something fundamental like let works.

I'm concerned about this, too.

It could be avoided by keeping `let` as-is, and introducing an explicit modifier for non-recursive `let`s. The problem is that it would be much more pervasive; you'd end up with lots of `let norec` in normal code.

phadej:
`let norec` would make doing "wrong" thing easier then the right thing, i.e. invalidate the motivation points.

ocharles:
Ok, time to try and get on top of this!

I've tried to read all the comments slowly to understand where this discussion has gone in my absence. It seems that there was initially a bit of discussion about what the keyword is - is it `let rec` or is it `letrec`? It seems like the agreement is now on `let rec`, as originally proposed. Great! However, my proposal originally suggested that `let rec` be the keyword, but it seems now that there is more support in `rec` being something separate that you can insert "into" a group of `let` bindings to form a new group. I'm fine with that. Is there enough support now that I should amend the proposal to reflect this?

The discussions around `where` seem hairier. @phadej has [put it pretty nicely though](https://github.com/ghc-proposals/ghc-proposals/pull/401#issuecomment-778350847):

> The proposed extension is `RecursiveLet `

And indeed, it was a goal (though not explicitly stated) to _only_ try and deal with `let` for now. I [echo @ndmitchell](https://github.com/ghc-proposals/ghc-proposals/pull/401#issuecomment-778347431) in that - for whatever reason - it's only `let` where I manage to burn myself. We can always refine this idea later with subsequent proposals once we've explored a small subset of the design in the wild.

Going all the way to extend this idea to the `module` level is well beyond the scope of what I'm proposing, and does increase the risk of this extension. If we want to do that right from the get-go we're biting off a huge feature with little evidence that any of it makes sense - _nothing_ will have been tried and tested before. My preference is take smaller steps.

There have been some comments around whether a (non-recursive) `let` should be able to refer to earlier bindings. My opinion is no, they cannot. I've seen some mentions that this ordering is "not very Haskell-like", but I am not really convinced that "being Haskell-like" is necessarily good - it's an implicit assumption that we've already got things right. I argue against having non-recursive `let`s be able to refer to other bindings because it's very explainable, and as such easy to write good error messages about. It's also simpler to implement, which suggests it's easier to do it right. The moment we start having to do some basic graph analysis to detect cycles the complexity increases, both in GHC code and in the model the user has to carry in their head, and there is a risk that errors are harder to describe. Precluding these prior references for now doesn't seem like a bad decision, it's just a conservative one. Should we decide to relax this restriction in the future, all existing code will continue to work fine. It's easier to add, much harder to take away.

So to summarize, I think I should:

* Take the suggestion that `rec` can occur as a binder in a `let` block, just [as @goldfirere described](https://github.com/ghc-proposals/ghc-proposals/pull/401#issuecomment-776107158).
* Clarify that this won't apply to `where` clauses
* Clarify that this won't apply to `module` declarations
* Clarify that a non-recursive let binding cannot refer to any other bindings in its same scope.

I understand my response is quite a lot of "no, let's not do that", which might not be satisfactory. I want to  emphasize that I'm really trying to say "no, let's not do that _yet_".

How does this sound?

tomjaguarpaw:
Could we have more detail on "a non-recursive let binding cannot refer to any other bindings in its same scope"?

I'd love to enable this extension on most of my source files. However, I'm worried that the above will cause problems. For example, consider the following:

```haskell
f a b c =
  let x = g a
      y = h x b
      z = j y c
  in q z
```

Some of these bindings refer to each other (non-recursively). If I enable `NoRecursiveLet` do I have to re-write it as follows?

```haskell
{-# LANGUAGE NoRecursiveLet #-}

f a b c =
  let x = g a
  in let y = h x b
  in let z = j y c
  in q z
```

If not, then what is meant by a non-recursive let binding cannot refer to any other bindings in its same scope"? I don't get it.

If so, then I fear that I won't be able to use `NoRecursiveLet` on existing code. For new code I doubt that I will want the eyesore of so many `in`s nor will I want to give up Haskell's liberal approach to order of bindings. Topologically sorting my bindings by hand is far too much work!

Furthermore, I don't think it's easier to remove this restriction than to add it. Suppose I have written

```haskell
{-# LANGUAGE NoRecursiveLet #-}
{-# OPTIONS_GHC -Wno-name-shadowing #-}

f =
  let x = foo
  in let x = bar
         y = baz x
  in y
```

Then as per this proposal (as I understand it) `y` will be `baz foo`. If later we wanted to weaken the restriction then `y` would be `baz bar`. Am I correct in this assessment? If so then weakening the restriction would break too much code to ever be implemented. That makes me think that this restriction is *not* a conservative decision, in the sense that we do have to decide once and for all what the permanent behaviour will be.

phadej:
@ocharles @tomjaguarpaw yes, the question is here whether the `let` should be using Scheme terms just `let` or `let*`.

I.e. 

```haskell
-- assume this are in scope:
x = 1
y = 2

example =
  let x = y
      y = x
  in (x, y)
```

What is `example`?

- With recursive let it's `(_|_, _|_)`
- With `let` semantics it is `(2, 1)`
- With `let*` semantics it is `(2, 2)`

I think the `let*` is what is least surprising, though simultaneous assignment is "cool" too.

tomjaguarpaw:
Yikes! You can count me out as a user of any semantics whereby the assignments within a single let change behaviour when reordered!

phadej:
@tomjaguarpaw so you want *simultaneous binding* if `let` is non-recursive, i.e. `scheme-let`?

Yet if I understand your [previous comment](https://github.com/ghc-proposals/ghc-proposals/pull/401#issuecomment-781500760), that is not what you want?

gridaphobe:
> ```haskell
> -- assume these are in scope:
> x = 1
> y = 2
> 
> example =
>   let x = y
>       y = x
>   in (x, y)
> ```
> 
> What is `example`?
> 
> * With recursive let it's `(_|_, _|_)`
> * With `scheme-let` semantics it is `(2, 1)`
> * With `scheme-let*` semantics it is `(2, 2)`

There's a fourth option, we could reject the binding group in `example` as the definitions are mutually recursive. In other words, the semantics of non-recursive `let` could be the same as `let rec`, with the exception that recursive definitions are rejected.

tomjaguarpaw:
There are two things that I am not willing to give up. Firstly, I don't want to be forced to rewrite

```haskell
f a b c =
  let x = g a
      y = h x b
      z = j y c
  in q z
```

as

```haskell
f a b c =
  let x = g a
  in let y = h x b
  in let z = j y c
  in q z
```

The former contains no recursion so I don't see why an extension that switches off recursion by default should break it. Secondly, I don't ever want

```haskell
let ...
    x = e1
    y = e2
    ...
in ...
```

and

```haskell
let ...
    y = e2
    x = e1
    ...
in ...
```

to mean different things.

An extension could give let bindings exactly the same semantics as they currently have, with the exception that if any let bound variable ever refers to itself (possibly indirectly) in its RHS then the program is rejected. The devil may be hiding in the details of "possibly indirectly", but I don't think so at first blush. (I believe this is what @gridaphobe is suggesting in the message immediately before mine.)

gridaphobe:
> An extension could give let bindings exactly the same semantics as they currently have, with the exception that if any let bound variable ever refers to itself (possibly indirectly) in its RHS then the program is rejected. The devil may be hiding in the details of "possibly indirectly", but I don't think so at first blush. (I believe this is what @gridaphobe is suggesting in the message immediately before mine.)

Yes. Very concretely, GHC already does the dependency analysis to split `let` declarations into minimal groups of definitions, which are tagged as `NonRec` or `Rec`. The implementation of option (4) could simply be to let GHC do its usual analysis and reject any non-recursive `let` that contains `Rec` binding groups.

tomjaguarpaw:
I suspect that @ocharles (and probably others in this thread) wants to be able to write

```haskell
f =
  let x = foo x
  in let x = bar x
  in let x = baz x
  in x
```

i.e. a *change* to the meaning of existing programs, whereas I want to decrease the space of valid programs. It may well be that these two desires are irreconcilable!

Ericson2314:
@tomjaguarpaw I think the "acylic solver" could consult the outside scope to break cycles, but something so fancy has its own costs re gotchas and learnability.

I'm starting to think this proposal should just be delayed. I am curious where the local types/instances and local modules goes, to see just how similar the top level and let/where in expression can become breaking any symmetry / adding new fancy things.

phadej:
To repeat what I said. if non-recursive let would have simultaneous binding semantics (i.e. not `let*`), then I'd rather suggest we forget this idea. Scheme is the only "popular" language which has it, AFAIK, and we could do

```
let (x, y) = (y, x) in ...
```

to make simultaneous assignment if `let` were non-recursive (type-inference gotchas aside).

phadej:
However, I encourage @ocharles to "complete" the proposal, and eventually submit it to committee, so it would be their decision (to do or not whatever with `let`).

phadej:
Thinking about @gridaphobe idea

If 

```haskell
x = 1
y = 2

example =
  let x = y
      y = x
  in (x, y)
```

is rejected because it's recursive, what about

```haskell
let rec xs = True : ys
        ys = False : xs

    zs  = map not xs
in ...
```

?

It is fine, and would be accepted, because the recursive group with `xs` and `ys`
is explicitly marked so.

I think the

```haskell
let zs = map not xs

    rec xs = True : ys
        ys = False : xs
in ...
```

would be accepted then as well. Do I understand @gridaphobe variant right?

I find it a bit surprising, but it makes sense.

But I guess that's exactly something to write in proposal as an
possible alternatives and let committee decide :P

eventhelix:
I would prefer keeping `let` and `where` the same and instead introducing a new syntax for non-recursive let.  I agree with @phadej that the syntax has to be as short as or shorter than recursive let to make doing the right thing easier. How about `val` as the keyword for non-recursive let?

Something like

```haskell
measurements (Rectangle {..}) =
    val area = width * height in
    val perimeter = 2*width + 2*height in
    (area, perimeter)
```

It should support shadowing in the ML style, which would be particularly useful with linear types.

ablygo:
Expanding on @tomjaguarpaw's point about people having irreconcilable goals, I think there aren't even immediately obvious semantics for what reassignment should even mean the way we're used to defining functions in let blocks. Otherwise what would the following do?
```
let null [] = True
    null _ = False
```
If we were to take lets to reassign by default with the extension then `null = const False`, since we've simply shadowed it in the second line. Even for recursive functions, like the `fib` example, even for imperative languages users expect function declarations to allow recursion, so if the desire is to be able to have `let x = x + 1` behave more traditionally we still have't actually achieved that; we've simply made Haskell's semantics differently surprising.

But even trying to follow the principle of least surprise doesn't quite work, because while we could use syntacticly vs. non-syntacticly declared functions to distinguish (e.g. `f x y = ...` as opposed to `f = \x y -> ...`), then eta-elimination suddenly has semantic meaning. But if we try to behave consistently in all cases then we're stuck trying to justify the following meaning `fib _ = 2` in the name of avoiding bugs:
```
let fib 0 = 1
    fib 1 = 1
    fib n = fib (n - 1) + fib (n - 2)
```
My worry is that the way we declare functions in Haskell already makes the current behaviour a local optimum: whether a default of fixed point or reassignment makes sense depends on a function/value distinction that we don't really have.

Trying to do both does really feel like two proposals in one, and I think the idea works better if the goal is made more clear than simply avoiding bugs, since re-assignment by default can also hide bugs. It's the "by default", not the "recursion", that creates the bug, and changing the default doesn't fix that issue. For re-assignment, `let* x = x + 1` would avoid stealing a keyword, though it's nice that `let ` is four characters. `let x <- x + 1` would also work, though it would be weird if sequential and non-sequential declarations were intermingled.

But I think in the end people need to agree on which problem this proposal is trying to solve. Both could be done, but I think it's important to be aware we're talking about solving two separate problems.

chrisdone:
Another angle, in case people are hesitant about syntax changes: 

Have a warning `-Wrecursive-bindings`, this will warn about ALL bindings that are either self-recursive or mutually recursive in any way, in any binding group: at the top-level (this can flag up accidental mutual-recursion), `instance` bindings, `where` or `let` bindings, etc.

**USE CASE 1**: If you're experiencing an infinite loop, you turn on `-Wrecursive-bindings` and run GHC again. You eyeball the warnings and see that one particularly clever `where` or `let` is the culprit. For most people, this will probably lead them directly to the bug.

**USE CASE 2**: For more paranoid people (like myself), you turn this warning on in your .cabal file to try to avoid the bug in the first place. For such people, you have a `{-# RECURSIVE foo #-}` annotation for bindings which will permit any recursion associated with this name, because I-know-what-I'm-doing. Later, if you get an infinite loop, well then, you can see exactly which declarations are suspicious by grepping for `RECURSIVE`, and, by dropping the annotation, GHC will point out all the interesting places for you to eyeball.

We can state mutual-recursion like this:

```haskell
{-# RECURSIVE knot, tied #-}
knot = tied
tied = knot
```

We can state that `print` itself is fine to be recursive, and that `ones` is intentionally codata.
```haskell
{-# RECURSIVE print #-}
print = \case ...
  App f x -> print f <> " " <> print x
  where {-# RECURSIVE ones #-} 
             ones = 1 : ones
             this = isfine
             isfine = ones
```

A pro is that it doesn't require changing the language syntax as much as adding another pragma. Gives you space to see what can be done. Seems like there's a lot of conversation about let syntax, which seems very narrow considering the wide-range of bindings available in Haskell, and syntax discussions tend to last forever.

I think if someone had a few days spare, they could write this as a GHC plugin that would sit in the renamer stage and loudly complain to the programmer about recursive binding groups. I'd love to let loose such a plugin on my codebases and see what % of declarations it flags up.

emilypi:
That's a 👍  from me dawg.

tomjaguarpaw:
I think that @chrisdone's suggestion addresses my wishes. Great idea, thanks! (I'm not sure whether it addresses @ocharles's use case.)

cartazio:
i support an extension like this:
i've written complicated codes where i do `x <- return $! expr ` as my idiom for non recursive / `let*` (in the scheme sense ) nested non recursive / pseudo imperative shadowing

cartazio:
https://agda.readthedocs.io/en/v2.6.1/language/mutual-recursion.html?highlight=mutual#old-syntax-keyword-mutual is one approach / experiment agda has. they alternativel had a "forward declare if you wanna have them recur" approach

edmundnoble:
Putting a vote in for a `let*` aimed at intentional "reassignment" with @ablygo and @cartazio. I don't think it would fork the language and it would certainly help in writing algorithms that naturally lend themselves to being written with shadowing.

cartazio:
i think @edmundnoble articulates a *very* good point. when we say we want nonrecursive  let, dont we usually just want to be able to write let*?

AshleyYakeley:
As an alternative, consider permitting `<-` to act as a non-recursive binding. For example
```
let
a = expr1
b = expr2
c <- expr3
d = expr4
in body
```
which would be sugar for

```
let
a = expr1
b = expr2
in case expr3 of
    c -> let
        d = expr4
        in body
```
Of course, this could also be used with existential types:

```
data T = forall a. MkT a (a -> Maybe a)

let
MkT init f <- foo
in body
```

Advantages of this approach:
* backwardly compatible
* no new keywords
* uses existing `<-` in a similar way to pattern guards and `do` notation
* can mix recursive and non-recursive bindings in a `let` block

evincarofautumn:
2¢ based on reading through the comments so far:

* I like the idea of allowing `rec` *blocks* (like I use in recursive `do` and `proc`) inside `let` blocks to indicate a recursive binding group, since it reuses existing syntax for an analogous idea.

* I’m also for punting on `where`. Making `let` *different* from `where` actually seems like a good thing to me, since it provides more than an arbitrary stylistic reason to choose one or the other.

* A possible solution to the issue of ordering:
  * All bindings in a `let` are mutually in scope, as they are now
  * Dependency cycles are allowed inside `rec`
  * Cycles may not cross a `rec` boundary

So bindings can still be reordered freely:

```
let { x = A; y = B[x]; } in C
=
let { y = B[x]; x = A; } in C
```

```
      ┌───────────┐
      │           │
    ┌─▼─┐     ┌───┼───┐
x → │ … │ y → │ … • … │
    └───┘     └───────┘

           ≅

        ┌───────────┐
        │           │
    ┌───┼───┐     ┌─▼─┐ 
y → │ … • … │ x → │ … │ 
    └───────┘     └───┘ 
```

Self-recursion is allowed inside a `rec`:

```
let
  rec fact x = if x <= 1 then 1 else x * fact (x - 1)
  gross = fact 5 + 24
in …
```

```
╔═══════════════════════════════════════╗
║          ┌──────────────────────┐     ║
║          │                      │     ║
║        ┌─▼──────────────────────┼───┐ ║
║ fact → │ λx. if … then … else … • … │ ║
║        └─▲──────────────────────────┘ ║
║          │                            ║
╚══════════╪════════════════════════════╝
         ┌─┼───────┐
 gross → │ • … + … │
         └─────────┘
```

Mutual recursion is also allowed inside a `rec`:

```
let { rec { x = A[y]; y = B[x]; }; z = C[x]; } in D
=
let { z = C[x]; rec { x = A[y]; y = B[x]; }; } in D
```

```
╔═════════════════════════════╗
║       ┌───────────────┐     ║
║       │               │     ║
║     ┌─▼─────┐     ┌───┼───┐ ║
║ x → │ … • … │ y → │ … • … │ ║
║     └─▲─┼───┘     └─▲─────┘ ║
║       │ │           │       ║
║       │ └───────────┘       ║
╚═══════╪═════════════════════╝
    ┌───┼───┐
z → │ … • … │
    └───────┘
```

In either case there may be incoming or outgoing edges that cross the `rec` boundary, but cycles *cannot* cross a `rec` boundary:

```
let { rec { x = A[y]; }; y = B[x]; } in C
```

```
╔═══════════════╗
║       ┌───────╫─────────┐
║       │       ║         │
║     ┌─▼─────┐ ║     ┌───┼───┐
║ x → │ … • … │ ║ y → │ … • … │
║     └───┼───┘ ║     └─▲─────┘
║         │     ║       │
║         └─────╫───────┘
╚═══════════════╝
```

There’s no *general* way to warn about code that differs in meaning based on whether `rec` is present, if the intent is explicitly to allow shadowing as a use case, however a good heuristic might be to warn about *things with syntactic parameters*, which I would *assume* are usually intended to be recursive functions:

```
let x = …
let x = … x …  -- Okay: meant to shadow/replace a variable.

let f = …
let f x = … f …  -- Warning: did you mean ‘let rec f x = … f …’?
```

(Also if you have trouble viewing these Unicode diagrams, please let me know if you’d like me to do ASCII versions!)

Tarmean:
Being able to let-bind with name shadowing in order would be a nice improvement for algorithms, especially in linear haskell. Between

    f x =
      let x = foo x
      in let (a,x) = bar x
      in let x = baz a x
      in x

and

    f x
      = foo x & \x ->
         bar x & \(a, x) ->
         baz a x & \x ->
         X

and

  f x =
     let!
      x = foo x
      (a,x) = bar x
      x = baz a x
    in x

I know which one I prefer. Maybe this should be a separate proposal. I feel most haskell usage uses let fairly rarely outside of CPS style so I'm not sure how many would turn this on by default, though.

 There is a tradeoff between not being surprised by a recursive let vs. being surprised by a non-recursive let. I would favour a new non-recursive let version like `let!` Or `let*` to make this behaviour obvious and the proposal backwards compatible.

Also, not really relevant but

   let
     !_ = unsafePerformIO (print 1)
     !_ = unsafePerformIO (print 2)
    in ()

currently prints 2 1. Hopefully nobody depends on this ordering either way.

nmaehlmann:
Disclaimer: I have no knowledge about GHC internals and write this from a user perspective. 

As far as I can see the problem of using recursion unintentionally is present regardless of the let environment.
I can write:
```
a :: Bool
a = a
```
and it compiles just fine. 

Instead of adding a language extension, which haskell necomers are not likely to enable by themselves, I think it would be nice to get a compiler warning or maybe even error in cases like this.
Unwanted recursion probably hard to detect in general, however the case of defining a zero arg function that references itself (regardless of wether its toplevel, let or where binding) seems to always be unwanted and may catch alot of potential errors already.

utdemir:
Apologies that this doesn't contribute much to the discussion, but I agree with @chrisdone's [comment](https://github.com/ghc-proposals/ghc-proposals/pull/401#issuecomment-790792340) about making this a compiler warning instead of a language extension; I also had a day to spare, so implemented a prototype GHC plugin (at renamer level) as per their suggestion to experiment with it: https://github.com/utdemir/recursive-let-plugin

AshleyYakeley:
I have two problems with the compiler-warning-only approach:

1. The established Haskell meaning of let-bindings is that they are recursive. If the compiler starts issuing warnings about heretofore correct usage, that implies that we've redefined the intent of that syntax.
2. There's still no way to use bindings to unpack existential types, which seems like an important use-case for non-recursive bindings. (It would certainly help me for my code.)

For this reason I would prefer new syntax for for the new kind of binding (and I happen to think `<-` is the cleanest).

sullyj3:
> If the compiler starts issuing warnings about heretofore correct usage, that implies that we've redefined the intent of that syntax.
I don't think this is really the end of the world if the warning is off by default.

HuwCampbell:
Apologies if this is noise.

I think a lot of the issues that @ocharles is having which motivate this change could be implemented with a warning on obviously _non-productive_ bottoms.

A simple recursive let binding which calls itself will obviously never produce anything of utility (unlike say the standard definition of `ones` or `cycle`, which, though recursion yield values lazily.

tomjaguarpaw:
Even a function boundary would render that kind of analysis impotent. Consider how you would detect the error in `let a = fromMaybe a Nothing` for example.

HuwCampbell:
That would certainly be possible to see after inlining and the case of case optimisation.

tomjaguarpaw:
If that's the correct answer to my challenge then I fear I must have misunderstood your original suggestion!

HuwCampbell:
No worries. I won't say I've fully fleshed this out to a proposal of its own.

What I'm thinking is a lightweight non-termination checker. Essentially a termination checker where it will fail where it knows it won't terminate; as opposed to failing unless it can't prove that it will.

chrisdone:
Ashley,

> The established Haskell meaning of let-bindings is that they are recursive. If the compiler starts issuing warnings about heretofore correct usage, that implies that we've redefined the intent of that syntax.

The compiler *only* issues warnings about correct usage. That’s why they’re warnings and not errors. Look at name shadowing warnings, unused do bindings, irrefutable pattern bindings on multi-constructor types, and orphan instances. We’re *refining* intent with a light-weight warning, after finding that certain edge cases of the syntax can be harmful, in an otherwise well established construct. 

> There's still no way to use bindings to unpack existential types, which seems like an important use-case for non-recursive bindings. (It would certainly help me for my code.)

This seems orthogonal to the original intent of the proposal by Ollie which was addressing accidental recursion in bindings. It happens rarely, but the cost of fixing it is high. A typical profile of a problem in need of a warning. Unpacking existentials without a case would be nice, but that seems like piggybacking!

chrisdone:
> Apologies that this doesn't contribute much to the discussion, but I agree with @chrisdone's comment about making this a compiler warning instead of a language extension; I also had a day to spare, so implemented a prototype GHC plugin (at renamer level) as per their suggestion to experiment with it if anyone is interested: https://github.com/utdemir/recursive-let-plugin

I look forward to trying this out, thanks! I’ll report back with feedback in the next few days.

AshleyYakeley:
> This seems orthogonal to the original intent of the proposal by Ollie which was addressing accidental recursion in bindings.

I believe a better way of addressing this is to create a new syntax for a new "non-recursive binding" intent, rather than making the existing syntax have an ambiguous intent (_recursive_ with the warning switched off, _non-recursive_ with the warning switched on). Warnings shouldn't change intent IMO.

Two different intents, two different unambiguous syntaxes. This approach allows mixing in the same file, even the same `let`-block, while avoiding awkward pragmas and warning flags. And as a very useful side-benefit, it allows unpacking of existential types.

AshleyYakeley:
Simple example (leaving aside the `where` issue). Readable! Unambiguous! Straightforward desugaring to `case`! No pragmas!

```haskell
print = \case ...
  App f x -> let
    ones = 1 : ones
    isfine <- ones
    this <- isfine
    in print f <> " " <> print x
```

tomjaguarpaw:
> I believe a better way of addressing this is to create a new syntax for a new "non-recursive binding" intent, rather than making the existing syntax have an ambiguous intent 

Could you elaborate on what aspect of @chrisdone's proposal makes intent ambiguous?

AshleyYakeley:
> Could you elaborate on what aspect of @chrisdone's proposal makes intent ambiguous?

If I come across `b = f a b c` is that intended to be a recursive binding, per existing Haskell? Or is it intended to be a non-recursive binding that happens to be incorrect? The compiler will make this choice of interpretations according to a warning flag.

tomjaguarpaw:
That's not my understanding. My understanding is that the semantics remain the same as existing Haskell but if the proposed warning will be raised (if it is enabled).

AshleyYakeley:
The program will compile the same way, true, but the warning implies a different intent (that is now being violated).

Rather than a `RECURSIVE` pragma, it would be better at least to have a `NONRECURSIVE` pragma instead. This would at least not change the meaning of existing code.

AshleyYakeley:
It's like this:

* no warning flag: compiler assumes all bindings are intended to be recursive
* warning flag: compiler assumes all bindings are intended to be non-recursive unless marked RECURSIVE, will complain if this is violated, but compile them as recursive nevertheless

AshleyYakeley:
Furthermore, it's a kind of half-assed non-recursive binding that doesn't even allow the one thing you'd really want non-recursive bindings to do, which is to unpack existential types.

Look, I get that a lot of people have a lot of existing prior code, full of bindings that were intended to be non-recursive, which they'd like to easily re-check for correctness without having to edit it all. But I don't believe it's worth unnecessarily disfiguring the language with warnings and pragmas for all _future_ code. Instead, we'd get in the habit of writing `pat <- expr` instead of `pat = expr`, and intent would be unambiguous.

chrisdone:
I don’t have anything new to add. I recognize @AshleyYakeley’s passion and conviction, but I don’t see any of my concerns or values reflected in his counter-arguments.

I’m not saying the warning idea is the right one, either. It’s just an idea.

maralorn:
I think it’s generally recognized, that Haskells biggest strength is converting runtime problems into compile time problems. In my experience the two nastiest runtime problems Haskell does not solve are accidentally diverging and memory leaks. 

The warning+exception solution helps with this. It can be the difference between a hanging production server and a compile time warning/error. That can save hours of debugging time or even a lot of money.

Now for coding convenience we could now enable other uses of the "now free" syntax. That might be nice. But it will bring a lot of bike-shedding for a minor improvement in coding ergonomics. I personally like the rust style re-use of variable names, but I don‘t think that's a defining feature of a language.

So my personal opinion is: Introducing a way (needn‘t be a warning, but that sounds a good fit) to prevent accidental recursion anywhere in the Modul (i.e. allowing less programs with -Werror) is vastly more important than having a nicer/different syntax which accepts a slightly different/nicer set of programs.

danidiaz:
I agree with @AshleyYakeley that it's better to express intent with mechanisms existing within the language, not through compiler warnings.

If we consider the current behaviour of `let` dangerous, let's add a warning, but have a separate non-recursive `let`.

The shortcomings of `var` in Javascript were solved by adding `let` and `const`, not by adding a language pragma that changed the behaviour of `var`.

chrisdone:
> I agree with @AshleyYakeley that it's better to express intent with mechanisms existing within the language, **not through compiler warnings.**

> If we consider the current behaviour of let dangerous, **let's add a warning**, but have a separate non-recursive let.

Are you for or against the compiler warnings? :thinking: 

If I'm making a good guess at your intended meaning, you're **specifically not in favour of a pragma** to disactivate warnings, but rather a new syntax where such a warning doesn't apply (like a new keyword or symbol). Is that correct?

danidiaz:
I have warmed to the pragma + warning solution. I find more palatable to think of the pragma not as a way to disable some warning, but as an optional clarification supplied by the programmer, a bit like function signatures. And, much like we have a warning for top-level declarations lacking signatures, we could have a warning for recursive lets that lack the explicit pragma.

As an analogy, in Java the [Override](https://docs.oracle.com/javase/tutorial/java/annotations/predefined.html) annotation is not mandatory, but overriding a method without adding it will get you a compiler warning. Also, adding it to a method that doesn't override another is a compiler error. 
 
Perhaps, instead of a pragma, the recently proposed [syntax for modifiers](https://github.com/goldfirere/ghc-proposals/blob/clarify-modifiers/proposals/0370-modifiers.rst) could be used instead.

AshleyYakeley:
So the key to understanding `<-` is that it is always syntax sugar for `case` with a single case. Of course, such `case` statements may be incomplete. If so, the compiler needs to create a code branch for non-matching cases.

In `do` notation, the two sides of the `<-` are of different types. If the case statement is incomplete, the compiler creates a `fail` branch, which in turn generates a `MonadFail` constraint, which may or may not be satisfiable.

With pattern guards, the two sides of the `<-` are of the same type. If the case statement is incomplete, the guarded case doesn't match on the non-matching branch. Note there is no "monady/applicative context" as such, at least, nothing needs to be an instance of `Applicative` or `Monad`.

With list comprehension, the two sides of the `<-` are of different types. If the case statement is incomplete, the element isn't drawn from the list on the non-matching branch. This is somewhat similar to pattern guards.

In the proposed `let` notation, the two sides of the `<-` are of the same type. If the case statement is incomplete, the compiler emits a warning and creates an `error` branch instead. If warnings-as-errors are switched on, then this is the same result as the `do` notation without a `MonadFail` constraint.

Like this:

|syntax|desugar|types|non-matching branch|
|-|-|-|-|
|`do` notation|to `case`|different|`fail` (cause error with no `MonadFail`)|
|pattern guard|to `case`|same|case not matched|
|list comprehension|to `case`|different|element not picked|
|`let` proposal|to `case`|same|`error` (also emit warning)|

This seems consistent to me.

chrisdone:
Your comment argues clearly in favor of my position. Your `let` is the only one that creates a hard runtime error—indicating a bug, whereas all the rest are perfectly legitimate control flows (failed pattern matches, yes, but you forgot `>>=` entirely), which is precisely why we use `<-` to denote it.

AshleyYakeley:
> Your `let` is the only one that creates a hard runtime error—indicating a bug,

But the compiler also generates a warning. So this bug is caught at compile time. Isn't _your_ proposal to generate a warning to catch a bug?

AshleyYakeley:
In addition, for `do` notation with no `MonadFail` constraint, there is no legitimate control flow for non-matching patterns.

AshleyYakeley:
TBH I just think you're not used to it. If `let` `<-` were in the language from the start, you'd think it the most natural thing in the world.

chrisdone:
To any Haskeller of any experience, `v <- e` *means* something: here lies control flow. **Whether or not** there is a pattern in `v`. Something “extra” is going on.

A simple pure binding `v = e` that *happens to be non-recursive* is clearly not control flow. 

That’s what I think. I still think you could propose a different operator or keyword for this.

AshleyYakeley:
It's not always control flow, though, is it?

So currently, for `do` notation, the non-matching generates `fail`. But I almost never want this, even for monads with `MonadFail` such as `IO`. Why? Because it will generate a `fail` message with a message such as "pattern not matched at line 10 char 14". This failure is difficult to catch, and the message is meaningless to present to the user of the program. Just like a hard `error`, this autogenerated `fail` is always indicative of a bug. In fact, I'm looking forward to the qualified do-notation extension just so I can always catch these bugs at compile time.

In such a case, `do` `<-` does not indicate any control flow. Or alternatively, it indicates the same trivial control flow as `let` `<-`. For example:

```
do
    pat <- expr
    body

let
pat = expr
in body
```
If the patterns are complete, these desugar to
```
expr >>= \case
    pat -> body

case expr of
    pat -> body
```
If the patterns are not complete:
```
expr >>= \case
    pat -> body
    _ -> fail "..." -- causes compile-time error if no MonadFail instance

case expr of
    pat -> body
    _ -> error "..." -- causes warning
```
I feel like this is the same kind of "control flow" for both.

chrisdone:
@utdemir 

> Apologies that this doesn't contribute much to the discussion, but I agree with @chrisdone's comment about making this a compiler warning instead of a language extension; I also had a day to spare, so implemented a prototype GHC plugin (at renamer level) as per their suggestion to experiment with it if anyone is interested: https://github.com/utdemir/recursive-let-plugin

I ran this on my compiler and it works nicely. [Here is an output](https://gist.github.com/chrisdone/3dee1f8ea17d39aea4ef9ec7d97c03a9). It found all my worker loops. I intentionally put an "accidental" self-reference:


```haskell
    refl =
      case name of
        HashGlobal g -> refl
```

And sure enough, it caught it:

```
src/Inflex/Resolver.hs:313:5: warning: Recursive definition at src/Inflex/Resolver.hs:313:5
    |
313 |     refl =
    |     ^^^^
```

Nice work! I will see how I can integrate this into my workflow.

AshleyYakeley:
The most useful thing for me would be if non-recursive `let` can unpack existential types, since I work with a lot of them. This seems like a really obvious use case for non-recursive bindings. I believe that should be in the proposal, and I certainly care about it more than syntax.

If people really don't like `let` `<-` syntax (despite, as I have pointed out, there being no single consistent "something extra" implied by `<-`), then I would suggest one small change to the proposed syntax, and make `rec` a block keyword, matching recursive `do` notation. For example, compare:

```
do
    a <- aa
    rec
        b <- bb c
        c <- cc b
    d <- dd
    expr

let
a = aa
rec
    b = bb c
    c = cc b
d = dd
in expr
```

The advantages of this relative to the proposal:

* can mix recursive and non-recursive bindings in a `let`
* consistency with `do` notation (arguably also an advantage relative to `let` `<-`)

Disadvantages relative to `let` `<-`:

* not backwardly compatible
* slightly more verbose

chrisdone:
As far as the submitted proposal itself goes, having digested more of the thread. I do support the general thrust. If we're putting our own spin on things, here's mine.

* Extension `RecursiveBindings` (not using the word "let"), on by default, that enables recursion across all bindings (`let`/`where`/top-level), not just lets. Disabling it disables said recursion *everywhere* within a module. (See below for trade-offs)
* Extension `LetRec` that enables `letrec` (or `rec { ...}`) syntax which would be the _only_ way to achieve recursion (self and mutual) when `RecursiveBindings` is turned off.

I could set `NoRecursiveBindings` at the project level and enable `LetRec` on a per-module by-need basis, or enable both at once. 

I do agree that the warning approach suffers from how to state intentional recursion: The pragma doesn't appeal much (although @danidiaz seems to be ok with it), which I understand. So an alternative to `NoRecursiveBindings`:

* A warning `-Wrecursive-bindings` that would complain rather than prevent, and it could be **shut up by enabling** `LetRec` and using `letrec` (or `rec`) syntax to be explicit. 

Regarding `<-` (or whatnot), I don't, personally, believe adding a new operator for binding is a good direction because: yes, I do have lots of code that I'd like to check (example: all of Hackage and my own packages), and secondly, `letrec` (and `rec { }`) is well established elsewhere, whereas an operator is an open question, doesn't seem worth re-hashing a debate over. 

The idea of a `let x = y; rec { a = b; b = a }` as suggested by @phadej seems flexible, but I don't feel very strongly about `letrec` vs `rec { .. }`. My vote on that would be with whatever is more popular. 

From taking a look at my own code, intentional self-recursion is rare and I don't mind sticking it in a `letrec` or `rec`. I understand that @goldfirere (who made IMO a compelling case against @ndmitchell's preference) uses `where` more than `let`, because so do I -- rarely using `let`, so I would indeed expect that what is true for `let` should be true for `where`: this perspective does permit the choice of either a new construct `letrec`, or a new enhancement `rec` that would work in either `let` or `where`. 

Explicitly, I _don't_ think that you should require a certain order of bindings like in Lisp (as suggested by @ocharles and @gridaphobe). Lisp is not pure, so order matters. Haskell does dependency analysis already, I don't think ordering buys us anything new. For people who like to write their code top-down, it's a regression. 

If I'm writing a parser, then all top-level modules are mutually recursive to a _very_ large extent. This is an example where I simply would enable `RecursiveBindings` or turn off `-Wrecursive-bindings`. I believe, in this case, a warning is simpler than an extension, where I would turn it on temporarily to help me root out a bug where recursion really wasn't intended. An extension could do that too, but it _feels_ more like the kind of thing warnings are intended for.

But I'm not married to the idea of a warning vs an extension; just seems more "warningy" and less "extending".

I think it is worth pointing out that you can create accidental loops at any level; a let, a where or top-level. Solving it for only one area would lead me to a false sense of security and even mislead me when debugging.

Tried to address all the points I saw and share my view so I didn't have to litter the thread with any further commentary.

proposals/0000-letrec.md
```diff
@@ -0,0 +1,116 @@
...
+
+With this proposal, if the author had disabled recursive let bindings with `-XNoRecursiveLet`, the definition would have been rejected. The options would be to change `let` to `let rec`, or to pause, reflect, and hopefully write the intended code.
+
+## Proposed Change Specification
+
+* A new language extension is added: `RecursiveLet`. This can be enabled or disabled via the normal mechanisims for extensions. By default the extension is enabled.
+
+* When `RecursiveLet` is _enabled_ (the default, or explicitly with `{-# language RecursiveLet #-}`), there are no changes.
+
+* When `RecursiveLet` is _disabled_ (`{-# language NoRecursiveLet -#}`), compilation of `let` expressions changes. With `NoRecursiveLet`, group of `let`-bound bindings cannot refer to themselves, either directly or mutually. For example, with `NoRecursiveLet`, the following will now fail to compile:
```
konsumlamm, line 56:
```suggestion
* When `RecursiveLet` is _disabled_ (`{-# language NoRecursiveLet #-}`), compilation of `let` expressions changes. With `NoRecursiveLet`, group of `let`-bound bindings cannot refer to themselves, either directly or mutually. For example, with `NoRecursiveLet`, the following will now fail to compile:
```

gbaz:
I would like to see some version of this proposal revived. In particular, I would be happy with any solution that allowed for a combination of flags giving a warning on recursive bindings -- even better if there was some combination that provided a warning and no workaround (i.e. if `let rec` and `-Wrecursive-let` were separately enabled).

I've played with a number of dsl and combinator things over time that only really "make sense" when recursion is not allowed (since they correspond to the stlc without fixpoints). It would be nice to, rather than simply asking users be careful to avoid explicit recursion, providing a discipline where they could turn on some flags and prevent it.

This for example can reduce the need for some of the "observable sharing" tricks done with EDSLs for hosted languages. But it would also allow embedding of languages with implicit complexity such that they guarantee "if you use these particular combinators, and you turn on these flags, then nothing you write can have a complexity beyond a certain class". There are some other related uses I can think of as well.

googleson78:
This would have saved me 30 extremely frustrating minutes today.

simonpj:
> This would have saved me 30 extremely frustrating minutes today.

I get bitten by accidental bottoms quite regularly too, and yes it's frustrating.  So I too would like to see this proposal revived.

I strongly urge re-imagining it as a *warning*.  That is, with `-Wrisky-recursion` you'd get warnings like this:
```
x = x    -- Warning
x = 1+x    -- Warning
x = 1:x     -- No warning; (lazy) constructor in the way
fac = \n -> ...fac ...     -- No warning: lambda in the way
```
That is, no warning provided the recursion is broken by a lazy data constructor or a lambda.   That means that 99% of the time we get no warnings.

But not always.  "Almost-circular programming" makes  you write stuff like
```
   (x,y) = f x z
```
where `f` produces an output `x` which is fed back into `f`.  This is sometimes precisely what you want.  We need a way to suppress the warning selectively when doing this.    One simple way might be this: write
```
   (x,y) = almostCircular (f x z)
```
where `almostcCircular` is a built-in identity function, but for the purposes of these warnings it breaks the loop just like a data constructor.

We just need a willing author.

googleson78:
My concrete grievance was of this form
```haskell
let (x, y) = ........ f x ..........
```
where I certainly did not want for this to happen.

The thread is very long already, and I don't know why the original intent was seemingly veered away from, but I would just love to have a keyword `letnorec` which I would just always use by default, until the moment came where I actually need a recursive binding.

sgraf812:
I think I agree with Simon that this would be a great warning. Alas, this means we need a (clever) guardedness/termination analysis.

But I would try without yet another builtin first. I could count on one hand the number of times where I *actually* wanted to write a guarded recursive data definition (one that wasn't syntactically guarded, such as `x = 1:x`). I think I have been bitten by this issue at least as many times. So I would simply add `{-# OPTIONS_GHC -Wno-risky-recursion #-}` to the top of the file when it fires and I don't want it to. It's not hard to put the "almost circular" definition into a different module anyway.

Perhaps we should see how often the warning triggers in practice (depends on the precision of the analysis) and then decide whether we need a builtin `almostCircular` (I'd choose the color `assumeGuarded` instead).

AshleyYakeley:
I'd like a form (`letnorec` or `letseq` or whatever) whereby I can unpack existential types instead of using `case`.

simonpj:
> I'd like a form (letnorec or letseq or whatever) whereby I can unpack existential types instead of using case.

That's a reasonable desire, but it's different, and significantly more complicated, that what we are discussing.   Let's not conflate the two.

> So I would simply add {-# OPTIONS_GHC -Wno-risky-recursion #-} to the top of the file when it fires

I would *much, much* rather have a way to suppress the warning *locally and precisely where needed*, rather than saying "somewhere in this module I use almost-circular programming".

AshleyYakeley:
Hmm, with the proposed warning flag set, would this code give a warning?
```
let
x = y
y = 1
in x
```

AshleyYakeley:
It sounds like there are three possible things?
1. Complaining about forward references (what I want)
2. Complaining about recursive references (this proposal)
3. Complaining about risky recursive references (Simon's suggestion)

tbidne:
Is writing a cotermination checker for GHC _relatively_ straightforward? I assumed it would be just as hard as a (structural) termination checker, and therefore probably require significant research.

It would be a great feature though.

noughtmare:
I'm not sure if a guardedness warning would really solve the problem. For example, it would not prevent me from writing code like this:

```
let x = []
    x' = 1 : x'
```

When I meant to write:

```
let x = []
    x' = 1 : x
```

Non recursive let bindings would probably have saved me in this case.

s-and-witch:
Did the thread already mention qualified do?
```haskell

module NoRec where

a >>= b = b a

{-# LANGUAGE QualifiedDo #-}
module M where
import NoRec qualified

foo = NoRec.do
  x <- 1
  x' <- 1 : x -- x' is not valid here
  map (+1) x'
```

maralorn:
I am a bit weary of overwhelming a good proposal with alternatives. Often proposals get stuck in limbo because of decision paralysis. But anyway …

As I said before in this thread I would love to get a warning for every explicit recursion I wrote. Now the question is how to opt-out of the warning. (Introducing a way for ghc to decide which recursions are good and which are bad would be awesome, but I wouldn’t begrudge my compiler for not solving that problem.)

Obvious possibilities are pragmas or modifiers on the module or function level but that’s too coarse grained. 

If we have some kind of marker which can be used inside of declarations or expressions, I can think of three possibilities:

1. A block marker, marking a number of declarations, which allows all bindings within that marker to be recursive. (This is how the proposal is currently written.)
2. A marker on bindings, which says that the variables in this binding may be used recursively. (very similar to the former.)
3. A marker on variable usage sites, (or theoretically more generally expressions) that says that this usage of a variable may introduce a recursion.

The third solution is probably too alien, but I like it because it feels very explicit and general. We could immediately introduce it for all expressions in a module without worrying whether the binding comes from a toplevel, where or let clause.
An example could be something like `ones = 1 : rec ones`.

Again probably to alien for most, but I would happily and immediately turn on an extension which forces me to do this for every recursive binding.

AshleyYakeley:
Would forward references still be allowed? e.g.
```
let
x = y
y = 5
in x
````

**Comment by sgraf812** [comment]
Re: turn `-Wrisky-recursion` off: Simon proposed a primitive `almostCircular`. But why not simply use `~`? That is, if I want to write

```
let (x,y) = f x z in ...
```

which will trigger the new, conservative (!) warning on `x` because the analysis gives up on `f`. I can turn it off by prefixing it with `~` (which makes no semantic difference):

```
let (~x,y) = f x z in ...
```

No language extension needed. Furthermore, this warning could potentially live in a separate linter, such as `stan`.

What I mean by "conservative" is this: 
Whenever the user has written a data `let` that diverges for some input, we emit a warning.
Conversely due to Rice's theorem, this means that we will sometimes warn even though there is no input for which the data `let` diverges, as in `f` above. That's OK; the user can turn the warning off by "breaking" the recursion through use of `~`.

(NB: This is *not* strictness analysis. Strictness analysis is MUST diverge, termination analysis is MAY diverge.)

An absolutely terrible and attractively simple termination analysis would simply warn for *every* recursive occurrence of `x`, even for `let x = 1:x`. Fine still; can always write `let ~x = 1:x`. I would even say it's good practice to prefix `~x` like this to emphasize which variable only occurs guarded.

**Comment by simonpj** [comment]
> Then I can turn the warning off by prefixing x's binding site with ~ (which makes no semantic difference):

I like that idea.

**Comment by maralorn** [comment]
> Re: turn `-Wrisky-recursion` off: Simon proposed a primitive `almostCircular`. But why not simply use `~`?

Can we make that work for top level bindings? (But maybe that should be out of scope anyway.)

**Comment by sgraf812** [comment]
@Ericson2314 and me have again been bitten by an accidental recursive binding: https://github.com/haskell/happy/pull/223/files#r1773713191. It slipped through 3 rounds of code review on my behalf and I spent over an hour debugging. If I knew what to look for, I should have immediately tried to grep the Core for `case .. of {}`, alas...

**Comment by tomjaguarpaw** [comment]
> @Ericson2314 and me have again been bitten by an accidental recursive binding

Could you point out where exactly it is? The URL looks like it should "jump to location", but it's not doing so for me.

**Comment by sgraf812** [comment]
I think it got lost in a force push. Basically, there is a module that wraps `IntSet` as a newtype `NameSet` which functions as a set of `Name`s (which is a newtype around `Int`). There are a bunch of `coerce`d functions of the form

```hs
empty :: NameSet
empty = coerce IntSet.empty

singleton :: Name -> NameSet
singleton = coerce IntSet.singleton

fromList :: [Name] -> NameSet
fromList = coerce IntSet.fromList

--

delete :: Name -> NameSet -> NameSet
delete = coerce IntSet.delete

--

member :: Name -> NameSet -> Bool
member = coerce IntSet.member
```

and I completely glossed over

```hs
difference :: NameSet -> NameSet -> NameSet
difference = coerce difference
```

which is *not like the previous definitions at all*! Furthermore, it could have easily been spotted as a risky recursion, prompting the user to write it as

```hs
difference :: NameSet -> NameSet -> NameSet
difference = coerce IntSet.difference
```

which it is in the most recent form of the PR.

maralorn:
We have been bitten by this in production again today. I would love to see this move forward.

Pitometsu-GeniusYield:
> Did the thread already mention qualified do?
> 
> ```haskell
> module NoRec where
> 
> a >>= b = b a
> 
> {-# LANGUAGE QualifiedDo #-}
> module M where
> import NoRec qualified
> 
> foo = NoRec.do
>   x <- []
>   x' <- 1 : x -- x' is not valid here
>   map (+1) x'
> ```

But `'x'` is still valid there.
E.g. in OCaml we can
```ocaml
let f = fun x ->
  let x = [3]
  and x' = 1::x (* the scope is not polluted by unnecessary binding 'x' *)
  in x' @ x
in f [2] (* - : int list = [1; 2; 3] *)
```

PR #402, opened 2021-02-11
Stable GADT constructor syntax (under review)

int-index:
The syntax of GADT constructor declarations is [documented in the User's Guide as "subject to change"](https://downloads.haskell.org/ghc/9.0.1/docs/html/users_guide/exts/gadt_syntax.html#formal-syntax-for-gadts). We propose to fix a couple of issues associated with it and to advertise the resulting syntax as stable (that is, a breaking change to
it would require another proposal).

[Rendered](https://github.com/serokell/ghc-proposals/blob/gadt-syntax/proposals/0000-gadt-syntax.rst)

### DIFF
--- a/proposals/0000-gadt-syntax.rst
+++ b/proposals/0000-gadt-syntax.rst
@@ -281,13 +281,13 @@ Clarifications and side conditions:
    written as ``∷``, ``→``, and ``⇒`` respectively.
 5. The ``btype`` production in ``gadt_prefix_sig`` and ``gadt_record_sig`` is
    taken as the result type of the GADT constructor; it must be a substitution
-   instance of the type being declared. This implies that ``data T where T1 ::
+   instance (allowing expansion of type synonyms, but not type families)
+   of the type being declared. This implies that ``data T where T1 ::
    (Int -> T)`` is rejected. The error message should not only say that ``(Int
    -> T)`` is not an instance of ``T`` but also helpfully explain where
    parentheses are and are not allowed in constructor signatures.
-6. While not a matter of syntax, we find it relevant to point out that type
-   synonym expansion is allowed in result types, while type family expansion is
-   not allowed. This matches the status quo.
+   (Note that the treatment of type synonyms vs. type families matches the
+   status quo.)
 
 Effect and Interactions
 -----------------------
--- a/proposals/0000-gadt-syntax.rst
+++ b/proposals/0000-gadt-syntax.rst
@@ -288,6 +288,9 @@ Clarifications and side conditions:
    parentheses are and are not allowed in constructor signatures.
    (Note that the treatment of type synonyms vs. type families matches the
    status quo.)
+6. If ``forall tvs ->`` quantification is used in a constructor's type, the use
+   of such constructor in an expression or a pattern is an error.  However, it
+   can be promoted and used in types under ``DataKinds``.
 
 Effect and Interactions
 -----------------------
--- a/proposals/0000-gadt-syntax.rst
+++ b/proposals/0000-gadt-syntax.rst
@@ -1,0 +1,324 @@
+Stable GADT constructor syntax
+==============================
+
+.. author:: Vladislav Zavialov
+.. date-accepted::
+.. ticket-url::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/402>`_.
+.. contents::
+
+The syntax of GADT constructor declarations is `documented in the User's Guide
+<https://downloads.haskell.org/ghc/9.0.1/docs/html/users_guide/exts/gadt_syntax.html#formal-syntax-for-gadts>`_
+as "subject to change". We propose to fix a couple of issues associated with it
+and to advertise the resulting syntax as stable (that is, a breaking change to
+it would require another proposal).
+
+
+Motivation
+----------
+
+First and foremost, let us do away with the misconception that the GADT
+constructor signatures are type signatures.
+
+1. **Constructor-specific syntax**
+
+   There are syntactic forms that are not valid in type signatures but are
+   currently allowed in GADT constructor signatures::
+
+     data T a where
+       S :: !a -> T a
+       R :: { fld :: a } -> T a
+
+   In ``S``, the field is marked as strict using a prefix ``!``. In ``R``, the
+   field is given a name using record syntax ``{ ... }``. The signatures of these
+   constructors are perfectly fine, but they are rejected if used as function type
+   signatures::
+
+     s :: !a -> T a
+     s = S
+
+     r :: { fld :: a } -> T a
+     r = R
+
+   ``s`` and ``r`` are rejected with the following error messages::
+
+     GadtConSyn.hs:7:6: error:
+         Unexpected strictness annotation: !a
+         strictness annotation cannot appear nested inside a type
+
+     GadtConSyn.hs:10:6: error:
+         Record syntax is illegal here: {fld :: a}
+
+2. **Constructor-specific limitations**
+
+   There are also type signatures that are not valid GADT constructor
+   signatures::
+
+     k :: a -> Eq a => F
+
+   This is accepted (assuming a ``data F`` declaration elsewhere), but an
+   analogous constructor signature is not::
+
+     data F where
+       K :: a -> Eq a => F
+
+   ``K`` is rejected with the following error message::
+
+     GadtConSyn.hs:16:3: error:
+         GADT constructor type signature cannot contain
+         nested ‘forall’s or contexts
+
+So, if GADT constructor signatures are syntactically not type signatures, then
+what are they?  The User's Guide provides a BNF-style grammar for GADT
+constructor declarations, see `6.4.7.1. Formal syntax for GADTs
+<https://downloads.haskell.org/ghc/9.0.1/docs/html/users_guide/exts/gadt_syntax.html#formal-syntax-for-gadts>`_::
+
+  gadt_con ::= conids '::' opt_forall opt_ctxt gadt_body
+
+  conids ::= conid
+          |  conid ',' conids
+
+  opt_forall ::= <empty>
+              |  'forall' tv_bndrs '.'
+
+  tv_bndrs ::= <empty>
+            |  tv_bndr tv_bndrs
+
+  tv_bndr ::= tyvar
+           |  '(' tyvar '::' ctype ')'
+
+  opt_ctxt ::= <empty>
+            |  btype '=>'
+            |  '(' ctxt ')' '=>'
+
+  ctxt ::= ctype
+        |  ctype ',' ctxt
+
+  gadt_body ::= prefix_gadt_body
+             |  record_gadt_body
+
+  prefix_gadt_body ::= '(' prefix_gadt_body ')'
+                    |  return_type
+                    |  opt_unpack btype '->' prefix_gadt_body
+
+  record_gadt_body ::= '{' fieldtypes '}' '->' return_type
+
+  fieldtypes ::= <empty>
+              |  fieldnames '::' opt_unpack ctype
+              |  fieldnames '::' opt_unpack ctype ',' fieldtypes
+
+  fieldnames ::= fieldname
+              |  fieldname ',' fieldnames
+
+  opt_unpack ::= opt_bang
+              :  {-# UNPACK #-} opt_bang
+              |  {-# NOUNPACK #-} opt_bang
+
+  opt_bang ::= <empty>
+            |  '!'
+            |  '~'
+
+Unfortunately, there are a couple of issues associated with it:
+
+* Right before this grammar is given, the User's Guide states "Note that this
+  grammar is subject to change in the future", so it is of limited use when
+  writing forward-compatible code.
+
+* It does not actually match the implementation, as it mistakenly states that a
+  strictness annotation is followed by a ``btype`` (it actually must be
+  followed by an ``atype``), and it does not mention linear type syntax at all.
+
+* It is too permissive with regards to parentheses, causing implementation
+  issues (see `#19192
+  <https://gitlab.haskell.org/ghc/ghc/-/issues/19192#note_329172>`_).
+
+In this proposal we aim to give an alternative specification that would fix the
+aforementioned issues.
+
+Let us first point out that GADT constructor syntax pursues two contradictory
+goals:
+
+1. On the one hand, GADT constructor syntax tries its best to mimic type
+   syntax::
+
+     data T x where
+       MkT :: forall a. Eq a => Maybe a -> T 0
+
+     ghci> :t +v MkT
+     MkT :: forall a. Eq a => Maybe a -> T 0
+
+   Notice how the type signature reported by ``:t`` is identical to the one in
+   the declaration.
+
+2. On the other hand, GADT constructor syntax allows the user to specify
+   additional information about the fields, such as their names (using record
+   syntax ``{ fld :: a }``) and strictness (``{-# UNPACK #-}``, ``!a``,
+   ``~a``). This information is not reflected in the constructor's type.
+
+It is impossible to define a constructor by its type signature and at the same
+time include information that is not part of its type signature. This inherent
+contradiction means that any solution will be a compromise.
+
+Perhaps if we were to drop one of these goals, we could come up with something
+glorious, but in practice we must take backwards-compatibility and ease of
+implementation into account. With that in mind, let us establish the following
+principles:
+
+1. A GADT constructor signature is either in *prefix style* or *record style*.
+   The two styles are completely distinct and we do not seek to unify their syntax.
+
+2. A record-style GADT constructor signature has the following parts in a
+   fixed order:
+
+   #. (optional) ``forall tvs.`` to bind type variables;
+   #. (optional) ``ctx =>`` to introduce constraints;
+   #. ``{ fld1 :: a, fld2 :: b, ... }`` to describe constructor fields, using
+      the same syntax as classic Haskell98-style record type declarations, such
+      as ``data R = MkR { fld1 :: a, fld2 :: b, ... }``;
+   #. ``->`` as a separator, which is special syntax rather than
+      the ``(->)`` type constructor;
+   #. the result type ``T``;
+
+3. A prefix-style GADT constructor signature has zero or more of the following
+   parts in a free order:
+
+   #. type variable telescope ``forall tvs.`` or ``forall tvs ->``;
+   #. constraint context ``ctx =>``;
+   #. parameter ``a ->``, ``a %p ->``, or ``a ⊸``, possibly with a strictness
+      annotation;
+
+   It then ends with the result type ``T``.
+
+
+Proposed Change Specification
+-----------------------------
+
+Changes from status quo
+~~~~~~~~~~~~~~~~~~~~~~~
+
+1. Drop support for parentheses around the tail of a prefix-style constructor
+   signature (fix `#19192
+   <https://gitlab.haskell.org/ghc/ghc/-/issues/19192>`_).
+2. Permit nested foralls and contexts (a free order of quantifiers) in
+   prefix-style constructor signatures (fix `#18389
+   <https://gitlab.haskell.org/ghc/ghc/-/issues/18389>`_).
+
+Grammar specification
+~~~~~~~~~~~~~~~~~~~~~
+
+Existing non-terminals
+^^^^^^^^^^^^^^^^^^^^^^
+
+In this section we describe the non-terminals, the definition of which is out
+of scope of this proposal (they are already present in the GHC grammar)
+
+1. ``atype`` - syntactically atomic types, e.g. ``T``, ``42``, ``x``, ``( ... )``, ``[ ... ]``
+2. ``ftype`` - application chain consisting of ``atype``, e.g. ``T @a b``
+3. ``btype`` - infix operator chain consisting of ``ftype``, e.g. ``T @a b + Q 42``
+4. ``forall_telescope`` - either ``forall tvs.`` or ``forall tvs ->``
+5. ``sig_vars`` - comma-separated field names, e.g. ``fld_1, fld_2, fld_n``, n >= 1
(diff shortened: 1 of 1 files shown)


AntC2:
While we're tidying up GADTs, there's similar similar-but-different nuances with Pattern signatures - which can also have strictness annotations and/or use record style; oh, also that curious double-`=>`.

> strictness ... not reflected in the constructor's type

Do we reflect strictness in the type of ordinary H2010 style constructors? We should see the same result from `:t` whichever style was used to declare.

int-index:
> Do we reflect strictness in the type of ordinary H2010 style constructors? We should see the same result from `:t` whichever style was used to declare.


Strictness annotations are not part of type signatures regardless of declaration style:

```haskell
ghci> data T = MkT !Int
ghci> :t MkT
MkT :: Int -> T
```

> While we're tidying up GADTs, there's similar similar-but-different nuances with Pattern signatures - which can also have strictness annotations and/or use record style; oh, also that curious double-`=>`.

Pattern signatures definitely deserve some attention, the `=>` arrow makes me feel uneasy. However, this particular proposal is about GADTs, and I don’t want to broaden its scope.

proposals/0000-gadt-syntax.rst
```diff
@@ -0,0 +1,305 @@
...
+
+.. author:: Vladislav Zavialov
+.. date-accepted::
+.. ticket-url::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/402>`_.
+.. contents::
+
+The syntax of GADT constructor declarations is documented in the User's Guide
```
simonpj, line 12: Make "documented in the user guide" into a link

proposals/0000-gadt-syntax.rst
```diff
@@ -0,0 +1,305 @@
...
+  data T where
+    T1 :: (Int -> T)
+    T2 :: (forall a. a -> T)
+    T3 :: forall a. (a -> T)
+    T4 :: forall a. (a -> (T))
+
+This makes it possible to avoid conflicts between ``btype`` and
+``gadt_prefix_sig`` in the LALR grammar.
+
+On the other hand, we now allow nested ``forall`` and contexts::
```
simonpj, line 281: This point comes up rather late and almost by-the-way.  But it will have a pretty major effect on the implementation.  It may also make it harder to specify pattern matching (see the [Type variables in patterns](https://www.microsoft.com/en-us/research/publication/type-variables-patterns/) paper).

I can see the motivation, but I'd like the proposal to draw attention to this change more strongly, certainly under Proposed Change Spec

simonpj:
I think this is on a good, consolidating path

proposals/0000-gadt-syntax.rst
int-index: I added the following paragraph to the Proposed Change Specification:

> Permit nested foralls and contexts (a free order of quantifiers) in prefix-style constructor signatures (fix #18389).

And the following paragraph to Costs and Drawbacks:

> The free order of quantifiers in prefix-style constructors will have major effect on the implementation.

monoidal:
Note that strictness annotations can be recovered with `:i MkT`. Strictness is not stored in types, e.g. `f !x y = y` and `g x y = y` have the same type.

In the proposal, `->.` should be the lollipop `⊸` (available only with UnicodeSyntax). We do not have `->.` since the linear types syntax has changed, but we had to name the token somehow.

proposals/0000-gadt-syntax.rst
```diff
@@ -0,0 +1,319 @@
...
+             |  record_gadt_body
+
+  prefix_gadt_body ::= '(' prefix_gadt_body ')'
+                    |  return_type
+                    |  opt_unpack btype '->' prefix_gadt_body
+
+  record_gadt_body ::= '{' fieldtypes '}' '->' return_type
+
+  fieldtypes ::= <empty>
+              |  fieldnames '::' opt_unpack ctype
```
int-index, line 109: @monoidal points out that I missed linearity syntax here.

goldfirere:
Looks good. I'm in support. Thanks for the very nice write-up.

proposals/0000-gadt-syntax.rst
int-index: Actually, I’m not sure if we want the syntax from #111 or #370. They seem to differ.

* #111 gives this example:
   ```haskell
   data R where
      R :: { f1 %π :: A, f2 %ρ :: B } -> R
   ```

* #370 gives this grammar:
  ```haskell
  fielddecl ::= vars '::' (type | '!' atype) {modifier}
  ```

  Which would look as:

   ```haskell
   data R where
      R :: { f1 :: A %π, f2 :: B %ρ } -> R
   ```

Which one do we want? @monoidal @goldfirere

goldfirere: I don't care. Maybe I didn't look at #111 closely enough when I wrote #370? Either syntax is compatible with how I see modifiers working in general: they affect the "next thing", if there is one. In the #111 syntax, the next thing is the `::`, which is a reasonable thing to affect. In the #370, there is no next thing, in which case the modifier affects the entire construct it appears in -- that is, the field declaration. So I don't much care. Thanks for noticing this discrepancy!

(If we go with the #111 syntax, I can change #370 accordingly.)

aspiwack: This is pretty much my fault for not noticing the discrepancy when #370 was raised.

If nobody is strongly opposed, let's go with #111 's syntax: it has been examined more carefully. This was not a random proposition, but one that underwent several rounds of testing for least amount of surprise, and such (I may make it sound a bit more serious than it actually was, but it was still kind of seriously tested).

int-index: Ok, I updated my proposal to use the #111 syntax.

proposals/0000-gadt-syntax.rst
```diff
@@ -0,0 +1,320 @@
...
+Proposed Change Specification
+-----------------------------
+
+Changes from status quo
+~~~~~~~~~~~~~~~~~~~~~~~
+
+1. Drop support for parentheses around the tail of a prefix-style constructor
+   signature (fix ``#19192``).
+2. Permit nested foralls and contexts (a free order of quantifiers) in
+   prefix-style constructor signatures (fix ``#18389``).
```
aspiwack, line 204: Very cheap nitpicking on my part. But it would be quite helpful if you added links to the issues here (and below).

aspiwack (commented):
It does make sense to me.

int-index:
@nomeata There doesn’t seem to be much discussion around this proposal, so I’d like to submit it to the committee.

goldfirere:
A few clarification questions:

* The grammar includes `~` as a `strictness_sigil`. Is this possible only with `-XStrictData` (as it is today)? (I actually would like `~` to be available generally, but that's properly a separate proposal.)
* I assume the Unicode arrow is still available (both thin arrow and fat arrow) in a `quantifier`. Might be good to say somewhere that `'->'` and `'=>'` include these possibilities.
* The proposal says that `data T where T1 :: (Int -> T)` would be rejected. But the grammar accepts this. I think you need to specify that the `btype` production of `gadt_prefix_sig` is taken as the result type of the GADT constructor; it must be a substitution instance of the type being declared.
* Is type synonym expansion allowed in result types? (It is today.) Is type family expansion allowed in result types? (It is not today.) These facts should be included.
* This proposal restricts record constructors to have all their type quantifiers up at the front. That strikes me as a reasonable stepping stone. But there should be some text describing this state of affairs, as it makes record constructor types strictly more restrictive than non-record constructor types.
* This proposal allows all quantified variables in a record constructor to be visible, or for all to be invisible. It does not allow a mix. This seems a bit strange. Maybe make `forall_telescope` recursive?
* It would be nice to have a future work section, which will perhaps inspire others to echo this proposal. Future work includes fixing pattern synonyms in a similar way and further generalizing record-constructor syntax.
* I am a little concerned about error messages. If someone writes `data T where MkT :: Int -> (Int -> T)` because they learned that `->` is right-associative but likes to be very explicit about parsing, and the definition is rejected, I would be unhappy if the error message says something like "Error: `(Int -> T)` is not an instance of `T`." True, but hardly helpful. The proposal does not need an exact wording of the message specified, but it should indicate that GHC will explicitly detect this case and helpfully explain where parentheses are and are not allowed in constructor signatures.

Just a few suggestions for hopefully-quick edits. I plan to recommend acceptance. Thanks!

Please ping me here with a mention when this is ready.

int-index:
> This proposal allows all quantified variables in a record constructor to be visible, or for all to be invisible. It does not allow a mix. This seems a bit strange.

The record syntax allows the `forall tvs. b` form only, not `forall tvs -> b`. The proposal already contained this side condition.

As to the other questions, I tried to incorporate the answers into the proposal. Let me know if I missed anything. Thanks!

int-index:
> Please ping me here with a mention when this is ready.

@goldfirere Ping. (Apologies, I forgot to include a mention in my previous comment)

int-index:
@goldfirere Ping.

goldfirere:
Great. Thanks. I have left a PR at https://github.com/serokell/ghc-proposals/pull/4 with a tiny wording clarification. I will now recommend acceptance to the committee.

goldfirere:
Hrm. I went to recommend acceptance, but I stumbled.

Suppose we have

```hs
data V a where
  MkV :: forall a -> a -> V a
```

That would seem to be accepted by this proposal. But now it seems we need #281 to be able to use `MkV` in a term. And we haven't begun to talk about a *pattern* with `MkV` in it. (Key question is about scoping -- which namespace gets the `a` in the pattern `MkV a x`?)

I don't think you intended for this to depend on #281, so perhaps you want to say that any constructor with a `forall ... ->` quantifier is (for now) banned in terms and patterns? Note that `forall ... ->` is still useful, if you are going to use the constructor only a compile-time context.

int-index:
> perhaps you want to say that any constructor with a `forall ... ->` quantifier is (for now) banned in terms and patterns? Note that `forall ... ->` is still useful, if you are going to use the constructor only a compile-time context.

@goldfirere Yes, that was the intent. That is exactly what #81 used to say before it was removed in #393. I think we should simply revert #393 and there will be no problem. But I can also mention it in this proposal, too.

goldfirere:

/remind me to accept in 2 weeks if there are no objections.

simonpj:
@int-index, what would go wrong if we allowed parens in `gadt_prefix_sig`?  Thus:
```
gadt_prefix_sig ::= btype
                 |  quantifier gadt_prefix_sig
                 | '(' gadt_prefix_sig ')'
```
Yes, we'd have to represent that in HsSyn, but `gadt_prefix_sig` is already a recursive production, that will require its own recursive type to represent it.  Adding parens seems innocuous.  

Any thoughts on this?

int-index:
> what would go wrong if we allowed parens in `gadt_prefix_sig`? 

It would create a conflict in the `happy` grammar. Upon encountering the opening parenthesis, the parser wouldn’t know whether to expect `gadt_prefix_sig` or `btype` next. Compare:

```
MkX :: Int -> (blah) -> X
MkY :: Int -> (blah)
```

In `MkX`, `blah` is a `btype`. In `MkY`, it is a `gadt_prefix_sig`. We find out about it only when we see (or don’t see) the `->` that follows.

simonpj:
Why does that same conflict not cause a problem with ordinary type signatures:
```
f:: Int -> (blah) -> X
f:: Int -> (blah)
```

int-index:
> Why does that same conflict not cause a problem with ordinary type signatures:

Ordinary type signatures use ordinary type syntax. They don’t need to support strictness annotations and record syntax (see “Constructor-specific syntax” in the proposal), so they don’t require the equivalent of `gadt_prefix_sig`.

Basically, when we go under parentheses, we need to know whether we accept `!a -> b` and `{ ... } -> b`. In function signatures, the answer is “No” and that’s the end of the story. In constructor signatures, the answer is “It depends”, hence the conflict.

gridaphobe:
> Basically, when we go under parentheses, we need to know whether we accept `!a -> b` and `{ ... } -> b`. In function signatures, the answer is “No” and that’s the end of the story. In constructor signatures, the answer is “It depends”, hence the conflict.

So the issue is that in constructor signatures the answer depends on whether we're parsing a top-level argument type or a nested type. Would it work if we always accepted those productions in GADT signatures, and then rejected signatures with nested strictness annotations or records and a post-processing check? It sounds like that might also provide better error messages.

int-index:
> Would it work if we always accepted those productions in GADT signatures, and then rejected signatures with nested strictness annotations or records and a post-processing check? It sounds like that might also provide better error messages.

It would be a tax on the implementation, but yes. There are two approaches to this:

1. In GHC HEAD, we produce `HsRecTy` and `HsBangTy` AST nodes for these productions. It’s really an abstraction leak, as things needed temporarily during parsing end up in the AST definition. Fixing it is actually what prompted me to write the proposal.

2. The alternative solution would be to use a `DisambECP`-style class to resolve this conflict (that’s how we resolve the conflicts between patterns, expressions, and arrow commands). That would keep the AST clean, but it would be an intrusive and complex change to the parser. I could implement it, but I just don’t know if supporting parentheses is worth it.

gridaphobe:
Is (2) using the trees-that-grow stuff? That's one option that came to mind.

Another would be to parse the superset of the grammar as a completely separate type, and then immediately convert it to the real AST or fail. The advantage of doing this would be that we can give a very descriptive error message, rather than the usual "unexpected blah" parse error.

int-index:
@gridaphobe The design space is explored and described here: https://github.com/ghc/ghc/blob/302854154626ef10363afdda3ff1db7160e0827f/compiler/GHC/Parser/PostProcess.hs#L1702-L2080

`Note [Ambiguous syntactic categories]` is the `DisambECP`-style approach I’m referring to, and `Note [Resolving parsing ambiguities: non-taken alternatives]` documents some other ways to solve this issue.

In any case, that’s mostly a matter of implementation. I’m pretty sure that all of these options will result in additional complexity. Is this complexity justified, are parentheses worth it?

simonpj:
>   Ordinary type signatures use ordinary type syntax. They don’t need to support strictness annotations and record syntax (see “Constructor-specific syntax” in the proposal), so they don’t require the equivalent of gadt_prefix_sig.

That still doesn't tell me why it's problematic to add one line to the BNF in your proposal, thus:
```
gadt_prefix_sig ::= btype
                 |  quantifier gadt_prefix_sig
                 | '(' gadt_prefix_sig ')'
```
I think you are saying that some parse is ambiguous, but I don't see it.

simonpj:
Iavor asks (on the ghc-steering-committee list) what is the benefit of allowing nested quantifiers.

And indeed, the proposal is silent on this point.  Can you that?

I think the only compelling reason comes when we have `forall n -> `.  So, if we allow nested parens (see above) the main motivation of the proposal is future compat with `forall n ->`.  Correct?

simonpj:
Richard says

> 1. Dropping the parentheses sounds trivial, but keeping them means we have to do extra work in the design of our data structures to remember the parentheses. Because constructor "types" aren't types, this information isn't otherwise needed.

Is it really true that we have to do significant "extra work"?   Doesn't the structure of `gadt_prefix_sig` suggest
```
data GadtPrefixSig p = BType (HsType p)
                                  |  Quant Quantifier (GadtPrfefixSIg p)
```
in which case adding
```
                                  | Par (GadtPrefixSig p)
```
seems pretty trivial.

(I think Richard's claim *used* to be true, when we stored `[ConArg]` but the nested quantifiers preclude that.

int-index:
> That still doesn't tell me why it's problematic to add one line to the BNF in your proposal. I think you are saying that some parse is ambiguous, but I don't see it.

No, there’s no ambiguity. It would be a valid specification, but not as easy to implement. Using a LALR parser (rather than, say, GLR) imposes its own set of constraints.

Adding the `| '(' gadt_prefix_sig ')'` production to `Parser.y` simply results in reduce/reduce conflicts. `happy` cannot produce a state machine with 1 token of lookahead that could handle it.

> what is the benefit of allowing nested quantifiers.

The idea is to give more freedom to API authors. For example, in type constructors quantifiers come in arbitrary order:

```
data T :: Bool -> forall k. k -> Type
```

The idea is to allow the same sort of flexibility in data constructors. I don’t have a concrete use case in mind, but it seems to make the language more regular.

simonpj:
> Adding the | '(' gadt_prefix_sig ')' production to Parser.y simply results in reduce/reduce conflicts

This part is very not-clear to me, yet is apparently central to the proposal.

Try this:
```
gadt_prefix_sig ::= btype               -- (1)
                 |  '!' atype '->' gadt_prefix_sig
                 |      atype '->' gadt_prefix_sig
                 | '(' gadt_prefix_sig ')'

btype ::= tycon                -- (2)
       | '(' type ')'

type ::= btype
       | btype '->' type
```
This is way more limited than reality, but it illustrates.  I can't see the reduce/reduce conflict.

int-index:
> This is way more limited than reality, but it illustrates. 

I might be missing the core idea behind your code snippet, but it does seem quite similar to what I’ve tried in my branch. My understanding is that source of the problem is that `'(' ktype ')'` is part of `atype`:

```
atype :: { LHsType GhcPs }
        : ntgtycon                    { ... }
        | tyvar %shift                { ... }
        | '*'                         { ... }
        | '(' ktype ')'               { ... }   -- CONFLICT HERE.
        | ...
```

And  `gadt_prefix_sig` can, in fact, start with `atype`. So we need to choose between these two productions:

1. '(' ktype ')'
2. '(' gadt_prefix_sig ')'

If you’re willing to get your hands dirty, here is my patch: https://gitlab.haskell.org/ghc/ghc/-/commit/32711799397ae76a87ba47b1ba6daa42f94fc965

If you add a `| '(' gadt_con_sig ')' { undefined }` production to  `gadt_con_sig`, you can see the issue for yourself.

goldfirere:
> (I think Richard's claim used to be true, when we stored `[ConArg]` but the nested quantifiers preclude that.)

That may be the case (that storing the parens in the AST is fairly easy). I may have been thinking of the current `[ConArg]`-based implementation.

My desire to drop the parentheses is on theoretical grounds, though, not practical ones: the thing after `::` is *not a type*, so there's no reason to believe that parentheses should go there. Or that `->` associates to the right there. `->` is just a piece of punctuation, separating elements in a list, quite unlike the `->` that appears in types, which is a right-associative binary operator.

simonpj:
> My desire to drop the parentheses is on theoretical grounds, though, not practical ones: the thing after :: is not a type, so there's no reason to believe that parentheses should go there.

But that argument would apply equally to nested quantifiers, wouldn't it?   

I'm not sure this line of thinking moves us on much.  Maybe we'd be best off simply asking

* Do we like nested parens in a data con sig? (Proposal says no, others disagree.  No technical issue, just a matter of taste.)
* Do we like nested quantifiers in a data con sig? (Proposal says yes, but only because "it seems to make the language more regular".  Might become more significant if we had dependency.)

Incidentally I note that, under the proposal, if you want record field names you can't have nested quantifiers.   That seems like an unfortunate fork in the road: you can have nested quantifiers, or named fields, but not both. In the status quo the quantifiers have to be at the front so the issue doesn't arise.

-----------------

TL;DR.  I can see merit in this proposal, but little urgency.  Meanwhile it's using up our cycles thinking about it.  Maybe we should park it (entirely without prejudice) until the needs become more pressing.

int-index:
> I can see merit in this proposal, but little urgency. Meanwhile it's using up our cycles thinking about it. Maybe we should park it (entirely without prejudice) until the needs become more pressing.

Whether we drop support for parentheses determines how invasive the changes I need to make in order to complete https://gitlab.haskell.org/ghc/ghc/-/issues/18782. Maybe it’s not urgent, but I’ve already sunk quite some time into it (with a few prototypes and writing this proposal), and I intend to finish it, so a resolution would be nice.

Just to be clear, keeping support for parentheses simply means more engineering work and complexity in the parser, it’s not a blocker. I simply remember talking to @goldfirere, and after hearing his ideological argument I figured it to be a good excuse to keep things simpler.

Let’s decide whether we want parentheses or not, this will inform my further implementation efforts.

goldfirere:
As proposal shepherd, in an attempt to make forward progress, I propose dropping the part of this proposal about the parentheses, retaining only the bit about quantifiers. My sense is that, with that change, the committee would approve. Furthermore, my sense is that @int-index is not particularly opposed to such a change, so this would seem to make everyone happy.

@int-index, please revise the text of the proposal, ping me, and then I will send this back through the committee.

Thanks!

int-index:
@goldfirere OK, I changed the proposal to retain support for parentheses in prefix-style signatures:

```haskell
data G where
  MkG :: !Int -> (!Int -> G)   -- OK
```

Note, however, that I did not _introduce_ support for parentheses record-style signatures. As of GHC 9.0, they are rejected:

```haskell
data G a where
  MkG :: ({ foo :: Int } -> G a)
```

```
<interactive>:3:24: error:
    • Record syntax is illegal here: {foo :: Int}
    • In the type ‘{foo :: Int}’
      In the definition of data constructor ‘MkG’
      In the data declaration for ‘G’
```

Let’s keep it this way, until somebody has a use case for parentheses there.

simonpj:
> Note, however, that I did not introduce support for parentheses record-style signatures. As of GHC 9.0, they are rejected:

yes i think that's fine.   Record syntax also does not support nested quantifiers, I note.

proposals/0000-gadt-syntax.rst
```diff
@@ -0,0 +1,327 @@
...
+    MkT :: Int -> forall a. Eq a => T a
+
+Implementing this proposal as described will fix
+`#18389 <https://gitlab.haskell.org/ghc/ghc/-/issues/18389>`_, and unblock
+`#18782 <https://gitlab.haskell.org/ghc/ghc/-/issues/18782>`_.
+
+Costs and Drawbacks
+-------------------
+
+* The proposed grammar allows parentheses in fewer places than the current
```
goldfirere, line 295: This bullet now appears out of date. Is that correct?

goldfirere:
I've prodded the committee to take another look.

/remind me to accept in 10 days if there is no further conversation.

goldfirere:
This proposal has now been accepted. Thanks, all!

PR #404, opened 2021-02-14
Generalize #229 to include consym

int-index:
Due to an oversight, the proposal talks about `varsym` specifically rather than operators in general (which can be either `varsym` or `consym`). In practice it manifests in an inconsistent behavior of `-Woperator-whitespace`, which I noticed while implementing #368

```haskell
ghci> f = \x y -> x: y
ghci> f = \x y -> x+ y

<interactive>:9:14: warning: [-Woperator-whitespace]
    The suffix use of a ‘+’ might be repurposed as special syntax
      by a future language extension.
    Suggested fix: add whitespace around it.
```

Note the lack of a warning in case of `:`. Likewise, it would not be triggered by any other operator that starts with `:`.

With the proposed amendment, this omission is fixed.

### DIFF
--- a/proposals/0229-whitespace-bang-patterns.rst
+++ b/proposals/0229-whitespace-bang-patterns.rst
@@ -107,14 +107,14 @@ Proposed Change Specification
   ``-XTemplateHaskell`` is also enabled, as well as ``⦇`` as opening and ``⦈``
   as closing if ``-XArrows`` is also enabled.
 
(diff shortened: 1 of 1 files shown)


nomeata:
@goldfirere, as the shepherd of #229, do you think this qualifies as “obviously a bugfix”, or should this be a proposal?

goldfirere:
Bugfix.

PR #405, opened 2021-02-16
Split RecordDotSyntax into two extensions

shayne-fletcher:
The functionality originally to be provided by the single language extension `RecordDotSyntax` is better provided by two orthogonal but related extensions `OverloadedRecordSelection` and `OverloadedRecordUpdate`.

See [!4532(comment 330639)](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4532#note_330639) for context.


phadej:
While it is not hard to deduce, I think proposal should be clear of the effects when only one extension is enabled.

Also if the only motivation is to get one part into 9.2 and the second later (into 9.4). Can't we wait with both for 9.4?

shayne-fletcher:
> While it is not hard to deduce, I think proposal should be clear of the effects when only one extension is enabled.
> 
> Also if the only motivation is to get one part into 9.2 and the second later (into 9.4). Can't we wait with both for 9.4?

Actually we're angling to get both parts in 9.2 however, the `OverloadedRecordUpdate` part will require `RebindableSyntax` or updates will be in error. The restriction of requiring `RebindableSyntax` would get lifted in 9.4.

shayne-fletcher:
> While it is not hard to deduce, I think proposal should be clear of the effects when only one extension is enabled.

Thanks for the feedback. I added this detail to section 2.1.

Topsii:
Just curious the proposal says
> We do not extend pattern matching, although it would be possible for `P{foo.bar=Just x}` to be defined.

Would this possible extension fall under `OverloadedRecordSelection` or `OverloadedRecordUpdate`?
From what I understand pattern matching uses `getField` instead of `setField`; hence it's `OverloadedRecordSelection`. On the other hand the linked comment and the proposal both state the expression `e{lbl₁.lbl₂ = val}` is equivalent to a use of `setField`. That confused me at first and might confuse others too if this proposal is ever extended to pattern matching. 

 Rephrasing it to 
> With the exception of pattern matching  `e{lbl₁.lbl₂ = val}` is equivalent to `e{lbl₁ = (e.lbl₁){lbl₂ = val}}`

does not really make it easier to understand at first glance. Before the split into two extensions this complexity was rather elegantly hidden.

Will there still be a `RecordDotSyntax` extension (later on?) that implies both `OverloadedRecordSelection` and `OverloadedRecordUpdate`? Don't mind me though, I am probably not going to use this extension(s) in the near future.

adamgundry:
> Also if the only motivation is to get one part into 9.2 and the second later (into 9.4).

I think this change is independently useful, even once both are implemented, because it means users have the option to switch on one feature without the other. There are enough changes coming regarding records (e.g. `NoFieldSelectors`, `DuplicateRecordFields` changes, the `HasField` stuff) that I think it makes sense to give users the flexibility to decide which pieces make up their preferred solution.

This may be particularly helpful when porting an existing codebase, because both extensions are non-conservative in different ways: `OverloadedRecordSelection` introduces new syntax but may break programs that do not use whitespace around `.` for composition, whereas `OverloadedRecordUpdate` re-purposes existing syntax in a way that may introduce ambiguity.

> Can't we wait with both for 9.4?

It's not yet certain that `OverloadedRecordUpdate` (without-`RebindableSyntax`) will make 9.4, because the implementation is blocked on changing the design of `setField` via a proposal that I'm still writing. Whereas `OverloadedRecordSelection` is close to being ready now, and there is demand for it even without update, so it would be a shame to delay.

adamgundry:
> > We do not extend pattern matching, although it would be possible for `P{foo.bar=Just x}` to be defined.
> 
> Would this possible extension fall under `OverloadedRecordSelection` or `OverloadedRecordUpdate`?

I think neither, because it isn't overloaded. The data constructor determines which type is meant. So it would really be a small syntactic extension to record pattern-matching syntax. But in any case that would require a separate proposal.

phadej:
@adamgundry so the @shayne-fletcher comment

> .. part will require RebindableSyntax or updates will be in error. The restriction of requiring RebindableSyntax would get lifted in 9.4.

is a white lie?

adamgundry:
@phadej I'd rather say it is a plausible prediction, just not a guarantee. :wink: I'm fairly optimistic we can hit 9.4, but it depends on whether my upcoming proposal gets mired in debate and whether the implementation hits any more blockers.

phadej:
Then I'll ask for `OverloadedRecordUpdate` part to not be merged for `9.2` if its future is dependent on yet unwritten proposal. The proposal doesn't specify the intermediate state, and as a GHC user I'd prefer it to not exist.

At the very least the intermediate state should be brought to the committee for discussion.

ndmitchell:
I wonder about the names of the extensions. I think `RecordDotSelection` might be a better name than `OverloadedRecordSelection`, because we don't overload the existing record selection, we overload the new record dot, in the way it was always suggested. In contrast, `OverloadedRecordUpdate` seems to capture that one well. Maybe the fact they are different is annoying enough they should both be `Overloaded...` or both be `RecordDot...`.

> Then I'll ask for OverloadedRecordUpdate part to not be merged for 9.2 if its future is dependent on yet unwritten proposal.

It's future might change based on future proposals. But as it stands currently, it's based on accepted but not yet fully implemented proposals. However, the `RecordDotSyntax` proposal lays out how it interacts with `RebindableSyntax`, so that part is entirely and accepted clear and fully implemented. (That isn't to say that one half shouldn't be marked more experimental, or even be hidden from users entirely.)

I note that `RecordDotSyntax` isn't in this proposal as an extension. Should it be? I know Adam is keen to give users a toolbox of lots of little pieces of syntax that can be put together to craft working records. I always felt that RecordDotSyntax should imply NoFieldSelectors. Or do we defer a comprehensive "records work" extension as a bundling of ~3 extensions for a future proposal?

simonpj:
I think it would be helpful to summarise the changes embodied by this PR.

shayne-fletcher:
> I think it would be helpful to summarise the changes embodied by this PR.

- Section 2.1 Language Extensions
  - Remove RecordDotSyntax
  - Add OverloadedRecordSelection, OverloadedRecordUpdate
    - Define OverloadedRecordSelection effects
    - Define OverloadedRecordUpdate effects
- Section 2.1.1 Syntax
  - Change occurence of RecordDotSyntax to OverloadedRecordUpdate
- Section 2.1.2 Precedence
  - Provide clarifying text for interpretations of `f M.n.x` and `f M.N.x`
- Section 2.3.1 Lexer
  - Add a sentence explaining '.' in a qualified name is never a field selection
- Section 2.3.2 Parsing
  - Remove `qvarid` from `field` production
- Global:
  - Replace occurences of "RecordDotSyntax" with "the extensions"

shayne-fletcher:
You can view the rendered version [here](https://github.com/shayne-fletcher/ghc-proposals/blob/amend-record-dot-syntax/proposals/0282-record-dot-syntax.rst).

shayne-fletcher:
> `RecordDotSelection` might be a better name than `OverloadedRecordSelection`

I understood the motivation for `OverloadedRecordSelection` is that `getField` can be overloaded.  I also believe @adamgundry is enamoured of having a set of similar named extensions e.g. `OverloadedLabels`, and so on.

In any case,  regarding specific names, let me know and I'll adjust... I'm easy!

adamgundry:
I don't feel terribly strongly about naming, but there is precedent for `Overloaded...` extensions meaning "interpret this piece of syntax via a typeclass". In some cases (`OverloadedLists`, `OverloadedStrings`) the syntax is already valid and it is generalised, while in the case of `OverloadedLabels` the syntax is valid only when the extension is enabled. Perhaps `OverloadedRecordDot`?

> I note that `RecordDotSyntax` isn't in this proposal as an extension. Should it be? I know Adam is keen to give users a toolbox of lots of little pieces of syntax that can be put together to craft working records. I always felt that RecordDotSyntax should imply NoFieldSelectors. Or do we defer a comprehensive "records work" extension as a bundling of ~3 extensions for a future proposal?

I wondered about this. I do generally prefer the approach of defining small orthogonal extensions first, then specifying larger features as the conjunction of such extensions. Moreover my feeling is that `RecordDotSyntax` shouldn't be added as an implemented extension until we have `OverloadedRecordUpdate` fully implemented. So I'd be inclined to defer, and leave open the possibility of changing the set of implied extensions via another proposal.

chreekat:
> I think it makes sense to give users the flexibility to decide which pieces make up their preferred solution.

As a user, the presence of many fine grained possibilities doesn't always help me, because I'm not paged in to the nuanced trade-offs.

> I wondered about this. I do generally prefer the approach of defining small orthogonal extensions first, then specifying larger features as the conjunction of such extensions.

This, however, makes sense. As long as we end up at a satisfying whole, I agree that it's good to build up pieces incrementally. So I hope this proposal is seen as an enabler for RecordDotSyntax rather than a replacement.

danidiaz:
If  `OverloadedRecordSelection` is enabled in a module but `OverloadedRecordUpdate` is not, will type-changing record updates be possible?

adamgundry:
> If `OverloadedRecordSelection` is enabled in a module but `OverloadedRecordUpdate` is not, will type-changing record updates be possible?

Yes, my understanding is that when `OverloadedRecordUpdate` is disabled, a record update expression will be type-checked in the usual way, i.e. the fields will need to uniquely determine a single record datatype but type-changing update of that datatype will be possible.

shayne-fletcher:
@nomeata I'd like to formally ask to turn consideration of this amendment over to the committee now please.

proposals/0282-record-dot-syntax.rst
```diff
@@ -68,15 +69,25 @@ For the specification we focus on the changes to the parsing rules, and
...
+This change adds new language extensions ``OverloadedRecordDot`` and
+``OverloadedRecordUpdate``.
+
+The effect of ``OverloadedRecordDot`` is to:
+  - alllow '``.``' in record exprssions;
+  - desguar field access via '``.``'to ``getField`` expressions;
+The effect of ``OverloadedRecordUpdate`` is to:
+  - desguar record updates to ``setField`` expressions.
+
+A record update expression like ``e{lbl₁.lbl₂ = val}`` requires both
```
simonpj, line 84: That is, `OverloadedRecordUpdate` makes no sense without `OverloadedRecordDot`?   Is that idea that the former will silently be a no-op without the lattter?  (That seems undesirable)

proposals/0282-record-dot-syntax.rst
```diff
@@ -93,9 +104,9 @@ Expression              Equivalent
 
 [Note: ``e{lbl = val}`` is the syntax of a standard H98 record update.
 It’s the nested form introduced by this proposal that is new :
-``e{lbl1.lbl2 = val}``. However, in the event ``RecordDotSyntax`` is in
-effect, note that we propose that ``e{lbl = val}`` desugar to
-``setField @"lbl" e val``].
+``e{lbl1.lbl2 = val}``. However, in the event
+``OverloadedRecordUpdate`` is in effect, note that we propose that
+``e{lbl = val}`` desugar to ``setField @"lbl" e val``].
```
simonpj, line 109: Don't say "we propose"

proposals/0282-record-dot-syntax.rst
```diff
@@ -107,8 +118,8 @@ thus, ``f r.a.b`` parses as ``f (r.a.b)``.
 Expression     Interpretation
 ============== ===================
 ``f r.x``      means ``f (r.x)``
-``f M.n.x``    means ``f (M.n.x)``
-``f M.N.x``    means ``f (M.N.x)``
+``f M.n.x``    means ``f (M.n.x)`` (means ``f (getField @"x" M.n)``)
+``f M.N.x``    means ``f (M.N.x)`` (``M.N.x`` is a qualified name, not a record field selection)
```
simonpj, line 122: I'd like to see a note after this table saying something like this:

"M.N.x" looks ambiguous.  It could mean
* `(M.N).x`; that is, select the `x` field from the (presumably nullary) data constructor `M.N`, or
* The qualified name `M.N.x`, meaning the `x` imported from `M.N`.

The ambiguity is resolved in favour of the former.  If the latter is desired, you can still write `(M.N).x`.

Actually this entire section on precedence has two parts.
1. How to disambiguate `M.N.x` (as written above)
2. How to disambiguate `f r.x` (the main payload of this section in the original proposal.

It'd be good to lay these two points out clearly.

simonpj:
Shayne I have made some suggestions above. Might you account for them before I ask the committee to look?

proposals/0282-record-dot-syntax.rst
```diff
@@ -68,15 +69,25 @@ For the specification we focus on the changes to the parsing rules, and
...
+  - desguar record updates to ``setField`` expressions.
+
+A record update expression like ``e{lbl₁.lbl₂ = val}`` requires both
+extensions be in effect.
 
 2.1.1 Syntax
 ^^^^^^^^^^^^
 
-In the event the language extension is enabled:
+In the event the language extensions are enabled:
```
simonpj, line 90: Which extensions?  There are two

simonpj:
It'd be helpful to add (to the first post in this conversation) links to
* The accepted proposal, rendered
* A rendered form of the proposal after applying this diff

proposals/0282-record-dot-syntax.rst
```diff
@@ -68,15 +69,25 @@ For the specification we focus on the changes to the parsing rules, and
...
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+2.1 Language extensions
+~~~~~~~~~~~~~~~~~~~~~~~
 
-This change adds a new language extension ``RecordDotSyntax``.
+This change adds new language extensions ``OverloadedRecordDot`` and
+``OverloadedRecordUpdate``.
+
+The effect of ``OverloadedRecordDot`` is to:
+  - alllow '``.``' in record exprssions;
```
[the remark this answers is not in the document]
shayne-fletcher: - Fixed

proposals/0282-record-dot-syntax.rst
shayne-fletcher: - No, `OverloadedRecordUpdate` makes sense without `OverloadedRecordDot`;
  - It means that record updates will desugar to `setField` expressions;
  - It means that "nested" updates will be syntax errors (`.` is not allowed);
- I have added a paragraph to clarify what happens in these two cases (section 2.1).

proposals/0282-record-dot-syntax.rst
shayne-fletcher: - Fixed

proposals/0282-record-dot-syntax.rst
shayne-fletcher: - Added the note;
- Split into two points.

proposals/0282-record-dot-syntax.rst
shayne-fletcher: - Changed to "In the event the language extensions ``OverloadedRecordDot` and `OverloadedRecordUpdate` are enabled"

shayne-fletcher:
> It'd be helpful to add (to the first post in this conversation) links to
> 
> * The accepted proposal, rendered
> * A rendered form of the proposal after applying this diff

- Done.

proposals/0282-record-dot-syntax.rst
```diff
@@ -68,15 +69,39 @@ For the specification we focus on the changes to the parsing rules, and
...
+
+This means:
+  - When ``OverloadedRecordDot`` is not enabled:
+
+    - ``e{lbl₁.lbl₂ = val}`` is a syntax error.
+  - When ``OverloadedRecordUpdate`` is enabled without ``OverloadedRecordDot``:
+
+    - ``e{lbl₁} = val`` will desguar to a ``setField`` expression;
+    - ``e{lbl₁.lbl₂ = val}`` is a syntax error.
+  - When both ``OverloadedRecordDot`` and ``OverloadedRecordUpdate`` are enabled:
```
simonpj, line 95: When ``OverloadedRecordDot`` is enabled without ``OverloadedRecordUpdate``:
    - ``e{lbl₁} = val`` is treated exactly as in Haskell 98
    - ``e{lbl₁.lbl₂ = val}`` is rejected

proposals/0282-record-dot-syntax.rst
```diff
@@ -68,15 +69,39 @@ For the specification we focus on the changes to the parsing rules, and
...
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+2.1 Language extensions
+~~~~~~~~~~~~~~~~~~~~~~~
 
-This change adds a new language extension ``RecordDotSyntax``.
+This change adds new language extensions ``OverloadedRecordDot`` and
+``OverloadedRecordUpdate``.
+
+The effect of ``OverloadedRecordDot`` is to:
+  - allow ``.`` in record expressions;
```
simonpj, line 79: I'm troubled about this way of specifying it, because we don't *really* allow dot in updates unless `OverloadedRecordUpdate` is on too.

How about this?

* If `OverloadedRecordDot` is on:
  * The expression `e.lbl` means `getField @"lbl" e`
  * The expression `.lbl` means `getField @"lbl"`
  
  If `OverloadedRecordDot` is not on, these expressions are parsed as uses of the function `(.)`

* If `OverloadedRecordUpdate` is on:
  * The expression `e { lbl = val }` means `setField @"lbl" e val`
  
  If `OverloadedRecordUpdate` is not on, this expression means just what it does in Haskell 98.

* If `OverloadedRecordDot` and `OverloadedRecordUpdate` are on:
  * The expression `e { lbl1.lbl2 = ef }` means `e{ lbl1 = (e.lbl1){lbl2 = val} }`
  
  If either extension is not on, the expression is rejected.

proposals/0282-record-dot-syntax.rst
```diff
@@ -107,13 +131,28 @@ thus, ``f r.a.b`` parses as ``f (r.a.b)``.
...
 
+Disambiguation of ``M.N.x`` is treated as follows.
+
+============== ===================
+Expression     Interpretation
+============== ===================
+``f M.N.x``    means ``f (M.N.x)`` (``M.N.x`` is a qualified name, not a record field selection)
+============== ===================
+
+Note: ``M.N.x`` looks ambiguous. It could mean:
```
simonpj, line 148: Put this first, and not as a "Note".  It's simply a disambiguation rule.

proposals/0282-record-dot-syntax.rst
```diff
@@ -107,13 +131,28 @@ thus, ``f r.a.b`` parses as ``f (r.a.b)``.
...
 ``f (g r) .x`` is illegal
+``f M.n.x``    means ``f (M.n.x)`` (that is, ``f (getField @"x" M.n)``)
 ============== ===================
 
+Disambiguation of ``M.N.x`` is treated as follows.
+
+============== ===================
+Expression     Interpretation
+============== ===================
+``f M.N.x``    means ``f (M.N.x)`` (``M.N.x`` is a qualified name, not a record field selection)
```
simonpj, line 145: Once that other stuff is first, this can just go in the table with the other entries

simonpj:
Shayne I have added more drafting suggestions

proposals/0282-record-dot-syntax.rst
shayne-fletcher: - Replaced the earlier text with your suggestion.

proposals/0282-record-dot-syntax.rst
shayne-fletcher: - Done.

proposals/0282-record-dot-syntax.rst
shayne-fletcher: - Done.

shayne-fletcher:
> Shayne I have added more drafting suggestions

@simonpj  Thank-you. Suggestions implemented.

proposals/0282-record-dot-syntax.rst
```diff
@@ -68,15 +69,37 @@ For the specification we focus on the changes to the parsing rules, and
...
+
+If ``OverloadedRecordUpdate`` is on:
+
+- The expression ``e{lbl = val}`` means ``setField @"lbl" val``;
+- Update expressions with qualified labels like ``r{M.x = val}`` are disallowed.
+
+If ``OverloadedRecordUpdate`` is not on, ``e{lbl = val}`` means just
+what it does in Haskell98.
+
+If ``OverloadedRecordDot`` and ``OverloadedRecordUpdate`` are on:
```
simonpj, line 94: are both on

proposals/0282-record-dot-syntax.rst
```diff
@@ -68,15 +69,37 @@ For the specification we focus on the changes to the parsing rules, and
...
+
+- The expression ``e{lbl₁.lbl₂ = val}`` means ``e{lbl₁ = (e.lbl₁){lbl2 = val}}``.
```
simonpj, line 96: Formatting bug here.  No bullet

proposals/0282-record-dot-syntax.rst
```diff
@@ -68,15 +69,37 @@ For the specification we focus on the changes to the parsing rules, and
...
```
simonpj, line 97: Add "Otherwise the expression ``e {lbl1.lbl2 = val}`` is illegal".

proposals/0282-record-dot-syntax.rst
shayne-fletcher: - Fixed

proposals/0282-record-dot-syntax.rst
shayne-fletcher: - Fixed

proposals/0282-record-dot-syntax.rst
shayne-fletcher: - Done

simonpj:
I understand the *spec* now.  But can you make clearer (maybe in this thread) what is the *motivation* for separating the two?

Presumably we don't want to force users to give two extensions indefinitely.  So sooner or later we'll have `RecordDotSyntax` which implies both `OverloadedRecordDot` and `OverloadedRecordUpdate`.  

Question 1: why not make `RecordDotSyntax` part of this proposal? 

Question 2: why would any user use `OverloadedRecordDot` or `OverloadedRecordUpdate`?  What's in it for them? Why not just use `RecordDotSyntax`?

ndmitchell:
Answer 1: We think that RecordDotSyntax will include these extensions plus some extension that allows multiple field names, e.g. NoFieldSelectors. Which that final extension is has not yet been determined.

simonpj:
Hmm.   So to complete your thought, you don't want to provide `RecordDotSyntax` now, because if you do, and we add `NoFieldSelectors` to `RecordDotSyntax` later, some programs that compile with `RecordDotSyntax` would then stop compiling.

I think it would be helpful to articulate this in the proposal.

It still doesn't answer the question "why separate OverloadedRecordDot and OverloadedRecordUpdate"?  Why not just have `OverloadedRecords` and then later `RecordDotSyntax` implies that plus `NoFieldSelectors`?

shayne-fletcher:
Having two extensions provides more fine grained control. For example, one can enable the use of `.` for for field selection now without affecting record updates. Conversely, one can enable overloaded record updates without giving `.` new meaning.

simonpj:
> Having two extensions provides more fine grained control.

Yes... but that's not a terribly *strong* motivation.   It's something we could do anytime.   But anyway if that is the only motivation for separating them, let's say so.

shayne-fletcher:
I've updated the "Alternatives within this proposal" section. See [7.7 Why two extensions and not just one?](https://github.com/shayne-fletcher/ghc-proposals/blob/amend-record-dot-syntax/proposals/0282-record-dot-syntax.rst#id26) and [7.8 Why not make RecordDotSyntax part of this proposal?](https://github.com/shayne-fletcher/ghc-proposals/blob/amend-record-dot-syntax/proposals/0282-record-dot-syntax.rst#id27) capturing these discussion points.

simonpj:
I recommend acceptance of this proposal, but invite the committee’s view on one point (the final bullet below).  Here is the thinking

* `RecordDotSyntax` is the extension that we will eventually want programmers to user.   It will probably ultimately imply `NoFieldSelectors`.  But we aren’t quite ready make that choice yet.  So we don’t want to specify exactly what `RecordDotSyntax` does yet.

* So we want another, less ambitious, extension to enable record-dot syntax itself, and its desugaring into `getField`; and similarly for record updates.

* This patch to the proposal goes just a little further, by dis-aggregating into two independent extensions, `OverloadedRecordDot` and `OverloadedRecordUpdates`.

* An alternative, if the committee prefers, would be to have a single extension (say, `OverloadedRecords`).

adamgundry:
Regarding motivation for two extensions, see my comment in https://github.com/ghc-proposals/ghc-proposals/pull/405#issuecomment-779872042. I think it plausible that users might want to start using record dot for selection even if they aren't using it for updates (e.g. because they are using lenses, or because they need type-changing update). Moreover overloaded updates won't be ready in time for GHC 9.2 (without `RebindableSyntax`), but selection probably will.

goldfirere:
I have not followed the previous debate, but am responding as a member of the committee. Please feel free to link to arguments above as appropriate.

What's the eventual goal here? Or, rather, what is the menu of possible eventual goals here? Simon presented this proposal to the committee as a stopgap, where we really want `-XRecordDotSyntax`, but we're not there yet, and so we'll go with the two extensions presented here, and then later `-XRecordDotSyntax` will imply these two and (possibly) `-XNoFieldSelectors`. Several questions well up:

* Why are we not ready for `-XRecordDotSyntax`? That is, what are we waiting to learn before proceeding? If we are waiting to learn how these extensions are used in practice, then it makes sense to do something small before we learn how to do something big. But I don't think that's the case -- it seems rather that we're waiting for experience about a *different* extension, `-XNoFieldSelectors`.... which (I think) is near completion.
* If we believe that orthogonal features should not be lumped together (Adam's comment), then why plan to have `-XRecordDotSyntax` lump orthogonal features together? It sounds a bit like `-fglasgow-exts` that we decided was a bad idea. Or maybe like user-definable extension names, which we also decided against.
* One motivation is about GHC 9.2. Does this mean that this proposal is merely to allow a bit more expressiveness in one GHC version? Then we will be saddled with the design herein for much longer than it will have proved useful.

So, while I don't have a problem, per se, with splitting the extension into two, I don't understand the stated motivation. It seems to me that the desire to get this feature into 9.2 is changing the design of our language in perpetuity, which is the tail wagging the dog.

shayne-fletcher:
This amendment does not change any design. It does not add or remove features. It only provides the means of enabling the two features of the original proposal independently of each other. This disentangling provides our users more flexibility, more options, more expressivity for all versions, certainly not just 9.2.

In part the motivation was informed by considering what we can provide in upcoming versions of the compiler. That's true but it's a detail and it's not a negative. In part it was this that led us to the realizations of why this amendment is an improvement over the original.

As for RecordDotSyntax being a bundle of extensions including these and possibly others, as Neil's comments above show (in this discussion and in the original proposal itself) we rather expect there will be case for that. We are however punting for now while all the development on records and user experience with the features that brings come into maturity. We are not proposing anything old or new on that subject here.

Thank-you for your feedback Richard!

yav:
I am not sure how this is supposed to work.  Consider the following example:
```
{-# Language OverloadedRecordDot #-}
data T = T { x :: Int }
data S = S { x :: Int }

f a = a { x = 2 }
```
The proposal says that this should me what it means in Haskell'98.   So, what is the type of `f`?

shayne-fletcher:
> I am not sure how this is supposed to work. Consider the following example:
```haskell
{-# LANGUAGE OverloadedRecordDot, DuplicateRecordFields #-}
data T = T { x :: Int }
data S = S { x :: Int }

f a = a { x = 2 }
```
> The proposal says that this should mean what it means in Haskell'98. So, what is the type of `f`?

It is ill-typed.
```
    • Record update is ambiguous, and requires a type signature
    • In the expression: a {x}
      In an equation for ‘f’: f a = a {x}
  |
5 | f a = a { x = 2 }
  |       ^^^^^^^^^^^
```

simonpj:
> I am not sure how this is supposed to work.

Your example does not involve any dots, so it'd be unaffected by this proposal, no?

adamgundry:
To expand a bit on the motivation, a couple of concrete cases where the split would be useful:
 * If I have a codebase that relies heavily on type-changing record updates (which are allowed with traditional record update), I can't use `OverloadedRecordUpdate` because it does not support type-changing update. Enabling the extension will cause my programs to stop type-checking. However I can enable and use `OverloadedRecordDot` without a problem.
 * If I have a codebase that makes heavy use of function compositions without whitespace, I can't turn on `OverloadedRecordDot` because it changes the parser. But it would be fine to use `OverloadedRecordUpdate`.

These are meaningfully distinct features, even though they were originally proposed together as one `RecordDotSyntax` extension. Given that we have various records-related changes coming soon (`DuplicateRecordFields` changes, `NoFieldSelectors`, my plans to propose changing `HasField`, ...) I think it makes sense to add the individual pieces first, then when we have a bit more experience of using them, figure out how it would be best to group them together.

yav:
OK, I am confused about the myriad of record extensions:

  * I thought that `OverloadedRecordDot` implies `DuplicateRecordFields`, but I guess it does not.  So I guess `OverlodedRecordDot` is just about how to desugar `x.r` and has nothing to do with datatypes (e.g., there is nothing about generating instances for the `Has`) classes.  Is that correct?

On the topic of splitting the two, I agree it would be useful to do so, but I'd like to raise the question if we should have `OverloadedRecordUpdate` at all.   In particular, it changes the meaning of current notation (`x { f = 2 }`) to something more restrictive.  Furthermore the loss of generality is important---type changing record updates are quite useful.   As such, I doubt this would ever be an extension that we'd want turned on by default.  So it would never really be part of a unified language, and my understanding is that this is something we'd like to avoid.  Perhaps a better design would be to have a *different* notation for this kind of monomorphic record update.

shayne-fletcher:
> OK, I am confused about the myriad of record extensions:
> 
> * I thought that `OverloadedRecordDot` implies `DuplicateRecordFields`, but I guess it does not.  So I guess `OverlodedRecordDot` is just about how to desugar `x.r` and has nothing to do with datatypes (e.g., there is nothing about generating instances for the `Has`) classes.  Is that correct?

You are correct.


> On the topic of splitting the two, I agree it would be useful to do so, 

Great!

> but I'd like to raise the question if we should have `OverloadedRecordUpdate` at all. In particular, it changes the meaning of current notation (`x { f = 2 }`) to something more restrictive. Furthermore the loss of generality is important---type changing record updates are quite useful. As such, I doubt this would ever be an extension that we'd want turned on by default. So it would never really be part of a unified language, and my understanding is that this is something we'd like to avoid. Perhaps a better design would be to have a _different_ notation for this kind of monomorphic record update.

Sorry, this was debated when the original proposal was considered and accepted. I don't think it falls into the category of things that are being considered with respect to this amendment.

yav:
@shayne-fletcher can you point me to something that outlines what's the long term plan for the language?   Which of these futures are we aiming for:

  1. Make `OverloadedUpdate` part of the language standard, thus loosing the ability to do type changing update
  2. Make `Language` pragmas part of the actual language specification, and so syntax means different things in different modules depending on the pragmas

Perhaps there is another option, I'd be interested to hear.  There was a lot of pressure to accept the original proposal, and most of the discussion was about the syntactic issues to do with `.` being used as both and operator and a record selector, and I don't recall discussing this particular topic.

shayne-fletcher:
> @shayne-fletcher can you point me to something that outlines what's the long term plan for the language? Which of these futures are we aiming for:
> 
> 1. Make `OverloadedUpdate` part of the language standard, thus loosing the ability to do type changing update
> 2. Make `Language` pragmas part of the actual language specification, and so syntax means different things in different modules depending on the pragmas
> 
> Perhaps there is another option, I'd be interested to hear. There was a lot of pressure to accept the original proposal, and most of the discussion was about the syntactic issues to do with `.` being used as both and operator and a record selector, and I don't recall discussing this particular topic.

Sadly I myself simply can't do this for you @yav , I must defer to others. I do think you have posed very good questions and I too look forward to hearing what can be said in response to them!

ndmitchell:
This point was discussed at length. But GitHub doesn't support the length the conversation grew to, so it's probably infeasible to point you at where. There is reference to it as a drawback in https://ghc-proposals.readthedocs.io/en/latest/proposals/0282-record-dot-syntax.html#id13.

> On the topic of splitting the two, I agree it would be useful to do so, but I'd like to raise the question if we should have OverloadedRecordUpdate at all. In particular, it changes the meaning of current notation (x { f = 2 }) to something more restrictive. Furthermore the loss of generality is important---type changing record updates are quite useful.

Restricting record updates is useful to get decent type checking. Otherwise, given unconstrained types, expressions like `x{f = 2, g = 3}` would end up with ambiguous types in the middle, as they desugar to two `setField` calls. I would personally happily (gladly) sacrifice polymorphic update (which I used ever other year or so) for duplicate field names (which I would use every day).

We could certainly come up with a new syntax for monomorphic vs polymorphic desugaring. And furthermore, I recommend `\Foo{..} -> Foo{f = 2, ..}` as the polymorphic variant 😉.

As to whether there is a way to have polymorphic update, but enough type constraints to make it have decent inference, that's connected to this proposal, but slightly different. If you switch out setField with something that allows type changing update, it would work. Given the RebindableSyntax plus this proposal, we'd have the freedom to experiment.

danidiaz:
If a typeclass like `SetField` exists, the "getter dot" of `OverloadedRecordSelection` is enough to [emulate](https://gist.github.com/danidiaz/a945ff36ddccac0851fea2a4485df350) a form of nested record update:

    class SetField (n :: Symbol) r v | n r -> v where
        modifyField :: (v -> v) -> r -> r

    newtype Setter a b = Setter ((b -> b) -> a)

    (.=) :: Setter a b -> b -> a
    Setter f .= b = f (const b)

    -- Setters in the original record turn into getters of the Setter newtype
    instance SetField n b c => HasField n (Setter a b) (Setter a c) where
        getField (Setter f) = Setter (f . modifyField @n)

    set :: a -> Setter a a
    set a = Setter (\f -> f a)

It could be used like this:

    person' = (set person).address.number .= 4

Annoyingly, it seems to require `UndecidableInstances`.

simonpj:
@yav asks

>     1. Make `OverloadedUpdate` part of the language standard, thus loosing the ability to do type changing update
> 
>     2. Make `Language` pragmas part of the actual language specification, and so syntax means different things in different modules depending on the pragmas

That particular question is a reasonable one, but it could be asked of many other extensions, including `OverloadedStrings`, `OverloadedLists` etc, all of which change the meaning of existing syntax.  I'm not sure we need answer it now.

About overloaded, non-type-changing record update

* We've already accepted it for `RecordDotSyntax`.  Of course, no decision is irrevocable.  We can always change our minds and decide that it should not be part of `RecordDotSyntax`, but that should be another proposal.

* In a way, this current small proposal makes a helpful step in that direction: by separating OverloadedRecordDot and OverloadedRecordUpdate it allows you to have `r.x` while still having the non-overloaded, type-changing record update.  That should be useful for those who want type-changing, but not overloaded, record update, Iavor perhaps among them.

  Perhaps we will then gain experience about the pros and cons, which can inform future decisions.  It's really difficult to anticipate what will matter and what won't; better to find out from experience.

proposals/0282-record-dot-syntax.rst
```diff
@@ -468,6 +500,21 @@ forms of update both of which fail in different corner cases. Instead,
 we recommend use of ``C{foo}`` as a pattern (with ``-XNamedFieldPuns``)
 to extract fields if necessary.
 
+7.7 Why two extensions and not just one?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Having two extensions provides fine grained control. For example, one
+can enable the use of ``.`` for field selection without affecting
+record updates. Conversely, one can enable overloaded record updates
+without giving ``.`` new meaning.
```
aspiwack, line 509: This looks like a pretty weak motivation to have two extensions. After all, extensions have a cost, so we must get a real benefit for that cost.

That being said, my understanding of the thread is that there are, indeed, stronger motivations. At least the fact that the record update syntax is not backward compatible should be brought up. I'm not entirely sure I managed to understand what motivated everyone involved. But it's alright: that's what this section is about, it should document the motivations which have accumulated through the threads (and objections as well). That would be really helpful!

simonpj:
Thanks Arnaud.   I think the alternatives and their motivations are pretty straightforward:

1. Add two extensions, as proposed.   Pro: flexibility for people like Iavor who want type-changing update, but would still like dot-notation.  Pro: orthogonal things are controlled by separate flags.   Con: each has to be documented separately: two flags with one para each, instead of one flag with two paras.  (The implementation cost is zero: it's only a question of which flag to test.)

2. Add one extension (OverloadedRecordFields, say) to do what OverloadedRecordDot and OverloadedRecordUpdate to in this proposal.  Pro: only one extension.   Con: Iavor and others might want dot-notation, but not want to give up type-changing update.

3. Use RecordDotSyntax, and be prepared to change what it means (e.g. add NoFieldSelectors) later.   Pro: only one extension.  Con: changing the meaning of an extension will break programs.

There is nothing deep here.  I don't think it matters much which we choose, but we should choose.  My own recommendation is (1), but if there is a consensus for (2) or (3), or someone wants to suggest another alternative, that would be equally fine.

arybczak:
> This point was discussed at length. But GitHub doesn't support the length the conversation grew to, so it's probably infeasible to point you at where. There is reference to it as a drawback in https://ghc-proposals.readthedocs.io/en/latest/proposals/0282-record-dot-syntax.html#id13.
> 
> > On the topic of splitting the two, I agree it would be useful to do so, but I'd like to raise the question if we should have OverloadedRecordUpdate at all. In particular, it changes the meaning of current notation (x { f = 2 }) to something more restrictive. Furthermore the loss of generality is important---type changing record updates are quite useful.
> 
> Restricting record updates is useful to get decent type checking. Otherwise, given unconstrained types, expressions like `x{f = 2, g = 3}` would end up with ambiguous types in the middle, as they desugar to two `setField` calls. I would personally happily (gladly) sacrifice polymorphic update (which I used ever other year or so) for duplicate field names (which I would use every day).
> 
> We could certainly come up with a new syntax for monomorphic vs polymorphic desugaring. And furthermore, I recommend `\Foo{..} -> Foo{f = 2, ..}` as the polymorphic variant wink.
> 
> As to whether there is a way to have polymorphic update, but enough type constraints to make it have decent inference, that's connected to this proposal, but slightly different. If you switch out setField with something that allows type changing update, it would work. Given the RebindableSyntax plus this proposal, we'd have the freedom to experiment.

There is no need to come up with a different syntax, type-changing updates with good type inference are possible.

 `optics-core-0.4` has them via generics with no special compiler support (though it exploits [dysfunctional dependencies](https://github.com/ghc-proposals/ghc-proposals/pull/374) in order to get them), so there is no reason not to have them in GHC proper (@adamgundry is working on a proposal to extend `SetField` to allow them, based on the experience we gained with `optics`).

Below snippet demonstrates that you can have type-changing updates compatible with overloaded fields with the same type-inference properties as the classic record update syntax.

<details><summary>type-changing.hs</summary>

```haskell
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE DeriveGeneric #-}
{-# LANGUAGE DuplicateRecordFields #-}
{-# LANGUAGE NamedWildCards #-}
module TypeChanging where

import GHC.Generics (Generic)
import Optics.Core

data Pet = Pet
  { name :: String
  } deriving (Show, Generic)

data User ph a = User
  { name :: String
  , age  :: a
  , pet  :: Maybe Pet
  } deriving (Show, Generic)

-- signature inferrable by ghc, can be written down
general :: ( Num b1
           , GField "name" s3 s1 a2 [a3]
           , GField "age" s1 s2 a1 b1
           , GField "pet" s2 b2 a4 (Maybe a5)
           ) => s3 -> b2
general user = user & gfield @"name" .~ []
                    & gfield @"age"  .~ 23
                    & gfield @"pet"  .~ Nothing


-- ../type-changing.hs:33:22: warning: [-Wpartial-type-signatures]
--     • Found type wildcard ‘_ph’ standing for ‘a’
--       Where: ‘a’ is a rigid type variable bound by
--                the inferred type of infer_output :: User a a1 -> User a2 Integer
--                at ../type-changing.hs:34:1-22
--     • In the type signature: infer_output :: User _ph _a -> _output
--    |
-- 33 | infer_output :: User _ph _a -> _output
--    |                      ^^^
--
-- ../type-changing.hs:33:26: warning: [-Wpartial-type-signatures]
--     • Found type wildcard ‘_a’ standing for ‘a1’
--       Where: ‘a1’ is a rigid type variable bound by
--                the inferred type of infer_output :: User a a1 -> User a2 Integer
--                at ../type-changing.hs:34:1-22
--     • In the type signature: infer_output :: User _ph _a -> _output
--    |
-- 33 | infer_output :: User _ph _a -> _output
--    |                          ^^
--
-- ../type-changing.hs:33:32: warning: [-Wpartial-type-signatures]
--     • Found type wildcard ‘_output’ standing for ‘User a2 Integer’
--       Where: ‘a2’ is a rigid type variable bound by
--                the inferred type of infer_output :: User a a1 -> User a2 Integer
--                at ../type-changing.hs:34:1-22
--     • In the type signature: infer_output :: User _ph _a -> _output
--    |
-- 33 | infer_output :: User _ph _a -> _output
infer_output :: User _ph _a -> _output
infer_output = general

-- ../type-changing.hs:36:16: warning: [-Wpartial-type-signatures]
--     • Found type wildcard ‘_input’ standing for ‘User a a1’
--       Where: ‘a’, ‘a1’ are rigid type variables bound by
--                the inferred type of infer_input :: User a a1 -> User a2 Integer
--                at ../type-changing.hs:37:1-21
--     • In the type signature: infer_input :: _input -> User _ph _a
--    |
-- 36 | infer_input :: _input -> User _ph _a
--    |                ^^^^^^
--
-- ../type-changing.hs:36:31: warning: [-Wpartial-type-signatures]
--     • Found type wildcard ‘_ph’ standing for ‘a2’
--       Where: ‘a2’ is a rigid type variable bound by
--                the inferred type of infer_input :: User a a1 -> User a2 Integer
--                at ../type-changing.hs:37:1-21
--     • In the type signature: infer_input :: _input -> User _ph _a
--    |
-- 36 | infer_input :: _input -> User _ph _a
--    |                               ^^^
--
-- ../type-changing.hs:36:35: warning: [-Wpartial-type-signatures]
--     • Found type wildcard ‘_a’ standing for ‘Integer’
--     • In the type signature: infer_input :: _input -> User _ph _a
--    |
-- 36 | infer_input :: _input -> User _ph _a
infer_input :: _input -> User _ph _a
infer_input = general
```

</details>

simonpj:
Just to close the loop on this proposal:

* The committee accepts this modification of the original proposal. Thank you for making various modifications in response to our suggestions.

* There is now work (led by Adam Gundry) on improving `OverloadedRecordUpdate`, so we'd like the user manual to flag that part of the design as experimental and liable to change.

Please merge the change into the original proposal.

I understand that the implementation will be out in GHC 9.2.

Thank you, Shayne, Neil, Adam, for doing this work!

nomeata:
@shayne-fletcher , can I press the merge button, or is there still something to be done here?

shayne-fletcher:
> @shayne-fletcher , can I press the merge button, or is there still something to be done here?

Thanks @nomeata , this is good to merge.

PR #408, opened 2021-03-05
initial: import renaming

phadej:


tek:
Why not reuse `as` instead of `to`? And why add the `renaming` specifier?

```haskell
import Foo (Bar as Baz)
```

seems like a straightforward solution to me.

proposals/0000-import-renaming.md
```diff
@@ -0,0 +1,143 @@
...
+import Data.List (foldl') renaming (foldr1 to unsafeFoldr1)
+```
+
+Renamed names are always imported, `importspec` specifies
+further names to be imported or hidden.
+
+We can rename with `ExplicitNamespaces`
+
+```haskell
+import GHC.Generics (type (:*:) to Prod)
```
JakobBruenker, line 88:
```suggestion
import GHC.Generics renaming (type (:*:) to Prod)
```
I think `renaming` is missing?

proposals/0000-import-renaming.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+We can rename import declaration
+
+```haskell
+import Data.Maybe (Maybe (Just, Nothing) to Option (Some, None))
```
JakobBruenker, line 94:
```suggestion
import Data.Maybe renaming (Maybe (Just, Nothing) to Option (Some, None))
```

proposals/0000-import-renaming.md
```diff
@@ -0,0 +1,143 @@
+---
+author: Oleg Grenrus
+date-accepted: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at pull request 408](https://github.com/ghc-proposals/ghc-proposals/pull/408)
+**After creating the pull request, edit this file again, update the number in
+the link, and delete this bold sentence.**
```
JakobBruenker, line 10:
```suggestion
```

proposals/0000-import-renaming.md
```diff
@@ -0,0 +1,141 @@
...
+## Costs and Drawbacks
+
+Allowing to *export* renamed names would allow a lot,
+but would require GHC interface files to contain
+not only bindings but also renamings.
+
+There is interesting interaction with `RecordDotSyntax`-like features,
+where constraint solver generates instances based on names in scope.
+Yet it wouldn't be possible to rename instances defined by hand.
+(TODO Check: if it's forbidden to shadow automatic instance, the renaming will break that invariant).
```
adamgundry, line 108: Indeed, at the moment hand-written `HasField` instances are required not to overlap with existing fields of the type. This proposal would allow you to introduce use sites where a hand-written `HasField` instance overlaps with the built-in constraint solver behaviour.

I don't think that's a huge problem though: we can just reject such programs as if the constraint solver got stuck due to two overlapping instances. The only dark corner I can see is that if you define a record pattern synonym and an accompanying manual `HasField` instance, renaming would affect the pattern synonym field but not the `HasField` instance.

```hs
module M where
  data T = MkT { foo :: Int }

  instance HasField "bar" T Bool where
    getField _ = True

module N where
  import M renaming (foo to bar)

  hmm = getField @"bar"
```

phadej: @adamgundry how record patterns and `HasField` interact? It's possible to have both now as well? To answer this questio myself: in GHC-9.0 pattern synonyms don't affect `HasField` resolution. It can be argued it should or shouldn't. Should because pattern synonyms are supposed to be indistinguishable from real constructors/fields. shouldn't because things would break & not compose :)

## An example

### A.hs

```haskell
module A where

data Record = Record { recFieldA :: Int, recFieldB :: Bool }
  deriving Show
```

### B.hs

Introduces pseudo fields using `HasField`

```haskell
{-# LANGUAGE DataKinds, MultiParamTypeClasses #-}
module B where

import A
import GHC.Records

instance HasField "recFieldC" Record Int where
    getField = recFieldA
``` 

### C.hs

Introduces pseudo fields using `PatternSynonyms`

```haskell
{-# LANGUAGE PatternSynonyms #-}
module C where

import A

pattern Pattern :: Int -> Bool -> Record
pattern Pattern{recFieldC,recFieldD} = Record recFieldC recFieldD
```

### D.hs

Use site

```haskell
{-# LANGUAGE DataKinds, TypeApplications #-}
module D where

import A
import GHC.Records

import B
import C

record :: Record
record = Record 42 True

example :: Int
example = getField @"recFieldC" record
```

This works because a solver for `HasField` doesn't look for pattern synonyms.

---

This (`ImportRenaming`) proposal would introduce yet another way
to do that

I.e.

```haskell
import A renaming (recFieldA to recFieldC)
```

I guess that **taking renaming into accounts** when solving for
`HasField` would be easier implementation wise,
and less surprising for users,
but it's also possible to consider only real names.

---

**Important observation** (not related to this proposa): With `RecordDotSyntax` which builds on top
of `HasField` the renaming the original fields is a breaking change,
when backward-compatibility is provided by `PatternSynonyms`.

adamgundry: Yes, I agree with all this. And indeed it is a bit of an awkward point that using a record pattern synonym for backwards compatibility can be broken by `HasField`; while working around this with a manual instance is possible, it's hardly pleasant. 

The underlying reason why pattern synonyms don't lead to `HasField` constraints being solved is that they can insert "fields" into arbitrary types (or even be completely polymorphic). Thus they may conflict with the existing solver behaviour or with manual `HasField` instances. But perhaps we should treat all these cases more like overlapping instances at use sites, rather than ruling them out too early.

goldfirere:
I like the general idea here.

What happens if you re-export a renamed thing? Is that allowed? If so, and if two modules both re-export the same renamed thing (to the same name), does an importing module know that importing the renamed name from both re-exporting modules is not ambiguous? (That's a confusing sentence. My guess is that you'll know what I'm getting at.)

I'm a strong 👍 on the `as` syntax inline in the import spec. That syntax resolves a lot of questions I had.

Might be worth pointing to https://agda.readthedocs.io/en/v2.6.1.3/language/module-system.html#name-modifiers, which is the description of this feature in Agda.

Ericson2314:
Along with Agda, Rust is also good prior art (https://doc.rust-lang.org/reference/items/use-declarations.html). It's less that those are good, than Haskell's imports are particularly clumsy.

simonpj:
One reason that we did not allow renaming in the original Haskell 98 spec is that you can achieve almost everything with new definitions.  Instead of
```
import X renaming( f to g, S to T )
```
you can have
```
import X 
g = f
type T = S
```
You didn't use to be able to do that with data constructors, but now you can, with pattern synonyms.  Record fields elude us (but are not visibly covered by the proposal either).

Maybe it's worth a new mechanism do something you can do already, maybe not.  But at least the proposal should point this out explicitly.

Ericson2314:
@simonpj my sense it that fresh definitions are a bit more heavy weight than "pure renames" for example, with `g = f`
- `{-# NOINLINE g #-}` etc. becomes possible.
- `g`'s inferred type may not be fully faithful to the original

I do sympathize that appealing to this stuff can feel like a retreat from equational reasoning, like admitting we have a `Rn` / just alpha-equivalence judgement equality that is stronger than the surface syntax `=`, but I think it's useful.

Ericson2314:
Perhaps more persuasively, let's say we get local modules (#283), and write:
```haskell
module A where

data D = X
f = D.X
```
and
```haskell
module B where
import A 
type E = D
g = E.X -- does this work?
```
I've liked this feature in Rust and C++, but I think it would be hard to implement in Haskell.
```haskell
module B where
import A renaming (D to E)
g = E.X -- does this work?
```
this seems more tractable.

Whether this is an argument for this or against #283 I won't weigh in on!

simonpj:
> @simonpj my sense it that fresh definitions are a bit more heavy weight than "pure renames" for example, with g = f

Yes, perhaps so. But I think the proposal would be stronger if it explicitly explained that much of what is proposed could be done with existing mechanisms -- it's just a matter of convenience, but convenience can matter.

It would be interesting to know if the proposal allows anything that really *cannot* be done today with auxiliary definition.

phadej:
One thing which cannot be renamed with new definition is instance definition, you have to write

```haskell
instance Functor F where
  fmap = ...
```

even if you really really wanted to rename `fmap` to `map`.

However, the record definition

```haskell
return MyData { field1 = ..., field2 = ... }
```

can be altered with `PatternSynonyms`, which however don't interact with `OverloadedRecordDot`, so there is a leak.

(I'll try to comment such situations, and welcome others to do that as well, so we get enough of examples).

PR #412, opened 2021-03-17
Explicit Splice Imports

mpickering:
This proposal suggest a refinement of the import syntax to distinguish between imports which can and can't be used in top-level splices. 

The gist is that an import of the form

```
import splice A
```

can only be used in top-level splices, and other imports can only be used outside of top-level splices. 

[Rendered](https://github.com/mpickering/ghc-proposals/blob/splice-imports/proposals/0000-splice-imports.rst)

### DIFF
--- a/proposals/0000-splice-imports.rst
+++ b/proposals/0000-splice-imports.rst
@@ -25,9 +25,10 @@ actually be used in a top-level splice.
 
 1. Using ``-fno-code``, any module imported by a module using ``TemplateHaskell`` has to be compiled to object
    code, this causes a significant slow down.
-2. IDEs such as haskell-language-server face a similar problem where all imports
-   have to be compiled.
-3. Proposals such as #14905 to increase build parrelism are far less effective
+2. IDEs such as haskell-language-server face a similar problem where normally
+   modules are just typechecked, but when ``TemplateHaskell`` is enabled, a large
+   number of modules have to be cautiously compiled to bytecode.
+3. Proposals such as `#14905 <https://gitlab.haskell.org/ghc/ghc/-/issues/14095>`_ to increase build parrelism are far less effective
    in projects which use ``TemplateHaskell``.
 
 
@@ -36,25 +37,27 @@ Proposed Change
 
 When the new language extension ``ExplicitSpliceImports`` is enabled then a
 new import modifier is added to the import syntax. An import is marked as a "splice"
-import when it is prefixed with ``splice``.
+import when it is prefixed with ``splice``::
 
-::
   {-# LANGUAGE ExplicitSpliceImports #-}
+  module C where
 
   import A
 
   import splice B
 
-Identifiers arising from splice imports are allowed to be used at negative levels, ie, unquoted in a top-level splice.
+The splice modifier indicates to the compiler that a dependent module needs to be compiled
+to object code before the current module can be renamed. A normal, non-splice, import
+needs to only be typechecked before the current module can be renamed.
 
-::
+Identifiers arising from splice imports are allowed to be used at negative levels, ie, unquoted in a top-level splice::
 
+  -- Accepted because B is a splice import
   foo = $(B.qux)
 
 
-But identifers from normal imports are rejected
+But identifers from normal imports are rejected::
 
-::
   -- Rejected, as A is not a splice import
   baz = $(A.zee)
 
@@ -63,13 +66,13 @@ level. For example, the following is legal::
 
   foo = $(B.qid [| A.zee |] )
 
-Because ``A.zee`` is used at level 0.
+Because ``A.zee`` is used at level 0 it doesn't need to be imported using a splice import.
 
 
-When ``TemplateHaskell`` is enabled but NOT ``SpliceImports``, then all imports
+When ``TemplateHaskell`` is enabled but NOT ``ExplicitSpliceImports``, then all imports
 are implicitly treated as splice imports, which matches the current behaviour.
 
-The prelude module is implicitly imported as a splice module so the following is
+The ``Prelude`` module is implicitly imported as a splice module so the following is
 allowed::
 
   zero = $(id [| 0 |])
@@ -90,9 +93,15 @@ Alternatives
   principle that we can learn about what's needed for a module just by looking
   at the import list.
 
+* The extension could only apply to **home** modules, because the benefits of
+  splice imports are when using GHC's ``--make`` mode. As the proposal stands,
+  for uniformity, any module used inside a top-level splice must be marked as
+  a splice module, even if it's an external module.
+
 * Another alternative would be to allow even finer grained control of splice
   imports so that the cases of usage at levels -1 or -2 could be distinguished.
-  This could be useful in some cross-compilation situations.
+  This could be useful in some cross-compilation situations. This is the approach
+  suggested in the `Stage Hygience for Template Haskell proposal <https://github.com/ghc-proposals/ghc-proposals/pull/243>`_.
 
 
 Unresolved Questions
--- a/proposals/0000-splice-imports.rst
+++ b/proposals/0000-splice-imports.rst
@@ -58,7 +58,7 @@ Definitions
 level
   Each expression exists at a level. The level is increased by 1 when
   inside a quote and decreased by 1 inside a splice. Therefore the level of
-  an expression can be calcated as the number of quotes surrounding an expression
+  an expression can be calculated as the number of quotes surrounding an expression
   subtract the number of splices.
 
 top-level splice
@@ -282,4 +282,3 @@ Alternatives
 
 Unresolved Questions
 --------------------
-
--- a/proposals/0000-splice-imports.rst
+++ b/proposals/0000-splice-imports.rst
@@ -110,15 +110,12 @@ Resolution of scopes (often called "renaming") is blind to whether or not an
 identifier was imported with ``splice``. This is important because it will allow
 GHC to emit errors advising the user to modify their import declarations.
 
-The typechecker will be modified to emit errors in the following cases:
-1. It is an error to reference a ``splice`` imported name from a negative
+The typechecker will be modified to emit errors in the following case:
+
+   It is an error to reference a ``splice`` imported name from a negative
    level, and it is an error to reference a non-``splice`` imported name from
    a non-negative level.
 
-2. It is an error to use a name that is both ``splice``-imported and non-``splice``
-   imported. Note that in some cases we could disambiguate these names by
-   inspecting the level at which they appeared, but we do not propose this.
-   See Ambiguity.
 
 Then,
 1. If a module is only available at compile time then the imports are only available in top-level splices.
@@ -333,11 +330,22 @@ Alternatives
 
   Practically, by far the most common situation is 2 stages.
 
- * Since ``ExplicitSpliceImports`` is a no-op when ``TemplateHaskell`` is
+* Since ``ExplicitSpliceImports`` is a no-op when ``TemplateHaskell`` is
    disabled, we could have ``ExplicitSpliceImports`` imply ``TemplateHaskell``.
    There is at least one case where this would be harmful: users may which to
    enable ``ExplicitSpliceImports`` globally for their project, but only
    carefully enable ``TemplateHaskell`` for a small number of modules.
+
+* There are several proposals or the syntax of splice imports. Some have objected
+  that the ``import splice`` suggestion is ungramatical, unlike ``import qualified`` or
+  ``import hiding``.
+
+  One possible alternative is ``$(import Foo)`` to represent a splice import, this
+  syntax clashes with the existing syntax for declaration splices and significantly
+  changes the structure of the import syntax.
+
+  Another alternative suggested was ``import for splice`` which restores the
+  gramatically nature of the import.
 
 
 
--- a/proposals/0000-splice-imports.rst
+++ b/proposals/0000-splice-imports.rst
@@ -33,6 +33,24 @@ actually be used in a top-level splice.
    for all dependencies.
 4. When cross-compiling, all packages need to be compiled for both host and target.
 
+Definitions
+-----------
+
+level
+  Each expression exists at a level. The level is increased by 1 when
+  inside a quote and decreased by 1 inside a splice. Therefore the level of
+  an expression can be calcated as the number of quotes surrounding an expression
+  subtract the number of splices.
+
+top-level splice
+  A splice, where the body is at a negative level or a top-level unadorned
+  declaration splice.
+
+
+home module
+  A module from the package that is currently being compiled.
+
+
 
 
 Proposed Change
@@ -81,6 +99,15 @@ If you require scenario (3) then two imports declarations can be used::
   import C
   import splice C
 
+The syntax for imports is changed in the follow way::
+
+  importdecl :: { LImportDecl GhcPs }
+     : 'import' maybe_src maybe_safe optsplice optqualified maybe_pkg modid optqualified maybeas maybeimpspec
+
+
+The ``splice`` keyword appears before the ``qualified`` keyword but after ``SOURCE``
+and ``SAFE`` pragmas.
+
 
 Specification of ``splice``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -110,10 +137,20 @@ Because ``A.zee`` is used at level 0 it doesn't need to be imported using a spli
 When ``TemplateHaskell`` is enabled but NOT ``ExplicitSpliceImports``, then all imports
 are implicitly additionally imported as splice imports, which matches the current behaviour.
 
-The ``Prelude`` module is implicitly also imported as a splice module so the following is
+If the ``Prelude`` module is implicitly imported then it is also imported as a splice module so the following is
 allowed::
 
   zero = $(id [| 0 |])
+
+If ``NoImplicitPrelude`` is enabled then you have to import ``Prelude`` as a splice
+module as well::
+
+  {-# LANGUAGE NoImplicitPrelude #-}
+
+  import splice Prelude
+
+
+Splice imports can't be rexported, unless they are also imported normally.
 
 
 Drawbacks
--- a/proposals/0000-splice-imports.rst
+++ b/proposals/0000-splice-imports.rst
@@ -14,32 +14,42 @@ Explicit Splice Imports Extension
 ==============
 
 This proposes a new extension, ``ExplicitSpliceImports``, which modifies the
-import syntax so that imports used inside top-level splices are marked explicitly.
+import syntax so that imports which are only used at compile-time are marked explicitly.
 
 
 Motivation
 ----------
 
-If a module enables the ``TemplateHaskell`` then all imported modules are required
-to first be compiled to object code before name resolution can take place. This
-is a major pessimisation because most of the imported identifers will not
-actually be used in a top-level splice.
+The primary goal of this proposal is to distinguish three different ways that
+modules are used:
 
-1. Using ``-fno-code``, any module imported by a module using ``TemplateHaskell`` has to be compiled to object
-   code, this causes a significant slow down.
-2. Projects such as haskell-language-server face a similar problem where normally
-   modules are just typechecked, but when ``TemplateHaskell`` is enabled, a large
-   number of modules have to be cautiously compiled to bytecode.
-3. Proposals such as `#14905 <https://gitlab.haskell.org/ghc/ghc/-/issues/14095>`_ to increase build parallelism are far less effective
+1. Modules whose code is executed only at compile time
+2. Modules whose code is executed only at runtime
+3. Modules whose code is executed both at compile time and runtime.
+
+Distinguishing these 3 different cases has several advantages
+
+1. If a module enables ``TemplateHaskell`` then all imported modules are required
+   to first be compiled to object code before name resolution can take place. This
+   is a major pessimisation because most of the imported identifers will not
+   actually be used in a top-level splice.
+   Proposals (such as `#14905 <https://gitlab.haskell.org/ghc/ghc/-/issues/14095>`_) to increase build parallelism are far less effective
    in projects which use ``TemplateHaskell`` because name resolution depends on code generation
    for all dependencies.
+2. Using ``-fno-code``, any module imported by a module using ``TemplateHaskell`` has to be compiled to object
+   code, this causes a significant slow down.
+3. Projects such as haskell-language-server face a similar problem where normally
+   modules are just typechecked, but when ``TemplateHaskell`` is enabled, a large
+   number of modules have to be cautiously compiled to bytecode.
 4. By using
-   splice imports you can separate the dependencies into those only needed at build-time and
-   those only needed at run-time. This allows you to omit linking against packages
+   splice imports you can separate the dependencies into those only needed at build-time (1) and
+   those only needed at runtime (2). This allows you to omit linking against packages
    only used during build time.
 5. When cross-compiling, all packages currently to be compiled for both host and target,
-   in a similar way to the previous point, only build dependencies need to be built
-   for the host and runtime dependencies for the target.
+   in a similar way to the previous point, only build dependencies (1) need to be built
+   for the host and runtime dependencies for the target (2). If a dependency
+   is used at both compile and runtime (3) then it must also be compiled both
+   for the host and target.
 
 
 Definitions
@@ -63,15 +73,19 @@ home module
 Proposed Change
 ---------------
 
-We would like to distinguish between three ways that an imported identifier can
-used.
+The ``splice`` modifier indicates that a module's imports are available at compile time.
+The absense of the ``splice`` modifier indicates that a module's imports are available
+at runtime.
 
-1. Only available in top-level splices
-2. Not available in top-level splices
-3. Available everwhere
+A ``splice`` imported module allows identifiers to be used at compile time. Compile time
+evaluation is indicated by a top-level splice. Therefore the ``splice`` modifier
+affects name resolution depending on whether we are inside a top-level splice or not.
 
-In order to do this
-the new language extension ``ExplicitSpliceImports`` is introduced which adds a
+1. If a module is only available at compile time then the imports are only available in top-level splices.
+2. If a module is only available at runtime then the imports are not available in top-level splices.
+3. If a module is availabe at both runtime and compile time then the imports are available everwhere.
+
+The new language extension ``ExplicitSpliceImports`` adds a
 new import modifier to the import syntax. An import is marked as a "splice"
 import when it is prefixed with ``splice``::
 
@@ -86,13 +100,13 @@ import when it is prefixed with ``splice``::
   import A
 
 
-The splice modifier indicates to the compiler that identifiers imported from
-the module can **only** be used inside top-level splices (1). When the extension is enabled,
-imports without the splice modifier are not available to be used in top-level splices (2).
-Therefore, in this example, identifiers from ``B`` can **only** be used in top-level splices
+The splice modifier indicates to the compiler that module B is only used at compile time
+and hence the imports can **only** be used inside top-level splices (1). When the extension is enabled,
+imports without the splice modifier are only available at runtime and therefore not available to be used in top-level splices (2).
+In this example, identifiers from ``B`` can **only** be used in top-level splices
 and identifiers from ``A`` can be used everywhere, apart from in top-level splices.
 
-This distinction is important for two reasons:
+To make some of the initial motivation explicit:
 
 1. Now when compiling module ``Main``, despite the fact ``TemplateHaskell`` is enabled,
    we know that only identifers from module ``B`` will be used in top-level splices so
@@ -215,7 +229,7 @@ module as well::
 Splice imports can't be rexported, unless they are also imported normally. Allowing
 splice imports to be exported would turn a build-time only import into a runtime
 export. Maintaining the distinction between things only needed at build-time and
-things only needed at run-time allows project dependencies to be separated in the
+things only needed at runtime allows project dependencies to be separated in the
 same way. This is important for cross-compilation.
 
 
--- a/proposals/0000-splice-imports.rst
+++ b/proposals/0000-splice-imports.rst
@@ -20,36 +20,46 @@ import syntax so that imports which are only used at compile-time are marked exp
 Motivation
 ----------
 
-The primary goal of this proposal is to distinguish three different ways that
-modules are used:
+The primary goal of this proposal is to distinguish, in the module header, three different ways that
+imported module imports are used:
 
-1. Modules whose code is executed only at compile time
-2. Modules whose code is executed only at runtime
-3. Modules whose code is executed both at compile time and runtime.
+1. Imported modules whose code is executed only at compile time;
+2. Imported modules whose code is executed only at runtime;
+3. Imported modules whose code is executed both at compile time and runtime.
 
-Distinguishing these 3 different cases has several advantages
+Distinguishing these 3 different cases has several advantages:
 
-1. If a module enables ``TemplateHaskell`` then all imported modules are required
-   to first be compiled to object code before name resolution can take place. This
-   is a major pessimisation because most of the imported identifers will not
+1. Currently, if a module enables ``TemplateHaskell``, then all imported modules
+   are compiled to object code before name resolution takes place. The ensures that any top level splices that may be encountered are able to be fully evaluated.
+   This is a pessimisation because most of the imported identifiers, which we have taken such pains to ensure we can run, will not
    actually be used in a top-level splice.
    Proposals (such as `#14905 <https://gitlab.haskell.org/ghc/ghc/-/issues/14095>`_) to increase build parallelism are far less effective
    in projects which use ``TemplateHaskell`` because name resolution depends on code generation
    for all dependencies.
-2. Using ``-fno-code``, any module imported by a module using ``TemplateHaskell`` has to be compiled to object
-   code, this causes a significant slow down.
-3. Projects such as haskell-language-server face a similar problem where normally
-   modules are just typechecked, but when ``TemplateHaskell`` is enabled, a large
+   By distinguishing imported modules whose code is executed only at runtime
+   (which in common cases will be a small fraction of imported modules), we are
+   able to improve this pessimisation.
+2. GHC offers an ``-fno-code`` flag that instructs the compiler to parse and
+   typecheck Haskell modules, but not to generate code. The intent is to offer
+   quick feedback to the user. Any module imports of a module using
+   ``TemplateHaskell`` must be compiled to object code, Because of .1 above.
+   This despite the fact that we will not generate object code for the module
+   itself. By distinguishing imported modules whose code is executed only at
+   runtime, we can reduce this unfortunate work significantly, and entirely in many
+   cases.
+3. Projects such as haskell-language-server face similar problems as 2., where they are interested only in the result of type-checking modules, but when ``TemplateHaskell`` is enabled a large
    number of modules have to be cautiously compiled to bytecode.
-4. By using
-   splice imports you can separate the dependencies into those only needed at build-time (1) and
-   those only needed at runtime (2). This allows you to omit linking against packages
-   only used during build time.
-5. When cross-compiling, all packages currently to be compiled for both host and target,
-   in a similar way to the previous point, only build dependencies (1) need to be built
-   for the host and runtime dependencies for the target (2). If a dependency
-   is used at both compile and runtime (3) then it must also be compiled both
-   for the host and target.
+4. By using splice imports we can separate the dependencies into those only needed at build-time (1) and
+   those only needed at runtime (2). We can then link only against those packages needed at runtime.
+5. Currently, when cross-compiling, in modules that use ``TemplateHaskell``, all
+   imported modules must be compiled for both host and target.
+   By distinguishing imported modules used at compile time(i.e. not used at
+   runtime), we can require only those modules to be compiled for the host.
+   Similarly, by distinguising imported modules used at runtime(i.e. not used at
+   compile time), we can require only those modules to be compiled for the
+   target. It can be very hard or impossible to make some packages available on
+   some cross-compile target platforms, so this change would significantly
+   improve the applicability of ``TemplateHaskell`` in these scenarios.
 
 
 Definitions
@@ -61,8 +71,29 @@ level
   an expression can be calculated as the number of quotes surrounding an expression
   subtract the number of splices.
 
+  example:  ``levels``::
+
+    -- foo is at level 0
+    foo = $(let
+      -- bar is at level 1
+      bar = $(let
+        -- baz is at level 2
+        baz = [|
+        -- qux is at level 1
+          qux = [|
+            -- quux is at level 0
+            quux = [|
+              quuz is at level -1
+              quuz = 0
+            |]
+          |]
+        |] in baz
+      ) in bar
+    )
+
+
 top-level splice
-  A splice, where the body is at a negative level or a unadorned
+  A splice, where the body is at a negative level or an unadorned
   declaration splice.
 
 
@@ -109,7 +140,7 @@ and identifiers from ``A`` can be used everywhere, apart from in top-level splic
 To make some of the initial motivation explicit:
 
 1. Now when compiling module ``Main``, despite the fact ``TemplateHaskell`` is enabled,
-   we know that only identifers from module ``B`` will be used in top-level splices so
+   we know that only identifiers from module ``B`` will be used in top-level splices so
    only ``B`` (and its dependencies) needs to compiled to object code before starting to compile ``Main``.
 2. When cross-compiling, only ``A`` needs to be built for the target and ``B``
    only for the host as it is only used at build-time.
@@ -142,7 +173,7 @@ Identifiers arising from splice imports can only be used at negative levels, ie,
   foo' =  B.qux
 
 
-But identifers from normal imports are rejected::
+But identifiers from normal imports are rejected::
 
   -- Rejected, as A is not a splice import and used at level -1
   baz = $(A.zee)
@@ -167,7 +198,7 @@ Resolution of scopes (often called "renaming") is blind to whether or not an
 identifier was imported with ``splice``.
 
 In the case of variables, variables which are splice imported can only be used
-inside a top-level quotation but reported as ambiguous if they clash with any
+inside a top-level quotation but are reported as ambiguous if they clash with any
 other variable in scope, for example::
 
   import A ( x )
@@ -178,7 +209,7 @@ other variable in scope, for example::
 In this case, there is no ambiguity because ``A.x`` isn't allowed to be used in
 the top-level splice, but we still produce an ambiguity error to prevent any confusing
 situations about what is in scope. This position is conservative and allows more
-flexibility in future if it's deemed the restriction should be relaxed.
+flexibility in the future if it's deemed that the restriction should be relaxed.
 
 For instances, a similar situation applies, splice and non-splice imports must
 have a consistent view of imported instances::
@@ -219,15 +250,21 @@ allowed::
   zero = $(id [| 0 |])
 
 If ``NoImplicitPrelude`` is enabled then you have to import ``Prelude`` as a splice
-module as well::
+module as well in order to use names from ``Prelude`` in negative level splices::
 
   {-# LANGUAGE NoImplicitPrelude #-}
 
   import splice Prelude
 
+  -- accepted
+  foo = $(id [|"foo"|])
 
-Splice imports can't be rexported, unless they are also imported normally. Allowing
-splice imports to be exported would turn a build-time only import into a runtime
+  -- rejected
+  foo = id $([|"foo"|])
+
+All exported names are at level 0. Splice imports can't be rexported, unless
+they are also imported normally.
+Allowing splice imports to be exported would turn a build-time only import into a runtime
 export. Maintaining the distinction between things only needed at build-time and
 things only needed at runtime allows project dependencies to be separated in the
 same way. This is important for cross-compilation.
@@ -247,7 +284,7 @@ Alternatives
 * ``splice`` imports could also bring identifiers into scope so that they
   can be used everywhere in a module, not **only** in top-level splices as
   the proposal suggest. This approach is not taken because it means that
-  build-time only dependencies can't be distinguished
+  build-time only dependencies can't be distinguished from runtime dependencies
 
 * Using a pragma rather than a syntactic modifier would fit in better with
   how ``SOURCE`` imports work and make writing backwards compatible code easier::
@@ -257,7 +294,7 @@ Alternatives
 * It might be proposed that an alternative would be to work out which modules
   need to be compiled based on usage inside a module. This would compromise the
   principle that we can learn about what's needed for a module just by looking
-  at the import list.
+  at the import list in the module header.
 
 * The extension could only apply to **home** modules, because the benefits of
   splice imports are when using GHC's ``--make`` mode. As the proposal stands,
@@ -269,8 +306,8 @@ Alternatives
   This could be useful in some cross-compilation situations. This is the approach
   suggested in the `Stage Hygience for Template Haskell proposal <https://github.com/ghc-proposals/ghc-proposals/pull/243>`_.
 
-  The syntax can be extended in a natural way to allow for this by adding an optional
-  integer component which specificies precisely what level the import should be allowed at::
+  The syntax in this proposal can be extended in a natural way to allow for this by adding an optional
+  integer component which specifies precisely what level the imported names should be allowed at::
 
     -- Can be used at -1
     import splice 1 A
--- a/proposals/0000-splice-imports.rst
+++ b/proposals/0000-splice-imports.rst
@@ -28,37 +28,75 @@ actually be used in a top-level splice.
 2. IDEs such as haskell-language-server face a similar problem where normally
    modules are just typechecked, but when ``TemplateHaskell`` is enabled, a large
    number of modules have to be cautiously compiled to bytecode.
-3. Proposals such as `#14905 <https://gitlab.haskell.org/ghc/ghc/-/issues/14095>`_ to increase build parrelism are far less effective
-   in projects which use ``TemplateHaskell``.
+3. Proposals such as `#14905 <https://gitlab.haskell.org/ghc/ghc/-/issues/14095>`_ to increase build parallelism are far less effective
+   in projects which use ``TemplateHaskell`` because renaming depends on code generation
+   for all dependencies.
+4. When cross-compiling, all packages need to be compiled for both host and target.
+
 
 
 Proposed Change
 ---------------
 
-When the new language extension ``ExplicitSpliceImports`` is enabled then a
-new import modifier is added to the import syntax. An import is marked as a "splice"
+We would like to distinguish between three ways that an imported identifier can
+used.
+
+1. Only available in splices
+2. Not available in splices
+3. Available everwhere
+
+In order to do this
+the new language extension ``ExplicitSpliceImports`` is introduced which adds a
+new import modifier to the import syntax. An import is marked as a "splice"
 import when it is prefixed with ``splice``::
 
   {-# LANGUAGE ExplicitSpliceImports #-}
-  module C where
+  {-# LANGUAGE TemplateHaskell #-}
+  module Main where
 
+  -- (1)
   import A
 
+  -- (2)
   import splice B
 
-The splice modifier indicates to the compiler that a dependent module needs to be compiled
-to object code before the current module can be renamed. A normal, non-splice, import
-needs to only be typechecked before the current module can be renamed.
 
-Identifiers arising from splice imports are allowed to be used at negative levels, ie, unquoted in a top-level splice::
+The splice modifier indicates to the compiler that identifiers imported from
+the module can **only** be used inside splices (1). When the extension is enabled,
+imports without the splice modifier are not available to be used in splices (2).
+Therefore, in this example, identifiers from ``B`` can **only** be used in top-level splices
+and identifiers from ``A`` can be used everywhere, apart from in top-level splices.
 
-  -- Accepted because B is a splice import
+This distinction is important for two reasons:
+
+1. Now when compiling module ``Main``, despite the fact ``TemplateHaskell`` is enabled,
+   we know that only identifers from module ``B`` will be used in top-level splices so
+   only ``B`` needs to compiled to object code before starting to compile ``Main``.
+2. When cross-compiling, only ``A`` needs to be built for the target and ``B``
+   only for the host as it is only used at build-time.
+
+If you require scenario (3) then two imports declarations can be used::
+
+  -- (3)
+  import C
+  import splice C
+
+
+Specification of ``splice``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Identifiers arising from splice imports can only be used at negative levels, ie, unquoted in a top-level splice::
+
+  -- Accepted, because B is a splice import and B.qux is used at level -1
   foo = $(B.qux)
+
+  -- Rejected, because B is a splice import and B.qux is used at level 0
+  foo' =  B.qux
 
 
 But identifers from normal imports are rejected::
 
-  -- Rejected, as A is not a splice import
+  -- Rejected, as A is not a splice import and used at level -1
   baz = $(A.zee)
 
 An identifier can appear inside a top-level splice, if it is at a non-negative
@@ -70,9 +108,9 @@ Because ``A.zee`` is used at level 0 it doesn't need to be imported using a spli
 
 
 When ``TemplateHaskell`` is enabled but NOT ``ExplicitSpliceImports``, then all imports
-are implicitly treated as splice imports, which matches the current behaviour.
+are implicitly additionally imported as splice imports, which matches the current behaviour.
 
-The ``Prelude`` module is implicitly imported as a splice module so the following is
+The ``Prelude`` module is implicitly also imported as a splice module so the following is
 allowed::
 
   zero = $(id [| 0 |])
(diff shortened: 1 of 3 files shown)


Ericson2314:
I talked to @mpickering and I support this.

I think this proposal is insufficient to ensure the correct native vs foreign versions of packages is always used for cross compilation, but also that that's fine. This indeed unblocks better parallelism in the presence of TH, and lays a foundation both in terms of the implementation and GHC user mind-share for something like my https://github.com/ghc-proposals/ghc-proposals/pull/243 to do the rest for cross.

That proposal, with its GHC and Cabal portions, had proved to much for people to digest all at once, and likewise to much for me as author assure every morsel was completely baked, so I think approving a design incrementally like this is not only good, but necessary.

goldfirere:
I'm in strong support of this directional of travel, but have a number of nitpicks/clarifying questions:

* Please define "top-level", as it has two possible interpretations: either splice-that-is-not-within-a-bracket or unnested-declaration-splice. I figured out that you mean the former, but I did not know this when starting my read of the proposal.
* Please define levels; the proposal refers to negative level but does not define this.
* Your note about `Prelude`: presumably this is not the case with `-XNoImplicitPrelude`, but it would be good to say.
* Presumably, an `implicit Prelude hiding ( head )` also eliminates `head` from being in scope in splices? Would be good to say something like "If `Prelude` is imported implicitly, then `import splice Prelude` is also in effect, implicitly. Otherwise, any `import` of `Prelude` is duplicated, but with `import splice`."
* Could you include a short BNF of `import` statements? Doing so would, for example, say whether `splice` comes before or after `qualified`.
* Can `splice` go *after* the module name? [This proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0190-module-qualified-syntax.rst) would probably want it to.
* What happens with ambiguity? For example:

```hs
import A ( x )
import splice B ( x )

foo = $( x ) x
```

Is that allowed? It's unambiguous, but perhaps confusing. I think I would prefer names from both bare `import`s and `import splice`s to be in scope, but there are usage restrictions on the names. That would mean that my example would be rejected as ambiguous. Others may disagree here.
* Relatedly, how does this interact with `-XRebindableSyntax`? If I have a splice-imported `fromInteger` and a regular-imported `fromInteger`, does the meaning of `0` in a splice differ from that outside the splice?
* How does this feature interact with re-exports? Presumably, you can't re-export a `splice`-imported thing.
* Please define (briefly) what a home module is. I *think* it's a module in the same package as the one being compiled, but I'm actually not sure.

Thanks -- this is such a nice simplification over #243, and I think it will be much easier to implement and understand.

Ericson2314:
> Otherwise, any import of Prelude is duplicated, but with import splice."

I would prefer not? It was my understanding that `import Prelude` wasn't special-cased except for it replacing the default implicit prelude. I view making one have to do the two separate imports as keeping the special-case as minimal as possible.

> It's unambiguous, but perhaps confusing. I think I would prefer names from both bare imports and import splices to be in scope, but there are usage restrictions on the names. 

So once we get to supporting cross, e.g. `Int` is already not the same `Int` anyways. In general, the stages become completely separate modules that just "happen to share the same module and have dependencies". I think this strong disambiguation rules might just confuse people as to how the phase separation works, rather than prevent confusion.

> How does this feature interact with re-exports? Presumably, you can't re-export a splice-imported thing.

Good catch. I agree.

> Please define (briefly) what a home module is. I think it's a module in the same package as the one being compiled, but I'm actually not sure.

For the record, it is.

-------------------------

> Please define (briefly) what a home module is. I think it's a module in the same package as the one being compiled, but I'm actually not sure.

I'm fine landing this as a first step, but I highly doubt we're going to be able to avoid the extra stuff from #243 and I think it's important that we agree on the general long term direction even if we just doing this for now.

For cross, this proposal solves the following:
```haskell
module C where
import splice B (bar)

x = $(bar) -- foo is from the native platform's A
```
However, it doesn't help us write `A`
```haskell
module B
import {- ?? -} A (foo)
bar = [| foo |] -- what platform?
```
If we follow the evaluation:
```haskell
x = $(bar) = $([| foo |]) = foo
```
we see that `foo` has to come from the foreign platform. But there's no type of import for `A` in `B` that accomplishes that.

That means we at least need `import`, `import quote`, and `import splice`. This corresponds to the `for-syntax` and `for-template` shorthands in Racket (https://docs.racket-lang.org/reference/require.html?q=require).

Finally, if one wants to support `$(... $(...) ...)` and `[| ... [| ... |] ... |]`, one ends up freely generating stages for each integer. That's the general case in Racket and #243.

mpickering:
Thanks for that example @Ericson2314, I now understand what you were staying about cross-compilation on IRC.

Are there not deeper problems here though? If you `import splice A`, then that says we need module `A` to be build for the host. Then looking at the interface file for `A` you find that there is a definition you can use `A.foo :: Q Exp`, at that point there's no information that `foo` contains identifiers from the previous module which was already compiled for a specific platform. 

In fact, I think the whole idea cross-stage persistence for top-level functions is built under the assumption that host = target. I think it's an open research design question about how it should work for heterogeneous cases.  For example, you wouldn't expect CSP to work when generating a C program from a Haskell program.

mpickering:
> I'm in strong support of this directional of travel, but have a number of nitpicks/clarifying questions:
> 
>     * Please define "top-level", as it has two possible interpretations: either splice-that-is-not-within-a-bracket or unnested-declaration-splice. I figured out that you mean the former, but I did not know this when starting my read of the proposal.

Done, I mean both. 
> 
>     * Please define levels; the proposal refers to negative level but does not define this.

Done
> 
>     * Your note about `Prelude`: presumably this is not the case with `-XNoImplicitPrelude`, but it would be good to say.

Clarified.

> 
>     * Presumably, an `implicit Prelude hiding ( head )` also eliminates `head` from being in scope in splices? Would be good to say something like "If `Prelude` is imported implicitly, then `import splice Prelude` is also in effect, implicitly. Otherwise, any `import` of `Prelude` is duplicated, but with `import splice`."


> 
>     * Could you include a short BNF of `import` statements? Doing so would, for example, say whether `splice` comes before or after `qualified`.

I choose (at random) before.
> 
>     * Can `splice` go _after_ the module name? [This proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0190-module-qualified-syntax.rst) would probably want it to.

I have no opinion about this. There are two things which make this awkward. 1. The extension is called `ImportQualifiedPost`, nothing to do with splice imports, would another extension have to be introduced? 2. The proposal already did nothing with package imports or source imports. 

> 
>     * What happens with ambiguity? For example:
> 
> 
> ```haskell
> import A ( x )
> import splice B ( x )
> 
> foo = $( x ) x
> ```
> 
> Is that allowed? It's unambiguous, but perhaps confusing. I think I would prefer names from both bare `import`s and `import splice`s to be in scope, but there are usage restrictions on the names. That would mean that my example would be rejected as ambiguous. Others may disagree here.

At least a warning would be expected I think.

> 
>     * Relatedly, how does this interact with `-XRebindableSyntax`? If I have a splice-imported `fromInteger` and a regular-imported `fromInteger`, does the meaning of `0` in a splice differ from that outside the splice?

This is a good question. I think that is a similar question to your one about ambiguity. You definitely need to splice import a module if it's instance is used in a top-level splice, but whether you allow different instances to be in scope in splices as the rest of the module is a choice of design. It seems like it would be less confusing to insist on the same instances in scope in both places but there's no reason from an implementation perspective. 

> 
>     * How does this feature interact with re-exports? Presumably, you can't re-export a `splice`-imported thing.

That is disallowed.
> 
>     * Please define (briefly) what a home module is. I _think_ it's a module in the same package as the one being compiled, but I'm actually not sure.

Done
> 
> 
> Thanks -- this is such a nice simplification over #243, and I think it will be much easier to implement and understand.

Ericson2314:
Glad it's clear now, @mpickering.

> In fact, I think the whole idea cross-stage persistence for top-level functions is built under the assumption that host = target.

Yes I agree, but conversely.

> I think it's an open research design question about how it should work for heterogeneous cases.

I think it's just sufficient to *not* have cross stage persistence. The racket way should work just fine, and the way build time dependencies work in Nixpkgs is analogous to Racket. I would say the research is done.

adamse:
The main uses of levels in this proposal seem to only go negative, what is the use for positive levels?

proposals/0000-splice-imports.rst
```diff
@@ -33,6 +33,24 @@ actually be used in a top-level splice.
...
+-----------
+
+level
+  Each expression exists at a level. The level is increased by 1 when
+  inside a quote and decreased by 1 inside a splice. Therefore the level of
+  an expression can be calcated as the number of quotes surrounding an expression
+  subtract the number of splices.
+
+top-level splice
+  A splice, where the body is at a negative level or a top-level unadorned
```
goldfirere, line 46: I'm confused by "or" here. It seems that the first set includes the second, making the "or" part redundant. Or am I missing something?

goldfirere:
I don't see any text in the updated proposal clarifying the ambiguity questions around e.g. `foo = $( x ) x`.

Good point about instances -- I hadn't thought about those.

Let's examine this:

```hs
module X where
data X = MkX

module Normal where
import X
instance Show X where show _ = "normal"

module Splice where
import X
instance Show X where show _ = "splice"

module Bottom where
import X (X(..))
import splice X (X(..))
import Normal ()
import splice Splice ()
import splice Language.Haskell.TH.Lib ( stringE )

s1 = show MkX
s2 = $( stringE (show MkX) )
```

Should this be accepted? If we allow different sets of instances to be in scope in different stages, then "yes", with `s1` being "normal" and `s2` being "splice". But if we require the same set of instances to be in scope in both stages, then this is rejected with two possible `Show X` instances.

Ericson2314:
FWIW in the cross case, @goldfirere's example is fine, even necessary. E.g.
```haskell
module X where
  data X = MkX

module Instance where
  import X

  #if aarch64_HOST_OS
  instance Show X where show _ = "are you reading this on a phone?"
  #else
  instance Show X where show _ = "are you reading this on a desktop?"
  #endif

module Bottom where
  import X (X(..))
  import splice X (X(..))
  import Instance ()
  import splice Instance ()
  import splice Language.Haskell.TH.Lib ( stringE )

  s1 = show MkX
  s2 = $( stringE (show MkX) )
```

It's only when we wish to "observe" build == host and "unlock" features like cross stage persistence that we have new coherence obligations across stages.

proposals/0000-splice-imports.rst
mpickering: I don't know how to describe a declaration splice which is not explicitly spliced. For example

```
module M where

import Lens

data A ...

deriveLens 'A
```

The call to `deriveLens 'A` should be understood to be at a negative level (hence `deriveLens` should be `splice` imported) but it isn't surrounded by splices.

mpickering:
I think the conservative position here is to disallow all these examples of ambiguity we have been discussing. If in the future like @Ericson2314 suggests we want to relax the requirements, then it won't break any user code to make the check less stringent. 

> The main uses of levels in this proposal seem to only go negative, what is the use for positive levels?

Positive levels ensure that code generation doesn't depend on information which is not yet available. For example, positive levels
disallow:

```
foo = [| \x -> $(x) |]
```

mpickering:
I updated the proposal now with comments about ambiguity.

Ericson2314:
@mpickering That's fair, but if we are trying to be conservative we should probably straight up ban cross stage persistence too?

mpickering:
@Ericson2314 Cross-stage persistence is to do with promoting identifiers upwards in levels, this proposal is only about splices, not about quotes.

Ericson2314:
@mpickering well what about local definitions using splice imports that are then spliced? That is "morally OK" provided those definitions are used at runtime or exported, but doesn't that work today as cross-stage persistence?

mpickering:
An example would be useful? Do you mean..

```
foo x = $$([|| x ||])
```

There's no CSP there, x is used at the level it is bound.

Ericson2314:
Nevermind, I think it's just ghci, which let me do
```haskell
> id' a = a
> $(id' [| 1 |])
```

simonpj:
Good stuff.  Some thoughts

* Could you number the sections for easy reference, please?  (There is a standard way to automate such numbering.)

* If I splice-import A then A must be compiled to executable code before importing it.  *But so must all modules that A depends on!*  This may seem obvious, but I really think it's worth stating explicitly.

* Under "Proposed change" I see "The splice modifier indicates to the compiler that identifiers imported from the module can only be used inside splices (1)."  I think you have (1) and (2) back to front.

* Sometimes you say "can only be used inside splices" and sometimes "can only be used in top-level splices". I think you mean the latter.  Worth being consistent.

* I wonder if you couldn't specify all this more neatly by saying that
  * Ordinary imports bind variables at level 0
  * Splice imports bind variables at level -1
  * The body of a module is processed starting at level 0
  Then the inability to refer to an ordinary import in a top-level splice becomes simply an ordinary staging error. And it makes sense: ordinary-imported modules can't be executed, and so are available "later" than splice-imported ones.

* That way of looking at the design would strongly suggest that splice-imported variables be available at level 0, i.e. outsides splices. This contradicts the proposal as written, but seems much nicer to me, because it rests on existing logic.   **I strongly urge this change**, unless I have missed something vital.

* Under "Ambiguity of...", I think it would be easier to say
  * Resolution of scopes (often called "renaming") is blind to whether or not an identifier was imported with `splice`.
  And then give examples.

* "Splice imports can't be re-exported, unless they are also imported normally.  Why can't splice-imports be re-exported? I think you might be worried about
  ```
  module M( f ) where { import splice S( f ) }
  module Y where { import M(f); $(f) }
  ```
  Here Y ordinary-imports `f`, but then uses it in a top-level splice.  But that's illegal (stage error) so Y would be rejected. What goes wrong if you delete this prohibition?

mpickering:
@simonpj Thanks for your comments. I have updated the proposal apart from your "strongly suggested change". Distinguishing between things which are only needed at build-time and things which are needed at runtime can further be exploited if we introduce the idea of build dependencies, which will not ultimately be linked into the final executable. I believe this is an important distinction to make and would prefer to keep the design as-is. 

> "Splice imports can't be re-exported, unless they are also imported normally. Why can't splice-imports be re-exported? I think you might be worried about

If you export a splice import in a situation like you describe, it turns from a build-time dependency into a runtime dependency. Splice imported things definitely won't exist at runtime (only the stuff they end up generating). Allowing a splice export would be similar to allowing:

```
module M (foo) where
import splice A

-- Stage error here
foo = $([| A.foo |])
```

simonpj:
> Distinguishing between things which are only needed at build-time and things which are needed at runtime can further be exploited if we introduce the idea of build dependencies, which will not ultimately be linked into the final executable. I believe this is an important distinction to make and would prefer to keep the design as-is.

If this is an important distinction, perhaps you can list it under "Motivation"?   I had no idea that this was one of your goals, and I don't think the proposal mentions it at all.

When saying "Splice imports can't be re-exported, unless they are also imported normally", link back to that particular motivation, since the sole reason for that rule is that motivation.

I'm sad to lose the idea that everything in the proposal can be explained by the existing stating mechanism.  It seems so simple and elegant!  Perhaps you could list this possibility under "alternatives" and explain why you chose not to adopt it.   (I suppose it is possible that the committee might take a different view.)

mpickering:
> I'm sad to lose the idea that everything in the proposal can be explained by the existing stating mechanism. It seems so simple and elegant! Perhaps you could list this possibility under "alternatives" and explain why you chose not to adopt it. (I suppose it is possible that the committee might take a different view.)

In fact, things can be explained in terms of levels still:

```
Level Specification of ``splice``                                               
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                                
                                                                                 
 * Ordinary imports introduce variables at all non-negative levels (>= 0)        
 * Splice imports introduce variables at all negative levels. (< 0) 
```

Ericson2314:
@simonpj The "cumulative stages" you proposed run afowl of cross compilation too: the build-time version of something may be quite different than it's run-time counterpart.

goldfirere:
I'm in favor of the proposal's intent as written, and against @simonpj's strongly urged change.

The key observation is that cross-stage persistence has costs. One particularly big cost is if the stages are actually executed on different architectures, where cross-stage persistence requires two separate binaries. But the idea of avoiding linking some dependencies in the final executable is a nice example of a benefit of avoiding cross-stage persistence. Note that, even between levels 0 and 1, there is a cost to persistence in the form of `Lift` instances.

simonpj:
> The key observation is that cross-stage persistence has costs. One particularly big cost is if the stages are actually executed on different architectures, where cross-stage persistence requires two separate binaries. But the idea of avoiding linking some dependencies in the final executable is a nice example of a benefit of avoiding cross-stage persistence. Note that, even between levels 0 and 1, there is a cost to persistence in the form of Lift instances.

I am open to being persuaded.  But the proposal does not present these arguments.  Could it, please?

mpickering:
@simonpj The proposal states:

> 4. By using splice imports you can separate the dependencies into those only needed at build-time and those only needed at run-time. This allows you to omit linking against packages only used during build time.
> 5. When cross-compiling, all packages currently to be compiled for both host and target, in a similar way to the previous point, only build dependencies need to be built for the host and runtime dependencies for the target.

and later

> 2. When cross-compiling, only A needs to be built for the target and B only for the host as it is only used at build-time.

Finally clarifying the point about re-exporting symbols.

> Splice imports can't be rexported, unless they are also imported normally. Allowing splice imports to be exported would turn a built-time only import into a runtime export. Maintaining the distinction between things only needed at build-time and things only needed at run-time allows project dependencies to be separated in the same way. This is important for cross-compilation.

Do you want some more explanation than this?

simonpj:
> Do you want some more explanation than this?

I was thinking about an explanation like this:

A major goal of this proposal is to distinguish:
A. Modules whose code is executed only at compile time
B. Modules whose code is executed only at runtime
C. Modules whose code is executed both at compile time and runtime.

There are several advantages to making this distinction:

1. If compiler is running on a platform (the host) that is different to
   the platform for which it is compiling (the target), then modules (C)
   that are executed both at compile time and runtime must be compiled
   twice, once for each platform.  If we know which modules are in
   (A,B) we can compile them only once

2. Modules (A) that are needed only at compile time are not needed in
   the final executable, providing a way to reduce the size of that executable.

3. Modules (A,C) that are needed at compile time must be compiled to executable
   code, regardless of `-fno-code`.

4. ...more from your motivation list....and make it a comprehensive list of all the things you want to achieve.  Then you can refer back to it, by number, from the body of the proposal.

mpickering:
I refocused the proposal now in the way that Simon suggested, so that it talks about distinguishing modules which can be used at compile time or runtime rather than modules which can be used in top-level splices or not.

proposals/0000-splice-imports.rst
```diff
@@ -0,0 +1,285 @@
...
+---------
+
+* The user has to be aware of the significance of using splice imports.
+
+
+
+Alternatives
+------------
+
+* ``splice`` imports could also bring indentifiers into scope so that they
```
masaeedu, line 247:
```suggestion
* ``splice`` imports could also bring identifiers into scope so that they
```

proposals/0000-splice-imports.rst
```diff
@@ -0,0 +1,285 @@
...
+   for the host and target.
+
+
+Definitions
+-----------
+
+level
+  Each expression exists at a level. The level is increased by 1 when
+  inside a quote and decreased by 1 inside a splice. Therefore the level of
+  an expression can be calcated as the number of quotes surrounding an expression
```
masaeedu, line 61:
```suggestion
  an expression can be calculated as the number of quotes surrounding an expression
```

proposals/0000-splice-imports.rst
```diff
@@ -0,0 +1,285 @@
...
+  subtract the number of splices.
```
masaeedu, line 62:
```suggestion
  minus the number of splices.
```

proposals/0000-splice-imports.rst
```diff
@@ -0,0 +1,285 @@
...
+
+top-level splice
+  A splice, where the body is at a negative level or a unadorned
```
masaeedu, line 65:
```suggestion
  A splice, where the body is at a negative level or an unadorned
```

proposals/0000-splice-imports.rst
```diff
@@ -0,0 +1,285 @@
...
+
+home module
+  A module from the package that is currently being compiled.
+
+
+Proposed Change
+---------------
+
+The ``splice`` modifier indicates that a module's imports are available at compile time.
+The absense of the ``splice`` modifier indicates that a module's imports are available
```
masaeedu, line 77:
```suggestion
The absence of the ``splice`` modifier indicates that a module's imports are available
```

masaeedu (commented):
Some typos

mpickering:
Thanks @masaeedu and @duog for some updates.

lykahb:
One of the alternatives listed in the proposal is this:
> It might be proposed that an alternative would be to work out which modules need to be compiled based on usage inside a module. This would compromise the principle that we can learn about what's needed for a module just by looking at the import list in the module header.

This alternative could be worth exploring more. We can require that all names used in the top-level splices are declared in imports. When the imports are explicit, it is easier to analyze the usage in the current module. It is more complex than looking at the import declarations alone, but still only relies on the data of one module. 

We can have rules to distinguish the imports:
1. Import is not qualified and does not explicitly list the names. For example, `import X` or `import X hiding (x)`. Classify it is as runtime-only.
2. Import is qualified or explicitly lists names. Then the analysis shows where the imported names are used across the module:
    1. Used only in a top-level splice
    2. Used in a top-level splice and runtime
    3. Used only in runtime

For example,

```
{-# LANGUAGE ExplicitSpliceImports #-}
import ModuleX (x)
import ModuleY as Y
import ModuleZ

-- This declaration only has variables that were imported explicitly.
-- Analysis marks ModuleX and ModuleY as used in splice.
foo = $(Y.f x)

-- Rejected, z is not declared in the imports
bar = $(Y.g z)

-- Analysis marks that ModuleX is used in runtime
baz = z x
```

There are three advantages to this approach:
- No syntax changes required
- Less friction when using imported names in TemplateHaskell and runtime
- It can be done as an optimization without a language extension. If a splice uses a name that was not found in the import declarations, GHC can fall back to the current behavior and compile all imports. The language extension would only disable this fallback and ensure that the modules with TemplateHaskell can be efficiently compiled.

Disadvantages:
- Analysis of usage may be too complex or slow
- The splices provide a way to the user to restrict themselves and get compiler errors if they accidentally depend on something at runtime that is supposed to only be a build time dependency. Thanks @masaeedu for this suggestion.

danidiaz:
In section 1.1, when explaining the first advantage,

> By distinguishing imported modules whose code is executed only at runtime (which in common cases will be a small fraction of imported modules), we are able to improve this pessimisation.

perhaps what's meant is "imported modules whose code is executed only at compile-time" ?

mpickering:
@lykahb The decision about whether a dependency needs to be compiled to object code or not is currently made just by reading the header -- changing this would be quite a difficult internal change I think. 

I think this is ready for committee review now?

mpickering:
@lykahb One option would be to implement a warning about imports only used inside splices and suggesting to change it to a splice import.

lykahb:
@mpickering If compiler could automatically figure out dependencies, it would have a better ergonomics than adding a new extension. Since it is too difficult to implement, I support this proposal.

Adding a warning is a good idea - it makes discovering the extension easier.

Ericson2314:
@lykahb I take it we are in practical agreement "too difficult", but for the cross case that's *not necessary better ergonomics*.

The only way to make the non-disambiguated way work is to build everything twice, which may not be possible and if it is also comes with an ergonomic penalty (more building and rebuilding).

Also, I value specificity on what's needed. "over-importing" means everything is more likely to get tangled over time. And I love the intent conveyed to the reader in nice precise imports.

Anyways, I don't want to pick on your comment in particular, but establish the counter-narrative that Phase-specific imports really are a virtue and not a implementation-specific kludge to be grudgingly accepted at best.

lykahb:
@Ericson2314 

> The only way to make the non-disambiguated way work is to build everything twice

Possibly it is true, but it is not obvious. The compiler could also build a module to a certain point before the dependencies are needed, classify the splice imports, and continue building the module without repeating any work.

The suggestion was only to explore whether we could keep the existing syntax and still _cheaply_ distinguish the imports. I just look for the opportunities to achieve the same goal of having a faster build without extending the language.

Also, the new syntax has its merits and I don't think of it as a kludge. I rather think that improving clarity of imports and increasing build parallelism are different concerns, and one is not necessarily required to solve the other.

Ericson2314:
Information about dependencies is first needed during name resolution. The packages for the different platforms might export different names (CPP, etc.) so we information from both.

mpickering:
Pinging @nomeata just to be sure the right person has been informed about the readiness of this proposal.

nomeata:
Thanks for the helpful ping!

int-index:
Hi. I won't be able to handle this until 10th of May. Sorry in advance for the delay.

mpickering:
Awesome thanks @nomeata and @int-index

nomeata:
Should we find a different shepherd, or is that delay ok to all involved?

mpickering:
I am in no particular hurry.. implementing this proposal will not be straightforward, still a lot of preliminary refactoring to do before that.

nomeata:
/remind @int-index look at this proposal on May 12th

proposals/0000-splice-imports.rst
```diff
@@ -0,0 +1,319 @@
...
+
+1. Currently, if a module enables ``TemplateHaskell``, then all imported modules
+   are compiled to object code before name resolution takes place. The ensures that any top level splices that may be encountered are able to be fully evaluated.
+   This is a pessimisation because most of the imported identifiers, which we have taken such pains to ensure we can run, will not
+   actually be used in a top-level splice.
+   Proposals (such as `#14905 <https://gitlab.haskell.org/ghc/ghc/-/issues/14095>`_) to increase build parallelism are far less effective
+   in projects which use ``TemplateHaskell`` because name resolution depends on code generation
+   for all dependencies.
+   By distinguishing imported modules whose code is executed only at runtime
+   (which in common cases will be a small fraction of imported modules), we are
```
int-index, line 40: Did you mean to say that a small fraction of imported modules will be used only at compile-time? (Currently the text says “at runtime”)

proposals/0000-splice-imports.rst
```diff
@@ -0,0 +1,319 @@
...
+    foo = $(let
+      -- bar is at level -1
+      bar = $(let
+        -- baz is at level -2
+        baz = [|
+        -- qux is at level -1
+          qux = [|
+            -- quux is at level 0
+            quux = [|
+              quuz is at level 1
```
int-index, line 84: this line should be a comment

proposals/0000-splice-imports.rst
```diff
@@ -0,0 +1,319 @@
...
+The new language extension ``ExplicitSpliceImports`` adds a
+new import modifier to the import syntax. An import is marked as a "splice"
+import when it is prefixed with ``splice``::
+
+  {-# LANGUAGE ExplicitSpliceImports #-}
+  {-# LANGUAGE TemplateHaskell #-}
+  module Main where
+
+  -- (1)
+  import splice B
```
int-index, line 126: Bikeshedding: this reads as if you’re importing a splice called `B`. I know splices can’t even have names, but that’s what I see when I read this line. Have other pseudo-keywords been considered? Racket seems to be used `for-syntax`, maybe we want `for-splices` or something?

Since `$` is used for splices, perhaps `import $B` could work.

proposals/0000-splice-imports.rst
```diff
@@ -0,0 +1,319 @@
...
+Proposed Change
+---------------
+
+The ``splice`` modifier indicates that a module's imports are available at compile time.
+The absence of the ``splice`` modifier indicates that a module's imports are available
+at runtime.
+
+A ``splice`` imported module allows identifiers to be used at compile time. Compile time
+evaluation is indicated by a top-level splice. Therefore the ``splice`` modifier
+affects name resolution depending on whether we are inside a top-level splice or not.
```
int-index, line 111: This bit about name resolution is quite worrisome. Let’s say I write

```
import Control.Lens

makeLenses ‘’Foo
```

Do I get an error that `makeLenses` is not in scope?

```
Test.hs:5:1: error:
    Variable not in scope:
      makeLenses
        :: Language.Haskell.TH.Syntax.Name
           -> Language.Haskell.TH.Lib.Internal.DecsQ
```

That’s not a good UX. I think there’s an alternative. Do name resolution as usual, brining all names into regardless of the splice level. But when executing the splice, issue an error if the name isn’t splice-imported:

```
Test.hs:5:1: error:
    ‘makeLenses’ is not available in top-level splices.
    Did you mean to use a splice-import?
```

proposals/0000-splice-imports.rst
```diff
@@ -0,0 +1,319 @@
...
+
+Because ``A.zee`` is used at level 0 it doesn't need to be imported using a splice import.
+
+Level Specification of ``splice``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+* Ordinary imports introduce variables at all non-negative levels (>= 0)
+* Splice imports introduce variables at all negative levels. (< 0)
+
+Ambiguity of instances and variables
```
int-index, line 192: This entire section seems to be in the spirit of my suggestion above about doing name resolution in a level-agnostic manner.

Let’s just explicitly say that levels affect object code generation and linking, but _not_ parsing or name resolution.

proposals/0000-splice-imports.rst
```diff
@@ -0,0 +1,319 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Resolution of scopes (often called "renaming") is blind to whether or not an
```
int-index, line 195: Actually, now I’m somewhat confused because there seems to be a contradiction between these statements in the proposal:

1. “Therefore the `splice` modifier affects name resolution depending on whether we are inside a top-level splice or not.”
2. “Resolution of scopes (often called "renaming") is blind to whether or not an identifier was imported with `splice`.”

Perhaps the design was changed during discussion (which I have not followed closely) but the text was not updated?

int-index:
Overall, I like the proposal, and I intend to recommend acceptance. But first it needs a little bit more polishing up, in particular the story around name resolution isn’t clear to me.

simonpj:
> But first it needs a little bit more polishing up, in particular the story around name resolution isn’t clear to me.

Great.  The right thing to do is to change the status to "Needs revision"; then the author can resubmit when ready, and the committee clock starts from then

proposals/0000-splice-imports.rst
mpickering: Yes, that is the proposed behavior. Updated now to make it clearer.

proposals/0000-splice-imports.rst
mpickering: This should have been clarified. Name resolution is not affected by splice imports (as you suggested above), typechecking is affected, and errors are produced if an import is used at the wrong level.

mpickering:
Proposal has been updated now. (Thanks to my colleague @duog). 

The key part her updated is here:

https://github.com/ghc-proposals/ghc-proposals/pull/412/commits/617170ce7f97794148e0a09fb323e9963d263ed3#diff-c19a0494f4b3cfb78328d72cf0ea016208c135ef6372e49551380ebe61744d99R109

which adds the following clarification. 

```
Resolution of scopes (often called "renaming") is blind to whether or not an
identifier was imported with ``splice``. This is important because it will allow
GHC to emit errors advising the user to modify their import declarations.

The typechecker will be modified to emit errors in the following cases:
1. It is an error to reference a ``splice`` imported name from a negative
   level, and it is an error to reference a non-``splice`` imported name from
   a non-negative level.

2. It is an error to use a name that is both ``splice``-imported and non-``splice``
   imported. Note that in some cases we could disambiguate these names by
   inspecting the level at which they appeared, but we do not propose this.
   See Ambiguity.
```

proposals/0000-splice-imports.rst
```diff
@@ -0,0 +1,345 @@
...
+identifier was imported with ``splice``. This is important because it will allow
+GHC to emit errors advising the user to modify their import declarations.
+
+The typechecker will be modified to emit errors in the following cases:
+1. It is an error to reference a ``splice`` imported name from a negative
+   level, and it is an error to reference a non-``splice`` imported name from
+   a non-negative level.
+
+2. It is an error to use a name that is both ``splice``-imported and non-``splice``
```
int-index, line 118: This is phrased as if I cannot do

```
  import C (x)
  import splice C (x)
```

But of course I can – several other places in the proposal assume that it is possible to import `x` using both a splice-import and a non-splice import. I think you meant to disallow different entities referring to the same name at different levels:

```
  import A (x)
  import splice B (x)  -- different ‘x’
```

That’s what your “Ambiguity” section describes.

Furthermore, I think this has nothing to do with typechecking. The ambiguity error will arise earlier, during name resolution. So you can simply drop this part.

proposals/0000-splice-imports.rst
```diff
@@ -0,0 +1,345 @@
...
+   inspecting the level at which they appeared, but we do not propose this.
+   See Ambiguity.
+
+Then,
+1. If a module is only available at compile time then the imports are only available in top-level splices.
+2. If a module is only available at runtime then the imports are not available in top-level splices.
+3. If a module is available at both runtime and compile time then the imports are available everywhere.
+
+The driver will be modified to ensure that, for modules with
+`-XTemplateHaskell`, object code is generated for ``splice`` imported modules,
```
int-index, line 129:
```suggestion
``-XTemplateHaskell``, object code is generated for ``splice`` imported modules,
```

I think you accidentally used Markdown instead of RST here.

int-index:
> Proposal has been updated now.

Thanks! I see you chose to ignore the comment about syntax I made at https://github.com/ghc-proposals/ghc-proposals/pull/412#discussion_r633253533, where I suggest `$(import M)` instead of `import splice M`. That is of course up to you as the proposal author which syntax you propose, but please include this suggestion in the “Alternatives” section.

Ericson2314:
I'm not so sure about that alternative. That looks to me like splicing a regular important. We can't do that today, but I could see us relaxing that restriction in the future. Similarly I think we will need quote-phase imports eventually, but `[|import M|]` looks like a regular important.

int-index:
> That looks to me like splicing a regular import.

Exactly. So what semantics what you assign to splicing a regular import if not the semantics of a splice-import introduced in this proposal?

mpickering:
@int-index I don't think your proposed syntax is very good. I much prefer the use of a keyword in the same was as safe or qualified imports rather than introducing operator soup. I will update the proposal with the alternative though.

int-index:
> operator soup

It might be operator soup, but it’s the same soup flavor as other splices (e.g. `$(makeLenses ‘’T)`), thus quite indicative of the meaning.

If `$( ... )` means “go up a level” and `[| ... |]` means “go down a level”, why not use this syntax to specify the level of an import?

>  I much prefer the use of a keyword in the same was as safe or qualified imports

That is very reasonable, but you will notice that import syntax is pseudo-grammatical, e.g. we have `import M hiding (T)` rather than `import M hide (T)`, so one might be tempted to read it as pseudo-English. But then `import splice` looks as if you’re importing a splice, whereas the meaning is that you’re importing _for_ splices.

Choosing another keyword would fix that, using `$( ... )` sidesteps the issue.

I like the proposal in any case, but this bit of syntax is a rough edge.

Ericson2314:
> So what semantics would you assign to splicing a regular import if not the semantics of a splice-import introduced in this proposal?

By "splicing a regular import", I meant that other semantics. Here is an example:

```haskell
import splice Config (buildingOnWindows)

$(if buildingForWindow then [| import Windows.X (foo) |] else [| import Unix.X (foo) |])

myFoo = foo
```

We can't express this today, but I think we will get a closer to being able to without ruining downsweep's minimalism after this.

int-index:
@Ericson2314 I don’t see any issue here, your example might as well have been written as

```
$(import Config (buildingOnWindows))

$(if buildingForWindow then [| import Windows.X (foo) |] else [| import Unix.X (foo) |])

myFoo = foo
```

Here, `import Config` is at level `-1`, and the use of `buildingForWindows` is also at level `-1`.

At the same time, `import Windows.X` / `import Unix.X` is at level `0`, and the use of `foo` is also at level `0`.

So the semantics is that a name can be used at the same level as the import from which it comes.

Ericson2314:
@int-index That's right, my apologies. Still, the issue breaks down in the other direction with quote imports:
```haskell
import quote Config (lensstuff)

mkPrism x = [| .... lensstuff ... $x ... |]
```
makes sense but
```haskell
[|import Config (lensstuff)|]

mkPrism x = [| .... lensstuff ... $x ... |]
```
does not. Even if we didn't have the "expressions at the top level are implicitly spliced" rule I still think it would be weird.

int-index:
> the issue breaks down in the other direction with quote imports

Ok, true. The `$(import ...)` idea doesn’t seem so appealing anymore. My criticism of `import splice` still stands, though. I’d be interested to see others come up with potential alternatives.

michaelpj:
`import for splice`/`import for quote`?

Could extend to `import for stage -1` for full stages if we want to get fancy.

mpickering:
`import safe` is similar to `import splice`, I suppose you would have preferred `import safely` if history had been different. I don't really want to get drawn into this discussion about syntax. The committee should decide what they think and not let this question make the proposal process drag out longer than necessary.

mpickering:
I updated the proposal now with the syntax alternatives.

int-index:

As you might already know, the committee’s deliberations are open and can be accessed at https://mail.haskell.org/pipermail/ghc-steering-committee/

proposals/0000-splice-imports.rst
```diff
@@ -0,0 +1,353 @@
...
+  integer component which specifies precisely what level the imported names should be allowed at::
+
+    -- Can be used at -1
+    import splice 1 A
+    -- Can be used at -2
+    import splice 2 A
+
+  Practically, by far the most common situation is 2 stages.
+
+* Since ``ExplicitSpliceImports`` is a no-op when ``TemplateHaskell`` is
```
int-index, line 333: The formatting is off here (see the Rendered version). I think it’s related to the fact that subsequent lines are indented more than this one.

proposals/0000-splice-imports.rst
```diff
@@ -0,0 +1,353 @@
...
+
+
+home module
+  A module from the package that is currently being compiled.
+
+
+Proposed Change
+---------------
+
+The ``splice`` modifier indicates that a module's imports are available at compile time.
```
simonpj, line 105: This is all in a funny order.  What splice modifier?  Surely you want to put the syntax change first?  Currently it is buried many paragraphs later.

proposals/0000-splice-imports.rst
```diff
@@ -0,0 +1,353 @@
...
+The absence of the ``splice`` modifier indicates that a module's imports are available
+at runtime.
```
simonpj, line 107: available *only* at runtime

simonpj:
I understand the motivation but I am unsure about the design.  In particular
```
import M( f )
import splice M ( g )

x = g True + $( f True )
```
I think that under the proposal this will be accepted.  
* Even though  the `import splice M` imports only `g`, I can use `f` in a splice.  
* And even though the `import M` imports only `f`, I can use `g` in a non-splice.

This seems deeply perplexing.   In effect the import list on the two `import M` decls are unioned, while the splice/non-splice story is a completely orthogonal per-module dependency thing.

I can't at the moment see a decent way to resolve this.  I thought of saying that `import splice M( f )` makes `f` in scope only in top-level splices.  I think that could work but it seems quite a bit more complicated.

int-index:
> I think that under the proposal this will be accepted.

Is that so? Indeed, that seems to be a weak spot in the specification, and I don’t think it’s the intent behind it.

> makes `f` in scope only in top-level splices

The proposal specifically says that it has no effect on scope – I believe it has to be this way for error messages to remain decent.

> I can't at the moment see a decent way to resolve this.

What about: for every imported _name_, track the way it was imported (normally, splice-imported, or both). At use sites, validate that its usage is allowed.

simonpj:
> What about: for every imported name, track the way it was imported (normally, splice-imported, or both). At use sites, validate that its usage is allowed.

Well yes, maybe.   But the proposal should spell this out.   And it'll complicate warning for redundant imports, and `-fprint-minimal-imports`.

proposals/0000-splice-imports.rst
mpickering: This is precisely the description of the `syntax` change. The paragraph states that a new modifier called `splice` is introduced to distinguish between imports availble at compile time or runtime.

mpickering: I have rearranged an reworded this section, I think it's got worse with all the clarifications!

mpickering:
> I understand the motivation but I am unsure about the design. In particular
> 
> ```
> import M( f )
> import splice M ( g )
> 
> x = g True + $( f True )
> ```
> 
> I think that under the proposal this will be accepted.
> 

The proposal should not accept this example, there was a mistake in the specification but it now reads (in the updated version)

```
 The typechecker will be modified to emit errors in the following case:          
                                                                                 
    It is an error to reference a non-``splice`` imported name from a negative       
    level, and it is an error to reference a ``splice`` imported name from   
    a non-negative level. 
```

So `g` is referenced at level 0, but it is (only) ``splice`` imported, hence error.
Likewise `f` is reference at level -1 but it is (only) non-``splice`` imported hence error.

Complications to other features can not be completely forseen at this stage of the proposal. Once the implementation proceeds then it will be obvious how to deal with these features.

proposals/0000-splice-imports.rst
```diff
@@ -0,0 +1,349 @@
...
+
+* It might be proposed that an alternative would be to work out which modules
+  need to be compiled based on usage inside a module. This would compromise the
+  principle that we can learn about what's needed for a module just by looking
+  at the import list in the module header.
+
+* The extension could only apply to **home** modules, because the benefits of
+  splice imports are when using GHC's ``--make`` mode. As the proposal stands,
+  for uniformity, any module used inside a top-level splice must be marked as
+  a splice module, even if it's an external module.
```
aspiwack, line 312: You haven't touched on it in the motivation, but it's worth mentioning that restricting the distinction to home modules could be rather costly for monorepos with several package, as per https://github.com/ghc-proposals/ghc-proposals/pull/243#issuecomment-534125177 .

proposals/0000-splice-imports.rst
```diff
@@ -0,0 +1,349 @@
...
+    -- Can be used at -1
+    import splice 1 A
+    -- Can be used at -2
+    import splice 2 A
+
+  Practically, by far the most common situation is 2 stages.
+
+* Since ``ExplicitSpliceImports`` is a no-op when ``TemplateHaskell`` is
+  disabled, we could have ``ExplicitSpliceImports`` imply ``TemplateHaskell``.
+  There is at least one case where this would be harmful: users may which to
```
aspiwack, line 331:
```suggestion
  There is at least one case where this would be harmful: users may wish to
```

proposals/0000-splice-imports.rst
```diff
@@ -0,0 +1,349 @@
...
+  enable ``ExplicitSpliceImports`` globally for their project, but only
+  carefully enable ``TemplateHaskell`` for a small number of modules.
+
+* There are several proposals or the syntax of splice imports. Some have objected
+  that the ``import splice`` suggestion is ungramatical, unlike ``import qualified`` or
```
aspiwack, line 336:
```suggestion
  that the ``import splice`` suggestion is ungrammatical, unlike ``import qualified`` or
```

aspiwack (commented):
You don't seem to mention it, but this phase distinction is also a staple of one of the most successful macro systems: Racket's. Where importing a module to be used in macro is written as `(require (for-syntax module))` (See for instance this [blog post](http://blog.ezyang.com/2016/07/what-template-haskell-gets-wrong-and-racket-gets-right/) by @ezyang).

Probably worth adding a word about this in the _Motivations_ section.

(some wibbles below)

proposals/0000-splice-imports.rst
```diff
@@ -0,0 +1,349 @@
...
+  s1 = show MkX
+  s2 = $( stringE (show MkX) )
+
+This program is also rejected because the instances defined in ``Normal`` and ``Splice`` overlap.
+
+
+Other Considerations
+~~~~~~~~~~~~~~~~~~~~
+
+When ``TemplateHaskell`` is disabled, then ``ExplicitSpliceImports`` is a no-op.
```
goldfirere, line 250: I don't think that's quite true. I could say

```hs
{-# LANGUAGE ExplicitSpliceImports #-}

import splice Foogle

x = 5
```

This module looks correct to me (if a bit silly). Without the `LANGUAGE` pragma, it would be rejected.

masaeedu: Wouldn't it still be rejected if `TemplateHaskell` is disabled? What would `import splice` mean in the absence of TH?

simonpj:
I'm generally in support.

* "The splice modifier indicates to the compiler that module B is only used at compile time and hence the imports can only be used inside top-level splices (1). When the extension is enabled, imports without the splice modifier are only available at runtime and therefore not available to be used in top-level splices (2)" .  I found this hard to grok.   Try this:
  * Entities imported via `import splice M` can be used only in top-level splices (i.e. at negative levels)
  * Entities imported via `import M` can be used only outside top-level splices (i.e. as levels >= 0)

* I'm worried about type-class instances.  If you use an instance, you need to link to the module from whence it came.   So to deliver on your goals:
  * Instances imported (perhaps transitively) by `import splice M` can be used only in top-level splices
  * Instances imported (perhaps transitively) by `import M` can be used only outside top-level splices.

  Do you agree?  This sounds as if it might be pretty hard to implement.  At the moment we do not record how an instance is imported.  For the rest of the proposal, implementation seems fairly easy.  But I'm a bit concerned that instances may not be.

  We could wonder about type-family instances too.  And (soon, maybe) defaults...

I can see the motivation, but I have lost a clear picture of how much implementation complexity this will impose.

Ericson2314:
@simonpj I am hoping we can make each level it's own virtual module, as far as the implementation is concered. So as opposed to trying hard to keep things separate, we are separate by default.

The challenge then because the interactions between names, e.g. quotes in the negative stage need to be name-resolved in the following stage. I think we can pull that off, however.

mpickering:
I started implementing the proposal and it's not clear to me now how splice imports should work. 

The problem is that if I splice import an identifier then, the result of running the splice can also contain identifiers which come from splice imported modules. Therefore we would need to link against the module where the splice import came from.

For example 

```
module A where

data X = X

x = [| x |]
```

then

```
module B where

import splice A

-- y = X, so need to link against A, which defines X.
y = $(x)
```

Two ways to solve this:

1. Allow splice imported identifiers to be used anywhere
2. Also check the generated code for whether all the identifiers are non-splice imported.

I think I prefer option 1.

Ericson2314:
This is why we need quote imports. If you can only use splice imports and splices and quote imports and in quotes, everything cancels out nicely, the phase separation is proper, and the driver has all the information it needs.

Ericson2314:
(A splice import means we must compile the import all the way to the end just to compile the current module half way (Tc). A quote import means we need only compile the import half way even if we are compiling the current module all the way to the end. But it crates a transitive dependency so quote import + splice import = (transitive) regular import, and then quote import gets forced the rest of the way.)

simonpj:
> Allow splice imported identifiers to be used anywhere

That does seem inevitable, but is quite a disappointing conclusion, given the goal of this proposal to cleanly separate what is needed at compile time from runtime.

Suppose we have
```
module CompileTime( foo ) where
   import RunTime( f ) -- Call this a quote-import
   import OtherStuff
   foo n = .....lots of code... [| f |] ....
```
If a module `M` splice-imports `CompileTime`, then it needs compile-time versions of `CompileTime` and `OtherStuff`; but not `RunTime`.   On the other hand, at run-time we will need `M` and `RunTime`.

Let's call `RunTime` a "quote import".  It imports things used only in quotes.

So a splice-import of `CompileTime` tells us that we have a run-time dependency on the quote-imports of `CompileTime`, rather than (as previously thought) no run-time dependencies.

This makes my head spin a bit, but maybe there is something here.

Ericson2314:
@aspiwack the racket solution is the "quote import", see https://docs.racket-lang.org/reference/require.html?q=require and the `for-template` and `for-syntax` shorthands in particular.

Ericson2314:
The phase-offsetting when working with transitive dependencies, behind being how Racket works, is also what we did with Nixpkgs. As far is I know, it's the one and only way to preserve the proper separation of phases across module boundaries.

Ericson2314:
> This makes my head spin a bit, but maybe there is something here.

No doubt further refining "compile-time" and "run-time" to specific compiler phases as outlined in https://github.com/ghc-proposals/ghc-proposals/pull/412#issuecomment-903930805 will make the head-spinning worse. But I do think it's a really good way to have the implementation and design police each other.

goldfirere:
I don't see it. That is, I don't see why `import splice M` must imply `import M` for this all to work. *Some* splice-imports will contain splices that expand out to include other definitions from `M`, but not all will. If a particular splice requires a particular definition, then that definition must be available in a regular import. I say "available" -- not necessarily in scope. Here is an example:

```hs
module A where

data X = MkX
```

```hs
module B where
import A

q = [| MkX |]
```

```hs
module C where
import splice B
import A ()   -- this is the key line

mkx = $q
```

The "key line" above says that `A` must be available at runtime in `C`. It's almost as if defining

```hs
module Z where
z = 5
```

in some module does two things: it creates an *entity* with original name `Z.z`. Entities have only original names, actually. The definition also creates a name `z`, exported from `Z` that refers to the entity `Z.z`. Entities are exported and imported just like instances: they are always exported (and re-exported) from a module and always imported. We have tighter control of names, though. Under this interpretation, all entities in the result of a splice must be in scope in a normal import.

Is this too complicated? Maybe. It does seem, to me, somewhat simpler than quote-imports.

Ericson2314:
@goldfirere If I understand you correctly, your design means that if `B` changes, then consumers like `C` will need to adjust their  `import A ()` and similar. Do we really want that?

I feel like we should just do the Racket thing -- TH's biggest problem is not any one technical issue, I would say, but rather the fact that it is only in partial dialogue other langauge's macro efforts, especially as macros are more important and more studied over there than over here. I am excited about this sort of thing as a way to repair that conversation as much as I am about it fixing any particular issue.

treeowl:
I do broadly think that "Just do what Racket does" is a very good default position on macro systems. That said, Racket has a serious module system and Haskell doesn't, so the situations are not entirely identical.

int-index:
Seems like there are some design questions to be resolved. Do we want quote imports or not? Assigning back for revision until the proposal takes a concrete stance on this. Hopefully the implementation efforts will also yield further insights.

mpickering:
How I understand "the racket way" is like this:

Quote and splice imports express relative offsets for the current module. 

* A splice import says I need this import a offset -1 to the current module
* A quote import says I need the import at offset +1 to the current module

When the module is compiled, we ask to compile it at a specific level, usually 0. This then fixes the level all the imports are required it, which fixes the level all it's imports are required at and so on. The result is a build graph which dictates at which levels all the modules must be compiled at. 

In a future where GHC is multi-target, then each level can be for a different target. 

## CSP

Cross-stage persistence adds the assumption that if a module is available at n then it is available at n + 1. Therefore an import which is implicitly persisted using top-level CSP adds a dependency on both `n` and `n + 1` implicitly.

## Current Restrictions

At the moment GHC is not multi-target in the sense that we can switch generating code for different architectures at runtime. There are however two different modes which can be configured at runtime.

* `-fno-code`, don't produce any object files
* Normal compilation, produce object files

It only makes sense to use -fno-code at the last stage, as you wouldn't be able to execute any of the prior stages if you don't generate any code. 

So if you set the top-level target to be compiled with `-fno-code` then if you encounter an `import splice` that will require the import is at stage -1, which is a object code stage. If that module `import quote`s a module, that module will only be required at level 0, a -fno-code stage, so that module doesn't have to be compiled to object code, just typechecked. 

## Example

Compiling the top module with `-fno-code`, 

```
module A where

import splice B

foo = $(b)
```

```
module B where

import quote C

b = [| c |]
```

```
module C where

c = 0
```

```
> ghc -fno-code A
[1 of 3] C (Nothing)
[2 of 3] B (B.o)
[3 of 3] A (Nothing)
```

simonpj:
> I started implementing the proposal and it's not clear to me now how splice imports should work.

In my head at least, this proposal is now back with the author for revision, in the light of this discovery.  It's not easy to see a way through here.  I agree that Racket is an excellent place to look.

Ericson2314:
> Cross-stage persistence

Is incompatible with cross compilation, so I rather have a design that doesn't assume/require it so we can at knobs for that later.

> At the moment GHC is not multi-target in the sense that we can switch generating code for different architectures at runtime.

Just a nit, since your `-fno-code` shows we should care about such things anyways, but GHC is *quite* close to that. In particular, for a few release now, everything but the native word size has been not hard-coded and instead controlled by the settings file. That last remaining hurdle we are also edging towards.

Ericson2314:
To add on to @mpickering's good example, with more fine-grained scheduling, I would hope to see something like:
```
> ghc -fno-code A
[1 of 4] C Tc
[2 of 4] B Tc
[2 of 4] B Cg (B.o)
[3 of 4] A Tc
```
```
> ghc A
[1 of 6] C Tc
[2 of 6] B Tc
[2 of 6] B Cg (B.o)
[3 of 6] A Tc
[2 of 6] C Cg (C.o)
[2 of 6] A Cg (A.o)
```

int-index:
Yeah, I agree that:

* We should copy what Racket does until there’s a reason to do otherwise. Or at least we should pick a forwards-compatible subset of it.
* Cross-compilation is an important use case to support, and the design should facilitate it, not hinder it. (I’m happy to trust @Ericson2314’s judgement in this regard).

sgraf812:
I'm all for *starting* with what Racket does, but we should at all times question whether we can make it simpler or need to do it differently due to our use case (that is: metaprogramming and not module systems).

So I'm all for sketching out how Racket-style quote imports and splice imports should work in Template Haskell.
Part of that is figuring out how cross-stage persistence as implemented by `Lift` should work, which Racket does not have or need, as far as I can tell. (Perhaps it is implicitly part of its module system, though.)

We should really figure out how cross-stage persistence interacts with `import splice`/`import quote`, and which role `Lift` plays for cross-compiling. For example, if I write

```hs
module Def where

n :: Int
n = 0
```
```hs
module Use where

import Def
import Language.Haskell.TH.Syntax

laterN :: Q Exp
laterN = [| n |]

nowN :: Q Exp
nowN = [| $(lift n) |]
  -- = lift n
```

The bindings `laterN` and `nowN` desugar to

```
nowN = lift $fLiftBoxedRepInt $fQuoteQ n

laterN
  = varE
      $fQuoteQ
      (mkNameG_v
         (unpackCString# "main"#)
         (unpackCString# "Def"#)
         (unpackCString# "n"#))
```

So `nowN` needs the `n` from stage 0 (it occurs at inside a quote inside a bracket, after all!), whereas `laterN` needs it only in stage 1 (quote stage). If it were only for `laterN`, we could use `import quote Def`, but not so with `laterN`.

Thus, implicit use of `lift` (such as in `f x = [| x + 1 |]`, which desugars to `f x = [| $(lift x) + 1 |]`) must be treated like `$(lift x)`, requiring `x` at stage offset -1. (Do we implicitly lift top-level identifiers such as `t` as well? Probably not!)

That would also allow us to understand `Lift` for what it is: It lifts some value from host stage 0 into a datum of the target stage 1, **without needing to run code on the target**. It is the job of the `Lift` instance to do this in a cross-compatible way (taking care of integer overflows, etc.). Explicit `lift`ing by the user is their way to signal that some logic should be executed on the host and then converted into some datum for the target.

To conclude, I don't think we need to ban cross-stage persistance/`Lift`. I think we can and should allow the following program, which requires `n` only on the host (note the `import splice`):

```hs
module OtherUse where

import splice Def
import Language.Haskell.TH.Syntax

copy :: Int
copy = $(lift n)
-- expands to `blub = I# 0#`
```

sgraf812:
I have a feeling that the `import quote`/`import splice` distinction could easily be *inferred* by the compiler in the renaming phase (it can be checked, after all!).

But the appeal of this extension is perhaps to make the information evident in the module header, so that a static build plan can be constructed that says "compile module A for the target, module B for the host and module C for both" à priori, just by looking at the header. 

This is a bit simpler than starting to rename module C, which would then need to be interrupted upon seeing `$(B.f ...)` and wait on full compilation of `B.f`. Arguably this would require some kind of coroutine mechanism that might not be so simple to implement across process boundaries. (But perhaps this would not be too complicated to implement after all!) 

Regardless, I would like to see an argument for explicit annotations vs. inference in the proposal!

I thought that another reason is assurance for the user that module A is *only* compiled for the host and module C is *only* compiled for the target. But alas, I don't think that is possible to state, because it depends on how modules A and C *are imported* (perhaps by quote or by splice themselves), not on whether the modules import by splice or by quote.

So yeah, perhaps simply *inferring* quote and splice imports and persisting this information in the interface file would be a viable alternative to this proposal, and would perhaps fix https://gitlab.haskell.org/ghc/ghc/-/issues/14095 as well. It would like to discuss this and clarify the issue in the Alternatives section.

---

Furthermore, I was recently made aware of the following staged metaprogramming system for OCaml: https://dl.acm.org/doi/pdf/10.1145/3607851 it would be good to verify that their notion of quote and splice imports does not deviate too much from ours without good reason.

michaelpj:
An observation that came up in the ecosystem workshop. HLS currently has an awful hack around TH. Notably, it has to compile any module dependencies which are needed for splice evaluation to bytecode, but it really wants to compile _only_ those modules. Today, it's not possible to know which modules are _necessary_ for splice evaluation until you actually typecheck the module. So HLS uses `hscCompileCoreExprHook` to do just that: typecheck the module, and then when you hit a splice, jump back out and ensure the build system has built the bytecode for the modules that you use.

However, if we had explicit splice imports we could get this information trivially from looking at the module header.

adamgundry:
I think we can consider this proposal superseded by https://github.com/ghc-proposals/ghc-proposals/pull/682, which is effectively a revised version. Thanks to @mpickering for your work on this and to everyone who commented,

PR #414, opened 2021-03-18
Add GC traceable pointer type

winterland1989:
We propose adding a new primitive type `GAddr#`, which is:

+ Allocated from a threaded local slab allocator, here slab means memory pool classified by size(16, 32, 64...).
+ It's GC traceable.
+ It can be added or substracted with offset, the result is still a `GAddr#`, which is traceable.
+ It can be passed to FFI like a pointer type, such as `char*` or `int16_t*` depend on types.

[Rendered](https://github.com/ZHaskell/ghc-proposals/blob/gaddr/proposals/0000-add-GC-traceable-pointers.md)

### DIFF
--- a/proposals/0000-add-GC-traceable-pointers.md
+++ b/proposals/0000-add-GC-traceable-pointers.md
@@ -62,8 +62,8 @@ data PrimArray a = PrimArray GAddr# Int#
 We don't need another slice type: the slicing operation can be implemented like:
 
 ```
-take :: Int -> PrimArray Word8 -> PrimArray Word8
-take (PrimArray gaddr len) (I# n) | n <=# len = PrimArray (gaddr `plusGAddr#` n) (len -# n)
+drop :: Int -> PrimArray Word8 -> PrimArray Word8
+drop (PrimArray gaddr len) (I# n) | n <=# len = PrimArray (gaddr `plusGAddr#` n) (len -# n)
                                   | otherwise = error "..."
 ```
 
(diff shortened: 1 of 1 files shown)


mpickering:
Perhaps this would get more discussion as a GHC ticket. I don't think ghc developers tend to look at these proposals too often.

bgamari:
@winterland1989 and I had a call to discuss this proposal. The conclusion was that while the proposal here does offer some intriguing benefits, it will likely be possible to get most of the benefits in a simpler way by building on what we have. Specifically,

 * allocating pinned bytearrays via the slab-style allocator used by the existing non-moving collector, allowing tracing of interior pointers and eliminating a major source of fragmentation
 * a primop for determining whether a `ByteArray#` is pinned, allowing users to opportunistically pass an array to an FFI call without copying
 * introduce a type,
   ```haskell
   data OffsetByteArray# :: TYPE (BoxedRep Unlifted) = OffsetByteArray# ByteArray# Int#
   ```
   and code generation logic to allow passing of `OffsetByteArray#` arguments via `UnliftedFFITypes`. This will eliminate the need for  C wrappers that some FFI applications 
 * somewhat orthgonal to the present proposal: introduce support for [off-heap `ByteArray#`s](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/2561), allowing `ByteArray#`s to be create from arbitrary off-heap buffers (e.g. enabling use of `mmap`'d `ByteArray`s).

aspiwack:
Reading @bgamari 's remarks, I'm wondering what the future of this proposal is. Should it be closed?

nomeata:
Probably. I’m closing, but please do shout out if you want to revive it!

PR #415, opened 2021-03-18
OPAQUE pragma (under review)

christiaanb:
Being able to mark a (global) binder _opaque_, thus:
```
{-# OPAQUE f #-}
```

What does OPAQUE mean?

* Don't inline it
* Type-class specialisation is a no-op, as is SpecConstr
* Don't perform any w/w on it, not even cast w/w in the early stage simplifier.
* Let any caller know that while the annotated function might be head strict, it will not unbox its arguments -- e.g. if `f :: Int -> Int -> blah` was strict, callers would normally expect its wrapper to inline and unbox, but it won't (For NOINLINE GHC carefully does w/w for this reason! See `Note [Worker-wrapper for NOINLINE functions]` in `GHC.Core.Opt.WorkWrap`).

To quote https://github.com/ghc-proposals/ghc-proposals/pull/415#issuecomment-802650289:

> The ultimate goal is that every call of `f` generates, well, a call of `f`, not of some name-mangled variant.

[Rendered](https://github.com/christiaanb/ghc-proposals/blob/master/proposals/0000-opaque-pragma.md)

### DIFF
--- a/proposals/0000-opaque-pragma.md
+++ b/proposals/0000-opaque-pragma.md
@@ -1,0 +1,120 @@
+---
+author: Christiaan Baaij
+date-accepted: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/0>).
+**After creating the pull request, edit this file again, update the number in
+the link, and delete this bold sentence.**
+
+# OPAQUE pragma
+
+To paraphrase https://gitlab.haskell.org/ghc/ghc/-/issues/19553#note_339736:
+
+Being able to mark a (global) binder _opaque_, thus:
+```
+{-# OPAQUE f #-}
+```
+
+What does OPAQUE mean?
+
+* Don't inline it
+* Don't expose its strictness -- e.g. if `f :: Int -> Int -> blah` was strict, callers would expect its wrapper to inline and unbox, but it won't (For NOINLINE GHC carefully does w/w for this reason! See `Note [Worker-wrapper for NOINLINE functions]` in `GHC.Core.Opt.WorkWrap`)
+* Don't perform any w/w on it, not even cast w/w in the early stage simplifier.
+
+## Motivation
+
+The base/prelude library of the Clash Haskell-to-Hardware compiler has multiple functions where Clash should _not_ translate the definitions of those functions to hardware, but use a specific-to-that-function hard-coded translation instead. An example of such a function would be multiplication for Clash' arbitrary-sized signed number type (See https://github.com/clash-lang/clash-compiler/blob/ab6dd31dc61fc96bb058e57dc117db6f5626d3a9/clash-prelude/src/Clash/Sized/Internal/Signed.hs):
+```haskell
+module Clash.Sized.Internal.Signed where
+
+newtype Signed (n :: Nat) = S { unsafeToInteger :: Integer}
+
+{-# NOINLINE times# #-}
+times# :: Signed m -> Signed n -> Signed (m + n)
+times# (S a) (S b) = S (a * b)
+```
+How does Clash know it should pick a hard-coded translation for that `times#` function? It reads in files such as https://github.com/clash-lang/clash-compiler/blob/ab6dd31dc61fc96bb058e57dc117db6f5626d3a9/clash-lib/prims/vhdl/Clash_Sized_Internal_Signed.primitives which contains entries such as:
+```json
+{ "BlackBox" :
+    { "name"      : "Clash.Sized.Internal.Signed.times#"
+    , "template"  : "~IF~AND[~SIZE[~TYP[0]],~SIZE[~TYP[1]]]~THEN~ARG[0] * ~ARG[1]~ELSEsigned'(~SIZE[~TYPO]-1 downto 0 => '0')~FI"
+    }
+}
+```
+which tells the Clash compiler that whenever it sees an application of `Clash.Sized.Internal.Signed.times#` it should use the string-template that's given by the `template` entry, as opposed to looking up the definition of `Clash.Sized.Internal.Signed.times#` and translating that.
+
+Clash uses the GHC API as a frontend, and also uses regular GHC+Cabal to install libraries. As such, Clash critically depends on GHC preserving applications of `Clash.Sized.Internal.Signed.times#`, and not inlining such functions.
+
+One might assume that the `NOINLINE` pragma is sufficient, but as https://gitlab.haskell.org/ghc/ghc/-/issues/19553 shows, it is not. That's because worker/wrapper transformations translate definitions such as the above `times#` to:
+```haskell
+times# :: Signed m -> Signed n -> Signed (m + n)
+times# = times#1 |> coercion
+
+times#1 :: Signed m -> Signed n -> Integer
+times#1 (S a) (S b) = a * b
+```
+where the `NOINLINE` pragma on `times#` is dropped. GHC actually does this for good reasons: https://gitlab.haskell.org/ghc/ghc/-/commit/6d49d5be904c0c01788fa7aae1b112d5b4dfaf1c
+
+All of this is framed in the Clash use case, but I can image that other GHC (API) users needs/wants to recognize applications of certain functions, and thus doesn't want those functions inlined (nor transformed by w/w).
+
+## Proposed Change Specification
+
+* Add a new `OPAQUE` pragma
+* Binders with an `OPAQUE` pragma are not inlined, just like `NOINLINE`.
+* Binders with an `OPAQUE` pragma do not have their strictness exposed, unlike `NOINLINE`
+* Binders with an `OPAQUE` pragma are not w/w transformed, not even the cast w/w that happens in the early stage simplifier (that normally even happens for `-O0` runs).
+* Unlike `NOINLINE`, `OPAQUE` does _not_ form a two word pragma together with `SPECIALIZE` _nor_ does it form a two word pragma with `CONLIKE`.
+* `OPAQUE[n]` respects the activation phase `n` in the same way as `NOINLINE[n]` with regards to when the annotated binder is allowed to be inlined.
+
+## Examples
+
+To indicate that `times#` really should not be inlined:
+```haskell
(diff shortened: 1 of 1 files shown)


treeowl:
This seems much nicer in general than the approach I've occasionally had to use of defining such a binding in a separate module compiled with `-O0`. A trick I've used would look something like this:

```haskell
{-# OPAQUE hidden1 #-}
{-# OPAQUE hidden2 #-}
hidden1, hidden2 :: a
(hidden1, hidden2) = undefined
```

These can be used in `RULES` with nonlinear patterns to detect when a function ignores its argument and treat it specially in that case:

```haskell
somefun :: (a -> b) -> ...
somefun f = ...
{-# NOINLINE [1] somefun #-}
somefun' f _ _ = somefun f
{-# INLINE [1] somefun' #-}

{-# RULES
"somefun" forall f. somefun f = somefun' f (f hidden1) (f hidden2)
"somefun2" forall f same. somefun' f same same = specialfun same
 #-}
```

simonpj:
The "discussed at this pull request" link is broken

simonpj:
> OPAQUE[n] respects the activation phase n in the same way as NOINLINE[n] with regards to when the annotated binder is allowed to be inlined.

Whoa!  Do we really need this?   Do you really want it inlined after phase n?   Are you expecting the strictness info to be exposed after phase n?   (Actually strictness analysis occurs after phase 0, the last phase, so that would mean strictness is never hidden; and worker wrapper would transform it.)

I think it'd be simpler not to allow a phase annotation -- at least until we have a live, motivated use-case

Other points:
* Are you allowed an INLINE or NOINLINE pragma too?  No -- it's mutually exclusive I think.
* Are you allowed a SPECIALISE pragma?   No -- I think not.  (Unlike NOINLINE.)
* We should specify that the type-class specialisation pass is a no-op, as is SpecConstr.  Yes?  Otherwise again you may get name-mangled variants.

There is no harm in letting head-strictness through:
```
{-# OPAQUE f #-} 
f :: Int -> Int
f x = x + 1
```
Then we can say that `f` is *strict* in `x`, but it doesn't *unbox* it.  Strictness allows more use of call-by value, but doesn't do w/w.

The ultimate goal is that every call of `f` generates, well, a call of `f`, not of some name-mangled variant.

christiaanb:
> > OPAQUE[n] respects the activation phase n in the same way as NOINLINE[n] with regards to when the annotated binder is allowed to be inlined.
> 
> Whoa! Do we really need this? Do you really want it inlined after phase n? Are you expecting the strictness info to be exposed after phase n? (Actually strictness analysis occurs after phase 0, the last phase, so that would mean strictness is never hidden; and worker wrapper would transform it.)
> 
> I think it'd be simpler not to allow a phase annotation -- at least until we have a live, motivated use-case

Correct, it is definitely not needed for the Clash use case. I guess I was trying to overly future-proof it for a future that may never be. I'll simplify the proposal and drop the possibility for phase annotations.

> Other points:
> 
> Are you allowed an INLINE or NOINLINE pragma too?  No -- it's mutually exclusive I think.

Ah, originally my thoughts were that adding NOINLINE on top of OPAQUE would be superfluous, as OPAQUE is somewhat of an NOINLINE++ (or NOINLINE-- depending on how you interpret it). But it is probably better to explicitly error and have the user explicitly decide between one or the other. Same for INLINE and INLINEABLE pragma: GHC should error out when a binder has both an OPAGUE and either an INLINE or INLINEABLE pragma.

Also, since the whole idea is not to inline OPAQUE binders, GHC should also error out when someone annotates a binder with both a OPAQUE and a CONLIKE pragma.

> Are you allowed a SPECIALISE pragma?   No -- I think not.  (Unlike NOINLINE.)

Indeed, for the same "no name mangling" reasons, a binder should not have both a OPAQUE and SPECIALISE pragma.

> We should specify that the type-class specialisation pass is a no-op, as is SpecConstr.  Yes?  
> Otherwise again you may get name-mangled variants.

Correct

> There is no harm in letting head-strictness through:
> 
> ```
> {-# OPAQUE f #-} 
> f :: Int -> Int
> f x = x + 1
> ```
> 
> Then we can say that `f` is _strict_ in `x`, but it doesn't _unbox_ it. Strictness allows more use of call-by value, but doesn't do w/w.

Correct

> The ultimate goal is that every call of `f` generates, well, a call of `f`, not of some name-mangled variant.

I'll update the proposal accordingly.

simonpj:
You can submit to the committee when you are ready

christiaanb:
@nomeata The proposal is ready to be submitted to the committee. The points raised in the discussion have been covered:
* Unlike NOINLINE, OPAQUE does not have an activation phase (it did in the original proposal)
* OPAQUE is now _explicitly_ mutually exclusive with INLINE, NOINLINE, INLINABLE, CONLIKE, and SPECIALISE
* We still let callers know arguments of OPAQUE binders are strict, but we'll also inform the caller that the arguments won't be unboxed (the initial proposal suggested stripping strictness information as well)

i-am-tom:
Hi! Thanks for your patience; I'm currently moving house, so it has taken me a while to catch up with everything!
This feels like a very reasonable proposal with a clear motivation, and I'll be recommending acceptance to the committee. /cc @nomeata

sgraf812:
Wow, this proposal completely slipped my attention.

My only question: What does this proposal provide that we couldn't get with a combination of [`GHC.Exts.lazy`](https://hackage.haskell.org/package/base-4.15.0.0/docs/GHC-Exts.html#v:lazy) and [`GHC.Exts.noinline`](https://hackage.haskell.org/package/base-4.15.0.0/docs/GHC-Exts.html#v:noinline)? I think it gives more guarantees wrt. Cast W/W and specialisation. But I expected both magic Ids to be mentioned.

christiaanb:
@sgraf812 Compiling the following with `-O`:
```haskell
{-# LANGUAGE DataKinds #-}
module Test where

import GHC.TypeNats
import GHC.Exts

newtype Signed (n :: Nat) = S { unsafeToInteger :: Integer}

{-# NOINLINE times #-}
times :: Signed m -> Signed n -> Signed (m + n)
times (S a) (S b) = S (a * b)

f = noinline times
```
results in the following simplifier output:
```
-- RHS size: {terms: 7, types: 6, coercions: 4, joins: 0/0}
Test.f1 [InlPrag=NOINLINE]
  :: forall {m :: Nat} {n :: Nat}. Signed m -> Signed n -> Integer
[GblId, Arity=2, Str=<1L><SL>, Unf=OtherCon []]
Test.f1
  = \ (@(m :: Nat))
      (@(n :: Nat))
      (ds :: Signed m)
      (ds1 :: Signed n) ->
      GHC.Num.Integer.integerMul
        (ds `cast` (Test.N:Signed[0] <m>_P :: Signed m ~R# Integer))
        (ds1 `cast` (Test.N:Signed[0] <n>_P :: Signed n ~R# Integer))

-- RHS size: {terms: 1, types: 0, coercions: 17, joins: 0/0}
times [InlPrag=[final]]
  :: forall (m :: Nat) (n :: Nat).
     Signed m -> Signed n -> Signed (m + n)
[GblId,
 Arity=2,
 Str=<1L><SL>,
 Unf=Unf{Src=<vanilla>, TopLvl=True, Value=True, ConLike=True,
         Guidance=ALWAYS_IF(arity=0,unsat_ok=True,boring_ok=True)}]
times
  = Test.f1
    `cast` (forall (m :: <Nat>_N) (n :: <Nat>_N).
            <Signed m>_R
            %<'Many>_N ->_R <Signed n>_R
            %<'Many>_N ->_R Sym (Test.N:Signed[0] <m + n>_P)
            :: (forall {m :: Nat} {n :: Nat}. Signed m -> Signed n -> Integer)
               ~R# (forall {m :: Nat} {n :: Nat}.
                    Signed m -> Signed n -> Signed (m + n)))

-- RHS size: {terms: 4, types: 14, coercions: 17, joins: 0/0}
f :: forall {m :: Nat} {n :: Nat}.
     Signed m -> Signed n -> Signed (m + n)
[GblId,
 Unf=Unf{Src=<vanilla>, TopLvl=True, Value=False, ConLike=False,
         WorkFree=False, Expandable=False, Guidance=IF_ARGS [] 20 0}]
f = \ (@(m :: Nat)) (@(n :: Nat)) ->
      noinline
        @(forall (m1 :: Nat) (n1 :: Nat).
          Signed m1 -> Signed n1 -> Signed (m1 + n1))
        (Test.f1
         `cast` (forall (m1 :: <Nat>_N) (n1 :: <Nat>_N).
                 <Signed m1>_R
                 %<'Many>_N ->_R <Signed n1>_R
                 %<'Many>_N ->_R Sym (Test.N:Signed[0] <m1 + n1>_P)
                 :: (forall {m1 :: Nat} {n1 :: Nat}.
                     Signed m1 -> Signed n1 -> Integer)
                    ~R# (forall {m1 :: Nat} {n1 :: Nat}.
                         Signed m1 -> Signed n1 -> Signed (m1 + n1))))
        @m
        @n
```
As you can see, the Core for `f` contains a call to a "name-mangled" variant of `times`, even though I annotated the binder `times` with a `NOINLINE` pragma, and annotated the call-site with a `GHC.Exts.noinline`.

simonpj:
The proposal (accepted though it is) has some markdown errors in the latter part.

Also can the sections be numbered please?

christiaanb:
@simonpj do you mean that RestructuredText https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0415-opaque-pragma.rst isn't rendered properly on github? That does seem to be the case yeah. The original markdown did render correctly though: https://github.com/christiaanb/ghc-proposals/blob/master/proposals/0000-opaque-pragma.md

I'll make a PR

nomeata:
Sorry if me converting this using `pandoc` failed, and thanks for cleaning it up

barci2:
I believe the proposal has been implemented in https://gitlab.haskell.org/ghc/ghc/-/issues/19553. Should it receive an "Implemented" label and have the tracking issue set in the proposal document?

nomeata:
Yes! I've set the label (on the phone right now). You can update the metadata with a PR. You can also include the GHC version number there if already known. Thanks for your help!

PR #418, opened 2021-04-18
Add rows to syntax table for punned updates

shayne-fletcher:
Bug-fix: As pointed out by @adamgundry  in the ["Refactor RecordUpd"](https://gitlab.haskell.org/ghc/ghc/-/issues/19463#note_347206) ticket, punned field update syntax is missing in the syntax table of [Section 2.2.1](https://github.com/shayne-fletcher/ghc-proposals/blob/master/proposals/0282-record-dot-syntax.rst#211-syntax). This fixes it.

### DIFF
--- a/proposals/0282-record-dot-syntax.rst
+++ b/proposals/0282-record-dot-syntax.rst
@@ -112,6 +112,8 @@ Expression              Equivalent
 ``e{lbl₁.lbl₂ = val}``  ``e{lbl₁ = (e.lbl₁){lbl₂ = val}}``
 ``e.lbl₁{lbl₂ = val}``  ``(e.lbl₁){lbl₂ = val}``
 ``e{lbl₁ = val₁}.val₂`` ``(e{lbl₁ = val₁}).val₂``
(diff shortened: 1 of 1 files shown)


simonpj:
Can you change 2.1.1 to make the "Note" into bullets, thus

Note
* **Updating nested fields**.   `e{lbl = val}` is the syntax of a standard H98 record update. It is the nested form introduced by this proposal that is new : `e{lbl1.lbl2 = val}`. However, in the event `OverloadedRecordUpdate` is in effect, note that `e{lbl = val}` desugars to `setField @"lbl" e val`
* **Punning**.  With `NamedFieldPuns`, the form `e { x, y }`  means `e { x=x, y=y }`.  With `OverloadedRecordUpdate` this behaviour is extended to nested updates: `e { a.b.c, x.y }`  means `e { a.b.c=c, x.y=y }`.  Note the variable that is referred to implicitly (here `c` and `y`) is the last chunk of the field to update.    So `c` is the last chunk of `a.b.c`, and `y` is the last chunk of `x.y`.

shayne-fletcher:
> Can you change 2.1.1 to make the "Note" into bullets, thus
[...]

Done in https://github.com/ghc-proposals/ghc-proposals/pull/419.

PR #420, opened 2021-04-21
no-implicit-binds: Adjust defaulting of `-XPatternSignatureBinds`

Ericson2314:
Today `-XPatternSignatureBinds` only matters when `-XScopeTypeVariables` is enabled, but proposal #238 would re-introduce `-XPatternSignatures`, with which `-XPatternSignatureBinds` would also be visible.

We cannot change the meaning of `-XScopeTypeVariables`, but I would rather `-XPatternSignatures` alone not cause any variables to be bound, as this is the more conservative, forwards compatible option, and one that I believe would best position `-XPatternSignatures` to become part of a future Haskell report even if the other extensions are hotly debated.

Finally, this means #238 can be a bit simpler and just assume the existence of `-XPatternSignatureBinds` without having to change any aspect of it.

[Rendered](https://github.com/Ericson2314/ghc-proposals/blob/no-implicit-forall/proposals/0285-no-implicit-binds.rst)

### DIFF
--- a/proposals/0285-no-implicit-binds.rst
+++ b/proposals/0285-no-implicit-binds.rst
@@ -55,7 +55,8 @@ It is on by default for backwards compatibility.
 When using ``-XNoImplicitForAll``, all variables in regular signatures, instances, and data declarations must be explicitly bound.
 
 Create ``-XPatternSignatureBinds`` to allow the implicit binding of free variables in pattern signatures.
-It is also on by default for backwards compatibility.
+It is implied by ``-XScopeTypeVariables`` by default for backwards compatibility.
+(Without ``-XScopeTypeVariables`` it currently has no effect, so we are free to have it off by default.)
 When using ``-XNoPatternSignatureBinds``, all variables in pattern signatures must be explicitly bound.
 
 Examples


Ericson2314:
This is a very small non-observable amendment to an accepted proposal, so I'd like to immediately submit it to the committee. CC @nomeata.

nomeata:
Do you mean `PatternSignatures` when you write `PatternSynonysm` in the PR description?

@simonmar, as the shepherd of #285, do you think that this qualifies as a “mostly editorial change” that you can just wave through?

nomeata:
/remind me to ping Simon in one week

Ericson2314:
Sorry, yes I do!

Ericson2314:
Did something happen with the reminder?

nomeata:
I’ll ask @simonmar by mail

Ericson2314:
#238 is now up for review, so I would like to submit this too (not sure what happened with emailing @simonmar).

nomeata:
I'm a bit lost in the various pattern related proposals, and since I'm traveling and only have a small phone screen, not well equipped to gain a good overview. @serras, maybe you are in a better spot? Maybe this proposal should be shepherded together with one of the others?

Ericson2314:
If @goldfirere is OK with it, I would propose combining all 3 of these PRs modifying existing proposals together. They are all trying to harmonize the design around this stuff so I think they are stronger together.

simonpj:
> If @goldfirere is OK with it, I would propose combining all 3 of these PRs modifying existing proposals together. 

That would help me, for one.  My head is spinning, so having a few rather than a blizzard would be v useful

goldfirere:
What is "all 3" -- that is, which 3 in specific? I'm lost, too! (I get that #238 is one of the others, but that leaves a mysterious third.)

Ericson2314:
The third is further afield, but https://github.com/ghc-proposals/ghc-proposals/pull/291. 

The connection is that we want `f @x = ...` and `f (Cons @x) = ...` to work analogously, e.g. not needing `-XScopeTypeVariables` where #291 required it for binding to be cautious, but #238 makes a point of quasi-deprecating `-XScopeTypeVariables` and offering ways to bind without it.

(I prefer the #238 approach to what I wrote in #291.)

nomeata:

Ericson2314:
@nomeata Sounds good. When I get back from vacation I will write the "meta proposal" I alluded to in https://github.com/ghc-proposals/ghc-proposals/pull/283#issuecomment-890392494 to address those procedural issues head on. Thanks to your email writing the motivation will be easier!

nomeata:
Thanks. Maybe collaborate with @goldfirere on that?

nomeata:

PR #425, opened 2021-06-07
Invisible binders in type declarations

int-index:
The [proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0425-decl-invis-binders.rst) has been accepted; the following discussion is mostly of historic interest.

This is my rewrite of #326 + #386.

We propose to allow invisible type variable binders (i.e. `@k`) in type declarations. Here are a few examples:

In a class declaration:
```haskell
class C @k (a :: k) where ...
--     ^^^^
```

In a data type declaration:
```haskell
data D @k @j (a :: k) (b :: j) = ...
--    ^^^^^^^
```

In a type family declaration:
```haskell
type family F @p @q (a :: p) (b :: q) where ...
--           ^^^^^^^
```

We then propose to use these binders to simplify scoping rules and arity inference.

[Rendered](https://github.com/serokell/ghc-proposals/blob/decl-invis-binders/proposals/0000-decl-invis-binders.rst)

### DIFF
--- a/proposals/0000-decl-invis-binders.rst
+++ b/proposals/0000-decl-invis-binders.rst
@@ -429,6 +429,8 @@ their headers::
 The occurrences of ``@`` must be *prefix*, as defined by
 `#229 <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0229-whitespace-bang-patterns.rst>`_.
 
+Guarded behind a new flag, ``-XTypeAbstractions``.
+
 Scope
 ~~~~~
 
@@ -622,8 +624,92 @@ inference and scoping rules.
 Costs and Drawbacks
 -------------------
 
-The change to arity inference is breaking, but most users are
-likely to be unaffected.
+The proposed changes break existing code in a few ways.
+
+Breakage 1: Scoping in Type Synonyms
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The first source of breakage is a change in the scoping of type synonyms. Consider::
+
+  type T1 = 'Nothing :: Maybe a
+
+This is no longer accepted, as ``a`` is not bound on the LHS. Instead, the user must write::
+
+  type T1 @a = 'Nothing :: Maybe a
+
+See "Argument 4: Quantification in Type Synonyms" for the motivation.
+
+There's no backwards-compatible way to rewrite this example, so we introduce
+the ``-Wimplicit-rhs-quantification`` warning, wait for three releases (in
+accordance with 3-Release-Policy), and only then make the change.
+
+Breakage 2: Arity Inference
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The second source of breakage is the change to arity inference. Consider::
+
+  type F :: Type -> forall k. Maybe k
+  type family F x
+
+This definition is currently assigned the arity of 2, but with the proposed
+changes will be assigned arity of 1. To keep the arity of 2, the user must rewrite it thus::
+
+  type F :: Type -> forall k. Maybe k
+  type family F x @k
+
+See "Argument 5: Arity Inference" for the motivation.
+
+We assume that this is an obscure situation and the change will go
+unnoticed by most users, because in order to encounter it, one must use
+``StandaloneKindSignatures`` – a relatively recent addition in itself.
+
+No migration strategy is provided.
+
+Breakage 3: Scoping in Instances
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The third source of breakage is the new requirement that variables mentioned on
+the RHS must also occur on the LHS. Consider::
+
+  type family F a :: k
+  type instance F Int = Any :: j -> j
+
+This is no longer accepted, as ``j`` is not mentioned on the LHS. The user must
+rewrite it as follows::
+
+  type family F a :: k
+  type instance F @(j -> j) Int = Any :: j -> j
+
+See "Addendum: Quantification in Type Family Instances" for the motivation.
+
+The ``@``-binders in type instances are already legal, so there's no need for a
+migration strategy: the fix is backwards-compatible.
+
+Breakage 4: Instantiation
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The fourth source of breakage is that instantiation of type/data family
+instances is fully determined by the left hand side, without looking at the
+right hand side. Consider::
+
+  type family F a :: k
+  type instance F Int = Char
+  type instance F Int = Maybe
+
+This is no longer accepted, as the LHS ``F Int`` does not fully determine if
+the instance matches, as evidenced by the presence of two instances with
+identical LHSs.
+
+The user must rewrite this code as follows::
+
+  type family F a :: k
+  type instance F @Type         Int = Char
+  type instance F @(Type->Type) Int = Maybe
+
+See "Addendum: Instantiation of Type Family Instances" for the motivation.
+
+The ``@``-binders in type instances are already legal, so there's no need for a
+migration strategy: the fix is backwards-compatible.
 
 Alternatives
 ------------
--- a/proposals/0000-decl-invis-binders.rst
+++ b/proposals/0000-decl-invis-binders.rst
@@ -1,0 +1,621 @@
+Invisible binders in type declarations
+======================================
+
+.. author:: Vladislav Zavialov
+.. co-author:: John Ericson, Simon Peyton Jones
+.. date-accepted::
+.. ticket-url::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/425>`_.
+.. contents::
+
+We propose to allow invisible type variable binders (i.e. ``@k``) in type
+declarations. Here are a few examples::
+
+  class C @k (a :: k) where ...
+         ^^^^
+
+  data D @k @j (a :: k) (b :: j) = ...
+        ^^^^^^^
+
+  type family F @p @q (a :: p) (b :: q) where ...
+               ^^^^^^^
+
+We then propose to use these binders to simplify scoping rules and arity
+inference.
+
+Relation to `#155 <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0155-type-lambda.rst>`_
+---------------------------------------------------------------------------------------------------------------
+
+This proposal is essentially the type-level counterpart to
+`#155 "Binding type variables in lambda-expressions" <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0155-type-lambda.rst>`_.
+While we do not have lambdas at the type level, we do have definition left-hand
+sides. Compare::
+
+  const :: a -> b -> a
+  const @c x _ = (x :: c)        -- from #155
+
+  type Const :: a -> b -> a
+  type Const @c x _ = (x :: c)   -- this proposal
+
+The subject of both proposals is the ``@c`` to the left of ``=``. The
+difference is that #155 is about functions, while this proposal is about
+``type``, ``data``, ``newtype``, ``class``, ``type family``, and ``data
+family`` declarations.
+
+#155 has been accepted, but it is not implemented at the time of writing.
+Nevertheless, for the purposes of presentation, **the text of this proposal
+assumes that term-level ``@t``-binders are already a part of the language.**
+
+Motivation
+----------
+
+The goal of this proposal is to tidy up the language and to simplify certain
+aspects of it related to name resolution, implicit quantification, and arity
+inference.
+
+We shall consider a number of various ways to arrive at the idea of
+``@k``-binders. While each individual argument may seem weak, together they
+form a compelling reason for the addition.
+
+Argument 1: Symmetry between Terms and Types
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Let us consider a kind-polymorphic class ``C``, such as the following::
+
+  class C (a :: k)
+
+In this declaration, ``a`` is a binding site for a type variable, whereas ``k``
+is a usage site. You can easily verify this claim by trying to duplicate the
+binder::
+
+  class Cx (a :: k) (a :: k)    -- rejected, "a" is bound twice
+  class Cy (a :: k) (b :: k)    -- ok
+
+Where is ``k`` bound, then? That is where implicit quantification comes into
+play. Compare with the following term-level definition::
+
+  f (x :: a) (y :: a) = undefined
+
+Here, ``x`` and ``y`` are variable binders, while ``a`` is a usage of an
+implicitly quantified type variable. However, in terms there is a way to bind
+``a`` explicitly::
+
+  f @a (x :: a) (y :: a) = undefined
+   ^^^^
+   explicit binder
+
+This proposal introduces the type-level equivalent of that feature::
+
+  class C @k (a :: k)
+         ^^^^
+         explicit binder
+
+As a consequence, the new syntax makes the language more uniform and
+consistent.
+
+Argument 2: Symmetry between ``class`` and ``instance``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Let us once again consider the kind-polymorphic class introduced earlier::
+
+  class C (a :: k)
+
+The kind of ``C`` is ``forall k. k -> Constraint``. With this definition, all
+of the following instances are permitted::
+
+  instance C Int
+  instance C Maybe
+  instance C False
+
+Note how ``Int``, ``Maybe``, and ``False`` are all of different kinds. That is
+possible because we can instantiate ``k`` differently in each instance. If we
+also enable ``TypeApplications``, we can make this clear::
+
+  instance C @Type           Int
+  instance C @(Type -> Type) Maybe
+  instance C @Bool           False
+
+This choice of syntax makes it apparent that ``C`` is in fact a multi-parameter
+type class of kind ``forall k. k -> Constraint``. The first parameter of ``C``
+is ``k``, the second parameter is ``a``, and both can be instantiated.
+
+One might expect that if it is possible to instantiate ``k`` by writing ``C
+@Type``, ``C @(Type -> Type)``, or ``C @Bool``, then the syntax to abstract
+over ``k`` would be ``C @k``. Unfortunately, that is not the case::
+
+  class C @k (a :: k)
+         ^^^^
+         rejected
+
+This proposal lifts this restriction.
+
+Argument 3: Deduplication with ``StandaloneKindSignatures``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Now let us consider a slightly more complicated example::
+
+  data P a b = MkP
+
+  class C (a :: i -> i -> i) where
+    p :: P a i
+
+The point of interest here is the ``i`` parameter of ``C``. You will notice two things about it::
+
+1. ``i`` is used as part of a lengthy, syntactically large type (``i -> i -> i``)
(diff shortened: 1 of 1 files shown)


Ericson2314:
Looking good so far! Per https://github.com/ghc-proposals/ghc-proposals/pull/386#issuecomment-814823108 I think you want the scoping and instantiation rules for data families too?

jvanbruegge:
Does this proposal touch in any way the way implicit pattern matching in type families work? I mean if I pattern match on types with different kinds, GHC implicitly pattern matches on their kinds first which can lead to surprising behavior (partiality when one does not expect it naively). With this proposal, it is possible to bind invisible arguments and explicitly pattern match on them, so would it be possible to disable that implicit pattern match completely and present an error to the user?

int-index:
@jvanbruegge The proposal disallows matching on implicit kind parameters on the RHS, so the following will become illegal:

```
type family F a :: k

type instance F Int = Char
type instance F Int = Maybe
```

You would need to rewrite it as:

```
type instance F @Type         Int = Char
type instance F @(Type->Type) Int = Maybe
```

However, if matching happens on the LHS, it is still fair game:

```
type family G (a :: k) :: Natural

type instance G Char = 0
type instance G Maybe = 1
type instance G True = 2
```

For clarity, we could rewrite it as:

```
type instance G @Type Char = 0
type instance G @(Type -> Type) Maybe = 1
type instance G @Bool True = 2
```

But in this case it is not required.

jvanbruegge:
Ok, cool. Yeay, I meant the first case

proposals/0000-decl-invis-binders.rst
```diff
@@ -0,0 +1,440 @@
...
+
+Support for ``@k``-binders in type declarations will have the following
+positive effects:
+
+* Term- and type-level syntax become more consistent with one another
+* ``class`` and ``instance`` declarations can be more symmetric
+* Standalone kind signatures will no longer lead to duplication
+* Implicit quantification rules in type synonyms can be simplified
+* Arity inference becomes much more straightforward
+
```
goldfirere, line 288: I think this proposal also strictly increases expressivity:

```hs
type family F a
type C :: forall a. F a -> Constraint
class C x where   -- no way to put a kind signature on x
  -- no way to bring a into scope here
```

Check me on the above, but I think there may be programs unwritable without this extension.

Relatedly, more simply, but more theoretically: This proposal means that every variable can have an explicit binding site. I'm pretty sure this is the last frontier (though I could be wrong).

proposals/0000-decl-invis-binders.rst
```diff
@@ -0,0 +1,440 @@
...
+quantifiers are included in the arity of a type synonym or type family.
+
+Primary Change: ``@k``-binders
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+1. Relax the syntactic check of ``data``, ``newtype``, ``type``, ``class``,
+   ``type family``, and ``data family`` declarations to allow ``@k``-binders in
+   their headers::
+
+     tv_bndr ::=
```
goldfirere, line 357: This could be misconstrued. What is `v`? What is `k`? Is `@_` one lexeme or two? I recommend expanding this out a bit, perhaps to:

```
tv_bndr ::=
        | tyvar
        | '(' tyvar '::' kind ')'
        | '@' tyvar
        | '@' '(' tyvar '::' kind ')'
        | '@' '_'
```

With your helpful comments, and with a note saying that all `@`-signs are prefix occurrences.

proposals/0000-decl-invis-binders.rst
```diff
@@ -0,0 +1,440 @@
...
+              | v           -- variable
+              | (v :: k)    -- variable with kind annotation
+       (NEW)  | @v          -- invisible variable
+       (NEW)  | @(v :: k)   -- invisible variable with kind annotation
```
goldfirere, line 361: What about *inferred* binders? I think this proposal needs a treatment of them. It just might be appropriate to have `@{k}` syntax for binding them, even though we don't do that in terms. The alternative would be to adopt the `@(..)` syntax we see in #238. Not quite sure what's best, but there should be something about inferred variables, I think.

proposals/0000-decl-invis-binders.rst
```diff
@@ -0,0 +1,440 @@
...
+
+Costs and Drawbacks
+-------------------
+
+TBD.
+
+Alternatives
+------------
+
+* We could also introduce ``@{k}``-binders for ``forall {x}.``-quantifiers, but
```
goldfirere, line 427: Aha. I'm glad you mention this. So does that mean that it is impossible to bring an inferred variable into scope, other than implicitly via a kind annotation? Perhaps, and perhaps that's OK. Regardless, it might be nice to mention the possibility of inferred variables when talking about zipping.

goldfirere (commented):
I like where this is going. Many thanks for writing this up!

proposals/0000-decl-invis-binders.rst
Ericson2314: @goldfirere I assume `(..)` from your #238 would apply here too? Then we consistently avoid that problem everywhere.

Ericson2314: Oh I se you already mentioned above in https://github.com/ghc-proposals/ghc-proposals/pull/425#discussion_r647683861

proposals/0000-decl-invis-binders.rst
```diff
@@ -0,0 +1,440 @@
...
+issue entirely::
+
+  type C :: forall i. (i -> i -> i) -> Constraint
+  class C @i a where
+    p :: P a i
+
+By writing ``@i``, we bind the type variable, making it scope over the class
+body; at the same time, we avoid repeating ``i -> i -> i``.
+
+Argument 4: Quantification in Type Synonyms
```
simonpj, line 186: argument 4a

proposals/0000-decl-invis-binders.rst
```diff
@@ -0,0 +1,440 @@
...
+  type T1 @a = 'Nothing :: Maybe a
+  type T2 @a = 'Just ('Nothing :: Maybe a)
+
+This way all the variables occurring on the RHS are bound on the LHS.
+We exploit the new syntax to allow a nice, simple, uniform scoping rule.
+To cite the User's Guide, "The reason for this exception [the strange, ad-hoc
+rule about top-level kind annotations] is that there may be no other way to
+bind k".
+
+The same rule applies to type family instances::
```
simonpj, line 215: Argument 4b: quantification in type family instances

proposals/0000-decl-invis-binders.rst
```diff
@@ -0,0 +1,440 @@
...
+   type family F a :: k
+
+   type instance F Int = Any :: j -> j
+
+(where ``Any :: forall p. p``). This type instance is currently legal, but
+under this proposal ``j`` would not be in scope. You would have to write::
+
+   type instance F @(j->j) Int = Any :: j -> j
+
+In type instances, this form is already allowed today.
```
simonpj, line 226: What does this sentence mean?  This *is* a `type instance` declaration!  Maybe you mean class `instance`?   Give an example

proposals/0000-decl-invis-binders.rst
```diff
@@ -0,0 +1,440 @@
...
+  -- arity = 2
+  type F :: Type -> forall k. Maybe k
+  type family F x
+
+  -- arity = 1
+  type G :: Type -> forall k. Maybe k
+  type family G x :: forall k. Maybe k
+
+With ``@``-binders we can do the opposite. We propose that by default, arity
+inference would include as few forall-bound variables as possible, to allow
```
simonpj, line 263: More than that, the arity would no longer be strangely influenced by the return kind signature.  Instead, you can simply count the binders to the left the `::` (if any).

Sigh: not quite.  In
```
F :: forall k. k -> Type
type family F a
```
F has arity 2, not 1.

proposals/0000-decl-invis-binders.rst
simonpj: > This proposal means that every variable can have an explicit binding site. 

Yes!  Say this loud and clear!

simonpj:
I have added some comments

I am happy to see that you have included
* In type family and data family instances, the instantiation is fully determined by the left hand side, without looking at the right hand side.

but you have zero text motivating, or even specifying, this point.  There is plenty to plunder in https://github.com/ghc-proposals/ghc-proposals/blob/wip/spj-instance-scoping/proposals/0000-instance-scoping.rst

proposals/0000-decl-invis-binders.rst
int-index: Yes, it is a type instance declaration, and it is already allowed

int-index:
This is ready for another round of review.

nomeata:
The title still says “Draft”. Oversight?

proposals/0000-decl-invis-binders.rst
```diff
@@ -0,0 +1,621 @@
...
+4. During arity inference, do not include trailing ``forall x.`` quantifiers in
+   the arity.
+
+Secondary Change: Scoping Rules
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+5. In type synonym declarations, require that every variable mentioned on the
+   RHS must be bound on the LHS.
+
+6. In type family and data family instances, require that every variable
```
goldfirere, line 545: This part is completely separable from the rest of the proposal, I think. Is this stated somewhere?

proposals/0000-decl-invis-binders.rst
```diff
@@ -0,0 +1,621 @@
...
+   mentioned on the RHS must also occur on the LHS.
+
+Secondary Change: Instantiation Rules
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+7. In type family and data family instances, the instantiation is fully
```
goldfirere, line 551: This, similarly, is completely separable form the rest of the proposal -- even the other point about instances. They are in sympathy with each other (and, indeed, the rest of the proposal), but there is no dependency between the rest of this proposal and these two points about instances. Right?

proposals/0000-decl-invis-binders.rst
```diff
@@ -0,0 +1,621 @@
...
+kind ``Relation a b``. So the declaration is rejected, which would have
+saved Jakob some time.
+
+Effect and Interactions
+-----------------------
+
+* Changes (1), (2), and (3) provide the programmer with a more principled way
+  of brining type variables into scope in certain corner cases.
+
+* Changes (4), (5), and (6) simplify arity inference and scoping rules, but they are
```
goldfirere, line 593: (6) does not need a migration strategy, because the fix is backward-compatible.

I'm not very bothered about (4), which is rather exotic.

For (5), I would say to issue a warning for a few releases before pulling the trigger. I don't think that would be hard to implement.

goldfirere (commented):
I'm quite happy with this! Thanks.

proposals/0000-decl-invis-binders.rst
int-index: It’s implied by putting it into a section titled “Secondary Change”.

goldfirere: Yes, but I would expect a "secondary change" to depend on the primary one. This one has no relationship, really, to the rest of the proposal. I'm all for it -- just got surprised at first at seeing a change to instance syntax in a proposal about type syntax.

int-index: True, this proposal bundles a few unrelated changes together.

int-index:
@nomeata I'd like to submit this proposal to the committee.

int-index:
@goldfirere In case this slipped your attention, do you plan to make a recommendation?

int-index:
What’s the status here? @simonpj, what do you think of the updated version, have I incorporated your suggestions successfully?

simonpj:
> What’s the status here? @simonpj, what do you think of the updated version, have I incorporated your suggestions successfully?

Yes, you have, thank you.

@goldfirere you are the shepherd. Are we done?

Ericson2314:
You all are probably thinking the same, but If we are going to try some data family promotion stuff for tuples, I think it is quite important we get this fixed first :). The cleanup from this should provide some spare "complexity budget" for the rest, I think.

goldfirere:
I've taken another pass through the specification. I have an easy question and a harder one.

Easy: Is this feature gate-kept behind a flag? The specification does not say so (but I think it should. Perhaps `-XTypeAbstractions` of #448.)

Harder: In light of recent commentary elsewhere about the pain of breaking changes, I've grown more concerned about the breakage induced by this proposal. How can we mitigate this pain?

I spot four independent sources of breakage here (in contrast to the claim in the [Costs and Drawbacks](https://github.com/serokell/ghc-proposals/blob/decl-invis-binders/proposals/0000-decl-invis-binders.rst#costs-and-drawbacks) section):

1. The change in scoping in type synonyms (requiring all variables on the RHS to be introduced in the LHS).
2. The elimination of arity inference.
3. The requirement that variables on the RHS of type/data instances are introduced on the LHS.
4. The requirement that the instantiation for type/data instances be determined by information on the LHS.

All four of these are independent of each other. All but (2) are essentially optional. (1) has a three-release warning; during these three releases, the fix can be applied (or not), but it is not backward compatible to today's GHC. (2) will break code right away; the fix is not backward compatible, but it is hard to imagine how to support a migration period -- and we believe the scenario addressed in (2) to be very rare. (3) and (4) will break code right away, but the fix is backward compatible for several GHC versions.

Should we pull the trigger? I think we should, and I'm happy enough with the plan as written. But I do think it would be appropriate to seek a little more guidance from outside our little bubble. To this end, could I request clear examples in the proposal demonstrating the breakage and fixes? Then, I am happy to seek out further input from people who are breakage-averse. In addition -- and regardless of what those people say -- I would like to plan, at least, to have very informative error messages in these cases, with (ideally) the suggested fix. Happily, I think doing so would be easy enough.

What do others think? Am I being overly cautious in requesting this extra step?

simonpj:
> The elimination of arity inference.

Arity inference is not eliminated is it?  It's just simplified.  But yes it'll give sightly different answers in obscure situations.

I think it would be helpful for the proposal to give concrete examples of all four forms of breakage.   (Indeed I think this might be helpful in any proposal that breaks backward compatibility.)   That will help to anchor our judgement about how rare it will be.

FWIW my sense is that it'll be pretty rare, and it's OK to pull the trigger.   Re (4) for example, who writes stuff like this?
```
type family F :: forall k. k where
  F = Int
  F = Monad
```
(I'm not even sure I got it right.) 

Your point about whether it's possible to fix the code in a way that is also accepted by previous versions of GHC is a good one: again, on each point it would help to have that laid out, along with the fix if there is one.   A backward-incompatibility that can be fixed without requiring compiler-version nonsense is less harmful than one that truly requires #ifdefs.

goldfirere:
I'm going to put this into needs-revision mode. My understanding from off-ticket communication is that @int-index has agreed to put these examples in the proposal.

It is worth saying that early voting in the committee was positive, and so I expect that this will indeed be accepted.

simonpj:
@int-index are you planning to make the small revisions asked for here?  It'd be great to land this proposal!

int-index:
> are you planning to make the small revisions asked for here?

On it.

int-index:
>  Is this feature gate-kept behind a flag?

Good point, I amended the proposal to require `-XTypeAbstractions` for the new feature.

> I spot four independent sources of breakage here

@goldfirere Thanks, I extended "Costs and Drawbacks" to explicitly describe those sources of breakage with examples.

int-index:
> merge with your branch

Done. I couldn’t find any option in the repository settings that might have been preventing a PR, so if you have more changes for me to pull, just let me know.

> I'd like to offer in my Discourse post to slow down and re-evaluate if head.hackage shows that these breakages will be burdensome. Do you think that's fair?

I’m not sure how we could test this without first implementing the proposal. What evaluation process exactly do you have in mind?

goldfirere:
> What evaluation process exactly do you have in mind?

A post-proposal-acceptance process. Or, if you prefer, I'd argue for conditional acceptance here, depending on what we learn during implementation. (Truth be told, I consider *all* proposals conditionally accepted on similar grounds. It is always possible to learn some new devastating fact during implementation.) Maybe this plan of mine makes it all more concrete:

1. Blast to Discourse (and maybe a few other places) seeking feedback on the breakage.
2. Discuss.
3. If the discussion suggests that doing so is sensible, advocate to the committee for acceptance.
4. Accept the proposal.
5. Implement (this is your part).
6. Assess breakage in practice on head.hackage, along at least two different axes: how many different places break, and how hard each breakage is to diagnose & fix. That is, even if each fix is trivial, if every package breaks, that's bad. Even if there are very few breakages, if the breakage requires the whole package to be rewritten, that's bad. (To be clear, I expect neither of these.)
7. Report on this breakage back out to the community, and discuss briefly.
8. Depending on the result of that discussion, merge.

We're operating here expecting breakage both to be relatively rare and easy to diagnose & fix. If these assumptions are correct, then these last steps will go swimmingly. If we're wrong, then it will be good to re-examine things.

My comment "Do you think that's fair?" is about setting expectations: I wouldn't want to have this plan in mind only to have you and your colleagues very upset when I try to take Step 7.

Ericson2314:
I would say we should be always able to merge the feature, but it might need a deprecation cycle. For example, simplified subsumption could have gotten a deprecation cycle rather than being cancelled outright as many people probably wished after being unable to upgrade to 9.0.

I too need to write a discorse post, but about how the sort of modularity work @hsyl20 and I are in to directly relates to our ability to do proper depreciation cycles without drowning in hard-to-test conditional code.

int-index:
> Maybe this plan of mine makes it all more concrete

@goldfirere That’s a fine plan, and thanks for taking the initiative here.

goldfirere:
I have created posts on [Discourse](https://discourse.haskell.org/t/feedback-requested-about-breakage-type-variable-binders-in-datatypes/4241), [Reddit](https://www.reddit.com/r/haskell/comments/tjm0ga/ghc_proposal_breakage_should_we_proceed/), and [Twitter](https://twitter.com/RaeHaskell/status/1506015143150645248). Will review in a few days.

/remind me to review responses on Thursday.

AntC2:
> @goldfirere I have created posts on ... 

Hehe Reddit thought it was spam. (What on earth triggered their algorithm?)

tomjaguarpaw:
Reddit's zealous spam filter snares me about 10% of the time I post a story. I messaged the mods to request they restore it.

goldfirere:
All seems in order here. In particular, the community outreach did not produce any red flags. I am (re-)recommending acceptance to the committee. Thanks!

goldfirere:
/remind me to accept this proposal (pending committee agreement) on April 4.

JakobBruenker:
> the community outreach did not produce any red flags

@goldfirere Note that the [reddit post](https://www.reddit.com/r/haskell/comments/tjm0ga/ghc_proposal_breakage_should_we_proceed/) is still marked as `[removed]`
(Which I imagine means the moderators haven't read @tomjaguarpaw's messsage yet)

goldfirere:
@JakobBruenker That's true (and I was aware of it). But given the quietness both on Discord and Twitter, I decided not to investigate. If you want to get Reddit unblocked, great! There's no reason that we can't revisit the committee discussion in light of new information from Reddit. But I'm afraid I'm not going to take another crack at Reddit on this front.

goldfirere:
@phadej points out (on [Reddit](https://www.reddit.com/r/haskell/comments/tjm0ga/ghc_proposal_breakage_should_we_proceed/), but the thread is mostly about something else) that this proposal violates the [Explicit Variables Proposal](https://github.com/goldfirere/ghc-proposals/blob/type-variables/principles.rst#212explicit-variable-principle-evp), included in #448, because there is no way to bring inferred variables into scope without unification. (Actually, the same is true within #448.) This suggests that this proposal might benefit from the [`@(..)` syntax](https://github.com/goldfirere/ghc-proposals/blob/type-variables/proposals/0448-type-variable-scoping.rst#58alternatives) proposed with #448.

I do not suggest we change this proposal at this point, just that there is an improvement possible in a later proposal.

Ericson2314:
That sounds good. If #448 pulls some of this under it's umbrella, that's fine, but there is enough separate stuff on arity not befitting #448 that I agree that's not a good reason to delay this any further.

PR #430, opened 2021-07-22
MonadFailDesugaring in v8.10.5 docs, note removal

ysangkok:
This flag isn't available in GHC master, so don't provide a broken link to Gitlab. Instead this link now goes to the current latest release which included the extension.


nomeata:
Thanks!

Issue #391, opened 2021-01-08
Explore ways to weaken or abandon the Instance Consistency Condition for fundeps


simonpj:
In [Functional dependencies in GHC](https://gitlab.haskell.org/ghc/ghc/-/wikis/Functional-dependencies-in-GHC) we identify two two major functional-dependency-related conditions on instance declarations
* The Coverage Condition
* The Instance Consistency Condition

Proposal #374 explores ways of weakening the Coverage Condition.

This ticket explores way to weaken (or even abandon) the Instance Consistency Condition. It is not yet a proposal ... a ticket will do for now.

---------------------
## Abandoning the instance consistency condition

Here is one concrete possibility, triggered by Examples 2, 3, and 4 on [the wiki page](https://gitlab.haskell.org/ghc/ghc/-/wikis/Functional-dependencies-in-GHC):

* Abandon the LICC altogether. It is too weak (Examples 2,3) and too strong (Example 4).
* Instead, when considering improvement of a Wanted constraint against the global instances, do the following:
  * Find all the instances that can possibly match the Wanted (where "can possibly match" is "unify with").
  * Among these instances see if there is one "best" instance, which is more specific than (is a substitution instance of) all the otheres
  * If there is just one such, add the fundeps from that instance.
  * "Add fundeps from instance" means (precisely as now): for each fundep, if the LHS tys match, then generate an equality with the instantiated RHS tys.

These rules are very like the [rules for overlapping instances](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/instances.html?highlight=overlapping%20instances#overlapping-instances).

#9210 is a very relevant ticket, with interesting discussion.

### Examples of how it works

Now consider Example 2:
* Only one instance unifies with `[W] C Bool [alpha] beta`
* So we take fudeps from that constraint alone, giving `beta ~ [alpha]`
* Now we can solve `[W] C Bool [alpha] [alpha]`
* Yielding `f :: a -> [a]` as desired

Now Example 3.  `[W] C alpha beta (gamma,delta)` unifies with both instances, so we get no fundeps at all.
If `alpha` gets unified with, say, `Int` by some other constraint, then it'll unify with just one instance and we can use the fundeps.

Example 4.  If we have `[W] TypeEq Int Int r`, that unifies with only one instance, so we'll get a fundep `r ~ True` as desired. Similarly if the first two arguments are apart.  All is good.

In the OP from #9210 we have
```
[...]

PR #433, opened 2021-08-06
Unsatisfiable

adamgundry:
This proposal introduces a new built-in constraint form, `Unsatisfiable`, as a mechanism for introducing custom type error messages. It is similar to the existing `TypeError` built-in type family, but it is more predictable when errors are reported.

[Rendered](https://github.com/adamgundry/ghc-proposals/blob/unsatisfiable/proposals/0000-unsatisfiable.rst)

### DIFF
--- a/proposals/0000-unsatisfiable.rst
+++ b/proposals/0000-unsatisfiable.rst
@@ -324,6 +324,11 @@ This provides a more flexible alternative to ``WARNING`` and ``DEPRECATED``
 pragmas, because warnings may be displayed only when definitions are used at
 particular types, and the messages may be constructed dynamically.
 
+(It would be possible to add a ``Deprecated`` constraint, similar to
+``Warning``, specifically for use with deprecation warnings, and controlled by a
+different flag.  However separating out deprecation warnings specifically does
+not seem strongly motivated.)
+
 
 
 Examples
@@ -591,3 +596,4 @@ Unresolved Questions
 --------------------
 
 Should the optional ``Warning`` feature be included?
+Should there be a separate ``Deprecated`` constraint as well?
--- a/proposals/0000-unsatisfiable.rst
+++ b/proposals/0000-unsatisfiable.rst
@@ -1,0 +1,469 @@
+Unsatisfiable
+=============
+
+.. author:: Adam Gundry
+.. date-accepted::
+.. ticket-url::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/433>`_.
+.. contents::
+
+This proposal introduces a new built-in constraint form, ``Unsatisfiable``, as a
+mechanism for introducing custom type error messages. It is similar to the
+existing ``TypeError`` built-in type family, but it is more predictable when
+errors are reported.
+
+
+Motivation
+----------
+
+GHC allows users to define custom type error messages using the ``TypeError``
+type family.  This is defined (in ``GHC.TypeLits``) as follows::
+
+  type family TypeError (a :: ErrorMessage) :: b where
+
+``TypeError`` is poly-kinded, so it can be instantiated wherever needed:
+typically either on the right-hand side of a type family definition to mark an
+invalid case, or in a context (instantiating it at kind ``Constraint``) to
+indicate that a constraint should not be solved.
+
+When GHC is reporting custom type errors it renders the user-supplied
+``ErrorMessage`` argument in place of the usual type error message. The details
+of the ``ErrorMessage`` type are not crucial here: it allows concatenation of
+literal strings and pretty-printed types.  For further details of the current
+design, see the `GHC user's guide on TypeError
+<https://downloads.haskell.org/ghc/9.0.1/docs/html/users_guide/exts/type_errors.html>`_.
+
+However, type families do not currently enjoy a clearly-defined operational
+semantics. It is not well specified under what circumstances GHC will reduce
+type families during the constraint solving stage of type inference, so it is
+not obvious when a ``TypeError`` will be encountered and reported as an error.
+In practice, GHC solves constraints normally, then when reporting unsolved
+constraints, it searches the constraint type for occurrences of ``TypeError``
+and reports any that it finds.  Unfortunately, this can lead to surprising
+behaviour, as the following subsections describe.
+
+
+
+Example: ``TypeError`` can report errors too lazily
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following is a tempting but problematic use of ``TypeError``, because it
+does not report an error when one would be expected::
+
+  type family NotCharBad x :: Type where
+    NotCharBad Char = TypeError (Text "Oops")
+    NotCharBad x    = x
+
+  f :: x -> [NotCharBad x] -> [NotCharBad x]
+  f _ r = r
+
+  oops :: Bool
+  oops = null (f 'x' [])
+
+The intention is that ``f`` should not be callable with ``x ~ Char``.  However
+type-checking ``oops`` does not depend on how ``NotCharBad Char`` reduces, so
+the ``TypeError`` is not triggered.
+
+See `#13775 <https://gitlab.haskell.org/ghc/ghc/-/issues/13775>`_ for another
+example of a type error being reported too lazily.
+
+Where it is important that certain inputs are prohibited, a better pattern is to
+use type families that return constraints::
+
+  type family NotCharGood x :: Constraint where
+    NotCharGood Char = TypeError (Text "Oops")
+    NotCharGood x    = ()
+
+  g :: NotCharGood x => x -> [x] -> [x]
+  g _ r = r
+
+Now type-checking ``null (g 'x' [])`` will need to solve the ``NotCharGood
+Char`` constraint, and hence will trigger the ``TypeError``.
+
+
+Example: ``TypeError`` can report errors too eagerly
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In the previous subsection, we saw a program that did not report a custom type
+error when one would be expected, and that this can be avoided by using
+``TypeError`` at kind ``Constraint``.  However, the opposite problem exists as
+well: sometimes custom type errors are reported when they are not expected.
+
+As a simple example, it is not possible to directly write a function which will
+trigger a type error when used, because this is rejected at the definition
+site::
+
+  foo :: TypeError (Text "Dont' call foo") => Int
+  foo = 0
+
+As a more substantial example, consider the following (where ``If`` is the type
+family from ``Data.Type.Bool``)::
+
+  -- Using a type family rather than a type synonym to work around #20181
+  type family ExpectTrue x :: Constraint where
+    ExpectTrue x = If x (() :: Constraint) (TypeError (Text "Input was False!"))
+
+  h :: ExpectTrue x => proxy x -> ()
+  h _ = ()
+
+  -- This works:
+  eg1 _ = h (Proxy @True)
+
+  -- This yields an error (expected):
+  eg2 _ = h (Proxy @False)
+
+  -- This yields an error (unexpected):
+  eg3 p = h p
+
+Here ``h`` is supposed to be called with ``x ~ True``.  As one would expect,
+explicitly instantiating ``x ~ True`` allows it to be called, while
+instantiating ``x ~ False`` triggers the error.  However, ``eg3`` demonstrates
+that the error is triggered also when ``x`` is unconstrained.  This is because
+GHC infers that::
+
+  eg3 :: If x () (TypeError (Text "Input was False!")) => proxy x -> ()
+
+Even though ``x`` is not yet known, and the ``If`` type family application may
+reduce to eliminate the ``TypeError`` depending on the value of ``x``, GHC still
+reports the occurrence of the ``TypeError``.
+
+See `#14771 <https://gitlab.haskell.org/ghc/ghc/-/issues/14771>`_ and `#16906
+<https://gitlab.haskell.org/ghc/ghc/-/issues/16906>`_ for more examples of
+errors being reported too eagerly.
+
+
+
+
+Example: ``TypeError`` can report too many errors
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+One of the uses of ``TypeError`` is to point out that a class instance is
+impossible, with a domain-specific error message. For example::
+
+  class ReflexiveEq a where
+      reflexiveEq :: a -> a -> Bool
+
+  instance TypeError (Text "Equality is not reflexive on Double") => ReflexiveEq Double where
+      reflexiveEq _ _ = False
+
+This displays the custom type error when trying to use the instance::
+
+  ghci> reflexiveEq 0 (0 :: Double)
+
+  <interactive>:28:1: error:
+      • Equality is not reflexive on Double
+      • In the expression: reflexiveEq 0 (0 :: Double)
+
+However, when superclasses are involved, the instance may need to carry
+additional constraints, which can result in too many errors being reported.  For
+example::
+
+  class Eq a => ReflexiveEq' a where
+      reflexiveEq' :: a -> a -> Bool
+      reflexiveEq' = (==)
+
+  instance (TypeError (Text "Can't compare functions with reflexiveEq"), Eq (a -> b)) => ReflexiveEq' (a -> b)
+
+  instance TypeError (Text "Can't compare functions with (==)") => Eq (a -> b) where
+      _ == _ = False
+
+Notice that the instance for ``ReflexiveEq'`` had to repeat the ``Eq (a -> b)``
+constraint in the instance context, because otherwise a type error is reported
+at the instance definition site.  But this approach means both errors are
+reported at the use site::
+
+  ghci> reflexiveEq' id not
+
+  <interactive>:36:1: error:
+      • Can't compare functions with (==)
+      • In the expression: reflexiveEq' id not
+        In an equation for ‘it’: it = reflexiveEq' id not
+
+  <interactive>:36:1: error:
+      • Can't compare functions with reflexiveEq
+      • In the expression: reflexiveEq' id not
+        In an equation for ‘it’: it = reflexiveEq' id not
+
+Ideally, when using a custom type error to report a domain-specific error
+message for a particular instance, it would not be necessary to add other
+constraints to the instance context.
+
+
+``TypeError`` and ``-fdefer-type-errors``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A further issue arises with ``-fdefer-type-errors``.  When this flag is enabled,
+GHC turns type errors into compile-time warnings, and replaces the erroneous
+code with runtime exceptions.  However, when using ``TypeError`` to prevent a
+definition from being called, there is nothing to guarantee that the error will
+be triggered at runtime.  For example::
+
+  ghci> let v = reflexiveEq 0 (0 :: Double)
+
+  <interactive>:32:9: warning: [-Wdeferred-type-errors]
+      • Equality is not reflexive on Double
+      • In the expression: reflexiveEq 0 (0 :: Double)
+        In an equation for ‘v’: v = reflexiveEq 0 (0 :: Double)
+  ghci> v
+  False
+
+We would expect evaluation of ``v`` to throw an error, but it does not!  Instead
+it uses the "redundant" method definition from the ``ReflexiveEq Double``
+instance to yield a value.  See `#16249
+<https://gitlab.haskell.org/ghc/ghc/-/issues/16249>`_ and `#18310
+<https://gitlab.haskell.org/ghc/ghc/-/issues/18310>`_ for discussion of this
+issue.
+
+Ideally, it would be possible for the definition of the ``reflexiveEq`` method
+in the ``ReflexiveEq Double`` instance to explicitly make use of the fact that
+the context is inconsistent, rather than giving a bogus definition that is not
+expected to be called.  Then under ``-fdefer-type-errors``, evaluation of the
+bogus evidence for the constraint would be forced, resulting in the expected
+runtime error.
+
+
+Proposed Change Specification
+-----------------------------
+
+The ``GHC.TypeError`` module (the planned new home for ``TypeError`` per `!6066
+<https://gitlab.haskell.org/ghc/ghc/-/merge_requests/6066>`_) is extended with
+the following definition::
+
+  class Unsatisfiable (e :: ErrorMessage) where
+    unsatisfiable :: a
+
+``Unsatisfiable`` constraints have the following properties:
+
+* During constraint solving, the solver treats ``Unsatisfiable`` Wanted
+  constraints like any other class with no instances.  At the end of constraint
+  solving, if a Wanted constraint of the form ``Unsatisfiable e`` remains
+  unsolved:
+
+  - Under ``-fno-defer-type-errors`` (the default), the type ``e`` is normalised
+    and rendered as a custom type error message.
+
+  - Under ``-fdefer-type-errors``, a warning is emitted instead of an error, and
+    the constraint is solved by producing a dictionary such that
+    ``unsatisfiable`` throws an appropriate runtime exception.
+
+* If an ``Unsatisfiable`` Given constraint is present during constraint solving,
+  the constraint solver will automatically solve all Wanted constraints. The
+  evidence for each constraint consists of an application of the
+  ``unsatisfiable`` class method at the appropriate type.
+
+* An ``Unsatisfiable`` constraint is never automatically generalised.
+
+* If an ``Unsatisfiable`` Given constraint is present during pattern-match coverage
+  checking, the match is trivially regarded as total.  See discussion on `#20180
+  <https://gitlab.haskell.org/ghc/ghc/-/issues/20180>`_.
+
+* If a class instance has an ``Unsatisfiable`` Given constraint in the context,
+  it bypasses the functional dependency check.
+
+* GHC will report an error if a user attempts to define an instance for
+  ``Unsatisfiable``.
+
+The rendering of ``ErrorMessage`` values works just as for ``TypeError``. But
+unlike ``TypeError``, mere presence of ``Unsatisfiable`` somewhere within a
+constraint type does not trigger an error.
+
+
+Examples
+--------
+
+For more substantial examples, see the `unsatisfiable package by Oleg Grenrus
+<https://hackage.haskell.org/package/unsatisfiable>`_.  This package implements
+a type-checker plugin that roughly corresponds to the design in this proposal.
+
+#. The following definitions are accepted (assuming ``DataKinds``,
+   ``FlexibleContexts`` and ``TypeApplications`` are enabled)::
+
+     type Msg = Text "Cannot call 'uncallable'."
+
+     uncallable :: Unsatisfiable Msg => ()
+     uncallable = unsatisfiable @Msg
+
+     uncallable' :: Unsatisfiable Msg => ()
+     uncallable' = uncallable
+
+   The definition of ``uncallable`` is accepted because ``Unsatisfiable`` appears
+   as a Given, not a Wanted.  In ``uncallable'``, the Wanted arising from the
+   occurrence of ``uncallable`` is solved using the Given in the context.
+
+#. The following definition results in a custom type error message (as the
+   ``Unsatisfiable Msg`` constraint is reported rather than being generalised
+   over)::
+
+     rejected = uncallable  -- error
+
+#. The following definition is accepted::
+
+     unusual :: Unsatisfiable Msg => Char
+     unusual = 42  -- no error
+
+   Here the presence of ``Unsatisfiable Msg`` in the context means that the ``Num
+   Char`` constraint arising from the body of ``unusual`` is discharged
+   automatically.
+
+#. The following definition is accepted::
+
+     k :: Unsatisfiable (Text "No") => ()
+     k = uncallable  -- no error
+
+   The Given ``Unsatisfiable (Text "No")`` solves the Wanted ``Unsatisfiable
+   Msg``, even though the messages are different. This shows that error messages
+   can be changed.
+
+#. Recall the following example from the Motivation, adapted for
+   ``Unsatisfiable``::
+
+     type ExpectTrue x = If x (() :: Constraint) (Unsatisfiable (Text "Input was False!"))
+
+     h :: ExpectTrue x => proxy x -> ()
+     h _ = ()
+
+     eg1 _ = h (Proxy @True)   -- no error
+
+     eg2 _ = h (Proxy @False)  -- error
+
+     eg3 p = h p               -- no error
+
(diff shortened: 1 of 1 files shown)


adamgundry:
@akrmn good question! 

> This made me wonder, what will happen if `-fdefer-type-errors` is used to call the instance for functions, `reflexiveEq id (id :: Int -> Int)`? Will that somehow use the default implementation of `reflexiveEq = (==)`, even though there's no `Eq` instance for functions? 

Indeed, the `ReflexiveEq (a -> b)` instance doesn't provide an implementation for `reflexiveEq` so it uses the default implementation. That requires an `Eq` dictionary, which is solved automatically (since there is an `Unsatisfiable` Given in the instance context) by inserting a call to `unsatisfiable`.

> Perhaps the specification should say something to the effect of "instances declared with `Unsatisfiable e` constraints have all their methods automatically defined as `unsatisfiable`"?

I did wonder about this. The Alternatives section says:

> The presence of an `Unsatisfiable` constraint in an instance context could allow the class methods to be omitted, with the compiler automatically inserting calls to `unsatisfiable`. This seems like unnecessary magic, however, because the user can write explicit calls to `unsatisfiable` instead, and in the presence of `-fdefer-type-errors` it can matter whether a class method is defined or calls `unsatisfiable`.

I'm not opposed to changing this and adding the automatic insertion of missing methods, if people think that is better. It's a little unclear to me what to do if there is a default method: do we use the default method, or do we override it? Doing either seems a bit arbitrary, although the difference is observable only with `-fdefer-type-errors`.

isovector:
This is a fantastic proposal. One issue I ran into when working on `polysemy` was that sometimes you only want to emit a custom error if the are no other errors to present --- that is to say, whether to report the error in report1 or report2. I had some WIP code towards this end in [!1739](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/1739).

rowanG077:
I like this proposal and see it for the first time. One thing I was hoping for is to also start a story on type warnings. I sometimes come across use cases where a custom warning is something I want. I tried my hand at implementing it in https://gitlab.haskell.org/ghc/ghc/-/merge_requests/2361 but I got stuck. I'd never worked with GHC so I didn't see a way forward.

Ericson2314:
> Given in the instance context) by inserting a call to `unsatisfiable`.

Do you plan on having the dictionary itself being diverging. or something else?

I would assume `unsatisfiable` is operationally pretty boring, i.e.
```
unsatisfiable :: UnsatisfiableDict e -> a
unsatisfiable = \case
```
but with the kinds fixed up. (And we should have absurd dictionary patterns or anyways so this can be written in surface Haskell.)

I would think the error message is done by the bottom value of `UnsatisfiableDict` itself, and an overlapping instance like
```
instance Unsatisfiable e => c
```
That could in fact be allowed syntax too, if we recognize `Unsatisfiable e` along with things like `Int ~ Bool` as false even with the open-world assumption.

adamgundry:
Thanks for the feedback!

@isovector 

> sometimes you only want to emit a custom error if the are no other errors to present --- that is to say, whether to report the error in report1 or report2.

Yes, I see this could be helpful, but I'm not sure how best to specify it. One can definitely end up in situations where introducing a custom message would be useful to explain e.g. an ambiguity or overlapping instances error. That's hard to express using something like `Unsatisfiable` or `TypeError` however. My current feeling is that a more promising approach would be some kind of "error reporting plugins" that could inspect the errors generated by GHC and rewrite them.

@rowanG077 
> One thing I was hoping for is to also start a story on type warnings.

I think this wouldn't be too difficult to add along the lines of `Unsatisfiable`: add something like
```hs
class Warning (e :: ErrorMessage)
```
such that a Wanted `Warning e` is solved automatically by the constraint solver, emitting the warning in the process (i.e. as if there was a global `instance Warning e`). Unlike `TypeWarning` this wouldn't be usable at arbitrary kinds, but OTOH it is clear at which points it gets triggered.

@Ericson2314 

> > Given in the instance context) by inserting a call to `unsatisfiable`.
> 
> Do you plan on having the dictionary itself being diverging. or something else?
>
> I would assume `unsatisfiable` is operationally pretty boring, i.e.
> 
> ```haskell
> unsatisfiable :: UnsatisfiableDict e -> a
> unsatisfiable = \case
> ```
> 
> but with the kinds fixed up. (And we should have absurd dictionary patterns or anyways so this can be written in surface Haskell.)
> 
> I would think the error message is done by the bottom value of `UnsatisfiableDict` itself, and an overlapping instance like
> 
> ```haskell
> instance Unsatisfiable e => c
> ```
> 
> That could in fact be allowed syntax too, if we recognize `Unsatisfiable e` along with things like `Int ~ Bool` as false even with the open-world assumption.

The dictionary type is precisely what one would expect from the class definition, i.e. the equivalent of `newtype UnsatisfiableDict e = UnsatisfiableDict (forall a . a)`, and `unsatisfiable` is just the usual class method selector function. The only time GHC will construct such a dictionary is under `-fdefer-type-errors`, where it inserts a runtime exception. Thus I don't think we need any more changes, and I'd prefer to keep this proposal as simple as possible. But perhaps I'm missing your point?

proposals/0000-unsatisfiable.rst
```diff
@@ -0,0 +1,469 @@
...
+
+Proposed Change Specification
+-----------------------------
+
+The ``GHC.TypeError`` module (the planned new home for ``TypeError`` per `!6066
+<https://gitlab.haskell.org/ghc/ghc/-/merge_requests/6066>`_) is extended with
+the following definition::
+
+  class Unsatisfiable (e :: ErrorMessage) where
+    unsatisfiable :: a
```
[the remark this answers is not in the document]
phadej: Usability. Consider if we had `undefined :: Void` or `error :: String -> Void`.

isovector:
@adamgundry 

> Yes, I see this could be helpful, but I'm not sure how best to specify it. 

My thought was parameterize `TypeError` as `Nat -> ErrorMessage -> k`, with the understanding that `TypeError n err` won't be emitted if there is some other `TypeError n' err'` (which might have come directly from GHC) such that `n' < n`.

Ericson2314:
@adamgundry my point was roughly whether the dictionary is bottom or not. I didn't think of non-bottom with a bottom field — seeing that I now understand the proposal better, but I might still prefer the dictionary being bottom.

adamgundry:
@isovector 
> My thought was parameterize `TypeError` as `Nat -> ErrorMessage -> k`, with the understanding that `TypeError n err` won't be emitted if there is some other `TypeError n' err'` (which might have come directly from GHC) such that `n' < n`.

I'm not hugely keen on this because the assignment of natural numbers to particular errors seems rather ad hoc; it's not clear to me how to choose the values in a principled way, especially once you have multiple libraries potentially reporting custom errors. I'd rather keep this proposal simple and discuss solutions to the problem of error prioritisation independently.

@Ericson2314 

> my point was roughly whether the dictionary is bottom or not. I didn't think of non-bottom with a bottom field — seeing that I now understand the proposal better, but I might still prefer the dictionary being bottom.

Since GHC uses newtypes for the dictionaries single-method classes, I think the dictionary is bottom, no? The field/method is just a coercion, with no extra box, i.e. `Unsatisfiable e` is representationally equal to `forall a . a` in Core.

simonpj:
I too like this proposal a lot.

I think you could give a bit more structure by defining
```
class Unsatisfiable (e :: ErrorMessage) where
  unsatisfiable :: HasCallStack => a
  unsatisfiable = error "Unsatisfiable constraint"
```
* Everyone knows what `error` does, so that helps to explain what will happen if you call `unsatisfiable` directly
* The `HasCallStack` constraint allows us to find out *which* use of `unsatisfible` is responsible.
* In the proposal you say that with `-defer-type-errors` "the constraint is solved by producing a dictionary such that unsatisfiable throws an appropriate runtime exception.".  The `HasCallStack` constraint for these implicit calls would make them much more informative, if I have this right.

There is another class of implicit calls: "If an Unsatisfiable Given constraint is present during constraint solving, the constraint solver will automatically solve all Wanted constraints. The evidence for each constraint consists of an application of the unsatisfiable class method at the appropriate type".   Here we might want to give a error message like "Unsatisfiable Num constraint", rather than just "Unsatisfiable constraint". Maybe `unsatisfiable` needs a String argument?  Not so nice, maybe not so important.

----------------

For the pattern match overlap check points, I'd love to see a compelling example.  The one in the proposal is highly artificial.

adamgundry:
@simonpj 

> I think you could give a bit more structure by defining
> 
> ```
> class Unsatisfiable (e :: ErrorMessage) where
>   unsatisfiable :: HasCallStack => a
>   unsatisfiable = error "Unsatisfiable constraint"
> ```
> 
> * Everyone knows what `error` does, so that helps to explain what will happen if you call `unsatisfiable` directly

I'm a bit confused by this, because it seems you are suggesting giving `unsatisfiable` a default method implementation, but we don't ever give an instance for `Unsatisfiable`, so that default method would never be used. 

Perhaps I should expand on the difference between `error` and `unsatisfiable`:
 * Typically `error` is used when the programmer claims (perhaps incorrectly) that a particular expression should not be evaluated. But it may well be, and if it is, an imprecise exception is thrown blaming the caller of `error` for their negligence.
 * In contrast, at the call site of `unsatisfiable`, the constraints in the context *guarantee* that the expression cannot be evaluated without somebody else having lied about their ability to produce an `Unsatisfiable` dictionary. Thus if the call is ever reached, evaluating the `Unsatisfiable` dictionary will expose the lie. (The liar may be using `-fdefer-type-errors`, calling `error` or supplying an infinite loop, and that is the behaviour we expect to see, not some error arising from `unsatisfiable` itself. The caller of `unsatisfiable` is not responsible for being lied to!)

> * The `HasCallStack` constraint allows us to find out _which_ use of `unsatisfible` is responsible.
> * In the proposal you say that with `-defer-type-errors` "the constraint is solved by producing a dictionary such that unsatisfiable throws an appropriate runtime exception.".  The `HasCallStack` constraint for these implicit calls would make them much more informative, if I have this right.

Given the above, I think `unsatisfiable` itself doesn't need a `HasCallStack` constraint. We typically aren't interested in how the lie was discovered (i.e. where `unsatisfiable` is called) but rather in who told the lie in the first place (i.e. where an `Unsatisfiable` dictionary is constructed). 

It could make sense to have `-fdefer-type-errors` display the call stack as part of the error it inserts. But I don't think that is specific to `Unsatisfiable`, as it would be sensible for any deferred type error? (Indeed, this proposal doesn't make any change to the `-fdefer-type-errors` mechanism; perhaps that could be made clearer.)

> There is another class of implicit calls: "If an Unsatisfiable Given constraint is present during constraint solving, the constraint solver will automatically solve all Wanted constraints. The evidence for each constraint consists of an application of the unsatisfiable class method at the appropriate type". Here we might want to give a error message like "Unsatisfiable Num constraint", rather than just "Unsatisfiable constraint". Maybe `unsatisfiable` needs a String argument? Not so nice, maybe not so important.

Happily, we don't need an extra argument, because it is the original `Unsatisfiable` Given that we want to call out as being a lie, regardless of how many innocent Wanteds were taken in by it.

That said, I think this sentence should be amended to say that the automatic solving of Wanteds from the `Unsatisfiable` Given happens at the end of solving, only for constraints that are otherwise unsolvable. 

> For the pattern match overlap check points, I'd love to see a compelling example. The one in the proposal is highly artificial.

Fair, I'll add some words. The motivation is just the same as the corresponding change to `TypeError` ([#20180](https://gitlab.haskell.org/ghc/ghc/-/issues/20180), [#11503](https://gitlab.haskell.org/ghc/ghc/-/issues/11503)), which has been implemented just recently thanks to @sheaf.

treeowl:
Would this tweak be more or less ergonomic?

```haskell
class Unsatisfiable (e :: ErrorMessage) where
  unsatisfiable' :: a

unsatisfiable :: forall e {rep} (a :: TYPE rep). Unsatisfiable e => a
unsatisfiable = unsatisfiable' @e (##)
```

In the grand scheme of things it doesn't matter, because you can always apply `(##)` (or whatever you like) inline, but it might be a bit more convenient.

adamgundry:
@treeowl good question. I don't have a clear opinion either way.
 * Having `unsatisfiable` be usable directly at unlifted types would be slightly more convenient and would be consistent with e.g. `error`, but it doesn't happen often, and as you say, it's not terribly difficult to work around.
 * This change would mean that an `import GHC.TypeError (Unsatisfiable (..))` would not import `unsatisfiable`, so you'd have to say `import GHC.TypeError (Unsatisfiable, unsatisfiable)` instead. That's mildly annoying, but does mean we have more freedom to tweak the class definition in the future without breaking backwards compatibility...

simonpj:
> I'm a bit confused by this, because it seems you are suggesting giving unsatisfiable a default method implementation, but we don't ever give an instance for Unsatisfiable, so that default method would never be used.

Indeed so, but in the Change Spec you say "the constraint is solved by
producing a dictionary such that unsatisfiable throws an appropriate
runtime exception."  So you *are* implicitly coughing up an instance for
`Unsatisfiable`. Giving it a default method might help the user know what is meant by the Delphic "throw the appropriate exception".

>  I think unsatisfiable itself doesn't need a HasCallStack constraint. We typically aren't interested in how the lie was discovered (i.e. where unsatisfiable is called) but rather in who told the lie in the first place (i.e. where an Unsatisfiable dictionary is constructed).

Let's be concrete.   What happens (with `-fdefer-type-errors`) in this program
```
f :: Unsatisfiable a => a -> a
f x = x+1

main = print (f True)
```
I don't want to get
```
Error: unsatisfiable constraint
```
That is what will happen if you elaborate to
```
unsatDict :: forall a. Unsatisfiable a
unsatDict @a = (error @a emptyStack "Unsatisfiable constraint") |> axUnsat
   -- axUnsat :: a ~ Unsatisfiable a

main = let (d :: Unsatisfiable Bool) = unsatDict @Bool
       in print (f d True)
```

I want to get
```
Error unsatisfiable constraint
  called from main at line 3, char 15.
```
That's what the `HasCallStack` constraint will give you. We'd elaborate to
```
unsatDict :: forall a. HasCallStack => Unsatisfiable a
unsatDict @a s = (error @a s "Unsatisfiable constraint") |> axUnsat
   -- axUnsat :: a ~ Unsatisfiable a

main = let (s :: CallStack)          = pushStack "called from Main" emptyStack
           (d :: Unsatisfiable Bool) = unsatDict @Bool s
       in print (f d True)
```
Maybe the details don't matter. But it *does* matter that we get a decent error message
and I think the proposal should be explicit on this score.

Writing out the elaboration might not be a bad way to do this, although I accept that
it's a bit implementation oriented for a spec.

adamgundry:
@simonpj My "throws an appropriate runtime exception" didn't explain further because it is not proposing any change to what `-fdefer-type-errors` does already. I'll reword the proposal to make it clearer that this sentence is purely descriptive of the status quo.

FWIW this is what we would get currently with `-fdefer-type-errors` on your example:
```hs
ghci> main
*** Exception: Unsatisfiable.hs:20:15: error:
    • No instance for (Unsatisfiable e0) arising from a use of ‘f’
    • In the first argument of ‘print’, namely ‘(f True)’
      In the expression: print (f True)
      In an equation for ‘main’: main = print (f True)
(deferred type error)
```
So this already includes a source location, but one might reasonably argue it could include more of the call stack.

adamgundry:
I realised my previous comment is potentially a bit misleading. Here's a hopefully clearer example:
```hs
f :: Unsatisfiable (Text "Blah blah") => a -> a
f x = x+1

main = print (f True)
```
This should give something like
```
*** Exception: Unsatisfiable.hs:24:15: error:
    • Blah blah
        arising from a use of ‘f’
    • In the first argument of ‘print’, namely ‘(f True)’
      In the expression: print (f True)
      In an equation for ‘main’: main = print (f True)
(deferred type error)
```
So the use of `Unsatisfiable` changes the unsolved constraint error message, and `-fdefer-type-errors` just does its usual thing of packaging up the unsolved constraint error as a runtime exception.

simonpj:
OK, well perhaps add some of this explanation to the proposal, including the error messages you expect in various situations

Ericson2314:
> Since GHC uses newtypes for the dictionaries single-method classes, I think the dictionary is bottom, no? The field/method is just a coercion, with no extra box, i.e. `Unsatisfiable e` is representationally equal to `forall a . a` in Core.

Ok fair point. That is indeed operationally the same as `data UnsatisfiableDict` (empty data decl). Now, if we kept Constraint and Type separate in Core, it might be a bit more legit to use an empty case on the dict than projection of `unsatisfiable` to *ex falso quodlibet* provide other dictionaries, because there would be no going from `Type` to `Constraint` skirting the rules for cannonicity, but this qualm is completely hypothetical today.

goldfirere:
What's the status here? I like the proposal! I would say that this new form should replace any use of `TypeError` at kind `Constraint`.

adamgundry:
The status is that I need to find some time to polish the proposal and submit to the committee. Apologies for letting this linger. I'll try to pick it up again soon.

adamgundry:
I've revised the proposal in the light of all the feedback. Thanks to everyone who has contributed so far! In particular:
 * Various points of the specification have been clarified.
 * The interaction with `-fdefer-type-errors` is now discussed in its own subsection.
 * The definition of ``unsatisfiable`` is now levity-polymorphic.
 * A new optional proposed change has been added, specifying a ``Warning`` class for custom warning messages.  Feedback on whether this is worth including is welcome.  I'm happy for the committee to make a final decision on this option.

I plan to submit this to the commitee soon.

rowanG077:
+1 on the `Warning` class. Thanks a lot!

treeowl:
I also like having a `Warning` class. We should also have a `Deprecated` class for instances that are going to go away.

adamgundry:
> We should also have a `Deprecated` class for instances that are going to go away.

We could, but what in practice is the difference from `Warning`, though? You could use a different warning group to control the messages, I suppose, but is it really worth it?

treeowl:
`Deprecated` seems much stronger to me. `Warning` seems more like something to be used in libraries focused on beginners.

treeowl:
You propose that having an `Unsatisfiable` given erases all wanteds. I don't know much about constraint solving, but that makes me slightly nervous in the presence of things like `UndecidableInstances` and `UndecidableSuperClasses`. Is it guaranteed that the constraint resolution process will never tie a knot that erases the last `Unsatisfiable` wanted?

adamgundry:
> `Deprecated` seems much stronger to me. `Warning` seems more like something to be used in libraries focused on beginners.

Interesting, I would have put them the other way round, with `Deprecated` meaning "this will probably go away in a future library release, but nothing else is wrong" but `Warning` being more general and meaning "this is potentially broken". That's why I see `Deprecated` as being just one kind of `Warning`, and hence subsumed by it.

Is there a practical difference between `Warning (Text "Using function foo at type Bar will be removed in the next release of the frobozz library")` and the corresponding `Deprecated`? The only one I can see is that a user might want to enable one class of warnings but not the other, and that seems like a weak motivation to add a separate class.

> You propose that having an `Unsatisfiable` given erases all wanteds. I don't know much about constraint solving, but that makes me slightly nervous in the presence of things like `UndecidableInstances` and `UndecidableSuperClasses`. Is it guaranteed that the constraint resolution process will never tie a knot that erases the last `Unsatisfiable` wanted?

`UndecidableInstances` and `UndecidableSuperClasses` are primarily about allowing the possibility of compile-time loops, rather than runtime. That said, I don't know if the constraint solver guarantees it will never solve an otherwise unsolvable constraint by constructing a knot that will loop at runtime, and it wouldn't surprise me if that is possible in corner cases. If that does happen, though, I don't think `Unsatisfiable` changes things significantly, because instead of solving `Unsatisfiable` by an accidental loop and using `Unsatisfiable` to solve some other constraint `C`, one could just as well solve `C` by the accidental loop.

treeowl:
There's a bit of a problem with the type of `unsatisfiable`: it requires the user to specify the error message. The way it's done in `trivial-constraints` is to have a boring class

```haskell
class Any => Bottom where
  no :: a
```

and then use that as a superclass of `Unsatisfiable`. That way, when you're under an `Unsatisfiable e` constraint, you can just say `unsatisfiable` without any type arguments.

treeowl:
@adamgundry, that sounds ... devious. I don't really care *how* it works, as long as it does.

nomeata:
/remind @i-am-tom to check progress in two weeks

simonpj:
>  It is similar to the existing TypeError built-in type family, but it is more predictable when errors are reported.

So does this subsume TypeError?   Probably not -- but anytime someone uses `TypeErorr @Constraint` GHC should probably suggest using Unsat instead.   Another warning?  

And maybe explain when TypeErorr is still useful (if it ever is).

proposals/0000-unsatisfiable.rst
```diff
@@ -0,0 +1,599 @@
...
+    message is replaced by the custom message that results from normalising and
+    rendering the type ``e``.
+
+* An ``Unsatisfiable`` constraint is never automatically generalised.
+
+* If an ``Unsatisfiable`` Given constraint is present during pattern-match coverage
+  checking, the match is trivially regarded as total.
+
+* If a class instance has an ``Unsatisfiable`` Given constraint in the context,
+  it bypasses the functional dependency check.
```
aspiwack, line 269: If we are going to include this sort of magic (which, I admit, sounds very nice), I would also recommend that not defining a method or an associated type doesn't trigger a warning when `Unsatisfiable` is given.

proposals/0000-unsatisfiable.rst
```diff
@@ -0,0 +1,599 @@
...
+     rejected = uncallable  -- error
+
+#. The following definition is accepted::
+
+     unusual :: Unsatisfiable Msg => Char
+     unusual = 42  -- no error
+
+   Here the presence of ``Unsatisfiable Msg`` in the context means that the ``Num
+   Char`` constraint arising from the body of ``unusual`` is discharged
+   automatically.
```
aspiwack, line 370: How would this example be desugared? I think something about desugaring is missing from the specification (it would handle this case, as well as `-fdefer-type-errors`)

proposals/0000-unsatisfiable.rst
```diff
@@ -0,0 +1,599 @@
...
+
+#. Assuming the ``Warning`` option is included, the following ``decode``
+   function will emit a warning if it is used at type ``Integer``, but the
+   program will otherwise compile normally::
+
+     type family WarnInteger a where
+       WarnInteger Integer = Warning (Text "Integer may require unbounded memory!")
+       WarnInteger a       = ()
+
+     decode :: (FromJSON a, WarnInteger a) => ByteString -> Maybe a
```
aspiwack, line 473: I like this. There are a bunch of instances like this out there that should be used quite carefully but that are all too easy to use without realising it. We could decorate them with such a warning.

proposals/0000-unsatisfiable.rst
```diff
@@ -0,0 +1,599 @@
...
+<https://github.com/ghc-proposals/ghc-proposals/pull/59>`_ and `proposal #278
+<https://github.com/ghc-proposals/ghc-proposals/pull/278>`_ were more ambitious
+attempts to improve custom type errors, but both have been abandoned due to
+their complexity.
+
+`#18978 <https://gitlab.haskell.org/ghc/ghc/-/issues/18978>`_ suggests
+introducing ``Annotate :: ErrorMessage -> Constraint -> Constraint`` where
+``Annotate e c`` renders the message ``e`` if the constraint ``c`` cannot be
+solved, and is equivalent to ``c`` otherwise.  This is similar to
+``WithMessage`` from `proposal #59
```
aspiwack, line 549: I think that having `Annotate e c` as a wanted is the same as having `Unsatifisiable e => c` as a wanted (it's different as a given, I think). It it exist, it should probably have the same restrictions.

aspiwack (commented):
Thanks for this proposal. I've made a handful of comments below.

goldfirere:
At first read, I was unsure about skipping the fundep consistency check -- that is, why would `Unsatisfiable` mean to skip anything with fundeps? But then I remembered my https://github.com/ghc-proposals/ghc-proposals/issues/391#issuecomment-760521361, which attempts to define a meaning of consistency. It indeed suggests that `Unsatisfiable` should block the consistency check.

Nothing here to act on, really, but I wanted to share the small insight.

proposals/0000-unsatisfiable.rst
aspiwack: I had misread this paragraph. But I think this is less radical than the alternative section is making it: I can _already_ skip defining methods, if I do so I get a warning (and a runtime error were I to use the missing method)

```
> instance Num Bool

<interactive>:1:10: warning: [-Wmissing-methods]
        ‘+’, ‘*’, ‘abs’, ‘signum’, ‘fromInteger’, and (either ‘negate’
                                                              or
                                                              ‘-’)
    • In the instance declaration for ‘Num Bool’

> True + False
*** Exception: <interactive>:1:10-17: No instance nor default method for class operation +
```

As for associated type: when defining an overlapping instance for error message purposes (a very real use-case), I don't have a choice, I _must_ leave the associated type undefined (again, GHC produces a warning). Because type family instances must be non-overlapping (well, the condition is a bit finer than this, but for this use-case it doesn't matter).

I think that this warning should be skipped when `Unsatisfiable` is in the context (I can, of course, deactivate it myself, but it makes sense in the context)

adamgundry: @aspiwack thanks, you've convinced me that omitting class methods when there is `Unsatisfiable` in the context is a reasonable thing to do. Ideally the omitted methods (where there is no default either) would be implemented via a call to `unsatisfiable` (rather than an error about the method being undefined).

I hadn't previously thought about associated type families. I think in the absence of [Constrained Type Families](https://arxiv.org/abs/1706.09715), the best we can do is say that we skip the warning for a missing associated type instance if there is an `Unsatisfiable` constraint in the context.

aspiwack: > Ideally the omitted methods (where there is no default either) would be implemented via a call to `unsatisfiable` (rather than an error about the method being undefined).

Indeed. Though it doesn't matter very much. If someone ends up changing the behaviour to use `unsatisifiable` in this case, then won't need a proposal.

> I hadn't previously thought about associated type families. I think in the absence of [Constrained Type Families](https://arxiv.org/abs/1706.09715), the best we can do is say that we skip the warning for a missing associated type instance if there is an `Unsatisfiable` constraint in the context.

I think so too.

phadej: > Indeed. Though it doesn't matter very much. If someone ends up changing the behaviour to use unsatisifiable in this case, then won't need a proposal.

Some committee members are/were vorried about what *precisely* happens with `-fdefer-type-errors`. I feel there is mixed wishes here.

adamgundry:
@goldfirere that's a nice suggestion. As a bikeshed colour choice, how about `custom:` in place of `type-warning=` so we get options like `-Werror=custom:foldable` and `-Wno-custom`?

Although now I'm wondering if we'll want to control how aggressive the warning is (i.e. is it on by default, requires `-Wall`, or isn't even in `-Wall`). It might be a bit controversial to introduce `instance Warning "foldable" Msg => Foldable (a, b)` if all warnings are on by default, but it's a much easier change to sell if such warnings are shown only to those who opt in to them.

treeowl:
@adamgundry, a few standard warning categories could help that, I think. Deprecation warnings, partiality warnings, performance warnings, and mistake-proneness warnings come to mind immediately. A `foldable` warning would be a mistake-proneness warning. I imagine we'd only want to enable deprecation warnings by default and probably even in `-Wall`.

adamgundry:
> what if we have the same warning flag in two different places but with different bins? That will surely be confusing to users

I'm not too worried about this. True, GHC currently satisfies the property that every warning flag belongs to a single bin. But I don't think there is anything fundamentally wrong or confusing about violating it, e.g. if `-Wcustom:code-smells` turns on some warnings with `-W` and gets more noisy under `-Wall`.

I'd be more concerned about the possibility that with `GetWarningBin`, we could easily have two independent libraries map the same symbol name with different bins, whereupon it would be impossible to use them together. To avoid that we could replace the symbol with something properly namespaced (an open data kind, if we had it), but that seems unnecessarily complex to me.

Returning the fundep idea, though, we could do something like this:
```hs
data WarningBin = Wdefault | W | Wall | Weverything | Wcustom Symbol

type WarningInBin :: Symbol  -> WarningBin  -> Constraint
class WarningInBin flag bin | flag -> bin  -- functional dependency optional

type Warning :: Symbol -> ErrorMessage -> Constraint
class Warning flag msg
```

That is, `Warning` takes only a flag name, and flags can be put in bins by writing instances for the `WarningInBin` class. When reporting a custom warning, the type-checker looks for instances of `WarningInBin "flag" bin` to determine which bin(s) to use (and uses a sensible default if there are no instances). I like the fact that bins need be defined only once for each flag (or not at all, if the user accepts the default), rather than cluttering up `Warning` use sites.

With the functional dependency we don't have a global guarantee that two independent libraries will not assign the same symbol to different bins, but (I argue) that's a virtue, and we do at least know that a single library won't put the same flag in two different bins.

I'm not sure this is really worth the bother either, to be honest. It still doesn't handle computed flag names, but then neither does anything else.

goldfirere:
You're taking advantage of the fact that fundeps aren't as principled as type families: GHC does not check for global consistency. But I see how that's an advantage.

If you're unconcerned about one flag living in multiple bins, I suppose I could be unconcerened, too.

adamgundry:
In the light of the above discussion, I think it might be sensible to remove ``Warning`` from the present proposal and instead submit a companion proposal that focuses exclusively on ``Warning``?

i-am-tom:
> > one of the things I miss about PureScript is the ability to define TODOs as type-level warnings that I can clean up as I go,
> 
> Could you point to a good description of this feature of PureScript? Sounds intriguing.

@simonpj not a particularly complex example, but hopefully illustrates the point:

```purescript
module Main where

import Prelude (pure)
import Effect (Effect)
import Prim.TypeError (class Warn, Text)

f :: forall x. Warn (Text "TODO: implement this") => x -> Effect x
f = pure

g :: Effect String -- Raises a warning (non-fatal)
g = f "hi"
```

`Warn` is documented (at least for end users) within the `Prim.TypeError` module here: https://github.com/purescript/documentation/blob/master/guides/Custom-Type-Errors.md

@adamgundry I think that might be sensible. Given the discussion here and on the mailing list so far, splitting the proposal should mean we have one half that we're generally happy with, and one that merits further discussion. I'll mark the proposal as "needs revision" for now, if you're ok to split it up? At that point, I'll resubmit the `Unsatisfiable` half to the committee, and we can continue to work out the details with `Warning`. (/cc @nomeata)

adamgundry:
I've created a separate proposal for `Warning` at #454. I'll revise this proposal in due course to remove `Warning` and take into account the other feedback from the committee.

adamgundry:
I have amended the proposal as follows:
 * The `Warning` constraint has now been removed in favour of #454.
 * An `Unsatisfiable` Given constraint in an instance context now allows class method and associated type definitions to be omitted without a warning.
 * The presentation has been revised in the light of detailed feedback from the committee. In particular, the motivation for some changes (and the effect of the proposal on the issues raised in the motivation section) is now more explicit. Moreover, the proposed change specification is more structured.

@nomeata @i-am-tom I'm not sure if the proposal process now requires further time for open discussion, or if the committee might reconsider this now?

nomeata:
That's up to @i-am-tom to assess and drive forward

adamgundry:
@i-am-tom ping?

i-am-tom:
@adamgundry Thanks, this one fell off my radar! I'll talk to the committee again today.

goldfirere:
This proposal has been accepted and merged. Thanks, all!

adamgundry:
This has been implemented for 9.8.1, thanks to @sheaf! https://gitlab.haskell.org/ghc/ghc/-/merge_requests/10315

PR #434, opened 2021-08-13
Fine-grained unused warnings

JakobBruenker:
This proposal aims to make warnings about unused names more precise, consistent, and configurable.

[Rendered](https://github.com/JakobBruenker/ghc-proposals/blob/fine-grained-unused/proposals/0000-fine-grained-unused-warnings.rst)

### DIFF
--- a/proposals/0000-fine-grained-unused-warnings.rst
+++ b/proposals/0000-fine-grained-unused-warnings.rst
@@ -1,0 +1,208 @@
+Fine-Grained Unused Warnings
+============================
+
+.. author:: Jakob Brünker
+.. date-accepted::
+.. ticket-url::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/434>`_.
+.. contents::
+
+This proposal aims to make warnings about unused names more precise,
+consistent, and configurable.
+
+
+Motivation
+----------
+
+::
+
+  a :: Int
+  a = 4
+    where bar = foo
+          foo = 3
+
+Compiling the above code with ``-Wall`` results in the
+these warnings:
+
+::
+
+      Defined but not used: `bar'
+      Defined but not used: `foo'
+
+Sometimes, I look at a warning like this and think "I could have sworn I used
+``foo`` somewhere" - but of course, what the warning is trying to say isn't
+actually that ``foo`` is unused, but rather that all of ``foo``'s usage sites
+are in unused bindings.
+
+Furthermore, while it's useful to see such a warning for CI, to immediately
+know all of the bindings that have to be removed to eliminate the warnings,
+during development, it would be enough for me to see a warning about the
+binding that's actually unused; the rest serves mostly to clutter up valuable
+screen real estate in ghcid.
+
+On the other hand, for a module like
+
+::
+
+  module Main (main) where
+
+  import Control.Monad (when)
+
+  main = pure ()
+
+  foo cond = when cond (putStrLn "foo")
+
+GHC will warn about ``foo`` being unused, but it will not warn about the import
+being unused, even though all it's usages are in unused bindings. But here,
+too, it would be useful to know that the import is unused during CI.
+
+A further inconsistency is that in
+
+::
+
+  import Control.Monad (when, when)
+  import Control.Monad (when, when)
+
+  foo cond = when cond (putStrLn "foo")
+
+GHC will warn about the second import, but not about the second occurrence in
+the import list of ``when``.
+
+Proposed Change Specification
+-----------------------------
+The overarching idea is to provide a consistent set of warnings about unused
+names, that are disabled by default, enabled by ``-Wall``, and can be
+configured to only show genuinely unused names.
+
+This affects the following warnings:
+
+* ``-Wunused-top-binds``: Warn if a top-level binding is not exported and does
+  not appear in another definition.
+* ``-Wunused-local-binds``: Warn if a local binding does not appear in another
+  definition.
+* ``-Wunused-pattern-binds``: Warn if a pattern does not bind any names and is
+  not a wild-card or bang pattern.
+* ``-Wunused-binds``: Alias for the three warnings above.
+* ``-Wunused-matches``: Warn if a name bound in a term-level pattern binding
+  does not appear in the right-hand side of a definition.
+* ``-Wunused-imports``:
+
+  * Warn if no bindings from an import are used and the import has no import
+    list or a non-empty import list.
+  * Warn if any individual name in an import list is not used.
+  * Warn if a name or module is imported twice. This does not apply if the
+    imports have different effects, e.g. one qualified and one unqualified
+    import. This is a change from the status quo, in which only repeated
+    modules would trigger this warning, not repeated individual names in the
+    import list. Furthermore, if a module has already been imported, the status
+    quo is to not trigger this warning if a second import has an empty import
+    list. Under this proposal, such a situation will also trigger this
+    warning.
+
+For each of these warnings, this proposal adds a transitive version, denoted by
+the name followed by a ``-transitive`` suffix (e.g. the transitive version of
+``-Wunused-binds`` is ``-Wunused-binds-transitive``. Rather than being
+triggered when a name is not used, this warnings will be triggered when a name
+is "transitively unused". A name is transitively unused if all of its usage
+sites are in unused or transitively unused definitions.
+
+An exception is ``-Wunused-pattern-binds``. A transitive version would not make
+sense, since the warning is about *not* binding to a name.
+
+The warning messages will be worded identically to their non-transitive
+counterparts, except that occurrences of "not used", "unused", and "redundant"
+will be replaced by "used only in unused or unreachable bindings", potentially
+accompanied by minor adjustments to the phrasing to make the message work
+grammatically. "Unreachable" in this case is effectively another way of saying
+"transitively unused but not (directly) unused".
+
+These warnings will have no effect if the corresponding non-transitive warning
+isn't enabled.
+
+Furthermore, the warning ``-Wunused-transitive`` is provided as an alias for
+
+* ``-Wunused-binds-transitive``
+* ``-Wunused-matches-transitive``
+* ``-Wunused-foralls-transitive``
+* ``-Wunused-imports-transitive``
+
+This means that all transitive warnings can be turned off at once with
+``-Wno-unused-transitive``.
+
+Related warnings that are not affected by this proposal are
+``-Wunused-packages`` and ``-Wunused-type-patterns``. These warnings are not
+enabled by ``-Wall``, and don't require a transitive counterpart.
+
+``-Wunused-do-bind`` also remains unaffected, and does not need a transitive
+version for the same reason as ``-Wunused-pattern-binds`` above.
+
+Examples
+--------
+
+::
+
+  module Main (main) where
+
+  import Prelude (Bool, IO, putStrLn, putStrLn)
+  -- warning: [-Wunused-imports]
+  -- The import of `putStrLn' from module `Prelude' is redundant
+
+  import Prelude ()
+  -- warning: [-Wunused-imports]
+  -- The import of `Prelude' is redundant
+
+  import Control.Monad (when)
+  -- warning: [-Wunused-imports-transitive]
+  -- `when', imported from module `Control.Monad`, is only used in unused or
+  -- unreachable bindings
+
+  foo :: forall a . (forall (b :: a) . Bool -> IO ())
+  -- warning: [-Wunused-foralls-transitive]
+  -- Quantified variable `a' is only used in unused or unreachable bindings
+
+  -- warning: [-Wunused-foralls]
+  -- Unused quantified variable `(b :: a)'
+
+  foo cond = when cond (putStrLn "foo")
+  -- warning: [-Wunused-top-binds-transitive]
+  -- Defined but only used in unused or unreachable bindings: `foo'
+
+  bar = foo
+  -- warning: [-Wunused-top-binds]
+  -- Defined but not used: `bar'
+
+  main = putStrLn "Hello, World!"
+
+Effect and Interactions
+-----------------------
+For the most part, the effects of this proposal are minor. Adding the
+transitive warnings to ``-Wall`` means that the only differences for existing
+code-bases are that the warning message GHC prints for the transitive warnings
+is different, and that GHC will print a few more warnings about unused names
+to be consistent with the previously existing warnings.
+
+Costs and Drawbacks
+-------------------
+None that I can see.
+
+Alternatives
+------------
+* It would be possible to (assuming ``-Wall``) make the transitive warnings
+  opt-in rather than opt-out. This would make the proposal less
+  backwards-compatible, since most of the transitive warnings are currently
+  part of the unused warnings and thus enabled by ``-Wall``.
+
+* Instead of having no effect when the corresponding unused warning is turned
+  off, transitive warnings could imply their non-transitive counterparts, or
+  GHC could show only transitive warnings (but no warnings about geniwuinely
+  unused bindings). Which of these behaviors would be the least surprising is
+  somewhat subjective.
+
+* Instead of having the ``-transitive`` suffix, the word ``unused`` in the
+  transitive warnings could be replaced by ``unreachable``.
+
+Implementation Plan
+-------------------
+I can implement it.


simonpj:
Looks plausible to me.

Is there a precendent for the transitive/intransitive distinction in any other programming language?

JakobBruenker:
> Is there a precendent for the transitive/intransitive distinction in any other programming language?

There is [a somewhat related discussion](https://github.com/rust-lang/rust/issues/39531) in the rust community, but I wasn't able to find any language that has implemented a similar concept.

nomeata:
Thanks for helping to refine the developer experience! From a procedural point of view, I would consider this a change to the compiler UX that doesn’t _require_ a full proposal process. If you are happy to implement and document it, and if the GHC devs (e.g. Simon) is happy with the direction, I’d say Go ahead!

You may also opt-into the full process nevertheless, e.g. if you think that the comittee review will benefit the design.

In either case you are welcome to use this space to draft the design.

simonpj:
I urged Jakob to write it up as a proposal, because doing so attracts very constructive feedback from a broad audience -- and the design work needs to be done anyway.

JakobBruenker:
@nomeata there's a couple of subjective points that might make it worth having the committee take a look:

@georgefst wrote on reddit

> I'd opt for the transitive warnings to be opt-in. They always just seem like clutter to me.

This is one thing I'm unsure about - In principle, I'd also prefer having the transitive warnings turned off by default, but on the other hand, it would hurt discoverability: It's easy to see how to turn them off if they're on (GHC tells you the warning name when a warning is triggered), but if they're turned off, the only way to find out they exist is to stumble across them in e.g. the User guide.

/u/thefybe wrote on reddit

> What are the semantics for recursive and especially mutually recursive functions?

Which the proposal doesn't address currently.
I think it makes sense to warn about both with just the regular unused warnings, not the transitive unused warnings.
Though probably with a more precise error message than they get currently, e.g. "Defined but only used recursively", and "Defined but only used in a mutually recursive set of bindings"

tomjaguarpaw:
> it would be enough for me to see a warning about the binding that's actually unused; the rest serves mostly to clutter up valuable screen real estate in ghcid

There is a tension here

* Listing *all* the unused bindings tells the user the set she has to delete to fix the problem
* Listing only the unused "leaves" tells the user the minimum set she has to use to fix the problem 

Could we resolve this tension, and avoid adding another layer of configurability, by warning once for each unused "leaf" and once *in total* for all the other therefore transitively unused bindings?  This seems to provide the best of both worlds. For example

```
warning: foo is unused

warning: bar is unused

warning: baz, quux, wibbles, xyzzy are transitively unused
```

JakobBruenker:
@tomjaguarpaw that's not a bad idea. How would you handle recursive and mutually recursive bindings with that?

tomjaguarpaw:
Indeed, it's actually a tree of bindings, but a tree of strongly connected components of the binding graph. I don't see the problem with non-mutually recursive bindings but for mutually recursive bindings I guess you have two options

1. Arbitrarily choose one representative of each "leaf" SCC
2. List all elements of each "leaf" SCC

For the program `foo = bar; bar = baz; baz = quux; quux = wibbles` 1 would show the warnings

```
warning: foo is unused

warning: baz, quux, wibbles are transitively unused
```

and 2 would show the warning

```
warning: foo, baz, quux, wibbles are an unused mutually-recursive group
```

I prefer 2 although I'm sure improvements could be made to each.

nomeata:
Yay, brainstorming time!

I tend to find it useful to get a list of all unreachable bindings - usually I use the warnings to know what to delete. Much more rarely than using them to find I missed something (and in that case the complete list is not too much a distraction).

In general I applaud any attempt to avoid adding configurability, when possible.

So maybe a good warning phrasing can satisfy all? List all unreachable bindings, but highlight the leaves (or roots?).

For mutually recursive I would _not_ pick a random one, but treat them equally - more in line with Haskell's (highly appreciated) irrelevance of ordering. (But I'm not entirely sure on this point yet.)

Maybe something like this; aggregating all warnings of a module into one (like the tabs warning):
```
The following bindings neither mentioned nor exported:
12:34: foo 
14:56: bar
Because of that, these bindings are unreachable:
34:2: Baz
```

This is concise, lists the possibility erroneously unused bindings first, but also shows the complete list. More careful phrasing of the first line would allow mutually recursive roots in the first group (which I think is fine from a UX POV).

Maybe the warnings needs to be structured to be easily parsable by generic tools. In that case I could live with just _one_ unused binding being recognized as the error location (and bet on LSP to ergonomically flag all locations for squiggly underlining).

JakobBruenker:
> Maybe the warnings needs to be structured to be easily parsable by generic tools. In that case I could live with just one unused binding being recognized as the error location (and bet on LSP to ergonomically flag all locations for squiggly underlining).

There is precedence for something similar for duplicate declarations:

```
File.hs:332:1: error:
    Multiple declarations of ‘a’
    Declared at: File.hs:328:1
                 File.hs:332:1
    |
332 | a = 6
    | ^
```

tomjaguarpaw:
> highlight the leaves (or roots?)

Yes, "roots" might be better use of terminology.

michaelpj:
Let me throw my hat into the ring. I agree that listing the "roots" of deadness is very helpful, but I also think that having per-binding warnings is good - this makes it *much* easier for tools. So how about:

Unused binding: `warning: foo is unused`
Transitively unused binding: `warning: foo is only used transitively by the following unused bindings: baz, qux`

For recursive groups, personally I would just disregard usage within the group and mark every member of the group as directly unused if none of them are used outside the group. If we really wanted to we could add `foo is unused (except recursively)` to the message, but I think that probably it would not confuse anyone if that was omitted.

sullyj3:
> `warning: foo is only used transitively by the following unused bindings: baz, qux`

This is a nice suggestion!
I think you could probably drop the word "transitively" without loss of meaning, which might be friendlier to beginners.

JakobBruenker:
@michaelpj If I understand correctly with that approach we would have the same amount of warnings as we have currently, which means being able to turn turn off transitive warnings as in the current version of the proposal would still be useful? (as opposed to @tomjaguarpaw's and @nomeata's suggestions which combined the warnings in part to eliminate the need for configurability)

michaelpj:
>  I think you could probably drop the word "transitively" without loss of meaning, which might be friendlier to beginners.

Yeah, you're probably right. Perhaps "indirectly": I think it's helpful to say something to avoid people thinking that `baz` and `qux` are the things that *directly* use `foo`.

> If I understand correctly with that approach we would have the same amount of warnings as we have currently, which means being able to turn turn off transitive warnings as in the current version of the proposal would still be useful?

Yep. Personally I would just have one flag that controls both, but I think people differ in how annoying they find the transitive warnings. If they're easily distinguishable then it doesn't seem so bad.

chreekat:
> this makes it *much* easier for tools.

Do we need a #378 - style statement of ambition about providing independent machine- and human-focused error output? :)

I could be wrong, but it looks like the proposal is concerned with human output ("cluttering ghcid screen real estate"). In my ideal future, we can consider that problem completely independently of what is or is not pleasing to a language server. But I'm not sure how realistic that is *today".

nomeata:
That discussion came up before. In an ideal world we'd have `stderr` only used by humans, and all machine-consumption goes via LSP. Until then, we also have to keep the `stderr`-parsed-by-language-agnosic-tools mode working reasonably well. But I’d say we can make careful pragmatic compromises here, e.g. in `stderr`-output group very similar warnings under the first occurrence (like with “tab used” warning). Presumably that’s better for the human reader, while not breaking and still useful the stderr-parsing tools.

sullyj3:
Yeah, having a separate output that's specifically designed for tooling strikes me as The Right Thing

goldfirere:
Just looking at old threads that I've fallen behind on. I really like what I see here. Just a gentle prod to reawaken the thread! :)

JakobBruenker:
I'll get back to it very soon :)

Jadefalke256:
In my opinion we should keep imports and unused top binds separated, but I'm fine and would prefer transitively unused binds to be combined in one warning.
My idea of what this feature currently should look like for a file `foo.hs`
```
import Data.List

bar :: a -> a
bar = foo

foo :: a -> a
foo = id

main :: IO ()
main = undefined
```
is to issue the warnings
```
[1 of 1] Compiling Main             ( foo.hs, foo.o )

foo.hs:1:1: warning: [-Wunused-imports]
    The import of ‘Data.List’ is redundant
      except perhaps to import instances from ‘Data.List’
    To import instances alone, use: import Data.List()
  |
1 | import Data.List
  | ^^^^^^^^^^^^^^^^

foo.hs:4:1: warning: [-Wunused-top-binds-transitive]
    Defined but not used: ‘bar’
  |
4 | bar = foo
  | ^^^
    as a consequence the following binds are transitively unused:
      declared at foo.hs:7:1:
      |
    7 | foo = id
      | ^^^
```
where the nesting of the error also makes it more clear which is the parent / root unused bind with its children being indented.

JakobBruenker:
Thank you @Jadefalke256, I've been working on rewriting the proposal to accommodate the discussed points, I'll post an updated version later today.

It sounds like people prefer avoiding added configurability, so I would probably say we don't have `-Wunused-top-binds-transitive`, and transitive binds just show up as part of the warning as long as `Wunused-top-binds` is enabled.

One thing I'm still a little unsure about is if you have something like

```
module Foo ()

import Data.List as L

foo = L.intercalate
```
But I think probably this should produce something like
```
Foo.hs:5:1: warning: [-Wunused-top-binds]
    Defined but not used: ‘foo’
  |
5 | foo = L.intercalate
  | ^^^
    as a consequence the following imports are transitively unused [-Wunused-imports]:
      imported at foo.hs:3:1:
      |
    3 | import Data.List as L
      | ^^^
```

In other words what I'm thinking is, transitively unused imports are reported if *both* `-Wunused-imports` is enabled, *and* the unused warning for whatever thing the import is used in is enabled; that seems like a decent compromise between configurability and simplicity.

JakobBruenker:
Okay, getting around to finishing the edit took me longer than I expected, but here it is.

@Jadefalke256 feel free to let me know if you have any thoughts, and anyone else as well, of course.

If there are no comments, I will ask the proposal to be sent to the committee in a week.

Jadefalke256:
Looks good to me! Thank you for writing this up.
I have started looking into the source to make a plan for tackling this!

JakobBruenker:
@nomeata Looks like this is ready, I'd like to submit to the committee

cdornan:
@JakobBruenker like everyone else I think this is solving an important problem. Furthermore it looks like the identified outstanding technical issues have been resolved. I will email the list today recommending acceptance.

adamgundry:
Do we know if this change will impact existing tools? (I don't know how many have moved to consuming structured errors rather than parsing error message texts. Obviously the former is preferable.)

It would be helpful for the proposal text to include the existing messages in the examples. I think this proposed changes mostly include more information (e.g. because currently `unused-imports` warnings don't show up at all if there are `unused-binds`) but it's a bit hard to tell.

In general I think GHC's error messages are excessively wordy. Could the messages proposed here be shortened? For example instead of
```
Foo.hs:5:1: warning: [-Wunused-top-binds]
    Defined but not used: ‘foo’
    as a consequence the following imports are transitively unused [-Wunused-imports]:
      Data.List imported at foo.hs:3:1:
    and the following top-level binds are transitively unused [-Wunused-top-binds]:
      bar defined at foo.hs:7:1:
    and the following local binds are transitively unused [-Wunused-local-binds]:
      baz defined at foo.hs:8:9:
```
what about
```
Foo.hs:5:1: warning: [-Wunused-top-binds]
    Defined but not used: ‘foo’
    so this import is unused (except perhaps to import instances) [-Wunused-imports]:
      Data.List imported at Foo.hs:3:1
    and this top-level binding is unused [-Wunused-top-binds]:
      ‘bar‘ defined at Foo.hs:7:1
    and these local bindings are unused [-Wunused-local-binds]:
      ‘baz‘ defined at Foo.hs:8:9
      ‘quux‘ defined at Foo.hs:9:10
```

simonpj:
What became of this suggestion?

- **Unused binding**: `warning: foo is unused`
- **Transitively unused binding**: `warning: foo is only used indirectly by the following unused bindings: baz, qux`

That idea seems like plausible to me -- perhaps with a flag `-freport-indirectly-unsed-bindings` to allow or suppress the warning on indirectly-unused bindings.  But the proposal does not describe this alternative, or say why an alternative was chosen.

cdornan:
@JakobBruenker while I liked your latest draft, some important points need to be addressed &mdash; at least you need to explain why your current proposl is preferrable. (Personally, I am hoping you will be able to defend your current proposal or synthesize enough of Simon's and Tom's suggestions to meet their requirements.)

Per Adam's comments we should get at least some HLS input on how disruptive this is likely to be.

simonmar:
> Do we know if this change will impact existing tools?

I think the emacs compile mode (which I use occasionally) will miss the nested warnings. It skips to the next top-level error/warning when you do `next-error` (C-x `)

The alternative proposal [mentioned by Simon above](https://github.com/ghc-proposals/ghc-proposals/pull/434#issuecomment-1712249862) would fix that.

aspiwack:
I find myself having the same reservations as the Simons, about how the warnings are formatted. I'll add the following remark to the discussion:

- The motivation of this proposal includes the fact that the transitively unused dependencies waste vertical space in the ghcid output. But I don't see, reading this proposal, how it addresses this problem.
- What happens is `a` is used in both, and only, in `x` and `y`, but `x` and `y` are both (transitively) unused? Does `a` appear as a consequence of `x` being unused? of `y`? of both? (note that if the answer is “both” it may make the situation worse than the status quo for ghcid)
- Even admitting that the formatting in this proposal helps ghcid, it seems to make it worse than the status quo for most other IDE. Simon M mentionned Emacs's compile mode, but the LSP also wants to have separate warnings for each identifier to be able to highlight them in the source.

JakobBruenker:
Thanks, these are some good points. I'll try to address everything/incorporate things into the proposal over the course of the next day or two.

cdornan:
@JakobBruenker just a quick check-in &mdash; are you still OK to make those revisions?

JakobBruenker:
@cdornan yep, sorry for the silence here, took me some time to figure out how to best address everything. I'll finish my changes over the weekend.

JakobBruenker:
Sorry it took so long but I've now committed the changes I've made. In the end it seemed like the only way to address all the points that were brought up was to go back to the earlier idea of having separate warnings. This should not affect tooling in any significant way.

JakobBruenker:
(cc @cdornan)

cdornan:
@JakobBruenker thanks very much &mdash; I will review this weekend (do feel free to ping me for progress updates).

proposals/0000-fine-grained-unused-warnings.rst
```diff
@@ -0,0 +1,205 @@
...
+  Foo.hs:3:1: warning: [-Wunused-imports, -freport-indirectly-used-bindings]:
+      The import of ‘Data.List’ is only used by the following unused binding: ‘foo’
+        except perhaps to import instances from ‘Data.List’
+      To import instances alone, use: import Data.List()
+
+  Foo.hs:5:1: warning: [-Wunused-top-binds]
+      Defined but not used: ‘foo’
+
+  Foo.hs:7:1: warning: [-Wunused-top-binds]
+      ‘bar' is defined but only used in the following unused binding: `foo’
```
adamgundry, line 151: Being pedantic, almost all occurrences of "only used" in the proposal should be "used only".

proposals/0000-fine-grained-unused-warnings.rst
```diff
@@ -0,0 +1,205 @@
...
+      ‘wibble' is defined but only used in the following unused binding: `worble’
+
+  Foo.hs:11:9: warning: [-Wunused-local-binds]
+      ‘worble' is defined but only used in the following unused binding: `wibble’
+
+  Foo.hs:12:9: warning: [-Wunused-local-binds]
+      Defined but not used: ‘wirble’
+
+  Foo.hs:13:15: warning: [-Wunused-foralls]
+      Foo quantified type variable ‘a' is only used in the following unused variable: `(b :: a)’
```
adamgundry, line 166: "Foo" in this message and the one below seems out of place?

proposals/0000-fine-grained-unused-warnings.rst
```diff
@@ -0,0 +1,205 @@
...
+and bar is unused, a question that arises is whether the warning about ``foo`` should reference ``bar``, or ``baz``.
+
+In this particular case, since there would be no warning about ``baz`` being unused, it seem more helpful if the warning references ``bar``.
+
+In general, a warning should only reference other bindings about which there is an "unused" warning.
+
+In practice, that typically means warnings about transitively unused top-level bindings will tend to reference other top-level bindings, and warnings about transitively unused
+local bindings will reference other local bindings. However, if ``baz`` were unused, it would be referenced in the warning instead, since using ``bar`` in that case would not get rid of the warning about ``foo``.
+
+If a binding ``b`` is only used recursively, the warning message will say that it is only used in ``b``.
```
adamgundry, line 104: This doesn't seem to match the example, and indeed I think it is better to say that `b` is unused than to say to it is used only by itself.

proposals/0000-fine-grained-unused-warnings.rst
```diff
@@ -0,0 +1,205 @@
...
+    where baz = foo * 2
+	
+and bar is unused, a question that arises is whether the warning about ``foo`` should reference ``bar``, or ``baz``.
+
+In this particular case, since there would be no warning about ``baz`` being unused, it seem more helpful if the warning references ``bar``.
+
+In general, a warning should only reference other bindings about which there is an "unused" warning.
+
+In practice, that typically means warnings about transitively unused top-level bindings will tend to reference other top-level bindings, and warnings about transitively unused
+local bindings will reference other local bindings. However, if ``baz`` were unused, it would be referenced in the warning instead, since using ``bar`` in that case would not get rid of the warning about ``foo``.
```
adamgundry, line 102: I'm finding this section a bit hard to follow. It would be nice to nail down the specification a bit more precisely: instead of discussing hypothetical designs and what "should" be the case, can we give a clear definition of which bindings are included in the list? (By all means unpack the meaning of the definition with examples and discussion of why it is the right design.)

proposals/0000-fine-grained-unused-warnings.rst
```diff
@@ -0,0 +1,205 @@
...
+      In the type signature for ‘far’
+
+  Foo.hs:14:1: warning: [-Wunused-top-binds]
+      Defined but not used: ‘far’
+
+Effect and Interactions
+-----------------------
+For the most part, the effects of this proposal are minor. The only differences for existing
+code-bases are that the warning message GHC prints for the transitive warnings
+is different.
```
adamgundry, line 180: And we get more warnings about indirectly unused imports and foralls, I think?

proposals/0000-fine-grained-unused-warnings.rst
```diff
@@ -0,0 +1,205 @@
...
+
+Since the warnings don't have any special formats, existing tools should be able to handle them without issues.
+
+Users that don't wish to see warnings about transitively unused bindings can turn those warnings off.
+
+Costs and Drawbacks
+-------------------
+The warning mechanism is somewhat more complicated and as a consequence might
+have a somewhat higher maintenance cost.
```
adamgundry, line 189: It is a bit of a shame that the new `-f` flag is somewhat separate from the `-W...` warning flag system, but changes which warnings are emitted. I wonder if there is a way to express this using warning groups (e.g. say that `-Wunused-imports` is split into `-Windirectly-unused-imports` + `-Wdirectly-unused-imports`, and have `-Wno-indirect-uses` to turn off all the "indirectly used" warnings at once). I haven't thought through all the consequences of this, though, and you may have considered it already.

adamgundry (commented):
Thanks for your efforts on this proposal @JakobBruenker! I'm generally in favour of improving the warnings along the lines you propose, modulo some discussion of the details. Apologies if I've missed anything that has been discussed already...

proposals/0000-fine-grained-unused-warnings.rst
```diff
@@ -0,0 +1,205 @@
...
+during development, these additional warnings take up a lot of valuable
+screen real estate in ghcid.
+
+Proposed Change Specification
+-----------------------------
+The overarching idea is to separate warnings about names that are genuinely
+unused, and warnings that are only "transitively unused" (or "indirectly unused"), i.e. only used by
+other (transitively) unused bindings.
+
+The warnings about transitively unused names will be reported if a new flag, `-freport-indirectly-unsed-bindings` is enabled (by default, it is) *and* the appropriate existing warning flags about unused bindings is enabled, i.e. one of
```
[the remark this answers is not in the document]
JakobBruenker: I actually did mean `indirectly-unused`. I've added that phrasing to alternatives, although I think I don't really like it, because I'm not sure what it means for a use to be indirect.

proposals/0000-fine-grained-unused-warnings.rst
```diff
@@ -0,0 +1,205 @@
...
+  
+  Foo.hs:3:1: warning: [-Wunused-imports, -freport-indirectly-used-bindings]:
+      The import of `Data.List' is only used by the following unused binding: ‘foo’
+        except perhaps to import instances from `Data.List'
+      To import instances alone, use: import Data.List()
+
+The warning about transitively unused bindings only appears if both the warning flag
+for the unused binding (in this case ``-Wunused-top-binds``) and the the one for
+the transitively unused binding itself (in this case ``-Wunused-imports``) are
+enabled.
```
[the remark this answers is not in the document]
JakobBruenker: Good point, in this particular case it doesn't look like that would be a change from the current behavior, but in general it could happen that the number of warnings a code base produces changes, in cases where two warning flags interact. I've changed the text accordingly.

proposals/0000-fine-grained-unused-warnings.rst
```diff
@@ -0,0 +1,205 @@
...
+  Foo.hs:6:1: warning: [-Wunused-top-binds]
+      ‘b1’ is defined but only used in the following unused bindings: ‘b2’, ‘b4’
+
+Related warnings that are not affected by this proposal are ``-Wunused-pattern-bindings``,
+``-Wunused-packages``, and ``-Wunused-do-bind``, since these are not about binding names.
+
+Examples
+--------
+
+NB: The only difference between these examples and how these warnings are currently presented is that the warnings about transitively unused bindings currently don't list the bindings in which they occur, and have the same phrasing as regular unused warnings - with the exception of the unused import warning and unused forall warnings, which don't currently occur if they are only transitively unused.
```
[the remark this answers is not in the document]
JakobBruenker: I've removed the note and added the current warnings for comparison.

proposals/0000-fine-grained-unused-warnings.rst
JakobBruenker: I've rewritten the section, let me know what you think

proposals/0000-fine-grained-unused-warnings.rst
JakobBruenker: I've added it to alternatives. A point that came up earlier in discussion is people actually wanting less configurability, so from that point of view n + 1 flags (as in the current text) may be better than 2*n + 1 flags (or 3*n + 1, since we'd probably also want to keep the existing `-Wunused-imports`?). It's also not clear to me that having separate "indirectly unused" warnings for each category is something that would be used in practice.

Perhaps the `-f` flag in the text could simply be a `-W` name instead? Not sure if it'd be violating any implicit rules about how `-W` flags are supposed to work.

JakobBruenker:
Thank you for the review @adamgundry. I have made some changes to the proposal to hopefully address your points.

JakobBruenker:
> do feel free to ping me for progress updates

@cdornan ping :)

cdornan:
@JakobBruenker I am sorry for the extended hiatus.  I have re-reviewed the proposal; I like it and I think it is ready to put before committee again.

proposals/0000-fine-grained-unused-warnings.rst
```diff
@@ -0,0 +1,223 @@
...
+Proposed Change Specification
+-----------------------------
+
+The proposed change aims to distinguish between genuinely (or directly) unused bindings and transitively (or indirectly) unused bindings. While the warnings for directly unused bindings remain unchanged, the warnings for indirectly unused bindings will now be controlled by a new flag, ``-freport-indirectly-unused-bindings``.
+
+**Criteria for Transitively Unused Bindings:**
+
+1. **Directly Unused Bindings:** A binding is considered directly unused if it is not referenced in any part of the code. These bindings will generate the usual warnings as before.
+
+2. **Transitively Unused Bindings:** A binding is classified as transitively unused if it is used only within other (directly or indirectly) unused bindings. The warning for these bindings will be reported only if ``-freport-indirectly-unused-bindings` and the relevant existing warning flags (e.g., ``-Wunused-top-binds``, ``-Wunused-local-binds``) are enabled.
```
aspiwack, line 51: What default value do you specify for `-freport-indirectly-unused-bindings`? My (rather strong) bias is to have it on by default.

(it does feel a little weird to me that this be a `-f` option, but maybe it's the best choice anyway)

proposals/0000-fine-grained-unused-warnings.rst
```diff
@@ -0,0 +1,223 @@
...
+    ::
+    
+      Foo.hs:6:1: warning: [-Wunused-top-binds]
+          ‘b1’ is defined but used only in the following unused bindings: ‘b2’, ‘b4’
+
+4. **Import and `forall` Bindings:** The proposal also extends to warnings about transitively unused imports and ``forall`` binds. Both are considered to be unused if they are used only in definitions or type declarations of unused bindings, with the same direct vs. indirect distinction.
+
+**Warning References and Messages:**
+
+- A binding will produce a warning if it is directly unused, or if it used only by bindings that are unused *and* produce a warning about being unused
```
aspiwack, line 68:
```suggestion
- A binding will produce a warning if either
  - it is directly unused, or
  -  it used only by bindings that are unused *and* produce a warning about being unused (and `-freport-indirectly-unused-bindings` is on)
```

cdornan:
@JakobBruenker I assumed that you intended `-freport-indirectly-unused-bindings` to be off by default but Arnaud's preference for switching it on by default makes a lot of sense. Can you add a note to clarify the default or did we miss it?

JakobBruenker:
@cdornan @adamgundry I agree it makes sense to enable it by default, since that matches the current behavior more closely. I've made a corresponding change in the proposal text.

proposals/0000-fine-grained-unused-warnings.rst
JakobBruenker: I'm definitely open to changing the flag prefix, although I'm not sure what would be better

proposals/0000-fine-grained-unused-warnings.rst
```diff
@@ -0,0 +1,225 @@
...
+Proposed Change Specification
+-----------------------------
+
+The proposed change aims to distinguish between genuinely (or directly) unused bindings and transitively (or indirectly) unused bindings. While the warnings for directly unused bindings remain unchanged, the warnings for indirectly unused bindings will now be controlled by a new flag, ``-freport-indirectly-unused-bindings``, which is enabled by default.
+
+**Criteria for Transitively Unused Bindings:**
+
+1. **Directly Unused Bindings:** A binding is considered directly unused if it is not referenced in any part of the code. These bindings will generate the usual warnings as before.
+
+2. **Transitively Unused Bindings:** A binding is classified as transitively unused if it is used only within other (directly or indirectly) unused bindings. The warning for these bindings will be reported only if ``-freport-indirectly-unused-bindings` and the relevant existing warning flags (e.g., ``-Wunused-top-binds``, ``-Wunused-local-binds``) are enabled.
```
simonpj, line 51: Typesetting bug: too much typewriter font

proposals/0000-fine-grained-unused-warnings.rst
```diff
@@ -0,0 +1,225 @@
...
+
+    An notable exception here is that local bindings are *not* considered transitively unused just because the top-level binding they are defined in is unused. They are only considered transitively unused if they are unused within the scope of the top-level definition. This is to avoid generating a lot of unhelpful warnings in these cases.
```
simonpj, line 53: Typesetting bug: this para comes out in grey with a sidebar

proposals/0000-fine-grained-unused-warnings.rst
```diff
@@ -0,0 +1,225 @@
...
+          ‘b1’ is defined but used only in the following unused bindings: ‘b2’, ‘b4’
+
+4. **Import and `forall` Bindings:** The proposal also extends to warnings about transitively unused imports and ``forall`` binds. Both are considered to be unused if they are used only in definitions or type declarations of unused bindings, with the same direct vs. indirect distinction.
+
+**Warning References and Messages:**
+
+- A binding will produce a warning if either
+  - it is directly unused, or
+  -  it used only by bindings that are unused *and* produce a warning about being unused (and `-freport-indirectly-unused-bindings` is on)
+    - This means that e.g. if a top-level bind is used only in an unused local bind, both ``-Wunused-top-binds`` *and* ``-Wunused-local-binds`` must be enabled.
```
simonpj, line 71: Typesetting bug: grey sidebar

proposals/0000-fine-grained-unused-warnings.rst
```diff
@@ -0,0 +1,225 @@
...
+    ::
+    
+      Foo.hs:6:1: warning: [-Wunused-top-binds]
+          ‘b1’ is defined but used only in the following unused bindings: ‘b2’, ‘b4’
+
+4. **Import and `forall` Bindings:** The proposal also extends to warnings about transitively unused imports and ``forall`` binds. Both are considered to be unused if they are used only in definitions or type declarations of unused bindings, with the same direct vs. indirect distinction.
+
+**Warning References and Messages:**
+
+- A binding will produce a warning if either
```
simonpj, line 68: Typesetting bug: list not laid out properly (add a blank line)

simonpj:
* There are a number of typesetting bugs which I have reported separately.

* Item 1 looks like a definition. But then its is modified by the first bullet of item 3 I think. It's not clear which binders in a mutually recursive group are "directly unused".   Let's be precise!

  Consider
  ```
  letrec { f = ...g..; g = ..f.. } in 3
  ```
  Which of `f` and `g` are directly unused?  It must be both or neither!  I think both should be.

  Maybe: *"A binding B is **directly unused** if it is referenced only in B's own strongly-connected component"*.  (You'd need to say a little about strongly connected components.)

* Note that definition would mean that both `foo` and `bar` are directly unused.
  ```
  foo = let bar = True in 3
  -- No references to foo
  ```
  I think that is fine.

* The definition of "transitively unsued" is also strange, becuase it relies on an unspecified "indirectly unused" term.   How about

  How about: *"A binding B is **transitively unused** if it is referenced only in B's own strongly-connected component, or the body of a transitively-unused binding in B's scope"*.

  The "in B's scope" part is to deal with your "notable exception".

* Search for "indirect" and replace with "transitive", or vice versa.

* "A binding will produce a warning if either ...". I found this bullet really hard to parse.  I *think* you mean:
  * If a top-level or local binding is directly unused, a warning will be produced if `-Wunused-top-binds` or `-Wunused-local-binds` (repectively) is on.
  * If a top-level or local binding is transitively unused, a warning will be produced if `-Wunused-top-binds` or `-Wunused-local-binds` (repectively) is on, *and* `-freport-indirectly-unused-bindings` is on.

* "The warnings for transitively unused bindings will reference all bindings they are used in that throw a warning".  This is a significant new requirement.  Please spell it out with an example right here.  (This might be quite a bit of work; I'm not sure.)

JakobBruenker:
@simonpj thanks! I've made some changes. A couple of comments:

- I went with "indirect" over "transitive"
- I used "the scope *B* is defined in" rather than "*B*'s scope" just to make it clear it's the scope of the binding itself, not the scope of *B*'s body.
- I slightly rephrased "A binding will produce a warning if either ...", though I'm not sure if it's better. I didn't want to reference `-Wunused-local-binds` directly in that bullet, since it's more general than just local bindings, it applies to all the different kinds of bindings (e.g. `-Wunused-pattern-binds`)

simonpj:
Please, I beg, can we use `-W` flags for warnings, not `-f` flags.  I suggest a flag design in the comment immediately preceding this one

cdornan:
@JakobBruenker it looks like everyone is generally in favour of the proposal but widespread unease about using `-f` rather than extending the warnings flags. How do feel about switching from `-f` to the `-W` design?

JakobBruenker:
Sure, I have no strong opinions on the flag. For consistency with `-Wunused-binds`, I have now changed it to `-Windirectly-unused-binds`. Let me know if you prefer a different option.

@simonpj Regarding "A binding will produce a warning if...", one point not captured by your suggestion is this:

  If you have a local bind that's used only by an unused top-level bind, that's an indirectly unused bind.
  But if `-Wunused-top-binds` is disabled, I believe we should not warn about this local bind (even if both `-Wunused-local-binds` and `-Windirectly-unused-binds` are on).


However, I believe I managed to move things around such that the wording is now much better:

- I've added a definition for "relevant warning flag", so the proposal can succinctly refer not just to warnings about top or local binds, but also others (such as variables bound in pattern matches)
- I've changed the definitions for directly and indirectly unused bindings such that the relevant warning flag must be enabled for the definition to hold.

These two points together let me simplify the bullets you mentioned.


I've also incorporated your suggested third bullet into the proposal (with minor changes). Thanks!

simonpj:
Thanks.  Writing specifications is hard, isn't it?

* "1. Relevant warning flags". I believe that all of these flags exist already, right?  Please say so. Moreover, I believe that the proposal plans to introduce one or more new warning flags.  Please so so, and list them.

* "2. A binding B is directly unused if it is referenced only in B's own strongly-connected component, and the relevant warning flag is enabled.".  But what *is* the "relevant warning flag"?  Please spell it out.  Maybe "...and
  * The binding is top-level and `-Wunused-top-binds` is on, or
  * The binding is not top-level and `-Wunused-local-binds` is on."

  Or, perhaps add a numbered item defining "relevant warning flag".  It sounds as if you want that defn to cover pattern bindings, which is a mildly puzzling since the proposal only covers top-level bindings and let/where bindings, doesn't it?

  PS: just before posting.  I have just realised that you intended item 1.  to be a *definition* of "relevant warning flag".    That wasn't at all clear to me. I thought you were simply listing flags that had some relevance to the proposal!  I suggest something more like:

  **Relevant warning flag**.  The **relevant warning flag** for a variable binding is defined as follows:
    * The relevant warning flag for a top-level binding is `-Wunused-top-binds`.
    * The relevant warning flag for a non-top-level binding is `-Wunused-local-binds`.
    * The relevant warning flag for a variable bound by pattern matching is ...
    * etc
    
* "3. ... or referenced only in the body of an indirectly unused binding within the scope B is defined in". I had a really hard time understanding this phrase.  How about this:

  "... or B is referenced only in the body or a (directly or indirectly) unused binding C, *and* C is in scope at the point where B's definition appears.  For example, suppose `foo1` and `foo2` appear nowhere else.
  ```
  bar1 = True
  foo1 = bar1

  foo2 = (bar2, foo1)
       where bar2 = True
  ```
  In this example
  * `foo2` is directly unused
  * `foo1` is indirectly unused, because it occurs only in the body of the unused `foo2`, *and* `foo1` is in scope at `foo2`'s definition.
  * Similarly, `bar1` is indirectly unused.
  * But `bar2` is *not* indirectly unused becuase, while it occurs in the body of the unused `foo2`, `bar2` is not in scop e at `foo2`'s definition site.

  Notice the `(directly or indirectly)" correction too.

* I would drop point (4).  By all means add it as an example, but the first bullet is just a consequence of (2).  (Oh, note the numbering change.)  And the second bullet is a forward reference to the as-yet-unexplained stuff about error messages.

* You say "If you have a local bind that's used only by an unused top-level bind, that's an indirectly unused bind.
But if -Wunused-top-binds is disabled, I believe we should not warn about this local bind (even if both -Wunused-local-binds and -Windirectly-unused-binds are on)."  I have not fully parsed that, but since it is (a) significant to you and (b) not obvious to me, please could you add an example (under Examples) that illustrates that specific point?

* Please could you add section numbers to your proposal.  (The RST template in the repo does so; it's a one-liner.)

JakobBruenker:
@simonpj 

> Writing specifications is hard, isn't it?

Indeed it is. I appreciate your help in hashing it out though!

> I suggest something more like [\<relevant warning flag suggestion>]

Done

> How about this [\<indirectly unused definition>]

Done

> Notice the `(directly or indirectly)" correction too.

Since we switched to using strongly connected components for the definition, directly unused binds are a subset of indirectly unused binds, so this is technically redundant. Nevertheless, I have added it for clarity.

> please could you add an example (under Examples) that illustrates that specific point?

Done

> Please could you add section numbers to your proposal

Done

simonpj:
Gettting better!

* I believe that although you define the "relevant warning flag" for
  * variables bound by a pattern match, and
  * type variables bound by a type pattern

  you never use these definitions.  Say so, and say that you put them in here just for completeness.

* In item (2) you might want to add somewhere that "A **binding** B includes value bindings, but also (at top level) type and class declarations."   (In contrast, I would not refer to an import or `forall` as a "binding".)

* I looked (for the first time) at item 4.  Please split it and make it more explicit.  Just saying "with the same direct vs indirect distinction" is too vague.  I think you mean something like

  * An **imported identifier**
    * is directly unused if it is not mentioned anywhere in the module
    * is indirectly unused if it is referenced only in (directly or indirectly) unused bindings.

  * A **forall-bound type variable**
    * is directly unused if it does not appear in the body of the type
    * is never indirectly unused
    For example, in the type `forall a. Int -> Int`, the variable `a` is directly unused.

With that, I am content.  Well done!

JakobBruenker:
@simonpj 

> you never use these definitions

I don't explicitly refer to them, but I had meant for proposal to affect them.

However, thinking about it again, type pattern variables really can't be indirectly unused, so I can add that note for them.

For term-level variables bound by a pattern match, I believe it makes sense to treat them just like local bindings (and I think the proposal implicitly does that), however I see now that I didn't have any examples including them. I've changed that now, and I've added a separate point (now point 4) about those variables.

> In item (2) you might want to add somewhere that "A binding B includes value bindings, but also (at top level) type and class declarations." (In contrast, I would not refer to an import or forall as a "binding".)

Done

> A forall-bound type variable [...] is never indirectly unused

I disagree with this; taking part of the first example in the example section:

```haskell
far :: forall a (b :: a) c . c
```

Here, `b` is directly unused, whereas `a` is indirectly unused.

I've now included this example also in the definition regarding `forall`-bound type variables, along with the phrasing

> - is indirectly unused if it only appears in the kind signature of other ``forall``-bound type variables in the body of the type

JakobBruenker:
@simonpj thanks, I've made those changes.

@cdornan let me know if there's anything else I should change to get this accepted.

cdornan:
Thnks @JakobBruenker and @simonpj &mdash; I am putting it to a vote now.

proposals/0000-fine-grained-unused-warnings.rst
```diff
@@ -0,0 +1,394 @@
...
+   - is directly unused if it does not appear in the body of the type
+   - is indirectly unused if it only appears in the kind signature of other (directly or indirectly) unused ``forall``-bound type variables in the body of the type
+
+   For example:
+
+   ::
+
+     far :: forall a (b :: a) c . c
+
+    Here, ``b`` is directly unused, but ``a`` is indirectly unused.
```
maralorn, line 124: Removing a space should fix the formating.
```suggestion
   Here, ``b`` is directly unused, but ``a`` is indirectly unused.
```

proposals/0000-fine-grained-unused-warnings.rst
```diff
@@ -0,0 +1,394 @@
...
+  into two (like ``-Windirectly-unused-imports`` and ``-Wdirectly-unused-imports``) and a warnings group like ``-Wno-indirect-uses`` to turn off all warnings about indirectly unused bindings at once.
+
+  * This would offer more configurability if users want to see some warnings about indirectly unused bindings but not others.
+
+  * It would require a higher number of warning flags.
+
+Implementation Plan
+-------------------
+
+`@Jadefalke256 <https://github.com/Jadefalke256>`_ has `expressed interest <https://gitlab.haskell.org/ghc/ghc/-/issues/20190#note_505317>`_ in implementing this proposal.
```
maralorn, line 394: The link to the user 404s. I hope they just deleted their github account and are still interested in implementing this?

proposals/0000-fine-grained-unused-warnings.rst
```diff
@@ -0,0 +1,394 @@
...
+  - ``quux`` is directly unused
+
+  - ``bar`` is indirectly unused, and will produce a warning stating
+
+    ::
+
+      warning: [-Wunused-top-binds, -Windirectly-unused-binds]
+          ‘bar' is defined but used only in the following unused bindings: ‘foo’, ‘quux’
+
+- The warning for an unused binding B will reference the innermost (directly or indirectly) unused binding(s) whose right-hand sides mention B. For example, suppose ``bar`` is
```
maralorn, line 157: I think I understand what this rule wants to say, but I think it is a bit unclear. I think because this is an amendment/adding precision to the previous point without explicitly saying so.

Maybe this formulation reflects the meaning more precisely?
```suggestion
- If the warning for an unused binding B would reference multiple nested bindings it will only reference the innermost (directly or indirectly) unused binding(s) of those. For example, suppose ``bar`` is
```

maralorn (approved):
This proposal is excellent. The attention to detail here serves as an example of how the community makes Haskell great. And once this proposal is implemented it will provide real value to all our users.

I added a few small remarks.

proposals/0000-fine-grained-unused-warnings.rst
JakobBruenker: Not sure, they might have renamed their account. I was planning on asking them again once the proposal is accepted. If they don't have time or motivation anymore, I will find time to implement it.

cdornan:
This proposal has been accepted &mdash; thanks ever so much @JakobBruenker and everyone for all of your hard work.

PR #436, opened 2021-09-10
Add support for total RecordWildCards

brandon-leapyear:
This is a proposal adding `Foo{..!}` syntax that warns if a record field is unused.

[Rendered](https://github.com/brandon-leapyear/ghc-proposals/blob/patch-1/proposals/0000-recordwildcards-total.rst)

### DIFF
--- a/proposals/0000-recordwildcards-total.rst
+++ b/proposals/0000-recordwildcards-total.rst
@@ -1,0 +1,117 @@
+Notes on reStructuredText - delete this section before submitting
+==================================================================
+
+The proposals are submitted in reStructuredText format.  To get inline code, enclose text in double backticks, ``like this``.  To get block code, use a double colon and indent by at least one space
+
+::
+
+ like this
+ and
+
+ this too
+
+To get hyperlinks, use backticks, angle brackets, and an underscore `like this <http://www.haskell.org/>`_.
+
+
+Proposal title
+==============
+
+.. author:: Brandon Chinn
+.. date-accepted:: Leave blank. This will be filled in when the proposal is accepted.
+.. ticket-url:: Leave blank. This will eventually be filled with the
+                ticket URL which will track the progress of the
+                implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. contents::
+
+``-XRecordWildCards`` does not give any warning or indication if you forget to use a field. This proposal adds syntax (optionally enabled by another extension) to give a warning if a field is unused.
+
+
+Motivation
+----------
+
+Some functions require all the fields of a record to be used, or at least considered, e.g. ``hash`` or ``toJSON``, or even something more security-facing. Currently, if one uses ``RecordWildCards`` and they add a field to the record later on, nothing will tell the developer to take a look at these functions.
+
+::
+
+ data Foo = Foo { a :: Int, b :: Bool }
+ 
+ instance ToJSON Foo where
+   toJSON Foo{..} =
+     object
+       [ "a" .= a
+       , "b" .= b
+       ]
+       
+ instance Hashable Foo where
+   hash Foo{..} = hash (a, b)
+ 
+ -- adding a 'c' field does not raise any indication of needing to look at `toJSON` or `hash`
+
+Existing workarounds are described in `this blog post <https://cs-syd.eu/posts/2021-09-10-undefined-trick>`_, all of which leave something to be desired:
+
+* Positional arguments break when fields are reordered, which matters in functions like `toJSON`
+* Using `undefined` is a bit of a hack, plus it fails `-Wunused-pattern-binds`. It also fails to explicitly say which field is unused (which isn't strictly necessary, but it's nice to have)
+
+The `safe-wild-cards <https://hackage.haskell.org/package/safe-wild-cards>`_ package provides a Template Haskell solution to this, but it would be nice to support this natively.
+
+Proposed Change Specification
+-----------------------------
+
+When ``-XRecordWildCards`` is enabled, in addition to supporting ``Foo{..}`` in a pattern match, also support ``Foo{..!}``, which desugars to ``Foo{a = a, b = b}``.
+
+Examples
+--------
+
+Now, the example in the Motivation section could be written as
+
+::
+
+ instance ToJSON Foo where
+   toJSON Foo{..!} =
+     object
+       [ "a" .= a
+       , "b" .= b
+       ]
+       
+ instance Hashable Foo where
+   hash Foo{..!} = hash (a, b)
+   
+Notice the lack of any change, except for swapping ``Foo{..}`` for ``Foo{..!}``.
+
+
+Effect and Interactions
+-----------------------
+The desugared syntax would raise an ``-Wunused-matches`` warning if a new field is added but not used.
+
+As far as I know, the proposed ``Foo{..!}`` syntax doesn't conflict with any current syntax.
+
+
+Costs and Drawbacks
+-------------------
+Since it's just syntax sugar, development "shouldn't" take long. Likewise, I don't see this adding much maintenance cost.
+
+It's completely optional, so novice users can avoid this syntax completely.
+
+Primary drawback is updating tooling/editor support for the new syntax.
+
+
+Alternatives
+------------
+
+* Add an additional extension, e.g. ``-XRecordWildCardsStrict``, that implies ``-XRecordWildCards`` and enables this extension. I'm not opposed to this, but since the change doesn't break existing ``-XRecordWildCards`` syntax, I don't see a reason to break out yet another extension.
+
+Unresolved Questions
+--------------------
+
+
+Implementation Plan
+-------------------
+
+Endorsements
+-------------


proposals/0436-recordwildcards-total.rst
```diff
@@ -0,0 +1,102 @@
...
+* Using `undefined` is a bit of a hack, plus it fails `-Wunused-pattern-binds`. It also fails to explicitly say which field is unused (which isn't strictly necessary, but it's nice to have)
+
+The `safe-wild-cards <https://hackage.haskell.org/package/safe-wild-cards>`_ package provides a Template Haskell solution to this, but it would be nice to support this natively.
+
+One could even think of this as the dual of the `Or pattern proposal <https://github.com/ghc-proposals/ghc-proposals/pull/43>`_, to warn on unused terms in product types as opposed to unused terms in sum types.
+
+Proposed Change Specification
+-----------------------------
+
+When ``-XRecordWildCards`` is enabled, in addition to supporting ``Foo{..}`` in a pattern match, also support ``Foo{..!}``, which desugars to ``Foo{a = a, b = b}``.
```
michaelpj, line 51: It would be nice to also support this in normal record pattern syntax, so I could write `Foo{a=anotherNameForA, b=anotherNameForB !}` and also be warned if another field is added. That might argue for a different syntax, however, since it looks stranger in non-wildcard case.

phadej:
```
Prelude> data Foo = Foo { foo :: Int, bar :: Char} deriving Show
Prelude> :set -Wall
Prelude> Foo { foo = 1 }

<interactive>:3:1: warning: [-Wmissing-fields]
    • Fields of ‘Foo’ not initialised: bar
    • In the expression: Foo {foo = 1}
      In an equation for ‘it’: it = Foo {foo = 1}
```

proposals/0436-recordwildcards-total.rst
```diff
@@ -0,0 +1,102 @@
...
+It's completely optional, so novice users can avoid this syntax completely.
+
+Primary drawback is updating tooling/editor support for the new syntax.
+
+
+Alternatives
+------------
+
+* Add an additional extension, e.g. ``-XRecordWildCardsStrict``, that implies ``-XRecordWildCards`` and enables this extension. I'm not opposed to this, but since the change doesn't break existing ``-XRecordWildCards`` syntax, I don't see a reason to break out yet another extension.
+
```
ysangkok, line 93: Expected to see this proposal compared to [linear record selectors](https://gitlab.haskell.org/ghc/ghc/-/issues/18570). Those would seem to provide the same benefit of making it impossible to forget using a field.

proposals/0436-recordwildcards-total.rst
```diff
@@ -0,0 +1,102 @@
...
+       
+ instance Hashable Foo where
+   hash Foo{..!} = hash (a, b)
+   
+Notice the lack of any change, except for swapping ``Foo{..}`` for ``Foo{..!}``.
+
+
+Effect and Interactions
+-----------------------
+The desugared syntax would raise an ``-Wunused-matches`` warning if a new field is added but not used.
```
ysangkok, line 75: Why can `-Wunused-matches` not simply be changed to trigger for any unused fields (as this proposal assumes a new syntax is needed to do)? Because it would trigger on too much of Hackage? I'd like to see this addressed. I don't think it is so bad to expand the scope of a warning, it doesn't even break the build, unless they have opted in to that with `-Werror`. If you think the existing warning cannot be changed, why not add a new warning?

I would in fact find a warning based approach advantageous, since GHC/HLS could suggest a fix for it, which would be to add the NamedFieldPuns extension and expand the `..` to the list of fields that are currently used. With HLS, it could be a one-click deal. I don't know if HLS policy allows suggesting new extensions for syntax that is *not* currently used, that is. Though I think it would make sense.

Since NamedFieldPuns is part of GHC2021, and RecordWildCards is not, I think it would be good to nudge people in that direction.

ysangkok (changes requested):

proposals/0436-recordwildcards-total.rst
```diff
@@ -0,0 +1,102 @@
...
+
+It's completely optional, so novice users can avoid this syntax completely.
+
+Primary drawback is updating tooling/editor support for the new syntax.
+
+
+Alternatives
+------------
+
+* Add an additional extension, e.g. ``-XRecordWildCardsStrict``, that implies ``-XRecordWildCards`` and enables this extension. I'm not opposed to this, but since the change doesn't break existing ``-XRecordWildCards`` syntax, I don't see a reason to break out yet another extension.
```
ysangkok, line 92: The use of "strict" is confusing because it is already in use: e.g. "strict evaluation" vs "lazy evaluation" which is not what is meant here.

proposals/0436-recordwildcards-total.rst
brandon-leapyear: Because in most* cases, you don't intend to use all the fields in RecordWildCards, you just don't want to list out all the fields. For example, if you have a Reader monad environment where the environment has many fields (e.g. ours has 10), you might use most but not all in one function.

This proposal is specifically for the use-case of functions like `hash` or `toJSON`, where the developer would like to be strict about the fact that the function _must_ account for all the fields. But in other functions, you're just using RecordWildCards to bring all the fields into scope, whether you use it or not

proposals/0436-recordwildcards-total.rst
brandon-leapyear: I wasn't aware that was an open proposal. For next time, how would I know what proposals to address when making a proposal?

I'm also not sure what linear record selectors has to do with this. It seems to only apply to records with only one field? IIUC it would also only work if the function was linear on the record data type, but this proposal is more general than linear types and should work for non-linear functions as well. e.g.
```hs
data Foo = Foo { a :: Int, b :: Int }

foo :: Foo -> Int
foo Foo{..!} = a + a + b
```
should work also (not sure why you'd want to do this, but my point is linearity is orthogonal to this proposal)

goldfirere:
I see the motivation here -- I definitely understand why you would want this.

However, I'm not keen to introduce new syntax for this use case. While we can say that this is syntax novices don't need to know, a large codebase will include the new syntax, so people will at least need to know how to read the syntax -- which is (at present) unsearchable (what would your search terms be?) and unsuggestive. (I wouldn't associate `!` with unused-match warnings.)

Instead, how about this:
* A new warning, off by default but included in `-Wall`, called `-Wunused-record-wildcards`, which would trigger on all matches of the form `RecordCon { .. }` where there is an unused field.
* A new facility for selectively enabling warnings over part of a module. Concretely, I suggest the [modifier](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0370-modifiers.rst) syntax:

```hs
toJSON :: Rec -> ...
%(Warnings "-Wunused-record-wildcards") toJSON (RecCon { .. }) = ...
```

also writable as

```hs
toJSON :: Rec -> ...
%(Warnings "-Wunused-record-wildcards")
toJSON (RecCon { .. }) = ...
```

or maybe even as

```hs
%(Warnings "-Wunused-record-wildcards")
toJSON :: Rec -> ...
toJSON (RecCon { .. }) = ...
```

if we decide to say that a `%Warnings` that affects a type signature also affects the definition.

This is clearly verbose. But the modifiers syntax critically allows any type-level expression after the `%`. So, a project could define

```
type CompleteRecordWildcard = Warnings "-Wunused-record-wildcards"   -- or something even shorter
```

and then use `%CompleteRecordWildcard` where appropriate.

That's *still* a small mouthful, but I think that's OK. Code is read much more often than it is written, and I think it's sensible for more obscure features to have long names. This makes them easy to read for the uninitiated and easy to search for the curious.

brandon-leapyear:
I think having scoped warnings is a really good idea regardless! There've been times when I wanted to turn off warnings like `-Worphans` or `-Wname-shadowing` for a specific section, but not the entire module.

With scoped warnings, I think this is a decent approach, but I still wouldn't want to make this a warning because then `-Weverything` would warn about all uses of RecordWildCards that don't use all the fields. Practically speaking, if this were a warning, then everyone using `-Weverything` should also add `-Wno-unused-record-wildcards`.

Abstractly, warnings are good for scenarios where developers **usually** want to avoid X, and X is rare enough that manually/verbosely enabling X is a sufficient trade-off. This proposal is far from the common RecordWildCards use-case (https://github.com/ghc-proposals/ghc-proposals/pull/436#discussion_r712349284), so a warning doesn't seem appropriate here.

> However, I'm not keen to introduce new syntax for this use case. While we can say that this is syntax novices don't need to know, a large codebase will include the new syntax, so people will at least need to know how to read the syntax -- which is (at present) unsearchable (what would your search terms be?) and unsuggestive. (I wouldn't associate ! with unused-match warnings.)

That is a really good point, but `RecordWildCards` today falls under the same issue, right?

If the modifier proposal goes through, what if we just added a separate modifier for this?
```hs
foo Foo{ %useAllRecordFields .. } = undefined
```

proposals/0436-recordwildcards-total.rst
aspiwack: I strongly agree with this sentiment.

By the way, for comparison, in Ocaml, record matching must be exhaustive by default. The following is an error:

```ocaml
match { a; b; c } with
| { a; b } -> …
```

If you want to drop some fields, you need special syntax

```ocaml
match { a; b; c } with
| { a; b; .. } -> …
```

(here, contrary to Haskell's syntax, the `..` do not bind anything. It just means “ignore the remaining fields”)

neongreen: > for comparison, in Ocaml, record matching must be exhaustive by default

Same in Rust. I prefer Rust's version to Haskell's (if you want to use many fields, just write `x.a` `x.b` `x.c` and then it will be clearer where the values are coming from), but having `{..!}` is definitely better than nothing.

jberryman:
I think @goldfirere 's alternative here would be totally unusable (I would simply put up with the status quo rather than add all that noise) and I like the original proposal. Documentation can be added to the RecordWildcard docs; I can't think of hardly any ghc extensions that meet this standard for searchability

Can I propose this syntax though?:

```
Foo{...}
```

This:
- doesn't mess with `!` which look like bang patterns
- is similar to JS's `===` strict equality

I support adding this to the existing extension, rather than creating a new one

JakobBruenker:
> Can I propose this syntax though?:
>```
> Foo{...}
>```

This would have interesting consequences for the discussion on https://github.com/ghc-proposals/ghc-proposals/pull/477

konsumlamm:
> Can I propose this syntax though?:
> 
> ```
> Foo{...}
> ```

Please no. This would only confuse people what the difference between `..` and `...` is.

> This:
> 
>   * is suggestive of JS's `===` strict equality

Is that really a good thing though? Afaik, `==` in JS is widely regarded as being a mistake and you should always use `===`.

jberryman:
On second thought, maybe this proposal could be expanded to any patterns involving curly braces (record patterns and NamedFieldPuns, for example). A possible syntax might be double braces (similar logic to `$$` for types splices?) like

``` hs
Foo{{ a, b, c }}  -- if `d` is added a warning is raised
```

proposals/0436-recordwildcards-total.rst
```diff
@@ -0,0 +1,125 @@
...
+* Using `undefined` is a bit of a hack, plus it fails `-Wunused-pattern-binds`. It also fails to explicitly say which field is unused (which isn't strictly necessary, but it's nice to have)
+
+The `safe-wild-cards <https://hackage.haskell.org/package/safe-wild-cards>`_ package provides a Template Haskell solution to this, but it would be nice to support this natively.
+
+One could even think of this as the dual of the `Or pattern proposal <https://github.com/ghc-proposals/ghc-proposals/pull/43>`_, to warn on unused terms in product types as opposed to unused terms in sum types.
+
+Proposed Change Specification
+-----------------------------
+
+When ``-XRecordWildCards`` is enabled, in addition to supporting ``Foo{..}`` in a pattern match, also support ``Foo{..!}``, which desugars to ``Foo{a = a, b = b}``.
```
LightAndLight, line 51: I agree with the motivation for this proposal (came here to check whether such a proposal existed), disagree with the proposed change.

Currently the `!` token in Haskell means that `seq` is involved, but the difference between `Foo{..}` and `Foo{..!}` is not a presence / absence of `seq`. So at the very least, I think the proposed `!` should be ruled out. More generally, though, I don't like using a new syntactic form to trigger the checks we need.

Here are some alternatives:

* Change [`-Wunused-record-wildcards`](https://downloads.haskell.org/ghc/latest/docs/users_guide/using-warnings.html#ghc-flag-Wunused-record-wildcards)

  Currently the flag warns when *none* of the fields in the pattern `Foo{..}` are used. This was the first flag I tried when I wanted completeness warnings for records, and I was surprised by its behaviour. I expected it to tell warn me for each field that was unused.

* Add `-Wunused-record-wildcard-fields`

  A backwards-compatible version of the previous point. While `-Wunused-record-wildcards` tells you that an entire `Foo{..}` is redundant, `-Wunused-record-wildcard-fields` will tell you each field that is unused.

* Add `-Wincomplete-record-patterns`

  I don't *need* a RecordWildCards-based solution. `-Wincomplete-record-patterns` will warn when a record pattern doesn't name all the fields associated with the constructor. Using the datatype from the proposal, `Foo{a = x}` would generate a "field `b` unused" warning, and so would `Foo{a}` under NamedFieldPuns. If RecordWildCards can actually be specified in terms of NamedFieldPuns (i.e. `Foo{..}` desugars to `Foo{a, b}` desugars to `Foo{a = a, b = b}`), then `-Wincomplete-record-patterns` should also give the warnings described in the previous two points.

guibou:
I've just opened https://gitlab.haskell.org/ghc/ghc/-/issues/25916 while unaware of this proposal.

Here are my few comments / opinions about this proposal and the current state of the discussion:

I really like the idea of a new warning. I agree here with @LightAndLight that the discussion should not be limited to WildCards, but also include plain old field matching, with or without `NamedFieldPuns`.

A dev aware of the existence of this new warning (not activated by default) can decide to turn it on on their codebase. True positive are bug to be fixed, and false positive do have easy workaround (e.g. `f Foo{a, b} = a + b` can be turned to `f foo = foo.a + foo.b`). I actually kinda like the idea that "pattern match" are exhaustive and record field access via selector / dot operator are limited to what's selected.

I'm not fan of a new syntax, it adds complexity in the discussion about the syntax (for the record, I feel confused about the `!` proposal, for its similarity with bang patterns) and it requires changes in parser and the different ASTs (HsExpr and template haskell). The implementation cost and the impact on the ecosystem is too high compared to the benefit. Note also that a library built with this flag AND `-Werror` may break because of a TH splice without any other solution than to disable the warning for the module using the TH splice. And the library could only be fixed with `CPP` code in order to stay backward compatible.

Unless I misunderstood how the proposal works, if we only add a warning without change of the Haskell language and without adding this warning by default or in `-Wall`, this can be done without the requirement of an accepted proposal?

adamgundry:
Thanks @guibou. I agree with you that any design that merely adds a warning will be rather more convincing than adding new syntax.

> Unless I misunderstood how the proposal works, if we only add a warning without change of the Haskell language and without adding this warning by default or in `-Wall`, this can be done without the requirement of an accepted proposal?

A proposal would be appropriate this case, because the design of the warning needs to be motivated and given a clear specification, for which the proposals process provides a framework. While off-by-default warnings do not have a huge impact on users unless they opt in, they still come with a maintenance cost and constrain future development, so it's important to get the design right.

PR #442, opened 2021-10-18
Codify proposals in an "Apsirational User's Guide"

Ericson2314:
This is a meta-proposal to try out an extension to the proposal process on an experimental basis.

We currently propose changes to an *implicit* state, the current design that is a roll-up of GHC and all unimplemented, accepted proposals. Instead, we should make that state *explicit* by codifying excepted proposals in a design document, which is to be copy of the GHC User's Guide known as the "Aspirational User's Guide". This will catch subtle design interactions earlier than implementation time, and make it easier to evaluate new proposals against an accurate baseline anyone can read.

[Rendered](https://github.com/Ericson2314/ghc-proposals/blob/aspirational-users-guide/proposals/0000-aspirational-users-guide.rst)

### DIFF
--- a/proposals/0000-aspirational-users-guide.rst
+++ b/proposals/0000-aspirational-users-guide.rst
@@ -15,45 +15,57 @@ Codify proposals in an "Apsirational User's Guide"
 .. contents::
 
 This is a meta-proposal to try out an extension to the proposal process on an experimental basis.
+
 We currently propose changes to an *implicit* state, the current design that is a roll-up of GHC and all unimplemented, accepted proposals.
-Instead, we should make that state *explicit* by codifying excepted proposals in a design document, which is a copy of the GHC User's Guide known as the "Aspirational User's Guide".
-This will catch subtle design interactions earlier than implementation design, and make it easier to evaluate new proposals against a known-to-all baseline.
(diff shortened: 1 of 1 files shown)


treeowl:
This sounds like a great idea. I've always found the process of writing documentation valuable for design. If you can't explain it for potential users, then the design is probably wrong. I haven't read this proposal yet, but I hope that a  PR/MR to the aspirational user's guide will become part of the proposal acceptance process.

nomeata:
> Since proposers can't just present a manual-diff as their proposed change specification

Why not… ah, you allude to that in section 3. I am reasonable optimistic that we can do the necessary `git` magic _somehow_, and would suggest we discuss this idea first under that optimistic assumption. _If_ we think we want it, we can sort out these technicalities.

Ericson2314:
The number of amendments we have (and there are more on the way!) might be taken as evidence we are already moving towards this :)

20230725_13h05m42s_grim

Ericson2314:
In #608, I just got confused and tripped up by the interactions between #448, #281, and #607; and also had trouble explaining the interaction between #448 and draft #270 (a similar thing; what I had forgotten is the salient parts of #270 were since in #281 and #607 so there was no point appealing to draft #270).

A single document describing how renaming works with all accepted proposals would helped me out immensely!

PR #445, opened 2021-10-21
Non-overlappable instances

treeowl:
Allow instances to be marked `NON_OVERLAPPABLE` so no one can
defeat code relying on their behavior.

[Rendered](https://github.com/treeowl/ghc-proposals/blob/non-overlappable/proposals/0000-non-overlappable.md)

### DIFF
--- a/proposals/0000-non-overlappable.md
+++ b/proposals/0000-non-overlappable.md
@@ -1,0 +1,192 @@
+---
+author: David Feuer
+date-accepted: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/445).
+
+# Non-overlappable instances
+
+Sometimes it is useful to ensure that a particular class instance
+is the absolute final word, and will never be overlapped. I propose
+to add a `NON_OVERLAPPABLE` pragma to enforce this.
+
+
+## Motivation
+
+John Ericson [suggested](https://github.com/haskell/containers/issues/706)
+a rather clever function for `Data.Map`, based on an idea by Dan Bornside.
+
+```haskell
+splitEither
+  :: Map (Either b c) v
+  -> (Map (Either b Void) v, Map (Either Void c) v)
+splitEither = unsafeCoerce . spanAntitone isLeft
+```
+
+Unlike a more conventional approach to splitting such a map, this accomplishes
+the split in `O(log n)` time, rather than `O(n)` time. It relies on the fact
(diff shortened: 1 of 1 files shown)


treeowl:
@Ericson2314 and @danbornside might be interested.

treeowl:
I think the fact that you haven't come up with an actual reason someone would want an overlapping `Ord` instance for `Either a b` is a direct result of the fact that there *is* no good reason for that.

nomeata:
Right, Overlapping instances is often a fishy thing to. That’s why you have to slap a verbose pragma (`{-# OVERLAPPING #-}` on it, and that’s why it is not part of Safe Haskell (I assume). Isn’t such a pragma a high enough hurdle for “I know what I am doing”?

If we’d introduce `NON_OVERLAPPABLE` in four years we might see someone asking for a `OVERLAPPING_EVEN_NON_OVERLAPPABLE` pragma…

Haskell is full of unsafe things, if you explicitly ask for it. `OVERLAPPING` is such an unsafe thing… I don't see why we need to make it harder.

As for breaking `Map` invariants – that’s already possible without `OVERLAPPING`, simply by providing a non-lawful instance. So “`splitEither` works, unless people write fishy overlapping instances” is probably the smaller problem than “`Data.Map` works, unless people write fishy non-lawful instances”.

treeowl:
@nomeata there's a significant escalation once someone writes code that depends on the invariant for safety, I think. As I see it, this plugs a gap that's existed since the local pragmas were created: and/or is very permissive!

treeowl:
@adamgundry, these things get tricky for sure, especially with GADTs, MPTCs, etc. To answer your sample question, trying to import both `B` and `C` and searching for an instance there (or beyond) would be an error. I would prefer if even the imports were an error, but I don't think GHC knows how to do that.

adamgundry:
> @adamgundry, these things get tricky for sure, especially with GADTs, MPTCs, etc. To answer your sample question, trying to import both `B` and `C` and searching for an instance there (or beyond) would be an error. I would prefer if even the imports were an error, but I don't think GHC knows how to do that.

I can see that GHC could report an error if it does end up trying to solve a relevant constraint in the module that imports both a `NON_OVERLAPPABLE` instance and an instance that actually overlaps it. I'm not sure that is enough for safety, though? Or at least it considerably weakens the guarantee that `NON_OVERLAPPABLE` is supposed to give.

Presumably GHC has some mechanism to ensure that the same type family cannot be given conflicting instances in different modules, so I guess something similar could be used to detect the conflicting instances. I'm a bit worried that doing so might be bad for compile-time performance though.

PR #446, opened 2021-10-22
Standalone deriving clauses grouped by strategy

treeowl:
Add support for multiple standalone `deriving` clauses for
each deriving strategy. This can considerably reduce boilerplate
for `DerivingVia`.

[Rendered](https://github.com/treeowl/ghc-proposals/blob/multi-standalone/proposals/0000-multiple-standalone-deriving.md)

### DIFF
--- a/proposals/0000-multiple-standalone-deriving.md
+++ b/proposals/0000-multiple-standalone-deriving.md
@@ -36,7 +36,7 @@ deriving via Generically (Foo a)
 ## Motivation
 
 Being able to write a deriving strategy once and use it multiple times
-can save a considerable amount of typing, especially when types get long
+can save a considerable amount of typing, especially when types get long.
 For example, one might wish to do something like
 
 ```haskell


aspiwack:
This is something that I've been wishing for often in the past couple of years. So I'm supportive.

I'm not sure that the bit about being instantiated at different kinds in a block is particularly justified. Though if it's easy to do, why not.

treeowl:
@aspiwack, it's more convenient to allow instantiations at multiple kinds. Is there any reason not to?

phadej:
@treeowl 

> @aspiwack, it's more convenient to allow instantiations at multiple kinds. Is there any reason not to?

If we take just two first instances from `Roger` example:

```haskell
deriving via Target Roger
  instance C Roger -- via Target @k Roger
  instance D Roger - via Target @Type Roger
```

If we require the same kinds, k` and `Type` would unify and we get

```haskell
deriving via Target @Type Roger instance C Roger -- not @k
deriving via Target @Type Roger instance D Roger
```

instances as a result.

It's possible that this is welcome behavior sometimes,
but it feels it would surprise many `PolyKinds` practioners.

aspiwack:
Maybe. I feel that having different kind instantiations is also kind of unexpected. Maybe I'm wrong. I personally try to be as explicit as possible about kind polymorphism.

RyanGlScott:
I'd also favor the different kind instantiation behavior as laid out in the proposal, as this is consistent with how `deriving` clauses work. If you write this:

```hs
data Roger (a :: k) = Roger
  deriving (C, D, E) via Target Roger
```

Then the derived instances will be via `Target @k Roger`, `Target @Type Roger`, and `Target @Bool Roger`.

i-am-tom:
Is this proposal dormant? Having just had a look through some of the open proposals, I'd echo @aspiwack's comment and say that I've also wanted this several times in the past and present 😄

PR #451, opened 2021-10-28
Sized literals (under review)

hsyl20:
Add support for sized primitive literals (``Int8#``, ``Word16#``...).

[Rendered](https://github.com/hsyl20/ghc-proposals/blob/sized-literals/proposals/0000-sized-literals.rst)

### DIFF
--- a/proposals/0000-sized-literals.rst
+++ b/proposals/0000-sized-literals.rst
@@ -1,0 +1,168 @@
+Sized primitive literals
+========================
+
+.. author:: Sylvain Henry
+.. date-accepted:: Leave blank. This will be filled in when the proposal is accepted.
+.. ticket-url:: Leave blank. This will eventually be filled with the
+                ticket URL which will track the progress of the
+                implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/451>`_.
+.. contents::
+
+Add syntax for sized primitive literals (Int8#, Word32#, etc.).
+
+
+Motivation
+----------
+
+GHC provides sized primitive types -- ``Int[8,16,32,64]#`` and
+``Word[8,16,32,64]#`` -- and their operations as primops. However it doesn't
+support *literals* for these types in Haskell syntax.
+
+It does however support primitive literals for ``Int#`` and ``Word#`` via the
+``MagicHash`` extension: ``123#`` is an ``Int#`` literal and ``123##`` is a
+``Word#`` literal.
+
+Internally in Core literals for sized primitives are supported but the only way
+to use them is to rely on constant folding. For example, the following code:
+
+::
+
+  import GHC.Word
+  import GHC.Exts
+  
+  -- from Int# and Word# literals
+  foo (# #) = intToInt64#   123#
+  bar (# #) = wordToWord32# 123##
+  -- from Integer literals
+  baz (# #) = let unW16 (W16# w) = w in unW16 (fromInteger 123)
+
+is compiled into the following Core:
+
+::
+
+  foo = \ _ -> 123#64
+  
+  bar = \ _ -> 123##32
+  
+  baz = \ _ -> 123##16
+
+Notice the literal suffixes that aren't allowed with Haskell syntax. This
+proposal is about adding support for these suffixes (not necessarily in this
+form) to Haskell.
+
+Adding support for these literals would make some codes easier to write. In
+particular, allowing case expression on sized literals.
+
+
+Proposed Change Specification
+-----------------------------
+
+We propose to borrow Rust's syntax for sized literals. A new extension (e.g.
+"SizedLiterals") allows the following forms:
+
+::
+
+  123i   -- Int# literal
+  123i8  -- Int8# literal
+  123i16 -- Int16# literal
+  123i32 -- Int32# literal
+  123i64 -- Int64# literal
+
+  123u   -- Word# literal
+  123u8  -- Word8# literal
+  123u16 -- Word16# literal
+  123u32 -- Word32# literal
+  123u64 -- Word64# literal
+
+The lexer has new lexemes for these literals.
+
+The lexer for primitive ``Int#`` and ``Word#`` has to take into account two
+extensions (``NegativeLiterals`` and ``BinaryLiterals``) in addition to
+``MagicHash``. It is currently defined as follows:
+
+::
+
+  -- Unboxed ints (:: Int#) and words (:: Word#)
+  -- It's simpler (and faster?) to give separate cases to the negatives,
+  -- especially considering octal/hexadecimal prefixes.
+  @decimal                          \# / { ifExtension MagicHashBit }        { tok_primint positive 0 1 decimal }
+  0[bB] @numspc @binary             \# / { ifExtension MagicHashBit `alexAndPred`
+                                           ifExtension BinaryLiteralsBit }   { tok_primint positive 2 3 binary }
+  0[oO] @numspc @octal              \# / { ifExtension MagicHashBit }        { tok_primint positive 2 3 octal }
+  0[xX] @numspc @hexadecimal        \# / { ifExtension MagicHashBit }        { tok_primint positive 2 3 hexadecimal }
+  @negative @decimal                \# / { negHashLitPred }                  { tok_primint negative 1 2 decimal }
+  @negative 0[bB] @numspc @binary   \# / { negHashLitPred `alexAndPred`
+                                           ifExtension BinaryLiteralsBit }   { tok_primint negative 3 4 binary }
+  @negative 0[oO] @numspc @octal    \# / { negHashLitPred }                  { tok_primint negative 3 4 octal }
+  @negative 0[xX] @numspc @hexadecimal \#
+                                       / { negHashLitPred }                  { tok_primint negative 3 4 hexadecimal }
+
+  @decimal                       \# \# / { ifExtension MagicHashBit }        { tok_primword 0 2 decimal }
+  0[bB] @numspc @binary          \# \# / { ifExtension MagicHashBit `alexAndPred`
+                                           ifExtension BinaryLiteralsBit }   { tok_primword 2 4 binary }
+  0[oO] @numspc @octal           \# \# / { ifExtension MagicHashBit }        { tok_primword 2 4 octal }
+  0[xX] @numspc @hexadecimal     \# \# / { ifExtension MagicHashBit }        { tok_primword 2 4 hexadecimal }
+
+We propose to extend it as follows for sized literals. We only show the
+``Int8#`` case to avoid cluttering this proposal: other cases are very similar.
+The call to ``tok_primint8`` returns a lexeme corresponding to ``Int8#``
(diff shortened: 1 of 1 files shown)


hsyl20:
Thanks for the feedback so far. I'll probably submit to the committee next week if no one objects.

hsyl20:
I'd like to submit this to the committee @nomeata.

nomeata:
Thanks! Pinging @goldfirere who said he can fill in for me while I am traveling.

goldfirere:
I have assigned @simonmar as the shepherd here.

noughtmare:
I may be a little late to the party, but I would actually like to keep the option open to have this syntax for boxed numbers. Normally, for boxed numbers the actual size of the literals does not matter so much, but sometimes you need to disambiguate. Also, disambiguation makes it clearer what the intent of the code is in general. So I would prefer keeping the unboxed hash symbol somewhere in the literal, e.g. `123#i8`. I also think this makes it clearer that the numbers are indeed unboxed.

konsumlamm:
> I may be a little late to the party, but I would actually like to keep the option open to have this syntax for boxed numbers. Normally, for boxed numbers the actual size of the literals does not matter so much, but sometimes you need to disambiguate. Also, disambiguation makes it clearer what the intent of the code is in general. So I would prefer keeping the unboxed hash symbol somewhere in the literal, e.g. `123#i8`. I also think this makes it clearer that the numbers are indeed unboxed.

I agree, I find it very confusing that `123i8` would be an `Int8#` and not an `Int8` (as I would expect based on the syntax).

hsyl20:
I've updated the proposal to reintroduce the hash. The proposal is under review but the change is trivial, so I hope it's ok. I've also mentioned the current proposal in the alternatives and the more conservative "double-hash" syntax too.

simonmar:
(shepherd recommendation)

I tentatively recommend that we accept the proposal. It fills a gap, and it's uncontroversial.

The syntax is debatable. The terminology `i8`, `u64` etc. doesn't appear anywhere else in Haskell or GHC. It does appear in Rust, and Thrift uses `i64`, but I would probably have chosen `123#8` and `123##64 or perhaps `123#Int8` and `123#Word64` which leaves room for `123#Char`.

nomeata:
The syntax `123#Int8` will soon create a desire for a generalization to arbitrary types after the `#`, and a library-accessible overloading mechanism, I expect. Not sure if that’s good or bad.

(I wonder if there is a place for module-qualitifed syntax overloading here, akin to qualified-do; e.g. `Int8.123`. But that would require separate modules for each of the types to exist. And looks odd.)

phadej:
https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0370-modifiers.rst

```haskell
%Int8# 123
```

int-index:
At this point might as well give a special treatment to `123 :: Int8#`

noughtmare:
> At this point might as well give a special treatment to 123 :: Int8#

Wouldn't that basically be #30?

goldfirere:
I would rather have a representation-polymorphic `Num` class (#30). I've experimented with such a thing in the past, and it works well. I suppose we could have a representation-polymorphic `Num#` class instead of `Num`, and have e.g. `123# :: Num# a => a` just to avoid messing with good ol' `Num`, but my first stop in the implementation would be to try to use `Num` to see if any problem arises.

Alternatively, we could do both: have these literals *and* have representation-polymorphic `Num` or `Num#`. But that would beg the question of whether we should also have non-overloaded literals for boxed numbers.

I'm not really against this -- there's not much downside to adopting this proposal -- but I don't see a need for an inconsistency between boxed literals and unboxed literals in this way. All else being equal, uniformity is nice.

phadej:
Going through a type-class is a problem this proposal trying to solve, doesn't it. It is valuable to mention e.g. TH Lift instances: it would be best if we are able to generate static code, without any redexes. That could be compiled into static data without relying on a simplifier. (Luckily *there are* size literals in `Core`, and there should be for all primitive types, like `ByteArray#`).

One way to think about proposals like these is in two parts:
1. How are we going to modify [template-haskell Lit](https://hackage.haskell.org/package/template-haskell-2.18.0.0/docs/Language-Haskell-TH-Syntax.html#t:Lit) and [GHC's HsLit](https://hackage.haskell.org/package/ghc-9.2.1/docs/Language-Haskell-Syntax-Lit.html#t:HsLit): for example adding `Int8L`, `Int16L`, ... constructors.
2. What is a concrete surface syntax for these literals. Without a concrete syntax, what would pretty-printer (template-haskell or HsExpr's one) produce.

IMHO getting 1. without 2. is already an improvement, especially if people cannot agree on a syntax, and that's the only missing piece (e.g. `ByteArray#` literals, no syntax, nothing goes forward). Then template-haskell prettyprinter would need to print some garbage (or some function application form), which is unfortunate, but still better then not having anything.

IMHO, even representation-polymorphic `Num` is a non-solution.

cc @treeowl who recently was surprised there is no literals for `ByteArray#`. I hope this proposal will help lead the way for `ByteArray#` literals (and "constructor" forms for `SmallArray#`, `Array#` etc).

goldfirere:
I don't quite understand why the argument above leads to the conclusion that overloaded unboxed literals are a non-solution. Maybe this will get at the nub of the argument: do we also need non-overloaded literals for `Int` and `Integer`? If not -- and you still say we do need such literals for `Int32#` and `Word64#` -- why the difference?

phadej:
@goldfirere they are good addition, but not sufficient IMO.

> do we also need non-overloaded literals for Int and Integer

No, they are not primitive:

```haskell
data Int = I# Int#
data Integer = IS !Int# | IP !BigNat# | IN !BigNat#
```

But we need `ByteArray#` literals eventually, as

```haskell
type BigNat# = WordArray#
type WordArray# = ByteArray#
```

goldfirere:
That makes sense. So your argument is:
- All types need to have computation-free ways of construction.
- For non-primitive types, this is via constructors.
- For primitive types, this must be via literals.

Put another way: Even if we had `Num#`, we would still need these literals in Core, to allow `fromInteger#` (or whatever) to inline (and avoid computation). If we have them in Core, then we should probably expose them in Haskell.

OK. I'm convinced now. I vote in support.

int-index:
> All types need to have computation-free ways of construction.
> For non-primitive types, this is via constructors.
> For primitive types, this must be via literals.

Yes, that was the intent behind my `123 :: Int8#` suggestion. It is not desugared into `fromInteger 123`, it is a compound syntactic form for an `Int8#` literal.

Ericson2314:
I do think the "ugliness" of the special syntax makes clear this is a bit esoteric, kind of like primops vs regular functions wrappers.

Someday, I hope can do `123 :: Int8#` with a typeclass but a method that is guaranteed to reduce at compile time, perhaps sharing an underlying mechanism with runtime-rep polymorphism which has similar needs. That, being extensible and yet having a performance guarantee, is better, and what I think we should we should save the "good" syntax like `123 :: Int8#` for.

simonpj:
I'm content with the proposal.  

* We need these literals in Core; and we need to pretty print them if only for deub purposes
* We could possibly get away without them in Haskell; as indeed we do today
* Experience shows that anything we can do in Core we will, in the end, want to expose in Haskell
* Case expressions are a good example.  There really isn't a good way to case-analysise a `Int8#` today.

It's not a huge deal.  But yes, I support.

jvanbruegge:
I dislike 2 because the singed/unsinged difference is too small. I strongly dislike 4 because if it looks like a type signature it should be one, not some parser magic.

Either 1 or 3 is fine, with a slight preference for 1 because i think that word is a bad name for unsigned ints.

Ericson2314:
3 I think should be `123#Word64#` etc. even though we aren't generalizing, just so it's really clear to anyone that stumbles across the syntax.

simonmar:
> 3 I think should be `123#Word64#` etc. even though we aren't generalizing, just so it's really clear to anyone that stumbles across the syntax.

Mumble. I see your point, but I'm not sure it's too hard to grasp that `<literal>#<type>` is syntax for a literal of type `<type>#`. And we'll never be using this syntax for non-primitive types, so the type would always have a `#` suffix.

nomeata:
> And we'll never be using this syntax for non-primitive types, so the type would always have a # suffix.

Famous last works?

doyougnu:
Ping.

 Looks like (3) was the winner with only 3 votes total? Is anything else blocking this proposal?

konsumlamm:
I actually like 2, mostly because it is consistent with the `Int#`/`Word#` literals (`123#`/`123##`) and that's what core currently uses, although I guess these could be changed too? I don't think we should introduce yet another syntax for unboxed integer literals, unless we deprecate the old one (or at least provide an alternative, like `123#i`/`123#u` or `123#Int`/`123#Word`).

I wouldn't say that 3 is the winner, there haven't been enough votes. A strawpoll would probably be better here.

Boarders:
Given that `##` is a pretty meaningless convention for unsigned literals / types I don't think it should be pushed into the future personally.

doyougnu:
> Syntax like 123#u16 would probably be very clear to a C or Rust developer and I don't see the reason why haskell needs to waste novelty on its own bespoke convention on this issue.

This is my feeling as well. Especially because Haskell and Rust have a lot of overlapping mind share, I think using this syntax is a good thing precisely because it has meaning to these developers.

doyougnu:
@hsyl20 @Boarders @konsumlamm @nomeata @simonmar @simonpj @Ericson2314 @int-index @jvanbruegge @goldfirere @noughtmare @phadej 

> A strawpoll would probably be better here.

Agreed. Please fill out this [Poll](https://strawpoll.vote/polls/mjxftfjm/vote?s=0). It uses ranked choice voting and results should be public.

nomeata:
Neat. But I'd use the ranked pairs method of determining the winner, as

> Ranked Pairs is special among voting systems in that it satisfies the Condorcet Criterion.

doyougnu:
Ping. After a week of voting we have the following (using ranked pairs):

```
Winner: Hash syntax extension, expecting generalizaiton: 123#Int8, 123#Word64
Runner up: Mimicking C/Rust syntax: 123#i8, 123#u64
Third place: Hash syntax extension 123#8, 123##64 
Fourth place: Type mimickry: 123 :: Int8# 123 :: Word64#
```
With 21 votes.

Note that it was possible to double vote in the poll, I assumed everyone would vote in good faith.

Is this sufficient to move this proposal forward? If not, what are the blockers?

simonmar:
@hsyl20 as the proposer, are you happy with this? If so, would you mind updating the proposal? Thanks!

hsyl20:
Yes, done!

simonmar:
Great, I think we can declare this proposal accepted.

Kleidukos:
@hsyl20 what is the ticket / MR in the GHC Gitlab for this GHC Proposal? :)

hsyl20:
@Kleidukos It has been added to the proposal: https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0451-sized-literals.rst

https://gitlab.haskell.org/ghc/ghc/-/issues/21422

ulysses4ever:
@Kleidukos by any chance, do you feel like working on this?

raehik:
I've picked this up at: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/8307

First time contributor, so I'd gratefully receive all help and feedback!

PR #448, opened 2021-10-28
Modern Scoped Type Variables (under review)

goldfirere:
This is the comprehensive proposal to update GHC's treatment of scoped type variables, obviating #126, #155, #285, #291, #238, and #420.

[Rendered](https://github.com/goldfirere/ghc-proposals/blob/type-variables/proposals/0448-type-variable-scoping.rst)
[Rendered principles](https://github.com/goldfirere/ghc-proposals/blob/type-variables/principles.rst)

PR #452, opened 2021-10-29
Articulate principles for GHC evolution

goldfirere:
This PR is an attempt to create a central place in the proposals repo for *principles*. We can then use these principles to guide our responses to individual proposals.

The principles written in the current document are just taken from #378. I do not wish to debate these principles in this PR -- that happened in #378. Instead, let's use this PR to debate whether articulating these principles centrally is a good idea, and whether the document in the form here is the best way to do so.

Concrete motivation: once this PR is accepted, then I will rephrase #448 to update it with all the principles in that proposal. Accepting #448 will thus change our set of principles.

[Rendered](https://github.com/goldfirere/ghc-proposals/blob/principles/principles.rst)

### DIFF
--- a/principles.rst
+++ b/principles.rst
@@ -24,9 +24,11 @@ a change to GHC.
 
 We urge proposers to resist the temptation to propose principles without an
 accompanying concrete change to GHC. Debating principles in the abstract does
-not seem a productive use of time, though it is possible to imagine exceptions. 
+not seem a productive use of time, though it is possible to imagine exceptions.
 For example, the justification for the rejection of a proposal may be distilled into
-a new principle here.
+a new principle here. Accordingly, this document is not (and presumably never will be)
+comprehensive: it contains the articulated prinicples guiding GHC's development
+but lacks the unarticulated principles, which will be added over time.
 
 All principles in this document are linked back to the PR that introduce them,
 so that readers can learn the context in which they were written.
--- a/principles.rst
+++ b/principles.rst
@@ -23,7 +23,9 @@ a change to GHC.
 
 We urge proposers to resist the temptation to propose principles without an
 accompanying concrete change to GHC. Debating principles in the abstract does
-not seem a productive use of time, though it is possible to imagine exceptions.
+not seem a productive use of time, though it is possible to imagine exceptions. 
+For example, the justification for the rejection of a proposal may be distilled into
+a new principle here.
 
 All principles in this document are linked back to the PR that introduce them,
 so that readers can learn the context in which they were written.
--- a/principles.rst
+++ b/principles.rst
@@ -80,5 +80,5 @@ not be affected by it.
 
 This principle is violated in various ways today: it is easy for GHC to generate error messages that refer to
 advanced features even when writing simple code. In addition, the existence of advanced features likely slow
-down GHC even when those features are not active. Yet this principle is an important to keep in mind going forward,
+down GHC even when those features are not active. Yet this principle is important to keep in mind going forward,
 as we hope not to make the current situation worse.
--- a/principles.rst
+++ b/principles.rst
@@ -1,0 +1,82 @@
+Principles for GHC
+==================
+
+This document lays out high-level principles for the evolution of GHC
+and the language it compiles. In much the way that we like to write
+*properties* of our functions -- instead of just writing down individual
+test cases -- this document strives to write *principles* of our proposals.
+
+Articulating these principles helps to guide future proposals: proposals
+that maintain the principles in this document are more likely to be accepted,
+while proposals that work against these principles are more likely to be rejected.
+Note that neither direction is (at all) a guarantee: there will be excpetions
+in both directions. Yet by writing down the principles, we can have an informed
+discussion of the tradeoffs of accepted or rejecting an individual proposal.
+
+How to update these principles
+------------------------------
+
+When making a proposal, following the `usual guidelines <https://github.com/ghc-proposals/ghc-proposals/#how-to-start-a-new-proposal>`_,
+feel free to include a diff against this listing of principles. We can then
+discuss the validity of the new principle(s) alongside the concrete proposal for
+a change to GHC.
+
+We urge proposers to resist the temptation to propose principles without an
+accompanying concrete change to GHC. Debating principles in the abstract does
+not seem a productive use of time, though it is possible to imagine exceptions.
+
+All principles in this document are linked back to the PR that introduce them,
+so that readers can learn the context in which they were written.
+
+Accepted principles
+-------------------
+
+.. _`#378`: https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0378-dependent-type-design.rst
+
+Syntax
+~~~~~~
+
+.. _SUP:
+
+**Syntactic Unification Principle (SUP).** (From `#378`_.) In the absence of punning, there is
+no difference between type-syntax and term-syntax.
+
+The SUP_ is a *long term* goal; today, there are many, many violations of this principle. (For example,
+``case`` can never appear in types, and ``Int -> Bool`` is meaningless in a term.) However, this principle
+should direct future proposals not to make the situation worse, and proposals that bring us closer to
+the SUP_ are to be valued.
+
+Name resolution
+~~~~~~~~~~~~~~~
+
+.. _LSP:
+
+**Lexical Scoping Principle (LSP)**. (From `#378`_.) For every *occurrence* of an
+identifier, it is possible to uniquely identify its *binding site*, without
+involving the type system.
+
+The LSP_ is true today, though Template Haskell splices may need to be run before
+completing name resolution (and running those splices requires type-checking them).
+
+Semantics
+~~~~~~~~~
+
+.. _PEP:
+
+**Predictable Erasure Principle (PEP)**. (From `#378`_.) The programmer knows, for sure, which bits of the program will be
+retained at runtime, and which will be erased.
+
+The PEP_ is true today: types are erased, while terms are retained.
+
+User experience
+~~~~~~~~~~~~~~~
+
+.. _OIP:
+
+**The Opt-In Principle (OIP):** (From `#378`_, slightly generalized.) Users who do not opt into an advanced feature will
+not be affected by it.
+
+This principle is violated in various ways today: it is easy for GHC to generate error messages that refer to
+advanced features even when writing simple code. In addition, the existence of advanced features likely slow
+down GHC even when those features are not active. Yet this principle is an important to keep in mind going forward,
+as we hope not to make the current situation worse.
--- a/principles.rst
+++ b/principles.rst
@@ -4,7 +4,8 @@ Principles for GHC
 This document lays out high-level principles for the evolution of GHC
 and the language it compiles. In much the way that we like to write
 *properties* of our functions -- instead of just writing down individual
-test cases -- this document strives to write *principles* of our proposals.
+test cases -- this document strives to write *principles* that our proposals
+ought to maintain (or build towards).
 
 Articulating these principles helps to guide future proposals: proposals
 that maintain the principles in this document are more likely to be accepted,
--- a/principles.rst
+++ b/principles.rst
@@ -1,3 +1,6 @@
+.. sectnum::
+.. highlight:: haskell
+
 Principles for GHC
 ==================
 
--- a/README.rst
+++ b/README.rst
@@ -14,7 +14,7 @@ GHC.
 * `≡ List of rejected proposals <https://github.com/ghc-proposals/ghc-proposals/pulls?q=is%3Apr+label%3A%22Rejected%22>`_
 * `≡ List of implemented proposals <https://github.com/ghc-proposals/ghc-proposals/pulls?q=is%3Apr+label%3A%22Implemented%22>`_
 * `≡ List of all proposals <https://github.com/ghc-proposals/ghc-proposals/pulls?q=>`_
-  
+
 What is a proposal?
 -------------------
 
@@ -41,6 +41,10 @@ Should the GHC maintainers deem a change significant or controversial enough to
 warrant that, they may, at their discretion, involve the committee and ask the
 contributor to write a formal proposal.
 
+Proposals may contain amendments to our principles_, which serve as a guideline
+for future proposals and for the general evolution of GHC.
+
+.. _principles: principles.rst
 
 The life cycle of a proposal
 -----------------------------------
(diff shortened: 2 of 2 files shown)


principles.rst
```diff
@@ -0,0 +1,82 @@
+Principles for GHC
+==================
+
+This document lays out high-level principles for the evolution of GHC
+and the language it compiles. In much the way that we like to write
+*properties* of our functions -- instead of just writing down individual
+test cases -- this document strives to write *principles* of our proposals.
```
nomeata, line 7: Is it really a property of the proposal? Or rather a property of the design points that proposals move to? (Which would make them an invariant of the proposal, to stretch the analogy. Assuming a principled starting point).

principles.rst
```diff
@@ -0,0 +1,82 @@
...
+------------------------------
+
+When making a proposal, following the `usual guidelines <https://github.com/ghc-proposals/ghc-proposals/#how-to-start-a-new-proposal>`_,
+feel free to include a diff against this listing of principles. We can then
+discuss the validity of the new principle(s) alongside the concrete proposal for
+a change to GHC.
+
+We urge proposers to resist the temptation to propose principles without an
+accompanying concrete change to GHC. Debating principles in the abstract does
+not seem a productive use of time, though it is possible to imagine exceptions.
```
nomeata, line 26:
```suggestion
We urge proposers to resist the temptation to propose principles without an
accompanying concrete change to GHC. Debating principles in the abstract does
not seem a productive use of time, though it is possible to imagine exceptions. 
For example, the justification for the rejection of a proposal may be distilled into
a new principle here.
```

principles.rst
goldfirere: You may be onto something here. A principle here is more like an invariant than a property, I suppose. Care to suggest new wording?

nomeata:
```suggestion
test cases -- this document strives to write *principles* that our proposals
ought to maintain (or build towards).
```

Ericson2314:
Heh, I like this, both on it's own, and because it seems like a slippery slope to the AUG idea :D.

In particular, I think it would be good to ship the principles with the User's Guide, because I think end users benefit from learning the "why" in addition to the "what", even if they have no interest in personally changing GHC. To that end, I think it would be good to separate the preamble on how the principles are modified from the proposals themselves, as the former doesn't need to go in the User's Guide.

simonpj:
Yes -- writing down principles is a Good Thing

principles.rst
```diff
@@ -0,0 +1,84 @@
...
+~~~~~~~~~~~~~~~
+
+.. _OIP:
+
+**The Opt-In Principle (OIP):** (From `#378`_, slightly generalized.) Users who do not opt into an advanced feature will
+not be affected by it.
+
+This principle is violated in various ways today: it is easy for GHC to generate error messages that refer to
+advanced features even when writing simple code. In addition, the existence of advanced features likely slow
+down GHC even when those features are not active. Yet this principle is an important to keep in mind going forward,
```
JakobBruenker, line 83:
```suggestion
down GHC even when those features are not active. Yet this principle is important to keep in mind going forward,
```

AntC2:
I'd like to see some version of the [**Contiguous Scoping Principle**](https://github.com/ghc-proposals/ghc-proposals/pull/448#issuecomment-957591916). I'm not sure the wording at that link is exactly right: what I was aiming for in the Principle to which Richard's is a counter is there should be some syntactically recognisable demarcation of scope. Arguably, with current `-XScopedTypeVariables`

```haskell
id :: forall a. a -> a
id (x :: a) = x
```

The scope for `a` is contiguous. The trouble is there's no guarantee someone later won't put `LOTS OF CODE` between the statements. I was looking for some syntactic connecting or bracketing. (Or that proposal's long-term drift may be for _dis-_connecting. The `forall a`'s scope stops at the `;`. Then you need `@a` to bring the tyvar into scope of the equation for `id`.)

The 'problem' to address is that code uses typically a very small repertoire of var names `x, y` and tyvar names `a, b`. Is this a fresh `a` shadowing an existing? Or are the two `a`'s scopes disjoint? This will get tougher with DH wanting to collapse the namespaces: is this term usage `x` the same or shadowing that type usage `x`?

AntC2:
**Compact, consistent Syntax Principle**. @int-index has work in-progress to remove ad-hockery and idiosyncrasy from Haskell's lexical syntax. But every extension/proposal comes with its own new syntax, that has to be squeezed into the context-free syntactic design space. [Last para here](https://github.com/ghc-proposals/ghc-proposals/pull/448#issuecomment-955951816). "It's not that there's anything particularly bad about `-XXxx`; it's that there's another piece of syntactic bloat "

Comparing the Language reports for 1990 vs 2010: the 'Lexical Syntax' Reference grew by about 33%; the 'Context-Free Syntax' (including import/export) by about 25%. Admittedly by 2010, both main compilers already allowed plenty of extensions not in the Language Report. GHC has kept up a remarkable job of not expanding the reserved words/symbols -- in contrast to some languages I could mention. I rather fear the full syntax GHC supports today is much larger than the 2010 Report. (For example, `pattern` synonyms seem to have introduced a bunch of surprises.) I have to remember what all those constructs mean -- or at least recognise enough to know where to go look them up.

You'd think that introducing extensions that generalise/unify (like GADTs) a bunch of existing extensions would enable the replaced extensions to eventually be deprecated/retired. But no, I have to keep in the back of my mind all that stuff in case I'm looking through some old code in a tutorial or old paper.

One example of syntactic awkwardness/a large design space footprint for small semantic power has been in Haskell since forever: record expressions in terms:

    fexp qcon stuff
    fexp aexp stuff

Should mean the `fexp` is applied to two arguments. **Except** that if `stuff` turns out to be `{ fbind1, ..., fbindn }`, that's `fexp` applied to a single argument record construction/update. I don't see how we can ever remove this idiosyncracy, nor free up some syntactic design space for better/different record systems. Every other proposal affecting term syntax has to leave a wide berth around this adjacency-that-isn't-application. For example some equivalent for this would be nice with `-XTypeApplications`:

    ... foo   MkBar @a { bar1 = blah1 , bar2 = blah2 }       -- parse error on input `{'
    ... foo $ MkBar @a { bar1 = blah1 , bar2 = blah2 }       -- also parse error on input `{' grrr
    ... foo ( MkBar @a { bar1 = blah1 , bar2 = blah2 } )     -- also parse error on input `{' grrr grrr

    ... foo $ MkBar @a blah1 blah2                           -- happy
    ... foo ( MkBar @a blah1 blah2 )                         -- happy


We should really have that `{ ... }` signals 'here's record stuff', and all the messy gubbins is inside the braces, not needing to backtrack leftwards.

Ericson2314:
@AntC2 I happen to agree, but I think this proposal is more about whether we should list these prinicples at all, not what those proposals should be. Certainly it is explicitly requested in the section text that new principles only be proposed alongside proposals that enforce them, rather than being proposed in the abstract.

AntC2:
> @Ericson2314  I happen to agree, ...

Thanks. ? You agree that we should try to minimise syntactic footprint? And/or that proposed syntax should be consistent with existing? The proposal-in-progress here already specifies a 'Syntactic Unification Principle', accepted as part of the Dependent Haskell work. I thought I was building on that.

(Or were you agreeing with the 'Contiguous Scoping Principle' -- which builds on the 'Lexical Scoping Principle' already here?)

> but I think this proposal is more about whether we should list these prinicples at all, not what those proposals should be. Certainly it is explicitly requested in the section text that new principles only be proposed alongside proposals that enforce them, rather than being proposed in the abstract.

#448 is where the 'Contiguous Scoping Principle' arose. The same group of proposals introduces several extra bits of syntax/is contributing to syntactic bloat, so it seems pretty concrete to me. (And to parallel @goldfirere identifying places where Contiguous Scoping doesn't hold today, I was identifying places where today's function & pattern syntax seems already awkward/inconsistent. We probably can't ever fix those; at least we shouldn't make it worse.) 

I carefully (I thought) didn't comment on "what those proposals should be" in terms of what features or how they work. The proposal considers several alternative syntaxes. Can we establish Principle(s) here (or over there, to be transferred to here) that will inform choosing amongst them? Judging from the reactions, folk have a gut feel/intuitions that something isn't consistent. Can we turn that into a Principle? So quite apart from the semantic innovations:

* `@type` in patterns follows type `@`pplications in terms. Consistent?
* Extended [`let` in patterns](https://github.com/goldfirere/ghc-proposals/blob/type-variables/proposals/0000-type-variable-scoping.rst#36let-in-patterns) has drawn several comments re "unfortunate" syntax. I presume it's aiming to build on `let` in `case` guards, which was modelled on `let` in list comprehensions. But would it be more consistent to position decls for patterns as part of `where` decls? I find the 'Alternative 2'  `let`s confusing because it looks at first glance like the pattern has extra arguments, but on closer inspection they're decls. (Or 'Alternative 1' introduces scopes outside the `(let ... in ...)` -- which was a surprise to some;  'Alternative 2' "even stranger" was the comment).
* `let` declaring inline functions to then use in View Patterns are liable to make a binding lhs bulky/hard to follow. I already find View Pattern calls hard to follow. How does the proposed syntax sit with others' intuitions of what's syntactically consistent?
* `let type ...` consistent? Are there already places with syntax `let type ...`? (I plain don't know.)

----

* If `@type` is to appear in both patterns and function applications, the 'Syntactic Unification Principle' suggests inline `let` could/should appear in both(?) Or is there some contrary principle needs making explicit? Does Haskell need so many ways of introducing/scoping decls?

goldfirere:
This PR is *only* about the format / location / whether it's a good idea about writing down these principles. It is *not* about the principles themselves, which are already accepted in #378. It is also *not* about potential future principles, which would come after this PR (if you're too eager, you can make a PR against this PR).

Given the general support expressed here, and the lack of debate on the issue at hand, I'd like to submit this to the committee @nomeata.

simonpj:
> This PR is only about the format / location / whether it's a good idea about writing down these principles

That's not quite right is it?  The first sentence says

> This document lays out high-level principles for the evolution of GHC and the language it compiles. 

So I believe that *this document * is the long term repository for these principles; and if, in the future, we want to add to them or modify them we should make PRs against this document.    Is that right?  Let's be very explicit about that.

nomeata:
Since I only have devices of limited ergonomics around, feel free to submit it yourself. I guess for shepherding, if you think it improper to do it yourself as the author, Richard, I would have suggest SPJ (chairs are a good choice for meta-PRs like this)

ocramz:
> The principles written in the current document are just taken from #378. 

The underlying assumption being that fully dependent types are the only guiding principle for GHC for the foreseeable future? 

I understand it's a very large project, but surely GHC has much more than the surface type system. Or am I mistaken?

int-index:
> The underlying assumption being that fully dependent types are the only guiding principle for GHC for the foreseeable future?

There’s no such underlying assumption. It’s just that the work on dependent types was effective at revealing the need for such principles. They narrow down the design space and clear up common misconception. If any other major addition to the language starts generating the same amount of discussion, it would be wise to distill its core ideas into such principles as well.

goldfirere:
Yes, I agree with @int-index here. The principles in this initial document are the only ones we've articulated, but far from the only ones we have. I will amend the document to clarify this.

goldfirere:
@simonpj I believe that the existing document explicitly states how to update it, in the second section.

goldfirere:
As part of my interim secretary duties while @nomeata is traveling, I have assigned this to @simonpj as shepherd.

AntC2:
> > The underlying assumption being that fully dependent types are the only guiding principle for GHC for the foreseeable future?
> 
> There’s no such underlying assumption. ...

Sorry, I see no evidence there's work other than Dependent Haskell going on. Proposals on other topics (such as FunDeps/Overlaps) are not getting any oxygen. Even if they get accepted, they just sit in 'pending action' because no GHC developer is interested.

My Haskell/GHC experience has not improved since around 2015 (Pattern Synonyms); in some respects it's got worse (`AllowAmbiguousTypes`). The discussion here has turned down my suggested **Compact, consistent Syntax Principle**. I don't want to use a programming language that's only getting more bloated -- Haskell's succinctness has been one of its major appeals for me. I'm out -- of GHC recent and future releases, that is; not out of Haskell.

harpocrates:
> Sorry, I see no evidence there's work other than Dependent Haskell going on. Proposals on other topics (such as FunDeps/Overlaps) are not getting any oxygen.

The GHC proposal process is supposed to address the concern of a written proposal "not getting any oxygen" by codifying a review process - the committee is supposed to follow a set discussion schedule per proposal. Do you have a proposal in mind that is starving? Perhaps you have suggestions for improving the process?

> Even if they get accepted, they just sit in 'pending action' because no GHC developer is interested.

I'm sympathetic to the concern, but I think it is unfair. Unless you are helping guide resources towards GHC (either your own time, your employer's time, your money, your employer's money, etc.) I don't see why you should expect to see developer time directed towards any particular feature you consider to be important. Have you considered getting involved? The development experience is actually pretty good.

I view the GHC proposal process as a place where language implementors, designers, and users can consult and agree on the direction of evolution of GHC before merge requests are made. However, there is no guarantee that implementation progress _will_ be made - just encouragement that a patch would be accepted.

> My Haskell/GHC experience has not improved since around 2015 (Pattern Synonyms); in some respects it's got worse (AllowAmbiguousTypes). The discussion here has turned down my suggested Compact, consistent Syntax Principle. I don't want to use a programming language that's only getting more bloated -- Haskell's succinctness has been one of its major appeals for me. I'm out -- of GHC recent and future releases, that is; not out of Haskell.

Whenever there isn't complete agreement on an issue, there will be some people who will not be satisfied with the proposed change. Serious suggestion: have you considered patching Hugs (or even GHC) to support some of the ideas you propose?

AntC2:
@harpocrates  (Response [moved to a discussion](https://github.com/ghc-proposals/ghc-proposals/discussions/462) -- too meta for here.)

On one point:

> Serious suggestion: have you considered patching Hugs (or even GHC) to support some of the ideas you propose?

Yes am doing. (See more detail in the discussion.) Feedback welcome.

goldfirere:
@AntC2 

> The discussion here has turned down my suggested Compact, consistent Syntax Principle. 

That is not my intent, at all. Instead, my intent is only to keep this particular ticket about the location, format, and whether to include a list of principles in this repo. If you would like to propose a principle, I would recommend waiting until this proposal #452 is accepted and then to make a PR amending the list of principles. If you're eager to get going beforehand, you could make an Issue to start debate, and then make the PR later. My goal is merely to try to keep the conversations organized, not to opine on any principle. (Along similar lines, you moved the discussion about general directions out of this ticket and into a discussion -- thank you!)

simonpj:
As shepherd, I am recommending acceptance of the proposal.

Remember, we are not debating the principles themselves, all of which are embodied in earlier proposals.  Rather we are discussing

* the usefulness of having  a place where we articulate the principles, and
* a process for updating them over time, namely by PRs against this proposal

The idea is that the latest incarnation of the proposal is always the current set of agreed principles.

simonpj:
@goldfirere can you please number the sections?  Thanks

Ericson2314:
Is this ready to go? I see all the responses in https://mail.haskell.org/pipermail/ghc-steering-committee/2021-December/thread.html look positive. I hope it is!

simonpj:
Yes, I declare this proposal accepted.  (It had slipped off my radar over Xmas -- apologies.)

PR #454, opened 2021-11-03
Custom type warnings

adamgundry:
This proposal introduces a new built-in constraint form, `Warning`, as a mechanism for introducing custom warnings using type information.  This was originally brought up in the discussion of proposal #433, but has been separated out so that its design can be discussed in greater detail.

### DIFF
--- a/proposals/0000-custom-type-warnings.rst
+++ b/proposals/0000-custom-type-warnings.rst
@@ -1,0 +1,310 @@
+Custom type warnings
+====================
+
+.. author:: Adam Gundry
+.. date-accepted::
+.. ticket-url::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/454>`_.
+.. sectnum::
+.. contents::
+
+This proposal introduces a new built-in constraint form, ``Warning``, as a
+mechanism for introducing custom warnings using type information.  This was
+originally brought up in the discussion of `proposal #433
+<https://github.com/ghc-proposals/ghc-proposals/pull/433>`_ but has been
+separated out so that its design can be discussed in greater detail.
+
+
+Motivation
+----------
+
+GHC currently allows users to define custom type error messages using the
+``TypeError`` type family.  `Proposal #433
+<https://github.com/ghc-proposals/ghc-proposals/pull/433>`_ supplements this
+with a type class, ``Unsatisfiable``::
+
+  class Unsatisfiable (e :: ErrorMessage) where
+    unsatisfiable :: a
+
+This class has more predictable behaviour and carries evidence of the
+unsatisfiability of the constraint, but may only be used at kind ``Constraint``
+(whereas ``TypeError`` is poly-kinded).  Both ``TypeError`` and
+``Unsatisfiable`` carry an ``ErrorMessage`` that may contain both literal
+strings and instructions to render types as part of the message.
+
+Custom type errors are a very useful feature for libraries that wish to provide
+good-quality domain-specific error messages to their users.  However, sometimes
+library authors may wish to define domain-specific warnings, making use of type
+information (e.g. warning when a definition is used at at particular type, or
+when a particular class instance is used).  See `#17027
+<https://gitlab.haskell.org/ghc/ghc/-/issues/17027>`_ for a request for a
+feature like this.
+
+Warnings may currently be specified only in a limited manner via the ``WARNING``
+and ``DEPRECATED`` pragmas, which emit a fixed warning message whenever a marked
+definition is used.  There is no way for these pragmas to make use of type
+information, either to control when the warning is emitted or to vary the
+message that is displayed.
+
+This proposal introduces a new built-in constraint form ``Warning``, somewhat
+similar to ``Unsatisfiable``, which allows user-defined warning messages.  These
+constraints have a clear semantics for how they are solved and hence when
+warning messages are emitted.  They allow users to specify how such warnings are
+controlled by GHC's command-line warning flags such as ``-Wall``.
+
+For example, the following ``decode`` function will emit a warning if it is used
+at type ``Integer``, but not at any other types::
+
+     type family WarnInteger a where
+       WarnInteger Integer = Warning "decode-integer" Wdefault (Text "Integer may require unbounded memory!")
+       WarnInteger a       = ()
+
+     decode :: (FromJSON a, WarnInteger a) => ByteString -> Maybe a
+     decode = ...
+
+This warning is enabled by default, but may be disabled by passing the
+command-line option ``-Wno-custom:decode-integer``.
+
+
+Proposed Change Specification
+-----------------------------
+
+The ``GHC.TypeError`` module (the planned new home for ``TypeError`` and
+``Unsatisfiable`` per `!6066
+<https://gitlab.haskell.org/ghc/ghc/-/merge_requests/6066>`_ and `proposal #433
+<https://github.com/ghc-proposals/ghc-proposals/pull/433>`_) is extended with
+the following definitions::
+
+    type Warning :: Symbol -> WarningBin -> ErrorMessage -> Constraint
+    class Warning flag bin msg
+
+    data WarningBin = Wdefault | W | Wall | Weverything
+
+The ``Warning`` class resembles the ``Unsatisfiable`` class from `proposal #433
+<https://github.com/ghc-proposals/ghc-proposals/pull/433>`_, but represents a
+constraint which can be solved at any time by causing a warning to be emitted.
+
+``Warning`` constraints have the following properties:
+
+* During constraint solving, the solver treats ``Warning`` constraints like any
+  other class with no instances.
+
+* At the end of constraint solving, if a Wanted constraint of the form ``Warning
+  "flag" bin msg`` remains unsolved, a warning diagnostic is emitted, using the
+  message that results from normalising and rendering the type ``msg``, and
+  provided the warning bin ``bin`` is enabled.  Note that if the flag parameter
+  is not a literal ``Symbol``, the constraint is not automatically solved.  (See
+  below for discussion of how users may control whether this warning is
+  displayed.)
+
+* If a Wanted ``Warning`` constraint is solved using another mechanism
+  (e.g. because an identical Given constraint is in scope, or because of an
+  ``Unsatisfiable`` Given constraint), then no warning is emitted.
+
+* A ``Warning`` constraint is never automatically generalised.
+
+* The presence of ``Warning`` constraints does not affect pattern-match coverage
+  checking or the functional dependency check.
+
+* GHC will report an error if a user attempts to define an instance for
+  ``Warning``.
+
+* The representation of a ``Warning flag bin msg`` constraint in Core is
+  equivalent to the dictionary ``data WarningDict flag bin msg = WarningDict``.
+  This is GHC's normal representation of a class with no methods.
+
+
+Flags to control warnings
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The first parameter to the ``Warning`` constraint represents the flag used to
+control whether the warning is displayed.  Individual warning flags may be
+enabled or disabled using the ``-Wcustom:flag`` or ``-Wno-custom:flag`` options,
+and their priority may be controlled using the ``-Werror=custom:flag`` or
+``-Wwarn=custom:flag`` options, just as for GHC's built-in warnings.
+
+In addition, every flag belongs to a *warning bin* such as ``-Wdefault`` or
+``-Wall``, given by the second parameter to the ``Warning`` constraint.
+Available warning bins are represented by the ``WarningBin`` datatype.
+
+For example, assuming a ``Warning "flag" Wall (Text "Message")`` constraint has
+been solved:
+
+* No warning flags: no warning displayed
+
+* ``-Wcustom:flag``: warning displayed
+
+* ``-Wall``: warning displayed
+
+* ``-Wall -Wno-custom:flag``: no warning displayed
+
+* ``-Werror=custom:flag``: error message displayed
+
+In general, it is possible for the same warning flag to be used multiple times
+with different bins, in which case GHC decides whether to display these
+independently, based on the command-line options.  For example, a single module
+might give rise to both ``Warning "flag" Wdefault (Text "Message 1")`` and
+``Warning "flag" Wall (Text "Message 2")``, and would display only one message
+by default but both messages under ``-Wall``.  This is a departure from GHC's
+normal behaviour that each warning flag belongs to exactly one bin, but it seems
+unlikely to cause significant difficulties.
+
+
+Examples
+--------
+
+#. The Motivation section gave an example of a ``decode`` function will emit a
+   warning if it is used at type ``Integer``, but not at other types::
+
+     type family WarnInteger a where
+       WarnInteger Integer = Warning "decode-integer" Wdefault (Text "Integer may require unbounded memory!")
+       WarnInteger a       = ()
+
+     decode :: (FromJSON a, WarnInteger a) => ByteString -> Maybe a
+     decode = ...
+
+     no_warning :: ByteString -> Maybe Int
+     no_warning = decode
+
+     emits_warning :: ByteString -> Maybe Integer
+     emits_warning = decode
+
+     no_warning_either :: WarnInteger Integer => ByteString -> Maybe Integer
+     no_warning_either = decode
+
+   In the final definition, the presence in the context of the ``WarnInteger
+   Integer`` Given constraint (which simplifies to an application of
+   ``Warning``) suppresses the warning that would otherwise result from calling
+   ``decode``; instead it will be deferred to call sites.
+
+   If the library author wishes for the warning to be enabled when ``-Wall`` is
+   in effect but not by default, they may use ``Warning "decode-integer" Wall``
+   in place of ``Warning "decode-integer" Wdefault``.
+
+#. Andreas Klebinger in `#17027
+   <https://gitlab.haskell.org/ghc/ghc/-/issues/17027>`_ gives an example which
+   can be slightly reformulated to work with ``Warning``::
+
+     type FitsIn :: Nat -> Type -> Constraint
+     type family FitsIn x ty where
+        FitsIn 0 ty       = ()
+        FitsIn 1 ty       = ty ~ Word8
+        ...
+        FitsIn a ty       = ( Warning "fits-in" W (Text "The byte size " :<>: ShowType a :<>:
+                                                   Text " is defaulted to Integer.")
+                            , ty ~ Integer )
+
+   This demonstrates that the warning message can depend on the type.
+
+#. It is possible to define a class instance that emits a warning if used, for
+   example::
+
+     instance Warning "foldable-pair" Wall (Text "Foldable on pairs may cause confusion.")
+           => Foldable ((,) a) where
+        ...
+
+
+Effect and Interactions
+-----------------------
+
+None.
+
+
+Costs and Drawbacks
+-------------------
+
+Where a definition uses a ``Warning`` constraint, a dictionary argument will
+potentially be passed at runtime.  In practice this should typically be
+specialised away, but in corner cases this may affect semantics or runtime
+performance.  For example, adding a class constraint to an otherwise
+non-polymorphic definition may lead to loss of sharing.
+
+The extra development cost of this feature should be relatively low, as it has
+been designed to fit with GHC's existing constraint solver and is similar to the
+``Unsatisfiable`` constraint.  It is a little unclear how much complexity the
+custom warning flags will introduce.
+
+This feature is primarily targeted at advanced users, but should help library
+authors make their libraries more accessible to beginners by providing suitable
+domain-specific warnings.  Thus the effect on accessibility of the language to
+novice users is expected to be a net positive.
+
+
+Alternatives
+------------
+
+``TypeError`` vs. ``Unsatisfiable``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This proposal makes ``Warning`` analogous to the ``Unsatisfiable`` class
+introduced in `proposal #433
+<https://github.com/ghc-proposals/ghc-proposals/pull/433>`_.  That is, all
+warnings must be constructed at kind ``Constraint``.  As discussed in the
+previous proposal, this makes it possible to clearly specify the points at which
+warnings should trigger.
+
+It would be possible to make an analogue of ``TypeError`` instead, i.e. a type
+family usable at any kind.  This would be more flexible, albeit potentially less
+predictable.  Since there is no evidence carried by the dictionary, the
+restriction to ``Constraint`` is less strongly motivated than for
+``Unsatisfiable``, however.
+
+
+Pragmas vs. classes
+~~~~~~~~~~~~~~~~~~~
+
+The ``WARNING`` and ``DEPRECATED`` pragmas already allow warnings to be attached
+to modules or identifiers, such that any use of the identifiers will trigger the
+warning.  They cannot currently be used on class instances.
+
+This proposal is more flexible than ``WARNING`` and ``DEPRECATED`` pragmas,
+because:
+
+* Warnings may be displayed only when definitions are used at particular types.
+
+* Messages may be constructed dynamically, rendering types as part of the
+  message.
+
+* Warning flags and bins may be specified for individual messages, giving
+  fine-grained control to users, rather than all warnings being controlled by a
+  pair of flags (``-Wwarnings-deprecations`` and ``-Wdeprecations``).
+
+* It is possible to selectively postpone a warning by adding the corresponding
+  constraint to the context.
+
+A reasonable alternative to this proposal would be to extend ``WARNING`` and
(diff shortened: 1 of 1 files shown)


treeowl:
Suppose I have

```haskell
blah :: Warning "The integer could be too big." => foobly
```

I should really be able to *satisfy* that `Warning` constraint, effectively saying "I checked; it's not too big." Only *unsatisfied* warning constraints should bubble up and be reported. This suggests that warnings should have individual identities, tagged with something in `Type`. So maybe something like

```haskell
data IntegerTooBig
instance IsWarning IntegerTooBig where
  type Msg IntegerTooBig = "Here's the message for an integer being too big"
  type Bin IntegerTooBig = "the-name-of-the-warning-bin"
```

So now I'd write something more like

```haskell
blah :: Warning IntegerTooBig => foobly

okay :: Integer -> foobly
okay x
  | tooBig x = error "The potato cannon cannot accept potatoes that large."
  | otherwise = suppressingWarning IntegerTooBig $ blah x
```

Using phantom parameters in the declarations of the warning tags would allow them to be attached to particular values, I imagine, in case really fine control is required.

With this scheme, warnings can also be propagated explicitly:

```haskell
bingle :: Warning IntegerTooBig => grung
bingle = ... blah ...
```

There won't be any warning about the *definition* of `bingle`, because its constraint satisfies the warning for `blah`. But *using* `bingle` where the warning isn't suppressed will produce it.

treeowl:
Previous comment updated.

treeowl:
A simple example of the generality:

```haskell
data Nat = Z | S Nat
data Vec n a where
  Nil :: Vec Z a
  (:<) :: a -> Vec n a -> Vec (S n) a

data TooShort (m :: Nat) (n :: Nat)
instance IsWarning (TooShort m n) where
  type Name (TooShort m n) = "vector-short"
  type Msg (TooShort m n) = "Vector may be too short")
  type Bin (TooShort m n) = "partial"

unsafeIndex :: Warning (TooShort m n) => SNat m -> Vec n a -> a
unsafeIndex = ...

safeIndex :: forall m n a. (m < n) ~ True => SNat m -> Vec n a -> a
safeIndex = suppressingWarning @(TooShort m n) unsafeIndex
```

ocramz:
It would be nice to have the constraints inside the `IsWarning` instance.

Would something like this work? 

```haskell
instance ((m < n) ~ True) => IsWarning (TooShort m n) where ...
```

adamgundry:
@treeowl interesting point. I can see the utility of `suppressingWarning`. I think in principle we could define
```hs
suppressingWarning :: forall flag bin msg . (Warning flag bin msg => r) -> r
```
but clearly that will be more of a pain to use, requiring more type applications, compared to
```hs
suppressingWarning :: forall warningKey . (Warning warningKey => r) -> r
```

I experimented with a similar approach using datatypes to identify warnings, but it appears we can get away without it in `base` (https://gist.github.com/adamgundry/b3b9a131003e5f016992f9e8183aa59b). Thus my inclination was towards the simpler option, i.e. putting the minimal API into `base` and leaving such wrappers to library code. But perhaps there is some value in more actively pushing people towards the "higher-level" API using datatypes, especially as it makes `suppressingWarning` more convenient? I'm somewhat on the fence here.

goldfirere:
I generally favor the idea of a minimal feature in GHC, allowing building on top in other APIs. Maybe one such approach becomes standard enough that we put it in `base` -- or maybe we are happy enough to put something in `base` right away, but then we should consult the CLC.

I'm happy with the proposal as stated. Thanks @adamgundry!

chshersh:
@adamgundry That's a great proposal! 👏🏻 I believe it can improve the UX of Haskell libraries significantly.

Here are several use cases for type-directed warnings that come to mind according to my experience:

1. The `Partial` constraint. With such proposal it should be possible to add a `Partial` constraint to each non-total function to warn about the fact that it's partial.
2. Unlawful instances. E.g. `Monad` for `ListT` transformer that is going to be removed from `mtl`. To have a smoother migration guide, a warning can be introduced first to warn people.
3. Functions like [`universe`](https://hackage.haskell.org/package/relude-1.0.0.1/docs/Relude-Enum.html#v:universe) or [`inverseMap`](https://hackage.haskell.org/package/relude-1.0.0.1/docs/Relude-Enum.html#v:inverseMap) can warn when used with too big data types (e.g. `Int`).
4. Warning that `nub` is quadratic and suggesting alternatives (I have several anecdotal use-cases when program runtime decreased from several seconds to milliseconds after replacing `nub` with `ordNub` or similar).
5. Warnings for `trace` functions so you don't leave them in production code.
6. `IsString` instance for something like ByteString can warn that it trims characters.
7. Functions that work with `Text` in `IO` (e.g. `readFile` from `text`) can warn that they are encoding-sensitive.

This feature becomes extremely helpful if you also put migration guides inside such warnings. Fortunately, it's not a problem if you use a package like [`type-errors-pretty`](https://hackage.haskell.org/package/type-errors-pretty) for writing long messages.

The only part I'm not sure is "warning bins". I usually always use `-Wall` and I believe it's considered one of the Haskell best-practices to always enable `-Wall` and even some extra warnings. I don't remember anyone using just `-W`. So, in my vision, I would just remove warning bins entirely and output all warnings from the library by default. This simplifies the `Warning` typeclass from having 3 type parameters to just 2: name and text.

However, I'd love to be able to disable all warnings from a particular library with something like `-Wno-custom:some-library/*`. And, generally speaking, having some sort of glob support would be nice so I can easily disable classes of errors (`-Wno-custom:text/text-io-*`).

adamgundry:
@chshersh thanks for the detailed list of use cases!

> The only part I'm not sure is "warning bins". I usually always use `-Wall` and I believe it's considered one of the Haskell best-practices to always enable `-Wall` and even some extra warnings. I don't remember anyone using just `-W`. So, in my vision, I would just remove warning bins entirely and output all warnings from the library by default. This simplifies the `Warning` typeclass from having 3 type parameters to just 2: name and text.

This is an interesting point. Perhaps the `WarningBin` stuff could indeed be disposed with and custom warnings always emitted (unless explicitly silenced). That's a nice simplification to the design, but of course it takes a little flexibility away from library authors to decide how important their warning is. How do others feel about this option?

> However, I'd love to be able to disable all warnings from a particular library with something like `-Wno-custom:some-library/*`. And, generally speaking, having some sort of glob support would be nice so I can easily disable classes of errors (`-Wno-custom:text/text-io-*`).

I suppose we could establish the convention that custom warnings from a particular library should mention the name of that library at the beginning of the flag name. I'm not sure how we would otherwise determine which library is responsible for originally defining any given warning?

I can see some appeal of glob support, especially if we drop warning bins from the constraints. But it feels somewhat complex to solve a problem that we don't yet have (libraries introducing many custom warnings, perhaps with several levels of categorization). So I'm inclined to leave that out of the proposal for the time being and see if the need for the feature develops in the future.

goldfirere:
Maybe `WarningBin` is indeed a step too far. I could see holding off here until someone asks for it.

Strongly in favor of establishing a convention around warning names.

I also like waiting for glob support. Seems easy to add later, and unnecessary to add now.

adamgundry:
Thanks everyone! I have revised the proposal in the light of feedback:
 * Warning bins have been removed, so `Warning` now has only two parameters and warnings are always enabled by default.
 * A naming convention for flags is recommended.
 * A new function `suppressWarning` is included, so users can suppress a single warning regardless of flags.  It has a carefully-chosen type so that it can be called with a type application using a single constraint synonym representing an application of `Warning`.
 * The alternatives section includes discussion of glob support and other ideas that have been raised during the discussion.

Any further feedback before I submit this to the committee?

goldfirere:
I like this proposal.

But I think `suppressWarning` has an awkward interface: if I understand correctly, the caller would need to provide the exact text (actually, worse: internal structure) of the warning in order to suppress it. Is there a way to do this where the caller of `suppressWarning` passes only the flag?

adamgundry:
> But I think `suppressWarning` has an awkward interface: if I understand correctly, the caller would need to provide the exact text (actually, worse: internal structure) of the warning in order to suppress it. Is there a way to do this where the caller of `suppressWarning` passes only the flag?

This is true and I agree it isn't ideal. Though to be precise, `suppressWarning` can be called with the entire `Warning flag msg` constraint as the type argument. Thus if a library exports a constraint synonym such as
```hs
type MyLibraryWarning = Warning "my-lib/my-warning" (Text "Blah blah")
```
then the user can write `suppressWarning @MyLibraryWarning`, which seems not too bad. It does depend on the library exposing suitable constraint synonyms though.

I don't see a good way to specify only the flag when calling `suppressWarning`. For inference to work we'd need some sort of functional dependency `flag -> msg`, but that would limit us to one message per flag (effectively preventing having warnings involving type parameters).

Another possibility is the approach @treeowl brought up in https://github.com/ghc-proposals/ghc-proposals/pull/454#issuecomment-968417439: define a datatype for each warning, and have the flag and message calculated by type families. That means those datatypes can be used to call `suppressWarning`. I'm still not convinced it is worth the extra complexity in `base`, however, and it can be done at the library level (per [this gist](https://gist.github.com/adamgundry/b3b9a131003e5f016992f9e8183aa59b)).

treeowl:
It's just as easy to implement the simple version using the datatype one.

adamgundry:
@goldfirere great suggestion, thanks! It didn't occur to me to use `QuantifiedConstraints` but it works perfectly here. I've updated the proposal to use this.

I've also discussed @treeowl's suggestion in more detail in the Alternatives (subsection 6.4). I still think we should stick with the approach recommended in the main text of the proposal, but if there are advantages to the alternative that I'm missing, please do point them out.

adamgundry:
@nomeata I would like to submit this proposal to the committee.

i-am-tom:
Hi @adamgundry, thanks for your patience :)

It looks like everything has been worked out! I must say, I do like the flag/message distinction and the ability to suppress based on flags (this feels more in line with the rest of Haskell - I suppress the flag, not the message).

I think @chshersh has done a really neat job of summing up some strong motivations for this extension, but I'd like to repeat my own from the other thread: it's really helpful to write something along the lines of...

```haskell
handleSpecificError :: Warning "todo" "SpecificError handler" => SpecificError -> IO ()
handleSpecificError = undefined
```

... and then let GHC manage my todo list :)

Anyway, the point I'm getting to is that I like this proposal a lot, and will submit it to the committee; thanks for your work on this!

Ericson2314:
On that note between void runtime rep, core coercisons (or do they use void runtime rep?), dependent haskell stuff including `exists`, Haskell has / is getting a lot of mechanisms to deal with erasability. The overlap makes me a bit nervous.

adamgundry:
> Can we guarantee erasure? Is `Warning flag msg => t` the same operationally as just `t`?

Indeed, my understanding is that `Warning flag msg => t` is operationally more or less equivalent to `() -> t` rather than `t`. We might prefer `(# #) -> t` but that requires the ability for constraints to be unlifted (https://gitlab.haskell.org/ghc/ghc/-/issues/19476). Unfortunately I'm not sure how to do better here.

treeowl:
Why can't we get erasure (without deferred type errors)?

i-am-tom:
Apologies for leaving this so long:

* @int-index are you satisfied with @adamgundry's answer?
* I think @treeowl's question is related, @int-index, and we're a bit too deep in the compiler for me to have any good answers; would you mind answering this one?

int-index:
> @int-index are you satisfied with @adamgundry's answer?

I’m satisfied with the answer in the sense that it is precise and clear, but not with its implications. It means that this feature, which is supposed to improve developer experience, will at the same time decrease the runtime performance (perhaps mitigated in some cases my the optimizer, but not always).

What we could try instead is to introduce a new quantifier for erased constraints. It would be a much more invasive change than what is proposed here, but I’d like to see this alternative elaborated and discussed before we accept this one.

> I think @treeowl's question is related ... would you mind answering this one?

Yes, so here’s my understanding. If we have a constraint `c => t`, then simply by virtue of `=>` being there, we are going to compile it to a function in Core, and then it’s up to the optimizer (specifically, specializer) to get rid of it when possible.

If we want to *guarantee* erasure, we need another construct. For example, `forall` is guaranteed to be erased.

An option we could explore here is that instead of `Warning flag msg => t` one would write `forall (w :: Warning flag msg). t`. Then the `w` variable would be erased, and that’s what we want. Yet we don’t want the variable itself, since it shouldn’t actually be used in the rest of the type, it’s more of a `forall (_ :: Warning flag msg). t`. But that also isn’t quite right, since forall-bound variables are inferred by unification, not constraint solving.

So what we actually want is (I conjecture) a new construction, `Warning flag msg =>. t`, where `=>` is exactly like `=>.`, but is guaranteed to be erased.

@adamgundry Would you be interested in exploring this option?

Ericson2314:
We will also need erased constraints for "promoted" methods to replace type families without a runtime cost, so I am all for it.

adamgundry:
This is an interesting point. I agree that we are missing a way to express erased constraints, and that in a more dependently-typed world, such a thing might be useful in other contexts.

That said, I don't know if we would really want `=>.` to be a separate quantifier. It seems to me a promising approach is the one described in [#19476](https://gitlab.haskell.org/ghc/ghc/-/issues/19476): generalise `Constraint` so that it can be used at runtime representations other than `LiftedRep`, and make `=>` polymorphic in a `RuntimeRep` parameter just like `->`. Then `Warning` could produce a `Constraint (TupleRep '[])` and we would have a guarantee that it had a zero-width runtime representation.

int-index:
@adamgundry Correct me if I’m wrong, but `(# #) -> t` is still not the same as `t`. Here’s a simple test to demonstrate it:

```
{-# LANGUAGE UnboxedTuples #-}

import Unsafe.Coerce

newtype T = MkT ((# #) -> Int)

x :: T
x = MkT (\_ -> 42)

main = print (unsafeCoerce x :: Int)
```

If I compile and run this program, it will print `0` rather than `42`. If I run `:main` in GHCi, it prints a number like `8589934593` or `4567545576`.

The problem isn’t the runtime representation of `(# #)`, which is indeed `TupleRep ‘[]`. The problem is the runtime representation of `->` itself. We need a new quantifier so we don’t have a function at all, rather than make its parameter zero-width.

adamgundry:
Yes, that's true, I elided that detail. I'm hazy about how much the difference between `(# #) -> t` and `t` matters in practice, though, when it comes to performance? We can't `coerce` from one to the other, but is calling a function with a zero-width argument significantly more expensive than merely evaluating the underlying closure?

int-index:
>  is calling a function with a zero-width argument significantly more expensive than merely evaluating the underlying closure?

Perhaps not, but generally I wouldn’t expect any change in runtime behavior at all from adding a warning.

treeowl:
I'm sure it depends on context. `(##) -> a` has one more bottom element than `a`.

aspiwack:
> Correct me if I’m wrong, but `(# #) -> t` is still not the same as `t`.

While this is true, it is pretty much irrelevant is the expression you are adding a warning to is a function or has other constraints. Which I expect to essentially always be the case.

Besides, an erasable quantifier would not solve this bit (from the proposal): “For example, adding a class constraint to an otherwise non-polymorphic definition may lead to loss of sharing.”. So even that is not a guarantee of unchanged runtime behaviour.

All in all, it's not worth losing hair over.

More interesting is whether the proposal strikes the right balance. The alternative of using a generically kinded warning à la `TypeError` in particular seems to me worth discussing. In the current proposal type families cannot emit warnings (note: however since warnings do not abort the computation, a type family emitting a warning would have to still return a value too, I'm not sure what that ought to look like), in exchange the behaviour is very clearly defined.

Is it what we want? Possibly. I'm inclined to say “yes” actually. But it's probably worth a little bit of ink

adamgundry:
> More interesting is whether the proposal strikes the right balance. The alternative of using a generically kinded warning à la `TypeError` in particular seems to me worth discussing. In the current proposal type families cannot emit warnings (note: however since warnings do not abort the computation, a type family emitting a warning would have to still return a value too, I'm not sure what that ought to look like), in exchange the behaviour is very clearly defined.

The proposal discusses this briefly in section 6.1, but I agree that the alternative could be fleshed out a bit. We could imagine adding `TypeWarning` roughly as proposed in [ghc#17027](https://gitlab.haskell.org/ghc/ghc/-/issues/17027):
```hs
type family TypeWarning (msg :: ErrorMessage) (x :: k) :: k
```
where the idea is that `TypeWarning msg x` would be provably equal to `x`, but use of this fact would emit `msg` as a warning.

However, much as in #433, the basic problem with this is that GHC does not have a clear semantics for when type family reduction occurs. Thus it is hard to specify precisely (and hard for the programmer to predict) when the warning should be emitted. There are a couple of plausible strategies:
 * Emit the warning when the rewriter needs to reduce an application of `TypeWarning msg x`. This is intuitively "right", but at the moment, the rules around when a stuck type family application gets its arguments rewritten are rather complex and opaque. It also differs from how `TypeError` works, which is based on unsolved constraints.
 * Do not reduce `TypeWarning` in the rewriter, but if an unsolved constraint mentions `TypeWarning` at the end of constraint solving, emit the warning, then try to solve the constraint with the `TypeWarning` application removed. This is more consistent with `TypeError`, but rather different from normal type family reduction.

The key place where `TypeWarning` can be used but `Warning` cannot is the RHS of a type family. I think it is always possible to rewrite a program using `TypeWarning` to one that uses `Warning`, by replacing type family applications with constrained variables (logic programming style), but this is a global transformation.

Of course, nothing stops us from having both the poly-kinded `TypeWarning` type family and the `Warning` class from the proposal, much as post-#433 we will have both `TypeError` and `Unsatisfiable`. But given the above I'm not extremely eager to add `TypeWarning`.

aspiwack:
You are making a good point, indeed. The complications for `TypeWarning` are quite hairy. It's fair to say that the `Warning` constraint gets a lot of of the benefits for a fraction of the cost. I'm essentially convinced at this point.

goldfirere:
For what it's worth, I'm not thrilled that adding these warnings incurs a runtime cost. In my view, any compile-time niceties that impose a runtime cost are bugs. Maybe they're very, very hard bugs to fix, but they're bugs.

Right now, we fully erase type-level quantification. So `nada :: forall a. Maybe a` really has the same exact runtime representation as `zilch :: Maybe Int`. It is not hard to imagine extending this idea to runtime-irrelevant constraints. That doesn't quite solve the "lost sharing" problem, but I'm sure we could fix that, too, if we tried. For me, this little argument says that we shouldn't stall on this particular point: that is, I think this feature is worthwhile and the runtime cost is both small and plausibly fixable. So I favor accepting this change, possibly with an open ticket that says we should make it even better.

nomeata:
And you are not worried that once we know how to make this feature free of runtime cost we will notice that we can't reasonably migrate to that form anymore?

goldfirere:
The question will be how GHC decides whether to erase a constraint. Is it by looking at some property of the constraint itself (say, its kind)? Or will it be by some syntax (say, on the `=>`)? I don't know. Either of these sound plausible to me, but the latter does seem more likely, meaning that users of `Warning` will have to update their code accordingly. Even this, though, is not so bad: when those users are ready to, they can just add the new little bit of syntax and make their library a tiny bit faster. The old syntax will remain available... just a little slower. And note that any library-writer who uses this syntax has *already* opted into the runtime hit. (This proposal does *not* suggest scrapping the current `{-# WARNING ... #-}` syntax.) So any migration simply means these libraries suffer either with CPP or a longer range of versions with the runtime hit. I'm fine with all of this.

goldfirere:
To be clear, we should adequately document the runtime hit. Even better if that documentation comes with actual data, preferably showing a ~0.01% slowdown or something similarly unscary.

i-am-tom:
While I see @int-index's point, and appreciate that a potential runtime hit here is undesirable, I'm still happy to recommend acceptance. Thanks to everyone who's explained the issues above, though I do have one more question: is there an example program that would suffer these penalties _without_ raising custom warnings? In other words, is it a runtime penalty only for programs that produced warnings while compiling? I'm struggling to come up with an example to the contrary.

adamgundry:
> is there an example program that would suffer these penalties _without_ raising custom warnings? In other words, is it a runtime penalty only for programs that produced warnings while compiling?

It seems likely there exist pathological cases where introducing a warning may affect operational semantics or runtime performance, without emitting a warning when compiling. But I've not been able to construct an example, and I have no idea how to meaningfully measure the "expected" runtime hit. In real-world code, I suspect there will almost always be no difference.

This proposal has been stalled for rather too long. I'd like to ask the committee to either accept or reject the proposal.

nomeata:
Pinging @i-am-tom explicitly, can you guide us to a conclusion?

adamgundry:
@i-am-tom ping?

simonpj:
Two comments on the propsoal

1.  The proposal says

"If a Wanted Warning constraint is solved using another mechanism (e.g. because an identical Given constraint is in scope, or because of an Unsatisfiable Given constraint), then no warning is emitted."

The parenthesis implies that Warning constraints can be solved by "another mechanism" and partenthetically gives examples of some such mechanisms.  

But I'd **much** prefer if the proposal explicity stated "A Warning constraint `Warning flag msg` can be solved by:..." and then give a series of bullets that specifies all the ways it can be solved.

2. I really struggled to understand 2.1 "Flags to control...".    For example

* Is `-Wcustom` new?  I guess so.
* `-Wcustom:flag` is all in one font, but I think you mean `-Wcustom:<flag>`, where `<flag>` is user-specified.  (When you give examples, please don't use "flag"; use "wombat" or "containers" or something.)
* "For example, assuming a Warning "flag" (Text "Message") constraint has been solved automatically:".  I thought that solving it automatically (as in (1) above) would kill off the warning. Only *unsolved* `Warning` constraints are reported.
* Are "warning bins" described somewhere.  If so, can you add a link?  If not can you explain?
* "...controlled, just as for GHC's builti-in warning". Please give a link to where these controls are described.
* 
Maybe you can review the wording of this section.

i-am-tom:
@nomeata @adamgundry Apologies, this fell off my radar. I'm certainly still in favour of acceptance - @int-index have you had any further thoughts here? Otherwise, pending @simonpj's changes, I'll happily resubmit for acceptance!

Ericson2314:
Could a `=>.` created after the fact be used with `Warning`? That would be ideal.

adamgundry:
Yes, I don't see `Warning` alone as particularly strong motivation for `=>.`, but it seems likely that future changes to GHC (either to allow constraints with a zero-width runtime representation, or indeed that are completely erased) could subsequently be used with `Warning` without too much difficulty.

I'll give this one more editing pass in due course.

i-am-tom:
Just commenting to say that I'm still keeping an eye on this issue (pending edits), and I agree that `=>.` is perhaps out-of-scope, but definitely worth exploring in a separate issue!

goldfirere:
Changed labels to make it clear that the next action is @adamgundry's. Please let me know if this is incorrect!

adamgundry:
Thanks, yes I'm planning to revise this to take account of committee feedback.

adamgundry:
I've unfortunately not yet found the opportunity to revise this proposal. However there has recently been quite a bit of discussion on https://github.com/haskell/core-libraries-committee/issues/87 (see also [Reddit discussion](https://www.reddit.com/r/haskell/comments/xbvy7o/add_warning_to_datalistheadtail/)) that suggests some form of custom warnings and local suppression of warnings would be useful. In particular, it would be nice to be able to add warnings to partial functions like `head`, provided they could be selectively disabled at the module level or at individual call sites.

I'm wondering whether using types as in this proposal is the best approach, or whether it would be reasonable to extend `WARNING` pragmas with a custom flag name, allow them to be attached to instances, and add another pragma to suppress particular warnings at call sites. That would avoid all the concerns about `Warning` constraints appearing in types and potentially changing semantics or performance. It is slightly less flexible, in particular because it would not be possible to construct the text of a warning from type information (as in the `FitsIn` example in the proposal). But perhaps that flexibility isn't really needed?

googleson78:
Personally I like the fact that `suppressWarning` can be used much more modularly than on a per-module basis. I could even imagine much fancier stuff being possible because of this, e.g. `suppressWarningAssert x`, which apart from removing the warning during compile time, would redisplay it again in a runtime exception, if `x` happens to throw.

Of course, this might be a bit contrived of an example, but having the warning be a (relatively) first class citizen of Haskell instead of a string might open up many such possibilities.

But in the end, I do agree that having **some** form of custom warnings is important to be able to move forward on other fronts, as https://github.com/haskell/core-libraries-committee/issues/87 demonstrates. Would it be perhaps possible to have the "extend WARNING approach" designed in such a way to be mostly compatible with potential future version(s) of `Warning`?

adamgundry:
I've opened https://github.com/ghc-proposals/ghc-proposals/pull/541 as an alternative to this proposal. It is more conservative, sticking closer to `WARNING` pragmas as they currently exist. But I think it would be easy enough to add `Warning` constraints on top later, if we wish.

adamgundry:
Status update: it looks like #541 is making progress. This proposal will need to be rewritten to use the flags agreed there, adding only the ability to define custom warnings via a type-based mechanism. It should also take account the concerns raised in the discussion that warnings in types may end up affecting runtime performance. (Perhaps it makes sense to take the conservative alternative of allowing `WARNING` pragmas to be attached to instances.)

I'm not currently intending to pick this up, so I'll mark it as dormant. If anyone would like to take it over, please feel free.

adamgundry:
> (Perhaps it makes sense to take the conservative alternative of allowing `WARNING` pragmas to be attached to instances.)

See https://github.com/ghc-proposals/ghc-proposals/pull/575 for a proposal doing just this.

PR #457, opened 2021-11-08
#281: forall unicode syntax

int-index:
I forgot about `UnicodeSyntax` when writing this part of the specification.


int-index:
This change fixes an omission and can probably bypass the proposal process.

@goldfirere As the shepherd of the original proposal, do you concur?

PR #409, opened 2021-03-08
Exportable named defaults

blamario:
This proposal is an adjusted version of an earlier [Haskell 2020 proposal](https://github.com/haskell/rfcs/pull/18) that received some approval, for whatever it's worth.

[Rendered](https://github.com/blamario/ghc-proposals/blob/exportable-named-default/proposals/0000-exportable-named-default.rst)

PR #458, opened 2021-11-12
Syntax for tuples, unboxed tuples, unboxed sums

int-index:
When writing #281, I thought I could postpone specifying the treatment of tuples under `NoListTupleTypeSyntax`. However, what slipped my attention was that we need some syntax to report inferred types in GHCi and error messages:

```
ghci> :set -XNoListTupleTypeSyntax
ghci> :t ('x', True)
```

If we printed `(Char, Bool)` here, we’d be flat out wrong, because under this extension this syntax actually stands for a pair of types, not for the type of a pair. Instead, we should print something like:

```
ghci> :t ('x', True)
Tuple2 Char Bool
```

There are plenty of options besides `Tuple2`. We could go with a type family taking a list of types (i.e. `Tuple [Char, Bool]`), a variadic data family (`Tuple Char Bool`), magic names starting with numbers (`2-Tuple Char Bool`), or just a different naming convention (`TupleOf2 Char Bool`).

After having discussed this with @goldfirere and @simonpj, we decided against `Tuple [Char, Bool]` and variadic `Tuple Char Bool`. The former does not interact well with partial application (which is needed to define higher-kinded instances such as `Functor`), and the latter would require additional complexity in the compiler (special type inference rules for the wired-in `Tuple` tycon, type families in data family return kind, data family promotion).

This leaves us with `Tuple2`, `2-Tuple`, and `TupleOf2`.

* I like `2-Tuple` the most (as that’s how we say in out loud), but it would require an ad-hoc extension to the lexer to allow a name starting with a number and with a hyphen. I bet I could make it work, but I’m not sure what others think of the idea.

* The `Tuple2` name is the most straightforward solution, we already have other names with numeric suffixes (e.g. `zipWith3`, `zipWith4`, etc). However, the numeric suffix convention is also used by classes, such as `Eq1` and `Eq2` in `Data.Functor.Classes`, with the number indicating the arity of a higher-kinded type variable. So this interpretation would suggest that `Tuple2` is defined as `data Tuple2 f g x y = T2 (f x y) (g x y)` or similar.

* The `TupleOf2` resolves the potential confusion with the `Eq1` / `Eq2` notation, but the name is a bit longer and does not look as nice.

This should probably be decided by a vote.

### DIFF
--- a/proposals/0281-visible-forall.rst
+++ b/proposals/0281-visible-forall.rst
@@ -630,9 +630,9 @@ Syntax
 
    * ``[]`` as the list type constructor
    * ``()`` as the unit type
-   * ``[a]`` the syntax of a list type
-   * ``(,)`` as the pair type constructor
-   * ``(a, b)`` as the syntax of a pair type
+   * ``[a]`` as the syntax of the list type
+   * ``(,)`` as the pair type constructor (likewise for ``(,,)``, ``(,,,)``, and so on)
+   * ``(a, b)`` as the syntax of the pair type (likewise for ``(a, b, c)``, ``(a, b, c, d)``, and so on)
 
    When the extension is on, these constructs retain their Haskell 2010
    meaning, which depends on whether we are in a type-level or term-level
@@ -654,23 +654,41 @@ Syntax
 
    Export the following synonym from the ``Data.Tuple`` module::
 
-     type Unit = ()
+     type Unit = ()   -- Tuple0 has a special name
 
-   For tuples of higher arities, users may want to define their own
-   synonyms, such as::
+     type Tuple2 = (,)
+     type Tuple3 = (,,)
+     type Tuple4 = (,,,)
 
-     type TupleOf2 = (,)
-     type TupleOf3 = (,,)
-     type TupleOf4 = (,,,)
      ... -- up to the maximum tuple arity
-
-   We do not propose adding them to ``Data.Tuple`` for the time being,
-   as better APIs are possible (e.g. based on variadic data families)
-   but are blocked by other technical issues. We leave it as future work.
 
    This change allows the use of built-in lists and tuples without any
    disambugation syntax (the ``'`` promotion syntax at the type level or the
    ``type`` herald at the term level).
+
+   Under ``NoListTupleTypeSyntax``, the synonyms are also to be used when
+   printing inferred types.
+
+   This extension interacts with ``UnboxedTuples``, ``UnboxedSums``, which also
+   rely on punning. ``ListTupleTypeSyntax`` has a similar effect on ``(# #)``,
+   ``(# , #)``, ``(# | #)``, and so on.
+
+   Export the following names from ``GHC.Exts``::
+
+     type Unit#    -- Tuple0# has a special name
+
+     type Tuple1#
+     type Tuple2#
+     type Tuple3#
+     type Tuple4#
+
+     ... -- up to the maximum unboxed tuple arity
+
+     type Sum2#
+     type Sum3#
+     type Sum4#
+
+     ... -- up to the maximum unboxed sum arity
 
 5. When ``ViewPatterns`` are enabled, interpret ``f (a -> b) = ...``
    as a view pattern, otherwise as ``f ((->) a b) = ...``.
@@ -1040,7 +1058,7 @@ Corner Cases
    One way to change this is to use synonyms
    from ``Data.Tuple`` and ``Data.List``::
 
-     x1 = g (TupleOf2 Int Bool)
+     x1 = g (Tuple2 Int Bool)
      x2 = g (List Int)
 
    Another way is to use the ``type`` herald::
@@ -1061,8 +1079,8 @@ Corner Cases
    One way to resolve this is to use synonyms
    from ``Data.Tuple`` and ``Data.List``::
 
-      a = f @(TupleOf2 Int Bool)
-      b = g  (TupleOf2 Int Bool)
+      a = f @(Tuple2 Int Bool)
+      b = g  (Tuple2 Int Bool)
 
    Another way is to use the ``type`` herald::
 
--- a/proposals/0281-visible-forall.rst
+++ b/proposals/0281-visible-forall.rst
@@ -659,7 +659,6 @@ Syntax
      type Tuple2 = (,)
      type Tuple3 = (,,)
      type Tuple4 = (,,,)
-
      ... -- up to the maximum tuple arity
 
    This change allows the use of built-in lists and tuples without any
@@ -681,14 +680,13 @@ Syntax
      type Tuple2#
      type Tuple3#
      type Tuple4#
-
-     ... -- up to the maximum unboxed tuple arity
+     ... -- and so on
 
      type Sum2#
      type Sum3#
      type Sum4#
+     ... -- and so on
 
-     ... -- up to the maximum unboxed sum arity
 
 5. When ``ViewPatterns`` are enabled, interpret ``f (a -> b) = ...``
    as a view pattern, otherwise as ``f ((->) a b) = ...``.


phadej:
> However, the numeric suffix convention is also used by classes, 

That reasoning is a stretch. If anything `Eq1` naming is wrong as it tries to encode kind<sup>&dagger;</sup> with a number.

<sup>&dagger;</sup> And not only kind but also variance, e.g. `class ShowP f where showP :: f a -> String`, i.e. 1-arity, kind polymorphic and phantom-variant is also useful (and exists by different names in different libraries). `Eq1` is like a *acetone* which preferred IUPAC name is *Propan-2-one*, except we don't have proper nomenclature for higher-kinded classes.

int-index:
OK. Then strike out `TupleOf2`.

glaebhoerl:
Would an alternative variation on the variadic version, where you write `Tuple 2 Char Bool` -- that is, the first parameter is a `Nat`, which then determines the rest of the kind -- change the picture at all? It seems like it would require less magic, but I don't know how *much* less. (Or maybe it would require *more*, due to dependency? I'm not up to date with the current state of Dependent Haskell features. But in any case, putting the arity up front does give the compiler strictly more information to work with than trying to figure that out on the fly.)

int-index:
@glaebhoerl It would require somewhat less magic. We could define:

```
type family TupleKind n where
  TupleKind 0 = Type
  TupleKind n = Type -> TupleKind (n - 1)

data family Tuple n :: TupleKind n

data instance Tuple 0 = ()
data instance Tuple 1 a = Solo a
data instance Tuple 2 a b = (a, b)
data instance Tuple 3 a b c = (a, b, c)
```

I agree it would be more elegant than having entirely separate `Tuple2`, `Tuple3`, `Tuple4`, etc.

There are still two engineering challenges here:

1. Promotion of data families (tuples can be promoted, we don’t want to lose that)
2. Type families in data family return kinds (currently disallowed)

monoidal:
1. I would not say that names are exported from `GHC.Prim` - this module is an internal implementation detail. Let's just use `GHC.Exts` in the specification, even if the implementation will use `GHC.Prim`.

2. Unboxed sums currently must be arity >= 2 (e.g. see `sumTyCon`).

   There are multiple problems if you want to add 0-sums and 1-sums:

* You'd need special syntax for constructors: `(# 1# | | #) :: Sum3# Int# a b`, `(# 1# | #) :: Sum2# Int# a`, but the logical syntax for 1-sum `(# 1# #)` is already taken by tuples.
* The 0-ary sum is tricky: it's an actually empty type, without `_|_`. What should be its runtime representation? Logically, it should be a memory location that cannot hold anything, but there's no such thing.
* The name `Void#`, while fitting, has already been taken in `GHC.Prim` until 9.0 by a different type. While this has been deprecated and moved to `GHC.Types` in 9.2, it's too early to reuse the name.

   I suggest removing 0-sums and 1-sums for now. It's a can of worms, and the utility is questionable.

3. You define a synonym `type Tuple1 = Solo`.

   Do you plan to print 1-tuples using `Tuple1` or `Solo`? In other words, is one of the names preferred?

   I would consider removing this synonym: it will lead to confusing rewriting. If you plan to keep it, for consistency I'd expect a synonym `type Tuple0 = Unit`.

int-index:
Thanks for taking a look @monoidal 

1. OK. Not sure what module would be best, but `GHC.Exts` is good enough, I suppose. Updated.
2. Good point. I don’t quite follow the reasoning about the runtime representation of `Void#`, we don’t need a representation as there are no values of this type, but I agree that as long as we constraint sums to have arity `>=2`, we don’t need names for arity 0 or 1. Removed.
3. The attempt to include `Solo` in this extension is a mistake on my behalf. Alphanumeric names require different treatment. Removed.
4. Quite right. Fixed.

AndreasPK:
> ... -- up to the maximum unboxed tuple arity

As I remember it there is no maximum arity for unboxed tuples.

int-index:
Can we export an infinite family of identifiers `Tuple2#`, `Tuple3#`, ..., `Tuple9223372036854775807#`, ...?

Well, somehow we manage to export `(#,#)`, `(#,,#)`, `(#,,,#)`, and so on. So it should be possible with alphanumeric names as well.

AndreasPK:
> Can we export an infinite family of identifiers `Tuple2#`, `Tuple3#`, ..., `Tuple9223372036854775807#`, ...?
> 
> Well, somehow we manage to export `(#,#)`, `(#,,#)`, `(#,,,#)`, and so on. So it should be possible with alphanumeric names as well.

They are not really exported like regular definitions. Instead the parser when it comes across a unboxed tuple dynamically creates the right type/dataCon on the fly. We could do the same thing for `Tuple<n>#` in theory. Not sure if that's worth it for unboxed tuples.

int-index:
Hm. I don’t want to introduce more things handled magically by the parser. I think it’s ok to virtually export `Tuple<n>#` from `GHC.Exts`.

That is, if you write

```
import GHC.Exts
```

then all of `Tuple2#`, `Tuple3#`, `Tuple4#`, and so on, are brought into scope. You can also:

* use it qualified: `GHC.Exts.Tuple2#` 
* import only some identifiers: `import GHC.Exts (Tuple3#)`
* import with an alias: `import GHC.Exts as E` and then write `E.Tuple3#`
* re-export the all of them in their entirety, e.g. `module M (module GHC.Exts)` would re-export `Tuple2#`, `Tuple3#`, ad infinitum.

The last point about re-export suggests that we might want a separate module just for unboxed tuples/sums. What about `GHC.Types.Unboxed`?

AndreasPK:
I have to say I'm not convinced about the effort/complexity/benefit tradeoff for unlifted tuples here. For regular tuples it's straight forward though.

int-index:
Consider this GHCi session:

```
GHCi, version 9.2.1: https://www.haskell.org/ghc/  :? for help
ghci> :set -XUnboxedTuples 
ghci> :t (# 'x', True #)
(# 'x', True #) :: (# Char, Bool #)
```

The purpose of NoTupleTypeSyntax is to eliminate ambiguity between list and tuple type/data constructors. So the `(# 'x', True #) :: (# Char, Bool #)` output would not be correct. We need `(# 'x', True #) :: Tuple2# Char Bool` or something like that.

monoidal:
> Instead the parser when it comes across a unboxed tuple dynamically creates the right type/dataCon on the fly.

I can't confirm. I created a file with `x = (#,,,,...,,#)` and got an error "A 64-tuple is too large for GHC".

I suspect GHC might manufacture unboxed tuples during compilation when needed (`tupleTyCon` has a clause for building larger tuples) but it doesn't seem we parse them.

AndreasPK:
> > Instead the parser when it comes across a unboxed tuple dynamically creates the right type/dataCon on the fly.
> 
> I can't confirm. I created a file with `x = (#,,,,...,,#)` and got an error "A 64-tuple is too large for GHC".
> 
> I suspect GHC might manufacture unboxed tuples during compilation when needed (`tupleTyCon` has a clause for building larger tuples) but it doesn't seem we parse them.

Ah that's right. I think nested unboxed tuples get "flattend out" somewhat early on and that's what I was thinking about. Sorry for the noise.

sheaf:
Let me try to summarise some of the options we have (up to renaming).

**1.** `Tuple3 a b c`: keep the current definition of tuples, but with a different name for the type constructors.

  - Pros:

    - Straightforward, simple to understand.
    - Least amount of breakage.
    - Simplest to implement (same behaviour as now).

  -  Cons;

     - Each tuple type has a separate definition, which can create boilerplate compared to having a single (parametric) definition.
     - If we use this as a stepping-stone with the intent to upgrade it later, we would break people's code twice.

**2.** `Tuple 3 a b c`: allow the arity to be specified as a parameter, as a built-in data family.

Here, we can either require that the dimension always be provided, or allow `Tuple` to appear on its own.

  - Pros:
    - Defines all tuples in a single family.
    - Users can still define instances for these tuples as before.

  - Cons:
    - The kind of `Tuple` can be confusing for newcomers. Enforcing application of `Tuple` to at least one argument attempts to hide this complexity somewhat.
    - Requires support for promoting data families (so that we can use the usual value-level tuple syntax at the type-level).
    - Requires support for type families in the declared return type of a data family (in order for instances to still work in the presence of `-XPolyKinds`).

**3.** `Tuple [a,b,c]`: use a list to specify the components, as a built-in data family.

  - Pros:
    - Brings tuples closer to being inductively defined.
    - Simple to understand.
 
  - Cons
    - Breaks existing instances that focus on final parameters, e.g. `instance Functor ((,,) a b)`.
    - Can't be partially applied (but we could provide newtypes, e.g. `newtype Tuple2 a b = Tuple2 (Tuple [a,b])`.
    - Requires support for promoting data families.

hsyl20:
@sheaf thanks for the summary.

Tuples were discussed at [HIW 2016](https://www.youtube.com/watch?v=6BqEt9d-amM&list=PLnqUlCo055hX1F0PCi9FjdllYQMwCQvps&index=11) with an emphasize on option 3 `Tuple [a,b,c]`. It would be great to have more inductive tuples indeed. In the discussion, the issue of having one infotable per tuple type was mentioned but I think we could lower tuples into `SmallArray`s before codegen to avoid this.

> Breaks existing instances that focus on final parameters, e.g. instance Functor ((,,) a b).

Shouldn't this be in "pros"? :trollface:

int-index:
> Shouldn't this be in "pros"?

It should be! That's how I'd like my tuples. And then a separate data type like `data Annotated e a = e :# a` with the `Functor`, `Foldable`, and `Traversable` instances that are currently defined for to pairs. But this would be an enormous breaking change.

Is there anyone willing to push it through?

goldfirere:
I think the "last parameter" argument is a small red herring. `instance Bifunctor (,)` is very reasonable today, and there would be no way to write it with `Tuple [...]` syntax.

sheaf:
> I think the "last parameter" argument is a small red herring. `instance Bifunctor (,)` is very reasonable today, and there would be no way to write it with `Tuple [...]` syntax.

I did intend to refer to omitting any number of parameters, not just a single one, but I agree it wasn't very clear and it was a bit misleading. I've edited the post.

goldfirere:
After some discussion today, I think the best way to move forward is to collect more input from the community. @sheaf summarizes the possibilities well in https://github.com/ghc-proposals/ghc-proposals/pull/458#issuecomment-974121585. I would like to orchestrate a vote, where different emoji reactions on a particular post indicate different preferences. But before we do that, we need to agree on what the ballot will look like. I propose:

1. `Tuple2`, `Tuple3`, `Tuple4`, ...
2. `Tuple 2`, `Tuple 3`, `Tuple 4`, ...  Writing `Tuple` on its own is an error: the following lexeme must be a literal number. If you like you can think of `Tuple 2` as a compound name.
3. `Tuple 2`, `Tuple 3`, `Tuple 4`, ...  Writing `Tuple` on its own is allowed. It would have kind `forall (n :: Nat) -> TupleKind n` where `type family TupleKind n where TupleKind 0 = Type; TupleKind n = Type -> TupleKind (n-1)`. `Tuple` would be a data family.
4. `2-Tuple`, `3-Tuple`, `4-Tuple`, ... This would require special lexing for the new names, and would conflict with uses of the `-` operator.

I have not included `Tuple [x,y,z]` as a possibility here for the primitive name of a tuple because there would be no way to write the instances we have today. No matter which of the above choices we make for the primitive names, we (or anyone else) could write a type family that takes a list to build a tuple. We might decide even to export such a definition from `base`, but I want to make that decision separately.

The question here is: **Is this the right ballot?** Please, do **not** respond to this post saying which option above you would like. Instead, please respond to this post saying whether this slate of options is the right slate to be considering. Feel free to suggest new additions to the ballot (though it will get awkward to have more than 6 in the GitHub voting scheme I envision). Sometime next week (presumably), I will kick off the actual vote (which will be open for at least a week).

Thanks!

/remind me Wednesday to kick off the vote if the conversation has led to a good resting place

nomeata:
I have a gut feeling that removing the nice special syntax for tuples and lists will cause an uproar against which the #EqMassacre will feel like a leisurely discussion of the weather over tea. So community involvement is certainly a good way to maybe mitigate that.

I wonder if we need to put options on the ballot that provide some similar symbolic ergonomics as what w e have now. Too bad that we can't do what ML languages do (`(a * b)`).

Maybe we need an option “None or the above, please try harder?”

Or maybe this fear is unwarranted, and we can see if it materializes.

int-index:
We actually can do `a * b` but that would work only for pairs.

goldfirere:
An interesting point @nomeata. Firstly, this is opt-in: you're affected only with `-XNoListTupleTypeSyntax`. But perhaps the set of people affected will come to grow over time, and so we should try to do something nice.

I'm content with a "try harder" option.

Actually, because I just thought of it, how about this: `(~ a, b, c ~)`. Note the space before the `a`, distinguishing this from a lazy pattern (which requires a `~` in prefix notation). Why `~`? To remind people that tuples are *lazy*.

Just for fun, we could go further. Suppose `(! a, b, c !)` are the types of strict tuples. Then we have something like this:

```hs
data (! a, b, c !) = MkStrictTuple3 !a !b !c
data (~ a, b, c ~) = MkLazyTuple3 a b c

type IsTuple3 :: (Type -> Type -> Type -> Type) -> Constraint
class IsTuple3 tup_type where
  mkTuple3 :: a -> b -> c -> tup_type a b c
  matchTuple3 :: (a -> b -> c -> r) -> tup_type a b c -> r
instance IsTuple3 (!,,!) where
  mkTuple3 = MkStrictTuple3
  matchTuple3 k (MkStrictTuple3 a b c) = k a b c
instance IsTuple3 (~,,~) where
  mkTuple3 = MkLazyTuple3
  matchTuple3 k (MkLazyTuple3 a b c) = k a b c
pattern (,,) :: IsTuple3 tup_type => a -> b -> c -> tup_type a b c
pattern (a,b,c) <- (matchTuple3 MkLazyTuple3 -> MkLazyTuple3 a b c)
  where (,,) = mkTuple3
{-# COMPLETE (,,) #-}
{-# INLINE (,,) #-}
```

With this scheme, we have `(,,)` as a shared data constructor for both strict and lazy tuples. (The pattern synonym definition uses `MkLazyTuple3` because using `MkStrictTuple3` would force the components of the lazy tuple, which we definitely do not want.) Topping us off with (#409) an exported default

```
default IsTuple3 ( (~,,~), (!,,!) )
```

I think this would even be backward compatible with what we have today. And I would imagine that straightforward inlining would get us today's performance, though this would need to be tested.

I am *not* proposing we do all of that. (You may want to. I would be in support -- once we can confirm that error messages, etc., don't leak the abstraction. If you can come up with syntax for mixed lazy/strict tuples, that would be fun and should work with the pattern synonym scheme above.) But it was why I chose the `(~ ... ~)` notation.

int-index:
I’m not sold on the Cthulhu operator `(~,,~)`. Perhaps it’s just me, but this is too heavy on punctuation (with tildes both at the start and the end).

Since the call is to “try harder”, let me suggest mixfix unicode. Write the product of two types as `A × B`, the product of three types as `A × B × C`, of four types `A × B × C × D`, and so on.

The prefix forms will be `(×)`, `(××)`, `(×××)`, and so on.

nomeata:
Oh, Unicode cross product is perfect (if you consider non-ASCII admissable), both semantially and visually. emoji-with-hearts-in-the-eyes.

Re opt-in: It's not really opt in if you have drank the cool-aid of single-namespace-so-no-punning-Haskell. So while it is practically opt-in due to how we move Haskell forward, let's try as hard as if it was _not_ opt-in.

tomjaguarpaw:
> instance `Bifunctor (,)` is very reasonable today

One of my hopes for dependent types is that they will provide a more coherent approach to "multi-type-variable-functorness". Rather than `Functor`, `Profunctor`, `Bifunctor`, `Trifunctor`, `Quadrafunctor`, `ThreeFunctorialParametersThenTwoCofunctorialParametersThenAnotherFunctorialOne`, ... ad infinitum, I hope that dependent types will open up a world with one functor notion that subsumes all of these. Please let's not eschew uniformity if the only reason is to continue support for crude special purpose hacks.

> and there would be no way to write it with Tuple [...] syntax

Good!

goldfirere:
As much as I love `A × B × C`, I don't see this happening. I agree with @nomeata that this new syntax is going to become the preferred standard at some point in the future, and it would be a significant departure from precedent to use Unicode. Separately, I don't know how to parse that: is `×` a binary operator? How does it produce a flat datatype? Or would we require parens, with `×` effectively replacing `,`?

Ericson2314:
Just to be clear, we do need *something* in the short term, right? Can we just do one of the simple options *and immediately deprecate it* just to get the Dependent Haskell pioneers unblocked? It's better to incubate stuff before we decide on the one right approach anyways.

int-index:
> Separately, I don't know how to parse that: is × a binary operator? How does it produce a flat datatype?

It’s an n-ary operator. We parse it as binary and then flatten it out immediately after fixity resolution.

> Or would we require parens, with `×` effectively replacing `,`?

I don’t see a need to require the parens.

> it would be a significant departure from precedent to use Unicode

Well, `*` is the ASCII counterpart of `×`. But then we couldn’t use both tuples and numeric multiplication in the same module unqualified. That’s bad.

What about `&`? That is, `A & B & C` looks fine to me. Ah, but that would conflict with `(&&)` in prefix form.

OK. It seems the ASCII space is exhausted. That’s why I suggested Unicode.

So what about this:
* without UnicodeSyntax, just go with the simple thing: `Tuple2`, `Tuple3`, etc;
* with UnicodeSyntax, support the pretty `A × B × C` syntax

We could recommend `A × B × C`, but ASCII-purists still have the option of avoiding it.

goldfirere:
It sounds like you're proposing (1) from https://github.com/ghc-proposals/ghc-proposals/pull/458#issuecomment-974241528, but with a separate feature to allow `×` to be used in the way you describe. That is, the raw tuple type would still be e.g. `Tuple3`, but `-XUnicodeSyntax` (or some new extension) would allow the new syntax. I'd definitely be in support of having this extra syntax... but it still doesn't negate the need to define what the raw tuple looks like.

I agree that the idea of the n-ary `×` belongs in the context of any vote, but I don't see it as one of the options in the vote.

int-index:
Yes, that is accurate.

Ericson2314:
would `(×)`, `(××)`, `(×××)` etc. exist with `-XUnicodeSyntax` too?

monoidal:
I don't have a strong opinion on A × B; I would prefer deferring the decision for later and keeping the required changes to minimum. It's easy to add things later, hard to remove.

However, if we go in this direction:
* It'd be good to have it for unboxed tuples for consistency. Possible option: `A ×# B ×# C` - mirroring that `(+#)` is the unboxed equivalent for `(+)`. We also need the unapplied version - either `(××#)` or `(×#×#)`. No idea about unboxed sums.
* Is the idea to support `A × B` with UnicodeSyntax only, or with UnicodeSyntax + NoListTupleSyntax? My intuitive choice would be the former option - just like we'd support Tuple2 with ListTupleSyntax. But then this is changing UnicodeSyntax which is an established extension. There are three packages https://hoogle.haskell.org/?hoogle=%28%C3%97%29&scope=set%3Astackage - looks minor but needs checking. It'd be weird if an extension called NoListTupleSyntax enabled additional tuple syntax.
* Would × be a normal identifier on the term level and special syntax on the type level? Or is the plan to reserve this syntax on both levels? (This is assuming no additional Dependent Haskell extensions.)

goldfirere:
Thanks for the recent conversation. It does not seem to change the original ballot, so I'm going to go ahead and commence voting.

**Rules**:

* This vote is non-binding. We are doing this to gauge the sentiment of the community, but new thoughts may come to light that e.g. invalidate one of the choices. We will try to follow the general will of the community, but take this vote as providing advice, not a decision.
* You *may* vote for more than one option.
* Vote by reacting with a GitHub emoji. You can do so by clicking the smiley face to the top-right of this post. Underneath there, you will have a choice of 8 emoji. I have labeled each of the options with the emoji you will use to vote.

**Fixed firmament**:

All options below have these properties in common:

* We propose to rename the family of built-in tuple types, using the naming scheme to be chosen below.
* Terms are entirely unchanged: tuple expressions are still denoted with parentheses and commas.
* The old names will be available by default and will be printed by GHC by default. This proposal is thus fully backward-compatible.
* When `-XNoListTupleTypeSyntax` is in effect, GHC will use the new names below for both input and output.
* Tuples of size 0 and 1 are special. A tuple of size 0 is called `Unit`, and a tuple of size 1 is called `Solo`.

**Choices**:

1. Vote with 👍. `Tuple2`, `Tuple3`, `Tuple4`, ... Accordingly, a tuple of `Int`, `Double`, and `Bool` would look like `Tuple3 Int Double Bool`.
2. Vote with 😄. `Tuple 2`, `Tuple 3`, `Tuple 4`, ... Writing `Tuple` on its own is an error: the following lexeme must be a literal number. If you like, you can think of `Tuple 2` as a compound name. Accordingly, a tuple of `Int`, `Double`, and `Bool` would look like `Tuple 3 Int Double Bool`.
3. Vote with 🎉. `Tuple 2`, `Tuple 3`, `Tuple 4`, ... Writing `Tuple` on its own is allowed. It would have kind `forall (n :: Nat) -> TupleKind n` where `type family TupleKind n where TupleKind 0 = Type; TupleKind n = Type -> TupleKind (n-1)`. `Tuple` would be a data family. Accordingly, a tuple of `Int`, `Double`, and `Bool` would look like `Tuple 3 Int Double Bool`.
4. Vote with ❤️. `2-Tuple`, `3-Tuple`, `4-Tuple`, ... This would require special lexing for the new names, and would conflict with uses of the `-` operator without any spaces around it. Accordingly, a tuple of `Int`, `Double`, and `Bool` would look like `3-Tuple Int Double Bool`.
5. Vote with 🚀. `(~ , ~)`, `(~ ,, ~)`, `(~ ,,, ~)`, ... This is a mixfix notation like the current tuple syntax, but with `~`s in the corners. Accordingly, a tuple of `Int`, `Double`, and `Bool` would look like `(~ Int, Double, Bool ~)`.
6. Vote with 👀. Keep looking. None of these is satisfactory. It would be better to stall the implementation of accepted proposal #281 than to use any of the above.

**Next steps**:

Some comments above (starting with https://github.com/ghc-proposals/ghc-proposals/pull/458#issuecomment-974506703) suggest using a mixfix version of the Unicode `×` to represent tuple types, and there is also some enthusiasm behind a `Tuple [Int, Double, Bool]` syntax. Both of these are very possible, but neither is a good enough answer to be the canonical name for tuple types (though it's conceivable to use `×` to represent tuples with `-XUnicodeSyntax` enabled). I would imagine that the mixfix `×` would be agreed upon if we choose one of the prefix syntaxes for tuples (choices 1-4).

Let the voting commence!

simonpj:
> neither is a good enough answer to be the canonical name for tuple types

Why is `Tuple [Int, Bool]` not a good enough answer?  You say above "because there would be no way to write the instances we have today."  But perhaps you mean rather

* there is no way to write an *unsaturated* two-tuple, written `(,)` today.  

We'd have to say `Tuple [_,_]` perhaps, or `Tuple [,]`.  What about a partial application to one argument?  `Tuple [Int, _]`  perhaps.  Or `Tuple [Int,]`.   Have I understood right?   The underscore notation or comma notation looks just about acceptable to me.

goldfirere:
Yes, that's right about unsaturated tuple types: these are the ones causing problems with the `Tuple [Int, Bool]` syntax.

I'm personally not all that excited about `Tuple [Int, _]` or `Tuple [Int, ]` or some such without a more general framework for list sections. Unless `Tuple` effectively becomes a keyword and the list construct is parsed specially in some way. Seems like it would be a strange bump in the surface of the language.

One idea got floated in a Tweag Haskeller conference call today: it seems that Scala has `Tuple2`, `Tuple3`, etc., for their tuples, with `(Int, Bool)` desugaring to `Tuple2 Int Bool`. It has had this for sometime, and every gets on with it. We could do the same for Haskell: use `Tuple [Int, Bool, Double]` syntax as our usual notation (including in pretty-printing!), but have this desugar to `Tuple3 Int Bool Double`. Really, this is a vote for 👍 but with the addition of special handling in the pretty-printer for fully saturated tuple types (the very common case) to use `Tuple` instead.

goldfirere:
@monoidal brought an unfortunate problem to my attention today: constraints. With `-XNoListTupleTypeSyntax`, presumably we still have

```
f1 :: (Show a, Eq a) => a -> String
```

But do we have

```
f2 :: (Show a, (Eq a, Monoid a)) => a -> a
```

?

Do we have

```
type ShowEq a = (Show a, Eq a)
```

? Does it matter whether we have `type ShowEq :: Type -> Constraint` as a standalone kind signature?

Do we have

```
x :: Dict (Show a, Eq a)
```

?

We can't really proceed without answers here, sadly.

monoidal:
Credit where it's due: the problem was found by @amesgen.

Some options:

(1) Continue to use `(,)`. But then, `(Show a, Eq a)` is still punning. Is this a single `Constraint` or is this a type-level tuple of kind `(Constraint, Constraint)`?

(2) Introduce `CTuple` parallel to `Tuple` and `Tuple#`. This would mean that `f :: (MonadX m, MonadY m, MonadZ m, Ord a) => m a -> m a` becomes `f :: CTuple4 (MonadX m) (MonadY m) (MonadZ m) (Ord a) => m a -> m a`, which is ugly. Advantage: currently GHC has hacks so that `()` is a Type while `() :: Constraint` is a Constraint. This would no longer be necessary.

(3) Reuse `Tuple` and make it kind-polymorphic.

Related random observation: unboxed tuples are representation-polymorphic. We can't have `Tuple# [Int#, Float#]` :(

(4) Use `CTuple [MonadX m, MonadY m, MonadZ m]`. Note that while partially-applied value tuples are common (e.g. they have `Functor` instances), partial application of constraint tuples is rare (if not impossible).

goldfirere:
> Some options:
> 
> (1) Continue to use `(,)`. But then, `(Show a, Eq a)` is still punning. Is this a single `Constraint` or is this a type-level tuple of kind `(Constraint, Constraint)`?

It would not be hard to keep `(Show a, Eq a) => ...` working: we have the `=>` to disambiguate. I think this could even be handled at the parser level. The other cases are still problematic, though.

> 
> (2) Introduce `CTuple` parallel to `Tuple` and `Tuple#`. This would mean that `f :: (MonadX m, MonadY m, MonadZ m, Ord a) => m a -> m a` becomes `f :: CTuple4 (MonadX m) (MonadY m) (MonadZ m) (Ord a) => m a -> m a`, which is ugly. Advantage: currently GHC has hacks so that `()` is a Type while `() :: Constraint` is a Constraint. This would no longer be necessary.

I like `CTuple` as the fully desugared version.

> 
> (3) Reuse `Tuple` and make it kind-polymorphic.

Hm. If we're going to write out a name, it might as well be `CTuple` instead of `Tuple`. Or maybe `Constraints` (note the plural). But kind-polymorphism would also work.

> 
> Related random observation: unboxed tuples are representation-polymorphic. We can't have `Tuple# [Int#, Float#]` :(

Yes we can: `Tuple# :: List (exists r. TYPE r) -> 😞`. I was all excited to use existentials. But the kind of the result depends on the kinds of the arguments. We'd need

```hs
type FList :: forall k. (k -> Type) -> List k -> Type
data FList f args where
  FNil :: FList f []
  FCons :: f x -> FList f xs -> FList f (x : xs)

type Tuple# :: forall reps. FList TYPE reps -> TYPE (TupleRep reps)
```

And now we'll need overloaded lists to keep the nice syntax. Urgh.

> 
> (4) Use `CTuple [MonadX m, MonadY m, MonadZ m]`. Note that while partially-applied value tuples are common (e.g. they have `Functor` instances), partial application of constraint tuples is rare (if not impossible).

Partial application of constraint tuples *is* possible, but indeed rare. But the `CTuple2`, `CTuple3`, etc syntax would work here.

JakobBruenker:
> It would not be hard to keep (Show a, Eq a) => ... working: we have the => to disambiguate. I think this could even be handled at the parser level. The other cases are still problematic, though.

So in other contexts, like

```haskell
type All :: (Type -> Constraint) -> [Type] -> Constraint
type family All c ts where
  All c '[] = ()
  All c (t:xs) = (c t, All c ts)
```

is the idea that you'd have to use `CTuple`/`Constraints` here with -XNoListTupleTypeSyntax?

JakobBruenker:
One thing that I find somewhat confusing about `CTuple` is that it looks a lot like `Tuple`, but
```haskell
                  (1,2) :: Tuple [Int, Int]
CTuple [Show a, Read a] :: Constraint
```
it's on the other side here.
I like `Constraints` though, it doesn't raise this concern.

int-index:
@JakobBruenker Think of this at three levels instead of two

```
  (1,2) :: Tuple  [Int, Int]       :: Type
        :: CTuple [Show a, Read a] :: Constraint
```

monoidal:
I understand the idea is to keep special form `(C1, C2, .., Cn) => ...`, but any other occurrence such as `Dict (Eq a, Show a)` would have to be rewritten using `CTuple` or `CTuple2` depending on the accepted variant. For example, under `-XNoListTupleSyntax`, both `f :: ((Eq a, Show a)) => ...` and `type T = () :: Constraint` would be a kind error.

The syntax `deriving (C1, C2)` would continue to be supported.

I prefer a separate name `CTuple` to tricks with kind polymorphism.

goldfirere:
I have filed #475, meant to replace this. @int-index, do you agree that we should close this one?

int-index:
Yes, thank you for taking over this!

PR #460, opened 2021-11-19
371-non-magical-eq: more graceful migration (under review)

int-index:
While implementing this proposal, I noticed that there are quite a few packages (just to name some: primitive, containers, Cabal, haddock, exceptions) that use `~` without enabling `TypeOperators`.

The proposal already includes a compatibility fallback, but does not cover this particular problem. This amendment extends the proposed compatibility fallback by making the usage of `~` without `TypeOperators` a warning instead of an error. In due course, the special case will be removed, but maintainers won’t have to modify their code immediately.

### DIFF
--- a/proposals/0371-non-magical-eq.md
+++ b/proposals/0371-non-magical-eq.md
@@ -35,11 +35,17 @@ Let `n` be the GHC version that implements this proposal.
 
 Starting with GHC `n`, include a compatibility fallback:
 
-* When the lookup for `~` fails because it is not in scope,
-  assume it refers to `Data.Type.Equality.~`.
-  When this happens and `-Wcompat` is in effect, emit a warning.
+1. When the lookup for `~` fails because it is not in scope,
+   assume it refers to `Data.Type.Equality.~`.
+   When this happens and `-Wtype-equality-out-of-scope` (included in
+   `-Wcompat`) is in effect, emit a warning.
 
-Starting with GHC `n+2`, enable the warning by default.
+2. When the use of `~` would have been rejected because `TypeOperators` are not
+   enabled, accept the program regardless.
+   When this happens and `-Wtype-equality-requires-operators` (enabled by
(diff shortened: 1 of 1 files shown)


phadej:
See https://github.com/haskell/core-libraries-committee/issues/12#issuecomment-971794515 and following discussion

> In fact from our perspective it would be better not to have the warning stage, it's just extra churn. We'll fix it when it breaks.

int-index:
Yes, the discussion seems relevant (but not conclusive). When @simonmar says “We'll fix it when it breaks”, the hidden assumption is that they have complete control over their code and can just fix it. But what about external dependencies?

What if the maintainer takes a while to respond? Should I just postpone the migration to a newer GHC version until the transitive closure of all my dependencies get updated, or should I create dozens of forks to use in the meantime?

The warning allows me to fix *my* code without waiting for others to fix *their* code.

int-index:
Another perspective here: https://gitlab.haskell.org/ghc/ghc/-/issues/18862#note_312815

> `Writing a ~ b does not require enabling -XTypeOperators.`
> ^ this was my only potential gripe. I see no problem with this if we don't have to enable it.

goldfirere:
I'm in support of this direction. Would you mind naming the warning in the proposal?

int-index:
> Would you mind naming the warning in the proposal?

Named them `-Wtype-equality-out-of-scope` and `-Wtype-equality-requires-operators`. Better suggestions are welcome. The latter one probably doesn’t need a flag at all, since enabling `TypeOperators` and disabling the warning is the same amount of effort, but I’m not sure if it’s a good idea to introduce a warning without a flag.

monoidal:
@int-index Is anything pending here, do you intend to submit?

int-index:
Let’s submit, hi @nomeata

goldfirere:

/remind me in a week to accept if there is no dissent from the committee.

PR #461, opened 2021-11-25
Class Backend - separate interface from representation

Icelandjack:
This is a work in progress proposal to bring "backends" to type classes. It aims to tackle three aspects of classes: presentation, representation and generalisation.

The feature works by specifying a type class as a frontend by adding a deriving clause to the end of it, any instance of `Front A` is virtual and is translated into `Back A` as if the user had written `deriving via Via A instance Back A`: 

```haskell
class Front a where
  ..
  
  deriving Back a
  via Via a
```
### Presentation
We don't commit to one class presentation; whether we use `Applicative` or a `Monoidal` interface or whether we want to bundle our types as "associated type families" (`Representable`) or exposed as multi-parameter type classes (`RepresentableOf`). They can be used interchangably.

```haskell
class Functor f => RepresentableOf rep f where
  indexOf :: f a -> (rep -> a)
  indexOf = index
  
  tabulateOf :: (rep -> a) -> f a
  tabulateOf = tabulate
  deriving Representable f
  via WrappedRepresentableOf rep f
```

The newtype instance describes how to translate an exposed class into a bundled class:

```haskel
type    WrappedRepresentableOf :: Type -> (Type -> Type) -> (Type -> Type)
newtype WrappedRepresentableOf rep f a = WrappedRepresentableOf (f a)
  deriving newtype (Functor, RepresentableOf rep)
  deriving anyclass Distributive -- (this can be derived with this proposal)

instance RepresentableOf rep f => Representable (WrappedRepresentableOf rep f) where
  type Rep (WrappedRepresentableOf rep f) = rep

  tabulate = tabulateOf
  index = indexOf
```

### Representation
Many type classes like `Traversable` and `Distributive` cannot be derive at all through `coerce`-based deriving. Most of these are derivable with a change of representation, for example by backing `traverse :: .. -> f (t b)` with [`backedTraverse :: .. -> Yoneda f (t b)`](https://hackage.haskell.org/package/kan-extensions-5.2.3/docs/Data-Functor-Yoneda.html).

### Generalisation
There is no way to currently evolve our ever-expanding menagerie of functors (`Functor`, `Contravariant`, `Bifunctor`, `Profunctor`, ..) into a unified backend. Switching to a [categorical `Functor`](https://www.reddit.com/r/haskell/comments/eoo16m/base_category_polymorphic_functor_and_functorof/) would break just about every instance. Similarly if we wanted to `Generic`, `Generic1` to evolve to a class of arbitrary kind `GenericK`.

Users would still want to write the `Contravariant` or `Generic1` instance they wrote yesterday, but have it translate into `Functor` and `GenericK`:

```haskell
instance Contravariant C where
  contramap :: (a' -> a) -> (C a -> C a')
  contramap = ..
  
-->

instance Functor C where
  type Source C = Op
  type Target C = (->)
  
  fmap :: Op a a' -> (C a -> C a')
  fmap (Op pre) = .. pre
```

### DIFF
--- a/proposals/classbackend.md
+++ b/proposals/classbackend.md
@@ -1,0 +1,222 @@
+---
+author: Iceland_jack
+date-accepted: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/0>).
+**After creating the pull request, edit this file again, update the number in
+the link, and delete this bold sentence.**
+
+# Type Class Backend
+
+This proposes a language extension `-XClassBackend` which separates
+the interface of a type class (frontend) from the concrete
+representation (backend). The frontend itself is virtual and is
+immediately translated into the backend.
+
+This allows deriving classes that could previously not be derived
+because of their representation (notably `Traversable`) and enables us
+to evolve and generalise type classes without breaking backwards
+compatibility.
+
+Any type class may have at most one backend specified. When defining
+or deriving a class `Front` an instance of its backend `Back` is
+automatially produced through the same mechanism used by
+`DerivingVia`. The methods of `Front` are defined in terms of methods
+of `Back`. The aim is to use `Front` without needing to be aware of
+`Back`.
+
+```haskell
+class Front a where
+  front :: a -> a
+  front = .. back ..
+
+  deriving Back a
+  via W a
+```
+
+## Motivation
+The need for type class backends comes up in different ways: for
+purposes of presentation, representation, evolution and
+generalisation.
+
+### Presentation Argument
+It is common to have more than one way to present a type class.
+
+Some type classes have different but equivalent interfaces that may be
+useful under different circumstances. One may want to define
+`Applicative` as a "lax monoidal functor" without adding redundant
+methods. Instead a new type class frontend (`Monoidal`) is defined
+with `Applicative` as a backend.
+
+```haskell
+class Functor f => Monoidal f where
+  unit :: f ()
+  (**) :: f a -> f b -> f (a, b)
+
+  deriving Applicative f
+  via WrappedMonoidal f
+```
+
+The definition of `WrappedMonoidal` shows how `Monoidal` is translated into `Applicative`:
+
+```haskell
+type    WrappedMonoidal :: (Type -> Type) -> Type -> Type
+newtype WrappedMonoidal f a = WrapMonoidal (f a)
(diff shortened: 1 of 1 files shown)


Icelandjack:
A backend in a sense does not generate a new type class. Similarly when we write instances of either `Front` or `Back` only instances of `Back` are created. This follows something Ryan Scott said to my wanting to derive multiple classes at once (`hiding Show everything via ..`, and `deriving superclasses newtype MonoidIO`). It has worked as a sensible principle for design:

> I'm of the firm opinion that all instances should be opt-in.

I apologize for the informal state of the proposal, you may need to fill in the blanks a bit. The idea is to establish an isomorphism between a frontend and its backend. The via type is used to map to the backend, and the methods of the frontend can be default methods that fully characterise that interface in terms of the backend. I didn't want to make the user provide two via types.

blamario:
I find the proposal very confusing. Most of all, I can't figure out what specific changes it proposes to GHC, as opposed to `base` classes. Does it require new syntax, a new deriving mechanism, ....? These should be stated up front.

Icelandjack:
Yes I will be more clear about the motivation and proposed changes, I will think about it more and then revisit it

Ericson2314:
I am also confused, but i reminded me of an idea of mine to separate what a class *requires* of its instances vs what it *provides* to its users. Is it like that or not really?

konsumlamm:
So if I understand correctly, all this does is adding a new `deriving` instance whenever the typeclass is implemented? Then why introduce new syntax for it, instead of just writing out the `deriving` instance manually?

i-am-tom:
@Icelandjack are you still working on this proposal?

Icelandjack:
@i-am-tom It's one thing I want repeately, still thinking about it and figuring out a nicer interface.

@konsumlamm It identifies one class interface with another, it's not just saving a deriving clause but can change the representation of a type class in a backwards compatible way. When you derive for a frontend, in actuality you only get the instance for the backend. Same when you write an instance by hand. It allows us to take the `Traversable` type class that can't be derived and make it a frontend for a typeclass that *can be derived*. This would automatically make every `Traversable` instance into a fronend instance that is translated into the derivable `Traversable` backend.

nomeata:
@Icelandjack, are you by any chance attending Haskell Symposium? If so, would you like to present this proposal at the [GHC Proposals session](https://icfp22.sigplan.org/home/haskellsymp-2022#GHC-Proposals-Session)?

PR #475, opened 2021-12-22
New syntax for tuples and lists to remove punning

goldfirere:
The [proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0475-tuple-syntax.rst) has been accepted; the following discussion is mostly of historic interest.

This proposal is informed by and supersedes #458 as a way to remove the punned list and tuple syntax.

For clarity, it is presented as a proposal distinct from #281 (and modifies the accepted #281 to avoid mention of this feature). However, note that rejecting this proposal *still* introduces `-XListTupleTypeSyntax`, as proposed in #281. It's just that accepting this proposal adds some more convenience, more details, and it changes the module structure slightly.

[Rendered](https://github.com/goldfirere/ghc-proposals/blob/tuple-syntax/proposals/0000-tuple-syntax.rst)

### DIFF
--- a/proposals/0000-tuple-syntax.rst
+++ b/proposals/0000-tuple-syntax.rst
@@ -6,7 +6,7 @@ Non-punning list and tuple syntax
 .. ticket-url::
 .. implemented::
 .. highlight:: haskell
-.. header::
+.. header::  This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/475>`_.
 .. contents::
 .. sectnum::
 
--- a/proposals/0000-tuple-syntax.rst
+++ b/proposals/0000-tuple-syntax.rst
@@ -37,7 +37,11 @@ to ::
   just :: Maybe Int
   just = Maybe 5
 
-even though the latter is nonsense.
+even though the latter is nonsense. With this proposal, users would have the
+option of writing ::
+
+  single :: List Int
+  single = [5]
 
 Puns also cause trouble when describing what we are talking about: conversations
 around these features invariably require people to say "``[]``-the-type" or
@@ -50,7 +54,18 @@ over time.
 Proposed Change Specification
 -----------------------------
 
-1. Add the following definitions to a new module ``GHC.Tuple.Prim``::
+1. Create a new module in ``base`` called ``GHC.Prelude``. This module exports definitions
+   peculiar to GHC that are safe to use in ordinary code.
+   This module is *not* intended as a replacement to ``Prelude``,
+   but instead a complement to it. This module is a counterpart to ``GHC.Exts``, which exports
+   many *unsafe* internals.
+
+   This proposal lists several definitions to be exported from ``GHC.Prelude``, but leaves other
+   definitions to be added in separate proposal(s).
+
+#. Export the following definitions from ``GHC.Tuple`` and ``GHC.Prelude``. (Implementation note:
+   These would probably end up in a new module ``GHC.Tuple.Prim``, due to the dependency from
+   the definitions below in ``GHC.Tuple`` on ``GHC.TypeLits``.) ::
 
      data Unit = ()
      data Solo a = MkSolo a    -- this is a change from today's `data Solo a = Solo a`
@@ -62,8 +77,9 @@ Proposed Change Specification
      -- ...
      data Tuple64 ... = (...)
 
-#. Add the following definitions to ``GHC.Classes`` (re-exported from ``GHC.Exts``). These
-   replace similar definitions today::
+#. Export the following definitions from ``GHC.Exts`` and ``GHC.Prelude``. These
+   replace similar definitions today. (Note that ``(...) =>`` is special syntax, and does not
+   construct tuples. See more on this point `below <#constraint-special-syntax>`_.)::
 
      class CUnit
      instance CUnit
@@ -100,6 +116,12 @@ Proposed Change Specification
        TupleN @4 = Tuple4
        -- ...
        TupleN @64 = Tuple64
+       TupleN @n  = TypeError (ShowType n :<>: Text " is too large; the maximum size for a tuple is 64.")
+
+     type Length :: [a] -> Nat   -- not exported
+     type family Length xs where
+       Length []     = 0
+       Length (_:xs) = 1 + Length xs
 
      type Tuple :: List Type -> Type
      type family Tuple ts where
@@ -108,6 +130,7 @@ Proposed Change Specification
        Tuple [a,b] = Tuple2 a b
        -- ...
        Tuple [...] = Tuple64 ...
+       Tuple ts    = TypeError (ShowType (Length ts) :<>: Text " is too large; the maximum size for a tuple is 64.")
 
      type CTupleNKind :: Nat -> Type
      type family CTupleNKind n = r | r -> n where
@@ -123,6 +146,7 @@ Proposed Change Specification
        CTupleN @4  = CTuple4
        -- ...
        CTupleN @64 = CTuple64
+       CTupleN @n  = TypeError (ShowType n :<>: Text " is too large; the maximum size for a tuple is 64.")
 
      type Constraints :: List Constraint -> Constraint
      type family Constraints cs where
@@ -131,11 +155,25 @@ Proposed Change Specification
        Constraints [a,b] = CTuple2 a b
        -- ...
        Constraints [...] = CTuple64 ...
+       Constraints ts    = TypeError (ShowType (Length ts) :<>: Text " is too large; the maximum size for a tuple is 64.")
 
-#. Export the following pseudo-definitions from ``GHC.Prim``. Note that ``GHC.Prim`` defines
-   types that cannot be defined in Haskell, so we say that we just export these
-   from ``GHC.Prim``, not define them there. Note that ``GHC.Exts`` re-exports
-   ``GHC.Prim``::
+     type TupleNKind# :: [RuntimeRep] -> [RuntimeRep] -> Type
+     type family TupleNKind# all_reps reps_to_go = r | r -> all_reps reps_to_go where
+       TupleNKind# all_reps '[]                      = TYPE (TupleRep all_reps)
+       TupleNKind# all_reps (first_rep : reps_to_go) = TYPE first_rep -> TupleNKind# all_reps reps_to_go
+
+     type TupleN# :: forall (reps :: [RuntimeRep]). TupleNKind# reps reps
+     type family TupleN# where
+       TupleN# @[]                 = Unit#
+       TupleN# @[rep1]             = Solo#
+       TupleN# @[rep1, rep2]       = Tuple2#
+       TupleN# @[rep1, rep2, rep3] = Tuple3#
+       -- ...
+       TupleN# @[...]              = Tuple64#
+       TupleN# @reps               = TypeError (ShowType (Length reps) :<>: Text " is too large; the maximum size for a tuple is 64.")
+
+#. Export the following pseudo-definitions from ``GHC.Exts``. (Implementation note:
+   These would likely be exported from ``GHC.Prim`` originally.) ::
 
      type Unit# :: TYPE (TupleRep [])
      data Unit# = (# #)
@@ -171,79 +209,80 @@ Proposed Change Specification
 
      data List a = [] | a : List a
 
-#. Re-export ``List`` from ``GHC.List``.
+#. Re-export ``List`` from ``GHC.List`` and ``GHC.Prelude``.
 
-#. Introduce a new extension ``-XListTupleTypeSyntax``; this extension is on by default.
+#. Introduce a new extension ``-XListTupleTypeSyntax``; this extension is part
+   of ``-XHaskell98``, ``-XHaskell2010``, and ``-XGHC2021``. It is thus on by default.
 
 #. With ``-XListTupleTypeSyntax``:
 
    1. An occurrence of ``[]`` in type-syntax (as defined in `#378`_) is a synonym
-      for ``GHC.Types.List``.
+      for ``GHC.List.List``.
 
-   #. An occurrence of ``[ty]`` in type-syntax is a synonym for ``GHC.Types.List ty``.
+   #. An occurrence of ``[ty]`` in type-syntax is a synonym for ``GHC.List.List ty``.
 
    #. An occurrence of ``()`` in type-syntax, where the type is not expected to be of kind ``Constraint``,
-      is a synonym for ``GHC.Tuple.Prim.Unit``.
+      is a synonym for ``GHC.Tuple.Unit``.
 
    #. An occurrence of ``(,,...,,)`` where there are *n* commas (for *n* ≧ 1) in type-syntax
-      is a synonym for ``GHC.Tuple.Prim.Tuple``\ *n+1*.
+      is a synonym for ``GHC.Tuple.Tuple``\ *n+1*.
 
    #. An occurrence of ``(ty1,ty2,...,tyn-1,tyn)`` (for *n* ≧ 2) in type-syntax, where neither the type
       is expected to be of kind ``Constraint`` and either none of the ``tyi`` are inferred to have kind ``Constraint``
       or there exists a ``tyi`` inferred to kind ``Type`` and none of the ``tyj`` (with j < i) are inferred to have
       kind ``Constraint``, is
-      a synonym for ``GHC.Tuple.Prim.Tuple``\ *n* ``ty1 ty2 ... tyn-1 tyn``.
+      a synonym for ``GHC.Tuple.Tuple``\ *n* ``ty1 ty2 ... tyn-1 tyn``.
 
-   #. An occurrence of ``(# #)`` in type-syntax is a synonym for ``GHC.Prim.Unit#``.
+   #. With ``-XUnboxedTuples``, an occurrence of ``(# #)`` in type-syntax is a synonym for ``GHC.Exts.Unit#``.
 
-   #. An occurrence of ``(#,,...,,#)`` where there are *n* commas (for *n* ≧ 1) in type-syntax
-      is a synonym for ``GHC.Prim.Tuple``\ *n+1*\ ``#``.
+   #. With ``-XUnboxedTuples``, an occurrence of ``(#,,...,,#)`` where there are *n* commas (for *n* ≧ 1) in type-syntax
+      is a synonym for ``GHC.Exts.Tuple``\ *n+1*\ ``#``.
 
-   #. An occurrence of ``(# ty1, ty2, ... , tyn-1, tyn #)`` (for *n* ≧ 2) in type-syntax is a synonym
-      for ``GHC.Prim.Tuple``\ *n*\ ``# ty1 ty2 ... tyn-1 tyn``.
+   #. With ``-XUnboxedTuples``, an occurrence of ``(# ty1, ty2, ... , tyn-1, tyn #)`` (for *n* ≧ 2) in type-syntax is a synonym
+      for ``GHC.Exts.Tuple``\ *n*\ ``# ty1 ty2 ... tyn-1 tyn``.
 
-   #. An occurrence of ``(# | | ... | | #)`` where there are *n* pipes (for *n* ≧ 1) in type-syntax
-      is a synonym for ``GHC.Prim.Sum``\ *n+1*\ ``#``.
+   #. With ``-XUnboxedSums``, an occurrence of ``(# | | ... | | #)`` where there are *n* pipes (for *n* ≧ 1) in type-syntax
+      is a synonym for ``GHC.Exts.Sum``\ *n+1*\ ``#``.
 
-   #. An occurrence of ``(# ty1 | ty2 | ... | tyn-1 | tyn #)`` (for *n* ≧ 2) in type-syntax is a
-      synonym for ``GHC.Prim.Sum``\ *n*\ ``# ty1 ty2 ... tyn-1 tyn``.
+   #. With ``-XUnboxedSums``, an occurrence of ``(# ty1 | ty2 | ... | tyn-1 | tyn #)`` (for *n* ≧ 2) in type-syntax is a
+      synonym for ``GHC.Exts.Sum``\ *n*\ ``# ty1 ty2 ... tyn-1 tyn``.
 
    #. An occurrence of ``()`` in type-syntax, where the type is expected to be of kind ``Constraint``,
-      is a synonym for ``GHC.Classes.CUnit``.
+      is a synonym for ``GHC.Tuple.CUnit``.
 
    #. An occurrence of ``(ty1, ty2, ..., tyn-1, tyn)`` (for *n* ≧ 2) in type-syntax, where the type is
-      expected to be of kind ``Constraint``, is a synonym for ``GHC.Classes.CTuple``\ *n* ``ty1 ty2 ... tyn-1 tyn``.
+      expected to be of kind ``Constraint``, is a synonym for ``GHC.Tuple.CTuple``\ *n* ``ty1 ty2 ... tyn-1 tyn``.
 
    #. An occurrence of ``(ty1, ty2, ..., tyn-1, tyn)`` (for *n* ≧ 2) in type-syntax, where the first
       ``tyi`` inferred to have kind ``Type`` or ``Constraint`` has kind ``Constraint``, is a synonym
-      for ``GHC.Classes.CTuple``\ *n* ``ty1 ty2 ... tyn-1 tyn``.
+      for ``GHC.Tuple.CTuple``\ *n* ``ty1 ty2 ... tyn-1 tyn``.
 
-   #. An unapplied occurrence of ``GHC.Types.List`` is pretty-printed as ``[]``.
+   #. An unapplied occurrence of ``GHC.List.List`` is pretty-printed as ``[]``.
 
-   #. An occurrence of ``GHC.Types.List ty`` is pretty-printed as ``[ty]``.
+   #. An occurrence of ``GHC.List.List ty`` is pretty-printed as ``[ty]``.
 
-   #. An occurrence of ``GHC.Tuple.Prim.Unit`` is pretty-printed as ``()``.
+   #. An occurrence of ``GHC.Tuple.Unit`` is pretty-printed as ``()``.
 
-   #. An occurrence of ``GHC.Tuple.Prim.Tuplen ty1 ty2 ... tyn`` is pretty-printed as ``(ty1, ty2, ..., tyn)``.
+   #. An occurrence of ``GHC.Tuple.Tuplen ty1 ty2 ... tyn`` is pretty-printed as ``(ty1, ty2, ..., tyn)``.
 
-   #. An occurrence of ``GHC.Tuple.Prim.Tuplen``, but not applied to a full *n* arguments, is pretty-printed as ``(,,...,,)``,
+   #. An occurrence of ``GHC.Tuple.Tuplen``, but not applied to a full *n* arguments, is pretty-printed as ``(,,...,,)``,
       where there are *n-1* commas.
 
-   #. An occurrence of ``GHC.Prim.Unit#`` is pretty-printed as ``(# #)``.
+   #. An occurrence of ``GHC.Exts.Unit#`` is pretty-printed as ``(# #)``.
 
-   #. An occurrence of ``GHC.Prim.Tuplen# ty1 ty2 ... tyn`` is pretty-printed as ``(# ty1, ty2, ..., tyn #)``.
+   #. An occurrence of ``GHC.Exts.Tuplen# ty1 ty2 ... tyn`` is pretty-printed as ``(# ty1, ty2, ..., tyn #)``.
 
-   #. An occurrence of ``GHC.Prim.Tuplen#``, but not applied to a full *n* arguments, is pretty-printed as ``(#,,...,,#)``,
+   #. An occurrence of ``GHC.Exts.Tuplen#``, but not applied to a full *n* arguments, is pretty-printed as ``(#,,...,,#)``,
       where there are *n-1* commas.
 
-   #. An occurrence of ``GHC.Prim.Sumn# ty1 ty2 ... tyn`` is pretty-printed as ``(# ty1 | ty2 | ... | tyn #)``.
+   #. An occurrence of ``GHC.Exts.Sumn# ty1 ty2 ... tyn`` is pretty-printed as ``(# ty1 | ty2 | ... | tyn #)``.
 
-   #. An occurrence of ``GHC.Prim.Sumn#``, but not applied to a full *n* arguments, is pretty-printed as ``(# | | ... | | #)``,
+   #. An occurrence of ``GHC.Exts.Sumn#``, but not applied to a full *n* arguments, is pretty-printed as ``(# | | ... | | #)``,
       where there are *n-1* pipes.
 
-   #. An occurrence of ``GHC.Classes.CUnit`` is pretty-printed as ``()``.
+   #. An occurrence of ``GHC.Tuple.CUnit`` is pretty-printed as ``()``.
 
-   #. An occurrence of ``GHC.Classes.CTuplen ty1 ty2 ... tyn`` is pretty-printed as ``(ty1, ty2, ..., tyn)``.
+   #. An occurrence of ``GHC.Tuple.CTuplen ty1 ty2 ... tyn`` is pretty-printed as ``(ty1, ty2, ..., tyn)``.
 
 #. With ``-XNoListTupleTypeSyntax``:
 
@@ -252,9 +291,9 @@ Proposed Change Specification
       never types or type constructors. (Note that ``(...) =>`` is special syntax, not an occurrence of any of the types
       listed above. See `below <#constraints-special-syntax>`_.)
 
-   #. An occurrence of ``GHC.Tuple.Prim.Tuplen ty1 ty2 ... tyn`` is pretty-printed as ``Tuple [ty1, ty2, ..., tyn]``.
+   #. An occurrence of ``GHC.Tuple.Tuplen ty1 ty2 ... tyn`` is pretty-printed as ``Tuple [ty1, ty2, ..., tyn]``.
 
-   #. An occurrence of ``GHC.Classes.CTuplen ty1 ty2 ... tyn`` is pretty-printed as ``Constraints [ty1, ty2, ..., tyn]``.
+   #. An occurrence of ``GHC.Tuple.CTuplen ty1 ty2 ... tyn`` is pretty-printed as ``Constraints [ty1, ty2, ..., tyn]``.
 
 Effect and Interactions
 -----------------------
@@ -272,14 +311,16 @@ Effect and Interactions
 
    .. _constraints-special-syntax:
 
-#. Note that the type syntax ``(ty1, ty2, ..., tyn) => ...`` is special syntax. The parser does *not*
+#. Note that the type syntax ``(ty1, ty2, ..., tyn) => ...`` is already special syntax. The parser does *not*
    parse a type to the left of the ``=>``. This syntax thus remains completely unaffected by ``-XListTupleTypeSyntax``
    and will continue to work with ``-XNoListTupleTypeSyntax``. Furthermore, because a type like ``(ty1, ty2, ... tyn) => ...``
    does not contain any uses of ``CTuplen``, it will also continue to pretty-print just as today.
 
    On the other hand, collections of constraints occurring not to the left of a ``=>`` are affected by
    this proposal, for example in ``Dict (Eq a, Show b)`` (which would be written ``Dict (Constraints [Eq a, Show b])``
-   under this proposal).
+   under this proposal). Another example is ``(Eq a, (Show a, Read a)) => a -> a``, which would not
+   be accepted under ``-XNoListTupleTypeSyntax``. Instead, the user should flatten the constraints or
+   write ``(Eq a, Constraints [Show a, Read a]) => a -> a``.
 
 #. An instance declaration like ``instance (C a, C b) => C (Tuple [a, b]) where ...`` would be
    rejected because it uses a type family in the instance head. We might choose to relax
@@ -293,7 +334,7 @@ Effect and Interactions
    above from modules not in the ``GHC.`` namespace, perhaps even including the
    ``Prelude``. This proposal does *not* make any such suggestions, and it does *not*
    depend on any such ideas being adopted in the future. Any such idea would
-   be evaluated by the Core Library Committee independently of this proposal.
+   be evaluated by the Core Libraries Committee independently of this proposal.
 
 Costs and Drawbacks
 -------------------
@@ -322,6 +363,8 @@ Alternatives
 #. Instead of introducing new names, we could use more mixfix bits of punctuation,
    such as ``(~ ty1, ty2 ~)`` for normal tuples and ``(% ty1, ty2 %)`` for constraint
    tuples. This was not as popular in a recent `straw poll <https://github.com/ghc-proposals/ghc-proposals/pull/458#issuecomment-982230541>`_.
+
+#. The name ``-XListTupleTypeSyntax`` is a mouthful. Maybe ``-XListTuplePuns`` would be better.
 
 Unresolved Questions
 --------------------
--- a/proposals/0000-tuple-syntax.rst
+++ b/proposals/0000-tuple-syntax.rst
@@ -218,7 +218,7 @@ Proposed Change Specification
 
    #. An unapplied occurrence of ``GHC.Types.List`` is pretty-printed as ``[]``.
 
-   #. An occurrence of ``GHC.Types.List ty`` is pretty-printed as ``[ty]]`.
+   #. An occurrence of ``GHC.Types.List ty`` is pretty-printed as ``[ty]``.
 
    #. An occurrence of ``GHC.Tuple.Prim.Unit`` is pretty-printed as ``()``.
 
--- a/proposals/0000-tuple-syntax.rst
+++ b/proposals/0000-tuple-syntax.rst
@@ -15,6 +15,9 @@ to write pun-free code. In addition, add a new extension that disables
 the built-in syntax for these types, thus requiring users to use the new
 names.
 
+Everything in this proposal is fully backward-compatible (except for the deprecation
+on the data constructor ``Solo``, but that has a migration period).
+
 .. _`#270`: https://github.com/ghc-proposals/ghc-proposals/pull/270
 .. _`#281`: https://github.com/ghc-proposals/ghc-proposals/pull/281
 .. _`#378`: https://github.com/ghc-proposals/ghc-proposals/pull/378
@@ -51,6 +54,12 @@ This proposal thus allows for a common way to refer to tuples and lists without
 punning. It is entirely opt-in, though we expect the new phrasing to become popular
 over time.
 
+Notation
+--------
+This proposal uses angle-bracket syntax ``<n>`` to denote the occurrence of a numeric
+variable in a bit of code. This allows us to e.g. refer to ``Tuple2`` and ``Tuple3`` with
+``Tuple<n>``.
+
 Proposed Change Specification
 -----------------------------
 
@@ -86,7 +95,7 @@ Proposed Change Specification
      data Tuple64 ... = (...)
 
 #. Export the following definitions from ``GHC.Exts`` and ``GHC.Prelude``. These
-   replace similar definitions (in ``GHC.Classes``) today. (Note that ``(...) =>`` is special syntax, and does not
+   the existing tuple definitions (in ``GHC.Classes``) today. (Note that ``(...) =>`` is special syntax, and does not
    construct tuples. See more on this point `below <#constraint-special-syntax>`_.)::
 
      class CUnit
@@ -126,7 +135,7 @@ Proposed Change Specification
        TupleN @64 = Tuple64
        TupleN @n  = TypeError (ShowType n :<>: Text " is too large; the maximum size for a tuple is 64.")
 
-     type Length :: [a] -> Nat   -- not exported
+     type Length :: List a -> Nat   -- not exported
      type family Length xs where
        Length []     = 0
        Length (_:xs) = 1 + Length xs
@@ -165,12 +174,12 @@ Proposed Change Specification
        Constraints [...] = CTuple64 ...
        Constraints ts    = TypeError (ShowType (Length ts) :<>: Text " is too large; the maximum size for a tuple is 64.")
 
-     type TupleNKind# :: [RuntimeRep] -> [RuntimeRep] -> Type
+     type TupleNKind# :: List RuntimeRep -> List RuntimeRep -> Type
      type family TupleNKind# all_reps reps_to_go = r | r -> all_reps reps_to_go where
        TupleNKind# all_reps '[]                      = TYPE (TupleRep all_reps)
        TupleNKind# all_reps (first_rep : reps_to_go) = TYPE first_rep -> TupleNKind# all_reps reps_to_go
 
-     type TupleN# :: forall (reps :: [RuntimeRep]). TupleNKind# reps reps
+     type TupleN# :: forall (reps :: List RuntimeRep). TupleNKind# reps reps
      type family TupleN# where
        TupleN# @[]                 = Unit#
        TupleN# @[rep1]             = Solo#
@@ -180,12 +189,12 @@ Proposed Change Specification
        TupleN# @[...]              = Tuple64#
        TupleN# @reps               = TypeError (ShowType (Length reps) :<>: Text " is too large; the maximum size for a tuple is 64.")
 
-     type SumNKind# :: [RuntimeRep] -> [RuntimeRep] -> Type
+     type SumNKind# :: List RuntimeRep -> List RuntimeRep -> Type
      type family SumNKind# all_reps reps_to_go = r | r -> all_reps reps_to_go where
       SumNKind# all_reps '[]                      = TYPE (SumRep all_reps)
       SumNKind# all_reps (first_rep : reps_to_go) = TYPE first_rep -> SumNKind# all_reps reps_to_go
 
-     type SumN# :: forall (reps :: [RuntimeRep]). SumNKind# reps reps
+     type SumN# :: forall (reps :: List RuntimeRep). SumNKind# reps reps
      type family SumN# where
       SumN# @[]                 = TypeError (Text "GHC does not support empty unboxed sums. Consider Data.Void.Void instead.")
       SumN# @[rep1]             = TypeError (Text "GHC does not support unary unboxed sums. Consider Data.Tuple.Solo# instead.")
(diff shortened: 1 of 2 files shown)


Ericson2314:
Great proposal!

For the effects and interaction note about `(...) =>`, syntax, maybe spell out a bit more strongly that it is already a special case? E.g. in `ghc-prim` today, as I just learened, we have
```haskell
class (c1, c2) => (c1, c2)
```
which looks like a nonsense impredicative definition.

Coincidentally, I opened https://gitlab.haskell.org/ghc/ghc/-/issues/20420 a few months back to jot down an idea for a superclass constraints. As it happens, this would also remove the need for the `(....) =>` hack since superclasses wouldn't be written in a tuply way.

I think this proposal is exactly right, and we shouldn't worry about `(....) =>` being weird since it already was, but if anyone is uncomfortable with that fact being more obvious to `-XNoListTupleTypeSyntax` user than it was before, I think the right answer is not to second guess the design here, but write a proposal for that alternative superclass syntax.

proposals/0281-visible-forall.rst
```diff
@@ -625,57 +625,10 @@ Syntax
    This change applies to ``∀`` (the ``UnicodeSyntax`` rendition of ``forall``)
    as well.
 
-4. Introduce a new extension, ``ListTupleTypeSyntax``, on by default,
```
int-index, line 628: Could you also amend the “Corner Cases” section that still talks about `TupleOf2`?

proposals/0000-tuple-syntax.rst
```diff
@@ -0,0 +1,324 @@
...
+     instance (...) => CTuple64 ...
+
+#. Remove existing tuple definitions from ``GHC.Tuple``.
+
+#. Add the following definitions to ``GHC.Tuple``, which re-exports all of ``GHC.Tuple.Prim``::
+
+     type TupleNKind :: Nat -> Type     -- Nat is from GHC.TypeLits
+     type family TupleNKind n = r | r -> n where
+       TupleNKind 0 = Type
+       TupleNKind n = Type -> TupleNKind (n-1)   -- this fails the injectivity check, but a little magic will allow this
```
int-index, line 92: Do you envision the magic to be specific to `TupleNKind` or all definition that try to use `Nat` inductively?

nomeata:
Initial gut feeling is that having 3 different ways to write what's now `(a, b)` feels unhelpful, and as if we couldn't make up our mind. What do we gain from having `Tuple 2` over only `Tuple2`? And what from `Constraints [c1, c2]` over `CTuple2 c1 c2` (besides less parentheses if the elements are complex expressions)?

proposals/0000-tuple-syntax.rst
```diff
@@ -0,0 +1,324 @@
...
+   #. An occurrence of ``(ty1, ty2, ..., tyn-1, tyn)`` (for *n* ≧ 2) in type-syntax, where the type is
+      expected to be of kind ``Constraint``, is a synonym for ``GHC.Classes.CTuple``\ *n*\ `` ty1 ty2 ... tyn-1 tyn``.
+
+   #. An occurrence of ``(ty1, ty2, ..., tyn-1, tyn)`` (for *n* ≧ 2) in type-syntax, where the first
+      ``tyi`` inferred to have kind ``Type`` or ``Constraint`` has kind ``Constraint``, is a synonym
+      for ``GHC.Classes.CTuple``\ *n*\ `` ty1 ty2 ... tyn-1 tyn``.
+
+   #. An unapplied occurrence of ``GHC.Types.List`` is pretty-printed as ``[]``.
+
+   #. An occurrence of ``GHC.Types.List ty`` is pretty-printed as ``[ty]]`.
```
int-index, line 221:
```suggestion
   #. An occurrence of ``GHC.Types.List ty`` is pretty-printed as ``[[ty]]``.
```

proposals/0281-visible-forall.rst
```diff
@@ -625,57 +625,10 @@ Syntax
...
-     type List = []
-
-   Export the following synonym from the ``Data.Tuple`` module::
-
-     type Unit = ()
-
-   For tuples of higher arities, users may want to define their own
-   synonyms, such as::
-
-     type TupleOf2 = (,)
```
int-index, line 662: The “Corner Cases” section that still talks about `TupleOf2`.

proposals/0000-tuple-syntax.rst
goldfirere: I think we should do what's necessary to get TupleNKind to compile. If we can do so generally, then all the better.

Right now, GHC has some built-in magic so that, if `n0-1 ~ m` and `n1-1 ~ m`, then GHC will know that `n0 ~ n1`. Note that this magic isn't available to non-built-in functions: it's using the fact that both calls to `-` take the same second argument. But my guess is that this particular existent magic isn't consulted during the injectivity check. So, either we consult it during the injectivity check, or we whip up some custom magic for `TupleNKind`. The former is preferable, but the latter is acceptable. But I don't wish to hold up this proposal on, say, a specification of exactly how to do this.

goldfirere:
`TupleN` was the most upvoted choice in the [straw poll](https://github.com/ghc-proposals/ghc-proposals/pull/458#issuecomment-982230541), so I thought that it would be nice here. And note that it's not `TupleN 2`, but just `TupleN`. That is, type inference will happily accept e.g. `snd :: TupleN a b -> b`. So the advantage is that you don't have to count the size of your tuple: you just list the components.

I agree that having both `Tuple` and `TupleN` (resp. `Constraints` and `CTupleN`) may be overkill. The committee should feel free to eliminate one.

goldfirere:
@Ericson2314 I don't see how to make the point more strongly than I have. But maybe the point came too late, and you were already nervous? I've added a forward reference from the proposal spec to this note. Feel free to suggest alternate concrete text.

angerman:
I think it would be helpful for simpletons like me of the Motivation, not only showed what we currently have, but also how those examples would look with this proposal implemented.

Instead of
```haskell
x :: [T]
x = [a,b,c]
```

We'd write
```
x :: List T
x = [a,b,c]
```
?

goldfirere:
Conversation with my colleague @monoidal today led to a few insights:

* It is annoying to have to import e.g. `Data.Kind`, `GHC.List`, and `GHC.Tuple` to write unpunned code. Maybe we should have a `GHC.Prelude` module? That's a bit like `GHC.Exts`, but I want one that is much tamer than `GHC.Exts` is.

* It is too low-level to specify e.g. that some symbols come from `GHC.Prim`. That is an implementation detail. Instead, the proposal should just specify what user-importable module a symbol is exported from. Indeed, we may want a principle around this.

* There should be a fall-through case in the type families calling `TypeError`.

* The proposal should note the need for `-XUnboxedTuples` and `-XUnboxedSums` where they arise.

* Example or comment around how e.g. `(Eq a, (Show a, Ord a)) => ...` will not work with `-XNoListTupleTypeSyntax`.

* Would a better extension name be `-XListTuplePuns`?

proposals/0000-tuple-syntax.rst
```diff
@@ -0,0 +1,329 @@
...
+   today. The exception is around unsaturated ``CTuplen``, which is not handled above. It is hard to have
+   an unsaturated constraint tuple, but possible by the use of a type family that decomposes one. Today's
+   GHC prints out e.g. ``ghc-prim-0.6.1:GHC.Classes.(%,%)``. Switching to ``GHC.Classes.CTuple2`` (which is
+   actually parseable) seems a positive improvement.
+
+#. With the definitions above, users can avoid puns in their lists and tuples.
+
+   .. _constraints-special-syntax:
+
+#. Note that the type syntax ``(ty1, ty2, ..., tyn) => ...`` is special syntax. The parser does *not*
```
Ericson2314, line 275: Oh I meant something simple like
```suggestion
#. Note that the type syntax ``(ty1, ty2, ..., tyn) => ...`` is already special syntax. The parser does *not*
```

leoconst:
Hi, first time ever commenting here (or indeed anywhere on GitHub!) so please let me know if I've wondered into the wrong place 😃.

This is regarding the naming (gosh I hope this isn't bikeshedding) of the new tuple type families. To reveal my bias, I'd love if `TupleN` was renamed to `Tuple` as that'd be my preferred use of an unpunned tuple in real code. That'd of course necessitate coming up with a good new name for the list-style tuple (the one currently named `Tuple`).

So, based on these selfish desires, I see 3 ways forwards:
1. Favours variadic-style - Rename `TupleN` -> `Tuple`, and `Tuple` -> something else.
2. Neutral - Leave `TupleN` as it is, and rename `Tuple` -> something else. This leaves us the ability to rename/alias the community-preferred tuple syntax to have the nice `Tuple` name with another proposal in the future.
3. Favours list-style - Leave the proposal as it is.

I think leaving asside my personal leaning towards option 1, I think option 2 is the most pragmatic, again relying on coming up with a good name for the list-style tuple. If there's no acceptable name for it, I'm very happy with option 3 - this is a fantastic proposal and a very interesting piece of dependant code and I certainly don't want to entirely halt progress on Dependant Haskell just for the sake of not typing `N` when I use tuples 😅 

Note: Anything that applies to to the boxed types I mentioned here of course apply to the unboxed variant too, that is `TupleN#` (I don't see a list-style `Tuple#` in the proposal or anything named as such on Hoogle, though there's probably a good reason for that I'm not privy to).

goldfirere:
Thanks for your contribution to the conversation, @leoconst!

I had not considered the idea of not using `Tuple` at all, saving it for later. My first impression is to lean against, just because once we release this, people will start using it. Then, when we introduce `Tuple` later, we will likely want to keep whatever we were using during the interim period. So we will end up with two names for the same concept, likely for a long time. This proposal already describes having 4 names for the same concept (`(Int,Bool)`, `Tuple2 Int Bool`, `Tuple [Int, Bool]`, and `TupleN Int Bool`), and so adding another seems bad. (Indeed, the 4 proposed here already may be too many.)

One other way to get parity between the list approach and the variadic approach is to give them both primary-sounding names, like `Tuple` and `Product`. Neither sounds more important than the other (to me). The downside here is that it's not obvious both names reduce to the same underlying thing.

Yet another possibility is not to have `Tuple` or `TupleN` at all. These definitions are possible outside of GHC (injectivity magic notwithstanding, but I bet we can add a little smarts to the injectivity test to get it to work) and so we could just leave it to libraries. The problem here is that tupling is so universal, and it's awkward to have competing names out in libraries.

I wish I saw a clearly-good answer here!

About `Tuple#`: the different components of the tuple (e.g. `Int#`, `Tree Integer`, further tuples) all have different kinds, and so we can't put them in a list. I was pleased I could get it to work with `TupleN#`, but until we have overloaded lists in types and a good solution for heterogeneous lists, `Tuple#` is off the table.

jvanbruegge:
I like the idea of naming `TupleN` to just `Tuple` and the current `Tuple` family to `Product`. On the term level, the `product` function also takes a list of arguments, I like the symmetry. And I agree, they should not be part of external libraries.

monoidal:
* > 11. With -XNoListTupleTypeSyntax:
  > i. Uses of [], [...], (), (,,...,,), (...,...,...), (# #), (#,,...,,#), (# ...,...,... #), (# | | ... | | #), and (# ... | ... | ... #) are now unambiguous. They always refer to data constructors, never types or type constructors.

  This suggests that `(# | | ... | | #)` and `(# ... | ... | ... #)` were ambiguous previously, but I believe there is no collision: the former is always a value (like `Left`), the latter is always a type (like `Either`). The extension as written disables the `(# ... | ... | ... #)` syntax rather than making it unambiguous. In theory, we don't have to change sums to get rid of punning, this is done only for consistency.
* If we're going to have `type family TupleN#` there should be also `SumN#`.
* Personally I'm not a fan of `TupleN` with an invisible parameter. Imagine having the same construct on the value level, a function `tuple` such that `tuple 1 2` is `(1,2)` but `tuple 1 2 3` is `(1,2,3)`. I'm also worried that inferring the length parameter would involve GHC going through 65 equations every time someone uses a tuple.
* If `NoListTupleTypeSyntax` is on, should `type T = '[]` be rejected?
* Renaming `Solo` to `MkSolo` is not backwards compatible. Is the plan to do this under a flag? That'd be too magical for me. I'd rather add a deprecated pattern synonym `Solo` for the transition period.
* `GHC.Prelude`: are there any alternative names? It's rather confusing to have `Prelude` and `GHC.Prelude` doing different things. Furthermore, the name `GHC.Prelude` is already used internally in GHC.

int-index:
@goldfirere What's the status here?

goldfirere:
> If `NoListTupleTypeSyntax` is on, should `type T = '[]` be rejected?

Why would that be rejected? Are you worried about the `'`? Dealing with that is the subject of another proposal (#270 I believe).

I've incorporated your other (very helpful) suggestions. And I have decided to recommend `-XListTuplePuns` as the extension name.

goldfirere:
I'd like to submit this to the committee, @nomeata. Thanks!

simonpj:
> An occurrence of GHC.Tuple.Tuplen ty1 ty2 ... tyn is pretty-printed as Tuple [ty1, ty2, ..., tyn].

I could not work out whether I can write `Tuple [Int, Bool]`  or `CTuple [Ord a, Show a]`  in my program and if so what flags are needed.  Where is this laid out?  Surely if it's pretty-printed that way I can write it that way!   (And is `Tuple` a keyword there, or some kind of built-in type family?)

simonpj:
Oh I found it -- it's a type family buried in the middle of point (5) of Section 2.   I wonder if it could be brought out more clearly.

One naming thing.  Do we want `Tuple` and `CTuple` or `Tuple` and `Constraints`? I lean to the former

proposals/0000-tuple-syntax.rst
```diff
@@ -0,0 +1,414 @@
...
+   be accepted under ``-XNoListTuplePuns``. Instead, the user should flatten the constraints or
+   write ``(Eq a, Constraints [Show a, Read a]) => a -> a``.
+
+#. An instance declaration like ``instance (C a, C b) => C (Tuple [a, b]) where ...`` would be
+   rejected because it uses a type family in the instance head. We might choose to relax
+   this restriction, by allowing type families in an instance head, as long as they can
+   reduce to a ground (i.e. type-family-free) type. This proposal does *not* include such
+   a lifting of the restriction, as the workaround is straightforward: just write
+   ``instance (C a, C b) => C (Tuple2 a b) where ...``. Still, we may decide to revisit
+   this in the future.
```
aspiwack, line 355: Alternatively, you don't define `Tuple2` and such (at least not primitively), and make `Tuple ts` a data family instead. Would this work?

proposals/0000-tuple-syntax.rst
```diff
@@ -0,0 +1,414 @@
...
+   thus think carefully about how to present these names to beginners, if
+   ``-XNoListTuplePuns`` catches on.
+
+Alternatives
+------------
+1. Instead of defining ``TupleN`` as a type family (as done here), it could be
+   a data family, effectively replacing the ``Tuple2``, ``Tuple3``, ..., definitions.
+   This design would seem to be too complicated to be the primitive definition
+   of tuples, however, when a very vanilla datatype like ``data Tuple2 a b = (a, b)``
+   suffices.
```
aspiwack, line 389: Ha :slightly_smiling_face: .

Still, I think that `Tuple`, is a more reasonable candidate to be made the primitive type than `TupleN`.

proposals/0000-tuple-syntax.rst
```diff
@@ -0,0 +1,414 @@
...
+
+#. Instead of introducing new names, we could use more mixfix bits of punctuation,
+   such as ``(~ ty1, ty2 ~)`` for normal tuples and ``(% ty1, ty2 %)`` for constraint
+   tuples. This was not as popular in a recent `straw poll <https://github.com/ghc-proposals/ghc-proposals/pull/458#issuecomment-982230541>`_.
```
aspiwack, line 393: It's worth noting that MLs do something like this. It's a bit of a weird syntax, though, as `a1 * a2 * … * an` is a single n-ary type constructor (so that, famously, `a * b * c` is not the same type as `a * (b * c)`). No solution is perfect. But I think that it ought to be mentioned in the proposal for completeness.

proposals/0000-tuple-syntax.rst
```diff
@@ -0,0 +1,414 @@
...
+
+#. Controlling the ``(# ... | ... | ... #)`` syntax for unboxed sum types with
+   ``-XNoListTuplePuns`` is not necessary to avoid punning, but is done only for
+   consistency. We could skip this, but I prefer keeping it as proposed.
```
aspiwack, line 397: This is an interesting case. It's not necessary, so if the choice is given, would one prefer using `(# a | b | c #)` or `Sum# [a, b, c]`? Would removing this syntax be considered an anti-feature?

If programmers prefer the former, then it raises an interesting question. It is arguable that unboxed sum are niche enough a feature not to deserve taking up syntactic real estate. However, tuples definitely are. So it would argue that the alternative, above, proposing a special syntax for tuples may have more weight to it than it first seems.

proposals/0000-tuple-syntax.rst
```diff
@@ -0,0 +1,414 @@
...
+
+#. There was an objection in the commentary about the name ``GHC.Prelude``. I continue
```
aspiwack, line 399: Ha again :slightly_smiling_face:

proposals/0000-tuple-syntax.rst
```diff
@@ -0,0 +1,414 @@
...
+
+     class (a, b) => CTuple2 a b
+     instance (a, b) => CTuple2 a b
+
+     -- ...
+
+     class (...) => CTuple64 ...
+     instance (...) => CTuple64 ...
+
+#. Remove existing tuple definitions from ``GHC.Tuple``.
```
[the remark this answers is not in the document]
goldfirere: No, given that, with `-XListTuplePuns`, any type-syntax use of `(...,...)` gets translated to use the new tuples.

proposals/0000-tuple-syntax.rst
```diff
@@ -0,0 +1,414 @@
...
+      is a synonym for ``GHC.Tuple.Unit``.
+
+   #. An occurrence of ``(,,...,,)`` where there are *n* commas (for *n* ≧ 1) in type-syntax
+      is a synonym for ``GHC.Tuple.Tuple``\ *n+1*.
+
+   #. An occurrence of ``(ty1,ty2,...,tyn-1,tyn)`` (for *n* ≧ 2) in type-syntax, where neither the type
+      is expected to be of kind ``Constraint`` and either none of the ``tyi`` are inferred to have kind ``Constraint``
+      or there exists a ``tyi`` inferred to kind ``Type`` and none of the ``tyj`` (with j < i) are inferred to have
+      kind ``Constraint``, is
+      a synonym for ``GHC.Tuple.Tuple``\ *n* ``ty1 ty2 ... tyn-1 tyn``.
```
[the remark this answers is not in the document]
goldfirere: Yes... but I don't quite know how else to get my point across. Maybe `Tuple<n>`? Would that be an improvement?

proposals/0000-tuple-syntax.rst
```diff
@@ -0,0 +1,414 @@
...
+      is a synonym for ``GHC.Exts.Sum``\ *n+1*\ ``#``.
+
+   #. With ``-XUnboxedSums``, an occurrence of ``(# ty1 | ty2 | ... | tyn-1 | tyn #)`` (for *n* ≧ 2) in type-syntax is a
+      synonym for ``GHC.Exts.Sum``\ *n*\ ``# ty1 ty2 ... tyn-1 tyn``.
+
+   #. An occurrence of ``()`` in type-syntax, where the type is expected to be of kind ``Constraint``,
+      is a synonym for ``GHC.Tuple.CUnit``.
+
+   #. An occurrence of ``(ty1, ty2, ..., tyn-1, tyn)`` (for *n* ≧ 2) in type-syntax, where the type is
+      expected to be of kind ``Constraint``, is a synonym for ``GHC.Tuple.CTuple``\ *n* ``ty1 ty2 ... tyn-1 tyn``.
```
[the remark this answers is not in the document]
goldfirere: It is what we do today. It *is* scary. But my goal here is not to de-scarify this aspect of GHC -- just to change concrete syntax.

proposals/0000-tuple-syntax.rst
goldfirere: No, because you can't translate what is now `instance Functor ((,) a)`.

proposals/0000-tuple-syntax.rst
goldfirere: I'd be quite happy with special syntax for tuples. But this seems to be the minority position, if https://github.com/ghc-proposals/ghc-proposals/pull/458#issuecomment-982230541 is to be believed.

proposals/0000-tuple-syntax.rst
```diff
@@ -0,0 +1,440 @@
...
+
+     type Tuple0 = Unit
+     type Tuple1 = Solo
+     data Tuple2 a b = (a, b)
+     data Tuple3 a b c = (a, b, c)
+     -- ...
+     data Tuple64 ... = (...)
+
+#. Export the following definitions from ``GHC.Exts`` and ``GHC.Prelude``. These
+   replace similar definitions (in ``GHC.Classes``) today. (Note that ``(...) =>`` is special syntax, and does not
```
aspiwack, line 89: This still sounded a bit mysterious to me. So, for anybody who is not in the know, today (as of Long Ago™), `GHC.Classes` hosts the following class definitions (which are not exported, hence not present in the Haddock). It would be helpful to any future person reading the text if this was explained in the proposal.

```haskell
class ()                                                                                                                                                                                                                                      
class (c1, c2)     => (c1, c2)                                                                                                                                                                                                                
class (c1, c2, c3) => (c1, c2, c3)                                                                                                                                                                                                            
class (c1, c2, c3, c4) => (c1, c2, c3, c4)                                                                                                                                                                                                    
class (c1, c2, c3, c4, c5) => (c1, c2, c3, c4, c5)                                                                                                                                                                                            
class (c1, c2, c3, c4, c5, c6) => (c1, c2, c3, c4, c5, c6)                                                                                                                                                                                    
class (c1, c2, c3, c4, c5, c6, c7) => (c1, c2, c3, c4, c5, c6, c7) 
…
class (c1, c2, c3, c4, c5, c6, c7, c8, c9, c10, c11, c12, c13, c14, c15, c16,                                                                                                                                                                 
       c17, c18, c19, c20, c21, c22, c23, c24, c25, c26, c27, c28, c29, c30,                                                                                                                                                                  
       c31, c32, c33, c34, c35, c36, c37, c38, c39, c40, c41, c42, c43, c44,                                                                                                                                                                  
       c45, c46, c47, c48, c49, c50, c51, c52, c53, c54, c55, c56, c57, c58,                                                                                                                                                                  
       c59, c60, c61, c62, c63, c64)                                                                                                                                                                                                          
   => (c1, c2, c3, c4, c5, c6, c7, c8, c9, c10, c11, c12, c13, c14, c15, c16,                                                                                                                                                                 
       c17, c18, c19, c20, c21, c22, c23, c24, c25, c26, c27, c28, c29, c30,                                                                                                                                                                  
       c31, c32, c33, c34, c35, c36, c37, c38, c39, c40, c41, c42, c43, c44,                                                                                                                                                                  
       c45, c46, c47, c48, c49, c50, c51, c52, c53, c54, c55, c56, c57, c58,                                                                                                                                                                  
       c59, c60, c61, c62, c63, c64) 
```

proposals/0000-tuple-syntax.rst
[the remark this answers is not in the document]
aspiwack: I certainly don't intend for you to de-scarify anything :slightly_smiling_face: . Though maybe you want to remind the reader that GHC already does this (because tuples of `Constraint`-s and tuples of `Type`-s are already two different type constructors) and it's merely a syntax change.

proposals/0000-tuple-syntax.rst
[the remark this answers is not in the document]
aspiwack: I'd certainly understand `GHC.Tuple.Tuple<n> ty1 ty2 ... tyn-1 tyn` myself. I'm not sure how universal the notation is, though.

proposals/0000-tuple-syntax.rst
```diff
@@ -0,0 +1,414 @@
...
+over time.
+
+Proposed Change Specification
+-----------------------------
+
+1. Create a new module in ``base`` called ``GHC.Prelude``. This module exports definitions
+   peculiar to GHC that are safe to use in ordinary code.
+   This module is *not* intended as a replacement to ``Prelude``,
+   but instead a complement to it. This module is a counterpart to ``GHC.Exts``, which exports
+   many *unsafe* internals.
```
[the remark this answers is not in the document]
aspiwack: No, I agree with you. Well, I'd still call this “be imported by default”, if the recommendation is to have it at the top of every file (as opposed to “imported implicitly” for the current `Prelude`). I'm re-reading your sentence, and I'm not sure what I understood yesterday as I read it. So… well… this was noise. Sorry.

proposals/0000-tuple-syntax.rst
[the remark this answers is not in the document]
goldfirere: So noted.

proposals/0000-tuple-syntax.rst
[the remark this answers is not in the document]
goldfirere: OK. I agree this is an improvement.

proposals/0000-tuple-syntax.rst
goldfirere: I've tried to clarify this.

proposals/0000-tuple-syntax.rst
```diff
@@ -0,0 +1,440 @@
...
+       TupleN @0 = Unit
+       TupleN @1 = Solo
+       TupleN @2 = Tuple2
+       TupleN @3 = Tuple3
+       TupleN @4 = Tuple4
+       -- ...
+       TupleN @64 = Tuple64
+       TupleN @n  = TypeError (ShowType n :<>: Text " is too large; the maximum size for a tuple is 64.")
+
+     type Length :: [a] -> Nat   -- not exported
```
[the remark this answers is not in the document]
goldfirere: Indeed! Old habits die hard, even in proposals about removing old habits. Thanks for pointing this out!

proposals/0000-tuple-syntax.rst
```diff
@@ -0,0 +1,440 @@
...
+     data Sum64# ... = ...
+
+#. Change ``GHC.Types`` to have the following definition::
+
+     data List a = [] | a : List a
+
+#. Re-export ``List`` from ``GHC.List`` and ``GHC.Prelude``.
+
+#. Introduce a new extension ``-XListTuplePuns``; this extension is part
+   of ``-XHaskell98``, ``-XHaskell2010``, and ``-XGHC2021``. It is thus on by default.
```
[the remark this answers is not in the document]
goldfirere: This is stated below, but perhaps too late. I've added a comment right at the top.

proposals/0000-tuple-syntax.rst
```diff
@@ -0,0 +1,440 @@
...
+      ``tyi`` inferred to have kind ``Type`` or ``Constraint`` has kind ``Constraint``, is a synonym
+      for ``GHC.Tuple.CTuple``\ *n* ``ty1 ty2 ... tyn-1 tyn``.
+
+   #. An unapplied occurrence of ``GHC.List.List`` is pretty-printed as ``[]``.
+
+   #. An occurrence of ``GHC.List.List ty`` is pretty-printed as ``[ty]``.
+
+   #. An occurrence of ``GHC.Tuple.Unit`` is pretty-printed as ``()``.
+
+   #. An occurrence of ``GHC.Tuple.Tuplen ty1 ty2 ... tyn`` is pretty-printed as ``(ty1, ty2, ..., tyn)``.
```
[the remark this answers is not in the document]
goldfirere: It's actually correct as it was, where the `n` was a variable. I think this is clearer in the new version.

proposals/0000-tuple-syntax.rst
```diff
@@ -0,0 +1,440 @@
...
+      is expected to be of kind ``Constraint`` and either none of the ``tyi`` are inferred to have kind ``Constraint``
+      or there exists a ``tyi`` inferred to kind ``Type`` and none of the ``tyj`` (with j < i) are inferred to have
+      kind ``Constraint``, is
+      a synonym for ``GHC.Tuple.Tuple``\ *n* ``ty1 ty2 ... tyn-1 tyn``.
+
+   #. With ``-XUnboxedTuples``, an occurrence of ``(# #)`` in type-syntax is a synonym for ``GHC.Exts.Unit#``.
+
+   #. With ``-XUnboxedTuples``, an occurrence of ``(#,,...,,#)`` where there are *n* commas (for *n* ≧ 1) in type-syntax
+      is a synonym for ``GHC.Exts.Tuple``\ *n+1*\ ``#``.
+
```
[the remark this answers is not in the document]
goldfirere: Quite right. Thanks!

proposals/0000-tuple-syntax.rst
```diff
@@ -0,0 +1,451 @@
...
+     type Tuple0 = Unit
+     type Tuple1 = Solo
+     data Tuple2 a b = (a, b)
+     data Tuple3 a b c = (a, b, c)
+     -- ...
+     data Tuple64 ... = (...)
+
+#. Export the following definitions from ``GHC.Exts`` and ``GHC.Prelude``. These
+   the existing tuple definitions (in ``GHC.Classes``) today. (Note that ``(...) =>`` is special syntax, and does not
```
konsumlamm, line 98:
```suggestion
#. Export the following definitions from ``GHC.Exts`` and ``GHC.Prelude``. These replace
   the existing tuple definitions (in ``GHC.Classes``) today. (Note that ``(...) =>`` is special syntax, and does not
```

proposals/0000-tuple-syntax.rst
```diff
@@ -0,0 +1,451 @@
...
+#. Three releases after this proposal is implemented, remove the ``Solo`` pattern synonym from ``GHC.Tuple``.
+
+Effect and Interactions
+-----------------------
+1. With ``-XListTuplePuns`` (which is on by default), all programs that are accepted today continue
+   to be accepted, and with the same meanings. Note that the peculiar dance around type tuples and constraint
+   tuples exists today; I have tried to describe the current implementation faithfully, above.
+
+#. With ``-XListTuplePuns`` (which is on by default), most pretty-printing will happen as it does
+   today. The exception is around unsaturated ``CTuplen``, which is not handled above. It is hard to have
```
konsumlamm, line 343:
```suggestion
   today. The exception is around unsaturated ``CTuple<n>``, which is not handled above. It is hard to have
```

proposals/0000-tuple-syntax.rst
```diff
@@ -0,0 +1,451 @@
...
+      or there exists a ``ty<i>`` inferred to kind ``Type`` and none of the ``ty<j>`` (with *j* < i) are inferred to have
+      kind ``Constraint``, is
+      a synonym for ``GHC.Tuple.Tuple<n>`` ``ty1 ty2 ... ty<n-1> ty<n>``. (This rule retains today's behavior.)
+
+   #. With ``-XUnboxedTuples``, an occurrence of ``(# #)`` in type-syntax is a synonym for ``GHC.Exts.Unit#``.
+
+   #. With ``-XUnboxedTuples``, an occurrence of ``(# ty #)`` in type-syntax is a synonym for ``GHC.Exts.Solo# ty``.
+
+   #. With ``-XUnboxedTuples``, an occurrence of ``(#,,...,,#)`` where there are *n* commas (for *n* ≧ 1) in type-syntax
+      is a synonym for ``GHC.Exts.Tuple<n+1>#.
```
konsumlamm, line 273:
```suggestion
      is a synonym for ``GHC.Exts.Tuple<n+1>#``.
```

monoidal:
> > If NoListTupleTypeSyntax is on, should type T = '[] be rejected?

> Why would that be rejected? Are you worried about the '? Dealing with that is the subject of another proposal (#270 I believe).

Yes, it's about the '. I don't think it falls under `-Wpuns`. Consider `type T = '[Int, Char]`. There's no punning here, two elements always mean a list and never the list type. Or did I misunderstand `-Wpuns`?

We could add a warning such as `-Wticked-promoted-constructors` that would uniformly warn about all occurrences of ticks; in that case, this would be orthogonal to this proposal.

However, I see value in disallowing `'[Int, Char]` with `-XNoListTuplePuns`. Ultimately, we don't want `-XNoListTuplePuns` users to use the tick. If we're going to have backwards incompatible changes, it's better to batch them. For example, we don't have "promoted numbers" `type T = '1`, because there is no punning involved. If there's no punning with lists under `-XNoListTuplePuns`, I would also disallow `type T = '[Int, Char]`.

int-index:
#270 used to include a warning for ticked promoted constructors (and that's why @goldfirere pointed to it), but its scope was later reduced and now it's just about puns, so I agree with @monoidal.

aspiwack:
> However, I see value in disallowing `'[Int, Char]` with `-XNoListTuplePuns`. Ultimately, we don't want `-XNoListTuplePuns` users to use the tick. If we're going to have backwards incompatible changes, it's better to batch them. For example, we don't have "promoted numbers" `type T = '1`, because there is no punning involved. If there's no punning with lists under `-XNoListTuplePuns`, I would also disallow `type T = '[Int, Char]`.

On the other hand `-XNoListTuplePuns` doesn't imply no puns. So the simple quote would still appear, potentially, on other constructors. I'm not sure whether it is a good idea to make it harder to have a uniform notation.

monoidal:
If the code relies on puns and simultaneously uses `-XNoListTuplePuns`, it'd have ticks on constructors but not on lists/tuples. I don't think this is "making it harder to have a uniform notation". Compare
`type T = '[ '[ 'C, 'D], '[ 'E]]` 
`type T = [['C, 'D], ['E]]`
I prefer the second version. Ticks just introduce noise.

From one angle, having optional ticks on lists mirrors optional ticks on constructors. But from the other perspective, the ticks themselves are a source of nonuniformity: there are two ways to write the same thing `C` or `'C`, and `'C` does not correspond to value-level syntax. For normal constructors, we have to keep ticks, but for lists with `-XNoListTuplePuns` we don't have to. I'd prefer not to extend the "bad" behavior - even if that'd be consistent with data constructors.

matpyz:
As much as I can't wait for Dependent Haskell, and I understand the rationale for avoiding puns, I humbly ask to introduce some alternative mixfix notation for tuples, like the one already proposed or some other, because some people are surely wery used to these. Would it really hurt to introduce syntax like `(*` Int, Bool, Ordering *)`? Use of  `*` suggests that we are working with a product and at the same time suggests that this type is lifted (ie. in kind `Type`, aka. `*`). For unboxed tuples, you could consider `(! Int, Bool, Ordering !)` (an obvious association with strictness and bangs).

Sincerely, a Haskeller of 10 years.

int-index:
@matpyz How is `(* Int, Bool, Ordering *)` an improvement over `Tuple [Int, Bool, Ordering]`?

matpyz:
@int-index No extra parentheses are required here: `Maybe (* Int, Bool *)`, while they are required here `Maybe (Tuple [Int, Bool])`.

goldfirere:
I'm quite sympathetic to @matpyz's plea, but in the end lean against, for the following reasons:
 - We do offer a mixfix syntax, via `Tuple`. I fully agree that it's less wieldy than `(* a, b, c *)`, but it does exist in this proposal.
 - The `Tuple` syntax is easy to search for, whereas `(* ... *)` is effectively impossible to search for.
 - The `Tuple` syntax is easy to read: once a reader knows what a tuple is -- even if they don't know this particular syntax -- they can easily guess what `Tuple [a, b, c]` would mean. One of my goals in language design is to optimize for reading over writing: code is read many more times than it is written.
 - Once this proposal is accepted, both `(a, b, c)` and `Tuple [a, b, c]` will be in active use (because the proposal is opt-in). It seems (for me) easier to understand `Tuple [a, b, c]` as equal to `(a, b, c)` than to have `(* a, b, c *)` the same as `(a, b, c)` (why have two syntaxes that are so similar for the same idea?). (This is pretty subjective, I know.)
 - On the other hand `(* a, b, c *)` is definitely easier to write.
 - The `(* a, b, c *)` syntax fits better with the history of Haskell than does the `Tuple [a, b, c]` syntax.

I don't mean to just shoot you down, @matpyz, but I'm still not convinced to make this change.

goldfirere:
@int-index @monoidal I don't understand how this proposal interacts with `'` and would prefer to keep `'` fully separate. That said, if you feel it belongs here, could one of you write a diff against this proposal or a written bullet point for me to include (say, as an Alternative)?

matpyz:
Thank you @goldfirere for your response. Readability/searchability point is quite convincing, I won't bother you all anymore with `(* ... *)`. However, proposal does not include a mixfix syntax like `Tuple# (Int, Bool, Ordering)` syntax for unboxed tuples (neither `Sum#` for sums). I believe it could be provided using type families (and `Tuple` itself):
```haskell
type family RepsToTypes :: forall (reps :: List TypeRep) -> List Type = r | r -> reps where
  RepsToTypes [] = []
  RepsToTypes (rep : reps) = TYPE rep : RepsToTypes reps
type family Tuple# :: forall (reps :: List TypeRep) . forall (t :: Tuple (RepsToTypes reps)) -> TYPE (TupleRep reps) = r | r -> t reps where
  Tuple# @[] () = Unit#
  Tuple# @[rep] (MkSolo (t :: TYPE rep)) = Solo# t
  Tuple# @[rep1, rep2] ((t1 :: TYPE rep1), (t2 :: TYPE rep2)) = Tuple2# t1 t2
  ...
type Sum# :: forall (reps :: List TypeRep) . forall (t :: Tuple (RepsToTypes reps)) -> TYPE (SumRep reps) = r | r -> t reps where
  Sum# @[rep1, rep2] ((t1 :: TYPE rep1), (t2 :: TYPE rep2)) = Sum2# t1 t2
  ...
```

goldfirere:
@matpyz I had thought mixfix for unlifted sums/tuples was impossible. I was wrong. Well done, @matpyz!

It does make me wonder whether, for consistency, we should have `Tuple (a, b, c)` instead of `Tuple [a, b, c]`. And, I think with a little cunning, we can get this to work even with `-XListTuplePuns`. I will attempt soon -- should be this week.

matpyz:
I've started appreciating `type Tuple :: List Type -> Type` as a powerful tool. While uniformity is nice, precision and expressiveness is nicer. One can offer pattern synonyms to give tuples a GADT-like feel:
```haskell
import Data.Kind (Constraint, Type)
import Data.Type.Equality (type (:~~:) (HRefl))
import Data.Typeable (Typeable)
import Type.Reflection qualified as TR
  ( eqTypeRep,
    typeRep,
    withTypeable,
    pattern App,
  )
...
data TupleView :: List Type -> Type where
  TupleViewNil :: TupleView '[]
  TupleViewCons :: Typeable ts => t -> Tuple ts -> TupleView (t ': ts)

{-# COMPLETE FromTupleView :: Tuple #-}

pattern FromTupleView :: Typeable ts => () => TupleView ts -> Tuple ts
pattern FromTupleView tv <-
  (tupleView -> tv)
  where
    FromTupleView TupleViewNil = ()
    FromTupleView (TupleViewCons x xs) = tupleCons x xs

pattern TupleNil :: Typeable ts => (ts ~ '[]) => Tuple ts
pattern TupleNil = FromTupleView TupleViewNil

pattern TupleCons :: Typeable tts => (tts ~ (t : ts), Typeable ts) => t -> Tuple ts -> Tuple tts
pattern TupleCons x xs = FromTupleView (TupleViewCons x xs)

type family Typeables types :: Constraint where
  Typeables '[] = ()
  Typeables (x ': xs) = (Typeable x, Typeables xs)

type Length :: List Type -> Type
data Length ts where
  Length0 :: Length '[]
  Length1 :: (ts ~ '[t], Typeables ts) => Length ts
  Length2 :: (ts ~ '[t1, t2], Typeables ts) => Length ts
  Length3 :: (ts ~ '[t1, t2, t3], Typeables ts) => Length ts
  Length4 :: (ts ~ '[t1, t2, t3, t4], Typeables ts) => Length ts

lengthOf :: forall ts. Typeable ts => Length ts
lengthOf = case TR.typeRep @ts of
  (TR.eqTypeRep (TR.typeRep @('[] @Type)) -> Just HRefl) -> Length0
  TR.App (TR.App (TR.eqTypeRep (TR.typeRep @((:) @Type)) -> Just HRefl) hd) tl ->
    TR.withTypeable hd succLength (TR.withTypeable tl lengthOf)
  _ -> error "lengthOf: impossible happened"

succLength :: forall t ts. Typeable t => Length ts -> Length (t ': ts)
succLength Length0 = Length1
succLength Length1 = Length2
succLength Length2 = Length3
succLength Length3 = Length4
succLength Length4 = error "succLength: n-tuples for n > 4 are not supported"

tupleCons :: forall t ts. Typeable ts => t -> Tuple ts -> Tuple (t : ts)
tupleCons x0 xs
  | Length0 <- lengthOf @ts, () <- xs = MkSolo x0
  | Length1 <- lengthOf @ts, MkSolo x1 <- xs = (x0, x1)
  | Length2 <- lengthOf @ts, (x1, x2) <- xs = (x0, x1, x2)
  | Length3 <- lengthOf @ts, (x1, x2, x3) <- xs = (x0, x1, x2, x3)
  | otherwise = error "tupleCons: n-tuples for n > 4 are not supported"

tupleView :: forall ts. Typeable ts => Tuple ts -> TupleView ts
tupleView xs
  | Length0 <- lengthOf @ts,
    () <- xs =
    TupleViewNil
  | Length1 <- lengthOf @ts,
    MkSolo x <- xs =
    TupleViewCons x ()
  | Length2 <- lengthOf @ts,
    (x, x1) <- xs =
    TupleViewCons x (MkSolo x1)
  | Length3 <- lengthOf @ts,
    (x, x1, x2) <- xs =
    TupleViewCons x (x1, x2)
  | Length4 <- lengthOf @ts,
    (x, x1, x2, x3) <- xs =
    TupleViewCons x (x1, x2, x3)
  | otherwise = error "tupleUncons: n-tuples for n > 4 are not supported"
```
It works even today in GHC 8.10.7!
```
ghci> TupleCons 8 (TupleCons True (TupleCons LT TupleNil))
(8,True,LT)
```

goldfirere:
I've now added syntax inspired by @matpyz as an Alternative and incorporated the suggestion by @monoidal and @int-index.

proposals/0000-tuple-syntax.rst
```diff
@@ -0,0 +1,562 @@
...
+  is not a tuple of types (resp. constraints).
+
+  A beautiful, unexpected consequence of this design is that it aids migration. Now, with or without
+  ``-XListTuplePuns``, a user can write e.g. ``Tuple (Int, Bool, Double)``, and this will be the types
+  of e.g. ``(1, True, 3.14)``. With ``-XListTuplePuns``, the argument to ``Tuple`` will be the tuple
+  type, of kind ``Type``. That is, the type will really be understood as ``Tuple (Tuple3 Int Bool Double)``.
+  The "``1``" equation fires, reducing to ``Tuple3 Int Bool Double``. On the other hand, with
+  ``-XNoListTuplePuns``, the user's type is understood as ``Tuple @3 (Int, Bool, Double)``, and the "``3``"
+  equation fires, reducing to ``Tuple3 Int Bool Double`` -- the same answer! And so, all users
+  can write ``Tuple`` before their tuples and not get hurt.
```
konsumlamm, line 557: This part needs to be indented by one additional space, currently it is not displayed correctly.

matpyz:
@goldfiere
1. Typo: `Sum#` should return `TYPE (SumRep reps)`.
2 . "no Solo hack" introduces non-uniformity. Consider `Tuple# (Int, Bool, Double)`. As it is proposed now, depending on `-XNoListTuplePuns`, is is either `Tuple3# Int Bool Double` or `Solo# (Int, Bool, Double)`. Tangentially, it also makes `Solo` a second class tuple, as it cannot be named uniformly (`Tuple [Int]`) just like other tuples. Instead, one could remove this hack and adjust warnings: with `-XListTuplePuns` GHC could abstain from emitting a warning when unticked promoted lists and tuples are used as arguments for `Tuple`, `Tuple#` and `Sum#` to ease transition to `-XNoListTuplePuns`. Then, consistently, `Tuple# (Int, Bool, Double) ~ Tuple3# Int Bool Double`.
4. I'd stick with list argument for `Tuple` to signify homogenous/hetoregenous runtime representation distinction. As a fellow Coq user, you should recognise an apparent vicious circle of defining an inductive type with index of the very same type (not a strictly positive occurence). I know Haskell will accept any recursion, but it smells a bit anyway.

int-index:
I also dislike using tuple constructors as inputs to `Tuple`. It’s probably fine to leave out `Tuple#` for now, with the hope that eventually we could give it this kind:

```haskell
type Tuple# :: Rec TYPE rs -> TYPE (TupleRep rs)
```
Here, `Rec` is a heterogeneous list parameterised by an interpretation, it’s a very useful abstraction (originally from https://hackage.haskell.org/package/vinyl-0.14.1/docs/Data-Vinyl-Core.html).

This would allow one to write `Tuple# (Int# :& Double# :& RNil)`, and to get from here to `Tuple# [Int#, Double#]` we’d just need to overload type-level list syntax to support heterogeneous collections.

goldfirere:
@matpyz 

1. Thanks.
2. I see your point here, especially for unboxed tuples, as I have now added to the proposal directly. For boxed tuples, though, I still think it is worthwhile.
3. I'm actually not bothered by the weird recursion here -- sorry. :)

@int-index

Yes, we might imagine this. Feel free to argue that this new idea is no good and we should have something better. But I thought it worthwhile to note how we could have mixfix syntax if we wanted it. Happy for the committee to decide this point.

matpyz:
It occured to me now: why not simply introduce `-XTickedListTupleTypes`?!

Here's the idea:
When this flag is on, type of lists is `'[]`, types of tuples are `'(,)` etc.
Also, unticked lists/tuples are interpreted in the same way in types as they've always been in terms.
Therefore, `[1,2,3] :: '[Int]`, `(1,2,3) :: '(Int, Int, Int)` would be correct as typing and kinding annotations.
This would affect users of `-XDataKinds`.
To ease the transition. GHC should for some time warn that `-XDataKinds` will soon imply `-XTickedListTupleTypes` and you should enable it to be future-proof. Later it should introduce implication, finally it should warn that it's redundant.

Ericson2314:
I agree with @int-index about :-1: on the tuple trick. Another reason is if we ever did want to switch to some sort of data family as *the* definition of tuples in the future, we would be stuck with terrible terrible impredicativity.

phadej:
Can the proposal be amended to have more direct intra-references.

There is a code like

```haskell
  TupleArgKind t_or_c 1 = t_or_c       -- *not* Solo t; see belo
  
  Tuple a = a    -- see below  
```

and I cannot find a reference, why *not* Solo anywhere below.

goldfirere:
@phadej Fixed now.

tek:
The question has come up whether we want to keep printing special syntax in `Typeable` or simply use the default representation, like `Tuple45`. Any opinions?

monoidal:
I vote for not changing the output of `Typeable` for now - in other words, let's keep the special syntax at least for a release or two, until `-XNoListTuplePuns` becomes battle-tested. The additional handling shouldn't a big burden, around 40 lines of code https://gitlab.haskell.org/ghc/ghc/-/merge_requests/8820/diffs#786588e27bcbc2a8360d2d0d3b2ce1d075797ffb_883_884.

In the future, we could group this change with making `Typeable` print `Type` instead of `*`, to avoid speading breakage over releases.

NathanielB123:
From a quick scroll I didn't see this mentioned anywhere: The pattern synonym `pattern Solo a = MkSolo a` is not actually backwards compatible with all code that uses that constructor (despite the `COMPLETE` pragma) because of https://gitlab.haskell.org/ghc/ghc/-/issues/15681.
Specifically,
```hs
f :: Monad m => Solo Int -> m Int
f x = do
  Solo y <- pure x
  pure y
```
does not compile with the pattern synonym, requiring a `MonadFail` constraint.
This is definitely an edge-case but I thought it would be worth mentioning.

nomeata:
Thanks for pointing this out. It seems there is hope that this works eventually (https://gitlab.haskell.org/ghc/ghc/-/issues/15681#note_501712)

proposals/0000-tuple-syntax.rst
```diff
@@ -0,0 +1,577 @@
...
+
+     type Sum# :: forall (reps :: List RuntimeRep). TupleArgKind# reps -> TYPE (SumRep reps)
+     type family Sum# ts where
+       Sum# () = TypeError (Text "GHC does not support empty unboxed sums. Consider Data.Void.Void instead.")
+       Sum# (a :: TYPE r) = TypeError (Text "GHC does not support unary unboxed sums. Consider Data.Tuple.Solo# instead.")
+       Sum# (a, b) = Sum2# a b
+       Sum# (a, b, c) = Sum3# a b c
+       -- ...
+       Sum# (a, b, ..., bk, bl) = Sum64# a b ... bk bl
+       Sum# @reps _ = TypeError (ShowType (Length reps) :<>: Text " is too large; the maximum size of a sum is 64.")
```
clyring, line 221: Today, unboxed sums are not limited to 64 alternatives.

```
ghci> :t (# () | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | #)
(# () | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | | #)
  :: (# () | b | c | d | e | f | g | h | i | j | k | l | m | n | o |
        p | q | r | s | t | u | v | w | x | y | z | t26 | t27 | t28 | t29 |
        t30 | t31 | t32 | t33 | t34 | t35 | t36 | t37 | t38 | t39 | t40 |
        t41 | t42 | t43 | t44 | t45 | t46 | t47 | t48 | t49 | t50 | t51 |
        t52 | t53 | t54 | t55 | t56 | t57 | t58 | t59 | t60 | t61 | t62 |
        t63 | t64 | t65 | t66 | t67 | t68 | t69 | t70 | t71 | t72 | t73 |
        t74 | t75 | t76 | t77 | t78 | t79 | t80 | t81 | t82 | t83 | t84 |
        t85 | t86 | t87 | t88 | t89 | t90 | t91 | t92 | t93 | t94 | t95 |
        t96 | t97 | t98 | t99 | t100 | t101 | t102 | t103 | t104 | t105 |
        t106 | t107 | t108 | t109 | t110 | t111 | t112 | t113 | t114 | t115
        | t116 | t117 | t118 | t119 | t120 | t121 | t122 | t123 | t124 |
        t125 | t126 | t127 | t128 #)
```

That's not a major problem for this `Sum#` type family because it can't be given as input a tuple wider than 64 entries anyway. But it poses other problems:

* The `Sum<n>#` family of type constructors exported from `Data.Sum.Experimental` is not complete.
* Pretty-printing of an unboxed sum type with more than 64 alternatives cannot use the `Sum#` type family with `-XNoListTuplePuns`.
* Because the `(# | | | #)` syntax is disallowed, it is not obvious how to write a "large" unboxed sum type or term when `-XNoListTuplePuns` is active.

PR #476, opened 2021-12-22
Removing -XNoPolyKinds

goldfirere:
Support for `-XNoPolyKinds` has become harder and harder to keep up. It's not clear that anyone is relying on the lack of kind polymorphism. I thus propose to drop support for `-XNoPolyKinds`. As the proposal describes, there is a remote possibility of breakage, which we would investigate before merging. I expect most users not to notice this change.

[Rendered](https://github.com/goldfirere/ghc-proposals/blob/poly-kinds/proposals/0000-poly-kinds.rst)

### DIFF
--- a/proposals/0000-poly-kinds.rst
+++ b/proposals/0000-poly-kinds.rst
@@ -1,0 +1,202 @@
+Drop support for ``-XNoPolyKinds``
+==================================
+
+.. author:: Richard Eisenberg
+.. date-accepted::
+.. ticket-url::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. contents::
+
+Since version 7.4, GHC has supported *kind polymorphism*, where a definition like ``data App f a = MkApp (f a)`` works
+with a variety of kinds. Kind polymorphism is a natural extension of the type
+polymorphism Haskell has had since its inception. In the light of experience
+with kind polymorphism, a mode where GHC *rejects* kind
+polymorphism has become increasingly hard to specify. This proposal thus suggests
+to have kind polymorphism be permanently on, much like type polymorphism is
+permanently on.
+
+Motivation
+----------
+Consider the following program::
+
+  data App f a = MkApp (f a)
+  data AtInt f = MkAtInt (f Int)
+
+  unApp :: App AtInt Maybe -> Maybe Int
+  unApp (MkApp (MkAtInt m)) = m
+
+Should that be accepted?
+
+The key question is: what is ``App``\ 's kind? After analyzing its definition,
+we see that ``f :: kappa -> Type`` and ``a :: kappa`` for some kind ``kappa``.
+In Haskell98, GHC insists that every datatype have one fixed, monomorphic kind.
+Thus GHC must choose how to default ``kappa``, and it chooses ``kappa := Type``,
+assigning ``App :: (Type -> Type) -> Type -> Type``. This means that the use
+``App AtInt Maybe`` is rejected.
+
+On the other hand, with ``-XPolyKinds``, GHC is allowed to *generalize* the kind
+of ``App``, assigning ``App :: forall k. (k -> Type) -> k -> Type``, and the program
+above is accepted. This is more general, composition behavior.
+
+Note that, with ``-XNoPolyKinds``, we could force acceptance of the above program,
+by defining e.g. ``data App f a = MkApp (f a) | Don'tUseThisConstructor (f Maybe)``.
+The extra, never-used ``Don'tUseThisConstructor`` tells GHC to choose ``kappa := Type -> Type``,
+but the kind is still monomorphic.
+
+Defaulting ``kappa`` is not so hard. But there are thornier cases. ::
+
+  {-# LANGUAGE PolyKinds #-}
+  module A where
+
+  data App f a = MkApp (f a)   -- infer App :: forall k. (k -> Type) -> k -> Type
+
+  ---------------------------------------
+
+  {-# LANGUAGE NoPolyKinds #-}
+  module B where
+  import A
+
+  type MyApp = App
+
+What kind should we assign to ``MyApp``? It can't be ``forall k. (k -> Type) -> k -> Type``,
+because that's kind-polymorphic, and ``-XNoPolyKinds`` has forbidden us from doing this.
+So GHC must again default, choosing ``MyApp :: (Type -> Type) -> Type -> Type``.
+
+A bit annoying, but still doable. It gets worse. ::
+
+  {-# LANGUAGE PolyKinds, ... #-}
+  module C where
+
+  data Nat = Zero | Succ Nat
+
+  type Vec :: Nat -> Type -> Type
+  data Vec n a where
+    Nil  :: Vec Zero a
+    (:>) :: a -> Vec n a -> Vec (Succ n) a
+  infixr 5 :>
+
+  type HList :: forall (n :: Nat). Vec n Type -> Type
+  data HList ts where
+    HNil  :: HList Nil
+    (:>>) :: t -> HList ts -> HList (t :> ts)
+
+  ---------------------------------
+
+  {-# LANGUAGE NoPolyKinds #-}
+  module D where
+  import C
+
+  type MyHList = HList
+
+Now what? ``HListN`` has a kind that is polymorphic in a variable
+whose kind is not even ``Type``. So we cannot default the variable to ``Type``, because
+that is ill-kinded. (Recall that ``Type :: Type``.) If we try this today, GHC issues
+an error, telling you not to engage in such shenanigans::
+
+  • Cannot default kind variable ‘n0’
+    of kind: Nat
+    Perhaps enable PolyKinds or add a kind signature
+  • In the type synonym declaration for ‘MyHList’
+
+As we continue to explore exactly when such an error gets triggered, it becomes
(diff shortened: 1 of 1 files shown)


monoidal:
I've tried implementing this. There are problems with backwards compatiblity.

1. hs-boot files:

```
-- A.hs-boot
class Integral a

-- A.hs
class Integral a where
  f :: a -> a
```

The kind of `Integral` is now different in the hs-boot file and in the hs file, error.

2. Phantom parameters become polykinded:

```
data Ptr a = Ptr Addr#

castPtr :: Ptr a -> Ptr b
castPtr = coerce
```

`Ptr` becomes polykinded and `coerce` can't change a kind (https://gitlab.haskell.org/ghc/ghc/-/issues/15441). Likewise:

```
data ForeignPtr a = ForeignPtr Addr# ForeignPtrContents

instance Eq (ForeignPtr a) where
    p == q  =  unsafeForeignPtrToPtr p == unsafeForeignPtrToPtr q
-- now fails: a is kind-poly, unsafeForeignPtrToPtr is kind-mono
```

```
-- Cabal:
class Typeable a => Structured a

newtype SymbolicPath from to = SymbolicPath FilePath

instance (Typeable from, Typeable to) => Structured (SymbolicPath from to)
-- now fails with obscure "no instance for Typeable k"
```

```
doSeq :: RemoteRef a -> IO (EvalStatus ())
doSeq ref = do
    sandboxIO evalOptsSeq $ do
      _ <- (void $ evaluate =<< localRef ref)
      return ()
-- now fails: 'a' is kind-poly
```

With those fixes, GHC builds. However, that experience makes me doubt we can remove `-XNoPolyKinds` without a gradual transition. Some ideas: issue a defaulting warning by default; enable `-XPolyKinds` in Haskell98/2010 but still support `-XNoPolyKinds` for some time; restrict `-XNoPolyKinds` to data declarations.

My branch is at `wip/no-nopolykinds`, except it needs an update to Cabal (kind annotation on the `SymbolicPath` type).

<details>

```
diff --git a/Cabal/src/Distribution/Utils/Path.hs b/Cabal/src/Distribution/Utils/Path.hs
index eb3d11deb..b4dc4ed65 100644
--- a/Cabal/src/Distribution/Utils/Path.hs
+++ b/Cabal/src/Distribution/Utils/Path.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE KindSignatures #-}
 module Distribution.Utils.Path (
     -- * Symbolic path
     SymbolicPath,
@@ -24,6 +25,7 @@ import Distribution.Utils.Generic (isAbsoluteOnAnyPlatform)
 import qualified Distribution.Compat.CharParsing as P
 -- import qualified Text.PrettyPrint                as Disp
 
+import Data.Kind (Type)
 -------------------------------------------------------------------------------
 -- * SymbolicPath
 -------------------------------------------------------------------------------
@@ -34,7 +36,7 @@ import qualified Distribution.Compat.CharParsing as P
 -- They are *symbolic* which means we cannot perform any 'IO'
 -- until we interpret them.
 --
-newtype SymbolicPath from to = SymbolicPath FilePath
+newtype SymbolicPath (from :: Type) (to :: Type) = SymbolicPath FilePath
   deriving (Generic, Show, Read, Eq, Ord, Typeable, Data)
 
 instance Binary (SymbolicPath from to)
```
</details>

phadej:
@monoidal Thank you for collecting these.

The proposal says

> This proposal thus suggests to have kind polymorphism be permanently on, much like type polymorphism is permanently on.

But specifying kind signatures is not a standard practice like specifying type signatures (yet?)!

E.g. in discussion around [`-Wmissing-kind-signatures`](https://downloads.haskell.org/ghc/9.2.1/docs/html/users_guide/using-warnings.html#ghc-flag--Wmissing-kind-signatures), in [the original issue Richard says](https://gitlab.haskell.org/ghc/ghc/-/issues/18992#note_393677)

> I would favor a slightly less chatty version that skips the warning whenever all type arguments and the result all have kind `Type`. This would be quiet on the vast majority of datatypes, but still bleat on monad transformers, stuff with `DataKinds`, etc.

I think that kind of trimmed down `-Wmissing-mono-kind-signatures` **in** `-Wall` is a prerequisite to this proposal.

That warning would warn about `Ptr`, `ForeignPtr`, `Structured`.

Also full `-Wmissing-kind-signatures` seems to be required for hs-boot (and maybe signature?) files.

phadej:
I made a quick experiment https://gitlab.haskell.org/ghc/ghc/-/merge_requests/7256 and realised that `-Wmissing-poly-kind-signatures` won't help people prepare. We'd need a warning which would say when the type would have different kind **if** `PolyKinds` were enabled. That would warn about `Ptr`, `ForeignPtr` and `Structured`; and also `Integral` in `Integral.hs-boot`.

goldfirere:
Thanks very much for this research @monoidal!

It is dispiriting, but better to be informed.

The hs-boot file issue is, I think, a red herring: we could say that any abstract type definitions in hs-boot files default all type parameters to have kind `Type`. (This is effectively what happens today.) This rule is much like how role inference defaults to `representational` in hs-boot files (instead of `phantom` the rest of the time). So maybe further challenges await here, but I think they are surmountable.

The error around `Structured` is strange -- looks like a bug. I will look into this this week if no one beats me to it.

But I agree with the high-level piece: I think we will need a gradual transition. I will either revise or retract this proposal during the week.

goldfirere:
I now understand the obscure error around `Typeable k`: it's a known limitation in the way that GHC solves constraints arising from superclasses in an instance declaration. In any case, if there's a bug here, it's either around the quality of the error message or, perhaps, an opportunity to Think Harder about a Hard Problem, but it's *not* a red herring here. (The curious may want to read [this Note](https://gitlab.haskell.org/ghc/ghc/-/blob/3219610e3ba6cb6a5cd1f4e32e2b4befea5bd384/compiler/GHC/Tc/TyCl/Instance.hs#L1485), paying particular attention to point (sc2), which is the issue here. As it turns out, this is up for revision shortly; see [GHC#20666](https://gitlab.haskell.org/ghc/ghc/-/issues/20666).)

goldfirere:
I was also a bit mystified around the `ForeignPtr` error. The reason it happens is because `Ptr` was monomorphised to get this far, so indeed `unsafeForeignPtrToPtr` is also kind-monomorphic.

Ericson2314:
> Think Harder about a Hard Problem

As I mentioned in https://github.com/haskellfoundation/tech-proposals/pull/21, I think in general GHC's implementation has too facile a notion of extension, such that the *true* language it implements is the "all extensions on" monster, and restrictions on that all tend to be a bit ad-hoc.

If we wish to have simplified versions of Haskell be better supported, with better errors and user interactions more broadly, I think at some point we are going to have to confront that problem head-on. Maybe `-XNoPolyKinds` just is a somewhat incoherent thought, but surely are other cases (`-XNoDependentTypes`, in some form?) that we do care about.

I hope researches take up the challenge, because the ultimate goal of both

 -  a modular notion of a what a language is
 -  great libraries / abstractions to get proper implementations of many different languages

would be extremely value. In some sense, it is *the* end goal of PL: we have learned how to make good enough languages to create wonderful ecosystems espousing the principles of good abstractions and code reuse, now we need to bring those same practices and principles to the design and production-quality implementation of the languages themselves.

goldfirere:
@Ericson2314 I agree! I think defining "language levels" within GHC would be a nice step forward. I'm not so worried about implementation (perhaps I should be), but instead worried about design: what features would go in the different levels? How would more advanced levels be discovered? How customizable would the interface be (could you be at level 2 but also include just `-XImpredicativeTypes`, for example)?

I don't quite want to tackle that in this ticket, but I think making concrete progress here would be wonderful.

Ericson2314:
Yeah I agree I don't want to scope-creep this one, but just rather point out the larger issue to keep in mind as why decide what to do in the shorter term.

adamgundry:
While I think `PolyKinds` is mostly a harmless extension, this would introduce yet more divergence between GHC (even in `Haskell98`/`Haskell2010` mode) and the underlying standards, and I'd expect the proposal to discuss that. This proposal seems essentially to be saying that `PolyKinds` is fixing a design bug in the Haskell standards, and hence should be made mandatory rather than being a "normal" optional extension. That's a plausible position to take, but perhaps it could be more strongly motivated (e.g. arguing that `PolyKinds` would be more consistent and a better default even for a `Haskell2010` programmer).

Ericson2314:
@adamgundry I suspect part of the answer that `-XNoPolyKinds` has sketchy rules for importing poly-kinded stuff, whereas baseline standard Haskell doesn't need to deal with that. Perhaps the answer is that GHC should be more strict in ensuring imported interfaces do not exceed the current language. That would mean keeping `-XNoPolyKinds`, but making it mean something more limiting than what it currently does.

phadej:
Thinking about this, I'm confused why `unsafeForeignPtrToPtr` is not polykinded if both `Ptr` and `ForeignPtr`  modules are `PolyKinds` enabled? Both types have phantom type parameter?

monoidal:
There is no problem with polykinded `unsafeForeignPtrToPtr` if `Ptr` and `ForeignPtr` are polykinded. However, I made `Ptr` monokinded because of the issue with `castPtr`. I could have added a kind signature on `castPtr` instead, then it would work.

Nevertheless, it makes no sense to have `Ptr Maybe`, the more general kind is not helpful here.

phadej:
@monoidal that's another way to say that "Sometimes the most general kind is not what users want for their types". And as users don't usually write kinds for their types, we circle back to having `PolyKinds` on by default is bad idea.

Though it is in GHC2021. (Anecdotally users run into that: `optics` doctests need to have `-XHaskell2010` to run).

PR #477, opened 2021-12-23
Unicode ellipsis.

kindaro:
## Motivation

We already have beautiful Unicode make-up for all reserved symbols. And they look great!

But no, not for all. The poor ellipsis `..` is left without. It wants to look like this: `…`.

## Proposed Change Specification

Make `…` accepted wherever `..` is accepted.

[Rendered](https://github.com/kindaro/ghc-proposals/blob/unicode-ellipsis/proposals/477-unicode-ellipsis.md)

### DIFF
--- a/proposals/477-unicode-ellipsis.md
+++ b/proposals/477-unicode-ellipsis.md
@@ -19,7 +19,9 @@ But no, not for all. The poor ellipsis `..` is left without. It wants to loo
 
 ## Proposed Change Specification
 
-Make `…` accepted wherever `..` is accepted.
+1. Make `…` accepted wherever `..` is accepted.
+
+2. Allow `…` in situations like `[False…]` where ASCII syntax `[False..]` would be a parse error. _(This is automatically the case with the patch proposed below.)_ _(Thanks to Jakob Brünker for [pointing this out](https://github.com/ghc-proposals/ghc-proposals/pull/477#issuecomment-1000255004).)_
 
 ## Examples
 
--- a/proposals/477-unicode-ellipsis.md
+++ b/proposals/477-unicode-ellipsis.md
@@ -1,0 +1,83 @@
+---
+author: Ignat Insarov
+date-accepted: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/0>).
+**After creating the pull request, edit this file again, update the number in
+the link, and delete this bold sentence.**
+
+# Proposal title
+
+Unicode ellipsis.
+
+## Motivation
+
+We already have beautiful Unicode make-up for all reserved symbols. And they look great!
+
+But no, not for all. The poor ellipsis `..` is left without. It wants to look like this: `…`.
+
+## Proposed Change Specification
+
+Make `…` accepted wherever `..` is accepted.
+
+## Examples
+
+```haskell
+{-# language UnicodeSyntax #-}
+{-# language RecordWildCards #-}
+
+module X where
+
+import Prelude (Bool (…))
+
+data Record = Record {value ∷ Bool}
+
+function ∷ Record → Bool
+function Record {…} = value
+
+booleans ∷ [Bool]
+booleans = [False…]
+```
+
+## Effect and Interactions
+
+Ellipsis gets its beautiful Unicode make-up.
+
+## Costs and Drawbacks
+
+There are no costs and no drawbacks.
+
+## Alternatives
+
+There are no alternatives.
+
+## Unresolved Questions
+
+There are no unresolved questions.
+
+## Implementation Plan
+
+Apply this patch:
+
+```diff
+diff --git a/compiler/GHC/Parser/Lexer.x b/compiler/GHC/Parser/Lexer.x
+index d74d17be8f..4135372d31 100644
+--- a/compiler/GHC/Parser/Lexer.x
++++ b/compiler/GHC/Parser/Lexer.x
+@@ -1063,6 +1063,7 @@ reservedSymsFM = listToUFM $
+        ,("-<<", ITLarrowtail NormalSyntax,  NormalSyntax,  xbit ArrowsBit)
+        ,(">>-", ITRarrowtail NormalSyntax,  NormalSyntax,  xbit ArrowsBit)
+ 
++       , ("…",  ITdotdot,                   UnicodeSyntax,  0 )
+        ,("∷",   ITdcolon UnicodeSyntax,     UnicodeSyntax, 0 )
+        ,("⇒",   ITdarrow UnicodeSyntax,     UnicodeSyntax, 0 )
+        ,("∀",   ITforall UnicodeSyntax,     UnicodeSyntax, 0 )
+```
+
+## Endorsements
+
+There are no endorsements.
+
--- a/proposals/477-unicode-ellipsis.md
+++ b/proposals/477-unicode-ellipsis.md
@@ -5,9 +5,7 @@ ticket-url: ""
 implemented: ""
 ---
 
-This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/0>).
-**After creating the pull request, edit this file again, update the number in
-the link, and delete this bold sentence.**
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/477).
 
 # Proposal title
 
--- a/proposals/477-unicode-ellipsis.md
+++ b/proposals/477-unicode-ellipsis.md
@@ -19,6 +19,8 @@ But no, not for all. The poor ellipsis `..` is left without. It wants to loo
 
 ## Proposed Change Specification
 
+When the `UnicodeSyntax` extension is enabled:
+
 1. Make `…` accepted wherever `..` is accepted.
 
 2. Allow `…` in situations like `[False…]` where ASCII syntax `[False..]` would be a parse error. _(This is automatically the case with the patch proposed below.)_ _(Thanks to Jakob Brünker for [pointing this out](https://github.com/ghc-proposals/ghc-proposals/pull/477#issuecomment-1000255004).)_


JakobBruenker:
The examples are not in line with the change specification "Make … accepted wherever .. is accepted.".

Example from the proposal:
```haskell
booleans = [False…]
```

Using `..`:
```haskell
ghci> booleans = [False..]

<interactive>:1:20: error: parse error on input `]'
```
This doesn't work because `False.` makes `False` look like the module name of a qualified name.

I see not particularly strong reason not to `[False…]` though, so I suppose my preferred way to reconcile this would be to rephrase the proposed changes.

kindaro:
Summary of major points raised during the discussion period:

1. The examples were not in line with the change specification. The proposed changes were rephrased so that the examples are now in line with the change specification.

I believe that the proposal is ready for review. I tag the committee secretary (currently @nomeata).

proposals/477-unicode-ellipsis.md
```diff
@@ -0,0 +1,83 @@
...
++++ b/compiler/GHC/Parser/Lexer.x
+@@ -1063,6 +1063,7 @@ reservedSymsFM = listToUFM $
+        ,("-<<", ITLarrowtail NormalSyntax,  NormalSyntax,  xbit ArrowsBit)
+        ,(">>-", ITRarrowtail NormalSyntax,  NormalSyntax,  xbit ArrowsBit)
+ 
++       , ("…",  ITdotdot,                   UnicodeSyntax,  0 )
+        ,("∷",   ITdcolon UnicodeSyntax,     UnicodeSyntax, 0 )
+        ,("⇒",   ITdarrow UnicodeSyntax,     UnicodeSyntax, 0 )
+        ,("∀",   ITforall UnicodeSyntax,     UnicodeSyntax, 0 )
+```
```
nomeata, line 78: Probably there is more to do: Mention this in `docs/users_guide/exts/unicode_syntax.rst`, adjust the pretty-printer in `compiler/GHC/Utils/Outputable.hs`, maybe a test case.

I assume that is implied here.

nomeata:
/remind me to check progress in two weeks.

AntC2:
Pasting in @int-index's [comment from the Steering Committee](https://mail.haskell.org/pipermail/ghc-steering-committee/2022-February/002725.html).

> Personally I use three dots to indicate omissions in the source text, like this:
>
>        f x = …
>
> So that is the meaning I’d like to reserve for three dots, not ranges.
>
> The proposal could use ‥ U+2025 instead.

As a further data point, GHC/i on my machine (Windows with `-XUnicodeSyntax`) gets indigestion with these characters. It appears to compile source code OK, but in error messages they display as `?`.

Neither can I get int-index's use for omissions to compile (8.10.7):

    parse error on input `?'
      |
    29 |     fooEllipsis x = ?                -- the source has ellipsis …
      |                     ^

Surrounding the ellipsis in `(…)` works ok. But equally I could go

    fooDotty x = (...)            -- three dots, as an operator
    footba x = (tba)

proposals/477-unicode-ellipsis.md
nomeata: These are the two I can think of, I expect that code review when actually building the feature might turn up more.

ulysses4ever:
Dear all, I'm reading the thread on ghc-steering-committee mailing list and totally agree that having different amounts of dots in Unicode v ASCII is confusing.

As a possible solution for this confusion, could we add support for **three ASCII dots**? Maybe activated with -XUnicodeSyntax (probably confusing), another extension (with a prospect to integrate it in -XGHCXXXX at some point) or unconditionally (preferably). I doubt this will break anything, and will be a nicer UX in general: I switch between couple languages a lot and always confuse where to use 2 v 3 dots, and I wish any option could work.

nomeata:
Hmm, I am not sure I agree that it's confusing. At least to me, from the start when I saw Haskell, the ascii `..` meant what's `\ldots` in LaTeX, and what I tend to write as three close-to-each other dots on paper and is semantically an ellipsis, which in Unicode has a codepoint with that names that in most fonts looks like what you expect it to…

So the `..` is an ASCII cludge like all others (probably because `...` is unnecessarily wide).

The analogue is `\` for λ - just because the left leg is missing we wouldn't use a unrelated look-alike like \ for the unicode variant, nor ask to be able to write `,\` in ASCII for a better approximation of the real letter.

TL;DR: no need to change ASCII-Haskell, and let's the semantically correct ellipsis symbol in Unicode-Haskell.

kindaro:
> As a possible solution for this confusion, could we add support for **three ASCII dots**?

The problem with three dots `...` is that it is a legitimate identifier for an operator. It axually appears [in many places on Hackage][1]. I expect that the conservative party of the Haskell state will push back hard against such a proposal.

* * *

I have zero interest in adding the Unicode two dots syntax to Haskell.

<details>
<summary>This is why.</summary>
My aim is not to add more Unicode — my aim is to add more beauty. I want my Haskell code to be more accessible for everyone.

* This proposal, were it accepted, would achieve my aim because the two dots in Haskell currently mean exactly the same as the ellipsis elsewhere in literature _(«an intentional omission of a word, sentence, or whole section from a text without altering its original meaning»)_ — so, if we replace the two dots with the ellipsis in a Haskell program, plausibly this Haskell program will be easier to understand.
* The Unicode two dots, on the other hand, are an essentially meaningless typographic construct. I should not propose adding meaningless typographic constructs to the syntax of Haskell.
</details>

If this proposal must be rejected as it stands because some people in the GHC steering committee prefer the Unicode two dots to the ellipsis for the sake of consistency with the current two dots syntax, then let this proposal be rejected. I understand the values behind this preference, I accept that it is valid but unfortunately it does not align with my interests.

[1]: https://hoogle.haskell.org/?hoogle=...

tomjaguarpaw:
> My aim is not to add more Unicode — my aim is to add more beauty. I want my Haskell code to be more accessible for everyone. 

I don't feel especially strongly about this proposal but I have to say I don't think this explanation holds much weight. For those who want their code to look beautiful they can obtain [a particular font with ligatures](https://github.com/i-tu/Hasklig/#hasklig) designed exactly to present the code (to them) in a "beautiful" way. Doing so has the great benefit that it does not impinge on everyone else!

I'm skeptical that adding support for Unicode ellipsis makes Haskell code more accessible for everyone and if that's being used as justification then I think I would like to see some evidence.

And personally I find `[1..]` *much* clearer to read and much easier to type (i.e. *much* more accessible accessible) than `[1…]` (but perhaps the former is just my font ...)

goldfirere:
I find the history interesting, but I urge us not to take it into account when making a decision about the future of our language. That is, just because a famous computer scientist (or three) used a certain notation in a certain publication years ago does not, by itself, say that the same choice is the right one today. Their use was in LaTeX for a book. Our use is in Unicode for an active programming language. It's possible that the same factors that led to their choice are still in play in the new context, but I see no evidence for that here.

Put another way: 99.9% of the users of this construct will be entirely unaware of this history. Given that fact, what design works best for them?

I don't actually have a good answer to my question (three-dot ellipsis is more standard and easier to type; two-dot ellipsis fits better with ASCII syntax), but the history should not be a factor, in my opinion.

nomeata:
It's not just history, it's also trying to align with current mathematical notation. For example https://en.m.wikipedia.org/wiki/Interval_(mathematics)#Integer_intervals. It's a bit like saying Monoid rather than Appendable. At least until there is evidence that using three dots is at least as common a notation for intervals “out there” than the two dot notation.

goldfirere:
I like your connection with Monoid vs Appendable. Perhaps it's telling that I always thought it should be Appendable... I don't think Haskell does favors for itself by using terminology unfamiliar to many programmers. (To be clear, I am *not* suggesting we should change this about `Monoid`!)

In any case, citing mathematical usage is more relevant for me than the history. But I still wonder how that fact affects the bulk of our users. After degrees in physics and computer science, and after teaching high school math (alongside computer science) for a number of years, I have never seen `..` used for interval notation outside of Haskell code. I *have* seen both `...` and `,` (as in the interval `[0, 2π)`), but never `..`.

tomjaguarpaw:
On the one hand I feel like I'm at risk of derailing the discussion, but given that it has become wider ranging already I'll accept that risk.

Perhaps we should reconsider whether we really like the `..` notation *at all*. Does it really pull its weight compared to the `enumFrom` and `enumFromTo`[1] functions? We must keep `..` even if we don't like it but we might conclude that we shouldn't embellish it with new special syntax, and perhaps even that we should discourage its use in favour of the named functions. If there were a compiler/linter flag that forbade `..` then I would be strongly inclined to turn it on in my own projects.

I'm not a Lisp aficionado, but it some circumstances the best syntax is no syntax. I really think that Haskell has too much syntax already.

(I offer my apologies to @kindaro for responding to his proposal intended to make something more beautiful by saying "perhaps we should be discouraging that thing anyway".) 

[1] and `enumFromThen`/`To` of course, but I cannot remember even seeing `[x,y,..z]`/`[x,y..]` used in practice

nomeata:
Gathering some evidence: The Wikipedia page mentions Pascal. Isabelle also uses two dots, as cab be seen at https://stackoverflow.com/q/54255213/946226, and it would be idiomatic Isabelle to define “pretty” LaTeX output with \\dots, but it [does not do so](https://isabelle.in.tum.de/library/HOL/HOL/Set_Interval.html).

ulysses4ever:
It would be cool to to gather a reference of interval notations in programming languages. It seems that numerical computing languages prefer an explicit function (usually called `range`), except in (generally, my favorite) Julia, where you use a colon instead (`1:10`).

kindaro:
Yo, reckon the two dots syntax has two meanings in Haskell:

* What is called _«interval notation»_ above.

  In Haskell, this notation is hardly about intervals. Real numbers are manifestly not enumerable.

* Omission of names in import and export lists, as well as in record constructors.

  To this case, the interval notation from _Concrete Mathematics_ is entirely irrelevant.

In both cases, the definition I have given above — applies. To recall: _«an intentional omission of a word, sentence, or whole section from a text without altering its original meaning»_. In contrast, the conversation above only accounts for the first case.

Further, it is not clear to me that the quote:

> Since then, the Hoare-Ramshaw interval notation has been incorporated into various programming languages, including Pascal and Haskell.

— Is authoritative. Whence does the author know that the intention of the designers of Haskell was to copy the Hoare-Ramshaw notation for intervals, and not to simply save the typing of one dot? How does this theory align with the two dot syntax also being used in import and export lists?

* * *

> I'm skeptical that adding support for Unicode ellipsis makes Haskell code more accessible for everyone and if that's being used as justification then I think I would like to see some evidence.

Brother Tom, you know as well as I that there is no such evidence. It is not feasible to get statistics for every small change of grammar. [We have to make a call to common sense or specialization of principles instead.][1] When a counterfactual change is in question, evidence is even harder to apply. Even if we could A/B test the next release of GHC, it will hardly inform us — of course the current Haskell programmers are more familiar with the current two dot syntax. If Simon followed this logic, his name would have been Bjarne.

The principle I am specializing here is that one entity should have one name. First, the entity in question is _«an intentional omission of a word, sentence, or whole section from a text without altering its original meaning»_ and the name for it is _«ellipsis»_; then, the entity is the ellipsis and the name for it `…`. This is all standard usage across all writing and typography. Why invent another name where the one in standard usage fits perfectly?

So, my claims:

1. The ellipsis is familiar to far more people than the two dots.
2. Having more than one name for the same entity leads to a confusion of discourse.

I say these claims are common sense.

[1]: https://astralcodexten.substack.com/p/the-phrase-no-evidence-is-a-red-flag

> And personally I find `[1..]` _much_ clearer to read and much easier to type (i.e. _much_ more accessible accessible) than `[1…]` (but perhaps the former is just my font ...)

How do you find that you find it so? And with such heavy emphasis? I want to hear from you, since my experience is the opposite. Possibly you can send me a screen shot so that I can appreciate your font?

Note though that no one is taking your way of writing it away. I only propose to enrich the experience of Haskell, not to impoverish it.

Other questions:
* Do you entertain Unicode syntax frequently overall?
* What is your input method?

kindaro:
@konsumlamm   I see a confused emoji you attached to my message above. What is confusing? Maybe I can repair my message so that your confusion changes to understanding?

tomjaguarpaw:
> you know as well as I that there is no such evidence

I don't think this is a fair criticism. Indeed I *believe* there is no evidence, because I *believe* the claim is false! I don't think that means, however, that counterfactually there could be no evidence if it were true. Evidence could be, for example, that on 10% of the keyboards of users worldwide there is no `.` key but there is a `…` key. That would be fairly convincing evidence that adding `…` would make Haskell more accessible! (Although perhaps I am misunderstanding what you meant by "accessible". If so then I would be grateful if you would clarify.)

I'm not sure I see the relevance of the Astral Codex Ten article, sorry.

> Possibly you can send me a screen shot so that I can appreciate your font?

Sure, I find the three `…` dots rather small and hard to read. I find the two `..` large and clear.

screenshot

I find `…` hard to type because there is no such key on my keyboard. I would have no idea how to enter that symbol.

> Note though that no one is taking your way of writing it away. I only propose to enrich the experience of Haskell, not to impoverish it.

Well, I fear that my experience with Haskell will indeed be impoverished by having to learn that `…` is a valid alternative to `..`, and to interpret it that way whenever I see it.

> Do you entertain Unicode syntax frequently overall?

No, I don't. I would prefer that Haskell stuck to ASCII. I could be persuaded to relent in certain situations if we really ran out of symbolic syntax (I understand that Agda and Lean have done so, and it makes sense for their mathematical use case).

> What is your input method?

[Standard ISO keyboard](https://en.wikipedia.org/wiki/ISO/IEC_9995).

tomjaguarpaw:
(I should reiterate: "I don't feel especially strongly about this proposal". That's why I have stayed out of the debate until recently. But I would prefer that if we claim that we're making claims about accessibility that they are backed up.)

int-index:
> Whence does the author know that the intention of the designers of Haskell was to copy the Hoare-Ramshaw notation for intervals, and not to simply save the typing of one dot? 

It’s a guess, to be honest. And I may have gotten the causality wrong, actually, especially now that I’m looking at the dates. But `..` having precedence in mathematical literature is a strong argument for me to stick with it.

int-index:
@kindaro I agree with your point that intervals are different from records and record wildcards. However, they happen to share syntax. I don’t object strongly to `import M (T(…))` or `fn R{…} = e`, but `[1…]` does look off to me.

Furthermore, I initially wrote `fn R{…} = ...` instead of `fn R{…} = e`, but then realized that would’ve been confusing. Might be my personal preference, but I often use `...` as meta-syntax to indicate an omission, and that relies on `...` not being actual syntax.

kindaro:
@tomjaguarpaw   Ah I see, so your argument against `…` is general. By the same line of reasoning, and as you say straight, you are against all Unicode syntax. Well, this train has left the station a long time ago. We have Unicode syntax. I only want to repair an omission. Whichever arguments were persuasive enough for the language extension `UnicodeSyntax` to be added, apply as well to the Unicode ellipsis. You would have to dig up those arguments and contest them.

Let me explain the reference to the post _«The Phrase "No Evidence" Is A Red Flag For Bad Science Communication»_. I said that `…` will increase accessibility. You said I should bring evidence. The post says that evidence is neither sufficient nor necessary, and that other epistemic instruments, such as common sense, should be admitted. Therefore, I bring no evidence, and instead appeal to common sense. Furthermore I point out that innovation cannot be evidence based in the first place, thus your objection is invalidated.

My story: a while ago I switched to variable pitch fonts and enhanced my standard keyboard with frequent Unicode characters. Input problem solved, output problem solved. It is not hard to do, message me if you are curious to try.

I am going now to summarize the trade that this proposal entails:

* People that hate Unicode get to be forced to remember that `…` stands for `..`.

  They did not complain when they were forced to remember that `..` stands for `…` though. I call this hypocrisy. If you want to remember less, then forget rather about `..` and support this proposal.

* People that love Unicode get to have a compete set of Unicode syntax.

* Everyone gets to have one name for one entity, as I describe above.

  You never engaged with that argument. Do you reject my premises? Do you reject general arguments overall?

> (I should reiterate: "I don't feel especially strongly about this proposal". That's why I have stayed out of the debate until recently. But I would prefer that if we claim that we're making claims about accessibility that they are backed up.)

I opened this whole proposal as a joke. No one would seriously talk about a trivial one line patch, I thought. A bureaucratic formality, I thought. And look where we are now, writing screen long epistemic critique to each other.

* * *

@int-index   Why is `..` having precedent in mathematical literature a strong argument for you to stick with it?

* I do not see said precedent in actual literature _(not that I read a lot, but I read some)_, nor does Richard @goldfirere _(as he says above)_. Do you see said precedent in actual literature? The notation I am used to looks like `1, …, 10` and I see `1… 10` as a shorthand for that and then `1.. 10` as a shorthand for the shorthand. I do not call for `1, …, 10` to be allowed, because it is not a low-hanging fruit. But at least we can eliminate one level of shorthanding with a trivial patch.

* Then, further, in Haskell we have `1..` and `1, 3.. 10` — is there a precedent in mathematical literature for that?

> Might be my personal preference, but I often use `...` as meta-syntax to indicate an omission, and that relies on `...` not being actual syntax.

Me too! I saw your electronic letter quoted above and I remember this objection. But I do not see how making `…` actual syntax impedes this. As I understand, it will fail to compile either way. So I do not see how the present proposal changes anything in this regard. Am I missing something?

AntC2:
> I opened this whole proposal as a joke.

If that's what you intended, I suggest the time to do that is April 1st, not Christmas Eve.

> A bureaucratic formality, I thought.

You mean a formality that it would get rejected as too trivial? Or would get accepted -- for much the same reason? For everything in Haskell, there's someone who'll want to debate it.

> appeal to common sense

My common sense says: Unicode chars go badly on my (Windows) machine; I don't need `..` syntax in lists or import/export specs sufficiently often to care; I can avoid writing Unicode syntax, but I can't avoid it if I'm using someone else's code; nothing's broken, don't fix it.

> ... long epistemic critique ... hypocrisy ...

Yes, for a "joke"/something trivial, the tone is getting too heated. Please keep comments light-hearted.

nomeata:
> For everything in Haskell, there's someone who'll want to debate it.

Especially when it's about syntax…

kindaro:
@AntC2   Yes, I mean that the problem does not warrant the ink and the paper being spent to discuss it. Or so I thought back then. I was expecting that there will either be immediate consensus to accept or a swift rejection for a good reason that will educate me.

Instead the big objection stems from people being unhappy about Unicode syntax in general. This does not educate me, and this is not a good reason to reject this proposal specifically — as I said above, this train has left the station a long time ago. The decision to accept Unicode was made a long time ago. I did not anticipate that I shall have to engage in re-litigation of `UnicodeSyntax`, and I do not want to.

I can see that to some _(experienced Haskell programmers)_ remembering the new `…` syntax is unwarranted effort. I can also see that to some _(people new to Haskell and programming in general)_ remembering the new `..` syntax is unwarranted effort. No one seems to be acknowledging the latter.

> My common sense says: Unicode chars go badly on my (Windows) machine; I don't need .. syntax in lists or import/export specs sufficiently often to care; I can avoid writing Unicode syntax, but I can't avoid it if I'm using someone else's code; nothing's broken, don't fix it.

I understand you and I recognize that many share this view, but this is nothing but an experience and an opinion. You are not trying to make the best possible judgement for what is good for Haskell, while accounting for all the nuance revealed in the preceding conversation. An appeal to common sense is not the same as an appeal to experience and opinion. My message was that in the absence of specific evidence we should apply other epistemic instruments, such as common sense or specialization of principles. It is common sense that a narrow view like in this quote from your message, where you intentionally ignore everything other than your own experience and opinion, is not an effective epistemic instrument. For example, you can see that I am engaging with every objection and make an effort to clarify what costs it reveals to be associated with the present proposal — even though I have some novel experience and a strong opinion. I take this to be an effective epistemic instrument and I ask you to give it a try.

> Yes, for a "joke"/something trivial, the tone is getting too heated. Please keep comments light-hearted.

Yes, you are right…  I did not recognize that _«hypocrisy»_ [has a shade of evil intent to it][1]. I do not think Tom has an evil intent. I edited the problematic message. _«Not my native language»_ defense.

[1]: https://www.merriam-webster.com/dictionary/hypocrisy

konsumlamm:
@kindaro since you want me to explain, here we go. My stance is similar to that of @tomjaguarpaw, I don't feel strongly about this proposal, as I don't use the `UnicodeSyntax` extension anyway. My confusion comes from not understanding some of your points.

> So, my claims:
> 
>     1. The ellipsis is familiar to far more people than the two dots.
> 
>     2. Having more than one name for the same entity leads to a confusion of discourse.
> 

The ellipsis may be more familiar to people in general, but I doubt that that's the case in the context of programming language syntax in particular. Most program languages use primarily ASCII, so the `…` character won't be more familiar as syntax. Isn't having more than one name for the same enttity (`..` and `…`) exactly what you're proposing here? Not that I think that's a bad thing, but it seems to contradict your argument.

> The ellipsis is not a Haskell syntax — it is everyone's syntax. Like numbers, like parentheses, like white space.

I'm not sure what exactly you mean by this. That Haskell didn't invent ellipsis? Or that ellipsis are so common that it is everyone's syntax?

kindaro:
Thank you @konsumlamm, let me try to clarify these points.

> > So, my claims:
> > 
> > 1. The ellipsis is familiar to far more people than the two dots.
> > 
> > 2. Having more than one name for the same entity leads to a confusion of discourse.
> 
> The ellipsis may be more familiar to people in general, but I doubt that that's the case in the context of programming language syntax in particular. Most program languages use primarily ASCII, so the `…` character won't be more familiar as syntax. Isn't having more than one name for the same enttity (`..` and `…`) exactly what you're proposing here? Not that I think that's a bad thing, but it seems to contradict your argument.

Yes, it is an unfortunate situation overall. Some time ago, programming languages were unable to use beautiful type setting, so they had to introduce new names for things. For example, consider that `<=` stands for the `≤` sign in many programming languages, while `=>` stands for `⇒`. Disaster! But there is a solution: we should come back to beautiful type setting, enabled by Unicode. Then we can leave things like `<=` and `..` in the past. I am hopeful that history will take this course, and with this proposal I aim to facilitate it.

Long story short, we have many names for the same thing because of legacy reasons and the way back is to:

1. Enable the beautiful type setting.
2. Gradually phase out the legacy type setting.

Specifically for `..` and `…` — we do not have the latter in the syntax of Haskell, but my case is that we mean the ellipsis when we type the two dots. This is presently being contested.

> > The ellipsis is not a Haskell syntax — it is everyone's syntax. Like numbers, like parentheses, like white space.
> 
> I'm not sure what exactly you mean by this. That Haskell didn't invent ellipsis? Or that ellipsis are so common that it is everyone's syntax?

Yes. Haskell did not invent the ellipsis, ellipsis is so common that it is everyone's syntax, and whenever we type the two dots in Haskell we can as well type the ellipsis since it has the same meaning.

kindaro:
To summarize the situation for now:

* Vladislav wants to use `…` as meta-syntax.

  - It is not clear why he thinks that this proposal will impede such usage. Waiting for him to comment on this.

  - It is not clear if it would be a big problem for many people even if this proposal will impede such usage. I also use `…` to denote code to be written sometimes, but I see no problem sticking to typed holes `_` for this purpose. Typed holes are also _(duh)_ typed and so give meaningful errors.

* Joachim has withdrawn his support in the light of Vladislav's claim that `..` is historically motivated.

  - However, Vladislav later retracted this claim:

    > > Whence does the author know that the intention of the designers of Haskell was to copy the Hoare-Ramshaw notation for intervals, and not to simply save the typing of one dot?
    > 
    > It’s a guess, to be honest. And I may have gotten the causality wrong, actually, especially now that I’m looking at the dates.

  - Also, Richard suggested that historical usage should not be taken into account.

* Joachim discovered that `..` is used in mathematical literature as notation for integral intervals. However, it seems to be rather obscure, as attested by Richard.

  - But see counterpoints №1 and №2 below.

* Joachim also discovered that `..` is used _(as notation for integral intervals, I suppose — I am not familiar with these languages)_ in Pascal and Isabelle.

  - But see counterpoints №1 and №2 below.

* No one else feels strongly about this proposal. _(Come forward if you do!)_

**Counterpoint №1:**   In Haskell, `..` is used as a notation for closed integral intervals, but also for things like `[1..]` and `[1, 3.. 10]` which are not exactly closed intervals, if intervals at all.

**Counterpoint №2:**   In Haskell, `..` is used as a notation for integral intervals, but also in import and export lists and in record syntax. To these latter cases the reasoning that refers to notation for intervals does not apply.

Did I miss anything?

tbagrel1:
AFAIK, Rust also uses the `1..3` syntax for integer ranges.

About mathematics notation, I always encountered either `⟦a, b⟧` or `1, …, n` in math courses for integer ranges; but neither `1..n` nor `1…n`. However, for what it's worth, [Wikipedia lists also `{a .. b}` and `a .. b` for integer ranges, but not `a … b`](https://en.wikipedia.org/wiki/Interval_(mathematics)#Integer_intervals).

Personally, I find that the ellipsis char (`…`) would be better suited if haskell required comma in ranges as in `1, …, n` or `1, 3, …, 9`. But given the current syntax of ranges (without commas), I prefer the `..` syntax that stands for "up to" in my mind (rather than standing for an ellipsis), at least in the `1..10` case. For `1, 3..10` my point is no longer valid, but I always found this syntax rather confusing so I don't use it.

> I can also see that to some (people new to Haskell and programming in general) remembering the new .. syntax is unwarranted effort. No one seems to be acknowledging the latter.

I disagree. For a beginner (I started haskell not so long ago), having to search how to typeset an unicode ellipsis would be a much bigger effort than just remembering that `..` is an ellipsis in one context (record destructuring), and the "to" keyword for integer ranges in another context. Most of the time, an ellipsis stands for a variable number of elements in a comma-separated list. But in integer ranges, commas are ommitted around `..` or `...`, so the ellipsis wouldn't have its usual well-understood-by-anyone meaning. Then, the user would need to attach a language-specific meaning to either `..` or `…`. It we stick to `..` and its unicode two-dots counterpart, at least we don't introduce *visual* breaking changes for experienced haskell users.

proposals/477-unicode-ellipsis.md
```diff
@@ -0,0 +1,83 @@
...
+```diff
+diff --git a/compiler/GHC/Parser/Lexer.x b/compiler/GHC/Parser/Lexer.x
+index d74d17be8f..4135372d31 100644
+--- a/compiler/GHC/Parser/Lexer.x
++++ b/compiler/GHC/Parser/Lexer.x
+@@ -1063,6 +1063,7 @@ reservedSymsFM = listToUFM $
+        ,("-<<", ITLarrowtail NormalSyntax,  NormalSyntax,  xbit ArrowsBit)
+        ,(">>-", ITRarrowtail NormalSyntax,  NormalSyntax,  xbit ArrowsBit)
+ 
++       , ("…",  ITdotdot,                   UnicodeSyntax,  0 )
```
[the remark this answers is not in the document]
kindaro: I also cannot say if this is needed.

kindaro:
Speaking of programming languages, [the situation is all over the place][1]. Some languages use the two dots, others use the three dots. Either choice is inconsistent with some widely used languages.

[1]: https://en.wikipedia.org/wiki/Ellipsis_(computer_programming)

AntC2:
> |  there's someone who'll want to debate it.
> @nomeata Especially when it's about syntax…

This proposal arrived Christmas Eve, and received little feedback -- for obvious reasons of timing. Then it got submitted to the Steering Committee. I suggest that was too early, and the proposal should be re-opened to elicit more general discussion -- well, it's getting that anyway ;-) -- and to see if a consensus arrives.

> @kindaro  people being unhappy about Unicode syntax in general.

I think you should understand why that is, rather than continually appealing to the beauty of Unicode syntax. (I've explained why I don't use it, and I've demonstrated GHCi doesn't handle either of those possible characters properly in my working environment. So my experience is not beautiful.)

> as I said above, this train has left the station a long time ago

I think you don't understand the status of extensions in GHC. `UnicodeSyntax` is not part of the Haskell Standard, so it's still standing at the station. (Yes we have a lot of platforms; also a lot of extensions still in the shunting-yard.) It hasn't even reached 'blessed' status of [GHC2021](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/control.html#extension-GHC2021). You could perhaps look back through the discussions around that, to understand why.

AntC2:
> @kindaro  | Vladislav wants to use `…` as meta-syntax.
>
> It is not clear why he thinks that this proposal will impede such usage. 

I'll guess. You can put arbitrary expressions in list range syntax:

    [1, 2+1 .. 9-2]                      ===> [1,3,5,7]          -- using two dots, taken as a range
    let (...) = (+) in [1, 2+1 ... 9-2]  ===> [1,10]             -- using three dots, taken as an operator

For that reason, you can't define a two-dot operator:

    let (..) = (+) in [1, 2+1 .. 9-2]    ===> <interactive>:2:6: error: parse error on input `..'

Vlad is defining Unicode U+2026 ELLIPSIS as an operator. (And you can define U+2025 TWO DOT LEADER as an operator -- so I'm unconvinced that's a workable alternative suggestion.)

kindaro:
@AntC2   Clearly we have a deep misunderstanding here. I am **not** going to address:
* The question of the legitimacy of Unicode syntax in general.
* The faults of your working environment in supporting Unicode.
* The situation with the Unicode syntax for lambda abstractions.
— As part of this proposal, because these points are out of its scope.

I invite you to talk to me privately if you think I can help you solve your technical or conceptual problems.

* * *

Overall, I am going to stay cool and wait for a message from Vladislav and Joachim that will hopefully clarify the situation.

nomeata:
So many words spilled over two or three small dots…

Let's see where we are. As @kindaro points out, the merits of `UnicodeSyntax` itself are not ok topic here. Hopefully that helps us to focus.

With that, the main question seems to be: is … the right symbol for where we currently use `..`. @kindaro argues yes, @int-index brought arguments forward that it might be the wrong symbol for _some_ of it's uses (ranges). I assume nobody contests that it would be an appropriate symbol for, say, the `..` in import lists.

One could propose to allow it only in these places, but I don't think it adds up: the (implementation and cognitive) complexity around that, where … can only be used instead of `..` in _some_ places are too large.

One could acknowledge that … isn't necessary the perfect symbol for ranges, but perfect is the enemy of good; the symbol isn't absurd here, and having the simple rule … = `..` is worth the little deviation from the preexisting syntax (which itself isn't _that_ ubiquitous outside of Haskell, it seems).

Or one could say that it was good to have talked about it, but given that the best path forward isn't fully clear, maybe the status quo is where we should stay for now.

@kindaro is that a fair assessment?

@int-index, what is your current sentiment about this?

int-index:
> @int-index, what is your current sentiment about this?

My stance softened somewhat once I realized that `..` is also used in imports/exports and record wildcards. I’d hate to see `…` used in ranges, and I still think it’s confusing that the ASCII counterpart of `…` would be two dots, not three, but I also must admit that `import M (T(…))` looks quite reasonable.

Overall, it seems that this change is motivated by aesthetics only and creates an unfortunate corner case (I continue to maintain that `…` in ranges isn’t right). That is simply unnecessary in my opinion, so I am still -1 on this. However, I no longer feel strongly about this, so if others are in favor, so be it.

nomeata:
Similar here, just leaning towards acceptance. Sounds like it's best solved by a committee vote, like most things aesthetics.

kindaro:
Aesthetics — yes, but of meaning, not of sight. Even in ranges the two dots denote an intentional omission _(of some enumerable values)_ without altering the original meaning — it is the rôle of the ellipsis in typography to denote that.

The motivation for inventing a new syntax _(the two dots)_ is not clear in the first place. Maybe it is simply that Donald Knuth _(if it is him who we credit with the two dots interval notation)_ had a new fancy hammer _(TeX)_ so for him nails were everywhere. Vladislav has not given any explanation for his favour of the two dots over the ellipsis as the syntax for ranges, nor has anyone else come forward.

I want it to be clear that this proposal is not to innovate, but to return to the roots, not to complicate, but to amend a complication.

Otherwise, the summary of Joachim is accurate. So far as everyone on the steering committee understands the case and has the good of Haskell in mind, I am fine with any outcome.

kindaro:
@nomeata   Hello Joachim, may I ask you to explain if there are some news I should attend to? I see a merge has occured but I am not sure what to infer from this event.

nomeata:
Ha, sorry for being so implicit: by merging the proposal and tagging it as accepted the committee declares it as accepted. You can submit a MR to GHC now if you want! :-)

kindaro:
[I have opened a merge request there.](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/8697)

PR #479, opened 2021-12-31
Magic constraint simplification type family

treeowl:
Add a magical type family to simplify constraints, primarily
for use in Template Haskell splices.

[Rendered](https://github.com/treeowl/ghc-proposals/blob/magic-simplify-constraint/proposals/0000-magic-simplify-constraint.md)


Ericson2314:
Longer term, what I really want for this is https://davidchristiansen.dk/pubs/tyde2020-predictable-macros-abstract.pdf, in which case we would have 
```haskell
simplifyConstraints :: [Type] -> m [Type] 
```
(One advantage is that when one does `-ddump-simpl`, one gets the simplified form. Another advantage is that it is more general to allow all sorts of "solving" interleaved with macro evaluation.)

But the above is a long ways off. As a stop-gap, this works fine, and the motivation is certainly familiar and resonates with me. Hell, it doesn't even need to simplify the constraint, just not issue the warnings!

CC @cgibbard, who wrote the bulk of our instance-generating TH.

treeowl:
Would simplifying the constraint cause trouble for undecidable instances, or only overlapping ones? I definitely want the warning suppression. I suppose one might want one of several things:

1. Full constraint simplification. This would actually be nice in hand-written code where you *want* the simplified constraint, but it's very long to write out.
2. Simplify constraints "gently", never letting them get (much?) bigger and limiting the number of simplification steps. I imagine this would avoid any potential undecidable instance breakage, if that's a risk.
3. Only simplify fully concrete instances (no type variables). This preserves any overlapping instances.

Ericson2314:
This might be a bit off topic, but one thing I have thought about is the interplay between simplifying constraints and modularity.

For example, if you have a `Ord (Maybe Foo)` and simplify it to `Ord Foo`, you are assuming the exististance of the `Ord a => Maybe a` instance. That becomes a "morally there" but unwritten dependency of your module, to be provided by one if its imports.

By the contravariance of `=>`, `Ord (Maybe a) => ...` is a more minimal than `(Ord a, forall b. Ord b => Ord (Maybe b)) => ...` so our usual simplification tendencies are almost exactly wrong!

The closest I can think of to a practical example where this comes up is comments I left in https://github.com/ghc-proposals/ghc-proposals/pull/276, where the idea is to get rid of today's Coercible decomposition going all the way back to the data type's parameters, and just get the coercions for the fields. This is likely to cause more breakage precisely because people write Coercible constraints minimized according to *assuming* today's decomposition rule. But the naive constraints per-field written by TH are in fact exactly what we want here, as they are also what the new weaker decomposition rule would give you!

adamgundry:
I'm worried about type soundness. Suppose:
 * module A defines class `K`,
 * module B imports A and defines `instance K Bool => K Int`,
 * module C imports A (but not B) and defines `instance K Char => K Int`,
 * module D imports B and C.

Now it would seem that `SimplifyConstraint (K Int)` reduces to `K Bool` from module B and `K Char` from module C. Thus in module D we have a potential type soundness bug.

Currently GHC won't eagerly detect such conflicting class instances, preferring to wait until they are arise during constraint solving. This is unlike type family instances, which are checked for consistency eagerly to avoid exactly this problem. I think `SimplifyConstraint` would force the same check to be performed for class instances? There's also a risk that using `INCOHERENT` instances would make it possible to violate type soundness within a single module (whereas at present they risk incoherence, but not unsoundness).

int-index:
Instead of a type family it could be a Template Haskell method `simplifyConstraint :: Cxt -> Q Cxt`. That wouldn’t risk unsoundness.

goldfirere:
Both @Ericson2314 and @int-index propose doing this right in Template Haskell, instead of as a type family. @Ericson2314 says doing so is a long ways off. But why? It looks easy enough to me. Well, maybe not all that easy, but I see no barriers to doing it. The hardest part would be impedance matching between TH and Core, but we can probably take advantage of existing code (in `GHC.ThToHs`, `GHC.Rename.HsType`, `GHC.Tc.Gen.HsType`, and `GHC.Tc.Gen.Splice`) that does some of this.

Advantages to doing this in TH:
- TH code could inspect the result.
- The simplification is done once, not every time GHC sees the constraint.
- There is no type soundness problem. (I think the type soundness problem is actually even worse than @adamgundry suggests, because you can get into trouble without overlapping instances: a downstream module might simply simplify a constraint further.)

It seems the goal here isn't writing a simple constraint -- it's having the runtime performance associated with a simple constraint. Then we have yet another alternative:
- Simplify constraints during compilation to find the most compact form for evidence, and use it.
This would, essentially, be just an optimization, possibly controlled by a flag. No need for language-level support or user intervention.

treeowl:
@goldfirere, @adamgundry, I believe an alternative to doing it in TH would be to require that the magical type families appear only as superclass or instance constraints. So

```haskell
instance (SimplifyConstraint (C x), D) => E x
class (SimplifyConstraint (F x), G) => H x
```

would be permitted, but things like

```haskell
type instance TF x = SimplifyConstraint (C x) -- Not even in a superclass/instance constraint
instance (SimplifyConstraint (J x) ~ K x) => L x  -- Not directly *as* a superclass/instance constraint
```

would not be. Would constraint *synonyms* be okay? Perhaps, but only if those are expanded eagerly enough to be sure there are no simplifications in illegal places.

A bigger problem seems to be that different situations may call for different sorts of simplification, as discussed above. For compatibility with overlapping instances, we need a version that removes fully concrete constraints (e.g., `Num Int`) and totally redundant ones (e.g., `(Eq (Either Int a), Eq (Either Int a))`) but doesn't simplify `Eq (Either Int a)` to `Eq a`. But should this version simplify constraint synonyms? What about constraint families?

If we go beyond that extremely gentle form of simplification, we'll may want a version that's guaranteed not to loop the constraint solver on undecidable instances. If we have

```haskell
instance Show (F (G x)) => Show (F x) where ...
```

and some splice generates `SimplifyWhatever (Show (F x))`, we mustn't fall over. But there's some doubt in my mind about whether it's reasonable for a TH splice to defeat instance overlap in such a manner.

Will we also want a version that is willing to risk looping, primarily for manual use, when a user wants to specify a potentially large simplified constraint by expansion from a small one?

goldfirere:
The restriction and these different designs all suggest (to me) that a type family is the wrong solution to this problem, when a TH-based solution would seem to satisfy your use-case. What is problematic about the TH approach? Or my "just make this an optimization" approach?

PR #480, opened 2022-01-09
Modify #285: Document RULES as affected by `-XNoImplicitForAll`

Ericson2314:
This feature is not yet implemented, but @goldfirere realized in #452 it
was not explicitly covered. @aspiwack thought perhaps it was simple
enough that we could do as a separate PR against the existing proposal,
and I thought that sounded reasonable so I am opining this.

Note that #448 has good positive effects with this change, but I think
it is OK in this case to ignore that for now. This is a "bug fix" that
can stand alone. When spooky effects downstream are bad, we shouldn't do
this, but then they are good, we can safely leave them be discussed
later, in this case in #448's motivation / effects and interactions
sections.


proposals/0285-no-implicit-binds.rst
```diff
@@ -194,6 +194,24 @@ This proposal applies to all alike:
...
+
+   This could be rewritten as::
+
+    {-# LANGUAGE NoImplicitForAll #-}
+    -- {-# LANGUAGE NoPatternSignatureBinds #-} -- Does not matter whether enabled or disabled
+
+    {-# RULES "name" forall a. forall (x :: Maybe a). foo x = 5 #-}
+
+   (The double-\ ``forall`` syntax separates type variables like ``a`` from term variables like ``x``.
+   This syntax is fairly obscure, but does in fact already exist.)
```
simonpj, line 213: "... does in fact already exist".   Can you say where it already exists, and give a link to the user manual entry that describes it?

simonpj:
Generally lgtm

proposals/0285-no-implicit-binds.rst
Ericson2314: I did some digging and it came from https://github.com/ghc-proposals/ghc-proposals/pull/55. (Wow, I didn't realize it's been enough years of having GHC proposals that stuff I'd need to dig up to remember would relate to them!)

It was implemented in [`512eeb9bb9a81e915bfab25ca16bc87c62252064`](https://gitlab.haskell.org/ghc/ghc/-/commit/512eeb9bb9a81e915bfab25ca16bc87c62252064).

It is documented in https://downloads.haskell.org/ghc/latest/docs/html/users_guide/exts/rewrite_rules.html#syntax . If you search for `map/id` you will get the exact spot.

-----

What of this information would you like contained within this PR proper?

simonpj: I was thinking just "...does is alreasdy part of the syntax for RULES (see the 6.19.1.1  "Rewrite rules, Syntax" <link>)."

Ericson2314:
```suggestion
   This syntax is alreasdy part of the syntax for ``RULES``, see `section 6.19.1.1 "Rewrite rules, Syntax" <https://downloads.haskell.org/ghc/latest/docs/html/users_guide/exts/rewrite_rules.html#syntax>`_ of the Users Guide.)
```

proposals/0285-no-implicit-binds.rst
```diff
@@ -194,6 +194,24 @@ This proposal applies to all alike:
...
+
+   This could be rewritten as::
+
+    {-# LANGUAGE NoImplicitForAll #-}
+    -- {-# LANGUAGE NoPatternSignatureBinds #-} -- Does not matter whether enabled or disabled
+
+    {-# RULES "name" forall a. forall (x :: Maybe a). foo x = 5 #-}
+
+   (The double-\ ``forall`` syntax separates type variables like ``a`` from term variables like ``x``.
+   This syntax is alreasdy part of the syntax for ``RULES``, see `section 6.19.1.1 "Rewrite rules, Syntax" <https://downloads.haskell.org/ghc/latest/docs/html/users_guide/exts/rewrite_rules.html#syntax>`_ of the Users Guide.)
```
Ericson2314, line 213:
```suggestion
   This syntax is already part of the syntax for ``RULES``, see `section 6.19.1.1 "Rewrite rules, Syntax" <https://downloads.haskell.org/ghc/latest/docs/html/users_guide/exts/rewrite_rules.html#syntax>`_ of the Users Guide.)
```

goldfirere:
I would much prefer *not* to have this debated by the committee. I was asked to include #285 in the mega-proposal #448. I have done so. This current proposal, #480, is covered by #448, also under consideration by the committee. Debating this one small piece separately seems to go against the spirit of the request that I write #448 in the first place.

goldfirere:
Just to be clear, the content of this proposal (though not the examples) is completely present in #448.

nomeata:
@Ericson2314, do you agree with Richard here?

Ericson2314:
I don't super care either way.

I thought the request of @aspiwack was easy enough to accommodate. In fact, this change isn't even normative, strictly speaking, as I wrote the spec itself in a general way not enumerating the instances where it applies. (In hindsight, I guess I anticipated forgetting some cases!)

Perhaps if this is quick unanimous thumbs-up, it saves time, but if there is any actual *debate* to be had just close it for #448? Just an idea in case @aspiwack still objects to dealing with in #448. Really it's between those two and I better not play awkward middleman any longer. :)

------

If I may be my broken-record self and bring up the "Aspiration Users Guide" idea again, the only thing I feel strongly here is a sense of tragedy over the choice of doing this or just doing #448 mattering *at all*.

If there was a single design document, the two options would be confluent in that the final state of the design document is the same. The git history would still faithfully record how decisions were arrived at, but the files in the top commit would just record the content of those decisions irrespective of how their were decided.

Since, however, the "file tree" is broken down by proposal, the history of decision making is recorded *twice*, and there is new view that just shows the decisions made absent how. The choice of #448 alone vs this will thus be observable.

This PR as the advantage of making clear the diff against prior consensus, leaves out the context of the larger variable binding picture. #448 has the advantage showing the whole picture, but cannot so readily make clear what's new vs what's old. There is no way to have the best of both worlds with the current system.

goldfirere:
For what it's worth -- and this is up for debate -- I was thinking that, if #448 is accepted, we would remove all the files it supersedes, including the proposal from #126, #155, #285, and a few others. Or maybe not delete -- I don't like the idea of removing others' work -- but put links at the top of pointing to the new text.

Ericson2314:
@goldfirere Oh! Very gld to year that.

> Or maybe not delete -- I don't like the idea of removing others' work -- but put links at the top of pointing to the new text.

That or simply linking prior version when it was intact.

If you were thinking of that, then I recommend what I wrote in https://github.com/ghc-proposals/ghc-proposals/pull/448#issuecomment-1009207109 if doing it *beforehand* on a branch with just the status quo parts, and then transforming that preliminary combined version into what you actually want on the #448 PR branch.

Once #448 is merged, the effect is the same as what you say, but while #448 is discussed we get to read the diff between the two.

If you are interested, I would be happy to in any way I can. Modern scoped type variables is the best chance I've gotten to put the "living documents with diffs are better" approach put to the test in the years since GHC proposals began, and with that in mind nothing might come up better for years to come, too.

aspiwack:
> For what it's worth -- and this is up for debate -- I was thinking that, if #448 is accepted, we would remove all the files it supersedes, including the proposal from #126, #155, #285, and a few others. Or maybe not delete -- I don't like the idea of removing others' work -- but put links at the top of pointing to the new text.

I think it would be hard as, unless I'm mistaken, a lot of the motivations, examples, alternatives, etc… for all these proposals haven't been ported to #448 .

The reason why I asked for this PR is that 1/ it was presented as a technical omission (which doesn't require debate, just a quick double-check), and this PR convinces me that it is (I don't even think it needs the committee to look at it, we could just merge as is, because of course this is what the committee accepted) 2/ as far as I can tell, this omission is the only reason why `-XNoImplicitForAll` is mentioned in #448 , and I'd like it to shed a bit of weight (because, as I said on the thread, #448 is pretty overwhelming in its current state).

goldfirere:
While others can, of course, make separate PRs (such as #480), I'm not willing to start removing parts of this proposal, given that I was asked explicitly to include all the proposals wrapped up in this one. @simonpj @nomeata, I believe you are the ones who requested this mega-proposal.

Maybe here's a way forward:
* Accept and merge #452 (the Principles PR)
* Reframe the principles in this proposal as a diff against the Proposals document.
* Remove the parts of this proposal that repeat other proposals. Instead, edit those proposals (in this PR) to refer to those new principles.

Another plan would be to make this PR solely about the principles. Then have other PRs take on the individual changes. Maybe even the first parts are only conditionally accepted until all the pieces are accepted (or rejected)? I still worry that doing this plan would then create "but we can't understand these principles in the abstract -- how would they affect language design in practice?"... leading to another mega-proposal... leading to another big glob that's hard to absorb.

nomeata:
With “this proposal“, you mean #448, not #480, right?

goldfirere:
Yes -- I've clarified. I see how that was confusing!

Ericson2314:
@goldfirere I think most/all the "this" should refer to #448?

I do basically agree: #448 should be comprehensive in somehow laying out the whole picture, regardless of whether #480 is merged. I do in fact like doing both --- fix the "bug" and still be comprehensive!

Ericson2314:
Also agree #452 should be accepted so we can start getting things more presentable either way, and I do like as a general policy when a new principle is added (with a new proposal, per the no "bare" principles rule), existing stuff that upholds it is made to connect to it.

PR #485, opened 2022-02-21
Add applyWhen and nTimes to Prelude

schuelermine:
This is a proposal augmenting Prelude by two general functions.

[Rendered](https://github.com/schuelermine/ghc-proposals/blob/patch-add-applyWhen-and-nTimes-to-Prelude/proposals/0000-add-applyWhen-and-nTimes-to-Prelude.md﻿)


nomeata:
Thanks for your contribution! It seems that this would rather require a proposal with the Core Libraries Committee, see https://github.com/haskell/core-libraries-committee

schuelermine:
OK, I will go there

PR #491, opened 2022-03-02
Support multiple hole meanings in types

treeowl:
Holes in a type can either mean that the user does not know what belongs
there or that they don't care. At present, this can only be controlled
for type signatures at the module level, and not at all for type
applications. I propose a simple way to distinguish these holes
syntactically.

[Rendered proposal](https://github.com/treeowl/ghc-proposals/blob/different-holes/proposals/0000-distinguish-partial-sigs.md)


adamgundry:
This seems like a good idea to me. Might we also consider `_` in patterns a "don't care" hole? That strikes me as a good reason to prefer `_` for "don't care" and use different syntax for "don't know" (as indeed the proposal does).

treeowl:
> This seems like a good idea to me. Might we also consider `_` in patterns a "don't care" hole? That strikes me as a good reason to prefer `_` for "don't care" and use different syntax for "don't know" (as indeed the proposal does).

Perhaps as a mnemonic device, but I think it's different enough not to want to lump it in the same category.

simonpj:
Is the **only** difference betweeen "don't know" and "don't care" that
* for the former you get a message saying "this hole stands for type Int->Int"
* whereas for the latter you do not get a messagse,
But they are in all other respects they are identical?

If so, please so so very explicitly.  Thanks!

treeowl:
@simonpj, that's a good question. Yes, that's the only immediate difference. It would be interesting to consider whether the intention each signals could be useful to improve type error messages (or the wording thereof) in some cases, but that's not part of the proposal.

simonpj:
> Yes, that's the only immediate difference

Are there any non-immediate differences?

Could you amend the proposal to state this property more clearly?  I read it quite carefully but could not figure it out for sure.  It's really a per-site message-suppression mechanism.

michaelpj:
> Yes, I think _ in terms should be considered a "don't care" hole for consistency, but GHC will need to emit an error/warning anyway since it doesn't have a way to infer a value.

This seems like it will be annoying to explain. We have this nice consistent syntax except where we don't.

I would be happy if we added the new syntax for term holes too, and gave appropriate messages corresponding to the intent:
- `__` in a term gives the message as today (?)
- `_` in a term says "I can't infer this value for you, did you mean .... ?"

Then we would nudge people towards using "don't know" term holes gradually, but it wouldn't break anyone using "don't care" holes.

treeowl:
@michaelpj, there is no difficulty, really. A term hole today is anything starting with an underscore. You can use `_`, `__`, `_banana`, etc

goldfirere:
This is a simpler restatement/redesign of part of #194. Interestingly, #194 also settled on using `__` for wildcards where we want a diagnostic and `_` for ones where we don't.

I'm not suggesting to close this in favor of #194, which grew in complexity. Simplicity is valuable! But I do think @treeowl may find a read of #194 interesting.

Anrock:
Bikeshedding: can we use `_?` for "don't knows"s? Double underscore is hard to distinguish visually from single underscore and question mark looks like a question "what type goes here?".

treeowl:
@Anrock that's a bit tricky because `?` is an operator, though it is possible. Also, it looks a little weird with named wildcards: `_?hello`. I wouldn't reject it out of hand, but it seems a bit odd.

Anrock:
@treeowl I guess the operator part can be fixed by requiring to have a space before operator character? And named wildcards by by parsing like `underscore`, `whatever string`, `question mark`. Though I'm not insisting, just an idea.

adamgundry:
@treeowl what's the status of this proposal, is it ready for submission? I think "don't care" holes would be a really useful feature, so it would be nice to make progress on this.

I'm wondering how often one would want to use a named wildcard for a "don't care" hole. Presumably it allows one to say "I don't care what this is, but it must be the same as that". But a possible simplification to the proposal might be to say that `_` is the only "don't care" hole, whereas named wildcards are always "don't know" holes? And it makes a certain kind of sense that the name gives a way to relate the answer back to the question being asked.

treeowl:
I forgot all about it. I definitely think we should offer named "don't care" holes.

nomeata:
@treeowl, are you by any chance attending Haskell Symposium? If so, would you like to present this proposal at the [GHC Proposals session](https://icfp22.sigplan.org/home/haskellsymp-2022#GHC-Proposals-Session)?

PR #492, opened 2022-03-13
Fix date-accepted on proposal 0366

andregrigon:
Seems like a typo, as the proposal was accepted on 2020 (see https://github.com/ghc-proposals/ghc-proposals/pull/366#issuecomment-727981091).


nomeata:
Thanks!

PR #493, opened 2022-03-23
Allow SPECIALISE to work with expressions

goldfirere:
[The proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0493-specialise-expressions.rst) has been accepted; the following discussion is mostly of historic interest.

---

@simonpj and I were surprised this morning when e.g. `{-# SPECIALISE (+) @Int #-}` didn't work. And then we realized that we could handle arbitrary expressions (headed by a top-level function), not just type specializations. And thus this proposal.


nomeata:
Wasn't the part about type applications not discussed before? Maybe a ticket?

Ericson2314:
This is great! An excellent example of the vantage of point of "why should types and terms be different?" finding existing gaps in the language.

> Create a new top-level binding behaving as f' vars = inline exp, where inline is from GHC.Exts and causes its argument to be unconditionally inlined (see the [documentation](https://hackage.haskell.org/package/base-4.16.0.0/docs/GHC-Exts.html#v:inline)).

I think what is missing is that the expresion must be able to reduce somehow.\

e.g.
```haskell
{-# SPECIALISE addMult #-}
````
should probably be an error or warning

Same with
```haskell
{-# SPECIALISE forall f. addMult f #-}
````
```haskell
{-# SPECIALISE (True, False) f #-}
````
Not so sure about
```haskell
{-# SPECIALISE (addMult @Double, False) f #-}
````
i.e. when the reduction isn't the outermost part of the expression.

**Comment by goldfirere** [comment]
Yes, `{-# SPECIALISE addMult #-}` and `{-# SPECIALISE forall f. addMult f #-}` should be warnings. I don't see a reason to make them errors. (For example, it's plausible that some automated tool would produce that, and it's harmless.)

The middle two of your examples are errors. `SPECIALISE` works only when the expression is "a top-level variable applied to arbitrary expressions".

**Comment by Ericson2314** [comment]
> a top-level variable applied to arbitrary expressions

Sorry! I just missed that completely. All makes sense now. I am less concerned about the warning now, to be honest, because

> The expression in a `SPECIALISE` pragma must be suitable for the left-hand side of a rewrite rule. 

Indeed prevents the silliest uses. `inline` (and `noinline`) I think could be further improved to be more fail-false, but that is orthogonal to this proposal. `SPECIALISE` has a nice desugaring in terms of `inline` and `RULES`, and that is good enough.

**Comment by simonpj** [comment]
Should allow underscores
```
{-# SPECIALISE foo _ (5 :: Int) #-}
```
rather than requiring
```
{-# SPECIALISE forall x. foo x (5 :: Int)  #-}
```

**Comment by AndreasPK** [comment]
I wanted to write a proposal like this today at some point but it seems I no longer have to!

I agree with simon on underscores. In fact he wrote this after we had talked about the proposal!
It's already common to use _ as "we don't care about the value of this variable" in many places and this would very naturally transfer to these pragmas.

The rule would be that in an specialize pattern any argument represented as an underscore will remain an argument of the specialized function. e.g `addMult _ 0` will produce a function with one argument.

This could be expanded to types with @_ ignoring a type argument.

This means we could go even further! We could then, having freed the forall syntax up give the pragma:

`{-# SPECIALIZE forall x. foo x y #-}` the meaning of "At every call site, *always* try to specialize for the first argument. And it would *finally* allow us to say: {-# SPECIALIZE forall a. foo @a #-} which tells GHC to specialize foo for every time it encounters if possible (and presumably warn if not possible).

And to me that makes perfect sense as well. Because we really do say for any argument to this function, specialize on that argument. But I can see how this might be harder to implement.

----------------

> In this case, GHC will create
> ```
> addMult5 :: Int -> Int -> Int
> addMult5 y = 5 * 2 + y
> {-# RULES "addMult/5" addMult 5 = addMult5 #-}
> ```
> GHC can then further optimize the right-hand side of addMult5 to avoid doing any multiplication at run time.

I think it should generate `addMult5 y = let x = 5 in 5 * 2 + y`. Otherwise it's unclear if this is allowed to duplicate the expression or not. And I think it should be allowed.

-----------

Motivation:

Maybe list some more advantages. Removing the branch is just one thing!

Strictness:

frob is (potentially) lazy in both x and y. If we specialise for the boolean flag one version would be strict in at least x and the other in at least y (independen of what large actually is). Which can potentially allow W/W to fire on this argument allowing good things to happen.

```
frob b x y = 
   if b then x `seq` large1
         else y `seq` large2
```

CPR:

If large1 can be CPRed but large2 can't. Then specializing for the large1 code path will produce a function that can be CPRed.

Inlining: The specialized functions are naturally smaller, so more likely to inline (generally considered a good thing).

Loop breakers:

```
{-# SPECIALIZE f True #-}
f b x y =
   if True then <expr not mentioning f>
           else  <expr mentioning f>
```
Specializing `f` here for `True` will generate a non-recursive specialization, so any call to `f True` is now a potential for inlining (of the specialized rhs)

-----

Code size:

It can also *reduce* code size. Currently one can achieve (most of the time) the same goal by forcing the target function to inline into every call site.  But there might be call sites where inlining has litte or no benefit at all! And I have seen (and written!) code like this in the past. So providing a manual specialize pragma would be much lighter in code size and would have at least some uses.

Although in practice we should expect code size to increase overall if this feature finds wide adoption.
------

> The syntax allowing multiple types ...

I don't know about this syntax. Perhaps it would be good to show what we will be losing if we give this up with an example?`

-----

> As today, SPECIALISE pragmas may be written only at top-level or in a class or instance declaration, never in a let or where.

Is there a good reason for this limitation? I think we should aim to allow specializing local functions on terms as well. Because why not! SpecConstr already does it so we know it's possible.

**Comment by AndreasPK** [comment]
> Create a new top-level binding behaving as f' vars = inline exp

This likely doesn't do what we want for recursive functions. If we have

```
loop b x y =
   if b then loop b x (y+1)
        else loop b (x+1) y
```

I think this would result in this function if specialized for True after one run of optimization:

```haskell
$sloop x y = loop True x (y+1)
```

Of course we would end up inlining `loop` in every iteration of the simplifier as well. But my main point is that the body will have a call to the unspecialized version. So you don't gain anything beyond the first iteration.

I think SpecConstr already handles this correctly so we do have code to do this correctly. But defining this with the semantics of `inline` isn't what we want.

nomeata:
> We could then, having freed the forall syntax up give the pragma:
> 
> `{-# SPECIALIZE forall x. foo x y #-}` the meaning of "At every call site, always try to specialize for the first argument. 

Is that `y` a “variable not in scope here”? I’d be wary of a syntax that changes meaning if things suddenly come into scope. Or am I confused?

goldfirere:
I don't think underscores are powerful enough to cover this feature. For example, I might want

```hs
{-# SPECIALISE forall x. f x x #-}
```

or perhaps more realistically

```hs
{-# SPECIALISE forall a. forall. g @a @a #-}
```

if `g` is a polymorphic function that behaves better when both types are the same. This repeated-variable case isn't covered by underscores.

Separately, I think it would be a nice generalization of RULE syntax to allow underscores on the left, but I think that can be described in a separate proposal that covers both this new feature and the existing one.

I *am* sympathetic to @AndreasPK's desire for the feature he describes, but I think the syntax should be something other than `forall`.

As for motivation, etc.: please add this! I've added you, @AndreasPK, as a collaborator to my fork. Just commit away. Thanks!

Ericson2314:
Suppose I have
```haskell
f :: (MonadFoo m, MonadBar m) => ...
```
I would like to be able to do something like
```haskell
{-# SPECIALIZE forall c. forall. f @(ReaderT c IO) # -}
```
Where some `(MonadFoo (ReaderT c IO), MonadBar (ReaderT c IO))` constraint would be inferred. Will this be possible?

I think this would be very nice for making MTL stuff be faster.

goldfirere:
@AndreasPK It sounds like you're proposing allowing a leading underscore to allow a variable to be implicitly brought into scope. I personally don't like this idea -- it seems at odds with much other design -- but you could consider making a separate proposal that affects RULES in the same way (for the same reasons). Given that this proposal builds on RULES syntax, I would expect any change to RULES syntax to also affect these new specializations.

@Ericson2314 The case you describe doesn't work too well: how would the specializer know whether to apply the rule? It would somehow need to decide based on whether the type variables meet the constraints, and that goes well beyond what today's rules can do.

Ericson2314:
@goldfirere I am certainly grasping around in the dark here, but I was hoping the rule could apply whenever the `m := ReaderT c IO` substitution is valid, ignoring the constraint.

My rough idea is that the constraints are "just" what one gets from that substitution and so they would need to be solved for anyways, and so the rule activation is safe to ignore the constraints knowing the presence/absence of the rule firing will never make the solving harder/easier.

No worries if this is just out of scope of this proposal, but I would be very interested if there was a way we could make something like that go.

Ericson2314:
Great, glad to hear it!

goldfirere:
I'm afraid to say I had lost track of this one. I'd like to submit to the committee, @nomeata.

@AndreasPK offers some tantalizing expansions of the idea here. But I deem these to be expansions of `RULES` syntax in general, and somewhat orthogonal to this proposal. It's all synergistic, in that this proposal makes that syntax more powerful, and thus @AndreasPK's ideas would be more widely applicable. But I still think it's a separate proposal, and that this one can stand alone.

adamgundry:
Shepherd here. This proposal seems basically solid to me: being able to use `SPECIALISE` with a type application rather than a type signature feels so natural I'm almost surprised it wasn't already possible, especially since it enables specialising functions with ambiguous types. The extension to specialising applications in a SpecConstr-like fashion seems reasonable.

One outstanding point seems to be @AndreasPK's https://github.com/ghc-proposals/ghc-proposals/pull/493#issuecomment-1153326382. The proposal could do with some explanation of how recursive functions interact with the extended `SPECIALISE`, and whether using `inline` gives the right semantics. I'm not an expert in this area, but a case I wondered about is unrolling loops over static data at compile time, like this:
```hs
{-# SPECIALISE loop [] #-}
{-# SPECIALISE forall x xs . loop (x:xs) #-}
loop :: [Int] -> Int
loop [] = 1
loop (x:xs) = x * loop xs

main = print (loop [1,2,3,4,5])
```
If I understand the proposal correctly, this will produce
```hs
loopNil = inline loop []
loopCons x xs = inline loop (x:xs)

{-# RULES "loop/loopNil" loop [] = loopNil #-}
{-# RULES "loop/loopCons" forall x xs . loop (x:xs) = loopCons x xs #-}
```
which does not optimise away the loop (presumably because the inliner sees the recursion and is reluctant to inline despite the `inline` magic function?). On the other hand if we were to produce
```hs
{-# RULES "loop/nil" loop [] = 1 #-}
{-# RULES "loop/cons" forall x xs . loop (x:xs) = x * loop xs #-}
```
then it seems the loop would get optimised away at compile time. Although that presumably introduces the corresponding risk that a programmer might accidentally turn a runtime infinite loop into a compile-time infinite loop...

simonpj:
I support acceptance.  But I think the proposal does itself no favours by piggy backing on `infixexp` and `inline`.   The recursion point above is one problem.  Another is that the proposal appears to allow
```
{-# SPECIALISE forall x. (if x then 3 else 4) #-}
```
which is silly.

Let me be more construtive.  Be explicit about syntax.  So
```
pragma ::= ...
        |  '{-#' specialise_keyword activation rule_foralls specexp '#-}'


specexp ::= lexp qop infixexp
         | specapp

specapp ::= qvar
         | specapp aexp
```
(I'm not sure if type arguments are in `aexp` but you can check.)

This still allows silly things like
```
{-# SPECIALISE f (let v=22 in Just v) #-}
```
Perhaps instead of `aexp` arguments we should insist on arguments that
consist of applications, literals, and variables (incl data constructors)?

To be fair, the grammar for rewrite rules is equally silly:
```

rule    :: { LRuleDecl GhcPs }
        : STRING rule_activation rule_foralls infixexp '=' exp
```
but there is no reason to perpetuate this!  The compiler actuallly rejects complicated LHSs,
but somehow the specification (i.e. this proposal) should say what is acceptable and what is not.

So, after desugaring infix syntax, the LHS looks like `f arg1 .. argn`.
Given the definition `f = rhs` (possibly recursive) make the new definition
```
  f' = \ bound-bars. rhs arg1 .. argn
```
If `f` was recursive, this defn will mention `f'`, but that is fine.  No need to involve `inline` at all.

adamgundry:
> If `f` was recursive, this defn will mention `f'`, but that is fine. No need to involve `inline` at all.

Did you mean that it will mention `f`? I don't see how it will mention `f'` directly. But I think that's probably fine, because presumably you have
```hs
{-# RULES "f/f'" forall bound-vars . f arg1 ... argn = f' bound-vars #-}
```
so then any recursive occurrences of the right call pattern will match the rule.

simonpj:
> Did you mean that it will mention f? I don't see how it will mention f' directly. 

Correct, fixed.  And yes the RULE may (or may not!) fire on the calls to `f` in `rhs`.

adamgundry:
@goldfirere Thanks for this proposal. I'm marking it as in need of revision, so that Simon's comments in https://github.com/ghc-proposals/ghc-proposals/pull/493#issuecomment-1563611832 can be addressed. Once that's done, I think it will be good to go.

rhendric:
> `SPECIALISE` has a nice desugaring in terms of `inline` and `RULES`, and that is good enough.

That observation hadn't occurred to me before reading this proposal, and it made me wonder if this feature could be implemented entirely in Template Haskell. I don't know if this is relevant to the proposal itself, or if it's only of interest to those of us eager to have this functionality, but I've sketched out a demo of how this might look [here](https://play.haskell.org/saved/BgCbY08r).

Intended usage:
```haskell
addMult :: Num a => a -> a -> a
addMult x y = x * 2 + y
{-# NOINLINE [1] addMult #-}
sPECIALISE NoInline [| addMult @Int |]
-- or, to get the equivalent of SPECIALISE forall
-- sPECIALISE NoInline [| \(x :: Int) -> addMult x |]
```

The `NOINLINE [1]` pragma seems to be necessary per `-Winline-rule-shadowing`, which is a bit of a downside. But the obvious upside is not having to hack GHC to try this idea out and iterate on it.

Big caveat that this is just a demo and may not do everything we'd want the full proposal to do, may cause GHC or the compiled program to run in circles, etc.

Rewbert:
I can jump in here and mention that I independently ended up in a place where I desire this functionality. I am trying to encourage GHC to perform a specific computation at compile time by manipulating the source with rewrite rules (it still feels wild to me that you can emit RRs with TH!). This is 100% donkey code, but it seems to work much better than I expected lol.

Short version: There are two problematic features that I can overcome individually.
1. I can apply my RRs to polymorphic functions by encouraging GHC to inline it (the parameters would then be at hand, and my RRs can fire).
2. I can apply my RRs to recursive functions because my rules manipulate the source, and don't actually run the function (if I had to actually run it, there is a chance that it wont terminate).

With my current strategy I can not handle the _combination_ of a polymorphic and recursive function, because GHC will not inline it. I do however know all possible parameters the function might be applied to (which is a relatively low number), and if I could just specialize it to the potential inputs and throw the old `-O3` into the mix, I am confident that I can overcome this as well.

I hope this is not too vague, but it would probably take more time than I have for me to synthesize an OK example that illustrates the problem.

---

I would not mind trying to hack on this. I've never contributed to GHC before, but this seems like a good place to begin. A lot of code is already in place, and the scope of the proposal is reasonable.

goldfirere:
I would love to have the time to push this through, but that's hard to come by. @simonpj has asked to take this one over, and I've granted that permission.

simonpj:
I have revised the proposal, including adding more examples. 

@adamgundry could you have another look?

adamgundry:
Thanks @simonpj! This proposal looks good, and I have recommended acceptance to the committee.

proposals/0000-specialise-expressions.rst
```diff
@@ -0,0 +1,248 @@
...
+Note that ``addMultInt`` uses no indirection in its use of addition and multiplication,
+unlike ``addMult``, which makes two calls to dictionary functions, confounding branch
+prediction and slowing down execution. The ``RULE`` means that GHC will replace
+all calls of ``addMult`` at ``Int`` to use the faster version.
+
+The syntax of today's ``SPECIALISE`` requires a function and its specialised type.
+
+This proposal suggests to generalise the syntax of ``SPECIALISE`` to allow an
+arbitrary expression, not just a function name and a type. Happily, this is backward
+compatible, because type signatures are expressions, too. This will allow, for example ::
```
adamgundry, line 37: Not quite an arbitrary expression, right?
```suggestion
This proposal suggests to generalise the syntax of ``SPECIALISE`` to allow a
function application with an optional type signature, not just a function name and a type. Happily, this is backward
compatible. This will allow, for example ::
```

proposals/0000-specialise-expressions.rst
```diff
@@ -0,0 +1,248 @@
...
+
+  {-# SPECIALISE addMult @Double #-}
+  {-# SPECIALISE addMult (5 :: Int) #-}
```
adamgundry, line 40: What about
```haskell
{-# SPECIALISE addMult 5 :: Int -> Int #-}
```
which I think is currently forbidden by the syntax, but could be accepted?

proposals/0000-specialise-expressions.rst
```diff
@@ -0,0 +1,248 @@
...
+     pragma ::= ...
+             |  '{-#' specialise_keyword activation rule_foralls specexp '#-}'
+             |  '{-#' specialise_keyword activation qvar '::' types1 '#-}'  -- as today
+
+     specialise_keyword ::= 'SPECIALISE' | 'SPECIALIZE' | 'SPECIALISE INLINE' | 'SPECIALISE INLINE'
+
+     specexp ::= qvar
+              |  specexp aexp
+              |  specexp '@' atype
+
```
adamgundry, line 95: This would allow `{-# SPECIALISE addMult 5 :: Int -> Int #-}`, although it makes `{-# SPECIALISE qvar :: type #-}` ambiguous. (Perhaps we can resolve the ambiguity by having the "backwards compatibility alternative" kick in only when there are at least two types?)

```suggestion
              |  specexp :: type
              
```

proposals/0000-specialise-expressions.rst
simonpj: yes!

proposals/0000-specialise-expressions.rst
simonpj: I have fixed this

simonpj:
Thanks @adamgundry for pointing this out... I have added an optional type sig for the new form, now labelled (1).

I would welcome opinions about whether to get rid of the multi-type form (2)
```
{-# SPECIALISE f :: Int -> Int
                          , Bool -> Bool
                         , Float -> Float #-}
```
This is implemented today, but is not documented in the user manual, and it's a tiresome exception in the new scheme.   

Even though it's a breaking change, I would argue for deleting (2).  We could easily do an impact study on Hackage... but of course Hackage isn't everything.

adamgundry:
I think it would be fine to deprecate the old syntax, introduce a warning recommending multiple pragmas instead, and remove it after a few releases.

nomeata:
As per our recent shift to more stability, it seems prudent to keep supporting the old syntax. GR1 applies, and I'm not sure GR2 does? (Yes it's tiresom…)

If it's actually tricky to support the union of the syntaxes, we could also introduce a language extension for just the new syntax, then the implementation has only to support one or the other variant.

simonpj:
> As per our recent shift to more stability, it seems prudent to keep supporting the old syntax. GR1 applies, and I'm not sure GR2 does? (Yes it's tiresome…)

It makes a good example:

* It's not *difficult* to continue to accept the old syntax.... just a small crufty wart, and I don't like those perpetuating themselves indefinitely.
* The impact of removal is likely to be vanishingly small.  After all, GHC has never claimed to do it.

So maybe (GR2) is a bit more nuanced than the current proposal says :-)

nomeata:
Sure, but isn't that exactly the attitude (which I personally share!) that we aim to change? No breaking change without compelling reason is rather clear on that.

I do agree that this will be annoying in plenty of cases like this, but if we add an implicit “unless we don't feel like it” to GR1, it's quite the status quo, and not the cultural shift that (according to my understanding) we want to undergo.

Or we declare `SPECIALISE` to be experimental and all is well :-)

adamgundry:
I think abstract principles are not necessarily going to help us much, because at the end of the day it's a judgement call. But one thing that is clear is that we shouldn't simply remove support without a deprecation warning (per GR3). So we could add that warning, without necessarily committing to a concrete schedule for removal. Then in the future the feature can be removed if it is actively getting in someone's way (e.g. by impeding some refactoring).

nomeata:
But then we already have spent the effort of supporting both together, have we? At that point at least my gut feeling is that implementing the warning, adding a flag for it, documenting it, etc. is more effort than just dragging the old syntax along forever.

adamgundry:
> my gut feeling is that implementing the warning, adding a flag for it, documenting it, etc. is more effort than just dragging the old syntax along forever.

That may be so, but I propose we leave that judgement call to whoever is implementing this. That is, we could accept the proposal on the basis that the old syntax is deprecated and *may* (but not *must*) be removed in the future, independently of the implementation of the rest of the proposal. Thus removing it would not require a further committee decision, merely implementation work to add the warning, etc. etc.. Like any proposal, that work will happen only if someone steps up to implement the change.

simonpj:
> That may be so, but I propose we leave that judgement call to whoever is implementing this. That is, we could accept the proposal on the basis that the old syntax is deprecated and may (but not must) be removed in the future, independently of the implementation of the rest of the proposal. Thus removing it would not require a further committee decision, merely implementation work to add the warning, etc. etc.. Like any proposal, that work will happen only if someone steps up to implement the change.

Sounds right to me.   

(Remember, this is only a corner of the proposal.)

If Adam's suggestion is acceptable, I'll amend the proposal to reflect it.

int-index:
We should follow the procedure: a deprecation warning first, then the breaking change. A warning flag is not that much of a tax on the implementation.

No specific wording in the proposal is required to say that some part of it will be implemented only if someone implements it. That's just the reality of it.

So I believe the proposal doesn't need to say "may (but not must)" or anything of this sort. It can simply prescribe to remove the alleged misfeature, specify the name of the warning flag, its group (Wcompat or Wall), and the deprecation period (3 releases, I imagine).

simonpj:
> If Adam's suggestion is acceptable, I'll amend the proposal to reflect it.

I have updated the proposal to say that we will deprecate the (undocumented) form
```
{-# SPECIALISE f :: Int->Int, Bool-> Bool #-}

File: proposals/0000-specialise-expressions.rst [hunk 4]
```diff
@@ -0,0 +1,254 @@
...
+specialised versions.
+
+Proposed Change Specification
+-----------------------------
+
+1. Here is the new BNF for ``SPECIALISE`` pragmas::
+
+     pragma ::= ...
+             |  '{-#' specialise_keyword activation rule_foralls specexp [ '::' type ]'#-}'  -- (1)
+             |  '{-#' specialise_keyword activation qvar '::' type ',' types1 '#-}'          -- (2)
```
adamgundry, line 90:
```suggestion
             |  '{-#' specialise_keyword activation qvar '::' type ',' types1 '#-}'          -- (2) DEPRECATED
```

proposals/0000-specialise-expressions.rst
```diff
@@ -0,0 +1,254 @@
...
+     types1 ::= types1 ',' type
+            |   type
+
+#. The first production (1) includes ``{-# SPECIALISE f :: type #-}`` as a special case
+   in which there are no arguments.  The ``:: type`` part is optional, because it is often
+   unnecessary if arguments are supplied.
+
+#. The second production (2) is there only to support the current (implemented but entirely undocumented)
+   possiblity of having multiple types in one ``SPECIALISE`` pragma. With this proposal, GHC will
+   deprecate this form, and remove it altogether after two compiler releases.
```
adamgundry, line 116:
```suggestion
#. The second production (2) is there only to support the current (implemented but entirely undocumented)
   possibility of having multiple types in one ``SPECIALISE`` pragma. With this proposal, GHC will
   deprecate this form. A new warning `-Wdeprecated-pragmas` (in `-Wdefault`) will be emitted when it is used,
   and GHC may remove support for the syntax altogether after at least three major releases with the warning.
```

proposals/0000-specialise-expressions.rst
```diff
@@ -0,0 +1,254 @@
...
+   users to avoid repetition with no runtime cost.
+
+#. Given how this builds on the existing machinery so nicely, the implementation burden
+   is expected to be small.
+
+#. The syntax allowing multiple types to be specified is not documented in the
+   `manual <https://downloads.haskell.org/ghc/latest/docs/html/users_guide/exts/pragmas.html?highlight=specialise#specialize-pragma>`_
+   and is rarely used, according to a `Hackage search <https://hackage-search.serokell.io/?q=SPECIALI%5BSZ%5DE.*%2C>`_.
+   However, there seems to be no great need to remove the syntax, and so
+   this proposal leaves it untouched.
```
adamgundry, line 239:
```suggestion
```

proposals/0000-specialise-expressions.rst
```diff
@@ -0,0 +1,254 @@
+Allow expressions in SPECIALISE pragmas
+=======================================
+
+.. author:: Richard Eisenberg
```
adamgundry, line 4:
```suggestion
.. author:: Richard Eisenberg, Simon Peyton Jones
```

adamgundry (commented):
Thanks @simonpj. I've made a few suggestions, primarily tweaks as a consequence of deprecating the multi-type form (including being more explicit about the deprecation process).

simonpj:
> Thanks @simonpj. I've made a few suggestions, primarily tweaks as a consequence of deprecating the multi-type form (including being more explicit about the deprecation process).

Thanks.  I have implemented these changes.

adamgundry:
Thanks @simonpj, I declare this proposal to be accepted.

PR #496, opened 2022-03-24
Allow record wildcards for empty records (under review)

Ericson2314:
Allow record wild cards for empty records. This removes and artificial restriction, making the language simpler and more predictable.

[Rendered](https://github.com/Ericson2314/ghc-proposals/blob/empty-record-wildcards/proposals/0000-empty-record-wildcards.rst)

### DIFF
--- a/proposals/0000-empty-record-wildcards.rst
+++ b/proposals/0000-empty-record-wildcards.rst
@@ -1,22 +1,7 @@
-Notes on reStructuredText - delete this section before submitting
-==================================================================
-
-The proposals are submitted in reStructuredText format.  To get inline code, enclose text in double backticks, ``like this``.  To get block code, use a double colon and indent by at least one space
-
-::
-
- like this
- and
-
- this too
-
-To get hyperlinks, use backticks, angle brackets, and an underscore `like this <http://www.haskell.org/>`_.
-
-
-Proposal title
+Allow record wildcards for empty records
 ==============
 
-.. author:: Your name
+.. author:: John Ericson (@Ericson2314)
 .. date-accepted:: Leave blank. This will be filled in when the proposal is accepted.
 .. ticket-url:: Leave blank. This will eventually be filled with the
                 ticket URL which will track the progress of the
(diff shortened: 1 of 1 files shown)


AntC2:
Yes current behaviour seems inconsistent. For example this record syntax is accepted, even though `Nothing` has no fields:

    fN Nothing{ } = 1

Ericson2314:
Good point thanks. @adamgundry mentioned it too in https://github.com/ghc-proposals/ghc-proposals/issues/484. I forgot to include it. Will do now.

AntC2:
Oh, and here's a weirder inconsistency. These are accepted even though the `..` can't denote any fields:

    data Foo = Foo { a :: Int }

    fa (Foo{ a, ..}) = 1

    xa = Foo{ a = 2, ..}

Ericson2314:
Cover it in which way? Do we want a change, or just noticed it matches how those are accepted?

AntC2:
> Cover it in which way? Do we want a change, ...

I think @goldfirere just means include it in the Motivation as another example of how the current banning `{..}` for constructors without fields is inconsistent (with H98 syntax, and with using `..`). So no, no change to the Proposed Change. (Except hehe, writing this out got me to thinking ...)

* In H98, using braces `MkFoo{ }, Nothing{ }, Just{ }` doesn't imply that `MkFoo` has any fields, or that they were declared with record syntax -- neither in patterns nor expressions.
* With `-XRecordWildCards`, `MkFoo{..}` currently insists that `MkFoo` does have fields and was declared with record syntax.
* Even though the `Foo{ a, .. }` example shows the `..` doesn't have to denote any field bindings.
* So `Nothing{ .. }` ought to imply only that _if_ there are fields, they were declared with record syntax; **Or** do we want to go stronger ...
* ?Propose `Just{ .. }` be accepted, but doesn't make any bindings. Hmm hmm what would that mean in a expression/build context? Note that `Just{ }` is valid in an expression ≡ `(Just undefined)`.

goldfirere:
Today I learned that `Just { }` is a valid expression in Haskell. Oh, Haskell.

Ericson2314:
I added `Foo { a = 1, .. }` to the motivation.

Ericson2314:
Oh, I think I need to submit this?

simonpj:
I don't seen any downside in this proposal.  I'd be content to see it accepted

gridaphobe:
I see one oddity that would remain if we accept this proposal as is.

```haskell
-- 1. Accepted before
data Foo = Foo { x :: Int }
f (x@Foo{..}) = 42

-- 2. Accepted after
data Foo = Foo {}
f (x@Foo{..}) = 42

-- 3. Rejected before
data Foo = Foo Int
f (x@Foo{..}) = 42

-- 4. Accepted after
data Foo = Foo
f (x@Foo{..}) = 42
```

(3) is rejected for good reason, there are fields that must be bound positionally. However (4), which syntactically looks like a positional datatype like (3) is now accepted. 

I think even with this oddity the proposal is worthwhile, and I'm not in favor of rejecting (4). Let's just note it as a remaining oddity. Alas.

simonpj:
The proposal says (right at the start):

Currently this is allowed:
```
data Foo = Foo { a :: Int }

f (Foo {..}) = 1

x = Foo {..}
```
But that's not true.  The x binding is rejected with
```
Puns.hs:13:5: warning: [-Wmissing-fields]
    • Fields of ‘Foo’ not initialised: a
    • In the expression: Foo {..}
      In an equation for ‘x’: x = Foo {..}
```
Could you correct the proposal, pls?

simonpj:
I'm in support of this proposal, notwithstanding Eric's point above.

AntC2:
> The `x` binding is rejected with ...

Heh, heh this works, and makes the point:

```haskell
data Foo = Foo { a :: Int }

f (Foo {..}) = let x = Foo {..} in 1
```

This works:

```haskell
data Baz b = Baz { b :: b }

xb = Baz {b = b}                      -- but { .. } rejected, same reason
```

So ... (?) We can tell `{ .. }` from what it's supposed to be desugared to. (The trick doesn't work for `Foo`, because of a type mismatch, of course.)

simonpj:
[12 days ago ](https://github.com/ghc-proposals/ghc-proposals/pull/496#issuecomment-1153303273)I pointed out what may be a bug in the propsoal text and asked

> Could you correct the proposal, pls?

Would be possible to make this fix?  THanks

tomjaguarpaw:
What should "allowed" mean here? It seems the proposal takes it to mean "compiles with or without warnings". Should it mean "compilles without warnings"? (Conversely, what should "rejected" mean?)

tomjaguarpaw:
Perhaps the clearest thing to do would be to rewrite to say

----

Currently this is allowed for `n >= 1`:

```haskell

data Foo = Foo { a1 :: tn, ..., an :: tn }

a1 = ...
...
an = ...

f (Foo {..}) = 1

x = Foo {..}
```

But this is not:

```haskell
data Foo = Foo { }

f (Foo {..}) = 1

x = Foo {..}
```

gridaphobe:
@Ericson2314 Feedback from the committee has been uniformly positive, but there are two outstanding requests for clarification to the proposal text:

* from @simonpj https://github.com/ghc-proposals/ghc-proposals/pull/496#issuecomment-1153303273
* from myself https://github.com/ghc-proposals/ghc-proposals/pull/496#issuecomment-1153221981

If you could update the text to clarify these points, I think we'd be ready to accept and merge, thanks!

Ericson2314:
Sorry for missing the activity here. Hope the last revision addresses both of those!

gridaphobe:
Thanks @Ericson2314!

@nomeata let's merge!

nineonine:
Please note that the "rendered" link in PR description is broken

goldfirere:
Fixed.

PR #500, opened 2022-04-06
Add implicit import proposal

TristanCacqueray:
The goal of this proposal is to enable direct access to any available module without requiring an import statement. This removes an unnecessary restriction, making the language simpler.

[Rendered](https://github.com/TristanCacqueray/ghc-proposals/blob/implicit-import/proposals/0000-implicit-import.rst)

### DIFF
--- a/proposals/0000-implicit-import.rst
+++ b/proposals/0000-implicit-import.rst
@@ -1,0 +1,119 @@
+Allow implicit qualified module import
+======================================
+
+.. author:: Tristan de Cacqueray
+.. date-accepted:: Leave blank. This will be filled in when the proposal is accepted.
+.. ticket-url:: Leave blank. This will eventually be filled with the
+                ticket URL which will track the progress of the
+                implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/497>`_.
+.. contents::
+
+The goal of this proposal is to enable direct access to any available module without requiring an import statement.
+This removes an unnecessary restriction, making the language simpler.
+
+
+Motivation
+----------
+The main purpose is to be able to directly use a package.
+This is particularly useful for single use functions, such as ``Servant.serve``.
+For example, this propose change would make the following program valid
+(assuming a cabal file with build depends and default extensions):
+
+::
+
+ module Main where
+
+ import Servant.API (Get, JSON, (:>))
+
+ type MyApi = "answer" :> Get '[JSON] Integer
+
+ server :: Servant.Server MyApi
+ server = pure 42
+
+ main :: IO ()
+ main = Network.Wai.Handler.Warp.run 8080 app
+   where
+     app :: Network.Wai.Application
+     app = Servant.serve (Data.Proxy.Proxy @MyApi) server
+
+
+The status quo is not satisfactory because it requires adding explicit imports:
+
+::
+
+ -- Without this proposal, these imports are required
+ import Data.Proxy
+ import qualified Network.Wai
+ import qualified Network.Wai.Handler.Warp
+ import qualified Servant
+
+Managing such imports has the following negative effects:
+
+- It increases the file length.
+- The author needs to jump to the top of the file each time a new module is used.
+- It complicates code refactoring when moving a definition requires moving (or copying) its import.
+
+
+Proposed Change Specification
+-----------------------------
+I propose an extension to enable implicit qualified imports.
+
+When the ``ImplicitQualifiedImport`` language extension is enabled,
+unknown fully qualified symbols are gracefully resolved early in the compilation pipeline by
+injecting implicit ``import qualified`` statement when necessary.
+
+
+Examples
+--------
+Here are some example functions that are often used locally and
+it would be useful to call them without having to add an import statement:
+
+- Control.Concurrent.threadDelay
+- Data.Char.isAlpha
+- Data.Foldable.traverse_
+- Data.Maybe.mapMaybe
+- Data.Set.fromList
+- Data.Text.pack
+- System.Environment.getArgs
+- Text.Printf.printf
+
+
+Effect and Interactions
+-----------------------
+The proposed change enables using any module without requiring an import statement.
+
+Interactions with existing language or compiler features:
+
+- Hidden imports (using ``hidding``) must not be available implicitely.
+- Modules available through multiple package will be disambiguated using the PackageImports extension.
+- Only unknown fully qualified names will be affected.
+
+
+Costs and Drawbacks
+-------------------
+TBD: estimate development and maintenance costs.
+
+This extension may improve the language's learnability for novice users by:
+
+- Reducing the length of code samples, and,
+- Simplying new module usage, e.g. for one-off experiments and temporary annotations.
+
+The main drawback is that the extension may reduce the readability of a module:
+its external requirements would no longer be explicitely listed in the import section.
+
+
+Alternatives
+------------
+Perhaps this could be supported by default, without an extension, if this new
+behavior does not conflict with already valid code.
+
+Another helpful solution would be to enable local import, e.g. in a function definition.
+
+
+Unresolved Questions
+--------------------
+TBD


TristanCacqueray:
Hi, first time contributor here. I could not find a previous proposal for this feature, so I figure I should try to make one. Though I don't know how GHC resolves symbols and manages imports, thus if this is possible, I will need help to write the specifications.

JakobBruenker:
Note that ghci already supports this, and when you run `:set` in ghci it claims to have enabled `-fimplicit-import-qualified`.

However, trying to use this flag in a file doesn't actually seem to do anything - even though ghc doesn't complain about a missing flag.

One question is whether this should use an extension as in the proposal, or just use the existing flag, but make it function outside of ghci.

nomeata:
> One question is whether this should use an extension as in the proposal, or just use the existing flag, but make it function outside of ghci.

If it exists, it should be an extension: It changes how to interpret the source file, and other tools would have to understand it as well.

Ericson2314:
How do fully qualified names leading to modules within the current package work? Currently GHC's "downsweep" figuring out module dependencies can function without reading entire modules. Shall we change it to do a pass looking for fully qualified imports?

JakobBruenker:
> If it comes, it should be an extension

In that case I would suggest removing `-fimplicit-import-qualified` as part of this proposal, and instead turning on the extension by default in GHCi (similarly to how MonomorphismRestriction is turned off in GHCi - though weirdly, that doesn't seem to be mentioned by `:set`?)

proposals/0000-implicit-import.rst
```diff
@@ -0,0 +1,119 @@
...
+     app :: Network.Wai.Application
+     app = Servant.serve (Data.Proxy.Proxy @MyApi) server
+
+
+The status quo is not satisfactory because it requires adding explicit imports:
+
+::
+
+ -- Without this proposal, these imports are required
+ import Data.Proxy
```
konsumlamm, line 49:
```suggestion
 import qualified Data.Proxy
```
This should also be `qualified`, since the example uses `Data.Proxy.Proxy`, right?

proposals/0000-implicit-import.rst
```diff
@@ -0,0 +1,119 @@
...
+- It complicates code refactoring when moving a definition requires moving (or copying) its import.
+
+
+Proposed Change Specification
+-----------------------------
+I propose an extension to enable implicit qualified imports.
+
+When the ``ImplicitQualifiedImport`` language extension is enabled,
+unknown fully qualified symbols are gracefully resolved early in the compilation pipeline by
+injecting implicit ``import qualified`` statement when necessary.
```
konsumlamm, line 67:
```suggestion
injecting an implicit ``import qualified`` statement when necessary.
```

proposals/0000-implicit-import.rst
```diff
@@ -0,0 +1,119 @@
...
+it would be useful to call them without having to add an import statement:
+
+- Control.Concurrent.threadDelay
+- Data.Char.isAlpha
+- Data.Foldable.traverse_
+- Data.Maybe.mapMaybe
+- Data.Set.fromList
+- Data.Text.pack
+- System.Environment.getArgs
+- Text.Printf.printf
```
konsumlamm, line 82:
```suggestion
- ``Control.Concurrent.threadDelay``
- ``Data.Char.isAlpha``
- ``Data.Foldable.traverse_``
- ``Data.Maybe.mapMaybe``
- ``Data.Set.fromList``
- ``Data.Text.pack``
- ``System.Environment.getArgs``
- ``Text.Printf.printf``
```
This is code, so it should be in code blocks. This also prevents `Data.Foldable.traverse_` from being interpreted as a link.

konsumlamm, line 82: I rarely see `threadDelay`, `getArgs` and `printf` being used, so perhaps these aren't the best examples, although this may be subjective.

proposals/0000-implicit-import.rst
```diff
@@ -0,0 +1,119 @@
...
+
+
+Effect and Interactions
+-----------------------
+The proposed change enables using any module without requiring an import statement.
+
+Interactions with existing language or compiler features:
+
+- Hidden imports (using ``hidding``) must not be available implicitely.
```
konsumlamm, line 91:
```suggestion
- Hidden imports (using ``hiding``) must not be available implicitly.
```

konsumlamm, line 91: What exactly does this mean? That with `import Data.Maybe hiding (mapMaybe)`, using `Data.Maybe.mapMaybe` should not be valid?

proposals/0000-implicit-import.rst
```diff
@@ -0,0 +1,119 @@
...
+- Modules available through multiple package will be disambiguated using the PackageImports extension.
+- Only unknown fully qualified names will be affected.
+
+
+Costs and Drawbacks
+-------------------
+TBD: estimate development and maintenance costs.
+
+This extension may improve the language's learnability for novice users by:
```
konsumlamm, line 100: Given that beginners usually don't use any language extensions, I don't think this would really affect beginners (unless it is enabled by default, of course). That would change however if this extension became part of a future `GHC20XX` language.

proposals/0000-implicit-import.rst
TristanCacqueray: It's not necessary, but your suggestion looks better.

proposals/0000-implicit-import.rst
TristanCacqueray: Thanks!

proposals/0000-implicit-import.rst
TristanCacqueray: If an import statement explicitly hides a definition, then this extension should not enable access to the definition, like in your example. Or perhaps this should only be the case for qualified imports, e.g. `import qualified Data.Maybe hiding (mapMaybe)`, then `Data.Maybe.mapMaybe` should not be valid.

proposals/0000-implicit-import.rst
TristanCacqueray: I'd be happy to use a different list :) I've added `Debug.Trace.trace` in the last commit.

proposals/0000-implicit-import.rst
TristanCacqueray: I can't think of a situation where this extension would not be useful, thus I'm hopeful it could be added to a future `GHC20XX` language.

TristanCacqueray:
Thank you all for the prompt feedback. I've incorporated your comments in the last commit.

JakobBruenker:
>  Shall we change it to do a pass looking for fully qualified imports?

I think this is an argument in favor of not enabling such a feature by default, since it'd likely be better if such a pass only has to be done if implicitly imported names are actually used in a module

proposals/0000-implicit-import.rst
```diff
@@ -0,0 +1,124 @@
...
+When the ``ImplicitQualifiedImport`` language extension is enabled,
+unknown fully qualified symbols are gracefully resolved early in the compilation pipeline by
+injecting an implicit ``import qualified`` statement when necessary.
+The pipeline might need an extra pass to collect missing imports.
+
+GHCi already implements this feature thanks to the flag ``-fimplicit-import-qualified``,
+but the flag does not seem to work with GHC.
+Moreover the flag is not documented in ``ghc --show-options``, but it is accepted.
+Thus this proposal deprecates the ``-fimplicit-import-qualified`` flag in favor of the
+``-XImplicitQualifiedImport`` modifier, so that the behavior is consistent between GHC and GHCi.
```
konsumlamm, line 74:
```suggestion
GHCi already implements this feature via the ``-fimplicit-import-qualified`` flag,
but it does not seem to work with GHC.
Moreover the flag is not documented in ``ghc --show-options``, but it is accepted.
Therefore this proposal deprecates the ``-fimplicit-import-qualified`` flag in favor of
``-XImplicitQualifiedImport``, so that the behavior is consistent between GHC and GHCi.
```

proposals/0000-implicit-import.rst
```diff
@@ -0,0 +1,124 @@
...
+- ``Text.Printf.printf``
+
+
+Effect and Interactions
+-----------------------
+The proposed change enables using any module without requiring an import statement.
+
+Interactions with existing language or compiler features:
+
+- Fully qualified import with hidden declarations are not affected: with ``import Data.Maybe hiding (mapMaybe)``, using ``Data.Maybe.mapMaybe`` should not be valid.
```
treeowl, line 99: What about explicitly limited imports? If I write

```haskell
import Control.Monad (liftM2)
```

will `Control.Monad.ap` be in scope?

How about renamed imports? If I write

```haskell
import Control.Monad as M hiding (ap)
```

I would expect that `M.ap` would not be valid but that `Control.Monad.ap` would be. Do you concur?

proposals/0000-implicit-import.rst
```diff
@@ -0,0 +1,124 @@
...
+-------------------
+TBD: estimate development and maintenance costs.
+
+This extension may improve the language's learnability for novice users by:
+
+- Reducing the length of code samples, and,
+- Simplying new module usage, e.g. for one-off experiments and temporary annotations.
+
+The main drawback is that the extension may reduce the readability of a module:
+its external requirements would no longer be explicitely listed in the import section.
```
treeowl, line 114:
```suggestion
its external requirements would no longer be explicitly listed in the import section.
```

treeowl: I think this is actually a pretty big deal, especially when trying to refactor code to change the module dependency order.

treeowl: Another potential downside: if a module uses this feature, it's harder to swap out its dependencies for ones with similar interfaces but different implementations. This may not be a big deal in context.

phadej:
How this would compare with https://github.com/ghc-proposals/ghc-proposals/pull/283 if it is undormanted.

I'm a fan of project specific fat preludes, and would e.g. re-export `Data.Text` as `T`. So what this proposal suggest that I could use `Data.Text.pack` everywhere doesn't add value, as I'd use `T.pack` everywhere.

Or using #283, my fat prelude could just re-export whole `Data.Text`.

I see #283 as superior. It gives more control, and still keeps imports contained in the header of the module.

proposals/0000-implicit-import.rst
TristanCacqueray: Thank you very much for the corrections!

proposals/0000-implicit-import.rst
TristanCacqueray: I agree, thank you for the suggestions, I'll add them to the next follow-up.

TristanCacqueray:
@phadej I guess the main difference with #283 is that this proposal doesn't need a curated list of module: If the fat prelude doesn't export, say `Servant`, then that does not address the issue raised in the motivation.

TristanCacqueray:
It seems like we can leverage the existing code used by GHCi. I took a stab at implementing this in GHC, and the change is rather minimal: https://gitlab.haskell.org/TristanCacqueray/ghc/-/tree/ImplicitQualifiedImport

guibou:
What about instances?

`import` statement also imports `instance`s and there is no control on this right now. An implicitely introduced `import qualified` will also implicitly import instances and ones may argue that it may be too much implicit. This should be discussed in the interaction section.

proposals/0000-implicit-import.rst
```diff
@@ -0,0 +1,124 @@
...
+ import qualified Data.Proxy
+ import qualified Network.Wai
+ import qualified Network.Wai.Handler.Warp
+ import qualified Servant
+
+Managing such imports has the following negative effects:
+
+- It increases the file length.
+- The author needs to jump to the top of the file each time a new module is used.
+- It complicates code refactoring when moving a definition requires moving (or copying) its import.
```
guibou, line 58: I'm sure you can add another argument: because unqualified is the default, user have the habit to introduce an unqualified import, which leads to import conflicts (now or in the future). Qualified locally is better (in my opinion) than unqualified globally, which is the default behavior.

TristanCacqueray: That is also my opinion, thank you for the suggestion. It's added in the last commit.

TristanCacqueray:
> What about instances?

If I understand correctly, the implementation proposed in the last commit should not affect instances. Only the names that throws ``Not in scope`` errors are handled differently.

Thank you for the feedback, I think all the suggestion are now part of the proposal. Please let me know if I missed something.

JakobBruenker:
> the implementation proposed in the last commit

GHCi seems to import instances from used modules:

```haskell
ghci> (++"!") <$> Data.Map.fromList [(1,"foo")]
fromList [(1,"foo!")]
```

`instance Functor (Map k)` is defined in `Data.Map.Internal`.

TristanCacqueray:
If I understand correctly, non orphaned instances are available whenever the typeclass or the type is in scope. 

While trying this out, I realize that the proposed implementation does not work for local module, where the implicit dependencies need to be declared in the ModuleGraph. Thus in the last commit I added a mention that the `GHC.Driver.Make.downsweep` process needs to be modified so that it parse the whole module.

TristanCacqueray:
In the last commit of my [ghc ImplicitQualifiedImport branch](https://gitlab.haskell.org/TristanCacqueray/ghc/-/tree/ImplicitQualifiedImport) I updated the downsweep pass to use the Lexer in order to add all the qualified names to the module graph. That seems to work as a proof of concept, though I can't tell what are the implications :)

nomeata:
I was wondering if there is a problem with Template Haskell, and the staging it introduces into the processing of a file. But it seems there is none: lexing still happens all at once, so that's good. And spliced code wouldn't have to be considered by this change (in a way, it already behaves as this), so also fine. Good!

I wonder if it is incompatible, at least morally, with local modules as envisioned by https://github.com/ghc-proposals/ghc-proposals/pull/283: with that, it's no longer clear how to map from what looks like a qualified name to the module one should import (it could be any prefix).

Overall it's an intriguing and interesting proposal, and my mind making up process is still ongoing. If I hadn't done some Ocaml programming recently I might have said “surely no”, but I'm not so sure anymore.

nomeata:
> Or perhaps I miss understood the issue with the prefixes?

Beyond the point that you mentioned, I see two possible issues:

 * You see `A.B.C.foo` in the code (with nothing imported explicitly). With Local Modules around, how would you know the name of the “real” module to import? I.e. which of these to emulate:
   ```
   import qualified A (module B (module C(foo)))
   import qualified A.B (module C(foo))
   import qualified A.B.C (foo)
   ```

 * Let's say you have a fat module that exports some (local) modules you common use (e.g. `Map`). In most of your code, you `import FatModule` (unqualified!), and then use, for example, `Map.bar`.
 
   Now in one module you forget to `import FatModule`, but still write `Map.bar`. How can you extension guess that it should now implicit `import FatModule`? Or – reasonably – it should no do that, but the error message “Could not find `Map.hs`” that comes from the implicit `import qualified Map (foo)` from this extension is also more confusing than helpful maybe.

TristanCacqueray:
Oh I see, thank you for providing the details. Well, assuming the implementation of local modules updates the existing downsweep process when looping imports to populate the cache with all the exported modules, here is a plausible solution:

- For `A.B.C.foo`, the `getImplicitImports` looks for `A.B.C`, `A.B` exporting `C`, and `A` exporting `B.C`. If more than one match is found then the compilation fails with `Multiple qualified names found for A.B.C, disambiguate by adding import X or Y`.

- If no match can be found, the extension does nothing and the renamer fails with a `Not in scope: 'Map.bar'` error. Then I guess we can also look for all the exported modules and perhaps suggest that the user can `import FatModule` to bring `Map` into scope. I agree that the extension should not automatically guess that the user meant to get `Map` from `FatModule`, that should definitely be an error.

nomeata:
Ah, so this only works with modules already loaded? If I write `main = Foo.bar` in `Test.hs` and run `ghc Test`, will it look for a file `Foo.hs`?

TristanCacqueray:
In the [PoC](https://gitlab.haskell.org/TristanCacqueray/ghc/-/commit/9b5208d7e532eefe45b54bde8f6d15160601f55e#1dab250036d04cfcf3530f6ff27889f723cc2dda_1631_1654) I'm using `findImportedModule` for qualified name tokens, and that seems to look for a file named after the qualified name: 

```
$ cat Test.hs
main = Foo.bar
$ cat Foo.hs
module Foo where
bar = putStrLn "go/"
$ ghc Test.hs
[1 of 3] Compiling Foo              ( Foo.hs, Foo.o )
[2 of 3] Compiling Main             ( Test.hs, Test.o )
[3 of 3] Linking Test [Objects changed]
```

If the file does not exist, the compilation fails with:

```
[1 of 2] Compiling Main             ( Test.hs, Test.o )

Test.hs:1:8: error:
    Not in scope: ‘Foo.bar’
    NB: no module named ‘Foo’ is imported.
  |
1 | main = Foo.bar
  |        ^^
```

nomeata:
Ok, and so if it's `Foo.Bar.baz` and we have local modules, it will look at both `Foo.hs` and `Foo/Bar.hs`? (But maybe it's too much a distraction to talk about what would happen with another extension that isn't even actively designed right now.)

TristanCacqueray:
That would be one solution, and I am fairly confident both proposal can live in harmony.

Moreover I think `ImplicitQualifiedImport` could be used in place of the `QualifiedReexports` suggested by @chshersh in https://github.com/ghc-proposals/ghc-proposals/pull/283#issuecomment-541525245 : the relude package may define the `BL`, `BS` and `Set` modules as standalone exposed-modules to achieve a similar result. 

Lastly, the [proof of concept](https://gitlab.haskell.org/TristanCacqueray/ghc/-/tree/ImplicitQualifiedImport) shows that this proposal can be implemented with minimal changes to the compiler. After all, the desired logic already exists for GHCi.

nomeata:
Right, the ease of implementation is good to know, but ultimately not the main deciding factor. I am curiously looking forward to the committee discussion, and am a bit surprised that we don't hear more options here. Has this proposal already hit the discourses and reddits?

simonpj:
Would it be possible to frame the "Proposed change specification" as a change to the language specification, making no reference to the implementation?  There should be no mention of "downsweep phase" or lexer or `ITq*` tokens. Put yourself in the position of a Haskell programmer who hasn't the foggiest idea how GHC is implemented.

Something more like this.

When `ImplicitQualifiedImport` is on
* A qualified name `M.N.x` is looked up in the top level environment (see [Import Declarations](https://www.haskell.org/onlinereport/haskell2010/haskellch5.html#x11-1010005.3) in the Hakell report.
* If that lookup fails, instead of reporting an out-of-scope error, behave as if an extra import declaration is added:
  ```
  import qualified M.N(x)
  ```
  If there is a module `M.N`, and it exports and entity `x`, the lookup succeeds.

Some examples would be useful.  For example:
```
module A.B( g ) where
  g = True

module C.D( f ) where
  f = "hello"

module M where
  import A.B as C.D
  foo = (C.D.f, C.D.g)
```
I think you intend that
* `C.D.g` binds to the `g` exported by `A.B` (as usual)
* `C.D.f` isn't in scope by the usual rules, so we try adding an extra import `import qualified C.D(f)`. That works, and binds to the `f` exported by `C.D`.
* Moreover, even if `C.D` exported `g` as well, the reference to `C.D.g` would continue to bind to the `g` from `A.B`.  Correct?

This makes it pretty hard to figure out the dependencies.  The occurence `C.D.f` in `foo`'s rhs induces a dependency on module `C.D`, but the very similar looking occurrence `C.D.g` does not.  That is going to make it hard to work out dependencies from a simple scan (as now).

It would be good to explain all this.

TristanCacqueray:
Oh I see, thank you @simonpj for pointing that out, I appreciate your help. I think it is confusing if the extension could mix different modules under the same name implicitly. Perhaps the example should fails because `A.B` does not export `f`?

Thus would it makes sense to change the second part of the specification to:

- If that lookup fails, and if there is no module `M.N` already in scope, then instead of reporting an out-of-scope error, behave as if an extra import declaration is added.

tomjaguarpaw:
This feels to me like one those ideas that is either completely revolutionary or will end up completely forgotten about after six months.  On the one hand, I can't imagine ever wanting to use this in practice. On the other hand, maybe after experiencing it I'll wonder how I ever lived without it!  So, I encourage the author to go ahead and try.

That said, I find the justification a bit weak:

> This removes an unnecessary restriction, making the language simpler.

For me, restrictions that allow you to easily know exactly what your module depends on are similar to restrictions that allow you to easily know if your function does IO. I don't buy that removing unnecessary restrictions is always better.

> Managing such imports has the following negative effects:
>
> * It increases the file length.

I'm quite doubtful that that's a negative effect. On what basis? Bytes on disk? Network transfer? Extra lines visible in the editor?

> * The author needs to jump to the top of the file each time a new module is added or removed.

Again, quite doubtful that this is a negative effect.

> * It encourages using unqualified imports because unqualified is the default, which can lead to import conflicts.

That's quite a tenuous chain of reasoning!

_That_ said, it seems like this is sufficiently interesting an idea to be worth a try (not that I have any say in the matter)! I hope my objections will be proved completely wrong and Haskell discovers the wonderful new[1] world of import-free programming.

[1] I don't think any other language does this, does it?

nomeata:
> [1] I don't think any other language does this, does it?

Ocaml does, and when I recently had to write some serious Ocaml, I didn't mind. There simply isn't a concept of importing; all available modules are in scope qualified, and you can `open` them if you want them unqualified. That experience makes my somewhat optimistic about this proposal. It very likely makes sense. Whether its worth the _change_ and the conceptual complexity of having both, I'm not sure.

tomjaguarpaw:
> >  [1] I don't think any other language does this, does it?
>
> Ocaml does

Good to know, thanks!

JakobBruenker:
Java also has this feature, in fact it's the only way to resolve name clashes in Java, as far as I'm aware.

chshersh:
I like this feature and I'd love to use it because it makes the Haskell development experience more hassle-free. And I think this feature would become especially handy and superior when more packages adopt [these module naming conventions](https://www.haskellforall.com/2021/05/module-organization-guidelines-for.html) when a package `foo` exports all public API under the single top-level `Foo` module.

Ideally, I want to use `Lens.view`, `ByteString.singleton` and `Maybe.mapMaybe` instead of `Control.Lens.view`, `Data.ByteString.singleton` and `Data.Maybe.mapMaybe` so qualified reexports still would be handier. But, well, Haskell already has some historical baggage and as noted, this can be workarounded on the user side.

It's nice observation that `relude` can implement kinda qualified reexports with this feature even though it looks rather ugly to have such "fake" modules in alternative standard library 😅 But implicit imports already work nice with my other packages like `co-log` or `validation-selective`. I can just type `Colog.log` and I already know that this comes from the `co-log` package. Writing `import qualified Colog` in the beginning of the file doesn't bring me anything, it's just extra waste of time.

I also always wanted the local imports feature because often I use an imported name only in a single place in the middle of a module even though I need to import it in the beginning of a file. But the implicit import proposal makes things even better because I don't even need to import stuff at all if I want to use it locally only in a single place.

Moreover, I think the proposed implicit import proposal can be good for IDE-based workflow when you type something like `ByteString.` and see all the functions from the module instead of typing smth like `singleton` and letting the IDE guess from which package (text, containers, bytestring, etc.) this name comes (if you want on-the-fly import addition in your IDE).

TristanCacqueray:
@simonpj I have updated the specification to address your comment: after testing different implementations, I think it is best to avoid adding implicit imports for existing qualified **or** renamed modules. To that effect, I have added another example and updated the [proof of concept](https://gitlab.haskell.org/TristanCacqueray/ghc/-/tree/ImplicitQualifiedImport) branch.

Please let me know if that makes sense.

simonpj:
Thanks. It's an improvement.  Some further suggestions under "Proposed change specification"

> "GHCi already implements this feature via..."

It seems a shame that the *specification* of a feature should refer to the feature before defining it.   I suggest you move this para to the end of the Proposed Change Spec section, and state the change up-front. Something like:

    Change 2: deprecate `-fimplicit-import-qualified`.   <explanation follows>

Incidentally, the explanation appears to be "the flag doesn't work and is not documented" to which a response might be "well, make it work, and document it".  But I think your motivation is deeper: it doesn't do what you want.   Perhpas says so?

----------------

Also in "Proposed change spec" you say "If that lookup fails, and if there is no matching qualified module or renamed module already defined, ..."  But what, precisely, is a "matching qualified module or renamed module"?  Can you make the language precise?  Something like:
```
If
* The lookup fails, and
* There is no user-written import declaration `import M.N ... ` or `imports X ... as M.N`
then instead of reporting an out-of-scope error, behave as if an extra import declaration is added: `import qualified M.N(x)`.
```
Then you might want to add a sentence or two to explain the second bullet -- and/or refer forward to a (explicitly-named) example that illustrates the reasoning.

simonpj:
> Ambiguous imports are forbidden

This is a good example, but you don't explain it.  "Ambiguious imports are forbidden is *way* too cryptic.  And the second bullet assumed different rules than those you give in Proposed Change Spec.  Indeed the second bullet (which might be better re-ordered as the third is the *reason* for you side condition under Proposed Change Spec.   Don't assume the reader is clever; jus lead us carefully by the hand

simonpj:
> Data.Maybe.fromJust is implicitely imported because Data.Maybe isn't already defined as a qualified module, or as a renamed module.

But Data.Maybe absolutely *is* imported explicitly!

TristanCacqueray:
> Incidentally, the explanation appears to be "the flag doesn't work and is not documented" to which a response might be "well, make it work, and document it". But I think your motivation is deeper: it doesn't do what you want. Perhpas says so?

You are correct, the motivation is deeper. I've reworded that part to clarify it.


> But what, precisely, is a "matching qualified module or renamed module"? Can you make the language precise?

Your suggestion looks much better, and I've added a few sentences to explain the reasoning.

In the last commit I have extended the examples with more explanations.
I meant to follow the principle of least surprise.
But perhaps this is not the case, and I would be happy to use a different specification.
That being said, I think the last version describes what this extension should do.


> Indeed the second bullet (which might be better re-ordered as the third is the reason for you side condition under Proposed Change Spec.
> But Data.Maybe absolutely is imported explicitly!

I stand by those examples: as long as the import is not qualified, adding an extra (implicit) qualified import should not cause any harm.
I hope this is clearer in the last commit.


Thank you so much for helping me out, studying how GHC works and observing the ghc-proposals process has been fascinating!

nomeata:
> Thank you so much for helping me out, studying how GHC works and observing the ghc-proposals process has been fascinating!

That is very good to hear! Processes, committee etc can easily become frustrating, so “fascinating” is quite remarkable :-)

goldfirere:
I will admit to an internal curmudgeonly reaction when I saw the title: "Oh, this is a bad idea, and in any case, it will never work." But after reading the proposal -- and reflecting on how this is nice in other languages -- I'm cautiously in support. At least in support of thinking about this seriously.

I have a few worries:
- Currently, GHC can do dependency analysis after parsing only a relatively simple prefix of source files. In particular, I believe this prefix can even be accurately parsed only with regular expressions. That is, there are no recursive constructs other than lists that appear before the end of the `import` section. With this change, that property would no longer hold. Does this matter? I'm not sure that it does, but my guess is that dependency analysis would slow down considerably if modules used this new feature. (Currently, in GHC, it already takes maybe 5 seconds or so to do the dependency analysis, which is a non-trivial wait when staring at a prompt.)
- I am worried about the interaction with #283. But actually Java has the same problems, as I demonstrate below. Java seems not to have suffered, so it's perhaps not worth worrying about this design corner.
- I am also worried about instances. Adding a new `import` imports new instances. With these new implicit imports, source files may import instances implicitly. In practice, I don't see this as being a big problem, as having extra instances usually does not cause trouble. Nevertheless, it might be nice to have a `-Wimplicit-orphan-imports` when orphan instances are implicitly imported, which could indeed surprise users.

For what it's worth, I would use this feature all the time. It would make me want #283 more... possibly to the point of omitting `import`s at the top and instead just importing symbols per function or per group of related functions.

-----------------------------------

I did this little experiment in Java:

```java
// file ./level1/level2.java
package level1;
public class level2
{
  public static final int x = 43;
}
```

```java
// file ./level1.java
public class level1
{
  public class level2
  {
    public static final int x = 42;
  }
}
```

```java
// file ./Main.java
public class Main
{
  public static void main(String[] args)
  {
    System.out.println(level1.level2.x);
  }
}
```

Compiling and running `Main` yields 42. If I delete or rename `level1.java` out of the way (and remove all `.class` files), then compiling and running `Main` yields 43. If I now restore `level1.java` but change `x` to `y`, I get an out-of-scope error, because `leve1.leve2` contains no `x`. If I change the name of the `level2` class to be something else, I get an out-of-scope error, because `leve1` contains no `level2`. Just to check, getting rid of `leve1` altogether allows `Main` to compile again. This reveal's Java's algorithm: when it sees `a.b.c`, first look for `a.class` (compiling it from `a.java` if necessary). Then, look for `b.c` within `a.class`, failing if `b.c` is not there. If `a.class` does not exist, then look for `a/b.class` (compiling it from `a/b.java` if necessary), and look for `c` in it. We could easily do the same if this proposal and #283 ever co-exist.

The problem is not as bad in Java as it would be for Haskell, because Java has a convention where packages begin with a lower-case letter while classes begin with an upper-case letter. This convention makes it easy for a reader of `a.b.C.D.e` to know what file `e` is really declared in. (I'd look in `a/b/C.java`, where I expect a local class `D` declaring `e`.) Haskell, of course, uses capital letters throughout, and so we lose this small signal. This does not worry me too much though -- I'm sure we could survive.

Ericson2314:
> Does this matter? I'm not sure that it does, but my guess is that dependency analysis would slow down considerably if modules used this new feature. (Currently, in GHC, it already takes maybe 5 seconds or so to do the dependency analysis, which is a non-trivial wait when staring at a prompt.)

Yes, it will slow down a lot. Error messages in the presence of cycles may be surprising, too.

Ericson2314:
> I will admit to an internal curmudgeon-only reaction when I saw the title

I am proudly still that curmudgeon. "I-want-to-type-less-driven design" is extremely suspicious, and it is prudent to have a knee-jerk reaction against it.

As the old joke goes, Java is verbose because IDEs have the pain. Well guess what, we have IDEs now too!

The real problem with Java was not that IDEs enabled more typing, but that the good stuff is mixed up between the boilerplate --- it is a *reading* disaster not writing disaster. And IDEs cannot abrogate the need to read and understand existing code.

Imports do not suffer from this problem. They are syntactically highly distinguished and separated, and so no one need bother reading them. Conversely, as I said in the previous comment, by removing them, we degrade the development experience in the presence of syntax errors, because now the imports are not formally decidable, and so IDE heuristics must be integrated into "downsweep" to continue to resolve names.

The *right* solution here is to abandon programming using text entirely. I encourage everyone to read https://github.com/unisonweb/unison. The world of text we are stuck with in comparison to that is a crime against programming.

--------

Absent going full Unison, I will offer a counter-proposal so I am not just a negative curmudgeon. I will do this in steps.

1. What if instead of getting rid of imports, we stored them in a separate `Mod.imports` file next to the regular `Mod.hs`file. Note how this somewhat resolves a "lock file" from a package manager, in that it is somewhat amendable to automatic updates.

2. When using an IDE, the IDE can do whatever heuristics it wants to keep the import graph up to date with edits. It should even be able to use the GHC API to pass in a in-memory version of the import graph without writing to disk, bypassing the `Mod.imports`. This is good for very low-latency operations (as is avoiding the disk in general).

3. Only when ~~"saving"~~ \[what a lame concept! What even is as that \[floppy disk] icon!] committing (Ah there we go, flushing an in-memory "journal" of edits to a version control normal form (with human annotation) is a legit concept), need the IDE update the `Mod.imports` files, persisting the latest import graph.

4. Simple batch compilation (e.g. of dependencies, or release builds) can use the "fast to read" `Mod.import` persisted format to avoid any regressions.

What is is doing is just separating the juicy human code from the mostly boilerplate `import`s, in a way that I believe satisfies @TristanCacqueray's of getting the boilerplate out of site / out of mind, without sacrificing other desirable tooling goals:

 - Batch compilation remains fast and simple
 - There is a *standardized semantics* for name resolution in the presence of syntax errors, because the notional `Mod.import`s are separate *parsing units*. This is even better than today where downsweep's "imports only" stuff is an unspecified hack.
 - The standalone `Mod.imports` format gets us closer to reviving https://github.com/ghc-proposals/ghc-proposals/pull/245

nomeata:
Thanks for the counter-proposal? But how about having your cake and eat it too, and make it a proposel on top fo @TristanCacqueray: Everything that you say, but if there is no `Mod.import` file (or the API client does not provide an in-memory one), `ghc` will parse the full file and look for explicit imports and (in the presence of this extension) implicit imports.

This way, IDE users live in the nice world you describe, but I can still use `ghc` directly without needing an IDE and get the same experience, and especially in one-off scripts this feature will be very useful.

TristanCacqueray:
@goldfirere Thank you for the feedback and for your cautious support.

- About slow down, please see my reply to @Ericson2314 below.

- About the interaction with `Local modules`, perhaps we can prevent the surprising issue by checking if there is only one path that can (implicitly) resolve a definition. For example, with `Level1.Level2.x`, if both `import Level1 (Level2.x)` and `import Level1.Level2(x)` match, then the import shall be explicitly defined, otherwise it is ambiguous.

- What if we change the specification from `behave as if an extra import declaration is added` to `look up *all* the names available in the environment`. That seems to be the behavior of `GHC.Rename.Env.lookupQualifiedNameGHCi`, and it does not seem like that would interfere with instances. I think it would be ideal if `ImplicitQualifiedImport` does not bring instances into scope implicitly: the user should add explicit imports in that case. It seems like this is the default behavior in Rust, where we need to add `use itertools::Itertools;` to get the extra Iterator instance provided by the `itertools` library.


@Ericson2314 Thank you for the feedback and for taking the time to write a counter-proposal.

> Yes, might slow it down.

Here is what I have in mind with regard to the implementation:

- Perform the lexer pass when building the module summaries. Later pass should be able to re-use this work, thus this change should not incur any slow down.
- In the downsweep, traverse the list of token to collect all the unique ITq* name and add them to the dependency list (as if they were regular import). It seems like the result of this pass should be cached as usual: it should not happen when a module is not changed.

I am hopeful this can be implemented efficiently to keep the overhead minimal, and I am looking forward how to benchmark this...


> we degrade the development experience in the presence of syntax errors, because now the imports are not formally decidable,

I am not sure to understand this: if we use the lexer for the implicit import resolution, how would syntax errors degrade the development experience?


> IDE heuristics must be integrated into "downsweep" to continue to resolve names.

Presently with HLS, when we type `Data.Set.`, the auto completion suggests *all* the names available. At least this would have to be updated so that HLS only suggests the names available in the `Data.Set` module, but that is not really related to this proposal. If I understand correctly, HLS operates like GHCi, and it already knows about all the available names.

> As the old joke goes, Java is verbose because IDEs have the pain. Well guess what, we have IDEs now too!

I think IDEs still require the user to accept imports addition or removal, which is helpful but not ideal.


> The right solution here is to abandon programming using text entirely.

I totally agree, but I think we need a new breed of code editor for that...
One idea I really like from Unison is how each expression is packaged individually, and I wish it would be practical to do the same in Haskell.


> Absent going full Unison, I will offer a counter-proposal so I am not just a negative curmudgeon.

That sounds like a good solution, and I'd be willing to work on this if the `ImplicitQualifiedImport` overhead becomes too big. Though I hope such optimization could be implemented as a follow-up.

TristanCacqueray:
> > Lexing

> Interesting idea, but not sure it will work in the presence of TH, where [| ... Foo.bar ... |] does not introduce a dependency.
> Did you have this in mind as a heuristic, or the definitional rule?

I haven't thought about that, though it doesn't sound complicated to handle such rule: if qualified tokens within a TH does not introduce a dependency, then we can skip the tokens that are between `ITopenExpQuote [|` and `ITcloseQuote |]`.


> we have to start fixing the tech debt in GHC

I hear you, and if this extension can negatively affect the renovation work, then I think we should put it in dormant state. I could not find past discussions about this feature and I'm happy this proposal now exists for future references.

Though if the extension can be added without using force-fit, and if it makes sense to have it sooner than latter,
then I would be thrilled to complete this work because I think such extension is worth the effort.

doyougnu:
I am not in support of this proposal. 

The proposal states:

> The goal of this proposal is to enable direct access to any available module without requiring an import statement. This removes an unnecessary restriction, making the language simpler.

I disagree that this makes the language simpler. Rather, it adds another language pragma (albeit a hopefully simply one) which increases the language complexity. More importantly though, this increases the amount of magic^{TM} and `spooky action at a distance` available in the language, which IMHO makes the language less declarative and harder to understand. In vanilla Haskell+GHC all Module imports have a clear and obvious location, if I need to know what a module is importing I know exactly where and how to find it. With this proposal if I need to know what a module imports I now have to read through the entire module because these imports are directly referenced and sprinkled throughout the module code. Perhaps I'm too much of a Writh fan with his variables declared before the procedures that use them design :)

> Managing such imports has the following negative effects:
> 
>   - It increases the file length.
>   - The author needs to jump to the top of the file each time a new module is added or removed.
>   - It encourages using unqualified imports because unqualified is the default, which can lead to import conflicts.

Taken in order:
  - Yes, this is indeed the case, but surely this is low in the ordering of negative effects. Is there a quantifiable negative effect from long file lengths? Does this increase compile times or something else that is high value? Does this slow down HLS or create long parsing times? 
  - Yes! But I usually don't do this anyway since I can apply code actions from HLS which alters the imports based on an error or warning. This seems like a high cost solution for a problem which can be solved by storing a mark in emacs/vim, jumping to imports, editing the imports and returning. But more importantly, with this proposal if I want to change an implicitly qualified import I now need to hunt to find the implicit import which IMO is much worse than jumping to a pre-defined location to edit. I suspect this proposal exacerbates this negative case rather than solving it.
  - Perhaps, but I think this is more of a matter of code and project style than a hard limitation.

Besides those points, I'm uncomfortable with the `other languages do it so it is good` argument that has been somewhat floated in the comments. In particular, the original motivation for implicit imports in OCaml was [as a solution](http://www.lpw25.net/papers/ml2014.pdf) for ad-hoc polymorphism. But Haskell+GHC has world-class support for ad-hoc polymorphism. IMO OCaml's solution is quite good because it reuses machinery that is well-supported and very powerful in the language, i.e., its module system and specifically functors. This same reasoning doesn't make sense in Haskell though.  If we are to argue for this proposal via a similar feature in another language as part of the motivation, then I think we should understand the original motivation for the feature and we should understand if the feature is considered an anti-feature by that languages community at the time of the proposal.

TristanCacqueray:
@doyougnu Thank you very much for taking the time to write these objections.

> In vanilla Haskell+GHC all Module imports have a clear and obvious location, if I need to know what a module is importing I know exactly where and how to find it.

Though the module names may not indicate which packages are used (unless they are imported with PackageImports). What if the proposal includes a command to print the direct dependency of a given module along with their package names, wouldn't that help?

> Is there a quantifiable negative effect from long file lengths?

I don't know how to quantify this, but I think shorter files are easier to read. The import section is also prone to merge conflict.

> I can apply code actions from HLS which alters the imports based on an error or warning.

I argue that HLS (or custom editor configuration) is also a high cost solution for managing qualified imports.

> if I want to change an implicitly qualified import I now need to hunt to find the implicit import.

I am not sure to understand this, if you want to remove the usage of a qualified import, you can add `import qualified A ()` to prevent its usage. Or you can add `import qualified B as A` to use another module.
In such situation, why not using HLS (or the replace function of your editor) to perform the refactor?

> Besides those points, I'm uncomfortable with the `other languages do it so it is good` argument that has been somewhat floated in the comments.

I agree and this was not part of the initial proposal. Though I think this extension can help users coming from these other languages. I got used to be able to invoke fully qualified name anywhere in the code, and coming back to Haskell, I find it a hassle to deal with the import section.

> In particular, the original motivation for implicit imports in OCaml.

Thank you for the link, I was not able to find it.

TristanCacqueray:
Please find a summary of the discussion so far. Please let me know if I missed something.

I believe the current poposal text accounts for the following points:

- Interaction with GHCi -fimplicit-import-qualified.
- Module dependencies analysis in the "downsweep" phase.
- Interaction with user-written imports.

The following points are not yet part of the proposal:

- Interactions with the `Local modules` proposal #283.
- Interactions with Template Haskell.
- Impact on the tech debt in GHC.
- Typeclass instances. The proposal indicates that they should not be introduced implicitly, but I think the specification needs to be changed from `behave as if an extra import declaration is added` to `look up *all* the names available in the environment`.
- https://github.com/ghc-proposals/ghc-proposals/pull/500#issuecomment-1099096625 and https://github.com/ghc-proposals/ghc-proposals/pull/500#issuecomment-1103348186 provide objections about the benefits and motivations of the proposal.
- Counter proposal `Mod.imports`.

goldfirere:
I would urge us not to worry about technical debt in this way. The proposal process is all about design. (We *can* talk about debt in the design.) This is *not* about implementation. It is possible that this proposal, if accepted, would have an unacceptable implementation burden. But let's focus on whether we want this in the language at all before worrying too much about the concrete implementation.

simonmar:
Perhaps a compromise here would be to allow implicit imports but only for modules outside the current package. The advantage of that would be that implicit imports wouldn't affect the dependency analysis and the implementation is much simpler. The (admittedly major) disadvantage is that specifying the semantics requires talking about packages, which otherwise are not part of the Haskell standard, so it feels wrong in that sense. There's also the disadvantage that using this extension liberally would make it harder to move the package boundaries, making refactoring harder. I guess I'm probably not in favour of this compromise, but I thought I'd mention it anyway.

santiweight:
What if you import this new namespace:

```
import *
```

In other words, you import the global namespace. This means that now the entire namespace contains all modules (it's a bit like a macro that qualified imports all possible modules).

Consequences:
 - No new pragma
 - The imports can still be pre-parsed for programs that don't use this namespace. GHC can be smart enough to parse the whole file if you do use this namespace

Maybe you could extend this syntax to support @simonmar's suggestion:

```
import "my-package" *
```

TristanCacqueray:
@santiweight If I understand correctly, such new import syntax would still need a pragma. And the advantage of this proposal being just a pragma is that it can be enabled globally by adding it to the list of default-extensions.

I am also not in favor of @simonmar's suggestion: even though that would work most of the time, I think the proposal would provide a better user experience if it also works with local modules.

TristanCacqueray:
@Ericson2314 After reading through [2022-ghc-modularity](https://hsyl20.fr/home/files/papers/2022-ghc-modularity.pdf) I think I better understand what you meant and I totally agree this work would make our life easier. Thank you so much for that!

I believe I have addressed the remaining salient points in the last commit, and unless there is an objection, I think this is ready for the committee.

nomeata:
Thanks, submission received and assigned to @Icelandjack. @Icelandjack, looks like you didn’t accept the repo invitation yet, so I couldn’t assign it to you here on Github. But you are on it :-)

simonpj:
> This demonstrates how the implicit dependency can be discovered using the Lexer

I would love to see this sentence (currently buried in Costs and Drawbacks) made much more explicit and promient. Explain what is the "implicit dependency" and why it matters.

I don't have a strong opinion about the utility (from a programmer's POV) of this proposal.  The design is simple enough (good) but the costs are not trivial: 
* More code paths -- a bit more complexity.  Nothing major, but more.
* Dependency computation needs to read every token in the file.

So I'm not very enthusiastic, but I'd be swayed by lots of people saying "this would make my life better"

TristanCacqueray:
@simonpj please find an update to the Costs and Drawbacks section in the last commit. I have removed the mention of the lexer because I think it is an implementation detail.

I believe the cost can be kept small by:
- updating the `getImports` code to use the lexer instead of the parser.
- sharing the tokenization work done by the make driver with the parsing layer.

Perhaps it is not optimal to do these changes now, and I would be happy to check on the upcoming modularization work and see how to accommodate this extension efficiently.

aspiwack:
@doyougnu 

> In particular, the original motivation for implicit imports in OCaml was [as a solution](http://www.lpw25.net/papers/ml2014.pdf) for ad-hoc polymorphism. But Haskell+GHC has world-class support for ad-hoc polymorphism. IMO OCaml's solution is quite good because it reuses machinery that is well-supported and very powerful in the language, i.e., its module system and specifically functors.

This is a misunderstanding. Ocaml has always had the ability to refer to other modules without an import statement. In fact, there is no import statement in Ocaml. The modular implicits that you are referring to are not about imports, but about discharging module argument automatically in the manner of type classes. This could happen with or without explicit import statements.

doyougnu:
> This is a misunderstanding. Ocaml has always had the ability to refer to other modules without an import statement. In fact, there is no import statement in Ocaml. The modular implicits that you are referring to are not about imports, but about discharging module argument automatically in the manner of type classes. This could happen with or without explicit import statements.

Aha! Thank you for the correction @aspiwack and taking the time to respond. I much prefer having correct information over being right!

nomeata:
Gentle ping about shepherding this proposal, @Icelandjack :-)

goldfirere:
This has been pending a recommendation from @Icelandjack for about 2.5 months -- @Icelandjack, do you think you can get to this recommendation? I think we may need to reassign if not. Thanks!

nomeata:
I’ve reassigned to Simon Marlow.

TristanCacqueray:
I'm still interested in pushing this forward as I think it is a valuable addition to the language. It seems like the most valuable use-case is to make the extension work with modules that are defined outside of the current package (e.g. `Debug.Trace`). Thus, It might be best to implement the proposal in two stages:

1. Introducing the extension and making it work by subsuming the `-fimplicit-import-qualified` GHCi flag.
2. Making the extension work for modules defined in the current package by changing the dependency "downsweep" process.

The first stage seems relatively straightforward to implement, and the only downside is that the user will get a `Not in scope` error when trying to use the feature with a local module. Therefore, If such implementation is feasible, I'd be happy to update the proposal with this two-stage plan.

simonmar:
Committee shepherd here. I've read through all the comments (phew).

One thing that would inform our discussion is to know the impact on compile time (and space) of doing a full lexical scan of every module during the downsweep. Would it be possible to get some numbers on that for something large, e.g. GHC itself?

TristanCacqueray:
@simonmar Thank you for following up on this proposal. That's a good idea, I'll try to get those performance numbers using the initial implementation.

TristanCacqueray:
Thanks to @mpickering suggestions, I think I managed to measure the cost of the lexical scan in this [make-lexical-analysis branch](https://gitlab.haskell.org/TristanCacqueray/ghc/-/tree/make-lexical-analysis) by calling `lexTokenStream` to count qualified names right after pre-processing the imports. Though I'm not familiar with the code base, and it would be great if someone could validate the test implementation.

I measured a full compiler build as well as a single rebuild of the GHC module:

```ShellSession
# baseline clean GHC build
2,892,047,868,432 bytes allocated in the heap 
Total   time  3400.765s  (3490.170s elapsed)

# lexical clean GHC build
2,914,811,164,216 bytes allocated in the heap 
Total   time  3469.633s  (3558.010s elapsed)

# baseline GHC module rebuild
    9,707,615,104 bytes allocated in the heap
Total   time    10.789s  ( 13.420s elapsed)

# lexical GHC module rebuild
   32,644,612,384 bytes allocated in the heap
Total   time   24.613s  ( 27.360s elapsed)
```

I noticed the rebuild does re-preprocessed all the dependencies, eventhough they didn't changed, thus perhaps this could be improved by caching the scan result?

simonpj:
I have reflected on this proposal and, while I see its merits, I lean against it.  It's a matter of taste but the cost/benefit ratio comes out too high for me -- more complexity [both design and implementation] without increasing expressiveness, and what is, to me, a small increase in convenience.  And stylistically I positively like listing dependencies!

But I accept that others may differ.

I'm a bit shocked that some compile time metric might *triple* (which seems to be what the above figues show).  That might be a show-stopper anyway; or it might be a measurement glitch, or overcome with cleverer implementation.    But design first!

TristanCacqueray:
It seems like the above figures show the cost of doing an extra lexical scan. Perhaps we could mitigate the overhead by sharing the lexical scan done in the downsweep process with the main pipeline? That may be a good improvement, for example, to replace the current getImport parser.

Unless we can reduce the cost/benefit ratio, then I'm perfectly happy to retract this proposal.
In any case, thank you all for taking the time to evaluate the proposal.

simonmar:
> sharing the lexical scan done in the downsweep process with the main pipeline

That's not going to save very much in the case when just one module needs to be recompiled, because we're only saving the lexical analysis of that one module.

To make this extension efficient, we would have to cache (on disk) the precomputed dependencies of each source file. That's a lot of extra complexity, and I think probably tips the balance into "not worth it" territory.

TristanCacqueray:
> To make this extension efficient, we would have to cache (on disk) the precomputed dependencies of each source file.

That sounds like a valuable feature on its own, for example, as suggested by @Ericson2314 [`Mod.import` counter proposal](https://github.com/ghc-proposals/ghc-proposals/pull/500#issuecomment-1101972865). I guess we should mark this extension as "dormant" until such feature is implemented.

nomeata:
Marking this as dormant as per author's request. Thanks for your contribution!

PR #507, opened 2022-05-16
Allow field puns to bind type variables

googleson78:
[Rendered]()

Currently, writing `{field :: a}` to bind the type `a` in a pattern is not allowed. 

This proposal introduces an extension to the existing `ScopedTypeVariables` and `NamedFieldPuns` extensions, that enables that behaviour.

I've marked this as a draft, since I'm not sure about a few things
* should I extend the proposal to account for the record construction case as well?
* is the grammar spec good enough?
* do we need more examples - this doesn't really introduce anything new, only syntax sugar, so it should be straightforward to understand?
* are there any corner case - for example some parse conflicts or such?
* as with the examples, I'm not sure what to write for effects and interactions, other than the fact that this steals the `{field :: a}` syntax


proposals/0000-scoped-type-variable-puns.md
```diff
@@ -0,0 +1,137 @@
...
+Proposed addition:
+```
+fbind
+        : qvar '=' texp
+        | qvar '::' ctype
+        | qvar
+        ...
+```
+
+TODO: I'm not sure if the Haskell report contains a specification for the grammar of field puns and scoped type variables.
```
goldfirere, line 68: No -- these features are not formalized, as far as I know. Feel free to delete this TODO.

proposals/0000-scoped-type-variable-puns.md
```diff
@@ -0,0 +1,137 @@
...
+If anyone wants to point me to where those are specified, I'd be happy to update this section.
+
+TODO: I'm not sure what corner cases might exist for this, since it is only a syntax sugar change.
```
goldfirere, line 71: I'm good at corner cases, but I'm struggling here too. Haskell does allow `data Rec = MkRec { a,b,c :: Int }`, and presumably you would not allow `f (MkRec { a,b,c :: x }) = ...`, (or, rather, you would, but the type signature would connect only to the final field, `c`), but that's not too cornerish.

proposals/0000-scoped-type-variable-puns.md
```diff
@@ -0,0 +1,137 @@
...
+Following this, the only things I can imagine are parser issues.
+
+The specification can, and almost always should, be illustrated with
```
goldfirere, line 74: Please delete this para.

proposals/0000-scoped-type-variable-puns.md
```diff
@@ -0,0 +1,137 @@
...
+`NamedFieldPuns` and `ScopedTypeVariables` are both firmly cemented as useful and ubiquitous language extensions.
+As we move towards improved type level programming, we should aim to make the UX as seamless as possible.
+I think that allowing the `MkField {field :: x}` syntax would be considered "obvious" and "natural" to anyone who has used both extensions,
+to the point where when I tried it I had no doubt it would work.
+
+Apart from the obvious convenience that this would provide, it is also less error-prone(as is usual with using field names instead of positional matches) compared to binding type variables in patterns, as described [here](https://gitlab.haskell.org/ghc/ghc/-/issues/18830#note_308891).
+
+## Proposed Change Specification
+
+I propose that a pattern match of the form
```
goldfirere, line 42: I like the idea of extending this to expressions, too. It keeps the language more symmetric between expressions and patterns.

proposals/0000-scoped-type-variable-puns.md
```diff
@@ -0,0 +1,137 @@
...
+
+-- x is able to be referred to in the body here, in the same way in which it would be if we wrote
+-- MkThing {field = field :: x}
+f :: Thing a -> ()
+f MkThing {field :: x} = ()
+```
+
+## Effect and Interactions
+
+TODO: not sure what to write here.
```
goldfirere, line 93: I agree there isn't much to write here.

goldfirere (commented):
I'm not a fan of `NamedFieldPuns`, but given that they already exist, this appears to be a user-friendly extension. Thanks for writing this up!

proposals/0000-scoped-type-variable-puns.md
AntC2: Not just "not formalised": neither field puns nor scoped type variables/PatternSignatures/pattern type annotations are allowed in the Report's syntax. So the formalising is in the GHC extensions documentation.

proposals/0000-scoped-type-variable-puns.md
AntC2: Presumably this'll interact with `-XExplicitForall` and `-XScopedTypeVariables` in that in

```haskell
    f :: forall a. Thing a -> a
    f MkThing{field :: a} = undefined :: a
```

the `a` appearing in the field annotation is the `a` `forall`'d in `f`'s signature.

(And that variety of scoping is contra one of the Principles for future GHC, [point 4. here](https://github.com/goldfirere/ghc-proposals/blob/type-variables/principles.rst#224contiguous-scoping-principle-csp).)

googleson78: This is however the same interaction as what is currently present if you write `MkThing {field = field :: a}`, so I'm not sure if I should mark it down as an interaction of this proposal.

googleson78:
Thanks for the feedback Richard! I've addressed your points and extended the proposal to cover the expression case as well, and am removing the draft status for this proposal.

AntC2:
> **Alternatives** ... the burden(annoyance) of effectively disabling `NamedFieldPuns` when you want to bind a type variable in a pattern match.

(You can use `NamedFieldPuns` and bind a type variable currently -- but it's clunky:

```haskell
f (MkThing{field} :: Thing x) = undefined :: x
```

and only if the type you want to bind is a parameter to the datatype.)

proposals/0000-scoped-type-variable-puns.md
googleson78: I tried to extend it to cover expressions as well.

proposals/0000-scoped-type-variable-puns.md
googleson78: Done

proposals/0000-scoped-type-variable-puns.md
googleson78: Thanks, I noted that down.

proposals/0000-scoped-type-variable-puns.md
googleson78: Done

aspiwack:
> I'm not a fan of `NamedFieldPuns`, but given that they already exist, this appears to be a user-friendly extension. Thanks for writing this up!

For the record, I'm a big fan of `NamedFieldPuns`.

googleson78:
Since there is not much discussion happening here, I think it's time I brought this up for review, @nomeata.

Ericson2314:
Also not a fan of `NamedFieldPuns` :) but yes consistency matter. I would rather this proposal be dealt with after https://github.com/ghc-proposals/ghc-proposals/pull/448 or we will just have to make that one even bigger to specify their interaction.

nomeata:
The proposal says “Allow field puns to bind type variables”, but really, it is “Allow field puns have type annotations”, and what happens with variables in these annotations depends on other factors (other extensions etc.), right?

I think @Ericson2314 has a point. For example, after #448, maybe we’d say that this feature in patterns requires `PatternSignatures` to be on. @googleson78, would that be ok for you, or do you think that’s an unreasonable delay?

I am somewhat distraught by 

> ```
> MkRec {x, y, z :: a}
> ```
> the `::` a signature binds only to the z, and not to all of the fields.

in all other places `x, y, z :: a` means the same as `x :: a, y ::a, z :: a`, _including_ in record definitions! This seems inconsistent.

But what else can we do?
```
MkRec {x, y, (z :: a)}
```
is odd because parentheses are not allowed there.


I am not fully convinced that having to write `foo = foo :: a` here is so bad. If you really need to bind a type variable, it seems to be a complex enough use case that the simple “`foo` stands for `foo = foo`” shortcut is simply not available. And with the type lambdas in #448 the need for this feature will be reduced.

Ericson2314:
> _including_ in record definitions!

Wow! I had no idea
```haskell
data Foo = Foo { a :: Int, x, y, z :: Int }
```
is valid!

It would seem to be the consistent syntax is thus
```haskell
foo Foo { a :: _, x, y, z :: Int }
```
the `:: _` stops the `:: Int` from reaching back any further.

googleson78:
@nomeata
I do not mind at all waiting, I definitely don't want to delay that proposal further.

> in all other places

Are there any places other than in a record definition? I'm not coming up with anything right now.

Regarding "we should have it be consistent with data definitions" - I think these are two semantically very different contexts, and I'm not sure has been any strive to make them mean syntactically and semantically consistent. For example, we can't write `data A = A {x, y, z}`, nor can we write `data A = A {x = foo}` in a record definition (I admit I can imagine a world where the second one means "default value for this field").

The only thing that comes to mind is pattern synonyms, where you do use the record construction syntax to define a new match. However, in there, adding a `::` doesn't currently parse (I assume exactly because it uses the parser for record matches/expressions).

If there are no other similarities between the syntax for a definition and matching/usage, then I don't see why we should strive to introduce one with this change.

The stronger reason why I'm opposed to unifying them is that it would make this significantly less useful - in the cases where you want to bind a type variable (and even in the ones you don't) you rarely have several fields of the same type, hence most of the time (including what originally prompted me to make this proposal) it would mean either field reordering or adding additional syntactic noise along the parens or the `:: _` that @Ericson2314 suggests. The `:: _`, btw, I would expect to mean "type hole", and not "ignore this annotation", hence would require turning on `PartialTypeSignatures`, which hurts the entire module.

> it seems to be a complex enough use case that the simple “`foo` stands for `foo = foo`” shortcut is simply not available

I don't understand "why not". The only reason coming to my mind is "our compiler is getting too complex". I understand scope creep and that you can't just add everything anyone wants, but this is not a brand new feature requiring tons of integration and something that will affect a big part of ghc - it is very localised to parsing+renaming, and in particular, specifically the record matching/construction parts of those. On the other hand, "death by a thousand cuts" seems like a more likely scenario, but again, to me personally, this seems more like missing functionality/interaction, rather than adding something new.

nomeata:
> Are there any places other than in a record definition? I'm not coming up with anything right now.

Anywhere where type signature are allowed!
```
module Example where
odd, even :: Int -> Bool
odd = ...
even = ...
```

googleson78:
Ah, I see what you mean now, I was thinking of only "record related" usages of `::`. Although, trying to use the same syntax in a pattern match doesn't parse hence decision on how it should behave isn't taken there (although I assume that will never be valid, because we would need type-directed parsing to be able to unambiguously parse it.) I see in this context how the `Rec {x, y, (z :: Int)}` syntax could make sense.

Ericson2314:
Re pattern synonyms, they currently always use named fields puns which isn't really correct. That should be fixed (been meaning to file something about it) but until then let's take any precedent with a grain of salt.

googleson78:
Now that #448 was accepted and merged, are there any suggestions on what I should be changing here? It seems like the proposal here should work with `PatternSignatures` in the "doesn't bind type variables" sort of way. Additionally, I guess allowing "type application patterns" for field puns might also make sense, but I'm not sure how good an idea including that here is.

Ericson2314:
I think you should start over and focus on just one thing: record gadt syntax for `forall` in data constructors, and `Foo { @a = Int, x = 1 }` to use it.

It's not nearly as exciting, but I think that's proper least-controversial next step to extend the language.

Then we can return to this, which seems like that + anonymous records.

PR #511, opened 2022-05-31
Deep Subsumption (under review)

mpickering:
The [simplified subsumption proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0287-simplify-subsumption.rst) argued in favour of simplifying the subsumption judgement in GHC's type system, by removing:

* Covariance and contravariance of function types
* Deep skolemisation
* Deep instantiation.

The change was motivated by wanting to simplify both the implementation and language semantics, as well as being a stepping-stone to the implementation of Quick-look impredicativity.

Unfortunately, the breakage study failed to accurately predict how annoying this change would be to users. Some common patterns found in libraries now require eta-expansion, when the compiler used to automatically insert the appropriate lambdas.

This proposal suggests adding a new language extension, `DeepSubsumption`, enabled by default in the common case when the language is set to `-XHaskell2010`, which recovers the previous subsumption rules. This would allow users to opt-in to deep subsumption as it was before GHC-9.0.

[Rendered](https://github.com/mpickering/ghc-proposals/blob/deep-subsumption/proposals/0000-deep-subsumption.rst)

### DIFF
--- a/proposals/0000-deep-subsumption.rst
+++ b/proposals/0000-deep-subsumption.rst
@@ -40,8 +40,10 @@ Proposed Change Specification
 -----------------------------
 
 We propose to add a language extension ``DeepSubsumption`` which provides a best
-effort attempt to restore the previous deep subsumption behaviour. The extension
-will implement the 4 subsumption rules which were removed by simplified subsumption.
+effort attempt to restore the previous deep subsumption behaviour.
+The extension will implement deep skolemisation and the co/contra subtyping
+rules, which were removed by simplified subsumption. It will not re-introduce
+deep instantiation.
 
 * The extension will be **on** by default, as-if it was added to the Haskell2010 extension set.
 * The extension will be **off** in the GHC2021 extension set.
--- a/proposals/0000-deep-subsumption.rst
+++ b/proposals/0000-deep-subsumption.rst
@@ -1,0 +1,224 @@
+Deep Subsumption
+==============
+
+.. author:: Matthew Pickering, Simon Peyton Jones, Jaro Reinders
+.. date-accepted:: Leave blank. This will be filled in when the proposal is accepted.
+.. ticket-url:: Leave blank. This will eventually be filled with the
+                ticket URL which will track the progress of the
+                implementation of the feature.
+.. implemented:: Leave blank. This will be filled in with the first GHC version which
+                 implements the described feature.
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. contents::
+
+``DeepSubsumption`` is a languge extension which allows users to opt-into deep
+subsumption as it existed before GHC-9.0.
+
+
+Motivation
+----------
+
+The `simplified subsumption proposal <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0287-simplify-subsumption.rst>`_
+argued to simplify GHC's type system by removing 4 cases from the subsumption judgement.
+The change was motivated by wanting to simplify both the implementation and language
+semantics.
+
+Unfortunately the breakage study failed to accurately predict how annoying this
+change would be to users. Some common patterns found it libraries now require
+"pointless" eta-expansion where the compiler used to automatically insert the
+appropiate casts due to the 4 removed subsumption rules.
+
+This proposal is about adding ``DeepSubsumption`` language extentsion which can
+be enabled to add back the removed subsumption rules to allow users to opt-in to
+deep subsumption as it was before GHC-9.0.
+
+
+Proposed Change Specification
+-----------------------------
+
+We propose to add a language extension ``DeepSubsumption`` which provides a best
+effort attempt to implement the restore the deep subsumption behaviour. The extension
+will implement the 4 subsumption rules which were removed by simplified subsumption.
+The extension is off by default and must be enabled in client code if they want
+the subsumption rules and associated eta-expansion.
+
+It is not recommended that people use this extension as it makes type inference
+less predictable and the language semantics more confusing. However, in a manner
+similar to ``NoMonoLocalBinds``, if the user really wants this then they are free to
+enable the extension with the understanding that they take responsibility if things
+go wrong!
+
+Example 1
+---------
+
+The example given by ParetoOptimalDev on `Discourse <https://discourse.haskell.org/t/r-haskell-was-simplified-subsumption-worth-it-for-industry-haskell/4486>`_ was analysed carefully by
+Jaro R.
+
+Certain libraries such as `pipes <https://hackage.haskell.org/package/pipes>`_ define a general data type but also
+export specialised type synonyms with universally quantified type variables. It
+is key to use a type synonym rather than a newtype so that the the specialised
+versions can still work with more general combinators.
+
+For example, pipes defines the following data types::
+
+  data Proxy x' x a b m r = ....
+
+  type Producer' b m r = forall x' x . Proxy x' x () b m r
+
+and also provides the ``fromHandle`` function, which uses the ``Producer'`` type synonym::
+
+  fromHandle :: MonadIO m => Handle -> Producer' ByteString m ()
+
+using the ``fromHandle`` function can lead to compilation failures with simplified
+subsumption::
+
+  readFreqSumFile file = readFreqSumProd $ withFile file ReadMode fromHandle
+
+
+Solution 1: Eta-expansion
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+As described in the simplfied subsumption proposal, the simplest fix is to eta-expand
+the call to ``fromHandle`` in the definition of ``readFreqSumFile``::
+
+  readFreqSumFile file = readFreqSumProd $ withFile file ReadMode (\x -> fromHandle x)
+
+However, ParetoOptimalDev isn't so satisfied by this solution because
+
+1. It required many such "pointless" changes to the code base.
+2. It seems "random" when you need to eta-expand or not, Haskell programmers expect
+   eta-equivalence to hold (even though it does not and never has).
+3. They view the benefits (simpler language, simpler semantics) as something that
+   this breaking change is not worth it. We have lived with deep subsumption for
+   many years.
+
+This led Jaro to explore some other alteratives.
+
+Solution 2: Newtype Wrapper
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Simon PJ suggests making these type synonyms into newtypes::
+
+  newtype Producer' b m r = MkProducer' (forall x' x. Proxy x' x () b m r)
+
+If you implement all the required constraints for this type then you can just write the original::
+
+  readFreqSumFile file = readFreqSumProd $ withFile file ReadMode PB.fromHandle
+
+But this is not quite a good solution here, because you can't
+automatically derive all the instances and you cannot compose these producers
+with other pipes.
+
+This interoperability could probably be restored by using the same tricks that
+the ``optics`` library uses to get their lenses to compose, but that seems like
+quite a big change here.
+
+Solution 3: Rewrite type synonyms in place
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Another simple change to resolve this is to just rewrite the type synonyms in-place::
+
+  fromHandle :: MonadIO m => Handle -> Proxy x' x () ByteString m ()
+
+Then you can also just write the original non-eta-expanded version. But a
+problem is that it is harder understand that the ``Proxy`` in this case really
+must be a producer. It is also another invasive change to rewrite all the type
+signatures of all downstream libraries which use this pattern.
+
+Solution 4: Deep Subsumption
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+With this proposal, the user enables ``DeepSubsumption`` and the program continues
+to typecheck as before::
+
+  {-# LANGUAGE DeepSubsumption #-}
+
+  ...
+
+  readFreqSumFile file = readFreqSumProd $ withFile file ReadMode PB.fromHandle
+
+
+Example 2
+---------
+
+Another consumer hit hard by the change is the `Integrated Haskell Platform<https://github.com/digitallyinduced/ihp/pull/1342>`_.
+
+In particular they define a type synonym which contains an implicit parameter::
+
+  type Html = (?context :: ControllerContext) => Html5.Html
+
+which is used to create the ``renderUser`` combinator::
+
+  renderUser :: User -> Html
+  renderUser user = [hsx|<li>{get #name user}</li>|]
+
+but now ``renderUser`` fails to typecheck in ``renderUsers`` without eta-expansion::
+
+  renderUsers :: [User] -> Html
+  renderUsers users = [hsx|
+    <ul>
+      {forEach users renderUser}
+    </ul>
+  |]
+
+the "solution" is to eta-expand the call to ``renderUser``::
+
+  renderUsers :: [User] -> Html
+  renderUsers users = [hsx|
+    <ul>
+      {forEach users (\x -> renderUser x)}
+    </ul>
+  |]
+
+But Marc Scholten is `not happy<https://github.com/digitallyinduced/ihp/pull/1342#issuecomment-1058870639>`_ with this proposal as
+
+  All of them break existing IHP apps / require a lot of changes when updating.
+
+So it seems the arguments for simplified subsumption are not persuasive enough for
+him to accept the breakage. This points to a solution where we can have the best of
+both worlds by adding a ``DeepSubsumption`` extension which can be enabled by
+clients such as IHP if they desire this behaviour.
+
+
+
+Effect and Interactions
+-----------------------
+
+* The ``DeepSubsumption`` language pragma has all the drawbacks identified in
+  the simplified subsumption proposal, but crucially allows users to opt-in to
+  the drawbacks if their value judgement is different to that of the steering committee.
+
+
+Costs and Drawbacks
+-------------------
+
+* We really do not recommend that people use this feature. It makes the language
+  more complicated and type inference less predictable.
+* Alejandro Serrano `suggests <https://github.com/ghc-proposals/ghc-proposals/pull/287#issuecomment-1128134798>`_ that reintroducing this feature will not alleviate any pain because
+  by the time it's introduced then everyone will have already felt it as all
+  libraries will be upgraded to the 9.0 series.
+
+
+
+Alternatives
+------------
+
+* The alternative is to do nothing, users must accept that simplified subsumption
+  is here to stay and update their code appropiately.
+
+Unresolved Questions
+--------------------
+
+* We need to decide whether we would want to backport this feature to the 9.2 branch.
+
+
+Implementation Plan
+-------------------
+
+* A draft patch has been prepared by Simon PJ. `!8210 <https://gitlab.haskell.org/ghc/ghc/-/merge_requests/8210>`_.
+
+Endorsements
+-------------
--- a/proposals/0000-deep-subsumption.rst
+++ b/proposals/0000-deep-subsumption.rst
@@ -34,7 +34,7 @@ semantics, as well as being a stepping-stone to the implementation of Quick-look
 Unfortunately, the breakage study failed to accurately predict how annoying this
 change would be to users. Some common patterns found in libraries now require
 eta-expansion, when the compiler used to automatically insert the
-appropiate lambdas.
+appropriate lambdas.
 
 This proposal suggests adding a new language extension, ``DeepSubsumption``,
 enabled by default in the common case when the language is set to ``-XHaskell2010``,
--- a/proposals/0000-deep-subsumption.rst
+++ b/proposals/0000-deep-subsumption.rst
@@ -18,11 +18,15 @@ Deep Subsumption
 subsumption as it existed before GHC-9.0.
 
 
-Motivation
+1. Motivation
 ----------
 
 The `simplified subsumption proposal <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0287-simplify-subsumption.rst>`_
-argued in favour of simplifying GHC's type system by removing 4 cases from the subsumption judgement.
+argued in favour of simplifying the subsumption judgement in GHC's type system, by removing:
+* Covariance and contravariance of function types
+* Deep skolemisation
+* Deep instantiation.
+
 The change was motivated by wanting to simplify both the implementation and language
 semantics, as well as being a stepping-stone to the implementation of Quick-look impredicativity.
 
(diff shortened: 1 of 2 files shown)


serras:
As I stated in the “Simplified Subsumption” thread, I really wonder whether adding this back makes sense, once the ecosystem has already moved to 9.0. Do we expect libraries to support “ghc < 9 && >= 9.6”, to give an example?

parsonsmatt:
I'd love this. Would immediately enable it and start using it again as soon as I could.

phadej:
There is no link to rendered version in the description.

gbaz:
> Also these codebases are usually compatible with single GHC version only, so they very well could prefer `DeepSubsumption` over `ImpredicativeTypes`, if there were a choice.

As I motivated in that proposal, I'd much prefer a feature that gives the functionality users want (type aliases able to abstract well over "floating" constraints) on purpose, rather than one that only gave it sort of accidentally, most of the time, due to some very quirky decisions.

Bodigrim:
I support the proposal, foremost because it is important to send a signal to the community that GHC cares and takes responsibility for the breakage caused, once it was recognized that the impact assessment was incomplete.

Ericson2314:
Exactly agree with @Bodigrim. Even if we deprecate `-XDeepSubsumption` and remove it soon, do what @gbaz says instead, etc. etc., it is right and proper to do this now now for community reasons.

I am squarely team "Haskell needs a lot of breaking changes", and I don't like how deep subsumption today messed up the operational semantics, but in order to make the amount of breaking changes I'd like to see happen politically feasible, we *must* always have nice and proper deprecation cycles.

simonpj:
> There were other reasons for simplified subsumption, like semantic correctness, but as a language implementer I do appreciate the value of reduced implementation complexity and I feel that it would be a shame to reintroduce that complexity given that the cost of simplified subsumption seems to have been paid down by the community at this point.

I'm happy to say that the implementation changes are 
* **modest** (a couple of hundred lines) 

and, crucially, 
* **highly localised**, so the tax is paid in one place, rather than making lots of things more complicated.

You can see for yourself.  Here's the patch: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/8210.  (It will need a substantial Note to explain it.)

The reasons to be cautious about `DeepSubsumption` (and why I think that Simplified Subumption was a good change, for all its pain) are laid out in [Section 2.1](https://github.com/mpickering/ghc-proposals/blob/deep-subsumption/proposals/0000-deep-subsumption.rst#21-deepsubsumption-is-not-recommended)

goldfirere:
This proposal has rather little cost for those that don't use it, but it has potentially huge gain for those that do -- especially if simplified subsumption broke their API. So I'm personally going to listen more to people who say "yes, thank you, I'll use this" than those that say "is this really the best design? I won't use it". I'd also pay close attention to "I really want to use this, but the design doesn't fit my use case. Could we do X instead?"

Is it strange to have deep subsumption in a bunch of releases, then missing from a few, and then back? Yes, absolutely.

Is deep subsumption just plain wrong to begin with? I tend to think so.

But will this proposal make some people happy (and make their code better in their opinion)? I think so, and happiness is important. This is why I favor doing this.

simonpj:
> It seems that where simplified subsumption broke stuff was gnarly bits inside packages

Yes, if a package uses gnarly things inside, but exports an API that doesn't have foralls to the right of arrows, the clients of that package are entirely unaffected by the package's choice of whether or not to use `DeepSubsumption`.

But if the package exports an API with forall to the right of an arrow, their *clients* must either use `DeepSubsumption` or be prepared to do judicious eta-expansion.  Example 1 (from Section 3 of the proposal)
```
readFreqSumFile file = readFreqSumProd $ withFile file ReadMode fromHandle
```
This is in *client* code.   Without `DeepSubsumption` they will have to write
```
readFreqSumFile file = readFreqSumProd $ withFile file ReadMode (\x -> fromHandle x)
```

jberryman:
> ecosystem has moved on

Curious has this actually been quantified?

Appreciate the proposal

parsonsmatt:
> Curious has this actually been quantified?

It would be very hard to do so - the entire reason why it was hard to do impact assessment is because we can still *define* these APIs for clients/end-users, it's the *use-site* that is broken. `type SqlReadT m a = forall backend. SqlBackendCanRead backend => ReaderT backend m a` compiles just fine with GHC 9.0, but any application that uses it will be broken.

Any company that has upgraded to GHC 9 has dealt with the migration. But I haven't deleted that type alias from the library, in part because we still support GHC 8.6, and in part because I've been hoping we'd be able to use it again.

cdsmith:
I appreciate the thoughtfulness that has gone into this proposal.

About section 2.1: are there easy solutions to the semantics problems that were identified with regard to deep subsumption, rather than merely stating them?  In one thread I found on ghc-devs, @simonpj said with regard to the strictness change, "BTW I think I did think about [eta expanding `f` to `seq f (\x -> f x)` instead of `\x -> f x` to preserve semantics] at one point, but discarded it as imposing real runtime overheads, and inhibiting optimisation."  If the semantics change is all that concerning to some, is it worth thinking about this again?  Maybe not, but it seems worth asking the question.

remiturk:
> Exactly agree with @Bodigrim. Even if we deprecate `-XDeepSubsumption` and remove it soon after, do what @gbaz says instead of this long term, etc. etc., it is right and proper to still do this now for community reasons.
> 
> I am squarely team "Haskell needs a lot of breaking changes", and I don't like how deep subsumption before and with this proposal messes up the operational semantics, but in order to make the amount of breaking changes I'd like to see happen politically feasible, we _must_ always have nice and proper deprecation cycles.

My I-don't-really-have-any-stake-in-this €0.02:
It would be Most Awesome(TM) if `-XDeepSubsumption` could be introduced and backported. Then I'd love to see some implementation of the explicitly floaty forall idea, assuming that the current pain is indeed mostly about type sysonyms. And finally `-XDeepSubsumption` could be deprecated and then removed.

I remember it took me quite a while to understand how `type T args = forall ...` works, because I had internalized the intuition that they could always be (textually/syntactically) inlined. Which actually makes me wonder, after 20 years of using Haskell, is `type T args = forall ...` the same as `type T args = (forall ...)` right now? Without a GHC at hand to check, I think they are the same, and that the weirdness is really the `:: ... -> (forall a. ...)` being equivalent/rewritten to `:: forall a. ... -> ...`?

AntC2:
> [Example 1. quoting @ParetoOptimalDev] Haskell programmers expect eta-equivalence to hold (even though it does not and never has).

Quite, re the "never has". That's the sort of thing that should get documented/spec'd -- even in the Language Report? Newbs who've learnt about eta-whatever/pointfree/combinators  need to be let down gently.

And yet my first Google hit for 'Haskell eta-equivalence' gets [this (very old) wiki page](https://wiki.haskell.org/Eta_conversion#:~:text=Haskell%20theoretical%20foundations&text=Converting%20from%20the%20first%20to,the%20process%20in%20either%20direction.) saying " the following two values are equivalent under η-conversion ... Pointfree programming ...", without actually saying not to rely on the equivalence in Haskell. The linked page discusses some 'Problems with Pointfree' but not lack of equivalence. And there's plenty further tutorials re Pointfree that assume equivalence.

The difficulty, again, is for innocent importers of gnarly packages. So thank you for introducing extra warnings. Should the docos for `ImpredicativeTypes` and/or `RankNTypes` also give warnings?

runeksvendsen:
> Quite, re the "never has". That's the sort of thing that should get documented/spec'd -- even in the Language Report? Newbs who've learnt about eta-whatever/pointfree/combinators need to be let down gently.
> 
> And yet my first Google hit for 'Haskell eta-equivalence' gets [this (very old) wiki page](https://wiki.haskell.org/Eta_conversion#:~:text=Haskell%20theoretical%20foundations&text=Converting%20from%20the%20first%20to,the%20process%20in%20either%20direction.) saying " the following two values are equivalent under η-conversion ... Pointfree programming ...", without actually saying not to rely on the equivalence in Haskell. The linked page discusses some 'Problems with Pointfree' but not lack of equivalence. And there's plenty further tutorials re Pointfree that assume equivalence.

I don't think there's any problem in saying "*the expressions `f` and `\x -> f x` are equivalent*" because this statement is true if `f` is a function (ie. not "bottom") and we understand "equivalent" to mean "*logically* equivalent" (ie. ignoring runtime characteristics). Similarly, I don't think there's anything wrong with saying "Newton's laws of motion hold" even though we know they don't as velocity approaches the speed of light.  
Cf. [Fast and Loose Reasoning is Morally Correct](https://www.cs.ox.ac.uk/jeremy.gibbons/publications/fast+loose.pdf).

ulysses4ever:
> [@simonpj](https://github.com/simonpj) said with regard to the strictness change, "BTW I think I did think about [eta expanding f to seq f (\x -> f x) instead of \x -> f x to preserve semantics] at one point, but discarded it as imposing real runtime overheads, and inhibiting optimisation." If the semantics change is all that concerning to some, is it worth thinking about this again?

Good point. Maybe the alternative translation could be hidden by yet another extension (`-XStrixtEta` or some such) implied by `-XDeepSubsumption` and can be disabled if desired (e.g. by `-XNoStroctEta`).

Ericson2314:
We discussed this more in https://github.com/ghc-proposals/ghc-proposals/pull/287 as of late. See https://github.com/ghc-proposals/ghc-proposals/pull/287#issuecomment-1125419649 for example. The problem seems to be trying to treat implicit dictionaries and implicit type arguments the same. Implicit dictionaries makes this hopeless, but I don't think type arguments form a problem because evaluation occurs under type lambdas (as made possible and necessitated by erasure).

Still, that is off topic for this one. I think we should first add back `DeepSubsumption` regardless of what we have to replace it, and then *separately* proposals alternatives to both it and today's simple subsumption. The point of adding back `DeepSubsumption` is not that we know no other option, but to retroactively do the deprecation cycle we should have all along, sending a strong message that research and industry can cooexist.

gbaz:
> I don't think there's any problem in saying "the expressions f and \x -> f x are equivalent" because this statement is true if f is a function (ie. not "bottom") and we understand "equivalent" to mean "logically equivalent" (ie. ignoring runtime characteristics). Similarly, I don't think there's anything wrong with saying "Newton's laws of motion hold" even though we know they don't as velocity approaches the speed of light.
> Cf. [Fast and Loose Reasoning is Morally Correct](https://www.cs.ox.ac.uk/jeremy.gibbons/publications/fast+loose.pdf).

That is _not_ what that paper says, and I wish people would stop citing it for that purpose. That paper establishes a specific result about certain forms of equational reasoning in a total fragment of a language establishing _related_ properties in the entirety of the partial language - but properties which typically only hold in cases where inputs of functions are already taken to be finite. It says effectively "if you can reason in a setting where bottom cannot be constructed, then this holds in a setting where bottom can be constructed, assuming you never allow bottom to be introduced."(It says a fair amount more, actually, but this is what people usually make use of). This is a pet peeve because people's mis-citation of this article is actually a common negative sterotype of haskell programmers among pl researchers!

But more generally, "ignoring runtime characteristics" is never what we want to teach _programmers_ to do in writing programs. It only makes sense insofar as we are validating typeclass laws or the like.

For example, equational reasoning via program derivation techniques can famously transform O(n^2) algorithms into O(n log n) algorithms or even O(n) algorithms (depending on the algorithm). Obviously we do not want to teach programmers that since there is some notion of "equivalence" then they can just disregard these differences in complexity. And this holds even in a total setting where one doesn't worry about bottoms!

So I think equational reasoning is great! But we should make sure to teach it the _right way_ with all the limits and tradeoffs in mind.

telser:
Let me start by saying that I've recently had to change some code due to the lack of `DeepSubsumption` and those changes were frustrating/annoying, but not showstoppers and not a large issue by any means. So overall the negative impact for me and the code I work with most frequently has been very small.

My understanding and the impact to me leads me to believe that `DeepSubsumption` is a potential source of bugs and/confusion at best and definitely not the path I'd want for the future. Especially if given the choice of `DeepSubsumption` or `ImpredicativeTypes`, the choice would be `ImpredicativeTypes` hands down. 

> The point of adding back `DeepSubsumption` is not that we know no other option, but to retroactively do the deprecation cycle we should have all along,

I think this is highly important. But further in that deprecation cycle, I think including the future removal of `DeepSubsumption` from the beginning is highly desirable. Even if that is some point far off, such as 5, or 10 releases past the introduction of `DeepSubsumption`, that gives the library ecosystem, industrial users, and any one else a set timeframe to adjust. This would also strongly signal that `DeepSubsumption` is thin ice and should thus be avoided if at all reasonable. 

> sending a strong message that research and industry can coexist.
 
This is also something I find important and further motivates me to suggest the idea of `DeepSubsumption` supported until version X out of the gate. Putting a plan in place allows industry to prepare for the change. Also giving industrial users, or would-be users, a sense of being "for industry" as opposed to not. 

Note: I think that a final support date should be included initially _even_ if we know with some work an alternative could be proposed. From my perspective the ability to plan on this and signal to the larger to community is hugely valuable. 

Finally, is it known how well tooling might be able to detect `DeepSubsumption` enabled but unused? I'm thinking of 'hlint' specifically, but any tool would do, I imagine. Given that it is not recommended, being able to guide users to turn it off in modules where it is truly not used seems like a net benefit to me. Perhaps GHC could even emit this as a warning similar to the `-Wdeep-subsumption` that would warn when it is used, a `-Wunused-deep-subsumption` could warn that it is _never_ used in a given module?

parsonsmatt:
If the plan is to remove `DeepSubsumption` then don't bother putting it back in. I miss the feature and want it back, I don't need a band-aid to help in migrating away from it.

AntC2:
> @runeksvendsen  we understand "equivalent" to mean "logically equivalent" (ie. ignoring runtime characteristics). Similarly, I don't think there's anything wrong with saying "Newton's laws of motion hold" ...

(Thanks @gbaz for your response.)

I think documentation/tutorials shouldn't just ignore complications -- runtime, compile time, or brain time. If you want to stretch an analogy to Newton's Laws: it's like saying to use the Laws to calculate the trajectory of your cannonball, without mentioning the Laws apply _in vacuo_; if you're firing your cannonball on a planet with an atmosphere, you need to adjust for air resistance, otherwise your cannonball will fall short of enemy lines and land on your own people.

Ericson2314:
@parsonsmatt in addition, rather than just unconditionally removing the thing a few releases later, one could imagine keeping it around until a replacement deemed suitable by current users was created. I think it is better not to stipulate exactly under what circumstances it could later be removed later in this proposal, but at the same time it's good to not just assume this is some sort of empty token gesture. It does have good symbolism, but I do not believe it is so empty to be *just* symbolism.

simonpj:
You might think I'd be in favour of eventual removal, as @telser suggests.  But if in fact

* The code to support the feature is reasonably short, and well localised
* Continuting to support the feature does not impose othe consequential costs (e.g. intellectual complexity) to the compiler
* It does not affect those who don't use it.
* Its downsides are clearly signalled

then I see little cost to keeping it indefinitely. I see `NoMonoLocalBinds` as being in the same camp.  It is sometimes convenient, but it's a warty bit of the type system whose precise consequences are hard to predict.   But we aren't proposing to drop `NoMonoLocalBinds` any time soon.

cdsmith:
I am in agreement with @parsonsmatt that I miss the feature and want it back, but that I question the value of adding it back with a promise to remove it later.  If this proposal came with a clear statement like "this will be removed in GHC 9.10", then I don't think it's worth bothering to add it at all.

On the other hand, if it were added back with a "buyer beware" label and the intention of continuing to learn more to discover the path forward, then I strongly support _that_ approach.  It's natural that since GHC maintainers currently have a good story right now for what things look like without deep subsumption, they are moving in that direction, and I'm happy to leave this little asterisk that may or may not get resolved in time.

ekmett:
I have some code that I haven't touched since 9.0 because it is just too painful to do the eta-expansion everywhere syntactically and makes it inelegant, and I have other things to work on that are more important to me for the moment. It is not a lot, but where it exists it is dense. I very much welcome a future world where I can just skip over this intervening time period for that code and carry on working with it.

m4dc4p:
Absolutely in favor of this. I upgrade our 500+ module codebase from 8.10 to 9.0 and subsumption probably caused 50% of the changes (~1k lines)

However, they were all simple eta expansions. If possible, I’d rather see an extension/bug fix that just gets of the need for me to eta expand lambdas on my own. From @simonpj example above, the following won’t be accepted under GHC 9:

> readFreqSumFile file = readFreqSumProd $ withFile file ReadMode fromHandle

And must be rewritten to:


> readFreqSumFile file = readFreqSumProd $ withFile file ReadMode (\x -> fromHandle x)

I would love if GHC were smart enough to just rewrite the fromHandle expansion on its own, without the other risks of DeepSubsumption. 

If that’s not possible I wouldn’t myself ever disabling this extension.

ekmett:
> I would love if GHC were smart enough to just rewrite the fromHandle expansion on its own, without the other risks of DeepSubsumption.

Those risks come precisely from that rewrite, though.

L-as:
Yes but now it's explicit, and you can see if work is unshared.
You could also make use of ` -Wauto-lambda-destroys-sharing` from the proposal.

strake:
I laud this proposal strongly. I am still on GHC 8.10 since:
- the brokenness simplified subsumption brings in my code is too rife
- it's not feasible to teach the linter which η-expansions are needful

This would allow me to stop keeping up my private fork of GHC, and (i hope) contribute some local work back upstream (it's fairly scant, to be honest, so take this for what it's worth!).

mpickering:
I added a small note to the proposal about the fact the extension will not be deprecated. 

Please can I submit this proposal to the committee for consideration @nomeata ?

nomeata:
Sorry, I missed the mention! Thanks for nudging.

aspiwack:
I've been assigned as the shepherd for this PR. You should hear back from me on Monday. Otherwise, please complain.

aspiwack:
I've recommended acceptance to the committee.

simonpj:
Thanks Arnaud.

@mpickering Maybe you can implement Arnaud's typographical suggestions above.

telser:
> I added a small note to the proposal about the fact the extension will not be deprecated.
> 

I definitely appreciate this and the points raised by others here that have made it clear to me this is the path that should be taken.

> I've recommended acceptance to the committee.

This is great! Hope to see this work soon.

mpickering:
> Thanks Arnaud.
> 
> @mpickering Maybe you can implement Arnaud's typographical suggestions above.

Thanks @aspiwack, I have applied your two suggestions. Does this get merged now?

proposals/0000-deep-subsumption.rst
```diff
@@ -0,0 +1,304 @@
...
+Given that we don't think that using ``DeepSubsumption`` is a good idea, we also
+propose to improve diagnostics to help users migrate to simplified
+subsumption.
+
+* When ``-XDeepSubsumption`` is off, the error message can be improved to suggest
+  eta-expansion (and optionally enabling ``DeepSubsumption``).
+
+* When ``-XDeepSubsumption`` is on:
+
+  * A new warning ``-Wdeep-subsumption``, in ``-Wcompat`` and ``-Weverything`` (but
```
gridaphobe, line 107: The [User Guide](https://downloads.haskell.org/ghc/latest/docs/html/users_guide/using-warnings.html?highlight=wcompat#ghc-flag--Wcompat) says that `-Wcompat` is for warnings that will be enabled by default in the future. Is that the plan for `-Wdeep-subsumption`? If so, over what timeframe?

I also think it would be weird if we landed in a world where a language extension enabled a feature where all uses incur a warning. This proposal in particular seems to be motivated by pain felt by industrial users, who also tend to set `-Werror` to keep their code warning-free. Are we not just going to cause the same pain via the warnings?

gridaphobe:
@mpickering @simonpj I'm concerned by the requirement that *clients* must enable `-XDeepSubsumption` in order to gain the usability benefits. This is something we try to avoid. The language complexity cost should be paid by the library author, not the user.

Is there an alternative where a module `M` enables `-XDeepSubsumption`, and then either

1. all exported functions from `M` will have Deep Subsumption applied at call sites, or
2. specific functions (or specific arguments) can be marked as candidates for Deep Subsumption with some pragma?

simonpj:
> Is there an alternative where a module M enables -XDeepSubsumption, and then either

I'm sorry, but I have no idea how to achieve that, with reasonable complexity.
* We have no mechanism for exporting extensions along with particular functions
* Deep subsumption is a property of, well, the subsumption check, not of a particular function call

Doubtless these objections could be overcome, at the cost of a permanent tax on compiler complexity.  But this proposal is intendended to be a mitigation of a mis-feature that users have (understandably) adopted.  I'm content to have a couple of hundred lines of code in one place to support this mitigation.  I am much, much, much less keen on investing in a brand new mechanism with unknown knock-on consequences (e.g. recompilation avoidance checks and hash summaries).  

I agree this is less than ideal.  It amounts to requiring libray authors to say "in your cabal file, if you depend directly on my library (i.e. call functions from it), then switch on -XDeepSubsumption".  That doesn't sound too bad to me, although I accept that it is not as good as "do nothing".

L-as:
What about when you want to use a library that uses it but don't want to enable it yourself? It's not like IHP, etc., are unusable, they're just less ergonomic.

proposals/0000-deep-subsumption.rst
cdsmith: I believe that `-XPartialTypeSignatures` is already a GHC extension that enables a feature, all of whose uses incur a warning.  That said, I agree that it's weird, in that case as well as this one.

gridaphobe:
> > Is there an alternative where a module M enables -XDeepSubsumption, and then either
> 
> I'm sorry, but I have no idea how to achieve that, with reasonable complexity.
> 
> * We have no mechanism for exporting extensions along with particular functions
> * Deep subsumption is a property of, well, the subsumption check, not of a particular function call

I’m not suggesting we export an extension, but attach a flag to the exported or annotated functions. The flag would

1. enable deep subsumption in a much more targeted way than a module-wide extension 
2. be the responsibility of the library author to understand and apply to the relevant definitions

This feels rather like how we deal with OVERLAPPING or INLINE pragmas, both mechanically and in spirit, so it doesn’t seem like a new pattern to me. 

I’m not sure about the implementation complexity of a pragma-based approach vs an extension-based approach. I can certainly imagine the pragma approach being significantly more involved and that might be reason enough to discard it. 

But I wouldn’t go so far as to call deep subsumption a misfeature. It’s clearly serving a real use for advanced users in avoiding the need for explicit eta expands. We should try to solve that user-level problem in a good, integrated way.

simonpj:
> I’m not suggesting we export an extension, but attach a flag to the exported or annotated functions.

Yes, that's exactly what I'm concerned about.  As I say above:

* We have no mechanism for exporting extensions along with particular functions.  We do for instances, and it's a pervasive complication, affecting several data types, many functions, and recompilations avoidance hash-code generation.  I am very un-keen to add all this complication to functions.

* Deep subsumption is a property of, well, the subsumption check, not of a particular function call.   So even if we had a per-function flag, its not clear how to use it.  Consider
  ```
  f (let v = ... in
     case v of
       True -> \y -> blah
       False -> g x)
  ```
  If `f` is one of these imported-and-labelled functions, we don't need to eta-expand `f`; we need to eta-expand the call to `g`, buried inside `f`'s argument.  I'm sure it's doable, but it would be a new and pervasive complication.

I can see three posssibilities:

1.  Keep the status quo: no deep subsumption.  That's the no-op, but it leave us with unhappy users.
2.  Adopt the proposal. This continues to regard deep subsumption as a mis-feature (as did the committee when accepting the simple subsumption proposal), but puts in place a modest change to mitigate the effects of the change.
3.  Adopt a new, and as yet unwritten, proposal that sees deep subsumption as a desirable feature (reversing the committee's earlier attitude), and puts in place language features to support it, perhaps along the lines you describe.

I think that (3) would be controversial.  Speaking for myself I really do regard deep subsumption as a mis-feature, and I wish that I had not implemented it.  (If I had not, we would not be having this conversation.)  But I do not have a monopoly on truth and perhaps others would support (3).

Nevertheless I urge that we adopt (2) right now, and leave (3) for a subsequent conversation. No one is arguing against (2) -- the only debate is about whether we should go further.   While it is not perfect, it gets us 80% of the way with 0% of the controversy.  Moreover, we have a patch ready to go, just waiting for the commitee's blessing before it lands in a released GHC.

Nothing preclude someone proposing (3) as a separate proposal -- but adopting (2) allows us to offer immediate relief to users (which many of them have welcomed) while that debate takes place.

Would that be acceptable?

parsonsmatt:
I'm totally fine with the language extension. The error message *should* be able to say whether or not deep subsumption would fix the issue and suggest that, and library authors can always say "For using this API, please enable `DeepSubsumption`"

gridaphobe:
> * We do for instances, and it's a pervasive complication, affecting several data types, many functions, and recompilations avoidance hash-code generation.  I am very un-keen to add all this complication to functions.

Thanks, I did not realize how much complexity that added.

> I can see three posssibilities:
> 
> 1. Keep the status quo: no deep subsumption.  That's the no-op, but it leave us with unhappy users.
> 2. Adopt the proposal. This continues to regard deep subsumption as a mis-feature (as did the committee when accepting the simple subsumption proposal), but puts in place a modest change to mitigate the effects of the change.
> 3. Adopt a new, and as yet unwritten, proposal that sees deep subsumption as a desirable feature (reversing the committee's earlier attitude), and puts in place language features to support it, perhaps along the lines you describe.
>
> I think that (3) would be controversial. Speaking for myself I really do regard deep subsumption as a mis-feature, and I wish that I had not implemented it. (If I had not, we would not be having this conversation.) But I do not have a monopoly on truth and perhaps others would support (3).

I think we should distinguish between the user-visible behavior (I don't need to manually eta-expand my code to satisfy the typechecker) and the internal mechanism that drives the behavior (deep subsumption). Deep subsumption may be undesirable as an implementation strategy, but surely you agree that in an ideal world eta-expansion would not affect whether your program typechecks? We may not know how to achieve that in a satisfactory manner today, that's fine. I just don't want to write off the entire idea as undesirable.

I'm ok with accepting the proposal as-is to ease the pain felt by industrial users today.

gridaphobe:
> * We do for instances, and it's a pervasive complication, affecting several data types, many functions, and recompilations avoidance hash-code generation.  I am very un-keen to add all this complication to functions.

Thanks, I did not realize how much complexity that added.

> I can see three posssibilities:
> 
> 1. Keep the status quo: no deep subsumption.  That's the no-op, but it leave us with unhappy users.
> 2. Adopt the proposal. This continues to regard deep subsumption as a mis-feature (as did the committee when accepting the simple subsumption proposal), but puts in place a modest change to mitigate the effects of the change.
> 3. Adopt a new, and as yet unwritten, proposal that sees deep subsumption as a desirable feature (reversing the committee's earlier attitude), and puts in place language features to support it, perhaps along the lines you describe.
>
> I think that (3) would be controversial. Speaking for myself I really do regard deep subsumption as a mis-feature, and I wish that I had not implemented it. (If I had not, we would not be having this conversation.) But I do not have a monopoly on truth and perhaps others would support (3).

I think we should distinguish between the user-visible behavior (I don't need to manually eta-expand my code to satisfy the typechecker) and the internal mechanism that drives the behavior (deep subsumption). Deep subsumption may be undesirable as an implementation strategy, but surely you agree that in an ideal world eta-expansion would not affect whether your program typechecks? We may not know how to achieve that in a satisfactory manner today, that's fine. I just don't want to write off the entire idea as undesirable.

I'm ok with accepting the proposal as-is to ease the pain felt by industrial users today.

simonpj:
> Deep subsumption may be undesirable as an implementation strategy, but surely you agree that in an ideal world eta-expansion would not affect whether your program typechecks? 

Indeed!  Impredicative polymoprhism is another example.  Surely in an ideal world we would accept every single program that could be typed in System F.  For example `(\x. (x True, x "hello")) id` should be typable, by inferring `x :: forall a. a->a`.  But  we don't know how to do that.  I tried for 15 yrs, and at the end of that road found Quick Look.  but even QL rejects that program.  QL is a compromise: it accepts more programs than before, but not all the ones we could possibly want.   (MLF goes further, at greater cost, but still doesn't get everything.)

Deep subsumption is more challenging still, because we'd have to go beyond System F to accept these programs without eta expansion!  Absent type classes and evidence passing, one could imagine a system (more expresive than System F)  in which we could accept deep-subsumptive programs without eta expansion.  I don't know what that system is --  it's a research question -- and it'd be a radical chang to GHC's intermediate language.  But the killer is that we *do* have evidence passing and type classes.   With them, I don't know where to begin: so definitely a research question.

So yes, in an ideal world.   But that ideal world is out of my immediate reach.

aspiwack:
Dear all,

The committee has reach a conclusion, and this proposal is now **accepted**. As usual, I'll let @nomeata merge the PR.

ChickenProp:
The proposal still says it's unresolved whether to backport to 9.2. Did the committee make a decision on that?

aspiwack:
Oh! Thanks for flagging that. The committee is in favour of backporting to 9.2.

mjrussell:
Im not sure how much feedback the committee gets on proposals like this after they've been accepted, but I just wanted to say that adding `DeepSubsumption` to GHC 9.2.4 had a significant & positive impact on upgrading a fairly large production codebase from 8.10.7 to 9.2.4. Now our team can take advantage of many of the benefits of 9.2 while we work on unwinding the DeepSubsumption required areas gradually from the codebase.

Everyone's efforts to advocate for this, get it merged into GHC and port the change to 9.2 made what is a tricky upgrade already due to breaking changes in libraries much smoother, so thank you very much!

simonpj:
> Im not sure how much feedback the committee gets on proposals like this after they've been accepted, but I just wanted to say that adding DeepSubsumption to GHC 9.2.4 had a significant & positive impact on upgrading a fairly large production codebase from 8.10.7 to 9.2.4. Now our team can take advantage of many of the benefits of 9.2 while we work on unwinding the DeepSubsumption required areas gradually from the codebase.

I'm delighted to hear this.  Thanks for telling us!

david-christiansen:
This has been implemented, right? Shouldn't it get the "Implemented" label?

nomeata:
Thanks! (I thought I added it, but sometimes I close the tab to quickly :-))

PR #473, opened 2021-12-10
First-class existential types

goldfirere:
Following on from the recent [ICFP paper](https://richarde.dev/papers/2021/exists/exists.pdf), I propose first-class existentials for GHC.

[Rendered](https://github.com/goldfirere/ghc-proposals/blob/existentials/proposals/0000-existentials.rst)

PR #512, opened 2022-06-02
NoFieldSelectors as datatype annotation

parsonsmatt:
This proposal allows you to enable or disable generation of record field selectors on a per datatype basis. It extends the `NoFieldSelectors` and `FieldSelectors` module-level language extensions in a very similar way that `Overlapping` and `Overlappable` instance annotations extend the `OverlappingInstances` pragma.

[Rendered](https://github.com/parsonsmatt/ghc-proposals/blob/matt/field-selectors-scoped/proposals/0512-nofieldselectors-per-datatype.md)

### DIFF
--- a/proposals/0512-nofieldselectors-per-datatype.md
+++ b/proposals/0512-nofieldselectors-per-datatype.md
@@ -71,6 +71,10 @@ Another motivation is consistency.
 The `OverlappingInstances` pragma is used to allow *all* instances in a module to be overlapping or overlappable.
 But this is too coarse grained - you usually want to specify a *single* instance as `{-# OVERLAPPABLE #-}` directly in the instance body.
 
+Another motivation: today it's generally considered Bad Practice to use record syntax for the constructors of datatypes with alternatives, because this generates partial field accessors.
+With NoFieldSelectors, we can avoid this, but at the cost of turning off field selector generation for the entire module, which we might not want.
+Being able to control field selector generation on a per-datatype level lets you use this trick while keeping other "normal" records the same.
+
 ## 1.2 Proposed Change Specification
 
 Introduce an annotation `{-# NoFieldSelectors #-}` that can appear on a datatype, a constructor, or a field.
--- a/proposals/0512-nofieldselectors-per-datatype.md
+++ b/proposals/0512-nofieldselectors-per-datatype.md
@@ -7,12 +7,12 @@ implemented: ""
 
 This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/512>).
 
-# `NoFieldSelector`s as a datatype annotation
+# `NoFieldSelector`s as a datatype and field annotation
 
 GHC recently implemented a feature `NoFieldSelectors` which disables generation of record fields on a per-module basis.
 The motivation for this is that `DuplicateRecordFields` and `OverloadedRecordDot` can now be used to access fields on a record in a way that permits sharing field names on records, and even allows polymorphism on the record in question.
 
-I propose that we allow `NoFieldSelectors` as datatype annotation.
+I propose that we allow a datatype and field level override of the module default.
 
 ## Motivation
 
@@ -73,11 +73,13 @@ But this is too coarse grained - you usually want to specify a *single* instance
 
 ## Proposed Change Specification
 
-Introduce an annotation `{-# NoFieldSelectors #-}` that can appear on a datatype.
+Introduce an annotation `{-# NoFieldSelectors #-}` that can appear on a datatype, a constructor, or a field.
 
 This works exactly the same as if `NoFieldSelectors` were present for the module it is defined in, without interacting with the other datatypes.
 
-For completeness and consistency, allowing `{-# FieldSelectors #-}` as a datatype annotation also make sense.
+For completeness and consistency, allowing `{-# FieldSelectors #-}` as a datatype or field annotation also make sense.
+
+When the Modifiers syntax has landed in GHC, then we can switch to using modifiers instead of pragmas.
 
 ## Examples
 
@@ -121,9 +123,37 @@ data MkPoint = MkPoint
     }
 ```
 
+The pragma may apply to a single constructor in a sum type.
+
+```haskell
+module OnlyOneSelector where
+
+data CoolSumType 
+    = HasFieldSelectors { x :: Int, y :: Int }
+    | Doesn'tHaveSelectors {-# NoFieldSelectors #-} { a :: Char, b :: Char }
+```
+
+Field level annotations will look similar to `{-# UNPACK #-}` pragmas.
+
+```haskell
+{-# language FieldSelectors #-}
+
+module Foo where
+
+-- This datatype does not have the `secretInformation` field selectors generated
+-- but the other two are fine
+data 
+    Foo 
+    = Foo 
+        { name :: String
+        , age :: Int
+        , secretInformation :: {-# NoFieldSelector #-} [String]
+        }
+```
+
 ## Effect and Interactions
 
-This interacts with the `NoFieldSelectors` and `FieldSelectors` language extension by allowing a per-datatype override.
+This interacts with the `NoFieldSelectors` and `FieldSelectors` language extension by allowing a per-datatype or per-field override.
 Otherwise, this has the same effects and interactions as the original `NoFieldSelectors` extension.
 
 Syntactically, a common trick is to `grep` for `^(type|newtype|data) TypeName` to find the declaration for a given type name. Putting an annotation between the `data` and type name would break this sort of search.
@@ -158,7 +188,7 @@ The `StrictData` analogy is already implemented here with `BangPatterns`.
 -- This datatpye generates the `age` field selector, but not the `name`
 -- selector.
 data Foo = Foo
-    { {-# NoFieldSelector #-} name :: String 
+    { name :: {-# NoFieldSelector #-} String 
     , age :: Int
     }
 ```
@@ -171,11 +201,36 @@ module X where
 data Point = Point
     { x :: Int
     , y :: Int
-    , {-# FieldSelector #-} z :: Int 
+    , z :: {-# FieldSelector #-} Int 
     }
 ```
 
-Syntactically, the `{-# Blah #-}` style is lifted from `{-# Overlapping #-}`.
+While `{-# FieldSelector #-}` on a field can borrow syntax from `UNPACK` pragmas, the pragma for a datatype or a constructor could be changed.
+The proposal currently puts the pragma before the type name, but after the constructor name:
+
+```haskell
+-- Current status
+data {-# NoFieldSelectors #-} Foo = Bar {-# FieldSelectors #-} { a :: Int }
+```
+
+But it could just as easily move the pragma before or after the type or constructor.
+
+```haskell
+-- Consistent after:
+data Foo {-# NoFieldSelectors #-} = Bar {-# FieldSelectors #-} { a :: Int }
+-- Consistent before:
+data {-# NoFieldSelectors #-} Foo = {-# FieldSelectors #-} Bar  { a :: Int }
+
+-- Inconsistent, but the other way
+data Foo {-# NoFieldSelectors #-} = {-# FieldSelectors #-} Bar { a :: Int }
+```
+
+Putting the pragma between the keyword and the type name has symmetry with `instance {-# OVERLAPPING #-}`.
+However, this makes it harder to `grep` for definitions.
+Putting the pragma before the field has symmetry with `{-# UNPACK #-}`, and it seems reasonable to stick with that convention for fields.
+There is no existing convention for constructors.
+
+Syntactically, the `{-# Blah #-}` style is lifted from `{-# Overlapping #-}` and `{-# Unpack #-}`.
 However, it's pretty ugly and verbose.
 A further extension could introduce new keywords that are only used in record declaration syntax with the same meaning, allowing you to write, eg,
 
@@ -185,6 +240,8 @@ data Point = Point
     , y :: lazy field Int
     }
 ```
+
+We could also use the modifiers syntax.
 
 ## Unresolved Questions
 
@@ -197,6 +254,8 @@ The datatype for a record declaration is extended with a field `Maybe FieldSelec
 When GHC is deciding whether or not to generate field selectors, this field is checked as a final source of truth.
 
 This change is also propagated to `template-haskell`.
+
+The implementation of this feature will be sponsored by Mercury.
 
 ## Endorsements
 
--- a/proposals/0512-nofieldselectors-per-datatype.md
+++ b/proposals/0512-nofieldselectors-per-datatype.md
@@ -5,7 +5,7 @@ ticket-url: "https://github.com/ghc-proposals/ghc-proposals/pull/512"
 implemented: ""
 ---
 
-This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/512>).
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/512).
 
 # `NoFieldSelector`s as a datatype and field annotation
 
--- a/proposals/0512-nofieldselectors-per-datatype.md
+++ b/proposals/0512-nofieldselectors-per-datatype.md
@@ -1,13 +1,11 @@
 ---
-author: Your name
+author: Matt Parsons
 date-accepted: ""
-ticket-url: ""
+ticket-url: "https://github.com/ghc-proposals/ghc-proposals/pull/512"
 implemented: ""
 ---
 
-This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/0>).
-**After creating the pull request, edit this file again, update the number in
-the link, and delete this bold sentence.**
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/512>).
 
 # `NoFieldSelector`s as a datatype annotation
 
--- a/proposals/0512-nofieldselectors-per-datatype.md
+++ b/proposals/0512-nofieldselectors-per-datatype.md
@@ -241,7 +241,9 @@ data Point = Point
     }
 ```
 
-We could also use the modifiers syntax.
+We could also use the modifiers syntax, when that is implemented.
+At that point, the pragma syntax could be deprecated and marked for deletion.
+Since pragmas are comments, GHC could reasonably provide a good warning message with a migration path, even after the pragma parsing/using code was deleted.
 
 ## 1.7 Unresolved Questions
 
--- a/proposals/0512-nofieldselectors-per-datatype.md
+++ b/proposals/0512-nofieldselectors-per-datatype.md
@@ -173,7 +173,7 @@ module Foo where
 -- but the other two are fine
 data Foo = Foo { name :: String
                , age :: Int
-               , %NoFieldSelectors secretInformation :: [String] }
+               , secretInformation %NoFieldSelectors :: [String] }
 ```
 
 ## 1.4 Effect and Interactions
--- a/proposals/0512-nofieldselectors-per-datatype.md
+++ b/proposals/0512-nofieldselectors-per-datatype.md
@@ -16,6 +16,14 @@ I propose that we allow a datatype and field level override of the module defaul
 
 ## 1.1 Motivation
 
+There are three main motivations for this proposal:
+
+1. Generating records in `TemplateHaskell` without field selectors
+2. Ability to suppress selectors for partial fields
+3. Consistency with `OverlappingInstances` pragma being specialized
+
+### 1.1.1 Template Haskell Record Generation
+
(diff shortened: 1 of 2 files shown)


nomeata:
Have you considered a syntax based on modifiers (https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0370-modifiers.rst) rather than pragmas? If so maybe explain in the proposal why pragmas are better than modifiers here?

(I like the sound of crickets, btw :-))

michaelpj:
Another motivation: today it's generally considered Bad Practice to use record syntax for the constructors of datatypes with alternatives, because this generates partial field accessors. With `NoFieldSelectors`, we can avoid this, but at the cost of turning off field selector generation for the entire module, which we might not want. Being able to control field selector generation on a per-datatype level lets you use this trick while keeping other "normal" records the same.

adamgundry:
I'm in favour of being able to control this at a fine-grained level. I don't have a strong opinion on syntax, but modifiers already seem to accommodate annotations on datatypes and individual fields. I don't anticipate this being difficult to implement in the renamer once the parser has been modified, because GHC already tracks `NoFieldSelectors` status on a per-field basis.

adamgundry:
I thought of another case in which this would be useful, perhaps even on a per-field basis. Consider the following module:
```hs
{-# LANGUAGE DuplicateRecordFields #-}

data S = MkS { foo :: Int }
data T = forall a . MkT { foo :: a }

hmm = foo
```

This is currently rejected (by GHC 9.2.2) with:
```
M.hs:6:7: error:
    Ambiguous occurrence ‘foo’
    It could refer to
       either the field ‘foo’, defined at M.hs:4:27
           or the field ‘foo’, defined at M.hs:3:16
```

However commenting out `S` shows that `foo` from `T` is not actually usable:
```
M.hs:6:7: error:
    • Cannot use record selector ‘foo’ as a function due to escaped type variables
      Probable fix: use pattern-matching syntax instead
    • In the expression: foo
      In an equation for ‘hmm’: hmm = foo
```

This is because the second `foo` gets a "naughty" record selector binding, to avoid making a definition where an existential would escape its scope. Now it would be better if GHC accepted the original module, with `foo` referring unambigiously to `S`. Unfortunately this is difficult, because it doesn't figure out which selectors are "naughty" until the type-checker, which runs too late to affect name resolution. However, with the new annotation in this proposal, users could explicitly mark such fields as not having selectors.

Ericson2314:
Nice observation!

1. We can make a warning for any time someone *doesn't* use `%(No FieldAccessors)` and GHC discovers at type checking time it is being asked to create such field selectors.

2. Conversely, if https://github.com/ghc-proposals/ghc-proposals/pull/473 is accepted such projections well be well on their way to being well typed. We need `foreach` too to be able to write down the type / make the projector a first-class function, but we'll have the power to make more ad-hoc per-field judgements that suffice for first-order usage.

parsonsmatt:
I'm curious if there's anything I need to do to help move this forward

googleson78:
You can [bring this forward to the committee](https://github.com/ghc-proposals/ghc-proposals#how-to-bring-a-proposal-before-the-committee) to ask for them to make a vote on it. I guess there also seem to be some voices (at least Joachim and Richard) who would prefer this proposal to be using modifiers instead, so maybe that's another direction to make prograss as well.

proposals/0512-nofieldselectors-per-datatype.md
```diff
@@ -0,0 +1,203 @@
...
+
+Since this is a specialization of an existing language behavior, I don't believe there is anything to be resolved here.
+
+
+## Implementation Plan
+
+The datatype for a record declaration is extended with a field `Maybe FieldSelectorOverride`.
+When GHC is deciding whether or not to generate field selectors, this field is checked as a final source of truth.
+
+This change is also propagated to `template-haskell`.
```
nomeata, line 199: Can you already say whether you plan to implement this yourself? Or know who will?

nomeata:
Right, once you are content with the amount of feedback you got and discussion ebbed down you can submit it to the committee. Ping me here when that’s the case.

parsonsmatt:
Are modifiers in GHC? If not, when's the expected timeline for the implementation?

SInce this appears to be a relatively small change, I'd love to get the `{-# #-}` style annotations in for an ASAP thing, and add `%(No FieldSelectors)` when that becomes available.

proposals/0512-nofieldselectors-per-datatype.md
parsonsmatt: I've assigned the implementation to @wavewave, Mercury's GHC engineer, if it gets approved/accepted.

simonpj:
I'm generally supportive too.  You'll need to modify the proposal to use the modifiers syntax. 

Modifiers are not part of GHC right now, so there may be some ground work to do here.

parsonsmatt:
@nomeata I think this is ready for review

nomeata:
Thanks! @Icelandjack, please shepherd this one to a conclusion

simonpj:
* The link in "This proposal is discussed at this pull request" is broken.
* Please add section numbers (it's a one-line change, I always forget what).
* You point out that `data {-# NoFieldSelectors #-} Foo = ....` breaks a certain sort of grep.  So why not put it afterwards?  `data Foo {-# NoFieldSelectors #-} = ...`, an alternative you suggest.

parsonsmatt:
> Please add section numbers (it's a one-line change, I always forget what).

I can't find any reference for section numbers in GitHub markdown. Could you point me to a relevant doc?

> You point out that data {-# NoFieldSelectors #-} Foo = .... breaks a certain sort of grep. So why not put it afterwards? data Foo {-# NoFieldSelectors #-} = ..., an alternative you suggest.

The initial selection is for symmetry with `instance {-# OVERLAPPABLE #-} cxt => cls typ ...`. However, putting it after the type is symmetric with the `module Foo {-# WARNING ... #-} (..) where`. I'd be happy to revise the proposal to go in either direction.

simonpj:
> I can't find any reference for section numbers in GitHub markdown. Could you point me to a relevant doc?

@nomeata can you advise -- and perhpas update the tempate to have section numbers?

nomeata:
We have it in the .rst template. I don't know if this is even possible with markdown. I wonder if we should stop supporting markdown proposals - I convert them using pandoc anyways, to keep the accepted proposals uniform.

simonpj:
Aha.  I thought the `.rst` template was the only path.   OK, Matt, I guess you can either number manually or switch to `.rst`.

nomeata:
@parsonsmatt, are you by any chance attending Haskell Symposium? If so, would you like to present this proposal at the [GHC Proposals session](https://icfp22.sigplan.org/home/haskellsymp-2022#GHC-Proposals-Session)?

parsonsmatt:
Unfortunately not! This is the first I've heard of it.

nomeata:
@int-index, can I leave this proposal to you?

int-index:
@nomeata Ok.

@parsonsmatt The proposal still uses the pragma syntax instead of the modifier syntax. The discussion shows that everyone is in favor of the modifier syntax. Do you intend to update the proposal text accordingly?

parsonsmatt:
100% in favor of adding a modifier syntax when that lands in GHC. Until then I'd like to be able to begin implementation of this feature asap.

int-index:
@parsonsmatt Features are implemented once but stay with us forever (most of them do). So I find it important to get them right. There’s no general framework for pragmas in GHC, each one is implemented separately, so adding another one is a perpetual maintenance burden. Modifiers, on the other hand, offer a generic and extensible framework, and the proposed feature would fit into this framework quite well.

What I hear you saying is “I don’t want to predicate this feature on modifiers because I’d like to volunteer my time implementing this feature, but I can’t volunteer my time to implement the entire modifiers framework”. I hope I’m not misinterpreting. If I got your position right, then let me point out that the implication of it is that you’re willing to impose maintenance burden for a syntax known to be suboptimal on all future GHC contributors in perpetuity because, quote, “I'd like to be able to begin implementation of this feature asap”. The argument would be strong if the “asap” part was justified, but as far as I can tell, there’s a readily available workaround – just enable `NoFieldSelectors` per module.

So instead of doing things “asap” I suggest we do them thoroughly. Implementing modifiers would facilitate other proposals as well, such as #390.

If you disagree and you’d like me to bring this before the committee anyway, I will do so (recommending rejection). The other committee members could override my recommendation. If you agree that modifiers are a better design, then the proposal text should match this vision.

parsonsmatt:
Presumably there will be some deprecation period for `{-# OVERLAPPING #-}` and friends, with a gentle push to move to the (at that point) available modifiers. Simply include `NoFieldSelectors` etc in that push.

I'm making a proposal to modify GHC, not some hypothetical future incarnation of GHC with an unspecified timeline.

int-index:
> Presumably there will be some deprecation period for `{-# OVERLAPPING #-}` and friends, with a gentle push to move to the (at that point) available modifiers. Simply include `NoFieldSelectors` etc in that push.

My point exactly – you are proposing a feature destined to be deprecated. But deprecation does not mean removal from GHC’s codebase. The syntax will stay, and the code supporting it will also stay. I see that as suboptimal.

> I'm making a proposal to modify GHC, not some hypothetical future incarnation of GHC with an unspecified timeline.

I’m making an argument for rejecting it.

parsonsmatt:
> My point exactly – you are proposing a feature destined to be deprecated. But deprecation does not mean removal from GHC’s codebase. The syntax will stay, and the code supporting it will also stay. I see that as suboptimal.

The [proposal for modifiers](https://github.com/ghc-proposals/ghc-proposals/pull/370) was merged about two years ago. There are no [PRs implementing it](https://gitlab.haskell.org/ghc/ghc/-/merge_requests?scope=all&state=opened&search=modifier) on GitLab. There aren't even any branches for it. When will it happen? Who knows! I don't. It's clearly not Richard's priority.

If this is sufficient reason to reject a proposal, then I know not to bother making any more.

int-index:
> If this is sufficient reason to reject a proposal, then I know not to bother making any more.

I just want to make it very clear that it is not in my power to reject a proposal single-handedly. As I have been assigned to shepherd it, I can make a _recommendation_ for rejection or acceptance. I do not speak for the entire committee.

So, no, this is not a sufficient reason to reject a proposal – it is my personal argument for doing so (which I will also present to the committee, unless the proposal is changed).

The options I am considering as successful outcomes for this proposal are:
1. You would find my argument compelling and decide to implement modifiers yourself (and then this proposal)
2. You would find my argument compelling and decide to wait for somebody else to implement modifiers (if ever)

Well, I see that I failed to convince you and only demotivated you. Which is not what I intended. Do you want me to proceed with the formal process?

nomeata:
This is indeed an unhealthy situation to be in, with an accepted proposal laying out a possible future, and one that other proposals can (and should be able to) build upon. It's the kind of situation that #517 hopes to help to avoid, but it's too late for that now anyways.

It would be good to know if anyone is about to implement modifiers? Maybe now, with this as extra motivation? @goldfirere or @aspiwack, do you know if it's high on anyone's TODO list?

aspiwack:
> It would be good to know if anyone is about to implement modifiers? Maybe now, with this as extra motivation? @goldfirere or @aspiwack, do you know if it's high on anyone's TODO list?

I'm not aware of what's been happening. If my memory serves me right, part of the discussion around modifiers were implementation-driven. So there would be a prototype somewhere? I don't know what became of it, if it does indeed exist.

---

I'm not familiar with the process to add pragmas to declaration, so I don't know, but I would like to point out that it may not actually be harder to add bespoke syntax to the parser which is compatible with the modifier syntax.

It is out of the question, for me, to mandate that @parsonsmatt implements the modifier syntax before adding a useful feature. But if we can be compatible with the foreseen future, then all the better. I will personally, not oppose the proposal on the ground that it uses the pragma syntax rather than the modifier syntax. But I would like, in that case, it to been argued in the proposal why the modifier syntax is not an reasonable option in the short term.

goldfirere:
While I'm sympathetic to @int-index's arguments above, I personally wouldn't want to reject a proposal solely on the grounds that it would be better to have after the modifiers syntax lands. I am not working on that implementation, and have no plans to.

int-index:
> There aren't even any branches for it. When will it happen? Who knows!

@parsonsmatt I have taken another look at the modifiers proposal, and it seems that there are no blockers to implementing it. In fact, I can probably do it in parallel with my other tasks. So I can volunteer to implement modifiers in GHC in 2023. Does that alleviate your worry?

parsonsmatt:
Frankly, no, I am not interested in blocking this proposal on an unimplemented proposal. If modifiers are implemented before this proposal is approved then I would be fine to rewrite the proposal in terms of modifiers. But I think the proposal is ready for review on its own merits as-is.

int-index:
Very well. Please describe your reasoning in the Alternatives section and I will make a recommendation based on the current proposal text.

simonpj:
Matthew, can I suggest that you add Michael's suggestion as an additional motivation.  He wrote

> Another motivation: today it's generally considered Bad Practice to use record syntax for the constructors of datatypes with alternatives, because this generates partial field accessors. With NoFieldSelectors, we can avoid this, but at the cost of turning off field selector generation for the entire module, which we might not want. Being able to control field selector generation on a per-datatype level lets you use this trick while keeping other "normal" records the same.

sol:
> Matthew, can I suggest that you add Michael's suggestion as an additional motivation. He wrote
> 
> > Another motivation: today it's generally considered Bad Practice to use record syntax for the constructors of datatypes with alternatives, because this generates partial field accessors. With NoFieldSelectors, we can avoid this, but at the cost of turning off field selector generation for the entire module, which we might not want. Being able to control field selector generation on a per-datatype level lets you use this trick while keeping other "normal" records the same.

I think there is hardly ever a useful application for partial field selectors.

Now, with this proposal, you will be able to tell the compiler to do "the sane thing" on a per-datatype level. But you still have to tell the compiler.

What you really want is the compiler to do "the sane thing" by default, that is to "never generate partial field selectors" (obviously guarded behind a language extension, for backwards compat)

Say, in an ideal world you would have both:

1. A per-datatype `{-# NoFieldSelectors #-}` as described in this proposal.
2. A dedicated language extension `NoPartialFieldSelectors` that disables field selectors for *sum record types*  in general.

That way you could avoid this "Bad Practice" by enabling `NoPartialFieldSelectors` globally vs. having to do it on a per-datatype basis.

Still, given that

1. We don't have `NoPartialFieldSelectors` today (nor do we have a proposal for it)
1. A per-datatype `{-# NoFieldSelectors #-}` will improve the situation for sum record types

I guess "yes", it makes sense to add "Michael's suggestion as an additional motivation". My point here is that while "yes", this proposal improves the situation for sum record types, for sum record types it is still only a half measure.

parsonsmatt:
> Matthew, can I suggest that you add Michael's suggestion as an additional motivation. He wrote

I don't really think this helps *that much* since GHC is happy to define `HasField` instances for partial records. I have a [related proposal](https://github.com/ghc-proposals/ghc-proposals/pull/535) that should improve *that* part of the story, at least

parsonsmatt:
@simonpj I've incorporated Michaele's motivation. Let me know if there are further changes you'd like to see.

Thanks!

simonpj:
> @simonpj I've incorporated Michaele's motivation. Let me know if there are further changes you'd like to see.

You write 

    Another motivation: today it's generally considered Bad Practice to use record syntax for the
   constructors of datatypes with alternatives, because this generates partial field accessors.

This is really hard going without an example.  What is a "datatype with alterantives". Maybe one with more than one data constructor?  Well, they certainly don't always produce partial field accessors:
```
data T = T1 { x :: Int }  | T2 { x :: Int }
```
Best just to show your example, with your desired per-datatype pragma.    (Oh and include an example of why you still want the field names for conctruction and pattern matching; i.e. you can't just use positional instead.)

Presentationally,j if it were me, I'd list my motivations in the movitation section

1. Generating record declarations in Template Haskell
2. Ability to suppress selectors for partial fields.
3. Consistency with OVERLAPPING for instance declarations

and then elaborate in a subsection for each.  Makes it easier to parse.

----------------

To be clear, I'm generally in support.

parsonsmatt:
That's a good suggestion, thanks! I've restructured the Motivation section to have more structure and an example for partial fields.

AntC2:
[Possibly related](https://discourse.haskell.org/t/field-selectors-for-newtypes/5812): with `-XNoFieldSelectors` set in general, allow `-XNewtypeFieldSelectors`. (I'm filling in some details from that discussion.)

Newtype field selectors are 'safe' because there's only one constructor (so no partial function); and little risk of `DuplicateFields`, if you follow the convention of naming the FieldSelector for its datatype (prefixed).

uhbif19:
Had need for that extension while refactoring GHC. My need was to enable field selectors for some types, there they will be total, while disable for or other.

int-index:
In December I initiated another round of discussion of this proposal in the committee, and the consensus has more or less emerged.

* Let's proceed with modifiers-style syntax. @parsonsmatt you can revise the proposal accordingly, or delegate this to me if you'd like to.
* I will implement #370. In the meantime, this particular feature can simply mimic the syntax of modifiers to be future-compatible. Supporting this particular use case does not really require a full implementation of modifiers, it's only a matter of using a forward-compatible syntax.

The proposal itself is now **conditionally accepted**: it needs to be revised to use syntax that is consistent with #370, but the overall idea is solid.

parsonsmatt:
I will delegate that modification to you. Thank you!

parsonsmatt:
There's a significant downside to the modifier syntax: backwards compatibility.

If I write `data Foo = Foo { unFoo :: {-# NoFieldSelector #-} Int }` today, then GHC can compile the module fine with a warning:

```
asdf.hs:1:27: warning: [GHC-42044] [-Wunrecognised-pragmas]
    Unrecognised pragma: NOFIELDSELECTOR
  |
1 | data Foo = Foo { unFoo :: {-# NoFieldSelector #-} Int }
  |                           ^^^^^^^^^^^^^^^^^^^
Ok, one module loaded.
```

For libraries that define records and don't want selector functions defined, then the pragma syntax will *parse* fine, but the exposed functions will differ.

---

I am really disheartened by the GHC proposal process, yet again. This proposal is an obvious extenion of existing behavior that uses existing conventions and no novel syntax, and was regarded as a good idea in all commentary within a week of opening it. We could have had this functionality in GHC 9.6 - an 8 month delay between proposal and release.

Instead, there was an *18 month syntax bikeshed* about whether we should use the syntax *that already exists and is already used consistently for this purpose*, or require implementing a novel syntax that is backwards incompatible with older GHCs.

At the absolute earliest, this could be implemented for GHC 9.10 - there are [22 days to get it in before the fork](https://gitlab.haskell.org/ghc/ghc/-/milestones/380#tab-issues). The implementation difficutly has increased substantially, since the parser must now be modified. Even if we get it in for GHC 9.10, we're looking at an estimated timeline of April 2024 for release - 22 months between proposal and release.

simonpj:
The proposal was submitted to the committee on 2 Oct 2022, and we let it lie for over a year.  This is down to the Steering Committee -- we promise a turn around in weeks not months, and this was an egregious exception.  As co-chair I really apologise for that.  

*Note to proposal authors*: you should absolutely feel empowered to ping the committee (or me personally) if your proposal is in the "under committee review" box for longer than a few weeks.

Running a painstaking process with volunteers alone is difficult, and there will be mistakes.  This was one of them.

We will be looking for new committee members soon.  Perhaps some of those who wish (rightly!) that the process was brisker might consider putting themselves forward.  

-------------

There is always the question of whether to continue down a path (pragmas) that we have been using, or to adopt a new one (modifiers) that we think will serve us better.  A good thing about this particular proposal is that Vlad has volunteered to implement the modifiers proposal.

Yes it means that older compilers won't understand the new syntax, but that is true of many new features, not just this one.

int-index:
> As co-chair I really apologise for that.

For the record, this incident is actually on me, not Simon. As a shepherd assigned to this proposal, I should've done a better job of driving the process, and Simon has repeatedly reminded me to act, which I didn't do in a timely fashion.

In retrospect, I've been less active as a committee member than I hoped I would be when I signed up for this role, mostly due to various personal issues that left little time and energy for work in general, not to mention volunteer work. I apologise for any inconveniences that it caused.

proposals/0512-nofieldselectors-per-datatype.md
```diff
@@ -0,0 +1,224 @@
...
+```haskell
+{-# language FieldSelectors #-}
+
+module Foo where
+
+-- This datatype does not have the `secretInformation` field selectors generated
+-- but the other two are fine
+data Foo = Foo { name :: String
+               , age :: Int
+               , %NoFieldSelectors secretInformation :: [String] }
```
int-index, line 176:
```suggestion
               , secretInformation %NoFieldSelectors :: [String] }
```

This was an oversight on my part. Fields are the only context where modifiers are written after, not before the modified entity.

int-index:
@adamgundry Let's merge this?

adamgundry:
Thanks @parsonsmatt and @int-index, I've marked this proposal as accepted and opened https://gitlab.haskell.org/ghc/ghc/-/issues/24481 to track its implementation.

PR #514, opened 2022-06-05
(more) levity-polymorphic $

wygulmage:
`($)` could be levity polymorphic in its function parameter's input as well result.

### DIFF
--- a/proposals/0000-more-levity-polymorphic-$.rst
+++ b/proposals/0000-more-levity-polymorphic-$.rst
@@ -1,0 +1,54 @@
+
+Levity-polymorphic ``$``
+========================
+
+.. author:: Keith Wygant
+.. date-accepted::
+.. ticket-url::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. contents::
+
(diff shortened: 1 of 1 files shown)


yy0zz:
Other alternative: write `I# do (2# *#) do 1# +# 8#` (with BlockArguments).

treeowl:
There's a much simpler possibility I brought up a few years ago, with all the same (potential) strictness and performance drawbacks:

```haskell
($) :: a -> a
($) f = f
```

wygulmage:
@treeowl, I avoided that because I assumed that the specialized signature is supposed to show the intended use of the function, and might help with type inference in edge cases. That said, it probably would be more newbie-friendly, since understanding `f `id` exp` gives insight into polymorphism while understanding GHC's implementation of polymorphic kinds is less generally useful.

simonpj:
Hmm.  I don't see why we couldn't do this.  The key is to define
```
($) f = f
```
and not
```
($) f x = f x
```
because the latter has a representation-polymorphic lambda, which we can't generate code for.

Mind you the reverse-apply function
```
revapp x f = f x
```
could not be rep-poly.  But I suppose that it is less common, and we'd just have to live with the assymetry.

So I see no technical problem at the GHC end.  Worth asking the Core Libraries Committee.

goldfirere:
I agree that this is possible, but I would say that this falls more squarely as a library issue than a language one. Perhaps check out https://github.com/haskell/core-libraries-committee

goldfirere:
No worries @wygulmage. It's very hard to figure out where to make proposals -- thanks for submitting one!

PR #71, opened 2017-08-04
Extend -Wall with incomplete-uni-patterns and incomplete-record-updates

tomjaguarpaw:
This suggestion [received great support on Haskell Reddit](https://www.reddit.com/r/haskell/comments/6q9tcp/ghc_warnings_you_should_use_in_addition_to_wall/dkvrk0e/).

PR #516, opened 2022-06-12
Introduce `-Wincomplete-record-selectors`

adamgundry:
The [proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0516-incomplete-record-selectors.rst) has been accepted; the following discussion is mostly of historic interest.

This proposal introduces a new warning flag, `-Wincomplete-record-selectors`, which emits a warning at use sites of partial selector functions or partial `OverloadedRecordDot` syntax.

[Rendered](https://github.com/adamgundry/ghc-proposals/blob/incomplete-record-selectors/proposals/0000-incomplete-record-selectors.rst).


proposals/0000-incomplete-record-selectors.rst
```diff
@@ -0,0 +1,177 @@
...
+4. Optionally, expressions such as ``x (T1 { x = 0, y = True })`` and ``case r
+   of { T2 _ -> 0 ; _ -> x r }`` may skip the warning, depending on the
+   capabilities of the pattern match coverage checker.
+
+
+Effect and Interactions
+-----------------------
+The ``NoFieldSelectors`` extension allows users to suppress field selector
+functions, thereby avoiding the risk of calling a partial selector function in
+an expression.  This does prevent use of ``OverloadedRecordDot`` for the field,
```
googleson78, line 138: I think you mean
```suggestion
an expression.  This does not prevent use of ``OverloadedRecordDot`` for the field,
```

Ericson2314:
Can we also make `-Wpartial-fields` not go off with `-XNoFieldSelectors` (if that isn't done already?) I agree partial field should not a be a problem, but I also still rather have definition-side validation.

Similarly, it would be nice to, rather than warning on solving, avoid creating `HasField` instances for partial fields altogether.

adamgundry:
> Can we also make `-Wpartial-fields` not go off with `-XNoFieldSelectors` (if that isn't done already?) I agree partial fields _themselves_ should not a be a problem, but I also still rather have definition-side than use-side validation nothing will go wrong.

Currently, even with `NoFieldSelectors`, there are still two sources of partiality from partial field declarations:
 * Record updates
 * `HasField` instances

Thus I think it is appropriate for `-Wpartial-fields` to continue warning about such cases.
 
> Similarly, it would be nice to, rather than warning on solving, avoid creating `HasField` instances for partial fields altogether.

Yes, that would be possible, but I'm not sure what the best way to control it would be. One could imagine adding yet more options/modifiers so that a declaration site could say "don't allow partial fields to be used with record selection, update or `HasField`". (#512 is somewhat related.)

But I still think we want `-Wincomplete-record-selectors` (and the existing `-Wincomplete-record-updates`), because they allow downstream code to avoid partiality even if it uses a library that exposes partial fields. And once we have those, I don't feel hugely motivated to add more options for control at the definition site. After all, the library author can choose to:
 * enable `-Werror=partial-fields` and avoid defining partial fields in their library, or
 * define partial fields in their library and tell their users to set  `-Werror=incomplete-record-selectors -Werror=incomplete-record-updates`.

goldfirere:
I'm generally in support. Thanks for writing up such a nice proposal!

There is a corner case that the proposal doesn't handle -- but I think it could:

```hs
data G a where
  MkG1 :: { x :: Int } -> G Bool
  MkG2 :: { y :: Double } -> G Char
```

According to the proposal, both `x` and `y` are partial record selectors. And using `x` or `y` on an argument of type `G a` should indeed warn. But what about this:

```hs
f :: G Bool -> Int
f r = getField @"x" r
```

That seems like it should *not* warn at all, but I believe the proposal would have a warning here.

Basically, I'm just asking if the pattern-match completeness check could be extended to cover `HasField` constraints.

Ericson2314:
@adamgundry it sounds like then you are leaning on `-Wpartial-fields` being *not* included in `-Wall` to justify it being somewhat redundant? IIUC, with this proposal:

1. All partiality that might result from partial fields finally gets warmed downstream/separately.
2. `-Wpartial-fields` has "false positives", doesn't catch anything that other warnings wouldn't.

If so, I suppose that is good. I would advocate deprecating `-Wparitial-fields` and letting linters do that instead perhaps.

adamgundry:
Thanks for the feedback! At long last I have revised this proposal as follows:
 * The Proposed Change Specification now reflects @goldfirere's observation that the pattern match checker may be able to determine that certain `HasField` uses are not partial. I've added to the examples as well. I have deliberately left the exact capabilities of the check to be implementation-defined, because I think trying to specify them here will be more trouble than it is worth.
 * The Effect and Interactions section now discusses `-Wpartial-fields` in response to @Ericson2314's comments. I argue there that we should leave `-Wpartial-fields` alone as there are cases where it is useful, even given the new `-Wincomplete-record-selectors`.

@nomeata I would like to submit this proposal to the commitee.

cdornan:
LGTM!

simonpj:
I'm generally supportive of this.

simonpj:
I'm generally in support.  Two points

* Please add section numbers. Doing so is easy.
* Why is the definition site flag called `-Wpartial-fields` and the use-site one called `-Wincomplete-record-selectors`? Better to have `-Wpartial-fields` and `-Wpartial-field-uses` or some consistent naming like that.  We could add `-Wpartial-field-bindings` as a synonym for `-Wpartial-fields.`
* The GADT section seems solid, because it is driven by type information, and will work for HasField constraints; but the "long distance information" section is very squishy, most especially so for HasField constraints.  I'd like to see that section put after the GADT section, and flagged more explictly as "the may use best efforts to help here, but isn't expected to".  In constrast, I think that using GADT info to suppress silly warning could (and perhpas should) be part of the spec.

Mind you, now I think about it, the warnings for solving HasField constraints for GADTs should really be postponed until the end of typechecking, because we may not know the full type of the record until then.  That's quite painful in implementation terms, and the benefit is small.  Gah.  One more effort-consuming corner.  Maybe we should make the GADT bit fully optional too.  That is: to be sure to avoid these warnings, just don't use partial fields at all.

(I'm not sure if we have a view about "best-efforts" warnings, where warnings might come and go according to how best efforts work out.  I could see uses being unhappy about that.)

AntC2:
> giving field names to heterogeneous multi-constructor datatypes, ... these are sometimes useful, 

"sometimes useful" doesn't sound like a strong motivation. Do we have metrics on how common are multi-constructor datatypes that use field names?

> common partial functions such as `head`,

Yes, `List` aka `data [a]` is a multi-constructor datatype, but doesn't use field names. Nor does `Data.Set` nor `mtl`.

I conjecture core libraries either define datatypes with multi-constructors but very few fields (so not needing field labels); or  for holding a block of systems values -- where you want field labels for sanity -- but with only a single constructor.

There's something of a move to represent an empty container/different constructors as a distinct type, so that partiality dangers get lifted into type-checking.

> @simonpj That's quite painful in implementation terms, and the benefit is small.

Given the examples of GADTs and 'long range information', and `HasField`; I'm wondering whether quite a bit of painful implementation will need to go into figuring out whether a usage might raise a pattern match failure -- to avoid crying wolf too often.

adamgundry:
@simonpj 

> * Why is the definition site flag called `-Wpartial-fields` and the use-site one called `-Wincomplete-record-selectors`? Better to have `-Wpartial-fields` and `-Wpartial-field-uses` or some consistent naming like that.  We could add `-Wpartial-field-bindings` as a synonym for `-Wpartial-fields.`

The naming was chosen for consistency with the existing `-Wincomplete-record-updates`. It's unfortunate that the existing `-Wpartial-fields` is inconsistent with this!

I'm not keen on `-Wpartial-field-uses` because it doesn't make clear that it refers to *selector* uses rather than updates. But I suppose we could call the new warning `-Wpartial-field-selectors`. Perhaps we should then also:
 * rename `-Wincomplete-record-updates` to `-Wpartial-field-updates`;
 * add `-Wpartial-field-bindings` with the effect of the current `-Wpartial-fields`;
 * change `-Wpartial-fields` to imply all of `-Wpartial-field-bindings`, `-Wpartial-field-selectors` and `-Wpartial-field-updates`?

> Mind you, now I think about it, the warnings for solving HasField constraints for GADTs should really be postponed until the end of typechecking, because we may not know the full type of the record until then. That's quite painful in implementation terms, and the benefit is small. Gah. One more effort-consuming corner. Maybe we should make the GADT bit fully optional too. That is: to be sure to avoid these warnings, just don't use partial fields at all.

Yes, I worried about this, which is why the proposal is worded to make this aspect implementation-defined...

> (I'm not sure if we have a view about "best-efforts" warnings, where warnings might come and go according to how best efforts work out. I could see uses being unhappy about that.)

... but I take this point, and if we are reluctant to have implementation-defined warnings, I'd be inclined to specify that the warnings are always emitted at use sites, which is simple and predictable.

--- 

@AntC2 

>> giving field names to heterogeneous multi-constructor datatypes, ... these are sometimes useful,
>
> "sometimes useful" doesn't sound like a strong motivation. Do we have metrics on how common are multi-constructor datatypes that use field names?

I have no metrics, but I have used such types myself in real-world commercial code. I don't think this pattern is necessarily that common in general libraries, but it is useful in domain-specific contexts. For example I used this when defining a large sum type representing the (many) email notifications that can be sent by a web application, like this:
```hs
data Notification
    = UserCreated { username :: Username }
    | PasswordReset { username :: Username, token :: ResetToken }
    | ...
```
The real thing has many constructors, with various different collections of fields, so it would be a bit painful to define a new record datatype for each separate constructor. But naming the fields is helpful, and by always pattern-matching on this datatype (never using the selector functions) we avoid partiality.

parsonsmatt:
> For ``HasField``, it would be possible to change its definition so that it would
not be solved at all for partial fields, or provide an alternative
implementation (either manually or automatically) returning a ``Maybe`` value.
This would avoid partiality when using ``OverloadedRecordDot``, without a need
for warnings.  It seems simplest to keep ``HasField`` consistent with existing
selector functions, however.

From my perspective, this alternatve seems like a clearly better option. Can still use the record selector without a warning, or you can `case` on the constructor with a `NamedFieldPun` to get it directly:

```haskell
data T = A { x :: Int } | B

foo :: T -> Maybe Int
foo t = t.x

bar :: T -> Int
bar t =
    case t of
        A { x } -> x
        B -> 0 
```

Alternatively, we could even *not* generate partial field selectors in the first place. So there just isn't a selector function `x` *or* an `instance HasField "x" A _`.

And a final alternative would be *customizable* field selection - `NoPartialFields` that wouldn't do the `HasField` *or* selector functions, `MaybePartialFields` which makes `HasField sym rec (Maybe ty)`, and I guess `PartialFields` is the current behavior.

This seems like a reasonable follow-up proposal, though, and is somewhat orthoganal to this proposal.

adamgundry:
@parsonsmatt thanks for writing up #535! Regardless of what is decided there, though, I think it makes sense to implement this proposal anyway, because there will inevitably be datatype definition sites that do not use the new features proposed in #535 at first, if at all, so it remains helpful to have `-Wincomplete-record-selectors` at use sites.

@gridaphobe thank you for recommending acceptance. I'd be grateful if the committee could reach consensus on the open points, then I'd be happy to amend the proposal accordingly. I think the main outstanding issues are:
 * Naming of the flag (see https://github.com/ghc-proposals/ghc-proposals/pull/516#issuecomment-1225921868 for my position on this, but I don't feel that strongly if the committee prefers an alternative);
 * Whether it is acceptable to have under-specified corner cases in which warnings may or may not be suppressed, or whether the proposal should err on the side of simplicity of specification and hence redundant warnings .

parsonsmatt:
> Regardless of what is decided there, though, I think it makes sense to implement this proposal anyway, because there will inevitably be datatype definition sites that do not use the new features proposed in https://github.com/ghc-proposals/ghc-proposals/pull/535 at first, if at all, so it remains helpful to have -Wincomplete-record-selectors at use sites.

Perfectly agree - warnings for partial accessors is a big improvement over the status quo, regardless of whether or not any alternatives come into play. Especially if you're using a datatype defined before `MaybePartialFields` was even available (ie all of them, so far). 😄

gridaphobe:
@adamgundry @simonpj 

> The naming was chosen for consistency with the existing -Wincomplete-record-updates. It's unfortunate that the existing -Wpartial-fields is inconsistent with this!

I agree with Adam. The "incomplete" terminology is also used for `-Wincomplete-patterns`. 

Why don't we just rename `-Wpartial-fields` to `-Wincomplete-record-definitions`? That would give us

* `-Wincomplete-record-definitions` to warn at the definition of a partial field
* `-Wincomplete-record-selectors` to warn at the use of a partial field as a selector
* `-Wincomplete-record-updates` to warn at the use of a partial field in a record update

gridaphobe:
@adamgundry - Simon has suggested on the list that `-Wincomplete-record-selectors` and `-Wincomplete-record-updates` (whatever the final names are) should be enabled by `-Wall`. Your proposal does not add `-Wincomplete-record-selectors` to `-Wall`, but I think that's just because of the precedent set by `-Wincomplete-record-updates`.

Would you object to adding both of those warnings to `-Wall`? I think it's sensible (but I agree with Simon that partial field **definitions** should not be warned about).

adamgundry:
Thanks @gridaphobe!

> Why don't we just rename `-Wpartial-fields` to `-Wincomplete-record-definitions`? That would give us
> 
> * `-Wincomplete-record-definitions` to warn at the definition of a partial field
> * `-Wincomplete-record-selectors` to warn at the use of a partial field as a selector
> * `-Wincomplete-record-updates` to warn at the use of a partial field in a record update

This sounds good to me. I presume we would retain `-Wpartial-fields` as a synonym for backwards compatibility. As we already have `-Wincomplete-patterns` and `-Wincomplete-uni-patterns` it seems more consistent to use the `incomplete` prefix than `partial` (the only other warning with that prefix is `-Wpartial-type-signatures` which is a rather different meaning of "partial").

> Simon has suggested on the list that `-Wincomplete-record-selectors` and `-Wincomplete-record-updates` (whatever the final names are) should be enabled by `-Wall`. Your proposal does not add `-Wincomplete-record-selectors` to `-Wall`, but I think that's just because of the precedent set by `-Wincomplete-record-updates`.
> 
> Would you object to adding both of those warnings to `-Wall`? I think it's sensible (but I agree with Simon that partial field **definitions** should not be warned about).

Yes, it was based on precedent, but in fact, it turns out I was wrong: #71 already added `-Wincomplete-record-updates` to `-Wall` as of GHC 9.2. Apologies that I missed this previously (presumably I was testing with an earlier version of GHC). Thus I agree we should have `-Wincomplete-record-selectors` in `-Wall` as well.

This does mean a breaking change for code using `-Wall -Werror`, but I think that is justifiable?

gridaphobe:
> This does mean a breaking change for code using `-Wall -Werror`, but I think that is justifiable?

Yes, I've always assumed that default warnings may get stricter with new compiler releases. I think the common guidance is to use `-Wall -Werror` for the main release build, but to have a "GHC Head" build that is warnings-only to get early warning of new/stricter warnings.

adamgundry:
I've changed the proposal to make the warning part of `-Wall`. However on further reflection I'm not sure we should rename `-Wpartial-fields`, at least not without separate discussion, so I've added a rationale to the Alternatives.

gridaphobe:
Thanks @adamgundry, this looks good to me! I sent it back to the committee for a final look.

adamgundry:
It seems this proposal has been accepted and is on `master` (https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0516-incomplete-record-selectors.rst), but this PR is still open. I'm not sure why it wasn't closed as merged?

nomeata:
Odd. Looks like I accidentially rebased the commits and thus github didn’t noticed that they are on master already? But as you say, I merged this with https://github.com/ghc-proposals/ghc-proposals/commit/baf2f191c090b29727258bc1399797bb954eda15.

I think we have to just close the PR, and live with the slightly confusing status. Thanks for noticing.

adamgundry:
It's [currently in GHC HEAD](https://gitlab.haskell.org/ghc/ghc/-/issues/18650#note_524746) but not yet released. As the [proposal document on `master`](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0516-incomplete-record-selectors.rst) says, it is slated for inclusion in 9.10.1, for which the alpha is expected soon (so the final release should be in a few months).

PR #517, opened 2022-06-13
Meta-proposal: Require implementors before proposal submission (under review)

nomeata:
This is a proposal to refine our proposal process, which came out of a discussion with @adamgundry at ZuriHac.

Currently we are happy to discuss and accept proposals that don't have an implementation plan, and we don't require proposal authors to do any implementation, to not raise the bar too high. But this leads to the odd situation that an ever growing list of proposals that are accepted, but nobody does anything about them.

So we wonder: is this useful? What is the value of proposing, discussing, refining and accepting a proposal if then nothing happens?

So here is a proposal to improve this situation: Before discussing a proposal, we require it to name someone (or somemany) who plan to implement it. Could be an established GHC developer, a new contributor, or even a paid consultant.

It does _not_ have to be the proposal author. But in that case, the proposal author is expected to rally support for the proposal and find someone able to implement who thinks it is good enough to join the effort.

Effects:

* Ideally, every accepted proposal has someone who will start implementing it in a timely manner.

* Since the first and easy step of implementing is to open a ticket in GHC's gitlab, accepted proposals will more likely show up in GHC’s issue tracker.

* The requirement to find an implementer (if the author’s does not do it themselves) means there is a higher incentive to beat the drum (mailing lists, twitter, forums…) which may lead to more general early community involvement, which would be great.

* This may attract new GHC developers.

* Of course, the implementors are still volunteers, who may lose interest, run out of time, or implementation may turn out to be too hard. That is ok. There is no repercussion for failing to implement a proposal you said you’d implement, no more than when you’d assign a GHC ticket to yourself and eventually unassign it.

* Some proposal authors may not be able to find an implementor, and then can’t be submitted to the committee. This may be a good thing: If there is not enough momentum behind it, it is probably not worth doing? And we'd spend our committee resources on the proposals that are more likely to actually go into GHC.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -60,8 +60,7 @@ This section outlines what stages a proposal may go through. The stage is identi
 
    `How to submit a proposal <#how-to-start-a-new-proposal>`_
 
-3. (No label.)  The wider community discusses the proposal in the commit section of the pull
-   request, while the author refines the proposal. This phase lasts as long as necessary.
+3. (No label.) The wider community discusses the proposal in the commit section of the pull request, while the author refines the proposal. If the author does not plan to implement the proposal themselves, this is also the phase where they try to find someone motivated and able to do so. This phase lasts as long as necessary.
 
    `Discussion goals <#discussion-goals>`_ •
    `How to comment on a proposal <#how-to-comment-on-a-proposal>`_ •
@@ -84,7 +83,10 @@ This section outlines what stages a proposal may go through. The stage is identi
 
    Acceptance of the proposal implies that the implementation will be accepted
    into GHC provided it is well-engineered, well-documented, and does not
-   complicate the code-base too much.
+   complicate the code-base too much. Those mentioned in the proposal as likely
+   implementers should now open a ticekt in `GHC’s issue tracker
+   <https://gitlab.haskell.org/ghc/ghc/-/issues/new>`_, and update the proposal’s
+   metadata to point to that ticket.
 
    `≡ List of accepted proposals <https://github.com/ghc-proposals/ghc-proposals/tree/master/proposals>`_ •
    `≡ List of proposals being revised <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3A%22Needs+revision%22>`_ •
@@ -97,10 +99,9 @@ This section outlines what stages a proposal may go through. The stage is identi
    `≡ List of dormant proposals <https://github.com/ghc-proposals/ghc-proposals/pulls?q=is%3Apr+label%3A%22dormant%22>`_
 
 
-8. Label: `Implemented <https://github.com/ghc-proposals/ghc-proposals/pulls?q=is%3Apr+label%3A%22Implemented%22>`_.   Once a proposal is accepted, it still has to be implemented.  The author
-   may do that, or someone else. We mark the proposal as “implemented” once it
-   hits GHC’s ``master`` branch (and we are happy to be nudged to do so by
-   email, GitHub issue, or a comment on the relevant pull request).
+8. Label: `Implemented <https://github.com/ghc-proposals/ghc-proposals/pulls?q=is%3Apr+label%3A%22Implemented%22>`_.
+
+   Once the implemenation hits GHC’s ``master`` branch, the implementors should update the metadata to indicate the version of GHC that will ship with this feature, and we label the PR as “implemented”.
 
    `≡ List of proposals pending implementation <https://github.com/ghc-proposals/ghc-proposals/pulls?q=is%3Apr+label%3A%22Accepted%22+-label%3A%22Implemented%22>`_ •
    `≡ List of implemented proposals <https://github.com/ghc-proposals/ghc-proposals/pulls?q=is%3Apr+label%3A%22Implemented%22>`_
--- a/README.rst
+++ b/README.rst
@@ -84,7 +84,7 @@ This section outlines what stages a proposal may go through. The stage is identi
    Acceptance of the proposal implies that the implementation will be accepted
    into GHC provided it is well-engineered, well-documented, and does not
    complicate the code-base too much. Those mentioned in the proposal as likely
-   implementers should now open a ticekt in `GHC’s issue tracker
+   implementers should now open a ticket in `GHC’s issue tracker
    <https://gitlab.haskell.org/ghc/ghc/-/issues/new>`_, and update the proposal’s
    metadata to point to that ticket.
 
--- a/README.rst
+++ b/README.rst
@@ -60,7 +60,7 @@ This section outlines what stages a proposal may go through. The stage is identi
 
    `How to submit a proposal <#how-to-start-a-new-proposal>`_
 
-3. (No label.) The wider community discusses the proposal in the commit section of the pull request, while the author refines the proposal. If the author does not plan to implement the proposal themselves, this is also the phase where they try to find someone motivated and able to do so. This phase lasts as long as necessary.
+3. (No label.) The wider community discusses the proposal in the conversation section of the pull request, while the author refines the proposal. If the author does not plan to implement the proposal themselves, this is also the phase where they try to find someone motivated and able to do so. This phase lasts as long as necessary.
 
    `Discussion goals <#discussion-goals>`_ •
    `How to comment on a proposal <#how-to-comment-on-a-proposal>`_ •
--- a/acceptance.rst
+++ b/acceptance.rst
@@ -30,7 +30,7 @@ Before accepting, the shepherd should check for the following
 3. Is the proposal sufficiently illustrated with examples? Examples help
    us understand the details that will be needed during implementation.
 
-4. Does the proposal indicate someone plans to implement it.
+4. Does the proposal indicate someone plans to implement it?
 
 Acceptance steps
 ----------------
--- a/acceptance.rst
+++ b/acceptance.rst
@@ -30,6 +30,8 @@ Before accepting, the shepherd should check for the following
 3. Is the proposal sufficiently illustrated with examples? Examples help
    us understand the details that will be needed during implementation.
 
+4. Does the proposal indicate someone plans to implement it.
+
 Acceptance steps
 ----------------
 
@@ -53,7 +55,7 @@ steps:
    where the word "proposal" links to the final rendered version, as found on https://github.com/ghc-proposals/ghc-proposals/tree/master/proposals
 
 4. If the PR title has "(under review)", remove it.
-   
+
 5. Set the PR to have the "Accepted" label.
 
 6. Comment on the PR that the proposal was accepted.
--- a/proposals/0000-template.md
+++ b/proposals/0000-template.md
@@ -108,6 +108,6 @@ the steering committee.
 
 ## Implementation Plan
 
-If accepted who will implement the change? If its is not the proposal authors,
-the proposal authors should someone who is motivated and able to work on an
-implementation of this proposal.
+If accepted who will implement the change? Before the proposal is submitted,
+the proposal authors should identify someone who is motivated and able to work on an
+implementation of this proposal. This may or may not be the proposal authors themselves.
--- a/proposals/0000-template.md
+++ b/proposals/0000-template.md
@@ -108,15 +108,6 @@ the steering committee.
 
 ## Implementation Plan
 
-(Optional) If accepted who will implement the change? Which other resources
-and prerequisites are required for implementation?
-
-## Endorsements
-
-(Optional) This section provides an opportunity for any third parties to express their
-support for the proposal, and to say why they would like to see it adopted.
-It is not mandatory for have any endorsements at all, but the more substantial
-the proposal is, the more desirable it is to offer evidence that there is
-significant demand from the community.  This section is one way to provide
-such evidence.
-
+If accepted who will implement the change? If its is not the proposal authors,
+the proposal authors should someone who is motivated and able to work on an
+implementation of this proposal.
--- a/proposals/0000-template.rst
+++ b/proposals/0000-template.rst
@@ -123,14 +123,6 @@ the steering committee.
 
 Implementation Plan
 -------------------
-(Optional) If accepted who will implement the change? Which other resources
-and prerequisites are required for implementation?
-
-Endorsements
--------------
-(Optional) This section provides an opportunity for any third parties to express their
-support for the proposal, and to say why they would like to see it adopted.
-It is not mandatory for have any endorsements at all, but the more substantial
-the proposal is, the more desirable it is to offer evidence that there is
-significant demand from the community.  This section is one way to provide
-such evidence.
+If accepted who will implement the change? If its is not the proposal authors,
+the proposal authors should someone who is motivated and able to work on an
+implementation of this proposal.


nomeata:
Oh, I also removed the “Endorsers” section. It was a nice idea, but doesn't seem to work in practice.

nomeata:
I think we are describing the same: garner interest among those willing and able to implement a change before submitting it here. At which point you go from a rough sketch to a full proposal is up to the author - the process only cares about the state of the proposal upon submission.

README.rst
```diff
@@ -84,7 +83,10 @@ This section outlines what stages a proposal may go through. The stage is identi
 
    Acceptance of the proposal implies that the implementation will be accepted
    into GHC provided it is well-engineered, well-documented, and does not
-   complicate the code-base too much.
+   complicate the code-base too much. Those mentioned in the proposal as likely
+   implementers should now open a ticekt in `GHC’s issue tracker
```
adamgundry, line 87:
```suggestion
   implementers should now open a ticket in `GHC’s issue tracker
```

proposals/0000-template.md
```diff
@@ -108,15 +108,6 @@ the steering committee.
...
-(Optional) This section provides an opportunity for any third parties to express their
-support for the proposal, and to say why they would like to see it adopted.
-It is not mandatory for have any endorsements at all, but the more substantial
-the proposal is, the more desirable it is to offer evidence that there is
-significant demand from the community.  This section is one way to provide
-such evidence.
-
+If accepted who will implement the change? If its is not the proposal authors,
+the proposal authors should someone who is motivated and able to work on an
+implementation of this proposal.
```
adamgundry, line 113:
```suggestion
If accepted who will implement the change? Before the proposal is submitted,
the proposal authors should identify someone who is motivated and able to work on an
implementation of this proposal. This may or may not be the proposal authors themselves.
```

simonpj:
I'm sympathetic to the idea of:

* Require a willing implementor before submitting to the committee

But let's make it very clear that this does *not* preclude making a PR for a proposal (no matter how sketchy), inviting discussion, refining it, etc; all without an implementor in sight,   It's just that before we put the proposal into a formal evaluation process, an implementor is required. 

Might not be suitable for all propsals; e.g. the one sketching the design direction for dependent Haskell, or articulating design principles.   But for most, requiring an implementor might
* Avoid accepted proposals languishing un-implemented forever
* Avoid consuming scarce committee time.  I don't think this is a big problem in practice, but it's a good principle that before asking busy volunteers to devote their most precious and unfungible resource (their time) to something, it would be helpful to have some expectation their efforts would have some tanglible outcome.

Since the implementation promise is (necessariliy) non-binding, and therefore could be lightly made, this change might be more of symbolic signpost than a hard hurdle to jump over; but no less valuable for that.

Ericson2314:
I understand the impetus not to accumulate a long backlog of miscellaneous pipe dreams, but a persistent problem in the Haskell ecosystem is that we are bad at doing long-term planing. Accumulating unimplemented proposals can result in a messy backlog, but it can also result in a long term roadmap and vision to rally around.

-----

This gets especially interesting as it relates to the Haskell Foundation.

I, for example, wrote https://discourse.haskell.org/t/pre-hftt-ongoing-focus-on-migration-tools/4626 in hopes of queueing up a number of items with a similar theme the HF might consider funding / making happen. I *also* have some yet-unpublished draft GHC proposals fleshing out features that I think would help with migrations stability. Currently in the "implementation plan" section, I write down what I just wrote above --- that I cannot afford to implement these myself but I would hope to bring them to the HF.

If this change as I understand it, there becomes a bit of a catch-22:
 - The steering committee won't want to approve anything which doesn't have an implementation plan
 - The foundation presumably won't want to fund any plan the steering committee doesn't approve.

The only solution to accepting and implementing such proposals thus to get GHC and HF representatives in a synchronous meeting and do both steps atomically, which IMO is a needless stumbling block the precludes momentum preliminarily accepted ideas building naturally.

-------

Turning attention from a possible future to the actual past, I have another example. I'll sheepishly admit my #285 is one of the "backlog" proposals with no implementation plan (well, other than me having once did some work with @RyanGlScott cleaning stuff up to lay the groundwork for such a feature). But that proposal has gone on to help motivate and influence #448, and also positively interacts with #509 (providing an alternative to fully qualifying everything to avoid capture via tighter-scoped, shadowing implicit bindings). Having written-down plans to refer to versus trying to keep the interaction between other proposals and would-be proposals in our heads seems good.

---------

Summing up, I think we do have a problem but it's not that we *have* a backlog at all, but that we have a *disorganized* backlog. I think the solution is once again something like the "Aspirational users guide" I proposed in https://github.com/ghc-proposals/ghc-proposals/pull/442.

A single document containing ramifications of accepted proposals ensures proposals are firstly consistent with another at all, but beyond that also that they interact well fulfilling a grander vision. And if there is a new proposal with a funding plan conflicting with a languishing proposal, we don't always have have to yield to the existing proposal. Having identified the conflict via the AUG, we can choose to accept the new proposal, or modify or even outright-revert the AUG changes corresponding to the languishing proposal instead.

In this manner, we can plan out ahead, ensure the plan is coherent, understand the difference between what's implemented and the plan, and finally still remain nimble modifying the plan as priorities and resources change. That seems much better to me than sticking with atomized proposals and merely narrowing our horizons to avoid the chaos of the existing process at scale.

I urge giving #442 another look --- unless I am mistaken the idea more stalled out than was rejected outright.

nomeata:
> The foundation presumably won't want to fund any plan the steering committee doesn't approve.

I’d expect that the foundation will happily say “we fund this, assuming it gets accepted“, which is certainly good enough to jump over the bar introduced by this proposal. Also, there is overlap between the people, and these rules are implemented by humans, so I wouldn’t worry too much on that front.

I am not sure a AUG (which I do see some merit in) would help on this front. In fact, it may be even be worse: Someone writes up their nice fancy vanity syntax proposal, it gets accepted, but nobody implements it. Currently, it means that the proposal just sits there and doesn't really hurt anybody. If we have the AUG, that changes would now be there, possibly interacting with other changes, making it possibly harder to move documented about implemented stuff into the real user guide. It seems that unimplemented proposals cause _more_ friction in the AUG world.

Here I am assuming that the motivation and decisions of implementors are not affected by whether accepted proposals are separate files or in one AUG. In an alternative world where the decisions of the committee would actually guide some development team that’s committed to implement all accepted proposals this would be different, and an AUG would work better there – because every discrepancy between the real UG and the AUG would be a TODO item on the GHC team. But that is not the case.

Ericson2314:
> I’d expect that the foundation will happily say “we fund this, assuming it gets accepted“, which is certainly good enough to jump over the bar introduced by this proposal. Also, there is overlap between the people, and these rules are implemented by humans, so I wouldn’t worry too much on that front.

Fair enough. We can do a such a "two phase commit" and that makes my prior concerns overblown.

> I am not sure a AUG (which I do see some merit in) would help on this front. In fact, it may be even be worse...

Well, what exactly is the problem we are trying to solve here?

1. Backlog of accepted & unimplemented proposals looks bad is confusing
2. Backlog of accepted & unimplemented proposals has unclear conflicts with new proposals
3. Deliberating proposals that won't be implemented wastes committee time
4. Something else?

AUG indeed increases the practical costs of accepted but unimplemented proposals, but that can be a *good* thing if the goal is to have less of them.

Instead of having a rule "someone agrees to implement" (which cannot be nor is intended to be enforced) there becomes the more open-ended trade off of "do we want to deal with any merge conflicts for the duration the proposal remains unimplemented?" which depends both on how long that duration is expected to be and how much the committee like the proposal.

In short, rather than revising the rules about whether proposals are eligible for acceptance in response to how the backlog is growing, the process can naturally impose a sort of "back pressure", automatically stabilizing itself.

> In an alternative world where the decisions of the committee would actually guide some development team that’s committed to implement all accepted proposals this would be different, and an AUG would work better there – because every discrepancy between the real UG and the AUG would be a TODO item on the GHC team. But that is not the case.

Well perhaps I am getting ahead to far ahead and it would be best to revisit this part later, but if the HF were to be move involved ensuring the general plans for our core infrastructure get executed, we could imagine instead of opting-*in* to submitting some accepted GHC proposals to the HF queue, that would indeed sort of be the default, and outside implementers are in essence "work stealing" from that queue in lieu of the HF popping the item from the front to execute itself.

In short, even if that is not the reality today as you say, I think a good mindset is operating as if it were the reality.

phadej:
I'd comment as a person which occasionally thinks about writing a proposal, but usually dismiss that idea as it is a massive time sink.

> Avoid consuming scarce committee time. I don't think this is a big problem in practice, but it's a good principle that before asking busy volunteers to devote their most precious and unfungible resource (their time) to something, it would be helpful to have some expectation their efforts would have some tanglible outcome.

Proposal *writers* are often also volunteers. Writing an initial pproposal and making revisions to it needs plenty of time too, So one outcome of this proposal might be that there will be just less proposals, and just issues in this repository or GHC issue tracker.

phadej:
Can @noemata or someone summarise which proposals don't have an implementor?

I suspect that "easy proposals" (like syntax changes, e.g. `\cases`) have implementors, but more tricky ones don't. I hope to be wrong with this

If there is such correlation between a type of proposal and it being implemented, then it is be a point against having the implementor requirement: if "add more stuff" is done, but "cleanup stuff" is not.

nomeata:
@phadej, everyone is a volunteer - authors, committee members, people implementing features. And it is already the case that the proposal proponents (i.e. typically the author) have to find someone to implement it - only that that now happens independent of and typically after the proposal process. So seems like a good thing if authors don't spend all that time on writing and revising before there is someone to implement it?

If the outcome is that there are less proposals and that ideas that sound nice, but are not compelling enough to attract implementor-volunteers, remain as issues rather than full time-sinking proposals, then that'd be a good outcome, wouldn't it?

nomeata:
Or let me frame it differently: One reason why the GHC Proposal Process isn’t a complete failure is that there is almost always someone responsible: First the author, then the secretary, then the shepherd, then the committee (but the shepherd ought to drive it). This way, we either have progress or someone to ~blame~ nudge.

But that property is no longer true once the proposal gets accepted: Then it just sits there, up for anyone to grab. What I’d like to see is a seamless transition of ownership that when a proposal is accepted, there will always a GHC ticket with an assignee, so that (unless someone loses interest/goes MIA etc. – the usual constraints in a volunteer-driven project) every accepted proposal will be implemented.

README.rst
```diff
@@ -60,8 +60,7 @@ This section outlines what stages a proposal may go through. The stage is identi
 
    `How to submit a proposal <#how-to-start-a-new-proposal>`_
 
-3. (No label.)  The wider community discusses the proposal in the commit section of the pull
-   request, while the author refines the proposal. This phase lasts as long as necessary.
+3. (No label.) The wider community discusses the proposal in the commit section of the pull request, while the author refines the proposal. If the author does not plan to implement the proposal themselves, this is also the phase where they try to find someone motivated and able to do so. This phase lasts as long as necessary.
```
goldfirere, line 63:
```suggestion
3. (No label.) The wider community discusses the proposal in the conversation section of the pull request, while the author refines the proposal. If the author does not plan to implement the proposal themselves, this is also the phase where they try to find someone motivated and able to do so. This phase lasts as long as necessary.
```

acceptance.rst
```diff
@@ -30,6 +30,8 @@ Before accepting, the shepherd should check for the following
 3. Is the proposal sufficiently illustrated with examples? Examples help
    us understand the details that will be needed during implementation.
 
+4. Does the proposal indicate someone plans to implement it.
```
goldfirere, line 33:
```suggestion
4. Does the proposal indicate someone plans to implement it?
```

And perhaps sympathetic changes to other nearby bullets-that-are-questions.

goldfirere (approved):
I think this is a nice step forward and will lead to fewer unhappy people.

To me, the critical problem this proposal addresses is that some folks write a proposal, invest significant time in seeing it through, it gets accepted, and then nothing happens. This creates a bad experience for those proposers, whose valuable time was unproductive. Other benefits (more effective time spent from the committee, smaller backlog) are, to me, secondary.

phadej:
@nomeata 

> But that property is no longer true once the proposal gets accepted: Then it just sits there, up for anyone to grab. 

But that is already true for many is-it-a-feature-or-a-bug issues on GHC tracker. There are issues with discussion ended with "this would require a GHC proposal".

Having half of ideas on GHC issue tracker and other half in this repository as issues will be worse than no. IMO this repository shouldn't have issues-to-be-proposals, all of that belongs to GHC issue tracker.

phadej:
Many proposals don't have `implemented` tag, e.g. https://github.com/ghc-proposals/ghc-proposals/pull/277 or https://github.com/ghc-proposals/ghc-proposals/pull/364 (I just looked at few).

phadej:
https://github.com/ghc-proposals/ghc-proposals/pull/371 and https://github.com/ghc-proposals/ghc-proposals/pull/229 are also implemented AFAIK.

phadej:
> Of course, the implementors are still volunteers, who may lose interest, run out of time, or implementation may turn out to be too hard. That is ok. There is no repercussion for failing to implement a proposal you said you’d implement, no more than when you’d assign a GHC ticket to yourself and eventually unassign it.

What about *Unsaturated Type Families*. This is feature very many (relatively speaking) are very excited about, and Csongor even had a PoC implemented. What's now?

But to be fair: Csongor only mentions in the proposal that there is (was?) a PoC, but he didn't say he'd make a proper patch. Would committee not even consider that proposal until Csongor found an actual implementor (him or someone else?).

I mean, many people are excited about that, but there are maybe five who can actually pull of the implementation!?

phadej:
There are also plenty of amendment proposals, which are tagged accepted, but not tagged implemented.

I think it would be beneficial to have a reverse link in changed proposal, so one can find an amendment, and then cross-check whether there is a missing tag (like *implemented*) or whether there its actually not implemented!

Also having an amendment tag might make things a bit clearer.

It looks like a lot more stuff is implemented that the quick tag-based scan shows right now.

nomeata:
> Also having an amendment tag might make things a bit clearer.

Good idea!

AntC2:
Can there be some way to record 'bright ideas' that can't find an implementor? To save people continually re-discovering (if not re-inventing) the wheel.

Case in point: #232 accepted three years ago, not implemented. Rediscovered #338 two years ago. Rediscovered #389 a year and a half ago. (And I wouldn't be surprised if there are tickets complaining about the disfluencies these proposals are aiming to fix.)

gbaz:
"github tickets as a union-find store: a natural experiment"

nomeata:
> Can there be some way to record 'bright ideas' that can't find an implementor?

I wish, but it seems to be a hard problem. Not even accepted proposals seem to work here. Maybe I am fatalistic here, but there will always be people who jump from idea to proposal rather quickly and miss prior discussion. In a way the present proposal helps a bit, too: it nudges people to talk more about their idea to attract implementors, and thus are more likely to be pointed to prior attempts earlier.

simonpj:
This propsal has been submitted to the commitee, and I am the shepherd.

On balance I recommend acceptance, with the these nuances clarified.  Requiring a willing implementor would avoid the committee spending precious cycles debating a proposal that no one is going to implement.

* An offer of implementation cannot be binding, so it is something of a soft constraint.  (An author could cynically volunteer themselves, without having any intention of carrying through, but we expect better of the Haskell community.)
* We should stress that nothing stops someone creating a proposal, making a PR, and debating it with the community, all without an implementor.  Only when it is *submitted to the committee* for review and approval is an implementor required.
* I wonder if a proposal that is accepted but not implemented (for whatever reason) should be un-accepted after, say, a year.  That would provide some incentive to get on with it; and the language context might be different by then.

Joachim suggests that this replaces the (never used) "Endorsements" section.  That seems sensible.

Only a few people have contributed above. I'd love to hear from more.

AntC2:
> @goldfirere the critical problem this proposal addresses is that some folks write a proposal, invest significant time in seeing it through, it gets accepted, and then nothing happens.

> @simonpj  wonder if a proposal that is accepted but not implemented (for whatever reason) should be un-accepted after, say, a year.

> Changes approved

@NiklausWirth "In the interest of increased quality of software products, we may be well advised to get rid of many facilities of modern, baroque programming languages that are widely advertised in the name of 'user-orientation,' 'scientific sophistication,' and 'progress.'" [ref](https://history.computer.org/pioneers/wirth.html)

Wirth's comment is quoted as re Pascal, but my memory is it was re Modula, and his motivation for introducing Oberon as a simplification. (He'd already resigned from the ALGOL 68 Language Committee on grounds it was heading in a direction he didn't want the ALGOL name to go.) 

In my view, GHC several years ago reached the level of 'Baroque'. My recent proposals/commentary on proposals -- including one "that is accepted [as of three years ago] but not implemented" have been aimed at preventing the Baroque/retaining the elegance which attracted me to Haskell in the first place.

Clearly I'm failing. This is my stop; I'm getting off the train. I don't want to program in today's GHC, and this thread gives me every expectation GHC is going to get worse.

AntC2:
(I see a trickle more discussion on the SC mailing list.)

> @aspiwack ... I also don't feel that the number of unimplemented proposals is so high as to be a problem.
>

I think that as well as the accepted-never-implemented proposals on github you might try to get a handle on the number of 'proto-proposals' that are lurking as Discussions/Issues/gitlab drafts, etc. Some of them are authored by people who are the likely implementors -- and that's fine. Others realistically don't stand a hope -- and the authors could have their attention gently drawn to this meta-requirement.

Speaking as someone who's given up putting effort into proposals after (as it turns out) wasting a huge amount of time (and wasting huge amounts of others' time I guess), I'd appreciate the Committee giving a firmer statement here. GHC is not going anywhere that helps me; and with the benefit of hindsight I can see it was never going to. I wish I'd known that (say) five years ago. Because I could have put those efforts into a different Haskell -- as I'm now doing.

nomeata:
We decided against this, for lack of consensus and clear evidence that we even have a problem.

arybczak:
> lack of consensus and clear evidence that we even have a problem.

This is interesting, because exhibit A is the modifier proposal, which still hasn't been implemented and multiple proposals were abandoned/stalled by YEARS because "we shouldn't use pragmas, we should use modifiers" and modifiers are still not implemented in GHC.

- https://github.com/ghc-proposals/ghc-proposals/pull/374
- https://github.com/ghc-proposals/ghc-proposals/pull/390
- https://github.com/ghc-proposals/ghc-proposals/pull/512

PR #518, opened 2022-06-13
Type vs Constraint proposal (under review)

simonpj:
This GHC proposal refines the type structure of `Type` and `Constraint`, to 
* allow unboxed and unlifed constraints
* solve a long-standing source of (still-unsolved) bugs in GHC

[Rendered proposal here](https://github.com/ghc-proposals/ghc-proposals/blob/spj/type-vs-constraint/proposals/0000-type-vs-constraint.rst)

### DIFF
--- a/proposals/0000-type-vs-constraint.rst
+++ b/proposals/0000-type-vs-constraint.rst
@@ -273,8 +273,11 @@ Accordingly:
   ``Data.Kind`` should have a stable API; the kinds of these type
   constructors will not change.
 
-* ``GHC.Prim`` exports ``SORT``, ``TypeOrConstraint``, ``IP``.
-  Users may import them from ``GHC.Prim``, but they should not complain if they change in future.
+* A module within ``ghc-prim`` exports: ``CONSTRAINT``, ``(=>)``, ``(==>)``, ``SORT``, ``TypeOrConstraint(..)``,
+  and ``IP``. If `#524`_ is accepted and implemented, importing the module from ``ghc-prim``
+  will require the user to enable ``-XUnstable``.
+
+.. _`#524`: https://github.com/ghc-proposals/ghc-proposals/pull/524
 
 Implementation notes
 :::::::::::::::::::::::::::::::::
--- a/proposals/0000-type-vs-constraint.rst
+++ b/proposals/0000-type-vs-constraint.rst
@@ -249,7 +249,7 @@ making them apart is unsound in the presence of the current ``newtype`` optimiza
 one-element classes.
 Accordingly, under this proposal,
 
-* ``TypeLike`` and ``ConstraintLike`` will be considered not *apart*.
+* ``TYPE`` and ``CONSTRAINT`` will be considered not *apart*.
 
 As a consequence, ``Type`` and ``Constraint`` are also not *apart*, just as today.
 This a wart, but it is an *existing* wart, and one that is not easy to fix.
@@ -258,9 +258,9 @@ As before, nothing prevents writing instances like::
 
   instance C (Proxy @Type a) where ...
 
-In particular, ``SORT`` and ``TypeLike`` and ``ConstraintLike`` (and the synonyms ``TYPE``, ``CONSTRAINT`` etc)
+In particular, ``TYPE``, ``CONSTRAINT``, ``Type`` and ``Constraint``
 are all allowed in instance heads. It's just that
-``TypeLike`` is not apart from ``ConstraintLike``
+``TYPE`` is not apart from ``CONSTRAINT``
 so that instance would irretrievably overlap with::
 
   instance C (Proxy @Constraint a) where ...
@@ -273,8 +273,7 @@ Future stability
 In the past it has not been very clear which parts of GHC's API are stable and which
 are unstable:
 
-* By "stable" we mean that efforts will be made to avoid change, and any
-changes should require a GHC proposal.
+* By "stable" we mean that efforts will be made to avoid change, and any changes should require a GHC proposal.
 
 * By "unstable" we mean that the API should be considered part of GHC's internal
   implementation.  Changes may be made to the unstable API without a proposal.
@@ -290,15 +289,14 @@ this proposal makes a modest start on clarifying the distinction.  In particular
 
 * The unstable API includes:
 
-  * The new type constructors ``CONSTRAINT``, ``(=>)``, ``(==>)``, and ``(-=>)`` are
+  * The new type constructors ``CONSTRAINT``, ``(=>)``, ``(==>)``, and ``(-=>)``;
     exported by ``GHC.Types``.
 
-  * The existing type constructors ``FUN`` and ``IP``, also exported by ``GHC.Types``.
+  * The existing type constructors ``FUN`` and ``IP``; also exported by ``GHC.Types``.
 
 * The stable API includes:
 
-  * ``Symbol``, ``Type``, ``TYPE``, ``Constraint``,
-  ``RuntimeRep``, ``Multiplicity``, ``Levity``, and ``(->)``; all exported by ``Data.Kind``
+  * ``Symbol``, ``Type``, ``TYPE``, ``Constraint``, ``RuntimeRep``, ``Multiplicity``, ``Levity``, and ``(->)``; all exported by ``Data.Kind``
 
 We keep ``CONSTRAINT`` in the unstable API for now, exposing it only though the possiblity
 of having unlifted implicit paramters.
--- a/proposals/0000-type-vs-constraint.rst
+++ b/proposals/0000-type-vs-constraint.rst
@@ -1,4 +1,4 @@
-Separating Type from Constraint
+Generalising Type and Constraint
 ==============================
 
 .. author:: Richard Eisenberg and Simon Peyton Jones
@@ -128,9 +128,38 @@ produce a coercion between ``Int`` and ``Bool``, thus::
 That would be Very, Very Bad.  So, although ``Type`` and ``Constraint`` are built
 with different (un-equal) primitive type constructors,
 
-* **GHC's type checker treats `Type` and `Constraint` as *not apart*.**
+* **GHC's type checker treats `Type` and `Constraint` as not "apart".**
 
 That in turn makes GHC complain that the above instances overlap, and are hence illegal.
+You can read more about what "apart" means in
+`Closed type families with overlapping equations <https://simon.peytonjones.org/closed-type-families/>`_,
+sections 3.2 and 3.3.
+
+Apartness affects type-class instances as well as type-family instances.
+Suppose we have::
+
+  instance {-# OVERLAPPABLE #-} C Int a  where ...
+  instance {-# OVERLAPPING #-}  C Int Bool where ...
+
+and we are trying to solve the constraint ``C Int (F gamma)``, where ``F`` is a type family and ``gamma``
+is an as-yet-unknown unification variable. Since ``(F gamma)`` is not
+apart from ``Bool``, it could be that (once we know ``gamma``) the second
+instance should be chosen. So GHC declines to commit to the first.
+
+If ``(F gamma)`` later simplifes to, say ``Char`` (which *is* apart from ``Bool``),
+then and only then GHC can commit to the first instance.
+
+Similarly with ``Type`` and ``Constraint``. Suppose we have::
+
+  instance D Type where...
+  instance D Constraint where ...
+
+and try to solve ``C Int Type``. This matches the first instance, but
+since ``Type`` is not apart from ``Constraint`` GHC thinks "oh, the
+second one could match in the future" and declines to commit to the
+first.  In fact, you can have either of these instances separately,
+but if they occur together neither will ever be chosen: they overlap
+irretrievably.
 
 
 Proposed Change Specification
@@ -220,7 +249,7 @@ Accordingly, under this proposal,
 As a consequence, ``Type`` and ``Constraint`` are also not *apart*, just as today.
 This a wart, but it is an *existing* wart, and one that is not easy to fix.
 
-Note that nothing prevents writing instances like::
+As before, nothing prevents writing instances like::
 
   instance C (Proxy @Type a) where ...
 
@@ -231,7 +260,7 @@ so that instance would irretrievably overlap with::
 
   instance C (Proxy @Constraint a) where ...
 
-But this is just the status quo; it is not a change.
+But this is just the status quo; it is not a change (see Sectionn 2.2).
 
 Future stability
 :::::::::::::::::::::::::::::::::
@@ -309,7 +338,7 @@ Indeed that was our first plan. But
 
 * Collapsing the two into one SORT witnesses the idea that types and constraints are the same thing at runtime.
 
-* We anticipate further movement in this area, perhpas via "Kinds are Calling Conventions". If so, ``SORT`` gives us wiggle room to do that; and (even more important) we don't want to duplicate any such changes across two distinct type constructors, ``SORT`` and ``CONSTRAINT``.
+* We anticipate further movement in this area, perhaps via "Kinds are Calling Conventions". If so, ``SORT`` gives us wiggle room to do that; and (even more important) we don't want to duplicate any such changes across two distinct type constructors, ``TYPE`` and ``CONSTRAINT``.
 
 So the balance came down pretty firmly in favour of the design we offer.
 
--- a/proposals/0000-type-vs-constraint.rst
+++ b/proposals/0000-type-vs-constraint.rst
@@ -220,6 +220,19 @@ Accordingly, under this proposal,
 As a consequence, ``Type`` and ``Constraint`` are also not *apart*, just as today.
 This a wart, but it is an *existing* wart, and one that is not easy to fix.
 
+Note that nothing prevents writing instances like::
+
+  instance C (Proxy @Type a) where ...
+
+In particular, ``SORT`` and ``TypeLike`` and ``ConstraintLike`` (and the synonyms ``TYPE``, ``CONSTRAINT`` etc)
+are all allowed in instance heads. It's just that
+``TypeLike`` is not apart from ``ConstraintLike``
+so that instance would irretrievably overlap with::
+
+  instance C (Proxy @Constraint a) where ...
+
+But this is just the status quo; it is not a change.
+
 Future stability
 :::::::::::::::::::::::::::::::::
 
@@ -277,16 +290,65 @@ Costs and Drawbacks
 
 Alternatives
 ------------
-* This seems to be the best way to support unlifted and unboxed constraints.
 
-* We have another putative fix to the type-checker bugs, but it would take a
-  major sea-change to the compiler, requiring a full separation between Haskell
-  types and Core types.
-  (Currently, GHC uses the same representation for both,
-  a considerable simplification, as only one type needs to be e.g. written to
-  interface files.) This other fix would likely require several solid weeks
-  of implementation work, and would have consequences (both implementation and end-user) that are hard to anticipate.
-  In constrast, the one presented here is simple, and we have a clear grasp of its consequences.
+Two different type constructors
+::::::::::::::::::::::::::::::::::
+
+We considered having two distinct primitive type constructors, ``TYPE`` exactly as now, and ``Constraint :: Type``.
+Indeed that was our first plan. But
+
+* In Core we would have to say "If ``e :: ty :: ki`` then ``ki`` must be ``TYPE rr`` or ``Constraint``.
+  Similarly for the types of binders. That "or" isn't fatal, but it's inelegant.
+
+* We anticipate that it will not be long before people want unlifted constraints. So then we'd add ``CONSTRAINT``, thus::
+
+    TYPE :: RuntimeRep -> Type
+    CONSTRAINT :: RuntimeRep -> Type
+
+  And now we have something isomorphic to what we propose (two types, vs one with a flag), except that our proposal allows all value-level terms with runtime rep rr to be treated uniformly. E.g. ``isUnliftedType`` has one case rather than two.
+
+* Collapsing the two into one SORT is witnessing the idea that types and constraints are the same thing at runtime.
+
+* We anticipate further movement in this area, perhpas via "Kinds are Calling Conventions". If so, ``SORT`` gives us wiggle room to do that; and (even more important) we don't want to duplicate any such changes across two distinct type constructors, ``SORT`` and ``CONSTRAINT``.
+
+So the balance came down pretty firmly in favour of the design we offer.
+
+Fully separate Haskell types from Core types
+::::::::::::::::::::::::::::::::::::::::::::
+
+An alternative approach would be to fully separate Haskell types from Core types,
+where ``Type`` and ``Constraint`` are distinct in the former but the same in the latter.
+(Currently, GHC uses the same representation for both,
+a considerable simplification, as only one type needs to be e.g. written to
+interface files.)
+This would be a major sea-change to the compiler, requiring weeks of effort, code duplication,
+and knock-on effects (both implementation-wise and end-user-visible) that are hard to predict and might be unwelcom.
+
+In constrast, the one presented here is simple, and we have a clear grasp of its consequences.
+
+Tuples
+:::::::::::::
+
+This proposal does not address the problem of tuples, although is tempting to do so.  In GHC today we have::
+
+      (,)   :: Type -> Type -> Type
+      (%,%) :: Constraint -> Constraint -> Constraint
+
+except that the latter is written, in source Haskell, as ``(c1,c2)``.  That leads to syntactic ambiguity. If I write::
+
+      type S a b a = ((a,b), c)
+
+do I intend to define a Constraint synonym, or a Type synonym?  It's tempting to give (,) a polymorphic type::
+
+      (,) :: forall (v:TypeOrConstraint). SORT v LiftedRep -> SORT v LiftedRep -> SORT v LiftedRep
+
+But, tempting as it is, we do not propose this.  The tuples problem
+is all about concrete syntax. It just so happens that we use the
+same parentheses and commas for normal tuples and constraint
+tuples. No one has asked for polymorphism here. And it seems
+overwrought to complexify our type system just to allow for this
+little concrete-syntax convenience. Letting tuple syntax drive the
+type system is the tail wagging the dog.  We leave this for the future.
 
 Unresolved Questions
 --------------------
--- a/proposals/0000-type-vs-constraint.rst
+++ b/proposals/0000-type-vs-constraint.rst
@@ -1,0 +1,280 @@
+Separating Type from Constraint
+==============================
+
+tmp/Unlifted Constraints
+====================
+
+.. author:: Richard Eisenberg and Simon Peyton Jones
+.. date-accepted:: 
+.. ticket-url:: 
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. contents::
+
+Since version 8.0, GHC has supported a rich structure of inhabited types through the
+new primitive ``TYPE``, allowing
+unlifted types to be treated as first-class citizens, leading to the possibility
+of unlifted datatypes and newtypes, already available.
+
+This proposal extends this treatment to constraints, whose inhabitants
+can usefully be viewed as implicit parameters (typically type class dictionaries).
+Just like how the existence of ``TYPE`` directly opened the door to new innovations
+in datatype and newtype definitions, we expect the similar treatment of constraints
+to allow for strict classes and unlifted and unboxed implicit parameters.
+
+In addition, the new treatment proposes here allows GHC's internal language, Core,
+to more properly tell the difference between types and constraints, eliminating a
+troubling class of bugs that keeps recurring.
+
+Motivation
+----------
+
+1. With this proposal, we immediately allow unlifted and unboxed implicit parameters
+   using the existing ``(?x :: Int#)`` syntax. This allows programs to abstract over
+   implicit parameters without suffering inefficiencies due to boxing.
+
+2. This proposal allows a fix to several type-checker bugs that is otherwise elusive.
+   With this proposal, we can eliminate this entire class of bugs, that we otherwise
+   believe will continue to haunt GHC.  Currently in GHC, `Type` and `Constraint` are
+   *distinct* in the typechecker, but *identical* in Core.  That leads to massive duplication;
+   many, many functions over types have two versions, one for the typechecker and one for Core.
+   It also leads to intellectual complexity, with two subtly-different forms of type equality.
+   Worse, nothing maintains the separation, which leads to those elusive bugs.
+
+3. Although we do not propose unlifted constraints in this proposal, we believe that
+   this paves the way toward new innovations in constraint representations.
+
+Background
+----------
+
+This section describes key aspects of the status quo.
+
+The current type structure
+:::::::::::::::::::::::::::
+
+The current root of our type system contains these definitions::
+
+  -- Primitive type constructors
+  type TYPE       :: RuntimeRep -> Type
+  type Constraint :: Type    -- The kind of constraints
+  type Symbol     :: Type    -- The kind of compile-time strings
+  type IP         :: Symbol -> Type -> Constraint   -- Implicit parameters
+
+  type FUN :: forall (m :: Multiplicity) ->
+              forall (r1 :: RuntimeRep) (r2 :: RuntimeRep).
+              TYPE r1 -> TYPE r2 -> Type
+  -- (=>) is not something that can be written unsaturated;
+  --      it's just some concrete syntax
+
+  -- Data type declarations, used only at the type level
+  data Multiplicity = Many | One
+  data Levity       = Lifted | Unlifted
+  data RuntimeRep   = BoxedRep Levity | IntRep | FloatRep | ...
+
+  -- Type synonyms
+  type LiftedRep   = BoxedRep Lifted
+  type UnliftedRep = BoxedRep Unlifted
+  type Type        = TYPE LiftedRep
+
+NB: in GHC, implicit parameters are internally represented as a special class,
+but that is not user-visible.
+
+Type and Constraint are not apart
+:::::::::::::::::::::::::::::::::::
+
+GHC has an optimization for one-element classes (where the element
+is either a superclass or a method), defining these in like a newtype, not a datatype.
+For example, if we have ::
+
+  class Default a where
+    def :: a
+
+the Core of the program will have a definition like ::
+
+  newtype Default a = MkDefault a
+
+In turn, this newtype gives rise to an axiom (coercion), like so::
+
+  axDefault :: Default a ~R# a
+
+where ``~R#`` represents primitive representational equality. Note that
+``axDefault`` is *heterogeneous*: the kind of ``Default a`` is ``Constraint``,
+whereas the kind of ``a`` is ``Type``.
+
+GHC allows us to extract out an equality relationship between *kinds* from an
+equality relationship on *types* -- and kind equalities are always nominal. To
+wit, Core allows ::
+
+  KindCo axDefault :: Constraint ~# Type
+
+For this reason, the type system must treat ``Constraint`` and ``Type`` as
+not *apart*. For this reason, the following is not allowed::
+
+  type family F a
+  type instance F Type = Int
+  type instance F Constraint = Bool
+
+If these instances were allowed, GHC could
+produce a coercion between ``Int`` and ``Bool``, thus::
+
+  Bool  ~#  F Constraint   -- By type instance F Constraint (backwards)
+        ~#  F Type         -- By KindCo axDefault
+        ~#  Int            -- By type instance F Type
+
+That would be Very, Very Bad.  So, although `Type` and `Constraint` are built
+with different (un-equal) primitive type constructors, GHC's type checker nevertheless
+treats them as *not apart*; that in turn makes GHC complain that the above
+instances overlap, and are hence illegal.
+
+
+Proposed Change Specification
+-----------------------------
+
+We propose the following new setup, not repeating any types that remains unchanged::
+
+  -- Primitive type constructors
+  type SORT :: TypeOrConstraint -> RuntimeRep -> Type
+  type IP   :: formal (r :: RuntimeRep). Symbol -> TYPE r -> CONSTRAINT r
+
+  -- Data types
+  data TypeOrConstraint = TypeLike | ConstraintLike
+
+  -- Synonyms
+  type TYPE       = SORT TypeLike
+  type CONSTRAINT = SORT ConstraintLike
+  type Constraint = CONSTRAINT LiftedRep
+
+  type (=>)  :: forall (r1 :: RuntimeRep) (r2 :: RuntimeRep).
+                CONSTRAINT r1 -> TYPE r2 -> Type  -- primitive
+  type (==>) :: forall (r1 :: RuntimeRep) (r2 :: RuntimeRep).
+                CONSTRAINT r1 -> CONSTRAINT r2 -> Constraint
+
+  -- Implicit parameters
+  class IP name ty | name -> ty where ...  -- Special case to allow with unusual kind
+
+Changes to the type structure
+:::::::::::::::::::::::::::::
+
+This proposal introduces ``(=>)`` and ``(==>)`` as proper type constructors, just like
+any other. Just like ``(->)``, they have kinds and can be abstracted over.
+Unlike ``FUN``, they do not take a ``Multiplicity`` argument; implicity, it is ``Many``.
+
+The ``==>`` arrow is used in two places:
+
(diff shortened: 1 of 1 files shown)


proposals/0000-type-vs-constraint.rst
```diff
@@ -0,0 +1,284 @@
...
+
+Type and Constraint are not apart
+:::::::::::::::::::::::::::::::::
+
+It remains the case that ``Type`` must not be apart from ``Constraint``, because
+making them apart is unsound in the presence of the current ``newtype`` optimization for
+one-element classes.
+Accordingly, under this proposal,
+
+  * ``TypeLike`` and ``ConstraintLike`` will be considered not *apart*.
```
adamgundry, line 200:
```suggestion
* ``TypeLike`` and ``ConstraintLike`` will be considered not *apart*.
```

proposals/0000-type-vs-constraint.rst
```diff
@@ -0,0 +1,284 @@
...
+Unlike ``FUN``, they do not take a ``Multiplicity`` argument; implicity, it is ``Many``.
+
+The ``==>`` arrow is used in two places:
+
+* In instance heads, like ``instance Eq a => Eq (Maybe a)``
+* In quantified constraints, like ``forall x. c x => Eq x``
+
+In order to be backward compatible,
+we allow writing ``=>`` instead of ``==>`` in instance heads and in quantified constraints.
+That is, the *concrete* syntax for the type construtor ``==>`` can still use ``=>``.
```
adamgundry, line 173:
```suggestion
That is, the *concrete* syntax for the type constructor ``==>`` can still use ``=>``.
```

proposals/0000-type-vs-constraint.rst
```diff
@@ -0,0 +1,284 @@
...
+However, users may also choose to write ``==>`` (imported from ``GHC.Exts``) if they choose.
```
adamgundry, line 174: What about `=>`? Can it be treated like a normal operator or does it still have special magic in the parser? There is an example later that suggests it can be used in a first-class manner, but currently something like `type T = (=>)` won't parse.

cdornan:
My observations.

  * The _Costs and Drawbacks_ as explained in &sect;6 appear (to me at least) to be negligible, and as such this proposal looks like a no-bainer. (I am more concerned with exploring the missed-opportunity costs of doing too little.)
  * You say the alternative putative fix would require several solid weeks of work &mdash; how much work is this fix estimated to need? (I am sorry if I missed it somewhere.)
  * Consider this thought experiment: if you knew what you know now when we introduced the key changes that led to this instability would you have proceeded with the changes? (Surely the correct answer must be no.)
  * Following on from the above thought experiment: have you considered just giving stuff up (temporarily at least), at the cost of limited end-user breakage to restore 'sanity'? (I suspect we are too far down the rabbit hole.)
  * I get the sense that Haskell and Core type representations really do need to break up! Can we really not envision some funky tools to exploit the similarity and make the separation easier to do progressively? (I am thinking of establishing two separate interfaces for working with types in Haskell and Core and mapping between them, with the actual separation deferred until everything is ready.)

Here I am playing the fool of course. I expect you will go ahead as planned with this proposal but I thought my best contribution would be asking some idiot questions to see what gives. I am expecting a short response!

Ericson2314:
@cdornan 

1. I agree we may well want to separate Core and Haskell types, but we can still do that after this proposal.

2. In the original GHC issue https://gitlab.haskell.org/ghc/ghc/-/issues/21623#note_435333, I hazard this guess that we can and will get rid of `TypeLike ~ ConstraintLike` in Core long term:
   > My uninformed guess is that long term we're going to get rid of heterogeneous equality, but get heterogenous representational coercions. E.g. it would seem CH-Coherence from https://richarde.dev/papers/2021/exists/exists.pdf requires them (though at the expression level)? If we can do that and get away with it, the problem is also solved, and `TypeLike ~ ConstraintLike` won't be derivable.

3. In that same thread (though I am having trouble finding it) I believe it was mention that `Core` knowing about coherence is valuable in its own right to e.g. enable more optimizations.

Bottom line is I think this proposal is an expedient solution that is in fact "barking up the *right* tree": forwards compatible with better things, and making decisions we would regret. (`TypeLike` being not apart from `ConstraintLike` I hope we will get rid of, but I view that as less correcting a mistake, than getting rid of something which is explicitly there to keep our options open / not allow code gets us into trouble in the short run.)

simonpj:
> I agree we may well want to separate Core and Haskell types, but we can still do that after this proposal.

I just want to record that I (spj) am very sceptical about the cost/benefit tradeoff of attempting this.  I think the costs would be very large, there may be unexpected and unwelcome consequences, and there will be lots of duplication both of code and intellectual effort.

I could be wrong in the end, but at least I don't think we should think of this as a realistic short-term path forwrard

AntC2:
>        type (=>)  :: forall (r1 :: RuntimeRep) (r2 :: RuntimeRep).
>                      CONSTRAINT r1 -> TYPE r2 -> Type  -- primitive

Pattern synonyms have that weird business with two `(...) => (...) => ...`. (And arguably there should be three sets of constraints; GADTs also could do with a similar mechanism(?).)

How does the typing for PatSyns get handled under this proposal?

proposals/0000-type-vs-constraint.rst
simonpj: yes

proposals/0000-type-vs-constraint.rst
simonpj: I have clarified.

proposals/0000-type-vs-constraint.rst
```diff
@@ -0,0 +1,284 @@
...
+Effect and Interactions
+-----------------------
+* We can fix type-checker tickets that have proved resistant to principled fixes.
+
+* The door is open to new innovations in strict classes.
+
+* This proposal is fully backward compatible.
+
+* This proposal is forward compatible with more glorious updates to the type/constraint
+  system we might imagine in the future, as detailed at TODO.
```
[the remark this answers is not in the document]
simonpj: I'm not sure, beyond "Kinds are calling conventions".  We'll have to wait for Richard to come back from holiday to find out.

simonpj:
> How does the typing for PatSyns get handled under this proposal?

We aren't proposing any changes to pattern synoyms.  The slightly odd concrete syntax for pattern synonym typse stays exactly as-is.

Ericson2314:
@AntC2 what that pattern syntax means (wanting constraints, then providing constraints) already cannot be written down as an expression type today, and is handled in a 2nd class manner instead. Arguable it would be nice to fix that some day, but this proposal makes the problem no worse at least.

aspiwack (commented):
Some comments below.

I'm still uneasy about this proposal. Having the `SORT` type may be a good thing, I don't really have an opinion on this. But I feel that there is a disconnect between the solution and the problem(s). I guess my question is: why does unifying `Constraint` and `Type` in this way help distinguish `Constraint` and `Type` in Core?

proposals/0000-type-vs-constraint.rst
```diff
@@ -0,0 +1,302 @@
...
+  type SORT :: TypeOrConstraint -> RuntimeRep -> Type
+  type IP   :: forall (r :: RuntimeRep). Symbol -> TYPE r -> CONSTRAINT r
+
+  type (=>)  :: forall (r1 :: RuntimeRep) (r2 :: RuntimeRep).
+                CONSTRAINT r1 -> TYPE r2 -> Type  -- primitive
+  type (==>) :: forall (r1 :: RuntimeRep) (r2 :: RuntimeRep).
+                CONSTRAINT r1 -> CONSTRAINT r2 -> Constraint
+
+  -- Data types
+  data TypeOrConstraint = TypeLike | ConstraintLike
```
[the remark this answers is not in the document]
adamgundry: Bear in mind that type families cannot appear in instance heads, and that the `TypeLike` data constructor occurs in `Type` by expanding type synonyms. Thus the drawback would be forbidding instances that match on `Type`, which is a pretty major restriction! See discussion starting from https://gitlab.haskell.org/ghc/ghc/-/issues/21623#note_432137

proposals/0000-type-vs-constraint.rst
```diff
@@ -0,0 +1,302 @@
...
+Costs and Drawbacks
+-------------------
+* This adds complexity to the root of our type system. However, we have learned
+  how to manage this complexity and protect users from seeing it. We do not expect
+  routine users to notice this change, but users who specify ``-fprint-explicit-runtime-reps``
+  will see some changes.
+
+Alternatives
+------------
+* This seems to be the best way to support unlifted and unboxed constraints.
```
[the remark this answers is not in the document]
simonpj: Yes  we considered having two distinct primitive type constructors, `TYPE` exactly as now, and `Constraint`.  Indeed that was our first plan.  But
* In Core we would have to say "If  `e :: ty :: k` then `k must be `TYPE <something>` or `Constraint`.  Similarly for the types of binders.  That "or" isn't fatal, but it's inelegant.
* We anticipate that it will not be long before people want unlifted constraints.  So then we'd add `CONSTRAINT`, thus:
  ```
  TYPE :: RuntimeRep -> Type
  CONSTRAINT :: RuntimeRep -> Type
  ```
  And now we have something isomorphic to what we propose (two types, vs one with a flag), except that our proposal allows all value-level terms with runtime rep `rr` to be treated uniformly.  E.g. `isUnliftedType` has one case rather than two.
* Collapsing the two into one `SORT` is witnessing the idea that types and constraints are the same thing at runtime.
* We anticipate further movement in this area, perhpas via Kinds are Calling Conventions.  If so, `SORT` gives us wiggle room to do that; and (even more important) we don't want to duplicate any such changes across two distinct type constructors, `SORT` and `CONSTRAINT`.

So the balance came down pretty firmly in favour of the design we offer.  Would it help to add the discussion above to "Alterantives"?

proposals/0000-type-vs-constraint.rst
```diff
@@ -0,0 +1,302 @@
...
+   using the existing ``(?x :: Int#)`` syntax. This allows programs to abstract over
+   implicit parameters without suffering inefficiencies due to boxing.
+
+2. This proposal allows a fix to several type-checker bugs that are otherwise elusive.
+   With this proposal, we can eliminate this entire class of bugs, that we otherwise
+   believe will continue to haunt GHC.  Currently in GHC, `Type` and `Constraint` are
+   *distinct* in the typechecker, but *identical* in Core.  That leads to massive duplication;
+   many, many functions over types have two versions, one for the typechecker and one for Core.
+   It also leads to intellectual complexity, with two subtly-different forms of type equality.
+   Worse, nothing maintains the separation, which leads to those elusive bugs.
```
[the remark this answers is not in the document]
simonpj: > Under this proposal there would still be at least one discrepancy

You mean that there are some ways in which the source language treats types differently to constraints?  Yes, absolutely. If they were treated identically, we would not need to distinguish them :-).    I'm not sure what you'd like to see here... can you suggest some words that would address your thought?

proposals/0000-type-vs-constraint.rst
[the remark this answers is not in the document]
adamgundry: I should have added: I think the proposal should somewhere say explicitly that instances matching on `Type` (or `Constraint`) remain allowed, hence matching on `TypeLike` or `ConstraintLike` must be allowed, thus motivating the use of a data kind with non-apart constructors rather than type families.

(I can't help but wonder if there is a connection to homotopy type theory lurking here somewhere, with `TypeOrConstraint` being the type with two points connected by a path.)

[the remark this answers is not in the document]
aspiwack: This makes sense, but I don't understand how we can be allowed to write instance of `TypeLike` and `ConstraintLike` if they are not apart.

proposals/0000-type-vs-constraint.rst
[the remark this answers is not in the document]
aspiwack: I think that it ought to be stated, indeed. As it stands, the problem and solution read pretty disconnected.

proposals/0000-type-vs-constraint.rst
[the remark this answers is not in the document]
aspiwack: I see. This is a little unfortunate. But as Simon says: not worse than today.

proposals/0000-type-vs-constraint.rst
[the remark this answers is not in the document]
aspiwack: No it was not what I meant at all. Sorry for the misunderstanding.

What I meant is:
- In the current situation: `Constraint /= Type` in the source language, but `Constraint = Type` in Core. This creates ample code duplication and subtle bugs.
- In the proposed situation: `x :: C :: Constraint` is invalid in the source language, but it is valid in the in Core. Does this create (ample?) code duplication and subtle bugs?

Maybe it doesn't. You've thought about it for quite a bit, you've probably come to the conclusion that it doesn't. But it's not obvious (at least not to me :slightly_smiling_face: ). This question, I guess, is part me trying to quench my curiosity, part due diligence.

proposals/0000-type-vs-constraint.rst
```diff
@@ -0,0 +1,365 @@
+Separating Type from Constraint
```
aspiwack, line 1: One more comment to object to the title. This proposal doesn't separate `Type` from `Constraint`: they are already separate. One of the motivations is that Core treats `Type` and `Constraint` as separate, but Core is out of scope for GHC proposal. At any rate, the real proposition is to make `Type` and `Constraint` instance of a common `SORT`. So maybe

_Generalising over Type and Constraint_

(I would normally have made that a suggestion, but, probably because I'm in a train with flaky connection right now, Github won't let me)

proposals/0000-type-vs-constraint.rst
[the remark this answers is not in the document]
Ericson2314: Changing the answer to decidable `Constraint ?= Type` is hugely more invasive than allowing something to be derived in the target language that isn't derivable in the source language. The latter is merely observing that compilation from Haskell to Core is not surjective, which is hardly surprising and to a large extent the the point of elaborating dictionary passing!

` x :: C :: Constraint` being admissible is a footgun insofar that explicit dictionary passing means it is possible to violate coherence, but this is already the case thanks to things like `IncoherentInstances` and, again, a cost of doing business with explicit dictionary passing unless we want to elaborate to "setoid hell" or similar (which would require a much fancier core language).

Overall I see nothing worth worrying about here. In your `Constraint` and `Type` alternative, we have the exact same issue as ` x :: C :: Constraint` is still admissable in Core. In making `Constraint = Type` everywhere we simply have a worse version of that foot-gun.

Neither the status quo nor any proposed alternative does any better with this issue.

Ericson2314:
@AntC2 "apartness" is used for type family LHS (patterns). It could also be used for instance heads but I suspect it is not the case that apartness is used for that..

proposals/0000-type-vs-constraint.rst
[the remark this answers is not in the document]
aspiwack: > Changing the answer to decidable `Constraint ?= Type` is hugely more invasive than allowing something to be derived in the target language that isn't derivable in the source language. The latter is merely observing that compilation from Haskell to Core is not surjective, which is hardly surprising and to a large extent the the point of elaborating dictionary passing!

This depends on implementation details. Please be mindful that I know this part of the GHC rather well. Vague hand-wavy answers are not what I'm after, here.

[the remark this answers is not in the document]
Ericson2314: We must be misunderstanding each other because nothing I wrote above was supposed to depend on implementation details beyond the basics of what Haskell and Core are, and the desugaring from the former to the latter.

`x :: C :: Constraint` is not a discrepancy in the sense of Haskell observing something which is *wrong* in Core, because that nothing in Haskell observes `x :: C :: Constraint` cannot be written; that is just a meta-theoretic property. In fact, because implicit parameters are constraints in surface Haskell (though they really shouldn't be), arbitrary  `x :: C :: Constraint` can be "resugared" back to surface Haskell well enough to round trip back to the Core with the same behavior modulo strictness.

The "invasiveness" I was referring to was from the user's perspective (the bugs from such a discrepancy where Core and Haskell conclude different and contradictory answers), not the implementation's. Sorry if that was unclear.

simonpj:
> @AntC2 "apartness" *is* used for type family LHS (patterns). It could also be used for instance heads but I suspect it is not the case that apartness is used for that.

In fact it is use for class-instance lookup.  Suppose we have
```
instance {-# OVERLAPPABLE #-} C Int a  where ...
instance {-# OVERLAPPING #-}  C Int Bool where ...
```
and we are trying to solve the constraint `C Int (F gamma)`, where `gamma` is an as-yet-unknown unification variable.  Since `(F gamma)` is not apart from `Bool`, it could be that (once we know `gamma`) the second instance should be chosen.  So we decline to commit to the first.  If `(F gamma)` later simplifes to, say `Char`, then and only then we choose the first.

Similarly with `Type` and `Constraint`, both today and with the proposal.  If we have
```
instance D Type where...
instance D Constraint where ...
```
and try to solve `C Int Type`.   This matches the first instance, but since `Type` is not apart from `Constraint` GHC thinks "oh, the second one could match in the future" and declines to commit to the first.

I'll add this explanation to the proposal.

proposals/0000-type-vs-constraint.rst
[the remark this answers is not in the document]
simonpj: > 

I think that it ought to be stated, indeed. As it stands, the problem and solution read pretty disconnected.

OK I added Section 7.1

proposals/0000-type-vs-constraint.rst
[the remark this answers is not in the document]
simonpj: Arnaud, I think your question is

> In the proposed situation: x :: C :: Constraint is invalid in the source language, but it is valid in the in Core. Does this create (ample?) code duplication and subtle bugs?

I'm confident that the answer is "no".  It is *already* the case that you can't have a term `(e :: ty :: Constraint)`, nor a binder whose type has that kind.  No change there.  It's easy to guarantee: when we start inferring the type of `e` or or `x` we simply give it a unification variable `alpha :: Type`. 

Nothing in the proposal affects any of this.  Does that reassure you?

simonpj:
The new apartness material is in section 2.2.

proposals/0000-type-vs-constraint.rst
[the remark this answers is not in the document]
goldfirere: It's in the proposal now: https://gitlab.haskell.org/ghc/ghc/-/issues/21623

goldfirere:
Just want to say that I've reviewed all of this (I wrote the initial text based on joint with with @simonpj and then sent it to him, who revised substantially and posted) and am pleased with the proposal as it stands. @simonpj shall we submit?

proposals/0000-type-vs-constraint.rst
```diff
@@ -0,0 +1,393 @@
...
+
+  instance C (Proxy @Constraint a) where ...
+
+But this is just the status quo; it is not a change (see Sectionn 2.2).
+
+Future stability
+:::::::::::::::::::::::::::::::::
+
+We anticipate that the kind of ``SORT`` may change again, for example to accommodate the ideas
+of `Kinds are calling conventions <https://simon.peytonjones.org/kinds-are-calling-conventions/>`_.
```
aspiwack, line 268: I think that the implicit argument, here, is that we want `CONSTRAINT` and `TYPE` to have the same arguments. The idea is something like: the type of evidence of `C :: CONSTRAINT args` is `DictC :: TYPE args`.

But it's not obvious to me that this is what we want. We could also have the type of evidence of `C :: Constraint` be `DictC :: ∃args. TYPE args`. This way, constraints are treated uniformly in the type checker, regardless of the kind of their evidence. We can easily have a pair `(C1, C2) :: Constraint` even if `C1` and `C2` have evidence of different kind.

Maybe the missing argument is: because we chose to distinguish `Constraint` and `Type` in Core, and having the type of evidence of `C` have the same kind as `C`, then we _need_ `CONSTRAINT` to have the same arguments as `TYPE` because it is going to be used in Core in all the same places.

I think this is true, and this is convincing to me, it shows how the design decision in the user-facing language follows from the choice of representation in the intermediate representation. Am I correct?

simonpj: I'm not following this, Arnaud.  You speak of an implicit argument, but all we are doing here is to advertise the SORT may change, but we will try to keep CONSTRAINT and TYPE stable.  So clients should try to use TYPE and CONSTRAINT, rather than SORT, to avoid being discombobulated if we change SORT later.

I don't see what you mean about "We couuld also have..."; we don't have existentials in types yet.  

I feel as if I'm missing your point, sorry!

simonpj:
@nomeata we'd like to submit this propsoal to the committee, please

proposals/0000-type-vs-constraint.rst
simonpj: Sorry, still not getting it.   What is `TYPE'` and `CONSTRAINT'`?  They aren't in the proposal.

As to the rest, are you thinking of the design alternative discussed in 7.1, perhaps?

proposals/0000-type-vs-constraint.rst
```diff
@@ -0,0 +1,393 @@
...
+to allow for strict classes and unlifted and unboxed implicit parameters.
+
+In addition, the new treatment proposed here allows GHC's internal language, Core,
+to more properly tell the difference between types and constraints, eliminating a
+troubling class of bugs that keeps recurring.
+
+Motivation
+----------
+
+1. With this proposal, we immediately allow unlifted and unboxed implicit parameters
```
gridaphobe, line 31: ImplicitParams seems to be widely viewed as a misfeature, or at least very niche. I'm not sure it would be accepted by the Steering Committee if it were proposed today.

With that in mind, I question how strong this motivation is. Is the lack of unlifted/unboxed ImplicitParams blocking real use cases?

proposals/0000-type-vs-constraint.rst
```diff
@@ -0,0 +1,393 @@
...
+We propose the following new setup, not repeating any types that remains unchanged::
+
+  -- Primitive type constructors
+  type SORT :: TypeOrConstraint -> RuntimeRep -> Type
+  type IP   :: forall (r :: RuntimeRep). Symbol -> TYPE r -> CONSTRAINT r
+
+  type (=>)  :: forall (r1 :: RuntimeRep) (r2 :: RuntimeRep).
+                CONSTRAINT r1 -> TYPE r2 -> Type  -- primitive
+  type (==>) :: forall (r1 :: RuntimeRep) (r2 :: RuntimeRep).
+                CONSTRAINT r1 -> CONSTRAINT r2 -> Constraint
```
gridaphobe, line 176: I'm not thrilled by the idea of adding another type of arrow to the existing mix, it feels like a lot to keep track of. And `=>` and `==>` feel very similar, in fact, are they not both instantiations of the following?

```
type (=>) :: forall (r1 :: RuntimeRep) (r2 :: RuntimeRep) (tc :: TypeOrConstraint).
             CONSTRAINT r1 -> SORT tc r2 -> SORT tc LiftedRep
```

Of course, this type is even more complex and abstract, and I don't particularly want to expose it to users..

I don't think I have any concrete suggestions here, just registering that the complexity feels bad.

proposals/0000-type-vs-constraint.rst
```diff
@@ -0,0 +1,393 @@
...
+
+  -- Data types
+  data TypeOrConstraint = TypeLike | ConstraintLike
```
gridaphobe, line 179: I don't think this would affect the proposal in a fundamental way, particularly since TypeOrConstraint will be an internal detail, but do we expect to ever add new cases here, e.g. as part of Kinds are Calling Conventions?

Perhaps another way of asking this question is: you cite Kinds are Calling Conventions (KCC) as an extension this proposal would better position us to support in the future, can you quickly sketch how you adapt the proposal to support KCC? Not being super familiar with that work it's hard for me to judge how much the current proposal buys us on that front.

gridaphobe:
Hi @goldfirere @simonpj!

I've left a few comments/questions inline, but here's my overall take on this proposal.

* This proposal is primarily about fixing a recurring class of tricky bugs in GHC.
    * You cite other motivations around ImplicitParams and future innovation in constraints, but these feel minor and hypothetical respectively. Bugs, on the other hand, are real and we should fix them in a good and sustainable way.
    * As I haven't been involved in GHC development in a long time, it's hard for me to judge the amount of pain these bugs are causing. But since we're in the area of type equality, I imagine the pain is substantial :)
    * The proposed solution, making `Type` and `Constraint` distinct in both Source Haskell and Core, seems sensible to me. I can't easily judge how much the new type structure will improve the maintainability of GHC, but it looks reasonable from a user-facing perspective. (I do agree that it's important to hide the new details from user-facing error messages to avoid confusion.)

So, this feels mostly like we're cleaning up an internal detail that should not affect users, and may prove to have benefits for future language extensions. As such, I'm inclined to trust your judgment that the new type structure will prove more robust and maintainable, and to recommend acceptance.

proposals/0000-type-vs-constraint.rst
simonpj: Well, it's pretty strong you like implicit parameters.  Ed Kmett wants this for sure.  I suppose someone could do some stats on Hackage to see how widely implicit parameters are used.   (Personally I don't see them as  a mis-feature, fwiw.)

proposals/0000-type-vs-constraint.rst
simonpj: Hmm.  We explicitly avoided any polymorphism in `TypeOrConstraint`, in an attempt to avoid complexity.  But you are right that there is a trade-off here: we could have a single, polymorphic type constructor in exchange for, well, it having to be polymorphic.  Indeed we could go further.  We could make FUN do it all
```
type FUN :: forall (m :: Multiplicity) ->
            forall (r1 :: RuntimeRep) (r2 :: RuntimeRep).
            forall (tc1 :: TypeOrConstraint) (tc2 :: TypeOrConstraint)
            SORT tc1 r1 -> SORT tc2 r2 -> Type 
```
These seem like possible futures.  We thought that being fully monomorphic was the most conservative choice -- and one that we could liberalise later -- but I think we'd be open to the committee's guidance on this.  @rae have I missed anything?

proposals/0000-type-vs-constraint.rst
simonpj: Adapting [Kind as Calling Conventions](https://www.microsoft.com/en-us/research/publication/kinds-are-calling-conventions/) to this proposal we might get:
```
type SORT :: Convention -> TypeOrConstraint -> RuntimeRep -> Type
data Convention = Eval levity | Call ArityDescription
data ArityDecription = ACons RuntimeRep ArityDescription | AZero | AConv Convention
```
Here we add a new `Convention` argument to `SORT` that describes the arity of the thing.

I'm not certain that this is the right thing to do, but it's plausible.

My long term plan is this: 
* in Core we make sure that the argument of a function has the *type* expected by the function
* in STG we make sure that the argument of a function has the *kind* expected by the function

So in STG we have a coarser well-typed-ness condition -- but that coarsening in turn allows us to do more CSE and unboxed sums and other such things.

But to call functions with the right calling convention we would really need to express the calling convention in the kind.  Hence KACC.

simonpj:
> As I haven't been involved in GHC development in a long time, it's hard for me to judge the amount of pain these bugs are causing.

For me, this pain is the primary motivation.  We have
* Multiple open bug that we cannot satisfactorily solve.
* Much code duplication
* No way to police the boundary between code that needs to be duplicate and code that doesn't.

It's a persistent and very painful problem, one that we have struggled with for years.  I am very keen to cauterise it at source.

goldfirere:
I agree with @simonpj here. I wanted to lead with the user-facing motivations in the proposal, but bug-fixing is the real motivation. However, I'm leery of imposing a user-visible language change just to fix what is, at its core, an implementation bug. So I put the user-facing motivation first, rightly or wrongly.

proposals/0000-type-vs-constraint.rst
gridaphobe: I do like ImplicitParams! But I've always felt I was in the minority in that opinion :)

proposals/0000-type-vs-constraint.rst
gridaphobe: > I have no idea what polymorphism over TypeOrConstraint would mean. That is, it's effectively polymorphism over visibility... and if a function is polymorphic over visibility, does it take an argument or not?

Hmm, interesting point. I agree that that's a whole can of worms that's best left unopened if we don't have a concrete usecase..

Let me take another approach. 

Here's what struck me about `=>` vs `==>`. If I understand Section 3.1 correctly, then there are two cases where `==>` is required by the type system: instance heads and quantified constraints. Both are syntactically identifiable, evidenced by the fact that we can continue to accept `=>` in those cases for backcompat purposes.

Is the introduction of `==>` as a stable type (see Section 3.4) premature? It seems like we don't actually need to make it public in order to implement the proposal. Suppose instead we leave `==>` in GHC.Prim. We'd be taking away the ability to abstract over `==>` _in an API-stable manner_, but I have a hard time wrapping my head around why you would care to do that in the first place. 

If users come back to us later saying they really do need `==>` as a first-class thing, we can then revisit the question of how polymorphic (or not) to make the various function arrows.

gridaphobe: @goldfirere also, if we don't know how to reason about polymorphism over `TypeOrConstraint`, we should probably ban it explicitly. I understand that `TypeOrConstraint` will only be exported from `GHC.Prim`, but a user could still write their own polymorphic function over it using non-stable APIs, and it sounds like we wouldn't know what to do with that definition.

proposals/0000-type-vs-constraint.rst
nomeata: But not alone with it - for some things they are great :-)

proposals/0000-type-vs-constraint.rst
simonpj: > Suppose instead we leave ==> in GHC.Prim.

That's an interesting idea.  Indeed we can use it for `(=>)` also.   That is:
* `(=>)` and `(==>)` may indeed exist in Core as `TyCons` but we don't expose them to the user, except maybe through an advertised-as-unsable GHC.Prim -- exactly like `SORT` in fact.
* The *only* places you can write `=>` is in the current, syntactically specified positions; e.g. `f :: Eq a => a -> a`.  For example, you cannot write `T (=>)` or `F ((==>) (Num a))`.

I can't seen a downside to this plan.  It leaves us a bit more wiggle room for the future, and no one is asking for a first class `(=>)` arrow.  @rae?

Maybe `FUN` is the same?  Currently it's exported by `Data.Kind`, but I guess we could change that.  That would be an un-forced change, but one that is consistent with the plan of being more explicit about API stability.   It could mean that some packages would need to change imports -- but how many packages use `FUN` directly?

When I look I see that `GHC.Exts` re-exports all of `GHC.Prim`.  So maybe `GHC.Exts` is advertised-unstable?   For a stable API you want modules without GHC in the module name.

simonpj: > if we don't know how to reason about polymorphism over TypeOrConstraint, we should probably ban it explicitly

There's nothing directly wrong with it.  E.g.
```
data T (a::TypeOrConstraint) where
   T1 :: T TypeLike
   T2 :: T ConstraintLike

f :: forall (a :: TypeOrConstraint). T a -> Int
f T1 = 2
f T2 = 3
```
This is fine; `TypeOrConstraint` is just a vanilla data type.  It's just polymorphic **arrows** that are a problem, and one that we propose to nip in the bud by, well, not providing polymophic arrows.

This does raise another question: maybe `TypeOrConstraint` should be be part of the `GHC.*` exports (unstable) and not the `Data.Kind` exports (stable); see discussion above.

proposals/0000-type-vs-constraint.rst
goldfirere: I based the design of my [`HUnit-approx`](https://hackage.haskell.org/package/HUnit-approx) library on implicit parameters. The feature shines there. And I thought the package was utterly unused, only to learn recently that a new-and-upcoming automatic differentiation library is using it. :) (I don't think that's public yet, so no link, sadly.)

proposals/0000-type-vs-constraint.rst
goldfirere: I think we currently do a poor job of separating API-stable bits from API-unstable bits. For example, the best way to get `+#` is to import `GHC.Exts`, but `+#` feels very stable to me. On the other hand, we're now proposing some unstable bits.

I could imagine having the unstable bits come from `GHC.Prim` and the stable ones come from `GHC.Exts`. But I don't think we should, for a few reasons:
- The export list of `GHC.Prim` is intimately tied with other implementation choices. We might, for example, take a definition that was previously primitive and make it not so; this would remove the export from `GHC.Prim`, even if we don't actually want to.
- The name `GHC.Prim` does not loudly say that it is unstable. ("Primitive", perhaps, but not unstable.) If someone spots `import GHC.Prim` on some blogpost, they're not going to know that they're building on sand.

I thus think we should make a new place to export unstable bits, perhaps `GHC.Exts.Unstable`. That is, `GHC.Exts` exports GHC-specific (and sometimes magical) things, but we roughly expect it to be stable over time. We might even commit to a migration strategy when making changes. On the other hand, `GHC.Exts.Unstable` can change at any time with no warning. Perhaps even better would be to have a new package `ghc-unstable` exporting unstable modules, and then it would be easier for e.g. an enterprise to scan its dependencies for something untoward. But I'm worried about the possibility of package-import cycles, so it's perhaps best to start small, with `GHC.Exts.Unstable`.

goldfirere: As to not exporting things: sure. But I would advocate for exporting `(=>)` from a stable API. That symbol has been around forever, and with the kind (effectively) we are ascribing it. Maybe `(==>)` isn't exported, or exported only from `GHC.Exts.Unstable`. I agree no one is asking for `(==>)`.

I *don't* think we should, at this point, move any definitions around in an unforced manner. That should be done as part of a broader attempt to solidify GHC's stability guarantees.

simonpj: > The name GHC.Prim does not loudly say that it is unstable. 

I agree that changes here are orthogonal to this proposal.  

But (separate to this proposal) my suggestion is that any `import GHC.xxx` signals "building on sand".  Any stable API should not start with "GHC".  That's a simple rule

goldfirere: I don't feel strongly about exporting `(=>)`. But anyone can write a type of kind `Constraint -> Type -> Type` if they want to, so it's plausible someone would want to abstract over types of that kind.

I'd be fine with `GHC.xxx` signaling instability, but that goes against a large amount of existing practice. For example, robust libraries (e.g. aeson) are built on `GHC.Generics`, and all primitive operations require imports from the `GHC.` namespace. So I personally think that ship has sailed and we should find a different way of signaling instability.

nomeata: `GHC.Stack` came to my mind as a stable(?) API in GHC..

proposals/0000-type-vs-constraint.rst
AntC2: Implicit Parameters: "[Most of the following, still rather incomplete, documentation is due to Jeff Lewis.](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/implicit_parameters.html)" -- and the text there in the User Guide seems to be the same as for Hugs. So it's been "rather incomplete" for something like two decades.

@EKmett likes them, but is also [wary of gotchas](https://old.reddit.com/r/haskell/comments/m8o88a/who_still_uses_readert/grkh17q/) -- at least as of a year ago.

If IPs are a major motivation here, is it time to show them some love?

proposals/0000-type-vs-constraint.rst
gridaphobe: I've always viewed `ghc-prim` (the package) as the unstable API, and the `GHC.*` modules within `base` as relatively stable. Looking again at the breadth of functionality exposed by `base`, that's probably too broad a brush to paint with, but it seems directionally correct to me.

Coincidentally, the proposal suggests exporting the unstable types from `GHC.Prim`, which lives in `ghc-prim`, so I think we're good here. We just need to update the proposal to export `(=>)` and `(==>)` from `GHC.Prim` rather than `Data.Kind`.

simonpj:
I believe that, aside from deciding where to export `(=>)` and `SORT` from (suggestion above), the proposal is broadly acceptable to the committee.  Is that right?  If so, perhaps we can tie the bow.

gridaphobe:
@simonpj agreed. Let's update the proposal to reflect `(=>)` and `(==>)` being unstable identifiers:

> The natural place would be from package ghc-prim, module GHC.Types -- that is where LiftedRep etc are currently defined.
> 
> Maybe (for now) we can signal that modules in ghc-prim are potentially unstable?

and we can merge this proposal.

Ericson2314:
> Maybe (for now) we can signal that modules in ghc-prim are potentially unstable?

This is definitely good to do regardless. Big +1 from me.

goldfirere:
I've posted #524 to create a mechanism with which we can protect unstable features. Assuming that proposal is accepted, I've modified the current one to move the new features into the unstable territory.

gridaphobe:
Thanks @goldfirere!

@nomeata I think we're good to merge here (once you're back from vacation).

goldfirere:
I don't think we're quite ready to merge, actually, given continued conversations around the `-XUnstable` mechanism. The proposal currently says that we're exporting e.g. `(==>)` from a module users can import without restraint or warning, and I think that may get us stuck in the future.

simonpj:
I think we can simply say, in this proposal, that `(=>)` and `(==>) ` are not part of the advertised API of the proposal.   The enforcement mechanism for this "not advertised API, subject to change" is separate matter, and applies to many, many existing GHC internals.

Let's just say this explicitly in the proposal and them land it.

goldfirere:
But saying this in this proposal (only) doesn't help our users, who will not generally read this proposal. I'd be fine with stating in the proposal that Set A of exports are stable, Set B of exports are unstable, and that we leave unspecified how users will discover the difference between these -- but that we commit to making this difference obvious to users before merging/releasing the change. If we don't do this, then I fear we'll be stuck either a) supporting the current design into perpetuity , or b) changing it later, breaking compatibility and angering our user base. Both of these seem like bad outcomes.

gbaz:
I believe the traditional way that warnings about instability have been transmitted through users is through comments in the haddocks?

simonpj:
> Set A of exports are stable, Set B of exports are unstable, and that we leave unspecified how users will discover the difference between these

Yes.  My point is that this is *precisely* the situation with vast swathes of `base`.  Nothing new here.

treeowl:
I have a silly question. Why do we even want `Type` and `Constraint` not to be absolutely the same, even in the surface syntax? It's clear that `=>` and `==>` are very different from `->` in Haskell. But can we collapse `Type` with `Constraint` and `=>` with `==>`? What goes wrong?

gridaphobe:
I don't think this proposal needs to specify the *mechanism* of communicating unstable APIs, it just needs to specify which types are stable or not.

gridaphobe:
@goldfirere are you content with the wording around stable/unstable exports now?

If so, @nomeata let's merge!

simonpj:
I have adjusted the text a bit to reflect the fact that we now propose to have two distinct type constructors, `TYPE` and `CONSTRAINT`, rather than having `SORT` plus `TypeOrConstraint`.  

I have also elaborated 3.4 about stability.

nomeata:
Will merge once I get a 🟢 from @goldfirere

goldfirere:
I continue to want a way for our users to discover what parts of the API are stable and what parts are unstable. This proposal specifies the distinction cleanly, but does not articulate how that will be made apparent to users. I continue to believe that if we do not make this difference more manifest, we will be sitting here in 3 years' time wanting to change this system but with too many users relying on the parts we meant to be unstable. That said, I feel like others do not share this concern, and so I accept that my thoughts are out of step with others' and do not wish to block this proposal.

Quite separately, I worry that, until the implementation is all finished (I have not reviewed it yet), there may still be changes to come here. After all, this is really driven by implementation concerns, not user-facing design questions. Does it make more sense to merge now and change later? or to wait until the implementation is finished and merge then?

simonpj:
> I continue to want a way for our users to discover what parts of the API are stable and what parts are unstable.

So do I.  Very much.  But that is the subject of #528.  Meanwhile this proposal is clear about what is stable and what is not.

> Quite separately, I worry that, until the implementation is all finished (I have not reviewed it yet), there may still be changes to come here. After all, this is really driven by implementation concerns, not user-facing design questions. Does it make more sense to merge now and change later? or to wait until the implementation is finished and merge then?

There's an obvious chicken and egg problem here.

However, since there have been no objections, I have invested a good week of work in the implementation, which is nearly complete.  https://gitlab.haskell.org/ghc/ghc/-/merge_requests/8750.

I have no objection to deferring merging until we have finished all the details in case there is a bit more to say in the proposal.

L-as:
Question: Would classes with this change be represented as unlifted constraints, since they're supposed to always not be bottom?

L-as:
Question: Would classes with this change be represented as unlifted constraints, since they're supposed to always not be bottom?

simonpj:
> Question: Would classes with this change be represented as unlifted constraints, since they're supposed to always not be bottom?

Good question.  But they are lifted now, and this proposal doesn't change that.

In fact making them unlifted would be tricky because of recursive dictionaries.  But in any case, this proposal should not make unrelated changes in Haskell; that should be another proposal..

Issue #383, opened 2020-12-04
Promoted methods before the rest of DH


Ericson2314:
Today's type families, while trying to do a generally-liked thing, are a bit weird. See #381 for #177 for examples of problems with them. Type families are also quite difficult to reform. @goldfirere concludes in part of a comment at the bottom of #177:

> The more I work on Haskell, the more I wonder whether type families are something of a misfeature: instead, we should just be able to use functions in types. Class methods would, naturally, induce constraints, just like proposed here. But I worry that the investment in this idea would end up being redundant once ordinary functions (and methods) can be used in types. (To be fair, I'm very pleased that Haskell got type families when it did. We would have never gotten to this point without them. It's just that I wonder about their future.)

To provide a bit of context, "constrained type families" as discussed in #177 suggest borrowing the idea of constraints to improve type families. But class methods at the type level naturally would keep their constraints too. And the openness of class methods can fill the role of the openness of type families.

I agree with @goldfirere's quote. But, I also think "better" type families are useful even if we don't have Dependent Haskell. That isn't to say I don't want dependent Haskell --- I do --- but rather than making users wait for Dependent Haskell to see these issues with families fixed, perhaps we could provide a DH-anticipating alternative[1] to type families. I think showing how the DH approach can both reveal these subtle problems and solve them can also build more support for the rest of DH.

Ericson2314:
Now I'll make an attempt to explore what doing this entails.

------

In #281, as a stop-gap, certain term-level syntax can be reinterpreted as type-level syntax via a "T2T mapping". This allows us to allow users to take advantage of progress towards full Dependent Haskell before the type- and term-level languages are actually the same. 

Likewise, one might be imagine certain definitions being usable at the type level before others are. But this only works for concrete definitions --- in fact concrete definitions that only refer to other definitions with "known promotability". It also leaves user no way to assert the promotability of a term-level definition other than trying to promote it.

An intriguing syntactic option is to go in reverse and *demote* type synonyms. I like this a lot, except that associated types are indistinguishable from hypothetical types synonyms in instances (to spoil where we're going), so I'll just use a `{-# PROMOTABLE #-}` pragma for now. I fully acknowledged this is ugly and hopefully we can do better.

---------

So, let's back to type families. We have a few cases:

 - Promotable functions aim to replace closed type families
    - Some restrictions for now
      - Partial ones require error cases, promoting bottom values or exceptions is more and boring work.
[...]

PR #521, opened 2022-06-22
Succinct associated type instances

L-as:
[Rendered](https://github.com/mlabs-haskell/ghc-proposals/blob/succinct-associated-type-instances/proposals/0521-succinct-associated-type-instances.md)

When declaring associated type family instances in Haskell,
you need to repeat the names of the arguments, unless
you don't use any of the type variables from the instance head.

In that case, you can replace the LHS with a `_`, which is
much more succinct and heavily reduces noise.

This proposal generalises this feature, allowing you to *also* refer
to the type variables bound in the instance head.
This allows you to make use of this succinct syntax no matter
the type instance you want to write:

```haskell
class C (a :: Type) where
  type F a :: Type

-- before
instance C [a] where
  type F [a] = a

-- also before
instance C (Maybe a) where
  type F _ = () -- couldn't refer to `a` before this proposal

-- now
instance C [a] where
  type F _ = a -- can refer to `a` now
```
While seemingly a small change, this can heavily reduce the amount of noise in code
with complex instance declarations.

In addition, this enables the same feature for *default type family instances*,
as reported in https://gitlab.haskell.org/ghc/ghc/-/issues/21702.

### DIFF
--- a/proposals/0521-succinct-associated-type-instances.md
+++ b/proposals/0521-succinct-associated-type-instances.md
@@ -31,7 +31,7 @@ instance C [a] where
 
 -- also before
 instance C (Maybe a) where
-  type F _ = () -- can't refer to `a` before this proposal
+  type F _ = () -- couldn't refer to `a` before this proposal
 
 -- now
 instance C [a] where


Ericson2314:
This is a good stop-gap, but just keep in mind the deficiency in the current syntax mirrors a deficiency in the current semantics.

We *really* ought to have
```haskell
class C (a :: Type) where
  type F :: Type

instance C [a] where
  type F = a
```
i.e. `F` is a plain field of the dictionary no arguments needed.

Instead, the way things work today is that `F` is barely associated with the class at all; the making it part of the class is a rather *aspirational* notational convenience, and the duplicated instance head leaks out the "truth".

As such,

> In addition, this enables the same feature for _default type family instances_, as reported in https://gitlab.haskell.org/ghc/ghc/-/issues/21702.

I fear we unfortunately cannot fix that with this change alone. Since the type family isn't really part of the class, nor is type family resolution at all connected to instance resolution, that restriction will have to persist.

------

Once we have more Dependent Haskell, we can hopefully replace type families with "promoted methods" per https://github.com/ghc-proposals/ghc-proposals/issues/383, and then these and many other problems go away.

L-as:
OTOH maybe an associated type family that isn't parameterised by all the class parameters is simply bad design, and the motivating example with fundeps should be solved by having a hidden type family for all fundeps.

Ericson2314:
Yeah fun deps should be sugar for type familes which in turn should be replaced by promoted methods :).

Any type family that doesn't use all parameters can go in a superclass with just the relevant parameter. And if there was a fun dep like your example, the elaboration of the fun dep would create exactly that superclass with the promoted method!

AntC2:
No I don't like this. `_` in a pattern means wildcard/don't care. But that's not what it means here. Remember Assoc types are just sugar for (stand alone) Type Family decls. Then

```haskell
-- proposed 
instance C [a] where
  type F _ = a

-- must desugar to
type instance F [a] = a
-- vs
type instance F _ = a         -- not valid
```

> much more succinct and heavily reduces noise

_much_ more, _heavily_ reduces? I think you're exaggerating. Please supply some realistic examples of this saving.

> @L-as ... the motivating example with fundeps should be solved by having a hidden type family for all fundeps.

No, the semantics of (Assoc) type families is subtly different from FunDeps -- which is the difficulty @Ericson2314 is alluding to. In common cases you can't observe the difference, but sometimes it bites. [Section 3 'The Totality Trap'](https://jgbm.github.io/pubs/morris-icfp2017-families-extended.pdf).

tomjaguarpaw:
As a user I think this is of negative benefit to me.  If I see `foo :: F [a] -> a` then I really want to look up the definition `type F [a] = a` and deduce that `foo :: a -> a`. Seeing `type F _ = a`, indirecting to `C [a]`, and then substituting back to `type F [a] = a` is pretty convoluted.

L-as:
> As a user of `F` I think this is of negative benefit to me. If I see `foo :: F [a] -> a` then I really want to look up the definition `type F [a] = a` and deduce that `foo :: a -> a`. Seeing `type F _ = a`, indirecting to `C [a]`, and then substituting back to `type F [a] = a` is pretty convoluted.

That can already happen though.
You can *already* have `type F _ = ()`.

tomjaguarpaw:
Interesting. I've never seen this. So I guess you are saying that the binding site for `a` should really be the instance head, so we should be able to use `a` regardless of whether it appears as a parameter in the `type` definition.

L-as:
Indeed, and you can't use a different "type instance head" in any case! Doing `type F [b] = ()` is not allowed, it has to be `type F [a] = ()` or `type F _ = ()`.

AntC2:
> Interesting. I've never seen this.

And for very good reasons: it's rare for a type function to completely ignore its type parameter(s) and always return a fixed type like `()`. Note that currently in a type family instance:

```haskell
type instance F _ = ()
```

you can't mention any type variables on RHS. The `_` is explicitly saying to ignore `F`'s argument. That's why I think @tomjaguarpaw's first intuition was correct: seeing the `_` but with a different meaning is very confusing.

There's another parallel with this binding at the head vs at the line business: GADTs:

```haskell
data G (a :: Type)  where
  MkG :: Ord b => b -> G b
```

The `a` in the head is a dummy. (I'm using it there to give a type-of-type.) It's common to re-use `a` in the constructor signatures, but not obligatory. But I must use some tyvar/`_` is not allowed/there's no presumption of using a tyvar from the head.

Omitting tyvars in Assoc type families (and only in those: not in stand-alone Type Families, not in GADTs) is just too confusing.

simonpj:
The original ticket is https://gitlab.haskell.org/ghc/ghc/-/issues/21702, and I think there is a sensible proposal there.  Namely, that in a class declaration you can declare a default instance like this
```
class C x y z where
   type F y :: Type           -- The type-family signature
   type F y = Maybe z    -- The default declaration; note the `z` here
```
Just like default methods, the default associated type is used to fill in any associated type decls that you omit in an instance decl. Thus if you write
```
instance C Int [a] (Tree a) where {}    -- No explicit declaration of F
```
then the default declaration from the class is used, suitably substituted, and the net result is just as if you had written
```
instance C Int [a] (Tree a) where
   type F [a] = Tree a
```
This is useful.  And you can't do it in GHC today.  

However I am entirely un-persuaded about the "_" business.  That seems like an entirely unrelated idea. (Not a stupid idea, but (a) unrelated (b) debatable and hence (c) distracting.  Nor does it let you do anything you can't do in GHC today.

My recommendation: refocus on the the original goal, in #21702

L-as:
I'll make another proposal for that then, albeit I do think they're quite related. They're both about making the type variables in the class or instance head available in the (default) instance.
The `_` feature already exists after all.

simonpj:
Thanks.  I think it would **greatly** increase the chances of making progress if you keep them separate.

AntC2:
> The original ticket is ...21702, and I think there is a sensible proposal there.

Yes I carefully read that ticket before commenting. (I also commented on that ticket.) With respect @simonpj, there's a difference between the o.p. on that ticket and what you've been writing (both on the ticket and just above here). Namely that @L-as's examples have a FunDep on the class. I think it's so redundant to have both a FunDep and an Associated type yielding the same result as to render this a 'feature' I would never use.

I also don't see how we could in general desugar from some arbitrary instance head to the needed type family instance by 'plugging in' to the assoc type's default.

```haskell
class C x y z where
   type F y :: Type              -- The type-family signature
   type F y = Maybe z            -- The default declaration

instance C (Either a b) Bool a   -- repeated tyvar in head needs FlexibleInstances
-- plug in to the default
type instance F Bool = Maybe a   -- where to get the `a` from?
```

Or in general is it expected that generating a type family instance from an Assoc type default might produce an invalid equation?

Now also consider the case where `C` has a FunDep `x y -> z`: the instance head is fine, thanks to the repeated tyvar `a`.

(And thank you for the guidance re the `_` business.)

simonpj:
> Or in general is it expected that generating a type family instance from an Assoc type default might produce an invalid equation?

Absolutely it might!   (And this should be explained in the proposal.)  With my example class C above
```
instance C Int Bool [v] where {}
```
would expand to
```
instance C Int Bool [v] where
  F Bool = [v]
```
and that is invalid, precisely as it would be if you'd written it by hand.  (Because in any type instance decl, the variables used on the RHS must be bound on the left.

The default declaration is like a macro: use this default declaration (suitably instantiated) if the user didn't write one.  But then treat it just as if the user *did* write it.

Thanks for pointing this out.

adamgundry:
> I think it's so redundant to have both a FunDep and an Associated type yielding the same result as to render this a 'feature' I would never use.

Another way to see this is as a way to get a bona fide fundep, with evidence and an explicit name for the corresponding type family. In particular, one can do this:
```
class b ~ F a => C a b where
  type F a :: Type
  type F a = b
```
Now any instances for `C` will be required to strictly respect the fundep (no bogus checks or overlapping instances funny business, even across modules), and GHC will be able to use `(C a b, C a b')` to conclude `b ~ b'` regardless of whether the constraints are wanteds or givens. Yet the instances don't even have to mention `F` explicitly. This seems very nice.

L-as:
> Another way to see this is as a way to get a bona fide fundep, with evidence and an explicit name for the corresponding type family. In particular, one can do this:
> 
> ```
> class b ~ F a => C a b where
>   type F a :: Type
>   type F a = b
> ```
> 
> Now any instances for `C` will be required to strictly respect the fundep (no bogus checks or overlapping instances funny business, even across modules), and GHC will be able to use `(C a b, C a b')` to conclude `b ~ b'` regardless of whether the constraints are wanteds or givens. Yet the instances don't even have to mention `F` explicitly. This seems very nice.

The use case in my situation was anyclass deriving. I wanted to have something like
```haskell
data A = A deriving anyclass (C B)
```
and have it define a type family, because of some type-level programming I was doing where you needed `F A` to be defined.
It's not "clean" in some sense, because I'm abusing multi-parameter type classes when I *only* want a type family.
That's why I also suggested this syntax, because then you could just do
```haskell
instance C (A a) where type F _ = B a`
```
Which isn't *as* succinct, and certainly doesn't work with derives, but if you need to repeat the instance head in the type instance,
it often becomes two lines of code, and adds a lot of noise if you have multiple of these kind of classes.

Perhaps the issue is that derives aren't powerful enough, but fixing them seems like a much more complex task.

AntC2:
> > Or in general is it expected that generating a type family instance from an Assoc type default might produce an invalid equation?
> 
> @simonpj Absolutely it might!

I'd assumed the reason this is currently rejected is as a simplistic but syntactically easy to check safety catch on your footgun.


> @adamgundry Another way to see this is as a way to get a bona fide fundep, with evidence and an explicit name for the corresponding type family.

I still see it as needless duplication/a code smell with the structure of your classes and instances.

I'm finding each of those justifications bizarre. I don't think I'll be using this feature. I'll shut up.

goldfirere:
I'm afraid I don't see how this proposal would work. For example:

```hs
class C a b where
  type F a
instance C (Maybe a) [b] where
  type F _ = b
```

That would seem to be accepted in this proposal. But what would be the value of `F (Maybe Int)`?

I agree with @simonpj: let's just focus on class declarations, not instance declarations.

L-as:
> I'm afraid I don't see how this proposal would work. For example:
> 
> ```haskell
> class C a b where
>   type F a
> instance C (Maybe a) [b] where
>   type F _ = b
> ```
> 
> That would seem to be accepted in this proposal. But what would be the value of `F (Maybe Int)`?
> 
> I agree with @simonpj: let's just focus on class declarations, not instance declarations.

`b` would be in scope in some sense, albeit would err since it's not bound on the LHS, as I think I noted in the proposal.

PR #523, opened 2022-07-22
`let` in types and patterns, now binding type variables

goldfirere:
This proposal includes the parts previously of #448 that were not accepted. The proposal extends `let` in three ways: allowing `let` to appear in types, allowing `let` to bind type variables (making them effectively local type synonyms), and allowing `let` amidst patterns. All are controlled by a new extension `-XExtendedLet`.

[Rendered](https://github.com/goldfirere/ghc-proposals/blob/extended-let/proposals/0000-extended-let.rst)

### DIFF
--- a/proposals/0000-extended-let.rst
+++ b/proposals/0000-extended-let.rst
@@ -178,6 +178,9 @@ Effects
 Effect and Interactions
 -----------------------
 
+.. _`Explicit Binding Principle`: ../principles.rst#explicit-binding-principle
+.. _`Syntactic Unification Principle`: ../principles.rst#syntactic-unification-principle
+
 1. The `Syntactic Unification Principle`_ is supported. The new ``let`` syntax in types is a strict subset
    of its syntax in terms, and the semantics are compatible. Note that allowing ``let`` in types brings
    us closer to getting this principle.
--- a/proposals/0000-extended-let.rst
+++ b/proposals/0000-extended-let.rst
@@ -11,11 +11,12 @@ Let Extensions
 .. contents::
 
 .. _`#448`: https://github.com/ghc-proposals/ghc-proposals/pull/448
+.. _`#378`: https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0378-dependent-type-design.rst
 
 This proposal suggests several extensions to Haskell's ``let`` syntax,
-allowing for local type synonyms (but not datatypes or classes), ``let`` in
-types, and ``let`` amongst patterns. The different features in this proposal
-may be accepted independent of the others, but I think they synergize nicely.
+allowing for local type synonyms (but not datatypes or classes) and ``let`` in
+types. The two different features in this proposal
+may be accepted independent of the other, but I think they synergize nicely.
 
 This proposal was spun off from `#448`_, but you do not need to read that
 proposal to understand this one.
@@ -43,6 +44,9 @@ Motivation
    type variable: in the scope of ``b``, ``b`` is synonymous with ``Bool``. Yet the only way
    to make such a ``b`` is via a pattern (or result, `#228`_) signature. Why force users
    to use matching instead of binding the variable directly?
+
+#. Supporting ``let``\ -binding in types gets us a bit closer to unifying the syntax
+   of types and terms, in sympathy with `#378`_.
 
 Proposed Change Specification
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -87,151 +91,10 @@ Effects
    is nothing stopping us from doing so, but it would seem to violate expectations of Haskellers
    and would be the first instance of a lower-case type variable being in scope at the top level.
 
-``let`` in patterns
--------------------
-
-This part of this proposal allows introducing a ``let``\ -binding in a pattern.
-The bound variable(s) scope over the same region of code as the pattern-bound
-variables do.
-
-The syntax for this feature is a bit awkward, and so this proposal presents
-two alternatives for discussion. I have a slight preference for Alternative 2.
-
-This goes beyond any previous proposal.
-
-Motivation
-~~~~~~~~~~
-
-1. This is needed to uphold the `Explicit Binding Principle`_.
-
-#. Though admittedly a weakish motivation, there is currently no way to share
-   expressions used in common in multiple view patterns. See `examples <#let-in-pattern-example>`_
-   below.
-
-Specification Alternative 1
-~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-1. With ``-XExtendedLet``, add a new form of pattern as follows::
-
-     pat → 'let' decls 'in' pat
-
-#. Any entites bound in ``decls`` scope over the same region of the program
-   that pattern-bound variables scope over, with the addition of the ``decls``
-   themselves (that is, the declarations can be recursive).
-
-Specification Alternative 2
-~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-0. **Background**. Here are some productions from the `Haskell 2010 Report`_ (to be changed
-   by this proposal)::
-
-     funlhs → var apat {apat}
-            | pat varop pat
-            | '(' funlhs ')' apat {apat}
-
-     apat → gcon               -- this one is unchanged, but provides context
-          | literal
-          | …
-
-     lexp → '\' apat1 … apatn '->' exp   (n ≧ 1)
-          | …
-
-     lpat → apat
-          | '-' (integer|float)
-          | gcon apat1 … apatn
-
-   (Recall that braces mean "0 or more".)
-
-1. With ``-XExtendedLet``, change the grammar to be ::
-
-     funlhs → var apats1
-            | pat varop pat
-            | '(' funlhs ')' apats
-
-     apats1 → apat
-            | apat apats
-            | '(' 'let' decls ')' apats
-
-     apats →
-           | apats1
-
-     lexp → '\' apats1 '->' exp
-          | …
-
-     lpat → apat
-          | '-' (integer|float)
-          | gcon apats
-
-   This allows phrases like ``f x (let y = g x x) (frob y -> True) = ...``, where we can include
-   a ``let`` construct in the middle of a list of patterns. The pattern grammar itself is unaffected.
-
-2. Any entities bound in ``decls`` scope over the same region of the program
-   that pattern-bound variables scope over, with the addition of the ``decls``
-   themselves (that is, the declarations can be recursive).
-
-Examples
-~~~~~~~~
-
-.. _let-in-pattern-example:
-
-1. Instead of ::
-
-     f :: Maybe Bool -> Bool -> Bool
-     f (x :: Maybe b) (y :: b) = ...
-
-   we can write (Alternative 1) ::
-
-     f :: Maybe Bool -> Bool -> Bool
-     f (let type b = Bool in x) (y :: b) = ...
-
-   or (Alternative 2) ::
-
-     f :: Maybe Bool -> Bool -> Bool
-     f (let type b = Bool) x (y :: b) = ...
-
-   Note that the ``b`` is in scope in the type signature for ``y``.
-
-   If we instead say (Alternative 1) ::
-
-     f (let type b = _ in (x :: Maybe b)) (y :: b) = ...
-
-   or (Alternative 2) ::
-
-     f (let type b = _) (x :: Maybe b) (y :: b) = ...
-
-   now the choice ``b ~ Bool`` is inferred, but we have an explicit binding
-   site for ``b``, in accordance with the `Explicit Binding Principle`_.
-
-#. Instead of ::
-
-     f x y z (frob x y z -> True) (frob x y z -> False) = ...
-
-   we can write (Alternative 1) ::
-
-     f x y z (let test = frob x y z in (test -> True)) (test -> False) = ...
-
-   or (Alternative 2) ::
-
-     f x y z (let test = frob x y z) (test -> True) (test -> False) = ...
-
-   avoiding some repetition.
-
-Effects
-~~~~~~~
-
-1. In concert with binding type variables in a ``let``, this helps uphold the `Explicit Binding Principle`_.
-   Without this feature, a line such as ``f (x :: Either b b) = ...``
-   has no binding site for ``b``. (Alternatively, we could say that the first ``b`` is a binding site,
-   but then we lose the `Local Lexical Scoping Principle`_, as the binding-site-vs-occurrence distinction depends on what is in scope.)
-
-   The ability to bind variables that would otherwise be pattern-bound is why this feature
-   allows binding type variables.
-
 ``let`` in types
 ----------------
 
-This part of the proposal allows ``let`` to be used in types. This part goes
-beyond any previous proposal.
+This part of the proposal allows ``let`` to be used in types.
 
 Motivation
 ~~~~~~~~~~
@@ -244,7 +107,7 @@ Motivation
 Proposed Change Specification
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-1. With ``-XExtendedLets``, expand the grammar for types to include the following::
+1. With ``-XExtendedLet``, expand the grammar for types to include the following::
 
      type → 'let' tdecls 'in' type
 
@@ -282,10 +145,6 @@ Effects
    aggressively. We can think about ways to preserve synonyms as we gain experience
    with the feature.
 
-#. This part of the proposal does not directly serve any of the principles outlined
-   at the top of this proposal, but now seems a convenient time to introduce this
-   extension, which should be relatively easy to implement.
-
 Effect and Interactions
 -----------------------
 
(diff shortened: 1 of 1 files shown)


konsumlamm:
### 1. `let`-binding types

It would be useful to also add some examples for `let`-binding types. Am I right in assuming that it adds the possibility to define type aliases in `let` expressions and `where` clauses (and nothing more)? Like `let type ... = ... in ...` or `... where type ... = ...`? If so, I think that would be a nice little feature, but it doesn't seem terribly useful to me (perhaps some examples could convince me otherwise). What would be more useful imo would be supporting local data types, e.g. a special result type for a local helper function (something like [this](https://github.com/lspitzner/pqueue/blob/1a1239abaacdcc1dfd18231aa74c55cdd79bf416/src/BinomialQueue/Internals.hs#L327)), but that probably has some issues (I'm not suggesting to do this, just poiting out that I'd find something like this more useful).

### 2. `let` in patterns

As I said in #448 (can't find my comment), I think it's a bad idea to use `let` here. The syntax for normal `let`s is `let <bindings> in <expression>`, where the bindings are only visible in the expression. However, in both alternatives, the variables bound by the `let` are also available in the following patterns. The suggestion after "If we instead say (Alternative 1)" fixes this, but now there are multiple patterns after the `in`, which looks very odd imo. I don't find the examples very convincing either, for example 2.4.2 could be rewritten using `where test = frob x y z`, which I think would be a lot more readable.

### 3. `let` in types

This looks fine to me. Perhaps there should be some motivation for using `let c = ... in ...` instead of `let type c = ... in ...` (probably to eventually support normal expressions on the type level, but it might seem odd compared to `let type c = ... in ...` proposed in 1)? Ii seems this goes one step into the direction of supporting normal expressions on the type level, without any future incompatibilities, is 
that correct?

---

I think the listed drawback is on point. Overall, I'm currently leaning towards "this is too much new weird syntax for very little gain".

AntC2:
Thank you @goldfirere this is great! Seldom have I felt a proposal is immediately natural. Hopefully we can eventually throw away the confusing `forall` scoping from signatures with `XScopedTypeVariables`.

`let type a = ...` makes immediate sense, whether or not I wish to allow punning.

I prefer your Alternative 2 in Patterns:

```haskell
        f :: Maybe Bool -> Bool -> Bool
        f (let type b = Bool in x) (y :: b) = ...                          -- Alternative 1

        f (let type b = Bool) x (y :: b) = ...                             -- Alternative 2
```

Because I'm used to reading `let z = ... in e` as meaning `z` has scope only within `e`. Whereas free-standing `let` (as in guards) scopes from that point forwards. Also Alternative 1 appears to be asymmetric whereas this under Alternative 2

```haskell
f (let type b = Bool)  (x :: Maybe b) (y :: b) = ...
```

treats `x` and `y` equally, and shows `b` bound to both vars.

You say under Costs and Drawbacks "This introduces yet more syntax to read", but `let` syntax is already in play in guards. I suppose what's proposed is a little odd in that what's declared escapes the closing parens `(let ... = ... )`, whereas guards don't use parens. Would anyone prefer less parenthetical syntax?

One q: we do currently allow this in patterns (`b, r` not in scope, and say there's no signature given for `g`):

```haskell
g (x :: b) (y :: b) :: r= ...
```

This binds `b` to whatever type gets inferred for `x`, and insists `y` be the same type (or _vice versa_); and binds `r` to the result type. But offends against the 'Explicit Binding Principle'. I feel I should write as:

```haskell
g (let type b; type r) (x :: b) (y :: b) :: r = ...
```

danidiaz:
In section [3.3](https://github.com/goldfirere/ghc-proposals/blob/extended-let/proposals/0000-extended-let.rst#id7):

    Instead of

        f :: forall a b. (c ~ Very Big Type a b) => c -> c -> c

Shouldn't it be

        f :: forall a b c. (c ~ Very Big Type a b) => c -> c -> c
?

I surmise that in the `let`-using version the `f :: forall a b.` is correct and the `c` wouldn't be available for type applications there.

Ericson2314:
Parts 1 and 3 are wondeful! But huge thumbs down to part 2.

A cardinal rule of pattern matching is:

> if a (closed) expression is exactly the same as a pattern, the pattern matches the expression

This is the foundation around which everything about pattern matching is built!

But
```haskell
 f (let type b = Bool in x) (y :: b)
```
as an expression has nothing to do with
```haskell
 f (let type b = Bool in x) (y :: b)
```
as a pattern.

Indeed even the renaming of the `let` binding *escaping* the `(..)` is extremely counter-intuitive.

------

I strongly urge you to split this proposal --- again, I know.

Parts 1 and 3 are classic examples of Dependent-Haskell-style asking "Wouldn't be easier/nicer if the pattern and expression syntax was more similar?". The answer is "yes!" and they are great natural additions to the language with a clear and obvious interpretation.

Parts 2 is not an inevitable design, because it isn't actually making pattern syntax closer to expression syntax given the syntax is interpreted completely differently. Even if one doesn't think the semantics here are the *wrong* ones, there are other semantics no further from expression syntax we might consider also. It simply is a less constrained design space.

Part 2 thus belongs in a separate proposal; it shouldn't hold up parts 1 and 3 in a traffic jam.

AntC2:
@Ericson2314 re Part 2, you seem not to have considered Alternative 2 -- which would not be valid in an expression, so no risk of confusion. Would using that alternative change your response?

> Part 2 thus belongs in a separate proposal; ... hold up ... in a traffic jam.

If you insist on splitting, my priorities would be: split parts 1 and 3 into a separate proposal; hold them up for a very long time. I want to get rid of the `forall` semi-global scoping from `-XScopedTypeVariables`.

goldfirere:
It seems that @Ericson2314 might prefer Alternative 2 of Part 2.

I am not willing to split up this proposal. (I think we spend too much time debating the proper formatting and presentation of proposals, and this busywork slows us down.) I *am* happy, though, for parts of this proposal to be accepted and other parts rejected/delayed. At that point, like with #448, I will reformat before merging.

gbaz:
Could an example be added to the first proposed change (let-binding types)?

Ericson2314:
From alternative 2:

```haskell
f x (let y = g x x) (frob y -> True) = ...
```

the `y` in `frob y` is a use or bind?

> I am happy, though, for parts of this proposal to be accepted and other parts rejected/delayed. At that point, like with https://github.com/ghc-proposals/ghc-proposals/pull/448, I will reformat before merging.

OK that is fine. "Lazy vs eager" splitting is no matter to me, as long as this stuff is not taken as a package deal.

goldfirere:
For the record, I don't really like either of the alternatives in the `let`-in-patterns bit.

My goal in that section is to retain the current behavior of pattern signature bindings, but to uphold the various scoping principles. Right now, we have

```hs
f :: Maybe (Either Int Bool) -> _
f (x :: a) (y :: a) = (x, y)
```

Here, `a` will be bound to `Maybe (Either Int Bool)`. But this happens via a pattern signature binding, which is problematic (because these currently depend on checking what variables are already in scope). I would rather something like

```hs
f :: Maybe (Either Int Bool) -> _
f (let type a = _) (x :: a) (y :: a) = (x, y)
```

Now, the binding for `a` is explicit: `a` will shadow any in-scope `a` that's already around. (This change would be *optional*, for those who want to be more explicit. People who prefer the old behavior will need to take no action to retain the old behavior.) But doing this requires the ability to bind a variable in the middle of a list of patterns. (We can't do it before the `f`, because the wildcard might want to capture a locally bound type variable.) Yet I can't find a better syntax for it all. :(

I guess one possibility is

```hs
f :: Maybe (Either Int Bool) -> _
f = let type a = _ in \cases
  (x :: a) (y :: a) -> (x, y)
```

Or even better

```hs
f :: Maybe (Either Int Bool) -> _
f = \cases
  (x :: a) (y :: a) -> (x, y)
  where
    type a = _
```

I actually quite like that last one. So maybe, since we have `\cases`, we don't need the `let`-in-patterns idea at all? Curious for some feedback here before I update the proposal.

Ericson2314:
> For the record, I don't really like either of the alternatives in the let-in-patterns bit.

OK whew :)

> I actually quite like that last one.

Me too!

I had just the `let type a = _ in` mind originally, but agreed the `where` which allows one to not sweat the exact details of the scoping is better.

> So maybe, since we have \cases, we don't need the let-in-patterns idea at all?

<3

tek:
what about top level let?
```haskell
f :: Maybe (Either Int Bool) -> _
let type a = _ in f (x :: a) (y :: a) = (x, y)
```

Ericson2314:
@tek The problem is that the the inferred definition of `a` depends on freshly quantified type variables. So `f` still needs to be the leftmost identifier.

Concretely:
```haskell
f :: forall x y. Maybe (Either x y) -> _
let type a = _ in f (x :: a) (y :: a) = (x, y)
```
becomes
```haskell
f :: forall x y. Maybe (Either x y) -> Maybe (Either x y) -> (Maybe (Either x y), Maybe (Either x y))
f = \ @x @y -> let type a = Maybe (Either x y) in \(x :: a) (y :: a) -> (x, y)
```

goldfirere:
OK. I've removed `let`-binding in patterns. It was just too ugly.

LeanderK:
As a GHC user, I am very excited for the proposal! But I have a more general question, since I am somewhat following the evolution of GHC. I am wondering whether I can use let-in-types in the foreseeable future, because there are a few similar proposals I am waiting for. For example [type-lambda](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0155-type-lambda.rst) proposal, which was accepted 3 years ago (or the Visible forall from last year). I am not sure whether there's progress or not. I would really like to have let in types, but there's no implementation plan which makes me quite pessimistic that it's going to be ready in the next years. I feel like the pace of accepting proposals is much faster than implementing them, so they pile up.

Also, with regard to the proposal I wonder: What I would really like is binding types in where-clauses to declutter the type-signatures. For me it seems like a similar idea, is it related? Does implementing one make the other more feasible?

AntC2:
> 4   Costs and Drawbacks
> This introduces yet more syntax to read, understand, parse, have errors about, etc. This drawback alone may make this proposal not worth accepting.

Yes, this. All I'm seeing is yet more syntax, and more verbose, for doing something we already have. Not worth it.

Ericson2314:
@LeanderK part of the bottleneck was that we had too many proposals balkanizing the spec. https://github.com/ghc-proposals/ghc-proposals/pull/448 fixes that. Implementation has started in https://gitlab.haskell.org/ghc/ghc/-/merge_requests/8629/diffs

Ericson2314:
@goldfirere do you want to also propose `= _` doing what we expect? Or is that to be a separate proposal?

Ericson2314:
@goldfirere Do you want to finish this one? If you want a co-author I would be happy to do mundane tasks like fixing a few formatting errors, and also take a crack at formalizing `let type foo = _` as mentioned above.

goldfirere:
Thanks for pinging here. I'll put this back in my queue, hopefully to address this week.

goldfirere:
I've pushed an implementation plan (though it's oddly not showing up in this PR -- it does show up on the Rendered link above). I think all the other outstanding parts are resolved now. I'd like to submit to the committee @nomeata.

jvanbruegge:
Just a quick note: Most of the proposal only talks about lets, so I originally wanted to ask if this also works with `where`. There is a single example later on that uses `where`, so I assume it does, but it might be good to specify that explicitly

jvanbruegge:
In addition to that, is `where` allowed in types as well? If yes I would assume explicit foralls to scope over the where clause similar to function arguments in top level definitions. E.g.:
```
foo :: forall a. Foo => a -> Bool
    where Foo = (Ord a, Enum a)
```
In fact, this would be similar to [Rust](https://doc.rust-lang.org/rust-by-example/generics/where.html) (albeit more general)

goldfirere:
Good point @jvanbruegge. I've fixed this.

Ericson2314:
If you do want to formalize `type foo = _` so that part of the motivation is directly actionable, I was thinking that it basically introduces a unification variable to define `foo` with, but *also* the restriction that the unified solution only uses variables that are in quantified at that spot (which I suppose can be done with those "levels", right?)

In other words:
```haskell
id :: forall a. a -> a
id = let type b = _ in \ @a ((x :: b) :: a) -> (x :: b) :: a
```
should be *rejected* because the `let type b = _ in` is outside the `\ @a ->`

goldfirere:
Good point, @Ericson2314. I've updated the proposal to mention and exemplify this nuance.

Ericson2314:
Thanks! And sorry I missed that wildcards were mentioned already!

proposals/0000-extended-let.rst
```diff
@@ -0,0 +1,236 @@
...
+     f = \cases
+       Nothing -> Just (Left 5)
+       x :: a -> x :: a
+       where
+         type a = _
+
+   While this second version is more verbose, it makes the binding site of ``a`` explicit, the
+   linchpin of the `Explicit Binding Principle`_.
+
+#. The existing trick of using e.g. ``f :: forall a. (a ~ SomeBigType). ... a ... a ... a `` to
```
Ericson2314, line 204:
```suggestion
#. The existing trick of using e.g. ``f :: forall a. (a ~ SomeBigType). ... a ... a ... a`` to
```
I think that fixes a rendering error.

proposals/0000-extended-let.rst
```diff
@@ -0,0 +1,236 @@
...
+1. This step further unifies term-level and type-level syntax, at low cost.
+
+#. An initial version of this feature will likely want to expand the synonyms
+   aggressively. We can think about ways to preserve synonyms as we gain experience
+   with the feature.
+
+#. Over time, we might decide to add ``where`` in more places.
+   
+Effect and Interactions
+-----------------------
```
Ericson2314, line 179:
```suggestion
-----------------------

.. _`Explicit Binding Principle`: ../principles.rst#explicit-binding-principle
.. _`Syntactic Unification Principle`: ../principles.rst#syntactic-unification-principle
```
So we don't have dangling links below.

Ericson2314:
3 little things, but after that I hope this can go to the committee!

goldfirere:
Yes, please submit to the committee @nomeata.

simonpj:
I want to argue against this proposal

1. It is poorly motivated.  Of the three motivations in 1.1
   - (1) lacks evidence, and any case (as the bullet shows) we have a workaround. (Type synonyms to abbreviate are another useful workaround.)
   - (2) isn't a motivation.  It's explains that pattern signatures introduce type variables that stand for types (not type variables), but does not motivate why we need to do so in other contexts.
   - (3) is very generic ... "gets us closer to types as terms"

2. It's not clear to me whether this is intended to affect Core types, or only source-langauge types.  If the former, then I'm much more relaxed -- it's just some surface syntax for a local synonym.  (But I still think it's a feature that is poorly motivated.)  If the latter I have lots of questions, some of which I have explored with Richard.

So it just adds complexity with very litle payoff.

In discussion with Richard, we did find one possible payoff.  Currently pattern signatures are funny: you can only tell whether  `(\(x::a) -> blah)` brings `a` into scope if you know whether or not `a` is already in scope.  Not a beautiful thing. With this proposal, Richard says you can write
```
   type a = _ in (\(x::a) -> blah)
```
(using a type wildcard) and now it's clear that `a` is bound by the `type` declaration.  Well, yes; but it is still not a thing of beauty, and it seems like using a sledgehammer to crack a nut. An alternative would be to abolish pattern signatures --- or at least abolish the rule that allows a pattern signature to bring a variable into scope.  *That rule was only present to allow us to give a name to existential type variables.*  E.g. 
```
data T = MkT [a] (a->Int)
f :: T -> [Int]
f (MkT (xs :: [a]) f) = let mf :: [a] -> Int
                            mf = map f
                        in mf xs
```
Here the pattern signature on `xs` brings `a` into scope, so that it can be mentioned in
the type signature for `mf`.  In the past there was no other way to do this. But now we
can say
```
f :: T -> [Int]
f (MkT @a xs f) = let mf :: [a] -> Int
                      mf = map f
                  in mf xs
```
So we could, if we chose, deprecate and ultimately abolish the ability for
pattern signatures to bring a new type variable into scope.  Instead of *adding*
complexity to the language, let's *remove* it.

nomeata:
❤️ to deprecating bringing type variables into scope via signatures in patterns

adamgundry:
Would it be worth considering the two features being introduced here separately? Having `let` in types seems like a straightforward application of the [Syntactic Unification Principle](https://github.com/ghc-proposals/ghc-proposals/blob/extended-let/principles.rst#syntactic-unification-principle) (at least when binding a tyvar). Whereas local type synonym declarations would be new at both the term and type levels, and introduce new complications (in particular the new treatment of wildcards, of which I'm a little suspicious).

adamgundry:
Shepherd here: I plan to give @goldfirere an opportunity to respond to the recent discussion and let that reach a conclusion before bringing this to the committee.

goldfirere:
I'm not quite sure how we'd go about removing the ability for pattern signatures to bring variables into scope. I think this feature has somewhat wide uptake. I think -- though I'm not sure -- that `singletons` makes essential use of this feature. @RyanGlScott will be able to answer that point authoritatively.

Beyond that, I agree that this proposal is weakly motivated. It brings us to the principles outlined in #448, but I tend to agree that the complexity isn't worth it. I will close.

RyanGlScott:
> I'm not quite sure how we'd go about removing the ability for pattern signatures to bring variables into scope. I think this feature has somewhat wide uptake. I think -- though I'm not sure -- that `singletons` makes essential use of this feature.

Indeed, `singletons-th` often generates code of the form:

```hs
sFoo :: Sing '(Succ x, x) -> Sing (Succ x)
sFoo (STuple2 (sY :: Sing y) _) = ... {- type involving y -} ...
```

Here, `y` stands for `Succ x`. (Example taken from https://github.com/goldfirere/singletons/issues/318#issuecomment-375922850)

I doubt that this is the only way one could generate this code, however. Provided that there is a suitable alternative, I'd be happy with deprecating the ability to bring type variables into scope via signatures in patterns.

Ericson2314:
I am sad to see this closed. I wouldn't mind dropping the wildcard part if we think `@` binding is good enough, but type synonyms in `where` clauses I have wanted on many occasions so I can close over parameters. Type synonyms at the top level aren't a good substitute because "lambda lifting" one needs to do can cancel out the tensity gains of binding a name in the first place!

> Having let in types seems like a straightforward application of the [Syntactic Unification Principle](https://github.com/ghc-proposals/ghc-proposals/blob/extended-let/principles.rst#syntactic-unification-principle) (at least when binding a tyvar).

Indeed, IMO a big point of having the principles is so we don't have to repeatedly restate the same motivation again and again. If consistent syntax between types and terms isn't motivation enough, we should rescind the principle as not one the committee actually believes in.

VitWW:
I found this Proposal accidentally.
I described a bit similar ideas in my #646 (Already has closed) and #643 (Already has switched to Draft) proposals.
Maybe they could be interesting for you. 

Typical code  from your proposal examples with mine proposal ideas:
```haskell
--sFoo :: Sing '(Succ x, x) -> Sing (Succ x)
sFoo :: forall x. forsome y <- Succ x. Sing '(y, x) -> Sing y
sFoo (STuple2 (sY :: Sing y) _) = ... {- type involving y -} ...

id :: forall a. forsome b <- a. a -> b
id @a (x :: b)  = x

f :: Maybe (Either Int (Char, Double, Bool)) -> Maybe (Either Int (Char, Double, Bool))
f = \cases
  Nothing -> Just (Left 5)
  (x :: forsome  a. a) -> x :: a
```

PR #526, opened 2022-07-28
Propose applicative comprehensions.

strake:


strake:
See also [earlier proposal](https://gitlab.haskell.org/ghc/ghc/-/wikis/applicative-comprehensions).

proposals/0000-applicative-comprehensions.rst
```diff
@@ -0,0 +1,151 @@
+Notes on reStructuredText - delete this section before submitting
+==================================================================
```
googleson78, line 2: Seems like this should be deleted

konsumlamm:
Can you please add a link to a rendered version of the proposal?

proposals/0000-applicative-comprehensions.rst
strake: Thanks! Done

proposals/0000-applicative-comprehensions.rst
```diff
@@ -0,0 +1,151 @@
...
+Also, discuss possibly contentious interactions with existing language or compiler
+features. Complete this section with potential interactions raised
+during the PR discussion.
+
+Costs
+-------------------
+
+This extension has been implemented and tested (in a feature branch).
+
+Other than the extension flag, it touches 3 modules in the compiler: one each in the typechecker, the renamer, and `HsToCore`. Given that, the author estimates he upkeep costs should be fairly low.
```
konsumlamm, line 105:
```suggestion
Other than the extension flag, it touches 3 modules in the compiler: one each in the typechecker, the renamer, and `HsToCore`. Given that, the author estimates the upkeep costs should be fairly low.
```

strake:
> Can you please add a link to a rendered version of the proposal?

Done

strake:
For me, on the contrary, code upkeep/readability is a major reason for me to want `ApplicativeComprehensions`. In cases where i am **already** naming values (as λ-binders), `ApplicativeComprehensions` leads to much lower upkeep costs, as the reader need not count arguments to figure out which is which. (And in other cases, a comprehension may not be appropriate at all!)

I've not used idiom brackets myself. Indeed, i've not used Idris much. I do think they could be very useful. But this and that proposal are not mutually exclusive, and furthermore this proposal:
- adds no new syntax
- seems (to me) a natural (in the lay sense) extension of a feature we already have
- already has a tested and working implementation ☺
So let's not let the perfect be the enemy of the good, here.

Ericson2314:
> adds no new syntax

It is still allowing existing syntax to mean new things. That isn't necessarily better?

> seems (to me) a natural (in the lay sense) extension of a feature we already have

It is, but if that feature is bad, then it is more work to deprecate than before

> So let's not let the perfect be the enemy of the good, here.

What pressing need is there when one can already use do notation? Syntax (including new meaning to existing syntax) has a cost. We get way more syntax proposals than anything else, and I think the bar should be fairly high.

konsumlamm:
> Thus, I think `ApplicativeDo` is an awkward middle ground --- it's both too magical and yet not concise enough to be great magic. I therefore don't think we should expand it into more parts of the language.

I think `ApplicativeComprehensions` would be better than `ApplicativeDo`, since it would be less magical. There is no ad-hoc check if the last statement is `pure ...` or `return ...` (by the nature of comprehensions), the only thing that needs to be checked is that no variable bound by a generator appears in another generator expression.

L-as:
I think this is interesting, but wouldn't something like `QualifiedComprehensions` make this redundant?

strake:
@L-as This changes the actual structure of the desugarized expression; i think `QualifiedComprehensions` would merely have some names in the expression resolve to other terms. So `QualifiedComprehensions` alone would not subsume `ApplicativeComprehensions`.

L-as:
Good point, and I imagine you could use both together theoretically.

nomeata:
@strake, are you by any chance attending Haskell Symposium? If so, would you like to present this proposal at the [GHC Proposals session](https://icfp22.sigplan.org/home/haskellsymp-2022#GHC-Proposals-Session)?

strake:
Summary of conversation so far:
- Optimal desugaring algorithm is O(n³).
- Desugaring may seem unpredictable.
- `ApplicativeComprehensions` is less magical than `ApplicativeDo`.
- `QualifiedComprehensions` would not subsume `ApplicativeComprehensions`, since the latter changes the structure of the desugaring, not merely the name resolution.

@nomeata I think this is ready for review.

konsumlamm:
> Summary of conversation so far:
> 
>    * Optimal desugaring algorithm is O(n³).
> 
>    * Desugaring may seem unpredictable.

These two points only apply to `ApplicativeDo` as far as I can tell. The desugaring for `ApplicativeComprehensions` seems pretty trivial and predictable to me, `[expr | x1 <- expr1, x2 <- expr2, ..., xN <- exprN]` simply desugars to `(\x1 x2 ... xN -> expr) <$> expr1 <*> expr2 <*> ... <*> exprN`, or am I missing something?

L-as:
This is also precisely why this proposal is much better than ApplicativeDo, and you could even deprecate ApplicativeDo IMHO.

tomjaguarpaw:
@L-as `ApplicativeDo` needs a special rule to deal with `pure` at the end. In what other ways is it worse than this proposal?

L-as:
@tomjaguarpaw see commentby Ericson above

tomjaguarpaw:
Do you mean this comment? https://github.com/ghc-proposals/ghc-proposals/pull/526#issuecomment-1198324299

If so I'm not sure I understand. That doesn't seem to provide concrete reasons why `ApplicativeDo` is harder to desugar than this proposal.

proposals/0000-applicative-comprehensions.rst
```diff
@@ -0,0 +1,178 @@
...
+
+Drawbacks
+-------------------
+
+Some criticisms of `ApplicativeDo` also apply here. For example:
```
konsumlamm, line 126:
```suggestion
Some criticisms of `ApplicativeDo` also apply here. For example:

```
Currently, the list is not rendered correctly.

strake:
> The desugaring for `ApplicativeComprehensions` seems pretty trivial and predictable to me, `[expr | x1 <- expr1, x2 <- expr2, ..., xN <- exprN]` simply desugars to `(\x1 x2 ... xN -> expr) <$> expr1 <*> expr2 <*> ... <*> exprN`, or am I missing something?

I believe the same proof (about its complexity) is valid for `ApplicativeComprehensions` as well as `ApplicativeDo`. I merely re-used the algorithm for the latter for the former; you can read about it [here](https://simonmar.github.io/bib/papers/applicativedo.pdf).

konsumlamm:
I read through the paper and what I missed is that `ApplicativeDo` (and by extension `ApplicativeComprehensions`) tries to use `Applicative` where possible, even if the whole expression needs a `Monad` constraint. Note that the optimal algorithm isn't enabled by default (https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/applicative_do.html#ghc-flag--foptimal-applicative-do).

tomjaguarpaw:
Ah, so `ApplicativeDo` has extra complexity around interactions with monadic statements in the same `do` block, whereas `ApplicativeComperhensions` has no such issue and can use a simpler desugaring?

simonmar:
(committee shepherd here)

Some comments having read the proposal and the discussion.

* The proposal itself is a relatively straightforward extension. Given that we already have `ApplicativeDo`, it should work in the same way - unpredictable `O(n^3)` desugaring algorithm and all. This is what the proposal is suggesting.
* Arguably `ApplicativeDo` should have been much simpler, supporting only cases where a `do` can be desugared into an `Applicative` expression, or possibly supporting explicit nested Applicatives with the extended syntax from the paper, but the unpredictable magic desugaring moved into a compiler plugin for use-cases like Haxl. In retrospect I think that might have been a better design. (I don't know whether it's possible with plugins as they stand, but if not then presumably the plugin interface could be extended). Still, that's something for future proposals.
* Regarding this proposal:
  * The typing rule is oversimplified because it doesn't take into account Monad/Applicative/Functor constraints. Suggest just removing it. The typing rule is exactly what you would get by typechecking the desugared code.
  * Are there any interactions with `ParallelListComprehensions` and `TransformListComp`?

simonpj:
I have always found ApplicativeDo confusing (even though I co-authored the paper about it!) so I'd welcome anything that makes it simpler and more uniform.   I'm concerned that we have around [20 open tickets around ApplicativeDo](https://gitlab.haskell.org/ghc/ghc/-/issues/?sort=updated_desc&state=opened&label_name%5B%5D=ApplicativeDo&first_page_size=20), and one who loves the feature enough to work on them.

So while I'd welcome something that made it simpler and more uniform, I'm anxious about extending a feature that is already rather wobbly.

Effort spent in consolidating the status quo (including simplifying it) would build confidence for future extensions.

goldfirere:
I just want to amplify this point of @simonpj's:

> Effort spent in consolidating the status quo (including simplifying it) would build confidence for future extensions.

It's always tempting to add something new to our wonderful language (I speak from experience), but I agree that it's probably best to build a better foundation here first. This isn't to say that there needs to be 0 bugs in the original applicative-do before extending it, but showing some sustained love toward what we have builds confidence that similar love will greet the new feature after it's implemented.

L-as:
> * Arguably `ApplicativeDo` should have been much simpler, supporting only cases where a `do` can be desugared into an `Applicative` expression, or [...]

I completely agree with this. I think using the same syntax for both applicative and monadic do/comprehensions is problematic,
and if you had a syntax that _only_ allowed independent binds, it seems the desugaring algorithm would also be O(n).
I think the use-case for automagically making code more parallel by supporting mixed use isn't that big, especially when the algorithm is somewhat opaque and unpredictable.

One possible solution could be using `QualifiedDo`, checking if the module in question exports a `>>=`, and if so, use a _purely monadic_ desugaring, otherwise _purely applicative_, i.e. no dependent binds.
You could also use this for a hypothetical `QualifiedComprehensions` language extension.
I'm not sure this is a good solution though, perhaps it would be better if there was a completely different syntax, e.g. `ado`, or `[[x | y <- z]]`.

googleson78:
> checking if the module in question exports a >>=, and if so, use a purely monadic desugaring, otherwise purely applicative, i.e. no dependent binds.

Is there some benefit to doing this over simply using a "library implementation" to disallow using `(>>=)`?:


<details>
<summary>Implementation</summary>

```haskell
-- Qualified.hs
{-# LANGUAGE ApplicativeDo #-}
{-# LANGUAGE QualifiedDo #-}

module Qualified where

import qualified Applicative

x = Applicative.do
  v <- getLine
  u <- getLine

  Applicative.pure $ v ++ u

-- Applicative.hs
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE MultiParamTypeClasses #-}

module Applicative where

import qualified Control.Applicative
import qualified Control.Monad
import qualified Data.Functor
import GHC.TypeLits

return :: Applicative f => a -> f a
return = Control.Applicative.pure

pure :: Applicative f => a -> f a
pure = Control.Applicative.pure

(>>=) :: NoCall => ()
(>>=) = error "bug"

(<*>) :: Applicative f => f (a -> b) -> f a -> f b
(<*>) = (Control.Applicative.<*>)

(>>) :: Applicative f => f a -> f b -> f b
(>>) = (Control.Applicative.*>)

fmap :: Functor f => (a -> b) -> f a -> f b
fmap = Data.Functor.fmap

class NoCall

instance TypeError (Text "no binds allowed") => NoCall
```
</details>

simonmar:
What do we want to do here? There seems to be general agreement that `ApplicativeDo` should be simpler but nobody to do the actual work in making that happen; the existing transformation would have to be split up and the optimal rearrangement stuff moved into a plugin which I suspect is a non-trivial amount of work.

Meanwhile the proposal here is actually not that big of a change, and wouldn't make it harder to simplify `ApplicativeDo` later. I don't think it would be problematic to accept it. But I agree all things being equal, time is better spent on making what we have less wobbly.

Ericson2314:
> But I agree all things being equal, time is better spent on making what we have less wobbly.

https://discourse.haskell.org/t/monadic-bang-a-plugin-for-more-concise-do-block-notation-inspired-by-idris/5556/15 does this have similar issues?

simonpj:
>  The desugaring for ApplicativeComprehensions seems pretty trivial and predictable to me, [expr | x1 <- expr1, x2 <- expr2, ..., xN <- exprN] simply desugars to (\x1 x2 ... xN -> expr) <$> expr1 <*> expr2 <*> ... <*> exprN, or am I missing something?

Is this claim true?  It seems to contradict other comments, such as

> Given that we already have ApplicativeDo, it should work in the same way - unpredictable O(n^3) desugaring algorithm and all

If the comprehension form simplified the design, that would be a Big Win.

> Arguably ApplicativeDo should have been much simpler, supporting only cases where a do can be desugared into an Applicative expression, or possibly supporting explicit nested Applicatives with the extended syntax from the paper, but the unpredictable magic desugaring moved into a compiler plugin for use-cases like Haxl

Could we consider given ApplicativeComprehensions the "simpler, predicatable design" that Simon describes here; and then in due course deprecating the complexities of ApplicativeDo?

If we think the ApplicativeDo design is too complicated (and I think there is some consensus on that) it would seem a pity to build more structure on top of a foundation that we regard as a bad idea.

----------------

I know that, through one lens, this proposal is only allowing a different syntactic presentation of something we already have, and can therefore be regarded as an easy win.  But I would love to use the energy here as a way to improve what we have rather than merely continue down a bad road.

JakobBruenker:
> Is this claim true? It seems to contradict other comments

It seems very likely to me that any do-block that can be desugared into an applicative expression only with `-foptimal-applicative-do` could also be written as a do block

konsumlamm:
> > The desugaring for ApplicativeComprehensions seems pretty trivial and predictable to me, [expr | x1 <- expr1, x2 <- expr2, ..., xN <- exprN] simply desugars to (\x1 x2 ... xN -> expr) <$> expr1 <_> expr2 <_> ... <*> exprN, or am I missing something?
> 
> Is this claim true? It seems to contradict other comments, such as

No, I corrected myself in https://github.com/ghc-proposals/ghc-proposals/pull/526#issuecomment-1272600769 (I realize I should have edited my original comment, I added a note now). I originally thought that `ApplicativeDo` (and by extension the proposed `ApplicativeComprehensions`) only applied to `Applicative`-only expressions, instead of also working when a part of it requires `Monad`.

> Could we consider given ApplicativeComprehensions the "simpler, predicatable design" that Simon describes here; and then in due course deprecating the complexities of ApplicativeDo?

Then my statement about trivial and predictable desugaring should apply. I think this would be very nice, but it might be confusing to people expecting it to work like `ApplicativeDo`.

simonpj:
> Then my statement about trivial and predictable desugaring should apply. I think this would be very nice, but it might be confusing to people expecting it to work like ApplicativeDo.

It might -- but my point is that if that is the direction in which we want to move the language, let's go for it.  (But we need to debate the if.)

simonmar:
The discussion so far has raised points about `ApplicativeDo` being complex and having unpredictable behaviour (in its defense, this is by design - the original motivating use case absolutely depends on doing the nesting optimisations described in the paper). But taking a step back, a better overall design would be to split the extension in two:

1. A simple desugaring that applies only to do-blocks that can be desugared into purely Applicative expressions
2. The more complex optimisation that chooses a nesting using the O(n^3) algorithm

Many use cases, such as the option-parsing examples only need (1). (2) should really be a compiler plugin, but given the complexity of the implementation it would likely require significant changes to the plugin interfaces. 

You could envisage (1) being something we would put in GHC20xx, whereas (2) definitely isn't.

Perhaps there's a (1.5) where we add some syntax to do-blocks to indicate applicative nesting, as suggested in the paper. That's also predictable and straightforward (but nobody is asking for it).

Anyway, returning to this proposal, what are the options?

1. Accept this proposal as is; it doesn't actually make things worse, but it does build on somewhat shaky ground
2. Reject the proposal until we've sorted out ApplicativeDo.
3. Request that the proposal only does the (1)-style desugaring, making it simple(r) to describe, and then accept it. But that would mean applicative comprehensions don't behave like applicative do-expressions.

Keen to hear thoughts.

nomeata:
Based on just that summary, I think (1) is reasonable - even if the status quo has issues, extending it consistently to a very related syntactic construct can't hurt much.

adamgundry:
> 3. Request that the proposal only does the (1)-style desugaring, making it simple(r) to describe, and then accept it. But that would mean applicative comprehensions don't behave like applicative do-expressions.

I'm attracted to the simplicity of this approach. We could simply say that `ApplicativeComprehensions` will produce an `Applicative` expression provided there are no data dependencies in the comprehension, and a `Monad` expression otherwise. That doesn't preclude doing something more clever later, but it provides an alternative to `ApplicativeDo` that is (hopefully) less controversial.

@strake how would you feel about such a simplication? Would it still be adequate for your use cases?

(We should also specify what `ApplicativeComprehensions NoMonadComprehensions` means; I think it could emit an error if a comprehension has a data dependency.)

simonpj:
> I'm attracted to the simplicity of this approach. We could simply say that ApplicativeComprehensions will produce an Applicative expression provided there are no data dependencies in the comprehension, and a Monad expression otherwise. That doesn't preclude doing something more clever later, but it provides an alternative to ApplicativeDo that is (hopefully) less controversial.

We get this
* (3) treats comprehension notation and do-notation differently, losing the curent consistency of the two
* But (3) also moves in the direction of distinguishing "simple, predictable" from "complex, less predictable" applicative desugaring, which is arguably something we want to do.

I would love to hear from folk who actively use applicative-do.  Our judgement will be better if informed by actual experience.  @strake I guess that includes you.  But could we seek some more users?  E.g. we could grep for ApplicativeDo in hackage and write to the maintainers of packages that use it?

 
And also leaving open t

simonmar:
I recommended acceptance to the committee.

simonpj:
I have looked at the proposal.  Two issues.

First, the specification leaves me befuddled.  
The entire technical spec is:  *"When ApplicativeComprehensions enabled, the compiler will try to desugar comprehensions as Applicative expressions. If it can not so desugar a particular comprehension, it would get a Monad constraint as usual.  In particular, when enabled, the compiler would desugar all comprehensions applicatively, but if a generator expression uses a term bound in a generator statement, that would be desugared as a monadic bind, as usual, and so need the Monad constraint."*.

What exactly does this mean?   It seems to say too little (there are many phrases I don't understand) or too much.  Concerning the "too much" point, I THINK that the spec may be this:

* The comprehension `[ e | Q ]` is typechecked and desugared precisely as if it was `do { Q; return e }`.

Is that right? If so, there is no need to talk about "desugar all comprehensions applicatively" or "term bound in a generator statement", because that is all covered by "just to what do-notation already does".  If it's not right, then what are the differences?

Second, I would love to narrow the proposal to restrict applicative comprehensions to the simple, predictable desugaring. That would break that simple link with do-notation.  But if that's a direction we want to move anyway, then we'll reduce future breakage if the new applicative comprehensions don't involve the bit we are considering removing or restricting.  There has bee little debate about this, including from the author.  (I asked a couple of weeks ago.)    I don't have a cast-iron view; I'd just like us to make a conscious choice, based on contributions from well-informed users of ApplicativeDo.

Are any of the contributors to this dicussion, committee members and author aside, actual users of ApplicativeDo?  I worry a bit that the proposal is not addressing an eager potential user base.   (That is not to say we should not do it, for the sake of uniformity; but doing it has a cost, thinking about it, implementing it, reviewing the code, and it'd help to have some enthusiastic support!)

adamgundry:
I concur with @simonpj's comments. One other thought occurs to me: if the primary motivation of this proposal is for consistency with `ApplicativeDo`, why do we need an extension named `ApplicativeComprehensions` at all? Instead, could we say that when both `MonadComprehensions` and `ApplicativeDo` are enabled:
 * first the `MonadComprehensions` desugaring applies to convert the comprehension into `do`-notation,
 * then `ApplicativeDo` applies to the desugared `do`-notation to produce `Applicative` combinators (if possible).

That would avoid the question of what `ApplicativeComphrensions NoMonadComprehensions` means, and leave space for a more restricted `ApplicativeComprehensions` extension and/or other alternatives to `ApplicativeDo` in the future.

nomeata:
> why do we need an extension named ApplicativeComprehensions at all?

Maybe because otherwise, the meaning of code using both `MonadComprehensions` and `ApplicativeDo` changes? Intuitively, I’d just 🤷🏻, but we are being training to pay more attention to that these days.

adamgundry:
Well it does generalise some existing code that works with `Monad` to use `Applicative` instead, but I'd be very surprised if that broke anything in practice, especially with `ApplicativeDo` and `MonadComprehensions` already in effect. So you'd need to have a non-law-abiding `Monad` instance that disagreed with `Applicative`, and be translating `do`-notation to (sometimes) use `Applicative` while relying on the fact that comprehension notation always uses `Monad`.

nomeata:
Exactly :-)

simonmar:
Putting the proposal in the dormant state for now; @strake if you're still keen to pursue it, would you be able to respond to the concerns above? I apologise for all the delays, confusion and back and forth here. If you're still interested in the proposal, I'll wake it up again and do my best to help drive it to a conclusion.

PR #528, opened 2022-08-01
Propose -expose-package-internals

goldfirere:
This is an alternative to #524 that more directly gets at the problem we have at the moment.

[Rendered](https://github.com/goldfirere/ghc-proposals/blob/package-internals/proposals/0000-package-internals.rst)

### DIFF
--- a/proposals/0000-package-internals.rst
+++ b/proposals/0000-package-internals.rst
@@ -6,9 +6,7 @@ Accessing package internal modules
 .. ticket-url::
 .. implemented::
 .. highlight:: haskell
-.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
-            **After creating the pull request, edit this file again, update the
-            number in the link, and delete this bold sentence.**
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/528>`_.
 .. sectnum::
 .. contents::
 
--- a/proposals/0000-package-internals.rst
+++ b/proposals/0000-package-internals.rst
@@ -1,0 +1,150 @@
+Accessing package internal modules
+==================================
+
+.. author:: Richard Eisenberg and Simon Peyton Jones
+.. date-accepted::
+.. ticket-url::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+Package authors include an ``exposed-modules`` stanza in ``.cabal`` files to
+list the modules available for import outside the package. Other modules are
+not available for import. Yet it is sometimes useful to access even unexposed
+("internal") modules. This proposal offers a way to do so.
+
+Note that internal modules of a package are *not* a part of that package's
+published API and thus changes therein would *not* require a major version
+bump. These modules do remain internal!
+
+Motivation
+----------
+
+1. Suppose package ``pack`` has several internal modules, whose exports are
+   not intended for the public API of ``pack``. Yet the author of ``pack``
+   still wants to test functions exported by these internal modules.
+
+   The author currently has two options, both bad:
+
+   1. Expose the internal modules, perhaps naming them with ``Internal`` or
+      similar.
+
+   2. Structure the project to recompile the entire package in order to run
+      the testsuite, instead of using the already-compiled package.
+
+   This proposal allows the modules to remain internal while still being
+   available for testing without recompilation.
+
+2. Suppose package ``pack`` exports a very general interface to an particular
+   algorithm. Because it is so general, ``pack`` offers no way of, say, using
+   real-world data to seed the algorithm. Instead, downstream packages such as
+   ``pack-json`` and ``pack-csv`` provide such access.
+
+   If an internal module in ``pack`` needs testing, then there is no choice but
+   to expose the internal module, so it can be tested in a downstream package.
+
+3. Suppose two packages ``pack1`` and ``pack2`` are developed together, with
+   ``pack2`` depending on ``pack1``. Because the two packages form logically one
+   whole (but are separated because, say, ``pack1`` has many fewer dependencies
+   than ``pack2`` or other reasons), ``pack2`` needs to import internal modules
+   from ``pack1``. Yet these internal modules are not intended to form part of
+   any public API.
+
+   This is the case within GHC's boot libraries, where the ``ghc-prim`` package
+   is not intended to have a public API. Yet users cannot easily learn that
+   ``ghc-prim`` is essentially an implementation detail and should instead depend
+   only on ``base`` (which re-exports much, but not all, of ``ghc-prim``). This
+   proposal is intended to protect users from unknowingly depending on ``ghc-prim``
+   modules, which might change frequently.
+
+Proposed Change Specification
+-----------------------------
+
+1. Add a new command-line flag ``-expose-package-internals <pkg>``, where ``<pkg>`` is understood
+   in the same way as the argument to the existing ``-package`` flag.
+
+#. It is an error for the ``<pkg>`` named in ``-expose-package-internals`` not to be exposed.
+
+#. Imports of modules from a package named in ``-expose-package-internals`` skip the exposed-module check.
+
+#. The error GHC produces when a non-exposed module is imported warns the user about
+   exposing internal modules, with something like ::
+
+    Could not load module 'Data.OldList';
+    it is a hidden module in the package 'base-4.15.1.0'.
+    If you wish to be able to access an internal module anyway, you may
+    specify '-expose-package-internals base-4.15.1.0'; however, be warned
+    that the API from internal modules may change between minor version releases.
+
+#. The `Haskell Package Versioning Policy <https://pvp.haskell.org/>`_ specifies that changes in
+   internal modules do *not* require a major version change.
+
+#. Haddock output does not change: only exposed modules are included in the documentation.
+   Users wishing to depend on internal modules are expected to consult the source code.
+
+#. For the next three releases of GHC, any import of a module exported from ``ghc-prim``
+   will emit a warning (controlled by ``-Wghc-prim-exports`` and part of ``-Wcompat``),
+   unless ``-expose-package-internals ghc-prim`` is specified. The warning will inform
+   users that all modules in ``ghc-prim`` will become internal after three releases.
+   When that happens, the warning itself will also be removed.
+
+
+Examples
+--------
+
+*Example.hs*\ ::
+
+  module Example (foldl) where
+  import Data.OldList   -- internal module in `base`
+
+::
+
+  > ghc -c Example.hs
+  Example.hs:2:1: error:
+    Could not load module 'Data.OldList';
+    it is a hidden module in the package 'base-4.15.1.0'.
+    If you wish to be able to access an internal module anyway, you may
+    specify '-expose-package-internals base-4.15.1.0'; however, be warned
+    that the API from internal modules may change between minor version releases.
+
+  > ghc -c Example.hs -expose-package-internals base
+  > ls
+  Example.hs Example.o Example.hi
+
+Effect and Interactions
+-----------------------
+
+1. All GHC users can now access internal modules from other packages, improving
+   the situations outlined in the Motivation_.
+
+#. The exports from ``ghc-prim`` are now protected; users can discover that
+   using names from ``ghc-prim`` is potentially troublesome.
+
+#. There may be a desire to integrate support for ``-expose-package-internals`` in ``.cabal``
+   files, but that is beyond the scope of this proposal.
+
+Costs and Drawbacks
+-------------------
+
+1. With this proposal, package authors now have no way to prevent access of
+   internal modules. However, the primary reason to have internal modules
+   is to present a stable, well-designed API to clients. Because internal
+   modules remain distinct from this public API, the public API is preserved.
+
+   Note that this does not pose a (new) security hazard, given that Template Haskell
+   can be used to access definitions from internal modules.
+
+Alternatives
+------------
+
+1. We could imagine a design where the command-line flag is per-module instead
+   of per-package, but that seems unnecessarily fine-grained.
+
+Unresolved Questions
+--------------------
+
+None at this time.
--- a/proposals/0000-package-internals.rst
+++ b/proposals/0000-package-internals.rst
@@ -37,7 +37,7 @@ Motivation
    This proposal allows the modules to remain internal while still being
    available for testing without recompilation.
 
-2. Suppose package ``pack`` exports a very general interface to an particular
+2. Suppose package ``pack`` exports a very general interface to a particular
    algorithm. Because it is so general, ``pack`` offers no way of, say, using
    real-world data to seed the algorithm. Instead, downstream packages such as
    ``pack-json`` and ``pack-csv`` provide such access.


proposals/0000-package-internals.rst
```diff
@@ -0,0 +1,148 @@
...
+   1. Expose the internal modules, perhaps naming them with ``Internal`` or
+      similar.
+
+   2. Structure the project to recompile the entire package in order to run
+      the testsuite, instead of using the already-compiled package.
+
+   This proposal allows the modules to remain internal while still being
+   available for testing without recompilation.
+
+2. Suppose package ``pack`` exports a very general interface to an particular
```
konsumlamm, line 40:
```suggestion
2. Suppose package ``pack`` exports a very general interface to a particular
```

konsumlamm:
One aspect of internal modules that this proposal completely ignores is unsafe internal functions, that could enable anything from violating internal invariants to violating memory safety if not used correctly. I think that the API might change between minor versions would really be the least of one's problems with this, but the proposal makes it sound like it would be the only one.

proposals/0000-package-internals.rst
```diff
@@ -0,0 +1,148 @@
...
+Effect and Interactions
+-----------------------
+
+1. All GHC users can now access internal modules from other packages, improving
+   the situations outlined in the Motivation_.
+
+#. The exports from ``ghc-prim`` are now protected; users can discover that
+   using names from ``ghc-prim`` is potentially troublesome.
+
+#. There may be a desire to integrate support for ``-expose-package-internals`` in ``.cabal``
```
nomeata, line 125: Probably not even needed, the `ghc-options` stanza would just work out of the box here I assume?

proposals/0000-package-internals.rst
```diff
@@ -0,0 +1,148 @@
...
+   modules remain distinct from this public API, the public API is preserved.
+
+   Note that this does not pose a (new) security hazard, given that Template Haskell
+   can be used to access definitions from internal modules.
+
+Alternatives
+------------
+
+1. We could imagine a design where the command-line flag is per-module instead
+   of per-package, but that seems unnecessarily fine-grained.
```
nomeata, line 143: Again, `{-# OPTIONS_GHC ... #-}` would immediately give the ability to configure this per module (and not need cabal support).

nomeata (commented):
Promising idea. Can likely be abused a lot - and probably will. But that's also true for modules called Internal.

Instead of a flag, one could imagine an annotation on import statement 
```
import Foo.Bar internal
```
similar to source imports for recursive imports. A bit more targeted this way maybe?

simonpj:
Several thoughts:


* Let's say that the proximate reason for offering this proposal is that GHC currently exposes many APIs that are unstable; but *users have no way to know what is unstable and what is not*.  Rather than invent a GHC-specific way to solve this tension, we propose a generic mechanism that can be used by any library, and that builds on an existing mechanism: the exposed/other module distinction in `.cabal` files.

* Let's point out that the whole existing exposed-module/other-module story addressed precisely the goal of separating the public API of a package from its internal APIs.  
  * Some description of the exposed-module/other-module story in Cabal and GHC would help to set the scene. Relying on cultural knowledge is fragile.
  * Can we point to the canonical description of what "exposed modules" and "other modules" means? I can't find such a description; it's alluded to in the [Cabal manual](https://cabal.readthedocs.io/en/3.4/cabal-package.html), but not very explicitly.
  * It is something of a bug that the PVP does not explicitly state that its gudance applies only to exposed modules.
  * The proposal addresses scenarios in which making it *impossible* to import an internal module is unacceptable for some reason.  Currently the effect is to force that module to become exposed; and that in turn willy-nilly expands the public API of the package, which is wrong.  The `ghc-prim` package is a case in point.

* We should say in the proposal that we invite the Core Libraries Committee to take a look at `base` and identify whether all the currently-exposed modules should be exposed.  (Answer: no. They expose all sorts of internal implementation details.)  And then begin a process to narrow the set of exposed modules, thereby more carefully defining the public API of `base`.  The proposal allows this to happen without making it impossible to write packages that deliberately depend on some internal aspect of `base`.

* "Add a new command-line flag -expose-package-internals <pkg>, where <pkg> is understood in the same way as the argument to the existing -package flag."  If you have said `-package-id blah-9.5.1` then it should be enough to say `-expose-package-internals blah`, not `-expose-package-internals blah-9.5.1`.

* "It is an error for the <pkg> named in -expose-package-internals not to be exposed."  Say what it means for a package to be "exposed".  I think you mean that either is is exposed by default (which ones are those?) or you have said `-package blah-9.5.1` earlier in the command line.

* "The Haskell Package Versioning Policy specifies that changes in internal modules do not require a major version change."  Does it specify this?  I think we mean "should be changed to specify....".

* Let's be more explicit about the responsiblities of the user: if you expose the internals of a package then 
  * You are taking responsibility for maintaining any internal invariants that the package requires.
  * You accept that the API of the internals may change with only a minor version bump.

phadej:
How this feature will affect Safe Haskell. Previously, violating module boundaries was considered unsafe.

There might be unexposed modules which are Safe or Trustworthy  annotated, but are unsafe if exposed as they would allow violating invariants.

phadej:
I can imagine a defensive reaction, where people would put whole lib into single module, and only export safe interface. This suggests that the feature is useful only when original author have thought about it, as single module libs can occur naturally as well. (I have refactored libs to move most of it into `Internal` modules to expose things which don't belong to stable API).

Also, the proposal should at least briefly discuss how this would affect build tools. It looks like cabal format have to be amended somehow, so Cabal would know to add `-expose-package-internals` to its `ghc` invocations.

simonpj:
> I can imagine a defensive reaction,

I suppose that a slightly more ambitious proposal would be to have a third category of module
* Exposed: part of the public API
* Internal: not part of the public API but you can import it if you say `-expose-package-internals`
* Private: cannot be imported ever, by anyone.

Personally I doubt that the distinction between the latter two pays its way; but others may feel differently, and there is no technical problem here.

nomeata:
> unusable before there is Cabal support.

Do you need Cabal support? Cabal allows you to specify arbitrary GHC flags, or you can use per module OPTIONS_GHC. So maybe Cabal support is not a blocker? Would be good to know though if this is true or if Cabal would get confused.

phadej:
> > unusable before there is Cabal support.
> 
> Do you need Cabal support? Cabal allows you to specify arbitrary GHC flags, or you can use per module OPTIONS_GHC. So maybe Cabal support is not a blocker? Would be good to know though if this is true or if Cabal would get confused.

good point, `ghc-options: -expose-package-internals ghc-prim` will be enough indeed.

marcosh:
I would just like to add a quick note.

I think it makes sense, more than talking about `internal` and `exposed` modules, to state things in terms of their stability:

- `stable` modules: their exposed API is considered for evolving the version of a package
- `unstable` modules: their code is not considered for evolving the version of a package (this is currently the case for `Internal` and not exposed packages)

This could intersect `publish` and `unpublished` modules (where a `stable` and `unpublished` module does not actually make sense), actually creating the classification mentioned [above](https://github.com/ghc-proposals/ghc-proposals/pull/528#issuecomment-1202344472).

Moreover, making explicit which modules concur in the definition of the API version could allow determining automatically whether a new version needs to be released.

Ericson2314:
This should be coordinated with Cabal:

 - Packages should have to declare their "peaking behind the curtain"

 - PVP doesn't apply to internal modules so arguably the meaning of `^>=`  could be clamped down.

proposals/0000-package-internals.rst
goldfirere: Not necessarily, I think. Maybe there are cases where the same package is included twice but with different unit-ids (either different versions of the same package or different dependencies), and then you'd want to supply the exactly-correct unit-id. Maybe this case is impossible? But having this info directly in the cabal file would also allow more inspection of it, in case a library does not want to depend on other libraries that depend on internals (because then the original library's dependencies might break unexpectedly).

proposals/0000-package-internals.rst
goldfirere: Interesting. I meant "per module" as meaning "per internal module", whereas you interpreted it as "per importing module". Either one works!

goldfirere:
@simonpj 

I've invited you to push directly to the proposal, as I won't have time to implement those changes this week.

chreekat:
In the spirit of ["don't bother asking"](https://twitter.com/apenwarr/status/1549200959980683269), you might consider reversing the behavior.

||Current Proposal|Reversed Behavior|
|--|--|--|
|**Default behavior** |Import of internal module *fails*|Import of internal module *succeeds* (with warning)|
|**Modified behavior** |Import of internal module *succeeds* (-expose-package-internals)|Import of internal module *fails* (-Werror)
|Bonus modified behavior |n/a|Import of internal module *succeeds* (-Wno-internal-modules)

JakobBruenker:
```haskell
> ghc -c Example.hs
Example.hs:2:1: error:
  Could not load module 'Data.OldList';
  it is a hidden module in the package 'base-4.15.1.0'.
  If you wish to be able to access an internal module anyway, you may
  specify '-expose-package-internals base-4.15.1.0'; however, be warned
  that the API from internal modules may change between minor version releases.
```

I'm a little worried that this would cause new users to use `-expose-package-internals` instead of importing the module they should be importing.

One way around that would be to have a pragma like `{-# USE_EXTERNAL Data.List, GHC.List #-}` such that the error message reads
```haskell
  Could not load module 'Data.OldList';
  it is a hidden module in the package 'base-4.15.1.0'.
  If possible, use 'Data.List' or 'GHC.List' instead.
  If you wish to be able to access an internal module anyway, you may
  specify '-expose-package-internals base-4.15.1.0'; however, be warned
  that the API from internal modules may change between minor version releases.
```

nomeata:
It seems tempting to design something big around this, but I think I'd rather build on existing feature. Especially those that don't need cabal integration etc.

So why not use the mechanism https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/exts/pragmas.html#warning-deprecated-pragma, which already is fine-grained (can be used on whole modules _and_ symbols), add a `INTERNAL` variant, and maybe its own `Wno-internals`.

Yes, this makes it opt-in for the providing library and does not allow importing _any_ hidden module. But maybe that's a good thing? The immediate motivation (ghc-prim) is well-served even without taking the ability to hide not export some modules away.

For the case of ghc-prim I'd say even `WARNING` may be enough. So I'm not sure we really need extra code for this.

goldfirere:
I like the idea of building on `DEPRECATED`. But I still think I'd want to control the warning so that I don't get warning-free access to multiple packages' internal modules all at once. (Indeed, I would want the same control for `DEPRECATED`!)

I also like @JakobBruenker's idea, but I, too, am worried about feature creep here.

JakobBruenker:
> I also like @JakobBruenker's idea, but I, too, am worried about feature creep here.

An non-feature-creepy alternative would be to just phrase the error message in a way that explicitly encourages looking for exposed module alternatives first

Ericson2314:
So zooming out a bit, while I am not opposed to this, I don't feel a sense of urgency either. This is because we can achieve something isomorphic to this with no new features!

Instead of have one package with some exposed and non-exposed internal modules we can have two packages:

 - `foo-unstable` with all modules exposed
 - and `foo` with just the stable modules exposed.

This gets us some immediate benefits of Cabal being aware:
1. We can also tell apart consumers that need the unstable parts from consumers that don't.
2. The version on `foo-unstable` will advanced much faster, so we can pin down the exact version of the internals we need. And we can do without complicating the PVP with new rules for stable vs unstable versions of libraries.

Now, granted, having two libraries instead of one is more bureaucracy that arbitrary libraries may find cumbersome. However, I believe the big motivation here is `base` etc., right? It is certainly not that big a cost for a single library, especially as compared to the cost of implementing/planning/bike-shedding this near feature.

-------

One might object, isn't this insufficient because this what we do today with `ghc-prim` and people still use it?

I think what we do today having `ghc-prim` *is* sufficient on a technical level. The problem is since there is *no* way people can use any part of `base` or `ghc-prim` without versions churn, they are somewhat *num* to the pain. I can't emphasize this enough:

**What should be sign as a nuisance and incentive not to use unstable features is instead accepted as part of the inevitable cost of a GHC upgrade.**

As soon as we add reexports/wrappers of stable functionality in a library that *doesn't* have a breaking change every release and encourage people to use that instead, I am confident norms will shift.

-------

I was previously (in #524) mentioning Rust's prior art as something we might emulate, and now here I am advocating we do something less than that. But there is one place I concede . Any `foo-unstable` library like `ghc-prim` will tip off the user its unstable when a breaking change upgrade causes then grief, but users rather learn earlier.

So here's the thing, `ghc-prim` *relative to itself* is a perfectly normal library. If it has a breaking change, it gets a major version bump, if it has a non-breaking change. It is only unstable relative to other libraries (it is not really obeying anything like a 3 release policy, and there are commonplace situations where one is forced to use a newer/older version e.g. where one is forced to use a newer/older compiler.

So if we add any new feature at all --- and to be clear, this is icing on the cake I'd only consider until after we've done the very easy step of creating a stable reexport shim over `base` --- I would add a way to mark an entire library "I am unstable (relative to other things)". There are no special rules / properties of those libraries *unto themselves*, this merely will create warnings on the GHC *and Cabal* level for anything that uses it.

That achieves what Rust does (no footgun of accidentally using unstable things) with simpler machinery (coarse-grained marking an entire library vs specific definitions) and better leveraging existing functionality to make up the difference (and give us good dependency expressiveness for free).

michaelpj:
> Let's say that the proximate reason for offering this proposal is that GHC currently exposes many APIs that are unstable; but users have no way to know what is unstable and what is not.

We do have a couple of things:
- Haddock module metadata has a stability field (but in practice it seems that essentially everything is marked "experimental").
- Cabal package metadata has a [stability field](https://cabal.readthedocs.io/en/3.6/cabal-package.html?highlight=stability#pkg-field-stability), but it 's freeform text.

It's unclear that these features are very meaningful or widely used, perhaps this could be an opportunity to make them more so. (See e.g. https://stackoverflow.com/questions/9012640/how-are-the-haddock-module-fields-portability-stability-and-maintainer-used)

> Instead of have one package with some exposed and non-exposed internal modules we can have two packages

Or perhaps, a package with two public library components? Assuming support for that ever gets finished.

Ericson2314:
> Or perhaps, a package with two public library components? Assuming support for that ever gets finished.

I much rather direct resources towards that than `-XUnstable` or `-expose-package-internals`. Better to prioritize the features with myriad use-cases than just one.

simonpj:
I like the DEPRECATED idea too; but it carries something of the wrong connotation.  I usually think of DEPRECATED meaning "this is old, out of date, and probably going away -- use something else".    Whereas  we might want to use UNSTABLE for "thismay well change; you are quite welcome to use it, but be prepared for upgrade pain".    But perhaps UNSTABLE could be a minor variant of DEPRECATED but basically use all the same mechanism. 

And as Richard says, at the moment we have only a global `-Wdeprecations` flag, and that is too coarse

Neither Richard nor I feel strongly about the details. But I really do think it would be helpufl to be able to be far more specific about what parts of the GHC libraries are stable and which are unstable.  Does anyone feel like driving this proposal forward?

nomeata:
>  But perhaps UNSTABLE could be a minor variant of DEPRECATED but basically use all the same mechanism.

That's the idea, I believe.

proposals/0000-package-internals.rst
```diff
@@ -0,0 +1,148 @@
...
+   not intended for the public API of ``pack``. Yet the author of ``pack``
+   still wants to test functions exported by these internal modules.
+
+   The author currently has two options, both bad:
+
+   1. Expose the internal modules, perhaps naming them with ``Internal`` or
+      similar.
+
+   2. Structure the project to recompile the entire package in order to run
+      the testsuite, instead of using the already-compiled package.
```
Bodigrim, line 35: The blessed way is to put these modules into an internal library, visible for test suite, but invisible for general public. For a real world example see https://hackage.haskell.org/package/attoparsec-0.14.4/attoparsec.cabal

proposals/0000-package-internals.rst
```diff
@@ -0,0 +1,148 @@
...
+
+   This proposal allows the modules to remain internal while still being
+   available for testing without recompilation.
+
+2. Suppose package ``pack`` exports a very general interface to a particular
+   algorithm. Because it is so general, ``pack`` offers no way of, say, using
+   real-world data to seed the algorithm. Instead, downstream packages such as
+   ``pack-json`` and ``pack-csv`` provide such access.
```
Bodigrim, line 43: I don't really understand this example, could you please elaborate? Is there a real world evidence of this problem?

proposals/0000-package-internals.rst
```diff
@@ -0,0 +1,148 @@
...
+
+   If an internal module in ``pack`` needs testing, then there is no choice but
+   to expose the internal module, so it can be tested in a downstream package.
+
+3. Suppose two packages ``pack1`` and ``pack2`` are developed together, with
+   ``pack2`` depending on ``pack1``. Because the two packages form logically one
+   whole (but are separated because, say, ``pack1`` has many fewer dependencies
+   than ``pack2`` or other reasons), ``pack2`` needs to import internal modules
+   from ``pack1``. Yet these internal modules are not intended to form part of
+   any public API.
```
Bodigrim, line 53: The blessed way to solve this is a package, containing multiple (public or internal) libraries. This has been implemented in Cabal 3.8, pending Hackage support.

proposals/0000-package-internals.rst
```diff
@@ -0,0 +1,148 @@
...
+
+   This is the case within GHC's boot libraries, where the ``ghc-prim`` package
+   is not intended to have a public API. Yet users cannot easily learn that
+   ``ghc-prim`` is essentially an implementation detail and should instead depend
+   only on ``base`` (which re-exports much, but not all, of ``ghc-prim``). This
+   proposal is intended to protect users from unknowingly depending on ``ghc-prim``
+   modules, which might change frequently.
```
Bodigrim, line 60: This sounds unconvincing to me, surely GHC or Cabal can easily emit a warning on importing `ghc-prim`.

proposals/0000-package-internals.rst
```diff
@@ -0,0 +1,148 @@
...
+   exposing internal modules, with something like ::
+
+    Could not load module 'Data.OldList';
+    it is a hidden module in the package 'base-4.15.1.0'.
+    If you wish to be able to access an internal module anyway, you may
+    specify '-expose-package-internals base-4.15.1.0'; however, be warned
+    that the API from internal modules may change between minor version releases.
+
+#. The `Haskell Package Versioning Policy <https://pvp.haskell.org/>`_ specifies that changes in
+   internal modules do *not* require a major version change.
```
Bodigrim, line 82: This proposal makes API dependent on GHC compilation flags, which is a huge deviation from PVP. Don't we have enough fragiltiy in the ecosystem already?

tomjaguarpaw:
> Let's say that the proximate reason for offering this proposal is that GHC currently exposes many APIs that are unstable; but users have no way to know what is unstable and what is not.

It's laudable and very much The Haskell Way™ to come up with a general solution to this problem. However, it may also be worth thinking more about how to ameliorate the immediate problem: the instability of the APIs that GHC exposes.

Now, even if we have the most perfect method for indicating which APIs are unstable through tooling, there is still a lot of work to do on an orthogonal concern: GHCHQ has to identify and commit to supporting those APIs that it wishes to keep stable, perhaps in coordination with the CLC and others. It also has to determine what exactly it means by "unstable". Given that each new release of GHC bumps the PVP-major version of `base` I'm not aware of any circumstances in which APIs *were* broken invalidly, according to the PVP[^1].

I think interested parties should proceed by identifying historical examples of when GHC changed an interface and it caused significant pain for some users. I would guess that the lessons to be learned are probably very specific, and  more to do with being mindful of the wide range of uses that GHC is put to, than a general method of communicating "stability".

When GHCHQ has identified APIs that it reserves the right to break without warning then I suggest that the first step is the very low tech approach of simply *documenting that very clearly in the Haddock*. When GHCHQ wants to *add* new such APIs then I recommend that it does so only in modules with `Internal` in the name. The `Internal` module convention is widely used in the Haskell community and I understand that works very effectively.

In short, I think that with a low-tech approach that costs 20% of the work we can achieve 80% of the benefit: just identify unstable APIs exposed by GHC and label them clearly as such. Sophisticated tooling can come later when we've extracted all possible benefit from the unsophisticated approach.

[^1]: I would be grateful to anyone who can point me to examples of where `base` did violate the PVP.

goldfirere:
Perhaps this proposal gets the details wrong, but I stand by my end goal: let there be a way for users to unavoidably confront the instability of unstable parts of their dependencies. It's true that my real motivation is around `ghc-prim`, and so @Bodigrim's suggestion of just adding a magical feature to `cabal` may be good enough. However, I disagree with @tomjaguarpaw's approach, because that seems to require reading Haddock output, and I'm not convinced that our users do so all that diligently. (My first model for many programmers facing a tough problem interacting with the guts of Haskell is that they search, find a blog post describing how to do X, and then copy that code. No Haddock required!)

In any case, I lack the bandwidth to see this properly through, and there doesn't seem to be all that much interest in my end goal (discoverable instability), so perhaps it's worth letting this one die.

If you are reading this more than a month from my posting date and see no further action here, feel free to close.

nomeata:
@goldfirere I know you just put this aside, but would you enjoy a discussion on these points at the [GHC Proposals session](https://icfp22.sigplan.org/home/haskellsymp-2022#GHC-Proposals-Session)?

simonpj:
> Perhaps this proposal gets the details wrong, but I stand by my end goal: let there be a way for users to unavoidably confront the instability of unstable parts of their dependencies

Me too.  And (as others have argued in other proposals) having a machine-checked warning is much, much better than a human checked naming convention.

tomjaguarpaw:
> having a machine-checked warning is much, much better than a human checked naming convention.

I think that having a machine-checked specification is most beneficial when there is some formally-definable semantics. By contrast, when the concepts in play are fuzzy, woolly, open to interpretation, a machine-checked specification is much less beneficial (and sometimes harmful due to the inflexibility it introduces). Drawing a hard line through a fuzzy region necessarily gets some things wrong. Can we formally specify what criteria need to be satisfied for a module to be classed as "internal" or "not internal"? (Bear in mind that `ghc-prim` has never broken the PVP[^1].) If not then I don't see how this proposal helps.

> the ``ghc-prim`` package is not intended to have a public API. Yet users cannot easily learn that `ghc-prim` is essentially an implementation detail and should instead depend only on `base`

Expanding and adapting slightly what I said before, I think we could get a huge amount of the desired benefit of this proposal for very little effort:

1. Put a big red warning at the top of [`ghc-prim`'s documentation front page](https://hackage.haskell.org/package/ghc-prim-0.9.0) that it is considered internal and will break without prior notice. (For what it's worth, this thread is the first time I am learning that `ghc-prim` should not be used by users, and a warning like that would have helped me learn years ago.)
2. In the module Haddocks leave links to the public versions of identifiers (exported from `base` usually) that users should choose in preference.
3. Where there is functionality with no stable alternative offer a channel (`ghc-devs@`?) for users to discuss how to make their code most robust to changes (sometimes this will require exporting a stable version). In fact if indeed nothing from `ghc-prim` should be considered stable I think it's very bad practice not to mention this fact prominently on the front page of the documentation.

I agree with Richard that this method is not perfect and it won't help when someone blindly copy/pastes code from a blog, but I do think it will go a long way for very little effort.

[^1]: to the best of my knowledge -- but if it has then efforts should start there!

tomjaguarpaw:
> I stand by my end goal: let there be a way for users to unavoidably confront the instability of unstable parts of their dependencies

Perhaps I can summarize my position by responding pithily to this: instability is not a binary condition. "Parts of dependencies" are not "either unstable or not". There are degrees of instability, it has many facets, it is a systemic quality that arises from the interaction of many parts of the system. Therefore adding a binary means of classifying stability under a "hard yes or no" seems doomed to failure.

nomeata:
This very much resonates with me. The solutions presented so far feel to specific to a general, vague and squishy problem.

Maybe we should go back to the idea of allowing libraries to attach custom warnings to symbols or modules (like the existing deprecation warnings), including a way to turn them off with relatively high precision (custom warning classes?).

Then a library author can make reliability convey some information to a user, even one that doesn't read haddocks.

Whether that is information about stability, or deprecation, or performance issues, or suggestions about which moon phase this code works best in, is then up to the library authors.

Such custom warnings seem rather straight forward, just a way to silence them in a targeted manner seems tricky.

simonpj:
> Can we formally specify what criteria need to be satisfied for a module to be classed as "internal" or "not internal"? 

Absolutely! 
1. If somehing is "internal" then as a library author I am free to change it without notice, and with only a minor version bump.
2. If something is "external" then as a library author I know that changing it may disrupt my clients.  I strive not to do so, to think about migration paths if I do, and I am obliged to make a major version bump.

As an author I want to be able to signal this distinction unambiguously.  As a client I want to be warned if I (accidentally or on purpose) I make use of an API that the author has unuambiguously signalled as unstable.

It's all very simple! 

I'm pretty agnostic about the precise mechanism.  E.g. if we collectively decide that the "umambiguous signal" is the module name includes the sub-string "Internal", that would satisfy (1).    (Side note: it would also require to add Internal to the name of many `base` modules, which might itself be a disruptive change, but the first thing is to agree the unambiguous signal.)   But then I still need (2): something that yells at me if I import a module whose name includes the string "Internal" unless I explicitly suppress that warning.

tomjaguarpaw:
I'll share a few more details regarding my thinking, because I think it will help understand why I am making the suggestions that I am.

A formal method of indicating instability has the perverse "Nash equilibrium" where *every* package declares *everything* to be unstable, to be technically correct whilst affording the most flexibility to the package author. What matters in practice is not formal adherence to a specification of instability but the actual severity of breaking changes per unit time, because the latter is directly linked to the effort involved in keeping working systems up-to-date. If we care primarily about abstract features of tooling then by all means let's pursue discussions about what they might look like. But on the other hand if we care primarily about the workload of Haskell community members keeping up with ecosystem evolution then we need to take a systemic approach, analyse the historic causes of churn, talk to people who have suffered from it to learn from their perspective, educate maintainers about how to make changes in a way that reduces user workload[^1], and educate users about how to cope with churn more easily. This is a social, software engineering problem, not a technical, computer science problem.  A binary flag "unstable or not" isn't going to make a big dent.

[^1]: For example [Opaleye's breakage policy](http://h2.jaguarpaw.co.uk/posts/opaleyes-api-breakage-policy/)

simonpj:
> But this is exactly what the [PVP](https://pvp.haskell.org/) formalises.

I don't think so!  The PVP says "Breaking change. If any entity was removed, or the types of any entities or the definitions of datatypes or classes were changed,..."   But in to vocabulary above, a change to the type of an entity that is (somehow) unambiguously flagged as uunstable is **not** a breaking change.  

The PVP says that any change to an externally-visible API is breaking.  But this entire thread is about changing that, by allowing an author to classify part of their API (which they want to expose for legitimate reasons discussed in the proposal) as "unstable" and hence not covered by the PVP.

I feel as if I'm not explaining this very well, for which I apologise. 

> every package declares everything to be unstable, to be technically correct whilst affording the most flexibility to the package author

That's entirely possible.  But then every client of that package would have to say `-XUseUnstableInterfaces`, and as a client I'm going to be reluctant to do that.  I'll look for another package. 

Indeed for an experimental package in the early stages of development, it might well make sense to describe a large fraction of the API as unstable.  Committed clients might be willing to absorb the instability.   But more conservative clients could seek packages with stable APIs.

tomjaguarpaw:
I believe I understand you quite well, so I think the problem in communication must be mine.

To check my understanding: your wish is for 

1. a way of marking some subset of an API as not beholden to the PVP, preferably in a machine-checkable way. Is that correct? (Or if not PVP, then some other notion of "stability" that encourage you to formalise.)

simonpj:
> a way of marking some subset of an API as not beholden to the PVP, preferably in a machine-checkable way. Is that correct? (Or if not PVP, then some other notion of "stability" that encourage you to formalise.)

Yes, that's right. But the "machine checked" part refers to clients.   As a client I'd like to be able to say "I don't depend on unstable APIs" without grepping my source tree.  A bit like saying "my code is warning-free".

But I have only a single perspective and I'd love to hear from actual clients! It doesn't really matter what I think; it matters what they think.

My perspective is as a supplier, where I *really* want to be able to signal unambiguously that an API is unstable, and that clients should regard it as part of GHC's internal mechanisms, not as an advertised API.  I am agnostic about the mechanism for this signalling, but I think that whatever it is, it should be a shared understanding between all library authors and all library clients.

tomjaguarpaw:
> But I have only a single perspective and I'd love to hear from actual clients! It doesn't really matter what I think; it matters what they think.

I agree, and the discussion could bear a few more actual clients. But I think this is the kind of thing where you'll have to actively reach out to them. Most actual such clients don't actively pay attention to the ghc-proposals repo, and I suspect that many who do learn of this discussion will feel uncomfortable participating in an unfamiliar space.

> I really want to be able to signal unambiguously that an API is unstable, and that clients should regard it as part of GHC's internal mechanisms

If a non-machine-checkable approach would be acceptable as a first step then I suggest simply just documenting that fact. For example, [`GHC.Classes`](https://hackage.haskell.org/package/ghc-prim-0.9.0/docs/GHC-Classes.html) doesn't make it clear that I shouldn't import it[^1]. I strongly suggest this first step. I don't think that "non-machine checkable" is important enough to be a deal breaker here.

But besides that, I still don't see that the notion of stability outlined in 1 above is particularly useful. As a client I expect `base` and `ghc-prim` to bump their major versions with every new GHC release, so I am expecting some degree of API breakage on every new GHC release, *regardless* of "a machine-checked stability flag". If I am the author of a library I can break my clients every day, even with no internals exposed, quite legally according to the PVP, by releasing a new major version every day. Yes, of course, if you're my client you're going to be reluctant to continue using my package and you'll look for another package. But that's my whole point: this binary notion of stability is nowhere nearly enough. So many informal and social aspects come into play that, to my mind, it's completely redundant, and possibly even counterproductive to try to capture stability in a binary indicator.

Based on my experience I believe what matters is

* amount of churn per unit time (not whether churn is "technically correct" according to some formal specification)
* that maintainers prioritise *not* breaking APIs where possible
* that users are accurately advised of breakage well in advance (it doesn't help much to be warned of breakage that never happens)
* that users have some element of "hand holding" that guides them to the recommended way of dealing with the breakage

I don't see that any "binary" notion of stability can help with those things.

But perhaps its time for other clients to chime in. I don't think I've said anything new here, so unless my repeating myself is particularly clarifying for others I'll refrain.

[^1]: except a very small, grey "internal" label in the corner, behind a sign saying "beware of the leopard".

goldfirere:
@tomjaguarpaw

> 
> * amount of churn per unit time (not whether churn is "technically correct" according to some formal specification)
> * that maintainers prioritise *not* breaking APIs where possible
> * that users are accurately advised of breakage well in advance (it doesn't help much to be warned of breakage that never happens)
> * that users have some element of "hand holding" that guides them to the recommended way of dealing with the breakage

This proposal addresses your first two points, but not your last two.

* If you want to reduce the amount of churn per unit time, then you first have to define "churn". For example, changing the name of a local variable in a function is not churn; it is completely invisible to users. A reasonable definition of "churn", to me, is "changes to publicly visible APIs" (including both static and dynamic semantics -- that is, types and runtime behavior). According to this definition, labeling parts of a package as unstable actively helps reduce churn, by removing some package-exported modules from the public API. Or, rather, if you like, it allows clients to decide how much churn to expose themselves to.
* Sometimes, for a library to advance, it must change aspects of exported functions. Of course we want libraries to prioritize not doing so. But having shielded parts of a package helps here, too: package authors can think about how to advance their library by changing only the internal parts. This gives more flexibility to package authors, while not affecting the public API, thus making it easier for a package author to prioritize not changing the public API.

It feels to me here that @tomjaguarpaw is asking for more than this proposal (or its line of thinking -- again, abstracting over the details) aims to deliver. All we want here is a way for users to be aware (without needing to look at Haddock) what is part of a package's public API and what is part of its internal, unstable API. Once again, there is a direct, formal consequence of this: changing the internal API does not require a version bump.

Regarding `ghc-prim` and the PVP: You're right that `ghc-prim` has not violated the PVP. But maybe with the ideas in this proposal, we could continue to evolve `ghc-prim` (and `base`) *without* major version bumps. That would reduce churn! And yet we currently have no way of doing so, without breaking the PVP.

For what it's worth, I'm content (barely) with just using a naming convention: perhaps we simply modify the PVP to except any modules ending in `.Internal`. Then we rename all of the `ghc-prim` modules and be done with it. Note that the `.Internal` idea works fine with my "copy from blog post" scenario, because the new code still has to add the imports. (I suppose the IDE might do so. But hopefully an IDE could be taught not to blindly add `.Internal` imports!)

nomeata:
> But maybe with the ideas in this proposal, we could continue to evolve ghc-prim (and base) without major version bumps. 

Even without, if all internal stuff is in ghc-prim (maybe rename to ghc-internals?) then a breaking change and PVP bump there does _not_ imply a major version bump of base.

More general, instead of marking some modules as internal, isn't it cleaner to keep them in a separate library, clearly named as internal (base-internals, ghc-internals) and then base only exports the stable parts. This

* The PVP applies as intended
* Users don't even _see_ the internal modules, so they are nudges away from them
* Access to internals requires them to put a thus named package into the cabal file, which seems a clear opt-in into the instability
* Library authors are free to make breaking changes to the -internals library (with major version bumps as they should, but they only affect those that explicitly use the internals)

Bodigrim:
I think this discussion mixes two different questions: one is `ghc-prim` and another is library stability in general. 

The peculiarity of `ghc-prim` is that it is not a Haskell package, and it is automatically generated from GHC primops. So there is absolutely no way to provide any compatibility wrappers or utilities. Ideally no one except `base` should consume `ghc-prim`: everyone else should use `GHC.Exts` from `base`. Historically it was not always true: some entities failed to be re-exported and dependency on `ghc-prim` ossified into many packages.

So `ghc-prim` story is irrelevant to any hypothetical stable/unstable API distinction, there are just zero reasons to depend on it, period. That's why I suggested above for `cabal check` to warn about `build-depends: ghc-prim`.

nomeata:
Extending the WARN pragma with a way to specify the warning category (for sufficiently targeted silencing) might not only solve this use case, but also others, like warning about use of partial functions, as discussed in https://github.com/haskell/core-libraries-committee/issues/87#issuecomment-1244508398

tomjaguarpaw:
> It feels to me here that @tomjaguarpaw is asking for more than this proposal

On the contrary, I want much *less*. Let's just note clearly in the docs of `ghc-prim` that it shouldn't be used. Here's me putting my commits where my mouth is: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/8989

gridaphobe:
While I am sympathetic to the desire to design a language feature to solve this problem, let me also point out that the Haskell ecosystem has had a very simple solution in place for many years.

1. Packages put internal definitions inside `Internal` modules, e.g. [`Data.ByteString.Internal`](https://hackage.haskell.org/package/bytestring-0.11.3.1/docs/Data-ByteString-Internal.html).
2. All stable entities are either defined in non-`Internal` modules, or re-exported from non-`Internal` modules.
3. `Internal` modules are still "exposed" via cabal, so downstream users can access them if necessary.
4. `Internal` modules are **not** bound by the PVP, and may change arbitrarily between releases without a major bump.
5. Clients of `Internal` modules are expected to understand this and deal with possible breakage. Or they should lobby to stabilize the exports they need.

As far as I can tell, this convention has worked quite well in the broader ecosystem. GHC's problem is that we have an existing module structure that does not adhere to this convention. We can change that!

nomeata:
So we could rename all modules in ghc-prim to GHC.Internal.* and be done with this? (Maybe also renaming ghc-prim to ghc-internals for good measure?)

simonpj:
> Internal modules are not bound by the PVP, and may change arbitrarily between releases without a major bump.

That convention makes sense, but I did not know about it.  Moreover it seems absent in [the PVP itself](https://pvp.haskell.org/).  Do you think it would make sense to update the PVP? 

Should we rename GHC.Base to GHC.Internal.Base?   That would be pretty disruptive, but perhaps worth it to conform to the convention.  Or perhaps we could define a different convention for the `base` library -- perhaps anything starting with `GHC.` is counted as "internal" and not subject to the PVP?

tomjaguarpaw:
> > Internal modules are not bound by the PVP, and may change arbitrarily between releases without a major bump.
>
> That convention makes sense, but I did not know about it. Moreover it seems absent in [the PVP itself](https://pvp.haskell.org/). Do you think it would make sense to update the PVP?

The convention is [mentioned in the text of the proposal](https://github.com/goldfirere/ghc-proposals/blob/package-internals/proposals/0000-package-internals.rst#id1). Is there some way we can amend the proposal to make the explanation of the convention stand out more clearly?

simonmar:
If we had `-expose-package-internals` would uses of it be rejected by Hackage? That would seem reasonable, I imagine `-expose-package-internals` is useful for local testing and experimentation but it's not for released packages on Hackage. Or are there intended use cases where we want this in released packages? If so, that seems somewhat worrying because it breaks our expectations about when a package should compile with respect to the versions of the packages it depends on.

tomjaguarpaw:
My understanding was that use of `expose-package-internals` was essential under some circumstances. For example, how otherwise would `base` expose stable parts of `ghc-prim`?

gbaz:
I support simply following the Internal convention rather than this flag. That said:

> That convention makes sense, but I did not know about it. Moreover it seems absent in [the PVP itself](https://pvp.haskell.org/). Do you think it would make sense to update the PVP?

It should be noted that not all uses of .Internal are about the PVP. Very often, even packages which exposes Internals modules will obey the PVP in all modules. They will nonetheless discourage use of the Internals module not for PVP reasons but because, e.g., use of the module may break invariant (creating an unbalanced "balanced tree" for example).

I do not think we should update the PVP proper. However, I do believe that some discussion of this phenomena "in the wild" should be added in the FAQ (and would support a PR to do so): https://pvp.haskell.org/faq/

In particular, it should be noted in the FAQ that if package A depends on package B's B.Internal module, and if the latter module makes no PVP guarantees, then package A must use tight bounds for B (i.e. precise bounds, not up to the next major version), and if package A exports anything that "exposes" B.Internal then package A as a whole cannot be PVP compliant.

Further, I do not think we should encourage this behavior. I think we should encourage accurate version numbering even with regards to internals modules. The worst case scenario is that a package's major version is bumped more frequently than one might desire. But what is the harm in this? (virtually none, I claim!). In general, authors should change code as necessary, and version numbers should simply _reflect_ this. The desire for a particular version number should typically not drive what code is produced or how it is structured.

simonmar:
> My understanding was that use of `expose-package-internals` was essential under some circumstances. For example, how otherwise would `base` expose stable parts of `ghc-prim`?

I'm probably missing something, but don't we already expose parts of `ghc-prim` from `base`? I would think that `ghc-prim` is itself an unstable / internal package, its major version changes all the time, but `base` provides a more stable abstraction over it. We don't need any extensions to do that.

tomjaguarpaw:
I see I was ambiguous. What I meant is that if this proposal passed, and all of `ghc-prim` was formally marked "internal", then `base`  would have to use `-expose-package-internals` in order to continuing reexporting the parts of `ghc-prim` that it currently does. Therefore it wouldn't be possible to reject uses of `-expose-package-internals` from Hackage.

simonmar:
> Specfically, what I meant is that if this proposal passed, and all of `ghc-prim` was formally marked "internal", then `base` would have to use `-expose-package-internals` in order to continuing reexporting the parts of `ghc-prim` that it currently does. Therefore it wouldn't be possible to reject uses of `-expose-package-internals` from Hackage.

Ah, I missed that detail of the proposal. So the proposal is actually to introduce a restriction (all modules of `ghc-prim` are internal) and a way to circumvent the restriction (`-expose-package-internals`) because there is exactly one legitimate use of `ghc-prim`. I am feeling even less convinced now! Why don't we just add a warning when someone uses `ghc-prim` if that's what we're trying to achieve?

proposals/0000-package-internals.rst
aspiwack: It is not raised in Richard's motivations, but private libraries, just as well as unexposed modules, run the risk of exposing functions without exposing enough type constructors to say the functions' types. So if we make private libraries, I will actually want a way to depend on them anyway, much in the manner that is proposed here.

goldfirere:
In light of all the comments here, I think this ticket has outlived its usefulness. I agree that a new feature is not the right solution to this problem. Thanks for the discussion!

PR #531, opened 2022-08-10
Fix typo "effect"->"affect" in proposal template

clyring:
(not a proposal)

### DIFF
--- a/proposals/0000-template.md
+++ b/proposals/0000-template.md
@@ -78,7 +78,7 @@ during the PR discussion.
 
 ## Costs and Drawbacks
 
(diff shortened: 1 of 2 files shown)


nomeata (approved):
I wonder if some language nerd can find a justification for how “effects” isn't wrong here.

 

But affects sounds at least as right.

chreekat:
> I wonder if some language nerd can find a justification for how “effects” isn't wrong here.
> 
> 
> 
> But affects sounds at least as right.

In the comic, the troll is using "to effect" in its capability as a synonym for "to cause". :)

PR #532, opened 2022-08-14
Clean up and simplify the treatment of implicit binding

Ericson2314:
The review process of #448, to be quite frank, felt lacking. Some last minute changes (which @goldfirere said elsewhere were done based on others unwritten concerns, i.e. weren't even something he was himself championing!) weakened the story around implicit binding, and when I had brought this up I didn't get much by way of replies.

Lest this appear too theoretical, I was also helping @tek implement some of this stuff, and the lack of clear story around implicit binding got in the way there too. One of the reasons variable binding is so inconsistent in GHC Haskell is that the implementation contains lots of repetitive and poorly abstracted code --- in such a situation there are few impediments to stop the various instances of should-have-been abstractions "drifting apart". @tek and I were able to do a great job deduplicate things, but the few remaining places were implicit variables could never be turned off meant we couldn't go all the way. Bummer!

But that's enough grumbling, here are the changes I would have liked to see happen. I have editing the existing proposals for my sanity; I new proposal atop the old one which is atop others is too much and would melt my brain. As such, the motivations are editted and should be self-contained, but I will provide an extra "meta-motivation here".

If it is better that this go in the repo, sure, I suppose I could make a rump all-motivation proposal for it instead. Just let me know.

--------

### Principles

A positive (to me) thing I appreciated near the end of the #448 review process was that @gridaphobe agreed with my assessment that some of the principles were insufficiently distinct. We were thus able to reach an agreement with @goldfirere to axe the "local lexical scoping principle", and fold it in to the "lexical scoping principle".

This was a great start, but upon rereading the principles (in order to connect the amended proposals to them), I realized that the "lexical scoping principle" an "explicit binding principle" still felt too close, and awkwardly dividing the labor between them.

I thus flipped their order ("explicit binding principle" first) and gave them the following roles (pithily summarized):

1. **Explicit Binding Principle**: Every implicit binding form needs an explicit equivalent.
2. **lexical scoping principle**: Must be possible to turn off implicit binding forms so only the explicit ones remain. 

I think this is much simpler and easy to understand.

### #425

I just updated to refer to principles. Most of this is about very nuanced arity, but the "easy parts" are very much part of the 448 story and thus relate to the same principle.

#### A single unified `-XImplicitBinds`

Originally this is how #285 was. Then someone convinced me pattern synonym binding and and implicit foralls are quite different. More recently, we realized some of the #285 examples were not covered by *either* extension. Too bad! Because those examples are of things people wanting the other restricted behavior would also want.

One solution was to make a *third* `-XNo` relating to binds, to pick up the missing things. Believe me, I was tempted! But, I know everyone is getting weary of type variable extensions :).

I think the better solution --- which I went with --- is just to recombine things. Yes, implicit foralls and implicit patern synonym binds are indeed *not* the same, but the *motivations* for why to disable them are. The same people that dislike one of them dislike all of them, and vice versa. Likewise, the same motivations around education and syntactic consistency that apply to one of them apply to all of them.

#### `-Werror=pattern-signature-binds` considered inadequate

Finally, note that `-XNoPatternSignatureBinds`, one of the former constituents of `-XNoImplicitBinds`, had been downgraded to a warning. This might seem fine (use `-Werror=...`!) but it really isn't, because it fails each step of the motivation:

##### Education

The point of disabling features with `-XNo` is so the student can be *completely unaware* they exist. But warnings must always be phrased in terms "that thing you did you might not have wanted to". That means making the student aware of the thing after all --- "that thing you did" is something the student was never taught and therefore should never be goaded by a warning into learning about after all.

Concretely, for educational purposes we want to get rid of implicit bindings, and get rid of the *concept* of implicit binding. We want the student to be *unable* to write them, and we want the compiler to *not* tell them with other configuration options feature exists, the same way a Haskell 98 users should not be told about "type families" or weather.

##### Single namespace syntactic uniformity

#270 has a very nice story about making single name-space code not fork-like by accepting fewer programs.
`-Werror=pattern-signature-binds` either breaks the "non fork-like" condition, or breaks the "single namespace" condition. Either is not acceptable.

This relates to the education case in that both are about being able to hide what "might have been" under other config settings.

### #523

This unmerged proposal is referenced in #448 (in PR form, with no implication that it is eventually accepted). The reason for this I think is worth elaborating on.

I think the reason we got into the confusing situations we have so far is because pattern signature binds are not obviously "syntactic sugar", in that there is no simple non-type-directed desugaring of what they do. I am steadfast that any such "weird" feature is "sugar in waiting" --- we simply need to create the much simpler primitives until it is sugar, but others are more "wait and see" and "by the book", and therefore don't want to ascribe to something the negative connotations of syntactic sugar until it is manifestly clear that it in fact is syntactic sugar.

#523 fixes this, by hinting at (it is not fully specified yet) the `let type var = _ in` syntax that can be used instead of pattern synonym binds. The desugaring is simple, not type directed, and only rename-directed in that we need to know what variables are as-of-yet not explicitly bound.

I don't call pattern signature binding "sugar" in the revised text, but I do call it "implicit', because any syntax that could be either a use or a binding based on the context (of in-scope variables) I define as "implicit". For those not comfortable with this yet, I suggest we hurry up and accept #523 so that it is also unambiguously "implicit" and "sugar" by having the explicit `let type var = _ in` syntax it can be desugared to.

--------------

Because the information is scattered around previous proposals and the baseline of what GHC implements, it was *very* difficult to right this proposal and keep everything in my brain.

I know I am a broken record about this, but please let's do something like the "aspirational users guide". The end goal of all this variable stuff is quite simple ---- simpler and more uniform than what we have before it! But the balkanized way we discuss and propose these things and (if need be) their migrations is extremely cognitively burdening.

I feel in writing this proposal, in discussing the weird interactions between proposals both unmerged and merged, code quality issues in GHC, etc., we are pushing against the limit of the complexity and nuance the current proposal process can deal with. And that's a real shame, because complexity we fail to deal with is not complexity that doesn't happen, but complexity that does happen, slipping by beneath our radar. I want to change the process to make there be *less* work and drudgery, not *more*.

I am *not at all confident* I got everything in my revisions correct, but I am quite exhausted thinking about it, and so I open this now in hopes that any lingering errors are easy to catch and fix with more eyeballs.

---------------

[Rendered Principles](https://github.com/Ericson2314/ghc-proposals/blob/type-variables/principles.rst)
[Rendered 425](https://github.com/Ericson2314/ghc-proposals/blob/type-variables/principles.rst)
[Rendered 448](https://github.com/Ericson2314/ghc-proposals/blob/type-variables/principles.rst)

### DIFF
--- a/principles.rst
+++ b/principles.rst
@@ -47,6 +47,7 @@ Accepted principles
 
 .. _`#281`: proposals/0281-visible-forall.rst
 .. _`#378`: proposals/0378-dependent-type-design.rst
+.. _`#425`: proposals/0425-decl-invis-binders.rst
 .. _`#448`: proposals/0448-type-variable-scoping.rst
 
 Syntax
@@ -158,6 +159,68 @@ there is an absence of punning.
 Name resolution and scoping
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+Explicit Binding Principle (EBP)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. _`Explicit Binding Principle`:
+
+**Principle**: Through the right combination of extensions, every implicit form of variable binding
+is must have an explicit equivalent that, regardless of the context, is unambiguously a binding site.
+
+Examples:
+
+#. Problem::
+
+     -- Assume no `a` in scope
+
+     id :: a -> a  -- The variable `a` has no explicit binding site.
+
+   Solution::
+
+     -- Assume no `a` in scope
+
+     id :: forall a. a -> a  -- The `a` in `forall a.` is an explicit binding site.
+
+   This is provided by ``-XExplicitForAll``, which predates the GHC proposal process.
+
+#. Problem::
+
+     -- Assume no `a` in scope
+
+     data Foo (a :: k)
+
+   Solution::
+
+     data Foo @k (a :: k)
+
+   This is provided by ``-XTypeAbstractions`` from `#425`_.
+
+#. Problem::
+
+     -- Assume no `b` in scope
+
+     f :: (Bool, Bool) -> Bool
+     f (x :: (b, b)) = ...   -- The variable `b` has no implicit binding site.
+
+   We could declare one or both of the ``b`` occurrences above a binding site,
+   as was the historical interpretation of this, but that doesn't help as this
+   syntax isn't unambiguously a binding site regardless of context (i.e.
+   regardless of whether there is a `b` already in scope).
+
+*Motivation:* The `Explicit Binding Principle`_ allows programmers to control exactly how variables come into
+scope.
+It ensures all short-hands can be explained in terms of an explicit, unambiguous equivalent that is easier to understand at the cost of being more verbose:
+
+- Positive-position signatures' free vars cause  implicit ``forall ... .``
+
+- Negative position free vars cause different sorts of binding:
+
+  - Signatures on term patterns (pattern signatures) cause implicit ``let type ... = _ in``
+
+  - Signatures on type variables (kind signatures) case implicit ``@...``
+
+From `#425`_, `#448`_.
+
 Lexical Scoping Principle (LSP)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
@@ -165,25 +228,39 @@ Lexical Scoping Principle (LSP)
 
 **Principle**:
 
-a. For every appearance of
-an identifier, it is possible to determine whether that appearance is a *binding site*
-or an *occurrence* without examining the context.
+a. For every appearance of an identifier, it is possible to determine whether that appearance is
+   a mere *occurrence*, and thus must be bound elsewhere for the program to be valid.
+   or the variable is a *binding site* (or causes an implicit binding site, basically the same thing).
+
+without examining the context.
 
 b. For every *occurrence* of an
 identifier, it is possible to uniquely identify its *binding site*, without
 involving the type system.
 
+This builds upon the `Explicit Binding Principle`_:
+whereas that former principle ensures that explicit alternatives to implicit binding constructs *exist at all*,
+this latter principle makes those explicit alternatives *compulsory*, because we must not have implicit binding in order to uphold this principle.
+
 The `Lexical Scoping Principle`_ is almost true today, with the following nuances:
 
-1. Template Haskell splices may need to be run before completing name resolution (and running those splices requires type-checking them).
+#. Template Haskell splices may need to be run before completing name resolution (and running those splices requires type-checking them).
 
-2. The `deprecated mechanism <https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/exts/duplicate_record_fields.html#selector-functions>`_ for disambiguating duplicate record fields violates the `Lexical Scoping Principle`_ by requiring the type system.
+#. The `deprecated mechanism <https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/exts/duplicate_record_fields.html#selector-functions>`_ for disambiguating duplicate record fields violates the `Lexical Scoping Principle`_ by requiring the type system.
 
-3. In a pattern signature, if we have ``f (x :: Maybe a)``, the ``a``
-   is an occurrence if ``a`` is already in scope, and it is a binding site otherwise.
+#. In a pattern signature, if we have ``f (x :: Maybe a)``, the ``a``
+   is an occurrence if ``a`` is already in scope, and is implicitly bound otherwise.
 
-4. In a type signature, any out-of-scope variable is implicitly bound. This is not
-   technically a violation of this principle (the seemingly-unbound identifier in the type signature
+#. In a type signature, if we have ``f :: a -> a``, the ``a``
+   is an occurrence if ``a`` is already in scope, and it implicitly bound otherwise.
+
+#. In a kind signature, if we have ``data Foo (a :: k)``, the ``k``
+   is an occurrence if ``k`` is already in scope (historically impossible), and it implicitly bound otherwise.
+
+   Technically this might be a violation of this principle as the ``k`` is in fact always a use with the implicit bind coming before,
+   but implicit foot-gun behavior
+
+   (the seemingly-unbound identifier in the type signature
    is always an occurrence), but it's worth noting here.
 
 *Motivation:* These principles mean that we can understand the binding
@@ -196,31 +273,6 @@ of mixing the term-level and type-level namespaces (`#270`_) and need to think a
 variables and imported term variables.
 
 (a) from `#448`_; (b) from `#378`_.
-
-Explicit Binding Principle (EBP)
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-.. _`Explicit Binding Principle`:
-
-**Principle**: Through the right combination of extensions and/or warning flags, it is possible
-for a Haskell programmer to ensure that all identifiers in a program have an explicit binding site.
-
-Examples::
-
-   id :: a -> a    -- the variable `a` has no explicit binding site, but we can write `forall a.` to provide one
-
-   f :: (Bool, Bool) -> Bool
-   f (x :: (b, b)) = ...   -- the variable `b` is bound to `Bool` by this
-                           -- pattern signature. But either the first b is a binding
-                           -- site, in violation of the Lexical Scoping Principle (a),
-                           -- or there is no explicit binding site, in violation of
-                           -- the Explicit Binding Principle.
-
-*Motivation:* The `Explicit Binding Principle`_ allows programmers to control exactly how variables come into
-scope. It also prevents the possibility of typos that accidentally introduce new
-variables.
-
-From `#448`_.
 
 Contiguous Scoping Principle (CSP)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/proposals/0425-decl-invis-binders.rst
+++ b/proposals/0425-decl-invis-binders.rst
@@ -621,6 +621,15 @@ The proposed changes provide the programmer with a more principled way of
 brining type variables into scope in certain corner cases, simplify arity
 inference and scoping rules.
 
+This gets us closer to upholding two principles:
+
+- The `Explicit Binding Principle`_, because the new ``@``-binders is explicit unlike previous alternatives
+
+- The `Lexical Scoping Principle`_, because the new simplified rules about variables needing to occur on the LHS in certain situations.
+
+.. _`Explicit Binding Principle`: ../principles.rst#explicit-binding-principle
+.. _`Lexical Scoping Principle`: ../principles.rst#lexical-scoping-principle
+
 Costs and Drawbacks
 -------------------
 
--- a/proposals/0448-type-variable-scoping.rst
+++ b/proposals/0448-type-variable-scoping.rst
@@ -27,6 +27,7 @@ together many existing proposals:
 .. _`#378`: https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0378-dependent-type-design.rst
 .. _`#402`: https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0402-gadt-syntax.rst
 .. _`#420`: https://github.com/ghc-proposals/ghc-proposals/pull/420
+.. _`#523`: https://github.com/ghc-proposals/ghc-proposals/pull/523
 .. _Type Variables in Patterns: https://richarde.dev/papers/2018/pat-tyvars/pat-tyvars.pdf
 .. _Kind Inference for Datatypes: https://richarde.dev/papers/2020/kind-inference/kind-inference.pdf
 .. _`Haskell 2010 Report`: https://www.haskell.org/onlinereport/haskell2010/haskellch10.html
@@ -104,12 +105,19 @@ synonym of ``-XScopedTypeVariables``. Under this change, that would no longer be
 
 This component of this proposal is taken
 from the not-yet-accepted proposal `#238`_, changing the name of what I now call
-``-XExtendedForAllScope``, and simplifying the binding story around pattern signatures
-(getting rid of ``-XPatternSignatureBinds``). This part of the proposal also introduces
-a new warning ``-Wpattern-signature-binds`` (available only in ``-Weverything``) as a
-new way of handling the pattern-signature-binding part of `#285`_.
+``-XExtendedForAllScope``.
 
-This component includes the ``-XNoImplicitForAll`` of `#285`_ unchanged.
+``-XImplicitBinds`` is the combination of ``-XImplicitBindsForAll`` and ``-XPatternSignatureBinds``) from accepted
+proposal `#285`_, and differ from that proposal in that:
+
+  - The extensions are combined for fewer knobs when the motivations are the same.
+
+  - ``RULES`` was forgotten as an example.
+
+  - Examples of implicit binds falsely categorized as pattern signature binds are now explicitly included.
+
+    "Pattern signatures" has a narrow meaning but @Ericson2314 misunderstood it to include other negative-position type signatures.
+    Now that the extensions are combined we side-step the phrase "pattern signature" more easily.
 
 Motivation for any extension shuffling
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -127,8 +135,16 @@ two reasons:
   It thus seems necessary to separate out the problematic ``-XExtendedForAllScope``
   from the other components of ``-XScopedTypeVariables``.
 
+A secondary goal is to clean up some issues with proposal `#285`_ while simplifying things:
+
+* ``-XImplicitForAll`` and ``-XPatternSignatureBinds`` have the exact same
+  motivation, and it is unclear why one would ever want one without the other.
+
+* ``-XImplicitForAll`` and ``-XPatternSignatureBinds`` *missed* some of the cases in the examples, which clearly are implicit binding forms meant to be turned off per the overall motivation, but nonetheless slipped through the cracks of the drafting process.
+
 Having separated out ``-XExtendedForAllScope``, it seemed strange to have a ``-XRumpEndOfOldScopedTypeVariables``
 extension, and so I've introduced separate ``-XMethodTypeVariables`` and ``-XPatternSignatures``.
+``-XPatternSignatures`` is more narrow than it's GHC 6.x namesake so as not to violate any principles by default, and to match ``-XKindSignatures`` which is narrow in the same fashion.
 
 Motivation for ``-XPatternSignatures``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -182,13 +198,27 @@ conservative; it may actually break my program somewhere!
 What I really want in this case is a pattern signature, and it would be nice if I could
 just state that ``PatternSignatures``.
 
-Motivation for ``-XNoImplicitForAll``
+Motivation for not doing binding
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Consistency
+"""""""""""
+
+Firstly, it matches ``KindSignatures``.
+``KindSignatures`` doesn't allow implicit binds for a rather roundabout reason: implicit binds would imply implicit kind-level foralls, which would require ``-XPolyKinds``.
+Given the other extensions being proposed here, we can retroactively reinterpret this as a simple syntactic rule: ``-XKindSignatures`` alone doesn't do implicit binding.
+Now, ``PatternSignature``
+
+Conservativism for standardization
+""""""""""""""""""""""""""""""""""
+
+With both of these signatures being very minimal, I think they would be easy uncontroversial candidates for a new language report.
+Conversely, all implicit binding constructs are very fraught with a complicated mix of upsides and downsides, we and should only standardize them with great care.
+
+Motivation for ``-XNoImplicitBinds``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-This is taken directly from `#285`_, updated to refer to warnings instead of language extensions.
-The "I" here is John Ericson.
-
-There are two independent motivations: education and consistency with a unified namespace.
+This is mostly taken  from `#285`_, but modified now that @Ericson2314 realizes both extensions share the same motivations not one having more than the other.
 
 Education
 ^^^^^^^^^
@@ -198,42 +228,74 @@ All other variables are explicitly bound, and the inconsistency means more to le
 Also, implicit syntax in general allows the beginner to not realize what they are doing.
 What are tedious tasks for the expert may be helpful learning steps to them.
 
-Further, the most beginnning students may be taught with both ``-XNoImplicitForAll`` and ``-XNoExplicitForAll``.
+Further, the most beginnning students may be taught with both ``-XImplicitBinds``, ``-XNoExplicitForAll``, and ``-XNoPolyKinds``.
 This means it's impossible to write forall types by any means.
 Combine with ``-Wmissing-signatures`` and ``-Wmissing-local-signatures``, so inferred polymorphic types of bindings are also prohibited, and a monomorphic custom prelude, and forall types are all but expunged entirely.
 
-I don't wish to argue whether these choices do or don't actually help learning, but just state that some people have opinions that they do and there is no technical reason GHC cannot accommodate them.
+@Ericson2314 doesn't wish to argue whether these choices do or don't actually help learning, but just state that some people have opinions that they do and there is no technical reason GHC cannot accommodate them.
+
+Consistency
+^^^^^^^^^^^
+
+Notice how today that out-of-scope variables in negative position signatures are implicitly bound in *different* ways depending on whether they are type variables (in pattern signatures) or kind variables (in negative position kind signatures).
+By banning implicit binding, we side-step that difference.
+
+After all, given::
+
+  data Foo (a :: k)
+
+desugars to::
+
+  data Foo @k (a :: k)
+
+a new Haskeller might conceivably think::
+
+  \(Foo (a :: k) -> ..
+
+desugars to::
+
+  \(Foo @k (a :: k) ->
+
+or::
+
+  \ @k (Foo (a :: k) ->
+
+which happen to be true in some simple common cases, but are in fact incorrect in general.
+
+That it takes a complicated example to show why these false desugarings aren't true in general make this is a huge educational stumbling block!
 
 Unified Namespace
 ^^^^^^^^^^^^^^^^^
 
 If `#270`_ is accepted, there will be a way to program Haskell with "morally" one namespace for types and terms alike.
-However, there is one exception to the unification of namespaces: lower case variables in type signatures bound "like terms" still are treated as free and implicitly bound with a ``forall`` instead::
+However, there is one exception to the unification of namespaces: lower case variables in type signatures bound "like terms" still are treated as free and implicitly bound instead::
 
   t = Int
   x :: t -- sugar for 'forall t. t', not 't ~ Int'
   x = 0
 
-Should the ``t`` in ``x :: t`` cause an implicit ``forall t.`` to be synthesized or not? With ``-XNoImplicitForAll``, we know
-it will not, and thus can refer to the ``t`` defined above, once such a reference is possible (left to another proposal).
+  t = Int
+  foo (x :: t) = 0 -- sugar for 'foo = let t = _ in \(x :: t) -> 0'
 
-Motivation for ``-Wpattern-signature-binds``
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Should the ``t`` in ``x :: t`` cause implicit ``forall t.`` and ``let t = I in`` to be synthesized or not?
 
-This is necessary in order to uphold the `Lexical Scoping Principle`_, part (a).
+With ``-XNoImplicitBinds``, we know it will not, and thus can refer to the ``t`` defined above, once such a reference is possible (left to another proposal).
+
+Without ``-XImplicitBinds`` we have no choice but do the implicit desugaring that violates the unified namespace abstraction
+--- otherwise we would be *forklike* in the that rather than giving new meaning to previously invalid programs, we would be *changing* the meaning of previously *valid* programs.
+Put another way, ``-XImplicitBinds`` and any ``-XUnifiedNamespace`` must be mutually exclusive, so we need a way for Haskell programmers to opt out of ``-XImplicitBinds`` first.
 
 Proposed Change Specification
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Points below up to and including the new (backward-compatible) definition of
-``-XScopedTypeVariables`` come from not-yet-accepted proposal `#238`_. The point
-about ``-XImplicitForAll`` is a restatement of (part of) accepted proposal `#285`_.
-The other part of `#285`_ has been modified to use ``-Wpattern-signature-binds``.
+``-XScopedTypeVariables`` come from not-yet-accepted proposal `#238`_.
+``-XImplicitBinds`` is a fixed and simplified (via combining extensions) version of accepted proposal `#285`_.
 
 1. Re-purpose deprecated extension ``-XPatternSignatures``. With ``-XPatternSignatures``, we
    allow type signatures in patterns. These signatures can mention in-scope
-   type variables as variable occurrences. A mention of an out-of-scope variable
-   binds the type variable as an alias of the type it is unified with (as today).
+   type variables as variable occurrences, but can not bind type variables without the
+   separate ``-XImplicitBinds`` extension. Do note that extension is on by default, however.
 
    The current ``-XPatternSignatures`` is just a synonym for ``-XScopedTypeVariables``.
    This change is thus not backward-compatible, but given that the existing extension
@@ -259,52 +321,64 @@ The other part of `#285`_ has been modified to use ``-Wpattern-signature-binds``
    this way, ``-XScopedTypeVariables`` does not change from its
    current meaning.
 
-#. Introduce ``-XImplicitForAll``, on by default. With ``-XImplicitForAll``,
-   an out-of-scope type variable mentioned in various constructs (listed below)
-   is implicitly brought into scope over the construct. With ``-XNoImplicitForAll``,
-   this implicit scoping does not happen, and the use of the variable is an error.
+#. Introduce ``-XImplicitBinds``. With ``-XImplicitBinds``, a few sorts of implicit bindings are enabled:
 
-   Constructs affected:
+   #. Implicit forall in positive position type signatures.
 
-   1. Type signatures for variable declarations, methods, and foreign imports & exports.
-      Example: ``let f :: a -> a; f = ... in ...`` becomes
-      ``let f :: forall a. a -> a; f = ... in ...``.
+      With this extension, out-of-scope type variables are implicitly quantified over the following constructs.
+      With ``-XNoImplicitBinds``, this implicit scoping does not happen, and the use of the variable is an error.
 
-   #. Kind signatures. Example: ``type T :: k -> Type`` becomes ``type T :: forall k. k -> Type``.
+      Constructs affected:
 
-   #. GADT constructor declarations. Example: ``MkG :: a -> Maybe b -> G (Either Int b)``
-      becomes ``MkG :: forall a b. a -> Maybe b -> G (Either Int b)``.
+      #. Type signatures for variable declarations, methods, and foreign imports & exports.
+         Example: ``let f :: a -> a; f = ... in ...`` becomes
+         ``let f :: forall a. a -> a; f = ... in ...``.
 
-   #. Pattern synonym signatures. Example: ``pattern P :: a -> Maybe a`` becomes
-      ``pattern P :: forall a. a -> Maybe a``. Implicit quantification in pattern synonyms
-      always produces *universal* variables, never existential ones.
+      #. Kind signatures. Example: ``type T :: k -> Type`` becomes ``type T :: forall k. k -> Type``.
 
-   #. Type annotations in expressions and ``SPECIALISE`` pragmas. Example:
-      ``Right True :: Either a Bool`` becomes ``Right True :: forall a. Either a Bool``.
+      #. GADT constructor declarations. Example: ``MkG :: a -> Maybe b -> G (Either Int b)``
+         becomes ``MkG :: forall a b. a -> Maybe b -> G (Either Int b)``.
 
-   #. Types in a ``deriving`` clause. Example: ``data T deriving (C a)`` becomes
-      ``data T deriving (forall a. C a)``.
+      #. Pattern synonym signatures. Example: ``pattern P :: a -> Maybe a`` becomes
+         ``pattern P :: forall a. a -> Maybe a``. Implicit quantification in pattern synonyms
+         always produces *universal* variables, never existential ones.
 
-   #. Instance heads, including standalone-deriving instances.
-      Example: ``instance Show a => Show (Maybe a)`` becomes
-      ``instance forall a. Show a => Show (Maybe a)``.
+      #. Type annotations in expressions and ``SPECIALISE`` pragmas. Example:
+         ``Right True :: Either a Bool`` becomes ``Right True :: forall a. Either a Bool``.
 
-   #. Type and data family instances, as well as closed type family equations.
-      Example: ``type instance F (Maybe a) = Int``
-      becomes ``type instance forall a. F (Maybe a) = Int``.
+      #. Types in a ``deriving`` clause. Example: ``data T deriving (C a)`` becomes
+         ``data T deriving (forall a. C a)``.
 
-   #. ``RULES`` pragmas.
-      Example: ``{-# RULES "name" forall (x :: Maybe a). foo x = 5 #-}``
-      becomes ``{-# RULES "name" forall a. forall (x :: Maybe a). foo x = 5 #-}``.
-      (The double-\ ``forall`` syntax separates type variables like ``a`` from
-      term variables like ``x``.)
+      #. Instance heads, including standalone-deriving instances.
+         Example: ``instance Show a => Show (Maybe a)`` becomes
+         ``instance forall a. Show a => Show (Maybe a)``.
 
-   This extension is part of accepted, unimplemented proposal `#285`_; the only change is including ``RULES`` pragmas, which @Ericson2314 simply forgot to include in `#285`_ (his own admission).
+      #. Type and data family instances, as well as closed type family equations.
+         Example: ``type instance F (Maybe a) = Int``
+         becomes ``type instance forall a. F (Maybe a) = Int``.
 
-   Being able to turn off this extension is necessary to uphold the `Explicit Binding Principle`_.
+      #. ``RULES`` pragmas.
+         Example: ``{-# RULES "name" forall (x :: Maybe a). foo x = 5 #-}``
+         becomes ``{-# RULES "name" forall a. forall (x :: Maybe a). foo x = 5 #-}``.
+         (The double-\ ``forall`` syntax separates type variables like ``a`` from
+         term variables like ``x``.)
 
-#. Introduce a new warning ``-Wpattern-signature-binds`` (available in ``-Weverything``) that
-   warns whenever an out-of-scope type variable is mentioned in a pattern signature.
+      This is the former ``-XImplicitForAll`` from accepted but unimplemented proposal `#285`_;
+      the only change is including ``RULES`` pragmas, which @Ericson2314 simply forgot to include in `#285`_ (his own admission).
+
+   #. Implicit binds in pattern signatures:
+
+      Out-of-scope type variables written in a pattern signature would be bound there and would remain in scope over the same region of code that term-level variables introduced in a pattern scope over.
+
+      This is the former ``-XPatternSignatureBinds`` from accepted, unimplemented proposal `#285`_.
+
+   #. Implicit binds in kind signatures:
+
+      Out-of-scope type variables written in a negative position kind signatures (positive ones are implicit foralls) are bound as implicit capital lambdas to the left of the parameter they occur in.
+
+      This was intended to be included in the former ``-XPatternSignatureBinds`` from accepted, unimplemented proposal `#285`_, but mistakenly wasn't as these are not "pattern signatures" in the current terminology.
+
+   This extension is on by default for backwards compatibility.
 
 Effects
 ~~~~~~~
@@ -316,8 +390,12 @@ Effects
 Alternatives
 ~~~~~~~~~~~~
 
-1. Previous versions of this proposal, along with the accepted `#285`_, use ``-XNoPatternSignatureBinds`` instead of ``-Wpattern-signature-binds``.
-   However, there seems to be no good reason this must be an extension, instead of a warning.
+1. We could further breaking down ``-XImplicitBinds``, like before.
+
+   But fixing the drafting error would require a *third* extension, ``-XNegativeSignatureBinds``, in addition to the original two.
+   This would allow more conservative defaults --- we must have Haskell98 implicit foralls but not the others which are all guarded behind language extensions today.
+
+   However, @Ericson2314 sense there is a weariness with too many extensions coming from this, and so didn't so it.
 
 Type arguments in constructor patterns
 --------------------------------------
@@ -340,7 +418,7 @@ shadowing any in-scope type variable ``a``. This design supports the
 `Visibility Orthogonality Principle`_, which states that the presence of
 an ``@`` should affect only whether a thing is visible or not, not other
 characteristics (like its shadowing and scoping behavior). Additionally,
-this choice edges us closer to the `Local Lexical Scoping Principle`_,
+this choice edges us closer to the `Lexical Scoping Principle`_,
 because we no longer have to check whether ``a`` is in scope before identifying
 the ``a`` in ``f (Just @a x) = ...`` is a binding site or an occurrence.
 
@@ -501,6 +579,162 @@ This shadowing behavior mimics what happens with term variables in patterns.
   f (Just @a _)  = (5 :: a)
 
 This is accepted. The type variable ``a`` is bound to ``Int``, by pattern-matching.
+
+``-XNoImplicitBinds``
+^^^^^^^^^^^^^^^^^^^^^
+
+Many of these examples also use ``-XTypeAbstractions`` from here and Proposal `#425`_.
+
+Basic examples
+""""""""""""""
+
+#. ::
+
+     f :: t -> ... -- error: `t` is not bound
+     f x = ...
+
+   This could be rewritten as::
+
+     f :: forall t. t -> ...
+     f x = ...
+
+#. ::
+
+     f (x :: t) = ... -- error: `t` is not bound
+
+   This could be rewritten as::
+
+     f :: forall t0. ...
+     f @t (x :: t) = ... -- OK
+
+#. ::
+
+     data Some where
+       MkSome :: forall t. t -> Some
+
+     f (MkSome (x :: t)) = ... -- error: `t` is not bound
+
+   This could be rewritten as::
+
+     data Some where
+       MkSome :: forall t. t -> Some
+
+     f (MkSome @t x) = ... -- OK
+
+Not just term definitions
+"""""""""""""""""""""""""
+
+Besides top level term bindings, we currently have signatures with implicit forall quantification for expressions, data declerations, family declarations, and instances [#class-forall]_.
+This proposal applies to all alike:
+
+#. ::
+
+     ... (id :: t -> t) -- error: `t` is not bound
+
+   This could be rewritten as::
+
+     ... (id :: forall t. t -> t) -- OK
+
+#. ::
+
+    data D :: k -> Type where -- error: `k` is not bound
+
+   This could be rewritten as::
+
+    data D :: forall k. k -> Type where -- OK
+
+#. ::
+
+    type family F :: k -> Type where -- error: `k` is not bound
+
+   This could be rewritten as::
+
+    type family F :: forall k. k -> Type where -- OK
+
+#. ::
+
+    instance Eq t => C t where -- error: `t` is not bound
+
+   This could be rewritten as::
+
+    instance forall t. Eq t => C t where -- OK
+
+When ``-XStandaloneKindSignatures`` is on, these new standalone signatures are affected as well.
+
+#. ::
+
+     type F :: k -> Type -- error: `k` is not bound
+     data F _ = ...
+
+   This could be rewritten as::
+
+     type F :: forall k. k -> Type -- OK
+     data F _ = ...
+
+#. ::
+
+     type F :: k -> k -- error: `k` is not bound
+     type family F where
+
+   This could be rewritten as::
+
+     type F :: forall k. k -> k -- OK
+     type family F where
+
+#. ::
+
+     type C :: (k -> Type) -> Constraint -- error: `k` is not bound
+     class C f where
+
+   This could be rewritten as::
+
+     type C :: forall k. (k -> Type) -> Constraint -- OK
+     class C f where
+
+#. ::
+
+     type D :: k -> Type -- error: `k` is not bound
+     data D where
+
+   This could be rewritten as::
+
+     type D :: forall k. k -> Type -- OK
+     data D where
+
+Pattern signatures in GADT declarations, family declarations, and class declarations are also restricted.
+I'll first use a hypothetical yet-unproposed ``@``-abstraction syntax to "fix" these examples to demonstrate the analogy to the previous examples.
+Then I'll put the inline signature or top-level signature workaround that exist today.
+
+#. ::
+
+     data D (y :: x) (z :: y) where -- error: `x` is not bound, `y` and `z` are fine
+
+   Could be be rewritten as::
+
+     data D @x (y :: x) (z :: y) where -- OK
+
+#. ::
+
+     type family F (y :: x) (z :: y) where -- error: `x` is not bound, `y` and `z` are fine
+
+   Could be be rewritten as::
+
+     type family F @x (y :: x) (z :: y) where -- OK
+
+#. ::
+
+     class Eq a => C (y :: x) (z :: y) where -- error: `x` is not bound, `y` and `z` are fine
+
+   Could be be rewritten as::
+
+     class Eq a => C @x (y :: x) (z :: y) where -- OK
+
+   Note that since there is no ``class F :: ...`` syntax analogous to ``data F :: ...``,
+   so ``-XStandaloneKindSignatures`` are the only way to write explicitly kind-polymorphic classes.
+
+Note that the variables to the left of the ``::`` are are deemed explicit bindings analogous to ``f (y :: x) (z :: z) = ...`` and permitted.
+However ``x`` to the right of the ``::`` is a use, not otherwise bound, and thus implicit binding today.
+It is not permitted as-is, and must be explicitly bound or discarded as done in the working alternatives.
 
 Effects
 ~~~~~~~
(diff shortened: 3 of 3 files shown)


AntC2:
> #523 fixes this, by hinting at (it is not fully specified yet) the `let type var = _` in syntax ...

Note that bit of #523 got taken out -- with no indication if/when it might be reawakened.

Otherwise, your 'meta-proposal' is so meta, I just don't understand what it's saying/or how it's different to all the other swirl of sub- and sub-sub-proposals. As I understand it, those 'principles' proposals are more setting a direction, and explicitly say they can be tweaked in light of concrete/detailed proposals. So perhaps what you want to refine would come across better in context of something concrete?

Perhaps some code examples would help?

principles.rst
```diff
@@ -158,32 +159,108 @@ there is an absence of punning.
...
+   syntax isn't unambiguously a binding site regardless of context (i.e.
+   regardless of whether there is a ``b`` already in scope).
+
+*Motivation:* The `Explicit Binding Principle`_ allows programmers to control exactly how variables come into
+scope.
+It ensures all short-hands can be explained in terms of an explicit, unambiguous equivalent that is easier to understand at the cost of being more verbose:
+
+- Positive-position signatures' free vars cause  implicit ``forall ... .``
+
+- Negative position free vars cause different sorts of binding:
```
tek, line 216: Could you include a code example illustrating the difference between negative and positive position free vars?

principles.rst
```diff
@@ -158,32 +159,108 @@ there is an absence of punning.
...
+#. Template Haskell splices may need to be run before completing name resolution (and running those splices requires type-checking them).
+
+#. The `deprecated mechanism <https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/exts/duplicate_record_fields.html#selector-functions>`_ for disambiguating duplicate record fields violates the `Lexical Scoping Principle`_ by requiring the type system.
+
+#. In a pattern signature, if we have ``f (x :: Maybe a)``, the ``a``
+   is an occurrence if ``a`` is already in scope, and is implicitly bound otherwise.
 
-2. The `deprecated mechanism <https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/exts/duplicate_record_fields.html#selector-functions>`_ for disambiguating duplicate record fields violates the `Lexical Scoping Principle`_ by requiring the type system.
+#. In a type signature, if we have ``f :: a -> a``, the ``a``
+   is an occurrence if ``a`` is already in scope, and it implicitly bound otherwise.
```
tek, line 255:
```suggestion
   is an occurrence if ``a`` is already in scope, and is implicitly bound otherwise.
```

principles.rst
```diff
@@ -158,32 +159,108 @@ there is an absence of punning.
...
 
-3. In a pattern signature, if we have ``f (x :: Maybe a)``, the ``a``
-   is an occurrence if ``a`` is already in scope, and it is a binding site otherwise.
+#. In a kind signature, if we have ``data Foo (a :: k)``, the ``k``
+   is an occurrence if ``k`` is already in scope (historically impossible), and it implicitly bound otherwise.
```
tek, line 258:
```suggestion
   is an occurrence if ``k`` is already in scope (historically impossible), and is implicitly bound otherwise.
```

principles.rst
```diff
@@ -158,32 +159,108 @@ there is an absence of punning.
...
+scope.
+It ensures all short-hands can be explained in terms of an explicit, unambiguous equivalent that is easier to understand at the cost of being more verbose:
+
+- Positive-position signatures' free vars cause  implicit ``forall ... .``
+
+- Negative position free vars cause different sorts of binding:
+
+  - Signatures on term patterns (pattern signatures) cause implicit ``let type ... = _ in``
+
+  - Signatures on type variables (kind signatures) case implicit ``@...``
```
tek, line 220:
```suggestion
  - Signatures on type variables (kind signatures) cause implicit ``@...``
```

principles.rst
```diff
@@ -158,32 +159,108 @@ there is an absence of punning.
...
+#. In a pattern signature, if we have ``f (x :: Maybe a)``, the ``a``
+   is an occurrence if ``a`` is already in scope, and is implicitly bound otherwise.
 
-2. The `deprecated mechanism <https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/exts/duplicate_record_fields.html#selector-functions>`_ for disambiguating duplicate record fields violates the `Lexical Scoping Principle`_ by requiring the type system.
+#. In a type signature, if we have ``f :: a -> a``, the ``a``
+   is an occurrence if ``a`` is already in scope, and it implicitly bound otherwise.
 
-3. In a pattern signature, if we have ``f (x :: Maybe a)``, the ``a``
-   is an occurrence if ``a`` is already in scope, and it is a binding site otherwise.
+#. In a kind signature, if we have ``data Foo (a :: k)``, the ``k``
```
tek, line 257: do you maybe want to include a code example with `data Foo @k (a :: k)` here?

proposals/0448-type-variable-scoping.rst
```diff
@@ -104,12 +105,19 @@ synonym of ``-XScopedTypeVariables``. Under this change, that would no longer be
...
 
-This component includes the ``-XNoImplicitForAll`` of `#285`_ unchanged.
+``-XImplicitBinds`` is the combination of ``-XImplicitBindsForAll`` and ``-XPatternSignatureBinds``) from accepted
+proposal `#285`_, and differ from that proposal in that:
+
+  - The extensions are combined for fewer knobs when the motivations are the same.
+
+  - ``RULES`` was forgotten as an example.
+
+  - Examples of implicit binds falsely categorized as pattern signature binds are now explicitly included.
```
tek, line 117: Concretely, this means that in `data A (a :: k)`, `k` is now in the same category as kind signature occurrences?

proposals/0448-type-variable-scoping.rst
```diff
@@ -127,8 +135,16 @@ two reasons:
...
+A secondary goal is to clean up some issues with proposal `#285`_ while simplifying things:
+
+* ``-XImplicitForAll`` and ``-XPatternSignatureBinds`` have the exact same
+  motivation, and it is unclear why one would ever want one without the other.
+
+* ``-XImplicitForAll`` and ``-XPatternSignatureBinds`` *missed* some of the cases in the examples, which clearly are implicit binding forms meant to be turned off per the overall motivation, but nonetheless slipped through the cracks of the drafting process.
+
 Having separated out ``-XExtendedForAllScope``, it seemed strange to have a ``-XRumpEndOfOldScopedTypeVariables``
 extension, and so I've introduced separate ``-XMethodTypeVariables`` and ``-XPatternSignatures``.
+``-XPatternSignatures`` is more narrow than it's GHC 6.x namesake so as not to violate any principles by default, and to match ``-XKindSignatures`` which is narrow in the same fashion.
```
tek, line 147:
```suggestion
``-XPatternSignatures`` is more narrow than its GHC 6.x namesake so as not to violate any principles by default, and to match ``-XKindSignatures`` which is narrow in the same fashion.
```

proposals/0448-type-variable-scoping.rst
```diff
@@ -182,13 +198,27 @@ conservative; it may actually break my program somewhere!
...
-This is taken directly from `#285`_, updated to refer to warnings instead of language extensions.
-The "I" here is John Ericson.
+Consistency
+"""""""""""
 
-There are two independent motivations: education and consistency with a unified namespace.
+Firstly, it matches ``KindSignatures``.
+``KindSignatures`` doesn't allow implicit binds for a rather roundabout reason: implicit binds would imply implicit kind-level foralls, which would require ``-XPolyKinds``.
+Given the other extensions being proposed here, we can retroactively reinterpret this as a simple syntactic rule: ``-XKindSignatures`` alone doesn't do implicit binding.
+Now, ``PatternSignature``
```
tek, line 210: You appear to have abandoned this sentence

proposals/0448-type-variable-scoping.rst
Ericson2314:
```suggestion
  - Examples of implicit binds falsely categorized as pattern signature binds are now properly included under the proposed change specification.
```

principles.rst
Ericson2314: I was a bit unsure about to what extent to include examples in the proposed change spec.

proposals/0448-type-variable-scoping.rst
```diff
@@ -1045,22 +1279,44 @@ Effects and Interactions
...
+   In many case, those different forms are chiefly distinguished by being confined to one of the type- or term- level, e.g. regular pattern for type synonym right hand sides.
+   A lack of inconsistency on this front therefore is part of the type and term syntax unification espoused by this principle.
+
+#. The `Explicit Binding Principle`_ is made to hold under ``-XNoImplicitBinds`` and ``-XPatternSignatures`` by side-stepping the need for new explicit syntax.
+
+   Making ``-XPatternSignatures`` not imply implicit bindings keeps that exntension in accordance with the `Explicit Binding Principle`_.
+   That principle says implicit binding constructions should have explicit counterparts they desugar to.
+
+   In general, solutions to `Lexical Scoping Principle`_ are also solutions to `Explicit Binding Principle`_.
+   It is just for implicit forms ones wishes to leave enabled that explicit syntax is needed, and explicit syntax forms an additional solution to the `Explicit Binding Principle`_ (and not the `Lexical Scoping Principle`_).
```
tek, line 1320:
```suggestion
   It is just for implicit forms one wishes to leave enabled that explicit syntax is needed, and explicit syntax forms an additional solution to the `Explicit Binding Principle`_ (and not the `Lexical Scoping Principle`_).
```

proposals/0532-type-variable-scoping.rst
```diff
@@ -0,0 +1,116 @@
...
+We were thus able to reach an agreement with @goldfirere to axe the "local lexical scoping principle", and fold it in to the "lexical scoping principle".
+
+This was a great start, but upon rereading the principles (in order to connect the amended proposals to them), I realized that the "lexical scoping principle" an "explicit binding principle" still felt too close, and awkwardly dividing the labor between them.
+
+I thus flipped their order ("explicit binding principle" first) and gave them the following roles (pithily summarized):
+
+#.
+**Explicit Binding Principle**: Every implicit binding form needs an explicit equivalent.
+#.
+**lexical scoping principle**: Must be possible to turn off implicit binding forms so only the explicit ones remain.
```
tek, line 48:
```suggestion
**Lexical scoping principle**: Must be possible to turn off implicit binding forms so only the explicit ones remain.
```

Ericson2314:
> > #523 fixes this, by hinting at (it is not fully specified yet) the `let type var = _` in syntax ...
> 
> Note that bit of #523 got taken out -- with no indication if/when it might be reawakened.

I am not referring to the removed `let` in patterns, but just regular `let ... in` in expressions.

> Otherwise, your 'meta-proposal' is so meta, I just don't understand what it's saying/or how it's different to all the other swirl of sub- and sub-sub-proposals. As I understand it, those 'principles' proposals are more setting a direction, and explicitly say they can be tweaked in light of concrete/detailed proposals. So perhaps what you want to refine would come across better in context of something concrete?

Yes it is "so meta" :). The underlying proposals should still stand alone. The meta-proposal is just supposed to be for someone thing that "well the old and new versions for 448 both seem fine to me, how did we decide between them?".

> Perhaps some code examples would help?

There are new examples in 448, do note.

principles.rst
tek: I don't remember what my intention was here, better disregard my comment

tek: ah, it was because of "historically impossible", to illustrate how it would be possible – but this has already been demonstrated above

proposals/0425-decl-invis-binders.rst
```diff
@@ -621,6 +621,15 @@ The proposed changes provide the programmer with a more principled way of
 brining type variables into scope in certain corner cases, simplify arity
 inference and scoping rules.
 
+This gets us closer to upholding two principles:
+
+- The `Explicit Binding Principle`_, because the new ``@``-binders is explicit unlike previous alternatives
```
tek, line 626:
```suggestion
- The `Explicit Binding Principle`_, because the new ``@``-binders are explicit unlike previous alternatives
```

proposals/0425-decl-invis-binders.rst
```diff
@@ -621,6 +621,15 @@ The proposed changes provide the programmer with a more principled way of
...
+
+- The `Lexical Scoping Principle`_, because the new simplified rules about variables needing to occur on the LHS in certain situations.
```
tek, line 628:
```suggestion
- The `Lexical Scoping Principle`_, because of the new simplified rules about variables needing to occur on the LHS in certain situations.
```

proposals/0448-type-variable-scoping.rst
Ericson2314:
```suggestion
   It is just for implicit forms which one wishes to leave enabled that explicit syntax is needed, and explicit syntax forms an additional solution to the `Explicit Binding Principle`_ (and not the `Lexical Scoping Principle`_).
```

proposals/0448-type-variable-scoping.rst
```diff
@@ -198,42 +228,74 @@ All other variables are explicitly bound, and the inconsistency means more to le
...
   t = Int
   x :: t -- sugar for 'forall t. t', not 't ~ Int'
   x = 0
 
-Should the ``t`` in ``x :: t`` cause an implicit ``forall t.`` to be synthesized or not? With ``-XNoImplicitForAll``, we know
-it will not, and thus can refer to the ``t`` defined above, once such a reference is possible (left to another proposal).
+  t = Int
+  foo (x :: t) = 0 -- sugar for 'foo = let t = _ in \(x :: t) -> 0'
+
+Should the ``t`` in ``x :: t`` cause implicit ``forall t.`` and ``let t = I in`` to be synthesized or not?
```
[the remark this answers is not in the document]
Ericson2314: Actually it shouldn't be, but I agree it was unclear. Hopefully better now.

proposals/0448-type-variable-scoping.rst
```diff
@@ -1045,22 +1279,44 @@ Effects and Interactions
...
    in `#126`_, where the choice between a binding site and an occurrence depends on whether a
    type variable is in scope.
 
-#. The `Explicit Variable Principle`_ is made to hold, by allowing explicit binders for type variables
-   for existentials and the variables bound by an inner ``forall`` in a higher-rank
-   type.
+#. The `Syntactic Unification Principle`_ is bolstersted by ``-XNoImplicitBinds``
+
+   As discussed in the "Consistency" section of the motivation for that extension, the different forms of implicit binding we have today work quite differently.
+   In many case, those different forms are chiefly distinguished by being confined to one of the type- or term- level, e.g. regular pattern for type synonym right hand sides.
```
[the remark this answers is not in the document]
Ericson2314: Rewrote and split. Hopefully better now.

Ericson2314:
In https://gitlab.haskell.org/ghc/ghc/-/merge_requests/8629#note_449873 @goldfirereprovided the missing context of how the previous revisions came about, and also signalled (If I understand correct) basic agreement with the thrust of this change. Horray!

Since this is an "in the weeds" modification to an existing proposal, and since @goldfirere (the original author, to be clear) doesn't need convincing, I am submitting this straight to the committee.

CC @nomeata

nomeata:
Since this builds on top of #448, I’d say @aspiwack is the natural shepherd.

aspiwack:
As the shepherd for this proposal, I am unwilling to bring this proposal to the committee in its current shape. It does too many things, and it will be impossible to have an effective discussion about it.

To summarize (I hope I'm not forgetting anything), the current PR:
1. modifies the current list of design principles
2. changes an already-accepted proposition to refer to these principles
3. modifies #448 to replace the extension `-XImplicitForall` with `-XImplicitBinds` and fold what was previously a warning into it
4. changes some language that @Ericson2314 doesn't like

In order to yield productive discussions and informed decisions, the proposal should be more focused:
- (1) and (3) should be different proposals.
- You shouldn't do (2): it doesn't make much sense to. It's akin to rewriting history and claiming that a proposal was meant to meet a goal that didn't exist at the time
- You shouldn't do (4) either: there is not enough value in this. Frankly this is not worth your time, or the time of the committee. Only change things that need changing.

I'll mark this proposal as https://github.com/ghc-proposals/ghc-proposals/labels/Needs%20revision in the meantime.

Ericson2314:
> changes some language that @Ericson2314 doesn't like

This is vague. I don't know what parts you want me to change.

> (1) and (3) should be different proposals.

The changes to #448 reference the modified proposals. It was also said that proposals should not be modified on their own, only in conjunction with the proposals.

Perhaps @gridaphobe can help me with the principles? I don't want to gratuitously change things, but while the current proposals list at least seems no longer redundant, I am still somewhat baffled by the division of labor between the **explicit binding principle** and **lexical scoping principle**. If it is indeed unclear, it should be fixed, and it helps ensure the motivation for the modified #448 makes sense. If I am just missing something and it is already makes sense then I can just drop this and adjust #448 accordingly.

proposals/0448-type-variable-scoping.rst
```diff
@@ -198,42 +233,74 @@ All other variables are explicitly bound, and the inconsistency means more to le
...
+which happen to be true in some simple common cases, but are in fact incorrect in general.
+
+That it takes a complicated example to show why these false desugarings aren't true in general make this is a huge educational stumbling block!
 
 Unified Namespace
 ^^^^^^^^^^^^^^^^^
 
 If `#270`_ is accepted, there will be a way to program Haskell with "morally" one namespace for types and terms alike.
-However, there is one exception to the unification of namespaces: lower case variables in type signatures bound "like terms" still are treated as free and implicitly bound with a ``forall`` instead::
+However, there is one exception to the unification of namespaces: lower case variables in type signatures bound "like terms" still are treated as free and implicitly bound instead::
```
aspiwack, line 276: .

proposals/0448-type-variable-scoping.rst
```diff
@@ -198,42 +233,74 @@ All other variables are explicitly bound, and the inconsistency means more to le
...
 
   t = Int
-  x :: t -- sugar for 'forall t. t', not 't ~ Int'
+  x :: t -- sugar for 'forall t. t', not a use of 't' resolving to 'Int'
```
aspiwack, line 279: .

proposals/0448-type-variable-scoping.rst
```diff
@@ -198,42 +233,74 @@ All other variables are explicitly bound, and the inconsistency means more to le
...
   x = 0
 
-Should the ``t`` in ``x :: t`` cause an implicit ``forall t.`` to be synthesized or not? With ``-XNoImplicitForAll``, we know
-it will not, and thus can refer to the ``t`` defined above, once such a reference is possible (left to another proposal).
+  t = Int
+  foo (x :: t) = 0 -- sugar for 'foo = let t = _ in \(x :: t) -> 0'
+
+Should the ``t`` in each ``x :: t`` cause implicit ``forall t.`` and ``let t = _ in`` to be synthesized or not?
```
aspiwack, line 285: .

proposals/0448-type-variable-scoping.rst
```diff
@@ -198,42 +233,74 @@ All other variables are explicitly bound, and the inconsistency means more to le
...
-The other part of `#285`_ has been modified to use ``-Wpattern-signature-binds``.
+``-XScopedTypeVariables`` come from not-yet-accepted proposal `#238`_.
+``-XImplicitBinds`` is a fixed and simplified (via combining extensions) version of accepted proposal `#285`_.
 
 1. Re-purpose deprecated extension ``-XPatternSignatures``. With ``-XPatternSignatures``, we
    allow type signatures in patterns. These signatures can mention in-scope
-   type variables as variable occurrences. A mention of an out-of-scope variable
-   binds the type variable as an alias of the type it is unified with (as today).
+   type variables as variable occurrences, but can not bind type variables without the
+   separate ``-XImplicitBinds`` extension. Do note that extension is on by default, however.
```
aspiwack, line 303: .

aspiwack (commented):
(apologies for the delayed response. I got distracted when I got the notification, and forgot to get back to you, sorry)

> The changes to https://github.com/ghc-proposals/ghc-proposals/pull/448 reference the modified proposals. It was also said that proposals should not be modified on their own, only in conjunction with the proposals.

I think that there are some occurrences of “proposal” that should be principle, but I'm sorry I can't make out which.

> > changes some language that @Ericson2314 doesn't like
> 
> This is vague. I don't know what parts you want me to change.

Fair enough. This was quite careless wording on my part, sorry. I meant changes that do not significantly affect the meaning of the proposal. Truth is, there are fewer such places than I initially thought. A lot of what I thought was rewordings of this nature was noise introduced by the diffing algorithm which is not really tailored for text. But there are still a few, I've highlighted some below.

The problem is not that these are bad. The problem is that they make it harder to understand your proposed change.

proposals/0448-type-variable-scoping.rst
Ericson2314: Ah this is also an artifact of the bad diffing I am afraid; there is a pattern signature binds examples added below which does *not* desugar to a `forall`. I therefore had to reword this to describe both examples at once.

proposals/0448-type-variable-scoping.rst
Ericson2314: Perhaps I was being too liberal modifying this because...it is in fact my own writing taking verbatim from #285 here that I am nit-picking!

The issue that is that `t` is *not* sugar for `t ~ Int`, because `t` is a type and `t ~ Int` is a constraint. I was in #285 trying to express that `t` is bound as `Int` but did so in what strikes me in hindsight as a sloppy manner.

I suppose I could do this in a separate PR or not at all. Let me know what you prefer.

proposals/0448-type-variable-scoping.rst
Ericson2314: The sentence that appears removed is on the line below.

Ericson2314:
> Apologies for the delayed response.

Likewise! I suspect I am not taking much long :/. Sorry about that.

> Truth is, there are fewer such places than I initially thought. A lot of what I thought was rewordings of this nature was noise introduced by the diffing algorithm which is not really tailored for text. But there are still a few, I've highlighted some below.
>
> The problem is not that these are bad. The problem is that they make it harder to understand your proposed change.

This makes sense, and is a huge relief to read!

I habitually break up prose in version control I write/edit on punctuation boundaries precisely because it makes the diff better *going forward*. But of course the first such edit until that is established is bad either way (either split line confuses it, or long line obscures it)!

Happy to respond to each of those, and rework things to lessen the diff.

Ericson2314:
One crazier-at-first option would be to first reformat the entire document to be broken up on punctuation and *not* line-length boundaries, and then reapply my edits. That would make one crazy diff but with no actual content changes. Then, if this is rebased on top, it should be much more readable.

If you like that, I would be happy to do that.

proposals/0448-type-variable-scoping.rst
Ericson2314: Isn't this more than a rewording? Before, implicit pattern signature binds were gated by `-XPatternSignatures` alone, but warned. Now, they are gated by the combination of `-XPatternSignatures` and `-XImplicitBinds`.

Yes, the difference in practice is subtle because the latter is on by default, but generally when I write these things I try to specify the effects of extensions and the defaulting of extensions separately. (Dealing with both at once I find confusing.)

Ericson2314:
@aspiwack ping.

@monoidal Reminded me it would be nice to resolve this in case we could make 9.6.

aspiwack:
Argl! I haven't replied here for two months. They've been crazy two months on my end, but nevertheless I apologise profusely. This is inexcusably long.

Honestly, I don't think that breaking at punctuation boundary would affect the diff in a significant way.

The way I see it is that there are three parts to this proposal
- A proposal to improve the principles
- `-XImplicitBindings`
- Some minor rewrites (which is where I'm getting a bit lost)

What you would like, I think, is to have `-XImplicitBindings` for 8.6. I think that the only way be able to expedite this process is to take `-XImplicitBindings` out as a standalone proposal. Which I can commit to a fast shepherding cycle.

Ericson2314:
I am sympathetic to trying to speed things along with next material but as I see it:

1. The proposals right now don't really make much sense and are overfit to the old plan; leaving them as is would bring us *out of compliance* with those principles. Is that OK? It feels "chaotic good" to me at best.
2. The 3 extraneous rewrites you flagged above are in fact more substantial things which the bad diffing disguised.

What I would recommend is that

1. I *try* reformatting to improve the diff. It might not work but if so it's just my time wasted.
2. @gridaphobe co-shepherds the proposal for the principles part, as @gridaphobe was also interested in the principles / shared my confusion.
3. You can just worry about `-XImplicitBinds`.

I think that approach would work OK?

aspiwack:
I don't think that this is a shepherding problem, really. The shepherd will spend more time on the proposal than the rest of the committee, hence will gain a deeper understanding. The problem that I'm envisioning is that there is a strong philosophical component to this proposal, which usually take a lot of time to converge into a collective decision. I'm not convinced that you need to alter the principles to motivate your change. If you insist that you do, I'll bring the whole proposal to the committee's attention.

Ericson2314:
@aspiwack What you say makes sense.

I *think* there really is an awkwardness with the existing principles that is brought out, but yes perhaps it could also be me shoe-horning my own philosophies in, blinded by fervor and thus missing that things  really are orthogonal.

I don't think I am doing that --- I certainly hope not! --- but perhaps the committee can help us decide :). I'll give it another pass in the meantime and make sure everything is as clear as it can be.

Ericson2314:
(FYI I made https://github.com/ghc-proposals/ghc-proposals/pull/553 for my hunch that the diff is bad now and could be better due to arbitrary line breaks. If that is merged, then I will push to this fixing conflicts and revealing what is hopefully a better diff.)

Ericson2314:
I realized in the 532 proper "summary of changes" I never did connect the principle changes and proposal proper changes! No wonder my intentions were not coming through! Pushed a new section to explicitly make the connection.

principles.rst
Ericson2314: I removed historically impossible because even more `@k` there is plain `k`: `data Foo k (a :: k)` has worked since 8.0! No need worrying about things older than that at this point. I don't think.

Ericson2314:
Per the above I believe we are now waiting on the committee and I have also made some revisions too, so I think the "needs revision" tag is inaccurate?

nomeata:
I take that as a resubmission, which puts the ball into @aspiwack' court.

Ericson2314:
I think per the above @aspiwack kindly offered to take it to the committee despite his reservations on the principles stuff, but sure with the new motivation material I added explaining why that material is needed, I suppose he can reconsider what his recommendation is too.

-----------

Reading the principles again, it now seems my amendment of the **Explicit Binding Principle** makes it quite close the **Explicit Variables Principle**. Perhaps there is only enough material (however it is sliced and diced) for 2 principles, not 3 as originally and as currently. But I will hold myself back from making yet more changes to the principles document!

proposals/0448-type-variable-scoping.rst
```diff
@@ -331,8 +415,12 @@ Effects
 Alternatives
 ~~~~~~~~~~~~
 
-1. Previous versions of this proposal, along with the accepted `#285`_, use ``-XNoPatternSignatureBinds`` instead of ``-Wpattern-signature-binds``.
-   However, there seems to be no good reason this must be an extension, instead of a warning.
+1. We could further break down ``-XImplicitBinds``, like before.
+
+   But fixing the drafting error would require a *third* extension, ``-XNegativeSignatureBinds``, in addition to the original two.
```
aspiwack, line 420: I think that you are unambiguously allowed to propose modification to unimplemented proposal.

proposals/0448-type-variable-scoping.rst
```diff
@@ -253,74 +323,88 @@ The other part of `#285`_ has been modified to use ``-Wpattern-signature-binds``
...
+         becomes ``{-# RULES "name" forall a. forall (x :: Maybe a). foo x = 5 #-}``.
+         (The double-\ ``forall`` syntax separates type variables like ``a`` from
+         term variables like ``x``.)
+
+      This is the former ``-XImplicitForAll`` from accepted but unimplemented proposal `#285`_;
+      the only change is including ``RULES`` pragmas, which @Ericson2314 simply forgot to include in `#285`_ (his own admission).
+
+   #. Implicit binds in pattern signatures:
+
+      Out-of-scope type variables written in a pattern signature would be bound there and would remain in scope over the same region of code that term-level variables introduced in a pattern scope over.
```
aspiwack, line 397: This is lacking an example.

proposals/0448-type-variable-scoping.rst
```diff
@@ -253,74 +323,88 @@ The other part of `#285`_ has been modified to use ``-Wpattern-signature-binds``
...
+
+      This is the former ``-XPatternSignatureBinds`` from accepted, unimplemented proposal `#285`_.
+
+   #. Implicit binds in kind signatures:
+
+      Out-of-scope type variables written in a negative position kind signature (positive ones are implicit foralls) are bound as implicit capital lambdas to the left of the parameter they occur in.
```
aspiwack, line 403: This is lacking an example as well.

proposals/0448-type-variable-scoping.rst
```diff
@@ -349,10 +437,8 @@ By contrast, the implementation errors in this case.
...
 Accordingly, ``f (Just @a x) = ...`` would always, unconditionally bind a new type variable ``a``, possibly shadowing any in-scope type variable ``a``.
-This design supports the `Visibility Orthogonality Principle`_,
-which states that the presence of an ``@`` should affect only whether a thing is visible or not, not other characteristics (like its shadowing and scoping behavior).
-Additionally, this choice edges us closer to the `Local Lexical Scoping Principle`_,
-because we no longer have to check whether ``a`` is in scope before identifying the ``a`` in ``f (Just @a x) = ...`` is a binding site or an occurrence.
+This design supports the `Visibility Orthogonality Principle`_, which states that the presence of an ``@`` should affect only whether a thing is visible or not, not other characteristics (like its shadowing and scoping behavior).
+Additionally, this choice edges us closer to the `Lexical Scoping Principle`_, because we no longer have to check whether ``a`` is in scope before identifying the ``a`` in ``f (Just @a x) = ...`` is a binding site or an occurrence.
```
aspiwack, line 441: This is no change here, correct?

proposals/0448-type-variable-scoping.rst
```diff
@@ -414,7 +500,19 @@ Proposed Change Specification
...
+
+   * Outside pattern signatures,
+     variable occurrences are considered binding sites,
+     shadowing any other in-scope variables.
+     It is an error to bring the same type variable into scope in two (or more) places within the same match group.
+
+   * Inside pattern signatures (i.e. on the right-hand side of ``pat :: sig``),
+     occurrences of in-scope type variables are usages,
+     whereas occurrences of out-of-scope type variables create implicit bindings.
+     It is allowed to mention the same out-of-scope variable more than once.
```
aspiwack, line 515: This is unrelated to the current principal, isn't it?

proposals/0448-type-variable-scoping.rst
```diff
@@ -429,8 +527,6 @@ Proposed Change Specification
    it does not trigger any behavior associated with partial type signatures.
    In particular, ``-XPartialTypeSignatures`` is not necessary, and no diagnostic is produced.
 
-#. As with term variables, it is an error to bring the same type variable into scope in two (or more) places within the same pattern.
-
```
aspiwack, line 433: Why did you remove this?

proposals/0448-type-variable-scoping.rst
```diff
@@ -977,20 +1238,42 @@ Effects and Interactions
...
 
-1. We get closer to the `Lexical Scoping Principle`_:
-   with ``-Werror=pattern-signature-binds``, type variables cannot be bound in pattern signatures,
-   closing one of the places where the `Lexical Scoping Principle`_ is currently violated.
+#. The `Explicit Variable Principle`_ is made to hold, by allowing explicit binders for type variables for existentials and the variables bound by an inner ``forall`` in a higher-rank
+   type.
+
+#. The `Lexical Scoping Principle`_ is made to hold with ``-XNoImplicitBinds``
+
+   Indeed, the purpose of ``-XNoImplicitBinds`` is to be the single extension which is both necessary and sufficient to do this.
```
aspiwack, line 1246: Well, not completely: the Template Haskell objection still holds I think.

proposals/0448-type-variable-scoping.rst
```diff
@@ -977,20 +1238,42 @@ Effects and Interactions
...
+
+   The `Lexical Scoping Principle`_, part (a), is upheld.
+   Binders occur in patterns, after ``forall``, in
+   ``let`` declarations, and a few other discrete places in the AST -- and
+   nowhere else. In particular, binders do not occur in pattern signatures.
+
+   The `Lexical Scoping Principle`_ part (a) is made to hold,
```
aspiwack, line 1253:
```suggestion
   The `Lexical Scoping Principle`_ part (b) is made to hold,
```

aspiwack (commented):
I have a few comments below. I'll make a proposal to the committee when they are addressed.

proposals/0425-decl-invis-binders.rst
```diff
@@ -621,6 +621,15 @@ The proposed changes provide the programmer with a more principled way of
...
 
+This gets us closer to upholding two principles:
+
+- The `Explicit Binding Principle`_, because the new ``@``-binders are explicit unlike previous alternatives
+
+- The `Lexical Scoping Principle`_, because of the new simplified rules about variables needing to occur on the LHS in certain situations.
+
+.. _`Explicit Binding Principle`: ../principles.rst#explicit-binding-principle
+.. _`Lexical Scoping Principle`: ../principles.rst#lexical-scoping-principle
+
```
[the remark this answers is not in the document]
Ericson2314: Done. Principles now refers to #425 enough that I can drop this while still making note of the relationship.

proposals/0448-type-variable-scoping.rst
```diff
@@ -179,13 +197,33 @@ it may actually break my program somewhere!
...
+This more narrow formulation of ``-XPatternSignatures`` matches ``KindSignatures``.
+``KindSignatures`` doesn't allow implicit binds for a rather roundabout reason: implicit binds would imply implicit kind-level foralls, which would require ``-XPolyKinds``.
+Given the other extensions being proposed here, we can retroactively reinterpret this as a simple syntactic rule: ``-XKindSignatures`` alone doesn't do implicit binding.
+Now, both extension just allow, respectively, term-level and type-level signatures, with no other functionality like implicit binding mechanisms also thrown in.
+
+Conservativism for standardization
+""""""""""""""""""""""""""""""""""
+
+With both of these signatures being very minimal, I think they would be easy uncontroversial candidates for a new language report.
+Conversely, all implicit binding constructs are very fraught with a complicated mix of upsides and downsides, we and should only standardize them with great care.
```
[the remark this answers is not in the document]
Ericson2314: It is possible I duplicated something by mistake. I will double check, thanks.

proposals/0448-type-variable-scoping.rst
Ericson2314: Oh whoops, I accidentally pushed my merging of https://github.com/ghc-proposals/ghc-proposals/pull/556 into this to fix comments.

proposals/0448-type-variable-scoping.rst
Ericson2314: Oh whoops, I accidentally pushed my merging of https://github.com/ghc-proposals/ghc-proposals/pull/556 into this to fix comments.

That one deleted this because it is not quite correct and the expanded above portion now covers it.

proposals/0448-type-variable-scoping.rst
Ericson2314:
```suggestion

#. The `Lexical Scoping Principle`_ outside of Template Haskell is made to hold with ``-XNoImplicitBinds``

   Indeed, the purpose of ``-XNoImplicitBinds`` is to be the single extension which is both necessary and sufficient to do this.
   (The Template Haskell issue is something would be solved by more rigorous notations of hygiene. That has little to do with ``-XNoImplicitBinds`` as currently specified, and is much more work of a very different nature.) 
```

proposals/0532-type-variable-scoping.rst
```diff
@@ -0,0 +1,124 @@
...
+We were thus able to reach an agreement with @goldfirere to axe the "local lexical scoping principle", and fold it in to the "lexical scoping principle".
+
+This was a great start, but upon rereading the principles (in order to connect the amended proposals to them), I realized that the "lexical scoping principle" an "explicit binding principle" still felt too close, and awkwardly dividing the labor between them.
+
+I thus flipped their order ("explicit binding principle" first) and gave them the following roles (pithily summarized):
+
+#.
+**Explicit Binding Principle**: Every implicit binding form needs an explicit equivalent.
+#.
+**Lexical scoping principle**: Must be possible to turn off implicit binding forms so only the explicit ones remain.
```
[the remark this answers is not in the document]
Ericson2314:
```suggestion
#. **Explicit Binding Principle**: Every implicit binding form needs an explicit equivalent.
#. **Lexical scoping principle**: Must be possible to turn off implicit binding forms so only the explicit ones remain.
```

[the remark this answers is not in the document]
Ericson2314: I was too eagerly adding newlines after `.`! :)

proposals/0448-type-variable-scoping.rst
Ericson2314: Forced pushed to remove merge.

principles.rst
```diff
@@ -168,26 +234,38 @@ Lexical Scoping Principle (LSP)
...
 The `Lexical Scoping Principle`_ is almost true today, with the following nuances:
 
-1. Template Haskell splices may need to be run before completing name resolution (and running those splices requires type-checking them).
+#. Template Haskell splices may need to be run before completing name resolution (and running those splices requires type-checking them).
 
-2. The `deprecated mechanism <https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/exts/duplicate_record_fields.html#selector-functions>`_ for disambiguating duplicate record fields violates the `Lexical Scoping Principle`_ by requiring the type system.
+#. The `deprecated mechanism <https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/exts/duplicate_record_fields.html#selector-functions>`_ for disambiguating duplicate record fields violates the `Lexical Scoping Principle`_ by requiring the type system.
 
-3. In a pattern signature,
+#. In a pattern signature,
```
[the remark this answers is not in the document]
Ericson2314: Great point. Did that.

proposals/0532-type-variable-scoping.rst
```diff
@@ -0,0 +1,124 @@
...
+.. author:: John Ericson
+.. date-accepted:: 
+.. ticket-url:: 
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/523>`_.
+.. sectnum::
+.. contents::
+
+I believe #448 landed with some ill-advised decisions, and I wish to correct them.
```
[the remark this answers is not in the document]
Ericson2314: Good point. Done.

proposals/0448-type-variable-scoping.rst
Ericson2314: Ah just one word. It was referring to the "Local Lexical Scoping Principle" which was never merged.

proposals/0448-type-variable-scoping.rst
Ericson2314: My point wasn't that I felt that I couldn't change it, but that I didn't want to be responsible for for *even more* language extensions. 448 already adds a lot already, and I didn't want to in modifying the proposal give credence to that objection to 448.

proposals/0532-type-variable-scoping.rst
```diff
@@ -0,0 +1,124 @@
...
+
+Concretely, for educational purposes we want to get rid of implicit bindings, and get rid of the *concept* of implicit binding.
+We want the student to be *unable* to write them, and we want the compiler to *not* tell them with other configuration options feature exists, the same way a Haskell 98 users should not be told about "type families" or weather.
+
+Single namespace syntactic uniformity
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+#270 has a very nice story about making single name-space code not fork-like by accepting fewer programs.
+``-Werror=pattern-signature-binds`` either breaks the "non fork-like" condition, or breaks the "single namespace" condition.
+Either is not acceptable.
```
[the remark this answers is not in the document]
Ericson2314: Yes good point. It was a bit tricky to write (my own memory went stale) but I have it now --- glad I made it explicit!

proposals/0448-type-variable-scoping.rst
```diff
@@ -179,13 +197,33 @@ it may actually break my program somewhere!
...
+We here discuss the motivation for this decision.
 
-There are two independent motivations: education and consistency with a unified namespace.
+Consistency
+"""""""""""
+
+This more narrow formulation of ``-XPatternSignatures`` matches ``KindSignatures``.
+``KindSignatures`` doesn't allow implicit binds for a rather roundabout reason: implicit binds would imply implicit kind-level foralls, which would require ``-XPolyKinds``.
+Given the other extensions being proposed here, we can retroactively reinterpret this as a simple syntactic rule: ``-XKindSignatures`` alone doesn't do implicit binding.
+Now, both extension just allow, respectively, term-level and type-level signatures, with no other functionality like implicit binding mechanisms also thrown in.
```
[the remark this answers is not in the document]
Ericson2314: Done!

principles.rst
```diff
@@ -160,6 +161,71 @@ there is an absence of punning.
...
+Explicit Binding Principle (EBP)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. _`Explicit Binding Principle`:
+
+**Principle**:
+Through the right combination of extensions,
+every implicit form of variable binding must have an explicit equivalent that,
+regardless of the context,
+is unambiguously a binding site.
```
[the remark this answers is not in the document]
Ericson2314: The key part is "regardless of the context, is unambiguously a binding site."

I agree actually that whether pattern signature binds are implicit or not is a squishy matter of opinion. So I am trying to step back say it is might be *explicit* but it is not *explicit and unambiguous*.

I think *explicit and unambiguous* is a more objective standard: the syntax needs to mean one thing, regardless of context. Pattern signature binds do not meet that standard.

principles.rst
```diff
@@ -168,26 +234,38 @@ Lexical Scoping Principle (LSP)
...
+#. In a pattern signature,
    if we have ``f (x :: Maybe a)``,
    the ``a`` is an occurrence if ``a`` is already in scope,
-   and it is a binding site otherwise.
+   and is implicitly bound otherwise.
+
+#. In a type signature,
+   if we have ``f :: a -> a``,
+   the ``a`` is an occurrence if ``a`` is already in scope,
+   and is implicitly bound otherwise.
```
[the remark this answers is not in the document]
Ericson2314: See what I wrote below in https://github.com/ghc-proposals/ghc-proposals/pull/532#discussion_r1035161855 . The **Explicit Binding Principle** is was bit wish-washy before on whether it or the **Lexical Scoping Principle** said it was no good. But now I am making the Explicit Binding Principle be about something else, "explicit *and unambiguous*" bindings, so it squarely falls under its purview.

-------

A natural follow-up question might be "that's great but who cares which principle catches the violation? Why does it matter?"

Well, my next thought it that banning implicit bindings, like promoting pun-free code (#270) is probably controversial. However, what would be less controversial is saying we have explicit option to teach/explain/desugar all our implicit options.

The **Explicit Binding Principle** I am trying to make the first, not second, principle, because it alone should be about that point. It (along with the Explicit Variable Principle, which is almost the same), says we need to have `\ @a -> ..`, `\(DataCon @a) -> ...`, `let type a = _ in ...`, etc. so long as we have their implicit counterparts. This I hope is not controversial! We should have syntax to be able state exactly what we mean. I would hope we could even standardize many of these at some point in a future language report.

The **Lexical Scoping Principle**, by contrast, has the controversial residual, which is merely *having* explicit counterparts isn't good enough, and for maximally easy *reading* we need to also be able to ensure the implicit/ambiguous ones are not used. I personally wouldn't mind if `-XImplicitBinds` went away and the language just became Agda style where we must be explicit, but I don't expect the Haskell community writ large to be persuaded of this any time soon. So the **Lexical Scoping Principle** keeps the controversial bits and tempers them down to being we should merely have the *option* to disable implicit/ambiguous constructs.

principles.rst
```diff
@@ -168,26 +234,38 @@ Lexical Scoping Principle (LSP)
...
 
-2. The `deprecated mechanism <https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/exts/duplicate_record_fields.html#selector-functions>`_ for disambiguating duplicate record fields violates the `Lexical Scoping Principle`_ by requiring the type system.
+#. The `deprecated mechanism <https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/exts/duplicate_record_fields.html#selector-functions>`_ for disambiguating duplicate record fields violates the `Lexical Scoping Principle`_ by requiring the type system.
 
-3. In a pattern signature,
+#. In a pattern signature,
    if we have ``f (x :: Maybe a)``,
    the ``a`` is an occurrence if ``a`` is already in scope,
-   and it is a binding site otherwise.
+   and is implicitly bound otherwise.
```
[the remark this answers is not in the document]
Ericson2314: See both replies below.

I don't mind putting it back the way it was, but I do think it's important for us deliberating the proposal to understand that even if isn't an "implicit binding", it *is* a violation of the **Explicit Binding Principle** because it is not explicit *and unambiguous*. (By which I mean the context doesn't matter as to whether it's a binding or not.)

Would it be better to rename this the **Explicit and Unambiguous Binding Principle**, or the **Context-free Binding Principle**?

proposals/0448-type-variable-scoping.rst
```diff
@@ -195,42 +233,74 @@ All other variables are explicitly bound, and the inconsistency means more to le
...
+Should the ``t`` in each ``x :: t`` cause implicit ``forall t.`` and ``let t = _ in`` to be synthesized or not?
 
-Motivation for ``-Wpattern-signature-binds``
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+With ``-XNoImplicitBinds``, we know it will not, and thus can refer to the ``t`` defined above, once such a reference is possible (left to another proposal).
 
-This is necessary in order to uphold the `Lexical Scoping Principle`_, part (a).
+Without ``-XImplicitBinds`` we have no choice but do the implicit desugaring that violates the unified namespace abstraction
+--- otherwise we would be *forklike* in the sense that rather than giving new meaning to previously invalid programs, we would be *changing* the meaning of previously *valid* programs.
+Put another way, ``-XImplicitBinds`` and any ``-XUnifiedNamespace`` must be mutually exclusive, so we need a way for Haskell programmers to opt out of ``-XImplicitBinds`` first.
```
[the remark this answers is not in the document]
Ericson2314: The issue is that we start with an invalid program, and we make it valid in two *different* ways, one with liberalized namespace lookup and the "term 2 type translate", the other with implicit binding, but there is no monotonic way to reconcile those things. Each extension is non-fork-like in isolation, but for the combination we are stuck.

Ericson2314:
Thank you for the through review @aspiwack. I believe I have either fixed or left replies (thankfully mostly the former!) on everything you pointed out, so passing the ball back to you.

principles.rst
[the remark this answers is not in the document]
aspiwack: I must register here that your proposal makes the bindings more confusing to me than they currently are. You are free to stick to your guns there, it will however colour my recommendation.

To me the context-freeness was the purview of the lexical binding principle; while the presence of a binding site in the text was the domain of the explicit binding principle. You're making the distinction less clear to me, I can't quite understand your point of view yet. I may very well be missing something.

proposals/0448-type-variable-scoping.rst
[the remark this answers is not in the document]
aspiwack: Ah, this is quite interesting. Can you make this more explicit and detailed in the proposal? It's not obvious to me yet what's happening.

proposals/0448-type-variable-scoping.rst
aspiwack: This is not what you're saying though. As it currently stands, I think it's rather misleading in fact.

proposals/0532-type-variable-scoping.rst
```diff
@@ -0,0 +1,161 @@
...
+
+   but the "illusion" of a single namespace is shattered, because ``a`` and ``B`` are *not* treated the same way.
+
+   Furthermore, as a practical matter, how *should* ``a`` be used here?
+   ``-Werror=pattern-signature-binds`` will catch and error on our pattern signature bind, but it won't allow us to refer to ``a``.
+   Not being able to use without rebinding it leaves an awkward tripping point that must be worked around with more variable indirection.
+
+With ``-XNoImplicitBinds`` instead, we go with option 2 to ``-Wpuns`` and ``-Wpun-binding`` stay genuine warnings that do not change behavior.
+``-XImplicitBinds`` however *can* change behavior, and it results in the same behavior (unification variable) regardless of ``-Wpuns`` and ``-Wpun-binding``.
+Documentation can make clear that, yes, the punning warnings are insufficient on their own to create the single-namespace "illusions", but that are intended to work with ``-XNoImplicitBinds`` in which case they do succeed in their intended purpose.
```
int-index, line 138: Hm, good point that `-Wpuns` and `-Wpun-bindings` are insufficient to maintain the illusion of a single namespace. I have not thought about this. But saying that `-XNoImplicitBinds` is required would be a step too far. I believe we just need a third warning: `-Wterm-variable-capture`. Good news is that it is already part of an accepted proposal (namely [#281](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0281-visible-forall.rst)) and there's an implementation in [!9420](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/9420).


Your example

```
a = Int
f (x :: a) = ...
```

would trigger `-Wterm-variable-capture`, whereas something like

```
f (x :: b) = ...   -- no term-level ‘b’ in scope
```

would continue to work under `-Wterm-variable-capture` (but not under `-XNoImplicitBinds`).

principles.rst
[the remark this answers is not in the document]
Ericson2314: Let me just recap your position as I understand it:

- Substantive changes seem mostly fine
   - substantive changes include `-XImplicitBinds` enabling pattern signature binds.
- Characterizing pattern signature binds as implicit binds seems bad or at least unnecessarily opinionated.
- Principles changes seem gratuitous
  - Principles changes are trying to demonstrate how pattern signature binds are dubious in the same way as the others.

So if you really want, I will relent and revert the principles changes in the interest of getting something done, but based on the above I hope it is clear why the amended proposal with the original principles feels shaky to me, and I worry how the committee would react.

If pattern signatures binds really are a perfectly good explicit binding, then it doesn't make sense that `-XImplicitBinds` would be required to use them! On the other hand if they are a dubvious implicit/contextual binding, then it does make sense that `-XImplictlBinds`/`-XContextualBinds` is a requirement to use them.

Ultimately, I don't really care what the names are, consider my proposal here subject to alpha equivalence. (That is why I bring up "contextual" vs "implicit".) I just want the principles and design decisions to live in harmony.

Also, I suspect you view might view the principles as a sort of append-only log, whereas I view the principles as a "living document" in which shuffling them around is a perfectly normal thing to do whether one is amending an existing proposal or writing a new proposal from scratch.

proposals/0532-type-variable-scoping.rst
Ericson2314: I think `-Wpun-bindings` would imply `-Wterm-variable-capture`? But sure `-XNoImplicitBinds` doesn't need to be formally required.

Basically the "issue" is that before there was term variable namespace lookup in types, `-XImplicitBinds` was purely monotonic in allowing more programs, but now it is not monotonic in that some valid programs will have their meaning changed. That need not be a show-stopper however, after all, extensions like `-XScopedTypeVariables` also change the meaning of valid programs and are thus not monotonic.

proposals/0448-type-variable-scoping.rst
```diff
@@ -331,8 +446,12 @@ Effects
...
 ~~~~~~~~~~~~
 
-1. Previous versions of this proposal, along with the accepted `#285`_, use ``-XNoPatternSignatureBinds`` instead of ``-Wpattern-signature-binds``.
-   However, there seems to be no good reason this must be an extension, instead of a warning.
+1. We could further break down ``-XImplicitBinds``, like before.
+
+   But fixing the drafting error would require a *third* extension, ``-XNegativeSignatureBinds``, in addition to the original two.
+   This would allow more conservative defaults --- we must have Haskell98 implicit foralls but not the others which are all guarded behind language extensions today.
+
+   However, @Ericson2314 sensed there is a weariness with too many extensions coming from this, and so didn't do it.
```
Ericson2314, line 454:
```suggestion
1. We could further break down ``-XImplicitBinds``, like before.

   This would allow more conservative defaults --- we must have Haskell98 implicit foralls but not the others which are all guarded behind language extensions today.
   
   But fixing the drafting error in `#285`_, where certain examples meant to be including under `-XPatternSignatureBinds` that wouldn't be because they are not "pattern signatures", in this manner would require a *third* extension, ``-XNegativeSignatureBinds``, in addition to the original two.
   @Ericson2314 sensed there is a weariness with so many extensions coming from this proposal (448) already, and so didn't want to pursue this route.
```

proposals/0448-type-variable-scoping.rst
Ericson2314: Ah I now gather that "But fixing the drafting error" is what is confusing. See https://github.com/ghc-proposals/ghc-proposals/pull/532/files#r1038961339 where I tried to reword this.

proposals/0532-type-variable-scoping.rst
int-index: Yes, I agree with your characterization of the issue around `-XImplicitBinds`, but you need not bring `-Wpuns` and `-Wpun-bindings` into this. Perhaps say that `-XImplicitBinds` may break programs that assume a single namespace, but `-Wterm-variable-capture` is available to detect such programs.

Ericson2314: https://github.com/Ericson2314/ghc-proposals/blob/type-variables/proposals/0448-type-variable-scoping.rst#unified-namespacing-and-extension-monotonicity has this!

proposals/0448-type-variable-scoping.rst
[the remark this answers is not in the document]
Ericson2314: I expanded upon it in https://github.com/Ericson2314/ghc-proposals/blob/type-variables/proposals/0448-type-variable-scoping.rst#unified-namespacing-and-extension-monotonicity

Ericson2314:
@aspiwack further thoughts? I think https://github.com/ghc-proposals/ghc-proposals/pull/532#discussion_r1038959470 is a good summary comment from my end.

proposals/0448-type-variable-scoping.rst
```diff
@@ -331,8 +449,43 @@ Effects
...
+But on the other hand, `270#`_ and `281#_`, the latter of which is accepted and partially implemented, adopt a model where variables in types resolving to variables defined in the term namespace as a fallback unconditionally.
+This is indeed backwards compatible, however it breaks the property of ``-XImplicitBinds`` being strictly *non-forklike* in allowing only more programs, not changing the meaning of existing programs.
+
+To wit, if ::
+
+  t = Int
+  x :: t -- out of scope, no type variable `t` in scope.
+  x = 0
+
+is an invalid program, we can *either* give make it valid by saying the second ``t`` is a use or implicit bind, but we cannot do *both*.
```
aspiwack, line 475: I assume you meant
```suggestion
is an invalid program, we can *either* make it valid by saying the second ``t`` is a use or implicit bind, but we cannot do *both*.
```

principles.rst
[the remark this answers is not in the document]
aspiwack: > Substantive changes seem mostly fine
> 
>     * substantive changes include `-XImplicitBinds` enabling pattern signature binds.

My personal opinion on whether I agree with the change is insubstantial, honestly. This is why we have a committee. My first job as a shepherd is make sure that the proposal is clear and consistent. It is.

> Characterizing pattern signature binds as implicit binds seems bad or at least unnecessarily opinionated.

I'd even say that it's factually incorrect. It doesn't mean that they're not a bad idea though.

> Principles changes seem gratuitous
> 
>     * Principles changes are trying to demonstrate how pattern signature binds are dubious in the same way as the others.

My point is that the principles already did this. But you can stick to your guns, let's see what the rest of the committee says.

I'll be submitting for committee discussion now: I really should have done this last week as I'll be on holiday for the 3 next weeks and won't be checking my emails. The committee's decision will most likely be in January. I'm really sorry about this.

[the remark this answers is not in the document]
Ericson2314: OK sounds good. Basically I don't want to be so stubborn that the principles changes sink the rest of the proposal.

I take it you are saying the Lexical Scoping Principle covers this case today (I agree actually), so then should the extension be called `-XNonLexicalBinds` or `-XLexicalBinds`? I'll go with the flow on whatever the committee wants re naming.

Enjoy your vacation!

aspiwack:
In view of this new remark, should I hold out my recommendation to the committee?

Ericson2314:
@aspiwack Nah I think it's find to send it. The bottom line is that more eyeballs is good.

-----

I chatted with @int-index earlier today bunch who pointed out that "implicit `forall`s" and "implicit binds in kind signatures" are in fact overlapping notions. That is an issue with what I wrote above but *not* with the proposal as written.

To resolve, I think it would be useful to basically think about what we want in terms of these two questions,

1. *Where are we trying to go?* Of which I hope the answer is something like "All post-Haskell-98 forms of implicit/contextual binding are deprecated, all implicit/contextual binding (including that from Haskell-98) one is able to disable."

2. *How do we want to get there? Of which are few routes and at least @int-index and I are not sure of the best answer.

Since these are more open-ended questions, it might be worth arranging a real-time discussion to answer them. Then I could update the proposal with whether the consensus of the meeting is. (I bring this up not circumnavigate the usual proposal process, but to complement it.)

aspiwack:
I've sent my recommendation.

Ericson2314:
Thank you so much, @aspiwack!

simonpj:
> Motivation: The [Explicit Binding Principle](https://github.com/Ericson2314/ghc-proposals/blob/type-variables/principles.rst#explicit-binding-principle) allows programmers to control exactly how variables come into scope. It ensures all short-hands can be explained in terms of an explicit, unambiguous equivalent that is easier to understand at the cost of being more verbose:

I found the bullets that followed this sentence to be quite confusing.  First use of "postive" and "negative.   

Moreover, this "motivation" sentence could move up, before "Examples", so that the examples simply illustrate the motivation.

--------------

In [the rich diff](https://github.com/ghc-proposals/ghc-proposals/pull/532/files?short_path=ddeba30#diff-ddeba30251d5cfc0e98c4f2759b8543b4ad6a806294dac10907930e1e843bdb2) I see


>     For every appearance of an identifier, it is possible to determine whether that appearance is a mere occurrence, and thus must be bound elsewhere for the program to be valid, or the variable is a binding site (or causes an implicit binding, which is close enough), without examining the context.
>    For every occurrence of an identifier, it is possible to uniquely identify its binding site, without involving the type system.

But I don't see that in the [rendered form](https://github.com/Ericson2314/ghc-proposals/blob/type-variables/principles.rst).  I like the extra clarification

simonpj:
John, I'm struggling to review this. You advise looking at the rich diff, but it's pretty long, and has a number of reorderings, examples, and other presentational improvments that make it hard to spot acutal changes.

Could you see your way to listing the actual changes you propose?  Other than presentational improvements, which are great.

E.g I think you are making no change whatsoever to the priniples, just improving presentation. Right? Or wrong?  Similarly the rest. 

Sorry to erect yet another obstacle, but I don't otherwise know how to assess this proposal.

Ericson2314:
@simonpj Just to check, did you see the [rendered meta-proposal](https://github.com/Ericson2314/ghc-proposals/blob/type-variables/proposals/0532-type-variable-scoping.rst), that summarizes and motivates all the changes made to the other existing proposals, and is meant to compliment reading the diff of the rest. I had a link to it above but not next to the other "rendered" links so I put it there too so it is more obvious.

simonpj:
> Just to check, did you see the [rendered meta-proposal](https://github.com/Ericson2314/ghc-proposals/blob/type-variables/proposals/0532-type-variable-scoping.rst), that summarizes and motivates all the changes made to the other existing proposals,

Thanks for pointing to it again.   I'm still finding it hard going.  
* It has a single section called "Motivation" and no list of "Proposed changes".
* All the links to "#448" are broken
* I think that section 1.1 relates to a particular accepted proposal, but there is no link.
* It's quite discursive, describing your thought processes.  But what I seek is simply a *list of changes*.  Following that list some motivating reasons is great, but the changes themselves are what I am after.   (You could say that the changes are in the diffs of the individual proposals, but many of those changes are presentational -- which is good -- so I'm finding it hard to know precisely what non-presentational changes you want to make.)

For example, I think 1.1 is about the Principles accepted propsals.  I think your changes are:
* **Combine two principles into one** (I think)
plus some re-ordering to improve presentation.

In 1.2, the link in the heading is again broken.  I think you want [this link](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0425-decl-invis-binders.rst).  I think your changes here are purely presentational -- no change in specification, right?

In 1.3 (link broken again) I'm not sure what the changes are.  I think maybe `-XImplicitBinds` is new, and replaces some other extensions which now disappear?  Maybe 1.3.2 is a second change.  But 1.3.3 doesn't describe a change at all .. maybe it is explanation for 1.3.2?

etc.  You can see how confused I am.  Sorry to be dim.

Ericson2314:
@simonpj very fair points! I have added a list to the top and fixed the links. Hopefully that helps!

Ericson2314:
Hi committee. I left this off with trying to set up a meeting to discuss some of the nuance in person. I sent some emails, but never actually proposed the meeting (being busy with other things); the fault is entirely mine.

The recent developments of PRs around this stuff --- plus spit-base getting my head back in GHC-land in general --- has thinking about that. Perhaps I should finally set up that meeting, or perhaps it would make sense to chit-chat about this stuff at Zurihac, or maybe both.

aspiwack:
I, personally, won't be there at ZuriHac, so no IRL chit-chatting for me I'm afraid. We'll figure some other way.

aspiwack:
(I've marked the proposal as “Needs revision” to signal that the next action in on John and not me, for internal bookkeeping in the community. The real status is: some discussion about scope is pending; but I say it's a good enough approximation)

Ericson2314:
Based on conversation with @simonpj then @int-index, I am going to put this on pause instead propose a series of tiny changes to #448 we can hopefully digest much more easily.

The first such change is https://github.com/ghc-proposals/ghc-proposals/pull/604

proposals/0448-type-variable-scoping.rst
```diff
@@ -552,8 +552,14 @@ Additionally, this choice edges us closer to the `Lexical Scoping Principle`_,
 because we no longer have to check whether ``a`` is in scope before identifying the ``a`` in ``f (Just @a x) = ...`` is a binding site or an occurrence.
 
 The other change in this restatement is the use of new extension ``-XTypeAbstractions`` instead of the current status of piggy-backing on the combination of ``-XTypeApplications`` and ``-XScopedTypeVariables`` (*both* need to be enabled today).
-This proposal suggests instead that ``-XScopedTypeVariables`` implies ``-XTypeAbstractions`` so that we remain backward-compatible with what is current implemented
-(though there may be some redundant enablings of ``-XTypeApplications`` that would no longer be needed).
+This proposal suggests instead that ``-XScopedTypeVariables`` and ``-XScopedTypeVariables`` should enable this feature in a deprecated manner, to encourage users to use ``-XTypeAbstractions`` instead.
```
nomeata, line 555:
```suggestion
This proposal suggests instead that ``-XScopedTypeVariables`` and ``-XTypeApppications`` should enable this feature in a deprecated manner, to encourage users to use ``-XTypeAbstractions`` instead.
```

Ericson2314: @nomeata Sorry for the noise, the other PR will take care of that. This one is on hold until it and one I have yet to open are resolved -- trying to avoid huge diffs and get the least controversial parts taken care of first.

I just pushed to this one to make sure I wasn't accumulating conflicts.

nomeata: This is a different PR? Ah, right! Ignore me then :-)

PR #533, opened 2022-08-24
All Threads Syscall

NicolasT:
Several Linux kernel security features exposed as syscalls are applied to the thread invoking such syscall only, then inherited by newly created threads. Since the GHC threaded RTS creates multiple threads before Haskell code is executed, we need a way to invoke some syscall in all running threads.


proposals/0000-all-threads-syscall.rst
```diff
@@ -0,0 +1,146 @@
...
+- First, the syscall is invoked in the calling thread. Keep track of the
+  result.
+
+- Then, set up machinery to invoke the syscall in all (other) running threads,
+  and wait for them to accomplish this task, blocking the calling thread.
+
+- If during the invocation of the syscall in any of the other threads the result
+  is *not* equal to the first invocation, abort the process.
+
+- Return the result.
```
hsyl20, line 70: Couldn't we have a more generic interface that doesn't mention syscalls at all? E.g.

```C
// callback for runOnAllThreads
// return false to abort before all threads have run the callback function
typedef bool (*run_callback)(void *user_data);

// run the given callback on all threads sequentially until one of them returns false
// or all of them have executed the callback.
void rts_runOnAllThreads(run_callback cb, void * user_data);
```

The business of keeping the result etc. would be done in the callback using the given `user_data` as storage.
By doing this, platform specific syscall calling could be done in a library instead of in the RTS.

NicolasT: Thanks for the feedback! Let me break this up in two parts:

- Running the action (whichever it may be, only syscalls or generic callbacks) in every thread sequentially should work, however, I'm not sure why that'd be desirable (vs. running them concurrently): once it turns out one of them returns "something else" (or `false`), the only way forward is `abort()` since the process would be in a very weird state. Hence, running them sequentially would only cause latency without other benefits?
- On allowing a generic callback to be passed in, this may seem enticing, indeed. However, with great power comes great responsibility. Depending on the implementation of the feature inside the RTS, one is very limited as to what one is allowed to do within such callback. In the design of the Glibc or Go runtime implementation of a similar feature, a signal and signal handler is used to interrupt running threads to perform the syscall. As such, the given callback would run inside a signal handler, and hence only async-signal-safe functions may be used (in the callback), a limitation the RTS itself can't enforce.

To conclude: I wouldn't oppose against using a more generic approach, if we deem the (non-enforcable) limitations of what can be done in the callback acceptable (I fail to see any other use-cases but triggering a syscall, though). As for running sequentially across threads, I'm not sure what the benefit would be.

hsyl20: Indeed we could run the callback in parallel. We should document this properly as users would then have to use atomic/locks to deal with their `user_data`. Not sure why I proposed to do it sequentially.

I don't think we would use a signal to implement this. We already have a thread synchronization mechanism (e.g. for major GC) so we could reuse it. We would have to handke threads not used for running capabilities, such as the ticker. Not sure about FFI threads: we could interrupt the interruptible ones and fail if there are non-interruptible ones running. In any case, we would have to document what's possible to do in the callback (blocking, etc.).
 
The generic approach seems better to me but you could mention both in the proposal as alternative implementations. Note that the generic approach has an interface similar to `rts_listThreads` which has been added in 9.6.

Ericson2314: > By doing this, platform specific syscall calling could be done in a library instead of in the RTS.

This alone is a huge benefit, but baking in less OS-specific stuff when an abstract interface will do is also a bonus to me.

nomeata:
@NicolasT, are you by any chance attending Haskell Symposium? If so, would you like to present this proposal at the [GHC Proposals session](https://icfp22.sigplan.org/home/haskellsymp-2022#GHC-Proposals-Session)?

NicolasT:
> @NicolasT, are you by any chance attending Haskell Symposium? If so, would you like to present this proposal at the [GHC Proposals session](https://icfp22.sigplan.org/home/haskellsymp-2022#GHC-Proposals-Session)?

I wish :smiley: However, no, I won't be there (been 10 years since my last ICFP).

proposals/0000-all-threads-syscall.rst
NicolasT: Sorry folks, this didn't drop off my radar. I'll integrate the comments shortly.

PR #535, opened 2022-09-02
Partial Field Behavior

parsonsmatt:


simonpj:
I'm fine with `NoSumSelectors`, but very strongly against `NoSumRecords`.  Data types with named field are extremely useful for pattern matching and construction -- and in those domains it would be silly to discourage (in any way) the use of named fields.

But providing a way to avoid partial functions, yes, by all means.

adamgundry:
Thanks for bringing forward this proposal! It would be nice to have a better story around partial fields.

Personally I would most like to see some version of the `MaybeFieldSelectors` route. I don't see much point in adding `NoSumRecords` to GHC (since it doesn't add any functionality, it could just as well be checked by a lint tool or a warning, and we already have `-Werror=partial-fields`). And `-Werror=incomplete-record-selectors` with #516 would give much of the effect of `NoPartialFieldSelectors` (albeit with a slightly different definition site vs. use site trade-off). Similarly `NoPartialRecordUpdate` is basically `-Werror=incomplete-record-updates`.

Having `MaybeFieldSelectors` change the type of selector functions (and hence `HasField` instances) defined in the current module seems relatively easy. I think the interaction with `-Werror=incomplete-record-selectors` is no problem: you don't get a warning if you use a selector defined under `MaybeFieldSelectors`, because it isn't incomplete! It might be a bit surprising to users that the types of selector functions change, and it isn't a conservative extension, but that seems unlikely to be a big deal.

The challenge is, indeed, updates. I think a plausible way to implement updates for partial fields would be to have them return the value unchanged if the field is not present, which is what one wants to define an affine traversal.

However, to support `OverloadedRecordUpdate` and to generate optics based on field names, we want to add something like
```hs
class SetField x s b where
  setField :: b -> s -> s
```
so we would need to generate something like
```hs
instance SetField "eventId" Message EventId where
  setField e = \case
    s@SendChat{} -> s
    Signup{user} -> Signup {eventId=e, user}
```
but now the field type is inconsistent with `HasField "eventId" Message (Maybe EventId)`. Maybe that's okay? But it seems a bit odd, and it means that a constraint like `(HasField x s b, SetField x s b)` will rule out partial fields entirely.

(Actually the accepted proposal #158 currently puts both getting and setting together in a single class, which won't work at all if we change the type of the "get" result like this, but splitting up the classes solves that problem amongst others.)

parsonsmatt:
My understanding is that #158 is Under Redesign in some way, and wouldn't be implemented as-is. Though the last time I went digging for the state of that proposal, I had to chase a bunch of GitLab issues and PR threads. 

Hm, given `(HasField x s (Maybe b), SetField x s b)`, we can naturally write:

```haskell
updateField :: (HasField x s t, SetField x s r) => (t -> r) -> s -> s
updateField k s = setField @x (k (getField @x s)) s
```

Which would work alright - for a partial field, it'd become `(Maybe typ -> typ) -> rec -> rec`, which even gives you the "am i modifying the right constructor" information when performing the partial update. 

I can see it being very nice and also very annoying if a partial update failed with no info.

AntC2:
> The current solution is to ... use a specific record type instead of a record.

Is it? The latest solution (as of 9.2) would be `NoFieldSelectors`. None nowhere. I've never liked field selectors. It was first proposed to do away with them around 2014. [And they play dubiously with field puns](https://gitlab.haskell.org/ghc/ghc/-/issues/22140) -- since folks seem to be [getting rid of punning](https://github.com/ghc-proposals/ghc-proposals/pull/270).

Always use pattern matching.

parsonsmatt:
> The latest solution (as of 9.2) would be `NoFieldSelectors`.

This does not forbid HasField access, so partiality remains.

adamgundry:
> My understanding is that #158 is Under Redesign in some way, and wouldn't be implemented as-is. Though the last time I went digging for the state of that proposal, I had to chase a bunch of GitLab issues and PR threads.

Yes. I've been trying to keep this wiki page up to date with the current state of things: https://gitlab.haskell.org/ghc/ghc/-/wikis/records/overloaded-record-fields


> Hm, given `(HasField x s (Maybe b), SetField x s b)`, we can naturally write:
> 
> ```haskell
> updateField :: (HasField x s t, SetField x s r) => (t -> r) -> s -> s
> updateField k s = setField @x (k (getField @x s)) s
> ```
> 
> Which would work alright - for a partial field, it'd become `(Maybe typ -> typ) -> rec -> rec`, which even gives you the "am i modifying the right constructor" information when performing the partial update.

This isn't quite the type you want: it requires the caller to provide a function that computes a `typ` even in the `Nothing` case, but in that case there's no value being updated and hence the result will never be used.

I suppose another strategy would be to provide
```hs
class SetField x s b where
  updateField :: (b -> b) -> s -> s
````
and hence get the right update behaviour without using `HasField` at all. That might make the strange interaction between `HasField` and `SetField` under `MaybeFieldSelectors` bite slightly less often, but I don't have a very good sense of whether it would be a problem in practice.

> I can see it being very nice and also very annoying if a partial update failed with no info.

Right. It does risk concealing bugs in some cases that would previously have thrown runtime errors due to incomplete updates and hence been more visible. So I suppose there's something to be said for `MaybeFieldSelectors` leading to partial fields being read-only, i.e. preventing their use in updates and `SetField`, and that would avoid the issues of `HasField` and `SetField` giving different types to the field.

**Comment by AntC2** [comment]
> [`noFieldSelectors`] does not forbid `HasField` access

I of course meant don't use `HasField` either/don't import `GHC.Records`. I said

>> Always use pattern matching.

I see the discussion on the Committee mailing list wrt pun-free code. Having a field selector (`:: T -> Int`) with the same name in the same namespace as a field label (`:: Int`) seems to me a _more_ egregious form of punning. And it's what prevented having same-named fields in different data types. (The field selector would be `:: T2 -> Char`.)

> [How do other languages handle this?](https://github.com/parsonsmatt/ghc-proposals/blob/matt/partial-field-behavior/proposals/0000-partial-field-behavior.md#other-languages)

I think @parsonsmatt's comparison could usefully include Hugs/Trex -- very similar to Purescript.

**Comment by parsonsmatt** [comment]
I've updated the proposal and I believe it is ready for review

**Comment by nomeata** [comment]
Meaning ready for review by the broader committee, or is this the “formal” submission to the committee?
In the former case I suggest, if you think it should see more attention, to beat the drum a bit (mailinglist, Discord, Mastodon, Twitter…).
In the latter case I suggest to always [explicitly ping me](https://github.com/ghc-proposals/ghc-proposals#how-to-bring-a-proposal-before-the-committee), as otherwise I might miss it.

**Comment by parsonsmatt** [comment]
Linked on [Twitter](https://twitter.com/mattoflambda/status/1605276734617800710) and [Reddit](https://www.reddit.com/r/haskell/comments/zqwj7w/rfc_ghc_proposal_to_make_partial_fields_safe/)

**Comment by Profpatsch** [comment]
Pinging @aspiwack as I remember he had some strong opinions on Partial Fields.

**Comment by Profpatsch** [comment]
Full +2 on `NoPartialFieldSelectors`, I’d love to enable that everywhere.

Not sure about the `MaybeFieldSelectors` part, I feel like it would be very confusing to suddenly see a `Maybe` introduced into the types without further explanation. Maybe it should be a `PartialField` type that is easy to convert to a `Maybe` on demand? Or maybe there is a way to attach this information to the type and show it when I hover over the accessor?

**Comment by cbarrett** [comment]
Seems good to me. Just a drive-by suggestion for further refinement, feel free to take it or leave it. 

In the same way that we can get a compile time error with a message when using the `HasField` constraint, it might provide a better developer experience and overall user experience to also provide a `partialField :: Record -> Either String a` function when `MaybeFieldSelectors` is on (or put this behind yet another extension, no strong opinion on that).

**Comment by adamgundry** [comment]
I suggest that before moving to a formal committee review we should try to gather consensus on which approach to take. While there is in principle nothing wrong with having both `NoPartialFieldSelectors` and `MaybeFieldSelectors`, the current tendency within the committee is to avoid multiple incompatible extensions. (I'm more relaxed about this personally, and indeed the current records story is replete with such choices...)

For `NoPartialFieldSelectors`, I'm interested in whether people consider `-Werror=incomplete-record-selectors -Werror=incomplete-record-updates` an adequate replacement (essentially: should this be a definition-site or use-site decision, or both)?

For `MaybeFieldSelectors`, the key open question seems to be the treatment of updates. Leaving them as partial seems like a missed opportunity (although there is always `-Werror=incomplete-record-updates`). But none of the alternatives are entirely convincing either.

**Comment by parsonsmatt** [comment]
> For NoPartialFieldSelectors, I'm interested in whether people consider -Werror=incomplete-record-selectors -Werror=incomplete-record-updates an adequate replacement (essentially: should this be a definition-site or use-site decision, or both)?

Do those warnings trigger with `HasField` use?

I think I'd *like* to have the option to disable field generation for things that are always an error, if only to free up the namespace. With `NoPartialFieldSelectors`, I'm free to handle how I want:

```haskell
{-# language NoPartialFieldSelectors #-}

data T = C { x :: Int } | D

x :: T -> Maybe Int
x = \case
    C { x } ->  Just x
    _ -> Nothing

instance HasField "x" T (Maybe Int) where
    getField _ = x
```

If we have #512, then `NoPartialFieldSelectors` becomes "syntax (extension?) sugar" for `{-# NoFieldSelector #-}` on partial fields.

> For MaybeFieldSelectors, the key open question seems to be the treatment of updates. Leaving them as partial seems like a missed opportunity (although there is always -Werror=incomplete-record-updates). But none of the alternatives are entirely convincing either.

Yeah ~ I feel like the safety aspect is the biggest problem. Until I can do `-Werror=incomplete-record-updates` (and some form of `NoPartialFieldSelectors`), I don't want to ever see a partial field, even though they *are* fine in many cases.

I feel like the options for a partial field update are no-op or runtime error. With no-op, you actually get a nice ability to do an update that is "constructor agnostic" - ie, suppose:

```haskell
data T 
    = A { a :: Char, x :: Int } 
    | B { b :: Double, x :: Int }
    | C { x :: Int }

change ::T -> T
change t = t { x = x.t + 1, a = maybe 'a' succ t.a, b = 0.0 }

-- >>> change (A 'x' 0)
-- A { a = 'y', x = 1 }
-- >>> change (B 3.0 1)
-- B { b = 0.0, x = 2 }
-- >>> change (C 3)
-- C { x = 4 }
```

The awkwardness *does* show up with `a = maybe 'a' succ t.a`. 

But I do feel fine punting that to a later extension - if this is a warranted impovement for what it covers, future work can always improve what remains.

adamgundry:
> > For NoPartialFieldSelectors, I'm interested in whether people consider -Werror=incomplete-record-selectors -Werror=incomplete-record-updates an adequate replacement (essentially: should this be a definition-site or use-site decision, or both)?
> 
> Do those warnings trigger with `HasField` use?

For `incomplete-record-selectors` they are specified to do so by #516, while for `incomplete-record-updates` doing so makes sense but will need to be part of a proposal to split `HasField` into separate get/set classes.

That said, you make good points about `NoPartialFieldSelectors` being useful to free up the namespace. That's quite a convincing reason to prefer it to the warnings.

The analogy with `NoFieldSelectors` isn't exact, because `NoFieldSelectors` releases the namespace but still allows `HasField` and updates. Maybe the extension should be called `NoPartialFields` since it covers all of the ways in which fields introduce partiality (selectors, updates and `HasField`)?

I'm coming round to the idea that we should just do that, and leave out the awkwardness of partial field updates with `MaybeFieldSelectors`. Saying "you can define partial fields, but they are not in scope as selector functions and you cannot use them in record updates" is simple, free from footguns, and mostly what I want when I define partial fields in practice.

michaelpj:
`NoPartialFieldSelectors` is great, big +1 form me.

`MaybeFieldSelectors` seems odd to me.  I would be fine with simply not being able to do record updates. `Control.Lens.TH` is there if I want to get a sensible `Traversal` for that case.

parsonsmatt:
I've amended the proposal such that `MaybeFieldSelectors` does not alter the behavior of record update.

m4dc4p:
I really like `NoPartialFieldSelectors`. I don't know that I'd ever use `MaybeFieldSelectors`.

I htink you missed a bit of text about record updating - at [https://github.com/ghc-proposals/ghc-proposals/pull/535 you ](https://github.com/ghc-proposals/ghc-proposals/pull/535/files#diff-85eb54dba7f08d3efda7caa11ea32d2189756e696fa8ab5e81732a0bb004006eR139) you say there is no record updating (unsafe), but later you say there is (https://github.com/ghc-proposals/ghc-proposals/pull/535/files#diff-85eb54dba7f08d3efda7caa11ea32d2189756e696fa8ab5e81732a0bb004006eR248).

Great proposal!

blamario:
+1 for `NoPartialFieldSelectors`.

I fear `MaybeFieldSelectors` might be dangerous in some situations. Consider adding a second constructor to an existing record declaration: with `NoPartialFieldSelectors` all existing field selectors will become partial and thus invalid and the compiler will force you to remove them; with `MaybeFieldSelectors` some of them may silently change the run-time logic with no compiler warning.

Ericson2314:
I agree with that others that `NoPartialFieldSelectors` is great, and `MaybeFieldSelectors` is fine but less urgent.

I also agree `-XNo` is better than `-Werror=` for cleaning up the design space. Ultimately I see records right now as going through a process of backing out of previous dubious design decisions before eventually landing on others ones, and `-XNo` is an integral part of that process.

vagarenko:
Note that GADTs record fields can be safe
```haskell
Prelude> :set -XDataKinds -XKindSignatures -XGADTs
Prelude> data Tag = A | B
Prelude> :{
Prelude| data Foo (a :: Tag) where
Prelude|   FooA :: { fooa :: () } -> Foo A
Prelude|   FooB :: { foob :: Int } -> Foo B
Prelude|   FooC :: { fooc :: Bool } -> Foo a
Prelude| :}
Prelude> :t fooa
fooa :: Foo 'A -> ()
Prelude> :t foob
foob :: Foo 'B -> Int
Prelude> :t fooc
fooc :: Foo a -> Bool
```
only `fooc` is partial here.

But if I add another constructor `FooD :: { food :: Char } -> Foo A` both `fooa` and `food` become partial.

adamgundry:
@vagarenko your example needs a slight modification, because `FooC` can be used at type `Foo A` or `Foo B` so `fooa` and `foob` are actually both partial. But the more general point is a good one. The proposal should specify what happens in the case of a GADT. I suppose `NoPartialFieldSelectors` could have some analysis to decide whether a GADT selector is known to be total (something like: every constructor either has the field, or has a return type that is apart from the return type of the constructors with the field). But I wonder if the complexity is justified?

proposals/0000-partial-field-behavior.md
```diff
@@ -0,0 +1,362 @@
...
+[TypeScript](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html) provides a compiler error if you use a partial field on a union type.
+However, if you refine the type using a `switch` statement, then you can use the fields, and this is guaranteed to be safe.
+Additionally, the compiler does not complain if you use a field that is present on every type.
+
+[Rust](https://doc.rust-lang.org/reference/items/enumerations.html) simply does not allow you to use dot notation on an enum with record fields, even if the field is shared among all types.
+To access named fields on an enum, you must use pattern matching.
+
+[F#](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/discriminated-unions) also does not treat named fields on a union as accessors, and only uses them for construction and pattern matching.
+
+[OCaml](https://stackoverflow.com/questions/22367396/does-ocaml-have-record-syntax-in-disjoint-unions) does not allow you to define named fields in a union at all.
```
aspiwack, line 97: Actually, this is no longer true (since ~2016): you can use labelled fields in sums in Ocaml (and you've been able to since about the time of this stack overflow thread, which is rather funny).

The feature goes by the name of [“inline records”](https://v2.ocaml.org/releases/5.0/htmlman/inlinerecords.html) and is pretty nicely designed in my opinion.

aspiwack (commented):
I've just skimmed the proposal for now, just a small remark on the state of the art.

parsonsmatt:
@nomeata Conversation has died down on this a bit - can we move it forward?

nomeata:
I'm currently traveling, so please allow a week delay (and please ping again if I somehow lose it)

AntC2:
@parsonsmatt, in the week before nomeata gets back to this, there's a few unfilled sections in the proposal. I suspect the Committee will ask you to expand.

In particular the new policy is to not proceed with proposals unless there's[ somebody(s) identified as willing to implement it](https://github.com/parsonsmatt/ghc-proposals/blob/matt/partial-field-behavior/proposals/0000-partial-field-behavior.md#implementation-plan); [particularly if it's a large amount of work](https://github.com/parsonsmatt/ghc-proposals/blob/matt/partial-field-behavior/proposals/0000-partial-field-behavior.md#costs-and-drawbacks).

JakobBruenker:
> In particular the new policy is to not proceed with proposals

Does it really? The [proposal](In particular the new policy is to not proceed with proposals) ends with a comment saying "We decided against this"

proposals/0000-partial-field-behavior.md
AntC2: Thanks @aspiwack ooh yes I like!

Especially `Error: This form is not allowed as the type of the inlined record could escape.` So that's essentially the same as pattern-matching on GADTs: you can only use the contents of the constructor (that is, the binding names) within scope of the match.

I see also you can match to the whole record (the `p` in `| Point p -> ...`). Does that mean records are first-class/stand-alone as in Javascript and purescript?

AntC2:
> Is it really?

See my comment immediately before. I'd say yes it is "really"/de facto. Even though not de jure.

adamgundry:
My understanding of the committee consensus is that an implementation plan is explicitly *not* required, and indeed there have been numerous proposals accepted without one (e.g. my #581). I don't think that's an obstacle in this case. (Of course that does mean that a proposal being accepted has no guarantee of being promptly implemented, which is unfortunate, but a reflection of the resources we have available.)

I think we will need to make a choice between adding `NoPartialFieldSelectors` or `MaybeFieldSelectors`, rather than adding both. But if I understand the first paragraph of the Proposed Change Specification correctly, @parsonsmatt is happy for the committee to make that choice as part of the deliberation process?

My sense is that `NoPartialFieldSelectors` is the way forward here, for a few reasons:
 * `NoPartialFieldSelectors` has a nice simple story about record update: you can't.
 * Enabling `NoPartialFieldSelectors` does not change the behaviour of existing programs that continue to compile, it just rules out some programs (whereas `MaybeFieldSelectors` might silently change types and hence behaviour).
 * To understand code using a selector function, the user would have to know whether `MaybeFieldSelectors` was enabled at the definition site and whether the field was partial.
 * As pointed out in https://github.com/ghc-proposals/ghc-proposals/pull/535#issuecomment-1364316803, under `MaybeFieldSelectors`, adding a constructor to an existing type might make a field partial and hence silently change behaviour elsewhere, which is rather troubling.

The `NoPartialFieldSelectors` proposal could do with a bit more clarity on a few points:
 * The interaction with `NoFieldSelectors`. I can see multiple reasonable options here:
    - `NoPartialFieldSelectors` also has the effect of enabling `NoFieldSelectors` for partial fields (only), so the user can define a different value using the name (e.g. a function that returns a `Maybe`).
    - `NoPartialFieldSelectors` is merely negative, preventing uses of partial fields but not freeing up the namespace. The user can separately enable `NoFieldSelectors` (or not) as they choose. This is a bit more orthogonal, if perhaps less convenient.
    - Maybe there is a case for two extensions, `NoPartialFields` (which is merely negative) and `NoPartialFieldSelectors` which enables `NoPartialFields` *and* selectively uses `NoFieldSelectors` for the partial fields?
 * Under `NoPartialFieldSelectors`, the proposal suggests that reusing the name of a partial field with a virtual `HasField` instance is forbidden. But I think we could allow this, and hence users could add back an instance using `Maybe` if they so choose?
 * At the risk of bikeshedding too early, I still think the name `NoPartialFields` is better as the change applies to more than just selectors. Though `NoPartialFieldSelectors` might make sense if there is a clearly defined interaction with `NoFieldSelectors` that warrants it.
 * It would be nice to see a more developed argument in the proposal text that `-Werror=incomplete-record-selectors -Werror=incomplete-record-updates` is not a sufficient alternative to `NoPartialFieldSelectors`. (Two plausible reasons came up in the discussion above: freeing up the namespace, and that we should let library authors make this choice at definition sites rather than requiring clients to set flags at use sites.)

nomeata:
I see that Adam already did a first review of the proposal, with some questions. I suggest that you resolve these before we send it before the full committee. Please ping me again when that has happened.

proposals/0000-partial-field-behavior.md
aspiwack: It depends on what you mean by “first-class” I suppose. In Ocaml, there are record types (which don't have constructors):

```ocaml
type foo = { bar: int; baz: bool }
```

These record types, however, are nominal: this is not a “row type” as in Purescript.

(Ocaml's object can serve as extensible records, but they won't have a pattern-matching syntax for instance.)

georgefst:
Having belatedly realised, after using it for a year, that `OverloadedRecordDot` inherits the partiality problems of field selectors, I'm strongly in favour of this.

One thing that I haven't seen mentioned so far is that currently GHC complains about two fields being defined with the same name and different types across different constructors in the same type, e.g. `data D = D1 {f :: ()} | D2 {f :: Bool}`. The only reason I can see for this restriction is that it makes it impossible to give an output type to the field selector `f` (or similarly, to choose `a` in `HasField "f" D a`). So with this extension enabled, we could, and I believe should, lift the restriction.

(I do have a non-contrived example where this would make things a lot more pleasant, where I currently have fields like `ta :: T a`/`tb :: T b`/`tc :: T c` in separate constructors, and where `T`'s parameter is just a phantom).

proposals/0000-partial-field-behavior.md
```diff
@@ -0,0 +1,362 @@
...
+
+`PartialFieldSelectors` is the current behavior of GHC.
+So let's talk about `NoPartialFieldSelectors`:
+
+#### No selector function
+
+Under `NoPartialFieldSelectors`, we wouldn't get a partial function.
+
+```haskell
+data OptionRec a = None | Some { fromSome :: a }
```
konsumlamm, line 131: Do I understand it correctly that a field present in all variants would not be affected? Perhaps add such a field in the example to clarify this.

PR #537, opened 2022-09-09
Disambiguating record updates using type signatures

adamgundry:
This proposal makes a small adjustment to name resolution for record updates when `DisambiguateRecordFields` is enabled: where multiple types have identically-named fields, expression type signatures can be used to disambiguate which type is meant.

[Rendered](https://github.com/adamgundry/ghc-proposals/blob/disambiguating-record-updates-using-type-signatures/proposals/0000-disambiguating-record-updates-using-type-signatures.rst).

### DIFF
--- a/proposals/0000-disambiguating-record-updates-using-type-signatures.rst
+++ b/proposals/0000-disambiguating-record-updates-using-type-signatures.rst
@@ -1,0 +1,546 @@
+Disambiguating record updates using type signatures
+===================================================
+
+.. author:: Adam Gundry
+.. date-accepted::
+.. ticket-url::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/537>`_.
+.. sectnum::
+.. contents::
+
+This proposal makes a small adjustment to name resolution for record updates
+when ``DisambiguateRecordFields`` is enabled: where multiple types have
+identically-named fields, expression type signatures can be used to disambiguate
+which type is meant.
+
+
+Motivation
+----------
+
+The ``DuplicateRecordFields`` extension makes it possible to define two records
+with the same field name in a single module.  This raises the question of how to
+interpret a record update such as ``r { f = v }`` if there are two fields ``f``
+in scope.  The original design of ``DuplicateRecordFields`` used a limited form
+of type-directed name resolution in this case, however support for this is
+scheduled to be removed following `proposal #366
+<https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0366-no-ambiguous-field-access.rst>`_.
+GHC 9.2 and 9.4 include a warning when such updates are used.  For example, this module: ::
+
+    {-# LANGUAGE DuplicateRecordFields #-}
+
+    data S = MkS { f :: Char }
+    data T = MkT { f :: Int }
+
+    foo :: S -> S
+    foo x = x { f = 'c' }
+
+results in the following warning on GHC 9.2: ::
+
+    M.hs:7:13: warning: [-Wambiguous-fields]
+        The record update x {f = 'c'} with type S is ambiguous.
+        This will not be supported by -XDuplicateRecordFields in future releases of GHC.
+      |
+    9 | foo x = x { f = 'c' }
+      |             ^^^^^^^
+
+The GHC developers would like to remove support for this (mis)feature and turn
+the warning into a fatal error.  However, the question arises as to how such
+updates should be written instead.
+
+One possible answer is provided by the ``OverloadedRecordUpdate`` extension
+(introduced in `proposal #282
+<https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0282-record-dot-syntax.rst>`_
+as amended by `proposal #405
+<https://github.com/ghc-proposals/ghc-proposals/pull/405>`_).  This changes the
+meaning of record update syntax to use typeclass polymorphism, based on the
+``HasField`` typeclass.  Thus under ``OverloadedRecordUpdate`` the program above
+would be accepted without warnings, and in fact the type of the update becomes
+more general, because we have: ::
+
+    foo :: HasField "f" s Char => s -> s
+    foo x = x { f = 'c' }
+
+However, ``OverloadedRecordUpdate`` is not yet fully implemented (pending a
+redesign that is in progress but is not yet complete).  Moreover, there are
+cases that cannot be handled by ``OverloadedRecordUpdate``, including:
+
+* Type-changing update
+
+* Updates to unlifted fields or data types
+
+* Multiple updates, where several fields must be changed simultaneously for the
+  update to be type-correct
+
+* Updates to fields with higher-rank types
+
+Some of these may be addressed by subsequent design changes to the ``HasField``
+class, but others seem fundamentally difficult (in particular, multiple updates
+and higher-rank types).  Thus there are cases where enabling
+``OverloadedRecordUpdate`` is not a conservative extension, i.e. it will break
+existing code.  (Changes to ``OverloadedRecordUpdate`` to improve this situation
+are a topic for future proposals.)
+
+All this suggests that we should have a mechanism for record updates where the
+programmer explicitly specifies which type is being updated.  This gives a
+migration path for existing code that relies on updates which were previously
+accepted by ``DuplicateRecordFields`` but will be rejected following complete
+implementation of `proposal #366
+<https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0366-no-ambiguous-field-access.rst>`_.
+
+There are various possible ways we could permit writing such updates:
+
+* with an expression type signature on the record being updated, e.g. ``(r :: T
+  a b) { f = v }`` (as in this proposal);
+
+* with new syntax, for example ``r { T | f = v }``, ``r T.{ f = v }`` or ``T { f = v .. r }``,
+  where ``T`` is a type constructor name without parameters (as in `proposal
+  #310 <https://github.com/Ericson2314/ghc-proposals/blob/disambiguate-record-update/proposals/0000-disambiguate-record-update.rst>`_);
+
+* by allowing a data type name to be used as if it were a module qualifier,
+  e.g. ``r { T.f = v }`` (a small part of Local Modules `proposal #283
+  <https://github.com/goldfirere/ghc-proposals/blob/local-modules/proposals/0000-local-modules.rst>`_).
+
+This proposal currently assumes the first of these options, but either of the
+alternatives are plausible, and are discussed further below.
+
+
+Proposed Change Specification
+-----------------------------
+
+If:
+
+1. ``DisambiguateRecordFields`` is enabled,
+
+2. a record update has an expression type signature for the record being updated, and
+
+3. the type is syntactically an application of a data type or newtype constructor ``T``;
+
+then name resolution for the fields being updated will consider only:
+
+a. fields of the type ``T`` (if it is a record type),
+b. record pattern synonym fields that are unassociated, and
+c. record pattern synonym fields that are associated with the type ``T``.
+
+That is, when resolving the field names ``f1, ..., fN`` in the record update
+``(r :: T a1 ... aM) { f1 = v1, ..., fN = vN }``, if ``T`` is a data type
+constructor then fields associated with type constructors other than ``T`` will
+be ignored.
+
+If a type signature is given but the type is not syntactically a data type
+(e.g. it is a type variable or type synonym), the normal name resolution rules
+apply.
+
+
+
+Examples
+--------
+
+Consider an expansion of the example from the motivation: ::
+
+    {-# LANGUAGE DuplicateRecordFields #-}
+
+    data S = MkS { f :: Char }
+    data T = MkT { f :: Int }
+
+    foo :: S -> S
+    foo x = x { f = 'c' }          -- rejected: f is ambiguous
+
+    foo2 x = (x :: S) { f = 'c' }  -- accepted: f refers to the field of type S
+
+    foo3 x = x { f = 'c' } :: S    -- rejected: outer type signature ignored
+
+
+    type T' = T
+
+    foo4 x = (x :: T') { f = 'c' } -- rejected: type synonym not expanded
+
+
+Pattern synonym fields
+~~~~~~~~~~~~~~~~~~~~~~
+
+Where there are unassociated pattern synonyms fields, even the presence of a type
+signature may not be enough to disambiguate: ::
+
+    {-# LANGUAGE DuplicateRecordFields, PatternSynonyms #-}
+
+    data T = MkT { k :: Int }
+
+    pattern MkP{k} = k
+    pattern MkQ{k} = (k, ())
+
+    foo x = (x :: T) { k = 3 }  -- rejected
+
+In this example, despite the presence of the type signature, it is still
+ambiguous whether ``k`` refers to the field of ``MkT`` or of ``MkP``.  It cannot
+be ``MkQ`` but that is evident only after type-checking.  Accordingly, this
+example will be rejected.
+
+(This program currently triggers a bug in GHC, see `GHC issue #21898
+<https://gitlab.haskell.org/ghc/ghc/-/issues/21898>`_.)
+
+
+Associated pattern synonyms
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When a pattern synonym field is exported, it may be "associated" with a data
+type constructor, like this: ::
+
+    {-# LANGUAGE DuplicateRecordFields, PatternSynonyms #-}
+    module M (T(MkP, g)) where
+      data T = MkT { f :: Int, x :: Bool }
+
+      pattern MkP{g} = MkT g True
+
+This means that importing ``T(..)`` will bring ``MkP`` and ``g`` into scope.
+This proposal takes advantage of these associations to help with disambiguation
+of pattern synonyms: ::
+
+    {-# LANGUAGE DisambiguateRecordFields #-}
+    module N where
+       import M (T(..))
+
+       data S = MkS { g :: Int }
+
+       foo x = x { g = 0 }  -- rejected: g could refer to the field of S or MkP
+
+       bar x = (x :: T) { g = 0 } -- accepted: must refer to MkP because S is ignored
+
+
+Data families
+~~~~~~~~~~~~~
+
+An awkward corner remains with this proposal, namely data families where the
+same field name is defined multiple times by different data instances, for
+example: ::
+
+    {-# LANGUAGE DuplicateRecordFields, TypeFamilies #-}
+
+    data family T a
+    data instance T Int  = MkTInt  { f :: Int  }
+    data instance T Bool = MkTBool { f :: Bool }
+
+    foo x = (x :: T Int) { f = 3 }  -- rejected
+
+Notice that the type constructor ``T`` and field name ``f`` are not sufficient
+to uniquely identify a field.  Even the record update ``(r :: T Int) { f = 3 }``
+would be rejected under this proposal, because name resolution alone is not able
+to distinguish ``T Int`` from ``T Bool``.
+
+The heart of the issue is that data families allow the definition of multiple
+*different* fields with the same label and belonging to the same type
+constructor for name resolution purposes.  Internally, GHC generates separate
+"representation types" for ``T Int`` and ``T Bool`` but these do not have names
+available to user code.
+
+
+Updates where only one constructor has all fields
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Where there is only one data constructor that has all the fields being updated,
+GHC is able to use this fact to identify the type being updated (see `GHC issue
+#21443 <https://gitlab.haskell.org/ghc/ghc/-/issues/21443>`_).  Following this
+proposal, a type signature may additionally rule out certain types.  For
+example, the following should be accepted: ::
+
+    {-# LANGUAGE DuplicateRecordFields, TypeFamilies, PartialTypeSignatures #-}
+
+    data S = MkS { f :: Int, g :: Int }
+
+    data family T a
+    data instance T Int  = MkTInt  { f :: Int, g :: Int  }
+    data instance T Bool = MkTBool { f :: Bool }
+
+    foo x = (x :: T _) { f = 3, g = 3 } -- accepted
+
+Notice that ``S`` can be ruled out by the type signature, while ``T Bool`` can
+be ruled out because it does not have all the fields being updated, and hence
+the update must refer to ``T Int``.
+
+
+Effect and Interactions
+-----------------------
+
+With this change, it is possible to write a type signature on a record
+expression being updated, and have the type constructor from that signature used
+for name resolution purposes.
+
+
+Partial type signatures
+~~~~~~~~~~~~~~~~~~~~~~~
+
+If a type being updated has many parameters, or they have complicated values, it
+might be annoying to be required to specify them, for (a contrived) example: ::
+
+    data T a b c d e = MkT { f :: a, ... }
+
+    wurble r = (r :: T (Int, Int, Int) Char Monad (,) [[[[[()]]]]]) { f = (1, 2, 3) }
+
+This can be made somewhat more compact using ``PartialTypeSignatures``: ::
+
+    wurble r = (r :: T _ _ _ _ _) { f = (1, 2, 3) }
+
+At present, this will require the user to specify
+``-Wno-partial-type-signatures`` to avoid many needless warnings, but
+this could be addressed using the approach in
+`proposal #491 <https://github.com/treeowl/ghc-proposals/blob/different-holes/proposals/0000-distinguish-partial-sigs.md>`_.
+
+
+``DisambiguateRecordFields`` vs ``DuplicateRecordFields``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The new behaviour will apply when the ``DisambiguateRecordFields`` extension is
+enabled.  ``DisambiguateRecordFields`` seems the natural home for this
+behaviour, as it controls whether GHC uses contextual information about fields
+during name resolution at use sites.  In particular, it has the effect of taking
+the data constructor name into account when resolving field names in record
+construction or pattern-matching, as well as ignoring non-fields when resolving
+a field name in an update (see also `GHC issue #22160
+<https://gitlab.haskell.org/ghc/ghc/-/issues/22160>`_).
+
+The main purpose of ``DuplicateRecordFields`` itself is to enable the definition
+(or re-export) of multiple types containing identically-named fields in a single
+module.  ``DuplicateRecordFields`` implies ``DisambiguateRecordFields``, so
+users enabling the former will automatically gain the latter behaviour.
+
+
+``OverloadedRecordUpdate``
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This proposal changes the specification of name resolution for traditional
+(non-overloaded) record updates.  When ``OverloadedRecordUpdate`` is in use, it
+does not resolve the field names in the same way, so it is not affected by this
+proposal.
+
+If a subsequent change to ``OverloadedRecordUpdate`` provides a way to write
+both non-overloaded and overloaded updates in a single file, then the
+non-overloaded updates will benefit from the changes described here.
+
+
+Costs and Drawbacks
+-------------------
+
+Since name resolution does not have access to type information, type signatures
+must mention the record type constructor explicitly rather than using a type
+synonym.  This means type synonyms are not entirely transparent.  Moreover,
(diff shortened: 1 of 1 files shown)


simonpj:
This proposal is a fix for accepted proposal #366.  That proposal simplifies the disambiguation rules for record updates, but it makes certain programs very difficult to write.  And that has caused us to hold back on implemeting #366. 

This proposal amends #366 by providing a way to resolve ambiguity when you really need "legacy" record update, in cases were overloaded record update won't do the job.

I'm in support.  Let's get this record stuff tidied up!

Profpatsch:
Maybe I’m annoying, but could we please make

```
let u = User { name = "misteruser", … }
in u @User { name = "otheruser" }
```

Work as well, syntactically? It’s just shorter to write & easier to read, and TypeApplications are in wide use in production code already.

adamgundry:
@Profpatsch 

> Maybe I’m annoying, but could we please make
> 
> ```
> let u = User { name = "misteruser", … }
> in u @User { name = "otheruser" }
> ```
> 
> work as well, syntactically? It’s just shorter to write & easier to read, and TypeApplications are in wide use in production code already.

I'm afraid this doesn't make sense to me, because `u` itself does not have a polymorphic type, so `u @User` is ill-typed. I don't see any benefit to this over `(u :: User)`, which is already valid syntax. The proposal discusses the rationale for not introducing new syntax in section 6.1.


> Oh, and the current warning message is not particularly great unfortunately:
> 
> ```
> The record update u {sites = validUserSites} with type User is ambiguous.
> This will not be supported by -XDuplicateRecordFields in future releases of GHC.
> ```
> 
> I’d expect the compiler to at least list the possible types this field could come from. Ideally, the haskell-language-server could propose the correct type from context, since it should have enough information to do so.

The warning already lists the (unique) type from which the field comes when resolved using type-directed disambiguation, namely `User`. It doesn't list other types that have a field of that name, but that information is available to HLS (for example) independently.

Profpatsch:
> I don't see any benefit to this over `(u :: User)`, which is already valid syntax.

The main benefit is that it’s just a lot nicer to type from left to right, does not require putting parens around the thing I just typed, and less verbose.

> I'm afraid this doesn't make sense to me, because `u` itself does not have a polymorphic type, so `u @User` is ill-typed.

Maybe not a polymorphic, but an ambiguous type. I agree it’s slightly different from the usual case of type application, but then again the whole update syntax is a special case.

> The warning already lists the (unique) type from which the field comes when resolved using type-directed disambiguation, namely `User`. It doesn't list other types that have a field of that name, but that information is available to HLS (for example) independently.

All I can tell you is that the warning message is not particularly helpful in practice, especially since HLS does not yet have any support for dealing with this.

proposals/0000-disambiguating-record-updates-using-type-signatures.rst
```diff
@@ -0,0 +1,546 @@
...
+    M.hs:7:13: warning: [-Wambiguous-fields]
+        The record update x {f = 'c'} with type S is ambiguous.
+        This will not be supported by -XDuplicateRecordFields in future releases of GHC.
+      |
+    9 | foo x = x { f = 'c' }
+      |             ^^^^^^^
+
+The GHC developers would like to remove support for this (mis)feature and turn
+the warning into a fatal error.  However, the question arises as to how such
+updates should be written instead.
```
konsumlamm, line 50: Why is this considered a misfeature? The type of `x` is clearly not ambiguous (it's `S`), so why does the warning say that it is ambiguous?

adamgundry: It's ambiguous at name resolution time. See https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0366-no-ambiguous-field-access.rst

konsumlamm: I see.

goldfirere:
I won't stand in the way, but I'm not a huge fan of this. Like the old disambiguation syntax, it reuses a type annotation to affect naming, which GHC struggles with. I would much rather build on existing syntax for affecting naming. Here are some ideas:

```hs
data T1 a b c = MkT1 { x :: Int }
data T2 d e f = MkT2 { x :: Bool }
```

```hs
f y = y T1.{ x = undefined } -- 1

f y = y { T1.x = undefined } -- 2

f y = y { T1 | x = undefined } -- 3

f y = y { x :: T1 = undefined } -- 4
```

NB: No type parameters. None of these have type annotations! (Even the last one.) Maybe (2) overlaps with a module qualification, and so it's impossible. But the others all seem plausible to me. (4) is strange, indeed, but maybe users will like seeing the `::` there: it's more obvious what's going on than (3), say. And (1) is strange because `T1` isn't a module. (Though I feel compelled to say that #283 would solve this problem.)

In thinking about this, here is what I don't like:
* Every time we have carefully tried to use a bit of syntax to mean something beyond what the syntax normally means, it's gone wrong. Examples: CUSKs depended on a very specific syntactic criterion that was always hard to recognize and work with; partial type signatures depend on the existence of a `_`, even though there are scenarios when we want a partial signature without a `_` (when we want to infer a kind) and don't want a partial signature even when we have a `_` (when we want to use visible kind application, skipping an argument); and the old disambiguation syntax was like this, too. In this particular case, what about `((e :: ty)) { ... }`, with two sets of parentheses? Would that use the new disambiguation feature? What about `(id (e :: ty)) { ... }`? I'm sure the answer there is "no". What about `(e :: ty) { ... } { ... }`? I'm sure the answer there is "no", too. Yes, these are strange cases, but the design here fundamentally leads to syntactic fragility.
* This design requires the user to write much more than they really need to: the type parameters of the type involved. And the only way to avoid doing so is to use a partial type signature, which then produces diagnostics that need to be suppressed. And it all creates extra unnecessary work for GHC, in checking the signature, and perhaps filling in the type-level holes.

I'm in full support of creating a new disambiguation mechanism here -- I just worry that we'll regret the concrete syntax being proposed.

goldfirere:
For those following along via email: see the edit at the top of the previous comment, which adds a little context.

adamgundry:
@goldfirere thanks for the comments. I agree that this is a somewhat subjective point.

For me, the main downside of introducing new syntax is that it is more *visible* complexity for users to learn and tools to support (and in particular, it won't be usable on older GHC versions) for what is a relatively rare case. That's why I like the idea of reusing the existing syntax, which will work across old/new versions and "obviously" mean the right thing to users who may not have encountered it before.

I accept that it leads to more *invisible* complexity, in particular the fragility that certain refactorings may cause a program to cease compiling. But GHC should be able to issue reasonable diagnostics in such situations, explaining precisely where a type annotation is required and why.

As an alternative, my preference would be for the "local modules lite" approach (your (2) and my section 6.2). That requires no new syntax, and while it requires quite a bit more specification and implementation work, it might be a nice direction of travel.

I'd be interested to know what others think.

simonpj:
> As an alternative, my preference would be for the "local modules lite" approach (your (2) and my section 6.2). That requires no new syntax, and while it requires quite a bit more specification and implementation work, it might be a nice direction of travel.

It might be good to flesh this out.  E.g.  is `T2.x` (where `T2` is the name of a data type) valid anywhere, or only in a field on the LHS of a record update.  "Anywhere" seems better.   I suppose that would also give an unambiguous way to refer to the record selector, even iwth DuplicateRecordSelecctors.

adamgundry:
> It might be good to flesh this out. E.g. is `T2.x` (where `T2` is the name of a data type) valid anywhere, or only in a field on the LHS of a record update. "Anywhere" seems better. I suppose that would also give an unambiguous way to refer to the record selector, even iwth DuplicateRecordSelecctors.

Yes, "anywhere" seems rather better in that it permits unambiguous references to selectors. I think actually the specification wouldn't be vastly difficult, but I'm worried about the impact on the implementation, at least if we allow a module qualifier as well as a datatype qualifier. Suppose `M.T.x` could refer either to a module `M.T` that exports some `x`, or to a module `M` that exports a datatype `T` with a field `x`. There's nothing in principle wrong with allowing both of these and resolving which is meant based on the import list. But from an implementation perspective, it seems a bit awkward that a `ModuleName` inside a `Qual` `RdrName` can no longer be a single string but must (possibly) be split up into a module part and a datatype name part.

Perhaps this is solvable with a bit of effort, though. Or perhaps the initial version could require the datatype name to be unqualified (so `T.x` might refer to either a type `T` in scope unqualified, or to a module `T`, but `M.T.x` always refers to a module `M.T`)?

tysonzero:
@Profpatsch @adamgundry

For `TypeApplications` perhaps the following would be more appropriate:

```
let u = User { name = "misteruser", … }
in u { name = "otheruser" } @User
```

I think it's reasonable to think of `{ ... }` as a weird postfix `forall r. r -> r` function, meaning `({ ... } @T)` is `T -> T` and `u { name = "otheruser"} @User` is morally parsed as `u ({name = "otheruser"} @User)`.

goldfirere:
I think @tysonzero's alternative is worth considering, as well. I don't love that it's postfix, but it fits the need here.

Still, my favored way forward is the module-prefix-y one.

adamgundry:
My opinion remains that new syntax is a red herring here. :smile_cat: 

Normal application after a record update is technically possible, albeit rarely well-typed. (For example, if you define `pattern P{x} = x` then `undefined { x = id }` is a complicated way of writing the identity function, so `undefined { x = id } 42` is valid.) Thus it would be strange if a postfix type application applied to the record update rather than the whole expression.

endgame:
This is very attractive to me as an Amazonka maintainer — if I understand correctly, it means the generated `amazonka-*` AWS service bindings will continue to compile cleanly when the old rules around record updates go away, since the generated code already contains type annotations on its record updates:

https://github.com/brendanhay/amazonka/blob/beaad7b9703e3c32be0d544af58cc4d21adac7a7/lib/services/amazonka-sts/gen/Amazonka/STS/AssumeRole.hs#L843

If I'm wrong about this, please let me know.

simonpj:
> This is very attractive to me as an Amazonka maintainer 

What is "this"?  The proposal as written?

As written it says that `(r :: T) { x = 5 }` will typecheck (when `x` is used in multiple records); but `r { x=5 } :: T` will not.  Your example link looked more like the latter than the former. But since it is generated code, presumably it would not be hard to switch?

endgame:
I misread the proposal and thought it would work as currently generated. Either way, I'm in favour of a backwards-compatible way of asserting which record type is involved in an update, provided that it isn't going to seed problems down the line.

My gut feeling is that using an existing token like `::` should make it easier to update the amazonka generator, since there's a chance it won't need new `haskell-src-exts`. But I haven't confirmed that `haskell-src-exts`' AST permits this.

adamgundry:
@endgame thanks for flagging up that Amazonka relies on type annotations for disambiguation; it strengthens the argument that we should do *something* rather than simply dropping the feature (the alternative in section 6.5). And perhaps we should after all adopt the alternative in section 6.7 to permit an "outer" type signature as well as an "inner" signature? That doesn't seem like a big ask from the implementation, and may reduce the breakage...

endgame:
Up to you - it probably won't be that hard to change the generator to use "inner" type annotations, if they are "more principled" or something. Consider a record with a type parameter, such that overwriting the field changes the type of the record - is it better to use inner or outer signature there?

But I would very much appreciate some way to do this that's backwards-compatible and doesn't require me to generate a custom import list for every record type in a generated service package (`amazonka-s3`, `amazonka-ec2`, etc.).

Ericson2314:
This is a good proposal. I prefer the `Type.field` alternative for most things, but we can introduce it later as a nicer syntax for the 95% case. Data families are very annoying and I don't see a better solution than `::` type signatures for them. (I don't think they common enough to warrant an entirely new syntax when `Type.field` is so much more intuitive.)

Perhaps it would make sense to steer people towards `Type.field` if we had both by making `::` disambiguation *only* work for data families (where there is no alternative). That's kinda weird, but I think the nudge is healthy.

Glad to see more progress on the long march through the record wilderness!

parsonsmatt:
I don't understand the direction of GHC wrt vanilla record updates.

GHC clearly *can* do this, even though (at the name level) it is ambiguous. Removing the feature (and turning a warning into an error) seems strictly worse from a UX perspective. And all the alternatives add a bunch of syntax to disambiguate something that *is not ambiguous* (from the whole program perspective). 

I understand the technical reason: GHC's renaming phase happens first, and the renamer considers the update to be ambiguous, so it has to do some weird punting to the type checker to properly resolve the name.

What I *don't* understand is: why are we allowing an internal implementation detail of the compiler to make the language *worse*? Why aren't we refactoring or changing the implementation to support this better?

It feels *absolutely bonkers* to me that GHC, with it's renown for type inference, is now requiring a type annotation on a *value* so that it can perform a simple field update. Like GHC is totally fine with

```haskell
module Main where

import Control.Monad

main = do
  print 10
  foobar 1 3

foobar x y = do
  when (x == 3) (print x)
  when (y / 4 < 0.2) (print y)
```

But somehow writing:

```haskell
setName :: String -> User -> User
setName newName u = u { name = newName }
```

is beyond what GHC should be capable of and you need to write `(u :: User)` even though you've already given the type ascription.

I do feel a bit bewildered by this, and I hope that isn't coming across as angry or upset. I'm mostly just confused why this is such a hard problem, especially since it seems to already be solved? and is the regression buying GHC users anything of value?

Ericson2314:
@parsonsmatt I can justify this as someone who thinks that `HasField` is bad: unhygienic ad-hoc overloading, and the old ambiguous fields this is straight-jacketting is bad for the same reason.

I am not sure that's how others feel, but it does make for some nice reasoning here. The renaming before type checking invariant isn't just the consequence of implementation details, but also a user-visibility feature.
 - It allows for simpler mental models, providing a baseline of "things I should be able to understand even if I am totally lost with the types".

It also affects what we can do when there *are no types*, i.e. we we have a program that doesn't type check:
 - People can mentally still do renaming in their head
 - GHC can still do renaming
 - HLS can still do renaming

That least point should be a killer app here. I want a best-in-class IDE, which means I do want al sorts of inlining, outlining, renaming, refactoring tools to work even if the code doesn't type check. Most of the time we working on Haskell programs the code doesn't type check, and I want my tools to be just as useful during that time!

This is also why I hope we get the `TyCon.field` and `Tycon.DataConstructor` stuff too. The `(u :: User arbitrarilyComplexArgs) { name = name }` syntax is still action at a distance, and kind of a cop-out. It is necessary for Data Families, but in the vast majority of cases `u { User.name = name }` is much easier to read and makes mental model stuff much clearer.

Ericson2314:
Basically these are my principles:

- Positional stuff is bad: positional stuff cause bugs and migration woes, is unsuitable to industry / programming in the large.
- Verbosity is fine: People spend much more time reading the code than writing it anyways. IDE autocomplete can help with the writing part. An *inability to abstract* is bad, but that is not necessarily the same thing.
- Extra defensive tasks is terrible: If there is an thing programmers should do that is annoying in the short term but worth it in the long term, this is a [Taylorist](https://en.wikipedia.org/wiki/Scientific_management) failure of language design.

In that regards the sole problem with old primitive
```haskell
data Foo = Foo { foo_bar :: String, foo_baz :: String }

y = x { foo_bar = "" }
```
is *not* that it is verbose, but that it is a waste of effort and sisyphean task to get *all of ones coworkers* and *all of the authors of ones dependencies* to do it.

On the contrary,
```haskell
data Foo = Foo { bar :: String, baz :: String }

y = x { Foo.bar = "" }
```
is hardly less verbose, but it *does* get rid of the problem of terrible incentivization, and that is the only problem here worth solving.

parsonsmatt:
So I understand the motivation:

```haskell
data User = User { name :: String }
data Dog = Dog { name :: String }
```

We'd like HLS to be able to rename `User { name }` into `User { userName }` without needing to know the types. So the following:

```haskell
changeName :: User -> User
changeName u = u { name = "new name" }
```

is ambiguous, such that HLS would have a hard time (only knowing names, not types) knowing to perform this change:

```diff
 changeName :: User -> User
-changeName u = u { name = "new name" }
+changeName u = u { userName = "new name" }
```

But, with `(u :: User) { name = .. }`, then HLS would be able to perform the global-rename? Same with `u { User.name = ... }`? But a bare `u { name = ... }` expression is ambiguous without running enough of the type checker to know what's up.

Ericson2314:
`u { User.name = .. }` it should be able to rename. `(u :: User) { name = .. }` let's just say it *won't* be able to rename (though with some elbow grease one could implement it without type checking), but that should only be needed for data families and thus rare as hell.

IDE can just track those places it couldn't procede and ask the user to manually resolve them,

Ericson2314:
Also,
```haskell
data User = User { name :: String }
type Customer = User

y = x { Customer.name = "" }
```
should *not* work in my book. We should be dispatching of Type Constructors not Types, except for data families where we are shit out of luck.

simonpj:
> It feels absolutely bonkers to me that GHC, with it's renown for type inference, is now requiring a type annotation on a value so that it can perform a simple field update.

It's a fair question.  The trouble is this.   Type inference generally is cleverly set up so that it doesn't matter in which order the type checker looks at the term -- the result is the same.  But to disambiguate record updates it does matter.  You give the example
```
setName :: String -> User -> User
setName newName u = u { name = newName }
```
But suppose it wsa
```
-- No type signature
--   f :: User -> Int
setName2 newName u =  (u { name = newName }, f u)
```
Now it's a bit more indirect.  We don't know what type `u` has until we see the call to `f`.  But we may not have visited the second component of the tuple yet.  Hmm.  Maybe we shoudlu defer that record update until we've sorted out the rest of the term. But no, perhaps something in the record update is needed to unravel *another* record update in the rest of the term.

Why can't we do some kind of clever order-independent thing for record updates, as we do for the rest of type inference?  That is *exactly* what the `HasField` class and overloaded record access and updates are all about.  But alas all of that technology works well with *monotypes*. And records often contain *polytypes*. Indeed the main reason we want to continue to support "legacy" records is because of polytyped record fields (and type-changing updates).  In effect, we need to know which field is being mentioned before doing type inference.

You might reasonably ask: couldn't we do some kind of top-down ad-hoc special case algorithm that would support `setName` while rejecting `setName2`.  Maybe we could.  But it imposes a significant pile of extra complexity on both the *specification* of the language and on its *implementation*.  For example would this be OK?
```
 ... f (u { name = newName })...
-- where f :: User -> Int
```
You could argue that it's obvious that we mean the `f` from the `User` type; but it's a bit less obvoius than before. One could perhaps write down a complete set of rules.  But it seems a high price (in complexity) for this particular piece of expressiveness.

So I guess my answer is: this isn't just an implementation detail peeping through, which was your legitimate concern. With duplicate record selectors there is something pretty deep and subtle going on, and I don't konw of a principled solution that still fully meets everyone's goals.  Instead we have currently have two solutions, each principled and explicable
* Overloaded record fields `HasField` etc, which let you write record-polymorphic functions.
* Named fields, with a simple and expliciable disambiguation mechanism, such as that described here.

What is tantalising here is that the programmer is in no doubt about which of the many fields `f` she has in mind; we just need a way for her to express that fact.  (Personally I'm qiute keen on using a qualified name rather than a type signature.)  If you have any other ideas for disambiguation, we are all ears.

parsonsmatt:
Ahhh, I think I have a better handle on the fundamental issue. Prior to `DuplicateRecordFields`, you simply couldn't define a duplicate name. `DuplicateRecordFields` allowed you to define duplicate *names* in a module, but *only if* those names were record fields. This is Fine if the names can be resolved to a type-class-like-thing, but record fields behave more like functions. So `HasField` is a way to try and make record fields more like type class methods and less like functions.

Thanks for the explanation, I feel much less confused 🙏🏻

simonpj:
> you simply couldn't define a duplicate name. 

Correct, if by "name" you mean "Entity called `foo` defined in module `Bar`".  So its "original name" (defining module + occurrence name) is `Bar.foo`.   And indeed if two records `S` and `T` both in module `Bar` define a field `wombat` then there are two quite different things: the `wombat` from `S` and the `wombat` from `T`.

> This is Fine if the names can be resolved to a type-class-like-thing

I would rather say that given
```
module M where
  class C a where
    op :: a -> a
```
then there is just one `op`, the class method, defined in module `M`, with original name `M.op`.  You can call it at differnet types, but `op` itself is just a record selector that picks out the field from the `M.C` record.  So again, no "duplicate name" problem.

I'm glad the explanation helped. I would love to have a single, principled design than did everything (fully general polytyped fields; type changing updates; duplicate record fields) but I don't know how to do that.  And Haskellers have been thinking about this challenge for a long time, so I'm not optimistic about a sudden technical breakthrough. 

Meanwhile, we content ourselves with a pair of principled designs, and accept that even then they are not perfect.  Your `setName` example is a good demonstration of what is unsatisfying.  As I say above, we *could* do that, but it's a complex thing to specify and implement.  So I'm content with something that has a decent workaround: in this caese, adding a signature or maybe a qualified name.

------------------------

The bit I'm least comfortable with is that the workaround fails for data families, and we have no decent workaround for that. 
```
data family D a
data instance D Int = D1 { x :: Int } | D2 { x::Int, y:: Bool }
data instance D Bool = D3 | D4 {x :: Bool} | D5 { x::Bool, y :: Char }
```
So there are two completely different `x` fields in scope.  How can we make sense of
```
f r = r { x = True }
```
We can't say `D.x` because that still doesn't tell us which `x`.   Nor can we say
```
f r = (r :: D Bool) { x = True }
```
because the `D` part of the signature (which is all this proposal looks at) isn't enough to say which `x`.

So, under this proposal, you have to write out the update yourself
```
f r = case r of
         D3 -> erorr "no x field"
         D4 _ -> D4 x
        D5 _ y -> D5 x y
```
Maybe this is a sufficiently rare case that we should not worry, but it still seems ... unclean.

One way to address the data-family problem, which the implementation uses internally, is to use the name of the first *data constructor* as a proxy unique name.  Thus we would say `D3.x` (using the first constructor as proxy):
```
f r = r { D3.x = True }
```
Or you could allow *any* data constructor as proxy, s you could say `D4.x`.  Even then it's arguably misleading because the *actual* constructor in any particular call might be `D5`.  It's a mosquito-like problem: small, but very irritating.

adamgundry:
I'm inclined not to worry about data families for now. As you say, there is a workaround, and generally speaking users can choose not to define the same field name for multiple instances of the same data family in the same module.

The current proposal doesn't support them because it seems easier to specify that name resolution depends only on the name of the type constructor. But we could imagine extending it to say that for data families we compare the parameters syntactically (so a `data instance D Bool` would match a record update with a signature for `D Bool`, but not for `D B` where `type B = Bool`).

---

Several people have expressed the opinion that they prefer the module prefix approach (the alternative described in section 6.2). Should we do that instead? My concern is that it means for a qualified name `M.T.foo` it is not obvious when parsing whether the module name is `M.T` or `M`. Presumably we could parse `M.T` and then have the renamer split it up and look for both `M` and `M.T`, but that feels like it might get tricky to implement? (In GHC implementation terms, we currently have `Qual ModuleName OccName :: RdrName`, where `ModuleName` is just a `FastString`, and that would no longer be enough.)

On the other hand, one thing we could do that fits nicely with the module prefix approach is say that even when `OverloadedRecordUpdate` is enabled, an update that mentions a prefix (e.g. `r { M.foo = x }`) is necessarily non-overloaded (and hence can mention higher-rank fields etc.). I have a [WIP proposal for this](https://github.com/adamgundry/ghc-proposals/blob/non-overloaded-updates/proposals/0000-overloaded-record-update-fallback.rst).

goldfirere:
Addressing the "let's just do inference, please" front: OCaml does this. That is, if you have two different record types with the same field name(s), OCaml will use types to disambiguate. But my limited experience over there suggests we should not follow that example. It's hard to predict when disambiguation will work, and in practice needs more type annotations than you would generally think. Anecdotally, OCaml users complain that its type inference isn't as good as it should be, and long-term OCaml maintainers believe that this "isn't as good as it should be" is really all about exactly this kind of disambiguation. By GHC maintaining a strict barrier between names and types, we avoid that small potential problem.

simonpj:
> Several people have expressed the opinion that they prefer the module prefix approach (the alternative described in section 6.2). Should we do that instead? 

My instinct is that we should.  It more clearly signals disambiguation than the type signature does.  And for parameterised types I don't really like being forced to write out the type arguments... or having to use type holes `T _ _`.

But it has ramifications.  Consider
```
data S = MkS { x :: Int }
data T = MkT { x :: Bool }
```
Then we might like to be able to do a record update with `r { S.x = 3 }`.  But we might equally like to do record *selection* with `S.x r`. It would seem funny to allow one but not the other.

> Presumably we could parse M.T and then have the renamer split it up and look for both M and M.T, but that feels like it might get tricky to implement? (In GHC implementation terms, we currently have Qual ModuleName OccName :: RdrName, where ModuleName is just a FastString, and that would no longer be enough.)

Yes.  I think we'd end up with `type ModuleName = [FastString]` rather than `type ModuleName = FastString`.   I don't think that would be hard to implement, and we might ultimately want it anyway (e.g. for Richard's now-parked local modules proposal), but someone would have to try it out to see how hard it was and what perf impact it had.  Notably, `FastString`s have a Unique, and we'd want `ModuleName` to have a `Unique` too. 

Alternatively, and more directly, consider bringing into scope a field name `x`, imported from data type `T` in module `P.Q`.  In the GlobalRdrEnv we add an entry for it as "module name `P.Q`, occurrrence name `x`".  But I suppose we could *also* add it as "module name `P.Q.T`, occurrence name `x`", by making a fake "module name" for it.  On reflection I think that would just work.

endgame:
With my (still relatively new) amazonka-maintainer hat on, I remain strongly in favour of syntax that is comprehensible to existing GHC releases, unless doing so is really going to constrain GHC in the future. Amazonka has not had a release for many years, and bringing it up to scratch has taken a long time. (It's been nearly 18 months, and I'm hoping for a Hackage release in Q1 2023.) I would really like to be able to ship something that supports both old and new GHCs, in case releases stall again for some reason.

Ericson2314:
@endgame that can be acomplished by allowing the `(x :: T) { ... }` syntax over a transitional period. But I hope we can still agree that `Tycon.x` that is still a nicer place to end up at.

adamgundry:
@endgame that's a good point about `(x :: T) { ... }` having the (not insignificant) advantage of working with both old and new GHCs. We could support both (perhaps even `x { ... } :: T` as well in the interests of backwards compatibility) but treat the type signature versions as deprecated and continue to issue a `-Wambiguous-fields` warning for a few releases?

endgame:
I am more than happy to rework the amazonka generator to emit `(x :: T) { ... }`, that's no trouble, and I definitely support taking the time to migrate towards a "definitely right" solution. Whether the accessors end up qualified by type or constructor name is up to you. But I think that it'd be really good to have a couple of major versions that supported some kind of transitional syntax as explicit qualification, as opposed to the type-directed `-XDuplicateRecordFields` that happens to work right now.

I think we might want a sequence of GHC releases that migrates something like:

1. Warn that type-directed disambiguation is going away, but supports `T.x` updates. Recommend `T.x` in the updates and note `(x :: T) { ... }` updates will work as an "if you absolutely must be compatible, but they won't be supported forever". If the message gets too long, put the details on https://errors.haskell.org?
2. Stop supporting type-directed `-XDuplicateRecordFields`, but support disambiguation using both `T.x` and `(x :: T) { ... }` updates. Warns that `(x :: T) { ... }` disambiguation will go away in a future release.
3. (After libraries like `haskell-src-exts` have had a chance to add support for `T.x` disambiguation) Remove `(x :: T) { ... }` disambiguation from GHC.

Ericson2314:
> After libraries like `haskell-src-exts` have had a chance to add support for T.x disambiguation

And I'll add that splitting out the GHC AST and Parser into a separate library is important so GHC development need not be blocked on such 3rd party steps! :)

AntC2:
Can I make a small request for when this proposal gets implemented/released. This is wrt how we document the functionality. (Apologies if this is the wrong place to ask.) From this thread, I'm not picking on this comment specifically, but I think it might be a common confusion that the docos don't dispel.

> @parsonsmatt This is Fine if the names can be resolved to a type-class-like-thing, but record fields behave more like functions.

H98 field accessors are indeed functions. `HasField`/`SetField` are indeed type-class-like things. But (to quote [the current docos](https://downloads.haskell.org/~ghc/9.4.3/docs/users_guide/exts/overloaded_record_dot.html)):

>```
> data Person = Person { name :: String }
> data Company = Company { name :: String, owner :: Person }
>
> main = do
>   let c = Company { name = "Acme Corp."
>                   , owner = Person { name = "Wile E. Coyote" } }
>   print $ c.name ++ " is run by " ++ c.owner.name
> ```

The `name` in `c.name` is neither a function nor a type-class method, it's a literal. I can't go

```
aliasName = name      -- presuming there's only one name in scope
print $ c.aliasName
```

Because that'll try to generate a pseudo-instance `Hasfield "aliasName" Person String`. Substitution of equals doesn't work -- unlike usual Haskell. I suggest changing this wording

> When `OverloadedRecordDot` is enabled one can write `a.b` to mean the `b` field of the `a` record expression.
> ...
> When considering `a.b`, the `b` field that is meant is determined by solving `HasField` constraints. 

to

When `OverloadedRecordDot` is enabled one can write `a.b` to mean the field named `b` of the `a` record expression.
...
When considering `a.b`, the `b` field that is meant is determined by solving `HasField` constraints for field name `"b"`.

Ericson2314:
@AntC2 It is fine and good, I think, to discuss how prospective features should be documented while they are being proposed, but *this* proposal has nothing to do with `OverloadedRecordDot`.

PR #539, opened 2022-09-15
[WIP] Extended Haskell Program Coverage (HPC)

Tritlo:
This is a WIP of the Extended Haskell Program Coverage (HPC) proposal, to be updated.

### DIFF
--- a/proposals/extended-hpc
+++ b/proposals/extended-hpc
@@ -1,0 +1,102 @@
+---
+author: David Binder, Matthías Páll Gissurarson
+date-accepted: ""
+ticket-url: ""
+implemented: ""
+---
+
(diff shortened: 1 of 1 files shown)


simonmar:
Would be good if the proposal could include a comparison with the [`:trace` command in GHCi](https://downloads.haskell.org/ghc/latest/docs/users_guide/ghci.html#tracing-and-history).

L-as:
You probably want to mark this as draft.

bgamari:
In my opinion the eventlog would be a better target for this sort of tracing than `.tix`; it already has an space and CPU-efficient representation, it handles concurrency well, and it already has a pretty good tooling story. For what it's worth, I have also wondered whether HPC itself shouldn't rather target the eventlog; certainly this is the design I would have chosen if I were to design HPC today.

PR #540, opened 2022-09-29
jsem: parallelism semaphores

sheaf:
This proposes a mechanism for dynamically scheduling parallelism across multiple GHC invocations through use of a system semaphore.

This change requires coordination with `cabal-install` and `stack`.

Rendered: todo.


aspiwack (commented):
I'm glad that this issue is being tackled. I've got a few comments below.

proposals/0540-jsem.rst
```diff
@@ -0,0 +1,342 @@
...
+We have concerns that rapidly adjusting the number of capabilities may have
+adverse effects: we propose that GHC will rate limit the release of tokens so as
+to reduce this thrashing. The exact mechanism of rate limiting is unspecified
+here.
+
+We also believe that GHC invocations should not be too eager to release
+resources, so we propose a minimum delay between acquiring a resource and
+releasing a resource. This allows us to prioritise completing a single large
+package before moving on to other work. The exact mechanism through which this
+is achieved is also left unspecified in this proposal.
```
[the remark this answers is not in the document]
mpickering: I tried to clarify that one issue is that if you are compiling N packages at once you have N copies of the EPS so you will load interface files multiple times (once per GHC session) which will use more memory than if you are compiling one package with N capabilities. Therefore we are trying to skew the system so that we prefer compiling 1 with N rather than N with 1.

proposals/0540-jsem.rst
```diff
@@ -0,0 +1,342 @@
+
+jsem: parallelism semaphores for GHC
```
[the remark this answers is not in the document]
aspiwack: My idea was more that the work “semaphore” is strange in this context, it's an implementation mechanism rather than a description of the feature. I would probably look at something along the idea of “job server”. Or maybe “shared jobs”.

But, as you rightly point out, it's not worth losing hair (and more importantly time) over.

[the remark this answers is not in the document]
nomeata: The flag is likely only used by tools? Then a long descriptive flag doesn't hurt. `-juse-jobserver` maybe?

proposals/0540-jsem.rst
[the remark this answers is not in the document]
aspiwack: Thanks, this makes sense.

The general idea is:
- Having several GHC running at the same time increases memory usage
- There is one GHC running by package
- If a package has enough parallelism to saturate the number of parallel job specified, it is therefore presumably better to use all the jobs on this packages: a single copy of GHC is running.
- To achieve this, you will make it so that a copy of GHC which already have jobs running is allowed to run new jobs, automatically having priority over other GHC instances.

Did I get this right?

proposals/0540-jsem.rst
```diff
@@ -0,0 +1,342 @@
...
+
+- We expect ``cabal-install`` and ``stack`` to be the only users of this feature
+  in the near term. We think the proposed protocol is adequate for this use case.
+
+- We can extend GHC to use the GNU make Jobserver protocol in the future, if
+  there are users for it.
+
+Notes:
+
+- rust's ``cargo`` does implement the GNU make Jobserver protocol [13]_ ,
```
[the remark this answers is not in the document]
nomeata: I was thinking the same. I skimmed the nix and dune PR, and basically it seems to be ”the protocol is bad”, with “killed processes don't give their tokens back” being one example. 

Is it true that if we implement a bespoke algorithm now, adding `-jmake-jobserver` later will be able to reuse half the work maybe?

Then I'd be ok with doing our own thing until someone shouts.

proposals/0540-jsem.rst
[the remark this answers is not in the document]
mpickering: Yes that's right.

proposals/0540-jsem.rst
```diff
@@ -0,0 +1,342 @@
...
+``N`` of capabilities, then parallel garbage collections in that GHC would
+recruit ``N`` OS threads. Instead we propose that GHC will adjust its number
+of capabilities, via ``setNumCapabilities`` [8]_ , to the number of modules
+it is compiling.
+Note that GHC already calls ``setNumCapabilities`` with the argument to ``-j``.
+
+We have concerns that rapidly adjusting the number of capabilities may have
+adverse effects: we propose that GHC will rate limit the release of tokens so as
+to reduce this thrashing. The exact mechanism of rate limiting is unspecified
+here.
```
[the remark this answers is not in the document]
mpickering: We don't have any answers about the rate-limiting at the moment so that will have to be omitted from the proposal.

proposals/0540-jsem.rst
```diff
@@ -0,0 +1,342 @@
...
+
+The implementation in GHC is self-contained and doesn't impact the rest of the
+compiler much. It does however add a new flag (which interacts with ``-j``),
+and a complete implementation requires coordination with ``cabal`` and ``stack``.
+
+Alternatives
+------------
+
+GNU make supports a Jobserver protocol [9]_ [5]_ which is the same as the
+GHC Jobserver protocol described above, except that:
```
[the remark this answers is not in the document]
mpickering: Done

proposals/0540-jsem.rst
```diff
@@ -0,0 +1,342 @@
...
+
+- Other communities have considered the Make jobserver, and decided some
+  aspects of the protocol are unsuitable:
+
+  - ocaml [10]_ [11]_
+
+  - nix [12]_
+
+- We expect ``cabal-install`` and ``stack`` to be the only users of this feature
+  in the near term. We think the proposed protocol is adequate for this use case.
```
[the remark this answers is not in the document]
mpickering: I am not sure that the make jobserver is a general solution either given how reluctant build systems (nix, ninja) are to implement the protocol (https://github.com/ninja-build/ninja/issues/1139). Hopefully in future there is a protocol that a wide range of tools adopt!

mpickering:
Could we submit this proposal to the committee please?

proposals/0540-jsem.rst
```diff
@@ -274,14 +274,26 @@ GHC Jobserver protocol described above, except that:
...
+However, we have decided to depart from this design, for the following reasons:
 
-  - ocaml [10]_ [11]_
+- Other communities have considered the Make jobserver, and decided that some
+  aspects of the protocol are unsuitable (OCaml [10]_ [11]_ , Nix [12]_ ,
+  ninja [16]_ ). To summarise:
 
-  - nix [12]_
+  - The protocol relies on spawned processes cooperating and returning tokens on
+    termination. If this doesn't happen, semaphore tokens can be lost entirel.
```
nomeata, line 288:
```suggestion
    termination. If this doesn't happen, semaphore tokens can be lost entirely.
```

mpickering:
Any progress on this one @nomeata ?

nomeata:
Sorry, I didn't realize this was submitted formally (it works better to explicitly ping me).

simonpj:
I understand the intent, but I'm totally baffled by the details.   Here are some questions. Please don't educate me here; instead, reply by improving the proposal!   

In Proposed Specification

* "Communicating through a semaphore",  in the preamble to Section 2.  What is "a semaphore".  Of course I know in a CS sense, but (guessing wildly) perhaps the operating systems you have in mind support the notion of a system-wide semaphore?  Or perhaps you mean that a process is running that cabal and ghc will talk to, and which will hand out tokens?  Or what?

* Since there may be many GHCs, and perhaps many cabals, running at once, they must presumably know which semaphore they are talking to.  How do they do that?  Perhaps cabal, having created a semaphore, passes its name to GHC via a command line argument?

* "The GHC Jobserver Protocol".  What is that? Does it exist already?  Or is it a new thing, proposed here?

* What are GHC's responsiblities? (This is the GHC steering commitee after all.)  Is it always a jobclient?    

* "There is a hidden cost in having a GHC program (such as ghc itself) run on fewer CPU cores than its capabilities:"  First mention of "capabilities".  How do these capabilities relate to the semaphores?  What is this "rate limit the release of tokens" stuff?  I'm pretty lost here.

* Examples.  How does cabal know how many tokesn let GHC grab (after all, in any invocation of GHC it doesn't know how many GHC will want) and how many to keep (for itself, presumably)?

* 4.1  You are assuming context that the reader may not have.  For a start, I think unit=package?  Worth saying.  Then I think you are contrasting: 
  * One invocation of GHC compiles exactly one unit, using --make
  * One invocation of GHC can compile many units

  And I think you are saying that if the latter is possible, it's better; and it completely negates the need for the proposal. So the proposal is need only if "it's not always possible to compile everything with a single GHC invocation, e.g. if the build plan involves non-Haskell dependencies somewhere in the middle."  Is that right?

sheaf:
Any updates from the committee?

In a build tool I implemented myself, going from `-jN` to `jsem` took only a couple of lines of changes. I was using `QSem` to control how many `ghc` invocations could be run concurrently; I simply changed that to creating a system semaphore and use the system semaphore instead. The tool can also re-use an existing system semaphore instead of creating it itself (jobclient vs jobserver).

I put up a package candidate [`semaphore-compat`](https://hackage.haskell.org/package/semaphore-compat-1.0.0/candidate/docs/System-Semaphore.html) which provides system a compatibility layer for semaphore functionality over `Win32` and `unix`, which is what I propose to be used in practice (in `ghc`, `hadrian`, `cabal` and `stack`) for system semaphores.

nomeata:
@gridaphobe, that ping went to you :-)

proposals/0540-jsem.rst
```diff
@@ -0,0 +1,384 @@
...
+   it is compiling.
+   Note that GHC already calls ``setNumCapabilities`` with the argument to ``-j``.
+
+   We have concerns that rapidly adjusting the number of capabilities may have
+   adverse effects: we propose that GHC will limit the rate at which it can
+   release tokens to the semaphore so as to reduce this thrashing.
+   The exact mechanism of rate limiting is unspecified here.
+
+B. We believe that GHC invocations should not be too eager to release
+   semaphore tokens. This skews the balance in favour of in-unit parallelism
```
gridaphobe, line 179: At this point I'm still quite unclear on the algorithm for GHC requesting tokens from the job-server. How does GHC decide how many tokens to request in the first place?

proposals/0540-jsem.rst
```diff
@@ -0,0 +1,384 @@
...
+1. To start, ``cabal`` would create a semaphore ``⟨sem⟩`` with 8 available tokens.
+
+2. Next, we compile the ``Bot`` unit, which is a large unit, with many modules,
+   which sits at the bottom of the dependency graph and must thus be compiled
+   before anything else.
+
+   a. ``cabal`` acquires one token from the semaphore and spawns one
+      ``ghc --make -jsem ⟨sem⟩`` invocation.
+   b. This invocation of ``ghc`` notices it has a lot of work to do (many modules
+      to compile from the ``Bot`` unit), so it requests more resources from the
```
gridaphobe, line 213: Does GHC request a specific number of tokens, or just "more"?

proposals/0540-jsem.rst
```diff
@@ -0,0 +1,384 @@
...
+   c. ``ghc`` finishes compiling the ``Bot`` unit, releasing the 7 tokens it
+      acquired.
+   d. ``cabal`` notices the ``ghc`` subprocess has terminated, and releases
+      the final (8th) token to the semaphore.
+
+3. After that, we move to compiling the middle units.
+
+   a. ``cabal`` will acquire tokens from the semaphore and spawn
+      ``ghc --make -jsem ⟨sem⟩`` invocations.
+   b. These ``ghc`` invocations won't query for more resources from the
```
gridaphobe, line 225: > These ``ghc`` invocations won't query for more resources

Why not? Because each unit contains a single module?

proposals/0540-jsem.rst
```diff
@@ -0,0 +1,384 @@
...
+
+   a. ``cabal`` acquires one token from the semaphore and spawns one
+      ``ghc --make -jsem ⟨sem⟩`` invocation.
+   b. This invocation of ``ghc`` notices it has a lot of work to do (many modules
+      to compile from the ``Bot`` unit), so it requests more resources from the
+      semaphore. As no other processes are competing for semaphore tokens, this
+      GHC invocation obtains the remaining 7 tokens.
+   c. ``ghc`` finishes compiling the ``Bot`` unit, releasing the 7 tokens it
+      acquired.
+   d. ``cabal`` notices the ``ghc`` subprocess has terminated, and releases
```
gridaphobe, line 218: What happens if GHC dies suddenly, before having released the additional 7 tokens? Does Cabal somehow force those to be released as well? I'm concerned about possible resource starvation issues.

proposals/0540-jsem.rst
```diff
@@ -0,0 +1,384 @@
...
+spawns ``ghc`` subprocesses. The ``ghc`` subprocesses are jobclients, and they
+communicate by use of the semaphore.
+
+Effects and Interactions
+------------------------
+
+Comparison with multiple home units
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Support for multiple home units (not yet fully implemented in ``cabal``)
```
gridaphobe, line 246: Can you add a link for the multiple home units work and proposal? I wasn't aware of that work.

proposals/0540-jsem.rst
```diff
@@ -0,0 +1,384 @@
...
+- modules are loaded directly into the home unit graph, which avoids having
+  to load the same interface files in different GHC invocations,
+- it doesn't require the entire unit to finish compiling before compilation
+  can start on another unit that depends on it: we can begin as soon as all
+  the modules we need have been compiled.
+
+However, it's not always possible to compile everything with a single GHC
+invocation, e.g. if the build plan involves non-Haskell dependencies somewhere
+in the middle. In comparison, the ``-jsem`` functionality can fit into any build
+system that one might be using, so it supports a wider range of use cases.
```
gridaphobe, line 265: Based on this brief description, it sounds like the multiple home units approach is expected to be strictly better than the job-server approach for pure-Haskell build plans. That's what I would expect the vast majority of Cabal and Stack build plans to be, since neither tool has a notion of multi-language builds.

The point about polyglot builds makes sense in principle, but then the work should be motivated from that perspective, maybe with an example of how it would help bazel's `rules_haskell`. And then I'd have a number of different questions.

proposals/0540-jsem.rst
```diff
@@ -0,0 +1,384 @@
...
+the scheduling.
+In order to implement parallelism at this level, it seems necessary to modify*
+the RTS and GHC's own thread scheduling algorithms. Not only this, the
+implementation would be specific to a platform.
+
+
+The other option for build systems is to use GHC in one-shot mode. In that case,
+the build system can control the scheduling of all the jobs. This is what
+``hadrian`` and ``rules_haskell`` do when building projects (``cabal``
+currently does not).
```
gridaphobe, line 327: This is what I wanted to ask about. Why not granularly split up the compilation by module and invoke GHC in one-shot mode? That would seem to substantially reduce the jobserver protocol complexity, and better make use of resources. I'd like to understand what the downsides are that you're seeing?

gridaphobe:
@sheaf apologies for my tardiness. I've left a bunch of questions inline on the proposal, but my biggest question is around the comparison the with "multiple home unit" work. 

In my view, the way to get optimal resource utilization is to break up the dependency graph into small units, ie individual modules, and then traverse the dependency graph in parallel. That can be done external to GHC (ala bazel / rules_haskell), or it can be done within GHC (`ghc --make`).

Your proposal is trying to find a middle ground, but it's an awkward middle ground because it requires this dance between the **package-level** build system (cabal/stack) and the **module-level** build system (ghc). Given the existing separation of responsibilities, I could see this as a reasonable cut-thru to get a good cost/benefit tradeoff.

**But**, then you say there's an in-progress plan for "multiple home units", which based on your proposal sounds like it is aiming to push the full package-level dependency graph into `ghc --make`. If it's true that such work is already in progress, then why shouldn't we just wait for it to be completed? Who are the concrete clients that will continue to benefit from `-jsem` once the multiple home units work is available?

mpickering:
@gridaphobe There isn't in particular an "in-progress" plan to get cabal to support multiple home units (past the repl) . If there was a client who wanted to fund work like this then it would be possible to investigate.

One big advantage of -jsem is that we have already implemented it in GHC and cabal. There is no more work to be done there. Making cabal support either using one-shot mode or multi home units is a much bigger project, with no clear sponsor and no time-line in mind. So it makes sense to me to accept this proposal so that people can benefit in the intervening years whilst cabal catches up.

proposals/0540-jsem.rst
mpickering: Well you can do that, but it's not how cabal-install works and modifying cabal-install to work like that is a completely separate project which has no financial backing nor good time-bound for how long it would take. Using `--make` mode is also faster than using one-shot mode.

sheaf:
Thanks for taking a look @gridaphobe, you raise many excellent points. I'll try to respond this week.

proposals/0540-jsem.rst
```diff
@@ -0,0 +1,384 @@
...
+
+As far as the GHC jobclient implementation is concerned, we have identified
+two salient points:
+
+A. There is a hidden cost in having a GHC program (such as ``ghc`` itself) run on
+   fewer CPU cores than its capabilities: the stop-the-world cost of garbage
+   collection becomes much more expensive. If we were to give GHC a fixed number
+   ``N`` of capabilities, then parallel garbage collections in that GHC would
+   recruit ``N`` OS threads. Instead we propose that GHC will adjust its number
+   of capabilities, via ``setNumCapabilities`` [8]_ , to the number of modules
```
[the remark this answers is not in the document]
mpickering: No, it should never exceed the number of available cores.

proposals/0540-jsem.rst
mpickering: With the current implementation GHC requests 1 per module.. but this can change in future so we have left it deliberately vague.

proposals/0540-jsem.rst
mpickering: more = 1 per module it needs to compile.

proposals/0540-jsem.rst
mpickering: If GHC dies suddenly then it will indeed not return the 7 tokens it has taken, but cabal can still return the one it owns so things can proceed for a bit longer with less parallelism. However, if GHC dies like this then the compilation has failed so cabal should find this out and report the failure to the user promptly.

proposals/0540-jsem.rst
mpickering: Yes exactly. (At least with the current implementation in GHC).

proposals/0540-jsem.rst
gridaphobe: > However, if GHC dies like this then the compilation has failed so cabal should find this out and report the failure to the user promptly

True, but we could rephrase the scenario a bit such that GHC fails to return the additional tokens, but does finish compilation successfully so Cabal continues with seriously limited parallelism.

What I'm curious about here is whether there's any benefit in establishing a parent-child relationship between tokens. Imagine that in step (d) when Cabal returns the implicit token, any tokens that GHC had requested are automatically returned as well, even if GHC failed to explicitly return them.

Is that overcomplicating things?

gridaphobe:
@mpickering 

> There isn't in particular an "in-progress" plan to get cabal to support multiple home units (past the repl) 

Thanks for the clarification, the proposal gave me the impression that there was an active multiple home units project. I agree that changes the calculus somewhat.

> One big advantage of -jsem is that we have already implemented it in GHC and cabal.

Since it's already implemented, could we see some stats on the perf improvements you've seen in practice?

gridaphobe:
@sheaf 

> Thanks for taking a look @gridaphobe, you raise many excellent points. I'll try to respond this week, and incorporate Matthew's responses into the proposal text.

friendly ping :)

proposals/0540-jsem.rst
sheaf: I'm not familiar with `rules_haskell`, so I can't comment. For me, the real point is that the changes to build tools are very small for this approach, whereas the changes required to `cabal` for multiple home units are complex and have many subtleties.

proposals/0540-jsem.rst
sheaf: The problem with your suggestion is that tokens are directly requested from the semaphore, so Cabal doesn't know who owns which tokens, and thus can't know how many tokens to release. This is by design: coordination is achieved via the semaphore only.

proposals/0540-jsem.rst
sheaf: Otherwise, I think the point is that this feature is not relevant to `rules_haskell`, because `rules_haskell` already fully controls all parallelism because it manually invokes GHC on a per-module basis. I've made a "One-shot mode"' subsection to the "Alternatives" section to try to make that clearer.

sheaf:
Thanks for the ping @gridaphobe; I apologise for the delay. I've updated the proposal in response to your very helpful comments. Could you take another look and let me know what you think?

proposals/0540-jsem.rst
```diff
@@ -158,37 +158,45 @@ Implementation of the protocol
...
+   to the number of tokens it is using (up to the number of available CPU cores).
+   This is because there is a hidden cost in having a GHC program
+   (such as ``ghc`` itself) run on fewer CPU cores than its capabilities: the
+   stop-the-world cost of garbage collection becomes much more expensive.
+   If we were to give GHC a fixed number ``N`` of capabilities, then parallel
+   garbage collections in that GHC would recruit ``N`` OS threads.
+   Note that GHC already calls ``setNumCapabilities`` with the argument
+   to ``-j``.
+
+D. GHC should rate-limit the release of semaphore tokens (the precise mechanism
```
gridaphobe, line 182: Does your current implementation have a view on how/when to release semaphore tokens?

I'm asking not because I think this should be specified (I agree it's an implementation detail), but because it seems like a very subtle and tricky knob to get right. That's a cost that the GHC Maintainers will have to bear and understand, so it would help if you already have a framework for how to reason about it.

gridaphobe:
@sheaf sorry again for the slowness. I left one more comment, and also recommended acceptance to the committee.

angerman:
Given that there is an implementation for this, are there performance comparisons (cpu/memory over time; wall clock) available for cabal/ghc flag combinations: -j, -j8/-j1, -j1/-j8, -j4/-j4, -j8/-j8, -jsem
(Assuming 8 core machine here).


I would expect -jsem to outperform them all?

sheaf:
> Given that there is an implementation for this, are there performance comparisons (cpu/memory over time; wall clock) available for cabal/ghc flag combinations: -j, -j8/-j1, -j1/-j8, -j4/-j4, -j8/-j8, -jsem (Assuming 8 core machine here).
> 
> I would expect -jsem to outperform them all?

I agree that this proposal is currently lacking benchmarks that would make it all much more convincing. I'll prepare some, thanks.

cdornan:
I have wanted this for many years &mdash; well done!

OI have one question -- the main issue I have at the moment with parallel GHC builds is the way error messages get interleaved. I have become better at unscrambling errors from different modules being merged arbitrarily but it could be confusing for the unwary. I couldn't work out whether this would be fixed by this proposal &mdash; is there a fix in the works?

simonpj:
I'm happy with the general idea here, but I'm hazy about the specifics.  In particular: **what changes would be needed for GHC**?

I think they are listed under "The GHC jobclient implementation would have the following characteristics:".  I'd prefer it to be more explicitly flagged: "The changes required to GHC are as follows...".

Moreover I think this protocol stuff happens when (*and only when*) GHC is given a `-jsem sem` flag. Correct?  Please say so.

Who creates the semaphore, and how?  What would happen if you give a `-jsem sem` flag but don't create the semaphore?  What happens if cabal/stack are not running.

Basically thumbs up but I'd love to see a bit more clarity.

gridaphobe:
@sheaf Could you update the proposal to address the comments by @angerman and @simonpj?

Apart from that, the committee feedback has been unanimously positive. So once these comments have been addressed we can consider the proposal accepted.

sheaf:
> I think they are listed under "The GHC jobclient implementation would have the following characteristics:". I'd prefer it to be more explicitly flagged: "The changes required to GHC are as follows...".

I added a paragraph at the end of `Implementation of the protocol` that is explicit about the changes required to GHC.

> Moreover I think this protocol stuff happens when (_and only when_) GHC is given a `-jsem sem` flag. Correct? Please say so.

Done.

> Who creates the semaphore?

The jobserver creates the semaphore, as explained in the section "GHC Jobserver Protocol".

> ... and how?

I've added a paragraph at the end of "Implementation of the protocol" that explains we are using the `Win32` and `unix` libraries, with a compatibility package that abstracts over the common functionality.

> What would happen if you give a -jsem sem flag but don't create the semaphore?

This is an error, e.g. "Semaphore does not exist". I've made that explicit in the "Changes to user manual" section.

> What happens if cabal/stack are not running.

GHC doesn't have to know anything about cabal or stack; as the section "GHC jobserver protocol" indicates, it only communicates via a semaphore. So nothing in particular would happen if cabal/stack are not running; all that matters is how many tokens the semaphore has. If nothing else is competing for resources on the semaphore, GHC would be able to acquire all tokens the semaphore has and would release them once it's done... effectively it would be like running with `-jN` where `N` is the number of tokens the semaphore has.

mpickering:
> I have wanted this for many years — well done!
> 
> I have one question -- the main issue I have at the moment with parallel GHC builds is the way error messages get interleaved. I have become better at unscrambling errors from different modules being merged arbitrarily but it could be confusing for the unwary. I couldn't work out whether this would be fixed by this proposal — is there a fix in the works?

If you use `ghc --make` mode then the errors always appear in the same order and not interleaved with other modules. This in itself can be confusing because it can look like the compiler isn't doing anything for a long while before a lot of output suddenly gets printed.

If you are using `cabal` then I think the output from building dependencies in parallel gets interleaved? But that isn't any worse with this patch, it's not something which we have worked on improving yet in any case.

proposals/0540-jsem.rst
```diff
@@ -0,0 +1,453 @@
...
+as it usually only involves switching over to using a system semaphore
+to control the behaviour of ``-j``.
+
+Alternatives
+------------
+
+Multiple home units
+~~~~~~~~~~~~~~~~~~~
+
+Support for [multiple home units](https://well-typed.com/blog/2022/01/multiple-home-units/)
```
PiDelport, line 312: Markdown → RST syntax?

```suggestion
Support for `multiple home units <https://well-typed.com/blog/2022/01/multiple-home-units/>`_
```

gridaphobe:
Thank you for the updates @sheaf!

@simonpj are you satisfied with the revisions to the proposal language? If so, I think we can merge.

sheaf:
> Small point: typesetting bug in the first sentence of 2.3.

Fixed.

> > I think they are listed under "The GHC jobclient implementation would have the following characteristics:". I'd prefer it to be more explicitly flagged: "The changes required to GHC are as follows...".
> 
> I still don't see that. How about this sentence
> 
> "The implementation of the jobclient protocol (in GHC) is more complex, while still remaining very non-invasive. Specifically, the following changes to GHC are needed. When given `-jsem` flag:"

The details about changes to GHC were at the end of that section, after the bullet points. I moved it to before the bullet points and added the phrasing you suggested.

> Please also add (perhaps following the bullets):
> 
> "With no `-jsem` flag, GHC decides by itself how many modules to compile concurrently."

That information is now repeated before the bullet points.

> I asked
> 
> > What would happen if you give a -jsem sem flag but don't create the semaphore? What happens if cabal/stack are not running.
> 
> I don't see an answer to this question. Perhaps GHC just blocks indefinitely waiting for tokens?

I answered all these questions in the "Effects and Interactions" section.

> 
> Another question: what if GHC has two tokens but could use more. Does it block waiting for more, or does it get on with the two, and _in parallel_ ask for more? I assume the latter, although that sounds more complex to implement. Please say.

I added an explanation to the first bullet point in "Implementation of the protocol" that says that resource acquisition is done in parallel, and reminded the reader of this fact in the "Effects and Interactions" section.

simonpj:
Typo: first sentence of 2.3 uses `jobclient` but jobserver, different fonts.

* You say "It does however add a new flag (which interacts with -j".  How does  it interact?  I think that if you say -jsem GHC totally ignores -j. Right?  Please specify in 2.3.
* "If -jsem=<sem> is passed but the semaphore does not exist, GHC should throw an error and exit immediately."  Surely that is part of 2.3? It's part of the specificaition of the changes to GHC, not an effect or an interatction.
* "GHC will acquire as many tokens as it finds available parallelism."  What does that mean?  How does it "find available parallelism".  Perhaps you mean that GHC will greedily ask for as many tokens as the server will give it?  Can you say so?

sheaf:
I updated it to make it more clear what a "unit of work" is:

```
   In the current implementation, a unit of work is the compilation of a single
   module. This means that GHC wants to have as many tokens as it can compile
   modules in parallel. In the future, we could envision GHC being able to use
   more than one token per module, e.g. if one is able to parallelise the
   simplifier workload.
```

simonpj:
> They override eachother. I made that explicit in the "Effects and Interactions" section.

I think you mean:
* If -jsem is not passed to GHC, we fall back to the existing mechanisms for concurrency.
* The -jsem and -j flags override eachother.

Could you put this info in 2.3 please, as an early bullet.  It is part of the *specification*. It's not an effect of the spec.    It also helps to make sense of bullet 3 "Note that GHC already calls setNumCapabilities with the argument to -j."  Except that you might want to say "Note that, with `-j` rather than `-jsem`, GHC calls ..."

One other question.  In `jsem <sem>` what is `<sem>`?  I suppose that it specifies the identity of the semaphore.   Is it an arbitrary string?  Or a number?   Who decides the semaphore name?   Perhpas `cabal`?   What happens if I have two builds going, in different shell windows, both using cabal and hence (I guess) using the same name?

sheaf:
> Could you put this info in 2.3 please, as an early bullet. It is part of the _specification_.

I moved the interactions between flags to the beginning of the "Proposed change specification".

> It also helps to make sense of bullet 3 "Note that GHC already calls setNumCapabilities with the argument to -j." Except that you might want to say "Note that, with `-j` rather than `-jsem`, GHC calls ..."

Done.

> One other question. In `jsem <sem>` what is `<sem>`? I suppose that it specifies the identity of the semaphore. Is it an arbitrary string? Or a number?

It's a string. I've added a parenthetical in "Proposed Change Specification".

> Who decides the semaphore name? Perhaps `cabal`?

The jobserver creates the semaphore and passes it on to jobclients. That was explained in "GHC Jobserver Protocol" but I've made the action of passing the name along more explicit.

>What happens if I have two builds going, in different shell windows, both using cabal and hence (I guess) using the same name?

Different cabal invocations will create different semaphores. If you want e.g. two invocations of cabal to cooperate over a single semaphore, they would need to be jobclients that get spawned by a parent jobserver, as described in "GHC Jobserver Protocol". I added a new bullet point under "Effects and Interactions" that describes this situation, but as of now we don't have any jobservers implemented which would spawn multiple `cabal` subprocesses that coordinate over a semaphore.

nomeata:
> Great. Pull the trigger.

Will do, after @sheaf confirmed that all nits have been fixed.

sheaf:
> > Different cabal invocations will create different semaphores
> 
> ...with different names. As a matter of interest, how does cabal (say) generate a fresh semaphore name, different from any other currently-running cabals? Or does the OS supply the semaphore name from some global pool? Worth saying.

In practice, we tack on a random unique number at the end of the semaphore name, querying the OS to check that the semaphore doesn't already exist (if it does, we try with a new random number). I've added this to the bullet point "Different jobserver invocations create distinct semaphores...".

> These two bullets confusingly overlap. Indeed the second imples the first. How about

OK, I updated that.

sheaf:
> Will do, after @sheaf confirmed that all nits have been fixed.

Yep, I think we're good to go here! Thanks.

PR #541, opened 2022-09-29
WARNING pragmas with categories

adamgundry:
This proposal extends the syntax of ``WARNING`` pragmas to allow them to specify one or more "warning categories".  Users can then enable or disable specific warning categories using command-line flags.  This will make it more reasonable to attach ``WARNING`` pragmas in cases where it is desirable to give users the option to disable the warning individually.

In particular, this is motivated by https://github.com/haskell/core-libraries-committee/issues/87.

[Rendered](https://github.com/adamgundry/ghc-proposals/blob/warning-pragmas/proposals/0000-warning-pragmas-with-categories.rst).

### DIFF
--- a/proposals/0000-warning-pragmas-with-categories.rst
+++ b/proposals/0000-warning-pragmas-with-categories.rst
@@ -1,0 +1,252 @@
+``WARNING`` pragmas with categories
+===================================
+
+.. author:: Adam Gundry
+.. date-accepted::
+.. ticket-url::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/541>`_.
+.. sectnum::
+.. contents::
+
+This proposal extends the syntax of ``WARNING`` pragmas to allow them to specify
+one or more "warning categories".  Users can then enable or disable specific
+warning categories using command-line flags.  This will make it more reasonable
+to attach ``WARNING`` pragmas in cases where it is desirable to give users the
+option to disable the warning individually.
+
+
+Motivation
+----------
+
+The recent `core-libraries proposal #87
+<https://github.com/haskell/core-libraries-committee/issues/87>`_, "Add {-#
+WARNING #-} to Data.List.{head,tail}", has lead to a great deal of debate.  That
+proposal seeks to discourage users from using certain partial ``Prelude``
+functions through the use of ``WARNING`` pragmas, which cause warnings to be
+emitted when the functions are used.  It has been controversial because, while
+there is broad consensus that partial functions should *usually* be avoided,
+there is less consensus that they should *never* be used.
+
+A key problem with ``WARNING`` pragmas as they currently exist is that there is
+no way to disable them selectively. It is possible to disable them entirely,
+with ``-Wno-warnings-deprecations``, but this will suppress all warnings from
+``WARNING`` pragmas, regardless of source.  Thus if the Core Libraries Committee
+decide to add a ``WARNING`` to partial functions, users who wish to continue
+using them will either need to put up with warnings that they consider
+unhelpful, suppress potentially-helpful warnings, or define their own
+non-standard versions of the functions.  This makes the change more
+controversial than it should be, as it potentially inconveniences existing
+users.
+
+Instead, this ghc-proposal extends ``WARNING`` pragmas so they can be annotated
+with one or more "warning categories".  Users can then enable or disable these
+categories individually.  For example, it would allow the core-libraries
+proposal to introduce::
+
+    {-# WARNING [head,partial] head "This is a partial function, it throws an error on empty lists." #-}
+    {-# WARNING [tail,partial] tail "This is a partial function, it throws an error on empty lists." #-}
+
+By default, ``-Wwarnings-deprecations`` is enabled, so the appropriate warning
+would be displayed at any use of ``head`` or ``tail``.  However, users may
+control this behaviour to their liking:
+
+* Users who prefer to use the functions unimpeded may specify
+  ``-Wno-warnings:partial`` to suppress all warnings annotated with the
+  ``partial`` warning category, while still seeing other warnings.
+
+* Users who do not wish to use the functions at all may specify
+  ``-Werror=warnings:partial`` so that any use of the functions will be reported
+  as an error, while other warnings are unaffected.
+
+* Since multiple warning categories can be applied to an identifier, users may
+  distinguish more finely if needed, for example ``-Werror -Wwarn=warnings:head
+  -Wno-warnings:tail`` could be used in a codebase where most warnings are
+  considered errors, except uses of ``head`` are merely warned about, and uses
+  of ``tail`` are not mentioned.
+
+
+Proposed Change Specification
+-----------------------------
+
+The `WARNING and DEPRECATED pragmas
+<https://downloads.haskell.org/ghc/9.4.1/docs/users_guide/exts/pragmas.html#warning-deprecated-pragma>`_
+may be immediately followed by a comma-separated list of warning categories in
+square brackets. Each warning category is an identifier beginning with a letter
+and followed by valid identifier characters or dashes.
+
+More formally, the grammar of declarations is extended as follows:
+
+============  =  =====================================================================
+*decl*        →  ``{-# WARNING`` [*categories*] *things* *strings* ``#-}``
+*decl*        →  ``{-# DEPRECATED`` [*categories*] *things* *strings* ``#-}``
+
+*categories*  →  ``[`` *category1*, ..., *categoryN* ``]``
+*category*    →  (*small* | *large*) { *small* | *large* | *digit* | ``'`` | ``-`` }
+
+*things*      →  *thing1*, ..., *thingN*
+*thing*       →  *varid* | *conid*
+*strings*     →  *string* | ``[`` *string1*, ..., *stringN* ``]``
+============  =  =====================================================================
+
+The list of categories can be omitted entirely, so this subsumes the existing
+syntax for ``WARNING`` and ``DEPRECATED`` pragmas.  That is, [*categories*]
+means that the presence of the *categories* non-terminal is optional; if present
+it is a list in square brackets (hence the use of typewriter font ``[``
+... ``]`` in the definition of the *categories* production).
+
+The *category* non-terminal subsumes both *varid* and *conid*, so it is always
+possible to use the name of the thing to which a warning is being attached as
+one of the categories.  The dash character (``-``) is permitted as a character
+in addition to identifier characters, since dashes are frequently used in
+warning names.
+
+There is no change to the existing rules for when ``WARNING`` or ``DEPRECATED``
+pragmas give rise to warnings, except that individual warning categories may be
+enabled or disabled using new ``-Wwarnings:<category>`` or
+``-Wno-warnings:<category>`` options, and their priority may be controlled using
+the ``-Werror=warnings:<category>`` or ``-Wwarn=warnings:<category>`` options.
+Here ``<category>`` represents the name of a *category* according to the grammar
+above.  The ``-Wwarnings-deprecations`` warning flag (and its synonym
+``-Wdeprecations``) controls the display of all warnings from ``WARNING`` or
+``DEPRECATED`` pragmas, regardless of category.  The command-line flags are
+processed from left to right, with later flags overriding previous ones, as at
+present.
+
+
+Examples
+--------
+
+Suppose the definitions of ``head`` and ``tail`` are annotated with::
+
+    {-# WARNING [head,partial] head "This is a partial function, it throws an error on empty lists." #-}
+    {-# WARNING [tail,partial] tail "This is a partial function, it throws an error on empty lists." #-}
+
+and the user program contains occurrences of both ``head`` and ``tail``::
+
+    module M where
+      foo = head
+      bar = tail
+
+This will result in the following warnings::
+
+    M.hs:2:7: warning: [-Wwarnings:head, -Wwarnings:partial]
+        In the use of ‘head’ (imported from Prelude):
+        "This is a partial function, it throws an error on empty lists."
+      |
+    2 | foo = head
+      |       ^^^^
+
+    M.hs:3:7: warning: [-Wwarnings:tail, -Wwarnings:partial]
+        In the use of ‘tail’ (imported from Prelude):
+        "This is a partial function, it throws an error on empty lists."
+      |
+    3 | bar = tail
+      |       ^^^^
+
+Notice that the message lists all the warning categories that apply.  In current
+version of GHC, this displays ``-Wdeprecations`` alone.
+
+The following examples show the effect of various combinations of warning
+flags:
+
+===============================  ===============================================
+Warning flags                    Result
+===============================  ===============================================
+None                             Warnings displayed
+``-Wno-warnings-deprecations``   No warnings
+``-Wno-warnings:partial``        No warnings
+``-Wno-warnings:tail``           Warning displayed for ``head`` but not ``tail``
+===============================  ===============================================
+
+Warning severity levels may be overridden by subsequent arguments on the
+command-line.  For example, ``-Wno-warnings-deprecations
+-Wwarnings:head -Werror=warnings:tail`` will result in warnings about ``head``
+and errors about ``tail``, but no other ``-Wwarnings-deprecations`` messages.
+On the other hand, ``-Werror=warnings:partial -Wno-warnings-deprecations`` will
+result in no warnings because the second option overrides the first.
+
+
+Effect and Interactions
+-----------------------
+
+This proposal should help resolve the controversy over whether ``head`` and
+``tail`` should be annotated with ``WARNING`` pragmas.  By annotating them with
+categorised warnings, users will be warned about their use by default, but may
+choose to override the warnings as they wish.
+
+These pragmas may be useful for libraries outside ``base`` as well, in
+particular where library authors wish to selectively discourage use of certain
+parts of their API.
+
+
+
+Costs and Drawbacks
+-------------------
+
+This is yet one more feature to implement, although the implementation cost
+should be fairly modest.
+
+Overall this should make the language more accessible to newcomers, as library
+authors will be able to use ``WARNING`` pragmas to discourage certain features
+even if those warnings can be reasonably be disabled in some contexts.
+
+This proposal does not provide a way to disable warnings at specific use sites,
+only at the module level.  In some cases, it would be nice to be able to mark
+individual uses as having been approved and the warning suppressed for that use
+alone, rather than for all uses in the module.
+
+It might be helpful to establish conventions around which categories exist, such
+as ``partial`` for warnings about partial functions.  Moreover it is unclear
+whether every identifier with an attached warning should have a category
+corresponding to the name of the identifier, as with the ``-Wwarnings:head`` and
+``-Wwarnings:tail`` examples.  These issues are currently left to individual
+library authors.
+
+This proposal does not provide a mechanism for organising or namespacing warning
+categories, as they are simply bare identifiers.  Thus if libraries use the
+names of their functions as categories, the names cannot be qualified to
+distinguish definitions from separate modules.
+
+
+Alternatives
+------------
+
+This proposal may be contrasted with `proposal #454
+<https://github.com/ghc-proposals/ghc-proposals/pull/454>`_, which introduces a
+built-in constraint ``Warning`` that can be used for custom warnings along
+similar lines to the existing support for custom type errors.  That proposal
+allows categorisation of warnings in a similar way, and moreover allows
+type-level programming to control the presence and content of warnings, and the
+suppression of warnings at individual use sites.
+
+However, this proposal is simpler, and by keeping the ``WARNING`` annotations as
+separate pragmas rather than requiring them to be part of the types, avoids the
+risk that introducing ``Warning`` constraints may have unexpected effects on
+program semantics.  Assuming this proposal is accepted, it would be fairly
+simple to change the ``Warning`` class to be controlled using the same flags.
+
+`Proposal #454 <https://github.com/ghc-proposals/ghc-proposals/pull/454>`_ does
+not allow multiple categories to be attached to a single warning, but it seems
+useful to include this feature, because it allows for multiple categories at
+different levels of granularity (potentially including a different category for
+every identifier).
+
+
+Unresolved Questions
+--------------------
+
+GHC currently supports both ``WARNING`` and ``DEPRECATED`` pragmas, which are
+synonymous.  Similarly the flags ``-Wwarnings-deprecations`` and
+``-Wdeprecations`` are synonymous, and there is no ``-Wwarnings`` flag.  This
+seems somewhat inconsistent.  Should we reduce the redundancy by considering
+deprecations one particular category of warnings?
+
+Could the use of a colon in the command-line flag name cause issues on some
+platforms?  Is there another syntax that would be preferable?
+
+
+Implementation Plan
+-------------------
+Support with the implementation of this proposal would be welcome.


alt-romes:
Thanks for putting this together and reacting fast to the conversation on the CLC.

I believe this solution promises a good compromise for adding the aforementioned head and tail partiality warnings without necessarily churning those who use `-Werror`.

However, if we don't agree that it is a solution to the warnings for head and tail, would it still be worth it to implement this finer grained warnings? I think Yes, I can imagine this coming up again and other ideas being blocked on -Werror + warnings pragmas being binary.

nomeata:
Thanks for writing this up. I hope it can also provide an alternative solution to the problem addressed by https://github.com/ghc-proposals/ghc-proposals/pull/528#issuecomment-1244557447

proposals/0000-warning-pragmas-with-categories.rst
```diff
@@ -0,0 +1,252 @@
...
+
+The list of categories can be omitted entirely, so this subsumes the existing
+syntax for ``WARNING`` and ``DEPRECATED`` pragmas.  That is, [*categories*]
+means that the presence of the *categories* non-terminal is optional; if present
+it is a list in square brackets (hence the use of typewriter font ``[``
+... ``]`` in the definition of the *categories* production).
+
+The *category* non-terminal subsumes both *varid* and *conid*, so it is always
+possible to use the name of the thing to which a warning is being attached as
+one of the categories.  The dash character (``-``) is permitted as a character
```
nomeata, line 101: Neat idea, but do we really want tools that deal with GHC command lines deal with all that complexity? Unicode emojis etc?

Also, a name can be many things. Can I use fully qualified names? Maybe even package qualified names?

Also, multiple categories is something we don't have yet. Does that mean it matters in which order I pass `-Werror=` or `-Wno-` etc. 

I'd aim for a single category only consisting of lower case letters, dashes, and maybe numbers, and keep it simple.

proposals/0000-warning-pragmas-with-categories.rst
```diff
@@ -0,0 +1,252 @@
...
+
+There is no change to the existing rules for when ``WARNING`` or ``DEPRECATED``
+pragmas give rise to warnings, except that individual warning categories may be
+enabled or disabled using new ``-Wwarnings:<category>`` or
+``-Wno-warnings:<category>`` options, and their priority may be controlled using
+the ``-Werror=warnings:<category>`` or ``-Wwarn=warnings:<category>`` options.
+Here ``<category>`` represents the name of a *category* according to the grammar
+above.  The ``-Wwarnings-deprecations`` warning flag (and its synonym
+``-Wdeprecations``) controls the display of all warnings from ``WARNING`` or
+``DEPRECATED`` pragmas, regardless of category.  The command-line flags are
```
nomeata, line 113: I'd have expected that no category now defaults to “`deprecation`”, and that flag still affects only that category - preserves the existing semantics, without making a flag with deprecation in the name disabling unrelated warnings (that happen to be specified with a related tool)

proposals/0000-warning-pragmas-with-categories.rst
```diff
@@ -0,0 +1,252 @@
...
+
+
+Unresolved Questions
+--------------------
+
+GHC currently supports both ``WARNING`` and ``DEPRECATED`` pragmas, which are
+synonymous.  Similarly the flags ``-Wwarnings-deprecations`` and
+``-Wdeprecations`` are synonymous, and there is no ``-Wwarnings`` flag.  This
+seems somewhat inconsistent.  Should we reduce the redundancy by considering
+deprecations one particular category of warnings?
```
nomeata, line 244: I'd say so

proposals/0000-warning-pragmas-with-categories.rst
```diff
@@ -0,0 +1,252 @@
...
+
+Could the use of a colon in the command-line flag name cause issues on some
+platforms?  Is there another syntax that would be preferable?
```
nomeata, line 247: I considered injecting the custom categories into the space of all categories via prefix `ext-` or `x-`. Concise, but a bit idiosyncratic.

proposals/0000-warning-pragmas-with-categories.rst
nomeata: And `-Wwarnings-deprecations` and `-Wdeprecations` simply become synonyms for `-Wwarnings:deprecation` then.

proposals/0000-warning-pragmas-with-categories.rst
adamgundry: > Neat idea, but do we really want tools that deal with GHC command lines deal with all that complexity? Unicode emojis etc?

Fair question. I don't have much insight into the details of cross-platform CLI handling, so I don't know what an appropriate set of restrictions on characters would be. (Obviously one could permit only ASCII lowercase letters, digits, and dashes, but might that be an annoying for restriction for users in some locales?)

One nice aspect of the present proposal is that every Haskell identifier name is a valid warning category, so library authors can choose to have the category name match the identifier name. That's also why I suggested allowing categories to start with uppercase letters. But I don't hugely care about the exact choices here.

> Also, a name can be many things. Can I use fully qualified names? Maybe even package qualified names?

No, this is mentioned briefly in the Costs and Drawbacks section.

> Also, multiple categories is something we don't have yet. Does that mean it matters in which order I pass -Werror= or -Wno- etc. (Answer below, I know, but I wish I would not have to ask :-))

Perhaps we don't directly have multiple categories, but we do have warning groups (e.g. `-W`, `-Wcompat`, `-Wall`) which amount to much the same thing, just with a restricted list of groups. In particular, we already have order-dependence (e.g. `-Wno-missing-signatures -Wall` is equivalent to `-Wall` and not `-Wall -Wno-missing-signatures`).

I think the "tagging" approach that multiple categories permits is a simple way to get behaviour like the existing warning groups, without needing a predefined taxonomy.

proposals/0000-warning-pragmas-with-categories.rst
adamgundry: Hmm, I quite like this idea. In particular I think we could say:
 * `{- # WARNING ... #-}` or `{- # DEPRECATED ... #-}` without a category list is a synonym for `{-# WARNING [deprecation] ... #-}` ( and perhaps `DEPRECATED` cannot have a category list)
 * `-Wwarnings` (new) enables all warnings from `WARNING` pragmas regardless of category
 * `-Wwarnings-deprecations` and `-Wdeprecations` become synonyms for `-Wwarnings:deprecation`

I'd vaguely worried about backwards compatibility here, but I guess the only potential surprise is that if a dependency starts using a new category, a downstream library that sets `-Wno-warnings-deprecations` or `-Wno-deprecations` will start seeing the warning until it sets -`Wno-warnings`. That seems okay.

proposals/0000-warning-pragmas-with-categories.rst
adamgundry: I'd also previously considered `custom:` but I quite like `x-`. I suggested an explicit `:` or such to make clear the distinction between parts of the flag name that were determined by user code, but perhaps that's unnecessary.

adamgundry:
> I see the high level goal here as: make available to library authors the warning machinery we already have in GHC.
> 
> Hence (as mentioned above), I'd be reluctant to add new features not there (e.g. suddenly a warning has multiple categories) when it's not strictly needed.

I would argue that since we already have groups like `-Wcompat`, this isn't really a new feature. And I think it is a useful one: it avoids the pressure to identify one true taxonomy of warnings! Perhaps a library starts by warning using the `partial` category, then later decides to remove the definition entirely and so wants to add a `compat` or `deprecation` category as well before the eventual removal.

> On the other hand, the existing machinery has the ability for some categories to by errors by default, on by default (`-Wdefault`), on only with `-Wall` or off even with `-Wall` (see https://ghc.gitlab.haskell.org/ghc/doc/users_guide/using-warnings.html). This is not possible with the proposal as it stands, right? (All custom warning categories fall into `-Wdefault`)
> 
> Maybe that's fine. I don't have a great idea to fix this. It is a property of the category, not the warning, so some flag as part of the `WARNING` doesn't cut it. I could imagine a convention around the category _name_ that indicates where it belongs. But that's a bit ad hoc…
> 
> So maybe it's ok if all custom warning categories fall into the “visible with -Wdefault” category. For now.

Yes, this is a good question. It was previously considered in the context of https://github.com/ghc-proposals/ghc-proposals/pull/454 and eventually removed on the basis of not being worth the extra complexity. Thus I was tempted not to worry about it initially.

That said, if we wanted to do this, then assuming we permit multiple categories I think we could have `all`, `default` etc. be category names and say that e.g. `-Wdefault` includes `-Wwarnings:default`? It might even be nice if there was an `error` category and the default behaviour included `-Werror=warnings:error`?

nomeata:
> Perhaps a library starts by warning using the partial category, then later decides to remove the definition entirely and so wants to add a compat or deprecation category as well before the eventual removal.

That would simply be two independent warnings - one in `partial` and one in `deprecation`, would it?

adamgundry:
> > Perhaps a library starts by warning using the partial category, then later decides to remove the definition entirely and so wants to add a compat or deprecation category as well before the eventual removal.
> 
> That would simply be two independent warnings - one in `partial` and one in `deprecation`, would it?

I was expecting we would want:
```
{-# WARNING [partial,deprecation] foo "This identifier is partial, and will be removed in the next version." #-}
```
since then either `-Wno-partial` or `-Wno-deprecation` would disable the (same) warning, whereas if we had
```
{-# WARNING [partial] foo "This identifier is partial, and will be removed in the next version." #-}
{-# WARNING [deprecation] foo "This identifier is partial, and will be removed in the next version." #-}
```
we would presumably get two warnings by default?

nomeata:
> That said, if we wanted to do this, then assuming we permit multiple categories I think we could have all, default etc. be category names and say that e.g. -Wdefault includes -Wwarnings:default? It might even be nice if there was an error category and the default behaviour included -Werror=warnings:error?

That doesn't quite work: I expect a category to belong to a category group. So `partial` is either part of `-Wdefault` or `-Wall`, but not that some partiality warnings are on by default and some only with `-Wall` - that would be a (odd) divergence from the current model. That's why I think the tagging model doesn't quite fit - it assigns data to individual warnings, but not to warning categories.

Assuming we want an open world of categories, without any central registry (that's clear, right), and assuming we want the namespace to be flat (not package-scoped for example) the only way to associate metadata to whole categories (default visibility, group membership, severity) seems to be to encode them in the name… 


Or, we add `WARNING`, `ERROR`, and `NOTICE` pragmas (for -Wdefault, -Werror, -Wall) and indicate that classification in the prefix (`xw-`, `xe-`, `xi-`). Possible, but maybe not worth the bother.

nomeata:
> we would presumably get two warnings by default?

And that is strictly better! Some user who has `-Wno-partial` set since a long time certainly wants to see the new deprecation warning!

proposals/0000-warning-pragmas-with-categories.rst
nomeata: The end user doesn't care too much about where the warning came from originally (built in or custom), so the least intrusive variant is best.

I'd have suggested to not mangle the name at all, allowing library authors to add warnings to existing built-in categories! The only reason I don't is that this would make it hard for GHC to complain about unrecognized warning categories.

proposals/0000-warning-pragmas-with-categories.rst
adamgundry: Strictly speaking we could do without it. But we have `-Wwarnings-deprecations` at the moment, and given all the flexibility to configure warnings it would seem odd if there wasn't a way to say "don't show me any messages from `WARNING` pragmas" without disabling all built-in warnings as well.

nomeata: I'd say users don't care too much if a warning comes from a pragma or some other source (built in? Plugin?), and would design the feature from that point of view. 

But it's not a big deal to add this warning category group, either, if you think it's useful.

adamgundry:
@nomeata okay, you've convinced me to drop multiple categories for simplicity. I've also adjusted the syntax in the light of your comments.

adamgundry:
@nomeata I would like to submit this proposal to the committee. There hasn't been much feedback on this proposal, despite me raising it several times on the CLC proposal thread, but I hope that is because this change is relatively uncontroversial. 

The only substantial discussion has been @nomeata's request to use a single warning category rather than multiple categories, which I've accepted, and some smaller adjustments in response to his feedback.

Since https://github.com/haskell/core-libraries-committee/issues/87 has been accepted, it would be good to try to get this implemented at the same time if possible, to simplify the migration story.

nomeata:
Should we maybe discuss the following point first:

Existing warning categories fall into one of these groups
 * on by default (-Wdefault)
 * off by default, on with -W aka -Wextra
 * off by default, on with -Wall
 * off by default, on with -Wcompat

I also thought we’d have the following group. @goldfirere told me that I am actually wrong, but shares the sentiment that we likely have that eventually
 * on by default, actually an error
(see https://downloads.haskell.org/ghc/latest/docs/users_guide/using-warnings.html)

But for custom warnings, how do we place them in these groups? Note that it is a property of the _category_, not an individual warning (`{-# WARNING #-})`, so it’d be a bit odd to set the category directly with the pragma, because what does it mean if different `WARNING` pragmas use the same category, but place them into a different group?

We could say they are all in `-Wdefault`, but I expect soon library authors want to be able to define off-by-default warnings, compat warnings or even error-by-default warnings.

Maybe we need a separate `{#- WARNING_GROUP partial: extra #-}` that declares that `-Wdefault` turns on all warnings in category `partial`, with names for the above groups?

And some sane thing that combines multiple such pragmas (e.g. things are in `-Wdefault` by default; if multiple pragmas are seen, then prefer `error` over `default` over `compat` over `extra` over `all` maybe?)

Is this sane, or is it too much spooky action at a distance?

tomjaguarpaw:
> Maybe we need a separate `{#- WARNING_GROUP partial: extra #-}`

If we start taking steps like this the warning system will become a small auxiliary language -- and will most likely end up a poorly-designed one given a presumable lack of interest in curating it.

adamgundry:
Ah yes, I tried to cover this in section 6.3 but it's still a bit of an open question.

I feel like separate `WARNING_GROUP` pragmas would be overly complex. My inclination is instead to have the group indicated at the definition site (e.g. `{-# WARNING [partial: extra] head "Blah" #-}` although I'm not hugely enamoured of that syntax). This means that some `partial` warnings may be in `-Wextra` while others are in `-Wdefault`, say, but I don't personally see that as a huge problem.

Another strategy, as you suggested before, would be to have different name prefixes for `partial: extra` vs `partial: default`.

nomeata:
> This means that some partial warnings may be in -Wextra while others are in -Wdefault, say, but I don't personally see that as a huge problem.

that’s precisely what I want to avoid, as I think it violates the principle of least surprise (from a user point of view), because I can no longer understand `-Wextra` as `[-Wwarn-$foo | $foo <- extra_categories ]`. Am I overthinking this?


But we could make progress with this proposal only supporting `-Wdefault` (as `WARNINGS` are now), and solve the severity problem later…

simonpj:
The proposal contains the example
```
{-# WARNING [partial-head] head "This is a partial function, it throws an error on empty lists." #-}
{-# WARNING [partial-tail] tail "This is a partial function, it throws an error on empty lists." #-}
```
Can I suppress both with `-Wno-x-partial-*`?   That is, can I enable or suppress groups of categories?  Could be useful to switch on/off all partial-function warnings, for example.

simonpj:
Reading [the manual](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/pragmas.html#warning-deprecated-pragma) makes me wonder: is there any real difference between WARNING and DEPRECATED?  I think the answer is "no", and the proposal usefully (but a bit quietly) defines DEPRECATE = WARNING [deprecations].  Great.

simonpj:
Suggstion.  Under "Proposed changes" can you enumerate the changes, in a numbered list?  Don't start a numbered item with "There is no change to the existing rules ".  Start with the change. By all means clarify by explaining non-changes as an elaboration.

"A new warning flag -Wextended-warnings controls the display of all warnings from WARNING pragmas, regardless of category. "   Controls it how?  Better to say straightforwardly "The new warning flag -Wextended-warnings switches on warnings from WARNING pragmas regardless of category; -Wno-extended-warnings switches them all off.

simonpj:
Generally I'm in support.

nomeata:
> Can I suppress both with -Wno-x-partial-*? 

No. I expect the real world application to use `partial` for both. The example shows traces of an earlier iteration where a warning could have any number of “tags” (`partial` and `head`) for finer control, but I argued that we should stick to single categories to match what we have in GHC so far. We could make the example a bit less confusing by using two less related categories (`partial` and `experimental` maybe.)

proposals/0000-warning-pragmas-with-categories.rst
```diff
@@ -0,0 +1,290 @@
...
+controversial than it should be, as it potentially inconveniences existing
+users.
+
+Instead, this ghc-proposal extends ``WARNING`` pragmas so they can be annotated
+with a "warning category".  Users can then enable or disable these
+categories individually.  For example, it would allow the core-libraries
+proposal to introduce::
+
+    {-# WARNING [partial] head "This is a partial function, it throws an error on empty lists." #-}
+    {-# WARNING [partial] tail "This is a partial function, it throws an error on empty lists." #-}
```
nomeata, line 49:
```suggestion
    {-# WARNING [x-partial] head "This is a partial function, it throws an error on empty lists." #-}
    {-# WARNING [x-partial] tail "This is a partial function, it throws an error on empty lists." #-}
```

proposals/0000-warning-pragmas-with-categories.rst
```diff
@@ -0,0 +1,290 @@
...
+it is a single category in square brackets (hence the use of typewriter font ``[``
+... ``]`` in the definition of the *categories* production).
+Similarly the list of *things* is optional as it may be omitted (for a ``WARNING`` on a module header).
+
+The *category* non-terminal subsumes both *varid* and *conid*, so it is
+possible to use the name of the thing to which a warning is being attached as
+the category, provided it is not an operator.  The dash character (``-``) is permitted as a character
+in addition to identifier characters, since dashes are frequently used in
+warning names.
+
```
nomeata, line 99:
```suggestion
The *category* non-terminal subsumes both *varid* and *conid*, so it is
possible to use the name of the thing to which a warning is being attached as
the category, provided it is not an operator.  The dash character (``-``) is permitted as a character
in addition to identifier characters, since dashes are frequently used in
warning names.

The compiler only allows a *category* that starts with `x-` or is `deprecations`.
(This restriction is not part of the syntax change because we may expect later
extensions to that list.)


```

proposals/0000-warning-pragmas-with-categories.rst
```diff
@@ -0,0 +1,290 @@
...
+There is no change to the existing rules for when ``WARNING``
+pragmas give rise to warnings, except that individual warning categories may be
+enabled or disabled using new ``-Wx-<category>`` or
+``-Wno-x-<category>`` options, and their priority may be controlled using
+the ``-Werror=x-<category>`` or ``-Wwarn=x-<category>`` options.
+Here ``<category>`` represents the name of a *category* according to the grammar
+above.  The command-line flags are
+processed from left to right, with later flags overriding previous ones, as at
+present.
```
nomeata, line 108:
```suggestion
There is no change to the existing rules for when ``WARNING``
pragmas give rise to warnings, except that individual warning categories may be
enabled or disabled using new ``-W<category>`` or
``-Wno-<category>`` options, and their priority may be controlled using
the ``-Werror=<category>`` or ``-Wwarn=<category>`` options.
Here ``<category>`` represents the name of a *category* according to the grammar
above.  The command-line flags are
processed from left to right, with later flags overriding previous ones, as at
present.
```

proposals/0000-warning-pragmas-with-categories.rst
```diff
@@ -0,0 +1,290 @@
...
+
+This proposal should help resolve the controversy over whether ``head`` and
+``tail`` should be annotated with ``WARNING`` pragmas.  By annotating them with
+categorised warnings, users will be warned about their use by default, but may
+choose to override the warnings as they wish.
+
+This approach also provides an alternative to `proposal #528
+<https://github.com/ghc-proposals/ghc-proposals/pull/528>`_, which is about
+discouraging users from importing from "internal" modules, without completely
+prohibiting their import.  For example, a ``WARNING [ghc-prim-internals]``
```
nomeata, line 187:
```suggestion
prohibiting their import.  For example, a ``WARNING [x-ghc-prim-internals]``
```

proposals/0000-warning-pragmas-with-categories.rst
```diff
@@ -0,0 +1,290 @@
...
+authors will be able to use ``WARNING`` pragmas to discourage certain features
+even if those warnings can be reasonably be disabled in some contexts.
+
+This proposal does not provide a way to disable warnings at specific use sites,
+only at the module level.  In some cases, it would be nice to be able to mark
+individual uses as having been approved and the warning suppressed for that use
+alone, rather than for all uses in the module.
+
+It might be helpful to establish conventions around which categories exist, such
+as ``partial`` for warnings about partial functions.  These issues are currently
```
nomeata, line 214:
```suggestion
as ``x-partial`` for warnings about partial functions.  These issues are currently
```

nomeata (commented):
Based on @aspiwack’s input I am suggesting to use the same string in the pragma and the command line flag.

I also changed the example to use categories that are more different.

Maybe I should have done that before submitting to the committee :-)

adamgundry:
Thanks! I've accepted @nomeata's changes implementing @aspiwack's suggestion that categories in `WARNING` pragmas include the prefix. I've also revised the Proposed Change Specification to start with an enumerated list of changes, as @simonpj requested, and made some other tweaks in response to his feedback (in particular adding a section 6.4 mentioning glob patterns as an alternative we are not currently planning to adopt).

simonpj:
> "A category is recognised if it either starts with x- or is deprecations. GHC will emit an error if a WARNING pragma uses a category that is not recognised."

Why do we insist on an `x-` prefix?   I can't see the motivation for this, and the proposal does not offer one.  It just states the restriction.

nomeata:
> The restriction to recognised categories in point 6 means that WARNING pragmas cannot extend the behaviour of existing GHC warning flags such as -Wtabs or groups such as -Wall. Moreover it means that GHC can still report unrecognised warning flags, rather than silently accepting them. In the future we may extend the set of recognised categories, if it appears useful to let users extend specific built-in warning categories (e.g. -Wcompat) or to allow prefixes other than x-.

proposals/0000-warning-pragmas-with-categories.rst
```diff
@@ -0,0 +1,332 @@
...
+``-Wno-warnings-deprecations`` does at present.  The naming is chosen to allow
+for other sources of "extended warnings" in the future.
+
+The restriction to recognised categories in point 6 means that ``WARNING``
+pragmas cannot extend the behaviour of existing GHC warning flags such as
+``-Wtabs`` or groups such as ``-Wall``.  Moreover it means that GHC can still
+report unrecognised warning flags, rather than silently accepting them.  In the
+future we may extend the set of recognised categories, if it appears useful to
+let users extend specific built-in warning categories (e.g. ``-Wcompat``) or to
+allow prefixes other than ``x-``.
```
nomeata, line 120:
```suggestion
Point 6 makes sure that the command line options `-Wx-partial` and `-Wno-x-partial`
can readily be distinguished from the existing large number of built-in warning
categories, such as `-Wtabs`, `-Wdodgy-import`, `-Winaccessible-code` etc. (See the
`user manual section <https://ghc.gitlab.haskell.org/ghc/doc/users_guide/using-warnings.html?highlight=warning#warnings-and-sanity-checking>`_ for a complete list).
This way GHC can still
report unrecognised warning flags, rather than silently accepting them.
```

proposals/0000-warning-pragmas-with-categories.rst
```diff
@@ -0,0 +1,331 @@
...
+
+
+
+Examples
+--------
+
+Suppose the definitions of ``head`` and ``nub`` are annotated with::
+
+    {-# WARNING [x-partial] head "This is a partial function, it throws an error on empty lists." #-}
+    {-# WARNING [x-quadratic] tail "The nub function has quadratic run-time complexity. If possible, use nubBy or nubOn." #-}
```
Vlix, line 159: Still says `tail`

nomeata: Thanks, fixed on master now.

PR #545, opened 2022-10-27
Deterministic operator parsing proposal

L-as:
[Rendered](https://github.com/L-as/ghc-proposals/blob/deterministic-operator-parsing/proposals/0545-deterministic-operator-parsing.md)

Introduce `DeterministicOperatorParsing` language extension
to determine associativity and precedence from the operator name
instead of from its definition.

This can help heavily simplify the implementation and use of Haskell tooling, e.g. Ormolu, apply-refact, retrie, hlint.

### DIFF
--- a/proposals/0545-deterministic-operator-parsing.md
+++ b/proposals/0545-deterministic-operator-parsing.md
@@ -61,9 +61,13 @@ then the next symbol of the operator is considered.
 ## Examples
 
 `a +| b *| c` is equivalent to `a +| (b *| c)`.
+
 `a +| b -> c` is equivalent to `(a +| b) -> c`.
+
 `a $$$ b $$$ c` is equivalent to `a $$$ (b $$$ c)`.
+
 `a ~~~ b ~~~ c` is invalid.
+
 `a +* b + c` is equivalent to `(a +* b) + c`.
 
 ## Effect and Interactions
--- a/proposals/0545-deterministic-operator-parsing.md
+++ b/proposals/0545-deterministic-operator-parsing.md
@@ -5,7 +5,7 @@ ticket-url: ""
 implemented: ""
 ---
 
-This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/545>).
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/545).
 
 # Deterministic operator parsing
 
--- a/proposals/0545-deterministic-operator-parsing.md
+++ b/proposals/0545-deterministic-operator-parsing.md
@@ -23,7 +23,7 @@ associativities for the custom operators.
 However, this means you can _not_ parse Haskell code without knowing how the operators
 were defined, e.g. `a $$$ b $$$ c` could mean either `(a $$$ b) $$$ c` or `a $$$ (b $$$ c)`.
 
-This means formatterse like Ormolu need to keep a database of all common operators,
+This means formatters like Ormolu need to keep a database of all common operators,
 and support specifying custom operators in the configuration file, rather than Just Working™.
 See e.g. https://github.com/tweag/ormolu/issues/927
 
--- a/proposals/0545-deterministic-operator-parsing.md
+++ b/proposals/0545-deterministic-operator-parsing.md
@@ -25,7 +25,16 @@ were defined, e.g. `a $$$ b $$$ c` could mean either `(a $$$ b) $$$ c` or `a $$$
 
 This means formatters like Ormolu need to keep a database of all common operators,
 and support specifying custom operators in the configuration file, rather than Just Working™.
-See e.g. https://github.com/tweag/ormolu/issues/927
+See e.g. https://github.com/tweag/ormolu/issues/927 (the issue isn't the performance, but rather
+the fact that it's even needed)
+
+In addition, it becomes easier for Haskellers to read Haskell source code.
+Without this extension, you can not know in what way to parse Haskell source
+code without remembering all the fixities involved, which can be a pain for some
+people when reading optics-heavy code.
+
+By making the fixity of the operator a function of its name rather than something to be declared,
+the (possibly robotic) Haskeller has less to keep in mind.
 
 Nim already does a similar thing: https://nim-lang.org/docs/manual.html#syntax-associativity
 
--- a/proposals/0545-deterministic-operator-parsing.md
+++ b/proposals/0545-deterministic-operator-parsing.md
@@ -13,6 +13,9 @@ Introduce `DeterministicOperatorParsing` language extension
 to determine associativity and precedence from the operator name
 instead of from its definition.
 
+This can help heavily simplify the implementation and use of Haskell tooling,
+e.g. Ormolu, apply-refact, retrie, hlint.
+
 ## Motivation
 
 Haskell supports custom operators and allows you to specify custom precedences and
@@ -137,14 +140,14 @@ Symbols from Unicode not in ASCII are subject to the default.
 
 ## Precedence table for custom operators not in `base`
 
-| Symbol    | Precedence (higher is tighter) |
-|:---------:|:------------------------------:|
-| ! .       | 9 |
-| # ^       | 8 |
-| % * /     | 7 |
-| - +       | 6 |
-| : \\      | 5 |
-| < > = ? ~ | 4 |
-| \|        | 2 |
-| & @       | 1 |
-| $         | 0 |
+| Symbol        | Precedence (higher is tighter) |
+|:-------------:|:------------------------------:|
+| ! . otherwise | 9 |
+| # ^           | 8 |
+| % * /         | 7 |
+| - +           | 6 |
+| : \\          | 5 |
+| < > = ? ~     | 4 |
+| \|            | 2 |
+| & @           | 1 |
+| $             | 0 |
--- a/proposals/0545-deterministic-operator-parsing.md
+++ b/proposals/0545-deterministic-operator-parsing.md
@@ -1,0 +1,145 @@
+---
+author: Las Safin
+date-accepted: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/545>).
+
+# Deterministic operator parsing
+
+Introduce `DeterministicOperatorParsing` language extension
+to determine associativity and precedence from the operator name
+instead of from its definition.
+
+## Motivation
+
+Haskell supports custom operators and allows you to specify custom precedences and
+associativities for the custom operators.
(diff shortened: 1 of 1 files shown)


proposals/0545-deterministic-operator-parsing.md
```diff
@@ -0,0 +1,150 @@
+---
+author: Las Safin
+date-accepted: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/545>).
+
+# Deterministic operator parsing
```
int-index, line 10: Parsing operators is already deterministic. You are probably looking for a different word.

L-as: Yes, I am, I was not sure what to call it. What do you think is better?

L-as: It is deterministic in the sense that it doesn't depend on the environment!

int-index: Lexical operator fixity, perhaps, as you determine their precedence and associativity from their lexical form instead of consulting the corresponding fixity declaration.

L-as: That does seem like a better name. Unless someone finds a better name this would probably be the name of the extension.

int-index: > It is deterministic in the sense that it doesn't depend on the environment!

This depends on whether you consider imported modules part of the environment. I see them as part of the program, not the environment. That is, a Haskell program is a _set_ of modules, not a single module, and you need to consider them all together to actually reason about code. Otherwise, you couldn't even tell what type a imported identifier has.

```
import M (x)
```

Would you say that it's non-deterministic whether `x :: Int` or `x :: Bool`? I would surely hope that is deterministic in _my_ builds 😅

treeowl:
Personally, I don't like this idea at all. One possible alternative would be a pragma like this:

```haskell
{-# EXPECT_FIXITIES
infixr 9 .
infixr 5 ++
infixl 4 <$
infixr 9 Cat..
 #-}
```

When GHC saw such a pragma in renaming, it would verify that the fixities match reality. Formatters and such could use the pragmas for parsing without renaming. It would be good to put in whateve may be needed to allow IDEs to populate and maintain the pragmas automatically.

proposals/0545-deterministic-operator-parsing.md
```diff
@@ -0,0 +1,157 @@
...
+any Haskell file can be successfully parsed, formatted, and manipulated,
+if `-XDeterministicOperatorParsing` is enabled.
+
+Formatters would be able to take advantage of this fact if told to use it,
+or if it's enabled via a language pragma.
+
+## Costs and Drawbacks
+
+Adding a new language extension adds complexity. In addition, the
+same code can mean _different_ things depending on whether the extension
```
int-index, line 85: This sounds like major trouble. Your goal is to help tooling, but now tooling needs to deal with modules _with_ and _without_ the extension, producing different results! Let me suggest an alternative.

Your ultimate goal seems to be able to determine the fixity of an operator locally, without looking at other modules. So let's keep fixity declarations, but require specifying them when importing operators. Instead of `import M (foo, bar, (+|))`, ask the user to write:

```
import M (foo, bar, infixl 6 +|)
```

And emit a warning (that could be made into an error with `-Werror`) for any operator occurrence whose fixity is not specified in the import list.

Make an exception for `Prelude`.

int-index: Alternatively, now that I see a similar suggestion from @treeowl, this does not need to be part of the import list.

Just permit duplicate fixity declarations at the top-level:

```
import M (foo, bar, (+|))

infixl 6 +|
```

No need to modify the import list syntax, no need to introduce a pragma. Just list the fixities of the operators used in the current module, and the compiler will check that those fixities agree with the ones specified when the operator was defined.

konsumlamm:
I don't really buy that this addresses the motivation. With this, if you want working formatting for custom operators, you need to enable an extension that potentially changes your operators to behave differently (so that the formatting is potentially incorrect for what you actually want)?

Wouldn't it be a better idea to have the fixed precedence/associativity table on the formatter side? This would have the same advantages and disadvantages afaict, but without introducing a weird extension.

L-as:
> Wouldn't it be a better idea to have the fixed precedence/associativity table on the formatter side? This would have the same advantages and disadvantages afaict, but without introducing a weird extension.

This is what is already happening and is a huge PITA for obvious reasons. See e.g. https://github.com/tweag/ormolu/issues/927
There are a _lot_ of operators, and users may define their own.

L-as:
> Personally, I don't like this idea at all. One possible alternative would be a pragma like this:

This would add a _lot_ of noise to Haskell programs.

phadej:
What about

```haskell
foo `seq` bar
m `div` n
```

These alphanumeric combinators have fixities, as they are often used infix.

L-as:
Ah, I had forgotten about that horrific feature. I think they should always be treated as if they have no fixity declaration.

proposals/0545-deterministic-operator-parsing.md
BinderDavid: > Alternatively, now that I see a similar suggestion [...] this does not need to be part of the import list.

Just to add that I found your first idea much better. Making it part of the import list intuitively conveys the idea that the fixity has to be declared in the module from which it is imported. Just repeating the fixity declaration at the toplevel intuitively conveys that the "ground source of truth" about the fixity of the operator is given at that place.

treeowl: I agree with @BinderDavid, for whatever it's worth.

Melvar:
I would find an `-XLisp` extension more palatable than this.
Defining a new operator would devolve into a choice between a sensible name with a random fixity, and a sensible fixity with a fixity selection sigil hanging off the front of the reasonable name. As presented at this time, the latter isn’t even possible in general since many of the 30 existing fixities do not have one of the 20 listed sigils assigned to them, e.g. `infixr 1` which is the fixity of `Prelude.=<<`.

proposals/0545-deterministic-operator-parsing.md
treeowl: On the other hand: is there anything wrong with allowing an operator to be imported with an intentionally different fixity? Could we mark whether the fixity should merely be checked or actually overridden? That would leave open the question of fixity for re-exported operators.

L-as:
> I would find an `-XLisp` extension more palatable than this. Defining a new operator would devolve into a choice between a sensible name with a random fixity, and a sensible fixity with a fixity selection sigil hanging off the front of the reasonable name. As presented at this time, the latter isn’t even possible in general since many of the 30 existing fixities do not have one of the 20 listed sigils assigned to them, e.g. `infixr 1` which is the fixity of `Prelude.=<<`.

As noted elsewhere, you can fix that specific problem by using the last symbol for associativity, but what do you mean by -XLisp? Uniform fixity?

proposals/0545-deterministic-operator-parsing.md
L-as: Allowing a different fixity would be very confusing.

Melvar:
By `-XLisp` I meant an actual lisp syntax frontend, and I am not a huge fan of lisp syntax.

proposals/0545-deterministic-operator-parsing.md
googleson78: I would say something like `Pragma-free` instead of `Deterministic`

goldfirere: We could call it "pure". A pure function evaluates to the same result in all contexts and program histories. With pure operator fixity, an operator has the same fixity in all contexts and program histories. Just like with functions, purity allows local reasoning.

tomjaguarpaw: But "purity" is typically taken to be a stronger concept than "determinacy", isn't it? That is nothing can be pure unless it is at least deterministic. If that's so then any argument that rejects "deterministic" as a description of this parsing method ought also to reject "pure".

goldfirere:
This is more-or-less exactly what [OCaml does](https://v2.ocaml.org/manual/expr.html#ss:precedence-and-associativity). Regardless of what OCamlers think of it, I think it's the wrong step for Haskell, because of the dissonance with 30 years of existing Haskell code.

However, the motivations behind this proposal are worth pursuing. It would not be hard for, say, GHC to include computed operator fixities in a companion file. Then a tool like Ormolu could use this companion file if available, or its own configuration otherwise. Not sure whether a .hie file is the right place, or plain .hi files. Or, as suggested above, the information could be encoded (and checked) in the source file itself.

L-as:
I've thought about this proposal a lot since submitting this, and some thoughts:

There are two core motivations, and wrt. formatters, tooling, etc., this is all _mostly_ fixable. Changing the language heavily purely for this doesn't seem worth it to me.
For the human motivation, if HLS were usable on my machine (performance issues), I perhaps wouldn't be complaining about this as much, since I could just look at the fixities on the go (I assume).

However, regardless of the above, I honestly no longer think precedence is that useful.
If we dropped all historical context, would we still be adding precedence to operators?
Operators themselves are useful in that it makes it easier to read binary operations, but doing `x + (y * z)` rather than `x + y * z` doesn't seem like a big issue.
In addition (here comes the heretical part), it seems to me that if you want to avoid parentheses, something like `x + $ y * z` could solve that, i.e. `$` becomes syntax. I don't think people would accept a proposal that adds such syntax, but you could add a warning for preventing depending on precedences.

Associativity is another matter, and is harder to avoid without adding the `$` syntax.

JakobBruenker:
> x + $ y * z 

I've wanted something like this at times, but with `-XBlockArguments`, you can write `x + do y * z`.

...Which looks a bit strange here, but can be reasonable in a lot of contexts, if you don't think of `do` as necessarily being monadic.

L-as:
My god! My eyes have been opened. I don't need this proposal anymore.

goldfirere:
> if HLS were usable

I continue to believe that having a widely-used, stable HLS would unlock and simplify much in the process of language design. For example, Coq would be completely and utterly unusable without something like Proof General. If no one had developed an interactive mode alongside the language, it would never have gotten very far. Ditto (but to a lesser extent) for languages like Agda and Idris. If we can get HLS to this point, I think language design would be a far smoother experience. Thanks to the HLS developers for getting us as far as we've gotten!

L-as:
What I really want is a Haskell-native structural editor. An IDE only for Haskell that doesn't operate on text but rather on the AST, and has full awareness of the types of the terms (and kinds of types).

Does such a thing already exist?

googleson78:
I think HLS's "tactics plugin"/wingman is the closest thing that exists currently :thinking: Or agda2hs :sweat_smile:

PR #546, opened 2022-10-31
Make Symbol a newtype over String (under review)

phadej:
[Rendered](https://github.com/phadej/ghc-proposals/blob/symbol-type/proposals/0000-symbol-newtype-string.rst)

I propose changing `data Symbol` definition to be

```haskell
newtype Symbol = MkSymbol String
```

and exporting this fact from `GHC.TypeLits`.

### DIFF
--- a/proposals/0000-symbol-newtype-string.rst
+++ b/proposals/0000-symbol-newtype-string.rst
@@ -85,6 +85,9 @@ Yet, it serves as specific purpose.
 Alternatives
 ------------
 
+One alternative is to keep ``Symbol`` semi-opaque. (Like ``Natural`` is not-completely opaque).
+And export the constructors only from ``GHC.Types`` and ``GHC.Exts`` modules.
+
 An alternative is to make ``String`` promote to ``Symbol``.
 That won't work out well, because lists of characters (i.e. ``String``)
 can be promoted already.
--- a/proposals/0000-symbol-newtype-string.rst
+++ b/proposals/0000-symbol-newtype-string.rst
@@ -1,0 +1,118 @@
+Make Symbol a newtype over String
+=================================
+
+.. author:: Oleg Grenrus
+.. date-accepted::
+.. ticket-url::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/546>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+I propose changing ``data Symbol`` definition
+to be
+
+::
+
+    newtype Symbol = MkSymbol String
+
+and exporting this fact from ``GHC.TypeLits``.
+
+
+Motivation
+----------
+
+The motivation is similar to accepted and implemented
+`Unify Nat and Natural proposal <https://github.com/ghc-proposals/ghc-proposals/pull/364>`_.
+
+Consider the data type with fields of type ``String``::
+
+  data Fields = Fields [String]
+
+``Fields`` is inhabitaed by terms, and since recently (`The Char Kind <https://github.com/ghc-proposals/ghc-proposals/pull/387>`_) also by 
+types. However type literals produce ``Symbol``, not ``String`` ::
+
+  fs = Fields ["foo", "bar"]           -- ok
+  type FS = Fields '[ "foo", "bar" ]   -- not ok
+
+Alternatively, we could declare the type with fields of type ``Symbol``::
+
+  data Fields = Fields [Symbol]
+
+Then it would have the opposite issue::
+
+  fs = Fields ["foo", "bar" ]         -- not ok
+  type FS = Fields '[ "foo", "bar" ]  -- ok
+
+As with ``Nat`` and ``Natural`` proposal, it's possible to go roundabout
+this (and e.g. ``singletons-base`` does). However, if ``Symbol`` were
+an ordinary ``newtype`` over ``String``, we would avoid this issue entirely.
+
+(Note: ``singletons-base`` demotes ``Symbol`` to ``Text`` which is
+incorrect, as ``Text`` cannot represent all ``String`` values)
+
+Proposed Change Specification
+-----------------------------
+
+* Change ``data Symbol`` to ``newtype Symbol = MkSymbol String``
+
+* Re-export full ``Symbol`` definition from ``GHC.TypeLits``.
+
+We don't propose any further changes in ``GHC.TypeLits`` interface.
+For example ``symbolVal`` will continue to return ``String`` value.
+A possible future ``GHC.Symbol`` interface mirroring
+``GHC.TypeLits`` functionlity, but using ``Symbol`` also for the term level
+can be made as CLC proposal, as that would be only a library change.
+
+Technically this is a library only change, but maybe it's only
+a happy coincidence that GHC internals don't need to be changed,
+when ``Symbol`` definition is changed.
+
+Effect and Interactions
+-----------------------
+
+This change is completely backwards-compatible.
+
+Costs and Drawbacks
+-------------------
+
+This proposal introduces "yet another string" type into Haskell ecosystem.
+Yet, it serves as specific purpose.
+
+Alternatives
+------------
+
+An alternative is to make ``String`` promote to ``Symbol``.
+That won't work out well, because lists of characters (i.e. ``String``)
+can be promoted already.
+
+The opposite alternative is to make ``type Symbol = String``.
+It was also meentioned as an alternative in The Char Kind proposal.
+It was rejected then, and the reasoning is still valid:
+First of all, we keep ``Symbol`` for type-checking efficiency.
+Moreover, we would also handle type families inside cons cells when solving
+``HasField`` constraints. For example, ``HasField T ('x' : F y : G z) ty``.
+
+It's possible to use some other type for ``Symbol`` "backend" than ``String``,
+however ``Text`` doesn't qualify as ``String`` (and type-level ``Symbol``)
+can contain values ``Text`` cannot represent. For example single surrogate codepoints.
+``SNat`` and ``KnownSymbol`` are backed by ``String``, so it is a natural
+choice.
+
+Unresolved Questions
+--------------------
+
+None.
+
+Implementation Plan
+-------------------
+
+The patch is very small. I'll write it.
+
+Endorsements
+-------------
+
+None atm.
--- a/proposals/0000-symbol-newtype-string.rst
+++ b/proposals/0000-symbol-newtype-string.rst
@@ -19,9 +19,6 @@ to be
 
     newtype Symbol = MkSymbol String
 
-and exporting this fact from ``GHC.TypeLits``.
-
-
 Motivation
 ----------
 
@@ -59,13 +56,12 @@ Proposed Change Specification
 
 * Change ``data Symbol`` to ``newtype Symbol = MkSymbol String``
 
-* Re-export full ``Symbol`` definition from ``GHC.TypeLits``.
-
-We don't propose any further changes in ``GHC.TypeLits`` interface.
-For example ``symbolVal`` will continue to return ``String`` value.
-A possible future ``GHC.Symbol`` interface mirroring
-``GHC.TypeLits`` functionlity, but using ``Symbol`` also for the term level
-can be made as CLC proposal, as that would be only a library change.
+In particular, we don't propose any changes to ``GHC.TypeLits`` interface.
+Firstly the implementation of ``Symbol`` stays internal.
+Also ``symbolVal`` will continue to return ``String`` value.
+In future the ``GHC.Symbol`` interface mirroring ``GHC.TypeLits`` functionlity
+could be vreated , but using ``Symbol`` also for the term level can be made as
+CLC proposal, as that would be only a library change.
 
 Technically this is a library only change, but maybe it's only
 a happy coincidence that GHC internals don't need to be changed,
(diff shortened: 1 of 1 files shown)


adamgundry:
Doing something like this seems a very good idea. 

As an alternative to exposing the implementation of `Symbol` as a newtype around `String`, we could keep the representation abstract but expose an interface for conversions?
```hs
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE TypeFamilies #-}

module GHC.Symbol (Symbol, toString, ToString, FromString) where

  import Data.String

  newtype Symbol = FromString String
    deriving IsString -- or just a fromString function?

  toString :: Symbol -> String
  toString (FromString s) = s

  type ToString :: Symbol -> String
  type family ToString s where
    ToString ('FromString s) = s

  type FromString :: String -> Symbol
  type family FromString s where
    FromString s = 'FromString s

```

I suppose a downside is that type family reduction will somewhat reveal the presence of the internal `FromString` constructor; punning with the type family may or may not be a good idea. And for `coerce` it might sometimes be useful to get hold of the constructor. And perhaps it is futile to plan for a future in which `String` is not the internal representation type...

Thus we could just as well have an internal/unstable module that exposes the newtype constructor? So perhaps for the present proposal, rather than modifying `GHC.TypeLits` (which is, de facto, relatively widely imported) the newtype could live in a new module `GHC.Symbol.Internal` or suchlike? Then there can be a CLC discussion to settle the design of `GHC.Symbol`, any new exports for `GHC.TypeLits`, and which instances to provide for `Symbol`.

int-index:
It still isn't clear to me what exactly goes wrong if we say `type Symbol = String`.

The example you cite is this:

```haskell
HasField T ('x' : F y : G z) ty
```

But then what stops the user from writing this?

```haskell
HasField T (MkSymbol ('x' : F y : G z)) ty
```

It seems to me that as soon as we expose some `C` such that `C :: String -> Symbol`, then we're paying the full cost of supporting a `String` in all positions where `Symbol` is currently supported, so we might as well cut the act and admit that `String ~ Symbol`.

phadej:
> But then what stops the user from writing this?

Nothing interesting I guess. That `MkSymbol ...` won't have `KnownSymbol` solved for it.

An alternative is what @adamgundry suggests, make the `Symbol` opaque. I'll add that as an alternative to the proposal.

phadej:
@adamgundry 

`ToString` and `FromString` can be written already today using `ConsSymbol` and `UnconsSymbol`. And IIRC the Char Kind proposal discussed this issue already. (Having less magical type families is a good thing).

phadej:
> so we might as well cut the act and admit that `String ~ Symbol`

The performance argument kicks in. I think we simply cannot afford that without very large rework of how GHC works. Just the size/cost of `Generic` instances for even small records encoding metadata with cons constructors for each character in type/constructor/field names (with type-applications) in them...

int-index:
> Nothing interesting I guess. That `MkSymbol ...` won't have `KnownSymbol` solved for it.

Oh, under your proposal, does `MkSymbol ('h' : 'e' : 'l' : 'l' : 'o' : '[]) ~ "hello"` hold?

adamgundry:
> `ToString` and `FromString` can be written already today using `ConsSymbol` and `UnconsSymbol`. And IIRC the Char Kind proposal discussed this issue already. (Having less magical type families is a good thing).

I wasn't proposing they would be magical type families, rather they would merely add and remove the newtype constructor (so that it can remain opaque).

phadej:
> > `ToString` and `FromString` can be written already today using `ConsSymbol` and `UnconsSymbol`. And IIRC the Char Kind proposal discussed this issue already. (Having less magical type families is a good thing).
> 
> I wasn't proposing they would be magical type families, rather they would merely add and remove the newtype constructor (so that it can remain opaque).

Their magic property comes from `Symbol` literals being magic.

```haskell
ToString "foobar"
```

won't compute, unless GHC learns to do it specifically. Thus I'd rather not add them. Conversion can be done with `UnconsSymbol` and `ConsSymbol`, these GHC knows how to reduce.

adamgundry:
Ah I see. But then @int-index's question in https://github.com/ghc-proposals/ghc-proposals/pull/546#issuecomment-1297685795 is indeed interesting.

phadej:
@adamgundry, @int-index

> Oh, under your proposal, does MkSymbol ('h' : 'e' : 'l' : 'l' : 'o' : '[]) ~ "hello" hold?

It doesn't.

I think that is not big issue. I was trying to show that there is a similar problem with `Natural` already:

```haskell
{-# LANGUAGE DataKinds, TypeFamilies #-}

import GHC.Num.Natural

type family Con (n :: Natural) :: Bool where
  Con (NS _) = True
  Con (NB _) = False
```

And if that worked, then `Con 42` wouldn't reduce.
That however doesn't compile (yet?!) as GHC rejects that definition with

```
N.hs:7:11: error:
    • Couldn't match a lifted type with an unlifted type
      When matching kinds
        k0 :: *
        GHC.Prim.Word# :: TYPE 'GHC.Types.WordRep
    • In the first argument of ‘NS’, namely ‘_’
      In the first argument of ‘Con’, namely ‘(NS _)’
      In the type family declaration for ‘Con’
  |
7 |   Con (NS _) = True
  |           ^
```

If such restriction is desired, than `Symbol` can be defined as:

```haskell
{-# LANGUAGE DataKinds, TypeFamilies #-}
{-# LANGUAGE MagicHash, UnboxedTuples #-}

data Symbol = MkSymbol (# String #)
```

Then we won't be able to promote `MkSymbol`. Or rather we, can, but we won't be able to do anything with it in types, as with cannot with `NS` or `NB`:

```
*Main GHC.Exts> :kind! MkSymbol
MkSymbol :: (# String #) -> Main.Symbol
= 'MkSymbol
*Main GHC.Exts> :m +GHC.Num.Natural 
*Main GHC.Exts GHC.Num.Natural> :kind! NS
NS :: Word# -> Natural
= 'NS
```

The Unboxed tuple indirection is ugly, but *does the job*, doesn't it?

Ericson2314:
> It doesn't.

I am confused. How is that prevented? Your proposal as written has `String` not `(# String #)` at this time, does it not?

Ericson2314:
IMO, I fear we are indefinitely locking ourselves into a design we may not want long term based on short-term implementation considerations. This isn't just me preferring `String ~ Symbol`, it is also me questioning what the future of `String` is.

It seems like the prudent thing to do is make sure we can "demote" `Symbol`, but make it demote to an abstract type which can be converted to and from a `String`. The conversations would be done with "opaque" functions and "opaque" type families to make sure in neither case dos the internal structure of `Symbol` leak.

The above design should unblock any short-term needs while keeping the design space maximally wide open going forward.

phadej:
> I am confused. How is that prevented?

In the current version of proposal it's not prevented. It just doesn't hold. You can write that expression, but it won't hold. I leave it to the committee to decide on the alternatives.

Ericson2314:
Reading the proposal I have no idea how one could tell that doesn't hold? Does this follow from some tricky bit of `DataKinds` I am ignorant of? Or is it a special case you intentionally made?

adamgundry:
If I understand correctly, the point is that the proposal introduces `MkSymbol` as a constructor for `Symbol` (whereas previously it had no constructors), but doesn't specify that symbol literals relate to it. Currently GHC models `Symbol` as an empty datatype at the term level (indeed, it accepts `case x :: Symbol of {}` as being total!) so this doesn't cause any fundamental problems, but it is a significant divergence between the term and type levels. It's rather like symbol literals are non-standard `Symbol` values.

Presumably not merely does `MkSymbol ('h' : 'e' : 'l' : 'l' : 'o' : '[]) ~ "hello"` fail to hold, but in fact `MkSymbol _` is apart from every symbol literal? That is, I can write the following and have it reduce for `Dodgy ""` and `Dodgy (MkSymbol '[])`?
```hs
  type family Dodgy (s :: Symbol) :: Type where
    Dodgy "" = Int
    Dodgy (MkSymbol '[]) = Bool
```

I agree with @Ericson2314 in https://github.com/ghc-proposals/ghc-proposals/pull/546#issuecomment-1298348623 that we should make `Symbol` an abstract type with both term-level functions and type families for converting to and from `String`. This could be along the lines of the design I sketched in https://github.com/ghc-proposals/ghc-proposals/pull/546#issuecomment-1297589946, but with the type families implemented using `ConsSymbol`/`UnconsSymbol` (or left to be implemented elsewhere).

@phadej observed that the same issue exists for `Natural` in https://github.com/ghc-proposals/ghc-proposals/pull/546#issuecomment-1298319948. This doesn't seem to have been considered in #364. I'm not convinced we should rely on the lack of ability to promote unlifted types as a workaround, and I think we should regard the current issue with `Natural` as a problem to be solved rather than an example to be imitated.

phadej:
@adamgundry So there are two options if I understand you right:

1. Define type and conversion functions in `GHC.Types`, 

```haskell
module GHC.Types (...
   Symbol,
   symbolToString,
   symbolFromString,
) where

data Symbol = ...
```

2. Define only type in GHC.Types but export the constuctor:

```haskell
module GHC.Types (...
  Symbol (..),
) where

newtype Symbol = MkSymbol String
```

I prefer the latter, because there is less library API to pick colors for, which I'd rather discuss in separate (CLC) proposal. And as you say, the fact `Natural` has constructors which (almost) can be promoted is an issue too, to be solved later.

adamgundry:
Right, I think the main thing is that the API exposed to users (via `GHC.TypeLits` or `GHC.Symbol` or wherever in `base`) should be abstract. I'm not particularly worried if the newtype constructor is visible via a `ghc-prim` import such as `GHC.Types`, provided it is clearly an internal and unstable implementation detail of GHC. This in theory what `ghc-prim` is for, although that could perhaps be advertised a bit more loudly.

Ericson2314:
I agree that the most important thing is that the internal representation stays unstable.

I also think it's waste of effort / misleading to expose (unstably) `MkSymbol` as the sole data constructor when every literal is apart from `MkSymbol _` as @adamgundry says — what do we gain from lying? What happens when we can write promoted `case "foo" of MkSymbol s -> ...`? — but this is merely about avoiding silliness.

Only stabilizing something we'll inevitably want to change later is catastrophic, and thus my first priority to avoid.

phadej:
I changed the proposal to keep `Symbol` implementation (semi) opaque.

@nomeata  I would like to submit this proposal to the committee.

Ericson2314:
Are there term level `String` <-> `Symbol` conversion functions? Don't we need those to achieve anything user-visible with this?

phadej:
> Are there term level String <-> Symbol conversion functions? Don't we need those to achieve anything user-visible with this?

There are, yet we don't need them to this to be useful. Please re-read the motivation section: this proposal is about making `Symbol` usable on *the term level*.

nomeata:
Thanks, assigned to @cdornan

proposals/0000-symbol-newtype-string.rst
```diff
@@ -0,0 +1,128 @@
...
+can contain values ``Text`` cannot represent. For example single surrogate codepoints.
+``SNat`` and ``KnownSymbol`` are backed by ``String``, so it is a natural
+choice.
+
+Unresolved Questions
+--------------------
+
+None.
+
+Implementation Plan
```
Ericson2314, line 120: If `newtype Symbol = MkSymbol String` is to be mentioned it, it should only be in this section.

Ericson2314:
I would like to also remind people of what @simonpj brought up in https://github.com/haskell/core-libraries-committee/issues/105. If we are to have a proper stable vs unstable boundary, we have to take things like writing abstract types seriously.

This ["(semi) opaque implementation"](https://github.com/ghc-proposals/ghc-proposals/pull/546#issuecomment-1319028028) winking and nodding is highly confusing and leaves users and implementers alike unsatisfied because the stability guarantees are not clear.

re-xyr:
The current propoal seems confusing to me, as in motivation it says the problem is we cannot write

```hs
data Fields = Fields [Symbol]
fs = Fields ["foo", "bar" ]         -- not ok
type FS = Fields '[ "foo", "bar" ]  -- ok
```

nor

```hs
data Fields = Fields [String]
fs = Fields ["foo", "bar" ]         -- ok
type FS = Fields '[ "foo", "bar" ]  -- not ok
```

But then it goes on to specify the proposed change that is

> - Change `data Symbol` to `newtype Symbol = MkSymbol String`
> 
> In particular, we don't propose any changes to `GHC.TypeLits` interface. *[...]*

But this doesn't do anything to solve the problem mentioned! Type-level double quotes still gives you `Symbol` while term-level still gives you `String` which are two apart types. 

Is the goal here just to allow `Symbol` to live in term level? Then shouldn't there be functions like

```hs
toSymbol :: String -> Symbol
fromSymbol :: Symbol -> String
```

specified in the proposal, instead of an *unexported* newtype definition?

phadej:
@re-xyr 

> Then shouldn't there be functions like ...

These are library concerns, and as `base` library and `GHC.TypeLits` are supervised by CLC, the ghc-proposals is not right place to discuss them. How the library part would look like is left out of this proposal. This proposal only makes ground so implementing a term-level `Symbol` library is possible.

I have previously asked how to coordinate proposals which cut through both GHC Steering Committee and CLC "jurisdictions" and got no answer, so I avoid making such.

Ericson2314:
This would not be satisfactory if there are very interesting changes on both sides, but in this case I think the language changes are boring/non-existent and merely exist to service the library changes, so we can cut the loop and unroll this way way reasonably.

phadej:
As said I said, I don't have time to sketch out the library part of this proposal.

Given there is no shepherd response either, I think I won't bother.

nomeata:
I’m sorry that we did not manage to handle your request in time. As per the latest [committee status](https://mail.haskell.org/pipermail/ghc-steering-committee/2022-November/002984.html), the backlog has in general grown over the last months, and we are generally behind. This could be improved with more and more active committee members, so I invite you to keep an eye open for the next round of nominations and join us to make the process less frustrating for proposal authors!

cdornan:
@phadej I am sorry for the delay &mdash; I was trying to close out another propsal then give this my full attention but that underwent some late complications.

I would be happy to try and make this process work for you &mdash; please let me know, otherwise I will try and find another sponsor as the proposal seemed worthy.

Ericson2314:
@phadej Did you have an implementation PR for this? That would help me write the CLC proposal.

cdornan:
@phadej @nomeata @adamgundry @Ericson2314 @int-index @re-xyr I have resubmitted this proposal myself as #562 to see if it can be rescued.

PR #116, opened 2018-03-26
Unboxed coercions

treeowl:
Offer access to unboxed coercions.

[Rendered](https://github.com/treeowl/ghc-proposals/blob/unboxed-coercions/proposals/0000-unboxed-coercions.rst)

PR #547, opened 2022-11-03
Remove kind constraints (under review)

goldfirere:
@simonpj and I have come to the conclusion that allowing constraints in kinds has outlived its usefulness. Constraints in kinds pose implementation challenges, are only minimally useful, and have a somewhat broken implementation. Let's just get rid of them.

[Rendered](https://github.com/goldfirere/ghc-proposals/blob/no-kind-equalities/proposals/0000-no-kind-equalities.rst)

### DIFF
--- a/proposals/0000-no-kind-equalities.rst
+++ b/proposals/0000-no-kind-equalities.rst
@@ -1,0 +1,138 @@
+Disallow kind equalities on type declarations
+=============================================
+
+.. author:: Richard Eisenberg and Simon Peyton Jones
+.. date-accepted::
+.. ticket-url::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+.. sectnum::
+.. contents::
+
+GHC currently allows types like ``T`` here::
+
+  type family F a where
+    F Int = False
+    F _   = True
+
+  type T :: forall a -> (F a ~ True) => Type
+  data T a = MkT a a
+
+Note that ``T``\'s kind includes an equality. This declaration means that we
+disallow ``T Int`` while allowing ``T`` to be instantiated at any other type
+(of kind ``Type``).
+
+This proposal seeks to remove this syntax, as it barely works, is hard to maintain,
+and is not so useful.
+
+Motivation
+----------
+
+A kind equality like the one in the introduction is very much like a datatype
+context: the equality is required to be satisfied in order to work with the type,
+the equality proof itself is never used. Put another way, an equality like this
+only restricts the programs you can write, never allowing new programs to be
+written.
+
+While it might be nice to eagerly forbid a type like ``T Int`` from being written,
+the benefit is slight (but real).
+
+On the other hand, the challenge in continuing to accept this syntax is real,
+and currently biting. For example, the declaration in the introduction is accepted,
+but the following is currently rejected::
+
+  type T :: forall a -> (F a ~ True) => Type
+  data T a where
+    MkT :: a -> a -> T a
+
+The reason this is rejected is that it tries to kind-check ``T a``. But because we
+don't know that ``F a ~ True``, the application ``T a`` is rejected. The "obvious"
+remedy (``MkT :: forall a -> F a ~ True => a -> a -> T a``) isn't currently available,
+as GADT syntax doesn't support visible dependent quantification (yet). Even if the
+syntax were available, it wouldn't work, because `GHC can't use an equality constraint <https://gitlab.haskell.org/ghc/ghc/-/issues/15710>`_
+in the same type signature it is introduced. We can imagine machinations that would
+accept ``T`` here, but it seems hard in the general case, most likely requiring a new
+algorithm to implement constraint generalization in kinds, a feature that is not otherwise
+needed.
+
+Even without trying to accept GADT-syntax definitions, it is challenging to
+maintain this feature. A concrete example is that Simon PJ and I (Richard)
+had to spend maybe a cumulative 8-10 hours of meetings and calls trying to
+figure out how to keep this feature working with the `work to separate
+Constraint and Type <https://gitlab.haskell.org/ghc/ghc/-/merge_requests/8750>`_.
+Instead of continuing down that path, we simply want to jettison the
+feature.
+
+It might also be interesting to note the history of this feature: I (Richard)
+added it as I was implementing ``-XTypeInType`` because it seemed easy to do
+so. No one asked for it. It just seemed like an easy addition. However, it has
+proved anything but easy. (This is not the first time we've been stymied supporting it.)
+So let's just send it back to the void whence it came.
+
+Proposed Change Specification
+-----------------------------
+
+Disallow the constraint arrow ``=>`` in the following contexts:
+
+* A kind annotation within a type: ``f ::  Int -> (Bool :: <<here>>) -> Double``
+* A visible kind application within a type: ``f :: Proxy @(<<here>>) True -> ...``
+* A kind annotation on a type variable binder: ``f :: forall (a :: <<here>>). a -> a``
+* A kind annotation on a ``data`` declaration: ``data T a b :: <<here>> where ...``
+* A kind annotation on a family declaration: ``type family F a b :: <<here>>`` (or ``data family``)
+* A standalone kind signature: ``type T :: <<here>>``
+
+In addition, any data constructor whose type includes a constraint is
+not available for use in types.
+
+Examples
+--------
+In addition to the example in the introduction, which would now be rejected,
+this proposal stops the use of any constrained data constructor in a type.
+For example::
+
+  data S a where
+    MkS :: F a ~ Int => a -> a -> S a
+  -- definition of S is OK
+
+  p :: Proxy (MkS True False)  -- this is rejected
+  p = Proxy
+
+Effect and Interactions
+-----------------------
+
+* Much code can be deleted. A partial list is on the `ticket tracking the implemenration of this proposal <https://gitlab.haskell.org/ghc/ghc/-/issues/22298#checklist-of-things-to-remove>`_.
+
+* The inconsistency in the Motivation_ section is gone.
+
+Costs and Drawbacks
+-------------------
+
+* Putting a constraint on a kind is very modestly useful, and this proposal destroys that use case. Instead, users are welcome to constrain data constructors instead of the type itself.
+
+* If and when we get full dependent types -- including the ability to have type class constraints in kinds (which is not permitted today) -- we may need to reintroduce
+this feature, thus making the work of removing it redundant. However, it is quite likely that the new implementation will be much more well-grounded than the
+current one. It is even possible that it will be easier to implement this feature correctly after the current incorrect implementation is ripped out.
+
+Alternatives
+------------
+
+* We don't have to do this. We could live with the inconsistency in the Motivation_, and
+Simon and I have indeed worked out how to land ``Constraint``\ -vs-\ ``Type`` even with
+this feature. But this is a poor plan, in my opinion.
+
+* We could fully implement the feature. However, this is quite a lot of work with very
+little payoff.
+
+Unresolved Questions
+--------------------
+None at this time.
+
+Implementation Plan
+-------------------
+I (Richard) will implement.
+
+Endorsements
+-------------
+The ideas behind this proposal were developed in concert with Simon PJ, though
+I wrote up this text without his review.
--- a/proposals/0000-no-kind-equalities.rst
+++ b/proposals/0000-no-kind-equalities.rst
@@ -127,8 +127,8 @@ Costs and Drawbacks
 * Putting a constraint on a kind is very modestly useful, and this proposal destroys that use case. Instead, users are welcome to constrain data constructors instead of the type itself.
 
 * If and when we get full dependent types -- including the ability to have type class constraints in kinds (which is not permitted today) -- we may need to reintroduce
-this feature, thus making the work of removing it redundant. However, it is quite likely that the new implementation will be much more well-grounded than the
-current one. It is even possible that it will be easier to implement this feature correctly after the current incorrect implementation is ripped out.
+  this feature, thus making the work of removing it redundant. However, it is quite likely that the new implementation will be much more well-grounded than the
+  current one. It is even possible that it will be easier to implement this feature correctly after the current incorrect implementation is ripped out.
 
 Alternatives
 ------------
--- a/proposals/0000-no-kind-equalities.rst
+++ b/proposals/0000-no-kind-equalities.rst
@@ -6,7 +6,7 @@ Disallow constraints in kinds
 .. ticket-url::
 .. implemented::
 .. highlight:: haskell
-.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/547>`_.
 .. sectnum::
 .. contents::
 
--- a/proposals/0000-no-kind-equalities.rst
+++ b/proposals/0000-no-kind-equalities.rst
@@ -70,6 +70,10 @@ so. No one asked for it. It just seemed like an easy addition. However, it has
 proved anything but easy. (This is not the first time we've been stymied supporting it.)
 So let's just send it back to the void whence it came.
 
+Ryan Scott's `blog post <https://ryanglscott.github.io/2021/08/01/equality-constraints-in-kinds/>`_
+provides additional background and gotchas around the existing feature. He even has to
+say "They're not totally useless" to defend the feature.
+
 Proposed Change Specification
 -----------------------------
 
@@ -100,6 +104,14 @@ For example::
   -- definition of S is OK
 
   p :: Proxy (MkS True False)  -- this is rejected
+  p = Proxy
+
+This proposal does *not* affect normal GADT constructors. So this still works::
+
+  data T a where
+    MkT :: Bool -> T Bool
+
+  p :: Proxy (MkT True)
   p = Proxy
 
 Effect and Interactions
--- a/proposals/0000-no-kind-equalities.rst
+++ b/proposals/0000-no-kind-equalities.rst
@@ -1,5 +1,5 @@
-Disallow kind equalities on type declarations
-=============================================
+Disallow constraints in kinds
+=============================
 
 .. author:: Richard Eisenberg and Simon Peyton Jones
 .. date-accepted::
--- a/proposals/0000-no-kind-equalities.rst
+++ b/proposals/0000-no-kind-equalities.rst
@@ -85,6 +85,10 @@ Disallow the constraint arrow ``=>`` in the following contexts:
 In addition, any data constructor whose type includes a constraint is
 not available for use in types.
 
+Note that class constraints are already forbidden in exactly these places,
+so this proposal changes only the status of equality constraints (both
+homogeneous, with ``~``, and heterogeneous, with ``~~``).
+
 Examples
 --------
 In addition to the example in the introduction, which would now be rejected,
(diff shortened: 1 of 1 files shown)


treeowl:
I assume we'll still be able to have kind constraints in types?

```haskell
flippedy :: forall {k} (a :: k) j. j ~ k => ....
```

On a more serious note, it is unfortunate that this *looks* like a step backwards for promotion, but it sounds like you think that'll likely require some other approach anyway.

proposals/0000-no-kind-equalities.rst
```diff
@@ -0,0 +1,138 @@
...
+Disallow the constraint arrow ``=>`` in the following contexts:
+
+* A kind annotation within a type: ``f ::  Int -> (Bool :: <<here>>) -> Double``
+* A visible kind application within a type: ``f :: Proxy @(<<here>>) True -> ...``
+* A kind annotation on a type variable binder: ``f :: forall (a :: <<here>>). a -> a``
+* A kind annotation on a ``data`` declaration: ``data T a b :: <<here>> where ...``
+* A kind annotation on a family declaration: ``type family F a b :: <<here>>`` (or ``data family``)
+* A standalone kind signature: ``type T :: <<here>>``
+
+In addition, any data constructor whose type includes a constraint is
```
nomeata, line 85: Does that include GADT constructors (whose signature don't look like they have constraints, but are to the best of my knowledge desugared to constructors with equality constraint parameters)?

simonpj: > Does that include GADT constructors (whose signature don't look like they have constraints, but are to the best of my knowledge desugared to constructors with equality constraint parameters)?

No, it doesn't include them.  It remains fine to have
```
data T a where
  T1 :: Int -> T Int
  ...
```
and (with `-XDataKinds`) use `T1` in a type, where it has kind, `Int -> T Int` (unsurprisingly).   But there is no equivalent, at the type level, of the GADT-aware pattern matching that we get with GADTs in terms.

simonpj:
I very strongly support this proposal.  It's a mis-feature, rather like contexts in data types
```
data Ord a => T a = T1 a | T2
```
which we removed from the language some time ago.  

It's an ill-specified feature that has never worked properly.  It is inconsistent in that it apperas to promise something (modest) but turns out not to deliver even on that modest promise.  It carrries seriously painful implementation consequences, in exchange for (essentialy) no user benefit. 

Let's just kill it.  If we want constraints in kinds we can add them later, only properly.

proposals/0000-no-kind-equalities.rst
nomeata: Thanks! So
```
data T a where
  T1 :: Int -> T Int
```
and
```
data T' a where
  T1' :: (Int ~ a) => Int -> T a
```
will behave different.

simonpj: Correct.  The latter can't be promoted; the former can.

We could retrench still further and prevent you promoting a GADT.  The ability to do so is not (in my understanding) particularly useful but it comes at close-to-zero cost. 

If anyone has a view here, do express it.

proposals/0000-no-kind-equalities.rst
```diff
@@ -0,0 +1,138 @@
...
+
+Alternatives
+------------
+
+* We don't have to do this. We could live with the inconsistency in the Motivation_, and
+Simon and I have indeed worked out how to land ``Constraint``\ -vs-\ ``Type`` even with
+this feature. But this is a poor plan, in my opinion.
+
+* We could fully implement the feature. However, this is quite a lot of work with very
+little payoff.
```
adamgundry, line 125: The formatting isn't quite right here.

proposals/0000-no-kind-equalities.rst
nomeata: I don't have a view, just questions:-)

(My personal mental model of GADTs is based only on the desugaring to the one with constraints, so it feels odd that it makes a difference, but the oddity may likely rest purely in my understanding here. And I have not used promoted GADTs yet, I think.)

adamgundry: The text could make clearer that the *change* here is to forbid promotion of constructors whose constraints contain only equalities (since constraints containing classes are already unpromoteable).

Presumably the rationale for this restriction is that following this proposal it would be impossible to write the kind of a promoted constructor with an equality constraint?

How feasible would it be to continue to support promotion of such constructors anyway? I guess kind checking would need to solve wanted equality constraints arising from calls to constrained constructors, but since there would be no way to introduce given constraints, that might not be too hard? But perhaps it isn't even desirable in the absence of a way to use the evidence in (type-level) pattern matching.

simonpj:
> How feasible would it be to continue to support promotion of such constructors anyway? 

The entire drive of the proposal is to outlaw kinds with a "=>" in them.  That is what gives rise to all the complications.  The current design generally outlaws such things, with the weird and inconsistent exception of `(t1 ~ t2)`.    

TL;DR: infeasible.   That's precisely what we want to outlaw.

proposals/0000-no-kind-equalities.rst
goldfirere: In truth, explicit equality constraints are already slightly different than GADT refinements: `MkG :: (a ~ Int) => a -> G a` has an extra field to store the equality at runtime, while `MkG :: Int -> G Int` does not. But this would make the difference more apparent, indeed.

> We could retrench still further and prevent you promoting a GADT. The ability to do so is not (in my understanding) particularly useful but it comes at close-to-zero cost.

No no no! This is one key part of `-XTypeInType`! I do not propose stopping GADT constructors from being used in types.

@adamgundry agreed about the clarification. On its way.

The rationale for the restriction is that the kind-checker doesn't need to know, at all, about constraints after this change. That is, when GHC processes a type (or a kind -- they're the same!), it must check the type for validity. This check is still completely separate from the check for terms, which comprise a different sublanguage. The validity check for types (kind-checking) will not know about constraints, a nice simplification. You surmise this in your comment; tracking constraints indeed isn't all that hard (that's why I slipped this feature in in the first place), but there are knock-on consequences, as we're now observing.

goldfirere:
> I'd be interested to see if this breaks anything in singletons, servant, and other type-heavy libraries

Me, too. Will have to check after the patch is implemented. I would be surprised if it does. The feature is so wonky that we would have seen more bug reports if anyone used it. But definitely worth checking!

proposals/0000-no-kind-equalities.rst
```diff
@@ -0,0 +1,142 @@
...
+    MkS :: F a ~ Int => a -> a -> S a
+  -- definition of S is OK
+
+  p :: Proxy (MkS True False)  -- this is rejected
+  p = Proxy
+
+Effect and Interactions
+-----------------------
+
+* Much code can be deleted. A partial list is on the `ticket tracking the implemenration of this proposal <https://gitlab.haskell.org/ghc/ghc/-/issues/22298#checklist-of-things-to-remove>`_.
```
JakobBruenker, line 108:
```suggestion
* Much code can be deleted. A partial list is on the `ticket tracking the implementation of this proposal <https://gitlab.haskell.org/ghc/ghc/-/issues/22298#checklist-of-things-to-remove>`_.
```

Ericson2314:
Hmm I am all for jettisoning unused half-baked features (*looks at Safe Haskell*), but I *also* very much *do* want promoted constraints long term. I think they are essential in order to deprecate type families which are all wrong.

Do you, @goldfirere and @simonpj anticipate we will be able to redo this properly sometime later?

simonpj:
> Do you, @goldfirere and @simonpj, anticipate we will be able to redo this properly sometime later?

For some value of "later", yes.  But it would be a pretty major change, adding the entire type-class infrastructure at the kind level.  Step by step.

Ericson2314:
@simonpj Agreed it is a major change. Just want to make sure the difficulties we saw here stem from the current implementation, and not the idea itself being trickier than it already sounds. Thanks!

AntC2:
>  a mis-feature, rather like contexts in data types ...

A chap going by 'Simonpj' appears to be disagreeing with you around 1999. GHC implemented them differently vs Hugs. The Report was "inexplicit".

I like the pre-1999 semantics. (I think Wadler was plain wrong.) I've re-implemented that semantics in Hugs - to give something like the double-constraints of Pattern Syns. I much prefer the simplicity vs GADTs. (And of course it avoids smuggling Constraint dictionaries inside the constructed data.)

RyanGlScott:
> I'd be interested to see if this breaks anything in `singletons`

As `singletons`' maintainer, I can confidently say that `singletons` does _not_ use kind constraints.

In fact, I'm not aware of _any_ Haskell library that uses them. At one point, I wrote a [blog post](https://ryanglscott.github.io/2021/08/01/equality-constraints-in-kinds/) about equality constraints in kinds, but I couldn't find any particularly compelling examples to include besides some rather contrived ones. If this proposal were implemented, that whole post would become moot. After reading the proposal, however, I'm convinced that making the blog post obsolete is a positive thing. Having this feature be a maintenance burden is a pretty compelling argument against its continued existence.

goldfirere:
It sounds like this one is ready for a decision. Please submit to the committee, @nomeata.

aspiwack:
:wave: Shepherd here.

The proposal is very well written, and I feel confident to make a recommendation to the committee, probably Monday or Tuesday. I'll ask for two changes:
- [ ] Add the non-example that `data T a where {MkInt :: Int -> T Int}` can be promoted. I didn't realise that they would while reading the text, so I think it would add a lot of clarity
- [ ] I'll be less insistent about it, but I believe that @RyanGlScott 's [blog post](https://ryanglscott.github.io/2021/08/01/equality-constraints-in-kinds/) (from supra https://github.com/ghc-proposals/ghc-proposals/pull/547#issuecomment-1308962393) makes great additional reading and the proposal would benefit from linking to it.

goldfirere:
Thanks, @aspiwack for the suggestions. Done.

proposals/0000-no-kind-equalities.rst
```diff
@@ -0,0 +1,154 @@
...
+* The inconsistency in the Motivation_ section is gone.
+
+Costs and Drawbacks
+-------------------
+
+* Putting a constraint on a kind is very modestly useful, and this proposal destroys that use case. Instead, users are welcome to constrain data constructors instead of the type itself.
+
+* If and when we get full dependent types -- including the ability to have type class constraints in kinds (which is not permitted today) -- we may need to reintroduce
+this feature, thus making the work of removing it redundant. However, it is quite likely that the new implementation will be much more well-grounded than the
+current one. It is even possible that it will be easier to implement this feature correctly after the current incorrect implementation is ripped out.
```
adamgundry, line 131: Broken list formatting:

```suggestion
* If and when we get full dependent types -- including the ability to have type class constraints in kinds (which is not permitted today) -- we may need to reintroduce
  this feature, thus making the work of removing it redundant. However, it is quite likely that the new implementation will be much more well-grounded than the
  current one. It is even possible that it will be easier to implement this feature correctly after the current incorrect implementation is ripped out.
```

aspiwack:
I've just marked this proposal as accepted.

The committee notes that the most likely source of breakage is the implicit use of the feature by defining a GADT with an explicit equality constraint then promoting it. But we still believe that this will be quite rare and relatively straightforward to fix if it happens in practice.

@nomeata I'll leave the merge to you.

proposals/0000-no-kind-equalities.rst
MangoIV: > In truth, explicit equality constraints are already slightly different than GADT refinements: MkG :: (a ~ Int) => a -> G a has an extra field to store the equality at runtime, while MkG :: Int -> G Int does not. But this would make the difference more apparent, indeed.

Hi, @goldfirere thank you for that information. Can you please clarify, why this is the case? Is it purely for performance reasons? If equality is special cased in GADTs, why is it not special cased in ADTs using -XExistentialQuantification (but no GADT syntax)? I.e. in detail I don't understand why 

```haskell
data Dict c where
  Dict :: c => Dict
  
data Dict' c = c => Dict' 

x :: Dict (Eq Int) 
x = Dict 

x' :: Dict' (Eq Int)
x' = Dict' 
``` desugar the same while 

```haskell
type SUnit :: () -> Type
data SUnit u where SUnit :: SUnit '()

type SUnit' :: () -> Type
data SUnit' u = u ~ '() => SUnit'

y :: SUnit '()
y = SUnit

y' :: SUnit' '()
y' = SUnit'
``` desugar differently (SUnit' gets an additional `Eq#`) while pattern matching on them desugar exactly the same? 

(please excuse that this question is rather for my own interest and less relevant for the conversation)

Comment (RyanGlScott) [reply proposals/0000-no-kind-equalities.rst:89]: > Can you please clarify, why this is the case? Is it purely for performance reasons?

In GHC, equality constraints of the form `(a ~ b)` are represented as a type class. A primitive, built-in type class, but a type class nonetheless. Moreover, the return kind of `(~)` is `Constraint`, so equality constraints are represented with pointers at runtime. Unless a lot of optimizations happen, this means that equality constraints will almost certainly incur some runtime cost.

There are no equality constraints in a type like `MkG :: Int -> G Int`, so GHC is free to pick a different approach here. And indeed it does: under the hood, the type of `MkG` looks something like this:

```hs
MkG :: forall a. (a ~# Int) => Int -> G a
```

Where `(~#)` is an _unlifted_ equality. In constrast to `(~)`, which is _lifted_ (i.e., represented with a pointer at runtime), an unlifted equality is erased at runtime. This is possible because `(a ~# Int)` represents a raw coercion from `a` to `Int`, and coercions are erased when compiling from GHC's core language to lower-level intermediate languages. If you're curious about the finer details, I invite you to read [`Note [The equality types story]`](https://gitlab.haskell.org/ghc/ghc/-/blob/756a66ec0875b675bd3256d46d57419827312426/compiler/GHC/Builtin/Types/Prim.hs#L978-1136) in the GHC source code.

> If equality is special cased in GADTs, why is it not special cased in ADTs using -XExistentialQuantification (but no GADT syntax)?

The ability to write something like `MkG :: Int -> G Int` is a feature that is exclusive to GADT syntax. The closest you can get with "traditional" data constructor syntax is something like this:

```hs
data G a = (a ~ Int) => MkG Int
```

But again, that will compile differently due to the explicit, lifted equality constraint. There is no equivalent syntax for traditional data constructors.

> desugar differently (SUnit' gets an additional `Eq#`) while pattern matching on them desugar exactly the same?

I'm not sure what you mean by "pattern matching on them desugar exactly the same". That's certainly not the case when I compile this example:

```hs
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE GADTs #-}
module Foo where

import Data.Kind
import Data.Type.Equality

type SUnit :: () -> Type
data SUnit u where SUnit :: SUnit '()

type SUnit' :: () -> Type
data SUnit' u = u ~ '() => SUnit'

z :: SUnit u -> u :~: '()
z SUnit = Refl

z' :: SUnit' u -> u :~: '()
z' SUnit' = Refl
```

Using GHC 9.4.4:

```
$ ghc-9.4.4 Foo.hs -fforce-recomp -ddump-simpl -dno-typeable-binds
[1 of 1] Compiling Foo              ( Foo.hs, Foo.o )

==================== Tidy Core ====================
Result size of Tidy Core
  = {terms: 20, types: 48, coercions: 9, joins: 0/0}

-- RHS size: {terms: 1, types: 1, coercions: 1, joins: 0/0}
Foo.$WSUnit [InlPrag=INLINE[final] CONLIKE] :: SUnit '()
[GblId[DataConWrapper],
 Caf=NoCafRefs,
 Unf=Unf{Src=InlineStable, TopLvl=True, Value=True, ConLike=True,
         Guidance=ALWAYS_IF(arity=0,unsat_ok=True,boring_ok=False)
         Tmpl= Foo.SUnit @'() @~(<'()>_N :: '() GHC.Prim.~# '())}]
Foo.$WSUnit = Foo.SUnit @'() @~(<'()>_N :: '() GHC.Prim.~# '())

-- RHS size: {terms: 6, types: 12, coercions: 4, joins: 0/0}
z :: forall (u :: ()). SUnit u -> u :~: '()
[GblId, Arity=1, Unf=OtherCon []]
z = \ (@(u_aGC :: ())) (ds_dGY :: SUnit u_aGC) ->
      case ds_dGY of { SUnit co_aGF ->
      (Data.Type.Equality.$WRefl @() @u_aGC)
      `cast` (((:~:) <()>_N <u_aGC>_N co_aGF)_R
              :: (u_aGC :~: u_aGC) ~R# (u_aGC :~: '()))
      }

-- RHS size: {terms: 10, types: 19, coercions: 4, joins: 0/0}
z' :: forall (u :: ()). SUnit' u -> u :~: '()
[GblId, Arity=1, Unf=OtherCon []]
z'
  = \ (@(u_aGs :: ())) (ds_dGT :: SUnit' u_aGs) ->
      case ds_dGT of { SUnit' $d~_aGv ->
      case GHC.Types.eq_sel @() @u_aGs @'() $d~_aGv of co_aGM
      { __DEFAULT ->
      (Data.Type.Equality.$WRefl @() @u_aGs)
      `cast` (((:~:) <()>_N <u_aGs>_N co_aGM)_R
              :: (u_aGs :~: u_aGs) ~R# (u_aGs :~: '()))
      }
      }
```

Note that `z` matches on `SUnit` and immediately binds a coercion (i.e., unlifted equality) `co_aGF`. In `z'`, however, matching on `SUnit'` doesn't provide immediate access to a coercion. Instead, it binds `$d~_aGv`, which is a lifted equality. To get access to the underlying coercion, it has to apply the `eq_sel` function. (In GHC's Core language, `eq_sel` converts an equality class dictionary into its underlying coercion.) Applying `eq_sel` can have a runtime cost, so there is definitely a difference.

MangoIV: Thank you for the long and elaborate answer, this definitely helps; the only question I'm now asking is, _why exactly_ can I not have a normal ADT witness an unlifted equality (and why does ghc not recognize that the ADT equivalent of the GADTs are isomorphic and _could_ be represented the same way)? Or am I missing something here?

MangoIV: (also apparently if you don't use the coercion (in your case to construct the Refl), it doesn't put in the second case (on ghc 925), I guess that's why I assumed that it would always look the same.

MangoIV: >  * We can't actually declare it that way because we don't have syntax for `~#`. And `~#` isn't a constraint, so even if we could write it, it wouldn't kind check.

(I still don't know why and why it's fundamentally impossible but I guess this explains it)

RyanGlScott: > _why exactly_ can I not have a normal ADT witness an unlifted equality

I'm not sure how to answer this question other than to repeat what I said above: there is simply no syntax for this in the language. You could propose such a syntax, I suppose, but that would be a separate proposal.

> and why does ghc not recognize that the ADT equivalent of the GADTs are isomorphic and could be represented the same way

If we're being precise, `MkG :: forall a. (a ~ Int) => Int -> G a` and `MkG :: forall a. (a ~# Int) => Int -> G a` aren't isomorphic. This is because `(a ~# Int)` is unlifted, which means that it cannot be inhabited by bottoming values, whereas `(a ~ Int)` (a lifted type) _can_ be inhabited by bottoming values. If the compiler knows that a dictionary of type `(a ~ Int)` is non-bottoming, it is within its rights to optimize it to an unlifted equality, but it is not guaranteed that this will always happen (or even be possible).

> (I still don't know why and why it's fundamentally impossible but I guess this explains it)

Is "fundamentally impossible" referring to the inability to use `(~#)` in the source language? This is simply because `(~#)` is a magical type that isn't directly accessible. But even if you _could_ write something like `(a ~# Int) => ...` directly in the source language, that wouldn't kind-check because the thing to the left of `=>` has to be of kind `Constraint`, whereas `(#~)` has return kind `TYPE (TupleRep '[])`. (See [this dormant proposal](https://github.com/ghc-proposals/ghc-proposals/pull/116), which ran into similar issues.) Thus far, no one has created a proposal to allow this syntax that addresses these issues.

MangoIV: Okay, that makes total sense, thank you. Also thank you for taking the time to explain this to me, this really helped.

Icelandjack:
```haskell
type  EnrFunctorOf :: forall enr s t. Cat enr -> forall s'. s ~ s' => EnrCat enr s -> forall t'. t ~ t' => EnrCat enr t -> (s -> t) -> Constraint
class EnrFunctorOf @enr @s @t enriched @s' source @t' target f where
  enrFmap :: (a -source→ a') -enriched→ (f a -target→ f a')
```

I ran into this use case, of allowing `s` and `t` to be quantified at both `EnrFunctorOf @_ @s @t enriched source target` and `EnrFunctorOf enriched @s source @t target`.

PR #549, opened 2022-11-10
Strengthen decomposition for newtypes

simonpj:
This proposal  makes a small technical change that will improve type inference in the presence of newtypes, by allowing them to be decomposed without losing completeness.

[Rendered form](https://github.com/ghc-proposals/ghc-proposals/blob/wip/spj-decompose-newtypes/proposals/0000-decompose-newtypes.rst)

### DIFF
--- a/proposals/0000-decompose-newtypes.rst
+++ b/proposals/0000-decompose-newtypes.rst
@@ -24,7 +24,7 @@ The problem
 Consider::
 
   type family F a
-  newtype T x = MkP (F x)
+  newtype T x = MkT (F x)
 
 and suppose we typecheck the function::
 
@@ -33,14 +33,27 @@ and suppose we typecheck the function::
 
 Then we end up trying to prove ``[W] (T Int) ~R (T Bool)``.  If ``MkT`` is in scope, it's easy: just
 unwrap the newtype.  But if not, GHC today will decomposing this to
-``[W] Int ~R Bool``, and complain that it can't prove that.  But it's totally wrong to claim that
+``[W] Int ~R Bool``, and complain that it can't prove that::
+
+    Foo.hs:12:7: error: [GHC-18872]
+        • Couldn't match type ‘Int’ with ‘Bool’
+            arising from a use of ‘coerce’
+        • In the expression: coerce x
+          In an equation for ‘f’: f x = coerce x
+       |
+    12 | f x = coerce x
+       |       ^^^^^^
+
+But it's totally wrong to claim that
 we need ``Int ~R Bool``, because of the type family.  (E.g. suppose ``F Int = Char`` and ``F Bool = Char``.)
 
 So the error message is bogus.  In more complicated situations we might unify type variables
-that should not be unified etc.
+that should not be unified, causing more misleading, far-away errors.
 
-In short, the decomposition step for newtypes is always *sound* (if typechecking succeeds, all is well),
-but it is not always *complete* (typechecking may fail when it shouldn't).
+In short,
+
+* the decomposition step for newtypes is always *sound* (if typechecking succeeds, all is well),
+* but it is not always *complete* (typechecking may fail when it shouldn't).
 
 So you may say "do not decompose newtype wanteds".  But then suppose
 we are trying to prove ``[W] (IO Int) ~R (IO Age)``, where::
@@ -49,10 +62,12 @@ we are trying to prove ``[W] (IO Int) ~R (IO Age)``, where::
    newtype IO a = MkIO (State -> (State,a))
 
 and ``MkAge`` is in scope, but ``MkIO`` is not (being rightly abstract).
-If we do not decompose newtypes, we'd be totally stuck, which is not good for the programmer.
+If we do not decompose newtypes, GHC would be totally stuck, which is not good for the programmer.
 
-So GHC has a weird special case, where it says "decompose the newtype unless there are
-in-scope Given equalities".  (Interesetd parties can look at
+So GHC has a weird special case, where it says "as a last resort, decompose the newtype unless there are
+in-scope Given equalities".  The "unless" part is an ad-hoc heuristic,
+intended to make incompleteness harder to trigger.
+(Interested parties can look at
 ``can_decompose`` in ``GHC.Tc.Solver.Canonical.canTyConApp``, and the
 discussion on `this merge request <https://gitlab.haskell.org/ghc/ghc/-/merge_requests/9282>`_.
 
@@ -75,11 +90,11 @@ Here's the CO-NTH rule in the paper, Fig 5::
 The CO-NTH rule governs completeness of decomposition for wanted constraints: if CO-NTH holds
 for each argument i, it means that nothing is lost by decomposing.
 
-But note "H is not a newtype"; that stops us decomposing newtypes (except via the ad-hoc
-rule above).
+But note "H is not a newtype"; that means **we can never decompose a newtype** except
+at risk of incompleteness.
 But in the case of the newtype ``IO`` above, it is easy to see that
 if we know ``(IO t1) ~R (IO t2)`` then we certainly know ``t1 ~R t2``, and vice versa.
-**Nothing is lost by decomposing ``IO``!**
+**Nothing is lost by decomposing ``IO``!**  CO-NTH is too conservative.
 
 The proposal below (see Proposed Change Specification) strengthens CO-NTH to allow
 this decomposition.
@@ -101,6 +116,23 @@ First, change the CO-NTH rule as follows::
   ---------------------------------------------------------------------------------------    CO-NTH
      G |- Nth(i) G : taus(i) ~rs(i) sigmas(i)
 
+Note the extra "or rs(i) is representational".  That says that if we know ``(T ty1) ~R (T ty2)``
+then we know ``ty1 ~R ty2`` *if ``T``'s argument has representational role*. When is that?
+Here's an example::
+
+  type family F
+  newtype N a b c = MkN ([a], F c)
+
+  --                   a            b      c
+  -- type role N representational phantom nominal
+
+Here
+* ``b`` has phantom role: it is not even mentioned in the RHS
+* ``c`` has nominal rule: it appears in the RHS but only in the argument of a type family
+* ``a`` has reprsentational role: it appears in the RHS, and *not* under a type family
+
+Representational role means "if you know the RHS type, then you know ``a``"
+
 Change 2: user role signatures
 ::::::::::::::::::::::::::::::
 
@@ -112,9 +144,10 @@ In GHC today, the user can declare explicit roles::
 Here the argument of ``P`` is not mentioned in its RHS, so its role would be inferres as ``phantom``.  But GHC today allows you to override a phantom role
 to ``representational``.
 
-If we continue to allow this, the new CO-NTH rule would be unsound.  Clearly, knowing `(P t1) ~R (P t2)` does **not** imply that `t1 ~R t2`.
+If we continue to allow this, the new CO-NTH rule would be unsound.  Clearly, knowing ``(P t1) ~R (P t2)`` does **not** imply that ``t1 ~R t2``.
 
-Conclusion: **the new CO-NTH requires that we do not allow a user to give a `representational` role for a phantom argument.**
+Conclusion: **the new CO-NTH requires that we do not allow a user to give a representational role for a phantom argument.**  Any such
+attempt would simply be rejected.
 
 With that change, CO-NTH is sound.
 
@@ -124,6 +157,9 @@ Effect and Interactions
 With the new rule, we can decompose ``(T s1 s2 s3) ~R (T t1 t2 t3)``,
 where ``T`` is a newtype, if *all three parameters are at representational role*.
 That strengthens type inference without introducing incompleteness.
+
+Moreover, arguments with representational role are very common: they are arguments that
+are mentioned, and not under type families.  The ``IO`` example above is a case in point.
 
 
 Costs and Drawbacks
--- a/proposals/0000-decompose-newtypes.rst
+++ b/proposals/0000-decompose-newtypes.rst
@@ -1,0 +1,148 @@
+Strengthen newtype decomposition
+================================
+
+.. author:: Simon Peyton Jones
+.. date-accepted::
+.. ticket-url::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+This proposal would strengthen the CO-NTH rule for coercions, thereby making type
+inference succeed more often.
+
+Motivation
+----------
+
+The problem
+:::::::::::
+
+Consider::
+
+  type family F a
+  newtype T x = MkP (F x)
+
+and suppose we typecheck the function::
+
+  f :: T Int -> T Bool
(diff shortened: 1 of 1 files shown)


proposals/0000-decompose-newtypes.rst
```diff
@@ -0,0 +1,182 @@
...
+  newtype P a = MkP Int
+  type role P representational
+
+Here the argument of ``P`` is not mentioned in its RHS, so its role would be inferres as ``phantom``.  But GHC today allows you to override a phantom role
+to ``representational``.
+
+If we continue to allow this, the new CO-NTH rule would be unsound.  Clearly, knowing ``(P t1) ~R (P t2)`` does **not** imply that ``t1 ~R t2``.
+
+Conclusion: **the new CO-NTH requires that we do not allow a user to give a representational role for a phantom argument.**  Any such
+attempt would simply be rejected.
```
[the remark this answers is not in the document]
simonpj: > To be clear, I take it that this restriction applies only to newtypes, not role ascriptions on datatypes?

Yes indeed.  I will clarify.

>  What about role ascriptions on classes (potentially represented via newtypes if they have a single method)?

Huh -- that is a fine question.  I think classes should be treated uniformly, unaffected the implementation choice to implement them with a newtype. I'm not confident that GHC does so.  But equalities *between class constraints* are rare... like `Eq t1 ~ Eq t2`. I'm not sure they can occur at all, actually.  Maybe with impredicative types....  So, very much a corner case, and one that might deserve study outside this proposal.

> What about type role P nominal, is that allowed?

Yes you can upgrade to `nominal`.

> But it would feel strange if the definition of P could not be given the "weaker" role representational but could be given the "stronger" role nominal.

Ah, but `representational` is very strong indeed!  It is not weaker (in this sense).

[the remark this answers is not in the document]
adamgundry: Thanks for clarifying!

> > What about role ascriptions on classes (potentially represented via newtypes if they have a single method)?
> 
> Huh -- that is a fine question. I think classes should be treated uniformly, unaffected the implementation choice to implement them with a newtype. I'm not confident that GHC does so. But equalities between class constraints are rare... like `Eq t1 ~ Eq t2`. I'm not sure they can occur at all, actually. Maybe with impredicative types.... So, very much a corner case, and one that might deserve study outside this proposal.

It's certainly a strange corner. I've only once used a `representational` role annotation on a class (https://well-typed.com/blog/2015/07/checked-exceptions/), and in that case the dictionary would be a unit datatype with a `phantom` parameter, as it happens.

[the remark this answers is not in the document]
phadej: I think that in example below occurs `Eq a ~ Eq b` constraint.

```haskell
{-# LANGUAGE RankNTypes #-}
{-# LANGUAGE ImpredicativeTypes #-}
module Constr where

import Data.Type.Equality
import Data.Type.Coercion

{-
-- RHS size: {terms: 11, types: 64, coercions: 6, joins: 0/0}
promote'
  :: forall a b.
     Coercion (Eq a => a -> a -> Bool) (Eq b => b -> b -> Bool)
     -> a :~: b
[GblId, Arity=1, Unf=OtherCon []]
promote'
  = \ (@a_aNJ)
      (@b_aNK)
      (ds_dPd
         :: Coercion
              (Eq a_aNJ => a_aNJ -> a_aNJ -> Bool)
              (Eq b_aNK => b_aNK -> b_aNK -> Bool)) ->
      case ds_dPd of { Coercion $dCoercible_aNN ->
      case GHC.Types.coercible_sel
             @(*)
             @(Eq a_aNJ => a_aNJ -> a_aNJ -> Bool)
             @(Eq b_aNK => b_aNK -> b_aNK -> Bool)
             $dCoercible_aNN
      of co_aO5
      { __DEFAULT ->
      (Data.Type.Equality.$WRefl @(*) @a_aNJ)
      `cast` (((:~:) <*>_N <a_aNJ>_N (Nth:0 (Nth:3 co_aO5)))_R
              :: (a_aNJ :~: a_aNJ) ~R# (a_aNJ :~: b_aNK))
      }
      }

-}
promote' :: Coercion (Eq a => a -> a -> Bool) (Eq b => b -> b -> Bool) -> a :~: b
promote' Coercion = Refl
```

Ericson2314:
I will need to think through this for, but I this seems nearly the exact opposite of the direction I had wanted to go with, which is getting rid of nth co, entirely. So I am skeptical.

proposals/0000-decompose-newtypes.rst
```diff
@@ -0,0 +1,192 @@
...
+Moreover, arguments with representational role are very common: they are arguments that
+are mentioned, and not under type families.  The ``IO`` example above is a case in point.
+
+
+Costs and Drawbacks
+-------------------
+
+Implementation is easy.
+
+The only cost is Change 2: you can't give a user-ascribed representational role to a phantom argument.
```
phadej, line 178: I found

```haskell
newtype UnliftedArray# (a :: TYPE 'UnliftedRep) = UnliftedArray# ArrayArray#
type role UnliftedArray# representational
```

in `primitive-unlifted`. (`primitive` uses `data` so there is no newtypes).

AFAICT, if this change is accepted, the `primitive-unlifted` interface will be worse, phantom role would be wrong, and nominal is unnecessarily strong. Using `data` is not an option (as then it won't be `primitive-unlifted`).

---

Another, example is 

```haskell
-- | Type-indexed list
newtype TypedList (f :: (k -> Type)) (xs :: [k]) = TypedList [Any]
  deriving (Typeable)
type role TypedList representational representational
```

from `dimensions`. Similar type occurs many times on hackage (`vinyl`'s `Rec`, `large-records`' `Rec`), but these use `TypedList [f Any]` like encoding, so `f` is actually not-phantom. Similarly `some` package has `newtype Some f = Some (f Any)`.

However, note the `xs` is also marked as representational. And it makes sense. One should be able to `coerce` `TypedList Maybe '[Int]` to `TypedList Maybe `[Down Int]`. 

---

[`data-elevator`](https://hackage.haskell.org/package/data-elevator-0.1.0.1/docs/Data-Elevator.html) defines

```haskell
type role Strict representational
newtype Strict (a :: LiftedType) where
  Strict_ :: Any @UnliftedType -> Strict a
```

treeowl: `prim-unlifted` is kind of a mess. The `BoxedRep` stuff was supposed to fix it, but I don't know if it can do so yet.

phadej: @treeowl it's still an example of `newtype WithSafeInterface a = Unsafe Any` pattern. These are newtypes for a reason. Maybe GHC should have `type family AnyOne (a :: k) :: l where`.

simonpj:
Those are very illuminating examples, @phadej, thank you. (I love the
GHC Proposals process -- you get such good feedback.)

One example you give is
```
newtype TypedList (f :: (k -> Type)) (xs :: [k]) = TypedList [Any]
type role TypedList representational representational
```
And you rightly say:

* one should be able to coerce `TypedList Maybe '[Int]` to `TypedList Maybe '[Age]`
* but one should not coerce `TypedList Maybe '[Int]` to `TypedList Maybe '[Bool]`.

and *that* is why you want a representational role for the argument. That is
quite compelling.

Let's go back to my original goal, though, of solving Wanteds.  Suppose I want to
solve this constraint:
```
[W] (TypedList t1)  ~R  (TypedList t2)
```
(I'll simplify to a single type parameter to save clutter.)  If I solve `[W] t1 ~R t2`, then I'm done.  But the goal of completeness asks "could there be any *other* way to solve it?  For example, what about unwrapping the newtype?  e.g. to solve `TypedList Int ~R TypedList Bool`, just unwrap:
```
  TypedList Int    ~R   [Any]    ~R    TypedList Bool
```
So if the inference engine commits to solving `[W] t1 ~R t2`, it might get
stuck, when an answer is available.

*BUT* of course that it not what the programmer intended.  Presumably there
is all sorts of unsafe-coerce magic happening under the hood, and you absolutely
must not unwrap.  So actually *decomposition does not threaten completeness*
in such cases.

My conclusion
* We should not change CO_NTH, for the reasons you say
* But the solver can still decompose newtype Wanteds with representational roles (which it does not do right now)

The only thing that the status quo does not get, which the proposal does, is this:
```
f :: Coercible (IO a) (IO b) => a -> b
f x = coerce x
```
(assuming IO is abstract).  We need to coerce x with `a ~R b`; and we can only
get that from `[G] (IO a) ~R (IO b)`.  The new CO-NTH rule would allow that;
the status quo does not.   But never mind!  No one has been asking for that anyway.

-------------

TL;DR: I think I can withdraw the proposal, and simply implement a change
to the implementation of the type inference engine (which of course does not
need a proposal).

Illuminating, as I say.

Does anyone else have thoughts they'd like to add?

adamgundry:
I'm trying to understand the new proposed behaviour.  Is this right?

I think it means that when the solver sees `[W] (T s_1 ... s_N) ~R (T t_1 ... t_N)`:
 * If `T`'s constructor is in scope, simplify by unwrapping the newtype using the axiom. (If the user gave `T` a conservative role signature, this might be slightly more permissive than the next bullet.)
 * If `T`'s constructor is not in scope, and none of the parameters have nominal role, simplify to `[W] s_i ~R_i t_i`. for each parameter `i = 1...N`. (In particular it is okay to have a mixture of representational and phantom roles, I think?)
 * If `T`'s constructor is not in scope, and at least one parameter has nominal role, do nothing. (Or can/should we try to solve "derived" wanteds for each representational parameter, even though we can't necessarily prove the original wanted yet?)

This avoids the problem in the Motivation, because in that case the parameter has nominal role. And while it is not quite complete, any proof of the wanted must ultimately either use `CO_TyConApp` (in which case decomposing is correct), the newtype axiom (which we will use if it is in scope), or a coercion variable corresponding to a Given (which we are agreeing to not worry about).

simonpj:
Pretty much right.

* If T's constructor is in scope, simplify by unwrapping the newtype using the axiom. (If the user gave T a conservative role signature, this might be slightly more permissive than the next bullet.).   **Yes: we already do this**
*  If T's constructor is not in scope, and none of the parameters have nominal role, simplify to [W] s_i ~R_i t_i. for each parameter i = 1...N. (In particular it is okay to have a mixture of representational and phantom roles, I think?)   **Actually I was proposing "if all of the parameters have representational role, then decompose**.  Phantom roles would lead to incompleteness.
* If T's constructor is not in scope, and at least one parameter has nominal role, do nothing. (Or can/should we try to solve "derived" wanteds for each representational parameter, even though we can't necessarily prove the original wanted yet?)   **I'd say "otherwise do nothing"**

adamgundry:
> * If T's constructor is not in scope, and none of the parameters have nominal role, simplify to [W] s_i ~R_i t_i. for each parameter i = 1...N. (In particular it is okay to have a mixture of representational and phantom roles, I think?)  **Actually I was proposing "if all of the parameters have representational role, then decompose**.  Phantom roles would lead to incompleteness.

That seems unnecessarily restrictive, though? For example, consider this newtype:
```hs
type N :: Type -> Type -> Type
newtype N a b = MkN a
type role N representational phantom
```
Currently GHC 9.4 accepts the following (without the constructor `MkN` in scope):
```hs
foo :: Coercible a b => N a c -> N b d
foo = coerce
```
But I think the proposed new rule would reject it? Or did I misunderstand how the proposed new rule relates to the existing implementation?

In general it seems unclear to me how many existing programs are accepted by the existing "weird special case" but will be rejected after this "change to the implementation of the type inference engine". :wink:

proposals/0000-decompose-newtypes.rst
```diff
@@ -0,0 +1,192 @@
...
+But note "H is not a newtype"; that means **we can never decompose a newtype** except
+at risk of incompleteness.
+But in the case of the newtype ``IO`` above, it is easy to see that
+if we know ``(IO t1) ~R (IO t2)`` then we certainly know ``t1 ~R t2``, and vice versa.
+**Nothing is lost by decomposing ``IO``!**  CO-NTH is too conservative.
+
+The proposal below (see Proposed Change Specification) strengthens CO-NTH to allow
+this decomposition.
+
+
```
goldfirere, line 100: I am unmoved by the current motivation, because I don't really see the problem today.

Instead, I think a different motivation would hold more water: You want `[G] (IO Int) ~R (IO Age)` to imply `[G] Int ~R Age`, which indeed we do not have today. But actually I find the usefulness of that implication to be minimal, so I'm still stuck on the motivation piece.

proposals/0000-decompose-newtypes.rst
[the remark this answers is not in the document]
goldfirere: Hmm.... my guess is that GHC gets this wrong today, in that I imagine it will decompose a newtype-class. We should check. ...checks.... Actually, GHC does get this right. `class C a where deflt :: a` does not decompose, whereas `class C2 a where deflt :: a; blah :: a -> a` does. So, nothing to see here.

goldfirere (commented):
I'm quite skeptical of this -- mostly because I don't buy the motivation. I think it's quite likely that the new change is sound. I'm not sure of the effect on inference, though.

Responses to other comments above:

* Ah. It seems others have pushed in a similar direction to me. Good. And you've discovered the actual thing your proposal delivers, which is decomposing givens... but no one is asking for this. Convergence!

* What's the change to the inference engine you'd like? I don't really see a change arising here.

simonpj:
See the discusion on https://gitlab.haskell.org/ghc/ghc/-/issues/22519.  We decided in the end to adopt a plan that does not involve changing the language at all.

The discussion on this proposal was really useful in reaching that conclusion, so thank you.

But this GHC proposal can now be parked, or put in abeyance.  Maybe closing altogether is more than we need do, since there are useful ideas here.  @nomeata can you change the status appropriately?

Ericson2314:
@simonpj you could convert it to a draft PR.

simonpj:
> @simonpj you could convert it to a draft PR.

What is "it"?   I'm not at all sure what you are asking for.

adamgundry:
@simonpj You pushed a commit 9ca47a77a8e6457200050a06f00ee7a3294f0f88 to the `wip/spj-decompose-newtypes` branch (which underlies this PR), but the commit contains a separate (draft) proposal. So I think it belongs on its own branch. (At the moment https://github.com/ghc-proposals/ghc-proposals/pull/549/files shows both the proposal discussed in this PR, and the new draft, but it should have only the former.)

Ericson2314:
By "it" I mean this PR. i.e. convert this PR to a draft PR.

But yes, I got a notification ad assumed the last comment was recent, when in fact that was the push, and the last comment was from months ago. Thanks @adamgundry!

simonpj:
> You pushed a commit https://github.com/ghc-proposals/ghc-proposals/commit/9ca47a77a8e6457200050a06f00ee7a3294f0f88 to the wip/spj-decompose-newtypes branch

Ah yes. That was a mistake, sorry.  I have reverted it.

> By "it" I mean this PR. i.e. convert this PR to a draft PR.

How do I do that?

Ericson2314:
It's in top right of the page underneath where "reviewers" are listed. See https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request#converting-a-pull-request-to-a-draft

simonpj:
OK thanks I have done that.   I don't know what the semantics of "draft" are, but it seems better than closing outright.  thanks

proposals/0000-decompose-newtypes.rst
konsumlamm, line 178:
```suggestion
The only cost is Change 2: in a `newtype` declaration, you can't give a user-ascribed representational role to a phantom argument.
```

PR #555, opened 2022-11-17
Higher Order Patterns in Rewrite Rules (under review)

noughtmare:
This is a proposal augmenting our existing rewrite rule mechanism with lightweight higher order matching using higher order patterns, which provides a way to match complex terms which contain locally bound variables. 

For example, consider the rule:
```haskell
"concatMap"   forall next f.   concatMap (\x -> Stream next (f x)) = concatMap' next f
```
Previously the `f x` part (which we call higher order pattern) only matched applications literally, like:
```haskell
concatMap (\x -> Stream next ((\y -> y * 2 + y) x)
```
This proposal allows it to match the beta-equivalent expression:
```haskell
concatMap (\x -> Stream next (x * 2 + x))
```
Which results in:
```haskell
concatMap' next (\x -> x * 2 + x)
```

[Rendered](https://github.com/noughtmare/ghc-proposals/blob/template-applications/proposals/0000-template-patterns.rst)


treeowl:
This looks quite nice, at least in simple cases! Could you go into a bit more detail about eta issues? Could you give some examples of rules that could be expressed by higher-order matching but not by this system?

noughtmare:
I'm not sure exactly which eta issues you mean. Do you mean the things discussed in the 'uniqueness of matching' section? Over at the [GHC issue](https://gitlab.haskell.org/ghc/ghc/-/issues/22465#uniqueness-of-matching) we have added some more info, I'll add that here as well. Is that enough?

The paper by De Moor and Sittampalam on higher order matching uses the example rule `forall f x. foo (f x)`. Here we apply a template variable to another template variable. This often leads to ambiguity. For example if we match that template against the term `0` we can get many possible substitutions:

```
[f :-> \a -> a, x :-> 0]
[f :-> \a -> 0]
[f :-> \g -> g 0, x :-> \a -> a]
[f :-> \g -> g (g 0), x :-> \a -> a]
...
```

I think that paper doesn't really present any practical examples, let alone one that could be useful in a rewrite rule.

simonpj:
Proposal needs section numbers.  Easy to add (there's an automated way)

nomeata:
It's a markdown proposal, I don't think we can have them there. Maybe we should remove the option for authors to use markdown?

treeowl:
@nomeata Please don't remove that option.

noughtmare:
Let's continue the discussion about markdown and RST at #560

jvanbruegge:
Given that there is already an implementation of this proposal, did you try to remove the FB machinery and look if fusion did not degrade?

noughtmare:
@simonpj I've changed this proposal to RST, so it now has section numbers and a table of contents.

noughtmare:
@jvanbruegge I think actually implementing those changes should be done separately from this proposal. However I did try it out and it does work like this:

```haskell
module T2 (foo) where

import Prelude hiding (foldr)
import GHC.Base (build, foldr)

mymap :: (a -> b) -> [a] -> [b]
{-# NOINLINE [0] mymap #-}
mymap f [] = []
mymap f (x:xs) = f x : mymap f xs

{-# RULES
"mymap"     [~1] forall f xs. mymap f xs                   = build (\c n -> foldr (\x ys -> c (f x) ys) n xs)
"mymapList" [1]  forall f.    foldr (\x ys -> f x : ys) [] = mymap f
#-}

foo :: Num a => [a] -> [a]
foo xs = mymap (\x -> x * 2 + x) xs
```

Where the rules fire and this is the resulting Core:

```haskell
foo
  = \ @a_aAz $dNum_aAA xs_ayJ ->
      let { lvl_sCI = fromInteger $dNum_aAA foo1 } in
      mymap
        (\ x_ayP -> + $dNum_aAA (* $dNum_aAA x_ayP lvl_sCI) x_ayP) xs_ayJ
```

But it doesn't work if I:

1. use `foldr` from the Prelude.
2. give `foo` a more concrete type e.g. `[Int] -> [Int]`.

I'm not so worried about point 1 because the `map` from `Data.List` doesn't use the `foldr` from prelude for its rules.

But point 2 is more worrying. I think another optimization interferes. This needs more investigation.

simonpj:
I have taken the liberty of word-smithing the proposal, including adding a related work section.

I have renamed the thing to "higher order patterns"; but that is still up for grabs of course.  "Template pattern" isn't working for me.

I don't think the proposal is tied to changes the the fusion machinery. It may *enable* such changes.  But whether or not we make such changes, this proposal adds very significant new expressiveness to the RULES machinery, while giving nothing up and introducing no new syntax.  The implementation is simple too.  What's not to like :-).

Can I invite anyone who wants to contribute to to so?   I think we can submit to the committee shortly

proposals/0000-template-patterns.rst
```diff
@@ -0,0 +1,376 @@
...
+	A sucessful match finds a *substitution* S: a binding for each template variable, such that applying S to the LHS yields the target.
+* 	After a successful match we replace the target expression with the substitution S applied to the RHS of the rule.
+
+
+In GHC today, a template variable ``v`` matches any expression ``e`` if
+
+* ``e`` has the same type as ``v``
+* No local binder of the template is free in ``e``.
+
+The change proposed here is that a **higher order pattern** matches any expression (of the same type):
```
sgraf812, line 147:
```suggestion
The change proposed here is to match an *entire application* of a template variable of a certain form against the target expression. We call this application form **higher order pattern**.
```

The sentence is broken. Perhaps use something like my suggestion?

proposals/0000-template-patterns.rst
```diff
@@ -0,0 +1,376 @@
...
+
+* 	**Higher order pattern (HOP)**.
+	A *higher order pattern* is an expression of form ``f x y z`` where:
```
sgraf812, line 150: Perhaps split into syntactic definition and its semantics wrt. matching.

```suggestion
* 	**Definition:** A **Higher order pattern (HOP)** is an expression of the form ``f x y z`` where:
```

proposals/0000-template-patterns.rst
```diff
@@ -0,0 +1,376 @@
...
+	- ``x``, ``y``, ``z`` are *local binders* (like ``y`` in rule "wombat" above; see definitions).
+	- The arguments ``x``, ``y``, ``z`` are *distinct* variables
+	- ``x``, ``y``, ``z`` must be term variables (not type applications).
+
+* 	A higher order pattern ``f x y z`` matches *any expression* ``e`` provided:
+
+	- The target has the same type as the template
+	- No local binder is free in ``e``, other than ``x``, ``y``, ``z``.
+
+*	If these two condition hold, the higher order pattern ``f x y z`` matches the target expression ``e``, yielding the substitution ``[f :-> \x y z. e]``.
```
sgraf812, line 162: Doesn't make much sense to have an additional bullet here, in particular because it refers to a sublist in the previous bullet

```suggestion
	If these two condition hold, the higher order pattern ``f x y z`` matches the target expression ``e``, yielding the substitution ``[f :-> \x y z. e]``.
```

proposals/0000-template-patterns.rst
```diff
@@ -0,0 +1,376 @@
...
+::
+
+	Template:    f x y
+	Target:      h (p+1) q      [p:->x, q:->y]
+
+The renaming ``[p:->x, q:->y]`` is done by the matcher (today) on the fly, to make the bound variables of the template and target "line up".
+
+Now, we can:
+
+* Either use the new template-application rule to succeed with ``[f :-> \x y. h (x+1) y]``.
```
sgraf812, line 185:
```suggestion
* Either use the new higher order pattern rule to succeed with ``[f :-> \x y. h (x+1) y]``.
```

proposals/0000-template-patterns.rst
```diff
@@ -0,0 +1,376 @@
...
+* Or use the existing decompose-application rule to match ``(f x)`` against ``(h (p+1))`` and ``y`` against ``q``.  This will succeed, with ``[f :-> \x. h (x+1)]``.
+
+Critically, *it doesn't matter which we do*.
+We get the same result either way.
+That's encouraging.
+
+More generally, we think that if a match exists it is unique (moudulo eta-reduction).
```
sgraf812, line 192: Doesn't eta-reduction have perf repercussions? E.g., if `h` has arity 1 there's a difference between the substitutions. Perhaps immaterial to the proposal, but we are bound to think about it in the implementation.

proposals/0000-template-patterns.rst
```diff
@@ -0,0 +1,376 @@
...
+Existing work on higher order unification has revealed that higher order unification of the template application form is a useful decidable subset of the general problem. This was first observed by Miller in `"Unification Under a Mixed Prefix" <https://repository.upenn.edu/cis_reports/454/>`_. Miller called these template applications 'patterns', but we decided against using that terminology because of the confusion with regular pattern matching in Haskell. Nowadays higher unification with Miller's pattern restriction is commonplace in dependently typed languages such as `Agda <http://www.cse.chalmers.se/~ulfn/papers/thesis.pdf>`_ and `Idris <https://www.type-driven.org.uk/edwinb/papers/impldtp.pdf>`_, which shows that higher order unification with this restriction is still very powerful.
+
+One notable extension of Miller's patterns is the extension by Duggan in his paper `"Unification with extended patterns" <https://doi.org/10.1016/S0304-3975(97)00141-2>`_. Duggan extends the simple template application form to allow projections of local binders as arguments to the template variables. Such an extension might also be applicable to the template application form in this proposal.
```
sgraf812, line 207: Perhaps cite https://cstheory.stackexchange.com/questions/50914/swapping-arguments-of-variables-in-higher-order-pattern-unification/50918#50918, too.

```suggestion
One notable extension of Miller's patterns is the extension by Duggan in his paper `"Unification with extended patterns" <https://doi.org/10.1016/S0304-3975(97)00141-2>`_. Duggan extends the simple higher order pattern form to allow projections of local binders as arguments to the template variables. Such an extension might also be applicable to the higher order pattern form in this proposal. András Kovács writes about further extensions `"in this StackOverflow" <https://cstheory.stackexchange.com/questions/50914/swapping-arguments-of-variables-in-higher-order-pattern-unification/50918#50918>`_.
```

proposals/0000-template-patterns.rst
```diff
@@ -0,0 +1,376 @@
...
+
+Existing work on higher order unification has revealed that higher order unification of the template application form is a useful decidable subset of the general problem. This was first observed by Miller in `"Unification Under a Mixed Prefix" <https://repository.upenn.edu/cis_reports/454/>`_. Miller called these template applications 'patterns', but we decided against using that terminology because of the confusion with regular pattern matching in Haskell. Nowadays higher unification with Miller's pattern restriction is commonplace in dependently typed languages such as `Agda <http://www.cse.chalmers.se/~ulfn/papers/thesis.pdf>`_ and `Idris <https://www.type-driven.org.uk/edwinb/papers/impldtp.pdf>`_, which shows that higher order unification with this restriction is still very powerful.
```
sgraf812, line 205: "template application"

```suggestion
The main related work on matching is `"Higher-order matching for program transformation" <https://www.sciencedirect.com/science/article/pii/S0304397500004023>`_ by De Moor and Sittampalam. This work identifies a subset of the higher order matching problem which is decidable and always has a finite set of possible substitutions. This subset is strictly larger than the subset of higher order patterns which we consider in this proposal. For example, their subset is able to find substitutions to match `forall f x. f x` to `0`. Our proposal does not support this form of higher order matching.

However De Moor and Sittampalam identify the (potential) importance of the special case that we exploit:

    There is a wealth of related work on higher-order matching and unification [5, 7, 11, 13, 16, 18, 24, 25], to name just a few. One important concept identified in some of these works (in particular [16, 18]) is that of a restricted notion of higher-order pattern. To wit, a restricted pattern is a normal term where every occurrence of a free function variable is applied to a list of distinct local variables, and nothing else. For such restricted patterns, much simpler and more efficient matching and unification algorithms are possible. Our algorithm returns all higher-order matches for rules where the pattern satisfies the above restriction; in fact there is at most one such match. We have not yet investigated the efficiency of our algorithm in this important special case.

Existing work on higher order unification has revealed that higher order unification of the higher order pattern form is a useful decidable subset of the general problem. This was first observed by Miller in `"Unification Under a Mixed Prefix" <https://repository.upenn.edu/cis_reports/454/>`_. Miller called our higher order patterns just 'patterns', but we decided against using that terminology because of the confusion with regular pattern matching in Haskell. Nowadays higher unification with Miller's pattern restriction is commonplace in dependently typed languages such as `Agda <http://www.cse.chalmers.se/~ulfn/papers/thesis.pdf>`_ and `Idris <https://www.type-driven.org.uk/edwinb/papers/impldtp.pdf>`_, which shows that higher order unification with this restriction is still very powerful.
```

proposals/0000-template-patterns.rst
```diff
@@ -0,0 +1,376 @@
...
+	::
+
+		foo (\x y z -> x * y + z)
+
+	But not every variable has to occur in the match. It would also match this expression where ``y`` does not occur:
+	::
+
+		foo (\x y z -> x * 2 + z)
+
+* 	Locally bound variables may only occur once.
```
sgraf812, line 237:
```suggestion
* 	Locally bound variables may only occur once in the higher order pattern. But that doesn't mean we reject such non-linear template variable applications.
```

proposals/0000-template-patterns.rst
```diff
@@ -0,0 +1,376 @@
...
+* 	Similarly if the template variable ``f`` is applied to non-variable arguments then it only matches a literal application.
+	Consider this rule:
+	::
+
+		{-# RULES "foo" forall f. foo (\x y -> f x 2 y) = "RULE FIRED" #-}
+
+	This would **not** match:
+	::
+
+		foo (\x y -> x * 2 + y)`
```
sgraf812, line 262:
```suggestion
		foo (\x _ y -> x * 2 + y)`
```

perhaps? otherwise, where does the `2` arg go?

proposals/0000-template-patterns.rst
```diff
@@ -0,0 +1,376 @@
...
+	But again it does contain the higher order pattern ``f x``, so it would match:
+	::
+
+		foo (\x y -> (bar x . baz) 2 y)
+
+Effect and Interactions
+-----------------------
+
+The main effect of this proposal is that rewrite rules involving higher order patterns now match more expressions.
+But the additional matches are guaranteed to be beta equivalent, so this change does not cause existing rules to become semantically incorrect.
```
sgraf812, line 273: What is it, beta or eta? I think eta. Or both? In what sense?

proposals/0000-template-patterns.rst
```diff
@@ -0,0 +1,376 @@
...
+The only contentious interactions could occur due to rules that now overlap under the new rules, for example:
+::
+
+	{-# RULES
+	"foo->bar"  forall f x.  foo x (\y. f y) = bar x f
+	"foo->baz"  forall   x.  foo x (\y. y * 2 + y) = baz x
+	#-}
+
+Previously, the rule ``"foo->baz"`` would always fire when encountering the expression ``foo x (\y. y * 2 + y)``, but now the rule ``"foo->bar"`` also matches.
+However, we do not expect that this occurs in practice.
```
sgraf812, line 284: Well, except when we implement our new `mapList` RULE that will overlap with every FB rule, no?

I think we should simply let "foo->baz" win. It's strictly more specific than "foo->bar", plus that's what backwards compatibility demands.

proposals/0000-template-patterns.rst
```diff
@@ -0,0 +1,376 @@
...
+
+Costs and Drawbacks
+-------------------
+
+1. 	The changes required for this proposal are small (the core of the change is an addition of just 22 lines of code).
+	Small changes can add up, but we think the benefits far outweigh this cost in this case.
+
+2. 	This proposal causes a silent change of behaviour of existing code.
+	It is possible to come up with an artificial system of rewrite rules that produces suboptimal results due to this change.
+	We do not expect this to happen in practice.
```
sgraf812, line 295: Point to the example. Is there one? I don't recall any, at least not where we let a *different* rewrite rule win vs. the status quo. Subsitutions might be different, too, but still equal modulo eta.

proposals/0000-template-patterns.rst
```diff
@@ -0,0 +1,376 @@
...
+
+		{-# RULES "foo" forall (f :: forall a. [a] -> Int). foo (\xs -> 1 + f xs) = 2 + foo f #-}
+
+	Here, the template variable ``f`` has a polymorphic type.
+	With explicit type abstractions and applications the rule looks like this:
+	::
+
+		{-# RULES "foo" forall (f :: forall a. [a] -> Int). foo (/\a. \(xs::[a]) -> 1 + f @a xs) = 2 + foo f #-}
+
+	The proposal could be change such that this rule would match the expression:
```
sgraf812, line 352:
```suggestion
	The proposal could be changed such that this rule would match the expression:
```

proposals/0000-template-patterns.rst
```diff
@@ -0,0 +1,376 @@
...
+
+
+	However, if we change the type of the template variable ``f`` to ``forall a. a -> Int``, then the rule with explicit type abstractions and applications looks like this:
+	::
+
+		{-# RULES "foo" forall (f :: forall a. a -> Int). foo (/\a. \(xs::[a]) -> 1 + f @[a] xs) = 2 + foo f #-}
+
+	(Note: we assume deep subsumption here for simplicity of presentation)
+
+	Now ``@[a]`` is no longer a plain locally bound variable, so this is no longer a higher order pattern.
```
sgraf812, line 365: If we implement sufficient extensions in the future, such as allowing invertible arguments in higher-order patterns, then we'll be able to allow this refactoring, because `[]` is injective (and thus invertible). E.g., we could push `\@a -> @[a]` into our subsitution and presume that `f` really abstracts over `@a`.

Perhaps it's best to wait until we have implemented the extension on the term level.

proposals/0000-template-patterns.rst
```diff
@@ -0,0 +1,376 @@
...
+
+	This seems fragile and we do not know of any practical programs that requires polymorphic template variables in higher order patterns.
+
+2. 	The name "higher order pattern" is still up for debate.
+	Suggestions are welcome.
+
+Implementation Plan
+-------------------
+
+The proposed changes have already been implemented in `#9343 <https://gitlab.haskell.org/ghc/ghc/-/merge_requests/9343>`_.
```
sgraf812, line 375:
```suggestion
The proposed changes have already been implemented in `!9343 <https://gitlab.haskell.org/ghc/ghc/-/merge_requests/9343>`_.
```

sgraf812 (commented):
A few bits and bobs, but all in all great!

proposals/0000-template-patterns.rst
simonpj: Good point.  And it will, already. I'll add words

proposals/0000-template-patterns.rst
simonpj: No I don't think so.  I don't see the problem here.

proposals/0000-template-patterns.rst
simonpj: Actually I'm not sure what this paragraph means.  @noughtmare ?

proposals/0000-template-patterns.rst
simonpj: I'm not sure what this means.  "Additional matches are beta equivalent".  What additional matches?  What does it mean for an additional match to be beta equivalent? @noughtmare ?

simonpj:
Thanks Sebastian!

proposals/0000-template-patterns.rst
noughtmare: I meant the example from the effects and interactions section right before it.

noughtmare: Ah, I see that is not really a problem due to GHC picking the most specific rule. 

I still think it should be possible to construct a set of rules that fires differently.

proposals/0000-template-patterns.rst
noughtmare: I mean that anything that now matches under the new rules must be beta-equivalant to an expression that would have matched under the old rules.

noughtmare: Currently we have a notion of syntactic matching, which means that the target must be syntactically equal to the template.

This proposal broadens/weakens that notion to semantic matching, so the target is allowed to be semantically beta-equivalent to the template in some cases.

So not all reason is lost.

simonpj: OK I have elaborated 3.1 to cover this.  See if you like it.

proposals/0000-template-patterns.rst
sgraf812: Oh, right. Sorry for the confusion.

noughtmare:
Thanks everyone for the discussion, reviews, and improvements! I think the proposal is now much clearer than the initial version.

I've removed the unresolved question because it is irrelevant as long as we don't consider type variables in a higher order pattern. I forgot that we ruled those out in this proposal. Such extensions will be left for future work.

@nomeata I think this proposal is ready for review.

simonpj:
> @nomeata I think this proposal is ready for review.

i.e. submitting to the committee

noughtmare:
Ah, I was using the terminology from [the documentation](https://github.com/ghc-proposals/ghc-proposals#how-to-bring-a-proposal-before-the-committee):

> Add a comment to the pull request, briefly summarizing the major points raised during the discussion period and stating your belief that the proposal is ready for review. In this comment, tag the committee secretary (currently @nomeata).

But I agree that "ready for review" is a bit ambiguous.

proposals/0000-template-patterns.rst
```diff
@@ -0,0 +1,367 @@
...
+* 	**Matching of higher order patterns (HOP-matching)**.
+        A higher order pattern ``f x y z`` (in the template) matches *any target expression* ``e`` provided:
+
+	- The target has the same type as the template
+	- No local binder is free in ``e``, other than ``x``, ``y``, ``z``.
+
+	If these two condition hold, the higher order pattern ``f x y z`` matches the target expression ``e``, yielding the substitution ``[f :-> \x y z. e]``.
+	Notice that this substitution is type preserving, and the RHS of the substitution has no free local binders.
+
+*       **Ambiguity breaking**.  When matching template ``(etmpl x)`` against target ``(etarget x)``, do *not* use HOP-matching even if ``(etmpl x)`` is a HOP; instead simply match ``etmpl`` (which is also a HOP) against ``etarget``.  See the next subsection for the justification for this refinement.
```
adamgundry, line 167: This point isn't completely clear to me. I think it would be clearer to specify in general how a template application is matched against a target application, rather than merely giving a negative rule for when HOP-matching does not apply.

I think the rule is something like this.... When matching `(etmpl1 etmpl2)` against target `(etarget1 etarget2)`
 * If `(etmpl1 etmpl2)` is not a HOP, decompose the application (i.e. match `etmpl1` against `etarget1` and `etmpl2` against `etarget2`.
 * If `(etmpl1 etmpl2)` is a HOP, and `etarget2` has been renamed to be equal to `etmpl2` (which is necessarily a locally bound variable), decompose the application.
 * Otherwise, `(etmpl1 etmpl2)` is a HOP, so use HOP-matching against `(etarget1 etarget2)`.

Is that right?

simonpj: Thanks Adam.

@noughtmare I think that the "Ambiguity breaking" bullet is unnecessay.  There is no reason to *specify* that we must not use HOP-matching in the ambiguous case -- it's an implementation choice, since it's going to match either way.  And it's confusing to imply that it might matter.  I think we can simply drop the bullet altogether.

Then at the end of 3.1 instead of "The refined rule picks (2), which yields success with as few lambdas as possible in the match." we can simply say "It makes no difference which choice is made, so the implementation is free to choose either".

To return to Adam's question I think the *specificaiton* can be:
* If the template is a HOP, use HOP-matching
* If not, try to decompose applications

Simple.  We shouldn't over-specify.

@noughtmare ok?  I didn't want to make this change without seeing if you had some reason for the way it is.

noughtmare: As the proposal says, HOP matches can cause eta-expansions compared to the status quo, which can have a performance impact. So, I still think the ambiguity breaking is a good idea.

Now that I'm thinking about it, there are also situations like matching the template `forall f. f x` with the target `foo x y x` (where `x` is local and `y` is not local), which I believe would not match due to ambiguity breaking. But there is a HOP match: `[f :-> \x -> foo x y x]`. 

So, it seems the ambiguity breaking rule does need to be modified, but I don't think removing it altogether is a good solution. Perhaps we can just say that HOP matches should be in maximally eta-reduced form? I believe that gets to the point without mentioning implementation details and it still seems relatively easy to implement.

simonpj: All good points.  Yes, I think we should modify

- If these two condition hold, the higher order pattern f x y z matches the target expression e, yielding the substitution [f :-> \x y z. e]. Notice that this substitution is type preserving, and the RHS of the substitution has no free local binders.

to say

- If these two condition hold, the higher order pattern f x y z matches the target expression e, yielding the substitution [f :-> e'], where e' is etaReduce( \x y z. e). Notice that this substitution is type preserving, and the RHS of the substitution has no free local binders.

Then we can add some words about `etaReduce`.   In the specification section we can just say "it's normal eta-reduction; e..g. `etaReduce( \x y z. f x y z) = f` and `etaReduce(\x y z.  f x y y z) = \x y. f x y y`.  In 3.1 we can motivate etaReduce.

OK?

noughtmare: There is also of course the opposite, for example matching `forall f. f x` to `(\y -> foo y) x`, but I assume eta reduction is not so much of a problem?

Other than that I think your suggestion is good.

simonpj: > There is also of course the opposite, for example matching forall f. f x to (\y -> foo y) x, but I assume eta reduction is not so much of a problem?

Well, the pattern is a HOP, so HOP-matching would yield
```
  f :-> etaReduce(   \x.   (\y. foo y) x )
={do the eta-reduction}
  fo :-> \x.foo y
```
Just the same as non-HOP matching.  What's the problem here?

noughtmare: First of all, a small typo is that it should be `f :-> \y. foo y`, but I thought you meant with `etaReduce` that it should do as much eta reduction as possible, so it would become `f :-> foo` in end, thus reducing too far.

If `etaReduce` only does one step then there is another problem if we have multiple HOP parameters, such as when we match `forall f. f x y` to `foo x y`, which should produce `f :-> foo`, but with only one eta reduction we will get `f :-> \x. foo x`.

simonpj: > so it would become f :-> foo in the end, thus reducing too far

Ah yes, how true.

What about:

"Matching of higher order patterns (HOP-matching). A higher order pattern `f x y z` (in the template) matches any target expression `e` provided:

1. The target has the same type as the template
2. No local binder is free in e, other than x, y, z.
3. The target `e` is not of form `(e' z)`, where `z` is not free in `e'`

If these three conditions hold,..."

Then we still need to explain (3), giving justification you give above.

Might you do that?  Joachim has merged the proposal but we can still make a new PR to clarify it.

simonpj:
This proposal has been implemented.  I'll add the implemented tag.

PR #556, opened 2022-11-17
Type arguments in constructor patterns: fix signature scoping

int-index:
The amendment has been accepted; the following discussion is mostly of historic interest.

----

At the term-level, we reject nonlinear variable bindings:
```haskell
k (x, x) = ()
-- Error: Conflicting definitions for ‘x’
```

However, pattern signatures are exempt from this restriction:
```haskell
u (Nothing :: Maybe (x, x)) = ()
```

Now consider the following program:
```haskell
data Proxy a = P
g :: Proxy (Nothing @(a, a)) -> ()
g (P @(Nothing :: Maybe (x, x))) = ()
```

Should we accept or reject `g`? Is it more like `u` or like `k`? Since multiple occurrences of `x` are on the right-hand side of `::`, one might expect it to be treated like `u`. Unfortunately, the proposal fails to take this into account and states the following:

> Any type variables mentioned in a type application are considered
> binding sites, shadowing any in-scope type variables.

Presumably, "any type variables" includes type variables in pattern signatures, so we would end up rejecting `g`. This, however, is in direct contradiction with the proposal's stated goal to uphold the VOP (Visibility Orthogonality Principle).

The fact that the pattern signature in `g` is hidden under a `@` should not affect its scoping rules! This amendment corrects this oversight and restores the VOP.

### DIFF
--- a/proposals/0448-type-variable-scoping.rst
+++ b/proposals/0448-type-variable-scoping.rst
@@ -414,7 +414,23 @@ Proposed Change Specification
    (Right now, pattern synonyms require all such quantifications to occur before any term arguments,
    but accepted proposal `#402`_ allows these quantifications to occur in any order in data constructors.)
 
-#. Any type variables mentioned in a type application are considered binding sites, shadowing any in-scope type variables.
+#. In accordance with the `Visibility Orthogonality Principle`_,
+   the rules that determine whether a variable occurrence is a binding site or a use site are not affected by the presence of a ``@``.
+   At the time of writing, the rules for patterns and pattern signatures are as follows:
+
+   * Outside pattern signatures,
+     variable occurrences are considered binding sites,
+     shadowing any other in-scope variables.
+     It is an error to bring the same type variable into scope in two (or more) places within the same match group.
+
+   * Inside pattern signatures (i.e. on the right-hand side of ``pat :: sig``),
+     occurrences of in-scope type variables are usages,
+     whereas occurrences of out-of-scope type variables create implicit bindings.
+     It is allowed to mention the same out-of-scope variable more than once.
+
+   Generalize these rules to apply not only to pattern signatures but also to
+   kind signatures in type applications in constructor patterns
+   (and to kind signatures in type variable patterns defined in the "Type arguments in lambda patterns" section).
 
 #. Typing follows the rules in `Type Variables in Patterns`_.
    In particular, see Figure 7, which we modify here in two ways:
@@ -428,8 +444,6 @@ Proposed Change Specification
 #. A wildcard ``_`` as a type argument says simply to skip that argument;
    it does not trigger any behavior associated with partial type signatures.
    In particular, ``-XPartialTypeSignatures`` is not necessary, and no diagnostic is produced.
-
-#. As with term variables, it is an error to bring the same type variable into scope in two (or more) places within the same pattern.
 
 Examples
 ~~~~~~~~
@@ -447,7 +461,7 @@ Here is an example (taken from `#15050 <https://gitlab.haskell.org/ghc/ghc/issue
 This should type-check, because the following code does::
 
     foo :: T Bool -> ()
-    foo (MkT (_ :: Int _)) = ()
+    foo (MkT (_ :: Int)) = ()
 
 Note that the data constructor expects up-to two type arguments (``forall b a.…``), but we are passing only one type argument, which then corresponds to the *first* type argument of of the data constructor.
 
@@ -499,6 +513,26 @@ This shadowing behavior mimics what happens with term variables in patterns.
 
 This is accepted.
 The type variable ``a`` is bound to ``Int``, by pattern-matching.
+
+Here is an example of pattern signatures within a type abstraction in a pattern::
+
+   data Proxy a = P
+   g2 :: Proxy (Nothing @(a, a)) -> ()
+   g2 (P @(Nothing :: Maybe (t, t))) = ()
+
+Note multiple occurrences of ``t`` in the pattern. Normally, we would disallow
+multiple bindings of a single variable::
+
+   f1 (P x) (P x) = x               -- Rejected (multiple bindings of ‘x’)
+   f2 (P @a x) (P @a y) = x         -- Rejected (multiple bindings of ‘a’)
+
+Pattern and kind signatures, however, are not subject to this restriction,
+since variable occurrences in pattern signatures are considered usages (not bindings)::
+
+   g1 (P x :: Proxy (a,a)) = x               -- Accepted (multiple occurrences of ‘a’ notwithstanding)
+
+   g2 :: Proxy (Nothing @(a, a)) -> ()
+   g2 (P @(Nothing :: Maybe (t, t))) = ()    -- Accepted (multiple occurrences of ‘t’ notwithstanding)
 
 Effects
 ~~~~~~~
@@ -580,6 +614,11 @@ Proposed Change Specification
    Conveniently, ``apat``\ s are used both in function left-hand sides and in lambda-expressions, so this change covers both use-cases.
 
    (Note that this does not subsume the new grammar for constructor patterns, which allow *types*, not just variables.)
+
+#. In accordance with the `Visibility Orthogonality Principle`_,
+   the rules that determine whether a variable occurrence is a binding site or a use site are not affected by the presence of a ``@``.
+   That is, name resolution in kind signatures in type variable patterns follows the rules for pattern signatures.
+   (The rules for pattern signatures are given in the "Type arguments in constructor patterns" section).
 
 #. A type variable pattern is not allowed in the following contexts:
 


proposals/0448-type-variable-scoping.rst
```diff
@@ -502,6 +509,15 @@ This shadowing behavior mimics what happens with term variables in patterns.
 
 This is accepted. The type variable ``a`` is bound to ``Int``, by pattern-matching.
 
+Here is an example of pattern signatures within a type abstraction in a pattern:
```
Ericson2314, line 512:
```suggestion
Here is an example of pattern signatures within a type abstraction in a pattern::
```

int-index:
Ping @goldfirere as the proposal author. Do you agree this is a fix?

Ericson2314 (approved):
I agree with this fix.

I think this is a pure oversight because the original paper didn't have non trivial kinds, nor `@(_ :: _)` in the grammar, so it simply didn't discuss this case at all.

simonpj:
You deleted "As with term variables, it is an error to bring the same type variable into scope in two (or more) places within the same pattern."  But why?  The example below
```
foo (MkT5 @_ @x @x)    = (() :: x ~ x => ())    -- not accepted
```
illustrates this principle, and it's a direct outworking of the VOP.  Why did you delete this?

simonpj:
You deleted "Any type variables mentioned in a type application are considered binding sites, shadowing any in-scope type variables."  But the following is true isn't it?

* Any type variables mentioned in a type application, **outside a pattern signature**, are considered binding sites, shadowing any in-scope type variables.

And it remains important, doesn't it? 

Perhaps you deleted it because it is a mere consequence of the VOP?  Both this point and the previous one are just outworking of the VOP, but I think it's a pity to lose them.  Make them sub-bullets of the VOP if you like.

simonpj:
Your example
```
data Proxy a = P
g :: Proxy (P @(a, a)) -> ()
g (P @(P :: Proxy (x, x))) = ()
```
is quite a subtle one.  g's type `Proxy (P @(a,a)) -> ()` uses `P` (a data constructor of `Proxy`) as a type argument to the type constructor of `Proxy`.  That's jolly confusing.  Could you use different data types for the two purposes?  Perhaps `g :: Proxy (Nothing @(a,a)) -> ()`.  That would be a notch less brain melting.

int-index:
> You deleted "As with term variables, it is an error to bring the same type variable into scope in two (or more) places within the same pattern." But why? 

I only moved it slightly, it's still there in the bullet that starts with "Outside pattern signatures,".

> You deleted "Any type variables mentioned in a type application are considered binding sites, shadowing any in-scope type variables." But the following is true isn't it?

I only moved it slightly, it's still there in the bullet that starts with "Outside pattern signatures,".

> The amendment appears to be two deletions, of sentences which are correct (or nearly so, see above).

Those sentences were moved, not deleted, to make it clear that they are only valid outside pattern signatures.

> Could you use different data types for the two purposes? Perhaps `g :: Proxy (Nothing @(a,a)) -> ()`.

Sure, I'll do that.

simonpj:
Ah I misunderstood.  I thought that VOP para was there before --- but in fact it's entirely new.  Thanks.

int-index:
Ping @goldfirere.

int-index:
@nomeata I don't know if @goldfirere actually has time to look at this, shall we put this through the usual proposal process? I'd like to submit to the committee.

nomeata:
I hope @goldfirere  has, because he’s the natural shepherd for this  :-)

simonpj:
> Normally, we would disallow a nonlinear binding such as (x, x), but pattern signatures are currently exempt from this restriction.

It's hard to know what "Normally" means.  I think it'd be better to be more direct, perhaps by adding a second binding to contrast with it. Somethign like this:

```
data Proxy a = P

f1 (P x) (P x) = x

f2 (P @a x) (P @a y) = x

g1 (P x :: Proxy (a,a)) = x

g2 :: Proxy (Nothing @(a, a)) -> ()
g2 (P @(Nothing :: Maybe (x, x))) = ()
```
Here, 
* `f1` is rejected because `x` is bound twice in a single LHS
* `f2` is rejected for the same reason: `a` is bound twice in a single LHS
* `g1` is accepted desepite the repeated `a`, because pattern signatures are exempted from the "variable bound twice" restriction.
* `g2` is accepted for the same reason.  The difference is that this time the pattern signature is inside a visible type argument in the pattern, but that does not affect scoping.

simonpj:
I'm in support.  I'd like to see the above clarification in the examples section though.

Ericson2314:
@nomeata If he isn't, @aspiwack would also be a very natural shepherd because he shepherded the original proposal and also is currently shepherding my https://github.com/ghc-proposals/ghc-proposals/pull/532.

proposals/0448-type-variable-scoping.rst
```diff
@@ -414,7 +414,19 @@ Proposed Change Specification
    (Right now, pattern synonyms require all such quantifications to occur before any term arguments,
    but accepted proposal `#402`_ allows these quantifications to occur in any order in data constructors.)
 
-#. Any type variables mentioned in a type application are considered binding sites, shadowing any in-scope type variables.
+#. In accordance with the `Visibility Orthogonality Principle`_,
+   the rules that determine whether a variable occurrence is a binding site or a use site are not affected by the presence of a ``@``.
+   At the time of writing, the rules are as follows:
+
+   * Outside pattern signatures,
```
goldfirere, line 421: I think we must be very careful around the term "pattern signature", because different people have different understandings of this term, and this has led to confusion elsewhere in this space.

I claim that a pattern signature is the `ty` in the grammar form `(pat :: ty)`, when that occurs in a pattern (as it must).

However, I have seen it claimed that the ty in `data T (a :: ty)` is a pattern signature. The example in this PR describes the `ki` in `(ty :: ki)` as a pattern signature, when it occurs in a visible type application in a pattern. These cases are not, to me, signatures on patterns, and are thus not pattern signatures.

One relevant nuance here is that these two are grammatically different:

```
x1 (MkT (a :: ty))
x2 (MkT @(a :: ty))
```

In `x1`, the `a` is a pattern, and thus `ty` is a pattern signature. In `x2`, the `a` is a type -- not a pattern, as we don't have type patterns in our grammar -- and thus (to me) the `ty` in `x2` is not a pattern signature. Perhaps the VOP suggests we should introduce type patterns in our grammar in order to avoid this nuance, but I don't think that doing so is worthwhile.

I'm not disagreeing with what I understand as the intended substance of this proposal, but I do object to the use of "pattern signature" to describe something that is not grammatically a signature of a pattern.

proposals/0448-type-variable-scoping.rst
```diff
@@ -414,7 +414,19 @@ Proposed Change Specification
...
+     variable occurrences are considered binding sites,
+     shadowing any other in-scope variables.
+     It is an error to bring the same type variable into scope in two (or more) places within the same match group.
+
+   * Inside pattern signatures (i.e. on the right-hand side of ``pat :: sig``),
+     occurrences of in-scope type variables are usages,
+     whereas occurrences of out-of-scope type variables create implicit bindings.
+     It is allowed to mention the same out-of-scope variable more than once.
```
goldfirere, line 429: This sentence doesn't have the intended payload, I think. The first such occurrence will bring the variable into scope, so the second occurrence is one of an out of scope variable.

goldfirere (commented):
Other than the language around "pattern signatures", I'm in support. Yes, this was an oversight on my part.

(Re "time to read": Yes! I'm managing to be able to keep up with this all, but maybe not quite as quick as before.)

proposals/0448-type-variable-scoping.rst
int-index: Would it be acceptable to say "pattern or kind signatures"?

proposals/0448-type-variable-scoping.rst
int-index: > The first such occurrence will bring the variable into scope

Not according to the definitions given here. The preceding sentence states quite clearly that occurrences of out-of-scope type variables create _implicit_ bindings. If the first occurrence brought the variable into scope, that would make it an explicit binding.

So I think the sentence says what I wanted it to say. Maybe I'm missing something.

goldfirere: OK.

proposals/0448-type-variable-scoping.rst
goldfirere: Yes, I think so.

int-index: Actually, those two paragraphs refer to pattern signatures specifically, since they describe the existing rules (the preceding sentence is "at the moment of writing, the rules are..."). Therefore, no change required. It is implied that we take the existing rules and generalize them in such a way that VOP is made to hold. For clarity, I stated so explicitly in a separate paragraph that follows. I also cross-referenced it from the section on type lambdas.

Ericson2314: For what it's worth, I was the person calling more things pattern signatures and I have stopped doing that. I am not sure anyone but me made that mistake.

goldfirere:
I'm happy with the new wording and would support this in a committee vote.

simonpj:
I'm in supporrt.  This proposal makes a good clarificaiton.

Ericson2314:
@nomeata it looks like this is an an "unofficially voted on" limbo state?

nomeata:
It's still pending Richard to make an recommendation on the mailing list so that the full committee will pay attention, I believe.

int-index:
@goldfirere This one awaits action from you. Do you intend to bring this before the committee (or simply merged as an amendment)?

goldfirere:
I declare this to be accepted by the committee.

PR #302, opened 2019-11-29
Lambda expressions with guards and multiple clauses (was: `\ of`, -XMultiWayLambda)

JakobBruenker:
This proposal introduces a new extension `-XMultiWayLambda` that allows a new expression introduced by `\ of`, similar to lambdas but with guards and multiple clauses, enabled by implicit layout.

[Rendered](https://github.com/JakobBruenker/ghc-proposals/blob/patch-1/proposals/0000-cases.md)

PR #559, opened 2022-11-23
GHC2023

nomeata:
[Rendered](https://github.com/ghc-proposals/ghc-proposals/blob/joachim/ghc2023/proposals/0000-ghc2023.rst)

This PR will guide the process of deciding if we want to have, and then defining, GHC2023, as per the process outlined in [#372](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0372-ghc-extensions.rst).

### DIFF
--- a/proposals/0000-ghc2023.rst
+++ b/proposals/0000-ghc2023.rst
@@ -98,7 +98,7 @@ this one probably isn’t yet.
 Alternatives
 ------------
 We could not do ``GHC2023`` and wait another year, or more, because we shy away from
-making what may look like a stability-threatning change.
+making what may look like a stability-threatening change.
 In my view that is worse: The fixes and improvements suggested above would reach our users later, we would not establish a regular and predicatable pattern, and in the worst case never dare to make a new release, which would make the ``GHC20xx`` idea fall into a similar pattern than the ``Haskell20xx`` report process, which at the moment is stalled.
 
 
--- a/proposals/0000-ghc2023.rst
+++ b/proposals/0000-ghc2023.rst
@@ -162,9 +162,9 @@ Why add RoleAnnotations?
 
 Roles are essential part of modern GHC Haskell.
 
-As `GeneralisedNewtypeDeriving` is in GHC2021 language, so should be `RoleAnnotations`. They are different sides of the same feature: without correct role annotations GND cannot be used safely.
+As `GeneralisedNewtypeDeriving`_ is in the GHC2021 language set, so should be `RoleAnnotations`_. They are different sides of the same feature: without correct role annotations GND cannot be used safely.
 
-Role annotations are required for correctly writing types with internal invariants like `Set` or "fast" implementations like `data Fin (n :: Nat) = UnsafeFin Int`.
+Role annotations are required for correctly writing types with internal invariants like ``Set`` or "fast" implementations like ``data Fin (n :: Nat) = UnsafeFin Int``.
 
 Alternatives
 ------------
@@ -187,3 +187,4 @@ I (Joachim) can prepare a MR.
 .. _LambdaCase: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/lambda_case.html#extension-LambdaCase
 .. _MonoLocalBinds: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/let_generalisation.html#extension-MonoLocalBinds
 .. _RoleAnnotations: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/roles.html#extension-RoleAnnotations
+.. _GeneralisedNewtypeDeriving: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/roles.html#extension-GeneralisedNewtypeDeriving
--- a/proposals/0000-ghc2023.rst
+++ b/proposals/0000-ghc2023.rst
@@ -26,6 +26,7 @@ The ``GHC2023`` language extension set comprises of all language extensions
 that ``GHC2021`` comprises, plus the following
 
 * `ExplicitNamespaces`_
+* `GADTs`_ (and the implied `MonoLocalBinds`_)
 * `LambdaCase`_
 
 
@@ -81,6 +82,65 @@ recently, e.g. in `#281 <https://github.com/ghc-proposals/ghc-proposals/pull/281
 thus may not yet be as stable as we want them for ``GHC20xx``.
 
 
+Why add GADTs and MonoLocalBinds?
+---------------------------------
+
+``GHC2021`` includes both `GADTSyntax`_ and `ExistentialQuantification`_, but
+does not include `GADTs`_ or `MonoLocalBinds`_.  Moreover, the combination of
+``GADTSyntax`` and ``ExistentialQuantification`` is enough to define GADTs and
+pattern match on them (see `GHC issue #21102
+<https://gitlab.haskell.org/ghc/ghc/-/issues/21102>`_ for detailed discussion).
+
+GHC 9.4 and later permits pattern-matching on an imported GADT regardless of
+which extensions are enabled, but doing so will emit a warning from
+``-Wgadt-mono-local-binds`` if ``MonoLocalBinds`` is disabled.  This is
+consistent with the principle that extensions are required at definition sites
+but not use sites.  (GHC 9.2 and previous versions required ``GADTs`` or
+``TypeFamilies`` to be enabled in order to pattern match on a GADT.)
+
+Enabling ``MonoLocalBinds`` is considered necessary for robust type inference
+when pattern matching on GADTs (see section 4.2 of `OutsideIn(X): Modular type
+inference with local assumptions
+<https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/jfp-outsidein.pdf>`_).
+Moreover, writing type signatures for polymorphic local bindings generally makes
+it easier to understand the code.  However, the type signature requirement makes
+it more difficult to factor out repeated code into a ``where`` clause (e.g. see
+`GHC issue #19396 <https://gitlab.haskell.org/ghc/ghc/-/issues/19396>`_), and
+this can surprise users and cause backwards incompatibility.
+
+Since ``ExistentialQuantification`` allows defining types with contexts that
+include equality constraints, there is not really a principled distinction
+between ``ExistentialQuantification`` and ``GADTs``.  (While there is a
+syntactic distinction between GADT syntax and "traditional" datatype syntax,
+both forms are capable of expressing simple ADTs, existentially quantified
+types, and GADTs.)
+
+Possible ways to resolve this infelicity include:
+
+* Add ``GADTs`` and ``MonoLocalBinds`` to ``GHC2023``.  This makes it clear
+  that GADTs/existentials are a core part of the language, and makes the type
+  inference compromises necessary to accommodate them.  Migration advice for
+  ``GHC2023`` should make clear that type signatures may need to be added for
+  local bindings (or ``NoMonoLocalBinds`` specified explicitly).  Given that
+  ``MonoLocalBinds`` is a simpler design which can safely be extended with
+  ``GADTs``, it makes sense to have it be part of the base language; users can
+  then opt-in explicitly to ``NoMonoLocalBinds`` as an extension if required.
+
+* Add ``GADTs`` but not ``MonoLocalBinds``.  This is mostly consistent with
+  ``GHC2021``, but means that type inference for local bindings may not be
+  predictable when using GADTs.  Moreover, including an extension but not the
+  extensions it implies is itself confusing (as with
+  ``TypeOperators``/``ExplicitNamespaces`` in ``GHC2021``).
+
+* Remove ``ExistentialQuantification`` from ``GHC2023``.  This means
+  existentials/GADTs are clearly treated as an extension, albeit an extension
+  that makes type inference "worse". Users will need to understand the impact
+  of ``MonoLocalBinds`` only if they import a GADT or define one after enabling
+  ``GADTs`` explicitly. Migration advice for ``GHC2023`` should make clear that
+  users may need to enable ``GADTs`` explicitly (and possibly give type
+  signatures for local bindings or specify ``NoMonoLocalBinds``).
+
+
 Why add LambdaCase?
 -------------------
 
@@ -110,7 +170,10 @@ The actual changes to GHC to add a new language edition are small, and likely
 I (Joachim) can prepare a MR.
 
 .. _ExplicitNamespaces: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/explicit_namespaces.html#extension-ExplicitNamespaces
+.. _ExistentialQuantification: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/existential_quantification.html#extension-ExistentialQuantification
+.. _GADTs: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/gadt.html#extension-GADTs
+.. _GADTSyntax: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/gadt_syntax.html#extension-GADTSyntax
 .. _TypeOperators: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/type_operators.html#extension-TypeOperators
 .. _OverloadedStrings: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/overloaded_strings.html#extension-OverloadedStrings
 .. _LambdaCase: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/lambda_case.html#extension-LambdaCase
-
+.. _MonoLocalBinds: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/let_generalisation.html#extension-MonoLocalBinds
--- a/proposals/0000-ghc2023.rst
+++ b/proposals/0000-ghc2023.rst
@@ -79,8 +79,11 @@ release.
 
 A counter-argument to `ExplicitNamespaces`_ is that it has seen changes
 recently, e.g. in `#281 <https://github.com/ghc-proposals/ghc-proposals/pull/281>`_ and
-thus may not yet be as stable as we want them for ``GHC20xx``.
+thus may not yet be as stable as we want them for ``GHC20xx``. To keep GHC20xx
+stable we could amend #281 to ask for a new extenion name for syntax added there.
 
+__Backward compatibility__: Enabling `ExplicitNamespaces`_ does not affect
+existing ``GHC2021`` code.
 
 Why add GADTs and MonoLocalBinds?
 ---------------------------------
@@ -141,6 +144,9 @@ Possible ways to resolve this infelicity include:
   signatures for local bindings or specify ``NoMonoLocalBinds``).
 
 
+__Backward compatibility__: Enabling ``GADTs`` seems to not break existing ``GHC2021`` code,
+but enabling ``MonoLocalBinds`` could does.
+
 Why add LambdaCase?
 -------------------
 
@@ -156,6 +162,12 @@ A counter-agument to adding `LambdaCase` is that just extended the meaning of
 wants to add extensions to ``GHC20xx`` that have been proven to be stable, then
 this one probably isn’t yet.
 
+__Backward compatibility__: Enabling `LambdaCase`_ does not affect
+existing ``GHC2021`` code, with the exception of lambda-bound variable names
+``casess`` (GHC already forbids the ``\case`` even without ``-XLambdaCase``).
+One could argue that this is ok (just as it was ok for GHC to break the use of ``\case``
+at some point in the past.)
+
 
 Why add RoleAnnotations?
 ------------------------
@@ -165,6 +177,9 @@ Roles are essential part of modern GHC Haskell.
 As `GeneralisedNewtypeDeriving`_ is in the GHC2021 language set, so should be `RoleAnnotations`_. They are different sides of the same feature: without correct role annotations GND cannot be used safely.
 
 Role annotations are required for correctly writing types with internal invariants like ``Set`` or "fast" implementations like ``data Fin (n :: Nat) = UnsafeFin Int``.
+
+__Backward compatibility__: Enabling `LambdaCase`_ does not affect
+existing ``GHC2021`` code.
 
 Alternatives
 ------------
--- a/proposals/0000-ghc2023.rst
+++ b/proposals/0000-ghc2023.rst
@@ -164,7 +164,7 @@ this one probably isn’t yet.
 
 __Backward compatibility__: Enabling `LambdaCase`_ does not affect
 existing ``GHC2021`` code, with the exception of lambda-bound variable names
-``casess`` (GHC already forbids the ``\case`` even without ``-XLambdaCase``).
+``cases`` (GHC already forbids the ``\case`` even without ``-XLambdaCase``).
 One could argue that this is ok (just as it was ok for GHC to break the use of ``\case``
 at some point in the past.)
 
@@ -178,7 +178,7 @@ As `GeneralisedNewtypeDeriving`_ is in the GHC2021 language set, so should be `R
 
 Role annotations are required for correctly writing types with internal invariants like ``Set`` or "fast" implementations like ``data Fin (n :: Nat) = UnsafeFin Int``.
 
-__Backward compatibility__: Enabling `LambdaCase`_ does not affect
+__Backward compatibility__: Enabling `RoleAnnotations`_ does not affect
 existing ``GHC2021`` code.
 
 Alternatives
--- a/proposals/0000-ghc2023.rst
+++ b/proposals/0000-ghc2023.rst
@@ -14,7 +14,7 @@ GHC2023
 Motivation
 ----------
 
-The `GHC20xx process proposal <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0372-ghc-extensions.rst#motivation>`_ motivates why we feel the need for ``GHC20xx`` language editoins in general.
+The `GHC20xx process proposal <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0372-ghc-extensions.rst#motivation>`_ motivates why we feel the need for ``GHC20xx`` language editions in general.
 
 The first edition, ``GHC2021``, is now two years old, and I propose we define
 ``GHC2023`` now, even if turns out to be a modest change, to keep the process alive.
@@ -40,18 +40,18 @@ instability.
 
 I believe that neither are good reasons to not define ``GHC2023``:
 
-* There is an inherent latecy between defining language editions and them
+* There is an inherent latency between defining language editions and them
 reaching the users. It reaches users who do not have to worry about supporting
 older GHC first, and thus library authors later. This is somewhat unavoidable,
 but not per se a reason to reduce the frequency.
 
 * Relatively frequent language editions do not introduce undue instability:
 Users who pin the language edition in their `.cabal` file or so are not
-affected by the exitance of a new one. Neither are users who pin their GHC
+affected by the existence of a new one. Neither are users who pin their GHC
 version. Only those users who upgrade their version of GHC _and_ ask for the
 latest edition may now have access to new features.
 .
-As `Arnaud explained on the mailing list <https://mail.haskell.org/pipermail/ghc-steering-committee/2022-November/002949.html>`_, Rust (generally not perceived as a languge with a perception of low stability) has a very similar model with their “language editions”.
+As `Arnaud explained on the mailing list <https://mail.haskell.org/pipermail/ghc-steering-committee/2022-November/002949.html>`_, Rust (generally not perceived as a language with a perception of low stability) has a very similar model with their “language editions”.
 
 Furthermore, a regular, not too slow cadence makes the whole process more
 predictable, which I expect *improves* the perception of stability, in the
@@ -64,7 +64,7 @@ Why add ExplicitNamespaces?
 As discussed in `issue #551
 <https://github.com/ghc-proposals/ghc-proposals/issues/551>`_, ``GHC2021``
 includes `TypeOperators`_, but does not include `ExplicitNamespaces`_. This was
-a very strange (and probably inadvertant) decision, given that the flag
+a very strange (and probably inadvertent) decision, given that the flag
 ``-XTypeOperators`` enables both the `TypeOperators`_ and the
 `ExplicitNamespaces`_ language extension.
 
--- a/proposals/0000-ghc2023.rst
+++ b/proposals/0000-ghc2023.rst
@@ -86,8 +86,7 @@ The latest `State of Haskell 2021 Survey results
 default”. It also missed ``GHC2021`` by just two votes. There is a whole style
 of writing Haskell that makes extensive use of ``\case``. And (unlike the runner up in the survey, `OverloadedStrings_`), it only enables *new* syntax, i.e. it does not change existing code.
 
-
-Why no add LambdaCase?
+Why not add LambdaCase?
 ----------------------
 
 We just extended the meaning of `LambdaCase`_ with ``\cases`` in `#302
--- a/proposals/0000-ghc2023.rst
+++ b/proposals/0000-ghc2023.rst
@@ -1,0 +1,109 @@
+GHC2023
+=======
+
+.. author:: Joachim Breitner
+.. date-accepted::
+.. ticket-url::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+.. sectnum::
+.. contents::
+
+
+Motivation
+----------
+
+The `GHC20xx process proposal <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0372-ghc-extensions.rst#motivation>`_ motivates why we feel the need for ``GHC20xx`` language editoins in general.
+
+The first edition, ``GHC2021``, is now two years old, and I propose we define
+``GHC2023`` now, even if turns out to be a modest change, to keep the process alive.
+
+Proposed Change Specification
+-----------------------------
+
+The ``GHC2023`` language extension set comprises of all language extensions
+that ``GHC2021`` comprises, plus the following
+
+* `ExplicitNamespaces`_
+* `LambdaCase`_
+
+
+(*The committe will vote on each extension separately. Until then, this is the
+list of nominated extensions, and will be updated after a vote.*)
+
+Why GHC2023 in the first place
+------------------------------
+There were concerns that defining ``GHC2023`` is too soon, given that
+``GHC2021`` has hardly reached the target audience, and that this introduces
+instability.
+
+I believe that neither are good reasons to not define ``GHC2023``:
+
+* There is an inherent latecy between defining language editions and them
+reaching the users. It reaches users who do not have to worry about supporting
+older GHC first, and thus library authors later. This is somewhat unavoidable,
+but not per se a reason to reduce the frequency.
+
+* Relatively frequent language editions do not introduce undue instability:
+Users who pin the language edition in their `.cabal` file or so are not
+affected by the exitance of a new one. Neither are users who pin their GHC
+version. Only those users who upgrade their version of GHC _and_ ask for the
+latest edition may now have access to new features.
+.
+As `Arnaud explained on the mailing list <https://mail.haskell.org/pipermail/ghc-steering-committee/2022-November/002949.html>`_, Rust (generally not perceived as a languge with a perception of low stability) has a very similar model with their “language editions”.
+
+Furthermore, a regular, not too slow cadence makes the whole process more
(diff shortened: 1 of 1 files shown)


phadej:
I'll note that `LambdaCase` can be considered *unstable* extension, as it changed quite recently: https://github.com/ghc-proposals/ghc-proposals/pull/302 `\cases` were added in GHC-9.4 to `LambdaCase` extension.

There isn't real experience of `\cases` use in the wild, but if added to GHC2023 we should be quite sure there won't be some small tweaks or/and missed things.

nomeata:
Yes, that’s exactly what I wrote in section “[Why not add LambdaCase?](https://github.com/ghc-proposals/ghc-proposals/blob/joachim/ghc2023/proposals/0000-ghc2023.rst#id7)”…

proposals/0000-ghc2023.rst
```diff
@@ -0,0 +1,114 @@
...
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/559>`_.
+.. sectnum::
+.. contents::
+
+
+Motivation
+----------
+
+The `GHC20xx process proposal <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0372-ghc-extensions.rst#motivation>`_ motivates why we feel the need for ``GHC20xx`` language editoins in general.
```
tek, line 17:
```suggestion
The `GHC20xx process proposal <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0372-ghc-extensions.rst#motivation>`_ motivates why we feel the need for ``GHC20xx`` language editions in general.
```

proposals/0000-ghc2023.rst
```diff
@@ -0,0 +1,114 @@
...
+
+Why GHC2023 in the first place
+------------------------------
+There were concerns that defining ``GHC2023`` is too soon, given that
+``GHC2021`` has hardly reached the target audience, and that this introduces
+instability.
+
+I believe that neither are good reasons to not define ``GHC2023``:
+
+* There is an inherent latecy between defining language editions and them
```
tek, line 43:
```suggestion
* There is an inherent latency between defining language editions and them
```

proposals/0000-ghc2023.rst
```diff
@@ -0,0 +1,114 @@
...
+reaching the users. It reaches users who do not have to worry about supporting
+older GHC first, and thus library authors later. This is somewhat unavoidable,
+but not per se a reason to reduce the frequency.
+
+* Relatively frequent language editions do not introduce undue instability:
+Users who pin the language edition in their `.cabal` file or so are not
+affected by the exitance of a new one. Neither are users who pin their GHC
```
tek, line 50:
```suggestion
affected by the existence of a new one. Neither are users who pin their GHC
```

proposals/0000-ghc2023.rst
```diff
@@ -0,0 +1,114 @@
...
+version. Only those users who upgrade their version of GHC _and_ ask for the
+latest edition may now have access to new features.
+.
+As `Arnaud explained on the mailing list <https://mail.haskell.org/pipermail/ghc-steering-committee/2022-November/002949.html>`_, Rust (generally not perceived as a languge with a perception of low stability) has a very similar model with their “language editions”.
```
tek, line 54:
```suggestion
As `Arnaud explained on the mailing list <https://mail.haskell.org/pipermail/ghc-steering-committee/2022-November/002949.html>`_, Rust (generally not perceived as a language with a perception of low stability) has a very similar model with their “language editions”.
```

proposals/0000-ghc2023.rst
```diff
@@ -0,0 +1,114 @@
...
+sense of delivering a stable flow of changes.
+
+
+Why add ExplicitNamespaces?
+---------------------------
+
+As discussed in `issue #551
+<https://github.com/ghc-proposals/ghc-proposals/issues/551>`_, ``GHC2021``
+includes `TypeOperators`_, but does not include `ExplicitNamespaces`_. This was
+a very strange (and probably inadvertant) decision, given that the flag
```
tek, line 67:
```suggestion
a very strange (and probably inadvertent) decision, given that the flag
```

proposals/0000-ghc2023.rst
```diff
@@ -0,0 +1,114 @@
...
+We just extended the meaning of `LambdaCase`_ with ``\cases`` in `#302
+<https://github.com/ghc-proposals/ghc-proposals/pull/302>`_, and if one only
+wants to add extensions to ``GHC20xx`` that have been proven to be stable, then
+this one probably isn’t yet.
+
+
+Alternatives
+------------
+We could not do ``GHC2023`` and wait another year, or more, because we shy away from
+making what may look like a stability-threatning change.
```
tek, line 101:
```suggestion
making what may look like a stability-threatening change.
```

proposals/0000-ghc2023.rst
```diff
@@ -0,0 +1,114 @@
...
+In my view that is worse: The fixes and improvements suggested above would reach our users later, we would not establish a regular and predicatable pattern, and in the worst case never dare to make a new release, which would make the ``GHC20xx`` idea fall into a similar pattern than the ``Haskell20xx`` report process, which at the moment is stalled.
```
tek, line 102:
```suggestion
In my view that is worse: The fixes and improvements suggested above would reach our users later, we would not establish a regular and predictable pattern, and in the worst case never dare to make a new release, which would make the ``GHC20xx`` idea fall into a similar pattern than the ``Haskell20xx`` report process, which at the moment is stalled.
```

phadej:
> and any possible changes to that in the future.

I wrote my comment to emphasize that if some new LambdaCase-ish feature would be introduced in the future, it should be then under a new extension, to keep GHC2023 stable. The same reasoning applies as for not adding `ExplicitNamespaces` to GHC2021 (which afaik extends the set of accepted programs, where `\cases` may clash e.g. in some DSL implementation, there is some occurences on Hackage! https://hackage-search.serokell.io/?q=%5C%5Ccases e.g. in JsContracts).

Concretely if `LambdaCase` were in GHC2021, I'd expected #302 to introduce new extension.

JakobBruenker:
> which afaik extends the set of accepted programs, where \cases may clash

Worth noting that this is also true of `PolyKinds`, which was included in GHC2021 (hence [this corresponding mention in the migration guide](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.2#polykinds))

phadej:
> > which afaik extends the set of accepted programs, where \cases may clash
> 
> Worth noting that this is also true of `PolyKinds`, which was included in GHC2021 (hence [this corresponding mention in the migration guide](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.2#polykinds))

No, I was talking about different thing. To repeat myself: if `GHC2021` had `LambdaCase`s in it, the addition of `\cases` would changed the `GHC2021` language. Therefore it's better to be sure that `LambdaCase` is "done". And also the `\cases` part is actually stable. 

The latter part is discussed in the proposal as @nomeata pointed out.

FWIW, I'd suggest a single section for each suggested extension with subsections for "why to include" and "why not to include". If it is hard to come up with any reasons to not include `ExplicitNamespaces` it will be good to write that explicitly.

JakobBruenker:
> if LambdaCase was in GHC2021 before that

If I recall correctly, the upcoming `\cases` was part of the reason `LambdaCase` didn't make it into GHC2021.

> just-add-whitespace

Adding whitespaces isn't enough, since both `\ case` and `\ cases` trigger `LambdaCase`, though there are [a few easy ways to fix it](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.4#cases-is-parsed-differently-when-xlambdacase-is-enabled)

nomeata:
> Adding whitespaces isn't enough

I stand corrected. I guess `LambdaCase` is older than the let's-get-whitespace-sensitive trend

proposals/0000-ghc2023.rst
```diff
@@ -0,0 +1,116 @@
...
+It seems to be a bad idea to retroactive *change* ``GHC2021`` to fix this (we
+do care about stability after all), but we should certainly fix this in the
+upcoming edition.
+
+And – at least if one accepts that regular releases of ``GHC20xx`` are a Good
+Thing™, adding `ExplicitNamespaces`_ alone should be sufficient to cut a new
+release.
+
+A counter-argument to `ExplicitNamespaces`_ is that it has seen changes
+recently, e.g. in `#281 <https://github.com/ghc-proposals/ghc-proposals/pull/302>_` and
```
adamgundry, line 80:
```suggestion
A counter-argument to `ExplicitNamespaces`_ is that it has seen changes
recently, e.g. in `#281 <https://github.com/ghc-proposals/ghc-proposals/pull/281>`_ and
```

Incidentally, I find it very surprising that #281 modified `ExplicitNamespaces`. We should add new features to new extensions rather than old ones!

nomeata: I am not actually sure. I think the main reason we have extensions in the first place is that we had thing thing called the report that was relatively sacrosanct. If we did not have that, would we even have started this whole business?

Also, I think of language extension flags a bit like library imports: it's a bit of red tape you put on top of the file to unlock certain features. Not so different from `import`! And, like there, some things are imported by default, somethings not…

Would you say that we should stop extending or changing libraries, and create new ones whenever we evolve a feature? Probably not.

tomjaguarpaw: Language extensions are more like importing a name than importing a module and yes, to export more features one should export more names, not change the behaviour of existing exported names, typically.

aspiwack: Tom, I'm not sure why you feel this way. I would certainly not be able to make such a confident statement (though I would rather tend to prefer Joachim's metaphor).

At any rate, it's generally considered perfectly fine to generalise the type of a function in subsequent versions of a library as it's backward compatible (though it's not forward compatible). So even in your metaphor, it would seem justified to change the meaning of an extension in a backward compatible way.

Rust editions, which are featured in the motivation section of the proposal do, as far as I know, incorporate backward compatible changes too with new compiler versions. In this mindset, it would be ok to have an extension change in a backward compatible way even if it is part of a `GHC20XX` extension set.

phadej:
> Rust editions, which are featured in the motivation section of the proposal do, as far as I know, incorporate backward compatible changes too with new compiler versions. In this mindset, it would be ok to have an extension change in a backward compatible way even if it is part of a GHC20XX extension set.

Then it should be ok to add `ExplicitNamespaces` to GHC2021, as AFAIK it is a backward compatible change.

aspiwack:
> Then it should be ok to add `ExplicitNamespaces` to GHC2021, as AFAIK it is a backward compatible addition.

By this reasoning, it would indeed. Now, there seem to be a pretty deep split in the community (including the GHC Steering Committee) on whether this reasoning is the one that we want to apply. So it comes up quite often. And there can't really be a conclusion, as there is no established answer.

I think it speaks to a missing [principle](https://github.com/ghc-proposals/ghc-proposals/blob/10b18465396c321ff5ea068f588694bfc4668263/principles.rst); but I don't know how to achieve sufficient consensus to move forward there.

proposals/0000-ghc2023.rst
tomjaguarpaw: Naturally it's hard to convey the fully subtlety of a breakage policy in a single sentence. I've written [a much more detailed exploration elsewhere](http://h2.jaguarpaw.co.uk/posts/opaleyes-api-breakage-policy/). I was just surprised by the analogy that Joachim drew. I think libraries should typically evolve  by adding *new* names, not changing the functionality of old ones, and I believe the same for language extensions.

aspiwack: I see. Thanks for the extra explanation: it's helpful. I don't think I fully agree, but I see where you're coming from now.

adamgundry:
I think when it comes to language "specifications" (interpreted loosely, so including `GHC2021`), it is important to maintain *forward compatibility* as well as *backward compatibility*. That is, if my code compiles under `GHC2021` as supported by GHC 9.10, I expect it to compile under GHC 9.2 as well. This means that a user can learn what `GHC2021` means and write code that conforms to it, independently of any particular compiler version.

(Now of course in reality there might be compiler bugs that break this property, and whether something is a bugfix or a minor new feature is often a judgement call. And changes to libraries introduce yet more complications. But that's the basic principle.)

---

Back to the main topic of this PR: I would suggest that for `GHC2023`  we should try to fix the infelicity that `GADTSyntax` and `ExistentialQuantification` are both included in `GHC2021`, and that this is enough to define a GADT, even though `GADTs` is notionally not part of `GHC2021`. (See https://gitlab.haskell.org/ghc/ghc/-/issues/21102). I think the simplest fix would be to simply enable `GADTs` (and hence `MonoLocalBinds`) as part of `GHC2023`.

safinaskar:
I hope you will address this: https://github.com/ghc-proposals/ghc-proposals/pull/380#issuecomment-750895034 .

Also, I think this would be good idea to remove GeneralisedNewtypeDeriving, because it is not compatible with Safe.

Also, I think we should add MonoLocalBinds, because it is simply good idea and because it is required by many other extensions anyway

adamgundry:
> Also, I think this would be good idea to remove GeneralisedNewtypeDeriving, because it is not compatible with Safe.

@safinaskar Do you use `Safe`? I've been under the impression it sees little use and may not be worth the effort of keeping around, so I would be interested to hear if you find it valuable.

(One might also argue that `GeneralizedNewtypeDeriving` could be considered `Safe` now that roles are relatively well established.)

tomjaguarpaw:
> I think the simplest fix would be to simply enable `GADTs` (and hence `MonoLocalBinds`)

Personally I love GADTs, but `MonoLocalBinds` causes some very puzzling behaviour, such as https://gitlab.haskell.org/ghc/ghc/-/issues/19396. In the absence of `MonoLocalBinds` it is easy to factor out repeated code into a `where` clause, without giving an explicit type signature. In the presence of `MonoLocalBinds` this procedure becomes unreliable. I am concerned about the impact on less-experienced users.

nomeata:
>  I think the simplest fix would be to simply enable GADTs (and hence MonoLocalBinds) as part of GHC2023.

Would you mind adding it, with the rationale, to the proposal text? (Via suggested edit, PR, or simply just push to the branch)

safinaskar:
> Do you use Safe?

@adamgundry . No. I wrote about Safe just to make sure this issue is not forgotten. Okey, I will not push this

Ericson2314:
Hmm the type variable rationalization will have to wait for GHC 2024 I'm afraid? Years and GHC releases are both infrequent enough getting everything to line up is tricky.

adamgundry:
I've added some text about `GADTs` and `MonoLocalBinds`. I still feel like the best option is to enable both, although I tried to set out some of the trade-offs and discuss other options.

Jashweii:
Is there a reason to add GADTs but not TypeFamil[yDependenc]ies? 
I would assume if we are going to have MonoLocalBinds we may as well get both.

Does the poll process apply to all extensions or just the ones proposed? Personally I would like BlockArguments.

adamgundry:
> Is there a reason to add GADTs but not TypeFamil[yDependenc]ies?

My opinion is that type families are still sufficiently unsatisfactory in some aspects of their design that we should keep them as an extension rather than blessing them in `GHC2023`, whereas GADTs are rather more robust. Others are welcome to disagree of course. :grin:

> Does the poll process apply to all extensions or just the ones proposed? Personally I would like BlockArguments.

We're still at the stage of discussing which extensions should be considered. I basically never use `BlockArguments` so I have no real opinion on it one way or the other.

nomeata:
> > Does the poll process apply to all extensions or just the ones proposed? Personally I would like BlockArguments.
> 
> We're still at the stage of discussing which extensions should be considered. I basically never use `BlockArguments` so I have no real opinion on it one way or the other.

My idea is to only put up those extensions for a vote where at least a committee member feels strongly enough about to put it on the list, together with an justification. This keeps the processes more focused than last time’s long ballot list.

I’m certainly sympathetic to `BlockArgument`. It’s precisely the kind of language feature that would make writing Haskell code a small, but noticeable bit more smooth and elegant, while the hurdle of actually enabling it is too big… But I wish there was a more principled way to make these decisions than just gut feeling.

Jashweii:
It's probably more of a community job like the "25 days of GHC extensions" blogposts, to encourage people to try them out then they might show up in a future poll.

AntC2:
Adding a link to [a current meta-discussion on the Steering Committee mailing list](https://mail.haskell.org/pipermail/ghc-steering-committee/2022-December/002996.html).

> should (ideally) every extension be on track to become on-by-default.

Presumably "on by default" there means included in `GHC202n`(?)

I presume including an extension in `GHC202n` means that the Committee thinks newbies should be faced with these extensions from the get go?

>  (or be discarded)

Yeah. Haskell used to be an experimental language. Some people I suspect never liked FunctionalDependencies; so TypeFamilies were introduced. Some people I suspect never liked TypeFamilies -- or at least found FunDeps more ergonomic: TypeFamilies have been available a decade and a half, I don't think we could just say FunDeps persist because 'old habits die hard'.

Behind all this is the Committee saying GHC/the language it supports has just gotten too complex? Speaking as a user, I don't think it's the plethora of options that makes it complex. Making them non-optional won't improve my GHC life. The complexity is because I have to understand what the option does even if I don't use it in my own code. I manage that by **switching off** every option, then wait to see the compiler complaining which options I need to switch on.

phadej:
I'd nominate `RoleAnnotations` to be included in GHC2023 too.

I don't really know a downside for not enabling it.

nomeata:
I guess the question is whether the role system is considered stable enough. Wasn't there recent discussion about changing something about it, by Simon? (But I think he withdrew that ideas).

I'm open to blessing that one, but don't have the full picture of what's going on in that area.

simonpj:
> As [Arnaud explained on the mailing list](https://mail.haskell.org/pipermail/ghc-steering-committee/2022-November/002949.html), Rust (generally not perceived as a language with a perception of low stability) has a very similar model with their “language editions”.

Arnaud points to the [Rust edition docs](https://doc.rust-lang.org/edition-guide/editions/index.html).  Indeed their slogan "stability without stagnation" is very much what we seek too.

However I note that, while there is no explcitly documented cadence, the Rust edition cadence seems to be 3 yrs (2015, 2018, 2021).  I think that some people think that two years would be a better cadence than one year for GHC.  I don't have a strong opinion myself, but I would love to hear from actual users.

nomeata:
NB, to not confuse tangerine to clementines, that rust bumps editions _only for breaking changes_, but otherwise adds new language features as they go. Most of our extensions are non-breaking (syntax guarded).

So following the rust model would mean to have a new GHC20xx with non-breaking extensions (e.g. `DerivingFunctor`)  frequently, but promise to hold back breaking changes and bundle them in less frequent releases.

simonpj:
> NB, to not confuse tangerine to clementines, that rust bumps editions only for breaking changes,

I see.  So you are saying that Rust silently extends the language with non-breaking changes, without changing editions or flags.  So a crate that works with one version of the compiler (version A) for edition 2021 might not work with another version of the compiler (version B) for the same edition 2021.  Because A accepts a feature that B doesn't. 

That seems... strange.

nomeata:
That is my understanding - „Vorwärts immer – rückwärts nimmer!“. And looking at the announcement of a recent rust release certainly seems to confirm that: https://blog.rust-lang.org/2022/12/15/Rust-1.66.0.html

phadej:
> I guess the question is whether the role system is considered stable enough. Wasn't there recent discussion about changing something about it, by Simon? (But I think he withdrew that ideas).

`RoleAnnotations` is de-facto stable. If someone wants to change how roles are declared they'll have to deal with quite massive breakage. *Search completed (565 matches across 158 packages)* on https://hackage-search.serokell.io/?q=%5Etype+role

nomeata:
Do you want to contribute a section on it to this proposal (via suggested edit or PR against my branch)? It's not clear get whether we even will have an attempt to define GHC2023, but if we do, we can certainly vote on this one. And if we don't I'll carry it over to the next attempt.

proposals/0000-ghc2023.rst
```diff
@@ -0,0 +1,179 @@
...
+Proposed Change Specification
+-----------------------------
+
+The ``GHC2023`` language extension set comprises of all language extensions
+that ``GHC2021`` comprises, plus the following
+
+* `ExplicitNamespaces`_
+* `GADTs`_ (and the implied `MonoLocalBinds`_)
+* `LambdaCase`_
+
```
phadej, line 31:
```suggestion
* `RoleAnnotations`_
```

proposals/0000-ghc2023.rst
```diff
@@ -0,0 +1,179 @@
...
+I (Joachim) can prepare a MR.
+
+.. _ExplicitNamespaces: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/explicit_namespaces.html#extension-ExplicitNamespaces
+.. _ExistentialQuantification: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/existential_quantification.html#extension-ExistentialQuantification
+.. _GADTs: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/gadt.html#extension-GADTs
+.. _GADTSyntax: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/gadt_syntax.html#extension-GADTSyntax
+.. _TypeOperators: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/type_operators.html#extension-TypeOperators
+.. _OverloadedStrings: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/overloaded_strings.html#extension-OverloadedStrings
+.. _LambdaCase: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/lambda_case.html#extension-LambdaCase
+.. _MonoLocalBinds: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/let_generalisation.html#extension-MonoLocalBinds
```
phadej, line 179:
```suggestion
.. _MonoLocalBinds: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/let_generalisation.html#extension-MonoLocalBinds
.. _RoleAnnotations: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/roles.html#extension-RoleAnnotations
```

safinaskar:
> So a crate that works with one version of the compiler (version A) for edition 2021 might not work with another version of the compiler (version B) for the same edition 2021

Exactly.

For this reason developers either use practice "we always use latest compiler", either they specify so-called "minimum supported Rust version" (MSRV) (MSRV is compiler version, not edition).

I highly recommend this very concentrated info on Rust evolution process (and editions and compiler versions and C++ comparison and related things) from Rust Library Team member: https://blog.m-ou.se/rust-standard/ .

I like Rust's approach more than Haskell's approach (Haskell has combinatoric explosion of language versions through combinations of language pragmas)

proposals/0000-ghc2023.rst
```diff
@@ -0,0 +1,179 @@
...
+default”. It also missed ``GHC2021`` by just two votes. There is a whole style
+of writing Haskell that makes extensive use of ``\case``. And (unlike the runner up in the survey, `OverloadedStrings_`), it only enables *new* syntax, i.e. it does not change existing code.
+
+A counter-agument to adding `LambdaCase` is that just extended the meaning of
+`LambdaCase`_ with ``\cases`` in `#302
+<https://github.com/ghc-proposals/ghc-proposals/pull/302>`_, and if one only
+wants to add extensions to ``GHC20xx`` that have been proven to be stable, then
+this one probably isn’t yet.
+
+
```
phadej, line 159:
```suggestion

Why add RoleAnnotations?
------------------------

Roles are essential part of modern GHC Haskell.

As `GeneralisedNewtypeDeriving` is in GHC2021 language, so should be `RoleAnnotations`. They are different sides of the same feature: without correct role annotations GND cannot be used safely.

Role annotations are required for correctly writing types with internal invariants like `Set` or "fast" implementations like `data Fin (n :: Nat) = UnsafeFin Int`.

````

Comment (nomeata) [reply proposals/0000-ghc2023.rst:159]: The latter is indeed a compelling point: `RoleAnnotations` aren’t really optional any more.

**Comment by nomeata** [comment]
> I highly recommend this very concentrated info on Rust evolution process (and editions and compiler versions and C++ comparison and related things) from Rust Library Team member: https://blog.m-ou.se/rust-standard/ .

Thanks a lot for that link, very helpful. There are lots of similarities (can mix editions / pragmas on a per-unit basis), and there are a few things to learn from (e.g. that it’s feasible to do have code update tools to make petty breaking changes like introducing new keywords less painful, and to do automated testing of large body of available code)!  Makes me even more bullish on “frequent GHC20xx releases with non-breaking changes are a good thing“.

**Comment by phadej** [comment]
> This means that everything that is released as part of stable Rust, [is there to stay](https://rust-lang.github.io/rfcs/1122-language-semver.html). Once a feature is stable, we cannot remove it.

If that includes standard libraries, i.e. `base` then that is very high standard. GHC is not there, and I'm not sure it should be.

(Note: I have no idea how large Rust standard library is).

> We are so committed to stability, that we compile and test every single crate on crates.io and every single Rust repository on GitHub with a Cargo.lock file

Is essentially compiling whole Stackage. GHC is not there, and I'm not sure it should be. (TH code, ghc plugins are brittle, but i'd say it's accepted that those may break with each major GHC).

>  and will be made available as an unstable feature. Unstable features can only be used on nightly versions of Rust, not on beta or stable versions.

In GHC this would mean that e.g. `LinearTypes` would be still unstable feature and not available in any stable "released" GHC, only in GHC-HEAD-yyyymmdd (but that one would be able to install with ghcup).

Do I understand correctly, that if I'm using stable Rust I only tell the edition, but I don't have any "feature toggles" / extensions. In stable Rust there are only stable features, and they are all "on"?

> A “Rust Edition” is basically a variant (or “dialect”) of the language. Unlike versions, all editions are supported and the latest version of the compiler supports all editions: currently Rust 2015, Rust 2018, and Rust 2021. In Rust 2015, async is a regular identifier, but in Rust 2018 and Rust 2021, async is a keyword.

I notice that **all editions are supported**. That is what I tried to talk about previously. If `LambdaCase` was included in GHC2021, then even with GHC-10 `cases` won't be a keyword in GHC2021 (as `cases` steals syntax), but would in future GHC202*. That implies that `\cases` should been a new extension.

---

It's good to understand the differences of language evolution.

@simonpj wrote

> I see. So you are saying that Rust silently extends the language with non-breaking changes, without changing editions or flags. So a crate that works with one version of the compiler (version A) for edition 2021 might not work with another version of the compiler (version B) for the same edition 2021. Because A accepts a feature that B doesn't.

> That seems... strange.

But it isn't. As Rust is very careful *to not break stuff*, upgrading to newer compiler is trivial.
That cannot be less true for GHC, and that's probably why the Rust practice seems strange.

That gives Rust "freedom" to extend the language, because they actually test that new stuff doesn't breaking anything. And that implies that there are no good reasons to not upgrade the rust compiler if a crate requires newer one.

---

That all said, I wonder how Rust community treats comments like "we have ten million lines of private Rust code and this feature will break some of it, but we cannot show it, and nothing breaks on crates.io". Is it just not an argument in Rust?

**Comment by phadej** [comment]
@safinaskar thanks for confirming my understanding.

> To enable such experimental APIs, you have to write `feature` pragma at the top of your file (and you have to use nightly compiler).

That's something which doesn't exist in GHC. Library contents cannot be controlled by extensions or language versions. That could helped keep Haskell2010 `Monad` without `Applicative` super class. OTOH I have no idea how that would work out in GHC.

Is using intrinsics and guarding parts of library behind a feature toggle something only possible in `std` and `core`?

> Procedural macros are powerful (like TH). Procedural macros are allowed to use additional APIs located in crate proc_macro

The API doesn't seem to include any kind of high level abstract syntax tree. The [TokenTree](https://doc.rust-lang.org/nightly/proc_macro/enum.TokenTree.html) looks more like lisp-lists. It's an interface I can believe is not impossible to keep stable.

https://doc.rust-lang.org/reference/procedural-macros.html confirms that

> This crate primarily contains a [TokenStream](https://doc.rust-lang.org/proc_macro/struct.TokenStream.html) type. Procedural macros operate over token streams instead of AST nodes, which is a far more stable interface over time for both the compiler and for procedural macros to target. A token stream is roughly equivalent to Vec<TokenTree> where a TokenTree can roughly be thought of as lexical token.

**Comment by nomeata** [comment]
> I notice that all editions are supported. That is what I tried to talk about previously.

I'm warming up to a policy where extensions that are “stable” (included in a GHC20xx) may only be changed in a backwards compatible way, or else a new extension name needs to be introduced, if that's what it takes to allow ourselves to make more of today's Haskell on by default.

(I'm wondering if it's with distinguishing more serious breakage from “this local identifier is now a keyword, please rename” breakage, and be more liberal about the latter. Even more so should we also get a code update tool eventually.)

**Comment by phadej** [comment]
> (I'm wondering if it's with distinguishing more serious breakage from “this local identifier is now a keyword, please rename” breakage, and be more liberal about the latter

The new keyword can be an exported identifier from the library e.g. `pattern` (e.g. selector of a record). New keyword "small" breakage may cascade very unfortunately. And renaming cannot be done automatically, if it's an exported identifier it's up to the maintainer to decide how to rename (e.g. `forAll` or `forall_`, latter seems weird but works better with `exists_`).

If it's breaking change, than it's breaking change, period.

**Comment by safinaskar** [comment]
> That could helped keep Haskell2010 `Monad` without `Applicative` super class

Things are more complicated. Haskell classes are called traits in Rust. Library feature attributes usually control visibility of items and nothing else. They cannot really control supertrait relationship. (In previous posts I wrote "pragmas" instead of "attributes", I'm sorry about that, these things are called "attributes" in Rust.)

Let me show you example. When Rust community planned Rust 2021, they wanted to make arrays (i. e. `[T; N]`) to have [IntoIterator](https://doc.rust-lang.org/nightly/std/iter/trait.IntoIterator.html) trait in Rust 2021 (and not to have in other editions).

Unfortunately, this was not possible.

The unit of compilation in Rust is crate. Crate consists of (possible multiple) files. Final executable is compiled from (possible multiple) crates. Editions are set per-crate. But you cannot really have `[T; N]` to have trait "IntoIterator" in one crate and not to have in other crate inside single executable!

So, Rust community did this: they released new Rust compiler, which made `[T; N]` to have "IntoIterator" trait in all Rust editions. But they made Rust 2015 and Rust 2018 parser to have special kludge, which looks like `[T; N]` has not `IntoIterator` trait. More info: https://doc.rust-lang.org/nightly/std/primitive.array.html#editions .

So, if Haskell was Rust, then we could create new edition, where `Monad` has `Applicative` superclass/supertrait. But this would be actually some hack implemented in parser and actually `Monad` would have `Applicative` superclass in all editions

**Comment by safinaskar** [comment]
> Is using intrinsics

I don't understand your question on intristics.

Anyway, let me tell you more on intristics. First of all, there are two different concepts: intristics and language items. Both things are unstable low-level interface between compiler and user code (and usually this "user code" is standard library). Both things usually are useful for standard library developers only.

Intristics are special functions implemented in compiler. They have prototype, but they don't have body (because body is in compiler). They are similar to Haskell's primops, i. e. [GHC.Prim](https://hackage.haskell.org/package/ghc-prim-0.9.0/docs/src/GHC.Prim.html) module a. k. a. `compiler/GHC/Builtin/primops.txt.pp`.

Language items are in some sense opposite things. They are items (not necessary functions) implemented in user code (usually standard library), but which have special meaning for compiler. They are similar to Haskell's well known names (`compiler/GHC/Builtin/Names.hs`). More info on lang items is here: https://manishearth.github.io/blog/2017/01/11/rust-tidbits-what-is-a-lang-item/

In short, intristics are special functions (full list of such functions is hard coded in compiler), which can be _called from user code_ (usually std lib) and which are _implemented in compiler_. And lang items are special items (full list of such items is hard coded in compiler), which can be _called_ (and used in other ways) _in compiler_ and which are _implemented in user code_ (usually std lib).

In fact, things are even more complicated: sometimes the compiler ignores lang item's implementation and provides its own (for example, this is what happens for `impl Add for isize` as we know from Manish's article above).

How to use intristics? Ultimate low-level way to use intristics is write something like this:
```rust
extern "rust-intrinsic" {
pub fn volatile_load<T>(src: *const T) -> T;
}
```
Then you can call this intristic. You can write this `extern "rust-intrinsic"` construct (and then call the intristic) everywhere (i. e. in any crate). There is no any connection to standard library. This requires a nightly compiler. And you have to add "intristic" feature attribute.

There is slightly more high-level way to use intristics. All intristics are listed in https://doc.rust-lang.org/nightly/core/intrinsics/index.html module (in `core`), so you can simply import them from there. This requires nightly compiler and special attribute, too. And this can be done from everywhere, too.

(I can also tell you how to define and use language items, if you want.)

> guarding parts of library behind a feature toggle something only possible in `std` and `core`?

It seems yes. (In std, core, proc_macro and other standard crates.) They are implemented in standard library using `#[unstable(feature = "...", issue = "...")]` construct. I just tried to use this construct in my code and saw a message "stability attributes may not be used outside of the standard library"

safinaskar:
> The API doesn't seem to include any kind of high level abstract syntax tree

Yes. Crate `proc_macro` is part of standard library and has the same stability guarantees like Rust compiler itself.

There are also widely used higher level libraries `syn` and `quote`. They are build on top of `proc_macro` and they provide AST. But they are just a normal libraries published on crates.io and use usual semver.

> Does Rust has orphan trait implementations

No.

Here is motivation for that `IntoIterator` decision: https://github.com/rust-lang/rust/pull/65819#issuecomment-548508659 (that thread may contain other useful info)

nomeata:
> The new keyword can be an exported identifier

That’s why I said _local_ (as in the case of LambdaCase), but maybe muttered that word too quietly. I agree that with exported identifiers things are a lot more tricky.

nomeata:
I plan to start a community vote in two weeks, and until then invite the committee and the community to add their nominations and refine the arguments for and against the extensions.

/remind me in two weeks

nomeata:
I plan to start a community vote in two weeks, and until then invite the committee and the community to add their nominations and refine the arguments for and against the extensions.

/remind me in two weeks to start a vote

nomeata:
/remind me to start a vote in in two weeks

nomeata:
It may make a difference for our assessment whether an extension is really just an extension (does not affect any existing ``GHC2021``), or can change the behavior of existing compiling code, so I annotated each nominated extensions accordingly and to the best of my knowledge.

proposals/0000-ghc2023.rst
```diff
@@ -156,6 +162,12 @@ A counter-agument to adding `LambdaCase` is that just extended the meaning of
 wants to add extensions to ``GHC20xx`` that have been proven to be stable, then
 this one probably isn’t yet.
 
+__Backward compatibility__: Enabling `LambdaCase`_ does not affect
+existing ``GHC2021`` code, with the exception of lambda-bound variable names
+``casess`` (GHC already forbids the ``\case`` even without ``-XLambdaCase``).
```
phadej, line 167:
```suggestion
``cases`` (GHC already forbids the ``\case`` even without ``-XLambdaCase``).
```

proposals/0000-ghc2023.rst
```diff
@@ -166,6 +178,9 @@ As `GeneralisedNewtypeDeriving`_ is in the GHC2021 language set, so should be `R
 
 Role annotations are required for correctly writing types with internal invariants like ``Set`` or "fast" implementations like ``data Fin (n :: Nat) = UnsafeFin Int``.
 
+__Backward compatibility__: Enabling `LambdaCase`_ does not affect
```
phadej, line 181:
```suggestion
__Backward compatibility__: Enabling `RoleAnnotations`_ does not affect
```

nomeata:
I’ve started a committee vote on whether we even want GHC2023; if that goes through we’ll proceed defining what it is.

chreekat:
Is it a [community vote](https://github.com/ghc-proposals/ghc-proposals/pull/559#issuecomment-1374796178) or a [committee vote](https://github.com/ghc-proposals/ghc-proposals/pull/559#issuecomment-1399474615)? :D

nomeata:
Oh, that was just a typo, sorry. It’s meant to be a committee vote all the time. Sorry …

(Not that I wouldn’t be curious what the community thinks. If there such a thing as “the” community.)

adamgundry:
I realise there's currently a debate on whether the next `GHC20xx` will be 2023 or later but something to put out there for whenever it is considered: should we remove `PolyKinds`? It can lead to rather hard to diagnose type errors (see for example https://mail.haskell.org/pipermail/haskell-cafe/2023-January/135847.html). I recently hit another case: upgrading a codebase to use `GHC2021`, I was puzzled by kind errors that turned out to result from `makeClassy` in `lens` generating a class with an extra type parameter thanks to `PolyKinds` being enabled on a datatype definition.

treeowl:
Personally, I'd like to see polykindedness by default (i.e., without an explicit `forall`) go away. That would help error messages in the common case and make it much easier to introduce polykindedness into a module gradually.

phadej:
@adamgundry I counter that by https://gitlab.haskell.org/ghc/ghc/-/merge_requests/7256

If we had `-Wmissing-poly-kind-signatures` then it would be a lot easier to analyse where polykinds are not desired. And I argue that for polykinded types your really ought to write their signature. (But `-Wmissing-kind-signatures` is a bit off limits).

(The `-Wmissing-poly-kind-signatures` idea came up from upgrading GHC itself).

nomeata:
I could not gather a strong majority for this on the committee, so let’s skip GHC2023.

PR #562, opened 2022-12-12
Make Symbol a newtype over String

cdornan:
I am restoring #546 to see if it can be rescued as it seemed to be a good idea.

[Rendered](https://github.com/ghc-proposals/ghc-proposals/blob/symbol-type/proposals/0000-symbol-newtype-string.rst).

### DIFF
--- a/proposals/0000-symbol-newtype-string.rst
+++ b/proposals/0000-symbol-newtype-string.rst
@@ -27,7 +27,7 @@ Consider the data type with fields of type ``String``::
 
   data Fields = Fields [String]
 
-``Fields`` is inhabitaed by terms, and since recently (`The Char Kind <https://github.com/ghc-proposals/ghc-proposals/pull/387>`_) also by 
+``Fields`` is inhabited by terms, and since recently (`The Char Kind <https://github.com/ghc-proposals/ghc-proposals/pull/387>`_) also by 
 types. However type literals produce ``Symbol``, not ``String`` ::
 
   fs = Fields ["foo", "bar"]           -- ok
--- a/proposals/0000-symbol-newtype-string.rst
+++ b/proposals/0000-symbol-newtype-string.rst
@@ -63,7 +63,7 @@ requires importing ``ghc-prim``.
 We don't propose any changes to ``GHC.TypeLits`` interface.
 Firstly the implementation of ``Symbol`` stays internal.
 Also ``symbolVal`` will continue to return ``String`` value.
-In future the ``GHC.Symbol`` interface mirroring ``GHC.TypeLits`` functionlity
+In future the ``GHC.Symbol`` interface mirroring ``GHC.TypeLits`` functionality
 could be vreated , but using ``Symbol`` also for the term level can be made as
 CLC proposal, as that would be only a library change.
 
--- a/proposals/0000-symbol-newtype-string.rst
+++ b/proposals/0000-symbol-newtype-string.rst
@@ -64,7 +64,7 @@ We don't propose any changes to ``GHC.TypeLits`` interface.
 Firstly the implementation of ``Symbol`` stays internal.
 Also ``symbolVal`` will continue to return ``String`` value.
 In future the ``GHC.Symbol`` interface mirroring ``GHC.TypeLits`` functionality
-could be vreated , but using ``Symbol`` also for the term level can be made as
+could be created , but using ``Symbol`` also for the term level can be made as
 CLC proposal, as that would be only a library change.
 
 Technically this is a library only change, but maybe it's only
--- a/proposals/0000-symbol-newtype-string.rst
+++ b/proposals/0000-symbol-newtype-string.rst
@@ -1,0 +1,128 @@
+Make Symbol a newtype over String
+=================================
+
+.. author:: Oleg Grenrus
+.. date-accepted::
+.. ticket-url::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/546>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+I propose changing ``data Symbol`` definition
+to be
+
+::
+
+    newtype Symbol = MkSymbol String
+
+Motivation
+----------
+
+The motivation is similar to accepted and implemented
+`Unify Nat and Natural proposal <https://github.com/ghc-proposals/ghc-proposals/pull/364>`_.
+
+Consider the data type with fields of type ``String``::
+
+  data Fields = Fields [String]
+
+``Fields`` is inhabitaed by terms, and since recently (`The Char Kind <https://github.com/ghc-proposals/ghc-proposals/pull/387>`_) also by 
+types. However type literals produce ``Symbol``, not ``String`` ::
+
+  fs = Fields ["foo", "bar"]           -- ok
+  type FS = Fields '[ "foo", "bar" ]   -- not ok
+
+Alternatively, we could declare the type with fields of type ``Symbol``::
+
+  data Fields = Fields [Symbol]
+
+Then it would have the opposite issue::
+
+  fs = Fields ["foo", "bar" ]         -- not ok
+  type FS = Fields '[ "foo", "bar" ]  -- ok
+
+As with ``Nat`` and ``Natural`` proposal, it's possible to go roundabout
+this (and e.g. ``singletons-base`` does). However, if ``Symbol`` were
+an ordinary ``newtype`` over ``String``, we would avoid this issue entirely.
+
+(Note: ``singletons-base`` demotes ``Symbol`` to ``Text`` which is
+incorrect, as ``Text`` cannot represent all ``String`` values)
+
+Proposed Change Specification
+-----------------------------
+
+* Change ``data Symbol`` to ``newtype Symbol = MkSymbol String``
+
+In particular, we don't propose any changes to ``GHC.TypeLits`` interface.
+Firstly the implementation of ``Symbol`` stays internal.
+Also ``symbolVal`` will continue to return ``String`` value.
+In future the ``GHC.Symbol`` interface mirroring ``GHC.TypeLits`` functionlity
+could be vreated , but using ``Symbol`` also for the term level can be made as
+CLC proposal, as that would be only a library change.
+
+Technically this is a library only change, but maybe it's only
+a happy coincidence that GHC internals don't need to be changed,
+when ``Symbol`` definition is changed.
+
+Effect and Interactions
+-----------------------
+
+This change is completely backwards-compatible.
+
+Costs and Drawbacks
+-------------------
+
+This proposal introduces "yet another string" type into Haskell ecosystem.
+Yet, it serves as specific purpose.
+
+Alternatives
+------------
+
+An alternative is to export ``Symbol`` definition from ``GHC.TypeLits``.
+This however raises questions like whether ``MkSymbol '[] ~ ""``.
+For now we pretend that ``MkSymbol`` should not be promoted,
+until that can be resticted.
+
+It is also possible to prevent ``MkSymbol`` usage on the type level by
+defining ``Symbol`` like::
+
+  data Symbol = MkSymbol (# String #)
+
+As unlifted types cannot be promoted, it won't be possible to successfully
+use ``MkSymbol`` on the type level. (No more success than with ``NS`` and
+``NB`` constructors of type ``Natural``).A
+
+An alternative is to make ``String`` promote to ``Symbol``.
+That won't work out well, because lists of characters (i.e. ``String``)
+can be promoted already.
+
+The opposite alternative is to make ``type Symbol = String``.
+It was also meentioned as an alternative in The Char Kind proposal.
+It was rejected then, and the reasoning is still valid:
+First of all, we keep ``Symbol`` for type-checking efficiency.
+Moreover, we would also handle type families inside cons cells when solving
+``HasField`` constraints. For example, ``HasField T ('x' : F y : G z) ty``.
+
+It's possible to use some other type for ``Symbol`` "backend" than ``String``,
+however ``Text`` doesn't qualify as ``String`` (and type-level ``Symbol``)
+can contain values ``Text`` cannot represent. For example single surrogate codepoints.
+``SNat`` and ``KnownSymbol`` are backed by ``String``, so it is a natural
+choice.
+
+Unresolved Questions
+--------------------
+
+None.
+
+Implementation Plan
+-------------------
+
+The patch is very small. I'll write it.
+
+Endorsements
+-------------
+
+None atm.
--- a/proposals/0000-symbol-newtype-string.rst
+++ b/proposals/0000-symbol-newtype-string.rst
@@ -1,7 +1,7 @@
 Make Symbol a newtype over String
 =================================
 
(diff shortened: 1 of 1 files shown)


Ericson2314:
With @adamgundry's suggestions, especially the notice on `ghc-prim`, this looks good to me.

re-xyr:
This all look pretty good, but I just have a tangentially related question: what are the performance implications of having `type Symbol = String`, specifically? In the last PR that was closed I saw mentions of `Generic` but I wasn't sure what that was exactly about.

proposals/0000-symbol-newtype-string.rst
```diff
@@ -0,0 +1,128 @@
...
+CLC proposal, as that would be only a library change.
+
+Technically this is a library only change, but maybe it's only
+a happy coincidence that GHC internals don't need to be changed,
+when ``Symbol`` definition is changed.
+
+Effect and Interactions
+-----------------------
+
+This change is completely backwards-compatible.
```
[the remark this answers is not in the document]
cdornan: I am incorporating these points.

proposals/0000-symbol-newtype-string.rst
```diff
@@ -0,0 +1,128 @@
+Make Symbol a newtype over String
+=================================
+
+.. author:: Oleg Grenrus
```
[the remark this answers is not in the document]
cdornan: duly extended

proposals/0000-symbol-newtype-string.rst
```diff
@@ -0,0 +1,128 @@
...
+  fs = Fields ["foo", "bar" ]         -- not ok
+  type FS = Fields '[ "foo", "bar" ]  -- ok
+
+As with ``Nat`` and ``Natural`` proposal, it's possible to go roundabout
+this (and e.g. ``singletons-base`` does). However, if ``Symbol`` were
+an ordinary ``newtype`` over ``String``, we would avoid this issue entirely.
+
+(Note: ``singletons-base`` demotes ``Symbol`` to ``Text`` which is
+incorrect, as ``Text`` cannot represent all ``String`` values)
+
```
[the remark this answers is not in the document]
cdornan: Adding the proposed qualification.

proposals/0000-symbol-newtype-string.rst
```diff
@@ -0,0 +1,128 @@
...
+Proposed Change Specification
+-----------------------------
+
+* Change ``data Symbol`` to ``newtype Symbol = MkSymbol String``
+
+In particular, we don't propose any changes to ``GHC.TypeLits`` interface.
```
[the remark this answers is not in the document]
cdornan: I have incorporated the explanation on where the change is taking place.

adamgundry:
> With @adamgundry's suggestions, especially the notice on ghc-prim, this looks good to me.

The credit here really goes to @phadej, I'm just interpreting. :smile: 

> what are the performance implications of having `type Symbol = String` instead, specifically? In the last PR that was closed I saw mentions of `Generic` but I wasn't sure what that was exactly about.

At the moment, GHC represents type-level `Symbol`s as packed strings. Naively changing the representation to be `[Char]` would significantly expand the AST size of the types being manipulated internally, which would hurt compile-time performance. In practice I think it would be possible to retain a packed representation, but then the constraint solver would need to be more complex (e.g. to solve an equality between a packed symbol and a cons). The added complexity doesn't seem worth it, compared to the alternative of having `Symbol` be a properly abstract type.

proposals/0000-symbol-newtype-string.rst
```diff
@@ -0,0 +1,134 @@
...
+----------
+
+The motivation is similar to accepted and implemented
+`Unify Nat and Natural proposal <https://github.com/ghc-proposals/ghc-proposals/pull/364>`_.
+
+Consider the data type with fields of type ``String``::
+
+  data Fields = Fields [String]
+
+``Fields`` is inhabitaed by terms, and since recently (`The Char Kind <https://github.com/ghc-proposals/ghc-proposals/pull/387>`_) also by
```
googleson78, line 30:
```suggestion
``Fields`` is inhabited by terms, and since recently (`The Char Kind <https://github.com/ghc-proposals/ghc-proposals/pull/387>`_) also by 
```

proposals/0000-symbol-newtype-string.rst
```diff
@@ -0,0 +1,134 @@
+Make Symbol a newtype over String
+=================================
+
+.. author:: Oleg Grenrus (with ammendments by Chris Dornan and Adam Gundry)
```
adamgundry, line 4:
```suggestion
.. author:: Oleg Grenrus (with amendments by Chris Dornan and Adam Gundry)
```

proposals/0000-symbol-newtype-string.rst
```diff
@@ -0,0 +1,134 @@
...
+
+* Change ``data Symbol`` to ``newtype Symbol = MkSymbol String``
+
+The definition being changed lives in ``GHC.Types``, and thus accessing the newtype constructor
+requires importing ``ghc-prim``.
+
+We don't propose any changes to ``GHC.TypeLits`` interface.
+Firstly the implementation of ``Symbol`` stays internal.
+Also ``symbolVal`` will continue to return ``String`` value.
+In future the ``GHC.Symbol`` interface mirroring ``GHC.TypeLits`` functionlity
```
JakobBruenker, line 66:
```suggestion
In future the ``GHC.Symbol`` interface mirroring ``GHC.TypeLits`` functionality
```

proposals/0000-symbol-newtype-string.rst
```diff
@@ -0,0 +1,134 @@
...
+could be vreated , but using ``Symbol`` also for the term level can be made as
```
JakobBruenker, line 67: s/vreated/revisited? I'm not 100% what word it is supposed to be

cdornan: 'created'

Ericson2314:
If GHC *actually* does `[Char]` for `Symbol`, things will get a lot slower, so we probably don't want it to do that, and have to explicitly convert between the two.

And frankly this is not so bad because at the term level we *also* don't want to promote the use of `String` Either.

I suspect at some point we we will settle on `Symbol = Text`, and the `newtype` plan keeps that future open.

cdornan:
In the hope of not burning up too much precious attention, can I suggest that folks focus on destroying this proposal as is rather than an extended discussion of all of our options and their various merits and demerits. I know that isn't what we usually do but I think it might be the best here.

simonpj:
I don't want to re-open the discssion. I just want a reason for the difference.  John has given one. OK, I'm in support.

Ericson2314:
Yes I was just trying to summarize it. And perhaps the apartness part (which admittedly did not come up before) is worth adding to this proposal now.

proposals/0000-symbol-newtype-string.rst
```diff
@@ -0,0 +1,134 @@
...
+
+  fs = Fields ["foo", "bar" ]         -- not ok
+  type FS = Fields '[ "foo", "bar" ]  -- ok
+
+As with ``Nat`` and ``Natural`` proposal, it's possible to go roundabout
+this (and e.g. ``singletons-base`` does). However, if ``Symbol`` were
+an ordinary ``newtype`` over ``String``, we would avoid this issue entirely.
+
+(Note: ``singletons-base`` demotes ``Symbol`` to ``Text`` which is
+incorrect, as ``Text`` cannot represent all ``String`` values)
```
aspiwack, line 50: Can't it? Do you have a reference for this claim? I'm infinitely curious now.

aspiwack (commented):
I fail to see how the proposed change addresses the motivations. The _Effects and interactions_ section doesn't address this point either. Could you elaborate?

proposals/0000-symbol-newtype-string.rst
re-xyr: From the haddock of `text`:

> A [Text](https://hackage.haskell.org/package/text-2.0.1/docs/Data-Text.html#t:Text) value is a sequence of Unicode scalar values, as defined in [§3.9, definition D76 of the Unicode 5.2 standard](http://www.unicode.org/versions/Unicode5.2.0/ch03.pdf#page=35). As such, a [Text](https://hackage.haskell.org/package/text-2.0.1/docs/Data-Text.html#t:Text) cannot contain values in the range U+D800 to U+DFFF inclusive. Haskell implementations admit all Unicode code points ([§3.4, definition D10](http://www.unicode.org/versions/Unicode5.2.0/ch03.pdf#page=13)) as [Char](https://hackage.haskell.org/package/base-4.14.1.0/docs/Data-Char.html#t:Char) values, including code points from this invalid range. This means that there are some [Char](https://hackage.haskell.org/package/base-4.14.1.0/docs/Data-Char.html#t:Char) values (corresponding to [Surrogate](https://hackage.haskell.org/package/base-4.14.1.0/docs/Data-Char.html#v:Surrogate) category) that are not valid Unicode scalar values, and the functions in this module must handle those cases.
> 
> Within this module, many functions construct a [Text](https://hackage.haskell.org/package/text-2.0.1/docs/Data-Text.html#t:Text) from one or more [Char](https://hackage.haskell.org/package/base-4.14.1.0/docs/Data-Char.html#t:Char) values. Those functions will substitute [Char](https://hackage.haskell.org/package/base-4.14.1.0/docs/Data-Char.html#t:Char) values that are not valid Unicode scalar values with the replacement character "�" (U+FFFD). Functions that perform this inspection and replacement are documented with the phrase "Performs replacement on invalid scalar values". The functions replace invalid scalar values, instead of dropping them, as a security measure. For details, see [Unicode Technical Report 36, §3.5](http://unicode.org/reports/tr36/#Deletion_of_Noncharacters).)

aspiwack: Very nice.

That being said, is it relevant here? I don't think we can create `Symbol` values with surrogate code points in them, can we?

re-xyr: I think we can?

```
ghci> :set -XDataKinds
ghci> type A = "\xd800"
ghci> :i A
type A :: GHC.Types.Symbol
type A = "\55296" :: GHC.Types.Symbol
```

re-xyr: Just a tangentially related idea, but we can make `type Symbol = Text` if we do replacement on `Symbol` literals. `Text` is uninhabited at type level so that isn't a problem either. Indeed this is breaking, so it's not very desirable.

aspiwack: Arfglbl. Indeed!

Though while `Symbol = Text` is technically breaking, it is probably not in a way that is depended upon in practice. So probably alright? Possibly maybe? But `Text` is not in base (yet?) so it's actually a pretty hard change to effect.

re-xyr:
> I fail to see how the proposed change addresses the motivations. The Effects and interactions section doesn't address this point either. 

I think this proposal just prepares for a future CLC proposal where we add `IsString Symbol`. Then we can use the same syntax for both term and type level `Symbol`s.

adamgundry:
Yes, this proposal addresses the motivation by allowing `Symbol` to be used sensibly at both the type and term levels. It requires a GHC-internal module import to use at the term level, but that's because specifying the term-level API for `Symbol` is a matter for the CLC (and the proposal says as much).

proposals/0000-symbol-newtype-string.rst
int-index: > But Text is not in base (yet?)

It shouldn't be hard to move the type declaration to `base` (`GHC.Types.Text` maybe?) but keep all the combinators for working on it in `text`. Just move these few lines of code:

```
data Text = Text
    {-# UNPACK #-} !Array -- bytearray encoded as UTF-8
    {-# UNPACK #-} !Int     -- offset in bytes (not in Char!), pointing to a start of UTF-8 sequence
    {-# UNPACK #-} !Int     -- length in bytes (not in Char!), pointing to an end of UTF-8 sequence

data Array = ByteArray ByteArray#
```

re-xyr: It wouldn't even need to be `base`; `ghc-prim` is enough. `text` also depends on `ghc-prim` already, so all of this can be done without CLC.

cdornan:
@re-xyr agreed; i am going to be afk until later &mdash; like to have a stab at clarifying; your input not the draft would be great; @Ericson2314 also, of course

goldfirere:
My concern here echoes @aspiwack's: the proposal does not seem to address the motivation. But, I do see how, if we had `IsString Symbol` and `-XOverloadedStrings`, then things would indeed be better. I would want a loud Haddock comment on the new `Symbol` definition that it is subject to change. (Exporting it only from ghc-prim is not enough, as I'm sure type-lit libraries will re-export!) With that caveat, I'm in support, in the hopes that we can do even better further in the future.

cdornan:
@goldfirere I think that reflects the sum of the discussion on the proposal &mdash; it really kicking off a process. I will add some language accordingly.

adamgundry:
In the light of the new consensus at https://github.com/haskellfoundation/tech-proposals/pull/51, I think we should perhaps close this proposal as out-of-scope for the GHC Steering Committee? As a change to GHC internal implementation details it almost doesn't require a proposal at all. However, it will necessarily make a very slight change to the `base` API, namely changing `Symbol` from a non-empty to an empty datatype, and thus might need to be run past the CLC.

adamgundry:
I don't know if @phadej has pushed a branch anywhere, but I believe literally the entire change envisaged here is replacing `data Symbol` in `GHC.Types` with `newtype Symbol = MkSymbol String`.

phadej:
@cdornan, @adamgundry is anyone working on this? Is anyone writing a CLC proposal, which apparently this change requires after all?

adamgundry:
Sorry for the delay @phadej. I'm not working on this, and thinking it over, I'm not convinced it needs a CLC proposal. The CLC seems to be quite burdened with the workload of responding to proposals, so I think it's better not to refer things that are not strictly in scope.

Assuming the constructor is exposed only from `ghc-internal`/`ghc-experimental`, then the only visible effect if you don't depend on those packages is that some (rare) programs that previously compiled without warnings will now compile with warnings. The fact that `case x :: Symbol of {}` currently does not emit a warning, when `Symbol` is morally an abstract type, seems to me a bug, and we can fix it by making the change you propose.

phadej:
> The fact that case x :: Symbol of {} currently does not emit a warning, when Symbol is morally an abstract type, seems to me a bug, ...

Or at least a missing feature. It's impossible to hide the fact that an empty type is empty. (There is no constructors to hide). Similarly, there is no way to make an empty `PatternSynonyms` match complete:

```haskell
module FastFin (Fin) where

import GHC.TypeNats

newtype Fin (n :: Nat) = MkFin Integer

{-# COMPLETE :: Fin 0 #-}
```

phadej:
> Assuming the constructor is exposed only from `ghc-internal`/`ghc-experimental`

I'd be happy with just exporting it from `ghc-prim`, where the type is defined currently. I don't know what's are the implications of re-exporting it from the `ghc-*` packages. As far as I understand the `ghc-internal` is as "don't use this package" as `ghc-prim` and AFAIU whether `Symbol` could be defined in `ghc-internal` or `ghc-prim` is arbitrary, but it is in `ghc-prim` because implicit params machinery (defined in `ghc-prim`) needs it.

adamgundry:
@phadej agreed on all counts. I wasn't meaning to imply a `ghc-internal` export would be preferable to `ghc-prim`; as you say they are both equally internal and the boundary between them is merely down to implementation details.

PR #564, opened 2022-12-16
Namespaces

kalhauge:
This is an compromise between #283 and #295. Essentially, it proposes creating a new entity called the `namespace` which can be used as an incremental way of introducing local modules. 

```haskell
namespace X where
  zero = 0

one = X:zero + 1
```

### DIFF
--- a/proposals/0000-namespaces.md
+++ b/proposals/0000-namespaces.md
@@ -1,0 +1,433 @@
+---
+author: Christian Gram Kalhauge
+date-accepted: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/0>).
+**After creating the pull request, edit this file again, update the number in
+the link, and delete this bold sentence.**
+
+# Namespaces
+
+One of the pain-points for building structured, succinct, and easy to read Haskell code is name conflicts. 
+The lack of a fine-grained namespace controls lead to developers either creating many small files or ad-hoc namespace measures as prefixing names to variable names.
+
+Many prior proposals have tried to come up with a solution to this, especially [#283](https://github.com/ghc-proposals/ghc-proposals/pull/283) and [#295](https://github.com/ghc-proposals/ghc-proposals/pull/295), however they have all gone dormant.
+This proposal, is yet another try.
+Contrary to the previous approaches, this proposal does not want to change the module system, but instead introduce a new construct: the namespace.
+A namespace is a (named) entity which captures the names of other entities (data types, types, and classes), so that they can easily be exported and referenced at will. 
+This proposal also introduces using another access qualifier for
+namespaces to differentiate them from modules (`:`).
+
+## Motivation
+
+The proposal will enable us to write succinct code and have better control of the exports.
+The motivation is a list of examples mirroring [#283](https://github.com/ghc-proposals/ghc-proposals/pull/283)
+
+### Example 1: Compact code
+
+If we want associate the same named function with two data structures, we'll have 
+to create two files, create a type class, or prefix the names of the function.
+
+With namespaces you can write this:
+
+```haskell
+module Pretty where
+
+space V2 where
+  data V2 = Mk { x :: Int, y :: Int }
+  pretty :: V2 -> Text
+  pretty = ...
+
+space V3 where
+  data V3 = Mk { x :: Int, y :: Int, z :: Int }
+  pretty :: V3 -> Text
+  pretty = ...
+
+main = print (V2:pretty (V2:Mk 0 1) <> V3:pretty (V3:Mk 0 1 0)) 
+```
+
+### Example 2: Associated Functions
+
+By using namespaces, we can associate specific functions with the data it operates
+on. The `= Set` notation means that, the namespace `Set` will act like the type 
+or class `Set` when not used as a qualifier, e.g., `Set:a`.
+
+```haskell
+-- in Data/Set.hs
+module Data.Set (Set) where
+
+space Set = Set (Set, fromList) where
+  data Set = ..
+
+  fromList :: [a] -> Set a
+  fromList = ... 
+
+-- in Main.hs
+import Data.Set (Set)
+
+mySet :: Set String
+mySet = Set:fromList ["A", "B"]
+```
+
+### Example 3: MyPrelude
+
+By using namespaces, we can collect names used often in a prelude and export them qualified.
+
+```haskell
+-- In MyPrelude.hs
+module MyPrelude ( BL, BS ) where
+
+import Data.ByteString.Lazy named as BL
+import Data.ByteString named as BS
+
+-- In Lib.hs
+module Lib where
+
+import MyPrelude (BL)
+
+toByteString :: Bool -> BL:ByteString
+toByteString b = case b of
+  True -> pack "True"
+  False -> pack "False"
+  where open BL (pack)
+```
+
+Also using the Hedgehog library could look like:
+```haskell
+module Spec where
+
+import Hedgehog (forAll, Gen, Range)
+```
+
+### Example 4: Local Data Types
+
+We can declare data types locally to a function without polluting the namespace.
+
+```haskell
+module Spec where
+
+open X (f) -- ^ instantly export f to the scope
+space X where
+  f = getCollection . foldMap MkCollection
+  data Collection = MkCollection { getCollection :: ... }
+```
+or if auto namespaces is turned on:
+```haskell
+module Spec where
+
+open X (f) -- ^ instantly export f to the scope
+space X where
+  f = Collection:get . foldMap Collection:Mk
+  data Collection = Mk { get :: ... }
+```
+
+## Proposed Change Specification
+
+Note: We're using the nomenclature and notation introduced by
+section 2 in [#283](https://github.com/ghc-proposals/ghc-proposals/pull/283), the section could be moved
+here for context if necessary.
+
+This proposal suggest doing the following in a nutshell:
+```haskell
+space A where
+  x, y :: Int
+  x = 0
+  y = 10
+```
+
+Create a new top-level-declaration, `space A where ...`, which can declare a new namespace.
+Every declaration inside the where clause is hidden from the rest of the module behind 
+the named-qualifier `A`. We can access the variables under `A` either by using 
+named quantification `A:x` or by opening the namespace. 
+Opening the namespace is a declaration which loads the variables into the relevant scope.
+It can be limited an import specification, e.g.:
+```haskell
+f = let open A in x
+f = x where open A
+f = x where open A (x)
+f = x where open A hiding (y)
+```
+
+A namespace is an entity so it can be parsed around and referenced like everything else in the module system. 
+This enables sharing of collections of names without real changes to the module system.
+
+### Details
+
+Adding a new language extension `-XNamespaces`, which activates the extension.
+This will enable the following syntactic changes. 
+We'll use pointers to [Haskell 2010](https://www.haskell.org/onlinereport/haskell2010/haskellpa1.html#haskellch5.html) syntax, when needed.
+
+#### Extensions to the Lexer ($2) and Expressions ($3)
+
+We'll add a new token `namid` to namespace which has the following production
+```
+namid → conid (namespace)
+```
+
+We'll add a lexing rule to that recognizes `nameid:varid` as three tokens 
+`namid`, `lookup_namespace`, and `varid`, and not as `namid`, `cons`, `varid`.
+Furthermore changes to all qualified names so they accept being qualified by namespaces.
+E.g., `A.B:X.z` is a valid qualified identifier, which should be read as the variable 
+`z` in the namespace `X`. 
+
+A Note on `Data.List.:` and on constructor operators. The choice of `:` as the delimiter is the 
+least worst choice. We'll use the same semantics as in Haskell 2010. `A::+` means the type operator 
+`:+` from `A:`, but we expect that people will mostly use operators unqualified.
+
+#### Extensions to Declarations and Bindings ($4)
+
+We'll add a new top-level namespace construct of the form:
+```
+topdecl 
+  → ...
+ | 'space' nameid ['=' conid] [exports] 'where' topdecls;
+```
+
+This will have the semantics of capturing the declarations in `topdecls` and encapsulating them from the rest of the module.
+The declarations in the `topdecls` can access the variables of the `topdecls`.
+
+It will then create a fresh variable `nameid` which is the namespace, which is limited by the `exports` 
+specification. 
+The `= conid` syntax allows the user to bind a type or class to the `nameid`.
+
+Example, borrowing the notation from [#283](https://github.com/ghc-proposals/ghc-proposals/pull/283), 
+where environments are mappings from qualified variables to original name `{ X |-> (M, m)}`.
+```haskell
+module M where
+space Truth = Bool (True, False) where
+  data Bool = True | False
+
+-- Environment outside Thruth is 
+-- { Truth |-> (M, Truth:Bool)
+-- , Truth:True -> (M, Truth:True)
+-- , Truth:False |-> (M, Truth:False)
+-- }
+```
+
+As is customs: It's okay to declare two namespaces of the same name, but referring to either would give an ambiguity error.
+
+We'll add a new declaration `open` construct of the form:
+```
+decl 
+  → ...
+ | 'open' nameid [impspec];
+```
+
+This will have the semantics of including everything in the namespace (restricted by the optional `impspec`) into the scope of the declaration.
+
+Continuing the example, opening truth gives the following:
+```haskell
+import M (Truth)
+open Truth (False);
+
+-- Environment is 
+-- { M.Truth |-> (M, Truth:Bool)
+-- , M.Truth:True -> (M, Truth:True)
+-- , M.Truth:False |-> (M, Truth:False)
+-- , Truth |-> (M, Truth:Bool)
+-- , Truth:True -> (M, Truth:True)
+-- , Truth:False |-> (M, Truth:False)
+-- , False |-> (M, Truth:False)
+-- }
+```
+
+#### Extensions to Modules ($5)
+
+We'll introduce a new import statement, of the form:
+
+`impdecl → 'import' modid 'named' ['as' nameid] [impspec]`
+
+Which saves the content of the module into a namespace.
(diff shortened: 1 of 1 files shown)


brandonchinn178:
Can't react on mobile, but 👍. I personally like `#`, I'd be ok with `\`, weakly against reusing colons

nomeata:
> Can't react on mobile

Rotate by 90° to widescreen, and the buttons likely appear

ocramz:
I find this proposal very confusing; the result looks like an entirely new language, at least on surface

kalhauge:
I have seen two kinds of constructive feedback across different different platforms. 

1. We don't like the `:` operator, use `#` instead. The `:` operator is used with lists and type constructors, using it with namespaces is one more problem. `#` is already used for advanced stuff which should not conflict with the notation, so it might be better for new developers. Do anybody have a good argument against using `#`, because otherwise I buy the argument and move the proposal to that operator instead?

2. Namespaces makes code harder to read. Specifically,
   - Searching code without an IDE, can be harder.
   - Documentation might be harder to get right.
   - The syntax would be prevalent, and not being used to it, it might look a little foreign. 

   First I want to say that the extension is optional and should not change the way we write code, even if turned on by default. A majority of Haskell code would still look the same. Though some library creators might assume that you can use namespaces to differentiate between similar named function, e.g., `State#run` and `StateT#run` (using the new syntax all ready :)).
I have also given an example in the proposal, to show how to use a library which exposes namespaces, without using the extension in your code.

   When that is said, it think the criticism is true. But I recon this is true for all extensions specifically the ones that handle naming. The question is if the extra syntactic load is worth the cost. My argument is; yes. As besides allowing Haskellers  to write code with the names they want, and enabling better tooling, it would be a consistent way of solving many of the issues Haskell has with name conflicts. 

@ocramz Is there anything I can do to make the proposal more clear to you? Would choosing a different operator help, or is it something more fundamental?

proposals/0000-namespaces.md
```diff
@@ -0,0 +1,778 @@
...
+module Vec where
+import Vec.V2 qualified as V2
+import Vec.V3 qualified as V3
+
+vec_0_1_1 :: V3.V3
+vec_0_1_1 = V3.fromV2 (V2.fromV2 (V2.Mk 0 1)) 1
+```
+
+However, in the last case you can't write a function `fromV3` in `Vec.V2` as it would 
+create a cyclic dependency.
```
chshersh, line 137: One technical clarification: GHC actually supports cyclic module dependencies so you **can** write `fromV3` in `Vec.V2`.

* https://downloads.haskell.org/ghc/latest/docs/users_guide/separate_compilation.html#how-to-compile-mutually-recursive-modules

kalhauge: Good catch! I'll note that it is possible using a `hs-boot` file.

konsumlamm:
There are several examples using `import ... named`, but I can't find any explanation of what that does.

In general, I'd really love having some sort of local modules or namespaces and I like this proposal. I think it might do a bit too much though, especially the `namespace Set = Set` syntax feels weird and ad hoc.

I'm not sure there's a good solution to the problem of which separator to use for namespaces. To me, `:` looks most fitting (after `.`, but that's ofc already used by modules) and alternatives like `#` or `\` look really off to me, although the clash with `(:)` is ofc a problem.

kalhauge:
@konsumlamm good catch, I have fixed it int the proposal, it was meant to be
```
import namespace X as ...
```
It changed to avoid an extra keyword.

> I think it might do a bit too much though, 

I really want to make the extension as small as possible. The core features are 
`namespace A where`, declaration `open A`, lookup`A:x`, and import and export of namespaces. Any features that takes away from that is definitely up for discussion. 

On the topic of `namespace Set = Set`. It is done to future proof the extension. I have added the following section:

> Using the `namespace Set = Set` notation is done to enable users to use the common Haskell
> pattern. 
> ```haskell
> import Data.Set (Set)
> import qualified Data.Set as Set
> ```
> Furthermore, it enables a straight forward syntactic description of turning all entities 
> into namespaces, see the `-XAutoNamespaces` extension.

However, if it is too confusing we can discuss removing it or moving it to another proposal.

proposals/0000-namespaces.md
```diff
@@ -0,0 +1,796 @@
...
+```haskell
+module Vec where
+
+namespace V2 where
+  data V2 = Mk { x :: Int, y :: Int }
+  fromV2 :: V2 -> V2
+
+namespace V3 where
+  data V3 = Mk { x :: Int, y :: Int, z :: Int }
+  fromV2 :: V2 -> Int -> V3
```
ocharles, line 80: Shouldn't this be

```suggestion
  fromV2 :: V2:V2 -> Int -> V3
```

I may be misunderstanding the proposal, but I don't see why `V2` is in scope here otherwise.

kalhauge: Yes, you are correct. That example have been changed a lot. It's fixed now.

alt-romes:
I‘m somewhat against this proposal.

I agree with the general motivation: our  module system isn’t perfect and it makes programmers choose a suboptimal solutions in designing programs.

However, I don’t yet see how creating a separate complex system (namespaces) that somewhat ignores modules and tries to do its own thing isn’t a language improvement; even more so when it’s proposed that eventually one could morph into the other. It seems we’re deviating from the core problem and putting a burden on code complexity and language complexity.

I’m set back in particular by not understanding the clear distinction between modules and namespaces, and why we need both!

kalhauge:
Thank you for your thoughtful feedback @alt-romes, and thank you for reading the proposal.

> I’m set back in particular by not understanding the clear distinction between modules and namespaces, and why we need both!

It's definitely my mistake for not explaining this better; also I understand why It's not obvious why we need both. In the current module systems, modules are tightly coupled with the files and with compilation units; and are mostly namespaces by accident. The thing is that modules=files, should not be modules=namespaces; because that makes the granularity of namespaces at the level of the files.

The problem now two fold: 1. we would like more granular modules=namespaces and 2. we want to stay backwards compatible.  The two other proposals that I link to in the descriptions tries to extend the current module system to make it better; however the solutions either became too complex or not compatible with the current way we write Haskell.

This proposal tries to fix this by adding the construct `namespace` to mean a collection of names without changing the meaning of a "module". That way a module can continue to mean file / compilation unit.

> even more so when it’s proposed that eventually one could morph into the other. It seems we’re deviating from the core problem and putting a burden on code complexity and language complexity.

Let me continue, and answer this part. The problem is that now we have two things that can mean a collection of names, 
one is a "namespace" and one is "module". The last part of the proposal just illustrates that the current "module" could focus on only being a file / compilation unit (by renaming it to file), and the namespace is sufficient to replace it. I'm sorry this was not clear from the text.

Does this clear up the difference?

AntC2:
> @kalhauge ... a module can continue to mean file / compilation unit.

That's not exactly what a 'module' means today. Language report § 5.3.3 Local aliases:

> It is legal for more than one module in scope to use the same qualifier, provided that all names can still be resolved unambiguously.

So the 'qualifier' (the _modid_ following `as` in an `import`) in effect is acting as a namespace. The weakness in current Haskell is it doesn't enforce consistency across _modid_ s in different modules in a project. OTOH, addressing that only needs using the same bunch of `import`s everywhere.

#283 addresses this by putting the `as` on the export. Does this proposal have a better way to enforce consistency?

kalhauge:
As I wrote above "modules" today are file / compilation units and namespaces by accident. 
You can create aliases of a module, but you can't use different namespaces inside of a file.

@AntC2 I'm not sure that you can get consistency in Haskell or that #283 `as` gives you it. 
The best example of consistency would be Java. Where every class is explicitly imported in the 
top of the file and not renamed.

Three ways that this proposal would improve consistency: 
1. Naturally more names could fit in a single file; and since we have fewer modules we have more consistency.
2. A module can export namespaces under standard names; users are less likely to rename those. Also see the prelude example. You could have one of these in each project that define all imports of the project; and only import that.
3. A developer can namespaces namespaces, this means that developer are less likely to rename them.

A big way it would reduce consistency is that you can rename and repackage namespaces. **With great power comes great responsibility.**

kalhauge:
Actually, the last example can be done even more cleanly, however not 
completely supported by the proposal yet: imports can't be nested in 
the namespaces yet, but maybe they should.

```haskell
-- External.hs
module External (Containers, MTL) where

namespace Containers where
  namespace Set = Set where import Data.Set
namespace MTL where
  namespace Writer = Writer where import Control.Monad.Writer
  namespace State = State where import Control.Monad.State

-- in file1
import External (Containers (Set), MTL (Writer))

tellNothing :: Writer (Set x) ()
tellNothing = tell (Set:empty)
 where open Writer (tell)

-- in file2
import External (MTL(State))

incr :: State Int Int
incr =  state (\i -> (i, i + 1)
  where open State (state)
```

noughtmare:
I don't fully understand why `.` can't be used for namespaces as well. You only give the example:

```
import Data.Set 
import Data

-- Is this a module lookup of 'x' from Data.Set or 
-- a namespace lookup of 'Set.x' from Data. 
Data.Set.x
```

Can't we just postpone failure until there is actual ambiguity like we already do when two qualified names conflict or when two module names from different packages conflict?

AntC2:
> I don't fully understand why `.` can't be used for namespaces as well.

That's what #283 tries to do. Note that currently, the `module` header and all the `import`s come at the top of the file. So we know all the possible module names before seeing usages. With this proposal (or #283) wanting to:

>  use different namespaces inside of a file.

Specifically, a datatype name and/or its constructors might generate namespaces (or `modid`s in #283). And the point of declaring that might be at the bottom of the file/after all its usages. So

`Data.Set.Baz.quux` might be

* item `quux` in namespace `Data.Set.Baz`;
* data constr.field `Baz.quux` in module `Data.Set`;
* ?data type.field `Baz.quux` in module `Data.Set` (I think you can make a type alone a namespace, without the constrs);
* data type.constr.field `Set.Baz.quux` in module `Data`.
* Or the coder didn't intend to switch on 283's Local modules at all/it's a typo attempting to use `.` as function composition, the compiler should give a helpful error message without suggesting a newbie switch on an advanced extension that will only lead to more pain.

At least using a different char to demark namespaces shows the coder did mean to use this new feature.

But still attempting to put a namespace prefix might suffer from a usage site appearing before the declaration, so the compiler must postpone name resolution until it's got to the end of the file and knows all namespace decls.

To correct something you said:

> or when two module names from different packages conflict?

I quoted from the [Language report § 5.3.3 above](https://github.com/ghc-proposals/ghc-proposals/pull/564#issuecomment-1358557508). There's currently no notion that module names "conflict" -- that is, if you `import` two different modules and give them the same `modid`, that's fine. Even if they contain same-named items, that's fine. It's only a problem if you try to refer to one of them.

kalhauge:
@noughtmare, I must admit that adding something that might introduce conflicts was disqualifying for Haskell. However, I can see that it is, as everything, a tradeoff.

I think @AntC2 answer explains the complexity of resolving names from two namespaces using the `.` operator. 

I would like to add that a disambiguate checker might break backwards compatibility:

```haskell
import Data.Set (Set)
import qualified Data.Set as Set

length :: Set.Set a -> Int
length = Set.size
```

Now enabling the extension produces two errors. Set.Set might point to the type of Set namespaced in module Set or in the associated namespace of Set and Set.size is both available in the module Set and the associated namespace for Set. The extension can therefore not be active by default.
Also, I expect there to be many more weird cases of seemingly valid Haskell programs that break with the extension on.

But, maybe a little breaking is acceptable to keep the `.` operator for all namespaces?
I lean towards having a new operator for a new thing, but I recognize that many people want to keep `.` 🙂

kalhauge:
Okay after a relatively positive reception on [reddit](https://www.reddit.com/r/haskell/comments/zp0g33/proposal_namespaces/), and a lukewarm reception here I have tried to take all the feedback you have given me and incorporated it in a new draft.

[Draft](https://github.com/kalhauge/ghc-proposals/blob/sketch/proposals/0000-namespaces.md)

I have not included it in the proposal yet, as I have only updated the motivation, and I'm still working out the details.
It's a little hard to parse the exact criticism captured by @ocramz comment, but here is what I have done:

1. I have made the motivation more incremental, explaining which features are added where and why. If the proposal still does not make sense to some of you, I would like to know where I loose you :). I'm guessing people are fine with the first couple of steps but less so when the  new operator gets introduced.
2. I have changed the default operator from `:` to `#`, it is slightly harder to read, but also slightly less destructive.
3. I have made namespaces unqualified by default; this enables the next step:
4. I realized that the `namespace` was unnecessary if we simply allow associated namespaces for all type-level entities. 
    ```haskell
    type A = A qualified where
      data A = B 
      x = 0
    --- same as the old notion: 
    namespace A = A where
       data A = B
       x = 0
    ```
    This makes the namespaces feel as more of an integrated part of Haskell, than something added later.
5. I have moved the `open` construct to it's own language extension. I have done this for two reasons. `open` is a keyword that might not be backward compatible and because the `open` construct can make tracability harder in the program.
6. Finally, have I added text about why `.` is a bad idea, but also suggest that it can be activated using an extra extension `-XAmbigiousNamespaces`

kalhauge:
@noughtmare I think you make some really good arguments; and I have a hard time coming up with some very bad examples; but this of cause does not mean they don't exist.

For now I have created an optional extension `-XAmbigiousNamespaces`, in which I discuss the drawbacks. I'm sure somebody with more knowledge about the two other proposals eventually will weigh in.

Merivuokko:
I find the new changes to this proposal in kalhauge/ghc-proposals to be very confusing. Especially the qualified types, as it messes up the concepts of first-class namespaces and types with local namespaces.

I would prefer that namespaces would be qualified (except for anonymous namespaces with export lists).
The notation `namespace X where` is very intuitive, and it resembles – for example:
C++: `namespace X { ... }'
Rust: `mod X { ... }`
Ada: `package X is ... end X;`
Ruby: `module X ... end`

I would not want to see the `type ... qualified where` syntax in Haskell, because:

1) It would be related to the `namespace ... where` syntax, but still it looks very different.
2) A declaration that starts with the keyword `type` should declare just a type or a typeclass, not variables.
3) The keyword `qualified` appears close to the end of line where it is easy to miss, if one is just quickly viewing the code to find declarations.

I would not want to see unqualified namespaces either. As I understand it, the whole point of namespaces is to allow definitions to be scoped logically to be accessed qualified in a natural way. Unqualified namespaces would only serve the purpose of keeping unwanted declarations away from the top-level scope, which is worthwhile, but qualified namespaces fulfill this goalas well.

I am also against the # notation. The dot operator in Haskell is messy and confusing. It is, however, already a part of an official language specification in double role, and we are not likely going to get rid of this confusing situation.

I believe that if we add a new namespace lookup operator ( # or : ), we are going to end up with much more confusion. Operators with double role are confusing to everybody, not just beginners. It also requires increased lexing work, both from humans and the compiler. # is verbose too. If we had a a free operator that could be defined as module/namespace lookup operator, I would support using it. But then I would wish that all module and namespace lookups would use that same operator, and dot would be left to refer to record lookup only.

In my opinion modules and namespaces should be made as similar as possible, when it comes to their usage and definition syntax. I would reserve the `import` keyword for importing compilation units and utilize the `open` keyword for bringing items from a module/namespace into scope.

Thus:
```
import Data.Set qualified as Set
open Set
```

would have the same meaning as

```
import Data.Set as Set
```

Modules could be viewed as compilation units that always contain an anonymous namespace. Compilation units can be imported with the `import` keyword which also brings the anonymous namespace from the module to scope, possibly naming it as well (with `as`), and possibly making this namespace only accessible qualified (by `qualified`).

A qualified import may thus create a hierarchy of namespaces, e.g. `import Data.Set qualified` would create two namespaces – `Data` which contains `Set`. A further import of Data.Vector would add `Vector`to the `Data` namespace.

Thus compilation unit names (module names) would be related, but still separate – but this would not really change the current situation.

Currently you cannot do this:

```
import Data.Text qualified as T
import T.Lazy qualified as TL
```

Inner namespaces would not change this in any way, because import always looks up a compilation unit. Thus it would not be possible to do:

```
import Data.Text
import Lazy -- Even though contents of Data.Text are in scope, there is no compilation unit named `Lazy` anywhere
```

Backwards compatibility is important to me, but not at all as important as having a good support for scoping definitions. I want the feature to be cleanly designed and intuitively understandable both to beginners and seasoned Haskellers.

kalhauge:
@Merivuokko, sorry I haven't been able to get back to you before now.

You bring up some good points: 

> I would not want to see the type ... qualified where syntax in Haskell, because:

Given some time, and looking back on this, I tend to agree. However, to minimize the effect of enabling the extension, this behavior is first introduced in the `-XAutoQualified` extension. Therefore a user of a namespaced haskell library can import only ability to access the namespace.

The reason why `type` is a namespace by itself solves two things. 

1) It creates a uniform definition of namespace: Every type has a namespace associated with it; which can be populated using the `where` keyword. Since the `type` keyword is already used for naming it is an obvious choice.
2) It creates a simple separation of type and value constructors. Type constructors are often toplevel `Type` and constructors would often be lower level `Type.Con`. This could be helpfull to create more idiomatic dependent Haskell.

> I am also against the # notation.

I have also changed my opinion on this as it looks like it has unambiguous meaning in some corner cases. I think the best option for an alternative syntax is still either `:` or `\`.

> I believe that if we add a new namespace lookup operator ( # or : ), we are going to end up with much more confusion.

My hope was that this extension would be an temporary measure until modules where no longer considered namespaces. That would free up the `.` notation. See the suggested extensions `-XAmbiguousNamespaces` and `-XNoModuleNamespaces`.

> Backwards compatibility is important to me, but not at all as important as having a good support for scoping definitions. I want the feature to be cleanly designed and intuitively understandable both to beginners and seasoned Haskellers.

I fundamentally agree. I think the problem is that for this change to take place it needs to be incremental so that the community can experiment with the feature and we can correct it before it becomes solidified. This is the reason for the confusing structure of this proposal.

I think that my primary thesis and the reason for this proposal is that I think we eventually should separate modules and namespaces, so that we can avoid all the nasty corner cases involved in the import of modules. 

The endgame for me looks like this, but I'm aware that it will take years and trial and error to get there:

```haskell
from "containers" import "Data/Set" (Set)
-- Imports the namespace/type Set from the "Data/Set.hs" module in the 
-- "containers" package, and all type instances (incuding orphans).
from "text" import "Data/Text" (Text) 
-- Import the Text type from "Data/Text.hs".

module Utils where
  listToSet :: Ord a => [a] -> Set a
  listToSet = foldMap (Set.singleton) 

-- We'll use Text a lot and we are lazy, so let's use T instead.
type T = Text

countUniqueWords :: T -> Int
countUniqueWords = Set.size . Utils.listToSet . T.words
```

PR #568, opened 2023-01-06
Add SPECREC proposal

AndreasPK:


proposals/0568-specrec.md
```diff
@@ -0,0 +1,247 @@
...
+
+# Transitive specialization support.
+
+Implement a `{-# SPECREC foo #-}` pragma. Using such a pragma on a function will encourage GHC to specialize
+all uses of `foo`, expose foos unfolding **and also do so for all use sites of `foo`** in a recursive fashion.
+
+That is if we have
+
+```haskell
+{-# SPECREC foo #-}
```
nomeata, line 18: Is that right, or should it be
```suggestion
{-# SPECREC elem #-}
```

proposals/0568-specrec.md
```diff
@@ -0,0 +1,247 @@
...
+It will act as if we had written:
+
+```haskell
+{-# SPECREC foo #-}
+elem :: Eq a => a -> [a] -> Bool
+elem = ...
+
+{-# SPECREC foo #-}
+elems :: (Foldable t, Eq a) => a -> t [a] -> Bool
+elems x = any (elem x)
```
nomeata, line 38: In the “as if” section I’d expect `SPECIALIZE`, not the `SPECREC` you are introducing.

phadej:
So this is recursive version of SPECIALIZABLE as in https://github.com/ghc-proposals/ghc-proposals/pull/357, not `SPECIALIZE`?

(As [`SPECIALIZE`](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/pragmas.html#specialize-pragma) asks to upfront specialize the binding)

---

Note: INLINE[ABLE] exposes unoptimize unfoldings, so there are no optimizations happening in between. So

> It's generally possible to achieve a similar effect by applying INLINE[ABLE] pragmas to every function in the call graph. This is somewhat error prone and it can be hard to verify that it works as expected.

Is not true. #357 `SPECIALIZABLE` would achieve similar effect.

---

I don't like having only rec variant. #357 seems to be better (and *that is not possible today*). 

---

There should also be pragma to unmark `SPECREC` effect. It might be pessimisation.

---

In example

```haskell
{-# SPECREC elem #-}
elem :: Eq a => a -> [a] -> Bool
elem = ...

elemList :: Eq b => [b] -> [[b]] -> Bool
elemList = elem
```

Is `elemList` still marked as `SPECREC` or not? The `elem` is already specialized / optimized in `elemList`? The proposal is not clear.

AndreasPK:
> So this is recursive version of SPECIALIZABLE as in https://github.com/ghc-proposals/ghc-proposals/pull/357, not SPECIALIZE?

> (As [SPECIALIZE](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/pragmas.html#specialize-pragma) asks to upfront specialize the binding)

Yes.

>> It's generally possible to achieve a similar effect by applying INLINE[ABLE] pragmas to every function in the call graph. This is somewhat error prone and it can be hard to verify that it works as expected.

> Is not true. https://github.com/ghc-proposals/ghc-proposals/pull/357 SPECIALIZABLE would achieve similar effect.

My definition of similar here was only in regards to "will these functions be specialized". Not about weither or not we expose the optimized/unoptimized unfoldings and differences that come with that. But there are differences yes. I would still say it's *similar* but not *the same effect*. I will adjust the language to be clearer.

> I don't like having only rec variant. https://github.com/ghc-proposals/ghc-proposals/pull/357 seems to be better (and that is not possible today either).

Do you mean having only SPECIALIZABLE is better than this proposal. Or that this proposal should be an extension of https://github.com/ghc-proposals/ghc-proposals/pull/357 (essentially having both)

> There should also be pragma to unmark SPECREC effect. It might be pessimisation.

I agree. Will update.

> I'd prefer pragma to be named SPECIALIZABLE_REC - being clear is more important than being concise.

Also seems reasonable.

proposals/0568-specrec.md
AndreasPK: I've expanded on the differences.

phadej:
> Do you mean having only SPECIALIZABLE is better than this proposal. Or that this proposal should be an extension of https://github.com/ghc-proposals/ghc-proposals/pull/357 (essentially having both)

Essentially having both.

simonpj:
So if I understand correctly, if I put `SPECIALISE_REC` on (say) `elem` then every function `f` (no matter how large) that calls `elem` will also be specialised, and every function `g` (not matter how large) that calls one of those functions `f`, etc.  So the transitive upward closure of all functions that call `elem`, not matter how indirectly, in other libraries, in client code, in code not yet written.   

That's a pretty big effect for a single pragma written in the `base` library. 

Moreover, it's very selective: you pick a particular function for this treatment. But which function(s)?  I want to pick the ones that are in the inner loop of my program; but the library author doesn't know which will be in the inner loop.

> Using -fexpose-all-unfoldings -fspecialise-aggressively / Whole program compilation...generally comes with a very heavy compile time cost and therefore is not as commonly used

Do you have data to support this claim?  Why do you think that `SPECIALISE_REC` will be cheaper? After all, it is more or less what you are suggesting!

My instinct is this:

* It is the client, not the library author, who should decide this.
* It should be easy for a client to say "please compile my entire program, libraries and all, doing as much type-class specialisation as you can".  Compile time increases, but runtime should decrease, and you get "specialisation is guaranteed" as desired by that Tweag post.

Bodigrim:
>  It is the client, not the library author, who should decide this.

It is difficult in practice to educate and encourage clients to enable `-fspecialise-aggressively`, even if they are direct consumers of your library, saying less about transitive consumers. For many applications it is absolutely crucial to specialise and you never want to impose 10x-100x penalty on those unlucky majority who did not enable `-fspecialise-aggressively`.

AndreasPK:
> Moreover, it's very selective: you pick a particular function for this treatment. But which function(s)? I want to pick the ones that are in the inner loop of my program; but the library author doesn't know which will be in the inner loop.

It's should be fairly safe to apply it to any potentially performance sensitive function that's overloaded. I don't think that this is any harder to judge than the same decision for INLINE or INLINEABLE.

> > Using -fexpose-all-unfoldings -fspecialise-aggressively / Whole program compilation...generally comes with a very heavy compile time cost and therefore is not as commonly used

> Do you have data to support this claim?

The compile time cost is easy to check. E.g. when building the Cabal library without any of it's dependencies it's build time increases by around 30% for `-fexpose-all-unfoldings -fspecialise-aggressively`.

It's harder to accurately judge use of this combination obviously. But for three concrete data points:

* When tweag worked on optimizing the backend pretty printer for GHC they choose adding INLINEABLE/SPECIALIZEABLE over these flags.
* GHC itself is not build with these flags.
* This combination of flags is currently broken for ghcs master branch. The lack of coverage of this feature to me also weakly implies that use is not all that common. But admittedly this is not a very strong link.

> Why do you think that SPECIALISE_REC will be cheaper? After all, it is more or less what you are suggesting!

`-fexpose-all-unfoldings` will expose *all* unfoldings while `SPECIALISE_REC` will only expose a subset of these bindings in any given module. So it seems like a reasonable inference.

Granted we narrow the gap by introducing a flag like `-fexpose-all-overloaded-unfoldings` or some such that would only expose unfoldings which are actually useful for specialization. Then I could see there being a smaller difference. And perhaps we should add such a flag. I will add this to the alternatives.

> It should be easy for a client to say "please compile my entire program, libraries and all, doing as much type-class specialisation as you can". Compile time increases, but runtime should decrease, and you get "specialisation is guaranteed" as desired by that Tweag post.

That is fair. But it's unclear to me if this is optimal. As @Bodigrim said in practice users are often unlikely to take active steps to ensure this. 

Broadly speaking this is a question of what the default should be. Should users expect code to monorphize by default or is it the responsibility of the user to use the right flags to ensure this.

phadej:
@AndreasPK 
> * When tweag worked on optimizing the backend pretty printer for GHC they choose adding INLINEABLE/SPECIALIZEABLE over these flags.

There isn't `SPECIALIZABLE` in GHC (is there?). Tweag used `SPECIALIZE` i.e. manually specializing things (up front?)

And then you rite `SPECIALISE_REC`, while proposal talks about `SPECIALIZEABLE_REC`.

Now it feels that these pragmas should have competely different base word, so they don't get confused.

AndreasPK:
> Now it feels that these pragmas should have competely different base word, so they don't get mixed up.

We could also use modifiers on `INLINEABLE` instead.

* `INLINEABLE` -> Expose the original rhs, keen to specialize only the specific binding.
* `INLINEABLE OPT`/`INLINEABLE UNOPT` -> Expose the optimized/unoptimized rhs, keen to specialize only the specific binding.
* `INLINEABLE REC` -> This proposal (which could be combined with OPT of course.

phadej:
> We could also use modifiers on INLINEABLE instead.

But that would miss *Using such a pragma on a function will encourage GHC to specialize
all uses of `foo`* won't it, i.e. `-fspecialize-aggressively` bit? 

Also `INLINEABLE REC` would be `INLINEABLE OPT REC` by default (to be consistent), is that good idea?

phadej:
FWIW, `OPAQUE` is one way to stop effect of proposed pragma propagating further. But it is a bit of a blunt one :)

Bodigrim:
> So if I understand correctly, if I put `SPECIALISE_REC` on (say) `elem` then every function `f` (no matter how large) that calls `elem` will also be specialised, and every function `g` (not matter how large) that calls one of those functions `f`, etc. So the transitive upward closure of all functions that call `elem`, not matter how indirectly, in other libraries, in client code, in code not yet written.

Think not about `elem`, think about `sort`:

```haskell
sort :: Vector v a => v a -> v a
```

The type class `Vector v a` encapsulates all basic vector manipulations. Thus it is critical to get it specialized so that, say, primitive unboxed vectors of `Int` are indexed efficiently. However, typically your `sort` is too large for GHC to specialize it automatically, and for the same reason you don't want to inline it everywhere it is used. 

Yes, we want all functions that call `sort`, no matter how indirectly and where, to be specialized, because otherwise performance penalties are too severe. It is reasonable to take this decision on the library's side.

AndreasPK:
> But that would miss _Using such a pragma on a function will encourage GHC to specialize all uses of `foo`_ won't it, i.e. `-fspecialize-aggressively` bit?

In that case this would depend on the used modifiers yes. If it's a different pragma or a "flag" on an existing pragma is really just bikeshedding.

AndreasPK:
For future reference there is a proof of concept here https://gitlab.haskell.org/ghc/ghc/-/merge_requests/9642

Haven't yet benchmarked anything with it yet.

jberryman:
(from the perspective of someone with medium sized mono-repo that builds several executables, with large number of transitive dependencies...)

I think I agree with Simon that ideally we would specialize from the top down: see a hot function in a tricky profile, say, and pass a single flag or annotation to specialize everything transitively.

In the `sort` example, what if we're sorting `Vector [[[[Int]]]]`? Under the current proposal I don't think those comparison ops are guaranteed to be all specialized (though incidentally they are likely inlined/exposed because likely small?). I think `sort` is an argument for top-down, but maybe I'm misunderstanding.

-----

But the proposal as-is is a huge improvement in terms of composability and making these optimizations less fragile.

The one thing I think it needs is some way for the top-level consumer (me/my team) to pass a ghc flag (e.g. in cabal.project) to manipulate a function or entire class in a library dependency. That is: please don't make us fork a library just to add a pragma to get this "transitive specialization".


 `-fexpose-all-overloaded-unfoldings` sounds interesting to try too.

-----

Optimized vs unoptimized unfoldings: I don't understand the subtleties of why you'd want one or the other here, but is there an opportunity to avoid compilation or optimization (now or in the future) if unoptimized unfoldings are exposed?

PR #572, opened 2023-01-23
By-laws: Allow out-of-season self-nominations

nomeata:
Wording suggested by @simonpj .

### DIFF
--- a/committee.rst
+++ b/committee.rst
@@ -143,8 +143,8 @@ is not fixed.
 The nomination and voting process is kept private, by using direct
 email to committee members, not the mailing list. 
 
-Additionally, anyone may self-nominate themselves anytime and trigger a vote. 
-Such "out-of-season" appointments are at the discretion of the committee, and
+Additionally, anyone may self-nominate themselves anytime (by email to the secretary). 
+Such "out-of-season" appointments are at the discretion of the secretary, and
 would typically be made to fill skills gaps on the committee, or if a previous
 season was under-subscribed so that the committee was short of members.
 
--- a/committee.rst
+++ b/committee.rst
@@ -112,6 +112,7 @@ of thoughtful contribution in a public space carries the most weight.
 Membership on the committee comes with a three-year term, extended so
 that a term expires only at the end of a nomination process. Call a
 member who has served for over three years an "expiring" member.
+
 A nomination process is triggered when the number of unexpiring members is
 about to drop below 9.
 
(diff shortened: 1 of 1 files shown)


committee.rst
```diff
@@ -142,6 +143,11 @@ is not fixed.
 The nomination and voting process is kept private, by using direct
 email to committee members, not the mailing list. 
 
+Additionally, anyone may self-nominate themselves anytime and trigger a vote. 
+Such "out-of-season" appointments are at the discretion of the committee, and
```
aspiwack, line 147: I'm not sure I'm following what the intent is here. The first sentence in the paragraph says that an out-of-season nomination triggers a vote. The second seems to mean that the committee decides whether it votes or not.

There is more than one way this can be interpreted. One is that, in case of an out-of-season nomination, the committee decides whether or not to hold a vote (that being said, how do they do that? are they holding a vote about the vote? I guess that the difference here would be that this is a decision about the relevance of an extra seat rather than on the particular person). Another decision is that the committee's vote (triggered by the first sentence) has, as an option, not to elect the nominee; though it would be trivial, since there are only two options in such an out-of-season vote (elect or not).

committee.rst
```diff
@@ -142,6 +143,11 @@ is not fixed.
 The nomination and voting process is kept private, by using direct
 email to committee members, not the mailing list. 
 
+Additionally, anyone may self-nominate themselves anytime and trigger a vote.
```
aspiwack, line 146: How, concretely, does a volunteer self-nominates out of season?

committee.rst
nomeata: Maybe it would be better to say “at the discretion of the secretary” or “the chair”? Some simple mechanism to avoid abuse of the process?

committee.rst
nomeata: Good point. By email to the secretary? (Public email would be bad.)

aspiwack: An email to the secretary is probably the right way, indeed.. We just need to make it explicit.

committee.rst
aspiwack: Fair enough. If the trigger is an email to the secretary, they can also be the filter. It would make sense.

committee.rst
nomeata:
```suggestion
Additionally, anyone may self-nominate themselves anytime (by email to the secretary). 
```

committee.rst
nomeata:
```suggestion
Such "out-of-season" appointments are at the discretion of the secretary, and
```

PR #573, opened 2023-01-25
Remove dot from characters allowed in overloaded labels

adamgundry:
This is necessary to avoid breaking existing code using the `(#.)` operator. See discussion at https://github.com/ghc-proposals/ghc-proposals/pull/170#issuecomment-1400837868.

cc @howtonotwin @goldfirere


int-index:
LGTM

simonpj:
I'm strongly supportive of fixing this.  I don't feel close enough to have a well-informed opinion about the details of the fix.

Thanks to Adam for proposing something concrete.  +1 from me.

adamgundry:
@nomeata This change is urgent to allow GHC 9.6 to be released without causing breaking changes. I don't know how much of the full process it needs to go through, but I'd like to submit to the committee.

PR #576, opened 2023-01-28
Adjust warning category syntax

adamgundry:
This adjusts the syntax of warning categories in #541 so that it is easier to parse and less ambiguous (both for computers and humans).

The original proposal allowed a module to be annotated with
```hs
    {-# WARNING [x-some-category] "message" #-}
    {-# WARNING [x-some-category] ["message1", "message2"] #-}
```
with square brackets used for both category and a list of message strings. This changes the syntax to use the `in` keyword and a double-quoted string, like this:
```hs
    {-# WARNING in "x-some-category" "message" #-}
    {-# WARNING in "x-some-category" ["message1", "message2"] #-}
```

This arose while working on the implementation (https://gitlab.haskell.org/ghc/ghc/-/merge_requests/9684).

[Rendered proposal with changes applied](https://github.com/adamgundry/ghc-proposals/blob/tweak-warning-category-syntax/proposals/0541-warning-pragmas-with-categories.rst) / [Rendered rich diff](https://github.com/ghc-proposals/ghc-proposals/pull/576/files?short_path=11cdc32#diff-11cdc32f68fb700f50705fee9fdca2fec3fadf6893fe455245aadaf4bc5e6de8)

### DIFF
--- a/proposals/0541-warning-pragmas-with-categories.rst
+++ b/proposals/0541-warning-pragmas-with-categories.rst
@@ -45,8 +45,8 @@ with a "warning category".  Users can then enable or disable these
 categories individually.  For example, it would allow the core-libraries
 proposal to introduce::
 
-    {-# WARNING [x-partial] head "This is a partial function, it throws an error on empty lists." #-}
-    {-# WARNING [x-partial] tail "This is a partial function, it throws an error on empty lists." #-}
+    {-# WARNING in "x-partial" head "This is a partial function, it throws an error on empty lists." #-}
+    {-# WARNING in "x-partial" tail "This is a partial function, it throws an error on empty lists." #-}
 
 Such warnings are enabled by default, so the appropriate warning would be
 displayed at any use of ``head`` or ``tail``.  However, users may control this
@@ -65,12 +65,11 @@ distinguish it from the built-in warning flags:
 Proposed Change Specification
 -----------------------------
 
-A warning *category* is an identifier beginning with a letter and followed by
-valid identifier characters or dashes.
+A warning *category* is a string consisting of valid identifier characters or dashes.
 
 #. A `WARNING pragma
    <https://downloads.haskell.org/ghc/9.4.1/docs/users_guide/exts/pragmas.html#warning-deprecated-pragma>`_
-   may be immediately followed by a single warning category in square brackets.
+   may be immediately followed by the ``in`` keyword and a single warning category in double quotes.
    See below for details of the grammatical changes.
 
 #. Individual warning categories may be enabled or disabled using new
@@ -125,10 +124,9 @@ Grammar of warning declarations
 The grammar of declarations is extended as follows:
 
 ============  =  =====================================================================
-*decl*        →  ``{-# WARNING`` [*categories*] [*things*] *strings* ``#-}``
+*decl*        →  ``{-# WARNING`` [``in`` ``"`` *category* ``"``] [*things*] *strings* ``#-}``
 
-*categories*  →  ``[`` *category* ``]``
-*category*    →  (*small* | *large*) { *small* | *large* | *digit* | ``'`` | ``-`` }
+*category*    →  { *small* | *large* | *digit* | ``'`` | ``-`` }
 
 *things*      →  *thing1*, ..., *thingN*
 *thing*       →  *varid* | *conid*
@@ -136,11 +134,14 @@ The grammar of declarations is extended as follows:
 ============  =  =====================================================================
 
 The category can be omitted entirely, so this subsumes the existing
-syntax for ``WARNING`` pragmas.  That is, [*categories*]
-means that the presence of the *categories* non-terminal is optional; if present
-it is a single category in square brackets (hence the use of typewriter font ``[``
-... ``]`` in the definition of the *categories* production).
+syntax for ``WARNING`` pragmas. If present,
+the ``in`` keyword is followed by a single category as a double-quoted string.
 Similarly the list of *things* is optional as it may be omitted (for a ``WARNING`` on a module header).
+
+The *strings* may be a single string or a list (with the latter giving a
+multi-line warning message).  Since a module header may have a pragma with no
+*things*, e.g. ``{-# WARNING "message" #-}`` or ``{-# WARNING ["message1", "message2"] #-}``,
+we use the ``in`` keyword to indicate the presence of a warning category.
 
 The *category* non-terminal subsumes both *varid* and *conid*, so it is
 possible to use the name of the thing to which a warning is being attached as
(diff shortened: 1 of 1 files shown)


nomeata:
I wonder why we needed the `["line 1" "line2"]` syntax anyways, when these are strings and you could put `"\n"` in them (and use line continuations for source level breaks). Oh well.

Would it be too much magic to use
```
{-# WARNING ["x-some-category" "message"] #-}
```
and tell a category from a message line via the `x-` prefix?

int-index:
> Incidentally, it would be very convenient if categories were lexed as string literals

Yes, I highly support using string literals instead of trying to hack the parser/lexer to accept those hyphens.

int-index:
What about

```
 {-# WARNING in "x-some-category" "message" #-}
```

The `in` keyword marks that the next string literal specifies the category, not the message. We could keep the square brackets if desired

```
{-# WARNING [in "x-some-category] "message" #-}
```

nomeata:
LGTM

adamgundry:
I like the idea of `in` and a string literal, not least because having hacked the lexer to accommodate categories as tokens, I've realised how painful it is!

There are a couple of corners to think about:
 * For warnings on identifiers we need to list them, so I guess we have a few options:
    ```hs
    {-# WARNING in "x-partial" head, tail "message..." #-}
    {-# WARNING head, tail in "x-partial" "message..." #-}
    {-# WARNING head, tail "message..." in "x-partial" #-}
    ```
 * It turns out that GHC allows semicolons to separate multiple warning definitions in a single pragma. So we'd have something like
     ```hs
   {-# WARNING in "x-foo" foo "message for foo..."
               ; in "x-bar" bar "message for bar..."
       #-}
      ```
    (although I wonder if we shouldn't just forbid such declarations, see https://gitlab.haskell.org/ghc/ghc/-/issues/22733).
 * Normal pragmas are case-insensitive, but presumably `in` would remain case sensitive?

nomeata:
I'd give a fixed order to keep things simple. I'd put the category first.

I think we can remove the semicolons.

Case sensitivity of stuff inside the pragma is fine, see `forall` in a RULE.

adamgundry:
Thanks, I've modified the proposed change to use the `WARNING in "x-category"` syntax.

nomeata:
As the shepherd of the previous version I'm tempted to just wave that through as an editorial change, or updat-with-implementation-specifics that needs no full committee vote. Any objections?

simonpj:
I have not really been following this, assuming it'd come to us when it settled down.

Can someone say why the double quotes round the category are needed?  Why
```
    {-# WARNING in "x-some-category" ["message1", "message2"] #-}
```
rather than
```
    {-# WARNING in x-some-category ["message1", "message2"] #-}
```
After all, we write `(f x)` not `("f" "x")`.  And the lexical syntax for categories looks rather like that for identifiers.  So why the double quotes?  Feels like extra clutter.

adamgundry:
> And the lexical syntax for categories looks rather like that for identifiers.

Similar, but not the same, though, because the lexical syntax of categories permits dashes, whereas `x-some-category` is lexed as three identifiers separated by the `-` operator. We could have the lexer enter a special state after it sees `WARNING in`, but it's much simpler to use a quoted string. And there is precedent for doing so (e.g. in the names of `RULES` and package names in `PackageImports`).

adamgundry:
FWIW the implementation of the proposal as modified by this PR is now ready for review at https://gitlab.haskell.org/ghc/ghc/-/merge_requests/9684

nomeata:
I’ve notified the committee that I would like to fast-track this.

simonpj:
> Similar, but not the same, though, because the lexical syntax of categories permits dashes, whereas x-some-category is lexed as three identifiers separated by the - operator. We could have the lexer enter a special state after it sees WARNING in, but it's much simpler to use a quoted string. And there is precedent for doing so (e.g. in the names of RULES and package names in PackageImports).

I'm not really convinced.  This feels like the implementation poking through too much.

The names of RULES is different: it really is a string including spaces.  you can have `RULES "Rule for map-fb (wombat %% fred)" ...`.  But here warning categories have a lexical syntax of their own.

If you want them to parse as identifiers, I suppose you could use underscores instead of hyphens:
```
    {-# WARNING in x_some_category ["message1", "message2"] #-}
```
But in fact there is structure inside: the "x" part is separate from the "some_category" part I think.  So maybe give it proper syntax:
```
    {-# WARNING in x:some_category ["message1", "message2"] #-}
    {-# WARNING in x:anotherCategory ["message1", "message2"] #-}
```
which lexes as "x", then ":", then "some_category".  etc.

int-index:
> But here warning categories have a lexical syntax of their own.

Their syntax is dictated by the command-line interface of GHC, since they will be used in flags such as -Wx-some-category` or `-Werror=x-another-category`.

So from the perspective of GHC's parser, it's an opaque string. Further validation is possible, but the point is: their syntax is not tied to Haskell syntax.

simonpj:
> I'm not really convinced. This feels like the implementation poking through too much.

Maybe I'm a lone voice.  What do others think?  It's not a huge deal either way I suppose.

For language extensions (which also appear on the command line) we don't say `{-# LANGUAGE "GADTs", "OverlappingInstances" #-}` and it would be clunky to do so.  Doesn't it seem inconsitent to require quotes for warning categories but not for language extensions?

adamgundry:
> But in fact there is structure inside: the "x" part is separate from the "some_category" part I think.

In my mind (and in the draft implementation) the `x-` prefix is part of the name, rather than a separate syntactic decoration. That allows a consistent treatment with the `deprecations` category, which doesn't have a prefix, and permits a possible future in which we have other prefixes or other recognised categories.

Moreover it allows us to present GHC as having, for example, one flag `-Werror=<category>` where `<category>` may be a built-in flag or a user-defined category. If the `x-` prefix was not part of the name, we would instead conceptually have two flags `-Werror=<built-in-category>` and `-Werror=x-<user-defined-category>`.


> Doesn't it seem inconsitent to require quotes for warning categories but not for language extensions?

We could say that we require quotes only when a warning category contains a dash (and hence cannot be lexed as a Haskell identifier). I believe none of the existing language extension names contain dashes. :smile: This would mean `{-# WARNING in deprecations ... #-}` would be accepted, but of course all other category names contain dashes.

nomeata:
Warning categories are usually passed as command line flags and the common style is with dashes (`-Wmissing-methods`). It would be strange if for custom categories we'd use a different convention like `-Wx_bad_asymptotics`. And even stranger to use underscores in the source but map them to dashes on the command line. (Not sure which one is proposed above.)

So if we don't want to deal with lexer complexity, I find the work-around with `in` and quotes quite reasonable.

konsumlamm:
> I'm surprised at the lack of feedback from everyone else in the community. I would love to see a broader range of opinions. To summarise, which do you prefer?

Fwiw, I quite like the currently suggested `{-# WARNING in "x-category" ... #-}` syntax. Normal warnings include dashes, so it is natural to be able to also define custom warnings with dashes.

PR #577, opened 2023-01-30
Amend the Higher Order Patterns proposal

noughtmare:
This amendment clears up one final point of discussion on the original proposal, discussed in [this thread](https://github.com/ghc-proposals/ghc-proposals/pull/555#discussion_r1081865219).

@simonpj @adamgundry

### DIFF
--- a/proposals/0555-template-patterns.rst
+++ b/proposals/0555-template-patterns.rst
@@ -156,16 +156,16 @@ The change proposed here makes matching more powerful by introducing the notion 
 	- ``x``, ``y``, ``z`` must be term variables (not type applications).
 
 * 	**Matching of higher order patterns (HOP-matching)**.
-        A higher order pattern ``f x y z`` (in the template) matches *any target expression* ``e`` provided:
+ 	A higher order pattern ``f x y z`` (in the template) matches *any target expression* ``e`` provided:
 
-	- The target has the same type as the template
-	- No local binder is free in ``e``, other than ``x``, ``y``, ``z``.
+	1. The target has the same type as the template.
+	2. No local binder is free in ``e``, other than ``x``, ``y``, ``z``.
+	3. ``e`` is not of the form ``(e' z)``, where ``z`` is not free in ``e'``.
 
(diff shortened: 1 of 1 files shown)


proposals/0555-template-patterns.rst
```diff
@@ -156,16 +156,16 @@ The change proposed here makes matching more powerful by introducing the notion
...
 
 * 	**Matching of higher order patterns (HOP-matching)**.
-        A higher order pattern ``f x y z`` (in the template) matches *any target expression* ``e`` provided:
+ 	A higher order pattern ``f x y z`` (in the template) matches *any target expression* ``e`` provided:
 
-	- The target has the same type as the template
-	- No local binder is free in ``e``, other than ``x``, ``y``, ``z``.
+	1. The target has the same type as the template.
+	2. No local binder is free in ``e``, other than ``x``, ``y``, ``z``.
+	3. ``e`` is not of the form ``(e' z)``, where ``z`` is not free in ``e'``.
```
treeowl, line 163:
```suggestion
	3. If ``e`` is of the form ``(e' z)``, then ``z`` is free in ``e'``.
```

simonpj: I can see why you want to invert it, but the intent is: decomposition doesn't apply.  That is, it is *not* the case that `e` looks like `e' x`, where `e'` doesn't mention x.  

It's a bit like saying `\x. e` is not eta-reducible.  It would be eta-reducible if `e` was of form `e' x` where `e'` does not mention `x`.  But the intent is always "the lambda isn't eta-reducible".

I'm inclined to leave it as it is.

sgraf812: Perhaps it is clearer to say
```suggestion
	3. ``\z.e`` cannot be eta-reduced. That is, ``e`` is not of the form ``(e' z)``, where ``z`` is not free in ``e'``.
```

sgraf812 (approved):
I think it would be reasonable to merge this amendment, with or without my suggestion.

adamgundry:
I'm just going to accept this as a minor editorial change, since higher-order patterns has been implemented for some time. Thanks for your work on this @noughtmare and @simonpj!

PR #579, opened 2023-02-10
Meta: Add Backward Compatibility section to template and review criteria

### LINKED ISSUE #578: Backwards compatibility
In the light of discussions about stability (e.g. https://discourse.haskell.org/t/language-library-and-compiler-stability-moved-from-ghc-9-6-migration-guide/5745), I think it would be a good idea if we made backwards compatibility a more explicit consideration in the proposals process.

I've tended to use the "Costs and Drawbacks" section of a proposal to discuss backwards compatibility concerns, but I realise now that the template doesn't direct authors to do that. Perhaps we should add a section to the proposal format for backwards compatibility? That would give somewhere for proposals to explain their impact on existing code and any steps to aid migration (e.g. introducing warnings before removing a feature).

I think we should also extend the review criteria to discuss backwards compatibility. @bgamari made [some good suggestions](https://discourse.haskell.org/t/language-library-and-compiler-stability-moved-from-ghc-9-6-migration-guide/5745/59):
>
> (3) I think points to an area where the GHC Proposals Process could be improved. The proposal recognizes that the change may break existing programs and says that the programs should adjust their whitespace to match the new whitespace rules. However, the proposal is silent on breakage extent, migration timeline, pre-breakage user warnings, or how tools might help aid migration. The first item is understandable as the proposal pre-dates head.hackage being an easily-used tool for characterising this sort of impact (although Ryan did [adapt](https://gitlab.haskell.org/ghc/head.hackage/-/merge_requests/62/diffs) head.hackage). However, today we could decide that the proposal process should do better in all of these areas:
>
> - we could require an approximate impact assessment for syntactic changes (e.g. how many occurrences of the affected token sequences does grep find on Hackage?)
> - alternatively, if grepping for affected syntax is too fiddly, the committee could instead require a second phase of approval such that, when an implementation is ready, a proper head.hackage study can be performed. In light of the results of this study the committee and proposal could design an appropriate migration strategy before final approval.
> - the committee could adopt a strategy similar to that of the CLC, requiring a warning period for any change that breaks user programs
> - with the continued progress of ghc-exactprint, I am hopeful that we are now in a better position to provide automated refactoring tools to facilitate mechanical migration for this sort of change. This particular change would be a great test-bed for this sort of tooling.

adamgundry:
Resolves #578.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -533,6 +533,13 @@ and any other relevant considerations, appropriately.
   absolutes, and regressions against the principles are possible, if a
   proposal is otherwise very strong.
 
+* *Backward compatibility.* Will the change break existing code, and if so, has
+  an adequate impact assessment been carried out to determine whether the
+  benefits outweigh the costs? Is there a clearly documented migration path?
+  Will users receive warnings in advance of the breaking change, and reasonable
+  error messages afterwards?
+
+
 How to build the proposals?
 ---------------------------
 
--- a/README.rst
+++ b/README.rst
@@ -84,7 +84,11 @@ This section outlines what stages a proposal may go through. The stage is identi
 
    Acceptance of the proposal implies that the implementation will be accepted
    into GHC provided it is well-engineered, well-documented, and does not
-   complicate the code-base too much.
+   complicate the code-base too much.  Moreover the GHC maintainers may decline
+   to accept (or revert) an implementation pending revisions to the proposal, if
+   there turn out to be significant gaps in the specification, unforeseen
+   interactions with existing features, or unexpected breaking changes not
+   covered by the backwards compatibility assessment.
 
    `≡ List of accepted proposals <https://github.com/ghc-proposals/ghc-proposals/tree/master/proposals>`_ •
    `≡ List of proposals being revised <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3A%22Needs+revision%22>`_ •
--- a/README.rst
+++ b/README.rst
@@ -83,12 +83,12 @@ This section outlines what stages a proposal may go through. The stage is identi
    author for review (label: `Needs revision <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3A%22Needs+revision%22>`_), or accepts it (label: `Accepted <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3A%22Accepted%22>`_).
 
    Acceptance of the proposal implies that the implementation will be accepted
-   into GHC provided it is well-engineered, well-documented, and does not
-   complicate the code-base too much.  Moreover the GHC maintainers may decline
-   to accept (or revert) an implementation pending revisions to the proposal, if
-   there turn out to be significant gaps in the specification, unforeseen
-   interactions with existing features, or unexpected breaking changes not
-   covered by the backwards compatibility assessment.
+   into GHC provided it is well-engineered, well-documented, conforms to the
+   specification and does not complicate the code-base too much.  However, the
+   GHC maintainers may reject an implementation if there turn out to be
+   significant gaps in the specification, unforeseen interactions with existing
+   features, or unexpected breaking changes not covered by the backwards
+   compatibility assessment.  In this case the proposal should be revised.
 
    `≡ List of accepted proposals <https://github.com/ghc-proposals/ghc-proposals/tree/master/proposals>`_ •
    `≡ List of proposals being revised <https://github.com/ghc-proposals/ghc-proposals/pulls?q=label%3A%22Needs+revision%22>`_ •
@@ -541,7 +541,9 @@ and any other relevant considerations, appropriately.
   an adequate impact assessment been carried out to determine whether the
   benefits outweigh the costs? Is there a clearly documented migration path?
   Will users receive warnings in advance of the breaking change, and reasonable
-  error messages afterwards?
+  error messages afterwards? See `the Backward Compatibility section of the
+  proposal template <proposals/0000-template.rst#backward-compatibility>`_
+  for specifics of how breakage is assessed.
 
 
 How to build the proposals?
--- a/proposals/0000-template.md
+++ b/proposals/0000-template.md
@@ -90,11 +90,10 @@ stop working? Assess the expected impact on existing code on the following scale
 
 0. No breakage
 1. Breakage only in extremely rare cases (e.g. for specifically-constructed
-   examples, but probably no Hackage packages)
+   examples, but probably no packages published in the Hackage package repository)
 2. Breakage in rare cases (e.g. a few Hackage packages may break, but probably
-   no Stackage packages)
-3. Breakage in uncommon cases (e.g. a few Stackage packages may break, but
-   probably no head.hackage packages)
+   no packages included in recent Stackage package sets)
+3. Breakage in uncommon cases (e.g. a few Stackage packages may break)
 4. Breakage in common cases
 
 (For the purposes of this assessment, GHC emitting new warnings is not
--- a/proposals/0000-template.md
+++ b/proposals/0000-template.md
@@ -86,13 +86,32 @@ drawbacks that cannot be resolved.
 ## Backward Compatibility
 
 Will your proposed change cause any existing programs to change behaviour or
-stop working? If so, assess the expected impact on existing code, and explain
-why the benefits of the change outweigh the costs of breakage.
+stop working? Assess the expected impact on existing code on the following scale:
 
+0. No breakage
+1. Breakage only in extremely rare cases (e.g. for specifically-constructed
+   examples, but probably no Hackage packages)
+2. Breakage in rare cases (e.g. a few Hackage packages may break, but probably
+   no Stackage packages)
+3. Breakage in uncommon cases (e.g. a few Stackage packages may break, but
+   probably no head.hackage packages)
+4. Breakage in common cases
+
+(For the purposes of this assessment, GHC emitting new warnings is not
+considered to be a breaking change, i.e. packages are assumed not to use
+`-Werror`.  Changing a warning into an error is considered a breaking change.)
+
+Explain why the benefits of the change outweigh the costs of breakage.
 Describe the migration path. Consider specifying a compatibility warning for one
 or more compiler releases before the change is fully implemented. Give examples
 of error messages that will be reported for previously-working code; do they
 make it easy for users to understand what needs to change and why?
+
+When the proposal is implemented, the implementers and/or GHC maintainers should
+test that the actual backwards compatibility impact of the implementation is no
+greater than the expected impact. If not, the proposal should be revised and the
+steering committee approve the change.
+
 
 
 ## Alternatives
--- a/proposals/0000-template.md
+++ b/proposals/0000-template.md
@@ -83,6 +83,18 @@ learnability of the language for novice users. Define and list any remaining
 drawbacks that cannot be resolved.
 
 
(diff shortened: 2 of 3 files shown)


nomeata:
You ought to update the copy of the template embedded in the link in the readme.

Do we want to differentiate between breaking code that uses an existing language standard or edition (Haskell2010, GHC3021) vs. code that explicitly opted into using an extension (which maybe comes with less guarantee of stability)? Not saying that the latter is necessarily less bad, but probably still a useful data point.

proposals/0000-template.rst
```diff
@@ -99,6 +99,18 @@ learnability of the language for novice users. Define and list any remaining
...
+Backward Compatibility
+----------------------
+Will your proposed change cause any existing programs to change behaviour or
+stop working? If so, assess the expected impact on existing code, and explain
+why the benefits of the change outweigh the costs of breakage.
+
+Describe the migration path. Consider specifying a compatibility warning for one
+or more compiler releases before the change is fully implemented. Give examples
+of error messages that will be reported for previously-working code; do they
+make it easy for users understand what needs to change and why?
```
konsumlamm, line 111:
```suggestion
make it easy for users to understand what needs to change and why?
```

adamgundry:
> Someone ought to update the copy of the template embedded in the link in the readme (before merging).

Hmm, I didn't realise such a copy existed. I found it in the source, but it seems that the links are not showing up in the GitHub web interface?

> Do we want to differentiate between breaking code that uses an existing language standard or edition (Haskell2010, GHC3021) vs. code that explicitly opted into using an extension (which maybe comes with less guarantee of stability)? Not saying that the latter is necessarily less bad, but probably still a useful data point.

This is an interesting question, and I think it belongs to our broader discussion of policies around extensions. While I assume we would be more reluctant to make a breaking change to a widely-used language edition than a recently-introduced and explicitly experimental extension, I would say that the scale of breakage is the important factor, not whether the extension is "officially blessed" by inclusion in a language edition.

nomeata:
>  I would say that the scale of breakage is the important factor, not whether the extension is "officially blessed" by inclusion in a language edition.

Ok, don't bite the bait then ;-)

Probably it will be clear anyway what language variant will be broken. And a backwards compatibility policy in light of extensions may soon emerge from the ongoing discussion on the role of extensions and language editions

angerman:
I'll take this bait
> explicitly experimental extension
which extensions are explicitly experimental today?

And while we are at this. Proposals are written often before an actual implementation exists, right? Thus we can hope that proposal authors (and reviewers) take utmost care to think about _anticipated breakage_. But do we validate _anticipated breakage_ against _actual breakage_? What do we do if actual breakage unexpectedly does not line up with the anticipated breakage?

adamgundry:
> > explicitly experimental extension
> > which extensions are explicitly experimental today?

I was speaking hypothetically, and there's not a very hard and fast classification. But for example `OverloadedRecordUpdate` is at the "very experimental and subject to change" end of the spectrum, as [mentioned in the manual](https://downloads.haskell.org/ghc/9.4.4/docs/users_guide/exts/overloaded_record_update.html). Perhaps we could more clearly communicate the stability of various extensions, but I don't think that will be easy.

> And while we are at this. Proposals are written often before an actual implementation exists, right? Thus we can hope that proposal authors (and reviewers) take utmost care to think about _anticipated breakage_. But do we validate _anticipated breakage_ against _actual breakage_? What do we do if actual breakage unexpectedly does not line up with the anticipated breakage?

At the moment this mostly relies in practice on review / head.hackage testing from the GHC developers, and of course ultimately just bug reports (against pre-releases if we are lucky, as we were with #573).

In general I feel that there is an unfortunate process gap after a proposal PR is accepted, because there is no guarantee that it will be implemented in a timely manner (or at all), and the GHC Steering Committee has little direct role in what happens subsequently. Thus (a) proposals can linger in an [accepted-but-not-implemented state](https://github.com/ghc-proposals/ghc-proposals/pulls?q=is%3Apr+label%3A%22Accepted%22+-label%3A%22Implemented%22) for a long time, as the context changes, and (b) sometimes when the proposal is implemented it is discovered that changes to the design are in order (e.g. #576). I'm not sure what to do to improve this though.

nomeata:
> I'm not sure what to do to improve this though.

I guess you joined after the discussion around https://github.com/ghc-proposals/ghc-proposals/pull/517 which I hoped would alleviate this a but

cdornan:
Clearly the right anser is a fabulously rich and generous benefactor who can fund everything at the rate that we can write proposals!

adamgundry:
A couple of points after talking this over with my colleagues at Well-Typed:
 * It would be helpful if the proposal explicitly indicated the expected level of breakage on a quantifiable scale. The committee could then incorporate this into its decision-making. Perhaps the scale could be something like:
   1. No breakage
   2. Potential breakage in obscure cases, but all of Stackage compiles with the change applied
   3. A small number of packages break, but all of head.hackage compiles with the change applied
   4. A significant number of packages break
 * This could then be validated by the implementer or GHC maintainers once an implementation was ready. Point 6 of "The life cycle of a proposal" should make explicit that an inaccurate backwards compatibility assessment is justification for GHC maintainers to decline to accept (or to revert) an implementation, pending revision of the proposal and agreement by the committee that the higher-than-expected breakage is acceptable.

simonpj:
Where are we on this?  Adam do you want to modify the wording in the light of the points you make?  (Which seem reasonable to me.)

adamgundry:
I've updated the text per discussion above. I'd particularly appreciate feedback on the rating scale used for assessing backwards incompatibilities; is it sufficiently well-defined to be useful?

proposals/0000-template.md
```diff
@@ -83,6 +83,37 @@ learnability of the language for novice users. Define and list any remaining
...
+## Backward Compatibility
+
+Will your proposed change cause any existing programs to change behaviour or
+stop working? Assess the expected impact on existing code on the following scale:
+
+0. No breakage
+1. Breakage only in extremely rare cases (e.g. for specifically-constructed
+   examples, but probably no Hackage packages)
+2. Breakage in rare cases (e.g. a few Hackage packages may break, but probably
+   no Stackage packages)
```
simonpj, line 95: It's not clear to me what the difference between Hackage and Stackage is.  Spell it out!

proposals/0000-template.md
```diff
@@ -83,6 +83,37 @@ learnability of the language for novice users. Define and list any remaining
 drawbacks that cannot be resolved.
 
 
+## Backward Compatibility
+
+Will your proposed change cause any existing programs to change behaviour or
+stop working? Assess the expected impact on existing code on the following scale:
```
simonpj, line 89: Can we point to the furtehr specifics info in the template.md file?  Or reproduce those specifics here if you prefer

README.rst
```diff
@@ -84,7 +84,11 @@ This section outlines what stages a proposal may go through. The stage is identi
 
    Acceptance of the proposal implies that the implementation will be accepted
    into GHC provided it is well-engineered, well-documented, and does not
-   complicate the code-base too much.
+   complicate the code-base too much.  Moreover the GHC maintainers may decline
+   to accept (or revert) an implementation pending revisions to the proposal, if
```
simonpj, line 88: This is a bit of a tortured sentence, hard to parse even for a native English speaker. Consider rewording.

simonpj (approved):
Good!  Some cosmetic suggestions.

proposals/0000-template.md
adamgundry: I've added a reference in the README.

README.rst
adamgundry: Indeed, I've reworded.

proposals/0000-template.md
adamgundry: I adjusted this a little to hopefully be clearer. I'm not sure it makes sense to give a full account of the relationship between Hackage and Stackage.

simonpj:
> I've updated the text per discussion above. I'd particularly appreciate feedback on the rating scale used for assessing backwards incompatibilities; is it sufficiently well-defined to be useful?

To me a 5-point scale seems a bit much.  4 would do.  But not worth too much polishing.

adamgundry:
@nomeata this one is ready for the committee to consider, I think (though I'm not sure what the right process is for such meta-proposals).

nomeata (changes requested):
Reminder to myself to update the copy of the template embedded in the link in the readme before merging

nomeata:
I was about to assign this to Simon for shepherding, but I think this doesn’t need a heavy process – it is just an invitation for the authors to write more, hardly controversial.

If @simonpj (as chari) is ok with it, I’ll just merge without committee process.

simonpj:
> To me a 5-point scale seems a bit much. 4 would do.

Adam didn't reply to this. @adamgundry ?

But in general yes, let's just merge once Adam says "done".  It's not controversial.  Other commmittee members, yell if you disagree.

goldfirere:
I support this change, but I actually think we should raise this on the committee list instead of just merging. It changes our process slightly and increases the burden on proposers. Again, I'm all for it -- but, for example, I would find it odd if the inverse of this patch were merged without process. So I think a quick airing on the list would be nice.

adamgundry:
> To me a 5-point scale seems a bit much. 4 would do.
> 
> Adam didn't reply to this. @adamgundry ?

I don't have a very strong feeling here about the exact number of entries. I guess we could indeed merge "rare" and "uncommon".

proposals/0000-template.md
```diff
@@ -83,6 +83,36 @@ learnability of the language for novice users. Define and list any remaining
...
+
+0. No breakage
```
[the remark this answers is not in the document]
adamgundry: Good questions. My answers would be:
 * If a proposal adds an extension, and turning on that extension breaks code, that isn't a backwards compatibility problem as such. Existing code will still compile, because it won't have the extension enabled. It might be a design consideration that influences the desirability of the extension, of course.
 * Changes to existing extensions are considered breaking, regardless of whether the extension has been advertised as "unstable", though we might be more willing to accept breaking changes to such features. We don't currently have a clear demarcation between "stable" and "unstable" extensions, and this is to some extent a matter of community consensus.

My intention is that whether something is a "breaking change" should be objective and correspond closely to whether code that compiled with GHC version X still compiles with version X+1. But acceptance of proposals is about weighing various factors, of which potential breakage should be only one.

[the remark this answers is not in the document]
nomeata: We probably want to know about both kinds of breakage: even if it's guarded by an extension the breakage becomes relevant when we consider turning the extension on by default, down the road, and so it may be relevant how much breakage to expect then (unless the extension is meant to stay a flag forever)

angerman:
I've had an interesting discussion around recent breakage (Bifunctor requiring a Functor instance now). Having warnings, about something _becoming_ a problem could also lead to more visibility of this change (outside of the selected circle of people with a strong interest in following proposals and compiler evolution, e.g. those who only apply the compiler, or are made to use the compiler). This _could_ lead to feedback from people who had not interacted with, or seen any of the changes to the compiler. This could be considered a public preview of a (breaking) change.

Are we willing to reconsider if there is significant pushback?

In any case, @adamgundry for this proposal, I think this is a very good step forward!

proposals/0000-template.md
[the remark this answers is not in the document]
adamgundry: Fair point. I can try to change the proposal text to reflect this.

angerman:
Thus specific instance is history. Let's look forward 😄

PR #65, opened 2017-07-29
Require namespacing fixity declarations for type names

RyanGlScott:
This proposes a way to clean up GHC's implementation of fixity declarations for names that live at both the value and type level, as the current implementation is rather unsatisfying (and sometimes broken).

[Rendered](https://github.com/RyanGlScott/ghc-proposals/blob/type-infix/0000-type-infix.rst)

PR #581, opened 2023-02-23
Namespace-specified imports

adamgundry:
The [proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0581-namespace-specified-imports.rst) has been accepted; the following discussion is mostly of historic interest.

This proposal allows imports to be limited to either the type or data namespace. Moreover it clarifies the specification of
`ExplicitNamespaces` and related extensions. This is heavily derived from #270. Thanks to @Hithroc and @cdornan for collaborating on this! This also includes changes originally prepared by @int-index amending proposal #65.

[Rendered](https://github.com/adamgundry/ghc-proposals/blob/namespace-specified-imports/proposals/0000-namespace-specified-imports.rst).
[Work-in-progress user's guide section with these changes](https://github.com/cdornan/ExplicitNamespaces-doc/pull/1).

### DIFF
--- a/proposals/0000-namespace-specified-imports.rst
+++ b/proposals/0000-namespace-specified-imports.rst
@@ -11,7 +11,7 @@ Namespace-specified imports
 .. contents::
 
 This proposal supports users who wish to write code without relying on punning,
-by allowing imports to be limited to either the type or data namespace.
+by allowing imports of all names in either the type or data namespace.
 Moreover it clarifies the specification of ``ExplicitNamespaces`` and related
 extensions.
 
@@ -165,14 +165,14 @@ Solution Overview
 ~~~~~~~~~~~~~~~~~
 
 To help programmers deal with the external code that uses punning we propose to
-introduce namespace-specified import syntax, guarded behind the
-``ExplicitNamespaces`` extension. The syntax introduces two keywords as part of
-imports, ``data`` and ``type``:
+extend the ``ExplicitNamespaces`` extension to allow the ``data`` and ``type``
+keywords to be used as part of import or export lists, potentially with a ``..``
+wildcard. For example:
 
 .. code:: haskell
 
-   import qualified Data.Proxy type as T   -- import only the type namespace
-   import qualified Data.Proxy data as D   -- import only the data namespace
+   import qualified Data.Proxy as T (type ..)   -- import only the type namespace
+   import qualified Data.Proxy as D (data ..)   -- import only the data namespace
 
 This avoids needing to name each item individually, but otherwise has the same
 effect as writing out an explicit import list, like this:
@@ -182,25 +182,24 @@ effect as writing out an explicit import list, like this:
    import qualified Data.Proxy as T (type Proxy)   -- import only the Proxy type
    import qualified Data.Proxy as D (data Proxy)   -- import only the Proxy constructor
 
-For consistency, this proposal introduces ``data`` as a namespace specifier
-within an import list, guarded by ``ExplicitNamespaces``.  (This replaces the
-existing use of ``pattern`` in import lists, guarded by ``PatternSynonyms``.)
-Moreover, this proposal modifies `proposal #65
+The ``data`` namespace specifier replaces the
+existing limited use of ``pattern`` in import lists, guarded by ``PatternSynonyms``.
+Moreover, for consistency this proposal modifies `proposal #65
 <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0065-type-infix.rst>`_
 to use ``data`` rather than ``value`` as the keyword for the data namespace in
 fixity declarations and pragmas.
 
 This proposal does not directly make changes to the tick syntax, or provide an
 equivalent at use sites. However it should reduce the need for disambiguating
-promoted data constructors using ticks, because namespace-specified qualified
+promoted data constructors using ticks, because qualified
 imports can be used instead.
 
 
 Proposed Change Specification
 -----------------------------
 
-Allow ``data`` namespace specifier in import/export lists
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Allow namespace specifiers with wildcards in import/export lists
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 When ``ExplicitNamespaces`` is enabled, anywhere the ``type`` keyword may appear
 in an import or export list, the ``data`` keyword may also appear.  Call such an
@@ -208,6 +207,10 @@ occurrence a *namespace specifier*. Any import/export of an identifier with a
 namespace specifier will be taken to refer only identifiers in the given
 namespace.  It is an error to use a namespace specifier if the identifier is not
 in scope in the given namespace.
+
+Moreover, a namespace specifier may be followed by a ``..`` wildcard instead of
+a single name. This is equivalent to importing or exporting all the available
+names in the corresponding namespace.
 
 More precisely, the existing grammar of import/export items accepted by GHC is
 essentially the following (after some minor simplifications): ::
@@ -223,12 +226,15 @@ essentially the following (after some minor simplifications): ::
                  |  qtycon
                  | 'type' oqtycon  -- with ExplicitNamespaces
 
-This proposal extends ``qcname_ext`` as follows: ::
+This proposal redefines ``qcname_ext`` as follows: ::
 
       qcname_ext -> qvar
                  |  qtycon
-                 | 'type' oqtycon  -- with ExplicitNamespaces
-                 | 'data' qvarcon  -- with ExplicitNamespaces
+                 | 'type' oqtycon_w_wildcard  -- with ExplicitNamespaces
+                 | 'data' qvarcon_w_wildcard  -- with ExplicitNamespaces
+
+      oqtycon_w_wildcard -> oqtycon | '..'
+      qvarcon_w_wildcard -> qvarcon | '..'
 
 Notice that:
 
@@ -236,8 +242,8 @@ Notice that:
   whereas ``type`` and ``data`` are valid either at the top or nested inside a
   type constructor or typeclass name.
 
-* ``data`` may be followed by a data constructor name or a variable name (with
-  the latter including record selectors, in particular).
+* ``data`` may be followed by a data constructor name, a variable name
+  (including record selectors, in particular), or a ```..`` wildcard.
 
 * Where a parent type constructor or class is exported together with its
   children, any namespace specifier on an individual import/export item will
@@ -245,10 +251,33 @@ Notice that:
   ``import M (type T(..))`` imports both ``T`` in the type namespace and any
   children in either namespace.
 
-* ``import M (data D(..))`` is syntactically valid, but not useful, as it is not
-  currently possible for identifiers in the data namespace to have children.
-  (We might imagine changing this e.g. for pattern synonym record fields, but
-  doing so is outside the scope of this proposal.)
+
+Dodgy import/export warnings
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The existing ``-Wdodgy-imports`` and ``-Wdodgy-exports`` flags (part of the
+``-W`` warning group) emit warnings if a ``..`` wildcard does not refer to any
+names in scope.  For example this arises in an export of ``T(..)`` if ``T`` is a
+type synonym or a data type with no constructors.
+
+Similarly, under this proposal a warning will be emitted if a ``type ..`` or
+``data ..`` item does not refer to any names in the corresponding namespace.
+
+For example:
+
+* ``import M (data D(..))``, ``import M (data D(type ..))`` and ``import M (data
+  D(data ..))`` are syntactically valid, but will always give rise to a warning,
+  as it is not currently possible for identifiers in the data namespace to have
+  children.  (We might imagine changing this, e.g. for record fields, but doing
+  so is outside the scope of this proposal.)
+
+* ``import M (type T (data ..))`` is accepted. When ``T`` is a type its only
+  sub-items are in the data namespace so this is somewhat redundant, but a class
+  may have both children in either namespace.
+
+* Similarly, ``import M (type T (type ..))`` is accepted but will give rise of a
+  warning when ``T`` is a type or a class that does not have any associated
+  types.
 
 
 Deprecate use of ``pattern`` in import/export lists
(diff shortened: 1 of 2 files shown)


proposals/0000-namespace-specified-imports.rst
```diff
@@ -0,0 +1,554 @@
...
+  <https://github.com/ghc-proposals/ghc-proposals/pull/214>`_ suggested using
+  prefix ``data.`` and ``type.`` systematically for this purpose.  However, that
+  would be syntactically noisy and there was difficulty gathering consensus for
+  that approach.
+
+* Rather than modifying ``ExplicitNamespaces`` to allow namespace-specified
+  imports, we could introduce a new extension ``NamespacedImports``.  This would
+  make it clearer whether code depends on the new feature, or whether the older
+  version of ``ExplicitNamespaces`` was enough.  However the general consensus
+  seems to favour reducing the number of extensions over avoiding change to
```
nomeata, line 514: That's debatable :-). We'll know more once we have concluded the ongoing policy discussion. But since we didn't add ExplicitNamespaces to GHC2023, likely fine here

proposals/0000-namespace-specified-imports.rst
```diff
@@ -0,0 +1,554 @@
...
+     , D(data f)    -- Accepted: exports data family D and field f
+     , pattern D    -- Accepted: exports data constructor D but emits warning
+     , T(data D)    -- Accepted: exports type T and data constructor D
+     , data f       -- Accepted: exports field f
+     , data v       -- Accepted: exports term v
+     , type T (..)  -- Accepted: exports type T and all its constructors
+     , T(pattern D) -- Rejected: pattern keyword cannot be used in sub-list
+     , data T       -- Rejected: T not in scope in data namespace
+     , type E       -- Rejected: E not in scope in type namespace
+     ) where
```
int-index, line 330: Can we also have a `TypeData` example here? I assume that if I have `type data K = T1`, then I must export/import it as `K(type T1)` rather than `K(data T1)`.

adamgundry: Good idea! Yes my reading of #106 is that the names are introduced in the type namespace, so would use `type` rather than `data`.

simonpj:
Looks good to me. One nit:
```
import qualified Data.Proxy as D (Proxy) -- Accepted (imports data constructor qualified)
```
Is that right?  It would import the *type* constructor `Proxy` wouldn't it?

Actually for all these Proxy examples it would help to remind us of the declaration of `Proxy` (which uses punning).

I think that
```
  import M type( T(..) )
```
where `T` is an algebraic data constructor, would import just `T` and not the data constructors.  (Possibly with a warning about a misleading `(..)`.

I assume that
```
  import M type ( T( K1, K2 ) )
```
would be an error because it explicitly imports K1, K2, which are in the data space?

Giving these as exmples would be useful.  And checking that the spec covers it.

proposals/0000-namespace-specified-imports.rst
adamgundry: I added `TypeData` to Effects and Interactions (https://github.com/adamgundry/ghc-proposals/blob/namespace-specified-imports/proposals/0000-namespace-specified-imports.rst#41typedata).

jvanbruegge:
I find it very weird that
```haskell
import Foo (type T(..))
```
also importa the data constructors. I would expect that it only exports the promoted constructors if using `DataKinds` and warn otherwise. Same for exports. Kind of like TypeData but for usage side, not definition

adamgundry:
@jvanbruegge that's already the case, though, and changing it sounds like it could lead to rather painful breakage. I think it makes sense if we understand the namespace specifier as binding tightly, i.e. `import Foo ( (type T) (..) )` rather than `import Foo ( type (T(..)) )`.

jvanbruegge:
That's very unfortunate. However a quick look at hackage shows only 5 packages that might be affected: https://hackage-search.serokell.io/?q=import.*%5C%28type+%5B%5E+%5D%2B%5C%28.%2B%5C%29%5C%29

So this might be something for consideration

proposals/0000-namespace-specified-imports.rst
```diff
@@ -0,0 +1,656 @@
...
+
+* Where a parent type constructor or class is exported together with its
+  children, any namespace specifier on an individual import/export item will
+  apply only to the parent; the children are unrestricted.  For example,
+  ``import M (type T(..))`` imports both ``T`` in the type namespace and any
+  children in either namespace.
+
+* ``import M (data D(..))`` is syntactically valid, but not useful, as it is not
+  currently possible for identifiers in the data namespace to have children.
+  (We might imagine changing this e.g. for pattern synonym record fields, but
```
goldfirere, line 250: Or regular data constructor record fields...

goldfirere:
Thanks for writing this up! I'm in support.

adamgundry:
Discussion on this one seems to have quietened down, so I'd like to submit it to the committee, @nomeata.

angerman:
As I've now been assigned Shepard for this, I've given this a few reads and here are my thoughts so far. The overall idea is great, I'm just having trouble with the syntax. Unless I'm mistaking this, permits to write the following to import _only_ the types (type namespace) from `Data.Proxy`
```
import Data.Proxy type
```
However, I find the postfix notation very hard to read; it feels to me like a half finished sentence. And this The intention seems to be to import _all_ type**s** from that module. 
```
import types Data.Proxy
```
to me appears much more obvious what it's going to do. This of course opens the issue of what does
```
import data Data.Proxy
```
mean? And here `values` seems a more appropriate wording, which we appear to want to get away from. From an aesthetics point, I like that type and data are same length, yet type and value seem more readable to me? Maybe I'm the only one.

nomeata:
I think the position in the proposal is motivated by `import Foo (explicit, list, of, items)`. But I should let the author answer.

angerman:
I don't want to bikeshed this too much. I only want to highlight my usability concerns. My primary focus here is on trying to make sure people _guess_ the right meaning most of the time, even if they are unfamiliar with much of haskell so far.

The example syntax given prominently in the proposal is:
```
import qualified Data.Proxy type as T   -- import only the type namespace
import qualified Data.Proxy data as D   -- import only the data namespace
```
and
```
import qualified Data.Proxy as T (type Proxy)   -- import only the Proxy type
import qualified Data.Proxy as D (data Proxy)   -- import only the Proxy constructor
```

with the change to the grammar being:
```
importdecl -> 'import' [src] ['safe'] ['qualified'] [package] modid [namespace] ['qualified'] ['as' modid] [impspec]

namespace  -> 'data'
           |  'type'
```

which again, would permit writing
```
import X type
```
(which to me is not clear what that's supposed to mean).  After reading the proposal, I know this is supposed to mean "import the type namespace (that is all types defined in module X) from module X", but the connection between the statement, and the meaning isn't very close IMO.

```
import X (type Proxy)
```
on the other hand reads ok to me.

As `ImportQualifiedPost` was brought up, this now ends up reading like
```
import X type qualified as Y
```
what is _type qualified_? And still _type_ is singular, yet I'm potentially getting many types imported.

Regarding the use in explicit import lists
```
import Data.Proxy type (Proxy(..))  -- Accepted: imports type constructor but not data constructor
import Data.Proxy (type Proxy(..))  -- Accepted: imports both type constructor and data constructor
```
seems like very fiddly :-/ I'm just moving the paren a few characters, and all of a sudden I'm additionally importing a data constructor, or not. I'm not sure how easy this is to spot during code reviews.

proposals/0000-namespace-specified-imports.rst
```diff
@@ -0,0 +1,656 @@
...
+
+If an import declaration uses a namespace specifier but no explicit import list,
+it is not an error for the declaration to bring no names into scope,
+e.g. because the ``data`` specifier was used on a module that exports only type
+names. (GHC may of course warn that such an import is redundant.)
+
+Where an import has a namespace specifier, any occurrence of an ellipsis
+``(..)`` will be taken to refer only to identifiers in that namespace.  For
+example, ``import M type (T(..))`` will import only type-level names (so the
+ellipsis will refer to nothing if ``T`` is a normal algebraic datatype).
```
eyeinsky, line 308: The ellipsis will refer to nothing, but the `T` type will still be imported, right? (Maybe say: "so for a normal algebraic data type `T` only the type `T` is imported and ellipsis will refer to nothing")

int-index:
@angerman What about this syntax:

```
import Data.Proxy (type ..) as T
import Data.Proxy (data ..) as D
```

The idea is to extend `qcname_ext` with two new forms:

```
  | 'type' '..'
  | 'data' '..'
```

which stand for "all types" and "all data".

michaelpj:
If we have `import Data.Proxy (type ...) as T` I would expect to be able to write `import Data.Proxy (...) as T` also! And what about `import Data.Proxy as T hiding (data ...)`?

int-index:
@michaelpj I don't know if you wrote `...` instead of `..` on purpose, but we must stick to the latter because

1. `..` is reserved syntax, `...` is a user-defined operator
2. `..` is already used in export/import items, e.g. `import Prelude(Bool(..))`

As to your actual suggestions, they look reasonable to me (i.e. they'd make the proposed syntax more regular), but I can't come up with an actual use case for them, since there are simpler ways of writing both of them.
1. `import Data.Proxy (..) as T` is equivalent to `import Data.Proxy as T`, i.e. just remove the `(..)` altogether
2. `import Data.Proxy as T hiding (data ..)` is equivalent to `import Data.Proxy (type ..) as T`, i.e. use the opposite namespace specifier

My preference would be to omit those unnecessary additions to keep the proposal simple, but ultimately it's up to @adamgundry and I don't feel strongly either way.

angerman:
@int-index that doesn't look too bad, and would be mostly in line with what we already have, reusing `..` as meaning wildcard.

```
import Data.Proxy as T (type ..) 
import qualified Data.Proxy as T (type ..)
import Data.Proxy qualified as T (type ..)
```

What about these two now:
```
import Data.Proxy type (Proxy(..))  -- Accepted: imports type constructor but not data constructor
import Data.Proxy (type Proxy(..))  -- Accepted: imports both type constructor and data constructor
```
so we now have
```
import Data.Proxy (     Proxy    )
import Data.Proxy (     Proxy(..))
import Data.Proxy (type ..       )
import Data.Proxy (data ..       )
import Data.Proxy (type Proxy    )  -- Accepted: imports type constructor but not data constructor
import Data.Proxy (type Proxy(..))  -- Accepted: imports both type constructor and data constructor
```
only?

Sorry if I'm getting a bit confused with your proposed syntax, the import one from the proposal.

int-index:
@angerman Here's the breakdown.

Today we have:

```
import Data.Proxy   -- import everything

import Data.Proxy (Proxy)          -- 1. import the type constructor only
import Data.Proxy (type Proxy)     -- 2. import the type constructor only
import Data.Proxy (pattern Proxy)  -- 3. import the data constructor only

import Data.Proxy (Proxy(Proxy))       -- 4. import both the type and data constructor
import Data.Proxy (Proxy(..))          -- 5. import both the type and data constructor
import Data.Proxy (type Proxy(Proxy))  -- 6. import both the type and data constructor
import Data.Proxy (type Proxy(..))     -- 7. import both the type and data constructor
```

My proposal is:
1. Replace (3) with `import Data.Proxy (data Proxy)`, i.e. use the `data` keyword instead of `pattern`
2. Allow these two new forms:
    ```
    import Data.Proxy (type ..)    -- 8. import all type constructors
    import Data.Proxy (data ..)    -- 9. import all data constructors and value bindings
    ```

All 9 forms also interact with `qualified` and `as Alias`, but those interactions aren't new and really are orthogonal to the proposed change.

int-index:
OK, so to list all the interactions, we end up with these combinations:

```
import qualified X (type ..) as Y
import qualified X (data ..) as Y
import X (type ..) qualified as Y
import X (data ..) qualified as Y
import X (type ..) as Y
import X (data ..) as Y
import qualified X (type ..)
import qualified X (data ..)
import X (type ..) qualified
import X (data ..) qualified
import X (type ..)
import X (data ..)
```

With @michaelpj's first suggestion, we also get

```
import qualified X (..) as Y
import X (..) qualified as Y
import X (..) as Y
import qualified X (..)
import X (..) qualified
import X (..)
```

And with @michaelpj's second suggestion, we also get

```
import qualified X as Y hiding (type ..)
import qualified X as Y hiding (data ..)
import X qualified as Y hiding (type ..)
import X qualified as Y hiding (data ..)
import X as Y hiding (type ..)
import X as Y hiding (data ..)
import qualified X hiding (type ..)
import qualified X hiding (data ..)
import X qualified hiding (type ..)
import X qualified hiding (data ..)
import X hiding (type ..)
import X hiding (data ..)
```

angerman:
@adamgundry @Hithroc @cdornan as the proposal authors, what's your take on this?

adamgundry:
I like the idea, thanks for the suggestion @int-index! In particular it gets rid of the subtle distinction between `import Data.Proxy (type Proxy(..))` and `import Data.Proxy type (Proxy(..))`, and indeed various corner cases like `import Data.Proxy data (data Proxy)` which are just redundant. Moreover `import Data.Proxy (type ..)` should hopefully be reasonably easy to understand for users who haven't seen it before.

I would be inclined not to permit `import M (..)` because it is entirely redundant. The implementation could always suggest `import M` in the error message. On the other hand, `import M hiding (type ..)` seems relatively natural and might be useful in some circumstances, so I'd suggest allowing it.

This change does blur the notion of an explicit import list or explicit export list. Presumably `-Wmissing-import-lists` would not fire for `import M (type .., data ..)` even though it is essentially equivalent to `import M` (and similarly for `-Wmissing-export-lists`). I think this is okay though.

I'm happy to rewrite the proposal to use this syntax, if we have consensus that this is the right approach. How do others feel about it?

Hithroc:
I'm okay with the proposed syntax.

adamgundry:
I've now revised the proposal to take account of the new syntax, and I'm very happy with the new design. Thanks @int-index for the suggestion! A few things occurred to me while doing so:

* It is now natural to permit `type ..` or `data ..` to appear as a child import item, for example:
   ```hs
   import M (C (type ..))
   ```
   This might be useful if `C` is a class with many children in both namespaces (e.g. both associated types and methods). Thus I'm inclined to allow it.

* We can now export all names in a single namespace:  
   ```hs
   module M (type ..) where ...
    ```

* We can now somewhat redundantly import both namespaces:
   ```hs
   import M (type.., data ..)
   ```

phadej:
> Presumably `-Wmissing-import-lists` would not fire for `import M (type .., data ..)`

I'd suggest keep it firing for these. The very important point of `-Wmissing-import-lists` is to find imports which may cause problems if dependencies add new bindings. The way to silence `-Wmissing-import-lists` should not be to change `import M` into `import M(data ..)`, but to spell the imports (or import qualified).

PR #583, opened 2023-03-25
HasField redesign

adamgundry:
This proposal reviews and makes recommendations for alterations to the `HasField` typeclass (and to the planned changes to it in #158).  In particular, it proposes implementing updates using a separate `SetField` class, and adds support for unlifted datatypes and fields.

In order to keep this proposal simple, it does not yet propose adding support for type-changing update, which is left for a future proposal.

[Rendered](https://github.com/adamgundry/ghc-proposals/blob/hasfield-redesign/proposals/0000-hasfield-redesign.rst).

cc @ndmitchell @shayne-fletcher @sheaf

### DIFF
--- a/proposals/0000-hasfield-redesign.rst
+++ b/proposals/0000-hasfield-redesign.rst
@@ -278,6 +278,13 @@ automatically. Specifically, this occurs when ``r`` is a concrete record type,
 ``x`` is a ``Symbol`` naming one of the fields of the record, the field is in
 scope and is not existentially quantified or higher-rank.
 
+When a constraint is solved automatically, GHC will generate a dictionary with
+an implementation of ``modifyField``, as if an instance for ``SetField``
+existed. It will not actually generate instances of ``SetField``, however,
+because instances have global scope whereas ``SetField`` constraints are solved
+automatically only if the field is in scope.  (This is identical to the
+behaviour for ``HasField``.)
+
 If ``R x y`` is a record type with a field ``f :: T x`` belonging to
 constructors ``MkR1`` and ``MkR2`` but not ``MkR3``, the generated dictionary
 for ``SetField "f" (R x) a`` will be equivalent to: ::
@@ -357,8 +364,9 @@ which point the representation of the argument is necessarily fixed).  See
 Examples
 --------
 
-For each example datatype, we give the automatically-generated instances that
-GHC makes available when the first field is in scope.
+For the first field of each example datatype, we describe the behaviour of the
+constraint solver by giving the corresponding instances (though GHC does not
+actually generate these instances).
 
 
 Simple record
--- a/proposals/0000-hasfield-redesign.rst
+++ b/proposals/0000-hasfield-redesign.rst
@@ -464,6 +464,47 @@ Splitting ``HasField`` into separate ``HasField`` and ``SetField`` classes means
 it is possible to define get-only or set-only virtual fields (although set-only
 fields must still have the ability to define ``modifyField``).
 
+Unlike the automatic constraint solving, which takes account of whether the
+field name is in scope, normal ``instance`` declarations are globally scoped and
+cannot be hidden at module boundaries.  This means that once a virtual field is
+defined, its existence cannot be hidden from client code, which may be
+undesirable as it may expose internal implementation details.
+
+Virtual fields are sometimes useful for backwards compatibility after a field
+has been refactored, since pattern synonym fields do not lead to automatic
+constraint solving for ``HasField``.
+
+It is sometimes useful to define virtual ``HasField`` instances that are
+polymorphic in the field name, to give a specific datatype a convenient syntax
+using ``OverloadedRecordDot``. For example, this is used by
+`esqueleto <https://hackage.haskell.org/package/esqueleto-3.5.10.0/docs/src/Database.Esqueleto.Internal.Internal.html#line-2276>`_.
+
+Various more general virtual field ``HasField`` instances have been proposed,
+some of which (to be non-orphan) would need to live in ``GHC.Records``, such as:
+
+* `Instances for tuples with numeric field names
+  <https://github.com/haskell/core-libraries-committee/issues/143>`_
+  (currently available in the
+  `tuple-fields package <https://hackage.haskell.org/package/tuple-fields>`_).
+
+* An `instance for Maybe <https://github.com/haskell/core-libraries-committee/issues/191>`_
+  or `for a general Functor <https://github.com/ghc-proposals/ghc-proposals/issues/600>`_.
+
+* `Unit datatypes with virtual fields based on MonadReader or MonadWriter
+  <https://github.com/ghc-proposals/ghc-proposals/pull/583#issuecomment-1646789620>`_.
+
+While these are undoubtedly convenient in some cases, some of them may lead to
+code that cannot be easily understood in terms of field selection and update,
+and (having been designed for ``RecordDotSyntax``) they may or may not interact
+well with uses of ``HasField``/``SetField`` in optics libraries. Thus we do not
+propose to add such instances to ``GHC.Records`` for now, pending further
+experimentation.  In some cases it may be more appropriate to define new
+operators, rather than overloading ``.`` with yet more potential
+interpretations.  The intent of ``HasField``/``SetField`` is to allow type
+information to help resolve otherwise ambiguous field names from Haskell
+records, not to be a general abstraction over all possible notions of record or
+uses of dot syntax.
+
 
 Costs and Drawbacks
 -------------------
--- a/proposals/0000-hasfield-redesign.rst
+++ b/proposals/0000-hasfield-redesign.rst
@@ -801,10 +801,13 @@ constraint solver being invoked.
 
 On the other hand, it would be possible in principle to solve constraints such
 as ``HasField "identity" Rank1 (a -> a)`` for arbitrary ``a``, making it appear
-as if the field has an infinite family of types.  However, this does not extend
+as if the field has an infinite family of types.  However, this would not extend
 to ``SetField``, because there we really need the value being set to be
-polymorphic.  Moreover, it violates the functional dependency ``x r -> a`` on
-the ``HasField`` class.
+polymorphic.  Moreover, it would violate the functional dependency ``x r -> a``
+on the ``HasField`` class, leading to a violation of confluence: given wanteds
+``HasField "identity" r (α -> α)`` and ``HasField "identity" r (β -> β)``,
+applying the fundep forces ``α ~ β``; whereas if we were first to learn ``r ~
+Rank1`` then we could solve both constraints without requiring ``α ~ β``.
 
 Accordingly, we propose that ``HasField`` or ``SetField`` constraints
 involving fields with higher-rank types should not be solved automatically.
--- a/proposals/0000-hasfield-redesign.rst
+++ b/proposals/0000-hasfield-redesign.rst
@@ -673,6 +673,11 @@ An alternative would be to rename ``HasField`` (e.g. to ``GetField``), at the
 cost of breaking any code with an explicit import like ``HasField(getField)``,
 or that defines a virtual field instance of ``HasField``.
 
+While we could use a type synonym ``type HasField = GetField`` for partial
+backwards compatibility, this would not allow defining instances, and would mean
+that a ``HasField(..)`` import could no longer import ``getField``.
+
+
 
 Downsides of keeping the classes independent
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/proposals/0000-hasfield-redesign.rst
+++ b/proposals/0000-hasfield-redesign.rst
@@ -848,8 +848,8 @@ corresponding feature for ``SetField``.  This relies on the fact that
 appropriate warning for selection and update.
 
 
-Updates to partial fields
-^^^^^^^^^^^^^^^^^^^^^^^^^
+Updates could ignore partial fields
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 In principle, it is not necessary for ``setField`` or ``modifyField`` to emit a
 runtime error if used with a field that is not present in the datatype; they
 could silently return the value unchanged instead.  This behaviour may be more
@@ -861,6 +861,20 @@ datatype definition.  Somewhat related is `proposal #535
 <https://github.com/ghc-proposals/ghc-proposals/pull/535>`_, which suggests an
 extension ``MaybeFieldSelectors`` to control whether partial fields can lead to
 runtime exceptions.
+
+
+Refrain from solving partial fields?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Another option would be for GHC to refrain from solving ``HasField`` or
+``SetField`` constraints automatically where the fields involved are partial.
+This would allow users to define virtual fields with the behaviour they want,
+without conflicting with the automatic solutions.  See `this comment from
+@pnotequalnp <https://github.com/ghc-proposals/ghc-proposals/pull/583#issuecomment-1489278894>`_
+for more motivation for this idea.
+
+However, this would make ``getField`` and ``setField`` less consistent with
+traditional record selectors and record updates.  Moreover it would lead to
+backwards incompatibility for ``HasField``.
 
 
 Affine traversals
--- a/proposals/0000-hasfield-redesign.rst
+++ b/proposals/0000-hasfield-redesign.rst
@@ -59,7 +59,7 @@ The accepted `proposal #158
 <https://github.com/ghc-proposals/ghc-proposals/pull/158>`_ plans to change the
 definition of ``HasField`` to support updates, which is necessary for the full
 implementation of the ``OverloadedRecordUpdate`` extension.
-An implementation of this proposal is available as `GHC merge request !3257
+An implementation of proposal #158 is available as `GHC merge request !3257
 <https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3257>`_, but has not yet
 been merged, because the compile-time performance cost of the selected
 implementation strategy is unacceptably high.  Such costs were not really
@@ -751,7 +751,7 @@ the specific laws are:
 Where the constraint solver automatically solves one of these constraints, the
 laws will be satisfied.
 
-Where a field is partial, ``getField`` is necessarily not defined.  In this case
+Where a field is absent, that is where ``getField`` is undefined,
 the laws permit ``modifyField`` to be defined (to be a no-op) or undefined.
 However it may not change the constructor so that the field is present.
 
--- a/proposals/0000-hasfield-redesign.rst
+++ b/proposals/0000-hasfield-redesign.rst
@@ -1,0 +1,1102 @@
+HasField redesign
+=================
+
+.. author:: Adam Gundry
+.. date-accepted::
+.. ticket-url::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/583>`_.
+.. sectnum::
+.. contents::
+
+
+This proposal recommends improvements to the design of the ``HasField``
+typeclass (differing from those planned under `#158
+<https://github.com/ghc-proposals/ghc-proposals/pull/158>`_).  In particular,
+the proposed design supports updates using a new ``SetField`` class, adds
+support for unlifted datatypes and fields, and specifies laws for the classes.
+
+In order to keep this proposal simple, it does not yet propose adding support
+for type-changing update, which is left for a future proposal.
+
+
+Motivation
+----------
+
+Following `proposal #6 <https://github.com/ghc-proposals/ghc-proposals/pull/6>`_,
+GHC 8.2 introduced a special built-in typeclass ``HasField`` in the
+``GHC.Records`` module, defined thus::
+
+  class HasField (x :: k) r a | x r -> a where
+    getField :: r -> a
+
+When the constraint solver sees a constraint of the form ``HasField "foo" T a``,
+where ``T`` is a concrete datatype and ``foo`` is a symbol corresponding to one
+of its fields, and this field is in scope, the constraint will be solved
+automatically with a dictionary derived from the record selector function for
+the field.
+
+This makes it possible to get a form of type-directed name resolution for field
+selection: given the expression ``getField @"foo" t``, the inferred type of
+``t`` can be used to determine which ``foo`` field is meant, even if there are
+multiple ``foo`` fields in scope and hence the expression ``foo t`` would be
+ambiguous.  (This arises in particular with the ``DuplicateRecordFields``
+extension, which has a somewhat ad hoc mechanism for disambiguating such
+expressions that has been removed in GHC 9.4, following `proposal #366
+<https://github.com/ghc-proposals/ghc-proposals/pull/366>`_.)
+
+GHC 9.2 includes support for using "record dot syntax" for selection with the
+``OverloadedRecordDot`` extension, e.g. ``t.foo`` can be used as syntactic sugar
+for ``getField @"foo" t``.  This is described in the accepted `proposal #282
+<https://github.com/ghc-proposals/ghc-proposals/pull/282>`_ (as modified by
+`proposal #405 <https://github.com/ghc-proposals/ghc-proposals/pull/405>`_).
+However, while the proposals describe both ``OverloadedRecordDot`` and another
+extension ``OverloadedRecordUpdate`` which allows type-based disambiguation of
+record update expressions, only the selection part is fully implemented so far.
+
+The accepted `proposal #158
+<https://github.com/ghc-proposals/ghc-proposals/pull/158>`_ plans to change the
+definition of ``HasField`` to support updates, which is necessary for the full
+implementation of the ``OverloadedRecordUpdate`` extension.
+An implementation of this proposal is available as `GHC merge request !3257
+<https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3257>`_, but has not yet
+been merged, because the compile-time performance cost of the selected
+implementation strategy is unacceptably high.  Such costs were not really
+considered in previous discussions, but it is not appropriate to slow down
+compilation of all programs with records for the benefit only of those using
+``HasField``.
+
+In the light of experience implementing these proposals, and discussion arising
+from `proposal #405 <https://github.com/ghc-proposals/ghc-proposals/pull/405>`_,
+it seems worth systematically re-evaluating the design choices surrounding
+``HasField`` and type-directed name resolution for field updates.
+
+
+Recap: Planned changes to HasField
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The accepted `proposal #158
+<https://github.com/ghc-proposals/ghc-proposals/pull/158>`_ plans to change the
+definitions in ``GHC.Records`` to the following::
+
+  class HasField (x :: k) r a | x r -> a where
+    hasField :: r -> (a -> r, a)
+
+  getField :: forall x r a . HasField x r a => r -> a
+  getField = snd . hasField @x
+
+  setField :: forall x r a . HasField x r a => r -> a -> r
+  setField = fst . hasField @x
+
+This makes it possible to both get and set fields, based on a single class.  The
+``OverloadedRecordDot`` extension would continue to desugar field selection
+syntax to call ``getField``, while the ``OverloadedRecordUpdate`` extension
+would desugar record update syntax to call ``setField``.
+
+Since ``setField`` has not yet been added to a released compiler, the version of
+``OverloadedRecordUpdate`` in GHC 9.2 requires ``RebindableSyntax`` to be
+enabled and a user-defined ``setField`` function to be in scope.  It provides no
+standard definition of this function.
+
+
+Design highlights
+~~~~~~~~~~~~~~~~~
+The essence of the new design is captured in the following definitions, which
+will replace the existing contents of ``GHC.Records``.  For a complete picture
+of the new contents of this module, including auxiliary definitions, see the
+`Proposed Change Specification`_.
+
+::
+
+  type HasField :: forall {k} {r_rep} {a_rep} . k -> TYPE r_rep -> TYPE a_rep -> Constraint
+  class HasField x r a | x r -> a where
+    getField :: r -> a
+
+  type SetField :: forall {k} {r_rep} {a_rep} . k -> TYPE r_rep -> TYPE a_rep -> Constraint
+  class SetField x r a | x r -> a where
+    modifyField :: (a -> a) -> r -> r
+
+  setField :: forall {k} {r_rep} {a_rep} (x :: k)(r :: TYPE r_rep)(a :: TYPE a_rep) . SetField x r a => a -> r -> r
+  setField = modifyField . const
+
+These are the key points of the new design.  Detailed justification for each
+point is deferred to subsequent sections.
+
+* The existing ``HasField x r a`` class continues to have a single method for
+  record field selection, ``getField :: r -> a``.
+
+* There is a new class ``SetField x r a`` for updates, rather than combining
+  both selection and update into the ``HasField`` class (as in `proposal #158
+  <https://github.com/ghc-proposals/ghc-proposals/pull/158>`_).
+
+* ``SetField x r a`` has a single method ``modifyField :: (a -> a) -> r -> r``
+  that allows a field value to be transformed (`proposal #158
+  <https://github.com/ghc-proposals/ghc-proposals/pull/158>`_ does not support
+  this operation).
+
+* The ``setField`` function is implemented in terms of ``modifyField``.
+
+* The order of arguments to ``setField :: a -> r -> r`` is reversed compared to
+  the status quo: it takes the new field value first, followed by the record
+  being updated.
+
+* The classes are representation-polymorphic, allowing support for unlifted
+  fields and datatypes.
+
+* The classes are polymorphic in the kind ``k`` of field labels.
+
+* Functional dependencies are used to allow type inference to determine the
+  field type from the record type and field name.
+
+As noted above, type-changing update is not being considered in this proposal,
+but may be addressed in a follow-up proposal.
+
+
+Motivation for changing the accepted design
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Why change the accepted design from `proposal #158 <https://github.com/ghc-proposals/ghc-proposals/pull/158>`_?
+Defining ``getField`` and ``modifyField`` in separate classes is a better design:
+
+* It gives more flexibility to users, in particular to define read-only or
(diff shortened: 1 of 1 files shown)


proposals/0000-hasfield-redesign.rst
```diff
@@ -0,0 +1,1102 @@
...
+``COMPLETE`` pragmas do give a notion of grouping for pattern synonyms, their
+purpose is currently limited to the pattern-match completeness checker, and is
+not clear that they should have a semantic impact.)
+
+
+Unresolved Questions
+--------------------
+Changing ``SetField`` to support type-changing update is deliberately left out
+of this proposal, so that it can be considered in detail as a subsequent
+proposal.
```
nomeata, line 1094: Is there a design for that already in someone's back's head? Is it going to be a nice extension of what we have here, or a major redesign? (I.e. is it really simpler to do that separately, of better to do all at once?)

adamgundry: There is a half-finished (well, perhaps more like 1/4 finished) design here: https://github.com/adamgundry/ghc-proposals/blob/type-changing-update/proposals/0000-setfield-type-changing-update.rst. It is explicitly structured as an extension to this design (partly because it was originally part of the same draft proposal, before I cut it to make the whole thing more manageable).

The scope of this proposal is something I've struggled with. Type-changing update brings in a lot of complications and potential for things getting derailed, which is why I think it makes sense to try to establish consensus about the (hopefully) less controversial changes first. I don't expect that any of the changes proposed here will need to be significantly rethought in order to add type-changing update later. Some of them are somewhat motivated by it (e.g. splitting into two classes means that `HasField` can be left entirely alone by type-changing update) but I've tried to argue in this proposal that they are independently justifiable, even if type-changing update never materialises.

nomeata: Thanks!

AntC2: Thanks Adam, I think it's sensible to postpone tackling type-changing update. Somebody needing it can revert to using record syntax for pattern-matching on the incoming record, and explicit building the new record with the data constructor prefix. As they can now, so it's not taking anything away.

(I am curious why people are itching for special syntax for update -- as opposed perfectly workmanlike syntax for building a new record. I suppose I should say "so-called 'update'", because there's no _in situ_ update going on like imperative/OO languages: the program is going to build a new record either way; all we're talking about is syntactic sugar.)

IMO the H98 syntax for record update is horribly confusing. I never use it. I wish we could take this opportunity to completely redesign it.

adamgundry: > IMO the H98 syntax for record update is horribly confusing. I never use it. I wish we could take this opportunity to completely redesign it.

My sense is that this ship has sailed, and that introducing a new syntax for updates now is not worth it (given that the existing one will remain for essentially forever). But I'm deliberately trying to steer away from such issues in this proposal, with `HasField` etc. functioning as a common substrate regardless of which syntax one uses.

konsumlamm:
Would `modifyField` be inserted anywhere by the compiler, or would it have to be explicitly used?

I'm a bit concerned about not being able to define "write-only" fields, where it doesn't make sense to pass a "current value" to `modifyField` (though ig one could just pass something like `undefined`). I imagine this could be useful for writing DSLs, although I have no concrete example.

---

Overall I really like this proposal (and I think it's vastly superior to #158).

adamgundry:
@simonpj 
> Huh. I think we should. It's not a full fundep, but our existing fundep machinery will work perfectly well for it: knowing the field name tells the _shape_ of the field type but not (of course) its instantation. And that is precisely what we want.

But if we do this, then we lose confluence: given wanteds `HasField "identity" r (α -> α)` and `HasField "identity" r (β -> β)` then applying the fundep forces `α ~ β`; whereas if we were first to learn `r ~ Rank1` then we could solve both constraints without requiring `α ~ β`.

@konsumlamm 
> Would `modifyField` be inserted anywhere by the compiler, or would it have to be explicitly used?

Only `setField` would be inserted by the compiler, as the desugaring of record update syntax under `OverloadedRecordUpdate` (this is #282). The idea of exposing `modifyField` is principally so it can be used by lens/optics libraries (which effectively provide a different syntax for update).

> I'm a bit concerned about not being able to define "write-only" fields, where it doesn't make sense to pass a "current value" to `modifyField` (though ig one could just pass something like `undefined`). I imagine this could be useful for writing DSLs, although I have no concrete example.

Yes I've wondered a little about this, but I don't have any compelling examples where one really wants to define `setField` but not `modifyField`. (And I don't think we necessarily need `SetField` to handle such cases if they exist: it functions primarily as a mechanism for getting access to setters for built-in record types, not as a universal abstraction over all notions of update.)

simonpj:
> But if we do this, then we lose confluence: given wanteds HasField "identity" r (α -> α) and HasField "identity" r (β -> β) then applying the fundep forces α ~ β; whereas if we were first to learn r ~ Rank1 then we could solve both constraints without requiring α ~ β.

You are so right.  This is where we want [wiggly arrows](https://gitlab.haskell.org/ghc/ghc/-/wikis/Functional-dependencies-in-GHC/Wiggly-arrows)

pnotequalnp:
I like this proposal. In regards to automatically solving instances for partial fields, I'd like to bring up three situations where I think it would make sense to define perfectly sensible, total instances that would conflict with the automatically solved ones with partial methods.

When one constructor's "field" has a layer of indirection.
```hs
data A
  = B { foo :: Int }
  | C { bar :: (Int, String) }

instance HasField "foo" A Int where
  getField = \case
    B {foo} -> foo
    C {bar = (foo, _)} -> foo

instance SetField "foo" A Int where
  modifyField f = \case
    B {foo} -> B {foo = f foo}
    C {bar = (foo, s)} -> C {bar = (f foo, s)}
```

When one constructor's "field" is virtual.
```hs
data A
  = B { foo :: Int }
  | C { bar :: Double }

instance HasField "foo" A Int where
  getField = \case
    B {foo} -> foo
    C {bar} -> round bar

instance SetField "foo" A Int where
  modifyField f = \case
    B {foo} -> B {foo = f foo}
    C {bar} -> C {bar = fromIntegral (f (round bar))}
```

The previously mentioned "affine traversal" case.
```hs
data A
  = B { foo :: Int }
  | C { bar :: String }

instance HasField "foo" A (Maybe Int) where
  getField = \case
    B {foo} -> Just foo
    C {} -> Nothing

instance SetField "foo" A Int where
  modifyField f = \case
    B {foo} -> B {foo = f foo}
    x@C {} -> x
```

Currently with this proposal, these three instances would be rejected at their definition by the restrictions on user-defined `HasField`/`SetField` instances.

With proposal #515, I believe these instances could be written with `-Wno-incoherent-hasfield-instances` and `{-# OVERLAPPING #-}` pragmas. If that proposal is accepted, then it may be a sufficient solution.

Without #515, or if introducing overlapping instances is too heavyweight a solution for what feels like a rather mundane problem, if we want to allow such instances, the ideal solution I think would be to add an option to avoid solving instances for partial fields. However, that's not feasible, since it would have to apply at the use site, rather than at the definition of the datatype and instances, meaning the compiler would not know whether to reject these instances at their definitions.

With the existence of `-Wincomplete-record-selectors` and `-Wincomplete-record-updates`, the concern of solving instances for partial fields causing unexpected runtime errors is I think unwarranted. However, the potential limitation on user-defined instances leads me to question why these instances are solved in the first place. I was unable to find discussion on this in the previous `HasField`-related proposals. My initial impression is that such instances would be desirable only in situations in which the particular constructor of a given value is an invariant of the surrounding code, in which case the partial methods encapsulate what would otherwise be some sort of `case x of A {foo} -> foo; _ -> error "impossible"`. Personally I wouldn't even use the generated `getField` in this situation, as using a custom exception type would likely help in any potential future debugging, for very little cost in upfront work, as well as allowing a higher level of abstraction potentially to recover from the broken invariant by catching the exception. I don't think such small convenience in such rare situations warrants solving instances for partial fields. I'm interested in hearing other people's thoughts on this, and if anyone has examples of these partial instances actually being relied on where not easily replaced by pattern matching (e.g. `case x of A {} -> f x.foo` seems like an unsatisfactory usage).

AntC2:
> @mniip In a large enough haskell program you may end up with a datatype that needs to hide its internals from most of the program to achieve some sort of encapsulation,

I think the mechanism for that is to declare Pattern Synonyms, and make them and their field selectors public. [The proposal does consider `HasField` for PatSyns](https://github.com/adamgundry/ghc-proposals/blob/hasfield-redesign/proposals/0000-hasfield-redesign.rst#712pattern-synonyms). I agree with Adam's assessment there's some fishhooks, so not to address it for a first release.

adamgundry:
@simonpj 

> > But if we do this, then we lose confluence: given wanteds HasField "identity" r (α -> α) and HasField "identity" r (β -> β) then applying the fundep forces α ~ β; whereas if we were first to learn r ~ Rank1 then we could solve both constraints without requiring α ~ β.
> 
> You are so right. This is where we want [wiggly arrows](https://gitlab.haskell.org/ghc/ghc/-/wikis/Functional-dependencies-in-GHC/Wiggly-arrows)

Interesting, I'd been thinking about using wiggly arrows (in spirit if not in concrete syntax) for the follow-up proposal on type-changing `SetField`, but I hadn't considered making `HasField` use them. That would amount to saying that we would no longer use wanteds `HasField f r α`, `HasField f r β` to infer `α ~ β`. In exchange we could use `HasField` on rank-1 polymorphic fields. I'll have to think further about that trade-off.


@pnotequalnp I think the main reason for `SetField` being solved for partial fields is for consistency with traditional record updates. You're right that we could instead refuse to solve such constraints automatically, perhaps controlled by some option at the definition site. (#535 is something like this.)

@mniip good point that we don't have a way to control scope for virtual fields. I would hope that libraries don't normally need to define such "internal-only" virtual fields.

Kleidukos:
Sorry if this has been already discussed somewhere else, but was there any consideration for generating accessors of type `Maybe a` for partial fields? In the absence of a more powerful proof system, wouldn't it be wiser to avoid partiality through `Maybe`? 

(I'm suggesting this because I quite appreciate [`(^?)` & `preview`](https://hackage.haskell.org/package/optics-core-0.4.1/docs/Optics-AffineFold.html#v:preview) and it would be good UX to promote safety and totality instead of partiality)

adamgundry:
> Sorry if this has been already discussed somewhere else, but was there any consideration for generating accessors of type `Maybe a` for partial fields? In the absence of a more powerful proof system, wouldn't it be wiser to avoid partiality through `Maybe`?

@Kleidukos there's some discussion in this direction on #535. We could plausibly do this, although the details need a bit of working out.

arybczak:
What's the status here? :slightly_smiling_face:

adamgundry:
Apologies, I think discussion has subsided, but I should review it and make any appropriate changes to the text before submitting to the committee. I'm afraid I've been rather distracted with other things...

adamgundry:
I'm still planning to come back to this, but for (my) future reference, the proposal could do with saying more about possible "virtual field" `HasField` instances, several of which have been suggested in different venues:
 * https://github.com/ghc-proposals/ghc-proposals/issues/600
 * https://github.com/haskell/core-libraries-committee/issues/143
 * https://github.com/haskell/core-libraries-committee/issues/191

googleson78:
Just to add another usage to virtual fields - I've heard of people using them for backwards compat, i.e. they rename a field but they add a `HasField` with the old field name and value, using the new actual record fields.

Personally I also have a couple of usages in the "style" of `PatternSynonyms`, where you don't want to add a new field(constructor), but you want to expose a value as a "field" because it intuitively feels like it might be one (constructor). I'm not so sold on this use case, but it's mostly an experiment so far for me.

mixphix:
I quite like virtual fields, particularly for surjections:

```hs
data Box = Box1 | Box2 | Box3

data Complicated = Complicated
  {
  ...
  ...
  ...
  }

toBox :: Complicated -> Box
toBox c
  | ... -> Box1
  | ... -> Box2
  | otherwise -> Box3

instance HasField "box" Complicated Box where
  getField = toBox
```

I just don't think `Functor`s of things should inherit their values' fields.

As for tuple accessors, I think we can do better than a plethora of orphan instances?

googleson78:
Just reposting here, since it's relevant to the virtual fields discussion:

https://github.com/haskell/core-libraries-committee/issues/191#issuecomment-1645747584

To be frank I hadn't though of this one, it's extremely clever and simultaneously a bitter reminder we don't have qualified re-exports :sweat_smile:

googleson78:
I have another use case that I often think about, and am employing partially*, something like this:
```haskell
data Env m r = Env

instance (MonadReader r m, HasField x r a) => HasField x (Env m r) (m a)
...

-- intended to be used like this:
-- bla = do
--   foo <- Env.foo
```

Which is a very minor convenience, compared to `asks (.foo)`, but I think it at least demonstrates that virtual fields can give a non-insignificant boost to the "Haskell as a framework for EDSL" usage that Haskell is somewhat known for (and we do indeed have quite a few nice EDSL libraries), in that a person now does not need to concern themselves what a Reader is or whatever.

Additionally, once we have split `SetField` and `GetField`, I can easily imagine* e.g.
```haskell
data Put w r = Put

instance (MonadWriter w m, HasField x r a, Monoid w) => SetField x (Put w r) (m ())
```

\* Unfortunately both of these instances are far from as straightforward as I've presented them here, due to issues with functional dependencies and type inference. My "partially employing" is a case where I do use the idea behind this particular instance, but specialised to a concrete `MonadReader`, which alleviates the type inference/fundep issues.

BinderDavid:
> As for tuple accessors, I think we can do better than a plethora of orphan instances?

(Context: I proposed these instances for tuples). If these instances are provided in `base`, then they wouldn't have to be orphan instances. I just currently provide them as orphan instances in https://hackage.haskell.org/package/tuple-fields, since the consensus in the responses to my proposal seemed to be that the RecordDot ecosystem / infrastructure is currently not stabilized enough to consider adding them to base at this point.

adamgundry:
Thanks everyone for your feedback. I've pushed some small updates to the proposal, exploring points that were raised in the discussion (especially about partial and virtual fields).

@nomeata I think this is ready to submit to the committee when convenient.

parsonsmatt:
> Moreover there is an implicit (and I think un-discussed) design choice here: it would be perfectly sensible for `modifyField` to be a no-op (rather than a crash) on a data constructor that lacks the relevant field. Have we disucssed that?

This was sort-of discussed in [Partial Field Behavior #535](https://github.com/ghc-proposals/ghc-proposals/pull/535). 

I see a few options when the field isn't on the constructor:

1. Fail loudly with `error`
2. Fail silently with a no-op update
3. Indicate in the types that the field wasn't present

I think the first option is probably *rarely* what people want, and shouldn't be the default. Alas.

The second option feels slightly nicer to me, but it *also* feels a bit out of character - we don't usually have a "permissive" style in Haskell, where we try our best to always prevent a runtime crash *now* at the expense of possible silent bugs slipping through.

In the third case, we need some mechanism for `modifyField` to "sequence" the `Maybe`-ness. If I have my `lens` hat on, then I'd think an interface like:

```haskell
modifyPartialField :: (ModifyField sym r r' t t') => (t -> t') -> r -> r'
```

would allow us to write, say,

```haskell
data T = A { a :: Int } | B

instance ModifyField "a" T (Maybe T) Int Int
```

which concretizes to:

```haskell
modifyField @"a" @A 
    :: (Int -> Int) 
    -> T -> Maybe T
```

And then you can recover the error or no-op behavior:

```haskell
modifyFieldError :: ModifyField sym r (Maybe r') t t' => (t -> t') -> r -> r'
modifyFieldError k r = fromMaybe (error "field not on constr") (modifyField k r)

modifyFieldNoop :: ModifyFieild sym r (Maybe r) t t => (t -> t) -> r -> r
modifyFieldNoop k r = fromMaybe r (modifyField k r)
```

---

OK, now that I've commented on the discussion, I guess I'll actually read the proposal 😅

parsonsmatt:
OK, having read the actual proposal: yeah! This seems like a good step forward.

I will say that I think solving type changing update *is* important, and I really would prefer to see that settled out before we commit to a design that may require significant overhaul in order to support type changing update later.

AntC2:
> 1. Fail loudly with `error`
> 2. Fail silently with a no-op update
> 3. Indicate in the types that the field wasn't present
> 
> I think the first option is probably _rarely_ what people want, and shouldn't be the default. Alas.

1. is what I would want. But rather than another fruitless debate, I think we say this proposal is really providing nicer syntax for existing behaviour; and 1. is existing behaviour.
2. I agree it would be nice to have a `modifyFieldNoop`. Does it matter where/how deep in a nested record the field access breaks down?
3. I don't see how we can reflect from a term (a particular constructor) back to a type. Again let's keep to a modest aim of nicifying existing behaviour.

AntC2:
> I see a few options when the field isn't on the constructor:

The proposal does mention [`-Wincomplete-record-selectors`](https://github.com/adamgundry/ghc-proposals/blob/hasfield-redesign/proposals/0000-hasfield-redesign.rst#23change-to--wincomplete-record-updates) -- and that's an idea that's been floating around for a long time in various forms. Can we expect that #516 will be released before/at the same time as this `HasField` redesign? I think the worry is warnings will be too noisy/won't be able to follow the code's logic restricting access to only the constructors with the field.

We need to get that warning out there, so to get some road miles on it.

adamgundry:
> I will say that I think solving type changing update _is_ important, and I really would prefer to see that settled out before we commit to a design that may require significant overhaul in order to support type changing update later.

I have a very rough plan for type-changing update (https://github.com/adamgundry/ghc-proposals/blob/type-changing-update/proposals/0000-setfield-type-changing-update.rst) but the details of how type inference works need clearer specification and tend to get bogged down in the details of how `FunctionalDependencies` should work.

The key change I'd envisage relative to this proposal would be to add parameters so we go from `SetField x r a` to `SetField x s t a b`. That would break "virtual field" instances for `SetField`, of course, but not much else needs to change. But is that much worse than not having `SetField` at all until we have a type-changing version? I think even non-type-changing `SetField` would be a useful step forward.

I suppose one thing we could do now would be to add the parameters for type-changing update already, but restrict the constraint solver behaviour, e.g. define the classes `SetField x s t a b` and `type SetField' x r a = SetField x r r a a`, but have the constraint solver always unify `s ~ t` and `a ~ b` when solving constraints. That would limit (but not necessarily eliminate) the breakage when generalising the constraint solver.

That's somewhat tempting. On the other hand, given that it sounds like the CLC may take the position that `SetField` belongs in a new module in `ghc-experimental` anyway (https://github.com/haskell/core-libraries-committee/issues/194#issuecomment-1674007990), perhaps we shouldn't be too shy about potentially changing it later?

parsonsmatt:
I just had an insight that the type-changing update would be *really really cool* for anonymous record support.

```haskell
data AnonRec (fields :: '[(Symbol, Type)])

instance SetField (fieldName :: Symbol) (AnonRec fields) (AnonRec ( '(fieldName, type) ': fields) type where
  setField = setAnonRecField @fieldName

anon :: AnonRec '[]

neat :: AnonRec '[("name", String)]
neat = anon { name = "neat" }
```

adamgundry:
I've updated the proposal following CLC feedback, so that `SetField` is now in a separate module (in the forthcoming `ghc-experimental` package).

proposals/0000-hasfield-redesign.rst
```diff
@@ -0,0 +1,1193 @@
...
+definition in the interests of generality and compatibility.  For example,
+the ``record-hasfield`` library makes use of the possibility to define label
+kinds other than ``Symbol``, allowing tuples of labels to be used for
+composition of fields.  In particular, it defines an instance like::
+
+  instance (HasField x1 r1 r2, HasField x2 r2 a2)
+      => HasField '(x1, x2) r1 a2
+
+which means ``getField @("foo", "bar")`` will be treated like the composition
+``getField @"bar" . getField @"foo"``.
```
aspiwack, line 1025: Another application I can imagine is to have instances for tuples

`instance HasField 1 (x,y)`

Not sure that it's particularly useful (unless we are able to use `u.1` as a notation), but putting it out there.

proposals/0000-hasfield-redesign.rst
```diff
@@ -0,0 +1,1193 @@
...
+
+
+
+Representation polymorphism
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
aspiwack, line 1030: This paragraph doesn't seem to be presenting an alternative, it restates something that is already in the proposed changes. Did I miss something?

proposals/0000-hasfield-redesign.rst
```diff
@@ -0,0 +1,1193 @@
...
+record update expressions, only the selection part is fully implemented so far.
+
+The accepted `proposal #158
+<https://github.com/ghc-proposals/ghc-proposals/pull/158>`_ plans to change the
+definition of ``HasField`` to support updates, which is necessary for the full
+implementation of the ``OverloadedRecordUpdate`` extension.
+An implementation of this proposal is available as `GHC merge request !3257
+<https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3257>`_, but has not yet
+been merged, because the compile-time performance cost of the selected
+implementation strategy is unacceptably high.  Such costs were not really
```
aspiwack, line 65: You don't touch upon it in the proposal, but I assume that using a van Laarhoven lens type would solve the performance issue. It does pose other problems (in particular is not easy to make a such a type representation parametric), but presumably a word should be said somewhere in the _Alternatives_ section.

aspiwack (commented):
As the shepherd for this proposal, I will soon make a recommendation to the committee.

Before I do, I have a few requests for clarification below.

proposals/0000-hasfield-redesign.rst
adamgundry: Well, the implicit alternative is "don't support representation polymorphism", just as the next section argues not to support multiplicity polymorphism.

proposals/0000-hasfield-redesign.rst
adamgundry: Nice idea. In fact I imagine one could use generics to implement an instance of `HasField (n :: Nat) t` for tuples and/or record types...

adamgundry (commented):
Thanks @aspiwack! I'll work on some updates in response to your comments.

aspiwack:
Thanks for the replies. They're very helpful. Make sure to fold it into the proposal and we'll be good to go.

adamgundry:
@aspiwack I've now updated the proposal text, taking into account your comments.

The only significant change is that `setField` is now a class method, with a default definition in terms of `modifyField` (and vice versa). Having thought about it, I'm increasingly of the opinion that it is a better design from a user perspective, because in most cases users can implement whichever they prefer. We can always revisit this if it turns out to be a problem from an implementation perspective.

proposals/0000-hasfield-redesign.rst
```diff
@@ -0,0 +1,1266 @@
...
+
+
+Setting vs modification
+~~~~~~~~~~~~~~~~~~~~~~~
+
+A previous iteration of the design supported only ``setField :: a -> r -> r`` and not
+``modifyField :: (a -> a) -> r -> r``.  The latter generalises ``setField`` to
+allow modifying any ``a`` values in the datatype (of which there may be none).
+
+It is easy to implement ``setField`` in terms of ``modifyField``, but not vice
```
nomeata, line 985: Is that true even when thinking about memory retention? `setField` can probably drop the reference to the value where `modifyField` can't easily. Do we need `modiftField'`?

nomeata: And performance-worried users would be encouraged to define `setField` explicitly, even if they also define `modifyField`? (Unless the selector thunk optimization applies reliably. Here the think would be `const x y`, right?)

parsonsmatt: In the interest of keeping concerns separate, why not provide `ModifyField` as a separate class entirely?

```haskell
class ModifyField ... where
    modifyField :: (a -> b) -> s -> t
    default modifyField :: (HasField sym s a, SetField sym s b, s ~ t, a ~ b) => (a -> b) -> s -> t
```

This allows potential virtual field use of `SetField` that don't have a corresponding `HasField`, and allows `modifyField` to be optimized in the case where there's a potentially more efficienct solution

adamgundry: > And performance-worried users would be encouraged to define `setField` explicitly, even if they also define `modifyField`? (Unless the selector thunk optimization applies reliably. Here the think would be `const x y`, right?)

Thinking about it further, I suppose the selector thunk optimization could help if you have `modifyField` defined in terms of `setField` (because the `getField` selector should be shortcutted) but not vice versa (because I don't think `const` is subject to the optimization, although I wonder if it could be).

I would encourage performance-worried users to make their datatype fields strict. :-)

> why not provide `ModifyField` as a separate class entirely?

We could, but that feels excessive to me. We don't need `modifyField` for `OverloadedRecordUpdate`, so GHC itself wouldn't make use of it. Users defining virtual fields would need to remember to add an extra instance per field (albeit one with a default method) that would only be used in certain situations, and we need laws relating three independent classes whose instances may or may not exist.

I suppose one potentially more compelling option would be to have `ModifyField` support type-changing update (as you hint in your code snippet), and add it independently of `SetField` which could remain non-type-changing. If we wanted to keep `OverloadedRecordUpdate` non-type-changing but add type-changing update support for the optics aficionados, that might not be a terrible approach?

proposals/0000-hasfield-redesign.rst
```diff
@@ -0,0 +1,1266 @@
...
+    modifyField :: (a -> a) -> r -> r
+    default modifyField :: (r_rep ~ LiftedRep, a_rep ~ LiftedRep, HasField x r a) => (a -> a) -> r -> r
+    modifyField f r = setField @x (f (getField @x r)) r
+
+    -- | Update function to set the field @x@ in the record @r@.
+    setField :: a -> r -> r
+    default setField :: a_rep ~ LiftedRep => a -> r -> r
+    setField v = modifyField @x (\ _ -> v)
+
+    {-# MINIMAL modifyField | setField #-}
```
konsumlamm, line 256: Is this really true? `modifyField` and `setField` don't always have default definitions.

adamgundry: I don't think that's how `DefaultSignatures` works, but I haven't actually verified it experimentally...

konsumlamm: You indeed get a type error if the default signature doesn't match, so there is no problem here.

simonpj:
Adam you write

> I suppose one thing we could do now would be to add the parameters for type-changing update already, but restrict the constraint solver behaviour, e.g. define the classes SetField x s t a b and type SetField' x r a = SetField x r r a a, but have the constraint solver always unify s ~ t and a ~ b when solving constraints. That would limit (but not necessarily eliminate) the breakage when generalising the constraint solver.

Can you say a bit more? What does the class look like?  Whey does the constraint solver have do to the "always unify..." bit?  

I'm asking because, other things being equal, future-proofing the proposal for type-changing update would be a good thing.  But are other things equal?

adamgundry:
The key change is that rather than having
```hs
class SetField x r a | x r -> a where
  modifyField :: (a -> a) -> r -> r
  setField :: a -> r -> r
```
we would have something like
```hs
class SetField x s t a b | x s -> a, x t -> b, x s -> t, x t -> s where
  modifyField :: (a -> b) -> s -> t
  setField :: b -> s -> t
```

Then when GHC solves a `SetField` constraint for a particular record type, it would look for something like `SetField "fld" (R v) t a b`, check that `R v` has a field `fld` of type `F v`, and emit wanteds `t ~ R v` , `a ~ F v`, `b ~ F v`. (It would also solve `SetField "fld" s R a b`, mutatis mutandis.)

This could later be generalised so that solving `SetField "fld" (R v) t a b` does not force the field types and record types to be unified, instead emitting wanteds `t ~ R β`, `a ~ F v`, `b ~ F β` for some fresh `β`.

The choice of functional dependencies would need some thought. I think that by having fundeps `x s -> t, x t -> s` we restrict updates to be non-type-changing (unless the user defines "dysfunctional" virtual instances). My provisional plan (sketched very roughly at https://github.com/adamgundry/ghc-proposals/blob/type-changing-update/proposals/0000-setfield-type-changing-update.rst) would be to relax these fundeps to be wiggly arrows (https://gitlab.haskell.org/ghc/ghc/-/wikis/Functional-dependencies-in-GHC/Wiggly-arrows). But figuring out the details of that is something I'm explicitly trying not to do in the present proposal!

Another option would be to have `s ~ t`, `a ~ b` as superclass constraints on `SetField`. That would entirely rule out type-changing instances (even dysfunctional ones), but it means that later removing those superclass constraints would potentially break more things.

Overall, I'm somewhat on the fence here. If we do manage to push through type-changing update in fairly short order, it would be annoying to have to change `SetField` to add extra parameters and thereby break previously-working code (albeit code relying on `ghc-experimental`). But it's also conceivable that progress will stall when trying to nail down type-changing update, in which case it may be annoying to have all these extra parameters that promise more than GHC is able to deliver.

When adding type-changing update, we could reduce the breakage by adding the generalised `SetField` with a different name or in a different module (i.e. have two different classes, `SetField` and `TypeChangingSetField`). But we'd still need to pick one for `OverloadedRecordDot` to use, and unless we added lots of magic it wouldn't be possible for instances or givens from one to apply to the other, so one of them would feel somewhat second-class.

goldfirere:
It seems to me that you could do what you described in the first paragraph -- but without the fundeps. That is, when the solver is working on a bona-fide record, it just emits extra wanteds. Otherwise, it doesn't -- no fundeps. In some sense, this is putting equality constraints on all the synthetic instances used to solve the `SetField` constraints. Could that work?

parsonsmatt:
I made a [GitHub repository](https://github.com/parsonsmatt/setfield-design-thoughts/blob/master/recupd.hs) with a draft of the design I'd done. 

This class supports type-changing update:

```haskell
class SetField sym s t a | sym s t -> a where
    setField :: a -> s -> t
```

Since you don't have `modifyField` as a class member, you don't need the `b` type parameter for the original. `modifyField` then has thist ype:

```haskell
modifyField :: forall sym s t a b. (HasField sym s a, SetField sym s t b) => (a -> b) -> s -> t
modifyField f s = setField @sym (f (getField @sym s)) s
```

Monomorphic updates have fantastic inference properties with a `(s ~ t)` constraint in the instance:

```haskell
data A = A { name :: String, age :: Int }

instance (A ~ t) => SetField "name" A t String where
    setField str (A _oldStr oldAge) = A str oldAge

instance (A ~ t) => SetField "age" A t Int where
    setField age (A oldStr _oldAge) = A oldStr age
```

This design permits "builder pattern" virtual instances:

```haskell
data OnlyName = OnlyName { name :: String }

instance (A ~ t) => SetField "age" OnlyName t Int where
    setField age (OnlyName oldStr) = A oldStr age

createA :: Int -> OnlyName -> A
createA age onlyName = setField @"age" age onlyName
-- with sugar,
    -- onlyName { age = age }
```

It allows multiple updates to be chained with a single extra constraint:

```haskell
multiUpdateFn
    :: forall nameS nameT ageS ageT a.
    ( SetField "name" nameS nameT String
    , SetField "age" ageS ageT a
    , ageT ~ nameS
    -- ^ A constraint that the output of updating the age is the same as
    -- the input to updating the name. This is necessary for GHC to provide.
    , Num a
    )
    => ageS
    -> nameT
multiUpdateFn x =
    (setField @"name" @nameS "hello")  (setField @"age" 5 x)
```

Polymorphic updates work fine:

```haskell
data PolyA a = PolyA { name :: a, age :: Int }

instance  SetField "name" (PolyA a) (PolyA b) b where
    setField str (PolyA _oldStr oldAge) = PolyA str oldAge

instance (PolyA a ~ t) => SetField "age" (PolyA a) t Int where
    setField age (PolyA oldStr _oldAge) = PolyA oldStr age
```

However, there *is* a problem: type changing update when the type variable is mentioned by multiple fields. You need to ensure that the record update expression has *all* of the fields represented.

int-index:
>  type changing update when the type variable is mentioned by multiple fields

Good observation, though it leads me to believe that `SetField` might be the wrong abstraction for type-changing updates, as type variables are often used in more than one field.

And while the type inference problems seem solvable to me (with enough fine-tuning), the multiple-fields problem seems to be inherent in the whole idea of `SetField` which only attempts to set a single field.

adamgundry:
> And while the type inference problem seems solvable to me (with enough fine-tuning), the multiple-fields problem seems to be inherent in the whole idea of `SetField` which only attempts to set a single field.

Yes, I agree with this. I've thought occasionally about how to handle multiple simultaneous field updates, but it seems to be difficult without introducing a lot of machinery. 

The best option I'm aware of is that given a datatype with a type parameter shared between fields, e.g.
```hs
data T a = MkT { f :: a, g :: a }
```
you can always add extra parameters so that each parameter belongs to a unique field, e.g.
```hs
data T' a1 a2 = MkT { f :: a1, g :: a2 }
type T = T' a a
```
and now you can compose two calls to (type-changing) `setField`. But of course this requires significant refactoring of the original datatype, which may well be infeasible in many cases.

parsonsmatt:
My draft allows for non-type-changing-update on a record with a duplicate:

```haskell
data SamePolyVar a = SamePolyVar { name :: a, age :: a }

instance (t ~ SamePolyVar a) => SetField "name" (SamePolyVar a) t a where
    setField str (SamePolyVar _oldStr oldAge) = SamePolyVar str oldAge

instance (t ~ SamePolyVar a) => SetField "age" (SamePolyVar a) t a where
    setField age (SamePolyVar oldStr _oldAge) = SamePolyVar oldStr age

-- | But type inference works fine, if the input type is known.
updateSamePolyVar :: SamePolyVar _ -> _ -- SamePolyVar Int
updateSamePolyVar =
    setField @"name" 10 . setField @"age" 20
```

In order to make it work for type-changing update in this context, there's not really that much magic necessary here - if GHC knows that "all polymorphic fields are being updated in one go" then it can "just" do a coercion.

```haskell
updateSamePolyVar :: SamePolyVar Char -> SamePolyVar Int
updateSamePolyVar x =
    setField @"name" 10 $ (setField @"age" 20 (unsafeCoerce x :: SamePolyVar Int))
```

I think GHC ought to be able to know that, given that the record update expression is all grouped in.

```haskell
-- sugared up,
foo :: SamePolyVar Char -> SamePolyVar Int
foo spv = spv { name = 10, age = 30 }
```

AntC2:
> the multiple-fields problem seems to be inherent in the whole idea of `SetField` which only attempts to set a single field.

Yes exactly. In full generality it needs to cope with updates to

```
data Dab a b = MkDab{ fa :: a, fb :: b, fab :: (a, b) }
```


> the record update expression is all grouped in. ...

The H98 syntax shows we're not really 'updating' anything, but building a fresh new value. The trouble is `SetField` working field-at-a-time wants to build a separate/intermediate new value from each field; and those intermediate values are ill-typed. (This was all well understood 6 months ago when this proposal started, which is why type-changing update is out of scope.)

aspiwack:
With apologies about the delay, the committee has decided to accept this proposal without modification.

@adamgundry , you're aware, as a member of the committee, but for the sake of transparency: there's been a bit of anxiety around the backward incompatible change and how it could be best handled. The consensus has eventually been that the extension is clearly enough marked as experimental so that the change is acceptable (and a smoother change would be hard to enact).

@nomeata as usual, I'll leave you the honor of merging the PR.

PR #585, opened 2023-03-30
Or-pattern syntax (amends #522)

knothed:
This is an amendment to proposal #522.

We propose three changes to the current `one of` syntax:
 1. change syntax from `one of 1, 2, ...` to `one of {1; 2; ...}` and make `one of` subject to the [layout rule](https://www.haskell.org/onlinereport/haskell2010/haskellch2.html#x7-210002.7). This means that we don't require the curly braces, so `one of 1;2;...` is also good.
 2. put production into `pat` instead of `apat` and without parentheses around it
 3. allow or-pattern to have 0 or 1 arguments.

Why?

### 1. `one of {...}` syntax

Similar to `case of`, we want to make use of layout blocks. Writing explicit `{}` around the pattern gives a much nicer grouping for code readers. Previously, one could misread `(one of 1, 2)` as `((one of 1), 2)`. Now, `one of {1;2}` makes this impossible.

Still, the `{}` are optional. This allows us to write or-patterns in a multilined fashion:
```hs
f (one of 1
          2
          3) -> ...
```

`of` layout automatically inserts `{}` and `;`.

### 2. put production unparenthesized into `pat`:

This allows one-of to appear:
- unparenthesized inside tuples: `f (3, one of {1;2}, 5) = 1`
- unparenthesized inside right sides of view-pats: `f (head -> one of {1;2}) = 3`

which feels more natural than having `()` around.

The rule is not too relaxed; this e.g. is still forbidden: `f one of 2;3 = 4`

### 3. or-pats with 0 or 1 arguments

This was requested by @eyeinski and seems like a nice thing to have. It may make sense to write an empty or-pattern and only later filling it with values. Also, similar to `-XEmptyCase`, the value is evaluated before the pattern match fails.

An empty or-pattern will give a redundant pattern match warning.

### DIFF
--- a/proposals/0522-or-patterns.rst
+++ b/proposals/0522-or-patterns.rst
@@ -92,8 +92,6 @@ Similar to ``case of``, ``of`` introduces a layout block and implicitly adds ope
   - ``one of {1;2}``
   - ``one of 1;2``
 
-When ``k=0``, we only allow ``one of {}``, not ``one of`` however.
-
 Some examples that this new grammar produces: ::
 
   case e of (one of T1; T2{}; T3 a b) -> ...
@@ -102,7 +100,7 @@ Some examples that this new grammar produces: ::
                     T2{}
                     T3 a b) -> ...
                     
-  case e of (one of {}) -> ...
+  case e of (one of) -> ...
 
   f :: (Int, Int) -> Int
   f (5, one of {6;7})
--- a/proposals/0522-or-patterns.rst
+++ b/proposals/0522-or-patterns.rst
@@ -1,4 +1,4 @@
-Or Patterns
+Or Patterns (Amended)
 ==============
 
 .. author:: David Knothe, Ömer Sinan Ağacan, Sebastian Graf 
@@ -6,7 +6,7 @@ Or Patterns
 .. ticket-url::
 .. implemented::
 .. highlight:: haskell
-.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/todo>`_.
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/585>`_.
 .. sectnum::
 .. contents::
 
--- a/proposals/0522-or-patterns.rst
+++ b/proposals/0522-or-patterns.rst
@@ -1,12 +1,12 @@
 Or Patterns
 ==============
 
-.. author:: David Knothe, Ömer Sinan Ağacan, Sebastian Graf
-.. date-accepted:: 2022-11-29
+.. author:: David Knothe, Ömer Sinan Ağacan, Sebastian Graf 
+.. date-accepted::
 .. ticket-url::
 .. implemented::
 .. highlight:: haskell
-.. header:: This proposal was `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/522>`_.
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/todo>`_.
 .. sectnum::
 .. contents::
 
@@ -62,7 +62,7 @@ function we get
 
     stringOfT :: T -> Maybe String
     stringOfT (T1 s)        = Just s
-    stringOfT (one of T2{}, T3{}) = Nothing
+    stringOfT (one of T2{}; T3{}) = Nothing
 
 This function doesn't match ``T4``, so we get our warning in the very first compile
 cycle or (even faster) in our IDE powered by a language server implementation.
@@ -76,45 +76,51 @@ Changes in the grammar
 
 We consider this as an extension to `Haskell 2010 grammar
 <https://www.haskell.org/onlinereport/haskell2010/haskellch10.html#x17-18000010.5>`_.
-The relevant non-terminal is ``apat``: ::
+The relevant non-terminal is ``pat``: ::
 
-  apat    →    var [ @ apat]                     (as pattern)
-          |    gcon                              (arity gcon  =  0)
-          |    qcon { fpat1 , … , fpatk }        (labeled pattern, k ≥ 0)
-          |    literal
-          |    _                                 (wildcard)
-          |    ( pat )                           (parenthesized pattern)
-          |    ( pat1 , … , patk )               (tuple pattern, k ≥ 2)
-          |    [ pat1 , … , patk ]               (list pattern, k ≥ 1)
-          |    ~ apat
+   pat    →    lpat qconop pat
+          |    lpat
 
 Or patterns extension adds one more production: ::
 
-          |    (one of pat1, …, patk )                (Or pattern, k ≥ 2)
+          |    one of { pat1 ; ... ; patk }                (Or pattern, k ≥ 0)
 
 ``one`` is a conditional keyword in patterns, but can still be used for variable patterns.
-The ``,`` between the parentheses have (shift) priority that is lower than any other ``apat``'s (reduction) priority.
+
+Similar to ``case of``, ``of`` introduces a layout block and implicitly adds opening and closing curly braces when not explicitly given. Therefore, both of these are equivalent:
+
+  - ``one of {1;2}``
+  - ``one of 1;2``
+
+When ``k=0``, we only allow ``one of {}``, not ``one of`` however.
 
 Some examples that this new grammar produces: ::
 
-  -- in expression context
-  case e of
-    (one of T1, T2{}, T3 a b) -> ...
+  case e of (one of T1; T2{}; T3 a b) -> ...
+  
+  case e of (one of T1
+                    T2{}
+                    T3 a b) -> ...
+                    
+  case e of (one of {}) -> ...
 
+  f :: (Int, Int) -> Int
+  f (5, one of {6;7})
+  f (fst -> one of 0) -> Int
+  
   -- in expression context
-  let (one of [x], (x : y : z)) = e1 in e2
+  let (one of [x]; (x : y : z)) = e1 in e2
 
   -- pattern guards in declarations
   f x y
-    | x@(one of T1 _, T2 a b) <- e1
+    | x@(one of T1 _; T2 a b) <- e1
     , guard x
     = e2
 
   -- nested Or patterns
   case e1 of
-    (one of (one of T1, T2), T3, T4) -> e2
-
-The new production doesn't add any ambiguities because of the mandatory parentheses, just like for tuples.
+    (one of one of {T1; T2}; T3; T4) -> e2
+    (one of one of one of {})
 
 NB: The new grammar allows Or patterns which bind variables. These will however be rejected in `2.2`_.
 
@@ -135,12 +141,10 @@ We give the static semantics in terms of *pattern types*. A pattern type has the
 Then the typing rule for Or patterns is:
 ::
 
-    Γ0, Σ0 ⊢ pat1 : τ ⤳ Γ0,Σ1,Ψ1    Γ0, Σ0 ⊢ pat2 : τ ⤳ Γ0,Σ2,Ψ2    
-    -------------------------------------------------------------
-                 Γ0, Σ0 ⊢ (one of pat1, pat2) : τ ⤳ Γ0,Σ0,∅
+          Γ0, Σ0 ⊢ pat_i : τ ⤳ Γ0,Σi,Ψi
+    -----------------------------------------
+    Γ0, Σ0 ⊢ (one of { pat_i }) : τ ⤳ Γ0,Σ0,∅
 
-
-An Or pattern consisting of more than two parts works the same.
 
 
 Dynamic semantics of Or pattern matching
@@ -150,24 +154,25 @@ Informal semantics in the style of `Haskell 2010 chapter 3.17.2: Informal
 Semantics of Pattern Matching
 <https://www.haskell.org/onlinereport/haskell2010/haskellch3.html#x8-600003.17.2>`_:
 
-- Matching the pattern ``(one of p1, …, pk)`` against the value ``v`` is the result of matching ``v`` against ``p1`` if it is not a failure, or the result of
-  matching ``(one of p2, …, pk)`` against ``v`` otherwise.
+- Matching the pattern ``one of {p1; ... ; pk}`` against the value ``v`` is the result of matching ``v`` against ``p1`` if it is not a failure, or the result of
+  matching ``(one of {p2; ... ; pk}`` against ``v`` otherwise. We require that ``p1``, …, ``pk`` bind no variables.
+- Matching the pattern ``one of {}`` against the value ``v`` fails after evaluating ``v``.
 
-  ``p1``, …, ``pk`` bind no variables.
-
-  NB: For k=1, the pattern ``(p1)`` is meant to denote a parenthesized pattern.
 
 Here are a few examples: ::
 
-    (\ (one of 1, 2) -> 3) 1 => 3
-    (\ (one of Left 0, Right 1) -> True) (Right 1) => True
-    (\ (one of (one of [1], [2, _]), (one of [3, _, _], [4, _, _, _])) -> True) [4, undefined, undefined, undefined] => True
-    (\ (one of 1, 2, 3) -> True) 3 => True
+    (\ (one of 1; 2) -> 3) 1 => 3
+    (\ (one of Left 0; Right 1) -> True) (Right 1) => True
+    (\ (one of (one of [1]; [2, _]); (one of [3, _, _]; [4, _, _, _])) -> True) [4, undefined, undefined, undefined] => True
+    (\ (one of 1; 2; 3) -> True) 3 => True
+    
+    g (one of {}) = 1
+    g (error "a") => error "a"
 
 We do not employ backtracking in Or patterns. The following would yield ``"no backtracking"``: ::
 
  case (True, error "backtracking") of
-   (one of (True, _), (_, True))
+   (one of (True, _); (_, True))
      | False -> error "inaccessible"
    _ -> error "no backtracking"
 
@@ -213,10 +218,10 @@ Examples
       where
         go (L _ pat) = go1 pat
 
-        go1 (one of WildPat{}, VarPat{}, LazyPat{})
+        go1 (one of WildPat{}; VarPat{}; LazyPat{})
           = True
 
-        go1 (one of PArrPat{}, ConPatIn{}, LitPat{}, NPat{}, NPlusKPat{}, ListPat {})
+        go1 (one of PArrPat{}; ConPatIn{}; LitPat{}; NPat{}; NPlusKPat{}; ListPat{})
           = False
 
         go1 (BangPat pat)       = go pat
@@ -294,13 +299,13 @@ So the following example would not type check because the Or pattern doesn't pro
      IsInt2 :: GADT Int
 
  foo :: a -> GADT a -> a
- foo x (one of IsInt1 {}, IsInt2 {}) = x + 1
+ foo x (one of IsInt1 {}; IsInt2 {}) = x + 1
 
 
 Considering view patterns, these do work seamlessly with Or patterns. As specified in `2.2`_, Or patterns will just merge the required constraints which come from view patterns. This would work: ::
 
  f :: (Eq a, Show a) => a -> a -> Bool
- f a (one of (== a) -> True, show -> "yes") = True
+ f a (one of (== a) -> True; show -> "yes") = True
  f _ _ = False
 
 Costs and Drawbacks
@@ -316,7 +321,7 @@ Alternatives
 Alternative syntax
 ~~~~~~~~~~~~~~~~~~
 
-We previously considered ``;``, ``;;`` and ``||`` as separators.
+We previously considered ``;``, ``;;`` and ``||`` as separators, without the `one of` prefix.
 
 One nice thing about using ``;`` for the separator is that it is also used
 for separating case alternatives, so it looks familiar. Example: ::
@@ -351,8 +356,8 @@ The `parent proposal <https://github.com/ghc-proposals/ghc-proposals/pull/43>`__
 
  data T = T1 Int | T2 Int | T3 | T4
 
- getInt (one of T1 a, T2 a) = Just a
- getInt (one of T3, T4) = Nothing
+ getInt (one of T1 a; T2 a) = Just a
+ getInt (one of T3; T4) = Nothing
 
 This is a non-goal of this proposal: with binding pattern variables come challenges like binding existential constraints. Correctly specifying the semantics is hard and caused the parent proposal to become dormant after no progress has been made.
 
@@ -365,7 +370,7 @@ We think the following semantics in terms of view patterns is equivalent.
 We could define the semantics of Or patterns as a simple desugaring to view
 patterns. The desugaring rule is: ::
 
-    (one of p1, …, pk)
+    (one of p1; ...; pk)
     =
     ((\x -> case x of p1 -> True; p2 -> True; …; pk -> True; _ -> False)
         -> True)
--- a/proposals/0522-or-patterns.rst
+++ b/proposals/0522-or-patterns.rst
@@ -87,7 +87,7 @@ Or patterns extension adds one more production: ::
 
 ``one`` is a conditional keyword in patterns, but can still be used for variable patterns.
 
-Similar to ``case of``, ``of`` introduces a layout block and implicitly adds opening and closing curly braces when not explicitly given. Therefore, both of these are equivalent:
+As stated in the Haskell report and for consistency with ``case of``, ``of`` introduces a layout block and implicitly adds opening and closing curly braces when not explicitly given. Therefore, both of these are equivalent:
 
   - ``one of {1;2}``
   - ``one of 1;2``


simonpj:
Looks plausible to me.

JakobBruenker:
> An empty or-pattern will give a redundant pattern match warning.

I'm not sure yet if I'm actually in favor of what I'm suggesting here, but it seems interesting enough to mention:

It seems like there are some cases where it's not actually redundant, e.g.

```haskell
absurd :: forall a. Void -> a
absurd (one of {}) = ...
```

IIUC, it doesn't actually matter what we put on the right-hand side here, since it will never actually be run.

Thus, we could decide to make the right-hand side optional if there is an empty `one of` in one of the patterns, yielding

```haskell
absurd :: forall a. Void -> a
absurd (one of {})
```

It seems like this would essentially give us absurd patterns, as discussed [elsewhere in this repository](https://github.com/ghc-proposals/ghc-proposals/discussions/423), and perhaps this would be the most natural way to implement them.

JakobBruenker:
As an aside, personally I'd be weakly in favor of also allowing `(one of)` (without the `{}`), purely for consistency with `-XEmptyCase`.

knothed:
> As an aside, personally I'd be weakly in favor of also allowing `(one of)` (without the `{}`), purely for consistency with `-XEmptyCase`.

Yes, consistency with `-XEmptyCase` is a relevant goal, so I think I'm also weakly in favor, even if it looks weird.

knothed:
I would like to go this into review @nomeata

proposals/0522-or-patterns.rst
```diff
@@ -150,24 +152,25 @@ Informal semantics in the style of `Haskell 2010 chapter 3.17.2: Informal
 Semantics of Pattern Matching
 <https://www.haskell.org/onlinereport/haskell2010/haskellch3.html#x8-600003.17.2>`_:
 
-- Matching the pattern ``(one of p1, …, pk)`` against the value ``v`` is the result of matching ``v`` against ``p1`` if it is not a failure, or the result of
-  matching ``(one of p2, …, pk)`` against ``v`` otherwise.
-
-  ``p1``, …, ``pk`` bind no variables.
+- Matching the pattern ``one of {p1; ... ; pk}`` against the value ``v`` is the result of matching ``v`` against ``p1`` if it is not a failure, or the result of
+  matching ``(one of {p2; ... ; pk}`` against ``v`` otherwise. We require that ``p1``, …, ``pk`` bind no variables.
+- Matching the pattern ``one of {}`` against the value ``v`` fails after evaluating ``v``.
```
goldfirere, line 157: This is surprising. Why is this pattern strict?

goldfirere:
I'm curious to know more about the history of all this. That is, the suggestions you're making smell as if they come from practical experience. Is that the case? Is there a (partial, buggy, work-in-progress perhaps) implementation somewhere?

In the introductory post, there is a comment about how the old form could be misread as `(one of 1), 2`. That's just as true of the new form, because the braces are optional. I don't think this is so bad, exactly, but it's also not better.

Other questions (for anyone, not just @knothed):
- What's easier for tooling? The original proposal has no interaction with layout, but the new one does. But actually maybe the revision makes things easier, because it's consistent with the other use of `of`. I don't know, but I'd love to hear from e.g. HLS authors to see how this might affect them.
- Why were the parentheses previously required but now are deemed to be superfluous? The presence of `of` makes the new form unambiguous, but that doesn't mean it's easy to parse using standard parsing technology. (But my hunch is that thew new version is no harder than the old one.... but actually the old one seems non-trivial to me, too.)
- How much do we see using layout for this feature? My first reaction is that I'd rather have punctuation to help me read, but I'd be curious to hear from others about how much they'd use this, preferably with some motivating examples.

This PR has gotten rather little feedback from the community. Has it been well advertised? The ideas here are all cosmetic, and cosmetic proposals tend to actually get more participation than others. I would hate to accept this only to find a large group of people enthusiastic about the original proposal but lukewarm on these changes; this group might reasonably feel disenfranchised to have participated in the previous proposal only to have things changed after-the-fact.

int-index:
The work-in-progress implementation can be found at https://gitlab.haskell.org/ghc/ghc/-/merge_requests/9229

tek:
> What's easier for tooling? The original proposal has no interaction with layout, but the new one does. But actually maybe the revision makes things easier, because it's consistent with the other use of `of`. I don't know, but I'd love to hear from e.g. HLS authors to see how this might affect them.

FYI I plugged this into the tree-sitter grammar for the [amended](https://github.com/tree-sitter/tree-sitter-haskell/blob/or_patterns/test/corpus/pat.txt#L488) and [original](https://github.com/tree-sitter/tree-sitter-haskell/blob/or_patterns_comma/test/corpus/pat.txt#L488) variants.

AntC2:
> @goldfirere  This PR has gotten rather little feedback from the community. Has it been well advertised? The ideas here are all cosmetic, ...

(I was less than lukewarm for the original proposal. I spotted this amendment, and saw it was cosmetic, so just left folks to get on with it. If you want some feedback ...)

Seeing braces and especially `;` in the middle of a pattern looks weird. `;` to me (ever since ALGOL-60) means 'end of statement'. I get it that usually the braces can be omitted, but then `;` not surrounded by some sort of bracketing looks weirder. Using multilines doesn't seem less weird -- especially because (I guess) the patterns themselves will be brief, so there'll be lots of whitespace but carrying relatively minor semantic significance.

I think [@tomjaguarpaw's observations](https://github.com/ghc-proposals/ghc-proposals/pull/522#issuecomment-1187015331) still apply. Put all this ugly stuff inside pattern synonyms -- which are already plenty syntactically fussy -- and only there, so that mainline code can be kept elegant.

sgraf812:
FWIW, David and I came up with this proposal while refining the implementation at https://gitlab.haskell.org/ghc/ghc/-/merge_requests/9229. 

(2) seems quite uncontentious; clearly we want as few parens as possible.
But if you remove the parens in `((one of 1,2,3), 4)` (which we find ugly), you get an ambiguous `(one of 1,2,3, 4)`.
That motivates (1), in two parts: We work around the issue by wrapping the list in `{}`, e.g., `(one of {1,2,3},4)`.  `{}` is a natural choice because it isn't taken in this context. It looks a bit better than the one with parens, but is horribly un-Haskelly due to our usage of `,` as a separator. So we propose to go all the way and have `(one of {1;2;3}, 4)`.

So (1) enables (2): If we reject (1) (or some variant of it), we'll have to stick to parens. (@knothed should clarify if I'm wrong here; it has been quite a while.)

Whether or not we want `one of` to be affected by layout is a different matter, but it seems less surprising to me than leaving it unaffected in *this single instance of `of`*, whereas the linked section of the Haskell report clearly states 

> The layout (or “off-side”) rule takes effect whenever the open brace is omitted after the keyword where, let, do, or of.

Another case  of "least surprise" and anticipating easier tooling is the generalisation of nilary and unary or patterns. Imagine writing some code generator or bash script that constructs an Or pattern based on some input; you clearly want it to behave well for all inputs, so in the current form of #522 the author would need to build in special cases for 0 or 1 alternatives, e.g., (3).

IMO, `one of {}` should be strict because then `one of ps` is strict for any instantiation of `ps`. Plus, `case x of {}` is also strict. The potential use case for absurd patterns

```hs
absurd :: forall a. Void -> a
absurd (one of {})
```

is nice, too. On the other hand, it's not much use in discussing syntax for such a narrow use case (narrow, because GHC isn't a theorem prover).

proposals/0522-or-patterns.rst
```diff
@@ -1,12 +1,12 @@
-Or Patterns
+Or Patterns (Amended)
```
sgraf812, line 1:
```suggestion
Or Patterns
```

sgraf812: I'm not sure if the meta data needs to be changed like that. Perhaps the committee will know what to do here?

AntC2:
The place where this would really useful is in Pattern Synonym decls -- because we're restricted to a single 'top equation' in those. Is it possible for the community to 'dip a toe in the water' by implementing this initially only in Pattern Decls, rather than in patterns-at-large?

> @sgraf812 `{}` is a natural choice because it isn't taken in this context.

I think you have that backwards: you looked for some syntax that "wasn't taken in this context"[**], because the original proposal "we find ugly". What's now proposed looks to me equally ugly and not at all "natural" -- because `{  ;  }` to me is sequencing/`do` notation -- or possibly record syntax, not that I like that much. On which, how do folks like

```
foo (one of { MkBar{ x = 1, y = 'A' };
              MkBar{ x = 2, y = 'B' } } )  = ...
```

(And would that be the way to lay it out? And how small of a typo would make it utterly baffling to the compiler?)

> is horribly un-Haskelly 

I'm not getting why yous think the now-proposed is more "Haskelly".

[**] I fear Haskell is turning into one of those languages where every corner of possible syntax is hijacked to mean something-or-other. Such that program text is [indistinguishable from line noise](https://foldoc.org/line+noise).

knothed:
@AntC2

> ```
> foo (one of { MkBar{ x = 1, y = 'A' };
>               MkBar{ x = 2, y = 'B' } } )  = ...
> ```
> (And would that be the way to lay it out? And how small of a typo would make it utterly baffling to the compiler?)

No, due to the virtual braces and semicolons introduced by `of` you could write it as
```hs
foo (one of MkBar{ x = 1, y = 'A' }
            MkBar{ x = 2, y = 'B' } )  = …

```
which is what makes this new syntax so appealing in my eyes.

proposals/0522-or-patterns.rst
nomeata: Usually not needed. We can mention the amending in the line with the acceptance date

nomeata:
Since we are talking about Haskelly aesthetics: Somehow the multi line layout within a set of parentheses rubs me the wrong way. I usually expect layout syntax at the end of a expression (`where`, `$ do`, cases with their alternatives, maybe patten guards when the `=` is then preceded by a line break), or if it's enclosed, then by keywords (`let … in`). I might call that “large syntax”, which contains “small syntax”, stuff you typically put on one line, and where parentheses are natural.

An parentheses enclosed pattern to me is “small syntax”, where layout (to me) doesn't work that well.

Are there places where or patterns work well as large syntax (not enclosed by parentheses, nothing following them on the right). I guess with lambda case one can write
```
foo = \case
  one of Foo 1 2
         Bar 3 4
    -> possibly large rhs
  Baz 5 6
    -> another possibly large rhs
```
That looks similar (but more ugly, unfortunately) than code I have written in ocaml, so I guess it's useful?

(The direct translation of the ocaml syntax would be to allow multiple patterns without keywords, `foo = \case { Foo; Bar -> rhs; Baz -> rhs2 }`.)

It doesn't work so well in function clauses or `\cases` due to the extra parentheses, does it? If the only place where layout based or patterns can be used without surrounding parentheses is with `\case`, I'm somewhat luke warm.

sgraf812:
>  What's now proposed looks to me equally ugly and not at all "natural" 

Then this is not an argument against the proposal. The nice thing about layout blocks is that if you don't like them for aesthetic reasons, you can always write your program in a form that doesn't require them.

Honestly, we should stop getting hung up about the `one of {1;2;3}` use case and focus on `one of 1;2;3`. Yes, there are semicolons, but I don't see how we could get around some kind of separator; `;` seems like the most sane of the alternatives considered in #522.

> I'm not getting why yous think the now-proposed is more "Haskelly".

By Haskelly, I mean that the proposed syntax reuses existing concepts of the syntax. It has nothing to do with whether I like the syntax (while I of course try to make the subjective best out of the "building blocks" I've got) or whether I encourage exploiting all the syntactic options in every program. The latter is what a style guide is for.

To reiterate, `one of {1;2;3}` is more Haskelly, because it reuses existing intuition about `case of` and what you can and can't do with layout blocks. I think the contextual keyword `one` is rather un-Haskelly, but that is a different discussion that happened in #522.

Plus, `((one or 1,2,3),4)` has a typo and hence completely different, accepted semantics compared to the intended `((one of 1,2,3),4)`. `(one or 1;2;3,4) is an immediate parse error, but that is just another reason why I like the amended syntax more than #522 and has nothing to do with Haskelly-ness.

> Somehow the multi line layout within a set of parentheses rubs me the wrong way. 

If people feel that way, they don't *need* to use multiple lines. At any rate, if I'm browsing someone else's code base who likes multi-line syntax, I can still very easily *understand* that it's all an Or pattern.

> I might call that “large syntax”, which contains “small syntax”

I agree that there is a "feel" to this distinction, and I try to stick to this distinction by never introducing layout blocks in a pattern guard, for example:

```hs
f mb_x | Right y <- case mb_x of {
   Nothing -> Right 3
   Just z -> z }
  = y+42
```

Really ugly, but still Haskell (today), hence Haskelly. Since expressions are allowed in view patterns, you can have "large syntax" in patterns, today, too. So the "large" vs. "small" distinction is one of style, not one of Haskelly or not. 

> The direct translation of the ocaml syntax would be to allow multiple patterns without keywords, `foo = \case { Foo; Bar -> rhs; Baz -> rhs2 }`

That is nice to look at. If we used the syntax `(p1; p2)` to denote or patterns, we could have (in a future were we allow nested matches)
```hs
foo = \case
  (Foo 1 2; Bar 3 4)
    -> possibly large rhs
  Baz 5 6
    -> another possibly large rhs
```
Very close, and I quite like it, much more so than a contextual keyword `one`. We rejected `(p1; p2)` though, because `;` is hard to search for, making it hard for people unused to Or patterns to understand the code (and this was the only real reason, I believe). Plus, it is a bit un-Haskelly because `;` indicate that there's surrounding layout block (but not as un-Haskelly as a new contextual keyword `one`).

I don't really want to dwell on `(p1; p2)` syntax for long in this thread; it is easy to open a counter proposal and have a concrete discussion there. I would be very open to that.

---

To illustrate what abominations are possible today, serving as an example for why we shouldn't reject a syntactic extension based on just *one* kind of program it enables: We have [code like this](https://gitlab.haskell.org/ghc/ghc/-/blob/2c04024617f1ee4c76844cfe0a886bab87c23bd0/compiler/GHC/Iface/Load.hs#L444-572) in GHC's code base, which the majority of Haskellers wouldn't identify as a well-formed program, I believe. This code intermingles `case of` with and without layout blocks, do blocks with and without explicit layout, all nested within another. It can't be quite un-Haskelly because it *is* Haskell. It does indeed rub me the wrong way, too. I believe that code is written that way for a lack of exiting do blocks (#327), indicating that a *lack* of syntactic expressivity can lead to code that is largely perceived as breaking style guides.

jdkr:
> ```
> foo = \case
>   one of Foo 1 2
>          Bar 3 4
>     -> possibly large rhs
>   Baz 5 6
>     -> another possibly large rhs
> ```
I think this might cause some confusion. Although there is a line break after `case`, the order of words is `case one of`. One might think that `one` is the expression that is matched against the patterns.

Also, `one` could be a legitimate variable name if you think of `one` as a unit element.

AntC2:
> @sgraf812 Then this is not an argument against the proposal.

The rules of the proposal process is #522 is accepted by the Committee; so I'm not 'allowed' to argue against it. I already ref'd on this thread agreement with @tomjaguarpaw's position. OTOH this amendment represents really abandoning #522 and introducing a new proposal. (Since #522 is only syntactic sugar for a semantics I can already express.) That sort of sophistry from you is just going to antagonise the discussion.

> We have [code like this] in GHC's code base, which the majority of Haskellers wouldn't identify as a well-formed program,

Evidence please! Have you surveyed some Haskellers? Stop putting words in peoples' mouths. I find it perfectly Haskelly. (It's not pretty; but then it's dealing with a gnarly sequence of lookups into an ugly low-level world.) It's cleverly using `{ ; }` amongst the nested `do` blocks to indent the 'statements' to the same visual level. Excellent example of what to do where indenting just can't cope. I think you're being wildly optimistic that you can use layout per this proposal to avoid explicit `{  ;  }`. The fact of the matter (as Landin already conceded way back with ISWIM) is that layout only copes with toy examples. I frankly think you should withdraw this proposal and look at more serious example code rather than trying to amend in mid-development. To take a case in point:

>     foo = \case
>       one of Foo 1 2
>              Bar 3 4
>         -> possibly large rhs
>       Baz 5 6
>         -> another possibly large rhs

> @jdkr I think this might cause some confusion.

Yes that example looks terrible (and it's only small). I want to use layout or explicit `{ ; }` to line up the data constructors vertically. But I can't because the ones under `one of` have to be indented. What I'd be looking for is syntax more like a `C` `switch` statement with multiple `case`s (that is, for matching multiple data constructors) leading to the same `large rhs`. Like in this [Example](https://www.tutorialspoint.com/cprogramming/switch_statement_in_c.htm#:~:text=Flow%20Diagram-,Example,-Live%20Demo) with (in effect) `one of` `B`, `C`:

```
   switch(grade) {
      case 'A' :
         printf("Excellent!\n" );
         break;
      case 'B' :
      case 'C' :
         printf("Well done\n" );
         break;
      case 'D' :
         ...
```

[BTW I was coding to that style in BCPL before `C` became well known. That (and ALGOL) is where I get my intuitions about `;`.]

My eye isn't going to remember how far indented I was before the `-> possibly large rhs`. I _want_ an explicit `;` to herald this is a new case. And I probably want that `;` well towards the left so that I have space for my `large rhs`. In IOW I want exactly the style of that `Iface/Load`.

AntC2:
> @jdkr Although there is a line break after `case`, the order of words is `case one of`. One might think that `one` is the expression

Yeah, this code is confusing several intuitions. Firstly `\case` is a whole nother thing vs `case`. Second, because `\case` "starts a layout" and then the linebreak I think what you put _isn't_ the order of words. Rather

```
foo = \case {
  ; one of {
    ; Foo 1 2
    ; Bar 3 4
    }     -> possibly large rhs
  ;   Baz 5 6
          -> another possibly large rhs
  }
```

I'm using explicit `{  :  }` exactly so I can line up the data constructors to reflect the program logic. Layout just won't hack it.

sgraf812:
> Evidence please! Have you surveyed some Haskellers? Stop putting words in peoples' mouths.

Yes, well, apologies for having done that. My use of "the majority" was intended to be rather colloquial. My point with that example was not that it isn't Haskelly (how could it) or that using layout is not clever. I was arguing that it is Haskell code written today (that is subjectively ugly) and that *although* there is incentive to write code with explicit layout blocks, it is *vastly* uncommon. I've never seen it outside GHC's code base. Hence IMO it doesn't make sense to focus on how `one of { MkT{ ... }; MkT2 {...} }` is perceived and we should focus on `one of MkT{ ... }; MkT2 {...}` instead.

> OTOH this amendment represents really abandoning https://github.com/ghc-proposals/ghc-proposals/pull/522 and introducing a new proposal. (Since https://github.com/ghc-proposals/ghc-proposals/pull/522 is only syntactic sugar for a semantics I can already express.)

That... is a curious and destructive interpretation that I hope is not shared by the committee, because it would jeopardise any incremental improvement to GHC whatsoever.

> Although there is a line break after case, the order of words is case one of. One might think that one is the expression that is matched against the patterns.

Yes, that is a good point to be considered by the committee. `one` is not recognised as a keyword by a reader that is unaware of Or patterns and the parens in #522 would help to disambiguate. That could be rectified either by a careful author of the code who wraps `one of ...` in optional parens *in this one case* (-> style guide/formatter) or by the language in using a different marker using an established keyword such as `case` (switching from `one of` to `case of` is not part of this proposal, though), or abandoning the prefix marker idea (see the discussion starting at https://github.com/ghc-proposals/ghc-proposals/pull/522#issuecomment-1295506494 for that) altogether in favor of `(p1;p2)` of `(p1;;p2)` syntax (again, a different proposal). 

IMO, the style guide solution is quite low cost compared to incurring parens on every Or pattern ever. Perhaps the committee disagrees and we should revise syntax (or leave it as is), in which case we'd like to know whether one of the alternative proposals ideas (`case of`, `(p1;;p2)`) might bear potential (which will of course not be treated as a committment that the fleshed out proposal will be accepted).

tek:
tree-sitter with [or keyword](https://github.com/tree-sitter/tree-sitter-haskell/blob/or_patterns_or/test/corpus/pat.txt#L488)

jdkr:
Just saw the conflict with the [or](https://hackage.haskell.org/package/base-4.18.0.0/docs/Prelude.html#v:or) function from the Prelude.

lierdakil:
> conflict with the or function from the Prelude

Why should it matter? You can't use a function in patterns, so there's no ambiguity.

knothed:
> Why should it matter? You can't use a function in patterns, so there's no ambiguity.

True, but `fun (T1 or T2) = 0` could either denote an or-pattern or a pattern match on the binary constructor `T1`.

jdkr:
> Why should it matter? You can't use a function in patterns, so there's no ambiguity.

Not sure, maybe it can be used. I just saw that `case` and `of` are reserved and you can't name variables and functions like that. On the other hand, you can name variables `as` or `forall`, for example.

lierdakil:
> `fun (T1 or T2) = 0` could either denote an or-pattern or a pattern match

Indeed, that's why you need braces or parenthesis. I don't see it being a big issue, we need to use those in numerous other cases with patterns also. E.g. `T1{} or T2{}` or `(T1) or (T2)` are or-patterns, while `T1 or T2` is matching on binary `T1` and binding its first argument to `or`. There is no parsing ambiguity here, although this does admittedly introduce some visual ambiguity.

> Not sure, maybe it can be used

Haskell already has enough conditional keywords nowadays such that it shouldn't be a huge issue.

`case` and `of` are keywords in the expression scope, and while `of` is probably a historical accident, having a variable named `case` would introduce parsing ambiguity. `or` would only be a keyword on the lhs of a pattern match outside of any patterns, so AFAICT this shouldn't be a technical issue.

jdkr:
> True, but `fun (T1 or T2) = 0` could either denote an or-pattern or a pattern match on the binary constructor `T1`.

Why can't this be distinguished? The second pattern variable for the binary constructor `T1` would be `T2` in this case. But this is a constructor itself. Also constuctors are always written in upper case.

sgraf812:
I'm not for or against using `or` or `\/`, but that should be a separate proposal. I'd rather we return focus to this proposal, which IMO is an incremental change for the better (even if it turns out to be just a local maximum).

Note that `or` and `\/` lack the prefix signalling that was deemed desirable in https://github.com/ghc-proposals/ghc-proposals/pull/522#issuecomment-1295506494. In writing up a proposal to use `or`, `\/` or `;;` (say), it would be best to be aware of the whole discussion in #522 (I certainly am not) and summarise the pros and cons in an amended section 6.1 "Alternative syntax" so that casual or forgetful readers (such as me) don't need to sift through 100s of posts of discussion spread over multiple issues.

knothed:
> Why can't this be distinguished? The second pattern variable for the binary constructor `T1` would be `T2` in this case. But this is a constructor itself. Also constuctors are always written in upper case.

Well, it could be, considering the arity of the constructor `T1` - if it is `0`, we get an `or`-pattern, if it is `2`, we get a constructor pattern on `T1`.
This however strikes me as cruelsome, a syntactically ambiguous pattern which is only resolved after looking up information on the used constructor.

lierdakil:
> return focus to this proposal

Fair enough, apologies for side-tracking.

Points (1) and (2) make sense. I'm not sold on (3), but this seems a natural consequence of (1) and (2).

Here's a question: is this valid under the amended syntax?

```hs
f (one of) = ...
```

It seems to me that it should be, based on how the layout rule is applied elsewhere (e.g. in `case ... of`), but I'm not convinced this is an intended consequence, and I reckon it would look pretty mystifying to the uninitiated.

int-index:
> I reckon it would look pretty mystifying to the uninitiated.

It would be valid but I expect no one would write it, just like today something like this is valid:

```haskell
f x
  | x < 0 = ...
  | let   = ...
```

Because of the interaction between nullary `let` and pattern guards, `let` can fulfil the role of `otherwise` for a catch-all clause. It's shorter, too! But people keep using `otherwise` for readability, and I think they'd write out `f (one of {}) = ...` explicitly, too.

int-index:
As to the amendment in question, I think it's a good idea to make `of` a layout keyword regardless of context. This makes lexing/parsing more predictable, so we should do it even if combining or-patterns and layout would be a rare occurrence.

lierdakil:
> I expect no one would write it

I pretty much guarantee someone will write it, but your point about it not being worse than `let` in pattern guards is convincing enough.

sgraf812:
ping @goldfirere as it seems dust has settled on the discussion

proposals/0522-or-patterns.rst
```diff
@@ -76,45 +76,49 @@ Changes in the grammar
...
 Or patterns extension adds one more production: ::
 
-          |    (one of pat1, …, patk )                (Or pattern, k ≥ 2)
+          |    one of { pat1 ; ... ; patk }                (Or pattern, k ≥ 0)
 
 ``one`` is a conditional keyword in patterns, but can still be used for variable patterns.
-The ``,`` between the parentheses have (shift) priority that is lower than any other ``apat``'s (reduction) priority.
 
-Some examples that this new grammar produces: ::
+Similar to ``case of``, ``of`` introduces a layout block and implicitly adds opening and closing curly braces when not explicitly given. Therefore, both of these are equivalent:
```
sgraf812, line 90:
```suggestion
As stated in the Haskell report and for consistency with ``case of``, ``of`` introduces a layout block and implicitly adds opening and closing curly braces when not explicitly given. Therefore, both of these are equivalent:
```

numlit2:
Careful with new syntax that will be used a lot. If it's ugly it will be everywhere.
`{x;y;z}` is ugly _and_ noisy.

In real code where data constructors have fields this will mean `(one of {x{};y;z{}}, a)`
or even `(one of {x{}; one of {y;z{}}}, a)`

This is too much.
Strong -1 on this proposal.

Someone proposed a syntax with `or` like `(T1{} or T2)`. This is concise and elegant.

simonpj:
>  I'm a big -1 on this proposal because it does not specify how the community tools will be supported in adding support for this new syntax.

Thanks for responding.  Can you expalin your reservations a bit more explicitly?   I'd like to understand better.

This proposal is about chainging syntax for a new (but accepted) not-yet-implented proposal, so there should not be any libraries to change.   What support would `ormolu` need?  Indeed how would `ormolu` be affected?  Thanks!

googleson78:
> how the community tools will be supported in adding support for this new syntax

At least for the following tools, I don't think there's anything to do:
* `hlint` - already uses `ghc-lib-parser`, so nothing to do here, maybe add some useful hints down the line
* `ormolu`, `fourmolu`, `hindent` - already uses `ghc-lib-parser`, so the only thing left to do is decide how to format the new syntax, which I don't think there can be any support given here, as the authors need to decide for themselves how to do so

In general, for a while now (2-3 years?) the best practice has been to use `ghc-lib-parser`, so as to not run into parse errors arbitrarily, so most tools have been doing that and should not have issues with at least not erroring out on the new syntax.

(btw, sorry for my ignorance, but what is `hpc`?)

s-and-witch:
I'm strongly disagreed with any option that includes two or more keywords like `one of ...`. I want to write less code, not more, so I don't understand why I will have to write something like `please consider one of {5;6;7}` in patterns. `or` it better option, but I also don't understand why semicolon separated patterns were dropped from the discussion. 
```
case x of
  1; 2; 3; 4 -> ...
  _ -> ....
````
Short and without tons of useless syntactic salt.

P.S. I found this in the discussion above:
> We rejected (p1; p2) though, because ; is hard to search for, making it hard for people unused to Or patterns to understand the code (and this was the only real reason, I believe).

What is meant by "...it hard for people unused to Or patterns to understand..." ? Who is the consumer of this feature? Everyone who knows Haskell a little will google extensions from the top of the file if they will found unknown syntax. Can someone explain me a scenario in which this would become a problem?

**Comment by googleson78** [comment]
> Can someone explain me a scenario in which this would become a problem?

Not that I'm particularly concerned about this case exactly, but it happens when you see `;` in a code review/snippet/file that is using globally enabled extensions/file that has 20 extensions enabled. I have personally had some issues with this like seeing view patterns for the first time (how exactly are you supposed to search for that?), and this is similarly unsearchable, but I don't think it's that big of an issue, since you can usually stumble around for a bit and learn what it is, and you only have to do it one time.

> because ; is hard to search for

This is a much bigger concern for me, because `;` is very hard to grep for, but OTOH I cannot think of a scenario off the top of my head when I would need to do that.

What I would be particularly unhappy with is having another operator thing (like commas in lists or records) that I cannot use layouting rules for instead. In order to be able to also use a layouting rule, we would need to have some layout introducing keyword. Perhaps some middle ground might be allowing `one of` to be omitted when it's a single line thing, but requiring `one of` when you want to use the layouted multiline version of this feature. This is getting a bit confusing though..

**Comment by goldfirere** [comment]
I'm not sensing a clear direction from the conversation here, so let me put my own ideas down:

One of the joys of programming with or-patterns in OCaml is that the separator for case alternatives is also used as the or-pattern operator. And I think we can do this in Haskell, too! Make the or-operator `;`. (Yes, I know this isn't a new idea. But wait.) Here is an example:

```hs
case ... of
  A -> dealWithA
  B -> dealWithB
  C
  D
  E -> dealWithTheRest
```

The layout rule inserts `;`s, and so this Just Works. It means that you can just delete the RHS of a match and get "fall-through" behavior. Naturally, you can still use `;` explicitly: `f (A ; B) = ...` works fine. I'd prefer `|` there, but we can't have everything. (Though I have to say the "`|` is reserved for an unimplemented proposal from years ago" isn't all that great a reason to avoid it.)

One objection is that `;` looks asymmetrical. But actually this is accurate here: the left-ward pattern gets checked first, which might matter because of laziness. It's rather contrived, but `Just (); Just _` has different laziness behavior than `Just _; Just ()`.

This little counter-proposal doesn't introduce a way of dropping into layout in the middle of a pattern -- so perhaps we want to use semicolons and an optional `one of` herald, which serves both to alert the reader and drop into layout. In this way, my suggestion goes beyond the current proposal.

What do we think?

s-and-witch:
> What do we think?

I think that your proposal is the best option from the discussed above, but I want to clarify for me the part about nested patterns.

Am I right saying that `one of` in that case becomes only the new layout introduction helper and can be avoided in case I insert ; everywhere? 

Examples:

```haskell
f (Solo (4; 5)) = ...
```
But
```haskell
f (Solo (one of 
                 4
                 5)) = ...
```
Will that work? I will expect that since `case ... of A; B -> ...` works.

sgraf812:
> The layout rule inserts ;s, and so this Just Works.

I think I'd like it. It was first brought up in https://github.com/ghc-proposals/ghc-proposals/pull/522#issuecomment-1186646295 along with a draw-back: What if the user wrote a case expression where the pattern spans over multiple lines?

```hs
f x = case x of
  T (Really, Long, Line)
     (Also, ABit, Long) -> 32
```

Then layout will insert a `;` after the first line, making your proposal introduce a breaking syntactic change (and making it quite complicated to implement in the first place, given that we have to retain compatibility to Haskell2010 in the parser). So I don't currently see how inserting `;` via layout in this position is possible.

There's also the risk of confusion with [`\cases`](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0302-cases.rst), which also introduces layout. E.g., I'm having a hard time understanding what

```hs
f = \cases
  []
  (x:xs) -> 42
```
should do (perhaps this needs `-XNondecreasingIndentation` today). Does it define a partially-defined binary function on lists as today or a totally-defined unary function on lists with this proposal?

I'd like a language where Richard's proposal works (yay, OCaml), but unfortunately that language doesn't seem to be Haskell.

If you take away the layout-in-case idea, then the proposal boils down to using `(T1; T2)` or `(T1 | T2)` as separator (although that rules out nilary patterns), which I'm sympathetic to, but a lot of people felt very enthusiastic about the prefix operator `one of` at the time: https://github.com/ghc-proposals/ghc-proposals/pull/522#issuecomment-1295506494

If we could sort out that discussion first (although my impression was that it *had* been sorted out), I think we'd gain a lot.

In the meantime I think the OP is a perfectly good proposal. Perhaps it could be accepted on the condition that the implementation in GHC is just a "preview" and syntax might change again? So that we have enough time to sort out the syntax but also can play around with the feature in a GHC release rather than having to maintain an MR on the side.

googleson78:
> on the condition that the implementation in GHC is just a "preview" and syntax might change again

Not sure how realistic this is. From my observations, no matter how hard you scream at people that a behaviour is very likely to change, they will still be unhappy when it does, or might not even know that it's likely to change.

s-and-witch:
> What if the user wrote a case expression where the pattern spans over multiple lines? ... Then layout will insert a ; after the first line

Nope, it will not. It will insert one only in case of this code
```haskell
f x = case x of
  T (Really, Long, Line)
  (Also, ABit, Long) -> 32
```
Which is a parse error for today. 

> There's also the risk of confusion with [\cases](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0302-cases.rst) ... (perhaps this needs -XNondecreasingIndentation today)

Yes, that compiles only with `-XNondecreasingIndentation`, but AFAIK, this extension breaks syntax by its design

https://play.haskell.org/saved/QEUejqR8

> a lot of people felt very enthusiastic about the prefix operator

I hope that they felt very enthusiastic about the way forward in discussion, because I'm disappointed with the trend of introducing 2 and more word keywords. They look good at mixfix position, not in the prefix.

sgraf812:
Yes, I realised shortly after posting that my mental model for layout was broken. I actually think now that layout-in-case-pat is viable.

Still, I think this PR is the wrong place to have the discussion. I opened https://github.com/ghc-proposals/ghc-proposals/issues/587 instead, where we can collect arguments in the topmost post.

I invite everyone to give their vote/arguments.

simonpj:
> One of the joys of programming with or-patterns in OCaml is that the separator for case alternatives is also used as the or-pattern operator. And I think we can do this in Haskell, too! Make the or-operator ;. 

Your sole example was a case expresion with the or-pattern at the top.  But or-patterns are compositional and can occur nested.  So I think you are saying that I can write
```
f (Just (A ; B)) = ...
```
to match on `Just A` or `Just B`.  Correct?

Then I think you are saying that *in addition* we support a special syntax for case expressions in which a case alternative can be a single pattern all by itself, with no `-> rhs` part; and then that pattern is or'd into the next pattern.  Thus the alterantives 
```
   p1 -> rhs1 ;
   p2 ;
   p3 -> rhs2
```
means the same as
```
  p1 -> rhs1 ;
  (p2 ; p3) -> rhs2
```
Very well, but that's an orthogonal addition to the proposal.  It's fully compatible with `one of` syntax:
```
   p1 -> rhs1 ;
   p2 ;
   p3 -> rhs2
```
means the same as
```
  p1 -> rhs1 ;
  one of { p2 ; p3 } -> rhs2
```
So, by all means, an extra proposal.   But one that is rather independent of the current thread isn't it?

sgraf812:
@simonpj could you offer your pro-`one` opinion over at https://github.com/ghc-proposals/ghc-proposals/issues/587 and vote correspondingly?

Whether or not we want layout can be made a secondary vote.

goldfirere:
@simonpj 

Yes, I mean the syntax to be available nestedly. But I don't think the two ideas are orthogonal: if `;` is the operator, then it all just works, no? That is,

```hs
case x of
  A
  B -> ...
```

is the same stream of tokens as

```hs
case x of
  A; B -> ...
```

The nice thing about this all is that readers just have to learn the one layout rule about inserting semicolons, and then everything else follows.

tek:
@goldfirere since `of` starts a layout, and the or pattern another one immediately, do you think the parser will be able to deal with this without problems?

nomeata:
Should I close this in favor of #609?

PR #586, opened 2023-05-13
Split -Wunused-imports (under review)

googleson78:
[Rendered](https://github.com/googleson78/ghc-proposals/blob/split-unused-imports/proposals/0000-split-wunused-imports.md)

We propose to split the `-Wunused-imports` warning into two parts:
* warn on actual unused imports
* warn on duplicate imports

while also not including the second in `-Wall`. This will greatly reduce "churn" for library authors.

### DIFF
--- a/proposals/0000-split-wunused-imports.md
+++ b/proposals/0000-split-wunused-imports.md
@@ -57,11 +57,11 @@ import Z
 ```
 A "duplicate import" looks like this:
 ```haskell
-import X
-import X (f)
+import X -- X exports f
+import Y (f)
 ...f is used here...
 ```
-Here, one of the imports is currently marked as unused.
+Here, one of the imports is currently marked as unused. Note that `X` and `Y` could be the same module just as well.
 
 The proposal is for `-Wunused-imports` to be changed so that only the "unused import" case emits a warning.
 
--- a/proposals/0000-split-wunused-imports.md
+++ b/proposals/0000-split-wunused-imports.md
@@ -179,8 +179,6 @@ The main cost is changing the behaviour of a warning without notice, even if we 
 
 Is this acceptable? From initial feedback given in the proposal discussion, it seems that it is.
 
-However, for an extension of this proposal which includes some mitigation for this issue, see [the unresolved question section on Meta warnings](#meta-warning).
-
 ## Alternatives
 
 ### Relaxed redundant imports
@@ -198,23 +196,6 @@ if someone wants to have duplicate import warnings as per the "relaxed redundant
 potentially amending the new `-Wduplicate-imports` warning instead.
 
 ## Unresolved Questions
-
-### Meta warning
-
-One big drawback of the proposed changed is that, as is, the default behaviour will change silently - the user was not asked and has not agreed for this to change.
-
-Therefore, we propose that we introduce a "meta warning", let's say `-Wmeta-unused-imports` (names here are placeholders for now), with the following properties:
-
-0. Whenever a user has `-Wunused-imports` on, and they don't have `-Wno-meta-unused-imports` on, emit a warning that highlights the changed behaviour of `-Wunused-imports`.
-1. It is on by default.
-2. It **is not made into an error by `-Werror`**. Optionally there could be `-Wmeta-error` or something along those lines for users who want to get errors for potential future meta warnings.
-
-This `-Wmeta-unused-improts` warning is only present for a few release cycles, and is removed afterwards, or alternatively, turned off by default.
-
-There are two issues with this approach:
-* On older ghcs, meta warnings are not something that exists, hence if you silence the meta warning with `-Wno-meta-unused-imports`, older ghcs will start warning about an unrecognised warning flag
-* In general it seems like quite a sledgehammer to apply here. Meta warnings are a new concept for ghc  in two ways - in being a new type of warning, but also a warning that doesn't error under `-Werror`.
-    Initial feedback has indicated that this cost may not be worth it.
 
 ### Niche `-Weverything` breakage
 
--- a/proposals/0000-split-wunused-imports.md
+++ b/proposals/0000-split-wunused-imports.md
@@ -1,0 +1,230 @@
+---
+author: Georgi Lyubenov, Torsten Schmits
+date-accepted: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/586).
+
+# Split `-Wunused-imports`
+
+The current iteration of `-Wunused-imports` fulfills two arguably different purposes - warn on "unused imports", and warn on "**duplicate** imports", the latter of which has proven to be a heavy burden on the ecosystem.
+
+We're suggesting that we split off the "duplicate imports" part from `-Wunused-imports` into another warning, called `-Wduplicate-imports`, and crucially, **do not** include `-Wduplicate-imports` into `-Wall`.
+
+We additionally introduce a new **"meta" warning**, which will be used to loudly communicate, but without erroring, that the behaviour of `-Wunused-imports` has changed, so that users can adapt accordingly.
+
+## Motivation
+
+`-Wunused-imports` has long been a thorn in the side of people who want both:
+* to not have unused imports (in the sense described in the [proposed change specification section](#proposed-change-specification))
+* to compile warning-free across different versions of their dependencies
+
+The most often case where this comes up is with `Prelude` and other widely used modules in `base`.
+In particular, when `Prelude` gets a new reexport, which previously only lived in another module from `base`,
+users of `-Wall` suddenly get new warnings, causing not only noise, but crucially breakage for users of `-Wall -Werror`.
+
+This effectively reduces the usability of the `-Wunused-imports` warning,
+because it forces library authors (who in general are doing this work for free) who want to enforce code quality (removing actual unused imports and potentially even unused dependencies)
+to also have to do extra work, with questionable benefits.
+
+This not only a theoretic concern:
+
+* Additional work was required in the implementation of the [export `liftA2` proposal](https://github.com/haskell/core-libraries-committee/issues/50):
+    * effort in `ghc` itself to maintain warning-freeness
+    * additional non-trivial sections in [the migration guide](https://github.com/haskell/core-libraries-committee/blob/main/guides/export-lifta2-prelude.md)
+    * two additional non-trivial(requiring `CPP` or turning off the warning entirely) merge requests, to `Cabal-syntax` and `containers` respectively
+
+* The implementation of the AMP proposal required very similar additional work to the `liftA2` case above, see
+    * the concerns raised in https://www.yesodweb.com/blog/2016/05/are-unused-import-warnings-harmful
+    * the migration guide for ghc 7.10 - https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/7.10#ghc-says-the-import-of-is-redundant
+
+* This is listed as a "bad part" of Haskell - https://www.snoyman.com/blog/2020/11/haskell-bad-parts-2/
+    * in particular, the article mentions the re-export of `(<>)` from `Prelude` as requiring the exact same `CPP` workarounds as already mentioned
+
+* In the [sized integer types proposal](https://github.com/haskell/core-libraries-committee/issues/156), a critical counterpoint is introducing new emissions of `-Wunused-imports`
+
+* The [export `foldl'` proposal](https://github.com/haskell/core-libraries-committee/issues/167) seems like it would encounter the same implementation issues as the export `liftA2` proposal, and similar concerns to the sized integer types proposal
+
+## Proposed Change Specification
+
+An "unused import" looks like this:
+```haskell
+import X (f)
+import Z
+...neither f, nor anything from Z is used, hence they are both unused imports...
+```
+A "duplicate import" looks like this:
+```haskell
+import X
+import X (f)
+...f is used here...
+```
+Here, one of the imports is currently marked as unused.
+
+The proposal is for `-Wunused-imports` to be changed so that only the "unused import" case emits a warning.
+
+We also propose to introduce another warning, `-Wduplicate-imports`, which warns on only the "duplicate import" case.
+
+Finally, since the majority of uses of `-Wunused-imports` seem to come from `-Wall` uses, we propose to **not** include the new `-Wduplicate-imports` in `-Wall`, so that
+library authors can benefit without having to do yet more work.
+
+The wideness of applicability of `-Wduplicate-imports` can also be debated, hence it is unclear if it should be in `-Wall` in general. For example, it makes some sense for app writers, but not so much for library authors.
+
+Expressing the proposed change via the current ghc implementation([source](https://gitlab.haskell.org/ghc/ghc/-/wikis/commentary/compiler/unused-imports)):
+
+Current:
+
+0. `-Wunused-imports` reports
+    * warnUnusedModules: import M, where nothing is used from M
+    * warnUnusedImports: import M(f), where f is unused, and M doesn't fall under warnUnusedModules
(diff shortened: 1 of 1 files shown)


Bodigrim:
Strongly in favor, even without a meta warning.

Kleidukos:
Excellent initiative, love to see proposals to empower library maintainers!

proposals/0000-split-wunused-imports.md
```diff
@@ -0,0 +1,230 @@
...
+
+We're in the camp of "yes, it is", **if** we include the meta warning, since anyone who is maintaining code and cares about keeping things warning free should be able to
+easily notice the meta warning and turn on `-Wduplicate-imports` (or turn off the meta warning).
+
+There's also the cost of introducing the new concept of a "meta warning". These could arguably warrant their own standalone proposal, where they are more precisely specified.
+
+## Alternatives
+
+### Don't include the meta warning and let users deal with the silently changed behaviour.
+This seems potentially frustrating. Who knows if someone has some niche use case that they really like for the "duplicate imports" part of `-Wunused-imports`?
```
nomeata, line 204: Unless we have actual evidence that this change would be harmful, I'd be very reluctant to introduce such a new concept. Feels like vorauseilender Gehorsam.

proposals/0000-split-wunused-imports.md
```diff
@@ -0,0 +1,230 @@
...
+import Z
+...neither f, nor anything from Z is used, hence they are both unused imports...
+```
+A "duplicate import" looks like this:
+```haskell
+import X
+import X (f)
+...f is used here...
+```
+Here, one of the imports is currently marked as unused.
```
nomeata, line 64: The proposed spec change only talks about imports of the same module, but (based on the example).it should also apply when importing different modules that re-export things, right? I think a more precise specification wouldn't hurt.

Also, one can argue that one really only want to change this for _different_ modules, and still always warn about ”blatant duplication” of the same module import, which is neither good style, not does it actually help with the problem we want to solve.

Thinking this to it's conclusion might result in a design much like Ben's design at https://gitlab.haskell.org/ghc/ghc/-/wikis/commentary/compiler/relaxed-unused-imports.

nomeata (commented):
Thanks for starting this!

proposals/0000-split-wunused-imports.md
```diff
@@ -0,0 +1,230 @@
...
+There's also the cost of introducing the new concept of a "meta warning". These could arguably warrant their own standalone proposal, where they are more precisely specified.
+
+## Alternatives
+
+### Don't include the meta warning and let users deal with the silently changed behaviour.
+This seems potentially frustrating. Who knows if someone has some niche use case that they really like for the "duplicate imports" part of `-Wunused-imports`?
+
+### Relaxed redundant imports
+
+We could also instead implement the spec that's suggested in the ["relaxed redundant imports" proposal](https://gitlab.haskell.org/ghc/ghc/-/wikis/commentary/compiler/relaxed-unused-imports), however
```
nomeata, line 208: The wiki page may look scary, with GHC code fragments etc, but here is the gist of @bgamar's idea in one sentence:

> Then, for each use of an imported name, we will attribute
that use to all import-items which implicitly provide it, AND one import-item
which explicitly provides it. Then, any import items with no
uses attributed to them are unused, and are warned about.

I find this matches my gut intuition for how things should work quite well: if I import something explicit, then there should be only one mention of it (but additional implicit imports are fine). If there are implicit imports it shouldn't matter if there is more than one.

So I'd keep it on the table for now.

proposals/0000-split-wunused-imports.md
googleson78: If I'm correctly following what "vorauseilender Gehorsam" means, I also speak partly for myself here - while this change in particular is not something that I would mind, I am not a fan of things changing silently under my feet.

I agree that it's a bit heavyweight to introduce an entirely new "type" of output (albeit very similar to existing ones) from GHC, I also feel that it might have been helpful in the past or be helpful in the future in any similar "transition periods".

Alternatively, I remember some discussion about extending/changing warning categories, and perhaps even excluding some warning categories from being errors by default. It might make sense to turn this into a category - "meta warnings" - that are not errors with `-Werror`. That way, this can fit into pre-existing concepts. I can't remember where this discussion was being held though.

proposals/0000-split-wunused-imports.md
googleson78: I read through that section and mostly understood what it was trying to accomplish, but regardless, in my eyes, we should strive to keep this as straightforward as possible - the language itself is already brimming with complexity required just to get started, and I would much prefer (as a user) to not have to spend extra time to understand the guard rails of the guard rails :sweat_smile: 

I would like to reiterate:
> They also seem to be mostly orthogonal to me - if someone wants to have duplicate import warnings as per the "relaxed redundant imports" spec, then we could have another proposal after this one, potentially amending the new `-Wduplicate-imports` warning instead.

Even if the relaxed version is a straight up upgrade, I don't see why these "two" functionalities are under the same flag in the first place.

(As an anecdote, I was explaining what `-Wredundant-imports` does to a non-Haskell programmer, and even in the middle of the explanation, when I got to the "duplicate" part which this proposal is about splitting off, they interrupted to ask "but that's not a redundant import, is it?")

proposals/0000-split-wunused-imports.md
googleson78: Yep, you're right, I should extend the proposed specification some, thanks!

proposals/0000-split-wunused-imports.md
nomeata: That's a fair point, one can indeed argue that all imports (not just the first one) are “used” in some way.

proposals/0000-split-wunused-imports.md
adamgundry: Thanks for writing up this proposal! It's a good idea to tackle this issue.

> Alternatively, I remember some discussion about extending/changing warning categories, and perhaps even excluding some warning categories from being errors by default. It might make sense to turn this into a category - "meta warnings" - that are not errors with `-Werror`. That way, this can fit into pre-existing concepts. I can't remember where this discussion was being held though.

There was some discussion in #541 about being able to classify the default severity of custom warnings from `WARNING` pragmas, perhaps that's what you are thinking of? I don't recall any discussion of errors that would not become warnings with `-Werror`.

I'm also pretty sceptical of the value of `-Wmeta-unused-imports`. It will be a very noisy way to herald the change, seems difficult to disable (since `-Wno-meta-unused-imports` will give an "unrecognised warning flag" warning on earlier compilers), and introduces complexity to the definition of `-Werror`. Warnings can and do change between GHC releases, so a mention in the release notes seems like sufficient notice to me.

proposals/0000-split-wunused-imports.md
```diff
@@ -0,0 +1,230 @@
...
+
+Proposed:
+* with `-Wunused-imports` - warn that the `Foo` import is unused
+* with `-Wduplicate-imports` - nothing
+
+## Effect and Interactions
+
+Unsure what to fill in here, it seems that the [Proposed Change Specification](#proposed-change-specification) covers the effects of this change.
+
+## Costs and Drawbacks
```
adamgundry, line 190: An obscure backwards compatibility point: with this proposal, compiling a module with duplicate imports will fail under `-Werror -Weverything -Wno-unused-imports`, whereas previously it would have succeeded (since `-Wno-unused-imports` previously suppressed both).

We could avoid this by making `-Wunused-imports` into a group that includes both `-Wreally-unused-imports` (in `-Wall`) and `-Wduplicate-imports` (in `-Weverything`). This would also mean that users who explicitly ask for `-Wunused-imports` continue to get both.

Is this worth it? I'm not sure. Adding a group feels a bit fiddly for a comparatively rare edge case.

proposals/0000-split-wunused-imports.md
googleson78: I'm resolving this to keep noisiness to a minimum, please add another comment if you feel that the proposed change specification still needs work.

proposals/0000-split-wunused-imports.md
googleson78: > since -Wno-meta-unused-imports will give an "unrecognised warning flag" warning on earlier compilers

I completely forgot this part. I admit that it does indeed greatly reduce my enthusiasm for the meta warning.

phadej:
*Relaxed redundant warnings* was (kind of) implemented once, and it was bad in the industrial setting where I need(ed) to aggressively prune imports, and perfectly drop (direct) dependencies (using fat prelude). If that alternative is chosen, there should still be a `-Waggressive-redundant-imports` to get as much as possible help from compiler to prune actually redundant imports.

Also in application development, I'll probably always enable `-Wredundant-imports`. (Redundant) imports (especially open imports - common in application development) do affect incremental compilation, so having precise imports has "measureable" benefits.

googleson78:
I pushed an amendment downgrading meta warnings from a part of the proposal to an "Unresolved Question".

@nomeata, I have a question here - is it allowed for me to potentially leave this as an unresolved question and up to a committee vote itself?

nomeata:
Yes, that's fine. Probably the shepherd will then make a judgement call, and the committee will debate it. But if you have a preference, feel free to put it first - the committee anyways may ask for changes to one aspect or another.

cdornan:
Thanks @googleson78 for this excellent intiative! I have been asked to shepherd the proposal.

My observations:

- bundling `-Wduplicate-imports` with `-Wunused-imports` was a mistake that we should fix;
- ideally `-Wall` should include `-Wunused-imports` but not `-Wduplicate-imports`;
- as @phadej says we should be careful retain access to `-Wunused-imports`/`-Waggressive-redundant-imports` as it represents an important use case;
- I agree that taking `-Wduplicate-imports` out of `-Wall` is disruptive but I am tempted to propose it as I don't see it as seriously disruptive and the alternatives of either keeping it in `-Wall` or finessing the issue with a complicated construction could have more downside;
- if folks insist that disrupting `-Wall` is unaceptable then:
  * adding the Meta warning looks seriously disproportionate in introducing an entirely untested concept (how would you feel about deleting it from the draft altogether?);
  * @adamgundry's proposal of forming a new group is tempting, if we are going to do anything, as it clearly addresses the issue in a way that is well understood.

I have tried to condense the issues into a few points and quite probably have missed out something. Please shout if you think I have missed anything. 

If we can agree on this then I can probably move quite quickly to a recommendation. The key to that I think is keeping to the spirit of simplicity of the original proposal.

simonpj:
I'm generally in strong support of this proposal.  I have a copule of comments/questions but I do not feel strongly about how they are resolved.

Please add section numbers to the proposal.  This only takes one line -- and I think the template has that line.

* I'm pretty relaxed about changing the meaning of the existing warning.   If anyone thinks it'll be disruptive, I'd love to hear from them.  
* I would rather not implement a whole new meta-mechanism.  That'll open up a whole new debate, and slow down this proposal.
* I would love to tighten up the definitions of "unused" and "duplicate", which are currently defined only by example.  How about this:
  * An import is **unused** if the entitiy it brings into scope is not mentioned in the body of the module.
  * Two imports are **duplicate** if they bring into scope the same entities.
  I'm being a bit sloppy here... what is an "import".  We might mean an entire `import M( a, b, c )` declaration; or we might mean just one item like the `b` in the above declaration.  Anyway, precision is valuable.
* One substantive point: these two seem very different to me.  Example 1:
  ```
  import Foo( f )
  import Bar( f, h )
  ```
  and Example 2:
  ```
  import Foo
  import Bar( f, h )
  ```
  Silencing duplicate-import warnings from (2) is the goal of this proposal.  Is it *really* the goal to silence the warning that `f` is imported *explicitly* by both declarations in (1); and indeed that the `import Foo( f )" is entirely redunant? I have welcomed GHC telling me that an import declaration was entirely unused.

  Is there a middle path?  Maybe not, but I wanted to ask.

nomeata:
> Is there a middle path? Maybe not, but I wanted to ask.

Yes, I believe that would be Ben’s proposal, here referenced as “Relaxed redundant imports”

googleson78:
I've dropped any mention of the "meta warning" from the proposal, given the so far unanimous feedback that it is not necessary and/or overkill. 

## @cdornan 

> If that alternative is chosen, there should still be a -Waggressive-redundant-imports to get as much as possible help from compiler to prune actually redundant imports.

> as @phadej says we should be careful retain access to -Wunused-imports/-Waggressive-redundant-imports as it represents an important use case;

Since I'm not sure I understand you correctly, I want to reiterate to be 100% sure we're on the same page - the current version of this proposal says *nothing* about changing the behaviour of the warning(s) - it merely suggests to split the existing `-Wunused-imports` into two new ones, `-Wunused-imports` and `-Wduplicate-imports`, which when combined, result in the same behaviour that we currently have.

---

> @adamgundry's proposal of forming a new group is tempting, if we are going to do anything, as it clearly addresses the issue in a way that is well understood.

I'm not sure I follow here. From what I understand, @adamgundry's group proposal only addresses the issue of users who turn on `-Weverything` and then turn off extensions selectively after that. It won't address the original issue which motivates not including `-Wduplicate-imports` in `-Wall`, which is that a lot of library authors turn on `-Wall` by default, hence for them to actually benefit from this proposal, they would need to go and make a change to all of their libraries.

## @simonpj 

> Please add section numbers to the proposal.

Done

---

> I would love to tighten up the definitions of "unused" and "duplicate", which are currently defined only by example.

I'll try to get around to describing the proposed change in a more formal manner this weekend.

---

> Is it really the goal to silence the warning that f is imported explicitly by both declarations in (1); and indeed that the `import Foo( f )" is entirely redunant? I have welcomed GHC telling me that an import declaration was entirely unused.

> Yes, I believe that would be Ben’s proposal, here referenced as “Relaxed redundant imports”

While I do agree that your `Example 1` is not the the issue that this proposal targets, I'm somewhat convinced that implementing "Relaxed redundant imports" should be done independently from splitting `-Wredundant-imports`. See my thoughts [in the proposal section on this topic](https://github.com/googleson78/ghc-proposals/blob/split-unused-imports/proposals/0000-split-wunused-imports.md#relaxed-redundant-imports) as well as [in this comment response to Joachim](https://github.com/ghc-proposals/ghc-proposals/pull/586#discussion_r1193110063)

cdornan:
> Since I'm not sure I understand you correctly, I want to reiterate to be 100% sure we're on the same page - the current version of this proposal says nothing about changing the behaviour of the warning(s) - it merely suggests to split the existing -Wunused-imports into two new ones, -Wunused-imports and -Wduplicate-imports, which when combined, result in the same behaviour that we currently have.

I understand that. Like Oleg, I was just capturing my position should we start making adjustments like Simon is proposing.

> I'm not sure I follow here. From what I understand, @adamgundry's group proposal only addresses the issue of users who turn on -Weverything and then turn off extensions selectively after that. It won't address the original issue which motivates not including -Wduplicate-imports in -Wall, which is that a lot of library authors turn on -Wall by default, hence for them to actually benefit from this proposal, they would need to go and make a change to all of their libraries.

I am sorry for the confusion. My point was ageneral one, that _if we insist on not changing the behaviour of `-Wall`_ (which is not my preference), then we will need to provide good tools for folks to easily achieve the same outcome, which is I think what Adam was driving at. I am really merely saying that I am in favour of forming a new group if we think it will help developers achieve precise control over which unused duplicate warnings are repoted (especially when amending existing scripts and configuration files). But if we conclude that that kind of control is not needed then that would be even better.

cdornan:
I am thinking this proposal is good to go and would like to recommend it to the committee for acceptance.

nomeata:
@cdornan, remember that the shepherd should announce their recommendation to the committee on the mailing list (it is not the case that all committee members follow all github discussion threads).

cdornan:
@nomeata of course -- I wanted to give this thread a heads up before announcing my recommendation on this list, which I will do soon (I got held up by other things on my end).

Thanks as always for the reminder! 👍

nomeata:
Ok, I didn't want to rush anyone!

PR #590, opened 2023-06-06
Class Scoped Functions

VitWW:
This proposal  adds Class Scoped Functions as a tool for renaming class methods names.

[Rendered](https://github.com/VitWW/ghc-proposals/blob/class-scoped-functions/proposals/0000-class-scoped-functions.rst)

### DIFF
--- a/proposals/0000-class-scoped-functions.rst
+++ b/proposals/0000-class-scoped-functions.rst
@@ -1,0 +1,171 @@
+Class Scoped Functions
+====================
+
+.. author:: Viktor WW
+.. date-accepted::
+.. ticket-url:: 
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+.. sectnum::
+.. contents::
+
+GHC has lack of support for renaming of class methods if class is already public.
+This proposal gives a way how to do this painlessly.
+
+Background
+----------
+
+GHC has lack of support for renaming of class methods if class is already public.
+
+Here is an real example of a ``Monoid a`` class in "base" module::
+
+  class Semigroup a => Monoid a where
+
+      -- __NOTE__: This method is redundant and has the default
+      -- implementation @'mappend' = ('<>')@ since /base-4.11.0.0/.
+      -- Should it be implemented manually, since 'mappend' is a synonym for
+      -- ('<>'), it is expected that the two functions are defined the same
+      -- way. In a future GHC release 'mappend' will be removed from 'Monoid'.
+      mappend :: a -> a -> a
+      mappend = (<>)
+
+      mempty :: a
+      mempty = mconcat []
+
+      mconcat :: [a] -> a
+      mconcat = foldr mappend mempty
+      
+      {-# MINIMAL mempty | mconcat #-}
+
+Motivation
+----------
+
+Once class become public and renaming or even removing methods becomes practically impossible 
+without breaking backward compatibility.
+
+Let we have::
+
+      -- Bar.hs:
+      module Bar where
+        
+            class Bar a where
+                  foo :: a -> a
+
+
+      -- Main.hs:
+      module Main where
+            import Bar (Bar(foo)) -- explicit
+            import Bar (Bar(..))  -- implicit
+
+            instance Bar T where
+                  foo = somefunc
+
+
+And we decided to rename ``foo`` into ``bar``. Is it possible? ::
+
+      class Bar a where
+            bar :: a -> a
+
+      foo :: Bar a => a -> a
+      foo = bar
+
+But what to do with all instances and imports? These changes fully broke backward compatibility.
+
+
+Proposed Change Specification
+-----------------------------
+
+Syntax
+~~~~~~
+
+We add a keyword, which is important in class declaration only.
+
+It is a Class Scoped function, so it could be named as ``function`` (PHP/JS/Lua-like), ``func`` (Go-like), 
+``fun`` (OCaml/F#/SML-like), ``fn`` (Rust-like), ``defun`` (Lisp-like), ``define`` (Scheme-like), ``def`` (Phyton/Ruby-like),
+``lambda`` (Lisp/Python-like),  ``sub`` (Perl-like) ::
+
+    class Bar a where
+
+        function foo
+        foo :: a-> a
+        foo = bar
+
+        bar :: a -> a
+        bar = foo
+
+        {-# MINIMAL foo | bar #-}
+
+    foo :: Bar a => a -> a
+    foo = bar
+
+    -- Main.hs:
+    import Bar (Bar(foo)) -- explicit
+    import Bar (Bar(..))  -- implicit
+
+    instance Bar T where
+        foo = somefunc
+
+
+Semantics
+~~~~~~~~~
+
+Class Scoped function (CSF) has simple rules:
+
+* CSF could be defined in where-part of classes and instances only
+* CSF has a scope of an instance(where it is defined),  a class (where it is defined) and it's instances
+* CSF has no scope outside of class and instances 
+* CSF is always shadows outside function with same name
+
+Now we can rewrite ``Monoid a`` class as follows::
+
+    class Semigroup a => Monoid a where
+
+        function mappend
+        mappend :: a -> a -> a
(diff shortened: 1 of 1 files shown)


alt-romes:
I think it could be a bit clearer, but if I understand correctly what you mean is: there is no way for a library developer to change the name of a class function, while providing a backwards compatible replacement, without breaking existing code because of explicit import lists.

That is:

* The library defines
```haskell
module Library where
class LibraryThing a where
     someFunc :: a -> ...
```
* The user depends on this class and on its methods, but **not on the rest of the module**
```haskell
import Library(LibraryThing(..))
use = use someFunc somehow
```
* The library author wants to change the name of the class function `someFunc` to `newFunc`, but in order to preserve backwards compatibility they also provide `someFunc` as a standalone function
* The code of the user will break when upgrading because they are not importing `someFunc` in their explicit import list.

In my opinion, the breakage on the import list doesn't justify the cost (both in language and implementation complexity) of adding a keyword and implementing this logic, and that the solution here would be a deprecation cycle and require the user to update its import list accordingly.

VitWW:
@alt-romes Thanks for feedback.

> If I understand correctly what you mean is: there is no way for a library developer to change the name of a class function, while providing a backwards compatible replacement, without breaking existing code because of explicit import lists.

Exactly!

> The user depends on this class and on its methods, but not on the rest of the module

This problem exists even if user import everything, but it become harder to solve if user use detailed import

> The library author changes the name of the class function `someFunc` to `newFunc`, but in order to preserve backwards compatibility they also provide `someFunc` as a standalone function.

Do you mean this ?:
```haskell
class Bar a where
      bar :: a -> a

foo :: Bar a => a -> a
foo = bar
```

It  is good solution **if only** user did't write nonempty instance:
```haskell
instance Bar T where
        foo = somefunc
```
What do you suggest to do in this case?!

proposals/0000-class-scoped-functions.rst
```diff
@@ -0,0 +1,176 @@
...
+-----------------------------
+
+Syntax
+~~~~~~
+
+We add a keyword, which is important in class declaration only.
+
+It is a Class Scoped function, so it could be named as ``function`` (PHP/JS/Lua-like), ``func`` (Go-like), 
+``fun`` (OCaml/F#/SML-like), ``fn`` (Rust-like), ``defun`` (Lisp-like), ``define`` (Scheme-like), ``def`` (Phyton/Ruby-like),
+``lambda`` (Lisp/Python-like),  ``sub`` (Perl-like) ::
```
konsumlamm, line 88:
```suggestion
It denotes a class scoped function, so it could be named ``function`` (PHP/JS/Lua-like), ``func`` (Go-like), 
``fun`` (OCaml/F#/SML-like), ``fn`` (Rust-like), ``defun`` (Lisp-like), ``define`` (Scheme-like), ``def`` (Python/Ruby-like),
``lambda`` (Lisp/Python-like),  ``sub`` (Perl-like) ::
```

proposals/0000-class-scoped-functions.rst
```diff
@@ -0,0 +1,176 @@
...
+
+    -- (A) case
+    instance Bar T where
+        foo = somefunc
+
+
+Semantics
+~~~~~~~~~
+
+Class Scoped function (CSF for (A) case) has simple rules:
```
konsumlamm, line 117:
```suggestion
Class scoped functions (CSF for (A) case) have simple rules:
```

proposals/0000-class-scoped-functions.rst
```diff
@@ -0,0 +1,176 @@
...
+
+* CSF could be defined in where-part of classes and instances only
+* CSF has a scope of an instance(where it is defined),  a class (where it is defined) and it's instances
+* CSF has no scope outside of class and instances 
+* CSF is always shadows outside function with same name
```
konsumlamm, line 122:
```suggestion
* CSFs can only be defined in classes and instances
* CSFs are only in scope in class and instance definitions
* CSFs always shadow outside functions with the same name
```

proposals/0000-class-scoped-functions.rst
```diff
@@ -0,0 +1,176 @@
...
+
+Now we can rewrite ``Monoid a`` class as follows::
```
konsumlamm, line 124:
```suggestion
Now we can rewrite the ``Monoid`` class as follows::
```

proposals/0000-class-scoped-functions.rst
```diff
@@ -0,0 +1,176 @@
...
+        {-# MINIMAL mempty | mconcat #-}
+
+    mappend :: Monoid a => a -> a -> a
+    mappend = (<>)
+
+
+Unfortunately, these changes require changes for detailed import ((B) case).
+
+So we need to have implicit extension "``ImportFromClassFunction``" for omitting these changes in import and we 
+need to explicit switch it off by "``NoImportFromClassFunction``" language pragma.
```
konsumlamm, line 148:
```suggestion
So we need to have the implicit extension "``ImportFromClassFunction``" for omitting these changes in import and we 
need to explicit switch it off by a "``NoImportFromClassFunction``" language pragma. 
```

konsumlamm:
What is the advantage of this over just adding a new method with default implementations? In fact, your proposal requires adding a new method with a default implementation and I don't see what we gain by duplicating it as a function and adding a "class scoped function" annotation.

I also find it weird to have a function keyword for this, when normal functions don't have a keyword.

VitWW:
@konsumlamm Thanks a lot!!

> I also find it weird to have a function keyword for this, when normal functions don't have a keyword.

We could chose a word with different meaning, for example `local` or `inner`
```haskell
class Semigroup a => Monoid a where

    inner mappend
```

> What is the advantage of this over just adding a new method with default implementations

Mainly it is useful tool for **backward compatible** refactoring class methods. 
We could freely rename methods, change argument order, amount of arguments, method signatures and even fully remove methods.

And it also would be useful in my (not yet published) Proposal "Hybrid class-instances and instance-classes".

konsumlamm:
> > What is the advantage of this over just adding a new method with default implementations
> 
> Mainly it is useful tool for **backward compatible** refactoring class methods. We could freely rename methods, change argument order, amount of arguments, method signatures and even fully remove methods.

Can you make an example for changing argument order or the amount of arguments? From your examples in the proposal, it looks like it does the same as adding a new method with a default definition.

VitWW:
> From your examples in the proposal, it looks like it does the same as adding a new method with a default definition.

Since this proposal is driven by backward and forward compatibility, we wish users could define in instances with old method only (backward compatibility) and with new method only (forward compatibility). That means default methods must be for both old and new method.
But if we don't care about compatibility, a default method is an option.

> Can you make an example for changing argument order or the amount of arguments?

1) Renaming
```haskell
class Foo a where

   function foo_old

   foo_old :: a -> a
   foo_old = foo_new

   foo_new :: a -> a
   foo_new = foo_old
```

2) Swap the order of arguments
```haskell
class Bar a where
   type Collect a

   function elem_old

   elem_old :: a -> Collect a -> Bool
   elem_old = flip elem_new

   elem_new :: Collect a -> a -> Bool
   elem_new = flip elem_old
```

3) Change amount of arguments
```haskell
class Bar a where
   type Collect a

   function nextN_old

   nextN_old :: Collect a -> Int -> (Collect a, Maybe a)
   nextN_old c m = go (c, Nothing) m
       where
       go r n = case n of
           | n <= 0    => r
           | otherwise => go (next_new $ fst r) (n - 1)

   next_new :: Collect a -> (Collect a, Maybe a)
   next_new c = nextN_old c 1
```

4) Remove method
```haskell
class Semigroup a => Monoid a where

   function mappend

   mappend :: a -> a -> a
   mappend = (<>)

   mempty :: a
   mempty = mconcat []

   mconcat :: [a] -> a
   mconcat = foldr mappend mempty
```

chreekat:
Can't you already literally do everything you've done in your examples? How does `function mappend` change anything? 

Oh, I guess it removes it from the documented list of class methods?

VitWW:
@chreekat Good point!
You are right.
The most benefits this feature get if we wish to get rid of class `Monoid` at all, if we wish `Monoid` to be an emergent property from `Semigroup` and `EmptyCat` classes.

But these require my (not yet published) Proposal "Hybrid class-instances and instance-classes". I hope to publish it  this weak.

PR #591, opened 2023-06-07
Class Instances of Injective Synonyms

VitWW:
This proposal shows why and how to add Class Instances of Injective Synonyms.

[Rendered](https://github.com/VitWW/ghc-proposals/blob/class-instances-of-synonyms/proposals/0000-class-instances-of-synonyms.rst)

### DIFF
--- a/proposals/0000-class-instances-of-synonyms.rst
+++ b/proposals/0000-class-instances-of-synonyms.rst
@@ -1,0 +1,138 @@
+Class Instances of Injective Synonyms 
+====================
+
+.. author:: Viktor WW
+.. date-accepted::
+.. ticket-url:: 
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+.. sectnum::
+.. contents::
+
+GHC Type Families has lack of class instances.
+
+This proposal gives reason why changing this is important.
+
+
+Background
+----------
+
+"Main reason" why class instances of Type Families are forbidden is written in the GHC documentation:
+
+`6.4.9.6. Type families and instance declarations <https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/type_families.html#type-families-and-instance-declarations>`_
+
+*Type synonym families may not appear (at all) in an instance head*
+
+*The reason for the latter restriction is that there is no way to check for instance matching. Consider* ::
+
+    type family F a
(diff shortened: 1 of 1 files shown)


googleson78:
Have you looked at this proposal - https://github.com/ghc-proposals/ghc-proposals/pull/218? It seems to also satisfy your current provided motivation.

VitWW:
@googleson78  Thanks for feedback!

RFC 218 (ApplyingVia) is good, I like it. But 1) It is not accepted 2) It use newtypes 3) It is too magical 

My proposal use Injective Type Families, it is much simpler in implementation than ApplyingVia and has no magic. And it does not contradict ApplyingVia and could coexist with that proposal

adamgundry:
Thanks for the proposal. Unfortunately I'm finding it hard to understand exactly what is being proposed. What does the prefix double quote mean here? Looking at this definition:
```hs
type family Distr (a :: SP) b = r | r -> b where
    Distr a b = b

(<+>) :: Semigroup (''Distr Sum a)  => Distr Sum a -> Distr Sum a -> Distr Sum a
(<+>) = (<>)
```
typechecking `(<+>)` will end up reducing `Distr Sum a` to `a` and then trying to solve a `Semigroup a` constraint; can we do that from the given constraint `Semigroup (''Distr Sum a)`? Does `''Distr Sum a` compute, or is it more like a newtype?

VitWW:
@adamgundry Thanks for feedback!

>  then trying to solve a Semigroup a constraint; can we do that from the given constraint Semigroup (''Distr Sum a)? Does ''Distr Sum a compute, or is it more like a newtype?

It is like a newtype, but without a newtype. It's more like  free "overloaded" types:
```haskell
a = (5 :: Distr Sum Int) <+> (4 :: Distr Sum Int)
-- a == (9 :: Distr Sum Int)
-- a == (9 :: Int)
```

VitWW:
@adamgundry, I think that maybe "Newtypes from Injective Synonyms" is more general and much better idea.
`''Distr Sum Int` is a newtype of `Int`
```haskell
(<+>) :: Semigroup (''Distr Sum a)  => ''Distr Sum a -> ''Distr Sum a -> ''Distr Sum a

a = (5 :: ''Distr Sum Int) <+> (4 :: ''Distr Sum Int)
```

VitWW:
It looks like this proposal is not efficient enough as using newtypes for same purposes. 
So, I close this RFC for a revision

Thanks everyone for feedback!

PR #592, opened 2023-06-08
Scoped kind variables in standalone kind signatures

int-index:
#448 defines `-XExtendedForAllScope` that controls the scoping behavior in type signatures for term-level definitions. (Before #448, this was controlled by `-XScopedTypeVariables`).

Now, it's somewhat surprising to me that #54 deliberately opts out of this feature. @simonpj my apologies that when we discussed this today, I said that the proposal didn't specify this. Apparently it does specify this, but we also have this on the [GHC Wiki](https://gitlab.haskell.org/ghc/ghc/-/wikis/ghc-kinds/kind-inference/tlks#scoped-kind-variables):

> ```haskell
> type T :: forall k. k-> Type
> data T a = MkT (Proxy (a :: k))
> ```
> Question: Do we allow the use of `k` in `a :: k` here, even though it is not mentioned in the declaration header?

> Vlad and RAE: yes, if `-XScopedTypeVariables` are enabled then the TLKS brings `k` into scope. (Despite the proposal saying otherwise, it needs to be amended).

So here's the amendment.

### DIFF
--- a/proposals/0054-kind-signatures.rst
+++ b/proposals/0054-kind-signatures.rst
@@ -76,18 +76,111 @@ Proposed Change Specification
 
    Associated types may not be given standalone kind signatures. (See "Costs and Drawbacks" for discussion.)
 
-   Unlike type signatures, the type variables brought into scope in a type-level kind
-   signature do *not* scope over the type definition.
-
    Standalone kind signatures are enabled with the extension ``-XStandaloneKindSignatures``.
 
-2. Introduce a new extension ``-XCUSKs``, on by default, that detects CUSKs as they
+2. With ``-XExtendedForAllScope``, type variables ``tvs`` bound in the
+   outermost ``forall tvs.`` in a standalone kind signature also scope over the
+   type declaration, as if bound by ``@``-binders: ::
+
+     type T1 :: forall k. k -> Type
+     data T1    a = MkT (Proxy (a :: k))   -- equivalent to...
+
+     type T2 :: forall k. k -> Type
+     data T2 @k a = MkT (Proxy (a :: k))
+     --      ^^
+     --      implicit @-binder arising from the use of ExtendedForAllScope
+
+   We say that ``-XExtendedForAllScope`` inserts *implicit ``@``-binders* at
+   the beginning of the type declaration header.
+
+   * **Aliasing**. If there are explicit, user-written ``@``-binders at the
+     beginning of a type declaration header (i.e. where implicit ``@``-binders
+     would have been inserted), aliasing takes place.
+
+     Given a sequence of explicit binders ``[@j1, @j2, ...]`` and a sequence of
+     implicit binders ``[@k1, @k2, ...]``, we do **not** concatenate them to
+     produce ``[@j1, @j2, ..., @k1, @k2, ...]``. Rather, we overlay (or zip)
+     them, producing pairs ``[(@j1, @k1), (@j2, @k2), ...]``, and then each
+     pair acts as a single binder, where one name is an alias for another::
+
+       type T3 :: forall k1 k2. k1 -> k2 -> Type
+       data T3 @j1 @j2 a = ...  -- `k1` is an alias for `j1`, and `k2` for `j2`
+       --      ^^^ ^^^
+       --      user-written @-binders
+
+     When there are too few explicit ``@``-binders, only the overlapping prefix
+     is aliased, while the remaining variables are bound by implicit
+     ``@``-binders::
+
+       type Mixed :: forall k1 k2 k3 k4. k1 -> k2 -> k3 -> k4 -> Type
+       data Mixed @j1 @j2 a = ...
+          -- `k1` is in scope as an alias for `j1`
+          -- `k2` is in scope as an alias for `j2`
+          -- `k3` is in scope via an implicit @-binder
+          -- `k4` is in scope via an implicit @-binder
+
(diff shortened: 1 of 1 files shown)


nomeata:
So with `-XNoExtendedForAllScope` one would write 

```
type T :: forall k. k-> Type
data T @k a = MkT (Proxy (a :: k))
```
Just like in terms, right?

int-index:
Yes, `@k`-binders will be available in GHC 9.8

nomeata:
Does this need a full proposal process, or is it a straight forward bugfix/clarification that we can simply merge?

int-index:
It's a major change to scoping rules, I think we should put it through the process.

proposals/0054-kind-signatures.rst
```diff
@@ -76,8 +76,8 @@ Proposed Change Specification
 
    Associated types may not be given standalone kind signatures. (See "Costs and Drawbacks" for discussion.)
 
-   Unlike type signatures, the type variables brought into scope in a type-level kind
-   signature do *not* scope over the type definition.
+   By analogy with type signatures, ``-XExtendedForAllScope`` determines whether the type variables
```
simonpj, line 79: Please give a couple of examples

simonpj:
Ryan's question made me want to be explicit about shadowing.  

With `-XExtendedForallScope`: in a term we can write
```
f :: forall a b c. blah
f @b @c @e = ...body...
```
The in `body`:
* `a` and `v` are two names for f's first type parameter
* `c` is the name for f's second type parameter (the `b` from `forall b` is shadowed)
* `e` is the name for f's third type parameter (the `c` from `forall c` is shadowed)

Right?  And all the same things will happen for explicit type argument in data type decls
```
T :: forall a b c. blah
data T @a @b @c x y z = MkT (....body...)
```
I think it could be helpful to give examples like this in the (amended) proposal text.   No surprises here, but explicit is good.

int-index:
The more I think about it, the less sure I am how this should work. There are some clear cut cases:

* H98 data/newtype declarations (both LHS and RHS):
   ```haskell
   type D :: forall k. k -> Type
   --              ^^ binding site
   data D (a :: k) = MkD (Proxy (a :: k))
   --          ^^ use site           ^^ use site
   ```

* Type synonyms (both LHS and RHS):
   ```haskell
   type S :: forall k. k -> Type
   --              ^^ binding site
   type S (a :: k) = Proxy (a :: k)
   --          ^^ use site      ^^ use site
   ```

* Class declarations (both LHS and RHS):
   ```haskell
   class C :: forall k. k -> Constraint
   --               ^^ binding site
   class C (a :: k) where
   --           ^^ use site
     f :: Proxy (a :: k) -> Proxy k
   --                ^^ use site  ^^ use site
   ```

But what about:

* GADTs. We could scope over the header, but what about constructor signatures?
* Open type/data families. I'd assume we want to scope over the declaration (which consists solely of a header), but what about their instances? Those can be in a different module!
* Closed type families. Again, we could probably scope over the header, but what about the instances?

int-index:
@simonpj I made the proposed amendment more precise by phrasing it in terms of implicit `@`-binders, and I also added examples for various corner cases. Could you take a look?

simonpj:
See #23884 for a motivating example where the authors expected the standalone kind signature to scope over the declaration.

simonpj:
@int-index it'd be good to progress this

int-index:
@simonpj See last comment. It's blocked only on your feedback.

simonpj:
Ah.  I had not realised that I failed to press submit.  Sorry

int-index (commented):
Thank you for the review @simonpj. All good points, I've pushed an update to address them.

simonpj:
OK good with me.

> As to the removal of CUSKs, it's mostly held back by the amount of breakage it would cause.

Are we even deprecating CUSKs now?  I think we should

int-index:
@adamgundry Let's submit to the committee.

int-index:
> Are we even deprecating CUSKs now? I think we should

It's just that disabling them by default would break a lot of code, it is significant. The right thing to do is to implement a warning first.

The last bit of progress I've made was to adapt the GHC test suite to the upcoming deprecation, see this commit

```
commit 679bbc9705371af8ca65a5bfeb702e14bc406f51
Author: Vladislav Zavialov <vlad.z.4096@gmail.com>
Date:   Thu Jun 8 12:22:16 2023 +0200

    testsuite: Do not require CUSKs
    
    Numerous tests make use of CUSKs (complete user-supplied kinds),
    a legacy feature scheduled for deprecation. In order to proceed
    with the said deprecation, the tests have been updated to use SAKS
    instead (standalone kind signatures).
    
    This also allows us to remove the Haskell2010 language pragmas that
    were added in 115cd3c85a8 to work around the lack of CUSKs in GHC2021.
```

Someone motivated to deprecate CUSKs could take up where I left off.

simonpj:
> Someone motivated to deprecate CUSKs could take up where I left off.

The first step is just to add a warning, on by default, to say that they are going away.  Do we have a ticket for this?

int-index:
> The first step is just to add a warning, on by default, to say that they are going away.

Yes

>  Do we have a ticket for this?

Not that I know of.

gridaphobe:
Thank you @int-index, this amendment is accepted!

int-index:
I created https://gitlab.haskell.org/ghc/ghc/-/issues/25656 to track the implementation of this amendment.

PR #595, opened 2023-06-15
Amend #134 (deprecated exports) in light of implementation

sheaf:
The implementation of the deprecated exports proposal in [GHC MR !10283](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/10283) has surfaced several subtleties that were not explicitly considered in the original proposal #134, which this amendment aims to clarify. Chiefly:

  - We should be more specific about how we decide exactly which identifiers in an export list are deprecated. In particular, specify how to handle a combination of implicit and explicit exports.
  - If an identifier is in scope in several different ways, each with a distinct deprecation message, GHC should emit *all* the messages as warnings.

[Rendered](https://github.com/sheaf/ghc-proposals/blob/amend-dep-exports/proposals/0134-deprecating-exports-proposal.rst)

### DIFF
--- a/proposals/0134-deprecating-exports-proposal.rst
+++ b/proposals/0134-deprecating-exports-proposal.rst
@@ -6,137 +6,285 @@ Deprecating Exports
 .. ticket-url:: https://gitlab.haskell.org/ghc/ghc/-/issues/4879
 .. implemented::
 .. highlight:: haskell
-.. header:: This proposal was `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/134>`_.
+.. header:: This proposal was `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/134>`_, with `an amendment discussed here <https://github.com/ghc-proposals/ghc-proposals/pull/595>`_.
 .. contents::
 
-This is my GSoC 2018 project for Haskell.org
-
-The project should be helpful with easing the transition between different versions of code.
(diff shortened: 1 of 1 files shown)


proposals/0134-deprecating-exports-proposal.rst
```diff
@@ -6,137 +6,285 @@ Deprecating Exports
...
     ) where
     ...
 
-The deprecation pragma would have an effect on an export that immediately follows it, only top level exports would be considered.
+The deprecation pragma has an effect on the export item that immediately
+follows it. Deprecation pragmas are only allowed on top-level exports items.
+This means we would reject the following: ::
 
-The semantics are also described in the aforementioned ticket as follows:
+    module M ( T ( {-# DEPRECATED "msg" #-} fld ) ) where { .. }
```
bgamari, line 52: I do wonder why this restriction is necessary. I would have thought that this would work and it does seem like something one might want to do.

bgamari: I see, line 108 explains this. It would be good to mention that this restriction does not limit expressiveness, as pointed out below.

adamgundry:
I see that this PR is still open, but the described changes have (I presume) been merged and released as part of GHC 9.8. Still it would be good to keep the main branch of the proposals repo up to date. Are there any changes you'd like to make to this before submitting it for committee review @sheaf?

PR #597, opened 2023-06-17
 Allow constraint tuples and synonyms in typeclass instance head (under review)

uhbif19:
```
Most of Haskell projects do a lot of `stock` and `anyclass` deriving
for repeating classes. Most of time there is some default list
of classes to be defived.

Making type synonym for that seems like the obvious thing to do,
jet it does not work.

It worked at some moment for any instances,
but was disabled due to wrong behaviour in some cases:
https://gitlab.haskell.org/ghc/ghc/-/issues/13267

Seems like only phantom param case is relevant for deriving instances case.
```

I think this should work not only for heads and not only for deriving instances, but lets begin with this case, cuz I seem to roughly understand how this works.

Some previous discussion (as part of unrelated issue): https://gitlab.haskell.org/ghc/ghc/-/issues/23522#note_505273

Rendered: https://github.com/uhbif19/ghc-proposals/blob/uhbif19/allow-constraint-type-synonyms-in-deriving-heads/proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md

### DIFF
--- a/proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
+++ b/proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
@@ -7,66 +7,65 @@ implemented: ""
 
 This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/597).
 
-# Allow constraint tuples and synonyms in typeclass instances
+# Allow constraint tuples and synonyms in typeclass instance head
 
 ## Motivation
 
 Most of Haskell projects do a lot of `stock` and `anyclass` deriving
-for repeating classes. Most of time there is some default list
-of classes to be defined.
+for same list of common classes. This involves a lot of code duplication.
 
-Making type synonym for that seems like the obvious thing to do,
-yet it does not work.
+Such usecase would be natually covered by defining type synonym
+for tuple of such reused classes, but GHC does not support deriving
+not for constaint synonyms, nor for tuples.
 
-There are two constructions forbidden now, but required for that usecase:
+This may be unexpected for user, because both of this constructs
+are supported in the instance context, but not in the head.
 
-* Constraint synonyms
-* Constraint tuples
+Another usecase is typeclasses which are semantically same as composition of
+multiple classes. They too could be replaced by constraint synonym instances.
 
-Synonyms worked at some moment,
-but was disabled due to wrong behaviour in some cases.
+Case of constaraint synonyms in instance head worked at some moment,
+but was disabled due to wrong user-facing error messages in some cases.
 Reasoning for that is covered in Note "Instances and constraint synonyms"
 and [issue 13267](https://gitlab.haskell.org/ghc/ghc/-/issues/13267).
 
-Main usecase only affect deriving instances head,
-but another usecase is usage of tuples/synonyms in context of any instance.
-It seems to be less common, but natural to cover in same proposal
-to allow tuples/synonyms anywhere in instance declaration,
-leading to more consistent semantics.
+Relevant SA question (was provided by @yaitskov):
+https://stackoverflow.com/questions/28037837/deriving-clause-with-arbitrary-constraint-aliases
 
 ## Proposed Change Specification
 
-Proposal requires two purely syntaxic rewrites for instance declarations
-to happen:
+Proposal requires two new constructions allowed in instance head:
+constraint tuple and constraint synonym.
 
-1. Apply all type synonyms in instance declarations.
-   This should be done recursively, until no type synonym is present anymore.
-2. For any constraint tuple in derived instance head,
+New semantics can be described by two purely syntaxic rewrites,
+translating all new constructions away from instance declaration:
+
+1. Apply all constraint synonyms in instance declaration.
+   This should be done recursively,
+   until no constraint synonym is present anymore.
+2. For any constraint tuple in instance head,
    split such declarations into multiple.
 
-After that rewrites happen, existing GHC semantics is applied to result,
-with same error messages or produced code.
+This should work the same for any kind of instances:
+stanalone derived instances with or without explicit context,
+non-standalone derived instances
+and user-defined instances.
 
-There are two cases not covered by this rewrites:
-
-1. Using tuples in instance context leads to nested tuples.
-   But they already have the same semantics as flatten tuples,
-   so this is concern only affecting AST handling.
-2. Syntax and need for user-defined instances of class tuples is not clear.
-   So tuples in user-defined instances head remain forbidden.
+Thus proposal do not introduce any changes to instance semantics,
+only allowed syntax and user error messages would be improved.
 
 ## Examples
 
-Main usecase:
+### Common derived instances usecase
 
 ```haskell
 type ToFromJSON x = (ToJSON x, FromJSON x)
 
--- Allowed by this proposal
-derive instance ... => (ToFromJSON MyData)
-
 -- Allowed by this proposal and working same as example above
 data MyData = MkMyData deriving (ToFromJSON)
+
+-- Allowed by this proposal
+derive instance ... => (ToFromJSON MyData)
 
 -- Should work the same with syntax allowed before proposal
 derive instance ... => ToJSON MyData
@@ -78,31 +77,27 @@ so rewriting implementation may be little different.
 But they have just the same samantics as standalone instances with holes,
 covered by examples before, so could be handled the same.
 
-Instance context case is simpler:
+### Tuple of classes instance usecase
+
+Example from @VitWW :
 
 ```haskell
-class ToFromJSON x => MyProtocolEncoding x where
-   ...
+type Num x = (Arithm x, Abs x)
+type Arithm x = (Add x, Sub x, Mult x)
 
-data MyContainer x = ...
+data MyData
 
--- Allowed by this proposal
-instance
-   (MyProtocolEncoding x, ToFromJSON (MyContainer x)) =>
-   MyContainer x where
-   ...
+instance Num MyData where
+    (+) = ...
+    (*) = ...
 
--- Should work the same with syntax allowed before proposal
-instance
-   (MyProtocolEncoding x, ToJSON (MyContainer x), FromJSON (MyContainer x)) =>
-   MyContainer x where
-   ...
+-- same as
+instance Add MyData where
+    (+) = ...
+
+instance Mult MyData where
+    (*) = ...
 ```
-
-Last example may be not the most natural,
-because `MyContainer` would probably have `ToFromJSON` conditional instance
-anyway. But my goal was to show instance context changes separately,
-while more natural examples involve instance head as well.
 
 ## Effect and Interactions
 
@@ -138,18 +133,19 @@ For user-defined instances one may define type class instead of synonym:
 class (ToJSON x, FromJSON x) => ToFromJSON x
 
 instance (ToJSON x, FromJSON x) => ToFromJSON x
+```
 
-While this trick works, Haskell begginers may be not aware of it,
-and it makes less clear API.
+While this trick works, Haskell begginers may be not aware of it.
+Also it makes less clear API, because `ToFromJSON` is should be same as
+typeclass tuple only by syntaxic convention.
 
 ## Unresolved Questions
 
-* Are there any other "wrong" cases not covered in existing issue?
-* Which implementation approach to use?
+The only question is, which implementation approach to use.
 
-I think all those questions are not blocking,
-because they never lead to bad semantics,
-only potential problems are slightly less-clear type errors.
+I think this question is not blocking,
+because all implementation paths lead to same semantics,
+only potential difference may be slightly less-clear type errors.
 Also implementation paths are similar any may be switched later.
 
 ## Implementation Plan
@@ -164,9 +160,15 @@ There are two principal solutions to that:
 1. Check that rewriting will be correct before applying it.
    In this context this probably amounts to checking for each referenced
    type synonym if it belongs to some list of known bad cases.
-2. Optimistically apply transformation and check for errors in later stages
+2. Optimistically apply transformation and check for errors in later stages.
 
 Also we can just do both.
+
+Tuple of constraints implementation, on other hand, is clear.
+There is slight difference in cases of user-defined and derived instances,
+because in user-defined instances we need to keep track of which
+method is relevant to each instance,
+but overall the only thing needed is to split instances.
 
 ### Potential problems
 
@@ -191,7 +193,7 @@ The potential problem for solution 2 would be error messages clarity.
 Since checks are applied to rewritten code,
 they might miss original context or show non-original AST.
 I believe such problems could be solved by means of
-tracking relation to original AST.
+tracking relation of transformed AST to original AST.
 
 ### Proposed solution
 
--- a/proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
+++ b/proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
@@ -1,0 +1,91 @@
+---
+author: Grigorii Gerasev
+date-accepted: ""
+ticket-url: ""
+implemented: ""
+---
+
+This proposal is [discussed at this pull request](https://github.com/ghc-proposals/ghc-proposals/pull/0>).
+**After creating the pull request, edit this file again, update the number in
+the link, and delete this bold sentence.**
+
+# Proposal title
+
+Allow constraints type synonyms in deriving heads again.
+
+## Motivation
+
+Most of Haskell projects do a lot of `stock` and `anyclass` deriving
+for repeating classes. Most of time there is some default list
+of classes to be defived.
+
+Making type synonym for that seems like the obvious thing to do,
+jet it does not work.
+
+It worked at some moment for any instances,
+but was disabled due to wrong behaviour in some cases:
+https://gitlab.haskell.org/ghc/ghc/-/issues/13267
+
+Seems like only phantom param case is relevant for deriving instances case.
+
+## Proposed Change Specification
+
+Accept type synonyms in deriving heads, by rewriting it to regular constraints.
+
+## Examples
+
+```
+type Synonym x = (Show x, Eq x)
+
+-- Should works exactly the same
+derive instance ... => (Synonym x)
+derive instance ... => (Show x, Eq x)
+```
+
+Non-standalone instances do not have explicit params,
+so rewriting implementation may be little different.
+But they have same seamantics as standalone instances with holes,
+so it should be the same.
+
+## Effect and Interactions
+
+I do not know any.
+May be some linters depend on instance head being class.
+
+## Costs and Drawbacks
+
+May be some rewriting implementations may make errors less clear,
+due to AST changed inside. Seems not likely and treatible.
+
+## Backward Compatibility
+
+No breakage. This was working before.
+
+## Alternatives
+
+1. Use CPP preprocessor variable, which breaks type-safety
+   and may which lead to surprising error messages
+2. Use Template Haskell, which have its own problems and is less simple for beginners
+
+## Unresolved Questions
+
+Are there any other "wrong" cases other than phantom params?
+
+## Implementation Plan
+
+GHC does some processing to convert AST into EarlyDeriveSpec.
+As part of that before or in `makeDerivSpecs` function
+one could rewrite type synonyms. So all types and later logic
+will remain the same.
+
+The only question is how to check if rewriting will be correct.
+Another question is handling of recursive type synonyms.
+
+1. If the phantom params is the only case, one could forbid only them
+2. If there might be complex cases GHC may just rewrite synonims and
+   check if result is correct for later stages usual way
+
+I (Gregory Gerasev @uhbif19) could try to implement that.
+
+## Endorsements
+
--- a/proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
+++ b/proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
@@ -13,29 +13,37 @@ This proposal is [discussed at this pull request](https://github.com/ghc-proposa
 
 Most of Haskell projects do a lot of `stock` and `anyclass` deriving
 for same list of common classes. This involves a lot of code duplication.
-
 Such usecase would be natually covered by defining type synonym
-for tuple of such reused classes, but GHC does not support deriving
-neither for constraint synonyms, nor for tuples.
+for tuple of such reused classes.
+Examples section has
+[code sample how this could look](#common-derived-instances-usecase).
 
-This may be unexpected for user, because both of this constructs
-are supported in the context of an instance, but not in the head of an instance
+But GHC does not support deriving neither for constraint synonyms,
+nor for tuples.
+This behaviour is inconsistent and may be unexpected for user,
+because both of this constructs are supported in the context of an instance,
+but not in the head of an instance.
+For example in instance looking like `instance Context MyData => Head MyData`,
+`Context` can be synonym to tuple of constraints, while `Head` cannot.
 
 Another usecase is typeclasses which are semantically same as composition of
 multiple classes. They too could be replaced by constraint synonym instances.
+See [detailed code sample](#tuple-of-classes-instance-usecase).
 
-Case of constraint synonyms in instance head worked at some point in time,
-but was disabled due to wrong user-facing error messages in some cases.
-Reasoning for that is covered in Note "Instances and constraint synonyms"
-and [issue 13267](https://gitlab.haskell.org/ghc/ghc/-/issues/13267).
-
-Relevant SA question (was provided by @yaitskov):
-https://stackoverflow.com/questions/28037837/deriving-clause-with-arbitrary-constraint-aliases
+There is already
+[existing SA question](https://stackoverflow.com/questions/28037837/deriving-clause-with-arbitrary-constraint-aliases),
+asking about GHC proposal with similar scope and usecase
+(was provided by @yaitskov).
 
 ## Proposed Change Specification
 
 The proposal requires two new declarations to be allowed in head of the instance:
 constraint tuple and constraint synonym.
+
+Case of constraint synonyms in instance head worked at some point in time,
+but was disabled due to wrong user-facing error messages in some cases.
+Reasoning for that is covered in Note "Instances and constraint synonyms"
+and [issue 13267](https://gitlab.haskell.org/ghc/ghc/-/issues/13267).
 
 New semantics can be described by two purely syntactic rewrites,
 translating all new constructions away from instance declaration:
@@ -47,9 +55,10 @@ translating all new constructions away from instance declaration:
    split such declarations into multiple.
 
 This should work the same for any kind of instances:
-stanalone derived instances with or without explicit context,
-non-standalone derived instances
-and user-defined instances.
+
+* Stanalone derived instances with or without explicit context
+* Non-standalone derived instances
+* User-defined instances
 
 Thus, the proposal does not introduce any changes to instance semantics,
 only the allowed syntax and error messages would be improved.
--- a/proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
+++ b/proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
@@ -16,15 +16,15 @@ for same list of common classes. This involves a lot of code duplication.
 
 Such usecase would be natually covered by defining type synonym
 for tuple of such reused classes, but GHC does not support deriving
-not for constaint synonyms, nor for tuples.
+neither for constraint synonyms, nor for tuples.
 
 This may be unexpected for user, because both of this constructs
-are supported in the instance context, but not in the head.
+are supported in the context of an instance, but not in the head of an instance
 
 Another usecase is typeclasses which are semantically same as composition of
 multiple classes. They too could be replaced by constraint synonym instances.
 
-Case of constaraint synonyms in instance head worked at some moment,
+Case of constraint synonyms in instance head worked at some point in time,
 but was disabled due to wrong user-facing error messages in some cases.
 Reasoning for that is covered in Note "Instances and constraint synonyms"
 and [issue 13267](https://gitlab.haskell.org/ghc/ghc/-/issues/13267).
@@ -34,10 +34,10 @@ https://stackoverflow.com/questions/28037837/deriving-clause-with-arbitrary-cons
 
 ## Proposed Change Specification
 
-Proposal requires two new constructions allowed in instance head:
+The proposal requires two new declarations to be allowed in head of the instance:
 constraint tuple and constraint synonym.
 
-New semantics can be described by two purely syntaxic rewrites,
+New semantics can be described by two purely syntactic rewrites,
 translating all new constructions away from instance declaration:
 
 1. Apply all constraint synonyms in instance declaration.
@@ -51,8 +51,8 @@ stanalone derived instances with or without explicit context,
 non-standalone derived instances
 and user-defined instances.
 
-Thus proposal do not introduce any changes to instance semantics,
-only allowed syntax and user error messages would be improved.
+Thus, the proposal does not introduce any changes to instance semantics,
+only the allowed syntax and error messages would be improved.
 
 ## Examples
 
@@ -61,7 +61,7 @@ only allowed syntax and user error messages would be improved.
 ```haskell
 type ToFromJSON x = (ToJSON x, FromJSON x)
 
--- Allowed by this proposal and working same as example above
+-- Allowed by this proposal and works the same as the example above
 data MyData = MkMyData deriving (ToFromJSON)
 
 -- Allowed by this proposal
@@ -73,9 +73,9 @@ derive instance ... => FromJSON MyData
 ```
 
 Non-standalone instances do not have explicit params,
-so rewriting implementation may be little different.
-But they have just the same samantics as standalone instances with holes,
-covered by examples before, so could be handled the same.
+therefore rewriting their implementation may be little different.
+However they have just the same samantics as standalone instances with holes,
+covered by examples before, and could therefore be handled the same.
 
 ### Tuple of classes instance usecase
 
@@ -101,33 +101,33 @@ instance Mult MyData where
 
 ## Effect and Interactions
 
-There is already extension named `TypeSynonymInstances`,
-but it does not cover constraint synonym case.
-Probably, constraint synonyms should be guarded by this extension as well
-and documentation updated on that.
+We already have the `TypeSynonymInstances` extension,
+but it does not cover the constraint synonym case.
+Constraint synonyms should probably be guarded by this extension as well
+and the documentation updated to reflect that.
 
 It may be, that some linters depend on current syntaxic restrictions,
-but I am not aware on any specific cases.
+the authors however, are not aware on any specific cases.
 
 ## Costs and Drawbacks
 
-Main concern is that rewriting implementations may make errors less clear.
-I propose to mitigate if this happens by adding context to errors,
-same as it done for instances resolving now.
+The primary concern is that rewriting implementations may make errors less clear.
+The authors propose to mitigate this by adding context to errors, in cases where this might happen.
+The same is already done for instances resolving today.
 
 ## Backward Compatibility
 
-No breakage. Part of this even was working before.
+No breakage. Part of this even used to work in the past.
 
 ## Alternatives
 
 For deriving instances there are two alternatives:
 
-1. Use CPP preprocessor variable, which breaks type-safety,
-   may lead to surprising error messages and is less composable
-2. Use Template Haskell, which have its own problems and is less simple for beginners
+1. Use CPP, which breaks type-safety, and
+   may lead to surprising error messages as well as being less composable
+2. Use Template Haskell, which has its own set of problems and is harder to learn for beginners
 
-For user-defined instances one may define type class instead of synonym:
+For user-defined instances one may define a type class instead of a synonym:
 
 ```haskell
 class (ToJSON x, FromJSON x) => ToFromJSON x
@@ -135,17 +135,17 @@ class (ToJSON x, FromJSON x) => ToFromJSON x
 instance (ToJSON x, FromJSON x) => ToFromJSON x
 ```
 
-While this trick works, Haskell begginers may be not aware of it.
-Also it makes less clear API, because `ToFromJSON` is should be same as
-typeclass tuple only by syntaxic convention.
+While this trick works, Haskell beginners may not be aware of it.
+It also leads to a less clear API, because `ToFromJSON` should be same as a
+typeclass tuple only by syntactic convention.
 
 ## Unresolved Questions
 
 The only question is, which implementation approach to use.
 
-I think this question is not blocking,
-because all implementation paths lead to same semantics,
-only potential difference may be slightly less-clear type errors.
+The authors do not believe the question of implementation to be blocking,
+because all implementation paths lead to same semantics.
+A potential difference in implementation paths could lead to slightly sub-optimal errors messages.
 Also implementation paths are similar any may be switched later.
 
 ## Implementation Plan
@@ -162,10 +162,10 @@ There are two principal solutions to that:
    type synonym if it belongs to some list of known bad cases.
 2. Optimistically apply transformation and check for errors in later stages.
 
-Also we can just do both.
+We could also do both.
 
 Tuple of constraints implementation, on other hand, is clear.
-There is slight difference in cases of user-defined and derived instances,
+There is a slight difference in cases of user-defined and derived instances,
 because in user-defined instances we need to keep track of which
 method is relevant to each instance,
 but overall the only thing needed is to split instances.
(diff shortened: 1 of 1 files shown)


proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,89 @@
...
+Accept type synonyms in deriving heads, by rewriting it to regular constraints.
+
+## Examples
+
+```
+type Synonym x = (Show x, Eq x)
+
+-- Should works exactly the same
+derive instance ... => (Synonym x)
+derive instance ... => (Show x, Eq x)
```
VitWW, line 40: correction:
```suggestion
## Examples

```haskell
type Synonym x = (Show x, Eq x)

-- Should works exactly the same
deriving instance ... => (Synonym x)
deriving instance ... => (Show x, Eq x)
```

VitWW (changes requested):

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
uhbif19: Thanks

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,89 @@
...
+Most of Haskell projects do a lot of `stock` and `anyclass` deriving
+for repeating classes. Most of time there is some default list
+of classes to be defived.
+
+Making type synonym for that seems like the obvious thing to do,
+jet it does not work.
+
+It worked at some moment for any instances,
+but was disabled due to wrong behaviour in some cases:
+https://gitlab.haskell.org/ghc/ghc/-/issues/13267
```
VitWW, line 25: Thanks for asking a review from me.
But I'm not a part of Committee, I'm not a contributor to GHC or famous Hackage library. My thoughts are just my thoughts.

I like your proposal. I wish even more: use type synonyms in instance heads
```haskell
type ReadShow x = (Show x, Read x)

data MyData

instance ReadShow MyData where
    show a = ...
    read a = ...

-- same as 
instance (Show MyData, Read MyData) where
    show a = ...
    read a = ...
```

VitWW (changes requested):

euonymos:
When and why it stopped working?

uhbif19:
@euonymos There is link on that issue in "motivation" part:

https://gitlab.haskell.org/ghc/ghc/-/issues/13267

euonymos:
I see. So they just removed all the strange things. To be honest I think aliases like that might pose a serious naming problem (and a reading problem consequently). How would you name them?

uhbif19:
> To be honest I think aliases like that might pose a serious naming problem (and a reading problem consequently).

How is this different from any other type synonym?

euonymos:
Because these ones are lists of unrelated things. How would you name (Show a, Eq a)? Aliases for single types might make sense, so may aliases for arrow types when we capture the whole idea like `Application = Request -> Response`, but a relation under a deriving strategy doesn't give a good way to name it I guess. That being said, I am not arguing it should not be done.

uhbif19:
> Because these ones are lists of unrelated things

Naming strategy is not related to Haskell semantics or necessity this feature, so is not in the scope of GHC proposal, I believe.

Also I already use such synonyms in multiple other situations and did seem them in other projects as well, so I do not think this changes status quo in any way. After all GHC already decided to allow such synonyms with `ConstraintKinds`, and it is even already mentioned in GHC docs: see "This allows type constraint synonyms" in https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/constraint_kind.html#the-constraint-kind

uhbif19:
@nomeata I believe no discussion happening because semantics of proposal is very simple. So I am asking to bring a proposal before the committee.

simonpj:
I can see the motivation here.

My concern is this.   Under this proposal we would have
```
type Wombat a = Eq a

data T1 a = MkT1 a deriving( Wombat )   -- Allowed

data T2 a = MkT2 a
instance Eq a => Wombat (S a) where ...    -- Not allowed

data T3 a = MkT3 a 
deriving instance Eq a => Wombat (S a)   -- I'm not sure ... the proposal is silent on this
```
It's strange to accept T1 but reject T2.  Argubly we should apply the same treatment to all instance heads.

------------------------
The entire "Proposed change specification" is this

* Accept type synonyms in deriving heads, by rewriting it to regular constraints.

This needs to be fleshed out a lot more and made more precise.  What is a "deriving head"?   What does "rewriting" mean? What about `deriving via` and other deriving mechanisms?  What about `deriving instance`?

I think I know, or can guess, but GHC proposals should be precise.  Please!

uhbif19:
> It's strange to accept T1 but reject T2. Argubly we should apply the same treatment to all instance heads.

Sure, and that is required feature as well. Probably same for instance context as well.

I decided to postpone it to follow-up proposal for two reasons:

1. May be there are some problems with regular instances, I am not aware of
2. I already understood how GHC deriving code works overall, so I have vague idea how to implement it in that case, but did not check regular instances case

Maybe it is better to extend proposal on all instances.

> What is a "deriving head"? 

I thought that is some known terminology. IIRC I taken it from GHC codebase or docs.

> What about deriving via and other deriving mechanisms?

I think they do not matter.
DeriveVia have some differences in implementation, but it seems like they do not matter for this case.
I will add that statement to proposal.

> What does "rewriting" mean?

I thought that type synonym application has known semantics, but I can include that to proposas.

> What about deriving via and other deriving mechanisms? What about deriving instance?

Standalone instances are already in examples. Actually I see them little bit simpler due to explicit type synonym param.

simonpj:
@uhbif19 it is helpful to reply here, but better still to update the proposal so that the questions do not arise in the first place.  No rush, but please do that before submitting it to the committee.  (They will just look at the the proposal, not proposal + clarifications in discussion.)  Strive for precision!

> Maybe it is better to extend proposal on all instances.

It's up to you, but I for one would be reluctant to accept an inconsitent state, in which synonyms were permitted in some places but not others.

uhbif19:
> but better still to update the proposal so that the questions do not arise in the first place

Sure, it was just quick reply for validation.

> It's up to you, but I for one would be reluctant to accept an inconsitent state, in which synonyms were permitted in some places but not others.

I do not see this as some serious inconsistency. Deriving is to me some meta-programming tool, which happens to be implemented inside GHC. As meta-programming it may do some additional steps before producing instance signature, it already does that for `_` context.

On code level IIRC this is checked separately as well and produce different errors.

uhbif19:
@simonpj

Seems that I missed more important point in previous replies.

Main case for that proposal is type synonyms to constraint tuples.

But non-derived instance for tuple constrains seem strange and I think they do not have usecases. On other side for context part of instance constraint synonym for tuple always makes sense, it just results in nested tuple.

So my proposal seem to make two different transformations on derived instances:

* Recrusively apply all type synonyms in instance head,
  until none are left.
* If instance head is tuple, split it into multiple deriving clauses.

That shows more bold inconsitency, that I missed: proposal only allows instances for tuple synonyms,
but not tuples themselves.

So there are various ways to pack this changes into proposal(s). My main motivating example does not work without both of this changes, so I just extended it to cover allowing constraint synonyms and tuples in any place of any instance.
Examples and semantics are clarified.

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,200 @@
...
+# Allow constraint tuples and synonyms in typeclass instances
+
+## Motivation
+
+Most of Haskell projects do a lot of `stock` and `anyclass` deriving
+for repeating classes. Most of time there is some default list
+of classes to be defined.
+
+Making type synonym for that seems like the obvious thing to do,
+because jet it does not work.
```
aspiwack, line 19:
```suggestion
Making type synonym for that seems like the obvious thing to do,
yet it does not work.
```
(I suppose?)

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,200 @@
...
+Synonyms worked at some moment,
+but was disabled due to wrong behaviour in some cases.
+Reasoning for that is covered in Note "Instances and constraint synonyms"
+and [issue 13267](https://gitlab.haskell.org/ghc/ghc/-/issues/13267).
+
+Main usecase only affect deriving instances head,
+but another usecase is usage of tuples/synonims in context of any instance.
+It seems to be less common, but natural to cover in same proposal
+to allow tuples/synonims anywhere in instance declaration,
+leading to more consistent semantics.
```
aspiwack, line 35:
```suggestion
Main usecase only affect deriving instances head,
but another usecase is usage of tuples/synonyms in context of any instance.
It seems to be less common, but natural to cover in same proposal
to allow tuples/synonyms anywhere in instance declaration,
leading to more consistent semantics.
```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Proposed Change Specification
+
+Proposal requires two purely syntaxic rewrites for in instance declarations
+to happen:
```
aspiwack, line 40:
```suggestion
Proposal requires two purely syntaxic rewrites for instance declarations
to happen:
```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,200 @@
...
+
+1. Apply all type synonyms in instance declarations.
+   This should be done recrusively, until no type synonim is present anymore.
```
aspiwack, line 43:
```suggestion
   This should be done recursively, until no type synonym is present anymore.
```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
aspiwack: Unless I'm misunderstanding, the secondary use-case is already possible in GHC today

```
> import GHC.Types
> type ShowEq :: Type -> Constraint; type ShowEq a = (Show a, Eq a)
> instance ShowEq a => Show (a -> b) where { show _ = "example" }
```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,200 @@
...
+   may lead to surprising error messages and is less composable
+2. Use Template Haskell, which have its own problems and is less simple for beginners
+
+For user-defined instances one may define type class instead of synonym:
+
+```haskellb
+class (ToJSON x, FromJSON x) => ToFromJSON x
+
+instance (ToJSON x, FromJSON x) => ToFromJSON x
+```
```
aspiwack, line 138: I don't think this is actually an alternative. I can't derive or define an instance of `ToFromJSON` without also (manually) deriving or defining `ToJSON` and `FromJSON` instances.

aspiwack:
```suggestion
```haskell
class (ToJSON x, FromJSON x) => ToFromJSON x

instance (ToJSON x, FromJSON x) => ToFromJSON x
```
```

File: proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md [hunk 8]
```diff
@@ -0,0 +1,200 @@
...
+Also implementation paths are similar any may be switched later.
+
+## Implementation Plan
+
+### Main approaches
+
+The main question is how to check that constraint synonym rewriting
+will lead to correct code.
+
+There are two principial solutions to that:
```
aspiwack, line 160:
```suggestion
There are two principal solutions to that:
```

aspiwack (commented):
I've got a question: do you intend for your proposal to support

```
data …
  deriving (TypeSynonym, …)
```

If you do, please add a corresponding example in the _Examples_ section.  In this case, there are probably some details to iron out in the _Specification_ section because, at least to me, it's not obvious what the semantics would be.

If you don't please call out explicitly that this is not supported by your proposal in the _Specification_ section.

A few remarks below (I took the opportunity to suggest fixes to a few typos).

uhbif19:
> If you do, please add a corresponding example in the Examples section

It was covered by text explanation in examples section. Added separate example for clarity. Semantics seems clear to me, because every non-standalone deriving should be equivalent to standalone deriving with inferred context.

aspiwack:
> Semantics seems clear to me, because every non-standalone deriving should be equivalent to standalone deriving with inferred context.

There's more than one way to go about it, though. So you want to be specific about your intended meaning. You haven't pushed your new example yet, though, maybe it'll go a long way to elucidate what you have in mind.

uhbif19:
> You haven't pushed your new example yet, though, maybe it'll go a long way to elucidate what you have in mind

Fixed. While example is trivial and I do not think it changes situation a lot.

Could you give examples of cases in which translation to standalone deriving is not clear?

VitWW:
> on-derived instance for constraint tuple appear strange to me (while @VitWW asked for them in comment before), and I am not sure if they are really important usecase.

Constraint tuple on instance head significantly helps both unify **without penalties** several classes and separate one into several.
I could create an RFC if it is necessary, but it looks like this RFC could absorb this sub-proposal.

Combine 2 classes example:
```haskell
type ReadShow x = (Show x, Read x)

data MyData

instance ReadShow MyData where
    show a = ...
    read a = ...

-- same as 
instance (Show MyData, Read MyData) where
    show a = ...
    read a = ...
```

Separate 1 class to several classes example:
```haskell
type Num x = (Arithm x, Abs x)
type Arithm x = (Add x, Sub x, Mult x)

data MyData

instance Num MyData where
    (+) = ...
    (*) = ...

-- same as 
instance Add MyData where
    (+) = ...

instance Mult MyData where
    (*) = ...
```

uhbif19:
@VitWW 

What is "RFC" in this context?

Okay, I will include that feature and motivation for it into current proposal. I agree that seems to simplify things and make them more consistent, I am just afraid to add things which cannot be explained merely by syntax rewriting (while this is not feature is not far from it, you just need to look up list of functions for given classes), because that feels like less obvious change to me.

uhbif19:
> for example, the interaction with (Num(..)) in import/export lists

Why there should be any interaction of that kind? The scope of this proposal is only allowed instances. Constraint synonyms were already exist and I did not seem any reasons to affect their semantics so far.

> And I expect it needs a even more tight and rigorous explanation of the new syntax and it's semantics.

What is incomplete with examples provided by @VitWW (once I will include their explanation in proposal)?

For me this proposal remains to be fully defined by rewriting new syntax into existing. So semantics should be defined by that. The only complication is that real GHC implementation may need to lookup class definition to do such rewriting (or may be do not actually rewrite AST at all) but that should not affect semantics in any way.

VitWW:
@uhbif19 Thanks!
RFC (request for comments) is a Proposal.
And context is to liberalize uses (allow to use) constraint tuples in deriving and instance heads as synonyms

yaitskov:
Today I realized the need in such extension in my app after a series of mistakes mixing ids I decided to wrap types for db key columns into newtypes and deriving section is duplicated for every key column. An average app could have dozens of tables:

```
type RowId = Int32

newtype MetricRowId = MetricRowId RowId
  deriving newtype
    ( Show, Eq, Ord, ToJSONKey, FromJSONKey, ToJSON, FromJSON
    , HasSqlEqualityCheck Postgres
    , FromBackendRow Postgres
    , HasSqlValueSyntax PgValueSyntax
    )

```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,202 @@
...
+   may lead to surprising error messages and is less composable
+2. Use Template Haskell, which have its own problems and is less simple for beginners
+
+For user-defined instances one may define type class instead of synonym:
+
+```haskell
+class (ToJSON x, FromJSON x) => ToFromJSON x
+
+instance (ToJSON x, FromJSON x) => ToFromJSON x
+
```
angerman, line 141: The code block needs to be closed here.

uhbif19:
Gonna update proposal as discussed this weekend, apologies for slowness.

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,202 @@
...
+instance
+   (MyProtocolEncoding x, ToFromJSON (MyContainer x)) =>
+   MyContainer x where
+   ...
+
+-- Should work the same with syntax allowed before proposal
+instance
+   (MyProtocolEncoding x, ToJSON (MyContainer x), FromJSON (MyContainer x)) =>
+   MyContainer x where
+   ...
```
konsumlamm, line 99: `MyContainer x` is a type, not an instance, what's going on here? Did you perhaps mean to write `MyProtocolEncoding (MyContainer x)`?

angerman:
> Gonna update proposal as discussed this weekend, apologies for slowness.

@uhbif19 all good. Looking forward to the updated proposal.

Ericson2314:
I think his proposal is very underspecified
```haskell
instance (A, B)
```
is the same as
```haskell
instance () A B
```
and the problem with that is it is an overlapping instance with
```haskell
instance (a, b) => (a, b)
```
I do not think it is worth adding some rather ad-hoc exceptions to overlapping instances checks.

VitWW:
> @Ericson2314 
> and the problem with that is it is an overlapping instance with
> 
> ```haskell
> instance (a, b) => (a, b)
> ```
> 
> I do not think it is worth adding some rather ad-hoc exceptions to overlapping instances checks.

In expression `instance (A, B)` both `A` and `B` are **NOT** types, but type classes (with params)!
So **no ambiguity** is here!

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
uhbif19: Thanks, solved.

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
uhbif19: This example was replaced with simpler one. Thanks for your finding.

uhbif19:
Updated proposal to allow all, not only derived, instance head, as @VitWW proposed and changed explanations somewhere. So stuff is more uniform and general now and there is second main motivation now.

I can add more low-level GHC-related details on how to implement that, but I think this is not really important for proposal, because it does not change semantics. And modulo user-facing errors issues, constraint synonym rewriting is already working for instance context, so it is clearly possible to implement.

Maybe I forgot to fix some comments from discussion, not sure.

FYI @nomeata @VitWW  @angerman

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,204 @@
...
+Probably, constraint synonyms should be guarded by this extension as well
+and documentation updated on that.
+
+It may be, that some linters depend on current syntaxic restrictions,
+but I am not aware on any specific cases.
+
+## Costs and Drawbacks
+
+Main concern is that rewriting implementations may make errors less clear.
+I propose to mitigate if this happens by adding context to errors,
```
angerman, line 115:
```suggestion
The authors propose to mitigate this by adding context to errors, in cases where this might happen.
```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,204 @@
...
+same as it done for instances resolving now.
```
angerman, line 116:
```suggestion
The same is already done for instances resolving today.
```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Backward Compatibility
+
+No breakage. Part of this even was working before.
```
angerman, line 120:
```suggestion
No breakage. Part of this even used to work in the past.
```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Alternatives
+
+For deriving instances there are two alternatives:
+
+1. Use CPP preprocessor variable, which breaks type-safety,
```
angerman, line 126:
```suggestion
1. Use CPP, which breaks type-safety, and
```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,204 @@
...
+   may lead to surprising error messages and is less composable
```
angerman, line 127:
```suggestion
   may lead to surprising error messages as well as being less composable
```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,204 @@
...
+2. Use Template Haskell, which have its own problems and is less simple for beginners
```
angerman, line 128:
```suggestion
2. Use Template Haskell, which has its own set of problems and is harder to learn for beginners
```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,204 @@
...
+
+For user-defined instances one may define type class instead of synonym:
```
angerman, line 130:
```suggestion
For user-defined instances one may define a type class instead of a synonym:
```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,204 @@
...
+
+```haskell
+class (ToJSON x, FromJSON x) => ToFromJSON x
+
+instance (ToJSON x, FromJSON x) => ToFromJSON x
+```
+
+While this trick works, Haskell begginers may be not aware of it.
```
angerman, line 138:
```suggestion
While this trick works, Haskell beginners may not be aware of it.
```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,204 @@
...
+Also it makes less clear API, because `ToFromJSON` is should be same as
+typeclass tuple only by syntaxic convention.
```
angerman, line 140:
```suggestion
It also leads to a less clear API, because `ToFromJSON` should be same as a
typeclass tuple only by syntactic convention.
```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Unresolved Questions
+
+The only question is, which implementation approach to use.
+
+I think this question is not blocking,
```
angerman, line 146:
```suggestion
The authors do not believe the question of implementation to be blocking,
```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,204 @@
...
+because all implementation paths lead to same semantics,
+only potential difference may be slightly less-clear type errors.
```
angerman, line 148:
```suggestion
because all implementation paths lead to same semantics.
A potential difference in implementation paths could lead to slightly sub-optimal errors messages.
```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,204 @@
...
+will lead to correct code.
+
+There are two principal solutions to that:
+
+1. Check that rewriting will be correct before applying it.
+   In this context this probably amounts to checking for each referenced
+   type synonym if it belongs to some list of known bad cases.
+2. Optimistically apply transformation and check for errors in later stages.
+
+Also we can just do both.
```
angerman, line 165:
```suggestion
We could also do both.
```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,204 @@
...
+
+Tuple of constraints implementation, on other hand, is clear.
+There is slight difference in cases of user-defined and derived instances,
```
angerman, line 168:
```suggestion
There is a slight difference in cases of user-defined and derived instances,
```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,204 @@
...
+are amplified by problem of cascading:
+handling synonyms application producing jet another synonyms.
+This would affect both calculation of referenced type synonyms
+and possible generative effects.
+In particular this leads to question if to check referenced type synonyms
+once or after each reduction pass.
+Another concern would be if reduction strategy affects user-facing errors.
+
+Solution 2, on other hand cannot be unsafe or too restrictive by design.
+Also it may lead to GHC code simplification,
```
angerman, line 188:
```suggestion
It may also lead to GHC code simplification,
```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,204 @@
...
+because now there are multiple codepaths
+need to handle type synonims differently.
+
+The potential problem for solution 2 would be error messages clarity.
+Since checks are applied to rewritten code,
+they might miss original context or show non-original AST.
+I believe such problems could be solved by means of
```
angerman, line 195:
```suggestion
The authors believe such problems could be solved by means of
```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,204 @@
...
+tracking relation of transformed AST to original AST.
+
+### Proposed solution
+
+I believe second approach is more simple to reason and implement.
```
angerman, line 200:
```suggestion
The authors believe the second approach is simpler to reason and implement.
```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,204 @@
...
+I (Gregory Gerasev @uhbif19) could try to provide such implementation.
```
angerman, line 201:
```suggestion
The authors (Gregory Gerasev @uhbif19) would volunteer to provide such implementation.
```

angerman (changes requested):
@uhbif19 I'm sorry for the delay. I think this is a good proposal, but could be made significantly stronger with some editing. I've tried to provide some suggestions.

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,204 @@
...
+# Allow constraint tuples and synonyms in typeclass instance head
+
+## Motivation
+
+Most of Haskell projects do a lot of `stock` and `anyclass` deriving
+for same list of common classes. This involves a lot of code duplication.
+
+Such usecase would be natually covered by defining type synonym
+for tuple of such reused classes, but GHC does not support deriving
+not for constaint synonyms, nor for tuples.
```
konsumlamm, line 19:
```suggestion
not for constraint synonyms, nor for tuples.
```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,204 @@
...
+
+This may be unexpected for user, because both of this constructs
+are supported in the instance context, but not in the head.
+
+Another usecase is typeclasses which are semantically same as composition of
+multiple classes. They too could be replaced by constraint synonym instances.
+
+Case of constaraint synonyms in instance head worked at some moment,
```
konsumlamm, line 27:
```suggestion
Case of constraint synonyms in instance head worked at some moment,
```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,204 @@
...
+This would affect both calculation of referenced type synonyms
+and possible generative effects.
+In particular this leads to question if to check referenced type synonyms
+once or after each reduction pass.
+Another concern would be if reduction strategy affects user-facing errors.
+
+Solution 2, on other hand cannot be unsafe or too restrictive by design.
+Also it may lead to GHC code simplification,
+because now there are multiple codepaths
+need to handle type synonims differently.
```
konsumlamm, line 190:
```suggestion
need to handle type synonyms differently.
```

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Motivation
+
+Most of Haskell projects do a lot of `stock` and `anyclass` deriving
+for same list of common classes. This involves a lot of code duplication.
+
+Such usecase would be natually covered by defining type synonym
+for tuple of such reused classes, but GHC does not support deriving
+not for constaint synonyms, nor for tuples.
+
```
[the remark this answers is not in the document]
uhbif19: Well examples are in example sections. I see it breaking the explanation as too, but I this structure was in the proposal template.

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,204 @@
...
+This may be unexpected for user, because both of this constructs
+are supported in the instance context, but not in the head.
```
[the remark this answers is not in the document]
uhbif19: Examples of using type synonyms in instance contexts?

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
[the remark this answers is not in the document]
angerman: The examples section is--I believe--for more in-depth examples. We can still put the parts of the examples that help the motivation into the motivation. See also e.g. https://github.com/strake/ghc-proposals/blob/acomp/proposals/0000-applicative-comprehensions.rst for one which has examples in the motivation.

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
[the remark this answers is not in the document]
angerman: Yes. You write it's supported in the context, but not in the head. Let's assume the reader of this document is uninitiated and does not know much about the intricacies and what is permitted and not today.

angerman:
@uhbif19 please ping me if you want me to give this another pass.

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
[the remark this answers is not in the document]
uhbif19: Sorry, but I do not see, how code from examples sections can be made shorter. Without context of specific classes they may be read multiple ways, and with them they are getting "in-depth" as you say.

proposals/0000-allow-constraint-type-synonyms-in-deriving-heads.md
```diff
@@ -0,0 +1,204 @@
...
+Another usecase is typeclasses which are semantically same as composition of
+multiple classes. They too could be replaced by constraint synonym instances.
+
+Case of constaraint synonyms in instance head worked at some moment,
+but was disabled due to wrong user-facing error messages in some cases.
+Reasoning for that is covered in Note "Instances and constraint synonyms"
+and [issue 13267](https://gitlab.haskell.org/ghc/ghc/-/issues/13267).
+
+Relevant SA question (was provided by @yaitskov):
+https://stackoverflow.com/questions/28037837/deriving-clause-with-arbitrary-constraint-aliases
```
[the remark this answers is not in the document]
uhbif19: Added small explanation. Is relevance stated clear now?

uhbif19:
@angerman 

I do not think that examples could be made into one-liners, so I just added links to them and re-arranged motivation section. Hopefully it is easier to follow now.

Other suggestions seem to be covered as well.

simonpj:
> I created a similar proposal as a ticket https://gitlab.haskell.org/ghc/ghc/-/issues/7543. Simon responded that we rename the class members before the typechecker runs, which makes the implementation difficult. This was 10 years ago but as far as I know this is still true. Perhaps support for deriving only will be simpler.

Yes, I *think* `deriving` is a bit simpler because `deriving` is dealt with in the typechecker not the renamer.  It should not be especially difficult to expand synonyms.  So it should not be hard to treat
```
type C a = (D a, E a)
data T b  = ... deriving( C )
```
just as if you had written
```
data T b = ... deriving( D, E )
```
But note that it's a bit tricky because it relies on the `a` being the last parameter, and not repeated.  E.g.
```
type BadC x y = D y x
data T b = ... deriving( BadC Int )
```
What would you rewrite that to?  Maybe
```
data T b = ... deriving( \b . D b Int)
```
But we don't have type level lambdas. So you would have to write down rules for when the abbreviation is OK.

----------------

Moreover the proposal ranges much wider: "The proposal requires two new declarations to be allowed in head of the instance: constraint tuple and constraint synonym".  That covers a lot more than `deriving`, but the proposal has no motivation for such an extension, and it isneed tricky as @monoidal says (https://gitlab.haskell.org/ghc/ghc/-/issues/7543).


With `DerivingStrategies` it would only makes sense of all of the synonymous constraints could be derived using the same strategy.

Colour me unconvinced, for now anyway.

VitWW:
> @simonpj  Moreover the proposal ranges much wider: "The proposal requires two new declarations to be allowed in head of the instance: constraint tuple and constraint synonym". That covers a lot more than `deriving`, but the proposal has no motivation for such an extension, and it isneed tricky

1) Motivation to use constraint synonym is ability to write much shorter code and more pretty code
2) Motivation to use constraint tuple is to allow to regroup classes without rewriting their realizations (derivings and instances). Which is crucial for libs

> So you would have to write down rules for when the abbreviation is OK

Agree

uhbif19:
> So you would have to write down rules for when the abbreviation is OK.

I think that this issue is covered in my proposal, while quite maybe that my answers are not best. The simplest solution I see is just perform rewriting optimistically and type-check after each pass. That seem to only affect error messages quality, not correctness.

As for renaming issue, I was not aware of it. I do not understand renamer well, but it appears strange, how deriving layer possibly can do something regular instance cannot. Like could not you affect scopes with adding deriving instance?

uhbif19:
@angerman

AFAIU you are shepherd for this proposal.

1. I am not clear, what is current status of proposal? Are some actions or changes awaited from me now?
2. In spite of latest @monoidal and @simonpj comments, in your opinion, would it help for proposal reviewing process if I split this proposal to "easy" deriving part and less clear general instance part?

simonpj:
> In spite of implementation issue mentioned in last @monoidal and @simonpj comments, in your opinion, would it help for proposal reviewing process, if I split this proposal to "easy" deriving part and less clear general instance part?

I think that would be helpful, yes.  Here's why.

-----------------------

First, consider
```
class S a = (C1 a, C2 a)

class C1 a where
  op1 :: a -> a
  op  :: a -> a

class C2 a where
  op2 :: a -> a
  op  :: a -> a

instance Eq a => S [a] where
  -- For C1
  op1 = ...
  op = ....

  -- For C2
  op2 = ...
  op = ....
```
Note that `C1` and `C2` both have a method `op`; that is fine in Haskell if `C1` and `C2` are declared in different modules.

Now in the `instance Eq a => S [a]`:
1. It is entirely unclear how to (as the proposal puts it) "split such declarations into multiple".  Which `op` belongs with with instance?
2. In implementation terms, the renamer (which knows nothing of types, we resolve each identifier occurrence to its binding site.
   But here, even for `op1` and `op2` it is far from clear what the right binding site is, knowing only "`S [a]`" and without the ability
   to expand synonyms (which the renamer does not have. (Remember there might be another top-level definition of `op1` or `op2`.)

3. It is unlikely that `Eq a` will be the right context for both `C1 [a]` and `C2 [a]`.  This is a serious problem.
4. Very little is gained, because we still have to write out the method declarations, of course.  Is itso much worse to say
   ```
   instance C1 a => C1 [a] where
     -- For C1
     op1 = ...
     op = ....

   instance C2 a => C2 [a] where
     -- For C2
     op2 = ...
     op = ....
   ```

Short summary: much pain, virtually no gain.  Thumbs down.

Ponts (1) and (2) do not apply to standalone `deriving`, but point (3) certainly does.  e.g.
```
   deriving instance Eq a => S [a]
```
is unlikely to be the Right Thing.

-----------------------

However, for vanilla `deriving`, *where the programmer specifies neither the instance context, nor the methods*, I can see some gain.
```
data T a = ... deriving( S )
```
is shorter than
```
data T a = ... deriving( C1, C2 )
```
But then you need to answer [my question above](https://github.com/ghc-proposals/ghc-proposals/pull/597#issuecomment-1839016127) about synonyms with out of order parameters.
I can think of one very simple (albeit a bit crude) restrition:

* The synonym declaration must look like
  ```
  type S a b c = (C1 a c, C2 c, C3 a a c, ...)
  ```
  in which the last parameter of the LHS (here `c`) must appear as the last parameter of each of the classes on the RHS, and only in that way.  So these would be illegal
  ```
  type S1 a b = (C1 b b, ...)  -- b appears twice
  type S2 a b = (C1 a, ...)    -- b does not appear at all
  type S3 a b = (C1 b a, ...)  -- b is not last
  ```

Even then one might wonder whether
* Can you have *nested* tuples?   E.g.
  ```
  data T a = ... deriving( S )
  type S a = (S1 a, S2 a)
  type S1 a = (C1 a, C2 a)
  type S2 a = (C3 a, C4 a)
  ```
* Can you have intermediate synonyms, perhpas with parameters in the "worong" order?
  ```
  data T a = ... deriving( S b )
  type S b a = (C a, S1 a b)
  type S1 a b = (C1 b a, C2 b a)
  ```

Then the rule would be:
* Start with the `deriving( S x )`
* Add a dummy parameter `a` to get `S x a`
* Expand synonyms recursively
* Flatten tuples
* Check that `a` appears only as the last parameter of each tuple element
* Knock off that `a` from each tuple element and behave as if the user had writtten
  `deriving( C1 x, C2 y, .. )`

All this needs to be spelled out.  Personally I wonder if the gain justifies the pain (where I'm think of documentation and language complexity pain, not implementation pain); but perhaps it does.

----------------------------------

Finally, one might might wonder, rather than a type/constraint *synonym*
```
type S a = (C1 a, C2 a)
```
you can use a type *class* with no methods, and suitable superclasses:
```
class (C1 a, C2 a) => S a where {}
```
Today, if you say
```
data T a = ... deriving( S )
```
you'll get a complaint that there is no instance for `C1 a` and `C2 a`.  So you have to say
```
data T a = ... deriving( S, C1, C2 )
```
which defeats the object of the exercise.  But, thinking aloud, it might be possible to extend `deriving` so that a (`stock`) attempt to derive `S` would also derive its superclasses. That would achieve the effect you desire.

Complication: what if you want to give a *manual* instance for `C1`, not a derived one? We don't want that to clash with the
auto-`derived`.  Now we have to say "unless there is a manual instance for the same type".  But then we get into overlapping instances etc.... Maybe not such a great idea afer all.

VitWW:
@simonpj Thanks for huge reaction!
1) A bit unfair to say what to do if types are unsuitable for derivings or instances. Nothing, an error must appears.

> Note that `C1` and `C2` both have a method `op`; that is fine in Haskell if `C1` and `C2` are declared in different modules. ... Which op belongs with with instance?

2) Yes, but in one module 2 `op` cannot coexists. In this situation user resolves name collision by himself. So we have something like `M1.op` and `M2.op` - no ambiguity

> Check that `a` appears only as the last parameter of each tuple element

3) Maybe I understand this wrong way, but verification of expanded types into classes in heads does NOT belong to renamer, but to a next checker.

> Is itso much worse to say
>  ```haskell
>  instance C1 a => C1 [a] where ...
>  instance C2 a => C2 [a] where ...
>  ```

4) Yes, we cannot write easy way such instances with tuple synonyms. But sometimes we could write similar instance:
```haskell
instance S a => S [a] where ...

-- must expand into
instance (C1 a, C2 a) => C1 [a] where ...
instance (C1 a, C2 a) => C2 [a] where ...
```

> Can you have nested tuples?

5) Yes, this sound reasonable

> Can you have intermediate synonyms, perhaps with parameters in the "wrong" order?

6) Sure, this sound reasonable.

> So these would be illegal
>  ```haskell
>  type S1 a b = (C1 b b, ...)  -- b appears twice
>  type S2 a b = (C1 a, ...)    -- b does not appear at all
>  type S3 a b = (C1 b a, ...)  -- b is not last
>  ```

7) I think you mean, that those types are legal, but using them in deriving head could produce an error. Yes

> Finally, one might might wonder, rather than a type/constraint synonym
> ```haskell
> type S a = (C1 a, C2 a)
> ```
> you can use a type class with no methods, and suitable superclasses:
> ```haskell
> class (C1 a, C2 a) => S a where {}
> ```

8) Such technique is already in use, but it requires additional trivial deriving or instance.

angerman:
@VitWW @uhbif19 are you planning to make any more changes to this proposal? Do you want to make additional changes to clarify what @simonpj raised?

michaelpj:
I think this proposal really needs to be split. I think the part about using synonyms in `deriving` is both better-motivated and less controversial, so deserves to be discussed on its own.

uhbif19:
> are you planning to make any more changes to this proposal?

Yes, I plan to split out deriving sub-proposal and aggregate current discussion in it. Most likely on next week.

angerman:
@uhbif19 I think you can just create a separate proposal for the sub-proposal, and keep this for the rest? That would be my approach, if you think the rest is different enough from this proposal, just open another one and close this.

uhbif19:
Okay, I kinda get convinced that @simonpj approach with extending deriving to handle `class (Eq x, Show x) => Synonym x` is much simpler to explain implementation-wise and would probably write new proposal with it. It seems like covers all recursion/errors questions by existing class-resolving semantics/pragmas.

I still feel that expanding type synonyms have more simple semantics, because synonyms should work like synonyms in any context, but it should be done as part more general discussion/proposal about how synonym expansion is handled generally.

The only question I have is how to specify strategy for such extension. `Synonym x` is empty class, so it seems like it should be extension of `anyclass` strategy to me. But this does not specify with which strategy to derive superclasses. One could have something like `anyclasss by stock` strategy, which seems most clear approach to me, but it complicates syntax. On other hand extending `stock` deriving to such case seems like less clear semantics to me.

uhbif19:
I would probably take a look on implementation and docs for both type synonyms and `anyclass` deriving and relevant extensions (`TypeSynonymInstances`, `LiberalTypeSynonyms`) to write about implementation problems more in depth.

PR #602, opened 2023-07-09
Add a section for CLC involvement to the proposal template

tek:
Motivated by a recent discussion, the idea is to include CLC concerns early in the process and make them a first-class aspect of GHC proposals.

### DIFF
--- a/proposals/0000-template.md
+++ b/proposals/0000-template.md
@@ -57,6 +57,43 @@ give a couple of examples and regard that as the specification! The
 examples should illustrate and elucidate a clearly-articulated
 specification that covers the general case.
 
+## Proposed Library Change Specification
+
+Specify the changes to libraries in the GHC repository, especially `base` and
+others under the purview of the
+[Core Libraries Committee](https://github.com/haskell/core-libraries-committee).
+
+Generally speaking, if your proposal adds new function or data types, the place
+to do so is in the `ghc-experimental` package, whose API is under the control of
+the GHC Steering Committee.
+After your proposal is implemented, stable, and widely used, you (or anyone
+else) can subsequently propose to move those types into `base` via a CLC
+proposal.
+
+Sometimes, however, your proposal necessarily changes something in `base`, whose
+API is curated by the CLC.
+In that case, assuming your proposal is accepted, at the point when it is
+implemented (by you or anyone else), CLC approval will be needed for these
+changes, via a CLC proposal made by the implementor.
+By signalling those changes now, at the proposal stage, the CLC will be alerted
+and have an opportunity to offer feedback, and agreement in principle.
+
+See [GHC base libraries](https://github.com/Ericson2314/tech-proposals/blob/ghc-base-libraries/proposals/accepted/051-ghc-base-libraries.rst?rgh-link-date=2023-07-09T17%3A01%3A15Z)
+for some useful context.
+
+Therefore, in this section:
+
+* If your proposal makes any changes to the API of `base` (including its
+  exports, types, semantics, and performance), please specify these changes
+  in this section.
+
+* If your proposal makes any change to the API of `ghc-experimental`, please
+  also specify these changes.
+
+If you propose to change both, use subsections, so that the changes are clearly
+distinguished.
+Similarly, if any other libraries are affected, please lay it all out here.
+
 ## Examples
 
 This section illustrates the specification through the use of examples of the
@@ -149,4 +186,3 @@ It is not mandatory for have any endorsements at all, but the more substantial
 the proposal is, the more desirable it is to offer evidence that there is
 significant demand from the community.  This section is one way to provide
 such evidence.
-
--- a/proposals/0000-template.rst
+++ b/proposals/0000-template.rst
@@ -74,6 +74,44 @@ give a couple of examples and regard that as the specification! The
 examples should illustrate and elucidate a clearly-articulated
 specification that covers the general case.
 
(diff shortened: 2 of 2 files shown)


proposals/0000-template.md
```diff
@@ -150,3 +150,13 @@ the proposal is, the more desirable it is to offer evidence that there is
 significant demand from the community.  This section is one way to provide
 such evidence.
 
+## Potential CLC Involvement
+
+(Optional) If this proposal includes changes that require approval by the
+[Core Libraries Committee](https://github.com/haskell/core-libraries-committee)
+(most crucially, concerning the API of `base`), the proposer or implementer will
+eventually have to submit an additional proposal to the CLC.
```
googleson78, line 158:
```suggestion
eventually have to submit an additional proposal to the CLC, in order to fully implement their GHC proposal.
```
or
```suggestion
eventually have to submit an additional proposal to the CLC, before they can fully implement their GHC proposal.
```

simonpj:
Thanks for putting this up. But I'd make this more prominent.  Specifically, add a new section after "Proposed change specification", for "Proposed library change specification", with wording something like this

## Proposed library change specification

Please specify changes to libraries, in this section.  

Generally speaking, if your proposal adds new function or data types, the place to do so is in the `ghc-experimental` package, whose API is under the control of the GHC Steering Committee.   After your proposal is implemented, stable, and widely used, you (or anyone else) can subsequently propose to move those types into `base` via a Core Libraries Committee (CLC) proposal.

Sometimes, however, your proposal necessarily changes something in `base`, whose API is curated by the CLC.  In that case, assuming your proposal is accepted, at the point when it is implemented (by you or anyone else), CLC approval will be needed for these changes, via a CLC proposal made by the implementor.  By signalling those changes now, at the proposal stage, the CLC will be alerted and have an opportunity to offer feedback, and agreement in principle.   

See [GHC base libraries](https://github.com/Ericson2314/tech-proposals/blob/ghc-base-libraries/proposals/accepted/051-ghc-base-libraries.rst) for some useful context.

Therefore, in this section:

* If your proposal makes any change to the API of `base` (including its exports, types, semantics, and performance), please specify these changes in this section.

* If your proposal makes any change to the API of `ghc-experimental`, please also specify these changes.  

If you propose to change both, use subections, so that the changes are clearly distinguished.  Similarly, if any other libraries are affected, please lay it all out here.

tek:
@simonpj that sounds great! I adopted it almost verbatim.

simonpj:
Great.   Any comments from other members of the steering committee?

simonpj:
@nomeata shall we merge this?  Just need to update the URL for the base-library plan, which should be https://github.com/Ericson2314/tech-proposals/blob/ghc-base-libraries/proposals/accepted/051-ghc-base-libraries.rst, I think. 

(I would have thought that the canonical link would be HF-only but I don't know how to get a canonical link the rendered plan.)

nomeata:
Arnaud suggested we hold off merging until base-experimental actually exists.

nomeata:
And here is the official link, I assume: https://github.com/haskellfoundation/tech-proposals/blob/main/proposals/accepted/051-ghc-base-libraries.rst

proposals/0000-template.md
```diff
@@ -57,6 +57,43 @@ give a couple of examples and regard that as the specification! The
 examples should illustrate and elucidate a clearly-articulated
 specification that covers the general case.
 
+## Proposed Library Change Specification
```
adamgundry, line 60: I note that the Proposed Change Specification section still has "the types and semantics of any new library interfaces" in the list of things the specification may include. Perhaps the template should make library changes a subsection (or two subsections) of the Proposed Change Specification?

```suggestion
### Library Changes
```

adamgundry (commented):
This looks good to me.

> Arnaud suggested we hold off merging until base-experimental actually exists.

Is it clear when `ghc-experimental` is going to be created? Presumably it depends on someone implementing a proposal that adds to it? I think we should merge something like this sooner rather than later, if necessary referring to `ghc-experimental` as something that will exist in the future.

tek:
> Is it clear when `ghc-experimental` is going to be created? Presumably it depends on someone implementing a proposal that adds to it? I think we should merge something like this sooner rather than later, if necessary referring to `ghc-experimental` as something that will exist in the future.

https://gitlab.haskell.org/ghc/ghc/-/merge_requests/8820 is waiting for `ghc-experimental` to be approved in order to be merged – we could include the creation of the package in that MR, if everyone considered it appropriate to not make this a standalone MR.

simonpj:
@bgamari has a MR in progress that creates `ghc-experimental`.   Ben, where is it at?

bgamari:
> @bgamari has a MR in progress that creates `ghc-experimental`. Ben, where is it at?

See https://gitlab.haskell.org/ghc/ghc/-/merge_requests/10914.

adamgundry:
I see that !10914 has now landed adding `ghc-experimental` to HEAD, so can we move this change forward?

nomeata:
Sure. Should I just merge it?

tek:
any opinions on Adam's suggestion about the subsection?

nomeata:
I can’t make up my mind. I guess for some proposals it’s silly to separate the two sections like this, for others it’s quite natural. @adamgundry, would you care to propose a concrete wording (maybe push directly to this branch), or we can just merge it like this

simonpj:
Let's just merge.   Life is short.  Thanks @nomeata

PR #603, opened 2023-07-11
Amend 475 to use ghc-experimental for exports

tek:
Motivated by recent conversations with the CLC about stability of the API of `base`, we devised the plan to introduce a new package, `ghc-experimental`, that shall become the initial home for exports of new entities that should be tried out by the community before committing to a stable API.

This amendment changes the modules exporting the new tuple- and sum-related entities to use that new package.

I'd appreciate feedback, in particular about whether the creation of this package should be part of the proposal or be handled separately, and whether the deprecation of `ghc-prim:GHC.Tuple` for the rename to `GHC.Internal.Tuple` should be included here.

### DIFF
--- a/proposals/0475-tuple-syntax.rst
+++ b/proposals/0475-tuple-syntax.rst
@@ -80,7 +80,7 @@ Proposed Change Specification
 
 #. Export the following definitions from ``Data.Tuple.Experimental`` and
    ``Prelude.Experimental``. (Implementation note: These would probably end up
-   in a new module ``Data.Tuple.Prim``, due to the dependency from
+   in a new module ``GHC.Tuple.Prim``, due to the dependency from
    the definitions below in ``Data.Tuple.Experimental`` on ``GHC.TypeLits``.) ::
 
      data Unit = ()
--- a/proposals/0475-tuple-syntax.rst
+++ b/proposals/0475-tuple-syntax.rst
@@ -63,30 +63,25 @@ variable in a bit of code. This allows us to e.g. refer to ``Tuple2`` and ``Tupl
 Proposed Change Specification
 -----------------------------
 
-1. Create three new modules in ``ghc-experimental`` called ``GHC.Prelude``,
-   ``GHC.Tuple`` and ``GHC.Sum``.
+1. Create three new modules in ``ghc-experimental`` called ``Prelude.Experimental``,
+   ``Data.Tuple.Experimental`` and ``Data.Sum.Experimental``.
    Since this new paradigm may evolve in response to users' experience, we want
    to avoid committing to an API in ``base`` too early, and therefore choose the
    new package as the home for the exports of the new entities.
 
-   ``GHC.Prelude`` exports definitions peculiar to GHC that are safe to use in
-   ordinary code, but whose API may evolve rapidly over subsequent releases.
+   ``Prelude.Experimental`` exports definitions peculiar to GHC that are safe to
+   use in ordinary code, but whose API may evolve rapidly over subsequent releases.
    This module is *not* intended as a replacement to ``Prelude``, but instead a
    complement to it. This module is a counterpart to ``GHC.Exts``, which exports
    many *unsafe* internals.
 
-   This proposal lists several definitions to be exported from ``GHC.Prelude``, but leaves other
-   definitions to be added in separate proposal(s).
+   This proposal lists several definitions to be exported from ``Prelude.Experimental``,
+   but leaves other definitions to be added in separate proposal(s).
 
-   ``GHC.Tuple`` will clash with the same module from ``ghc-prim``.
-   The plan is to accept that a user of both packages will have to use
-   package-qualified imports for now, and to rename the module in ``ghc-prim``
-   in a separate proposal, while re-exporting the names from ``ghc-prim``'s
-   module in ``ghc-experimental``.
-
-#. Export the following definitions from ``GHC.Tuple`` and ``GHC.Prelude``. (Implementation note:
-   These would probably end up in a new module ``GHC.Tuple.Prim``, due to the dependency from
-   the definitions below in ``GHC.Tuple`` on ``GHC.TypeLits``.) ::
+#. Export the following definitions from ``Data.Tuple.Experimental`` and
+   ``Prelude.Experimental``. (Implementation note: These would probably end up
+   in a new module ``Data.Tuple.Experimental.Prim``, due to the dependency from
+   the definitions below in ``Data.Tuple.Experimental`` on ``GHC.TypeLits``.) ::
 
      data Unit = ()
      data Solo a = MkSolo a    -- this is a change from today's `data Solo a = Solo a`
@@ -106,7 +101,7 @@ Proposed Change Specification
      -- ...
      data Tuple64 ... = (...)
 
-#. Export the following definitions from ``GHC.Tuple`` and ``GHC.Prelude``. These replace
+#. Export the following definitions from ``Data.Tuple.Experimental`` and ``Prelude.Experimental``. These replace
    the existing tuple definitions (in ``GHC.Classes``) today. (Note that ``(...) =>`` is special syntax, and does not
    construct tuples. See more on this point `below <#constraint-special-syntax>`_.)::
 
@@ -127,7 +122,7 @@ Proposed Change Specification
      class (...) => CTuple64 ...
      instance (...) => CTuple64 ...
 
-#. Export the following pseudo-definitions from ``GHC.Tuple``. (Implementation note:
+#. Export the following pseudo-definitions from ``Data.Tuple.Experimental``. (Implementation note:
    These would likely be exported from ``GHC.Prim`` originally.) ::
 
      type Unit# :: TYPE (TupleRep [])
@@ -147,7 +142,7 @@ Proposed Change Specification
      type Tuple64# :: TYPE r1 -> ... -> TYPE r64 -> TYPE (TupleRep [r1, ..., r64])
      data Tuple64# ... = (# ... #)
 
-#. Export the following pseudo-definitions from ``GHC.Sum``. (Implementation note:
+#. Export the following pseudo-definitions from ``Data.Sum.Experimental``. (Implementation note:
    These would likely be exported from ``GHC.Prim`` originally.) ::
 
      -- NB: There are no 0-sums or 1-sums in Haskell, today or tomorrow.
@@ -163,9 +158,9 @@ Proposed Change Specification
      type Sum64# :: TYPE r1 -> ... -> TYPE r64 -> TYPE (SumRep [r1, ..., r64])
      data Sum64# ... = ...
 
-#. Remove existing tuple definitions from ``GHC.Tuple`` in ``ghc-prim``.
+#. Remove existing tuple definitions from ``Data.Tuple.Experimental`` in ``ghc-prim``.
 
-#. Export the following definitions from ``GHC.Tuple``::
+#. Export the following definitions from ``Data.Tuple.Experimental``::
 
      type Length :: List a -> Nat   -- not exported
      type family Length xs where
@@ -239,7 +234,7 @@ Proposed Change Specification
 
      data List a = [] | a : List a
 
-#. Re-export ``List`` from ``GHC.List`` and ``GHC.Prelude``.
+#. Re-export ``List`` from ``GHC.List`` and ``Prelude.Experimental``.
 
 #. Introduce a new extension ``-XListTuplePuns``; this extension is part
    of ``-XHaskell98``, ``-XHaskell2010``, and ``-XGHC2021``. It is thus on by default.
@@ -252,16 +247,16 @@ Proposed Change Specification
    #. An occurrence of ``[ty]`` in type-syntax is a synonym for ``GHC.List.List ty``.
 
    #. An occurrence of ``()`` in type-syntax, where the type is not expected to be of kind ``Constraint``,
-      is a synonym for ``GHC.Tuple.Unit``.
+      is a synonym for ``Data.Tuple.Experimental.Unit``.
 
    #. An occurrence of ``(,,...,,)`` where there are *n* commas (for *n* ≧ 1) in type-syntax
-      is a synonym for ``GHC.Tuple.Tuple<n+1>``.
+      is a synonym for ``Data.Tuple.Experimental.Tuple<n+1>``.
 
    #. An occurrence of ``(ty1,ty2,...,ty<n-1>,ty<n>)`` (for *n* ≧ 2) in type-syntax, where neither the type
       is expected to be of kind ``Constraint`` and either none of the ``ty<i>`` are inferred to have kind ``Constraint``
       or there exists a ``ty<i>`` inferred to kind ``Type`` and none of the ``ty<j>`` (with *j* < i) are inferred to have
       kind ``Constraint``, is
-      a synonym for ``GHC.Tuple.Tuple<n>`` ``ty1 ty2 ... ty<n-1> ty<n>``. (This rule retains today's behavior.)
+      a synonym for ``Data.Tuple.Experimental.Tuple<n>`` ``ty1 ty2 ... ty<n-1> ty<n>``. (This rule retains today's behavior.)
 
    #. With ``-XUnboxedTuples``, an occurrence of ``(# #)`` in type-syntax is a synonym for ``GHC.Exts.Unit#``.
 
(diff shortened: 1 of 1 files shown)


Ericson2314 (approved):

simonpj:
Looks good to me.

proposals/0475-tuple-syntax.rst
```diff
@@ -63,15 +63,27 @@ variable in a bit of code. This allows us to e.g. refer to ``Tuple2`` and ``Tupl
...
+   ``GHC.Prelude`` exports definitions peculiar to GHC that are safe to use in
+   ordinary code, but whose API may evolve rapidly over subsequent releases.
+   This module is *not* intended as a replacement to ``Prelude``, but instead a
+   complement to it. This module is a counterpart to ``GHC.Exts``, which exports
    many *unsafe* internals.
 
    This proposal lists several definitions to be exported from ``GHC.Prelude``, but leaves other
    definitions to be added in separate proposal(s).
 
+   ``GHC.Tuple`` will clash with the same module from ``ghc-prim``.
```
int-index, line 81: Can we avoid the clashes? Nothing compels us to create a conflicting `GHC.Tuple` in `ghc-experimental`, at least not for the sake of this proposal. Just leave it in `ghc-prim`?

The new `GHC.Sum` and `GHC.Prelude` can go to `ghc-experimental`, of course, as there is no conflict.

proposals/0475-tuple-syntax.rst
```diff
@@ -63,15 +63,27 @@ variable in a bit of code. This allows us to e.g. refer to ``Tuple2`` and ``Tupl
 Proposed Change Specification
 -----------------------------
 
-1. Create a new module in ``base`` called ``GHC.Prelude``. This module exports definitions
-   peculiar to GHC that are safe to use in ordinary code.
-   This module is *not* intended as a replacement to ``Prelude``,
-   but instead a complement to it. This module is a counterpart to ``GHC.Exts``, which exports
+1. Create three new modules in ``ghc-experimental`` called ``GHC.Prelude``,
```
int-index, line 66: `GHC.Prelude` is actually taken by the GHC API for the Prelude used by GHC itself. Are we absolutely sure we want to repurpose this name, when we could simply pick a different one? Need opinion from @goldfirere

proposals/0475-tuple-syntax.rst
simonpj: > Can we avoid the clashes? Nothing compels us to create a conflicting GHC.Tuple in ghc-experimental, at least not for the sake of this proposal. Just leave it in ghc-prim?

Good qn.  But if we leave it in `ghc-prim` then clients of the new tuple stuff will be force to depend on `ghc-prim` which we do not want.  Rather, I think we should transition `ghc-prim:GHC.Tuple` to become `ghc-prim:GHC.Internals.Tuple`.  Quite how fast we do that isn't clear to me.  

@goldfirere ?

proposals/0475-tuple-syntax.rst
simonpj: Huh -- that is annoying.

It demonstrates that we are using the module prefix "GHC." in two different ways:
* Modules of GHC itself (e.g. GHC.Core.Opt.Simplify)
* Modules in `ghc-experimental`.
These are not the same purpose.

Maybe `ghc-experimental` should start with `GHCX.`?  Thus `GHCX.Prelude`, `GHCX.Tuple`.   Or some other prefix?

int-index: `GHCX.*` is just a slightly different take on `GHC.Exts.*` that @tek previously suggested, right? I believe `GHC.Exts.*` is good and descriptive. For example, `GHC.Exts.Tuple` is "GHC's extensions to tuples". When those get stabilized, they can be reexported from `Data.Tuple`

JakobBruenker: I wonder about the broader situation here - would the idea be that all `GHC.Exts.*` modules end in non-`GHC.Exts.*` modules if/when brought into `base`? Since if anything is added to `base` *also* under the `GHC.Exts.*` moniker, you would still end up with a naming conflict during the deprecation period from `ghc-experimental`

proposals/0475-tuple-syntax.rst
tek: our new `GHC.Tuple` depends on `base` (it needs `TypeNats` and possible `TypeError`), so leaving it in `ghc-prim` is a no-go.

proposals/0475-tuple-syntax.rst
tek: `GHC.Exts` exposes some unsafe operations, as in "crash the process" unsafe, which the original text uses as contrast for the purpose of `GHC.Prelude`…would calling the new namespace `GHC.Exts.*` associate `ghc-experimental` with that notion?

On the other hand, `GHC.Exts` might be a candidate for migration to `ghc-internals`.

There is also a large number of `GHC.*` modules in `base` that might be a source for additional conflicts later.

nomeata: I find the prefix `GHC.` for things in ghc-experimental strange, as they are neither about GHC (e.g. a plugins) or part of GHC (e.g. GHC.Prelude). It's as if I start to prefix my library modules with `Nomeata.`. I expect developers to find this naming confusing (me already included.)

How about we append `Experimental` to the module name from base where things might end up in? `Data.Tuple.Experimental`?

Ericson2314: The only thing is should GHC's experiments have a monopoly on such names?

`Data.Tuple.GlasgowExperiment` lol

What it is comes first, "compilation" is not relevant and thus unmentioned, who experiment and so not to conflict with other experiments comes last.

simonpj:
> How about we append Experimental to the module name from base where things might end up in? Data.Tuple.Experimental?

Or prefix it, so that `ghc-experimental` modules are in the `Experimental.*` name space.  Thus `Experimental.Data.Tuple`.

(I can't speak for others but I'm quite open to renaming the library itself too, incidentally, but just `experimental` sounds a bit too generic.)

numlit2:
> Or prefix it, so that `ghc-experimental` modules are in the `Experimental.*` name space. Thus `Experimental.Data.Tuple`.

`Experimental.*` is very good. I know I will import something unstable. `GHC.Exts.*`doesn't tell me anything.

adamgundry:
This somewhat overlaps with the discussion at https://github.com/haskellfoundation/tech-proposals/pull/53.

I would suggest we use `Prelude.Experimental`, `Data.Tuple.Experimental` and `Data.Sum.Experimental`. Moreover we could have `Prelude.Experimental` export `Type` (cf. https://github.com/haskell/core-libraries-committee/issues/193).

tek:
That seems like a reasonable schema to me, and it saves us the discussion about `GHC.Prelude`.

But exporting `Type` from there sounds like it would become its permanent home – the problem articulated in the other discussion isn't that `Type` isn't available in `base`, but that it isn't implicitly in scope, no?

adamgundry:
> the problem articulated in the other discussion isn't that `Type` isn't available in `base`, but that it isn't implicitly in scope, no?

Indeed, so the benefits are limited. I was just thinking about what else `Prelude.Experimental` could reasonably export, as a "staging ground" for things that may be added to `Prelude` later. It's fine if you prefer to leave that discussion for a separate proposal.

tek:
MR now reflects these changes: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/8820/diffs?commit_id=9be5e1645df92e217a2ce5f806940cba016277f7

proposals/0475-tuple-syntax.rst
tek: now called `Prelude.Experimental`.

tek:
@nomeata would you submit, please?

nomeata:
I am not sure if this needs the full committee process. Isn’t it the only reasonable choice given the (accepted) plan around the use of ghc-experimental?

If @simonpj or @adamgundry could sanity-check the diff I’m inclined to just merge as it.

adamgundry (commented):
Thank you for pushing this on @tek! I have a few minor nitpicks but I agree that a full committee review seems unnecessary given that most of this is essentially predetermined.

I'm wondering whether it helps or hinders to give all the fully-qualified module names referring to modules in `ghc-prim`. Arguably, where things live in `ghc-prim`/`ghc-internals` are implementation details that needn't be part of the specification, provided we are clear about what is being exposed from `base` (nothing?) and `ghc-experimental`. But I'll leave this up to your preference.

proposals/0475-tuple-syntax.rst
```diff
@@ -63,18 +63,25 @@ variable in a bit of code. This allows us to e.g. refer to ``Tuple2`` and ``Tupl
...
-   definitions to be added in separate proposal(s).
+   This proposal lists several definitions to be exported from ``Prelude.Experimental``,
+   but leaves other definitions to be added in separate proposal(s).
 
-#. Export the following definitions from ``GHC.Tuple`` and ``GHC.Prelude``. (Implementation note:
-   These would probably end up in a new module ``GHC.Tuple.Prim``, due to the dependency from
-   the definitions below in ``GHC.Tuple`` on ``GHC.TypeLits``.) ::
+#. Export the following definitions from ``Data.Tuple.Experimental`` and
+   ``Prelude.Experimental``. (Implementation note: These would probably end up
+   in a new module ``Data.Tuple.Prim``, due to the dependency from
```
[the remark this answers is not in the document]
tek: ah, now that we use `Data.Tuple.Experimental` instead of `GHC.Tuple` (which was supposed to move to `base`), we could probably use the latter for the new definitions instead of `GHC.Tuple.Prim`. I'll change that and see if it creates any issues.

proposals/0475-tuple-syntax.rst
```diff
@@ -265,41 +277,41 @@ Proposed Change Specification
...
 
-   #. An occurrence of ``GHC.Tuple.Tuple<n> ty1 ty2 ... ty<n>`` is pretty-printed as ``(ty1, ty2, ..., ty<n>)``.
+   #. An occurrence of ``GHC.Tuple.Prim.Tuple<n> ty1 ty2 ... ty<n>`` is pretty-printed as ``(ty1, ty2, ..., ty<n>)``.
 
-   #. An occurrence of ``GHC.Tuple.Tuple<n>``, but not applied to a full *n* arguments, is pretty-printed as ``(,,...,,)``,
+   #. An occurrence of ``GHC.Tuple.Prim.Tuple<n>``, but not applied to a full *n* arguments, is pretty-printed as ``(,,...,,)``,
       where there are *n-1* commas.
 
-   #. An occurrence of ``GHC.Exts.Unit#`` is pretty-printed as ``(# #)``.
+   #. An occurrence of ``GHC.Types.Unit#`` is pretty-printed as ``(# #)``.
```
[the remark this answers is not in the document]
tek: those should all have been changed to `GHC.Types`, I overlooked a few occurrences.
`GHC.Prim` is a virtual module, so my assumption is that Richard hadn't determined the precise module yet at the point of writing it.

The straightforward change would have been to use `Data.Tuple.Experimental` instead of `GHC.Exts` – I used `GHC.Types` because I now know what we landed on eventually…do you think that's too much retconning? We could also change the other sentence to say "would likely be exported from `GHC.Types` originally".

proposals/0475-tuple-syntax.rst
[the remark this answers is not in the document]
tek: ✅

proposals/0475-tuple-syntax.rst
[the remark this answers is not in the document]
tek: ah no, there's no equivalent of `GHC.Exts` with the new schema…

simonpj:
Please could the propsal say clearly
1. What are the changes to the API of `base`?
2. What are the changes to the API of `ghc-experimental`?
3. What other changes to ghc-internal packages are there?  E.g. to the union of `ghc-internal` and `ghc-prim`?  True, this doesn't need to be part of the specification, since it's an implementation matter, and it should be explained as such.

Currently all these changes are mixed together, and only module names are given, not packages, so I can't work out the answer.  Could you simply summarise in one place?

I think there is quite a bit of (1) changes, and they need agreement from CLC.  E.g. changes to `GHC.List`.

I notice mention of `GHC.Tuple.Prim`.  What package is that in?  And I thought we'd agreed that ghc-internal modules would usually look like `GHC.Internal.Tuple`?

adamgundry:
> I think there is quite a bit of (1) changes, and they need agreement from CLC. E.g. changes to `GHC.List`.

Good point. I was assuming there would be no `base` changes, which I think would be the case if point 10 was adapted so that `GHC.List` would not re-export `List`? (And the references to `GHC.List.List` changed to `GHC.Types.List`.)

simonpj:
I think that, more than just edinting point 10, it would help the reader a lot if there was a brief summy of

- Changes to ghc-experimental
- Changes to base

The existing structure is fine; just a summary is needed so that the reader can quickly get that overview.

tek:
I used the libraries section we added to the template last week. Other than `List`, there's no change to base.

proposals/0475-tuple-syntax.rst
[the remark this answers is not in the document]
tek: ah, but `GHC.Tuple.Prim` is already in master due to an older MR that implements only a small part of the proposal – should we discard it anyway and break the API of `ghc-prim`?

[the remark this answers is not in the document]
tek: @adamgundry what do you think?

proposals/0475-tuple-syntax.rst
```diff
@@ -94,8 +102,9 @@ Proposed Change Specification
      -- ...
      data Tuple64 ... = (...)
 
-#. Export the following definitions from ``GHC.Exts`` and ``GHC.Prelude``. These replace
-   the existing tuple definitions (in ``GHC.Classes``) today. (Note that ``(...) =>`` is special syntax, and does not
+#. Replace the existing constraint tuples in ``GHC.Classes`` with the following
+   definitions and export them from ``Data.Tuple.Experimental`` and ``Prelude.Experimental``. ::
```
adamgundry, line 106:
```suggestion
   definitions and export them from ``Data.Tuple.Experimental`` and ``Prelude.Experimental``.
```

proposals/0475-tuple-syntax.rst
```diff
@@ -328,6 +340,24 @@ Proposed Change Specification
...
+
+1. ``base``:
+   - ``GHC.List`` will export ``GHC.Types.List``.
+     The proposal and the implementation of the ``List`` part predate the
+     stability practice of introducing experimental entities in ``ghc-experimental``.
+
+#. ``ghc-experimental``:
+   - ``Data.Tuple.Experimental`` will export all tuple-related entities.
+   - ``Data.Sum.Experimental`` will export all sum-related entities.
+   - ``Prelude.Experimental`` will export all of the above.
```
adamgundry, line 354: The RST list syntax is somehow not quite right here.

proposals/0475-tuple-syntax.rst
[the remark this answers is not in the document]
adamgundry: I think `ghc-prim` has no stability guarantees, so you should feel free to remove `GHC.Tuple.Prim` unless there is some implementation reason to keep it.

adamgundry (commented):
Apart from a couple of RST syntax nits this looks good to me, thanks @tek!

@simonpj are you happy with the revised state? AFAICS there is no need for further CLC consultation as the only change to `base` is the export of `List` from `GHC.List`, which has already been implemented.

proposals/0475-tuple-syntax.rst
```diff
@@ -63,18 +63,26 @@ variable in a bit of code. This allows us to e.g. refer to ``Tuple2`` and ``Tupl
...
+   This module is *not* intended as a replacement to ``Prelude``, but instead a
+   complement to it. This module is a counterpart to ``GHC.Exts``, which exports
    many *unsafe* internals.
 
-   This proposal lists several definitions to be exported from ``GHC.Prelude``, but leaves other
-   definitions to be added in separate proposal(s).
+   This proposal lists several definitions to be exported from ``Prelude.Experimental``,
+   but leaves other definitions to be added in separate proposal(s).
+
+   All of the wired-in types must be declared in ``ghc-prim``, and re-exported from
```
simonpj, line 81: This isn't part of the specification -- it's an implementation note.  Mabye
```
Implementation note: many type definitions will be "wired-in", defined in `ghc-prim`, and simply re-exported from `ghc-experimental`.

File: proposals/0475-tuple-syntax.rst [hunk 8]
```diff
@@ -63,18 +63,26 @@ variable in a bit of code. This allows us to e.g. refer to ``Tuple2`` and ``Tupl
...
+   ``ghc-experimental``.
 
-#. Export the following definitions from ``GHC.Tuple`` and ``GHC.Prelude``. (Implementation note:
-   These would probably end up in a new module ``GHC.Tuple.Prim``, due to the dependency from
-   the definitions below in ``GHC.Tuple`` on ``GHC.TypeLits``.) ::
+#. Replace the existing tuples in ``GHC.Tuple`` with the following definitions
+   and export them from ``Data.Tuple.Experimental`` and ``Prelude.Experimental``. ::
```
simonpj, line 85: I would not talk about replacing anything.  That's an implementation matter.  Perhaps just
```
Export the following type definitions from `Data.Tuple.Experimental`.

File: proposals/0475-tuple-syntax.rst [hunk 9]
```diff
@@ -63,18 +63,26 @@ variable in a bit of code. This allows us to e.g. refer to ``Tuple2`` and ``Tupl
...
+   ``Data.Tuple.Experimental`` and ``Data.Sum.Experimental``.
+   Since this new paradigm may evolve in response to users' experience, we want
+   to avoid committing to an API in ``base`` too early, and therefore choose the
+   new package as the home for the exports of the new entities.
+
+   ``Prelude.Experimental`` exports definitions peculiar to GHC that are safe to
+   use in ordinary code, but whose API may evolve rapidly over subsequent releases.
+   This module is *not* intended as a replacement to ``Prelude``, but instead a
+   complement to it. This module is a counterpart to ``GHC.Exts``, which exports
    many *unsafe* internals.
```
simonpj, line 76: Perhpas add
```
`Prelude.Experimental` simple re-exports the all the exports of `Data.Tuple.Experimental` and `Data.Sum.Experimental`, thus
   module Prelude.Experimental (
       module Data.Tuple.Experimental,
       module Data.Sum.Experimental
    ) where 
     import Data.Tuple.Experimental
    import Data.Sum.Experimental

Now you don't need to talk about `Prelude.Experimental` any futher.

In fact you could specify `Prelude.Experimental` as a separate bullet.

File: proposals/0475-tuple-syntax.rst (hunk 5 above)
Comment (simonpj) [inline proposals/0475-tuple-syntax.rst:106]: Same comment.  Don't talk about replacing or about `GHC.Classes`.  Those are implementation matters. Simply write down the definitions visible through `Data.Tuple.Experimental`.

**Review by simonpj — COMMENTED** [review]
Generally fine, thank you.  Just cosmetic suggestions.

**Comment by simonpj** [comment]
> @simonpj are you happy with the revised state? AFAICS there is no need for further CLC consultation as the only change to base is the export of List from GHC.List, which has already been implemented.

I agree we don't need to *consult* but I think it would be polite to send an email to the CLC mailing list, drawing their attention to these changes.

**Comment by tek** [comment]
@adamgundry @simonpj I believe I have implemented all your suggestions, and sent an email to the CLC.

File: proposals/0475-tuple-syntax.rst [hunk 10]
```diff
@@ -63,18 +63,35 @@ variable in a bit of code. This allows us to e.g. refer to ``Tuple2`` and ``Tupl
...
-   This module is *not* intended as a replacement to ``Prelude``,
-   but instead a complement to it. This module is a counterpart to ``GHC.Exts``, which exports
+1. Create two new modules in ``ghc-experimental`` called ``Data.Tuple.Experimental``
+   and ``Data.Sum.Experimental``.
+   Since this new paradigm may evolve in response to users' experience, we want
+   to avoid committing to an API in ``base`` too early, and therefore choose the
+   new package as the home for the exports of the new entities.
+
+   Implementation note: Many type definitions will be "wired-in", defined in ``ghc-prim``,
+   and simply re-exported from `ghc-experimental`.
```
adamgundry, line 73:
```suggestion
   and simply re-exported from ``ghc-experimental``.
```

adamgundry (approved):

nomeata:
@adamgundry , shall I merge?

tomjaguarpaw:
There's something I don't understand here. Apparently there is one change to `base`: "`GHC.List` will export `GHC.Types.List`". But `GHC.List` [already exports](https://hackage.haskell.org/package/base-4.18.0.0/docs/GHC-List.html#t:List) `GHC.Types.List` so I don't understand why that's part of the proposal. Perhaps it's supposed to be explained by the following text

> The proposal and the implementation of the ``List`` part predate the stability practice of introducing experimental entities in ``ghc-experimental``.

but I don't understand what "the ``List`` part" means or why that should explain this particular aspect of the proposal.

tomjaguarpaw:
(Please don't hold up moving forward with this proposal on my account. It's probably just me being dim or having forgotten something I knew about before. But the CLC was explicitly informed of this proposal and this CLC member doesn't understand the one small part that seems to impinge on the CLC.)

tek:
@tomjaguarpaw that's just to say that the part of the proposal that introduces `List` and exports it from `base` was already merged to master and released before we started the split-base discussion and decided to refrain from exporting these new entities from `base`.

tomjaguarpaw:
So #475 was partially implemented and now you're revising it so that the rest of the implementation uses `ghc-experimental` instead?

tek:
@tomjaguarpaw that's correct

tomjaguarpaw:
Great, thanks for clarifying!

tek:
small tweak: mentioned `Solo` deprecation explicitly in the library changes section.

adamgundry:
> @adamgundry , shall I merge?

LGTM, yes please!

PR #605, opened 2023-07-16
Amend 281 visible forall to work without ScopedTypeVariables

int-index:
This amendment removes the requirement to enable the `ScopedTypeVariables` extension to write code of the following form:

```haskell
f :: forall x -> Show x => x -> String
f (type t) = show @t
```

Here `type t` binds a type variable named `t`, which is then used on the right-hand side in a type application `@t`. This entire arrangement already relies on `ExplicitNamespaces` to use the `type` namespace specifier and `RequiredTypeArguments` to typecheck it against the `forall x ->`. There is no need to require `ScopedTypeVariables` on top of that, especially since we are trying to move away from it (#448)

### DIFF
--- a/proposals/0281-visible-forall.rst
+++ b/proposals/0281-visible-forall.rst
@@ -343,8 +343,6 @@ Part I: Proposed Change Specification
    follows type-level name resolution rules (i.e. uses of punned identifiers
    resolve to the type namespace), both at binding sites and at use sites.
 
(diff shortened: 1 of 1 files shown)


Ericson2314 (approved):
Good catch!

simonpj:
I support this.

int-index:

PR #608, opened 2023-08-03
Amend #448 so `-XPatternSignatures` does not include implicit bindings

Ericson2314:
*2nd in a series of #448 amendments, following up from #604, building towards #532*

This proposal scales back #448 's reintroduced `-XPatternSignatures` so that it does not include pattern signature bindings. Pattern signature binds have multiple drawbacks that pattern signatures themselves do not have, so it makes sense to want them to be enabled separately. Pattern signature bindings are instead are enabled by `-XScopedTypeVariables`, which is unchanged.

This unblocks https://gitlab.haskell.org/ghc/ghc/-/merge_requests/10385 by @s-and-witch. Unlike #604 this does not effect GHC 9.8, so it isn't so critically urgent to resolve.

_(Instead of a rendered link, go to the Files Changed tab and toggle the rich text diff option. Rending the whole proposal is not that useful when this is a small change to a much larger proposal.)_


nomeata:
This would essentially be https://github.com/ghc-proposals/ghc-proposals/pull/119, right?

Ericson2314:
Yes!

AntC2:
> (Rending the whole proposal is not that useful when this is a small change to a much larger proposal.)

I agree, so thank you for that. Never the less I'm having a hard time reconnecting all the dots to follow which particular convoluted position we've reached. Bear with me ...

>        data T = MkT [a] (a->Int)
>        f :: T -> [Int]
>        f (MkT (xs :: [a]) f) = let mf :: [a] -> Int
>                                    mf = map f
>                                in mf xs

This example appears to be ill-typed. (I'm being pedantic, but since this is one in a series of mods to a mod to a reintroduced mod, I feel pedantry is the point of the exercise.) I think we should have

> ... let mf :: [a] -> [Int]               -- result from map is still a list

Also I'm not happy with that form of `data` decl. tyvar `a` is existential(?) So shouldn't there be an explicit `forall a. ...` on rhs of the decl?

> Here the pattern signature on ``xs`` brings ``a`` into scope, ...

Yes, but this never caused me conceptual difficulties. It's only with `ScopedTypeVariables` in force that you have to nervously look around for a signature somewhere that might already have introduced `a`.  Without that extension, you can be confident the leftmost appearance in the equation for `f` is what's introducing the tyvar.

Ericson2314:
The type errors are now fixed. Thanks for pointing them out to me. They are original in https://github.com/ghc-proposals/ghc-proposals/pull/523#issuecomment-1346449731 but i hope @simonpj doesn't mind editorializing the quote to fix them.

> Yes, but this never caused me conceptual difficulties. It's only with `ScopedTypeVariables` in force that you have to nervously look around for a signature somewhere that might already have introduced `a`. Without that extension, you can be confident the leftmost appearance in the equation for `f` is what's introducing the tyvar.

What should:
```haskell
f (x :: a) = undefined
  where
    g (y :: a) = undefined
```
With `PatternSignatures` but without `ScopedTyepVariables` do?

In my opinion, the interpretation in this amended proposal is the only one that is both unsurprising and doesn't require "nervously looking around" for outer binders, as you put it well.

Ericson2314:
@nomeata I would like to submit to the committee!

AntC2:
> What should:
> ...
> With PatternSignatures but without ScopedTypeVariables do?

Good question! And I agree needs a little 'looking around' -- but only a little. I'll try a variation

```
ff (x :: a) = g (x, x) 
      where
        g ( x@(_, y) :: b) = y  
```

This compiles, infers `ff :: a -> a`. `x` appears in two different binding sites at two different types. But we know those are binding sites syntactically, so that's two different `x`s. `b` doesn't appear anywhere leftwards so that must be a binding. If I change `g`'s `:: b` to `:: a`, GHC complains `Occurs check ... infinite type ...`. So that `a` isn't appearing leftmost/can't be a binding site.

Now try (in a Haskell that supports not only `PatternSignatures` with binds but also `ResultSignatures` with binds):

```
ff (x ) :: a = g (x, x)     -- this `:: a` is a ResultSignature (binding)
      where
        g ( x@(_, y :: a) ) :: b = (y :: b) :: a 
```

This compiles, infers `ff :: a -> a`. My 'leftmost appearance' rule seems to be working fine.

I can change that top line to `ff (x :: a) :: a = g (x, x) ...`. Still compiles happily. I can change the `:: b` on `g`'s result to `:: a`. Still happy.

So what worries me with this proposal is you're taking away the ability for `PatternSignatures` to be binding sites. And forcing me to use the uglifying and complexifying `@a` mechanism.

Complexifying because why are you forcing me to distinguish a binding from a usage in _types_? Types don't work like terms. Types work by unification not evaluation (beta reduction). Unification means if I have an `:: a` here, I have the same `:: a` everywhere in that scope.

AntC2:
(This is probably a q for some other proposal in this whirl but ...)

Are we sure in patterns we can distinguish `@` in as-patterns vs `@` for tyvar introduction? Note `let foo@a = 42 in ...` is valid (but admittedly silly).

AntC2:
Another way to mentalise `PatternSignatures` is: I want everything marked `:: a` to be the same type. I don't care how you get there: you're a compiler; that's your job.

Ericson2314:
>Types work by unification not evaluation (beta reduction).

I don't think that is true (there is a mix of unification and special-cased "bidirectional typechecking" methods going on). Moreover, this philosophy --- types are not like terms, the scoping / name resolution of type need not be explicit / types should not be thought of as substitution is decidedly *not* the philosophy of #448. I don't expect you to start liking this stuff when you didn't before, but I also don't think an amendment proposal like this one is the right place to re-litigate these issues.

As an aside, let me add that "types working like beta reduction" isn't just some fancy Martin Löf thing, it is also a System F thing. My strong desire to see parsing and renaming for types and terms work the same way can thus be departed from any Dependent Haskell notions.

proposals/0448-type-variable-scoping.rst
```diff
@@ -527,6 +622,7 @@ The type variable ``a`` is bound to ``Int``, by pattern-matching.
 
 Here is an example of pattern signatures within a type abstraction in a pattern::
 
+   {-# LANGUAGE ScopedTypeVariables #-} -- for pattern signture bindings
```
adamgundry, line 625:
```suggestion
   {-# LANGUAGE ScopedTypeVariables #-} -- for pattern signature bindings
```

proposals/0448-type-variable-scoping.rst
```diff
@@ -540,6 +636,8 @@ multiple bindings of a single variable::
 Pattern and kind signatures, however, are not subject to this restriction,
 since variable occurrences in pattern signatures are considered usages (not bindings)::
 
+   {-# LANGUAGE ScopedTypeVariables #-} -- for pattern signture bindings
```
adamgundry, line 639:
```suggestion
   {-# LANGUAGE ScopedTypeVariables #-} -- for pattern signature bindings
```

adamgundry (commented):
Thanks very much for this @Ericson2314! I supported this idea [five years ago](https://github.com/ghc-proposals/ghc-proposals/pull/119#issuecomment-380591618) and I'm still strongly in favour.

Ericson2314:
@goldfirere Had a chance to read this one yet? (Sorry I didn't think to mention in person at ICFP.)

I recall, hopefully correctly, from the other threads that you also wanted this, but were worried others on the steering commit did not, but then they in fact were OK with this.

goldfirere:
Sorry for losing track of this one. I'm in support of having a mechanism to forbid binding type variables in pattern signatures, but I disagree with the specifics here: with this amendment, `-XScopedTypeVariables` would imply `-XPatternSignatures`, `-XMethodTypeVariables`, and `-XExtendedForAllScope`, and the ability to bind variables in pattern signatures.

And, as far as I can see, the behavior specified in this proposal is already available, via `-XPatternSignatures -Werror=pattern-signature-binds`. What's not good enough about `-XPatternSignatures -Werror=pattern-signature-binds`? I continue to think that we abuse our extensions mechanism to perform duties better handled by the warning mechanism or a linter.

To be clear, I agree that we should support Haskellers who want to prevent their pattern signatures from binding variables. But I argue that the existing #448 already accomplishes this goal and would need to be convinced that the status quo is somehow insufficient before recommending acceptance.

(I don't mean to just unilaterally block this proposal, and so I'm happy to forward to the committee with a "reject" recommendation; perhaps others will disagree with me.)

simonpj:
>
> I'm in support of having a mechanism to forbid binding type variables in
> pattern signatures, but I disagree with the specifics here
>

Maybe you can put that text on the proposal's discussion thread, with your
shepherd hat on, inviting the author to respond?

Simon

Ericson2314:
@goldfirere The most notable is the interaction with other features: For example, having contextual bindings like this breaks `-Wpuns` and `-Wpun-values`'s ability to simulate a single namespace:

```haskell
x = 1

f @x = 1 -- is this a bind or a use? Can't change based on warnings alone!
```

This an other interactions was detailed in a the much larger https://github.com/ghc-proposals/ghc-proposals/pull/532, but I had pulled them out because I thought this change was uncontroversial. I suppose I can add them back if you like.

goldfirere:
I'm afraid I still don't get it.

Even with the proposed amendment, the `-XScopedTypeVariables` language -- where `(pat :: a)` might bind `a` or be a usage of `a`, depending on whether `a` is in scope -- will still exist. You perhaps won't use it, but it will exist. So a future proposal that blends namespaces will have to have *some* opinion on what to do with `let a = ... in ... (pat :: a) ...`. Will that proposal be fork-like? Or, put more directly, would that future proposal introduce an extension that changes the semantics of existing programs? I think it will have to.

Already, we're saying that you're more future-compatible with dependent types if you have `-Wpuns` and `-Wpun-bindings`. Maybe we just add `-Wpattern-signature-binds` to that list. I just don't see the practical difference between putting the feature behind a warning vs an extension -- other than the fact that warnings strike me as lighter-weight, and thus preferable (other things being equal).

Note that the situation with `f @a = ...` is different from what we're debating here, because `f @a` should unambiguously bind `a` (shadowing any other `a` in scope) -- and thus it would be problematic if `f @a = ...` could be an occurrence of an in-scope `a`. We've debated that point before, and I agree a warning is insufficient there. For better or worse, this case here, in a signature, is different.

Ericson2314:
@goldfire Sorry let me be more explicit:

1. The issue is that the `x` is my example will be considered a bind, and get a warning, because pattern signature binds only considers the type namespace (absent new unproposed language extensions) as doing otherwise would be a breaking change.

2. That leaves the single namespace user *unable* to use the term-namespace-bound `x` in a pattern signature (Because the implicit bind gets in the way).

3. Beyond the lack of expressivity, this might be highly confusing to a user taught with single namespaces and hitherto unaware that there was even such a thing as type vs term namespaces.

Ericson2314:
The extension version of `-Wpattern-signature-binds`, this, fixes being able to use variables in pattern signatures regardless of the namespace they are bound in. The hypothetical extension version of `-Wpuns` and `-Wpun-bindings` would "fix" shadowing, allowing variables bound in a narrower scope in the "wrong" namespace to take shadowing. There is no way to fix either of these things (from a single namespace perspective) without extensions.

I think most people agree shadowing is of dubious utility (unless someone is trying to teach how substitution works), so leaving `-Wpuns` and `-Wpun-bindings` as warnings seems fine to me. In the other and hand being unable to use variables bound in the other namespace *at all* in pattern signatures strikes me as very frustrating limitation I could not justify to a student/coworker. (And I would like to use and teach single-namespace Haskell regardless of dependent features because I consider it more human-friendly.)

goldfirere:
I am a huge fan of single-namespace Haskell. But despite your kind re-explanations, I still don't see how this proposal gets us closer to that goal.

You say "the single namespace user [is] unable to use the term-namespace-bound x in a pattern signature (because the implicit bind gets in the way)." How does this proposal change that?

It sounds to me as if you're imagining some future `-XUnifiedNamespace` extension and further imagining particular interactions with this proposal. I'm in support of designing today in order to make room for future extensions. Yet this proposal does not, say, remove the `-XScopedTypeVariables`-powered behavior relating to binding. So, in some sense, the future proposal will have to grapple with backward incompatibility. So I don't see how the current proposal moves the needle on that potential future conversation. Maybe if `-XScopedTypeVariables` were a niche or newly-introduced extension, I might feel differently. But it is pervasive, meaning that a future proposal that disagrees with `-XScopedTypeVariables` is going to be problematic. (Not impossible -- but it will be a lively debate.)

Ericson2314:
Today there are two choices:

1. No pattern signatures at all
2. Pattern signatures with implicit binds, which, regardless of warning, will not allow uses of term-namespace-bound variables.

The point of this proposal is to take the opportunity of the newly reintroduced, yet-unimplemented `-XPatternSignatures` for pattern signatures without pattern binds. That directly solves the problem:

- Term bound `x` can now be used in a pattern signature, with `-XPatternSignatures` and without `-XScopedTypeVariables`.

I am not imagining a `-XUnifiedNamespace` future because @int-index's most recent plan is that we do not need a such a thing after all; that `-Wpuns` and `-Wpun-bindings` are sufficient.

I don't think we need to do anything about about `-XScopedTypeVariables`? #448 was already about giving users other options in hopes that they migrate away from it, and this doesn't change that. `-XScopedTypeVariables` is of course unchanged, it simply does something more wr.t. pattern signatures (adds implicit binds) than the underlying `-XScopedTypeVariables`.

Ericson2314:
Stepping back, this proposal has two separate "constituencies", if you will:

1. People that may or may not care about single namespaces and Dependent Haskell, but think implicit binding is bad. They can also be served by a warning, as you point out. (Still, no one but @AntC2 has come out actively in support of implicit binding. And @AntC2 already doesn't like `-XScopedTypeVariables` too for other reasons IIUC. If no almost no one like implicit binding, why should our *brand new" `-XPatternSignatures` include it? Old code might use implicit binds, but old code isn't directly affected by new extensions!)

2. People that strongly support the single namespace but may or not care about implicit binding. The problem with our current implicit binding is that it is hostile to cross namespace usages as described. This issue can *not* be solved by warnings — the problem is what you can *not* do, not things you can do but don't want to do, which can indeed be solved with linting. In other words, the problem isn't actually the implicit binding itself, the problem is that it steals syntax from the mere usage.

These are rather separate points, our conversation has been about 2 and the proposal as written has been about 1. I'm just trying to kill two birds with one stone, and prevent needing yet more extensions / axes of variation — because I have yet to meet anyone that doesn't want `-XScopedTypeVariables` but does want some sort of implicit binding in pattern signatures.

AntC2:
You called? ;-)

> @Ericson2314 If no almost no one like implicit binding, why should our *brand new" -XPatternSignatures include it?

I think I'm the only person with experience using `-XPatternSignatures` -- and that's in a form with implicit bindings. I don't see how you can say "almost no one like implicit binding": how could they experience it? (And yes we haven't met in person, but you're counting me as not even an "anyone"? ;-)

To be precise: in the form I use, all tyvars appearing within the function's equation are scoped _only_ within that single equation. That's consistent with the var scoping in [this example from the wild](https://gitlab.haskell.org/ghc/ghc/-/issues/12203#note_121599) (which I find confusing enough already)

```
pattern ConsK :: a -> D k (StreamK k a) -> StreamK k a
pattern ConsK x xs <- StreamK (Cons x ((UnsafeD . StreamK) -> xs)) where   -- this xs different to
    ConsK x (UnsafeD (StreamK xs)) = StreamK (Cons x xs)                   -- that xs (different types)
```

I'm not the only person who's found `-XScopedTypeVariables` awkward/dysergonomic. And now the tyranny of backward compatibility means GHC has to keep supporting it, even if we arrive at a better design. If the proposal here in effect forces using `-XScopedTypeVariables`, it'll be more of a straitjacket. Combining with the single namespace extensions/promoted variables will make it just more confusing:

```
x = 1
...      -- pages of code
f :: forall x. x -> ...
f (Just y) :: x = ...
f  x       :: y = ...
...      -- more pages of code
data Cart = Cart{ x, y :: Double }
```

(Yes of course my immediate reaction is: don't do that. So this is more what if I forgot I'd named a module-wide variable `x`, or I'd picked up code written by someone else, or I'd imported a module with an `x`.)

Which `x` is that in `:: x`? Might it even be the field selector function from `Cart` promoted to a type? The advantage I see _without_ `-XScopedTypeVariables` and without datavar promotion is that must be a binding/I don't need to nervously scan up and down the program looking for name clashes.

I don't have experience with `-XPatternSignatures + -XScopedTypeVariables + promoted variables` -- but then neither does anyone else. (Nor do we yet have `-Wpuns, -Wpun-bindings` -- this is getting all rather hard to imagine.) My guess is that `-XPatternSignatues` with implicit binding is more ergonomic than with `-XScopedTypeVariables`, so it provides a longer-term route to deprecate `-XScopedTypeVariables`.

------------

> @goldfirere Note that the situation with `f @a = ...` is different from what we're debating here, because `f @a` should unambiguously bind `a` (shadowing any other `a` in scope)

That makes good sense. So I can go like this?:

```
f :: forall c d e. d -> (c, Maybe e)
f @a x :: (a, b) = ...
```

in which the `a` in the Pattern Signature is explicitly bound from the `@a`, but the `b` is implicitly bound? (`b` to be resolved to `Maybe e` from the signature.)

goldfirere:
@Ericson2314 

> Term bound `x` can now be used in a pattern signature, with -XPatternSignatures and without -XScopedTypeVariables.

I don't see how you mean this. We can't currently use a term-bound `x` in a pattern signature, regardless of this proposal. Today, the namespaces remain separate. Or have I missed a beat somewhere?

Separately, you argue that implicit binding is poor for most. Maybe that's true, in which case I would agree that we shouldn't go to pains to support a feature that few want. But you'd have to convince me that few like implicit binding, which is a different conversation.

@AntC2 

> So I can go like this?

With implicit binding in pattern signatures, yes, exactly as you say.

Ericson2314:
@goldfirere 

> I don't see how you mean this. We can't currently use a term-bound `x` in a pattern signature, regardless of this proposal. Today, the namespaces remain separate. Or have I missed a beat somewhere?

@int-index's plan is that we should *always* fallback on checking in the other namespace if something isn't found in the "right" namespace; in other words, his plan to do what we do for `-XDataKinds` but in both directions. To support this are many other proposals, but no `-XUniifiedNamespaces`.

> Separately, you argue that implicit binding is poor for most. Maybe that's true, in which case I would agree that we shouldn't go to pains to support a feature that few want. But you'd have to convince me that few like implicit binding, which is a different conversation.

Well, my evidence was seeing people who are usually less excited about single namespaces and dependent things (such as SPJ, via the comment I quoted in the amended proposal body). But sure, this is not representative, so I just created this poll https://discourse.haskell.org/t/need-feedback-on-pattern-signature-behavior/7704

AntC2:
Re the points being made on the Discourse poll about reading code you didn't write (or not recently) and the rules of thumb for trying to keep track of where the tyvars are coming from ...

Wild thought: could there be a warning for where a function type signature with explicit `forall` doesn't appear immediately before its definition? (There might be blank lines or comments intervening, but no other definitions.) Notionally they're syntactically bound together like

```
f :: forall a. a -> a
in                                        -- borrowing a keyword
f (x :: a) = ...
```

simonpj:
Let me attempt to summarise, and John can tell me if I have this right.

* In #448, `-XPatternSignatures` enables
  1. The possibility of a type signature in a pattern, e.g. `f (x::Int) = <body>`
  2. Pattern signatures can bring type variables into scope, e.g. `f (x::a) = <body>`, where the type variable `a` is in scope in `<body>`

* John proposes instead that `-XPatternSigatures` enables only (i).  Instead, `-XScopedTypeVariables` enables
  * `-XPatternSignatures`
  * `-XMethodTypeVariables`
  * `-XExtendedForAllScope`
  * *and* (ii) that pattern signatures bring type variables into scope

* That fourth bullet seems a bit non-uniform. We could instead:
  * Add language extension `-XPatternSignatureBinds` which makes pattern signatures bring type variables into scope
  * `-XScopedTypeVariables` is then just a synonym for `-XPatternSignatures` + `-XMethodTypeVariables` + `-XExtendedForAllScope` + `-XPatternSignatureBinds`.

* Rather separately, Richard is arguing for the possiblity of using *warnings* rather than *language extensions*.  This really a separate conversation and I have opened #615 to discuss it.

Does that accurately summarise where we are at? I don't think there any disagreements of principle here: in particular, we all agree that it'd be good to allow programmers to control whether or not `PatternSignatures` brings variables into scope.  We have differences of taste about exactly how to do that, which I have attempted to summarise above.

Ericson2314:
@simonpj Thanks, I think that is good summary. There is just one thing missing, which is the interaction between the other single namespace proposals. Concretely, if we go with a warning `-Wpattern-signature-binds` instead, then #270 (which should be unblocked now that #581 is merged) has to have some sort of extension after all and can't be just warnings.

simonpj:
> Concretely, if we go with a warning -Wpattern-signature-binds 

That is point (4) in my list, right?  If we adopted (3) instead would that caveat go away? If not can you explain?

int-index:
I have not devoted as much time as @Ericson2314 has to thinking about the interactions of T2T, `-Wpuns`, and pattern signature binds.

What are concrete examples of code that make it impossible to reject implicit binds with a warning rather than an extension?

Ericson2314:
This program:
```haskell
t = List Unit
f (x :: t) = x <> x
```
With pattern signature binds, we have this inferred type:
```haskell
f :: Semigroup t => t -> t
```
Without pattern signature binds `t` and with T2T, `t` will be a *use* not a *bind*, thus we'll get
```haskell
f :: List Unit -> List Unit
```

simonpj:
> Without pattern signature binds t and with T2T, t will be a use not a bind, thus we'll get

I don't think so.  See https://github.com/ghc-proposals/ghc-proposals/pull/608.  I think that means that
```
t = List Unit
f :: t -> t
```
is simply rejected: a term variable occurring in a type.  Right Vlad?

I still don't (yet) see how this intersects [my summary above](https://github.com/ghc-proposals/ghc-proposals/pull/608#issuecomment-1744614105)

Ericson2314:
Oh hmm https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0281-visible-forall.rst#id15 does actually say that there is an extension opt-in. So we could get away with a warning after all. #281 never mentioned pattern signature binds, so that is a bit ambiguous, but I think it is a pretty straightforward reading that the same stuff should apply.

Ericson2314:
@simonpj Ah two things:

1. I didn't realize you were outright suggesting adding back `-XPatternSignatureBinds`, as opposed to saying that we're blocked on that other issue you made for Richard's larger point. I am perfectly fine adding back `-XPatternSignatureBinds`; the only reason I didn't is because I have also pondered the idea few of the implicit binding extensions should be combined because they have different defaults (which draft #532) does. But we can add it back for now.

   (I don't think `-XPatternSignatureBinds` will make Richard happier with this PR, but I don't think you necessarily do either.)
   
2. OK I am checking out #207. One question I have is when `type ..` is required in patterns. E.g. is
   ```haskell
   sizeOfWrapper :: forall a -> Int
   sizeOfWrapper t = sizeOf t
   ```
   OK, or do we need to do
   ```haskell
   sizeOfWrapper :: forall a -> Int
   sizeOfWrapper (type t) = sizeOf t
   ```
   ?

simonpj:
> One question I have is when type .. is required in patterns

@int-index @goldfirere what do you say?  #281 seems silent on the question of a pattern-to-type mapping for patterns.  It speaks only of term-to-type mapping for terms.  Should this not be explicit in the proposal?

Ericson2314:
For the record, my interpretation of #281 and #607 is that skipping the `(type ...)` in pattern position is allowed. #281 part II does have

> The syntactic descriptions here applying to expressions apply equally to patterns, though we will continue to discuss only expressions.

and 

> With the proposed changes, the namespace of an identifier is no longer tied to whether it stands for a type variable or a term variable.

> Variables and constructors (regardless of their namespace) are mapped directly, without modification.

So my interpretation is that with `\x -> ` we bind the identifier `x` in the term namespace regardless of whether it (semantically) denotes type or term variable. And then we look it up in type position (semantically) it must be a type variable or else the T2T lookup fails.

But yes, it would be good to make this explicit in #281 so we don't need to come back to it.

Ericson2314:
OK @simonpj I've reworked this:

1. As you requested, `-XPatternSignatureBinds` is back, `-XScopedTypeVariables` is once again just implying other extensions, and doing nothing on its own.
2. I've made the motivation relating to single vs multiple namespaces. My referring to unmerged #270 has been replaced with accepted and almost implemented #281, so I feel much better about this.

simonpj:
Thanks.  I found the new 3.2.2 hard to follow.   But I am content with the actual payload in 3.5.

@rae

Ericson2314:
What's the status at this, at first @simonpj was saying let's not block on https://github.com/ghc-proposals/ghc-proposals/issues/615, but now that has become https://github.com/ghc-proposals/ghc-proposals/pull/620.

proposals/0448-type-variable-scoping.rst
```diff
@@ -179,6 +183,139 @@ it may actually break my program somewhere!
...
+The error message is completely different, but the effect with respect to merely whether the program was rejected is the same.
+(Choosing a way to disable the implicit binding part of ``-XPolyKinds`` is left as future work --- possibly a future amendment to this proposal.)
+
+Now, both extensions (``-XPatternSignatures`` and ``-XKindSignatures``) just allow, respectively, term-level and type-level signatures, with no other functionality like implicit binding mechanisms also thrown in.
+
+Conservativism for standardization
+""""""""""""""""""""""""""""""""""
+
+With both of these extensions being very minimal, I think they would be easy uncontroversial candidates for a new language report.
+Conversely, all implicit binding constructs are very fraught with a complicated mix of upsides and downsides, we and should only standardize them with great care.
```
goldfirere, line 317:
```suggestion
Conversely, all implicit binding constructs are very fraught with a complicated mix of upsides and downsides, and we should only standardize them with great care.
```

goldfirere (commented):
Just some typos, etc. Nothing interesting in the review.

goldfirere:
OK. I'll recommend acceptance. Though I've come to think that we're fighting the wrong battle here, in extension shuffling. #628 paves a different way forward. Yet those ideas should not stop others from making progress.

Ericson2314:
Great!

And to be clear, I would be very happy to someday reformulate this stuff in https://github.com/ghc-proposals/ghc-proposals/pull/628 approach. The gist is that I am hoping new editions going forward can have as little implicit binding as possible, and no pattern signature binds in particular (as `@` binding obviates it). Editions should indeed remove the need to have a `-XPatternSignatures` (without implicit binding) at all --- it should be a mandatory part of all new editions whether they fall on the "student/conservative -- fancy types" spectrum.

Ericson2314:
I see there was a lot of discussion of this on the mailing list https://mail.haskell.org/pipermail/ghc-steering-committee/2024-January/thread.html

I just want to answer a few things:

## Deprecating `ScopedTypeVariables`?

In my view, #448 (the proposal I am amending) was *always* about opening the door to someday deprecating `ScopedTypeVariables`. The proliferation of extensions was essentially generated by the Venn diagram of (a) what we have today, and (b) what we conceivably want tomorrow.

Yes, actually removing support for `ScopedTypeVariables` is inconceivable as it is one of the widely extensions. But a large part of the motivation editions process is to untangle what is (a) supported/allowed vs (b) promoted/endorsed. The principles explain how `ScopedTypeVariables` fall short. The more dependent stuff we get, the worse the problem gets.

My amendment (this proposal) basically seeks to claim that the same sort of problems that apply the whole of `ScopedTypeVariables` also apply to the *new* (as of the orignial #448) `PatternSignatures`. It feels odd to me to make a new extension only for it too include more than what may want in the future, and thus we would have to go through the whole #448-style extensions splitting process again.

## Warnings vs extension diagnostics

The philosophical parts have already been discussed elsewhere; I don't mean to relitigate that. I do want to point out a practical difference. `-Wpattern-signature-binds` presumably makes a message like this:

> error: pattern signature binds ‘foo’

With the proposal, `-XPatternSignatures` without `-XScopedTypeVariables` presumably makes a warning like this:

> error: Not in scope: type variable ‘foo’

If the goal is, as I believe, to treat this sort of implicit binding as a mistake, I think the second sort of message is the right one to end up. Editions (assuming that is the end goal) should not explain things in terms of other editions. Editions should be self-contained including their mental modal, and the prior editions or sibling editions should be mere "counterfactuals" that one need not understand.

In short, I advocate [*Damnatio memoriae*](https://en.wikipedia.org/wiki/Damnatio_memoriae) for the features we don't want when they are disabled.

Whether we should do that right away with extensions, or wait for editions, is a separate question that applies equally well to this proposal or the original #448.

## Is `PatternSignatures` deprecated enough to change its meaning?

@simonmar brought this up. I don't claim to know the answer --- maybe it is in fact no! But that's a criticism that applies to the original proposal, which reintroduced `PatternSignatures` with changed meaning. This proposal does change its meaning from the current plan, but it doesn't change the fact the plan is for it to change meaning somehow.

I have no qualms staying safe and deciding we should use `PatternSignatures2` or something instead. No qualms at all! But as that is an orthogonal solution to an orthogonal problem (present in the original #448), I don't think it should hold this up.

Ericson2314:
Thanks @adamgundry for relaying this to the list :)

@simonpj writes

> I see. So you are proposing
>
> - PatternSignatures = SimplePatternSignatures + PatternSignatureBinds

I think that proposal was to keep `PatternSignatures = ScopedTypeVariables`, as it not proposed in the preexisting #448 but it as currently implemented. This is because that deprecation process was never "finished", so it would be a still-not-quite acceptable change to repurpose `PatternSignatures` for something else.

If we can change `PatternSignatures` at all, I would advocate for going all the way to `PatternSignatures` being the only knob and no `SimplePatternSignatures` or `PatternSignatureBinds`, simply for sake of having fewer knobs and because I don't think there is much demand for *just* pattern signature binds but not other forms of implicit binding.

Ericson2314:
Hey just checking it looked like a positive consensus was being reached on the mailing list end of Jan / beginning of Feb, but was there ever a final vote?

goldfirere:
It does look like there's a build failure. @Ericson2314 do you want to investigate this before we merge?

nomeata:
The readthedocs failure can be ignored, or probably be solved by merging `master` into this branch.

VitWW:
> It does look like there's a build failure.

The reason of readthedocs failure - this git-fork is not updated for too long, last update was before Feb 2024.

Ericson2314:
I am not back from a trip, and just pushed a merge with master. I'll watch this to make sure CI passes.

Issue #598, opened 2023-06-20
Or patterns syntax: `(p1|p2)` or `(p1;p2)`?


sgraf812:
In the previous [vote on Or pattern syntax](https://github.com/ghc-proposals/ghc-proposals/issues/587), infix operator syntax won by a large margin.

In this vote, we want to settle if we rather want `(p1|p2)`, `(p1; p2)` or something entirely different, such as `(p1 \/ p2)` (or `;;`).
Note that all operators will need mandatory parentheses around them to avoid ambiguities.

**As in #587, you can vote by reacting to this post via emoji in one (or two, if you want to express dissent for one of the options) of three ways**:

Vote :tada: for `(p1|p2)`. Pros:

- Familiarity/trodden path: This is the operator that OCaml uses for Or patterns.
- Compared to `(p1;p2)`: No risk of confusion with record field matches `(T1{x=_;y=_}; T2{})`

Vote :heart: for `(p1; p2)`. Pros:

- Asymmetry expresses left bias of first match semantics well
- Compared to `(p1|p2)`: Does not steal syntax from a potential "guards-in-patterns" extension as sketched out [here](https://gitlab.haskell.org/ghc/ghc/-/wikis/view-patterns-alternative). That proposal might be a viable alternative to view pattern syntax, which [is incompatible with type-kinded patterns `f (a -> b)`](https://github.com/ghc-proposals/ghc-proposals/issues/598#issuecomment-1600648262) in a future Dependent Haskell, according to @int-index.
- Could harmonise well with layout in case expressions (see below), if we want that, in which case the outer parens could be omitted. Whether or not we want that would be another vote.

Vote :rocket: for some other operator, such as `(p1 \/ p2)`/`(p1 ' p2)`/you name it. (If this wins, we'll have another vote.)

---

### Infix `;` and layout in `case`

Proposed by @rae in https://github.com/ghc-proposals/ghc-proposals/pull/585#issuecomment-1550485615.
We could use layout to insert `;` in `case` expressions rather nicely:

```hs
case ... of
  A -> dealWithA
  B -> dealWithB
  C
  D
  E -> dealWithTheRest
```

which is quite like

```hs
case ... of
  A -> dealWithA
  B -> dealWithB
  C; D; E -> dealWithTheRest
```

Cute. Note that neither program is currently valid Haskell (not even with `-XNondecreasingIndentation`, however that works), so there is no risk of introducing a breaking change.

---

[...]

Issue #587, opened 2023-05-17
Or patterns: Prefix signalling or just infix?


sgraf812:
Many users are quite unhappy with the syntax proposed in the currently accepted proposal #522. #585 tried to improve matters while sticking to the accepted design.

Still, users are unhappy with the prefix operator `one of` of proposal #522. During the discussion in #522, this post https://github.com/ghc-proposals/ghc-proposals/pull/522#issuecomment-1295506494 was the moment we adopted the prefix operator `one of (T1;T2)` rather than the infix operator `(T1; T2)`.

:tada: Pros for the prefix operator:

1. Easy to google: `one of Haskell` will hopefully lead to the user guide page on Or patterns. An infix operator like `(T1; T2)` would be really hard to google and might lead to confusion in e.g., code reviews (where people don't scroll up to the extension section or the extension is hidden in a cabal file). 
2. Clear signalling: `one of` means "here begins an or pattern". With infix operators, you are kind of thrown into the middle of it and have to parse the extent of the pattern.
3. Nullary Or patterns via `one of {}`, although the advantage is rather weak; perhaps it's useful for code generators

:heart: Pros for the infix operator (might be `;`, but in theory other operators such as `\/` and `|` are on the table, too):

1. Cute, less noise than `one of`.
2. Precedent: OCaml (for `|`). No other language uses a prefix operator such as `one of`.
3. If the infix operator is `;`: Harmonises well with layout in case expressions (see below). `one of` would harmonise well, too (#585), although not quite as well.

### Infix `;` and layout in `case`

Proposed by @rae in https://github.com/ghc-proposals/ghc-proposals/pull/585#issuecomment-1550485615.
We could use layout to insert `;` in `case` expressions rather nicely:

```hs
case ... of
  A -> dealWithA
  B -> dealWithB
  C
  D
  E -> dealWithTheRest
```

which is quite like

```hs
case ... of
  A -> dealWithA
  B -> dealWithB
  C; D; E -> dealWithTheRest
```

Cute. Note that neither program is currently valid Haskell (not even with `-XNondecreasingIndentation`, however that works), so there is no risk of introducing a breaking change.

Richard goes on to propose `one of { T1; T2 }` in patterns as an allowed alternative to `T1; T2` to allow for layout, e.g., 

```hs
f (one of T1
          T2) = 42
```

Not sure if that pulls its weight, though.

Note also that this layout rule might lead to confusion in conjunction with `-XLambdaCase`:

```hs
[...]

PR #609, opened 2023-08-04
Amend or patterns (#522) to use p1 ; p2

knothed:
Proposal #522 defines _or-patterns_, which match multiple patterns simultaneously. For example,

```
f (one of 3, 4, 5) = …
```
would match on all `3`, `4` and `5`.

After trying to amend #522 to adjust the syntax (in #585), some unhappiness was raised about the current `one of` syntax. We then made two votes to let the community decide the exact syntax that _or-patterns_ should follow.

[Vote 1](https://github.com/ghc-proposals/ghc-proposals/issues/587) decided that _or-patterns_ should have infix syntax and no prefix signalling like `one of` (with 55-15 votes).

In [Vote 2](https://github.com/ghc-proposals/ghc-proposals/issues/598), the options `(p1; p2)` and `(p1 | p2)` were compared. The result was roughly balanced, with a little bit more support towards `;` (48-43 votes).

So this amended proposal proposes the `(p1; p2)` syntax with `(p1 | p2)` being suggested as an alternative.

### Layout Rule

A good reason to vote for `(p1; p2)` is the use of the layout rule introduced by `of` (sources: [@bradrn](https://github.com/ghc-proposals/ghc-proposals/issues/598#issuecomment-1660466684), [@lierdakil](https://github.com/ghc-proposals/ghc-proposals/issues/598#issuecomment-1599210728)).
This makes it possible to write

```
f x = case x of
  A
  B
  C -> 3
  D
  E -> 4
  F -> 5
```

Or-Patterns could then appear *without* parentheses in some places and could make use of automatic `;` insertion when they are inside a layout block introduced by `of`.

Edit: this is now also a part of this proposal.

### DIFF
--- a/proposals/0522-or-patterns.rst
+++ b/proposals/0522-or-patterns.rst
@@ -128,8 +128,8 @@ We give the static semantics in terms of *pattern types*. A pattern type has the
 Then the typing rule for Or patterns is:
 ::
 
-          Γ0, Σ0 ⊢ pat_i : τ ⤳ Γ0,Σi,Ψi
-    -----------------------------------------
+      Γ0, Σ0 ⊢ pat_i : τ ⤳ Γ0,Σi,Ψi
+    ---------------------------------
     Γ0, Σ0 ⊢ ( pat_i ) : τ ⤳ Γ0,Σ0,∅
 
 
--- a/proposals/0522-or-patterns.rst
+++ b/proposals/0522-or-patterns.rst
@@ -74,42 +74,74 @@ Proposed Change Specification
 Changes in the grammar
 ~~~~~~~~~~~~~~~~~~~~~~
 
-We consider this as an extension to `Haskell 2010 grammar
-<https://www.haskell.org/onlinereport/haskell2010/haskellch10.html#x17-18000010.5>`_.
-This proposal adds one more production to the nonterminal ``apat``: ::
+We consider this as an extension to GHC's [Parser.y](https://gitlab.haskell.org/ghc/ghc/-/blob/master/compiler/GHC/Parser.y).
+This proposal adds one more production to the nonterminal ``aexp``: ::
 
-    apat -> ( orpats )
+    aexp -> ( orpats )
+
+It also adds one more production to the nonterminal ``pat``: ::
+
+    pat -> pat ';' orpats
 
 where ``orpats`` is a new nonterminal with the following productions: ::
 
-    orpats -> pat
-            | pat ';' orpats
+    orpats -> exp
+            | exp ';' orpats
+
+In addition, the nonterminal ``pattern_synonym_decl`` currently has three productions: ::
+
+    pattern_synonym_decl -> 'pattern' pattern_synonym_lhs '=' pat
+    pattern_synonym_decl -> 'pattern' pattern_synonym_lhs '<-' pat
+    pattern_synonym_decl -> 'pattern' pattern_synonym_lhs '<-' pat where_decls
+
+We propose to change the ``pat`` nonterminal in all of these productions to the ``infixexp`` nonterminal.
 
 
-N.B.: In the concrete GHC Parser, we use the nonterminals `aexp` and `exp` instead of `apat` and `pat` because of the ECP parsing.
+N.B.: The mixture of `exp` and `pat` nonterminals is due to the ECP parsing.
 
 
-Some examples that this new grammar produces: ::
+**Some examples that this new grammar produces:**
+
+Or-patterns with parentheses: ::
 
   case e of (T1; T2{}; T3 a b) -> ...
 
   f :: (Int, Int) -> Int
   f (5, (6;7)) = 2
 
-  -- in expression context
-  let ([x]; (x : y : z)) = e1 in e2
+Unparenthesized or-patterns:
 
-  -- pattern guards in declarations
-  f x y
-    | x@(T1 _; T2 a b) <- e1
-    , guard x
-    = e2
+  case e of
+    1; 2; 3 -> x
+    4; (5; 6) -> y
 
-  -- nested Or patterns
-  case e1 of
-    ((T1; T2); T3; T4) -> e2
+Unparenthesized or-patterns using layout:
 
-NB: The new grammar allows Or patterns which bind variables. These will however be rejected in `2.2`_.
+  case e of
+    1
+    2
+    3 -> a
+    4
+    5;6 -> b
+    7;8 -> c
+
+  f x = case x of
+    A _ _; B _
+    C -> 3
+    (D; E (Just _) Nothing)
+     -> 4
+    F -> 5
+
+N.B.: Unparenthesized or-patterns only work in some places where patterns are expected. For example, in ::
+
+  g x = do
+    A; B <- x
+    return 1
+
+the ``A; B <- x`` is interpreted as two statements. Parentheses would have to be used around ``A; B`` to make it denote an or-pattern.
+
+
+N.B.: The new grammar allows or-patterns which bind variables. These will however be rejected in `2.2`_.
 
 .. _2.2:
 
@@ -291,12 +323,24 @@ Considering view patterns, these do work seamlessly with Or patterns. As specifi
  f a (((== a) -> True); (show -> "yes")) = True
  f _ _ = False
 
+.. 4.2:
 Costs and Drawbacks
 -------------------
 The cost is a small implementation overhead. Also, as Or patterns are syntactic sugar, they add to the amount of syntax Haskell beginners have to learn.
 We believe however that the mentioned advantages more than compensate for these disadvantages.
 Or patterns are available in all of the top seven programming languages on the TIOBE index (Python, Java, Javascript, C#, C, etc.), which suspects that the concept won't be particularly troublesome for beginners to learn.
 
+Additionally, the changes to the ``pattern_synonym_decl`` nonterminal are required to allow unparenthesized or-patterns in the syntax. This breaks pattern synonym declarations using a type annotation such as: ::
+
+  pattern X <- 42 :: Int
+
+One would now have to parenthesize the right side of the pattern synonym and write ::
+
+  pattern X <- (42 :: Int)
+
+A search on Hackage shows that this syntax is currently only used in very few places.
+
+The syntax change also makes 3 tests in the `patsyn` testsuite fail, out of 250 total `patsyn` tests. The fix is always to enclose the right side in parentheses.
 
 Alternatives
 ------------
@@ -304,28 +348,44 @@ Alternatives
 There have been proposed a **lot** of alternatives in regard to the exact syntax of or-patterns (see the discussion [here](https://github.com/ghc-proposals/ghc-proposals/pull/585)).
 
 After performing two community votes ([Vote 1](https://github.com/ghc-proposals/ghc-proposals/issues/587), [Vote 2](https://github.com/ghc-proposals/ghc-proposals/issues/598)), the relative majority voted for the here-proposed `(p1; p2)` syntax, with `(p1 | p2)` being close behind (with 48-43 votes).
-So, a suitable alternative would be to use the syntax `(p1 | p2)`. On the positive side, the `|` seperator naturally resembles an *or*-operation, while on the negative side this syntax could be better used by a future "guards in patterns" proposal.
+So, a suitable alternative would be to use the syntax `(p1 | p2)`.
 
-Layout Rule
-~~~~~~~~~~~
+While we like both options `(p1 ; p2)` and `(p1 | p2)`, as both convey the alternative nature of the *or*-pattern pretty well, we tend towards the semicolon.
 
-It would be nice if or-patterns could make use of the layout rule introduced by `of` to allow writing alternatives of the or-patterns simply below each other, the `;` being inserted automatically:
+While `|` is a pretty natural choice regarding an *or* operation, the semicolon does a better job in showing the asymmetry of the pattern as later alternatives are only evaluated when earlier ones fail to match.
 
-```
-f x = case x of
-  A
-  B
-  C -> 3
-  D
-  E -> 4
-  F -> 5
-```
+Also, the `(p1 | p2)` syntax could be better used by a future "guards in patterns" proposal.
 
-Note that, in this case, or-patterns would not require parentheses.
-For this to work however, a minor change to the `pattern_synonym_decl` production is required (as outlined [`here`](https://github.com/ghc-proposals/ghc-proposals/issues/598#issuecomment-1602600326)).
+Another great advantage of `;` over `|` is the use of the layout rule: in a layout context introduced by `of`, semicolons are automatically inserted into equally-indented lines. This makes it possible to write ::
 
-This change (or-patterns without mandatory parentheses and required changes to `pattern_synonym_decl`) will be proposed in a follow-up proposal - it is not part of this one.
+  f x = case x of
+    1
+    2
+    3 -> x
 
+where the or-pattern is implicitly parsed as `1; 2; 3`.
+This resembles the `switch/case`-syntax known from languages like C and Java.
+
+No unparenthesized or-patterns
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+As described in `4.2`_, the changes to the `pattern_synonym_decl` nonterminal are breaking changes in some rare cases.
+We could avoid introducing these breaking changes by requiring or-patterns to be parenthesised. This means, the proposal would not include the following production ::
+
+      pat -> pat ';' orpats
+
+and would not include the changes to the `pattern_synonym_decl` nonterminal.
+
+Beware that it would still be possible to use the layout rule even with parenthesized or-patterns as follows: ::
+
+    case a of
+      (A
+      B
+      C) -> 1
+
+This is an artifact of the layout rule and is not intended to be used. It could even be made into a syntax error.
+
+When disallowing the unparenthesized syntax `p1; p2`, we do not see much advantage of the `;` separator over the `|` separator, except that the unparenthesized syntax could be added some time in the future.
 
 Binding pattern variables
 ~~~~~~~~~~~~~~~~~~
--- a/proposals/0522-or-patterns.rst
+++ b/proposals/0522-or-patterns.rst
@@ -6,7 +6,7 @@ Or Patterns
 .. ticket-url::
 .. implemented::
 .. highlight:: haskell
-.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/609>`_.
+.. header:: The old proposal was `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/522>`_. This amendment proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/609>`_.
 .. sectnum::
 .. contents::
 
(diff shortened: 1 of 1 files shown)


proposals/0522-or-patterns.rst
```diff
@@ -62,7 +62,7 @@ function we get
 
     stringOfT :: T -> Maybe String
     stringOfT (T1 s)        = Just s
-    stringOfT (one of T2{}, T3{}) = Nothing
+    stringOfT (T2{} ; T3{}) = Nothing
```
konsumlamm, line 65:
```suggestion
    stringOfT (T2{}; T3{}) = Nothing
```

proposals/0522-or-patterns.rst
```diff
@@ -150,25 +141,22 @@ Informal semantics in the style of `Haskell 2010 chapter 3.17.2: Informal
 Semantics of Pattern Matching
 <https://www.haskell.org/onlinereport/haskell2010/haskellch3.html#x8-600003.17.2>`_:
 
-- Matching the pattern ``(one of p1, …, pk)`` against the value ``v`` is the result of matching ``v`` against ``p1`` if it is not a failure, or the result of
-  matching ``(one of p2, …, pk)`` against ``v`` otherwise.
-
-  ``p1``, …, ``pk`` bind no variables.
+- Matching the pattern ``(p1; ... ; pk)`` against the value ``v`` is the result of matching ``v`` against ``p1`` if it is not a failure, or the result of
+  matching ``(p2; ... ; pk)`` against ``v`` otherwise. We require that ``p1``, …, ``pk`` bind no variables.
```
konsumlamm, line 145:
```suggestion
- Matching the pattern ``(p1; ...; pk)`` against the value ``v`` is the result of matching ``v`` against ``p1`` if it is not a failure, or the result of
  matching ``(p2; ...; pk)`` against ``v`` otherwise. We require that ``p1``, …, ``pk`` bind no variables.
```

proposals/0522-or-patterns.rst
knothed: Both of them are valid syntax.

nomeata:
> This change will be proposed in a follow-up proposal - it is not part of this one.

Why so piecemeal? Is there a good reason you can't propose the whole thing, either with the layout rule (which looks good to me based on your example) or with a rationale not to have it?

knothed:
> Is there a good reason you can't propose the whole thing

The only reason to split this up into two proposals is to avoid having one „too large“ proposal which has more potential for discord, so we thought two smaller proposals could go through the process easier.

But you’re right, it does seem very sensible to combine both proposals.

proposals/0522-or-patterns.rst
konsumlamm: I know, but the rest of the text has no space before `;`.

nomeata:
From the committee side, ideally all the discord has been resolved by way of cunning reasoning and uncontestable arguments (or, if not possible, lots of bike shedding and popular votes) before it reaches us :-). Of course the proposal can list alternatives with pros and cons, but we'd like to hear from the proposal authors what they genuinely think is best (and why)!

simonpj:
FWIW I'm leaning (pretty strongly) towards not wanting to have layout involved in or-patterns.  For me it's too subtle, and the use-cases aren't compelling.

I suppose that one advantage of adopting infix ";" is that if, in the light of some years of experience, layout after all seems desirable, it'd be easier to add it.

knothed:
> FWIW I'm leaning (pretty strongly) towards not wanting to have layout involved in or-patterns. 

Well, sounds like another vote :D

nomeata:
I don't expect layout in _nested_ patterns to be particularly pretty, but when the disjunction is outermost, then layout as in
```
f x = case x of
  A
  B
  C -> 3
  D
  E -> 4
  F -> 5
```
is something I'd welcome, at least if there are no particular bad consequences (besides “looks unusual when not used to it yet”).

AntC2:
> @simonpj For me it's too subtle, and the use-cases aren't compelling.

Yeah, @knothed and @nomeata have shown the 'happy path'. Case discriminants aren't always so neat as a single constructor. (How) does this work?

```
f x = case x of
  A _ _; B _
  C -> 3
  {D; E (Just _) Nothing }
  -> 4
  F -> 5
```

googleson78:
@AntC2 

Which part of that match are you concerned about in particular?

(To me) It seems to directly translate as:

1. If the the constructor of `x` is `A` or `B` or `C`, return `3`
2. If the constructor of `x` is `D` or it's `E`, for which the first argument of `E`'s constructor is `Just` and the second one's constructor is `Nothing`, return `4`
3. If the constructor of `x` is `F`, return `5`.

To me the two parts with issues in that program are:
1. The `->` for the second match needs to be indented from the match, currently, not sure if you meant to suggest that it would not need to be after the layout proposal.
2. More importantly, the usual usage of `{}` to denote "this is usually layouted, but you can use {} and ; to write it on the same line" is a bit weird here, since we would already allow writing single line or-patterns without `{}`. On the other hand, not allowing `{}` to surround or patterns seems a bit inconsistent. I would personally prefer the former where you can write `X ; Y` or `{X ; Y}` and both work.

sgraf812:
> FWIW I'm leaning (pretty strongly) towards not wanting to have layout involved in or-patterns. For me it's too subtle, and the use-cases aren't compelling.

I don't think the choice is "should `(p1; p2)` participate in layout?", it rather is "should we be able to leave off parens in `(p1; p2)`?".
Layout is activated by `of`, not by the use of an Or pattern. So even in the proposal as written you can (dubiously) exploit layout to write

```hs
f x = case x of
  (A
  B
  C) -> 3
  (D
  E -> 4
  F) -> 5
```

(NB: Since layout is sensitive to indentation, the "nicer" syntax where the `B` and `C` lines are indented by an additional space won't be accepted.)

As long as we use `;`, the same ASCII character as the "line separator" in other syntactic sorts, I see absolutely no way to deactivate layout *just* for Or patterns that respects the phase separation between the lexer and the parser.

> I suppose that one advantage of adopting infix ";" is that if, in the light of some years of experience, layout after all seems desirable, it'd be easier to add it.

The advantage is rather that exploiting layout as above becomes (subjectively) *readable* only if you leave off parens. As the Alternatives section states, that would need an insignificant (as far as hackage is concerned) breaking change to PatSyn declarations.
But again, leaving off parens where it is unambiguous might in itself be a win (or it might not; subjective), and that could be determined after "some years of experience" with the baseline, mandatory-parens version. 
(The proposal takes some care in requiring parens around ViewPats that this upgrade path is actually possible.)

Whether or not people want to exploit layout is an optional secondary concern. I don't mind exotic formatting; it would be banned in style guides/auto formatters anyway. On the bright side, code like @AntC2 posted above (which is not quite syntactically correct due to `{}` and misplaced line breaks) gives IOCCC hackers something to work with.

AntC2:
> 2\. More importantly, the usual usage of `{}` to denote "this is usually layouted, but you can use {} and ; to write it on the same line" is a bit weird here,

Hmm. I'm not convinced by your claim "usually layouted". I tend to use layout only for quick-and-dirty hacking of ad-hoc code. For longer-lived code with case branches of nested constructors, I prefer to format everything 'properly'. In particular the rhs of the `->` might be quite lengthy. So I like a sentinel `;` to say that branch is finished, on to the next.

Using a match to a bare constructor is what's weird here IMO. So showing only the 'happy path' is not demonstrating how messy it might look in practice.

And it would be inconsistent with the rest of Haskell to allow `;` separators but not `{  }` grouping. Ok so thanks for the answer we still need to indent the `-> ...`.

> I would personally prefer the former where you can write `X ; Y` or `{X ; Y}` and both work.

Emm? I'm expecting both would work; which is why I demonstrated both together. Is mix-and-match not the case? I was asking if anybody thought it looked unacceptably weird..

AntC2:
> ```haskell
> f x = case x of
>   (A
>   B
>   C) -> 3
>   (D
>   E -> 4
>   F) -> 5
> ```

Are you sure that's what you wanted to put? There's a mis-placed parens(?)

And yeah, not being able to indent each constructor into the same column is ugly.

proposals/0522-or-patterns.rst
```diff
@@ -294,53 +314,79 @@ So the following example would not type check because the Or pattern doesn't pro
...
 
+Additionally, the changes to the ``pattern_synonym_decl`` nonterminal are required to allow unparenthesized Or patterns in the syntax. This breaks pattern synonym declarations using a type annotation such as: ::
+
+  pattern X <- 42 :: Int
+
+One would now have to parenthesize the right side of the pattern synonym and write ::
+
+  pattern X <- (42 :: Int)
+
+A search on Hackage shows that this syntax is currently only used in very few places.
```
sgraf812, line 342:
```suggestion
A search on Hackage shows that this syntax is currently only used (without enclosing parentheses) in `haskell-src-exts-simple <https://hackage-search.serokell.io/?q=%5Epattern.*%3C-.*%3A%3A%5B%5E%5C%29%5D*%24>`_.
```

nomeata:
I'm traveling right now, please allow for a week delay

knothed:
Are there updates yet?

nomeata:
I've assigned it to @cdornan, I hope he'll have a look soon

cdornan:
@knothed sorry about the delay &mdash; I have reviewed the proposal and I like it very much.

The surface syntax you have settled on looks really nice and the way it can exploit layout is really neat. The adjustments to the grammar are simple and elegant as are the semantics. It solves a really important problem. A _lot_ of time and thought has gone in to this over multiple iterations with input collected from the community. You have an implementation patch. What is not to like?

I guess we will find out soon enough!

I will send out an email to the list this week summarizing it and recommending acceptance.

proposals/0522-or-patterns.rst
```diff
@@ -74,49 +74,80 @@ Proposed Change Specification
 Changes in the grammar
 ~~~~~~~~~~~~~~~~~~~~~~
 
+We consider this as an extension to GHC's .
 We consider this as an extension to `Haskell 2010 grammar
```
cdornan, line 78: Implementation of the proposal should not disrupt the processing of modules where the extension is not active, but implementing it by adjusting `Parser.y` in the way you are suggesting would be disruptive wouldn't it? Could you please clarify.

cdornan:
@knothed feedback from the committee has been generally positive but we do need to clarify whether the proposal would be generally disruptive, even when the extension is not active.

knothed:
@cdornan In conjunction with `-XPatternSynonyms`, the proposal would be disruptive even when `-XOrPatterns` is not active. The surface of disruption is extremely small however as outlined in the proposal.

I've slightly clarified the proposal to highlight the alternative (or-patterns with required parentheses) that avoids this disruptiveness. I believe it is now in the hands of the committee to decide which variant of or-patterns is best.

nomeata:
Is it possible to change the parsing only if `-XOrPatterns` is actually on? Otherwise we might have to go through a deprecation period for the old syntax.

sgraf812:
A simple solution I think would be to use `exp` instead of `infixpat` in `pattern_synonym_decl` (note that today, `pat -> exp` is the only production of `pat`) until we have deprecated the old syntax (which we hopefully will?). A more heavy-weight solution would be to parameterise the `pat` production by its use of `orpats`, but that kind of amounts to the same thing as inlining its old definition into the use site in `pattern_synonym_decl`. I guess we could simply introduce and use `pat_syn_pat : exp` as well.

The following patch to (somewhat old) GHC master is accepted by happy without conflict (the `%shift` declaration in `orpats` is needed in order to resolve associativity of `;`)

```
diff --git a/compiler/GHC/Parser.y b/compiler/GHC/Parser.y
index ac9976815f..82a18ef622 100644
--- a/compiler/GHC/Parser.y
+++ b/compiler/GHC/Parser.y
@@ -1627,18 +1627,18 @@ role : VARID             { sL1 $1 $ Just $ getVARID $1 }

 -- Glasgow extension: pattern synonyms
 pattern_synonym_decl :: { LHsDecl GhcPs }
-        : 'pattern' pattern_synonym_lhs '=' pat
+        : 'pattern' pattern_synonym_lhs '=' exp
          {%      let (name, args, as ) = $2 in
                  acsA (\cs -> sLL $1 (reLoc $>) . ValD noExtField $ mkPatSynBind name args $4
                                                     ImplicitBidirectional
                       (EpAnn (glR $1) (as ++ [mj AnnPattern $1, mj AnnEqual $3]) cs)) }

-        | 'pattern' pattern_synonym_lhs '<-' pat
+        | 'pattern' pattern_synonym_lhs '<-' exp
          {%    let (name, args, as) = $2 in
                acsA (\cs -> sLL $1 (reLoc $>) . ValD noExtField $ mkPatSynBind name args $4 Unidirectional
                        (EpAnn (glR $1) (as ++ [mj AnnPattern $1,mu AnnLarrow $3]) cs)) }

-        | 'pattern' pattern_synonym_lhs '<-' pat where_decls
+        | 'pattern' pattern_synonym_lhs '<-' exp where_decls
             {% do { let (name, args, as) = $2
                   ; mg <- mkPatSynMatchGroup name $5
                   ; acsA (\cs -> sLL $1 (reLoc $>) . ValD noExtField $
@@ -2882,8 +2882,14 @@ aexp    :: { ECP }
                            fmap ecpFromExp $
                            acsA (\cs -> sLLlA $1 $> $ HsProc (EpAnn (glR $1) [mj AnnProc $1,mu AnnRarrow $3] cs) p (sLLa $1 (reLoc $>) $ HsCmdTop noExtField cmd)) }

+        | '(' orpats ')' { undefined }
+
         | aexp1                 { $1 }

+orpats :: { LPat GhcPs }
+orpats : exp %shift { undefined }
+       | exp ';' orpats { undefined }
+
 aexp1   :: { ECP }
         : aexp1 '{' fbinds '}' { ECP $
                                    getBit OverloadedRecordUpdateBit >>= \ overloaded ->
@@ -3318,7 +3324,8 @@ gdpat   :: { forall b. DisambECP b => PV (LGRHS GhcPs (LocatedA b)) }
 -- Bangs inside are parsed as infix operator applications, so that
 -- we parse them right when bang-patterns are off
 pat     :: { LPat GhcPs }
-pat     :  exp          {% (checkPattern <=< runPV) (unECP $1) }
+pat     : exp          {% (checkPattern <=< runPV) (unECP $1) }
+        | pat ';' exp  {% undefined }

 -- 'pats1' does the same thing as 'pat', but returns it as a singleton
 -- list so that it can be used with a parameterized production rule
```

cdornan:
@sgraf812 I like simple solutions (as long as they aren't _too_ simple, of course) and like the idea of using '`exp` instead of `infixpat` in `pattern_synonym_decl` until we have deprecated the old syntax'.

nomeata:
> until we have deprecated the old syntax (which we hopefully will?). 

It doesn't seem to be part of the proposal yet, unless I missed it.

I’m torn between deprecating the (arguably somewhat confusing) syntax, and containing _all_ changes to when `OrPatterns` is enabled, so that no deprecation is necessary. Both are reasonable, but let’s make sure it’s clear what the proposal proposes.

adamgundry:
In the long term, it would be better if turning on `OrPatterns` wouldn't result in code randomly breaking. Conservative extensions are good! So I think we could just as well guard the parser change behind `-XOrPatterns` and still introduce a deprecation warning for the conflicting syntax?

knothed:
> I think we could just as well guard the parser change behind `-XOrPatterns`

How would that work? The grammar rules in Parser.y cannot change depending on the active extensions AFAIK.

sgraf812:
> It doesn't seem to be part of the proposal yet, unless I missed it.

Ah, so deprecation should be part of the proposal, too? Fair enough. Arguably it is an independent concern, though; we could roll with the `exp` change without risking any breaking change and propose the deprecation later on.

However I think the fact that the `exp` change works is a lucky coincidence; IMO we should strive to use some kind of `*pat` non-terminal there. I suppose we could have

```
pat -> pat_syn_pat | orpats
pat_syn_pat -> exp

pattern_synonym_decl -> ... pat_syn_pat ...
```

and that would not incur any breaking changes until someone touches `exp` the next time. I suppose they could then propose to deprecate `pat_syn_pat -> exp` in favor of `pat_syn_pat -> infixexp` or something like that.

TLDR; I think we could just introduce `pat_syn_pat` as alluded to above, then we don't risk any breaking changes. @knothed would you be up for that?

cdornan:
> Ah, so deprecation should be part of the proposal, too? Fair enough. Arguably it is an independent concern, though; we could roll with the exp change without risking any breaking change and propose the deprecation later on.

I can see both sides of this &mdash; in any case I think it would be a good idea to at least provide a roadmap in the proposal explaining how we see this playing out. Whether the depracation is trigered by this proposal or a follow-up isn't a major deal though I could see why @nometa might prefer to get it all out of the way.

nomeata:
I think it's information we need - @angerman (rightfully) pays close attention to whether there is change to existing code, and if so, how it's going to be managed.

cdornan:
@nomeata I have been asking for this information -- sorry if I gave the contrary impression.

@sgraf812 @knothed could you add a prominent section summarising the general disruption to modules (that do not enable this extension). Ideally this won't be any disruption at all, otherwise we will have to look at a depracation cycle before we can roll it out.

If you recommend that we consider disrupting the language at a later stage to clean up the grammar then could you say that also &mdash; though maybe it might be better to leave that to a later proposal (I leave it to your judgement).

I am very sorry for this late disruption &mdash; I think part of the reason for it is that you have gone so thoroughly into the details of how to realise this that some of the consequences &mdash; entirely plain to you &mdash; have become slight obscured. I am sorry for my part &mdash; I am so used to proposed language extensions only affecting the language under the extension that I failed to appreciate the issue until Moritz pointed it out &mdash; that is my bad.

yav:
Does the change to the parser lead to changing the meaning of existing programs?  At a quick glance, it seems that some programs that were previously syntactically invalid would now be parsed as or-patterns, but we could always reject these after the parser, if the extension is not enabled.

knothed:
I like the 
```
pat -> pat_syn_pat | orpats
pat_syn_pat -> exp
```
solution. It seems to work without any breaking changes.

> Does the change to the parser lead to changing the meaning of existing programs? 

I‘m gonna run the testsuite tomorrow and check. If all is good, I‘ll change the proposal to use this syntax.

aspiwack:
Something that is missing in this conversation about pattern synonym syntax is the problem that needs to be solved. The proposal should be clear about why you want to change the pattern synonym syntax (regardless of whether the change is backward compatible).

knothed:
I've updated the proposal with the change to the pattern synonym syntax and with a section why it is needed.

The change is non-breaking; no syntax is stolen. So we should be fine now!

nomeata:
Good job! Happy to see the proposal process leading to improvements, and not just friction.

knothed:
Are there any updates on this? Maybe I‘ve missed to state that the proposal is ready for review again. @cdornan

cdornan:
@knothed thanks for those revisions. I am sending it back to the committee for consideration.

simonpj:
I am happy with the payload, but please can you amend 2.1.1 to give the actual new BNF.   It's too hard to apply diffs, expressed in English, to a fragment of BNF.  Just give me the new BNF!  

Eg the text talks of "three pattern_synonym_decl rules", but the fragment actually given only shows one.  Please just write it out.

Also in 2.1 you say "NB: for the concrete implementation...". This is useful, but it belongs in Section 8 about the implementation. As I understand it, this ECP business makes no difference to the user-level specification about which programs are syntactically correct, and which are not. Right?

 And can you expand it a bit.  What is ECP?  Why do you need to modify several productions?  Giving a bit more detail would help here.

proposals/0522-or-patterns.rst
```diff
@@ -1,12 +1,12 @@
 Or Patterns
 ==============
 
-.. author:: David Knothe, Ömer Sinan Ağacan, Sebastian Graf
-.. date-accepted:: 2022-11-29
+.. author:: Sebastian Graf, David Knothe
+.. date-accepted::
```
adamgundry, line 5: I'm not sure why Ömer's name is being removed from the authors list, and the date-accepted removed, was that deliberate?

cdornan:
@knothed could you address Simon's and Adam's concerns &mdash; after that we can merge.

knothed:
@cdornan Done

PR #601, opened 2023-07-04
Extension lifecycle framework proposal

david-christiansen:
The [proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0601-extension-lifecycle-framework.md) has been accepted; the following discussion is mostly of historic interest.

---

This proposal suggests a defined lifecycle for language extensions, to more clearly communicate tacit community knowledge about which extensions make sense in which contexts. Additionally, it proposes a set of compiler warning flags that empower users of GHC to choose their own level of tolerance for language extension changes.

This proposal is a product of the Haskell Foundation's Stability Working Group.

Issue #551, opened 2022-11-13
GHC2021, TypeOperators and ExplicitNamespaces


nomeata:
In https://discourse.haskell.org/t/quo-vadit-ghc2023/5220/42?u=nomeata it was pointed out that `GHC2021` contains `TypeOperators`, but not `ExplicitNamespaces`, despite the fact that `-XTypeOperators` will also enable `ExplicitNamespaces`.

This could be

 * Intentional, given that `TypeOperators` got 8 votes, but `ExplicitNamespaces` only 4 (see <https://github.com/ghc-proposals/ghc-proposals/blob/ghc2021/proposals/0000-ghc2021.rst>). (I don’t think it was intentional in my votes)
 * An accident, and we should fix it by making `-XGHC2021` imply `ExplicitNamespaces`.
 * An accident, but we should not fix it in `GHC2021`, but rather the next `GHC20xx`

Opinions?

ysangkok:
I'd like to point out that it is also enabled by `TypeFamilies`. So it is even more popular than it seems.

It seems the vote was conducted with confusing data. I am thinking of e.g. the "Popularity" and "Proliferation" columns being lower for ExplicitNamespaces than for TypeOperators/TypeFamilies. I think it would be possible to use e.g. `LANGUAGE TypeFamilies NoExplicitNamespaces`, but that is probably not very common. So the numbers don't seem to account for extensions implying other extensions, is that true? This is unintuitive.

If one accepts that the numbers were methodically wrong, it calls all of GHC2021 into question. I submit that we shouldn't consider disabling anything already enabled by GHC2021, so let's stick with this one extension.

Among people who voted no on ExplicitNamespaces:

* Voted yes on TypeFamilies but no on TypeOperators: CC, SM
* Voted yes on TypeFamilies and TypeOperators: AM, VB

In case these all of these four people are reachable
---

Can't we just repoll these four people? If they all say yes, with the existing votes, we'd have 8 votes. If any of them thinks changing GHC2021 is impermissible, they can lay down that veto (7 votes would not be sufficient).

Why speculate on the intentions on the voters when we can just ask them?

In case any of these four people is not reachable
---

The whole committee could vote on it. I think it is better than letting non-committee members decide on something that the committee was supposed to decide.

If the GHC2021 committee can't decide on it, the results table will be complicated to read in any case. Or of course, it could be just left like this, but then we need another page documenting what GHC2021 actually means.

jumper149:
[...]

PR #613, opened 2023-09-27
GHC2024

nomeata:
[Rendered](https://github.com/ghc-proposals/ghc-proposals/blob/joachim/ghc2024/proposals/0000-ghc2024.rst)

This PR will guide the process of deciding if we want to have, and then defining, GHC2024, as per the process outlined in [#372](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0372-ghc-extensions.rst).

### DIFF
--- a/proposals/0000-ghc2024.rst
+++ b/proposals/0000-ghc2024.rst
@@ -31,6 +31,7 @@ that ``GHC2021`` comprises (`official list <https://downloads.haskell.org/ghc/la
 * `GADTs`_ (and the implied `MonoLocalBinds`_)
 * `LambdaCase`_
 * `RoleAnnotations`_
+* `TypeFamilies`_
 
 (*The committe will vote on each extension separately. Until then, this is the
 list of nominated extensions, and will be updated after a vote.*)
@@ -235,12 +236,38 @@ At the moment, using ``GHC2021`` together with `Safe`_ causes a warning, because
 *Backward compatibility*: Enabling `RoleAnnotations`_ does not affect
 existing ``GHC2021`` code.
 
+
+Why add TypeFamilies?
+---------------------
+
+Type families are one of the most used features of GHC. The reason for
+not including `TypeFamilies`_ in ``GHC2021`` was that type families
+don't work so well without `MonoLocalBinds`_, and it was considered at
+the time that ``MonoLocalBinds`` was too steep a change.
+
+But if we add ``MonoLocalBinds`` to ``GHC2024``, there is no
+obstacle to make this very popular feature.
+
+*Counterarguments*: `A reason not to include
+<https://github.com/ghc-proposals/ghc-proposals/pull/613#issuecomment-1759556663>`
+that the semantics of type families (in particular the strictness of
+its evaluation) is unsatisfactory and would like it to change before
+they become a default. But there are two possibility: either the
+semantic change is backward compatible, in which case including
+``TypeFamilies`` in ``GHC2024`` won't cause any issue; or the semantic
+change isn't backward compatible, in which case the massive popularity
+of type families makes it impossible to incorporate the change in the
+``TypeFamilies`` extension, and ``GHC2024`` is safe.
+
+*Backward compatibility*: Assuming that ``MonoLocalBinds`` is enabled,
+enabling ``TypeFamilies`` doesn't affect existing ``GHC2021`` code
+further.
+
 Alternatives
 ------------
 We could not do ``GHC2024`` and wait yet another year, or more, because we shy away from
 making what may look like a stability-threatening change.
 In my view that is worse: The fixes and improvements suggested above would reach our users later, we would not establish a regular and predictable pattern, and in the worst case never dare to make a new release, which would make the ``GHC20xx`` idea fall into a similar pattern than the ``Haskell20xx`` report process, which at the moment is stalled.
-
 
 Implementation Plan
 -------------------
@@ -260,3 +287,4 @@ Implementation Plan
 .. _RoleAnnotations: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/roles.html#extension-RoleAnnotations
 .. _Safe: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/safe_haskell.html#extension-Safe
 .. _GeneralisedNewtypeDeriving: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/roles.html#extension-GeneralisedNewtypeDeriving
+.. _TypeFamilies: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/type_families.html#extension-TypeFamilies
--- a/proposals/0000-ghc2024.rst
+++ b/proposals/0000-ghc2024.rst
@@ -23,7 +23,7 @@ Proposed Change Specification
 -----------------------------
 
 The ``GHC2024`` language extension set comprises of all language extensions
-that ``GHC2021`` comprises, plus the following
+that ``GHC2021`` comprises (`official list <https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/control.html#extension-GHC2021>`_), plus the following
 
 * `DerivingStrategies`_
 * `DisambiguateRecordFields`_
--- a/proposals/0000-ghc2024.rst
+++ b/proposals/0000-ghc2024.rst
@@ -350,7 +350,7 @@ Why add BlockArguments?
 
 `BlockArguments`_ denoises common idioms, e.g. `when` for `forM` in do blocks, by not requiring
 parentheses or `$`. The resulting non-verbose syntax is arguably very Haskelly, and some
-people working with languages that have this already (Lean, PureScript) report beeing happy about it.
+people working with languages that have this already (Agda, PureScript, Lean) report beeing happy about it.
 
 *Backward compatibility*: No incompatibilities (accoridng to the original proposal).
 
--- a/proposals/0000-ghc2024.rst
+++ b/proposals/0000-ghc2024.rst
@@ -25,12 +25,14 @@ Proposed Change Specification
 The ``GHC2024`` language extension set comprises of all language extensions
 that ``GHC2021`` comprises (`official list <https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/control.html#extension-GHC2021>`_), plus the following
 
+* `DataKinds`_
 * `DerivingStrategies`_
 * `DisambiguateRecordFields`_
 * `ExplicitNamespaces`_
 * `GADTs`_ (and the implied `MonoLocalBinds`_)
 * `LambdaCase`_
 * `RoleAnnotations`_
+* `TypeData`_
 * `TypeFamilies`_
 
 (*The committe will vote on each extension separately. Until then, this is the
@@ -54,6 +56,53 @@ Furthermore, a regular, not too slow cadence makes the whole process more
 predictable, which I expect *improves* the perception of stability, in the
 sense of delivering a stable flow of changes.
 
+
+Why add DataKinds and TypeData?
+-------------------------------
+
+We're considering ``GADTs`` and ``TypeFamilies`` for adoption. Both of
+them greatly benefit from ``DataKinds`` and ``TypeData`` (so do
+phantom types in plain Haskell 2010 for that matter). They let us
+inject more types at the type level. On the other hand if we reject ``GADTs`` and
+``TypeFamilies`` from ``GHC2024``, then we ``DataKinds`` and
+``TypeData`` may not pull their weight.
+
+That being said, `it's been suggested
+<https://github.com/ghc-proposals/ghc-proposals/pull/613#issuecomment-1761015976>`_
+that having ``DataKinds`` on yields better error messages when
+mistakenly using a data constructor in types:
+
+::
+
+  foo :: Just Int
+  foo = Just 0
+
+With ``DataKinds`` the error message is::
+
+  <interactive>:2:19: error: [GHC-83865]
+      • Expected a type, but ‘Just Int’ has kind ‘Maybe (*)’
+      • In the type signature: foo :: Int -> Just Int
+
+With ``NoDataKinds``::
+
+  <interactive>:4:19: error: [GHC-76037]
+      Not in scope: type constructor or class ‘Just’
+      Suggested fix:
+        Perhaps you intended to use DataKinds
+        to refer to the data constructor of that name?
+
+Both ``DataKinds`` and ``TypeData`` are used the same way at the
+type-level, the former lets us reuse types that we are using in
+computations at the type level, whereas the latter lets us define type
+purely for type level computations.
+
+*Counterarguments*: ``TypeData`` is still rather new, being only
+released in GHC 9.6. It's probably too early to consider that we have
+sufficient feedback on its merits. Nevertheless, it's innocuous enough
+to deserve a vote.
+
+*Backward compatibility*: Enabling `DataKinds`_ or `TypeData`_ does
+not affect existing ``GHC2021`` code.
 
 Why add DerivingStrategies?
 ---------------------------
@@ -288,3 +337,5 @@ Implementation Plan
 .. _Safe: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/safe_haskell.html#extension-Safe
 .. _GeneralisedNewtypeDeriving: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/roles.html#extension-GeneralisedNewtypeDeriving
 .. _TypeFamilies: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/type_families.html#extension-TypeFamilies
+.. _DataKinds: https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/data_kinds.html?highlight=datakinds#extension-DataKinds
+.. _TypeData: https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/type_data.html#extension-TypeData
--- a/proposals/0000-ghc2024.rst
+++ b/proposals/0000-ghc2024.rst
@@ -2,11 +2,11 @@ GHC2024
 =======
 
 .. author:: Joachim Breitner
-.. date-accepted::
-.. ticket-url::
+.. date-accepted:: 2024-01-09
+.. ticket-url:: https://gitlab.haskell.org/ghc/ghc/-/issues/24320
 .. implemented::
 .. highlight:: haskell
-.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/613>`_.
+.. header:: This proposal was `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/613>`_.
 .. sectnum::
 .. contents::
 
--- a/proposals/0000-ghc2024.rst
+++ b/proposals/0000-ghc2024.rst
@@ -26,6 +26,7 @@ The ``GHC2024`` language extension set comprises of all language extensions
 that ``GHC2021`` comprises (`official list <https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/control.html#extension-GHC2021>`_), plus the following
 
 * `DataKinds`_
+* `DefaultSignatures`_
 * `DerivingStrategies`_
 * `DisambiguateRecordFields`_
 * `ExplicitNamespaces`_
@@ -103,6 +104,28 @@ to deserve a vote.
 
 *Backward compatibility*: Enabling `DataKinds`_ or `TypeData`_ does
 not affect existing ``GHC2021`` code.
+
+Why add DefaultTypeSignatures?
+------------------------------
+
+There is no strong argument for inclusion of
+`DefaultTypeSignatures`_. But nor is there for rejecting
+it. ``DefaultTypeSignatures`` seems quite stable and there hasn't been
+any pushback against the feature. It's a situational, but useful
+extension that we don't really have expectations to change.
+
+*Counterarguments*: `It's been pointed out
+<https://github.com/ghc-proposals/ghc-proposals/pull/613#issuecomment-1759419147>`_
+that the most common use of default type signatures is with generic
+deriving and ``DeriveAnyClass`` (which is not considered for inclusion
+in ``GHC2024`` because it's seen as too error prone). An alternative,
+for this use-case would be to use deriving-via with the
+``Generically`` type class. But there are other uses, and
+`Deriving-via doesn't always work
+<https://github.com/ghc-proposals/ghc-proposals/pull/613#issuecomment-1760024807>`_.
+
+*Backward compatibility*: Enabling `DefaultTypeSignatures`_ does not
+affect existing ``GHC2021`` code.
 
 Why add DerivingStrategies?
 ---------------------------
@@ -339,3 +362,4 @@ Implementation Plan
 .. _TypeFamilies: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/type_families.html#extension-TypeFamilies
 .. _DataKinds: https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/data_kinds.html?highlight=datakinds#extension-DataKinds
 .. _TypeData: https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/type_data.html#extension-TypeData
+.. _DefaultSignatures: https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/default_signatures.html#extension-DefaultSignatures
--- a/proposals/0000-ghc2024.rst
+++ b/proposals/0000-ghc2024.rst
@@ -36,6 +36,7 @@ that ``GHC2021`` comprises (`official list <https://downloads.haskell.org/ghc/la
 * `RoleAnnotations`_
 * `TypeData`_
 * `TypeFamilies`_
+* `BlockArguments`_
 
 (*The committe will vote on each extension separately. Until then, this is the
 list of nominated extensions, and will be updated after a vote.*)
@@ -344,6 +345,15 @@ of type families makes it impossible to incorporate the change in the
 enabling ``TypeFamilies`` doesn't affect existing ``GHC2021`` code
 further.
 
+Why add BlockArguments?
+-----------------------
+
+`BlockArguments`_ denoises common idioms, e.g. `when` for `forM` in do blocks, by not requiring
+parentheses or `$`. The resulting non-verbose syntax is arguably very Haskelly, and some
+people working with languages that have this already (Lean, PureScript) report beeing happy about it.
+
+*Backward compatibility*: No incompatibilities (accoridng to the original proposal).
+
 Alternatives
 ------------
 We could not do ``GHC2024`` and wait yet another year, or more, because we shy away from
@@ -372,3 +382,4 @@ Implementation Plan
 .. _DataKinds: https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/data_kinds.html?highlight=datakinds#extension-DataKinds
 .. _TypeData: https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/type_data.html#extension-TypeData
 .. _DefaultSignatures: https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/default_signatures.html#extension-DefaultSignatures
+.. _BlockArguments: https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/block_arguments.html#extension-BlockArguments
--- a/proposals/0000-ghc2024.rst
+++ b/proposals/0000-ghc2024.rst
@@ -106,11 +106,11 @@ to deserve a vote.
 *Backward compatibility*: Enabling `DataKinds`_ or `TypeData`_ does
 not affect existing ``GHC2021`` code.
 
-Why add DefaultTypeSignatures?
+Why add DefaultSignatures?
 ------------------------------
 
 There is no strong argument for inclusion of
-`DefaultTypeSignatures`_. But nor is there for rejecting
+`DefaultSignatures`_. But nor is there for rejecting
 it. ``DefaultTypeSignatures`` seems quite stable and there hasn't been
 any pushback against the feature. It's a situational, but useful
 extension that we don't really have expectations to change.
@@ -125,7 +125,7 @@ for this use-case would be to use deriving-via with the
 `Deriving-via doesn't always work
 <https://github.com/ghc-proposals/ghc-proposals/pull/613#issuecomment-1760024807>`_.
 
-*Backward compatibility*: Enabling `DefaultTypeSignatures`_ does not
+*Backward compatibility*: Enabling `DefaultSignatures`_ does not
 affect existing ``GHC2021`` code.
 
 Why add DerivingStrategies?
--- a/proposals/0000-ghc2024.rst
+++ b/proposals/0000-ghc2024.rst
@@ -22,24 +22,58 @@ The first edition, ``GHC2021``, is now three years old, and I propose we define
 Proposed Change Specification
 -----------------------------
 
-The ``GHC2024`` language extension set comprises of all language extensions
+The ``GHC2024`` language extensions set comprises of all language extensions
 that ``GHC2021`` comprises (`official list <https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/control.html#extension-GHC2021>`_), plus the following
 
 * `DataKinds`_
-* `DefaultSignatures`_
 * `DerivingStrategies`_
 * `DisambiguateRecordFields`_
 * `ExplicitNamespaces`_
 * `GADTs`_ (and the implied `MonoLocalBinds`_)
-* `ImpredicativeTypes`_
 * `LambdaCase`_
 * `RoleAnnotations`_
-* `TypeData`_
-* `TypeFamilies`_
-* `BlockArguments`_
 
-(*The committe will vote on each extension separately. Until then, this is the
-list of nominated extensions, and will be updated after a vote.*)
+Tally
+-----
+
+The committee held a vote among nominated extensions with the following result. As per the GHC20xx process, quorum was ⅔, so 7 votes.
+
+============================= ========= =======================================
+                    Extension     Votes In favor
+============================= ========= =======================================
+        `DerivingStrategies`_   **10**  VZ, AS, SPJ, SM, AG, RE, ES, MA, CD, JB
+  `DisambiguateRecordFields`_   **10**  VZ, AS, SPJ, SM, AG, RE, ES, MA, CD, JB
+ `GADTs`_ + `MonoLocalBinds`_   **10**  VZ, AS, SPJ, SM, AG, RE, ES, MA, CD, JB
+                `LambdaCase`_   **10**  VZ, AS, SPJ, SM, AG, RE, ES, MA, CD, JB
+        `ExplicitNamespaces`_    **9**  AS, SPJ, SM, AG, RE, ES, MA, CD, JB
+           `RoleAnnotations`_    **8**  AS, SPJ, SM, AG, RE, ES, CD, JB
+                 `DataKinds`_    **8**  AS, SPJ, SM, AG, RE, ES, CD, JB
+              `TypeFamilies`_        6  AS, SPJ, SM, RE, ES, JB
+         `DefaultSignatures`_        4  AS, SM, RE, ES
+                  `TypeData`_        4  SPJ, RE, ES, JB
+            `BlockArguments`_        4  VZ, RE, ES, JB
+        `ImpredicativeTypes`_        1  AS
+============================= ========= =======================================
+
+NB: `ImpredicativeTypes`_ was accidentally missing on the original ballot, and
+some committee members may have forgotten to vote. But with explicit votes against from
+AG, SPJ, RE, JB it could not have made quorum.
+
+The voting committee members are
+
+=== ======================
+SM  Simon Marlow
+SPJ Simon Peyton-Jones
+JB  Joachim Breitner
+RE  Richard Eisenberg
+VZ  Vladislav Zavialov
+ES  Eric Seidel
+CD  Chris Dornan
+AS  Arnaud Spiwack
+AG  Adam Gundry
+MA  Moritz Angermann
+=== ======================
+
 
 Why GHC2024 in the first place
 ------------------------------
@@ -383,3 +417,4 @@ Implementation Plan
 .. _TypeData: https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/type_data.html#extension-TypeData
 .. _DefaultSignatures: https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/default_signatures.html#extension-DefaultSignatures
 .. _BlockArguments: https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/block_arguments.html#extension-BlockArguments
+.. _ImpredicativeTypes: https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/impredicative_types.html#extension-ImpredicativeTypes
--- a/proposals/0000-ghc2024.rst
+++ b/proposals/0000-ghc2024.rst
@@ -38,22 +38,22 @@ Tally
 
 The committee held a vote among nominated extensions with the following result. As per the GHC20xx process, quorum was ⅔, so 7 votes.
 
-============================= ========= =======================================
-                    Extension     Votes In favor
-============================= ========= =======================================
-        `DerivingStrategies`_   **10**  VZ, AS, SPJ, SM, AG, RE, ES, MA, CD, JB
-  `DisambiguateRecordFields`_   **10**  VZ, AS, SPJ, SM, AG, RE, ES, MA, CD, JB
- `GADTs`_ + `MonoLocalBinds`_   **10**  VZ, AS, SPJ, SM, AG, RE, ES, MA, CD, JB
-                `LambdaCase`_   **10**  VZ, AS, SPJ, SM, AG, RE, ES, MA, CD, JB
-        `ExplicitNamespaces`_    **9**  AS, SPJ, SM, AG, RE, ES, MA, CD, JB
-           `RoleAnnotations`_    **8**  AS, SPJ, SM, AG, RE, ES, CD, JB
-                 `DataKinds`_    **8**  AS, SPJ, SM, AG, RE, ES, CD, JB
-              `TypeFamilies`_        6  AS, SPJ, SM, RE, ES, JB
-         `DefaultSignatures`_        4  AS, SM, RE, ES
-                  `TypeData`_        4  SPJ, RE, ES, JB
-            `BlockArguments`_        4  VZ, RE, ES, JB
-        `ImpredicativeTypes`_        1  AS
-============================= ========= =======================================
+============================= ========= ==  ==  ===  ==  ==  ==  ==  ==  ==  ==
+                    Extension     Votes VZ  AS  SPJ  SM  AG  RE  ES  MA  CD  JB
+============================= ========= ==  ==  ===  ==  ==  ==  ==  ==  ==  ==
+        `DerivingStrategies`_   **10**  ✓    ✓  ✓    ✓   ✓   ✓   ✓   ✓   ✓   ✓
+  `DisambiguateRecordFields`_   **10**  ✓    ✓  ✓    ✓   ✓   ✓   ✓   ✓   ✓   ✓
+ `GADTs`_ + `MonoLocalBinds`_   **10**  ✓    ✓  ✓    ✓   ✓   ✓   ✓   ✓   ✓   ✓
+                `LambdaCase`_   **10**  ✓    ✓  ✓    ✓   ✓   ✓   ✓   ✓   ✓   ✓
+        `ExplicitNamespaces`_    **9**       ✓  ✓    ✓   ✓   ✓   ✓   ✓   ✓   ✓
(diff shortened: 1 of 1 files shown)


nomeata:
Fall is coming close, and I’d like to start the process of defining GHC2024. 

In last year’s attempt (https://github.com/ghc-proposals/ghc-proposals/pull/559) the vote was not to define a new language edition, but I hope that we can have it now. I still think that we should strive to produce a _regular_ stream of “This is modern haskell” language extensions, given that they do little harm (existing code is unaffected), and we don't want a new language edition release to be something that scares people. At least no more than a GHC release itself.

I hope we can do this without too much entanglement with the extension classificaiton proccess (#601) and the [stability guarantee discussion](https://docs.google.com/document/d/1wtbAK6cUhiAmM6eHV5TLh8azEdNtsmGwm47ZulgaZds/view). We should only include extensions that we consider “stable” in some sense, but I hope we can make that call without finalizing the other pieces.

I’ve copied the proposal over from last year, with four suggestions. The committee will eventually vote upon them individually, so including one that doesn't make it shouldn’t affect  the other extensions’ prospect.

I invite interested parties who have good arguments for additional changes to submit them as PRs or “suggested edits” against this branch. Committee members can just commit directly. If you have further refinements to the pros and cons of an extension, please also add them.

adamgundry:
Thanks for kicking this off @nomeata! I have made some small edits and suggested adding `DisambiguateRecordFields`.

divarvel:
Would `DerivingStrategies` be a good candidate for inclusion in GHC2024? With GND part of GHC2021 I think it makes sense to have a way to be explicit about it.

I personally use it systematically with newtypes and even have contributed a hlint hint to enforce it.

Kleidukos:
Agreed, DerivingStrategies is part of the dialect of a number of organisations using Haskell and I can't imagine writing (and publishing) Haskell code without it.

adamgundry:
I agree `DerivingStrategies` would be a good addition; I've added it to the proposal text.

amesgen:
Another potential candidate for inclusion: [`LexicalNegation`](https://downloads.haskell.org/ghc/9.6.3/docs/users_guide/exts/lexical_negation.html). It solves a small but very real syntactic anomaly (see the linked examples), and I used it every since it became available and think there is no reason not to enable it everywhere. Still, given that it was introduced in 9.0, it is maybe too young? (OTOH, eg `ImportQualifiedPost` was added in 8.10 and is already in `GHC2021`).

nomeata:
Certainly worth considering! It received [few votes in 2021](https://github.com/ghc-proposals/ghc-proposals/blob/ghc2021/proposals/0000-ghc2021.rst#id3), but maybe just because there were so many extensions considered back then. Do you want to propose concrete wording for the section?

phadej:
@amesgen 

> This means that (- x) is the right operator section of subtraction, whereas (-x) is the negation of x. 

That is terrible. I don't see that kind of syntax go through in untyped language where there are no types to catch mistakes. But having types doesn't make it a good syntax. I'm not a fan of whitespace sensitive expression syntax. It's already terrible with `@`. I'd wish GHC steer back to be whitespace insensitive (sans layout).

phadej:
I also opened a GHC issue, https://gitlab.haskell.org/ghc/ghc/-/issues/24062 `NegativeLiterals` alone can change the meaning of programs. That's a corner case, but I actually used it this month.

amesgen:
> > This means that (- x) is the right operator section of subtraction, whereas (-x) is the negation of x.
> 
> That is terrible. I don't see that kind of syntax go through in untyped language where there are no types to catch mistakes. But having types doesn't make it a good syntax. I'm not a fan of whitespace sensitive expression syntax. It's already terrible with `@`. I'd wish GHC steer back to be whitespace insensitive (sans layout).

Thanks for mentioning that! Just FTR: There are even more operators that are already whitespace-sensitive since [proposal 229](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0229-whitespace-bang-patterns.rst) (which also introduced `LexicalNegation`).

In that light, I think it makes to still wait until a future iteration of the `GHCXXXX` process whether there is any movement to go back to whitespace insensitivity, and only to revisit `LexicalNegation` (hence embracing whitespace sensitivity) if that turned out not to be the case.

phadej:
> any movement

I don't expect any large movement. It feels we are approaching a local optima of the syntax evolution: there is very little one can do without *breaking something*. 

Off-topic: I agree that `-identifier` and `- identifier` can be understood as `@tyvar` and `@ opapply`, that would be consistent. But then `-literal` will be different than `let x = literal in -literal`, but `typelet x = literal in Proxy @x` and `Proxy @literal` work the same (if we had `typelet, but say emulating with `~` constraint).

So i back up my complaint. I'd say that `LexicalNegation` **without** `NegativeLiterals` make sense, it's consistent with other whitespace sensitivity in GHC (which I don't like, but it's there). `NegativeLiterals` however break code (it was an opportunity to introduce a class(es) *just for* literal desugaring, but it was missed).

amesgen:
> So i back up my complaint. I'd say that `LexicalNegation` **without** `NegativeLiterals` make sense, it's consistent with other whitespace sensitivity in GHC (which I don't like, but it's there). `NegativeLiterals` however break code (it was an opportunity to introduce a class(es) _just for_ literal desugaring, but it was missed: `fromNatural` for normal desugaring, separate fromInteger for `NegativeLiterals` - then I could opt my type to not work with `NegativeLiterals`).

That makes a lot of sense; I don't know the exact historical details, but judging from 

> Under [`LexicalNegation`](https://downloads.haskell.org/ghc/9.6.3/docs/users_guide/exts/lexical_negation.html#extension-LexicalNegation), negated literals are desugared without negate. That is, `-123` stands for `fromInteger (-123)` rather than `negate (fromInteger 123)`. This makes [`LexicalNegation`](https://downloads.haskell.org/ghc/9.6.3/docs/users_guide/exts/lexical_negation.html#extension-LexicalNegation) a valid replacement for [`NegativeLiterals`](https://downloads.haskell.org/ghc/9.6.3/docs/users_guide/exts/negative_literals.html#extension-NegativeLiterals).

([src](https://downloads.haskell.org/ghc/9.6.3/docs/users_guide/exts/lexical_negation.html)) the motivation might just have been for `LexicalNegation` to be a strict superset of `NegativeLiterals`? I think [proposal 229](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0229-whitespace-bang-patterns.rst) does not specify this, and I also don't see any related discussion in the [implementation PR](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/2768).

In that case, maybe `LexicalNegation` and `NegativeLiterals` should be made fully orthogonal, ie such that with `-XLexicalNegation -XNoNegativeLiterals`, `-1` desugars to `negate (fromInteger 1)`. But this is getting off-topic for this PR :smile:

nomeata:
We should probably set out a timeline for this process.

I suggest to collect nominations and rationales until the end of November, start the committee discussion November 1st with a vote early December. Hopefully this means the implementation makes it to [9.10](https://gitlab.haskell.org/ghc/ghc/-/milestones/380#tab-issues).

@adamgundry, you seem to visibly care the most. WDYT?

nomeata:
> Another potential candidate for inclusion: LexicalNegation. It solves a small but very real syntactic anomaly (see the linked examples, or some anecdotal evidence here), and I used it every since it became available and think there is no reason not to enable it everywhere

It seems to be a bit more contentious than initially thought. Do you still want to push for it, or shall we use that discussion bandwidth for something else?

adamgundry:
> I suggest to collect nominations and rationales until the end of October, start the committee discussion November 1st with a vote early December. Hopefully this means the implementation makes it to [9.10](https://gitlab.haskell.org/ghc/ghc/-/milestones/380#tab-issues).

Thanks @nomeata, this sounds good to me given the 9.10 schedule.

dbeacham:
Is it worth having a list of all extensions annotated with which are included and nominated for GHC2021 and GHC2014 respectively? Discovery of what's missing from GHC2021 is a bit difficult otherwise.

Sorry if I've missed such a list.

[All current extension in 9.6.3](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/table.html)

nomeata:
> included in GHC20201

See <https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/control.html#extension-GHC2021> for the official list (will add link to the proposal)

> nominated for GHC2024

see the proposal <https://github.com/ghc-proposals/ghc-proposals/blob/joachim/ghc2024/proposals/0000-ghc2024.rst#proposed-change-specification>

JakobBruenker:
> -XTypeInType has become absolutely standard.

-XTypeInType is actually deprecated and now only acts as a synonym for -XPolyKinds, -XDataKinds, and -XKindSignatures

aspiwack:
@JakobBruenker Oh, right. I'd _completely_ forgotten about that, thanks. Well, only `-XDataKinds` is missing, and I was already proposing inclusion, so it's just one less line in my suggestion.

nomeata:
@aspiwack if you are feeling confident about your nomination, please edit the proposal directly, and maybe include some rationale.

jvanbruegge:
I'd like to make a case against `DefaultSignatures` and `DeriveAnyClass` (I agree on the rest of the suggestions from @aspiwack ).
`DeriveAnyClass` is only useful together with `DefaultSignatures` and dangerous if used without, creating runtime errors (especially without `-Wall`).

`DefaultSignatures` latter is mainly used to provide an instance based on `GHC.Generics`. The main problem with it is that it does not compose well. The "blessed" instance needs to be added by the author of the type class and you cannot easily add other default implementations after the fact (e.g. based on another generic representation).

Given that the `Generically` newtype is now in base, GHC should promote using `DerivingVia` instead and eventually deprecate `DefaultSignatures` and `DeriveAnyClass`

aspiwack:
> `DefaultSignatures` latter is mainly used to provide an instance based on `GHC.Generics`. The main problem with it is that it does not compose well. The "blessed" instance needs to be added by the author of the type class and you cannot easily add other default implementations after the fact (e.g. based on another generic representation).

Default methods don't compose particularly well in general, but they're still useful. I don't see any harm that having `-XDefaultTypeSignature` can create, and I can say with high confidence that I use them often, and not particularly for generic-deriving and derive any class (though commonly in type classes that have attached type families).

I agree, on the other hand, that `-XDeriveAnyClass` is dangerous. It wouldn't be if we had to explicitly call the `anyclass` strategy. But it's not possible without `-XDeriveAnyClass` activated, which is unfortunate. I don't imagine that we can solve this conundrum satisfactorily in time for GHC2024. So it'll be left unsolved.

Kleidukos:
@aspiwack I'd like to challenge (in a friendly manner) your nomination for `DataKinds`, as it appears that a certain number of times, [`TypeData`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/type_data.html#extension-TypeData) provides the behaviour that is expected without polluting the types namespace. Would you be open to this alteration of your original nomination?

Kleidukos:
> Given that the `Generically` newtype is now in base, GHC should promote using `DerivingVia` instead

@jvanbruegge This is an interesting project. Would you be open to write something on how to adopt `Generically`? I'm afraid right now the reference for it is a curious mix of example, mathematical terminology and explanation (reference, guide and explanation being three different kinds of documentation that ought to live their own life, cf the [Diátaxis Framework](https://diataxis.fr/)).

I'd be more than happy to integrate it in a visible place to promote such a migration.

MangoIV:
I’d be a huge fan of a GHC2024 that includes so many nice language extensions, I would like to leave a plea though:
Could we have a community survey for this? Best would even be if that could be targeted, I.e. who of the stakeholders want what extensions for what reasons. I think this will make the decision for the committee much easier, too.

MangoIV:
> Given that the Generically newtype is now in base, GHC should promote using DerivingVia instead and eventually deprecate DefaultSignatures and DeriveAnyClass

please mind that deriving via Gernically cannot achieve the same things as deriving anyclass. It’s strictly less powerful.

MangoIV:
Also may I suggest `-XBlockArguments`?

Ericson2314:
I agree with @phadej that we shouldn't include `TypeFamilies`. I don't like their design on a language level, and, as @cgibbard says to me, "most code I see that uses type families I wish it did not" so I don't really feel the need to promote them as "the best we have today either".

isomorpheme:
> > Given that the Generically newtype is now in base, GHC should promote using DerivingVia instead and eventually deprecate DefaultSignatures and DeriveAnyClass
> 
> please mind that deriving via Generically cannot achieve the same things as deriving anyclass. It’s strictly less powerful. 

@MangoIV Do you have an illustrative example?

nomeata:
> Also may I suggest -XBlockArguments?

You can, preferably as a “suggested edit” or a PR against the text with a rationale why this extension should really be the default these days.

(I’m personally a bit hesitant to include syntactic extensions with high contention potential just yet, and would lean towards demonstrating that we _can_ get a continous stream of GHC20xx out, and after having some practice tackle some of the possibly bigger ones. But that plan of crawl-before-you-work would have worked better if we had had GHC2022 and GHC2023, so I also sympathize with not wanting to wait much longer to get stuff like `BlockArguments` in.)

MangoIV:
> > > Given that the Generically newtype is now in base, GHC should promote using DerivingVia instead and eventually deprecate DefaultSignatures and DeriveAnyClass
> > 
> > 
> > please mind that deriving via Generically cannot achieve the same things as deriving anyclass. It’s strictly less powerful.
> 
> @MangoIV Do you have an illustrative example?

```hs
{-# LANGUAGE DefaultSignatures #-}
{-# LANGUAGE DeriveAnyClass #-}
{-# LANGUAGE DerivingStrategies #-}
{-# LANGUAGE DerivingVia #-}

module Anyclass where

class C a where
  c :: proxy a -> Int
  default c :: proxy a -> Int
  c = const 42

newtype Foo a = MkFoo a

instance C Int where
  c = const 42

newtype A = MkA Int
  deriving anyclass (C)

newtype B = MkB Int
  deriving (C) via Foo B
```

MangoIV:
> You can, preferably as a “suggested edit” or a PR against the text with a rationale why this extension should really be the default these days.

@nomeata thank you, I first wanted to motivate a discussion on whether it would be useful (and wanted) to do so here and then if it turns out to be add it to the proposal.

aspiwack:
@Kleidukos 

> @aspiwack I'd like to challenge (in a friendly manner) your nomination for `DataKinds`, as it appears that a certain number of times, [`TypeData`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/type_data.html#extension-TypeData) provides the behaviour that is expected without polluting the terms namespace. Would you be open to this alteration of your original nomination?

Is there any downside to having both? (with the understanding that `-XTypeData` is probably too fresh to make it into GHC2024, but thinking about the future)

MangoIV:
I'd say it would even be advantageous to have `-XDataKinds` on, especially for learners, what I see often is 
```hs 
foo :: Just Int 
foo = _
```
if `-XDataKinds` is off, this just asks you to enable it, offering no help in solving the issue, whereas when it's on, it gives you helpful information. "Expected a type, but Just Int has kind Maybe Type"

tomjaguarpaw:
More direct link to ["opt-in principle"](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0378-dependent-type-design.rst#learnability-of-haskell).

aspiwack:
@JakobBruenker 

> It seems worth pointing out that under "opt-in principle", [#378](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0378-dependent-type-design.rst) uses -XDataKinds as an example of an extension that probably shouldn't be enabled by default.

I don't see anywhere in the opt-in principle saying that `-XDataKinds` shouldn't be enabled by default. It laments that the error message _without_ `-XDataKinds` violates the principle laid out. That being said, the principle does posit the existence of at least one extension that would be perpetually off by default, so I see how you may want to extend it to `-XDataKinds` being perpetually off by default.

Now, I believe that the vast majority of the the GHC committee, including the author of the opt-in principle paragraph cited above, has shifted to wanting to minimise the number of extensions which are perpetually off by default. Maybe the principle itself should be rephrased some to reflect these new priorities.

But for the case at hand, we can go back to first principles. The opt-in principle is largely concerned with the learnability of Haskell, so the question is: is the error messages you get with `-XDataKinds` turned on helpful to beginners?

JakobBruenker:
@aspiwack I will admit after thinking about it for a bit I can't think of any situations that newcomers are likely to run into where -XDataKinds makes the error message worse. I'm curious if others can think of any such examples.

phadej:
@JakobBruenker I do following mistake often enough, I "learned" both errors; and I don't think the either error is particularly great (the suggested fix for `-XNoDataKinds` is not helpful).

```
Prelude> :set -XDataKinds 
Prelude> let foo :: Int -> Just Int; foo = Just

<interactive>:2:19: error: [GHC-83865]
    • Expected a type, but ‘Just Int’ has kind ‘Maybe (*)’
    • In the type signature: foo :: Int -> Just Int
```

```
Prelude> :set -XNoDataKinds 
Prelude> let foo :: Int -> Just Int; foo = Just

<interactive>:4:19: error: [GHC-76037]
    Not in scope: type constructor or class ‘Just’
    Suggested fix:
      Perhaps you intended to use DataKinds
      to refer to the data constructor of that name?
```

telser:
I think we should be _very_ conservative in adding new extensions to the default behavior.

As someone who is not part of the committee I'm including some thoughts on extensions discussed so far below.

Of the extensions in the proposal:
- `DerivingStrategies`
  The primary drawback here seems to be the increased complexity in the language, which is the cost of being more explicit. Something that I would _prefer_ is to see `-Wmissing-deriving-strategies` move to `-Wcompat` for a few releases first and if we want a world where strategies are the default path. That would allow for a world where this is certainly hammered on by the community before being a default.

- `DisambiguateRecordFields`
  This easily feels like one of the lowest cost/risk extensions to add.
  
- `ExplicitNamespaces`
  This has been confusing in GHC2021 and I would hope to see that situation fixed at a minimum for the next revision. Amending the prior proposals to have a new extension and avoid changing this one definitely feels like the right path to me.
  
- `GADTs+MonoLocalBinds`
  This situation is sticky and we need to be very careful in proceeding to balance the stability of the ecosystem with an understandable desire for the feature. I think we would need some case studies to understand how wide the impact of `MonoLocalBinds` would be, for industry use, not just hackage. Repeating the situation of `DeepSubsumption` should be avoided, but so should the `GHC2021` and `ExplicitNamespaces`.
  
- `LambdaCase`
  The relative instability here definitely feels like a disqualifier to me. And unlike `ExplicitNamespaces`, I don't think there is a good way to split behavior, particularly not retroactively now.
  
- `RoleAnotations`
  Is this stable? It clearly plays into GND, but it isn't clear to me the impact on error messages, type inference, or anything else. Given the wording included here, it feels like the new warning `-Wmissing-role-annotations` should be on by default to ensure correctness. But that again would need some time in `-Wcompat`. Also, how does the role annotation work with incoherent pragmas instead of as an extension? This feels like exactly the kind of thing I want as a default, but that needs more polish before sending to the masses. Some of that is solved by documentation, but it also isn't clear to me at this point if the syntax in the current extension is burdensome for big applications, confusing to newcomers or otherwise in need of improvement.
  
For others:

I again think we should have a high bar to pass to be on by default. Error messages, documentation, backwards compatibility, and stability all need to be heavily considered. And I certainly don't think we have any discussed so far that meet that high bar.

simonpj:
>
> ExplicitNamespaces
> This has been confusing in GHC2021 and I would hope to see that situation
> fixed at a minimum for the next revision. Amending the prior proposals to
> have a new extension and avoid changing this one definitely feels like the
> right path to me
>

I'm behind the curve here.  What problem has arisen?  What are possible
solutions?

We'd need a proposal from someone if the best solution is another
extension.  (The danger is in grandfathering in bad designs; but we should
evaluate that when we have a concrete proposal.)

Simon

simonpj:
>
> GADTs+MonoLocalBinds
> This situation is sticky and we need to be very careful in proceeding to
> balance the stability of the ecosystem with an understandable desire for
> the feature. I think we would need some case studies to understand how wide
> the impact of MonoLocalBinds would be, for industry use, not just
> hackage. Repeating the situation of DeepSubsumption should be avoided,
> but so should the GHC2021 and ExplicitNamespaces.
>

MonoLocalBinds will certainly break some programs that compile with
NoMonoLocalBinds.  But the solution is easy: compile with -XGHC2021.
That's what (GR1) says.

This would not repeat the DeepSubsunption episode, because GHC2021 doesn't
include GADTs.  (Deep subsumption was initially a change you could not
avoid.)

Is that not right?

Simon

AntC2:
> @aspiwack ... we can go back to first principles. The opt-in principle is largely concerned with the learnability of Haskell, so the question is: is the error messages you get ...

As a side-note to al this fretting about principles, I discover

```
{-# LANGUAGE   Haskell2010,
               NoEmptyDataDecls    #-}         -- (A)

module H2010  where

    data EmptyD     -- empty data decls incl in H2010
                    -- but rejected here
    
```

(Tested at GHC 8.10.7. Presumably in GHC 9, setting `GHC20xx` but `NoEmptyDataDecls` or `NoScopedTypeVariables` will give you most but not all of some language level.) Furthermore:

```
{-# LANGUAGE   Haskell2010,
               NoEmptyDataDecls
               EmptyDataDecls      #-}         -- (B)

```

vs

```
{-# LANGUAGE   Haskell2010,
               EmptyDataDecls,
               NoEmptyDataDecls    #-}         -- (C)
```

GHC accepts (B) and its EmptyDataDecl with no mention of the contradictory extensions; rejects (C)'s EmptyDataDecl, again with no mention of the contradiction. So it's the last-mentioned polarity of some extension that 'sticks'(?)

Then does it really matter what's included in `GHC20xx`? If I (as beyond the learner level) take strong exception to some poorly-designed (IMO) extension -- like `-XScopedTypeVariables` [**] -- I can make my module header less cluttered by taking most but not all of some level.

Then the discussion here, which is chiefly concerned for learners, can leave those with qualms to fine-tune for themselves.

[**] Actually not: `-XScopedTypeVariables` is nowadays equivalent to `-XExplicitForAll`; but `-XNoScopedTypeVariables` seems to be a no-op: you have to set `-XNoExplicitForAll`.

nomeata:
Right, at least at the moment GHC20xx is merely a set of flags that you can further tune, as in GHC, the last flag wins. I think good defaults matter (a lot), and not just for learners, but anyone who thinks they don't is welcome to ignore this discussion.

safinaskar:
@nomeata and others, please, be sure not to repeat mistakes done with GHC2021:

- Be sure not to get into misunderstanding, such as this: https://mail.haskell.org/pipermail/ghc-steering-committee/2020-December/001997.html . And here is response by nomeata: https://mail.haskell.org/pipermail/ghc-steering-committee/2020-December/002001.html ("Sorry for the confusion, next time round will be easier"). Also keep in mind a lot of corner cases, such as this: https://mail.haskell.org/pipermail/ghc-steering-committee/2020-December/002014.html
- (Also on implications between extensions by nomeata: https://mail.haskell.org/pipermail/ghc-steering-committee/2020-December/002073.html )
- Very helpful comment on this topic: https://github.com/ghc-proposals/ghc-proposals/pull/380#issuecomment-750273391
- And despite all these precautions taken above GHC2021 still turned out to be inconsistent, and this was found after two years (!!!) after implementing GHC2021: https://github.com/ghc-proposals/ghc-proposals/issues/551 . "it was pointed out that GHC2021 contains TypeOperators, but not ExplicitNamespaces, despite the fact that -XTypeOperators will also enable ExplicitNamespaces". Fix this and make sure nothing similar will happen again

simonpj:
@safinaskar thanks for these pointers.   What are the general lessons? I can see one:

* If X is in GHC20204, and X implies Y, then Y should be in GHC2024.

Another possible one is

* GHC2024 should be a superset of GHC2021. 

I am not sure whether Joachim esposes this one.  But if we don't follow it, we should be super clear what we are removing from GHC2021 and why.

Any others?

TeofilC:
Regarding DataKinds, I made a GHC ticket about that somewhat unhelpful error message a while ago: https://gitlab.haskell.org/ghc/ghc/-/issues/21031

rhendric:
I would like to *very meekly* object to DataKinds being included by default. It's entirely possible there's an argument that will convince me that DataKinds should be enabled whenever TypeData is, or that GADTs/TypeFamilies require both of them to be their best selves, but based on what I know now, personally I like having TypeData on by default and DataKinds off by default.

Both extensions are ‘opt-in’ in correct code, but when you've made a mistake and used a data constructor where you meant to use a type constructor, having DataKinds enabled means the compiler will assume that you meant what you wrote and start unification, and maybe that leads to an error in a different location; whereas having DataKinds disabled means the compiler points out the incorrect constructor where it occurs.

TypeData has no such issue—TypeData gets involved if and only if your code uses the keywords `type data`, which is harder to do by accident than using a constructor from the wrong namespace.

(I'm only here because [the poll](https://discourse.haskell.org/t/ghc2024-community-input/8168) didn't give me the option to support one but not the other.)

Kleidukos:
Seconded, `TypeData` is an excellent refinement of `DataKinds` and should be encouraged.

MangoIV:
I think there’s a strong point for DataKinds, namely promoted lists and Symbols which are in DataKinds and probably the most used promoted types.

AntC2:
> based on what I know now, personally I like having TypeData on by default and DataKinds off by default.

Could I draw your attention to @nomeata's ["GHC20xx is merely a set of flags that you can further tune, "](https://github.com/ghc-proposals/ghc-proposals/pull/613#issuecomment-1782367590).

I understand this debate to be setting defaults for the bulk of users -- especially beginners who might get overwhelmed by having to choose on an extension-by-extension basis. As I understand it, you could set `-GHC2024` _and_ `-XNoDataKinds` (providing that comes later amongst the options). Would that work for you?

(In my case, I'm always going to set `-XOverlappingInstances` and `-XFunctionalDependencies`, so `-GHC2024` would be of marginal benefit -- because I'd also override to `-XNoScopedTypeVariables`.)

rhendric:
> As I understand it, you could set `-GHC2024` _and_ `-XNoDataKinds` (providing that comes later amongst the options). Would that work for you?

Of course it'd work. It'd also work for others to set `-GHC2024` and `-XDataKinds`. The question is whether one is ‘better’, with, I think, a bit of a bias toward the Haskell beginner's experience.

For beginners in particular (but also for me, and I'm no beginner), I believe that being told ‘You didn't mean to use `Just` here, unless `Just` is the name of a type that you haven't imported’ would be better than being told ‘You have a type error: `Maybe a` is not the same as `Type`’. (Neither error message on offer is quite that clear, but that's a problem to be solved in the error messages, not by enabling advanced language features by default.)

But that's just my opinion, and the outcome won't stop me from customizing my extension list either way.

aspiwack:
For the sake of clarity, I would like to stress that GHC20XX editions are emphatically not targeted at beginners. Of course beginners are taken into account, but most Haskell programmers are not beginners; and with these editions we want to make their lives better. It's a discussion about defaults, not just about learnability.

So for beginners experience to influence the default, I'd say it should be noticeably worse. Like really bad. And in the case of DataKind, the available evidence say that it may be a little worse sometimes, and a little better other times. This, to me, is too weak an argument against inclusion.

AntC2:
> GHC20XX editions are emphatically not targeted at beginners.

Oh! That's news to me. Non-beginners can readily hand-craft their options, I would have thought. (That's why I don't use `GHC20XX`.) H2010 is not fit-for-purpose for beginners. If `GHC20XX` is not targeted at providing beginners with an 'assured' platform, I can't see the point of it at all -- especially all the energy discussing it seems to drain.

Kleidukos:
Indeed there seems to be a disconnect, GHC2021 has been most useful to bundle extensions as a language edition to provide quality of life improvements in a seamless manner for beginners.
I'm surprised to hear about it only now.

AntC2:
> I'm surprised to hear about it only now.

I suppose the [docos](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/control.html#extension-GHC2021) say merely

> GHC blesses a number of extensions, beyond Haskell 2010, to be suitable to turned on by default. These extensions are considered to be stable and conservative.

OTOH that "stable and conservative" might be useful to beginners especially seems to be a common misconception ([@nomeata  above](https://github.com/ghc-proposals/ghc-proposals/pull/613#issuecomment-1782367590)):

>  I think good defaults matter (a lot), and not just for learners, ...

That isn't "not at all for learners".

"good" is doing a lot of heavy lifting in that comment -- as is @aspiwack's "we want to make their [most Haskell programmers] lives better."

`GHC20XX` is not making my Haskell life better; it introduces extensions I actively hate; I don't use it. It doesn't seem to make Haskell programmers' lives better in general either -- it just seems to waste committee cycles on bikeshedding. Can I suggest we just stop `GHC2024` here and now. Might have been a nice idea. Turns out it's just an extra layer of complication.

georgefst:
> `GHC20XX` is not making my Haskell life better; it introduces extensions I actively hate; I don't use it. It doesn't seem to make Haskell programmers' lives better in general either -- it just seems to waste committee cycles on bikeshedding. Can I suggest we just stop `GHC2024` here and now. Might have been a nice idea. Turns out it's just an extra layer of complication.

Well it's made my life better! I write a lot of fairly small scripts and it's crazy to keep putting the same twenty or so pragmas at the top of each one just to enable Modern Haskell. For one thing, it's an embarrassment when introducing the language to people: "look how concise and readable this code is! oh ignore all those lines at the top, they don't matter".

I'm really glad to see this happening again, especially as I had been hoping for a 2022 or 2023 edition, having found the 2021 set pretty conservative for my tastes (I'd even like to see `NoFieldSelectors` and `OverloadedRecordDot` being discussed, but maybe the world isn't ready for that yet).

georgefst:
On the other hand, I am concerned by the moderate level of support for `GADTs`, in the Discourse poll and elsewhere, given the ramifications of `MonoLocalBinds`. I agree with @tomjaguarpaw [here](https://discourse.haskell.org/t/how-to-learn-language-extensions/7252/67) that it can make for a pretty horrible user experience. The inability to trivially factor out repeated expressions seems contrary to some of the core ideas of Haskell.

In this case I am particularly concerned for beginners, but it's something that really annoys me in my own code too. I find that it comes up often in practice.

simonpj:
> On the other hand, I am concerned by the moderate level of support for GADTs, in the Discourse poll and elsewhere, given the ramifications of MonoLocalBinds. I agree with @tomjaguarpaw [here](https://discourse.haskell.org/t/how-to-learn-language-extensions/7252/67) that it can make for a pretty horrible user experience. The inability to trivially factor out repeated expressions seems contrary to some of the core ideas of Haskell.

I hear this, but I'd love to see concrete evidence.  After all:
* You can still factor out repeated expressions
* But if you want to use the *same* expression with two different types, you need a type signature
* If it's not a function defn (and repeated expressions aren't) then it may well fall under the monomorphism restriction anyway

So yes, it's easy to show examples that break with `MonoLocalBinds`.  But are they common?   War stories please!

I note that we don't *have* to insist that `GADTs` implies `MonoLocalBinds`.  It's the kosher thing to do; it makes it possible to write typing rules that specify which programs typecheck and which do not.  But there nothing *unsound* about `NoMonoLocalBinds` and it usually works fine. It's just less principled.

proposals/0000-ghc2024.rst
```diff
@@ -0,0 +1,385 @@
...
+*Backward compatibility*: Assuming that ``MonoLocalBinds`` is enabled,
+enabling ``TypeFamilies`` doesn't affect existing ``GHC2021`` code
+further.
+
+Why add BlockArguments?
+-----------------------
+
+`BlockArguments`_ denoises common idioms, e.g. `when` for `forM` in do blocks, by not requiring
+parentheses or `$`. The resulting non-verbose syntax is arguably very Haskelly, and some
+people working with languages that have this already (Lean, PureScript) report beeing happy about it.
```
nomeata, line 353:
```suggestion
people working with languages that have this already (Agda, PureScript, Lean) report beeing happy about it.
```

phadej:
I'd suggest adding `ViewPatterns`. Can someone remind me why they aren't in GHC2021?

nomeata:
One reason is that a few people (including me) would rather see https://gitlab.haskell.org/ghc/ghc/-/wikis/view-patterns-alternative fleshed out, proposed, and implemented. Not that there are any indications in that direction… but I think that’s one of the reasons.

phadej:
@nomeata thanks. Then I won't suggest addition `ViewPatterns`. I want to point out however that `ViewPatterns` are necessary for writing (any interesting) `PatternSynonyms` (which aren't suggested either, but may be at some point).

simonpj:
Question: when GHC 9.X has GHC2024 as its default language edition, could that change (alone) break any packages?

Another way to say it: do all pacakges in Hackage specify a language edition?  See [this question](https://github.com/ghc-proposals/ghc-proposals/pull/625#issuecomment-1849853947)

phadej:
> Another way to say it: do all pacakges in Hackage specify a language edition? 

Yes

simonpj:
> Cabal made it possible to leave default language unspecified, and it's up to Hackage to require it or not for uploaded packages. In a way, it's not different than leaving base (or some other dependency) versions unspecified.

Right.  But Hackage *does* do such a check, right? Where is it done?  [Joachim couldn't find it](https://github.com/ghc-proposals/ghc-proposals/pull/625#issuecomment-1849938371).

Why do you say "Yes, so far, but they may not in the future."?  Of course, anything is possible in the future! But do you have any reason to suppose that Hackage might lift this requirement?

andreasabel:
> Who are the folk who maintain the Hackage infrastructure?

Those are @gbaz and @davean, if I am not mistaken.

tomjaguarpaw:
> >  Who are the folk who maintain the Hackage infrastructure?
>
> Those are @gbaz and @davean, if I am not mistaken.

This is an interesting question. @gbaz and @davean are [Haskell Infrastructure Admins](https://github.com/haskell-infra/haskell-admins#haskell-infrastructure-admins), and that might imply they are also Hackage administrators (contact details under "[Reporting Problems](https://hackage.haskell.org/)"). The group that determines Hackage policy might be another group entirely. I'm not sure.

gbaz:
Hackage does _not_ do a check for a language version existing now. In general we do _not_ want hackage to perform any checks cabal does not. We at most want hackage to turn some cabal _warnings_ into hackage _errors_ -- in fact, the cabal codebase specifically is designed to warn on things that may cause hackage upload errors, and the intention is that every package that `cabal check` passes warning-free should be uploadable to hackage. The only additional checks hackage does, outside of some ad-hoc things that have been done at times to check things ahead of new cabal releases, have to do with tarball size, well-formedness, etc -- things that are not about the _cabal file_ per-se.

Here's the issue where default-language was made optional by cabal: https://github.com/haskell/cabal/issues/6288

There's no real motivation for it, and I think it was an unfortunate decision.

If we want to force users to specify default-langauge, then the order of operations is we should make at _least_ a warning in cabal. I would suggest that interested parties just open a cabal ticket to make it entirely non-optional again. I see no good reason why it was made optional, and many good reasons to require it.

_however_, note that all cabal packages have an implicit bound on the ghc they work with, as set by their bound on base. And this in turn proxies to limiting to a range of which "default-language" options will be implicitly selected as the compiler default by the compiler versions that correspond.

So as long as changes to ghc's implicit default-language correspond to major version bumps in base, then the situation as it stands is workable.

Finally -- note that `cabal init` generates a file with a default-language automatically. So it is _possible_ that there are packages on hackage with no default language, but it is likely there are very few. Somebody would need to grep the full tarball to be certain.

phadej:
> I see no good reason why it was made optional,

Because it's not a hard requirement. People complained that writing minimal `.cabal` file is too much work. E.g. people were *very vocal* about `cabal install --lib` not working. They wanted a global installed library environment which just works, but...

The issue is that:
- `ghci` command just works,
- but `ghci` with libraries doesn't just work
- so people want a way to get a repl with libraries (on Hackage, or locally or ...) available **as easily as possible**.
- so they thought they want the "old way" of just installing stuff and dealing with cabal (dependency) hell (they don't want that).

I think that writing a bare minimum `.cabal` file is *good enough* way to achieve the repl with dependencies. Creating a package description doesn't mean you need to upload it Hackage. It can be *just* a local build specification. Creating a package definition should be easy. Uploading to Hackage may require a bit more work.

Or to put it differently: `.cabal` files are not *only* meant to be descriptions of packages to be uploaded to Hackage.

**And that means that `.cabal` spec will always have a design dilemma of being easy for throw away stuff, but robust for persistent stuff (e.g. for distribution, like via Hackage).**

(recall auto populating `exposed-modules`!)

IMHO a compromise is to allow `.cabal` file to not require much, and make `cabal check` be the strict(er) check. I admit that it was an oversight to not make `cabal check` warn that not specifying `default-language` would result in Hackage rejecting the package.

---

Your experience may vary, but for myself `cabal init` is awkward for throw away experiments: it asks too much and wrong stuff, and non-interactive way doesn't ask stuff and adds stuff to `.cabal` file which I don't want (It adds comments which are just annoying, and defaults to Haskell2010, where GHC2010 is better default for quick experiments). Throwing manually together a minimal `.cabal` file like

```cabal
cabal-version: 2.0
name: foo
version: 0

library
  build-depends: base, ...
  exposed-modules: Foo
```

is just better user experience for myself. TBH, I have a saved code-snippet in my text editor which has a bit more stuff, but I'm not always using *my* text editor, so occasionally I write `.cabal` files manually. Also `cabal init` errors on packages named `foo`, because package of such name exists: https://hackage.haskell.org/package/foo. That annoys me *a lot*, as `foo` is my go to throw away package name.

gbaz:
I don't agree with all the above reasoning, but I think its not worth hashing out further, especially in this ticket. We have a pretty clear plan of action that we can all agree on -- a pr to the cabal repo that creates a warning for missing language of at least "suspicious" severity -- which will lead to it being classified by `isHackageDistError` as something hackage would reject, and will lead to hackage, when it moves to using the new version of the cabal library, indeed rejecting the package.

nomeata:
> and will lead to hackage, when it moves to using the new version of the cabal library, indeed rejecting the package.

So the authoritative place for the checks that hackage performs is in the Cabal library after all? Then my [answer on the other thread](https://github.com/ghc-proposals/ghc-proposals/pull/625#issuecomment-1849880458) wasn't missing the mark after all, phew.

phadej:
> So the authoritative place

I'd argue it's not *authoritative*, and shouldn't be.  `hackage-server` can do additional checks on packages, and add them *quickly*, independently of `Cabal` (and it does, e.g. recent `PackageInfo` stuff). `hackage-server` can also ignore some warning checks issued by `cabal check`.

The analogy is that GHC has various warnings and has various severities for warnings (`-Wdefault`, `-Wall`, ...). And it's arguable that `-Wdefault` errors may cause distribution issues. But it is users (e.g. in their CI setup) who make some `-Werror` choices. And maybe run HLint which adds even more warnings. And in some other world, Hackage could actually build packages on upload and maybe reject some if there are some - non fatal - issues reported by GHC + other inspection tools.

But we leave in the real world unfortunately. E.g. Stackage maintainers rejected to add `-Werror=missing-methods` to their setup, so Stackage could become that warning-free, "because they track upstream". And now we are in limbo, with `-Werror=missing-methods` like behavior is many years away. Recall, the motivation for `-Werror=missing-methods` is to make change assessments easier, and if only Stackage is used as the assessment package set - we wouldn't need having `-Werror=missing-methods` in GHC.

Now as I bring that up, I notice that I have been myself inconsistent. In `Cabal` I was ok to be lenient, with GHC I *now* want to be more strict. I think I'd be fine if it's Stackage that is cranked up strict (does actual code checks, in addition to just building code and running some tests), Hackage less so (doesn't build stuff nowadays, cannot check "everything"), and then Cabal and GHC even less so (E.g. even allowing symbols with just type signatures, but no actual binding!)

For the record: https://github.com/haskell/cabal/issues/6288 no-one objected making `default-language` optional back then (e.g. not even Herbert who is usually quite strict about things). I can only guess and try to remember that people were annoyed by a warning that you can omit default-language, and nothing actually broke (and there weren't GHC2021 on the horizon in 2019).

adamgundry:
I've opened https://github.com/haskell/cabal/issues/9620 to track the Cabal-level warning discussed above.

The fact that changing GHC's default language to `GHC2024` will break existing programs has also come up in discussion on the implementation MR (https://gitlab.haskell.org/ghc/ghc/-/merge_requests/11882#note_543068). I wonder if we should have GHC emit a warning when compiling a module without an explicit language edition, to alert users that it may break in future releases? (After all, not all existing programs are Cabal packages.) Of course it's potentially a bit late to introduce the warning in 9.10 if we are also changing the default in 9.10...

mpickering:
Given the discussion on this ticket, I am quite surprised that a GHC MR has been proposed already which bumps the default extension to GHC2024.

* This causes breakage for standalone scripts (as can be seen widely in GHC's testsuite).
* This also potentially causes breakage for **packages** which don't specify a default language.

It seems that these issues with the ecosystem need to be patched up first before such changes can be introduced by default.

It seems that for 9.10 it would be ok to introduce GHC2024 as an extension set but I am currently strongly opposed to making it the default.

nomeata:
It was always the explicit plan that GHC when run as-is gives you the latest edition, for a good experience in ghci and throw-away scripts, and that code that should last more than one GHC release should be explicit in the edition. This is how we did it for GHC2021 (which enabled many more extensions than GHC2024). Of course it’s something we could have communicated more loudly, but that’s always the case.

It’s a bit of unfortunate timing that `cabal` dropped the requirement to specify a language editions around the time we set out these rules and GHC2021 was introduced (under the assumption that Cabal requires a language to be specified), but do we have evidence that people don’t specify an edition? (I doubt it)

The use case of just firing up `ghc` or `ghci` remains a useful one, and I would not want that to warn me about lack of `-XGHC2024`. At least not in `-Wdefault`.

The GHC testsuite is hardly representative here; it tests weird corners by construction, and it is written to work with just the current GHC, with the least effort.

mpickering:
The environment is now very different than when `GHC2021` was proposed and merged, there is clearly a focus now around stability so it is disappointing that the steering committee is quite gung-ho about introducing this breakage without waiting for suitable mitigations to exist in the ecosystem.

It seems that at least cabal developers, and also ghc developers (myself) are not aware of this expectation that everyone must provide a language edition for any stability guarantee. It seems to make other stability efforts a bit pointless if every few releases many standalone scripts will break because a new non-conservative language edition will be released and quickly made the default. 

Is this change going to break people compiling xmonad configs? Will this change render instructions about using GHC directly in university courses redundant? The proposal is quiet about any kind of breakage impact and whether the ecosystem is ready to absorb such breakage, even if "it was always the plan", that doesn't mean everything is lined up and ready for the change.

nomeata:
Of course you have a point, and it’s not completely unexpected. The worry that we’ll have this “OHG GHC2024 suddenly happens what do we do why is not everything behaving the way it did since GHC2021” effect was one of the main reasons I tried to push for *more frequent* GHC20xx releases, so that it wouldn’t have come as a surprise anymore, no more than a GHC release itself. But I failed to convince back then.

Anyways, I will not stand in the way of a decision to make GHC2021 not the default until the world is ready, should someone make a push for that.

alt-romes:
I share Matthew's opinion. MonoLocalBinds, more precisely, is quite dangerous by default for non-cabal packages, because it is mostly in scripts that eg top-level definitions go by without signatures whatsoever.

Simon raised this point earlier in the thread, but it got somewhat resolved/cleared as "to upload to hackage, a default-language is mandatory". However, that fails to account for scripts and lecture-notes/material and such.

Even though GHC2021 introduced a lot of extensions, did it introduce one with as much potential for breakage as MonoLocalBinds? I think it is a fair extension to include in GHC2024, just not by default when running the compiler!

When part of a package, this is much less of a problem, first, because `default-language` is required for hackage and often used generally, and, second, because packages are more "robust" to MonoLocalBinds in the sense that package's don't often have top-level bindings without type annotations, even if a few local lets don't.

It's unfortunate that we seemingly overlooked this before the proposal was accepted.

And I agree with @nomeata that we shouldn't introduce a warning by default to require a language extension -- that is indeed to burdensome for a quick ghci session.

MangoIV:
> The environment is now very different than when `GHC2021` was proposed and merged, there is clearly a focus now around stability so it is disappointing that the steering committee is quite gung-ho about introducing this breakage without waiting for suitable mitigations to exist in the ecosystem.
> 
> 
> 
> It seems that at least cabal developers, and also ghc developers (myself) are not aware of this expectation that everyone must provide a language edition for any stability guarantee. It seems to make other stability efforts a bit pointless if every few releases many standalone scripts will break because a new non-conservative language edition will be released and quickly made the default. 
> 
> 
> 
> Is this change going to break people compiling xmonad configs? Will this change render instructions about using GHC directly in university courses redundant? The proposal is quiet about any kind of breakage impact and whether the ecosystem is ready to absorb such breakage, even if "it was always the plan", that doesn't mean everything is lined up and ready for the change.

simonpj:
> It seems that at least cabal developers, and also ghc developers (myself) are not aware of this expectation that everyone must provide a language edition for any stability guarantee

It's an absolutely key part of [GHC Proposal #625](https://github.com/ghc-proposals/ghc-proposals/pull/625), for which I have widely sought feedback.  See GR1.

A standalone script specifies what compiler to use (Python, Haskell, bash etc).   Think of `ghc -XGHC2021` as a different compiler to `ghc -XGHC2010`.  Every script should specify a language edition, to specify the language in which it is written.   That's not true today, but I see no other way to ensure their stability.

However GR3 says that, whenever possible we should release a compiler that *warns* of an upcoming change, rather than simply introducing it.  Sometimes it it not feasible to do this.  For example adding `foldl'` to `Prelude` (a recent change) broke a lot of packages.

In the case of `MonoLocalBinds` I think that, with some non-trivial work, one could warn about bindings that will be rejected with `MonoLocalBinds`.  Then we could release a compiler that warns, and make `GHC2024` the default in the release after.  Would that help?

PR #616, opened 2023-10-03
Ghc CPP

alanz:
This is a proposal to introduce a new **GhcCPP** pragma to perform a subset of **CPP** processing as an alternative, as part of the normal GHC compilation process.

This subset will be just enough to control conditional compilation,and will only be active in lines introduced with a ``#`` and preprocessor keyword.

It will be tooling friendly, in that parts of the code disabled by the processor will be saved as comments, if comments are being processed.

[Rendered](https://github.com/alanz/ghc-proposals/blob/ghc-cpp/proposals/0000-ghc-cpp.rst)

### DIFF
--- a/proposals/0000-ghc-cpp.rst
+++ b/proposals/0000-ghc-cpp.rst
@@ -1,0 +1,221 @@
+Ghc CPP
+=======
+
+.. author:: Alan Zimmerman
+.. date-accepted:: 
+.. ticket-url::
+.. implemented:: 
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/0>`_.
+            **After creating the pull request, edit this file again, update the
+            number in the link, and delete this bold sentence.**
+.. sectnum::
+.. contents::
+
+Haskell code uses the CPP pragma extensively. I'm proposing to
+introduce a new GhcCPP pragma to perform a subset of CPP processing as
+an alternative, as part of the normal GHC compilation process.
+
+This subset will be just enough to control conditional compilation,
+and will only be active in lines introduced with a ``#`` and
+preprocessor keyword.
+
+It will be tooling friendly, in that parts of the code disabled by the
+processor will be saved as comments, if comments are being processed.
+
+
+Motivation
+----------
+
+Haskell code uses the CPP pragma to manage version changes, when the
+compiler or a library changes.
+
+CPP gets the job done, but it is not haskell-friendly, and so has a
+side-effect of mangling otherwise normal haskell code, which is not
+intended to be processed for the version-selection process.
+
+It is also not tooling-friendly. It runs as a separate pass over the
+source code, emitting updated source for compilation which can
+actually change pretty much anything, if a person decides to use the
+full power of CPP to redefine existing Haskell, or introduce macros to
+generate boilerplate code.
+
+The primary motivation for this proposal is to make it possible to
+safely perform source-code transformations on code containing
+preprocessor directives.
+
+Currently in ``ghc-exactprint`` when a file uses CPP, the original
+file and preprocessor output are diffed at the token level and the
+"missing" parts inserted as comments. This works in some cases, but
+for many it fails, for reasons such as
+
+- A comparison to a numeric constant
+
+  ::
+   #if __GLASGOW_HASKELL__ < 707
+
+  Shows up as a stray ``906`` if compiled with GHC 9.6.x
+
+- Preprocessor directives inside haskell comments are also processed.
+  This makes the diff fail too.
+
+- Trailing backslashes are eliminated, and the lines joined up. This
+  breaks some ASCII art comments, as well as multi-line strings.
+
+By implementing a subset of CPP processing inside GHC, we can avoid
+these drawbacks, and provide conditional compilation facilities that
+cover most of the "normal" current needs.
+
+The processing will be explicitly limited to lines introduced by a
+standard CPP directive, such as ``#define FOO``, and operate at the
+GHC Lexer level. This means that it will not touch directives
+appearing inside comments, nor will CPP macros be applies anywhere but
+in other CPP directives.
+
+
+Proposed Change Specification
+-----------------------------
+
+When enabled, the lexer has the additional rules
+
+::
+   cpp -> '#' cppkeyword [ anysymbol {any} ] newline
+         | '#' cppkeyword [ anysymbol {any} ] '\' newline cppcont
+   cppcont -> [ anysymbol {any} ] '\' newline cppcont
+         |  [ anysymbol {any} ] newline
+   cppkeyword -> 'define' | 'include' | 'undef' | 'error' | 'ifdef'
+                 | 'ifndef' | 'if' | 'elif' | 'else' | 'endif'
+
+   ccomment = '/*'  [ anysymbol {any} ] '*/'
+   whitestuff -> whitechar | comment | ncomment | ccomment
+
+Note: as per
+https://gcc.gnu.org/onlinedocs/cpp/Initial-processing.html the
+CPP-style comments do not nest.
+
+The only tokens emitted are ``cpp`` and ``cppcont``. The additional
+comment type is stored with the others if they are being kept.
+
+These will be passed to the actual preprocessor for processing. This will
+
+- Accumulate continued ``cpp`` and ``cppcont`` tokens until a full CPP
+  directive line is constructed.
+- Apply the standard CPP processing to this line. (Based on
+  https://gcc.gnu.org/onlinedocs/cpp/Overview.html)
+- Any preprocessor state will be inserted into an (opaque) field in
+  the parser state, made available to it by get and set operators.
+
+The result of this processing is to put the preprocessor into one of two states
+- Normal
+- Ignoring
+
+In **Normal** mode it passes all non-cpp tokens on to the parser as
+normal.
+
+In **Ignoring** mode it converts all non-cpp tokens into comments,
+which are stored or not as per normal comment processing.
+
+In either case, the ``cpp`` and ``cppcont`` tokens are converted to
+comments, as above, and not passed on to the parser.
+
+Example
+
+.. code:: haskell
+   {-# LANGUAGE GhcCPP #-}
+   #define FOO
+   #ifndef FOO
+   x = 1
+   #else
+   x = 5
+   #endif
+
+Results in the following token stream (showing comments as they are lexed)
+
+::
+   Comment ({-# LANGUAGE GhcCPP #-})
+   Comment (ITcpp "#define FOO")
+   Comment (ITcpp "#ifndef FOO")
+   Comment (ITvarid "x")
+   Comment (ITequal)
+   Comment (ITinteger 1)
+   Comment (ITcpp "#else")
+   ITvarid "x"
+   ITequal
+   ITinteger 5
+   Comment (ITcpp "#endif")
+   ITeof
+
+
+
+Proposed Library Change Specification
+-------------------------------------
+
+N/A
+
+Examples
+--------
+
+Effect and Interactions
+-----------------------
+
+There may be potential interactions between this extension and CPP.
+It would probably be wise to emit a warning if both are enabled at the same time.
+
+
+Costs and Drawbacks
+-------------------
+The main cost is some additional complexity in the lexer, to emit the
+preprocessor tokens when the pragma is enabled. This should be a fixed
+change though, and the actual preprocessor processing happens in
+another layer. So it should not adversely affect ongoing language
+evolution in the GHC lexer and parser.
+
+
+Backward Compatibility
+----------------------
+
+This proposal introduces a new pragma, ``GhcCPP``, as an alternative
+to the existing ``CPP`` pragma.
+
+It explicitly does not set out to be a drop-in replacement for
+``CPP``, as it has a reduced and changed functionality.
+
+As it is a pragma, it is up to potential users of the feature to turn
+it on in a case-by-case basis as part of their development process.
+They can always fall back to ``CPP`` if the effects are not what they
+expect.
+
+Will your proposed change cause any existing programs to change behaviour or
+stop working? Assess the expected impact on existing code on the following scale:
+
+0. No breakage
+
+
+
+Alternatives
+------------
+List alternative designs to your proposed change. Both existing
+workarounds, or alternative choices for the changes. Explain
+the reasons for choosing the proposed change over these alternative:
+*e.g.* they can be cheaper but insufficient, or better but too
+expensive. Or something else.
+
+The PR discussion often raises other potential designs, and they should be
+added to this section. Similarly, if the proposed change
+specification changes significantly, the old one should be listed in
+this section.
+
+Unresolved Questions
+--------------------
+TBD
+
+Implementation Plan
+-------------------
+
+If accepted, the author (@alanz) will implement it.
+There is already some preliminary work on
+https://gitlab.haskell.org/ghc/ghc/-/tree/wip/az/ghc-cpp
+
+Endorsements
+-------------
+Yes please
--- a/proposals/0000-ghc-cpp.rst
+++ b/proposals/0000-ghc-cpp.rst
@@ -10,8 +10,8 @@ Ghc CPP
 .. sectnum::
 .. contents::
 
-Haskell code uses the CPP pragma extensively. I'm proposing to
-introduce a new GhcCPP pragma to perform a subset of CPP processing as
+Haskell code uses the ``CPP`` extension extensively. I'm proposing to
+introduce a new ``GhcCPP`` extension to perform a subset of CPP processing as
 an alternative, as part of the normal GHC compilation process.
 
 This subset will be just enough to control conditional compilation,
@@ -25,10 +25,10 @@ processor will be saved as comments, if comments are being processed.
 Motivation
 ----------
 
-Haskell code uses the CPP pragma to manage version changes, when the
+Haskell code uses the ``CPP`` extension to manage version changes, when the
 compiler or a library changes.
 
-CPP gets the job done, but it is not haskell-friendly, and so has a
+CPP gets the job done, but it is not Haskell-friendly, and so has a
 side-effect of mangling otherwise normal haskell code, which is not
 intended to be processed for the version-selection process.
 
@@ -184,8 +184,8 @@ evolution in the GHC lexer and parser.
 Backward Compatibility
 ----------------------
 
-This proposal introduces a new pragma, ``GhcCPP``, as an alternative
-to the existing ``CPP`` pragma.
+This proposal introduces a new extension, ``GhcCPP``, as an alternative
+to the existing ``CPP`` extension.
 
 It explicitly does not set out to be a drop-in replacement for
 ``CPP``, as it has a reduced and changed functionality.
--- a/proposals/0000-ghc-cpp.rst
+++ b/proposals/0000-ghc-cpp.rst
@@ -49,9 +49,10 @@ file and preprocessor output are diffed at the token level and the
 "missing" parts inserted as comments. This works in some cases, but
 for many it fails, for reasons such as
 
-- A comparison to a numeric constant
+- A comparison to a numeric constant:
 
-  ::
+.. code:: haskell
+
    #if __GLASGOW_HASKELL__ < 707
 
   Shows up as a stray ``906`` if compiled with GHC 9.6.x
@@ -76,9 +77,10 @@ in other CPP directives.
 Proposed Change Specification
 -----------------------------
 
-When enabled, the lexer has the additional rules
+When enabled, the lexer has the additional rules:
 
-::
+.. code:: haskell
+
    cpp -> '#' cppkeyword [ anysymbol {any} ] newline
          | '#' cppkeyword [ anysymbol {any} ] '\' newline cppcont
    cppcont -> [ anysymbol {any} ] '\' newline cppcont
@@ -118,9 +120,10 @@ which are stored or not as per normal comment processing.
 In either case, the ``cpp`` and ``cppcont`` tokens are converted to
 comments, as above, and not passed on to the parser.
 
-Example
+Example:
 
 .. code:: haskell
+
    {-# LANGUAGE GhcCPP #-}
    #define FOO
    #ifndef FOO
@@ -129,9 +132,10 @@ Example
    x = 5
    #endif
 
-Results in the following token stream (showing comments as they are lexed)
+Results in the following token stream (showing comments as they are lexed):
 
-::
+.. code:: haskell
+
    Comment ({-# LANGUAGE GhcCPP #-})
    Comment (ITcpp "#define FOO")
    Comment (ITcpp "#ifndef FOO")
--- a/proposals/0000-ghc-cpp.rst
+++ b/proposals/0000-ghc-cpp.rst
@@ -28,15 +28,12 @@ Motivation
 Haskell code uses the ``CPP`` extension to manage version changes, when the
 compiler or a library changes.
 
-CPP gets the job done, but it is not Haskell-friendly, and so has a
-side-effect of mangling otherwise normal haskell code, which is not
-intended to be processed for the version-selection process.
-
-It is also not tooling-friendly. It runs as a separate pass over the
-source code, emitting updated source for compilation which can
-actually change pretty much anything, if a person decides to use the
-full power of CPP to redefine existing Haskell, or introduce macros to
-generate boilerplate code.
+CPP gets the job done, but it is not Haskell-native, and so has a
+side-effect of not being tooling-friendly. It runs as a separate pass
+over the source code, emitting updated source for compilation which
+can actually change pretty much anything, if a person decides to use
+the full power of CPP to redefine existing Haskell, or introduce
+macros to generate boilerplate code.
 
 The primary motivation for this proposal is to make it possible to
 safely perform source-code transformations on code containing
@@ -54,6 +51,10 @@ for many it fails, for reasons such as
    #if __GLASGOW_HASKELL__ < 707
 
 Shows up as a stray ``906`` if compiled with GHC 9.6.x
+
+.. code:: haskell
+
+   #if __GLASGOW_HASKELL__906 < 707
 
 - Preprocessor directives inside haskell comments are also processed.
   This makes the diff fail too.
@@ -91,7 +92,8 @@ When enabled, the lexer has the additional rules:
 
 Note: as per
 https://gcc.gnu.org/onlinedocs/cpp/Initial-processing.html the
-CPP-style comments do not nest.
+CPP-style comments do not nest. They are needed because they are valid
+CPP syntax, and are emitted in files such as ``cabal_macros.h``.
 
 The only tokens emitted are ``cpp`` and ``cppcont``. The additional
 comment type is stored with the others if they are being kept.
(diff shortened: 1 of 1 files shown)


nomeata:
Interesting idea. I am wondering: why exactly these features of CPP? Why not expanding macros in the source as well? Is there a risk that people will ask for more and more features to be added to GhcCPP, until it's almost CPP again? Or is there a good reason to assume that the proposed feature set is a particular attractive sweet spot, also from the user pov? (I don't doubt that it is, but it could be made clearer.)

alanz:
> Why not expanding macros in the source as well?

Because then the source as seen in the editor and the source seen by the compiler are different, so managing tools around this gets tricky.

Plus I have a gut feel that we need a way for conditional compilation, but that CPP is too powerful, and so gets abused.

This approach does not rule out CPP if it is actually needed, just tries to provide a weaker variant, just strong enough for conditional compilation.

phadej:
> Why not expanding macros in the source as well?

Cause then you'll need lex (and actually preserve in parse tree for refactoring tools) them. 

```haskell
#if __GLASGOW_HASKELL__ >=800
#define FOO(x,y) (foo x y noArg)
#else
#define FOO(x,y) (foo x y)
#endif

...
main :: IO ()
main = print FOO(x,y)
```

This is not unseen code, but while you can lex `FOO(x,y)` successfully, and even expand it, the HsExpr will look terrible. Or actually `HsPar`, `HsType` etc. as every construct would need to preserve the original `FOO(x,y)` structure, so if some e.g. renaming refactoring `x -> z` can produce

```haskell
main :: IO ()
main = print FOO(z,y)
```

and not the expanded version.

---

And this example is one which can be rewritten as

```haskell
#if __GLASGOW_HASKELL__ >=800
myFoo x y = foo x y noArg
#else
myFoo x y = foo x y
#endif

...
main :: IO ()
main = print (myFoo x y)
```

which would fall into GhcCPP supported realm. With `type`, `pattern` (pattern synonyms) etc there is a lot one can shim on Haskell level instead of text-expansion lexer level.

---

Something you cannot, like Prim instance definitions in `primitive` https://github.com/haskell/primitive/blob/0b64fae48295f566fc6fec8e86f268924eb27683/Data/Primitive/Types.hs#L292 but I'd argue that's a situation like with `alex`, `happy` or `hsc2hs` files, for which tools just cannot do much, and the best you can do is to put as little code in these files as possible.

phadej:
> There may be potential interactions between this extension and CPP. It would probably be wise to emit a warning if both are enabled at the same time.

For library code which won't be able to drop the old GHC support immediately, I imaginepeople wanting to use something like

```
{-# LANGUAGE CPP #-}
#if __GLASGOW_HASKELL__ >= VERSION_WITH_GHCCPP
{-# LANGUAGE GhcCPP #-}
#endif

```

and rip the GhcCPP benefits while developing their libs using recent GHC, but not forcing users to use it.

I'd welcome a better story than a warning (which these users would turn off).

alanz:
> I imagine people wanting to use something like ..

The key is to not have them run, and by enabling CPP for the choice it already runs.

So I imagine the initial scan for pragmas would have to be aware of this idiom, and, do the selection to enable only GhcCPP

Or if both are enabled silently drop CPP.

phadej:
> Or if both are enabled silently drop CPP.

That is an option.

I'd also imagine that (in the future) people would have

```
default-extension: GhcCPP
```

in their cabal files (or GhcCPP will even be a part of GHC20xx), but in some file they would actually need `CPP`.

But then they could have

```haskell
{-# LANGUAGE NoGhcCPP #-}
{-# LANGUAGE CPP #-}
```

and not have `CPP` silently dropped.

---

I'd expect `GhcCPP` be almost free i.e. could be enabled for every file i.e. not run `cpp` at all if there is no `#pragmas`. I think the proposal does say so, just not explicitly. (And GHC could record in the interface files if `cpp` were used, so e.g. if `cabal_macros.h` changes, there won't be recompilation - as `cabal_macros.h` doesn't produce any code - only CPP definitions).

Ericson2314:
I would like us to end up with something nicer like Rust's `#[cfg(..)]` eventually, but I certainly see the merits of this as an intermediate step.

alanz:
> I would like us to end up with something nicer

Longer term, once we have something like this, that belongs to GHC, we can plan on how to evolve it, without taking it as a given

AntC2:
[`cpphs`](https://hackage.haskell.org/package/cpphs-1.17/src/docs/index.html) is still available, and still works. I suggest you compare how it works to your proposal.

`cpphs` also aims to avoid "a side-effect of mangling otherwise normal haskell code, which is not intended to be processed for the version-selection process." In particular, it's aware that in Haskell line breaks and indentation are significant; also that `'` might be part of an identifier, not necessarily a `Char` literal.

But how to drive `cpphs` isn't abundantly obvious/the docos are unclear. [I wrote up some notes](https://stackoverflow.com/a/73768057/2840542).

(And as per comments on that StackOverflow q, are you sure it wouldn't be more Haskelly to extend Template Haskell?)

Thinking in a whole-application context, where there are modules in all sorts of languages as well as Haskell: `CPP`, for all its limitations, does work consistently on any source. Will installations want to break their heads on Haskell CPP being different to `C` CPP or perl/python/js/other glue languages CPP?

alanz:
> [cpphs](https://hackage.haskell.org/package/cpphs-1.17/src/docs/index.html) is still available, and still works

Thanks, I will do so.

However, the key feature for me of my proposed approach is to get an AST with Exact Print annotations matching the AST GHC sees, and the rest captured clearly in comments.

So any approach involving a separate preprocessor, and then reconstructing the missing parts as comments (as we do now, mostly unsuccessfully), will not be achieving my primary goal.

And the only assumption I want to have to make is that the lexer can process the code, which makes TH troublesome.

For managing versions of GHC, there are necessarily changes.  This means that to rely on Template Haskell we need to actually parse the code sufficiently to be able to run the TH.

AntC2:
> the key feature for me of my proposed approach is to get an AST with Exact Print annotations matching the AST GHC sees, and the rest captured clearly in comments.

Thanks. Ah, yes that's valuable. We don't get traceability back to original source by using an alternative preprocessor such as `cpphs`. (But see `cpphs`'s `--linepragma` option, which at least tries.)

> perform a subset of CPP processing as an alternative, as part of the normal GHC compilation process.
>
> This subset will be just enough to control conditional compilation, ...

I don't have a strong feel for how much folk are using macros as well as conditional compilation directives. But if they are, the proposal as stands seems to block using macros(?) Or can you first run `cpp` only to expand macros, before passing the result to `GhcCPP`? (Looks like not with `cpphs` off the shelf, but perhaps it wouldn't be too hard to add a `--onlymacro` option? `cpphs` docos say "Numbering of lines in the output is preserved so that any later processor can give meaningful error messages." -- so they're on the same page -- metaphorically, that is. ;-)

(Re Template Haskell, Yeah I was just making sure you'd considered and rejected it.)

bgamari:
> I don't have a strong feel for how much folk are using macros as well as conditional compilation directives. But if they are, the proposal as stands seems to block using macros(?) 

My understanding is that users needing to use macros will continue to use `CPP`. However, my sense is that most (nearly all?) users use `CPP` only for simple conditional compilation, for which `GhcCPP` will suffice.

proposals/0000-ghc-cpp.rst
```diff
@@ -0,0 +1,231 @@
...
+.. date-accepted:: 
+.. ticket-url::
+.. implemented:: 
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/616>`_.
+.. sectnum::
+.. contents::
+
+Haskell code uses the CPP pragma extensively. I'm proposing to
+introduce a new GhcCPP pragma to perform a subset of CPP processing as
```
konsumlamm, line 14:
```suggestion
Haskell code uses the ``CPP`` extension extensively. I'm proposing to
introduce a new ``GhcCPP`` extension to perform a subset of CPP processing as
```
{-# LANGUAGE #-}` is a pragma, `CPP` is an extension.

proposals/0000-ghc-cpp.rst
```diff
@@ -0,0 +1,231 @@
...
+preprocessor keyword.
+
+It will be tooling friendly, in that parts of the code disabled by the
+processor will be saved as comments, if comments are being processed.
+
+
+Motivation
+----------
+
+Haskell code uses the CPP pragma to manage version changes, when the
```
konsumlamm, line 28:
```suggestion
Haskell code uses the ``CPP`` extension to manage version changes, when the
```

proposals/0000-ghc-cpp.rst
```diff
@@ -0,0 +1,231 @@
...
+compiler or a library changes.
+
+CPP gets the job done, but it is not haskell-friendly, and so has a
```
konsumlamm, line 31:
```suggestion
CPP gets the job done, but it is not Haskell-friendly, and so has a
```

proposals/0000-ghc-cpp.rst
```diff
@@ -0,0 +1,231 @@
...
+side-effect of mangling otherwise normal haskell code, which is not
```
konsumlamm, line 32: What do you mean by mangling?

proposals/0000-ghc-cpp.rst
```diff
@@ -0,0 +1,231 @@
...
+"missing" parts inserted as comments. This works in some cases, but
+for many it fails, for reasons such as
+
+- A comparison to a numeric constant:
+
+.. code:: haskell
+
+   #if __GLASGOW_HASKELL__ < 707
+
+Shows up as a stray ``906`` if compiled with GHC 9.6.x
```
konsumlamm, line 56:
```suggestion
Shows up as a stray ``906`` if compiled with GHC 9.6.x
```
The whole ``#if __GLASGOW_HASKELL__ < 707`` shows up as ``906`` or just the ``__GLASGOW_HASKELL__``?

proposals/0000-ghc-cpp.rst
```diff
@@ -0,0 +1,231 @@
...
+.. code:: haskell
+
+   cpp -> '#' cppkeyword [ anysymbol {any} ] newline
+         | '#' cppkeyword [ anysymbol {any} ] '\' newline cppcont
+   cppcont -> [ anysymbol {any} ] '\' newline cppcont
+         |  [ anysymbol {any} ] newline
+   cppkeyword -> 'define' | 'include' | 'undef' | 'error' | 'ifdef'
+                 | 'ifndef' | 'if' | 'elif' | 'else' | 'endif'
+
+   ccomment = '/*'  [ anysymbol {any} ] '*/'
```
konsumlamm, line 89: What's the point of supporting another type of comment?

proposals/0000-ghc-cpp.rst
```diff
@@ -0,0 +1,231 @@
...
+change though, and the actual preprocessor processing happens in
+another layer. So it should not adversely affect ongoing language
+evolution in the GHC lexer and parser.
+
+
+Backward Compatibility
+----------------------
+
+This proposal introduces a new pragma, ``GhcCPP``, as an alternative
+to the existing ``CPP`` pragma.
```
konsumlamm, line 188:
```suggestion
This proposal introduces a new extension, ``GhcCPP``, as an alternative
to the existing ``CPP`` extension.
```

konsumlamm (commented):
Some name bikeshedding: GHC and CPP are both acronyms, yet the extension name is `GhcCPP`. It might make more sense to choose something like `GHC_CPP` or something different altogether.

proposals/0000-ghc-cpp.rst
phadej: `#include` files (unfortunately for us) have C-style comments. E.g. `cabal_macros.h`, `ghcautoconf.h`

adamgundry:
I like this idea in general. If I understand correctly, unchosen alternatives will still be lexed (so they will need to be lexically correct) but not parsed (because the token stream from the lexer represents them as comments). Is that right, and is that what we want? I can imagine other reasonable positions:
 * Ignored branches are not required to be lexically correct, which might be helpful if there are changes in the lexer that need to be worked around between versions (but hopefully this is rare in practice).
 * Ignored branches are still parsed, so the compiler will catch syntax errors (though the same issue applies regarding changes in the parser, and I can imagine this being difficult to implement if the conditional doesn't enclose an entire declaration).
 * Ignored branches are renamed and possibly even typechecked (this goes in the direction of variational programming, but given the complexity and the fact that this wouldn't help for working around library API changes, I guess it is unlikely to be worthwhile for the time being).

alanz:
> My understanding is that users needing to use macros will continue to use CPP

This proposal does not disable macros, it just limits their applicability to be within preprocessor directives.  They should have the same feature set as normal CPP ones. Or at least enough to serve the purpose of conditional compilation.

I also see the rollout process being gradual, with the initial version having just enough to cover "standard" usage, for some definition of standard.  I see that as at the very least dealing with the cabal macro definitions, and other simple conditionals.

alanz:
> Some name bikeshedding

I am expecting a discussion of the name, what I have used is provisional.

Ericson2314:
@adamgundry Yes I definitely would to push on the variational stuff long term, as that 2019 HIW presentation from @mpickering and I discussed, but I don't think that need be part of the initial proposal. This is merely defining the language, while those things are more and more sophisticate processing of the same language.

phadej:
One thing which this proposal won't support is using CPP to do conditional `doctest` (or more general documentation) assembling. We can do

```haskell
{-

#if __GLASGOW_HASKELL >= 900

On newer GHCs >= 9 we also get
>>> foo
"foo"

#endif

-}
```

I think it's essential to use `{- ... -}` comments, as CPP inserts empty lines (and `#line` / `{-# LINE -}` pragmas) which break `--` line comment blocks.

proposals/0000-ghc-cpp.rst
alanz: Using the `ghc-exactprint` diffing process, and printing,  this ends up something like

```
#if __GLASGOW_HASKELL__906 < 707
```

The point being it makes source-to-source tooling impossible

proposals/0000-ghc-cpp.rst
alanz: I will remove this

Ericson2314:
We should test pretty rigorously that this and a text-based implementation commute:

https://q.uiver.app/#q=WzAsNCxbMCwwLCJcXG1hdGhybXtPcmlnaW5hbH0iXSxbMCwyLCJcXG1hdGhybXtUb2tlbnN9Il0sWzMsMCwiXFxtYXRocm17UmVkdWNlZCBUZXh0fSJdLFszLDIsIlxcbWF0aHJte0FTVH0iXSxbMCwxLCJcXG1hdGhybXtMZXh9Il0sWzAsMiwiXFxtYXRocm17R0hDUFBfe3RleHR9fSIsMl0sWzIsMywiXFxtYXRocm17UGFyc2UgXFxjZG90IExleH0iLDJdLFsxLDMsIlxcbWF0aHJte1BhcnNlIFxcY2RvdCBHSENDUFBfe3Rva2Vuc319ICJdXQ==

alanz:
> One thing which this proposal won't support is using CPP to do conditional doctest ...

I think that would be out of scope of this proposal, but could be addressed in a follow-up

Ericson2314:
Kind of like @adamgundry's point, I think the proposal might go into too much detail on how it will work internally. E.g. perhaps it would be better to anticipate some the variational stuff and separate this from the other tokens. otherwise e.g. `#if` inside comments will not work.

michaelpj:
Does this have any effect on recompilation checking? I might hope that if we have a restricted subset of CPP then we can also get better recompilation checking, e.g. perhaps the GHC-CPP preprocessor can tell us which flags it actually looked at.

andreabedini:
Why sticking to CPP syntax? Could we use annotations instead? A simple tool could translate back to CPP for old compilers (but if the change is small enough it could be implemented as a patch release)

Ericson2314:
We should already have good recompilation checking by being able to remember the result of running C++ (or its hash) and only continue if it changed.

alanz:
> We should test pretty rigorously that this and a text-based implementation commute

One of the test tools I have in mind is to compare dumped ASTs for files using CPP and GhcCPP to make sure they are identical.  And then run them on hackage and see what proportion passes.

alanz:
> hy sticking to CPP syntax? Could we use annotations instead?

Because the motivation is to enable tooling, to process the files.  If it is as easy as changing a pragma at the top of a file, it will lead to a faster uptake.

But once we have the feature, and concept of built-in preprocessor, it should be possible to experiment with alternative versions too.

phadej:
> Why sticking to CPP syntax?

We still want be able to process say `cabal_macros.h` (though nowadays most of definitions there are provided by GHC itself), `ghcautoconf.h`, packages' own `autoconf` outputs (like `time`). TL;DR there are a lot external definitions used for conditional compilation.

---

Also if the new syntax were introduced, I wouldn't use it until the **oldest** GHC I support the new syntax. That will take probably at least three, for some packages up to five years. Because I'm happy with CPP, I wouldn't change just because I could.

phadej:
> Why sticking to CPP syntax?

What's wrong with CPP syntax? It doesn't conflict with anything in Haskell, it's not terrible. It also exists.

alanz:
> Why sticking to CPP syntax?

It also lets the proposal focus on the essence (for me) of the thing. To be able to do source to source transformations of a larger class of source files.

Otherwise it would become a discussion of some new language, which would never end.

Ericson2314:
@alanz So per my commutative diagram, anything that e.g. didn't support `#if` within `{- ... -}` would be wrong. I am thinking that / the two-pass version should be the defined correct one, and then your single pass version is fine to implement (and great for testing) but with the understanding that it is not quite correct. And then sometimes layer with enough yak shaving it comes correct.

This is a round-about way to say I don't want anyone to *rely* on the fact that `#if` within `{- ... -}` is ignored, e.g. expect it to show up in haddock.

It should be pretty easy to slim down `cpphs` (or just rewrite from scatch) to make the actually "production" implementation of `GhcCpp`. And then we can reconcile with your version for exact-print purposes over time.

alanz:
> So per my commutative diagram

I am working on the assumption that the initial versions of this will only commute for a subset of all Haskell code currently using CPP.  And for the first attempt, applying expansions inside comments is excluded.

And to be sure I understand, I take your diagram to say (top path)
- Run GHC Lexer and GhcCPP preprocessor to generate tokens.
- Dump these tokens to a text file, discarding the comment ones
- Parse this text file with regular GHC, dumping the AST

bottom path
- run GHC with GhcCpp enabled for the file, as intended with this proposal, and dump the AST
- These two ASTs should be identical

For my side, I am more interested initially on comparing the AST for a file using the GhcCpp extension vs the same one using CPP

Ericson2314:
@alanz So it seems to me that independently of Haskell, you've come up with a nice minimal variant of CPP (only operate on `#` lines). Maybe let's call that `CPP--` :grin:.

I think the two pass approach of "run CPP--, parse Haskell" is a much simpler mental model for understanding what is going on. And likewise the small differences between that and the one pass approach you have implemented today where `#if` is ignored within `{- -}` are not so intuitive.

So I am thinking we should take "run CPP--, parse Haskell" as the definition of what is supposed to happen, and we perhaps implement it that way for things that can observe the difference (e.g. Haddock, doctests, etc. where the contents of comments matters). And then over time we can make the one-pass way actually do the same thing as it and then switch over.

alanz:
> Make sure to fix multiline strings as well 

That is one of the motivating problems for this proposal

aryah47:
Great to finally see a concrete proposal for a slimmed-down cpp for haskell specific purposes, ie conditional compilation, thx for doing this! I'm a bit hesitant with having another kind of comment, though I guess its well motivated by backward compatibility. 

Still, should this say become part of a future standard, would we then have /* and */ potential operator syntax occupied by just yet another kind of comment? Not terrible, but somewhat unsatisfying.. Still, overall, excited to see this feature land!

alanz:
@phadej wrote

> For library code which won't be able to drop the old GHC support immediately, I imaginepeople wanting to use something like

```
{-# LANGUAGE CPP #-}
#if __GLASGOW_HASKELL__ >= VERSION_WITH_GHCCPP
{-# LANGUAGE GhcCPP #-}
#endif
```

It looks like this will not work. See  [getOptions'](https://github.com/ghc/ghc/blob/c318be56f69eb87f1cbb219dd660f09498857ece/compiler/GHC/Parser/Header.hs#L259) which pulls out the pragmas, *before* CPP can run.

And trying to load in ghci

```
{-# LANGUAGE CPP #-}
#if __GLASGOW_HASKELL__ >= 22
{-# LANGUAGE NONEXISTENT #-}
#endif
```

gives an error about the unsupported extension NONEXISTENT.

So, any suggestions as to how to enable this in a backward compatible way? It can't simply replace `CPP` because it is deliberately a subset, so will not work in all cases.

Perhaps use a new-style language pragma comment to flag that it is to be used. So

```
{-# LANGUAGE CPP #-}
{-! LANGUAGE GHC_CPP !-}
```

Or similar.  If that appears, CPP is turned off, and GHC_CPP turned on.

phadej:
```
#if __GLASGOW_HASKELL__ >= 22
{-# LANGUAGE NONEXISTENT #-}
#endif
```

22? We have 910 for GHC-9.10, so that case is always true.

VitWW:
It is better to have different syntax for better distinguishing CPP from GHC_CPP.
For example, for GHC_CPP must use double `##` :
```haskell
##if __GLASGOW_HASKELL__ >=800
myFoo x y = foo x y noArg
##else
myFoo x y = foo x y
##endif
```

alanz:
> It is better to have different syntax for better distinguishing CPP from GHC_CPP.

The goal is to have it as a drop-in replacement for normal CPP, for the 90%+ usages that do not do anything unusual.

proposals/0000-ghc-cpp.rst
```diff
@@ -0,0 +1,244 @@
...
+
+When enabled, the lexer has the additional rules:
+
+.. code:: haskell
+
+   cpp -> '#' cppkeyword [ anysymbol {any} ] newline
+         | '#' cppkeyword [ anysymbol {any} ] '\' newline cppcont
+   cppcont -> [ anysymbol {any} ] '\' newline cppcont
+         |  [ anysymbol {any} ] newline
+   cppkeyword -> 'define' | 'include' | 'undef' | 'error' | 'ifdef'
```
blamario, line 87: Your presentation slides here at ZuriHac say

> Pure, so no include files

Does that mean `include` should be excluded from the supported keyword list?
Also why is `undef` in the list? And why is `line` missing?

alanz: There are two stages, the first is to tokenise valid CPP directives, the second is to interpret them (or not).
So the lexer will scan these correctly, but we choose not to process them when they occur.

alanz:
FYI, I did a talk on my progress with this at HIW2025.  Slides [here](https://alanz.github.io/hiw2025/)

proposals/0000-ghc-cpp.rst
phadej: include files are quite important. While GHC does provide `MIN_VERSION_pkgname` macros nowadays, it still doens't provide all the macros which are in `cabal_macros.h` today, nor the potential future ones

Including `MachDeps.h` to get `WORD_SIZE_IN_BITS` is very common.

Packages use common include to reduce repetition, e.g. `containers`:

```
/*
 * Common macros for containers
 */

#ifndef HASKELL_CONTAINERS_H
#define HASKELL_CONTAINERS_H

/*
 * On GHC and MicroHs, include MachDeps.h to get WORD_SIZE_IN_BITS macro.
 */
#if defined(__GLASGOW_HASKELL__) || defined(__MHS__)
#include "MachDeps.h"
#endif

#if defined(__GLASGOW_HASKELL__) || defined(__MHS__)
#define DEFINE_PATTERN_SYNONYMS 1
#endif

#ifdef __GLASGOW_HASKELL__
# define USE_ST_MONAD 1
#ifndef WORDS_BIGENDIAN
/*
 * Unboxed arrays are broken on big-endian architectures.
 * See https://gitlab.haskell.org/ghc/ghc/-/issues/16998
 */
# define USE_UNBOXED_ARRAYS 1
#endif
#endif

#endif
```

There are many other examples where such common include file is used to translate version bounds to feature-check definitions (like `DEFINE_PATTERN_SYNONYMS `) above.

So, IMO, includes are essential. Whatever preprocessor mechanism is proposed it should have some kind of "modularization" approach.

blamario: If there's a new pragma (I suggested `CPP_HS_SUBSET`), it can be given the list of additional required preprocessor keywords as arguments, like

```
{-# CPP_HS_SUBSET +include #-}
```

This seems like a good thing to inform the compiler about, because it has implications for the build environment.

noughtmare: It seems feasible to have an `#include` directive which only includes `#define`ed macros (like `WORD_SIZE_IN_BITS`) to be used in conditional directives (e.g. `#ifdef` and `#if`). And perhaps it could also support transitive includes of such macros.

proposals/0000-ghc-cpp.rst
```diff
@@ -0,0 +1,244 @@
...
+- Ignoring
+
+In **Normal** mode it passes all non-cpp tokens on to the parser as
+normal.
+
+In **Ignoring** mode it converts all non-cpp tokens into comments,
+which are stored or not as per normal comment processing.
+
+In either case, the ``cpp`` and ``cppcont`` tokens are converted to
+comments, as above, and not passed on to the parser.
```
noughtmare, line 125:
```suggestion
In either case, the ``cpp`` and ``cppcont`` tokens are converted to
comments, as below, and not passed on to the parser.
```

proposals/0000-ghc-cpp.rst
```diff
@@ -0,0 +1,244 @@
...
+- Any preprocessor state will be inserted into an (opaque) field in
+  the parser state, made available to it by get and set operators.
+- Any include file processing will temporarily switch the lexer input,
+  then continue.  See https://gcc.gnu.org/onlinedocs/cpp/Header-Files.html
+
+
+The result of this processing is to put the preprocessor into one of two states
+
+- Normal
+- Ignoring
```
noughtmare, line 116: I'm guessing this is done to support the conditional directives and the `ignoring` mode is used in the branch that is not taken. It would be useful to spell that out explicitly.

proposals/0000-ghc-cpp.rst
```diff
@@ -0,0 +1,244 @@
...
+These will be passed to the actual preprocessor for processing. This will
+
+- Accumulate continued ``cpp`` and ``cppcont`` tokens until a full CPP
+  directive line is constructed.
+- Apply the standard CPP processing to this line. (Based on
+  https://gcc.gnu.org/onlinedocs/cpp/Overview.html)
+- Any preprocessor state will be inserted into an (opaque) field in
+  the parser state, made available to it by get and set operators.
+- Any include file processing will temporarily switch the lexer input,
+  then continue.  See https://gcc.gnu.org/onlinedocs/cpp/Header-Files.html
```
noughtmare, line 110: This screams for a non-trivial example. Let's consider the example below:

```
#define FOO
#ifndef FOO
x = 1
#else
x = 5
#endif
```

1. The tokenizer will see `#define` and then `FOO` and then a newline which completes the first line. (I assume that newlines always terminate a "directive line".) `FOO` is added as a defined constant.
2. `#ifndef FOO` is lexed similarly. Now since `FOO` is defined we switch to Ignoring mode.
3. We lex `x = 1` as normal Haskell, but wrap it in a comment.
4. We lex `#else` and switch back to Normal mode.
5. We lex `x = 5` as normal.
6. We lex `#endif` and do nothing since we are already in Normal mode.

noughtmare: I guess newlines can still be escaped with a backslash.

PR #626, opened 2023-11-29
#281 Visible forall in types of terms: data constructors, patterns, polytypes, drop the binder restriction

int-index:
A revision of #281 with clarifications and fixes

1. **Data constructors**
    The original intent behind the proposal was to allow `forall a ->` in data constructors, too, but this was only implied rather than stated explicitly. This revision makes this point clear. Assumes [#402 "Stable GADT constructor syntax"](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0402-gadt-syntax.rst) (accepted, not implemented).

2. **Patterns**
    The proposal states in Part 2
    > The syntactic descriptions here applying to expressions apply equally to patterns, though we will continue to discuss only expressions.

    It is meant to apply to T2T, too, but this was only implied rather than stated explicitly. The amendment solves this by introducing the notions of "T2T in Expressions" and "T2T in Patterns".

3. **Polytypes**
    The typing rules in the proposal used monotypes `rho` for required type arguments, even though the intent was to allow polytypes `sigma`. (Thanks to @JakobBruenker for spotting this)

4. **Drop the binder restriction**
   At the moment, the proposal requires a type pattern to be either a variable or a wildcard. This means that we allow this

   ```haskell
   data T a where
     Typed :: forall a -> a -> T a

   g_var :: T Int -> Int
   g_var (Typed a x) = x :: a    -- `a` is unified with `Int`, OK
   ```

   but not this:

   ```haskell
   g_con :: T Int -> Int
   g_con (Typed Int x) = x    -- `Int` is unified with `Int`, somehow not OK?
   ```

   Now, why is that? Both programs seem perfectly fine to me.

   The reason is that at the time of writing the proposal we didn't quite know how to address [#18986](https://gitlab.haskell.org/ghc/ghc/-/issues/18986). The restriction was meant to sidestep this issue.

   Fortunately, the issue has been fixed in GHC commit [2afbddb0](https://gitlab.haskell.org/ghc/ghc/-/commit/2afbddb0f24df20cfc42279bb8695f91bbd3c1e0). There are also ideas how to make the solution even better (see [#23639](https://gitlab.haskell.org/ghc/ghc/-/issues/23639)). So the restriction is wholly unnecessary now. In fact, it only makes things worse. Imagine we replaced `forall a ->` with `forall a.`

   ```haskell
   data T a where
     Typed :: forall a. a -> T a

   g_var :: T Int -> Int
   g_var (Typed @a x) = x :: a  -- OK

   g_con :: T Int -> Int
   g_con (Typed @Int x) = x     -- OK
   ```

   Both definitions are accepted. That's an inconsistency between visible and invisible forall, and resolving it is as simple as removing the restriction.

### DIFF
--- a/proposals/0281-visible-forall.rst
+++ b/proposals/0281-visible-forall.rst
@@ -712,7 +712,7 @@ Syntax
      infixpat2 ::=
             | infixpat
       (NEW) | infixpat      '->'  infixpat2      -- only with RequiredTypeArguments and /without/ ViewPatterns (conflict described below)
-      (MOV) | infixexp      '->'  infixpat2      -- only with ViewPatterns and /without/ RequiredTypeArguments (conflict described below)
+      (MOV) | infixexp      '->'  infixpat2      -- only with ViewPatterns (takes precedence over RequiredTypeArguments, conflict described below)
       (NEW) | infixpat mult '->'  infixpat2      -- only with RequiredTypeArguments
       (NEW) | infixpat      '->.' infixpat2      -- only with RequiredTypeArguments
       (NEW) | infixpat      '=>'  infixpat2      -- only with RequiredTypeArguments
@@ -748,9 +748,19 @@ Syntax
    as well.
 
 4. Adding ``p1 -> p2`` to the syntax of patterns (see the BNF of "types-in-terms"
-   above) is incompatible with ``ViewPatterns``. When both ``RequiredTypeArguments``
-   and ``ViewPatterns`` are enabled, the conflict is resolved in favor of ``ViewPatterns``.
-   The programmer can write ``type (t1 -> t2)`` or ``(->) t1 t2`` instead.
+   above) is incompatible with ``ViewPatterns``.
+
+   * When both ``RequiredTypeArguments`` and ``ViewPatterns`` are enabled,
+     the conflict is resolved in favor of ``ViewPatterns``. The programmer can
+     write ``type (t1 -> t2)`` or ``(->) t1 t2`` instead.
+
+   * When only ``ViewPatterns`` is enabled,
+     the conflict is resolved in favor of ``ViewPatterns``.
+
+   * When only ``RequiredTypeArguments`` is enabled, the conflict is not
+     resolved either way. It is reported to the user in an error message.
+     This can be reconsidered if an alternative syntax for view patterns is
+     implemented in GHC.
 
    To make the grammar of Haskell more regular and to simplify the
    implementation of "types-in-terms", change the precedence of view patterns
@@ -1510,12 +1520,6 @@ Alternatives
    (accepted, not implemented). In principle, nothing precludes us from
    extending the syntax of existential quantification in Haskell98-style data
    declarations in a similar manner, but nothing compels us to do so either.
-
-9. We could say that ``a -> b`` is a view pattern regardless of enabled
-   extensions and thereby make this proposal less fork-like; the problem is that
-   it would create an asymmetry: ``(t1 -> t2)`` would be allowed in types,
-   ``(e1 -> e2)`` would be allowed in expressions, but ``(p1 -> p2)`` would not
-   be allowed in patterns (here the arrow stands for the function type).
 
 Unresolved Questions
 --------------------
--- a/proposals/0281-visible-forall.rst
+++ b/proposals/0281-visible-forall.rst
@@ -358,9 +358,9 @@ Part I: Proposed Change Specification
      follow the rules shown in Figure 4 of "A quick look at impredicativity",
      extended as follows::
 
-        G |- sigma[a := rho];                 pis  ~>  Theta; phis; rho_r
-        ------------------------------------------------------------------  ITVDQ
-        G |- (forall a -> sigma); (type rho), pis  ~>  Theta; phis; rho_r
+        G |- sigma_b[a := sigma_a];                     pis  ~>  Theta; phis; rho_r
+        ---------------------------------------------------------------------------  ITVDQ
+        G |- (forall a -> sigma_b);     (type sigma_a), pis  ~>  Theta; phis; rho_r
 
    * In checking mode, in a function binding ``f (type x) = ...`` or a lambda
      ``\(type x) -> ...``, the ``x`` is a fresh skolem.
@@ -374,22 +374,57 @@ Part I: Proposed Change Specification
      x = f (type Int)   -- OK
      x = type Int       -- invalid use of a type in a term
 
-   This is checked during type checking, so Template Haskell is unaffected, and
-   ``[e| type Int |]`` is allowed (but different from ``[t| Int |]``).
+   This is checked during type checking, so Template Haskell is unaffected:
+   ``[e| type Int |]`` and ``[p| type Int |]`` are both allowed but different
+   from ``[t| Int |]``.
 
-   Furthermore, any pattern of form ``type t`` must be either a variable or a
-   wildcard::
-
-     f (type x)   = ...    -- OK
-     f (type _)   = ...    -- OK
-     f (type Int) = ...    -- invalid use of a type in a term
-
-   This is also checked during type checking, so Template Haskell must be able
-   to represent patterns such as ``[p| type Int |]``.
 
 6. **Erasure**. In types of terms, ``forall a ->`` is an erased quantifier.
    Making ``forall a ->`` erased in types of types is out of scope of this
    proposal.
+
+7. **Data constructors**. When ``RequiredTypeArguments`` is in effect, allow
+   ``forall a ->`` in data constructor declarations.
+
+   * The flavour of the parent declaration (data, data instance, newtype,
+     newtype instance) does not affect the validity of ``forall a ->``.
+
+   * The GADT constructor syntax (GHC Proposal `#402 <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0402-gadt-syntax.rst>`_)
+     supports ``forall a ->`` quantification both for universals and existentials::
+
+       data T a b where
+         MkT1 :: forall a b.   forall x y.   (x, y, a, b) -> T a b  -- OK
+         MkT2 :: forall a b.   forall x y -> (x, y, a, b) -> T a b  -- OK
+         MkT3 :: forall a b -> forall x y.   (x, y, a, b) -> T a b  -- OK
+         MkT4 :: forall a b -> forall x y -> (x, y, a, b) -> T a b  -- OK
+
+     In Haskell98-style syntax, all data constructors have invisible universal
+     quantifiers, and it is illegal to specify visible existential quantifiers::
+
+       data T a b =
+         forall x y.   MkT1 (x, y, a, b)    -- OK, `forall a b.` is added by GHC
+         forall x y -> MkT2 (x, y, a, b)    -- Illegal syntax
+
+     There is no inherent reason to rule out ``forall a ->`` in Haskell98-style
+     existentials other than to save on implementation costs (see "Alternatives").
+
+   * The use of ``forall a ->`` in the type of a data constructor is compatible
+     with ``DataKinds`` promotion of the said constructor.
+
+   * Pattern matching on a data constructor ``MkT :: forall a -> ...`` implies
+     the use of type patterns in positions that correspond to required type
+     parameters::
+
+       data Ex where { MkEx :: forall a -> Show a => a -> Ex }
+
+       f :: Ex -> String
+       f (MkEx (type a) x) = show (x :: a)
+       --      ^^^^^^^^
+       --      type pattern corresponding to (forall a ->) in the type of MkEx
+
+   * The rules for checking type patterns in constructors are derived from the
+     rules for checking type abstractions ``@a``, mutatis mutandis; for the
+     latter, see GHC Proposal `#448 <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0448-type-variable-scoping.rst>`_.
 
 Part I: Examples and Explanation
 --------------------------------
@@ -436,12 +471,36 @@ Part I: Examples and Explanation
      -- Usage
      str = symbolValVis (type "Hello, World")
 
+5. A type pattern in a constructor pattern:
+   ::
+
+     data T a b where
+       MkT :: forall a. forall b -> (a,b) -> T a b
+       --               ^^^^^^^^^^^^
+       --               visible forall in a data constructor
+
+     f :: T a b -> (b, a)
+     f (MkT @a (type b) c) = (snd c :: b, fst c :: a)
+     --        ^^^^^^^^
+     --        the corresponding type pattern
+
+6. A type pattern in a constructor pattern that requires unification:
+   ::
+
+       data U a where { MkU :: forall a -> U a }
+
+       f :: U (Maybe Int) -> Int
+       f (MkU (type (Maybe a))) = 42 :: a
+
+   The type pattern ``type (Maybe a)`` is unified with ``Maybe Int`` in the
+   signature. As a result, ``a`` stands for ``Int`` in ``42 :: a``.
+
 Note that as long as we limit ourselves to part I of this proposal, we need the
 ``type`` marker in all of the above examples, even when the argument is a
-syntactically valid term. If the programer were to write ``symbolValVis "Hello,
-World"``, they would get an error message stating that a term argument was
+syntactically valid term. If the programer were to write ``symbolValVis "Hello, World"``,
+they would get an error message stating that a term argument was
 received where a type argument was expected. That's because our typing rule
-``ITVDQ`` explicitly requires the argument to be of form ``type rho``.
+``ITVDQ`` explicitly requires the argument to be of form ``type sigma``.
 
 Could we extend our system to permit arguments without the ``type`` prefix?
 That is precisely the subject of part II.
@@ -570,14 +629,20 @@ information has been fully determined:
   1. Namespace selection syntax (under ``-XDataKinds``)
   2. Name quotation syntax (under ``-XTemplateHaskell``)
 
+* The meaning of ``->`` has been determined. It can stand for one of the following:
+
+  1. Part of lambda, case-of, or multi-if syntax, as in ``\x -> ...``
+  2. The function type constructor ``(->)``, as in ``Int -> Bool``
+  3. A view pattern ``f (e -> p) = ...`` (under ``-XViewPatterns``)
+
 Part II: Proposed Change Specification
 --------------------------------------
 
 Syntax
 ~~~~~~
 
-1. Extend the term syntax with several constructs that
-   previously could only occur at the type level:
+1. Extend the term syntax (expressions and patterns) with several constructs
+   that previously could only occur at the type level:
 
    * Function arrows: ``a -> b``
    * Multiplicity-polymorphic function arrows: ``a %m -> b`` (under ``-XLinearTypes``)
@@ -587,33 +652,92 @@ Syntax
 
    We will call them **types-in-terms**.
 
-   Grammatically, their constituents are terms, not types::
+   To that end, we change the grammar of expressions and patterns as follows.
+   Start with these nonterminals based on the `Haskell 2010 Report <https://www.haskell.org/onlinereport/haskell2010/haskellch10.html#x17-18000010.5>`_,
+   modified to account for pattern signatures, view patterns, and the ``type`` herald::
 
-                   proposed grammar:                      as opposed to:
-         ┌────────────────────────────────────┬───────────────────────────────────────┐
-         │                                    │                                       │
-         │  exp ::=                           │    exp ::=                            │
-         │      | exp₀ '->' exp₁              │        | type₀ '->' type₁             │
-         │      | exp₀ '=>' exp₁              │        | type₀ '=>' type₁             │
-         │      | 'forall' tv_bndrs '.'  exp  │        | 'forall' tv_bndrs '.'  type  │
-         │      | 'forall' tv_bndrs '->' exp  │        | 'forall' tv_bndrs '->' type  │
-         │                                    │                                       │
-         └────────────────────────────────────┴───────────────────────────────────────┘
+     exp ::=
+       | 'type' ktype                    -- only with ExplicitNamespaces (see Part I of #281)
+       | infixexp :: [context =>] type
+       | infixexp
 
-   This is a necessity to avoid parsing conflicts, with the following
-   consequences:
+     viewpat ::=
+       | pat
+       | exp '->' viewpat                -- only with ViewPatterns
 
-   1. The ``'`` symbol signifies Template Haskell name quotation rather than ``DataKinds`` promotion.
-   2. The ``*`` symbol is treated as an infix operator regardless of ``-XStarIsType``.
-   3. Built-in syntax for tuples and lists is interpreted as in terms.
-      That is, ``[a]`` is a singleton list rather than the type of a list,
-      and ``(a, b)`` is a pair rather than the type of a pair.
+     pat ::=
+       | 'type' ktype                    -- only with ExplicitNamespaces (see Part I of #281)
+       | infixpat :: [context =>] type   -- only with PatternSignatures  (see #448)
+       | infixpat
 
-2. The syntactic descriptions here applying to expressions apply equally to patterns, though
-   we will continue to discuss only expressions.
+     infixexp ::=
+       | lexp qop infixexp
+       | - infixexp
+       | lexp
 
-3. Make ``forall`` a keyword at the term level. Not guarded by any extension
-   (same motivation as `#193 <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0193-forall-keyword.rst>`_).
+     infixpat ::=
+       | lpat qconop infixpat
+       | lpat
+
+   Introduce the following changes:
+
+   * add new non-terminals ``infixexp2`` and ``infixpat2`` that include terms-in-types
+   * update ``exp`` and ``pat`` to use the new non-terminals in place
+     of ``infixexp`` and ``infixpat`` respectively
+   * rearrange ``viewpat`` to avoid parsing conflicts
+
+   ::
+
+     exp ::=
+       | 'type' ktype                    -- only with ExplicitNamespaces (see Part I of #281)
+       | infixexp2 :: [context =>] type
+       | infixexp2
+
+     viewpat ::= pat  -- the other production has been moved to infixpat2
+
+     pat ::=
+       | 'type' ktype                    -- only with ExplicitNamespaces (see Part I of #281)
+       | infixpat2 :: [context =>] type  -- only with PatternSignatures  (see #448)
+       | infixpat2
+
+     infixexp2 ::=
+            | infixexp
+      (NEW) | infixexp      '->'  infixexp2      -- only with RequiredTypeArguments
+      (NEW) | infixexp mult '->'  infixexp2      -- only with RequiredTypeArguments
+      (NEW) | infixexp      '->.' infixexp2      -- only with RequiredTypeArguments
+      (NEW) | infixexp      '=>'  infixexp2      -- only with RequiredTypeArguments
+      (NEW) | 'forall' tv_bndrs '.'  infixexp2   -- only with RequiredTypeArguments
+      (NEW) | 'forall' tv_bndrs '->' infixexp2   -- only with RequiredTypeArguments
+
+     infixpat2 ::=
+            | infixpat
+      (NEW) | infixpat      '->'  infixpat2      -- only with RequiredTypeArguments and /without/ ViewPatterns (conflict described below)
+      (MOV) | infixexp      '->'  infixpat2      -- only with ViewPatterns and /without/ RequiredTypeArguments (conflict described below)
+      (NEW) | infixpat mult '->'  infixpat2      -- only with RequiredTypeArguments
+      (NEW) | infixpat      '->.' infixpat2      -- only with RequiredTypeArguments
+      (NEW) | infixpat      '=>'  infixpat2      -- only with RequiredTypeArguments
+      (NEW) | 'forall' tv_bndrs '.'  infixpat2   -- only with RequiredTypeArguments
+      (NEW) | 'forall' tv_bndrs '->' infixpat2   -- only with RequiredTypeArguments
+
+     -- infixexp and infixpat are unchanged
+
+   Note that the constituents of terms-in-types use the term syntax.
+   For example::
+
+                   proposed grammar:                    as opposed to:
+         ┌──────────────────────────────────────┬───────────────────────────────────┐
+         │  infixexp2 ::=                       │  infixexp2 ::=                    │
+         │    | infixexp '->' infixexp2         │    | btype '->' ctype             │
+         │    | 'forall' tv_bndrs '.' infixexp2 │    | 'forall' tv_bndrs '.' ctype  │
+         │    | ...                             │    | ...                          │
+         └──────────────────────────────────────┴───────────────────────────────────┘
+
+   The use of term syntax on the LHS of ``e1 -> e2`` and ``e1 => e2`` is a
+   necessity to avoid parsing conflicts. The use of term syntax on the RHS is
+   done for consistency with the LHS.
+
+3. Make ``forall`` a keyword at the term level (in expressions and patterns).
+   Not guarded by any extension (same motivation as `#193 <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0193-forall-keyword.rst>`_).
    This implies ``forall`` is no longer a valid identifier.
 
    For three releases before this change takes place, include a new warning
@@ -623,8 +747,26 @@ Syntax
    This change applies to ``∀`` (the ``UnicodeSyntax`` rendition of ``forall``)
    as well.
 
-4. When ``ViewPatterns`` are enabled, interpret ``f (a -> b) = ...``
-   as a view pattern, otherwise as ``f ((->) a b) = ...``.
+4. Adding ``p1 -> p2`` to the syntax of patterns (see the BNF of "types-in-terms"
+   above) is incompatible with ``ViewPatterns``. When both ``RequiredTypeArguments``
+   and ``ViewPatterns`` are enabled, the conflict is resolved in favor of ``ViewPatterns``.
+   The programmer can write ``type (t1 -> t2)`` or ``(->) t1 t2`` instead.
+
+   To make the grammar of Haskell more regular and to simplify the
+   implementation of "types-in-terms", change the precedence of view patterns
+   relative to pattern signatures as follows::
+
+        f (e -> p :: t)   = ...     -- user-written function LHS
+        f (e -> (p :: t)) = ...     -- old parse
+        f ((e -> p) :: t) = ...     -- new parse
+
+   The new parse is consistent with kind signatures, e.g. ``Int -> Bool :: Type``
+   is parsed as ``(Int -> Bool) :: Type``, not ``Int -> (Bool :: Type)``.
+
+   For three releases before this change takes place, include a new warning
+   ``-Wview-pattern-signatures`` in ``-Wdefault`` to warn on affected code.
+   The warning text will suggest to parenthesize the RHS of the view pattern.
+   See "Effect and Interactions" for the impact analysis.
 
 5. ``case ... of x -> y -> z`` is an error. We require parentheses to
    disambiguate:
@@ -700,16 +842,20 @@ Type checking
 8. Generalize the ``ITVDQ`` rule introduced earlier
    by using ``t2t``::
 
-     rho = t2t(e)
-     G |- sigma[a := rho];         pis  ~>  Theta; phis; rho_r
-     ---------------------------------------------------------- ITVDQ-T2T
-     G |- (forall a -> sigma);  e, pis  ~>  Theta; phis; rho_r
+     sigma_a = t2t(e)
+     G |- sigma_b[a := sigma_a];         pis  ~>  Theta; phis; rho_r
+     --------------------------------------------------------------- ITVDQ-T2T
+     G |- (forall a -> sigma_b);      e, pis  ~>  Theta; phis; rho_r
 
-   ``t2t`` transforms term arguments into type arguments, see "T2T-Mapping"
-   below for an informal definition.
+   ``t2t`` transforms term arguments into type arguments, see the "T2T-Mapping"
+   section for an informal definition of ``t2t``.
 
-   In other words, given ``f :: forall a -> t``, the ``x`` in ``f x`` is
-   parsed and renamed as a term, but then mapped to a type.
+   That is, given ``f :: forall a -> t``, the argument ``e`` in ``f e``
+   is parsed and renamed as a term, but then mapped to a type.
+
+   In the same way, generalize the type checking rule for patterns to invoke
+   ``t2t`` to transform a pattern to a type when the function or constructor has
+   a visible forall in its type.
 
 9.  Any uses of terms in types are ill-typed:
     ::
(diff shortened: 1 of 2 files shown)


Ericson2314 (approved):

nomeata:
Remind me again: if I can specify `Int`, I should also be able to specify an in scope type variable (else we lose the ability to abstract), shouldn't I? But won't that look the same as binding a fresh variable? Or are the two indistinguishable anyways?

int-index:
> if I can specify `Int`, I should also be able to specify an in scope type variable

In a type yes, but not in a type pattern! I see the problem you're getting it, but we have the exact same problem in term-level patterns. I can pattern-match on literals:

```
isMyFavoriteNumber 42 = True
isMyFavoriteNumber _  = False
```

But I can't abstract over them:

```
favNum = 42

isMyFavoriteNumber favNum = True    -- Oh no! Doesn't do what I intended
isMyFavoriteNumber _      = False
```

Type patterns attempt to be as close as possible to term patterns, and these are the rules of the games: lowercase names bind fresh variables.

simonpj:
Question
```
data T a where
  Typed :: forall a -> a -> T a
```
Do we already allow a visible `forall a -> ...` in the type of a data constructor?  Presumably with `RequiredTypeArguments`?  The user manual does not say (see [manual entry[(https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/required_type_arguments.html?highlight=requiredtype#extension-RequiredTypeArguments))

int-index:
> Do we already allow a visible `forall a -> ...` in the type of a data constructor?

Not yet, as `Note [TyVarBinders in DataCons]` explains. But we will, see ticket [#18389](https://gitlab.haskell.org/ghc/ghc/-/issues/18389) and the accepted GHC Proposal [#402](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0402-gadt-syntax.rst)

simonpj:
> At the moment, the proposal requires a type pattern to be either a variable or a wildcard

Really?  What proposal?  Here is [the relevant passage of proposal 448](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0448-type-variable-scoping.rst#42proposed-change-specification) which is, I believe, what you are referring to.

That section does not restrict type patterns in the way  you say, does it?

int-index:
> Really? What proposal? 

The one I'm amending, 281 :-)

See the diff. (This is a PR, not an GitHub issue)

simonpj:
Richard and I had a good discussion about this.  Here is a summary.

* **Type patterns**. [GHC proposal 448](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0448-type-variable-scoping.rst) says (not very explicitly):
  * Section 5.2.  In general, an _arbitrary pattern_ can be of form `@a` where `a` is a type variable.  E.g
    ```
    f1 :: forall a. a -> blah
    f1 @a (x::a) = ...
    ```

  * Section 4.2. The *argument of a constructor pattern* can take the more general form `@ty` where `ty` is a type.  e.g.
    ```
    f2 :: Maybe Int -> blah
    f2 (Just @Int x) = ...
    ```
    Section 4.2 has a special syntax
    ```
    pat ::= gcon tyapp_or_pat1 … tyapp_or_patk | ...
    tyapp_or_pat ::= apat
                  |  '@' atype    -- The new bit
    ```

* **Polymorphic kinds**. Consider
  ```
  data T f where
    MkT :: forall (f:: forall k. k-> Type). f Int -> f Maybe -> T f

  g :: T Proxy -> blah    -- Proxy :: forall k. k -> Type
  g (MkT @b x y) = ...
  ```
  Does `g` typecheck, binding `b` to `Proxy`, and giving `b` the poly-kind `forall k. k->Type`?  Surely it should, but doing so took some fancy footwork, described in `Note [Type patterns: binders and unifiers]` of GHC.Tc.Gen.HsType and [GHC ticket](https://gitlab.haskell.org/ghc/ghc/-/issues/18986).

* **Data constructors with visible type arguments**.  [GHC proposal 281 on visible forall](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0281-visible-forall.rst) does not explicitly discuss data constructors.  But one might reasonably interpret it as allowing visible foralls in data constructors:
  ```
  data T a where
    Typed :: forall a -> a -> T a
  ```
  **Action**: this PR (which amends 281) should make explicit that data constructors can have explicit foralls.
  Here [GHC proposal 402](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0402-gadt-syntax.rst) is relevant, and [GHC ticket #18389](https://gitlab.haskell.org/ghc/ghc/-/issues/18389).

  Not yet implemented.

* Now consider our `f2` example above, but this time suppose that a data constructor has a visible type argument, rather than invisible as in `f2`:
  ```
  f2vis :: T Int -> Int
  f2vis (Typed (type Int) x) = x
  ```
  If `f2` is OK, surely `f2vis` should be OK. That is, it should be fine to have a non-tyvar type pattern there. The natural thing is to further extend `tyapp_or_pat`, like this:
  ```
  pat ::= gcon tyapp_or_pat1 … tyapp_or_patk | ...
  tyapp_or_pat ::= apat
                |  '@' atype
                |  'type' atype    -- The new bit
  ```
  **This is the restriction that Vlad wants to lift, and we agree**.

* On the other hand, it makes no sense to allow `\(type Int) ->  blah`.  **So let's keep it uniform with Section 5.2 of 448**: a general pattern can be `@a` or `type a`; in neither case can we have an arbitrary type.  (In constructor argument patterns, both are liberalised uniformly.)

  This is different to what Vlad proposes -- but only superficially; `\(type Int) -> blah` will certainly be rejected somewhere!

* When we lack the `type` herald, of course the *parser* or *renamer* cannot do the rejection; we have to wait till the type checker.  That's fine; it's an implementation matter.

* One other point came up:
  * Suppose `K1` as an invisible type argument, while `K2` and `K3` has a visible type arg. Consider
    ```
    f1 (K1 @(Int -> Int ))      = ...
    f2 (K2 (type (Int -> Int))) = ...
    f3 (K3 (Int -> Int))        = ...
    ```
    In `f1` and `f2` it's clear that we are parsing types so the `->` is a function arrow.
    But in `f3`, it could be a view pattern; and we won't find out until the type checker; but it affects scoping so that the renamer needs to know. Section 7.1 of 281 says "When `ViewPatterns` are enabled, interpret `f (a -> b) = ...` as a view pattern, otherwise as `f ((->) a b) = ....`, but that seems unpleasantly fork-like.

    We suggest always interpreting `(a->b)` as a view patterns (regardless of `ViewPatterns`). If you want
    a function type, use previx `(->)` or the `type` herald.

    **Action**: this could be an amendments in the current PR.

  * Section 7.1 says "The syntactic descriptions here applying to expressions apply equally to patterns, though we will continue to discuss only expressions.", but that is too cryptic. We need to describe the P2T mapping, perhaps not in full glory, but more than we have now.  In 7.1 we need to say that we need to extend patterns with type syntx.

    **Action**: explain P2T.

simonpj:
@int-index and I had a useful conversation and I updated my writeup in the previous comment as a result.

Main takeaway: the insights of the above writeup cost me a couple of hours to figure out.  I would love them to be absolutely clear in the revised version of 281.  It is tricky territory!

int-index:
@nomeata Alright, let's go over the examples. We start with these definitions (I took the freedom to rename them to my liking)

```haskell
data InvT a where InvTyped :: forall a.   a -> InvT a
data VisT a where VisTyped :: forall a -> a -> VisT a
```

The only difference between them `InvTyped` and `VisTyped` is the visibility of the forall-bound variable. One would reasonably expect the following pattern matches to behave identically:

```haskell
f (InvTyped @t x) = ...
g (VisTyped t x) = ...
```

Per the VOP, we shouldn't have any differences in the way we parse, rename, and type-check `@t` in `f` and `t` in `g` respectively. Of course, punning is an obstacle to fully achieving this goal, but at least we could uphold VOP in the absence of punning.

Now let me add a third definition

```haskell
data StrT a where StrTyped :: forall {a}. String -> a -> StrT a 
```

Here we have an ordinary `String` value in the position where `VisT` has a required type parameter. So let's compare the following pattern matches:

```haskell
g (VisTyped t x) = ...
h (StrTyped s x) = ...
```

This time the LSP (Lexical Scoping Principle) kicks in. There shouldn't be any differences in the way we parse and rename `t` in `g` and `s` in `h`, because we must do so _without involving the type system_.

Let's now line up all three examples we have

```
f (InvTyped @t x) = ...
g (VisTyped t x) = ...
h (StrTyped s x) = ...
```

It follows, then, that renaming of `@t` in `f`, renaming of `t` in `g`, and renaming of `s` in `h` should all use the same algorithm to decide between usages and bindings.

I want to draw particular attention to this. We have here:
* `@t` – a type abstraction
* `t` – a term pattern that turns into a type pattern via T2T
* `s` – a term pattern
And our goal is to come up with a procedure to decide between usages and bindings that treats all three cases consistently.

But of course, we already have such a procedure. In the term pattern `s`, the `s` is a binding, end of story. That's how term pattern works, a lowercase name binds a variable (unless it's in a pattern signature `p :: sig`, where other rules apply).

So this decision propagates from `s` to `t` via the LSP. And then it propagates from `t` to `@t` via the VOP. This means that in `s`, `t`, and `@t`, any lowercase name is a binding, not usage (unless it's in a kind signature `t :: ksig`, where the rules for `p :: sig` apply again).

Based on this transitive application of our design principles, it is rather easy to address all of your examples.

```haskell
foo2 :: forall a. a -> a
foo2 @a x = … -- nomeata: here, a is brought into scope
              -- int-index: it does not matter
  where
    f :: InvT a -> blah
    f (InvTyped @a x) = ...
      -- nomeata: does this refer to a or bind a?
      -- int-index: bind `a`; imagine it was a term pattern (StrTyped s x) = ...,
      --                      then the `s` would be a binding, even if there was
      --                      another variable named `s` in scope!

    f' :: InvT Int -> blah
    f' (InvTyped @a x) = ...
      -- nomeata: will this give a type error, a not Int, or bind a?
      -- int-index bind `a`; again, by analogy with (StrTyped s x) = ...

    fv :: VisT a -> blah
    fv (VisTyped a x) = ...
      -- nomeata: does this refer to a or bind a?
      -- int-index bind `a`; again, by analogy with (StrTyped s x) = ...

    fv :: VisT Int -> blah
    fv (VisTyped a x) = ...
      -- nomeata: will this give a type error, a not Int, or bind a? 
      -- int-index bind `a`; again, by analogy with (StrTyped s x) = ...

    fx :: VisT a -> blah
    fx (VisTyped (type a) x) = ...
      -- nomeata: maybe the (type) herald changes this behaviour?
      -- int-index: it does not, and I'd find it surprising if it did,
      --            as it's purpose is to switch namespaces resolution,
      --            not binding/usage behavior

    fx :: VisT Int -> blah
    fx (VisTyped (type a) x) = ...
      -- nomeata: maybe the (type) herald changes this behaviour?
      -- int-index: it does not
```

nomeata:
There is no syntactic difference, but a semantic one:

* The argument to `Typed` is a type: it gets unified with what is there, and can't fail at runtime.
* The `x` is a term: it gets compared, and can fail at runtime, and this guides which branch to take. (It would be not un reasonable to allow `x` here with an Eq instance around, but it'd still be something quite different).

This is not a minor difference, but a crucial aspect! 

I don't have a great solution.

The least bad I can think of that sticks to the LSP and SUP is a form of `XNoShadowing`, so that we can tell lexically whether `x` is an occurrence or a binder. For terms this would be an error (unless and until someone wants the Eq-bases interpretation). 

The next best might be some syntax annotation for occurrences. `g .x`, hunting at that `x` is a bit like a qualified name `Prelude.pi`, only that the local namespace is the empty string. Or something like that.

In any case I believe that we will want to use abstracted types (and thus, SUP and all, terms) in patterns somehow.

int-index:
> There is no syntactic difference, but a semantic one

Absolutely, but that's what the LSP is about: a semantic difference can't affect our decisions between a usage and a binding. We should be able to look at the syntax alone to figure that out.

> a form of `XNoShadowing`

Sure, that'd be fine as well. We don't have to commit to an `Eq`-based desugaring, though it's something I'd fancy. This is about semantics, and we are here to discuss syntax and name resolution. Once we agree that `g x = ...` can be a use of an in-scope `x`, we are free to assign it either `Eq`-based semantics, some other semantics or outright reject.

The point is: semantics do not affect syntax (and name resolution).

>  some syntax annotation for occurrences. `g .x`

Yes, that'd work too. There's some wiggle room in the design space, which is why I offered to collaborate on a separate proposal. The issue is to tricky to address it in this amendment.

nomeata:
> The point is: semantics do not affect syntax (and name resolution).

Well, in so far as semantics put up requirements for syntax. If the semantics says “sometimes it is useful and necessary to refer to a variable in scope”, then the syntax has to allow it. 

> Yes, that'd work too. There's some wiggle room in the design space, which is why I offered to collaborate on a separate proposal. The issue is too tricky to address it in this amendment.

Fair enough. But only if we are confident that we can decide on this proposal without knowing where we will go in the end? 

For example, how about

```
data a * b = Tuple a b
foo (*) = …
  bar (MkT (*) = …
```
Is this a mention of the type constructor `(*)`? A mention of the term `(*)`? A binding of `(*)`?

nomeata:
> Is that the case for (erased) dependent quantification? I'm struggling to come up with an example where it's useful

I have at least a snarky answer: Give me an example where it’s useful to refer to `Int` and I’ll turn it into an example where it is useful to refer to `a`. :-)

int-index:
> Well, in so far as semantics put up requirements for syntax. If the semantics says “sometimes it is useful and necessary to refer to a variable in scope”, then the syntax has to allow it.

Yes, when we design the syntax, we are driven by the needs of the desired semantics. But when we consider an individual code snippet like `g (VisTyped t x) = ...` or `h (StrTyped s x) = ...`, we should be able to reason about its syntax without taking the semantics of individual elements into consideration, e.g. here it doesn't matter that `VisTyped` and `StrTyped` have different types.

I think we're on the same page here, right?

> Is this a mention of the type constructor `(*)`? A mention of the term `(*)`? A binding of `(*)`?

It is, sure enough, a binding of `(*)` in your example `bar (MkT (*)) = …`, because this operator does not start with `:`. If, on the other hand, you had `bar (MkT (:*)) = ...`, it would be a mention of `(:*)`. This is what the term-level lexical rules dictate, and we are in term-level syntax.

Sadly, `TypeOperators` are not consistent with term-level operators (a known wart in the Haskell syntax). So when we switch to type-level syntax (either via `@` or `type`), the renamer will treat all operators as usages, even if it does not start with `:`.

Thus in your other example `MkT @((*))` it is a mention of the `(*)` type operator, while it should be a binding of a new type variable named `(*)`.

int-index:
@simonpj I'm not sure what parts of your comment are just observations and what parts are meant to be incorporated in the proposal, so I followed the suggestions that you marked as "Action", of which there are three

1. (Data constructors) **Action:** this PR (which amends 281) should make explicit that data constructors can have explicit foralls.
2. (View patterns) **Action:** this could be an amendments in the current PR.
3. (Patterns) **Action:** explain P2T.

Here is how I addressed them

1. Data constructors
    * In part 1 of the proposal, I have added section 4.7 "Data constructors" that allows visible forall in data constructors and refers to #402 and #448 where appropriate
    * I have extended section 5 "Part I: Examples and Explanation" with two new examples that show the use of type patterns, one that simply binds a variable and one that requires unification
    * In section 11 "Alternatives" I explained why we do not allow `forall a ->` in Haskell98-style constructors
2. View patterns
    * In section 6.1 "The Resolved Syntax Tree" I specified that the meaning of `->` should have been resolved between lambdas `\a -> b`, function types `Int -> Bool`, and view patterns `e -> p`
    * In section 11 "Alternatives" I have mentioned your suggestion of making the proposal less fork-like, where I explained why I don't like this idea
3. Patterns
    * In part 2 section 7.4 "Type checking", item 10, there is now a mention of type patterns and T2T in patterns
    * Section 7.6 "T2T-Mapping" now specifies both T2T in expressions and T2T in patterns, with commonalities factored out
    * I have extended section 8 "Part II: Examples" with two examples that use T2T in patterns

Does that address your comments or would you like more details incorporated into the proposal?

simonpj:
@nomeata, I think that you are perhaps suggesting that
* The choices we have embodied in the already-accepted proposals #448 have some consequences you don't like.
* It's not clear what other design choices would do better.

Is that right?   I'm not clear about what you don't like.  Do you have some examples where you think the behavior is "wrong"?   For example you say

> Because it clashes with the expectations that anywhere where I can mention a type like Int I should be able to mention a type like Maybe a, TF a or simply a, where a is scope.

Can you turn that into a concrete example?  For example
```
f :: forall a. blah
f x = stuff
     where
        g (MkT @(Maybe a) (Just x)) = ...
```
In the binding of `g`, both `a` and `x` are in scope; but in patterns lower-case variables are binders, so so the pattern `MkT @(Maybe a) (Just x)` re-binds both `a` and `x`.  This seems entirely in line with expectations, doesn't it?

There is a wart around `TypeOperators` because `*` can be a type *constructor* like `Int`, and yet it looks "lower case" (does not start with a `:`.   So, in a pattern, say `Just (a * b)`, is `*` an *occurrence* like `Just`, or a *binder* like `a` and `b`?  Vlad?  But that's a distraction.

nomeata:
> In the binding of `g`, both `a` and `x` are in scope; but in patterns lower-case variables are binders, so so the pattern `MkT @(Maybe a) (Just x)` re-binds both `a` and `x`. This seems entirely in line with expectations, doesn't it

Not with my expectations, no. In local contexts there are types that I cannot write down without mentioning a type variable. So _if_ it is useful to say `Maybe` or `Int` in these positions, then it follows that it is useful to refer to an in-scope abstract type `a` in some way. Maybe in a syntactically different way (`g (MkT @(Maybe .a))` or so, but it has to be possible, else we lose the ability to abstract.

I’d say that any example where it is necessary to mention type constructors in patterns (is there one?) can be turned into one where it is necessary to have type variable occurrences.

proposals/0281-visible-forall.rst
```diff
@@ -700,10 +750,10 @@ Type checking
...
-     G |- sigma[a := rho];         pis  ~>  Theta; phis; rho_r
-     ---------------------------------------------------------- ITVDQ-T2T
-     G |- (forall a -> sigma);  e, pis  ~>  Theta; phis; rho_r
+     sigma_a = t2t(e)
+     G |- sigma_b[a := sigma_a];         pis  ~>  Theta; phis; rho_r
+     --------------------------------------------------------------- ITVDQ-T2T
+     G |- (forall a -> sigma_b);      e, pis  ~>  Theta; phis; rho_r
 
    ``t2t`` transforms term arguments into type arguments, see "T2T-Mapping"
    below for an informal definition.
```
simonpj, line 759: We need a new point

9.  In the same way, generalise the typechecking rule for patterns, to invoke `p2t` to transform a pattern to a type, when the constructor has a visible forall in its type (see Section 7.6).

proposals/0281-visible-forall.rst
```diff
@@ -731,23 +781,18 @@ Type checking
...
+       f (MkEx a x) = show (x :: a)
+       --     ^^^
+       --     the `a` corresponds to `forall a ->` in the type of `MkEx`
 
-    This is a conservative decision that can be revised in a later proposal.
+    In general, we permit term patterns in positions where type patterns are
+    expected by applying the T2T transformation, see "T2T-Mapping" below.
 
 Namespaces vs Semantics
 ~~~~~~~~~~~~~~~~~~~~~~~
```
simonpj, line 798: This is an odd, out of place section.  Why not make it item (6) of 7.1?   And please give an example or two.

There is a numbering bug... section numbers jump from 7.1 to 7.4. Peculiar.

nomeata:
There is maybe another way I can try to explain my concern.

Types are not matched, they are unified; if I write `MkT @(Maybe a)`, then `Maybe a =?= t` is thrown into the mix, where `t` is the type variable bound by `MkT`. Therefore, the analogue in term land isn’t term matching, but is the (hypothetical) feature of comparing against arbitrary expressions:
```
foo n (fib 5) = "yes, the second arg is fib of 5
```
We don’t have it, but it’s not hard to imagine (it is not implementation-wise not a big step from literal matching or pattern synonyms). Imagine we had that, then surely we’d expect
```
foo n (fib n) = "yes, the second arg is fib of first arg
```
to work as well, wouldn’t we? We wouldn’t expect this to bring a new `n` into scope that’s somehow resolved to be a preimage of the first argument under `fib`? Or even _if_ we’d expect that, wouldn’t we expect to be able to express both of these possible interpretations?

I think this is – roughly – what’s going on the type level, with `Maybe` instead of `fib`.

proposals/0281-visible-forall.rst
```diff
@@ -1242,6 +1341,33 @@ Alternatives
...
+
+   A similar limitation in GADTs is addressed by GHC Proposal `#402 <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0402-gadt-syntax.rst>`_
+   (accepted, not implemented). In principle, nothing precludes us from
+   extending the syntax of existential quantification in Haskell98-style data
+   declarations in a similar manner, but nothing compels us to do so either.
+
+9. We could say that ``a -> b`` is a view pattern regardless of enabled
+   extensions and thereby make this proposal less fork-like; the problem is
+   that it would create an asymmetry between use sites ``f (MkT (Int -> Bool))``
+   and definition sites ``f (MkT (a -> b)) = ...``.
```
simonpj, line 1369: I did not understand the asymmetry.  Could you spell it out please?

Perhaps you mean: `(e1 -> e2)` would be allowed in terms but `(p1 -> p2)` would not be allowed in patterns?   Better to say it like that, if that's what you mean.

And yes, that's asymmetric, but that's what happens if you add `ViewPatterns`.  The question is whether the asymmetry changes with other (apparently unrelated) extensions.

int-index:
> So, in a pattern, say `Just (a * b)`, is `*` an occurrence like `Just`, or `a` binder like `a` and `b`? Vlad?

This depends on whether we are in **term syntax** (e.g. `f pat = ...`) or **type syntax** (`f (type pat) = ...` or `f @pat = ...`).

1. **Term syntax**
   We are in patterns and `*` is "lowercase" (not starting with `:`), so it is a parse error. Let us consider all four cases: alphanumeric uppercase `K`, alphanumeric lowercase `k`, operator "lowercase" `*`, and operator "uppercase" `:*`

   Based on your example, here's what we get
   ```
   f (Just (K a b))  = ... -- usage of K
   f (Just (a :* b)) = ... -- usage of (:*)
   f (Just (k a b))  = ... -- parse error
   f (Just (a * b))  = ... -- parse error
   ```

   But hey, I thought we were discussing usage-vs-binding instead of usage-vs-parse-error? This happens because one does not simply apply one `VarPat` to another `VarPat` in term syntax. We could get the comparison we wanted if we discarded `a` and `b`

   ```
   g (Just K)    = ... -- usage of K
   g (Just (:*)) = ... -- usage of (:*)
   g (Just k)    = ... -- binding of k
   g (Just (*))  = ... -- binding of (*)
   ```

   That's better. Now we see the rules in patterns that use term syntax: lowercase identifiers are binders, uppercase identifiers are usages.

2. **Type syntax**

   First of all, let's get `StarIsType` out of the way. If it is enabled, then `*` is an occurrence and it stands for `Data.Kind.Type`. We are concerned with `NoStarIsType` and `TypeOperators`.

   Here's what we get if we adapt our examples above to use type syntax

   ```
   f (Just @(K a b))  = ... -- usage of K
   f (Just @(a :* b)) = ... -- usage of (:*)
   f (Just @(k a b))  = ... -- binding of k
   f (Just @(a * b))  = ... -- usage of (*)  :-(

   g (Just @K)    = ... -- usage of K
   g (Just @(:*)) = ... -- usage of (:*)
   g (Just @k)    = ... -- binding of k
   g (Just @(*))  = ... -- usage of (*)  :-(
   ```

   This time we don't get parse errors because type pattern syntax is more permissive than term pattern syntax. But we also get a discrepancy between the treatment of "lowercase" operators: they are bindings in term syntax and usages in type syntax. I've marked the inconsistencies with a sad face.

We could try to address it, but not without breaking existing programs.

nomeata:
> As it happens, f doesn't have a type sig, but xs :: [Maybe <something>]. When we pattern match, we may want a name for <something>.

Would
```
f xs = ..(isJust x)... case xs of (:) @(_ a) x _ -> ....
```
work here just as well? Or is it _necessary_ to mention `Maybe`?

simonpj:
If you have the `isJust` you can replace the `Maybe` with `_`.   And maybe  you can regardless.  But I would regard that as making the program worse (less easy to understand).

I'm sure you asked the question for a reason, perhaps relating to some proposed alternative set of rules that you have in your head.  But I'm not sure what they are.   (And of course the right place to debate them would be on a new proposal thread or issue... we are just piggy backing here on Vlad's tidy up.)

nomeata:
Thanks for bearing with me. Now that we have established that it’s necessary or desireable to desirable to have an occurrence `Maybe` in 
```
f xs = ..(isJust x)... case xs of (:) @(Maybe a) x _ -> ....
``
then doesn't it follow immediately that it is necessary or desirable to have an _occurrence_ of `m` in a more abstract context:
```
withF :: forall (m : Type -> Type). (forall a. m a -> Bool) -> …
withF isJust = …
  where
    f xs = ..(isJust x)... case xs of (:) @(m a) x _ -> ....
```

I’ll anticipate possible rebuttals:

* It still works with `m` being a binding
 
  Yes, but more by accident, because in this example, the `Maybe` wasn’t really _necessary_. I expect that we can construct examples where one really _has_ to specify the type in a pattern.

* You can’t abstract over constructors in terms, why should you here?
  
  Because `Maybe` is more like `fib` than like `Just`. Surely `type MyMaybe = Maybe` has more of a `my_fib = fib` feel than introducing a type variable. (I understand that I am not substantiating this point of view as much as I wish.)

* Granted, but this is too obscure.

  That’s hard to rerebut. If we don’t value the ability to abstract enough to take it into account in our planning, then I guess that’s an option.


> we are just piggy backing here on Vlad's tidy up

I doµ’t think it’s tidying up: This is where we are introducing occurrences of types bound at the top-level (aka `Maybe`) in terms, and that’s when I'd say we have to think about occurrences of types bound locally (aka `m`).  But if I am really the only one to think that’s important, I am willing to concede and silently hope for a told-you-so moment a few years down :-)

**Comment by simonpj** [comment]
> I don’t think it’s just tidying up: This is where we are introducing occurrences of types bound at the top-level (aka Maybe) in terms for the first time, and that’s when I'd say we have to think about occurrences of types bound locally (aka m).

Not at all!  This scoping stuff is all in #448, and has been in GHC Haskell for way longer e.g. `f (x :: Maybe a) = ...`

> Now that we have established that it’s necessary or desireable to desirable to have an occurrence Maybe in

I don't get this.  I think we established that it is *not* necessary, but I would argue that it is desirable.
```
withF :: forall (m : Type -> Type). (forall a. m a -> Bool) -> …
withF isJust = …
  where
    f xs = ..(isJust x)... case xs of (:) @(m a) x _ -> ....
```
Yes, as I mentioned earlier, it would be nice to have a way to distinguish *binders* from *occurrences* in patterns. We don't have a way to do that right now.  But if you succeed in making the case, we could add it later, so that you could write `@(.m a)` or something.  

But that is for the future.  I'm still terribly hazy about what change, if any, you are proposing for this MR.  (Or I suppose for #448.)

**Comment by nomeata** [comment]
I admit I got lost which proposal is is setting up a an overall goal, and which proposals just fleshed out something that is already decided. But yes, I maybe should have complained about #448 back then. I’ll be quiet now.

**Comment by int-index** [comment]
> This fails with
> 
> ```
> given.hs:20:13: error: [GHC-69710]
>     Type variable ‘a’ is already in scope.
>     Type applications in patterns must bind fresh variables, without shadowing.
> ```
> 
> So if I switch from `forall a.` to `forall a ->` and omit the `@` – will this work the same with the present proposal? Presumably not, but what about the VOP then?

That was the old implementation. #448 says it's a valid binding and GHC HEAD should accept it.

File: proposals/0281-visible-forall.rst (hunk 2 above)
Comment (int-index) [reply proposals/0281-visible-forall.rst:798]: I agree it's oddly placed. It's doesn't even need to be part of the spec, as it's just an observation. I'll move it to Effect and Interactions.

File: proposals/0281-visible-forall.rst (hunk 1 above)
Comment (int-index) [reply proposals/0281-visible-forall.rst:759]: I expect each point to stand alone, so if it starts with "in the same way", it can't be a separate point. I'll just add this sentence to point 8.

File: proposals/0281-visible-forall.rst [hunk 4]
```diff
@@ -570,6 +614,12 @@ information has been fully determined:
   1. Namespace selection syntax (under ``-XDataKinds``)
   2. Name quotation syntax (under ``-XTemplateHaskell``)
 
+* The meaning of ``->`` has been determined. It can stand for one of the following:
+
+  1. Part of lambda, case-of, or multi-if syntax, as in ``\x -> ...``
+  2. The function type constructor ``(->)``, as in ``Int -> Bool``
+  3. A view pattern ``f (e -> p) = ...`` (under ``-XViewPatterns``)
+
```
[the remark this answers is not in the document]
int-index: > We need to say "Extend the syntax of both terms and patterns with several constructs that..."

The dichotomies I use throughout the proposal is (term-vs-type)-syntax and expression-vs-pattern. So I can't say "terms and patterns", that'd be a category error.

I edited the sentence to say "extend the term syntax (expressions and patterns)".

> Then the little table with "proposed" vs "as opposed to" needs to be duplicated for patterns. Or maybe add a third column.

I don't think that's necessary. GHC uses the same grammar for expressions and patterns. We can pretend otherwise when it's convenient, but here it's more convenient to acknowledge this and avoid the duplication of nonterminals.

I added the following sentences: "Patterns reuse the expression grammar in GHC. The new "terms-in-types" productions are also valid in patterns."

> Then new line 653 "This is a necessity..." What is "this"? Please clarify.

Clarified as follows: "In `e1 -> e2` and `e1 => e2`, using the expression grammar for the LHS is a necessity to avoid parsing conflicts"

> Then the following (i), (ii), (iii) is a mystery. We have extended the syntax of terms. Very well. If we later want to re-interpret that term as a type, we have work to do (Section 7.6). But picking out three particular cases here seems quite confusing. What is special about them? It's all handled in 7.6 isn't it?

Nothing special.  Moved these examples to "Effect and Interactions".

> Can we just delete this cryptic utterance, whose meaning is unclear, and instead spell it out where needed?

OK, done.

> What is this "interpret" business? 

Well, the production is always there. The GHC parser accepts `p1 -> p2` today! It just interprets it as a view pattern, not as a function arrow. I'm using "interprets" not in the sense of an interpreter or compiler, but in the sense of assigning meaning to syntax (in the same way a person interprets what they read or hear). 

Anyway, I rephrased it in a way that better aligns with your suggestion.

> `case ... of x -> (y -> z)` cannot possibly make sense, right?

Not with this proposal alone, no. But it will make sense if we allow to return types from functions (it's far-fetched but I like to think it'll be possible in DH)

```haskell
f :: Bool -> Type
f b = case b of
  True  -> Int
  False -> (Int -> String)
```

It's not something I'm proposing here, but it's in the back of my mind when I'm designing the syntax. I also expect to be able to match on types

```haskell
g :: Type -> Bool
g (Int -> String) = True
g _ = False
```

Both examples above are already possible with `TypeRep`, of course, but the syntax is nowhere as nice.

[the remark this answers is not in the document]
simonpj: > I don't think that's necessary. GHC uses the same grammar for expressions and patterns. We can pretend otherwise when it's convenient, but here it's more convenient to acknowledge this and avoid the duplication of nonterminals.

But GHC's implementation is, well, just an implementation. The Haskell Report uses separate syntax for each.  I really think it would be better to be explicit, especially since (p1 -> p2) is treated differently to (e1 -> e2).

I really think it'd be better to be explicit about this.  It's a very confusing area, as we have discovered through our extended interaction.  The more explicit the better.  A bit of duplication does no harm.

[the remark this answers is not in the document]
int-index: I added a proper BNF. Having done that, I agree it's an improvement.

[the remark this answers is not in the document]
simonpj: Thanks! 

Can you add a pointer (in a comment) to point (4) to the production for `(->)` for patterns?

It looks as if these are these give the *complete* syntax for `exp` and `pat`.  Is that right?  Doesn't seem plausible. What did the previous syntax for `exp` and `pat` look like?

proposals/0281-visible-forall.rst
```diff
@@ -587,33 +637,43 @@ Syntax
...
+     pat ::=
+       | infixpat
+       | infixpat      '->'  pat
+       | infixpat mult '->'  pat
+       | infixpat      '->.' pat
+       | infixpat      '=>'  pat
+       | 'forall' tv_bndrs '.'  pat
+       | 'forall' tv_bndrs '->' pat
+
+   Note that the constituents of terms-in-types use the term syntax::
```
simonpj, line 661: I think you mean "for example" here, correct?  If so say so.

proposals/0281-visible-forall.rst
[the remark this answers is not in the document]
int-index: I pushed another update. Took me longer than expected because I've discovered a precedence issue and had to come up with a mitigation plan

```
f (e -> p :: t)   = ...     -- user-written function LHS
f (e -> (p :: t)) = ...     -- old parse
f ((e -> p) :: t) = ...     -- new parse
```

See the new commit.

int-index:
@simonpj Summary of clarifications

1. **Clarification:** `forall a ->` is also allowed in data constructors (but only in GADT syntax)
2. **Clarification:** Patterns undergo T2T just like expressions
3. **Clarification:** Patterns are type-checked using the rules for `@t`-abstractions from [#448](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0448-type-variable-scoping.rst), modulo visibility
4. **Clarification:** `ViewPatterns` and types-in-terms are in conflict syntactically (is `Int -> Bool` a view pattern?)
5. **Clarification:** Types-in-terms have an actual BNF now, based on Haskell2010+PatternSignatures+ViewPatterns

Summary of changes
1. **Change:** Use `sigma` instead of `rho` in the typing rules
2. **Change:** Drop the binder restriction, i.e. given `MkT :: forall a -> blah` and `f (MkT pat) = ...`, the `pat` is no longer limited to a variable or wildcard
3. **Change:** A new warning `-Wview-pattern-signatures`
4. **Change:** Check for `RequiredTypeArguments` when a data constructor is defined, not when it's used


Unless you have further comments, I intend to do an impact analysis on Stackage, and maybe strengthen the amendment by introducing a breaking change to the precedence of `e -> p :: t` (see the section on `-Wview-pattern-signatures`). I will submit to the committee after that.

simonpj:
Great summary.   I agree with your plan.

I don't understand the problem with view patterns.  What happens today?  And why does it become non-viable with `RequiredTypeArguments`?  Perhaps you point is that:
* Today, `p1 -> p2 :: ty` parses as `p1 -> (p2 :: ty)`, in both patterns and expressions.  The latter because patterns are parsed as expressions in GHC, and then converted to patterns.  (Actually that is not true.. there is the subtle machinery of `DisambECP` which I forget every time: see `Note [Ambiguous syntactic categories]`.)
* If we want to convert a pattern to a type, it would be funny if `t1 -> t2 :: t3` thereby parsed as `t1 -> (t2 :: t3)`.
* We could rejig at that point, couldn't we?  As part of P2T.

I don't understand your proposed solution. You talk of emitting a warning and rejigging. But I don't know where the rejigging place; nor what happens in P2T.

Tricky stuff.

int-index:
> I don't understand the problem with view patterns. What happens today? And why does it become non-viable with `RequiredTypeArguments`?

Today the parser does not support "terms-in-types", I have not implemented that part of the proposal yet. But the goal is to accept the following:

```
f :: forall a -> ...   -- f expects a required type argument

x = f (Int -> Bool :: Type)
```

**Question 1:** is this `Int -> (Bool :: Type)` or `(Int -> Bool) :: Type`? The BNF for expressions should tell us that.

Now let's think about view patterns. 

```
f (isJust -> True :: Bool) = ...
```

**Question 2:** is this `isJust -> (True :: Bool)` or `(isJust -> True) :: Bool`? The BNF for patterns should tell us that.

And the problem is that if our answers to these questions differ, we need to do some fancy footwork in the parser. I'd prefer to avoid that.

Now, do we have that problem? Turns out we do. We want `(a -> b) :: c` for types-in-terms but the `ViewPatterns` of today use `a -> (b :: c)`.

The parser can deal with that but I find the entire situation rather unfortunate, so I propose to emit a warning and possibly remove the fancy footwork in the future, in favor of the newly proposed `(a -> b) :: c` precedence.

simonpj:
> DisambECP takes place during parsing, as soon as we discover if we have a HsExpr or Pat. And we can rejig there.

Hmm.  So you are saying that the parser behaves *as if* we has a parser for expressions and another parser for patterns.

Your Q1 and Q2.  I think you are saying:
* In patterns we want `(e1 -> p2 :: ty)` to parse as a view pattern `(e1->(p2 :: ty))`, because that is what happens today.
* In expressions we want `(e1 -> e2 :: ty)` to parse as `((e1->e2) :: ty)`, because that would make sense for a Required type argument.

But since the parser behaves as if we had a parser for expressions and another parser for patterns, that should be easy to do, shouldn't it?

So perhaps you are saying that although it's easy to do, it is tiresome, so you want to change the pattern syntax precedence, with a warning to encourage you to add parens.

I'm guessing a bit as you can see.

int-index:
I have performed the impact analysis and it turns out that only 2 out of 3318 packages are affected by the change to the precedence of `e -> p :: t`.

Based on this result, I decided to include this breaking change in the amendment.

Now I'd like to submit to the committee, @nomeata

aspiwack:
I've submitted my recommendation to the committee.

proposals/0281-visible-forall.rst
```diff
@@ -1143,6 +1327,69 @@ Effect and Interactions
...
+       f (#)  = ...   -- binding a local variable (#)
+
+       -- Type syntax
+       f (type (:#)) = ...   -- usage of type constructor (:#)
+       f (type (#))  = ...   -- usage of type constructor (#)
+
+     To be consistent with terms, ``type (#)`` would have to be a binding of a
+     local type variable named ``(#)``.
+
+  Accounting for these discrepancies is a non-goal for the T2T transformation
```
nomeata, line 1381: Just checking: Do I understand it correctly that this means that users who need to refer to a type called (`#`) they can just do that in implicit type  applications (`MkF @(#)`), but cannot do it in required type application applications (`MkF (#)`)? But that’s ok because they can work-around it with an explicit namespace, like `MkF (type (#))`?

proposals/0281-visible-forall.rst
```diff
@@ -623,8 +747,26 @@ Syntax
    This change applies to ``∀`` (the ``UnicodeSyntax`` rendition of ``forall``)
    as well.
 
-4. When ``ViewPatterns`` are enabled, interpret ``f (a -> b) = ...``
-   as a view pattern, otherwise as ``f ((->) a b) = ...``.
+4. Adding ``p1 -> p2`` to the syntax of patterns (see the BNF of "types-in-terms"
+   above) is incompatible with ``ViewPatterns``. When both ``RequiredTypeArguments``
+   and ``ViewPatterns`` are enabled, the conflict is resolved in favor of ``ViewPatterns``.
+   The programmer can write ``type (t1 -> t2)`` or ``(->) t1 t2`` instead.
```
goldfirere, line 753: I think I'd favor making this treatment not dependent on `ViewPatterns`. As written, this proposal makes `ViewPatterns` fork-like, in that enabling `ViewPatterns` rejects some previously-valid programs. It is easy to avoid this, by unconditionally refusing to accept `p1 -> p2` as a function type pattern. It's a wart, yes, but I'd rather have a consistent wart than an inconsistent one.

(Really, the mistake here was allowing `ViewPatterns` not to have a keyword herald. This decision could only have happened because the implementation uses the same grammar for expressions and patterns. Just reading the Report, it would look impossible to parse a view pattern as implemented today, and a keyword herald would have been used. This might be a small lesson for the future.)

simonpj: Just to be clear, consider
```
data T a where
  K :: forall a -> a -> T a

f (K (Int->Int) x) = ...
```

* Richard proposes that the type pattern `(Int -> Int)` is rejected regardless of `ViewPatterns` because `Int -> Int` could be a view pattern.  
* Vlad suggests accepting it iff `ViewPattens` is off.

I don't feel strongly either way, but I'd be content to have it always-rejected.  You'd have to write
```
f (K ((->) Int Int) x) = ...
```
instead, which works regardless of `ViewPatterns`.

int-index: > this proposal makes `ViewPatterns` fork-like, in that enabling `ViewPatterns` rejects some previously-valid programs

True, and I agree it's a problem. However, we only have that problem because `ViewPatterns` stole syntax that it had no business stealing, and we're now dealing with the consequences. So the question is: what's the best way to solve this problem?

Instead of limiting `RequiredTypeArguments` as you suggest, I'd rather have an alternative syntax for `ViewPatterns` that does not have this issue in the first place. There's already a draft (not mine) at https://gitlab.haskell.org/ghc/ghc/-/wikis/view-patterns-alternative, and I think it's quite promising.

phadej: > problem because ViewPatterns

`ViewPatterns` has been around for over a decade (it is in GHC-7.0, I don't know when it was added). I'd consider it as one of the most stable extensions GHC has. The existence of `ViewPatternsAlternative` (wiki page created 12 years ago as well), and as far as I can tell zero progress on it is the same as *it doesn't exist*. *Someone* has to put work into advancing it first. Until then `ViewPatterns` should continue to work. `PatternSynonyms` usage relies on `ViewPatterns`.

I actually hope that GHC steering committee finds a way to avoid stalemates. alternative view patterns wiki page  has existed for 12 years, but there are no visible progress. But just because wiki page exists, `ViewPatterns` is not added to GHC2021 nor GHC2024, because "something might happen", but evidence says contrary. (FWIW, similarly "Safe Haskell" might improve, and not be in conflict with GHC202x, but so far there have been zero progress as well). That is uncomfortable situation. cc @adamgundry 
 @nomeata.

int-index: > Someone has to put work into advancing it first.

This is exactly what's happening here. I'm trying to advance the "view patterns alternative" by providing a motivation for it. The syntax chosen for `ViewPatterns` has led to obstacles for `RequiredTypeArguments`, and I'm not willing to compromise `RequiredTypeArguments` because of that. This motivates the "view patterns alternative".

Motivation precedes implementation.

> Until then `ViewPatterns` should continue to work. `PatternSynonyms` usage relies on `ViewPatterns`.

Quite right, and no one is proposing to break `ViewPatterns` until we have the alternative. Is that your concern? You'll be happy to find that the amendment already says "the conflict is resolved in favor of `ViewPatterns`".

Ericson2314: Yes, the question here is not of breakage, but of "fork-like-ness" vs "accumulated defects".

I do think "fork-like-ness" is preferable too. As discussed elsewhere, the thing people don't like is breakage when their software is updated. Supporting multiple "contradictory" things for an extended period of time however is quite popular: maintainers of old code and writers of new code both get what they want.

goldfirere: It seems to me there's a straightforward answer here: don't accept `Int -> Int` as a type pattern. If we end up doing an alternative syntax for view patterns, we can always add `Int -> Int` as a type pattern later. But if we accept `Int -> Int` as a type pattern now, we'll never be able to remove that, even if view patterns never get revised. I agree that this would be allowing a mistake around view patterns to make a new feature worse -- but I also think it's better to get 98% of the new feature in rather than wait to figure out view patterns before making any progress.

tomjaguarpaw: I strongly agree with @goldfirere, regardless of the specific issue at hand. As a general principle, if there is a choice between options A and B, both improvements, and A is significantly easier to enact than B yet doesn't make it more difficult to enact B later, then A should be chosen. It's important to unlock value as quickly as possible, even (or maybe especially) partial value!

int-index: @goldfirere I'm not looking forward to dropping this from the specification only to resubmit it again later. Would it be fine if I said that this change is conditional on having an alternative to view patterns?

int-index: I pushed a change, adding this sentence

> When only ``RequiredTypeArguments`` is enabled, the conflict is not resolved either way. It is reported to the user in an error message. This can be reconsidered if an alternative syntax for view patterns is implemented in GHC.

goldfirere: Yes I'm quite happy with that.

aspiwack:
The committee has accepted this proposal amendment. Thanks @int-index for the work that went into it, and for your patience.

@nomeata I'll leave it to you to merge, as usual.

PR #631, opened 2024-01-18
Proposal: Meaningful main return types

shlevy:
[Rendered](https://github.com/shlevy/ghc-proposals/blob/io-exitcode/proposals/0631-main-return-types.rst)

### DIFF
--- a/proposals/0631-main-return-types.rst
+++ b/proposals/0631-main-return-types.rst
@@ -1,0 +1,159 @@
+Set program exit code by main return type
+==============
+
+.. author:: Shea Levy
+.. date-accepted::
+.. ticket-url::
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/631>`_.
+.. sectnum::
+.. contents::
+
+To reduce surprise and match common programming language behavior,
+this proposal would allow programs to signal their exit code
+based on the return value from ``main``.
+
+Motivation
+----------
+Many languages, perhaps most notably ``C``, allow specifying a program's exit
+code by returning an appropriate value from ``main``. ``Haskell98`` *does*
+allow a ``main`` which yields values of any type, but surprisingly it does not
+do anything with the result: any successful execution will result in a successful
+program exit, even if ``main`` yields ``1`` or ``ExitFailure 1``.
+
+Before this proposal, the following program would exit with code ``0`` even
+when there are no results:
+
+::
+
+ main :: IO ExitCode
+ main = do
+   results <- doSomeWork
+   case results of
+     [] -> pure (ExitFailure 1)
+     _ -> print results >> pure ExitSuccess
+
+With this proposal in place, the program would exit with code ``1`` in that
+case.
+
+This proposal follows `Rust <https://doc.rust-lang.org/std/process/trait.Termination.html>`_
+in using a typeclass for return types which can signal exit codes.
+
+This is a minor quality of life improvement, but one that I've seen hit
+newcomers now and then and matches reasonable expectations of program
+behavior.
+
+Proposed Change Specification
+-----------------------------
+Add a new typeclass, ``System.Exit.Termination``, with a function
+``report :: Termination a => a -> ExitCode``. If, for a given program,
+``(main >>= exitWith . report) :: IO ()`` type-checks, then
+the resulting program will behave as if that had been written for ``main``
+instead.
+
+Proposed Library Change Specification
+-------------------------------------
+
+Add the ``Termination`` typeclass to ``System.Exit``, with the
+``report`` function/method, as described above.
+
+Add a ``Termination`` instance for ``ExitCode``.
+
+Examples
+--------
+
+::
+
+ data ExitReason
+   = AllsWell
+   | NoResults
+
+ instance Termination ExitReason where
+   report AllsWell = ExitSuccess
+   report NoResults = ExitFailure 1
+
+ main = do
+   results <- doSomeWork
+   case results of
+     [] -> pure NoResults
+     _ -> print results >> pure AllsWell
+
+With this proposal in place, the program would exit with code ``1`` if
+``doSomeWork`` doesn't yield results.
+
+Effect and Interactions
+-----------------------
+This change will allow users to write ``main`` programs which yield an
+``ExitCode`` and have that code respected.
+
+No known interactions.
+
+Costs and Drawbacks
+-------------------
+Development seems likely to be minimal, and maintenance basically
+non-existent. This will have no *negative* impact on novice users
+learning the language.
+
+
+Backward Compatibility
+----------------------
+Technically, this will cause a change in behavior in extremely rare cases
+(this is level ``1`` on the breakage scale): If a program currently has
+``main :: IO ExitCode`` and some branch currently yields ``ExitFailure n``,
+the program will currently exit with ``0`` on that branch. After this proposal,
+it will exit with ``n``. But the latter is almost certainly the desired behavior for
+writing code like that, and I'd be surprised if it exists in any real program.
+
+Because this behavior change, if it ever actually matters, is likely in the direction
+of *improving* the program behavior, it's not clear that there are any real costs
+to this "breakage". Still, we could add a warning to the ``Termination ExitCode``
+instance and encourage users to use a custom type, or see the alternatives
+for `an option <#no-exitcode-instance>`_ with no backwards incompatibility.
+
+Alternatives
+------------
+
+Retain the status quo
+^^^^^^^^^^^^^^^^^^^^^
+
+While the current behavior is functional and any surprise is likely to be
+caught early in testing, this change will make learning Haskell a bit
+easier and arguably makes for ``main`` functions which better match
+the Haskell ethos of well-typed structured interfaces.
+
+Do it just for ExitCode
+^^^^^^^^^^^^^^^^^^^^^^^
+
+Rather than a typeclass, we could just have ``main :: IO ExitCode`` be
+respected. Using the typeclass allows using domain-specific types
+to exit codes as in `the example <#Examples>`_, letting the user
+give semantic meaning to the exit status in the typical Haskell
+way.
+
+No ExitCode instance
+^^^^^^^^^^^^^^^^^^^^^
+
+To preserve full backwards compatibility and encourage custom domain-specific
+types, we could avoid having a ``Termination ExitCode`` instance, or have one
+which is ``const ExitSuccess`` with a warning emitted if it's ever used. Having
+an ``ExitCode`` instance reduces surprise and reduces overhead for simple
+programs.
+
+Add Int instance
+^^^^^^^^^^^^^^^^
+
+Some users may expect ``main :: IO Int`` to work, and we could add a
+``Termination Int`` instance to satisfy that. But this is much more likely
+to cause behavior changes in real programs, and perpetuates a practice of
+semantically loose types.
+
+Unresolved Questions
+--------------------
+The name of ``Termination`` and ``report`` are copied from ``Rust``, but
+perhaps we want a different paint color.
+
+
+Implementation Plan
+-------------------
+I'd (@shlevy) be willing to implement this if accepted.
--- a/proposals/0631-main-return-types.rst
+++ b/proposals/0631-main-return-types.rst
@@ -2,11 +2,11 @@ Meaningful main return types
 =========================================
 
 .. author:: Shea Levy
-.. date-accepted::
-.. ticket-url::
+.. date-accepted:: 2024-05-21
+.. ticket-url:: https://gitlab.haskell.org/ghc/ghc/-/issues/24848
 .. implemented::
 .. highlight:: haskell
-.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/631>`_.
+.. header:: This proposal was `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/631>`_.
 .. sectnum::
 .. contents::
 
--- a/proposals/0631-main-return-types.rst
+++ b/proposals/0631-main-return-types.rst
@@ -53,7 +53,7 @@ instead. Otherwise, the resulting program will behave as if ``main >> exitWith E
 had been written, but the compiler will emit a warning.
 
 Proposed Library Change Specification
--------------------------------------
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Add a new module to ``ghc-experimental``:
 
--- a/proposals/0631-main-return-types.rst
+++ b/proposals/0631-main-return-types.rst
@@ -54,6 +54,7 @@ Add a new module to ``ghc-experimental``:
  module GHC.IO.Exit where
 
  import GHC.IO.Exception (ExitCode (..))
+ import Data.Void (Void, absurd)
 
  class Termination e where
    report :: e -> ExitCode
@@ -61,12 +62,15 @@ Add a new module to ``ghc-experimental``:
  instance Termination ExitCode where
    report = id
 
- -- Not strictly required, but users might expect
- -- 'Termination' to be the class of /allowed/
- -- main return types (and perhaps some day it could
- -- become that).
+ -- The remaining instances are not strictly required,
+ -- but users might expect 'Termination' to be the class of
+ -- /allowed/ main return types (and perhaps some day it
+ -- could become that).
  instance Termination () where
    report = const ExitSuccess
+
+ instance Termination Void where
+   report = absurd
 
 If, for a given program, ``(main >>= exitWith . report) :: IO ()`` type-checks,
 then the resulting program will behave as if that had been written for ``main``
--- a/proposals/0631-main-return-types.rst
+++ b/proposals/0631-main-return-types.rst
@@ -1,5 +1,5 @@
 Set program exit code by main return type
-==============
+=========================================
 
 .. author:: Shea Levy
 .. date-accepted::
@@ -154,7 +154,7 @@ give semantic meaning to the exit status in the typical Haskell
 way.
 
 No ExitCode instance
-^^^^^^^^^^^^^^^^^^^^^
+^^^^^^^^^^^^^^^^^^^^
 
 To preserve full backwards compatibility and encourage custom domain-specific
 types, we could avoid having a ``Termination ExitCode`` instance, or have one
@@ -175,6 +175,29 @@ semantically loose types.
  instance Termination Int where
    report 0 = ExitSuccess
    report n = ExitFailure n
+
+Restrict main to IO ()
+^^^^^^^^^^^^^^^^^^^^^^
+
+If ``main`` *had* to be ``IO ()``, this would also reduce surprise,
+or at least make it apparent at compile time. This is
+backwards-incompatible, but would likely not break that many
+programs and the fix would be straightforward.
+
+This alternative would break more programs than the proposal,
+and would miss out on the added benefit of program behavior
+being specified by more normal Haskell control flow. Also,
+this appears to have been the behavior in Haskell 1.4, and
+presumably the Haskell 98 authors changed this for a reason.
+
+Restrict main to IO Void
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+This would force programmers to be explicit about exit codes,
+and indicate that program exit is something different than
+normal ``IO`` completion. It might have been a reasonable
+choice when Haskell was new, but as it would break almost
+every program out there today it's not worth the churn.
 
 Unresolved Questions
 --------------------
--- a/proposals/0631-main-return-types.rst
+++ b/proposals/0631-main-return-types.rst
@@ -40,12 +40,26 @@ When ``-XMeaningfulMainReturn``, GHC emits an error if ``main`` has a type that 
 ``IO ()``, ``IO Void``, or ``IO ExitCode``. If ``main``'s type unifies with ``IO ExitCode``, the
 resulting program behaves as if its entrypoint were ``realMain = main >>= exitWith``.
 
-``-XMeaningfulMainReturn`` will become the default in the next GHC language edition.
+``-XMeaningfulMainReturn`` is intended to become the default in the next GHC language edition.
 
 Examples
 --------
+Before this proposal, the following program would exit with code ``0`` even
+when there are no results:
 
 ::
+
+ main :: IO ExitCode
+ main = do
+   results <- doSomeWork
+   case results of
+     [] -> pure (ExitFailure 1)
+     _ -> print results >> pure ExitSuccess
+
+With this proposal in place, the following examples would behave as indicated:
+
+::
+
    {-# LANGUAGE NoMeaningfulMainReturn #-}
 
    main :: IO ()
@@ -54,6 +68,7 @@ Examples
 This is accepted as it is now.
 
 ::
+
    {-# LANGUAGE NoMeaningfulMainReturn #-}
 
    main = main
@@ -61,6 +76,7 @@ This is accepted as it is now.
 This unifies with ``IO ()``, so it is accepted without warning.
 
 ::
+
    {-# LANGUAGE NoMeaningfulMainReturn #-}
 
    main :: IO Int
@@ -69,6 +85,7 @@ This unifies with ``IO ()``, so it is accepted without warning.
 Warning emitted, program exit code is ``0``, not ``1``
 
 ::
+
    {-# LANGUAGE MeaningfulMainReturn #-}
 
    main :: IO Int
@@ -77,6 +94,7 @@ Warning emitted, program exit code is ``0``, not ``1``
 Compile error, ``Int`` is potentially ambiguous.
 
 ::
+
    {-# LANGUAGE MeaningfulMainReturn #-}
 
    main :: IO Void
@@ -85,6 +103,7 @@ Compile error, ``Int`` is potentially ambiguous.
 Successful compilation, program exit code is ``0``
 
 ::
+
    {-# LANGUAGE MeaningfulMainReturn #-}
 
    main :: IO ExitCode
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
I sympathize with the desire for something like this. On the other hand it's very easy to simulate in a library; just make sure you always use `main = mainWithTermination $ do ...` (see below)! That's not really satisfying though, because the aim is to avoid the surprise when you *don't* know how `main` uses its result (i.e., it doesn't). But would you at least be willing to mention this alternative in the proposal?

----

```.hs
import System.Exit

class Termination a where
  report :: a -> ExitCode

mainWithTermination :: Termination a => IO a -> IO ()
mainWithTermination realMain = exitWith =<< (report <$> realMain)
```

aspiwack:
Other potential alternatives that address part of the motivation by reducing the surprising behaviours

- Restrict `main` to always be of type `IO ()`, other types are an error. Otherwise leave the current behaviour unchanged
  - This isn't backward compatible, but it probably affects very few programs (though more than your proposal)
- Restrict `main` to always be of type `IO Void`, other types are an error.
  - Then we're always forced to terminate the program with `exitWith`, forcing the programmer to be explicit about their exit codes
  - It's incompatible with at least 99.9% of Haskell executables out there, so it's probably not particularly feasible. But I propose it for completeness as it could have been an absolutely reasonable design when Haskell was still new.

---

Some comment on the form:

Please be precise in the definition of the `Termination` type class (right now you say that a function exist, but you really need to list the methods).

As a complement to @nfrisby 's comment above: please list, in the specification section, all the instances of `Termination` that you expect should be exported by default.

Do be aware that in the current scheme, [GHC proposals are not allowed to modify `base`](https://github.com/haskellfoundation/tech-proposals/blob/main/proposals/accepted/051-ghc-base-libraries.rst). You'll need to find a place (either in an existing module or a new one) in `ghc-experimental`. I'm not actually sure whether we have naming conventions there quite yet.

shlevy:
> I sympathize with the desire for something like this. On the other hand it's very easy to simulate in a library; just make sure you always use `main = mainWithTermination $ do ...` (see below)! That's not really satisfying though, because the aim is to avoid the surprise when you _don't_ know how `main` uses its result (i.e., it doesn't). But would you at least be willing to mention this alternative in the proposal?
> 
> ```haskell
> import System.Exit
> 
> class Termination a where
>   report :: a -> ExitCode
> 
> mainWithTermination :: Termination a => IO a -> IO ()
> mainWithTermination realMain = exitWith =<< (report <$> realMain)
> ```

Addressed in 3bce7f3448a513e4db40a5beb4c38bc39da1097d

shlevy:
> > Add a `Termination` instance for `ExitCode`.
> 
> Which other instances of `Termination` do you think should be defined out of the box? You did mention that an `Int` instance has pros and cons. I would also expect at least `()`, for example :)

Addressed in 402ec5f94396b0d2566275e0cd54e4719956e5da

shlevy:
> Other potential alternatives that address part of the motivation by reducing the surprising behaviours
> 
>     * Restrict `main` to always be of type `IO ()`, other types are an error. Otherwise leave the current behaviour unchanged
>       
>       * This isn't backward compatible, but it probably affects very few programs (though more than your proposal)
> 
>     * Restrict `main` to always be of type `IO Void`, other types are an error.
>       
>       * Then we're always forced to terminate the program with `exitWith`, forcing the programmer to be explicit about their exit codes
>       * It's incompatible with at least 99.9% of Haskell executables out there, so it's probably not particularly feasible. But I propose it for completeness as it could have been an absolutely reasonable design when Haskell was still new.
> 

Addressed in 4aa742bdfafe9514337758a5c37338222a74c1e6

> 
> Some comment on the form:
> 
> Please be precise in the definition of the `Termination` type class (right now you say that a function exist, but you really need to list the methods).
> 
> As a complement to @nfrisby 's comment above: please list, in the specification section, all the instances of `Termination` that you expect should be exported by default.
> 
> Do be aware that in the current scheme, [GHC proposals are not allowed to modify `base`](https://github.com/haskellfoundation/tech-proposals/blob/main/proposals/accepted/051-ghc-base-libraries.rst). You'll need to find a place (either in an existing module or a new one) in `ghc-experimental`. I'm not actually sure whether we have naming conventions there quite yet.

Addressed in 402ec5f94396b0d2566275e0cd54e4719956e5da

Ericson2314:
I had something to do with Rust supporting this :) https://github.com/rust-lang/rfcs/issues/1176 (though https://github.com/rust-lang/rfcs/pull/1937 actually got the job done). I would support it in Haskell for the same reasons!

shlevy:
What is the process at this point?

Ericson2314:
You can ask the secretary (currently @nomeata but about to change) to submit to the committee. (It's common to forget this step for the initial submission, however. I wonder if that is a sign it should be different..)

nomeata:
I’ll defer to @adamgundry already :-)

shlevy:
OK, @adamgundry can we submit to the committee?

simonpj:
I would love to hear from users about this.   To me it seems a bit indirect.  It amounts to this:

* To make your program return an appropriate exit-code
* ..make `main` return `ExitCode`
* ...or, if you prefer
  * make`main` return a value of your own type, say `T`
  * make `T` an instance of `Termination`, where `report` maps `T` to `ExitCode`

But instead of saying `retun (ExitFailure 1)` , why not just say `exitWith (ExitFailure 1)`?  That seems more direct.

And it's even more indirect to say `return BadThings` and then given an instance `report BadThings = ExitFailure 1`.

To me this seems like adding a bit more complexity without adding any new modularity or expressiveness.  But I could be persuaded by user testimony and/or compelling examples.

If this proposal is eventually adopted, please could we not call the class `Termination`?  For me that connotes some kind of check that a function does not diverge.  Perhaps `ExitStatus`?

sullyj3:
One nice thing about avoiding `exitWith` in favour of returning is that you're guaranteed that there's no code after an exit call which might mistakenly not be executed

```
main :: IO ()
main = do
  exitSuccess
  putStrLn "Wait, I wanted this to be printed!" -- Compiles successfully
```

```
main :: IO ExitCode
main = do
  pure ExitSuccess
  putStrLn "Wait, I wanted this to be printed!" -- Type error
```

Of course in simple examples like this you can see that `exitSuccess` should go at the bottom, but you can imagine in code with a lot of branching it could be more difficult to spot. Refactors would need to be responsible for keeping it so, and they could easily miss it.

tomjaguarpaw:
> One nice thing about avoiding `exitWith` in favour of returning is that you're guaranteed that there's no code after an exit call which might mistakenly not be executed

That issue arises in myriad other cases, for example `throw`, `throwIO`, `undefined`, `error`, ... so trying to improve it in one single case seems futile.

shlevy:
@simonpj What are your thoughts on the alternative of just requiring `main :: IO ()`? I see the concern about indirectness, but the counter is that if `main` *can* yield a value users will expect that to mean something.

simonpj:
> @simonpj What are your thoughts on the alternative of just requiring main :: IO ()? I see the concern about indirectness, but the counter is that if main can yield a value users will expect that to mean something.

I'm not a user so my opinion should not count for much.  If our users want to reject programs with `main :: IO Int`, say, that's fine with me.  Maybe we should have insisted on `IO ()` from day 1.

But making that change will break people's code, and then they'll complain, and we'll need deprecations and flags to control it.   All that takes time and attention and implementation effort, all of which is in limited supply.   Is it important enough?  To me the pain isn't justified by the gain.  But this is not a strongly held view.

shlevy:
> I would love to hear from users about this.

Is there a good mechanism to solicit more input beyond just opening the proposal?

tomjaguarpaw:
I'd post the proposal to Discourse, Reddit and Twitter.

geekosaur:
I was just thinking the other day that it was surprising that nobody had proposed at least `main :: IO ()` yet, with the recent trajectory of Haskell.

My recollection re Haskell98 is that there was an intent to allow programs that looped by reinvoking `main`, but I'm not thinking of many uses of a `main` that produces a value that don't have problems on the initial invocation.

shlevy:
I've requested further feedback https://discourse.haskell.org/t/feedback-requested-simple-proposal-to-set-program-exit-code-by-mains-value/8687

proposals/0631-main-return-types.rst
```diff
@@ -0,0 +1,214 @@
...
+This is a minor quality of life improvement, but one that I've seen hit
+newcomers now and then and matches reasonable expectations of program
+behavior.
+
+Proposed Change Specification
+-----------------------------
+
+If, for a given program, ``(main >>= exitWith . GHC.IO.Exit.report) :: IO ()`` type-checks,
+then the resulting program will behave as if that had been written for ``main``
+instead.
```
aspiwack, line 52: What happens if it doesn't typecheck?

proposals/0631-main-return-types.rst
```diff
@@ -0,0 +1,214 @@
...
+and always define ``main = mainWithTermination $ do { ... }``.
+
+Do it just for ExitCode
+^^^^^^^^^^^^^^^^^^^^^^^
+
+Rather than a typeclass, we could just have ``main :: IO ExitCode`` be
+respected. Using the typeclass allows using domain-specific types
+to exit codes as in `the example <#Examples>`_, letting the user
+give semantic meaning to the exit status in the typical Haskell
+way.
```
aspiwack, line 158: I don't understand what this alternative means.

aspiwack (commented):
Shepherdly :wave: 

There are a couple of points below that need to be tightened, and then I'll make my recommendation for the committee.

proposals/0631-main-return-types.rst
shlevy: Is it acceptable to say “otherwise, the program behaves as it would have had this proposal not been written”? I don’t know of a formal specification of `main`’s current behavior, which presumably would have to include what the runtime does when exceptions are thrown etc.?

proposals/0631-main-return-types.rst
shlevy: Fixed.

proposals/0631-main-return-types.rst
aspiwack: It's acceptable. I'm not sure that it's a good idea. Maybe record the option of failing when `IO A` and `A` doesn't have a `Termination` instance as an alternative.

---

From a pure theoretical property, both alternatives can be implemented in the other by means of an overlappable instance:

Old behaviour when no instance:
```haskell
instance {-# OVERLAPPABLE #-} Termination a where
  report _ = ExitSuccess
```

Error when no instance
```haskell
instance {-# OVERLAPPABLE #-} TypeError <some error message> => Termination a where
  report _ = error "unreachable"
```

So the question of which alternative we want is, beyond implementation considerations, largely a question of what we consider the best UX.

shlevy: Addressed

shlevy:
Renamed to `ExitStatus` and `toExitCode`

simonpj:
Cosmetics: could you please put "Proposed library change" under "2. Proposed change specification".  Maybe have two subsections, 2.1 and 2.2.  I like having the entire proposed change in one section.  THanks

shlevy:
Will do, FYI this structure is from the template https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0000-template.rst

sullyj3:
> > One nice thing about avoiding `exitWith` in favour of returning is that you're guaranteed that there's no code after an exit call which might mistakenly not be executed
> 
> That issue arises in myriad other cases, for example `throw`, `throwIO`, `undefined`, `error`, ... so trying to improve it in one single case seems futile.

Those all have in common that they're crashes or failure states. What distinguishes the one single case is that it's a happy path - and it's nice to have a single exit point at least in the happy path.

shlevy:
> Cosmetics: could you please put "Proposed library change" under "2. Proposed change specification". Maybe have two subsections, 2.1 and 2.2. I like having the entire proposed change in one section. THanks

Done

adamgundry:
My impression from the committee discussion so far is that in the long-term, requiring an `ExitStatus` instance (alternative 7.5) is the preferred design. The question is how we get there, given that this is a somewhat breaking change.

How about the following:

1. Introduce a warning by default when `main` has a type other than `IO ()` or `IO Void` (or perhaps a few other types...) and yet the RTS will always return exit code 0. This will alert users that returning any other type is misleading, but won't break anything.

2. Introduce a new off-by-default extension that unconditionally requires an `ExitStatus` instance (and does not emit the warning). We can then plan to enable the extension as part of the next language edition.

This seems to follow our stability principles nicely, because nothing breaks until the user upgrades their language edition. The only question is whether the additional implementation effort is justified.

maralorn:
1. Before we activate this feature for everyone, e.g. before we publish it in a language edition we should probably move the class out of ghc-experimental. That can of course wait a bit.
2. But before we do that:
  * When this feature is in effect how do the ExitStatus instances propagate into `main`?
  * Will we require importing `GHC.IO.Exit`?
  * Will those instances be available from the Prelude? (if yes, always or only when a potential extension is enabled)
  * Can the "magic exitWith call" use ExitStatus and ExitStatus instances without them being in scope in main? (e.g. it can use all ExitStatus instances defined in `GHC.IO.Exit` plus those which the user brought into scope.)

We should consider which solution is best and for which solutions we need to consult the CLC.

maralorn:
I think warning by -Wdefault on non-canonical return types is an excellent idea. We should do this as soon as possible. If we gate the feature behind an extension the warning should hint to that extension.

I lean slightly against making this warning an error even when gated behind an extension. It would probably break some programs that likely didn’t care about their return type but had a random non `IO ()` action at the end of their main. We would need to weigh that when adding an extension to a language edition. This is the long ongoing discussion about which warnings should be -Werror by-default. But unless we have a guideline for when to do that I think this "code smell" is benign enough to let the user decide on their own time.

simonpj:
> 1.    Introduce a warning by default when main has a type other than IO () or IO Void (or perhaps a few other types...) and yet the RTS will always return exit code 0. This will alert users that returning any other type is misleading, but won't break anything.
> 2.  Introduce a new off-by-default extension that unconditionally requires an ExitStatus instance (and does not emit the warning). We can then plan to enable the extension as part of the next language edition.

Would it not be simpler to do (1) but never do (2)?  As @maralorn says, the warning is enough incentive to move.  Simpler: we don't need a new language extension at all.

> Will we require importing GHC.IO.Exit?

Definitely not.  E.g. when you use a list comprehension you are implicitly using `map`, but you don't have to import it.

> When this feature is in effect how do the ExitStatus instances propagate into main?

For the ones defined in GHC.IO.Exit, you get them automatically; for any user-defined ones, the module that defines the instance must be in the transitive closure of your imports.  That is likely true (orphan instances aside) because the return *type* is definitely in that transitive closure. 

Once settled, these points should be spelled out explicitly in the proposal.

tomjaguarpaw:
> Would it not be simpler to do (1) but never do (2)?

Personally I'd prefer the approach of doing 1 (because it addresses a real confusion) and then coming back after some time has passed to reflect on whether we really need 2. The notion of type classes controlling something as simple as an exit code seems like far too much complexity to me.

shlevy:
I agree with the idea of warning (and perhaps eventually erroring?) on non-canonical exit codes (perhaps accepting only `IO ()` and `IO Void`, maybe also for `IO a` or something which would typecheck if given the signature `IO a`).

Is the consensus that, if we have the typeclass-based approach at all, it should be behind an extension? If so, then I think in that case we should require an `ExitStatus` instance when the extension is on. But it sounds like there’s an open question whether we should simply stop at the warning and be done with it. How do we resolve that?

Re @maralorn’s resolution questions, I think the behavior should be basically as if we had `import GHC.IO.Exit ()` at the top of `module Main`, i.e. I agree with @simonpj.

maralorn:
> I agree with the idea of warning (and perhaps eventually erroring?) on non-canonical exit codes (perhaps accepting only `IO ()` and `IO Void`, maybe also for `IO a` or something which would typecheck if given the signature `IO a`).
> 
> Is the consensus that, if we have the typeclass-based approach at all, it should be behind an extension? If so, then I think in that case we should require an `ExitStatus` instance when the extension is on. But it sounds like there’s an open question whether we should simply stop at the warning and be done with it. How do we resolve that?

I see no problem with gating the feature behind an extension. It also has the advantage that we can flag that extension experimental for a while so concerns about the typeclass approach can be watched. Even with the extension I am slightly in favor of only warning on types without the class, with my reasoning given above: It will make it easier to make the extension default on later. I think the warning is more important than the return-type support, but I am generally in favor of something like it.

> Re @maralorn’s resolution questions, I think the behavior should be basically as if we had `import GHC.IO.Exit ()` at the top of `module Main`, i.e. I agree with @simonpj.

Yeah, okay. Since there is no way to use those instances without transitively depending on `GHC.IO.Exit` that import seems unobservable so I guess this does not affect the Prelude or anything concerning the CLC.

simonpj:
> Is the consensus that, if we have the typeclass-based approach at all, it should be behind an extension?

I'm not sure what is gained by having an extension.  Suppose we just 
* have the type class
* instances for () and Void
* Warning if there is no instance; program returns exit code 0

Now all programs behave exactly as today, except in a few (rather rare) case we get a warning.  You can suppress the warning by adding `return ()` at the end, which is fully back-compatible.

So why do we need an extension?

maralorn:
We don’t need to, to prevent breakage. I just thought it is good practice to guard language changes behind an extension, e.g. so that we can change it if problems arise in a testing period. When we are satisfied with the extension, we can subsume the extension in the next language edition and don’t need to worry about it anymore.

That being said I have no strong opinion about this and am fine either way.

adamgundry:
Assuming we give an instance for `ExitCode` (which the proposal currently has) as well as `()` and `Void` then there is a behaviour change here (as stated up front in the Motivation section). In the absence of the `ExitCode` instance I agree this doesn't break backwards compatibility, but it also seems less well motivated to use a typeclass at all.

Perhaps it depends on one's stance about extensions. Given that it makes a change from something explicitly specified in Haskell 2010, I would personally favour having an extension (even perhaps one that is on by default, if the concern is that waiting for a new language edition is too long).

simonpj:
> Assuming we give an instance for ExitCode (which the proposal currently has) as well as () and Void then there is a behaviour change

What is the behaviour change?  

Perhaps it's this: If you return ExitCode (a brand new type that does not exist at the moment) you can get a non-zero return code.  But  an old compiler doesn't understand ExitCode anyway, so it'll reject the new program anyway.

> Given that it makes a change from something explicitly specified in Haskell 2010

What is the change?  Perhaps that you get zero exit code from any type, including new ones?

I don't think it's worth the extra fuss of an extension (documentation, code ...) for the sake of box-ticking.   

Arguing against myself, some existing extensions like `PatternGuards` could simply be on all the time, because they use new syntax.  The only reason for an extension so that you don't accidentally use it.   So `-XPatternGuards` unlocks the new syntax.  I suppose in the same way `-XExitCodes` could enable this type class thing.

But there are other things like name-shadowing that are on all the time; you can tell if you use it accidentally, via a warning flag.

I don't think it matter much either way.

maralorn:
Consider the program:

```
import System.Exit
main = pure ExitFailure
```
This program builds and runs successfully today. With the proposed change it will compile successfully without a warning and then fail at runtime (as was probably the intention of the author, tbf.). Technically speaking that is the worst kind of breakage that we can introduce.

I agree that it doesn’t matter much. But it is an observable behavior change that is not gated behind new syntax.

shlevy:
> > Assuming we give an instance for ExitCode (which the proposal currently has) as well as () and Void then there is a behaviour change
> 
> What is the behaviour change?
> 
> Perhaps it's this: If you return ExitCode (a brand new type that does not exist at the moment) you can get a non-zero return code. But an old compiler doesn't understand ExitCode anyway, so it'll reject the new program anyway.

No, `ExitCode` is not a new type. But as I mention in the proposal, if someone does have an existing `main :: IO ExitCode` they almost certainly _want_ it to be have as this proposal suggests.

shlevy:
I do want to question whether the technical backwards incompatibility here makes sense to worry about. If there are any actual Haskell programs with `main :: IO ExitCode` (and there probably aren’t!), they are almost certainly all buggy with the current behavior. That’s not a GHC bug of course, but I stand by the claim that this is an *improvement* for those users, if they exist. The fact that it’s also in a formal sense a breaking behavior change should not be an obstacle in my view.

simonpj:
> This program builds and runs successfully today. With the proposed change it will compile successfully without a warning and then fail at runtime (as was probably the intention of the author, tbf.). Technically speaking that is the worst kind of breakage that we can introduce.

Ah I had entirely missed the fact that `ExitCode` already exists.  Very well, you are correct.

Thank you.   Now I'm on the fence. I don't have a strong opinion either way.  But now I do agree that it should technically be an extension because of the behaviour change.  Not a big deal either way

maralorn:
@shlevy I suggest you modify (or leave) the proposal the way you are most convinced of. The committee can then decide if they want to pick an alternative.

aspiwack:
@tomjaguarpaw 

> The notion of type classes controlling something as simple as an exit code seems like far too much complexity to me.

On the contrary, I believe that type classes are precisely the right way to have a function work at multiple types. In practice, based on how people around me use this feature in Rust, only the `()`, `Void` and `ExitStatus` instances (provided by default) are likely to be used at all, though. So the type class itself is rather a technicality.

tomjaguarpaw:
> >  The notion of type classes controlling something as simple as an exit code seems like far too much complexity to me.
> 
> On the contrary, I believe that type classes are precisely the right way to have a function work at multiple types

I agree. I'm disputing whether `main :: IO <t>` should work for multiple types `<t>` at all. I don't see a strong need. However, if this feature has been proved out in other languages, such as Rust, then so be it.

shlevy:
> @tomjaguarpaw
> 
> > The notion of type classes controlling something as simple as an exit code seems like far too much complexity to me.
> 
> On the contrary, I believe that type classes are precisely the right way to have a function work at multiple types. In practice, based on how people around me use this feature in Rust, only the `()`, `Void` and `ExitStatus` instances (provided by default) are likely to be used at all, though. So the type class itself is rather a technicality.

For what it’s worth, I’ve done the closest equivalent C offers to domain-specific returns from main: https://github.com/shlevy/long-shebang/blob/6c83125ab9c7f63b16917599645ac75960f2c87d/long-shebang.c#L16-L30

shlevy:
However, I think the alternatives of insisting on `IO ()` or, modulo compatibility, `IO Void`, are pretty reasonable as well. I think if I were starting from scratch with a new language I’d lean toward the latter to emphasize that program exit is simply different from function return.

aspiwack:
@tomjaguarpaw 
> I agree. I'm disputing whether `main :: IO <t>` should work for multiple types `<t>` at all. I don't see a strong need. However, if this feature has been proved out in other languages, such as Rust, then so be it.

I had misunderstood your point then. Rust does give me confidence that the type-class thing is a reasonable approach. It's probably the easiest one to transition to. I expect it to be at any rate.

This sort of proposal is difficult to deal with, because the problem is small, yet real and really worth fixing. But the design space is very large, the constraints aren't clear. A very fertile ground for bikeshedding :slightly_smiling_face: .

tomjaguarpaw:
> This sort of proposal is difficult to deal with, because the problem is small, yet real and really worth fixing. But the design space is very large, the constraints aren't clear. A very fertile ground for bikeshedding 🙂 .

Yes indeed! My preference remains to not even try fixing this at the language or compiler level, but at the library level, as per https://github.com/ghc-proposals/ghc-proposals/pull/631#issuecomment-1898452290.

aspiwack:
@tomjaguarpaw 
> Yes indeed! My preference remains to not even try fixing this at the language or compiler level, but at the library level, as per [#631 (comment)](https://github.com/ghc-proposals/ghc-proposals/pull/631#issuecomment-1898452290).

It implements the feature, of course. But the problem is not that we can't return an `ExitCode`, it's that we can and it does something unexpected.

Just adding a warning, like @adamgundry proposes, addresses the issue satisfactorily in my eyes. @shlevy as the proposer what do you think about that alternative?

shlevy:
Updated. My preference is still to have the typeclass but I added the [warn on ambiguous main](https://github.com/shlevy/ghc-proposals/blob/io-exitcode/proposals/0631-main-return-types.rst#warn-error-on-ambiguous-main) alternative.

aspiwack:
@shlevy , my apologies for having drawn out the process so long. I think I still have quite a bit to learn on managing committee debate on this sort of small proposal with a very open-ended design space.

The committee prefers the following alternative:
- An extension is introduced (I'll leave the choice of a name to you). This extension is off by default at first, but the intent is to turn it on by default on the next language extension.
- When the extension is off, a warning (on by default) is emitted when `main` has a type other than `IO ()` or `IO Void`
- When the extension is on, a warning (the same) is emitted when `main` has a type other than `IO ()`, `IO Void`, `IO ExitCode`
- When the extension is on and `main` has type `IO ExitCode`, then the application exits with the code returned by `main` (no type class is introduced).

The reasoning is that this is the minimal change to achieve what you're looking for. We can always change the behaviour later to use a type class, whereas it would be difficult to remove the type class if we don't want it after all.

Are you ok with this plan? If so, please modify the proposal to reflect this, and I'll mark the proposal as accepted.

shlevy:
@aspiwack Re point 3, I prefer this be an error, not a warning. But if that has already been discussed and rejected or will take too long to cycle through, I’ll update the proposal to match your suggestion.

shlevy:
> I think I still have quite a bit to learn on managing committee debate on this sort of small proposal with a very open-ended design space.

One thought on this: I think a quick call earlier on might have enabled us to get through a bunch of the open questions and choices. Then the other thought is that ultimately I think the committee should just say “we’ve heard the options and choices, this is the one we like”

aspiwack:
> @aspiwack Re point 3, I prefer this be an error, not a warning. But if that has already been discussed and rejected or will take too long to cycle through, I’ll update the proposal to match your suggestion.

Ok for an error. I don't believe the committee has an opinion on this, so just go for it.

> One thought on this: I think a quick call earlier on might have enabled us to get through a bunch of the open questions and choices.

I've been reluctant to add calls to the committee's process. But I'll definitely try next time.

> Then the other thought is that ultimately I think the committee should just say “we’ve heard the options and choices, this is the one we like”

Well, the issue, really, was that the committee couldn't decide on what we collectively liked :smile: .

simonpj:
> Ok for an error. I don't believe the committee has an opinion on this, so just go for it.

yes I agree.  Go for it.

shlevy:
Updated!

aspiwack (commented):
Some wibbles below, then I'll mark the proposal as accepted. Thanks @shlevy , and my apologies again for the time we took with this.

shlevy:
Wibbles unwobbled

adamgundry:
Thanks @shlevy and @aspiwack! I've opened https://gitlab.haskell.org/ghc/ghc/-/issues/24848 to track this.

PR #633, opened 2024-01-26
Add table of contents to the README

s-and-witch:
See preview here: https://github.com/ghc-proposals/ghc-proposals/blob/75fb64a7d71d9350025195cfb7b6bfcbeceb55e3/README.rst#table-of-contents

This patch also adds links from all subsequent sections to the TOC

I'm using level 1 header here because `:local:` directive builds TOC from headings of a lower level and without `:local:` `contents` directive will spawn text like `Contents` above the TOC that looks like plain text.


s-and-witch:
Just a README update, CC @nomeata

nomeata:
LGTM

PR #570, opened 2023-01-11
Native string interpolation syntax

brandonchinn178:
The [proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0570-string-interpolation.rst) has been accepted; the following discussion is mostly of historic interest.

---

## Status

* GHC branch with prototype: [`wip/interpolated-strings`](https://gitlab.haskell.org/ghc/ghc/-/compare/wip%2Fstrings...wip%2Finterpolated-strings?from_project_id=1&straight=false)
* Opened CLC proposal: https://github.com/haskell/core-libraries-committee/issues/416
* Last major revision: May 15 2026
* HIW 2026 slides: https://brandonchinn178.github.io/presentations/2026-06-05-hiw-string-interpolation.html

## Related discussions

* #569
* https://gitlab.haskell.org/ghc/ghc/-/issues/20862
* https://www.reddit.com/r/haskell/comments/i3f3ip/state_of_string_interpolation_in_haskell/

PR #569, opened 2023-01-08
Multiline string literals

brandonchinn178:
Resolves #260.

Doesn't address QuasiQuote deprecation because ~~I disagree that QuasiQuotes are **only** useful for multiline strings; QuasiQuotes are being used for much more than multiline strings~~ that can be a separate proposal [^1].

[Rendered](https://github.com/brandonchinn178/ghc-proposals/blob/multiline-strings/proposals/0569-multiline-strings.rst)

[^1]: I originally didn't understand what the relation was; I now understand that it's due to the fact that `[foo|...|]` could be written with `$(foo """...""")`; the only thing restricting it before was lack of multiline support

PR #637, opened 2024-02-16
Modify multiline string proposal to remove one trailing newline

brandonchinn178:
After thinking through this more, I think it makes more sense for exactly one trailing newline to be stripped, the same as exactly one leading newline is stripped.

Current proposal: https://github.com/ghc-proposals/ghc-proposals/blob/424b22fd7fb15c56aa4fbdfa1ed86d3f5245eda2/proposals/0569-multiline-strings.rst

### Abstract motivation

The abstract reason why I think the trailing newline should be omitted is to be consistent with the leading newline. One behavior I think most people in the original thread agreed with is that most of the time, the developer doesn't intend the newline right after the `"""` delimiter to be included. If a leading newline is intended, I think most people think it's more obvious when there's an explicitly blank line in between:

```
x = """

  This has exactly one leading newline
  """
```
I think it makes the overall feature more consistent and intuitive to have the same behavior for both beginning and end.

### Practical motivation

Practically speaking, I think it composes a bit better, e.g.
```
s1 =
  """
  line 1
  line 2
  """

s2 = "line 3"

s3 =
  """
  line 4
  line 5
  """

putStrLn $ unlines [s1, s2, s3]

{-
Old behavior:
   "line 1\n"
++ "line 2\n"
++ "\n"
++ "line 3\n"
++ "line 4\n"
++ "line 5\n"
++ "\n"

New behavior:
   "line 1\n"
++ "line 2\n"
++ "line 3\n"
++ "line 4\n"
++ "line 5\n"
-}
```

But then again, perhaps this is contrived, and people would just use `concat` instead.

### DIFF
--- a/proposals/0569-multiline-strings.rst
+++ b/proposals/0569-multiline-strings.rst
@@ -119,6 +119,8 @@ A working prototype is available at `brandonchinn178/string-syntax <https://gith
 
    #. If the first character of the string is a newline, remove it
 
+   #. If the last character of the string is a newline, remove it
+
 #. After parsing, it becomes indistinguishable to the equivalent single-quoted string (modulo annotations for exact-printing)
 
 Common whitespace prefix calculation
@@ -207,6 +209,10 @@ Step 2vi - Since the first character is a newline character, we remove it and ar
 
   "..abc\\n\\n..def\\n\\nghi\\n....\\njkl\\n"
 
+Step 2vii - Since the last character is a newline character, we remove it and are left with the final result::
+
+  "..abc\\n\\n..def\\n\\nghi\\n....\\njkl"
+
 Step 3 - This gets treated as a normal string from now on, with the escaped ``\\n`` characters interpreted as usual.
 
 Ignore leading characters
@@ -230,7 +236,7 @@ The common prefix calculation ignores all characters preceding the first newline
     """
 
   -- equivalent to
-  s' = "Line 1\n   Line 2\nLine 3\n"
+  s' = "Line 1\n   Line 2\nLine 3"
 
 This implies that normal strings could also be written using ``"""``
 
@@ -255,7 +261,7 @@ String gaps are collapsed first and not included in the whitespace calculation
       """
 
   -- equivalent to
-  s' = "a b c d e\nf g\n"
+  s' = "a b c d e\nf g"
 
 Mixing tabs and spaces
 ~~~~~~~~~~~~~~~~~~~~~~
@@ -272,7 +278,7 @@ In the following example, each line has 16 leading spaces after expanding tabs.
   ⇥"""
 
   -- equivalent to
-  s' = "a\nb\nc\n"
+  s' = "a\nb\nc"
 
 Leading newline
 ~~~~~~~~~~~~~~~
@@ -290,27 +296,24 @@ The specification strips exactly one leading newline, which is the behavior of l
     """
 
   -- equivalent to
-  s' = "\na\nb\nc\n"
+  s' = "\na\nb\nc"
 
 Trailing newline
 ~~~~~~~~~~~~~~~~
 
-As mentioned in the example in *Section 3.1 General Walkthrough*, trailing newlines are naturally included without any explicit rules. As a bonus, it does the same thing that ``unlines`` does. To avoid a trailing newline, put the closing ``"""`` immediately after the last line, or use a string gap:
+Similarly to a single leading newline being removed, a single trailing newline will also be removed. To keep the trailing newline, add a blank line after the last line:
 
 ::
 
-  x =
+  s =
     """
     a
     b
-    c"""
 
-  x2 =
     """
-    a
-    b
-    c\
-    \"""
+
+  -- equivalent to
+  s' = "a\nb\n"
 
 Indent every line
 ~~~~~~~~~~~~~~~~~
@@ -328,7 +331,7 @@ In the following example, desugaring ``s1`` into ``s1'`` removes the 2 spaces be
       c
     """
 
-  s1' = "a\nb\nc\n"
+  s1' = "a\nb\nc"
 
   s2 =
     """
@@ -344,7 +347,7 @@ In the following example, desugaring ``s1`` into ``s1'`` removes the 2 spaces be
     \&  c
     """
 
-  s2' = "  a\n  b\n  c\n"
+  s2' = "  a\n  b\n  c"
 
 Escaping triple quotes
 ~~~~~~~~~~~~~~~~~~~~~~
@@ -379,7 +382,7 @@ Since escaped characters are resolved *after* calculating the common whitespace 
 
 ::
 
-  s' = " name\tage\n Alice\t20\n Bob\t30\n\t40\n"
+  s' = " name\tage\n Alice\t20\n Bob\t30\n\t40"
 
 Example from Fourmolu
 ~~~~~~~~~~~~~~~~~~~~~
@@ -434,6 +437,7 @@ With multiline strings:
       Aeson.Null -> pure PrintStyleInherit
       Aeson.String "" -> pure PrintStyleInherit
       _ -> PrintStyleOverride <$> Aeson.parseJSON v
+
     """
 
   adtParsePrinterOptType =
@@ -441,6 +445,7 @@ With multiline strings:
     \\s -> case s of
       "" -> pure PrintStyleInherit
       _ -> PrintStyleOverride <$> parsePrinterOptType s
+
     """
 
 While the double backslash is still required, I think the overall style is much better (could be resolved in a later proposal adding raw strings).
@@ -529,6 +534,7 @@ With multiline strings:
               [ "unknown value: " <> show s
               , "Valid values are: %s"
               ]
+
     """
     fieldTypeName
     fieldTypeName
--- a/proposals/0569-multiline-strings.rst
+++ b/proposals/0569-multiline-strings.rst
@@ -230,7 +230,7 @@ The common prefix calculation ignores all characters preceding the first newline
     """
 
   -- equivalent to
-  s' = "Line 1\n   Line 2\nLine 3"
+  s' = "Line 1\n   Line 2\nLine 3\n"
 
 This implies that normal strings could also be written using ``"""``
 
@@ -328,7 +328,7 @@ In the following example, desugaring ``s1`` into ``s1'`` removes the 2 spaces be
       c
     """
 
-  s1' = "a\nb\nc"
+  s1' = "a\nb\nc\n"
 
   s2 =
     """
@@ -344,7 +344,7 @@ In the following example, desugaring ``s1`` into ``s1'`` removes the 2 spaces be
     \&  c
     """
 
-  s2' = "  a\n  b\n  c"
+  s2' = "  a\n  b\n  c\n"
 
 Escaping triple quotes
 ~~~~~~~~~~~~~~~~~~~~~~
@@ -379,7 +379,7 @@ Since escaped characters are resolved *after* calculating the common whitespace 
 
 ::
 
-  s' = " name\tage\n Alice\t20\n Bob\t30\n\t40"
+  s' = " name\tage\n Alice\t20\n Bob\t30\n\t40\n"
 
 Example from Fourmolu
 ~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


jvanbruegge:
I would prefer not to have these extra trailing newlines like in the original proposal

brandonchinn178:
@jvanbruegge I don't understand. These trailing newlines are a consequence of the spec as accepted by the proposal. All of the other examples have them. It was an oversight to omit them for these specific examples.

This PR should be merged because it updates the examples to match the behavior as currently described. Making these not end with a trailing newline is a change to the spec that requires additional discussion.

brandonchinn178:
@gridaphobe What's the process for modifying an in-progress proposal? Do I need to make a separate proposal, or is this fine?

cc-ing some people involved in the first discussion:
@adamgundry @maxigit @nomeata @ChickenProp @konsumlamm

gridaphobe:
@brandonchinn178 see https://github.com/ghc-proposals/ghc-proposals?tab=readme-ov-file#how-to-amend-an-accepted-proposal, it's basically the same process.

gridaphobe:
@brandonchinn178 could you provide some motivation for this change? Maybe you found some examples while working through the implementation where stripping a single trailing newline made more sense?

Just going off the real-world examples you took from fourmolu, the proposed change seems to make things worse. It also seems like many other languages preserve the trailing newline (I checked Java, Python, Ruby, Scala).

jvanbruegge:
The new rule is e.g. what nix does.
This allows to put the closing quite on a new line (similar how the opening quote can be on a new line:
```
foo = """
    bar
  """"
-- equivalent to "bar"
```
otherwise only the opening quote could be on its own line

brandonchinn178:
I personally don't care too much about Nix, so I'm not too motivated by the Nix behavior.

The fourmolu examples add an extra newline just to keep parity with the unlines example. In the actual fourmolu code, we actually use an `unlines_` helper that omits the ending newline.

### Abstract motivation

The abstract reason why I think the trailing newline should be omitted is to be consistent with the leading newline. One behavior I think most people in the original thread agreed with is that most of the time, the developer doesn't intend the newline right after the `"""` delimiter to be included. If a leading newline is intended, I think most people think it's more obvious when there's an explicitly blank line in between:

```
x = """

  This has exactly one leading newline
  """
```
I think it makes the overall feature more consistent and intuitive to have the same behavior for both beginning and end.

### Practical motivation

Practically speaking, I think it composes a bit better, e.g.
```
s1 =
  """
  line 1
  line 2
  """

s2 = "line 3"

s3 =
  """
  line 4
  line 5
  """

putStrLn $ unlines [s1, s2, s3]

{-
Old behavior:
   "line 1\n"
++ "line 2\n"
++ "\n"
++ "line 3\n"
++ "line 4\n"
++ "line 5\n"
++ "\n"

New behavior:
   "line 1\n"
++ "line 2\n"
++ "line 3\n"
++ "line 4\n"
++ "line 5\n"
-}
```

But then again, perhaps this is contrived, and people would just use `concat` instead.

maxigit:
I would prefer the new proposal : one leading and one trailing newline removed. It's consistent, easy to remember, allows the openning and closing quotes to have the same indentation and if people wants extra newlines they can add them easily.

I never liked the current proposal forcing me to close the quotes at the end of the last line (I pretty much never write strings which end up with a newline)

VitWW:
This behavior is more intuitive clear.

Some notes:
```haskell
s1 =
  """
  line 1
  """

s1 == "line1"

s2 =
  """
  line 1

  """

s2 == "line1\n"
```

And we have 1 interesting consequence: 2 alternatives to wrtie empty string
```haskell
s3 =
  """
  """

s4 =
  """

  """

s3 == "" == s4
```

brandonchinn178:
@adamgundry Can we bring this up for review? I'm just about finished with the PR, so I'd like to bring this into the implementation. Comments seem to be generally in favor, with only one comment raising a question about motivation, which I updated in the PR description.

erikd:
Other than commenting that I like the proposal in its current state, I have nothing to add.

brandonchinn178:
@gridaphobe Any idea when we can do next steps? I just got approval to merge the [MR](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/12089)

simonpj:
Thoughts

* Eric Siedel is the shepherd for this change.  Our [process](https://github.com/ghc-proposals/ghc-proposals#committee-process-for-responding-to-a-proposal) asks that the shepherd comes up with a recommendation to the committee within 2 weeks.  So far it has been 9 days.
* After that it depends on how controversial, or otherwise, the proposal is; but a few weeks at most.  In the case of this, very small proposal, I think it should be much shorter, depending on availability of committee members.
* In the interim I suggest you include the change in your MR.  Provided you promise to change it back if the committee wants, I think you can safely merge it too.  It's a very tiny change!

As to the content

* The OP at the top of the thread doesn't actually point to the existing proposal, which makes it harder to review.  Could it do so please?
* In [the spec](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0569-multiline-strings.rst), it says
  * Near the beginning: (ii) split the string by newlines
  * At the end: (vi) if the string starts with a newline, remove it. (This is the bit you are modifying.)
  Could the "remove leading and trailing newline" bit be put at the beginning with step (ii)?  That seems simpler to me.  Get rid of them up front, precisely when you are dealing with "split by newlines".  If there is a subtle reason why not could the proposal spell out that reason with an example?


Generally I'm thumbs up for this change.

proposals/0569-multiline-strings.rst
```diff
@@ -112,6 +112,7 @@ A working prototype is available at `brandonchinn178/string-syntax <https://gith
 
       * See the "Common whitespace prefix calculation" section below for the specification of the calculation
       * If a line only contains whitespace, remove all of the whitespace
+      * Don't remove common whitespace immediately after the delimiter, see the example in *Section 3.2 Ignore leading characters*
```
brandonchinn178, line 115: This isn't a change in the spec, it's a clarification that follows from the "Ignore leading characters" section:

> This means that characters immediately after the ``"""`` delimiter will be included verbatim

brandonchinn178:
@simonpj Thanks for noting the timeframe, I missed that in the doc, sorry about that. I'll go ahead and merge the MR with this change, but I'll definitely revert the change if the proposal is rejected.

I updated the PR description with a link to the original proposal, and I added a commit with the answers to your question.

ad-si:
My 2 cents: If I have a multiline text, i almost always want to have a trailing newline (save to a file, insert into a document, …) so it should be the default for the most common writing style without the empty newline before the closing """

simonpj:
> My 2 cents: If I have a multiline text, i almost always want to have a trailing newline (save to a file, insert into a document, …) so it should be the default for the most common writing style without the empty newline before the closing """

That is a good point.  @brandonchinn178 ?   I have no experience to draw on here.

jvanbruegge:
I'd prefer no trailing newline. Also for symmetry, as the first newline after """ is removed.

brandonchinn178:
@ad-si I do agree that writing to files should always end with a newline, but I'm still compelled by the poor composition between multiline strings and normal strings, as described in the PR description. IMO it's better for the default behavior to strip for composability, and then be explicit about the trailing newline in the instances when you need it:
```hs
writeFile "foo.txt"
  """
  my file content
  on multiple lines
  here
  
  """
```
Thoughts?

akhra:
another +1 for symmetry, and I'm also cold on specializing around what's ultimately an arbitrary (and IO-centric!) convention. I'd rather be explicit about sanitization and target-formatting.

maxigit:
> My 2 cents: If I have a multiline text, i almost always want to have a trailing newline (save to a file, insert into a document, …) so it should be the default for the most common writing style without the empty newline before the closing """

I think both use cases (with or without trailing newline) will happend and I don't know which is the most common (In my particular case it will be WITHOUT). However, the benefit of removing it is it allows syntaxing features (the closing quote) to stay to the left (and indent them, etc ...) regardless of the length of the line. I prefer


    longline = """
                              my really long                                                                                           ....                                               line
                       """

over 


    longline = """
                                my really long                                                                                           ....                                               line"""

In the first example I can see easily that my string is closed.longline = """
                 my really long                                                                                           ....                                               line
                       """


Ultimately, I think both ways are fine and the sooner in GHC the better ...

VitWW:
This Proposal is changing in strange way.
It's very good design of "heredoc/newdoc" string-syntax at PHP: [string.heredoc](https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc).
In brief, additional rules for multiline quotes:
1) _After_ opening `"""` quote is allowed just spaces/tabs/newline symbols in same line, compile error otherwise
2) _Before_ closing `"""` quote is allowed just spaces/tabs symbols in same line, otherwise it is not a closing quote

So, next must be an errors:
```haskell
  s1 = """hello world"""

  s2 =
    """    hello
    world
    """

  s3 =
    """
    a
    b
    c"""
```

VitWW:
To those, who disagree:
#569 : do you see a conflict here?

> 3.6   Trailing newline
> As mentioned in the example in Section 3.1 General Walkthrough, trailing newlines are naturally included without any explicit rules. As a bonus, it does the same thing that unlines does. To avoid a trailing newline, put the closing """ immediately after the last line, or use a string gap:
>  ```haskell
>  x1 =
>    """
>    a
>    b
>    c"""
>  
>  x2 =
>    """
>    a
>    b
>    c\
>    \"""
>   ````
> 3.8   Escaping triple quotes
> Only three literal """ characters in a row will end the multiline string, so escaping any or all of the quote characters will not terminate the string:
>  ```haskell
>  x3 =
>    """
>    This is a literal multiline string:
>    \"\"\"
>    Hello
>      world!
>    \"""
>    """
>  ```

But `\"""` in `x3` and in `x2` are the same or different?

brandonchinn178:
@VitWW 

No, there's no conflict or ambiguity. It's the same as
```hs
s1 = "a\"b"

s2 = "a\  \"
```
The first escapes the quote, the second ends a string gap.

And the original proposal discussion already contains a bunch of comments about keeping characters before/after delimiters verbatim; I don't really want to rehash that discussion here, as nothing's changing on that front.

VitWW:
> I don't really want to rehash that discussion here, as nothing's changing on that front.

Main motivation to has multiline-strings in languages - is to have easy readable texts with avoiding escapes as much as possible.

And allowing closing quote not in the new line  - it is a request to escape every internal `"""`

But Ok, changing just trailing end-line is still better then to remain original proposal as is.

gridaphobe:
I was initially skeptical of this amendment, thinking it would be better for consistency to have the trailing newline. But thinking more through the ways we use strings:

* The proposal mentions string concatenation as a motivation. I would add string interpolation as a use case (even though we still don't have first-class interpolation in Haskell). Embedding the trailing newline in the string leads to contortions like
```
str1 = """
       hello
       """
-- trailing newline in `str1`, need to avoid a superfluous newline
str2 = "${str1}world"
-- no trailing newline, naturally reflects the desired output
str2 = """
       $str1
       world
       """
```
* We typically use `putStrLn` for printing, and it's frustrating to have to reason about when `putStr` would be more appropriate due to trailing newlines.


I recommend we accept this amendment.

brandonchinn178:
Excellent points! String interpolation was also a hidden motivation for me 🙂 #570

simonpj:
> Excellent points! String interpolation was also a hidden motivation for me 🙂 https://github.com/ghc-proposals/ghc-proposals/pull/570

Is that an explicit motivation in the proposal, supported with examples?  It would be good if it was, and no too late to add it if not.   It all helps understanding

ad-si:
I just know I will be writing a lot of this in the future 😮‍💨:

```hs
s1 =
  """
  line 1
  line 2\n
  """
```
(Using an empty newline before the closing `"""` looks too much like an accident in my opinion)

On the other hand, I guess you could argue that this explicit newline is actually a win as it's now abundantly clear that there is a trailing newline.

VitWW:
Original proposal has a minor bug in rules.  It wrongly removes whitespaces in short lines, and non-empty first lines.

So we need to change next line:
```
   #. Remove common whitespace prefix in every line
```
into
```
   #. Empty first line if first line only contains whitespaces and if multi-line consists more then 1 line

   #. Empty every line (excluding first line) if line only contains whitespaces

   #. Remove common whitespace prefix in every line (excluding first line and empty ones)
```

gridaphobe:
Thank you @brandonchinn178, I am declaring the amendment accepted.

PR #638, opened 2024-02-21
Prefix form for MkSolo# (amendment of #475: Non-punning list and tuple syntax)

int-index:
This amendment to [proposal #475](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0475-tuple-syntax.rst) has been accepted; the following discussion is mostly of historic interest.

----

At the moment, the proposal defines

```
type Unit# :: TYPE (TupleRep [])
data Unit# = (# #)

type Solo# :: TYPE rep -> TYPE (TupleRep [rep])
data Solo# a = (# a #)
```

Alright, so what are the prefix forms of those constructors? The data constructor of `Unit#` is `(# #)`, and the data constructor of `Solo#` is... also `(# #)`! This is an ambiguity.

We can resolve this ambiguity by renaming the data constructor of `Solo#` to `MkSolo#`, retaining the mixfix `(# a #)` syntax of course.

Please review: @goldfirere, @tek

### DIFF
--- a/proposals/0475-tuple-syntax.rst
+++ b/proposals/0475-tuple-syntax.rst
@@ -335,7 +335,7 @@ Proposed Change Specification
 
    #. An occurrence of ``(# x #)`` is a synonym for ``GHC.Types.MkSolo# x`` in both term- and type-syntax.
       (It is not valid in types at the time of writing, as GHC is unable to promote unboxed tuples, but this concern
-       is separate from syntax.)
+      is separate from syntax.)
 
    #. A use of ``(# ... | ... | ... #)``, where each of the ``...`` is filled in, (among other arities) is now disallowed.
 
--- a/proposals/0475-tuple-syntax.rst
+++ b/proposals/0475-tuple-syntax.rst
@@ -142,7 +142,7 @@ Proposed Change Specification
      data Unit# = (# #)
 
      type Solo# :: TYPE rep -> TYPE (TupleRep [rep])
-     data Solo# a = (# a #)
+     data Solo# a = MkSolo# a
 
      type Tuple0# = Unit#
      type Tuple1# = Solo#
@@ -275,6 +275,8 @@ Proposed Change Specification
 
    #. With ``-XUnboxedTuples``, an occurrence of ``(# ty #)`` in type-syntax is a synonym for ``GHC.Types.Solo# ty``.
 
+   #. With ``-XUnboxedTuples``, an occurrence of ``(# x #)`` in term-syntax is a synonym for ``GHC.Types.MkSolo# x``.
+
    #. With ``-XUnboxedTuples``, an occurrence of ``(#,,...,,#)`` where there are *n* commas (for *n* ≧ 1) in type-syntax
       is a synonym for ``GHC.Types.Tuple<n+1>#``.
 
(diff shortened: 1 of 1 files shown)


tek:
It seems like a definite oversight to me, I can't find any arguments for not having `MkSolo#`.

proposals/0475-tuple-syntax.rst
```diff
@@ -463,6 +469,11 @@ Effect and Interactions
    means either ``Tuple2# Int Bool`` (with ``-XNoListTuplePuns``) or ``Solo# (Tuple2 Int Bool)`` (with
    ``-XListTuplePuns``). Furthermore, ``Tuple# (Int#, Double#)`` would not kind-check with ``-XListTuplePuns``. So we avoid these problems by simply erroring in the 1-element case.
 
+#. The name of the data constructor for unboxed 1-tuples is ``MkSolo#`` rather
+   than ``(# #)`` to distinguish it from the data constructor for 0-tuples.
+   The ambiguity only arises when the name is used unapplied.
+   When applied to an argument, ``MkSolo# a`` can be written and pretty-printed as ``(# a #)``.
```
goldfirere, line 475: "can be" or "is" (at least for printing)? That is, I would want the prefix `MkSolo#` to be a rare sight for programmers.

goldfirere (commented):
I think we need to fix this. But I'm a little worried about programmer confusion here. I think that the pretty-printer should continue to print applied unary tuples using mixfix syntax. Is there ever a case where we have to print an un-applied constructor where the user didn't write this? If not, then no problem. If there is, we should be careful in an error message to tell the user of the relationship between `MkSolo# x` and `(# x #)`.

proposals/0475-tuple-syntax.rst
int-index: Updated the text to reflect this preference

int-index:
@adamgundry Let's submit this for committee consideration.

simonpj:
Reading this amendment makes me wonder:

1. The canonical names for the tuple *type* constructors are `Tuple2`, `Tuple3`, ... and `Tuple2#`, `Tuple3#`, ...
1. The canonical names for the tuple *data* constructors are `(,)`, `(,,)`, .. and `(#,#)`, `(#,,#)`... 
1. The proposal (amended) adds `MkSolo` and `MkSolo#` as the canonical names for unit tuples.  Some change is forced, since the existing canonical names don't work.
1. The proposal also adds `Unit`, `Solo` and `Unit#`, `Solo#` as the canonical names for the type constructors, with `Tuple0`, `Tuple1`, `Tuple0#` and `Tuple1#` as synonyms.  **This is an un-forced change.

**Concerning (3)**, an alternative could be to stick with `Tuple0`, `Tuple1`, `Tuple0#` and `Tuple1#` as the canonical names for the type constructors, with (if you like) `Unit` and `Solo` as type synonyms.  Maybe that would be more uniform?  Thus
```
type Solo = Tuple1;  data Tuple1 a = MkSolo a
```
rather than
```
type Tuple1 = Solo; data Solo a = MkSolo a
```

**Concerning (4)**, we are free to invent different canonical names for our unit data constructors.   Rather than `MkSolo` we could have `(_)` for example, and similarly `(#_#)`.  As others have said, they are usually printed distifx.  Ah: the distfix form of `(_) x` would naturally be `( x )`, and that's ambiguous.   Bother.  Maybe that's the main argument for `MkSolo`.  And if `MkSolo` then clearly `MkSolo#`.

I am not advocating strongly here for any one path.  This is detail stuff.  But we will be stuck with this forever; I hadn't seen the non-uniformity so clearly before; and uniformity is a virtue.

tek:
> **Concerning (3)**, an alternative could be to stick with `Tuple0`, `Tuple1`, `Tuple0#` and `Tuple1#` as the canonical names for the type constructors, with (if you like) `Unit` and `Solo` as type synonyms. Maybe that would be more uniform? 

Compelling idea. Note however that the new names are already in [GHC 9.8](https://hackage.haskell.org/package/ghc-prim-0.11.0/docs/GHC-Tuple-Prim.html)

tek:
> Of course, we could also add special logic just for those two types to prefer type synonyms in user-facing compiler output.

FWIW the tuple printing logic is as specialized as it can get already

int-index:
What about class instances? Say I want to declare

```
instance C Unit where
  ...

instance C (Solo a) where
  ...
```

Do I have to turn on `TypeSynonymInstances`?

Tritlo:
This proposed amendment has been accepted. The alternative described by Simon is left for a future amendment to #475.

PR #641, opened 2024-03-08
Wildcard binders in type declarations (amend 425)

int-index:
The amendment to the [proposal](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0425-decl-invis-binders.rst) has been accepted; the following discussion is mostly of historic interest.

---

This proposal amendment is based on https://gitlab.haskell.org/ghc/ghc/-/issues/23501

**Background.** A term-level pattern can take many forms, and some of those forms are irrefutable (i.e. no actual runtime matching required):
* a variable `v`
* a wildcard `_`
* parenthesized variable `(v)`
* parenthesized wildcard `(_)`
* parenthesized variable with type annotation `(v :: t)`
* parenthesized wildcard with type annotation `(_ :: t)`
* nested parentheses are allowed: `((v))`
* nested type annotations are allowed: `(((v :: t1) :: t2) :: t3)`

All those forms aren't truly patterns in the sense of _pattern matching_ (there is no matching). It would be more fair to call them _binders_, as they simply bind _any_ input argument to a name. It's only with data constructors that we get actual patterns, e.g. `Just x`.

In function equations, patterns can be prefixed with `@` to bind invisible arguments. That is, we also have `@tv`, `@_`, `@(_)`, `@(tv :: t)`, `@(_ :: k)`, `@((tv))`, `@(((tv :: k1) :: k2) :: k3)`, etc.

**Proposed amendment.** The key idea is that binders in type declarations, such as `type T @a b = ...`, are rather similar to those irrefutable term-level patterns (binders). So it is rather strange that in types, the grammar is limited to only five arbitrarily picked variants: `tv`, `(tv :: k)`, `@tv`, `@(tv :: k)`, and `@_`. Where are all the other forms, e.g. `_`, `(_ :: k)`, or `((tv :: k1) :: k2)`? I propose that we use a grammar that is as general as possible, which would allow `type T _ = ...` among other perfectly sensible declarations.

### DIFF
--- a/proposals/0425-decl-invis-binders.rst
+++ b/proposals/0425-decl-invis-binders.rst
@@ -7,7 +7,7 @@ Invisible binders in type declarations
 .. ticket-url::
 .. implemented::
 .. highlight:: haskell
-.. header:: This proposal was `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/425>`_.
+.. header:: This proposal was `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/425>`_ and `amended by #641 <https://github.com/ghc-proposals/ghc-proposals/pull/641>`_.
 .. contents::
 
 We propose to allow invisible type variable binders (i.e. ``@k``) in type
@@ -415,21 +415,29 @@ Proposed Change Specification: Declarations
 Syntax
 ~~~~~~
 
-Relax the syntactic check of ``data``, ``newtype``, ``type``, ``class``,
-``type family``, and ``data family`` declarations to allow ``@k``-binders in
-their headers::
+Generalize the grammar of type variable binders in ``data``, ``newtype``,
+``type``, ``class``, ``type family``, and ``data family`` declarations to allow
+``@k``-binders, wildcards, and combinations thereof::
 
+  -- Old grammar (for reference):
   tv_bndr ::=
            | tyvar                         -- variable
-           | '(' tyvar '::' kind ')'       -- variable with kind annotation
-    (NEW)  | '@' tyvar                     -- invisible variable
-    (NEW)  | '@' '(' tyvar '::' kind ')'   -- invisible variable with kind annotation
-    (NEW)  | '@' '_'                       -- wildcard (to skip an invisible quantifier)
+           | '(' tyvar '::' kind ')'       -- variable with a kind annotation
+
+  -- New grammar:
+  tv_bndr   ::=
+             |     a_tv_bndr               -- binder for a visible argument
+             | '@' a_tv_bndr               -- binder for an invisible argument
+  a_tv_bndr ::=
+             | tyvar                       -- variable
+             | '_'                         -- wildcard
+             | '(' tyvar '::' kind ')'     -- variable with a kind annotation
+             | '(' '_'   '::' kind ')'     -- wildcard with a kind annotation
 
 The occurrences of ``@`` must be *prefix*, as defined by
 `#229 <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0229-whitespace-bang-patterns.rst>`_.
 
-Guarded behind a new flag, ``-XTypeAbstractions``.
+New forms of binders are guarded behind a new flag, ``-XTypeAbstractions``.
 
 Scope
 ~~~~~
@@ -753,6 +761,23 @@ Alternatives
 
   Some thoughts on the topic can be found in GitHub comment `326/634791269
   <https://github.com/ghc-proposals/ghc-proposals/pull/326#issuecomment-634791269>`_.
+
+* The grammar of binders could be further generalized to allow nested
+  parentheses and nested kind annotations::
+
+    tv_bndr   ::=
+               |     a_tv_bndr             -- binder for a visible argument
+               | '@' a_tv_bndr             -- binder for an invisible argument
+    a_tv_bndr ::=
+               | tyvar                     -- variable
+               | '_'                       -- wildcard
+               | '(' p_tv_bndr ')'         -- parenthesized binder
+    p_tv_bndr ::=
+               | a_tv_bndr                 -- plain binder
+               | a_tv_bndr '::' kind       -- binder with a kind annotation
+
+  The committee decided against this option to avoid the introduction of
+  another recursive data type to the AST.
 
 Unresolved Questions
 --------------------


TeofilC:
Will this entail a breaking change for folks using the `.Lib` functions in `template-haskell`?
I'm assuming as it changes syntax it will be a breaking change for folks directly using the AST.

int-index:
> Will this entail a breaking change for folks using the `.Lib` functions in `template-haskell`? I'm assuming as it changes syntax it will be a breaking change for folks directly using the AST.

That's right. The current definition cannot represent a wildcard binder

```
data TyVarBndr flag = PlainTV  Name flag
                    | KindedTV Name flag Kind
```

In fact, it can't represent the invisible wildcard binder `@_` either, so the proposal in its current form also requires a breaking change, amendment or not.

My goal is to minimize breaking changes: I wouldn't want us to make another one a year later, and then another one again, etc. So in this amendment I'm trying to come up with a future-proof, general definition of a binder.

VitWW:
Does "breaking change" means that what is **allowable** now, in future it would mean something else or it would be forbidden?

As I understood the proposal, some possibilities are forbidden now, and in future they would be allowable. This is not a "breaking change".

int-index:
@VitWW True, the change to the _language_ is backwards compatible. Strictly more programs will be considered valid Haskell. However, as you might be aware, GHC has metaprogramming facilities that are exposed to the user via the [template-haskell](https://hackage.haskell.org/package/template-haskell) library. Allowing new syntax often entails modifying `template-haskell`, and in this case it's a breaking change to the API of this library.

simonpj:
I have no objections in principle -- except that it add un-forced complexity.  It particular means that the syntax (and hence data type) for  "binders" becomes recursive.

It's nice to have, perhaps, but there are many other things I think would bring much more value for our invested cycles.  (E.g. the recent thread on a stable TH API: #24021.)

I'd be inclined to wait until our users ask for it.

int-index:
> I'd be inclined to wait until our users ask for it.

I find that it is in the spirit of Haskell to support corner cases, e.g. `let {} in e` is rather useless because it's the same as `e`, but GHC still supports it. So when we talk about supporting `tv` and `(tv :: k)`, how come we don't support `(tv)`? Isn't it strange that parentheses and a kind annotation form a single fused construct? And if we support `(tv)`, then what about `((tv))`? Maybe not really useful, but has the same feel as `let {} in e`.

I say we just list our building blocks and support them in full generality. The building blocks are: type variables, wildcards, parentheses, kind annotations. Any combination thereof should be valid.

We don't need users to ask for it. It just feels like the right thing to support corner cases (at least to me).

tomjaguarpaw:
I for one would welcome the increased consistency, all else being equal.

maralorn:
I’d also argue that this is not about corner cases but the principle of least surprise. Let’s have consistent syntax.

int-index:
OK, it's been quiet for a week. Let's submit this to the committee. @adamgundry, over to you.

AntC2:
> how come we don't support `(tv)`?

Hmm. Haskell syntax has some odd corners of inconsistency. In terms, `(v)` means the same as `v`, but `(+)` doesn't mean the same as `+`; `MkT (<expr1>) (<expr2)` means apply constructor to both exprs, but `MkT {<stuff>} (<expr2>)` binds `MkT` syntactically tighter to stuff, doesn't apply `MkT` to expr2 whether or not stuff supplies all fields -- so is probably a syntax error, ?but reported as a type error; `MkT {}` binds tightly and 'applies' `MkT` to no arguments and is valid even for constructors defined without fields; `v {<stuff>}` is a field update; `(v) {<stuff>}` is invalid syntax IIRC.

In declarations:

```
MkT x y   = ...  -- } declares and binds x, y     -- this I don't like, because ...
(MkT x y) = ...  -- } means the same
foo x y   = ...  -- declares foo                  -- ... it reminds me too much of this
(foo x y) = ...  -- syntax error
```

"A foolish consistency is the hobgoblin ..."? Let sleeping dogs lie?

maralorn:
`(v) { foo = 5 }` is not a syntax error. Also operator sections are an integral part of Haskell, but as a pattern `(+)` is a syntax error. I’d say your examples are only peripherally connected to the proposal.

AntC2:
> `(v) { foo = 5 }` is not a syntax error.

Ah, apologies then the other way round: usually in expressions `MkT` is same as `(MkT)`; but `MkT { foo = 5 }` is not the same as `(MkT) { foo = d }`. (That second is treated as field update but `(MkT)` wasn't applied to any fields/is of wrong type.)

simonpj:
Vlad writes (in email):

>  So the current accepted proposal #425 is no good in its current form: it is rather inconsistent to have `@_` but not `_`.

Ah, I had not realised that.  (It is not mentioned in the OP of this discussion, nor in the amendment; it is mentioned implicitly in [#23501](https://gitlab.haskell.org/ghc/ghc/-/issues/23501): " I think it would also be fine to allow _ binders without the @, even though this is not included in the proposal".)   If I missed this point, perhaps others did too.

Summarising, thenthe proposal of this thread really has three components:

1. Add wildcards to *all* binders in type declarations, not just `@_`.
2. Allow kind annotations for wildcards
3. Support all this recursively.

For (1) and (2) consider
```
data T1 @_         = MkT1  -- Allowed by #425
data T2 @(_::Nat)  = MkT2  -- Not currently allowed by #425
data T3 _          = MkT3  -- Not currently allowed by #425
data T4 (_ :: Nat) = MkT4  -- Not currently allowed by #425
```
I agree that all four make sense.  A proposal to support all four is OK with me.

As I say I am not keen on (3).

None of these changes would be breaking changes.

As to TH, I think we are not *forced* to change anything.  When converting into TH we could just use a fresh variable.  So if the programmer uses a TH quotation `[d| type T @_ = ... |]`, the generated AST would actually be equivalent to `[d| type T @_w1 = ... |]` where `_w1` is a fresh name.  If that becomes a problem for some reason, we can make a breaking change in TH; but *until* it becomes a problem I see little motivation to *force* a breaking change on our users.

maralorn:
I feel like I have no calibration how serious a TH change is?  Can anyone put numbers on how often we have breaking changes in TH? My feeling until know was that it has to be relatively common. Also, even if the change is technically breaking, how much users would be affected given that probably most usages of TH don't use the whole grammar?

goldfirere:
For what it's worth, I agree with @simonpj here. Let's do this -- but not recursively. Datatype declarations remain different from functions, as the former have a fixed arity. If/when we can say `type Maybe = \ (a :: Type) -> data Nothing | Just a`, I'll happily change my stance.

simonpj:
> I feel like I have no calibration how serious a TH change is? Can anyone put numbers on how often we have breaking changes in TH? My feeling until know was that it has to be relatively common. Also, even if the change is technically breaking, how much users would be affected given that probably most usages of TH don't use the whole grammar?

It's hard to quantify this.   But people complain, legitimately, when we break their code, so we need a Good Reason to do so. Other things being equal, not-breaking is better than breaking.  I'm not arguing that we should never make this change to TH; just that we should await a concrete reason to do so, since a no-op on the TH front is a viable alternative.

simonpj:
Dear @int-index, as you know I wrote to the steering committee a couple of weeks ago with a concrete proposal that they accepted.  It appears below.  I declare this plan accepted.

Would you like to update the MR to reflect these decisions, and then we can merge it?  (By all means add some "alternative designs" to the end of  #425, to reflect the not-taken paths, if you like.)

Thanks!  

Simon


In [proposal #641](https://github.com/ghc-proposals/ghc-proposals/pull/641) Vlad proposes to amend [proposal #425 ](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0425-decl-invis-binders.rst)to permit more wildcard binder forms in type declarations:

We have since exchanged some email, and I invite you to look at my latest post on the ticket, namely:
      https://github.com/ghc-proposals/ghc-proposals/pull/641#issuecomment-2091099038

In the light of these exchanges  I now propose:

* Add wildcards to all binders in type declarations, not just @_.
* Allow kind annotations for wildcards

These changes just fill out the design of #425 in two obvious ways.  In particular:
```
data T1 @_         = MkT1  -- Allowed by #425
data T2 @(_::Nat)  = MkT2  -- Not currently allowed by #425
data T3 _          = MkT3  -- Not currently allowed by #425
data T4 (_ :: Nat) = MkT4  -- Not currently allowed by #425
```
However I propose that we reject
* Do all this recursively
* Make changes to Template Haskell

There is no obvious user need for
```
  data T2 @((k::Nat)::Nat) = MkT2
```
New recursive structures just add (modest but palpable) complexity to the specification and to the implementation.  It is a cost without (so far as I can see) a user need or tangible benefit.

As to Template Haskell, neither #425 nor the above enhancement forces us to make a change to TH.  Currently binders in TH data type declarations cannot have wildcards.  There is no pressing reason to introduce them: one can always use a type variable instead.  And our General Rules ask us to avoid making breaking changes without a strong reason to do so.

simonpj:
Meanwhile I'll put it back into "needs revision", but it need not stay there long!

int-index:
> Meanwhile I'll put it back into "needs revision", but it need not stay there long!

I've revised the amendment, please check if the new syntax matches what you had in mind.

simonpj:
> I've revised the amendment, please check if the new syntax matches what you had in mind.

Looks good to me.  

You say nothing about TH.  I know that other proposals typically don't do so, but it did form part of our debate.   It's hard to see what to say, where, especially given that this is a diff to an existing proposal that also says nothing about TH.  So I suppose we can just leave that.

TL;DR let's merge this.  @adamgundry could you do that -- I'm not sure of the precise steps.

Thanks @int-index .

Issue #478, opened 2021-12-23
Why template-haskell does not provide basic QuasiQuoters?


brandonchinn178:
Is there a reason why `template-haskell` doesn't provide QuasiQuoters corresponding to TH quotes `[e| ... |]`, `[d| ... |]`, etc.? A lot of packages provide QuasiQuoters that could be written as "`[e| ... |]` + extra stuff"; for example, a text interpolation library could be implemented as
```hs
-- the QuasiQuoter corresponding to [e| ... |]
eQQ :: QuasiQuoter

-- Interpolate: (let x = 1 in [text|x is: #{x}|]) == "x is: 1"
text :: QuasiQuoter
text =
  eQQ
    { quoteExp =
        appE [| concat |]
        . listE
        . map
          ( \case
            Left raw -> litE $ stringL raw
            Right interpolated -> quoteExp eQQ interpolated
          )
        . parse
    }

-- Parses "x is: #{x}" into [Left "x is: ", Right "x"]
parse :: String -> [Either String String]
```
Whereas currently, it'd have to pull in `haskell-src-meta` to parse the string.

As another example, we have a use case for easily writing lists of hlists:
```hs
rows =
  eQQ
    { quoteExp = listE . map (toHList . quoteExp eQQ) . lines
    }

-- generates [1 :& "hello" :& True :& Nil, 2 :& "world" :& False :& Nil]
[rows|
  [1, "hello", True]
  [2, "world", False]
|]
```
Today, the best way to do this without `haskell-src-meta` is to do
```hs
rows = listE . map toHList

$( rows
     [ [| [1, "hello", True] |]
     , [| [2, "world", False] |]
     ]
 )
```

Related: #260

treeowl:
This is out of scope for a GHC proposal. You can open an issue [for the core libraries committee](https://github.com/haskell/core-libraries-committee).

brandonchinn178:
ah my bad

phadej:
FWIW, https://github.com/haskell/core-libraries-committee/blob/main/PROPOSALS.md is only for `base` changes.

[README](https://github.com/haskell/core-libraries-committee#suggesting-changes-to-core-libraries) says

[...]

PR #642, opened 2024-03-20
DownRanked Existentials

VitWW:
**DownRanked Existentials**

This proposal adds handiness and flexible control to existential types by using `DownRanked Existentials` .

An alternative to #473 (de facto `UpRanked Existential`)

[Rendered](https://github.com/VitWW/ghc-proposals/blob/down-ranked-exists/proposals/0000-down_ranked_exists.rst)

### DIFF
--- a/proposals/0000-down_ranked_exists.rst
+++ b/proposals/0000-down_ranked_exists.rst
@@ -1,0 +1,440 @@
+DownRanked Existentials
+==========================
+
+.. author:: Viktor WW
+.. date-accepted::
+.. ticket-url:: 
+.. implemented::
+.. highlight:: haskell
+.. header:: This proposal is `discussed at this pull request <https://github.com/ghc-proposals/ghc-proposals/pull/642>`_.
+.. sectnum::
+.. contents::
+
+This proposal introduces down-ranked existentials into GHC
+
+.. _`#473`: https://github.com/ghc-proposals/ghc-proposals/pull/473
+.. _`#81`:  https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0081-forall-arrow.rst
+.. _`#281`: https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0281-visible-forall.rst
+.. _`#281rd`: https://ghc-proposals.readthedocs.io/en/latest/proposals/0281-visible-forall.html
+.. _`#378`: https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0378-dependent-type-design.rst
+.. _`#378rd`: https://ghc-proposals.readthedocs.io/en/latest/proposals/0378-dependent-type-design.html
+
+
+Motivation
+----------
+
+Richly typed programming invariably uses its share of existential types, and this proposal makes it vastly easier to work with existentials.
+
+Alternative Proposal is "First-class existential types" `#473`_ and was added as ``ExistentialQuantification`` extension.
+
+It suggests to add *de facto* **UpRanked Existential**
+
+- **UpRanked Existential** rule: Any N-Ranked ``forall`` *type_variable* is INSTEAD (N+1)-Ranked ``exists`` *type_variable* 
+::
+
+  --UpRanked Existentials:
+  cmpg :: forall a. a -> a ->(forall b. b -> b -> Bool) -> Bool
+  cmpg x y f = f x y
+
+  -- same as
+  cmpg :: forall a. a -> a ->((exists b. b -> b) -> Bool) -> Bool
+
+  -- same as
+  cmpg :: (exists a. a -> a ->((exists b. b -> b) -> Bool)) -> Bool
+
+Unfortunately, working with High-Ranked Types is complicated in Haskell and it's Core.
+
+This Proposal suggest to add the opposite: **DownRanked Existential**  
+
+- **DownRanked Existential** rule: Any N-Ranked (``forall`` or ``exists``) *type_variable* is ALSO (N-1)-Ranked ``exists`` *type_variable* 
+::
+
+  --DownRanked Existentials:
+  cmpg2 :: forall a. a -> a ->(forall b. b -> b -> (b, Bool)) -> (a, Bool)
+  cmpg2 x y f = f x y
+
+  -- same as (same-name sugaring version)
+  cmpg2 :: forall a. exists b. a -> a ->(forall b. b -> b -> (b, Bool)) -> (b, Bool)
+
+  -- same as (1-variable sugaring version)
+  cmpg2 :: forall a. exists c <- b. a -> a ->(forall b. b -> b -> (b, Bool)) -> (c, Bool)
+
+  -- same as (full version)
+  cmpg2 :: forall a. exists c | c <- b. a -> a ->(forall b. b -> b -> (b, Bool)) -> (c, Bool)
+
+If ∀a: f a *then* ∃b: ∀a, a ≡ b : f b
+
+*Note: using same keyword "exists" for both UpRanked and DownRanked Existential Quantifiers is incompatible and inconsistent idea*
+
+
+Proposed Change Specification
+-----------------------------
+
+Roles
+~~~~~
+
+1. Direct-catch a Data-quantifier 
+::
+
+  -- hidden existential type
+  data Box = forall a. MkBox a
+
+  -- open existential type
+  data exists a. Ex = forall a. MkEx a  -- NEW!
+
+2. Extractor from Data-quantifier 
+::
+
+  fromBox :: Box -> ???
+  fromBox (MkBox x) = x       -- Error!
+
+  fromEx :: exists a. Ex -> a
+  fromEx (MkEx x) = x         -- OK! NEW!
+
+3. Indirect Data-existential 
+::
+
+  toBox :: forall a. a -> Box
+  toBox = MkBox
+
+  toEx :: forall a. a -> exists b. Ex
+  toEx = MkEx                 -- NEW!
+
+4. Direct Non-data catch and extract 
+::
+
+  upd :: forall a. exists b. a ->(forall b. b -> b) -> b
+
+5. Existential Boundaries are the same as a escaper type variable boundaries 
+::
+
+  data Doc = forall a. Show a => MkDoc a
+  
+  data exists a. Show a => DocE = forall a. Show a => MkDocE a  -- NEW!
+
+
+Extension
+~~~~~~~~~
+
+Introduce a new extension -XDownRankedExistential.
+
+1. Introduce a new extension ``-XDownRankedExistential``.
+
+#. With ``-XDownRankedExistential``, ``exists`` is a keyword in both types and terms.
+
+#. With ``-XDownRankedExistential``, introduce a new type for existentials.
+
+
+Syntax
+~~~~~~
+
+Syntax of ``exists`` quantifier has 3 forms.
+
+1. The **Full form** is 
+::
+
+  exists a1 a2 a3 | a1 <- b1, a2 <- b2, a3 <- b3.
+
+Where ``| ... <- ...`` is a binding part.
+
+2. Sugared **Same-name form**. If for some ``N`` we have same names ``aN == bN`` then we could omit to write this specific binding of type variable. 
+::
+
+  exists a1 a2 a3. 
+
+  -- desugars into
+  exists a1 a2 a3 | a1 <- a1, a2 <- a2, a3 <- a3.
+
+
+  -- partial same-name
+  exists a1 a2 a3 | a1 <- b1.
+
+  -- desugars into
+  exists a1 a2 a3 | a1 <- b1, a2 <- a2, a3 <- a3.
+
+3. Sugared **One-variable form**. If ``exists`` has just one variable we could write binding shortly. 
+::
+
+  exists a1 <- b1.
+
+  -- desugars into
+  exists a1 | a1 <- b1.
+
+Grammar
+~~~~~~~
+
+1. The grammar is modified as follows (baseline: GHC's parser)::
+
+        ctype → quantifiers_telescope ctype   -- NEW!
+              -- forall_telescope ctype       -- REMOVE!
+              | context '=>' ctype
+              | ...
+
+        quantifiers_telescope → exists_telescope forall_telescope -- NEW!
+
+        -- just for comparison
+        forall_telescope → 'forall' tv_bndrs '.'
+                         | 'forall' tv_bndrs '->'
+                         | {- empty -}
+
+        -- NEW!
+        exists_telescope → 'exists' tv_bind '.'
+                         | 'exists' tv_bndrs_ex '.'
+                         | {- empty -}
+
+        tv_bndrs_ex → tv_bndr tv_bndrs tv_bind_ex
+                    | {- empty -}
+
+        tv_bind_ex → '|' tv_bind tv_bind_ex_next
+                   | {- empty -}
+
+        tv_bind_ex_next → ',' tv_bind tv_bind_ex_next
+                        | {- empty -}
+  
+        tv_bind → tv_bndr '<-' tv_bndr
+
+2. The grammar is modified for ``data`` declaration too.
+
+
+Examples
+--------
+
+Filter Nat-Vectors
+~~~~~~~~~~~~~~~~~~
+
+We could use boxing/unboxing existential types for Vectors ::
+
+  data Nat = Zero | Succ Nat
+
+  type Vec :: Nat -> Type -> Type
+  data Vec n a where
+    VNil :: Vec Zero a
+    (:>) :: a -> Vec n a -> Vec (Succ n) a
+  infixr 5 :>
+
+  data exists n. VecE a = forall n. MkVecE (Vec n a)
+
+  vec2E :: forall a n. Vec n a -> exists m. VecE a
+  vec2E = MkVecE
+
+  vecEFrom :: forall a. exists m. VecE a -> Vec m a
+  vecEFrom (MkVecE x) = x
+
+  fromList :: forall a. [a] -> exists n. VecE a
+  fromList []     = vec2E VNil                
+  fromList (x:xs) = vec2E $ x :> vecEFrom $ fromList xs
+
+  filter :: forall a n. (a -> Bool) -> Vec n a -> exists m. VecE a
+  filter p VNil = vec2E VNil
+  filter p (x :> xs)
+    | p x       = vec2E $ x :> $ vecEFrom $ filter p xs
+    | otherwise = filter p xs  
+
+Poly-existentials
+~~~~~~~~~~~~~~~~~
+
+Poly-existentials data ::
+
+  -- Sum-Type existential
+  data exists a b. ExEither = forall a. MkExLeft a | forall b. MkExRight b
+
+  -- Partly existential
+  data exists a. ExLeftEither = forall a. MkExLeft a | forall b. MkExRight b
+
+
+  -- Phantom-existential Type
+  data exists a. L = forall a. Con a (exists b. L) | Nil
+
+  -- Head, next-to-Head existential
+  -- we catch `b` twice and not from `forall`, but from `exists`
+  data exists a b. L = exists b. forall a. Con a (exists b c. L) | Nil
+
+  -- Head-next-next existential
+  data exists a b c. L = exists b c. forall a. Con a (exists b c d. L) | Nil  
+
+Poly-existentials could have an ambiguity existential-errors :: 
+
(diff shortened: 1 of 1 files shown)


VitWW:
Typical code with `DownRankedExistentials`  looks like this:

```haskell
data Nat = Zero | Succ Nat

type Vec :: Nat -> Type -> Type
data Vec n a where
  VNil :: Vec Zero a
  (:>) :: a -> Vec n a -> Vec (Succ n) a
infixr 5 :>

data exists n. VecE a = forall n. MkVecE (Vec n a)

vec2E :: forall a n. Vec n a -> exists m. VecE a
vec2E = MkVecE

vecEFrom :: forall a. exists m. VecE a -> Vec m a
vecEFrom (MkVecE x) = x

fromList :: forall a. [a] -> exists n. VecE a
fromList []     = vec2E VNil
fromList (x:xs) = vec2E $ x :> vecEFrom $ fromList xs

filter :: forall a n. (a -> Bool) -> Vec n a -> exists m. VecE a
filter p VNil = vec2E VNil
filter p (x :> xs)
  | p x       = vec2E $ x :> $ vecEFrom $ filter p xs
  | otherwise = filter p xs
```

AntC2:
Thank you @VitWW. I'm afraid I'm not following whether the proposal here is just alternative syntax for the same behaviour as #473 proposes, or whether there's some different semantics. I have to say your code examples seem more cluttered that #473's corresponding examples. Specifically why do declarations use two variables where #473 seems to use one? as [here](https://github.com/VitWW/ghc-proposals/blob/down-ranked-exists/proposals/0000-down_ranked_exists.rst#syntax)

> `exists a1 <- b1.`

Re current existential types in GHC (which are not "first-class"), I think #478 is not changing anything around them: we have GADT syntax for the `data`; we pattern match on the constructor to make available the wrapped value at the exposed type. So then what is your `Box` example trying make less "hidden"?:

> ```
> -- hidden existential type
> data Box = forall a. MkBox a
> ```

Ok, this is current syntax with [`ExistentialQuantification`, implies `ExplicitForAll`](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/existential_quantification.html). Although it uses `forall`, I know this is existential because the bound `a` doesn't escape to LHS of the declaration. The type for the constructor is `MkBox :: forall a. a -> Box` -- again I know `a` is existential because it doesn't appear in the result type `Box`. Then I prefer to declare that more explicitly with GADT syntax (and as is used in #473 -- and I think all `data` 'advanced' Haskell extensions avoid H98 syntax as being too inexplicit):

```
data Box  where MkBox :: forall a. a -> Box
```

> ```
> -- open existential type
> data exists a. Ex = forall a. MkEx a  -- NEW!
> ```

Now I'm confused. Is that the same `a` but <s>declared</s> bound twice? The `forall a`'s scope has escaped to LHS. Is this following a model from some other programming language or notation, and which doesn't cause [variable shadowing/masking](https://en.wikipedia.org/wiki/Variable_shadowing)? Can I express this using GADT syntax? Can I write:

```
data exists a. Ex2 = forall b. MkEx2 b
```

Do I now have an existential phantom `a`; and a conventional GHC existential `b`?

Note [Point 3. sub-Point 5. in #473 ](https://github.com/goldfirere/ghc-proposals/blob/existentials/proposals/0473-existentials.rst#proposed-change-specification). I can't pretend I understand that restriction, but I presume it's something to do with the way the program must unwrap a datatype/constructor via pattern matching to expose the existential in a type-safe way.

VitWW:
> Thank you @VitWW. I'm afraid I'm not following whether the proposal here is just alternative syntax for the same behaviour as #473 proposes, or whether there's some different semantics.

Thanks!
I mention, that  " #473 (de facto UpRankedExistentials)", but this is "DownRanked Existentials" Proposal.
So not, this is not "alternative syntax", it is different Existential type with different Existential quantifier!

>Specifically why do declarations use two variables where 473 seems to use one? as **here**

> > `exists a1 <- b1.`

Ok, this is unnecessary part, I'll remove it in next commit. Just simple use `exists a1.`

> > `-- open existential type`
> > `data exists a. Ex = forall a. MkEx a`
>
> Now I'm confused. Is that the same a but ~declared~ bound twice?

Yes, exactly! This is the core idea.
The rule: *if* ∀a: f a *then* ∃b: ∀a, a ≡ b : f b
This existential type is Dependent from Higher-Ranked (Exactly N+1 Ranked) `forall`.

> Can I express this using GADT syntax? Can I write:
> `data exists a. Ex2 = forall b. MkEx2 b`
>
>Do I now have an existential phantom a; and a conventional GHC existential b?

Yes, you could even create `data exists a. UnitE = MkUnit` with phantom existential type.  This existential type just "unboxible" (or bottom-unboxible) while normal existential type could unbox:
```haskell
fromEx :: exists a. Ex -> a
fromEx (MkEx x) = x         -- OK! NEW!
```

LeanderK:
i am also not entirely following the proposal (and what's the difference to the other). Your examples capture my main need...something like:

```haskell
fromList :: forall a. [a] -> exists n. Vec n
fromList ....
```
But in your example there's a constant wrapping and unwrapping going on? Like in the type-signature:
```haskell
filter :: forall a n. (a -> Bool) -> Vec n a -> exists m. VecE a
```
and especially this code
```haskell
vec2E $ x :> vecEFrom $ fromList xs
```
I don't see why this is such an improvement from the status quo tbh. Yes, we get `exists` in the type-signature but it pollutes the code with the wrapping all the time. I think this will get old really quickly with `Vec n` types and then we're back to where we started, longing for a nice solution. And since it's incompatible with a nice solution we have painted ourselves into a corner, otherwise I would have said at least it's something. I would rather wait for some proper solution (without the plumbing), are a proposal that is at least forward-compatible with a proper solution.

VitWW:
> But in your example there's a constant wrapping and unwrapping going on? Like in the type-signature:
> and especially this code
> 
> ```haskell
> vec2E $ x :> vecEFrom $ fromList xs
> ```

Yeah, It's look like existential Monad and maybe it is not a joke )))
In simplified form the return type is wrapped: `filter :: p -> a -> m a`
So yes, we need to unwrap, calculate and wrap again.

Why do we do this? Because this function has 3 different return types `Vec 0 a`, `Vec n a`, `Vec (n-1) a`.
So, we must hide second argument-type into existential type `exists m. VecE a`
```haskell
filter :: forall a n. (a -> Bool) -> Vec n a -> exists m. VecE a
filter p VNil = MkVecE VNil    -- :: Vec 0 a
filter p (x :> xs)
  | p x       = MkVecE $ x :> $ vecEFrom $ filter p xs  -- :: Vec n a
  | otherwise = filter p xs     -- :: Vec (n-1) a
```

> I don't see why this is such an improvement from the status quo tbh. Yes, we get `exists` in the type-signature but it pollutes the code with the wrapping all the time. 

I also love magic. )) 
But I think it is an optimization problem, and with some Rule-pragmas this could become better

> Also, since it's `data`, wouldn't it have a runtime cost?

Good question!
At least now, "A newtype constructor cannot have existential type variables".
I think in theory it is possible to coerce `Vec n a <<->> exists m. VecE a`, but I'm not so sure.

AntC2:
> `exists` in the type-signature but it pollutes the code with the wrapping all the time.

I think this comes with the territory. #473 uses exactly the same technique. And it's the same as in current GHC how you can't 'export' the types inside a GADT: [_pattern matching causes type refinement_](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/gadt.html#:~:text=The%20key%20point%20about%20GADTs%20is%20that%20pattern%20matching%20causes%20type%20refinement.). I also see on that GADTs page, and as in the ICFP paper behind #473, the vast amount of type algebra needed to guarantee type safety.

In contrast here, we have a few scratchy 'rule's of predicate logic. Firstly, those look wrong to me; secondly we will need a comprehensive proof they preserve type safety -- oh, and bring along evidence of the type derivation.

> The Main rule: if ∀a: f a then ∃b: ∀a, a ≡ b : f b
> Second rule: if ∃a: f a then ∃b: ∃a, a ≡ b : f b

What's the `≡` doing there? It's type unification? If `a` unifies with `b`, then we can substitute one for the other? That seems emm unhelpful/merely clutter that introduces another type variable. I apologise for not being George Boole nor Aristotle nor Chrysippus.

VitWW:
> What's the `≡` doing there? It's type unification? If `a` unifies with `b`

No, this is just equality (set equality) `a` and `b`.

> In contrast here, we have a few scratchy 'rule's of predicate logic. Firstly, those look wrong to me; secondly we will need a comprehensive proof they preserve type safety -- oh, and bring along evidence of the type derivation.

Yes, I think that type safeness guarantee comes from 2 rules, I've described.
But this is a bit complicated to me to write a formal and full proof of that.
So, everyone is invited to discussion.

VitWW:
> But in your example there's a constant wrapping and unwrapping going on? 
> and especially this code
> 
> ```haskell
> vec2E $ x :> vecEFrom $ fromList xs
> ```
> Yes, we get `exists` in the type-signature but it pollutes the code with the wrapping all the time. 
> Also, since it's `data`, wouldn't it have a runtime cost?

@LeanderK I was not pay attention about next: yes, functions `filter` and `fromList` constantly use wrapping and unwrapping.
But, this wrapping and unwrapping is O(1) operation and it is not recursive.
So, yes, wrapping and unwrapping has a runtime cost, but is is quite minimal for `filter` and `fromList`.

VitWW:
I would like to submit this Proposal to the committee, @adamgundry

proposals/0000-down_ranked_exists.rst
```diff
@@ -0,0 +1,462 @@
...
+~~~~~
+
+1. Boxing type variable into a Data declaration 
+::
+
+  -- hidden conventional existential GHC type
+  data Box = forall a. MkBox a
+
+  -- open existential type
+  data exists a. Ex = forall a. MkEx { unEx :: a }  -- NEW!
```
maralorn, line 93: Why is `Ex` quantified with `exists a.` but has no free `a`?
Why do you shadow `a` on the right hand side?

proposals/0000-down_ranked_exists.rst
```diff
@@ -0,0 +1,462 @@
...
+
+Main and second rules guarantee us that unboxing give us same type as boxing or it is bottom type (for phantom or partly phantom existentials)
+
+3. Direct capture type variable by Data-constructor
+::
+
+  toBox :: forall a. a -> Box
+  toBox = MkBox
+
+  toEx :: forall a. a -> exists a. Ex
```
maralorn, line 117: Again, why shadowing `a` and quantifying with `exists a` when `Ex` contains no free `a`?

proposals/0000-down_ranked_exists.rst
```diff
@@ -0,0 +1,462 @@
...
+  toEx = MkEx
+
+4. Absorption (indirect capture) different types into one inner type
+::
+
+  fromEither :: forall a b. Either a b -> exists c. Ex
```
maralorn, line 123: Same

maralorn (commented):
@VitWW 

I am sorry, but I cannot follow your proposal at all.

The "changed specification" section does not make clear to me what actually is to be changed. It introduces syntax by example with no obvious interpretation to me and it does not say much about the semantics.

#473 is tackling a difficult problem, is quite long and intricate and backed by an ICFP paper. It discusses implementation of existential types by changing core which is a big step. Your proposal does not say anything about a needed change to core and instead just says "implementation unclear"

Implementing a feature like this is not just about inventing a new syntax, we also need semantics and at least a solid argument that these semantics can be implemented.

Besides the fact that the proposal needs to be made a lot clearer, before the committee can consider your proposal you need to explain how your proposal is either a lot less expressive or how you sidestep the implementation difficulties of existentials.

VitWW:
Thank you!
@maralorn I didn't see [Pending shepherd recommendation](https://github.com/ghc-proposals/ghc-proposals/labels/Pending%20shepherd%20recommendation) label, but I think you are the shepherd of this Proposal. Am I right?

> Your proposal does not say anything about a needed change to core and instead just says "implementation unclear"

I'm a bit confused. "Implementation Plan" part of proposal describes **who** will implement the change if accepted, but not **which lines**.

> I am sorry, but I cannot follow your proposal at all.

I'm not sure what you don't understand. Ok, I describe here in short what do I want to achieve and how, and I hope we find common language to make this Proposal more clear.

1) I've showed that #473 has common motivation with my Proposal and common purpose, but they are different in approaches, use different quantifiers and different rules. I've shown that my Proposal is not just a syntax sugar over #473.

2) We already could create existential types with `ExistentialQuantification` extension:
```haskell
data Box = forall a. MkBox a
```

3) Let we have `Either a b` type and we wish to unbox it.
```haskell
fromEither :: forall a b. Either a b -> ???
fromEither (Left x)  = x
fromEither (Right y) = y
```
We have several ways to do this, but we choose to "absorb" types with existential `Box` type
```haskell
fromEither :: forall a b. Either a b -> Box
fromEither (Left x)  = MkBox x
fromEither (Right y) = MkBox y
```
4) I wish to change `Box` type to be unboxible, but preserve "absorption rule".
```haskell
data exists a. Box = forall a. MkBox a

fromEither :: forall a b. Either a b -> exists c. Box
fromEither (Left x)  = MkBox x
fromEither (Right y) = MkBox y

fromBox :: exists a. Box -> a
fromBox (MkBox x) = x
```
In brief, that's all in this Proposal in abilities.

5) How to achieve this? I've describe 2 rules:
The Main rule : *if* ∀a: f a *then* ∃b: ∀a, a ≡ b : f b
Second rule __: *if* ∃a: f a *then* ∃b: ∃a, a ≡ b : f b

Both of them guarantee, that `exists a` capture **specific type of type variable** from +1 Ranked `forall` (or `exists`) if it is not phantom. And it is always possible to unbox it.

6) If existential is phantom, then unboxig is always bottom ⊥type

I hope all 6 points of my explanation are clear. Don't hesitate, ask what is unclear.

proposals/0000-down_ranked_exists.rst
VitWW: 1) We don't shadow right `a`, but we capture right `a`
2) We have 2 alternative options: `exists a. Box` and `exists a. Box a`. I'm not against both of them. And I'm in favor in second one. 
Main criteria to chose - to preserve "absorption" rule.

maralorn:
I can’t get into the details of your response right now. But to clear up the confusion: No, I am not the shepherd for this proposal. None has been assigned, afaict.

VitWW:
> No, I am not the shepherd for this proposal.

Ok. Anyway your feedback is welcome and I'm glad to see your response!

maralorn:
> I'm a bit confused. "Implementation Plan" part of proposal describes **who** will implement the change if accepted, but not **which lines**.

You are right, that part of the proposal template is normally about who implements the proposal.
The statement "unclear" just resonated with me, because it was unclear to me _how_ this proposal would be implemented.

On the rest of your points, I won’t go into too much detail, I am not an expert on the topic. I just had another look at #473, and I feel like topics like "how does this play with constraints", "how does this compile to core" and "how does this affect type inference" are essential questions which your proposal does not say anything about.

I will stand by my point that `data exists a. Box = forall a. MkBox a` is weird syntax. I suspect your intuition is, that this should basically mean the same as `data Box = forall a. MkBox a` but you want the `a` be "visible" to the outside.
For me that would maybe represented by  something like `data exists a. Box a = MkBox a`? But then I wonder why the datatype at all. Then we could just write `exists a. a`.

My main question is how you think a type like `exists a. Box` can be used. You present two "rules". I guess I can see how they could be use to introduce an `exists` but I don’t see how they can eliminate one.

simonpj:
I had a look at the proposal.  I'm afraid that I could not understand it at all.  I'm not sure what problem it solves, or how it solves it.

One way to tackle this difficulty might be as follows.  I'm 99% sure that the proposal would require a change to GHC's Core language.  That's a big deal!  (The same was true about #473.)  But the good thing is that the Core language is
* explicitly typed, so no type inference questions
* very small, so much easier to specify changes than specifying changes to Haskell.

Moreover, Core already has a specification, [here](https://gitlab.haskell.org/ghc/ghc/-/blob/master/docs/core-spec/core-spec.pdf).  So a good way forward would be to write the new syntax of types, and the new typing judgements, for Core, using the existing doc as a baseline.  That's what Richard did in his paper.

It may not be easy to do this, but  this proposal is not like a superficial syntax change to the surface language; it's a deep fundamental change to the type system of Core, and as such it needs a formal description before we can go far with it.  Type systems are hard!

VitWW:
> @maralorn  I will stand by my point that `data exists a. Box = forall a. MkBox a` is weird syntax. I suspect your intuition is, that this should basically mean the same as `data Box = forall a. MkBox a` but you want the `a` be "visible" to the outside.
> For me that would maybe be represented by something like `data exists a. Box a = MkBox a`? But then I wonder why the datatype at all. Then we could just write `exists a. a`.

Ok, I get it, wierd. So I've changed into `data exists a. Box a = forall a. MkBox a`

> @AntC2  If you feel the 7 points add explanation, please merge them into the proposal text so readers get a consistent picture.

Ok, I merge them.

> You've put 'UpRanked Existential' in bold like it's a technical term. The term doesn't appear in 473, nor in the ICFP paper behind it, nor does Google find any hits. The proposal then contrasts 'DownRanked Existential' also in bold like it's a technical term. But no Google hits for that either. Please either give in words a definition for both those terms (not just some code that's supposed to illustrate whatever they mean); or (preferably) take those 'technical terms' out of the proposal altogether and use words that will help readers understand.

... But I **write explicitly in words** what do I mean by these terms:

UpRanked Existential rule: Any N-Ranked `forall` type_variable is INSTEAD (N+1)-Ranked `exists` type_variable

DownRanked Existential rule: Any N-Ranked `forall` (or `exists`) type_variable is ALSO (N-1)-Ranked `exists` type_variable

Do you see:
N-Rank -> (N+1)-Rank -> upRanked
N-Rank -> (N-1)-Rank -> downRanked

adamgundry:
> I would like to submit this Proposal to the committee, @adamgundry

Apologies for the delay in responding here. In the light of a brief review and the comments so far (e.g. https://github.com/ghc-proposals/ghc-proposals/pull/642#pullrequestreview-2028145186, https://github.com/ghc-proposals/ghc-proposals/pull/642#issuecomment-2084749162), I think this proposal would need substantial further work before being ready for committee consideration. I'm therefore not going to assign a shepherd at this point. Writing a well-motivated and clearly specified proposal is a difficult task, unfortunately, and I'm sorry I don't see a way to move this forward.

VitWW:
Agree, there are some things to think about

VitWW:
@maralorn @simonpj 
Ok, I've add section about changes to Core language:

We wish to write next:
```haskell
data Box = forall a. MkBox a

fromBox :: Box -> exists a. a
fromBox (MkBox x)  = x
```

So I suggest to add existential type:
```haskell
Type → n
     | Type1 → Type2
     | Type1 Type2
     | ∀ n. Type
     | ∃ n, Type1. Type2  -- NEW!
     | ...
```

We add Dissolve expression typing rule:
```haskell
Γ; ├`tm e : ∃α^κ, β^κ. τ
----------------------------
Γ; ∆ ├`tm e : τ[ α^κ → β^κ ]

--Tm DissolveExType
```

And we several lifting rules to Existential type:
```haskell
Γ ├`tm e : τ
Γ ├`tm τ : k
-------------------------
Γ; ∆ ├`tm e : ∃α^κ, τ. α

--Tm LiftTyEx

Γ ├`tm e : τ1 τ2
Γ ├`tm τ1 : k
-----------------------------
Γ; ∆ ├`tm e : ∃α^κ, τ1. α τ2

--Tm LiftExTyApp1

Γ ├`tm e : τ1 τ2
Γ ├`tm τ2 : k
-----------------------------
Γ; ∆ ├`tm e : ∃α^κ, τ2. τ1 α

--Tm LiftExTyApp2

Γ ├`tm e : τ1 → τ2
Γ ├`tm τ1 : k
-------------------------------
Γ; ∆ ├`tm e : ∃α^κ, τ1. α → τ2

--Tm LiftExTyFun1

Γ ├`tm e : τ1 → τ2
Γ ├`tm τ2 : k
-------------------------------
Γ; ∆ ├`tm e : ∃α^κ, τ2. τ1 → α

--Tm LiftExTyFun2
```

And we add rules for coerce and Constraints binders absolutely same as for `∀ n. τ` type.

Is it some "fatal error" in proposed rules?

VitWW:
Interestingly, but it looks like it would be possible to write next without `ScopedTypeVariables` extension
```haskell
f2 :: forall a. [a] -> [a]
f2 xs = ys ++ ys
   where
     ys :: exists b. [b] 
     ys = reverse xs
```

P.S. @simonpj Do you think my proposed changes ( https://github.com/ghc-proposals/ghc-proposals/pull/642#issuecomment-2136029543 ) to Core Language make sense ?

